PR #411, opened 2018-02-20
Derive more traits for events::ModifiersState

jturner314:


tomaka:
Thanks! Please add an entry in the changelog as well!

jturner314:
Okay, I added an entry to the changelog. Thanks for creating `winit`; it's making my window-management code simple!

Issue #414, opened 2018-03-04
panic after dropping window on X11


wez:
Possibly related to https://github.com/tomaka/winit/issues/79, but I'm raising this separately because there is no way for an application to trap this fault.

The issue is that the internal maps are no longer consistent / are invalidated after the drop and the event processing uses `unwrap` when accessing them.

Example program:

```
extern crate winit;

fn main() {
    let mut events_loop = winit::EventsLoop::new();

    let _window1 = Some(winit::Window::new(&events_loop).unwrap());
    let mut window2 = Some(winit::Window::new(&events_loop).unwrap());
    let _window3 = Some(winit::Window::new(&events_loop).unwrap());

    let mut num_windows = 3;

    drop(window2.take());

    events_loop.run_forever(|event| {
        match event {
            winit::Event::WindowEvent { event: winit::WindowEvent::Closed, .. } => {
                num_windows -= 1;
                if num_windows == 0 {
                    return winit::ControlFlow::Break;
                }
            },
            _ => (),
        }
        winit::ControlFlow::Continue
    })
}
```

Stack trace:

```
Breakpoint 1, rust_panic () at libstd/panicking.rs:433
433	libstd/panicking.rs: No such file or directory.
(gdb) bt
#0  rust_panic () at libstd/panicking.rs:433
#1  0x00005555556d369f in std::panicking::rust_panic_with_hook () at libstd/panicking.rs:418
#2  0x00005555556d3423 in std::panicking::begin_panic_fmt () at libstd/panicking.rs:349
#3  0x00005555556d3353 in rust_begin_unwind () at libstd/panicking.rs:325
#4  0x0000555555712be1 in core::panicking::panic_fmt () at libcore/panicking.rs:72
#5  0x0000555555712b07 in core::panicking::panic () at libcore/panicking.rs:51
#6  0x00005555555784aa in <core::option::Option<T>>::unwrap (self=...) at /checkout/src/libcore/macros.rs:20
#7  0x000055555557a2cb in winit::platform::platform::x11::EventsLoop::process_event (self=0x7fffffffda40, xev=0x7fffffffd8b8, callback=0x7fffffffd980)
    at /home/wez/.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.11.1/src/platform/linux/x11/mod.rs:360
#8  0x0000555555578e57 in winit::platform::platform::x11::EventsLoop::run_forever (self=0x7fffffffda40, callback=...)
[...]

Issue #259, opened 2017-08-15
Panic in linux::x11::mod.rs::process_event


ktff:
Panic occurs in `unwrap()` for lines like

`let window_data = windows.get_mut(&WindowId(xev.event)).unwrap()`

It is caused by following events: 
1. Create `Window` with `EventsLoop`
2. User creates some events (ex. mouse movement)
3. `Window` is drooped
4. Processing events on `EventsLoop`

Which creates situation where there are events waiting to be processed for a window that doesn't exist any more.

This is strictly minimal example. Real example creates new `Window` in step 3 and it doesn't matter if it's with old or new `EventsLoop`. Also moving 4. before 3. doesn't help because 2. can happen in between processing events and drooping `Window`. 

I haven't looked if other platform implementations suffer from the same issue.

Quick and dirty fix would be to just ignore those events. 
I am willing to implement this.

Relevant backtrace:
```
thread 'Updater[1]' panicked at 'called `Option::unwrap()` on a `None` value', /checkout/src/libcore/option.rs:335:20
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
stack backtrace:
   0: std::sys::imp::backtrace::tracing::imp::unwind_backtrace
             at /checkout/src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::_print
             at /checkout/src/libstd/sys_common/backtrace.rs:71
   2: std::panicking::default_hook::{{closure}}
             at /checkout/src/libstd/sys_common/backtrace.rs:60
             at /checkout/src/libstd/panicking.rs:380
   3: std::panicking::default_hook
             at /checkout/src/libstd/panicking.rs:396
   4: std::panicking::rust_panic_with_hook
             at /checkout/src/libstd/panicking.rs:610
   5: std::panicking::begin_panic
             at /checkout/src/libstd/panicking.rs:571
   6: std::panicking::begin_panic_fmt
             at /checkout/src/libstd/panicking.rs:521
   7: rust_begin_unwind
             at /checkout/src/libstd/panicking.rs:497
   8: core::panicking::panic_fmt
             at /checkout/src/libcore/panicking.rs:71
   9: core::panicking::panic
             at /checkout/src/libcore/panicking.rs:51
  10: <core::option::Option<T>>::unwrap
             at /checkout/src/libcore/macros.rs:22
  11: winit::platform::platform::x11::EventsLoop::process_event
[...]

Issue #13, opened 2016-07-05
Add method to close window


humb1t:
You provided a possibility to create/show/hide/resize window:
`pub fn hide(&self) {
        self.window.hide()
}`
But I sure that 'pub fn close' method is also can be provided.

LFalch:
Can't you just drop the object?

humb1t:
Yes, I can. But this is not the same.

francesca64:
@humb1t in what way should it be different?

Osspial:
It looks like the consensus as per #434 is that we aren't going to add a `close` method, so I'm going to close this issue.

PR #416, opened 2018-03-04
x11: Destroy dropped windows; handle WM_DELETE_WINDOW

### LINKED ISSUE #79: Dropping Window doesn't close it (X11)
Original: https://github.com/tomaka/glutin/issues/805

francesca64:
Fixes #79 #259 #414 #421

This changes the implementation of `Drop` for `Window` to send a `WM_DELETE_WINDOW` `ClientMessage`, offloading all the cleanup and window destruction to the event loop. Unsurprisingly, this entails that the event loop now handles `WM_DELETE_WINDOW` using the behavior that was previously contained in `Window`'s `Drop` implementation, along with destroying the `Window`. Not only does this mean that dropped windows are closed, but also that clicking the × button on the window actually closes it now.

The previous implemention of `Drop` was also broken, as the event loop would be (seemingly permenanently) frozen after its invocation. That was caused specifically by the mutex locking, and is no longer an issue now that the locking is done in the event loop.

While I don't have full confidence that it makes sense for the `Drop` implementation to behave this way, this is nonetheless a significant improvement. The previous behavior led to inconsistent state, panics, and event loop breakage, along with not actually destroying the window.

This additionally makes the assumption that users don't need `Focused` or `CursorLeft` events for the destroyed window, as `Closed` is adequate to indicate unfocus, and users may not expect to receive events for closed/dropped windows. In my testing, those specific events were sent immediately after the window was destroyed, though this sort of behavior could be WM-specific. I've opted to explicitly suppress those events in the case of the window no longer existing.


Ralith:
It's by design that clicking the X doesn't automatically close the window. It's supposed to issue a request, that the application can respond to as appropriate. For example, an application might want to prompt the user to save their work.

francesca64:
@Ralith I agree with you 100% that winit should work that way. However, I'm not convinced winit currently codifies that idea.

For one thing, in the absence of #13, the current API doesn't even enable us to manually close windows (aside from dropping, but I don't see that documented as a solution, and I imagine we want the canonical solution to be more explicit). The [multiwindow example](https://github.com/tomaka/winit/blob/master/examples/multiwindow.rs) seems to suggest that `Closed` indicates a window's already been closed, and [the documentation](https://docs.rs/winit/0.11.1/winit/enum.WindowEvent.html#variant.Closed) isn't explicit about the meaning.

Now, I'm not too familiar with the other backends, but I tried to investigate this. On Windows, it looks like [the window's state is removed at the same time as sending `Closed`](https://github.com/tomaka/winit/blob/4c62d71950077499fba4f0081a14fae59c5b7ede/src/platform/windows/events_loop.rs#L359) and [it's stated here that the `DefWindowProcW` call from above results in the window ultimately being destroyed](https://github.com/tomaka/winit/blob/4c62d71950077499fba4f0081a14fae59c5b7ede/src/platform/windows/window.rs#L307). This is more clear on OS X, where [`find_and_remove_window`](https://github.com/tomaka/winit/blob/4c62d71950077499fba4f0081a14fae59c5b7ede/src/platform/macos/events_loop.rs#L92) is used to remove the window [at the same time as sending `Closed`](https://github.com/tomaka/winit/blob/4c62d71950077499fba4f0081a14fae59c5b7ede/src/platform/macos/window.rs#L76). Of course, I haven't tested the behavior of these, so I could fortunately be wrong in how I'm interpreting this.

Either way, I think there's a definite need for winit to take a more explicit stance on what the intended behavior is. I designed this PR the way I did to make it more consistent with [existing expectations](https://github.com/tomaka/winit/issues/79#issuecomment-336773212) and the other backends, with the mindset that larger API concerns could be ironed out in future work. However, if you're of the mindset that this would be introducing a *de facto* regression, well, I can't dispute that. I can get to work on making a new PR that addresses these concerns.

Ralith:
I can't speak to the state of the other backends, and I'm not in a position to unilaterally reject this approach. Consistency is good, and incremental improvements are worth making.

I would, however, like to note for posterity that "drop = close" is natural from an RAII perspective, and in the overwhelmingly common case of one-window applications that terminate then a close is requested, it Just Works.

potocpav:
This **fixes** a `panic` that I incounter on multiple windows created sequentially:

```rust

extern crate glium;

fn spawn() {
    let mut events_loop = glium::glutin::EventsLoop::new();
    let window = glium::glutin::WindowBuilder::new();
    let context = glium::glutin::ContextBuilder::new();
    let display = glium::Display::new(window, context, &events_loop).unwrap();

        if let glium::glutin::Event::WindowEvent { event: glium::glutin::WindowEvent::Closed, .. } = event {
            return glium::glutin::ControlFlow::Break;
        }
        glium::glutin::ControlFlow::Continue
    });
}

fn main() {
    spawn();
    spawn();
}
```

After closing the first win, this produces on my X11 setup:

```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', libcore/option.rs:335:21
(...)
   9: <core::option::Option<T>>::unwrap
             at /checkout/src/libcore/macros.rs:20
  10: winit::platform::platform::x11::EventsLoop::process_event
             at /home/pavel/.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.11.2/src/platform/linux/x11/mod.rs:719
  11: winit::platform::platform::x11::EventsLoop::run_forever
             at /home/pavel/.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.11.2/src/platform/linux/x11/mod.rs:188
  12: winit::platform::platform::EventsLoop::run_forever
             at /home/pavel/.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.11.2/src/platform/linux/mod.rs:399
  13: winit::EventsLoop::run_forever
             at /home/pavel/.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.11.2/src/lib.rs:231
  14: multi_win::spawn
             at src/main.rs:10
  15: multi_win::main
             at src/main.rs:20
```

francesca64:
I've verified that this is consistent with the behavior on Windows. I haven't had the opportunity to test on OS X, but the code there is easier to interpret, so I'd be surprised if it's not the case there as well.

Considering the difficulty of making multi-platform design changes, combined with the significance of the existing problems in the X11 backend, I'm still in favor of this being merged. I'll also create an issue concerning winit's concept of `Closed`.

thiolliere:
I have an issue using it in vulkano:
I add 
```
[patch.crates-io]
winit = { git = "https://github.com/francesca64/winit", branch = "x11-drop-window" }
```
to the root `Cargo.toml` and then if I run triangle example and close the window then it does close the window but doesn't end the processus.

tomaka:
> I have an issue using it in vulkano:

I'm not sure vulkano is a good example of this, as it stores the window in an `Arc` which may not be dropped when you want.

PR #424, opened 2018-03-19
Try XOpenIM with different locale modifiers

hcpl:
Implements the solution suggested in https://github.com/tomaka/winit/issues/277#issuecomment-337751136.


hcpl:
Please let me know if you think this needs to be handled more gracefully, like returning `CreationError` instead of panicking.

francesca64:
Thanks for tackling this!

Do you think you could rebase [my commit for making XIM optional](https://github.com/francesca64/winit/tree/x11-xim-optional) into your branch? That way, the issue should be completely resolved.

hcpl:
@francesca64 thanks for the commit! I can't build your branch yet though:

```
error[E0308]: mismatched types
   --> src/platform/linux/x11/mod.rs:453:79
    |
453 |                         let mut count = (self.display.xlib.Xutf8LookupString)(window_data.ic, xkev,
    |                                                                               ^^^^^^^^^^^^^^ expected *-ptr, found enum `std::option::Option`
    |
    = note: expected type `*mut x11_dl::xlib::_XIC`
               found type `std::option::Option<*mut x11_dl::xlib::_XIC>`
    = help: here are some functions which might fulfill your needs:
            - .unwrap()

error[E0308]: mismatched types
   --> src/platform/linux/x11/mod.rs:460:75
    |
460 |                             count = (self.display.xlib.Xutf8LookupString)(window_data.ic, xkev,
    |                                                                           ^^^^^^^^^^^^^^ expected *-ptr, found enum `std::option::Option`
    |
    = note: expected type `*mut x11_dl::xlib::_XIC`
               found type `std::option::Option<*mut x11_dl::xlib::_XIC>`
    = help: here are some functions which might fulfill your needs:
            - .unwrap()

error: aborting due to 2 previous errors

error: Could not compile `winit`.
```

Can you fix these first? I'll try to rebase when it'll be ok to do so.

francesca64:
Oh, whoops. I picked that commit out of my libxkbcommon branch, where that section of code doesn't exist. Sorting that out would be more involved than what I'm able to commit to right now, but fortunately, after this PR, I assume it would be rather unusual for users not to be able to get a valid input method.

I'm going to go ahead and give you a proper review now; sorry about that.

src/platform/linux/x11/mod.rs
```diff
@@ -954,10 +954,28 @@ impl Window {
...
 
-            let im = (x_events_loop.display.xlib.XOpenIM)(x_events_loop.display.display, ptr::null_mut(), ptr::null_mut(), ptr::null_mut());
+            let open_im = || (x_events_loop.display.xlib.XOpenIM)(
+                x_events_loop.display.display,
+                ptr::null_mut(),
+                ptr::null_mut(),
+                ptr::null_mut(),
+            );
+
+            let mut im = open_im();
```
francesca64: After doing some research, I believe this call should be preceded by an empty `XSetLocaleModifiers`, [as is done in alacritty](https://github.com/jwilm/alacritty/blob/6debc4f3351446417d0c4e38173cd9ef0faa71d5/src/window.rs#L335). Otherwise, the user's `XMODIFIERS` environment variable is never used. This is also done in [GLFW](https://github.com/glfw/glfw/blob/6e69f63ffa4059d0f16344e6b038285c1ca78c16/src/x11_init.c#L944), [GTK](https://github.com/GNOME/gtk/blob/695d141f32d8aed14b57cc065554974d541f8ba6/gtk/gtkimcontextxim.c#L364), and so on.

jwilm:
Thanks @hcpl and @francesca64 for working on this feature. It looks like all of the issues in this thread have been addressed, and the latest review was an approval. Can it be merged?

PR #430, opened 2018-03-21
Add methods to get the position of a window's client area, relative to the desktop

Osspial:
This differs from the current `get_position`, in that the current method gets the window's top-left position, including window decorations. This adds a `get_inner_position` method, which gets the client area's top-left position, not including decorations, relative to the desktop.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 - Overhauled X11 window geometry calculations. `get_position` and `set_position` are more universally accurate across different window managers, and `get_outer_size` actually works now.
+- Corrected `get_position` on macOS to return outer frame position, not content area position.
 
 # Version 0.12.0 (2018-04-06)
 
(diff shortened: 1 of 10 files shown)


francesca64:
For me under Xfwm4, `get_position` and `get_inner_position` return the same values. However, since that's the result of more fundamental issues with geometry calculation in the X11 backend, I think it's fine for now.

francesca64:
Okay, with #438 merged, this works on X11 now.

@mitchmindtree are you sure they were both outer position? I changed the relevant code (and badgered one of my friends into testing it), and it seems that the existing `get_position` implementation was returning the content area position (judging by both the values returned and the use of `contentRectForFrameRect`). So, I changed `get_inner_position` to be identical to the current `get_position`, and changed `get_position` to just use the frame rect instead of requesting the content rect.

(Also, it seems that on macOS, `set_position` sets the content area position. On X11, we set the outer window position, and I'm not sure about the other backends, but that's a concern for a separate issue.)

francesca64:
Okay, `set_position` works the same way on Windows as it does on X11, and [the docs](https://docs.rs/winit/0.12.0/winit/struct.Window.html#method.set_position) specify that `set_position` uses the same coordinates as `get_position`. I've corrected the implementation of `set_position` on macOS accordingly.

francesca64:
@robsaunders do you think you could take a look at this?

francesca64:
Alright, after getting a chance to test this on macOS hands-on, I've confirmed that the new implementation is correct, and consistent with Windows and X11.

Issue #343, opened 2017-11-05
Clarification on dead keys


k0nserv:
I did some reasearch based on https://github.com/tomaka/winit/issues/263 using a simple macOS app that just logs keyDown events. With my layout(based on Swedish Pro) ^, ¨, ~ don't work correctly since they are dead keys. Handling dead keys involves quite a bit more [work](https://stackoverflow.com/questions/33203020/get-the-char-accent-in-nsevent) that needs to be done in winit. I'm curious if handling dead keys is considered in scope for winit or if that's something that you expect applications themselves to deal with.

Also see https://github.com/jwilm/alacritty/issues/682

vberger:
For comparison, the wayland platforms automatically handles dead keys. By that I mean that for example:

- User presses `^` key (which is a dead key on my azerty keyboard)
- winit generates a `KeyPressed` event for key `^` but no `ReceivedCharacter` event
- User then presses `e` key
- winit generates a `KeyPressed` event for key `e`, then a `ReceivedCharacted('ê')`

tomaka:
What @vberger described is what should happen IMO. That's also currently the behaviour on Windows.

PR #432, opened 2018-03-23
x11: Don't panic when using dead keys

### LINKED ISSUE #366: Deadkeys on Linux panics
On X11, even the most simple program crashes, when using a dead key. For example when inputting `^ + a`, which should result in `â`. I used the german layout. and tested with latest version and with master.:

```thread 'main' panicked at 'attempt to subtract with overflow'```
[`src/platform/linux/x11/mod.rs:415:35`](https://github.com/tomaka/winit/blob/8f18dab06108592b04bda260be30b9b912a67d43/src/platform/linux/x11/mod.rs#L410-L419)

francesca64:
Fixes #366 

I was originally planning to fix this by rewriting keyboard handling to use XInput2+XKB+libxkbcommon, and I do indeed have a fully functional implementation of that. However, I never managed to figure out how to handle IME, and I could seldom find the motivation to try to conquer yet another dimension of X11 eroticism. That only got worse when I discovered older libxkbcommon versions, which are still in use, [don't support compose sequences](https://github.com/francesca64/xkbcommon-dl/issues/1). As a result, this fairly important issue has gone unresolved for much longer than I ever intended.

This PR fixes the issue using a bold new solution: a single `if` that checks if the keycode is non-zero. A `KeyPress` with a keycode of 0 is received at the end of compose sequences and pre-edit sequences, which has our desired UTF8 string associated with it. The panic was simply the result of trying to send a `KeyboardInput` event for this input, which involved trying to calculate a scancode for this nonexistent key. So, now we just don't try to send a `KeyboardInput` event in that case, and just like that, everything works.

I should note that the behavior of the X11 backend isn't consistent with [the expectations laid out in #343](https://github.com/tomaka/winit/issues/343#issuecomment-342106540). While I was able to easily achieve that behavior using XInput2, vanilla X11 seems less cooperative. For individual keys pressed during a compose/pre-edit sequence, X11 only generates `KeyRelease` events, so the user receives `KeyboardInput` events for key releases with no corresponding press events. In the case of compose, the `ReceivedCharacter` is sent to the user before the `KeyboardInput` for the release of the final key in the sequence. For pre-editing, a `KeyboardInput` for the release of the return key is sent after the `ReceivedCharacter`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 - Added subclass to macos windows so they can be made resizable even with no decorations.
+- Dead keys now work properly on X11, no longer resulting in a panic.
 
 # Version 0.11.3 (2018-03-28)
 
--- a/src/platform/linux/x11/mod.rs
+++ b/src/platform/linux/x11/mod.rs
@@ -16,7 +16,7 @@ use std::sync::{Arc, Mutex, Weak};
 use std::sync::atomic::{self, AtomicBool};
 use std::collections::HashMap;
 use std::ffi::CStr;
-use std::os::raw::{c_char, c_int, c_long, c_uchar, c_ulong};
+use std::os::raw::{c_char, c_int, c_long, c_uchar, c_uint, c_ulong};
 
 use libc;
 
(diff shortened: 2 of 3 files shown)


b-r-u:
Thanks for wandering through the arcane landscape of X11 APIs!
Despite the drawbacks from your last paragraph, this looks like a really simple solution to the problem.

Ralith:
> However, I never managed to figure out how to handle IME

If it's any consolation, that's why I gave up on xkbcommon-ifying the main keyboard input code path myself. It may simply be impossible to support anything other than XCompose with that approach, which I imagine eastern users would be rather unsatisfied with. At this point I'm 100% approving of bodging X11 support into something approximately sane and focusing on Wayland for doing it Right™.

francesca64:
The event filtering hack unfortunately had to go. It resulted in backspace and arrow key inputs being doubled, so the original proviso once again applies.

PR #439, opened 2018-03-30
x11: Only access XIM from the event loop thread

francesca64:
XIM isn't thread-safe at all. Any call made to it from another thread will result in the event loop freezing (this is why the old implementation of `Drop` for `Window` had that problem).

XIM is now confined to one thread, and the existing API is maintained using channels. In testing this with Alacritty, I initially thought the occasional slight lag on updating the spot location was due to this change, but it's present without it as well.


francesca64:
Since I didn't specify, this issue is encountered by trying to call `send_xim_spot` from another thread. The reason it seems that no one's ever had this problem is because there's presumably a lack of applications handling windows in separate threads that also support XIM.

PR #442, opened 2018-04-01
Expose get_physical_extents for the MonitorId

fschutt:
At least on Windows, X11, Wayland, macOS and iOS, this new function returns the physical size of the screen in millimeter. On Android, I can't implement it because there is no way for me to the access to the JavaVM. This is waiting on https://github.com/tomaka/android-rs-glue/pull/116. I need this for two reasons:

1. As a poor mans DPI value until proper DPI support is added - for rasterizing the font with freetype, I need the correct screen DPI - see: https://github.com/servo/webrender/issues/2596
2. I am working on a GIS application where correct scaling is important - i.e. I have to know how large the physical monitor is so I can calculate the map scale accordingly. 

I haven't tested that the function actually returns the value in millimeter, though, just that it compiles.


francesca64:
Sorry for the wait.
- This needs a CHANGELOG entry.
- Looks good on X11, and returned values seem correct (though I must admit I didn't check with a ruler!)
- Wayland implementation looks solid, though I've yet to test it.
- I tested this on Windows 10, and it always returns `None`. On the first call, it fails at "reset to whatever DPI awareness was set before", and on all subsequent calls, it fails at "couldn't set to the required DPI awareness, reset and quit". On that note, I'm not really sure what the value is in returning `None` when it can't be reset, since we'd still have usable dimensions, and the process is still in an unexpected state. Also, is there a reason [`VERTSIZE`/`HORZSIZE`](https://msdn.microsoft.com/en-us/library/windows/desktop/dd144877(v=vs.85).aspx) aren't used?
- The logic looks good on macOS (I'll try to test it later), though since [the size values are 0 for an invalid display](https://developer.apple.com/documentation/coregraphics/1456599-cgdisplayscreensize), would it make sense to return `None` in that case?
- Is it possible for me to test the Android implementation if I use the android-rs-glue PR?
- I have no idea about iOS.

fschutt:
Also relevant: https://ofekshilon.com/2011/11/13/reading-monitor-physical-dimensions-or-getting-the-edid-the-right-way/

Windows "smartly" enlarges the monitors size in order to correct for the distance of the eye to the monitor. Which is - in my opinion - a complete shitpile of design. It may work, it may not work. Reading the EDID data from the windows registry seems to be the only proper way to do it, as ridicoulous as this may sound.

Note: this function is named `get_physical_extents()`, it doesn't have anything to do with the actual DPI itself.

fschutt:
It seems that this is GLFW's solution:

```c
    if (IsWindows8Point1OrGreater())
    {
        widthMM  = GetDeviceCaps(dc, HORZSIZE);
        heightMM = GetDeviceCaps(dc, VERTSIZE);
    }
    else
    {
        widthMM  = (int) (dm.dmPelsWidth * 25.4f / GetDeviceCaps(dc, LOGPIXELSX));
        heightMM = (int) (dm.dmPelsHeight * 25.4f / GetDeviceCaps(dc, LOGPIXELSY));
    }
```

https://github.com/glfw/glfw/blob/8d98a6e37d122f45439263c96dd50ac2f16b7f5f/src/win32_monitor.c#L82-L91

francesca64:
I've verified that this works as expected on macOS. For Windows, I assume that's not finished yet.

fschutt:
Yeah, I haven't had much time to work on this... I'm basically trying to port the solution presented here: https://ofekshilon.com/2014/06/19/reading-specific-monitor-dimensions/ - this way it'll work on all versions on Windows and with multiple monitors, too.

src/platform/linux/mod.rs
```diff
@@ -102,6 +102,14 @@ impl MonitorId {
             &MonitorId::Wayland(ref m) => m.get_hidpi_factor(),
         }
     }
+
+    #[inline]
+    pub fn get_physical_extents(&self) -> Option<(u64, u64)> {
+        match self {
```
torkleyy: Should match on `*self`

src/platform/windows/monitor.rs
```diff
@@ -52,39 +56,265 @@ fn wchar_as_string(wchar: &[wchar_t]) -> String {
...
+    use winapi::um::handleapi::INVALID_HANDLE_VALUE;
+    use winapi::um::setupapi::{DIGCF_PRESENT, DIGCF_PROFILE, SP_DEVINFO_DATA};
+    use winapi::um::winnt::KEY_READ;
+    use winapi::um::winreg::RegCloseKey;
+    use winapi::um::cfgmgr32::MAX_DEVICE_ID_LEN;
+
+    // Find the display_device with the correct monitor name
+    let display_device = get_display_device_from_monitor_name(monitor_name)?;
+    let device_id = wchar_as_string(&display_device.DeviceID[..]);
+    println!("device id - {:?}, monitor name: {:?}", device_id, monitor_name);
```
torkleyy: Should be removed or replaced by logging.

francesca64: Thanks for helping out, but this PR is WIP, so I don't think this is pertinent.

(Also, winit has no logging infrastructure whatsoever, so your proposed alternative would be something of an undertaking. If you want winit to have logging sooner, then I recommend opening an issue so everyone has a chance to argue about whether or not to use `slog`. It's something I've been meaning to do, but I've been busy.)

src/platform/ios/mod.rs
```diff
@@ -150,6 +150,23 @@ impl MonitorId {
...
         1.0
     }
+
+    pub fn get_physical_extents(&self) -> Option<(u64, u64)> {
+        unsafe {
+            if setjmp(mem::transmute(&mut jmpbuf)) != 0 {
+                let app: id = msg_send![Class::get("UIApplication")?, sharedApplication];
+                let delegate: id = msg_send![app, delegate];
+                let state: *mut c_void = *(&*delegate).get_ivar("glutinState");
+                let state = state as *mut DelegateState;
```
francesca64: Okay, I've had a chance to work on iOS now. I don't believe this needs to be wrapped in `setjmp`, and you can just use `self.delegate_state` instead of re-retrieving everything.

src/platform/ios/mod.rs
```diff
@@ -150,6 +150,23 @@ impl MonitorId {
...
+                let scale = (*state).scale;
+                let width = (*state).size.0;
+                let height = (*state).size.1;
+                Some(((width as f32 * scale) as u64, (height as f32 * scale) as u64))
```
francesca64: and note that converting a float to an integer doesn't automatically round. Is this what you wanted?

src/platform/linux/x11/monitor.rs
```diff
@@ -86,6 +89,7 @@ pub fn get_available_monitors(x: &Arc<XConnection>) -> Vec<MonitorId> {
                         dimensions: ((*crtc).width as u32, (*crtc).height as u32),
                         position: ((*crtc).x as i32, (*crtc).y as i32),
                         primary: true,
+                        extents_mm: (x_mm as u64, y_mm as u64),
```
francesca64: Fun fact: these can sometimes be 0. #543

fschutt:
Oh I haven't, it's just that the reason why I needed this in the first place, which is webrender, has "fixed itself" because I noticed that I mixed up pixels and points: https://github.com/servo/webrender/issues/2596#issuecomment-391591026

So for my use-case, I don't really need it anymore... I'll definitely wait until proper DPI is done, then maybe I can get some redundancy out.

francesca64:
Proper DPI is over here #548, so it will probably be out in a week or so (though I should know better than to say things like that).

ZeGentzy:
This PR hasn't had any activity for over a year! If this is still relevant, please rebase this PR against master and reopen it. Thanks!

Issue #105, opened 2016-12-24
Handle hidpi scaling.


nical:
The wording here is intentionally vague because I don't know very well what the possibilities are and what makes sens for winit. It is safe to say that it would be good for winit to expose some way to indicate whether the application is "dpi-aware" and avoid having the window upscaled by the window manager for apps that don't want this behavior.

On Fedora 25 with wayland (now enabled by default), winit windows are all upscaled on hidpi screens.

cc @vberger

vberger:
I believe this needs some discussion, as DPI is not handled the same way on all platforms.

The wayland way is:

- Each monitor has a scale factor, defaulting to 1.0, but it can be higher for HiDPI screens (typically 2.0)
- Each window known on which monitor(s) it is displayed, and what their scale factor are
- Each window draws in its buffer with the scale factor of its choice and notify the server about it (if not set, defaults to 1.0)
- If the scale factors of the screen(s) and the one chosen by the application do not match, the server scales the contents of the window appropriately

tomaka:
From what I gathered, on Win32 each thread has a HiDPI-awareness level:

- A window on a thread that is not HiDPI-aware (the default) will be scaled by the operating system.
- A window on a thread that is HiDPI-aware is supposed to query the HiDPI level of the monitor and is assumed to be scaled based on this information. For example if you create a window of size 1280x1024 on a monitor that reports a resolution of 1280x1024 and a HiDPI level 2.0, the window will occupy half of the screen (and if the thread was not HiDPI-aware, the window would occupy the whole screen).
- A window on a thread that is per-monitor-HiDPI-aware is the same as HiDPI-aware, except that it handles changing the current monitor itself. If you are only HiDPI-aware, Windows will automatically scale the window if you move it from one monitor to another that doesn't have the same HiDPI level. But if you are per-monitor-aware, it's the application that is responsible for this.

I couldn't find whether moving a window from a thread to another, or changing the HiDPI awareness level during the execution, modifies the existing windows. The HiDPI awareness is supposed to be set once at initialization.

The older way to deal with that was per application, and you could only set it once (all further calls would fail). Maybe this is also the case for threads.

kryptan:
We need to support 3 things to support high DPI:

1. Notifying OS that we are high DPI aware.
2. Getting scaling factor of the window. There is `Window::hidpi_factor()` function for this.
3. Receive notifications from the OS about change of the scale factor.

On Windows situation is as follows:

# Notifying OS that we are high DPI aware

[...]

PR #443, opened 2018-04-01
[macOS] implement HiDPIFactorChanged event on relevant events:

### LINKED ISSUE #441: Use viewDidChangeBackingProperties on MacOS to detect when the DPI / monitor has changed
I came across this interesting blog post: http://supermegaultragroovy.com/2012/10/24/coding-for-high-resolution-on-os-x-read-this/

It seems that macOS can call an application when the DPI or the monitor changes, given that `viewDidChangeBackingProperties` is implemented. In these cases, glium should emit a [HiDPIFactorChanged](https://docs.rs/winit/0.11.3/winit/enum.WindowEvent.html#variant.HiDPIFactorChanged) event, so that the application knows that it should update the DPI.

sodiumjoe:
* moving window to another screen
* changing resolution of screen

implement #105 for macos
fixes #441 (using `windowDidChangeBackingProperties` instead of `viewDidChangeBackingProperties`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Implement `WindowEvent::HiDPIFactorChanged` for macOS
+
 # Version 0.11.3 (2018-03-28)
 
 - Added `set_min_dimensions` and `set_max_dimensions` methods to `Window`, and implemented on Windows, X11, Wayland, and OSX.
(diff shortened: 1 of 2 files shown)


francesca64:
@trishume would you be able to review this?

francesca64:
Thanks, to both of you!

PR #444, opened 2018-04-02
[Windows] Adjust min_dimensions and max_dimensions using AdjustWindowRectEx.

### LINKED ISSUE #196: Win32: with_dimensions excludes decoration, with_max_dimensions includes them
I have a reproduction case [here that works with the latest master.](https://gist.github.com/LPGhatguy/8d12e3a9e0fad838a881f200e3687c6c)

It looks like, on Win32 at least, `with_max_dimensions` and `with_min_dimensions` include the window decorations, but `with_dimensions` does not. Only including the client area seems like the correct approach here, so this seems like a bug.

I'm taking a look at this.

Lymia:
This fixes #196, although it does not update the documentation as suggested in the comments.

### DIFF
--- a/src/platform/windows/window.rs
+++ b/src/platform/windows/window.rs
@@ -622,6 +622,14 @@ pub struct WindowWrapper(HWND, HDC);
 // https://github.com/retep998/winapi-rs/issues/396
 unsafe impl Send for WindowWrapper {}
 
+pub unsafe fn adjust_size(
+    (x, y): (u32, u32), style: DWORD, ex_style: DWORD,
+) -> (LONG, LONG) {
+    let mut rect = RECT { left: 0, right: x as LONG, top: 0, bottom: y as LONG };
+    winuser::AdjustWindowRectEx(&mut rect, style, 0, ex_style);
+    (rect.right - rect.left, rect.bottom - rect.top)
+}
+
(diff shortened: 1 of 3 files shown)


francesca64:
Sorry for the wait; you're at the top of my list now. Could you rebase/merge with the current master?

Lymia:
Rebased onto master.

Lymia:
I don't know what's going on with AppVeyor. That looks like a sporadic failure.

src/platform/windows/window.rs
```diff
@@ -622,6 +622,14 @@ pub struct WindowWrapper(HWND, HDC);
 // https://github.com/retep998/winapi-rs/issues/396
 unsafe impl Send for WindowWrapper {}
 
+pub unsafe fn adjust_size(
+    (x, y): (u32, u32), style: DWORD, ex_style: DWORD,
```
francesca64: Ooh, I never knew you could destructure a tuple in the function parameters. Thanks for teaching me something new today!

francesca64:
Thanks! Everything looks good, and it worked great when I tested it on Windows 10.

(I won't fuss about not making the documentation change, since the docs sorta need an overhaul soon anyway.)

francesca64:
...whoops, the CHANGELOG entry was under the wrong section from before. I'll have to remember to correct that.

Issue #391, opened 2018-01-27
with_visibility(false) followed by window.show() hangs


gillenc:
On Windows the following code opens a window but the code hangs at the window.show() line.
```rust
extern crate winit;

fn main() {
    let mut events_loop = winit::EventsLoop::new();
    let window = winit::WindowBuilder::new().with_visibility(false).build(&events_loop).unwrap();
    window.show();

    events_loop.run_forever(|event| {
        match event {
            winit::Event::WindowEvent { event: winit::WindowEvent::Closed, .. } => winit::ControlFlow::Break,
            _ => winit::ControlFlow::Continue,
        }
    });
}
```

Issue #94, opened 2016-11-26
Unnecessary usage of thread-local storage on Windows


Osspial:
Currently, the win32 window callback relies on thread-local storage to stash data for the callback function. This isn't necessary - by using the [win32 subclassing API](https://msdn.microsoft.com/en-us/library/windows/desktop/bb773183(v=vs.85).aspx), it is possible to pass a user-defined pointer into the window function with the `dwRefData` parameter in a subclass. By boxing up the callback data and freeing it when the window is destroyed, the use of the `thread_local` macro can be avoided.

ZeGentzy:
Did this get solved or was it decided against? The commit that references this issue doesn't appear at a glance to solve it.

Osspial:
This was fixed by the EventLoop 2.0 rework.

PR #445, opened 2018-04-02
Remove child thread from win32 event loop

### LINKED ISSUE #415: Race condition: run_forever can dispatch events after the window is destroyed
# Background

While working on https://github.com/tomaka/glutin/pull/998, my test program was crashing every time I closed the window:

```
Assertion failed: false, file gfx/angle/checkout/src/libANGLE/renderer/d3d/SurfaceD3D.cpp, line 307
error: process didn't exit successfully: `target\debug\mozangle-with-glutin-example.exe` (exit code: 3221226505)
```

This is ANGLE somewhere inside `eglSwapBuffers`, calling Windows’s `GetClientRect` with the pointer returned by `glutin::Window::platform_window` that Glutin had previously passed to `eglCreateWindowSurface`, and asserting that this call succeeds. Here it was failing, with a debugger showing that `GetLastError()` returned 1400 which is `ERROR_INVALID_WINDOW_HANDLE` and showing indeed the that pointer as "unreadable memory".

# Conditions

Trying to reproduce this without ANGLE, I found out after a while that this problem only happens when:

1. The closure passed to `events_loop.run_forever(…)` does some non-trivial work for each event
2. The cursor is moved over the window, such that a bunch of events are queued
3. The "close window" button is clicked after that, without waiting.

As a result the closure keeps being called to handle events, even though the window is already gone and the `HWND` pointer is invalid.

I don’t know if this problem is Windows-specific or if a similar race condition happens on other platforms.

# Test case

* Steps: run the program below on Windows, mouse over the window for a bit, then close it within a couple seconds.

* Actual result: `GetClientRect failed. GetLastError: 1400` is printed dozens of times.

* Expected result: no output, `GetClientRect` always succeeds. Either the OS window is kept alive until all events are dispatched (or maybe as long as the `winit::Window` exists), or queued events are dropped when the window is destroyed.

  Alternatively, `winit::Window` gains a new method that tests whether the OS window still exists, and docs and examples point out that is should be used before doing anything with the window. (Maybe `glutin::GlWindow::swap_buffers` also uses it to become a no-op on a closed window.)

```rust
extern crate winapi;
extern crate winit;

use winit::os::windows::WindowExt;

fn main() {
    let mut events_loop = winit::EventsLoop::new();

    let window = winit::WindowBuilder::new()
        .build(&events_loop)
        .unwrap();

//        println!("{:?}", event);

        match event {
            winit::Event::WindowEvent { event: winit::WindowEvent::Closed, .. } => {
                return winit::ControlFlow::Break
            },
            _ => {}
        }
        unsafe {
            let hwnd = window.get_hwnd() as _;
            let mut rect = winapi::shared::windef::RECT { left: 0, top: 0, right: 0, bottom: 0 };
            if winapi::um::winuser::GetClientRect(hwnd, &mut rect) == 0 {
                println!("GetClientRect failed. GetLastError: {:?}",
                         winapi::um::errhandlingapi::GetLastError());
            }
        }
        std::thread::sleep(std::time::Duration::from_millis(10)); // Pretend to do some work
        winit::ControlFlow::Continue
    });
}
```

Osspial:
This PR removes child thread creation from the win32 event loop, which has the side effect of getting rid of all the hacks present beforehand to get inter-thread communication working properly. It also updates the window callback to use win32's subclassing API to pass winit's data to windows, rather than going through thread-local storage.

At first glance, this closes #415, #391, and #94, though further testing is required and it may close other issues I'm not seeing.

OUTSTANDING ISSUES:
* [ ] How do we handle Windows' modal loops (e.g. on resizing, moving)? Currently, the users' event handling code is paused, which is bad (not to mention inconsistent with other platforms).
* [ ] `EventsLoop` handles events from ALL windows on a thread, not just windows created within the loop.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,7 @@ features = [
     "libloaderapi",
     "windowsx",
     "hidusage",
+    "commctrl"
 ]
 
 [target.'cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "openbsd"))'.dependencies]
--- a/src/platform/windows/events_loop.rs
+++ b/src/platform/windows/events_loop.rs
@@ -224,7 +224,7 @@ pub unsafe extern "system" fn callback(
                 window_id: SuperWindowId(WindowId(window)),
                 event: Closed
             });
-            winuser::DefWindowProcW(window, msg, wparam, lparam)
+            commctrl::DefSubclassProc(window, msg, wparam, lparam)
         },
         winuser::WM_DESTROY => {
             Box::from_raw(subclass_input);
@@ -364,7 +364,7 @@ pub unsafe extern "system" fn callback(
             use events::ElementState::Pressed;
             use events::VirtualKeyCode;
             if msg == winuser::WM_SYSKEYDOWN && wparam as i32 == winuser::VK_F4 {
-                winuser::DefWindowProcW(window, msg, wparam, lparam)
+                commctrl::DefSubclassProc(window, msg, wparam, lparam)
             } else {
                 let (scancode, vkey) = event::vkeycode_to_element(wparam, lparam);
                 subclass_input.event_queue.borrow_mut().push_back(Event::WindowEvent {
@@ -561,7 +561,7 @@ pub unsafe extern "system" fn callback(
 
                 0
             } else {
-                winuser::DefWindowProcW(window, msg, wparam, lparam)
+                commctrl::DefSubclassProc(window, msg, wparam, lparam)
             }
         },
 
@@ -612,7 +612,7 @@ pub unsafe extern "system" fn callback(
             };
 
             if call_def_window_proc {
-                winuser::DefWindowProcW(window, msg, wparam, lparam)
+                commctrl::DefSubclassProc(window, msg, wparam, lparam)
             } else {
                 0
             }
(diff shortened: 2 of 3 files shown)


Osspial:
There are two main ways I can see to handle the modal loops:

* Re-implement the modal loop code, in a non-blocking way, within winit.
* Adjust the event loop API to give winit total control over the application's main loop, even while polling events, so that the user's provided `callback` can be called within the window procedure.

I'm in favor of the second option, as the first one would involve a *lot* of work on our end both to create the initial implementation, with proper window snapping support and such, as well as support additional features Microsoft may add in the future.

fschutt:
Any update on this? I'm currently waiting on this because of #506.

If winit ever gets menus / integration with the OS I **need** to be on the same thread as the HWND, otherwise, Windows just locks up the application. 

I'm happy with any solution, but please just make sure that the `.get_hwnd()` call (which I need to implement Win32-specific functions that act on / modify the HWND) runs on the same thread as the user code.

francesca64:
@fschutt this is blocked on #459, where the current status is "waiting for people to respond to me" https://github.com/tomaka/winit/issues/459#issuecomment-389220019 (It's also ostensibly blocked by HiDPI support, which is awaiting responses too https://github.com/tomaka/winit/issues/105#issuecomment-390372062. The rationale for getting this in first is because #459 is very much a breaking change and IMO #105 is too important to gate behind that API migration.)

Most people would assume that the best way to help winit grow is by working on the code, but I'd say the best thing someone can do is push API discussions forward. I encourage you to participate in both of those discussions, since doing so will likely lead to a faster resolution.

Osspial:
Closing; obsoleted by #638.

PR #452, opened 2018-04-09
Doc typo fix for struct winit::Window

virome:


virome:
Whoops, didn't [ci skip] this, sorry!

virome:
Did you spot the 'matches' on the line below? With that, I think it reads fine.
What's an example of a linking verb you were thinking of?

matt-dombrowski:
I would like to preface this by stating that I have not attempted to learn this particular dialect by going through previous comments, so take this with a grain of salt. The phrase I'm keying in on is "if nothing". It sounds like it is missing an object compliment after it (i.e. "if nothing has happened" or "if nothing is present"). I think the parsimonious way to make this flow better is to say "if it is nothing" (which is the linking verb I was thinking of). I can tell that the line of thinking for this comment is a basic branch structured as "if x than y or if not x than z", but that doesn't translate well into natural language. Personally I would rephrase to more explicitly follow the schema of "if x than y, or if not possible z" (for example "...or the primary monitor if there is no current monitor"). I am aware that this new schema is semantically identical to the previous one, but it translates better into natural language. It is a tad redundant, but it does more explicitly explain what is going on while still being succinct. 

I did not spot the "matches". I believe it is semantically null, so would be in favor of axing it entirely.

virome:
True, it isn't exactly clear what it means for there to be no matches.

I'm not familiar enough with the codebase to know why there wouldn't be any matches for the current monitor. I'll look into that, and then maybe this sentence could be rewritten to be a bit more descriptive.

francesca64:
I was initially going to tell you that this was fine as-is, but...

> I'm not familiar enough with the codebase to know why there wouldn't be any matches for the current monitor.

...then I realized that I don't actually know this either. I've thus investigated:

- **Windows**: stubbed as `unimplemented!()`, though #457 implements it. The monitor that the majority of the window is on is returned.
- **macOS**: stubbed as `unimplemented!()`.
- **X11**: the monitor that the majority of the window is on is returned. There's a [default value](https://github.com/tomaka/winit/blob/4005bf11e470a87113c18641d9b9df8620253a89/src/platform/linux/x11/window.rs#L532-L541) that's used when `get_position` or `get_outer_size` return `None` (though it's dubious that it makes sense to return a monitor at all in that case, since AFAIK this would really only be hit for a window that's been closed). That default value is the first monitor returned by [`get_available_monitors`](https://github.com/tomaka/winit/blob/107a1e7332992dc4fa5c9141c443c0635a007d9a/src/platform/linux/x11/monitor.rs#L22), which is the first monitor enumerated by [`XRRGetMonitors`](https://cgit.freedesktop.org/xorg/proto/randrproto/tree/randrproto.txt#n1614) on XRandR 1.5 and onward, and [`XRRGetScreenResources`](https://cgit.freedesktop.org/xorg/proto/randrproto/tree/randrproto.txt#n662) otherwise. Neither of these claim to always have the primary monitor as the first one.
- **Wayland**: due to inavailability of positioning info, the window's most recent monitor is returned.
- **Android, iOS, and emscripten**: these implementations don't really do anything, so it's not an issue here.

So you've inadvertently caused the discovery that the doc comment isn't even accurate, which I'd say is a victory for winit!

virome:
In that case, do you think it's even necessary to mention the possiblity of 'no matches' in the documentation? Or something akin to "if a window's monitor cannot not be determined for certain, a platform-specifc best guess is returned."

matt-dombrowski:
Oh, I'll bet that that was what the "matches" comment was for!

francesca64:
I personally don't think it's necessary. There aren't really any *reasonable* cases where there won't be a match.

matt-dombrowski:
I probably should have been more clear, I still believe that the "matches" comment is semantically null. Any use it may have had is met with a descriptive comment about the purpose of the function. I merely now understand it's (probable) former purpose.

francesca64:
@virome I see that you updated your commit, so I assume you're ready for me to merge this. For future reference, GitHub didn't send me a notification when that happened; IIRC it only notifies me if you *add* commits, so it's always good to comment to give me a heads up.

Anyway, thanks for the contribution!

virome:
@francesca64 Good to know!

PR #453, opened 2018-04-09
Fix some more unconstrained types in msg_send results

SimonSapin:
The error I was investigating https://github.com/servo/servo/pull/20474#issuecomment-379802897 turned out to be already be fixed by https://github.com/tomaka/winit/pull/428, but there was a few more cases of the same problem.


francesca64:
Thanks! I think this should have a CHANGELOG entry though.

SimonSapin:
I’m already several yaks deep and I don’t directly need this change so this is only a drive-by PR and I won’t work on it further, sorry. Feel free to take over.

PR #454, opened 2018-04-09
Windows: Implement Refresh event

### LINKED ISSUE #180: The 'Refresh' event isn't generated on Windows
There is the 'WM_PAINT' window message which Windows sends to an application when its window needs to be redrawn. Logically this message should be processed and the 'winit::WindowEvent::Refresh' event should be emitted as a consequence.

francesca64:
Fixes #180 

This is based on [the snippet](https://github.com/tomaka/winit/issues/180#issuecomment-302977662) provided by @LPGhatguy, with one key difference: rather than returning 0, we pass the event along to the default handler. The reason for this is made clear in [the docs](https://msdn.microsoft.com/en-us/library/windows/desktop/dd145137(v=vs.85).aspx):

> If an application processes a `WM_PAINT` message but does not call `BeginPaint` or otherwise clear the update region, the application continues to receive `WM_PAINT` messages as long as the region is not empty. In all cases, an application must clear the update region before returning from the `WM_PAINT` message.

Using the original snippet, event processing slows down massively and my laptop fan gets quite loud. With this, `Refresh` appears to have the same behavior on Windows as on X11.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - Overhauled X11 window geometry calculations. `get_position` and `set_position` are more universally accurate across different window managers, and `get_outer_size` actually works now.
 - Fixed SIGSEGV/SIGILL crashes on macOS caused by stabilization of the `!` (never) type.
+- Implemented `Refresh` event on Windows.
 
 # Version 0.12.0 (2018-04-06)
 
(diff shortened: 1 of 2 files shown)


tomaka:
Thanks! Feel free to merge after conflicts are resolved.

Issue #300, opened 2017-09-27
True fullscreen on macOS


rennis250:
Hey there,

I'm able to run the fullscreen examples, but they don't open a true fullscreen window on macOS, just a standard window that is maximised to full size. I took a look at the code and it seems that winit is following the same process as described here:

https://developer.apple.com/library/content/documentation/GraphicsImaging/Conceptual/OpenGL-MacProgGuide/opengl_fullscreen/opengl_cgl.html

One discrepancy is that Apple's documentation recommends that the fullscreen window should be deferred, whereas winit allows sets this parameter to NO in ```src/platform/macos/window.rs```, line 435.

Also, it recommends that when the window is displayed, the ```makeKeyAndOrderFront``` parameter should be ```self```, whereas winit always sets this to ```nil``` in ```src/platform/macos/window.rs```, line 333.

I'm going to try playing with these on my own, but wanted to open this in case someone else already has dealt with this and can provide an answer.

Thanks and best,
Rob

rennis250:
The problem is line 432. It should only request ```appkit::NSBorderlessWindowMask``` and nothing else for a true fullscreen window.

francesca64:
Fixed in #465.

> Also, it recommends that when the window is displayed, the makeKeyAndOrderFront parameter should be self, whereas winit always sets this to nil in src/platform/macos/window.rs, line 333.

What would you expect `self` to point to here?

(It also doesn't seem to make a difference, since it works just as well if I pass arbitrary integers to it.)

Issue #252, opened 2017-08-04
[META] Feature parity between platforms


vberger:
This is intended to be a meta-issue to track feature parity between the various backends.

This is a first draft, please all who have the right to, update this message with what you find missing (be it a feature I forgot or the status of a feature for a platform or a platform entirely)

Each section includes a collapsed description of the features it lists.

Legend: 

- ✔️ : Works as intended
- ▢ : Mostly works but some bugs are known
- ❌ : Missing feature or large bugs making it unusable
- **N/A** : Not applicable for this platform
- ❓ : Unknown status

----------

**Display features**

| Feature | Windows | MacOS | Linux x11 | Linux Wayland | Android | iOS | Emscripten |
| :--- | :---: | :---: | :---: | :---: | :---: | :---: | :---: |
| Window initialization | ✔️ | ✔️ | ▢ #5 | ✔️ | ▢ #33 | ▢ #33 | ❓ |
| Providing pointer to init OpenGL | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ❓ |
| Providing pointer to init vulkan | ✔️ | ✔️  | ✔️ |  ✔️ | ✔️ | ❓ | **N/A** |
| Window decorations | ✔️ | ✔️ | ✔️ | ▢ #306   | **N/A** | **N/A** | **N/A** |
| Window decorations toggle | ✔️ | ✔️ | ✔️ | ✔️ | **N/A** | **N/A** | **N/A** |
| Window resizing | ✔️ | ▢ #219 | ✔️ | ▢ #306 | **N/A** | **N/A** | ❓ |
| Window transparency | ▢ #260 | ✔️ | ✔️ | ✔️ | **N/A** | **N/A** | **N/A** |
| Window maximization | ✔️ | ✔️ | ✔️ | ✔️ | **N/A** | **N/A** | **N/A** |
| Window maximization toggle | ✔️ | ✔️ | ✔️ | ✔️ | **N/A** | **N/A** | **N/A** |
| Fullscreen | ✔️ | ✔️ | ✔️ | ✔️ | **N/A** | **N/A** | ❌ |
| Fullscreen toggle | ✔️ | ✔️ | ✔️ | ✔️ | **N/A** | **N/A** | ❌ |
| HiDPI support #105 | ✔️ | ✔️ | ✔️ | ✔️ | ▢ | ✔️ | ✔️ |

<details>
<summary>Expand details of features</summary>

[...]

Issue #129, opened 2017-01-31
Would it make sense to add a way to maximize the window?


pedrocr:
Windowing systems often have a WM hint or similar mechanism to get the operating system to resize the window to it's maximum dimensions on the screen. Would adding this to the Window API make sense?

francesca64:
Thanks to #457, this is now implemented on Windows.

francesca64:
This was also implemented on Wayland back in #383.

francesca64:
With #465, this is now implemented everywhere.

Issue #72, opened 2016-11-03
Is it possible to make an existing window fullscreen?


tomaka:
Original: https://github.com/tomaka/glutin/issues/782

francesca64:
This has been in the API as `Window::set_fullscreen` since #130, but it's only implemented on X11.

francesca64:
Thanks to #457, this is now implemented on Windows.

francesca64:
This was also implemented on Wayland back in #383.

francesca64:
With #465, this is now implemented everywhere.

PR #465, opened 2018-04-14
 Implement set_maximized, get_current_monitor, set_fullscreen and set_decorations for MacOS

edwin0cheng:
This PR implemented following functions in MacOS:
- `Window::get_current_monitor`
- `Window::set_fullscreen`
- `Window::set_maximized`
- `Window::set_decorations`
- `WindowBuilder::with_maximized`

And the maximized and fullscreen behaviors are following safari's conventions. (See https://github.com/tomaka/winit/issues/460).

This PR should f-ixed #300, #72, #66, #129, and Part of #252

### DIFF
--- a/src/platform/macos/window.rs
+++ b/src/platform/macos/window.rs
@@ -347,8 +347,7 @@ impl WindowDelegate {
                         withObject:nil
                         afterDelay: 0.5                        
                     ];
-                }
-                else {
+                } else {
                     state.restore_state_from_fullscreen();
                 }
             }
@@ -574,7 +573,7 @@ impl Window2 {
         }
 
         // Make key have to be after set fullscreen
-        // to prevent normal zie window brefly appears
+        // to prevent normal size window brefly appears
         unsafe {
             if win_attribs.visible {
                 window.window.makeKeyAndOrderFront_(nil);
--- a/src/platform/macos/window.rs
+++ b/src/platform/macos/window.rs
@@ -12,8 +12,9 @@ use objc::declare::ClassDecl;
 
 use cocoa;
 use cocoa::base::{id, nil};
-use cocoa::foundation::{NSPoint, NSRect, NSSize, NSString};
-use cocoa::appkit::{self, NSApplication, NSColor, NSView, NSWindow, NSWindowStyleMask, NSWindowButton};
+use cocoa::foundation::{NSDictionary, NSPoint, NSRect, NSSize, NSString};
+use cocoa::appkit::{self, NSApplication, NSColor, NSScreen, NSView, NSWindow, NSWindowButton,
+                    NSWindowStyleMask};
 
 use core_graphics::display::CGDisplay;
 
@@ -22,7 +23,7 @@ use std::ops::Deref;
 use std::os::raw::c_void;
 use std::sync::Weak;
 
-use super::events_loop::Shared;
+use super::events_loop::{EventsLoop,Shared};
 
 use window::MonitorId as RootMonitorId;
 
@@ -512,7 +513,7 @@ impl Window2 {
 
     fn create_view(window: id) -> Option<IdRef> {
         unsafe {
-            let view = IdRef::new(NSView::alloc(nil).init());
+            let view = IdRef::new(NSView::init(NSView::alloc(nil)));
             view.non_nil().map(|view| {
                 view.setWantsBestResolutionOpenGLSurface_(YES);
                 window.setContentView_(*view);
@@ -717,7 +718,18 @@ impl Window2 {
 
     #[inline]
     pub fn get_current_monitor(&self) -> RootMonitorId {
-        unimplemented!()
+        unsafe {
+            let screen = NSScreen::mainScreen(nil);
+            let desc = NSScreen::deviceDescription(screen);
+            let key = IdRef::new(NSString::alloc(nil).init_str("NSScreenNumber"));
+            
+            let value = NSDictionary::valueForKey_(desc, *key);
+            let display_id = msg_send![value,unsignedIntegerValue];
+
+            RootMonitorId {
+                inner : EventsLoop::make_monitor_from_display(display_id)
+            }
+        }
     }
 }
 
--- a/src/platform/macos/window.rs
+++ b/src/platform/macos/window.rs
@@ -33,6 +33,8 @@ pub struct Id(pub usize);
 struct DelegateState {
     view: IdRef,
     window: IdRef,
+    win_attribs: WindowAttributes,
+    handle_with_fullscreen: bool,
     shared: Weak<Shared>,
 }
 
@@ -195,6 +197,57 @@ impl WindowDelegate {
             }
         }
 
+        /// Invoked when entered fullscreen
+        extern fn window_did_enter_fullscreen(this: &Object, _: Sel, _: id){
+            unsafe {
+                let state: *mut c_void = *this.get_ivar("winitState");
+                let state = &mut *(state as *mut DelegateState);
+                state.win_attribs.fullscreen = Some(get_current_monitor());
+                
+                state.handle_with_fullscreen = false;
+            }
+        }
+
+        /// Invoked when exited fullscreen
+        extern fn window_did_exit_fullscreen(this: &Object, _: Sel, _: id){
+            unsafe {
+                let state: *mut c_void = *this.get_ivar("winitState");
+                let state = &mut *(state as *mut DelegateState);
+                state.win_attribs.fullscreen = None;
+            }
+        }
+
+        /// Invoked when fail to enter fullscreen
+        ///
+        /// When this window launch from a fullscreen app (e.g. launch from VS Code
+        /// terminal), it creates a new virtual destkop and d an transition
+        /// animation. This animation takes one second and cannot be disable without
+        /// elevated privileges. In this animation time, all toggleFullscreen events
+        /// will be failed. In this implementation, we will try again by using
+        /// performSelector:withObject:afterDelay: until window_did_enter_fullscreen.
+        /// It should be fine as we only do this at initialzation (i.e with_fullscreen
+        /// was set).
+        ///
+        /// From Apple doc:
+        /// In some cases, the transition to enter full-screen mode can fail,
+        /// due to being in the midst of handling some other animation or user gesture.
+        /// This method indicates that there was an error, and you should clean up any
+        /// work you may have done to prepare to enter full-screen mode.
+        extern "C" fn window_did_fail_to_enter_fullscreen(this: &Object, _: Sel, _: id) {
+            unsafe {
+                let state: *mut c_void = *this.get_ivar("winitState");
+                let state = &mut *(state as *mut DelegateState);
+
+                if state.handle_with_fullscreen {
+                    let _: () = msg_send![*state.window, 
+                        performSelector:sel!(toggleFullScreen:)
+                        withObject:nil
+                        afterDelay: 0.01                        
+                    ];
+                }
+            }
+        }
+
         static mut DELEGATE_CLASS: *const Class = 0 as *const Class;
         static INIT: std::sync::Once = std::sync::ONCE_INIT;
 
@@ -228,6 +281,14 @@ impl WindowDelegate {
                 conclude_drag_operation as extern fn(&Object, Sel, id));
            decl.add_method(sel!(draggingExited:),
                 dragging_exited as extern fn(&Object, Sel, id));
+
+            // callbacks for fullscreen events
+            decl.add_method(sel!(windowDidEnterFullScreen:),
+                window_did_enter_fullscreen as extern fn(&Object, Sel, id));
+            decl.add_method(sel!(windowDidExitFullScreen:),
+                window_did_exit_fullscreen as extern fn(&Object, Sel, id));
+            decl.add_method(sel!(windowDidFailToEnterFullScreen:),
+                window_did_fail_to_enter_fullscreen as extern fn(&Object, Sel, id));
 
             // Store internal state as user data
             decl.add_ivar::<*mut c_void>("winitState");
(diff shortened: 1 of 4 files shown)


francesca64:
I installed High Sierra in preparation for this! I'll review it as soon as I get a chance.

edwin0cheng:
This implementation are quite straight forward, and should be following standard macos convention, which is just calling `NSWindow::toggleFullscreen:` for fullscreen and `NSWindow::zoom` for maximized.

However, `set_decoration` and `with_fullscreen` are little bit tricky:
- Calling `toggleFullscreen:` and `zoom:` on a non-resizable window in MacOS is no-op by default, i have to do it manually (use `setFrame` for `zoom`) or set it to resizable tempority and restore it back. (for `toggleFullScreen`
- Normally, the implementation of `with_fullscreen` is just calling `toggleFullscreen` after window initialization, but when a winit app are launched from a fullscreen app (e.g. from a FullScreen VS Code Terminal), it creates a new virtual desktop and a transition animation, which tasks one sec and cannot be disable. In this animation time, all `toggleFullscreen` events will be failed. The PR just make another `toggleFullscreen` call with delay to do it again until it success.

src/platform/macos/window.rs
```diff
@@ -33,6 +35,77 @@ struct DelegateState {
     view: IdRef,
     window: IdRef,
     shared: Weak<Shared>,
+
+    win_attribs: RefCell<WindowAttributes>,
```
francesca64: Is it safe to be using cells in async callbacks?

src/platform/macos/window.rs
```diff
@@ -701,23 +862,76 @@ impl Window2 {
...
+        let state = &self.delegate.state;
+        let win_attribs = state.win_attribs.borrow_mut();
+
+        let current = win_attribs.fullscreen.clone();
+        match (current.clone(), monitor.clone()) {
+            (None, None) => {
+                return;
+            }
+            (Some(_), Some(_)) => {
+                return;
```
francesca64: When the two monitors aren't the same, this should be `unimplemented!()`.

src/platform/macos/window.rs
```diff
@@ -701,23 +862,76 @@ impl Window2 {
...
-    pub fn set_fullscreen(&self, _monitor: Option<RootMonitorId>) {
-        unimplemented!()
+    /// TODO: Right now set_fullscreen do not work on switching monitors
+    /// in fullscreen mode
+    pub fn set_fullscreen(&self, monitor: Option<RootMonitorId>) {
+        let state = &self.delegate.state;
+        let win_attribs = state.win_attribs.borrow_mut();
+
+        let current = win_attribs.fullscreen.clone();
+        match (current.clone(), monitor.clone()) {
```
francesca64: None of these clones are necessary.

src/platform/macos/window.rs
```diff
@@ -12,17 +12,19 @@ use objc::declare::ClassDecl;
 
 use cocoa;
 use cocoa::base::{id, nil};
-use cocoa::foundation::{NSPoint, NSRect, NSSize, NSString};
-use cocoa::appkit::{self, NSApplication, NSColor, NSView, NSWindow, NSWindowStyleMask, NSWindowButton};
+use cocoa::foundation::{NSDictionary, NSPoint, NSRect, NSSize, NSString};
+use cocoa::appkit::{self, NSApplication, NSColor, NSScreen, NSView, NSWindow, NSWindowButton,
+                    NSWindowStyleMask};
```
francesca64: [Indentation](https://github.com/rust-lang-nursery/fmt-rfcs/blob/586f18f4ead07bf79f892656537e05306ae30dd4/guide/principles.md#overarching-guidelines)

francesca64:
- When using `with_fullscreen`, the window briefly appears at normal size before becoming fullscreen. Setting fullscreen prior to making the window key fixes this.
- This currently disregards whatever monitor is passed for setting fullscreen, always using whichever monitor has the keyboard focus. [`enterFullScreenMode`](https://developer.apple.com/documentation/appkit/nsview/1483780-enterfullscreenmode) should work for this, but it appears to be absent from core-foundation.
- Similarly, `get_current_monitor` always returns the currently focused monitor, rather than the monitor the window is on. [`CGGetDisplaysWithRect`](https://developer.apple.com/documentation/coregraphics/1456071-cggetdisplayswithrect) could be used to find the monitor that the majority of the window is on.
- `set_decorations` doesn't respect the more granular style settings provided by `WindowBuilderExt`. The best way to handle this isn't obvious, as "decorated" isn't a binary state in this case.
- `set_fullscreen` similarly has no awareness of the extra attributes from `WindowBuilderExt`, i.e. using `with_titlebar_hidden` makes entering fullscreen impossible.
- The titlebar popping in when switching to fullscreen from an undecorated window can be made less jarring by setting the titlebar to transparent, and the title and buttons to hidden, though that small visual improvement might not be worth the added complexity.
- Using the button to set maximization causes the wrong maximization state to be restored when exiting fullscreen.

servo/core-foundation-rs#204 will give us some more options (literally), though nothing that it looks like we presently need.

edwin0cheng:
> This currently disregards whatever monitor is passed for setting fullscreen, always using whichever monitor has the keyboard focus. enterFullScreenMode should work for this, but it appears to be absent from core-foundation.

We can use msg_send! macro to send the selector to use it, i will give it a try. And we could expose the fullscreen options in WindowExt and WindowBuilderExt, but i think i would recommend to create another PR to address it.

> Similarly, get_current_monitor always returns the currently focused monitor, rather than the monitor the window is on. CGGetDisplaysWithRect could be used to find the monitor that the majority of the window is on.

I want to use this but it is absent in current CoreFoundation binding. (Which is c api which i cannot use msg_send! here.

> The titlebar popping in when switching to fullscreen from an undecorated window can be made less jarring by setting the titlebar to transparent, and the title and buttons to hidden, though that small visual improvement might not be worth the added complexity.

Indeed, we would make another PR to address this too.

edwin0cheng:
Ok, seem to be we cannot use ` enterFullScreenMode` as all keys of [`NSViewFullScreenModeOptionKey`](https://developer.apple.com/documentation/appkit/nsviewfullscreenmodeoptionkey?language=objc) are not bound in rust cocoa library.

edwin0cheng:
>- When using with_fullscreen, the window briefly appears at normal size before becoming fullscreen. Setting fullscreen prior to making the window key fixes this.
>- Using the button to set maximization causes the wrong maximization state to be restored when exiting fullscreen.

These bugs should be fixed.

edwin0cheng:
@francesca64 It should be fine now, except for the points i mentioned in previous comments, we need to wait upstream cocoa crate to add corresponding binding.

src/platform/macos/window.rs
francesca64: Sorry, I misunderstood how the delegate methods work.

src/platform/macos/window.rs
```diff
@@ -370,15 +549,44 @@ impl Window2 {
...
+            unsafe {
+                if monitor.inner != get_current_monitor().inner {
+                    unimplemented!();
+                }
+            }
+            window.set_fullscreen(Some(monitor.clone()));
+        }
+
+        // Make key have to be after set fullscreen
+        // to prevent normal zie window brefly appears
```
francesca64: zie -> size

src/platform/macos/window.rs
```diff
@@ -194,6 +288,72 @@ impl WindowDelegate {
...
+                let state = &mut *(state as *mut DelegateState);
+
+                if state.handle_with_fullscreen {
+                    let _: () = msg_send![*state.window, 
+                        performSelector:sel!(toggleFullScreen:)
+                        withObject:nil
+                        afterDelay: 0.5                        
+                    ];
+                }
+                else {
```
francesca64: `} else {`

francesca64:
Thanks, and addressing the rest in a later PR seems reasonable. Two more nitpicks and we should be good to merge.

edwin0cheng:
Fixed, and thank you so much.

francesca64:
Are you noticing that when using `with_fullscreen`, it now takes two tries to succeed? Things worked fine as of 88428752, and the problem starts in 1a808b02.

edwin0cheng:
No, i just tested it and it only take 1 try. Could you mind to detail step to reproduce it?

francesca64:
Okay, I've tried a bit more, and I think I know what's happening. I run the fullscreen example, press F twice, then quit, and I get this event log:
```
WindowEvent { window_id: _, event: Resized(2560, 1440) }
WindowEvent { window_id: _, event: Focused(true) }
WindowEvent { window_id: _, event: Resized(2560, 1440) }
WindowEvent { window_id: _, event: HiDPIFactorChanged(1.0) }
WindowEvent { window_id: _, event: ReceivedCharacter('f') }
WindowEvent { window_id: _, event: KeyboardInput { device_id: _, input: KeyboardInput { scancode: 3, state: Released, virtual_keycode: Some(F), modifiers: ModifiersState { shift: false, ctrl: false, alt: false, logo: false } } } }
WindowEvent { window_id: _, event: KeyboardInput { device_id: _, input: KeyboardInput { scancode: 3, state: Pressed, virtual_keycode: Some(F), modifiers: ModifiersState { shift: false, ctrl: false, alt: false, logo: false } } } }
WindowEvent { window_id: _, event: ReceivedCharacter('f') }
WindowEvent { window_id: _, event: Resized(2560, 1440) }
WindowEvent { window_id: _, event: Resized(2560, 1440) }
WindowEvent { window_id: _, event: HiDPIFactorChanged(1.0) }
WindowEvent { window_id: _, event: Resized(2560, 1315) }
WindowEvent { window_id: _, event: ReceivedCharacter('\u{1b}') }
WindowEvent { window_id: _, event: KeyboardInput { device_id: _, input: KeyboardInput { scancode: 53, state: Released, virtual_keycode: Some(Escape), modifiers: ModifiersState { shift: false, ctrl: false, alt: false, logo: false } } } }
```
As you can see, the first F press never generates a `Pressed` event! Besides pressing F a second time to get it to work, I can also just click and then press F - the key is that the first event is being ignored. If you've been keeping up on the recent PRs, this sounds like the result of #466. If I drop that commit, then I get the expected event log:
```
WindowEvent { window_id: _, event: Resized(2560, 1440) }
WindowEvent { window_id: _, event: Focused(true) }
WindowEvent { window_id: _, event: Resized(2560, 1440) }
WindowEvent { window_id: _, event: HiDPIFactorChanged(1.0) }
WindowEvent { window_id: _, event: Resized(2560, 1440) }
WindowEvent { window_id: _, event: HiDPIFactorChanged(1.0) }
WindowEvent { window_id: _, event: KeyboardInput { device_id: _, input: KeyboardInput { scancode: 3, state: Pressed, virtual_keycode: Some(F), modifiers: ModifiersState { shift: false, ctrl: false, alt: false, logo: false } } } }
WindowEvent { window_id: _, event: ReceivedCharacter('f') }
WindowEvent { window_id: _, event: Resized(2560, 1440) }
WindowEvent { window_id: _, event: Resized(2560, 1440) }
WindowEvent { window_id: _, event: HiDPIFactorChanged(1.0) }
WindowEvent { window_id: _, event: KeyboardInput { device_id: _, input: KeyboardInput { scancode: 3, state: Released, virtual_keycode: Some(F), modifiers: ModifiersState { shift: false, ctrl: false, alt: false, logo: false } } } }
WindowEvent { window_id: _, event: Resized(2560, 1315) }
WindowEvent { window_id: _, event: KeyboardInput { device_id: _, input: KeyboardInput { scancode: 53, state: Pressed, virtual_keycode: Some(Escape), modifiers: ModifiersState { shift: false, ctrl: false, alt: false, logo: false } } } }
```
@sodiumjoe @swiftcoder

sodiumjoe:
Interesting, that does seem like the likely culprit. I guess since full screen doesn't invoke the modal tracking loop, so it doesn't eat the MouseDown event?

sodiumjoe:
I'm actually seeing other weird downstream effects related to the hidpifactorchanged event that goes away once i revert that commit. we may want to revert and rethink that?

edwin0cheng:
I checked my code, and didn't think of some special problem. However, maybe i dont know how the discard event work. How do you make sure if window_did_resize called, and the next event must be the event you want discard?

francesca64:
@edwin0cheng I don't believe this is your fault, but rather an unforeseen interaction between this and the other PR. We didn't notice it before because your branch is based prior to that PR being merged (which is also why I ended up thinking it was a problem in one of your later commits, since switching back to your branch to test that meant the issue was no longer present). The moral of the story is that we should always rebase before testing. But anyway, I don't think you need to take any further action.

@sodiumjoe oh geeze, but I guess it's not surprising; #466 is admittedly a hack. The other moral of the story is that we should have some sort of test suite. I guess it should be reverted, and them maybe we should glob fixing it into a larger "what are we going to do about the fundamental conflicts between Cocoa and winit's design" issue? Assuming I'm correct in interpreting that the disparity is the reason we have to deal with this.

sodiumjoe:
I'd be interested to hear @swiftcoder 's opinion. It seems like we could keep the hack contained by checking the event type [here](https://github.com/tomaka/winit/blob/master/src/platform/macos/events_loop.rs#L227) and only discarding MouseUp events?

francesca64:
Alright, I agree that we should wait for @swiftcoder. Just please make some note in #468 in case tomaka swoops in and merges it.

sodiumjoe:
I just closed it, I'll reopen once we figure this out. I can also try the filtering thing I mentioned real quick.

sodiumjoe:
i just tried a really naive filtering:

```
                match event {
                    // Call the user's callback.
                    Some(event) => match event {
                        Event::WindowEvent { event: WindowEvent::MouseInput { state: ElementState::Released, .. }, .. } => {
                            if !self.shared.should_discard_next_event() {
                                self.shared.user_callback.call_with_event(event);
                            }
                        },
                        _ => self.shared.user_callback.call_with_event(event),                    },
                    None => break,
                }
```

and that doesn't appear to fix anything, so I'll wait for @swiftcoder

swiftcoder:
Your filtering would work, but you need to restrict to pressed events, not release.

That said, we are now layering a hack on top of a hack. I hadn't anticipated that `windowDidResize` would be invoked in other cases, but it makes sense.  Maybe need a better approach for the key presses... Feel free to back out my patch in the meantime, I'll have a think this evening about a more comprehensive solution.

sodiumjoe:
Derp, you're right. Changing that to `Pressed` seems to fix most of the issues, though there are still some weird artifacts in my branch of alacritty that's using the `HiDpiFactorChanged` event.

It also appears to be causing some weird issues with the relatively new WindowBuilder `                .with_titlebar_transparent(true)` and `.with_fullsize_content_view(true)`, which seems really weird.

I think it makes the most sense to back that change out for now.

PR #470, opened 2018-04-18
Revert 466 window resize mouse up fix

francesca64:
Relevant discussion [here](https://github.com/tomaka/winit/pull/465#issuecomment-382124999)


francesca64:
I find it pretty surprising that GitHub generated an empty "revert" commit for every commit since then, since it makes this PR seem much scarier than it actually is.

PR #476, opened 2018-04-20
Replace Closed event with CloseRequested and Destroyed

### LINKED ISSUE #434: API change proposal for window closing
Currently, the `Closed` event [indicates when a window has been destroyed](https://github.com/tomaka/winit/pull/416#issuecomment-370523045), and clicking the × button (or closing the window through some other external means) leads to the unconditional destruction of the window. Ideally, application developers should have the opportunity to decide how to handle external requests for the window to close, such as prompting the user about unsaved work, but that's not presently possible.

## Proposal

A new event named `Destroyed` should be created, taking the place of the existing `Closed` event. This is consistent with the naming scheme used in both Win32 and X11.

The existing `Closed` event should be removed in favor of `CloseRequested`. This prevents existing code from missing the API change, and prevents ambiguity in future code. When × is pressed/etc., the only behavior that should occur is this event being sent, allowing the application developer to decide what action to take.

A `close` method could be added to `Window` (#13). In the simplest case, `CloseRequested` would be handled by calling `close` for that window. However, since this method would have the same effect as dropping the window, it could be preferable to leave out this method and just make dropping the canonical solution.

francesca64:
Resolves #434

This is implemented and tested on Windows, macOS, X11, and Wayland.

Migrating to the new events is easy, since most people just need to replace all of their `Closed` with `CloseRequested`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Corrected `get_position` on macOS to return outer frame position, not content area position.
 - Corrected `set_position` on macOS to set outer frame position, not content area position.
 - Added `get_inner_position` method to `Window`, which gets the position of the window's client area. This is implemented on all applicable platforms (all desktop platforms other than Wayland, where this isn't possible).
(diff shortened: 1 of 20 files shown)


vberger:
LGTM for the wayland part.

PR #478, opened 2018-04-20
macOS: Implement Moved

### LINKED ISSUE #67: Event::Moved isn't working
Original: https://github.com/tomaka/glutin/issues/724

francesca64:
Fixes #67

This also accounts for the position changing during resizes, which I'm not sure if we do on other platforms.


francesca64:
I've confirmed that `Moved` is emitted when resizing on Windows and X11 as well, though in that case X11 gives us an additional `Moved` that has the nested window position relative to the parent window.

PR #479, opened 2018-04-21
Windows: Position fixes

francesca64:
- `Moved` would give incorrect values for negative positions (relevant with multiple monitors). This is because `WORD` (`u16`) was being cast directly to `i32`, so the wrap-around would never be interpreted as a negative number, instead giving us values up around `u16::MAX`. Fixing that is as easy as casting to `i16` first, but it turns out that [`WM_MOVE`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms632631(v=vs.85).aspx) supplies client area positions. So, now we use [`WM_WINDOWPOSCHANGED`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms632652(v=vs.85).aspx) instead, which gives us window positions.

- `get_position` was using `GetWindowPlacement`, which gives us [coordinates relative to the taskbar](https://msdn.microsoft.com/en-us/library/windows/desktop/ms632611(v=vs.85).aspx) instead of relative to the screen. I never noticed this before, since I had my taskbar hidden, in which case it doesn't make a difference. Otherwise, you'd get positions that are dependent on the taskbar state, and if you pass them to `set_position` (which uses `SetWindowPos` under the hood, which expects screen coordinates) then the window will shift upwards by the size of the taskbar. I changed `get_position` to use `GetWindowRect`, which removes these problems and makes behavior more consistent with other backends.


francesca64:
It looks like `Moved` on X11 has also been using client area positions, which is odd, since the docs for [`ConfigureNotify`](https://tronche.com/gui/x/xlib/events/window-state-change/configure.html) would suggest otherwise. If I trust [GLFW's implementation](https://github.com/glfw/glfw/blob/58cc4b2c5c2c9a245e09451437dd6f5af4d60c84/src/x11_window.c#L1502), then `ConfigureNotify` using client area position is consistent across WMs.

While obtaining reliable X11 window positions is a [solved problem](https://github.com/tomaka/winit/pull/438), it doesn't sound like a good idea to do that much querying every time the window moves. That could be solved very effectively by caching the frame extents (which would also make the geometry getters on X11 considerably cheaper), but managing that state would introduce a new category of potential bugs, and for best results I'd have to add shared state between `EventsLoop` and `Window2`.

At present, I'm not convinced that's worth the effort, since A) client area positions are already the *de facto* standard for this event, and B) I doubt very many people really use `Moved`. So, unless anyone feels strongly about this, my plan is to:
1. Change the implementation of `Moved` in this PR back to what it was before, just corrected to use `GET_X_LPARAM`/`GET_Y_LPARAM` so negative values are interpreted properly.
2. Change #478 to use client area positions.
3. Document that `Moved` supplies client area positions.
4. Investigate this more in future work (assuming we have agreement that `Moved` ought to supply window positions, that is).

francesca64:
Okay, despite what I said, I've figured this out on X11, so there's no obstacle to consistently using window positions (once I finish that, at least).

Issue #158, opened 2017-03-19
[macOS] `NSWindow`s are not `release`d when closed, causing odd zombie window behaviour in some cases.


mitchmindtree:
In the constructor for `Window`, the following `NSWindow` method is called:
```rust
window.setReleasedWhenClosed_(NO);
```
Instead, we rely on the `Drop` implementation for `IdRef` to release the `NSWindow` after `Window`'s drop method is called:
```rust
impl Drop for IdRef {
    fn drop(&mut self) {
        if self.0 != nil {
            let _: () = unsafe { msg_send![self.0, release] };
        }
    }
}
```
In most cases this is fine, as a window is normally dropped immediately after it is closed. However, in some multi-window applications (like the `multiwindow.rs` example), a `Window` is not dropped until *all* `Window`s are closed. This means a user might have many windows that are "closed" but have not been `release`d.

Documentation on what it means to have a "closed yet not released" window is quite sparse, however after doing some experiments it seems that cocoa does not handle this well. For example, when running the `multiwindow.rs` example, the following strange behaviour occurs:

1. Click one of the windows to focus it.
2. Close that window via the x button.
3. Swipe sideways to switch to a separate desktop. This causes a `Focus(false)` event to be produced for the window that was just closed. We shouldn't receive any events for the window that was just closed.
4. Swipe back to the original desktop. This causes a `Focus(true)` event to be produced for the window that was closed *and causes it to reappear* (though without its toolbar)!

   ![screen shot 2017-03-19 at 9 18 22 pm](https://cloud.githubusercontent.com/assets/4587373/24080012/e6cf1ce4-0ce9-11e7-858c-586e7a77c6db.png)

5. Click one of the other windows to focus it.
6. Close the window via the x button.
7. Swipe sideways to a different desktop (again causing a `Focus(false)` event to be produced for the window we just closed).
8. Swipe back to the original desktop. Not only does the window we just closed reappear, now the *first* window we closed regains its toolbar again!

   ![screen shot 2017-03-19 at 9 25 11 pm](https://cloud.githubusercontent.com/assets/4587373/24080037/91703c0a-0cea-11e7-9deb-750cacd05346.png)

[...]

PR #481, opened 2018-04-22
Macos multi windows leak

bogdad:
made multi window leak less memory

created a simple test which spawns windows and recreates them on "return" key press, the title of the window is the number of windows created.

before the change
a lot of objects created

after the change:
not a lot of objects created

and more significant:

total persisted before:

a lot persisted

and total persisted after

not a lot persisted

The gist of the change is the adding NSAutorelease pools, as in most places where we are calling objc it uses NSObject:autorelease which puts the object in the top most autorelease pool on the stack, or it is never released if no such autorelease pool exist, 
or if the topmost autorelease pool is outside of the event loop, i.e. gets freed after the thread terminates.

This is continuation of https://github.com/tomaka/winit/pull/303 which was an incorrect attempt to do the same thing.
This probably fixes https://github.com/tomaka/winit/issues/158

I know its hard to check that without instruments and looking at the lists of allocations,
not sure how to best approach approving this pr


edwin0cheng:
Just a little reminder, same as https://github.com/tomaka/winit/pull/428:

>Due to the recent changes in the Rust compiler, unconstrained type
variables are now deduced to `!` instead of `()`. There are some
occurrences where `msg_send!` is used without constraining its return
type (relying on the assumption that they would be deduced to be `()`).
As a result, the macOS port of winit stopped working in nightly.

So all `msg_send!` call should be set a return value expliclty, just like this:

`
let _ : () = msg_send![...];
`

bogdad:
@edwin0cheng hi! thank you, fixed!

francesca64:
Alright, I'm back on macOS, so I can review this now. First of all, those statistics look great!

I can still reproduce #158 on this branch, *but* after rebasing it seems completely fixed. Be sure to double-check that.

Before this can merge:
- `examples/recreate_window_leak.rs` needs to be removed, since it's not useful for users. You can embed it in your PR description or as a gist if you still want it to be accessible.
- This should have a CHANGELOG entry. I know it's not straightforward to write one for something abstract like this, but a good overview of what this improves will both A) make the next winit release seem cooler, and B) give people an idea of what things to check if they experience regressions.

Note that at the time of writing, I haven't actually finished reviewing this yet, so there could be other things. I just wanted to let you know this was being worked on.

bogdad:
Hi!

Thank you for working on it, everything seems reasonable, will do. I missed
checking 158, nice it is fixed!

Please give me few days to update.

src/platform/macos/window.rs
```diff
@@ -505,13 +515,28 @@ unsafe fn get_current_monitor() -> RootMonitorId {
 
 impl Drop for Window2 {
     fn drop(&mut self) {
+        // Remove this window from the `EventLoop`s list of windows.
+        let id = self.id();
+        if let Some(shared) = self.delegate.state.shared.upgrade() {
```
bogdad: I noticed that the shared.windows weak<Window> for this id is already none at this point, but we still need to remove it from the array itself

src/platform/macos/events_loop.rs
```diff
@@ -94,7 +94,7 @@ impl Shared {
         if let Ok(mut windows) = self.windows.lock() {
             windows.retain(|w| match w.upgrade() {
                 Some(w) => w.id() != id,
-                None => true,
+                None => false,
```
bogdad: not need to keep dead week ptrs here

src/platform/macos/window.rs
```diff
@@ -452,9 +452,15 @@ impl WindowDelegate {
         unsafe {
             let delegate = IdRef::new(msg_send![WindowDelegate::class(), new]);
 
+            // setDelegate uses autorelease on objects,
+            // so need autorelease
+            let autoreleasepool = NSAutoreleasePool::new(cocoa::base::nil);
```
francesca64: Is there any reason you went with `cocoa::base::nil` instead of using the imported `nil`? It's a minor gripe, but having that consistent makes it easier for people to tell the same type is being used throughout.

francesca64:
> Rebased, added change log entry (not really sure it is a good one, please advise)

Thanks. The CHANGELOG entry looks good.

> checked the multiwindow example,

Thanks for doing that. It gives me good peace of mind.

> Can also confirm that this branch (without rebasing) alone did not fix 158 issue, which is interesting, trying to figure out why meanwhile.

Prior to #476, windows in the multiwindow example weren't dropped until the program exited. Prior to #481, the drop implementation wasn't releasing windows correctly. With both, windows drop at the correct time and are dropped the right way.

src/platform/macos/window.rs
```diff
@@ -638,11 +679,30 @@ impl Window2 {
         }
     }
 
+    fn class() -> *const Class {
+
```
francesca64: Extra newline

src/platform/macos/window.rs
```diff
@@ -505,13 +515,33 @@ unsafe fn get_current_monitor() -> RootMonitorId {
...
+        // Window ->
+        // Rc<Window2> (makes Weak<..> in shared.windows none) ->
+        // Window2
+        // needed to remove the element from array
+        let id = self.id();
+        if let Some(shared) = self.delegate.state.shared.upgrade() {
+            shared.find_and_remove_window(id);
+        }
+
+
```
francesca64: Extra newline

src/platform/macos/window.rs
```diff
@@ -538,20 +568,29 @@ impl Window2 {
...
+        let autoreleasepool = unsafe {
+            NSAutoreleasePool::new(cocoa::base::nil)
+        };
 
         let app = match Window2::create_app(pl_attribs.activation_policy) {
             Some(app) => app,
-            None      => { return Err(OsError(format!("Couldn't create NSApplication"))); },
+            None      => {
+                let _: () = unsafe { msg_send![autoreleasepool, drain] };
+                return Err(OsError(format!("Couldn't create NSApplication"))); },
```
francesca64: The `},` should be on a newline.

francesca64:
In terms of functionality, everything looks good. I pretty much just gave you some formatting nitpicks (sorry about those, but I'm hoping to get winit to look more cohesive internally).

src/platform/macos/window.rs
```diff
@@ -505,13 +515,33 @@ unsafe fn get_current_monitor() -> RootMonitorId {
 
 impl Drop for Window2 {
     fn drop(&mut self) {
+        // Remove this window from the `EventLoop`s list of windows.
+        // The destructor order is:
+        // Window ->
+        // Rc<Window2> (makes Weak<..> in shared.windows none) ->
```
francesca64: It would also be nice to have `none` be `None`, for clarity. We should be good to merge after this.

bogdad:
Fixed all code formatting errors, sorry about that, and thank you for all the work and explanation!

francesca64:
Thanks!

Issue #338, opened 2017-11-02
DeviceIds contain no data on most platforms


Xaeroxe:
This needs to be fixed: https://github.com/tomaka/winit/blob/master/src/platform/windows/mod.rs#L22

~~Since Windows makes up 96% of gamers at the moment this effectively makes the DeviceId useless.~~

EDIT: winit is not exclusively a gaming library.

Xaeroxe:
Having just browsed the Windows source, I can now tell you this isn't as bad as it first appears because the only places this is used is in keyboard and mouse input, and thanks to the distinction between those events it's actually impossible to confuse these.  So the DeviceId, while useless, also isn't really needed.

Xaeroxe:
After even further scouring this is literally only used for the X11 backend.  DeviceId is just a dummy value for every other platform.

Xaeroxe:
So, I'd like to propose that we restructure events.  Since 99% of the time DeviceEvents are just raw mouse input we should label them as such, then create a new event for non-mouse devices that will only be emitted on X11 (and maybe other platforms if support for that is ever added)

Xaeroxe:
`DeviceEvent::Added` and `DeviceEvent::Removed` are completely dead.  Never emitted anywhere.

torkleyy:
Device ids can be useful when using multiseat AFAIK.

Xaeroxe:
@torkleyy Sure, but that's only meaningful if the device id actually carries any data.  It doesn't 99% of the time, throughout most of winit it's just dead characters.

torkleyy:
So what is it that you're suggesting?

Xaeroxe:
Key: 
- M Modify
- R Remove
- A Add

R DeviceEvent::Added

R DeviceEvent::Removed

R WindowEvent::AxisMotion

M WindowEvent Remove all DeviceIds from this enum, as they carry 0 useful data

M winit::Touch remove the DeviceId as it's also useless

M DeviceEvent::Button and DeviceEvent::Motion Either start emitting these for devices other than the mouse on platforms other than X11, or just remove it entirely.  Having this behave as it does today is dangerous as it can lead developers to believe they have support for arbitrary input devices on all platforms when in fact they are far from it.

A DeviceEvent::MouseMoved largely replaces today's DeviceEvent::Motion, but includes semantics for this being the mouse.  Also includes both x and y, rather than just one axis.

Ralith:
The proposed changes include significant regressions.

[...]

PR #482, opened 2018-04-23
Windows: Implement DeviceEvents

### LINKED ISSUE #467: Windows: winit DeviceEvent for KeyboardInput does not work
DeviceEvents for KeyboardInput don't seem to be working on Windows. My friend on Linux seems to have no issue with it.

A sample program to reproduce the issue:
```
extern crate winit;

fn main() {

    let mut events = winit::EventsLoop::new();
    let window = winit::Window::new(&events);
    
    while true {
        let mut keypressed = false;
        events.poll_events(|event| {
            match event {
                winit::Event::DeviceEvent {
                    event: winit::DeviceEvent::Key(KeyboardInput),
                    .. 
                } => {
                    eprintln!("Key pressed");
                },
                _ => {}
            }
        });
    }
}
```

francesca64:
Fixes #467

All variants other than `Text` have been implemented. While `Text` can be implemented using `ToUnicode`, that doesn't play nice with dead keys, IME, etc.

Most of the mouse `DeviceEvent`s were already implemented, but due to the flags that were used when registering for raw input events, they only worked when the window was in the foreground.

This is also a step forward for #338, as `DeviceId`s are no longer useless on Windows. On `DeviceEvent`s, the `DeviceId` contains that device's handle. While that handle could ostensibly be used by developers to query device information, my actual reason for choosing it is because it's simply a very easy way to handle this. As a fun bonus, this enabled me to create this method:
```rust
DevideIdExt::get_persistent_identifier() -> Option<String>
```
Using this gives you a unique identifier for the device that persists across replugs/reboots/etc., so it's ideal for something like device-specific configuration.

There's a notable caveat to the new `DeviceId`s, which is that the value will always be 0 for a `WindowEvent`. There doesn't seem to be any straightforward way around this limitation.

I was concerned that multi-window applications would receive *n* copies of every `DeviceEvent`, but Windows only sends them to one window per application.

Lastly, there's a chance that these additions will cause antivirus/etc. software to detect winit applications as keyloggers. I don't know how likely that is to actually happen to people, but if it does become an issue, the raw input code is neatly sequestered and would be easy to make optional during compilation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 - Corrected `get_position` on Windows to be relative to the screen rather than to the taskbar.
 - Corrected `Moved` event on Windows to use position values equivalent to those returned by `get_position`. It previously supplied client area positions instead of window positions, and would additionally interpret negative values as being very large (around `u16::MAX`).
+- On Windows, implemented all variants of `DeviceEvent` other than `Text`. Mouse `DeviceEvent`s are now received even if the window isn't in the foreground.
+- `DeviceId` on Windows is no longer a unit struct, and now contains a `u32`. For `WindowEvent`s, this will always be 0, but on `DeviceEvent`s it will be the handle to that device. `DeviceIdExt::get_persistent_identifier` can be used to acquire a unique identifier for that device that persists across replugs/reboots/etc.
 
 # Version 0.13.1 (2018-04-26)
 
--- a/src/os/windows.rs
+++ b/src/os/windows.rs
@@ -3,6 +3,7 @@
 use std::os::raw::c_void;
 use libc;
 use MonitorId;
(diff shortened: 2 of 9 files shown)


francesca64:
(I need to make sure I remember to re-export `DeviceIdExt` from glutin)

Ralith:
> get_persistent_identifier

This can be implemented on Linux using `udev_device_get_sysattr_value` with `"idVendor"`, `"idProduct"`, and `"serial"`, on the device node obtained by `XIGetProperty` on the `"Device Node"` atom. Would be pretty sweet to have as a portable API.

francesca64:
@Ralith thanks for the info. That sounds like a fun thing to implement, so I'll add it to my to-do list.

PR #484, opened 2018-04-26
Set minimum x11-dl version to include Z

branan:
Without this pin, an existing cargo.lock for an older winit will not
update the x11-dl dependency, and thus will select a version that is
missing required new XIM features.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,5 +48,5 @@ wayland-client = { version = "0.12.0", features = ["dlopen"] }
 wayland-protocols = { version = "0.12.0", features = ["unstable_protocols"] }
 wayland-kbd = "0.13.0"
 wayland-window = "0.13.0"
-x11-dl = "2.17"
+x11-dl = "2.17.5"
 percent-encoding = "1.0"


francesca64:
Thanks! This is definitely a good idea. However, shouldn't it be `=2.17.5`? It looks like cargo normally ignores the patch number completely: rust-lang/cargo#1418

When I tested this locally, it wouldn't pull in the new version. Constraining with `=` made it work. You should double check what I'm saying, though.

@Daggerbot in the future, could you please follow semver more closely? Cargo seems to make assumptions wrt semver.

branan:
That issue is related to Cargo pulling in a newer version of a package when you specify an older one, which is the [documented behavior](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html). It doesn't totally ignore Zs, it just treats them as a minimum. I don't see any reason to hard-pin to a specific Z here - we just want to be sure that winit is on a new enough version to include the bits it needs.

In my local testing this behaved correctly. I just rolled back to my pre-winit-update commit and tried the process again, to be sure nothing else I did accidentally fixed it

* Update my app to point to local winit with this patch
* Update my local vulkano to point to winit with this patch
* Build app - fails to compile due to the Closed/CloseRequested change
* Check Cargo.toml - It does contain the updated x11-dl as expected \o/

francesca64:
Alright, I tried again, and it works as expected. I'm not sure what I was doing wrong the first time.

branan:
Thanks! Please let me know when this fix is released so I can update vulkano-rs/vulkano#953

francesca64:
Do you want a release now? While I'll feel slightly embarrassed to need to release 0.13.1 a day after 0.13.0, it seems like the best thing to do.

branan:
I'm not in any rush, personally. I've got my local environment sorted.

That said, I think it's good to keep the ecosystem as up-to-date as possible, and I think 0.13.1 is necessary for that. Adding 0.13.0 dependencies to other crates is just going to cause trouble for people. It's up to you how urgent that is for your release schedule, though ¯\\\_(ツ)_/¯

francesca64:
Well, I want to get winit 0.13 in as many hands as possible, so here we go: #486

PR #487, opened 2018-04-28
macOS: fix subtle regression introduced in 0474dc986

sodiumjoe:


francesca64:
Thanks! Could you add a CHANGELOG entry?

sodiumjoe:
@francesca64 done! I always forget to do that 🤦‍♂️

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 
+- Fix regression of Window platform extensions for macOS.
```
francesca64: I was hoping for something a bit more specific. My impression from testing this is that it fixes `with_fullsize_content_view` (...and be sure to let me know if I'm wrong on that).

sodiumjoe: that's correct, I'll add more detail

francesca64:
Looks good! I'll merge once CI passes.

francesca64:
Looks like AppVeyor is stalled, which is coincidentally the only one I can't restart builds on. Since I doubt your changelog edit broke Windows builds, I'll go ahead and merge.

francesca64:
Well, we did just have a release a day ago. Besides that, note that I don't have write access on glutin, so getting winit releases downstream involves locking the tomaka mutex.

I think now that winit's level of activity is higher, it could be good to have an issue for tracking releases? I'm not really sure. I just know that I have various things in mind I'm aiming for with winit 0.14, and that I don't really know what other people want.

sodiumjoe:
Yeah, I figured it would be best to get as much into the glutin release as possible, since that appears to be the bottleneck.

I'm not sure what an issue for tracking releases would entail, exactly.

If there's other stuff you want to get in before doing another release, it makes sense to wait.

I'll defer to your judgement.

francesca64:
I'm not really sure what it would entail either.

While it's appealing to try to squeeze as much into the glutin bump as possible, I also don't want to have too much change at once, or bisecting regressions could be harder. Granted, there's not much risk of that with the current additions, but I'm 2/3 of the way into changing `Moved` from using client area positions to window positions, and the fix for X11 involves some refactoring and added caching logic.

Besides the things I'm presently working on, I do aspire to get the DPI PRs for [Wayland](https://github.com/tomaka/winit/pull/341) and [Windows](https://github.com/tomaka/winit/pull/332) merged for 0.14, though I keep getting sidetracked. Be sure to let me know if my "make releases impressive" mentality results in releases being too infrequent, since I'll always end up finding another thing I want to add.

sodiumjoe:
The bisecting point is a good one, but also seems pretty easily solved by pointing a dependency at a local version and bisecting on the dependency's commits, versus bisecting across release versions (this is what I tend to do from alacritty for example, in fact I do this frequently transitively through glutin to winit).

It does seem like a bit of a treadmill of work with, and any release point will be pretty arbitrary. Personally I think it makes sense for a project at this stage to err on the side of too frequent releases, rather than too infrequent. The glutin bottleneck does present a problem for this.

The only specific concern I have is releasing glutin with the regression to the macos window platform attributes, it would be nice for the next glutin release to include this fix. But that's mostly impatience on my part.

edwin0cheng:
Just for your information : 

I did that change in 0474dc986, and the reason to change the order of `!pl_attrs.titlebar_transparent` is:
1. The default implemenation of `pl_attrs.titlebar_transparent` is false, so that if even a normal user (which will not use `pl_attrs.titlebar_transparent` flag) would affect by that, which mean the `attrs.decorations` flag  may not be working.
2. As i don't want to change the original code too much, I do not change the if then structure. It should be changed the implementation to add the flag one by one, instead of mutual exclusive if then structure.

sodiumjoe:
@edwin0cheng that's a good point. The reason it was that way when I added those flags is I also didn't want to change the structure too much, but I think you're right, we should change it to check the flags individually. I'll take a pass at that. Thanks for the note!

PR #488, opened 2018-04-28
macos platform attributes regression

sodiumjoe:
fixed !decorations case and refactored logic to be a little easier to
read

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- Fix `.with_decorations(false)` in macOS
 - Fix regression of Window platform extensions for macOS where `NSFullSizeContentViewWindowMask` was not being correctly applied to `.fullsize_content_view`.
 - Corrected `get_position` on Windows to be relative to the screen rather than to the taskbar.
 - Corrected `Moved` event on Windows to use position values equivalent to those returned by `get_position`. It previously supplied client area positions instead of window positions, and would additionally interpret negative values as being very large (around `u16::MAX`).
--- a/src/platform/macos/window.rs
+++ b/src/platform/macos/window.rs
@@ -658,32 +658,24 @@ impl Window2 {
                 }
             };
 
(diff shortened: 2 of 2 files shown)


sodiumjoe:
I tested this with all the permutations of `.with_decorations` and platform attributes that make sense.

@edwin0cheng can you take a look?

edwin0cheng:
Do we need to handle `pl_attrs.titlebar_transparent = true` here ? Or do i miss something ?

sodiumjoe:
@edwin0cheng nice catch, I missed that the close button was broken. I had accidentally duplicated `NSMinitiaturizableMask` instead of including `NSClosableMask`. This should now be fixed. I've tested with all the combinations.

`titlebar_transparent` is handled later on L704. It's a little weird, because generally you would want to use that with `fullsize_content_view`, but since we offer those separately, I figured it's up to the consumer to use those together.

sodiumjoe:
@edwin0cheng to be clear, I believe it *is* now the case that the final `else` arm of the conditional represents the default case.

edwin0cheng:
>to be clear, I believe it is now the case that the final else arm of the conditional represents the default case.
Maybe we could add a comment to describe that.

edwin0cheng:
Anyway, Look good for me, I tested other weird cases, couldn't find any bugs too.

sodiumjoe:
@edwin0cheng added that comment, thanks for testing!

sodiumjoe:
@francesca64 sorry about the churn on this issue, but I think this is good to merge now.

francesca64:
@edwin0cheng thanks for reviewing this!

@sodiumjoe you have nothing to apologize for. The logic is a lot easier to follow now, so I'm happy.

Thinking more about releases, I've noticed that the changelog is already getting pretty long. I think it would make sense to cut another release around the end of the week.

sodiumjoe:
@francesca64 sounds great, thanks!

PR #489, opened 2018-04-28
x11: Always receive Awakened event in run_forever

### LINKED ISSUE #462: Awakened event doesn't always work
See the comment at the top of this code block:

```rust
//! This demonstrates a situation where the the Awakened event doesn't make it from the proxy to the
//! event loop (at least with the X11 backend). If you remove the `run_forever` call, the Awakened
//! event starts to work again. Maybe waiting for events invalidates the Awakened event.
//! Click the window that appears to call the proxy wakeup() method.

extern crate glutin;

fn main() {
    let window_builder = glutin::WindowBuilder::new()
		.with_title("A window")
		.with_dimensions(800, 600);
	
	let mut events_loop = glutin::EventsLoop::new();
	
	let context_builder = glutin::ContextBuilder::new();
	let _gl_window = glutin::GlWindow::new(window_builder, context_builder, &events_loop).unwrap();
	
	let events_proxy = events_loop.create_proxy();
	
	loop {
		let mut events = vec![];
		
		// Wait for events.
			println!("event: {:?}", event);
			events.push(event);
			glutin::ControlFlow::Break
		});
		
		// Just poll events without waiting.
		events_loop.poll_events(|event| {
			println!("event: {:?}", event);
			events.push(event);
		});
		
		for event in &events {
			match *event {
				glutin::Event::WindowEvent{event: glutin::WindowEvent::MouseInput{state, ..}, ..} => {
					match state {
						glutin::ElementState::Pressed => {
							println!("wakeup()");
							events_proxy.wakeup().unwrap();
						}
						_ => {}
					}
				}
				_ => {}
			}
		}
	}
}
```

When it works, it should print `wakeup()` then `event: Awakened` to the console when the window is clicked. When it does not, it will only print `wakeup()`.

b-r-u:
Do not reset the pending_wakeup boolean at the start of run_forever so
that each call to EventsLoopProxy::wakeup results in an Awakened event.

Fixes #462

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - Corrected `get_position` on Windows to be relative to the screen rather than to the taskbar.
 - Corrected `Moved` event on Windows to use position values equivalent to those returned by `get_position`. It previously supplied client area positions instead of window positions, and would additionally interpret negative values as being very large (around `u16::MAX`).
 - Implemented `Moved` event on macOS.
+- Corrected `run_forever` on X11 to stop discarding `Awakened` events.
 
 # Version 0.13.1 (2018-04-26)
 
--- a/src/platform/linux/x11/mod.rs
+++ b/src/platform/linux/x11/mod.rs
@@ -190,8 +190,6 @@ impl EventsLoop {
     pub fn run_forever<F>(&mut self, mut callback: F)
         where F: FnMut(Event) -> ControlFlow
     {
-        self.pending_wakeup.store(false, atomic::Ordering::Relaxed);
-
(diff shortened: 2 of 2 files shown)


francesca64:
Looks good to me (pending your confirmation that I understand this feature https://github.com/tomaka/winit/issues/462#issuecomment-385184981).

Also, this needs a CHANGELOG entry.

b-r-u:
Done!
Would you like me to squash the commits?

francesca64:
I'll let you in on a winit secret, which is that "Squash and merge" is the only option enabled for merging PRs. So, you don't have to worry about it.

b-r-u:
Good to know :)

PR #341, opened 2017-11-05
Implement hidpi support for wayland

vberger:
This fixes as few bug regarding size managing in wayland and implements proper HiDPI support.

This follows the workflow as I described in https://github.com/tomaka/winit/issues/105#issuecomment-341964817 :

1. the window is created with requested size assuming a 1.0 dpi factor
2. once the dpi is known, appropriate `HiDPIFactorChanged` and `Resized` events are generated.

As this kind of behavior should be uniform over platforms, I think we need to decide what we wish to do regarding that before merging this PR or #332

PR #332, opened 2017-10-25
Implement high-DPI support on Windows

kryptan:
I tested this only on Windows 10. A also tested codepaths for older Windows versions by commenting out code for newer versions.

Currently it uses the `DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE` mode on Windows 10. There is also newer and better [`DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2`](https://msdn.microsoft.com/en-us/library/windows/desktop/mt791579(v=vs.85).aspx) but I had an issue with it. `V2` enables correct scaling of window's non-client area (title bar and borders) but `glViewport` is then unable to correctly set viewport and works as if non-client area was not scaled. I see nothing wrong in winit or glutin so I suspect that this is a bug in either Windows or in OpenGL drivers (I have latest NVidia drivers installed). Before updating to the latest NVidia drivers I also had another issue where moving window from one monitor to another with different DPI would make it black (at least when using glutin), however it was fixed after driver update.

cc @retep998

PR #490, opened 2018-04-29
wayland: migrate to smithay-client-toolkit

### LINKED ISSUE #475: Wayland: segfault after dropping window
Repro:
```rust
extern crate winit;

fn main() {
    let mut events_loop = winit::EventsLoop::new();

    let mut window = winit::WindowBuilder::new()
        .build(&events_loop)
        .ok();

        match event {
            winit::Event::WindowEvent { event: winit::WindowEvent::Closed, .. } => {
                window.take();
            },
            _ => (),
        }
        winit::ControlFlow::Continue
    });
}
```

Valgrind:
```
==3984== Invalid read of size 8
==3984==    at 0x66058C0: wl_proxy_get_listener (in /usr/lib/libwayland-client.so.0.3.0)
==3984==    by 0x21FAD8: <wayland_client::generated::client::wl_surface::WlSurface as wayland_client::Proxy>::from_ptr_initialized (wayland_api.rs:3274)
==3984==    by 0x1CCB93: <wayland_client::generated::client::wl_keyboard::WlKeyboard as wayland_client::Implementable<ID>>::__dispatch_msg (wayland_api.rs:4708)
==3984==    by 0x1C6CB3: wayland_client::event_queue::dispatch_func::{{closure}} (event_queue.rs:343)
==3984==    by 0x1C29C4: std::panicking::try::do_call (panicking.rs:305)
==3984==    by 0x2A1F8E: __rust_maybe_catch_panic (lib.rs:102)
==3984==    by 0x1C26B5: std::panicking::try (panicking.rs:284)
==3984==    by 0x1CE021: std::panic::catch_unwind (panic.rs:361)
==3984==    by 0x1C4FA2: wayland_client::event_queue::dispatch_func (event_queue.rs:341)
==3984==    by 0x6605623: ??? (in /usr/lib/libwayland-client.so.0.3.0)
==3984==    by 0x66069B3: wl_display_dispatch_queue_pending (in /usr/lib/libwayland-client.so.0.3.0)
==3984==    by 0x2229E5: wayland_client::event_queue::EventQueue::dispatch (event_queue.rs:151)
==3984==  Address 0x8 is not stack'd, malloc'd or (recently) free'd
```

Running with `WAYLAND_DEBUG=1` reveals the culprit:
```
[2853632.257] wl_keyboard@20.leave(4860, nil)
```

### LINKED ISSUE #485: Gracefully handle wl_seat < 5 on wayland
Currently the wayland backend binds the first seat it finds and requests version 5:

https://github.com/tomaka/winit/blob/011720848a43c52e7f0444302494c7b3ce39033c/src/platform/linux/wayland/event_loop.rs#L312-L315

This was made in the past as a lazy move (I am the culprit), but not all wayland compositors actually support this version 5 yet, so we should rather gracefully handle older versions.

vberger:
This PR migrates the wayland backend to using [Smithay's Client Toolkit](https://crates.io/crates/smithay-client-toolkit), which deprecates both wayland-kbd and wayland-window. It at the same time upgrades wayland-client to version 0.20, which significantly reworked its API. See [this blog post](https://smithay.github.io/wayland-rs-v-0-20.html) for some backgound on these changes and their reasons.

This PR moves a lot of code around, making it I expect rather difficult to review, however it brings very little functional changes, and I did not attempt to refactor the backend further than simply porting it to the new wayland-client API.

It also slightly simplifies the backend, as some of the logic is now handled by Smithay's Client Toolkit.

I've tested it using glutin on weston, more tests on other compositors would be welcome.

A few notes & unresolved questions:

### Surface painting

On wayland, a surface is defined by the contents attached to it. Notably, a surface is not drawn before anything is drawn to it. I worked around this in previous version of the backed by having the decorations be the main surface, and the actual surface contents a sub-surface. This way the window exists even if winit's user has not drawn anything yet. However it appeared this approach is horribly inefficient and can be source of lags during resizing notably, so with Smithay's Client Toolkit I reverted back to having the decorations drawn as a subsurface, and the contents as a main surface.

This means that with this PR, the window is not displayed at all until the client draws something on it. 

**The question is:** are we okay with that, or should I add some mechanism to ensure the decorations are displayed anyway?

### Seat handling

Wayland has a notion of "seats", as a set of pointer/leyboard/touchscreen. In particular, it is possible for the compositor to advertise several of them. If this is the case, that means that there are several Independent pointers on the screen, each possibly associated with a different keyboard, each having a different focus, etc...

The previous implementation used only the first seat, ignoring the rest. This one uses them all, an merges all their inputs (which can lead to confusing stuff I guess, especially regarding pointer motion).

However, cases of problem should be extremely rare, I don't know of any common setup where there would actually be several seats (this would be some kind of screen sharing with independent inputs ?).

**The question is:** What would be a proper way to handle this multi-seat thing in winit? Do we have some way in the API to note that actually there are more than one pointer and that each should be tracked independently?

Also:

- fixes #485 
- fixes #475

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - Fixed memory leak on X11 every time the mouse entered the window.
 - On X11, drag and drop now works reliably in release mode.
 - Added `WindowBuilderExt::with_resize_increments` and `WindowBuilderExt::with_base_size` to X11, allowing for more optional hints to be set.
+- Rework of the wayland backend, migrating it to use [Smithay's Client Toolkit](https://github.com/Smithay/client-toolkit).
 
 # Version 0.13.1 (2018-04-26)
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,8 +44,8 @@ features = [
 ]
 
 [target.'cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "openbsd"))'.dependencies]
-wayland-client = { version = "0.20.0", features = [ "dlopen", "egl", "cursor"] }
-smithay-client-toolkit = "0.2.0"
+wayland-client = { version = "0.20.2", features = [ "dlopen", "egl", "cursor"] }
+smithay-client-toolkit = "0.2.1"
 x11-dl = "2.17.5"
 parking_lot = "0.5"
 percent-encoding = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,10 +44,8 @@ features = [
 ]
 
 [target.'cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "openbsd"))'.dependencies]
-wayland-client = { version = "0.12.0", features = ["dlopen"] }
-wayland-protocols = { version = "0.12.0", features = ["unstable_protocols"] }
-wayland-kbd = "0.13.0"
-wayland-window = "0.13.0"
+wayland-client = { version = "0.20.0", features = [ "dlopen", "egl", "cursor"] }
+smithay-client-toolkit = "0.2.0"
 x11-dl = "2.17.5"
 parking_lot = "0.5"
 percent-encoding = "1.0"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -105,8 +105,7 @@ extern crate parking_lot;
 #[cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "openbsd"))]
 extern crate percent_encoding;
 #[cfg(any(target_os = "linux", target_os = "freebsd", target_os = "dragonfly", target_os = "openbsd"))]
-#[macro_use]
-extern crate wayland_client;
+extern crate smithay_client_toolkit as sctk;
 
 pub use events::*;
 pub use window::{AvailableMonitorsIter, MonitorId};
--- a/src/os/unix.rs
+++ b/src/os/unix.rs
@@ -173,18 +173,16 @@ impl WindowExt for Window {
 
     #[inline]
     fn get_wayland_surface(&self) -> Option<*mut raw::c_void> {
-        use wayland_client::Proxy;
         match self.window {
-            LinuxWindow::Wayland(ref w) => Some(w.get_surface().ptr() as *mut _),
+            LinuxWindow::Wayland(ref w) => Some(w.get_surface().c_ptr() as *mut _),
             _ => None
         }
     }
 
     #[inline]
     fn get_wayland_display(&self) -> Option<*mut raw::c_void> {
-        use wayland_client::Proxy;
         match self.window {
-            LinuxWindow::Wayland(ref w) => Some(w.get_display().ptr() as *mut _),
+            LinuxWindow::Wayland(ref w) => Some(w.get_display().c_ptr() as *mut _),
             _ => None
         }
     }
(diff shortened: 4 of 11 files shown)


vberger:
> Are you hoping for a review?

Not necessarily, this is also why I refrained from further refactoring. I'll leave that to an future PR when I (or somebody else actually) get to it.

> While this does have unfortunate implications for our example programs, I doubt it has any impact outside of that. Doing the least hacky thing sounds like a good idea to me.

Given past experience,  it has cause issues with old wayland compositors that do not support the `xdg_shell` protocol, as in their case the client has to draw at least once before entering the event loop otherwise it gets stuck (in block on waiting for events that will not arrive until the window is drawn...). Though compositors not supporting this protocol should be very rare.

vberger:
(oops, misclicked)

vberger:
> There's also the potential of having multiple seats using the same display, which requires a multi-focus WM to work properly. Generally speaking, WMs don't implement multi-focus, and many applications probably wouldn't behave well even if they did.

This is actually to this situation that the "multi-seat" thing in wayland refers: if the wayland server advertises more than one "seat", this means we are in a multi-focus environment.

Regarding distinguishing devices more precisely, this will need to be done using a protocol extension, as its not possible using the core protocol (a seat only advertise a single "virtual pointer" and "virtual keyboard"). I'm not sure what the current state of things is, afair it is still being discussed in the wayland-devel mailing lists...

> **Also, an aside.** I stole the libxkbcommon bindings from wayland-kbd a while ago: xkbcommon-dl. I'd like to publish them so I can use them in the X11 backend, so I want to make sure you're okay with me doing that.

Sure, go ahead, I'll probably update smithay's-client-toolkit to use it then, to avoid code-duplication.

francesca64:
Well, after trying some things, it looks like I officially can't test Wayland stuff. I can run Weston at least, but can't create OpenGL contexts, so with this PR and onward that doesn't amount to anything.

I do have another machine sitting around, so long-term I can run Wayland on that. For now, I can't offer you any peace of mind.

> Given past experience, it has cause issues with old wayland compositors that do not support the xdg_shell protocol

Can it be assumed that Wayland users are generally less likely to be using old versions of stuff (or more willing to upgrade for things to work)? Either way, since you said it should be very rare, I think this is fine as-is.

Are you ready to merge this? I'd like to release 0.14 in the next few days, and then my big goal for 0.15 is "HiDPI for everyone", so if you rebase #341 I'll rebase #332. I plan on just following the API you proposed, assuming Windows doesn't object.

> Sure, go ahead, I'll probably update smithay's-client-toolkit to use it then, to avoid code-duplication.

Thanks. The first time, I accidentally ran `cargo publish` from my libxkbcommon checkout rather than the bindings, and cargo definitely wasn't happy about that. The only API change (IIRC) is that compose functionality is in a separate module, since libxkbcommon < 0.5 doesn't have those functions. https://crates.io/crates/xkbcommon-dl

francesca64:
Oh, and do you think this should have a CHANGELOG entry?

vberger:
I'll give a last check, to verify I'm good with all this, and this can be merged I think. I'll also add a changelog entry, it can't hurt to add one even if there are normally no user-facing changes.

I indeed planned to rebase #341 once this is merged, however regarding HiDPI, I don't think the concerns I've raised regarding window creation ( https://github.com/tomaka/winit/issues/105#issuecomment-341964817 ) have been addressed, as notably #341 and #332 make different assumptions regarding whether the initial size of the window in the WindowBuilder is configured taking into account the DPI of the monitors or not.

vberger:
Okay, done, and I have rebased as well. This is good to be merged for me.

PR #495, opened 2018-05-02
Add Copy/Paste keys

chrisduerr:
This is only a tiny update which introduces the `Copy` and `Paste` keys
which are present on X11/Wayland/Windows. I'm not sure if this exists on
MacOS too, but I'm not able to test that and it doesn't have names but
just matches on the hex key values.

The "Copy" element is a reserved keyword in Rust but shouldn't cause any
conflicts in this scenario, this behavior falls in line with
https://docs.rs/winit/0.13.1/winit/enum.MouseCursor.html#variant.Copy,
but it would be possible to rename it. However `Copy` seems like the
most intuitive choice.

I'd be willing to add a few other keys too, however I wanted to keep this minimal
to what I actually want and need. I'm not quite sure why the current selection of
keys is the way it is, so I'm keeping this PR to the bare minimum of what I would
like to see introduced.

### DIFF
--- a/src/platform/linux/wayland/keyboard.rs
+++ b/src/platform/linux/wayland/keyboard.rs
@@ -290,10 +290,10 @@ fn keysym_to_vkey(keysym: u32) -> Option<VirtualKeyCode> {
         // => Some(VirtualKeyCode::WebSearch),
         // => Some(VirtualKeyCode::WebStop),
         // => Some(VirtualKeyCode::Yen),
-        // fallback
         keysyms::XKB_KEY_XF86Copy => Some(VirtualKeyCode::Copy),
         keysyms::XKB_KEY_XF86Paste => Some(VirtualKeyCode::Paste),
         keysyms::XKB_KEY_CUT => Some(VirtualKeyCode::Cut),
+        // fallback
         _ => None
     }
 }
--- a/src/platform/linux/wayland/keyboard.rs
+++ b/src/platform/linux/wayland/keyboard.rs
@@ -293,6 +293,7 @@ fn keysym_to_vkey(keysym: u32) -> Option<VirtualKeyCode> {
         // fallback
         keysyms::XKB_KEY_XF86Copy => Some(VirtualKeyCode::Copy),
         keysyms::XKB_KEY_XF86Paste => Some(VirtualKeyCode::Paste),
+        keysyms::XKB_KEY_CUT => Some(VirtualKeyCode::Cut),
         _ => None
     }
 }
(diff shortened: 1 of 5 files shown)


francesca64:
Thanks for contributing!

> I'm not quite sure why the current selection of keys is the way it is

There's not really any reason, to be honest: https://github.com/tomaka/winit/issues/71#issuecomment-350502148

Anyway, I'm guessing you don't have access to Windows, judging by the build failure? Either way, `WM_COPY` and `WM_PASTE` are event types, not keycodes, so it doesn't make sense to match against them there.

francesca64:
Oh, and generally speaking, all user-facing changes need to add an entry to the CHANGELOG.

src/platform/linux/wayland/keyboard.rs
```diff
@@ -291,6 +291,8 @@ fn keysym_to_vkey(keysym: u32) -> Option<VirtualKeyCode> {
         // => Some(VirtualKeyCode::WebStop),
         // => Some(VirtualKeyCode::Yen),
         // fallback
```
francesca64: This comment goes above `_ => None`

francesca64:
I also can't test this, since my keyboard doesn't have media keys, but on X11 and Wayland it looks harmless enough.

I think it would be good to handle `XF86XK_Cut` too, for the sake of completeness.

chrisduerr:
There were two definitions for Windows with `Paste` and I really just took a guess. :P

There doesn't seem to be a `VK_OEM_PASTE`, but there is an `APPCOMMAND_PASTE`, I'm not sure if this is a key but since it doesn't have `VK` in it and there's no `APPCOMMAND` anywhere in the Windows configs I'd just assume that there is no `Paste` on Windows? That seems a bit odd though because if anything I'd expect that the `Copy` Key would be missing.

Thanks for the heads-up on the other issues, I'll fix them right now.

chrisduerr:
Went for the cheap route to just skip `Cut` and `Paste` on Windows, still added `Copy` just as a nice bonus. Basically only implements these features on Linux now but that should still be an improvement and woud do the job for me.

If anyone knows how this works on MacOS/Windows, please do share. I don't have any MacOS or Windows machine available for testing.

chrisduerr:
@francesca64 Thanks for all the super quick feedback, I removed Windos support and reworked the CHANGELOG entry, please let me know if this is good now or if some other changes should be made.

francesca64:
> Thanks for all the super quick feedback

No problem. Reviewing PRs gives me a break from documenting the upcoming window icon API.

Everything looks perfect now. Time to merge!

PR #502, opened 2018-05-07
Windows: Fix panic when calling set_fullscreen(None)

### LINKED ISSUE #501: Crash when None is passed to set_fullscreen on Windows
Glutin / Winit panics if one passes None to set_fullscreen(..) on a window 
that hasn't been fullscreened prior.

Noticed this while testing code on windows while using glutin:

```rust
// ...

let window = glutin::GlWindow::new(window_builder, context_builder,
&events_loop).unwrap();

window.set_fullscreen(None); // will panic on windows

// ...
```

An easy fix would be something like this:

https://github.com/ladlecrow/winit/blob/3dd07152f088545e6c1898f21449d663adfd7314/src/platform/windows/window.rs#L429-L431

I'm unsure if if this is the right approach though.

ladlecrow:
Fixes #501 

This circumvents a panic when one calls set_fullscreen(None) on a window that has never been in fullscreen mode prior to calling that method.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 - Rework of the wayland backend, migrating it to use [Smithay's Client Toolkit](https://github.com/Smithay/client-toolkit).
 - Added `WindowBuilder::with_window_icon` and `Window::set_window_icon`, finally making it possible to set the window icon on Windows and X11. The `icon_loading` feature can be enabled to allow for icons to be easily loaded; see example program `window_icon.rs` for usage.
 - Windows additionally has `WindowBuilderExt::with_taskbar_icon` and `WindowExt::set_taskbar_icon`.
+- Fix panic when trying to call `set_fullscreen(None)` on a window that has not been fullscreened prior.
 
 # Version 0.13.1 (2018-04-26)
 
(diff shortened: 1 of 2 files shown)


ladlecrow:
Build and tested with Windows 7 (64bit)

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 - On X11, drag and drop now works reliably in release mode.
 - Added `WindowBuilderExt::with_resize_increments` and `WindowBuilderExt::with_base_size` to X11, allowing for more optional hints to be set.
 - Rework of the wayland backend, migrating it to use [Smithay's Client Toolkit](https://github.com/Smithay/client-toolkit).
+- Fix panic when trying to call `set_fullscreen(None)` on a window that has not been fullscreened prior.
```
francesca64: You need to specify which platform this is for.

francesca64: (...and don't forget to rebase.)

ladlecrow: Woops, my bad.

francesca64:
Alright, I still want @edwin0cheng to verify that this is good, but then we should be good to merge. Thanks for the contribution!

edwin0cheng:
Look good for me, thanks

francesca64:
@edwin0cheng thanks!

PR #503, opened 2018-05-08
Release winit 0.14.0

francesca64:
I want to get #502 in first, but after that, I feel like we're ready for another release. The CHANGELOG section is somehow even longer than last time.

If anyone has any more regressions they'd like to find, now would be a good time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+# Version 0.14.0 (2018-05-09)
+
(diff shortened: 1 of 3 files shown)


fschutt:
Could you wait a day? I have a very simple PR in the works regarding menu bars and context menus on Windows, that could be included...

francesca64:
@fschutt yes, so long as it's as simple as you say.

fschutt:
@francesca64 See #506

PR #509, opened 2018-05-10
compile with icon_loading feature on docs.rs

tinaun:
these functions should be more visible now.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,9 @@ repository = "https://github.com/tomaka/winit"
 documentation = "https://docs.rs/winit"
 categories = ["gui"]
 
+[package.metadata.docs.rs]
+features = ["icon_loading"]
+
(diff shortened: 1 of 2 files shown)


francesca64:
Thanks! I didn't know about that option.

Do you think the documentation makes it clear enough that those methods are behind a feature?

...I've also noticed that I accidentally made `Icon::to_cardinals` public, which is baaaad.

tinaun:
i thought about listing the methods that icon_loading enables in the struct level docs, but theres no real preferred way of doing this i can find

francesca64:
Yeah, I couldn't find any "official" recommendation, but I like to err on the side of safety.

I hope you don't mind that I made the change myself, but I'd feel like I were being annoying if I told you what sentence I wanted verbatim.

tinaun:
less work for me :v

Issue #131, opened 2017-02-01
osx: keyboard events not sent back to application


paulrouget:
My application use cocoa textfield. Because of:

- https://github.com/tomaka/winit/blob/f1d70d351ef9c46ab684d70bd99dcad7d8c9b015/src/platform/macos/mod.rs#L745
- PR: https://github.com/tomaka/glutin/pull/384

… textfields are not editable.

We need to find a better way to address the alert sound issue.

paulrouget:
This appears to solve this issue. I'll do a PR later.

```diff
diff --git a/src/platform/macos/mod.rs b/src/platform/macos/mod.rs
index bd1ed0b6..d3758aba 100644
--- a/src/platform/macos/mod.rs
+++ b/src/platform/macos/mod.rs
@@ -407,7 +407,21 @@ impl Window {
                 appkit::NSFullSizeContentViewWindowMask as NSUInteger
             };

-            let window = IdRef::new(NSWindow::alloc(nil).initWithContentRect_styleMask_backing_defer_(
+
+            use std::sync::{Once, ONCE_INIT};
+            static INIT: Once = ONCE_INIT;
+            INIT.call_once(|| {
+                extern fn on_key_down(_this: &Object, _sel: Sel, _id: id) { }
+                extern fn alloc(_this: &Object, _sel: Sel, _id: id) -> id { unsafe { NSWindow::alloc(nil) } }
+                let superclass = Class::get("NSWindow").unwrap();
+                let mut decl = ClassDecl::new("WinitNSWindow", superclass).unwrap();
+                decl.add_method(sel!(keyDown:), on_key_down as extern fn(&Object, Sel, id));
+                decl.add_method(sel!(alloc:), alloc as extern fn(&Object, Sel, id) -> id);
+                decl.register();
+            });
+            let window: id = msg_send![Class::get("WinitNSWindow").unwrap(), alloc];
+
+            let window = IdRef::new(window.initWithContentRect_styleMask_backing_defer_(
                 frame,
                 masks,
                 appkit::NSBackingStoreBuffered,
```

francesca64:
Fixed via #511

PR #511, opened 2018-05-10
fix(mac_platform): forward keyevent to system

### LINKED ISSUE #496: macOS: IME input source switch doesn't work on some key
Reproduce steps: 

- specify `f18` as input source switch key in mac os prefs 
- `cargo run --example window`, focus to winit example window, try to switch input source via assigned keyboard shortcut. 

winit itself does receive key event around those
```
WindowEvent { window_id: WindowId(Id(140455570389456)), event: KeyboardInput { device_id: DeviceId(DeviceId), input: KeyboardInput { scancode: 79, state: Pressed, virtual_keycode: None, modifiers: ModifiersState { shift: false, ctrl: false, alt: false, logo: false } } } }
WindowEvent { window_id: WindowId(Id(140455570389456)), event: ReceivedCharacter('\u{f715}') }
WindowEvent { window_id: WindowId(Id(140455570389456)), event: KeyboardInput { device_id: DeviceId(DeviceId), input: KeyboardInput { scancode: 79, state: Released, virtual_keycode: None, modifiers: ModifiersState { shift: false, ctrl: false, alt: false, logo: false } } } }
```

but system doesn't switch input only if winit window got focused. For some other typical key combinations (i.e option+space) switching works, so hard to guess what exactly causes this problem.

kwonoj:
- closes https://github.com/tomaka/winit/issues/496

This PR intend to fix #496 - reason system doesn't switch input method is winit doesn't forward NSKeyDown to system specifically, so only application aware of it.

Still, it is bit unsure about historical reason why keydown event was prevented from first time. Creating PR to discuss if there's better / viable approach than this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - `Icon::to_cardinals` is no longer public, since it was never supposed to be.
 - Wayland: improve diagnostics if initialization fails
+- Fix some system event key doesn't work when focused, do not block keyevent forward to system on macOS
 
 # Version 0.14.0 (2018-05-09)
 
--- a/src/platform/macos/events_loop.rs
+++ b/src/platform/macos/events_loop.rs
@@ -294,10 +294,7 @@ impl EventsLoop {
 
         // FIXME: Document this. Why do we do this? Seems like it passes on events to window/app.
         // If we don't do this, window does not become main for some reason.
-        match event_type {
-            appkit::NSKeyDown => (),
-            _ => appkit::NSApp().sendEvent_(ns_event),
-        }
+        appkit::NSApp().sendEvent_(ns_event);
 
         let windows = self.shared.windows.lock().unwrap();
         let maybe_window = windows.iter()
--- a/src/platform/macos/window.rs
+++ b/src/platform/macos/window.rs
@@ -415,13 +415,13 @@ impl WindowDelegate {
             // callbacks for drag and drop events
             decl.add_method(sel!(draggingEntered:),
                 dragging_entered as extern fn(&Object, Sel, id) -> BOOL);
-           decl.add_method(sel!(prepareForDragOperation:),
+            decl.add_method(sel!(prepareForDragOperation:),
                 prepare_for_drag_operation as extern fn(&Object, Sel, id));
-           decl.add_method(sel!(performDragOperation:),
+            decl.add_method(sel!(performDragOperation:),
                 perform_drag_operation as extern fn(&Object, Sel, id) -> BOOL);
-           decl.add_method(sel!(concludeDragOperation:),
+            decl.add_method(sel!(concludeDragOperation:),
                 conclude_drag_operation as extern fn(&Object, Sel, id));
-           decl.add_method(sel!(draggingExited:),
+            decl.add_method(sel!(draggingExited:),
                 dragging_exited as extern fn(&Object, Sel, id));
 
             // callbacks for fullscreen events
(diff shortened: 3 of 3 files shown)


francesca64:
Thanks for contributing!

> Still, it is bit unsure about historical reason why keydown event was prevented from first time.

Well, I dug into the blame, and I found the reason.

44e954895fd6c3497c8b5483e1b3efe160429329 (tomaka/glutin#384)

Related: #131

kwonoj:
womp, what would be best way to resolve? prolly convert #131 into mergeable PR instead of this change?

francesca64:
The macOS backend has changed a lot since #131, so most of the relevant code is already in place: 
https://github.com/tomaka/winit/blob/102dd07456f6c283f05adf678f84d6c0e7c3925e/src/platform/macos/window.rs#L688-L694

As far as I can tell, it's just a matter of adding `extern fn on_key_down(_this: &Object, _sel: Sel, _id: id) { }` and `decl.add_method(sel!(keyDown:), on_key_down as extern fn(&Object, Sel, id));`. (The lines regarding `alloc` are already handled in other places.)

kwonoj:
Let me try to read diffs as suggested and try to update PR. (if I can't, I'll call off and will let you know too) appreciate for help @francesca64 🙇

kwonoj:
@francesca64 ok, updated PR - manual verification on local system works for forwarding input method change correctly and also doesn't raise alert sound.

CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 # Unreleased
+- Fix some system event key doesn't work when focused, do not block keyevent forward to system
```
francesca64: You need to specify which platform this change is for.

src/platform/macos/window.rs
```diff
@@ -415,14 +415,14 @@ impl WindowDelegate {
...
-           decl.add_method(sel!(prepareForDragOperation:),
-                prepare_for_drag_operation as extern fn(&Object, Sel, id));
-           decl.add_method(sel!(performDragOperation:),
-                perform_drag_operation as extern fn(&Object, Sel, id) -> BOOL);
-           decl.add_method(sel!(concludeDragOperation:),
-                conclude_drag_operation as extern fn(&Object, Sel, id));
-           decl.add_method(sel!(draggingExited:),
-                dragging_exited as extern fn(&Object, Sel, id));
+            decl.add_method(sel!(prepareForDragOperation:),
+                    prepare_for_drag_operation as extern fn(&Object, Sel, id));
```
francesca64: This line should only be indented once.

src/platform/macos/window.rs
```diff
@@ -686,10 +686,13 @@ impl Window2 {
         static INIT: std::sync::Once = std::sync::ONCE_INIT;
 
         INIT.call_once(|| unsafe {
+            extern fn on_key_down(_this: &Object, _sel: Sel, _id: id) { }
```
francesca64: `{ }` should be `{}`

CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 # Unreleased
```
francesca64: There should be a newline between the heading and the change list.

src/platform/macos/events_loop.rs
```diff
@@ -295,7 +295,6 @@ impl EventsLoop {
         // FIXME: Document this. Why do we do this? Seems like it passes on events to window/app.
         // If we don't do this, window does not become main for some reason.
         match event_type {
```
francesca64: This `match` block can be replaced with `appkit::NSApp().sendEvent_(ns_event);` now.

kwonoj:
Updated PR per suggestion.

francesca64:
Okay, I tested this, and everything seems great!

PR #515, opened 2018-05-13
X11: Fix get_current_monitor

### LINKED ISSUE #64: [Linux] Multiple monitors identified as a single one
Original: https://github.com/tomaka/glutin/issues/678

francesca64:
Fixes #64 

This also adds an `eprintln!` to the X11 error handler. Since many X11 errors occur asynchronously, this should make debugging a bit easier. I still plan on adding proper logging in future work, but for now, this serves as a convenient preview for the panic that will happen soon afterwards.

Another notable change is that the monitor list is cached now. `XRRGetScreenResources` is [supposedly very slow](https://people.gnome.org/~ssp/randr/TODO), and XRandR can notify us when we need to invalidate the cached list. Previously, Alacritty would query and reconstruct the monitor list 3 times, but now it only needs to once, so hooray for minor startup time improvements. (This will be more important when full HiDPI support for X11 lands, since detecting DPI changes involves checking which monitor a window is on whenever it moves.)

The rest of this is just some refactoring to reduce code duplication between the XRandR 1.5 and <1.5 paths.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - On X11, `Window::get_current_monitor` now reliably returns the correct monitor.
 - On X11, `Window::hidpi_factor` returns values from XRandR rather than the inaccurate values previously queried from the core protocol.
 - On X11, the primary monitor is detected correctly even when using versions of XRandR less than 1.5.
+- `MonitorId` now implements `Debug`.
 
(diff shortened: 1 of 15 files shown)


francesca64:
Bonus: `MonitorId` now implements `Debug` on all platforms.

PR #518, opened 2018-05-15
macOS: Implement NSTextInputClient

francesca64:
Issues this hopefully fixes:
- #263
- jwilm/alacritty#61
- jwilm/alacritty#209
- jwilm/alacritty#372
- jwilm/alacritty#748 ✔️
- jwilm/alacritty#764 ✔️
- jwilm/alacritty#935 ✔️
- jwilm/alacritty#1040 ✔️
- jwilm/alacritty#1041 ✔️
- ~~jwilm/alacritty#1064~~ ❌

(I'm not sure about all of these. Hopefully we can get feedback on most of them.)

Alacritty branch for easy testing: https://github.com/francesca64/alacritty/tree/macos-kb

@kwonoj are you interested in trying this out?

PR #520, opened 2018-05-15
macOS: Fix re-enabling decorations after the window is built without them

### LINKED ISSUE #517: Yet another macOS decoration regression
1. Create window `with_decorations(false)`
2. `set_decorations(true)` - only the buttons appear

francesca64:
Fixes #517

These properties are changed *only* when using `with_decorations(false)`, and *not* when calling `set_decorations(false)`: 
https://github.com/tomaka/winit/blob/d86f53a02c0b17f76551c6b8917ecafe4f1f5164/src/platform/macos/window.rs#L781-L784

Changing those properties back to the default when calling `set_decorations(true)` solves the problem. @sodiumjoe do you know why those calls are used for `with_decorations(false)`? It seems to work fine without them, especially considering they're never used when disabling decorations after the fact.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 Cargo.lock
 target/
+rls/
 *~
 #*#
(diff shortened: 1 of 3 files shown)


sodiumjoe:
That might've been vestigial code from when I was trying to implement `decoration=false` as resizeable. It seems safe to remove that block to me.

francesca64:
Ah, thanks. Also, fun fact from #518: using `setStyleMask:NSBorderlessWindowMask` seems to unset the first responder. It's really weird.

PR #526, opened 2018-05-17
Windows: CursorState improvements

### LINKED ISSUE #523: set_cursor_state(CursorState::Grab) not working correctly on Windows 8.1
When I call `display.gl_window().window().set_cursor_state(CursorState::Grab)` every frame, at first it works, the cursor is hidden and can't be moved out of the window with the mouse.
But I can still ALT+TAB out of the Window, and then, when I move the cursor back over the window, it's not being grabbed even though I call that function every frame!
After that it behaves like `Hide`, i.e. the cursor is hidden while over the window, but it's not grabbed.

francesca64:
Fixes #523

Prior to changing the cursor state, we now check the current grab state, since it can be invalidated by alt-tabbing and other things.

`CursorState::Hide` is also implemented now.

The cursor name is now wrapped in a `Cursor` struct to allow multithreaded access.

`Window::set_cursor_state` has been reworked to use `execute_in_thread`. Two unneeded `transmute` calls were also removed.

The `WM_SETCURSOR` handler is much more readable now.

`MonitorId::get_adapter_name` has been removed, since it's dead code and appears to be a relic from 4 years ago.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@
 - Added the `Window::set_ime_spot(x: i32, y: i32)` method, which is implemented on X11 and macOS.
 - **Breaking**: `os::unix::WindowExt::send_xim_spot(x: i16, y: i16)` no longer exists. Switch to the new `Window::set_ime_spot(x: i32, y: i32)`, which has equivalent functionality.
 - Fixed detection of `Pause` and `Scroll` keys on Windows.
+- On Windows, alt-tabbing while the cursor is grabbed no longer makes it impossible to re-grab the window.
+- Implemented `CursorState::Hide` on Windows.
 
 # Version 0.14.0 (2018-05-09)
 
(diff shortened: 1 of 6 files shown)


francesca64:
...apparently `CursorState::Hide` was already implemented, but the code was written in such a way as to make that quite unclear.

My implementation doesn't automatically hide the cursor when grabbing, which I think is a better way to go, since that gives you more flexibility. The Windows and macOS backends both make an additional call to handle this for the user (this is absent from the X11 backend, which is presumably why X11 is the only one to have `MouseCursor::NoneCursor` implemented). 

Any thoughts on this? I think it's strictly the right behavior (it's a more direct mapping to the behavior of at least 3 platforms), though it requires the user to also call `set_cursor(MouseCursor::NoneCursor)` if they want to grab and hide the cursor (and explicitly set it back to the default afterwards). I wouldn't be opposed to adding something like `CursorState::GrabAndHide` if it seems like enough of an ergonomic issue (which would also allow for slightly more instantaneous cursor hiding on Windows, since cursor changes don't apply until the cursor moves, whereas cursor hiding doesn't have that problem).

francesca64:
So, I've thought more about, and I'm going to reduce this PR to just the bug fix and cleanup. I'd rather make a proper proposal about `CursorState`, since I'm now convinced that more thought needs to go into it, and I can give that issue a fancy title like "False dichotomies in `CursorState`", which is always a plus.

PR #529, opened 2018-05-19
X11: Improve hint support

### LINKED ISSUE #257: Add platform-specific window code
What I'm proposing is that platform-specific window code be added to winit, such as override-redirect, setting Atoms, etc. on Xorg. This would be very useful to my current project, and allow it to be implemented in a much less hacky way. I'd be willing to write a PR for this if you would be willing to potentially merge it.

francesca64:
Fixes #257

After this, Alacritty also won't have to directly make calls to Xlib anymore. The hints needed there seemed reasonable to provide out of the box, i.e. setting a hint that contains the PID. There's also a behavior change included here, in that `WM_CLASS` is no longer based on the window title. It was a semantically incorrect and rather surprising thing to do in the first place; now it defaults to the executable name (which the spec specifies as a fallback) and you can explicitly set `WM_CLASS` using `WindowBuilderExt::with_class`.


francesca64:
Unless any surprise PRs/bug reports come in, I plan on releasing 0.15.0 after this.

PR #532, opened 2018-05-21
X11: Fix primary monitor fallback regression

### LINKED ISSUE #507: winit panicks failing to find x11 monitor
copied from https://github.com/tomaka/glutin/issues/1018:

When I iterate over available monitors with events loop, nothing is returned, and attempting to call `get_primary_monitor()` will crash:

```
extern crate glium;
use glium::glutin;

fn main () {
  {
    let events_loop = glutin::EventsLoop::new();
    let mut available_monitors_count = 0; 
    for i in events_loop.get_available_monitors() {
      available_monitors_count += 1;
    }
    println!("avilable monitors count: {}", available_monitors_count);  // prints '0'
    let primary_monitor = events_loop.get_primary_monitor(); // crash 1
  }
  let events_loop = glutin::EventsLoop::new();  // crash 2
}
```

crash 1 with: 

```
thread 'main' panicked at '[winit] Failed to find any x11 monitor', libcore/option.rs:916:5
```

If crash 1 is commented out, then code will reach crash 2 with:

```
thread 'main' panicked at 'Failed to call XInternAtom: XError { description: "BadRequest (invalid request code or no such operation)", error_code: 1, request_code: 140, minor_code: 42 }', libcore/result.rs:945:5
``` 

Previously the iter loop would show some available monitors, and the primary monitor would be returned, but I'm not sure exactly what has changed since the last time it worked. This is on end of life LTS Ubuntu (14.04) so it may be something out of date.

francesca64:
Fixes #507

Apparently I understand X11 esoterica better than I understand iterators.

### DIFF
--- a/src/platform/linux/x11/monitor.rs
+++ b/src/platform/linux/x11/monitor.rs
@@ -19,7 +19,20 @@ const FORCE_RANDR_COMPAT: bool = false;
 const DISABLE_MONITOR_LIST_CACHING: bool = false;
 
 lazy_static! {
-    static ref MONITORS: Mutex<Option<Vec<MonitorId>>> = Mutex::new(None);
+    static ref XRANDR_VERSION: Mutex<Option<(c_int, c_int)>> = Mutex::default();
+    static ref MONITORS: Mutex<Option<Vec<MonitorId>>> = Mutex::default();
+}
+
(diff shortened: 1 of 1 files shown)


francesca64:
Also, I didn't put a CHANGELOG entry on this, since this regression is "self contained" within the current unreleased version.

Issue #531, opened 2018-05-20
Minimum supported Rust version


francesca64:
I've noticed that we don't seem to pin to a specific stable Rust version on CI. Now that 1.26 changed `match` behavior, it's [very likely](https://github.com/tomaka/winit/pull/530/files#diff-688dd1ea014c7e6b749bcda522984f51L55) someone's accidentally going to make winit unusable for anyone using an older version.

So, how far back do we want to support?

vberger:
[`wayland-client` requires rust >= 1.20.0](https://github.com/Smithay/wayland-rs/blob/master/.travis.yml#L10), and this is tested so unlikely to change unless absolutely needed.

So, I guess this puts a lower-bound onto what winit can require.

b-r-u:
`parking_lot` requires Rust 1.24.0 and with some small changes `winit` also compiles with 1.24.0.

francesca64:
1.20.0 is what I was hoping for, since I have some ideas for using associated constants.

I didn't realize that `parking_lot` requires 1.24.0, or I would've raised this issue sooner. I'd say "well, it's probably fine, since no one's complained about it", but that doesn't mean anything, since almost no one's used winit 0.14.0 in the first place.

Though, I don't know how important it really is that we support older versions; I think the main argument is for people who install Rust from a package manager instead of `rustup`?

PR #539, opened 2018-05-29
appveyor.yml: Test additional Rust channels

b-r-u:
In addition to nightly, also test the current stable version and Rust
1.24.1.

This might help with #531.


francesca64:
Thanks!

Issue #540, opened 2018-05-31
with_resizable / set_resizable


dannyfritz:
Be able to disable resizing of the window. This would include removing the resize cursors when hovering over the edge of the window.

Related to #65 and https://github.com/tomaka/glutin/issues/877

|                        | Windows | X11 | Wayland | MacOS | iOS | Emscripten | Android |
| --------------- | -----------| ---- | ----------| ---------| --- | ------------ | --------- |
| `with_resizable` | ✔#541 | ✔#556 | ✔ #565 | ✔#553 | NA | NA | NA |
| `set_resizable`   | ✔#558 | ✔#558  | ✔ #565 | ✔#553 | NA | NA | NA |

francesca64:
Thanks for reporting this. AFAIK this should be straightforward to implement on X11, Windows, and macOS.

@vberger is this possible on Wayland?

dannyfritz:
I'm playing around with this to see if I can get it to work for Windows. Removing `WS_SIZEBOX` from the `styles` seems to mess up a lot more than just preventing resizing.

vberger:
Right now it is already doable (I believe on all platforms) by setting a min_size and max_size, both with the same values.

It'll require some work on smithay's client toolkit to remove the "resize" cursor hint from the decorations as well on wayland, but it should not be a big thing.

dannyfritz:
I made [a commit to attempt `X11`](https://github.com/dannyfritz/winit/commit/50ba0fdb90d16bd610b15f11946481785e2c50f2), I haven't played around with it too much, but it seems using `_NET_WM_ALLOWED_ACTIONS` and `_NET_WM_ACTION_RESIZE` isn't having any effect.

dannyfritz:
[Here](https://github.com/dannyfritz/winit/commit/ad3f52adb91a14ab12d2499c0ffd08175e283b4a) is where I'm at now.

It crashes intermitedly with a `TypeMismatch(0)` Error from `xconn::get_property`.
I'm not sure if it is just my X11 server, but when it does successfully run without erroring and removes `_NET_WM_ACTION_RESIZE`, it doesn't seem to take hold as I can still resize the window.

---

`ffi::XA_ATOM` took me the longest time to figure out. I was doing this for the longest time:

```rust
let array_atom = unsafe { self.xconn.get_atom_unchecked(b"ATOM[]\0") };
```

francesca64:
[...]

PR #541, opened 2018-05-31
Windows: implement with_resizable (#540)

dannyfritz:
Only implemented `WindowBuilder::with_resizable` for MS Windows.
regarding #540

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - On X11, the `Moved` event is no longer sent when the window is resized without changing position.
 - `MouseCursor` and `CursorState` now implement `Default`.
+- `WindowBuilder::with_resizable` implemented for Windows.
 
 # Version 0.15.0 (2018-05-22)
 
--- a/examples/no_resize.rs
+++ b/examples/no_resize.rs
@@ -4,7 +4,7 @@ fn main() {
     let mut events_loop = winit::EventsLoop::new();
 
     let _window = winit::WindowBuilder::new()
-        .with_title("A non-resiable window!")
+        .with_title("A non-resizable window!")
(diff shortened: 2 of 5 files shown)


francesca64:
That was fast! I'll test this a bit later, but for now:
1. You need to add an entry to the CHANGELOG.
2. The documentation should mention that it's only implemented on Windows (even though I'll surely add the others before the next release).

dannyfritz:
Alright, modified docs and CHANGELOG. I might get to X11 this afternoon.

francesca64:
Cool. [`_NET_WM_ALLOWED_ACTIONS`](https://specifications.freedesktop.org/wm-spec/1.3/ar01s05.html) should be the way to go on X11. Let me know if you don't understand all the business with atoms and properties, though it's hopefully intuitive given the existing code in the backend.

francesca64:
Alright, I finally tested this out. I just corrected a little typo, so I'll merge this once CI passes again (I know it doesn't matter, but I'm paranoid).

francesca64:
Though, it's still possible to maximize the window. I'm guessing you probably don't want that.

dannyfritz:
I had considered it, but I came across this deep in the issue tracker saying probably not to: https://github.com/tomaka/winit/pull/127#issuecomment-276293336

francesca64:
That was a slightly different issue, since it was about min/max being equal being implicitly prescribed the semantics that you've explicitly introduced here.

Disabling resizing on macOS automatically disables the green button (which is used both for maximization and fullscreen), so it's probably more consistent to disable the maximize button here as well (I wouldn't worry about `set_fullscreen`/`set_maximized` though, since it's fine if those still work, and likely preferable). I took care of the macOS implementation here: #553

PR #542, opened 2018-05-31
Added a GitHub PULL_REQUEST_TEMPLATE

dannyfritz:
I figured this could be useful for making sure contributors are aware of what changes they have to make.


francesca64:
Thanks! This is a really good idea, since about 50% of the time I have to tell someone to add a CHANGELOG entry.

I've made some changes to better reflect my expectations. Hopefully from now on things will be a bit more efficient around here.

PR #544, opened 2018-06-01
Fix DPI with 0 width/hight reported by xorg

### LINKED ISSUE #543: HiDPI: Physical size from xrandr may be unreliable (can be 0mm x 0mm)
I ran into this issue when trying to run `Alacritty`, which uses `winit`, on my GPD Pocket with ArchLinux running GNOME 3.28.2 on Xorg. When starting the program, `winit` will report the following error:

```
[winit X11 error] XError {
    description: "BadValue (integer parameter out of range for operation)",
    error_code: 2,
    request_code: 12,
    minor_code: 0
}
```

Further investigation showed that `XConnection::get_output_info` returned `inf` for the HiDPI factor. Reading the source code I realized that the value is calculated from the resolution and the physical size of the screen, which are all fetched from `xrandr`. Trying to run `xrandr` directly to see what's happening, and I ended up with

```
Screen 0: minimum 320 x 200, current 2496 x 1560, maximum 8192 x 8192
DSI-1 connected primary 2496x1560+0+0 right (normal left inverted right x axis y axis) 0mm x 0mm
   1200x1920     60.38*+
   1920x1200     59.88    59.95  
   1920x1080     60.01    59.97    59.96    59.93  
   1600x1200     60.00  
   1680x1050     59.95    59.88  
   1400x1050     59.98  
   1600x900      59.99    59.94    59.95    59.82  
   1280x1024     60.02  
   1400x900      59.96    59.88  
   1280x960      60.00  
   1440x810      60.00    59.97  
   1368x768      59.88    59.85  
   1280x800      59.99    59.97    59.81    59.91  
   1280x720      60.00    59.99    59.86    59.74  
   1024x768      60.04    60.00  
   960x720       60.00  
   928x696       60.05  
   896x672       60.01  
   1024x576      59.95    59.96    59.90    59.82  
   960x600       59.93    60.00  
   960x540       59.96    59.99    59.63    59.82  
   800x600       60.00    60.32    56.25  
   840x525       60.01    59.88  
   864x486       59.92    59.57  
   700x525       59.98  
   800x450       59.95    59.82  
   640x512       60.02  
   700x450       59.96    59.88  
   640x480       60.00    59.94  
   720x405       59.51    58.99  
   684x384       59.88    59.85  
   640x400       59.88    59.98  
   640x360       59.86    59.83    59.84    59.32  
   512x384       60.00  
   512x288       60.00    59.92  
   480x270       59.63    59.82  
   400x300       60.32    56.34  
   432x243       59.92    59.57  
   320x240       60.05  
   360x202       59.51    59.13  
   320x180       59.84    59.32
```

Note that the physical size reported by `xrandr` is absurd, since it is shown as `0mm x 0mm`. However, using `xdpyinfo` I got the normal physical size reported without problem:

```
screen #0:
  dimensions:    2496x1560 pixels (207x130 millimeters)
  resolution:    306x305 dots per inch
  depths (7):    24, 1, 4, 8, 15, 16, 32
  root window id:    0x16f
  depth of root window:    24 planes
```

I believe this should be some issue with xrandr, however I have saw some other cases with similar problems without a fix:

<https://bbs.archlinux.org/viewtopic.php?id=232288>
<http://xpra.org/trac/ticket/728>

I also confirmed this problem with @farseerfc who also owns the same computer as mine.

Since all the other programs on my system work just fine (including the ones using GTK3, QT5, Java Swing, which all support HiDPI), I am wondering if it is possible for `winit` to fallback to some other mechanism to determine hidpi factor or simply set it to 1 if the physical screen size cannot be detected. Although this should be a problem of somewhere else, I think we can at least get `winit` to run on these strange devices...

chrisduerr:
There is an issue with xorg reporting 0 width or height which leads to a division by zero with floating point numbers. This will result in a DPI factor of +inf. As a workaround to this when either width or height are 0 the DPI is now set to a fixed value (1.0).

In #543 the option of setting this as an environment variable has been discussed and I'd be happy to add this to the PR if it is desired.

This fixes #543.

### DIFF
--- a/src/platform/linux/x11/util/randr.rs
+++ b/src/platform/linux/x11/util/randr.rs
@@ -53,6 +53,11 @@ pub fn calc_dpi_factor(
     (width_px, height_px): (u32, u32),
     (width_mm, height_mm): (u64, u64),
 ) -> f64 {
+    // See http://xpra.org/trac/ticket/728 for more information
+    if width_mm == 0 || width_mm == 0 {
+        return 1.0;
+    }
+
     let ppmm = (
         (width_px as f64 * height_px as f64) / (width_mm as f64 * height_mm as f64)
     ).sqrt();
--- a/src/platform/linux/x11/util/randr.rs
+++ b/src/platform/linux/x11/util/randr.rs
@@ -1,4 +1,4 @@
-use std::{slice, env};
+use std::{env, slice};
 use std::str::FromStr;
 
 use super::*;
--- a/src/platform/linux/x11/util/randr.rs
+++ b/src/platform/linux/x11/util/randr.rs
@@ -57,6 +57,10 @@ pub fn calc_dpi_factor(
     // Override DPI if `WINIT_HIDPI_FACTOR` variable is set
     if let Ok(dpi_factor_str) = env::var("WINIT_HIDPI_FACTOR") {
         if let Ok(dpi_factor) = f64::from_str(&dpi_factor_str) {
+            if dpi_factor <= 0. {
+                panic!("Expected `WINIT_HIDPI_FACTOR` to be bigger than 0, got '{}'", dpi_factor);
+            }
+
             return dpi_factor;
         }
     }
--- a/src/platform/linux/x11/util/randr.rs
+++ b/src/platform/linux/x11/util/randr.rs
@@ -1,5 +1,5 @@
-use std::str::FromStr;
 use std::{slice, env};
+use std::str::FromStr;
 
 use super::*;
 use super::ffi::{
--- a/src/platform/linux/x11/util/randr.rs
+++ b/src/platform/linux/x11/util/randr.rs
@@ -1,4 +1,5 @@
-use std::slice;
+use std::str::FromStr;
+use std::{slice, env};
 
 use super::*;
 use super::ffi::{
@@ -53,6 +54,13 @@ pub fn calc_dpi_factor(
     (width_px, height_px): (u32, u32),
     (width_mm, height_mm): (u64, u64),
 ) -> f64 {
+    // Override DPI if `WINIT_HIDPI_FACTOR` variable is set
+    if let Ok(dpi_factor_str) = env::var("WINIT_HIDPI_FACTOR") {
+        if let Ok(dpi_factor) = f64::from_str(&dpi_factor_str) {
+            return dpi_factor;
+        }
+    }
+
     // See http://xpra.org/trac/ticket/728 for more information
     if width_mm == 0 || width_mm == 0 {
         return 1.0;
--- a/src/window.rs
+++ b/src/window.rs
@@ -346,7 +346,7 @@ impl Window {
     /// and two for a retina display.
     ///
     /// ## Platform-specific
-    /// On X11 the DPI factor can be overwritten using the `WINIT_HIDPI_FACTOR` environment
+    /// On X11 the DPI factor can be overridden using the `WINIT_HIDPI_FACTOR` environment
     /// variable.
     #[inline]
     pub fn hidpi_factor(&self) -> f32 {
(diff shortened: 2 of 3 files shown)


francesca64:
Thanks! With the number of regressions being reported, it looks like I'll have to put out a 0.15.1

Also, I'd like a CHANGELOG entry (I really need to merge #542), though I guess you should wait until the situation with env vars is decided.

chrisduerr:
This took quite a while for such a simple change, but the PR should be good to go now. Any comments are appreciated @francesca64.

src/window.rs
```diff
@@ -344,6 +344,10 @@ impl Window {
     /// Returns the ratio between the backing framebuffer resolution and the
     /// window size in screen pixels. This is typically one for a normal display
     /// and two for a retina display.
+    ///
+    /// ## Platform-specific
+    /// On X11 the DPI factor can be overwritten using the `WINIT_HIDPI_FACTOR` environment
```
francesca64: overwritten -> overridden

src/platform/linux/x11/util/randr.rs
```diff
@@ -1,4 +1,5 @@
-use std::slice;
+use std::str::FromStr;
+use std::{slice, env};
```
francesca64: This is the nitpickiest of nitpicks, but I'm trying to cultivate a consistent `use` style.
```rust
use std::{env, slice};
use std::str::FromStr;
```

chrisduerr: I love it :D. I was just following the default `rustfmt` import ordering.

francesca64: Thanks, but you didn't alphabetize the order of `env` and `slice`.

chrisduerr: Oh I didn't even notice that, sorry.

francesca64:
Oh, right, there's something important. There's nothing to stop users from specifying a DPI factor of 0.0 (or a negative DPI factor) which could cause all sorts of unexpected problems.

chrisduerr:
@francesca64 Is there a problem with that? If the user wants to break the application' that should be fine, right? It might be nice to mention in the documentation and it wouldn't be an issue to just make sure it's bigger than 0, but I don't see anything wrong with allowing the user to mess stuff up if he wants to do so.

francesca64:
They'd be semantically incorrect DPI factors, and 0.0 would result in countless divide by zeros throughout the entire software stack, whereas negative values would invert all positions and sizes (and in the case of `u32` casts, produce 0). It's not a minor breakage, but a "turn off sane laws of the universe" switch.

I can think of no reason to allow this. Besides, #548 asserts that the DPI factor is greater than 0 on every physical/logical conversion, so anyone trying to break things wouldn't get very far. Adding a check here and exiting with a helpful message seems practical.

chrisduerr:
So how would you go about this? Panic with an error or fix the value to 1.0 if the input is invalid?

francesca64:
I'm partial to panicking, since I like to be loud and explicit.

chrisduerr:
Loud and explicit sounds great to me, I'll put that in.

francesca64:
Awesome, thanks!

Tomorrow I'll look into the macOS regressions. I'll probably cut 0.15.1 around Friday, so if I miss anything let me know before then.

b-r-u:
We might as well go all in and use [`f64::is_normal`](https://doc.rust-lang.org/std/primitive.f64.html#method.is_normal) to protect against inf, NaN and subnormals!

chrisduerr:
Oh yeah that would have totally been a good idea, forgot about that. @francesca64 feel free to add that yourself if you're interested in that.

francesca64:
Thanks @b-r-u! One of the coolest things about Rust is that there's always something new to learn.

I'll add that in when I add the `DisplayWidthMM`/`DisplayHeightMM` fallback, and I'll be sure to include it in #548's assertions as well.

PR #547, opened 2018-06-02
Test against Rust 1.24.1 on Travis

### LINKED ISSUE #531: Minimum supported Rust version

francesca64:
Fixes #531

Does this look good to everyone? I don't really know how to test the config in advance, so I'm worried I'll break things.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,12 @@ matrix:
       addons:
         apt:
           packages: *i686_packages
+    - env: TARGET=i686-unknown-linux-gnu
+      os: linux
+      rust: 1.24.1
+      addons:
+        apt:
+          packages: *i686_packages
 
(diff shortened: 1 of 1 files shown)


b-r-u:
It looks good and I guess we will find out :)
https://travis-ci.org/tomaka/winit/builds/387236272

francesca64:
Ah, cool! I didn't realize it would use the new config.

PR #548, opened 2018-06-03
DPI for everyone

### LINKED ISSUE #62: Setting Cursor Position Inaccurate
Original: https://github.com/tomaka/glutin/issues/663

### LINKED ISSUE #105: Handle hidpi scaling.

### LINKED ISSUE #169: X11 HiDPI scaling algorithm is broken
As a well-known fact, X11 locks the DPI to 96 and the current code gets incorrect screen size, resulting in the scale factor always being 1.

The defacto standard is to use Xft.dpi to get font DPI.

### LINKED ISSUE #315: Move `hidpi_factor` to MonitorId
This would allow the device-pixel-ratio to be determined before creating a window. Alacritty's use case for this is showing a correctly dimensioned window at startup without having a flash of the incorrectly sized window. This is because terminal size is usually dictacted by lines/columns requested, and that requires the scale factor to compute.

### LINKED ISSUE #337: HiDPI macOS, mouse position is in physical pixels whereas windows creation happens in logical pixels
I'm using macOS with a HiDPI display, with the scale factor of 2.0.

There is an unfortunate mismatch in the API where the input mouse events are reported in physical pixels (they are scaled according to the scale factor reported by the window), whereas the window creation API naively requests a window of the size of the input values, without specifying the scale factor. A window of those logical pixel dimensions is created, but the physical pixels are double of that.

This leads the user of the library having to scale the mouse values down, or window dimensions up by themselves – and possibly fail to do that, if they are developing with scale factor of 1.0 and don't notice the problem.

Besides, the library documentation speaks mostly of "pixels", but "points" are mentioned too in a few APIs. I guess that the distinction refers to physical and logical pixels, but a explanation of those would be warranted in the documentation itself. (Note: please correct me if I'm wrong! Also, I'd be interested to know how if the distinction was actually considered when writing the docs – do the docs actually reflect some intention of separating those, or was `pixel` meant to be used as a general term referring both?)

I want to help with the current situation (at least for the macOS implementations), but I'm unaware what plans @tomaka has for the API.

Should a `WindowBuilder.with_dimensions(800, 600)` create a window of those logical dimensions? In that case, the docs should be changed to refer points, and methods should be added to get the scale factor. Or if `with_dimensions` was meant to create a window of those physical pixel dimensions, the implementation itself should be fixed, since at the moment it doesn't do that.

I'm going to send PRs once I get input from @tomaka what was the intention.

### LINKED ISSUE #469: 4k HiDPI monitor detected as 2560x1440 on Windows 10
OS: Windows 10 Home, Version 1708, build 16299.371
winit version: 0.11.3
GPU: Nvidia GTX 1080, driver 385.54
Scaling factor: 150%

```rust
events_loop.get_primary_monitor()
```

produces `MonitorId` with name = "\\.\DISPLAY1", dimensions = (2560, 1440), position = (0, 0), hidpi_factor = 1

```rust
winit::WindowBuilder::new().with_fullscreen(Some(events_loop.get_primary_monitor()))
```

changes monitor resolution to 2560x1440 instead of using native 3840x2160.

francesca64:
Fixes #62
Fixes #105
Fixes #169 
Fixes #315
Fixes #332 (big thanks to @kryptan for the initial work on Windows!)
Fixes #337
Fixes #469
Fixes tomaka/glutin#1025

Alright, it was quite the undertaking, but I believe I've finished this on X11, Windows, macOS, iOS, Android, and Emscripten.

Some caveats:
- The Intel OpenGL driver on Windows still has some problems in mixed-DPI environments. It doesn't happen with DirectX or Vulkan (and OpenGL with Nvidia works fine), and it will only affect people who either A) have multiple monitors of differing scale factors, or B) change the scale factor after login. The issue itself is just that there's a small black/background-colored gap below the titlebar; this doesn't happen without DPI awareness, but I think this is an acceptable trade-off. Besides, if users disagree, it's still possible to disable DPI awareness using `EventsLoopExt`.
- Android still returns a DPI factor of 1.0. As far as I can tell, getting the real value won't be possible until tomaka/android-rs-glue#116 is sorted out. However, once that's filled in the code I've added here should do all the right conversions.
- I couldn't really test Emscripten, since I was unable to get anything to actually run in my browser.
- iOS was an adventure, and I ended up reworking glutin's iOS backend (it's apparently been broken for quite some time). I still never got anything to render, but I can at least say that it reported the correct sizes and DPI factor. I've unfortunately introduced a small amount of GL-specific code to the iOS backend here; I plan on doing some restructuring and fixing that before this makes it into a release, but I didn't want to make this PR any more huge and monolithic than it is already.
- My solution for Wayland was just to comment things out in `src/platform/linux/mod.rs` and say "@vberger will figure this out later"

That said, this PR is functionally complete. The only reason I've labeled it WIP is because it's still completely undocumented! I just wanted an opportunity for feedback before I start fussing over doc comments. (I also noticed that I left a `// TODO: Adjust min+max size` in the Windows backend, whoops)

I won't reiterate what I wrote in `CHANGELOG.md`. Besides there, the other files I recommend looking at are `src/dpi.rs`and `src/window.rs`. You can obviously also test this, and I've provided some partial migrations of various things to help with that:
- [glutin](https://github.com/francesca64/glutin/tree/super-dpi) - everything should work except for the examples, and note that `GlWindow::resize` doesn't yet explicitly take `PhysicalSize`.
- [gfx pre-ll](https://github.com/francesca64/gfx/tree/super-dpi) - everything should work except for the examples.
- [alacritty](https://github.com/francesca64/alacritty/tree/super-dpi) - it seems fine on macOS, though on X11 it seems to "double upscale" on HiDPI displays. This is almost certainly a result of the math in the auto-resize Alacritty does after creation, but I couldn't find where it does that.

Migrating can be tedious, but isn't hard, and the more explicit API resulted in me fixing some bugs in my own applications. Overall, I'm pretty satisfied, but if anyone has objections, I'm in no rush to merge this. I want to release a 0.15.1 with various regression fixes before introducing any breaking changes like this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,10 @@
 
 - **Breaking:** Removed `VirtualKeyCode::LMenu` and `VirtualKeyCode::RMenu`; Windows now generates `VirtualKeyCode::LAlt` and `VirtualKeyCode::RAlt` instead.
 - On X11, exiting fullscreen no longer leaves the window in the monitor's top left corner.
-- On X11, exiting fullscreen no longer leaves the window in the monitor's top left corner.
-- The Windows backend is now DPI aware. `WindowEvent::HiDPIFactorChanged` is implemented, and `MonitorId::get_hidpi_factor` and `Window::hidpi_factor` return accurate values. Window creation along with returned positions and sizes now correctly use pixels rather than points.
+- The Windows backend is now DPI aware. `WindowEvent::HiDpiFactorChanged` is implemented, and `MonitorId::get_hidpi_factor` and `Window::hidpi_factor` return accurate values.
+- `WindowEvent::HiDpiFactorChanged` implemented on X11.
+- On macOS, `Window::set_cursor_position` is now relative to the client area.
+- On macOS, setting the maximum and minimum dimensions now applies to the client area rather than to the window.
 - **Breaking:** All deprecated methods have been removed.
 
 # Version 0.15.1 (2018-06-13)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - On Emscripten, `MonitorId::get_hidpi_factor` now returns the same value as `Window::get_hidpi_factor` (it previously would always return 1.0).
 - **Breaking:** The entire API for sizes, positions, etc. has changed. In the majority of cases, winit produces and consumes positions and sizes as `LogicalPosition` and `LogicalSize`, respectively. The notable exception is `MonitorId` methods, which deal in `PhysicalPosition` and `PhysicalSize`. See the documentation for specifics and explanations of the types. Additionally, winit automatically conserves logical size when the DPI factor changes.
 - **Breaking:** All deprecated methods have been removed. For `Window::platform_display` and `Window::platform_window`, switch to the appropriate platform-specific `WindowExt` methods. For `Window::get_inner_size_points` and `Window::get_inner_size_pixels`, use the `LogicalSize` returned by `Window::get_inner_size` and convert as needed.
+- HiDPI support for Wayland.
 
 # Version 0.15.1 (2018-06-13)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,11 @@
 
 - **Breaking:** Removed `VirtualKeyCode::LMenu` and `VirtualKeyCode::RMenu`; Windows now generates `VirtualKeyCode::LAlt` and `VirtualKeyCode::RAlt` instead.
 - On X11, exiting fullscreen no longer leaves the window in the monitor's top left corner.
-- **Breaking:** `Window::hidpi_factor` has been renamed to `Window::get_hidpi_factor` for better consistency. `WindowEvent::HiDPIFactorChanged` has been renamed to ``WindowEvent::HiDpiFactorChanged``. DPI factors are always represented as `f64` instead of `f32` now.
+- **Breaking:** `Window::get_hidpi_factor` has been renamed to `Window::get_hidpi_factor` for better consistency. `WindowEvent::HiDPIFactorChanged` has been renamed to ``WindowEvent::HiDpiFactorChanged``. DPI factors are always represented as `f64` instead of `f32` now.
 - The Windows backend is now DPI aware. `WindowEvent::HiDpiFactorChanged` is implemented, and `MonitorId::get_hidpi_factor` and `Window::hidpi_factor` return accurate values.
-- `WindowEvent::HiDpiFactorChanged` implemented on X11.
+- Implemented `WindowEvent::HiDpiFactorChanged` on X11.
 - On macOS, `Window::set_cursor_position` is now relative to the client area.
-- On macOS, setting the maximum and minimum dimensions now applies to the client area rather than to the window.
+- On macOS, setting the maximum and minimum dimensions now applies to the client area dimensions rather than to the window dimensions.
 - On iOS, `MonitorId::get_dimensions` has been implemented and both `MonitorId::get_hidpi_factor` and `Window::get_hidpi_factor` return accurate values.
 - On Emscripten, `MonitorId::get_hidpi_factor` now returns the same value as `Window::get_hidpi_factor` (it previously would always return 1.0).
 - **Breaking:** The entire API for sizes, positions, etc. has changed. In the majority of cases, winit produces and consumes positions and sizes as `LogicalPosition` and `LogicalSize`, respectively. The notable exception is `MonitorId` methods, which deal in `PhysicalPosition` and `PhysicalSize`. See the documentation for specifics and explanations of the types. Additionally, winit automatically conserves logical size when the DPI factor changes.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - On macOS, `Window::set_cursor_position` is now relative to the client area.
 - On macOS, setting the maximum and minimum dimensions now applies to the client area rather than to the window.
 - On iOS, `MonitorId::get_dimensions` has been implemented and both `MonitorId::get_hidpi_factor` and `Window::get_hidpi_factor` return accurate values.
+- On Emscripten, `MonitorId::get_hidpi_factor` now returns the same value as `Window::get_hidpi_factor` (it previously would always return 1.0).
 - **Breaking:** The entire API for sizes, positions, etc. has changed. In the majority of cases, winit produces and consumes positions and sizes as `LogicalPosition` and `LogicalSize`, respectively. The notable exception is `MonitorId` methods, which deal in `PhysicalPosition` and `PhysicalSize`. See the documentation for specifics and explanations of the types. Additionally, winit automatically conserves logical size when the DPI factor changes.
 - **Breaking:** All deprecated methods have been removed. For `Window::platform_display` and `Window::platform_window`, switch to the appropriate platform-specific `WindowExt` methods. For `Window::get_inner_size_points` and `Window::get_inner_size_pixels`, use the `LogicalSize` returned by `Window::get_inner_size` and convert as needed.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,9 @@
 - `WindowEvent::HiDpiFactorChanged` implemented on X11.
 - On macOS, `Window::set_cursor_position` is now relative to the client area.
 - On macOS, setting the maximum and minimum dimensions now applies to the client area rather than to the window.
-- **Breaking:** All deprecated methods have been removed.
+- **Breaking:** The entire API for sizes, positions, etc. has changed. In the majority of cases, winit produces and consumes positions and sizes as `LogicalPosition` and `LogicalSize`, respectively. The notable exception is `MonitorId` methods, which deal in `PhysicalPosition` and `PhysicalSize`. See the documentation for specifics and explanations of the types. Additionally, winit automatically conserves logical size when the DPI factor changes.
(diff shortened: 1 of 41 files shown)


vberger:
> My solution for Wayland was just to comment things out in `src/platform/linux/mod.rs` and say "@vberger will figure this out later"

Do you want me to take my #341 PR and integrate these changes into a PR to your PR branch ?

francesca64:
I'm glad I merged #547, since this failed on 1.24...

@vberger that's probably a good idea; besides making things cleaner, you integrating the new types before this merges would be an effective form of review for them.

CHANGELOG.md
```diff
@@ -3,6 +3,16 @@
 - On X11, the `Moved` event is no longer sent when the window is resized without changing position.
 - `MouseCursor` and `CursorState` now implement `Default`.
 - `WindowBuilder::with_resizable` implemented for Windows.
+- On X11, exiting fullscreen no longer leaves the window in the monitor's top left corner.
+- **Breaking:** `Window::get_hidpi_factor` has been renamed to `Window::get_hidpi_factor` for better consistency. `WindowEvent::HiDPIFactorChanged` has been renamed to ``WindowEvent::HiDpiFactorChanged``. DPI factors are always represented as `f64` instead of `f32` now.
```
rukai: I think it should be "`Window::hidpi_factor` has been renamed"

francesca64: Thanks!

rukai:
@vberger I migrated pf sandbox to use this branch https://github.com/rukai/PF_Sandbox/tree/super-dpi
And since your commit I'm hitting this issue.
```
error[E0277]: the trait bound `*mut wayland_sys::client::wl_display: std::marker::Send` is not satisfied in `wayland_client::display::DisplayInner`
   --> pf_sandbox/src/vulkan/mod.rs:221:25
    |
221 |         let draw_text = DrawText::new(device.clone(), queue.clone(), swapchain.clone(), &images);
    |                         ^^^^^^^^^^^^^ `*mut wayland_sys::client::wl_display` cannot be sent between threads safely
    |
    = help: within `wayland_client::display::DisplayInner`, the trait `std::marker::Send` is not implemented for `*mut wayland_sys::client::wl_display`
    = note: required because it appears within the type `wayland_client::display::DisplayInner`
    = note: required because of the requirements on the impl of `std::marker::Send` for `std::sync::Arc<wayland_client::display::DisplayInner>`
    = note: required because it appears within the type `wayland_client::display::Display`
    = note: required because of the requirements on the impl of `std::marker::Send` for `std::sync::Arc<wayland_client::display::Display>`
    = note: required because it appears within the type `winit::platform::platform::wayland::window::Window`
    = note: required because it appears within the type `winit::platform::platform::Window`
    = note: required because it appears within the type `winit::Window`
    = note: required by `vulkano_text::DrawText::new`
```
The issue probably occurs with any vulkano project but I havnt had time to test.
It looks like a similar issue to https://github.com/tomaka/winit/issues/472

francesca64:
You have no idea how much of a relief it is when a notification is for something that isn't my fault.

Shouldn't this have been caught by this test? https://github.com/tomaka/winit/blob/19dd961752319276fa170c07e46963de9e3b589c/tests/send_objects.rs#L11-L15

rukai:
running `cargo test` locally, it does actually fail although it fails in an example before it reaches your test:
```
error[E0277]: the trait bound `*mut wayland_sys::client::wl_display: std::marker::Send` is not satisfied in `wayland_client::display::DisplayInner`
  --> examples/proxy.rs:13:5
   |
13 |     std::thread::spawn(move || {
   |     ^^^^^^^^^^^^^^^^^^ `*mut wayland_sys::client::wl_display` cannot be sent between threads safely
   |
   = help: within `wayland_client::display::DisplayInner`, the trait `std::marker::Send` is not implemented for `*mut wayland_sys::client::wl_display`
   = note: required because it appears within the type `wayland_client::display::DisplayInner`
   = note: required because of the requirements on the impl of `std::marker::Send` for `std::sync::Arc<wayland_client::display::DisplayInner>`
   = note: required because it appears within the type `wayland_client::display::Display`
   = note: required because of the requirements on the impl of `std::marker::Send` for `std::sync::Weak<wayland_client::display::Display>`
   = note: required because it appears within the type `winit::platform::platform::wayland::event_loop::EventsLoopProxy`
   = note: required because it appears within the type `winit::platform::platform::EventsLoopProxy`
   = note: required because it appears within the type `winit::EventsLoopProxy`
   = note: required because it appears within the type `[closure@examples/proxy.rs:13:24: 19:6 proxy:winit::EventsLoopProxy]`
   = note: required by `std::thread::spawn`
```

francesca64:
Weird, `cargo test` passes for me, and I can run the proxy example fine in Weston.

@vberger feel free to just push directly to this branch.

rukai:
To confirm the issue, I ran the following commands on a Digtal ocean vps of mine
```
rustup default stable
rustup update
git clone -b super-dpi https://github.com/francesca64/winit
cd winit
cargo test
```
I still get the proxy.rs build failures I posted above.

Maybe you need to `cargo update` or maybe you have modified files that arnt checked in or maybe you are using the wrong branch?

Doesnt explain why the travis is passing though!?!

vberger:
Aaaah, ok, this is an unforseen consequence of my last release of wayland-client, sorry for that. I definitly need better testing... ><

I'll correct this asap.

vberger:
Alright, wayland-client 0.20.9 is yanked, wayland-client 0.20.10 is released with the fix, and I added tests for that. All should be good again.

rukai:
Thanks, that fixes it.

kryptan:
> The Intel OpenGL driver on Windows still has some problems in mixed-DPI environments. ... The issue itself is just that there's a small black/background-colored gap below the titlebar;

But I think coordinates of mouse events will still be computed correctly while rendering is incorrect. This will create a mismatch

src/platform/macos/window.rs
```diff
@@ -861,55 +890,38 @@ impl Window2 {
...
+            NSWindow::setContentSize_(*self.window, NSSize::new(size.width as CGFloat, size.height as CGFloat));
         }
     }
 
-    pub fn set_min_dimensions(&self, dimensions: Option<(u32, u32)>) {
+    pub fn set_min_dimensions(&self, dimensions: Option<LogicalSize>) {
         unsafe {
-            let (width, height) = dimensions.unwrap_or((0, 0));
-            nswindow_set_min_dimensions(self.window.0, width.into(), height.into());
+            let dimensions = dimensions.unwrap_or_else(|| (!0, !0).into());
```
yvt: Hello, I think this line should be:
```rust
let dimensions = dimensions.unwrap_or_else(|| (0, 0).into());
```

Otherwise `window.set_min_dimensions(None)` [crashes the application](https://gist.github.com/yvt/e54982a9199c95fb4678461a66e48edf) by changing the window size to a huge value.

francesca64: Whoops, good catch. Looks like I got carried away with copy-pasting.

francesca64:
@kryptan that's really unfortunate. Hopefully Intel fixes this soon, since AFAIK there's nothing we can do about it, and I don't think it makes sense to postpone this for something that hopefully only affects a small subset of users.

src/platform/linux/x11/util/randr.rs
```diff
@@ -2,20 +2,40 @@ use std::{env, slice};
...
+    // See http://xpra.org/trac/ticket/728 for more information.
+    if width_mm == 0 || width_mm == 0 {
+        return 1.0;
+    }
+
+    let ppmm = (
+        (width_px as f64 * height_px as f64) / (width_mm as f64 * height_mm as f64)
+    ).sqrt();
+    // Quantize 1/12 step size
+    ((ppmm * (12.0 * 25.4 / 96.0)).round() / 12.0).max(1.0)
```
chrisduerr: I don't really have a very deep understanding of this topic, but wouldn't the `96.0` here need to by the actual x11 screen DPI to properly support DPI on x11?

As far as I understand it 96.0 is only the default screen DPI on x11 and it could change based on a user's configuration. I'm not sure if in this context it's a different value that's always constant, but thought I'd just ask to properly understand this topic.

chrisduerr: Actually it might be possible that this `96.0` here is just used to calculate the DPI factor you'd need to apply to a screen with 96 DPI to get the "proper" DPI scaling. So this is probably all working as intended.

francesca64: I recommend reading the relevant issue #169 along with this blog post: http://wok.oblomov.eu/tecnologia/mixed-dpi-x11/

After that, you'll have an understanding with the same depth as mine, or probably better, since I know for a fact that I glossed over some things. I might not get a chance to look into that constant today, since I still have a Windows bug to investigate and a burning desire to relax for the rest of the day.

francesca64: Ah, GitHub only *sometimes* wants to update the page automatically, so I didn't see that. I do encourage being skeptical of this math, since I don't know who added it or when; I just know that it makes my eyes not hurt when using winit apps on my 1440p monitor.

chrisduerr: Thanks a lot for the clarification and the helpful links to the documentation. Until now I've mainly just made guesses based on the math in alacritty and winit and took a short read through the freetype documentation.

Looks like this will give me some more insights, very much appreciated!

vberger:
:tada: :tada:

PR #551, opened 2018-06-05
Windows creates Alt event instead of Menu event. (to match X11)

rukai:
Fixes https://github.com/tomaka/winit/issues/549

LMenu + RMenu is from the original batch of events copied from some windows list by tomaka https://github.com/tomaka/winit/commit/24baf4d10405129d1631efe82ae2f7a0cf18132b
While LAlt + RAlt were added later. https://github.com/tomaka/winit/commit/f86af01a99064ac1476498ea48370d4193ceda97

Because LAlt + RAlt is used on every other platform, and its a better description, I propose we delete LMenu + RMenu and point the windows key events LMenu + RMenu to LAlt + RAlt instead.

State of other platforms:
*   x11 uses LAlt + RAlt https://github.com/tomaka/winit/blob/master/src/platform/linux/x11/events.rs#L168
*   wayland uses LAlt + RAlt https://github.com/tomaka/winit/blob/master/src/platform/linux/wayland/keyboard.rs#L247
*   empscripton uses LAlt + RAlt https://github.com/tomaka/winit/blob/master/src/platform/emscripten/mod.rs#L637
*   mac os looks looks like it isnt implemented? https://github.com/tomaka/winit/blob/master/src/platform/macos/events_loop.rs#L605 whatever the case, I shouldnt touch mac os as I dont have access to a machine to test on.
*   android doesn't use VirtualKeyCode
*   ios doesn't use VirtualKeyCode


francesca64:
Thanks! This looks like a thorough investigation, and I'm definitely in favor of this change.

> mac os looks looks like it isnt implemented? 

Ha, that would explain why I get `virtual_keycode: None` when pressing the alt key on macOS. I'm surprised I never noticed. I'll tack a quick fix for that onto this, if you don't mind.

francesca64:
I just remembered that I'm not supposed to merge a breaking change before the 0.15.1 release, so this will have to wait until after I fix a few more bugs.

francesca64:
...`RWin` and `LWin` are also swapped on macOS. Don't tell anyone this, but I have a lot of trouble distinguishing left from right.

PR #554, opened 2018-06-06
X11: Fix super fun race conditions

### LINKED ISSUE #458: X11 crash: `[xcb] Unknown sequence number while processing queue`
The following error occurred at startup when running a native GUI program I'm working on. I've run a lot of GUI windows on X11 in the past year and I've never seen this error before, so thought I'd post it:

```
    Finished release [optimized + debuginfo] target(s) in 18.95 secs
     Running `target/release/audio_server`
[xcb] Unknown sequence number while processing queue
[xcb] Most likely this is a multi-threaded client and XInitThreads has not been called
[xcb] Aborting, sorry about that.
audio_server: xcb_io.c:259: poll_for_event: Assertion `!xcb_xlib_threads_sequence_lost' failed.
```

`rustc 1.26.0-nightly (9c9424de5 2018-03-27)`
`winit 0.12.0`

It has only occurred once and I'm not entirely sure how to recreate it yet.

Curiously, the error suggests calling `XInitThreads` however it looks to me like [we already do this](https://github.com/tomaka/winit/blob/107a1e7332992dc4fa5c9141c443c0635a007d9a/src/platform/linux/x11/xdisplay.rs#L39).

The only thing I can think to note is that I do use an [EventsLoopProxy](https://docs.rs/winit/0.12.0/winit/struct.EventsLoopProxy.html) to wakeup the main (GUI) thread from a separate audio monitoring thread. There's a chance that `EventsLoopProxy::wakeup` gets called prior to the a call to `EventsLoop::run_forever`. I have no idea if this has anything to do with the issue yet, just a thought.

francesca64:
Fixes #458 
Fixes amethyst/amethyst#707
Fixes jwilm/alacritty#1344

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - On X11, `with_override_redirect` now actually enables override redirect.
 - macOS now generates `VirtualKeyCode::LAlt` and `VirtualKeyCode::RAlt` instead of `None` for both.
 - On macOS, `VirtualKeyCode::RWin` and `VirtualKeyCode::LWin` are no longer switched.
+- Fixed race conditions when creating an `EventsLoop` on X11, most commonly manifesting as "[xcb] Unknown sequence number while processing queue".
 
 # Version 0.15.0 (2018-05-22)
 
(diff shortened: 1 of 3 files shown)


francesca64:
The Appveyor failure just seems to be a random network issue.

PR #556, opened 2018-06-07
X11: implement with_resizable (#540)

dannyfritz:

Setting the min/max to the same value creates a non-resizable window with X11.

regarding #540


francesca64:
Thanks! Looks good.

This still shows resize cursors on Xfwm4, but since it's fine on KWin, Mutter, Marco, Compiz, IceWM, Awesome, and Openbox, I'll let you off the hook.

dannyfritz:
Dang, do you have all those window managers installed locally? I only tested on Mutter with Gnome 3.

PR #557, opened 2018-06-07
Windows: Window::set_resizable

dannyfritz:

Related to #540.

I did expose the method through the wrapper `Window` class, so this PR might be blocked on the release of `0.16`.  As this will cause Android, iOS, and Emscripten to fail without at least stubs for methods.


dannyfritz:
Closing to combine with #558

Issue #463, opened 2018-04-14
macOS: winit emits a MouseInput event after resizing


sodiumjoe:
repro example:

```
extern crate winit;

fn main() {
    let mut events_loop = winit::EventsLoop::new();

    let _window = winit::WindowBuilder::new()
        .with_title("A fantastic window!")
        .build(&events_loop)
        .unwrap();

    events_loop.run_forever(|event| {
        match event {
            winit::Event::WindowEvent { event: winit::WindowEvent::Resized(x, y), .. } => {
                println!("resized: {},{}", x, y);
                winit::ControlFlow::Continue
            }
            winit::Event::WindowEvent { event: winit::WindowEvent::MouseInput { state, .. }, .. } => {
                println!("mouse input: {:?}", state);
                winit::ControlFlow::Continue
            }
            winit::Event::WindowEvent { event: winit::WindowEvent::Closed, .. } => {
                winit::ControlFlow::Break
            },
            _ => winit::ControlFlow::Continue,
        }
    });
}
```

sodiumjoe:
For some reason, after resizing, on mouse up, `appkit::NSLeftMouseUp` fires, and `appkit::NSLeftMouseDown` never fires. Seems like it might have something to do with resize blocking the event loop.

sodiumjoe:
So, with the `println`s added here: https://github.com/tomaka/winit/compare/master...sodiumjoe:463-resize?expand=1

the output looks like this on resize:
```
WindowEvent { window_id: WindowId(Id(140514968137152)), event: CursorEntered { device_id: DeviceId(DeviceId) } }
resizing
WindowEvent { window_id: WindowId(Id(140514968137152)), event: Resized(2040, 1526) }
resizing
WindowEvent { window_id: WindowId(Id(140514968137152)), event: Resized(2040, 1524) }
resizing
WindowEvent { window_id: WindowId(Id(140514968137152)), event: Resized(2040, 1522) }
WindowEvent { window_id: WindowId(Id(140514968137152)), event: MouseInput { device_id: DeviceId(DeviceId), state: Pressed, button: Left, modifiers: ModifiersState { shift: false, ctrl: false, alt: false, logo: false } } }
WindowEvent { window_id: WindowId(Id(140514968137152)), event: TouchpadPressure { device_id: DeviceId(DeviceId), pressure: 0.0, stage: 1 } }
[...]

PR #561, opened 2018-06-09
macOS: Only detect clicks+motion within client area

### LINKED ISSUE #560: CursorMoved and MouseInput events happening in title bar
On macOS High Sierra, it seems like I am getting window events that I shouldn't be, when clicking on the title bar / window decoration area.

Is this a bug or intentional? I think these window events should be restricted to happen only in the inner window. I even loaded up a C++ SDL project to check how they handle it, and it does ignore the equivalent events in the title bar.

francesca64:
Fixes #560


francesca64:
I was pleasantly surprised to find that this also fixes #463!

PR #565, opened 2018-06-13
wayland: implement set_resizable

vberger:

I believe this should be pretty trivial to review. :)

I tested it on weston, based on the glutin examples.


francesca64:
Excellent! Thanks.

PR #567, opened 2018-06-15
Mirror monitor list methods on Window

### LINKED ISSUE #555: `get_available_monitors()` / `get_primary_monitor()` ergonomics
Right now, `get_available_monitors()` and `get_primary_monitor()` are methods of `EventsLoop` - that means knowing which monitors are connected to the system is only possible in a context where `EventsLoop` can be borrowed, which creates some issues when writing abstractions.

If at all possible, it would be very useful if `Window` also had these methods - after all, it already has `get_current_monitor()`.

Concrete usecase (from [`ggez`](https://github.com/ggez/ggez)):

We have a struct, `Context`, that encapsulates away many of the inter-operation boilerplate of the engine, so it makes sense for it to own the `Window`. It cannot also own the `EventsLoop`, because `Context` needs to be available to make callbacks on based on the events within the loop.

Some of the callbacks are user-defined, and, it being a game engine and all, being able to tell the app to go fullscreen from within such a callback is desirable. Ideally, on any monitor, not just the current one, and that needs the data from `get_available_monitors()` and/or `get_primary_monitor()`.

(Current workaround is to cache data from `get_available_monitors()` within the `Context` at initialization - `EventsLoop` is borrowed during that time, to initialize a `Window`.)

francesca64:
Fixes #555

@vberger did I do this correctly on Wayland?


vberger:
the `self.monitors` field of the wayland window only tracks the monitors on which (part of) the window is currently displayed, so I think what you did is not what was intended.

Rather you'd need to store in the `Window` a handle of the `self.env.outputs` field of the EventLoop (it is of type [`sctk::OutputMgr`](https://docs.rs/smithay-client-toolkit/0.2.3/smithay_client_toolkit/output/struct.OutputMgr.html) which is already an `Arc`-like object, and thus can be cheaply cloned, and then mirror the implementations of the two functions from `EventLoop`.

vberger:
:+1:

Issue #538, opened 2018-05-28
Blur-Behind / Glassy Windows


Lisoph:
Moved from https://github.com/tomaka/glutin/issues/875

## About
Glassy windows are similar to transparent windows. They let the user see through the client area of the window, exposing whatever is behind it. The difference to transparent windows is, that glassy windows have a blur-filter applied to them.

The effect can look something like [this](http://i.stack.imgur.com/WWz1O.png) or [this](https://pbs.twimg.com/media/BpUou7pIIAE5cyL.jpg:large).

Transparency and blur are mutually exclusive.
## Implementation
On Windows, the WinAPI function `DwmEnableBlurBehindWindow` can be used to create the blur-behind effect (I think). On OS X, `NSVisualEffectView` is the way to do it.
Personally I don't know how to achieve the effect on X11, but I'm sure it's possible there as well.

Library wise this could be exposed via a new function to the WindowBuilder struct. The question being is: What do we specifiy as the parameters? Since the blur effect can be customized to achieve a specific look and feel, passing a bool is probably not sufficient.

### Further
Both of the Windows and OS X functions above only affect the client area of the window, however on both systems the title bar (and on Windows the border) can be made blurred as well.

It would probably be a good idea to implement seperate functions for controlling the client area and title bar / border blur.

Generally, I propose 2 functions similar to this:
```Rust
fn with_client_blur(self, tbd.) -> WindowBuilder<'a>;
fn with_decoration_blur(self, tbd.) -> WindowBuilder<'a>;
```
Names and signatures are of course debatable.

Comments and thoughts appreciated!

### Appendix 1
I have done some more research and found that on Windows 10, the undocumented function `SetWindowCompositionAttribute` needs to be used, as `DwmEnableBlurBehindWindow` is deprecated and does not properly work anymore on Windows 10. This function is is implemented in user32.dll.

francesca64:
> I'd be happy to implement this

Great! I haven't really had any substantial PRs to review in weeks now, and it's been getting lonely.

> Any pointers / suggestions to someone new to this codebase?

[...]

PR #568, opened 2018-06-15
Blur-Behind / Glassy Windows implementation

Lisoph:
Implements #538.

### Backend progress:
- [x] macOS
- [ ] Windows
- [ ] X
- [ ] Wayland
---


src/os/macos.rs
```diff
@@ -16,6 +16,9 @@ pub trait WindowExt {
     ///
     /// The pointer will become invalid when the `Window` is destroyed.
     fn get_nsview(&self) -> *mut c_void;
+
+    /// Just for testing purposes.
```
Lisoph: Should probably change that comment.

src/os/macos.rs
```diff
@@ -157,3 +165,27 @@ impl MonitorIdExt for MonitorId {
         self.inner.get_nsscreen().map(|s| s as *mut c_void)
     }
 }
+
+#[repr(i64)]
```
francesca64: Are these actually supposed to be `i64`, or are they an architecture-dependent size? I'd appreciate a comment including the `NSWhatever` name of the enum.

src/platform/macos/view.rs
```diff
@@ -38,8 +39,14 @@ pub fn new_view(window: id, shared: Weak<Shared>) -> IdRef {
     unsafe {
         // This is free'd in `dealloc`
         let state_ptr = Box::into_raw(Box::new(state)) as *mut c_void;
-        let view: id = msg_send![VIEW_CLASS.0, alloc];
-        IdRef::new(msg_send![view, initWithWinit:state_ptr])
+        let view_class = if win_attribs.blur { VISUAL_EFFECT_VIEW_CLASS.0 } else { VIEW_CLASS.0 };
+        let view: id = msg_send![view_class, alloc];
+        msg_send![view, initWithWinit:state_ptr];
+        if win_attribs.blur {
+            msg_send![view, setMaterial: BlurMaterial::Light as i64];
```
francesca64: You have to do `let _: () = msg_send!`, as otherwise segsegv/sigill will be triggered at runtime when the `!` (never) type is stabilized.

src/platform/macos/window.rs
```diff
@@ -3,6 +3,7 @@ use std::ops::Deref;
 use std::os::raw::c_void;
 use std::sync::Weak;
 use std::cell::{Cell, RefCell};
+use os::macos::BlurMaterial;
```
francesca64: I've been grouping imports into blocks, ordered:
1. `std`
2. Crates
3. Internal

With a newline between each block.

(I know this isn't consistently applied throughout the repo yet, but we're getting there.)

src/platform/macos/window.rs
```diff
@@ -582,6 +583,12 @@ impl WindowExt for Window2 {
     fn get_nsview(&self) -> *mut c_void {
         *self.view as *mut c_void
     }
+
+    #[inline]
+    fn set_blur_material(&self, material: BlurMaterial) {
+        let view = self.get_nsview() as *mut objc::runtime::Object;
+        unsafe { msg_send![view, setMaterial: material as i64]; }
```
francesca64: In every other usage, we call `msg_send!` on `*self.view`. Doing that is much cleaner (and has less potential overhead) than calling `get_nsview` like this.

Additionally, with `setMaterial: material`, from what I've seen Obj-C style has us writing `setMaterial:material`. This applies to the other occurrences as well, naturally.

src/lib.rs
```diff
@@ -482,6 +482,11 @@ pub struct WindowAttributes {
     /// [iOS only] Enable multitouch,
     /// see [multipleTouchEnabled](https://developer.apple.com/documentation/uikit/uiview/1622519-multipletouchenabled)
     pub multitouch: bool,
+
+    /// Whether the window should be blurry. This is similar to transparency.
```
francesca64: "blurry" has a pretty strong negative connotation. "blurred" sounds more neutral, but that has associations with input focus... "the window should have a blur effect" probably sounds the coolest, and then the second sentence should go on to explain things more.

src/window.rs
```diff
@@ -96,6 +96,15 @@ impl WindowBuilder {
     #[inline]
     pub fn with_transparency(mut self, transparent: bool) -> WindowBuilder {
         self.window.transparent = transparent;
+        self.window.blur = !transparent;
```
francesca64: What's the intention with the negation here?

Lisoph:
Once the changes are merged, I will submit a PR to glutin for compability with these changes. For the time being, in order to follow along, clone glutin and apply [these changes](https://github.com/Lisoph/glutin/commit/7f06a079592c78ead89107964b20752d6c9abe37). Also check for newer commits.

src/os/macos.rs
Lisoph: It's an `i64` based on the current `NSVisualEffectView.setMaterial` method signature. The method takes a c-style enum, which is represented as an `i64`. I don't think this is set in stone but I doubt it will ever change.

I have added some valuable documentation in my new commit.

src/window.rs
Lisoph: Transparency and blur are mutually exclusive. The negation ensures that transprency and blur are always set to the opposite of each other, so that both can never be `true` at the same time. This avoids running code paths for transparency in backends when blur is enabled (and vice versa), which in turn avoids potential bugs and improves performance.

Writing this comment, I noticed that this obviously creates a bug. When for example `with_blur(false)` is called, transparency will be enabled, which is wrong. I'm going to fix this.

src/os/macos.rs
```diff
@@ -16,6 +16,13 @@ pub trait WindowExt {
     ///
     /// The pointer will become invalid when the `Window` is destroyed.
     fn get_nsview(&self) -> *mut c_void;
+
+    /// For windows created with the [blurred](WindowBuilder::with_blur) option,
+    /// this controls the appearance of the blur effect.
+    /// 
+    /// Marked as unsafe because depending on the version of macOS and the `BlurMaterial` variant passed,
+    /// this might cause a crash.
```
francesca64: Is there a lowest common denominator of supported enum variants? I don't think people will be comfortable using a method that can crash based on criteria that are only vaguely explained.

Lisoph: The lowest common denominator appears to be `AppearanceBased`, `Light`, `Dark`, `MediumLight` and `UltraDark` ([source](https://developer.apple.com/documentation/appkit/nsvisualeffectview/material?changes=latest_major)), but those are now marked as deprecated. There are some marked as "modified", I think that means their underlying integer value has changed, which would disqualify them.  
This is all quite problematic since this swift enum is our only source of truth and it appears Apple is comfortable changing the order of constants willy nilly - they only keep the swift code consistent, but apparently not the enum's binary representation.

Honestly I don't know what to do here. This is why I was originally unsure about the `BlurMaterial` enum in the first place.  
The only alternative I see is exposing the raw `i64` and removing the enum, but that won't solve the crash / macOS-version dependency problem either.

francesca64: How impractical would it be for us to translate the values depending on the version of macOS?

Also, shouldn't it be `isize` instead of `i64`?

Lisoph: I don't think that would be too much work, definitely doable.

Yeah you could be right with `isize` over `i64`. The right size is whatever the size of the swift enum is. `isize` should be a good bet.

OvermindDL1:
Just for note, KDE5's wayland client supports blue-behind/glassy windows as long as the proper transparency is set for any given pixel.

vberger:
Does it actually blur the contents under it ? This PR is not about windows with a (partially) transparent background, but about asking the display server to blur the contents behind the transparent window.

OvermindDL1:
It is proper blur, came out in Plasma 5.13, example video (this is apparently an early version of it, bit more refined in the release version):
https://www.youtube.com/watch?v=xroL6ZWg2Gs

The settings are global, but any properly transparent window (or parts there-of) will have the system-set blur.   I'm unsure if there are per-window settings that are controllable, though knowing KDE Plasma that wouldn't surprise me if there is.

OvermindDL1:
Ah, correction, apparently it's enabled on a window when the window has the `_KDE_NET_WM_BLUR_BEHIND_REGION` property enabled on it (potentially other options?).

vberger:
Ah I see. Though, it seems that this `_KDE_NET_WM_BLUR_BEHIND_REGION` property is an x11 mechanism, not a wayland one. So it would not be applicable to native wayland clients.

OvermindDL1:
Using the API itself blue is enabled by `KWindowEffects::enableBlurBehind(winId, true);`, and this would enable it on both X11 and Wayland.

For note, KDE is only performing bug fixes for X11 currently, new development is exclusively Wayland.  The Blur uses the existing compositor functionality so it works on both.

OvermindDL1:
It seems you can even force blur on when it is default disabled via setting the `WindowForceBlurRole` attribute to true via the API.

OvermindDL1:
Documention I found this at starts with:
https://api.kde.org/4.x-api/kde-workspace-apidocs/kwin/libkwineffects/html/group__kwineffects.html
and
https://api.kde.org/4.x-api/kde-workspace-apidocs/kwin/libkwineffects/html/namespaceKWin.html

So at least KDE should be covered it seems once the calls are made.  Just remember to feature detect for Plasma API 5.13 or higher to use it.

vberger:
Ah so this would need to link winit against Qt/KDE... Not the kind of solution I was hoping for :/

OvermindDL1:
Dynamically at most, though KDE does have a habit of exposing most things over DBus as well, so it might be available there.

What is winit linking to as it stands?  Is it just direct X11/Wayland interfaces?  If so then I'm unsure how you would be able to control effects such as blur as those are handled by the compositor on each, of which Plasma is KDE's compositor (I think there is another DE or two that uses plasma as well) so compositor interfaces would have to be called regardless.  Plasma's Compositor is significantly the most complete on Wayland (it also has backwards compat with X11) and it even has a testing framework that does not require an active framebuffer (it has a virtual framebuffer, along the lines xvfb for X11 but one that works on Wayland as well) that is fantastic for performing automated testing (of which KDE has a substantial amount of), useful if you don't already have automated wayland testing.

vberger:
> What is winit linking to as it stands? Is it just direct X11/Wayland interfaces?

Currently yes. Ideally we would like to keep it that way as much as possible I believe.

For wayland, the ideal situation would be if Plasma exposed some KDE-specific protocol extension, I think. Though if not we'll need to work around that.

I think it may be worth opening an other issue to discuss how we should integrate WM-specific features in winit in general for X11/Wayland. This is getting out of scope for this PR, with only touches MacOS specific code.

OvermindDL1:
> For wayland, the ideal situation would be if Plasma exposed some KDE-specific protocol extension, I think. Though if not we'll need to work around that.

I'm like 90% certain it does as I'm pretty sure I ran across it a few months ago on their Wayland interfaces...

> I think it may be worth opening an other issue to discuss how we should integrate WM-specific features in winit in general for X11/Wayland. This is getting out of scope for this PR, with only touches MacOS specific code.

Perhaps the title should be changed to reflect it then?  The title implies nothing about Apple-specific interfaces at all?  ^.^;

:+1:

Lisoph:
> > I think it may be worth opening an other issue to discuss how we should integrate WM-specific features
> > in winit in general for X11/Wayland. This is getting out of scope for this PR, with only touches MacOS
> > specific code.

> Perhaps the title should be changed to reflect it then? The title implies nothing about Apple-specific
> Interfaces at all? ^.^;

This PR is for the implementation on all platforms. I'm currently working on the windows backend. The code is written, but it doesn't work... doh!

OvermindDL1:
> Thanks for the discussions, by the way. That will definitely help me on the Linux side.

Sure.  :-)

I'm mostly in the KDE world so that is what I primarily know (it's the only setup that does things like HiDPI and such 'mostly' right and so forth), but yeah it is the compositor that handles all that, and as far as I know there are 2 such compositors out, KDE's Plasma is one, and Compiz is the other.  I'm unsure if there are others but those are the two big ones, and only Plasma has full production support for Wayland at the current time.

Lisoph:
I just pushed a raw Windows implementation. If anyone want's to try it, just remember to resize the window. The client are needs to be redrawn for the blur effect to become visible.

By the way I accidentally `rustfmt`ed some files, which might have killed some manual formatting.

francesca64:
> By the way I accidentally `rustfmt`ed some files, which might have killed some manual formatting.

I won't review this until you revert that.

vberger:
@Lisoph 

> This PR is for the implementation on all platforms. [...]
> Thanks for the discussions, by the way. That will definitely help me on the Linux side.

Alright, my bad.

If you plan to do the wayland implementation as well, I expect this will require some digging and possible PRs to [wayland-protocols](https://github.com/Smithay/wayland-rs) and/or [sctk](https://github.com/Smithay/client-toolkit). I right now am quite overwhelmed but a lot of stuff and don't have the time to do all this digging around, but if someone manages to fond out what needs to be done, actually doing it should be pretty quick and straightforward.

sp3d:
@vberger, digging suggests that https://github.com/KDE/kwayland/blob/master/src/client/protocols/blur.xml is the source of truth for the KDE blur region protocol, so it should just be a matter of binding it.

fschutt:
Regarding KDE on X11 at least, `_KDE_NET_WM_BLUR_BEHIND_REGION` might be worth a look. As far as I am aware, it is a regular X11 Atom property like so many others.

Using xprop:

```
xprop -f _KDE_NET_WM_BLUR_BEHIND_REGION 32c -set _KDE_NET_WM_BLUR_BEHIND_REGION 0 -id $x11_window_id
```

friktor:
Hey. Can it be better to accept the ready changes for blur in mac os? I don`t think that in the near future anyone would want to be engaged in the implementation of blur for windows. But if you do not actualized this PR and do not accept it - over time, this work will depreciate, and someone will have to write it from scratch. @francesca64 @Lisoph

Lisoph:
@friktor That's fine with me. Waiting for @francesca64's opinion.

friktor:
@Lisoph Good) It's just that this PR has been active for almost a year now, and would it be better to decide the fate of this feature.

jansol:
@Lisoph did you ever revert the formatting you mentioned? GitHub isn't very good at showing forced pushes sometimes...

Osspial:
@Lisoph just fyi - @francesca64 is no longer working on Winit, I should be able to test this out on Windows, though.

src/platform/windows/events_loop.rs
```diff
@@ -473,7 +473,8 @@ pub unsafe extern "system" fn callback(
                 window_id: SuperWindowId(WindowId(window)),
                 event: Refresh,
             });
-            winuser::DefWindowProcW(window, msg, wparam, lparam)
+            //winuser::DefWindowProcW(window, msg, wparam, lparam) // TODO: Why was this here?
```
Lisoph: Like the comment says - DefWindowProc doesn't appear to make much sense for WM_PAINT. In fact it broke the redrawing code in the example. Can we remove this without breaking anything or am I missing something?

This line is also present in the most recent commit from upstream.

Lisoph:
> 
> 
> @Lisoph just fyi - @francesca64 is no longer working on Winit, I should be able to test this out on Windows, though.

@Osspial Thanks for the info! I'm still actively working on this (just fixed the modern Windows 10 implementation). Can we change to a different reviewer if @francesca64 is no longer working on winit? I'll try to merge with master and then take a look at the formatting.

ZeGentzy:
I can test this for linux, once it's impl'd. Can this be rebased against eventloops v2?

ZeGentzy:
Can someone rebase this PR against master and reopen it? It seams to have not aged well, so I'm closing it. Thanks.

PR #572, opened 2018-06-17
Shares the XConnection between all event pools instead of just all event pools on the same thread.

ZeGentzy:
Shares the XConnection between all event loops instead of just all event 
loops on the same thread.

This is needed for adding shared context support to [tomaka/glutin](https://github.com/tomaka/glutin), as contexts
must be made with the same native display (and therefore the same
connection.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - `EventsLoop::get_available_monitors` and `EventsLoop::get_primary_monitor` now have identical counterparts on `Window`, so this information can be acquired without an `EventsLoop` borrow.
 - `AvailableMonitorsIter` now implements `Debug`.
 - Fixed quirk on macOS where certain keys would generate characters at twice the normal rate when held down.
+- On X11, all event loops now share the same `XConnection`.
 
 # Version 0.15.1 (2018-06-13)
 
--- a/src/platform/linux/mod.rs
+++ b/src/platform/linux/mod.rs
@@ -6,6 +6,7 @@ use std::ffi::CStr;
 use std::os::raw::*;
 use std::sync::Arc;
 
+use parking_lot::Mutex;
 use sctk::reexports::client::ConnectError;
 
 use {
@@ -50,9 +51,9 @@ pub struct PlatformSpecificWindowBuilderAttributes {
     pub x11_window_type: x11::util::WindowType,
 }
 
-thread_local!(
-    pub static X11_BACKEND: Result<Arc<XConnection>, XNotSupported> = {
-        XConnection::new(Some(x_error_callback)).map(Arc::new)
+lazy_static!(
+    pub static ref X11_BACKEND: Mutex<Result<Arc<XConnection>, XNotSupported>> = {
+        Mutex::new(XConnection::new(Some(x_error_callback)).map(Arc::new))
     };
 );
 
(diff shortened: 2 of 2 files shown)


francesca64:
This doesn't look like it segfaults on [`multithread_window`](https://github.com/azriel91/multithread_window) any more often than on master, and I don't see any [XCB abort errors/etc](https://github.com/tomaka/winit/issues/458).

The segfault seems to happen most often when moving the cursor around, and especially across monitors... anyway, since it's not a regression, I don't expect you to fix that.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 - HiDPI support for Wayland.
 - `EventsLoop::get_available_monitors` and `EventsLoop::get_primary_monitor` now have identical counterparts on `Window`, so this information can be acquired without an `EventsLoop` borrow.
 - `AvailableMonitorsIter` now implements `Debug`.
+- In X11, all events loops will now share the same XConnection.
```
francesca64: "events loops" -> "event loops" (I know that's the actual type name, but it's a very awkward type name).

The other entries don't use the future tense.

Need backticks around `XConnection`.

src/platform/linux/mod.rs
```diff
@@ -26,6 +26,8 @@ use self::x11::{XConnection, XError};
 use self::x11::ffi::XVisualInfo;
 pub use self::x11::XNotSupported;
 
+use parking_lot::Mutex;
```
francesca64: This should be on line 9, above the `sctk` import.

src/platform/linux/mod.rs
```diff
@@ -441,14 +442,13 @@ r#"Failed to initialize any backend!
...
-        X11_BACKEND.with(|result| {
-            result
-                .as_ref()
-                .map(Arc::clone)
-                .map(x11::EventsLoop::new)
-                .map(EventsLoop::X)
-                .map_err(|err| err.clone())
-        })
+        let xconn = X11_BACKEND.lock();
+        xconn
```
francesca64: You can just call `lock()` as part of this method chain, which looks cooler.

src/platform/linux/mod.rs
```diff
@@ -357,29 +359,28 @@ unsafe extern "C" fn x_error_callback(
...
-                request_code: (*event).request_code,
-                minor_code: (*event).minor_code,
-            };
-
-            eprintln!("[winit X11 error] {:#?}", error);
-
-            *xconn.latest_error.lock() = Some(error);
-        }
-    });
+    let xconn = X11_BACKEND.lock();
```
francesca64: This variable name should be suffixed with `_lock`.

francesca64:
Thanks!

I tested this again, and now my impression is that this *does* segfault more often than on master. Are you seeing that too?

ZeGentzy:
No, I tested with this script:
```
#!/bin/bash

RUNS=1000

run(){
    f="$(cat <(cargo run 2>&1) | tail -n +2)"
    ret="$(diff <(echo "$f") <(cat suc) | wc -l)"
    if [ $ret -eq "0" ]; then
        echo "g" >> good
        echo g
    else
        echo "b" >> bad
        echo b
    fi
}

touch good
rm good
touch good

touch bad
rm bad
touch bad

N=16
for i in `seq 1 $RUNS`
do
    (
        run
    ) &

    # allow only to execute $N jobs in parallel
    if [[ $(jobs -r -p | wc -l) -gt $N ]]; then
        # wait only for first job
        wait -n
    fi
done
wait

echo GOOD $(wc -l good) VS BAD $(wc -l bad)
```
Commit\Thread Number   | 16               | 1
                                         | Good  | Bad    | Good | Bad
upstream/master              | 7         | 993    | 929    | 71
origin/share_xconnection | 10       | 990    | 901    | 99
Difference                        |  0.3%  | -0.3% | -2%    | 2%

ZeGentzy:
Fixed the table, sorry for that, github is wierd

PR #575, opened 2018-06-19
windows feature WS_EX_NOREDIRECTIONBITMAP

ekicyou:

Windows additionally has `WindowBuilderExt::with_no_redirection_bitmap`.
In order to use Direct Composition API, implemented the function to set WS_EX_NOREDIRECTIONBITMAP extended attribute.

ref: [MSDN Windows with C++ - High-Performance Window Layering Using the Windows Composition Engine](https://msdn.microsoft.com/magazine/dn745861.aspx)


src/os/windows.rs
```diff
@@ -55,6 +55,9 @@ pub trait WindowBuilderExt {
 
     /// This sets `ICON_BIG`. A good ceiling here is 256x256.
     fn with_taskbar_icon(self, taskbar_icon: Option<Icon>) -> WindowBuilder;
+
+    /// This sets `WS_EX_NOREDIRECTIONBITMAP`.
+    fn with_no_redirection_bitmap(self) -> WindowBuilder;
```
francesca64: It's conventional for builder methods to take a `bool`, for more flexibility.

francesca64:
Well, I don't really know how to test this, since it just makes my OpenGL app transparent.

My only question is if this is mutually exclusive with `with_transparency`. Having both made the client area black.

ekicyou:
> My only question is if this is mutually exclusive with with_transparency. Having both made the client area black.

Excuse my translation.

It becomes exclusive designation.

Both can be used as a means to realize a transparent window, but the realization method is different.

WS_EX_NOREDIRECTIONBITMAP specifies not to display the GDI bitmap.
The GDI bitmap is the old rendering API, not GPU optimized.
By using only composition surfaces directly managed by DWM for screen display, you can intercept maximum GPU optimization of Windows.

As a composition surface, you can specify a swap chain.

I am sorry, I can not prepare a small sample now.
The trial at hand is in trial with the [modification](https://github.com/ekicyou/webrender) of [[servo/webrender]](https://github.com/servo/webrender).

ekicyou:
Thank you. Check the source and fix it.

francesca64:
Alright, looks great.

PR #577, opened 2018-06-22
Wayland: Fix window creation dimensions

francesca64:
In retrospect, I was actually the one who put in this dummy code (https://github.com/francesca64/winit/commit/2dbe28e30079b3532cbca9aedea4f08f0a18d1ab#diff-7d4b937dfd8134b6a46e1cfc22ca80faR27). This explains a lot, since 64 is clearly a cheeky reference to @francesca64.

Sorry @vberger!


francesca64:
CI failure is because of rust-lang/rust#51699

PR #588, opened 2018-07-03
Disable maximize button on non-resizable windows

matanui159:

WIN32: Implemented and tested. I created a constant to prevent repetitive code, if this is not preferable I can change it.

MacOS: According to https://github.com/tomaka/winit/pull/541#issuecomment-394798514, this is already automatically done on MacOS (can someone check this?)

X11: This is a bit difficult. GLFW disables resizing it by setting the maximum and minimum size to the base size. Maybe this also disables maximizing?

Wayland: According to https://github.com/tomaka/winit/pull/127#issuecomment-296431711, this can be done by not drawing the maximize button. Not sure if this is already done or not, haven't look too far into the source code.


matanui159:
Looking further into the code it seems that X11 already does the same thing as GLFW, does this disable maximization on Linux? I don't have a Linux distro setup right now so if someone else can check that would be great (I've created an example called `non_resizable`). Wayland also seems to managed by client-toolkit.

francesca64:
Thanks for making a PR!

> this is already automatically done on MacOS (can someone check this?)

Aw, don't you trust me?

> Looking further into the code it seems that X11 already does the same thing as GLFW, does this disable maximization on Linux?

It unfortunately varies by WM. It works on KWin, which has a big market share, but not on Xfwm (though I special-cased that entire method into a no-op on Xfwm, so it doesn't matter). I haven't checked Mutter, but the important bits are 1) it's supposed to disable maximization, and 2) it's not consistently done. So, don't worry about it.

examples/non_resizable.rs
```diff
@@ -0,0 +1,22 @@
+extern crate winit;
```
francesca64: This entire example is redundant, since we already have this: https://github.com/tomaka/winit/blob/master/examples/resizable.rs

src/platform/windows/window.rs
```diff
@@ -951,7 +953,7 @@ unsafe fn init(
         };
 
         if !attributes.resizable {
-            style &= !winuser::WS_SIZEBOX;
+            style &= !WS_RESIZABLE as u32;
```
francesca64: I believe the type is `DWORD`, not `u32`.

CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 # Unreleased
+- Disabled maximize button on non-resizable windows (currently Windows only).
```
francesca64: There should be a newline between the header and the entry list.

src/platform/windows/window.rs
```diff
@@ -33,6 +33,8 @@ use platform::platform::monitor::get_available_monitors;
 use platform::platform::raw_input::register_all_mice_and_keyboards_for_raw_input;
 use platform::platform::util;
 
+const WS_RESIZABLE: LONG = (winuser::WS_SIZEBOX | winuser::WS_MAXIMIZEBOX) as LONG;
```
francesca64: Isn't it more semantically appropriate to store this as a `DWORD`? This ought to be the canonical form for this piece of data.

src/platform/windows/window.rs
matanui159: K thx, my IDE just told me it was u32 which is probably the same thing.

src/platform/windows/window.rs
matanui159: Sometimes it was used as a LONG and sometimes it was used as a DWORD so I didn't really know which to use, will switch it out

matanui159:
What documentation changes would you like me to make?

matanui159:
Also for linux, do you think it would be possible to check when the window is resized and then reset its size? Its a bit forceful but it might work for WM's that don't want to cooperate.

francesca64:
> What documentation changes would you like me to make?

Just that you stated maximization is only disabled on Windows.

> Also for x11, do you think it would be possible to check when the window is resized and then reset its size? Its a bit forceful but it might work for WM's that don't want to cooperate.

That would work, but it's simply too aggressive of a solution. I could imagine it causing really weird glitches when using tiling WMs, forming a continuous loop of the WM trying to resize the window to place it somewhere while the window keeps changing back. One of the concessions we have to make when developing for X11 is that there's no strong set of federated guidelines for how WMs should behave, and the standards that do exist are more *may* than *must*. This is an unfixable problem, and trying to fix unfixable problems just creates more problems.

It's of my opinion that `with_resizable`/`set_resizable` should basically never be used in serious applications anyway. Even if an application disables resizing, it's still a malformed winit program if it doesn't handle `Resized`, so the only conceivable point is indicating to the user that you prefer they keep the window at the size you chose.

francesca64:
Alright, I finally got around to testing this, and everything's good to go! (though maintainers are naturally paranoid, so I'll wait for the CI to pass before merging...)

About the CHANGELOG, I figured I'd come across as really annoying if I insisted you write it a particular way, so I hope you don't mind me having written it directly.

matanui159:
Nah its fine, I'm not a big fan of documentation anyway

PR #595, opened 2018-07-09
set the UIViewController's view to the one that was just created

mtak-:
This, plus a bunch of other unrelated changes, let me get a triangle on screen on iOS with vulkano. I suspect this will fix a lot issues dependencies like glium have on iOS.

Let me know if you need more information from me.

### DIFF
--- a/src/platform/ios/mod.rs
+++ b/src/platform/ios/mod.rs
@@ -489,7 +489,7 @@ fn create_delegate_class() {
             let view: id = msg_send![view_class, alloc];
             let view: id = msg_send![view, initForGl:&bounds];
 
-            msg_send![view_controller, setView: view];
+            let _: () = msg_send![view_controller, setView:view];
 
             let _: () = msg_send![window, setRootViewController:view_controller];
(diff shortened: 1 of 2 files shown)


src/platform/ios/mod.rs
```diff
@@ -489,6 +489,8 @@ fn create_delegate_class() {
             let view: id = msg_send![view_class, alloc];
             let view: id = msg_send![view, initForGl:&bounds];
 
+            msg_send![view_controller, setView: view];
```
francesca64: You need to prefix this with `let _: () = ` like in below, since the stabilization of `!` (the never type) will change the type inference from `()` to `!`, resulting in sigsegv/sigill.

I also prefer `setView:view`, since AFAIK it's more consistent with conventional Obj-C style, though correct me if I'm wrong on that.

mtak-:
FWIW, the vulkano code I'm working on eventually stackoverflows. But not before rendering several frames

francesca64:
Sorry for the delay.

screen shot 2018-07-13 at 13 42 41

This is great! Not only can I actually render stuff now, but touch events work again too!

PR #598, opened 2018-07-10
Clarify DPI docs to highlight WindowEvent::HiDpiFactorChanged

### LINKED ISSUE #594: get_hidpi_factor returns incorrect value for Gnome on Wayland
In wayland `get_hidpi_factor` returns 1.0 instead of the correct value equal to 2.0. I am using Gnome 3.28.2 and winit v0.16.1.

vberger:
This is an attempt to clarify the relation between `WindowEvent::HiDpiFactorChanged` and `Window::get_hidpi_factor`.

Fixes #594

### DIFF
--- a/src/dpi.rs
+++ b/src/dpi.rs
@@ -30,7 +30,7 @@
 //!
 //! There are two ways to get the DPI factor:
 //! - You can track the `WindowEvent::HiDpiFactorChanged` event of your windows. This event is sent any
-//!   time the HiDPI factor changes, be it because the window moved to an other monitor, or the
+//!   time the DPI factor changes, be it because the window moved to an other monitor, or the
 //!   user changed the configuration of their screen.
 //! - You can also retrieve the DPI factor of a monitor by calling
 //!   [`MonitorId::get_hidpi_factor`](../struct.MonitorId.html#method.get_hidpi_factor), or the
@@ -38,7 +38,7 @@
 //!   [`Window::get_hidpi_factor`](../struct.Window.html#method.get_hidpi_factor), which is roughly equivalent
 //!   to `window.get_current_monitor().get_hidpi_factor()`.
 //!
-//! Depending on the platorms, the actual DPI factor applied on a window may only be known after
+//! Depending on the platform, the window's actual DPI factor may only be known after
 //! the event loop has started and your window has been drawn once. To properly handle these cases,
 //! the most robust way is to monitor the `WindowEvent::HiDpiFactorChanged` event and dynamically
 //! adapt your drawing logic to follow the DPI factor.
--- a/src/dpi.rs
+++ b/src/dpi.rs
@@ -19,9 +19,9 @@
 //! small.
 //!
 //! That's a description of what happens when the button is 100x100 *physical* pixels. Instead, let's try using 100x100
-//! *logical* pixels. To map logical pixels to physical pixels, we simply multiply by the DPI (dots per inch) factor. 
-//! On a "typical" desktop display, the DPI factor will be 1.0, so 100x100 logical pixels equates to 100x100 physical 
-//! pixels. However, a 1440p display may have a DPI factor of 1.25, so the button is rendered as 125x125 physical pixels. 
+//! *logical* pixels. To map logical pixels to physical pixels, we simply multiply by the DPI (dots per inch) factor.
+//! On a "typical" desktop display, the DPI factor will be 1.0, so 100x100 logical pixels equates to 100x100 physical
+//! pixels. However, a 1440p display may have a DPI factor of 1.25, so the button is rendered as 125x125 physical pixels.
 //! Ideally, the button now has approximately the same perceived size across varying displays.
 //!
 //! Failure to account for the DPI factor can create a badly degraded user experience. Most notably, it can make users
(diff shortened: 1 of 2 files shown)


src/dpi.rs
```diff
@@ -28,10 +28,20 @@
 //! feel like they have bad eyesight, which will potentially cause them to think about growing elderly, resulting in
 //! them entering an existential panic. Once users enter that state, they will no longer be focused on your application.
 //!
-//! There are two ways to get the DPI factor: either by calling
-//! [`MonitorId::get_hidpi_factor`](../struct.MonitorId.html#method.get_hidpi_factor), or
-//! [`Window::get_hidpi_factor`](../struct.Window.html#method.get_hidpi_factor). You'll almost always use the latter,
-//! which is basically equivalent to `window.get_current_monitor().get_hidpi_factor()` anyway.
+//! There are two ways to get the DPI factor:
+//! - track the `WindowEvent::HiDpiFactorChanged` event of your windows. This event is sent any
```
francesca64: I think it looks better when list items like this start with a capital letter, though that's probably subjective.

vberger:
Welp, my English is still clearly not as good as I'd like it to be. Is that better?

src/window.rs
```diff
@@ -304,6 +304,10 @@ impl Window {
     ///
     /// See the [`dpi`](dpi/index.html) module for more information.
     ///
+    /// Note that this value can change depending on user action (for example if the window is
+    /// moved to another screen), as such tracking the `WindowEvent::HiDpiFactorChanged` events is
```
francesca64: So, I was wondering if "as such" here ought to be preceded by a semicolon rather than a comma, but after doing some digging, it turns out that "as such" is quite the can of worms... https://www.chronicle.com/blogs/linguafranca/2015/04/24/academic-writing-as-such/

Since I use "as such" rather often, it will take a lot of soul searching before I can decide how I feel about it... ~~as such~~, I'm not going to make a fuss about this. I mostly just found it interesting.

(Though preceding with a semicolon + putting a comma after "as such" is probably still a good idea?)

src/dpi.rs
```diff
@@ -28,10 +28,20 @@
...
+//! - You can track the `WindowEvent::HiDpiFactorChanged` event of your windows. This event is sent any
+//!   time the HiDPI factor changes, be it because the window moved to an other monitor, or the
+//!   user changed the configuration of their screen.
+//! - You can also retrieve the DPI factor of a monitor by calling
+//!   [`MonitorId::get_hidpi_factor`](../struct.MonitorId.html#method.get_hidpi_factor), or the
+//!   current DPI factor applied to a window by calling
+//!   [`Window::get_hidpi_factor`](../struct.Window.html#method.get_hidpi_factor), which is roughly equivalent
+//!   to `window.get_current_monitor().get_hidpi_factor()`.
+//!
+//! Depending on the platorms, the actual DPI factor applied on a window may only be known after
```
francesca64: I think "platform" instead of "platforms" would sound more natural here. I also think "applied to" is probably better... man, I'm starting to feel like I seem obnoxious for making grammar nitpicks.

src/dpi.rs
```diff
@@ -28,10 +28,20 @@
 //! feel like they have bad eyesight, which will potentially cause them to think about growing elderly, resulting in
 //! them entering an existential panic. Once users enter that state, they will no longer be focused on your application.
 //!
-//! There are two ways to get the DPI factor: either by calling
-//! [`MonitorId::get_hidpi_factor`](../struct.MonitorId.html#method.get_hidpi_factor), or
-//! [`Window::get_hidpi_factor`](../struct.Window.html#method.get_hidpi_factor). You'll almost always use the latter,
-//! which is basically equivalent to `window.get_current_monitor().get_hidpi_factor()` anyway.
+//! There are two ways to get the DPI factor:
+//! - You can track the `WindowEvent::HiDpiFactorChanged` event of your windows. This event is sent any
+//!   time the HiDPI factor changes, be it because the window moved to an other monitor, or the
```
francesca64: This is the only time in the docs that "HiDPI factor" is said instead of "DPI factor", so it could cause people to assume there's a distinction... though, arguably I should've been writing "HiDPI factor" the whole time, since that corresponds to the actual API naming.

francesca64:
Okay, sorry for the delay! I guess I'm rather protective of this documentation, so hopefully I'm not coming across as overly critical.

vberger:
No worries, I've changed the text following your comments. :)

src/dpi.rs
```diff
@@ -28,10 +28,20 @@
 //! feel like they have bad eyesight, which will potentially cause them to think about growing elderly, resulting in
 //! them entering an existential panic. Once users enter that state, they will no longer be focused on your application.
 //!
-//! There are two ways to get the DPI factor: either by calling
-//! [`MonitorId::get_hidpi_factor`](../struct.MonitorId.html#method.get_hidpi_factor), or
-//! [`Window::get_hidpi_factor`](../struct.Window.html#method.get_hidpi_factor). You'll almost always use the latter,
-//! which is basically equivalent to `window.get_current_monitor().get_hidpi_factor()` anyway.
+//! There are two ways to get the DPI factor:
+//! - You can track the `WindowEvent::HiDpiFactorChanged` event of your windows. This event is sent any
+//!   time the DPI factor changes, be it because the window moved to an other monitor, or the
```
francesca64: One last "an other" -> "another"... and the "or the" at the end would sound more balanced as "or because the".

Issue #599, opened 2018-07-12
Possible unsoundness on windows


andrewhickman:
This is just something that stuck out as odd to me while browsing the source, so I could be completely wrong here, but the windows backend seems to have some thread-safety issues. `Window` is marked with `unsafe impl Send + Sync` but contains some `Cell` and `RefCell`. Looking at, for example, [`set_resizable`](https://github.com/tomaka/winit/blob/c1ef1acfc0a6d372519c3cf1a1c1e0696373fa61/src/platform/windows/window.rs#L285) (which is exposed in the public API) these seem to get used without additional synchronisation.

francesca64:
Hm, this looks problematic... worse still, I think I was the one who let this pass code review! Then because of the magic of technical debt, people who added more members over time used `Cell` too.

Do you want to fix this, or should I?

andrewhickman:
I'm happy to fix this :)

francesca64:
Fixed via #601

PR #601, opened 2018-07-13
Fix unsoundness on Windows

andrewhickman:
This PR fixes the unsound use of Cell and RefCell on windows (See #599). The first commit just replaces `Cell<bool>` with `AtomicBool` and moves `RefCell` into the `Mutex<WindowState>`. That should take care of the safety issues but I think there is still potential for the saved state to get out of sync with what has been executed. The second commit attempts to address this by moving all the window state variables into `WindowState` and moving all calls to `execute_in_thread` to within the lock of `WindowState`.

I think there are still some similar soundness issues on the macos backend ([e.g. here](https://github.com/tomaka/winit/blob/c1ef1acfc0a6d372519c3cf1a1c1e0696373fa61/src/platform/macos/window.rs#L529)) but I'm not currently able to test any fixes there.


francesca64:
Sorry for the delay! This looks good, though I still haven't tested it out yet.

For soundness stuff like this, I think it's good to have a CHANGELOG entry, so people using older versions are aware of the unsoundness.

> moving all calls to `execute_in_thread` to within the lock of `WindowState`

If this is something we want to require from now on, then it would make sense for the parameters of `execute_in_thread` to enforce this.

andrewhickman:
I've added a comment to `execute_in_thread`. There is one case where it is called outside of the lock (in Window::new) so I haven't fiddled with enforcing it further for now.

CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 # Unreleased
+- Fixed some unsoundness around thread-safety on Windows.
```
francesca64: There should be a newline between the header and the list, and this entry is a bit vague.

src/platform/windows/events_loop.rs
```diff
@@ -57,6 +57,7 @@ use platform::platform::dpi::{
     enable_non_client_dpi_scaling,
     get_hwnd_scale_factor,
 };
+use platform::platform::icon::WinIcon;
```
francesca64: Keeping with alphabetical order, this should go after `platform::platform::event` below.

src/platform/windows/window.rs
```diff
@@ -1,19 +1,19 @@
...
 
-use std::cell::{Cell, RefCell};
+use std::cell::Cell;
 use std::ffi::OsStr;
-use std::{io, mem, ptr};
 use std::os::windows::ffi::OsStrExt;
-use std::sync::{Arc, Mutex};
 use std::sync::mpsc::channel;
+use std::sync::{Arc, Mutex};
+use std::{io, mem, ptr};
```
francesca64: I prefer to have the top-level brackety imports sorted above more specific ones... this also applies to the `winapi::um` below.

src/platform/windows/window.rs
```diff
@@ -283,32 +272,23 @@ impl Window {
...
-            winuser::SetWindowLongW(
-                self.window.0,
-                winuser::GWL_STYLE,
-                style as _,
-            );
+            winuser::SetWindowLongW(self.window.0, winuser::GWL_STYLE, style as _);
         };
-        self.resizable.replace(resizable);
+    }
+        }
```
francesca64: There's some very confusing bracket indentation happening here.

src/platform/windows/window.rs
```diff
@@ -1029,6 +1003,13 @@ unsafe fn init(
             mouse_in_window: false,
             saved_window_info: None,
             dpi_factor,
+            fullscreen: attributes.fullscreen.clone(),
+            window_icon: window_icon,
```
francesca64: Just `window_icon,` would be cleaner.

francesca64:
I finally got around to testing this out, and things seem peachy. Just a few nitpicks, and this will make it into a release pretty soon.

francesca64:
Thanks!

Issue #597, opened 2018-07-10
libxkbcommon.so is not present OOTB on most distros, causing wayland keyboard handling to silently fall back to raw scancodes only.


whentze:
I realize this was my own fault, but after spending so many hours tracking down the root cause of this problem, I think it should at least be documented so others can avoid it.

- [This comment](https://github.com/tomaka/winit/blob/v0.13.0/src/platform/linux/wayland/keyboard.rs#L85) from winit's wayland keyboard handling code states that, on systems with a wayland compositor, libxkbcommon is most likely installed anyway. This is correct, however
- wayland-client-toolkit tries to [dlopen "libxkbcommon.so"](https://github.com/Smithay/client-toolkit/blob/master/src/keyboard/ffi.rs#L257) when initializing the keyboard, and
- On many distros, you will usually only have "libxkbcommon0" installed, so there will only be "libxkbcommon.so.0", not an unversioned .so file. To get "libxkbcommon.so", which would be a symlink to "libxkbcommon.so.0", you usually need the libxkbcommon-dev or -devel package, which is *not* usually installed.
This is the case on [Debian](https://packages.debian.org/da/jessie/amd64/libxkbcommon0/filelist), [Ubuntu](https://packages.ubuntu.com/bionic/amd64/libxkbcommon0/filelist), [Fedora](https://koji.fedoraproject.org/koji/rpminfo?rpmID=12673378#filelist), [CentOS](https://centos.pkgs.org/7/centos-x86_64/libxkbcommon-0.7.1-1.el7.x86_64.rpm.html) and [OpenBSD](https://cvsweb.openbsd.org/cgi-bin/cvsweb/~checkout~/ports/x11/xkbcommon/pkg/PLIST?rev=1.2&content-type=text/plain).
- On those plattforms, winit just silently falls back to `raw_keyboard_impl()`, which is not very useful.

There are several options for making this better:
-Change the client toolkit so it looks for "libxkbcommon.so.0" instead. This has the benefit of avoiding breakage in case an incompatible "libxkbcommon.so.1" is released.
-Explicitly document that the unversioned "libxkbcommon.so" is a dependency, and hint users towards installing the -dev package.
-Emit a warning (log?) message when falling back to raw input, also possibly hinting towards the -dev package.

vberger:
> -Change the client toolkit so it looks for "libxkbcommon.so.0" instead. This has the benefit of avoiding breakage in case an incompatible "libxkbcommon.so.1" is released.

[...]

PR #602, opened 2018-07-14
Update wayland-client and client-toolkit

trimental:

A few fixes, improvements and decoration changes

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,8 +54,8 @@ features = [
 ]
 
 [target.'cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "openbsd"))'.dependencies]
-wayland-client = { version = "0.20.6", features = [ "dlopen", "egl", "cursor"] }
-smithay-client-toolkit = "0.2.2"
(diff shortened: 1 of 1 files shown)


vberger:
These upgrates pull a few interesting things:

- soundness fixes in `wayland-client`
- fix for #597 and several improvements on the decorations look and interactions with the server, from `smithay-client-toolkit`

However, upgrading the point-releases dependencies like this is not generally necessary, as `cargo update` already pulls them as compatible versions. It would allow us to avoid having some users coming about a bug that they actually only need to run `cargo update` to get fixed, though.

trimental:
Ah I didn’t know that 👍, I was mostly trying to get the 2.4 client toolkit fix as without it I can’t use the wayland backend

francesca64:
I guess there'd be no harm in merging this?

trimental:
By the sounds of it if anything broke then it would already be broken now anyway because of the updating, I didn’t know that cargo updated minor versions automatically, I left this open because I thought it wouldn’t hurt to update it anyway.

francesca64:
Right. Merging PRs is fun anyway.

PR #603, opened 2018-07-16
Support NetBSD platform

iquiw:
  NetBSD only


francesca64:
Thanks!

PR #605, opened 2018-07-17
Use class macro

### LINKED ISSUE #604: Use new objc class macro
[objc added a `class` macro](https://github.com/SSheldon/rust-objc/issues/65) which caches access to the class. So basically we can simply replace all `Class::get("x").unwrap()` with `class!(x)`. The `unwrap` isn't needed anymore because the macro will panic if it fails. Note also that this macro should eventually (through procedural macros) be able to link directly to the class.

I'd be happy to send a PR which performs this replacement, but cases like https://github.com/tomaka/winit/blob/master/src/platform/ios/mod.rs#L173 need to be clarified (do these ever actually fail?).

grovesNL:
- [X] Tested on all platforms changed (tested macOS, not able to test on iOS)

Fixes #604

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,10 +26,10 @@ image = { version = "0.19", optional = true }
 version = "0.2"
 
 [target.'cfg(target_os = "ios")'.dependencies]
-objc = "0.2"
+objc = "0.2.3"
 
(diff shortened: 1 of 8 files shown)


francesca64:
Thanks! Looks great.

PR #606, opened 2018-07-19
[x11-backend] Retrieve DPI from Xft.dpi XResource

semtexzv:
DPI Scaling on X11 is in rough shape. But I think calculating actual display DPI ourselves is not a way to go. The "Xft.dpi"  XResource is at least somewhat standard on several linuxes ( [Arch](https://wiki.archlinux.org/index.php/HiDPI#X_Resources) ). And is used in both [GLFW](https://github.com/glfw/glfw/blob/7ef34eb06de54dd9186d3d21a401b2ef819b59e7/src/x11_init.c#L802) and [QT](https://code.qt.io/cgit/qt/qtbase.git/tree/src/plugins/platforms/xcb/qxcbscreen.cpp?id=b393641888f18e3accf094c91808e2189c2138cc#n860) .

 This commit changes the behavior of DPI scaling on X11. The primary value of DPI scaling factor is retrieved from Xft.dpi XResource. The previous behavior is used as a fallback option.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - On iOS, the view is now set correctly. This makes it possible to render things (instead of being stuck on a black screen), and touch events work again.
 - Added NetBSD support.
+- X11 - Dpi scaling factor behavior changed. First, winit tries to read it from "Xft.dpi" XResource, and uses DPI calculation from xrandr dimensions as fallback behavior.
 
 # Version 0.16.2 (2018-07-07)
 
--- a/src/platform/linux/x11/util/randr.rs
+++ b/src/platform/linux/x11/util/randr.rs
@@ -79,6 +79,24 @@ impl From<*mut ffi::XRRCrtcInfo> for MonitorRepr {
 }
 
 impl XConnection {
+    // Retrieve DPI from Xft.dpi property
+    pub unsafe fn get_xft_dpi(&self) -> Option<f64> {
+        (self.xlib.XrmInitialize)();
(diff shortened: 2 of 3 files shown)


francesca64:
How does this work with respect to multiple monitors / mixed DPI?

semtexzv:
It doesn't. It is a global override of DPI setting. Both GLFW and QT use it as such.

francesca64:
Alright, and this won't regress mixed DPI for people who don't have this var set?

pedrocr:
Just hit this issue running my app in Ubuntu 18.04. I just checked and Xft.dpi is also set by GNOME's display settings. Can we get this merged then?

chrisduerr:
Note that QT does support mixed DPI with multiple monitors similar to what winit is doing right now with a command line flag. I'm not sure if the Xft.dpi scaling is considered, but it would be possible to apply the Xft.dpi scaling on top of what winit has right now as a way to increase/decrease the scale globally, while the relation between monitors is calculated automatically to always be correct.

mitchmindtree:
I can confirm that this PR does fix the expected behaviour on my hidpi laptop, however I can't comment on the mixed-dpi behaviour as I don't have another monitor available to test on.

francesca64:
I'm definitely in favor of moving forward on this (and exploring that Qt multi-monitor stuff, though possibly not until a later PR) but it looks like @semtexzv isn't here anymore. Would anyone like to resurrect this PR?

Osspial:
Closing, since the rebased version of this got merged.

PR #609, opened 2018-07-24
iOS Abstract Out the UIView type from winit

mtak-:

There are a lot of changes to iOS in this pull request.
- The `UIApplication` `setjmp` hacks now happen in `Window::new`. This allows `PlatformSpecificWindowBuilderAttributes` to customize the specific `UIView` class.
- The biggest breaking change here is that `winit` no longer sets up a `GLKView`. Putting responsibility on other libraries like `glutin` and `vulkano-win` to specify the exact `UIView` subclass.
- `EventsLoop` now explicitly checks that it's created on the main thread. AFAICT this was required before, but it was just UB to create it anywhere else.
- Only one `Window` can ever be created.
- `EventsLoop` cannot be polled until a `Window` has been created - in order for the `CFRunLoopRunInMode` calls to work.

I have been testing it with `glutin`. https://github.com/mtak-/glutin/commit/1a3066f8582ee2881890c7e7297419c8badb2fb3

Let me know what needs to be changed/etc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- **Breaking:** On iOS, `UIView` is now the default root view. `WindowBuilderExt::with_root_view_class` can be used to set the root view objective-c class to `GLKView` (OpenGLES) or `MTKView` (Metal/MoltenVK).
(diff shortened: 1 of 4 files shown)


francesca64:
This looks fantastic, thanks!

PR #610, opened 2018-07-25
Fix key state on MacOS

paulrouget:
To reproduce the issue:
- on Mac, `cargo run --example window`
- press the `Cmd` key
- See that we get two consecutive `state:Release` `KeyboardInput` events

We should get `Pressed` then `Release`.


francesca64:
Good catch, thanks!

PR #612, opened 2018-07-26
Experiment: Update to Rust 2018 edition.

icefoxen:
...on Linux, at least.  Rather a PITA but it seems to work fine.  I don't really expect this to be merged, at least not for a while, but it may be interesting for us to have in our back pocket.


Osspial:
This has gotten out-of-date enough that I can't see it being merged in its current form, so I'm closing this and opening an issue at #912.

PR #613, opened 2018-07-26
iOS: Fix the `longjmp`/`setjmp` ffi

mtak-:

- The length of the `JmpBuf` was incorrectly set to `27` `c_int`'s. In reality the length is architecture dependent and larger than `27` on most platforms.
- The longjmp function now returns `!`
- Tweaked a `panic!` message.

Before this change, sometimes there would be random crashes every few launches with unhelpful stack traces:

```
malloc: *** error for object 0x1d00ceb60: Invalid pointer dequeued from free list
*** set a breakpoint in malloc_error_break to debug
```

After this change, I am unable to reproduce the crash.

### DIFF
--- a/src/platform/ios/ffi.rs
+++ b/src/platform/ios/ffi.rs
@@ -64,10 +64,20 @@ extern {
 
 extern {
     pub fn setjmp(env: *mut c_void) -> c_int;
-    pub fn longjmp(env: *mut c_void, val: c_int);
+    pub fn longjmp(env: *mut c_void, val: c_int) -> !;
 }
 
-pub type JmpBuf = [c_int; 27];
+// values taken from "setjmp.h" header in xcode iPhoneOS/iPhoneSimulator SDK
+#[cfg(any(target_arch = "x86_64"))]
+pub const JBLEN: usize = ((9 * 2) + 3 + 16);
+#[cfg(any(target_arch = "x86"))]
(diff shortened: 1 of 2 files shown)


src/platform/ios/ffi.rs
```diff
@@ -64,10 +64,20 @@ extern {
...
-    pub fn longjmp(env: *mut c_void, val: c_int);
+    pub fn longjmp(env: *mut c_void, val: c_int) -> !;
 }
 
-pub type JmpBuf = [c_int; 27];
+// values taken from "setjmp.h" header in xcode iPhoneOS/iPhoneSimulator SDK
+#[cfg(any(target_arch = "x86_64"))]
+pub const JBLEN: usize = ((9 * 2) + 3 + 16);
+#[cfg(any(target_arch = "x86"))]
+pub const JBLEN: usize = (18);
```
francesca64: What's the significance of the parentheses?

francesca64:
I'm really glad the iOS backend is getting some love! A CHANGELOG entry would be nice, but everything looks fine.

src/platform/ios/ffi.rs
mtak-: I copied it directly from the "setjmp.h" header where they `#define _JBLEN (value)`. The extra parentheses don't make sense in rust :). I'll remove them.

Issue #607, opened 2018-07-20
Win 7: assertion failed: validate_hidpi_factor(dpi_factor)


sinesc:
I am getting frequent validate_hidpi_factor assertion failure panics on Windows 7 in dpi.rs in line 273:9 and 153:9 (random which) using winit 0.16.2. This does not seem to occur on Windows 10 or Linux.

```
thread 'main' panicked at 'assertion failed: validate_hidpi_factor(dpi_factor)', C:\Users\sinesc\.cargo\registry\src\github.com-1ecc6299db9ec823\winit-0.16.2\src\dpi.rs:273:9
stack backtrace:
   0:        0x14018cad3 - std::sys::windows::backtrace::unwind_backtrace
                               at C:\projects\rust\src\libstd\sys\windows\backtrace\mod.rs:95
   1:        0x14019676e - std::sys_common::backtrace::print
                               at C:\projects\rust\src\libstd\sys_common\backtrace.rs:59
   2:        0x140186758 - std::panicking::default_hook::{{closure}}
                               at C:\projects\rust\src\libstd\panicking.rs:211
   3:        0x140186411 - std::panicking::default_hook
                               at C:\projects\rust\src\libstd\panicking.rs:227
   4:        0x140186f5c - std::panicking::rust_panic_with_hook
                               at C:\projects\rust\src\libstd\panicking.rs:475
   5:        0x13fd656c2 - std::panicking::begin_panic<str*>
                               at C:\projects\rust\src\libstd\panicking.rs:409
   6:        0x13fd362a3 - winit::dpi::PhysicalSize::to_logical
                               at C:\Users\sinesc\.cargo\registry\src\github.com-1ecc6299db9ec823\winit-0.16.2\src\dpi.rs:273
   7:        0x13fd35fca - winit::dpi::LogicalSize::from_physical<(u32, u32)>
                               at C:\Users\sinesc\.cargo\registry\src\github.com-1ecc6299db9ec823\winit-0.16.2\src\dpi.rs:208
   8:        0x13fbff8c7 - winit::platform::platform::window::{{impl}}::get_inner_size::{{closure}}
                               at C:\Users\sinesc\.cargo\registry\src\github.com-1ecc6299db9ec823\winit-0.16.2\src\platform\windows\window.rs:194
   9:        0x13fbf1570 - core::option::Option<(u32, u32)>::map<(u32, u32),winit::dpi::LogicalSize,closure>
                               at C:\projects\rust\src\libcore\option.rs:414
[...]

PR #608, opened 2018-07-21
Update cocoa and core-graphics

paulrouget:
- [x] Tested on all platforms changed
- [x] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users
- [ ] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior
- [ ] Created an example program if it would help users understand this functionality

PR #618, opened 2018-07-30
fix assertion failed: validate_hidpi_factor(dpi_factor) (#607)

sinesc:
use cached window dpi to fix panic on win 7

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - On iOS, the `UIApplication` is not started until `Window::new` is called.
 - Fixed thread unsafety with cursor hiding on macOS.
 - On iOS, fixed the size of the `JmpBuf` type used for `setjmp`/`longjmp` calls. Previously this was a buffer overflow on most architectures.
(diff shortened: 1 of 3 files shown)


sinesc:
The build failure seems to be unrelated. Not sure what to do here.

paulrouget:
@francesca64 can we land this and #608?

francesca64:
@paulroget yeah, I just didn't get a chance to review this yesterday.

I still need to test this, but the code looks good.

@sinesc don't worry about the build failure, but this needs a CHANGELOG entry.

francesca64:
Okay, I tested this on Windows 10 (checking both the modern and legacy code paths), and everything works fine. I'll merge after a CHANGELOG entry is added.

francesca64:
Thanks! All winit PRs get auto-squashed, so it's not a problem.

francesca64:
This appears to have broken fullscreen and maximization on Windows. This is presumably because they trigger a resize, which causes us to deadlock on the window state.

francesca64:
Alright, I've already made a fix: #622

PR #621, opened 2018-08-05
Maintenance/620/fix x11 release mode compilation

azriel91:
- [x] Tested on all platforms changed
- [x] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users
- n/a ~~Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior~~
- n/a ~~Created an example program if it would help users understand this functionality~~

Fixes #620 (tested it fails compilation before the fix, and works after)

PR #622, opened 2018-08-05
Windows: Fix fullscreen deadlock + release 0.17.1

francesca64:
The change to `get_hidpi_factor` in #618 produces a deadlock when entering fullscreen mode, and this fixes that.

Since both this and #621 are rather onerous problems, I'm also releasing 0.17.1 now.


felixrabe:
@francesca64 I only see "Release winit 0.17.1" on your branch `win-fullscreen-deadlock` but not on `master`.

francesca64:
All PRs get squashed, and you can see the relevant changes here: b2b740fed709dfc279652015959631c245364d2d

felixrabe:
Ah right thx, forgot about the squashing.

PR #627, opened 2018-08-13
[NOMERGE] Adds a way to create windows and events pools from a subset of their raw parts.

ZeGentzy:
Adds a way to create windows and events pools from a subset of their raw parts.

A new API which allows us to make windows and events pools from a subset of their raw parts. Should be useful when interfacing with C libraries. 

~Currently the raw parts we require just so happen to match the ones we receive from gfx-rs/portability in its `Vk*SurfaceCreateInfo` structs, as my use case for this api will be constructing `Window`s and `EventsPool`s from that data (and that will just make my life easier :D).~ We now also require the dims of the surface for the wayland backend.

I currently only plan to implement this functionality on linux and windows. Currently untested.

Implemented:
 - [x] Linux - X11
 - [x] Linux - Wayland
 - [x] Windows

Tested:
 - [x] Linux - X11
 - [ ] Linux - Wayland
 - [ ] Windows

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,7 +54,7 @@ features = [
 ]
 
 [target.'cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "openbsd", target_os = "netbsd"))'.dependencies]
-wayland-client = { version = "0.20.10", features = [ "dlopen", "egl", "cursor"] }
+wayland-client = { version = "0.20.12", features = [ "dlopen", "egl", "cursor"] }
 smithay-client-toolkit = "0.3.0"
 x11-dl = "2.18.3"
 parking_lot = "0.6"
--- a/examples/raw_window.rs
+++ b/examples/raw_window.rs
@@ -1,7 +1,5 @@
 extern crate winit;
 
-use std::sync::Mutex;
-
 fn main() {
     let mut evlp = winit::EventsLoop::new();
     let win = winit::WindowBuilder::new()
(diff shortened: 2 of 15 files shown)


vberger:
From a quick glance on your wayland code, it appears that you made the code behave such that a window made from raw parts is known by winit to be external, and many (most?) actions on it are not possible.

Could you detail the example you have in mind regarding why one would create a window from raw parts, and what they would do this it?

This is especially relevant regarding resizing. It appears from your code that you assume winit won't handle the resizing of the surface, though on wayland the dimensions of a surface is defined by the dimensions of the buffer you attach to it, so it's not really possible to have some other part handle the resizing without some integration.

ZeGentzy:
Hi @vberger, thanks for your interest.

Yes. Most functionality would be unavailable. The goal in mind is to make a openGL context on the raw parts of a window which gfx-hal receives from vulkan applications. An example of this is [here](https://github.com/gfx-rs/gfx/pull/2347/files#diff-d6f63be366bd926bf6799561e7d273f2R648)  (nonfunctional) and [here](https://github.com/tomaka/glutin/pull/1058/files#diff-f8d3bd20ee4a9d20ac560ca3c4518ad9R1) (functional).

Alternatively, I could've just added another set of functions to glutin which took the raw parts instead of a winit window, however, I figured a `new_from_raw_parts` function might be more useful.

I sadly haven't looked into how resizing would work on Wayland, yet.

ZeGentzy:
I've came to the conclusion that it works on X11, which is good enough for my intended use I guess. I don't intend to have this get merged in, as it doesn't work on Wayland, never mind other platforms.

Osspial:
Since there don't seem to be any intentions to get this into a mergeable state, I'm going to close this.

PR #371, opened 2017-12-22
wayland: Implement key repetition

mcoffin:
**NOTE**: This still needs cleanup. I'm just getting the base implementation out there to get comments before I spend a bunch of time cleaning it up.

This implements key repitition for wayland. I could use some comments on the high-level approach, as there are plenty of cleanup tasks that came from code-churn while debugging.

I wound up having to pass an `EventsLoopProxy` handle through the `wl_seat` and `wl_keyboard` implementations so that the underlying timer event thread would be able to wake up the `EventsLoop` that's reading from the sink.

Comments appreciated. I'm using this locally with jwilm/alacritty (with a few patches to update glutin to use this version of winit), and it seems to work alright. The threading overhead is a bit of a bummer, but it seems to be fairly minimal.

# TODO List

- [x] Audit + fix error handling with internal queues + async wakeups of `EventsLoop`.
- [x] Cleanup (read: unwrap) the `HashMap` usage for storing state in the timer thread, since we only track one key repeat at a time (I didn't know this initially until I actually played around with it in Xorg)
- [ ] Break up the monolith of nesting that occurs inside `KeyboardIData::new`
- [x] Think about the currently hard-coded channel capacities to make sure they're sensible.
- [ ] Dynamically configure the `tokio_timer::Timer` implementation based on the current `RepeatInfo` config. (This will just allow for more efficiency in the timer thread, and potentially more granular repeat intervals. 10ms is a decent preset, but dynamic-configuration is probably more correct).

Issue #310, opened 2017-10-07
 A way to disable key repeats


ryanisaacg:
At least on Windows, key events seem to respond with a delay (especially when a key is being mashed) due to key repeat handling. Is there any way to disable this?

(Originally [I asked in glutin but was referred here](https://github.com/tomaka/glutin/issues/943))

vberger:
For the record, the usecase of disabling key repeats is also of interest for [smithay](https//github.com/Smithay/smithay), otherwise we would need to filter-out repeated events...

ryanisaacg:
@vberger I think you're missing the ":" in the "https://" of your link

vberger:
Indeed, I've corrected it. 😅

francesca64:
We actually can't implement this on X11 without resorting to heuristics, though AFAIK it's straightforward to implement on Windows.

francesca64:
It's also easy to implement on macOS. Since there's a precedent for how to handle this on X11, we might as well go for it (though I don't consider it a priority at present). 

What's the ideal here? Adding an `is_repeat` field to `KeyboardInput`? Or it being configurable to completely strip out repeats? Both?

fschutt:
On windows, key repetition is currently way too fast, I'm not sure what's up with that, if it's a bug or not - is there any way to disable key repetition in winit? It's just because if I type "Hello World", I get "HHeeello WWWorrrrld" because the delay for key repetition is extremely low for some reason.

francesca64:
@fschutt I'm not able to reproduce that on Windows 10, though making it possible to filter key repeats would be easy on all platforms other than X11. I'm curious about why that's happening at all, though.

zpgaal:
I'd suggest (without knowing the internals of winit) to add a new ElementState: Repeat.
It can be discarded in the event loop if required.

Osspial:
Closing in favor of #753 and #812

Kittenlover229:
It seems like necroposting, but I've spend a decent amount researching relevant issues with no avail. Perhaps a hacky solution could be outlined *at least as a temporary fix*? It seems like the proposed bodge solution from GLFW doesn't exactly work, since winit doesn't provide the event timestamps.

kchibisov:
`KeyEvent` has `is_repeat` field, so simply ignore repeats.

Kittenlover229:
[...]

PR #628, opened 2018-08-13
Add key repetition for the wayland backend

trimental:

This PR adds the feature of key repetition to the wayland backend. This PR is requested prematurely so that it may be reviewed before the release of smithay-client-toolkit 0.3.0, currently it points to the git version as a dependency.

I have tested the changes and added a small update to the changelog (might need more explanation and better wording?), however I cannot find any documentation of key repetition (I assume this is just standard behavior).

Please look to https://github.com/Smithay/client-toolkit/pull/16 for further information about the new `map_keyboard_with_repeat()` function

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,7 +55,7 @@ features = [
 
 [target.'cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "openbsd", target_os = "netbsd"))'.dependencies]
 wayland-client = { version = "0.20.10", features = [ "dlopen", "egl", "cursor"] }
-smithay-client-toolkit = "0.2.6"
+smithay-client-toolkit = { git = "https://github.com/Smithay/client-toolkit" }
 x11-dl = "2.18.3"
 parking_lot = "0.6"
 percent-encoding = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,7 +55,7 @@ features = [
 
 [target.'cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "openbsd", target_os = "netbsd"))'.dependencies]
 wayland-client = { version = "0.20.10", features = [ "dlopen", "egl", "cursor"] }
-smithay-client-toolkit = { git = "https://github.com/Smithay/client-toolkit" }
+smithay-client-toolkit = "0.3.0"
 x11-dl = "2.18.3"
 parking_lot = "0.6"
 percent-encoding = "1.0"
--- a/src/platform/linux/wayland/keyboard.rs
+++ b/src/platform/linux/wayland/keyboard.rs
@@ -1,77 +1,110 @@
 use std::sync::{Arc, Mutex};
 
-use {ElementState, KeyboardInput, ModifiersState, VirtualKeyCode, WindowEvent};
-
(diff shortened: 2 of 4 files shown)


vberger:
This is a simpler implementation of key repetition, based on a thread.

It has been developed in #371 that no background thread should be needed. I do have some plans to introduce some real, proper event loop capability to my wayland libs (in a way making it possible to have good support for secondary event sources, including timers). But I don't know when I'll have the bandwidth to do this (this is related to https://github.com/Smithay/wayland-rs/issues/185 for those interested).

In the meantime, this works, and leaves room to easily implement stuff like disabling or discriminating key repetition events, depending on what is decided wrt #310.

Cargo.toml
```diff
@@ -55,7 +55,7 @@ features = [
 
 [target.'cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "openbsd", target_os = "netbsd"))'.dependencies]
 wayland-client = { version = "0.20.10", features = [ "dlopen", "egl", "cursor"] }
-smithay-client-toolkit = "0.2.6"
+smithay-client-toolkit = { git = "https://github.com/Smithay/client-toolkit" }
```
SirCmpwn: You can use 0.3.0 now.

trimental: I'm going to change this pull request to 'Upgrade to smithay-client-toolkit 0.3.0' as 0.3.0 contains few breaking changes from https://github.com/Smithay/client-toolkit/pull/25

trimental: Sorry my bad, it seems none of those breaking changes apply

francesca64:
Oh, that would be fine. The checklist system isn't perfect, though I don't really know if there's a better way... anyway, is this PR ready for merge?

src/platform/linux/wayland/keyboard.rs
```diff
@@ -1,77 +1,110 @@
...
 
 pub fn init_keyboard(
     keyboard: NewProxy<wl_keyboard::WlKeyboard>,
     sink: Arc<Mutex<EventsLoopSink>>,
+    events_loop_proxy: EventsLoopProxy,
 ) -> Proxy<wl_keyboard::WlKeyboard> {
-    // { variables to be captured by the closure
-    let mut target = None;
+    // { variables to be captured by the closures
+    let target = Arc::new(Mutex::new(None));
```
francesca64: `Default::default()` should be equivalent to this.

trimental:
I believe it should be in a state to merge after implementing your review point. I just want to be sure though that my understanding is correct that there is no difference between key press and key repeat events in winit and that they should be indistinguishable.

src/platform/linux/wayland/keyboard.rs
francesca64: Oh, since the type isn't inferred in this case, then it's preferable the way it already is. Sorry!

francesca64:
That's correct; our API presently has no indication of a key press being a repeat.

SirCmpwn:
Thanks! I'm sure you guys have your own release cycle but I wanted to mention that I'm looking forward to seeing this tagged for alacritty's sake.

francesca64:
This is going to be included in a new release as part of #630, which will merge as soon as the CI passes!

SirCmpwn:
:tada:

Issue #633, opened 2018-08-22
Race condition on Windows freezing window


DefinitelyNotRobot:
This is happening on Windows platform. There seems to be some form of race condition going on that freezes/deadlocks the events loop. Here is the code that causes it:

```
        use winit::*;
        let mut events_loop = winit::EventsLoop::new();
        let window = WindowBuilder::new().build(&events_loop).unwrap();
        std::thread::sleep_ms(20); // Somehow helps to trigger it more often
        window.set_fullscreen(Some(events_loop.get_primary_monitor()));
        window.set_title("test");
        loop {
            events_loop.poll_events(|_| {});
        }
```

francesca64:
This looks like it's the same as #574 or #617

francesca64:
Fixed via #634

Issue #617, opened 2018-07-29
Windows 10: Window freezes when updating title on window resized


jannik17:
Simple example to reproduce:

```
extern crate winit;

fn main() {
    let mut events_loop = winit::EventsLoop::new();
    let window = winit::Window::new(&events_loop).unwrap();

    events_loop.run_forever(|event| {
        match event {
            winit::Event::WindowEvent {
                event: winit::WindowEvent::CloseRequested,
                ..
            } => winit::ControlFlow::Break,
            winit::Event::WindowEvent {
                event: winit::WindowEvent::Resized(logical_size),
                ..
            } => {
                println!("About to freeze...");
                window.set_title("Resized");
                winit::ControlFlow::Continue
            }
            _ => winit::ControlFlow::Continue,
        }
    });
}
```

ArturKovacs:
Yes I also experienced this. I think this is related to #459.

fschutt:
This is likely because winit uses two threads right now on Windows. If you call any Win32 methods from a different thread than the thread where the window was created on, Windows / Win32 freezes your application. This is also a reason why you can't do `SetMenu(HWND)` from a different thread.

I mean, that's one way to solve race conditions...

Osspial:
I'd close this, but I don't currently have the permission to do that. @tomaka ping

Issue #574, opened 2018-06-18
cursor grab fails in fullscreen on windows


thiolliere:
When I add the line `window.set_cursor(winit::CursorState:Grab).unwrap()` after the window build in fullscreen example and run it on windows then it fails.

The window does not respond and windows ask to quit or wait.

rukai:
I think you mean: `window.set_cursor_state(winit::CursorState::Grab).unwrap()`

I have the same issue on Windows 10
It also occurs with `CursorState::Hide`.

The issue occurs when `with_fullscreen(Some(..))` is used to create a window while cursor is Grab or Hide.
The issue also occurs when the cursor is Grab or Hide and then `set_fullscreen(Some(..))` is called.

francesca64:
It looked like a deadlock was happening somewhere, and the results are a bit amusing.

As you might've heard by now, we block during resize events on Windows. However, we only unblock in `run_forever`/`poll_events`. Entering fullscreen triggers a resize, and thus a block. This means that all event processing is frozen in the duration between creating a fullscreen window and starting `run_forever`/`poll_events` execution, and due to some other fun complexities of our Windows implementation, cursor grabbing requires posting a message to the event queue. `set_cursor_state` doesn't return until it gets a result, but since event processing is still held up in the `WM_SIZE` handler, your program freezes.

Here's a really goofy temporary workaround you can use:
```rust
extern crate winit;

use winit::{ControlFlow, Event, WindowEvent};

fn main() {
    let mut event_loop = winit::EventsLoop::new();
    let monitor = event_loop.get_primary_monitor();
    let window = winit::WindowBuilder::new()
        .with_fullscreen(Some(monitor))
        .build(&event_loop)
        .unwrap();

    let mut run_count = Some(0);

    event_loop.run_forever(|event| {
        match event {
            Event::WindowEvent { event, .. } => match event {
                WindowEvent::Resized(_) => {
                    if let Some(count) = run_count {
                        let dpi_factor = window.get_hidpi_factor();
                        if (dpi_factor == 1.0 && count == 0)
                        || (dpi_factor != 1.0 && count == 1) {
                            window.grab_cursor(true).unwrap();
                            run_count = None;
                        }
                    }
                },
                WindowEvent::CloseRequested
                | WindowEvent::KeyboardInput {
                    input: winit::KeyboardInput {
                        virtual_keycode: Some(winit::VirtualKeyCode::Escape),
                        ..
                    }, ..
                } => return ControlFlow::Break,
[...]

PR #634, opened 2018-08-23
Remove resize block on Windows

Osspial:
Reintroduces some visual bugs when resizing, but fixes the freezes in #574, #617, and #633.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,11 @@
 # Unreleased
+- On Windows, fix freezes when performing certain actions after a window resize has been triggered.
 
 # Version 0.17.2 (2018-08-19)
 
 - On macOS, fix `<C-Tab>` so applications receive the event.
 - On macOS, fix `<Cmd-{key}>` so applications receive the event.
 - On Wayland, key press events will now be repeated.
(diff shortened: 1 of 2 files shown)


Osspial:
It looks like travis failed because of some network errors during the darwin/ios tests. This doesn't touch any of the code for those platforms, so a proper run shouldn't fail them.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - On macOS, fix `<C-Tab>` so applications receive the event.
 - On macOS, fix `<Cmd-{key}>` so applications receive the event.
 - On Wayland, key press events will now be repeated.
+- On Windows, fix freezes when performing certain actions after a window resize has been triggered.
```
atouchet: 0.17.2 is already out. This should go under Unreleased.

CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 # Unreleased
+- On Windows, fix freezes when performing certain actions after a window resize has been triggered.
```
francesca64: I'd appreciate if this also mentioned the reintroduction of the visual bug. (There should also be a newline between the heading and the list.)

PR #635, opened 2018-08-23
Wayland: commit frame surface on resize

trimental:

This PR commits the surface of the wayland window frame on resize  as suggested by @vberger.
This is required by https://github.com/Smithay/client-toolkit/commit/acdfde234e0afa3513b276493696811986aa2322

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Fixed graphical glitches when resizing on Wayland
+
 # Version 0.17.2 (2018-08-19)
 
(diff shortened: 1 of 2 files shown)


vberger:
For the record, decoration drawing by SCTK is blocked by commits on the main surface. SCTK itself avoids committing the surface by itself, because depending on the client's behavior it may introduce graphical glitches, notably during resizes (the decorations size and content size could have a few pixels of mismatch).

In this specific case, winit can commit without risk of glitches, because the decorations simply need to be refreshed, but not resized. This is used to implement button highlighting notably.

francesca64:
Thanks!

PR #638, opened 2018-08-24
Event Loop 2.0 API and Windows implementation

Osspial:
Changes winit's API as discussed in #459 and implements said changes on Windows.

This PR ***does not*** provide implementations for the new API on any non-Windows platform. Said platforms have not been tested, and there should be no expectations that they even compile. However, it *does* provide a working version of the API on Windows as well as API docs.

Before this is merged, we should thoroughly review the code changes and make sure all the documentation is correct/includes working links.

Changelog entry is included below for convenience and potential further discussion:
- Change all occurrences of `EventsLoop` to `EventLoop`.
- Previously flat API is now exposed through `event`, `event_loop`, `monitor`, and `window` modules.
- `os` module changes:
  - Renamed to `platform`.
  - All traits now have platform-specific suffixes.
  - Exposes new `desktop` module on Windows, Mac, and Linux.
- Changes to event loop types:
  - `EventLoopProxy::wakeup` has been removed in favor of `send_event`.
  - **Major:** New `run` method drives winit event loop.
    - Returns `!` to ensure API behaves identically across all supported platforms.
      - This allows `emscripten` implementation to work without lying about the API.
    - `ControlFlow`'s variants have been replaced with `Wait`, `WaitUntil(Instant)`, `Poll`, and `Exit`.
      - Is read after `EventsCleared` is processed.
      - `Wait` waits until new events are available.
      - `WaitUntil` waits until either new events are available or the provided time has been reached.
      - `Poll` instantly resumes the event loop.
      - `Exit` aborts the event loop.
    - Takes a closure that implements `'static + FnMut(Event<T>, &EventLoop<T>, &mut ControlFlow)`.
      - `&EventLoop<T>` is provided to allow new `Window`s to be created.
  - **Major:** `platform::desktop` module exposes `EventLoopExtDesktop` trait with `run_return` method.
    - Behaves identically to `run`, but returns control flow to the calling context can take non-`'static` closures.
  - `EventLoop`'s `poll_events` and `run_forever` methods have been removed in favor of `run` and `run_return`.
- Changes to events:
  - Remove `Event::Awakened` in favor of `Event::UserEvent(T)`.
    - Can be sent with `EventLoopProxy::send_event`.
  - Rename `WindowEvent::Refresh` to `WindowEvent::RedrawRequested`.
    - `RedrawRequested` can be sent by the user with the `Window::request_redraw` method.
  - `EventLoop`, `EventLoopProxy`, and `Event` are now generic over `T`, for use in `UserEvent`.
  - **Major:** Add `NewEvents(StartCause)`, `EventsCleared`, and `LoopDestroyed` variants to `Event`.
    - `NewEvents` is emitted when new events are ready to be processed by event loop.
      - `StartCause` describes why new events are available, with `ResumeTimeReached`, `Poll`, `WaitCancelled`, and `Init` (sent once at start of loop).
    - `EventsCleared` is emitted when all available events have been processed.
      - Can be used to perform logic that depends on all events being processed (e.g. an iteration of a game loop).
    - `LoopDestroyed` is emitted when the `run` or `run_return` method is about to exit.
- Rename `MonitorId` to `MonitorHandle`.


Osspial:
Who would be up for reviewing the changes made here? I don't have a good idea of who has access to a Windows machine, and I'd rather this didn't wait around in the backlog forever.

hardiesoft:
Hello, I've tried some experiments with this branch, and run into an issue where I get a panic if I try to spawn a win32 context menu on handling a right-click event using the TrackPopupMenuEx winapi function.
It's something to do with the context menu creating its own modal run loop I think, and then the panic occurs within SubclassInput::send_event when it tries to mutably borrow an already borrowed value.

It's pretty easy to replicate, but let me know if you'd like a code snippet.  I also note that this is already a great improvement to the current master winit event loop on win32 which has the second thread; since TrackPopupMenuEx needs to be called on the thread owning the Window, it just silently fails to create the context menu, and would require some inter-thread message passing shenanigans to work at all.  Overall I find the API design of this branch much nicer for my use-cases.

Cheers

Osspial:
@hardiesoft Thanks! The reason that's happening is because, as you said, `TrackPopupMenuEx` tries to send its own events. That *would* call into the event loop closure that called `TrackPopupMenuEx`, violating Rust's aliasing rules and causing undefined behavior, but the implementation has a `RefCell` that triggers the panic when we attempt to borrow it mutably twice.

The best way I can think of to solve this without changing the API would be to buffer up events if new events are generating during the closure then send them all off when the closure returns, although that would pause user event processing while the `TrackPopupMenuEx` function is being called.

It may also be worth adding a function to `EventLoop` that lets you buffer up a series of functions to be called after the event closure returns. With that API, you could do something like this:

```rust
    .build(&event_loop)
    .unwrap();

event_loop.run(move |event, event_loop, control_flow| {
    match event {
        Event::WindowEvent {
            ..
        } => *control_flow = ControlFlow::Exit,
        Event::WindowEvent {
            event: WindowEvent::MouseInput {
                state: winit::event::ElementState::Released,
                button: winit::event::MouseButton::Right,
                ..
            },
            ..
        } => {
            event_loop.queue_function(|| unsafe {
                use std::ptr;
                use winapi::um::winuser::{self, *};

                let menu = winuser::CreatePopupMenu();
                winuser::InsertMenuA(menu, 0, MF_BYPOSITION | MF_STRING, 0, "Hello\0".as_ptr() as _);
                winuser::TrackPopupMenuEx(
                    menu,
                    0,
                    512,
                    512,
                    window.get_hwnd() as _,
                    ptr::null_mut()
                );
                winuser::DestroyMenu(menu);
            });
        }
    }
});
```

And `winit` would call the popup-creation method after the main event loop closure returns, allowing user event processing to continue while `TrackPopupMenuEx` runs its own loop.

hardiesoft:
Hi @Osspial I do like your idea of queuing a function to be run after the main loop closure; for full flexibility ideally I don't want to stop other user event processing while a context menu (or OS file dialog etc) is open.  The main issue I see here is just one of documenting this properly if this is the direction you decide to go in - perhaps providing an example that addresses this use-case?  Would the proposed API be useful outside of Windows, or would it essentially be a platform specific thing to work around this issue?

If you decide it's not worth the specialised API to handle this I'd be okay with it just buffering the events (or even throwing them away) and blocking.  If I care about things like animations continuing to run in my window I'd probably have my render loop off the main thread, and I generally don't expect to receive input to my window while a context menu or other OS UI element has focus.  IIRC, macos will just block the calling thread in this case too, so it's sort of a lowest common denominator behaviour.

src/event_loop.rs
```diff
@@ -0,0 +1,155 @@
...
+    #[inline]
+    pub fn get_primary_monitor(&self) -> MonitorHandle {
+        MonitorHandle { inner: self.event_loop.get_primary_monitor() }
+    }
+
+    /// Hijacks the calling thread and initializes the `winit` event loop with the provided
+    /// closure. Since the closure is `'static`, it must be a `move` closure if it needs to
+    /// access any data from the calling context.
+    ///
+    /// Any values not passed to this function will *not* be dropped.
```
vberger: I thnk we need to specify somewhere the behavior of the `&mut ControlFlow`, specifically regarding these points:

- What is its initial value?
- Is it reset to its initial value at the beginning of each event loop run?

src/event_loop.rs
```diff
@@ -0,0 +1,155 @@
...
+            event_loop_proxy: self.event_loop.create_proxy(),
+        }
+    }
+}
+
+/// Used to send custom events to `EventLoop`.
+#[derive(Clone)]
+pub struct EventLoopProxy<T> {
+    event_loop_proxy: platform_impl::EventLoopProxy<T>,
+}
```
vberger: Do we guarantee that `EventLoopProxy<T>` is `Send` if `T` is `Send`?

src/platform/unix.rs
```diff
@@ -5,61 +5,61 @@ use std::ptr;
...
 
-pub use platform::XNotSupported;
-pub use platform::x11::util::WindowType as XWindowType;
+pub use platform_impl::XNotSupported;
+pub use platform_impl::x11::util::WindowType as XWindowType;
 
-/// Additional methods on `EventsLoop` that are specific to Linux.
-pub trait EventsLoopExt {
-    /// Builds a new `EventsLoop` that is forced to use X11.
+/// Additional methods on `EventLoop` that are specific to Linux.
```
vberger: If the trait is named `EventLoopUnix`, then these are methods that are specific to unix-like OSes, not only Linux. I believe *BSD uses x11 too for example.

caitp:
> @hardiesoft Thanks! The reason that's happening is because, as you said, TrackPopupMenuEx tries to send its own events. That would call into the event loop closure that called TrackPopupMenuEx, violating Rust's aliasing rules and causing undefined behavior, but the implementation has a RefCell that triggers the panic when we attempt to borrow it mutably twice.

This model doesn't seem to be compatible with [SendMessage](https://docs.microsoft.com/en-us/windows/desktop/api/winuser/nf-winuser-sendmessage), which has legitimate uses. Is there any way we can make event handlers reentrant?

Osspial:
@caitp The only way I can think of doing that would be to make the run functions take a `Fn` instead of a `FnMut`, but that carries API limitations that clearly aren't practical for any reasonable application.

dylanede:
How is continuous rendering meant to be handled with the new API? I'm trying to do my rendering in response to `WindowEvent::RedrawRequested`. I then call `Window::request_redraw()` after rendering in order to get schedule the next render. This doesn't seem to actually work though. If I instead put the redraw request in response to `Event::EventsCleared`, rendering only happens continuously whilst other events are firing (e.g. mouse movement). If I put the call in response to both `RedrawRequested` and `EventsCleared`, rendering is continuous, but the window is unresponsive.

dylanede:
I've also managed to get winit to hit unreachable code inside `EventLoopRunner::process_event`.

JohnDoneth:
I just ran all of the examples, and tried to break them in order to provide some useful feedback. This is what I have discovered.

Errors in examples:
- cursor_grab - pressing H a bunch of times causes the window to become unresponsive.
- handling_close - doesn't seem to work anymore when attempting to enter Y in the terminal.
- fullscreen - becomes unresponsive after entering fullscreen.

The other examples seemed to work correctly. I'm excited to see this PR progress! 🎉

dylanede:
Using a proxy in combination with `Poll` seems to cause that unreachable code to be hit as well.

Osspial:
@dylanede @JohnDoneth 

Thanks for the reports! I've pushed fixes that seem to fix all of those issues. I haven't tested the `Poll`/`EventLoopProxy` crash, but that should have been fixed as well.

src/event_loop.rs
Osspial: I've updated the docs to clarify this.

hardiesoft:
I also still hit that unreachable code, but only in --release mode.

dylanede:
@Osspial I think the problem I mentioned with `Poll` and proxies is in fact just the issue @hardiesoft has reported above, since it only happened in release mode and I have not boiled it down yet.

Osspial:
@hardiesoft Your example should be fixed in the latest commit. @dylanede, can you run your `Poll` code again and see if that crashes?

dylanede:
@hardiesoft well done! That's stopped the crashes for me.

vberger:
What is the status of this? Can it be merged into `eventloop-2.0` to bootstrap the implementation of the other platforms?

Osspial:
@vberger AFAIK the main thing this is waiting on is a review of the code. @francesca64 volunteered to do it, but it doesn't seem like she's found the time to make that happen.

est31:
Ping  @francesca64

Osspial:
The API this exposes is pretty stable, and it's taking a while time to get a code review done. If a review can't happen soonish I think it may make sense to merge this as-is so that people can start working on implementations for other platforms, then do a review later as required.

Osspial:
@mtak- I wrote up an answer to that in the API change thread, but it's been buried pretty thoroughly (*I* couldn't even find it at first) so I can't blame you for not finding it 😄! Regardless, I've pasted that here.

> Why do this? With the return-based solution, there are cases where there are several "correct" ways to handle control flow. Consider the following event loop:
> 
> ```rust
> event_loop.run(|event| {
>     match event {
>         Event::LoopEnd => ControlFlow::Poll,
>         _ => ControlFlow::Wait
>     }
> });
> ```
> 
> That situation leads to return patterns like `Wait, Wait, Wait, {...}, Poll`. What do you do with the `Wait`s that are returned before the `Poll`? You could take that as "halt the event loop between each call to the function", which is pretty clearly nonsense. However, it's the only solution that doesn't involve throwing away values. Other solutions could be to have some sort of "control flow precedence pattern" where returning one type overrides future returns of another type, or to ignore all of the return values except for the one from `LoopEnd` (in which case what's the point of having it return anyway!).
> 
> As a persistent parameter, there are no values we have to ignore and it's clear when the control flow of the loop actually changes. Under this, `Wait`, `WaitTimeout`, and `Poll` would only impact control flow when all of a loop's events have been cleared, and `Break` would essentially kill the loop immediately (barring destruction events).

I really need to make a post summarizing all the rationale discussion from that thread...

Also @francesca64, I've been rebasing this onto master and have gotten to a point where, to include the new windows drop handler implementation, it looks I need to use `crossbeam_channel`'s `select` macro to get `FileDropHandlerData`'s event sending working. However, that has a minimum Rust version of `1.26`. How should we handle that?

mtak-:
Ooops! Burned again by the Hidden Items while cmd+f'ing. Thanks that makes a lot of sense! I've started on an iOS implementation.

Osspial:
@mtak- Thanks for working on that! I just rebased this PR onto master, so you may want to update to the new code before you get too far into your implementation.

vberger:
I'm ok with merging this as it is now, so that other platforms can start working on the API.

francesca64:
When testing on Windows 10, I've noticed that `set_decorations` causes the window to freeze. `with_decorations` works fine, though. This can be reproduced in the fullscreen example app by pressing D.

Osspial:
@francesca64 Thanks! That should be fixed now.

src/platform_impl/windows/event_loop.rs
```diff
@@ -0,0 +1,1603 @@
...
+    {
+        unsafe{ winuser::IsGUIThread(1); }
+
+        assert_eq!(mem::size_of::<RootEventLoop<T>>(), mem::size_of::<EventLoop<T>>());
+        let self_ptr = self as *const EventLoop<T>;
+
+        let mut runner = unsafe{ EventLoopRunner::new(
+            self,
+            move |event, control_flow| {
+                let event_loop_ref = &*(self_ptr as *const RootEventLoop<T>);
```
vberger: This kind of trick cannot be done for x11/wayland (as there is an internal layer of `EventLoop` type that is an enum and does the dispatch between x11 and wayland). As such, I'm not sure how I can implement `run_return` for wayland without some seriously UB black magic.

Osspial: The original design added a new `EventLoopWindows` type used for window creation that got passed to the callback where `EventLoop` currently gets passed. I removed it because I realized you could pass a immutable reference to `EventLoop` to achieve the same effect, but it may make sense to add that type back in to reduce the amount of unsafe shenanigans that have to take place.

vberger: I think that'd be better, yes.

Osspial: I've re-added that type as `EventLoopWindowTarget`.

vberger: Ah, nevermind, I missed some subtlety, I can work. Won't be extremely pretty, but that's the fault of the 2-in-1 x11/Wayland backend

vberger:
Okay, I have a wayland implementation that compiles. Do we stick to the initial plan, and I make a PR for it to the `eventloop-2.0` branch, once this is merged?

Osspial:
@vberger That sounds good.

Also, I just went though the code and fixed some places where freezes happened after setting window state, so those issues should be resolved going forward. Let me know if anyone runs into another one.

francesca64:
While the fix worked, many methods still don't work correctly if called from another thread:
```rust
extern crate winit;

fn main() {
    let event_loop = winit::event_loop::EventLoop::new();
    let window = winit::window::WindowBuilder::new()
        .with_dimensions((400, 300).into())
        .build(&event_loop)
        .unwrap();
    let (tx, rx) = std::sync::mpsc::channel();
    std::thread::spawn(move || {
        loop {
            match rx.try_recv() {
                Ok(event) => match event {
                    winit::event::WindowEvent::CloseRequested => break,
                    winit::event::WindowEvent::KeyboardInput { input: winit::event::KeyboardInput {
                        state: winit::event::ElementState::Pressed,
                        virtual_keycode: Some(key),
                        modifiers,
                        ..
                    }, .. } => {
                        use winit::event::VirtualKeyCode::*;
                        match key {
                            Escape => break,
                            // These do nothing
                            F => {
                                let monitor = match modifiers.shift {
                                    true => None,
                                    false => Some(window.get_current_monitor()),
                                };
                                window.set_fullscreen(monitor);
                            },
                            M => window.set_maximized(!modifiers.shift),
                            D => window.set_decorations(!modifiers.shift),
                            A => window.set_always_on_top(!modifiers.shift),
                            // These freeze (i.e. the close button/escape won't work)
                            G => window.grab_cursor(!modifiers.shift).unwrap(),
                            H => window.hide_cursor(!modifiers.shift),
                            _ => (),
                        }
                    },
                    _ => (),
                },
                Err(_) => (),
            }
        }
        drop(window);
    });
        *control_flow = winit::event_loop::ControlFlow::Wait;
        match event {
            winit::event::Event::WindowEvent { event, .. } => {
                if let winit::event::WindowEvent::Destroyed = event {
                    *control_flow = winit::event_loop::ControlFlow::Exit;
                } else {
                    let _ = tx.send(event);
                }
            },
            _ => (),
        }
    })
}
```

francesca64:
Also, entering/exiting fullscreen no longer generates a `Resized` event.

Osspial:
@francesca64 Both of those issues should be fixed now.

francesca64:
I found a pretty niche regression. When I use screen sharing in Discord, winit programs will panic as soon as I try to select an application window:
```
thread 'main' panicked at 'already borrowed: BorrowMutError', libcore\result.rs:945:5
stack backtrace:
   0: std::sys::windows::backtrace::unwind_backtrace
             at C:\projects\rust\src\libstd\sys\windows\backtrace\mod.rs:65
   1: std::sys_common::backtrace::_print
             at C:\projects\rust\src\libstd\sys_common\backtrace.rs:71
   2: std::sys_common::backtrace::print
             at C:\projects\rust\src\libstd\sys_common\backtrace.rs:59
   3: std::panicking::default_hook::{{closure}}
             at C:\projects\rust\src\libstd\panicking.rs:211
   4: std::panicking::default_hook
             at C:\projects\rust\src\libstd\panicking.rs:227
   5: std::panicking::rust_panic_with_hook
             at C:\projects\rust\src\libstd\panicking.rs:511
   6: std::panicking::continue_panic_fmt
             at C:\projects\rust\src\libstd\panicking.rs:426
   7: std::panicking::rust_begin_panic
             at C:\projects\rust\src\libstd\panicking.rs:337
   8: core::panicking::panic_fmt
             at C:\projects\rust\src\libcore\panicking.rs:92
   9: core::result::unwrap_failed<core::cell::BorrowMutError>
             at C:\projects\rust\src\libcore\macros.rs:26
  10: core::result::Result<core::cell::RefMut<core::option::Option<winit::platform_impl::platform::event_loop::EventLoopRunner<()>>>, core::cell::BorrowMutError>::expect<core::cell::RefMut<core::option::Option<winit::platform_impl::platform::event_loop::EventLoopRunner<()>>>,core::cell::BorrowMutError>
             at C:\projects\rust\src\libcore\result.rs:809
  11: core::cell::RefCell<core::option::Option<winit::platform_impl::platform::event_loop::EventLoopRunner<()>>>::borrow_mut<core::option::Option<winit::platform_impl::platform::event_loop::EventLoopRunner<()>>>
             at C:\projects\rust\src\libcore\cell.rs:826
  12: winit::platform_impl::platform::event_loop::public_window_callback<()>
             at .\src\platform_impl\windows\event_loop.rs:880
  13: DefSubclassProc
  14: DefSubclassProc
  15: CallWindowProcW
  16: CallWindowProcW
  17: IsWindowVisible
  18: KiUserCallbackDispatcher
  19: NtUserGetMessage
  20: GetMessageW
  21: winit::platform_impl::platform::event_loop::EventLoop<()>::run_return<(),closure>
             at .\src\platform_impl\windows\event_loop.rs:245
  22: winit::platform_impl::platform::event_loop::EventLoop<()>::run<(),closure>
             at .\src\platform_impl\windows\event_loop.rs:190
  23: winit::event_loop::EventLoop<()>::run<(),closure>
             at .\src\event_loop.rs:145
  24: window::main
             at .\examples\window.rs:14
  25: std::rt::lang_start::{{closure}}<()>
             at C:\projects\rust\src\libstd\rt.rs:74
  26: std::rt::lang_start_internal::{{closure}}
             at C:\projects\rust\src\libstd\rt.rs:59
  27: std::panicking::try::do_call<closure,i32>
             at C:\projects\rust\src\libstd\panicking.rs:310
  28: panic_unwind::__rust_maybe_catch_panic
             at C:\projects\rust\src\libpanic_unwind\lib.rs:105
  29: std::panicking::try
             at C:\projects\rust\src\libstd\panicking.rs:289
  30: std::panic::catch_unwind
             at C:\projects\rust\src\libstd\panic.rs:392
  31: std::rt::lang_start_internal
             at C:\projects\rust\src\libstd\rt.rs:58
  32: std::rt::lang_start<()>
             at C:\projects\rust\src\libstd\rt.rs:74
  33: main
  34: invoke_main
             at f:\dd\vctools\crt\vcstartup\src\startup\exe_common.inl:78
  35: __scrt_common_main_seh
             at f:\dd\vctools\crt\vcstartup\src\startup\exe_common.inl:288
  36: BaseThreadInitThunk
  37: RtlUserThreadStart
error: process didn't exit successfully: `target\debug\examples\window.exe` (exit code: 3221226505)
```
This doesn't happen on master, and can be reproduced with the `window` example.

Osspial:
@francesca64 That too works now.

mtak-:
iOS is [ready](https://github.com/mtak-/winit/tree/events_loop_2.0_ios). Just waiting on a merge.

francesca64:
I've been working on a macOS implementation, and it's just about complete.

 @Osspial I realize in retrospect that rebasing could be a pain, especially since more things will continue to change on master... feel free to just force-push that branch back (which I think you can do now?) if you'd rather merge as-is. Alternatively, we could work towards one last pre-EV2 and then freeze master and rebase everything.

Osspial:
@francesca64 Thanks for reviewing this! Sorry it took so long to respond to those - for some reason I didn't see that you posted the review.

Anyhow, we've got a lot of unreleased changes, so I think it's a good idea to go through any recent PRs that aren't much effort to merge, merge those, then freeze master and get ready to release this.

Osspial:
@francesca64 It's... been a while. Any status updates on the X11 and MacOS backends? From what you said it sounded like they were pretty much ready to go, but it's been over a month since we've heard anything about it.

If those two are ready, I'm tempted to go ahead and release what we have (Windows, MacOS, X11/Wayland, and iOS). Emscripten is already broken anyway (notwithstanding @iceiix's work in #760 and #767), and I don't want to hold up everyone else for Android support - that can be patched in in a future update, when you have time to work on it.

Osspial:
Also, this is now merged against the latest master! It was a whole lot less of a PITA than rebasing, but it still takes time so I don't want to do that sort of thing yet again. (pro tip: if you've got a massive PR like this use `git merge` not `git rebase` if you enjoy having time to do other things)

vberger:
Is there anything blocking the merge of this in the `eventloop-2.0` branch right now, so we can setup the other PRs to it and hopefully merge the whole thing to master when it is ready?

Or has this part of the initial plan been completely dropped?

francesca64:
@Osspial 

> From what you said it sounded like they were pretty much ready to go

macOS is indeed complete (aside from some minor correctness issues that I know how to fix but haven't gotten around to yet), but I haven't done any work on X11 aside from the rough one I did during the API discussion. An X11 implementation should be easy, though.

> I don't want to hold up everyone else for Android support - that can be patched in in a future update, when you have time to work on it.

That's reasonable. I'll definitely be handling the Android implementation, but it's not the priority right now.

> Emscripten is already broken anyway

I'm honestly not sure how many people will want to use/maintain Emscripten after #589 anyway, so I'm fine with this too.

chris-morgan:
Just want to say: I’m delighted with this, because my Win32 EdgeHTML experimentation (Windows.Web.UI.Interop.WebViewControl, which incidentally requires a single-threaded apartment) had not been working, in a very opaque manner, and I finally today tried minimising and translating a known-good C++ implementation, and that worked; the key difference then was the use of winit, which I had erroneously thought a completely harmless component. I began to get suspicious of its event loop implementation (though I still don’t have any sort of grasp on *why* it makes an IAsyncOperation seem to be working but never complete, even when polling on the original thread). So I came to the issue tracker, and found recent discussion about all this and this very-recently-merged work; and guess what? It all works now. So with this work I will *finally* be able to resume work on a good web view control (MSHTML is disqualified from that classification, naturally).

So: thanks for sorting this out!

Osspial:
I just realized that, instead of aborting the entire process at the end of `run`, we could potentially use `libunwind` to unwind the stack. I just tested out building and I wasn't able to get `libunwind` compiling on Windows MSVC (so we'd have to sort that out), but it's a potential avenue forwards for allowing execution of `run` on non-main threads.

Osspial:
@francesca64 Do you have any sort of timeline for when you think the MacOS and X11 backends will be done by? I really don't want to be pushy, but these changes have been in progress for the better part of a year I'd like to move on from this. There's a bunch of stuff I'd like to make, both within Winit and without, that are either waiting on this to get going or would be significantly improved from having these overhauls on crates.io. I'm tired of having to choose between either writing Winit stuff that will make EL2.0 harder to merge, or writing stuff against the EL2.0 branch and not knowing when that code will see the light of day.

Osspial:
If there have been any personal or work issues that have made it hard to work on this, that's okay. Stuff happens. But (and I'll admit that I absolutely need to get better on this too) going radio silent doesn't really work.

francesca64:
@Osspial I'm not interested in working on an X11 implementation at this time. As for macOS, that's something I have to coordinate with @mtak- (going forward, macOS and iOS will likely have shared components), though there shouldn't be much left that needs to be done.

vberger:
I may be able to work on the x11 migration, with two caveats:

- It'll be a lot of `I-have-no-idea-what-Im-doing.gif`
- It won't be before two weeks

So if someone feels more competent and/or has more time, don't hesitate. :)

grovesNL:
winit Events Loop 2.0 with wasm32-unknown-unknown seems relevant to the `requestAnimationFrame` discussion in https://github.com/rustwasm/gloo/issues/1 if anyone has any ideas to contribute there.

Osspial:
@francesca64 If you don't have the time to finish the macOS implementation up, could you post a link to your current work and whatever remaining issues it has so that someone else could potentially pick it up and finish it?

vberger:
For the record, I'm in the process of porting the X11 backend to eventloop 2.0.

I'm finished with the "fixing imports and moving stuff around" part, and I'm down to the hard part: implementing `run`, `run_return` and `request_redraw`.

ZeGentzy:
I intend to port the MacOS via to use of `cargo check --target x86_64-apple-darwin`, and if I can get a VM to work, some minimal testing. Probably will take 2-3 weeks and will be not tested to actually be functional - only to compile.

ryanisaacg:
@ZeGentzy I have a Mac but not much time on my hands right now, but if you ping me during development I'll be able to build and run your branch within a day or two.

Osspial:
@ZeGentzy I finally managed to find @francesca64's WIP implementation! This branch should provide a good start: https://github.com/francesca64/winit/tree/macos-ev2

src/platform_impl/linux/wayland/event_loop.rs
```diff
@@ -323,24 +323,80 @@ impl SeatManager {
 
     fn remove_seat(&mut self, id: u32) {
         use self::wl_seat::RequestsTrait as SeatRequests;
+<<<<<<< HEAD:src/platform_impl/linux/wayland/event_loop.rs
```
omni-viral: Is this intentional?

vberger: This was an othersight that was fixed in #790

omni-viral: Yeah. I just wonder how CI didn't find it.

vberger: CI didn't pass at all for the other platforms than windows, which were not ported yet in this PR.

nvzqz:
Is there a reason for why extension traits were renamed to include the name of the operating system? This isn't usually the standard convention (and feels awkward imo) so I'm curious as to the reasoning for this.

PR #641, opened 2018-08-27
Add support for F16-F24

overlazy:

This should resolve this [issue](https://github.com/jwilm/alacritty/issues/1525).


Osspial:
If we're going to be merging this, is there any reason to not expose `F20`-`F24` as well?

overlazy:
I think we can add F20. As for F21-F24 I didn't find any useful reference about them on mac os. There is no information about these keys in Events.h file either ( F20 is the last key code presented in it).

francesca64:
[Qt's implementation](https://github.com/qt/qtbase/blob/c5307203f5c0b0e588cc93e70764c090dd4c2ce0/src/corelib/kernel/qcore_mac_objc.mm#L357) uses [these unicode constants](https://developer.apple.com/documentation/appkit/1535851-function_key_unicodes?language=objc). It looks like GLFW and GTK only go up to F20 on macOS, and SDL only goes up to F12. While the majority of macOS users won't have these higher function keys at all, I still think it would be super great to support them!

For implementing this here, matching on the `characters` attribute of the keyboard event looks like the right way to go. This would be handled in our [view class's `keyDown` handler](https://github.com/tomaka/winit/blob/a70bc2082908a189d613f42534c9f39fd7df572f/src/platform/macos/view.rs#L347) and then again for `keyUp`. Basically, if `to_virtual_key_code` returns `None`, the `characters` would be passed to a new function that checks for these function keys.

OvermindDL1:
What about supporting all the XFree86 standard keycodes?  I have a highly extended keyboard that has not only everything from F1 to F24 but also a great deal of other keys that put out other 'unusual' but standard XFree86 codes.  I have to admit it is highly irritating that many engines (though not all) don't respond to many of my keys and would like anything based on winit to work properly.  :-)

src/platform/macos/view.rs
```diff
@@ -351,8 +351,27 @@ extern fn key_down(this: &Object, _sel: Sel, event: id) {
         let state = &mut *(state_ptr as *mut ViewState);
         let window_id = WindowId(get_window_id(state.window));
 
+        state.raw_characters = {
```
francesca64: It would be really nice to have the code for reading the characters to a `String` put into a function, since it's repeated now.

src/platform/macos/view.rs
```diff
@@ -351,8 +351,27 @@ extern fn key_down(this: &Object, _sel: Sel, event: id) {
...
+            let string = str::from_utf8_unchecked(slice);
+            Some(string.to_owned())
+        };
+
         let keycode: c_ushort = msg_send![event, keyCode];
-        let virtual_keycode = to_virtual_key_code(keycode);
+        let virtual_keycode = {
+            let mut v_code = to_virtual_key_code(keycode);
+            if v_code.is_none() {
+                // We are checkig here for F21-F24 keys, since their keycode
```
francesca64: checkig -> checking

src/platform/macos/events_loop.rs
```diff
@@ -680,6 +680,23 @@ pub fn to_virtual_key_code(code: c_ushort) -> Option<events::VirtualKeyCode> {
     })
 }
 
+pub fn to_virtual_key_code_ex(
```
francesca64: The suffix `ex` seems like a Windows naming convention to me, and feels a bit out of place in macOS code. Either way, a more explicit/specific name would be nice, since this function is exclusively for function keys. It would also be good to have a comment briefly explaining why a separate function is needed, paired with a link to the [relevant docs](https://developer.apple.com/documentation/appkit/1535851-function_key_unicodes?language=objc).

src/platform/macos/view.rs
```diff
@@ -351,8 +351,27 @@ extern fn key_down(this: &Object, _sel: Sel, event: id) {
...
+            );
+            let string = str::from_utf8_unchecked(slice);
+            Some(string.to_owned())
+        };
+
         let keycode: c_ushort = msg_send![event, keyCode];
-        let virtual_keycode = to_virtual_key_code(keycode);
+        let virtual_keycode = {
+            let mut v_code = to_virtual_key_code(keycode);
+            if v_code.is_none() {
```
francesca64: It's cleaner to use [`or_else`](https://doc.rust-lang.org/core/option/enum.Option.html#method.or_else).

src/platform/macos/view.rs
```diff
@@ -416,8 +425,27 @@ extern fn key_up(this: &Object, _sel: Sel, event: id) {
 
         state.last_insert = None;
 
+        // We need characters here to check for additional keys such as
+        // F21-F24.
+        state.raw_characters = {
```
francesca64: I'd prefer you just bind the characters to a local variable here, since using the `raw_characters` field (which is specifically used for special handling in `keyDown`) could cause confusion.

francesca64:
@OvermindDL1 I think that would be good to put into a separate PR, especially if it's mostly Linux-specific. It would be a really easy first contribution to make, if you're interested, especially since you'd be the most qualified to test it anyway.

PR #643, opened 2018-09-05
Expose platform.

ZeGentzy:

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -121,7 +121,7 @@ pub use icon::*;
 pub mod dpi;
 mod events;
 mod icon;
-mod platform;
+pub mod platform;
 mod window;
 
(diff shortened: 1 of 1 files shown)


Osspial:
Are we sure we want to do this? The `platform` module is an implementation detail, and all the actual platform-specific APIs are exposed through the `os` module. I'd figure that anything which needs an exposed `platform` module should instead be done by extending `os`'s OS-specific traits.

vberger:
I agree with @Osspial , I believe most of the APIs in platform were designed with the intent to be crate-internal. There is for example a large mix of `pub` and `pub(crate)` in them.

Even assuming it was *wanted* to expose the module, I doubt it can reasonably be exposed without a significant cleanup & refactor.

Issue #403, opened 2018-02-08
Pop-up menus


Shookbelf:
Hi, is there any way to create pop-ups like for right click menus with winit? Since right-click menus can also go across the borders of a window, I suppose they need to have their own window and surface.

It would be nice if functionality like this could be included.

paulirotta:
This is done as a higher level function of glutin or other UI lib that creates the windows and text on them.

Unless an extension binds to the native UI. React Native without the HTML garbage. It would be a most excellent direction but not a small mandate/extension. As a consultant I see a lot of wasted money and lives porting the same app to many platforms just to get the native components.

Shookbelf:
I am not talking about the drawing of the menu content. My current understanding is that those are basically specialized windows. If I want to spawn a window like that it's currently not possible with winit, because it will behave like a regular window.  
Perhaps it would be enough to make it possible to hide windows from the task manager.  
I think the ability to define parent windows is somehow related.

Please correct me if my understanding is wrong.

Osspial:
Do you mean having some way to create windows that aren't placed on the taskbar?

paulirotta:
I believe that is correct- small "windows" without edge decoration or appearing in the taskbar for things like menus. It is then up to a higher level lib (out of scope- you choose what to do) to populate these.

Shookbelf:
@paulirotta Unfortunately I don't know exactly how it is done on different platforms.  
This could be enough if the following features are supported:
- The window should not show up in task bars or similar
- It should not have borders
- Focus should be handled correctly

Most things can probably be handled by users of winit if the basics work.

There might be more subtle behaviour on different platforms though.

In xdg_shell there is a protocol xdg_popup for exactly this type of windows.

Osspial:
Borderless window handling is already supported. As far as the other two features go, implementing them on Windows seems to be pretty easy, and I've done so in these two branches: [focus](https://github.com/osspial/winit/tree/focusable) and [taskbar](https://github.com/osspial/winit/tree/popup_window).

I don't know how it would be handled on other platforms, but I can't imagine it's too much of a challenge.

Osspial:
Something worth noting is that Wayland doesn't let you specify where a window will be placed on-screen, which complicates right-click menu creation - it's expected that a menu would be created at the mouse pointer, but there's currently no way to ensure that happens on Wayland.

@vberger Does Wayland provide any way to work around this?

vberger:
[...]

PR #646, opened 2018-09-09
Update wm spec hints

svenstaro:

This just puts in all hints from [here](https://specifications.freedesktop.org/wm-spec/wm-spec-1.5.html#idm140200472629520). The old version used wm-spec v1.3.

This will help #403 at least a little for modern WMs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 - Fixed graphical glitches when resizing on Wayland.
 - On Windows, fix freezes when performing certain actions after a window resize has been triggered. Reintroduces some visual artifacts when resizing.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 - Fixed graphical glitches when resizing on Wayland.
 - On Windows, fix freezes when performing certain actions after a window resize has been triggered. Reintroduces some visual artifacts when resizing.
+- Updated window manager hints under X11 to v1.5 of "Extended Window Manager Hints" ([see here](https://specifications.freedesktop.org/wm-spec/wm-spec-1.5.html#idm140200472629520))
```
francesca64: I'd prefer if "Extended Window Manager Hints" were the link text (instead of "see here") with no quotation marks, and if it ended in a period.

svenstaro:
@francesca64 ok, how about now?

francesca64:
Looks great, thanks!

PR #648, opened 2018-09-09
Test against 1.26.2

### LINKED ISSUE #645: New versions of `parking_lot` are failing builds for 1.24.1
The minimum compiler version that `winit` tests is `1.24.1`, but `parking_lot` has been updated and the version we're pulling no longer supports that version of the compiler.

Should we bump the minimum version, or make `parking_lot`'s version explicitly support `1.24.1`?

Osspial:
Bumps minimum tested version to `1.26.2`. Fixes #645.

Should this have a changelog entry? I haven't put on in here since there don't seem to be any for previous version bumps, but I can add it in if necessary.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ matrix:
           packages: *i686_packages
     - env: TARGET=i686-unknown-linux-gnu
       os: linux
-      rust: 1.24.1
+      rust: 1.26.2
       addons:
         apt:
           packages: *i686_packages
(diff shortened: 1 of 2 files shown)


Osspial:
@francesca64

francesca64:
I'm definitely in favor of having this mentioned in the CHANGELOG. In fact, I think it would be great to have our minimum version requirement detailed in the README too.

Osspial:
Cool! I guess we don't have much of a reason to do this now.

It may be worth looking at raising the minimum version once Rust 2018 comes out, though.

PR #649, opened 2018-09-11
Put badges next to each other instead of below eachother

svenstaro:
This makes it look more like other projects have it. I always thought the way winit had it looked odd.
However, I see a certain logic in which you have arranged them so just close this if you want to keep it like you have it.


francesca64:
Looks good to me, thanks!

PR #651, opened 2018-09-12
Make `parking_lot` optional

### LINKED ISSUE #650: Make `parking_lot` dependency optional
Right now winit depends on `parking_lot` for Mutex handling - while I realize that `parking_lot` offers faster Mutex implementations than the standard library, it is also a pretty heavy dependency:

```
parking_lot v0.6.3
├── lock_api v0.1.3
│   ├── owning_ref v0.3.3
│   │   └── stable_deref_trait v1.1.1
│   └── scopeguard v0.3.3
└── parking_lot_core v0.2.14
    ├── rand v0.4.3
    └── smallvec v0.6.5
        └── unreachable v1.0.0
            └── void v1.0.2
```

This adds a lot of overhead to the compile time. winit is currently using `parking_lot` pretty much only for the `Mutex`, which can be exchanged with the `std::sync::Mutex`, so a simple feature flag should be possible so that crates depending on winit can turn off the `parking_lot` mutex and use the standard library mutex - many crates don't *need* that extra performance and would rather go with better compile times.

fschutt:
Fixes #650 - `parking_lot` is now an optional dependency, replacing `parking_lot::Mutex` by `std::sync::Mutex` if the feature `parking_lot_mutex` is disabled.

TODO:

- [ ] Windows / Mac should not need to download `parking_lot` if the feature is enabled
- [ ] Make `parking_lot` the default implementation for Linux (opt-out instead of opt-in)
- [ ] Document change in `CHANGELOG.md`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ features = ["icon_loading"]
 
 [features]
 icon_loading = ["image"]
+parking_lot_mutex = ["parking_lot"]
 
 [dependencies]
 lazy_static = "1"
@@ -57,5 +58,5 @@ features = [
 wayland-client = { version = "0.20.10", features = [ "dlopen", "egl", "cursor"] }
 smithay-client-toolkit = "0.3.0"
 x11-dl = "2.18.3"
-parking_lot = "0.6"
 percent-encoding = "1.0"
+parking_lot = { version = "0.6", optional = true }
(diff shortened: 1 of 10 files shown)


fschutt:
I'll close this in favor of a custom `Mutex` implementation. I'll also need to see if it even makes sense to keep `parking_lot` (i.e. how much traffic these mutexes actually receive in practical use). Afaik, nobody has yet tested that `parking_lot` actually makes things faster.

Issue #304, opened 2017-10-04
Configuration file support


WaDelma:
I wanted to create configuration file for keybindings, but I am not sure how to approach it.

Preferably I would want to to save the `VirtualKeyCode` to the configuration file as it's human readable and load it, but I can only go one way: from the code to string (by using `Debug` implementation which isn't the best). I could implement giant match for going the other way, but that seems to be a thing that `winit` should be doing (maybe with something like `str_to_enum_derive` crate).

As serialising/deserialising `VirtualKeyCode` isn't the easiest right now I thought I could do it with scan codes. I don't think that hardcoding the scan codes would be the best idea even just for the default settings, so I looked way to convert `VirtualKeyCode` to one, but there doesn't seem to be way to do this either.

So I am left wondering how configuration file should be done.

francesca64:
@WaDelma does the [`serde` feature](https://github.com/tomaka/winit/pull/652) added in 0.18.0 resolve this for you?

PR #652, opened 2018-09-12
Add optional Serde implementations and missing derivable traits

Osspial:
This is useful for things like configuration files, especially for saving keyboard/mouse bindings and such. This addresses #304.

Implementations are added for:
- All types in `dpi`.
- `KeyboardInput`
- `ElementState`
- `MouseButton`
- `MouseScrollDelta`
- `VirtualKeyCode`
- `ModifiersState`
- `ControlFlow`
- `MouseCursor`
- `WindowAttributes`

This PR also fills in a couple of API holes where some types were missing standard derivable implementations (`PartialEq`, `Eq`, and `Hash`), though that can be moved out of this PR if we decide we don't want to add Serde support.


src/lib.rs
```diff
@@ -419,6 +426,8 @@ pub struct WindowAttributes {
     /// Whether the window should be set as fullscreen upon creation.
     ///
     /// The default is `None`.
+    #[cfg_attr(feature = "serde", doc("\nThis field does not currently get serialized or deserialized."))]
+    #[cfg_attr(feature = "serde", serde(skip))] // TODO: FIGURE OUT HOW TO SERIALIZE THIS
     pub fullscreen: Option<MonitorId>,
```
vberger: I'm not sure this can be serialized in any meaningful way. This would require that the monitors have a consistent ids as seen by the app between two executions. Do any platform provide this kind of guarantee ?

Osspial: I can't say I know enough about the platform-specific APIs to give this a firm answer. However, I'd be really surprised if they didn't - being able to enumerate over a consistent list of monitors and choose the right one is really important on multi monitor setups, especially since the first monitor reported isn't always the one users want your application to be on.

Since you're brining this up, does Wayland not provide APIs to do this consistently?

vberger: To be honest I'm not sure.

- On the one hand there are the global ids, that uniquely identify a monitor, but have no guarantee to be stable across reboots, not even to still be valid at all.
- On the other hand, the wayland server reports the "name" of the monitor (in the form of model an make), which is stable across reboots, but are not unique identifiers (two monitors of the same model would have the same name I believe).

Osspial: Hmm. Is the order in which displays are reported consistent across reboots? If that's the case, we could see about serializing an index into the reported display list, although that would be rather hacky.

If we can't find a good solution for this soonish, I'd say that we should make a separate issue for figuring out how to proceed and then merge this as-is.

vberger: I think leaving this as a separate issue is probably a good approach here.

Osspial: I'm okay with that.

Osspial:
Is there anything blocking this from getting merged? @tomaka, you gave this a thumbs-down; is there any particular reason you object to this?

Osspial:
...?

Osspial:
@francesca64 @tomaka 

Unless there are any objections, I don't think there's anything blocking this from getting merged.

jwilm:
Alacritty copies the `VirtualKeyCode` enum, derives serde traits on it, and adds a conversion with the actual winit type. We would love to see serde support for this type!

ekse:
Another vote in favor of the conversion of VirtualKeyCode for the veloren project, see https://github.com/tomaka/glutin/issues/1074.

Osspial:
@mitchmindtree I would be alright with removing the `WindowAttributes` implementation. I don't think there are any other types where implementing serialization traits is a real issue (except maybe `ControlFlow`, but I don't think that's a big deal).

CHANGELOG.md
```diff
@@ -21,6 +21,8 @@
 - On X11, now a `Resized` event will always be generated after a DPI change to ensure the window's logical size is consistent with the new DPI.
 - Added further clarifications to the DPI docs.
 - On Linux, if neither X11 nor Wayland manage to initialize, the corresponding panic now consists of a single line only.
+- Add optional `serde` feature with implementations of `Serialize`/`Deserialize` for DPI types, various event types, and `WindowAttributes`.
```
vberger: Given this PR does not longer derive serialize traits for `WindowAttributes`, this changelog line should be updated.

vberger:
@mitchmindtree 
> maybe this PR should add a line to the .travis.yml for this?

I'd suggest adding a travis lines to test this, as well as a test file that asserts that the types inquestion do indeed implement the serialize/deserialize traits, to ease future maintenance/refactors/whatever should happen.

Osspial:
Test lines/files have been added. I also added in lines for the `icon_loading` feature, since they didn't seem to be in there.

Osspial:
...huh, the `icon_loading` feature breaks `1.24.1`.

francesca64:
Oh gosh! Since we didn't have `icon_loading` tested on CI before, #683 must've broken 1.24.1 compatibility without me realizing.

From looking at the logs, it would be easy to make `image` 0.20.0 work on 1.24.1. `image` doesn't pin to any version currently, and PistonDevelopers/image#482 already exists for discussing that. I'll take care of that either today or tomorrow, though it's not something that blocks this PR.

CHANGELOG.md
francesca64: @Osspial once you take care of this, we'll be good to merge.

PR #655, opened 2018-09-16
Wayland: use `init_from_env()` to create windows and allow server-sid…

trimental:

`init_from_env()` allows the creation of windows with server-side decorations when appropriate (when the compositor allows and supports it). Some compositors will never support server-side decorations on wayland like gnome's mutter however other compositors like kde's kwin actively support and encourage its usage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 # Unreleased
+- On Wayland, windows will use server-side decorations when appropriate.
 - Added support for F16-F24 keys.
 - Fixed graphical glitches when resizing on Wayland.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 # Unreleased
+- On Wayland, windows will use server-side decorations when appropriate.
```
vberger: I'd rather word it "when available", as otherwise the notion of what "appropriate" means is unclear.

vberger:
LGTM apart from the changelog nit. :+1:

PR #658, opened 2018-09-19
Don't include NUL byte in _NET_WM_NAME

sometoby:
> The contents of the property are not required to be null-terminated;
> any terminating null should not be included in text_prop.nitems.

https://tronche.com/gui/x/xlib/ICC/client-to-window-manager/XmbTextPropertyToTextList.html

- [ ] n/a <strike>Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior</strike>
- [ ] n/a <strike>Created an example program if it would help users understand this functionality</strike>


francesca64:
Thanks! Looks good.

PR #659, opened 2018-09-19
Add X11-specific with_gtk_theme_variant option

sometoby:

This change provides applications that use winit with the ability to set the `_GTK_THEME_VARIANT` property for X11 windows. [Alacritty](https://github.com/jwilm/alacritty) for example could make use of this feature to let users enable dark window decorations specifically for Alacritty without affecting other GNOME apps. The stock GNOME terminal already provides this functionality.

If my understanding of winit's current Wayland support is accurate, implementing this feature for native Wayland isn't yet possible. I'd very happily be corrected on that and amend my changes accordingly.

Since I'm making these changes specifically to enable the above-mentioned feature in Alacritty, I've only tested them on top of the 0.15.1 release that Alacritty depends on — getting it to build with winit 0.17.2 seemed nontrivial at first glance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Fixed graphical glitches when resizing on Wayland.
 - On Windows, fix freezes when performing certain actions after a window resize has been triggered. Reintroduces some visual artifacts when resizing.
 - Updated window manager hints under X11 to v1.5 of [Extended Window Manager Hints](https://specifications.freedesktop.org/wm-spec/wm-spec-1.5.html#idm140200472629520).
+- Added `WindowBuilderExt::with_gtk_theme_variant` to X11-specific `WindowBuilder` functions.
 
 # Version 0.17.2 (2018-08-19)
 
--- a/src/platform/linux/x11/window.rs
+++ b/src/platform/linux/x11/window.rs
@@ -272,6 +272,10 @@ impl UnownedWindow {
                 window.set_window_type(pl_attribs.x11_window_type).queue();
             }
 
+            if let Some(variant) = pl_attribs.gtk_theme_variant {
+                window.set_gtk_theme_variant(variant).queue();
+            }
+
             // set size hints
             {
                 let mut min_dimensions = window_attrs.min_dimensions;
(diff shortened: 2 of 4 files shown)


vberger:
> If my understanding of winit's current Wayland support is accurate, implementing this feature for native Wayland isn't yet possible.

To be honest, I think you can remove the "yet", or at least it cannot be implemented in its current form for wayland.

Reasons for that are:

- the GNOME project [insists that wayland clients draw their own decorations](https://blogs.gnome.org/tbernard/2018/01/26/csd-initiative/) (as opposite to KDE or sway that draw server-side decorations for example), and thus does not draw borders on its wayland clients, only the x11 ones.
- winit does not use GTK to create its window and draw its decorations.

Winit uses [SCTK](https://github.com/smithay/client-toolkit) to draw its decorations on wayland. The toolkit uses server-side decorations if the server supports them, and draws its own if not. Currently the toolkit has a single theme, but if anyone is willing to either contribute new themes or make this theme more customization (by choosing the colors for example), this would certainly be welcome.

src/platform/linux/x11/window.rs
```diff
@@ -457,6 +461,19 @@ impl UnownedWindow {
         )
     }
 
+    fn set_gtk_theme_variant(&self, variant: String) -> util::Flusher {
+        let hint_atom = unsafe { self.xconn.get_atom_unchecked(b"_GTK_THEME_VARIANT\0") };
+        let utf8_atom = unsafe { self.xconn.get_atom_unchecked(b"UTF8_STRING\0") };
+        let variant = CString::new(variant).expect("_GTK_THEME_VARIANT contained null byte");
```
francesca64: `_GTK_THEME_VARIANT` should be in back-ticks here.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 - Fixed graphical glitches when resizing on Wayland.
 - On Windows, fix freezes when performing certain actions after a window resize has been triggered. Reintroduces some visual artifacts when resizing.
 - Updated window manager hints under X11 to v1.5 of [Extended Window Manager Hints](https://specifications.freedesktop.org/wm-spec/wm-spec-1.5.html#idm140200472629520).
+- Added `with_gtk_theme_variant` to X11-specific `WindowBuilder` functions
```
francesca64: It's conventional in the CHANGELOG to fully qualify this, i.e. `WindowBuilderExt::with_gtk_theme_variant`. Ending in a period would also be nice.

francesca64:
Thanks for the PR! The implementation looks solid; I'll test this shortly.

sometoby:
Speedily amended the changes based on your feedback, cheers :)

francesca64:
Looks great, thanks!

PR #668, opened 2018-10-03
x11: compute resize logical size with new dpi

### LINKED ISSUE #663: Resize events with DPI update in winit 0.17
So I've been trying to get the breaking winit/glutin update to work with Alacritty and I've run into issues where changes would work on X11/i3, but I'd run into issues with Wayland. I've tested out a couple of different ways to handle this, but I think I've run into an issue with how the events are ordered.

This all heavily depends on the order of events coming in, I wasn't able to reproduce it when logging all events for example. So I'm trying to outline the issue, but I'm not sure if everything I'm saying is comeletely correct.

Currently it seems like winit sends Resize events first and sends DPR change events after that. On i3 I'm sometimes running into the scenario that I'm just receiving a DPR change event when moving a window between screens with different DPRs, but the window got scaled correctly already, so this DPR event is just useful for converting the next resize event to physical sizes. However on Sway when starting the window on a 2.0 DPI screen, the window seems to be first resized to a smaller version with 1.0 DPR and then I receive a 2.0 DPR event which requires that I actually scale the previous viewport with the new DPR.

These two behaviors seem to be in conflict with each other. One requiring that a standalone DPR change is ignored, the other requiring that the viewport should be resized.

The common behavior when both DPR and Resize events are received at the same time (polling until the queue is empty), seems to be that the DPR change event should be ignored. So it doesn't seem like the behavior of i3 is incorrect, if anything this should be related to Wayland?

Even though it might not be useful, here's the debug output I've generated which lead me to this error: 
[bug_logs.txt](https://github.com/tomaka/winit/files/2416354/bug_logs.txt)

If this is actually an issue with winit and not the way I've been making use of the events, there are several ways this could be changed. I think always sending DPR events before resize events would resolve this issue, even though it could potentially still cause timing issues. Always sending the DPR with a resize event would also allow converting logical to physical sizes without having to rely on correct event order, however it seems like DPR is not stored internally, so that might cause some overhead. These were just my first two thoughts, I'm sure there are other solutions too.

vberger:
This change does 2 things to the x11 backend:

- change the order of processing of the contents of events so that checking the current monitor and retrieving the dpi is done *before* computing the logical size of the window
- whenever the dpi changes, always queue a `Resized` event with the new, correct logical size
  - given on x11 the physical size is constant by default, a dpi change changes the logical size, so any previous logical size becomes wrong, hence the need for this event

This should hopefully fix #663, but I don't really have the environment to test it right now.

cc @chrisduerr 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@
 - On Windows 10, fix transparency (#260).
 - on macOS, fix modifiers during key repeat.
 - Implemented the `Debug` trait for `Window`, `EventsLoop`, `EventsLoopProxy` and `WindowBuilder`.
+- X11 backend will now always generate a `Resized` event after a DPI change, to ensure consistency
+  of the logical size of the window
+- Some clarifications of the DPI docs
 
 # Version 0.17.2 (2018-08-19)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,11 +14,10 @@
 - On Windows, `Window::set_cursor` now applies immediately instead of requiring specific events to occur first.
 - On Windows, fix `Window::set_maximized`.
 - On Windows 10, fix transparency (#260).
-- on macOS, fix modifiers during key repeat.
+- On macOS, fix modifiers during key repeat.
 - Implemented the `Debug` trait for `Window`, `EventsLoop`, `EventsLoopProxy` and `WindowBuilder`.
-- X11 backend will now always generate a `Resized` event after a DPI change, to ensure consistency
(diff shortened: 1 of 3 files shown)


chrisduerr:
I've linked Alacritty's glutin update PR against this change and was able to successfully use it on both Sway and i3 without any issues.

vberger:
Alright, I've also clarified a little the DPI docs to clarify how to interpret the interaction between the `Resized` and `HiDPIFactorChanged` events.

vberger:
Yes, this seems good to me, thanks!

PR #671, opened 2018-10-09
Upgrade to smithay-client-toolkit 0.4

trimental:
- [x] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users (Just about the decorations change)

Upgrades to smithay-client-toolkit to version 4.0, biggest changes are

* Now uses wayland 0.21 (doesn't use pure rust implementation because of c lib dependency that can hopefully be removed in the future)

* Better 'ConceptFrame' for decorations which looks more visual appealing then 'BasicFrame'

* Window titles will now be displayed in the header of the window

* Fix wayland key repetition when keyboard loses focus

Fixes https://github.com/tomaka/winit/issues/657

### DIFF
--- a/src/platform/linux/wayland/keyboard.rs
+++ b/src/platform/linux/wayland/keyboard.rs
@@ -5,11 +5,14 @@ use sctk::keyboard::{
     self, map_keyboard_auto_with_repeat, Event as KbEvent, KeyRepeatEvent, KeyRepeatKind,
 };
 use sctk::reexports::client::protocol::wl_keyboard;
-use sctk::reexports::client::{NewProxy, Proxy};
+use sctk::reexports::client::Proxy;
+use sctk::reexports::client::protocol::wl_seat;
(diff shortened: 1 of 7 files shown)


src/platform/linux/wayland/keyboard.rs
```diff
@@ -120,45 +127,47 @@ pub fn init_keyboard(
...
+                                        modifiers: ModifiersState::default(),
+                                    },
                                 },
-                            },
-                            wid,
-                        );
+                                wid,
+                            );
+                        }
                     }
```
vberger: We probably  should handle `wl_keyboard::Event::Modifiers` too, to keep symmetry with the SCTK impl.

vberger: Actually, no need, I don't think it's possible that we receive this event if `libxkbcommon` is not available, given it uses its format.

vberger:
For the record, here is how alacritty looks like in weston with the new decorations:

alacritty-new-decorations

Issue #670, opened 2018-10-07
window.set_maximized does nothing on Windows 10


Bzomak:
Using the fullscreen example, toggling decorations and fullscreen works, but there is no noticable effect with the set_maximized method.

Bzomak:
Okay, I've done some playing around, and it does trigger sometimes, but there seems to be some sort of interaction between the fullscreen state that is getting muddled. Time to dive further into the code...

francesca64:
Fixed via #672

PR #672, opened 2018-10-10
Windows: Fix window.set_maximized()

Bzomak:

Should fix issue #670
Tested on my Windows 10 machine, using the fullscreen example.
Restores the correct window.set_maxmized functionality, that appears to have been broken in PR #601


francesca64:
Thanks! Looks great.

Issue #260, opened 2017-08-17
Window transparency doesn't work on Windows 10


sambercx:
The ```DwmEnableBlurBehindWindow``` method for making windows transparent on Windows doesn't seem to work on Windows 10 anymore. On Windows 10 the ```DwmExtendFrameIntoClientArea``` function can be called with the margins set to -1 to make it fully transparent.

EDIT: The new method only works if the window decorations are disabled.

LYP951018:
I'm trying to implement this by `GetProcAddress` in runtime. But should we cache `kernel32` module handle in winit? Otherwise we may need to call `LoadLibrary`every time the method is called.

retep998:
Use `lazy_static` to lazily load `kernel32.dll` and keep the handle to it around and do the same for each of the function pointers. Never let go.

Lisoph:
I will attempt to fix this as part of #568.

subnomo:
I've enabled transparency using layered windows. It works with window decorations, as well. See my patch:

```diff
diff --git a/src/platform/windows/window.rs b/src/platform/windows/window.rs
index 259376f4..a227c090 100644
--- a/src/platform/windows/window.rs
+++ b/src/platform/windows/window.rs
@@ -905,6 +905,10 @@ unsafe fn init(
         ex_style |= winuser::WS_EX_NOREDIRECTIONBITMAP;
     }
 
+    if attributes.transparent {
+        ex_style |= winuser::WS_EX_LAYERED;
+    }
+
     // adjusting the window coordinates using the style
     winuser::AdjustWindowRectEx(&mut rect, style, 0, ex_style);
 
@@ -959,6 +963,11 @@ unsafe fn init(
                                               format!("{}", io::Error::last_os_error()))));
         }
 
+        if attributes.transparent {
+            let opacity = 100; // transparency value 0-255
+            winuser::SetLayeredWindowAttributes(handle, 0x0, opacity, winuser::LWA_ALPHA);
+        }
+
         WindowWrapper(handle)
     };
```

PR #675, opened 2018-10-14
Windows: Fix transparency

subnomo:

Fixed transparency in Windows 10 (#260). It also works with window decorations, by using `WS_EX_LAYERED`. Tested with the transparency example on my Windows 10 machine.

### DIFF
--- a/src/platform/windows/window.rs
+++ b/src/platform/windows/window.rs
@@ -14,6 +14,7 @@ use winapi::um::{combaseapi, dwmapi, libloaderapi, winuser};
 use winapi::um::objbase::COINIT_MULTITHREADED;
 use winapi::um::shobjidl_core::{CLSID_TaskbarList, ITaskbarList2};
 use winapi::um::winnt::{LONG, LPCWSTR};
+use winapi::um::wingdi::CreateRectRgn;
 
 use {
     CreationError,
@@ -894,7 +895,7 @@ unsafe fn init(
             }
         )
     } else {
-        (winuser::WS_EX_APPWINDOW | winuser::WS_EX_WINDOWEDGE,
+        (winuser::WS_EX_APPWINDOW | winuser::WS_EX_WINDOWEDGE | winuser::WS_EX_LAYERED,
             winuser::WS_OVERLAPPEDWINDOW | winuser::WS_CLIPSIBLINGS | winuser::WS_CLIPCHILDREN)
     };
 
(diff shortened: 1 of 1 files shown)


src/platform/windows/window.rs
```diff
@@ -1019,13 +1023,22 @@ unsafe fn init(
     // making the window transparent
     if attributes.transparent && !pl_attribs.no_redirection_bitmap {
         let bb = dwmapi::DWM_BLURBEHIND {
-            dwFlags: 0x1, // FIXME: DWM_BB_ENABLE;
+            dwFlags: dwmapi::DWM_BB_ENABLE | dwmapi::DWM_BB_BLURREGION,
             fEnable: 1,
-            hRgnBlur: ptr::null_mut(),
+            hRgnBlur: CreateRectRgn(0, 0, -1, -1), // makes the window transparent
```
francesca64: Do we need to manually free this region using `DeleteObject`?

src/platform/windows/window.rs
```diff
@@ -1019,13 +1023,22 @@ unsafe fn init(
...
 
         dwmapi::DwmEnableBlurBehindWindow(real_window.0, &bb);
+
+        if attributes.decorations {
+            // HACK: When opaque (opacity 255), there is a trail whenever
+            // the transparent window is moved. By reducing it to 254,
+            // the window is rendered properly.
+            let opacity = 254;
+
+            winuser::SetLayeredWindowAttributes(real_window.0, 0x0030c100, opacity, winuser::LWA_ALPHA);
```
francesca64: The meaning of the `0x0030c100` constant should be documented.

src/platform/windows/window.rs
```diff
@@ -14,6 +14,7 @@ use winapi::um::{combaseapi, dwmapi, libloaderapi, winuser};
 use winapi::um::objbase::COINIT_MULTITHREADED;
 use winapi::um::shobjidl_core::{CLSID_TaskbarList, ITaskbarList2};
 use winapi::um::winnt::{LONG, LPCWSTR};
+use winapi::um::wingdi::CreateRectRgn;
```
francesca64: Since `wingdi` is alphabetically before `winnt`, this import should be before it.

src/platform/windows/window.rs
```diff
@@ -1019,13 +1023,22 @@ unsafe fn init(
     // making the window transparent
     if attributes.transparent && !pl_attribs.no_redirection_bitmap {
         let bb = dwmapi::DWM_BLURBEHIND {
-            dwFlags: 0x1, // FIXME: DWM_BB_ENABLE;
+            dwFlags: dwmapi::DWM_BB_ENABLE | dwmapi::DWM_BB_BLURREGION,
```
francesca64: These constants appear to be absent from winapi 0.3.5. It would be great if you'd change Cargo.toml to use winapi 0.3.6, since then people won't have to manually run `cargo update`.

src/platform/windows/window.rs
subnomo: Yep, good catch.

subnomo:
@francesca64 Done!

PR #676, opened 2018-10-15
Provide current modifiers state with pointer events on wayland

trimental:

Tracks modifiers state from keyboards and then uses that to provide the modifiers state for pointer events

Fixes https://github.com/tomaka/winit/issues/653

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- On Wayland, pointer events will now provide the current modifiers state
 - On Wayland, titles will now be displayed in the window header decoration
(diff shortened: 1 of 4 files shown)


vberger:
Given the state tracker is internal to the seat/pointer/keyboard impls, couldn't it be a `Rc<Cell<ModifiersState>>` rather than a `Arc<Mutex<ModifiersState>>` ?

trimental:
For the record, the modifiers tracker must be thread safe to send it to `map_keyboard_auto_with_repeat`

francesca64:
Thanks for the PR!

PR #679, opened 2018-10-20
version 0.17.3

sodiumjoe:
- ~[ ] Tested on all platforms changed~
- ~[ ] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users~
- ~[ ] Created an example program if it would help users understand this functionality~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+# Version 0.17.3 (2018-10-19)
+
 - On Wayland, pointer events will now provide the current modifiers state.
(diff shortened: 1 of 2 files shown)


sodiumjoe:
I'm not really sure how winit handles semver, so I just opened https://github.com/tomaka/winit/pull/680 as well. I'll let @francesca64 go with whichever one is correct and close the other one 😄

vberger:
Hmm, the changelog entry for #641 should have marked the change as breaking.

Winit has respected semver up to now, and I don't think there are plans for this changing, so the path forward `0.18.0`.

PR #680, opened 2018-10-20
version 0.18.0

sodiumjoe:
- ~[ ] Tested on all platforms changed~
- ~[ ] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users~
- ~[ ] Created an example program if it would help users understand this functionality~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 - On Wayland, key repetition is now ended when keyboard loses focus.
 - On Wayland, windows will now use more stylish and modern client side decorations.
 - On Wayland, windows will use server-side decorations when available.
-- Added support for F16-F24 keys.
+- **Breaking:** Added support for F16-F24 keys (variants were added to the `VirtualKeyCode` enum).
 - Fixed graphical glitches when resizing on Wayland.
 - On Windows, fix freezes when performing certain actions after a window resize has been triggered. Reintroduces some visual artifacts when resizing.
 - Updated window manager hints under X11 to v1.5 of [Extended Window Manager Hints](https://specifications.freedesktop.org/wm-spec/wm-spec-1.5.html#idm140200472629520).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 
-# Version 0.18.0 (2018-10-26)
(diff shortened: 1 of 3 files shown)


vberger:
I've added a note in the changelog precising that F16-F24 was a breaking change. That's unfortunate, I was hoping we could have this release without breaking changes.

While we are at it, I'll check if there are other breaking PRs soon-to-be-merged that we might want to merge before the release.

sodiumjoe:
fixed readme

sodiumjoe:
rebased master

francesca64:
Sorry I've been silent on this! There are two concerns that need to be addressed:
1. #683 broke Rust 1.24.1 compatibility, which will be resolved by PistonDevelopers/image#828.
2. #685 could be considered an upcoming breaking change, since it requires `cocoa` 0.18. We're currently using `cocoa` 0.17, and a type from that is part of our public API [in a very minor way](https://github.com/tomaka/winit/blob/214e157e5df4e3853f4c83269b04b88146d623d8/src/os/macos.rs#L50). I doubt that impl has ever been used by consumers (or was ever intended to be), so it's probably a good idea to just prevent it from being exposed.

Osspial:
Could we also merge #652 for 0.18? I think all issues with that have been resolved.

rukai:
@sodiumjoe can you please set the version of image crate to `0.20.1`. This will ensure users of winit have the correct version of image without having to run `cargo update`.

sodiumjoe:
ok, bumped `image` dep.

I'm on macos, but I'm not sure what the second thing is. Just delete that `impl` and update the changelog?

francesca64:
That impl is used internally here: https://github.com/tomaka/winit/blob/0fca8e8cb57f0b4b03bc12b1b17bec5855e7ec7e/src/platform/macos/window.rs#L720

So we either need to prevent the impl from being exposed (if that's possible?) or just change it into a regular `pub(crate)` method. Either way, it should live in the `window` module, since having private stuff in the `os` modules is always a bit confusing.

vberger:
> So we either need to prevent the impl from being exposed (if that's possible?)

I don't think that is possible, unless the type which has the impl is private, of course.

sodiumjoe:
rebased master again, i think that was all the blockers

Issue #678, opened 2018-10-17
Unit testing: Panic messages should not contain control characters


nausicaea:
When writing unit tests for a library that depends on `winit`, I discovered that under certain circumstances, `winit` fails to initialize and panics with a message that is built from a raw string (see [/src/platform/linux/mod.rs](https://github.com/tomaka/winit/blob/master/src/platform/linux/mod.rs#L433-L436)).

This makes cases that trigger such a panic hard to test for (or at the least prone to errors), because the message contains newlines and multiple successive spaces. Depending on how the source of `winit` is edited, those spaces could at some point become `\t` characters, which would make my tests fail, even though functionally, nothing had changed.

I know that this could also be seen as a cosmetic issue, but in my humble opinion, it would be nice to have single-line panic messages that include as few control characters as possible.

Regardless, thank you for this invaluable library!

francesca64:
Oh, I'm sorry! I used the formatting there to try to be fancy, but if it's an inconvenience at all I have no problem with it being removed. I'd gladly accept a PR to that effect; otherwise, I'll change it myself when I get a chance.

nausicaea:
Uhm, I'm a bit unsure on how to proceed. I've never contributed to a project other than my own and I wanted to ask whether there is a procedure that I'm expected to follow when creating a pull request. Do you have any conventions in place (branching, updating the changelog, testing, checklists?)?

francesca64:
(Sorry that this isn't documented anywhere!)
- Creating a separate branch is nice, just because it makes it a bit easier for me to checkout things.
- A changelog entry would be quite nice, and the style/format should match what's already in the changelog.
- You don't really need to worry about testing, at least in this case. If it works correctly for you, works correctly for me, and passes CI, then you're all good.
- The [PR template](https://github.com/tomaka/winit/blob/master/.github/PULL_REQUEST_TEMPLATE.md) includes a little checklist, though only the first two items would apply to this kind of revision.

francesca64:
Fixed via #681

PR #681, opened 2018-10-21
Unit testing: Panic messages should not contain control characters

nausicaea:
This PR refers to issue #678 .

The relevant panic message should primarily appear on linux systems without either Wayland or X11. This is true for example on the Windows Subsystem for Linux, on which I tested the changes. Please let me know if I should test on additional platforms.

The binary used to trigger the panic is the example `window`. After the change, the panic message looks as follows:

```bash
$ cargo run --example window
thread 'main' panicked at 'Failed to initialize any backend! Wayland status: XdgRuntimeDirNotSet X11 status: XOpenDisplayFailed', src/platform/linux/mod.rs:437:9
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

- [ ] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior (not applicable)
- [ ] Created an example program if it would help users understand this functionality (not applicable)


francesca64:
Thanks for the PR! Everything looks good.

PR #683, opened 2018-10-22
update to image 0.20

rukai:

Tested on x11/kde and windows.
This should be merged before the release of 0.18.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- **Breaking:** `image` crate upgraded to 0.20. This is exposed as part of the `icon_loading` API.
 - On Wayland, pointer events will now provide the current modifiers state.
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 
+- Image crate upgraded to 0.20, this is exposed as part of the icon_loading api **breaking**
```
francesca64:
```suggestion
- **Breaking:** `image` crate upgraded to 0.20. This is exposed as part of the `icon_loading` API.
```

Thanks for the PR! I just have some tiny nitpicks to make this more consistent with other entries.

rukai:
Just noticed there was a warning triggered when building the example without the `icon_loading` feature, so I fixed that as well by moving the `use winit::Icon` into `fn main() { }`

francesca64:
Alright, awesome!

PR #685, opened 2018-10-25
Make views explicitly layer-backed on macOS Mojave.

pcwalton:
This is an alternate fix for tomaka/glutin#1069.

On Mojave, views automatically become layer-backed shortly after being added to a window. Changing the layer-backedness of a view breaks the association between the view and its associated OpenGL context. To work around this, on Mojave we explicitly make the view layer-backed up front so that AppKit doesn't do it itself and break the association with its context.

This was breaking the `window` example in `glutin`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,9 +31,9 @@ objc = "0.2.3"
 
 [target.'cfg(target_os = "macos")'.dependencies]
 objc = "0.2.3"
-cocoa = "0.17"
+cocoa = "0.18.4"
 core-foundation = "0.6"
-core-graphics = "0.16"
+core-graphics = "0.17.3"
 
 [target.'cfg(target_os = "windows")'.dependencies.winapi]
 version = "0.3.6"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 - Add optional `serde` feature with implementations of `Serialize`/`Deserialize` for DPI types and various event types.
 - Add `PartialEq`, `Eq`, and `Hash` implementations on public types that could have them but were missing them.
 - On X11, drag-and-drop receiving an unsupported drop type can no longer cause the WM to freeze.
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -30,9 +30,15 @@ objc = "0.2.3"
...
 core-foundation = "0.6"
-core-graphics = "0.16"
+
+[target.'cfg(target_os = "macos")'.dependencies.core-graphics]
+git = "https://github.com/pcwalton/rust-core-foundation"
+branch = "new-appkit-versions"
+
+[target.'cfg(target_os = "macos")'.dependencies.cocoa]
+git = "https://github.com/pcwalton/rust-core-foundation"
+branch = "new-appkit-versions"
```
mitchmindtree: Just for others' reference, it looks like the PR that these git branches are pending on is [here](https://github.com/servo/core-foundation-rs/pull/266).

francesca64:
Thanks for the PR; the code looks good and I love the comments! I don't have Mojave to test this against, but I'm nonetheless fine with merging once we have a `cocoa` release to use.

In the mean time, I think it would be great to add a CHANGELOG entry for this.

gnzlbg:
I think there is a new version of `cocoa` already (0.18.3).

francesca64:
@gnzlbg the latest version I see on crates.io is 0.18.1: https://crates.io/crates/cocoa

gnzlbg:
@francesca64 probably has not been published yet: https://github.com/servo/core-foundation-rs/blob/master/cocoa/Cargo.toml#L7

atouchet:
cocoa 0.18.3 has been published now.

atouchet:
@pcwalton could you update this now that the new version of cocoa has been published?

pcwalton:
Rebased and updated to use crates.io versions of `core-foundation`  and `cocoa`.

PR #690, opened 2018-11-05
Remove From<NSApplicationActivationPolicy> impl from ActivationPolicy

Osspial:
- [x] ~~Created an example program if it would help users understand this functionality~~

This is blocking `0.18` and a PR didn't seem to be coming, so I threw this together. With cross-compilation Rust compiles it fine, but since I don't have convenient access to a MacOS machine it's untested.

cc #680


vberger:
The change seems clear enough and the build passes, LGTM.

francesca64:
Thanks for taking the initiative; this looks solid!

Issue #579, opened 2018-06-27
Add Window::get_fullscreen() -> bool


rukai:
`set_fullscreen` takes an `Option<WindowId>`.
However we dont need to make `get_fullscreen()` return an `Option<WindowId>` because we can use `get_current_monitor() -> WindowId` instead.

The window manager (at least on X11) can set the fullscreen state so just remembering the state in winit will not be enough.

francesca64:
The way I'd want to address this is similar to #208, since:

- Maximization events are necessary and it would make sense for the fullscreen API to follow the same pattern.
- There's also a fullscreen button on macOS.
- I tend to favor events over getters, since people presumably usually only want to know about changes, and only as often as they happen.

So I'd go with `WindowEvent::Fullscreen(Option<MonitorId>)`. However, I'm not opposed to having the getter as well, though I think it really should return `Option<MonitorId>`, for consistency with `set_fullscreen`. That also makes it easier to do dumb stuff with combinators.

rukai:
Thanks, that all makes sense.

francesca64:
Are there any WMs you recommend for testing this?

rukai:
Not sure if your asking me directly? I only use i3.

Issue #473, opened 2018-04-19
Added support for fullscreen options and proper fullscreen behaviors in macos


edwin0cheng:
This is  a tracking issue for improvements of PR https://github.com/tomaka/winit/pull/465 :

- Currently disregards whatever monitor is passed for setting fullscreen, always using whichever monitor has the keyboard focus. `enterFullScreenMode` should work for this, but it appears to be absent from core-foundation.
- Similarly, get_current_monitor always returns the currently focused monitor, rather than the monitor the window is on. `CGGetDisplaysWithRect` could be used to find the monitor that the majority of the window is on. But it is absent from core-foundation too. 
- We should let user to control the FullScreen Options by expose it in `WindowBuilderExt`. 

Here is the related PR for in core-foundation:

- [x] PR for CGGetDisplaysWithRect https://github.com/servo/core-foundation-rs/pull/207
- [x] PR for Fullscreen options: https://github.com/servo/core-foundation-rs/pull/204
- [x] PR for enterFullScreenMode related: https://github.com/servo/core-foundation-rs/pull/208

And related discussion: https://github.com/tomaka/winit/issues/460#issuecomment-382212377

francesca64:
Is anything else blocking this? If you don't have time at the moment, I'd be happy to implement this.

francesca64:
I've only addressed `get_current_monitor`. #521 

`enterFullScreenMode` works pretty differently from `toggleFullScreen`. It pops our view out into a "private" fullscreen window, which not only means our window delegate's `windowDidEnterFullScreen`/etc. isn't triggered, but we also don't get events at all. If you fullscreen to a different screen than the window is on, then the window remains visible where it was in addition to having the fullscreen view.

Getting that to work correctly seems like it will be involved, so I'm not presently up to pursuing this further.

edwin0cheng:
Thanks for your help !!

- The main problem is, i found that i do not have another screen to let me implement/test these features.   Have to wait until i come back home next month. 😅

[...]

PR #692, opened 2018-11-08
feat: add macos simple fullscreen

acheronfail:
This change adds a new "simple" fullscreen mode for macOS.
This is usually a highly requested feature on macOS since many people do not like having to depend on a separate space for fullscreen applications.

Many applications provide a simple fullscreen mode like this (VLC, iTerm, etc).

See the demo here: https://www.dropbox.com/s/0c9aawble06xq2t/simple_fullscreen.mov?dl=0

DISCLAIMER: I'm mainly adding this so I can add this functionality to alacritty (https://github.com/jwilm/alacritty/issues/34).

---


francesca64:
Thanks for the PR! I haven't had a chance to review this in detail yet, but two things for now:
- This method should be exposed via `WindowExt` instead: https://github.com/tomaka/winit/blob/master/src/os/macos.rs#L7
- What's your reason for favoring this approach over using `enterFullScreenMode` (as discussed in #473)?

acheronfail:
> This method should be exposed via `WindowExt` instead: /src/os/macos.rs@master#L7

I'll get onto that. 👍 

> What's your reason for favoring this approach over using enterFullScreenMode (as discussed in #473)?

[`enterFullScreenMode`](https://developer.apple.com/documentation/appkit/nsview/1483780-enterfullscreenmode?language=objc) is more aimed towards games or kiosks. Basically, macOS has three distict fullscreen modes:

* Standard fullscreen (happens when you click the green fullscreen button)
    * This moves the window to its own space
* [`enterFullScreenMode`](https://developer.apple.com/documentation/appkit/nsview/1483780-enterfullscreenmode?language=objc)
    * This takes over the monitor, so _no spaces or other applications may be used at the same time_ (think of how fullscreen mode works in kiosk applications)
* "Simple Fullscreen"
    * The window just fills the screen completely (hiding the dock and menu bar)
    * This used to be how all fullscreen windows worked pre-macOS Lion
    * This is a popular and widely requested feature, that many apps have implemented manually since Apple removed the functionality.
        * Some apps are: VLC, iTerm2, [The `electron` framework](https://electronjs.org/docs/api/browser-window#winsetsimplefullscreenflag-macos), and various other applications
    * This means you can still use other applications, spaces, and all other functionality, while having your application in fullscreen mode.

francesca64:
Sorry I've taken so long! I have access to a better macOS device now, so things should be faster in the future.

I definitely see that value in this functionality. I've noticed that your implementation doesn't adhere to winit's expected fullscreen behavior: in fullscreen mode, toggling decorations/maximization/etc. shouldn't take effect until after fullscreen has been exited (but the state is still tracked internally).

examples/simple_fullscreen.rs
```diff
@@ -0,0 +1,61 @@
+extern crate winit;
+
+use winit::{ControlFlow, Event, WindowEvent};
```
francesca64: Putting this inside of `main` will prevent warnings from being emitted outside of macOS, which would be nice.

Though, instead of having a standalone `simple_fullscreen` example, it could be cool to offer it as an option on `fullscreen` that's only prompted for on macOS.

foo
```diff
@@ -0,0 +1 @@
+asd
```
francesca64: Whoops!

src/os/macos.rs
```diff
@@ -22,6 +22,13 @@ pub trait WindowExt {
     /// - `false`: the dock icon will only bounce once.
     /// - `true`: the dock icon will bounce until the application is focused.
     fn request_user_attention(&self, is_critical: bool);
+
+    /// Toggles a fullscreen mode that doesn't require a new macOS space.
+    ///
+    /// This is how fullscreen used to work on macOS in versions before Lion.
+    /// And allows the user to have a fullscreen window without using another
+    /// space or taking control over the entire monitor.
+    fn set_simple_fullscreen(&self, enable: bool);
```
francesca64: We usually try to give parameters specific names, i.e. `fullscreen` instead of `enable`.

foo
acheronfail: My bad! 🤦‍♂️

acheronfail:
> I've noticed that your implementation doesn't adhere to winit's expected fullscreen behavior: in fullscreen mode, toggling decorations/maximization/etc. shouldn't take effect until after fullscreen has been exited (but the state is still tracked internally).

@francesca64 Do you mean that if the user tries to maximize the window while it's fullscreen, nothing should happen _until_ fullscreen is exited?

Something like:

```
normal window (640x480)
set_simple_fullscreen(true)
maximize (does nothing since window is currently fullscreened)
set_simple_fullscreen(false)
window is now maximized
```

?

acheronfail:
Actually, I see what you mean. If I enable the simple fullscreen, the decorations can still be toggled. I'll fix that up.

* [ ] ensure decorations and maximizing does not affect screen while in fullscreen

src/platform/macos/window.rs
```diff
@@ -94,22 +96,30 @@ impl DelegateState {
...
+            .unwrap_or_else(|| self.window.styleMask());
+        if resizable {
+            base_mask | NSWindowStyleMask::NSResizableWindowMask
+        } else {
+            base_mask & !NSWindowStyleMask::NSResizableWindowMask
+        }
+    }
+
+    fn saved_standard_frame(&self) -> NSRect {
+        self.standard_frame.get().unwrap_or(NSRect::new(
```
francesca64:
```suggestion
        self.standard_frame.get().unwrap_or_else(|| NSRect::new(
```
`unwrap_or` will unconditionally construct `NSRect`: https://doc.rust-lang.org/core/option/enum.Option.html#method.unwrap_or

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Unreleased
 
 - On X11, fixed panic caused by dropping the window before running the event loop.
+- on macOS, added a new simple fullscreen mode which does not require a separate space
```
francesca64: It would be awesome if this mentioned the method name somewhere (`WindowExt::set_simple_fullscreen`)

francesca64:
This looks solid! It's definitely consistent with `set_fullscreen` now.

I've noticed that the *existing* macOS fullscreen behavior (wrt decorations and maximization toggling) isn't consistent with other platforms, **but** that's not anything you have to worry about... I'm just writing this as a reminder to myself.

Anyway, I know I've kept you waiting for a long time already, but as per [our new policies](https://github.com/tomaka/winit/blob/master/CONTRIBUTING.md#making-a-pull-request) we still need one more approval.

mtak-:
@francesca64 I don't mind at all!

@acheronfail There's a few differences between this and the existing fullscreen behavior.

- Calling `set_simple_fullscreen(true)` twice gets the window stuck in fullscreen mode AFAICT:
```rs
window.set_simple_fullscreen(true);
window.set_simple_fullscreen(true);
window.set_simple_fullscreen(false); // still fullscreen
```
- Calling `set_simple_fullscreen(false)` on a _not_ fullscreen window moves it.
- In the example, maximizing the window by clicking the green button, then pressing `F` gets the window stuck in a weird state. I'm guessing the OS fullscreen doesn't interact well with `simple_fullscreen`?

acheronfail:
@mtak- nice finds!

> Calling set_simple_fullscreen(true) twice gets the window stuck in fullscreen mode AFAICT

You were right, I've just fixed this. Now it won't try to fullscreen if it's already fullscreen (and it won't try to "un"fullscreen if it's not already, etc).

> Calling set_simple_fullscreen(false) on a not fullscreen window moves it

This is fixed in the latest commit. 🙂 

> In the example, maximizing the window by clicking the green button, then pressing F gets the window stuck in a weird state. I'm guessing the OS fullscreen doesn't interact well with simple_fullscreen?

Yep, the same fix has solved this issue as well. Although the main problem with that was the example was just using a toggle `is_fullscreen = !is_fullscreen` and the above caused that to get out of sync.

The change I made also gives it a return type to indicate whether the requested transition was successful:

```rust
```

You can see where in the example this is used.

acheronfail:
I've just found another edge-case, I'll push another commit soon!

acheronfail:
@mtak- @francesca64 Just one question:

At the moment this behaviour occurs:

```rust
window.set_simple_fullscreen(true);   // simple fullscreen
window.set_fullscreen(None); // this does nothing
```

Using `set_fullscreen(None)` (which usually disables the native fullscreen) will do nothing, if the screen is already in simple fullscreen.

The inverse also performs the same:

```rust
window.set_fullscreen(Some(monitor)); // native fullscreen
window.set_simple_fullscreen(false);   // this does nothing
```

**Question**: Is this the behaviour we want? Or does it make sense to you that `set_fullscreen` should also disable simple fullscreen?

mtak-:
@acheronfail Thanks for the extremely speedy response!

This transition is a corner case, but still a bit wonky:
```
fullscreen true kind simple // fullscreens
fullscreen true kind native // fullscreens - probably should do nothing
fullscreen false kind simple // does nothing
fullscreen false kind native // unfullscreens, but leaves the window in an odd state
```
I'm guessing that the existing fullscreen code should be checking for simple fullscreen.

> Question: Is this the behaviour we want? Or does it make sense to you that set_fullscreen should also disable simple fullscreen?

I agree with the do nothing approach on `false`/`None` if that _mode_ is not currently active.

This is the way I imagine the two modes interacting.

native | simple | supported | valid transitions
--- | --- | --- | ---
true | true | no | N/A
true | false | yes | `set_fullscreen(None)`/`set_fullscreen(Some(other))`
false | true | yes | `set_simple_fullscreen(false)`
false | false | yes | `set_fullscreen(Some(..))` `set_simple_fullscreen(true)`

With all invalid transitions being ignored/do nothing.

@francesca64 @acheronfail Does that table seem right to you?

acheronfail:
@mtak- That seems right to me.

I just updated it and verified that it behaves the same as the above table.

The only question I have, is whether or not users should be able to tell if the window is in fullscreen (or simple fullscreen) or not?
Perhaps something like:

```rust
fn get_fullscreen() -> Option<MonitorId>;
fn get_simple_fullscreen() -> bool
```

That would be the only thing that I can think of.
Of course, they can always keep track of the window state itself... But we already have that information in the `DelegateState`. Just a though. 🤷‍♂️

francesca64:
> The only question I have, is whether or not users should be able to tell if the window is in fullscreen (or simple fullscreen) or not?

That was proposed in #579, which it looks like I forgot to ever follow up on.

PR #674, opened 2018-10-13
[META] Add a CONTRIBUTING.md

vberger:
This PR introduces a `CONTRIBUTING.md` file based on the discussion in #669. This is for now very WIP and intended to serve as a basis for discussion, like an RFC.

Regarding the most fundamental still-open questions:

- who exactly is a "maintainer" of winit, who has commit right and is willing to dedicate time to winit?
- For now I've left #252 as the "official list of features", but we might want to make it more formal maybe?

Rendered version of the file: https://github.com/vberger/winit/blob/contributing/CONTRIBUTING.md

Issue #642, opened 2018-08-28
How are double clicks handled?


portstrom:
I searched the documentation and found nothing about double clicks. I think there should either be some state for this in the mouse input event or at least a function to get the double click time from the system.

jwilm:
@portstrom this is currently left up for the application to solve. In Alacritty, we record a timestamp whenever a click is received, and using a configurable timeout, watch for double and triple clicks.

fasihrana:
Should be able to read system setting i think but would that be in the winit scope?

Issue #624, opened 2018-08-09
Joystick events


LaylConway:
The recommended method of getting joystick events on Windows, for devices that do not fall under the XInput API, is to use the raw input API (WM_INPUT). I saw that winit already does this for mouse & keyboard events, but not yet for joysticks.

Since I've got a need for this in my own project, I hope to soon give a shot at implementing this on Windows, but if anyone else who is more familiar with this API can implement it faster than I could, that would be appreciated.

francesca64:
Funnily enough, I implemented both this and version-agnostic XInput support a while ago... https://github.com/francesca64/winit/tree/win-joy

I just never merged it, since I was concerned about breaking the API later on, and it's missing some special-casing for rumble/etc. I expected to iron out the design more by implementing joystick support on Linux and macOS, but never got around to it.

So, I guess I should rebase it and make a PR.

Osspial:
@francesca64 I'd be up for taking over that PR, rebasing it against the event loop 2.0 code, and getting it ready to release on Windows, if you're okay with it.

francesca64:
@Osspial absolutely! Thanks a bunch.

PR #695, opened 2018-11-10
Winit Features and Scope

Osspial:
[Rendered](https://github.com/Osspial/winit/blob/scope/FEATURES.md)

Adds `FEATURES.md` in place of `SCOPE.md`, as discussed in #674. Subsumes #252. I've replaced the tier 1/2/3 structure with a new Core, Platform, and Usability tier structure, as I think that aligns more with what we're trying to accomplish.

Discussion points highlighted in the document:
* Do we want to support WASM without Emscripten?
* I've introduced the concept of upgrading features from Platform to Core, as some features (window resize increments, child windows) may end up becoming core. Do we want to have this be an official process as I've described?
* Do we want to support child windows?
* Do we want to support gamepad and joystick events?
* Do we want to support device movement (gyroscope and joystick) events?
* The *`1.0` and stability* section is based on my understanding of our end goal for Winit. Is is accurate?
* Some features in the core matrix are marked as partially complete, but don't have an associated issue listed. Do such issues exist?

### DIFF
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -118,7 +118,7 @@ Tables detailing feature compatibility across platforms can be found in the wiki
 ## Usability
 * `serde`: Enables serialization/deserialization of certain types with Serde. (Maintainer: @Osspial)
 
-# Compatibility Matrix - Move to wiki on merge
+# Compatibility Matrix
 
 Each section includes a collapsed description of the features it lists.
 
@@ -131,22 +131,22 @@ Legend:
 - ❓: Unknown status
 
 ## Windowing
-|Feature                          |Windows|MacOS |Linux x11|Linux Wayland|Android|iOS    |Emscripten|
-|-------------------------------- | ----- | ---- | ------- | ----------- | ----- | ----- | -------- |
-|Window initialization            |✔️    |✔️    |▢#5      |✔️          |▢#33   |▢#33  |❓        |
-|Providing pointer to init OpenGL |✔️    |✔️    |✔️       |✔️          |✔️     |✔️    |❓        |
-|Providing pointer to init Vulkan |✔️    |✔️    |✔️       |✔️          |✔️     |❓     |**N/A**   |
-|Window decorations               |✔️    |✔️    |✔️       |▢#306       |**N/A**|**N/A**|**N/A**   |
-|Window decorations toggle        |✔️    |✔️    |✔️       |✔️          |**N/A**|**N/A**|**N/A**   |
-|Window resizing                  |✔️    |▢#219 |✔️       |▢#306       |**N/A**|**N/A**|❓        |
-|Window resize increments         |❌    |❌    |❌       |❌          |❌    |❌     |❌        |
-|Window transparency              |✔️    |✔️    |✔️       |✔️          |**N/A**|**N/A**|**N/A**   |
-|Window maximization              |✔️    |✔️    |✔️       |✔️          |**N/A**|**N/A**|**N/A**   |
-|Window maximization toggle       |✔️    |✔️    |✔️       |✔️          |**N/A**|**N/A**|**N/A**   |
-|Fullscreen                       |✔️    |✔️    |✔️       |✔️          |**N/A**|**N/A**|❌        |
-|Fullscreen toggle                |✔️    |✔️    |✔️       |✔️          |**N/A**|**N/A**|❌        |
-|HiDPI support                    |✔️    |✔️    |✔️       |✔️          |▢#721 |✔️    |✔️         |
-|Popup windows                    |❌    |❌    |❌       |❌          |❌    |❌     |❌        |
+|Feature                          |Windows|MacOS   |Linux x11   |Linux Wayland  |Android|iOS    |Emscripten|
+|-------------------------------- | ----- | ----   | -------    | -----------   | ----- | ----- | -------- |
+|Window initialization            |✔️     |✔️     |▢[#5]      |✔️             |▢[#33]|▢[#33] |❓        |
+|Providing pointer to init OpenGL |✔️     |✔️     |✔️         |✔️             |✔️     |✔️    |❓        |
+|Providing pointer to init Vulkan |✔️     |✔️     |✔️         |✔️             |✔️     |❓     |**N/A**   |
+|Window decorations               |✔️     |✔️     |✔️         |▢[#306]        |**N/A**|**N/A**|**N/A**   |
+|Window decorations toggle        |✔️     |✔️     |✔️         |✔️             |**N/A**|**N/A**|**N/A**   |
+|Window resizing                  |✔️     |▢[#219]|✔️         |▢[#306]        |**N/A**|**N/A**|❓        |
+|Window resize increments         |❌     |❌     |❌         |❌             |❌    |❌     |❌        |
+|Window transparency              |✔️     |✔️     |✔️         |✔️             |**N/A**|**N/A**|**N/A**   |
+|Window maximization              |✔️     |✔️     |✔️         |✔️             |**N/A**|**N/A**|**N/A**   |
+|Window maximization toggle       |✔️     |✔️     |✔️         |✔️             |**N/A**|**N/A**|**N/A**   |
+|Fullscreen                       |✔️     |✔️     |✔️         |✔️             |**N/A**|**N/A**|❌        |
+|Fullscreen toggle                |✔️     |✔️     |✔️         |✔️             |**N/A**|**N/A**|❌        |
+|HiDPI support                    |✔️     |✔️     |✔️         |✔️             |▢[#721]|✔️    |✔️         |
+|Popup windows                    |❌     |❌     |❌         |❌             |❌    |❌     |❌        |
 
 ## System information
 |Feature      |Windows|MacOS |Linux x11|Linux Wayland|Android|iOS    |Emscripten|
@@ -154,25 +154,41 @@ Legend:
 |Monitor list |✔️    |✔️    |✔️       |✔️          |**N/A**|**N/A**|**N/A**   |
 
 ## Input handling
-|Feature                 |Windows|MacOS |Linux x11|Linux Wayland|Android|iOS    |Emscripten|
-|----------------------- | ----- | ---- | ------- | ----------- | ----- | ----- | -------- |
-|Mouse events            |✔️    |▢#63  |✔️       |✔️          |**N/A**|**N/A**|✔️       |
-|Mouse set location      |✔️    |✔️    |✔️       |❓           |**N/A**|**N/A**|**N/A**  |
-|Cursor grab             |✔️    |▢#165 |▢#242    |❌#306      |**N/A**|**N/A**|✔️       |
-|Cursor icon             |✔️    |✔️    |✔️       |❌#306      |**N/A**|**N/A**|❌       |
-|Touch events            |✔️    |❌    |✔️       |✔️          |✔️    |✔️     |✔️       |
-|Multitouch              |❓     |❌    |✔️       |✔️          |❓     |❌     |❌       |
-|Keyboard events         |✔️    |✔️    |✔️       |✔️          |❓     |❌     |✔️       |
-|Drag & Drop             |▢#720 |▢#720️ |▢#720️    |❌#306      |**N/A**|**N/A**|❓        |
-|Raw Device Events       |▢#750 |▢#750 |▢#750    |❌          |❌    |❌     |❌       |
-|Gamepad/Joystick events |❌#804|❌    |❌       |❌          |❌    |❌     |❌       |
-|Device movement events  |❓     |❓     |❓       |❓           |❌    |❌     |❌       |
+|Feature                 |Windows   |MacOS   |Linux x11|Linux Wayland|Android|iOS    |Emscripten|
+|----------------------- | -----    | ----   | ------- | ----------- | ----- | ----- | -------- |
+|Mouse events            |✔️       |▢[#63]  |✔️       |✔️          |**N/A**|**N/A**|✔️       |
+|Mouse set location      |✔️       |✔️      |✔️       |❓           |**N/A**|**N/A**|**N/A**  |
+|Cursor grab             |✔️       |▢[#165] |▢[#242]  |❌[#306]    |**N/A**|**N/A**|✔️       |
+|Cursor icon             |✔️       |✔️      |✔️       |❌[#306]    |**N/A**|**N/A**|❌       |
+|Touch events            |✔️       |❌      |✔️       |✔️          |✔️    |✔️     |✔️       |
+|Multitouch              |❓        |❌      |✔️       |✔️          |❓     |❌     |❌       |
+|Keyboard events         |✔️       |✔️      |✔️       |✔️          |❓     |❌     |✔️       |
+|Drag & Drop             |▢[#720]  |▢[#720] |▢[#720]  |❌[#306]    |**N/A**|**N/A**|❓        |
+|Raw Device Events       |▢[#750]  |▢[#750] |▢[#750]  |❌          |❌    |❌     |❌       |
+|Gamepad/Joystick events |❌[#804] |❌      |❌       |❌          |❌    |❌     |❌       |
+|Device movement events  |❓        |❓       |❓       |❓           |❌    |❌     |❌       |
 
 ## Pending API Reworks
 Changes in the API that have been agreed upon but aren't implemented across all platforms.
 
-|Feature                         |Windows|MacOS |Linux x11|Linux Wayland|Android|iOS    |Emscripten|
-|------------------------------  | ----- | ---- | ------- | ----------- | ----- | ----- | -------- |
-|New API for HiDPI (#315 #319)   |✔️    |✔️    |✔️       |✔️          |▢#721  |✔️    |✔️        |
-|Event Loop 2.0 (#459)           |✔️    |❌    |❌       |✔️          |❌     |❌    |❌        |
-|Keyboard Input (#812)           |❌    |❌    |❌       |❌          |❌    |❌     |❌        |
+|Feature                             |Windows|MacOS |Linux x11|Linux Wayland|Android|iOS    |Emscripten|
+|------------------------------      | ----- | ---- | ------- | ----------- | ----- | ----- | -------- |
+|New API for HiDPI ([#315] [#319])   |✔️    |✔️    |✔️       |✔️          |▢[#721]|✔️    |✔️       |
+|Event Loop 2.0 ([#459])             |✔️    |❌    |❌       |✔️          |❌     |❌    |❌       |
+|Keyboard Input ([#812])             |❌    |❌    |❌       |❌          |❌     |❌     |❌      |
+
+[#165]: https://github.com/tomaka/winit/issues/165
+[#219]: https://github.com/tomaka/winit/issues/219
+[#242]: https://github.com/tomaka/winit/issues/242
+[#306]: https://github.com/tomaka/winit/issues/306
+[#315]: https://github.com/tomaka/winit/issues/315
+[#319]: https://github.com/tomaka/winit/issues/319
+[#33]: https://github.com/tomaka/winit/issues/33
+[#459]: https://github.com/tomaka/winit/issues/459
+[#5]: https://github.com/tomaka/winit/issues/5
+[#63]: https://github.com/tomaka/winit/issues/63
+[#720]: https://github.com/tomaka/winit/issues/720
+[#721]: https://github.com/tomaka/winit/issues/721
+[#750]: https://github.com/tomaka/winit/issues/750
+[#804]: https://github.com/tomaka/winit/issues/804
+[#812]: https://github.com/tomaka/winit/issues/812
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -15,10 +15,9 @@ be used to create both games and applications. It supports the main graphical pl
   - via Emscripten
   - via WASM ***//DISCUSS: DO WE WANT TO SUPPORT THIS?***
 
-Most platforms expose capabilities that cannot be meaningfully transposed to the others. Winit does not
-aim to support every single functionality of every platform, but rather to abstract the set of
-capabilities that is common to all platforms. In this context, APIs exposed in winit can be split into
-different "support levels":
+Most platforms expose capabilities that cannot be meaningfully transposed onto others. Winit does not
+aim to support every single feature of every platform, but rather to abstract over the common features
+available everywhere. In this context, APIs exposed in winit can be split into different "support tiers":
 
 - **Core:** Features that are essential to providing a well-formed abstraction over each platform's
   windowing and input APIs.
@@ -35,9 +34,7 @@ feature and may be asked to support the feature should it break in the future.
 
 Winit ***does not*** directly expose functionality for drawing inside windows or creating native
 menus, but ***does*** commit to providing APIs that higher-level crates can use to implement that
-functionality. When requesting or implementing a new Winit feature, you should consider whether or
-not it's directly related to window creation or input handling. If it isn't, it may be worth creating
-a separate crate that extends Winit's API to add that functionality.
+functionality.
 
 ## `1.0` and stability
 
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -13,7 +13,7 @@ be used to create both games and applications. It supports the main graphical pl
   - Android
 - Web
   - via Emscripten
-  - via WASM ***//DISCUSS: DO WE WANT TO SUPPORT THIS?***
+  - via WASM
 
 Most platforms expose capabilities that cannot be meaningfully transposed onto others. Winit does not
 aim to support every single feature of every platform, but rather to abstract over the common features
@@ -74,7 +74,6 @@ Tables detailing feature compatibility across platforms can be found in the wiki
 - **Popup windows**: Windows can be created relative to the client area of other windows, and parent
   windows can be disabled in favor of popup windows. This feature also guarantees that popup windows
   get drawn above their owner.
-  ***//DISCUSS: SHOULD THIS BE SUPPORTED?***
 
 
 ### System Information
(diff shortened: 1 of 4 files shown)


Osspial:
cc @francesca64 @vberger @mitchmindtree @tomaka.

Relevant issues that may or may not be addressed in the document right now:
- Input handling and filtering
  - Double-clicks: #642 
  - IME: #508 
  - Key repeats: #310, #52
- Device/gamepad events: #624, #119, #212
  - Pen/tablet input: #99 
- Window information:
  - Window creation w/ physical dimensions: #615 
  - Fixed window aspect ratio: #584 
  - `Window::get_fullscreen()` #579 
- Window appearance:
  - Glassy windows: #538 
- `wasm32-unknown-unknown`: #395

Other relevant issues that I'm inclined to close as stuff we won't be changing from its current API state:
- Screensaver toggle: #616 (out of scope)
- Native widgets and menus: #446, #427, #403 (out of scope)
- Windows other messages: #450 (supported by subclassing)
- Configuration files: #304 (supported by serde)

Osspial:
My assorted thoughts on some of those issues, in no particular order:

Generalized gamepad and joystick support is something I think we should support. It's a feature that most people *expect* us to support, falls well within the category of *cross-platform input handling abstraction*, and is implemented within the platform's event loop in several places.

Proper child window support is really tricky to get right outside of Winit, and requires platform-specific code to do properly. On Wayland it's impossible without platform calls, and Winit doesn't expose any way of converting client-space coordinates to screen-space coordinates. Additionally, disabling a window in favor of a child window is a common behavior for desktop applications (see file dialogs and *are you sure you want to close without saving* dialogs) so it makes sense to expose that here.

Gyroscopes and accelerometers are fairly commonly used in mobile games, so I see no reason not to expose them as events.

Double-click and key repeat notifications are both things that could be handled outside of Winit, but neither of them take up a significant amount of API surface area on our end. I think it makes for a better abstraction to track those numbers internally and deliver them in click and keypress events.

MatusT:
I would like to see added what everyone usually forgets about in every windowing API(or any new cross-platform GUI library) and that is split between Win32 and WinRT APIs on Windows. 

I don't know if anyone else is interested but I would love to have a windowing library in Rust with which I could make a UWP native application on Windows(as with WinRT/C++) that I could deploy to Microsoft Store. Bonus: render with vulkan on directx 12 through gfx-hal into the window.

Although all this is stuck on https://github.com/contextfree/winrt-rust/issues/53 which I have no idea how to solve.

FEATURES.md
```diff
@@ -0,0 +1,182 @@
...
+Winit ***does not*** directly expose functionality for drawing inside windows or creating native
+menus, but ***does*** commit to providing APIs that higher-level crates can use to implement that
+functionality.
+
+## `1.0` and stability
+
+When all core features are implemented to the satisfaction of the Winit maintainers, Winit 1.0 will
+be released and the library will enter maintenance mode. For the most part, new core features will not
+be added past this point. New platform features may be accepted and exposed through point releases.
+***//DISCUSS: IS THIS ACCURATE?***
```
vberger: This seems alright to me.

FEATURES.md
```diff
@@ -0,0 +1,182 @@
...
+
+### Tier upgrades
+Some platform features could in theory be exposed across multiple platforms, but have not gone
+through the implementation work necessary to function on all platforms. When one of these features
+gets implemented across all platforms, a PR can be opened to upgrade the feature to a core feature.
+If that gets accepted, the platform-specific functions gets deprecated and become permanently
+exposed through the core, cross-platform API.
+***//DISCUSS: DO WE WANT TO HAVE THIS BE AN OFFICIAL PROCESS, OR NO?***
```
vberger: I am *a priori* not opposed to this kind of approach. I like the idea of implementing things platform-wise, and then unifying them in the core API. The other way would mean exposing APIs full of `unimplemented!()` or in general not working waiting for stuff to be implemented.

FEATURES.md
```diff
@@ -0,0 +1,182 @@
...
+Each section includes a collapsed description of the features it lists.
+
+Legend:
+
+- ✔️: Works as intended
+- ▢: Mostly works but some bugs are known
+- ❌: Missing feature or large bugs making it unusable
+- **N/A**: Not applicable for this platform
+- ❓: Unknown status
+
```
vberger: From a purely practical point of view, do we really want to have the feature parity table in this file? This would make it quite tedious to update, if we want to keep it in sync with the new issues / PRs affecting it.

Would it make sens to have the description of each feature here, and have the actual table as a wiki page for example?

FEATURES.md
```diff
@@ -0,0 +1,182 @@
...
+- **Window transaprency**: Winit allows the creation of windows with a transparent background.
+- **Window maximization**: The windows created by winit can be maximized upon creation.
+- **Window maximization toggle**: The windows created by winit can be maximized and unmaximized after
+  creation.
+- **Fullscreen**: The windows created by winit support being fullscreen.
+- **Fullscreen toggle**: The windows created by winit can be switched to and from fullscreen after
+  creation.
+- **Child windows**: Windows can be created relative to the client area of other windows, and parent
+  windows can be disabled in favor of child windows.
+  ***//DISCUSS: SHOULD THIS BE SUPPORTED?***
```
vberger: I believe this can be in scope for winit. As you said (at least on wayland) creating child window require direct platform interaction in general, and can be difficult to expose without exposing a lot of winit's internals.

FEATURES.md
```diff
@@ -0,0 +1,182 @@
...
+- **Mouse events**: Generating mouse events associated with pointer motion, click, and scrolling events.
+- **Mouse set location**: Forcibly changing the location of the pointer.
+- **Cursor grab**: Locking the cursor so it cannot exit the client area of a window.
+- **Cursor icon**: Changing the cursor icon, or hiding the cursor.
+- **Touch events**: Single-touch events.
+- **Multitouch**: Multi-touch events, including cancellation of a gesture.
+- **Keyboard events**: Properly processing keyboard events using the user-specified keymap and
+  translating keypresses into UTF-8 characters, handling dead keys and IMEs.
+- **Drag & Drop**: Dragging content into winit, detecting when content enters, drops, or if the drop is cancelled.
+  ***//DISCUSS: WINIT SUPPORTS FILE DROPS, BUT NOT TEXT OR IMAGE DROPS***
```
vberger: Yes, this is a point I discussed in an earlier issue / PR. I don't know how this works for other platforms, but in wayland, drag'n'drop is unified for all kind of contents. The way it works is that the application starting the d'n'd and the application receiving it negociate a mime type for the data transfer. Dragging files from a file explorer is just a special case of this.

Can this kind of general behavior be transposed to other platforms?

FEATURES.md
```diff
@@ -0,0 +1,182 @@
...
+- **Clipboard**: Winit supports copy-pasting content to and from winit.
+  ***//DISCUSS: AN ISSUE SHOULD BE OPENED REGARDING HOW THIS API SHOULD WORK***
```
vberger: Agreed. I believe this was already decided that this is in scope for winit.

FEATURES.md
```diff
@@ -0,0 +1,182 @@
...
+- **Raw Device Events**: Capturing input from input devices without any OS filtering.
+- **Gamepad/Joystick events**: Capturing input from gampads and joysticks.
+  ***//DISCUSS: SHOULD THIS BE SUPPORTED?***
```
vberger: While on X11 this can be done without winit, on wayland this requires at least some collaboration from winit to be possible. I don't know about other platforms.

FEATURES.md
```diff
@@ -0,0 +1,182 @@
...
+- **Device movement events:**: Capturing input from the device gyroscope and accelerometer.
+  ***//DISCUSS: SHOULD THIS BE SUPPORTED?***
```
vberger: I suppose this is mostly relevant for iOS/Android platforms?

I'd say that if we decide Gamepad/Joystick is in scope for winit, then this should be as well.

vberger:
> On Wayland it's impossible without platform calls, and Winit doesn't expose any way of converting client-space coordinates to screen-space coordinates.

Screen-space coordinates are by design not available to clients on wayland. So while it'd be possible to define a parent surface for dialogs, hey cannot be explicitly placed wrt to their parent.

It's an other matter if we are considering tooltips. These are fundamentally different from windows in wayland, and they can be placed wrt to their parent. I think not all platforms make such a distinction, this will require some discussion to unify the platforms on that.

FEATURES.md
Osspial: Yes, I think it would. I don't think you can make a PR against both a wiki and the main repository in the same document, so I'll move the parity table to a *move into wiki on merge* section.

FEATURES.md
Osspial: I assume? I know that it works on Windows, and I'm pretty sure I've done it before on macOS and X11 but I can't quite recall.

FEATURES.md
Osspial: Yeah, and it looks like it's supported on Emscripten as well. I don't know if it works on desktop platforms, though.

I agree that it makes sense to talk about these in the same category as gamepad/joystick events.

Osspial:
@MatusT WinRT compatibility is a feature that I'm not opposed to adding, and it's certainly something that Winit should, in the long run, aim to support! The thing is, implementing it correctly is a huge undertaking (in terms of work seems like it's essentially adding support for another operating system) and I think it's a niche enough target that I'm not willing to put in the work to learning how the API works and supporting it. If somebody is willing to make a PR for it and agrees to maintain it going forward, I don't think you'll find much objection to it getting merged. However, without that I'm not comfortable promising future support for it.

@vberger Could Wayland tooltips be implemented by a downstream library with what Winit currently exposes? That's starting to nudge into the realm of GUI stuff, and my gut instinct is to say that we shouldn't support it in this codebase.

vberger:
> @vberger Could Wayland tooltips be implemented by a downstream library with what Winit currently exposes? That's starting to nudge into the realm of GUI stuff, and my gut instinct is to say that we shouldn't support it in this codebase.

No, winit does not expose enough to make tooltips yet. Though if we support making then in platforms where tooltips and just regular windows with a parent (I think I understood some platforms behave like this?), we may need to support them on wayland as well, for platform compatibility.

FEATURES.md
francesca64: I can confirm that it's similar on X11 [and macOS](https://github.com/tomaka/winit/pull/244#issuecomment-320134156), so I think we can reasonably provide this functionality.

Also, what about knowing the hover position and being able to decide whether to accept or reject a drop (based on both position and type)? Anyone building GUIs on top of winit will demand this, and it's at least easy to implement on X11.

FEATURES.md
francesca64: This is something I've gone back and forth on a lot in my head, though I've ultimately been leaning in favor of it. As @Osspial said, it's something people expect, and it's still largely an open problem in the Rust ecosystem.

I'm pretty happy with the results I've had [implementing Raw Input and XInput gamepad support on Windows](https://github.com/francesca64/winit/tree/win-joy). Since supporting Raw Input requires a window to receive events (which is why it was said to be out of scope for gilrs), and we already have a bunch of Raw Input abstractions in master, winit is really the simplest place this could be handled.

While XInput is polling-based, XInput devices also send Raw Input events, so I was able to rely on that both to only poll when I know we won't have to wait, and to dynamically register which XInput devices are connected and disconnected (attempting to poll a nonexistent XInput device is *very slow*, and AFAIK there's no part of the XInput API for enumerating devices). I think all of this is a great boon, and it's something I don't think could be implemented as elegantly if done outside of winit.

There's the issue of other platforms, though. On Linux, X11 and evdev are rather orthogonal, and different implementations will be required for BSD/etc. I haven't researched macOS at all, or the situation with the non-desktop platforms. Before pursuing a feature like this, I'd want us to have a clear view of how it would be integrated into the event loop on each platform. It's also worth noting that device-specific code is often needed to support force feedback, though Chromium is a good reference for that and it doesn't seem *that* extensive.

I think that any gamepad support provided by winit should be "low level", which is to say, a stream of events that's been transformed only enough to have a consistent format across platforms. Something like SDL mappings would have to be applied downstream, though we'd still have to research what info we'd need to provide for that to be possible.

FEATURES.md
Osspial: From skimming the Windows code, that also seems to be how Windows does it. The one question I have is how do we implement communication from the callback back to the operating system to accept or reject the drop? The best way I can think of doing it would be to pass a mutable reference along with the event that the callback can modify (as otherwise we're getting into the realm of having different return types for different events), but that isn't possible on Windows before Event Loop 2.0 gets merged.

Osspial:
@vberger Hmm. How much more of the Wayland internals would we need to expose to make it possible to implement tooltips externally?

FEATURES.md
Osspial: Another thing - Winit supports dragging content *into* windows. Should we support dragging content *from* windows? In principle I'm not opposed to it, but I don't know what the API would look like and it starts to turn Winit from *windowing and user input* to *general cross-platform desktop interfacing* which is a substantially bigger project.

Perhaps we could publicly expose some of Winit's internals in a permanently unstable way (say, hide them behind a feature or make them all `#[doc(hidden)]`) and split that into a separate crate, alongside non-windowing stuff like clipboard handling? Doing that would have issues and I'm honestly not entirely sold on it myself, but it would allow us to keep the core of Winit simple and build more specialized libraries around it.

FEATURES.md
Osspial: I've opened an issue to discuss how that should work at #704

vberger:
@Osspial 

The "nice" way (for users) would imply exposing SCTK as a public dependency of winit.

The "conservative" way would be to just provide a raw pointer to the C shell surface, but it can be of 3 different possible types, so either the method would return an enum, or there would be 3 methods.

Osspial:
@vberger I'd be most comfortable with exposing the raw pointer and wrapping it with an enum.

FEATURES.md
Osspial: The more I think about it, the less comfortable I am declaring this as in-scope for Winit. Right now, Winit does a decent job of keeping the scope of its API small as exclusively a *window creation* library, and adding clipboard support starts to turn it into a grab-bag of assorted desktop interfacing tools. I don't want to be in a situation where Winit starts snowballing and gaining more and more features that aren't directly related to windowing/user input because it doesn't expose enough of its internals.

@vberger If we exposed Winit's pointers to the Wayland environment through `unix::EventsLoopExt` (so the Wayland registry, display, compositor, subcompositor, shell, data device manager, and decorations manager), could the implementation work for this be moved out of Winit? Admittedly that's a fairly significant number of internals, but they're only necessary to expose because Wayland doesn't use global state nearly as much as other APIs. If we can do that, we can move substantial amounts of design and maintenance work outside Winit's core.

vberger:
This will require some work from sctk, which does not currently expose it. I'm not extremely thrilled about that to be honest, but I understand the tradeoff.

To make sure we're on the same page, you're proposing that everything about tooltips (creating them and drawing their contents) is out of scope for winit?

For example, stuff like "creating a second winit window and using a platform-dependent routine to make it a tooltip"  would not be possible on wayland this way. I don't know how creating tooltips work in other platforms, so I don't know if it may be something users might want to do.

Osspial:
@vberger Yes, that's what I'm proposing. In general my feeling is that all platform-specific GUI and I/O work is out of Winit's scope, and we should focus exclusively on event loop management, renderable window creation, and input handling. If more specialized stuff can't be done with Winit's API that means that Winit isn't exposing enough of its internals.

Also, what part does sctk not expose? It looks like the shell surfaces are exposed through `Proxy`s, which let you obtain a pointer to the underlying C type. If there's more work that has to be done I'd be happy to try and do some of it, as I have access to a machine that can run Wayland and the Rust side of things seems fairly understandable.

Osspial:
Oooooh, I was misinterpreting SCTK's shell global as the shell surface! That makes sense.

FEATURES.md
```diff
@@ -0,0 +1,184 @@
...
+  - macOS
+  - Unix
+    - via X11
+    - via Wayland
+- Mobile
+  - iOS
+  - Android
+- Web
+  - via Emscripten
+  - via WASM ***//DISCUSS: DO WE WANT TO SUPPORT THIS?***
```
est31: Emscripten is tough to install, and furthermore there is a lot of attention on the wasm without emscripten ecosystem. I think that it certainly should be supported.

Osspial:
@vberger Since we're going to be exposing those, I think it makes sense to expose the rest of SCTK's Wayland environment objects. If we do that, is it possible to cleanly support a Wayland clipboard outside of Winit? If it is (or if it only requires us to expose a little bit more to do so) I'd advocate for declaring clipboard support as out of scope and implementing it in a downstream library, since AFAIK Wayland is the only platform where clipboard support can't already be implemented downstream.

vberger:
@Osspial Well, exposing all the environment objects means exposing 7-8 objects in total. those can already be retrieved via the display (that is already exposed), just inducing some slight overhead.

As such, clipboard handling can and has been implemented outside of winit by @trimental, see https://github.com/jwilm/alacritty/pull/1583.

The absolute minimum winit needs to expose is the display, the surface (both are already exposed), and the shell surface (for tooltips support mostly), I think.

Though, having the clipboard implemented outside of winit requires some nontrivial considerations which boils down to forcing the user to manage a second event loop (and thus having to manually advance it from the winit one, or spinning a new thread to run it). The implementation would overall be much simpler if it were in winit.

However, I agree that as clipboard support is *possible* outside of winit, it could be possible to keep it out of its scope, and reconsider the question later, once all the rest is completely consolidated. But this opens the question of drag'n'drop.

The machinery from copy/paste and drag'n'drop is mostly the same in wayland, so from my point of view it'd b strange to support one and not the other. But maybe the separation between them is more pronounced in other platforms?

FEATURES.md
```diff
@@ -0,0 +1,184 @@
...
+### Windowing
+- **Window initialization**: Winit allows the creation of a window
+- **Pointer to OpenGL**: Winit provides the necessary pointers to initialize a working opengl context
+- **Pointer to Vulkan**: Same as OpenGL but for Vulkan
+- **Window decorations**: The windows created by winit are properly decorated, and the decorations can
+  be deactivated
+- **Window decorations toggle**: Decorations can be turned on or off after window creation
+- **Window resizing**: The windows created by winit can be resized and generate the appropriate events
+  when they are. The application can precisely control its window size if wanted.
+- **Window transaprency**: Winit allows the creation of windows with a transparent background.
```
mitchmindtree: transaprency -> transparency

FEATURES.md
```diff
@@ -0,0 +1,184 @@
...
+- **Fullscreen**: The windows created by winit support being fullscreen.
+- **Fullscreen toggle**: The windows created by winit can be switched to and from fullscreen after
+  creation.
+- **Child windows**: Windows can be created relative to the client area of other windows, and parent
+  windows can be disabled in favor of child windows.
+  ***//DISCUSS: SHOULD THIS BE SUPPORTED?***
+
+
+### System Information
+- **Monitor list**: Retrieve the list of monitors and their metada, including which one is primary is applicable
```
mitchmindtree: ~metada~ metadata
is primary ~is~ if applicable

FEATURES.md
mitchmindtree: > Proper child window support is really tricky to get right outside of Winit, and requires platform-specific code to do properly. On Wayland it's impossible without platform calls, and Winit doesn't expose any way of converting client-space coordinates to screen-space coordinates. Additionally, disabling a window in favor of a child window is a common behavior for desktop applications (see file dialogs and are you sure you want to close without saving dialogs) so it makes sense to expose that here.

What it really means for a window to be a child window is still slightly unclear to me, maybe a description of what a child window can offer that another regular window cannot would be worth adding?

Reading the current entry, It seems to me that the relative positioning could be solved using the difference between `get_position` on two regular windows. Disabling a window in favour of another (e.g. disallowing user input for a popup) sounds to me like something could be handled by application logic (e.g. filter out input events associated with `WindowId`s that aren't `popup_window_id`).

I guess I just want to make sure we're not exposing something that could otherwise be done using application logic in order to reduce the winit surface area. I'm also a little concerned that the concept of a "child" window and what that means might be different per platform, though I haven't actually looked into this at all yet - it could very well have a very specific meaning across each platform.

All that said, I have read some users mention that child windows are simply essential in some use cases. One example that comes to mind is the VST API, though I don't know much about it.

Osspial:
@vberger If we also expose our internal `EventQueue` could it be supported without having to spin up another event loop? My general feeling here is that clipboard support is a feature that would be easiest to implement directly inside of Winit, but is leaning more towards data I/O which I read as falling outside Winit's scope of windowing/user input. 

It's also nuanced enough that getting the balance between usability and flexibility in a cross-platform way is a tricky problem (every platform seems to handle multiple-formats/custom-formats in different and slightly incompatible ways, which is super annoying). That'll involve a bit of iteration, and I feel like it'll be easier to do that iteration if it isn't tied directly into Winit. 

I guess the other reason I'm so reluctant to include this directly in Winit is that the clipboard API is entirely independent from window handling on Windows and macOS, although I suppose providing the minimum possible API that works everywhere is kinda the point here :P.

Osspial:
@mitchmindtree I'd reply directly to your feedback about child windows but for whatever reason Github isn't letting me :shrug:. It seems like there's also a bit of a mismatch between what I'm referring to in the document and how the Windows documentation refers to it, so I'm going to be referring to them as *Owned windows* from now on, and will update this document to reflect that.

Owned windows are essentially pop-up windows. They're guaranteed to be drawn above their owning window, but can still be covered up by other windows that don't own them. Specifying their position is tricky from a Winit user's perspective because there isn't a cross-platform way of creating a Window relative to the client area of another Window without using platform-specific API calls - by design, Wayland doesn't let you control the position of a window relative to the desktop, and right now Winit's API *only* lets you control the position of a window relative to the desktop.

Disabling an owned window in favor of its owner can already be done to an extent by application logic, but most desktop shells also have window-disable UI that can't easily be replicated by application logic. For example, on Windows the file save dialog is implemented as an owned window that disables the parent window. When the dialog is open Windows unfocuses the parent window, and if you try to interact with the parent window it plays a sound effect and flashes the owned window to draw attention to it. That effect could *mostly* be replicated by an application, but each platform has different effects for doing the same thing so they'd have to have a bunch of platform-specific code to make it feel native. It's easiest for us to handle that inside Winit.

vberger:
@Osspial 

> If we also expose our internal EventQueue could it be supported without having to spin up another event loop? 

Well, that would be technically possible, but that's not something I'm really willing to do, because the `EventQueue` is not really shareable.

> I guess the other reason I'm so reluctant to include this directly in Winit is that the clipboard API is entirely independent from window handling on Windows and macOS [...]

Okay, that answers my question about other platforms. Like I said, it's currently already possible to implement clipboard handling on wayland with what winit exposes, so that seems acceptable to keep that out of winit, at least for now.

Osspial:
In the feature matrix, raw device events for Windows, MacOS, and Linux X11 are marked as having some bugs. What bugs are they? I wasn't able to find issues for them.

@francesca64

francesca64:
@Osspial 

> Dicussion on this has slowed down quite a bit

I haven't really had a chance to work through everything here, so you might have to wait a bit longer.

> In the feature matrix, raw device events for Windows, MacOS, and Linux X11

I think the reason that was done is because the `DeviceEvent` API isn't fully designed (though I wasn't the one who filled in that row).

X11 is pretty much the reference implementation (though does have this rather onerous bug: #483), but since the API was designed with X11 in mind, I've found some of the concepts don't map well to other platforms. Windows has almost all of the events implemented, though the `Text` event was excluded for being sorta impossible to do correctly. macOS is probably missing most of the events... Emscripten only has `MouseMotion`, and all other platforms seem to not implement `DeviceEvent` at all.

@Ralith was the person who started the `DeviceEvent` concept IIRC, though no one currently active has worked on the design any further. There's still a lot of discussion I'd want to see on `DeviceEvent` in particular before we'd be able to hit 1.0.

Osspial:
Alright, I can open an issue for further discussion of the device events API. I agree that it needs some more work before we declare it done.

The main remaining undecided issue I see is moving the resize increments feature into the core Window API. My feeling on that is that we should have some sort of `Sizing{new_rect: &mut LogicalSize}` event that lets people control the Window's size exactly as it gets resized by the user, but that's dependent on getting Events Loop 2.0 done so we can have non-`'static` events. Does anyone object to adding that to the core Windowing API section?

vberger:
> My feeling on that is that we should have some sort of `Sizing{new_rect: &mut LogicalSize}` event that lets people control the Window's size exactly as it gets resized by the user

:+1: for this, with one small caveat, depending on the environment, there may be a need to distinguish "the system suggested this new size for your window but you can adapt it" vs "the system imposes you this new size and you must follow it". Is this a wayland-only nuance, or is it relevant for other platforms?

Osspial:
Reviving this now, to aid with the marketing push discussed in #830. I've done some minor status updates, but I don't see anything blocking this getting merged. I will do so on April 11th if no objections are raised.

madsmtm:
To any interested parties: we're reopening the clipboard discussion of this again in https://github.com/rust-windowing/winit/pull/2156, feel free to drop by and give input!

Laocoon7:
Is #584 on anyones radar? Interest has been expressed in #850 about a month ago, so I'm not the only person looking for this feature. I would be willing for it to work under the same conditions as glfwSetWindowAspectRatio(); specifically "If the window is full screen, the aspect ratio only takes effect once it is made windowed. If the window is not resizable, this function does nothing."

John-Nagle:
I gather nothing was ever done about double-clicking.

I'd argue that should be in Winit's scope, because there's much platform-specific stuff involved in double-click recognition, especially where that interacts with "accessibility". Winit should provide the same double-click recognition rules as other programs on the same platform, using the current system settings. Otherwise, the Winit user has to have per-platform code, which is rarely needed for Winit. That's the whole point.

Game controllers, etc. probably don't need to be handled by Winit. They're usually handled through the "gilrs" crate, which can handle multiple joysticks, force feedback, and other devices way beyond Winit's department. That crate provides a consistent interface for HID devices across platforms.

kchibisov:
>Winit should provide the same double-click recognition rules as other programs on the same platform, using the current system settings

there's no such rules on X11 and it's all toolkit dependant. Getting that through e.g. gtk is not an option, because e.g. QT is different.

Maybe on macOS/Windows the threshold can be somehow exposed, but it's also, UI toolkit specific most of the time.

John-Nagle:
MacOS and Windows have documented APIs for double click time. Working on figuring out where the Gnome "Settings" program and KDE settings programs for Linux store that info.

No idea how this works in browser WASM.

Egui uses this:

    const MAX_DOUBLE_CLICK_DELAY: f64 = 0.3; // TODO(emilk): move to settings

More info to follow.

kchibisov:
macOS and Windows are GUI toolkits, more or less. X11/Wayland are protocols, protocols don't have settings in general, and you can change settings on the fly.

I'm pretty sure gtk uses dbus, kde probably something else. The also hardcode it sometimes or change the value based on the UI element (I have a feeling that it's different on decorations vs text).

I remember that it's was 300, but I think it's like 400 now in gtk.

John-Nagle:
I'm collecting data on where to get that value. Currently, I've found references for Windows, MacOS, and KDE. Working on finding out where Gnome Settings puts that info. 

[doubleclickapi.txt](https://github.com/rust-windowing/winit/files/14563570/doubleclickapi.txt)

PR #700, opened 2018-11-11
Introduce WindowBuilderExt::with_app_id for wayland

vberger:
This was discussed some time ago, and allows applications to specify an application id that the wayland compositor will use to find the matching `.desktop` file.

### DIFF
--- a/src/os/unix.rs
+++ b/src/os/unix.rs
@@ -225,6 +225,13 @@ pub trait WindowBuilderExt {
     fn with_resize_increments(self, increments: LogicalSize) -> WindowBuilder;
     /// Build window with base size hint. Only implemented on X11.
     fn with_base_size(self, base_size: LogicalSize) -> WindowBuilder;
+
+    /// Build window with a given application ID. It should match the `.desktop` file distributed with
+    /// your program. Only relevant on Wayland.
+    ///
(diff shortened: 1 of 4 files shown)


src/os/unix.rs
```diff
@@ -225,6 +225,13 @@ pub trait WindowBuilderExt {
     fn with_resize_increments(self, increments: LogicalSize) -> WindowBuilder;
     /// Build window with base size hint. Only implemented on X11.
     fn with_base_size(self, base_size: LogicalSize) -> WindowBuilder;
+
+    /// Build window with a given app-id. It should match the `.desktop`
```
francesca64:
```suggestion
    /// Build window with a given `app-id`. It should match the `.desktop`
```
Should `app-id` be in backticks?

src/os/unix.rs
```diff
@@ -225,6 +225,13 @@ pub trait WindowBuilderExt {
...
+    /// file distributed with your program. Only relevant on Wayland.
+    ///
+    /// For details about application id conventions, see the
```
francesca64:
```suggestion
    /// For details about application ID conventions, see the
```
To be more consistent with the stylization you used in the CHANGELOG

PR #701, opened 2018-11-11
travis: freeze dependencies that silently broke 1.24.1 compat

### LINKED ISSUE #698: smithay-client-toolkit fails 1.24.1 Linux builds
The root cause seems to be from [`stb_truetype`](https://crates.io/crates/stb_truetype), which is depended on by [`smithay-client-toolkit`](https://crates.io/crates/smithay-client-toolkit) via [`andrew`](https://crates.io/crates/andrew) via [`rusttype`](https://crates.io/crates/rusttype).

vberger:
Fixes #698

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,6 +63,17 @@ install:
   - rustup self update
   - rustup target add $TARGET; true
 
+before_script:
+  - |
+      if [ "$TRAVIS_RUST_VERSION" == "1.24.1" ]; then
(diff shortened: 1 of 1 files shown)


francesca64:
Until we reach a consensus, we should merge this anyway so contributors won't have to deal with spurious CI failures (...and so I won't get a scary build failure email every time I merge a PR!)

PR #703, opened 2018-11-13
On Windows, catch window callback panics and forward them to the calling thread

### LINKED ISSUE #699: Windows backend invokes undefined behavior upon panic
Right now, panics in the Windows callback function invokes undefined behavior by unwinding into foreign code (see [here](https://doc.rust-lang.org/beta/nomicon/print.html#what-unsafe-rust-can-do)). This is going to become even more of a problem once the Event Loop 2.0 code gets merged and we're running *all* user code in the Windows callback. There should be a panic handler installed that catches the panic, kills the Windows event loop, then resumes panicking from the main event loop.

Osspial:
- [x] ~~Created an example program if it would help users understand this functionality~~

Fixes #699.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 # Unreleased
+- On Windows, catch panics in event loop child thread and forward them to the parent thread.
 
 - On X11, fixed panic caused by dropping the window before running the event loop.
 - Introduce `WindowBuilderExt::with_app_id` to allow setting the application ID on Wayland.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,9 @@
 
 - On X11, fixed panic caused by dropping the window before running the event loop.
 - Introduce `WindowBuilderExt::with_app_id` to allow setting the application ID on Wayland.
-<<<<<<< HEAD
 - On Windows, catch panics in event loop child thread and forward them to the parent thread. This prevents an invocation of undefined behavior due to unwinding into foreign code.
-=======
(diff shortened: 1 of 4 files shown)


francesca64:
It looks like there's something wrong with how you rebased this? It adds a lot of noise to the diff.

Osspial:
Huh, I don't know what was up with that. It looks like that's been fixed now.

src/platform/windows/events_loop.rs
```diff
@@ -370,45 +391,79 @@ lazy_static! {
...
+    // If a panic has been triggered, cancel all future operations in the function.
+    if CONTEXT_STASH.with(|stash| stash.borrow().as_ref().map(|s| s.panic_error.is_some()).unwrap_or(false)) {
+        return error;
+    }
+
+    let callback_result = panic::catch_unwind(f);
+    match callback_result {
+        Ok(lresult) => lresult,
+        Err(err) => CONTEXT_STASH.with(|context_stash| {
+            println!("catch panic!");
```
francesca64: Was this `println!` left in intentionally?

Osspial:
Thanks for the review! Those changes have been applied. Also, I'd like to wait for #710 to be reviewed/merged before merging this, since we should also be catching panics created by the `thread_event_target_callback` introduced there.

CHANGELOG.md
```diff
@@ -2,8 +2,12 @@
 
 - On X11, fixed panic caused by dropping the window before running the event loop.
 - Introduce `WindowBuilderExt::with_app_id` to allow setting the application ID on Wayland.
+<<<<<<< HEAD
```
francesca64: ​Whoops!

PR #713, opened 2018-11-18
Add ordering traits to VirtualKeyCode

oskargustafsson:
Motivation:
This allows VirtualKeyCode variants to be stored in a BTreeSet.
Unlike HashSets, BTreeSets implement Ord and Hash, allowing them to be
keys in a {Hash|BTree}Map. This is nice, e.g. when implementing
keyboard shortcuts functionality, which maps a set of pressed keys to
some action.


Osspial:
That rationale seems reasonable enough. I don't see any real reason not to support this.

francesca64:
I'm also not opposed to this. The only downside is that reordering variants would be a breaking change, but we also never reorder variants, so it's not really an issue.

How about you, @vberger?

vberger:
I don't have any objection either.

francesca64:
Alright! Thanks for the PR!

PR #718, opened 2018-11-20
On Windows, fix window shrinking when leaving fullscreen in some situations

### LINKED ISSUE #715: Windows: Toggling between decorations and fullscreen shrinks window
1. Turn decorations off
2. Enter fullscreen
3. Exit fullscreen
4. Turn decorations back on
5. Repeat cycle

Each time you exit fullscreen, the window will get a bit smaller (presumably by the same width as the border that flashes?). Step 4 is the key to being able to repeat this endlessly, potentially ending up with a very shrunken window.

This can be reproduced in the `fullscreen` example.

Osspial:

Fixes #715. This seems to have been caused by the window style getting improperly set when leaving fullscreen mode.

The Windows fullscreen window restore code is a bit convoluted in general so I'd like to take some time to refactor and simplify that a bit, but this can be fixed without doing that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - Introduce `WindowBuilderExt::with_app_id` to allow setting the application ID on Wayland.
 - On Windows, fix issue where resizing or moving window combined with grabbing the cursor would freeze program.
 - On Windows, fix issue where resizing or moving window would eat `Awakened` events.
+- On Windows, exiting fullscreen after entering fullscreen with disabled decorations no longer shrinks window.
 
 # Version 0.18.0 (2018-11-07)
 
--- a/examples/fullscreen.rs
+++ b/examples/fullscreen.rs
@@ -36,7 +36,7 @@ fn main() {
     let mut decorations = true;
 
     events_loop.run_forever(|event| {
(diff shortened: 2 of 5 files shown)


francesca64:
This introduces another fun bug that makes the window grow:
1. Enter fullscreen with decorations on
2. Exit fullscreen with decorations off
3. Repeat

Osspial:
@francesca64 Heh, funny that this would add a bug that does exactly the opposite of the original bug! Regardless, Windows now saves the client rect instead of the adjusted window rect when we enter fullscreen so this and any similar issues should be fixed.

francesca64:
Awesome! This next one isn't a regression, but might make sense to tackle here too:
1. Maximize with decorations off
2. Un-maximize with decorations on
3. Repeat

This progressively makes the window shorter by the same height as the title bar.

Osspial:
We're using the official Windows restore-size-from-maximize function there, so fixing that would involve a bit more work than I'm willing to do in this PR. I'm working on a bigger refactor to the Windows style-flags code though (which touches that), so I'll see about fixing it in that.

francesca64:
Alright, sounds great!

PR #724, opened 2018-11-22
Use smithay-client-toolkit's dpi handling.

dvc94ch:
This depends on https://github.com/smithay/client-toolkit/pull/46


dvc94ch:
ping @vberger

dvc94ch:
Done

francesca64:
@vberger should this have a CHANGELOG entry?

dvc94ch:
> should this have a CHANGELOG entry?

It doesn't really change anything from a winit user perspective, it works identically to the current version but is IMO easier to read. The only gain for winit is dpi issues affect all smitahy-client-toolkit users and should be opened or moved to the smithay-client-toolkit repo, so less code to maintain, and for smithay-client-toolkit users the advantage is they have tried and tested dpi handling.

PR #730, opened 2018-11-29
Refactor win32 window state code

### LINKED ISSUE #723: On Windows, `hide_cursor` hides cursor for all Winit windows
It should only hide the cursor for the window that `hide_cursor` gets called on.

Osspial:

This PR refactors the Windows state-setting code to make it significantly easier to read and modify going forward. It also fixes #723 and a few other miscellaneous untracked issues I found when doing the refactor.

### TODO:
- [x] Mask out `WS_CHILD` when `WS_POPUP` is enabled
- [x] Fix `set_maximized`
- [ ] Fix the unfixed window shrinking bug in #718

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,8 +22,8 @@
 - Removed minimum supported Rust version guarantee.
 - On Windows, fix malformed function pointer typecast that could invoke undefined behavior.
 - Refactored Windows state/flag-setting code.
-  - Hiding the cursor no longer hides the cursor for all Winit windows - just the one `hide_cursor` was called on.
-  - Beforehand, cursor grabs would get perpetually canceled when the grabbing window lost focus. Now, cursor grabs get automatically re-initialized when the window regains focus and the mouse moves over the client area.
+- On Windows, hiding the cursor no longer hides the cursor for all Winit windows - just the one `hide_cursor` was called on.
+- On Windows, cursor grabs used to get perpetually canceled when the grabbing window lost focus. Now, cursor grabs automatically get re-initialized when the window regains focus and the mouse moves over the client area.
 
 # Version 0.18.0 (2018-11-07)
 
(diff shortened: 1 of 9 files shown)


JasonHise:
Was about to submit an issue but it looks like this is still open and is probably the same thing?  Our particular use case is that the window is shrinking when toggling back and forth between windowed and borderless fullscreen on windows, my shot in the dark guess would be that there is an inconsistency between whether the window still/already has a border when the client rect gets saved and restored, respectively.

aloucks:
I experienced something similar where the size/position wasn't restoring correctly, but eventually sorted it out by setting everything manually instead of using the `set_fullscreen` API. The logic boils down to:

```rust
enum WindowMode {
    Fullscreen {
        last_position: LogicalPosition,
        last_size: LogicalSize,
    },
    Windowed,
}

pub fn toggle_window_mode(&mut self) {
    match self.window_mode {
        WindowMode::Windowed => {
            let last_position = self.window.outer_position().unwrap();
            let last_size = self.window.inner_size();
            let monitor = self.window.current_monitor();

            let dpi_factor = monitor.hidpi_factor();
            let x = monitor.position().x as _;
            let inner_physical_size = monitor.size();

            self.window.set_visible(false);
            self.window.set_decorations(false);
            self.window.set_outer_position(LogicalPosition::from((x, 0)));
            self.window.set_inner_size(inner_physical_size.to_logical(dpi_factor));
            self.window.set_visible(true);

            self.window_mode = WindowMode::Fullscreen {
                last_position,
                last_size,
            };
        }
        WindowMode::Fullscreen {
            last_position,
            last_size,
        } => {
            self.window.set_visible(false);
            self.window.set_decorations(true);
            self.window.set_inner_size(last_size);
            self.window.set_outer_position(last_position);
            self.window.set_visible(true);

            self.window_mode = WindowMode::Windowed;
        }
    }
}
```

Osspial:
@JasonHise Actually, looking back at this, it looks like that's a separate bug entirely! It looks like there was a small oversight in some code that arrived later and we weren't handling the window rect restore correctly. It's a fairly simple fix though, so I've made a PR here: #1172.

Sorry about dismissing your issue earlier!

JasonHise:
@Osspial awesome, thanks!

Issue #706, opened 2018-11-14
High CPU load when moving window between monitors


NickeZ:
Initially reported here: https://github.com/jwilm/alacritty/issues/1790

Which operating system does the issue occur on?
Ubuntu 18.04.1 Linux (kernel version 4.15)

If on linux, are you using X11 or Wayland?
* X11
* I3 version 4.14.
* (I don't have any compositor)

Using the window example I can cause excessive cpu load. Should the example application `window.rs` be an empty see-through window?

Is there anything I can do to debug it when it happens?

![high_load](https://user-images.githubusercontent.com/492753/48475201-9311eb00-e7fc-11e8-8b6c-eb14e653b37f.png)

francesca64:
Thanks for reporting!

> Should the example application window.rs be an empty see-through window?

If nothing is rendered to it, yes. Though, in my experience, rather than being see-through it usually merely has the same appearance of whatever was behind it when it was opened.

> Is there anything I can do to debug it when it happens?

Does anything look odd about the frequency/latency of the events being printed out? My best guess is that something's going wrong when we try to listen for new events. If not, I'll look into more rigorous ways to investigate this.

chrisduerr:
I've just checked this to make sure Alacritty's not doing anything wrong and I've come to the conclusion that this must be something with winit.

Note that the `run_forever` closure isn't actually executed continuously, which could trigger a problem (like an infinite loop without any content). In fact in my testing it was completely silent after moving the example to a different screen, however the CPU was still running hot.

NickeZ:
I updated the topic because the reproducible way to get to this state is by moving the window between monitors. I totally forgot that I moved the example window to a second screen...

> Does anything look odd about the frequency/latency of the events being printed out? My best guess is that something's going wrong when we try to listen for new events. If not, I'll look into more rigorous ways to investigate this.

It doesn't print anything special when CPU usage goes up. Just the regular keyboard/mouse events.

NickeZ:
I ran it with strace. I get continuous, very high frequency, output of these:

```
poll([{fd=3, events=POLLIN}], 1, -1)    = 1 ([{fd=3, revents=POLLIN}])
[...]

PR #737, opened 2018-12-07
Fix high CPU usage on tiling WMs when moving windows across monitors (#706)

Jasu:
This commit restricts an Xfwm4-specific DPI-preserving hack to Xfwm4 only, fixing #706.

The hack saves and restores the DPI-adjusted size until the actual size matches. On tiling WMs like i3 this fails, since the size is constrained by the layout. This in turn causes a never-ending XResizeWindow vs. XConfigureWindow fight between the WM and the client, making the WM, winit client, and Xorg consume all CPU cycles available.

**Update:** I also tested this on fluxbox, and it worked (preserved the DPI.)
**Update 2:** I got paranoid that I might have removed the only point where the DPI is preserved, so I tested this on one more WM, icewm, and it also worked on it.

- [X] Tested on all platforms changed **Tested on Xfwm4, fluxbox, icewm, and i3**
- [ ] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior **Not applicable**
- [ ] Created an example program if it would help users understand this functionality **Not applicable**

### DIFF
--- a/src/platform/linux/x11/mod.rs
+++ b/src/platform/linux/x11/mod.rs
@@ -499,10 +499,12 @@ impl EventsLoop {
                     }
 
                     // This is a hack to ensure that the DPI adjusted resize is actually applied on all WMs. KWin
-                    // doesn't need this, but Xfwm does.
+                    // doesn't need this, but Xfwm does. The hack should not be run on other WMs, since tiling
+                    // WMs constrain the window size, making the resize fail. This would cause an endless stream of
+                    // XResizeWindow requests, making Xorg, the winit client, and the WM consume 100% of CPU.
                     if let Some(adjusted_size) = shared_state_lock.dpi_adjusted {
                         let rounded_size = (adjusted_size.0.round() as u32, adjusted_size.1.round() as u32);
-                        if new_inner_size == rounded_size {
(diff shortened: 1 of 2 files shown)


src/platform/linux/x11/mod.rs
```diff
@@ -499,20 +499,24 @@ impl EventsLoop {
...
-                            shared_state_lock.dpi_adjusted = None;
-                        } else {
-                            unsafe {
-                                (self.xconn.xlib.XResizeWindow)(
-                                    self.xconn.display,
-                                    xwindow,
-                                    rounded_size.0 as c_uint,
-                                    rounded_size.1 as c_uint,
-                                );
+                        if util::wm_name_is_one_of(&["Xfwm4"]) {
```
Jasu: I wonder if this should be combined with the `new_inner_size == rounded_size` two lines below? Any opinions?

Jasu: I realized that with the check here the WM name comparison would be done on every `XConfigureWindow`. I combined the check with the one below and force-pushed. I re-tested with Xfwm4 and i3.

vberger:
LGTM

PR #738, opened 2018-12-14
Add support for generating dummy DeviceIDs and WindowIDs

### LINKED ISSUE #330: Support for creating dummy DeviceIds and WindowIds
I have a rather large set of code that processes winit events and stores input state, as well as generating input events for key bindings.  I'd like to add test cases to it to make sure the input handler always works but I can't generate winit::Events in a headless environment because DeviceId and WindowId are opaque types.  It'd be nice to be able to generate dummy types for these based on some integer.

Xaeroxe:
Fixes #330 

  - [x] Windows
  - [x] Android
  - [x] Emscripten
  - [x] iOS
  - [x] Linux
  - [x] OSX

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 - On Wayland, fix resizing and DPI calculation when a `wl_output` is removed without sending a `leave` event to the `wl_surface`, such as disconnecting a monitor from a laptop.
 - On Wayland, DPI calculation is handled by smithay-client-toolkit.
 - On X11, `WindowBuilder::with_min_dimensions` and `WindowBuilder::with_max_dimensions` now correctly account for DPI.
-- Added support for generating dummy DeviceIds and WindowIds to better support unit testing.
+- Added support for generating dummy `DeviceId`s and `WindowId`s to better support unit testing.
 
 # Version 0.18.0 (2018-11-07)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - On Wayland, fix resizing and DPI calculation when a `wl_output` is removed without sending a `leave` event to the `wl_surface`, such as disconnecting a monitor from a laptop.
 - On Wayland, DPI calculation is handled by smithay-client-toolkit.
 - On X11, `WindowBuilder::with_min_dimensions` and `WindowBuilder::with_max_dimensions` now correctly account for DPI.
+- Added support for generating dummy DeviceIds and WindowIds to better support unit testing.
 
 # Version 0.18.0 (2018-11-07)
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -170,8 +170,9 @@ impl std::fmt::Debug for Window {
 pub struct WindowId(platform::WindowId);
 
 impl WindowId {
-    /// Returns a dummy `WindowId`, useful for unit testing. Absolutely no guarantees are made
-    /// about this `WindowId`, it may be equal to a real `WindowId`.
+    /// Returns a dummy `WindowId`, useful for unit testing. The only guarantee made about the return
+    /// value of this function is that it will always be equal to itself and to future values returned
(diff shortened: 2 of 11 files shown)


Xaeroxe:
CI failure seems unrelated to this PR: https://travis-ci.org/tomaka/winit/jobs/468126803#L684

src/lib.rs
```diff
@@ -169,6 +169,16 @@ impl std::fmt::Debug for Window {
 #[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct WindowId(platform::WindowId);
 
+impl WindowId {
+    /// Returns a dummy WindowId, useful for unit testing. Absolutely no guarantees are made
+    /// about this window id, it may be equal to a real WindowId.
+    ///
+    /// **Passing this into a winit function will result in undefined behavior.**
```
Osspial: It doesn't look like Winit has any publicly-facing functions that take `WindowId`s, so this warning doesn't currently apply anywhere. However, this function should be marked unsafe for forwards-compatibility reasons since it could cause undefined behavior issues if we *do* introduce such functions. Once we approach 1.0, we can look at making this safe if we know we aren't going to introduce other functions that make it unsafe.

src/lib.rs
```diff
@@ -177,6 +187,16 @@ pub struct WindowId(platform::WindowId);
 #[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct DeviceId(platform::DeviceId);
 
+impl DeviceId {
+    /// Returns a dummy DeviceId, useful for unit testing. Absolutely no guarantees are made
+    /// about this device id, it may be equal to a real DeviceId.
+    ///
+    /// **Passing this into a winit function will result in undefined behavior.**
```
Osspial: See `WindowId` comments.

src/lib.rs
```diff
@@ -169,6 +169,16 @@ impl std::fmt::Debug for Window {
 #[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct WindowId(platform::WindowId);
 
+impl WindowId {
+    /// Returns a dummy WindowId, useful for unit testing. Absolutely no guarantees are made
+    /// about this window id, it may be equal to a real WindowId.
```
Osspial: WindowId should be surrounded in backticks so they appear as inline code blocks

```
`like this`
```

Xaeroxe:
@Osspial comments resolved, thank you!

Xaeroxe:
Is it worth providing the guarantee that the return value of the `dummy` function will always at minimum be equal to future return values of the `dummy` function?

Xaeroxe:
which is to say `dummy()` is deterministic, and the return value is always guaranteed to be equal to itself.

Osspial:
@Xaeroxe Yes, I think that makes sense.

Xaeroxe:
@Osspial what are the next steps on getting this merged?

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 - On Wayland, fix resizing and DPI calculation when a `wl_output` is removed without sending a `leave` event to the `wl_surface`, such as disconnecting a monitor from a laptop.
 - On Wayland, DPI calculation is handled by smithay-client-toolkit.
 - On X11, `WindowBuilder::with_min_dimensions` and `WindowBuilder::with_max_dimensions` now correctly account for DPI.
+- Added support for generating dummy DeviceIds and WindowIds to better support unit testing.
```
francesca64:
```suggestion
- Added support for generating dummy `DeviceId`s and `WindowId`s to better support unit testing.
```

Xaeroxe:
Thank you for your reviews and for merging this!

I'm eager to get started writing some unit tests in Amethyst using this new functionality.  When are we planning on releasing the next version of winit to crates.io?

francesca64:
That will optimistically happen before the end of the month.

PR #739, opened 2018-12-18
Support Yen in macOS

lambdalisue:
    - Tested on macOS Mojave (10.14.2)
    - Not sure if it should
    - Not sure if I should
    - Not sure if I should
---

Thanks for your project. I had some problems to remap `Yen` key in macOS (I am using alacritty) so I made this PR.
This PR does

- Support `Yen` key of JIS in macOS
- Add comments for some "unknown" keycoes

---

01-jis-keylayout

key codes 2018-12-18 10-42-31


lambdalisue:
Done 👍

PR #740, opened 2018-12-20
Improve `set_cursor` on macOS

francesca64:
This changes cursor setting to follow the correct macOS conventions. The previous implementation had various problems, i.e. the cursor would be unset when the window lost focus.

This also implements all `MouseCursor` variants, which was quite the research project. Having to load some of them from WebKit is pretty silly, but it seems to be the most practical solution.


src/platform/macos/util/cursor.rs
```diff
@@ -0,0 +1,133 @@
...
+    ];
+    let pdf_path: id = msg_send![cursor_path,
+        stringByAppendingPathComponent:cursor_pdf
+    ];
+    let info_path: id = msg_send![cursor_path,
+        stringByAppendingPathComponent:cursor_plist
+    ];
+
+    let image = NSImage::alloc(nil).initByReferencingFile_(pdf_path);
+    let info = NSDictionary::dictionaryWithContentsOfFile_(
```
mtak-: Maybe we need checks for `nil` on `image` and `info`?

PR #742, opened 2018-12-20
macOS: Implement `Refresh`

francesca64:


src/platform/macos/view.rs
```diff
@@ -154,6 +158,27 @@ extern fn init_with_winit(this: &Object, _sel: Sel, state: *mut c_void) -> id {
     }
 }
 
+extern fn draw_rect(this: &Object, _sel: Sel, rect: id) {
```
mtak-: `rect` should be of type `NSRect`.

PR #746, opened 2018-12-22
Remove MSRV guarantee

### LINKED ISSUE #743: Linux build broken on 1.28.0
[`stb_truetype` updated to Rust 2018 (1.31.0)](https://github.com/redox-os/stb_truetype-rs/blob/master/Cargo.toml#L5), and once again broke our compatibility guarantees with a patch release.

So, we need to revisit the discussion from #698. We can either...
- Pin dependencies to older versions as necessary
- Move forward to 1.31.0

Osspial:
As per discussion in #743. Closes #743.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 - On macOS, fixed unsoundness in drag-and-drop that could result in drops being rejected.
 - On macOS, implemented `WindowEvent::Refresh`.
 - On macOS, all `MouseCursor` variants are now implemented and the cursor will no longer reset after unfocusing.
+- **Breaking:** Removed minimum supported Rust version guarantee.
 
 # Version 0.18.0 (2018-11-07)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,11 +19,10 @@
 - On macOS, fixed unsoundness in drag-and-drop that could result in drops being rejected.
 - On macOS, implemented `WindowEvent::Refresh`.
 - On macOS, all `MouseCursor` variants are now implemented and the cursor will no longer reset after unfocusing.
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 - On Wayland, DPI calculation is handled by smithay-client-toolkit.
 - On X11, `WindowBuilder::with_min_dimensions` and `WindowBuilder::with_max_dimensions` now correctly account for DPI.
 - Added support for generating dummy `DeviceId`s and `WindowId`s to better support unit testing.
+- **Breaking:** Removed minimum supported Rust version guarantee.
```
francesca64: I'm not sure how much of a breaking change this is. I mean, compatibility is already broken in 0.18.0...

On that same note, should we remove the entry from 0.18.0 about the 1.28.0 change? That was something I added *after* release, so it's always felt a bit weasely, and it's not even true anymore.

Osspial: If we remove the `1.28.0` line from the changelog, I feel like we should remove all MSRV lines from the changelog. I'd be fine with doing that, but I don't feel too strongly either way.

francesca64: Do we have any other MSRV lines in the changelog, though? The only other mention of MSRV changes I see is under 0.10.0, and that feels a bit different to me.

Osspial: Eh, fair enough. Just to be clear about where we stand: this shouldn't be marked as a breaking change but still be in the changelog, and the `1.28` change should be removed?

francesca64: That's my position, yes.

Osspial: SGTM. Those changes have been made.

Issue #748, opened 2018-12-29
BadDrawable with Vulkan/XLib window on Linux


kvark:
Getting this when running any of the Ash or gfx-hal examples:
```
[winit X11 error] XError {
    description: "BadDrawable (invalid Pixmap or Window parameter)",
    error_code: 9,
    request_code: 149,
    minor_code: 4
}
```
Apparently happens on `poll_events` without the user code doing anything, so may be inside `winit`.

Linux: 4.19.8-104.current #1 SMP PREEMPT Wed Dec 12 01:23:21 UTC 2018 x86_64 GNU/Linux

kvark:
@Osspial note that it's super annoying, since it prevents me from running anything with Vulkan at the moment. Let me know if you have ideas, or if you don't have capacity to investigate.

Osspial:
@kvark I've got a Linux machine, but I don't know if it supports Vulkan and I know basically nothing about how the X11 backend works. If Vulkan *does* work I'll see if I can reproduce this, but I'm afraid I don't have the expertise to do much else.

kvark:
@Osspial does't sound like you'd be the best person to look into this problem, unless you do want to get more understanding of X11 and Vulkan. Should I reach out to other contributors? Or should I drop the hope for help and try to figure this out myself?

Osspial:
@kvark It doesn't look like my Linux machine supports Vulkan, so I can't look into this. @francesca64 expressed interest in looking into this, so you should talk to her.

As far as other information that would be useful: if you run a Vulkan application that uses SDL to create windows, do you get a similar error? If you do, it's probably not an issue on our end.

kvark:
Looks like even `vkcube` isn't showing anything, so it's not `winit`, phew!
Sorry about the trouble. I didn't realize my system is so dysfunctional at this point.

yorickdewid:
Its likely Vulkan does not load the driver for the correct card. Try

```sh
VK_ICD_FILENAMES=/usr/share/vulkan/icd.d/nvidia_icd.json vkcube
```

Uniformbuffer3:
I have recently opened an issue on mesa about the problem and a developer have merged a patch that should fix the problem: https://gitlab.freedesktop.org/mesa/mesa/-/issues/4688

PR #749, opened 2018-12-30
Version 0.18.1

Osspial:


rukai:
Looks like there are no breaking changes? Can this be 0.18.1?

Osspial:
This has some minor API additions which makes me lean towards releasing this with a full version increment.

francesca64:
@Osspial are you sure you're not just doing this so `EventLoop` 2.0 will be 0.20? That would be pretty cool, so I wouldn't blame you.

Joking aside, I'm not opposed to this, but I personally would've gone with 0.18.1 just for the downstream convenience. I do think it would be nice to establish guidelines for release versioning; AFAIK the most recent discussion is here: https://github.com/tomaka/winit/pull/468#issuecomment-382607437

Anyway, before releasing, I still want to review #730 and investigate #748. Is this intended to be our last (planned) pre-`EventLoop` 2.0 release?

Osspial:
@francesca64 I... hadn't been thinking about that, but now I'm disappointed that I wasn't!

Anyhow, my modus operandi with `0.x.y` releases has been to treat them as standard semver, bumping the `x` number for releases with any API additions and bumping `y` for releases that only have bugfixes. I can see the rationale for treating `y` as the minor version number though, and wouldn't be opposed to releasing this as `0.18.1`.

I was planning for this to be the last pre-EL2.0 release, yes. If you want to hold off on releasing this until we handle #730 and #748, I'd be fine with it.

vberger:
Regarding version numbers, even if this release is 0.18.1, nothing prevents us from releasing eventloop 2.0 as 0.20 and skipping 0.19 in the process. (With wayland-client I went directly from 0.14 to 0.20 with my last big rewrite to mark the fact that it was fundamentally different.)

On this regard, if there are no breaking change, I think it is better if this is released as 0.18.1, so that `cargo update` works out of the box for anyone depending on `winit = "0.18"`, as that is the way cargo interprets semver for pre-1.0.

Xaeroxe:
As a winit user with a few dependencies that also depend on winit releasing this as a patch version would make me very happy,

Xaeroxe:
I just double checked for semver breakage in this release, and I couldn't find any.

Osspial:
Alright, this has been reformatted as `0.18.1`.

Osspial:
@francesca64 Would you be up for releasing the current slate of changes now, then releasing `0.18.2` once the issues you brought up have been addressed?

chrisduerr:
I have nothing to say here, but I'd be in favor of publishing a new version as quickly as possible since things like high resource usage when moving windows in between monitor's is a pretty serious issues which affects a decent number of users.

Just my 2 cents from the perspective of a winit enduser. Otherwise the solution might be to link against the git version of winit temporarily, but that doesn't really help anyone.

Osspial:
Since there hasn't been any comment from @francesca64 and there's a definite benefit for releasing these changes now, I'm going to go ahead and merge this and leave those other two issues for a future patch release.

Issue #752, opened 2019-01-04
Virtual Keycode is incorrect on macOS non-qwerty layouts


Toqozz:
The `virtual_keycode` field seems to be incorrect when using a non-qwerty layout.  It seems to always return values corresponding to the qwerty (or perhaps physical?) layout.

I'm running the following program:
```
extern crate winit;

fn main() {
    let mut events_loop = winit::EventsLoop::new();
    let window = winit::Window::new(&events_loop).unwrap();

    events_loop.run_forever(|event| {
        println!("{:?}", event);
        match event {
            winit::Event::WindowEvent {
                event: winit::WindowEvent::CloseRequested, ..
            } => winit::ControlFlow::Break,
            _ => winit::ControlFlow::Continue,
        }
    });
}
```

Pressing a dvorak `P` here (qwerty 'R')  results in the following lines:
```
WindowEvent { window_id: WindowId(Id(140217433672064)), event: KeyboardInput { device_id: DeviceId(DeviceId), input: KeyboardInput { scancode: 15, state: Pressed, virtual_keycode: Some(R), modifiers: ModifiersState { shift: false, ctrl: false, alt: false, logo: false } } } }

WindowEvent { window_id: WindowId(Id(140217433672064)), event: ReceivedCharacter('p') }
```

As you can see, we receive `virtual_keycode: Some(R)`.  Also note that right after the window receives the `P` character.  (Shouldn't these always be the same?)

The issue doesn't seem to occur on Linux/X11.

```
winit version: 0.18
MacBook Pro (2017)
macOS version: 10.14 (Mojave)
```

francesca64:
Thanks for reporting this!

We're currently using `NSEvent keyCode` to [generate `VirtualKeyCode`](https://github.com/tomaka/winit/blob/45a428141329377a1669698e4ca6561834c0dbcc/src/platform/macos/events_loop.rs#L547): https://github.com/tomaka/winit/blob/139686ddce870614f12eb3a0ed33a9adfee1e47a/src/platform/macos/view.rs#L415

This is wrong, since that's really just a scancode: https://stackoverflow.com/a/18413469/5435443

[...]

PR #755, opened 2019-01-08
Fix incorrect keycodes when using a non-US keyboard layout.

Toqozz:
This commit fixes the issue described in #752, and uses the advised
method to fix it.

I'm relatively new to contributing and Rust (this is my first real contribution), and I'm sure my lack of experience is shining through here.  Regardless, I hope that this PR fixes/assists in fixing the issue adequately!

### DIFF
--- a/src/platform/macos/events_loop.rs
+++ b/src/platform/macos/events_loop.rs
@@ -544,68 +544,67 @@ impl Proxy {
     }
 }
 
-pub fn windows_key_code_for_char_code(c: char) -> Option<c_ushort> {
+pub fn char_to_keycode(c: char) -> Option<events::VirtualKeyCode> {
     // We only translate keys that are affected by keyboard layout.
     //
     // Note that since keys are translated in a somewhat "dumb" way (reading character)
     // there is a concern that some combination, i.e. Cmd+char, causes the wrong
     // letter to be received, and so we receive the wrong key.
     //
-    // Codes taken from `to_virtual_key_code`.
-    // It may be better to map these codes to something more immediately meaningful.
+    // Implementation reference: https://github.com/WebKit/webkit/blob/82bae82cf0f329dbe21059ef0986c4e92fea4ba6/Source/WebCore/platform/cocoa/KeyEventCocoa.mm#L626
     Some(match c {
-        'a' | 'A' => 0x00,
-        's' | 'S' => 0x01,
-        'd' | 'D' => 0x02,
-        'f' | 'F' => 0x03,
-        'h' | 'H' => 0x04,
-        'g' | 'G' => 0x05,
-        'z' | 'Z' => 0x06,
-        'x' | 'X' => 0x07,
-        'c' | 'C' => 0x08,
-        'v' | 'V' => 0x09,
-        'b' | 'B' => 0x0b,
-        'q' | 'Q' => 0x0c,
-        'w' | 'W' => 0x0d,
-        'e' | 'E' => 0x0e,
-        'r' | 'R' => 0x0f,
-        'y' | 'Y' => 0x10,
-        't' | 'T' => 0x11,
-        '1' | '!' => 0x12,
-        '2' | '@' => 0x13,
-        '3' | '#' => 0x14,
-        '4' | '$' => 0x15,
-        '6' | '^' => 0x16,
-        '5' | '%' => 0x17,
-        '=' | '+' => 0x18,
-        '9' | '(' => 0x19,
-        '7' | '&' => 0x1a,
-        '-' | '_' => 0x1b,
-        '8' | '*' => 0x1c,
-        '0' | ')' => 0x1d,
-        ']' | '}' => 0x1e,
-        'o' | 'O' => 0x1f,
-        'u' | 'U' => 0x20,
-        '[' | '{' => 0x21,
-        'i' | 'I' => 0x22,
-        'p' | 'P' => 0x23,
-        'l' | 'L' => 0x25,
-        'j' | 'J' => 0x26,
-        '\''| '"' => 0x27,
-        'k' | 'K' => 0x28,
-        ';' | ':' => 0x29,
-        '\\'| '|' => 0x2a,
-        ',' | '<' => 0x2b,
-        '/' | '?' => 0x2c,
-        'n' | 'N' => 0x2d,
-        'm' | 'M' => 0x2e,
-        '.' | '>' => 0x2f,
-        '`' | '~' => 0x32,
+        'a' | 'A' => events::VirtualKeyCode::A,
+        'b' | 'B' => events::VirtualKeyCode::B,
+        'c' | 'C' => events::VirtualKeyCode::C,
+        'd' | 'D' => events::VirtualKeyCode::D,
+        'e' | 'E' => events::VirtualKeyCode::E,
+        'f' | 'F' => events::VirtualKeyCode::F,
+        'g' | 'G' => events::VirtualKeyCode::G,
+        'h' | 'H' => events::VirtualKeyCode::H,
+        'i' | 'I' => events::VirtualKeyCode::I,
+        'j' | 'J' => events::VirtualKeyCode::J,
+        'k' | 'K' => events::VirtualKeyCode::K,
+        'l' | 'L' => events::VirtualKeyCode::L,
+        'm' | 'M' => events::VirtualKeyCode::M,
+        'n' | 'N' => events::VirtualKeyCode::N,
+        'o' | 'O' => events::VirtualKeyCode::O,
+        'p' | 'P' => events::VirtualKeyCode::P,
+        'q' | 'Q' => events::VirtualKeyCode::Q,
+        'r' | 'R' => events::VirtualKeyCode::R,
+        's' | 'S' => events::VirtualKeyCode::S,
+        't' | 'T' => events::VirtualKeyCode::T,
+        'u' | 'U' => events::VirtualKeyCode::U,
+        'v' | 'V' => events::VirtualKeyCode::V,
+        'w' | 'W' => events::VirtualKeyCode::W,
+        'x' | 'X' => events::VirtualKeyCode::X,
+        'y' | 'Y' => events::VirtualKeyCode::Y,
+        'z' | 'Z' => events::VirtualKeyCode::Z,
+        '1' | '!' => events::VirtualKeyCode::Key1,
+        '2' | '@' => events::VirtualKeyCode::Key2,
+        '3' | '#' => events::VirtualKeyCode::Key3,
+        '4' | '$' => events::VirtualKeyCode::Key4,
+        '5' | '%' => events::VirtualKeyCode::Key5,
+        '6' | '^' => events::VirtualKeyCode::Key6,
+        '7' | '&' => events::VirtualKeyCode::Key7,
+        '8' | '*' => events::VirtualKeyCode::Key8,
+        '9' | '(' => events::VirtualKeyCode::Key9,
+        '0' | ')' => events::VirtualKeyCode::Key0,
+        '=' | '+' => events::VirtualKeyCode::Equals,
+        '-' | '_' => events::VirtualKeyCode::Minus,
+        ']' | '}' => events::VirtualKeyCode::RBracket,
+        '[' | '{' => events::VirtualKeyCode::LBracket,
+        '\''| '"' => events::VirtualKeyCode::Apostrophe,
+        ';' | ':' => events::VirtualKeyCode::Semicolon,
+        '\\'| '|' => events::VirtualKeyCode::Backslash,
+        ',' | '<' => events::VirtualKeyCode::Comma,
+        '/' | '?' => events::VirtualKeyCode::Slash,
+        '.' | '>' => events::VirtualKeyCode::Period,
+        '`' | '~' => events::VirtualKeyCode::Grave,
         _ => return None,
     })
 }
 
-pub fn to_virtual_key_code(code: c_ushort) -> Option<events::VirtualKeyCode> {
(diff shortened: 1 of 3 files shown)


src/platform/macos/events_loop.rs
```diff
@@ -544,6 +544,67 @@ impl Proxy {
     }
 }
 
+pub fn windows_key_code_for_char_code(c: char) -> Option<c_ushort> {
```
francesca64:
```suggestion
pub fn char_to_keycode(c: char) -> Option<c_ushort> {
```

This function name made sense in Chromium, where they're explicit about their virtual keycodes being the same as what Windows has. While winit's `VirtualKeyCode` is also based on Windows keycodes, we 1) don't enforce that, and 2) most people don't know they're originally from Windows. So, a different function name would be easier to understand here. The name I entered in is just a suggestion, so feel free to go with something different.

Great job writing comments! For something inobvious like this, we usually also add [a link](https://github.com/WebKit/webkit/blob/82bae82cf0f329dbe21059ef0986c4e92fea4ba6/Source/WebCore/platform/cocoa/KeyEventCocoa.mm#L626) to whatever we referenced to learn how to implement it.

src/platform/macos/events_loop.rs
```diff
@@ -544,6 +544,67 @@ impl Proxy {
...
+    // We only translate keys that are affected by keyboard layout.
+    //
+    // Note that since keys are translated in a somewhat "dumb" way (reading character)
+    // there is a concern that some combination, i.e. Cmd+char, causes the wrong
+    // letter to be received, and so we receive the wrong key.
```
francesca64: Have you found any cases where this actually happens?

src/platform/macos/events_loop.rs
```diff
@@ -544,6 +544,67 @@ impl Proxy {
...
+    //
+    // Codes taken from to_virtual_key_code.
+    // It may be better to map these codes to something more immediately meaningful.
```
francesca64: Is there any reason not to just have this function directly return `Option<VirtualKeyCode>`? Remapping the scancodes is a bit unintuitive.

src/platform/macos/events_loop.rs
```diff
@@ -544,6 +544,67 @@ impl Proxy {
...
+        '/' | '?' => 0x2c,
+        'n' | 'N' => 0x2d,
+        'm' | 'M' => 0x2e,
+        '.' | '>' => 0x2f,
+        '`' | '~' => 0x32,
+        _ => return None,
+    })
+}
+
 pub fn to_virtual_key_code(code: c_ushort) -> Option<events::VirtualKeyCode> {
```
francesca64:
```suggestion
pub fn scancode_to_keycode(code: c_ushort) -> Option<events::VirtualKeyCode> {
```
It would be good to give this function a name that makes the difference explicit between it and the char-based function. Once again, feel free to try a different naming pattern than what I suggested.

`check_additional_virtual_key_codes` should probably also be renamed to be 1) consistent with whatever names the other ones will have, and 2) make the growing number of keycode functions less confusing... that could just be `check_function_keys`, or something like that.

src/platform/macos/view.rs
```diff
@@ -412,15 +450,16 @@ extern fn key_down(this: &Object, _sel: Sel, event: id) {
...
         // can vary, but we know that they are encoded
         // in characters property.
+        let keycode = to_translated_macos_scancode(event);
         let virtual_keycode = to_virtual_key_code(keycode)
             .or_else(|| {
                 check_additional_virtual_key_codes(&state.raw_characters)
             });
-        let scancode = keycode as u32;
+
+        let scancode: u32 = msg_send![event, keyCode];
```
francesca64: I think it would be good to put all the `keyCode` calls into a function called `get_scancode` or something, and then have comments explicitly noting that `keyCode` returns the scancode. That way, there won't be much room for confusion in the future (the reason this bug never got caught is largely due to how AppKit's usage of the label `keyCode` differs from ours).

src/platform/macos/view.rs
```diff
@@ -436,6 +475,7 @@ extern fn key_down(this: &Object, _sel: Sel, event: id) {
             },
         };
 
+        // NOTE: this code is run twice.
```
francesca64: Why is this code run twice?

src/platform/macos/view.rs
```diff
@@ -403,6 +403,44 @@ fn get_characters(event: id) -> Option<String> {
...
+    }
+}
+
+// Translate macos scancode into the corresponding windows key code.
+fn to_translated_macos_scancode(event: id) -> c_ushort {
+    // Cmd switches Roman letters for Dvorak-QWERTY layout, so try modified characters first.
+    let characters = get_characters(event);
+    if let Some(chars) = characters {
+        let code = chars.chars().next().map_or(0, |c| windows_key_code_for_char_code(c).unwrap_or(0));
+        if code != 0 {
```
francesca64: This pattern (using `0` as a magic constant) is super unidiomatic, so it would be way nicer to use `Option` throughout this function. You can use [`and_then`](https://doc.rust-lang.org/core/option/enum.Option.html#method.and_then) to chain the result of the first attempt to get a keycode with the result of the second attempt, and then use [`unwrap_or_else`](https://doc.rust-lang.org/core/option/enum.Option.html#method.unwrap_or_else) to default to the scancode at the end (though I'm not sure how relevant this advice will be after changing `windows_key_code_for_char_code` to return `Option<VirtualKeyCode>`).

There's also some repetition in this function that can be simplified out to make things easier to understand (maybe just using an inline function?). That will make it easier to change this into an `Option` chain, too.

src/platform/macos/view.rs
```diff
@@ -403,6 +403,44 @@ fn get_characters(event: id) -> Option<String> {
...
+            return code;
+        }
+    }
+
+    // Translate characters that weren't caught in the above operation.
```
francesca64: Are there any examples of when this is triggered?

src/platform/macos/view.rs
```diff
@@ -403,6 +403,44 @@ fn get_characters(event: id) -> Option<String> {
...
+    let raw_characters = get_characters_ignoring_modifiers(event);
+    if let Some(chars) = raw_characters {
+        let code = chars.chars().next().map_or(0, |c| windows_key_code_for_char_code(c).unwrap_or(0));
+        if code != 0 {
+            return code;
+        }
+    }
+
+    unsafe {
+        // We've checked all layout-related keys, so the fall-through to scancode.
```
francesca64:
```suggestion
        // We've checked all layout-related keys, so we fall-through to scancode.
```

src/platform/macos/view.rs
```diff
@@ -403,6 +403,44 @@ fn get_characters(event: id) -> Option<String> {
     }
 }
 
+fn get_characters_ignoring_modifiers(event: id) -> Option<String> {
```
francesca64: It would be neater to just add an `ignore_modifiers` flag to `get_characters`, since that results in less unsafe code to audit.

Toqozz:
Thanks for the in-depth reply!  I'll get to work on these issues now.

src/platform/macos/events_loop.rs
Toqozz: No, I haven't been able to personally observe this.

My concern is in things like these: https://www.webnots.com/option-or-alt-key-shortcuts-to-insert-symbols-in-mac-os-x/.  If one of those codes resulted in a match for `char_to_keycode()` then the key could potentially be incorrect.  This doesn't seem to happen right now, but I don't think it's far fetched to be concerned about it -- what if these codes become user-configurable?
I'm also not sure how those codes work in languages other than english, but maybe I'm going out of scope there.

src/platform/macos/events_loop.rs
Toqozz: The original thinking was so that we can more easily "fall through" to `to_virtual_key_code()` -- the last stop to get a keycode, which must be passed through.  I realise now that it may be better (and more concise) to simply do all checking inside `to_translated_macos_scancode()` (which would then be renamed) and return a virtual keycode from there.

src/platform/macos/view.rs
Toqozz: IMO it doesn't have a good reason to be run twice.  The ~10 lines below the comment here seem to be lifted entirely from the `get_characters()` method, which is used to set `state.raw_characters` earlier.  My best guess is that this has happened as a somewhat lazy way to bring `string` into scope without dealing with `Option`/ownership.
Also, does `get_characters()` need to return an `Option` when it only ever seems to return `Some()`?

I'm not sure on the etiquette here.  I wanted to fix this but wasn't entirely sure what the best way to go about it would be.  Thoughts?

src/platform/macos/view.rs
Toqozz: Yes.  Mostly keys such as backspace and enter.  I'll elaborate with these in the comment.

src/platform/macos/events_loop.rs
Toqozz: Thanks for the reasoning.  How would you suggest I link to the reference?

Is
```
// Implementation reference: https://github.com/WebKit/webkit/blob/82bae82cf0f329dbe21059ef0986
    c4e92fea4ba6/Source/WebCore/platform/cocoa/KeyEventCocoa.mm#L626
```
enough?

src/platform/macos/events_loop.rs
Toqozz: Would it make sense to remove the `check_additional_virtual_key_codes()` function and just move its checks to `char_to_keycode()`?  This would avoid doubling up on string checks (both functions take virtually the same input) but loses the fact that `char_to_keycode()` only checks layout related keys.

I don't have a keyboard with those function keys to test anything, and it's probably not worthwhile, so I'll leave that TBD.

francesca64: The only complication is that the function keys use UTF-16, but I'm definitely in favor of merging them ***if*** it makes things cleaner.

I'm not really able to test those keys either, and neither was the person who implemented it IIRC. It's straightforward enough that I'm not too worried, though.

src/platform/macos/events_loop.rs
francesca64: Yeah, that works! If in doubt, there's a bunch of examples: https://github.com/tomaka/winit/search?l=Rust&q=https

src/platform/macos/events_loop.rs
francesca64: Ah, that makes sense.

> but maybe I'm going out of scope there.

Supporting other languages is very much in scope. From a cursory look, the Hiragana input method uses the same altcodes as the U.S. input method. On 390 Sebulshik, there don't seem to be any altcodes at all... it's probably worth checking out [`InputMethodKit`](https://developer.apple.com/documentation/inputmethodkit?language=objc) to see if there's an API provided for specifying altcodes, since if not, I'm inclined to assume there's only one set of possible values.

Either way, I'm not too worried about this. It doesn't seem likely that Apple will make them user-configurable any time soon, though that's just my gut feeling on this.

src/platform/macos/view.rs
francesca64: Oh, I see. I was confused and thought the comment was about a change you made.

Yeah, this redundancy is definitely silly. Also, IIRC `get_characters` came later: it was something I nagged someone into making to simplify their PR, but I didn't nag them into using it everywhere... there's a lot of room for refactoring here, and it would be very much appreciated!

Um, I can't remember any reason for `get_characters` returning an `Option`, so that's pretty funny. While we could keep it this way and check for `nil`, it's probably fine to just use `assert_ne!(characters, nil)` and then always return `String`.

src/platform/macos/events_loop.rs
Toqozz: When I press my F keys I get characters like `"\u{f703}"`.  It seems like we could match against these for those keys and avoid UTF-16.  Or is there a specific reason for the larger encoding size?

francesca64: The fact that the docs are explicit about this: https://developer.apple.com/documentation/appkit/1535851-function-key_unicodes?preferredLanguage=occ

Toqozz:
I've applied changes which (I believe) resolve most of what we've talked about here.  There were some minor details that I was uncertain about (i.e. where to put `get_scancode()`), but these should be able to be easily resolved.

I was quite busy last week -- hence the delay -- but I should be able to resolve anything more quite quickly, so please let me know.

Thanks :)

src/platform/macos/events_loop.rs
Toqozz: I've decided to leave `check_additional_virtual_key_codes()` (now `check_function_keys()`) as it is for now, minus the change from `Option<&String>` to `&String` to align with `get_characters()`.  

I don't think it would make things any cleaner and makes the `char_to_keycode()` code less purposeful (imo).

chrisduerr:
I just wanted to chime in and say that I've been using Alacritty with your winit branch for a while (on macOS) and it seems to be working great, haven't run into any issues.

chrisduerr:
I don't think the remaining two 'TODO's are relevant for this PR so it might be a bit misleading to keep them in. Build also is just failing because there's a merge conflict in the CHANGELOG file.

Is there anything left to be done other than fixing the merge conflict? The functionality seems good to go and it would be a shame to forget about a QoL change like this.

Toqozz:
What are the remaining two TODO's that you refer to?  Did I miss something?

I would definitely still like to push this; things can be very annoying/confusing for non-qwerty users without it.  I might even consider it crucial, since so many projects rely on winit.

chrisduerr:
@Toqozz I'm talking about the list in the original comment of this PR. This is part of winit's PR template, but unfortunately if you look at the list of PRs, it shows up as '2 of 4' tasks done. So that can easily confuse maintainers to think there's still progress the author needs to complete.

francesca64:
Sorry for the delay! Thanks so much for this great PR!

PR #759, opened 2019-01-14
Redox support (legacy)

jackpot51:

### DIFF
--- a/src/os/redox.rs
+++ b/src/os/redox.rs
@@ -1,0 +1,15 @@
+#![cfg(target_os = "redox")]
+
+use std::sync::{Arc, Mutex};
+
+use Window;
+
+pub trait WindowExt {
+    fn get_orbclient_window(&self) -> Arc<Mutex<orbclient::Window>>;
+}
+
+impl WindowExt for Window {
+    fn get_orbclient_window(&self) -> Arc<Mutex<orbclient::Window>> {
+        self.window.get_orbclient_window()
+    }
+}
--- a/src/platform/redox/mod.rs
+++ b/src/platform/redox/mod.rs
@@ -1,0 +1,540 @@
+use orbclient::{EventOption, Renderer};
+use std::collections::VecDeque;
+use std::fmt;
+use std::sync::{Arc, Mutex};
+
+use {CreationError, MouseCursor, WindowAttributes, VirtualKeyCode};
+use dpi::{LogicalPosition, LogicalSize, PhysicalPosition, PhysicalSize};
+use window::MonitorId as RootMonitorId;
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct DeviceId;
+
+fn convert_scancode(scancode: u8) -> Option<VirtualKeyCode> {
+    match scancode {
+        orbclient::K_A => Some(VirtualKeyCode::A),
+        orbclient::K_B => Some(VirtualKeyCode::B),
+        orbclient::K_C => Some(VirtualKeyCode::C),
+        orbclient::K_D => Some(VirtualKeyCode::D),
+        orbclient::K_E => Some(VirtualKeyCode::E),
+        orbclient::K_F => Some(VirtualKeyCode::F),
+        orbclient::K_G => Some(VirtualKeyCode::G),
+        orbclient::K_H => Some(VirtualKeyCode::H),
+        orbclient::K_I => Some(VirtualKeyCode::I),
+        orbclient::K_J => Some(VirtualKeyCode::J),
+        orbclient::K_K => Some(VirtualKeyCode::K),
+        orbclient::K_L => Some(VirtualKeyCode::L),
+        orbclient::K_M => Some(VirtualKeyCode::M),
+        orbclient::K_N => Some(VirtualKeyCode::N),
+        orbclient::K_O => Some(VirtualKeyCode::O),
+        orbclient::K_P => Some(VirtualKeyCode::P),
+        orbclient::K_Q => Some(VirtualKeyCode::Q),
+        orbclient::K_R => Some(VirtualKeyCode::R),
+        orbclient::K_S => Some(VirtualKeyCode::S),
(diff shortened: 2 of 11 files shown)


vberger:
Wow, thanks for this effort!

Before nitpicking the review, I have some larger questions:

- several functions are mocked-out in this PR
  - I suppose the ones for which you changed the documentation noting them as "doing nothing", is because they map to some functionality that is not provided by Redox?
  - Regarding the ones noted as "TODO", is it because you want to leave them as a future PR, or because they are waiting for implementation work in Redox?
- Are you additionally offering to maintain this code in the future?

jackpot51:
Thanks!

I could have been a bit more clear. The functions like set_always_on_top that are marked TODO have support in the underlying window manager (orbital), as demonstrated in the Window::new function, but that support is not exposed yet in orbclient. The ones marked N/A are not supported by orbital, and may never be. I suppose it would be helpful to add more information about each stub. Where applicable, I added notes in the documentation about functions that do not currently function on Redox.

Yes, I am offering to maintain the Redox support for winit. I could add my name to the README if you like.

vberger:
Thanks for the clarification, I'll take some time to review this properly, hopefully tomorrow.

In the meantime, you can add a new volumn for redox in the table at the bottom of CONTRIBUTING.md and a line with your name, that'd be perfect.

jackpot51:
Finished that. I also added more information about each TODO

Osspial:
One nitpick that I have - it doesn't look like this is getting tested on our automated testing system. Since Travis doesn't support Redox we clearly can't get a native compiler going on there, but it should be possible to set up a configuration that cross-compiles to Redox and makes sure everything builds.

jackpot51:
It would be very difficult to set up qemu with Redox right now. Automatic cross compilation should be easy

jackpot51:
I added CI support for Redox. Since a cross compiler is required to do cargo build, complicated to set up, and also would not catch many more errors than cargo check, I have made the Redox test use cargo check.

src/os/redox.rs
```diff
@@ -0,0 +1,15 @@
...
+
+pub trait WindowExt {
+    fn get_orbclient_window(&self) -> Arc<Mutex<orbclient::Window>>;
+}
+
+impl WindowExt for Window {
+    fn get_orbclient_window(&self) -> Arc<Mutex<orbclient::Window>> {
+        self.window.get_orbclient_window()
+    }
+}
```
vberger: This means that `orbclient` becomes a public dependency of `winit`. Given `orbclient` is the official API of Redox (I assume it is at least?), I'd tend to be okay with that.

However this means that any breaking change to `orbclient` means a breaking release of `winit` to update it, which is problematic. For this reason `winit` has tried to reduce public dependencies to a minimum.

So, what is the stability story of `orbclient`? What do you think @Osspial @francesca64 ?

src/platform/redox/mod.rs
```diff
@@ -0,0 +1,554 @@
...
+                    _ => ()
+                }
+            }
+        }
+    }
+
+    pub fn run_forever<F>(&mut self, mut callback: F)
+        where F: FnMut(::Event) -> ::ControlFlow,
+    {
+        // Yeah that's a very bad implementation.
```
vberger: How will it be possible to improve it in the future?

src/platform/redox/mod.rs
```diff
@@ -0,0 +1,554 @@
...
+        EventsLoopProxy
+    }
+}
+
+#[derive(Clone)]
+pub struct EventsLoopProxy;
+
+impl EventsLoopProxy {
+    pub fn wakeup(&self) -> Result<(), ::EventsLoopClosed> {
+        unimplemented!()
```
vberger: Is it not possible to implement this currently?

src/platform/redox/mod.rs
jackpot51: Not sure, I copied this specific function from the iOS module

src/platform/redox/mod.rs
jackpot51: I was unsure what exactly this is supposed to do. The simpler modules like ios and android do not implement it, and the more complicated modules like windows and x11/wayland are very confusing to read.

With a little more information I am sure I could implement it.

src/platform/redox/mod.rs
vberger: Oh ok.

I guess this is related to the use of `thread::sleep()`, likely because `self.poll_events()` is non-blocking on iOS. If this methods is a blocking call on Redox, the `sleep` call can likely be removed, and with that I don't see any obvious way why the implementation would be "bad".

jackpot51: The methods are non-blocking because the window is opened with the Async WindowFlag

src/platform/redox/mod.rs
jackpot51: Ok. I will work on implementing it

src/platform/redox/mod.rs
vberger: Well in that case, the implementation could be improved by blocking (for example the wayland backend uses epoll to monitor the wayland socket) rather than loop-sleeping.

src/os/redox.rs
francesca64: I'm fine with this. There hasn't been an incompatible change in almost 2 years: https://crates.io/crates/orbclient/versions

Osspial:
This has been sitting here, untouched, for the past six months. What's the status on it?

jackpot51:
@Osspial I have not yet fixed the remaining comment about EventsLoopProxy. If this does not merge against master anymore, I will rebase it.

Osspial:
@jackpot51 Actually, we have merged a completely reorganized public API onto master. It'll take some reworking to get this functioning again...

jackpot51:
@Osspial I see. This is still critical for me, so when I can, I will add Redox support to the master version. For now, you could re-review this for the legacy version if it interests you.

ZeGentzy:
@jackpot51 Hey, this has gone untouched for over a month now. Could you rebase this against master and open a new PR. I'm closing this for now. Thanks.

PR #766, opened 2019-01-18
Set ime spot with logical position

bet4it:

It seems that we should send xim with the logical position rather than physical position.

Before this commit:
ime_spot_before

After this commit:
ime_spot_after


bet4it:
This was a bug of alacritty and has been fixed in  jwilm/alacritty#2065

PR #770, opened 2019-01-25
On wayland, fix `with_title()` not setting the windows title

trimental:

Wayland windows currently did not set their title when built with the `with_title()` function.


rkanati:
This is a one-line fix, and I can confirm it fixes the problem. Please merge.

PR #772, opened 2019-01-25
Fix second window on Linux

seppo0010:
I ran into a problem when having two open windows on Linux. On Windows
it worked just fine.

This was the stacktrace reported by valgrind

```
==10814== Invalid read of size 8
==10814==    at 0x6E4ADD4: XCloseIM (in /usr/lib/x86_64-linux-gnu/libX11.so.6.3.0)
==10814==    by 0x17F1EB0: winit::platform::platform::x11::ime::inner::close_im::hbc9884d53ebb135d (inner.rs:12)
==10814==    by 0x17F2165: winit::platform::platform::x11::ime::inner::ImeInner::close_im_if_necessary::h66b984514f534c73 (inner.rs:53)
==10814==    by 0x1795FD6: winit::platform::platform::x11::ime::callbacks::replace_im::h24ec173b1f14fafb (callbacks.rs:132)
==10814==    by 0x1796457: winit::platform::platform::x11::ime::callbacks::xim_instantiate_callback::h7d468b9bb7725d0d (callbacks.rs:149)
==10814==    by 0x6E63D24: _XimRegisterIMInstantiateCallback (in /usr/lib/x86_64-linux-gnu/libX11.so.6.3.0)
==10814==    by 0x6E4AE8B: XRegisterIMInstantiateCallback (in /usr/lib/x86_64-linux-gnu/libX11.so.6.3.0)
==10814==    by 0x17953A1: winit::platform::platform::x11::ime::callbacks::set_instantiate_callback::hc2cfe1836f29d39a (callbacks.rs:39)
==10814==    by 0x1797573: winit::platform::platform::x11::ime::Ime::new::_$u7b$$u7b$closure$u7d$$u7d$::h29e3c3099f3d0bd3 (mod.rs:57)
==10814==    by 0x17FDBD2: core::ops::function::impls::_$LT$impl$u20$core..ops..function..Fn$LT$A$GT$$u20$for$u20$$RF$F$GT$::call::h7296aa50456f6259 (function.rs:247)
==10814==    by 0x17F3FC2: winit::platform::platform::x11::ime::input_method::PotentialInputMethods::open_im::h09a3f2c789e7c5e2 (input_method.rs:269)
==10814==    by 0x1796AE9: winit::platform::platform::x11::ime::Ime::new::h7ab2dd81306c8874 (mod.rs:56)
==10814==  Address 0x0 is not stack'd, malloc'd or (recently) free'd
```

I added a guard against null and my code now works fine.

Tested on Linux, Ubuntu 16.04.

### DIFF
--- a/src/platform/linux/x11/ime/inner.rs
+++ b/src/platform/linux/x11/ime/inner.rs
@@ -49,7 +49,7 @@ impl ImeInner {
     }
 
     pub unsafe fn close_im_if_necessary(&self) -> Result<bool, XError> {
-        if !self.is_destroyed {
+        if !self.is_destroyed && self.im != ptr::null_mut() {
             close_im(&self.xconn, self.im).map(|_| true)
         } else {
             Ok(false)


mitchmindtree:
Can you offer anymore information on how to reproduce this error? E.g. does this happen when running the `multiwindow.rs` example?

I'm currently developing multiple multi-window installations (one with 4 windows) on Ubuntu 18.04 LTS and I'm yet to run into this error.

seppo0010:
Hi @mitchmindtree. I'm sorry I did not provide more details. The multiwindow example does work, unfortunately.

I'm using nphysics, nphysics-testbed and amethyst, so I have too many layers to debug what's wrong since both of those libraries use winit themselves.

If you wanna take a look at my crash you can clone this repo, in the merge-objects branch: https://github.com/seppo0010/building-stuff/tree/merge-objects and then run `cargo run --features "testbed"`.

If you want to run it with valgrind as I did, you will have to run `cargo build --features "testbed" && cp target/debug/building-stuff . && valgrind ./building-stuff`.

francesca64:
I think this should also log a `warn!` if the pointer's null, since it's not something that's supposed to happen (the only reason I'm not asking for an `assert!` is because IM isn't important for many users)

(A CHANGELOG entry would be nice too)

tdaffin:
I ran into this problem while trying to track down a bug in Piston/conrod.
I've posted a reproducible example here:
https://github.com/tdaffin/conrod_bug/tree/seg_fault

ZeGentzy:
Is the SEGV caused by this bug, or does it just so happen to also reproduce this bug? If the former, can you file it separately? I'll take a look at it soon(tm).

tdaffin:
The segv in my example occurred when I was trying to create two windows at once in order to demonstrate some other bug, but the segv is not related to the other bug.  Also, the segv happens in a very similar call stack as reported here.
I may also be useful if I note that when I applied the patch here to a local copy of winit the crash did not occur anymore but the two windows did not really behave as expected..

ZeGentzy:
Apologizes, but this PR has fallen out of date now, and I didn't grab the time to test it before it did.
Can someone rebase this to master w/ [francesca's suggested changes](https://github.com/rust-windowing/winit/pull/772#issuecomment-466681664) then open a new PR. That way I can ping the appropriate testers. Thanks. Anyways, closing this issue.

PR #775, opened 2019-01-26
Wayland: add `set_wayland_theme()` to control client decoration color…

trimental:

I've created a new function `set_wayland_theme()` in `winit::os::unix::WindowExt` that takes a struct of `winit::os::unix::WaylandTheme` to allow for theming of the client side wayland decorations.

Using `WaylandTheme` allows for us to not expose SCTKs internal data types or traits

Fixes https://github.com/tomaka/winit/issues/765

### DIFF
--- a/src/os/unix.rs
+++ b/src/os/unix.rs
@@ -4,6 +4,8 @@ use std::os::raw;
 use std::ptr;
 use std::sync::Arc;
 
+use sctk::window::{ButtonState, Theme};
+
 use {
     EventsLoop,
     LogicalSize,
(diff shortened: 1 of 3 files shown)


src/platform/linux/wayland/window.rs
```diff
@@ -235,6 +235,11 @@ impl Window {
         }
     }
 
+
+    pub fn set_theme<T: Theme>(&mut self, theme: T) {
```
vberger: Given there is locking inside, I don't think this method needs to take `&mut self`, `&self` should be enough.

src/os/unix.rs
```diff
@@ -25,6 +27,74 @@ pub use platform::x11;
 pub use platform::XNotSupported;
 pub use platform::x11::util::WindowType as XWindowType;
 
+/// Theme for wayland client side decorations
+///
+/// Colors must be in native endian ARGB8888 format
```
vberger: Given endianness is handled by SCTK, this comment would probably be clearer without talking about endianness.

trimental: Yeah, I forgot SCTK now handles it. Also the comments wrong, the colors must only be in big endian anyway.

PR #780, opened 2019-01-27
Add wayland specific code to examples for windows to appear

trimental:

Adds wayland specific code to the examples to allow for the windows to appear.

Once a version of SCTK with https://github.com/Smithay/client-toolkit/commit/74316f4d71f21ae5b53339ee39b8b0249341dbae is released the git dependency can be replaced to that.

Fixes https://github.com/tomaka/winit/issues/776

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,7 +63,7 @@ features = [
 
 [target.'cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "openbsd", target_os = "netbsd"))'.dependencies]
 wayland-client = { version = "0.21", features = [ "dlopen", "egl", "cursor"] }
-smithay-client-toolkit = "0.4.3"
+smithay-client-toolkit = { git = "https://github.com/smithay/client-toolkit" }
 x11-dl = "2.18.3"
 parking_lot = "0.7"
 percent-encoding = "1.0"
--- a/examples/helpers/mod.rs
+++ b/examples/helpers/mod.rs
@@ -1,0 +1,45 @@
+#[cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "netbsd", target_os = "openbsd"))]
+extern crate smithay_client_toolkit as sctk;
+
+#[cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "netbsd", target_os = "openbsd"))]
+use winit::os::unix::WindowExt;
+
+// Wayland requires the commiting of a surface to display a window
+pub fn init_wayland(window: &winit::Window) {
+    #[cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "netbsd", target_os = "openbsd"))]
+    {
+        if let Some(winit_display) = window.get_wayland_display() {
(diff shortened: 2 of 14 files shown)


trimental:
Noted that no pixels are drawn to the screen so there can be the effect of a transparent window. I didn't include drawing due to extra code overhead but that can easily be added if required.

vberger:
Given this code repeats itself a lot, I'd rather have it in an helper module that each example imports.

So, a possibility would be to have a `helpers` folder in the examples, and this code in `helpers/mod.rs` which is then imported by the examples. (the subfolder is necessary so that cargo does not try to interpret this file as an example to compile)

trimental:
Yes that sounds like a better implementation. I still haven't included the drawing though as that would mean handling the resize event on each example to redraw the window.

Cargo.toml
```diff
@@ -63,7 +63,7 @@ features = [
 
 [target.'cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "openbsd", target_os = "netbsd"))'.dependencies]
 wayland-client = { version = "0.21", features = [ "dlopen", "egl", "cursor"] }
-smithay-client-toolkit = "0.4.3"
+smithay-client-toolkit = { git = "https://github.com/smithay/client-toolkit" }
```
Osspial: Why is the `smithay-client-toolkit` dependency changed to refer to the `git` version instead of the `crates.io` version? I'm pretty sure will prevent us from pushing this update to `crates.io`.

examples/helpers/mod.rs
```diff
@@ -0,0 +1,45 @@
+#[cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "netbsd", target_os = "openbsd"))]
+extern crate smithay_client_toolkit as sctk;
+
+#[cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "netbsd", target_os = "openbsd"))]
+use winit::os::unix::WindowExt;
+
+// Wayland requires the commiting of a surface to display a window
+pub fn init_wayland(window: &winit::Window) {
```
Osspial: This is emitting `unused_variables` warnings on non-linux platforms. Could you add a `let _ = window;` line somewhere to suppress that warning?

Cargo.toml
vberger: This requires changes from SCTK that are not yet released, though given how things are coming, next version of SCTK will be a breaking change (upgrade to wayland-client 0.22).

I plan to make the upgrade as part of the event loop 2.0 refactor, so I'd say this PR should wait until the eventloop 2.0 story is done.

examples/helpers/mod.rs
vberger: For clarity, I'd actually rather move the `#[cfg(...)]` in the call site in the examples, so that people can clearly see that this code is for some platform-specific stuff.

Cargo.toml
Osspial: Oh wait, that was... written in original PR comment. Whoops. I guess I should read everything before making these sorts of comments :sweat_smile:.

Osspial:
Is this PR necessary, now that #835 has been merged?

trimental:
Nope :+1:

PR #782, opened 2019-01-27
x11: thread safe replacement for XNextEvent

### LINKED ISSUE #779: Using `XNextEvent` in `EventsLoop` is not thread safe
The `EventsLoop` on X11 backend uses `XNextEvent` to poll and wait for events which is not thread safe.

`XNextEvent` will hold the global Xlib mutex while blocked waiting for input, which causes a deadlock when other threads try to use X11, e.g. when calling `make_current` for a GL context.

The workaround is to use `select` or `poll` to wait for input on the X11 socket file descriptor and then use a non-blocking function like `XCheckIfEvent` to grab the events from the event queue. This is a rather ugly workaround to an ugly Xlib limitation.

Here is my work-in-progress fix to this problem: https://github.com/rikusalminen/winit/tree/waitevent

~This patch is not quite complete and I'm seeing some strange issues with `CloseRequested` events which I can't quite explain. Some further debugging is required. Any help with testing and debugging is welcome.~ edit: fixed!

~This bug also introduces an error handling conondrum. `XNextEvent` and other X11 functions do not return an error code, but `poll` will return error codes, some of which may be fatal errors. `EventsLoop::runForever` does not return a `Result` or other error code. I don't know what the right solution is (which would be compatible with other backend). Please advice how to proceed with this.~ edit: solved with `assert!` and `panic!` for fatal error conditions.

Additionally, a multi-threaded example/test case would be useful to verify that this problem is solved.

This issue is the root cause in several open issues for winit.  At least the following bugs look like they are caused by this: #757, #420, #239 .

NOTE: I also fixed this same issue in GLFW back in 2012:
https://github.com/glfw/glfw/commit/3d6221c490fab8bf70ba527430a2518c821c1027#diff-0f3106b1408d89cbecbe2f57a783331b

rikusalminen:
XNextEvent will block for input while holding the global Xlib mutex.

This will cause a deadlock in even the most trivial multi-threaded
application because OpenGL functions will need to hold the Xlib mutex
too.

Add EventsLoop::poll_one_event and EventsLoop::wait_for_input to provide
thread-safe functions to poll and wait events from the X11 event queue
using unix select(2) and XCheckIfEvent.

This is a somewhat ugly workaround to an ugly problem.

Fixes #779

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - Added serde serialization to `os::unix::XWindowType`.
 - **Breaking:** `image` crate upgraded to 0.21. This is exposed as part of the `icon_loading` API.
+- On X11, make event loop thread safe by replacing XNextEvent with select(2) and XCheckIfEvent
 
 # Version 0.18.1 (2018-12-30)
 
--- a/src/platform/linux/x11/mod.rs
+++ b/src/platform/linux/x11/mod.rs
@@ -19,6 +19,7 @@ use std::collections::HashMap;
 use std::ffi::CStr;
 use std::ops::Deref;
 use std::os::raw::*;
+use libc::{select, fd_set, FD_SET, FD_ZERO, FD_ISSET, EINTR, EINVAL, ENOMEM, EBADF, __errno_location};
 use std::sync::{Arc, mpsc, Weak};
 use std::sync::atomic::{self, AtomicBool};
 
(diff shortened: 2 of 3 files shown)


jansol:
@francesca64 ping (fixed version of #781 )

vberger:
I tested some of the winit examples with this on x11, and it seems to work smoothly.

PR #784, opened 2019-01-31
Add monitor changed event

saschagrunert:
Hey, my idea was to add another event if the monitor name has changed. I implemented it first for X11, because I am not sure if the approach is the right one. WDYT?

### DIFF
--- a/src/events.rs
+++ b/src/events.rs
@@ -37,19 +37,19 @@ pub enum WindowEvent {
     Destroyed,
 
     /// A file has been dropped into the window.
-    /// 
+    ///
     /// When the user drops multiple files at once, this event will be emitted for each file
     /// separately.
     DroppedFile(PathBuf),
 
     /// A file is being hovered over the window.
-    /// 
+    ///
     /// When the user hovers multiple files at once, this event will be emitted for each file
     /// separately.
     HoveredFile(PathBuf),
 
     /// A file was hovered, but has exited the window.
-    /// 
+    ///
     /// There will be a single `HoveredFileCancelled` event triggered even if multiple files were
     /// hovered.
     HoveredFileCancelled,
@@ -115,6 +115,9 @@ pub enum WindowEvent {
     ///
     /// For more information about DPI in general, see the [`dpi`](dpi/index.html) module.
     HiDpiFactorChanged(f64),
+
+    /// The window has moved to another monitor
+    MonitorChanged(String),
 }
 
 /// Represents raw hardware events that are not associated with any particular window.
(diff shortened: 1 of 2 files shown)


src/events.rs
```diff
@@ -115,6 +115,9 @@ pub enum WindowEvent {
     ///
     /// For more information about DPI in general, see the [`dpi`](dpi/index.html) module.
     HiDpiFactorChanged(f64),
+
+    /// The window has moved to another monitor
+    MonitorChanged(Option<String>),
```
Osspial: Two points:
- The `MonitorChanged` event should contain a [`MonitorId`](https://docs.rs/winit/0.18.1/winit/struct.MonitorId.html), not a `String`.
- Why does `MonitorChanged` contain an `Option` instead of a raw value? Your code never emits a `None`,

Osspial:
Hi, and thanks for the PR!

I'd be down with adding this. The approach seems about right (at least from my limited knowledge of the X11 backend), but there are a couple of sticking points that came up which I've added review comments for.

I'd like to get some idea of what support looks like on other platforms before merging this, though. Off the top of my head Windows probably supports it and Wayland probably doesn't, but we'd need to look into how MacOS handles it. @francesca64

src/events.rs
saschagrunert: > Two points:
> 
> * The `MonitorChanged` event should contain a [`MonitorId`]
(https://docs.rs/winit/0.18.1/winit/struct.MonitorId.html), not a `String`.

This is currently not possible because MonitorId does not implement `PartialEq`. The main reason that this is not derivable is that [`sctk::output::OutputMgr`](https://docs.rs/smithay-client-toolkit/0.4.4/smithay_client_toolkit/output/struct.OutputMgr.html) does not implement PartialEq, too.

> * Why does `MonitorChanged` contain an `Option` instead of a raw value? Your code never emits a `None`,

I will go for a `String` for now, you're right.

Osspial: You should be able to implement `PartialEq` for `MonitorId` by comparing the backend's monitor ID - it looks like every platform has some method for retrieving uniquely identifying information. For Wayland, that would be `MonitorHandle::get_native_identifier`.

francesca64:
@Osspial this would be trivial to implement on macOS, so no worries there: https://github.com/tomaka/winit/blob/84c812e5689b2561fbcde8c07efa05ff0dc52ddb/src/platform/macos/window.rs#L261

The X11 implementation here looks good too.

ZeGentzy:
Hey, this hasn't had activity for a good amount of time, and looks like its fallen out of date. Can you rebase this against master, and submit a new PR. Anyways, closing this one, thanks.

PR #790, opened 2019-02-05
Wayland's eventloop 2.0

vberger:
This is the port of the wayland backend to events loop 2.0.

Notably:

- the X11-specific code in platform_impl / platform has been commented-out
- it also updates the wayland dependencies:
  - wayland-client to 0.22
  - SCTK to 0.5

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,9 +64,9 @@ features = [
 ]
 
 [target.'cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "openbsd", target_os = "netbsd"))'.dependencies]
-wayland-client = { version = "0.22.0", features = [ "dlopen", "egl", "cursor", "eventloop"] }
+wayland-client = { version = "0.23.0", features = [ "dlopen", "egl", "cursor", "eventloop"] }
 calloop = "0.4.2"
-smithay-client-toolkit = "0.5"
+smithay-client-toolkit = "0.6"
 x11-dl = "2.18.3"
 parking_lot = "0.7"
 percent-encoding = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,8 +64,9 @@ features = [
 ]
 
 [target.'cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "openbsd", target_os = "netbsd"))'.dependencies]
-wayland-client = { version = "0.21", features = [ "dlopen", "egl", "cursor"] }
-smithay-client-toolkit = "0.4.3"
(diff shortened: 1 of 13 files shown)


vberger:
No sure who'd be best to review this ? It's honestly a pretty boring diff, as the port to the new eventloop setup didn't require any deep refactor of the wayland code.

Most of the diff is:

- the changes in `wayland/event_loop.rs`, which are needed to match the new API, this includes the use of `calloop` to drive the events loop
- a lot of search&replace to match the new types of wayand-client 0.22
- the resolution of what I believe to be forgotten merge conflicts in #638

Osspial:
I read over the diff, and nothing in particular stood out to me as erroneous in your code (for what that's worth, as someone without much knowledge of the Linux backend).

However, there was one thing I noticed in the public cross-platform API that seems a bit misleading, but not technically unsafe - the `T` in `EventLoopProxy<T>` isn't marked as `'static`. It's impossible to create an `EventLoopProxy` where `T` *isn't* `'static`, but I must have looked over adding the trait bound when I created it so we should probably add that in.

vberger:
Do you want it hadded to this PR, or can we merge it and do it in a second time?

Osspial:
Could you change the API here and add in the bounds on Linux? I'll push another commit that gets it working on Windows after this has been merged.

vberger:
Alright, will do that

vberger:
This should do it, I also added the bounds in the windows code, it seemed simple enough.

Osspial:
It looks like one of the tests was broken, but that should be fixed now. Once the CI finishes this should be ready to merge.

Osspial:
Actually, I went ahead and tested this on my Linux machine and a couple of points came up:

* I pulled in some of the work from #780 to my own branch to do tests, but even with that code a window didn't appear. I needed to make some modifications to the helper script to get it working so this may be an issue on my end, but you should check the code [here](https://github.com/Osspial/winit/tree/evl2_x) to make sure this isn't an issue with the backend.
* There's some special-casing of the `RedrawRequested` event that I'm not sure is being handled - in some cases, it can be sent after `EventsCleared` but before `NewEvents` (see [here](https://github.com/tomaka/winit/blob/eventloop-2.0/src/window.rs#L366-L370)). Is that behavior implemented? I didn't notice it, but I could also just be blind.

vberger:
I added the proper code to handle `RedrawRequested` special cases.

However I'm not clear on what you did actually change to make the window appear? I does display a window with transparent contents for me.

vberger:
This should be good now, are you ok with that @Osspial ?

Osspial:
@vberger  :+1:

vberger:
Alright, so I'll merge this into the `eventloop-2.0` branch, so that the x11 work can be done on top of it.

PR #792, opened 2019-02-13
Handle horizontal wheel input (Windows)

torkeldanielsson:
- [-] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior
- [-] Created an example program if it would help users understand this functionality

CHANGELOG, documentation and example should not be needed for this one? Or let me know if you think otherwise and I will fix it :)


Osspial:
Hi, and thanks for the PR!

You shouldn't need to add any documentation or examples, but since this is an improvement of functionality on Windows you should add a CHANGELOG entry. Other than that, this looks good!

torkeldanielsson:
Ok, added a note about it! :)

Issue #696, opened 2018-11-10
Discussion: mobile and web Window creation APIs


Osspial:
Right now, Winit exposes window creation and modification functionality the same way on all platforms. Is that something that makes sense? On mobile, only one function in `WindowBuilder` (`with_multitouch`) actually does anything, and only three functions in `Window` (`get_hidpi_factor`, `get_inner_size`, and `get_outer_size`) do anything useful. A few more functions are implemented for Emscripten, but it has much more in common with mobile platforms than it does with desktop platforms.

Should we expose an entirely different window structure for these platforms that better reflects the underlying capabilities? I don't have much experience with those platforms, but I've written up an example of what I think that sort of API would look like:

* Bikeshedding aside, I'm calling it `View`.
* For creation, `View` exposes `list_views(&EventsLoop) -> impl Iterator<View>`, which gets an iterator over all windows the OS provides.
* `View` exposes the following methods:
  * `set_multitouch(&self, bool)`: enables or disables multitouch support.
  * `get_hidpi_factor(&self) -> f32`: gets the HiDPI factor.
  * `get_size(&self) -> LogicalSize`: gets the window's size.
* Winit adds two new public platform modules:
  * `mobile`
    * Exposes extension trait `ViewExt` with the following methods:
      * `set_valid_orientations(ValidScreenOrientations)`: used to set the valid screen orientations.
    * Adds `ValidScreenOrientations` struct, defined as followed:
      ```rust
      pub struct ValidScreenOrientations {
          pub portrait: bool,
          pub landscape: bool
      }
      ```
  * `emscripten`
    * Exposes extension trait `ViewExt` with the following methods:
      * `set_fullscreen(bool)`: sets if the window is fullscreen or not
      * `grab_cursor(bool)`: grabs or ungrabs the cursor
      * `hide_cursor(bool)`: hides or shows the cursor

That design may be infeasible one way or another, so any API corrections are appreciated.

Relevant to #33 
cc @francesca64 @seivan @mtak- @willglynn

EDIT: Renamed `WindowBorrowed` to `View` since that's a better name for it, and added a reference to `EventsLoop` in `list_views`.

mtak-:
I completely understand the motivation for this, and I'll provide more feedback on this issue once I've cobbled together an EventLoop2.0 implementation for iOS.

Some early feedback:
[...]

PR #797, opened 2019-02-20
stdweb support for eventloop 2.0

ryanisaacg:
Unresolved design questions:
- [x] Should multiple canvases be supported? I'm leaning no
- [x] Should fullscreen be supported? If it should be, should it be the size of the window or some attempt to make the canvas actually fullscreen?
- [x] Is the window icon synonymous with the favicon? I'm leaning yes
- [x] Should file dropping be supported? It's a JS `File` object which isn't as useful as it is on desktop
- [x] Should show / hide be supported?
- [ ] How should refresh rate (opaque to a web page) be handled by VideoMode?

Implementation:
- [x] Creating a canvas / setting window title
- [x] DOM -> winit event binding support
- [x] Event cause
- [x] ControlFlow handling
    - [x] Poll
    - [x] Wait
    - [x] WaitUntil
    - [x] Exit
- [x] Pointer events
- [x] Keyboard events
- [x] Requesting redraws
- [x] Canvas positioning within the window
- [x] KeyboardEvent::which PR to stdweb
- [ ] Pointer capture API (stdweb PR)
- [x] DPI handling (the browser handles it)
- [x] Handle on close events (stdweb PR)
- [x] Mouse wheel events
- [ ] Fullscreen (stdweb PR)
- [ ] Monitor APIs
- [ ] Add logging

Testing / docs:
- [ ] Add stdweb to CI


ryanisaacg:
Ok, so I've made a bunch of progress on this (most of it was just churning through various bad designs that didn't work at all) and it's in a state where you can make a window, run an event loop, and receive events! 🎉 

I made a roadmap in the PR text to give an indication of where I'm at on the implantation, and the unresolved questions I have about the design. If any maintainers (or users for that matter) have opinions, please let me know!

solson:
Apologies if this question is naive, but would it make sense for winit to support wasm-bindgen's web-sys directly, rather than stdweb? According to [this thread](https://github.com/rustwasm/team/issues/226#issuecomment-418475778), stdweb plans to become a wrapper around web-sys that's a bit higher level (but also more expensive), while web-sys provides the bare minimum only-pay-for-what-you-use interface for Web APIs.

Either way, it would be fine to support both, or stdweb now and web-sys later, so I wouldn't block the PR over this.

ryanisaacg:
Essentially: stdweb is what I'm more familiar with and what my downstream project uses. I would be open to supporting both (for example the rand crate allows either via feature flags) if someone else contributes the web-sys backend.

Osspial:
General thoughts on the API questions you have:

> Should multiple canvases be supported? I'm leaning no

I'd agree with that for now. We can revisit this once we figure out #696, but for now only having a single canvas is simplest. 

> Should fullscreen be supported? If it should be, should it be the size of the window or some attempt to make the canvas actually fullscreen?

Yes, and I'd say to actually make the canvas fullscreen. This is a feature that's commonly supported across browsers (video fullscreening, for instance), it's useful to have, and it's consistent with the other backends.

> Is the window icon synonymous with the favicon? I'm leaning yes

Eeeeh, I'm less sure about this. If we do end up supporting multiple canvases at some point there's no clean solution for what to do when multiple icons get set on different `Window` objects. This feels like it should be handled by the parent webpage.

> Should file dropping be supported? It's a JS File object which isn't as useful as it is on desktop
Should show / hide be supported?

I'd say no for now. Using the same File API across both desktop platforms and web platforms implies that the data they provide can be used the same way, which doesn't seem to be the case (I'd be surprised if `std::fs::File::open` works on the web, but I may be wrong there). We can come back to this later though, and see about maybe exposing a different API to make clear that it has different semantics.

ryanisaacg:
Just a heads up: my progress on this is going to slow down for a while, as I'll probably not have as much time to work on it for the next few weeks.

blm768:
I've taken a stab at implementing `wasm-bindgen`/`web-sys` support over in [my branch](https://github.com/blm768/winit/tree/web-sys). It's a very rough copy/paste job right now, but it at least compiles. Still needs to be cleaned up and tested.

So far, it looks like the amount of code that needs to be changed between `stdweb` and `wasm-bindgen`/`web-sys` is fairly minimal. It's probably worth considering whether the two should be separate backends, a single backend sprinkled with `#[cfg]` directives, or something in between.

ryanisaacg:
To be clear: this is different than the web-sys effort from #845?

It looks like your code is based on mine (let me know if that's not correct.) If that's the case, once I'm finished I would love to help finish the web-sys version and get them into a state where they're easy to co-maintain. I agree that cfg directives would probably be a good way to do this.

blm768:
> To be clear: this is different than the web-sys effort from #845?

Yep. To be honest, I had no idea that existed until just now. Probably worth trying to merge the two efforts somehow.

> It looks like your code is based on mine (let me know if that's not correct.) 

100% correct. Depending on just how the differences work out, some of the differences might be worth breaking out into their own sub-modules or helper functions, but many of them are probably simpler to just `#[cfg]` out.

blm768:
After a little testing, it's looking like the basic canvas creation and event handling seem to be working (at least with a slightly generous definition of "working") with the `web-sys` version. At this point, my thought is that I should focus on exploring the integration of the three existing efforts so we can avoid having too much duplication of effort. It's probably too early in the design and implementation process to just merge everything into one massive PR, but some early consideration for integration will probably pay off in the long run.

While I'm thinking about it, I've got a couple of thoughts I've been considering:

* I noticed that both this PR and #845 use an exception to bail out of the main event loop. If it's not too late in the design process for `eventloop-2.0`, it seems like providing a "sanctioned" way to bail out of the event loop early would be useful, especially if `winit` ever picks up another callback-driven backend. I'm not familiar enough with `winit` yet to propose any particular design, but it seems like it's worth some consideration.
* When handling the closures in `wasm-bindgen`, it's necessary to keep the `Closure` objects alive on the Rust side at least as long as they might be used on the JavaScript side; after the Rust object is dropped, trying to call the JavaScript closure produces an exception. The `Closure`s can just be leaked, but it seems better to clean them up properly. This requires a [small design change](https://github.com/blm768/winit/commit/4f7e75daa76eafdfc389044522230d97ac4ab659), which may have some bearing on the integration efforts. It might be worth including something similar to this change in the `stdweb` backend since it allows for de-registration of the event handlers when the event loop and/or window go out of scope on the Rust side.

ZeGentzy:
@ryanisaacg Are you intending to maintain the stdweb backend after this PR lands?

ryanisaacg:
@ZeGentzy Yes, but without a guarantee that I'll have a lot of time for it. I'll do my best to keep it updated, though; at the very least I'll be using it downstream.

ZeGentzy:
Okay. Should prolly add a new column and row to the table in https://github.com/rust-windowing/winit/blob/master/CONTRIBUTING.md then.

Also to the tables in https://github.com/rust-windowing/winit/blob/master/FEATURES.md when you are done.

@Osspial can probably add you to rust-windowing

examples/window.rs
```diff
@@ -10,9 +13,10 @@ fn main() {
         .with_title("A fantastic window!")
         .build(&event_loop)
         .unwrap();
+    console!(log, "Built window!");
 
     event_loop.run(|event, _, control_flow| {
-        println!("{:?}", event);
+        console!(log, format!("{:?}", event));
```
felixrabe: You might want to revert the changes in this file.

src/platform_impl/stdweb/event_loop.rs
```diff
@@ -0,0 +1,468 @@
...
+                        },
+                        ControlFlowStatus::Exit => { return; }
+                    }, false)
+                };
+                let mut control = runner.control.to_control_flow();
+                // Handle starting a new batch of events
+                //
+                // The user is informed via Event::NewEvents that there is a batch of events to process
+                // However, there is only one of these per batch of events
+                self.handle_event(Event::NewEvents(start_cause), &mut control);
```
hecrj: Calling this here after borrowing the `runner` causes a panic on my end. I think it's because `handle_event` wants to borrow it again mutably.

examples/window.rs
ryanisaacg: Yup, before merge I'll be doing some cleanup (this is still actively in development and I use this example to test.)

src/platform_impl/stdweb/event_loop.rs
ryanisaacg: Thanks for the heads up, should be fixed now.

The work that's left is logging, adding the onclose event once it's merged into stdweb, adding fullscreen mode and pointer capture support to stdweb, and adding support for those to this backend once they're merged into stdweb. If you want to tackle any of those, feel free!

hecrj:
I am currently working on unifying both the `stdweb` and `web_sys` efforts. I am just restructuring the work by @ryanisaacg and @blm768 while removing duplication. Take a look here: https://github.com/hecrj/winit/tree/web/src/platform_impl/web

In summary, I have created a new `web` platform that contains two backends: `stdweb` and `web_sys`. Most of the code in the `web` platform is backend agnostic. Each backend simply provides `Canvas` and `Timeout` types which is where the actual wiring happens. These types are also useful to avoid leaking closures.

The `web_sys` backend is already usable. I have been able to use it successfully with `wgpu` and WebGL. Here is an example that renders a triangle and outputs the window events on console: http://hello-triangle.surge.sh/

The `stdweb` backend should be quite straightforward to implement. I will work on that later today.

Additionally, I have made some other changes:
  * Added support for the `ReceivedCharacter` event.
  * Made the keyboard and focused events local to the canvas by adding a `tabindex` attribute. Now you need to click on the canvas to interact with it. This should allow us to easily implement support for multiple windows in the same document, and it also removes awkward global listeners.
  * Because of the previous change and given that a document has only one title, `set_title` now only changes the `alt` attribute of the canvas.

Let me know what you think!

blm768:
@hecrj: I haven't looked through your changes thoroughly, but I'm thrilled to learn that it's working (particularly with `wgpu`, which should be really useful in making my own toy project more portable).

Have you happened to look at the changes in #845? If I recall correctly, they had already started on multi-canvas support, and the `WebsysWindowBuilderExt` trait might be a useful starting pattern. For the single-canvas cases, it may also be possible to extend that trait to provide some mechanism to use the document title as the window title, ensure that the canvas has focus so it receives events immediately, etc.

That raises a question about `winit` in general, actually: does it generally report any input events for windows that don't have focus (i.e. "mouse over" events that happen "in passing")? If so, we'll probably want to retain some kind of support for those events.

felixrabe:
(Just noticed that this is a Draft PR that I didn't know about previously. Quick intro with *very* helpful screenshot here: https://github.blog/2019-02-14-introducing-draft-pull-requests/)

hecrj:
@blm768 

> For the single-canvas cases, it may also be possible to extend that trait to provide some mechanism to use the document title as the window title, ensure that the canvas has focus so it receives events immediately, etc.

I see that #845 uses platform-specific builder attributes to take control of an already existing canvas in the document. Maybe we could take this approach and simply assume that the `Window` should take control of the whole document when the attribute is not provided.

> That raises a question about winit in general, actually: does it generally report any input events for windows that don't have focus (i.e. "mouse over" events that happen "in passing")? If so, we'll probably want to retain some kind of support for those events.

Not sure what the desired behavior is in this case, but the current one reports mouse events just fine even when the canvas loses focus. Right now, the only events that are focus-dependent are keyboard events.

ZeGentzy:
@ryanisaacg  As we discussed on Discord, I've made a new web branch with @hecrj's fork. It should be noted that it was four commits behind your branch, so you might want to push those changes to it, if they are still applicable.

Anyways, closing. Thanks!

ZeGentzy:
@ryanisaacg Can you open a tracking issue with the things at the top? Thanks.

ryanisaacg:
@ZeGentzy sure, opened #1072

PR #799, opened 2019-02-22
Remove icon loading feature

### LINKED ISSUE #661: Is it possible to decouple the `Icon` functionality from a specific version of the `image` crate?
As it is now, `WindowBuilder::with_window_icon()` and `winit::Icon` have API's that use various types from the `image` crate: `image::image::ImageError`, `image::image::ImageFormat`, some of the `From` impl's for `Icon`, etc.

These all use the version of `image` that `winit` uses, 0.19, and so if you want to use that functionality you have to use the same version of `image`.  This isn't really a *huge* problem, but is kind of inconvenient.  Perhaps `winit` could at least re-export that version of the `image` crate?  I dunno what the best solution is without just newtype'ing all the types to hide them.

Osspial:

Removes the `icon_loading` feature from the `eventloop-2.0` branch. This does *not* remove window icons as a feature - it just removes the `image` crate integration. This is because we've gotten [so](https://github.com/tomaka/winit/issues/707) [many](https://github.com/tomaka/winit/pull/758) [issues](https://github.com/tomaka/winit/issues/661)/[prs](https://github.com/tomaka/winit/pull/683) about the version of the `image` crate that maintaining our support for it seems to be far more trouble than it's worth. Not only that, but keeping the dependency will just lock us into a never-ending series of breaking changes every time `image` wants to bump its major version number - preventing us from ever stabilizing until `image` reaches `1.0`, and potentially forcing a major bump *after* that if `image` ever decides to make more breaking changes after that.

Yes, this PR forces users to add some boilerplate code, but it [really isn't that much](https://github.com/Osspial/winit/blob/b994918177bc18f48a484e40f04ef190a99419f3/examples/window_icon.rs#L56-L68). 

Discussion is encouraged, of course, but I've got a strong preference towards merging this. Closes #661.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,11 +51,9 @@ install:
 script:
   - cargo build --target $TARGET --verbose
   - cargo build --target $TARGET --features serde --verbose
-  - cargo build --target $TARGET --features icon_loading --verbose
   # Running iOS apps on OSX requires the simulator so we skip that for now
   - if [ "$TARGET" != "x86_64-apple-ios" ]; then cargo test --target $TARGET --verbose; fi
   - if [ "$TARGET" != "x86_64-apple-ios" ]; then cargo test --target $TARGET --features serde --verbose; fi
-  - if [ "$TARGET" != "x86_64-apple-ios" ]; then cargo test --target $TARGET --features icon_loading --verbose; fi
 
 after_success:
   - |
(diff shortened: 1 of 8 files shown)


Osspial:
There seems to be overwhelming support for this so I'll go ahead and merge it.

Issue #795, opened 2019-02-14
Seat identifiers on x11/wayland


vberger:
Wayland is a multi-seat aware protocol, and I believe x11 is too. From this, I think it could be reasonable to add a platform specific extension trait `unix::DeviceIdExt` like so:

```rust
pub trait DeviceIdExt {
    /// Get the name of the seat this device is associated with
    ///
    /// Some systems can be multi-seat, meaning that there may be more than
    /// one keyboard / pointer, and these should be treated as different users interacting
    /// with the app.
    ///
    /// If this kind of multi-user interaction would be meaningful with your app, this method allows
    /// you to retrieve a seat-unique identifier from this `DeviceId`.
    fn seat(&self) -> &str;
}
```

While arguably pretty niche, this seat-awareness can be necessary for downstream apps. For example, on wayland, there is one clipboard per seat, and the app should be careful to trigger the copy and paste actions from the appropriate seat.

Thoughts on this design, @francesca64 ?

pickfire:
I believe this is the same for X11 which is usually called multi-pointer https://wiki.archlinux.org/index.php/Multi-pointer_X. Quick scanning at the docs, seems like not possible to identify which keyboard active using `KeyboardEvent` alone.

notgull:
Is there a way to get the name of the seat from multi-pointer mode?  `xtrace` seems to indicate that the seat name isn't sent over the wire (at least by default) and the [Xinput2](https://www.x.org/releases/X11R7.5/doc/inputproto/XI2proto.txt) specification doesn't mention seats at all. It seems that master devices take priority here, but the seat name doesn't match up with the real seat (e.g. `"seat0"`).

kchibisov:
I'm not sure about X11 here. I know that with Wayland you have seat IDs.

PR #804, opened 2019-03-06
Overhaul device events API and add gamepad support on Windows

### LINKED ISSUE #119: Gamepad support
Surprised nobody's tackled this one yet. We already handle keyboard and mouse input, so gamepad input seems to be the logical next step.

Anyone looking for inspiration might want to see [how SDL2 does this](https://hg.libsdl.org/SDL/file/c2b0d3d180f4/src/joystick): basically they maintain a list of controllers and map them as closely to the XInput layout as possible. That doesn't have to be the strategy we use here, and there is some legacy code in there as well (mainly deprecated Windows APIs like DirectInput), but I think it's a reasonable starting point.

### LINKED ISSUE #212: Consider expanding scope of raw input
In light of [wayland's requirement that winit participate in the raw input stack](https://lists.freedesktop.org/archives/wayland-devel/2017-April/033819.html) (see also [the more mature relative pointer motion spec](https://github.com/wayland-project/wayland-protocols/blob/master/unstable/relative-pointer/relative-pointer-unstable-v1.xml)), `DeviceEvent` and its generic axis/button cases are being retained. On non-wayland platforms, however, winit is not required to, and currently does not, participate in the raw input stack for non-mouse/keyboard devices. We could:

- Continue to only involve winit in input handling on each platform only as much as that platform requires.
This keeps winit smaller, and forces portable downstream code that wants non-mouse/keyboard input to involve an external crate for unhandled inputs and process events from both sources.

- Involve winit in input handling on all platforms for every device class for which winit is required to be involved on any platform
This would mean communicating with evdev (for actual input) and udev (for hotplug support) on Linux, for example.

I think the strongest arguments for such an expansion in scope are consistency and ease of use. If a developer on Wayland someday writes a gamepad-using application on top of winit, they will be frustrated to find that it doesn't work on other platforms. On the other hand, if/when good dedicated crates exist they may not be difficult to drop in, and can presumably no-op on Wayland, removing the need for an application developer to write platform-specific code. `DeviceInput` documentation could recommend this pattern, hopefully reducing potential for surprise.

Because Wayland is (afaik) the only platform where winit *must* be involved, and hasn't yet even adopted the direct input extension, the potential for surprise will remain very low for the near future: everyone already needs external code to handle non-mouse/keyboard input. As such, this is a low priority issue.

### LINKED ISSUE #624: Joystick events

### LINKED ISSUE #750: Device Events API Discussion
The Device Events API currently exists, but there seems to be informal consensus between the maintainers that it needs to be reworked somewhat before we can declare it stable. This issue is meant to house discussion on what changes should be made.

Osspial:

This started out as an attempt to merge @francesca64's gamepad integration (❤️) from #624 onto the EventLoop 2.0 branch, but ballooned to a complete overhaul of the device event API since it was the easiest way to expose a clean rumble API and [we agreed that it](https://github.com/tomaka/winit/pull/695) [needed to be done anyway](https://github.com/tomaka/winit/issues/750).

This is a draft PR because it makes absolutely no attempt to to implement the device event changes on non-Windows platforms, which needs to be done before this can be merged. This also means that it's waiting on X11 and MacOS EL2.0 implementations, since meaningful work to get this working on those backends can't start until those backends work.

Feedback on the API is encouraged, and the [API docs are available here](https://osspial.github.io/winit/el2-win-joy/winit/). I expect the non-breaking changes to be fairly uncontroversial, and these are my reasons for the breaking changes:
- Splitting up `DeviceEvent` into `MouseEvent`, `KeyboardEvent`, and `GamepadEvent`:
  - Previously, the `DeviceEvent` enum was a bit of a mess; there were several different event sources aggregated under one type, most of which could only be emitted from one specific kind of device. Splitting the API up clarifies which events get emitted by which types of devices and makes it more easily learnable.
- Removing `DeviceEvent::Text` variant:
  - No backend implemented it, and its utility compared to `WindowEvent::ReceivedCharacter` was questionable.
- Splitting up `DeviceId`:
  - See reasons for splitting `DeviceEvent`. Also, I had to split off `GamepadHandle` so I could add a `rumble` function, and splitting up `MouseId` and `KeyboardId` makes the API more consistent.
- Removing device IDs from `WindowEvent` variants:
  - This was only implemented on X11, and all the other platforms just returned stub IDs. This was misleading at best, and I couldn't see a compelling reason to keep it in the face of that.

Closes #750, closes #624, closes #119, closes #212 (?).


src/event/device.rs
```diff
@@ -0,0 +1,303 @@
...
+    // @francesca64 review: why were these variants here? I don't see how it makes sense for the dpad or hat switch
+    // to have axes, since they're both four separate buttons.
+
+    // /// This is supposed to have a specialized meaning, referring to a point-of-view switch present on joysticks used
+    // /// for flight simulation. However, Xbox 360 controllers (and their derivatives) use a hat switch for the D-pad.
+    // HatSwitch,
+    // DPadUp,
+    // DPadDown,
+    // DPadLeft,
+    // DPadRight,
```
Osspial: Copied from comment:
@francesca64 review: why were these variants here? I don't see how it makes sense for the dpad or hat switch to have axes, since they're both four separate buttons.

src/platform_impl/windows/raw_input.rs
```diff
@@ -231,5 +423,265 @@ pub fn get_raw_mouse_button_state(button_flags: USHORT) -> [Option<ElementState>
...
+                //     0x33 => Some(GamepadAxis::LeftTrigger),
+                //     0x34 => Some(GamepadAxis::RightTrigger),
+                //     0x35 => Some(GamepadAxis::RightStickY),
+                //     0x39 => Some(GamepadAxis::HatSwitch),
+                //     0x90 => Some(GamepadAxis::DPadUp),
+                //     0x91 => Some(GamepadAxis::DPadDown),
+                //     0x92 => Some(GamepadAxis::DPadRight),
+                //     0x93 => Some(GamepadAxis::DPadLeft),
+                //     _ => None,
+                // },
```
Osspial: Copied from comment:
@francesca64 when reviewing - where did you get these values? I've commented them out because the HOTAS controller I've been using to test the raw input backend has been getting axis hints when it shouldn't, and this seems to be the culprit.

vberger:
I can't review this PR in detail right now, I do absolutely not have the time to do so, but your initial message raised a few questions & remarks for me:

> - Splitting up `DeviceEvent` into `MouseEvent`, `KeyboardEvent`, and `GamepadEvent`: 

Is there a plan for `OtherEvent` or something like that, to expose devices that generate events but are not mouses, keyboards or gamepads? Or would these just be unsupported?

> - Removing device IDs from WindowEvent variants:
>   - This was only implemented on X11, and all the other platforms just returned stub IDs.

Wayland could have supported them, and they actually be necessary to support multi-seat systems if it is ever planned, see #795 . Noting that not supporting them properly would mean the clipboard story on wayland would remain hackish forever.

Osspial:
> Is there a plan for `OtherEvent` or something like that, to expose devices that generate events but are not mouses, keyboards or gamepads? Or would these just be unsupported?

I'd be down for having that, but it wasn't something I was originally considering when writing this PR. Incidentally, most of the legwork is already done for it on Windows, so I could conceivably add basic support onto this. If we did that, we'd need to consider at some point whether or not we expose a cross-platform API for accessing and modifying those devices beyond platform-specific traits, and what said API would look like.

> Wayland could have supported them, and they actually be necessary to support multi-seat systems if it is ever planned, see #795 . Noting that not supporting them properly would mean the clipboard story on wayland would remain hackish forever.

Oh, I didn't realize that! It seems important to have them, then. I'd rather expose them as an `Option<{Device}>` than the previous form, though, as it isn't possible on all platforms.

Osspial:
Wait, couldn't the active seat identifier be exposed through `WindowExtUnix` or `EventLoopExtUnix`? My (admittedly casual) understanding of multi-seat is that each program runs within the context of a single seat, so there shouldn't be a situation in which multiple seats are trying to interact with a single application at once.

vberger:
> My (admittedly casual) understanding of multi-seat is that each program runs within the context of a single seat

No, that'd be a multi-session setup. Here, several seat can be connected to the same session, interacting with the same windows. See https://www.youtube.com/watch?v=WO2L_ihO_rI for example

Osspial:
Ah, OK. That's a bit of confusing terminology then, as [X11 multi-seat is what Wayland calls multi-session](https://wiki.archlinux.org/index.php/xorg_multiseat), and [Wayland multi-seat is what X11 calls multi-pointer](https://wiki.archlinux.org/index.php/Multi-pointer_X).

Should we expose the Seat ID directly through `WindowEvent`'s variants, rather than through the Device ID? That makes it more clear that multi-seat configurations are possible, instead of relegating it to a platform extension that most people won't check to see exists. That also allows us to provide a sensible non-`None` default on non-linux platforms, since on those there's only one seat active.

Osspial:
Regarding the `OtherEvent` point you made - I've added `HidId`/`HidEvent` to the public API. Unfortunately, I don't have any devices that declare themselves as HID devices, so I've been unable to test that it works.

src/platform_impl/windows/raw_input.rs
francesca64: I got them from here: http://www.freebsddiary.org/APC/usb_hid_usages.php

I'm confident those values aren't reliable, of course. Being compatible with SDL mappings is basically the only way to provide layout info for RAWINPUT/etc.

src/event/device.rs
francesca64: I don't remember, but if there's any validity to them being here, it probably has to do with weird differences in how Xbox controllers work under RAWINPUT or something. There are lots of articles that complain about quirks like that, so if nothing comes up from a cursory search, then you can just assume this is some cruft I left in.

aloucks:
WIll it be more confusing to have Device events broken out individually while WindowEvents remain bundled under a common enum? This might make it more difficult to understand the subtitles between similar events (e.g. MouseMotion vs CursorMoved).

Osspial:
@aloucks That's a good point! It's probably worth taking a closer look at `WindowEvent` and examining it in the broader context of the API, but it wasn't something I considered when working on this PR.

aleksijuvani:
I finished reading through #695, but I'm still curious as to why it was decided to support gamepads in winit. Regardless of whether people expect winit to support it or not, gilrs already exists for talking to gamepads, and it looks to me like it does this pretty well. It seems like a waste to duplicate this effort. I thought this was previously declared as "out of scope" for winit?

Osspial:
@aleksijuvani It was indeed declared out of scope previously, but that was by a different person that's no longer actively involved with the project and I personally disagree with it. As for why we'd implement this ourselves instead of just directing people to `gilrs`: `gilrs` only supports a polling-based API, which is incompatible with non-game applications. Media center applications, for example, may still want to have responsive controller support while not constantly running a loop at 60 hertz. Besides, I can see no reason ideologically for gamepad input to be out of scope - Winit is an "window creation and input handling" library, and gamepad input is certainly used as an input method for a non-trivial number of applications.

Osspial:
Once I've rebased this, I'm going to merge this onto the `gamepad-device-events` branch so new contributors can easily make PRs against it.

ZeGentzy:
Can we merge this instead against https://github.com/rust-windowing/winit/tree/dpi-overhaul?  Makes my life easier, as I don't have to maintain two copies of glutin.

Osspial:
@ZeGentzy I'd rather not, since that makes it impossible to implement just the DPI changes or gamepad support in a single PR - you'd have to deal with the error messages from both at one time, which sounds like a bad experience for new contributors.

Why would you have to maintain additional copies of glutin, anyway? The intent isn't for people to actively use those branches in their applications - it's so we have a centralized place to develop these features, and then merge them onto master. Glutin isn't necessary for either of those processes.

(also I'd be surprised if the breaking changes in this patch break things on Glutin's end)

ZeGentzy:
People usually (I think?) only want to implement something if they want to use it in their application. Generally, from my experience at least, you port your application to this shiny new unimplemented api, then fiddle with the dependency till it compiles and works. Like, what better way to test if gampad support works then seeing if it works in your application?

Now, if these applications are dependent on glutin, and this PR doesn't break anything in glutin, thenI guess it's as simple as a `[replace]` in the `Cargo.toml`. However, if this pr does break something, now some poor soul is going to fiddling with glutin and winit.

Besides, what if they want to run their application against both the dpi overhaul and with gamepad support? I see little joy for the poor soul who was to merge the two branches together, as git is a PITA, nor for the poor soul in the end who has to merge the two branches back into master.

And, of course, now we got to backport changes to two branches, just not one, to avoid them getting too out of sync from master. This is like the eventsloopv2-master split, but with twice the pain.

Osspial:
Remember, `dpi-overhaul` and `gamepad-device-events` aren't intended as long-lived branches that diverge significantly from `master`: they're small branches oriented around specific features that exist to make it easy for new contributors to make significant improvements to Winit's API even if they aren't familiar with our codebase. Part of the reason `dpi-overhaul` is a great place for new contributors is that implementing its changes for the most part just involves following several related error messages, and making localized changes around those error messages. Since both this branch and `dpi-overhaul` break compilation on non-Windows platforms, combining them prematurely will make it impossible for new contributors to focus on a single problem at a time.

> Now, if these applications are dependent on glutin, and this PR doesn't break anything in glutin, then I guess it's as simple as a `[replace]` in the `Cargo.toml`. 

I just tested it, and I can confirm that this PR doesn't break anything in Glutin.

> Like, what better way to test if gampad support works then seeing if it works in your application?

I'll agree with that, but see point above.

> This is like the eventsloopv2-master split, but with twice the pain.

The difference between these two branches and the EL2 branch is that the changes in these branches are *much* less fundamental than the changes introduced by the EL2 branch. We aren't in a situation where changes to `master` are inherently incompatible with changes to these branches - the vast majority of changes will be localized enough that merging will either be completely painless or only involve trivial tweaks. Indeed, I just tested merging `dpi-overhaul` and `el2-win-joy`, and git merged them without complaints.

Besides, these branches should get merged much more quickly than than the EL2 branch did, since the changes are much easier to implement and can be implemented without a deep understanding of Winit's structure.

> I see little joy for the poor soul who was to merge the two branches together (and then demerge them to send in PRs separately), as git is a PITA, nor for the poor soul in the end who has to merge the two branches back, one-by-one, into master, hoping for the changes from each don't conflict with each other.

I'd argue that merging two WIP branches, fixing the issues raised by both branches, than de-merging them is a gross misuse of git as a tool.

PR #805, opened 2019-03-08
Add additional numpad key mappings

chrisduerr:
Since some platforms have already used the existing `Add`, `Subtract`
and `Divide` codes to map numpad keys, the X11 and Wayland platform has
been updated to achieve parity between platforms. On macOS only the
`Subtract` numpad key had to be added.

Since the numpad key is different from the normal keys, an alternative
option would be to add new `NumpadAdd`, `NumpadSubtract` and
`NumpadDivide` actions, however I think in this case it should be fine
to map them to the same virtual key code.


Osspial:
Thanks for the PR!

This looks good to me, but I'm going to cc @vberger and @francesca64 so they can take a look, as the maintainers of the changed platforms.

Osspial:
Thanks! Merging

PR #810, opened 2019-03-09
fix command key event left and right reverse on macOS

TakWolf:
https://github.com/tomaka/winit/issues/808

------------------

I only swapped the VirtualKeyCode mapping.


Osspial:
Thanks! Sorry it took so long to get back on this - I kept putting it off and, well, here we are.

All that's missing is a changelog entry. Add that in, and I'll go ahead and merge this.

TakWolf:
I just made changes based on the results. I don't have a clear documentation reference but https://github.com/phracker/MacOSX-SDKs/blob/master/MacOSX10.6.sdk/System/Library/Frameworks/Carbon.framework/Versions/A/Frameworks/HIToolbox.framework/Versions/A/Headers/Events.h#L273

I'm not sure it is ok on all keyboard style.It is best to have another ont to test this.

Osspial:
Fair enough. @mtak- are you able to test this?

chrisduerr:
@Osspial Just to potentially speed things up a bit, I've tested this on my macbook at work and I was able to confirm both that the current behavior is wrong (LWin and RWin swapped) and that this PR fixes that.

Osspial:
@chrisduerr Thanks for the test! I'll merge this now.

PR #815, opened 2019-03-17
Fix build on FreeBSD

t6:
```
error[E0432]: unresolved import `libc::__errno_location`
  --> src/platform/linux/x11/mod.rs:22:85
   |
22 | use libc::{select, fd_set, FD_SET, FD_ZERO, FD_ISSET, EINTR, EINVAL, ENOMEM, EBADF, __errno_location};
   |                                                                                     ^^^^^^^^^^^^^^^^ no `__errno_location` in the root
```
__errno_location is called __error on FreeBSD and __errno on Open- and NetBSD.

Note that I only tested on FreeBSD.  The OpenBSD and NetBSD bits are a pure guess on my part based on libc documentation and their errno.h [2,3,4,5].

[1] FreeBSD: https://rust-lang.github.io/libc/x86_64-unknown-freebsd/libc/fn.__error.html
[2] NetBSD: https://rust-lang.github.io/libc/x86_64-unknown-netbsd/libc/fn.__errno.html
[3] NetBSD: https://github.com/NetBSD/src/blob/trunk/include/errno.h
[4] OpenBSD: https://rust-lang.github.io/libc/x86_64-unknown-openbsd/libc/fn.__errno.html
[5] OpenBSD: https://github.com/OpenBSD/src/blob/master/include/errno.h

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 - On Windows, fix `CursorMoved(0, 0)` getting dispatched on window focus.
+- On FreeBSD, NetBSD, and OpenBSD, fix build of X11 backend.
 
 # Version 0.19.0 (2019-03-06)
 
--- a/src/platform/linux/x11/mod.rs
+++ b/src/platform/linux/x11/mod.rs
@@ -23,9 +23,9 @@ use libc::{select, fd_set, FD_SET, FD_ZERO, FD_ISSET, EINTR, EINVAL, ENOMEM, EBA
 #[cfg(target_os = "linux")]
 use libc::__errno_location;
 #[cfg(target_os = "freebsd")]
-use libc::__error;
+use libc::__error as __errno_location;
 #[cfg(any(target_os = "netbsd", target_os = "openbsd"))]
-use libc::__errno;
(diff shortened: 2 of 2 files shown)


t6:
An alternative solution might be to use the errno crate to access `errno` in a cross-platform way.

https://crates.io/crates/errno

vberger:
Hi, thanks for the PR!

Just as a stylistic nitpick, maybe it'd be better to just rename the imports, to avoid littering `#[cfg()]` markers all over the source, given this is apparently just a name change.

I'm thinking something like:

```rust
#[cfg(target_os = "linux")]
use libc::__errno_location;
#[cfg(target_os = "freebsd")]
use libc::__error as __errno_location;
#[cfg(any(target_os = "netbsd", target_os = "openbsd"))]
use libc::__errno as __errno_location;
```

Do we have any strong stylistic constraints wrt to this ? @francesca64 @Osspial ?

t6:
I've changed it as you've suggested.  Anything else I can to do to get this merged?

Osspial:
The only thing I can see blocking this is a lack of changelog entry. We should be able to merge this once that's added in.

t6:
Added one.

Issue #817, opened 2019-03-18
Focus event order on Windows


chrisduerr:
While trying to fix https://github.com/jwilm/alacritty/issues/2170, I've just noticed that the event order of the `Focused` event is a bit different on Windows compared to the other platforms.

Even though window focus on Windows can only be changed when clicking on a window, the `Focused` events comes *before* the `MouseInput` event. They aren't coming in the same bulk either so the order isn't only messed up but they're coming in completely disconnected from each other in short succession.

The issue with this is that it makes it impossible to determine if a `MouseInput` event was on a focused or unfocused window. So clicks on unfocused windows can't easily be ignored.

I'm not sure if this is an issue with winit or a constraint due to the Windows API, but this is probably the best place to ask.

Osspial:
I'm not aware of any clean way to deliver `MouseInput` before `Focused` - it would require some pretty hacky stuff on our end to get working, and I'm not comfortable with introducing that to the codebase. *However*, it should still be possible to fix the downstream behavior easily!

When I was looking into this, I found that someone had put in code that resulted in a [`CursorMoved(0, 0)` event getting dispatched whenever the window gained focus](https://github.com/tomaka/winit/blob/master/src/platform/windows/events_loop.rs#L1061-L1069), which is pretty clearly causing the selection bug in Alacritty. No idea why that was there - the [relevant documentation](https://docs.microsoft.com/en-us/windows/desktop/inputdev/wm-setfocus) explicitly says that the variable they're trying to read as mouse position is unused, but removing that event dispatch is easy. I'll get a PR going to fix that ASAP.

chrisduerr:
As far as I'm concerned, fixing the issue is good enough at this point.

However by default on Windows the first click on a focused window is usually ignored, so selections should not be cancelled. At least cmd.exe and powershell behave this way.

For Alacritty that might not be a huge deal, but it might still be something that winit should support.

Always ignoring the first click after the focus has changed from unfocused to focused doesn't seem like an option either, since it's possible to focus a window without using the mouse.

Are you aware of any option a user of winit might have to resolve this? It seems impossible to determine if a click was what caused the focus or not.

Osspial:
[...]

PR #819, opened 2019-03-19
On Windows, fix CursorMoved(0, 0) getting sent on focus

Osspial:
The code that was sending it shouldn't even have been there to begin with. This PR removes said code.


Osspial:
Fixes the downstream bug in #817 (https://github.com/jwilm/alacritty/issues/2170).

PR #838, opened 2019-04-12
Feat/fullscreen getters

### LINKED ISSUE #579: Add Window::get_fullscreen() -> bool

acheronfail:
This change adds the `window.get_fullscreen()` method - this is useful for many reasons and makes using this library much easier.

It also adds a `get_simple_fullscreen()` method to the macos `WindowExt` as well, so macos users can distinguish which type of fullscreen mode is currently active.

I've tested this on the following platforms:

- [x] macOS
- [x] Linux (X11)
- [x] Windows 10

Closes #579

### DIFF
--- a/src/platform/linux/wayland/window.rs
+++ b/src/platform/linux/wayland/window.rs
@@ -7,7 +7,7 @@ use platform::{MonitorId as PlatformMonitorId, PlatformSpecificWindowBuilderAttr
 use window::MonitorId as RootMonitorId;
 
 use sctk::surface::{get_dpi_factor, get_outputs};
-use sctk::window::{ConceptFrame, Event as WEvent, Window as SWindow, Theme};
+use sctk::window::{ConceptFrame, Event as WEvent, State as WState, Window as SWindow, Theme};
 use sctk::reexports::client::{Display, Proxy};
 use sctk::reexports::client::protocol::{wl_seat, wl_surface};
 use sctk::reexports::client::protocol::wl_surface::RequestsTrait as SurfaceRequests;
@@ -24,7 +24,7 @@ pub struct Window {
     kill_switch: (Arc<Mutex<bool>>, Arc<Mutex<bool>>),
     display: Arc<Display>,
     need_frame_refresh: Arc<Mutex<bool>>,
-    fullscreen: Arc<Mutex<Option<RootMonitorId>>>,
+    fullscreen: Arc<Mutex<bool>>,
 }
 
 impl Window {
@@ -32,6 +32,7 @@ impl Window {
         let (width, height) = attributes.dimensions.map(Into::into).unwrap_or((800, 600));
         // Create the window
         let size = Arc::new(Mutex::new((width, height)));
+        let fullscreen = Arc::new(Mutex::new(false));
 
         let window_store = evlp.store.clone();
         let surface = evlp.env.create_surface(move |dpi, surface| {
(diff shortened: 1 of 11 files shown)


rukai:
Related issue: https://github.com/tomaka/winit/issues/579

Thanks for doing this.

src/platform/linux/x11/window.rs
```diff
@@ -38,6 +38,7 @@ pub struct SharedState {
     pub last_monitor: Option<X11MonitorId>,
     pub dpi_adjusted: Option<(f64, f64)>,
     // Used to restore position after exiting fullscreen.
```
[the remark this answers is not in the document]
acheronfail: Good spot - will fix.

acheronfail:
> One design question - if we're going to be exposing the API, is there any technical reason we can't expose it as get_fullscreen() -> Option<MonitorId>?

Good idea, originally I made it a `bool`  since macOS has two fullscreen modes (one of them doesn't use a `MonitorId`). But, that doesn't apply since macOs users can just use the `WindowExt` in order to test for the other fullscreen mode anyway.

I've updated `get_fullscreen` to return an `Option<MonitorId>` now. 😄

src/platform/linux/wayland/window.rs
```diff
@@ -235,6 +242,8 @@ impl Window {
         } else {
             self.frame.lock().unwrap().unset_fullscreen();
         }
+
+        *(self.fullscreen.lock().unwrap()) = monitor.clone();
```
vberger: Actually, this would not be the proper place to fill this value. Wayland is all about requesting permission, and we are not sure that we will be set fullscreen on the monitor we requested (or set fullscreen at all actually). We need to track it from the `Configure` events we get.

At this line:

https://github.com/rust-windowing/winit/blob/b94572621a4da21af909950d8a00211b053c4d5f/src/platform/linux/wayland/window.rs#L47-L49

The [`Configure` variant](https://docs.rs/smithay-client-toolkit/0.6.1/smithay_client_toolkit/window/enum.Event.html#variant.Configure) has an other field, `states`. Which is a `Vec`. If we are set fullscreen, it'll contain [`State::Fullscreen`](https://docs.rs/smithay-client-toolkit/0.6.1/smithay_client_toolkit/window/enum.State.html#variant.Fullscreen) as one of its values.

You can then obtain the monitor we are fullscren on using `get_outputs(&my_surface)` in a way similar to the `get_current_monitor` method:

https://github.com/rust-windowing/winit/blob/b94572621a4da21af909950d8a00211b053c4d5f/src/platform/linux/wayland/window.rs#L273-L279

acheronfail: After trying to implement this myself, I'm not sure that it's possible with the current architecture.

My original plan was to check for `State::Fullscreen`, toggle a `fullscreen` `bool` and then make `get_fullscreen()` something like:

```rs
pub fn get_fullscreen(&self) -> Option<MonitorId> {
  if self.fullscreen {
    Some(self.get_current_monitor())
  else {
    None
  }
}

However, the closure that's checking the for the `event::Configure` variant only has access to the `InternalWindow` structs. However, the `get_fullscreen()` method only has access to the `Window` struct.

I don't know how to share any information between these two scopes with the current implementation, so I'm not sure where to go from here.

@vberger any chance you could shed some light on a potential solution?

Comment (vberger) [reply src/platform/linux/wayland/window.rs]: You can share the `fullscreen` boolean between the `Window` and `InternalWindow`, in the same way their `size` field is shared.

Comment (acheronfail) [reply src/platform/linux/wayland/window.rs]: Ah! Of course, I had missed that. I'll do that now. :)

File: src/platform/linux/wayland/window.rs [hunk 3]
```diff
@@ -45,12 +47,22 @@ impl Window {
...
+                WEvent::Configure { new_size, states } => {
                     let mut store = window_store.lock().unwrap();
+
+                    let mut is_fullscreen = false;
+                    for state in states {
+                        if let WState::Fullscreen = state {
+                            is_fullscreen = true;
+                            break;
+                        }
+                    }
```
vberger: `WState` implements `PartialEq`, so you should be able to contract this to:

```rust
let is_fullscreen = states.contains(&WState::Fullscreen);
```

acheronfail: Good point! That's much nicer. I've updated the commit with that change. :)

PR #839, opened 2019-04-13
Prevent the event loop from pausing when entering modal loop (eventloop-2.0)

aloucks:
After clicking the window title bar or border (for a drag or resize), the event loop pauses until the mouse is moved. This change relays the `WM_NCLBUTTONDOWN` message to the dummy window where it queues a redraw and consumes the message. This effectively jumpstarts the modal loop and it continues to fire draw requests.

### DIFF
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -799,6 +799,19 @@ unsafe extern "system" fn public_window_callback<T>(
             commctrl::DefSubclassProc(window, msg, wparam, lparam)
         },
 
+        winuser::WM_NCLBUTTONDOWN => {
+            let mut runner = subclass_input.event_loop_runner.runner.borrow_mut();
+            if let Some(ref mut runner) = *runner {
+                let hwnd = runner.modal_redraw_window;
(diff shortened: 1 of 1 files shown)


src/platform_impl/windows/event_loop.rs
```diff
@@ -799,6 +799,19 @@ unsafe extern "system" fn public_window_callback<T>(
...
+            if let Some(ref mut runner) = *runner {
+                let hwnd = runner.modal_redraw_window;
+                winuser::PostMessageW(hwnd, winuser::WM_NCLBUTTONDOWN, 0, 0);
+                if wparam == winuser::HTCAPTION as _ {
+                    winuser::PostMessageW(hwnd, winuser::WM_MOUSEMOVE, 0, 0);
+                }
+            }
+            drop(runner);
+            commctrl::DefSubclassProc(window, msg, wparam, lparam)
+        },
```
Osspial: Why do you forward the `WM_NCLBUTTONDOWN` message to the dummy window and trigger a redraw in there, instead of asking it to redraw directly from here?

aloucks: I checked again and also came to the conclusion that it's not necessary. I've pushed a new commit to remove the forwarding. I think it was leftover from sorting out the relationship with the dummy modal window.

Issue #728, opened 2018-11-26
Function naming inconsistency


Osspial:
Somewhat nitpicky, but there's an amount of inconsistency in our function naming schemes:

* `WindowBuilder::with_dimensions` vs `Window::set_inner_size`
* `WindowBuilder::with_visibility` vs `Window::show`/`Window::hide`
* `Window::get_position` and `Window::get_inner_position` vs `Window::get_outer_size` and `Window::get_inner_size` (`get_position` should be `get_outer_position`)
* `Window::set_position` (should be `set_outer_position`)
* `grab_cursor` and `hide_cursor` vs all other state-setting functions (should be `set_cursor_grab` and `set_cursor_visibility`)
* `set_{min/max}_dimensions` vs `get_{whatever}_size` (`set_dimensions` should be `set_size`)

Additionally, there's inconsistency in the word forms of the functions:

* `with_visibility` vs `with_maximized` (`visibility` is a noun, while `maximized` is a past-tense verb. Could be `maximization`)
* `with_visibility` vs `with_resizable` (`resizable` is an adjective. Could be `resizability`)

I don't feel too strongly about the word-form inconsistency since that's hard to get right and being consistent may force us to use uncommon forms of the words, but the other inconsistencies should be addressed. Being consistent will reduce the mental overhead of using the API and should reduce the amount of time users spend referring to the API docs to find which naming variation is used for each specific function.

francesca64:
This all sounds good to me. I also agree that word-form consistency likely wouldn't be practical to enforce, but I'm in favor of having the other changes be part of the `EventLoop` 2.0 release.

aloucks:
The inner/outer size and position inconsistency is particularly confusing.

PR #857, opened 2019-04-28
Add window centering.

mralve:
Added window centering on windows and linux (x11) at least.

Added convenient functions to be able to center the window quickly, also made it default on windows, but this is togglable.

- [ - ] Created an example program if it would help users understand this functionality
- [x] Updated [feature matrix](https://github.com/tomaka/winit/blob/master/FEATURES.md), if new features were added or implemented

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 - Added to `WindowAttributes` the attributes `center_window: bool` and `start_monitor: i16`. Window centering is by default enabled.
-- Added the following convenience functions for managing window centering: center(), and set_centered(). Refer to the documentation for how to use them.
+- Added the following convenience functions for managing window centering: `center()`, and `set_centered()`. Refer to the documentation for how to use them.
 # Version 0.19.1 (2019-04-08)
 
 - On Wayland, added a `get_wayland_display` function to `EventsLoopExt`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,5 @@
 # Unreleased
 
-- Added in WindowAttributes `center_window: bool` `start_monitor: i16`, Window centering is by default enabled on windows.
-- Added convenience functions to manage window centering, `center()` ,`set_center_monitor()`, `set_centered()` see more in window struct.
-
 # Version 0.19.1 (2019-04-08)
 
 - On Wayland, added a `get_wayland_display` function to `EventsLoopExt`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+- Added in WindowAttributes `center_window: bool` `start_monitor: i16`, Window centering is by default enabled on windows.
+- Added convenience functions to manage window centering, `center()` ,`set_center_monitor()`, `set_centered()` see more in window struct.
+
 # Version 0.19.1 (2019-04-08)
 
 - On Wayland, added a `get_wayland_display` function to `EventsLoopExt`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
-
+- Added in WindowAttributes `center_window: bool` `start_monitor: i16`, Window centering is by default enabled on windows.
+- Added convenience functions to manage window centering, `center()` ,`set_center_monitor()`, `set_centered()` see more in window struct.
 # Version 0.19.1 (2019-04-08)
 
 - On Wayland, added a `get_wayland_display` function to `EventsLoopExt`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Unreleased
-- Added to `WindowAttributes` the attributes `center_window: bool` and `start_monitor: i16`. Window centering is by default enabled.
-- Added the following convenience functions for managing window centering: `center()`, and `set_centered()`. Refer to the documentation for how to use them.
+
+- Added to `WindowAttributes` the `center_window: Option<MonitorId>`.
+- Added the following function for managing window centering `set_centered()`.
+
 # Version 0.19.1 (2019-04-08)
 
 - On Wayland, added a `get_wayland_display` function to `EventsLoopExt`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
-- Added to `WindowAttributes` the attributes `center_window: bool` and `monitor_index: i16`. Window centering is by default enabled.
-- Added the following convenience functions for managing window centering: center(), set_target_monitor(), and set_centered(). Refer to the documentation for how to use them.
+- Added to `WindowAttributes` the attributes `center_window: bool` and `start_monitor: i16`. Window centering is by default enabled.
+- Added the following convenience functions for managing window centering: center(), and set_centered(). Refer to the documentation for how to use them.
 # Version 0.19.1 (2019-04-08)
 
 - On Wayland, added a `get_wayland_display` function to `EventsLoopExt`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
-- Added in WindowAttributes `center_window: bool` `start_monitor: i16`, Window centering is by default enabled on windows.
-- Added convenience functions to manage window centering, `center()` ,`set_center_monitor()`, `set_centered()` see more in window struct.
+- Added to `WindowAttributes` the attributes `center_window: bool` and `monitor_index: i16`. Window centering is by default enabled.
+- Added the following convenience functions for managing window centering: center(), set_target_monitor(), and set_centered(). Refer to the documentation for how to use them.
 # Version 0.19.1 (2019-04-08)
 
 - On Wayland, added a `get_wayland_display` function to `EventsLoopExt`.
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -154,7 +154,7 @@ Legend:
 |Window transparency              |✔️     |✔️     |✔️         |✔️             |**N/A**|**N/A**|**N/A**   |
 |Window maximization              |✔️     |✔️     |✔️         |✔️             |**N/A**|**N/A**|**N/A**   |
 |Window maximization toggle       |✔️     |✔️     |✔️         |✔️             |**N/A**|**N/A**|**N/A**   |
-|Window centering                 |✔️     |❌     |✔️         |❌             |**N/A**|**N/A**|**N/A**   |
+|Window centering                 |✔️     |❌     |✔️         |**N/A**         |**N/A**|**N/A**|**N/A**   |
 |Fullscreen                       |✔️     |✔️     |✔️         |✔️             |**N/A**|**N/A**|❌        |
 |Fullscreen toggle                |✔️     |✔️     |✔️         |✔️             |**N/A**|**N/A**|❌        |
 |HiDPI support                    |✔️     |✔️     |✔️         |✔️             |▢[#721]|✔️    |✔️         |
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -88,7 +88,7 @@ If your PR makes notable changes to Winit's features, please update this section
 - **Popup / modal windows**: Windows can be created relative to the client area of other windows, and parent
   windows can be disabled in favor of popup windows. This feature also guarantees that popup windows
   get drawn above their owner.
-
+- **Window centering**: Sets the window position to be in the center of the given monitor.
 
 ### System Information
 - **Monitor list**: Retrieve the list of monitors and their metadata, including which one is primary.
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -81,6 +81,7 @@ If your PR makes notable changes to Winit's features, please update this section
 - **Window maximization**: The windows created by winit can be maximized upon creation.
 - **Window maximization toggle**: The windows created by winit can be maximized and unmaximized after
   creation.
+- **Window centering**: Sets the window position to be in the center of the given monitor.
 - **Fullscreen**: The windows created by winit can be put into fullscreen mode.
 - **Fullscreen toggle**: The windows created by winit can be switched to and from fullscreen after
   creation.
@@ -88,7 +89,6 @@ If your PR makes notable changes to Winit's features, please update this section
 - **Popup / modal windows**: Windows can be created relative to the client area of other windows, and parent
   windows can be disabled in favor of popup windows. This feature also guarantees that popup windows
   get drawn above their owner.
-- **Window centering**: Sets the window position to be in the center of the given monitor.
 
 ### System Information
 - **Monitor list**: Retrieve the list of monitors and their metadata, including which one is primary.
@@ -154,6 +154,7 @@ Legend:
 |Window transparency              |✔️     |✔️     |✔️         |✔️             |**N/A**|**N/A**|**N/A**   |
 |Window maximization              |✔️     |✔️     |✔️         |✔️             |**N/A**|**N/A**|**N/A**   |
 |Window maximization toggle       |✔️     |✔️     |✔️         |✔️             |**N/A**|**N/A**|**N/A**   |
+|Window centering                 |✔️     |❌     |✔️         |❌             |**N/A**|**N/A**|**N/A**   |
 |Fullscreen                       |✔️     |✔️     |✔️         |✔️             |**N/A**|**N/A**|❌        |
 |Fullscreen toggle                |✔️     |✔️     |✔️         |✔️             |**N/A**|**N/A**|❌        |
 |HiDPI support                    |✔️     |✔️     |✔️         |✔️             |▢[#721]|✔️    |✔️         |
--- a/examples/window.rs
+++ b/examples/window.rs
@@ -8,7 +8,6 @@ fn main() {
         .build(&events_loop)
         .unwrap();
 
-    _window.center();
 
     events_loop.run_forever(|event| {
         println!("{:?}", event);
--- a/examples/window.rs
+++ b/examples/window.rs
@@ -8,6 +8,8 @@ fn main() {
         .build(&events_loop)
         .unwrap();
 
+    _window.center();
+
     events_loop.run_forever(|event| {
         println!("{:?}", event);
 
--- a/examples/window_icon.rs
+++ b/examples/window_icon.rs
@@ -27,11 +27,9 @@ fn main() {
         // At present, this only does anything on Windows and X11, so if you want to save load
         // time, you can put icon loading behind a function that returns `None` on other platforms.
         .with_window_icon(Some(icon))
-        .with_centering(true)
         .build(&events_loop)
         .unwrap();
     
-    //window.center();
 
     events_loop.run_forever(|event| {
         if let winit::Event::WindowEvent { event, .. } = event {
--- a/examples/window_icon.rs
+++ b/examples/window_icon.rs
@@ -27,8 +27,11 @@ fn main() {
         // At present, this only does anything on Windows and X11, so if you want to save load
         // time, you can put icon loading behind a function that returns `None` on other platforms.
         .with_window_icon(Some(icon))
+        .with_centering(true)
         .build(&events_loop)
         .unwrap();
+    
+    //window.center();
 
     events_loop.run_forever(|event| {
         if let winit::Event::WindowEvent { event, .. } = event {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -97,9 +97,9 @@ extern crate image;
 extern crate serde;
 
 #[cfg(target_os = "windows")]
-extern crate backtrace;
-#[cfg(target_os = "windows")]
 extern crate winapi;
+#[cfg(target_os = "windows")]
+extern crate backtrace;
 #[macro_use]
 #[cfg(target_os = "windows")]
 extern crate bitflags;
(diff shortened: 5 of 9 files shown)


mralve:
Is currently testing/working on Linux

vberger:
It looks like your PR changes a lot of unrelated code, I suspect you've run rustfmt on the files?

Could you avoid doing so? It makes the pull request pretty difficult to review.

mralve:
@vberger yeah exactly sorry, it's my vscode having auto fmt on, I will see if I can correct it in a new commit.

mralve:
@vberger Now this should be ready for review, although I don't possess the tools for mac so it has to be done by someone else. ^_^

examples/window.rs
```diff
@@ -8,6 +8,8 @@ fn main() {
         .build(&events_loop)
         .unwrap();
 
+    _window.center();
```
ZeGentzy: Nit: remove the underscore.

examples/window_icon.rs
```diff
@@ -27,8 +27,11 @@ fn main() {
         // At present, this only does anything on Windows and X11, so if you want to save load
         // time, you can put icon loading behind a function that returns `None` on other platforms.
         .with_window_icon(Some(icon))
+        .with_centering(true)
         .build(&events_loop)
         .unwrap();
+    
+    //window.center();
```
ZeGentzy: Nit: either uncomment, or remove.

src/lib.rs
```diff
@@ -510,6 +510,16 @@ pub struct WindowAttributes {
     /// The default is `false`.
     pub always_on_top: bool,
 
+    /// Whether the window will be centered on screen.
+    ///
+    /// The default is true.
```
ZeGentzy: true -> \`true\`

src/lib.rs
```diff
@@ -510,6 +510,16 @@ pub struct WindowAttributes {
...
+    pub center_window: bool,
+
+    /// Decides what monitor the window will start on. Users primary monitor is -1.
```
ZeGentzy: what -> which
Users -> The user's
-1 -> \`-1`

src/lib.rs
```diff
@@ -510,6 +510,16 @@ pub struct WindowAttributes {
...
+    ///
+    /// The default is -1,
```
ZeGentzy: -1 -> \`-1`

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
-
+- Added in WindowAttributes `center_window: bool` `start_monitor: i16`, Window centering is by default enabled on windows.
```
ZeGentzy: "Added to \`WindowAttributes\` the attributes  \`center_window: bool\` and \`start_monitor: i16`. Window centering is by default enabled on windows."

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
...
+- Added convenience functions to manage window centering, `center()` ,`set_center_monitor()`, `set_centered()` see more in window struct.
```
ZeGentzy: Change to this instead:
"Added the following convenience functions for managing window centering: `center()` ,`set_center_monitor()`, and `set_centered()`. Refer to the documentation for how to use them."

src/platform/windows/window.rs
```diff
@@ -715,15 +715,43 @@ unsafe fn init(
...
 
     if let Some(dimensions) = attributes.dimensions {
         win.set_inner_size(dimensions);
     }
 
+    if attributes.center_window && !&attributes.fullscreen.is_some(){
+        let window_size = win.get_outer_size();
+
+        let monitor;
+        if attributes.start_monitor == -1 {
```
ZeGentzy: Why is `start_monitor` an `i16`? This should be an `Option<MonitorId>` instead.

src/window.rs
```diff
@@ -380,6 +395,39 @@ impl Window {
...
     }
+        /// Centers the window on the current monitor the window lives in.
+    pub fn center(&self) {
+        self.set_centered(self.get_current_monitor())
+    }
+
+    /// Centers the window on the specified target monitor, falls back to primary monitor if out range.
+    pub fn set_center_monitor(&self, monitor_target: usize) {
+        if monitor_target < self.get_available_monitors().count() {
+            self.set_centered(self.get_available_monitors().nth(monitor_target).unwrap())
```
ZeGentzy: This should be accepting a `Option<MonitorId>`, not usize. 
`self.set_centered(monitor_target.unwrap_or_else(|_| self.get_primary_monitor()))`

src/window.rs
```diff
@@ -69,6 +69,21 @@ impl WindowBuilder {
         self
     }
 
+        /// Sets so the window will start centered on screen.
```
ZeGentzy: Why the weird formatting?
Also, "Sets so" -> "Sets whether or not"

src/window.rs
```diff
@@ -69,6 +69,21 @@ impl WindowBuilder {
...
+    #[inline]
+    pub fn with_centering(mut self, centered: bool) -> WindowBuilder {
+        self.window.center_window = centered;
+        self
+    }
+
+    /// Sets what monitor the window will start in.
```
ZeGentzy: "what" -> "which"

src/window.rs
```diff
@@ -69,6 +69,21 @@ impl WindowBuilder {
...
+    /// where -1 is the users primary monitor.
```
ZeGentzy: -1 -> \`-1`

src/window.rs
ZeGentzy: Turn the period at the end to a comma.

src/window.rs
ZeGentzy: users -> user's

src/platform/linux/x11/window.rs
```diff
@@ -380,6 +380,35 @@ impl UnownedWindow {
                 window.set_always_on_top_inner(window_attrs.always_on_top).queue();
             }
 
+            if window_attrs.center_window && !window_attrs.fullscreen.is_some() 
+            {
+                let window_size = window.get_outer_size();
+
+                let monitor;
+                if window_attrs.start_monitor == -1 {
```
ZeGentzy: As mentioned, this code would look better if you just stored a `MonitorId` here.

src/window.rs
mralve: well, the goal with set_center_monitor() is to only supply a simple monitor index, and if that is out of range then it falls back to the primary monitor, set_center_monitor is better renamed to set_target_monitor() and monitor_target > monitor_index to reflect that idea better. otherwise, if you want to use a MonitorID then simply use set_centered() instead.

src/platform/windows/window.rs
mralve: This is done for simplicity. only supplying a monitors index in display config is better IMO, start_monitor is renamed to monitor_index to better reflect that.

vberger: We don't really have any guarantee that the order will be the same across starts of the application, so configuring apps by monitor index cannot work reliably. Integrating such an API in winit would be a false promise that this would work.

FEATURES.md
```diff
@@ -88,7 +88,7 @@ If your PR makes notable changes to Winit's features, please update this section
 - **Popup / modal windows**: Windows can be created relative to the client area of other windows, and parent
   windows can be disabled in favor of popup windows. This feature also guarantees that popup windows
   get drawn above their owner.
-
+- **Window centering**: Sets the window position to be in the center of the given monitor.
```
vberger: If you add an entry here, you need to also add it to the compatibility matrix below.

src/window.rs
```diff
@@ -380,6 +395,40 @@ impl Window {
     pub fn set_decorations(&self, decorations: bool) {
         self.window.set_decorations(decorations)
     }
+    
+    /// Centers the window on the current monitor the window lives in.
```
vberger: Please specify the platform-dependent behavior of this function. For example it cannot work at all on Wayland, as the app cannot control its placement on the monitor.

src/window.rs
```diff
@@ -380,6 +395,40 @@ impl Window {
...
+    pub fn set_target_monitor(&self, monitor_index: usize) {
+        if monitor_index < self.get_available_monitors().count() {
+            self.set_centered(self.get_available_monitors().nth(monitor_index).unwrap())
+        } else {
+            self.set_centered(self.get_primary_monitor());
+        }
+    }
+
+    /// Sets the window position to be in the center of the given monitor.
+    /// Will do nothing if the window is in fullscreen.
```
vberger: Same here, this method behavior is platform dependent.

src/window.rs
```diff
@@ -380,6 +395,40 @@ impl Window {
...
+        let monitor_size = monitor.get_dimensions();
+        let monitor_position = monitor.get_position();
+
+        let mut monitor_window_position: LogicalPosition = (0.0, 0.0).into();
+        monitor_window_position.x =
+            monitor_position.x + (monitor_size.width * 0.5) - (&window_size.width * 0.5);
+        monitor_window_position.y =
+            monitor_position.y + (monitor_size.height * 0.5) - (&window_size.height * 0.5);
+
+        &self.set_position(monitor_window_position);
```
vberger:
```suggestion
        self.set_position(monitor_window_position);
```

src/platform/windows/window.rs
mralve: Ahh, I see, well then it's better to drop the set_center_monitor() function and use the Option<MonitorId> directly in the display config?

src/window.rs
mralve: dropping set_center_monitor() as set_centered() already do this.

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
-
+- Added to `WindowAttributes` the attributes `center_window: bool` and `start_monitor: i16`. Window centering is by default enabled.
+- Added the following convenience functions for managing window centering: center(), and set_centered(). Refer to the documentation for how to use them.
```
ZeGentzy: Oh, I forgot. We should surround the two functions with backticks (`)

src/platform/linux/x11/window.rs
```diff
@@ -380,6 +380,34 @@ impl UnownedWindow {
...
             }
 
+            if window_attrs.center_window && !window_attrs.fullscreen.is_some() 
+            {
+                let window_size = window.get_outer_size();
+
+                let monitor_size;
+                let monitor_position;
+
+                if window_attrs.start_monitor.is_none() {
```
ZeGentzy: Haven't tested if this compiles, but how about something like this?
```
let monitor = match window_attrs.start_monitor {
    None => window.get_primary_monitor(),
    Some(m) => m.inner,
};

monitor_size = monitor.get_dimensions();
monitor_position = monitor.get_position();
```

src/platform/linux/x11/window.rs
```diff
@@ -380,6 +380,34 @@ impl UnownedWindow {
...
+                    monitor_size = monitor.get_dimensions();
+                    monitor_position = monitor.get_position();
+                }
+
+
+                let mut monitor_window_position: LogicalPosition = (0.0, 0.0).into();
+                monitor_window_position.x =
+                    monitor_position.x + (monitor_size.width * 0.5) - (&window_size.unwrap().width * 0.5);
+                monitor_window_position.y =
+                    monitor_position.y + (monitor_size.height * 0.5) - (&window_size.unwrap().height * 0.5);
```
ZeGentzy: What if `monitor_position.{x,y}` != 0? Won't this not centre the window?

src/platform/windows/window.rs
```diff
@@ -715,15 +715,39 @@ unsafe fn init(
...
     }
 
     if let Some(dimensions) = attributes.dimensions {
         win.set_inner_size(dimensions);
     }
 
+    if attributes.center_window && !&attributes.fullscreen.is_some(){
+        let window_size = win.get_outer_size();
+
+        let monitor;
```
ZeGentzy: Haven't tested if this compiles, but how about something like this?
```
let monitor = match window_attrs.start_monitor {
    None => window.get_primary_monitor(),
    Some(m) => m.inner,
};
```

src/platform/windows/window.rs
```diff
@@ -715,15 +715,39 @@ unsafe fn init(
...
+        } else {
+            monitor = attributes.start_monitor.unwrap().inner;
+        }
+
+        let monitor_size = monitor.get_dimensions();
+        let monitor_position = monitor.get_position();
+
+
+        let mut monitor_window_position: LogicalPosition = (0.0, 0.0).into();
+        monitor_window_position.x =
```
ZeGentzy: What if `monitor_position.{x,y}` != 0? Won't this not centre the window?

src/window.rs
```diff
@@ -380,6 +395,45 @@ impl Window {
     pub fn set_decorations(&self, decorations: bool) {
         self.window.set_decorations(decorations)
     }
+    
+    /// Centers the window on the current monitor the window lives in.
+    /// 
+    /// ## Platform-specific
+    /// 
+    /// This only works on Windows and X11.
+    /// it's not tested on macOS.
```
ZeGentzy: You mean "It's not implemented on MacOS"? I don't see a MacOS impl anywhere.

At minimum, "it's" -> "It's"

src/window.rs
```diff
@@ -380,6 +395,45 @@ impl Window {
...
+        self.set_centered(self.get_current_monitor())
+    }
+
+    /// Sets the window position to be in the center of the given monitor.
+    /// Will do nothing if the window is in fullscreen.
+    /// 
+    /// ## Platform-specific
+    /// 
+    /// This only works on Windows and X11.
+    /// it's not tested on macOS.
```
ZeGentzy: Same as above.

src/window.rs
```diff
@@ -380,6 +395,45 @@ impl Window {
...
+        };
+
+        let monitor_size = monitor.get_dimensions();
+        let monitor_position = monitor.get_position();
+
+        let mut monitor_window_position: LogicalPosition = (0.0, 0.0).into();
+        monitor_window_position.x =
+            monitor_position.x + (monitor_size.width * 0.5) - (&window_size.width * 0.5);
+        monitor_window_position.y =
+            monitor_position.y + (monitor_size.height * 0.5) - (&window_size.height * 0.5);
```
ZeGentzy: Same as above.

src/window.rs
ZeGentzy: Nvm, I guess the default impls bellow act as MacOS's.

Anyways, "it's" -> "It's", "macOS" -> "MacOS"

src/platform/linux/x11/window.rs
mralve: I don't see any problem here, it centers on both of my monitors. have tried entering custom values to test, but that works as expected.

src/platform/windows/window.rs
mralve: ahh right, matching is so nice, I'm just too used to write ifs, coming from c#

src/platform/linux/x11/window.rs
ZeGentzy: Ah, I thought `monitor_position` was the window's position on the monitor. Looking at the code again, I realize I was mistaken.

src/lib.rs
```diff
@@ -531,6 +541,8 @@ impl Default for WindowAttributes {
             title: "winit window".to_owned(),
             maximized: false,
             fullscreen: None,
+            center_window: true,
```
aloucks: I don't think it's right to set `center_window: true` as the default behavior. Let the platform decide where to put it. This is better off as opt-in behavior.

src/lib.rs
```diff
@@ -510,6 +510,16 @@ pub struct WindowAttributes {
...
 
+    /// Whether the window will be centered on screen.
+    ///
+    /// The default is `true`.
+    pub center_window: bool,
+
+    /// Decides which monitor the window will start on. The user's primary monitor is `None`.
+    ///
+    /// The default is `None`,
+    pub start_monitor: Option<MonitorId>,
```
aloucks: I don't think it's unreasonable to require the user to select the monitor first (e.g. `EventsLoop::get_primary_monitor()`).

Why not remove `start_monitor` and only have:

```rust
center_window: Option<MonitorId>
```

src/window.rs
```diff
@@ -380,6 +395,45 @@ impl Window {
...
+    
+    /// Centers the window on the current monitor the window lives in.
+    /// 
+    /// ## Platform-specific
+    /// 
+    /// This only works on Windows and X11.
+    /// It's not tested on macOS.
+    /// 
+    /// This has no effect on Android, Wayland or iOS.
+    pub fn center(&self) {
```
aloucks: Why is this needed in addition to `set_centered`? The naming of state changing functions is already inconsistent (see also #728). I think having multiple ways to do the same thing is probably not a great idea, especially when getting the default monitor is already really easy.

src/window.rs
```diff
@@ -380,6 +395,45 @@ impl Window {
     pub fn set_decorations(&self, decorations: bool) {
         self.window.set_decorations(decorations)
     }
+    
+    /// Centers the window on the current monitor the window lives in.
+    /// 
+    /// ## Platform-specific
+    /// 
+    /// This only works on Windows and X11.
+    /// It's not tested on macOS.
```
Osspial: It's misleading to say that this isn't tested on macOS - it just isn't implemented.

src/window.rs
```diff
@@ -380,6 +395,45 @@ impl Window {
...
+        self.set_centered(self.get_current_monitor())
+    }
+
+    /// Sets the window position to be in the center of the given monitor.
+    /// Will do nothing if the window is in fullscreen.
+    /// 
+    /// ## Platform-specific
+    /// 
+    /// This only works on Windows and X11.
+    /// It's not tested on macOS.
```
Osspial: See previous comment on macOS.

FEATURES.md
```diff
@@ -154,6 +154,7 @@ Legend:
 |Window transparency              |✔️     |✔️     |✔️         |✔️             |**N/A**|**N/A**|**N/A**   |
 |Window maximization              |✔️     |✔️     |✔️         |✔️             |**N/A**|**N/A**|**N/A**   |
 |Window maximization toggle       |✔️     |✔️     |✔️         |✔️             |**N/A**|**N/A**|**N/A**   |
+|Window centering                 |✔️     |❌     |✔️         |❌             |**N/A**|**N/A**|**N/A**   |
```
Osspial: This should *probably* be **N/A** on Wayland, since it doesn't seem like the sort of thing Wayland would support. cc @vberger

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
-
+- Added to `WindowAttributes` the attributes `center_window: bool` and `start_monitor: i16`. Window centering is by default enabled.
```
Osspial: `start_monitor` is `Option<MonitorId>`, not `i16`. Window centering should default to being disabled.

Osspial:
On a higher level, I'd like to ask: why is this PR necessary? The [`set_centered` method's implementation](https://github.com/rust-windowing/winit/blob/e1edc1d6b7e42b8d7632b18dfd87f5c75b831e63/src/window.rs#L420-L436) shows that it's entirely possible to implement this without access to any of Winit's internals, and bringing this into our codebase increases the codebase's complexity. Admittedly, It's not a substantial increase in this case, but it sets an uncomfortable prescient for pulling in code that's *convenient* but that can be implemented entirely outside of Winit.

mralve:
@Osspial Whilst you bring up great points I would understand if you reject this pr, But I would argue that for a "window creation and management library." as stated in the readme, a library that provides basic functionality for window creation and management, being able to control the window starting monitor, and if it's going to be centered on the screen out of the box is seen as rather basic IMO, while yes being convenient i would see this as clearing up unnecessary redundancy in the view from multiple projects needing this trivial functionality.

Osspial:
So, it seems like I was lacking a bit of nuance with that initial response! To clarify myself: I'm entirely okay with adding a field in `WindowAttributes` to set the window's default monitor. I just don't like bringing `set_centered` into Winit's codebase, since it can be entirely implemented with publicly-visible, non-platform-specific API calls.

Also, given the above I'd rather refer to this feature as *default monitor* rather than *window centering*, since that exposes looser API guarantees and opens the door for the API to be implemented on platforms that have multiple displays but not necessarily a desktop surface.

mralve:
Great having a default monitor property, would kinda require window centering by default unless there is better a way to give the os responsibility to place the window on the right screen and by its deafult window position?

src/platform/windows/window.rs
```diff
@@ -715,15 +715,37 @@ unsafe fn init(
...
+    if let Some(_) = &attributes.fullscreen {
+        win.set_fullscreen(attributes.fullscreen.clone());
         force_window_active(win.window.0);
     }
 
     if let Some(dimensions) = attributes.dimensions {
         win.set_inner_size(dimensions);
     }
 
+    if attributes.center_window.is_some() && !&attributes.fullscreen.is_some(){
```
ZeGentzy: We can replace all this code with something along the lines of this, right (pseudo code):
```
if attributes.fullscreen.is_none() {
    if let Some(mon) = attributes.center_window {
        win.set_centered(mon);
    }
}
```

src/platform/linux/x11/window.rs
```diff
@@ -380,6 +380,27 @@ impl UnownedWindow {
                 window.set_always_on_top_inner(window_attrs.always_on_top).queue();
             }
 
+            if window_attrs.center_window.is_some() && !window_attrs.fullscreen.is_some()
```
ZeGentzy: https://github.com/rust-windowing/winit/pull/857/files#r286254032

ZeGentzy:
I'll chip in: maybe instead of `set_centered` we could rename it to `set_monitor`, and instead of `center_window`, `default_monitor`, as suggested. 

The new function will by default try to place the window in the center. If that's not possible, just anywhere on the monitor will do. Not sure if the latter is possible on Wayland, but if it's not, I guess it can be a no-op.

ZeGentzy:
Closing for inactivity.

PR #868, opened 2019-05-17
[WIP] Add support for OS specific message.

crlf0710:

### DIFF
--- a/src/platform/windows.rs
+++ b/src/platform/windows.rs
@@ -126,31 +126,16 @@ impl DeviceIdExtWindows for DeviceId {
 #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct OsSpecificWindowEvent {
     pub(crate) window: HWND,
-    pub(crate) msg: UINT,
-    pub(crate) wparam: WPARAM,
-    pub(crate) lparam: LPARAM,
+    /// The message identifier of the window event (`WM_*`).
+    pub message: UINT,
+    /// The first parameter of the window event (`WPARAM`).
+    pub wparam: WPARAM,
+    /// The second parameter of the window event(`LPARAM`).
+    pub lparam: LPARAM,
     pub(crate) retval: *mut Option<LRESULT>,
 }
 
 impl OsSpecificWindowEvent {
-    /// The message identifier of the window event (`WM_*`).
-    #[inline]
-    pub fn message(&self) -> u32 {
-        self.msg
-    }
-
-    /// The first parameter of the window event (`WPARAM`).
-    #[inline]
-    pub fn wparam(&self) -> usize {
-        self.wparam
-    }
-
-    /// The second parameter of the window event(`LPARAM`).
-    #[inline]
-    pub fn lparam(&self) -> isize {
-        self.lparam
-    }
-
     /// Marks the current event processed, returning the specified value
     /// as the result of the window procedure to the system.
     /// How the value is interpreted is dependent on the message's type.
@@ -177,7 +162,7 @@ impl OsSpecificWindowEvent {
     #[inline]
     pub unsafe fn set_overriden_reply(&self, f: impl FnOnce(isize) -> isize) {
         use winapi::um::commctrl;
-        let retval = f(commctrl::DefSubclassProc(self.window, self.msg, self.wparam, self.lparam));
+        let retval = f(commctrl::DefSubclassProc(self.window, self.message, self.wparam, self.lparam));
         *self.retval = Some(retval);
     }
 }
--- a/src/platform/windows.rs
+++ b/src/platform/windows.rs
@@ -159,6 +159,10 @@ impl OsSpecificWindowEvent {
     /// Calling this function outside this event's call to the closure passed to
     /// [`EventLoop::run`](../../event_loop/struct.EventLoop.html#method.run)
     /// will result in undefined behavior.
+    ///
+    /// The current `message`, `wparam`, `lparam` values within this struct will be
+    /// used. Calling this function after setting these to incorrect
+    /// values will result in undefined behavior.
     #[inline]
     pub unsafe fn set_overriden_reply(&self, f: impl FnOnce(isize) -> isize) {
         use winapi::um::commctrl;
--- a/src/platform/windows.rs
+++ b/src/platform/windows.rs
@@ -4,6 +4,9 @@ use std::os::raw::c_void;
 
 use libc;
 use winapi::shared::windef::HWND;
+use winapi::shared::minwindef::{UINT, WPARAM, LPARAM, LRESULT};
+
+use std::fmt;
 
 use event::DeviceId;
 use monitor::MonitorHandle;
@@ -116,5 +119,72 @@ impl DeviceIdExtWindows for DeviceId {
     #[inline]
     fn get_persistent_identifier(&self) -> Option<String> {
         self.0.get_persistent_identifier()
+    }
+}
+
+/// Unprocessed window event that are specific to Windows.
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct OsSpecificWindowEvent {
+    pub(crate) window: HWND,
+    pub(crate) msg: UINT,
+    pub(crate) wparam: WPARAM,
+    pub(crate) lparam: LPARAM,
+    pub(crate) retval: *mut Option<LRESULT>,
+}
+
+impl OsSpecificWindowEvent {
+    /// The message identifier of the window event (`WM_*`).
+    #[inline]
+    pub fn message(&self) -> u32 {
+        self.msg
+    }
+
+    /// The first parameter of the window event (`WPARAM`).
+    #[inline]
+    pub fn wparam(&self) -> usize {
+        self.wparam
+    }
+
+    /// The second parameter of the window event(`LPARAM`).
+    #[inline]
+    pub fn lparam(&self) -> isize {
+        self.lparam
+    }
+
+    /// Marks the current event processed, returning the specified value
+    /// as the result of the window procedure to the system.
+    /// How the value is interpreted is dependent on the message's type.
+    /// The next window procedure (or subclass window
+    /// procedure) won't be called for this message.
+    ///
+    /// Calling this function outside this event's call to the closure passed to
+    /// [`EventLoop::run`](../../event_loop/struct.EventLoop.html#method.run)
+    /// will result in undefined behavior.
+    #[inline]
+    pub unsafe fn set_reply(&self, val: isize) {
+        *self.retval = Some(val);
+    }
+
+    /// Immediately calls the next window procedure(or subclass window
+    /// procedure), and maps its return value through the user-passed closure.
+    /// How the value is interpreted is dependent on the message's type.
+    ///
+    /// This function must not be called more than once for each message.
+    ///
+    /// Calling this function outside this event's call to the closure passed to
+    /// [`EventLoop::run`](../../event_loop/struct.EventLoop.html#method.run)
+    /// will result in undefined behavior.
+    #[inline]
+    pub unsafe fn set_overriden_reply(&self, f: impl FnOnce(isize) -> isize) {
+        use winapi::um::commctrl;
+        let retval = f(commctrl::DefSubclassProc(self.window, self.msg, self.wparam, self.lparam));
+        *self.retval = Some(retval);
+    }
+}
+
+
+impl fmt::Debug for OsSpecificWindowEvent {
+    fn fmt(&self, fmtr: &mut fmt::Formatter) -> fmt::Result {
+        fmtr.pad("OsSpecificWindowEvent { .. }")
     }
 }
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -1497,7 +1497,25 @@ unsafe extern "system" fn public_window_callback<T>(
                 );
                 0
             } else {
-                commctrl::DefSubclassProc(window, msg, wparam, lparam)
+                use platform_impl::platform::OsSpecificWindowEvent;
+                let mut retval: Option<LRESULT> = None;
+
+                subclass_input.send_event(Event::WindowEvent {
+                    window_id: RootWindowId(WindowId(window)),
+                    event: WindowEvent::OsSpecific(OsSpecificWindowEvent {
+                        window,
+                        msg,
+                        wparam,
+                        lparam,
+                        retval: &mut retval as _,
+                    }),
+                });
+
+                if let Some(retval) = retval {
+                    retval
+                } else {
+                    commctrl::DefSubclassProc(window, msg, wparam, lparam)
+                }
             }
         }
     }
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -1504,7 +1504,7 @@ unsafe extern "system" fn public_window_callback<T>(
                     window_id: RootWindowId(WindowId(window)),
                     event: WindowEvent::OsSpecific(OsSpecificWindowEvent {
                         window,
-                        msg,
+                        message: msg,
                         wparam,
                         lparam,
                         retval: &mut retval as _,
--- a/src/platform_impl/windows/mod.rs
+++ b/src/platform_impl/windows/mod.rs
@@ -65,6 +65,8 @@ impl WindowId {
     }
 }
 
+pub type OsSpecificWindowEvent = crate::platform::windows::OsSpecificWindowEvent;
+
 mod dpi;
 mod drop_handler;
 mod event;
--- a/src/event.rs
+++ b/src/event.rs
@@ -177,6 +177,9 @@ pub enum WindowEvent {
     ///
     /// For more information about DPI in general, see the [`dpi`](dpi/index.html) module.
     HiDpiFactorChanged(f64),
+
+    /// Subscribed OS specific message.
+    OsSpecific(platform_impl::OsSpecificWindowEvent),
 }
 
 /// Identifier of an input device.
@@ -499,7 +502,7 @@ pub enum VirtualKeyCode {
     Multiply,
     Mute,
     MyComputer,
-    NavigateForward, // also called "Prior"
+    NavigateForward,  // also called "Prior"
     NavigateBackward, // also called "Next"
     NextTrack,
     NoConvert,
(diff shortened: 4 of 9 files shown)


crlf0710:
Hi, this PR is not fully ready yet, but i'd like to see whether everyone supports going in this direction.

src/platform/windows.rs
```diff
@@ -104,6 +105,36 @@ impl MonitorHandleExtWindows for MonitorHandle {
     }
 }
 
+/// Additional methods on `OsSpecificWindowEvent` that are specific to Windows.
+pub trait OsSpecificWindowEventExtWindows {
```
ZeGentzy: I'd rather not expose extensions.

Here is my proposal:

In `src/platform_impl/windows/mod.rs` we say `pub type OsSpecificWindowEvent = platform::windows::OsSpecificWindowEvent`.

We then define `OsSpecific(OsSpecificWindowEvent),` as `OsSpecific(platform_impl::OsSpecificWindowEvent),` and get rid of the `OsSpecificWindowEvent` in `src/events.rs`.

In this file we can then define `OsSpecificWindowEvent` as something like following (forgive the fake types):
```
pub enum Command {
    Menu(id),
    Accelerator(id),
    Control(code, id, handle),
}

pub enum OsSpecificWindowEvent {
    Command(Command),
    Notify(id, NMHDR_ptr)
}
```

crlf0710: Hi, this will introduce quite a few concepts that `winit` itself consider out of scope(menus, controls, etc), and will grow more and more(and introduce lots of breaking change) whenever downstream crates has new requirements, like handling control colors and owner draws etc. Is it ok to go this way?

crlf0710: Just for your interest: I've just started https://github.com/crlf0710/winit-featurepack to add os-specific missing features for `winit` for those functionality excluded.

ZeGentzy: Ah, I wasn't exactly sure what those two commands were for. If that's the case, instead of `
pub enum OsSpecificWindowEvent`, we can have this:
```
pub struct OsSpecificWindowEvent {
    message: UINT,
    wparam: WPARAM,
    lparam: LPARAM,
}
```

src/platform_impl/windows/event_loop.rs
```diff
@@ -1403,6 +1403,22 @@ unsafe extern "system" fn public_window_callback<T>(
 
             0
         },
+        winuser::WM_COMMAND | winuser::WM_NOTIFY => {
```
ZeGentzy: Why only expose these two events? Why not move this to the end of the match statement and have `_ => {`?

src/platform/windows.rs
crlf0710: ok

src/platform_impl/windows/event_loop.rs
crlf0710: because... well, the wndproc is a complex mechanism. for many messages, return value is a "reply" to the system/caller.  There's the `DefWindowProc` function which generates the reply for us, but user will want to override this behavior for some messages. If we handle everything now, i think we'll need to invent a reply mechanism API now.

Maybe i was just lazy. i think i will give it a try.

Osspial:
FWIW, this functionality is already exposed on Windows through the native [Window Subclassing APIs](https://docs.microsoft.com/en-us/windows/desktop/Controls/subclassing-overview#subclassing_v6). That's a more flexible solution since it lets you completely override and modify all events a window receives, and doesn't require API design work on our end.

crlf0710:
@Osspial I might be wrong, but my argument is this:
1. It's certainly not zero cost to subclass a window, and even more so if you subclass it twice, once within winit, and once outside.
2. I believe one large part of the core of Rust-style API design is encapsulating `push`-style APIs and providing `poll`-style APIs - change them into queues, filters and/or iterators.  While it's ok to subclass the window another time outside the scope of winit, a new SUBCLASSPROC is another `push`-style API entry-point, and it's hard if not impossible to make the SUBCLASSPROC reuse the existing event queue of winit itself.
It definitely hurts ergonomics when the downstream crate needs to deal with the two queues together - largely eliminating the whole point of using the `winit` crate.

crlf0710:
So i've update the PR to remove the additional trait. PTAL.

src/platform_impl/windows/mod.rs
```diff
@@ -65,6 +68,51 @@ impl WindowId {
...
+    window: HWND,
+    msg: UINT,
+    wparam: WPARAM,
+    lparam: LPARAM,
+    retval: *mut Option<LRESULT>,
+}
+
+impl OsSpecificWindowEvent {
+    #[inline]
+    pub fn message(&self) -> u32 {
```
ZeGentzy: What's the point of these setter-getters? Why not just make the member public? The only one which I think makes sense is `set_overriden_reply`, since it looks non-trivial, however, it's got no documentation that describes what it does!

src/platform_impl/windows/mod.rs
```diff
@@ -65,6 +68,51 @@ impl WindowId {
     }
 }
 
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct OsSpecificWindowEvent {
+    window: HWND,
+    msg: UINT,
+    wparam: WPARAM,
+    lparam: LPARAM,
+    retval: *mut Option<LRESULT>,
```
ZeGentzy: Why `*mut` and not `&mut`? Sounds like a recipe for disaster, and `winit` tries to be as safe as it can be?

crlf0710: @ZeGentzy  I know, but if we add a `'a` there, then we'll get `'a` everywhere (`WindowEvent`, `Event`, ...) and on every platform, i think?

crlf0710: And currently this field is only accessed from `unsafe fn` members btw.

src/platform_impl/windows/mod.rs
crlf0710: yes, maybe i should just expose them.

crlf0710: Public fields does make more consistency with current `winit` apis, but after i start to make these member public, it occurs to me that then people will be able to set up random values for these fields, and use the unsafe fn `set_overriden_reply` to send them to the system API. (Which may crash the process.) While nobody will likely do this, does it make better sense to keep them read-only (only readable with getter) in the current way?

crlf0710:
Added some doc comments. I'm not english native speaker, feel free to point out the wording that needs to be improved.

src/platform_impl/windows/mod.rs
crlf0710: Also, `&'a mut T` prevents cloning and copying.

src/platform_impl/windows/mod.rs
ZeGentzy: Users fiddling with the `window`, `msg` `wparam` and `lparam` fields shouldn't do anything, right, since they are all copy?

Don't the values get implicitly cloned [here](https://github.com/rust-windowing/winit/pull/868/files#diff-d4d8c8dcf07874a35d08c6a70e0a6dafR1505), and hen don't the originals get sent [here](https://github.com/rust-windowing/winit/pull/868/files#diff-d4d8c8dcf07874a35d08c6a70e0a6dafR1505).

The only members which users fiddling with can do damage is `retval`, and I guess we can keep the getter/setter for that (and it is unsafe).

ZeGentzy: I'd rather `retval` becomes `*mut LRESULT,`, and have it set by default to `commctrl::DefSubclassProc(self.window, self.msg, self.wparam, self.lparam)`. Then we can just have this:
```
    #[inline]
    pub unsafe fn set_reply(&self, val: isize) {
        *self.retval = val;
    }

    #[inline]
    pub unsafe fn set_overriden_reply(&self, f: impl FnOnce(isize) -> isize) {
        use winapi::um::commctrl;
        let retval = f(*self.retval);
        *self.retval = Some(retval);
    }
}
```

If users want to improperly fiddle with unsafe functions, that's their call.

ZeGentzy: If we do that, we only have to return `retval` [here](https://github.com/rust-windowing/winit/pull/868/files#diff-d4d8c8dcf07874a35d08c6a70e0a6dafR1514)

src/platform_impl/windows/mod.rs
```diff
@@ -65,6 +68,72 @@ impl WindowId {
...
+        self.wparam
+    }
+
+    /// The second parameter of the window event(`LPARAM`).
+    #[inline]
+    pub fn lparam(&self) -> isize {
+        self.lparam
+    }
+
+    /// Mark the current event processed, returning specified value
```
ZeGentzy: Mark -> Marks
returning -> returning the

The value will be interpreted according to the meaning of the message. -> How the value is interpreted is dependent on the message's type.

It is undefined behavior to call this function unless during the event loop processing this corresponding event. -> ```Calling this function outside this event's call to [`event_handler`] will result in undefined behavior.```

Then append:
```[`event_handler`]: ../I/don't/know/the/proper/address/:)/../struct.EventLoop.html#method.run_whatever_the_function_name_was```

File: src/platform_impl/windows/mod.rs (hunk 3 above)
Comment (ZeGentzy) [reply src/platform_impl/windows/mod.rs]: Hmmm, reading over the docs you wrote, I was under the assumption that `commctrl::DefSubclassProc` was some mapping function, and not a syscall.

We should keep `retval` as an `Option` and just let users fiddle with them all they want. We can then state in the docs for `set_overriden_reply` that fiddling with them incorrectly could result in UB.

File: src/platform_impl/windows/mod.rs [hunk 6]
```diff
@@ -65,6 +68,72 @@ impl WindowId {
...
+    /// procedure) won't be called for this message.
+    ///
+    /// It is undefined behavior to call this function unless during the
+    /// event loop processing this corresponding event.
+    #[inline]
+    pub unsafe fn set_reply(&self, val: isize) {
+        *self.retval = Some(val);
+    }
+
+    /// Immediately call the next window procedure(or subclass window
```
ZeGentzy: call -> calls
map -> maps
a closure -> the user-passed closure.
The value will be interpreted according to the meaning of the message. -> How the value is interpreted is dependent on the message's type.

It is undefined behavior to call this function unless during the event loop processing this corresponding event. -> ```Calling this function outside this event's call to [`event_handler`] will result in undefined behavior.```

Then append:
```[`event_handler`]: ../I/don't/know/the/proper/address/:)/../struct.EventLoop.html#method.run_whatever_the_function_name_was```

File: src/platform_impl/windows/mod.rs (hunk 3 above)
Comment (crlf0710) [reply src/platform_impl/windows/mod.rs]: All `commctrl::DefSubclassProc()` does is "call the next handler, or default handler, if any", which at most times is what user want but sometimes is what user want to avoid(for example, the default handler might fill the window background with a color, causing flash). And when the user calls it, user may still decide to change the value it returns (this potential change is what i meant with the word mapping). Maybe the function names needs some bike shedding too?

File: src/platform_impl/windows/mod.rs (hunk 5 above)
Comment (crlf0710) [reply src/platform_impl/windows/mod.rs]: done. a few more lines are added. ptal again.

File: src/platform_impl/windows/mod.rs (hunk 3 above)
Comment (crlf0710) [reply src/platform_impl/windows/mod.rs]: Changed these fields to public.

File: src/platform/windows.rs [hunk 7]
```diff
@@ -118,3 +121,55 @@ impl DeviceIdExtWindows for DeviceId {
...
+
+impl OsSpecificWindowEvent {
+    /// Marks the current event processed, returning the specified value
+    /// as the result of the window procedure to the system.
+    /// How the value is interpreted is dependent on the message's type.
+    /// The next window procedure (or subclass window
+    /// procedure) won't be called for this message.
+    ///
+    /// Calling this function outside this event's call to the closure passed to
+    /// [`EventLoop::run`](../../event_loop/struct.EventLoop.html#method.run)
```
ZeGentzy: You sure this works? I swear that you have to do this:
```
    /// Calling this function outside this event's call to the closure passed to
    /// [`EventLoop::run`] will result in undefined behavior.
    ///
    ///[`EventLoop::run`]: (../../event_loop/struct.EventLoop.html#method.run)
```

src/platform/windows.rs
```diff
@@ -118,3 +121,55 @@ impl DeviceIdExtWindows for DeviceId {
...
+    pub unsafe fn set_reply(&self, val: isize) {
+        *self.retval = Some(val);
+    }
+
+    /// Immediately calls the next window procedure(or subclass window
+    /// procedure), and maps its return value through the user-passed closure.
+    /// How the value is interpreted is dependent on the message's type.
+    ///
+    /// This function must not be called more than once for each message.
+    ///
```
ZeGentzy: Can we mention that improperly messing with the params or messages could result in UB?

src/platform/windows.rs
crlf0710: Yes, it does. i tested it locally both with stable and nightly compilers. i'm using 1.34.0. Not sure about earlier compiler versions though.

src/platform/windows.rs
crlf0710: Sure, will add it this evening(in my timezone)

crlf0710: done.

crlf0710: done

src/platform_impl/windows/mod.rs
Osspial: FWIW, we're probably going to add a lifetime to `WindowEvent` eventually, since there are some events that can only logically be handled through borrowing.

Osspial: Also, I realize now that this actually is a pretty big safety footgun. See, `send_event` doesn't always dispatch an event immediately: if `send_event` is called inside an active invocation of the user's event callback (which happens with certain `Window` functions), the event will get buffered and dispatched after the active callback invocation returns. If that happens when handling this type of event, `retval` will point to an invalid value when it's passed to the user's callback, and attempts to modify it will result in undefined behavior. The current API doesn't have any mechanisms to prevent that from happening.

Osspial:
This implementation seems mostly sound, barring the issue with `retval` mentioned above. However, while I'm not opposed to adding platform-specific window events, but I don't like the current method of doing it - since the current implementation publicly exposes every event `public_window_callback` doesn't explicitly handle, it makes handling a new window event within Winit a breaking change (e.g. if a downstream user relies on `WM_SETTEXT` being exposed, and then we start handling `WM_SETTEXT` within `public_window_callback`, the downstream code no longer works due to our internal code change). That's a pretty heavy burden for our codebase to carry, and it's not a design I'm comfortable exposing.

I'd be fine with merging a design that only exposes a select few events through an `OsSpecificWindowEvent` enum, since that's much easier to manage on our end.

-------

> It's certainly not zero cost to subclass a window, and even more so if you subclass it twice, once within winit, and once outside.

It isn't zero-cost, but the API was explicitly designed for layering subclasses, so the cost should be negligible.

>  While it's ok to subclass the window another time outside the scope of winit, a new SUBCLASSPROC is another push-style API entry-point, and it's hard if not impossible to make the SUBCLASSPROC reuse the existing event queue of winit itself.

Actually, it's fairly easy to do! You can use `EventLoopProxy::send_event` to dispatch arbitrary user-defined events to Winit's event queue.

src/platform_impl/windows/mod.rs
crlf0710: This will apply the same to the wparam and lparam arguments. If the handling is deferred, wparam and lparam (NMHDR or any pointers) will no longer be usable.

Is there anyway to prevent such message handling deferring?

crlf0710:
> Actually, it's fairly easy to do! You can use EventLoopProxy::send_event to dispatch arbitrary user-defined events to Winit's event queue.

Well, no. `EventLoopProxy::send_event` is not available for use to me if i'm a thin wrapper over winit. If i'm a more heavy-weight framework, i can provide my `EventLoop` above winit's `EventLoop`, and always specify `winit`'s `UserEvent` for my own use. In that way i'm managing all the `UserEvent`s so there will be no problem.

But if i'm a thin wrapper, i don't want to interfere with `UserEvent` at all, and i'll leave it completely to the user. I'm just wanting to inject more extra messages into winit message queue in a supplementary way. Does this make sense?

src/platform_impl/windows/mod.rs
Osspial: You could add a new variant of `send_event` that doesn't buffer events it can't dispatch. However, we can't get rid of deferred dispatch entirely, since it prevents the user callback from getting double-borrowed.

Osspial:
@crlf0710 I can understand what you're saying there. However, I still stand by what I said previously - forwarding all unhandled win32 events into the standard Winit event callback exposes the structure of Winit's internals to such a degree that I'm uncomfortable merging it.

In regards to using `UserEvent` - could you structure your thin wrapper code like this?

```rust
pub enum MyExtendedEvent {/*variants*/}

pub fn extend_winit_events<T>(winit_event_loop: &EventLoop<T>, winit_window: Window)
    where T: 'static + From<MyExtendedEvent>
{
    /* extract `EventLoopProxy` from `winit_event_loop` and subclass `window` */
}

struct SubclassInput<T: 'static + From<MyExtendedEvent>> {
    p: EventLoopProxy<T>
}

unsafe extern "system" fn my_subclass_callback<T>(
    window: HWND,
    msg: UINT,
    wparam: WPARAM,
    lparam: LPARAM,
    _: UINT_PTR,
    subclass_input_ptr: DWORD_PTR
) -> LRESULT
    where T: 'static + From<MyExtendedEvent>
{
    let subclass_input = &*(subclass_input_ptr as *const SubclassInput<T>);

    let event: MyExtendedEvent = /* generate event */;
    subclass_input.p.send_event(T::from(event));
}
```

That way, you can send custom events into Winit's event queue, users can still have their own custom events (by wrapping around your events), and Winit doesn't have to worry about which win32 events get publicly exposed.

crlf0710:
@Osspial Thank you, i think this can work, although it's still a little strange to user though. 

I still want a way to add a small "extension point" to `winit` itself, to make out-of-scopes features naturally integrated. However it's not clear to me what that would look like, so i'll do some experiments myself first. If i come up with something i myself am satisfied with, i'll open a PR again.

At the same time, it's quite obvious that the current approach just went into its own dead end. So i'll just go ahead and close this. Thank you @Osspial and @ZeGentzy for your time.

Osspial:
@crlf0710 Glad to help out! I'll be happy to see what you come up with for an extension API - since we're pushing for a small Winit as a base building block, I'd definitely like to make it as easy as possible to build new mechanisms on top of it.

PR #884, opened 2019-05-25
Update parking_lot to 0.8 for windows too (eventloop-2.0)

aloucks:
The previous attempt to update parking_lot missed the windows platform.

The parking_lot dependency is now no longer specified twice to help prevent that mistake from happening again.


aloucks:
This fixes the linux build errors from #882 (https://travis-ci.org/rust-windowing/winit/builds/537194992) and the macos compile error is fixed in #885.

PR #886, opened 2019-05-26
Refine function names and type signatures

### LINKED ISSUE #728: Function naming inconsistency

### LINKED ISSUE #794: Replace Option<T> in Window function return types with T
In #434, we decided that an active `Window` struct ideally should only represent an open window - the only way of closing the window is to drop it. This stance is contrary to several of the `Window` function APIs that return `Option<T>` (specifically, `get_position`, `get_inner_position`, `get_inner_size`, and `get_outer_size`). The docs for those functions say that the functions will return `None` if the backing window no longer exists, but the API says that the backing window *must* exist if the `Window` struct exists.

As such, this issue tracks the removal of `Option` from those APIs. It also opens the question of what exactly we should do in the unlikely case that the OS *does* choose to close the window without our consent. As far as I can tell, there are three possible options:
* Panic
* Return the last known value
* Return a stub value

As I'm writing this I don't feel all that strongly about which option we take, but if nobody weighs in I'd be fine with making a choice. Since the Event Loop 2.0 release seems to be our *break absolutely everything* release I'd support rolling these changes into that update.

Osspial:
  - [x] Test on iOS
  - [x] Add `multitouch` changelog entry.

This PR addresses the function naming and type signature issues brought up in #728 and #794. It also improves the typing for our error types (namely, by introducing types in the first place) and makes our getter APIs follow the [Rust naming conventions](https://rust-lang-nursery.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter).

Closes #728, closes #794.


src/error.rs
```diff
@@ -0,0 +1,43 @@
...
+
+pub enum ExternalError {
+    NotSupported(NotSupportedError),
+    Os(OsError),
+}
+
+/// TODO: MANUALLY IMPLEMENT TRAITS
+#[derive(Debug)]
+pub struct NotSupportedError {
+    _marker: (),
```
ZeGentzy: What's the point of this marker?

src/lib.rs
```diff
@@ -115,6 +115,8 @@ extern crate smithay_client_toolkit as sctk;
 extern crate calloop;
 
 pub mod dpi;
+#[macro_use]
```
ZeGentzy: Do we really want to expose the macro? From the looks of it, the macro won't even compile if users use it.

src/platform/ios.rs
```diff
@@ -23,43 +23,28 @@ pub trait WindowExtIOS {
...
 
     /// Returns a pointer to the `UIView` that is used by this window.
     ///
     /// The pointer will become invalid when the `Window` is destroyed.
-    fn get_uiview(&self) -> *mut c_void;
-
-    /// Sets the HiDpi factor used by this window.
-    /// 
-    /// This translates to `-[UIWindow setContentScaleFactor:hidpi_factor]`.
-    fn set_hidpi_factor(&self, hidpi_factor: f64);
```
ZeGentzy: What happened to these?

src/platform_impl/ios/mod.rs
```diff
@@ -58,13 +58,194 @@
...
 #![cfg(target_os = "ios")]
 
-// TODO: (mtak-) UIKit requires main thread for virtually all function/method calls. This could be
-// worked around in the future by using GCD (grand central dispatch) and/or caching of values like
-// window size/position.
-macro_rules! assert_main_thread {
-    ($($t:tt)*) => {
-        if !msg_send![class!(NSThread), isMainThread] {
-            panic!($($t)*);
+use std::{fmt, mem, ptr};
```
ZeGentzy: All these changes look misplaced. git mistake?

src/platform_impl/ios/mod.rs
```diff
@@ -91,8 +272,346 @@ pub struct DeviceId {
 
 impl DeviceId {
     pub unsafe fn dummy() -> Self {
-        DeviceId {
-            uiscreen: std::ptr::null_mut(),
+        DeviceId
```
ZeGentzy: All these changes look misplaced. git mistake?

src/platform_impl/linux/x11/window.rs
```diff
@@ -347,26 +346,24 @@ impl UnownedWindow {
...
                     | ffi::XI_LeaveMask
                     | ffi::XI_FocusInMask
-                    | ffi::XI_FocusOutMask;
-                if window_attrs.multitouch {
-                    mask |= ffi::XI_TouchBeginMask
-                        | ffi::XI_TouchUpdateMask
-                        | ffi::XI_TouchEndMask;
-                }
+                    | ffi::XI_FocusOutMask
+                    | ffi::XI_TouchBeginMask
```
ZeGentzy: Uhhh, this intended? Was the if statement unnecessary? I thought this PR only renamed stuff. Is this mentioned in the changelog?

src/platform_impl/macos/window.rs
```diff
@@ -497,44 +496,43 @@ impl UnownedWindow {
...
         let point = appkit::CGPoint {
             x: (cursor_position.x + window_position.x) as CGFloat,
             y: (cursor_position.y + window_position.y) as CGFloat,
         };
         CGDisplay::warp_mouse_cursor_position(point)
-            .map_err(|e| format!("`CGWarpMouseCursorPosition` failed: {:?}", e))?;
+            .map_err(|e| ExternalError::Os(os_error!(OsError::CGError(e))))?;
         CGDisplay::associate_mouse_and_mouse_cursor_position(true)
-            .map_err(|e| format!("`CGAssociateMouseAndMouseCursorPosition` failed: {:?}", e))?;
+            .map_err(|e| ExternalError::Os(os_error!(OsError::CGError(e))))?;
```
ZeGentzy: I sorta like having fn names in the errors. Helps when users send logs.

src/platform_impl/windows/window.rs
```diff
@@ -325,56 +314,56 @@ impl Window {
...
         self.window_state.lock().dpi_factor
     }
 
-    fn set_cursor_position_physical(&self, x: i32, y: i32) -> Result<(), String> {
+    fn set_cursor_position_physical(&self, x: i32, y: i32) -> Result<(), ExternalError> {
         let mut point = POINT { x, y };
         unsafe {
             if winuser::ClientToScreen(self.window.0, &mut point) == 0 {
-                return Err("`ClientToScreen` failed".to_owned());
+                return Err(ExternalError::Os(os_error!(io::Error::last_os_error())));
```
ZeGentzy: I like when we got function names in the errors. Helps with log readability.

src/platform_impl/windows/window.rs
```diff
@@ -587,22 +576,18 @@ unsafe fn init<T: 'static>(
         let icon = attributes.window_icon
             .take()
             .map(WinIcon::from_icon);
-        if icon.is_some() {
-            Some(icon.unwrap().map_err(|err| {
-                CreationError::OsError(format!("Failed to create `ICON_SMALL`: {:?}", err))
-            })?)
+        if let Some(icon) = icon {
+            Some(icon.map_err(|e| os_error!(e))?)
```
ZeGentzy: I like when we got function names in the errors. Helps with log readability.

src/platform_impl/windows/window.rs
```diff
@@ -670,8 +655,7 @@ unsafe fn init<T: 'static>(
         );
 
         if handle.is_null() {
-            return Err(CreationError::OsError(format!("CreateWindowEx function failed: {}",
-                                              format!("{}", io::Error::last_os_error()))));
+            return Err(os_error!(io::Error::last_os_error()));
```
ZeGentzy: I like when we got function names in the errors. Helps with log readability.

src/error.rs
```diff
@@ -0,0 +1,43 @@
+use platform_impl;
+
+pub enum ExternalError {
+    NotSupported(NotSupportedError),
+    Os(OsError),
+}
+
+/// TODO: MANUALLY IMPLEMENT TRAITS
```
ZeGentzy: But why?

src/platform_impl/linux/x11/window.rs
```diff
@@ -1161,10 +1158,10 @@ impl UnownedWindow {
                 ffi::GrabNotViewable => Err("Cursor could not be grabbed: grab location not viewable"),
                 ffi::GrabFrozen => Err("Cursor could not be grabbed: frozen by another client"),
                 _ => unreachable!(),
-            }.map_err(|err| err.to_owned())
+            }.map_err(|err| ExternalError::Os(os_error!(OsError::XMisc(err))))
         } else {
             self.xconn.flush_requests()
-                .map_err(|err| format!("Failed to call `XUngrabPointer`: {:?}", err))
+                .map_err(|err| ExternalError::Os(os_error!(OsError::XError(err))))
```
ZeGentzy: I like when we get function names in the errors. Helps with log readability.

src/platform_impl/linux/x11/window.rs
```diff
@@ -1204,13 +1201,13 @@ impl UnownedWindow {
                 x,
                 y,
             );
-            self.xconn.flush_requests().map_err(|e| format!("`XWarpPointer` failed: {:?}", e))
+            self.xconn.flush_requests().map_err(|e| ExternalError::Os(os_error!(OsError::XError(e))))
```
ZeGentzy: I like when we get function names in the errors. Helps with log readability.

src/platform_impl/windows/window.rs
```diff
@@ -587,22 +576,18 @@ unsafe fn init<T: 'static>(
...
-        let icon = pl_attribs.taskbar_icon
+        let icon = attributes.window_icon
             .take()
             .map(WinIcon::from_icon);
-        if icon.is_some() {
-            Some(icon.unwrap().map_err(|err| {
-                CreationError::OsError(format!("Failed to create `ICON_BIG`: {:?}", err))
-            })?)
+        if let Some(icon) = icon {
+            Some(icon.map_err(|e| os_error!(e))?)
```
ZeGentzy: I like when we get function names in the errors. Helps with log readability.

src/error.rs
```diff
@@ -0,0 +1,43 @@
+use platform_impl;
+
```
aloucks: Can you implement `std::error::Error` for this so that conversion with `?` will work?

src/error.rs
Osspial: That was mainly referring to `Debug` impls, since otherwise the `_marker` field will show up in the auto-generated impl.

src/lib.rs
Osspial: `#[macro_use]` doesn't publicly export macros - that's `#[macro_export]`. That just lets the macro be used elsewhere in the codebase.

src/error.rs
Osspial: So that `NotSupportedError` is an opaque type. That has two purposes:
1. Users can't create instances of it
2. We can freely add fields to it in the future without making breaking changes.

src/platform_impl/linux/x11/window.rs
Osspial: Oh right, I... also removed the `multitouch` field, since it's only used by a few backends and I don't think it makes sense to ever leave it disabled. We can move that discussion to another PR if you'd like - I just have a bad habit of having several small, somewhat unrelated changes katamari into a single larger commit.

ZeGentzy: If it gets mentioned in the change log, I'm fine with it staying in this PR.

src/platform/ios.rs
Osspial: That's a good question! Looks like stuff got messed up on the iOS backend when I tried to rebase it - I'll just go ahead and re-apply them from scratch. That should be less work than figuring out how to fix this.

src/platform_impl/macos/window.rs
Osspial: I can totally get that. I've set up the `os_error` macro to include the filename and line number, so we know where the error originated from (that's why it's a macro, actually!). Do you think that's an adequate replacement? If not, I'd be down with replacing or supplementing those fields with a `description` field that can include the source function name.

src/error.rs
Osspial: That's the plan. I didn't do it for the initial PR since I was on a machine where that wasn't easy to do (platform-specific `Display` implementations are necessary), but those should definitely get in before this gets merged.

src/platform_impl/macos/window.rs
ZeGentzy: Ah, I did not notice that `os_error` is a macro. If that's the case, I think it is fine as it is, however, a description field could never hurt.

Osspial:
This PR should be ready to merge. I'll give it a day or so to get any final feedback, though.

felixrabe:
Good stuff!

Btw there is a preference to switch off squashing (or enable other merge strategies) if you prefer: https://help.github.com/en/articles/configuring-commit-squashing-for-pull-requests

PR #887, opened 2019-05-26
Fix Windows backend invoking unreachable! with Exit and run_return

### LINKED ISSUE #867: EventLoop 2.0 hitting unreachable code during resize on Windows
When resizing a window on Windows, the following `unreachable!()` statement is hit:
https://github.com/rust-windowing/winit/blob/eventloop-2.0/src/platform_impl/windows/event_loop.rs#L412

This causes the program to immediately panic without any chance for recovery.

This was tested using the `run_return` event loop method. It was suggested to me that it would freeze and act like `run` during resize on Windows, but apparently there are some more severe issues?

Osspial:
Fixes #867.

@chrisduerr Can you confirm that this fixes the issue?


chrisduerr:
Can confirm that this fixes the crash!

PR #888, opened 2019-05-29
Build window with requested dimensions

zakorgy:

This caused me a headache in the past few weeks, because on Windows sometimes the window size was `1024 x 768` instead of what I requested. This fixes that issue. I didn't dig myself in the details, if anyone want's to reproduce it on Windows here are the steps:
1. clone https://github.com/szeged/webrender
2. `cd webrender/wrench`
3.  remove the `wrapper.resize()` from [here](https://github.com/szeged/webrender/blob/master/wrench/src/main.rs#L434)
4. Run any of the tests e.g. `cargo run --features=dx12 show reftests\aa\aa-dist-bug.yaml`
5. The window size should be `1920x1080` but it's `1024x768`

cc @Osspial

### DIFF
--- a/src/window.rs
+++ b/src/window.rs
@@ -148,8 +148,8 @@ impl WindowBuilder {
                 // resizing the window to the dimensions of the monitor when fullscreen
                 LogicalSize::from_physical(monitor.get_dimensions(), 1.0)
             } else {
-                // default dimensions
-                (1024, 768).into()
+                // use requested or fallback to default dimensions
+                self.window.dimensions.unwrap_or((1024, 768).into())
             }
         }));
 


src/window.rs
```diff
@@ -148,8 +148,8 @@ impl WindowBuilder {
                 // resizing the window to the dimensions of the monitor when fullscreen
                 LogicalSize::from_physical(monitor.get_dimensions(), 1.0)
             } else {
-                // default dimensions
-                (1024, 768).into()
+                // use requested or fallback to default dimensions
+                self.window.dimensions.unwrap_or((1024, 768).into())
```
vberger: I don't understand, this code is already in a `self.window.dimensions.unwrap_or_else(|| { ... })`, how can this change make any difference?

zakorgy: Ooops! I didn't notice that this is wrapped in a `unwrap_or_else`. Even though this still fixes my issue. I will try to figure out what is happening here.

zakorgy:
Switching back to stable rust from nightly (1.36.1) fixed the issue on my side. Sorry for the false alarm, I will close this.

PR #895, opened 2019-06-03
WIP - Make EL2 DPI changes and implement on Windows

Osspial:
- [ ] Tested on all platforms changed
  - [x] Windows
  - [ ] macOS
  - [ ] iOS
  - [ ] Linux X11
  - [ ] Linux Wayland
- [x] Change `HiDPIFactorChanged` event.
- [x] Remove `LogicalSize` and `LogicalPosition` from `WindowEvent`s.
- [ ] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users
- [x] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior
- [x] Update examples
- [ ] Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented

Makes changes to the public API as discussed in #837, and implements them on Windows.

I was originally going to try and implement these changes on the other platforms, but given the number of DPI problems we've had in the past I figure it's unwise to attempt to hack on systems I don't understand and am unable to test.

PR #896, opened 2019-06-03
Support listing available video modes for a monitor

### LINKED ISSUE #877: Support listing available video modes for a monitor
https://github.com/rust-windowing/glutin/issues/742

Should be trivial to implement.

aleksijuvani:
Resolves #877.

     - [x]  Windows
     - [x]  macOS
     - [x]  X11
     - [x]  Wayland
     - [x]  iOS
     - [ ]  Android
     - [ ]  Emscripten

Adds a new method `get_video_modes` to the `MonitorId` struct in the public API, and implements it on the Windows platform. I'm going to look into implementing this on macOS as well..

### DIFF
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -156,6 +156,7 @@ Legend:
 |Window maximization toggle       |✔️     |✔️     |✔️         |✔️             |**N/A**|**N/A**|**N/A**   |
 |Fullscreen                       |✔️     |✔️     |✔️         |✔️             |**N/A**|**N/A**|❌        |
 |Fullscreen toggle                |✔️     |✔️     |✔️         |✔️             |**N/A**|**N/A**|❌        |
+|Video mode query                 |✔️     |✔️     |✔️         |✔️             |❌    |✔️     |❌        |
 |HiDPI support                    |✔️     |✔️     |✔️         |✔️             |▢[#721]|✔️    |✔️         |
 |Popup windows                    |❌     |❌     |❌         |❌             |❌    |❌     |❌        |
 
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -92,6 +92,7 @@ If your PR makes notable changes to Winit's features, please update this section
 
 ### System Information
 - **Monitor list**: Retrieve the list of monitors and their metadata, including which one is primary.
+- **Video mode query**: Monitors can be queried for their supported fullscreen video modes (consisting of resolution, refresh rate, and bit depth).
 
 ### Input Handling
 - **Mouse events**: Generating mouse events associated with pointer motion, click, and scrolling events.
@@ -156,14 +157,14 @@ Legend:
 |Window maximization toggle       |✔️     |✔️     |✔️         |✔️             |**N/A**|**N/A**|**N/A**   |
 |Fullscreen                       |✔️     |✔️     |✔️         |✔️             |**N/A**|**N/A**|❌        |
 |Fullscreen toggle                |✔️     |✔️     |✔️         |✔️             |**N/A**|**N/A**|❌        |
-|Video mode query                 |✔️     |✔️     |✔️         |✔️             |❌    |✔️     |❌        |
 |HiDPI support                    |✔️     |✔️     |✔️         |✔️             |▢[#721]|✔️    |✔️         |
 |Popup windows                    |❌     |❌     |❌         |❌             |❌    |❌     |❌        |
 
 ### System information
-|Feature      |Windows|MacOS |Linux x11|Linux Wayland|Android|iOS    |Emscripten|
-|------------ | ----- | ---- | ------- | ----------- | ----- | ----- | -------- |
-|Monitor list |✔️    |✔️    |✔️       |✔️          |**N/A**|**N/A**|**N/A**   |
+|Feature          |Windows|MacOS |Linux x11|Linux Wayland|Android|iOS    |Emscripten|
+|---------------- | ----- | ---- | ------- | ----------- | ----- | ----- | -------- |
+|Monitor list     |✔️    |✔️    |✔️       |✔️          |**N/A**|**N/A**|**N/A**   |
+|Video mode query |✔️    |✔️    |✔️       |✔️          |❌      |✔️     |❌         |
 
 ### Input handling
 |Feature                 |Windows   |MacOS   |Linux x11|Linux Wayland|Android|iOS    |Emscripten|
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -84,6 +84,8 @@ extern crate log;
 #[cfg(feature = "serde")]
 #[macro_use]
 extern crate serde;
+#[macro_use]
+extern crate derivative;
 
 #[cfg(target_os = "windows")]
 extern crate winapi;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -101,6 +101,8 @@ extern crate dispatch;
 extern crate core_foundation;
 #[cfg(target_os = "macos")]
 extern crate core_graphics;
+#[cfg(target_os = "macos")]
+extern crate core_video_sys;
 #[cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "netbsd", target_os = "openbsd"))]
 extern crate x11_dl;
 #[cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "netbsd", target_os = "openbsd", target_os = "windows"))]
--- a/src/monitor.rs
+++ b/src/monitor.rs
@@ -44,6 +44,27 @@ impl Iterator for AvailableMonitorsIter {
     }
 }
 
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct VideoMode {
+    pub(crate) dimensions: (u32, u32),
+    pub(crate) bit_depth: u16,
+    pub(crate) refresh_rate: u16,
+}
+
+impl VideoMode {
+    pub fn dimensions(&self) -> PhysicalSize {
+        self.dimensions.into()
+    }
+
+    pub fn bit_depth(&self) -> u16 {
+        self.bit_depth
+    }
+
+    pub fn refresh_rate(&self) -> u16 {
+        self.refresh_rate
+    }
+}
+
 /// Handle to a monitor.
 ///
 /// Allows you to retrieve information about a given monitor and can be used in [`Window`] creation.
@@ -87,5 +108,11 @@ impl MonitorHandle {
     #[inline]
     pub fn hidpi_factor(&self) -> f64 {
         self.inner.hidpi_factor()
+    }
+
+    /// Returns all fullscreen video modes supported by this monitor.
+    #[inline]
+    pub fn video_modes(&self) -> impl Iterator<Item = VideoMode> {
+        self.inner.video_modes()
     }
 }
--- a/src/monitor.rs
+++ b/src/monitor.rs
@@ -44,6 +44,12 @@ impl Iterator for AvailableMonitorsIter {
     }
 }
 
+/// Describes a fullscreen video mode of a monitor.
+///
+/// Can be acquired with:
+/// - [`MonitorHandle::video_modes`][monitor_get].
+///
+/// [monitor_get]: ../monitor/struct.MonitorHandle.html#method.video_modes
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
 pub struct VideoMode {
     pub(crate) dimensions: (u32, u32),
@@ -52,14 +58,26 @@ pub struct VideoMode {
 }
 
 impl VideoMode {
+    /// Returns the resolution of this video mode.
     pub fn dimensions(&self) -> PhysicalSize {
         self.dimensions.into()
     }
 
+    /// Returns the bit depth of this video mode, as in how many bits you have
+    /// available per color. This is generally 24 bits or 32 bits on modern
+    /// systems, depending on whether the alpha channel is counted or not.
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **Wayland:** Always returns 32.
+    /// - **iOS:** Always returns 32.
     pub fn bit_depth(&self) -> u16 {
         self.bit_depth
     }
 
+    /// Returns the refresh rate of this video mode. **Note**: the returned
+    /// refresh rate is an integer approximation, and you shouldn't rely on this
+    /// value to be exact.
     pub fn refresh_rate(&self) -> u16 {
         self.refresh_rate
     }
--- a/src/platform_impl/ios/monitor.rs
+++ b/src/platform_impl/ios/monitor.rs
@@ -1,18 +1,13 @@
 use std::{
-    collections::VecDeque,
+    collections::{HashSet, VecDeque},
     fmt,
     ops::{Deref, DerefMut},
 };
 
 use dpi::{PhysicalPosition, PhysicalSize};
+use monitor::VideoMode;
 
-use platform_impl::platform::ffi::{
-    id,
-    nil,
-    CGFloat,
-    CGRect,
-    NSUInteger,
-};
+use platform_impl::platform::ffi::{id, nil, CGFloat, CGRect, CGSize, NSInteger, NSUInteger};
 
 pub struct Inner {
     uiscreen: id,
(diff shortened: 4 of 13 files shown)


Osspial:
Hey, thanks for the PR!

FYI before you get much further into implementing this on macOS - the vast majority of work is currently being done on the `eventloop-2.0` branch, and we're expecting to merge that onto master Soon. I'd suggest reworking this PR against that branch. Fortunately, looking at these changes I wouldn't expect that to be too difficult to do.

aleksijuvani:
Thanks for the heads up. I'll take a look.

src/lib.rs
```diff
@@ -540,3 +542,42 @@ impl Default for WindowAttributes {
         }
     }
 }
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct VideoMode {
+    dimensions: (u32, u32),
+    bit_depth: u16,
+    frequency: u16,
```
ZeGentzy: you mean refresh rate? Frequency is a pretty vague name.

src/lib.rs
```diff
@@ -540,3 +542,42 @@ impl Default for WindowAttributes {
         }
     }
 }
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct VideoMode {
+    dimensions: (u32, u32),
```
ZeGentzy: Can this become `PhysicalSize` instead of `(u32, u32)`

src/lib.rs
```diff
@@ -540,3 +542,42 @@ impl Default for WindowAttributes {
...
+    bit_depth: u16,
```
ZeGentzy: Is it possible to expose the number of bits for each channel separately? At least distinguish the number of color bits from alpha bits?

src/lib.rs
aleksijuvani: This needs to implement `Hash` which `PhysicalSize` does not (and can not, as it's essentially a tuple of `f64`). Hence, we're doing the conversion from `(u32, u32)` to `PhysicalSize` in the getter method.

src/lib.rs
aleksijuvani: I'm open to bikeshedding this name, but I've commonly seen frequency used as a synonym for the display refresh rate. It is measured in hertz after all, which indicate a frequency..

src/lib.rs
Osspial: `PhysicalSize` is stored as `u32`s and not `f64`s on `eventloop-2.0`, so this should be possible once rebased.

src/lib.rs
aleksijuvani: WinAPI does not expose this information as far as I can tell. SDL has some heuristics for this, but I think those were written with 16-bit and 8-bit modes in mind for older Windows systems. There's still some code for greyscale support in there.. Windows 8 and up no longer supports these modes. What do you need this information for?

src/lib.rs
aleksijuvani: Looks like this would depend on the changes made in #895 being merged first.

aleksijuvani:
I've reworked this against the `eventloop-2.0` branch and added support for macOS. There's a bit of a quirk there, as we can't currently obtain the refresh rate of the monitor without some patches upstream against the core-video-sys crate for example.

I'm also not a fan of allocating an additional `HashSet` on macOS instead of just re-using the iterator we get from calling `map` on the `Vec` returned from `CGDisplayMode::all_display_modes`, but I couldn't come up with a clean way to do this.

How do we proceed with implementing this for the rest of the platforms? As for the Android and iOS platforms I think this comes down to just returning the screen size, and additionally obtaining the refresh rate and bit depth if applicable. I haven't looked into what Emscripten is doing.

src/lib.rs
ZeGentzy: Yeah, but frequency could be frequency of anything. Sound is also measured in hertz. I'd be okay with something like `refresh_frequency` or something.

src/lib.rs
ZeGentzy: No, reason, I just noticed glfw exposed the separate bits. That's OK.

src/monitor.rs
```diff
@@ -44,6 +47,51 @@ impl Iterator for AvailableMonitorsIter {
...
+    }
+}
+
+/// An iterator over all available video modes.
+///
+/// Can be acquired with:
+/// - [`MonitorHandle::video_modes`][monitor_get].
+///
+/// [monitor_get]: ../monitor/struct.MonitorHandle.html#method.video_modes
+pub struct VideoModesIter<'a> {
```
ZeGentzy: `type VideoModesIter<'a> = HashSetIter<'a, VideoMode>;`? No need to define our own struct and everything, I think.

ZeGentzy:
> I've reworked this against the `eventloop-2.0` branch and added support for macOS. There's a bit of a quirk there, as we can't currently obtain the refresh rate of the monitor without some patches upstream against the core-video-sys crate for example.

Filled upstream already, yeah?

> I'm also not a fan of allocating an additional `HashSet` on macOS instead of just re-using the iterator we get from calling `map` on the `Vec` returned from `CGDisplayMode::all_display_modes`, but I couldn't come up with a clean way to do this.

I guess you could use a vec, then sort it at the end (or always keep it sorted) and dedup it, however that sounds like a pain to maintain.

src/lib.rs
aleksijuvani: I see. I don't think I've seen it called `refresh_frequency` before. `refresh_rate` is pretty self-explainatory and would work for me.

src/monitor.rs
aleksijuvani: I think we ought to newtype it to avoid a breaking change if this type changes in the future, as the underlying collection, and hence the iterator type is an implementation detail. That is essentially what we're doing here.

aleksijuvani:
> > I've reworked this against the `eventloop-2.0` branch and added support for macOS. There's a bit of a quirk there, as we can't currently obtain the refresh rate of the monitor without some patches upstream against the core-video-sys crate for example.
> 
> Filled upstream already, yeah?

I'll look into filing a pull request against core-video-sys for this today.

src/monitor.rs
ZeGentzy: Would an type alias cause breaking changes if the details get changed? I wouldn't think so, but I haven't put much thought into the matter.

aleksijuvani: I think it could, because it essentially acts the same as if the alias and the aliased type were equivalent.

As an example, let's say there's a function in user code that takes a `std::collections::hash_set::Iter` parameter (for some reason) instead of a generic `Iterator` parameter. If the alias in winit were suddenly to point to a `std::collections::vec_deque::Iter` instead, this would become a type check error. This is probably not how you would usually write a function that takes an iterator parameter, but it illustrates the problem.

I think different iterators do also expose different methods on them, so if some methods were suddenly to disappear, or new ones to appear entirely, this would be more visible of a breaking change.

aleksijuvani:
We're now reporting the display refresh rate correctly on macOS. I've filed LuoZijun/rust-core-video-sys#2 to expose the necessary functionality from the Core Video API.

aleksijuvani:
LuoZijun/rust-core-video-sys#2 has been merged, so all that remains is implementing this for the remaining platforms. I suppose I could take a look at either iOS or Linux today, so that we can ship this as soon as possible.

aleksijuvani:
I've finished the iOS implementation. I couldn't get the Emscripten or Android platforms to compile. It looks like their implementation in eventloop-2.0 hasn't been finished.

aleksijuvani:
I've finished implementing this on Wayland, and found a cleaner way to return iterators from the `video_modes` method. I've refactored all platforms accordingly. I'm going to look into implementing this on X11 next.

aleksijuvani:
X11 implementation is done. This is now implemented on all platforms except for Android and Emscripten that don't seem to be ported over to eventloop-2.0 yet. Should be ready for final review!

src/platform_impl/windows/monitor.rs
```diff
@@ -1,21 +1,23 @@
...
-use super::{EventLoop, util};
+use super::{util, EventLoop};
 use dpi::{PhysicalPosition, PhysicalSize};
+use monitor::VideoMode;
 use platform_impl::platform::dpi::{dpi_to_scale_factor, get_monitor_dpi};
 use platform_impl::platform::window::Window;
 
 /// Win32 implementation of the main `MonitorHandle` object.
-#[derive(Debug, Clone)]
+#[derive(Clone)]
```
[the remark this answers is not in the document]
aleksijuvani: I would've normally used derivative here, but it's not used elsewhere in winit, so I didn't want to introduce an additional dependency.

src/platform_impl/ios/monitor.rs
```diff
@@ -134,6 +129,27 @@ impl Inner {
...
+        let available_mode_count: NSUInteger = unsafe { msg_send![available_modes, count] };
+
+        let mut modes = HashSet::with_capacity(available_mode_count);
+
+        for i in 0..available_mode_count {
+            let mode: id = unsafe { msg_send![available_modes, objectAtIndex: i] };
+            let size: CGSize = unsafe { msg_send![mode, size] };
+            modes.insert(VideoMode {
+                dimensions: (size.width as u32, size.height as u32),
+                bit_depth: 32,
```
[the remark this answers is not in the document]
aleksijuvani: There's no way to query the bit depth on iOS as far as I can tell. I believe the bit depth is always 32 bits.

src/platform_impl/windows/monitor.rs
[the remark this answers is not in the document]
ZeGentzy: While @Osspial might have a different opinion, I think derivative is the better option for maintainability here.

src/platform_impl/linux/wayland/event_loop.rs
```diff
@@ -584,6 +585,20 @@ impl MonitorHandle {
...
+    #[inline]
+    pub fn video_modes(&self) -> impl Iterator<Item = VideoMode>
+    {
+        self.mgr
+            .with_info(&self.proxy, |_, info| info.modes.clone())
+            .unwrap_or(vec![])
+            .into_iter()
+            .map(|x| VideoMode {
+                dimensions: (x.dimensions.0 as u32, x.dimensions.1 as u32),
+                refresh_rate: (x.refresh_rate as f32 / 1000.0 + 0.5) as u16,
```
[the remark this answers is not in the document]
aleksijuvani: We're receiving the refresh rate from Wayland in millihertz, so we must divide it by a thousand to get it in hertz. We then effectively `ceil` the float before casting it into an integer, but I suppose we could've used `f32::ceil` here also.

src/platform_impl/linux/x11/util/randr.rs
```diff
@@ -114,6 +115,33 @@ impl XConnection {
...
+        let resource_modes = slice::from_raw_parts((*resources).modes, (*resources).nmode as usize);
+
+        let modes = resource_modes
+            .iter()
+            // XRROutputInfo contains an array of mode ids that correspond to
+            // modes in the array in XRRScreenResources
+            .filter(|x| output_modes.iter().any(|id| x.id == *id))
+            .map(|x| {
+                let refresh_rate = if x.dotClock > 0 && x.hTotal > 0 && x.vTotal > 0 {
+                    x.dotClock as u64 * 1000 / (x.hTotal as u64 * x.vTotal as u64)
```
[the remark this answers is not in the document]
aleksijuvani: I don't know what's going on here, but this is the same line of code that SDL, GTK and GLFW use to calculate the display refresh rate. I couldn't find any documentation for this.

src/platform_impl/linux/x11/util/randr.rs
```diff
@@ -114,6 +115,33 @@ impl XConnection {
...
+                } else {
+                    0
+                };
+
+                VideoMode {
+                    dimensions: (x.width, x.height),
+                    refresh_rate: (refresh_rate as f32 / 1000.0 + 0.5) as u16,
```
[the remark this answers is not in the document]
aleksijuvani: Same thing as in Wayland. We're converting from millihertz to hertz and ceiling the value.

src/platform_impl/macos/monitor.rs
```diff
@@ -93,6 +102,44 @@ impl MonitorHandle {
...
+            assert!(time.flags & kCVTimeIsIndefinite == 0);
+
+            time.timeScale as i64 / time.timeValue
+        };
+
+        CGDisplayMode::all_display_modes(self.0, std::ptr::null())
+            .expect("failed to obtain list of display modes")
+            .into_iter()
+            .map(move |mode| {
+                let cg_refresh_rate = (mode.refresh_rate() + 0.5) as i64;
```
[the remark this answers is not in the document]
aleksijuvani: Yes, that should work. I'll see if I can change these to use `ceil` instead.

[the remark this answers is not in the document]
aleksijuvani: Actually, this isn't a `ceil`, but `round`. Same goes for the other occurrences. I've changed the code to use `round`.

aloucks:
It makes it difficult to see what's going on when you force-push after comments have started. The commits will all get squashed in the end anyway.

aleksijuvani:
Sorry about that. I didn't realize this repository squashes pull requests upon merge. While making changes to the Wayland platform, I noticed that I needed to restructure the other platforms as well, and I decided to squash it all into a single commit at that point. There shouldn't be any more large changes landing to this branch at this point.

src/platform_impl/linux/wayland/event_loop.rs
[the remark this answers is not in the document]
vberger: In all, I'm not sure it is the right approach to represent refresh rates as `u16` in Hertz. There are displays whose refresh rate is not an integer in this unit, no?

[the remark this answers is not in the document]
aleksijuvani: The refresh rate is usually not an exact integer, but very close (e.g. 59.996 Hz). We could report the display refresh rate as a float, but Windows and iOS only report it as an integer, so it might be misleading to return a float on these platforms. I assume this is what you meant?

src/platform_impl/windows/monitor.rs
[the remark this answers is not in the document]
Osspial: I'm fine with using derivative here.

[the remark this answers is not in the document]
aleksijuvani: I've changed this to use derivative.

src/platform_impl/linux/wayland/event_loop.rs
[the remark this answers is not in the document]
vberger: I don't get what would be misleading about using a float while some platform always return integer values, could you elaborate on that?

We already have precedent with the DPI factor values, which are always integer on Wayland for example.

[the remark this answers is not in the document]
aleksijuvani: To elaborate, I feel like that returning a float here implies that this is an exact refresh rate, when in fact it's truncated or rounded (on at least two platforms). Returning an integer type does not have this implication to me.

If we look at other windowing libraries, there's precedent of using integers to represent the refresh rate in SDL and GLFW. I don't feel particularly strongly about this either way, so I'm fine with changing this to return floats if we all agree on that.

FEATURES.md
```diff
@@ -156,6 +156,7 @@ Legend:
 |Window maximization toggle       |✔️     |✔️     |✔️         |✔️             |**N/A**|**N/A**|**N/A**   |
 |Fullscreen                       |✔️     |✔️     |✔️         |✔️             |**N/A**|**N/A**|❌        |
 |Fullscreen toggle                |✔️     |✔️     |✔️         |✔️             |**N/A**|**N/A**|❌        |
+|Video mode query                 |✔️     |✔️     |✔️         |✔️             |❌    |✔️     |❌        |
```
Osspial: Could you add a brief blurb describing this feature further up in the document, for the sake of consistency with the rest of the features?

src/platform_impl/windows/monitor.rs
```diff
@@ -170,4 +175,44 @@ impl MonitorHandle {
...
+        loop {
+            unsafe {
+                let mut mode: wingdi::DEVMODEW = mem::zeroed();
+                mode.dmSize = mem::size_of_val(&mode) as WORD;
+                if winuser::EnumDisplaySettingsExW(
+                    self.monitor_info.szDevice.as_ptr(),
+                    i,
+                    &mut mode,
+                    0,
+                ) == 0
```
Osspial: Multi-line `if` clauses harm code readability. Could you pull this into a separate variable?

FEATURES.md
aleksijuvani: Done! I also noticed that this should probably be under the system information section rather than under windowing, so I've moved it there.

src/platform_impl/windows/monitor.rs
aleksijuvani: I've moved `device_name` into a separate variable so this fits into a single line.

src/platform_impl/linux/wayland/event_loop.rs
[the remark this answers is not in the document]
vberger: I personally have the other expectation, imo an integer refresh rate suggests that it is exact...

Anyway, I don't know much about which uses an app would have for the refresh rate. If rounding it doe snot pose any functionality issue, then I'm fine with it.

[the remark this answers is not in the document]
aleksijuvani: Having slept on it, I think we should leave it as a rounded integer, so at least the behaviour is consistent across platforms. If the user requires an exact value, there's not much point in us returning an exact value on some platforms, and a rounded value on others. Furthermore, changing this to a floating point value would complicate the implementation (though slightly) for `VideoMode` as we would now have to implement `Hash` and `Eq` by hand. If there later appears an use-case where an exact value is required, we can revisit this. For now, perhaps we can document it in the function description that the returned value is an approximation?

felixrabe:
I just noticed on my old macOS 10.9 MBAir that the `core-video-sys` crate brings in `metal` as a dependency, which means I can't compile at all.

Is there some way to do this without depending on Metal?

aleksijuvani:
We don't actually need Metal for this at all, but core-video-sys happens to pull that in for other reasons. I've opened LuoZijun/rust-core-video-sys#3 to make the Metal dependency optional. I will follow this up with a pull request to winit that then disables the "metal" feature for the core-video-sys dependency.

PR #897, opened 2019-06-03
Direct new contributions to eventloop-2.0

Osspial:
This is a temporary change while `eventloop-2.0` is separate from `master`.


aloucks:
It might be more effective to put this at the top of the README.

felixrabe:
or both

ZeGentzy:
Can I just say that I love the caps.

PR #898, opened 2019-06-03
Fix Windows fullscreen toggle on windows that started invisible

repi:
If one created a window with `with_visibility(false)` and then done `show()` then `set_fullscreen` wasn't working properly and hid the window because the internal `WindowFlags` was still stored as `WindowFlags::VISIBLE = false`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- On Windows, fix `set_fullscreen` on windows that were created invisible (`with_visibility(false)`)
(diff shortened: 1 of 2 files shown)


Osspial:
Thanks for the PR! Could you rebase this onto the `eventloop-2.0` branch?

repi:
Thanks, we haven't upgraded and tested 0.20 yet but great if it has been fixed there with/after the big refactor. Can confirm later when we upgrade

Osspial:
It looks like this particular issue has been addressed, so I'm going to close this PR. However, it looks like the test-case you describe is still showing some funky behavior which I'll look into.

repi:
Thanks!

PR #906, opened 2019-06-10
Make `log` an optional default dependency

nvzqz:
Not all users of `winit` intend on logging events via `log`. This allows for opting out of the dependency.

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@
 - Removed `serde` implementations from `ControlFlow`.
 - Rename several functions to improve both internal consistency and compliance with Rust API guidelines.
 - Remove `WindowBuilder::multitouch` field, since it was only implemented on a few platforms. Multitouch is always enabled now.
+- **Breaking:** Make `log` an optional default dependency
 
 # Version 0.19.1 (2019-04-08)
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ features = ["serde"]
 [dependencies]
 lazy_static = "1"
 libc = "0.2"
-log = "0.4"
+log = { version = "0.4", optional = true }
 serde = { version = "1", optional = true, features = ["serde_derive"] }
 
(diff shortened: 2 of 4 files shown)


nvzqz:
@ZeGentzy rebase to which branch? This is already based off of `eventloop-2.0`.

ZeGentzy:
/me has a brain fart

I assumed they were all based off master.

lgtm. Any objections, @Osspial ?

vberger:
I'm not strongly opposed to this, but I'm not sure I understand the need. Is `log` such a large dependency that it changes that much?

nvzqz:
It's not really a matter of `log` being a large dependency. It's more so about reducing the total number of dependencies for the end user if they don't need this feature. Also, some people want to not waste CPU cycles, so the logging here becomes a no-op at call sites when the `log`  feature is disabled. The substitute macros make it such that calling upon `log`'s macros remains the same within `winit` without littering the codebase with `#[cfg(feature = "log")]`.

Osspial:
A couple things: 

a) `log` is a fairly ubiquitous crate (as of now, there are 2873 direct dependencies on `crates.io`, making it one of the most depended-on crates on the platform!). Pretty much any serious project will likely depend on other crates that have it as a dependency, and removing it entirely from your dependency tree seems like a game of dependency whack-a-mole that'll be pretty hard to win. 

b) Most crates have us as a dependency through Glutin, so they couldn't even disable this feature if they wanted to.

As such, I don't think this is worth the additional complexity on our end.

-----

Also, if you want to compile `log` out of your crate completely, [the crate already has a standard API for it](https://docs.rs/log/0.4.6/log/#compile-time-filters).

nvzqz:
I wasn't aware of `log`'s ability to be compiled out downstream.

Also, I realize that having an optional `log` feature results in dependent crates not being able to conditionally enable `log` and use it themselves since optional dependencies can't enable features in other dependencies.

Closing because you apparently can opt out of logging.

PR #907, opened 2019-06-10
Drop platform name from extension traits

nvzqz:
As mentioned in https://github.com/rust-windowing/winit/issues/459#issuecomment-500286307, I feel as though adding the name of the platform to the `Ext` traits feels awkward and inconsistent with the Rust ecosystem. Also, these traits are already within their own modules specific to each platform.

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@
       - `Exit` aborts the event loop.
     - Takes a closure that implements `'static + FnMut(Event<T>, &EventLoop<T>, &mut ControlFlow)`.
       - `&EventLoop<T>` is provided to allow new `Window`s to be created.
-  - **Major:** `platform::desktop` module exposes `EventLoopExtDesktop` trait with `run_return` method.
+  - **Major:** `platform::desktop` module exposes `EventLoopExt` trait with `run_return` method.
     - Behaves identically to `run`, but returns control flow to the calling context and can take non-`'static` closures.
   - `EventLoop`'s `poll_events` and `run_forever` methods have been removed in favor of `run` and `run_return`.
 - Changes to events:
--- a/examples/fullscreen.rs
+++ b/examples/fullscreen.rs
@@ -70,8 +70,8 @@ fn main() {
                         #[cfg(target_os = "macos")]
                         {
                             if macos_use_simple_fullscreen {
-                                use winit::platform::macos::WindowExtMacOS;
-                                if WindowExtMacOS::set_simple_fullscreen(&window, !is_fullscreen) {
+                                use winit::platform::macos::WindowExt;
+                                if WindowExt::set_simple_fullscreen(&window, !is_fullscreen) {
                                     is_fullscreen = !is_fullscreen;
                                 }
                                 return;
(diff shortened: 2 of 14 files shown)


vberger:
I believe the reason these extension traits have the platform name in their name is because they are not exclusive and more than one may be imported at the same time.

For example someone may need to have both `EventLoopExtDesktop` and `EventLoopExtUnix` in scope, which is not possible if they have the same name (in which case this forces the user to rename the import with something like `use ...::EventLoopExt as EventLoopExtDesktop`.

aleksijuvani:
I believe it's possible to use syntax such as `use winit::platform::{desktop::EventLoopExt as _, macos::EventLoopExt as _}` to bring multiple traits with the same name into scope as of Rust 1.33.

nvzqz:
If you don't need to refer to the traits by name, you can just do a glob import ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=8139a5a81dafccddb0cf05d13f2c508f)).

```rust
mod a {
    pub trait Trait {}
}

mod b {
    pub trait Trait {}
}

use a::*;
use b::*;
```

felixrabe:
See also previous discussion on #459 on Aug 15, 2018: https://github.com/rust-windowing/winit/issues/459#issuecomment-413129522 and https://github.com/rust-windowing/winit/issues/459#issuecomment-413183172 (at the end).

felixrabe:
My suggestion: Use plain `EventLoopExt` as the name for exclusive platform-specific traits, and `EventLoopDesktopExt` (as an example for Desktop) for traits that are available in addition.

felixrabe:
One more unsolicited idea concerning a possible module structure here:

    mod platform
        mod desktop
            trait EventLoopExt
            trait WindowExt

            mod macos
                trait EventLoopExt: super::EventLoopExt
                trait WindowExt: super::WindowExt

            mod unix (excl. macos, maybe also excl. fuchsia and redox)
                trait EventLoopExt: super::EventLoopExt
                trait WindowExt: super::WindowExt

        mod mobile
            trait EventLoopExt
            trait WindowExt

            mod ios
                :

nvzqz:
Another approach (which isn't pretty but just putting it out there) is to have `fn run_return` and `type UserEvent` go on each OS extension trait directly, and remove `desktop::EventLoopExt`. I'm not in favor of this approach but it is an option.

Osspial:
This is a pretty bikesheddy change. As such I don't feel all that strongly about it and I'm fine with any solution that addresses the problem @vberger mentioned, but since it's so low-impact I don't see a compelling reason to change it.

Osspial:
Hey,

I'd like to apologize for being so dismissive of this change in my comment yesterday. Not much discussion was had over the exact naming conventions for this in the original #459 thread, and we mostly settled on this because nobody proposed anything else. I'm open to changes here, but I'll still warn against getting too heated or spending too much time in this discussion - like many design choices, there isn't really a "right" answer here for the naming convention and it's a relatively minor part of the API to boot.

Regardless, here's my two cents on why I'm fine with the current situation: I like internal consistency in designs. It doesn't make a whole lot of sense to me to have a `Desktop` suffix on the desktop platform traits while not having a suffix on any of the other platforms, and having that inconsistency would increase the mental overhead for users of our API. Despite its clunkiness, remembering *there's a platform suffix for every platform trait* is easier than remembering *there's a platform suffix on these platform traits but not these other platform traits*.

To that end, I'd be okay with either:
* The current design
* The suffixless design, in light of `use WhateverExt as _`.

Both have benefits. Both are slightly clunky. I don't feel that strongly about it, and am willing to go with whatever develops a consensus, but if no consensus develops we're just going to default to the current solution.

ZeGentzy:
Closing for inactivity.

PR #909, opened 2019-06-11
Fix initial dimensions of a fullscreen window on Windows

aleksijuvani:
Previously, creating a fullscreen window on Windows would result in the window spanning more than a single monitor since the DPI factor was not taken into account. A workaround was to manually specify the size for the window, but this is no longer necessary.

### DIFF
--- a/src/window.rs
+++ b/src/window.rs
@@ -287,7 +287,7 @@ impl WindowBuilder {
         self.window.inner_size = Some(self.window.inner_size.unwrap_or_else(|| {
             if let Some(ref monitor) = self.window.fullscreen {
                 // resizing the window to the dimensions of the monitor when fullscreen
-                LogicalSize::from_physical(monitor.size(), 1.0)
+                LogicalSize::from_physical(monitor.size(), monitor.hidpi_factor()) // DPI factor applies here since this is a borderless window and not real fullscreen
             } else {
                 // default dimensions
                 (1024, 768).into()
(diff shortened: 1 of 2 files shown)


vberger:
I think this may be in conflict with https://github.com/rust-windowing/winit/pull/895

aleksijuvani:
I think you're right. From a cursory glance at the changes of #895, it would also appear to fix the issue. However, the fullscreen example in that branch appears to be broken, so I cannot test this.

aleksijuvani:
Yes, it looks like #895 now fixes this issue. I'll keep this pull request around in case we want to merge this before #895 lands.

src/window.rs
```diff
@@ -287,7 +287,7 @@ impl WindowBuilder {
         self.window.inner_size = Some(self.window.inner_size.unwrap_or_else(|| {
             if let Some(ref monitor) = self.window.fullscreen {
                 // resizing the window to the dimensions of the monitor when fullscreen
-                LogicalSize::from_physical(monitor.size(), 1.0)
+                LogicalSize::from_physical(monitor.dimensions(), monitor.hidpi_factor()) // DPI factor applies here since this is a borderless window and not real fullscreen
```
aleksijuvani: @Osspial I think this ought to be `.size()` now! I can fix this, unless you're already handling it?

Osspial: You're right! Feel free to go ahead and commit that change.

aleksijuvani: Done!

aleksijuvani:
I've rebased this against `master` again. It's kind if ridiculous that git is not able to resolve this one-line merge automatically..

Osspial:
Yeah, git merging on CHANGELOG has always been somewhat shaky, and I'm not entirely sure why...

I wonder if there's any way to change the merge algorithm git uses for that specific file.

PR #911, opened 2019-06-12
Change Monitor dimensions functions to size functions

Osspial:
  - [x] Test on macOS
  - [x] Test on iOS
  - [ ] Test on X11

So I realize that three separate PRs for this is more than is ideal, but there are a bunch of small gaps that slip through. I've looked over all of our publicly exposed functions though and I'm 95% sure this covers all of them.


Osspial:
So this build is failing, but I'm pretty sure it's just because Travis isn't actually doing the build on 3/4 of the Linux platforms.

Issue #891, opened 2019-05-30
Is ControlFlow supposed to be sticky? (eventloop-2.0)


aloucks:
Based on the doc comment, I would believe that it is intended to be sticky. However, the x11 and wayland implementations are resetting it per loop iteration. It doesn't look like Windows is doing this.

https://github.com/rust-windowing/winit/blob/0df436901a4835b47c1daf58a5260b839cf9dd6a/src/event_loop.rs#L65-L69

https://github.com/rust-windowing/winit/blob/0df436901a4835b47c1daf58a5260b839cf9dd6a/src/platform_impl/linux/x11/mod.rs#L283-L286

https://github.com/rust-windowing/winit/blob/0df436901a4835b47c1daf58a5260b839cf9dd6a/src/platform_impl/linux/x11/mod.rs#L289-L291

https://github.com/rust-windowing/winit/blob/0df436901a4835b47c1daf58a5260b839cf9dd6a/src/platform_impl/linux/x11/mod.rs#L301-L306

Osspial:
It is indeed intended to be sticky.

vberger:
Ah, ok. I assumed it was only supposed to be sticky during a run of the event loop, rather than forever.

aloucks:
> I assumed it was only supposed to be sticky during a run of the event loop, rather than forever.

Yes, that's how I understand it; Control flow is sticky for the duration of `run_return` and reset in between calls to that. The current implementation resets the control flow on each tick of the loop.

This part is correct:

https://github.com/rust-windowing/winit/blob/0df436901a4835b47c1daf58a5260b839cf9dd6a/src/platform_impl/linux/x11/mod.rs#L226-L229

vberger:
Well, the fix is simple enough, I just don't have the time to work on it for now.

Issue #830, opened 2019-04-07
Preparation for contributor marketing push


Osspial:
Winit needs more people actively working to improve it; the current crop of maintainers has been doing great work, but due to the nature of volunteer work said efforts aren't guaranteed and we don't have any contingency plans in the event that a platform maintainer stops being able to work on their platform. That is unacceptable, given Winit's foundational place in the Rust windowing, GUI, and gamedev ecosystems. I'd like to start a broader marketing push to get more people working on each of the backends (which I'm happy to helm), but there are a few questions I'd like to get resolved before getting started on that:

1. Should a single person be allowed to maintain more than one platform?
2. What is the maintainership status of the @francesca64's (macOS, X11, and Android) platforms?
3. Should we implement an automated mechanism for assigning contributors to bugs?
4. In `CONTRIBUTING.md`, how should we list individuals that are able to contribute code to a platform, but don't necessarily have the time to maintain it? We don't have a role listed explicitly for doing that.
5. Should we move Winit and related projects into a new GitHub organization?

I've got personal stances on most of those issues, but I'm going to reserve those for a separate comment so to keep the starting post relatively unbiased.

cc @tomaka @francesca64 @mitchmindtree @vberger @mtak- @ZeGentzy @ryanisaacg

Osspial:
My stances:

1. I'd say no. Doing so creates a massive single point of failure upon which on the entire Winit project can stall, which we've run into trying to get the EventLoop-2.0 changes out the door. Reducing the amount of work somebody is responsible for makes it easier to replace them if they can't do their work.
2. @francesca64 has been MIA for several months, and from private conversations with her it doesn't sound like she has the time to contribute to Winit. I'd like to market all of these platforms as needing new maintainers given the extended inactivity on them, but I haven't gotten her explicit permission to do that and I don't like the idea of forcibly unlisting somebody from work they volunteered to do.
3. The current bug-fixing system is incredibly ad-hoc - when a bug for a given platform gets reported, we rely on somebody stepping up to find a solution. With the current number of active contributors, this isn't sustainable: oftentimes, it takes several weeks for the maintainer to find the time to investigate a platform's bugs, and there are some bugs in the backlog that have been sitting untouched for many months. Given that, I feel like adding an automated bug-assignment system is essential, but doing that requires having a greater network of people to which we can assign bugs.
[...]

PR #913, opened 2019-06-13
Release 0.20.0 Alpha 1

### LINKED ISSUE #219: MacOS can't wake up during resize, breaks timeout schemes
As the [comment describes](https://github.com/tomaka/winit/blob/a08347eef0e9d103be3d1aa817fbfc0fb1796985/src/platform/macos/events_loop.rs#L242-L247), `nextEventMatchingMask:untilDate:inMode:dequeue:` blocks during a resize and does not return events until the resize is complete. That includes not returning events sent via `postEvent:atStart:`, which means [`Proxy::wakeup()`](https://github.com/tomaka/winit/blob/a08347eef0e9d103be3d1aa817fbfc0fb1796985/src/platform/macos/events_loop.rs#L533-L544) doesn't actually cause `run_forever()` to wake up.

This property is especially unfortunate with `glutin_window`, which implements its own `fn wait_event_timeout(&mut self, timeout: Duration)` by calling `run_forever()` and interrupting when the timeout expires. Since it _can't_ interrupt, a `wait_event_timeout()` that wants to yield 60 fps to e.g. drive a render loop will simply hang entirely until the user is done resizing.

I've been experimenting with ways to get `nextEventMatchingMask:` to be less thorny, but another complimentary approach would be to expose a new `run_until_timeout()` call in addition to `poll_events()` and `run_forever()` -- or perhaps to deprecate `poll_events()` and `run_forever()` in favor of `run_until_timeout(Some(0))` and `run_until_timeout(None)`. A timeout-aware API would let `glutin_window` pass its deadline all the way down to `nextEventMatchingMask:untilDate:`, avoid spawning needless timeout threads, and sidestep some of the resize-related blocking issues all at once.

### LINKED ISSUE #276: run_forever cannot bulk process events, unable to empty event queue
In a program that uses `poll_events` in an infinite loop, `poll_events` will process events until the event queue is empty. If multiple events are enqueued while the program is rendering, the program can process each of them before beginning the relatively expensive task of rendering the next frame.

With `run_forever` though, this sort of "bulk event processing" cannot happen. The callback does not know if more events are in the queue, so it must unconditionally render a frame to the user. But while rendering a frame, more events can be enqueued. If rendering takes too long, the input queue becomes filled with events, and the `run_forever` callback cannot process them fast enough as it has to process older events first.

This severely affects interactivity; the program appears to the user that the program is constantly "catching up" with what the user did seconds ago (this is particularly noticeable when moving the mouse, which causes many events).

---

There are a couple workarounds to this issue. The obvious one is to use `poll_events` instead. However, if the program only needs to re-render in response to user input (common in editors), rendering all of the time wastes CPU and power.

A more complicated workaround is to have the event loop forward all of its events to the program's own queue (such as [`std::sync::mpsc`](https://doc.rust-lang.org/std/sync/mpsc/)), which can implement a "check if empty function". This introduces complexity in the program, though, and it would be better if `winit` had a simpler interface.

---

The API described in #231 would solve this issue. The program can call `get_event` with a short timeout until no events are left, then render the frame and call `get_event` with an infinite timeout.

### LINKED ISSUE #413: How do I build a new window inside run_forever or poll_events?
I'd like to be able to spawn a new window in response to user input.

Since the event loop is borrowed mutably by `run_forever` and `poll_events`, there doesn't appear to be a way to spawn new windows in response to input events.

This may be another facet of https://github.com/tomaka/winit/issues/231

### LINKED ISSUE #459: API Change Proposal: Events Loop 2.0
# UPDATE 8/23/18:
A PR for Windows is available at #638

# UPDATE 7/12/18:
I've adapted the agreed-upon changes into a changelog here: https://gist.github.com/Osspial/d3a3390807fa95a469d9c9b677db4715. The original proposal is below.

------

# Events Loop 2.0

This issue proposes a series of changes to winit's `EventsLoop` API, based around the experiences I've had with creating a GUI library that uses winit as a window creation backend. The main goal is to give winit complete control over the event loop, both for event polling and event waiting. These changes should let us fix various implementation bugs on the win32 backend, as well as fix what I believe are design bugs in the current API.

## Technical Motivation:

**EDIT:** The technical motivation for the Windows backend is included below. However, pretty much every platform winit supports is hacky or broken in some way by necessity of the current API we expose.

Right now, the `win32` backend creates a background thread, in which the events loop resides. The biggest problem with this is that it introduces various unexpected synchronization bugs that require a good amount of effort to fix and increase the complexity of the backend (see #415 and #391). However, it's necessary to cleanly handle `win32`'s various modal event loops.

So, what are these modal loops, and why do they matter? On Windows, when a user begins resizing a window, Windows suspends the program's main event loop and begins its own loop, inside of the resize event's handler. Unfortunately, this freezes any processing that's going on outside the Windows event loop, which is currently worked around by having the loop exist in its own thread, introducing quite a few bugs as mentioned above. Inside the loop, Windows does a bunch of additional processing, most notably handling window snapping (although it probably does other stuff too that isn't obvious).

Windows doesn't publicly expose the APIs used to handle snapping, so re-implementing the modal loops ourselves so that they don't suspend the main event loop isn't really an option. However, Windows should still process events we push to the Windows events queue, letting us implement event polling inside the Windows event loop with custom events ([using the same mechanisms used for wakeup messages](https://github.com/tomaka/winit/blob/bdc01fee1a54470b07a897592927c26d1af60341/src/platform/windows/events_loop.rs#L312-L328)), as long as winit's API lets it assume complete control over program control flow. Right now, it doesn't, but this change would let us remove the second thread and simplify much of the implementation.

Actually removing the second thread isn't that difficult (see #445), but these bugs present pretty major blockers that should be resolved before that's merged.

## Design Details

**EDIT:** The original design has been removed from this post, as it is outdated and may cause confusion for new readers. However, it does contain some useful information about the rationale behind the changes, and can be accessed here: https://gist.github.com/Osspial/53efa87797899872650f6bece14736c0

### LINKED ISSUE #728: Function naming inconsistency

### LINKED ISSUE #794: Replace Option<T> in Window function return types with T

### LINKED ISSUE #825: OpenBSD: Need to bump wayland-client and smithay-client-toolkit version numbers
Hello,

I found out today that winit no longer builds on OpenBSD, since it uses an older version of wayland-client (v0.21) and an older version of smithay-client-toolkit (v0.4.3) that both depend on an older nix crate (v0.12) that does not compile on OpenBSD. Bumping wayland-client to v0.23.1 and smithay-client-toolkit to v0.6.0 fixes this issue for me.

Thanks,
Rob

Osspial:
As per https://github.com/rust-windowing/winit/issues/830#issuecomment-501180361. I want to get #911 merged and #891 resolved before this, but... yeah. We've still got a long way for Winit to go, but we've made it this far, and I can only see progress speeding up from here. I cannot express enough thanks to the people that have helped get it here, from those writing the code, to the people testing stuff out, the people contributing their thoughts in the issues, and everyone in-between.

It's kinda surreal to be able to do this.

🎉

Closes #825, closes #276, closes #794, closes #728, closes #413, closes #219, closes #459.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+# 0.20.0 Alpha 1
+
 - Changes below are considered **breaking**.
 - Change all occurrences of `EventsLoop` to `EventLoop`.
 - Previously flat API is now exposed through `event`, `event_loop`, `monitor`, and `window` modules.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "winit"
-version = "0.19.1"
+version = "0.20.0-alpha1"
 authors = ["The winit contributors", "Pierre Krieger <pierre.krieger1708@gmail.com>"]
 description = "Cross-platform window creation library."
 keywords = ["windowing"]
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 
 ```toml
 [dependencies]
-winit = "0.19.1"
+winit = "0.20.0-alpha1"
 ```
 
 ## [Documentation](https://docs.rs/winit)


aleksijuvani:
I've rebased #909 if we want to get that in as well. I assume we're fine with the Android and Emscripten backends not compiling for this release?

Osspial:
@aleksijuvani Thanks! We should get that PR in, too.

I'm fine with Android not compiling, since waiting for it to get done without anyone on board that can actually do it is a recipe for waiting forever. As far as Emscripten goes... well, I'm not actually sure what we're going to end up *doing* with the Emscripten backend. We've got a couple people working on `wasm` backends, and once those get merged I can't see Emscripten getting much use since it's such a pain to set up. I'm inclined to just remove it, but there hasn't been any formal talk about that, so I'm not going to pass judgement on it just yet.

aleksijuvani:
Right, I figured that was the case! Regarding Android, technically, I think anyone *could* do it, seeing as how the Android simulator is freely available for download. It's just that it's hard to justify putting in the effort for a platform that the person doing the work isn't personally invested in.

vberger:
Going to do a quick a PR for #891

vberger:
> It's kinda surreal to be able to do this.

And pretty relieving as well! Glad to see we finally reached this point. :tada:

felixrabe:
From my POV, you can just merge this. It's not as if you were instantly making a release, but preparing for it. (IMO, immediately after a release, Cargo.toml version and CHANGELOG header should reflect an upcoming version instead of the last released one anyway.)

Osspial:
@felixrabe The CI is set up to automatically release when the version gets incremented on `master`.

vberger:
@Osspial The two issues you linked in the original post are now resolved though, are there other things you want to finish before releasing the alpha?

Osspial:
@vberger Nope, stuff looks good! I'll pull together a post for this and release this/announce the marking push at some point later today.

Osspial:
@ZeGentzy is Glutin ready for this release?

ZeGentzy:
It's short an iOS backend. Beyond that, I guess it's releaseable, although I didn't get in nearly the amount of things I wanted done, due to exams hoovering up my time.

Osspial:
[First draft of release post, if anyone would like to feedback it.](https://gist.github.com/Osspial/667344cda503df3490a58a8ccf91c700) That combines the contributor marketing push with some new stuff detailing the major changes we've made and why.

There are still placeholder links in there that I'd like to fix, but I'm holding off on doing those until I've made all the things to link to.

vberger:
This drafts looks pretty awesome ! :+1: I personally have no changes to suggest.

Osspial:
Alright, stuff is pretty much ready! Took longer than I expected, but hell, what doesn't?

I'll hold off on posting this until tomorrow morning so we can catch the reddit/hackernews visibility wave, but we should get the alpha onto crates.io sometime later tonight so we can be sure that everything is working as we'd like.

WRT the issue list [here](https://gist.github.com/Osspial/fddc1fdcbd65a0f6510c9669b21b4901): it's probably best to just tag those issues and link to our issue page, with the appropriate filters. That'll be more accurate than maintaining a separate list of issues we need help on.

ZeGentzy:
> WRT the issue list [here](https://gist.github.com/Osspial/fddc1fdcbd65a0f6510c9669b21b4901): it's probably best to just tag those issues and link to our issue page, with the appropriate filters. That'll be more accurate than maintaining a separate list of issues we need help on.
On it.

Osspial:
I'm using GitHub's pinned issues feature to highlight important issues related to this release. If anyone thinks other issues are more important than what's been pinned, please raise your concerns here.

ZeGentzy:
And done!

Osspial:
@ZeGentzy Thanks!

One thing - why do we want a `Contributor Push Issue` tag? I'm not entirely sure what it's intended to communicate.

ZeGentzy:
"This issue looks relatively still up to date to us, and should be ready-ish for solving." The name `Contributor Push Issue` doesn't exactly convey that, but I couldn't think of one better.

I'm sure it's quickly going to fall out of sync, but oh well.

aloucks:
This might be a good time to run `cargo fmt` (once all outstanding 0.20-alpha PRs are merged). I'd also recommend adding `cargo fmt --check` to CI to enforce it.

Osspial:
@ZeGentzy if that's the case, I think I'd rather just remove the tag. I'm not sure it conveys enough information to be useful, and it's unclear when to apply it - the meanings you've applied to it seem to be already covered by other tags.

@aloucks That's a pretty good idea! I'll go ahead and pull that together.

Osspial:
https://crates.io/crates/winit
https://crates.io/crates/glutin
https://users.rust-lang.org/t/winit-0-20-the-state-of-windowing-in-rust-and-a-request-for-help/29485
https://www.reddit.com/r/rust/comments/c3cmqb/winit_020_the_state_of_windowing_in_rust_and_a/
https://news.ycombinator.com/item?id=20243853

It is done.

aloucks:
Nice!

One thing -- you might consider re-formatting using rustfmt with only stable options. I accidentally ran `cargo fmt` after pulling down the latest code and nearly every file was modified. I didn't realize what the problem was until I went spunking through this PR and found that it used `+nightly`. Is requiring nightly the right choice? It could cause some friction with PRs.

aleksijuvani:
If I remember correctly, I used to use rustfmt with `+nightly` and I found that it occasionally made significant formatting changes that broke the build when used in conjunction with `--check`. I would recommend switching to stable rustfmt.

Osspial:
@aloucks @aleksijuvani I'd be fine with switching to stable `rustfmt`. I normally live on nightly, so I didn't notice that some of the options were unstable 😄!

That being said, if we're sticking with stable `rustfmt` I think we should disable formatting for the examples. Without the unstable `force_multiline_blocks` flag, some of the examples become significantly harder to read. Compare the `match event` clauses in these two cases:

with nightly:
```rust

window.set_title("A fantastic window!");

let mut cursor_idx = 0;

    match event {
        Event::WindowEvent {
            event:
                WindowEvent::KeyboardInput {
                    input:
                        KeyboardInput {
                            state: ElementState::Pressed,
                            ..
                        },
                    ..
                },
            ..
        } => {
            println!("Setting cursor to \"{:?}\"", CURSORS[cursor_idx]);
            window.set_cursor_icon(CURSORS[cursor_idx]);
            if cursor_idx < CURSORS.len() - 1 {
                cursor_idx += 1;
            } else {
                cursor_idx = 0;
            }
        },
        Event::WindowEvent {
            ..
        } => {
            return;
        },
        _ => (),
    }
});
```

with stable:
```rust

window.set_title("A fantastic window!");

let mut cursor_idx = 0;

event_loop.run(move |event, _, control_flow| match event {
    Event::WindowEvent {
        event:
            WindowEvent::KeyboardInput {
                input:
                    KeyboardInput {
                        state: ElementState::Pressed,
                        ..
                    },
                ..
            },
        ..
    } => {
        println!("Setting cursor to \"{:?}\"", CURSORS[cursor_idx]);
        window.set_cursor_icon(CURSORS[cursor_idx]);
        if cursor_idx < CURSORS.len() - 1 {
            cursor_idx += 1;
        } else {
            cursor_idx = 0;
        }
    }
    Event::WindowEvent {
        ..
    } => {
        return;
    }
    _ => (),
});
```

aloucks:
I had a hard time spotting the difference even when I put those two examples in side-by-side windows :)

I personally don't think the stable option here is more difficult to read. I actually prefer the the slight vertical compression, although I could barely notice the difference.

I would leave rustfmt turned on for everything and use `#[rustfmt::skip]` where needed (I tend to use it for things like matrix definition when I want columns lined up nicely). 

Another option that I'd recommend is to set `max_width = 120` (the default is 100), but I guess that's a little more subjective.

aleksijuvani:
I can't tell the difference between the two at a quick glance either. The stable option looks fine to me. As for the line width, I personally use `max_width = 80`, but I realize that might not be enough for everyone. I would probably just leave it as the default. I think `max_width = 120` is too much when working in a two-column editor layout. Another option probably worth mentioning is `newline_style`, which can enforce consistent line endings, but this will result in difficult merge conflicts if there's already mixed line endings. All in all, the defaults are very reasonable to me.

aloucks:
> All in all, the defaults are very reasonable to me.

Agreed

felixrabe:
> Another option probably worth mentioning is `newline_style`, which can enforce consistent line endings, but this will result in difficult merge conflicts if there's already mixed line endings.

The merge trouble for any whitespace-only change like this is easily contained:

- Apply the change programmatically in a separate git commit on the whole project. (I actually use a script called [`git-exec`](https://github.com/felixrabe/bin/blob/master/git-exec) that helps apply commits whose messages start with `$ ...` in that way.)

- Do the same to any branch before merging. Rebasing is a bit cumbersome but doable: Apply whitespace to base, go to first commit, apply whitespace, then make a commit diffing first+white to second+white. Repeat. (I'm using this strategy regularly. I wonder if anyone came up with a more fluent way to do this?)

- This alpha period is a good time anyway to make such "messy" changes.

Osspial:
Eh, I suppose that's just me being overly picky. I'd be fine with killing the non-stable options for now.

I'd still like to enable `merge_imports` at some point when that's stable, but that can wait.

PR #921, opened 2019-06-13
Update example in README.md and add move prefix to window.rs example closure

Osspial:
I just realized that the example in `README.md` was still using the old API! Wouldn't have wanted to let that slip through...

This also makes the event closure in `window.rs` a `move` closure. That change isn't *technically* necessary, but most practical modifications to `window.rs` are going to require it, so I figure we might as well add it in to reduce friction for new users using it as a base.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,18 +32,27 @@ another library.
 
 ```rust
 extern crate winit;
+use winit::window::WindowBuilder;
+use winit::event::{Event, WindowEvent};
+use winit::event_loop::{EventLoop, ControlFlow};
 
 fn main() {
-    let mut event_loop = winit::EventLoop::new();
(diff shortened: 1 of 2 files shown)


felixrabe:
For me, it rather confuses me: "Why is this 'move'?" I prefer to not write it if it is not required.

vberger:
Well, the provided closure must be `'static`, so as long as it captures anything it must be `move`. This example that is so simple that it does not need it, but I expect it to be the exception, not the norm.

Osspial:
What we *could* do is make the example trivially capture `window`, so the closure has to be `move`. I've done that in the latest commit.

Osspial:
Are there any complaints with the new example changes? If not, I'll merge this.

Issue #912, opened 2019-06-13
Update to Rust 2018 Edition


Osspial:
See #612, but in issue form.

crlf0710:
One thing that previously mentioned but not yet done is migrating to 2018 module structure.

Tried to do the migration locally and filed https://github.com/rust-lang/rust/issues/61924 .

felixrabe:
^ The issue created by @crlf0710 was marked as a duplicate of rust-lang/rust#59345.

murarth:
Closed by #924

PR #924, opened 2019-06-15
Migrate to 2018 edition.

crlf0710:
- [X] Created an example program if it would help users understand this functionality
- [X] Updated [feature matrix](https://github.com/tomaka/winit/blob/master/FEATURES.md), if new features were added or implemented

### DIFF
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -13,9 +13,9 @@ use std::{fmt, error};
 use std::time::Instant;
 use std::ops::Deref;
 
-use platform_impl;
-use event::Event;
-use monitor::{AvailableMonitorsIter, MonitorHandle};
+use crate::platform_impl;
+use crate::event::Event;
+use crate::monitor::{AvailableMonitorsIter, MonitorHandle};
 
 /// Provides a way to retrieve events from the system and from the windows that were registered to
 /// the events loop.
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -146,10 +146,8 @@ impl<T> EventLoop<T> {
     }
 
     /// Returns the list of all the monitors available on the system.
-    ///
-    // Note: should be replaced with `-> impl Iterator` once stable.
     #[inline]
-    pub fn available_monitors(&self) -> AvailableMonitorsIter {
+    pub fn available_monitors(&self) -> impl Iterator<Item = MonitorHandle> {
         let data = self.event_loop.available_monitors();
         AvailableMonitorsIter{ data: data.into_iter() }
     }
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -46,13 +46,13 @@ pub struct EventLoopWindowTarget<T: 'static> {
 }
 
 impl<T> fmt::Debug for EventLoop<T> {
-    fn fmt(&self, fmtr: &mut fmt::Formatter) -> fmt::Result {
+    fn fmt(&self, fmtr: &mut fmt::Formatter<'_>) -> fmt::Result {
         fmtr.pad("EventLoop { .. }")
     }
 }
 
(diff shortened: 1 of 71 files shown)


crlf0710:
OK, seems the CI passed. The windows and linux migration was done by cargo-fix, and macos, android, ios, emscripten migration was done manually. Not really sure if there's anything that's covered by CI. If everything's right, this should close #912.

aleksijuvani:
I would imagine all this causing a lot of merge conflicts, so that's something to keep in mind. I also think we can do better here. We should be able to remove the `extern crate` declarations from `lib.rs` for example. I think we should also avoid sprinkling `crate::` everywhere and instead prefer to `use` the types in question, or use `super::` where applicable.

crlf0710:
1. I have remove most of the `extern crate`s, but i'm leaving those with `macro_use`s alone.
2. I've also changed trait objects to use `dyn` syntax.
3. I've addressed the above mentioned `impl Iterator` change.

I haven't renamed the `mod.rs` yet because last time i heard that some IDE plugins doesn't support the `{module_name}.rs` project structure. Let's leave it for a subsequent PR.

src/event_loop.rs
```diff
@@ -149,7 +149,7 @@ impl<T> EventLoop<T> {
     ///
     // Note: should be replaced with `-> impl Iterator` once stable.
```
felixrabe: This comment no longer applies.

Osspial:
@crlf0710 Do you know which plugins, exactly? The module name change is a pretty significant ergonomics improvement for non-IDE users, and I don't think it makes sense to delay for `non-specific IDE` to support it. If we *do* end up delaying, we should at least know exactly what we're delaying for, so the change doesn't just languish in the backlog.

felixrabe:
@Osspial The `foo/mod.rs` -> `foo.rs` change is not mandatory. It is just a structure option added by 2018. I'm also for postponing this into an own issue to discuss this separately.

Osspial:
@felixrabe It's not strictly mandatory, but neither is removing `extern crate` - they're both aesthetic changes that make the code easier to read and maintain, which is why 2018 edition introduced them. I'm not necessarily opposed to delaying and discussing it, but I don't see what sort of discussion there can *be* (beyond the *this breaks the editor workflow for this IDE* issue @crlf0710 brought up) - it's clearly the direction the language is pushing, and the new structure wouldn't have been added to the language if it wasn't intended for people to adopt it.

CHANGELOG.md
```diff
@@ -296,7 +296,7 @@
 # Version 0.11.1 (2018-02-19)
 
 - Fixed windows not receiving mouse events when click-dragging the mouse outside the client area of a window, on Windows platforms.
-- Added method `os::android::EventsLoopExt:set_suspend_callback(Option<Box<Fn(bool) -> ()>>)` that allows glutin to register a callback when a suspend event happens
+- Added method `os::android::EventsLoopExt:set_suspend_callback(Option<Box<dyn Fn(bool) -> ()>>)` that allows glutin to register a callback when a suspend event happens
```
Osspial: We shouldn't retroactively modify the changelog.

src/event_loop.rs
crlf0710: Done

CHANGELOG.md
crlf0710: Done.

crlf0710:
> Do you know which plugins, exactly?

It's either RLS or Intellj-Rust, but i can't remember which. There was also some hassle dealing with the #[path = "blah"] things.

If you wish, we can experiment again immediately in another PR after this PR is merged, and see whether there's still problems with the newest version of these plugins. I just want do these uncertain things within a separately PR so it can be easily reviewed or even reverted later if it is proved problematic.

Osspial:
@crlf0710 That sounds good to me.

PR #925, opened 2019-06-15
Add exclusive fullscreen mode

### LINKED ISSUE #717: Windows: Fullscreen mode is borderless window, not exclusive fullscreen
In exclusive fullscreen mode, the Window completely takes over drawing to a monitor and doesn't allow other windows to draw on that monitor, while the current mode just draws the window over the taskbar and without borders. Winit should definitely expose both modes, but the current API may be a bit misleading.

Where does this stand versus other operating systems? Do the other backends fully take over rendering to the monitor, or just draw over all other elements on the screen?

aleksijuvani:
Resolves #717. This uses `CGConfigureDisplayWithDisplayMode` on macOS for now, but I think what I'm going to end up doing is move to use `enterFullScreenMode` instead, since that seems to be the preferred way to do fullscreen on macOS. Implementing this on X11 should be fairly straight-forward. I'm not sure how to do this on Wayland however.

     - [x]  Windows
     - [x]  macOS
     - [x]  X11
     - [x]  Wayland
     - [x]  iOS
- [x] Updated [feature matrix](https://github.com/tomaka/winit/blob/master/FEATURES.md), if new features were added or implemented


vberger:
> I'm not sure how to do this on Wayland however.

That'll be pretty simple actually: the wayland server will never give any app direct control over the monitor, hence "exclusive fullscreen" does not exist, only regular fullscreen.

examples/fullscreen.rs
```diff
@@ -118,14 +83,39 @@ fn prompt_for_monitor(event_loop: &EventLoop<()>) -> MonitorHandle {
...
 
     println!("Using {:?}", monitor.name());
 
     monitor
 }
+
+fn prompt_for_video_mode(monitor: &MonitorHandle) -> VideoMode {
+    // Video modes are returned in a random order, so we must store them in
+    // order to be able to later pick the nth video mode reliably
+    let mut video_modes: Vec<_> = monitor.video_modes().collect();
```
Osspial: For the sake of consistency when running this example, could the video modes get sorted here?

src/platform_impl/windows/window.rs
```diff
@@ -389,23 +389,49 @@ impl Window {
...
+                        }
+
+                        mark_fullscreen(window.0, false);
+                        return;
+                    });
+                }
+                Some(fullscreen) => {
+                    let (position, size): ((i32, i32), (u32, u32)) = match fullscreen {
+                        Fullscreen::Exclusive(ref video_mode) => (
+                            video_mode.monitor().position().into(),
```
Osspial: Right now, this retrieves the monitor's position before the video mode transition. What ends up happening is that (at least on my multi-monitor setup), the monitor's position ends up changing after the video mode changes, so the window gets set to the wrong position. We should be retrieving this value after we've changed the video mode.

src/monitor.rs
```diff
@@ -50,17 +50,16 @@ impl Iterator for AvailableMonitorsIter {
 /// - [`MonitorHandle::video_modes`][monitor_get].
 ///
 /// [monitor_get]: ../monitor/struct.MonitorHandle.html#method.video_modes
-#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+#[derive(Derivative)]
+#[derivative(Clone, Debug = "transparent")]
 pub struct VideoMode {
```
Osspial: I realize this specific feedback would've been better-placed in the original PR, but it's only occurred to me now: is it possible to add a function to `Monitor` to retrieve the currently-active video mode? That way, applications can launch in fullscreen with the sensible default of *whatever fullscreen settings the monitor currently has*, instead of having to guess which mode to use.

src/platform_impl/windows/window_state.rs
```diff
@@ -24,7 +23,7 @@ pub struct WindowState {
     pub saved_window: Option<SavedWindow>,
     pub dpi_factor: f64,
 
-    pub fullscreen: Option<MonitorHandle>,
+    pub fullscreen: Option<Fullscreen>,
```
Osspial: Placing this comment here, since github won't let me do it in the actual flags struct: exclusive fullscreen windows should set the `ALWAYS_ON_TOP` flag. The best way to do that is probably to add a `FULLSCREEN_OR_MASK` field, include `ALWAYS_ON_TOP` in it, and apply the mask in `fn mask`.

src/platform_impl/windows/window_state.rs
```diff
@@ -123,6 +122,53 @@ impl WindowState {
         this.window_flags.set(WindowFlags::MARKER_FULLSCREEN, is_fullscreen);
         let new_flags = this.window_flags;
 
+        // Change video mode if necessary
+        if old_flags & WindowFlags::MARKER_FULLSCREEN != new_flags & WindowFlags::MARKER_FULLSCREEN
```
Osspial: Right now, this won't update the video mode if the video mode is changed while fullscreen is active.

src/platform_impl/windows/window_state.rs
aleksijuvani: I tried doing this earlier, but it didn't seem to make any difference in behaviour whatsoever. It made me think that maybe the window style for it is not getting applied correctly.. I didn't look into it any further.

src/platform_impl/windows/window.rs
aleksijuvani: Interesting! This didn't seem to occur on my system. I'll take a look.

aleksijuvani:
The macOS implementation is starting to look pretty good. The only remaining issue there is that upon exiting exclusive fullscreen and re-entering it, the video mode change seems to fail. Perhaps the list of video modes is invalidated upon capturing the display.

With these new fullscreen modes, I'm inclined to deprecate `set_simple_fullscreen` because that fullscreen behaviour seems to me very strange and unidiomatic, and most certainly not what the user is going to expect. I won't touch upon that in this pull request however, and will leave that open for discussion.

src/monitor.rs
aleksijuvani: Shouldn't be a problem, but I'm inclined to do this in a separate pull request since I will have to implement it for all four platforms. I think the borderless mode would suffice for a sensible default in any case.

examples/fullscreen.rs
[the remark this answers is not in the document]
aleksijuvani: Sure, that seems reasonable. Coincidentally, the list of video modes has a consistent order on macOS, but it's randomized on Windows due to it being a `HashMap` under the hood.

src/platform_impl/windows/window_state.rs
Osspial: Really? This patch seems to work on my machine:

```patch
diff --git a/src/platform_impl/windows/window_state.rs b/src/platform_impl/windows/window_state.rs
index 67317c89..9aabd91d 100644
--- a/src/platform_impl/windows/window_state.rs
+++ b/src/platform_impl/windows/window_state.rs
@@ -75,6 +75,7 @@ bitflags! {
             WindowFlags::RESIZABLE.bits |
             WindowFlags::MAXIMIZED.bits
         );
+        const FULLSCREEN_OR_MASK  = WindowFlags::ALWAYS_ON_TOP.bits;
         const NO_DECORATIONS_AND_MASK = !WindowFlags::RESIZABLE.bits;
         const INVISIBLE_AND_MASK = !WindowFlags::MAXIMIZED.bits;
     }
@@ -210,6 +211,7 @@ impl WindowFlags {
     fn mask(mut self) -> WindowFlags {
         if self.contains(WindowFlags::MARKER_FULLSCREEN) {
             self &= WindowFlags::FULLSCREEN_AND_MASK;
+            self |= WindowFlags::FULLSCREEN_OR_MASK;
         }
         if !self.contains(WindowFlags::VISIBLE) {
             self &= WindowFlags::INVISIBLE_AND_MASK;
```

aleksijuvani:
I've addressed most of your notes regarding the Windows implementation. I still have to fix the case where the video mode is changed while already in exclusive fullscreen mode. I realized that this hasn't been tested on the macOS implementation either, so I will have to create a test case for this. As for sorting the video modes, I've implemented this for Windows. It may be better to do this on the public API level (so that it's consistent across platforms and that the user can take advantage of the `Ord` trait implementation). I'll take a look at this tomorrow.

I've marked exclusive fullscreen `unimplemented!()` on iOS for now, as I can't seem to be able to detect external displays consistently with the simulator (and they report only a single video mode). I would have to test this on a physical device, which I would need to go out and buy some cables for..

Once I've worked out the remaining kinks, all that remains is the X11 implementation. I think I may have to end up doing a proper Linux install for this, as the virtual machine video driver only reports a single video mode. I'll also have to mark this `unimplemented!()` for Wayland. Or perhaps we want to modify the API so that `set_fullscreen` returns `Result`, with it always being an `Err` on Wayland?

While writing this, I was reminded that we could potentially have `set_fullscreen` called on a window when a different window may already be in fullscreen mode.. I think it might be good for us to track this and disallow `set_fullscreen` if there already is a fullscreen window. It'd be good to return an `Err` for this reason as well.

felixrabe:
Q project people: Could we merge this in a few smaller commits than just one mega-commit if accepted?

Q everyone: Maybe split this up into smaller chunks merged in separate PRs or something?

Why I suggest this: Mega-commits are hard to read.

aleksijuvani:
The macOS and Windows implementations have had all issues addressed and should be ready for final review. I will still have to do the implementation for X11.

I don't think splitting this into multiple pull requests makes any sense. This just adds friction for no benefit, as they will all be merged in anyway. I could squash these commits into a single commit per backend, and we could merge those commits as is without squashing again, but I don't think that's hardly worth the effort. I would suggest squash-merging this as usual in a single commit. If you wish to only inspect a single backend in the commit, you can do so easily by ignoring files in the other folders.

aleksijuvani:
It doesn't look like core-foundation-rs is actively maintained, so instead of waiting for a review for servo/core-foundation-rs/pull/318, I might have to pull this functionality into our `ffi.rs` for macOS.

With the exception of that, this should be all done! Implemented on all supported platforms (i.e. macOS, Windows and X11), with the exception of iOS, but I feel like it's quite a niche use-case there anyway.

src/monitor.rs
```diff
@@ -52,17 +52,41 @@ impl Iterator for AvailableMonitorsIter {
...
+    fn partial_cmp(&self, other: &VideoMode) -> Option<std::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for VideoMode {
+    fn cmp(&self, other: &VideoMode) -> std::cmp::Ordering {
+        // TODO: we can impl `Ord` for `PhysicalSize` once we switch from `f32`
+        // to `u32` there
+        let size: (u32, u32) = self.size().into();
```
ZeGentzy: What is this code sorting by? Can we at least get a comment for the users?

examples/multithreaded.rs
```diff
@@ -19,6 +19,8 @@ fn main() {
             .with_inner_size(WINDOW_SIZE.into())
             .build(&event_loop)
             .unwrap();
+        let video_modes: Vec<_> = window.current_monitor().video_modes().collect();
```
ZeGentzy: What when they move it between monitors....

aleksijuvani: The reason this is here in the first place is so that we can cycle between video modes with the arrow keys. The window will be placed fullscreen on the monitor it first appeared on. Video modes can differ between monitors and tracking this per monitor seems like it would unnecessarily complicate this example.

ZeGentzy:
Tested on X11 with this glutin example:
<details>
```
mod support;

use std::io::{stdin, stdout, Write};
use glutin::event::{ElementState, Event, KeyboardInput, VirtualKeyCode, WindowEvent};
use glutin::event_loop::{ControlFlow, EventLoop};
use glutin::monitor::{MonitorHandle, VideoMode};
use glutin::window::{Fullscreen, WindowBuilder};

fn main() {
    let el = EventLoop::new();

    print!("Please choose the fullscreen mode: (1) exclusive, (2) borderless: ");
    stdout().flush().unwrap();

    let mut num = String::new();
    stdin().read_line(&mut num).unwrap();
    let num = num.trim().parse().ok().expect("Please enter a number");

    let monitor = Some(match num {
        1 => Fullscreen::Exclusive(prompt_for_video_mode(&prompt_for_monitor(&el))),
        2 => Fullscreen::Borderless(prompt_for_monitor(&el)),
        _ => unreachable!("Please enter a valid number"),
    });

    let mut is_fullscreen = monitor.is_some();
    let mut is_maximized = false;
    let mut decorations = true;
    let mut video_mode = 0usize;

    let wb = WindowBuilder::new()
        .with_title("Hello world!")
        .with_fullscreen(monitor);
    let windowed_context = glutin::ContextBuilder::new()
        .build_windowed(wb, &el)
        .unwrap();

    let windowed_context = unsafe { windowed_context.make_current().unwrap() };

    let gl = support::load(&windowed_context.context());

    el.run(move |event, _, control_flow| {

        match event {
                WindowEvent::RedrawRequested => {
                    gl.draw_frame([1.0, 0.5, 0.7, 1.0]);
                    windowed_context.swap_buffers().unwrap();
                }
                WindowEvent::KeyboardInput {
                    input:
                        KeyboardInput {
                            virtual_keycode: Some(key),
                            state,
                            modifiers,
                            ..
                        },
                    ..
                } => match (key, state) {
                    (VirtualKeyCode::Escape, _) => *control_flow = ControlFlow::Exit,
                    (VirtualKeyCode::F, ElementState::Pressed) => {
                        is_fullscreen = !is_fullscreen;
                        if !is_fullscreen {
                            windowed_context.window().set_fullscreen(None);
                        } else {
                            let monitor = windowed_context.window().current_monitor();
                            let video_modes: Vec<_> = windowed_context.window().current_monitor().video_modes().collect();
                            let video_mode = (video_modes.len() - 1).min(video_mode);
                            println!("Exclusive: {:?}", modifiers.alt);
                            windowed_context.window().set_fullscreen(
                                Some(match modifiers.alt {
                                    true => {
                                        println!(
                                            "Chosen video mode: {}",
                                            video_modes.iter().nth(video_mode).unwrap()
                                        );
                                        Fullscreen::Exclusive(video_modes.iter().nth(video_mode).unwrap().clone())
                                    }
                                    false => Fullscreen::Borderless(monitor),
                                })
                            )
                        }
                    }
                    (VirtualKeyCode::Left, ElementState::Pressed)
                        | (VirtualKeyCode::Right, ElementState::Pressed) => {
                        let video_modes: Vec<_> = windowed_context.window().current_monitor().video_modes().collect();
                        video_mode = match key {
                            VirtualKeyCode::Left => video_mode.saturating_sub(1),
                            VirtualKeyCode::Right => (video_modes.len() - 1).min(video_mode + 1),
                            _ => unreachable!(),
                        };
                        println!(
                            "Picking video mode: {}",
                            video_modes.iter().nth(video_mode).unwrap()
                        );
                    }
                    (VirtualKeyCode::S, ElementState::Pressed) => {
                        println!("windowed_context.window().fullscreen {:?}", windowed_context.window().fullscreen());
                    }
                    (VirtualKeyCode::M, ElementState::Pressed) => {
                        is_maximized = !is_maximized;
                        windowed_context.window().set_maximized(is_maximized);
                    }
                    (VirtualKeyCode::D, ElementState::Pressed) => {
                        decorations = !decorations;
                        windowed_context.window().set_decorations(decorations);
                    }
                    _ => (),
                },
                _ => (),
            },
            _ => {}
        }
    });
}

// Enumerate monitors and prompt user to choose one
fn prompt_for_monitor(el: &EventLoop<()>) -> MonitorHandle {
    for (num, monitor) in el.available_monitors().enumerate() {
        println!("Monitor #{}: {:?}", num, monitor.name());
    }

    print!("Please write the number of the monitor to use: ");
    stdout().flush().unwrap();

    let mut num = String::new();
    stdin().read_line(&mut num).unwrap();
    let num = num.trim().parse().ok().expect("Please enter a number");
    let monitor = el
        .available_monitors()
        .nth(num)
        .expect("Please enter a valid ID");

    println!("Using {:?}", monitor.name());

    monitor
}

fn prompt_for_video_mode(monitor: &MonitorHandle) -> VideoMode {
    for (i, video_mode) in monitor.video_modes().enumerate() {
        println!("Video mode #{}: {}", i, video_mode);
    }

    print!("Please write the number of the video mode to use: ");
    stdout().flush().unwrap();

    let mut num = String::new();
    stdin().read_line(&mut num).unwrap();
    let num = num.trim().parse().ok().expect("Please enter a number");
    let video_mode = monitor
        .video_modes()
        .nth(num)
        .expect("Please enter a valid ID");

    println!("Using {}", video_mode);

    video_mode
}
```
</details>

At the start sometimes it landed on the wrong window. Not sure if this is caused by this PR or #978. :/

examples/multithreaded.rs
ZeGentzy: There is no additional tracking really required. Look at my glutin example. Just do this call every time before it is needed. Before calling fullscreen print out the one which is actually chosen.

Yes, that means the one that gets picked in the end might be slightly different then the one they wanted if they move screens after selecting, but that strikes me as better then their window jumping to a completely different window.

I don't want to be debugging only to have the window fly to my main screen, which I like to put my code on.

If that really concerns you, we can on every `WindowEvent::Moved` event check if the current window has changed, and if it has just update `video_modes` and print out the new video mode. That would be the best solution IMHO.

ZeGentzy:
This seams broken in weston, I typed 2 then 0 and nothing happens:
```
Please choose the fullscreen mode: (1) exclusive, (2) borderless: Monitor #0: Some("none (weston-X11)")
Please write the number of the monitor to use: Using Some("none (weston-X11)")
OpenGL version 4.5 (Core Profile) Mesa 19.1.0
dbg_callback 0x8248 0x8251 0x826B "Shader Stats: SGPRS: 16 VGPRS: 8 Code Size: 84 LDS: 0 Scratch: 0 Max Waves: 10 Spilled SGPRs: 0 Spilled VGPRs: 0 PrivMem VGPRs: 0"
dbg_callback 0x8248 0x8251 0x826B "Shader Stats: SGPRS: 8 VGPRS: 24 Code Size: 52 LDS: 0 Scratch: 0 Max Waves: 10 Spilled SGPRs: 0 Spilled VGPRs: 0 PrivMem VGPRs: 0"
dbg_callback 0x8248 0x8251 0x826B "Shader Stats: SGPRS: 16 VGPRS: 8 Code Size: 72 LDS: 0 Scratch: 0 Max Waves: 10 Spilled SGPRs: 0 Spilled VGPRs: 0 PrivMem VGPRs: 0"
dbg_callback 0x8248 0x8251 0x826B "LLVM diagnostic (remark): <unknown>:0:0: 3 instructions in function"
dbg_callback 0x8248 0x8251 0x826B "LLVM diagnostic (remark): <unknown>:0:0: 4 instructions in function"
dbg_callback 0x8248 0x8251 0x826B "Shader Stats: SGPRS: 8 VGPRS: 24 Code Size: 40 LDS: 0 Scratch: 0 Max Waves: 10 Spilled SGPRs: 0 Spilled VGPRs: 0 PrivMem VGPRs: 0"
dbg_callback 0x8248 0x8251 0x826B "Shader Stats: SGPRS: 8 VGPRS: 8 Code Size: 84 LDS: 0 Scratch: 0 Max Waves: 10 Spilled SGPRs: 0 Spilled VGPRs: 0 PrivMem VGPRs: 0"
```

examples/fullscreen.rs
```diff
@@ -1,56 +1,35 @@
...
+    stdout().flush().unwrap();
 
-        #[cfg(not(target_os = "macos"))]
-        Some(prompt_for_monitor(&event_loop))
-    };
+    let mut num = String::new();
+    stdin().read_line(&mut num).unwrap();
+    let num = num.trim().parse().ok().expect("Please enter a number");
+
+    let fullscreen = Some(match num {
```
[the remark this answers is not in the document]
aleksijuvani: `fullscreen` is an option because an `Option<Fullscreen>` is what we have to pass into `with_fullscreen` and `set_fullscreen`. It doesn't really matter if we make it into `Some` here or at the call sites.

examples/fullscreen.rs
```diff
@@ -67,35 +46,14 @@ fn main() {
...
-                            }
-                        }
-
-                        is_fullscreen = !is_fullscreen;
-                        if !is_fullscreen {
+                        if window.fullscreen().is_some() {
                             window.set_fullscreen(None);
                         } else {
-                            window.set_fullscreen(Some(window.current_monitor()));
+                            window.set_fullscreen(fullscreen.clone());
```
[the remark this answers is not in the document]
aleksijuvani: If we want to use the current monitor here, I think we ought to re-query the video mode from the user as well. I would prefer for this to use whatever values were input by the user on start up, so that we don't have to worry about the monitor changing here.

examples/fullscreen.rs
```diff
@@ -67,35 +46,14 @@ fn main() {
...
-                                if WindowExtMacOS::set_simple_fullscreen(&window, !is_fullscreen) {
-                                    is_fullscreen = !is_fullscreen;
-                                }
-                                return;
-                            }
-                        }
-
-                        is_fullscreen = !is_fullscreen;
-                        if !is_fullscreen {
+                        if window.fullscreen().is_some() {
```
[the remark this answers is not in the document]
aleksijuvani: This is not always true. This toggles fullscreen based on the window's current fullscreen state.

aleksijuvani:
I'm getting a protocol error upon running the fullscreen example on Weston in the `master` branch: `zxdg_shell_v6@13: xdg_surface buffer does not match the configured state`. It's not caused by the changes in this pull request.

aleksijuvani:
> At the start sometimes it landed on the wrong window. Not sure if this is caused by this PR or #978. :/

I haven't touched the window positioning code, so this is most likely the same issue.

src/monitor.rs
Osspial: Looks like it's sorting by Monitor->Size->Refresh Rate, which seems like the natural way to sort video modes (thinking in terms of how you'd sort video modes in a settings list). I personally don't feel too strongly about requiring docs here, since this is the most intuitive way to sort the video modes anyway, but I'm inclined leave that detail up to @aleksijuvani.

Osspial:
> 
> 
> I'm getting a protocol error upon running the fullscreen example on Weston in the `master` branch: `zxdg_shell_v6@13: xdg_surface buffer does not match the configured state`. It's not caused by the changes in this pull request.

My guess is that this should get fixed by https://github.com/rust-windowing/winit/pull/981.

examples/fullscreen.rs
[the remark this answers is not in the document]
ZeGentzy: I was too groggy this morning. Some reason I read this as `fullscreen.is_some()`, instead of `window.fullscreen().is_some()`.

examples/fullscreen.rs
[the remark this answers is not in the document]
ZeGentzy: Yikes, that's a silly complaint from me. [I misread this code](https://github.com/rust-windowing/winit/pull/925/files/b6eb841ff094efa3a85a9dfea4d4b114b9b05bcd#r298030273), it should be fine actually.

src/monitor.rs
ZeGentzy: Ah, that makes sense. Was tired this morning, spent five minutes staring at it and couldn't figure out what it did. Yeah, I don't think this needs any docs.

vberger:
This is pretty likely that #981 will solve the protocol error indeed.

src/platform_impl/ios/monitor.rs
```diff
@@ -1,18 +1,44 @@
...
-    monitor::VideoMode,
+    monitor::{MonitorHandle as RootMonitorHandle, VideoMode as RootVideoMode},
+    platform_impl::platform::ffi::{id, nil, CGFloat, CGRect, CGSize, NSInteger, NSUInteger},
 };
 
-use crate::platform_impl::platform::ffi::{
-    id, nil, CGFloat, CGRect, CGSize, NSInteger, NSUInteger,
-};
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct VideoMode {
```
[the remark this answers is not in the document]
aleksijuvani: I don't think it's the right solution to move these fields to the `RootVideoMode`. If we look at how the rest of the rest of the library is architectured, the `RootMonitorHandle` and `RootWindow` only contain a single field, which is the platform-specific implementation. This makes a lot of sense to me, and it'd be a shame to add fields to the "root" types that may or may not be applicable to all platforms. It's boilerplate, yes, but this makes future changes significantly easier in the platform-specific backends.

In fact, we don't really need to the `size`, `bit_depth` and `refresh_rate` fields at all on most platforms, since we already have the native mode handle that contains this information. I'd be inclined to modify the functions to fetch the values from the native mode handle (i.e. `DEVMODEW`, `CGDisplayModeRef`, etc.) instead. I think Wayland is the only one that actually requires us to store the values because of its callback-oriented nature (and incidentally X11, because changing this would require a significant refactor, as we need to store the `XRRScreenResources` instead of re-querying it on every call - although I think this refactor is inevitable anyway as the current `MonitorHandle` returns out of date information if the display configuration changes, but I digress).

examples/fullscreen.rs
```diff
@@ -1,56 +1,35 @@
...
-        Some(prompt_for_monitor(&event_loop))
-    };
+    let mut num = String::new();
+    stdin().read_line(&mut num).unwrap();
+    let num = num.trim().parse().ok().expect("Please enter a number");
+
+    let fullscreen = Some(match num {
+        1 => Fullscreen::Exclusive(prompt_for_video_mode(&prompt_for_monitor(&event_loop))),
+        2 => Fullscreen::Borderless(prompt_for_monitor(&event_loop)),
+        _ => unreachable!("Please enter a valid number"),
```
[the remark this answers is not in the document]
aleksijuvani: Done!

src/platform_impl/linux/x11/util/randr.rs
```diff
@@ -132,6 +94,10 @@ impl XConnection {
                     size: (x.width, x.height),
                     refresh_rate: (refresh_rate as f32 / 1000.0).round() as u16,
                     bit_depth: bit_depth as u16,
+                    native_mode: x.id,
+                    // This is populated in `MonitorHandle::video_modes` as the
+                    // video mode is returned to the user
+                    monitor: None,
```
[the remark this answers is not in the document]
aleksijuvani: This field is only visible internally to the crate, and we `unwrap()` the value before returning it to the user in `VideoMode::monitor()`. As I mentioned earlier, this ought to be refactored away in the future anyway, but I think that's outside the scope of this pull request.

examples/multithreaded.rs
aleksijuvani: Done!

aleksijuvani:
I will be leaving on a two week long trip in a few days and will be unable to make any further changes for quite some time, so I would like to get this merged before I leave. Do we have a consensus on the remaining review notes, or do you feel strongly that these ought to be changed? I will change these if that's what it takes to get this merged, but I still feel like long-term maintainability would suffer from these changes.

Osspial:
I'm still running into the *window in wrong position* problem in my latest tests ([see video here](https://gfycat.com/hardtofindredgannet)), so I suspect my initial diagnosis for that wasn't the correct one. If you're unable to reproduce this, I could look into it on my machine, then just push directly to the branch and merge it once I've figured out what's going on.

aleksijuvani:
I'm not able to reproduce this on my system. I suspect that may have to do with the virtual layout in which your monitors are configured. Thank you for offering to look into this! It would be helpful. Feel free to commit directly to this branch. I also have an opportunity tomorrow to make changes, and maybe a few hours on Friday, but then I'll be unreachable until the end of July or so.

Osspial:
Alright, this turned out to be one hell of a bug to track down.

It looks like the underlying cause was that sometimes, the call to `ChangeDisplaySettingsExW` would hang for such a long time that Windows would erroneously think our program had frozen, taking away control over the Winit window*. When that function finally returned, we'd still lack control over our window, causing `SetWindowPos` to fail and the window not getting moved to the proper position.

If you're interested in reproducing this bug, just apply this patch, run the `fullscreen` example in exclusive mode, and furiously click on the window before it properly gets fullscreened:

```patch
diff --git a/src/platform_impl/windows/window.rs b/src/platform_impl/windows/window.rs
index 756fc5ad..b8362e1f 100644
--- a/src/platform_impl/windows/window.rs
+++ b/src/platform_impl/windows/window.rs
@@ -480,6 +480,7 @@ impl Window {
                     let mut native_video_mode = video_mode.video_mode.native_video_mode.clone();

                     let res = unsafe {
+                        std::thread::sleep_ms(6000);
                         winuser::ChangeDisplaySettingsExW(
                             display_name.as_ptr(),
                             &mut native_video_mode,
```

The furious clicking is necessary to make Windows detect a freeze in our program. I'm not sure what steps you need to take to reproduce this without that patch, since this seems to be caused by various timing flukes outside of our control, but after inspecting my [ETW logs](https://randomascii.wordpress.com/2015/09/24/etw-central/) I'm confident that this is the correct diagnosis.

Fortunately, the solution is pretty simple - just insert a call to `PeekMessageW` after `ChangeDisplaySettingsExW` to inform Windows that our program is still running properly. I've pushed that change to this branch.

<sup>* This is the same infrastructure that causes the window to gray out and `(Not Responding)` to appear in the title bar. Funnily enough, `windbg` disables that behavior in debugged applications, making this bug impossible to reproduce when running in a debugger. :/</sup>

aleksijuvani:
Thank you for tracking this down! I just got back home from my trip last evening. It's interesting that `ChangeDisplaySettingsExW` would take such a long time to complete. This didn't occur on my system. I wonder if that is dependent on the display and drivers being used.

Osspial:
> It's interesting that `ChangeDisplaySettingsExW` would take such a long time to complete. This didn't occur on my system. I wonder if that is dependent on the display and drivers being used.

@aleksijuvani I did a bit more investigation, and I don't think I was entirely correct with my original assessment! `ChangeDisplaySettingsExW` takes some time to execute, but that only makes the problem more visible instead of directly causing it - the root cause is that we ask for the monitor between creating the `EventLoop` and creating the `Window`. If the user takes a while to enter their desired monitor configuration, the program freeze handler kicks in and the `SetWindowPos` call doesn't work. Incidentally, that also means it's possible to reproduce this bug on master without exclusive fullscreen - just wait five or more seconds before entering the desired monitor configuration

That's why I was able to reproduce it, but you weren't - whenever the display selection interaction happened, I took a good amount of time to select the correct display and configuration, which took long enough that Windows thought the program had frozen. I'm assuming you didn't do that, so you never saw the bug.

aleksijuvani:
I'm still not able to reproduce this even after waiting quite a bit before applying the desired display configuration. Which Windows version are you running on? I'm on Windows 10 version 1809. It seems odd to me that Windows would outright ignore API calls if it thinks that the window is frozen. Surely it should continue to allow the window to be manipulated?

aleksijuvani:
I neglected to do this before, but I've just now also tested reproducing this by inserting the `std::thread::sleep_ms` call, and this does allow me to reproduce the issue. Either way, introducing the `PeekMessageW` call seems to fix the issue. Shall we go ahead with merging this?

Osspial:
@aleksijuvani I've merged this now. Thank you so much for implementing this everywhere!

PR #922, opened 2019-06-14
Fix thread-safety of `set_maximized` and `set_title` on macOS

aleksijuvani:
Fixes #905. See https://github.com/rust-windowing/winit/issues/905#issuecomment-501990074. I'm not sure if this should be added to the changelog, or if this is specific to the `eventloop-2.0` changes, seeing as how the "multithreaded" example does not exist in the previous branch.

- [x] Tested on all platforms changed
- [ ] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users
- [ ] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior
- [ ] Created an example program if it would help users understand this functionality
- [ ] Updated [feature matrix](https://github.com/tomaka/winit/blob/master/FEATURES.md), if new features were added or implemented

PR #929, opened 2019-06-17
Fix warnings

felixrabe:
As mentioned on https://github.com/rust-windowing/winit/pull/922#issuecomment-502852085. This should probably not be merged as-is.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -86,8 +86,6 @@ extern crate log;
 #[macro_use]
 extern crate serde;
 #[macro_use]
-extern crate derivative;
-#[macro_use]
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -86,8 +86,6 @@ extern crate log;
 #[macro_use]
 extern crate serde;
 #[macro_use]
-extern crate derivative;
```
Osspial: This crate should be `cfg`-ed to Windows, instead of removed.

felixrabe: Brought it back with a cfg per your request.

PR #933, opened 2019-06-17
CONTRIBUTING.md: Linkify project members

felixrabe:


ZeGentzy:
I'd merge this, but it won't let me till the tests pass :/

ZeGentzy:
Why is this CI not done?

@Osspial think of all the trees that are dieing to this unskippable CI!

Osspial:
@ZeGentzy Eh, trees are overrated anyway. It's not like they've done anything for us.

Osspial:
Thinking about it, we should move that table into the wiki. That'll make it easier for people to add themselves as testers, so they don't have to go through a full PR process (which seems overblown for that purpose).

felixrabe:
Ah, you were faster than me: https://github.com/rust-windowing/winit/wiki/Testers-and-Contributors :)

PR #945, opened 2019-06-20
Implement fix described in #851

### LINKED ISSUE #851: Transparent and borderless window on Windows 10
Both `WindowBuilder::with_transparency(true)` and `WindowBuilder::with_decorations(false)` are working well by themselves on Windows 10, but their interaction leads to "invisible" unresponsive window (it "screenshots" image that was under window before it appeared).

Culprit is `WS_EX_LAYERED` flag [here](https://github.com/rust-windowing/winit/blob/master/src/platform/windows/window_state.rs#L201). I've removed it and window appeared like it should. But i'm not sure if it's proper fix for all versions of Windows.

Osspial:

Closes #851


ZeGentzy:
Finally. :heart:

Issue #963, opened 2019-06-22
Wrong XRandR check regarding 0mm display size


Abendstolz:
In [x11/util/randr.rs ](https://github.com/rust-windowing/winit/blob/master/src/platform_impl/linux/x11/util/randr.rs#L25) there is a bug as the check should read:

`if width_mm == 0 || height_mm == 0`

All credits go to https://github.com/rust-lang/rust-clippy

ZeGentzy:
Duplicate: #769

Issue #961, opened 2019-06-22
Forward porting from winit-legacy (tracking issue)


felixrabe:
(Edit: Working on this in PR #966.)

I just looked at the Git history more closely and noticed that some (all?) of the changes on the [winit-legacy](https://github.com/rust-windowing/winit/tree/winit-legacy) branch have not been ported to EL2.

My [forward-porting](https://github.com/felixrabe/winit/tree/forward-porting) branch contains some of my commits that I've noticed are lacking on current (EL2) master. If I get to it, I might expand it to more forward-portable stuff.

ZeGentzy:
Yeah, some changes probably slipped through the cracks. Open a PR with the cherrypicked changes, not a tracking issue. There's hardly anything to track: either it's merged, or it's not (which in that case, merge it).

felixrabe:
Can we keep this open until the PR closes this issue?

Issue #769, opened 2019-01-22
xrandr reports width_mm is zero bug


dannyfritz:
I'm pretty sure this is a bug: https://github.com/tomaka/winit/blob/3c59283b3fbf409353413c0c0d7d39ae85287e60/src/platform/linux/x11/util/randr.rs#L26

But I'm unsure if the answer is `width_px` or `height_mm`. So that is why this is an issue and not a PR. Easy fix regardless. :)

Osspial:
Hey, thanks for reporting this! Sorry it took so long to get back on it.

You'd want to change that to `height_mm`. You're right that that's a fairly simple fix, though.

ZeGentzy:
Needs to be fixed on evlpv2 branch :/

ZeGentzy:
Fix monitor width & height sanity check #861

PR #964, opened 2019-06-22
[#963] Change XRandR display size check to also take height into account

Abendstolz:
I won't tick any checkboxes as this is a bug, not a feature, so there is no documentation involved. 
Also it only impacts X11 and the fix is trivial.

### DIFF
--- a/src/platform_impl/linux/x11/util/randr.rs
+++ b/src/platform_impl/linux/x11/util/randr.rs
@@ -22,7 +22,7 @@ pub fn calc_dpi_factor(
     }
 
     // See http://xpra.org/trac/ticket/728 for more information.
-    if width_mm == 0 || width_mm == 0 {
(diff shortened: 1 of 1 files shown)


felixrabe:
Sorry for the double-approve — GitHub is feeling no pressure today to indicate a clicked button has indeed been clicked (so I click it again) until after the page reloads (and the non-idempotent whatever happened twice 👎 ).

vberger:
Why am I pretty sure that this had already been fixed? Maybe something that slipped through the backports...

ZeGentzy:
Yeah, @vberger, this didn't get backported, which is why I reopened #769. It's PR even had a `Needs backporting to event loop v2` tag, which got deleted by someone.

felixrabe:
@vberger I'm working through missing forward-ports right now, see #961.

This basically re-fixes PR #861, originally merged as 594cd18567.

Issue #955, opened 2019-06-21
Capturing mouse button clicks on Mac?


swiftcoder:
I'm trying to implement a traditional 3D camera on Mac, by grabbing and hiding the cursor, and  using `DeviceEvent::MouseMotion` to track movement. However, I can't figure out a way to capture mouse button clicks in this scenario.

`DeviceEvent::Button` does not fire at all for mouse button clicks on Mac. Is this the expected behaviour?

`WindowEvent::MouseInput` does fire, but only if the mouse was inside the area of the window before capturing the cursor. And if the mouse was outside the window, capturing the cursor doesn't stop the click from falling through to whatever application is in the background.

Any advice on making this work? Or guidelines as to the expected behaviour so that I can patch winit?

swiftcoder:
As a workaround, moving the cursor to an arbitrary position inside the window seems to work:
```
window.set_cursor_position((100, 100).into()).unwrap();
window.hide_cursor(true);
window.grab_cursor(true).unwrap();
```
With the downside that the cursor is now somewhere unexpected to the user when the program exits.

vbogaevsky:
@swiftcoder have you tried it on other platforms?

Osspial:
It looks like `DeviceEvent::Button` isn't implemented on macOS. From searching the code it looks like it should be implemented in [this function](https://github.com/rust-windowing/winit/blob/master/src/platform_impl/macos/app.rs#L61-L106), and I don't know why it hasn't been implemented yet. Should be a pretty easy PR to make, though!

swiftcoder:
Looking through the code, I don't think `DeviceEvent::Button` is implemented on Windows either?

Osspial:
[It's implemented here](https://github.com/rust-windowing/winit/blob/master/src/platform_impl/windows/event_loop.rs#L1362-L1365). For some reason, a lot of the Windows events are imported by the event's variant, instead of used through the event - that's why it was hard to find this.

swiftcoder:
I've added logging to the cursor_grab example in #967. Unfortunately it highlights that event intercepting at the `[NSApplication sendEvent:]` level doesn't capture mouse clicks outside of the window.

I think the correct fix here is to automatically move the cursor to an arbitrary position inside the window when `set_cursor_grab(true)` is called, and reset to the original position when `set_cursor_grab(false)` is called (or the program exits)?

markhildreth:
[...]

PR #967, opened 2019-06-23
Implement DeviceEvent::Button on Mac

swiftcoder:
- [ ] Updated [feature matrix](https://github.com/tomaka/winit/blob/master/FEATURES.md), if new features were added or implemented

Addresses #955.

This should do the trick, but I don't have a solid test case handy, since my application is still on 0.19.1 (via glium/glutin), and this code has all been refactored since that version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- On Mac, implement `DeviceEvent::Button`.
 - Change `Event::Suspended(true / false)` to `Event::Suspended` and `Event::Resumed`.
 - On X11, fix sanity check which checks that a monitor's reported width and height (in millimeters) are non-zero when calculating the DPI factor.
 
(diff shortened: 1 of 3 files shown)


felixrabe:
I'll give it a try later.

felixrabe:
Ok, tested this with commit 9dd15d00d812 using your example. Seems to work as expected:

- Without your changes: DeviceEvent::Button does not get reported.
- With your changes: DeviceEvent::Button does get reported.

Didn't see any regressions, so looks fine from me.

felixrabe:
I've [rebased against master, example commit first](/felixrabe/winit/tree/device-button-test-without-changes), so you can force-push this PR on that and all is smooth. (Yeah, I was bored and had nothing else to do.) :)

Osspial:
Merging. Thanks for the PR!

Issue #936, opened 2019-06-19
Invisible window takes focus from existing window


chichid:
Consider this code, where I'm creating an invisible window, after a visible window. Should the first window appear disabled (the second window takes focus from it)? 

``
let el = winit::EventsLoop::new();
let win = winit::WindowBuilder::new().with_visibility(true).build(&el).unwrap();
let win = winit::WindowBuilder::new().with_visibility(false).build(&el).unwrap();    
``

I tried this on Windows 10, and it results in a disabled window. I would expect an invisible window not to take focus from any other windows.

Osspial:
Hi, and thanks for the bug report!

I can confirm that this is undesirable behavior. My best guess as to what's causing it is one of the direct or indirect calls to [`set_window_flags`](https://github.com/rust-windowing/winit/blob/master/src/platform_impl/windows/window_state.rs#L116-L128) in the [this part of the window `init` function](https://github.com/rust-windowing/winit/blob/06244dd4929d1acf1f4650ecb38ca73c1ec22743/src/platform_impl/windows/window.rs#L722-L754). I'd be up for tackling this at some point, but this should also be a good first issue to investigate if anyone would be interested.

chichid:
I'm interested, and already investigating it, I haven't done Win32 in ages though.

chichid:
PR created

PR #968, opened 2019-06-23
Fix: Invisible windows steal focus from visible windows Win32

chichid:
Fixes: https://github.com/rust-windowing/winit/issues/936
Updated the multiwindow example adding an invisible window

- [x] Tested on all platforms changed (Windows only)
- [x] `cargo fmt` has been run on this branch (Tried running it but it modifies a lot of files...)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - Change `Event::Suspended(true / false)` to `Event::Suspended` and `Event::Resumed`.
 - On X11, fix sanity check which checks that a monitor's reported width and height (in millimeters) are non-zero when calculating the DPI factor.
+- On Windows, when a window is initially invisible, it won't take focus from the existing visible windows.
 
 # 0.20.0 Alpha 1
 
--- a/src/platform_impl/windows/window.rs
+++ b/src/platform_impl/windows/window.rs
@@ -271,7 +271,8 @@ impl Window {
                 winuser::SWP_ASYNCWINDOWPOS
                     | winuser::SWP_NOZORDER
                     | winuser::SWP_NOREPOSITION
-                    | winuser::SWP_NOMOVE,
+                    | winuser::SWP_NOMOVE
+                    | winuser::SWP_NOACTIVATE,
             );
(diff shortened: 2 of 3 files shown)


felixrabe:
I see no changes to docs in this diff, despite "updated documentation". Would it make sense to add a mention about the new behavior to the docs?

examples/multiwindow.rs
```diff
@@ -14,6 +14,12 @@ fn main() {
         windows.insert(window.id(), window);
     }
 
+    let invisible_window = WindowBuilder::new()
+        .with_visible(false)
+        .build(&event_loop)
+        .unwrap();
+    windows.insert(invisible_window.id(), invisible_window);
+
```
Osspial: The changes made in this example shouldn't get merged onto `master`.

Osspial:
@felixrabe wrt documentation - I don't think that's strictly necessary, since this seems like the expected behavior to begin with. Still, I wouldn't deny docs spelling this out explicitly.

examples/multiwindow.rs
chichid: I thought they could be useful to showcase the feature, but I reverted that! YW, the feature will be important for me in one of my projects!

felixrabe: TIL, YW is YAAIDK (yet another abbreviation I don't know). :)

chichid: So is TIL! :)

Osspial: The `cargo fmt` check is failing. Go ahead and rebase against master - we were relying on nightly rustfmt settings beforehand, but we've updated our config to get rid of those flags and only use stable settings. Run `cargo fmt` once this is rebased and Travis should work.

Sorry about the confusion!

felixrabe: @chichid (why does your name not get expanded?) TIL is "things / today I learned". So what is "YW"?

chichid: YW = You're welcome

chichid: @Osspial No Problem, I will rebase then, once I'm back home! :)

PR #973, opened 2019-06-23
Wayland: Add relative pointer movement

### LINKED ISSUE #962: Add pointer grab for Wayland
pointer_constraints_v1 extension has been around for almost a year. It is already supported in wlroots.

Cherser-s:
Partially resolves #962. Added support for zwp-relative-pointer and added DeviceEventsSink in addition to existing event queue.
- [ ] Added pointer constraints support

### DIFF
--- a/src/platform_impl/linux/wayland/event_loop.rs
+++ b/src/platform_impl/linux/wayland/event_loop.rs
@@ -31,64 +31,37 @@ use smithay_client_toolkit::{
     Environment,
 };
 
-pub struct WindowEventsSink {
-    buffer: VecDeque<(crate::event::WindowEvent, crate::window::WindowId)>,
+pub struct WindowEventsSink<T> {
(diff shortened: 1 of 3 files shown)


src/platform_impl/linux/wayland/event_loop.rs
```diff
@@ -492,7 +519,20 @@ impl SeatData {
...
+                        self.relative_pointer_manager_proxy.as_ref().map_or_else(
+                            || None,
+                            |manager| {
+                                super::pointer::implement_relative_pointer(
+                                    self.sink.clone(),
+                                    self.pointer.as_ref().unwrap(),
+                                    manager,
+                                )
+                                .ok()
+                            },
```
vberger: Can't this be simplified using `Option::and_then` rather than `map_or_else`?

Cherser-s: Yes, next commit fixes it.

Issue #855, opened 2019-04-26
Request for slightly better event overview documentation


AlterionX:
The current docs leave something to be desired in differentiating the many different types of `Event`s in the system.

There are two primary inputs (`DeviceEvent` and `WindowEvent`) but the there's no higher level documentation to differentiate the two. If you go into the docs and read each individual class's documentation it becomes rather obvious, but it's confusing at first, so I think it should get (at least briefly) mentioned when explaining the event loop.

Additionally, it's slightly confusing that certain events are duplicated, as this phrase: "Once a window has been created, it will generate events. For example whenever the user moves the window, resizes the window, moves the mouse, etc. an event is generated." makes it seem like there is one and exactly one event being created where there are, in fact, multiple.

I ended up writing an abstraction for the entire input loop to filter out events like these, but it was a bit frustrating working with the thing before I got used to it.

alasher:
@AlterionX Check out my pull request and see if it's close to what you were thinking. I'm new to the project so feel free to clarify anything that may be misleading or incorrect. Thanks!

AlterionX:
@alasher Someone else got there before I did! But what they said is what I'm thinking as well.

A mouse movement or keyboard press is also duplicated through the `WindowEvent` interface through this: https://docs.rs/winit/0.20.0-alpha1/winit/event/enum.WindowEvent.html#variant.KeyboardInput or other variants of `WindowEvent`.

AlterionX:
Merged with #976

PR #976, opened 2019-06-24
Update docs to differentiate DeviceEvents and WindowEvents

alasher:
New to this project, thought I'd try to address #855. All input is appreciated! Many of the below regulations do not apply for this docfix, but I did generate the documentation with `cargo doc` and verified the links work appropriately. I changed the `WindowEvent` and `DeviceEvent` links to their dedicated docs page, instead of the sublink on the generic `Event` page. This way when the user clicks they can see all the different types of Window/Device, which I think makes more sense in the context they are linked in.


Osspial:
Hi, and congratulations on the first PR!

There's one clarification I'd like to make the docs to make, but otherwise these seem like great changes. Right now, the patched documentation for `DeviceEvent` implies that mouse and keyboard events will *only* be dispatched through `DeviceEvent`. Instead, it'd be more accurate to say that *unfiltered* mouse and keyboard events get dispatched through `DeviceEvent`, as users will receive `WindowEvent`s for mouse and keyboard input (as filtered by the window manager) if the window is focused.

AlterionX:
I had commented on the relevant issue, but an example of `WindowEvent` being used for keyboard input is here: https://docs.rs/winit/0.20.0-alpha1/winit/event/enum.WindowEvent.html#variant.KeyboardInput

Just for reference.

alasher:
Thanks for the feedback! I've updated the description again. Hopefully it's not too wordy, but I think the text now accurately describes the key differences between DeviceEvents and WindowEvents. Feel free to nit-pick anything I wrote to make it more clear.

Osspial:
That looks good. Thank you for looking into this!

PR #977, opened 2019-06-24
Implement `_NET_WM_PING` protocol for X11

hoyon:
The spec for the protocol can be found [here](https://specifications.freedesktop.org/wm-spec/1.4/ar01s06.html) but essentially what it does is allow the desktop environment to send a ping message to a window when the user attempts to close the window. 

If the window neither responds nor closes in a reasonable amount of time, it is considered unresponsive and the user is given the option to kill the program which sends `SIGKILL` to the process.

I've tested this using gnome with X11.


ZeGentzy:
Oh, apparently I can do it my self. But now I got to wait for that damn CI for modifying the changelog....

:rage:

PR #980, opened 2019-06-25
Improve handling of file paths in the windows DnD handler

superfuchs:
This changes `FileDropHandler::iterate_filenames` to use the exact length returned by `DragQueryFileW` instead of relying on MAX_PATH.

I've additionally replaced all calls to `mem::uninitialized` with `mem::zeroed` (on windows only).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- On Windows, support paths longer than MAX_PATH (260 characters) in `WindowEvent::DroppedFile`
+and `WindowEvent::HoveredFile`.
 - On Mac, implement `DeviceEvent::Button`.
 - Change `Event::Suspended(true / false)` to `Event::Suspended` and `Event::Resumed`.
 - On X11, fix sanity check which checks that a monitor's reported width and height (in millimeters) are non-zero when calculating the DPI factor.
--- a/src/platform_impl/windows/drop_handler.rs
+++ b/src/platform_impl/windows/drop_handler.rs
@@ -200,15 +200,14 @@ impl FileDropHandler {
 
             for i in 0..item_count {
                 // Get the length of the string NOT including the terminating null character
-                let character_count =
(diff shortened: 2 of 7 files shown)


Osspial:
Hi, and thanks for the PR! Sorry about the delay in looking at this.

Couple points:
* I'm assuming you're making this because [it seems you can have files longer than MAX_PATH](https://docs.microsoft.com/en-us/windows/desktop/fileio/naming-a-file#maximum-path-length-limitation). Could you throw a comment in here documenting that reasoning for switching away?
* Could you add a CHANGELOG entry? This is a meaningful change that will affect Winit's behaviors in some cases, so we should document that the change was made.
* Why change from `uninitialized` to `zeroed`? All of the places I can see it being used get immediately overwritten by the Windows API, so it shouldn't be having any ill effects.

superfuchs:
Yes, exactly. If you want to test this, you have to first [enable](https://blogs.msdn.microsoft.com/jeremykuhne/2016/07/30/net-4-6-2-and-long-paths-on-windows-10/) long path support on Windows 10 and then use a file manager that supports it (explorer.exe doesn't), like [One Commander](http://onecommander.com/). Note also that this alone won't help you much, if you then want to e.g. read that file. See this [issue](https://github.com/BurntSushi/ripgrep/issues/364), for example.

I did the change from `uninitialized` to `zeroed` because `uninitialized` will be [deprecated ](https://github.com/rust-lang/rust/pull/60445#issue-275104154) in favor of `MaybeUninit`. We could wait for `MaybeUninit` to reach the stable branch instead, of course.

Added the CHANGELOG entry and the comment, as requested.

aleksijuvani:
> Why change from `uninitialized` to `zeroed`? All of the places I can see it being used get immediately overwritten by the Windows API, so it shouldn't be having any ill effects.

The Windows API generally expects structs to be zeroed rather than uninitialized, and this is mentioned in their API documentation. I remember having run into this before because I hadn't zeroed some struct that it expected to be zeroed. We should always use `zeroed` in this context rather than `uninitialized` or `MaybeUninit`.

PR #835, opened 2019-04-07
wayland: use an invisible surface as shell surface

vberger:
**Note:** This work is on the eventloop-2.0 branch

This decorelates the window management from the actual user content,
meaning:

- the created window no longer needs the user to draw something to
  start existing
- it reduces our need to do roundtrips during initialization to
  avoid protocol errors

depends on #833

PR #981, opened 2019-06-25
Revert "wayland: use an invisible surface as shell surface (#835)"

vberger:
This reverts commit 65587ef43a7d001ba3e1fa6c7ecd12bd47fffa84.

It introduced sublte bugs in its interaction with OpenGL and glutin,
so we should better revert it for now.


milkey-mouse:
#835 also caused problems with vulkano for me; namely, resizing the window/recreating the swapchain was really slow for some reason. Fortunately this PR fixes it for vulkan as well.

PR #982, opened 2019-06-25
Remove `XFlush` call in event loop

### LINKED ISSUE #865: EventLoop 2.0 swallowing events
So I've just ported Alacritty to the EventLoop 2.0 rework and in general it has been surprisingly painless. However it does seem like there was a regression which is (hopefully) not caused by errors during the porting process.

It seems like when using `run_return` to check for events, it might under some circumstances swallow certain events. This can be reproduced by printing continuous output (like `yes`) and trying to cancel it by pressing Ctrl+C (which is handled as a `ReceivedChar` event). This event is never received. It's possible to print all glutin events with the `--print-events` flag. The only received events are `EventsCleared` and `LoopDestroyed`.

These events are not always swallowed, but it appears only to happen when the event loop is called repeatedly without any events. Under normal circumstances when Alacritty is idle and the control flow is `Wait`, it seems like events are received properly (so the event loop is actively running, waiting for events).

The EventLoop 2.0 branch for Alacritty can be found here:
https://github.com/jwilm/alacritty/pull/2438

murarth:
Internally, `XFlush` calls `_XSend` to write data. It then calls
`XEventsQueued(display, QueuedAfterReading)`, which reads data from the
X server connection. This prevents the event loop source callback from
being run, as there is no longer data waiting on the socket.

Ideally, we would want to call `_XSend` directly to ensure that no
output is buffered by Xlib. However, this function is not exported as
part of Xlib's public API.

Testing with the `XFlush` call removed does not appear to adversely
affect the performance of an application. If any bugs should eventually
arise from this change, perhaps another function may be used in place of
`XFlush`, such as `XPending`, which writes buffered output but does not
so aggressively read from the X server connection.

Closes #865

### DIFF
--- a/src/platform_impl/linux/x11/mod.rs
+++ b/src/platform_impl/linux/x11/mod.rs
@@ -291,11 +291,6 @@ impl<T: 'static> EventLoop<T> {
                 );
             }
 
-            // flush the X11 connection
-            unsafe {
-                (wt.xconn.xlib.XFlush)(wt.xconn.display);
-            }
-
             match control_flow {
                 ControlFlow::Exit => break,
                 ControlFlow::Poll => {


Osspial:
Now that this has been merged, I think we should release an `alpha2` so X11 users can properly use the new event loop.

aloucks:
There is still another flush call:

https://github.com/rust-windowing/winit/blob/master/src/platform_impl/linux/x11/util/mod.rs#L82

murarth:
@aloucks That one appears to be called only when necessary. It doesn't seem to be interfering with the event loop.

PR #984, opened 2019-06-26
Fix broken links to mod DPI on various documentation pages

alasher:
Looking around in the documentation I saw several links to mod DPI were broken. Tested and confirmed these new links work. 👍🏻

Other tasks not relevant for this tiny docfix

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -198,7 +198,7 @@ pub enum WindowEvent {
     /// * Changing the display's DPI factor (e.g. in Control Panel on Windows).
     /// * Moving the window to a display with a different DPI factor.
     ///
-    /// For more information about DPI in general, see the [`dpi`](dpi/index.html) module.
+    /// For more information about DPI in general, see the [`dpi`](../dpi/index.html) module.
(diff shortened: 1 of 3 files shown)


Osspial:
Adding onto this - I recently found out that rustdoc lets you pass rust paths instead of URLs for intra-doc links (i.e. you can do `[this](crate::dpi)` and it'll link to the `dpi` module). We should probably switch to that link format at some point since it's significantly easier to catch errors with.

alasher:
@Osspial Nice! I like it, it's much less fragile than a relative link. I admittedly didn't do much research before issuing the fix, but I agree it'd be worth replacing all links with this format.

Osspial:
@alasher To be clear - I'm not disappointed in this PR! At various points I've explicitly searched for that functionality to no avail, but it's historically been really hard to find. The only way I found out about it was that rustdoc wasn't warning about links I wrote that I thought it should be complaining about.

felixrabe:
@Osspial The nice linking is a very recent improvement of rustdoc. I'm not sure whether it is even already stable.

Issue #806, opened 2019-03-08
Window starting position


chrisduerr:
I'm not sure about support of other platforms, but I believe at least XCB should support setting the starting position of a Window, so in theory it should be possible to have a `with_position` method on at least the X11 backend.

Most of the other functions relevant at startup have a matching `with_*` function call for their `set_*` calls, so I'm not sure if this is due to a lack of support for some platforms, however it seems there are some clear advantages for the platforms that support it since it allows spawning the window without things jumping around.

repi:
A `with_position` would indeed be nice. 

Currently we work around it by directly setting the position of the window with `window.set_position`, which works but is less elegant.

chrisduerr:
The big issue with `window.set_position` is that it will resize the window, no matter how early you do it. So unfortunately that creates this flickering effect where the window is first spawned at one size and then later changed to another.

Osspial:
This should work on all non-Wayland desktop platforms, so I'd be fine with this getting added.

Abendstolz:
As the API was changed lately ([commit](https://github.com/rust-windowing/winit/commit/0df436901a4835b47c1daf58a5260b839cf9dd6a#diff-2ce3f8d26915325bbd02de9e0056281eR416)) and there is only a `set_outer_position` I assume there should only be a `with_outer_position`. 

Is that assumption correct?

vberger:
> Another question: As by @Osspial's comment this doesn't work on a Wayland desktop - does or should set_outer_position? Because the "platform-specific" part of the docs there only mention iOS.

Wayland apps cannot control their location on the desktop, at all.

Abendstolz:
> > Another question: As by @Osspial's comment this doesn't work on a Wayland desktop - does or should set_outer_position? Because the "platform-specific" part of the docs there only mention iOS.
> 
> Wayland apps cannot control their location on the desktop, at all.

So should that be documented in `set_outer_position` as well or is it considered common knowledge for anyone targeting Wayland? (Which you might do without yourself knowing)

felixrabe:
It should be documented, because using winit should not require reading up on surprising implementation differences on different platforms. Everything that could cause someone opening an issue later on along the lines of "why does xyz not work as expected?" should be explicitly documented.

chrisduerr:
Usually better than documenting things is encoding it in the API. So if this doesn't work on Wayland, it just shouldn't be implemented for that platform.

vberger:
[...]

Issue #978, opened 2019-06-24
Fullscreen at startup inconsistent on X11


chrisduerr:
This has been reported to Alacritty in https://github.com/jwilm/alacritty/issues/2561.

On X11, it seems like setting the startup mode after the window is `show()`n on an originally invisible window has some inconsistent behavior. I've tested this myself on i3.

When the window is created and set to fullscreen on `window.get_current_monitor()` right at startup, the window is fullscreened on the incorrect monitor.

Additionally, if the position is set right after `show()` too, the window is fullscreened on the incorrect monitor and when leaving fullscreen mode the position is not what it has been set to.

If `set_fullscreen` is manually called on a keyboard event at a later point in time, both of these work without any issue.

chrisduerr:
Applying the following diff to glutin makes it possible to reproduce the issue:

<details>

```diff
diff --git a/glutin_examples/examples/fullscreen.rs b/glutin_examples/examples/fullscreen.rs
index 25cbc72..bdc03d6 100644
--- a/glutin_examples/examples/fullscreen.rs
+++ b/glutin_examples/examples/fullscreen.rs
@@ -9,52 +9,26 @@ use glutin::window::WindowBuilder;
 use std::io::Write;
 
 fn main() {
+    let mut is_fullscreen = true;
     let el = EventLoop::new();
 
     #[cfg(target_os = "macos")]
     let mut macos_use_simple_fullscreen = false;
-
-    let monitor = {
-        // On macOS there are two fullscreen modes "native" and "simple"
-        #[cfg(target_os = "macos")]
-        {
-            print!(
-                "Please choose the fullscreen mode: (1) native, (2) simple: "
-            );
-            std::io::stdout().flush().unwrap();
-
-            let mut num = String::new();
-            std::io::stdin().read_line(&mut num).unwrap();
-            let num = num.trim().parse().ok().expect("Please enter a number");
-            match num {
-                2 => macos_use_simple_fullscreen = true,
-                _ => {}
-            }
-
-            // Prompt for monitor when using native fullscreen
-            if !macos_use_simple_fullscreen {
-                Some(prompt_for_monitor(&el))
-            } else {
-                None
-            }
-        }
-
-        #[cfg(not(target_os = "macos"))]
-        Some(prompt_for_monitor(&el))
-    };
-
-    let mut is_fullscreen = monitor.is_some();
     let mut is_maximized = false;
     let mut decorations = true;
 
     let wb = WindowBuilder::new()
[...]

PR #989, opened 2019-06-26
X11: Remember position setting in fullscreen

murarth:
Window position that is assigned while a window is set to fullscreen
will be used as the window's position when fullscreen is disabled.

### DIFF
--- a/src/monitor.rs
+++ b/src/monitor.rs
@@ -13,7 +13,7 @@
 use std::collections::vec_deque::IntoIter as VecDequeIter;
 
 use crate::{
-    dpi::{PhysicalPosition, PhysicalSize},
+    dpi::{LogicalPosition, PhysicalPosition, PhysicalSize},
     platform_impl,
 };
 
@@ -115,6 +115,17 @@ impl MonitorHandle {
     #[inline]
     pub fn position(&self) -> PhysicalPosition {
         self.inner.position()
+    }
+
+    #[inline]
+    pub(crate) fn relative_position(&self, position: LogicalPosition) -> PhysicalPosition {
+        let phys = self.position();
+        let dpi = self.hidpi_factor();
+
+        PhysicalPosition{
+            x: position.x + phys.x * dpi,
+            y: position.y + phys.y * dpi,
+        }
     }
 
     /// Returns the DPI factor that can be used to map logical pixels to physical pixels, and vice versa.
--- a/src/monitor.rs
+++ b/src/monitor.rs
@@ -122,7 +122,7 @@ impl MonitorHandle {
         let phys = self.position();
         let dpi = self.hidpi_factor();
 
-        PhysicalPosition{
+        PhysicalPosition {
             x: position.x + phys.x * dpi,
             y: position.y + phys.y * dpi,
         }
--- a/src/platform_impl/linux/mod.rs
+++ b/src/platform_impl/linux/mod.rs
@@ -214,6 +214,14 @@ impl Window {
     }
 
     #[inline]
+    pub fn set_global_outer_position(&self, position: PhysicalPosition) {
+        match self {
+            &Window::X(ref w) => w.set_global_outer_position(position),
+            &Window::Wayland(ref w) => w.set_global_outer_position(position),
+        }
+    }
+
+    #[inline]
     pub fn inner_size(&self) -> LogicalSize {
         match self {
             &Window::X(ref w) => w.inner_size(),
--- a/src/platform_impl/linux/wayland/window.rs
+++ b/src/platform_impl/linux/wayland/window.rs
@@ -5,7 +5,7 @@ use std::{
 };
 
 use crate::{
-    dpi::{LogicalPosition, LogicalSize},
+    dpi::{LogicalPosition, LogicalSize, PhysicalPosition},
     error::{ExternalError, NotSupportedError, OsError as RootOsError},
     monitor::MonitorHandle as RootMonitorHandle,
     platform_impl::{
@@ -221,6 +221,11 @@ impl Window {
 
     #[inline]
     pub fn set_outer_position(&self, _pos: LogicalPosition) {
+        // Not possible with wayland
+    }
+
+    #[inline]
+    pub fn set_global_outer_position(&self, _pos: PhysicalPosition) {
         // Not possible with wayland
     }
 
--- a/src/platform_impl/linux/x11/monitor.rs
+++ b/src/platform_impl/linux/x11/monitor.rs
@@ -10,7 +10,7 @@ use super::{
     util, XConnection, XError,
 };
 use crate::{
-    dpi::{PhysicalPosition, PhysicalSize},
+    dpi::{LogicalPosition, PhysicalPosition, PhysicalSize},
     monitor::VideoMode,
 };
 
@@ -104,6 +104,17 @@ impl MonitorHandle {
     #[inline]
     pub fn hidpi_factor(&self) -> f64 {
         self.hidpi_factor
+    }
+
+    #[inline]
+    pub fn relative_position(&self, position: LogicalPosition) -> PhysicalPosition {
+        let phys = self.position();
+        let dpi = self.hidpi_factor();
+
+        PhysicalPosition{
+            x: position.x + phys.x * dpi,
+            y: position.y + phys.y * dpi,
+        }
     }
 
     #[inline]
--- a/src/platform_impl/linux/x11/monitor.rs
+++ b/src/platform_impl/linux/x11/monitor.rs
@@ -111,7 +111,7 @@ impl MonitorHandle {
         let phys = self.position();
         let dpi = self.hidpi_factor();
 
-        PhysicalPosition{
+        PhysicalPosition {
             x: position.x + phys.x * dpi,
             y: position.y + phys.y * dpi,
         }
--- a/src/platform_impl/linux/x11/util/hint.rs
+++ b/src/platform_impl/linux/x11/util/hint.rs
@@ -116,6 +116,21 @@ impl<'a> NormalHints<'a> {
         }
     }
 
+    pub fn get_position(&self) -> Option<(u32, u32)> {
+        self.getter(ffi::PPosition, &self.size_hints.x, &self.size_hints.y)
+    }
+
+    // WARNING: This hint is obsolete
+    pub fn set_position(&mut self, position: Option<(u32, u32)>) {
+        if let Some((x, y)) = position {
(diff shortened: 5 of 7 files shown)


murarth:
This solves one aspect of #978, that position setting was not honored if set while in fullscreen mode.

chrisduerr:
> This solves one aspect of #978, that position setting was not honored if set while in fullscreen mode.

Based on my testing, this does not resolve this specific issue.

murarth:
I've pushed another commit to add the method `with_outer_position` to `WindowBuilder`. It seems that WMs will ignore position information unless it is passed in this fashion at the time of window creation. Also, Xlib docs describe these values as "obsolete" and say that some WMs will completely ignore them, so I've added a note to the public method. However, testing on Gnome Classic, this hint was effective.

If this change is acceptable, I'll add it to the changelog. Of course, maintainers of other platforms will have to be notified to implement this feature. I'm not sure how that's usually done here.

chrisduerr:
While a `with_outer_position` method would be great and would resolve https://github.com/rust-windowing/winit/issues/806, I'm not sure how it's supposed to be used.

If I want to spawn a window at a specific logical position, that depends on the monitor it's spawned on. Is the DPI always calculated based on the current active monitor, or the monitor where it would land? If it depends on the active monitor, how is any application supposed to reason about it? If it depends on where it's going to land, how do we know which DPI to use for calculating the position where it's spawned to get the DPI for calculating the position where it will be spawned?

Also, does this mean the `set_outer_position` call will be ignored on X11 when it's done too quickly?

Also, based on my testing this still does not solve the original issue. It just changes which window the fullscreen window will be displayed on, but after leaving fullscreen mode it still goes directly back to 0x0 of the current monitor.

murarth:
If `with_outer_position` were to require a `MonitorHandle`, that would allow winit to translate the user-provided position to the true location on the X display, taking position, size, and DPI of the monitor into account. `set_outer_position` may have to assume a position on the current monitor. Perhaps another method could be added (called, e.g. `set_outer_position_on_monitor`), which would also accept a `MonitorHandle` for relative positioning.

>Also, does this mean the `set_outer_position` call will be ignored on X11 when it's done too quickly?

Possibly, yes. This is due to the asynchronous nature of the X protocol -- after the window is created, the WM will likely update the position and size of the new window (at minimum, to add space for window decorations). Any changes with an earlier timestamp than the WM update will be ignored by the X server.

>[...] after leaving fullscreen mode it still goes directly back to 0x0 of the current monitor.

Setting position and fullscreen on window creation (I think that's what you're describing here) is a bit tricky because X doesn't seem to track non-fullscreen position/size when a window is fullscreen. As far as X is concerned, fullscreen is the window's size. winit tracks a `restore_position` to apply when fullscreen mode is exited and this is set on window creation (in commit https://github.com/rust-windowing/winit/pull/989/commits/7dcad895887759cb416db4ee0418eddd73515ac8, added after my previous comment), but I may have missed an edge case somewhere.

murarth:
I've pushed a commit with the above mentioned change to `WindowBuilder::with_outer_position`. It also corrects `set_outer_position` by interpreting the input position as relative to the current monitor. I have not implemented the proposed `set_outer_position_on_monitor` method, but it would be straightforward to do so, if desired.

chrisduerr:
> If with_outer_position were to require a MonitorHandle, that would allow winit to translate the user-provided position to the true location on the X display, taking position, size, and DPI of the monitor into account.

Surer, this would work, but that would kinda defeat the point of setting the absolute position on a global, rather than a per-monitor level. If someone has it in their config that a window should always spawn 100x100 pixels away from the top-left screen of the top-left monitor, that couldn't be easily done with this PR.

One would basically have to get all available monitors, their hidpi factor, then calculate things manually themselves.

Also, if this doesn't work on Wayland at all and it isn't guaranteed to work on X11, I feel like the purpose of the original issue is a bit lost. That would still raise the question on how to set the position on the platforms where it doesn't work right after startup, which would mean that set_outer_position would have to be called again and we're back at square one.

I think the with_outer_position issues could probably be resolved on X11 by just taking in a physical position. Since that's likely what X11 expects anyways. If Wayland is not supported, most of the reason why everything needs to be done in logical positions wouldn't matter for the remaining platforms as far as I know.

On Wayland, hopefully the `set_outer_position` call should just work though, so doing it after startup there might suffice?

murarth:
>[...] setting the absolute position on a global, rather than a per-monitor level.

If that's what `set_outer_position` is intended to do, fair enough. Its behavior with respect to multiple screens isn't documented, so I took a guess based on the X11 implementation using the current monitor DPI to translate logical to physical position. winit could be more flexible, if desired, by providing two methods: one to set monitor-local `LogicalPosition` and another to set global `PhysicalPosition`.

It's true that `with_outer_position`/`set_outer_position` are not supported for every platform, but there are existing `Window` methods with documented limitations on specific platforms. It seems sensible to me to provide the functionality where available and allow application authors to decide what to do about limited availability of features.

murarth:
I've just pushed a commit implementing `set_global_outer_position`, as described above. `set_outer_position` now sets a position relative to the current monitor. I believe this needs to be the behavior on all platforms (if it isn't already) because otherwise it is fundamentally broken on X11. Using the current monitor's DPI factor to translate the input `LogicalPosition` to a `PhysicalPosition` in global screen space would produce unexpected results in an environment with multiple screens of varying DPI factor.

I haven't implemented `set_global_outer_position` on any platforms other than X11/Wayland. If I should add `unimplemented!()` methods for other platforms so that this will compile, let me know.

chrisduerr:
@murarth This still doesn't solve the issue with fullscreen at startup not preserving position, right? Do you have any idea why that might be happening or how we could fix it?

murarth:
@chrisduerr It works for me, using the `WindowBuilder` methods. Are you able to set position without fullscreen? You could be using one of the WMs that ignores hints.

chrisduerr:
Yes, I am able to set the position without fullscreen. It just doesn't work when done right at startup and fullscreen is set at startup too. If only position is set, or both are set at a later time, everything works.

murarth:
Huh. Well, I don't know what to say. It's working for me. Here's the code I'm using:

<details>

```rust
use std::io::{self, Write};

use winit::{
    event::{ElementState, Event, KeyboardInput, VirtualKeyCode, WindowEvent},
    monitor::MonitorHandle,
    window::WindowBuilder,
};

fn main() {

    let monitor = prompt_for_monitor(&event_loop);

        .with_inner_size((200, 200).into())
        .with_outer_position(monitor.clone(), (100, 100).into())
        .with_fullscreen(Some(monitor))
        .build(&event_loop)
        .unwrap();

        match event {
            Event::WindowEvent {
                window_id,
            Event::WindowEvent {
                event:
                    WindowEvent::KeyboardInput {
                        input:
                            KeyboardInput {
                                state: ElementState::Pressed,
                                virtual_keycode: Some(key),
                                ..
                            },
                        ..
                    },
                window_id,
            } if window_id == window.id() => match key {
                VirtualKeyCode::Q => {
                }
                VirtualKeyCode::F => {
                    let is_fullscreen = window.fullscreen().is_some();

                    if is_fullscreen {
                        window.set_fullscreen(None);
                    } else {
                        let monitor = window.current_monitor();
                        window.set_fullscreen(Some(monitor));
                    }
                }
                _ => (),
            },
            _ => (),
        }
    });
}

// Enumerate monitors and prompt user to choose one
fn prompt_for_monitor(event_loop: &EventLoop<()>) -> MonitorHandle {
    for (num, monitor) in event_loop.available_monitors().enumerate() {
        println!("Monitor #{}: {:?}", num, monitor.name());
    }

    print!("Please write the number of the monitor to use: ");
    io::stdout().flush().unwrap();

    let mut num = String::new();
    io::stdin().read_line(&mut num).unwrap();
    let num = num.trim().parse().ok().expect("Please enter a number");
    let monitor = event_loop
        .available_monitors()
        .nth(num)
        .expect("Please enter a valid ID");

    println!("Using {:?}", monitor.name());

    monitor
}
```

</details>

chrisduerr:
@murarth Does it work if you set `with_global_outer_position` at startup and run `set_fullscreen` right after setting an invisibly launched window to visible? Your example doesn't really help much since I don't think most applications know exactly which monitor they want to get launched on.

murarth:
This works for me, too.

<details>

```rust
        .with_inner_size((200, 200).into())
        .with_global_outer_position((300, 300).into())
        .with_visible(false)
        .build(&event_loop)
        .unwrap();

    window.set_visible(true);
    window.set_fullscreen(Some(window.current_monitor()));
```

</details>

>I don't think most applications know exactly which monitor they want to get launched on.

If you have the desired `PhysicalPosition`, you can iterate through the available monitors and select the one containing that physical point, using `position` and `size` methods on `MonitorHandle`. winit uses maximum overlapping area in `current_monitor`, but as long as the window is to be placed within only one screen's boundaries, it's the same result.

chrisduerr:
> This works for me, too.

The example you provided fullscreens on the correct monitor in i3, however it does not restore the proper position after leaving fullscreen. It just returns to 0x0 of the monitor it was fullscreened on.

> If you have the desired PhysicalPosition, you can iterate through the available monitors and select the one containing that physical point, using position and size methods on MonitorHandle. winit uses maximum overlapping area in current_monitor, but as long as the window is to be placed within only one screen's boundaries, it's the same result.

I'm aware that this is possible, however this is quite a bit of work and requires you to want to set the position. If I want to fullscreen at startup, but I don't know a specific position (I want it on the current monitor), then this will not work.

murarth:
>If I want to fullscreen at startup, but I don't know a specific position (I want it on the current monitor), then this will not work.

It sounds like this is fundamentally in conflict with winit's API. You must specify a monitor to build a window in fullscreen mode and there's no such thing as a "current monitor" without a visible window.

The X11-specific bit of this is just that a window is not truly visible until some time after the call to `.set_visible(true)`. So, the only option may be to wait until the window is moved/resized/exposed and set fullscreen at that point.

chrisduerr:
> It sounds like this is fundamentally in conflict with winit's API. You must specify a monitor to build a window in fullscreen mode and there's no such thing as a "current monitor" without a visible window.

Exactly, which is why I think it needs to be possible to call `set_fullscreen(current_monitor)` right after startup, but the linked issue has more discussion related to that. However your example with setting fullscreen after visible still doesn't work, so this PR should likely focus on the position part.

ZeGentzy:
What is the status on this PR?

murarth:
I ended up trying to do a few different things at once with this PR and it's become something of a mess. I'm going to close it and take another look at each part separately.

Issue #985, opened 2019-06-26
Add `Window::set_minimized` method


felixrabe:
I was surprised to find that such a method was missing: https://docs.rs/winit/0.20.0-alpha1/winit/window/struct.Window.html

Osspial:
Yeah, we should absolutely expose this. This should be trivial to implement on Windows.

kchibisov:
I guess this bug should be closed now.

Issue #983, opened 2019-06-25
MacOS minimize support


chrisduerr:
This has been reported to Alacritty in https://github.com/jwilm/alacritty/issues/2534.

Currently winit exposes an API on all platforms to set the window visibility using the `set_visible` method. If this is called on macOS, the window is hidden without adding it to the dock, this is working as intended.

However, on macOS a different method of hiding windows exists too, this is called `miniaturize` [in their docs](https://developer.apple.com/documentation/appkit/nswindow/1419426-miniaturize?language=objc)  which makes the window invisible and adds it to the dock. I believe this is what happens when the minimize button is clicked in the window decorations, but as far as I can tell there's no way to call this explicitly from winit.

The generated cocoa-rs bindings provide [this method](https://github.com/servo/cocoa-rs/blob/ea0667fdd2cffbdc442e50bbf6a2977ba7c2d322/src/appkit.rs#L1018) on `NsWindow`. So I think it should be relatively trivial to implement this using the [`WindowsExtMacOS`](https://github.com/rust-windowing/winit/blob/master/src/platform/macos.rs#L12), since I don't think it's available in this fashion on any other platform.

Alternatively, this could be declared as out of scope, so downstream consumers would have to use the [`ns_window`](https://github.com/rust-windowing/winit/blob/master/src/platform/macos.rs#L16) call instead to do this themselves.

felixrabe:
I think that's perfectly in scope. I just looked at https://docs.rs/winit/0.20.0-alpha1/winit/window/struct.Window.html because I wondered whether this could not be handled by some platform-independent `set_minimized` method. Turns out, there isn't one.

felixrabe:
Related: #783 (Minimizing window on Windows causes WindowEvent::Resized(size=0) event)

felixrabe:
Opened #985 (Add `Window::set_minimized` method) about the missing method.

kchibisov:
I guess this bug should be closed now.

PR #990, opened 2019-06-27
Window::set_minimized (#985)

SuperiorJT:
Gets started on supporting `Window::set_minimized` (#985, #983). I'm very new to this repo and a beginner in Rust so it might not be exactly how you guys want it. I've tested it on macOS via keyboard events and was able to make it minimize similar to other macOS applications. I'm willing to work on the Windows side of this as well.

  - [x] Windows
  - [x] macOS
  - [x] X11
  - [x] Wayland
  - [x] iOS


src/platform_impl/linux/wayland/window.rs
```diff
@@ -278,6 +278,10 @@ impl Window {
         *(self.need_frame_refresh.lock().unwrap()) = true;
     }
 
+    pub fn set_minimized(&self, minimized: bool) {
+        unimplemented!();
```
vberger:
```suggestion
        // An app cannot un-minimize itself on Wayland
        if minimized {
            self.frame.lock().unwrap().set_minimized();
        }
```

SuperiorJT:
On the Windows implementation do we want to create a `WindowFlag` for `MINIMIZED`? I'm looking at the way `set_maximized` works and it looks like we could do something similar, but I don't know if we should track that in our `WindowState`.

Osspial:
@SuperiorJT We should track that in `WindowFlag`s, yeah.

ZeGentzy:
@SuperiorJT What is the status on this PR?

SuperiorJT:
Implementations are made for windows, macOS and wayland. We need someone to develop the x11 and iOS (if applicable) implementation since I am unable to. Once all is implemented, then I can go through and fmt and update required documentation.

murarth:
I have an implementation for X11 ready: https://github.com/murarth/winit/tree/x11-set-minimize

SuperiorJT:
I added @murarth's X11 implementation, formatted, and added documentation. It's ready for review.

src/platform_impl/android/mod.rs
```diff
@@ -359,6 +359,11 @@ impl Window {
         Err(ExternalError::NotSupported(NotSupportedError::new()))
     }
 
+    #[inline]
+    pub fn set_minimized(&self, _minimized: bool) {
+        unimplemented!()
+    }
```
murarth: Do we want this to panic or should it be a no-op? `set_maximized` below is a no-op.

Osspial: I don't think it particular matters, since the Android backend doesn't even compile right now.

Osspial:
I've tested this out on Windows, and this seems to cause glitches if you do the following:

1. Maximize the window
2. Minimize the window via `set_minimized`
3. Alt-tab back into the window

If you do that, the minimize button in the window's titlebar has the window restore icon instead of the window minimize icon, and fails to minimize the window when pressed.

SuperiorJT:
I've fixed the maximize issue. Windows didn't like the window styles being set after minimization. Let me know if you find any more issues.

murarth:
@goddessfreya I wrote the X11 implementation myself, so I can confirm it works. If we want to ping our X11 testers to get another set of eyes on it, that's fine by me.

goddessfreya:
@murarth Haha, I'm ever so slightly blind. I see. Did you test for weird behavior evolving fullscreened/unmapped windows? Should be good to go on the X11 side, I don't see much that could go wrong.

murarth:
@goddessfreya I've tested for any weird interactions and found none.

SuperiorJT:
I added the stub for WASM. Let me know if I missed anything.

PR #991, opened 2019-06-27
Handle WM_POINTER* events in favor of WM_TOUCH

### LINKED ISSUE #975: Handle WM_POINTER* events in favor of WM_TOUCH
As WM_TOUCH have some [well document problems](http://the-witness.net/news/2012/10/wm_touch-is-totally-bananas/), I think it would be more appropriate to handle the WM_POINTERDOWN, WM_POINTERUPDATE and WM_POINTERUP event family whenever possible. Even though these events are only sent by Windows 8.1 or higher, they are more robust, easier to set up (no need to call Un-/RegisterTouchWindow) and less error prone.

dam4rus:
Fixes #975 

Haven't tested on Windows 7, but it should work, since the POINTER events are only fired on Windows 8 or higher. It should be tested anyway because GetPointerFrameInfoHistory and SkipPointerFrameMessages doesn't exist in User32.dll on pre-Windows 8.

### DIFF
--- a/src/platform_impl/windows/dpi.rs
+++ b/src/platform_impl/windows/dpi.rs
@@ -9,13 +9,12 @@ use winapi::{
         winerror::S_OK,
     },
     um::{
-        libloaderapi::{GetProcAddress, LoadLibraryA},
         shellscalingapi::{
             MDT_EFFECTIVE_DPI, MONITOR_DPI_TYPE, PROCESS_DPI_AWARENESS,
             PROCESS_PER_MONITOR_DPI_AWARE,
         },
         wingdi::{GetDeviceCaps, LOGPIXELSX},
-        winnt::{HRESULT, LPCSTR},
+        winnt::HRESULT,
         winuser::{self, MONITOR_DEFAULTTONEAREST},
     },
 };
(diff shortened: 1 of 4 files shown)


felixrabe:
Some meta suggestions regarding this PR:

- Change PR title to 'Handle WM_POINTER* events in favor of WM_TOUCH'.
- Add 'Fixes #975' to the description.

ZeGentzy:
Hey @0xpr03, and @FliegendeWurst,

Can one of you two test this on windows 7?

Hey @arthrp, and @notriddle,

Can one of you two test this on windows 8 or 10?

0xpr03:
Build & Tests pass on win 7

0xpr03:
Update: All examples are broken with (window example):
```
error: process didn't exit successfully: target\debug\examples\window.exe (exit code: 0xc0000139, STATUS_ENTRYPOINT_NOT_FOUND)
```

dam4rus:
As i mentioned, the functions needed by these events are not available on Windows 7 and older. I couldn't find anything about how Rust links FFI functions and I assumed that it uses dynamic linking. Because handling WM_TOUCH already breaks pre-Windows 7 compatibility, i guess that it's not priority for winit to support Windows XP. 

So we should load those functions dynamically through LoadLibrary and GetProcAddress. I've found an example usage of GetProcAddress in 'platform_impl/windows/dpi.rs', where a 'get_function' macro is defined to ease this process. Maybe this macro and it's accompanying function should be moved into 'platform_impl/windows/util.rs' and made visible within the whole module.

0xpr03:
We probably don't want support for anything older, but windows 7 is still supported and will be for longer on the industry side, so I'd personally not drop w7 support for now.

Osspial:
@ZeGentzy thanks for pinging testers!

@dam4rus Sorry about the delay in reviewing. The dynamic function loading machinery in `dpi` should indeed be used for this, and moving it into `util` is a good idea.

dam4rus:
@Osspial Should i create a separate PR for moving the dynamic function loading macro?

Osspial:
@dam4rus Nah, adding it in this PR is fine.

dam4rus:
@0xpr03 It should work now. Can you test it?

0xpr03:
@dam4rus yes, current version works with examples

Osspial:
I'll go ahead and merge this now. Thank you for implementing this!

dam4rus:
WM_TOUCH already handled multiple touch points (AFAIK) and so does WM_POINTER. So, it depends on what you mean by multitouch support? This just sends individual touch events for every touch point like the WM_TOUCH implementation. The client can store these events to handle simultaneous touch points. Even so, it's kinda hard to tell what's a "touch event frame" with the current implementation.

Osspial:
Ah, I suppose that should just be marked with a checkmark instead of a question mark, then. I'll merge this now and correct that.

PR #992, opened 2019-06-27
Remove comment that contradicts code

felixrabe:


vberger:
Well, it *is* private to the crate though, so I don't get how this comment is wrong?

felixrabe:
It was changed from private to `pub(crate)`, but the comment was not updated.

I see no point in mentioning the visibility of a field in its doc comment in general, because then you have to update the same thing in two places if you need to change it.

PR #994, opened 2019-06-28
Fix issues with redraw_requested when called during EventsCleared

Osspial:

If the user calls `request_redraw` multiple times during the `EventsCleared` handler, multiple redraw events would get dispatched. This fixes that.

Example displaying that behavior here:

```rust
use winit::{
    event::{Event, WindowEvent},
    window::WindowBuilder,
};

fn main() {

        .build(&event_loop)
        .unwrap();

        println!("{:?}", event);

        match event {
            Event::WindowEvent {
                window_id,
            Event::EventsCleared => {
                window.request_redraw();
                window.request_redraw();
            }
        }
    });
}
```

All the `request_redraw` calls should be consolidated into a single event, but they weren't here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 - On X11, fix sanity check which checks that a monitor's reported width and height (in millimeters) are non-zero when calculating the DPI factor.
 - On X11, implement `_NET_WM_PING` to allow desktop environment to kill unresponsive programs.
 - On Windows, when a window is initially invisible, it won't take focus from the existing visible windows.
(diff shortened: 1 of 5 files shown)


Osspial:
@notriddle I'm pinging you to test this, since you're on the Windows testers list in the wiki. Would you be able to give this a go-around to make sure it doesn't introduce problematic behavior?

notriddle:
Well, using your example, I can see that `RedrawRequested` is being enqueued exactly once. So it seems to work fine at first glance.

For more thorough testing, I'm also trying [alacritty's patch](https://github.com/chrisduerr/alacritty/tree/evlp2), which has no obvious problems other than the already-known resize problems.

PR #996, opened 2019-06-28
Release Alpha 2

Osspial:
I'd like to get a couple of the latest bugfixes merged before releasing this but we should be able to merge this later today.


Osspial:
@ZeGentzy right, that took a while. I wanted to get #994 ready before releasing this, since there were quite a few bugs in win32's handling of `RedrawRequested`. That should be good to go now, though.

Osspial:
Doesn't look like we need to release a new version of Glutin for downstream users to use the new alpha.

PR #997, opened 2019-06-28
macOS: Dpi overhaul

vbogaevsky:

Updates objc to current version.
Fixed every error, so it compiles. Any suggestions for a proper way to test it?
I am not entirely sure I've handled `WindowEvent::HiDpiFactorChanged` creation correctly.

### DIFF
--- a/src/platform_impl/macos/app_state.rs
+++ b/src/platform_impl/macos/app_state.rs
@@ -13,7 +13,7 @@ use std::{
 use cocoa::{
     appkit::{NSApp, NSWindow},
     base::nil,
-    foundation::{NSRect, NSSize},
+    foundation::NSSize,
 };
 
 use crate::{
@@ -187,10 +187,14 @@ impl Handler {
         }
     }
 
-    fn handle_hidpi_factor_changed_event(&self, ns_window: IdRef, hidpi_factor: f64) {
-        let ns_size = unsafe { NSWindow::frame(*ns_window).size };
-        let new_size = LogicalSize::new(ns_size.width, ns_size.height).to_physical(hidpi_factor);
-        let new_inner_size = &mut Some(new_size);
+    fn handle_hidpi_factor_changed_event(
+        &self,
+        ns_window: IdRef,
+        suggested_size: LogicalSize,
+        hidpi_factor: f64,
+    ) {
+        let size = suggested_size.to_physical(hidpi_factor);
+        let new_inner_size = &mut Some(size);
         let event = Event::WindowEvent {
             window_id: WindowId(get_window_id(*ns_window)),
             event: WindowEvent::HiDpiFactorChanged {
@@ -201,21 +205,21 @@ impl Handler {
 
         self.handle_nonuser_event(event);
 
-        let origin = unsafe { NSWindow::frame(*ns_window).origin };
-        if let Some(physical_size) = new_inner_size {
-            let logical_size = physical_size.to_logical(hidpi_factor);
-            let size = NSSize::new(logical_size.width, logical_size.height);
-            let rect = NSRect::new(origin, size);
-            unsafe { ns_window.setFrame_display_(rect, cocoa::base::YES) };
-        };
+        // let origin = unsafe { NSWindow::frame(*ns_window).origin };
+        let physical_size = new_inner_size.unwrap_or(size);
+        let logical_size = physical_size.to_logical(hidpi_factor);
+        let size = NSSize::new(logical_size.width, logical_size.height);
+        // let rect = NSRect::new(origin, size);
+        unsafe { NSWindow::setContentSize_(*ns_window, size) };
     }
 
     fn handle_event(&self, proxy: EventProxy) {
         match proxy {
             EventProxy::HiDpiFactorChangedProxy {
                 ns_window,
+                suggested_size,
                 hidpi_factor,
-            } => self.handle_hidpi_factor_changed_event(ns_window, hidpi_factor),
+            } => self.handle_hidpi_factor_changed_event(ns_window, suggested_size, hidpi_factor),
         }
     }
 }
--- a/src/platform_impl/macos/app_state.rs
+++ b/src/platform_impl/macos/app_state.rs
@@ -274,7 +274,7 @@ impl AppState {
             HANDLER.set_in_callback(true);
             // Since we do not push_back, we don't need to take_deferred
             // for event in HANDLER.take_deferred() {
-                HANDLER.handle_nonuser_event(event);
+            HANDLER.handle_nonuser_event(event);
             // }
             HANDLER.set_in_callback(false);
         }
--- a/src/platform_impl/macos/app_state.rs
+++ b/src/platform_impl/macos/app_state.rs
@@ -23,8 +23,8 @@ lazy_static! {
     static ref HANDLER: Handler = Default::default();
 }
 
-impl Event<Never> {
-    fn userify<T: 'static>(self) -> Event<T> {
+impl<'a, Never> Event<'a, Never> {
+    fn userify<T: 'static>(self) -> Event<'a, T> {
         self.map_nonuser_event()
             // `Never` can't be constructed, so the `UserEvent` variant can't
             // be present here.
@@ -33,7 +33,8 @@ impl Event<Never> {
 }
 
 pub trait EventHandler: Debug {
-    fn handle_nonuser_event(&mut self, event: Event<Never>, control_flow: &mut ControlFlow);
+    // Not sure probably it should accept Event<'static, Never>
+    fn handle_nonuser_event(&mut self, event: Event<'_, Never>, control_flow: &mut ControlFlow);
     fn handle_user_events(&mut self, control_flow: &mut ControlFlow);
 }
 
@@ -54,10 +55,10 @@ impl<F, T> Debug for EventLoopHandler<F, T> {
 
 impl<F, T> EventHandler for EventLoopHandler<F, T>
 where
-    F: 'static + FnMut(Event<T>, &RootWindowTarget<T>, &mut ControlFlow),
+    F: 'static + FnMut(Event<'_, T>, &RootWindowTarget<T>, &mut ControlFlow),
     T: 'static,
 {
-    fn handle_nonuser_event(&mut self, event: Event<Never>, control_flow: &mut ControlFlow) {
+    fn handle_nonuser_event(&mut self, event: Event<'_, Never>, control_flow: &mut ControlFlow) {
         (self.callback)(event.userify(), &self.window_target, control_flow);
         self.will_exit |= *control_flow == ControlFlow::Exit;
         if self.will_exit {
@@ -86,8 +87,8 @@ struct Handler {
     control_flow_prev: Mutex<ControlFlow>,
     start_time: Mutex<Option<Instant>>,
     callback: Mutex<Option<Box<dyn EventHandler>>>,
-    pending_events: Mutex<VecDeque<Event<Never>>>,
-    deferred_events: Mutex<VecDeque<Event<Never>>>,
+    pending_events: Mutex<VecDeque<Event<'static, Never>>>,
+    deferred_events: Mutex<VecDeque<Event<'static, Never>>>,
     pending_redraw: Mutex<Vec<WindowId>>,
     waker: Mutex<EventLoopWaker>,
 }
@@ -96,11 +97,11 @@ unsafe impl Send for Handler {}
 unsafe impl Sync for Handler {}
 
 impl Handler {
-    fn events<'a>(&'a self) -> MutexGuard<'a, VecDeque<Event<Never>>> {
+    fn events<'a>(&'a self) -> MutexGuard<'a, VecDeque<Event<'static, Never>>> {
         self.pending_events.lock().unwrap()
     }
 
-    fn deferred<'a>(&'a self) -> MutexGuard<'a, VecDeque<Event<Never>>> {
+    fn deferred<'a>(&'a self) -> MutexGuard<'a, VecDeque<Event<'static, Never>>> {
         self.deferred_events.lock().unwrap()
     }
 
@@ -144,11 +145,11 @@ impl Handler {
         *self.start_time.lock().unwrap() = Some(Instant::now());
     }
 
-    fn take_events(&self) -> VecDeque<Event<Never>> {
+    fn take_events(&self) -> VecDeque<Event<'_, Never>> {
         mem::replace(&mut *self.events(), Default::default())
     }
 
-    fn take_deferred(&self) -> VecDeque<Event<Never>> {
+    fn take_deferred(&self) -> VecDeque<Event<'_, Never>> {
         mem::replace(&mut *self.deferred(), Default::default())
     }
 
@@ -164,7 +165,7 @@ impl Handler {
         self.in_callback.store(in_callback, Ordering::Release);
     }
 
-    fn handle_nonuser_event(&self, event: Event<Never>) {
+    fn handle_nonuser_event(&self, event: Event<'_, Never>) {
         if let Some(ref mut callback) = *self.callback.lock().unwrap() {
             callback.handle_nonuser_event(event, &mut *self.control_flow.lock().unwrap());
         }
@@ -182,7 +183,7 @@ pub enum AppState {}
 impl AppState {
     pub fn set_callback<F, T>(callback: F, window_target: RootWindowTarget<T>)
     where
-        F: 'static + FnMut(Event<T>, &RootWindowTarget<T>, &mut ControlFlow),
+        F: 'static + FnMut(Event<'_, T>, &RootWindowTarget<T>, &mut ControlFlow),
         T: 'static,
     {
         *HANDLER.callback.lock().unwrap() = Some(Box::new(EventLoopHandler {
@@ -247,21 +248,21 @@ impl AppState {
         }
     }
 
-    pub fn queue_event(event: Event<Never>) {
+    pub fn queue_event(event: Event<'static, Never>) {
         if !unsafe { msg_send![class!(NSThread), isMainThread] } {
             panic!("Event queued from different thread: {:#?}", event);
         }
         HANDLER.events().push_back(event);
     }
 
-    pub fn queue_events(mut events: VecDeque<Event<Never>>) {
+    pub fn queue_events(mut events: VecDeque<Event<'static, Never>>) {
         if !unsafe { msg_send![class!(NSThread), isMainThread] } {
             panic!("Events queued from different thread: {:#?}", events);
         }
         HANDLER.events().append(&mut events);
     }
 
-    pub fn send_event_immediately(event: Event<Never>) {
+    pub fn send_event_immediately(event: Event<'static, Never>) {
         if !unsafe { msg_send![class!(NSThread), isMainThread] } {
             panic!("Event sent from different thread: {:#?}", event);
         }
--- a/src/platform_impl/macos/app_state.rs
+++ b/src/platform_impl/macos/app_state.rs
@@ -20,7 +20,7 @@ use crate::{
 };
 
 lazy_static! {
-    static ref HANDLER: Handler = Default::default();
+    static ref HANDLER: Handler<'static> = Default::default();
 }
 
 impl<'a, Never> Event<'a, Never> {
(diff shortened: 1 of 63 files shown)


src/platform_impl/macos/window.rs
```diff
@@ -122,10 +122,17 @@ fn create_window(
         let frame = match screen {
             Some(screen) => appkit::NSScreen::frame(screen),
             None => {
-                let (width, height) = attrs
-                    .inner_size
-                    .map(|logical| (logical.width, logical.height))
-                    .unwrap_or_else(|| (800.0, 600.0));
+                // Not sure what dpi_factor we need here.
+                // If i understand correctly, here we fail to create NSScreen frame, so we've got
+                // no dpi_factor
```
aleksijuvani: This code path determines the initial dimensions of the window when it's not in fullscreen mode, from the size provided by the user, or if not provided, a default value of 800 by 600 logical pixels is used. So this would incorrectly use a DPI factor of 1.0 for the dimensions provided by the user. Perhaps we can defer setting the window frame until the `NSWindow` is initialized and the appropriate `backingScaleFactor` can be used?

src/platform_impl/macos/window.rs
```diff
@@ -543,14 +569,13 @@ impl UnownedWindow {
...
-        cursor_position: LogicalPosition,
-    ) -> Result<(), ExternalError> {
+    pub fn set_cursor_position(&self, cursor_position: Position) -> Result<(), ExternalError> {
         let window_position = self.inner_position().unwrap();
+        let dpi_factor = self.hidpi_factor();
+        let logical_cursor_position = cursor_position.to_logical(dpi_factor);
         let point = appkit::CGPoint {
-            x: (cursor_position.x + window_position.x) as CGFloat,
-            y: (cursor_position.y + window_position.y) as CGFloat,
+            x: (logical_cursor_position.x + window_position.x) as CGFloat,
```
aleksijuvani: Since `inner_position` is now returning a `PhysicalPosition` instead of a `LogicalPosition`, this is now adding together the logical position of the cursor and the physical position of the window. The window's position ought to be converted to a logical position also.

src/platform_impl/macos/window.rs
```diff
@@ -424,7 +433,7 @@ impl UnownedWindow {
             .into())
```
aleksijuvani: I believe this needs to be converted from logical coordinates to physical coordinates.

src/platform_impl/macos/window.rs
```diff
@@ -435,7 +444,9 @@ impl UnownedWindow {
...
```
aleksijuvani: I think this also needs to be converted to physical coordinates.

aleksijuvani:
Looking at this pull request made me realize that winit may be conflating the concept of screen-space coordinates and window-space coordinates (i.e. it's using the `backingScaleFactor` of the window to determine its dimensions, instead of the `backingScaleFactor` of the screen that it is on). This is in fact mentioned in the documentation for `NSWindow`'s `backingScaleFactor` quite explicitly:

> It is important to note that the value of this property does not represent anything concrete, such as pixel density or physical size, because it can vary based on the configured display mode. For example, the display may be in a mirrored configuration that is still high-resolution scaled, resulting in pixel geometry that may not match the native resolution of the display device.

I'll have to look into this later and verify if this is correct, but I think that's outside the scope of this pull request.

vbogaevsky:
@aleksijuvani I think you are correct, we do have to use `backingScaleFactotr` of the screen, not the window.

aleksijuvani:
From a quick test, these two values appear to be equivalent on my current configuration, so perhaps this is not a problem (or at least, not as big of a problem as I thought). But like I said, I feel like addressing this is outside the scope of this pull request, so I would suggest to continue using `backingScaleFactor` of the window for the time being, and this can be changed at a later time if it turns out to be incorrect.

src/platform_impl/macos/window.rs
vbogaevsky: The trouble is we need a `frame` to initialise an `NSWindow`. Also in this branch of `match` `screen` is `None`, so we cannot query it for `backingScaleFactor`.
The `screen` being `None` is kind of odd, by the way.

aleksijuvani: I think we can pick the screen with [`init:contentRect:styleMask:backing:defer:screen:`](https://developer.apple.com/documentation/appkit/nswindow/1419755-init) for the window. We could set this to the main screen (i.e. the screen that has the current keyboard focus, `NSScreen:main`). We would then be able to use the `backingScaleFactor` of the main screen here.

src/platform_impl/macos/window_delegate.rs
```diff
@@ -275,10 +288,18 @@ extern "C" fn window_did_move(this: &Object, _: Sel, _: id) {
...
+        if state.previous_dpi_factor != hidpi_factor {
+            state.previous_dpi_factor = hidpi_factor;
+            state.emit_event(
+                WindowEvent::HiDpiFactorChanged {
+                    hidpi_factor,
+                    new_inner_size: &mut Some(inner_size(ns_size, hidpi_factor)),
+                }
+                .to_static()
+                .unwrap(),
+            );
```
Osspial: WRT `HiDpiFactorChanged` (these comments also apply to the other places in the code where it gets dispatched):

The `new_inner_size` field is supposed to take a mutable reference to the new window size suggested by the OS. That size can then be modified by the user if they would like to resize the window in a different way in response to the DPI change, and the window's size after the DPI change gets set to whatever value is pointed to by `new_inner_size` after the event handler. If the OS doesn't suggest a size, we set it to `None`, and if the user wants to ignore any sizing changes, they can set `new_inner_size` to `None` in their callback and we should leave the window's size as it was before the DPI factor change.

With that in mind, there are a couple concrete problems I can point to in the current implementation that must be solved before this gets merged:

1. `new_inner_size` must reference a value that later gets read by the code dispatching the event. Right now, the event handler doesn't do that, so any new value the user passes to `new_inner_size` is getting ignored.
2. `HiDpiFactorChanged` cannot use `emit_event`. This is because `emit_event` queues the event for dispatch, rather than dispatching it immediately, and thus requires an event with a `'static` lifetime. Skimming the code it doesn't look like the function you need exists right now, but you should be able to adapt the code in [`emit_resize_event`](https://github.com/rust-windowing/winit/blob/3a449af496485e0e106b5b01234480507121aa94/src/platform_impl/macos/window_delegate.rs#L84-L92), since that code dispatches the event without going through an event queue.

As far as testing this goes, it'll probably be easiest if you've got another monitor somewhere in your house that you can hook up, since it seems like that's the main place this event with get dispatched from. If that's not an option, it's worth looking into where [`windowDidChangeBackingProperties`](https://developer.apple.com/documentation/appkit/nswindowdelegate/1419517-windowdidchangebackingproperties) gets dispatched since that's the other place this event can get emitted from, but how exactly you trigger that seems less clear from a cursory glance.

Also, it may be helpful to look at the [win32 implementation for reference](https://github.com/rust-windowing/winit/blob/3a449af496485e0e106b5b01234480507121aa94/src/platform_impl/windows/event_loop.rs#L1568-L1590) if you'd like a concrete code example.

As always, if any part of that was unclear or if you've got any questions, feel free to ask and I'll answer them to the best of my ability.

vbogaevsky: Hi! Thanks for adding some clarity. I was quite confused why `HiDpiFactorChanged` should've been `'static`.
Do we get a new `WindowDelegateState` every time dpi factor changes?

Osspial: @vbogaevsky Doesn't look like it. I think `WindowDelegateState` is created alongside the window, and the `HiDpiFactorChanged` event emitted there is just used to notify the event loop that the window starts out with a non-`1.0` DPI factor.

vbogaevsky: @Osspial `emit_resize_event` utilises `AppState::send_event_immediately` which as `emit_event` expects event to be `'static` too. As I understand, it works for  `emit_resize_event` because `WindowEvent::Resized` holds no references, so it can be evaluated as `'static`, however we have to give `WindowEvent::HiDpiFactorChanged` a reference to an `Option` holding `new_inner_size`.
The only logical solution I see is to make `WindowDelegateState` `'static', then it will be holding a '`static' value which we will be able to reference.

vbogaevsky: I hope I found a solution to this problem with `'static`

vbogaevsky:
I've tested the emission of `HiDpiFactorChanged` by dropping a test window created on built-in display (with retina) to a plugged in display without retina.

src/platform_impl/macos/window_delegate.rs
```diff
@@ -275,10 +296,12 @@ extern "C" fn window_did_move(this: &Object, _: Sel, _: id) {
...
-        let dpi_factor = unsafe { NSWindow::backingScaleFactor(*state.ns_window) } as f64;
-        if state.previous_dpi_factor != dpi_factor {
-            state.previous_dpi_factor = dpi_factor;
-            state.emit_event(WindowEvent::HiDpiFactorChanged(dpi_factor));
+        let hidpi_factor = unsafe { NSWindow::backingScaleFactor(*state.ns_window) } as f64;
+        let ns_size = unsafe { NSWindow::frame(*state.ns_window).size };
+        let new_size = inner_size(ns_size, hidpi_factor);
+        if state.previous_dpi_factor != hidpi_factor {
+            state.previous_dpi_factor = hidpi_factor;
+            state.emit_hidpi_factor_changed_event(&mut Some(new_size));
```
Osspial: Right now, this throws away any changes the user makes to the `new_inner_rect_opt` field. We should read the value the user places in that field, and change the window's size to match it if the user wants a different size than the size suggested by the OS.

src/platform_impl/macos/window_delegate.rs
```diff
@@ -289,10 +312,12 @@ extern "C" fn window_did_change_screen(this: &Object, _: Sel, _: id) {
...
```
Osspial: ditto

src/platform_impl/macos/window_delegate.rs
Osspial: I've set up some examples to verify that stuff is behaving correctly.

This example should set the window's physical size to `256x256` whenever the DPI factor changes, and will panic if that doesn't happen:

```rust
use winit::{
    dpi::PhysicalSize,
    event::{Event, WindowEvent},
    window::WindowBuilder,
};

fn main() {

        .build(&event_loop)
        .unwrap();

    let mut assert_size = None;

        println!("{:?}", event);

        if let Some(size) = assert_size.take() {
            assert_eq!(size, window.inner_size());
        }

        match event {
            Event::WindowEvent {
                window_id,
            Event::WindowEvent {
                event: WindowEvent::HiDpiFactorChanged{new_inner_size, ..},
                ..
            } => {
                assert_size = Some(PhysicalSize::new(256, 256));
                *new_inner_size = assert_size;
            }
        }
    });
}
```

This example should keep the window's physical size consistent when the DPI factor changes.

Incidentally, both of those examples panic on Windows, so it looks like I've got some bugfixing to do as well!

src/platform_impl/macos/app_state.rs
```diff
@@ -79,36 +80,38 @@ where
...
     ready: AtomicBool,
     in_callback: AtomicBool,
     control_flow: Mutex<ControlFlow>,
     control_flow_prev: Mutex<ControlFlow>,
     start_time: Mutex<Option<Instant>>,
     callback: Mutex<Option<Box<dyn EventHandler>>>,
-    pending_events: Mutex<VecDeque<Event<Never>>>,
-    deferred_events: Mutex<VecDeque<Event<Never>>>,
+    pending_events: Mutex<VecDeque<Event<'a, Never>>>,
+    // We don't need it any more. See line 270
```
ZeGentzy: Please don't refer to things by line number. Those are constantly changing!

src/platform_impl/macos/window_delegate.rs
ZeGentzy: Can these be turned into tests, @Osspial ?

src/platform_impl/macos/app_state.rs
vbogaevsky: @ZeGentzy These are PR only, I'll remove them before it's merged.

src/platform_impl/macos/app_state.rs
```diff
@@ -79,36 +80,38 @@ where
...
+struct Handler<'a> {
     ready: AtomicBool,
     in_callback: AtomicBool,
     control_flow: Mutex<ControlFlow>,
     control_flow_prev: Mutex<ControlFlow>,
     start_time: Mutex<Option<Instant>>,
     callback: Mutex<Option<Box<dyn EventHandler>>>,
-    pending_events: Mutex<VecDeque<Event<Never>>>,
-    deferred_events: Mutex<VecDeque<Event<Never>>>,
+    pending_events: Mutex<VecDeque<Event<'a, Never>>>,
```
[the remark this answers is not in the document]
vbogaevsky: rustc demanded a lifetime. `Event` in `pending_events` and `deferred_events` get `'static` through `lazy_static!` initialization of `Handler`. I guess, it's better to remove `'a` and set `Event` to have `'static` explicitly in order to not pollute code with lifetimes.

src/platform_impl/macos/app_state.rs
```diff
@@ -79,36 +80,38 @@ where
...
+    // We don't need it any more. See line 270
+    // deferred_events: Mutex<VecDeque<Event<'a, Never>>>,
```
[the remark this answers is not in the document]
vbogaevsky: A-ha, now I see.
It eludes me how can we make `WindowEvent::HiDpiChanged` `'static` then.

src/platform_impl/macos/window_delegate.rs
Osspial: Ideally yes, but doing so would require a lot of work that I'm not willing to do right now. The only way I can see to automatically trigger `HiDpiFactorChanged` would be to modify the system's monitor DPI configuration, and that involves diving into the platform-specific APIs.

src/platform_impl/macos/app_state.rs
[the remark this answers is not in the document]
Osspial: Right now on Windows we just throw away the `HiDpiFactorChanged` event when it can't be dispatched, since the set of circumstances I described is pretty rare, but it occurs to me now that a better solution would be to re-post the OS event at the end of the OS event queue and deal with it later. I'm not entirely sure if that's possible on macOS, so another solution would be to push a new event type to `deferred_events` with no lifetime that gets transformed into a proper `HiDpiFactorChanged` event when it gets dispatched.

src/platform_impl/macos/app_state.rs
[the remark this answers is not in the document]
Osspial: Ah, OK. I'd switch to `'static` since having the lifetime in `Handler` states that the `Event`'s lifetime will last as long as the `Handler`, which for lifetime-d events isn't the case.

Osspial:
@vbogaevsky Any update on the status of this?

vbogaevsky:
@Osspial I'm still tinkering with it. I hope I'll have a working version by Monday.

src/platform_impl/macos/app_state.rs
```diff
@@ -245,29 +280,42 @@ impl AppState {
...
+        HANDLER.deferred().push_back(wrapper);
         if !HANDLER.get_in_callback() {
             HANDLER.set_in_callback(true);
-            for event in HANDLER.take_deferred() {
-                HANDLER.handle_nonuser_event(event);
+            for wrapper in HANDLER.take_deferred() {
+                match wrapper {
+                    EventWrapper::StaticEvent(event) => HANDLER.handle_nonuser_event(event),
+                    EventWrapper::EventProxy(proxy) => {
+                        let event = HANDLER.make_event(&proxy);
```
vbogaevsky: Got stuck here with "cannot borrow data  in a `&` reference as mutable on `HANDLER` and "use of moved value" on `event`.

vbogaevsky:
@Osspial, I think, I've found a solution. It compiles, but I have not tested it yet, because I currently don't have an access to a second monitor and it appears there is no way to change dpi of a built in display.

src/platform_impl/macos/app_state.rs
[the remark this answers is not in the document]
vbogaevsky: @Osspial, I went with second option and made a proxy event and a wrapper to pass to `deferred_events`.

aleksijuvani:
I can try this out some time next week. I ordered some cables to test this out with external monitors, but it'll be a few days until they arrive.

vbogaevsky:
@aleksijuvani, great!

src/platform_impl/macos/window_delegate.rs
vbogaevsky: @Osspial, I've tested current implementation by setting different screen resolutions, and when it should have changed size of the window, it keeps the promotions of the window.
Is it how we do want to keep it?
On the second test it performs correctly.

vbogaevsky: @Osspial, in the first example, on hidpi change we should get a square window (256 x 256), but we get a rectangular (256 x 212). Do we want to keep it that way?

src/platform_impl/macos/window_delegate.rs
```diff
@@ -289,12 +297,8 @@ extern "C" fn window_did_change_screen(this: &Object, _: Sel, _: id) {
 extern "C" fn window_did_change_backing_properties(this: &Object, _: Sel, _: id) {
     trace!("Triggered `windowDidChangeBackingProperties:`");
     with_state(this, |state| {
-        let dpi_factor = unsafe { NSWindow::backingScaleFactor(*state.ns_window) } as f64;
-        if state.previous_dpi_factor != dpi_factor {
-            state.previous_dpi_factor = dpi_factor;
-            state.emit_event(WindowEvent::HiDpiFactorChanged(dpi_factor));
-            state.emit_resize_event();
-        }
+        state.emit_static_hidpi_factor_changed_event();
```
aleksijuvani: This needs to keep track of the previous backing scale factor and not emit the event if the scale factor has not changed.

src/platform_impl/macos/window_delegate.rs
```diff
@@ -275,12 +287,8 @@ extern "C" fn window_did_move(this: &Object, _: Sel, _: id) {
 extern "C" fn window_did_change_screen(this: &Object, _: Sel, _: id) {
```
aleksijuvani: We can get rid of this callback. Tracking this in `windowDidChangeScreen` isn't necessary because `windowDidChangeBackingProperties` will be invoked if the backing scale factor changes.

src/platform_impl/macos/window.rs
```diff
@@ -451,35 +470,50 @@ impl UnownedWindow {
...
         }
     }
 
-    pub fn set_max_inner_size(&self, dimensions: Option<LogicalSize>) {
+    pub fn set_max_inner_size(&self, dimensions: Option<Size>) {
         unsafe {
-            let dimensions = dimensions.unwrap_or_else(|| (!0, !0).into());
-            set_max_inner_size(*self.ns_window, dimensions);
+            let dimensions = dimensions.unwrap_or(Logical(LogicalSize {
+                width: 0.0,
```
aleksijuvani: The documentation for [`maxSize`](https://developer.apple.com/documentation/appkit/nswindow/1419595-maxsize?language=objc) seems to imply that this should be `std::f32::MAX` instead.

src/platform_impl/macos/window_delegate.rs
aleksijuvani: The latter test doesn't seem correct to me. Shouldn't we be comparing the logical size, not the physical size? The physical size will always be different between two monitors with different scaling factors.

src/platform_impl/macos/window_delegate.rs
```diff
@@ -73,22 +73,34 @@ impl WindowDelegateState {
...
+            ns_window: IdRef::retain(*self.ns_window),
+            hidpi_factor,
+        });
+        AppState::send_event_immediately(wrapper);
+    }
+
     pub fn emit_resize_event(&mut self) {
         let rect = unsafe { NSView::frame(*self.ns_view) };
         let size = LogicalSize::new(rect.size.width as f64, rect.size.height as f64);
+        let hidpi_factor = self.previous_dpi_factor;
```
aleksijuvani:
```suggestion
        let hidpi_factor = unsafe { NSWindow::backingScaleFactor(*self.ns_window) } as f64;
```

src/platform_impl/macos/window_delegate.rs
vbogaevsky: @aleksijuvani, could you test it with the latest commit? I've tuned it all up a bit, and when I change resolution through System Preferences on my laptop both tests pass.

src/platform_impl/macos/window.rs
vbogaevsky: `LogicalSize` expects dimensions to be `f64`.
Docs also note "that he window server limits window sizes to 10,000."
Maybe we should default to `10000.0`?

src/platform_impl/macos/window_delegate.rs
vbogaevsky: I added tracking into `pub fn emit_static_hidpi_factor_changed_event()`

src/platform_impl/macos/window.rs
aleksijuvani: I think this needs to be `std::f32::MAX`. It's fine that we aren't using the full range of `f64` in `LogicalSize`. The docs mention explicitly that resetting to the default value can only be done by setting this to `FLT_MAX`, which equals `std::f32::MAX`. It might not matter, but it might also be that AppKit is doing an explicit comparison for the default value somewhere, which is why I'd prefer to set this to the exact value. As a sidenote, the data type of `CGFloat` is `f64` on 64-bit operating systems and `f32` on 32-bit operating systems, which is likely why `FLT_MAX` was chosen in AppKit for the default instead of `DBL_MAX`.

src/platform_impl/macos/window_delegate.rs
aleksijuvani: Looks good to me. No extraneous events are emitted anymore. Additionally, modifying the latter test here so that `let assert_size = window.inner_size().to_logical(window.hidpi_factor())` and `assert_eq!(assert_size, window.inner_size().to_logical(window.hidpi_factor()))` will make it pass. Both of the tests pass now when dragging the window between monitors with different scaling factors.

src/platform_impl/macos/window.rs
vbogaevsky: Ok, I set it tp use f32.

vbogaevsky:
@Osspial @aleksijuvani looks like all is good, can we merge?

ZeGentzy:
Merging. Cheers!

src/platform_impl/macos/window_delegate.rs
Osspial: @aleksijuvani The expected behavior when setting `HiDpiFactorChanged` to `None` was that the window's physical size would remain consistent, which is what the second test tested for. That was the intended, and as of #1130 the exhibited, behavior.

PR #1011, opened 2019-07-05
Fix transparent window with decorations

baysmith:
- [x] Tested on all platforms changed _(but not all Windows versions; Windows 10 only)_
- [x] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users _(not a noticeable change for most users)_
- [x] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior _(no user-facing changes)_
- [x] Created or updated an example program if it would help users understand this functionality _(no changes in functionality)_
- [x] Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented _(no features added)_


Osspial:
Could you still add a changelog entry? I'd disagree that this isn't a noticeable change for users, since it fixes a concrete bug that users (such as yourself!) can run into.

baysmith:
Done. Comment added to changelog.

PR #1014, opened 2019-07-05
On Windows, Unset maximized when transforming window

Osspial:

This PR causes calls to automatically unset `maximized` when calling `set_inner_size` or `set_outer_position`. I'm proposing this change for a few reasons:

1. Sizing or moving a window in the desktop shell unsets `maximized`, so doing this improves our consistency with the rest of the OS.
2. Displaying a window with maximized decorations when it isn't maximized is incorrect behavior.


ZeGentzy:
Is this good for merging?

PR #1017, opened 2019-07-06
PULL_REQUEST_TEMPLATE.md: Add entry on warnings

felixrabe:


Osspial:
We should also add [`-D warnings` to the CI](https://github.com/rust-unofficial/patterns/blob/master/anti_patterns/deny-warnings.md) so it fails in case of warnings.

rukai:
Uh, the page you linked to explains why that is bad idea...

felixrabe:
I'll test the idea in this PR, then force-revert, and then open a new PR so we can discuss this separately.

**Don't merge yet.**

felixrabe:
AppVeyor failing as expected.
Waiting for Travis.

felixrabe:
I'm not going to fix the `mem::uninitialized()` deprecation warnings right now, hoping someone else beats me to it.

Osspial:
@rukai Hmm, probably should've done more than just skim that article.

We could do `-D warnings -A deprecated`. I'm fairly OK with Travis catching warnings for compiler mis-features, since we'd want to fix those ASAP.

felixrabe:
I think `mem::uninitialized()` should be easy to fix, and I don't think we have other outstanding warnings (besides my warnings PR #1020), so `-A deprecated` is not needed (and not helpful for "OK with Travis catching warnings for compiler mis-features").

Osspial:
@felixrabe The `-A deprecated` is mainly for future-proofing PRs so they don't fail unnecessarily, since there's a decent chance we're using APIs internally that will get deprecated sometime in the future.

PR #1026, opened 2019-07-08
Add new `EventLoopWindowTargetExtUnix` trait.

ZeGentzy:

https://github.com/rust-windowing/glutin/issues/1184

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 
 # 0.20.0 Alpha 2 (2019-07-09)
 
+- **Breaking:** On Linux, the functions `is_wayland`, `is_x11`, `xlib_xconnection` and `wayland_display` have been moved to a new `EventLoopWindowTargetExtUnix` trait.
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 
+- **Breaking:** On Linux, the functions `is_wayland`, `is_x11`, `xlib_xconnection` and `wayland_display` have been moved to a new `EventLoopWindowTargetExtUnix` trait.
```
felixrabe: I think we usually add changes to the bottom of the list, not the top.

Also, that would be the first change marked as "Breaking". Because we're in "Alpha", I didn't mark my bool->enum change breaking, and there's at least one other change in the list that would be considered breaking.

So... I think you can remove "Breaking" here, and we should decide whether we label breaking changes individually or slap the "all below are breaking" label on top again.

ZeGentzy: Fixed. I'll push it later.

PR #1027, opened 2019-07-08
Replace `std::mem::uninitialized` with `MaybeUninit`

### LINKED ISSUE #1025: Replace `mem::uninitialized<T>()` with `mem::MaybeUninit<T>`
See [the Rust 1.36 announcement](https://blog.rust-lang.org/2019/07/04/Rust-1.36.0.html#maybeuninitt%3E-instead-of-mem::uninitialized) concerning this change. `uninitialized` will be deprecated in **1.38**, which means nightly already generates deprecation warnings. (CI tests nightly too.)

Fixing this should be easy enough (hint: "good first issue"), by running `rg uninitialized` and replacing all found uses with [`MaybeUninit`](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html) (has nice examples showing equivalent code for `uninitialized` and `MaybeUninit`).

murarth:
Closes #1025

### DIFF
--- a/src/platform_impl/linux/mod.rs
+++ b/src/platform_impl/linux/mod.rs
@@ -1,6 +1,6 @@
 #![cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "netbsd", target_os = "openbsd"))]
 
-use std::{collections::VecDeque, env, ffi::CStr, fmt, mem, os::raw::*, sync::Arc};
+use std::{collections::VecDeque, env, ffi::CStr, fmt, mem::MaybeUninit, os::raw::*, sync::Arc};
 
 use parking_lot::Mutex;
 use smithay_client_toolkit::reexports::client::ConnectError;
@@ -410,7 +410,7 @@ unsafe extern "C" fn x_error_callback(
 ) -> c_int {
     let xconn_lock = X11_BACKEND.lock();
     if let Ok(ref xconn) = *xconn_lock {
-        let mut buf: [c_char; 1024] = mem::uninitialized();
+        let mut buf: [c_char; 1024] = MaybeUninit::uninit().assume_init();
         (xconn.xlib.XGetErrorText)(
             display,
             (*event).error_code as c_int,
--- a/src/platform_impl/linux/mod.rs
+++ b/src/platform_impl/linux/mod.rs
@@ -410,14 +410,16 @@ unsafe extern "C" fn x_error_callback(
 ) -> c_int {
     let xconn_lock = X11_BACKEND.lock();
     if let Ok(ref xconn) = *xconn_lock {
-        let mut buf: [c_char; 1024] = MaybeUninit::uninit().assume_init();
+        // `assume_init` is safe here because the array consists of `MaybeUninit` values,
+        // which do not require initialization.
+        let mut buf: [MaybeUninit<c_char>; 1024] = MaybeUninit::uninit().assume_init();
         (xconn.xlib.XGetErrorText)(
             display,
             (*event).error_code as c_int,
-            buf.as_mut_ptr(),
+            buf.as_mut_ptr() as *mut c_char,
             buf.len() as c_int,
         );
-        let description = CStr::from_ptr(buf.as_ptr()).to_string_lossy();
+        let description = CStr::from_ptr(buf.as_ptr() as *const c_char).to_string_lossy();
 
         let error = XError {
             description: description.into_owned(),
--- a/src/platform_impl/linux/x11/mod.rs
+++ b/src/platform_impl/linux/x11/mod.rs
@@ -20,7 +20,7 @@ use std::{
     cell::RefCell,
     collections::{HashMap, HashSet, VecDeque},
     ffi::CStr,
-    mem,
+    mem::{self, MaybeUninit},
     ops::Deref,
     os::raw::*,
     rc::Rc,
@@ -100,11 +100,7 @@ impl<T: 'static> EventLoop<T> {
             .expect("Failed to query XRandR extension");
 
         let xi2ext = unsafe {
-            let mut result = XExtension {
-                opcode: mem::uninitialized(),
-                first_event_id: mem::uninitialized(),
-                first_error_id: mem::uninitialized(),
-            };
+            let mut result: XExtension = MaybeUninit::uninit().assume_init();
             let res = (xconn.xlib.XQueryExtension)(
                 xconn.display,
                 b"XInputExtension\0".as_ptr() as *const c_char,
@@ -204,8 +200,9 @@ impl<T: 'static> EventLoop<T> {
                 move |evt, &mut ()| {
                     if evt.readiness.is_readable() {
                         // process all pending events
-                        let mut xev = unsafe { mem::uninitialized() };
-                        while unsafe { processor.poll_one_event(&mut xev) } {
+                        let mut xev = MaybeUninit::uninit();
+                        while unsafe { processor.poll_one_event(xev.as_mut_ptr()) } {
+                            let mut xev = unsafe { xev.assume_init() };
                             processor.process_event(&mut xev, &mut callback);
                         }
                     }
@@ -395,9 +392,10 @@ struct DeviceInfo<'a> {
 impl<'a> DeviceInfo<'a> {
     fn get(xconn: &'a XConnection, device: c_int) -> Option<Self> {
         unsafe {
-            let mut count = mem::uninitialized();
-            let info = (xconn.xinput2.XIQueryDevice)(xconn.display, device, &mut count);
+            let mut count = MaybeUninit::uninit();
+            let info = (xconn.xinput2.XIQueryDevice)(xconn.display, device, count.as_mut_ptr());
             xconn.check_errors().ok().and_then(|_| {
+                let count = count.assume_init();
                 if info.is_null() || count == 0 {
                     None
                 } else {
--- a/src/platform_impl/linux/x11/mod.rs
+++ b/src/platform_impl/linux/x11/mod.rs
@@ -100,27 +100,21 @@ impl<T: 'static> EventLoop<T> {
             .expect("Failed to query XRandR extension");
 
         let xi2ext = unsafe {
-            let mut opcode = MaybeUninit::uninit();
-            let mut first_event_id = MaybeUninit::uninit();
-            let mut first_error_id = MaybeUninit::uninit();
+            let mut ext = XExtension::default();
 
             let res = (xconn.xlib.XQueryExtension)(
                 xconn.display,
                 b"XInputExtension\0".as_ptr() as *const c_char,
-                opcode.as_mut_ptr(),
-                first_event_id.as_mut_ptr(),
-                first_error_id.as_mut_ptr(),
+                &mut ext.opcode,
+                &mut ext.first_event_id,
+                &mut ext.first_error_id,
             );
 
             if res == ffi::False {
                 panic!("X server missing XInput extension");
             }
 
-            XExtension {
-                opcode: opcode.assume_init(),
-                first_event_id: first_event_id.assume_init(),
-                first_error_id: first_error_id.assume_init(),
-            }
+            ext
         };
 
         unsafe {
@@ -401,11 +395,9 @@ struct DeviceInfo<'a> {
 impl<'a> DeviceInfo<'a> {
     fn get(xconn: &'a XConnection, device: c_int) -> Option<Self> {
         unsafe {
-            let mut count = MaybeUninit::uninit();
-            let info = (xconn.xinput2.XIQueryDevice)(xconn.display, device, count.as_mut_ptr());
+            let mut count = 0;
+            let info = (xconn.xinput2.XIQueryDevice)(xconn.display, device, &mut count);
             xconn.check_errors().ok()?;
-
-            let count = count.assume_init();
 
             if info.is_null() || count == 0 {
                 None
@@ -520,7 +512,7 @@ impl<'a> Drop for GenericEventCookie<'a> {
     }
 }
 
-#[derive(Debug, Copy, Clone)]
+#[derive(Debug, Default, Copy, Clone)]
 struct XExtension {
     opcode: c_int,
     first_event_id: c_int,
--- a/src/platform_impl/linux/x11/mod.rs
+++ b/src/platform_impl/linux/x11/mod.rs
@@ -100,18 +100,27 @@ impl<T: 'static> EventLoop<T> {
             .expect("Failed to query XRandR extension");
 
         let xi2ext = unsafe {
-            let mut result: XExtension = MaybeUninit::uninit().assume_init();
+            let mut opcode = MaybeUninit::uninit();
+            let mut first_event_id = MaybeUninit::uninit();
+            let mut first_error_id = MaybeUninit::uninit();
+
             let res = (xconn.xlib.XQueryExtension)(
                 xconn.display,
                 b"XInputExtension\0".as_ptr() as *const c_char,
-                &mut result.opcode as *mut c_int,
-                &mut result.first_event_id as *mut c_int,
-                &mut result.first_error_id as *mut c_int,
+                opcode.as_mut_ptr(),
+                first_event_id.as_mut_ptr(),
+                first_error_id.as_mut_ptr(),
             );
+
             if res == ffi::False {
                 panic!("X server missing XInput extension");
             }
-            result
+
+            XExtension {
+                opcode: opcode.assume_init(),
+                first_event_id: first_event_id.assume_init(),
+                first_error_id: first_error_id.assume_init(),
+            }
         };
 
         unsafe {
@@ -394,18 +403,19 @@ impl<'a> DeviceInfo<'a> {
         unsafe {
             let mut count = MaybeUninit::uninit();
             let info = (xconn.xinput2.XIQueryDevice)(xconn.display, device, count.as_mut_ptr());
-            xconn.check_errors().ok().and_then(|_| {
-                let count = count.assume_init();
-                if info.is_null() || count == 0 {
-                    None
-                } else {
-                    Some(DeviceInfo {
-                        xconn,
-                        info,
-                        count: count as usize,
-                    })
-                }
-            })
+            xconn.check_errors().ok()?;
+
+            let count = count.assume_init();
+
+            if info.is_null() || count == 0 {
+                None
+            } else {
+                Some(DeviceInfo {
+                    xconn,
+                    info,
+                    count: count as usize,
+                })
+            }
         }
     }
 }
--- a/src/platform_impl/linux/x11/util/client_msg.rs
+++ b/src/platform_impl/linux/x11/util/client_msg.rs
@@ -30,7 +30,7 @@ impl XConnection {
         event_mask: Option<c_long>,
         data: ClientMsgPayload,
     ) -> Flusher<'_> {
-        let mut event: ffi::XClientMessageEvent = unsafe { mem::uninitialized() };
+        let mut event: ffi::XClientMessageEvent = unsafe { MaybeUninit::uninit().assume_init() };
         event.type_ = ffi::ClientMessage;
         event.display = self.display;
         event.window = window;
@@ -54,7 +54,7 @@ impl XConnection {
         let format = T::FORMAT;
         let size_of_t = mem::size_of::<T>();
         debug_assert_eq!(size_of_t, format.get_actual_size());
-        let mut event: ffi::XClientMessageEvent = unsafe { mem::uninitialized() };
+        let mut event: ffi::XClientMessageEvent = unsafe { MaybeUninit::uninit().assume_init() };
         event.type_ = ffi::ClientMessage;
         event.display = self.display;
         event.window = window;
--- a/src/platform_impl/linux/x11/util/client_msg.rs
+++ b/src/platform_impl/linux/x11/util/client_msg.rs
@@ -30,13 +30,17 @@ impl XConnection {
         event_mask: Option<c_long>,
         data: ClientMsgPayload,
     ) -> Flusher<'_> {
-        let mut event: ffi::XClientMessageEvent = unsafe { MaybeUninit::uninit().assume_init() };
-        event.type_ = ffi::ClientMessage;
-        event.display = self.display;
-        event.window = window;
-        event.message_type = message_type;
-        event.format = c_long::FORMAT as c_int;
-        event.data = unsafe { mem::transmute(data) };
+        let event = ffi::XClientMessageEvent {
+            type_: ffi::ClientMessage,
+            display: self.display,
+            window,
+            message_type,
+            format: c_long::FORMAT as c_int,
+            data: unsafe { mem::transmute(data) },
+            // These fields are ignored by `XSendEvent`
+            serial: 0,
+            send_event: 0,
+        };
         self.send_event(target_window, event_mask, event)
     }
 
@@ -54,12 +58,17 @@ impl XConnection {
         let format = T::FORMAT;
         let size_of_t = mem::size_of::<T>();
         debug_assert_eq!(size_of_t, format.get_actual_size());
-        let mut event: ffi::XClientMessageEvent = unsafe { MaybeUninit::uninit().assume_init() };
-        event.type_ = ffi::ClientMessage;
-        event.display = self.display;
-        event.window = window;
-        event.message_type = message_type;
-        event.format = format as c_int;
+        let mut event = ffi::XClientMessageEvent {
+            type_: ffi::ClientMessage,
+            display: self.display,
+            window,
+            message_type,
+            format: format as c_int,
+            data: ffi::ClientMessageData::new(),
+            // These fields are ignored by `XSendEvent`
+            serial: 0,
+            send_event: 0,
+        };
 
         let t_per_payload = format.get_payload_size() / size_of_t;
         assert!(t_per_payload > 0);
--- a/src/platform_impl/linux/x11/util/geometry.rs
+++ b/src/platform_impl/linux/x11/util/geometry.rs
@@ -183,9 +183,7 @@ impl XConnection {
         window: ffi::Window,
         root: ffi::Window,
     ) -> Result<TranslatedCoords, XError> {
-        let mut x_rel_root = MaybeUninit::uninit();
-        let mut y_rel_root = MaybeUninit::uninit();
-        let mut child = MaybeUninit::uninit();
+        let mut coords: TranslatedCoords = unsafe { MaybeUninit::zeroed().assume_init() };
 
         unsafe {
             (self.xlib.XTranslateCoordinates)(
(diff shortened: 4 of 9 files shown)


murarth:
All uses of `mem::uninitialized` were in Xlib FFI code, so Linux X11 is the only platform affected by this change.

src/platform_impl/linux/x11/util/client_msg.rs
```diff
@@ -54,7 +54,7 @@ impl XConnection {
         let format = T::FORMAT;
         let size_of_t = mem::size_of::<T>();
         debug_assert_eq!(size_of_t, format.get_actual_size());
-        let mut event: ffi::XClientMessageEvent = unsafe { mem::uninitialized() };
+        let mut event: ffi::XClientMessageEvent = unsafe { MaybeUninit::uninit().assume_init() };
```
ZeGentzy: This causes UB, according to the rust docs:
> It is up to the caller to guarantee that the MaybeUninit<T> really is in an initialized state. Calling this when the content is not yet fully initialized causes immediate undefined behavior.

src/platform_impl/linux/x11/util/geometry.rs
```diff
@@ -183,7 +183,8 @@ impl XConnection {
         window: ffi::Window,
         root: ffi::Window,
     ) -> Result<TranslatedCoords, XError> {
-        let mut translated_coords: TranslatedCoords = unsafe { mem::uninitialized() };
+        let mut translated_coords: TranslatedCoords =
+            unsafe { MaybeUninit::uninit().assume_init() };
```
ZeGentzy: This causes UB, according to the rust docs:
> It is up to the caller to guarantee that the MaybeUninit<T> really is in an initialized state. Calling this when the content is not yet fully initialized causes immediate undefined behavior.

src/platform_impl/linux/x11/util/geometry.rs
```diff
@@ -202,7 +203,7 @@ impl XConnection {
 
     // This is adequate for inner_size
     pub fn get_geometry(&self, window: ffi::Window) -> Result<Geometry, XError> {
-        let mut geometry: Geometry = unsafe { mem::uninitialized() };
+        let mut geometry: Geometry = unsafe { MaybeUninit::uninit().assume_init() };
```
ZeGentzy: This causes UB, according to the rust docs:
> It is up to the caller to guarantee that the MaybeUninit<T> really is in an initialized state. Calling this when the content is not yet fully initialized causes immediate undefined behavior.

src/platform_impl/linux/x11/mod.rs
```diff
@@ -395,9 +392,10 @@ struct DeviceInfo<'a> {
 impl<'a> DeviceInfo<'a> {
     fn get(xconn: &'a XConnection, device: c_int) -> Option<Self> {
         unsafe {
-            let mut count = mem::uninitialized();
-            let info = (xconn.xinput2.XIQueryDevice)(xconn.display, device, &mut count);
+            let mut count = MaybeUninit::uninit();
+            let info = (xconn.xinput2.XIQueryDevice)(xconn.display, device, count.as_mut_ptr());
             xconn.check_errors().ok().and_then(|_| {
+                let count = count.assume_init();
```
ZeGentzy: Is this copying `count`? Can we put the `let count = count.assume_init();` outside the lambda just to stay on the safe side?

src/platform_impl/linux/x11/util/geometry.rs
```diff
@@ -264,27 +265,27 @@ impl XConnection {
...
+                nchildren.as_mut_ptr(),
             );
 
             // The list of children isn't used
             if children != ptr::null_mut() {
                 (self.xlib.XFree)(children as *mut _);
             }
 
-            parent
+            parent.assume_init()
```
ZeGentzy: Afaik, `assume_init` should be called on `root` and `nchildren` so that `Drop` is called on them, else we got a memory leak.

src/platform_impl/linux/x11/util/hint.rs
```diff
@@ -317,13 +317,13 @@ impl XConnection {
...
         let size_hints = self.alloc_size_hints();
-        let mut supplied_by_user: c_long = unsafe { mem::uninitialized() };
+        let mut supplied_by_user = MaybeUninit::uninit();
         unsafe {
             (self.xlib.XGetWMNormalHints)(
                 self.display,
                 window,
                 size_hints.ptr,
-                &mut supplied_by_user,
+                supplied_by_user.as_mut_ptr(),
```
ZeGentzy: Afaik, `assume_init` should be called on `supplied_by_user` so that `Drop` is called on it, else we got a memory leak.

src/platform_impl/linux/x11/util/input.rs
```diff
@@ -93,7 +93,7 @@ impl XConnection {
         device_id: c_int,
     ) -> Result<PointerState<'_>, XError> {
         unsafe {
-            let mut pointer_state: PointerState<'_> = mem::uninitialized();
+            let mut pointer_state: PointerState<'_> = MaybeUninit::uninit().assume_init();
```
ZeGentzy: This causes UB, according to the rust docs:
> It is up to the caller to guarantee that the MaybeUninit<T> really is in an initialized state. Calling this when the content is not yet fully initialized causes immediate undefined behavior.

src/platform_impl/linux/x11/util/input.rs
```diff
@@ -141,7 +141,7 @@ impl XConnection {
     }
 
     pub fn lookup_utf8(&self, ic: ffi::XIC, key_event: &mut ffi::XKeyEvent) -> String {
-        let mut buffer: [u8; TEXT_BUFFER_SIZE] = unsafe { mem::uninitialized() };
+        let mut buffer: [u8; TEXT_BUFFER_SIZE] = unsafe { MaybeUninit::uninit().assume_init() };
```
ZeGentzy: This causes UB, according to the rust docs:
> It is up to the caller to guarantee that the MaybeUninit<T> really is in an initialized state. Calling this when the content is not yet fully initialized causes immediate undefined behavior.

src/platform_impl/linux/x11/util/window_property.rs
```diff
@@ -45,10 +45,10 @@ impl XConnection {
...
         while !done {
             unsafe {
-                let mut actual_type: ffi::Atom = mem::uninitialized();
-                let mut actual_format: c_int = mem::uninitialized();
-                let mut quantity_returned: c_ulong = mem::uninitialized();
-                let mut bytes_after: c_ulong = mem::uninitialized();
+                let mut actual_type: ffi::Atom = MaybeUninit::uninit().assume_init();
+                let mut actual_format: c_int = MaybeUninit::uninit().assume_init();
+                let mut quantity_returned: c_ulong = MaybeUninit::uninit().assume_init();
+                let mut bytes_after: c_ulong = MaybeUninit::uninit().assume_init();
```
ZeGentzy: This causes UB, according to the rust docs:
> It is up to the caller to guarantee that the MaybeUninit<T> really is in an initialized state. Calling this when the content is not yet fully initialized causes immediate undefined behavior.

src/platform_impl/linux/x11/window.rs
```diff
@@ -453,7 +462,7 @@ impl UnownedWindow {
                 // 64 would suffice for Linux, but 256 will be enough everywhere (as per SUSv2). For instance, this is
                 // the limit defined by OpenBSD.
                 const MAXHOSTNAMELEN: usize = 256;
-                let mut hostname: [c_char; MAXHOSTNAMELEN] = mem::uninitialized();
+                let mut hostname: [c_char; MAXHOSTNAMELEN] = MaybeUninit::uninit().assume_init();
```
ZeGentzy: This causes UB, according to the rust docs:
> It is up to the caller to guarantee that the MaybeUninit<T> really is in an initialized state. Calling this when the content is not yet fully initialized causes immediate undefined behavior.

src/platform_impl/linux/x11/window.rs
```diff
@@ -410,11 +419,11 @@ impl UnownedWindow {
                 unsafe {
                     // XSetInputFocus generates an error if the window is not visible, so we wait
                     // until we receive VisibilityNotify.
-                    let mut event = mem::uninitialized();
+                    let mut event = MaybeUninit::uninit();
                     (xconn.xlib.XIfEvent)(
                         // This will flush the request buffer IF it blocks.
                         xconn.display,
-                        &mut event as *mut ffi::XEvent,
+                        event.as_mut_ptr(),
```
ZeGentzy: Afaik, `assume_init` should be called on `event` so that `Drop` is called on it, else we got a memory leak.

src/platform_impl/linux/x11/window.rs
```diff
@@ -1134,13 +1143,13 @@ impl UnownedWindow {
...
-            let dummy_color: ffi::XColor = mem::uninitialized();
+            let mut dummy_color = MaybeUninit::uninit();
             let cursor = (self.xconn.xlib.XCreatePixmapCursor)(
                 self.xconn.display,
                 pixmap,
                 pixmap,
-                &dummy_color as *const _ as *mut _,
-                &dummy_color as *const _ as *mut _,
+                dummy_color.as_mut_ptr(),
+                dummy_color.as_mut_ptr(),
```
ZeGentzy: Afaik, `assume_init` should be called on `dummy_color` so that `Drop` is called on it, else we got a memory leak.

ZeGentzy:
I'm not sure who thought `MaybeUninit` is better than `mem::uninitialized`... it seams easy to mess up, unlike the prior `mem::uninitialized` api...

murarth:
I was not aware that even an uninitialized `i32` is undefined behavior. I'll address those issues accordingly.

src/platform_impl/linux/x11/window.rs
murarth: Even though `dummy_color` is an [`XColor`](https://docs.rs/x11-dl/2.18.3/x11_dl/xlib/struct.XColor.html) instance, containing only plain old data?

felixrabe:
> ~I'm not sure who thought `MaybeUninit` is better than `mem::uninitialized`... it seams easy to mess up, unlike the prior `mem::uninitialized` api...~

You know you can "delete" a comment :)

src/platform_impl/linux/x11/util/input.rs
```diff
@@ -25,16 +25,16 @@ impl From<ffi::XIModifierState> for ModifiersState {
 
 pub struct PointerState<'a> {
```
murarth: It turns out that some of these private fields are unused. rustc wasn't detecting this before because the struct was being filled with `&mut pointer_state.field` references and this was seen as a use. I've commented out the fields as it should have no effect on the program. I could instead make them public or add `#[allow(dead_code)]` to the struct.

francesca64: I kept them in both 1) to make allocation easier, and 2) so that contributors can more easily discover/utilize functionality they might need (the idea is to provide a robust internal API). I'd prefer they be kept in, though with comments explaining the rationale.

murarth: Fair enough. I've restored the fields and marked them `pub` to silence the warnings.

murarth:
@ZeGentzy I believe I have resolved all of the undefined behavior issues.

aloucks:
There should probably be a more thorough audit of `mem::uninitialized`.

I think that most of these cases should either be replaced with `mem::zeroed()` (e.g. for native struct initialization) or simply use `0` or `ptr::null_mut()` to initialize integers and pointers. Arrays should probably just use `[0; the_size]` as well. Uninitialized memory is playing with fire and a recipe for UB. We should try to avoid `MaybeUninit` unless it's absolutely necessary.

francesca64:
@aloucks in what ways is `mem::zeroed` safer than `MaybeUninit`?

ZeGentzy:
Reading from unitialized memory triggers UB. Reading from zero-initialized memory only triggers UB if that is not a valid state. Since most (all?) of these variables are integers, accidentally reading from them if they get zeroed-initialized, instead of uninitialized, will help us avoid accidental UB.

aloucks:
> @aloucks in what ways is `mem::zeroed` safer than `MaybeUninit`?

In general, you should avoid `mem::zeroed()` too, but sometimes it's you're only option. (e.g. https://github.com/retep998/winapi-rs#how-do-i-create-an-instance-of-a-union). It's usually fine for initializing native structs (because they generally don't contain references). 

Why would you prefer `MaybeUninit` over initializing pointers and integers to zero/null?

aloucks:
> Reading from zero-initialized memory only triggers UB if that is not a valid state. 

I don't think you even have to read from it for it to be considered UB. The addition of `MaybeUninit` doesn't make uninitialized memory safe. It was just impossible to use `mem::uninitialized()` in some cases before without triggering UB (e.g. with bools). `MaybeUninit` provides a for _a way_ to to do it safely, but it's still [extremely easy to get it wrong](https://doc.rust-lang.org/beta/std/mem/union.MaybeUninit.html).

ZeGentzy:
@aloucks According to the docs:

> It depends on T whether that already makes for proper initialization. For example, MaybeUninit<usize>::zeroed() is initialized, but MaybeUninit<&'static i32>::zeroed() is not because references must not be null.

So, operations that would normally trigger UB on uninitialized memory will not trigger UB on zero-initialized types, if zero is a valid initialization of the type.

src/platform_impl/linux/x11/mod.rs
```diff
@@ -100,22 +100,27 @@ impl<T: 'static> EventLoop<T> {
             .expect("Failed to query XRandR extension");
 
         let xi2ext = unsafe {
-            let mut result = XExtension {
```
ZeGentzy: I think it should be safe to zero-initialize the whole struct, then pass pointers to the individual members in it.

src/platform_impl/linux/x11/util/geometry.rs
```diff
@@ -183,41 +183,71 @@ impl XConnection {
         window: ffi::Window,
         root: ffi::Window,
     ) -> Result<TranslatedCoords, XError> {
-        let mut translated_coords: TranslatedCoords = unsafe { mem::uninitialized() };
+        let mut x_rel_root = MaybeUninit::uninit();
```
ZeGentzy: I think it should be safe to zero-initialize the whole struct, then pass pointers to the individual members in it.

src/platform_impl/linux/x11/util/geometry.rs
```diff
@@ -183,41 +183,71 @@ impl XConnection {
...
+                y_rel_root: y_rel_root.assume_init(),
+                child: child.assume_init(),
+            })
+        }
     }
 
     // This is adequate for inner_size
     pub fn get_geometry(&self, window: ffi::Window) -> Result<Geometry, XError> {
-        let mut geometry: Geometry = unsafe { mem::uninitialized() };
+        let mut root = MaybeUninit::uninit();
```
ZeGentzy: I think it should be safe to zero-initialize the whole struct, then pass pointers to the individual members in it.

src/platform_impl/linux/x11/util/input.rs
```diff
@@ -93,46 +94,64 @@ impl XConnection {
         device_id: c_int,
     ) -> Result<PointerState<'_>, XError> {
         unsafe {
-            let mut pointer_state: PointerState<'_> = mem::uninitialized();
-            pointer_state.xconn = self;
-            pointer_state.relative_to_window = (self.xinput2.XIQueryPointer)(
+            let mut root = MaybeUninit::uninit();
```
ZeGentzy: I think it should be safe to zero-initialize the whole struct, then pass pointers to the individual members in it.

ZeGentzy:
> Why would you prefer MaybeUninit over initializing pointers and integers to zero/null?

I'd rather we zero initialize whole integer-only structs then manually set their contents to zero.

src/platform_impl/linux/x11/util/input.rs
murarth: `zeroed` is not safe for `PointerState`, as it contains a reference field, `&XConnection`.

I've instead zero-initialized each value before passing a reference to the FFI call.

aloucks:
> So, operations that would normally trigger UB on uninitialized memory will not trigger UB on zero-initialized types, if zero is a valid initialization of the type.

Yes, absolutely - that's why I was saying prefer it over uninitialized and prefer `0` or `ptr::null_mut()` for non-struct members.

src/platform_impl/linux/x11/util/input.rs
```diff
@@ -93,46 +94,64 @@ impl XConnection {
...
         unsafe {
-            let mut pointer_state: PointerState<'_> = mem::uninitialized();
-            pointer_state.xconn = self;
-            pointer_state.relative_to_window = (self.xinput2.XIQueryPointer)(
+            let mut root = 0;
+            let mut child = 0;
+            let mut root_x = 0.0;
+            let mut root_y = 0.0;
+            let mut win_x = 0.0;
+            let mut win_y = 0.0;
```
ZeGentzy: Can we do this instead:
```
let ps = PointerState {
                xconn: self,
                root: 0,
                child: 0,
                root_x: 0.0,
                root_y: 0.0,
                win_x: 0.0,
                win_y: 0.0,
                buttons: MaybeUninit::zeroed().assume_init(),
                modifiers: MaybeUninit::zeroed().assume_init(),
                group: MaybeUninit::zeroed().assume_init(),
                relative_to_window: false,
            };
```

src/platform_impl/linux/x11/util/input.rs
ZeGentzy: Afaik, if we set the `&XConnection` to self before calling `assume_init`, it should be safe.

It's the first example here: https://doc.rust-lang.org/nightly/std/mem/union.MaybeUninit.html#examples

aloucks: If `PointerState` contains a reference, `MaybeUninit` is probably the correct choice as long as the reference is filled with valid data before `assume_init()`. Apologies if my comments were misunderstood. The point I've been trying to make is that uninitialized data shouldn't be used as a blanket solution - it does have some valid cases. You just have to be extra careful - it's a footgun waiting to go off.

src/platform_impl/linux/x11/mod.rs
```diff
@@ -100,22 +100,27 @@ impl<T: 'static> EventLoop<T> {
...
-                opcode: mem::uninitialized(),
-                first_event_id: mem::uninitialized(),
-                first_error_id: mem::uninitialized(),
-            };
+            let mut opcode = MaybeUninit::uninit();
```
ZeGentzy: I think it should be safe to zero-initialize the whole struct, then pass pointers to the individual members in it.

aloucks:
`MaybeUninit::zeroed().assume_init()` is no different than `mem::zeroed()`.

murarth:
> `MaybeUninit::zeroed().assume_init()` is no different than `mem::zeroed()`.

Yes, but since `mem::uninitialized` has been deprecated, `mem::zeroed` may also be deprecated in the future and using `MaybeUninit::zeroed` now avoids making another change in the future.

That said, if brevity is preferred, I'm not opposed to changing it.

aloucks:
What I had originally meant by auditing the use of `mem::uninitialized()` is that many of these cases don't require it at all and can actually be expressed with safe code.

For example, `XConnection::query_pointer` can be implemented without it and the only `unsafe` part is where it calls the `xinput2.XIQueryPointer` ffi function.

```rust
pub fn query_pointer(
    &self,
    window: ffi::Window,
    device_id: c_int,
) -> Result<PointerState<'_>, XError> {
    let mut root = 0;
    let mut child = 0;
    let mut root_x = 0.0;
    let mut root_y = 0.0;
    let mut win_x = 0.0;
    let mut win_y = 0.0;
    let mut group = Default::default();
    let mut buttons = Default::default();
    let mut modifiers = Default::default();

    let relative_to_window = unsafe { (self.xinput2.XIQueryPointer)(
        self.display,
        device_id,
        window,
        &mut root,
        &mut child,
        &mut root_x,
        &mut root_y,
        &mut win_x,
        &mut win_y,
        &mut buttons,
        &mut modifiers,
        &mut group,
    ) == ffi::True };

    if let Err(err) = self.check_errors() {
        Err(err)
    } else {
        Ok(PointerState {
            xconn: self,
            root,
            child,
            root_x,
            root_y,
            win_x,
            relative_to_window,
            buttons,
            modifiers,
            win_y,
            group
        })
    }
}
```

Interestingly enough now that we're creating that struct with 100% safe code we find that most of those fields aren't even being used.

```
warning: field is never used: `root`
  --> src/platform_impl/linux/x11/util/input.rs:28:5
   |
28 |     root: ffi::Window,
   |     ^^^^^^^^^^^^^^^^^
   |
   = note: #[warn(dead_code)] on by default

warning: field is never used: `child`
  --> src/platform_impl/linux/x11/util/input.rs:29:5
   |
29 |     child: ffi::Window,
   |     ^^^^^^^^^^^^^^^^^^

warning: field is never used: `win_x`
  --> src/platform_impl/linux/x11/util/input.rs:32:5
   |
32 |     win_x: c_double,
   |     ^^^^^^^^^^^^^^^

warning: field is never used: `win_y`
  --> src/platform_impl/linux/x11/util/input.rs:33:5
   |
33 |     win_y: c_double,
   |     ^^^^^^^^^^^^^^^

warning: field is never used: `group`
  --> src/platform_impl/linux/x11/util/input.rs:36:5
   |
36 |     group: ffi::XIGroupState,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^

warning: field is never used: `relative_to_window`
  --> src/platform_impl/linux/x11/util/input.rs:37:5
   |
37 |     relative_to_window: bool,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^

```

src/platform_impl/linux/x11/util/input.rs
murarth: But the field cannot be accessed without, at least, dereferencing a raw pointer to a mutable reference.

The documentation of `MaybeUninit` does not (to my knowledge) explicitly state that this would be undefined; however, the section on [field-by-field initialization](https://doc.rust-lang.org/nightly/std/mem/union.MaybeUninit.html#initializing-a-struct-field-by-field) states that initialization in this way is not possible. So, I take that to mean that it is undefined behavior to create a `&mut T` for uninitialized data.

aloucks: @murarth is correct, it's undefined behavior. I left an example below of how this specific case can be solved without any usage of uninitialized memory. 

> There is currently no supported way to create a raw pointer or reference to a field of a struct inside MaybeUninit<Struct>. That means it is not possible to create a struct by calling MaybeUninit::uninit::<Struct>() and then writing to its fields.

murarth:
I've pushed a commit reforming usage of `MaybeUninit`, following these guidelines:

* Prefer safe zero-initialization using `Default`, when possible
* Zero-initialize integers and floats using `0` or `0.0`
* Use `MaybeUninit::uninit` for large byte buffers and union types
* Use `MaybeUninit::uninit` when the resulting value is ignored

src/platform_impl/linux/mod.rs
```diff
@@ -410,14 +410,16 @@ unsafe extern "C" fn x_error_callback(
 ) -> c_int {
     let xconn_lock = X11_BACKEND.lock();
     if let Ok(ref xconn) = *xconn_lock {
-        let mut buf: [c_char; 1024] = mem::uninitialized();
+        // `assume_init` is safe here because the array consists of `MaybeUninit` values,
+        // which do not require initialization.
```
aloucks: This could be replaced with

```rust
let mut buf = [0 as c_char; 1024];
```

src/platform_impl/linux/x11/util/hint.rs
```diff
@@ -317,13 +317,13 @@ impl XConnection {
 
     pub fn get_normal_hints(&self, window: ffi::Window) -> Result<NormalHints<'_>, XError> {
         let size_hints = self.alloc_size_hints();
-        let mut supplied_by_user: c_long = unsafe { mem::uninitialized() };
+        let mut supplied_by_user = MaybeUninit::uninit();
```
aloucks: This is just an integer. There is no need for uninit.

You can just set it normally (and change the param back to `&mut supplied_by_user` below:

```rust
let mut supplied_by_user = 0;
```

src/platform_impl/linux/x11/util/input.rs
```diff
@@ -141,21 +160,28 @@ impl XConnection {
...
-        let mut buffer: [u8; TEXT_BUFFER_SIZE] = unsafe { mem::uninitialized() };
-        let (_, status, count) = self.lookup_utf8_inner(ic, key_event, &mut buffer);
-        // The buffer overflowed, so we'll make a new one on the heap.
-        if status == ffi::XBufferOverflow {
-            let mut buffer = Vec::with_capacity(count as usize);
-            unsafe { buffer.set_len(count as usize) };
-            let (_, _, new_count) = self.lookup_utf8_inner(ic, key_event, &mut buffer);
+        // `assume_init` is safe here because the array consists of `MaybeUninit` values,
+        // which do not require initialization.
+        let mut buffer: [MaybeUninit<u8>; TEXT_BUFFER_SIZE] =
```
aloucks: This could be replaced with

```rust
let mut buffer = [0_u8; TEXT_BUFFER_SIZE];
```

src/platform_impl/linux/x11/window.rs
```diff
@@ -449,19 +459,22 @@ impl UnownedWindow {
         let pid_atom = unsafe { self.xconn.get_atom_unchecked(b"_NET_WM_PID\0") };
         let client_machine_atom = unsafe { self.xconn.get_atom_unchecked(b"WM_CLIENT_MACHINE\0") };
         unsafe {
-            let (hostname, hostname_length) = {
-                // 64 would suffice for Linux, but 256 will be enough everywhere (as per SUSv2). For instance, this is
-                // the limit defined by OpenBSD.
-                const MAXHOSTNAMELEN: usize = 256;
-                let mut hostname: [c_char; MAXHOSTNAMELEN] = mem::uninitialized();
```
aloucks: This could be replaced with

```rust
let mut hostname = [0 as c_char; MAXHOSTNAMELEN];
```

src/platform_impl/linux/x11/window.rs
```diff
@@ -1134,13 +1147,13 @@ impl UnownedWindow {
         let cursor = unsafe {
             // We don't care about this color, since it only fills bytes
             // in the pixmap which are not 0 in the mask.
-            let dummy_color: ffi::XColor = mem::uninitialized();
```
aloucks: Why not `mem::zeroed()` and be done with it?

src/platform_impl/linux/x11/window.rs
```diff
@@ -410,11 +420,11 @@ impl UnownedWindow {
                 unsafe {
                     // XSetInputFocus generates an error if the window is not visible, so we wait
                     // until we receive VisibilityNotify.
-                    let mut event = mem::uninitialized();
```
aloucks: Why not `mem::zeroed()` and be done with it?

src/platform_impl/linux/x11/util/hint.rs
murarth: The value is ignored. There's no `unsafe`. There's no undefined behavior.

Why zero-initialize a variable that is never read?

src/platform_impl/linux/mod.rs
murarth: Why zero-initialize a KB of memory? If we can't trust a C library to copy data into a buffer and accurately report the number of bytes written, why are we using C libraries at all?

src/platform_impl/linux/x11/window.rs
murarth: `XEvent` has no `Default` impl. It's a union type with a size of 192 bytes.

src/platform_impl/linux/x11/window.rs
murarth: Same as above, the value is not used.

murarth:
Apologies if I've missed something. Are there any unresolved issues with this PR?

PR #1031, opened 2019-07-09
Add NewEvents(Init) callback to x11

rikusalminen:
On X11, `Event::NewEvents(StartCause::Init)` was not delivered as first event.

̃~~I added a test case to run on an event loop without any windows attached and check that first event is `NewEvents(Init)`, then `ControlFlow::Exit` and check that last event is `LoopDestroyed`.~~

I have ran this test on Linux/x11 platform only.
**Help wanted**: please run the new test on macos and windows.

### DIFF
--- a/src/platform_impl/linux/x11/mod.rs
+++ b/src/platform_impl/linux/x11/mod.rs
@@ -249,6 +249,12 @@ impl<T: 'static> EventLoop<T> {
         let mut control_flow = ControlFlow::default();
         let wt = get_xtarget(&self.target);
 
+        callback(
+            crate::event::Event::NewEvents(crate::event::StartCause::Init),
+            &self.target,
+            &mut control_flow,
+        );
+
         loop {
             // Empty the event buffer
             {


felixrabe:
Thx for this.

You didn't actually "create or update an example program", so you can remove that line from the description.

felixrabe:
Also, you can try to work from the failing Travis builds to see if you can fix them, if they look straightforward to you.

felixrabe:
Based on this output

```
failures:
---- event_loop_wait stdout ----
thread 'event_loop_wait' panicked at 'On macOS, `EventLoop` must be created on the main thread!', src/platform_impl/macos/event_loop.rs:45:17
---- event_loop_basic stdout ----
thread 'event_loop_basic' panicked at 'On macOS, `EventLoop` must be created on the main thread!', src/platform_impl/macos/event_loop.rs:45:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

it looks like tests are not run on the main thread. I don't know for sure, but perhaps Rust tests are run in multiple threads for speed ups. Maybe there is an option to get them to run on a single (main) thread instead?

@rikusalminen Can you insert an assertion into the tests you added to make sure they are run from the main thread?

rikusalminen:
Unfortunately it does not seem like there's a way to make `cargo test` use the main thread without some hacks (this is the best I can come up with: https://stackoverflow.com/questions/43458194/is-there-any-way-to-tell-cargo-to-run-its-tests-on-the-main-thread)

@felixrabe please advice what to try next? Make the test(s) skip on macos?

In any case, the actual fix patch can be merged (or cherry-picked) while thinking about a solution for running tests on macos.

Are there plans for creating tests that actually use the windowing API?

rikusalminen:
I also noticed that Linux travis tests will fail because no backends can be initialized.

felixrabe:
I suggest you revert the tests (or force-push on a branch that does not have tests ; also update the title of this PR), and open a separate PR for just the tests. Then we can merge this change (if the project owners think it is good), and discuss how to approach testing in the new PR.

rikusalminen:
I have removed the test case from this pull request.

I'll consider what to do with it, maybe it could be put on the examples folder instead of automatic testing.

Osspial:
@rikusalminen There's been talk at various points of creating an interactive test suite since a large portion of Winit is hard to automatically test, so it may make sense to put this test in there if somebody starts work to create it.

PR #1033, opened 2019-07-09
Fix RequestRedraw ordering issues when ControlFlow is set to Poll

### LINKED ISSUE #1008: On Windows, when control_flow is set to Poll calling request_redraw in EventsCleared doesn't dispatch RedrawRequested immediately
This can be reproduced with this program:

```rust
use winit::{
    event::{Event, WindowEvent},
    window::WindowBuilder,
};

fn main() {

        .build(&event_loop)
        .unwrap();

        println!("{:?}", event);

        match event {
            Event::WindowEvent {
                window_id,
            Event::EventsCleared => {
                window.request_redraw();
            }
            _ => *control_flow = ControlFlow::Poll,
        }
    });
}
```

`RequestRedraw` should be dispatched between `EventsCleared` and `NewEvents`, but it isn't.

Osspial:

This PR still has some major issues, and should not be merged in its current state. The main issue I'm aware of is that resizing the window completely breaks the event loop, but I'll look into fixing that.

Fixes #1008.

### DIFF
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -12,11 +12,10 @@
 //! The closure passed to the `execute_in_thread` method takes an `Inserter` that you can use to
 //! add a `WindowState` entry to a list of window to be used by the callback.
 
+mod runner;
+
 use parking_lot::Mutex;
 use std::{
-    any::Any,
-    cell::RefCell,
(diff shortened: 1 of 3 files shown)


Osspial:
Closing for now, since fixing this bug properly is incredibly difficult with the current API.

PR #1040, opened 2019-07-10
Fix `web` errors

hecrj:
I have fixed some issues I found with the new `web` branch. It should compile now!


Osspial:
Merging. Thank you for fixing this!

PR #1042, opened 2019-07-11
Touch events emit screen coordinates instead of client coordinates on Windows

### LINKED ISSUE #1002: Touch events emit screen coordinates instead of client coordinates on Windows
As reported in issue #975 the location provided by WM_TOUCH and WM_POINTER events are screen coordinates instead of window client coordinates. These events should emit client coordinates for consistency with the rest of the API and other backends.

This is a change that will break clients using the touch events on Windows though.

dam4rus:
Fixes #1002

I think this change should be reflected in CHANGELOG.md. Should i add it under "Unreleased"?

### DIFF
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -1457,8 +1457,8 @@ unsafe extern "system" fn public_window_callback<T>(
                         continue;
                     }
 
-                    let x = location.x as f64;
-                    let y = location.y as f64;
+                    let x = location.x as f64 + (input.x % 100) as f64 / 100f64;
+                    let y = location.y as f64 + (input.y % 100) as f64 / 100f64;
                     let location = LogicalPosition::from_physical((x, y), dpi_factor);
                     subclass_input.send_event(Event::WindowEvent {
                         window_id: RootWindowId(WindowId(window)),
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -1448,8 +1448,17 @@ unsafe extern "system" fn public_window_callback<T>(
             {
                 let dpi_factor = hwnd_scale_factor(window);
                 for input in &inputs {
-                    let x = (input.x as f64) / 100f64;
-                    let y = (input.y as f64) / 100f64;
+                    let mut location = POINT {
+                        x: input.x / 100,
+                        y: input.y / 100,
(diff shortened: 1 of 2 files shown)


src/platform_impl/windows/event_loop.rs
```diff
@@ -1448,8 +1448,17 @@ unsafe extern "system" fn public_window_callback<T>(
             {
                 let dpi_factor = hwnd_scale_factor(window);
                 for input in &inputs {
-                    let x = (input.x as f64) / 100f64;
-                    let y = (input.y as f64) / 100f64;
+                    let mut location = POINT {
+                        x: input.x / 100,
+                        y: input.y / 100,
+                    };
```
Osspial: Doing the conversion like this loses the subpixel precision that the windows API provides. It should be possible to do the conversion to client-space and bring over the remainder.

src/platform_impl/windows/event_loop.rs
```diff
@@ -1510,8 +1519,17 @@ unsafe extern "system" fn public_window_callback<T>(
                 // The information retrieved appears in reverse chronological order, with the most recent entry in the first
                 // row of the returned array
                 for pointer_info in pointer_infos.iter().rev() {
-                    let x = pointer_info.ptPixelLocation.x as f64;
-                    let y = pointer_info.ptPixelLocation.y as f64;
+                    let mut location = POINT {
+                        x: pointer_info.ptPixelLocation.x,
+                        y: pointer_info.ptPixelLocation.y,
+                    };
```
Osspial: This may be better-suited to another PR, but would it be possible to use `ptHimetricLocation` to get sub-pixel precision when handling this event? I should've caught that in the original PR, but I didn't look closely enough and didn't realize it was quite possible.

Osspial:
> I think this change should be reflected in CHANGELOG.md. Should i add it under "Unreleased"?

Yep!

src/platform_impl/windows/event_loop.rs
dam4rus: Sure, i will look at it. I can open a new PR if you wish.

Osspial: Do whatever works best for you. Either option is fine by me.

dam4rus: I've implemented it in a separate branch so i'll open a separate PR for it.

Issue #159, opened 2017-03-23
Support for creating child windows


Boscop:
It would be very useful to have support for creating child windows with a given parent handle.
I implemented support for child windows on Windows in my glutin fork like 8 months ago:
https://github.com/Boscop/glutin/blob/master/src/window.rs#L205
https://github.com/Boscop/glutin/blob/master/src/api/win32/init.rs#L76-L84
https://github.com/Boscop/glutin/blob/master/src/api/win32/callback.rs#L23
https://github.com/Boscop/glutin/blob/master/src/api/win32/mod.rs#L471-L473
(Child windows have to run in the same thread as the parent window, so I'm using a HashMap with HWND as key for event dispatching of all the windows in the same thread.)

It would also be useful to have support for timers on the window.
Either via callback:
https://github.com/Boscop/glutin/blob/master/src/window.rs#L508-L516
Or events:
https://github.com/Boscop/glutin/blob/master/src/events.rs#L61
https://github.com/Boscop/glutin/blob/master/src/api/win32/callback.rs#L342-L346

I implemented this because it's necessary to create a child window for writing GUIs for VST plugins. The plugin host creates a window and [passes it](https://github.com/overdrivenpotato/rust-vst2/blob/master/src/editor.rs#L23) to the plugin, which has to create a child window on that for it's GUI and register a timer on the HWND so that it refreshes it's GUI every few milliseconds.
This is what I do in the plugin:
```rust
	fn open(&mut self, parent: *mut c_void) {
		let mut display = WindowBuilder::new().with_dimensions(WINDOW_WIDTH, WINDOW_HEIGHT).with_parent(WindowID::new(parent)).build_glium().unwrap();
		{
			let window = display.get_window().unwrap();
			unsafe { user32::SetWindowLongA(window.platform_window() as winapi::HWND, winapi::winuser::GWLP_USERDATA, self as *mut _ as winapi::LONG); }
			window.set_timer(WIN_GUI_FRAME_TIMER, 25 /* ms */, Some(handle_timer)).unwrap();
		}
// ...
const WIN_GUI_FRAME_TIMER: winapi::UINT_PTR = 4242;
unsafe extern "system" fn handle_timer(hwnd: winapi::HWND, _: winapi::UINT, timer_id: winapi::UINT_PTR, elapsed_ms: winapi::DWORD) {
	if timer_id != WIN_GUI_FRAME_TIMER {
		return;
	}
[...]

Issue #93, opened 2016-11-24
allow creation of child window


k-brac:
It would be nice if we could create child window.
I can see 2 features : 
- use of the safe rust API (mostly already existing)
- unsafe API for FFI

-> I am experimenting in developing a plugin in rust and I would like to use for example conrod for GUI but the host application only gives me a void * as a parent window.

This change would allow me (and others) to embed their rust development into other applications.

If you think this is interesting I can look into this issue (for windows at least)

Since winit is already using CreateWindowEx I think it is just a matter of API design.
I don't know how this works for other platforms.

tomaka:
This needs to be done for Servo anyway.
Since the concept of a parent window doesn't exist everywhere, it must be done with the platform-specific builder.

francesca64:
Closing in light of #96 + #159

PR #1044, opened 2019-07-13
[Draft] Per-window event handling and enable creation of windows without an event loop for embedded windows

l0calh05t:
While creating child windows has become possible with platform-specific APIs (https://github.com/rust-windowing/winit/issues/93), use of winit for embedded windows such as plugin windows where external code provides a container window (see https://github.com/rust-windowing/winit/issues/159) remains blocked by the fact that such applications control the main event loop in external code, therefore `winit::EventLoop` cannot be used in such cases.

Therefore, I suggest adding `WindowBuilder.build_without_event_loop` (no equivalent `Window.new_without_event_loop` as this only really makes sense when also using `WindowBuilderExtWindows.with_parent_window` or other similar platform-specific functions). Furthermore, as lack of an event loop prevents calling the event handler passed to `EventLoop.run`, I suggest adding `Window.set_event_handler` to set an optional per-window event handler. This should make it possible to remain compatible with the existing API while gaining the ability to support embedded windows.
An added benefit is that it should make the handling of multi-window applications easier, as the user wouldn't have to perform his own event forwarding depending on `window_id`.

Currently, this PR is a suggestion for a possible API extension without an implementation. Suggestions for improvements to the API and ideas on how this could be implemented are very welcome.

### DIFF
--- a/src/window.rs
+++ b/src/window.rs
@@ -4,6 +4,7 @@ use std::fmt;
 use crate::{
     dpi::{LogicalPosition, LogicalSize},
     error::{ExternalError, NotSupportedError, OsError},
+    event::Event,
     event_loop::EventLoopWindowTarget,
     monitor::{AvailableMonitorsIter, MonitorHandle},
     platform_impl,
@@ -295,6 +296,20 @@ impl WindowBuilder {
         platform_impl::Window::new(&window_target.p, self.window, self.platform_specific)
             .map(|window| Window { window })
     }
+
+    /// Builds the window without an event loop.
+    ///
+    /// Intended for use with embedded windows where external code controls the main event loop.
+    /// Embedded windows should be created with `WindowBuilderExtWindows.with_parent_window` or
+    /// other similar platform-specific functions.
+    ///
+    /// Call `Window.set_event_handler` to set an event handler for the window.
+    ///
+    /// Possible causes of error include denied permission, incompatible system, and lack of memory.
+    #[inline]
+    pub fn build_without_event_loop(self) -> Result<Window, OsError> {
+        unimplemented!();
+    }
 }
 
 /// Base Window functions.
@@ -356,6 +371,23 @@ impl Window {
     #[inline]
     pub fn request_redraw(&self) {
         self.window.request_redraw()
+    }
+
+    /// Sets the event handler for a window and returns any previously set event handler.
+    ///
+    /// If an event handler is set, the handler is called for events with `window_id == window.id()`.
+    ///
+    /// This function is mainly intended for embedded windows with an external parent window such as
+    /// when using `WindowBuilderExtWindows.with_parent_window` where external code controls the main
+    /// event loop. However, it can also be used with `EventLoop.run` for per-window event handling.
+    /// In such cases, the event handler is called before the main event handler passed to
+    /// `EventLoop.run` is called.
+    #[inline]
+    pub fn set_event_handler<T: 'static>(
+        &self,
+        _event_handler: Option<Box<dyn FnMut(Event<T>) -> bool>>,
+    ) -> Option<Box<dyn FnMut(Event<T>) -> bool>> {
+        unimplemented!();
     }
 }
 


src/window.rs
```diff
@@ -295,6 +296,20 @@ impl WindowBuilder {
...
+    ///
+    /// Intended for use with embedded windows where external code controls the main event loop.
+    /// Embedded windows should be created with `WindowBuilderExtWindows.with_parent_window` or
+    /// other similar platform-specific functions.
+    ///
+    /// Call `Window.set_event_handler` to set an event handler for the window.
+    ///
+    /// Possible causes of error include denied permission, incompatible system, and lack of memory.
+    #[inline]
+    pub fn build_without_event_loop(self) -> Result<Window, OsError> {
```
ZeGentzy: After some pondering, I'm 80% certain this function can not be implemented with X11. Pinging @murarth to investigate this. Surely, no eventsloop = no `XConnection`, right?

ZeGentzy: And while we're at it, no `EventLoop` = no connection to the Wayland display server... I'd be surpized if this is implementable on any of the other platforms.

murarth: I haven't done any testing and I'm not totally clear on the details of the proposed API (I skimmed through the comments of the above mentioned issues, but I'm not familiar with the use of VST plugins or any other type of GUI plugin), but I think this could be possible.

While it's true that, presently, the `EventLoop` is responsible for opening the `XConnection`, that doesn't necessarily need to be the case.

ZeGentzy: Ok, but what do you think should be responsible for opening it. Keep in mind, we might bump into this, but for X11 if we do something too wierd: https://github.com/rust-windowing/glutin/issues/1152

murarth: It appears there's already a safeguard against this for X11. All `XConnection` instances are derived from the `lazy_static` here: https://github.com/rust-windowing/winit/blob/master/src/platform_impl/linux/mod.rs#L48

ZeGentzy: Right- but the "external code [that] controls the main event loop" might not be run by winit. So, that external code might be using a different X11 display.

murarth: True, but that is not an issue unique to winit. Any host program using an X (or Wayland) display server that expects a plugin to create a subwindow to its window must be expected to provide sufficient information to do so. If it doesn't, that would be a problem for any plugin.

The winit-specific element of this would be if, say, a host program provides an Xlib `Display*` that it expects the plugin to use, then we would need something like a `WindowBuilder::from_x_display` method to make use of it.

ZeGentzy: Yes. Which means this function is unfeasible for X11. I suspect for similar reasons it will be unfeasible on every single other platform. This needs to be a platform specific extension like glutin's `RawContextExt`.

murarth: Oh, right. Yes. I agree. (I forgot that we were talking about this specific platform-independent interface.)

src/window.rs
```diff
@@ -357,6 +372,23 @@ impl Window {
...
+    ///
+    /// This function is mainly intended for embedded windows with an external parent window such as
+    /// when using `WindowBuilderExtWindows.with_parent_window` where external code controls the main
+    /// event loop. However, it can also be used with `EventLoop.run` for per-window event handling.
+    /// In such cases, the event handler is called before the main event handler passed to
+    /// `EventLoop.run` is called.
+    #[inline]
+    pub fn set_event_handler<T: 'static>(
+        &self,
+        _event_handler: Option<Box<dyn FnMut(Event<T>) -> bool>>,
```
ZeGentzy: Presumably this callback should also be static? See https://github.com/rust-windowing/winit/blob/master/src/event_loop.rs#L138

Won't that interfere with `run`?

What is this function mean to be returning?

Osspial:
I've skimmed the API you've proposed, and I'll say that it would be *really* difficult to implement, on Windows at least. Most of the Windows backend has been designed with the assumption that `EventLoop` will be there, and getting rid of that assumption will probably involve several painful and difficult-to-maintain changes to the existing codebase.

Anyhow, I read through #159, and https://github.com/rust-windowing/winit/issues/159#issuecomment-511124405 gives the clearest summary of the technical requirements this proposal is trying to solve. With that context in mind, I've got a few questions about how this proposal goes about solving those problems, as well as some suggestions that may make it easier to integrate into our codebase:

a) The current proposal requires stripping out the `EventLoop` infrastructure completely, which involves scrapping much of Winit's internal structure and the various API conveniences granted by `EventLoop`. Why is that necessary? Would it be possible to have some sort of `EventLoop::run_externally_driven` function that doesn't actually drive the event loop, but hooks into an event loop that's already been set up?

b) If we do end up going forward with the API as proposed, how do the two new functions interact with the current `EventLoop` functionality?

crsaracco:
@l0calh05t are you on the `vst-rs` discord? Maybe we can take a look at trying this draft idea "internally" first (with actual VSTs as proof that our solution works), wrap our heads around what is exactly needed (and what would be the least painful changeset to `winit`), and then try an actual proposal afterwards?

src/window.rs
l0calh05t: > Yes. Which means this function is unfeasible for X11. I suspect for similar reasons it will be unfeasible on every single other platform. This needs to be a platform specific extension like glutin's `RawContextExt`.

This really shouldn't be a platform-specific extension, or rather, if it is there would have to be versions for all major platforms for it to be useful. Embedded/externally hosted windows are definitely possible on x11, windows, and mac. There are VST (and other) plugins for all three.

Looking at https://github.com/steinbergmedia/vstgui/blob/59e9dcfe61e4c4b15c5598ebe7d2e9a52fe6a33d/vstgui/lib/platform/linux/x11platform.cpp it seems that on X11 (but not on Win32 or Cocoa), an own `xcb_connect` call is issued and an event loop is started. I'm not sure how this is done without breaking the existing host loop (I'm no X11 expert). Maybe the `xcb_aux_sync` call is important here, or the host is expected to run separate threads for each plugin window. But they certainly don't on Mac or Win32.

> True, but that is not an issue unique to winit. Any host program using an X (or Wayland) display server that expects a plugin to create a subwindow to its window must be expected to provide sufficient information to do so. If it doesn't, that would be a problem for any plugin.

The APIs I know mostly just provide a single platform-specific pointer/handle/integer to the parent window. But somehow there are plugins on X11 anyways. So it is not unfeasible.

src/window.rs
l0calh05t: I'm not sure about the lifetime. I doubt it should be `'static`, as each Window could have its own handler and you don't know a priori how many windows you will have.

Why would it interfere with `run`?

My initial thought with the `bool` return value was that events that are handled by an `event_handler` would no longer be passed to the event handler passed to `run`, but it doesn't necessarily have to be implemented that way. In the embedded case, the `run` handler can't really exist anyways - at least not as is.

l0calh05t:
> I've looked over the API you've proposed, and I'll say that it would be _really_ difficult to implement, on Windows at least. Most of the Windows backend has been designed with the assumption that `EventLoop` will be there, and getting rid of that assumption will probably involve several painful and difficult-to-maintain changes to the existing codebase.

It is entirely possible that the current architecture may be a showstopper here. In principle it is fairly easy to implement on Windows. Usually, an `HWND` doesn't need to know anything about the event loop driving it. You have the per-`HWND` `GWLP_USERDATA` and can attach all information you need for a specific window instance. You would then have a `WndProc` for your window class that simply fetches the right callback from the user data.
 
> Anyhow, I read through #159, and [#159 (comment)](https://github.com/rust-windowing/winit/issues/159#issuecomment-511124405) gives the clearest summary of the technical requirements this proposal is trying to solve. With that context in mind, I've got a few questions about how this proposal goes about solving those problems, as well as some suggestions that may make it easier to integrate into our codebase:
> 
> a) The current proposal requires stripping out the `EventLoop` infrastructure completely, which involves scrapping much of Winit's internal structure and the various API conveniences granted by `EventLoop`. Why is that necessary? Would it be possible to have some sort of `EventLoop::run_externally_driven` function that doesn't actually drive the event loop, but hooks into an event loop that's already been set up?

The intent was to preserve `EventLoop` as much as possible by making its existence optional. However, the `Window` itself could no longer rely on the existence of such an object. How would `EventLoop::run_externally_driven` look like? It would have to return control to the calling thread immediately on most platforms, I believe.

> b) If we do end up going forward with the API as proposed, how do the two new functions interact with the current `EventLoop` functionality?

The intent was to make the existence of `EventLoop` optional, by providing

1. a way to create a `Window` without requiring an `EventLoop` instance
2. a way to handle events passed to a window, without a "global" event handler passed to `EventLoop::run`,

and to do this without requiring existing users of winit to change how they use the API.

l0calh05t:
> @l0calh05t are you on the `vst-rs` discord? Maybe we can take a look at trying this draft idea "internally" first (with actual VSTs as proof that our solution works), wrap our heads around what is exactly needed (and what would be the least painful changeset to `winit`), and then try an actual proposal afterwards?
> 
> I'd be willing to contribute to the Linux side of such an effort.

No, I don't use discord at all currently.

I would be willing to look into Windows, but from looking at winit, the current architecture makes things very difficult.

goddessfreya:
Closing for inactivity.

Issue #1029, opened 2019-07-08
Reduce code duplication between macOS and iOS


felixrabe:
There's duplicated code in:

- [src/platform_impl/macos](/rust-windowing/winit/tree/master/src/platform_impl/macos)
- [src/platform_impl/ios](/rust-windowing/winit/tree/master/src/platform_impl/ios)

Maybe introduce src/platform_impl/**common_apple** as a module that can hold common code for Apple platforms.

alasher:
Since "common_apple" isn't itself a device, I think that may not be the most organized place to put the common lib. What if we did something similar to the Xorg/Wayland setup? Something like this:

- platform_impl/
    - apple/
        - common/
        - ios/
        - macos/
        - mod.rs
    - ...

I might be interested in working on it, I'm interested in the project and have been trying to find ways to help here and there. So far documentation only, but this seems like a good way to help clean up and learn more about the project.

ZeGentzy:
I'd personally rather have `src/api/common_apple`, but afaik, this is a minor detail to be bikesheding on.

felixrabe:
I also think `platform_impl/apple/...` makes more sense than my original suggestion.

@alasher As a first step, you could maybe start just moving the existing modules to their new place, without introducing a `common` module yet, to test out how the new layout feels.

Then, like in a separate PR, tackle moving the common bits to a `common` module.

(I personally prefer having two separate PRs / master commits for this.)

PR #1059, opened 2019-07-22
Refactor iOS and macOS device folders

alasher:
    - Nothing added now, might highlight in release notes after more refactoring is done
    - Looked around at built docs and couldn't find any broken links - online docs are build for Unix target anyway.

Part 1 of an Apple refactor suggestion from issue #1029. All thoughts on this format are welcome. Unfortunately log info is lost, unless you use `git log --follow`. I didn't make an `apple` module or anything yet, I just wanted to make the minimum amount of changes in this commit to get every file in its new folder.

One issue I had was getting iOS tested, does anybody actually test Glutin/winit using this target? I can successfully build the winit iOS target, but I really wanted to get an example tested. I was able to get another [separate Rust project](https://github.com/kennytm/rust-ios-android) working on iOS using `cargo lipo`, but it looks like winit/glutin doesn't support this tool out of the box. It looks like we need a header to map the native Objective C/Swift code to the Rust winit static library definitions, but I thought I'd ask for advice before going that route. Any advice on this would be very useful, because I'd like to set up a local Glutin demo app on iOS to test any future refactoring changes I make.


felixrabe:
Don't worry about the `git log`, it looks like a clean commit to me, judging from https://github.com/rust-windowing/winit/pull/1059/files:

- It is clear what was moved.
- It is clear what was changed.

mtak-:
This repo might be helpful. https://github.com/mtak-/rust-on-mobile

The instructions should be compatible with winit as we use a very similar setup for our ios project.

aleksijuvani:
It's possible to test the iOS backend of winit with the Simulator on macOS. If the examples only output to the console, you can use `simctl spawn` to launch the executable and observe the output.

`rustup target add x86_64-apple-ios`
`cargo build --example video_modes --target x86_64-apple-ios`
`xcrun simctl spawn "iPhone 6s" target/x86_64-apple-ios/debug/examples/video_modes`

For graphical applications you will need to create an application bundle. This can be done by creating a new folder with the ".app" suffix, e.g. "Window.app". Inside this folder you will need to copy the executable (i.e. `cp target/x86_64-apple-ios/debug/examples/window Window.app`), and [create an "Info.plist" file](https://gist.github.com/aleksijuvani/14fb32b8a1dbbe99e314d891f8c5e38b).

You can then use `xcrun simctl install "iPhone 6s" Window.app` to install the application on to the simulator and use `xcrun simctl launch "iPhone 6s" com.foo.Window` to launch it, or launch it by hand from inside the simulator.

goddessfreya:
Closing for inactivity.

Issue #1041, opened 2019-07-10
Rethinking RedrawRequested


Osspial:
The current `RedrawRequested` event is currently hard to understand, which makes it difficult to implement and difficult to use. This issue will outline why I believe that is the case, and how I think we should fix it.

## The current problems

`RedrawRequested` is extremely inconsistent with the rest of our API. This is exhibited in a couple ways:

1. It's inconsistent with itself, since it *sometimes* provides ordering guarantees.
    - It's explicitly guaranteed to be the last event emitted if `request_redraw` is called, but isn't explicitly guaranteed to be the last event if the OS requested the redraw.
2. It's inconsistent with `WindowEvent`, since no other window event provides any sort of ordering guarantees.
3. It's inconsistent with `EventsCleared`, since it can be dispatched after `EventsCleared` if queued with `request_redraw` during the `EventsCleared` handler.

Some of that inconsistency is intentional (point 3), and some isn't (point 1), but the end result is that the API is difficult to explain and difficult to implement. Even beyond the complexity those inconsistencies introduce, the current model can result in duplication of `RedrawRequested` events when using [the recommended event loop structure described in the root docs](https://docs.rs/winit/0.20.0-alpha2/winit/#event-handling). Consider the following chain of events, with an application using the linked event loop structure:

1. The user resizes the window, resulting in the OS triggering a `RequestRedraw` event.
2. The event loop dispatches `RequestRedraw` between `NewEvents` and `EventsCleared`, as part of the standard control flow.
3. `EventsCleared` gets dispatched, upon which the user updates the application's state and calls `request_redraw`.
4. `RequestRedraw` gets dispatched immediately after `EventsCleared`.

The above situation has a couple problems: the first `RedrawRequested` is dispatched before the application has a chance to update its internal state, and the application goes through the expensive process of rendering a frame twice in a single iteration of the event loop. Ideally, our API wouldn't exhibit either issue, allowing the user to update their application state before all redraws occur and allowing Winit to unify redraw requests from the user and OS into a single event.

## Proposed changes

* Rename `EventsCleared` to `MainEventsCleared`
* Remove `RedrawRequested` from `WindowEvent`.
* Add `RedrawRequested(WindowId)` to `Event`.
    * Guarantee that `RedrawRequested` get dispatched between `MainEventsCleared` and `RedrawEventsCleared`.
* Add `RedrawEventsCleared` to `Event`.

[...]

PR #1062, opened 2019-07-23
Implement changes to `RedrawRequested` event

murarth:
Implements the changes described in #1041 for the X11 platform and for
platform-independent public-facing code.

### DIFF
--- a/src/platform_impl/linux/wayland/event_loop.rs
+++ b/src/platform_impl/linux/wayland/event_loop.rs
@@ -294,7 +294,7 @@ impl<T: 'static> EventLoop<T> {
             // send Events cleared
             {
                 sticky_exit_callback(
-                    crate::event::Event::EventsCleared,
+                    crate::event::Event::MainEventsCleared,
                     &self.window_target,
                     &mut control_flow,
                     &mut callback,
@@ -438,7 +438,8 @@ impl<T> EventLoop<T> {
                     );
                 }
                 if refresh {
-                    sink.send_window_event(crate::event::WindowEvent::RedrawRequested, wid);
+                    unimplemented!()
+                    //sink.send_window_event(crate::event::WindowEvent::RedrawRequested, wid);
                 }
                 if closed {
                     sink.send_window_event(crate::event::WindowEvent::CloseRequested, wid);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 - On Windows, location of `WindowEvent::Touch` are window client coordinates instead of screen coordinates.
 - On X11, fix delayed events after window redraw.
+- Changes to the `RedrawRequested` event (#1041):
+  - `RedrawRequested` has been moved from `WindowEvent` to `Event`.
+  - `EventsCleared` has been renamed to `MainEventsCleared`.
+  - `RedrawRequested` is now issued only after `MainEventsCleared`.
+  - `RedrawEventsCleared` is issued after each set of `RedrawRequested` events.
 
 # 0.20.0 Alpha 2 (2019-07-09)
 
--- a/examples/request_redraw.rs
+++ b/examples/request_redraw.rs
@@ -19,14 +19,11 @@ fn main() {
             event: WindowEvent::CloseRequested,
             ..
         } => *control_flow = ControlFlow::Exit,
-        Event::EventsCleared => {
+        Event::MainEventsCleared => {
             window.request_redraw();
             *control_flow = ControlFlow::WaitUntil(Instant::now() + Duration::new(1, 0))
         }
-        Event::WindowEvent {
-            event: WindowEvent::RedrawRequested,
-            ..
-        } => {
+        Event::RedrawRequested(_) => {
             println!("{:?}", event);
         }
         _ => (),
--- a/src/event.rs
+++ b/src/event.rs
@@ -29,9 +29,22 @@ pub enum Event<T> {
     UserEvent(T),
     /// Emitted when new events arrive from the OS to be processed.
     NewEvents(StartCause),
-    /// Emitted when all of the event loop's events have been processed and control flow is about
-    /// to be taken away from the program.
-    EventsCleared,
+    /// Emitted when all events (except for `RedrawRequested`) have been reported.
+    ///
+    /// This event is followed by zero or more instances of `RedrawRequested`
+    /// and, finally, `RedrawEventsCleared`.
+    MainEventsCleared,
+
+    /// The OS or application has requested that a window be redrawn.
+    ///
+    /// Emitted only after `MainEventsCleared`.
+    RedrawRequested(WindowId),
+
+    /// Emitted after any `RedrawRequested` events.
+    ///
+    /// If there are no `RedrawRequested` events, it is reported immediately after
+    /// `MainEventsCleared`.
+    RedrawEventsCleared,
 
     /// Emitted when the event loop is being shut down. This is irreversable - if this event is
     /// emitted, it is guaranteed to be the last event emitted.
@@ -52,7 +65,9 @@ impl<T> Event<T> {
             WindowEvent { window_id, event } => Ok(WindowEvent { window_id, event }),
             DeviceEvent { device_id, event } => Ok(DeviceEvent { device_id, event }),
             NewEvents(cause) => Ok(NewEvents(cause)),
-            EventsCleared => Ok(EventsCleared),
+            MainEventsCleared => Ok(MainEventsCleared),
+            RedrawRequested(wid) => Ok(RedrawRequested(wid)),
+            RedrawEventsCleared => Ok(RedrawEventsCleared),
             LoopDestroyed => Ok(LoopDestroyed),
             Suspended => Ok(Suspended),
             Resumed => Ok(Resumed),
@@ -183,9 +198,6 @@ pub enum WindowEvent {
         axis: AxisId,
         value: f64,
     },
-
-    /// The OS or application has requested that the window be redrawn.
-    RedrawRequested,
 
     /// Touch event has been received
     Touch(Touch),
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -48,19 +48,16 @@
 //!
 //! event_loop.run(move |event, _, control_flow| {
 //!     match event {
-//!         Event::EventsCleared => {
+//!         Event::MainEventsCleared => {
 //!             // Application update code.
 //!     
 //!             // Queue a RedrawRequested event.
 //!             window.request_redraw();
 //!         },
-//!         Event::WindowEvent {
-//!             event: WindowEvent::RedrawRequested,
-//!             ..
-//!         } => {
+//!         Event::RedrawRequested(_) => {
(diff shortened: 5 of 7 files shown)


murarth:
I've implemented the proposed changes for X11 only, inserting an `unimplemented!()` in the Wayland code so that this code will compile. Thus, this changeset will only compile on Linux and will only run as expected on X11.

I'm not sure how incomplete implementations are usually done, so please let me know if I should do something differently with this PR.

src/platform_impl/linux/wayland/event_loop.rs
```diff
@@ -438,7 +438,8 @@ impl<T> EventLoop<T> {
                     );
                 }
                 if refresh {
-                    sink.send_window_event(crate::event::WindowEvent::RedrawRequested, wid);
+                    unimplemented!()
+                    //sink.send_window_event(crate::event::WindowEvent::RedrawRequested, wid);
```
ZeGentzy: Yike.

ZeGentzy:
Yeah, it would be a very long while...

@FliegendeWurst, since you are on the testers list, would you like to test this? Please make sure the examples don't freeze, and the ordering of the events is consistent with what is proposed here: https://github.com/rust-windowing/winit/issues/1041

Thanks :)

FliegendeWurst:
The 'request_redraw' example no longer compiles since the event naming changed. Other examples ('cursor', 'cursor_grab', 'transparent', 'timer', 'window_icon', 'window_run_return') still seem to work. I adjusted the 'resizable' example to print debug information on all events. You can download the full log below. It seems like the event ordering works as intended, but some events are sometimes in two separate 'event cycles' (see comments below) and in the same cycle at other times. I'm not sure that's intended.

```
NewEvents(WaitCancelled { start: Instant { tv_sec: 3708, tv_nsec: 833814233 }, requested_resume: None })  
DeviceEvent { device_id: DeviceId(X(DeviceId(10))), event: Motion { axis: 0, value: -23.0 } }  
DeviceEvent { device_id: DeviceId(X(DeviceId(10))), event: Motion { axis: 1, value: -5.0 } }  
DeviceEvent { device_id: DeviceId(X(DeviceId(10))), event: MouseMotion { delta: (-23.0, -5.0) } }  
WindowEvent { window_id: WindowId(X(WindowId(153092097))), event: CursorEntered { device_id: DeviceId(X(DeviceId(2))) } }  
WindowEvent { window_id: WindowId(X(WindowId(153092097))), event: CursorMoved { device_id: DeviceId(X(DeviceId(2))), position: LogicalPosition { x: 527.0769230769231, y: 138.46153846153848 }, modifiers: ModifiersState { shift: false, ctrl: false, alt: false, logo: false } } }  
WindowEvent { window_id: WindowId(X(WindowId(153092097))), event: CursorLeft { device_id: DeviceId(X(DeviceId(2))) } }  
MainEventsCleared  
RedrawEventsCleared  
NewEvents(WaitCancelled { start: Instant { tv_sec: 3708, tv_nsec: 835169670 }, requested_resume: None })  
WindowEvent { window_id: WindowId(X(WindowId(153092097))), event: Resized(LogicalSize { width: 525.2307692307693, height: 200.30769230769232 }) }  
MainEventsCleared  
RedrawRequested(WindowId(X(WindowId(153092097)))) // Redraw immediately after Resized event  
RedrawEventsCleared  
NewEvents(WaitCancelled { start: Instant { tv_sec: 3708, tv_nsec: 844635764 }, requested_resume: None })  
DeviceEvent { device_id: DeviceId(X(DeviceId(10))), event: Motion { axis: 0, value: -67.0 } }  
DeviceEvent { device_id: DeviceId(X(DeviceId(10))), event: Motion { axis: 1, value: 0.0 } }  
DeviceEvent { device_id: DeviceId(X(DeviceId(10))), event: MouseMotion { delta: (-67.0, 0.0) } }  
MainEventsCleared  
RedrawEventsCleared  
NewEvents(WaitCancelled { start: Instant { tv_sec: 3708, tv_nsec: 856688181 }, requested_resume: None })  
DeviceEvent { device_id: DeviceId(X(DeviceId(10))), event: Motion { axis: 0, value: -83.0 } }  
DeviceEvent { device_id: DeviceId(X(DeviceId(10))), event: Motion { axis: 1, value: 3.5714285713620484 } }  
DeviceEvent { device_id: DeviceId(X(DeviceId(10))), event: MouseMotion { delta: (-83.0, 3.5714285713620484) } }  
MainEventsCleared  
RedrawEventsCleared  
NewEvents(WaitCancelled { start: Instant { tv_sec: 3708, tv_nsec: 867070153 }, requested_resume: None })  
MainEventsCleared  
RedrawEventsCleared  
NewEvents(WaitCancelled { start: Instant { tv_sec: 3708, tv_nsec: 868223645 }, requested_resume: None })  
DeviceEvent { device_id: DeviceId(X(DeviceId(10))), event: Motion { axis: 0, value: -92.0 } }  
DeviceEvent { device_id: DeviceId(X(DeviceId(10))), event: Motion { axis: 1, value: 3.5714285713620484 } }  
DeviceEvent { device_id: DeviceId(X(DeviceId(10))), event: MouseMotion { delta: (-92.0, 3.5714285713620484) } }  
MainEventsCleared  
RedrawEventsCleared  
NewEvents(WaitCancelled { start: Instant { tv_sec: 3708, tv_nsec: 868638730 }, requested_resume: None })  
WindowEvent { window_id: WindowId(X(WindowId(153092097))), event: Resized(LogicalSize { width: 493.84615384615387, height: 200.30769230769232 }) }  
MainEventsCleared  
// no redraw event after resized event?  
RedrawEventsCleared  
NewEvents(WaitCancelled { start: Instant { tv_sec: 3708, tv_nsec: 868749321 }, requested_resume: None })  
MainEventsCleared  
RedrawRequested(WindowId(X(WindowId(153092097)))) // here it is, apparently  
RedrawEventsCleared
```

[log.txt](https://github.com/rust-windowing/winit/files/3435203/log.txt)

ZeGentzy:
Yeah, it should be on the same cycle, otherwise applications that use both `Poll` and issue `request_redraw` on every `MainEventsCleared` will do one unnecessary redraw. I can't think of an obvious solution to this beyond just waiting for a `RedrawRequested` event to land, which seams unreasonable.

Maybe we can make it so that `MainEventsCleared` only gets sent if there are new main events?  Same with `RedrawEventsCleared`? And then we can add an `AllEventsCleared`, which is sent after all the events are cleared (both redraw and main), where users can preferably put stuff like setting control_flow, or checking some logic they want done every time the events loop is about to end a cycle.
How do you feel about this solution, @Osspial?

murarth:
I've fixed the example. I'm not sure what can be done about the issue of related events being split across separate event loop iterations.

Osspial:
IIRC X11 is an asynchronous API. Perhaps we're seeing the redraw event splitting because of synchronization issues between Winit and the X11 server, roughly as so?:

1. X11 posts a `Resize` event
2. Winit drains the event buffer, pulling the `Resize` event
3. X11 posts a `Redraw` event while Winit is processing the event buffer
4. Winit finishes event processing, starts a new iteration of the event loop, checks the event buffer, and finally sees the `Redraw` event
5. Winit dispatches `Redraw` in a new iteration of the event loop

Also re:@ZeGentzy - I'm strongly opposed to only sending `MainEventsCleared`/`RedrawEventsCleared` when the relevant events exist. Those events are incredibly useful as marker events that help applications react to the event loop, and dispatching them inconsistently means downstream applications that use a `Poll` model (like games) can't put their update logic in `MainEventsCleared`. That'd be a major usability headache.

ZeGentzy:
@Osspial There is no guarantee that after an `ConfigureNotify` notify event (`Resized`) we will receive a `Expose` event (`RedrawRequested`).  For that reason, we simply can't wait till the `Expose` event arrives, as it may never.

Referring to the docs, every X function that X clients may call says that XYZ *may* generate Expose events. Two functions of interest are:

> [...] Changing the size of a mapped window *may* lose its contents and generate Expose events. [...]

-- XResizeWindow, emphasis mine.


> [...] Moving and resizing a mapped window *may* generate an Expose event on the window. [...]

-- XMoveResizeWindow, emphasis mine.

There is no reason to suggest that `Expose` events caused by the X server are anymore guaranteed. In fact:

> The X server *can* report Expose events to clients wanting information about when the contents of window regions have been lost. *The circumstances in which the X server generates Expose events are not as definite as those for other events.* [then it goes on to list some reasons that the event *may* be generated.]

-- §10.9.1 Expose Events,  emphasis mine.

I did some googling, and I found this document:
> [...]  A ConfigureNotify event is always followed by one or more Expose events. 

-- http://csweb.cs.wfu.edu/~torgerse/Kokua/Irix_6.5.21_doc_cd/usr/share/Insight/library/SGI_bookshelves/SGI_Developer/books/XLib_PG/sgi_html/ch03.html

However, that claim seems to contradict this mozilla bug here which mentions the expose event arriving before the configure one: https://bugzilla.mozilla.org/show_bug.cgi?id=1280653#c11

Fwiw, I tested this out with this small patch against master (not this PR) and found that an `Expose` event always happened after the `ConfigureNotify` w/ i3wm, however, there is no reason to believe that this will hold true for all hardware/software configurations. Waiting for an `Expose` event seams destined to cause issues on some obscure setups.

```
diff --git a/examples/window.rs b/examples/window.rs
index 12e0593e..e0949ba0 100644
--- a/examples/window.rs
+++ b/examples/window.rs
@@ -13,13 +13,21 @@ fn main() {
         .unwrap();
 
         match event {
             Event::WindowEvent {
                 window_id,
+                event: WindowEvent::Resized(_),
+                window_id,
+            } => println!("{:?}", event),
+                event: WindowEvent::RedrawRequested,
+                window_id,
+            } => println!("{:?}", event),
+            Event::EventsCleared => println!("{:?}", event),
         }
     });
```

In fact, I could reproduce this redraw splitting on the master, so this isn't even a regression from what's currently happening on master:
```
EventsCleared
WindowEvent { window_id: WindowId(X(WindowId(69206017))), event: Resized(LogicalSize { width: 377.53846153846155, height: 375.69230769230774 }) }
WindowEvent { window_id: WindowId(X(WindowId(69206017))), event: RedrawRequested }
EventsCleared
EventsCleared
WindowEvent { window_id: WindowId(X(WindowId(69206017))), event: Resized(LogicalSize { width: 357.2307692307693, height: 359.0769230769231 }) }
WindowEvent { window_id: WindowId(X(WindowId(69206017))), event: RedrawRequested }
EventsCleared
EventsCleared
EventsCleared
WindowEvent { window_id: WindowId(X(WindowId(69206017))), event: Resized(LogicalSize { width: 343.3846153846154, height: 348.0 }) }
EventsCleared
WindowEvent { window_id: WindowId(X(WindowId(69206017))), event: RedrawRequested }
EventsCleared
```

Given that, I'm approving this PR.

murarth:
Fixed the doctests. The remaining CI failures are on Windows and Mac.

ZeGentzy:
One more change, we should only issue `RequestRedraw` if `Expose`'s count == 0.

PR #1066, opened 2019-07-23
Update the percent-encoding crate to 2.0

SimonSapin:


SimonSapin:
I see that the version number is 0.20.0-alpha2. Does that mean that 0.20 is not ready yet? How do you feel about backporting this PR and publishing 0.19.2? (I’m hoping to de-duplicate dependencies in Servo, which is using 0.19.1 at the moment.)

PR #1076, opened 2019-07-29
Update the percent-encoding crate to 2.0 in winit v0.19.2

SimonSapin:
This is the same as https://github.com/rust-windowing/winit/pull/1066, but targeting the `winit-legacy` branch.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "winit"
-version = "0.19.1"
+version = "0.19.2"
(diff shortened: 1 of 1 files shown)


SimonSapin:
Since v0.20.0 is not ready yet, would you agree to publish this change (and others in the branch since v0.19.1) on crates.io as v0.19.2?

Osspial:
@SimonSapin That seems reasonable.

SimonSapin:
Thank you!

PR #1083, opened 2019-07-31
Fix armv7-apple-ios compile target

aleksijuvani:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,11 +34,27 @@ matrix:
       os: osx
       rust: stable
 
-    # iOS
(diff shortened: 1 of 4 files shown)


ZeGentzy:
Shouldn't we also be testing `x86_64-apple-ios`?

aleksijuvani:
Sure! I believe it's only used for the Simulator. I dropped it initially because I was worried it would inflate the build times for Travis, but it looks like it's so fast that it doesn't even matter. I'll add it back.

aleksijuvani:
Should be all good now!

PR #1084, opened 2019-08-01
macOS/iOS: Fix auto trait impls of `EventLoopProxy`

yvt:
This pull request fixes `EventLoopProxy<T>`'s auto trait implementations.

- `EventLoopProxy<T>: Send` *only if* `T: Send`, because otherwise it would allow sending unsendable objects.
- `EventLoopProxy<T>: !Sync` because of `std::sync::mpsc::Sender` used under the hood. This restriction is congruent with other platforms' `EventLoopProxy`.

----------------------

    - [x] macOS
    - [ ] iOS — No experience with running Rust code on iOS whatsoever, can someone please test this?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - On iOS, add support for hiding the home indicator.
 - On iOS, add support for deferring system gestures.
 - On iOS, fix a crash that occurred while acquiring a monitor's name.
+- On macOS and iOS, corrected the auto trait impls of `EventLoopProxy`.
 
 # 0.20.0 Alpha 2 (2019-07-09)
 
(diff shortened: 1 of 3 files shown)


mtak-:
Nice catch! This breaks our build (for good reasons), but unfortunately there's no great workaround because the `Clone` impl on `EventLoopProxy` requires `T: Clone`. Would be great to get that fixed as well - even if it's only for the macos/ios backends.

ZeGentzy:
Needs a changelog entry, then lgtm.

yvt:
@mtak- Since `EventLoopProxy` wraps each platform's implementation, other platforms had to be updated as well (otherwise they'll break), so I opened a separate PR: #1086

yvt:
Added a changelog entry.

mtak-:
Thanks!

PR #1088, opened 2019-08-02
Disable overscan compensation for external displays on iOS

aleksijuvani:

r? @mtak-

### DIFF
--- a/src/platform_impl/ios/window.rs
+++ b/src/platform_impl/ios/window.rs
@@ -16,7 +16,7 @@ use crate::{
         event_loop,
         ffi::{
             id, CGFloat, CGPoint, CGRect, CGSize, UIEdgeInsets, UIInterfaceOrientationMask,
-            UIRectEdge,
+            UIRectEdge, UIScreenOverscanCompensation,
         },
         monitor, view, EventLoopWindowTarget, MonitorHandle,
     },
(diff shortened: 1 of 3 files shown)


src/platform_impl/ios/window.rs
```diff
@@ -165,7 +165,7 @@ impl Inner {
             let uiscreen = match monitor {
                 Some(Fullscreen::Exclusive(video_mode)) => {
                     let uiscreen = video_mode.video_mode.monitor.ui_screen() as id;
-                    let () = msg_send![uiscreen, setCurrentMode: video_mode.video_mode.screen_mode];
+                    msg_send![uiscreen, setCurrentMode: video_mode.video_mode.screen_mode];
```
mtak-: `winit` has been using explicit return types to guard against `msg_send!` calls deducing to the `!` type after never is stabilized (presently they deduce to `()` when the type is not explicitly stated). Previous attempts to stabilize `!` have caused many bugs in winit. See https://github.com/rust-windowing/winit/pull/428 and the other github activity referencing it. If/when never is stabilized, these calls are likely to deduce to `!` causing UB.

I'll open an issue on `gfx` as well, as I know they also have unconstrained types stemming from `msg_send`.

mtak-: https://github.com/gfx-rs/gfx/issues/2939

aleksijuvani: Thank you! I've addressed the issue here.

mtak-:
Tested on AppleTV. Works. Thanks!

aleksijuvani:
Thank you for testing this! I've rebased this against `master`.

Issue #1072, opened 2019-07-26
Web support tracking issue


ryanisaacg:
- [x] On-close event listening (#1155)
- [x] Support fullscreen APIs (#1142)
- [x] ~~Support cursor grab~~
- [x] ~~Support maximizing a canvas~~
- [x] ~~Finalize decision on how to handle VideoMode~~
- [x] Decide if tab title / favicon should be exposed by some cross-platform extension, treated as window icon and title, or neither (They should be left to the user to handle)
- [x] Use raw-window-handle (#1192)
    - [x] Add web support to [raw-window-handle](https://github.com/rust-windowing/raw-window-handle/pull/26)
- [x] Update documentation to reflect web support
- [x] Fix the RedrawRequested implementation on web not conforming to the expected behavior ( #1181)

hecrj:
Before proceeding, I think we need to get on the same page on how to tackle these last features.

> On-close event listening

How would this work? We need to keep in mind that right now canvases have a "local" scope. We could attach a `beforeunload` event listener to the `window` and proxy the event to all the existing canvases. That would be global thing, so it may be awkward to implement/maintain, and I feel it could lead to weird bugs overall.

Ideally, we should be able to detect if a particular canvas has been "closed". But when do we consider a canvas to be closed? When it's removed from the DOM? That wouldn't work because a user may want to reinsert the canvas and keep using it.

I am not sure there is a reasonable implementation for this. I am inclined to simply state that a canvas cannot be closed, and ask our users to take care of any unload wiring themselves (the same way we ask them to insert the canvas freely).

> Support fullscreen APIs
> Support maximizing a canvas

These ones are also tricky. Anything that touches global state (i.e. `document`, `window`, etc.) will have erratic/unintuitive behavior when dealing with multiple canvases (unless we emulate a window manager ourselves). What should happen if our users try to maximize/toggle fullscreen on multiple canvases on the same page?

I think I would simply let the users handle this too. At least for now. Users will be able to make better informed decisions as they will know exactly how many canvases they will be dealing with.

> Decide if tab title / favicon should be exposed by some cross-platform extension, treated as window icon and title, or neither

I think this should be left to the users of the API, as we don't currently have a concept of "main" canvas.

ryanisaacg:
@hecrj

> > On-close event listening

> How would this work? 

[...]

PR #1089, opened 2019-08-02
Stop appending canvas to document in web platform

hecrj:
The changes here stop `winit` from automatically appending the window canvas to the document root in the web platform. I discussed this with @ryanisaacg on Discord and we seemed to agree.

I think this approach has multiple benefits:

  * It reduces the scope of `winit` to simply providing a canvas with the proper event listeners attached.
  * We get multiple canvas support for free, given that all the event listeners are currently local to the canvas (however, we still need to think if we should provide fullscreen support and how in #1072).
  * The user decides how to embed the canvas freely in their own web document. This allows us to leave some functionality out of `winit`'s scope, like changing the document/tab title or updating the favicon. The user is expected to implement this on top of `winit` if necessary.

Additionally, I have updated an unnecessary `TODO` comment and fixed some panics on canvas creation.

### DIFF
--- a/src/platform_impl/web/stdweb/canvas.rs
+++ b/src/platform_impl/web/stdweb/canvas.rs
@@ -56,7 +56,7 @@ impl Canvas {
         // https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex
         canvas
             .set_attribute("tabindex", "0")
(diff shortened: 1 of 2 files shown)


ZeGentzy:
lgtm, merging, thanks.

PR #1090, opened 2019-08-03
Add touch pressure information for touch events on iOS

aleksijuvani:

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -324,16 +324,71 @@ pub struct Touch {
     pub device_id: DeviceId,
     pub phase: TouchPhase,
     pub location: LogicalPosition,
-    /// Force of the touch, between 0.0 and 1.0, where 1.0 is the maximum
-    /// possible force. Returns `None` if the platform does not provide this
-    /// data.
+    /// Describes how hard the screen was pressed. May be `None` if the platform
+    /// does not support pressure sensitivity.
     ///
     /// ## Platform-specific
     ///
     /// - Only available on **iOS**.
-    pub force: Option<f64>,
+    pub force: Option<Force>,
     /// Unique identifier of a finger.
     pub id: u64,
+}
+
+/// Describes the force of a touch event
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum Force {
+    /// On iOS, the force is calibrated so that the same number corresponds to
+    /// roughly the same amount of pressure on the screen regardless of the
+    /// device.
+    Calibrated {
+        /// The force of the touch, where a value of 1.0 represents the force of
+        /// an average touch (predetermined by the system, not user-specific).
+        ///
+        /// The force reported by Apple Pencil is measured along the axis of the
+        /// pencil. If you want a force perpendicular to the device, you need to
+        /// calculate this value using the `altitude_angle` value.
+        force: f64,
+        /// The maximum possible force for a touch.
+        ///
+        /// The value of this field is sufficiently high to provide a wide
+        /// dynamic range for values of the `force` field.
+        max_possible_force: f64,
+        /// The altitude (in radians) of the stylus.
+        ///
+        /// A value of 0 radians indicates that the stylus is parallel to the
+        /// surface. The value of this property is Pi/2 when the stylus is
+        /// perpendicular to the surface.
+        altitude_angle: Option<f64>,
+    },
+    /// If the platform reports the force as normalized, we have no way of
+    /// knowing how much pressure 1.0 corresponds to – we know it's the maximum
+    /// amount of force, but as to how much force, you might either have to
+    /// press really really hard, or not hard at all, depending on the device.
+    Normalized(f64),
+}
+
+impl Force {
+    /// Returns the force normalized to the range between 0.0 and 1.0 inclusive.
+    /// Instead of normalizing the force, you should prefer to handle
+    /// `Force::Calibrated` so that the amount of force the user has to apply is
+    /// consistent across devices.
+    pub fn normalized(&self) -> f64 {
+        match self {
+            Force::Calibrated {
+                force,
+                max_possible_force,
+                altitude_angle,
+            } => {
+                let force = match altitude_angle {
+                    Some(altitude_angle) => force / altitude_angle.sin(),
+                    None => *force,
+                };
+                force / max_possible_force
+            }
+            Force::Normalized(force) => *force,
+        }
+    }
 }
 
 /// Hardware-dependent keyboard scan code.
--- a/src/event.rs
+++ b/src/event.rs
@@ -303,27 +303,36 @@ pub enum TouchPhase {
     Cancelled,
 }
 
-/// Represents touch event
+/// Represents a touch event
 ///
-/// Every time user touches screen new Start event with some finger id is generated.
-/// When the finger is removed from the screen End event with same id is generated.
+/// Every time the user touches the screen, a new `Start` event with an unique
+/// identifier for the finger is generated. When the finger is lifted, an `End`
+/// event is generated with the same finger id.
 ///
-/// For every id there will be at least 2 events with phases Start and End (or Cancelled).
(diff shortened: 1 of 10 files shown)


src/event.rs
```diff
@@ -303,27 +303,37 @@ pub enum TouchPhase {
...
-    /// unique identifier of a finger.
+    /// Force of the touch, between 0.0 and 1.0, where 1.0 is the maximum
+    /// possible force
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **Windows:** Always returns 1.0.
+    /// - **Wayland:** Always returns 1.0.
+    /// - **X11:** Always returns 1.0.
+    pub force: f64,
```
Osspial: We should expose this as an `Option` to make it more obvious when the platform isn't providing pressure data.

src/event.rs
```diff
@@ -303,27 +303,37 @@ pub enum TouchPhase {
...
     pub location: LogicalPosition,
-    /// unique identifier of a finger.
+    /// Force of the touch, between 0.0 and 1.0, where 1.0 is the maximum
+    /// possible force
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **Windows:** Always returns 1.0.
+    /// - **Wayland:** Always returns 1.0.
+    /// - **X11:** Always returns 1.0.
```
Osspial: Would it make more sense to document this as "only on iOS"?

Also, this documentation implies that this is implemented on Android, when it isn't.

src/platform_impl/ios/view.rs
```diff
@@ -218,6 +218,8 @@ unsafe fn get_window_class() -> &'static Class {
                         break;
                     }
                     let location: CGPoint = msg_send![touch, locationInView: nil];
+                    let force: CGFloat = msg_send![touch, force];
+                    let max_force: CGFloat = msg_send![touch, maximumPossibleForce];
```
mtak-: This needs a check to be sure force touch is supported on the device/in the os settings. Otherwise `force` is `NaN`.
```rust
let trait_collection: id = msg_send![object, traitCollection];
let touch_capability: UIForceTouchCapability = msg_send![trait_collection, forceTouchCapability];
let force = if touch_capability == UIForceTouchCapability::Available {
    let force: CGFloat = msg_send![touch, force];
    let max_force: CGFloat = msg_send![touch, maximumPossibleForce];
    Some((force / max_force) as _)
} else {
    None
};
```

src/platform_impl/ios/view.rs
```diff
@@ -235,6 +237,7 @@ unsafe fn get_window_class() -> &'static Class {
                             device_id: RootDeviceId(DeviceId { uiscreen }),
                             id: touch_id,
                             location: (location.x as f64, location.y as f64).into(),
+                            force: (force / max_force) as _,
```
mtak-: This is probably lossy and inconsistent across devices. From the [Apple docs](https://developer.apple.com/documentation/uikit/uitouch/1618110-force?language=objc).

> The force of the touch, where a value of 1.0 represents the force of an average touch (predetermined by the system, not user-specific).

If apple provides an API for `maximumPossibleForce` then it's likely to be different across devices, meanwhile `1.0` is always the force of an average touch. This means users will have to reverse the above logic to figure out if the touch is a force touch or not.

A couple of other questions:
- Do other platforms do things similarly to iOS (1.0 is average, everything is a float)?
- What scale is iOS using (linear/logarithmic/etc)?

aleksijuvani: I have similar concerns here. I didn't look at other platforms at length, but on Windows it appears that the force is reported as an integer between 0 and 1024 where 1024 is the maximum pressure applicable. I don't know what scale iOS is using for the force here. I'm not sure if it's possible to make this behave identically across platforms. What do you suggest?

src/platform_impl/ios/view.rs
aleksijuvani: Good catch! I've added the capability check.

src/event.rs
aleksijuvani: Done.

src/event.rs
aleksijuvani: Sure.

src/platform_impl/ios/view.rs
mtak-: On iOS it [https://forums.developer.apple.com/thread/31775](looks like) values are calibrated to be consistent across devices, however different devices may have different max force.

Android appears to have the 0-1 clamping behavior, but touch pressure scaling can cause > 1 values to be returned.

> I'm not sure if it's possible to make this behave identically across platforms.

Agreed. That's too bad :/.

> What do you suggest?

I suggest we just pass the float along with no scaling for now, and add some documentation explaining the values it might have on each platform.

Osspial: > I suggest we just pass the float along with no scaling for now, and add some documentation explaining the values it might have on each platform.

I strongly disagree with that suggestion. If we're exposing an API that's common across platforms, the semantics of that API should differ as little as possible between platforms - doing otherwise creates an API that's scary to use and will likely cause compatibility bugs downstream.

I'm not opposed to exposing a raw pressure value that is platform dependent (some sort of `TouchExt` trait would probably be best for this), but platform-dependent behaviors should be opt-in when possible, not the default.

aleksijuvani: We should then also add a function returning the maximum possible force on iOS.

aleksijuvani: If we don't normalize the value, the user will need to normalize the returned value in their application using platform-specific extension traits if a value in the 0..=1 range is required. This happens to be my use case.

Apple cautions against normalizing the force in the [forum thread](https://forums.developer.apple.com/thread/31775) linked above. They say this is because they configure this range by setting `maximumPossibleForce` so that, say, a value of 3.0 will roughly match the same force across all their devices. I don't think this is an issue for us, as we don't expect a value of 0.5 to match the same amount of force being applied to the device across different devices anyway, say, iPhones and Microsoft Surface tablets.

I think I'm leaning towards normalizing the force to the 0..=1 range, and providing platform-specific extensions for iOS to access the raw values for the purpose of detecting the force touch activation point.

mtak-: The `force: f64` value is difficult to use correctly without more information regardless of what we do. `winit` users should know something about the os, os settings, or input device.

IMO constraining the value from `0-1` would make folks assume that the value is useful without platform specific code, and gains little. iOS is the only platform that uses touch force as a core part of its UX design and making `force` easy to use _incorrectly_ on that platform by constraining the range feels wrong to me. Are you sure your usage of `force` won't cause someone with a different device to have to mash the screen really hard? Max possible force also can differ from touch to touch. [Apple pencil](https://developer.apple.com/documentation/uikit/pencil_interactions/handling_input_from_apple_pencil/computing_the_perpendicular_force_of_apple_pencil?language=objc) requires more care to get a correct force value.

Maybe force could be more than just a float, so that it's obvious that some nuanced interpretation of the value is required.

Names chosen arbitrarily:
```rust
enum Force {
  IOS { force: f64, max_possible: f64, altitude: Option<f64> },
  Normalized { force: f64 },
  ...
}
impl Force {
  fn normalized(&self) -> f64;
}
```

aleksijuvani: That makes a lot of sense to me. I've added a commit that adds a variant for the calibrated force returned on iOS. I think the ergonomics of this are pretty good, and I've added some documentation to make this easier to understand for users. What do you think, @Osspial?

Osspial: I'm down with that design.

aleksijuvani: I've rebased this against `master`. Should be good to merge, unless something else comes up?

Osspial:
Merging. We'll probably want to do another design pass on the `Force` enum once we have stuff working on other platforms, but it works well enough now.

Thanks for the PR!

PR #1092, opened 2019-08-04
Replace `set_decorations` with `set_prefers_status_bar_hidden` on iOS

aleksijuvani:
I don't think the semantics of `set_decorations` really match with the intent of wanting to hide the status bar. On desktop platforms for example, fullscreen windows are expected to automatically get their decorations disabled. On iOS however, all windows are fullscreen but the status bar is still enabled by default. It's also a common use case to have a regular windowed application (with decorations!) on desktop platforms, but to also hide the status bar on iOS. It looks confusing to have the `set_decorations` call be only invoked for iOS. This pull request makes `set_decorations` a no-op on iOS and adds a new extension function `set_prefers_status_bar_hidden` to `WindowExtIOS` and `with_prefers_status_bar_hidden` to `WindowBuilderExtIOS`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@
 - On iOS, fix a crash that occurred while acquiring a monitor's name.
 - On iOS, fix armv7-apple-ios compile target.
 - Removed the `T: Clone` requirement from the `Clone` impl of `EventLoopProxy<T>`.
+- On iOS, add `set_prefers_status_bar_hidden` extension function instead of
+  hijacking `set_decorations` for this purpose.
 
 # 0.20.0 Alpha 2 (2019-07-09)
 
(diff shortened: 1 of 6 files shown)


mtak-:
Loving all these iOS PRs!

Mind updating the [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md) to replace the ✔️'s in window decorations with `**N/A**`, and adding a note about hiding the status bar to the `### iOS` section?

The platform specific docs for [`set_decorations`](https://github.com/rust-windowing/winit/blob/8a1c5277eb96d25d2c5e821eb24aa4a4e3020b74/src/window.rs#L588) should also be removed.

Thanks!

aleksijuvani:
Thank you! I've now updated the documentation appropriately.

PR #1094, opened 2019-08-05
[WIP] iOS os version checking around certain APIs

mtak-:

I am waiting on #1090 to be finished and merged so I can add some version checking around the force touch logic.

This PR also documents in `FEATURES.md` that `winit` requires atleast iOS version 8.0.


mtak-:
This is ready for review!

src/platform_impl/ios/monitor.rs
```diff
@@ -35,15 +38,20 @@ impl Drop for VideoMode {
...
 
 impl VideoMode {
     unsafe fn retained_new(uiscreen: id, screen_mode: id) -> VideoMode {
         assert_main_thread!("`VideoMode` can only be created on the main thread on iOS");
-        let refresh_rate: NSInteger = msg_send![uiscreen, maximumFramesPerSecond];
+        let refresh_rate: NSInteger = if app_state::capabilities().maximum_frames_per_second {
+            msg_send![uiscreen, maximumFramesPerSecond]
+        } else {
+            log::warn!("`-[UIScreen maximumFramesPerSecond]` is unsupported, defaulting to 60 fps");
+            60
```
francesca64: How was this determined to be a reasonable default? (I know the answer, but I want a comment.)

It could also be good to have the warning contain info on both 1) the version the user has, and 2) the version the user would need for this feature.

src/platform_impl/ios/monitor.rs
```diff
@@ -35,15 +38,20 @@ impl Drop for VideoMode {
     fn drop(&mut self) {
         unsafe {
             assert_main_thread!("`VideoMode` can only be dropped on the main thread on iOS");
-            msg_send![self.screen_mode, release];
+            msg_send![self.screen_mode, release]
```
francesca64: Piggybacking off of `drop`'s return type for this is very cute, but I'm worried it's a bit too opaque - it's a case where re-ordering the lines would change the semantics in surprising ways.

src/platform_impl/ios/app_state.rs
```diff
@@ -618,3 +619,54 @@ impl EventLoopWaker {
...
+    pub safe_area: bool,
+    pub home_indicator_hidden: bool,
+    pub defer_system_gestures: bool,
+    pub maximum_frames_per_second: bool,
+    pub force_touch: bool,
+}
+
+impl NSOperatingSystemVersion {
+    fn meets_requirements(&self, required_major: NSInteger, required_minor: NSInteger) -> bool {
+        (self.major, self.minor) >= (required_major, required_minor)
```
francesca64: Omg, I didn't realize you could do this!

src/platform_impl/ios/app_state.rs
```diff
@@ -618,3 +619,54 @@ impl EventLoopWaker {
...
+}
+
+impl From<NSOperatingSystemVersion> for Capabilities {
+    fn from(os_version: NSOperatingSystemVersion) -> Capabilities {
+        assert!(
+            os_version.major >= 8,
+            "`winit` requires iOS version 8 or greater"
+        );
+
+        let safe_area = os_version.meets_requirements(11, 0);
```
francesca64: It would be good to have comments documenting the source for all of these requirements.

src/platform_impl/ios/app_state.rs
```diff
@@ -618,3 +619,54 @@ impl EventLoopWaker {
...
+        }
+    }
+}
+
+pub fn capabilities() -> Capabilities {
+    lazy_static! {
+        static ref CAPABILITIES: Capabilities = {
+            let version: NSOperatingSystemVersion = unsafe {
+                let process_info: id = msg_send![class!(NSProcessInfo), processInfo];
+                msg_send![process_info, operatingSystemVersion]
```
francesca64: On iOS <8, will this prevent us from reaching the assertion? I don't remember how unhappy `msg_send` gets about things like that.

src/platform_impl/ios/app_state.rs
```diff
@@ -618,3 +619,54 @@ impl EventLoopWaker {
...
+    fn meets_requirements(&self, required_major: NSInteger, required_minor: NSInteger) -> bool {
+        (self.major, self.minor) >= (required_major, required_minor)
+    }
+}
+
+impl From<NSOperatingSystemVersion> for Capabilities {
+    fn from(os_version: NSOperatingSystemVersion) -> Capabilities {
+        assert!(
+            os_version.major >= 8,
+            "`winit` requires iOS version 8 or greater"
```
francesca64: There should be comments documenting why that's the requirement (note that I don't think the assertion needs to detail that explanation).

src/platform_impl/ios/view.rs
```diff
@@ -27,24 +27,49 @@ macro_rules! add_property {
...
-                $after_set
-            }
+                $setter_name
+            } else {
+                #[allow(non_snake_case)]
+                extern "C" fn $setter_name($object: &mut Object, _: Sel, value: $t) {
+                    unsafe {
+                        $object.set_ivar::<$t>(VAR_NAME, value);
+                    }
+                    log::warn!(concat!("`", stringify!($getter_name), "` is has no effect on this version of iOS"))
```
francesca64:
```suggestion
                    log::warn!(concat!("`", stringify!($getter_name), "` has no effect on this version of iOS"))
```
It would also be good to add the same info that I mentioned in my comment on `maximumFramesPerSecond` - unless plumbing that info through would be super painful?

src/platform_impl/ios/view.rs
```diff
@@ -213,30 +240,36 @@ unsafe fn get_window_class() -> &'static Class {
...
-                        msg_send![trait_collection, forceTouchCapability];
-                    let force = if touch_capability == UIForceTouchCapability::Available {
-                        let force: CGFloat = msg_send![touch, force];
-                        let max_possible_force: CGFloat = msg_send![touch, maximumPossibleForce];
-                        let altitude_angle: Option<f64> = if touch_type == UITouchType::Pencil {
-                            let angle: CGFloat = msg_send![touch, altitudeAngle];
-                            Some(angle as _)
+                    let force = if supports_force {
+                        let trait_collection: id = msg_send![object, traitCollection];
+                        let touch_capability: UIForceTouchCapability =
```
francesca64: It might be good to mention that this is checking for _device-level_ support, and to make it more explicit that `Capabilities` is specifically OS-level support (which would probably also involve adding some comments on `Capabilities`).

src/platform_impl/ios/app_state.rs
mtak-: Ah. Yes. That's silly. The [alternative](https://developer.apple.com/documentation/foundation/nsprocessinfo/1408730-operatingsystemversionstring?language=objc) is a pretty terrible API, but I'll move to that.

src/platform_impl/ios/monitor.rs
```diff
@@ -35,15 +38,27 @@ impl Drop for VideoMode {
...
 impl VideoMode {
     unsafe fn retained_new(uiscreen: id, screen_mode: id) -> VideoMode {
         assert_main_thread!("`VideoMode` can only be created on the main thread on iOS");
-        let refresh_rate: NSInteger = msg_send![uiscreen, maximumFramesPerSecond];
+        let os_capabilities = app_state::os_capabilities();
+        let refresh_rate: NSInteger = if os_capabilities.maximum_frames_per_second {
+            msg_send![uiscreen, maximumFramesPerSecond]
+        } else {
+            // All iOS devices support 60 fps, and on devices where `maximumFramesPerSecond` is not
+            // supported, they are all guaranteed to have 60hz refresh rates. This does not
```
francesca64: What's the source for this guarantee?

PR #1100, opened 2019-08-11
create 'custom events' example

### LINKED ISSUE #953: Add example showing custom events
Coming from https://gist.github.com/Osspial/667344cda503df3490a58a8ccf91c700, I was wondering whether `CustomEvent` was actually required to derive `Copy`:

```rust
#[derive(Debug, Clone, Copy)]
enum CustomEvent {
    Timer,
}
```

I didn't find `examples/custom_events.rs`. Maybe the example from the above mentioned announcement gist could be extracted there and then linked?

(Note: I'm not looking for clarification on deriving `Copy` here, it was just what led me to look for an example.)

satrix321:
* fixes #953

  - [x] Windows
  - [ ] macOS
  - [ ] X11
  - [ ] Wayland
  - [ ] iOS

### DIFF
--- a/examples/custom_events.rs
+++ b/examples/custom_events.rs
@@ -1,0 +1,51 @@
+use std::time::{Duration, Instant};
+use winit::{
+    event::{Event, StartCause, WindowEvent},
+    event_loop::{ControlFlow, EventLoop},
+    window::WindowBuilder,
+};
+
+#[derive(Debug, Clone, Copy)]
+enum CustomEvent {
+    Timer,
+}
+
+fn main() {
+    let event_loop = EventLoop::<CustomEvent>::with_user_event();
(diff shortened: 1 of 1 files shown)


Osspial:
Hi, and thanks for the PR!

We've actually got an example that shows how to use custom events already, in [`proxy.rs](https://github.com/rust-windowing/winit/blob/31ada5a052c79e16c7ad53ef030962e5f63a45d5/examples/proxy.rs). However, that example is pretty terribly named so it's easy to miss.

I think it might make more sense to rename that example and modify it so it's using a custom type, instead of `i32`.

satrix321:
Hi.

Makes sense, in that case I'll close this PR and open a new one for modifying the proxy.rs example.

PR #1134, opened 2019-08-30
Add touch pressure information for touch events on Windows

dam4rus:
Related to PR #1090

### DIFF
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -1516,7 +1516,7 @@ unsafe extern "system" fn public_window_callback<T>(
                                 continue;
                             },
                             location,
-                            force: None, // TODO
+                            force: None, // WM_TOUCH doesn't support pressure information
                             id: input.dwID as u64,
                             device_id: DEVICE_ID,
                         }),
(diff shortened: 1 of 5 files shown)


src/platform_impl/windows/event_loop.rs
```diff
@@ -1587,6 +1604,34 @@ unsafe extern "system" fn public_window_callback<T>(
                         continue;
                     }
 
+                    let force = match pointer_info.pointerType {
+                        winuser::PT_TOUCH => {
+                            let mut touch_info: winuser::POINTER_TOUCH_INFO = mem::uninitialized();
```
Osspial: We've moved away from using `mem::uninitialized`, since it's unclear if it's possible to use safely. Could you replace this with `MaybeUninit`?

src/platform_impl/windows/event_loop.rs
```diff
@@ -1587,6 +1604,34 @@ unsafe extern "system" fn public_window_callback<T>(
...
+                                    pointer_info.pointerId,
+                                    &mut touch_info as *mut _,
+                                ) {
+                                    0 => None,
+                                    _ => normalize_pointer_pressure(touch_info.pressure),
+                                }
+                            })
+                        }
+                        winuser::PT_PEN => {
+                            let mut pen_info: winuser::POINTER_PEN_INFO = mem::uninitialized();
```
Osspial: Ditto

src/platform_impl/windows/event_loop.rs
dam4rus: Sure. I've seen the efforts to replace `mem::unitialized` but didn't know if it's finalized or not.

PR #1142, opened 2019-09-04
Add web fullscreen support

ryanisaacg:

PR progress:
- [x] Request fullscreen makes a call to the web APIs
- [x] Fullscreen requests are aware of security requirements that fullscreen can only happen in specific event handlers
- [x] The canvas is resized when it enters and exits fullscreen, and resize events are sent to the user

This makes progress on #1072 

Currently there is an issue that the event is not processed within a user handler, so the browser rejects the attempt to fullscreen.


TannerRogalsky:
I don't think this should block this PR or feature but it's maybe worth noting that the fullscreen API works on iOS Safari on iPads only. The iPhone has no fullscreen API.

src/platform_impl/web/web_sys/canvas.rs
```diff
@@ -87,9 +122,25 @@ impl Canvas {
...
+        *self.intended_size.borrow()
+    }
+
+    pub fn window_size(&self) -> LogicalSize {
+        let window = web_sys::window().expect("Failed to obtain window");
+        let width = window.inner_width().expect("Failed to get width").as_f64().expect("Failed to get width as f64");
+        let height = window.inner_height().expect("Failed to get height").as_f64().expect("Failed to get height as f64");
+
+        LogicalSize { width, height }
     }
```
hecrj: This should probably be a backend static method: `web_sys::window_size()`.

src/platform_impl/web/event_loop/window_target.rs
```diff
@@ -163,5 +162,23 @@ impl<T> WindowTarget<T> {
                 },
             });
         });
+
+        let runner = self.runner.clone();
+        let event_canvas = canvas.clone();
```
hecrj: Could we avoid cloning the whole canvas if we create the `intended_size` here and move it into the closure? We could  keep there the size before entering fullscreen by reading the size of the `raw` canvas.

src/platform_impl/web/stdweb/canvas.rs
```diff
@@ -236,4 +295,39 @@ impl Canvas {
...
+
+        self.add_event(move |event: E| {
+            handler(event);
+
+            if *wants_fullscreen.borrow() {
+                canvas.request_fullscreen();
+                *wants_fullscreen.borrow_mut() = false;
+            }
+        })
+    }
```
hecrj: I understand this is an API limitation, but it will be a bit confusing if users are buffering events and processing them in batches. If a fullscreen request happens as a result of this event processing, it will not be applied until the user interacts again, right? This is a bit unfortunate.

src/platform_impl/web/web_sys/canvas.rs
ryanisaacg: 👍 Good point

src/platform_impl/web/event_loop/window_target.rs
ryanisaacg: We do need to keep track of `intended_size` as it changes, because it can be changed by the user at any point.

src/platform_impl/web/stdweb/canvas.rs
ryanisaacg: Yeah it is an unfortunate fact of the web fullscreen APIs. I'm not sure of a better way to work around it.

src/platform_impl/web/event_loop/window_target.rs
hecrj: I was thinking of ignoring user resizes when fullscreen is enabled. Does that break something? Or would it be unintuitive?

ryanisaacg: That works for when fullscreen is enabled but the problem comes in when you have something like:

```
create window 1 at size of 800x600
set window 1's size to 640x480
fullscreen window 1
un-fullscreen window 1
```

You probably want it to return to `640x480`, which means the intended size can change during the lifetime of the program.

hecrj: Couldn't we clone `raw` instead and use it to read the `intended_size`, store it while fullscreen mode is enabled, and restore the correct size when disabled? We could wrap the raw canvas in a new type if necessary. It's just that implementing `Clone` for the current `Canvas` type feels a bit hacky because of all the event listeners.

ryanisaacg: Great point, I used that design and was able to remove the Clone impl from `Canvas`.

src/platform_impl/web/event_loop/window_target.rs
```diff
@@ -165,5 +165,33 @@ impl<T> WindowTarget<T> {
...
+                    width: raw.width() as f64,
+                    height: raw.height() as f64,
+                };
+
+                backend::window_size()
+            } else {
+                intended_size
+            };
+            raw.set_width(new_size.width as u32);
+            raw.set_width(new_size.height as u32);
```
hecrj: Whoops, the second one should be `set_height` I think.

Issue #1140, opened 2019-08-31
Cannot combine x11 window types.


Toqozz:
In X11, it is possible to have multiple `_NET_WM_WINDOW_TYPE` properties, so that you can have a combination of window types.  `winit`, however, only allows a single `XWindowType`.

```rust
let window = WindowBuilder::new()
    .with_title("A fantastic window!")
    .with_x11_window_type(XWindowType::Notification)    // this property is overridden
    .with_x11_window_type(XWindowType::Utility)         // by this one
    .build(&event_loop)
    .unwrap();
```

In Xlib, if you wanted to have a window which had types `NOTIFICATION` and `UTILITY` you'd do something like this:

```c
property[2] = XInternAtom(dsp, "_NET_WM_WINDOW_TYPE", false);
property[0] = XInternAtom(dsp, "_NET_WM_WINDOW_TYPE_NOTIFICATION", false);
property[1] = XInternAtom(dsp, "_NET_WM_WINDOW_TYPE_UTILITY", false);
XChangeProperty(dsp, win, property[2], XA_ATOM, 32, PropModeReplace, (unsigned char *) property, 2L);
```

From a user perspective, I think the most obvious behaviour would be for `WindowBuilder.with_x11_window_type()` to combine window types rather than replace them, but it seems to just overwrite the property:
https://github.com/rust-windowing/winit/blob/cf0b8babbdb39abc443c94c8a94281201e6e9b0c/src/platform/unix.rs#L367-L370

I would be happy to contribute to fix this issue if a preferred method is described.  I locally did a quick and dirty fix by pushing window types to a `Vec`, which works but allows doubling up on properties.  Perhaps a `HashSet` or bitwise OR would work better.

`winit = "0.20.0-alpha2"`

Osspial:
Thanks for the issue!

I see no reason not to do this. If you'd like to submit a PR, could you expose this by converting `XWindowType` to a [`bitflags`](https://docs.rs/bitflags/1.1.0/bitflags/) type? Also, I'd like to keep `with_x11_window_type`'s current behavior for the sake of consistency with the rest of the API. Moving to a `bitflags` type would allow us to do that while still making it easy to set multiple window types.

murarth:
Closed by #1147

PR #1147, opened 2019-09-06
Allow using multiple `XWindowType`s on X11 (#1140)

Toqozz:

---

I'm not familiar with rustdoc.  Does documentation need to be updated manually for this change?  Note that the change is breaking due to the case change of `XWindowType`.

`get_atoms()` is quite ugly, but seems necessary.

Please let me know of any changes/improvements necessary and I'll update as soon as I can.


Cargo.toml
```diff
@@ -44,7 +44,7 @@ version = "0.1.3"
 default_features = false
 features = ["display_link"]
 
-[target.'cfg(any(target_os = "ios", target_os = "windows"))'.dependencies]
+[target.'cfg(any(target_os = "ios", target_os = "windows", target_os = "linux"))'.dependencies]
```
Osspial: Quick note - this will cause the build to fail for BSD and Dragonfly systems, since those systems use our Linux backend. Replace this with `target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "openbsd", target_os = "netbsd"`, as seen later down in this file.

src/lib.rs
```diff
@@ -123,7 +123,7 @@ extern crate serde;
 #[macro_use]
 extern crate derivative;
 #[macro_use]
-#[cfg(any(target_os = "ios", target_os = "windows"))]
+#[cfg(any(target_os = "ios", target_os = "windows", target_os = "linux"))]
```
Osspial: Ditto

Cargo.toml
Toqozz: Ahh!  I didn't even think of that.  Should be resolved now.

src/platform_impl/linux/x11/util/hint.rs
```diff
@@ -20,75 +20,104 @@ impl From<bool> for StateOperation {
...
-            &DropdownMenu => b"_NET_WM_WINDOW_TYPE_DROPDOWN_MENU\0",
-            &PopupMenu => b"_NET_WM_WINDOW_TYPE_POPUP_MENU\0",
-            &Tooltip => b"_NET_WM_WINDOW_TYPE_TOOLTIP\0",
-            &Notification => b"_NET_WM_WINDOW_TYPE_NOTIFICATION\0",
-            &Combo => b"_NET_WM_WINDOW_TYPE_COMBO\0",
-            &Dnd => b"_NET_WM_WINDOW_TYPE_DND\0",
-            &Normal => b"_NET_WM_WINDOW_TYPE_NORMAL\0",
-        };
-        unsafe { xconn.get_atom_unchecked(atom_name) }
+    pub(crate) fn get_atoms(&self, xconn: &Arc<XConnection>) -> Vec<ffi::Atom> {
```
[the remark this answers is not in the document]
Toqozz: So requiring a slice even if they only want a single window type?  Or should this be hidden from users unless they want multiple window types?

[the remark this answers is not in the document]
Toqozz: How would this change face the user?  --  `with_x11_window_type()`.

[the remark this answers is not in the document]
ZeGentzy: Changed my mind on the `AsRef`, but we should still be accepting an slice of them.

So:
`fn with_x11_window_type(self, x11_window_type: &[utils::WindowType]) -> WindowBuilder`
`fn set_window_type(&self, window_type: &[util::WindowType]) -> util::Flusher<'_>`

and so forth.

Toqozz:
I've pushed the slice variant.  A static lifetime seems like the neatest way to do this, but stops users adding window types dynamically -- which is fine, IMO.  I admit I'm not too confident with lifetimes so maybe this has worse consequences.

Something I noticed is that the window type is not set at all if the window type is the default:
https://github.com/rust-windowing/winit/blob/b03e589987d7c95b526ad5d85ae8f1f3d0f5d050/src/platform_impl/linux/x11/window.rs#L299-L301

Either this behaviour or the documentation should be changed, since the documentation states: ```Build window with `_NET_WM_WINDOW_TYPE` hint; defaults to `Normal`. Only relevant on X11.```
Is it ok to just hack this into the `if` statement?  Manually `impl Default for PlatformSpecificWindowBuilderAttributes` sounds more elegant but a couple things already seem to work around this in the same file.  Note that the code is different with this commit, so we can't just remove the `if`:
https://github.com/rust-windowing/winit/blob/1e83d9d2c3567f1a5543cdb8d3661f619ba86aa4/src/platform_impl/linux/x11/window.rs#L301-L303

src/platform_impl/linux/x11/util/hint.rs
```diff
@@ -20,104 +20,75 @@ impl From<bool> for StateOperation {
...
-        }
-        if self.contains(WindowType::NORMAL) {
-            vec.push(unsafe { xconn.get_atom_unchecked(b"_NET_WM_WINDOW_TYPE_NORMAL\0") });
-        }
-
-        vec
+    pub(crate) fn as_atom(&self, xconn: &Arc<XConnection>) -> ffi::Atom {
+        use self::WindowType::*;
+        let atom_name: &[u8] = match self {
+            &Desktop => b"_NET_WM_WINDOW_TYPE_DESKTOP\0",
```
ZeGentzy: `match *self`, then no need for the `&` at the start of everything.

src/platform/unix.rs
```diff
@@ -320,8 +320,8 @@ pub trait WindowBuilderExtUnix {
     fn with_class(self, class: String, instance: String) -> WindowBuilder;
     /// Build window with override-redirect flag; defaults to false. Only relevant on X11.
     fn with_override_redirect(self, override_redirect: bool) -> WindowBuilder;
-    /// Build window with `_NET_WM_WINDOW_TYPE` hint; defaults to `Normal`. Only relevant on X11.
-    fn with_x11_window_type(self, x11_window_type: XWindowType) -> WindowBuilder;
+    /// Build window with `_NET_WM_WINDOW_TYPE` hints; defaults to `Normal`. Only relevant on X11.
+    fn with_x11_window_type(self, x11_window_type: &'static [x11::util::WindowType]) -> WindowBuilder;
```
ZeGentzy: `&'static`? Adding a lifetime to the `WindowBuilder` I think is better then `&'static`. glutin does something similar with `ContextBuilder`, so it wouldn't be out of place.

ZeGentzy:
> I've pushed the slice variant. A static lifetime seems like the neatest way to do this, but stops users adding window types dynamically -- which is fine, IMO. I admit I'm not too confident with lifetimes so maybe this has worse consequences.

I'd rather not have it `&'static`. I think giving `WindowBuilder` a lifetime like glutin's `ContextBuilder`  is the best option, but if that cannot be done (any objections to adding a lifetime, @Osspial?), I'd rather we accept a vec. 

> Something I noticed is that the window type is not set at all if the window type is the default:
> 
> https://github.com/rust-windowing/winit/blob/b03e589987d7c95b526ad5d85ae8f1f3d0f5d050/src/platform_impl/linux/x11/window.rs#L299-L301
> 
> Either this behaviour or the documentation should be changed, since the documentation states: `` Build window with `_NET_WM_WINDOW_TYPE` hint; defaults to `Normal`. Only relevant on X11. ``
> 
> Is it ok to just hack this into the `if` statement? Manually `impl Default for PlatformSpecificWindowBuilderAttributes` sounds more elegant but a couple things already seem to work around this in the same file. Note that the code is different with this commit, so we can't just remove the `if`:
> 
> https://github.com/rust-windowing/winit/blob/1e83d9d2c3567f1a5543cdb8d3661f619ba86aa4/src/platform_impl/linux/x11/window.rs#L301-L303

I don't really know. cc  @murarth

src/platform/unix.rs
Toqozz: You're right, it's less of a deal than I thought it was.

Osspial:
> any objections to adding a lifetime, @Osspial?

Not in principal, but I don't think forcing it via platform-specific extension traits is a good place to do it - it'll force us to hack in lifetimes for `PlatformSpecificWindowBuilderAttributes` on all other platforms to support a relatively minor feature. Creating a window is a fairly rare and expensive operation anyway, so I don't think asking for another allocation will have much impact.

murarth:
> https://github.com/rust-windowing/winit/blob/1e83d9d2c3567f1a5543cdb8d3661f619ba86aa4/src/platform_impl/linux/x11/window.rs#L301-L303

According to the WM spec (at the bottomf of [this section](https://standards.freedesktop.org/wm-spec/wm-spec-latest.html#idm140200472629520)), if no `_NET_WM_WINDOW_TYPE` hint is provided, the WM **must** assume a type of `_NET_WM_WINDOW_TYPE_NORMAL`, so it's perfectly fine for winit documentation to state that `Normal` is the default and to provide no hints to the WM if the user has not provided any to winit.

The replacement `if` statement appears correct to me and the comment preceding it should be removed or changed to indicate that the WM is required to assume `_NET_WM_WINDOW_TYPE_NORMAL` if none is provided.

Toqozz:
> According to the WM spec (at the bottom of [this section](https://standards.freedesktop.org/wm-spec/wm-spec-latest.html#idm140200472629520)), if no `_NET_WM_WINDOW_TYPE` hint is provided, the WM **must** assume a type of `_NET_WM_WINDOW_TYPE_NORMAL`, so it's perfectly fine for winit documentation to state that `Normal` is the default and to provide no hints to the WM if the application has not provided any to winit.

I think this is a bit obtuse.  If the documentation says the window type defaults to `Normal`, one would expect winit to apply the `Normal` hint, right?  At the very least the wording should be changed to something like ```Build window with `_NET_WM_WINDOW_TYPE` hint.  If no hint is set, the window manager will (should) assume `Normal`. Only relevant on X11.```, but this sounds weird because we'd be doing something weird.

Pretty much every application I can `xprop` has the `_NET_WM_WINDOW_TYPE_NORMAL` atom.  I don't think it's even possible to leave it unset in gtk.  It doesn't make sense for winit to be an exception here.

The more I think about it the more I want to `impl Default for PlatformSpecificWindowBuilderAttributes`.  This way users would be able to set no window hints if desired, while having a sensible default that aligns with the rest of the ecosystem.


> Creating a window is a fairly rare and expensive operation anyway, so I don't think asking for another allocation will have much impact.

So just to confirm, `Vec<WindowType>`?

murarth:
> _NET_WM_WINDOW_TYPE_NORMAL indicates that this is a normal, top-level window, either managed or override-redirect. **Managed windows with neither _NET_WM_WINDOW_TYPE nor WM_TRANSIENT_FOR set MUST be taken as this type.** Override-redirect windows without _NET_WM_WINDOW_TYPE, must be taken as this type, whether or not they have WM_TRANSIENT_FOR set.

Toqozz:
> > _NET_WM_WINDOW_TYPE_NORMAL indicates that this is a normal, top-level window, either managed or override-redirect. **Managed windows with neither _NET_WM_WINDOW_TYPE nor WM_TRANSIENT_FOR set MUST be taken as this type.** Override-redirect windows without _NET_WM_WINDOW_TYPE, must be taken as this type, whether or not they have WM_TRANSIENT_FOR set.
> 
> (Edit: Accidentally submitted too soon.)
> 
> The "MUST" language in the spec is pretty clear, but if you want to pass `_NET_WM_WINDOW_TYPE_NORMAL` by default, I'm not opposed to that.

I edited my post a while after submitting to include a relevant section from the documentation, probably after you started typing.  Sorry!

murarth:
I think you may be misreading that portion of the spec. I don't think it's saying that the client MUST supply `_NET_WM_WINDOW_TYPE`, but that, if present, it MUST include one of the given standard atoms.

This is how the section begins:

> **This SHOULD be set by the Client** before mapping to a list of atoms indicating the functional type of the window.

A `SHOULD`; not a `MUST`. However, you're right that it is expected of a client and there's no reason not to provide the hint, so I concede the point. I agree that the default value should be `vec![WindowType::Normal]` and a manual `impl Default for PlatformSpecificWindowBuilderAttributes` is the way to go.

Toqozz:
> I think you may be misreading that portion of the spec. ...

Ahhh it seems like I am.  It's almost like I'm reading some legislation with this spec -- thanks for understanding anyway.

Hopefully we've arrived at an implementation here.  Funnily enough we've almost gone right back to what I originally had to fix the issue locally.

ZeGentzy:
Good to merge, @murarth ?

murarth:
@Toqozz This needs an entry in `CHANGELOG.md`. (It looks it had one, but it was removed in one of your revert commits.)

Otherwise, it looks great.

Toqozz:
Oops.  Resolved.

PR #1153, opened 2019-09-08
wayland: dispatch pending events before NewEvents

### LINKED ISSUE #1148: Wayland dropping events when blocking in event loop
While trying to refactor the glutin update for Alacritty a bit, I've noticed an issue with the way the event loop operates when it is blocked.

On Wayland, when the window is frequently redrawn and the event loop is blocked for short periods of time, it's possible that key release events are dropped, leading to infinite `ReceiveChar` events.

I was able to reproduce this with this minimal examlpe in Sway:

```rust
mod support;

use glutin::event::{Event, WindowEvent};
use glutin::event_loop::{ControlFlow, EventLoop};
use glutin::window::WindowBuilder;
use glutin::ContextBuilder;

fn main() {
    let el = EventLoop::new();
    let wb = WindowBuilder::new().with_title("A fantastic window!");

    let windowed_context =
        ContextBuilder::new().build_windowed(wb, &el).unwrap();

    let windowed_context = unsafe { windowed_context.make_current().unwrap() };

    println!(
        "Pixel format of the window's GL context: {:?}",
        windowed_context.get_pixel_format()
    );

    let gl = support::load(&windowed_context.context());

    el.run(move |event, _, control_flow| {
        println!("{:?}", event);

        match event {
            Event::LoopDestroyed => return,
            Event::WindowEvent { ref event, .. } => match event {
                WindowEvent::ReceivedCharacter(c) => {
                    std::thread::sleep_ms(10);
                    gl.draw_frame([1.0, 0.5, 0.7, 1.0]);
                    windowed_context.swap_buffers().unwrap();
                }
                WindowEvent::Resized(logical_size) => {
                    let dpi_factor =
                        windowed_context.window().hidpi_factor();
                    windowed_context
                        .resize(logical_size.to_physical(dpi_factor));
                }
                WindowEvent::RedrawRequested => {
                    std::thread::sleep_ms(10);
                    gl.draw_frame([1.0, 0.5, 0.7, 1.0]);
                    windowed_context.swap_buffers().unwrap();
                }
                WindowEvent::CloseRequested => {
                    *control_flow = ControlFlow::Exit
                }
                _ => (),
            },
            _ => (),
        }
    });
}
```

Pinging @vberger since I've already mentioned this on #glutin and this might be relevant to one of the Wayland libraries.

According to @kchibisov, the `sleep_ms(10)` might not be enough depending on the system, he's had to change it to `30` for example. High key repeat rates are also necessary, the `30` ms were with a repeat rate of `60` and my tests were done with the maximum Sway repeat rate (I've just set it to `600`). Even then a reproduction is a bit spurious, though very reliable in my case (I can reproduce in at least 90% of attempts).

vberger:
When we are woken up from calloop(mio) wait, always attempt to
dispatch pending events even though calloop should have done it.

When using OpenGL, mesa shares our wayland socket, and also reads
from it, especially if vsync is enabled as it'll do blocking reads.
When doing so, it may enqueue events in the internal buffer of our
event queue.

As the socket has been read, mio will thus not notify it to calloop
as readable, and thus calloop will not know it needs to dispatch. In some
cases this can lead to some events being delivered much later than
they should. Combined with key repetition this can actually cause some
flooding of the event queue making this effect event worse.

Fixes #1148

### DIFF
--- a/src/platform_impl/linux/wayland/event_loop.rs
+++ b/src/platform_impl/linux/wayland/event_loop.rs
@@ -314,6 +314,25 @@ impl<T: 'static> EventLoop<T> {
             // send pending events to the server
             self.display.flush().expect("Wayland connection lost.");
 
+            // During the run of the user callback, some other code monitoring and reading the
+            // wayland socket may have been run (mesa for example does this with vsync), if that
+            // is the case, some events may have been enqueued in our event queue.
+            //
+            // If some messages are there, the event loop needs to behave as if it was instantly
+            // woken up by messages arriving from the wayland socket, to avoid getting stuck.
+            let instant_wakeup = {
+                let window_target = match self.window_target.p {
+                    crate::platform_impl::EventLoopWindowTarget::Wayland(ref wt) => wt,
+                    _ => unreachable!(),
+                };
+                let dispatched = window_target
+                    .evq
+                    .borrow_mut()
+                    .dispatch_pending()
+                    .expect("Wayland connection lost.");
+                dispatched > 0
+            };
+
             match control_flow {
                 ControlFlow::Exit => break,
                 ControlFlow::Poll => {
@@ -328,7 +347,12 @@ impl<T: 'static> EventLoop<T> {
                     );
                 }
                 ControlFlow::Wait => {
-                    self.inner_loop.dispatch(None, &mut ()).unwrap();
+                    let timeout = if instant_wakeup {
+                        Some(::std::time::Duration::from_millis(0))
+                    } else {
+                        None
+                    };
+                    self.inner_loop.dispatch(timeout, &mut ()).unwrap();
                     callback(
                         crate::event::Event::NewEvents(crate::event::StartCause::WaitCancelled {
                             start: Instant::now(),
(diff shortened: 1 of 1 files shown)


vberger:
Let it be know that this is the result of 5 epic hours of println-debugging and at least 20 recompilations of alacritty. :fire:

kchibisov:
I've mentioned one bug with this PR I'm observing in alacritty in #smithay , so probably needs some investigation.

Just for the reference, with this patch applied alacritty can freeze during resize/startup, so looks like

Screenshot_Mon 09 Sep 2019 01:58:39 AM MSK

The left instance is dead, so can't resize/kill with wm hotkey.

backtrace:
[backtrace.txt](https://github.com/rust-windowing/winit/files/3588592/backtrace.txt)

vberger:
Yes, can reproduce. I think I have a guess about what happens, looks like the patch will have to be more subtle.

vberger:
New version of the patch, which is less brutal in its approach to dispatching and should not get stuck any more.

My few attempts with alacritty locally seems to indicate that this does indeed solve it, can you confirm @kchibisov ?

kchibisov:
Seems like I can't repro anymore, was spamming terminals with repeat_rate 255 and instantly closing them, fullscreening, resizing and so on. Nice!

Osspial:
@vberger Hey, it doesn't look like this ever got a changelog entry. Could you add that in?

vberger:
Ah sorry about that, will do.

Issue #705, opened 2018-11-13
crash on suspend/resume on laptop


chutz:
Found via https://github.com/jwilm/alacritty/ occurs in the example program as well, when I suspend and resume my laptop, winit based programs crash with the following message:
```
thread 'main' panicked at '[winit] Failed to find any monitors using XRandR.', libcore/option.rs:989:5
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

vberger:
Given the error message, this can only have been generated at either of these two places:

https://github.com/tomaka/winit/blob/2f7321a07628f24f26b7c368e9b4be20b3ff4dfc/src/platform/linux/x11/monitor.rs#L107-L111

or

https://github.com/tomaka/winit/blob/2f7321a07628f24f26b7c368e9b4be20b3ff4dfc/src/platform/linux/x11/monitor.rs#L220-L225

which means that `self.get_available_monitors()` has either returned an empty list or that the list did not contain a monitor marked primary.

A proper backtrace with debug symbols would indeed help.

francesca64:
> or that the list did not contain a monitor marked primary.

I should've included a comment in `get_primary_monitor` explaining this, but if no monitors are reported as primary, we always mark the first one:
https://github.com/tomaka/winit/blob/8dcd5143935cedcefeef5f08e17d5b076dcfbfbf/src/platform/linux/x11/monitor.rs#L191-L197

So this can only be the result of the monitor list being empty. Since winit is aware of any changes to what monitors are connected (something added in recent versions), it's likely that resuming has us receive a state update before the hardware is available again, resulting in an empty list.

chutz:
That sounds like a pretty reasonable theory, often the GPUs take longer to come back than the CPU, so it is pretty probable that the monitor list could be empty for a period of time while the GPU is brought back up.

I am at a conference at the moment, so rebuilding with debug symbols is a little difficult. I am happy to test patches etc though if you have a theoretical fix.

francesca64:
I unfortunately can't reproduce this on my desktop.

> so rebuilding with debug symbols is a little difficult

I doubt debug symbols would help anyway, so no worries.

The most straightforward fix would be to make the monitor check more lenient, i.e. not to panic until the monitor data would actually need to be accessed. I do have some concerns over the correctness of doing that, but at the same time, I doubt the protocol gives us any real way to deal with this problem.

jonpas:
[...]

Issue #1007, opened 2019-07-04
Breaks on xrandr < 1.5


jarredbarber:
I'm trying to run using the Cygwin X11 server packaged with MobaXterm, which supports RANDR 1.4;

```
jarred@greywind:~$ xrandr --version
xrandr program version       1.5.0
Server reports RandR version 1.4

jarred@greywind:~/projects/winit$ xrandr --listmonitors
RandR 1.5 not supported

jarred@greywind:~/projects/winit$ cargo run --example window
   Compiling winit v0.20.0-alpha1 (/home/jarred/projects/winit)
    Finished dev [unoptimized + debuginfo] target(s) in 12.09s
     Running `target/debug/examples/window`
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: OsError { line: 106, file: "src/platform_impl/linux/x11/window.rs", error: XMisc("No monitors were detected.") }', src/libcore/result.rs:997:5
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
jarred@greywind:~/projects/winit$
```

I tried changing the error on line 106 in `window.rs` to just return a default DPI scale factor of 1.0, but that results in deeper "can't find monitor" errors:

```
jarred@greywind:~/projects/winit$ cargo run --example window
   Compiling winit v0.20.0-alpha1 (/home/jarred/projects/winit)
    Finished dev [unoptimized + debuginfo] target(s) in 10.21s
     Running `target/debug/examples/window`
EventsCleared
thread 'main' panicked at '[winit] Failed to find any monitors using XRandR.', src/libcore/option.rs:1034:5
```

I tried changing `FORCE_RANDR_COMPAT` to `true` in `monitors.rs`, but that didn't seem to fix anything.

ZeGentzy:
cc @murarth

felixrabe:
This is where the `platform = X11` vs `platform = Windows` breaks down, because this is both.

Maybe we need a `platform = Cygwin` (or `CygwinX11`) too.

ZeGentzy:
My, I didn't notice this said Cygwin.

jarredbarber:
It's running under WSL, so I guess X11 is more appropriate - the code thinks it's all linux

murarth:
I don't have a Windows installation, so I don't have any way to investigate this.

All I can contribute is that I've just tested the xrandr < 1.5 compatibility code and it still works for me, so this does appear to be a Cygwin-specific issue.

[...]

PR #1158, opened 2019-09-11
X11: Fix panic when no monitors are available

murarth:


murarth:
Fixes #705 by returning a dummy `MonitorHandle` when no monitors are available. This seemed like the most straightforward fix, as the platform-independent facade would not permit returning an `Option<MonitorHandle>` from `Window::current_monitor`. The dummy `MonitorHandle` has an `id` value of `0` (an invalid `XID` value) and a few checks are added to avoid caching this dummy handle or using it as though it were valid.

I was unable to reproduce the original issue myself, so I hope that someone still experiencing the issue will test this and let me know if any new bugs occur as a result of this change.

vberger:
Overall, what would happen if a user unknowingly tries to use this dummy monitor handle as it were the main monitor for their app, by passing it to `set_fullscreen`, or by calling the methods of `MonitorHandle` to get its name / position / dpi factor?

murarth:
Replaced a `return Err` with a default DPI factor of 1.0 to avoid another panic. This should address #1007.

ZeGentzy:
Should be good to merge when CI is done?

murarth:
@ZeGentzy Yes, I think so.

PR #1165, opened 2019-09-15
Fix surrogate pair handling on Windows

### LINKED ISSUE #1164: Cannot enter unicode characters like emojis on Windows
When entering emojis using the Windows emoji picker or the compose sequence, the right unicode character is not sent to the application.

A `ReceivedCharacter` event is triggered, however it seems like there might be issues with UTF-16 and UTF-8 conversion, since it's received as `'\u{d83d}'` and `'\u{de00}'` which is the UTF-16 representation of the chosen emoji (see https://unicode-table.com/de/1F600/).

More details can be found in the Alacritty issue: https://github.com/jwilm/alacritty/issues/2796

Osspial:

Should fix #1164 and https://github.com/jwilm/alacritty/issues/2796.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,8 @@
 - On Windows, add touch pressure information for touch events.
 - On macOS, differentiate between `CursorIcon::Grab` and `CursorIcon::Grabbing`.
 - On Wayland, fix event processing sometimes stalling when using OpenGL with vsync.
-- Officially remove the Emscripten backend
+- Officially remove the Emscripten backend.
+- On Windows, fix handling of surrogate pairs when dispatching `ReceivedCharacter`.
 
 # 0.20.0 Alpha 3 (2019-08-14)
 
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -1031,11 +1031,29 @@ unsafe extern "system" fn public_window_callback<T>(
 
         winuser::WM_CHAR => {
             use crate::event::WindowEvent::ReceivedCharacter;
-            let chr: char = mem::transmute(wparam as u32);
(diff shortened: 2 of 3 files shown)


chris-morgan:
This patch mangles mismatched high/low surrogates. Admittedly this is a corner case and indicates buggy software somewhere anyway, so if there’d be a concrete performance hazard to doing it that way I could live without it, but I prefer the style of the patch I wrote for gVim: on a normal character, flush any pending high surrogate, and on a low surrogate with no pending high surrogate, send the low surrogate.

Osspial:
Ah yeah, that first case is a problem. I'll go ahead and address that. I've gotta ask, though - why would it be a good idea to send the mismatched low surrogate? There's nothing sensible you can do with that data, and the Unicode standard [explicitly says not to encode surrogate pairs if you aren't UTF-16](https://unicodebook.readthedocs.io/unicode_encodings.html#utf-16-surrogate-pairs).

chris-morgan:
Mismatched surrogates are exceptional, when you have a buggy IME. I don’t particularly expect to encounter them in the wild, but I’m confident that some users will encounter them at some point in time, and I prefer to pass through bad input rather than swallowing it or turning into even worse input. That way you can build tools like gVim atop it, and receive the mismatched surrogates and do something sane with them—or at least inspect what they are.

I have just realised, however, that `char` is a Unicode scalar rather than a Unicode code point, so we actually can’t represent mismatched surrogates in a `char`. So anything that transmuted a surrogate into a `char` was actually wrong already. Hmm, perhaps winit actually *should* just drop such input on the floor.

Osspial:
Actually, now that you bring that up, I have no idea why we're using a transmute there. The standard library provides perfectly good char conversion methods so we really should be using those.

PR #1166, opened 2019-09-16
Flush high surrogate if not followed by low surrogate

Osspial:
Addresses https://github.com/rust-windowing/winit/pull/1165#issuecomment-531628475. Shouldn't have an impact on properly-formed input, but will cause us to ignore all malformed surrogate pairs.

### DIFF
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -1031,12 +1031,12 @@ unsafe extern "system" fn public_window_callback<T>(
 
         winuser::WM_CHAR => {
             use crate::event::WindowEvent::ReceivedCharacter;
-            let high_surrogate = 0xD800 <= wparam && wparam <= 0xDBFF;
-            let low_surrogate = 0xDC00 <= wparam && wparam <= 0xDFFF;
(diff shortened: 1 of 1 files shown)


src/platform_impl/windows/event_loop.rs
```diff
@@ -1048,6 +1048,8 @@ unsafe extern "system" fn public_window_callback<T>(
                     }
                 }
             } else {
+                subclass_input.window_state.lock().high_surrogate = None;
```
chris-morgan: You will note that you are now using `subclass_input.window_state.lock()` on all three branches. Consider extracting it.

Osspial: I prefer the current style, since it unlocks the mutex immediately after setting `high_surrogate` and it easily ensures that `window_state` isn't locked while the event handler is running. If we pulled that lock out of the `if` block, we'll have to explicitly drop it before the `send_event` call and overall that doesn't seem like much of a readability win.

Incidentally, this review made me realize that `window_state` was getting locked during the `ReceivedCharacter` handler for surrogate pairs, so I've been able to fix that before this went out live to anyone!

PR #1168, opened 2019-09-16
Fix panic upon closing the app on iOS

aleksijuvani:


aleksijuvani:
r? @mtak- 
I'm not sure if this should be waiting for both `ProcessingEvents` and `ProcessingRedraws`. For me this was only ever called with `ProcessingEvents`, but I'm not using a directly Metal-backed view (instead it's a sublayer), so perhaps I'm triggering a different code path than your application.

Osspial:
Is this good to merge? cc @mtak- again

Osspial:
Are there any example Winit Metal applications this could be tested with?

aleksijuvani:
It looks like glutin is creating a directly GL-backed view for the window, so this can be tested with the glutin examples. The changes here look correct. The panic can be reproduced with these examples, and the issue is fixed after applying the changes in this pull request.

PR #1173, opened 2019-09-17
Fix events not being emitted during modal loops on macOS

### LINKED ISSUE #1162: EventLoop in 0.20.0-alpha3 gets clogged on resize events on macos
I was running into this issue on my codebase, but @Osspial suggested I run the window example to see if that gets stuck too. Looking at the printouts from that, when I continuously resize the window, it gets stuck on a bunch of them. When I let go, it continues to render again.

aleksijuvani:
Fixes #1162

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 - On Windows, fix handling of surrogate pairs when dispatching `ReceivedCharacter`.
 - On macOS 10.15, fix freeze upon exiting exclusive fullscreen mode.
 - On iOS, fix null window on initial `HiDpiFactorChanged` event.
+- On macOS, fix events not being emitted during modal loops, such as when windows are being resized
+  by the user.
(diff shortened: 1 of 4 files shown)


vbogaevsky:
when this PR is merged we well have to fix `HiDpiChangedEvent` implemented in #997 since it is built around `deferred_events` queue in `app_state::Hander`.

aleksijuvani:
I've removed `deferred_events` here as it was only ever used for sending out the resize events during the modal resize loop. Is there a reason the `HiDpiFactorChanged` event should be sent immediately, or can we refactor #997 to go through the regular event queue, i.e. `emit_event` instead of `send_events_immediately`?

vbogaevsky:
I think, it should not be a problem. If we can process resize events with `emit_event`, we should be able to do the same for `HiDpiFactorChanged`.

PR #1176, opened 2019-09-18
Prevent stealing focus on new windows

### LINKED ISSUE #1160: No way to avoid focus stealing on X11.
There seems to be no way to map a window without focusing it.  I've only tested this on X11, but it may be present on other platforms as well given there is no `WindowBuilder` option to disable the behaviour.

On X11, this code is the culprit:
https://github.com/rust-windowing/winit/blob/b03e589987d7c95b526ad5d85ae8f1f3d0f5d050/src/platform_impl/linux/x11/window.rs#L353-L358

[`XMapRaised`](https://tronche.com/gui/x/xlib/window/XMapRaised.html) pushes the window to the top of the stack as well as maps the window.

There should be an option to disable this behaviour and map the window normally with `XMapWindow()`.

murarth:
Previously, `XSetInputFocus` was called unconditionally on new windows (so long as they were visible upon creation). This was a bugfix for fullscreen windows not receiving input and locking up the user's display (rust-windowing/glutin#509). This PR moves the `XSetInputFocus` call into `Window::set_fullscreen`, preventing non-fullscreen windows from stealing focus, which may not be desired, e.g. for tooltip or notification type windows.

Fixes #1160

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@
 - On macOS, fix events not being emitted during modal loops, such as when windows are being resized
   by the user.
 - On Windows, fix hovering the mouse over the active window creating an endless stream of CursorMoved events.
+- On X11, prevent stealing input focus when creating a new window.
+  Only steal input focus when entering fullscreen mode.
 
 # 0.20.0 Alpha 3 (2019-08-14)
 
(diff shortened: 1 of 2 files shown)


murarth:
Focus stealing in X11 is determined by the window manager, influenced by window type hint. Some window types should never steal focus, such as tooltips and notifications, while others usually do take focus.

This PR corrects a bug introduced by a ~4 year old bugfix which gives *all* new windows focus. The new behavior gives control back to the WM, while preventing a regression on the original bug -- a window will always gain focus when entering fullscreen mode so that it doesn't lock up the user's display.

I suppose the title is misleading -- this PR does not prevent any new window from stealing focus; it only no longer gives focus to windows that are not intended to ever have input focus.

Osspial:
Ah, thanks for the clarification! This seems good to go, then.

Issue #1178, opened 2019-09-21
Not receiving DeviceEvent::MouseMotion events on Wayland


andersrein:
The DeviceEvent::MouseMotion events are not being sent to the EventLoop on Wayland. I tracked it down implement_relative_pointer never being called. This was in turn caused by the relative_pointer_manager_proxy not being set on the SeatData object. In the end it turns out that "zwp_relative_pointer_manager_v1" GlobalEvent was fired after the relevant "wl_seat" GlobalEvent.

andersrein:
Fixed in #1179

PR #1179, opened 2019-09-21
Wayland: Switched to using a reference to relative_pointer_manager_proxy when creating SeatData

andersrein:
Changed SeatData to obtain a reference to the relative pointer manager proxy in SeatManager. This solves the problem that when the GlobalEvent for the "zwp_relative_pointer_manager_v1" interface comes after the "wl_seat" interfaces the SeatData still has a way to obtain the relative_pointer_manager_proxy that was created when receiving "zwp_relative_pointer_manager_v1". This is a fix for #1178

### DIFF
--- a/src/platform_impl/linux/wayland/event_loop.rs
+++ b/src/platform_impl/linux/wayland/event_loop.rs
@@ -148,7 +148,7 @@ impl<T: 'static> EventLoop<T> {
 
         let mut seat_manager = SeatManager {
             sink: sink.clone(),
-            relative_pointer_manager_proxy: Arc::new(Mutex::new(None)),
+            relative_pointer_manager_proxy: Rc::new(RefCell::new(None)),
             store: store.clone(),
             seats: seats.clone(),
             kbd_sender,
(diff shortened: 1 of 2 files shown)


andersrein:
#1179

andersrein:
#1178

src/platform_impl/linux/wayland/event_loop.rs
```diff
@@ -148,7 +148,7 @@ impl<T: 'static> EventLoop<T> {
 
         let mut seat_manager = SeatManager {
             sink: sink.clone(),
-            relative_pointer_manager_proxy: None,
+            relative_pointer_manager_proxy: Arc::new(Mutex::new(None)),
```
vberger: Maybe I'm missing something, but why is `Arc<Mutex<_>>` necessary? Wouldn't `Rc<RefCell<_>>` be sufficient? It seems to me that the `SeatManager` and `SeatData` always remain on the same thread ad the `EventLoop`.

vberger:
Apart from the little point I raised, this looks good! :+1:

src/platform_impl/linux/wayland/event_loop.rs
andersrein: Of course. That makes senses. Updated the pull request to use RefCell and Rc instead.

PR #1180, opened 2019-09-21
Wayland support for set_cursor_grab and set_cursor_visible

andersrein:
Added support for set_cursor_grab and set_cursor_visible. There is currently a know limitation that if when making the cursor visible it does not take effect until the cursor re-enters the window surface.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - On Windows, fix handling of surrogate pairs when dispatching `ReceivedCharacter`.
 - On macOS 10.15, fix freeze upon exiting exclusive fullscreen mode.
 - On iOS, fix null window on initial `HiDpiFactorChanged` event.
+- On Windows, fix hovering the mouse over the active window creating an endless stream of CursorMoved events.
 
 # 0.20.0 Alpha 3 (2019-08-14)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 - On X11, prevent stealing input focus when creating a new window.
   Only steal input focus when entering fullscreen mode.
 - Added support for set_cursor_visible and set_cursor_grab on wayland
+- On Wayland, fixed DeviceEvents for relative mouse movement is not always produced
 
 # 0.20.0 Alpha 3 (2019-08-14)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,11 +11,13 @@
 - Officially remove the Emscripten backend.
 - On Windows, fix handling of surrogate pairs when dispatching `ReceivedCharacter`.
 - On macOS 10.15, fix freeze upon exiting exclusive fullscreen mode.
+- On X11, allow setting mulitple `XWindowType`s.
 - On iOS, fix null window on initial `HiDpiFactorChanged` event.
 - On Windows, fix fullscreen window shrinking upon getting restored to a normal window.
 - On macOS, fix events not being emitted during modal loops, such as when windows are being resized
   by the user.
 - On Windows, fix hovering the mouse over the active window creating an endless stream of CursorMoved events.
+- On X11, return dummy monitor data to avoid panicking when no monitors exist.
 - On X11, prevent stealing input focus when creating a new window.
   Only steal input focus when entering fullscreen mode.
 - Added support for set_cursor_visible and set_cursor_grab on wayland
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - On Windows, fix handling of surrogate pairs when dispatching `ReceivedCharacter`.
 - On macOS 10.15, fix freeze upon exiting exclusive fullscreen mode.
 - On iOS, fix null window on initial `HiDpiFactorChanged` event.
+- On Windows, fix fullscreen window shrinking upon getting restored to a normal window.
 - On macOS, fix events not being emitted during modal loops, such as when windows are being resized
   by the user.
 - On Windows, fix hovering the mouse over the active window creating an endless stream of CursorMoved events.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,8 @@
 - On Windows, add touch pressure information for touch events.
 - On macOS, differentiate between `CursorIcon::Grab` and `CursorIcon::Grabbing`.
 - On Wayland, fix event processing sometimes stalling when using OpenGL with vsync.
-- Officially remove the Emscripten backend
+- Officially remove the Emscripten backend.
+- On Windows, fix handling of surrogate pairs when dispatching `ReceivedCharacter`.
 - On macOS 10.15, fix freeze upon exiting exclusive fullscreen mode.
 
 # 0.20.0 Alpha 3 (2019-08-14)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@
 - On macOS, fix events not being emitted during modal loops, such as when windows are being resized
   by the user.
 - On Windows, fix hovering the mouse over the active window creating an endless stream of CursorMoved events.
+- On X11, prevent stealing input focus when creating a new window.
+  Only steal input focus when entering fullscreen mode.
 
 # 0.20.0 Alpha 3 (2019-08-14)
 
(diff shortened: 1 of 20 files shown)


src/platform_impl/linux/wayland/event_loop.rs
```diff
@@ -69,6 +74,68 @@ impl<T> WindowEventsSink<T> {
...
+    }
+
+    fn register_pointer(&mut self, pointer: wl_pointer::WlPointer) {
+        self.pointers.push(pointer);
+    }
+
+    pub fn set_cursor_visible(&mut self, visible: bool) {
+        // If we want to hide the cursor do it immediately. There is a known
+        // limitation that the cursor will not become visible again until it
+        // enters a new surface
```
vberger: Please, explain a bit more this "known limitation", especially where does it come from? Is it an inherent limitation from the wayland protocol? Is it a `TODO` that should be improved in the future?

It seems that the rest of your code makes no attempt to set the pointer visible when requested, so I'd really prefer this to be explained.

src/platform_impl/linux/wayland/pointer.rs
```diff
@@ -241,3 +253,13 @@ pub fn implement_relative_pointer<T: 'static>(
         )
     })
 }
+
+pub fn implement_locked_pointer(
+    surface: &WlSurface,
+    pointer: &WlPointer,
+    constraints: &ZwpPointerConstraintsV1,
+) -> Result<ZwpLockedPointerV1, ()> {
+    constraints.lock_pointer(surface, pointer, None, Lifetime::Persistent.to_raw(), |c| {
```
vberger: This need for `to_raw()` here is a deficiency in the code generated by wayland-protocols, it self caused by deficiency in the protocol XML files :/ I have "fixing this" somewhere in my todo list.

src/platform_impl/linux/wayland/event_loop.rs
andersrein: What I meant was that it is a known limitation in my winit implementation because I was too lazy to figure out how to obtain the default cursor surface which I would need to pass into set_cursor call. 

I figured it would be good enough for now because wayland will revert to the default cursor on the next PtrEvent::Enter. 

It's definitively a `TODO` though. Should I change the comment to me more clear? Or do we absolute need the feature of being able to make the cursor visible again right away? Maybe obtaining the default cursor WlSurface is trivial, but I am having a hard time finding documentation on these thing.

kchibisov: I'm not sure if I'm correct here, but I was doing kind of the same thing when implementing this `set_visible`, expect I was using `AutoThemer` and `AutoPointer` from SCTK to change cursor, which is quite easy to manage cursor and overcome this limitation iirc. Since my PR is kind of WIP and private, and your one was earlier you can try adopting `AutoPointer`/`AutoThemer` approach.

andersrein:
In light of the comment about Arc Mutex vs Rc RefCell in #1179 I switched to using Rc and RefCell most places

src/platform_impl/linux/wayland/event_loop.rs
kchibisov: fwiw without setting cursor visible right away it's quite buggy feature and will need immediate fix in my opinion (At least from alacritty usage point of view).

vberger: Indeed, `AutoThemer` provided by SCTK should make the process of making the cursor visible back again easily, and also change the icon to other things (grabbing cursor, loading icon, ...).

I'm not opposed to first merging this as is, and adding AutoThemer support in a future PR, but in that case please add a clear `TODO` comment explaining what remains to be done, so that we won't forget about this.

andersrein: I updated the comments in the code and also added documentation about this in the public API.

andersrein: It seems I have gotten it to work using ThemeManager and ThemedPointer. Currently I am holding a vector of both the registered WlPointers and a vector of ThemedPointers which does not seem optimal to me. There reason being that I don't want the entire thing to stop working if for some reason the WlShm or ThemeManager fails to initialize. Is this a valid concern? Or can I assume that this will always work?

vberger: `WlShm` cannot fail to initialize, or it is a bug from the server, so we don't need to worry about this.

wrt to `ThemedPointer`, this is what [`AutoThemer`](https://docs.rs/smithay-client-toolkit/0.6.4/smithay_client_toolkit/pointer/enum.AutoThemer.html) is for: it gracefully handles a failure to initialize a `ThemeManager`. And in that case, you only need to keep the `AutoPointer`s, as they give you access to the `WlPointer`s via deref.

andersrein: I've switched to using `AutoThemer` and `AutoPointer`. Now the `CursorManager` only stores a vector of `AutoPointer` objects which I think is better.

I encountered some difficulties implementing this. The first problem was that the second call to `sync_roundtrip` inside `from_display_with_cb` triggered the registering of the pointer, but I was only initializing the `AutoThemer` after creating the environment. This caused caused panic when registering the pointer in `CursorManager` because the `AutoThemer` was not set when registering the pointer.

To work around that I moved the initialization of the `AutoThemer` into the `Environment` callback. Then there was the problem that you need both shm and compositor to initialize `AutoThemer`. So I had to set up the compositor interface as well and only trigger the code to initialize the `AutoThemer` once both the compositor and the shm interface was initialized.

Could it be a good idea to have some way for the caller to run some code after all the global interface callbacks have been run, but before the second call to `sync_roundtrip` ? Or maybe I am just using this API completely wrong?

Either way, I think all the mentioned issues has been resolved in this pull request.

vberger: This is likely that SCTK's API has room for improvement. I've actually opened an issue to gather feedback there: https://github.com/Smithay/client-toolkit/issues/77.

src/platform_impl/linux/wayland/event_loop.rs
```diff
@@ -172,6 +265,39 @@ impl<T: 'static> EventLoop<T> {
...
+                        (*compositor_cell.borrow_mut()) = Some(compositor);
+                    }
+
+                    if compositor_cell.borrow().is_some() && shm_cell.borrow().is_some() {
+                        let compositor = compositor_cell.borrow_mut().take().unwrap();
+                        let shm = shm_cell.borrow_mut().take().unwrap();
+                        let auto_themer = AutoThemer::init(None, compositor, &shm);
+                        cursor_manager2.borrow_mut().set_auto_themer(auto_themer);
+                    }
+
```
vberger: The `Environment` callback is mainly targeted for globals that can appear after the initial setup (like seats or outputs). For functionality globals that are either always present or absent depending on whether the server supports it, it'll probably be easier to try & get them though `[env.manager.instanciate_exact(..)`](https://docs.rs/wayland-client/0.23.5/wayland_client/struct.GlobalManager.html#method.instantiate_exact) once the `Environment` has been created.

kchibisov:
fwiw I can confirm that this PR works fine in alacritty. Ideally we should set to a previous cursor instead of `left_ptr`, but since this functionality is not implemented anyway, so we can go with `left_ptr` for now and add it later.

src/platform_impl/linux/wayland/event_loop.rs
andersrein: I would like to use the environment to obtain the compositor and shm, but I need the AutoThemer when the pointers are registered. This happens during initialization of the Environment.

vberger: Ah right. Well, this is the kind of dependencies between globals that could probably be better handled by SCTK directly I guess.

andersrein:
Any updates on this? What remains to do in order to get this reviewed and merged in?

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@
 - On X11, return dummy monitor data to avoid panicking when no monitors exist.
 - On X11, prevent stealing input focus when creating a new window.
   Only steal input focus when entering fullscreen mode.
+- Added support for set_cursor_visible and set_cursor_grab on wayland
```
kchibisov:
```suggestion
- On Wayland, add support for set_cursor_visible and set_cursor_grab.
```

Just to make things consistent with the rest of the CHANGELOG

ZeGentzy:
Thanks! Merging!

PR #1182, opened 2019-09-23
X11: Fix window position setting ignored when fullscreen

murarth:
When `Window::set_outer_position` is called while a window is fullscreen,
the given position will be saved and will be assigned to the window when
it leaves fullscreen mode.

This addresses one of the issues described in #978.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 - On X11, prevent stealing input focus when creating a new window.
   Only steal input focus when entering fullscreen mode.
 - On Wayland, fixed DeviceEvents for relative mouse movement is not always produced
+- On X11, fix `Window::set_outer_position` being ignored when called while window is fullscreen.
 
 # 0.20.0 Alpha 3 (2019-08-14)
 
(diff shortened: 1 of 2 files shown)


Osspial:
I tested `set_outer_position` on Windows when fullscreened, and we get similarly nonsensical behavior there. I feel like we should standardize and document the behavior of window position/size manipulation functions on fullscreen windows across platforms, to avoid gotchyas when developers are only regularly testing one platform.

Is there any reason `set_outer_position` and `set_inner_size` shouldn't just return hard errors when the window is in fullscreen mode? I don't think overloading `set_outer_position` as `set_restore_position` when the window is fullscreened is a good idea.

murarth:
> Is there any reason `set_outer_position` and `set_inner_size` shouldn't just return hard errors when the window is in fullscreen mode? I don't think overloading `set_outer_position` as `set_restore_position` when the window is fullscreened is a good idea.

cc @chrisduerr, author of the original issue

chrisduerr:
Doesn't seem to help for me. Window is still spawned on incorrect monitor and it still has the incorrect size after unfullscreening.

murarth:
Ah, right. The issue was that setting position *at startup* was ignored, which has a different root cause. I will close this PR and approach the issue from a different angle.

PR #1186, opened 2019-09-26
Prevent EventLoop from getting initialized outside the main thread in cross-platform functions

### LINKED ISSUE #1107: I've got "Unexpected GetClientRect failure"
# Unexpected GetClientRect failure
I 've got this panic on x86_64-pc-windows-msvc. I think this error has occurred just because `event_loop` dropped while `window` was alive. 
Writing this panic in the document may be good solution.

# Code
```
use winit::window::Window;
use winit::event_loop::EventLoop;

use std::thread;
use std::sync::mpsc;

fn main() {
    let (window_sender, window_receiver) = mpsc::channel();
    let handle = thread::spawn(move || {
        let window = Window::new(&event_loop).unwrap();

        window_sender.send(window).unwrap();
    });

    handle.join().unwrap();
    let window = window_receiver.recv().unwrap();
    let inner_window_size = window.inner_size(); // panic here!
    println!("{:?}", inner_window_size);
}
```

Osspial:
I've seen a decent number of issues recently where people have decided to initialize `EventLoop` outside of the main thread, without being aware that that does not work on all platforms. This PR clarifies that that does not work everywhere, and requires you to explicitly opt in to compatibility-breaking behavior.

This only applies to `EventLoop::new` and `EventLoop::with_user_event`. This PR also adds platform-specific functions to the `EventLoopExtWindows` trait for initializing the event loop outside of the main thread, and documents the precautions that must be taken when using them.

I wouldn't be surprised if this is a controversial change, but I strongly believe that it's the right direction to take for the reasons outlined above. @murarth Would you be able to implement this on Linux? It looks like the implementation would be fairly simple ([see this post](https://stackoverflow.com/questions/20530218/check-if-current-thread-is-main-thread)), but I'm not comfortable implementing semantic changes on platforms I can't test.

Closes #1107.


murarth:
I can implement this for Linux, but `gettid` is not portable to our supported BSD platforms.

I've been doing some searching, but I've found only one potentially helpful SO post: https://stackoverflow.com/questions/4867839/how-can-i-tell-if-pthread-self-is-the-main-first-thread-in-the-process

So, here's what we have so far:
* [x] Linux: `syscall(SYS_gettid) == getpid()`
* [ ] Dragonfly: Unknown
* [ ] FreeBSD: The answer suggests `pthread_main_np`, but it isn't defined in the `libc` crate for FreeBSD.
* [ ] NetBSD: The answer suggests `_lwp_self`, but I'm not sure what that is or how to use it. Nothing by that name exists in `libc`. It's possible that they're referring to the underlying system call implementing `pthread_self`; i.e. `pthread_self()` always returns `1` for the main thread.
* [x] OpenBSD: `pthread_main_np() == 1`

Osspial:
`_lwp_self` seems to be a normal libc function. I would guess that the situation with NetBSD and FreeBSD are oversights, and we can probably submit PRs to libc adding those functions. I set up a Dragonfly VM, and it looks like it also uses `pthread_main_np`.

src/event_loop.rs
```diff
@@ -30,6 +30,9 @@ use crate::{
 /// forbidding it), as such it is neither `Send` nor `Sync`. If you need cross-thread access, the
 /// `Window` created from this `EventLoop` _can_ be sent to an other thread, and the
 /// `EventLoopProxy` allows you to wake up an `EventLoop` from another thread.
+///
+/// ***For cross-platform compatibility, the `EventLoop` must be created on the main thread.*** This
+/// isn't strictly necessary on some platforms, like Windows, but attempting to create
```
mitchmindtree: Looks like this sentence doesn't quite finish?

Great initiative on this PR btw! I have also seen many folks run into this in the past.

murarth:
I wonder if it might be better to add the `new_any_thread` method to the `EventLoopExtDesktop` trait or to create a new trait (e.g. `EventLoopAnyThread`). This would facilitate applications that require this platform-dependent functionality but want to remain otherwise platform-agnostic.

Osspial:
@murarth Unfortunately, we can't add it to `EventLoopExtDesktop` since macOS can only have the event loop on the main thread. We could potentially add a new trait for that, but imo that's not worth the additional complexity.

murarth:
@Osspial Fair enough.

I've submitted a PR to `libc` for the functions we need (rust-lang/libc#1533).

Osspial:
@murarth It looks like libc `0.2.64` has been released with the changes we need.

murarth:
@Osspial Great. I've added the Linux/BSD implementation to this PR.

However, I've only tested it myself on Linux.

Osspial:
@murarth I'm not going to ask you to set up BSD environments to test this out. As long as we can successfully cross-compile this to the BSD platforms we support, I'll be alright with merging it.

src/event_loop.rs
Osspial: Thanks for catching that! I've updated the docs to not end in the middle of the

Issue #1120, opened 2019-08-25
`NSInternalInconsistencyException` upon touching the screen on iOS 13.0


aleksijuvani:
As of iOS 13.0, applications crash with `Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'Invalid parameter not satisfying: sourceNode'` in `UIGestureGraphEdge`. As far as I can tell, this appears to be because we are creating the root view controller before `UIApplicationMain`, which violates some invariants in UIKit. This no longer works in iOS 13.0. Deferring initialization of windows and their view controllers until after `UIApplicationMain` fixes the issue.

aleksijuvani:
I don't seem to be able to reproduce this anymore on my iPhone 6s after updating to the latest iOS 13.1 beta. Perhaps this has been addressed by Apple? Simulator still crashes however and there hasn't been a new version, so we'll need to wait until a new Simulator runtime is released to see.

aleksijuvani:
I can confirm that this is no longer an issue with the iOS 13.2 Simulator runtime.

We probably ought to do something about the iOS situation in the future, but that's going to be a big chunk of work that I'm not quite ready to tackle right now. I feel like we should restructure the entire way `EventLoop::run` works, so that the callback is not called per-event but instead you get a callback that is called once when the application starts and then you loop over events in that closure as they become available.

This would solve a lot of issues with the iOS backend (namely, being unable to iterate over the list of monitors before `Event::NewEvents(StartCause::Init)`, and I also worry that this same issue might come back to bite us in the future).

PR #1187, opened 2019-09-26
Release Alpha 4

Osspial:
I'd like to get #1168, #1171, #1175, #1180, #1186, and possibly #1121 in before this releases, but most of those PRs seem close to getting merged so that shouldn't be too much of a hassle.

- [x] Update release date in CHANGELOG
- [x] #1168
- [x] #1171
- [x] #1175
- [x] #1180
- [x] #1186
- [x] #1121

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 - Add web support via the 'stdweb' or 'web-sys' features
 - On Windows, implemented function to get HINSTANCE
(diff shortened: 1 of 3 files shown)


aleksijuvani:
We should wait on #1121 until a new Simulator runtime is released to see if this was fixed by Apple or if we still need to address this in winit. See https://github.com/rust-windowing/winit/issues/1120#issuecomment-531188824.

PR #1189, opened 2019-09-26
Implement `WindowBuilder::with_outer_position` for X11

murarth:

### DIFF
--- a/src/window.rs
+++ b/src/window.rs
@@ -114,6 +114,12 @@ pub struct WindowAttributes {
     /// The default is `None`.
     pub max_inner_size: Option<LogicalSize>,
 
+    /// The desired position of the window. If this is `None`,
+    /// some platform-specific position will be chosen.
+    ///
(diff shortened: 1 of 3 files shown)


src/window.rs
```diff
@@ -114,6 +114,12 @@ pub struct WindowAttributes {
     /// The default is `None`.
     pub max_inner_size: Option<LogicalSize>,
 
+    /// The desired position of the window. If this is `None`,
+    /// some platform-specific position will be chosen.
+    ///
+    /// The default is `None`.
```
Osspial: We mention in `with_outer_position` that this doesn't work on Wayland. We should mention that here, too.

murarth:
Merging to a new branch makes sense to me.

Also, I haven't added a changelog entry because I figured we could add one when it's all done instead of one for each platform as they're implemented. If this isn't the way it's usually done, let me know.

PR #1197, opened 2019-10-01
Implements a new CursorIcon element for external icons

Carbonhell:
Made the PR to show my current code. Any sort of feedback is appreciated, I'm a newbie both in rust and windows api so I expect a lot of issues in my code.


src/window.rs
```diff
@@ -785,6 +785,9 @@ pub enum CursorIcon {
     NwseResize,
     ColResize,
     RowResize,
+
+    /// Represents a custom cursor which uses an image.
```
ZeGentzy: Can we get a comment saying it's currently only windows only?

ZeGentzy:
A changelog entry also would not hurt.

Carbonhell:
Done. I hope I've put the changelog entry in the right place (Unreleased).

ZeGentzy:
Hey @0xpr03, want to test this?

0xpr03:
Which platform ?

kchibisov:
@0xpr03 Windows.

0xpr03:
Only on VM, back on windows next week, sry

kchibisov:
Ok, I can test this one myself ( I have windows around ). Probably VM is enough though.

kchibisov:
From my testing with cursor.rs example it seems fine.

It handles visible change fine and seems like was loaded correctly. I'm not familiar with Windows at all though. Let me know if you're expecting a different kind of testing here.

src/window.rs
```diff
@@ -788,6 +788,10 @@ pub enum CursorIcon {
     NwseResize,
     ColResize,
     RowResize,
+
+    /// Represents a custom cursor which uses an image. At the current time, it's only implemented for Windows.
+    #[cfg(target_os = "windows")]
+    Custom(&'static str),
```
ZeGentzy: I think we should mention the accepted formats in this things docs. 

Also, for consistency's sake, we should remove the `#[cfg(target_os = "windows")]`, and instead have other platforms panic with `unimplemented!` when they receive this enum variant.

0xpr03: But unimplemented means it looks like it's available ? And then crashes during run..

Carbonhell: uh i'm gonna wait before fixing the other issues then, personally i think compiling this only under windows made more sense, instead of panic attacks, not sure

murarth:
It would be great for this to be a cross-platform API, but we need to address the fact that different platforms will expect different file formats. The file provided for the example, `tailless-pointer.cur`, is a Windows resource file. X11 can also set the pointer to an image, but it expects (naturally) an X11 cursor file. A developer cannot simply provide one file to winit and expect the same result on every supported platform.

There are two ways I can see to address this:
* The simple solution is to expose this API through platform-specific traits; `WindowExtWindows`, `WindowExtUnix`, etc.
* A more complex, but cross-platform solution would be to implement a separate crate that would build a platform-specific cursor file from an image and the necessary metadata. Winit users could use a build script to generate a cursor file for the native platform when the project is built and provide this file to winit at runtime.

src/window.rs
```diff
@@ -788,6 +788,9 @@ pub enum CursorIcon {
     NwseResize,
     ColResize,
     RowResize,
+
+    /// Represents a custom cursor which uses an image. At the current time, it's only implemented for Windows. The argument has to be a path pointing to a .cur or .ani file.
+    Custom(&'static str),
```
Osspial: Is there any particular reason we aren't using Winit's [`Icon`](https://docs.rs/winit/0.20.0-alpha3/winit/window/struct.Icon.html) type here? Even though the winapi docs for `SetCursor` don't explicitly say so, the function does accept the `HICON` handle provided by `platform_impl::windows::icon::WinIcon`, and it should make this variant easier to port between platforms.

murarth: On X11 (and I imagine Windows is similar), cursor data may contain information that isn't represented by `Icon`. Specifically, it may contain a position (this is used to implement the crosshair, whose pointer position is in the center of the cursor image instead of the top left) and it may contain multiple images that are cycled through in an animation.

Osspial: We could change `Custom`'s signature to `Custom{ hotspot: PhysicalPosition, icon: Icon }` (or some other name for `hotspot` - that's the Windows terminology). WRT animated cursors - honestly, that seems like a niche enough feature that I don't think we need to officially support it, although API-wise we could use `Vec<Icon>`, `framerate: u32` if we really needed to.

ffoglia9:
Thank you all for the useful info, sadly I don't have enough time nor experience in Rust to finish this for now. I'll try to pick it up again in future.

Carbonhell:
Thank you all for the useful info, but for now I lack the time and the proper knowledge of Rust to complete this. I'll try to pick it up again in future.

PR #1201, opened 2019-10-02
Remove derivative dependency

alexheretic:
I wondered if removing derivative could improve compile times, as it has [elsewhere](https://github.com/gfx-rs/gfx/pull/3029). For winit it's not a _huge_ deal for the compile times, but it's still an improvement and one less dependency so maybe it's attractive? Otherwise feel free to close this.

_Tested with linux, rust 1.38, ryzen 2700x_
### Before
```sh
$ hyperfine --max-runs 3 'cargo clean; cargo build --example window'
Benchmark #1: cargo clean; cargo build --example window
  Time (mean ± σ):     23.551 s ±  0.197 s    [User: 174.218 s, System: 9.170 s]
  Range (min … max):   23.353 s … 23.748 s    3 runs
```
### After
```sh
$ hyperfine --max-runs 3 'cargo clean; cargo build --example window' 
Benchmark #1: cargo clean; cargo build --example window
  Time (mean ± σ):     22.894 s ±  0.176 s    [User: 166.748 s, System: 8.805 s]
  Range (min … max):   22.692 s … 23.014 s    3 runs
```

- [x] ~Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior~


src/platform_impl/macos/monitor.rs
```diff
@@ -22,17 +22,44 @@ use core_video_sys::{
     CVDisplayLinkGetNominalOutputVideoRefreshPeriod, CVDisplayLinkRelease,
 };
 
-#[derive(Derivative)]
-#[derivative(Debug, Clone, PartialEq, Hash)]
+#[derive(Clone, Eq)]
```
alexheretic: Technically `Eq` wasn't here before, but it looks like it should have been so I added it.

Osspial:
@alexheretic Ah, thanks for catching that.

PR #1204, opened 2019-10-03
Wayland support for set_cursor_icon

kchibisov:
- [x] Tested on all platforms changed
- [x] Compilation warnings were addressed
- [x] `cargo fmt` has been run on this branch
- [x] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users
- [ ] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior
- [ ] Created or updated an example program if it would help users understand this functionality
- [x] Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented

Cursors' names were copied from X11 implementation, however some of them were truncated, but I've checked these cursors on X11 and Wayland and they were the same. I've also noticed that animated cursors don't work, but I feel like it's a SCTK problem.

P.s.
I've also added cursor grab to feature matrix, since it was recently implemented, but not added to matrix.

PR #1206, opened 2019-10-04
Wayland: Fix panic when calling set_cursor_grab from a different thread than evlp's one.

kchibisov:
CursorManagers\` `grab_pointer` can only be called from a thread on which `pointer_constraints_proxy` event queue is located, so calling it directly from a Window doesn't work well, in case you've sent your window to another thread, so we need to pass cursor grab updates to the event loop and call this function from there.

The code is basically a copy of @andersrein approach for handling a cursor grab thing.

Fixes regression introdced in 5ced36e319d55178dd943de033e4aacdbd0d0327

Some explanation why things were changed in #1204.

However @andersrein approach is not a good one for `set_cursor_icon` and `set_cursor_visible`, since you can face cursor icon update delays when pressing and holding Shift for e.g., the cursor icon will be changed with a bit of a delay according to my testing. I didn't find it at the first place while testing alacritty, since this case was quite odd and I completely forgot about it. So cursor_grab was auto converted to follow set_cursor_icon.

### DIFF
--- a/src/platform_impl/linux/wayland/event_loop.rs
+++ b/src/platform_impl/linux/wayland/event_loop.rs
@@ -132,10 +132,11 @@ impl CursorManager {
     }
 
     pub fn set_cursor_icon(&mut self, cursor: CursorIcon) {
-        if self.cursor_visible && cursor != self.current_cursor {
+        if cursor != self.current_cursor {
             self.current_cursor = cursor;
-
-            self.set_cursor_icon_impl(cursor);
+            if self.cursor_visible {
+                self.set_cursor_icon_impl(cursor);
+            }
         }
     }
 
@@ -191,7 +192,11 @@ impl CursorManager {
         }
     }
 
-    pub fn grab_pointer(&mut self, surface: Option<&WlSurface>) {
+    // This function can only be called from a thread on which `pointer_constraints_proxy` event
+    // queue is located, so calling it directly from a Window doesn't work well, in case
+    // you've sent your window to another thread, so we need to pass cursor grab updates to
+    // the event loop and call this function from there.
+    fn grab_pointer(&mut self, surface: Option<&WlSurface>) {
         for locked_pointer in self.locked_pointers.drain(..) {
             locked_pointer.destroy();
         }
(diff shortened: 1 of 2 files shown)


src/platform_impl/linux/wayland/window.rs
```diff
@@ -441,6 +441,8 @@ impl WindowStore {
             bool,
             bool,
             bool,
+            Option<bool>,
+            &wl_surface::WlSurface,
```
andersrein: Is there any particular reason we don't just pass a reference to the entire internal window instead of many bit and pieces of it?

src/platform_impl/linux/wayland/event_loop.rs
```diff
@@ -132,10 +132,11 @@ impl CursorManager {
     }
 
     pub fn set_cursor_icon(&mut self, cursor: CursorIcon) {
-        if self.cursor_visible && cursor != self.current_cursor {
+        if cursor != self.current_cursor {
             self.current_cursor = cursor;
-
-            self.set_cursor_icon_impl(cursor);
+            if self.cursor_visible {
```
kchibisov: Follow X11 behavior on cursor icon updates when it's in invisible state.

src/platform_impl/linux/wayland/window.rs
kchibisov: They wont be the same and we can't really pass it, since we want to mutate existing one and send something different (`replace`, `take`). `make_wid` doesn't make sense in event loop proxy either. We can try to do it ofc, but I'd like to leave it as it is.

@vberger probably have some explanation on why things are written this way though.

vberger: To be quite honest I don't remember exactly the reason I set it like that originally, most likely I didn't find a better way. And then things grew on top of it without changing the pattern.

There may certainly be a better way to structure this, but I agree this would be for an other PR. In general I suspect there is room for some refactoring / reorganization of winit's wayland backend, and I also suspect this would better be done in link with https://github.com/Smithay/client-toolkit/issues/77.

PR #1207, opened 2019-10-04
Use consistent return types for available_monitors()

### LINKED ISSUE #1111: Inconsistent return types for available_monitors()
The event loop and window both have a method `available_monitors()` both return `AvailableMonitorsIter` but the event loop declares the return type as `impl Iterator<Item = MonitorHandle>` instead of `AvailableMonitorsIter`.

The documentation of `AvailableMonitorsIter` states that it can be acquired with `EventLoop::available_monitors` which is wrong.

Is there a reason for this?

cronokirby:
Fixes #1111 

This just changes the return types to be consistent between `event_loop` and `window`. This doesn't go as far as to introduce something like `MonitorHandle::available_monitors(&EventLoop)`, but that could be amended if desired. I just didn't want to introduce a breaking change so quickly

### DIFF
--- a/src/monitor.rs
+++ b/src/monitor.rs
@@ -1,13 +1,12 @@
 //! Types useful for interacting with a user's monitors.
 //!
 //! If you want to get basic information about a monitor, you can use the [`MonitorHandle`][monitor_id]
-//! type. This is retreived from an [`AvailableMonitorsIter`][monitor_iter], which can be acquired
-//! with:
+//! type. This is retreived from one of the following methods, which return an iterator of
+//! [`MonitorHandle`][monitor_id]:
 //! - [`EventLoop::available_monitors`][loop_get]
 //! - [`Window::available_monitors`][window_get].
 //!
 //! [monitor_id]: ./struct.MonitorHandle.html
-//! [monitor_iter]: ./struct.AvailableMonitorsIter.html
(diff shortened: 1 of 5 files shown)


src/window.rs
```diff
@@ -703,7 +703,7 @@ impl Window {
     ///
     /// **iOS:** Can only be called on the main thread.
     #[inline]
-    pub fn available_monitors(&self) -> AvailableMonitorsIter {
+    pub fn available_monitors(&self) -> impl Iterator<Item = MonitorHandle> {
         let data = self.window.available_monitors();
         AvailableMonitorsIter {
```
aleksijuvani: We should remove the `AvailableMonitorsIter` type since it is no longer required. Then this function can simply be `self.window.available_monitors().into_iter().map(|inner| MonitorHandle { inner })`.

cronokirby: Good idea :)

src/monitor.rs
```diff
@@ -10,42 +10,11 @@
 //! [monitor_iter]: ./struct.AvailableMonitorsIter.html
```
aleksijuvani: Let's remove references to `AvailableMonitorsIter` in this top-level comment as well! Looks good to me otherwise.

Issue #686, opened 2018-10-25
Add WindowExt::get_hinstance() for win32 windows


akatechis:
I'm trying to create a Vulkan Surface for drawing to a winit window using [Surface::from_hwnd()](https://docs.rs/vulkano/0.10.0/vulkano/swapchain/struct.Surface.html#method.from_hwnd) from the vulkano crate.

Is there a better/preferred way to do this?

As a sidenote: I see that [docs.rs](https://docs.rs/winit/0.17.2/winit/os/index.html) only shows the `unix` submodule under `os`, presumably because the docs are built on a unix-like machine?

vberger:
To create vulkan surfaces, there is the [vulkano-win](https://docs.rs/vulkano-win/) crate that bridges vulkano and winit, and it apparently works without needing this addition. (I know nothing about vulkan or vulkano though.)

Regarding docs.rs, the top bar has a drop-down menu for changing the platform, it just displays the linux one by default.

akatechis:
I'll take a look! Thanks for both answers 👍

vberger:
I'm closing this issue. Feel free to reopen if needed.

nifker:
I want to use ash myself and add allow window creation on windows - should I use win32 then or might this still be added?

nifker:
@vberger I want to use ash myself and want to allow window creation on windows - should I use win32 then or might this still be added?

vberger:
Better ask @Osspial, I know pretty much nothing about win32.

Osspial:
@nifker I suppose there wouldn't be much harm to adding this, but it's already possible to do this externally to Winit via [`GetWindowLong(hwnd, GWL_HINSTANCE)`](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getwindowlonga), so I don't think it's strictly necessary for us to expose. I'd accept a PR implementing it, but I don't want to write such a PR myself.

nifker:
@Osspial Ive also seen using this function to get HINSTANCE: https://docs.rs/winapi/0.3.8/winapi/um/libloaderapi/fn.GetModuleHandleW.html

PR #1213, opened 2019-10-05
Implemented a HINSTANCE getter function for Windows

nifker:

Issue: #686

### DIFF
--- a/src/platform_impl/ios/view.rs
+++ b/src/platform_impl/ios/view.rs
@@ -500,7 +500,9 @@ pub unsafe fn create_window(
             let () = msg_send![uiscreen, setCurrentMode: video_mode.video_mode.screen_mode];
             msg_send![window, setScreen:video_mode.monitor().ui_screen()]
         }
-        Some(Fullscreen::Borderless(ref monitor)) => msg_send![window, setScreen:monitor.ui_screen()],
+        Some(Fullscreen::Borderless(ref monitor)) => {
+            msg_send![window, setScreen:monitor.ui_screen()]
+        }
         None => (),
     }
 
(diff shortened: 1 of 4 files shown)


src/platform_impl/windows/window.rs
```diff
@@ -825,7 +830,9 @@ unsafe fn init<T: 'static>(
             return Err(os_error!(io::Error::last_os_error()));
         }
 
-        WindowWrapper(handle)
+        let hinstance = libloaderapi::GetModuleHandleW(std::ptr::null() as *const _);
```
Osspial: Strictly speaking, it's possible to change a window's HINSTANCE via [`SetWindowLong`](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setwindowlonga), so we should probably use `GetWindowLong` to prevent bugs if somebody (future Winit developers or downstream users) decides to do that.

src/platform_impl/ios/view.rs
```diff
@@ -500,7 +500,9 @@ pub unsafe fn create_window(
             let () = msg_send![uiscreen, setCurrentMode: video_mode.video_mode.screen_mode];
             msg_send![window, setScreen:video_mode.monitor().ui_screen()]
         }
-        Some(Fullscreen::Borderless(ref monitor)) => msg_send![window, setScreen:monitor.ui_screen()],
+        Some(Fullscreen::Borderless(ref monitor)) => {
+            msg_send![window, setScreen:monitor.ui_screen()]
+        }
```
Osspial: The `rustfmt` rules changed in the latest stable release, which is why CI is complaining. Update to the newest stable and re-format and this change, as well as the associated CI error, should go away.

src/platform_impl/windows/window.rs
nifker: I implemented it using GetWindowLong now

PR #1217, opened 2019-10-08
Add support for Windows Dark Mode

### LINKED ISSUE #1193: Add support for Windows dark theme
This has been reported to Alacritty in https://github.com/jwilm/alacritty/issues/2829.

davidhewitt:
I had a go at adding simple support for Windows 10's "Dark Mode". The moment the title bar will set to light / dark automatically according to system theme. I did not add an option for users of `winit` to disable this behaviour, as I felt it's likely desirable for most use cases. I can change that if you prefer otherwise!

Please advise on any documentation you'd like me to update.

The "window" example when Dark Mode is enabled:

image

Closes #1193

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -215,6 +215,8 @@ pub enum WindowEvent {
     ///
     /// Applications might wish to react to this to change the theme of the content of the window
     /// when the system changes the window theme.
+    ///
+    /// At the moment this is only supported on Windows.
     ThemeChanged(Theme),
 }
 
--- a/src/event.rs
+++ b/src/event.rs
@@ -210,6 +210,13 @@ pub enum WindowEvent {
     ///
     /// For more information about DPI in general, see the [`dpi`](crate::dpi) module.
     HiDpiFactorChanged(f64),
+
+    /// The window theme has changed into (or out of) Dark Mode.
+    ///
+    /// The parameter is true if the window is now in a dark theme, and false if it is not.
+    /// Applications might wish to react to this to change the theme of the content of the window
+    /// when the system changes the window theme.
+    DarkModeChanged(bool),
 }
 
 /// Identifier of an input device.
--- a/src/event.rs
+++ b/src/event.rs
@@ -10,7 +10,7 @@ use std::path::PathBuf;
 use crate::{
     dpi::{LogicalPosition, LogicalSize},
     platform_impl,
-    window::WindowId,
+    window::{Theme, WindowId},
 };
 
 /// Describes a generic event.
@@ -211,12 +211,11 @@ pub enum WindowEvent {
     /// For more information about DPI in general, see the [`dpi`](crate::dpi) module.
     HiDpiFactorChanged(f64),
 
-    /// The window theme has changed into (or out of) Dark Mode.
+    /// The system window theme has changed.
     ///
-    /// The parameter is true if the window is now in a dark theme, and false if it is not.
     /// Applications might wish to react to this to change the theme of the content of the window
     /// when the system changes the window theme.
-    DarkModeChanged(bool),
+    ThemeChanged(Theme),
 }
 
 /// Identifier of an input device.
--- a/src/platform_impl/windows/dark_mode.rs
+++ b/src/platform_impl/windows/dark_mode.rs
@@ -73,7 +73,7 @@ lazy_static! {
         // We won't try to do anything for windows versions < 17763
         // (Windows 10 October 2018 update)
         match *WIN10_BUILD_VERSION {
-            Some(v) => v > 18362,
+            Some(v) => v >= 17763,
             None => false
         }
     };
--- a/src/platform_impl/windows/dark_mode.rs
+++ b/src/platform_impl/windows/dark_mode.rs
@@ -7,7 +7,7 @@ use winapi::{
     shared::{
         basetsd::SIZE_T,
         minwindef::{BOOL, DWORD, UINT, ULONG, WORD},
-        ntdef::{LPSTR, PVOID, WCHAR},
+        ntdef::{LPSTR, NTSTATUS, NT_SUCCESS, PVOID, WCHAR},
         windef::HWND,
     },
     um::{libloaderapi, uxtheme, winuser},
@@ -29,7 +29,7 @@ lazy_static! {
             szCSDVersion: [WCHAR; 128],
         }
 
-        type RtlGetVersion = unsafe extern "system" fn (*mut OSVERSIONINFOW);
+        type RtlGetVersion = unsafe extern "system" fn (*mut OSVERSIONINFOW) -> NTSTATUS;
         let handle = get_function!("ntdll.dll", RtlGetVersion);
 
         if let Some(rtl_get_version) = handle {
@@ -43,7 +43,8 @@ lazy_static! {
                     szCSDVersion: [0; 128],
                 };
 
-                (rtl_get_version)(&mut vi as _);
+                let status = (rtl_get_version)(&mut vi as _);
+                assert!(NT_SUCCESS(status));
 
                 if vi.dwMajorVersion == 10 && vi.dwMinorVersion == 0 {
                     Some(vi.dwBuildNumber)
@@ -109,11 +110,10 @@ fn set_dark_mode_for_window(hwnd: HWND, is_dark_mode: bool) {
 
     #[allow(non_snake_case)]
     #[repr(C)]
-    struct WINDOWCOMPOSITIONATTRIBDATA
-    {
+    struct WINDOWCOMPOSITIONATTRIBDATA {
         Attrib: WINDOWCOMPOSITIONATTRIB,
         pvData: PVOID,
-        cbData: SIZE_T
+        cbData: SIZE_T,
     }
 
     lazy_static! {
@@ -124,20 +124,17 @@ fn set_dark_mode_for_window(hwnd: HWND, is_dark_mode: bool) {
     if let Some(set_window_composition_attribute) = *SET_WINDOW_COMPOSITION_ATTRIBUTE {
         unsafe {
             // SetWindowCompositionAttribute needs a bigbool (i32), not bool.
-            let mut is_dark_mode_bigbool = is_dark_mode as BOOL;
+            let mut is_dark_mode_bigbool: BOOL = is_dark_mode.into();
 
             let mut data = WINDOWCOMPOSITIONATTRIBDATA {
                 Attrib: WCA_USEDARKMODECOLORS,
                 pvData: &mut is_dark_mode_bigbool as *mut _ as _,
-                cbData: std::mem::size_of_val(&is_dark_mode_bigbool) as _
+                cbData: std::mem::size_of_val(&is_dark_mode_bigbool) as _,
             };
 
             assert_eq!(
                 1,
-                set_window_composition_attribute(
-                    hwnd,
-                    &mut data as *mut _
-                )
+                set_window_composition_attribute(hwnd, &mut data as *mut _)
             );
         }
     }
@@ -152,15 +149,12 @@ fn should_apps_use_dark_mode() -> bool {
     lazy_static! {
         static ref SHOULD_APPS_USE_DARK_MODE: Option<ShouldAppsUseDarkMode> = {
             unsafe {
-
                 const UXTHEME_SHOULDAPPSUSEDARKMODE_ORDINAL: WORD = 132;
 
-                let module = libloaderapi::LoadLibraryA(
-                    "uxtheme.dll\0".as_ptr() as _
-                );
+                let module = libloaderapi::LoadLibraryA("uxtheme.dll\0".as_ptr() as _);
 
                 if module.is_null() {
-                    return None
+                    return None;
                 }
 
                 let handle = libloaderapi::GetProcAddress(
--- a/src/platform_impl/windows/dark_mode.rs
+++ b/src/platform_impl/windows/dark_mode.rs
@@ -43,19 +43,28 @@ lazy_static! {
             }
         }
     };
+    static ref DARK_THEME_NAME: Vec<u16> = widestring("DarkMode_Explorer");
+    static ref LIGHT_THEME_NAME: Vec<u16> = widestring("");
 }
 
-pub fn try_dark_mode(hwnd: HWND) {
-    let theme_name: Vec<_> = widestring("DarkMode_Explorer");
-
+/// Attempt to set dark mode on a window, if necessary.
+/// Returns true if dark mode was set, false if not.
+pub fn try_dark_mode(hwnd: HWND) -> bool {
     // According to Windows Terminal source, should be BOOL (32-bit int)
     // to be appropriately sized as a parameter for DwmSetWindowAttribute
-    let is_dark_mode: BOOL = should_use_dark_mode() as _;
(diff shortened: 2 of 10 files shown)


davidhewitt:
(I rebased onto master to fix the merge conflict)

Osspial:
For the most part, this seems good. Could you add a function to `WindowExtWindows` to allow users to retrieve whether dark mode is active or not? We should let users change their theme based on the OS option if they can.

davidhewitt:
Thanks for the review! I've added a field to `WindowState` to track the current state of dark mode.

Regarding the event, I was also wondering about such. I know MacOS also has a dark mode though know nothing about its integration in APIs. So the message might be relevant on at least two platforms.

chrisduerr:
> I know MacOS also has a dark mode though know nothing about its integration in APIs

Dark mode on macOS is enabled by setting a property in the info.plist file. I'm not sure how you query for it as an application, but to support it there's no support necessary from winit (Alacritty already sets this property).

davidhewitt:
A quick google for macOS suggests we might be able to recieve a system broadcast for it on that platform too: https://stackoverflow.com/questions/39048894/how-to-detect-switch-between-macos-default-dark-mode-using-swift-3

So it might be appropriate to add a `WindowEvent` called `SystemThemeChanged` or `DarkModeChanged` - which might never get sent on some platforms (for now) but could be sent on (at least) Windows and macOS.

I don't have a mac so can't do the implementation for that platform. If you agree we should add a new `WindowEvent` I can do the Windows implementation as part of this PR and then open an issue for someone else to follow up with mac support.

Osspial:
@davidhewitt I'm fine with taking that approach.

davidhewitt:
Great - have now added an event `DarkModeChanged` and confirmed it fires during the `window` example.

davidhewitt:
Just a heads up; I recently updated my system to the latest Windows Insiders and it appears that this definition: `const DWMWA_USE_IMMERSIVE_DARK_MODE: DWORD = 19;` now needs to be `20` for the `window` example to run successfully on my system. I've no idea if this is an intended change or a bug or what.

The windows terminal source doesn't appear to have changed from `19` to `20`, but that still works fine, which is puzzling. Might be worth holding off on merging this until I figure out what's going on.

src/event.rs
```diff
@@ -204,6 +204,13 @@ pub enum WindowEvent {
     ///
     /// For more information about DPI in general, see the [`dpi`](../dpi/index.html) module.
     HiDpiFactorChanged(f64),
+
+    /// The window theme has changed into (or out of) Dark Mode.
+    ///
+    /// The parameter is true if the window is now in a dark theme, and false if it is not.
+    /// Applications might wish to react to this to change the theme of the content of the window
+    /// when the system changes the window theme.
+    DarkModeChanged(bool),
```
Osspial: I don't think we should have this be a `bool`, since I can see a there isn't a clear semantic mapping between `true`/`false` and `Dark`/`Light`. Could we have this be `ThemeChanged(Theme)` where `Theme` is defined as follows?

```rust
enum Theme {
    Light,
    Dark,
}
```

davidhewitt:
I've updated the to use a `ThemeChanged` event, and created a workaround my problem. The best solution I could come up with is checking the Windows 10 build version and subsequently choosing the appropriate value for `DWMA_USE_IMMERSIVE_DARK_MODE`.

goddessfreya:
@Osspial This is, afaik, good for merge?

Osspial:
@davidhewitt Sorry it took so long to get back to you on this. I have no idea what's going on with dark mode on your system, since `8ba2055` completely broke this for me. There are two primary issues:

- `WIN10_BUILD_VERSION` is defined as `4026549603` on my system, far outside the range defined in the source file.
- Setting `DWMWA_USE_IMMERSIVE_DARK_MODE` to `20` causes the `DwmSetWindowAttribute` to fail. `19` works fine.

I'm inclined to remove the windows version check and publish this with `DWMWA_USE_IMMERSIVE_DARK_MODE = 19`.

davidhewitt:
> I'm inclined to remove the windows version check and publish this with DWMWA_USE_IMMERSIVE_DARK_MODE = 19.

I would happily undo the version check! Strange about my system then. Are you on a stable windows 10 release or an insiders build?

davidhewitt:
Rebased on master.

@Osspial I've changed the method used for querying the Windows 10 build version. Does this new method now return a build version within the expected range on your machine?

(I'm currently on Windows 10 insiders build 19018, and `WIN10_BUILD_VERSION` is coming out as `19018`)

If this new method doesn't work on your box, then I'll remove the version checking and hard-code `19` again. Though it might be great if someone else on Windows 10 insiders can report their build number and also whether `19` or `20` is the correct definition of `DWMWA_USE_IMMERSIVE_DARK_MODE` for them.

chelmich:
On my machine `WIN10_BUILD_VERSION` is reported as `18362`. This matches what is displayed by `winver`. I can confirm that the dark theme works as intended, including changing the theme while a window is open.

daxpedda:
I had access to a machine with windows build 17763 on it, tried this out and it works as intended, `WIN10_BUILD_VERSION` was also reported correctly.

A tiny thing that I noticed: it doesn't change theme when the window isn't focused, but it will switch to the correct one as soon as it's focused again.

davidhewitt:
Thanks guys. Anyone else also on Windows 10 insiders builds greater than  `18363`? Looks like if there was an API change (if there is indeed one and it's not my system) it ocurred on a build later than that.

> A tiny thing that I noticed: it doesn't change theme when the window isn't focused, but it will switch to the correct one as soon as it's focused again.

Interesting. Did you happen to notice at what time the `ThemeChanged` event is being fired in the `window` example? Is it when it switches into focus, or when you change the dark theme setting? (On my system the ThemeChanged event fires when I change the system setting, regardless of focus.)

daxpedda:
> Did you happen to notice at what time the `ThemeChanged` event is being fired in the `window` example?

It fires as soon as I change the system setting, even though it will only actually switch theme when I focus the window again.

Same behaviour as on your system.

davidhewitt:
> It fires as soon as I change the system setting, even though it will only actually switch theme when I focus the window again. 

Cool. So we can probably make it switch theme sooner if we force Windows to redraw the title bar.

Past experiments I've had into trying to do that have never been very successful... anyone know a nice API to force the title bar to redraw? 😄

src/platform_impl/windows/event_loop.rs
```diff
@@ -1895,6 +1896,27 @@ unsafe extern "system" fn public_window_callback<T>(
...
+            let is_dark_mode = try_dark_mode(window);
+            let mut window_state = subclass_input.window_state.lock();
+            let changed = window_state.is_dark_mode != is_dark_mode;
+
+            if changed {
+                use crate::window::Theme::*;
+                let theme = if is_dark_mode { Dark } else { Light };
+
+                window_state.is_dark_mode = is_dark_mode;
+                subclass_input.send_event(Event::WindowEvent {
```
daxpedda:
```suggestion
                mem::drop(window_state);
                subclass_input.send_event(Event::WindowEvent {
```

Before sending an event we should drop `window_state`, trying to do anything on `ThemeChanged` like changing window size will otherwise crash the program.

davidhewitt: Good catch - committed!

daxpedda:
> Past experiments I've had into trying to do that have never been very successful... anyone know a nice API to force the title bar to redraw? 😄

I was playing around with this for a bit and as far as I could find, using `SetWindowPos` with `SWP_FRAMECHANGED` should fix this, but it doesn't. The only thing I could find is resizing the window to a new size, any size does it, just not the original one.

So the only workaround I could come up with, is if the window isn't focused, we want to resize the window by one pixel and than back again.

daxpedda:
> Given that this focus issue only applies to an out-of-date Windows 10 build

I didn't know this issue is not present in newer builds? I will try on a newer build now and report back.

davidhewitt:
> I didn't know this issue is not present in newer builds? I will try on a newer build now and report back.

I'm now doubting myself. Pretty sure on my system (most recent Windows Insiders build) the frame changes whether focussed or not. I'll check later.

daxpedda:
You were right, this only affects older versions of windows. I would consider this a bug on windows, not our problem.

davidhewitt:
I asked about the support for insiders builds at https://github.com/ysc3839/win32-darkmode/issues/2

From that I learned of a different solution which works for builds `18362` (April 2019) and up, including my latest windows insiders build.

So I've changed the implementation here to that new solution. I've also moved the `lazy_static!` calls around to put the statics closer to where they're used, instead of global.

I've kept the windows 10 build version check in, so that this code will no-op rather than crash on older windows 10 builds.

This is hopefully now good to go!

src/platform_impl/windows/dark_mode.rs
```diff
@@ -0,0 +1,257 @@
...
+                }
+            }
+        }
+    };
+
+    static ref DARK_MODE_SUPPORTED: bool = {
+        // We won't try to do anything for windows versions < 17763
+        // (Windows 10 October 2018 update)
+        match *WIN10_BUILD_VERSION {
+            Some(v) => v > 18362,
```
daxpedda:
```suggestion
            Some(v) => >= 17763,
```

I think is what you meant?

daxpedda:
I tried on `17763` and `18362`, both work with the fix I suggested.

davidhewitt:
> I tried on `17763` and `18362`, both work with the fix I suggested.

Thanks! The example I was using only used this method from `18362`, so I was being conservative with putting `18362` in the check. But I forgot to update the comment. Fortunately it turns out `17763` also works 😄 

Out of curiosity, how are you testing so many Win10 versions so quickly? VMs?

daxpedda:
> Out of curiosity, how are you testing so many Win10 versions so quickly? VMs?

Hahaha, that would be epic! 😄
My main windows machine is running LTSC, which is currently stuck at `17763`, but my tablet is running `18362`, so we just happen to be fortunate 😁.

src/platform_impl/windows/dark_mode.rs
```diff
@@ -0,0 +1,257 @@
...
+
+fn should_apps_use_dark_mode() -> bool {
+    type ShouldAppsUseDarkMode = unsafe extern "system" fn() -> bool;
+    lazy_static! {
+        static ref SHOULD_APPS_USE_DARK_MODE: ShouldAppsUseDarkMode = {
+            unsafe {
+
+                const UXTHEME_SHOULDAPPSUSEDARKMODE_ORDINAL: WORD = 132;
+
+                let module = libloaderapi::LoadLibraryExW(
```
daxpedda: We are using `LoadLibraryA` in the rest of the library, not sure if this is important to anybody to keep this uniform.

davidhewitt:
Thanks; I switched to use `get_function!` and `LoadLibraryA`.

src/platform_impl/windows/dark_mode.rs
```diff
@@ -0,0 +1,222 @@
...
+        struct OSVERSIONINFOW {
+            dwOSVersionInfoSize: ULONG,
+            dwMajorVersion: ULONG,
+            dwMinorVersion: ULONG,
+            dwBuildNumber: ULONG,
+            dwPlatformId: ULONG,
+            szCSDVersion: [WCHAR; 128],
+        }
+
+        type RtlGetVersion = unsafe extern "system" fn (*mut OSVERSIONINFOW);
```
daxpedda: `RtlGetVersion` has a return type, it's supposed to be `winapi::shared::ntdef::NTSTATUS`.

src/platform_impl/windows/dark_mode.rs
```diff
@@ -0,0 +1,222 @@
...
+                let mut vi = OSVERSIONINFOW {
+                    dwOSVersionInfoSize: 0,
+                    dwMajorVersion: 0,
+                    dwMinorVersion: 0,
+                    dwBuildNumber: 0,
+                    dwPlatformId: 0,
+                    szCSDVersion: [0; 128],
+                };
+
+                (rtl_get_version)(&mut vi as _);
```
daxpedda: We should use `winapi::shared::ntdef::NT_SUCCESS` to check if the return value is a success.

src/platform_impl/windows/dark_mode.rs
```diff
@@ -0,0 +1,222 @@
...
+
+    lazy_static! {
+        static ref SET_WINDOW_COMPOSITION_ATTRIBUTE: Option<SetWindowCompositionAttribute> =
+            get_function!("user32.dll", SetWindowCompositionAttribute);
+    }
+
+    if let Some(set_window_composition_attribute) = *SET_WINDOW_COMPOSITION_ATTRIBUTE {
+        unsafe {
+            // SetWindowCompositionAttribute needs a bigbool (i32), not bool.
+            let mut is_dark_mode_bigbool = is_dark_mode as BOOL;
```
daxpedda:
```suggestion
            let mut is_dark_mode_bigbool = is_dark_mode.into();
```
Simple widening, doesn't need `as`.

src/platform_impl/windows/dark_mode.rs
```diff
@@ -0,0 +1,222 @@
...
+
+            let mut data = WINDOWCOMPOSITIONATTRIBDATA {
+                Attrib: WCA_USEDARKMODECOLORS,
+                pvData: &mut is_dark_mode_bigbool as *mut _ as _,
+                cbData: std::mem::size_of_val(&is_dark_mode_bigbool) as _
```
daxpedda:
```suggestion
                cbData: std::mem::size_of_val(&is_dark_mode_bigbool).into()
```
The underlying type is `usize`.

davidhewitt: Unfortunately windows wants `u32` for the cbData value, so we need to do a narrowing cast here. Size of a bool is always going to fit safely in a `u32`, so this narrowing `as` isn't dangerous.

davidhewitt:
Thanks again, have addressed all those suggestions.

src/event.rs
```diff
@@ -201,6 +201,12 @@ pub enum WindowEvent {
     ///
     /// For more information about DPI in general, see the [`dpi`](crate::dpi) module.
     HiDpiFactorChanged(f64),
+
+    /// The system window theme has changed.
+    ///
+    /// Applications might wish to react to this to change the theme of the content of the window
+    /// when the system changes the window theme.
+    ThemeChanged(Theme),
```
goddessfreya: We should probably mention that it is currently only implemented for windows.

davidhewitt: Thanks, have done so!

VideoPlayerCode:
Looking really good now. Great work @davidhewitt

src/platform_impl/windows/dark_mode.rs
Osspial: I have to thoroughly disagree with this comment, actually. Casting via `as` makes it clear that it's *just* a vanilla cast with nothing fancy going on, which is much less clear with an `into()` call.

Osspial:
Sorry about the delay reviewing this. Once my nit is addressed this should be good to go. Thank you for researching and implementing this!

davidhewitt:
Nit fixed, and rebased - expecting the CI to be green and fingers crossed this is good to merge!

PR #1220, opened 2019-10-11
Add web targets to travis

ryanisaacg:


Osspial:
Could you add lines testing this on nightly?

ryanisaacg:
Any idea why appveyor is failing? This shouldn't be affecting it afaik

Osspial:
@ryanisaacg Looks like it's failing to configure `gcc` properly on nightly. It's probably a configuration error somewhere on Appveyor or `rustup-init`'s side, though if it persists for a few days we should probably take a closer look at it.

ryanisaacg:
@Osspial Are we good to merge this even with appveyor failing?

Osspial:
@ryanisaacg yep

PR #1227, opened 2019-10-17
Fix so the compiler can infer msg_send! return types

SSheldon:
Currently, due to a quirk in Rust's type inference interacting with the structure of the `msg_send!` macro, a `()` return type will be inferred when the compiler cannot otherwise determine the return type. This behavior is expected to change, and in the future could resolve to a `!` return type, which results in undefined behavior.

Linting has previously been added for this in rust-lang/rust#39216, but it did not catch these cases due to SSheldon/rust-objc#62. An upcoming version of objc will be fixed to stop hiding these errors, at which point they will become compile errors.

This change fixes these errors and allows winit to compile with the fixed version of objc. It also bumps cocoa to ensure users of winit are on a version that has addressed those issues.

  - I ran `cargo test` on macos, close enough?
  - No compilation warnings added by this change!
  - No, it's such a li'l change and I don't have rustfmt installed
  - Probably not valuable to users
  - No user-facing changes!
  - No functionality change!
  - No new features!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ version = "0.2"
 objc = "0.2.3"
 
 [target.'cfg(target_os = "macos")'.dependencies]
-cocoa = "0.18.4"
+cocoa = "0.19.1"
 core-foundation = "0.6"
 core-graphics = "0.17.3"
 dispatch = "0.1.4"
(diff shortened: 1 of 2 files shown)


SSheldon:
Awesome, thank you!

@Osspial, do y'all have plans for publishing another alpha of 0.20? Will that be soon? I'm wondering if I should wait for the next alpha with these fixes before I publish objc 0.2.7. (At this point cocoa has published fixes for the 0.18 and 0.19 releases, and every other dependency I've checked is looking good, so I think winit 0.20 is the last place that would be impacted by the new objc version.)

Osspial:
@SSheldon I just uploaded Alpha 4 about ten minutes ago!

Issue #1082, opened 2019-07-31
Tracking issue: Rethinking RedrawRequested 


goddessfreya:
This issue tracks the implementation of the RedrawRequested API changes that were discussed in #1041. PRs implementing this change should be made against the `redraw-requested-2.0` branch, and we'll merge that branch onto `master` once it's implemented on all platforms.

- [x] Public API changed
- [x] Implemented on all platforms
  - [x] Windows - #1050
  - [x] macOS - #1235
  - [x] iOS - #1299
  - [x] Linux X11 - #1062 
  - [x] Linux Wayland - #1237 
  - [x] Web - #1301
- [ ] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users
- [ ] Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented.

mtak-:
#1133 gets iOS 90% of the way there. There will need to be a separate (trivial) pr to actually send out the events.

hecrj:
I can try to get this working on macOS. Is there anything I should be specially aware of?

Osspial:
@mtak- the `redraw-requested-2.0` branch now contains #1133. Would it be possible to open a PR getting the iOS implementation the rest of the way there?

Osspial:
@aleksijuvani You've got some experience with iOS programming. Could you see about implementing this on iOS?

@ryanisaacg Could you implement this on the web?

aleksijuvani:
I'm no longer working with Rust these days, so I don't have a strong incentive to work on this. I can help with reviewing the pull request when it's done, but I won't be implementing this myself.

Osspial:
Wow, that's... all the platforms done. I'll get this ready to merge within the next day or so.

PR #1237, opened 2019-10-22
Implemented revamped `RedrawRequested` for linux wayland

heghe:
Implement the RedrawRequested for linux wayland. Partial fix for #1082.

- [x] Tested on all platforms changed (I tested only with the examples, but they seems to work)

### DIFF
--- a/src/platform_impl/linux/wayland/event_loop.rs
+++ b/src/platform_impl/linux/wayland/event_loop.rs
@@ -500,24 +500,34 @@ impl<T: 'static> EventLoop<T> {
                     );
                 }
             }
-            // do a second run of post-dispatch-triggers, to handle user-generated "request-redraw"
-            // in response of resize & friends
-            self.post_dispatch_triggers();
+            // send Events cleared
             {
-                let mut guard = sink.lock().unwrap();
-                guard.empty_with(|evt| {
+                sticky_exit_callback(
+                    crate::event::Event::MainEventsCleared,
+                    &self.window_target,
+                    &mut control_flow,
+                    &mut callback,
+                );
+            }
+
+            // handle request-redraw
+            {
+                self.redraw_triggers(|wid, window_target| {
                     sticky_exit_callback(
-                        evt,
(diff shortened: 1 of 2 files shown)


Osspial:
Hi, and congratulations on your first PR!

I just rebased `redraw-requested-2.0` branch against master. It looks like that introduced a few merge conflicts, but they don't seem like they'll be too much effort to resolve.

heghe:
I rebased against the new `redraw-requested-2.0` and fix the conflicts.

src/platform_impl/linux/wayland/event_loop.rs
```diff
@@ -695,10 +724,6 @@ impl<T> EventLoop<T> {
                         wid,
                     );
                 }
-                if refresh {
-                    unimplemented!()
-                    //sink.send_window_event(crate::event::WindowEvent::RedrawRequested, wid);
-                }
```
vberger: The `refresh` needs to be handled here, because the `for_each` method will clear the flag after being called. This means that if the `refresh` flag had been set *before* `MainEventsCleared`, this will be lost.

I think the best way to handle this would be to remove the handling of `refresh` from this `for_each` method, and instead add to `WindowStore` a similar method specifically for handling the refreshes.

heghe:
I've pushed a new version with @vberger proposal. Everything seems ok. Tested with the examples and with glutin examples on sway.

Osspial:
Looks like the CI formatting pass was failing because of formatting issues in the Windows implementation. I've fixed that on the main branch.

heghe:
Yeah, I saw that. Didn't do the fixes myself since ware out of scope for this PR. Is there any way to retrigger the CI besides a new push? or it's fine like this.

Osspial:
I think you have to rebase or merge against our `redraw-requested-2.0`. Github doesn't automatically rebase PRs when the upstream branch changes.

heghe:
Understood. I'll let that for later, in case I need to do more changes
after review.

PR #1245, opened 2019-10-27
X11: Fix events not being reported when using `run_return`

### LINKED ISSUE #1243: No window or device events after returning ControlFlow::Exit in run_return
I'm working on Linux/X11, and I'm using a more traditional poll-style event loop, using `run_return`. After polling for events, the program does other logic such as drawing. In the code below, `run_return` should return control if there are no more events, or if a quit event is received. As I understand it, `ControlFlow::Exit` is used to return from the function, similar to a `break` statement.

```
loop {
	event_loop.run_return(|event, _, control_flow| {
		match event {
			Event::WindowEvent { event, window_id: _ } => {
				match event {
					WindowEvent::CloseRequested => {
						should_quit = true;
					},
					_ => {},
				}
			},
			Event::EventsCleared => {
			},
			_ => {},
		}
	});

	if should_quit {
		return;
	}

	// Draw and logic goes here
}
```

However, with this setup, my window never seems to receive any other events. If I print out the received events, all I get is `NewEvents(Init)`, `EventsCleared` and `LoopDestroyed`, each time `run_return` is called. There are never any other events received. If I do nothing in response to `EventsCleared` instead, then I do receive other events, but then `run_return` continues until I quit the window, so nothing ever gets drawn on it.

If I'm not receiving window or device events in subsequent polls after exiting the loop the first time, then either that seems to be a bug or I'm doing something wrong here.

murarth:
Fixes #1243

### DIFF
--- a/examples/timer.rs
+++ b/examples/timer.rs
@@ -17,7 +17,10 @@ fn main() {
     let timer_length = Duration::new(1, 0);
 
     event_loop.run(move |event, _, control_flow| {
-        println!("{:?}", event);
+        if let Event::WindowEvent { event, .. } = &event {
+            // Print only Window events to reduce noise
+            println!("{:?}", event);
+        }
 
         match event {
             Event::NewEvents(StartCause::Init) => {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 - On macOS, fix application termination on `ControlFlow::Exit`
+- On X11, fix events not being reported when using `run_return`.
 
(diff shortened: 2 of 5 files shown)


examples/timer.rs
```diff
@@ -17,7 +17,10 @@ fn main() {
     let timer_length = Duration::new(1, 0);
 
     event_loop.run(move |event, _, control_flow| {
-        println!("{:?}", event);
+        if let Event::WindowEvent { event, .. } = &event {
```
goddessfreya: Personally, I like getting all the events.

PR #1248, opened 2019-10-28
X11: Fix window creation hangs when another application is fullscreen

### LINKED ISSUE #1210: Window creation hangs on X11 when another application is fullscreen
Most likely happening [here](https://github.com/rust-windowing/winit/blob/master/src/platform_impl/linux/x11/window.rs#L883); it seems like visibility isn't guaranteed to occur at any particular time in this case. Execution resumes once fullscreen is released.

murarth:
Fixes #1210

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 - On macOS, fix application termination on `ControlFlow::Exit`
+- On X11, fix window creation hanging when another window is fullscreen.
 
 # 0.20.0 Alpha 4 (2019-10-18)
 
--- a/examples/window_debug.rs
+++ b/examples/window_debug.rs
@@ -1,0 +1,121 @@
+// This example is used by developers to test various window functions.
+
+use winit::{
+    dpi::{LogicalSize, PhysicalSize},
+    event::{DeviceEvent, ElementState, Event, KeyboardInput, VirtualKeyCode, WindowEvent},
+    event_loop::{ControlFlow, EventLoop},
+    window::{Fullscreen, WindowBuilder},
+};
+
+fn main() {
+    let event_loop = EventLoop::new();
+
+    let window = WindowBuilder::new()
+        .with_title("A fantastic window!")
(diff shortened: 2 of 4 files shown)


murarth:
@goddessfreya I'm not sure I'm understanding the issue you're describing.

> 5- Notice that the window is still visible... which is unexpected, since it shouldn't be. In fact, it doesn't matter how many times we press v, for it is still going to be visible.

Are you sure this isn't just the window_debug window being drawn into the fullscreen window because winit doesn't draw anything into the windows?

> delay all calls to XMapRaised until after the window is made visible, or recall XMapRaised in `visibility_notify`.

`XMapRaised` is what causes a window to become visible. `VisibilityNotify` shouldn't be received at all until after `XMapRaised` is called. And it should have no effect after it is received.

goddessfreya:
> Are you sure this isn't just the window_debug window being drawn into the fullscreen window because winit doesn't draw anything into the windows?

No, this is not the case, because I can see two title bars, one for the window_debug window, and one for the window that used to be fullscreen. If I press v, the `window_debug`'s title bar disappears.

> `XMapRaised` is what causes a window to become visible. `VisibilityNotify` shouldn't be received at all until after `XMapRaised` is called. And it should have no effect after it is received.

My bad, was slightly distracted. I mean (re)calling `XUnmapWindow` which I believe is the function that makes them invisible.

If I remember correctly, the windows stay unmapped until the fullscreen window unfullscreens, so `XUnmapWindow` does nothing, see:

> If the specified window is already unmapped, XUnmapWindow() has no effect.
-- https://tronche.com/gui/x/xlib/window/XUnmapWindow.html

murarth:
Okay, I think I get what you're saying now. winit's window gets unmapped if another window is fullscreen, which means the `XUnmapWindow` is ignored and it doesn't stay unmapped.

I've added another `XUnmapWindow` call in `visibility_notify` when visibility is `No`. I think this should fix the issue. (As usual, this is some WM-specific stuff and my WM isn't experiencing the issue.)

PR #1252, opened 2019-10-30
X11: Fix misreporting DPI factor at startup

### LINKED ISSUE #998: EventLoop 2.0 misreporting DPI Factor at startup on X11
While looking at some remaining bugs in the EventLoop 2 version of Alacritty, I've noticed that the DPI Factor is always `1` at startup, regardless of which monitor the window is spawned on.

If the monitor is moved between monitors, the DPI is always updated correctly. So starting on a monitor with a 2.0 DPI and moving it to a different monitor and back will 'fix' the incorrectly reported DPI.

murarth:
Fixes #998

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 - On macOS, fix application termination on `ControlFlow::Exit`
+- On X11, fix misreporting DPI factor at startup.
 
 # 0.20.0 Alpha 4 (2019-10-18)
 
--- a/src/platform_impl/linux/x11/event_processor.rs
+++ b/src/platform_impl/linux/x11/event_processor.rs
@@ -377,14 +377,8 @@ impl<T: 'static> EventProcessor<T> {
                         let (width, height) = shared_state_lock
                             .dpi_adjusted
                             .unwrap_or_else(|| (xev.width as f64, xev.height as f64));
-                        let last_hidpi_factor =
-                            shared_state_lock.guessed_dpi.take().unwrap_or_else(|| {
-                                shared_state_lock
-                                    .last_monitor
-                                    .as_ref()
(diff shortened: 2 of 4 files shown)


chrisduerr:
Can confirm that this fixes the issue for me.

chrisduerr:
> While this PR didn't change master's behavior (or at least as far as I can tell), I can't confidently say it works.

I can. I'm not sure how your xrandr setup affects things though, since for me my monitors just have different DPI by default.

When starting on my primary monitor with 1.16666666666666667 DPI, I always get 1.0 reported from winit at startup. This PR fixes that.

murarth:
@goddessfreya There are some necessary steps to reproducing the issue that may be missing from your test case:

* Build a window with `.with_visible(false)`
* Call `window.set_visible(true)` immediately after building (before entering `EventLoop::run`)
* Call `window.current_monitor()` (again, before entering the event loop)

On master, in this scenario, `window.current_monitor()` will return the monitor at `0, 0`, while, internally, winit will guess the window's DPI based on the monitor that currently contains the mouse pointer. If these are two different monitors with two different DPI factors, then the application will have received the wrong DPI factor (through the `current_monitor()` handle). Additionally, winit will not correct this by issuing a `HiDpiFactorChanged` event, if its internal guess had been accurate.

This PR fixes that inconsistency by assigning a window's `current_monitor` to the monitor that is guessed based on mouse pointer position. If this guess proves accurate, the application will have received the correct DPI factor. If the guess is not accurate (i.e. winit receives a `ConfigureNotify` event placing the window on a monitor other than the one guessed), winit will issue a `HiDpiFactorChanged` event to correct it, ultimately resulting in the application receiving the correct DPI factor.

I hope this makes clear exactly what this PR is aiming to fix.

chutz:
I can confirm that this fixes https://github.com/jwilm/alacritty/issues/2938 for me

PR #1256, opened 2019-11-01
Extended LogicalSize type conversions

nifker:

### DIFF
--- a/src/dpi.rs
+++ b/src/dpi.rs
@@ -241,79 +241,88 @@ impl LogicalSize {
 }
 
 /// Represents all types into/from which LogicalSize can be converted
-pub trait TypeConversion{
-	/// If necessary(if the output is an integer) it will round the result
+pub trait TypeConversion {
+    /// If necessary(if the output is an integer) it will round the result
     fn get(v: f64) -> Self;
 }
 
-impl TypeConversion for u32{
-    fn get(v: f64) -> Self{
+impl TypeConversion for u32 {
+    fn get(v: f64) -> Self {
         v.round() as u32
     }
 }
-impl TypeConversion for i32{
-    fn get(v: f64) -> Self{
+impl TypeConversion for i32 {
+    fn get(v: f64) -> Self {
         v.round() as i32
     }
 }
-impl TypeConversion for i64{
-    fn get(v: f64) -> Self{
+impl TypeConversion for i64 {
+    fn get(v: f64) -> Self {
         v.round() as i64
     }
 }
-impl TypeConversion for u64{
-    fn get(v: f64) -> Self{
+impl TypeConversion for u64 {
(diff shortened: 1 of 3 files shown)


src/dpi.rs
```diff
@@ -240,32 +240,89 @@ impl LogicalSize {
     }
 }
 
-impl From<(f64, f64)> for LogicalSize {
+/// Represents all types into/from which LogicalSize can be converted
```
goddessfreya: Why precisely is `FromPrimitive` and `ToPrimitive` not satisfactory?

nifker: Because I would either need to 1. downcast at run-time or like I did, add a method which provides an equivalent getter function for all typoes even if some of them need to be handled differently(which is then done in the getter).
Or do you have a better idea to set this up?

goddessfreya: Yeah, so why isn't num's `ToPrimitive` work? It lets you convert `f64`s to whatever you want, see https://docs.rs/num-traits/0.2.5/src/num_traits/cast.rs.html#323-348 and https://docs.rs/num/0.2.0/num/cast/trait.ToPrimitive.html

nifker: Does it round the float on its own like the previous did?

goddessfreya: :woman_shrugging:

nifker: E.g. this here: https://github.com/rust-windowing/winit/blob/master/src/dpi.rs#L146

nifker: As I wasnt sure if ToPrimitive rounds it on their own.

nifker: Ive just changed it - I hope its better now.

goddessfreya: Afaik, some reading of the `num` source code reveals that `ToPrimitive` rounds down.

nifker: What do you suggest, how to do it? I guess rounding the size down is probably not the best.

goddessfreya: I'm honestly not sure. I'd rather not add a trait, but I'm also not sure `ToPrimitive` will work.

nifker: Is there a test unit for the into/from conversion? Otherwise we could create one based on the previous functionality and see if it works the same as previously.

goddessfreya: To my knowledge there is no unit test, but adding one should be trivial.

aloucks:
Doesn't the DPI overhaul supersede this PR? https://github.com/rust-windowing/winit/pull/1277

@Osspial

nifker:
@aloucks It seems like but Im gonna test it quickly.

Osspial:
@nifker I see no reason not to add array conversions. Go for it 👍

nifker:
As of #1283 I am closing this

PR #1258, opened 2019-11-04
fix broken / missing `raw-window-handle` implementation on android

Noxime:
  - build check was performed and passed (`cargo build --target=aarch64-linux-android`). on a side note, should build checks like these be added to travis? Testing on mobile is hard but making sure something builds isn't very tricky.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+# Version 0.19.5 (2019-11-04)
+
+- On Android, fix the missing `raw-window-handle` support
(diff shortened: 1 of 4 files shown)


Osspial:
Thanks for this PR.

> should build checks like these be added to travis? Testing on mobile is hard but making sure something builds isn't very tricky.

If you'd like to add a Travis Android build to `winit-legacy`, go for it. I expect we'll be figuring out CI on `master` once the Android backend PR is ready, but we'll cross that bridge when we get to it.

Issue #448, opened 2018-04-05
HoveredFile and HoveredFileCancelled events not implemented on Windows


francesca64:

dannyfritz:
Researched this a bit and found these 2 articles explaining how to do it:
https://www.codeproject.com/articles/840/how-to-implement-drag-and-drop-between-your-progra
https://msdn.microsoft.com/en-us/library/4w4z9dwh.aspx

Will have to switch from `WM_DROPFILES` to `OleDropTarget`.

francesca64:
Oh, great! Do you want to implement it?

dannyfritz:
I am looking at implementing it. I'm just secretly hoping it requires I do a PR to `winapi` so I can finally do something with Rust and creating FFI bindings.

francesca64:
I think you're in luck; I'm 99% sure that it requires making a PR to `winapi`. That's a big part of why I never worked on this, since I'm too lazy for that.

retep998:
[`IDropTarget`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms679679)

dannyfritz:
https://github.com/retep998/winapi-rs/pull/631 was merged. Implementation in `winit` is next. If anybody wants to take that up please do. If I start working on it I will post here.

ArturKovacs:
Hi! I seem to have a working implementation. I needed to add one or two things to winapi-rs as well so first I'm going to create a PR for winapi-rs probably tomorrow.

ArturKovacs:
retep998/winapi-rs#651

francesca64:
Fixed via #662

PR #1260, opened 2019-11-05
Windows: Survive a OleInitialize failure. Fixes #1255

### LINKED ISSUE #1255: Windows: 0.20 crashes if CoInitializeEx has been called with MULTITHREADED
The drag/drop support for Windows use OLE drag drop, which requires that COM has been initialized with a single-threaded apartment model. [OleInitialize](https://docs.microsoft.com/en-us/windows/win32/api/ole2/nf-ole2-oleinitialize) is called and we panic if that fails.

So if someone has called CoInitialize(MULTITHREADED), which for example is currently required (for unclear reasons) when trying to use [cpal](https://github.com/RustAudio/cpal), winit just dies.

winit should handle the failure of OleInitialize better and just discard drag/drop support (or fall back to WM_DROPFILES, it's unclear to me why OLE drag/drop is used at all).

hrydgard:
The new file drag/drop code uses OLE.

OleInitialize fails if CoInitialize(COINIT_MULTITHREADED) has previously been called on the current thread, and currently winit panics when this happens (see #1255).

There are many reasons that CoInitialize(COINIT_MULTITHREADED) might have been called before winit is initialized, for example [cpal might be in use](https://github.com/RustAudio/cpal/pull/348). OLE drag/drop requires COINIT_APARTMENTTHREADED but if you don't care about drag/drop, that might not be a concern.

So let's survive OleInitialize failing. Should probably warn about the lost drag/drop functionality in the log somehow - what would be the appropriate way?

Not sure which documentation to update?

### DIFF
--- a/src/platform_impl/windows/window.rs
+++ b/src/platform_impl/windows/window.rs
@@ -74,35 +74,38 @@ impl Window {
         // done. you owe me -- ossi
         unsafe {
             init(w_attr, pl_attr, event_loop).map(|win| {
-                let file_drop_handler = {
-                    use winapi::shared::winerror::{OLE_E_WRONGCOMPOBJ, RPC_E_CHANGED_MODE, S_OK};
-
-                    let ole_init_result = ole2::OleInitialize(ptr::null_mut());
-                    // It is ok if the initialize result is `S_FALSE` because it might happen that
-                    // multiple windows are created on the same thread.
-                    if ole_init_result == OLE_E_WRONGCOMPOBJ {
+                use winapi::shared::winerror::{OLE_E_WRONGCOMPOBJ, RPC_E_CHANGED_MODE, S_OK};
+                let file_drop_handler = match ole2::OleInitialize(ptr::null_mut()) {
+                    OLE_E_WRONGCOMPOBJ => {
                         panic!("OleInitialize failed! Result was: `OLE_E_WRONGCOMPOBJ`");
-                    } else if ole_init_result == RPC_E_CHANGED_MODE {
-                        panic!("OleInitialize failed! Result was: `RPC_E_CHANGED_MODE`");
                     }
+                    RPC_E_CHANGED_MODE => {
+                        // OleInitialize failed because someone is using COINIT_MULTITHREADED, which
+                        // OLE is not compatible with, but there's no reason we can't stumble along
+                        // without drag/drop support.
+                        None
+                    }
+                    _ => {
+                        // It is ok if the initialize result is `S_FALSE` because it might happen that
+                        // multiple windows are created on the same thread.
+                        let file_drop_runner = event_loop.runner_shared.clone();
+                        let file_drop_handler = FileDropHandler::new(
+                            win.window.0,
+                            Box::new(move |event| {
+                                if let Ok(e) = event.map_nonuser_event() {
+                                    file_drop_runner.send_event(e)
+                                }
+                            }),
+                        );
+                        let handler_interface_ptr =
+                            &mut (*file_drop_handler.data).interface as LPDROPTARGET;
 
-                    let file_drop_runner = event_loop.runner_shared.clone();
-                    let file_drop_handler = FileDropHandler::new(
-                        win.window.0,
-                        Box::new(move |event| {
-                            if let Ok(e) = event.map_nonuser_event() {
-                                file_drop_runner.send_event(e)
-                            }
-                        }),
-                    );
-                    let handler_interface_ptr =
-                        &mut (*file_drop_handler.data).interface as LPDROPTARGET;
-
-                    assert_eq!(
-                        ole2::RegisterDragDrop(win.window.0, handler_interface_ptr),
-                        S_OK
-                    );
-                    file_drop_handler
+                        assert_eq!(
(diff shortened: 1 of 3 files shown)


hrydgard:
Turns out drag/drop is not documented in winit at all. Documenting it seems like a separate issue, so I checked off the final box here. Please go ahead and merge.

src/platform_impl/windows/window.rs
```diff
@@ -74,35 +74,38 @@ impl Window {
...
+                    OLE_E_WRONGCOMPOBJ => {
                         panic!("OleInitialize failed! Result was: `OLE_E_WRONGCOMPOBJ`");
-                    } else if ole_init_result == RPC_E_CHANGED_MODE {
-                        panic!("OleInitialize failed! Result was: `RPC_E_CHANGED_MODE`");
                     }
+                    RPC_E_CHANGED_MODE => {
+                        // OleInitialize failed because someone is using COINIT_MULTITHREADED, which
+                        // OLE is not compatible with, but there's no reason we can't stumble along
+                        // without drag/drop support.
+                        None
```
goddessfreya: We should probably add a `warn!` here mentioning drag and drop is borked.

goddessfreya:
Hey @0xpr03, I noticed you are on the testers list. Want to give this PR a test drive?

0xpr03:
I'm back tomorrow night, then I can test this.

hrydgard:
I added a warn! log as requested.

hrydgard:
Just for the record, the Travis failures look unrelated? Should I just rebase this on master maybe?

goddessfreya:
@hrydgard Yeah, a quick rebase should do it.

goddessfreya:
Nvm, I restarted the build and everything's fine now :p

hrydgard:
@goddessfreya it still says `Changes requested` by you, but everything is OK now right?

icefoxen:
Just ran into this myself.  "clear, unambiguous, show-stopping error message" is a bit of an exaggeration.    Well, "show-stopping" isn't.

hrydgard:
I'm opposed to creating extra threads just to resolve this type of issue when it's only necessary if you really want to use OLE drag/drop. In fact, I think I'm in fact questioning the utility of OLE drag/drop in the first place - what's wrong with good old WM_DROPFILES? Actual OLE object use is pretty rare and not something that is worth sacrificing simple usage of a common library like cpal for.

goddessfreya:
@Osspial Out of interest, why is it that drag and drop support is dependent on `COINIT_MULTITHREADED`?

hrydgard:
@goddessfreya  it's the opposite, it's dependent on COINIT_APARTMENTTHREADED.

Osspial:
@hrydgard Don't you have to create an extra thread anyway? [Cpal's method for starting audio output takes control over the calling thread](https://docs.rs/cpal/0.10.0/cpal/traits/trait.EventLoopTrait.html#tymethod.run), and as such there's no way to use Winit's event loop and Cpal's event loop on the same thread at the same time.

> And the error that happens is everything but clear and unambigious if you're not very familiar with Windows programming.

That's... fair. It's more accurate to say that it's *relatively* clear and unambiguous, at least compared to the alternative presented here.

> In fact, I think I'm in fact questioning the utility of OLE drag/drop in the first place - what's wrong with good old `WM_DROPFILES`? 

That's a good question! The discussion around that is in #448 - TL;DR, it's impossible to implement `HoveredFile` and `HoveredFileCancelled` without the OLE API.

Osspial:
Since https://github.com/RustAudio/cpal/pull/354 has been merged, I'm going to go ahead and close this.

hrydgard:
Don't know, looks like time to have another look! We worked around it before by forking cpal to use COINIT_APARTMENTTHREADED, which it already survived just fine for our use case at least.

azriel91:
Heya @hrydgard, have you managed to find time to check?
I haven't managed to understand if the `cpal` fix means this one isn't necessary

hrydgard:
We haven't yet switched to the new cpal yet, it fell down the list of priorities a little. Hope to get to it soon.

kcking:
I can confirm this is still an issue on the latest cpal, but changing to COINIT_APARTMENTTHREADED in a fork seems to fix this issue.

ArturKovacs:
The annoyance here is that `OleInitialize` calls `CoInitializeEx` with `COINIT_APARTMENTTHREADED` and according to the msdn docs there is no way around this because the ole functionality is simply not thread-safe. However, any other crate within the same program could offer a choice to call `CoInitializeEx` with `COINIT_APARTMENTTHREADED` to match Winit's behaviour. But I know that we are discussing Winit and not "any other crate" so here's my proposal.

I think there should be an option in Winit that allows to skip the ole initialization and thereby disable drag-drop and file-hover events. If this option is not set before initialization, any failure of the ole initialization triggers a panic like before.

kotauskas:
Any progress on this?

Osspial:
@kotauskas #1524 implements a function that disables OLE drag-drop, which lets Winit function with `COINIT_MULTITHREADED`. I'm inclined to close this PR once that's merged.

hrydgard:
Looks good. I'm happy with closing this once that's in.

PR #1262, opened 2019-11-05
X11: Fix modifiers being reported after release

### LINKED ISSUE #1259: X11 reporting modifiers after release
It looks like in some cases the state of modifiers is not released correctly, leading to key events having incorrect modifiers attached.

More detail can be found in https://github.com/jwilm/alacritty/issues/2906.

murarth:
* Moves `ModifiersChanged` variant from `WindowEvent` to `DeviceEvent`

Fixes #1259

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 - On macOS, fix application termination on `ControlFlow::Exit`
+- On X11, fix key modifiers being incorrectly reported.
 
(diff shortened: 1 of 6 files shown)


chrisduerr:
I can't reproduce the bug on my machine, but I can verify that the modifiers now behave the same way on X11 as they do on macOS. So when pressing shift, the modifiers state has `shift` set both for the shift key press itself and for all following events.

So everything looking good with regards to functionality from a consumer perspective. Thanks for the quick work @murarth, nicely done.

PR #1263, opened 2019-11-07
Determine im before close it.

uhuntu:

### DIFF
--- a/src/platform_impl/linux/x11/ime/inner.rs
+++ b/src/platform_impl/linux/x11/ime/inner.rs
@@ -5,7 +5,9 @@ use super::{ffi, XConnection, XError};
 use super::{context::ImeContext, input_method::PotentialInputMethods};
 
 pub unsafe fn close_im(xconn: &Arc<XConnection>, im: ffi::XIM) -> Result<(), XError> {
-    (xconn.xlib.XCloseIM)(im);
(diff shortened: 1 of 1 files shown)


src/platform_impl/linux/x11/ime/inner.rs
```diff
@@ -5,7 +5,9 @@ use super::{ffi, XConnection, XError};
 use super::{context::ImeContext, input_method::PotentialInputMethods};
 
 pub unsafe fn close_im(xconn: &Arc<XConnection>, im: ffi::XIM) -> Result<(), XError> {
-    (xconn.xlib.XCloseIM)(im);
+    if !im.is_null() {
```
goddessfreya: Huh, I swear someone had already done this.

Anyways, we should issue a warning if it is null.

murarth:
I don't think it's possible for a null pointer to reach this point. The `open_im` function returns `None` if the raw pointer returned by Xlib is null.

I suppose there's no harm in double-checking. But perhaps a better solution would be to have an `XIM` wrapper type so that a raw pointer is not floating around the various parts of winit code.

goddessfreya:
That's a good catch, @murarth. I think wrapping the pointer in `std::ptr::NonNull` would be satisfactory.

PR #1271, opened 2019-11-10
Fix invalid_value lint triggering on mem::zeroed of CFRunLoopSourceContext

thomcc:
Without you get 
```
warning: the type `platform_impl::platform::observer::CFRunLoopSourceContext` does not permit zero-initialization
   --> src/platform_impl/macos/event_loop.rs:134:55
    |
134 |             let mut context: CFRunLoopSourceContext = mem::zeroed();
    |                                                       ^^^^^^^^^^^^^
    |                                                       |
    |                                                       this code causes undefined behavior when executed
    |                                                       help: use `MaybeUninit<T>` instead
    |
    = note: `#[warn(invalid_value)]` on by default
note: Function pointers must be non-null (in this struct field)
   --> src/platform_impl/macos/observer.rs:96:5
    |
96  |     pub retain: extern "C" fn(*const c_void) -> *const c_void,
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

    - Was that type public?

### DIFF
--- a/src/platform_impl/macos/event_loop.rs
+++ b/src/platform_impl/macos/event_loop.rs
@@ -132,7 +132,7 @@ impl<T> Proxy<T> {
             // process user events through the normal OS EventLoop mechanisms.
             let rl = CFRunLoopGetMain();
             let mut context: CFRunLoopSourceContext = mem::zeroed();
-            context.perform = event_loop_proxy_handler;
+            context.perform = Some(event_loop_proxy_handler);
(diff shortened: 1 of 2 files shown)


retep998:
Oh man, so before this change this was super duper undefined behavior. It's amazing nobody has reported issues about this.

PR #1275, opened 2019-11-12
Add web examples

ryanisaacg:

### DIFF
--- a/examples/stdweb_window.rs
+++ b/examples/stdweb_window.rs
@@ -1,18 +1,19 @@
-use winit::{
-    event::{Event, WindowEvent},
-    event_loop::{ControlFlow, EventLoop},
-    platform::web::WindowExtStdweb,
-    window::WindowBuilder,
-};
-// Note: stdweb is aliased to std_web in winit's Cargo.toml
-// In most cases, it should not be necessary to include this line
-use std_web as stdweb;
-use stdweb::{
-    traits::*,
-    web::document
-};
-
+#[cfg(all(target_arch = "wasm32", feature = "stdweb"))]
 fn main() {
+    use winit::{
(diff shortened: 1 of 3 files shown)


Osspial:
I ran `cargo web start --example stdweb_window --features stdweb`, but the code included in this PR doesn't seem to be enough to instantiate a webpage that contains a Winit window. Additionally, `cargo web` is giving me this message:

```
If you need to serve any extra files put them in the 'static' directory
in the root of your crate; they will be served alongside your application.
You can also put a 'stdweb_window-static' directory in your 'examples' directory.
Your application is being served at '/stdweb_window.js'. It will be automatically
rebuilt if you make any changes in your code.
```

Would it be possible to include the static files in this PR? Additionally, could we add web setup documentation to the crate's root docs?

ryanisaacg:
@Osspial 

> I ran cargo web start --example stdweb_window --features stdweb, but the code included in this PR doesn't seem to be enough to instantiate a webpage that contains a Winit window. 

Nothing is rendered in the window, but it's there for me at least. Are you not seeing events logged / a canvas in the DOM?

We could insert some code to use webgl or HTML5 canvas to color-fill the canvas to make it obvious.

> Additionally, cargo web is giving me this message:

Cargo web will give you this message if you don't have a `static` directory. It's just a message that warns you where to put your static assets; this PR has none.

Osspial:
@ryanisaacg ah, alright. I just checked the console log and events are indeed getting printed.

> We could insert some code to use webgl or HTML5 canvas to color-fill the canvas to make it obvious.

Would it be easier to set the page's background color, so it contrasts against the canvas' default uncleared color?

ryanisaacg:
#1301 does this better, so I'll just make the background color change after we merge that.

Issue #1135, opened 2019-08-30
Make PhysicalPosition integer, make dedicated Input* types as necessary


kennylevinsen:
Splitting out from a comment in #939.

As part of the DPI overhaul, PhysicalSize has been converted to an integer type to better represent the physical pixels it is designed to map to.

PhysicalPosition was left as a float to allow certain types of high resolution input to have sub-pixel resolution. However, as certain types of devices also report input *geometries*, I believe that serving sub-pixel resolution input devices require fully floating types.

Considering that using floating types for physical pixel addressing is awkward and confusing, I believe that these are simply two problems that need individual solutions. Therefore, I propose converting PhysicalPosition to integer as well, and introducing separate types for sub-pixel resolution input as necessary.

This leaves us with clean types that have clearly defined purposes, rather that having a situation where we are hammering everything into the same shape.

ZeGentzy:
Personally, I was suprized to learn that `PhysicalPosition` wasn't an integer and thought that it was something overlooked when drafting the dpi overhaul. It's not obvious to me, and I imagine others, that `PhysicalPosition` sometimes could report sub-pixel numbers, and so I think splitting `PhysicalPosition` would be a good idea. I suggest we also make a respective `InputSize` struct, or whatever we want to name it, just for consistency's sake, even if unused in the codebase, but beyond that I fully support this idea.

I imagine you folks would be interested in this discussion: @Osspial @mtak- @murarth @aleksijuvani @ryanisaacg

ZeGentzy:
Ah, and @mb64 and @philip-alldredge, since you two appear to be working on the android backend, which I imagine reports touch related stuff with ssub-pixel resolution.

mtak-:
That makes sense to me. I'm seeing non-integral values for [majorRadius](https://developer.apple.com/documentation/uikit/uitouch/1618106-majorradius?language=objc) on iOS.

aleksijuvani:
It feels a bit iffy to me to introduce new types when we could be just using floats everywhere. It's not surprising to me that floats would be cast towards zero where sub-pixel accuracy is not available. That said, I don't feel particularly strongly about this.

Osspial:
[...]

PR #1277, opened 2019-11-14
Make size/position types generic over pixel type

Osspial:

Implements the changes suggested in https://github.com/rust-windowing/winit/issues/1135#issuecomment-528543352. I haven't gotten around to updating the type signatures on non-Windows platforms but I'll go through the CI logs after they fail and update them where necessary.

### DIFF
--- a/src/dpi.rs
+++ b/src/dpi.rs
@@ -155,7 +155,10 @@ impl<P> LogicalPosition<P> {
 
 impl<P: Pixel> LogicalPosition<P> {
     #[inline]
-    pub fn from_physical<T: Into<PhysicalPosition<X>>, X: Pixel>(physical: T, dpi_factor: f64) -> Self {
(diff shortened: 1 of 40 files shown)


src/dpi.rs
```diff
@@ -297,7 +297,7 @@ impl<P: Pixel, X: Pixel> From<(X, X)> for LogicalSize<P> {
 }
 
 impl<P: Pixel, X: Pixel> Into<(X, X)> for LogicalSize<P> {
-    fn into(self: LogicalSize<P>) -> (X, X) {
+    fn into(self: LogicalSize<f64><P>) -> (X, X) {
```
goddessfreya: Has this syntax ever been possible? :eyes:

Osspial: No, that was an error with my regexes.

goddessfreya: Heh, I suspected so :p

PR #1279, opened 2019-11-15
X11: Fix incorrect modifiers when events are missed

murarth:
* Syncs modifier state with state data in X key/button/motion events.
* Fixes modifier state in XWayland, as xinput2 raw input events will
  not be received when a window does not have focus.
* Removes `impl From<_> for ModifiersState` on X11/Wayland API types,
  replacing them with `pub(crate)` methods.

Addresses an issue brought up in #1259

### DIFF
--- a/src/platform_impl/linux/x11/event_processor.rs
+++ b/src/platform_impl/linux/x11/event_processor.rs
@@ -120,6 +120,26 @@ impl<T: 'static> EventProcessor<T> {
             return;
         }
 
+        // We can't call a `&mut self` method because of the above borrow,
+        // so we use this macro for repeated modifier state updates.
+        macro_rules! update_modifiers {
+            ( $state:expr , $modifier:expr ) => {{
+                match ($state, $modifier) {
+                    (state, modifier) => {
+                        if let Some(modifiers) =
+                            self.device_mod_state.update_state(&state, modifier)
+                        {
+                            let device_id = mkdid(util::VIRTUAL_CORE_KEYBOARD);
+                            callback(Event::DeviceEvent {
+                                device_id,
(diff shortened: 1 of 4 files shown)


src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -687,7 +709,18 @@ impl<T: 'static> EventProcessor<T> {
                         let window_id = mkwid(xev.event);
                         let new_cursor_pos = (xev.event_x, xev.event_y);
 
-                        let modifiers = ModifiersState::from(xev.mods);
+                        let modifiers = ModifiersState::from_x11(&xev.mods);
+
+                        if let Some(modifiers) =
+                            self.device_mod_state.update_state(&modifiers, None)
```
kchibisov: I feel like this check and later modifiers changed event send should be extracted to an `inline` function, since they are mostly identical (I'm about XI_Motion/XI_ButtonPress/ffi::KeyPress) and share the same logic.

kchibisov:
Besides my comment, the PR works fine under XWayland now.

src/platform_impl/linux/x11/event_processor.rs
murarth: Fair point. I've replaced the repeated `if let` with a macro invocation, using a macro defined at the top of the function. As noted in the comment, a `&mut self` method can't be used because `&self.target` is borrowed for the entire body of the function.

murarth:
I realized a possible bug in updating state: Keycodes were not removed from the set when their modifier was removed by an update.

So, I pushed a new commit to fix it. Requesting re-review.

PR #1283, opened 2019-11-21
Dpi Type conversions into/from arrays

nifker:


src/platform_impl/ios/view.rs
```diff
@@ -508,9 +508,7 @@ pub unsafe fn create_window(
             let () = msg_send![uiscreen, setCurrentMode: video_mode.video_mode.screen_mode];
             msg_send![window, setScreen:video_mode.monitor().ui_screen()]
         }
-        Some(Fullscreen::Borderless(ref monitor)) => {
```
murarth: It seems you're not using the latest version of rustc/rustfmt. This change needs to be reverted to pass `cargo fmt --check`.

nifker: It seems like --check is an invalid argument.

nifker: Seems like its fixed now

PR #1285, opened 2019-11-22
Stop refocusing window when switching from fullscreen to windowed

### LINKED ISSUE #1284: Windows: Switching from borderless fullscreen to windowed when not focused will refocus immediately
<details>
  <summary>Old post</summary>
I wanted to add the ability to tab-out or to open another window on top of a borderless fullscreen window.

The easiest way I could come up with was when the window is unfocused I could switch the window from borderless fullscreen to just a borderless window. Turned out the issue is that switching from fullscreen to not-fullscreen will send another focus event right after, so any focus trigger wouldn't work.

I played around with the windows internals quite a bit to try to figure this out, trying to add `SWP_NOACTIVATE` everywhere, to no avail.

So the other solution I came up with is to just remove `ALWAYS_ON_TOP` from borderless, which does solve the problem.

The feature I need is the following:
-  I want the use to be able to unfocus or alt-tab out of a borderless fullscreen window
-  while still being able see the borderless window, so no minimizing or hiding
-  while still having the fullscreen marker, because it gives the OS and third-party programs the correct indicators

The current limitation that hinder this are:
-  trying to switch in and out of fullscreen with the focus event doesn't work, because Windows will always send another focus event right after, maybe somebody can fix that?
-  there is no way to tell a borderless fullscreen window to not be always on top

My proposed solutions:
1.  make `set_always_on_top` work in combination with windows that are in fullscreen mode, `set_fullscreen` will always set `ALWAYS_ON_TOP` though.
2.  make a third Fullscreen mode, maybe `LightBorderless` (this is terrible somehow).
3.  change fullscreen mode to not force `ALWAYS_ON_TOP`
4.  somehow fix windows not to refocus when changing between fullscreen and not-fullscreen

Obviously the best solution would be four, but I tried and I don't think its possible.
I think the easiest solution would be 1., has the least amount of breaking changes.

Would love to make a PR as soon as there is some consensus.
</details>

Switching from borderless fullscreen to windowed when a window is not focused will immediatly focus the window again.

Apparently the issue is not using `SWP_NOACTIVATE` on `SetWindowPos` in certain places.

daxpedda:

This should stop some functionality on unfocused windows to automatically focus them.
This affects for example switching from fullscreen to windowed or changing the outer position of windows.

Fixes #1284.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - On X11, fix misreporting DPI factor at startup.
 - On X11, fix events not being reported when using `run_return`.
 - On X11, fix key modifiers being incorrectly reported.
+- On Windows, fix focusing unfocused windows when switching from fullscreen to windowed.
 
 # 0.20.0 Alpha 4 (2019-10-18)
 
--- a/src/platform_impl/windows/window.rs
+++ b/src/platform_impl/windows/window.rs
@@ -202,7 +202,10 @@ impl Window {
                 y as c_int,
                 0,
                 0,
-                winuser::SWP_ASYNCWINDOWPOS | winuser::SWP_NOZORDER | winuser::SWP_NOSIZE,
+                winuser::SWP_ASYNCWINDOWPOS
+                    | winuser::SWP_NOZORDER
+                    | winuser::SWP_NOSIZE
+                    | winuser::SWP_NOACTIVATE,
             );
             winuser::UpdateWindow(self.window.0);
         }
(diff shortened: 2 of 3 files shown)


goddessfreya:
Hey @notriddle, you appear to be on the testers list. Want to give this a test?

notriddle:
It seems to work. I applied this patch to one of the examples:

```patch
diff --git a/examples/fullscreen.rs b/examples/fullscreen.rs
index d4b83bb0..44983154 100644
--- a/examples/fullscreen.rs
+++ b/examples/fullscreen.rs
@@ -35,6 +35,9 @@ fn main() {
         match event {
+               WindowEvent::Focused(b) => {
+                       window.set_fullscreen(if b { fullscreen.clone() } else { None })
+               }
                 WindowEvent::KeyboardInput {
                     input:
                         KeyboardInput {
```

It works. So it seems like this patch fixes the original bug. For context, I'm running Windows 10.

PR #1294, opened 2019-11-30
Have EventLoopClosed contain the original event

### LINKED ISSUE #1292: EventLoopClosed should contain the original event
This is a feature request.

Other, more low-level `channel`-alike crates provide each provide a `send` method with an `Err(...(T))` result for the failure case to recover the original event without a need to `clone` it before-hand, but `winit` (as a high-level interface/wrapper) doesn't. I think this is a bit inconsistent. If there are some underlying platform implementations, where this is not possible, I think `EventLoopProxy::send_event` should return an `Option`-alike type (or a custom enum) in the error case.

Examples:
* [`calloop`](https://docs.rs/calloop/0.4.4/calloop/channel/enum.SendError.html)
* [`crossbeam-channel`](https://docs.rs/crossbeam-channel/0.4.0/crossbeam_channel/struct.SendError.html)

zserik:
      (via CI, shouldn't change behavoir of user code except involved types when using `EventLoopProxy`, because the return value of `EventLoopProxy::send_event` changed)
- Created or updated an example program if it would help users understand this functionality
      (appropriate example to modify would be `examples/custom_events.rs`, but I'm unsure which modification would be appropriate.)
- Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented
      (no appropriate section found, where this should be added)

Fixes #1292.

### DIFF
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -199,7 +199,7 @@ impl<T: 'static> EventLoopProxy<T> {
     /// function.
     ///
     /// Returns an `Err` if the associated `EventLoop` no longer exists.
-    pub fn send_event(&self, event: T) -> Result<(), EventLoopClosed> {
+    pub fn send_event(&self, event: T) -> Result<(), EventLoopClosed<T>> {
         self.event_loop_proxy.send_event(event)
     }
 }
@@ -211,17 +211,17 @@ impl<T: 'static> fmt::Debug for EventLoopProxy<T> {
 }
 
 /// The error that is returned when an `EventLoopProxy` attempts to wake up an `EventLoop` that
-/// no longer exists.
-#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
-pub struct EventLoopClosed;
+/// no longer exists. Contains the original event given to `send_event`.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+pub struct EventLoopClosed<T>(pub T);
 
-impl fmt::Display for EventLoopClosed {
+impl<T: fmt::Debug> fmt::Display for EventLoopClosed<T> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(f, "{}", error::Error::description(self))
     }
 }
 
-impl error::Error for EventLoopClosed {
+impl<T: fmt::Debug> error::Error for EventLoopClosed<T> {
     fn description(&self) -> &str {
         "Tried to wake up a closed `EventLoop`"
     }
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -157,7 +157,7 @@ impl EventLoop {
 }
 
 impl EventLoopProxy {
-    pub fn wakeup(&self) -> Result<(), ::EventLoopClosed> {
+    pub fn wakeup(&self) -> Result<(), ::EventLoopClosed<()>> {
         android_glue::wake_event_loop();
         Ok(())
     }
--- a/src/platform_impl/linux/mod.rs
+++ b/src/platform_impl/linux/mod.rs
@@ -650,7 +650,7 @@ impl<T: 'static> EventLoop<T> {
 }
 
 impl<T: 'static> EventLoopProxy<T> {
-    pub fn send_event(&self, event: T) -> Result<(), EventLoopClosed> {
+    pub fn send_event(&self, event: T) -> Result<(), EventLoopClosed<T>> {
         match *self {
             EventLoopProxy::Wayland(ref proxy) => proxy.send_event(event),
             EventLoopProxy::X(ref proxy) => proxy.send_event(event),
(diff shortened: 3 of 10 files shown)


src/platform_impl/android/mod.rs
```diff
@@ -157,7 +157,7 @@ impl EventLoop {
 }
 
 impl EventLoopProxy {
-    pub fn wakeup(&self) -> Result<(), ::EventLoopClosed> {
+    pub fn wakeup(&self) -> Result<(), ::EventLoopClosed<()>> {
```
zserik: I haven't found an appropriate `send_event` method for Android, this was the most-similiar method.

src/platform_impl/linux/wayland/event_loop.rs
```diff
@@ -282,8 +282,13 @@ impl<T: 'static> Clone for EventLoopProxy<T> {
...
 
 impl<T: 'static> EventLoopProxy<T> {
-    pub fn send_event(&self, event: T) -> Result<(), EventLoopClosed> {
-        self.user_sender.send(event).map_err(|_| EventLoopClosed)
+    pub fn send_event(&self, event: T) -> Result<(), EventLoopClosed<T>> {
+        self.user_sender.send(event).map_err(|e| {
+            EventLoopClosed(match e {
+                ::calloop::channel::SendError::Disconnected(x) => Some(x),
+                _ => None,
+            })
```
zserik: This `match` statement could be replaced with an `if let` statement.

src/event_loop.rs
```diff
@@ -212,16 +212,23 @@ impl<T: 'static> fmt::Debug for EventLoopProxy<T> {
...
+/// Optionally contains the original event given to `send_event` if the platform supports it.
+#[derive(Copy, Clone, PartialEq, Eq, Hash)]
+pub struct EventLoopClosed<T>(pub Option<T>);
 
-impl fmt::Display for EventLoopClosed {
+impl<T> fmt::Debug for EventLoopClosed<T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "EventLoopClosed {{ .. }}")
+    }
+}
```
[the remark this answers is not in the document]
zserik: It would require that `T` derives `Debug`, thus the `std::error::Error` impl would require `T: Debug`, too.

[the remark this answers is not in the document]
zserik: I implemented it similiar to `EventLoopProxy`.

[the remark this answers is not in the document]
Osspial: I'd prefer you do a normal `#[derive(Debug)]` here. `EventLoopProxy` has a manual implementation because it doesn't actually contain any instances of `T`, so it never needs to print `T` out, whereas this is essentially a newtype.

src/platform_impl/linux/mod.rs
```diff
@@ -504,20 +510,31 @@ pub enum EventLoop<T: 'static> {
     X(x11::EventLoop<T>),
 }
 
-pub enum EventLoopProxy<T: 'static> {
-    X(x11::EventLoopProxy<T>),
-    Wayland(wayland::EventLoopProxy<T>),
+/// A handle that can be sent across threads and used to wake up the `EventLoop`.
+pub struct EventLoopProxy<T: 'static> {
+    user_sender: calloop::channel::Sender<T>,
 }
```
Osspial: cc @murarth @goddessfreya @vberger for review. What do you guys think of this change?

Osspial: cc @murarth @goddessfreya @vberger for review. What do you guys think of this change?

murarth: Seems like a sensible change to me. I don't anticipate a divergence in the X11 and Wayland implementations of this basic communication channel.

zserik: I moved the offending code change into a separate commit.

src/platform_impl/linux/wayland/event_loop.rs
```diff
@@ -282,8 +282,14 @@ impl<T: 'static> Clone for EventLoopProxy<T> {
...
 
 impl<T: 'static> EventLoopProxy<T> {
-    pub fn send_event(&self, event: T) -> Result<(), EventLoopClosed> {
-        self.user_sender.send(event).map_err(|_| EventLoopClosed)
+    pub fn send_event(&self, event: T) -> Result<(), EventLoopClosed<T>> {
+        self.user_sender.send(event).map_err(|e| {
+            EventLoopClosed(if let ::calloop::channel::SendError::Disconnected(x) = e {
+                x
+            } else {
+                unreachable!()
```
goddessfreya: nit: should still return `Err` and not panic. Same for x11.

zserik: https://github.com/rust-windowing/winit/pull/1294#discussion_r352815218

zserik: https://twitter.com/ZelleTTY/status/1202678011981697027

zserik: https://github.com/rust-windowing/winit/pull/1294#discussion_r352815218

zserik: How? This was already considered and dismissed.

goddessfreya: Ah, I did not notice. Apologies.

PR #1297, opened 2019-11-30
X11: Report `CursorMoved` when touch event occurs

### LINKED ISSUE #1261: On X11, touch events can move the cursor without CursorMoved event
This is exemplified in [KAS](https://github.com/dhardy/kas) (easy to see in the calculator example):

-   cursor position is updated when a `CursorMoved` and `CursorLeft` events are received but not otherwise; other than as reported here this appears to be sufficient
-   widgets are highlighted based on the cursor position, as updated above
-   on X11, touch events (first finger only if multiple presses) move the mouse cursor; this may be because until relatively recently touch events simply emulated mouse clicks on X11
-   my app observes all touch events correctly
-   when the first touch event starts, my app observes the mouse position being moved via `CursorMoved`
-   when this touch moves (`TouchPhase::Moved`), X11 visually moves the mouse cursor, but my app does not receive `CursorMoved` events

Result: pressing a button with the finger and dragging off of it results in the start button being highlighted (because the app believes the mouse is hovering over it) *and* the button under the finger being highlighted (because the app highlights all drags).

I guess this is an X11 bug. What's your experience with issues like this?

murarth:
Closes #1261


murarth:
@dhardy Can you confirm that this PR resolves the issue?

dhardy:
Ah, sorry. Yes, it does, but it causes another more subtle bug because now all touch events fire `CursorMoved` events (potentially leaving my UI highlighting the wrong button after multiple touch events end).

The solution would be to only fire `CursorMoved` events for the "first" touch event (i.e. only for touch events (identified by [id](https://docs.rs/winit/0.20.0-alpha4/winit/event/struct.Touch.html#structfield.id)) where no other touch was active at the start time).

murarth:
@dhardy I've implemented the fix you described. Does everything seem to be working as expected now?

dhardy:
~I tested the commit I made and it worked perfectly, but I just tested this version and it doesn't. The issue is that `first_touch` will set a new `id` as soon as the previous "first touch" ends; to match X11 behaviour only a new touch (i.e. `Phase::Begin`) should be able to become a "first touch".~

I didn't think to test a third touch event started after the first ends. Both commits fail in this case. Instead, a touch should only be able to become a "first commit" if there are no other active touch events.

murarth:
@dhardy Oh, what a silly mistake on my part. I'm glad you caught it. I've pushed another fix.

dhardy:
Just tested, and it passes my obscure test, but not the much simpler test of touch and drag across buttons: the first button remains highlighted. I will investigate tomorrow. (Also, touch-cancel events seem to be causing issues but this may be my UI).

src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -1216,3 +1241,19 @@ impl<T: 'static> EventProcessor<T> {
         }
     }
 }
+
+fn is_first_touch(m: &mut HashSet<u64>, id: u64, phase: TouchPhase) -> bool {
+    let is_first = m.is_empty();
```
dhardy: This logic is incorrect except on TouchPhase::Started. You also don't actually need a hashmap here.

murarth: Fixed again. Hopefully.

murarth: Also, I changed the code from using a `HashSet` to a simple count of concurrent running touch events. As long as winit doesn't receive a duplicate `Ended` event, it should work as well as the `HashSet`.

dhardy:
Is there any reason you keep re-writing my patches? I already tested mine, now I have to test this one too.

murarth:
@dhardy I didn't realize you were writing any patches. It seems we each arrived at the same solution independently.

dhardy:
The first I created a pull-request against your fork, and the second I linked above.

murarth:
Well, I'm sorry, but I didn't see either of those. I just saw the one comment on my code, so I wrote my own fix, which appears to have essentially the same logic.

If you've tested your version and can confirm that it resolves the issue, I think this PR is ready to be reviewed and merged.

src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -1216,3 +1238,23 @@ impl<T: 'static> EventProcessor<T> {
...
+    match phase {
+        TouchPhase::Started => {
+            if *num == 0 {
+                *first = Some(id);
+            }
+            *num += 1;
+        }
+        TouchPhase::Cancelled | TouchPhase::Ended => {
+            if *first == Some(id) {
+                *first = None;
```
dhardy: My understanding is that this is correct, but unnecessary: if `num == 0` then `first == Some(id)` is never evaluated (except on `TouchPhase::Ended` for the last event, when the above ids are still valid). As soon as the next event starts, `first` is re-assigned. Thus in my version I didn't bother using an `Option` here.

murarth:
@goddessfreya I believe this PR is now ready for review.

PR #1235, opened 2019-10-21
Implement revamped `RedrawRequested` on macOS

hecrj:
Related to #1082.

I have tested this with https://github.com/hecrj/iced/pull/22 and it seems to work well.

However, I am not very familiar with macOS windowing APIs, thus I am not sure if any further changes are necessary.

The PR most likely won't compile because it needs a fix that was merged to `master` recently: https://github.com/rust-windowing/winit/commit/af3ef52252dfad178d897b036876c2575c20f1f3. I will be happy to rebase the changes once `redraw-requested-2.0` catches up!

- [x] Tested on all platforms changed
- [x] Compilation warnings were addressed
- [x] `cargo fmt` has been run on this branch
- [ ] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users
- [ ] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior
- [ ] Created or updated an example program if it would help users understand this functionality
- [ ] Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented

PR #1299, opened 2019-12-01
Implement revamped `RedrawRequested` on iOS.

simlay:
Hi Friends,

Saw the need to help with iOS work for the RedrawRequest 2.0 API in #1082 from @Osspial's request and thought I'd do it.

Honestly, it seemed like too little work but it's quite similar to #1235.

I've tested this with for https://github.com/hecrj/iced/pull/57 in the [ios-examples repo](https://github.com/iced-rs/ios-examples).

- [ ] Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented.

### DIFF
--- a/src/platform_impl/ios/app_state.rs
+++ b/src/platform_impl/ios/app_state.rs
@@ -779,7 +779,7 @@ pub unsafe fn handle_main_events_cleared() {
         .main_events_cleared_transition()
         .into_iter()
         .map(|window| Event::RedrawRequested(RootWindowId(window.into())))
-        .map(|_: Event<Never>| Event::RedrawEventsCleared);
+        .chain(std::iter::once(Event::RedrawEventsCleared));
     drop(this);
 
     handle_nonuser_events(redraw_events);
--- a/src/platform_impl/ios/app_state.rs
+++ b/src/platform_impl/ios/app_state.rs
@@ -778,7 +778,8 @@ pub unsafe fn handle_main_events_cleared() {
     let redraw_events = this
         .main_events_cleared_transition()
         .into_iter()
-        .map(|window| Event::RedrawRequested(RootWindowId(window.into())));
+        .map(|window| Event::RedrawRequested(RootWindowId(window.into())))
+        .map(|_: Event<Never>| Event::RedrawEventsCleared);
     drop(this);
 
     handle_nonuser_events(redraw_events);
--- a/src/platform_impl/ios/app_state.rs
+++ b/src/platform_impl/ios/app_state.rs
@@ -12,7 +12,7 @@ use std::{
 use objc::runtime::{BOOL, YES};
 
 use crate::{
-    event::{Event, StartCause, WindowEvent},
+    event::{Event, StartCause},
     event_loop::ControlFlow,
     platform_impl::platform::{
         event_loop::{EventHandler, Never},
@@ -51,11 +51,7 @@ enum UserCallbackTransitionResult<'a> {
 
 impl Event<Never> {
     fn is_redraw(&self) -> bool {
-        if let Event::WindowEvent {
-            window_id: _,
-            event: WindowEvent::RedrawRequested,
-        } = self
-        {
+        if let Event::RedrawRequested(_) = self {
             true
         } else {
             false
@@ -776,16 +772,13 @@ pub unsafe fn handle_main_events_cleared() {
 
     // User events are always sent out at the end of the "MainEventLoop"
     handle_user_events();
-    handle_nonuser_event(Event::EventsCleared);
+    handle_nonuser_event(Event::MainEventsCleared);
 
     let mut this = AppState::get_mut();
     let redraw_events = this
         .main_events_cleared_transition()
         .into_iter()
-        .map(|window| Event::WindowEvent {
-            window_id: RootWindowId(window.into()),
-            event: WindowEvent::RedrawRequested,
-        });
+        .map(|window| Event::RedrawRequested(RootWindowId(window.into())));
     drop(this);
 
     handle_nonuser_events(redraw_events);
--- a/src/platform_impl/ios/app_state.rs
+++ b/src/platform_impl/ios/app_state.rs
@@ -775,10 +775,11 @@ pub unsafe fn handle_main_events_cleared() {
     handle_nonuser_event(Event::MainEventsCleared);
 
     let mut this = AppState::get_mut();
-    let mut redraw_events : Vec<Event<Never>> = this
+    let mut redraw_events: Vec<Event<Never>> = this
         .main_events_cleared_transition()
         .into_iter()
-        .map(|window| Event::RedrawRequested(RootWindowId(window.into()))).collect();
+        .map(|window| Event::RedrawRequested(RootWindowId(window.into())))
+        .collect();
 
     if !redraw_events.is_empty() {
         redraw_events.push(Event::RedrawEventsCleared);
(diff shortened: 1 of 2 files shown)


hecrj:
Very happy to see this! :smile:

src/platform_impl/ios/app_state.rs
```diff
@@ -776,16 +772,13 @@ pub unsafe fn handle_main_events_cleared() {
...
 
     let mut this = AppState::get_mut();
     let redraw_events = this
         .main_events_cleared_transition()
         .into_iter()
-        .map(|window| Event::WindowEvent {
-            window_id: RootWindowId(window.into()),
-            event: WindowEvent::RedrawRequested,
-        });
+        .map(|window| Event::RedrawRequested(RootWindowId(window.into())));
```
hecrj: I think we may need to send a `RedrawEventsCleared` after this.

simlay: Good catch. All the other platforms send `RedrawEventsCleared` afterward.

Osspial:
Thank you so much for taking the time to implement this!

@aleksijuvani You mentioned that you'd be willing to review the iOS `RedrawRequested` PR. Could you test this out and do a quick review pass over it?

src/platform_impl/ios/app_state.rs
```diff
@@ -776,16 +772,14 @@ pub unsafe fn handle_main_events_cleared() {
...
     let mut this = AppState::get_mut();
     let redraw_events = this
         .main_events_cleared_transition()
         .into_iter()
-        .map(|window| Event::WindowEvent {
-            window_id: RootWindowId(window.into()),
-            event: WindowEvent::RedrawRequested,
-        });
+        .map(|window| Event::RedrawRequested(RootWindowId(window.into())))
+        .map(|_: Event<Never>| Event::RedrawEventsCleared);
```
Osspial: You want to use `chain(std::iter::once(Event::RedrawEventsCleared)` here, not `map`. `map` turns all the `RedrawRequested` events into `RedrawEventsCleared` events.

simlay: Yup! I was just testing out similar logic.

simlay:
I adjusted the incorrect logic and tested this with the [ios-example](https://github.com/rust-windowing/glutin) in the glutin project. It now runs `RedrawEventsCleared` after `MainEventsCleared` and after every `RedrawRequested`.

src/platform_impl/ios/app_state.rs
```diff
@@ -776,16 +772,14 @@ pub unsafe fn handle_main_events_cleared() {
...
     let mut this = AppState::get_mut();
     let redraw_events = this
         .main_events_cleared_transition()
         .into_iter()
-        .map(|window| Event::WindowEvent {
-            window_id: RootWindowId(window.into()),
-            event: WindowEvent::RedrawRequested,
-        });
+        .map(|window| Event::RedrawRequested(RootWindowId(window.into())))
+        .chain(std::iter::once(Event::RedrawEventsCleared));
```
aleksijuvani: There are two separate paths that can be taken for issuing `RedrawRequested` events, depending on how the Metal layer was set up. The redraw can either be queued, or it can be issued immediately. See the discussion [here.](https://github.com/rust-windowing/winit/pull/1133/files/b7de1d43cb55c2c10a7197d0edaf6324c3c342a1#r319364974)

Right now this is only taking the queued redraws into account. That is to say, if a redraw event is emitted from [`draw_rect`](https://github.com/rust-windowing/winit/blob/400f75a2b33fc765a3173992cccc7d94defafde5/src/platform_impl/ios/view.rs#L105-L108) (this is invoked by `setNeedsDisplay`), `RedrawEventsCleared` and `RedrawRequested` will be emitted in the wrong order.

Perhaps we can change `request_redraw` to always call `queue_gl_or_metal_redraw` and get rid of the `draw_rect` code path. This would ensure that the events are issued in the correct order.

simlay: Ah. Good catch. I was unaware of the `setNeedsDisplay` path. I'm a little apprehensive to remove the `draw_rect` codepath because it would prevent using winit for a UIKit backed app if I understand correctly. That being said, I'm not a huge fan of duplicating the lines to the [draw_rect](https://github.com/rust-windowing/winit/blob/400f75a2b33fc765a3173992cccc7d94defafde5/src/platform_impl/ios/view.rs#L105-L108) logic but   eh.

simlay: In any case, I've added the `RedrawRequested` and `RedrawEventsCleared` the `draw_rect` function as seen below. That said, I don't quite understand how `RedrawEventsCleared` and `RedrawRequested` would have been emitted in the wrong order rather than `RedrawEventsCleared` not being emitted previously. I'll look into testing this.

aleksijuvani: > That said, I don't quite understand how `RedrawEventsCleared` and `RedrawRequested` would have been emitted in the wrong order rather than `RedrawEventsCleared` not being emitted previously. I'll look into testing this.

My understanding is that there would be no queued `RedrawRequested` events so the iterator is empty—the empty iterator is then chained with a `std::iter::once(RedrawEventsCleared)`, so a `RedrawEventsCleared` is always emitted in that code path. The actual `RedrawEvents` is emitted from the other code path at a different time, hence the events appearing out of order.

> I'm a little apprehensive to remove the `draw_rect` codepath because it would prevent using winit for a UIKit backed app if I understand correctly.

In practice this would mean that you would no longer receive `RedrawRequested` events from the system and would have to issue them yourself. I'm not sure if you were receiving these events from the system in the first place however. The rules for when `drawRect` gets called are a bit complicated and depends on how things were set up w.r.t. layers and such.

aleksijuvani: As for testing this, I believe you can invoke the `drawRect` code path with gfx-hal examples. If the Metal layer is a sublayer of the winit created layer (which is what gfx-hal does on iOS), you'll be able to call `request_redraw` and invoke the `drawRect` code path.

Taking a quick glance at the new changes—I think this would result in `RedrawEventsCleared` being emitted, followed by `RedrawRequested` events, and then another `RedrawEventsCleared`. We would essentially be receiving duplicate `RedrawEventsCleared` events because both code paths emit it (provided that `draw_rect` is invoked that is).

simlay: > Taking a quick glance at the new changes—I think this would result in RedrawEventsCleared being emitted, followed by RedrawRequested events, and then another RedrawEventsCleared. We would essentially be receiving duplicate RedrawEventsCleared events because both code paths emit it (provided that draw_rect is invoked that is).

I was afraid that would be the case. I think it was worse than that actually. The chain operator added `RedrawEventsCleared` to every call for `handle_main_events_cleared` even without a `RedrawRequested` event the `.main_events_cleared_transition` function returned nothing.

Anyway, I've updated that issue and tested the Metal backend codepath using the gluten iOS example.

For the UIKit backend, I tested out the events using a some UIKit bindings here https://github.com/simlay/rust-uikit/blob/add-winit/ios-example/rust/src/lib.rs#L75. I found a few issues with some of the logs warning with something similar to the following:
```
processing non `RedrawRequested` event after the main event loop: RedrawEventsCleared
```
My guess as to why this is happening is because of the objective-c runtime is asynchronous so, I'm guessing this is suppose to be the behavior. Aside from that, the events came in order and weren't duplicated.

PR #1301, opened 2019-12-02
Implement revamped `RedrawRequested` on Web

hecrj:
Related to #1082.

I figured I would give this a shot and it was easier than I expected. Once this and #1299 are merged, I think #1082 will be ready to go!

I have also added a `web` example to test things out. It seems to work fine.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -117,3 +117,6 @@ package = "stdweb"
 version = "=0.4.20"
 optional = true
 features = ["experimental_features_which_may_break_on_minor_version_bumps"]
+
+[target.'cfg(target_arch = "wasm32")'.dev-dependencies]
+console_log = "0.1"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -29,6 +29,7 @@ raw-window-handle = "0.3"
 [dev-dependencies]
 image = "0.21"
 env_logger = "0.5"
+console_log = "0.1"
```
simlay: I think it might be good to do to `target.'cfg(target_arch = "wasm32")'.dev-dependencies]` and maybe even with the `sys-web` feature flag but I don't feel too strongly about it.

hecrj: IIRC, it's not possible to check enabled features in the Cargo manifest.

In any case, I have changed to only enable it on Wasm.

PR #1306, opened 2019-12-04
Change ModifiersState to a bitflags struct

Osspial:

This addresses a pet peeve I've had for a while. The `ModifiersState` struct is logically a series of bitflags, but we've been exposing it as a series of `bool`s. This PR addresses that.

If people have objections, please bring them up, but in the absence of discussion I'm inclined to merge this.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -700,16 +700,24 @@ bitflags! {
     /// Each flag represents a modifier and is set if this modifier is active.
     #[derive(Default)]
     pub struct ModifiersState: u32 {
-        // We're using every other bit here so that, when we distinguish between left and right
-        // modifiers, the left and right bits can be adjacent to eachother.
+        // left and right modifiers are currently commented out, but we should be able to support
+        // them in a future release
         /// The "shift" key.
-        const SHIFT = 0b01 << 0;
+        const SHIFT = 0b100 << 0;
+        // const LSHIFT = 0b010 << 0;
+        // const RSHIFT = 0b001 << 0;
         /// The "control" key.
-        const CTRL  = 0b01 << 2;
+        const CTRL = 0b100 << 3;
+        // const LCTRL = 0b010 << 3;
+        // const RCTRL = 0b001 << 3;
         /// The "alt" key.
-        const ALT   = 0b01 << 4;
+        const ALT = 0b100 << 6;
+        // const LALT = 0b010 << 6;
+        // const RALT = 0b001 << 6;
         /// This is the "windows" key on PC and "command" key on Mac.
-        const LOGO  = 0b01 << 6;
+        const LOGO = 0b100 << 9;
+        // const LLOGO = 0b010 << 9;
+        // const RLOGO = 0b001 << 9;
     }
 }
 
--- a/src/event.rs
+++ b/src/event.rs
@@ -675,21 +675,66 @@ pub enum VirtualKeyCode {
     Cut,
 }
 
-/// Represents the current state of the keyboard modifiers
-///
-/// Each field of this struct represents a modifier and is `true` if this modifier is active.
-#[derive(Default, Debug, Hash, PartialEq, Eq, Clone, Copy)]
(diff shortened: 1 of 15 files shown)


src/platform_impl/windows/event.rs
```diff
@@ -17,13 +17,22 @@ fn key_pressed(vkey: c_int) -> bool {
...
+        ModifiersState::CTRL,
+        key_pressed(winuser::VK_CONTROL) && !filter_out_altgr,
+    );
+    mods.set(
+        ModifiersState::ALT,
+        key_pressed(winuser::VK_MENU) && !filter_out_altgr,
+    );
+    mods.set(
+        ModifiersState::LOGO,
+        key_pressed(winuser::VK_LWIN) || key_pressed(winuser::VK_RWIN),
```
goddessfreya: And this code stores both the left and right win in the left bit, when I'd expect them to be stored in the two separate bits. The code above doesn't use things like `VK_LCONTROL`, `VK_RMENU`, ect... instead using the generic keycodes.

src/platform_impl/linux/x11/util/input.rs
```diff
@@ -17,12 +17,12 @@ impl ModifiersState {
     }
 
     pub(crate) fn from_x11_mask(mask: c_uint) -> Self {
-        ModifiersState {
-            alt: mask & ffi::Mod1Mask != 0,
-            shift: mask & ffi::ShiftMask != 0,
-            ctrl: mask & ffi::ControlMask != 0,
-            logo: mask & ffi::Mod4Mask != 0,
-        }
+        let mut m = ModifiersState::empty();
```
goddessfreya: And how would you distinguish between the left and right keys on x11?

src/event.rs
```diff
@@ -645,21 +645,91 @@ pub enum VirtualKeyCode {
...
+
+bitflags! {
+    /// Represents the current state of the keyboard modifiers
     ///
-    /// This is the "windows" key on PC and "command" key on Mac.
-    pub logo: bool,
+    /// Each flag represents a modifier and is set if this modifier is active.
+    #[derive(Default)]
+    pub struct ModifiersState: u32 {
+        // We're using every other bit here so that, when we distinguish between left and right
```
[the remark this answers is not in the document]
Osspial: It'd be more accurate to say "when we eventually when we distinguish...". Left shift and right shift aren't exposed separately right now.

src/event.rs
```diff
@@ -645,21 +645,91 @@ pub enum VirtualKeyCode {
...
-    pub shift: bool,
-    /// The "control" key
-    pub ctrl: bool,
-    /// The "alt" key
-    pub alt: bool,
-    /// The "logo" key
+impl ModifiersState {
+    /// Returns `true` if the shift key is pressed.
+    pub fn shift(&self) -> bool {
+        self.intersects(Self::SHIFT)
```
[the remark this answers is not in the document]
Osspial: We're currently only filling in one of those bits.

src/platform_impl/linux/x11/util/input.rs
murarth: X11 can check against hard-coded keycode values for left and right Ctrl, Shift, etc. However, it would be helpful if the bit mask had three bits for each modifier; e.g. `0b001` => ctrl modifier set, `0b010` => left ctrl, `0b100` => right ctrl. This way, X11 code could express that a modifier is set even when the modifier key does not correspond to a known keycode. Documentation would also need to reflect that `mods.ctrl()` may be `true` when both `mods.left_ctrl()` and `mods.right_ctrl()` are `false`.

src/platform_impl/windows/event.rs
Osspial: The `LOGO` bit in the design is side-agnostic.

src/platform_impl/linux/x11/util/input.rs
Osspial: I've updated the patterns to use three bits. This PR should be ready to review again.

Issue #1272, opened 2019-11-10
WindowEvent does not register Key release if looses focus.


goddessfreya:
As mentioned in https://github.com/rust-windowing/winit/pull/1262#pullrequestreview-314593736, however over the window, hold any key, move off the window release the key. You won't get the released key event. Ever.

https://gist.github.com/goddessfreya/39ae1f768cde5bdbc8c9aefa25984824
https://gist.github.com/goddessfreya/0192aa8d9920dba7c933c98190c42d90

cc @murarth

Osspial:
This is a problem on Windows as well. I wouldn't be surprised if we had this issue everywhere.

IMO the simplest solution is to automatically send a `KeyUp` event for all pressed keys whenever the active window loses focus.

murarth:
Is this really an issue winit needs to solve? I feel like applications should be expected to handle this, if appropriate, by responding to `Focused` events. If we start generating artificial key release events, there's no way to opt out of it.

OvermindDL1:
It could be opt in perhaps, by setting a flag or so, or opt-out?

Rua:
Automatically sending `KeyUp` events just shifts the problem back to `KeyDown` events. What if the window loses focus, the user releases the key, and then presses it again before refocusing the window? Then you have a `KeyUp` event without a preceding `KeyDown` event.

goddessfreya:
@Rua IMO, when a window loses focus, we should send a `KeyUp` for all keys currently held down. When the window regains focus, all keys currently held down should send a `KeyDown`. I'm not sure how feasible this is on other platforms, but it should be achievable with X11. This should make it so that every `KeyUp` has a matching `KeyDown` and vice versa.

@murarth If we are expecting users to listen to the `Focused` events and track the keys via `DeviceEvent::Key`, then what use is `WidnowEvent::KeyboardInput`?

murarth:
@goddessfreya I'm not suggesting that application developers should use `DeviceEvent::Key` for all key press responses. I just think that, if an application's behavior is sensitive to focus loss while a key is held, it may be better handled by the application than by winit generating artificial events. I just feel that, if we go down the road of generating artificial events, we may encounter unforeseen complications.

On the other hand, I can't really think of a use case where an application would want the current behavior. And if there's only one "correct" way for applications to workaround this issue, it would make sense for winit to exhibit that behavior itself.

Anyway, if the consensus is in favor of fixing this, I concede the point and will implement this for X11.

[...]

PR #1307, opened 2019-12-04
 Implement windows focus key press/release on Windows 

Osspial:

Implements #1272 on Windows.


src/event.rs
```diff
@@ -132,6 +132,15 @@ pub enum WindowEvent {
     KeyboardInput {
         device_id: DeviceId,
         input: KeyboardInput,
+        /// If `true`, the event was generated synthetically by winit
+        /// in one of the following circumstances:
+        ///
+        /// * **X11**: Synthetic key press events are generated for all keys pressed
```
goddessfreya: @Osspial Probably want to update the docs.

PR #1318, opened 2019-12-14
macOS: Unbundled window activation hack

francesca64:

The explanation for this PR is actually in a wall of comments, so check the diff for context. I have a tendency to view winit as a Rosetta stone of windowing, hence the thorough commenting...

I scanned through the issue tracker, and it doesn't look like there's an open issue for this. You'd think someone from Alacritty would've reported this already, but since Alacritty comes so nicely bundled, only someone developing Alacritty would be able to hit this. I also suspect that there aren't many people developing with winit on macOS in general.

As an exciting bonus, I removed the commented-out cruft I left in `app_delegate.rs` last year. It was just copy-pasted from the iOS backend, which I was using as a reference. It was all intended to be removed, but I forgot to ever clean it up.

Solution derived from this Godot PR: godotengine/godot#17187
Equivalent GLFW issue: glfw/glfw#1515


francesca64:
@aleksijuvani this PR exists specifically because that doesn't work, though you're free to try yourself.

francesca64:
@vbogaevsky would you be able to take a look at this?

vbogaevsky:
@francesca64, I'll review and test it this night.

francesca64:
@vbogaevsky that sounds like the opposite of what I'd expect; what you observed on `macos-unbundled` is what's expected on `master`, and what you observed on `master` is exactly what's expected on `macos-unbundled`. Can you verify that the branches weren't swapped? 

(Enabling info logging can help verify what's happening as well, though that would require plugging in `env_logger`...)

francesca64:
@vbogaevsky in which case? I'm having trouble following you, so being more explicit about context would help me a lot.

vbogaevsky:
@francesca64 in both cases (master and PR branch).
On PR branch the app is active i.e. `msg_send![NSApp(), isActive]` and `State::get_activated(this)` return `YES` and `true` respectively.
But `` in OS dropdown menu is inaccessible. To make it accessible you have to switch to another app and then back.

francesca64:
@vbogaevsky ah, thanks. I'm able to repro that, but it doesn't seem to be a regression?

How acceptable is this PR?

vbogaevsky:
@francesca64 It's not a regression indeed, and I believe we can merge after resolving conflicts in `src/platform_impl/macos/app.rs`

francesca64:
@vbogaevsky thanks!

(I didn't notice you'd already solved the conflicts, so double thanks! Now I feel silly for having made those merge commits.)

PR #1323, opened 2019-12-18
FIX Crash on macOS when starting maximized without decorations

### LINKED ISSUE #1288: Crash on macOS when starting maximized without decorations
On the latest release of macOS, when running the winit example after adding `.with_maximized(true).with_decorations(false)`, it will crash with the error `illegal hardware instruction`.

This has been reported to Alacritty in https://github.com/jwilm/alacritty/issues/2877.

hatoo:

This PR fixes #1288 
It was crashed in `set_style_mask_sync`.
As far as I investigated, `dispatch_sync_f` must not be called in the main thread, so I added `if` clause to detect the main thread and call callback directly then.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- on macOS, fix crash when starting maximized without decorations
+
 # # 0.20.0 Alpha 5 (2019-12-09)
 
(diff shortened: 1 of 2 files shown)


goddessfreya:
Thanks @aleksijuvani. Will merge when CI completes.

PR #1324, opened 2019-12-19
On Wayland, under mutter(GNOME Wayland), fix CSD being behind the status bar, when starting window in maximized mode

### LINKED ISSUE #1312: Gnome Wayland decorations behind system bar
It seems like somewhere between the version glutin master currently uses (0.20.0-alpha4) and the latest master, the window decorations on Wayland with gnome have been broken.

Previously when starting a maximized window, the window bar would be visible, as it should be. However with the latest master, the window is 'offset', so the top of the screen is the top of the inner window (no decorations) and the decorations are behind the system bar.

This might be better explained with the screenshots from the original Alacritty issue: https://github.com/jwilm/alacritty/issues/3070

kchibisov:
GNOME places maximized window CSD behind the status bar, and in the same time it moves any window crossing the "bounding box"(screen edges + status bar's bottom edge on top) into it on the next resize. So the main problem is GNOME placing window in a wrong place initially.

That gif shows the behavior I'm talking about. You don't need maximize to repro it.

jump

The question why mutter places our window in a wrong place initially is still flying in the air...

cc @Will-W

Fixes #1312

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 - On Wayland, fix cursor icon updates on window borders when using CSD.
+- On Wayland, under mutter(GNOME Wayland), fix CSD being behind the status bar, when starting window in maximized mode.
 
 # # 0.20.0 Alpha 5 (2019-12-09)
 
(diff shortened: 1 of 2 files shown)


Will-W:
This PR shows the correct behaviour for me.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Unreleased
 
 - On Wayland, fix cursor icon updates on window borders when using CSD.
+- On Wayland, under mutter(GNOME Wayland), fix CSD being behind the status bar, when stating window in maximized mode.
```
vberger: Is this a typo for "starting" (missing "r") ?

kchibisov: Yes.

kchibisov: I had to force push, because I have to alter commit message(the same typo).

Issue #1195, opened 2019-09-30
Fullscreen Titlebar Won't Show on macOS


nixpulvis:
While in an exclusive fullscreen mode on macOS, the titlebar (while hovering over the top of the screen) doesn't slide down as expected. I've confirmed this issue with `cargo run --example window` from this repo. This issue seems to have been introduced in 5bc3cf18d98aef0a5f0a51c0cf655d37af833730 with the merge of #925.

This was originally brought to our attention in https://github.com/jwilm/alacritty/pull/2438#issuecomment-536649468.

aleksijuvani:
This is because we currently set the `NSApplicationPresentationOptions` as such that the menu bar is hidden in exclusive fullscreen. See [`window:willUseFullScreenPresentationOptions:`](https://github.com/rust-windowing/winit/blob/18a0119b06e5922d4a4759b5167b413fdd29179f/src/platform_impl/macos/window_delegate.rs#L412-L430). We should check if the window is in borderless fullscreen mode and not set these flags in that case. Additionally, a way should be added to set `NSApplicationPresentationOptions` to user-specified options when the window is in borderless fullscreen.

Osspial:
@vbogaevsky could you look into resolving this?

nixpulvis:
I've taken a short look at this, and I'm by no means a real macOS developer but...

Simply passing through the currently unused `_proposed_options` as the return value of [`window:willUseFullScreenPresentationOption`](https://github.com/rust-windowing/winit/blob/18a0119b06e5922d4a4759b5167b413fdd29179f/src/platform_impl/macos/window_delegate.rs#L412-L430) seems to give us the behavior we want in Alacritty. Though it seems we don't even try to allow a user in a non exclusive fullscreen mode to even attempt to transition into an exclusive fullscreen mode (thus sidestepping the whole `CGShieldingWindowLevel() + 1` hack).

I'd be curious to know what a macOS user/developer has to say about this, since the whole way fullscreen works on macOS seems a bit crazy to me, it's hard for me to know how I'd expect this all to work.

I think I can say however that for exclusive fullscreen (not necessarily the same for non exclusive) it's not acceptable to always hide the menubar. This is partially supported by the fact that the green button in the traffic light UI typically means "toggle exclusive fullscreen", and a user who clicks that button to get into this mode, will be helpless to exit it.

aleksijuvani:
[...]

PR #1326, opened 2019-12-19
In MacOS, only disable menu bar in exclusive fullscreen mode

cixel:
My attempt at a quick patch for #1195, which causes jwilm/alacritty#3023. It seems to fix the issue in Alacritty.

@aleksijuvani  mentioned wanting `proposed_options` to be be replaced with user configured options via something like `window.set_fullscreen_presentation_options`. I'm happy to take a crack at this, but I'm inexperienced with Rust and figured a quick patch might suffice for now.

### DIFF
--- a/src/platform_impl/macos/window.rs
+++ b/src/platform_impl/macos/window.rs
@@ -263,7 +263,10 @@ pub struct SharedState {
     is_simple_fullscreen: bool,
     pub saved_style: Option<NSWindowStyleMask>,
     /// Presentation options saved before entering `set_simple_fullscreen`, and
-    /// restored upon exiting it
+    /// restored upon exiting it. Also used when transitioning from Borderless to
+    /// Exclusive fullscreen in `set_fullscreen` because we need to disable the menu
+    /// bar in exclusive fullscreen but want to restore the original options when
+    /// transitioning back to borderless fullscreen.
     save_presentation_opts: Option<NSApplicationPresentationOptions>,
     pub saved_desktop_display_mode: Option<(CGDisplay, CGDisplayMode)>,
 }
@@ -765,6 +768,14 @@ impl UnownedWindow {
             let mut fade_token = ffi::kCGDisplayFadeReservationInvalidToken;
 
             unsafe {
+                let app = NSApp();
+                trace!("Locked shared state in `set_fullscreen`");
+
+                let mut shared_state_lock = self.shared_state.lock().unwrap();
+                shared_state_lock.save_presentation_opts = Some(app.presentationOptions_());
+            }
+
+            unsafe {
                 // Fade to black (and wait for the fade to complete) to hide the
                 // flicker from capturing the display and switching display mode
                 if ffi::CGAcquireDisplayFadeReservation(5.0, &mut fade_token)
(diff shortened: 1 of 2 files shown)


cixel:
Ok, I spent a bit more time with this. Here's where I'm at.

In the transition from Borderless-->Exclusive, I used the `shared_state` lock to save `app.presentationOptions_()` the way `set_simple_fullscreen` does. Then, I disable the menu bar the same way `window:willUseFullscreenPresentationOptions` does. 

In the transition from Exclusive-->Borderless, I read `save_presentation_opts` from shared_state, and if they have been set, I give them to `app.setPresentationOptions_()`.

I'm still noticing a little bit of weirdness, but I'm not sure how to interpret it. Here's a [screen recording](https://gfycat.com/gregariousgiftedbaldeagle) to demonstrate.
In the video:
1. start --multithreaded example
1. Press F to enter borderless fullscreen. Notice the menu bar is hidden.
1. Press alt-F to enter exclusive fullscreen. This works as intended.
1. Press F to enter borderless fullscreen. This still has no menu bar.
1. Move to a different desktop and select a different window (in this case my terminal)
1. Move back to the fullscreen window. Notice that the menu bar now appears.
1. Press alt-F to enter exclusive fullscreen. This is still hiding the menu bar as expected.
1. Press F to restore borderless fullscreen. The menu bar still appears as it did before (before your proposed change it did not).

I'm not sure why I have to go through the ceremony of switching desktops and selecting another window to get the menu bar to appear in the first place, but once it appears, it seems to be saved and restored correctly when transitioning between Borderless<-->Exclusive modes.

Any thoughts? Also, should I be loading/retreiving `shared_state.save_presentation_opts` with _every_ transition to Exclusive, not just ones between Borderless and Exclusive?

Osspial:
@vbogaevsky could you review this?

felipesere:
I've tested this on my Mac with the merge-conflict resolved and it worked like a charm.
Is there anything holiding this up? I'd love to see it land and then used in Alacritty.

ArturKovacs:
I assume one of you fixed the problem that the menubar was hidden when first entering borderless fullscreen, because I could not reproduce it.

Please merge with the master first because there are a few issues that show up otherwise.

After merging with master I do get one strange behaviour. Running `cargo run --example multithreaded`:

1. Press <kbd>esc</kbd> twice to close the first two windows
2. Press <kbd>alt</kbd>+<kbd>F</kbd>. The window goes into exclusive fullscreen
3. Press <kbd>F</kbd>. The window seems to change to borderless fullscreen which is indicated by the fact that I can use <kbd>cmd</kbd>+<kbd>tab</kbd> to switch between applications. But now the menu bar isn't shown.

Let me know if you need help investigating this.

felipesere:
If I go from <kbd>alt</kbd>+<kbd>F</kbd> back to borderless fullscreen then the menu is blacked out.
If I go directly to borderless fullscreen then its fine and can get to the menu.

This feels odd...

madsmtm:
I get the same reproduction as @ArturKovacs on MacOS 10.14.6 Mojave, with the latest `master` - this seems like it can be fixed, but damn is it hard to keep track of all these state changes!

irh:
> If I go from alt+F back to borderless fullscreen then the menu is blacked out.
If I go directly to borderless fullscreen then its fine and can get to the menu.

I think I have a fix for this, see https://github.com/irh/winit/commit/99d3054c05945291a66d236091cb12654dc843b1

Are there any other issues remaining? I'd be happy to spend some more time on helping to push this forward.

madsmtm:
Closing since this was merged in https://github.com/rust-windowing/winit/pull/2053, and remaining issues are tracked in https://github.com/rust-windowing/winit/issues/2068

PR #1328, opened 2019-12-21
Basic android event loop 2.0

dvc94ch:


katyo:
@dvc94ch Curiously, but sometimes instant is less than start, so panic occurs here.
https://github.com/dvc94ch/winit/blob/a6ad0f75669b60f4752cf3b665eafa5098866da0/src/platform_impl/android/mod.rs#L217-L218
In particular it occurs after suspending the app. It seems android's scheduler throttles the app's time.
When I changed this piece to:
```rust
let duration = if instant <= start { Duration::default() } else { instant - start };
first_event = convert(looper.poll_all_timeout(duration).unwrap());
```
the demo runs fine.

katyo:
@dvc94ch It still has troubles with orientation change. Sometimes window size given in resize event is unusual (like 720x720) and several resize events happened. It seems we need workaround like delaying resize event from the moment when it actually happened.

dvc94ch:
@katyo I'm having the same problem, it's driving me nuts :)

dvc94ch:
@katyo I think I've fixed it, can you check if it works for you now? On a ConfigChange we rerender 10x in a 10ms interval and that works on my phone.

katyo:
@dvc94ch I tested on three different devices on from api level 16 to 28. Orientation change works fine everywhere!

dvc94ch:
I think this will probably need changes in the future, but it's enough of an improvement on the status quo that it should be merged before it gets stale.

katyo:
@dvc94ch Unfortunately it still work too buggy. In particular after recent commit.
Sometime it has missing first redraw. Also the problem with orientation change appeared again.
I will try find some time in a new year weekend to play around it.

Osspial:
Going forward, who's going to be responsible for maintaining the Android backend? There aren't any listed maintainers in [the wiki document](https://github.com/rust-windowing/winit/wiki/Testers-and-Contributors), and having a backend without active developers working on it makes it difficult for Winit's API to evolve.

dvc94ch:
It's my intention of maintaining it, I'm just currently distracted with https://github.com/flutter-rs/cargo-flutter and will get back to this in a couple of days.

katyo:
Also I in a process of implementing some missing features related to user interaction.
But it little bit hard because I need JNI for some things and I have really strange problems with it (SIGABRT).

dvc94ch:
@katyo did you resolve the SIGABRT issue? I'm having the same problem when trying to call `jni.AttachCurrentThread`. I'm begining to think that the android-ndk + native-app-glue is a halfassed piece of shit and that we might need a java entrypoint.

katyo:
@dvc94ch: I don't sure, but, it seems, SIGABRT gone away after I fixed pointer to activity object.

dvc94ch:
Closing in favor of https://github.com/goddessfreya/winit/pull/2

PR #1330, opened 2019-12-21
Reimplement NativeDisplayMode on iOS for #1310

### LINKED ISSUE #1310: multithreaded doesn't compile on iOS
Discovered in #1309. Attempting to run the example gives the following error:

```
   --> examples/multithreaded.rs:30:9
    |
30  |         thread::spawn(move || {
    |         ^^^^^^^^^^^^^ `*mut objc::runtime::Object` cannot be sent between threads safely
    |
    = help: within `winit::monitor::VideoMode`, the trait `std::marker::Send` is not implemented for `*mut objc::runtime::Object`
    = note: required because it appears within the type `winit::platform_impl::platform::monitor::VideoMode`
    = note: required because it appears within the type `winit::monitor::VideoMode`
    = note: required because of the requirements on the impl of `std::marker::Send` for `std::ptr::Unique<winit::monitor::VideoMode>`
    = note: required because it appears within the type `alloc::raw_vec::RawVec<winit::monitor::VideoMode>`
    = note: required because it appears within the type `std::vec::Vec<winit::monitor::VideoMode>`
    = note: required because it appears within the type `[closure@examples/multithreaded.rs:30:23: 140:10 rx:std::sync::mpsc::Receiver<winit::event::WindowEvent>, video_modes:std::vec::Vec<winit::monitor::VideoMode>, video_mode_id:usize, window:winit::window::Window]`
```

simlay:
Closes #1310.

I pretty much did exactly as https://github.com/rust-windowing/winit/issues/1310#issuecomment-561571449 said. I'm not big on the name NativeDisplayMode nor that it's nearly the exact same code from the macOS logic but I figure it's an okay first pass.

### DIFF
--- a/src/platform_impl/ios/monitor.rs
+++ b/src/platform_impl/ios/monitor.rs
@@ -18,8 +18,30 @@ pub struct VideoMode {
     pub(crate) size: (u32, u32),
     pub(crate) bit_depth: u16,
     pub(crate) refresh_rate: u16,
-    pub(crate) screen_mode: id,
+    pub(crate) screen_mode: NativeDisplayMode,
     pub(crate) monitor: MonitorHandle,
+}
+
+#[derive(Debug, PartialEq, Eq, Hash)]
+pub struct NativeDisplayMode(pub id);
+
+unsafe impl Send for NativeDisplayMode {}
+
+impl Drop for NativeDisplayMode {
+    fn drop(&mut self) {
+        unsafe {
+            let () = msg_send![self.0, release];
+        }
+    }
+}
+
+impl Clone for NativeDisplayMode {
+    fn clone(&self) -> Self {
+        unsafe {
+            let () = msg_send![self.0, retain];
+        }
+        NativeDisplayMode(self.0)
+    }
 }
 
 impl Clone for VideoMode {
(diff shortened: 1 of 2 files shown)


src/platform_impl/ios/monitor.rs
```diff
@@ -18,31 +18,44 @@ pub struct VideoMode {
...
+        unsafe {
+            let () = msg_send![self.0, release];
+        }
+    }
+}
+
+impl Clone for NativeDisplayMode {
+    fn clone(&self) -> Self {
+        unsafe {
+            let () = msg_send![self.0, retain];
```
aleksijuvani:
```suggestion
            let _: id = msg_send![self.0, retain];
```

src/platform_impl/ios/monitor.rs
```diff
@@ -18,31 +18,44 @@ pub struct VideoMode {
...
 
+#[derive(Debug, PartialEq, Eq, Hash)]
+pub struct NativeDisplayMode(pub id);
+
+unsafe impl Send for NativeDisplayMode {}
+
+impl Drop for NativeDisplayMode {
+    fn drop(&mut self) {
+        unsafe {
+            let () = msg_send![self.0, release];
```
goddessfreya: @aleksijuvani @simlay I was just merging master back into my `winit-glutin-next` and noticed that the `assert_main_thread` has been removed. Is this intentional?

aleksijuvani: I think it's fine to pass this around multiple threads. If we see a crash here we'll revisit.

PR #1335, opened 2019-12-24
MacOS fix `CursorEntered` and `CursorLeft` events fired at old window size.

hatoo:
- [x] Tested on all platforms changed
- [x] Compilation warnings were addressed
- [x] `cargo fmt` has been run on this branch
- [x] `cargo doc` builds successfully
- [x] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users
- [ ] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior
- [ ] Created or updated an example program if it would help users understand this functionality
- [ ] Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented

This PR fixes #1128 
`CursorEntered` and `CursorLeft` events fired at old window size because `TrackingRect` is not updated once initialized.
This adds an event for `NSViewFrameDidChangeNotification` to update `TrackingRect` for each window resize.

PR #1338, opened 2019-12-26
Relase alpha 6

Osspial:


goddessfreya:
As discussed on discord, hoping #1335 lands before the release.

chrisduerr:
Unless it holds it back forever, https://github.com/rust-windowing/winit/pull/1344 might also be interesting. That way the feature could be documented and "released" for desktop platforms (which is likely enough for most people) and existing users like Alacritty might benefit from the bugfixes?

The PR itself doesn't seem too controversial, so it might be worthwhile to consider.

Osspial:
@chrisduerr That's a good call. I'll update that PR with the proper documentation.

Osspial:
Closing in favor of #1352.

kchibisov:
I am wondering whether it's a good idea to have some kind of stable winit version before `dpi-overhaul` gets merged, since dpi-overhaul is a massive change and having some kind of stable starting point (alpha6) before testing it downstream(e.g. in alacritty) will be a good idea imho.

I'm saying this, because in alacritty we were going to make release right after alpha6 and stick to some kind of stable winit releases afterwards, but now we're likely cut a new version with winit's git (we really don't want to do that, since it breaks bisecting).

I'm not saying that winit should do whatever popular downstream is doing right now and help them. I'm saying that dpi-overhaul wasn't likely closely tested in downstream clients at all and updating to it should take some time (even merging it will take some time) and stall downstream users' releases...

Osspial:
@kchibisov I'd be fine with releasing `alpha6` without the DPI overhaul so that you have a release to pin to. However, the DPI overhaul is a *major* usability improvement for Winit that's been widely requested and worked on for quite a long time, and I don't want it to bitrot on the backburner when it's ready for release.

kchibisov:
>I'd be fine with releasing alpha6 without the DPI overhaul so that you have a release to pin to.

Thanks, this will help us a lot with our winit "handling" and we can finally move to official winit releases instead of git master and forked glutin.

chrisduerr:
Just to clarify, I'm all for making improvements available as quickly as possible, but I'd assume it will still take some time until DPI overhaul is merged and fully working.

So since there are a lot of bugfixes since the last release already, which are known to work pretty well, I think it makes sense to get those out to the user as quickly as possible so DPI overhaul can then be tested and ironed out without any pressure.

PR #1342, opened 2019-12-28
[MacOS] Fix memory management

hatoo:

I'm completely newbie about Cocoa's memory management.
Please review this carefully.

This PR contains 2 changes.
1. Fix memory leak on `run_return`
I noticed that `leaks` command shows some leaks of `NSAutoreleasePool` object when run `window_run_return` example.
I fixed it by inserting `pool.drain();` to end.
src/platform_impl/macos/event_loop.rs
2. Reduce memory usage on  `run_return`
I noticed that memory usage of `window_run_return` example is very higher than `window` example.
I fixed it by using `NSAutoreleasePool`.
src/platform_impl/macos/app_state.rs

### DIFF
--- a/src/platform_impl/macos/app_state.rs
+++ b/src/platform_impl/macos/app_state.rs
@@ -11,7 +11,11 @@ use std::{
     time::Instant,
 };
 
-use cocoa::{appkit::NSApp, base::nil, foundation::NSString};
+use cocoa::{
+    appkit::NSApp,
+    base::nil,
+    foundation::{NSAutoreleasePool, NSString},
+};
 
 use crate::{
     event::{Event, StartCause, WindowEvent},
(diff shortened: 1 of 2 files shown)


chrisduerr:
Just out of curiosity, is there a reason why this isn't part of the `NSAutoreleasePool`'s drop by default?

And does it make sense to implement a wrapper type in winit that just calls `drain` in the destructor.

goddessfreya:
@chrisduerr Apparently there is more than one way to dispose of an `NSAutoreleasePool`, `drain` and `release`, so I imagine they didn't want to call any specific disposal method in case users want to use the other one:

> You create an NSAutoreleasePool object with the usual alloc and init messages and dispose of it with drain (or release—to understand the difference, see Garbage Collection). [...]

> #### Garbage Collection
> In a garbage-collected environment, there is no need for autorelease pools. You may, however, write a framework that is designed to work in both a garbage-collected and reference-counted environment. In this case, you can use autorelease pools to hint to the collector that collection may be appropriate. In a garbage-collected environment, sending a drain message to a pool triggers garbage collection if necessary; release, however, is a no-op. In a reference-counted environment, drain has the same effect as release. Typically, therefore, you should use drain instead of release.

PR #1346, opened 2019-12-30
Remove Option from HiDpiFactorChanged in favor of a bare PhysicalSize

Osspial:

Thinking about it, the `Option` wrapper isn't really necessary - if the user doesn't want the window to resize, they can just set the size to `Window::inner_size()`.

Not 100% sure if this'll compile on macOS and iOS, but I'll update the PR if the CI fails.


goddessfreya:
> Not 100% sure if this'll compile on macOS and iOS, but I'll update the PR if the CI fails.

Fyi, you can check macOS builds with `cargo check --target x86_64-apple-darwin`. Still to figure out a method for ios.

PR #1348, opened 2020-01-01
On Macos, Hide cursor only inside window

### LINKED ISSUE #1276: Make `set_cursor_visible` on macOS only hide cursor within the window?
I noticed only macOS hides the cursor even when the cursor leaves the window, is there anyway to only hide within the window?

In SDL `SDL_ShowCursor(false)` hides cursor only within the window by default. I'm not sure how they did it, but a quick browse through the source and I see they only use `[NSCursor hide]` /  `[NSCursor unhide]` when switching relative mouse mode, not for `SDL_ShowCursor`.

hatoo:

This PR fixes #1276 
As same as SDL does, This PR hides cursor by setting the invisible cursor instead of hiding cursor globally.


src/platform_impl/macos/util/cursor.rs
```diff
@@ -126,3 +127,34 @@ pub unsafe fn load_webkit_cursor(cursor_name: &str) -> id {
...
     ]
 }
+
+pub unsafe fn invisible_cursor() -> id {
+    static CURSOR_BYTES: &[u8] = &[
+        0x47, 0x49, 0x46, 0x38, 0x37, 0x61, 0x10, 0x00, 0x10, 0x00, 0x80, 0x00, 0x00, 0x00, 0x00,
+        0x00, 0x00, 0x00, 0x00, 0x21, 0xF9, 0x04, 0x01, 0x00, 0x00, 0x01, 0x00, 0x2C, 0x00, 0x00,
+        0x00, 0x00, 0x10, 0x00, 0x10, 0x00, 0x00, 0x02, 0x0E, 0x8C, 0x8F, 0xA9, 0xCB, 0xED, 0x0F,
+        0xA3, 0x9C, 0xB4, 0xDA, 0x8B, 0xB3, 0x3E, 0x05, 0x00, 0x3B,
+    ];
```
vbogaevsky: From support perspective it would be good to have a comment on what exactly `CURSOR_BYTES` are, and how we use them.

Issue #1141, opened 2019-08-31
`WINIT_HIDPI_FACTOR` does not work when `Xft.dpi` is set


Toqozz:
This is a bit of a weird one.  I **think** it's tied to `Xft.dpi` but I'm not positive.

On X11, the `WINIT_HIDPI_FACTOR` environment variable seems to be ignored if `Xft.dpi` is set in the user's `.Xresources` file:

Xresources:
```
Xft.dpi: 80
```
Program:
```rust
std::env::set_var("WINIT_HIDPI_FACTOR", "1.0");

let event_loop = EventLoop::new();

let window = WindowBuilder::new()
    .with_title("A fantastic window!")
    .build(&event_loop)
    .unwrap()

dbg!(window.hidpi_factor());
```
Output:
```
[src/main.rs:20] window.hidpi_factor() = 0.8333333333333334
```

This behaviour continues even if you remove `Xft.dpi: 80` from the `.Xresources` file.  The only way to get winit to respect `WINIT_HIDPI_FACTOR` again is to set an invalid value like `Xft.dpi: foo` -- perhaps the value is being cached somewhere?

My guess is that the hidpi factor gets overwritten by the Xresources config regardless of `WINIT_HIDPI_FACTOR` at some point.

`winit = "0.20.0-alpha3"`

Toqozz:
Indeed;
https://github.com/rust-windowing/winit/blob/5bc3cf18d98aef0a5f0a51c0cf655d37af833730/src/platform_impl/linux/x11/util/randr.rs#L109-L119

`WINIT_HIDPI_FACTOR` is accounted for in `calc_dpi_factor()`, but doesn't run if an xft dpi exists.

I'd be interested in contributing a fix for this.

We probably want to move checking for `WINIT_HIDPI_FACTOR` out of `calc_dpi_factor()`.  Maybe something like `get_dpi()` which first checks for `WINIT_HIDPI_FACTOR` and then proceeds as it does currently?

Osspial:
FWIW, I don't think this will be relevant once the #939 work is complete, since we're largely stopping doing the HiDPI factor transformation internally.

Toqozz:
Great!

rofrol:
If https://github.com/rust-windowing/winit/issues/939 is close, can this also be closed?

murarth:
Yes. Closed by #1354.

Issue #836, opened 2019-04-09
Allow ignoring Xft.dpi


chrisduerr:
Winit has recently gotten the ability to set a fixed DPI on all displays in X by using the Xft.dpi file, however this makes it impossible to have dynamic scaling in winit while manually setting the DPI for other applications, since as soon as the file is found all dynamic scaling is disabled.

This has been reported to Alacritty in https://github.com/jwilm/alacritty/issues/2271.

I see three ways to solve this issue:
 - Keeping things the way they are right now, user just needs to know and work around this
 - Introduce an env variable to force dynamic scaling and ignore Xft.dpi
 - Introduce an env variable which applies Xft.dpi on top of dynamic scaling

I'm unsure which of these would be the best option, though the first seems unfortunate. I think that the second would be my personal favourite but maybe we can even make this a Rust function and skip pointless environment variables?

lespea:
What if get_output_info took another parameter to force calculation of the dpi by using the width/height of the screen?

chrisduerr:
With the release of Alacritty 0.4.0, this has been brought up again and I think it's a very good point. I personally like winit's scaling too, so I would be annoyed if I had a monitor that required me to use Xft.dpi for other applications without being able to ignore it for Alacritty.

Fundamentally this should be trivial to fix, I can send a PR for this myself so that's no problem. The only question is about *how*.

I'm not sure what GTK's `gdk_monitor_get_scale_factor` does, but based on the behavior of GTK, I'd expect it to not be what we would want. Also having per-monitor scaling is not the issue, that's already working today. The issue is with being able to ignore `Xft.dpi` and even when using `gdk_monitor_get_scale_factor` we'd run either into the scenario where `Xft.dpi` is always ignored (which we do not want), or `Xft.dpi` is never ignored, which is described in this issue.

So I think fundamentally keeping this issue open to look into GDK's monitor DPI function will likely not bear any fruits and just delay an actual solution even further.

I understand that environment variables aren't great, however even the `WINIT_HPDPI_FACTOR` variable is already used by many because it's just something people need. So I think perfect is the enemy of good in this way, especially because perfect does not exist with X11 DPI.

I'd propose either making the decision to expose it in the API, or to offer an environment variable and then just go with it. I'm fine with either, however I don't think waiting will improve anything with X11.

kchibisov:
I would like to add some additional notes to what @chrisduerr already mentioned.

[...]

PR #1349, opened 2020-01-03
Update the DPI module docs

Osspial:

Significantly rewrites the DPI docs to better suit Winit's current state. Feedback is appreciated - both for content and spelling/grammar. Please bring up any understandability regressions compared to the previous docs.

### DIFF
--- a/src/dpi.rs
+++ b/src/dpi.rs
@@ -33,9 +33,9 @@
 //! approximate DPI (at the time, 72 pixels per inch), but [Microsoft decided to report that the DPI
 //! was roughly 1/3 bigger than the screen's actual DPI (so, 96 pixels per inch) in order to make
 //! text more legible][microsoft_dpi]. As a result, the exact DPI as defined by the OS doesn't carry
-//! a whole lot of weight when designing cross-platform UIs. However, if you're in a situation where
-//! having an exact DPI value matters it should be safe to use `96 * scale_factor` on desktops and
-//! the web and `160 * scale_factor` on mobile platforms.
+//! a whole lot of weight when designing cross-platform UIs. Scaled pixels should generally be used
+//! as the base unit for on-screen UI measurement, instead of DPI-dependent units such as
+//! [points][points] or [picas][picas].
 //!
 //! ### Position and Size types
 //!
@@ -91,6 +91,8 @@
 //! - **Web:** The scale factor is the ratio between CSS pixels and the physical device pixels.
 //!
 //! [microsoft_dpi]: https://blogs.msdn.microsoft.com/fontblog/2005/11/08/where-does-96-dpi-come-from-in-windows/
+//! [points]: https://en.wikipedia.org/wiki/Point_(typography)
+//! [picas]: https://en.wikipedia.org/wiki/Pica_(typography)
 //! [windows_1]: https://docs.microsoft.com/en-us/windows/win32/hidpi/high-dpi-desktop-application-development-on-windows
 //! [apple_1]: https://developer.apple.com/library/archive/documentation/DeviceInformation/Reference/iOSDeviceCompatibility/Displays/Displays.html
 //! [apple_2]: https://developer.apple.com/design/human-interface-guidelines/macos/icons-and-images/image-size-and-resolution/
--- a/src/dpi.rs
+++ b/src/dpi.rs
@@ -9,10 +9,10 @@
 //! mobile screens range from below 720p to above 1440p, and desktop screens range from 720p to 5K
 //! and beyond.
 //!
-//! Given that, it's naive for 2D content to assume that it'll only be displayed on screens with a
-//! consistent pixel density. If you were to render a 96-pixel-square image on a 1080p screen, then
-//! render the same image on a similarly-sized 4K screen, the 4K rendition would only take up about
-//! a quarter of the physical space as it did on the 1080p screen. That issue is especially
+//! Given that, it's a mistake for 2D content to assume that it'll only be displayed on screens with
+//! a consistent pixel density. If you were to render a 96-pixel-square image on a 1080p screen,
+//! then render the same image on a similarly-sized 4K screen, the 4K rendition would only take up
+//! about a quarter of the physical space as it did on the 1080p screen. That issue is especially
 //! problematic with text rendering, where quarter-sized text becomes a significant legibility
 //! problem.
 //!
@@ -28,8 +28,8 @@
 //! for example, a button that's normally 50 pixels across would be 100 pixels across on a device
 //! with a scale factor of `2.0`, or 75 pixels across with a scale factor of `1.5`.
 //!
-//! The scale factor correlates with, but no direct relationship to, the screen's actual DPI (dots
-//! per inch). Operating systems used to define the scale factor in terms of the screen's
+//! The scale factor correlates with, but no has direct relationship to, the screen's actual DPI
+//! (dots per inch). Operating systems used to define the scale factor in terms of the screen's
 //! approximate DPI (at the time, 72 pixels per inch), but [Microsoft decided to report that the DPI
 //! was roughly 1/3 bigger than the screen's actual DPI (so, 96 pixels per inch) in order to make
 //! text more legible][microsoft_dpi]. As a result, the exact DPI as defined by the OS doesn't carry
@@ -49,8 +49,8 @@
 //! API to have integer precision where appropriate (e.g. most window manipulation functions) and
 //! floating precision when necessary (e.g. logical sizes for fractional scale factors and touch
 //! input). If `P` is a floating-point type, please do not cast the values with `as {int}`. Doing so
-//! will truncate the fractional part of the float, rather than round to the nearest integer - the
-//! proper behavior. Use the provided `cast` function or `From`/`Into` conversions, which handle the
+//! will truncate the fractional part of the float, rather than properly round to the nearest
+//! integer. Use the provided `cast` function or `From`/`Into` conversions, which handle the
 //! rounding properly. Note that precision loss will still occur when rounding from a float to an
 //! int, although rounding lessens the problem.
 //!
--- a/src/dpi.rs
+++ b/src/dpi.rs
@@ -6,7 +6,7 @@
 //! 1920x1080 is a common resolution for both desktop and mobile screens, despite mobile screens
 //! normally being less than a quarter the size of their desktop counterparts. What's more, neither
 //! desktop nor mobile screens are consistent resolutions within their own size classes - common
-//! mobile screens range from below 720p to above 1440p, and desktop screens range from 1080p to 5K
+//! mobile screens range from below 720p to above 1440p, and desktop screens range from 720p to 5K
 //! and beyond.
 //!
 //! Given that, it's naive for 2D content to assume that it'll only be displayed on screens with a
--- a/src/dpi.rs
+++ b/src/dpi.rs
@@ -63,9 +63,11 @@
 //! - **macOS:** "retina displays" have a scale factor of 2.0. Otherwise, the scale factor is 1.0.
 //!   Intermediate scale factors are never used. It's possible for any display to use that 2.0 scale
 //!   factor, given the use of the command line.
-//! - **X11:** On X11, we calcuate the scale factor based on the millimeter dimensions provided by
-//!   XRandR. This is almost certainly the wrong approach, but many man-hours have been spent trying
-//!   to figure out how to do it better and X11 provides no easy answer.
+//! - **X11:** Many man-hours have been spent trying to figure out how to handle DPI in X11. Winit
+//!   currently uses a three-pronged approach.
+//!   + Calcuate the scale factor based on the millimeter monitor dimensions provided by XRandR.
+//!   + Use the value set in `Xft.dpi` in Xresources. Overrides XRandR's monitor size.
+//!   + Use the value in the `WINIT_X11_SCALE_FACTOR` environment variable. Overrides `Xft.dpi` and XRandR.
 //! - **Wayland:** On Wayland, scale factors are set per-screen by the server, and are always
 //!   integers (most often 1 or 2).
 //! - **iOS:** Scale factors are set by Apple to the value that best suits the device, and range
--- a/src/dpi.rs
+++ b/src/dpi.rs
@@ -1,80 +1,83 @@
-//! DPI is important, so read the docs for this module if you don't want to be confused.
+//! Pixel types and UI scaling information.
 //!
-//! Originally, `winit` dealt entirely in physical pixels (excluding unintentional inconsistencies), but now all
-//! window-related functions both produce and consume logical pixels. Monitor-related functions still use physical
-//! pixels, as do any context-related functions in `glutin`.
+//! ## Why should I care about UI scaling?
 //!
-//! If you've never heard of these terms before, then you're not alone, and this documentation will explain the
-//! concepts.
+//! Modern computer screens don't have a consistent relationship between resolution and size.
+//! 1920x1080 is a common resolution for both desktop and mobile screens, despite mobile screens
+//! normally being less than a quarter the size of their desktop counterparts. What's more, neither
+//! desktop nor mobile screens are consistent resolutions within their own size classes - common
+//! mobile screens range from below 720p to above 1440p, and desktop screens range from 1080p to 5K
+//! and beyond.
 //!
-//! Modern screens have a defined physical resolution, most commonly 1920x1080. Indepedent of that is the amount of
-//! space the screen occupies, which is to say, the height and width in millimeters. The relationship between these two
-//! measurements is the *pixel density*. Mobile screens require a high pixel density, as they're held close to the
-//! eyes. Larger displays also require a higher pixel density, hence the growing presence of 1440p and 4K displays.
+//! Given that, it's naive for 2D content to assume that it'll only be displayed on screens with a
+//! consistent pixel density. If you were to render a 96-pixel-square image on a 1080p screen, then
+//! render the same image on a similarly-sized 4K screen, the 4K rendition would only take up about
+//! a quarter of the physical space as it did on the 1080p screen. That issue is especially
+//! problematic with text rendering, where quarter-sized text becomes a significant legibility
+//! problem.
 //!
-//! So, this presents a problem. Let's say we want to render a square 100px button. It will occupy 100x100 of the
-//! screen's pixels, which in many cases, seems perfectly fine. However, because this size doesn't account for the
-//! screen's dimensions or pixel density, the button's size can vary quite a bit. On a 4K display, it would be unusably
-//! small.
+//! Failure to account for the scale factor can create a significantly degraded user experience.
+//! Most notably, it can make users feel like they have bad eyesight, which will potentially cause
+//! them to think about growing elderly, resulting in them having an existential crisis. Once users
+//! enter that state, they will no longer be focused on your application.
 //!
-//! That's a description of what happens when the button is 100x100 *physical* pixels. Instead, let's try using 100x100
-//! *logical* pixels. To map logical pixels to physical pixels, we simply multiply by the DPI (dots per inch) factor.
-//! On a "typical" desktop display, the scale factor will be 1.0, so 100x100 logical pixels equates to 100x100 physical
-//! pixels. However, a 1440p display may have a scale factor of 1.25, so the button is rendered as 125x125 physical pixels.
-//! Ideally, the button now has approximately the same perceived size across varying displays.
+//! ## How should I handle it?
 //!
-//! Failure to account for the scale factor can create a badly degraded user experience. Most notably, it can make users
-//! feel like they have bad eyesight, which will potentially cause them to think about growing elderly, resulting in
-//! them entering an existential panic. Once users enter that state, they will no longer be focused on your application.
+//! The solution to this problem is to account for the device's *scale factor*. The scale factor is
+//! the factor UI elements should be scaled by to be consistent with the rest of the user's system -
+//! for example, a button that's normally 50 pixels across would be 100 pixels across on a device
+//! with a scale factor of `2.0`, or 75 pixels across with a scale factor of `1.5`.
 //!
-//! There are two ways to get the scale factor:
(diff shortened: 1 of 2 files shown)


icefoxen:
https://github.com/rust-windowing/winit/blob/52020cbfbe40504e745385393f38c4918e14ce58/src/dpi.rs#L12-L17 -- Would demonstrate the problem better if you made the comparison between a 1080p computer screen vs 1080p phone screen, *especially* since you talk about text right thereafter which is a very vivid example of the problem.

https://github.com/rust-windowing/winit/blob/52020cbfbe40504e745385393f38c4918e14ce58/src/dpi.rs#L46-L51 -- Might be nice to add a brief example of what could cause this to happen -- a window being dragged from one screen to a different one for example.  Hard for someone not used to this to think of a case that causes it otherwise.

Looks very good besides that!  :heart:

src/dpi.rs
```diff
@@ -1,80 +1,83 @@
-//! DPI is important, so read the docs for this module if you don't want to be confused.
```
Lokathor: Quality sentence, please restore.

src/dpi.rs
```diff
@@ -1,80 +1,83 @@
...
+//! problem.
 //!
-//! So, this presents a problem. Let's say we want to render a square 100px button. It will occupy 100x100 of the
-//! screen's pixels, which in many cases, seems perfectly fine. However, because this size doesn't account for the
-//! screen's dimensions or pixel density, the button's size can vary quite a bit. On a 4K display, it would be unusably
-//! small.
+//! Failure to account for the scale factor can create a significantly degraded user experience.
+//! Most notably, it can make users feel like they have bad eyesight, which will potentially cause
+//! them to think about growing elderly, resulting in them having an existential crisis. Once users
+//! enter that state, they will no longer be focused on your application.
```
Lokathor: ", and they could even `panic!`."

src/dpi.rs
```diff
@@ -1,80 +1,83 @@
...
+//!   details.
+//! - **macOS:** "retina displays" have a scale factor of 2.0. Otherwise, the scale factor is 1.0.
+//!   Intermediate scale factors are never used. It's possible for any display to use that 2.0 scale
+//!   factor, given the use of the command line.
+//! - **X11:** On X11, we calcuate the scale factor based on the millimeter dimensions provided by
+//!   XRandR. This is almost certainly the wrong approach, but many man-hours have been spent trying
+//!   to figure out how to do it better and X11 provides no easy answer.
+//! - **Wayland:** On Wayland, scale factors are set per-screen by the server, and are always
+//!   integers (most often 1 or 2).
+//! - **iOS:** Scale factors are set by Apple to the value that best suits the device. See [this
```
Lokathor: This should also say the common values.

src/dpi.rs
```diff
@@ -1,80 +1,83 @@
...
-//! pixels, as do any context-related functions in `glutin`.
+//! ## Why should I care about UI scaling?
 //!
-//! If you've never heard of these terms before, then you're not alone, and this documentation will explain the
-//! concepts.
+//! Modern computer screens don't have a consistent relationship between resolution and size.
+//! 1920x1080 is a common resolution for both desktop and mobile screens, despite mobile screens
+//! normally being less than a quarter the size of their desktop counterparts. What's more, neither
+//! desktop nor mobile screens are consistent resolutions within their own size classes - common
+//! mobile screens range from below 720p to above 1440p, and desktop screens range from 1080p to 5K
```
Ralith: I'm not sure sub-1080p desktop/laptop screens are actually uncommon yet. For a random example, see the primacy of 1366x768 at https://gs.statcounter.com/screen-resolution-stats/desktop/worldwide.

src/dpi.rs
Osspial: Dry humor is no place for puns.

src/dpi.rs
Osspial: ok

src/dpi.rs
```diff
@@ -1,80 +1,83 @@
...
-//!   [`Window::hidpi_factor`](../window/struct.Window.html#method.hidpi_factor), which is roughly equivalent
-//!   to `window.current_monitor().hidpi_factor()`.
+//! The scale factor correlates with, but no direct relationship to, the screen's actual DPI (dots
+//! per inch). Operating systems used to define the scale factor in terms of the screen's
+//! approximate DPI (at the time, 72 pixels per inch), but [Microsoft decided to report that the DPI
+//! was roughly 1/3 bigger than the screen's actual DPI (so, 96 pixels per inch) in order to make
+//! text more legible][microsoft_dpi]. As a result, the exact DPI as defined by the OS doesn't carry
+//! a whole lot of weight when designing cross-platform UIs. However, if you're in a situation where
+//! having an exact DPI value matters it should be safe to use `96 * scale_factor` on desktops and
+//! the web and `160 * scale_factor` on mobile platforms.
```
Ralith: This is a bit misleading as it ignores the fact that scale factor should account for the device's expected viewing distance as well, completely independent of the actual DPI.

src/dpi.rs
```diff
@@ -1,80 +1,83 @@
...
-//! - **Wayland:** On Wayland, DPI factors are set per-screen by the server, and are always integers (most often 1 or 2).
-//! - **iOS:** DPI factors are both constant and device-specific on iOS.
-//! - **Android:** This feature isn't yet implemented on Android, so the DPI factor will always be returned as 1.0.
-//! - **Web:** DPI factors are handled by the browser and will always be 1.0 for your application.
+//! Winit will dispatch a [`HiDpiFactorChanged`](crate::event::WindowEvent::HiDpiFactorChanged)
+//! event whenever a window's scale factor has changed. This gives you a chance to rescale your
+//! application's UI elements and adjust how the platform changes the window's size to reflect the
+//! new scale factor. If a window hasn't received a
+//! [`HiDpiFactorChanged`](crate::event::WindowEvent::HiDpiFactorChanged) event, then its
+//! scale factor is `1.0`.
```
Ralith: Should the event be generated unconditionally during startup so this assumption doesn't need to be encoded in every downstream application?

src/dpi.rs
```diff
@@ -126,19 +129,19 @@ impl Pixel for f64 {
...
     dpi_factor.is_sign_positive() && dpi_factor.is_normal()
 }
 
 /// A position represented in logical pixels.
 ///
-/// The position is stored as floats, so please be careful. Casting floats to integers truncates the fractional part,
-/// which can cause noticable issues. To help with that, an `Into<(i32, i32)>` implementation is provided which
-/// does the rounding for you.
+/// The position is stored as floats, so please be careful. Casting floats to integers truncates the
+/// fractional part, which can cause noticable issues. To help with that, an `Into<(i32, i32)>`
```
Ralith: This should discuss the issues, and emphasize that some issues (i.e. user-visible loss of precision) are not solved by using the `Into` impl.

src/dpi.rs
```diff
@@ -205,9 +208,9 @@ impl<P: Pixel, X: Pixel> Into<[X; 2]> for LogicalPosition<P> {
 
 /// A position represented in physical pixels.
 ///
-/// The position is stored as floats, so please be careful. Casting floats to integers truncates the fractional part,
-/// which can cause noticable issues. To help with that, an `Into<(i32, i32)>` implementation is provided which
-/// does the rounding for you.
+/// The position is stored as floats, so please be careful. Casting floats to integers truncates the
+/// fractional part, which can cause noticable issues. To help with that, an `Into<(i32, i32)>`
```
Ralith: As above.

src/dpi.rs
```diff
@@ -274,9 +277,9 @@ impl<P: Pixel, X: Pixel> Into<[X; 2]> for PhysicalPosition<P> {
 
 /// A size represented in logical pixels.
 ///
-/// The size is stored as floats, so please be careful. Casting floats to integers truncates the fractional part,
-/// which can cause noticable issues. To help with that, an `Into<(u32, u32)>` implementation is provided which
-/// does the rounding for you.
+/// The size is stored as floats, so please be careful. Casting floats to integers truncates the
+/// fractional part, which can cause noticable issues. To help with that, an `Into<(u32, u32)>`
```
Ralith: As above.

src/dpi.rs
Osspial: Oh wow, they're even more common than 1080p screens. Good catch

src/dpi.rs
Osspial: Possibly, but that'll take implementation work on each platform and isn't implemented currently. Feel free to open a dedicated issue for it, though.

src/dpi.rs
Osspial: Could you elaborate on how this is misleading? This is mainly intended as a quick rule of thumb if you absolutely need some DPI value to plug into an equation or API that asks for it - not as a hard rule.

src/dpi.rs
Osspial: Huh, those docs are completely wrong now - they're stored generically, not as floats. Will revise, thanks.

src/dpi.rs
```diff
@@ -1,80 +1,85 @@
...
-//! changes, and [`HiDpiFactorChanged`](../enum.WindowEvent.html#variant.HiDpiFactorChanged) events
-//! whenever the DPI factor changes. Receiving either of these events means that the physical size of your window has
-//! changed, and you should recompute it using the latest values you received for each. If the logical size and the
-//! DPI factor change simultaneously, `winit` will send both events together; thus, it's recommended to buffer
-//! these events and process them at the end of the queue.
+//! - **Windows:** On Windows 8 and 10, per-monitor scaling is readily configured by users from the
+//!   display settings. While users are free to select any option they want, they're only given a
+//!   selection of "nice" scale factors, i.e. 1.0, 1.25, 1.5... on Windows 7, the scale factor is
+//!   global and changing it requires logging out. See [this article][windows_1] for technical
+//!   details.
```
kchibisov: FWIW Windows 7 is EOL in 2 weeks or something, not sure how mentioning it worth here.

src/dpi.rs
```diff
@@ -1,80 +1,85 @@
...
+//! - **macOS:** "retina displays" have a scale factor of 2.0. Otherwise, the scale factor is 1.0.
+//!   Intermediate scale factors are never used. It's possible for any display to use that 2.0 scale
+//!   factor, given the use of the command line.
+//! - **X11:** On X11, we calcuate the scale factor based on the millimeter dimensions provided by
+//!   XRandR. This is almost certainly the wrong approach, but many man-hours have been spent trying
+//!   to figure out how to do it better and X11 provides no easy answer.
```
kchibisov: We're also using `Xft.dpi` thing and an `env` variable, so I guess you should mention at least `Xft.dpi`.

src/dpi.rs
Osspial: Still used be a significant number of people according to the [Steam Hardware Survey](https://store.steampowered.com/hwsurvey/Steam-Hardware-Software-Survey-Welcome-to-Steam), so as much as Microsoft may want it to be it's not a dead product yet.

src/dpi.rs
Osspial: Oh, missed that when going through the source code! Thanks for bringing it up. Honestly, we should probably remove the screen size DPI code and just fall back to `1.0`, since literally nobody expects the current behavior.

kchibisov: @chrisduerr will be very unhappy if you remove it and a lot of other alacritty users.

kchibisov: besides this joke part. There's no other way around to use multi monitor setups on X11 with different monitors DPI. So with `Xft.dpi` each monitor will have a scaling factor of `Xft.dpi / 96.0`. This is why we brought https://github.com/rust-windowing/winit/issues/836 , it's a great feature to have a actually working multi monitor mixed DPI setup working while having Xft.dpi or something in Xresources, since some apps relies on it.

Osspial: :/

Osspial: Eh, I guess we should just accept that DPI scaling on X11 has always, and will always be, a clusterfuck. I'm just glad I don't have to personally deal with any of it.

src/dpi.rs
```diff
@@ -63,9 +63,11 @@
...
 //!   Intermediate scale factors are never used. It's possible for any display to use that 2.0 scale
 //!   factor, given the use of the command line.
-//! - **X11:** On X11, we calcuate the scale factor based on the millimeter dimensions provided by
-//!   XRandR. This is almost certainly the wrong approach, but many man-hours have been spent trying
-//!   to figure out how to do it better and X11 provides no easy answer.
+//! - **X11:** Many man-hours have been spent trying to figure out how to handle DPI in X11. Winit
+//!   currently uses a three-pronged approach.
+//!   + Calcuate the scale factor based on the millimeter monitor dimensions provided by XRandR.
+//!   + Use the value set in `Xft.dpi` in Xresources. Overrides XRandR's monitor size.
+//!   + Use the value in the `WINIT_X11_SCALE_FACTOR` environment variable. Overrides `Xft.dpi` and XRandR.
```
kchibisov: This is not true, WINIT_X11_SCALE_FACTOR doesn't overrides Xft.dpi. We do have a bug about it https://github.com/rust-windowing/winit/issues/1141 .

kchibisov: I meant that the situation is not like we have a bug that we should fix. The situation is more like "should we override it or not", I can send patches for both #1141 and #836 , but we're kind of not sure what is fine for winit here.

src/dpi.rs
Osspial: I've added a bit to the module docs.

src/dpi.rs
Osspial: Ah, that's a weird choice. IMO `WINIT_X11_SCALE_FACTOR` should override `Xft.dpi`, since it's more specific. I'd be fine accepting a PR to fix it, and implementing the solution you described in https://github.com/rust-windowing/winit/issues/836#issuecomment-562912905.

kchibisov: Then I'll send a PR to fix both of them.

src/dpi.rs
Ralith: I don't think it's appropriate to recommend a value that is expressly not DPI (e.g. wildly varying for the same panel depending on type of installation) be thought of as "kinda-sorta DPI", let alone used with APIs that require DPI. If the user wants to invent a plausible DPI based on scaling factor that's their prerogative, but it's not a generally reasonable thing to do with a scale factor.

Osspial: These aren't just values pulled out of thin air - the `96` is based on the DPI semantics for Windows, X11, and CSS, while `160` is the the documented base DPI factor for Android and (based on measurements) approximately the DPI factor for iOS. I'd be fine with making it more clear that these values aren't hard rules, but they'll be helpful numbers for most applications and as such I believe it's absolutely appropriate to include them.

Ralith: This information would not be helpful for any GUI application I've written or made meaningful contributions to. Android runs on both TVs and watches, whose DPI factors are going to be both wildly different *and* wildly inappropriate as the basis for user interface scaling. This paragraph encourages people to continue to think about user interface scaling in terms of DPI, which is wrong.

Osspial: I'm not arguing that the DPI should be used as the basis for user interface scaling. My intended usecase for that section is mainly for managing converting font sizes to pixel sizes , since font sizes are generally specified in points, rather than some DPI-agnostic unit. I'd assume that that information wouldn't have been useful to you since you'd likely have been using UI toolkits that automatically managed those conversions for you. Personally speaking, I can point to several specific times where I've been writing font handling code and would've liked to have a clearly documented DPI value so I could be certain that I was doing the conversions correctly.

These are the values that have been officially documented by OS authors and standard writers. This documentation isn't inventing anything new - it's just reporting the values that GUI toolkit designers have used as the bases for font scaling.

Lokathor: Since winit has no obvious UI kit, we should be focused on docs that help folks build some good UI kit, and then consumers of that kit won't even read your docs since they'll read the kit's docs.

src/dpi.rs
chrisduerr: I think with @kchibisov's PR things work perfectly fine. You have the choice to use Xft.dpi if you need it for other applications, you have the choice to still use dynamic scaling only for winit applications by passing `0` to the env variable, you have the choice to pick any arbitrary factor and if you don't like any of it, you can just fall back to the 'default' X behavior by passing `0`.

src/dpi.rs
Osspial: Alright, I've done some more reading into HTML/CSS best practices, and I've updated the documentation to reflect that. Thank you for bringing this up.

src/dpi.rs
```diff
@@ -1,80 +1,102 @@
...
+//! normally being less than a quarter the size of their desktop counterparts. What's more, neither
+//! desktop nor mobile screens are consistent resolutions within their own size classes - common
+//! mobile screens range from below 720p to above 1440p, and desktop screens range from 720p to 5K
+//! and beyond.
 //!
-//! Modern screens have a defined physical resolution, most commonly 1920x1080. Indepedent of that is the amount of
-//! space the screen occupies, which is to say, the height and width in millimeters. The relationship between these two
-//! measurements is the *pixel density*. Mobile screens require a high pixel density, as they're held close to the
-//! eyes. Larger displays also require a higher pixel density, hence the growing presence of 1440p and 4K displays.
+//! Given that, it's naive for 2D content to assume that it'll only be displayed on screens with a
```
goddessfreya: I'd avoid calling people who don't really think that much/at all about DPI naive.

src/dpi.rs
```diff
@@ -1,80 +1,102 @@
...
-//! There are two ways to get the DPI factor:
-//! - You can track the [`HiDpiFactorChanged`](../enum.WindowEvent.html#variant.HiDpiFactorChanged) event of your
-//!   windows. This event is sent any time the DPI factor changes, either because the window moved to another monitor,
-//!   or because the user changed the configuration of their screen.
-//! - You can also retrieve the DPI factor of a monitor by calling
-//!   [`MonitorHandle::hidpi_factor`](../monitor/struct.MonitorHandle.html#method.hidpi_factor), or the
-//!   current DPI factor applied to a window by calling
-//!   [`Window::hidpi_factor`](../window/struct.Window.html#method.hidpi_factor), which is roughly equivalent
-//!   to `window.current_monitor().hidpi_factor()`.
+//! The scale factor correlates with, but no direct relationship to, the screen's actual DPI (dots
```
goddessfreya: "but no" => "but has no"

src/dpi.rs
```diff
@@ -1,80 +1,102 @@
...
-//! The window's logical size is conserved across DPI changes, resulting in the physical size changing instead. This
-//! may be surprising on X11, but is quite standard elsewhere. Physical size changes always produce a
-//! [`Resized`](../event/enum.WindowEvent.html#variant.Resized) event, even on platforms where no resize actually occurs,
-//! such as macOS and Wayland. As a result, it's not necessary to separately handle
-//! [`HiDpiFactorChanged`](../event/enum.WindowEvent.html#variant.HiDpiFactorChanged) if you're only listening for size.
+//! Winit's position and size types types are generic over their exact pixel type, `P`, to allow the
+//! API to have integer precision where appropriate (e.g. most window manipulation functions) and
+//! floating precision when necessary (e.g. logical sizes for fractional scale factors and touch
+//! input). If `P` is a floating-point type, please do not cast the values with `as {int}`. Doing so
+//! will truncate the fractional part of the float, rather than round to the nearest integer - the
```
goddessfreya: My opinion, but, the hyphen use makes the sentence read clunky.

src/dpi.rs
```diff
@@ -1,80 +1,102 @@
...
+//!   display settings. While users are free to select any option they want, they're only given a
+//!   selection of "nice" scale factors, i.e. 1.0, 1.25, 1.5... on Windows 7, the scale factor is
+//!   global and changing it requires logging out. See [this article][windows_1] for technical
+//!   details.
+//! - **macOS:** "retina displays" have a scale factor of 2.0. Otherwise, the scale factor is 1.0.
+//!   Intermediate scale factors are never used. It's possible for any display to use that 2.0 scale
+//!   factor, given the use of the command line.
+//! - **X11:** Many man-hours have been spent trying to figure out how to handle DPI in X11. Winit
+//!   currently uses a three-pronged approach.
+//!   + Calcuate the scale factor based on the millimeter monitor dimensions provided by XRandR.
```
goddessfreya: Why this odd reverse ordering? Why not something like "Winit first looks for the value set in the `WINIT_X11_SCALE_FACTOR` environment variable, using it if present. If not present, Winit looks for [...]"

Osspial: There was a reason, but I'm too tired to articulate what it was right now. I've updated the docs to hopefully be better-worded.

src/dpi.rs
```diff
@@ -1,80 +1,106 @@
...
-//!   [`Window::scale_factor`](crate::window::Window::scale_factor), which is roughly equivalent
-//!   to `window.current_monitor().scale_factor()`.
+//! The scale factor correlates with, but no has direct relationship to, the screen's actual DPI
+//! (dots per inch). Operating systems used to define the scale factor in terms of the screen's
+//! approximate DPI (at the time, 72 pixels per inch), but [Microsoft decided to report that the DPI
+//! was roughly 1/3 bigger than the screen's actual DPI (so, 96 pixels per inch) in order to make
+//! text more legible][microsoft_dpi]. As a result, the exact DPI as defined by the OS doesn't carry
+//! a whole lot of weight when designing cross-platform UIs. Scaled pixels should generally be used
+//! as the base unit for on-screen UI measurement, instead of DPI-dependent units such as
+//! [points][points] or [picas][picas].
```
Ralith: This is still confusing. winit doesn't expose points or picas so it's not clear why they're being discussed, and "correlates with, but has no direct relationship to" is a contradiction in terms.

Osspial: I chose to talk about points and picas because, before going through this review process, I thought they were the units you were supposed to use for font size, and I don't think that's an uncommon misconception to have.

My thoughts when writing that phrase was `correlates = relates to, though not necessarily directly` and `direct relationship = there is a clear mathematical rule you can use to derive property B from property A`. I'll admit that that's a somewhat lawerly application or language though.

PR #1354, opened 2020-01-03
On X11, make `WINIT_HIDPI_FACTOR` dominate `Xft.dpi` in some cases

### LINKED ISSUE #836: Allow ignoring Xft.dpi

kchibisov:
This commit makes `WINIT_HIDPI_FACTOR` dominate `Xft.dpi` in general and
adds a special value `0` for `WINIT_HIDPI_FACTOR` to use winit computed
DPI factor with randr over Xft.dpi.

We agreed with @Osspial [here](https://github.com/rust-windowing/winit/pull/1349#discussion_r362718880) that the proposed [solution](https://github.com/rust-windowing/winit/issues/836#issuecomment-562912905) for #836 makes sense and also that `WINIT_DPI_FACTOR` should dominate `Xft.dpi`.

However I'm wondering whether I should send this PR to dpi-overhaul branch instead?

Fixes #836 #1141

cc @chrisduerr

### DIFF
--- a/src/platform_impl/linux/x11/util/randr.rs
+++ b/src/platform_impl/linux/x11/util/randr.rs
@@ -109,7 +109,10 @@ impl XConnection {
                 } else if let Ok(dpi) = f64::from_str(&var) {
                     EnvVarDPI::Scale(dpi)
                 } else {
-                    EnvVarDPI::NotSet
+                    panic!(
+                        "`WINIT_HIDPI_FACTOR` invalid; DPI factors must be either normal floats greater than 0, or `randr`. Got `{}`",
+                        var
+                    );
                 }
             },
         );
@@ -125,7 +128,7 @@ impl XConnection {
             EnvVarDPI::Scale(dpi_override) => {
                 if !validate_hidpi_factor(dpi_override) {
                     panic!(
-                        "`WINIT_HIDPI_FACTOR` invalid; DPI factors must be either normal floats greater then 0 or `randr` string. Got `{}`",
+                        "`WINIT_HIDPI_FACTOR` invalid; DPI factors must be either normal floats greater than 0, or `randr`. Got `{}`",
                         dpi_override,
                     );
                 }
--- a/src/platform_impl/linux/x11/util/randr.rs
+++ b/src/platform_impl/linux/x11/util/randr.rs
@@ -6,6 +6,13 @@ use super::{
 };
 use crate::{dpi::validate_hidpi_factor, platform_impl::platform::x11::VideoMode};
 
+/// Represents values of `WINIT_HIDPI_FACTOR`.
+pub enum EnvVarDPI {
+    Randr,
+    Scale(f64),
+    NotSet,
+}
+
 pub fn calc_dpi_factor(
     (width_px, height_px): (u32, u32),
     (width_mm, height_mm): (u64, u64),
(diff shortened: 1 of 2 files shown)


murarth:
I wonder if it might be better to use some string, e.g., `auto` instead of a `0` value to trigger this special behavior. I believe this would more clearly express user intent and would also leave open the possibility of adding some other special case in the future, if the need should arise.

kchibisov:
@murarth I was also thinking about it, but decided to avoid for some reasons.

- I don't like multi type variables it should be float or string, not string and float.
- It's easier to handle zero then check for string, match it and in the same time handle arbitrary floats.

Osspial:
> However I'm wondering whether I should send this PR to dpi-overhaul branch instead?

Basing it on master is fine. Rebasing `dpi-overhaul` onto this should be easy, and targeting `master` will let alacritty use it while it's still on the `alpha6` release.

kchibisov:
>also leave open the possibility of adding some other special case in the future, if the need should arise.

DPI on X11 isn't vary that much tbh, like you can get one from `Xft.dpi`,RandR thing and just set in winit in the current winit state. I really can't imagine anything new to this variable tbh.

Well, if you feel really strong on using `auto` instead of `0`, I'll be fine with changing things, just let me know.

murarth:
> * I don't like multi type variables it should be float or string, not string and float.

To me, this feels more like a case of `enum { Auto, Scale(f64) }` versus `f64` with a hidden special value `0.0`. I believe general Rust conventions tend to favor the former over the latter.

> DPI on X11 isn't vary that much tbh, like you can get one from `Xft.dpi`,RandR thing and just set in winit in the current winit state. I really can't imagine anything new to this variable tbh.

You may be right, but I tend to think there are often unforeseen use cases in a complex situation like this, with multiple interacting software systems.

> Well, if you feel really strong on using `auto` instead of `0`, I'll be fine with changing things, just let me know.

Unless it conflicts with any existing software (I don't believe any software other winit would be interpreting this value, anyway), I would say use `auto`. If anyone objects to this on the basis of it making use of this env variable more difficult, I will concede the point.

Osspial:
Small nit - I think `physical` would be a better term than `auto`, since `auto` to me implies "do whatever the default automatic behavior is" rather than "ignore `Xft.dpi`".

kchibisov:
>Small nit - I think physical would be a better term than auto, since auto to me implies "do whatever the default automatic behavior is" rather than "ignore Xft.dpi".

Hm, I'm not sure about `physical`, but maybe `randr` will make more sense. The variable is X11 only anyway.

murarth:
I'm not picky about which string is used. Whatever makes it clear to users of this env var. `randr` is fine with me.

src/platform_impl/linux/x11/util/randr.rs
```diff
@@ -107,16 +100,50 @@ impl XConnection {
...
+        // Override DPI if `WINIT_HIDPI_FACTOR` variable is set
+        let dpi_env = env::var("WINIT_HIDPI_FACTOR").ok().map_or_else(
+            || EnvVarDPI::NotSet,
+            |var| {
+                if var.to_lowercase() == "randr" {
+                    EnvVarDPI::Randr
+                } else if let Ok(dpi) = f64::from_str(&var) {
+                    EnvVarDPI::Scale(dpi)
+                } else {
+                    EnvVarDPI::NotSet
```
kchibisov: Should we blow up here? I think like it makes sense not to do so for backwards compatibility.

Osspial: Why do you say we shouldn't? Winit blew up before if you set the environment variable to an invalid value, and this doesn't break any previously-invalid values.

src/platform_impl/linux/x11/util/randr.rs
```diff
@@ -94,38 +101,49 @@ impl XConnection {
...
-            } else {
+        let dpi_env = env::var("WINIT_HIDPI_FACTOR").ok().map_or_else(
+            || EnvVarDPI::NotSet,
+            |var| {
+                if var.to_lowercase() == "randr" {
+                    EnvVarDPI::Randr
+                } else if let Ok(dpi) = f64::from_str(&var) {
+                    EnvVarDPI::Scale(dpi)
+                } else {
+                    EnvVarDPI::NotSet
```
murarth: It looks like this is treating an invalid env var value the same as an absent env var. Is that the desired behavior?

src/platform_impl/linux/x11/util/randr.rs
kchibisov: Not really, we allowed arbitrary string before.

src/platform_impl/linux/x11/util/randr.rs
kchibisov: We allowed arbitrary strings, so the idea was to follow the old behavior and not to blow up here.  If you feel like we should blowup let me know, however I can see an issue if we add one more env var and user will have 2 winit apps with different winit version.

src/platform_impl/linux/x11/util/randr.rs
murarth: Oops, didn't see the existing comment chain for this.

It seems that the previous implementation was also silently ignoring any string that didn't parse to a valid `f64`, but I don't think there's any reason to continue supporting that behavior. winit should panic on an invalid value in this env var. After all, we are already panicking on a valid `f64` that constitutes an invalid DPI factor.

chrisduerr: I think I'm in favor of exploding as a downstream. I'd much rather have it blow up in the face of my users than having to troubleshoot for hours why it doesn't do what it's supposed to do only because there's a zero width character or some funkyness somewhere.

kchibisov: Oh, wait, let's at least not blow up on empty string. What do you think?

murarth: Yes, I think it's reasonable to treat an empty var and an absent var as the same.

PR #1375, opened 2020-01-08
Don't discard high-precision cursor position data

Ralith:
Most platforms (X11, wayland, macos, stdweb, ...) provide physical positions in f64 units, which can contain meaningful fractional data. For example, this can be empirically observed on modern X11 using a typical laptop touchpad. This is useful for e.g. content creation tools, where cursor motion might map to brush strokes on a canvas with higher-than-screen resolution, or positioning of an object in a vector space.

I have not audited other uses of `PhysicalPosition`. IMO there's a strong case for making it non-generic; it was difficult to trace the types of all uses even for this one event, and 64 bits ought to be enough for anyone.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 - On Windows, fix bug where `RedrawRequested` would only get emitted every other iteration of the event loop.
 - On X11, fix deadlock on window state when handling certain window events.
 - `WindowBuilder` now implements `Default`.
-- `WindowEvent::CursorMoved` changed to `f64` units, preserving high-precision data supplied by most backends
+- **Breaking:** `WindowEvent::CursorMoved` changed to `f64` units, preserving high-precision data supplied by most backends
 
 # 0.20.0 (2020-01-05)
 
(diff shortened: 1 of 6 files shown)


Ralith:
It's also possible to get similar higher-quality data on Windows via e.g. `WM_POINTER`, though winit's current implementation does not do so.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 - On Windows, fix bug where `RedrawRequested` would only get emitted every other iteration of the event loop.
 - On X11, fix deadlock on window state when handling certain window events.
 - `WindowBuilder` now implements `Default`.
+- `WindowEvent::CursorMoved` changed to `f64` units, preserving high-precision data supplied by most backends
```
murarth: Now that we're out of alpha, I think we should be marking breaking API changes.

```suggestion
- **Breaking:** `WindowEvent::CursorMoved` changed to `f64` units, preserving high-precision data supplied by most backends
```

murarth:
This seems like a sensible change to me. Are there any objections?

Osspial:
Actually, I'd like to hold off on merging this for now. I'd like to get the various bugfixes we've done on master before doing a breaking release, since it hasn't even been a week since the last breaking release.

Ralith:
Maybe make a 0.20.x branch?

PR #1378, opened 2020-01-09
Return physical position in `CursorMoved` on macOS

### LINKED ISSUE #1371: CursorMoved position is returns a PhysicalPosition with logical values inside it
I am using winit 0.20 on MacOS. I have a 13" MacBook Pro connected to a 1440p non-retina display.

Repro Steps:
 - Create a 900x600 window logically sized on a retina-scaled display.
 - Print the position field of the CursorMoved event when this event is fired
 - Move the mouse to the bottom right of the window and observe the output
 - Example code provided below

Expected:
 - When the window is on the retina display, the returned physical coordinates should be (1800, 1200) and when converted to logical coordinates should be (900, 600)

Observed
 - The PhysicalPosition returned in the CursorMoved event is returning (900, 600) and when converted to logical coordinates is (450, 300)

Notes
 - The window is sized as expected on both displays (1800x1200 on the retina display and 900x600 on the non-retina display in raw pixels)
 - When the window is moved to the non-retina display, the physical and logical coordinates from the event are the expected value of (900, 600)
 - scale_factor correctly changes to 2 on the Retina display and 1 on the non-retina display

```rust
use winit::{
    event::{Event, WindowEvent},
    window::WindowBuilder,
};

fn main() {
    simple_logger::init().unwrap();

        .with_inner_size(winit::dpi::LogicalSize::new(900.0, 600.0))
        .build(&event_loop)
        .unwrap();

        //println!("{:?}", event);

        match event {
            Event::WindowEvent {
                window_id,
            Event::MainEventsCleared => {
                window.request_redraw();
            },
            Event::WindowEvent {
                event:
                    WindowEvent::CursorMoved {
                        device_id,
                        position,
                        ..
                    },
                ..
            } => {
                let logical_position = position.to_logical::<i32>(window.scale_factor());
                println!("Scale Factor: {}    Physical {:?}    Logical: {:?}", window.scale_factor(), position, logical_position);
            },
            _ => (),
        }
    });
}
```

hecrj:
Fixes #1371.

I found the issue described in #1371 when upgrading `iced_winit`: https://github.com/hecrj/iced/pull/147

The fix takes the same approach used in `window_delegate`. I am not sure if there is a better way to obtain the cursor position using physical coordinates.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - On Windows, fix bug where `RedrawRequested` would only get emitted every other iteration of the event loop.
 - On X11, fix deadlock on window state when handling certain window events.
 - `WindowBuilder` now implements `Default`.
+- On macOS, fix `CursorMoved` event reporting the cursor position using logical coordinates.
 
 # 0.20.0 (2020-01-05)
 
--- a/src/platform_impl/macos/view.rs
+++ b/src/platform_impl/macos/view.rs
@@ -17,6 +17,7 @@ use objc::{
 };
 
 use crate::{
+    dpi::LogicalPosition,
     event::{
         DeviceEvent, ElementState, Event, KeyboardInput, ModifiersState, MouseButton,
         MouseScrollDelta, TouchPhase, VirtualKeyCode, WindowEvent,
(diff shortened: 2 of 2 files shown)


aclysma:
#1371 no longer reproduces for me with this PR. Thanks!

PR #1381, opened 2020-01-09
Move `ModifiersChanged` variant to `WindowEvent`

### LINKED ISSUE #1365: [Wayland] No ModifiersChanged event send when window doesn't have keyboard focus, but at least has pointer focus, on modifer release
For e.g. if you open a URL in Alacritty holding some modifier your keyboard focus will move to your browser, but pointer could be still over Alacritty and will highlight URL even if you release all your modifiers. So to sync your modifiers you should bring keyboard focus back to the window.

I feel like winit can't do anything about it, knowing some Wayland security concerns, however it's better to keep this issue around, maybe there's something can be done.

cc @chrisduerr

murarth:
Fixes #1365

I've only implemented the change for X11 and Wayland. Others will need to add Mac and Windows implementations to this PR.


kchibisov:
@vberger I have a question about Wayland implementation here. Is it correct that compositor *must* re advertise `Modifiers` on each `KbdEvent::Enter`, since I don't see a point about sending such events from `winit`. My small testing with `glutin` example shows that for e.g. sway is sending such event and `winit` for a second is sending wrong event with my old modifier which I'm already not holding. So I feel like the right way to address that issue would be to rely on compositor for `KbdEvent::Enter` and send empty mods for `Leave`?

vberger:
@kchibisov Actually, given what you describe I suspect there might be some bug in SCTK.

Could you share a `WAYLAND_DEBUG=1` trace featuring the following action sequence under Gnome+Wayland? I'd like to compare it to what happens in sway:

- focus the window with no modifier pressed
- press a modifier
- unfocus the window without releasing the modifier
- refocus the windows while the modifier is still pressed

kchibisov:
@vberger Here are 2 logs from glutin window example, one from gnome and one from sway. I was doing the thing you've described here and using `ALT` modifier.

[gnome-debug.txt](https://github.com/rust-windowing/winit/files/4051476/gnome-debug.txt)

[sway-debug.txt](https://github.com/rust-windowing/winit/files/4051479/sway-debug.txt)

vberger:
@kchibisov Ok, thanks. The main difference seems to be that Gnome sends the `modifiers` event before the `enter` event, while sway sends it after. In any case, this is out of topic for this issue, we should move this somewhere else.

rye:
So, after digging at macOS a bit, I have a working reproduction that I can work with:

<details><summary>🔎 trace of <code>window_debug</code> with debugging added within the <code>flags_changed</code> function to print out the computed modifiers</summary><pre>
2020-01-22 09:03:08,041 TRACE [winit::platform_impl::platform::app_delegate] Triggered `applicationDidFinishLaunching`
2020-01-22 09:03:08,042 TRACE [winit::platform_impl::platform::app_delegate] Completed `applicationDidFinishLaunching`
2020-01-22 09:03:08,091 TRACE [winit::platform_impl::platform::window_delegate] Triggered `windowDidBecomeKey:`
2020-01-22 09:03:08,091 TRACE [winit::platform_impl::platform::window_delegate] Completed `windowDidBecomeKey:`
2020-01-22 09:03:08,093 TRACE [winit::platform_impl::platform::app_delegate] Triggered `applicationDidBecomeActive`
2020-01-22 09:03:08,093 TRACE [winit::platform_impl::platform::app_delegate] Completed `applicationDidBecomeActive`
2020-01-22 09:03:08,101 TRACE [winit::platform_impl::platform::view] Triggered `validAttributesForMarkedText`
2020-01-22 09:03:08,101 TRACE [winit::platform_impl::platform::view] Completed `validAttributesForMarkedText`
2020-01-22 09:03:08,101 TRACE [winit::platform_impl::platform::view] Triggered `validAttributesForMarkedText`
2020-01-22 09:03:08,101 TRACE [winit::platform_impl::platform::view] Completed `validAttributesForMarkedText`
2020-01-22 09:03:08,101 TRACE [winit::platform_impl::platform::view] Triggered `validAttributesForMarkedText`
2020-01-22 09:03:08,101 TRACE [winit::platform_impl::platform::view] Completed `validAttributesForMarkedText`
2020-01-22 09:03:09,162 TRACE [winit::platform_impl::platform::view] Triggered `flagsChanged`
2020-01-22 09:03:09,162 TRACE [winit::platform_impl::platform::view] Modifiers: LOGO
2020-01-22 09:03:09,162 TRACE [winit::platform_impl::platform::view] Completed `flagsChanged`
2020-01-22 09:03:10,794 TRACE [winit::platform_impl::platform::window_delegate] Triggered `windowDidResignKey:`
2020-01-22 09:03:10,794 TRACE [winit::platform_impl::platform::window_delegate] Completed `windowDidResignKey:`
2020-01-22 09:03:10,795 TRACE [winit::platform_impl::platform::app_delegate] Triggered `applicationDidResignActive`
2020-01-22 09:03:10,795 TRACE [winit::platform_impl::platform::app_delegate] Completed `applicationDidResignActive`
2020-01-22 09:03:12,543 TRACE [winit::platform_impl::platform::window_delegate] Triggered `windowDidBecomeKey:`
2020-01-22 09:03:12,544 TRACE [winit::platform_impl::platform::window_delegate] Completed `windowDidBecomeKey:`
2020-01-22 09:03:12,545 TRACE [winit::platform_impl::platform::app_delegate] Triggered `applicationDidBecomeActive`
2020-01-22 09:03:12,545 TRACE [winit::platform_impl::platform::app_delegate] Completed `applicationDidBecomeActive`
2020-01-22 09:03:16,191 TRACE [winit::platform_impl::platform::view] Triggered `flagsChanged`
2020-01-22 09:03:16,191 TRACE [winit::platform_impl::platform::view] Modifiers: LOGO
2020-01-22 09:03:16,191 TRACE [winit::platform_impl::platform::view] Completed `flagsChanged`
2020-01-22 09:03:18,579 TRACE [winit::platform_impl::platform::window_delegate] Triggered `windowDidResignKey:`
2020-01-22 09:03:18,579 TRACE [winit::platform_impl::platform::window_delegate] Completed `windowDidResignKey:`
2020-01-22 09:03:18,580 TRACE [winit::platform_impl::platform::app_delegate] Triggered `applicationDidResignActive`
2020-01-22 09:03:18,580 TRACE [winit::platform_impl::platform::app_delegate] Completed `applicationDidResignActive`
2020-01-22 09:03:27,096 TRACE [winit::platform_impl::platform::window_delegate] Triggered `windowDidBecomeKey:`
2020-01-22 09:03:27,097 TRACE [winit::platform_impl::platform::window_delegate] Completed `windowDidBecomeKey:`
2020-01-22 09:03:27,098 TRACE [winit::platform_impl::platform::app_delegate] Triggered `applicationDidBecomeActive`
2020-01-22 09:03:27,098 TRACE [winit::platform_impl::platform::app_delegate] Completed `applicationDidBecomeActive`
2020-01-22 09:03:27,346 TRACE [winit::platform_impl::platform::view] Triggered `flagsChanged`
2020-01-22 09:03:27,346 TRACE [winit::platform_impl::platform::view] Modifiers: (empty)
2020-01-22 09:03:27,346 TRACE [winit::platform_impl::platform::view] Completed `flagsChanged`
2020-01-22 09:03:28,707 TRACE [winit::platform_impl::platform::activation_hack] Triggered `activationHackMouseMoved`
2020-01-22 09:03:28,707 TRACE [winit::platform_impl::platform::activation_hack] Completed `activationHackMouseMoved`
2020-01-22 09:03:30,025 TRACE [winit::platform_impl::platform::view] Triggered `keyDown`
2020-01-22 09:03:30,026 TRACE [winit::platform_impl::platform::view] Triggered `hasMarkedText`
2020-01-22 09:03:30,026 TRACE [winit::platform_impl::platform::view] Completed `hasMarkedText`
2020-01-22 09:03:30,026 TRACE [winit::platform_impl::platform::view] Triggered `validAttributesForMarkedText`
2020-01-22 09:03:30,026 TRACE [winit::platform_impl::platform::view] Completed `validAttributesForMarkedText`
2020-01-22 09:03:30,026 TRACE [winit::platform_impl::platform::view] Triggered `validAttributesForMarkedText`
2020-01-22 09:03:30,026 TRACE [winit::platform_impl::platform::view] Completed `validAttributesForMarkedText`
2020-01-22 09:03:30,026 TRACE [winit::platform_impl::platform::view] Triggered `validAttributesForMarkedText`
2020-01-22 09:03:30,026 TRACE [winit::platform_impl::platform::view] Completed `validAttributesForMarkedText`
2020-01-22 09:03:30,026 TRACE [winit::platform_impl::platform::view] Triggered `validAttributesForMarkedText`
2020-01-22 09:03:30,026 TRACE [winit::platform_impl::platform::view] Completed `validAttributesForMarkedText`
2020-01-22 09:03:30,026 TRACE [winit::platform_impl::platform::view] Triggered `insertText`
2020-01-22 09:03:30,026 TRACE [winit::platform_impl::platform::view] Completed `insertText`
2020-01-22 09:03:30,027 TRACE [winit::platform_impl::platform::view] Completed `keyDown`
2020-01-22 09:03:30,027 TRACE [winit::platform_impl::platform::window] Dropping `UnownedWindow` (0x7fe2a0c62d30)</pre></details>

In this example I first switched (via <kbd>⌘-TAB</kbd>) rather quickly, and then rather slowly.

It seems that the _release_ of <kbd>⌘</kbd> (`LOGO`) is fired when you wait a considerable (~1s) amount of time after the window becomes "key" before releasing it, but the same event is not fired when you switch quickly, which is why this behaves "stuck," and also why it seems somewhat inconsistent.

@murarth, I am now set up locally and can PR the fix onto your fork's branch. Do I have it right that the fix is to just send a `ModifiersChanged(empty)` when a window loses focus? Also, I notice there's some tracking logic for modifiers in the Wayland impl, but I don't see the same on macOS, is that needed? (I don't think it is, since switching via <kbd>⌘-TAB</kbd> basically requires you to release <kbd>⌘</kbd> before you can do anything else, which on winit's end would trigger another event anyhow.)

murarth:
@rye That sounds right. Thank you for contributing the Mac and Windows implementations.

rye:
@murarth Okay, sweet. No problem, I'll try to figure out where the event should get fired in the Windows implementation and PR that for you as well.

src/platform_impl/macos/window_delegate.rs
```diff
@@ -319,6 +319,7 @@ extern "C" fn window_did_become_key(this: &Object, _: Sel, _: id) {
 extern "C" fn window_did_resign_key(this: &Object, _: Sel, _: id) {
     trace!("Triggered `windowDidResignKey:`");
     with_state(this, |state| {
+        state.emit_event(WindowEvent::ModifiersChanged(ModifiersState::empty()));
```
chrisduerr: I've played around with things on macOS too, but haven't gotten a chance to send a PR yet. But based on my understanding, this would break when focus is lost with a modifier pressed and the modifier is still held when the window is focused again, right?

Based on my testing, there is no flags changed emitted when the window is focused or unfocused, so sending an empty on unfocus only resets things to none, but it doesn't correctly make sure that the modifier is reapplied on enter?

rye: That is correct—the changes I made did not re-apply the former pre-unfocus modifiers once focus is regained. The Wayland implementation does have and use a `modifiers_tracker`, but I didn't find something that was already built that I could use, besides the [`modifiers` field on `ViewState`](https://github.com/rust-windowing/winit/blob/b28fd6b4fa4ac0d1e6232d5aa6d290e3c8d8e6cb/src/platform_impl/macos/view.rs#L59). I also wasn't sure _if_ we even needed it (modifiers tracking), since in most of the cases I tested I had to release all modifiers to regain focus anyhow. (The only exception is mousing out of the window, but personally I'd still expect to have to freshly hold down a modifier if I want to use it, I guess.)

Is there a way we can figure out (e.g. by querying or looking at a field) what the current modifiers are when focus is regained? I'm not sure what the best approach here is as I have absolutely zero experience with macOS window development. What is "correct" doesn't seem to be clear to me, besides clearing modifiers when the window is unfocused.

I was primarily targeting the <kbd>⌘-TAB</kbd> use case here, so I'd definitely be open to additional ideas if you have them.

chrisduerr: > since in most of the cases I tested I had to release all modifiers to regain focus anyhow

This is definitely not the case. In my testing it was quite trivial to do this.

> Is there a way we can figure out (e.g. by querying or looking at a field) what the current modifiers are when focus is regained?

When I looked at it, it didn't seem like the `state` we have on the focus callback would give us access to it. I have no experience with macOS either, though I was able to write a reliably working solution that uses key events in addition to flag changes. You can check if `state.modifiers != event_mods` and then emit an event and set `state.modifiers` to `event_mods`.

That seemed to work pretty well, though I have not tested it sufficiently to determine if this solution would work reliably in all scenarios. But based on my initial estimate, it at least seemed to give good feedback for the key events.

The problem of course is that we'd have to wait until a key event to update the modifiers, which means that they are not correct until another key is pressed.

rye: >This is definitely not the case.

Yeah, I realized this after I wrote it. The mouse is a thing, after all. :)

As I thought about it more, I realized it'd work to do this check within `mouse_click`, `key_down`, and `key_up` within the `view` module. That check would be fairly straightforward to do, and since these all come from `NSEvent` the `modifierFlags` should definitely be available, and checkable via `event_mods` as you suggested.

I'll go ahead and PR that as well.

chrisduerr: > As I thought about it more, I realized it'd work to do this check within mouse_click, key_down, and key_up within the view module. That check would be fairly straightforward to do, and since these all come from NSEvent the modifierFlags should definitely be available, and checkable via event_mods as you suggested.

That's not entirely accurate though. If a modifier is released while the window is not in focus, you won't accurately be able to tell pressed modifiers until one of these actions occurs, even though your window already has full focus. So that's definitely a bit strange.

I feel like there should be a solution without caveats like this, but I haven't had the time to look more into that yet.

rye: @chrisduerr, yeah, that's right. The fundamental problem is that we don't seem to be able to handle a `NSEvent` when focus is regained, because such an event isn't emitted. If such an event were emitted, we could just grab the modifiers out of it.

I've done quite a bit of digging now and I'm still getting more and more confident that our answer will be to just check modifiers and queue up an event to update `state.modifiers` if necessary.

While the docs I linked above do say that

>You can retrieve the window object in question by sending `object` to `notification`,

the "window object" one would retrieve would be a `NSWindow`, which [does not expose any modifiers information](https://developer.apple.com/documentation/appkit/nswindow). As such, in the context of handling a didResignKey or didBecomeKey event, we simply _don't have access_ to the modifier information.

So I think this is a protocol-level issue: our notification of gaining or losing focus gets fired on the `NSWindowDelegate`, not on the `NSView`, where keyboard, mouse, and "flags" events get fired, and indeed no event is consistently fired after regaining focus on our `NSView` that would let us get updated modifiers. Unless I'm missing something, that'd mean we don't really have a way, besides periodically checking the provided modifiers with an event that fires and sending a fresh event before sending the event that we got.

---

I've got [another PR into this branch](https://github.com/murarth/winit/pull/5) that will add this check in a couple of key spots where I think user interaction is most likely. And of course, if `ModifiersChanged` gets fired, then `state.modifiers` will be consistent and events will not get doubly fired.

src/platform_impl/macos/view.rs
```diff
@@ -623,6 +623,18 @@ fn retrieve_keycode(event: id) -> Option<VirtualKeyCode> {
     })
 }
 
+// Update `state.modifiers` if `event` has something different
+fn update_potentially_stale_modifiers(state: &mut ViewState, event: id) {
+    if state.modifiers != event_mods(event) {
+        state.modifiers = event_mods(event);
```
chrisduerr: I'd suggest assigning `event_mods(event)` to a variable, so it's not executed twice.

src/platform_impl/macos/view.rs
```diff
@@ -762,9 +776,10 @@ extern "C" fn flags_changed(this: &Object, _sel: Sel, event: id) {
             }));
         }
 
-        AppState::queue_event(EventWrapper::StaticEvent(Event::DeviceEvent {
-            device_id: DEVICE_ID,
-            event: DeviceEvent::ModifiersChanged(state.modifiers),
+        AppState::queue_event(EventWrapper::StaticEvent(Event::WindowEvent {
+            // TODO Maybe memoize get_window_id if it's safe to reuse?
+            window_id: WindowId(get_window_id(state.ns_window)),
```
chrisduerr: The window ID of the state isn't going to change. If it is extracted it should probably be used in the `for event in events` loop too. Would probably be a good change.

src/platform_impl/macos/window_delegate.rs
chrisduerr: How does this code interact with the new changes? I'd imagine it might not be necessary anymore?

What happens if a modifier is held, the window is unfocused and the modifier is still held when entering again? Does this cause an empty event without another modifier event because the internal window state never let it go?

rye: >I'd imagine it might not be necessary anymore?

That's an interesting question, and I think you're right that it's probably not be necessary to keep it. Anecdotally, commenting out this line did not change the fact that Alacritty is fixed by this PR.

One point of contention is that the Wayland impl _is_ emitting an empty event and then another to restore it when focus is restored. But the Wayland event lifecycle is not the same as on macOS.

>What happens if a modifier is held, the window is unfocused and the modifier is still held when entering again? Does this cause an empty event without another modifier event because the internal window state never let it go?

In this case, the window losing focus triggers an emptying event, and regaining focus does not trigger an event until selected further user interaction happens. (Either keypress, mouse click, mouse enter, mouse exit, scrolling, or pressure change.) You're right that we'd be firing an empty event. I suppose since we now have a check before actually sending out events for user interaction, we no longer need to send out empty modifiers state, since windows might not want to get a synthetic event like that.

I should also do this check (potentially updating stale modifiers) on `mouse_motion`, I think, since it's pretty lightweight.

src/platform_impl/macos/view.rs
```diff
@@ -890,6 +911,8 @@ fn mouse_motion(this: &Object, event: id) {
             return;
         }
 
+        update_potentially_stale_modifiers(state, event);
```
chrisduerr: Why have you decided to add this on `key_up` and `mouse_motion`?

I think `key_up` shouldn't be super critical, but I'd imagine the less we do on `mouse_motion` the better.

rye: I added this just for consistency's sake, but I did also have reservations about adding it on `mouse_motion`. I'd be happy to remove it from one or both locations, what do you think?

chrisduerr: I'd only add it to scenarios where it can actually trigger a relevant bug. I'm not sure about a scenario with mouse motion where it could trigger an incorrect modifier state.

But maybe consistency isn't that bad either, but it's generally a very ugly solution imho.

rye: Alright, I'll also add removing it with `mouse_motion` but I'll keep it in `key_up` for now.

kchibisov:
@vbogaevsky 

If you're marking something as WIP, but in reality others thinking that it's not, please request changes and comment why it's WIP. This PR implements ModifiersChanged event change on **all** platforms affected, so I don't see how it is WIP, knowing that both Wayland/X11/macOS backends were tested downstream. If you're not satisfied with proposed changes for macOS, please, go into a bit more details on it, otherwise things will stall for a bit longer than it should.

vbogaevsky:
@kchibisov 
I've missed @chrisduerr 's review

chrisduerr:
@Osspial It seems like the only reason this failed CI, is because the armv7-apple-ios toolchain wasn't available. Just to make sure you didn't miss this, since it's very important for macOS Alacritty users.

murarth:
@vbogaevsky You're right. This PR was incomplete when I first submitted it. I'll try to mark such PRs as Draft in the future. Have you taken a look at the Mac implementation now? I'd like to get your approval to merge this.

Since Osspial is unavailable for now, I'd like to get a Windows tester to check this. @ryanisaacg Can you check that the interaction with keyboard modifiers and window gain/lose focus is correct? (e.g. It doesn't trigger inaccurate results such as those described in #1365)

ryanisaacg:
I'm not 100% sure what the expected behavior is, but the current behavior of this branch on Windows is that the modifiers are reported to the window regardless of focus.

murarth:
@ryanisaacg The intended behavior is:

* Only a focused window receives `ModifiersChanged` events.
* If a window loses focus while modifier keys are held, it receives `ModifiersChanged` with an empty modifiers set immediately before the `Focused(false)` event.
* If modifier keys are held when a window gains focus, it receives `ModifiersChanged` with the correct set of modifiers immediately after the `Focused(true)` event.

I think that about sums it up.

@rye Can you fix the Windows implementation so that only a currently-focused window receives the `ModifiersChanged` events?

chrisduerr:
@murarth I don't think there's fundamentally anything wrong with unfocused windows receiving `ModifiersChanged` events, is there? The only reason why most platforms do not do this, is because they can't.

Always sending it seems much better than trying to retroactively fix the modifier state once focus is regained, which leads to issues like the ones on macOS and Linux.

murarth:
@chrisduerr I think, bearing in mind that only one window can have keyboard focus (to my knowledge) and a process may have multiple windows, it makes sense that only one window should be receiving keyboard events such as `ModifiersChanged`. Consequently, if no window has keyboard focus, no window should receive keyboard events.

And since we've established that some platforms receive `ModifiersChanged(<empty>)` when focus is lost, ensuring that only the active window has a non-empty modifier state, I think other platforms should have the same behavior, for the sake of platform parity.

chrisduerr:
> And since we've established that some platforms receive ModifiersChanged(<empty>) when focus is lost

IIRC, this is not the case on any platform, is it? We just decided that on Wayland it makes sense to send a `ModifiersChanged(<empty>)`, because all the modifiers are readvertised on focus. So I don't think any platform should send this by default.

I think currently macOS does this too though, but it does not receive the modifiers on focus again, but instead when the first event with modifiers is generated.

I suppose though fundamentally when you are able to always get the modifiers regardless of focus, it should be trivial to implement Wayland's behavior on Windows, simply by holding back new events until the window is focused again.

rye:
>@rye Can you fix the Windows implementation so that only a currently-focused window receives the `ModifiersChanged` events?

Probably, but I don't have a Windows machine to test on so I'd be adding that with no way of knowing if my changes were correct. Would someone be willing to take that on?

vbogaevsky:
I've done some tests on macOS.
 - when window is in focus `ModifiersChanged` events are triggered
 - when window is out of focus `ModifiersChanged` events are not triggered
 - when window is in focus, modifier key is pressed and the focus is moved from the window, no event is triggered 
 - if modifier key is released when window is out of focus focus is returned to the window with `cmd + tab`, then ModifiersChanged(empty) is triggered after `Focused(true)` immediately.
 - `fn` button generates empty `ModifiersChanged` event (it is not a regression)
 - if modifier key is released when window is out of focus and the focus is returned to the window with click on the app icon in dock, then `ModifiersChanged(empty)` is triggered only after some other event (ie `MouseMotion`).

murarth:
@chrisduerr It was my understanding from the discussion in #1365 that all platforms should exhibit the behavior [I described here](https://github.com/rust-windowing/winit/issues/1365#issuecomment-571832570). Looking back, I see that I did not make that clear. I mentioned changes to Mac and Windows, but I did not explicitly state my expectation that platforms other than X11 would be modified, if necessary, to exhibit the new behavior.

I understand that not every platform needs to issue `ModifiersChanged(empty)` when focus is lost, but I think that, with `ModifiersChanged` being a window event, and in light of the fact that a process may have multiple windows, the behavior I've described seems most sensible. Any alternative, I think, would mean some platforms would issue `ModifiersChanged` for each open window, a behavior that would not be useful and could result in buggy platform-inconsistent behavior; or which window would receive the event would change according to unspecified and platform-dependent rules.

Anyway, I just think the whole situation becomes simpler and less bug-prone if all platforms have the same behavior and, unless I'm mistaken, minimal work is required to create this behavior on each platform and no platform would lose information that is particularly valuable to applications.

----

> * when window is in focus, modifier key is pressed and the focus is moved from the window, no event is triggered

@rye Can you fix this? MacOS should issue `ModifiersChanged(empty)` immediately before `Focused(false)` if the previous modifier state was non-empty. It should also set `state.modifiers` to an empty to set to avoid an unnecessary `ModifiersChanged(empty)` event upon re-focus.

rye:
>@rye Can you fix this? MacOS should issue `ModifiersChanged(empty)` immediately before `Focused(false)` if the previous modifier state was non-empty. It should also set `state.modifiers` to an empty to set to avoid an unnecessary `ModifiersChanged(empty)` event upon re-focus.

I can.

Part of the problem as I've noted before, at least from the macOS side of things, is that `FlagsChanged` (read: modifier keys changed) is dispatched on the `NSView`, whereas `DidResignKey`/`DidBecomeKey` (read: unfocus/focus) is handled by the `NSWindowDelegate`. Thus message passing (and hence, the addition of yet more unsafe code in the focus loss handler, which is a relatively hot path) _must_ be done in order to actually interact with the `ViewState` from within the window delegate.

It took me a few hours of crawling through unsafe functions to figure out how to do it, but I think I figured it out and in my testing my change seems to fully satisfy this need. Will PR into this branch shortly.

chrisduerr:
@rye I haven't looked at the code, but if we emit an event on focus loss, wouldn't it be possible to do the same on focus? That should make it possible to get rid of the events that are setup everywhere and instead only use the flags changed, focus and unfocus hooks.

chrisduerr:
@rye I haven't looked at the code, but if we emit an event on focus loss, wouldn't it be possible to do the same on focus? That should make it possible to get rid of the events that are setup everywhere and instead only use the flags changed, focus and unfocus hooks.

rye:
@chrisduerr I seem to remember trying that and I think I [concluded](https://github.com/rust-windowing/winit/pull/1381#discussion_r369854300) that
>in the context of handling a `didResignKey` or `didBecomeKey` event, we simply _don't have access_ to the modifier information.

More explicitly, the _type_ of the event `Object` we receive during `didResignKey` and `didBecomeKey` handling is `NSNotification`, not `NSEvent`. ([`NSEvent`'s have a `modifierFlags` field](https://developer.apple.com/documentation/appkit/nsevent?language=objc); [`NSNotification`'s do not](https://developer.apple.com/documentation/foundation/nsnotification).)

It'd be possible to possibly stash the stale modifiers state and re-emit it, but then again, we don't have _up-to-date_ modifiers information until another event is fired on our view.

chrisduerr:
@rye Gotcha, so your unsafety shenanigans are just used to send the event and update the modifier state. I thought they might have allowed you to access the current modifier state of the window.

rye:
@chrisduerr exactly, yep; unsafety here is used to get (and update) the current view's `modifiers` field, which is behind a raw pointer that needs referencing. I pretty much followed the pattern from the way the `with_state` helper is structured, since that seems to work well enough. I used the `as_ref` cast in particular because I wanted to avoid dereferencing something null-y.

chrisduerr:
I've taken a look at the Windows code and swallowing the modifiers when the window isn't focused and sending the modifier state on focus is trivial.

The only problem is checking if the window is focused in the `WM_INPUT` callback, since the callback is handled in an invisible window with no reference to the real window as far as I can tell.

I also found out that this hasn't even been handled correctly in the current version of the code, which means that the `ModifiersChanged` event is emitted with the window ID of the invisible window opposed to the actual one.

chrisduerr:
@murarth Seems like the github runner failure was caused by https://github.com/rustwasm/wasm-bindgen/issues/2009, which should be fixed in the latest nightly.

Might be worth restarting to get all greens?

chrisduerr:
@ryanisaacg Could you take another look at my latest Windows changes, just to verify it does actually work? I think this is still missing a Windows review.

The previous comments by @murarth should help understand the intended behavior, if that is still uncertain.

ryanisaacg:
Sure, I'll take a look this weekend.

chrisduerr:
Just for information, I've got some patches that should further improve macOS handling and one that should fix an outstanding bug with mouse reports. But it might take a bit until I can get up the PRs for that. It'll definitely be sometime this weekend though.

ryanisaacg:
Windows looks good to me; matches expected behavior.

chrisduerr:
@vbogaevsky could you give it another test on macOS? I've made some changes that should address the remaining differences between Linux/Windows and macOS.

chrisduerr:
@rye You've implemented the initial macOS version, could you maybe help with testing the latest changes if @vbogaevsky isn't around to help out?

I'd really like to get this tested asap, so outstanding issues can be resolved quickly and this can finally be merged.

rye:
@chrisduerr, I certainly will. I'll double check macOS in the next hour or so.

chrisduerr:
Thanks a ton for the quick response. I think the main thing are the issues mentioned in @vbogaevsky last review and the expected platform behavior mentioned by @murarth.

rye:
- Giving it a once-over, the code itself for the macOS impl looks good to me now, and agree with the changes @chrisduerr made to them. One nit: `event` is now unused in `mouse_entered`, so can we rename it to start with `_`?

  - I agree (as an end-user and in principle) that generating a `mouse_motion` event doesn't need to/shouldn't happen within the handler for an entirely different event, `mouse_entered`. Immediately after a `mouse_entered` event is fired, mouse motion events are subsequently fired if they are produced; I can't really see the need to call `mouse_motion` in this spot.

- The `window` example now behaves as I'd expect, and events now seem to be fired according to the desired behavior. (As another data point Alacritty behaves entirely as expected, too.)

chrisduerr:
The `_` prefix must have gotten lost in one of my changes, of course it doesn't make any sense to have it without when it is never used. @murarth could you maybe fix this directly? I don't think there's much need to make a PR just for that change. Unfortunately I can't suggest a change since the line isn't affected by this PR.

PR #1385, opened 2020-01-10
On Wayland, fix coordinates in mouse events when scale factor isn't 1

kchibisov:
On Wayland the mouse events are in *logical* coordinates by default, however winit decides to use *physical* ones after dpi-overhaul, so we should multiply coordinates by surface scale factor before propagating downstream.

### DIFF
--- a/src/platform_impl/linux/wayland/pointer.rs
+++ b/src/platform_impl/linux/wayland/pointer.rs
@@ -7,6 +7,7 @@ use crate::event::{
 
 use super::{
     event_loop::{CursorManager, EventsSink},
+    make_wid,
     window::WindowStore,
     DeviceId,
 };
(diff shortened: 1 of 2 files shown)


src/platform_impl/linux/wayland/pointer.rs
```diff
@@ -101,13 +110,16 @@ pub fn implement_pointer(
                         surface_y,
                         ..
                     } => {
-                        if let Some(wid) = mouse_focus {
+                        if let (Some(wid), Some(surface)) = (mouse_focus, current_surface.as_ref())
```
vberger: This could be make simpler by putting both `surface` and `wid` in the same option as a tuple.

Or even by storing only `surface` and recreating the `wid` on the fly using `make_wid` whenever needed, as it is known to be a `surface` corresponding to a winit window thanks to the initial check of `find_wid` in `PtrEvent::Enter`.

kchibisov: >Or even by storing only surface and recreating the wid on the fly using make_wid whenever needed, as it is known to be a surface corresponding to a winit window thanks to the initial check of find_wid in PtrEvent::Enter.

I like this idea.

PR #1298, opened 2019-11-30
[DRAFT] Move error and dpi types into a separate crate + split out x11 symbol loading code + xinerama & multihead support

goddessfreya:
See https://github.com/rust-windowing/winit_types and https://github.com/rust-windowing/glutin_x11_sym.

@murarth @Osspial Are you folks familiar with how to setup CI for glutin_x11_sym, glutin_sys and winit_types?

Platforms:
- [x] X11
- [x] Wayland
- [x] Macos
- [x] Windows
- [x] iOS
- [x] Web

Other: 
- [ ] Tested on all platforms changed
- [ ] Compilation warnings were addressed
- [ ] `cargo fmt` has been run on this branch
- [ ] `cargo doc` builds successfully
- [ ] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users
- [ ] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior
- [ ] Created or updated an example program if it would help users understand this functionality
- [ ] Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented

PR #1403, opened 2020-01-17
On X11, if RANDR based scale factor is higher than 20 reset it to 1

kchibisov:
Some video drivers could set display metrics to odd values, which can result in
extra large scale factors (e.g. winit is sending 720 for 5k screen on nvidia
binary drivers), so let's just drop them to prevent clients from using them.
The value 20 was picked, because the DPR for 8k @ 5 inch is ~18.36.

This PR is more of a proposal. The original issue was here https://github.com/alacritty/alacritty/issues/3214 , as you can see in a log, DPR from winit is 718, which is literally insane. I see 2 options on how to fix such issues.

1 - Make winit compute DPI only if a randr value in use, otherwise return just 1 all the time. Qt is probably doing something similar IIRC
2 - Just try to ignore odd values (this PR is doing it).

The proposed `1` change is a breaking change, so I guess `2` option should fix some problems right now, however I feel like we should also implement `1` at some point to be more `native`, since auto DPI adjustments aren't something common for X11.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 - On Wayland, bump `smithay-client-toolkit` to 0.15.1.
 - On Wayland, implement `request_user_attention` with `xdg_activation_v1`.
 - On X11, emit missing `WindowEvent::ScaleFactorChanged` when the only monitor gets reconnected.
+- On X11, if RANDR based scale factor is higher than 20 reset it to 1
 
 # 0.25.0 (2021-05-15)
 
--- a/src/platform_impl/linux/x11/util/randr.rs
+++ b/src/platform_impl/linux/x11/util/randr.rs
@@ -27,7 +27,11 @@ pub fn calc_dpi_factor(
     // Quantize 1/12 step size
     let dpi_factor = ((ppmm * (12.0 * 25.4 / 96.0)).round() / 12.0).max(1.0);
     assert!(validate_scale_factor(dpi_factor));
-    dpi_factor
+    if dpi_factor <= 20. {
+        dpi_factor
+    } else {
+        1.
+    }
 }
 
 impl XConnection {


goddessfreya (dismissed):
I don't think it should be Winit's job to work around bugs in NVIDIA's proprietary driver. RANDR misreporting values should be filed with them.

kchibisov:
>I don't think it should be Winit's job to work around bugs in NVIDIA's proprietary driver. 

Well, then all clients should do the same job as e.g. alacritty here and filter out bad values from X11, etc. It doesn't happen all the time and also monitor could report wrong values, so not NVIDIA binary driver is at fault here. The RANDR based calculations isn't common for X11 and I don't think that anyone is using them by default. However winit enforces it on a users by default, so if something   all downstream clients should workaround the problem. Also you're already filtering out some RANDR metrics, so how extra large DPI differ here? Just think of it like your monitor returned a bad value instead of something normal, like 16mmx9mm.

So you're not working around bugs in NVIDIA driver at first place, you're also working around bad monitors, etc.

Alternative, don't use RANDR calculation by default.

goddessfreya:
@kchibisov The original issue mentions that the GTX1050 ran fine with Nouveau and reported proper values, while the 2050 w/ the proprietary driver did not run fine. I'm guessing they tested with the same monitor, so it's either a driver issue or a hardware issue that NVIDIA's drivers should be working around.

Either way, if this was, for example, the monitor misreporting the data, that still, IMO, wouldn''t be Winit's job to correct. If a monitor wants to report a DPI of 720, it can go right ahead. It is up to clients to decide if they trust the number being forward by Winit.

kchibisov:
>Either way, if this was, for example, the monitor misreporting the data, that still, IMO, wouldn't be Winit's job to correct. If a monitor wants to report a DPI of 720, it can go right ahead. It is up to clients to decide if they trust the number being forward by Winit.

I'd agree with you, if winit wasn't a cross platform toolkit, but X11 only toolkit, but its not an X11 only toolkit, so this will enforce alacritty(not only alacritty, I think) to have a special case for X11 in its codebase and make things harder to maintain, because winit is fine with reporting odd values from xrandr. Than, also don't use other quirks for randr, like you're checking 0mm sizes and filtering them out. Maybe monitor is fine with reporting 0.

Also we had already similar issue in the past about the xrandr here.

https://github.com/jwilm/alacritty/issues/1501

and some distros already referring to it, because or default behavior isn't something normal for X11 apps.

https://wiki.gentoo.org/wiki/Alacritty

So if you really don't want to filter out things, just don't use randr by default like Qt is doing with its environment variables. e.g. `QT_AUTO_SCREEN_SCALE_FACTOR=1`.

goddessfreya:
@kchibisov I think there is a major difference between filtering out 0x0mm monitors and filtering out numbers larger than an arbitrary cutoff. The former is a physical impossibility, as monitors have to be larger than 0x0mm, and so winit can confidently say that a 0x0mm sized monitor is a bug. The cutoff, on the other hand, is something arbitrary that should be decided by clients based on their use cases.

I'd like to point out that the proposed cutoff point of 20 would not have resolved any of the problems mentioned by users in alacritty/alacritty#1501.

> So if you really don't want to filter out things, just don't use randr by default like Qt is doing with its environment variables. e.g. `QT_AUTO_SCREEN_SCALE_FACTOR=1`.

Afaik, we check for the `WINIT_X11_SCALE_FACTOR` env var first, then try use `Xft.dpi` before using RANDR. If your monitor is misreporting it's EDID values users have many options at their disposal. In addition to the two methods mentioned above, users could probably attempt to follow [these instructions,](https://wiki.archlinux.org/index.php/Kernel_mode_setting#Forcing_modes_and_EDID) although I've never tried them myself.

-----

To get a scaling factor of 720 on a 5k monitor, NVIDIA's proprietary drivers would have be reporting a monitor surface area of 1.39mm^2. I've never see a monitor with a faulty EDID that's reporting something that small. If NVIDIA's proprietary driver is shitting itself at the simple task of reporting semi-realistic monitor sizes, users, as I mentioned at the start, should file a bug with them. Until then, users can use the options above to work around their broken driver. Downstream libraries, like winit, shouldn't have to work around NVIDIA's proprietary driver's inability to provide realistic screen sizes.

kchibisov:
>I'd like to point out that the proposed cutoff point of 20 would not have resolved any of the problems mentioned by users in alacritty/alacritty#1501.

I'm using this as an argument to switch default behavior on X11, to use randr computations only with WINIT_X11_SCALE_FACTOR=randr, like Qt is doing, so user wont be surprised that App blowups, since they should manually toggle this behavior.

>Afaik, we check for the WINIT_X11_SCALE_FACTOR env var first, then try use Xft.dpi before using RANDR. If your monitor is misreporting it's EDID values users have many options at their disposal. In addition to the two methods mentioned above, users could probably attempt to follow these instructions, although I've never tried them myself.

I'm trying to say, that I don't like when users should tweak the env variables to make apps run. This is why I'm referring to Qt here, since you can **opt-in** to auto dpi, not **opt-out** like with winit.

>To get a scaling factor of 720 on a 5k monitor, NVIDIA's proprietary drivers would have be reporting a monitor surface area of 1.39mm^2. I've never see a monitor with a faulty EDID that's reporting something that small. If NVIDIA's proprietary driver is shitting itself at the simple task of reporting semi-realistic monitor sizes, users, as I mentioned at the start, should file a bug with them. Until then, users can use the options above to work around their broken driver. Downstream libraries, like winit, shouldn't have to work around NVIDIA's proprietary driver's inability to provide realistic screen sizes.

I know, insane thing, but then clients should workaround values from winit. Though I'm not insisting to filtering out. I'm fine if winit switches the default behavior to report DPI of `1` and use randr based computations only if user requested it with env varialbe. (this should solve our old issue about DPI)

I don't have that much to say on this issue, we can just filter out things in alacritty, this is not the end of the world if winit decides to pass broken values to us.

kchibisov:
Just to make it clear.

I'm fine if you decide to avoid filtering DPI values from randr, but in this case, I feel like winit should follow Qt here and switch default behavior to match Qt. By this I mean the following.

1) if env variable is presented use it.
2) if env variable is missing and Xft.dpi isn't set, just use `1` and never use randr based computations.

IIRC some systems could be without RANDR, so maybe it's not a bad idea to follow Qt here?

However I'm not familiar with X11 that much.

Anyway, feel free to close this PR if you don't like the suggested idea, I don't have much arguments left.

goddessfreya:
> IIRC some systems could be without RANDR, so maybe it's not a bad idea to follow Qt here?

Any system which has updated it's Xorg since after 2010- hell, since after 2008- should reasonably support RANDR.

I will note that I have added Xinerama and multihead support [here](https://github.com/rust-windowing/winit/pull/1298/files#diff-2308d46eeecc9cb95b9ff378e099d22eR1), though, for those interested.

> [..] I feel like winit should follow Qt here and switch default behavior to match Qt [..] Anyway, feel free to close this PR if you don't like the suggested idea, I don't have much arguments left.

I'd like to hear what @murarth's opinion on this mater is first, they are the one maintaining the platform after all.

kchibisov:
@murarth Do you have any opinion on a problem that was brought here, sorry for pinging, just don't want this thing to stall for a long.

murarth:
I usually stay out of discussions about DPI matters because I don't use any high DPI displays and am not familiar with the expectations of users or application developers with regard to high DPI monitors. So, I'll try to speak generally about my perspective on this issue and leave room for the input of others.

> 1 - Make winit compute DPI only if a randr value in use, otherwise return just 1 all the time. Qt is probably doing something similar IIRC

If this is something application developers may prefer, I think an issue should be opened to propose this change and get further feedback from application developers, with special consideration for how this may affect platform parity.

> 2 - Just try to ignore odd values (this PR is doing it).

I'm not opposed, in principle, to accepting a workaround such as this. winit has a few workarounds for bugs or unusual behavior of window managers, so I don't think it's unreasonable to expect a workaround for buggy behavior from some other software. However, I agree that using `scale_factor > 20` as a detection of buggy behavior is arbitrary and it won't solve the issue for all users experiencing it.

The only other option appears to be checking for very small screen size. In one of the above mentioned issues, a user reported that `xrandr` was reporting `160mm x 90mm`. Can we assume that any screen of that size (or smaller) should have a `1.0` scale factor?

If there's no reliable way to identify this buggy behavior that won't be prone to both false positives and false negatives, it may have to remain unresolved and application developers will have to continue instructing users to use `WINIT_X11_SCALE_FACTOR=1` when experiencing this bug.

kchibisov:
I've pinged you because you were asked to give some opinion on this by @goddessfreya , no more no less. The issue isn't related to DPI, but rather to strange behavior of randr in some cases. So it could happen even on LowDPi monitors, since randr can report whatever it wants. 

>If this is something application developers may prefer, I think an issue should be opened to propose this change and get further feedback from application developers, with special consideration for how this may affect platform parity.

Fair enough, I'll open an issue about it at some point.

>I'm not opposed, in principle, to accepting a workaround such as this. winit has a few workarounds for bugs or unusual behavior of window managers, so I don't think it's unreasonable to expect a workaround for buggy behavior from some other software.

This is what I'm trying to say here, you shouldn't force solving platforms specific bugs in cross platform toolkit on application developers, since it isn't going to end good. Since as app developer you should have a special check on some platforms that DPI make sense, instead of using it right away, which likely lead to a wrapping of call and things to be overly complex, like you can call scale_factor more than once in your app and you should workaround every call to it...

>However, I agree that using scale_factor > 20 as a detection of buggy behavior is arbitrary and it won't solve the issue for all users experiencing it.

The value 20 was picked, because the DPR for 8k @ 5 inch is ~18.36 . So it does the same thing as you've said about filtering some small screens, but in a bit different way, so you don't match by screen size, but by DPR of some theoretical small screen with large DPR, since screens with bigger DPR are likely has less diagonal/larger resolution. Screens with dimensions 160mmx90mm are also strange for X11, so maybe filtering by some physical metrics could also work (let's say less than 200mmx200mm). This won't solve issue in general, but make winit workaround some completely odd values.

goddessfreya:
> Screens with dimensions 160mmx90mm are also strange for X11, so maybe filtering by some physical metrics could also work (let's say less than 200mmx200mm). This won't solve issue in general, but make winit workaround some completely odd values.

Filtering for screen size strikes me as a really bad idea. [There](https://www.amazon.ca/Eyoyo-Inch-IPS-Portable-Resolution/dp/B07216T9KG/ref=pd_sbs_147_2/130-3067647-4049266?_encoding=UTF8&pd_rd_i=B07216T9KG&pd_rd_r=edb0f2be-4edc-46e8-820d-0822892b3cc3&pd_rd_w=FaBbl&pd_rd_wg=BIQAg&pf_rd_p=dbebb38c-0e3d-4a67-ac15-432d7c7a2789&pf_rd_r=J7WBVEFQABBBE1VGHZ4Q&psc=1&refRID=J7WBVEFQABBBE1VGHZ4Q) [are](https://www.amazon.ca/Padarsey-inch-Monitor-HDMI-Raspberry/dp/B07DQNMRRM/ref=pd_sbs_147_5/130-3067647-4049266?_encoding=UTF8&pd_rd_i=B07DQNMRRM&pd_rd_r=edb0f2be-4edc-46e8-820d-0822892b3cc3&pd_rd_w=FaBbl&pd_rd_wg=BIQAg&pf_rd_p=dbebb38c-0e3d-4a67-ac15-432d7c7a2789&pf_rd_r=J7WBVEFQABBBE1VGHZ4Q&psc=1&refRID=J7WBVEFQABBBE1VGHZ4Q) [many](https://www.amazon.com/Performance-RCA-Touchscreen-Quad-Core-Processor/dp/B06XCKNR17/ref=sxin_3_osp104-40e8cc31_cov?ascsubtag=40e8cc31-b3c6-41a9-8345-711e3002bd3b&creativeASIN=B06XCKNR17&cv_ct_cx=laptop&cv_ct_id=amzn1.osp.40e8cc31-b3c6-41a9-8345-711e3002bd3b&cv_ct_pg=search&cv_ct_wn=osp-search&keywords=laptop&linkCode=oas&pd_rd_i=B06XCKNR17&pd_rd_r=1d93b595-283a-4954-99c5-15cbd03db138&pd_rd_w=ywTQn&pd_rd_wg=VygY2&pf_rd_p=eb3e5cda-5ec9-4d94-919d-310a5d641b8b&pf_rd_r=YWY9HRRH5DYTSK762JTP&qid=1579997735&sr=1-2-32a32192-7547-4d9b-b4f8-fe31bfe05040&tag=scrippsonsite-20) [monitors](https://www.amazon.ca/dp/B07LGRQRMW/ref=twister_B07JR93W3Z?_encoding=UTF8&psc=1), [laptops](https://www.amazon.com/Eleduino-Touchscreen-1920x1200-Raspberry-Computer/dp/B07ZSHG9VT/ref=sr_1_1?keywords=9+inch+touch+screen&qid=1579998190&sr=8-1), [touchscreens](https://www.amazon.com/dp/B07CNNP9LN/ref=psdc_1292115011_t2_B07ZSHG9VT) [smaller](https://www.amazon.com/Eviciv-Portable-Monitor-Display-1024X600/dp/B07L6WT77H/ref=pd_cp_147_2/144-3139777-4138749?_encoding=UTF8&pd_rd_i=B07L6WT77H&pd_rd_r=1ba9e404-3eee-445a-a3d3-017a64334a12&pd_rd_w=y7b45&pd_rd_wg=IkqzT&pf_rd_p=e44de6bb-cc27-4696-9c22-3a1bddefabbd&pf_rd_r=BZF7JJ2X5GCDY7W2J1S9&psc=1&refRID=BZF7JJ2X5GCDY7W2J1S9) than 10".

kchibisov:
I feel like we misspoke a bit here, since I was voting for 20mm x 20mm, and just followed 160mm x 90mm, so ended up in a bad metrics, since original issue was with metrics like 16mmx9mm.

NickeZ:
There are definitely android devices with screens smaller than 20x20mm.

kchibisov:
I doubt you wont run X11 on this thing, but that's why I've initially suggested DPR based filtering and not metrics based

kchibisov:
covers https://github.com/rust-windowing/winit/pull/1931

PR #1404, opened 2020-01-17
add composition event support

### LINKED ISSUE #1293: Improve IME event handling
Reference: https://github.com/servo/servo/issues/24724

When I tried to implement compositions event handling for the browser engine, I realized that current event interface doesn't have enough information about IME events.
To implement this, we need to know what is the current content in IME when the user edit something in IME. But seems `glutin::WindowEvent::KeyboardInput` doesn't have interface to get that information. It would be nice to send more information when IME status was updated.

Currently, winit submits character information when IME edit completes. But to implement compositions event handling, those information should be available when updating the status as well. In addition, winit sends charcter information for each characters (e.g. if the content in IME was "あああ", it sends 3 `ReceiveCharacter('あ')` event) . It would be nice to have send this information as a string.

I tested this on my Linux with X11, but all other platforms might have the same issues.

garasubo:
Fix #1293

Added new window events to get information about preedit status in IME: `CompositionStart`, `CompositionUpdate`, `CompositionEnd`.

This change requires x11-dl crate. I've already send a pull request for it (https://github.com/erlepereira/x11-rs/pull/107)

TODO:
- I'm not sure `ImeContextClientData` could be accessed at the same time. It might need to be wrapped with `Mutex`
- This feature only available in X11. I don't have other platform, so need to help implementing this feature for other platforms.
- We also need to provide a way to disable this feature, since some people don't want to receive these new events.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,7 +78,7 @@ wayland-client = { version = "0.23.0", features = [ "dlopen", "egl", "cursor", "
 mio = "0.6"
 mio-extras = "2.0"
 smithay-client-toolkit = "^0.6.6"
-x11-dl = "2.18.4"
+x11-dl = "2.18.5"
 percent-encoding = "2.0"
 
 [target.'cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "openbsd", target_os = "netbsd", target_os = "windows"))'.dependencies.parking_lot]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,7 +78,7 @@ wayland-client = { version = "0.23.0", features = [ "dlopen", "egl", "cursor", "
 mio = "0.6"
 mio-extras = "2.0"
 smithay-client-toolkit = "^0.6.6"
-x11-dl = "2.18.3"
+x11-dl = "2.18.4"
 percent-encoding = "2.0"
 
 [target.'cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "openbsd", target_os = "netbsd", target_os = "windows"))'.dependencies.parking_lot]
--- a/src/event.rs
+++ b/src/event.rs
@@ -235,6 +235,11 @@ pub enum WindowEvent<'a> {
         is_synthetic: bool,
     },
 
+    /// An event from IME
+    Composition {
+        event: CompositionEvent,
+    },
+
     /// The cursor has moved on the window.
     CursorMoved {
         device_id: DeviceId,
@@ -504,6 +509,14 @@ pub struct KeyboardInput {
     /// this device are not being delivered to the application, e.g. due to keyboard focus being elsewhere.
     #[deprecated = "Deprecated in favor of DeviceEvent::ModifiersChanged"]
     pub modifiers: ModifiersState,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum CompositionEvent {
+    CompositionStart(String),
+    CompositionUpdate(String),
+    CompositionEnd(String),
 }
 
 /// Describes touch-screen input state.
--- a/src/event.rs
+++ b/src/event.rs
@@ -513,9 +513,9 @@ pub struct KeyboardInput {
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub enum CompositionEvent {
-    CompositionStart,
+    CompositionStart(String),
     CompositionUpdate(String, usize),
-    CompositionEnd,
+    CompositionEnd(String),
 }
 
 /// Describes touch-screen input state.
--- a/src/event.rs
+++ b/src/event.rs
@@ -236,9 +236,7 @@ pub enum WindowEvent<'a> {
     },
 
     /// An event from IME
-    Composition {
-        event: CompositionEvent,
-    },
+    Composition { event: CompositionEvent },
 
     /// The cursor has moved on the window.
     CursorMoved {
--- a/src/event.rs
+++ b/src/event.rs
@@ -344,6 +344,7 @@ impl<'a> WindowEvent<'a> {
                 input,
                 is_synthetic,
             }),
+            Composition { event } => Some(Composition { event }),
             #[allow(deprecated)]
             CursorMoved {
                 device_id,
--- a/src/event.rs
+++ b/src/event.rs
@@ -514,9 +514,9 @@ pub struct KeyboardInput {
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub enum CompositionEvent {
-    CompositionStart(String),
-    CompositionUpdate(String),
-    CompositionEnd(String),
+    CompositionStart(),
+    CompositionUpdate(String, usize),
+    CompositionEnd(),
 }
 
 /// Describes touch-screen input state.
--- a/src/event.rs
+++ b/src/event.rs
@@ -236,7 +236,7 @@ pub enum WindowEvent<'a> {
     },
 
     /// An event from IME
-    Composition { event: CompositionEvent },
+    Composition(CompositionEvent),
 
     /// The cursor has moved on the window.
     CursorMoved {
@@ -344,7 +344,7 @@ impl<'a> WindowEvent<'a> {
                 input,
                 is_synthetic,
             }),
-            Composition { event } => Some(Composition { event }),
+            Composition(event) => Some(Composition(event)),
             #[allow(deprecated)]
             CursorMoved {
                 device_id,
@@ -513,9 +513,9 @@ pub struct KeyboardInput {
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub enum CompositionEvent {
-    CompositionStart(),
+    CompositionStart,
     CompositionUpdate(String, usize),
-    CompositionEnd(),
+    CompositionEnd,
 }
 
 /// Describes touch-screen input state.
--- a/src/platform_impl/linux/x11/event_processor.rs
+++ b/src/platform_impl/linux/x11/event_processor.rs
@@ -19,10 +19,12 @@ use crate::{
     },
     event_loop::EventLoopWindowTarget as RootELW,
 };
+use crate::platform_impl::platform::x11::ime::ImeEventReceiver;
 
 pub(super) struct EventProcessor<T: 'static> {
     pub(super) dnd: Dnd,
     pub(super) ime_receiver: ImeReceiver,
+    pub(super) ime_event_receiver: ImeEventReceiver,
     pub(super) randr_event_offset: c_int,
     pub(super) devices: RefCell<HashMap<DeviceId, Device>>,
     pub(super) xi2ext: XExtension,
@@ -1185,6 +1187,16 @@ impl<T: 'static> EventProcessor<T> {
         match self.ime_receiver.try_recv() {
             Ok((window_id, x, y)) => {
                 wt.ime.borrow_mut().send_xim_spot(window_id, x, y);
+            }
+            Err(_) => (),
+        }
+        match self.ime_event_receiver.try_recv() {
+            Ok((window, event)) => {
+                debug!("get composition event");
+                callback(Event::WindowEvent {
+                    window_id: mkwid(window),
+                    event: WindowEvent::Composition { event },
+                });
             }
             Err(_) => (),
         }
--- a/src/platform_impl/linux/x11/event_processor.rs
+++ b/src/platform_impl/linux/x11/event_processor.rs
@@ -1195,7 +1195,7 @@ impl<T: 'static> EventProcessor<T> {
                 debug!("get composition event");
                 callback(Event::WindowEvent {
                     window_id: mkwid(window),
-                    event: WindowEvent::Composition { event },
+                    event: WindowEvent::Composition(event),
                 });
             }
             Err(_) => (),
--- a/src/platform_impl/linux/x11/event_processor.rs
+++ b/src/platform_impl/linux/x11/event_processor.rs
@@ -12,6 +12,7 @@ use super::{
 
 use util::modifiers::{ModifierKeyState, ModifierKeymap};
 
+use crate::platform_impl::platform::x11::ime::ImeEventReceiver;
 use crate::{
     dpi::{PhysicalPosition, PhysicalSize},
     event::{
@@ -19,7 +20,6 @@ use crate::{
     },
     event_loop::EventLoopWindowTarget as RootELW,
 };
-use crate::platform_impl::platform::x11::ime::ImeEventReceiver;
 
 pub(super) struct EventProcessor<T: 'static> {
     pub(super) dnd: Dnd,
--- a/src/platform_impl/linux/x11/event_processor.rs
+++ b/src/platform_impl/linux/x11/event_processor.rs
@@ -12,11 +12,12 @@ use super::{
 
 use util::modifiers::{ModifierKeyState, ModifierKeymap};
 
-use crate::platform_impl::platform::x11::ime::ImeEventReceiver;
+use crate::platform_impl::platform::x11::ime::{ImeEvent, ImeEventReceiver};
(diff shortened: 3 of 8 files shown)


murarth:
This PR does not successfully build on any platform. You need to update your winit fork and rebase this branch against the current master.

garasubo:
@murarth I know that. This failure is because we need to update x11-dl crate, as mentioned in the description. If you want to try to build in your local:
1. Pull this MR in your local
2. Update Cargo.toml to use your local x11-dl

garasubo:
Only the failure is in armv7-apple-ios. It seems the issue in CI?
```
error: component 'rust-std' for target 'armv7-apple-ios' is unavailable for download for channel nightly
Sometimes not all components are available in any given nightly.
error: backtrace:
error: stack backtrace:
   0: backtrace::capture::Backtrace::new_unresolved
   1: error_chain::backtrace::imp::InternalBacktrace::new
   2: rustup::dist::manifestation::Manifestation::update
   3: rustup::dist::dist::try_update_from_dist_
   4: rustup::toolchain::Toolchain::install
   5: rustup::toolchain::Toolchain::install_from_dist
   6: rustup_init::rustup_mode::update
   7: rustup_init::rustup_mode::main
   8: rustup_init::run_rustup_inner
   9: rustup_init::main
  10: std::rt::lang_start::{{closure}}
  11: main
```

Cargo.toml
```diff
@@ -78,7 +78,7 @@ wayland-client = { version = "0.23.0", features = [ "dlopen", "egl", "cursor", "
 mio = "0.6"
 mio-extras = "2.0"
 smithay-client-toolkit = "0.6"
-x11-dl = "2.18.3"
+x11-dl = "2.18.4"
```
murarth: It's actually version `2.18.5` that's required now.

src/event.rs
```diff
@@ -235,6 +235,9 @@ pub enum WindowEvent<'a> {
         is_synthetic: bool,
     },
 
+    /// An event from IME
+    Composition { event: CompositionEvent },
```
murarth: We generally prefer tuple enum variants when there's only one field.

src/event.rs
```diff
@@ -506,6 +510,14 @@ pub struct KeyboardInput {
     pub modifiers: ModifiersState,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum CompositionEvent {
+    CompositionStart(),
+    CompositionUpdate(String, usize),
+    CompositionEnd(),
```
murarth: * `CompositionStart` should be a unit variant rather than an empty tuple variant.
* The cursor position field in `CompositionUpdate` appears to be a character-wise index. Unless there's a compelling reason not to, I think we should convert it to a byte-wise index, as that's more typical of Rust conventions.
* I think `CompositionEnd` should also contain the composition text, if possible.

src/platform_impl/linux/x11/ime/context.rs
```diff
@@ -1,41 +1,239 @@
...
+            ffi::XNPreeditDrawCallback_0.as_ptr() as *const _,
+            &preedit_callbacks.draw_callback as *const _,
             ptr::null_mut::<()>(),
         ),
     )
     .expect("XVaCreateNestedList returned NULL")
 }
 
+fn create_xim_callback(client_data: ffi::XPointer, callback: XIMProcNonnull) -> ffi::XIMCallback {
+    let mut xim_callback: ffi::XIMCallback = unsafe { MaybeUninit::uninit().assume_init() };
```
murarth: This is not a safe usage of `MaybeUninit`. Also, it appears to be totally unnecessary. The `XIMCallback` struct instance can be constructed in the usual manner.

src/platform_impl/linux/x11/ime/context.rs
```diff
@@ -44,11 +242,22 @@ impl ImeContext {
         im: ffi::XIM,
         window: ffi::Window,
         ic_spot: Option<ffi::XPoint>,
+        event_sender: ImeEventSender,
     ) -> Result<Self, ImeContextCreationError> {
+        let client_data = Box::new(ImeContextClientData {
+            window,
+            tag: 753,
```
murarth: What is the purpose of the `tag` field?

src/platform_impl/linux/x11/ime/context.rs
```diff
@@ -1,41 +1,239 @@
...
+        client_data
+            .event_sender
+            .send((
+                client_data.window,
+                CompositionUpdate(
+                    String::from_iter(client_data.text.clone()),
+                    client_data.cursor_pos,
+                ),
+            ))
+            .expect("failed to send composition update event");
```
murarth: This code generates a `CompositionUpdate` before each `CompositionEnd`, always with an empty buffer. I don't believe this is desired behavior. (No such event is generated by Firefox [in this demo](https://developer.mozilla.org/en-US/docs/Web/API/Element/compositionstart_event#Result).)

src/platform_impl/linux/x11/ime/context.rs
```diff
@@ -1,41 +1,239 @@
...
+    client_data.text.split_off(chg_range.start);
+    client_data.text.append(&mut new_chars);
+    client_data.text.append(&mut old_text_tail);
+
+    client_data
+        .event_sender
+        .send((
+            client_data.window,
+            CompositionUpdate(
+                String::from_iter(client_data.text.clone()),
```
murarth: Instead of cloning the `Vec`, we can build a `String` using `client_data.text.iter().collect()` to avoid an extra allocation.

src/platform_impl/linux/x11/ime/context.rs
garasubo: This is just for debug. Sorry, I forgot to clean it up. Will remove it.

src/platform_impl/linux/x11/ime/context.rs
garasubo: I think it depends on the IME behaviour. At least, in my environment, when I hit enter to decide the preedit string, firstly, preedit_update_callback is called to delete all preedit string, and then preedit_done_callback is called. In this case, CompositionUpdate won't be generated here.
This might be a little bit different from Firefox demo. Firefox itself uses GTK to manage those IME events, and I referred to the logic in GTK. So, this behaviour is similar to GTK. It seems Firefox fileter those event to align Web specifications.
I'm not sure what is the best way. I though that it was better to expose those events as much as possible, and each application using winit should filter them as they need. But maybe it is better to follow Web specification to provide unified interfaces.

Do you think those filtering should happen inside the library?

src/platform_impl/linux/x11/ime/context.rs
garasubo: Fixed

src/platform_impl/linux/x11/ime/context.rs
garasubo: Fixed

src/event.rs
garasubo: > CompositionStart should be a unit variant rather than an empty tuple variant.
Fixed this part.

> The cursor position field in CompositionUpdate appears to be a character-wise index. Unless there's a compelling reason not to, I think we should convert it to a byte-wise index, as that's more typical of Rust conventions.
I think the use case of cursor position is for rendering cursor in preedit string. In this senario, character-wise index will be more useful for users. 

> I think CompositionEnd should also contain the composition text, if possible.
 If we follow the Web specification way, it should be better to contain the text, but original XIM event doesn't contain the composition text. I'm not sure which is the better way.

Cargo.toml
garasubo: Fixed

src/event.rs
garasubo: Fixed

src/platform_impl/linux/x11/ime/context.rs
murarth: I see you've fixed one instance, but there are two remaining instances of the `String::from_iter` usage.

src/event.rs
murarth: > I think the use case of cursor position is for rendering cursor in preedit string. In this senario, character-wise index will be more useful for users.

This is a fair point. My concern is that character-wise indexing presents a potential source of confusion: If an IME system were to generate text containing combining characters, would a character-wise index count such combining characters or treat them as part of the base character that they modify? (I apologize if I'm using confusing or incorrect terminology; I'm not entirely familiar with IME or Unicode concepts.)

> If we follow the Web specification way, it should be better to contain the text, but original XIM event doesn't contain the composition text. I'm not sure which is the better way.

I checked the [UI events spec](https://w3c.github.io/uievents/#event-type-compositionupdate) for clarification on this and my understanding is that a `CompositionUpdate` event with an empty buffer value is generated when the user has cancelled the IME input. So, I think that, if the IME has actually been completed, a `CompositionUpdate("", 0)` event should not be generated before `CompositionEnd`. In my testing, I am receiving an empty update event immediately before the end event, after either cancelling or completing input.

> If we follow the Web specification way, it should be better to contain the text, but original XIM event doesn't contain the composition text. I'm not sure which is the better way.

I think it would be best to adhere as closely to the web spec as we can. At least, that's my tendency here as I'm unfamiliar with the use case of IME APIs and so I'd prefer to follow the guidance of others. To that end, I think we may also want to add `String, usize` fields to the `CompositionStart` event, as the web spec indicates that some IME systems may provide this information in the start event.

src/platform_impl/linux/x11/ime/context.rs
murarth: I checked the [UI events spec](https://w3c.github.io/uievents/#event-type-compositionupdate) for clarification on this and my understanding is that a `CompositionUpdate` event with an empty buffer value is generated when the user has cancelled the IME input. So, I think that, if the IME has actually been completed, a `CompositionUpdate("", 0)` event should not be generated before `CompositionEnd`. In my testing, I am receiving an empty update event immediately before the end event, after either cancelling or completing input.

src/event.rs
garasubo: > My concern is that character-wise indexing presents a potential source of confusion
That's a good point. I'm not very familiar with IME or Unicode concept either, but character-wise index might not the best way. Byte-wise index might not be the best way, but I agree that it will be less confusion.

> I think it would be best to adhere as closely to the web spec as we can.
OK. I'll revise my change to follow the web spec as much as possible.

src/platform_impl/linux/x11/ime/context.rs
garasubo: Good catch. I replaced 2 of them as well

src/platform_impl/linux/x11/ime/context.rs
garasubo: Changed. I added new states in event_processor and convert original events to events following this event s spec.

garasubo:
- Rebased my branch to resolve conflict
- Change the event structure to follow web UI specification.

src/platform_impl/linux/x11/ime/context.rs
```diff
@@ -59,22 +243,29 @@ impl ImeContext {
...
             ffi::XNInputStyle_0.as_ptr() as *const _,
-            ffi::XIMPreeditNothing | ffi::XIMStatusNothing,
+            ffi::XIMPreeditCallbacks | ffi::XIMStatusNothing,
             ffi::XNClientWindow_0.as_ptr() as *const _,
             window,
+            ffi::XNPreeditAttributes_0.as_ptr(),
+            pre_edit_attr.ptr,
             ptr::null_mut::<()>(),
         );
+        println!("set preedit call back");
```
murarth: Forgot to delete this?

src/platform_impl/linux/x11/ime/context.rs
```diff
@@ -87,18 +278,20 @@ impl ImeContext {
...
             ffi::XNInputStyle_0.as_ptr() as *const _,
-            ffi::XIMPreeditNothing | ffi::XIMStatusNothing,
+            ffi::XIMPreeditCallbacks | ffi::XIMStatusNothing,
             ffi::XNClientWindow_0.as_ptr() as *const _,
             window,
             ffi::XNPreeditAttributes_0.as_ptr() as *const _,
             pre_edit_attr.ptr,
             ptr::null_mut::<()>(),
         );
+        println!("set preedit call back 2");
```
murarth: And this?

murarth:
I think this PR is ready to merge (once those two `println!`s are gone).

I've created a new branch on winit so that we can merge this new API into `master` only once it's implemented on all platforms.

src/platform_impl/linux/x11/ime/context.rs
garasubo: Removed

src/platform_impl/linux/x11/ime/context.rs
garasubo: Removed

src/platform_impl/linux/x11/ime/context.rs
```diff
@@ -1,41 +1,216 @@
...
+) {
+    let client_data = unsafe { &mut *(client_data as *mut ImeContextClientData) };
+    let call_data = unsafe { &mut *(call_data as *mut XIMPreeditCaretCallbackStruct) };
+    client_data.cursor_pos = call_data.position as usize;
+
+    client_data
+        .event_sender
+        .send((
+            client_data.window,
+            ImeEvent::Update(client_data.text.iter().collect(), client_data.cursor_pos),
```
murarth: The cursor position field here also needs to be converted to a byte offset, correct?

I didn't notice this in my previous review because I didn't see any erroneous events generated in my testing. When is the "caret" callback actually called?

garasubo: Fix byte offset calculation.

This event will be called when you move your cusor in IME

garasubo:
The checks were failed, but the errors comes from where I didn't change
```
##[error]    --> C:\Users\runneradmin\.cargo\registry\src\github.com-1ecc6299db9ec823\js-sys-0.3.35\src\lib.rs:4898:60
     |
4898 |                 let mem = buf.unchecked_ref::<WebAssembly::Memory>();
     |                                                            ^^^^^^ this struct is private
     |
note: the struct `Memory` is defined here
    --> C:\Users\runneradmin\.cargo\registry\src\github.com-1ecc6299db9ec823\js-sys-0.3.35\src\lib.rs:3516:5
     |
3516 |     #[wasm_bindgen]
     |     ^^^^^^^^^^^^^^^
     = note: this error originates in an attribute macro (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0603]: struct `Memory` is private
##[error]    --> C:\Users\runneradmin\.cargo\registry\src\github.com-1ecc6299db9ec823\js-sys-0.3.35\src\lib.rs:4908:60
     |
4908 |                 let mem = buf.unchecked_ref::<WebAssembly::Memory>();
     |                                                            ^^^^^^ this struct is private
     |
note: the struct `Memory` is defined here
    --> C:\Users\runneradmin\.cargo\registry\src\github.com-1ecc6299db9ec823\js-sys-0.3.35\src\lib.rs:3516:5
     |
3516 |     #[wasm_bindgen]
     |     ^^^^^^^^^^^^^^^
     = note: this error originates in an attribute macro (in Nightly builds, run with -Z macro-backtrace for more info)

error: aborting due to 3 previous errors
```

garasubo:
@murarth The CI tests were passed. Could you merge it?

murarth:
Thank you for your work on this implementation, @garasubo.

A final note: I had named the new branch `composition-events` (with a trailing `s`), but I now feel that that name may be confusing and `composition-event` (no trailing `s`) is a better name. I've created the new branch `composition-event` and deleted the old one.

src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -1188,6 +1197,40 @@ impl<T: 'static> EventProcessor<T> {
             }
             Err(_) => (),
         }
+        match self.ime_event_receiver.try_recv() {
```
kaiwk, line 1200: Hi @garasubo and @murarth , I'm implementing this for Mac, should I trigger both `WindowEvent::ReceivedCharacter` and `CompositionEvent::End` at the same time, or just ignore `WindowEvent::ReceivedCharacter` during composing?

I'm not sure how XIM work, but looks like here, in `process_event`, both x event and ime event are handled?

garasubo: Thanks for get interested in this work. I don't have Mac dev environment, so that would be great.

I think we should handle both to keep compatibility, so in my implementation, both events are handled.
In XIM, IME events and receiving characters events are independent. So, I didn't touch any logic related to ReceivedCharacter event.

PR #1408, opened 2020-01-18
Fix: deadlock when requesting redraw on X11

### LINKED ISSUE #1407: Hang when window.request_redraw() called
I'm updating KAS from winit 0.20.0-alpha5 to 0.20.0. Code (with debug stuff) [here](https://github.com/dhardy/kas/commits/debug).

The example apps hang before drawing; I traced it to [`event_loop.rs:210`](https://github.com/dhardy/kas/blob/34d8498fd2e38eeeff2209a0932da17c6456f563/kas-wgpu/src/event_loop.rs#L210):
```rust
w.window.request_redraw();
```
This appears to block. (It's possible it's been called more than once before drawing; I assume this is valid from the API.)

Running on X11.

dhardy:
Fixes #1407.

Can I ask why this wasn't caught before the release? It should be obvious from any app that requests redraws on X11.

### DIFF
--- a/src/platform_impl/linux/x11/mod.rs
+++ b/src/platform_impl/linux/x11/mod.rs
@@ -395,7 +395,6 @@ impl<T: 'static> EventLoop<T> {
         let mut xev = MaybeUninit::uninit();
 
         let wt = get_xtarget(&self.target);
-        let mut pending_redraws = wt.pending_redraws.lock().unwrap();
 
         while unsafe { self.event_processor.poll_one_event(xev.as_mut_ptr()) } {
             let mut xev = unsafe { xev.assume_init() };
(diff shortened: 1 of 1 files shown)


goddessfreya:
Well, none of the glutin and winit examples caught it, at least on my system. Strikes me as a highly timing dependent thing. Anyways, thanks for the PR!

dhardy:
It is not remotely timing dependent. A Mutex was locked before starting an event handler, which allowed the user to call a method which also tried to lock that same mutex.

dhardy:
No, four of the examples use `request_redraw` and none of them hang without this patch. Maybe it's only certain event handlers that cause this to happen. :confused:

PR #1410, opened 2020-01-19
Add ability to create Icons from embedded resources on Windows

davidhewitt:
To achieve this, I needed to refactor the `Icon` internals a little so that on Windows the canonical representation is the handle rather than RGBA.

I exposed this functionality as an `IconExtWindows` platform trait.

### DIFF
--- a/src/platform/windows.rs
+++ b/src/platform/windows.rs
@@ -1,16 +1,19 @@
 #![cfg(target_os = "windows")]
 
 use std::os::raw::c_void;
+use std::path::Path;
 
 use libc;
+use winapi::shared::minwindef::WORD;
 use winapi::shared::windef::HWND;
 
 use crate::{
+    dpi::PhysicalSize,
     event::DeviceId,
     event_loop::EventLoop,
     monitor::MonitorHandle,
-    platform_impl::EventLoop as WindowsEventLoop,
-    window::{Icon, Window, WindowBuilder},
+    platform_impl::{EventLoop as WindowsEventLoop, WinIcon},
+    window::{BadIcon, Icon, Window, WindowBuilder},
 };
 
 /// Additional methods on `EventLoop` that are specific to Windows.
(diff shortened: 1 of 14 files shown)


src/platform/windows.rs
```diff
@@ -171,3 +173,30 @@ impl DeviceIdExtWindows for DeviceId {
...
     }
 }
+
+/// Additional methods on `Icon` that are specific to Windows.
+pub trait IconExtWindows: Sized {
+    /// Create an icon from a file path.
+    ///
+    /// Specify `size` to load a specific size from the file, or `None` to load the default
+    /// icon size from the file.
+    fn from_path<P: AsRef<Path>>(path: P, size: Option<(usize, usize)>) -> Result<Self, BadIcon>;
```
Osspial:
```suggestion
    /// Create an icon from a file path.
    ///
    /// Specify `size` to load a specific icon size from the file, or `None` to load the default
    /// icon size from the file.
    fn from_path<P: AsRef<Path>>(path: P, size: Option<PhysicalSize<u32>>) -> Result<Self, BadIcon>;
```

Also, what happens if the icon file doesn't contain the specified size?

src/platform/windows.rs
```diff
@@ -171,3 +173,30 @@ impl DeviceIdExtWindows for DeviceId {
...
+
+    /// Create an icon from a resource embedded in this executable or library.
+    ///
+    /// Specify `size` to load a specific size from the file, or `None` to load the default
+    /// icon size from the file.
+    fn from_resource(ordinal: WORD, size: Option<(usize, usize)>) -> Result<Self, BadIcon>;
```
Osspial: `from_path` ditto. Also, the parameter name `ordinal` is unclear - `resource_id` or something similar would be better.

src/platform_impl/windows/icon.rs
```diff
@@ -1,108 +1,162 @@
...
-        assert_eq!(pixel_count, (width * height) as usize);
-        let mut and_mask = Vec::with_capacity(pixel_count);
-        let pixels = rgba.as_mut_ptr() as *mut Pixel; // how not to write idiomatic Rust
-        for pixel_index in 0..pixel_count {
-            let pixel = unsafe { &mut *pixels.offset(pixel_index as isize) };
-            and_mask.push(pixel.a.wrapping_sub(std::u8::MAX)); // invert alpha channel
-            pixel.to_bgra();
-        }
-        assert_eq!(and_mask.len(), pixel_count);
+    pub fn from_resource(ordinal: WORD, size: Option<(usize, usize)>) -> Result<Self, BadIcon> {
```
Osspial: `size` type and `ordinal` ditto.

src/platform_impl/windows/icon.rs
```diff
@@ -1,108 +1,162 @@
...
-    #[allow(dead_code)]
-    pub fn from_path<P: AsRef<Path>>(path: P) -> Result<Self, io::Error> {
-        let wide_path: Vec<u16> = path.as_ref().as_os_str().encode_wide().collect();
+    pub fn as_raw_handle(&self) -> HICON {
+        self.inner.handle
+    }
+
+    pub fn from_path<P: AsRef<Path>>(
+        path: P,
+        size: Option<(usize, usize)>,
```
Osspial: `size` type ditto.

src/platform_impl/windows/icon.rs
```diff
@@ -1,108 +1,162 @@
...
                 hwnd,
                 winuser::WM_SETICON,
                 icon_type as WPARAM,
-                self.handle as LPARAM,
+                self.as_raw_handle() as LPARAM,
             );
         }
     }
+
+    fn _new(handle: HICON) -> Self {
```
Osspial: Could you rename this to `from_handle`?

src/platform_impl/windows/icon.rs
```diff
@@ -1,108 +1,162 @@
...
+        size: Option<(usize, usize)>,
+    ) -> Result<Self, BadIcon> {
+        let wide_path: Vec<u16> = path
+            .as_ref()
+            .as_os_str()
+            .encode_wide()
+            .chain(once(0))
+            .collect();
+        // 0 indicates that we want to use the actual width and height
+        let (width, height) = size.unwrap_or((0, 0));
```
Osspial: We want to use `LR_DEFAULTSIZE` as the default for this, not `0`. Icon files can contain multiple sizes, and if we specify `0` and the default size Windows will load a large version of the icon and scale down instead of loading the appropriate-sized icon for the system configuration.

src/platform_impl/windows/icon.rs
```diff
@@ -1,108 +1,162 @@
...
-        let pixels = rgba.as_mut_ptr() as *mut Pixel; // how not to write idiomatic Rust
-        for pixel_index in 0..pixel_count {
-            let pixel = unsafe { &mut *pixels.offset(pixel_index as isize) };
-            and_mask.push(pixel.a.wrapping_sub(std::u8::MAX)); // invert alpha channel
-            pixel.to_bgra();
-        }
-        assert_eq!(and_mask.len(), pixel_count);
+    pub fn from_resource(ordinal: WORD, size: Option<(usize, usize)>) -> Result<Self, BadIcon> {
+        // 0 indicates that we want to use the actual width and height
+        let (width, height) = size.unwrap_or((0, 0));
```
Osspial: `LR_DEFAULTSIZE` ditto.

src/platform/windows.rs
davidhewitt: From some quick experiments, it looks like windows will attempt to scale the icon if the desired size isn't present.

davidhewitt:
Thanks, have addressed all feedback and rebased on master.

Osspial:
Will merge once the CI finishes.

PR #1422, opened 2020-01-28
Fix panic at redraw event in non redraw phase

filnet:

### DIFF
--- a/src/platform_impl/windows/event_loop/runner.rs
+++ b/src/platform_impl/windows/event_loop/runner.rs
@@ -419,7 +419,8 @@ impl<T> EventLoopRunner<T> {
                 self.call_event_handler(event)
             }
             (RunnerState::New, Event::RedrawRequested(_))
-            | (RunnerState::Idle(..), Event::RedrawRequested(_)) => {
+            | (RunnerState::Idle(..), Event::RedrawRequested(_))
+            | (RunnerState::HandlingEvents, Event::RedrawRequested(_)) => {
                 self.new_events();
                 self.main_events_cleared();
                 self.call_event_handler(event);
--- a/src/platform_impl/windows/event_loop/runner.rs
+++ b/src/platform_impl/windows/event_loop/runner.rs
@@ -424,11 +424,11 @@ impl<T> EventLoopRunner<T> {
                 self.call_event_handler(event);
             }
             (_, Event::RedrawRequested(_)) => {
-                panic!("redraw event in non-redraw phase");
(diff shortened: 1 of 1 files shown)


Kethku:
I tested this by building my app targeting your PR branch and I am still getting redraw event in non-redraw phase any time I try to move the window. I don't think this fixes the problem

filnet:
@Kethku this PR also provides more information when the panic happens.

Could you reproduce the issue with this fix and report back the panic message ?

Kethku:
I just pulled the resent change and the error seems to have gone away. I am trying reverting the most recent change to confirm the issue reappears.

Kethku:
Ok it looks like on this machine I don't encounter the error which is very concerning. Remoting in to try from my other machine now.

Kethku:
Arg. I can't get the error to reproduce in any environment anymore...

filnet:
I had the same problem. Sometimes I can reproduce the issue 3 times in a row and sometimes it just won't reproduce. Very frustrating.

Kethku:
Alrighty then. I can neither confirm nor deny at this moment if this PR fixes the issue

Kethku:
I have updated my app to target this hotfix and will run with it until I find the error. I will report back in a couple of days if I have not seen it

filnet:
This PR needs to be reviewed thoroughly. Although it seems to fix the issue, it feels like a workaround of the symptom. The root cause of the issue might still be there.

One thing, for example, it that the panic cannot be reached anymore. So why keep it ?

tmpvar:
With this patch I was able to drag/resize a winit window (on Windows 10) and it did not panic.

I was able to consistently repro on master @ 4eddd1e5bc336e12ba06110f2ccadd64cadfd845 using the resizable example

```
     Running `target\debug\examples\resizable.exe`
Resizable: true
thread 'main' panicked at 'redraw event in non-redraw phase', src\platform_impl\windows\event_loop\runner.rs:427:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
error: process didn't exit successfully: `target\debug\examples\resizable.exe` (exit code: 0xc0000409, STATUS_STACK_BUFFER_OVERRUN)
```

This may not be an apples to apples comparison though as this branch is based on `0.20.0` while master is on `0.21.0`. As a test, I've rebased this branch on  4eddd1e5bc336e12ba06110f2ccadd64cadfd845 and still cannot repro the panic!

filnet:
Superseded by https://github.com/rust-windowing/winit/pull/1461

PR #1430, opened 2020-02-02
On macOS, Fix `set_simple_screen` to remember frame excluding title bar

### LINKED ISSUE #1420: macOS sends resize event including window decorations after SimpleFullscreen
It was reported to alacritty in https://github.com/alacritty/alacritty/issues/3248 that after toggling `SimpleFullscreen` and going back to normal window, the original window size is a bit larger. As you can see on a log on our issue after SimpleFullscreen, the height of a window is equal to a `height_before_toggle + 22.0`. It was discovered that setting window decorations to `none` resolves this issue completely, so I guess winit is sending resize event with window decorations included or something like that.

hatoo:

This PR fixes #1420 

The window size that restored from simple_fullscreen become bit larger because when become simple_fullscreen, it remembers the frame which contains title bar and restore it by `NSWindow::setFrame_display_` which assumes the argment does not contains title bar.

This PR fix it by remember the frame which does not contains title bar.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- On macOS, fix `set_simple_screen` to remember frame excluding title bar.
 - On macOS, fix set_minimized(true) works only with decorations.
 - On macOS, add `hide_application` to `EventLoopWindowTarget` via a new `EventLoopWindowTargetExtMacOS` trait. `hide_application` will hide the entire application by calling `-[NSApplication hide: nil]`.
 - On macOS, fix not sending ReceivedCharacter event for specific keys combinations.
--- a/src/platform_impl/macos/window.rs
+++ b/src/platform_impl/macos/window.rs
@@ -1028,7 +1028,11 @@ impl WindowExtMacOS for UnownedWindow {
 
             if fullscreen {
                 // Remember the original window's settings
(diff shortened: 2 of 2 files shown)


goddessfreya:
@kchibisov @msanders Does this resolve your issues?

kchibisov:
note: I don't have macOS machine and can't test it myself.

@msanders to simplify your testing you can apply this patch to alacritty. It make it use this winit from this PR.

```diff
diff --git a/Cargo.toml b/Cargo.toml
index 415ceb9..df4367f 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,3 +13,4 @@ incremental = false

 [patch.crates-io]
 servo-freetype-sys = { path = "servo-freetype-proxy" }
+winit = { git = "https://github.com/hatoo/winit", branch = "macos-fix-simple_fullscreen" }

```

msanders:
Tested and confirmed it fixes the resizing issue. Thank you!

It looks like the behavior with the prompt was just a fluke, although the window configuration bug still exists (can open a separate ticket for it if that would be preferred).

kchibisov:
>Tested and confirmed it fixes the resizing issue. Thank you!

Thx for testing.

>the window configuration bug still exists

What bug?

Issue #768, opened 2019-01-21
Blocking alt keys on macOS


chrisduerr:
In a terminal emulator like Alacritty, it's possible to send control characters directly to the shell using the alt and control keys. However this gets problematic with keyboard layouts that use the alt key for accessing alternative key characters. This is mainly an issue on macOS since other platforms usually use the `AltGr` key.

The concrete problem is that when alt is held down, the modified key is received, instead of the normal key. So for example Option + F will generate `ƒ` instead of `f`. In iTerm it's possible to pick between sending escapes and sending modified keys, however Alacritty currently can't offer that since winit will always send the modified keys.

Now I'm not really familiar with how this works under the hood in winit, but ideally I'd like an option in the window builder (maybe possible to change at runtime?) to prevent alt from modifying the characters, so that Option + F will still send `f`.

Alternatively it would be necessary for Alacritty to define its own mappings and ignore the `ReceivedCharacter` event, which would likely result in a bit of duplication between winit and Alacritty.

More information about this issue can be found here:
https://github.com/jwilm/alacritty/issues/2017

eraserhd:
Also jwilm/alacritty#1610

kjmph:
I caught this sight of this bug from Alacritty. See my misguided post from alacritty/alacritty#2017. I opened a pull request #1436. However, that is meant to be an example pull request, as I'm not a maintainer of winit, I'm not as familiar with how options work.

kjmph:
If anyone would like to test, the pull request went through a few rounds. I have an open pull request that fixes this issue:

https://github.com/rust-windowing/winit/pull/1449

If anyone can test to provide support for the pull request, there is a single change to alacritty:

```patch
diff --git a/alacritty/src/window.rs b/alacritty/src/window.rs
index 71327ba..7bc555a 100644
--- a/alacritty/src/window.rs
+++ b/alacritty/src/window.rs
@@ -265,7 +265,8 @@ impl Window {
             .with_title(title)
             .with_visible(false)
             .with_transparent(true)
-            .with_maximized(window_config.startup_mode() == StartupMode::Maximized);
+            .with_maximized(window_config.startup_mode() == StartupMode::Maximized)
+            .with_ignore_alt_modifier(true);

         match window_config.decorations {
             Decorations::Full => window,
```

kchibisov:
If someone is going to test it on alacritty they could apply the following patch to use mentioned PR in addition to the patch above.

```diff
[...]

PR #1436, opened 2020-02-04
Fix for dropping diacritics from downstream events

kjmph:
This is meant to be a example pull request. This is most likely an
option that the consumer of winit should specify. If that option was
ALT AS META, perhaps a sythetic escape could be inserted in the event
stream?


kjmph:
This references #768.

src/platform_impl/macos/event.rs
```diff
@@ -89,6 +89,106 @@ pub fn char_to_keycode(c: char) -> Option<VirtualKeyCode> {
...
+        VirtualKeyCode::Apostrophe => '\'',
+        VirtualKeyCode::Comma if m.shift() => '<',
+        VirtualKeyCode::Comma => ',',
+        VirtualKeyCode::Period if m.shift() => '>',
+        VirtualKeyCode::Period => '.',
+        VirtualKeyCode::Slash if m.shift() => '?',
+        VirtualKeyCode::Slash => '/',
+        VirtualKeyCode::Grave if m.shift() => '~',
+        VirtualKeyCode::Grave => '`',
+        _ => return None,
```
kchibisov: This is not the right way to do this. You're assuming QUERTY here for US keyboard. However there're keyboard for which this isn't true. Hardcoding isn't an option.

src/platform_impl/macos/view.rs
```diff
@@ -632,44 +632,91 @@ extern "C" fn key_down(this: &Object, _sel: Sel, event: id) {
...
+                        true
+                    }
+                };
+
+                if pass_along {
+                    // Some keys (and only *some*, with no known reason) don't trigger `insertText`, while others do...
+                    // So, we don't give repeats the opportunity to trigger that, since otherwise our hack will cause some
+                    // keys to generate twice as many characters.
+                    let array: id = msg_send![class!(NSArray), arrayWithObject: event];
+                    let _: () = msg_send![this, interpretKeyEvents: array];
```
kchibisov: It seems to me that you're manually trying to regenerate things. This isn't the right way to solve this issue. IIRC it should be done with masks like iterm is doing here https://github.com/gnachman/iTerm2/blob/30e34ee31060e61787dc6e1968a487ab05d0aff5/sources/iTermKeyboardHandler.h#L16 . I guess tracking its code is a right way to go.

src/platform_impl/macos/event.rs
kjmph: Yes, you are right, it is assuming QWERTY. I'm hoping to get attention to the issues Mac users face.

src/platform_impl/macos/view.rs
kjmph: Excellent! I'll research more about that approach. Thanks for the pointers.

kchibisov: I don't have macOS systems tbh. But It's good to take a look what other terminals are doing here, since they should also solve this problem somehow.

kjmph: That's fine if you aren't using macOS, I'll try and represent a dev env. Note, if I follow the `winit` code correctly, it is translating the scan code to the virtual key-code since the retrieved character isn't in `char_to_keycode`. The scan code is supposed to independent of placement, yes? So, the big issue is with shift modifier? I'll still look into masks, I'm just making sure I follow the dependency on keyboard layouts.

kchibisov: scancode is bound to place IIRC, this mapping is from Event.h(apple thing) though. But I could be wrong here, however you can verify that by using `dvorak` layout, I guess. So translating ascii chars to virtual key codes is the right way to go. However scancode for `return`, `backspace` and friends is likely identical regardless of the layout. Yes, shift modifier could be a problem here and for `Option` mode you're getting strange thing in the end anyway.

The problem is that we're getting strange characters from `Alt + a` input for e.g. so we should find out how to filter out them. I feel like Masks should do something about how macOS sending events to us. So the only way to go is to understand what this code in iterm is about and try to and clearly explain/port here.

kjmph: Okay, `Dvorak` works fine. However, `British - PC` has a different enough mapping to exhibit that once the `Shift` modifier is pressed, the incorrect key is sent to upstream. I will see if I can fix it.

kjmph: Okay, found it in Apple's documentation:

https://developer.apple.com/documentation/appkit/nsevent/1524605-charactersignoringmodifiers

Which makes much more sense.. Now, looking back at the code, I see that this can be accomplished very easily like so:

```patch
diff --git a/src/platform_impl/macos/view.rs b/src/platform_impl/macos/view.rs
index 741ecb4..7889898 100644
--- a/src/platform_impl/macos/view.rs
+++ b/src/platform_impl/macos/view.rs
@@ -510,10 +510,12 @@ extern "C" fn insert_text(this: &Object, _sel: Sel, string: id, _replacement_ran
         state.is_key_down = true;

         // We don't need this now, but it's here if that changes.
-        //let event: id = msg_send![NSApp(), currentEvent];
+        let event: id = msg_send![NSApp(), currentEvent];
+        let ev_mods = event_mods(event);
+        let raw_string = get_characters(event, ev_mods.alt());

         let mut events = VecDeque::with_capacity(characters.len());
-        for character in string.chars().filter(|c| !is_corporate_character(*c)) {
+        for character in raw_string.chars().filter(|c| !is_corporate_character(*c)) {
             events.push_back(EventWrapper::StaticEvent(Event::WindowEvent {
                 window_id: WindowId(get_window_id(state.ns_window)),
                 event: WindowEvent::ReceivedCharacter(character),
@@ -624,7 +626,8 @@ extern "C" fn key_down(this: &Object, _sel: Sel, event: id) {
         let window_id = WindowId(get_window_id(state.ns_window));
-        let characters = get_characters(event, false);
+        let ev_mods = event_mods(event);
+        let characters = get_characters(event, ev_mods.alt());

         state.raw_characters = Some(characters.clone());
```

Of course, it still has to be cleaned up, because I don't understand the insertText event fully. It seems there is an interaction between keyDown and insertText, and it is a bit odd. Plus, we don't want to overwrite the string being passed in to insertText, right?

So many questions.. However, the above patch does produce the correct keys, and works as well with the `British - PC` layout. Plus, it appears key repeats work just fine, this allowed me to turn off most of my alacritty config, which is a big plus.

kchibisov: >Plus, we don't want to overwrite the string being passed in to insertText, right?

I have no idea about this one, so someone who knows macOS should answer, I guess. But generally when you're not sure about something like that, I'd recommend to test and see what happens.

>Okay, found it in Apple's documentation:

Well, it says the thing we actually want in alacritty AFAICS. However we should keep old behavior under option, I guess, but let just try to implement something that works for us, and hide new behavior under option later during impl.

>Plus, it appears key repeats work just fine, this allowed me to turn off most of my alacritty config, which is a big plus.

Could you verify that some keys don't create twice as much events as they should on repeat? I'd recommend to press just all of them. I just wonder why you've mentioned key repeat here.

>So many questions.. However, the above patch does produce the correct keys, and works as well with the British - PC layout.

I'll assume that it just produces correct keys without mods or with shift, and no strange unicode thing when Alt is involved, right?. Have you verified Alt + Shift behavior.

Could you provide me `--print-events` log from alacritty during you pressing `s, shift + s, shift + alt + s, alt + s`? Just want to check.

kjmph: > > Plus, we don't want to overwrite the string being passed in to insertText, right?
> 
> I have no idea about this one, so someone who knows macOS should answer, I guess. But generally when you're not sure about something like that, I'd recommend to test and see what happens.

I have tested it, it works well. I support I mean more particularly, why use insertText at all? It causes the control flow to bounce around due to some key presses not triggering insertText. Copy and paste still work without insertText. (I tested this statement, I enjoy removing that pass_along style code.)

> > Okay, found it in Apple's documentation:
> 
> Well, it says the thing we actually want in alacritty AFAICS. However we should keep old behavior under option, I guess, but let just try to implement something that works for us, and hide new behavior under option later during impl.

Yes, I agree. I think we should close this pull request and open a new one with options to get this behavior. I'll go read how to get options into winit.

> > Plus, it appears key repeats work just fine, this allowed me to turn off most of my alacritty config, which is a big plus.
> 
> Could you verify that some keys don't create twice as much events as they should on repeat? I'd recommend to press just all of them. I just wonder why you've mentioned key repeat here.

At one point, I had key repeats not working. However, this is no longer reproducible. It may have been my hacks at the time while learning the project.

> > So many questions.. However, the above patch does produce the correct keys, and works as well with the British - PC layout.
> 
> I'll assume that it just produces correct keys without mods or with shift, and no strange unicode thing when Alt is involved, right?. Have you verified Alt + Shift behavior.
> 
> Could you provide me `--print-events` log from alacritty during you pressing `s, shift + s, shift + alt + s, alt + s`? Just want to check.

Absolutely. Trying to attach is failing.. let me see..

kjmph: Couldn't upload the file for some reason. You can read it here though:

https://gist.github.com/kjmph/f9b0b84e6ad9d62e1a329364a004f059

kjmph: Wow. `Alt-i` is very elucidating. It doesn't trigger `insert_text` since it isn't a printable character. Yet, when the event triggers `key_down` the `state` object doesn't have the `is_key_down` set to true, so it calls `interpretKeyEvents` and places the event back in the queue. Apple then interprets that as the circumflex diacritic, and will use it to append to the next character inserted. I went through the `winit` bugs, and there were a lot that referenced this code. By removing the `insert_text` and `do_command_by_selector`, a lot of code is cleaned up. I'm unable to reproduce any problems with this. I'll open a pull request showing these changes.

kjmph: I can't seem to invite you to review @kchibisov. Here is the improved pull request: https://github.com/rust-windowing/winit/pull/1449

PR #1449, opened 2020-02-10
Fix Alt modifier on macOS

kjmph:
The complications between insertText and keyDown were causing
difficulties, especially when diacritics are used. This commit cleans
out the duplication of events and the state that needs to be captured
between event handlers.


src/platform/macos.rs
```diff
@@ -131,6 +131,7 @@ pub trait WindowBuilderExtMacOS {
     /// Build window with `resizeIncrements` property. Values must not be 0.
     fn with_resize_increments(self, increments: LogicalSize<f64>) -> WindowBuilder;
     fn with_disallow_hidpi(self, disallow_hidpi: bool) -> WindowBuilder;
+    fn with_ignore_alt_modifier(self, ignore_alt_modifier: bool) -> WindowBuilder;
```
kchibisov: I feel like it should be `with_option_as_alt`, since it's more clear to the users. Getting inspiration from other apps could be a good idea.

src/platform_impl/macos/view.rs
```diff
@@ -489,79 +482,6 @@ extern "C" fn first_rect_for_character_range(
     }
 }
 
-extern "C" fn insert_text(this: &Object, _sel: Sel, string: id, _replacement_range: NSRange) {
-    trace!("Triggered `insertText`");
-    unsafe {
-        let state_ptr: *mut c_void = *this.get_ivar("winitState");
-        let state = &mut *(state_ptr as *mut ViewState);
```
kchibisov: I'm not sure about this change tbh. However I can't prove it myself that this thing is requited. So I need a time to go through the history to understand what this thing is about. I'd suggest you to do the same, since there should be a reason for it.

src/platform_impl/macos/view.rs
```diff
@@ -489,79 +482,6 @@ extern "C" fn first_rect_for_character_range(
...
-            }));
-        }
-
-        AppState::queue_events(events);
-    }
-    trace!("Completed `insertText`");
-}
-
-extern "C" fn do_command_by_selector(this: &Object, _sel: Sel, command: Sel) {
-    trace!("Triggered `doCommandBySelector`");
```
kchibisov: Same as for insert text here.

kchibisov:
Also could you compare `ReceivedCharacter` from `Control + s` on your PR to the one without it?

kjmph:
Do you mean this?

Without:

```
[2020-02-10 00:01] [INFO] glutin event: WindowEvent { window_id: WindowId(Id(140689493250816)), event: ReceivedCharacter('\u{13}') }
```

With:

```
[2020-02-10 00:01] [INFO] glutin event: WindowEvent { window_id: WindowId(Id(140352641466896)), event: ReceivedCharacter('\u{13}') }
```

src/platform/macos.rs
kjmph: Yes, we should make this flag as descriptive as possible.

src/platform_impl/macos/view.rs
kjmph: I've been going through old issues, and seeing that I can't reproduce any of them. I'll keep digging.

src/platform_impl/macos/view.rs
kjmph: Yes, I will continue to dig.

kchibisov:
>Do you mean this?

Yeah, I was a bit sleepy during reading your log from https://github.com/rust-windowing/winit/pull/1436#discussion_r376818844 , I've somehow assumed that you pressed `Ctrl + s` there, but it was `Ctrl + c`.

chrisduerr:
Especially for Alacritty's live config reload it would be interesting to change this at runtime. Do you think it would be possible to add a `set_ignore_alt_modifier` on Window instead of just `with_ignore_alt_modifier` on the WindowBuilder @kjmph?

kjmph:
Very interesting, as the boolean `ignore_alt_modifier` is checked during event processing, then yes it can be moved to a live configuration parameter. If you want me to attempt that, it may have to wait a day before I have time to modify the patch.

chrisduerr:
I think that would be very worthwhile and is not uncommon with window parameters. Especially for Alacritty it would be excellent.

vbogaevsky:
@kjmph are you still working on this PR?

kjmph:
Yes, I'll pull it together, the day job is busy at the moment. I'll dedicate time tonight.

kjmph:
Okay, it works now.. Shall I commit the changes to winit to this pull request? I know there were a couple of users testing this pull request, just curious what is the process around here.

The changes are a bit more extensive to alacritty, as this needs to be processed by the `NSView` which is attached to the `Window`. Previously, this change leveraged `WindowBuilder`.

Here is the patch to alacritty:

```patch
diff --git a/alacritty.yml b/alacritty.yml
index e552166..63c11b5 100644
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -347,6 +347,9 @@
 # Send ESC (\x1b) before characters when alt is pressed.
 #alt_send_esc: true
 
+# (macOS Only) When Option is pressed treat as alt modifier. Don't send diacritical marks.
+#option_as_alt: true
+
 #debug:
   # Display the time it takes to redraw each frame.
   #render_timer: false
diff --git a/alacritty/src/event.rs b/alacritty/src/event.rs
index cd4f82a..b22e3f1 100644
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -504,6 +504,10 @@ impl<N: Notify + OnResize> Processor<N> {
                             processor.ctx.display_update_pending.font = Some(font);
                         }
 
+                        if cfg!(target_os = "macos") && processor.ctx.config.option_as_alt() != config.option_as_alt() {
+                            processor.ctx.window.set_option_as_alt(config.option_as_alt());
+                        }
+
                         *processor.ctx.config = config;
 
                         processor.ctx.terminal.dirty = true;
diff --git a/alacritty/src/window.rs b/alacritty/src/window.rs
index 71327ba..6cd9041 100644
--- a/alacritty/src/window.rs
+++ b/alacritty/src/window.rs
@@ -151,6 +151,11 @@ impl Window {
         // Set OpenGL symbol loader. This call MUST be after window.make_current on windows.
         gl::load_with(|symbol| windowed_context.get_proc_address(symbol) as *const _);
 
+        // get_platform_window could be invoked with Config instead of
+        // WindowConfig if this call is misplaced.
+        #[cfg(target_os = "macos")]
+        windowed_context.window().set_option_as_alt(config.option_as_alt());
+
         // On X11, embed the window inside another if the parent ID has been set
         #[cfg(not(any(target_os = "macos", windows)))]
         {
@@ -356,6 +361,11 @@ impl Window {
         self.window().set_simple_fullscreen(simple_fullscreen);
     }
 
+    #[cfg(target_os = "macos")]
+    pub fn set_option_as_alt(&mut self, option_as_alt: bool) {
+        self.window().set_option_as_alt(option_as_alt);
+    }
+
     #[cfg(not(any(target_os = "macos", target_os = "windows")))]
     pub fn wayland_display(&self) -> Option<*mut c_void> {
         self.window().wayland_display()
diff --git a/alacritty_terminal/src/config/mod.rs b/alacritty_terminal/src/config/mod.rs
index fd049af..a9dabc4 100644
--- a/alacritty_terminal/src/config/mod.rs
+++ b/alacritty_terminal/src/config/mod.rs
@@ -117,6 +117,10 @@ pub struct Config<T> {
     #[serde(default, deserialize_with = "failure_default")]
     alt_send_esc: DefaultTrueBool,
 
+    /// Option on macOS is treated as Alt modifier
+    #[serde(default, deserialize_with = "failure_default")]
+    option_as_alt: DefaultTrueBool,
+
     /// Shell startup directory
     #[serde(default, deserialize_with = "option_explicit_none")]
     pub working_directory: Option<PathBuf>,
@@ -197,6 +201,12 @@ impl<T> Config<T> {
         self.alt_send_esc.0
     }
 
+    /// Option on macOS is treated as Alt modifier
+    #[inline]
+    pub fn option_as_alt(&self) -> bool {
+        self.option_as_alt.0
+    }
+
     /// Keep the log file after quitting Alacritty
     #[inline]
     pub fn persistent_logging(&self) -> bool {
```

kjmph:
Note, for reference, here is the commit to modify this branch to the new config style:

```patch
diff --git a/src/platform/macos.rs b/src/platform/macos.rs
index c600d37..2cd10a1 100644
--- a/src/platform/macos.rs
+++ b/src/platform/macos.rs
@@ -56,6 +56,14 @@ pub trait WindowExtMacOS {
     /// And allows the user to have a fullscreen window without using another
     /// space or taking control over the entire monitor.
+
+    /// Toggles the `Option` key being interpretted as an `Alt` modifier.
+    ///
+    /// This will ignore diacritical marks and accent characters from
+    /// being processed as received characters. Instead, the input
+    /// device's raw character will be placed in event queues with the
+    /// Alt modifier set.
+    fn set_option_as_alt(&self, option_as_alt: bool);
 }
 
 impl WindowExtMacOS for Window {
@@ -83,6 +91,11 @@ impl WindowExtMacOS for Window {
     fn set_simple_fullscreen(&self, fullscreen: bool) -> bool {
         self.window.set_simple_fullscreen(fullscreen)
     }
+
+    #[inline]
+    fn set_option_as_alt(&self, option_as_alt: bool) {
+        self.window.set_option_as_alt(option_as_alt)
+    }
 }
 
 /// Corresponds to `NSApplicationActivationPolicy`.
@@ -131,7 +144,6 @@ pub trait WindowBuilderExtMacOS {
     /// Build window with `resizeIncrements` property. Values must not be 0.
     fn with_resize_increments(self, increments: LogicalSize<f64>) -> WindowBuilder;
     fn with_disallow_hidpi(self, disallow_hidpi: bool) -> WindowBuilder;
-    fn with_ignore_alt_modifier(self, ignore_alt_modifier: bool) -> WindowBuilder;
 }
 
 impl WindowBuilderExtMacOS for WindowBuilder {
@@ -191,12 +203,6 @@ impl WindowBuilderExtMacOS for WindowBuilder {
         self.platform_specific.disallow_hidpi = disallow_hidpi;
         self
     }
-
-    #[inline]
-    fn with_ignore_alt_modifier(mut self, ignore_alt_modifier: bool) -> WindowBuilder {
-        self.platform_specific.ignore_alt_modifier = ignore_alt_modifier;
-        self
-    }
 }
 
 /// Additional methods on `MonitorHandle` that are specific to MacOS.
diff --git a/src/platform_impl/macos/view.rs b/src/platform_impl/macos/view.rs
index 3a48991..6857d7f 100644
--- a/src/platform_impl/macos/view.rs
+++ b/src/platform_impl/macos/view.rs
@@ -31,7 +31,7 @@ use crate::{
         ffi::*,
         util::{self, IdRef},
         window::get_window_id,
-        PlatformSpecificWindowBuilderAttributes, DEVICE_ID,
+        DEVICE_ID,
     },
     window::WindowId,
 };
@@ -65,10 +65,7 @@ impl ViewState {
     }
 }
 
-pub fn new_view(
-    ns_window: id,
-    pl_attribs: &PlatformSpecificWindowBuilderAttributes,
-) -> (IdRef, Weak<Mutex<CursorState>>) {
+pub fn new_view(ns_window: id) -> (IdRef, Weak<Mutex<CursorState>>) {
     let cursor_state = Default::default();
     let cursor_access = Arc::downgrade(&cursor_state);
     let state = ViewState {
@@ -77,7 +74,7 @@ pub fn new_view(
         ime_spot: None,
         tracking_rect: None,
-        ignore_alt_modifier: pl_attribs.ignore_alt_modifier,
+        ignore_alt_modifier: false,
     };
     unsafe {
@@ -101,6 +98,12 @@ pub unsafe fn set_ime_position(ns_view: id, input_context: id, x: f64, y: f64) {
     let _: () = msg_send![input_context, invalidateCharacterCoordinates];
 }
 
+pub unsafe fn set_option_as_alt(ns_view: id, option_as_alt: bool) {
+    let state_ptr: *mut c_void = *(*ns_view).get_mut_ivar("winitState");
+    let state = &mut *(state_ptr as *mut ViewState);
+    state.ignore_alt_modifier = option_as_alt;
+}
+
 struct ViewClass(*const Class);
 unsafe impl Send for ViewClass {}
 unsafe impl Sync for ViewClass {}
diff --git a/src/platform_impl/macos/window.rs b/src/platform_impl/macos/window.rs
index 9577511..7412033 100644
--- a/src/platform_impl/macos/window.rs
+++ b/src/platform_impl/macos/window.rs
@@ -23,7 +23,7 @@ use crate::{
         monitor::{self, MonitorHandle, VideoMode},
         util::{self, IdRef},
         view::CursorState,
-        view::{self, new_view},
+        view::{self, new_view, set_option_as_alt},
         window_delegate::new_delegate,
         OsError,
     },
@@ -95,7 +95,7 @@ unsafe fn create_view(
     ns_window: id,
     pl_attribs: &PlatformSpecificWindowBuilderAttributes,
 ) -> Option<(IdRef, Weak<Mutex<CursorState>>)> {
-    let (ns_view, cursor_state) = new_view(ns_window, pl_attribs);
+    let (ns_view, cursor_state) = new_view(ns_window);
     ns_view.non_nil().map(|ns_view| {
         if !pl_attribs.disallow_hidpi {
             ns_view.setWantsBestResolutionOpenGLSurface_(YES);
@@ -1088,6 +1088,13 @@ impl WindowExtMacOS for UnownedWindow {
             }
         }
     }
+
+    #[inline]
+    fn set_option_as_alt(&self, option_as_alt: bool) {
+        unsafe {
+            set_option_as_alt(*self.ns_view, option_as_alt);
+        }
+    }
 }
 
 impl Drop for UnownedWindow {
```

kjmph:
Note, `ignore_alt_modifier` in the internal state is set to false by default. All macOS users of winit would see no behavioral differences with this change unless they called `set_option_as_alt` on the `Window`. This may be better for all downstream users of winit.

kjmph:
As the downstream issue reported successful testing, I'll push the changes to this branch now.

kjmph:
Is there anything I can do to advance this?

kchibisov:
well, we're waiting on macOS winit maintainers to take a look at this code. I'd recommend to test IME/virtual keyboards, since it seems to me that `insert_text` was purely for them.

kjmph:
Ok! That sounds good. I'll test that.

kjmph:
I tested with a Virtual Keyboard, and it appears to send proper key codes. However, I do notice with Input modified to Chinese, things like popups suggesting character combinations don't show up in Alacritty. This occurs irrespective of this patch. I'd still love to hear if this patch works for Chinese/Cangjie users (or any input modifier user), yet it produces valid output.

kjmph:
I merged against master, and rebuilt Alacritty. Still continues to work great for me here. (I've now moved through a few different keyboards as my setup evolved.) What can I do to help get this merged?

src/platform_impl/macos/view.rs
```diff
@@ -100,6 +98,12 @@ pub unsafe fn set_ime_position(ns_view: id, input_context: id, x: f64, y: f64) {
     let _: () = msg_send![input_context, invalidateCharacterCoordinates];
 }
 
+pub unsafe fn set_option_as_alt(ns_view: id, option_as_alt: bool) {
+    let state_ptr: *mut c_void = *(*ns_view).get_mut_ivar("winitState");
+    let state = &mut *(state_ptr as *mut ViewState);
+    state.ignore_alt_modifier = option_as_alt;
```
chrisduerr, line 104: This doesn't make sense to me. Why are alt modifiers ignored when option is treated as alt? That seems opposite to what would be logical.

src/platform/macos.rs
```diff
@@ -56,6 +56,14 @@ pub trait WindowExtMacOS {
...
     /// space or taking control over the entire monitor.
     fn set_simple_fullscreen(&self, fullscreen: bool) -> bool;
+
+    /// Toggles the `Option` key being interpretted as an `Alt` modifier.
+    ///
+    /// This will ignore diacritical marks and accent characters from
+    /// being processed as received characters. Instead, the input
+    /// device's raw character will be placed in event queues with the
+    /// Alt modifier set.
+    fn set_option_as_alt(&self, option_as_alt: bool);
```
chrisduerr, line 66: Why was the `with_option_as_alt` configuration removed? I was suggesting both since that seems to be the most common scenario for features like these.

src/platform_impl/macos/view.rs
kjmph: I think there is a terminology clash here. According to Apple documentation the event has the property charactersIgnoringModifiers, which is the raw character being pressed. See more here: https://developer.apple.com/documentation/appkit/nsevent/1524605-charactersignoringmodifiers

However, the ev_mods are associated with the window event. And the upper layers will interpret the Option key press as an Alt modifier, and thus, if `alt_send_esc` is true, will send the escape character with the raw character. So, the flag is named `ignore_alt_modifier` as an indicator that the raw character needs to be retrieved. We can change this name to `ignore_option_modifier` or `ignore_diacritical_marks` if you think that any of those are clearer?

src/platform/macos.rs
kjmph: I must have misunderstood this comment? https://github.com/rust-windowing/winit/pull/1449#issuecomment-584730213

You would like the ability for WindowBuilder and the configuration to modify this behavior? Okay, I can bring that back.

src/platform_impl/macos/view.rs
chrisduerr: I think it's probably better to have a more descriptive name and state exactly what the modifier should do. The option key is alt in both cases, plenty of platforms support alternative keyboard input with alt (it's alternate after all) and raw character modification is obviously necessary to. So instead of just saying that "option is alt" which is not clear at all, we should state what the option key actually does and do the same for the other variable here.

src/platform/macos.rs
chrisduerr: Yes. You want to be able to do it at startup, but also modify it later at any time. So we need both.

chrisduerr:
@kjmph Are you willing in finishing this off? Since I can now review your changes officially we should be able to push things through relatively quickly. Otherwise I'd just take this PR over if you wouldn't mind.

kjmph:
Thanks for the ping, I've had a bit on my plate with the day job. It should be pretty straight-forward to bring back the original WindowBuilder code snippet. If you have the time, it would be awesome if you could cover me on this. Note, I've been running this patchset for quite a while, and it works great. Daily usage of Alacritty..

kjmph:
@chrisduerr, I updated to master and brought back the Platform specific WindowBuilder attributes. I'm going and updating the Alacritty patches now. However, I'm unable to build Alacritty's master branch, as:

```
error[E0425]: cannot find value `Add` in this scope
   --> alacritty/src/config/bindings.rs:524:9
    |
524 |         Add,    ModifiersState::LOGO; Action::IncreaseFontSize;
    |         ^^^ not found in this scope

error: aborting due to previous error
```

This is unrelated to this patch, I believe. It seems to be something with `glutin::event::VirtualKeyCode`? I'll poke at it a bit.

kjmph:
Ah, it appears Alacritty can't build with Winit's master branch (irrespective of these changes).. This is the first I've encountered that. What do you suggest I do?

kjmph:
Sorry to spam the thread, it looks like according to this commit: https://github.com/rust-windowing/winit/commit/d612a1b5a10cc63413b0d7026eedcdab78d2d181 The virtual key code `Add` is prefixed as `VirtualKeyCode::NumpadAdd`. Alacritty continues to build now, I'm running against master again.

chrisduerr:
I'll get around to looking into this eventually myself, though because of some of the major changes in the PR I need to take the time to setup things like IME. And since I don't run macOS unless I'm at work, that might take a bit.

kjmph:
Absolutely @chrisduerr. Let me know if I can help again. I think you want to make a pass at some of the naming, I'm not sure it is a 100% match for the project. Really, just want a flag that ignores diacritical marks from inputs and uses raw characters.

jkp:
How are we getting on with this patch-set...its few months old now but it would be fantastic to get this in if it fixes the issue. The existing workarounds are somewhat limited so a comprehensive solution would be amazing.

kjmph:
From my perspective (dev) I've completed all the asks from the maintainers. The patch runs really well for me, I'm unable to expose any bugs, and there are two ways to enable ignoring diacritical marks (window building, and config updates). I feel like it is solidified minus any naming changes the maintainers would like to make.

chrisduerr:
It's waiting on a review of at least me and a macOS maintainer. I'm not sure if any of the macOS maintainers are available though.

kjmph:
I think it is a good opportunity to fix this issue, as a number of macOS devs are frustrated that Big Sur still doesn't have 24bit color support in Terminal.app. Always nice to have a reason to try switching, and stay for the awesomeness that is Alacritty. :)

src/platform/macos.rs
chrisduerr: This should still be changed.

chrisduerr (changes requested):
This breaks IME.

chrisduerr:
Besides IME, this also breaks applications that setup global shortcuts like `skhdrc`.

src/platform/macos.rs
kjmph: Okay, I'll make the change to `with_option_as_alt`

kjmph:
Thanks for the update, any sort of reproducible test case we can use to isolate the problem?

chrisduerr:
I mean just run any kind of IME and try to input text. It didn't work at all for me.

kjmph:
I tried the virtual keyboards builtin to macOS for both Chinese and Japanese. They functioned when tested, at the time. It does appear that it is broken now. Will investigate why.

saysjonathan:
IME appears to work for me on macOS 11.1 with this change merged against current master, [38fcceb](https://github.com/rust-windowing/winit/tree/38fccebe1fbc4226c75d6180e5317bd93c024951). Also tried Japanese and Chinese. I have not yet tried `skhd`.

Is there any help needed to continue progress?

saysjonathan:
`skhd` is working for me without issue with @kjmph's branch against [38fcceb](https://github.com/rust-windowing/winit/tree/38fccebe1fbc4226c75d6180e5317bd93c024951).

src/platform_impl/macos/view.rs
ArturKovacs: To add to @chrisduerr's feedback, I'd like to suggest a specific alternative name: `alt_ignored_in_text` this naming could be used throughout, e.g. the window builder would have `with_alt_ignored_in_text`.

ArturKovacs (changes requested):

src/platform_impl/macos/view.rs
chrisduerr: @ArturKovacs What exactly is "alt ignored in text" supposed to mean?

ArturKovacs: It's supposed to mean that text coming from the keyboard (ie `ReceivedCharacters`) will not be affected by <kbd>alt</kbd>. In other words, text ignores the <kbd>alt</kbd> key.

chrisduerr: I think assuming that people will make the connection between "text" and the ReceivedCharacter event is a bit much. I certainly wasn't able to.

ArturKovacs: Consider accompanying your critiques with suggested solutions. In my opinion that speeds up such discussions.

And what do you think about `alt_ignored_in_chars`?

chrisduerr: You can just call it `alt_dead_keys` or whatever. But style questions like these are really irrelevant as long as the functionality doesn't even work yet.

kjmph:
I have minimal time to keep working on this. I was trying to start conversation amongst some of the MacOS developers, because I'm not a mac dev. For what it is worth, the official released binary for Alacritty also doesn't work with virtual keyboards, as it passes through the arrow keys.

https://user-images.githubusercontent.com/1075649/105254890-844bbf00-5b50-11eb-89d0-a9bf438450ab.mov

Contrast that to Terminal:

https://user-images.githubusercontent.com/1075649/105254944-9ded0680-5b50-11eb-8a11-40c88869ffd6.mov

I'll try to find some time, yet I have read through all the public Apple developer documentation and an answer doesn't look clear to me. It is highly possible that I just haven't spent a long enough time to grok how to ignore diacritical marks in one case, and handle virtual keyboard characters in another case.

ArturKovacs:
> I have minimal time to keep working on this.

@chrisduerr earlier [offered to take over](https://github.com/rust-windowing/winit/pull/1449#issuecomment-690627032) this PR. I'm guessing that the offer still stands and I think they'd be more than happy to take over if you don't feel like you have sufficient time working on this. Is that correct @chrisduerr? Otherwise I can take over if you'd like that @kjmph.

chrisduerr:
Yes, I'm not into macOS development either but this issue causes a lot of trouble downstream. So the sooner this is resolved the less work I have to put into support tickets, so there's definitely some time I can invest to resolve that. Of course every time I pick up one thing I can't do another which is why outside contributions are always very helpful.

chrisduerr:
I've decided to take a little peak at what one would have to do to get this working and have some questions about this implementation.

Looking at iTerm, it seems like it does some special checking for the modifier key and skips the `insert_text` functionality should the modifier be ignored:

https://gitlab.com/gnachman/iterm2/-/blob/master/sources/iTermStandardKeyMapper.m#L435
https://gitlab.com/gnachman/iterm2/-/blob/master/sources/iTermKeyboardHandler.m#L252

Winit already seems to be doing something similar with key repetition:

https://github.com/rust-windowing/winit/blob/master/src/platform_impl/macos/view.rs#L673

So is there a reason why this part of winit's code wasn't extended, to handle unmodified alt keys without propagation to `insert_text`? I've tried it out myself in a small local patch and it doesn't exhibit the same IME issues while still apparently removing the alt modifications:

<details>

<summary>Patch</summary>

```diff
diff --git a/examples/window.rs b/examples/window.rs
index 783578de..e5b01b83 100644
--- a/examples/window.rs
+++ b/examples/window.rs
@@ -1,6 +1,7 @@
+use winit::platform::macos::WindowExtMacOS;
 use simple_logger::SimpleLogger;
 use winit::{
-    event::{Event, WindowEvent},
+    event::{Event, WindowEvent, ElementState, VirtualKeyCode, KeyboardInput},
     event_loop::{ControlFlow, EventLoop},
     window::WindowBuilder,
 };
@@ -14,6 +15,7 @@ fn main() {
         .with_inner_size(winit::dpi::LogicalSize::new(128.0, 128.0))
         .build(&event_loop)
         .unwrap();
+    let mut option_alternative_input = true;
 
     event_loop.run(move |event, _, control_flow| {
         *control_flow = ControlFlow::Wait;
@@ -24,6 +26,25 @@ fn main() {
                 event: WindowEvent::CloseRequested,
                 window_id,
             } if window_id == window.id() => *control_flow = ControlFlow::Exit,
+            Event::WindowEvent { event, .. } => match event {
+                WindowEvent::KeyboardInput {
+                    input:
+                        KeyboardInput {
+                            virtual_keycode: Some(virtual_code),
+                            state,
+                            ..
+                        },
+                    ..
+                } => match (virtual_code, state) {
+                    (VirtualKeyCode::X, ElementState::Pressed) => {
+                        option_alternative_input = !option_alternative_input;
+                        println!("ReceivedCharacter: TOGGLING TO: {}", option_alternative_input);
+                        window.set_option_alternative_input(option_alternative_input);
+                    },
+                    _ => (),
+                },
+                _ => (),
+            },
             Event::MainEventsCleared => {
                 window.request_redraw();
             }
diff --git a/src/platform/macos.rs b/src/platform/macos.rs
index c96f176a..3f4c8f62 100644
--- a/src/platform/macos.rs
+++ b/src/platform/macos.rs
@@ -63,13 +63,13 @@ pub trait WindowExtMacOS {
     /// Sets whether or not the window has shadow.
     fn set_has_shadow(&self, has_shadow: bool);
 
-    /// Toggles the `Option` key being interpretted as an `Alt` modifier.
+    /// Sets whether the Option key can be used for alternative keyboard input.
     ///
-    /// This will ignore diacritical marks and accent characters from
-    /// being processed as received characters. Instead, the input
-    /// device's raw character will be placed in event queues with the
-    /// Alt modifier set.
-    fn set_option_as_alt(&self, option_as_alt: bool);
+    /// When set to `true`, the option key can be used like AltGr to input
+    /// additional characters and dead keys.
+    ///
+    /// By default, alternative keyboard input is enabled.
+    fn set_option_alternative_input(&self, option_alternative_input: bool);
 }
 
 impl WindowExtMacOS for Window {
@@ -109,8 +109,8 @@ impl WindowExtMacOS for Window {
     }
 
     #[inline]
-    fn set_option_as_alt(&self, option_as_alt: bool) {
-        self.window.set_option_as_alt(option_as_alt)
+    fn set_option_alternative_input(&self, option_alternative_input: bool) {
+        self.window.set_option_alternative_input(option_alternative_input)
     }
 }
 
@@ -161,7 +161,11 @@ pub trait WindowBuilderExtMacOS {
     fn with_resize_increments(self, increments: LogicalSize<f64>) -> WindowBuilder;
     fn with_disallow_hidpi(self, disallow_hidpi: bool) -> WindowBuilder;
     fn with_has_shadow(self, has_shadow: bool) -> WindowBuilder;
-    fn with_ignore_alt_modifier(self, ignore_alt_modifier: bool) -> WindowBuilder;
+
+    /// Sets whether the Option key can be used for alternative keyboard input.
+    ///
+    /// See [`WindowExtMacOs::set_option_alternative_input`] for details.
+    fn with_option_alternative_input(self, option_alternative_input: bool) -> WindowBuilder;
 }
 
 impl WindowBuilderExtMacOS for WindowBuilder {
@@ -229,8 +233,8 @@ impl WindowBuilderExtMacOS for WindowBuilder {
     }
 
     #[inline]
-    fn with_ignore_alt_modifier(mut self, ignore_alt_modifier: bool) -> WindowBuilder {
-        self.platform_specific.ignore_alt_modifier = ignore_alt_modifier;
+    fn with_option_alternative_input(mut self, option_alternative_input: bool) -> WindowBuilder {
+        self.platform_specific.option_alternative_input = option_alternative_input;
         self
     }
 }
diff --git a/src/platform_impl/macos/view.rs b/src/platform_impl/macos/view.rs
index ad83db91..d61b28d1 100644
--- a/src/platform_impl/macos/view.rs
+++ b/src/platform_impl/macos/view.rs
@@ -54,9 +54,11 @@ pub(super) struct ViewState {
     ns_window: id,
     pub cursor_state: Arc<Mutex<CursorState>>,
     ime_spot: Option<(f64, f64)>,
+    raw_characters: Option<String>,
+    is_key_down: bool,
     pub(super) modifiers: ModifiersState,
     tracking_rect: Option<NSInteger>,
-    ignore_alt_modifier: bool,
+    option_alternative_input: bool,
 }
 
 impl ViewState {
@@ -75,9 +77,11 @@ pub fn new_view(
         ns_window,
         cursor_state,
         ime_spot: None,
+        raw_characters: None,
+        is_key_down: false,
         modifiers: Default::default(),
         tracking_rect: None,
-        ignore_alt_modifier: pl_attribs.ignore_alt_modifier,
+        option_alternative_input: pl_attribs.option_alternative_input,
     };
     unsafe {
         // This is free'd in `dealloc`
@@ -101,10 +105,10 @@ pub unsafe fn set_ime_position(ns_view: id, input_context: id, x: f64, y: f64) {
     let _: () = msg_send![input_context, invalidateCharacterCoordinates];
 }
 
-pub unsafe fn set_option_as_alt(ns_view: id, option_as_alt: bool) {
+pub unsafe fn set_option_alternative_input(ns_view: id, option_alternative_input: bool) {
     let state_ptr: *mut c_void = *(*ns_view).get_mut_ivar("winitState");
     let state = &mut *(state_ptr as *mut ViewState);
-    state.ignore_alt_modifier = option_as_alt;
+    state.option_alternative_input = option_alternative_input;
 }
 
 struct ViewClass(*const Class);
@@ -166,6 +170,10 @@ lazy_static! {
             attributed_substring_for_proposed_range
                 as extern "C" fn(&Object, Sel, NSRange, *mut c_void) -> id,
         );
+        decl.add_method(
+            sel!(insertText:replacementRange:),
+            insert_text as extern "C" fn(&Object, Sel, id, NSRange),
+        );
         decl.add_method(
             sel!(characterIndexForPoint:),
             character_index_for_point as extern "C" fn(&Object, Sel, NSPoint) -> NSUInteger,
@@ -175,6 +183,10 @@ lazy_static! {
             first_rect_for_character_range
                 as extern "C" fn(&Object, Sel, NSRange, *mut c_void) -> NSRect,
         );
+        decl.add_method(
+            sel!(doCommandBySelector:),
+            do_command_by_selector as extern "C" fn(&Object, Sel, Sel),
+        );
         decl.add_method(sel!(keyDown:), key_down as extern "C" fn(&Object, Sel, id));
         decl.add_method(sel!(keyUp:), key_up as extern "C" fn(&Object, Sel, id));
         decl.add_method(
@@ -493,6 +505,81 @@ extern "C" fn first_rect_for_character_range(
     }
 }
 
+extern "C" fn insert_text(this: &Object, _sel: Sel, string: id, _replacement_range: NSRange) {
+    println!("ReceivedCharacter: Insert Text");
+    trace!("Triggered `insertText`");
+    unsafe {
+        let state_ptr: *mut c_void = *this.get_ivar("winitState");
+        let state = &mut *(state_ptr as *mut ViewState);
+
+        let has_attr = msg_send![string, isKindOfClass: class!(NSAttributedString)];
+        let characters = if has_attr {
+            // This is a *mut NSAttributedString
+            msg_send![string, string]
+        } else {
+            // This is already a *mut NSString
+            string
+        };
+
+        let slice =
+            slice::from_raw_parts(characters.UTF8String() as *const c_uchar, characters.len());
+        let string = str::from_utf8_unchecked(slice);
+        state.is_key_down = true;
+
+        // We don't need this now, but it's here if that changes.
+        //let event: id = msg_send![NSApp(), currentEvent];
+
+        let mut events = VecDeque::with_capacity(characters.len());
+        for character in string.chars().filter(|c| !is_corporate_character(*c)) {
+            events.push_back(EventWrapper::StaticEvent(Event::WindowEvent {
+                window_id: WindowId(get_window_id(state.ns_window)),
+                event: WindowEvent::ReceivedCharacter(character),
+            }));
+        }
+
+        AppState::queue_events(events);
+    }
+    trace!("Completed `insertText`");
+}
+
+extern "C" fn do_command_by_selector(this: &Object, _sel: Sel, command: Sel) {
+    println!("ReceivedCharacter: Do Command");
+    trace!("Triggered `doCommandBySelector`");
+    // Basically, we're sent this message whenever a keyboard event that doesn't generate a "human readable" character
+    // happens, i.e. newlines, tabs, and Ctrl+C.
+    unsafe {
+        let state_ptr: *mut c_void = *this.get_ivar("winitState");
+        let state = &mut *(state_ptr as *mut ViewState);
+
+        let mut events = VecDeque::with_capacity(1);
+        if command == sel!(insertNewline:) {
+            // The `else` condition would emit the same character, but I'm keeping this here both...
+            // 1) as a reminder for how `doCommandBySelector` works
+            // 2) to make our use of carriage return explicit
+            events.push_back(EventWrapper::StaticEvent(Event::WindowEvent {
+                window_id: WindowId(get_window_id(state.ns_window)),
+                event: WindowEvent::ReceivedCharacter('\r'),
+            }));
+        } else {
+            let raw_characters = state.raw_characters.take();
+            if let Some(raw_characters) = raw_characters {
+                for character in raw_characters
+                    .chars()
+                    .filter(|c| !is_corporate_character(*c))
+                {
+                    events.push_back(EventWrapper::StaticEvent(Event::WindowEvent {
+                        window_id: WindowId(get_window_id(state.ns_window)),
+                        event: WindowEvent::ReceivedCharacter(character),
+                    }));
+                }
+            }
+        };
+
+        AppState::queue_events(events);
+    }
+    trace!("Completed `doCommandBySelector`");
+}
+
 fn get_characters(event: id, ignore_modifiers: bool) -> String {
     unsafe {
         let characters: id = if ignore_modifiers {
@@ -563,20 +650,26 @@ fn update_potentially_stale_modifiers(state: &mut ViewState, event: id) {
 }
 
 extern "C" fn key_down(this: &Object, _sel: Sel, event: id) {
+    println!("ReceivedCharacter: DOWN DOWN DOWN");
     trace!("Triggered `keyDown`");
     unsafe {
         let state_ptr: *mut c_void = *this.get_ivar("winitState");
         let state = &mut *(state_ptr as *mut ViewState);
         let window_id = WindowId(get_window_id(state.ns_window));
-        let ev_mods = event_mods(event);
-        let characters = get_characters(
-            event,
-            state.ignore_alt_modifier && ev_mods.alt() && !ev_mods.ctrl(),
-        );
+        let mods = event_mods(event);
+        let ignore_modifiers = !state.option_alternative_input && mods.alt() && !mods.ctrl();
+        println!("ReceivedCharacter: Ignoring: {}", ignore_modifiers);
+        let characters = get_characters(event, ignore_modifiers);
+        println!("ReceivedCharacter: String: {}", characters);
+
+        // TODO: Do we always want with mods here?
+        state.raw_characters = Some(characters.clone());
 
         let scancode = get_scancode(event) as u32;
         let virtual_keycode = retrieve_keycode(event);
 
+        let is_repeat = msg_send![event, isARepeat];
+
         update_potentially_stale_modifiers(state, event);
 
         #[allow(deprecated)]
@@ -588,18 +681,35 @@ extern "C" fn key_down(this: &Object, _sel: Sel, event: id) {
                     state: ElementState::Pressed,
                     scancode,
                     virtual_keycode,
-                    modifiers: ev_mods,
+                    modifiers: event_mods(event),
                 },
                 is_synthetic: false,
             },
         };
 
-        AppState::queue_event(EventWrapper::StaticEvent(window_event));
-        for character in characters.chars().filter(|c| !is_corporate_character(*c)) {
-            AppState::queue_event(EventWrapper::StaticEvent(Event::WindowEvent {
-                window_id,
-                event: WindowEvent::ReceivedCharacter(character),
-            }));
+        let pass_along = {
+            AppState::queue_event(EventWrapper::StaticEvent(window_event));
+            // Emit `ReceivedCharacter` directly for repeated keys and keys ignoring the modifier
+            // to skip the `input_text` processing.
+            if (is_repeat && state.is_key_down) || ignore_modifiers {
+                for character in characters.chars().filter(|c| !is_corporate_character(*c)) {
+                    AppState::queue_event(EventWrapper::StaticEvent(Event::WindowEvent {
+                        window_id,
+                        event: WindowEvent::ReceivedCharacter(character),
+                    }));
+                }
+                false
+            } else {
+                true
+            }
+        };
+
+        if pass_along {
+            // Some keys (and only *some*, with no known reason) don't trigger `insertText`, while others do...
+            // So, we don't give repeats the opportunity to trigger that, since otherwise our hack will cause some
+            // keys to generate twice as many characters.
+            let array: id = msg_send![class!(NSArray), arrayWithObject: event];
+            let _: () = msg_send![this, interpretKeyEvents: array];
         }
     }
     trace!("Completed `keyDown`");
@@ -611,6 +721,8 @@ extern "C" fn key_up(this: &Object, _sel: Sel, event: id) {
         let state_ptr: *mut c_void = *this.get_ivar("winitState");
         let state = &mut *(state_ptr as *mut ViewState);
 
+        state.is_key_down = false;
+
         let scancode = get_scancode(event) as u32;
         let virtual_keycode = retrieve_keycode(event);
 
diff --git a/src/platform_impl/macos/window.rs b/src/platform_impl/macos/window.rs
index 598980da..5961801b 100644
--- a/src/platform_impl/macos/window.rs
+++ b/src/platform_impl/macos/window.rs
@@ -24,7 +24,7 @@ use crate::{
         monitor::{self, MonitorHandle, VideoMode},
         util::{self, IdRef},
         view::CursorState,
-        view::{self, new_view, set_option_as_alt},
+        view::{self, new_view, set_option_alternative_input},
         window_delegate::new_delegate,
         OsError,
     },
@@ -72,7 +72,7 @@ pub struct PlatformSpecificWindowBuilderAttributes {
     pub resize_increments: Option<LogicalSize<f64>>,
     pub disallow_hidpi: bool,
     pub has_shadow: bool,
-    pub ignore_alt_modifier: bool,
+    pub option_alternative_input: bool,
 }
 
 impl Default for PlatformSpecificWindowBuilderAttributes {
@@ -89,7 +89,7 @@ impl Default for PlatformSpecificWindowBuilderAttributes {
             resize_increments: None,
             disallow_hidpi: false,
             has_shadow: true,
-            ignore_alt_modifier: false,
+            option_alternative_input: true,
         }
     }
 }
@@ -1147,9 +1147,9 @@ impl WindowExtMacOS for UnownedWindow {
     }
 
     #[inline]
-    fn set_option_as_alt(&self, option_as_alt: bool) {
+    fn set_option_alternative_input(&self, option_alternative_input: bool) {
         unsafe {
-            set_option_as_alt(*self.ns_view, option_as_alt);
+            set_option_alternative_input(*self.ns_view, option_alternative_input);
         }
     }
 }
```

</details>

saysjonathan:
@chrisduerr I've applied this patch but it looks like it reverts the fix for ignoring modifiers. In my use case, `alt-c` sends `ç` instead of `^[c`.

My test environment is this patch and `alacritty` using a patched version of `winit`.

chrisduerr:
@saysjonathan Did you fix all the compilation issues? I've changed a bunch of stuff in winit so Alacritty will not just compile with my patch.

I've tested it with winit's `window` application and it seemed to work for me.

saysjonathan:
Intersting, I did not have any compilation issues with Alacritty.

I have @kjmph's change merged against [38fcceb](https://github.com/rust-windowing/winit/tree/38fccebe1fbc4226c75d6180e5317bd93c024951), with your patch applied on top. You can see it in this [branch](https://github.com/saysjonathan/winit/tree/alt-fix-new).

Here's a patch of my Alacritty changes:
```patch
diff --git a/Cargo.toml b/Cargo.toml
index 5bd8497..e634709 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,4 +11,4 @@ debug = 1
 incremental = false
 
 [patch.crates-io]
-winit = { git = "https://github.com/saysjonathan/winit", branch = "alt-fix" }
+winit = { git = "https://github.com/saysjonathan/winit", branch = "alt-fix-new" }
diff --git a/alacritty/src/window.rs b/alacritty/src/window.rs
index e43d527..0326234 100644
--- a/alacritty/src/window.rs
+++ b/alacritty/src/window.rs
@@ -331,7 +331,7 @@ impl Window {
             .with_transparent(true)
             .with_maximized(window_config.maximized())
             .with_fullscreen(window_config.fullscreen())
-            .with_ignore_alt_modifier(true);
+            .with_option_alternative_input(true);
 
         match window_config.decorations {a
             Decorations::Full => window,
```

With those changes I did not experience any issues with compilation. I'll try the `winit` example `window` app as well.

saysjonathan:
Doesn't work for me with the `window` example:

```
2021-01-21 14:13:29,669 TRACE [winit::platform_impl::platform::view] Triggered `insertText`
2021-01-21 14:13:29,669 TRACE [winit::platform_impl::platform::view] Completed `insertText`
2021-01-21 14:13:29,669 TRACE [winit::platform_impl::platform::view] Completed `keyDown`
WindowEvent { window_id: WindowId(Id(140548673512288)), event: KeyboardInput { device_id: DeviceId(DeviceId), input: KeyboardInput { scancode: 8, state: Pressed, virtual_keycode: Some(C), modifiers: ALT }, is_synthetic: false } }
WindowEvent { window_id: WindowId(Id(140548673512288)), event: ReceivedCharacter('ç') }
```

Shouldn't the `RecievedCharacter` here be `^[c` and not `ç`?

chrisduerr:
I mean if you want alternative input disabled for your option, you'll have to pass `false` not `true`.

saysjonathan:
Ah! Helps to fully read the code. Trying now with `window` and `alacritty`.

saysjonathan:
I get the same output in from the `window` example but it is in fact working for me with Alacritty after passing `false` to `with_option_alternative_input`.

```
2021-01-21 14:21:09,968 TRACE [winit::platform_impl::platform::view] Completed `keyDown`
WindowEvent { window_id: WindowId(Id(140204247800544)), event: KeyboardInput { device_id: DeviceId(DeviceId), input: KeyboardInput { scancode: 8, state: Pressed, virtual_keycode: Some(C), modifiers: ALT }, is_synthetic: false } }
WindowEvent { window_id: WindowId(Id(140204247800544)), event: ReceivedCharacter('ç') }
```

chrisduerr:
> I get the same output in from the window example but it is in fact working for me with Alacritty after passing false to with_option_alternative_input.

With my patched window example you can press `x` to toggle the behavior of the option key.

saysjonathan:
Verified `window` is also working. Looks good and sorry for the false failure cases!

```
2021-01-21 14:58:04,392 TRACE [winit::platform_impl::platform::view] Completed `keyDown`
WindowEvent { window_id: WindowId(Id(140308047369808)), event: KeyboardInput { device_id: DeviceId(DeviceId), input: KeyboardInput { scancode: 8, state: Pressed, virtual_keycode: Some(C), modifiers: ALT }, is_synthetic: false } }
WindowEvent { window_id: WindowId(Id(140308047369808)), event: ReceivedCharacter('c') }
```

ArturKovacs:
@kjmph considering that you haven't shown any activity in more than a month and you haven't responded to my previous question, I'm assuing you don't have the time to continue working on this, so @chrisduerr I'd like to ask you to take over this PR. If you can't push to this branch I'm not sure what's the best way to do it, but I'm open to suggestions.

kjmph:
My apologies, to be more clear, I have limited time to take on this work. Please feel free to take over this PR @ArturKovacs. Let me know how I may be of service in transitioning the work to you. I think I succeeded in showing that this can be fixed on MacOS, yet as I’m not as involved in the community, I could use the help getting it available to all other users.

mspielberg:
@chrisduerr It's been quite a while, but are you still interested in taking this over the finish line based on your approach in https://github.com/rust-windowing/winit/pull/1449#issuecomment-764644210?

chrisduerr:
Considering that I don't even have access to a macOS machine, I couldn't finish it up even if I was interested.

ayax79:
Hi, 

I would love to help get this over the line. I'm new to winnit, but I have a mac and I am proficient in rust. 

What is left besides fixing the compilation errors, merging in master changes, and testing?

chrisduerr:
@ayax79 See the comment history. IIRC it broke some IME stuff like emoji input, I'm not sure if my suggested patch is a proper solution but it might be. So you'll probably have to do a bunch of testing and validate that it works properly and then clean it all up.

Considering this patch is ancient, it might be easier to start from scratch and just yoink the new code you need from the patches.

kjmph:
I agree with @chrisduerr ; this patch is out of date as the underlying code was rewritten, best to start a new change lifting the pieces as needed. Let me know what I can do that's best for the project. Note, the patch did improve input handling with the Alt modifier. Releases already have IME issues and this patch doesn't degrade any capabilities. It may be easier to breakout a second issue to address IME handling, and let a fix through for the Alt handling. Cheers.

ayax79:
I compared the patch to master. It won't take me long to reimplement.

ayax79:
I implemented the old fix here: https://github.com/ayax79/winit/tree/fix-alt-modifier-new

I verified with a debugger that everything is working the same as the original fix. There still seems to be some oddness. For instance, once I load up zellij, I still see characters. 

When I have a few minutes, I am going to continue debugging with showkey.

madsmtm:
> I reimplemented the fix here: https://github.com/ayax79/winit/tree/fix-alt-modifier-new

Feel free to open a draft PR, makes it easy for others to follow your progress

ayax79:
I'll try to by end of the day. After rereading this thread, I realized that kjmph's repo didn't have the patch from @chrisduerr. I will be implementing that patch. I'll pull request once it is in.

ayax79:
WIP pull request. This is should have all the patches applied, isn't working yet. Still seeing characters with modifiers: https://github.com/rust-windowing/winit/pull/2576

madsmtm:
Closing in favour of https://github.com/rust-windowing/winit/pull/2576 then.

Issue #1055, opened 2019-07-21
winit::event::Event extensions and time enhancements.


cheako:
https://github.com/rust-windowing/winit/issues/459#issuecomment-513512731

winit::event::Event would logically have:
1. Something to emit when polling: Update
1. Something to emit when a timer expires and obviously this needs to contain a handle to the timer that's expiring: Timer(T)

Consider redoing winit::event_loop::ControlFlow as a trait instance that has callbacks for ```fn set_timer(T, Instant) -> RunningTimerAPI;```, ```fn set_sleeping(&mut self);```, ```fn set_polling(&mut self);```, and ```fn set_exiting(self);```.
#1018

cheako:
...Also a ```fn create_tick(&mut self, T, Duration) -> Rc<TickAPI>;```, you know your average JavaScript timer.  Emits an event every Duration until told to stop....  also it would be nice to provide the TickAPI when emitting ```winit::event::Event::Tick(Rc<TickAPI>, T)```.

cheako:
Ohh, I forgot.  ```winit::event::Event::Update(Duration)```  Since the last update, an inverse Tick.  It should be simple to measure this oneself, but providing it would be nice.  

Sarcasticly, of course Real Time applications are going to need an estimate as to how long until the next Update, so if you could provide that as well, that would be great.  The app will provide how long you can take in calculating the estimated next Update time, obviously.

ZeGentzy:
> [#459 (comment)](https://github.com/rust-windowing/winit/issues/459#issuecomment-513512731)
> 
> winit::event::Event would logically have:
> 
>     1. Something to emit when polling: Update
> 
>     2. Something to emit when a timer expires and obviously this needs to contain a handle to the timer that's expiring: Timer(T)
> 

Are these two cases not covered by https://docs.rs/winit/0.20.0-alpha2/winit/event/enum.Event.html#variant.NewEvents ?

I'm basically at loss to what you want.

cheako:
I was asked to open a new issue for this so you can ignore the implied want for anything.

Ohhhhh  "NewEvents"... but um don't call them that, what would you do the next time you go to add events.  I mean sure there's still "NewerEvents" but you'll just have to trust me you'll soon run out of names with that scheme.

cheako:
[...]

Issue #1018, opened 2019-07-06
`control_flow` by `&mut` rationale


mickvangelderen:
I would like to understand the rationale for having `control_flow` as a mutable parameter rather than the return value in the `EventLoop::run` and similar methods. I am confused because when returned by value it is harder to forget to set it. What are the downsides?

Osspial:
Thanks for the question!

I've actually answered this before, but the answer is [buried pretty deeply](https://github.com/rust-windowing/winit/issues/459#issuecomment-404663870) in the original API discussion so it's pretty hard to find. I'll add it to the FAQ in the announcement post, but I'll also copy it here for posterity:

> Instead of having the user return ` ControlFlow`, make the `EventHandler` function take an `&mut ControlFlow` parameter, which defaults to `ControlFlow::Poll`. The values stored here is persistent between all calls to the function.
> 
> Why do this? With the return-based solution, there are cases where there are several "correct" ways to handle control flow. Consider the following event loop:
> 
> ```rust
> event_loop.run(|event| {
>     match event {
>         Event::LoopEnd => ControlFlow::Poll,
>         _ => ControlFlow::Wait
>     }
> });
> ```
> 
> That situation leads to return patterns like `Wait, Wait, Wait, {...}, Poll`. What do you do with the `Wait`s that are returned before the `Poll`? You could take that as "halt the event loop between each call to the function", which is pretty clearly nonsense. However, it's the only solution that doesn't involve throwing away values. Other solutions could be to have some sort of "control flow precedence pattern" where returning one type overrides future returns of another type, or to ignore all of the return values except for the one from `LoopEnd` (in which case what's the point of having it return anyway!).
> 
> As a persistent parameter, there are no values we have to ignore and it's clear when the control flow of the loop actually changes. Under this, `Wait`, `WaitTimeout`, and `Poll` would only impact control flow when all of a loop's events have been cleared, and `Break` would essentially kill the loop immediately (barring destruction events).

mickvangelderen:
Thank you for answering. 

Now I'm wondering if I should put my update and render code (which are coupled 1:1 for now) in `NewEvents(Poll)` or in `EventsCleared`. I would rather see run take an iterator of 0..n available events. This will make it easier to accumulate multiple updates into a single value (like mouse dx dy).

```rust
event_loop.run(|events| {
    let mut keep_running = true;
    let mut mouse_dx = 0; // accumulate multiple mouse move events.
    for event in events {
[...]

PR #1450, opened 2020-02-10
Add an example showing how to switch between ControlFlow::Poll and Co…

dabreegster:
…ntrolFlow::Wait

I'm trying to figure out how to write a game loop that updates a simulation based on how much real time has passed, but lets the user pause. From #1055 and #1018, it seems like other people have been confused by this too. I've reduced the use case to a hopefully minimal stopwatch example that the user can pause/resume. If you just let it run, you can see that the total amount of update time measured by the stopwatch slowly diverges from the time since the beginning of the program.

I'm pretty sure I'm not using WaitUntil correctly. Should the "time has passed, go update by some delta-time" instead happen from a NewEvents(StartCause::ResumeTimeReached) case? I tried some variations of that without any luck. (I can update with some other attempts if that'd be helpful.) Any hints? Hopefully an example like this one could help people in the future understand how to use ControlFlow a little better. Thanks!

- [X ] `cargo fmt` has been run on this branch
- [X ] Created or updated an example program if it would help users understand this functionality
- [X ] Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - On android added support for `run_return`.
 - On MacOS, Fixed fullscreen and dialog support for `run_return`.
 - On Windows, fix bug where we'd try to emit `MainEventsCleared` events during nested win32 event loops.
+- On Web, use mouse events if pointer events aren't supported. This affects Safari.
 - On Windows, `set_ime_position` is now a no-op instead of a runtime crash.
 - On Android, `set_fullscreen` is now a no-op instead of a runtime crash.
 - On iOS and Android, `set_inner_size` is now a no-op instead of a runtime crash.
--- a/src/platform_impl/web/web_sys/canvas.rs
+++ b/src/platform_impl/web/web_sys/canvas.rs
@@ -9,8 +9,8 @@ use std::rc::Rc;
 
 use wasm_bindgen::{closure::Closure, JsCast};
 use web_sys::{
-    Event, FocusEvent, HtmlCanvasElement, KeyboardEvent, MediaQueryListEvent, PointerEvent,
-    WheelEvent,
+    Event, FocusEvent, HtmlCanvasElement, KeyboardEvent, MediaQueryListEvent, MouseEvent,
+    PointerEvent, WheelEvent,
 };
 
 pub struct Canvas {
@@ -24,8 +24,14 @@ pub struct Canvas {
     on_cursor_leave: Option<Closure<dyn FnMut(PointerEvent)>>,
     on_cursor_enter: Option<Closure<dyn FnMut(PointerEvent)>>,
     on_cursor_move: Option<Closure<dyn FnMut(PointerEvent)>>,
-    on_mouse_press: Option<Closure<dyn FnMut(PointerEvent)>>,
-    on_mouse_release: Option<Closure<dyn FnMut(PointerEvent)>>,
+    on_pointer_press: Option<Closure<dyn FnMut(PointerEvent)>>,
+    on_pointer_release: Option<Closure<dyn FnMut(PointerEvent)>>,
+    // Fallback events when pointer event support is missing
+    on_mouse_leave: Option<Closure<dyn FnMut(MouseEvent)>>,
+    on_mouse_enter: Option<Closure<dyn FnMut(MouseEvent)>>,
+    on_mouse_move: Option<Closure<dyn FnMut(MouseEvent)>>,
+    on_mouse_press: Option<Closure<dyn FnMut(MouseEvent)>>,
+    on_mouse_release: Option<Closure<dyn FnMut(MouseEvent)>>,
     on_mouse_wheel: Option<Closure<dyn FnMut(WheelEvent)>>,
     on_fullscreen_change: Option<Closure<dyn FnMut(Event)>>,
     wants_fullscreen: Rc<RefCell<bool>>,
@@ -76,8 +82,13 @@ impl Canvas {
             on_cursor_leave: None,
             on_cursor_enter: None,
             on_cursor_move: None,
-            on_mouse_release: None,
+            on_pointer_release: None,
+            on_pointer_press: None,
+            on_mouse_leave: None,
+            on_mouse_enter: None,
+            on_mouse_move: None,
             on_mouse_press: None,
+            on_mouse_release: None,
             on_mouse_wheel: None,
             on_fullscreen_change: None,
             wants_fullscreen: Rc::new(RefCell::new(false)),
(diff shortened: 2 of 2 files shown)


examples/pauseable.rs
```diff
@@ -0,0 +1,98 @@
...
+                    let dt = last_update.elapsed();
+                    total += dt;
+                    println!(
+                        "- Update: +{:?} for a total of {:?}. It's been {:?} since the very start",
+                        dt,
+                        total,
+                        started_at.elapsed()
+                    );
+                    // Pretend that updating the application takes about 20ms
+                    thread::sleep(Duration::from_millis(20));
```
goddessfreya: Won't `last_update` diverge by 20ms because the `let now = ...` is below this line?

dabreegster: Ah, thanks! I'm not sure why it felt more natural to measure the last update after applying it. Switching the order seems to make this example work as expected.

examples/pauseable.rs
```diff
@@ -0,0 +1,101 @@
...
+                    println!("- Pausing");
+                }
+            }
+            Event::WindowEvent { .. } => {
+                println!("- Handling some input");
+                // Pretend that handling the event takes about 1ms
+                thread::sleep(Duration::from_millis(1));
+                window.request_redraw();
+            }
+            Event::RedrawRequested(_) => {
```
filnet: move that block to after MainEventsCleared

examples/pauseable.rs
```diff
@@ -0,0 +1,101 @@
...
+                } else {
+                    *control_flow = ControlFlow::Wait;
+                    println!("- Pausing");
+                }
+            }
+            Event::WindowEvent { .. } => {
+                println!("- Handling some input");
+                // Pretend that handling the event takes about 1ms
+                thread::sleep(Duration::from_millis(1));
+                window.request_redraw();
```
filnet: calling `request_redraw` might result in undefined behavior
see https://docs.rs/winit/0.21.0/winit/window/struct.Window.html#method.request_redraw

best would be to call it in MainEventsCleared

examples/pauseable.rs
```diff
@@ -0,0 +1,101 @@
...
+                    println!("- Resuming!");
+                } else {
+                    *control_flow = ControlFlow::Wait;
+                    println!("- Pausing");
+                }
+            }
+            Event::WindowEvent { .. } => {
+                println!("- Handling some input");
+                // Pretend that handling the event takes about 1ms
+                thread::sleep(Duration::from_millis(1));
```
filnet: would be better to call this in `MainEventsCleared`

examples/pauseable.rs
dabreegster: Done

examples/pauseable.rs
dabreegster: Done. Thanks for pointing out the caveat with request_redraw; it's a little unintuitive. I would expect to call it after changing any internal state that I want to be redrawn, and all of the requests during each WindowEvent to get batched together reasonably.

examples/pauseable.rs
```diff
@@ -0,0 +1,103 @@
...
+                        started_at.elapsed()
+                    );
+                    // Pretend that updating the application takes about 20ms
+                    thread::sleep(Duration::from_millis(20));
+                }
+                // Redraw since the application state was updated either by time passing or some
+                // input being handled previously.
+                window.request_redraw();
+            }
+            Event::WindowEvent { .. } => {
```
filnet: That block should be moved to before `MainEventsCleared` ;)

I found this piece documentation useful : https://docs.rs/winit/0.21.0/winit/event/index.html
You should try to organize your code in the order the events are sent.

examples/pauseable.rs
```diff
@@ -0,0 +1,101 @@
...
+            Event::MainEventsCleared => {
+                if running {
+                    let dt = last_update.elapsed();
+                    let now = Instant::now();
+                    // We want an update every update_frequency, so calculate the next time before
+                    // we process the update, which takes some amount of time. (If processing the
+                    // update takes more than update_frequency, then the next update will occur
+                    // immediately.)
+                    last_update = now;
+                    *control_flow = ControlFlow::WaitUntil(now + update_frequency);
```
Osspial: > Should the "time has passed, go update by some delta-time" instead happen from a `NewEvents(StartCause::ResumeTimeReached)` case?

Yep! I've written a diff that shows how I'd structure the code with the timer updating in `NewEvents`:
```diff
diff --git a/examples/pauseable.rs b/examples/pauseable.rs
index 3892882a..9675d159 100644
--- a/examples/pauseable.rs
+++ b/examples/pauseable.rs
@@ -1,7 +1,7 @@
 use std::thread;
 use std::time::{Duration, Instant};
 use winit::{
-    event::{ElementState, Event, KeyboardInput, VirtualKeyCode, WindowEvent},
+    event::{ElementState, Event, KeyboardInput, VirtualKeyCode, WindowEvent, StartCause},
     event_loop::{ControlFlow, EventLoop},
     window::WindowBuilder,
 };
@@ -27,11 +27,33 @@ fn main() {
     let started_at = Instant::now();
 
     let mut running = true;
-    let mut last_update = Instant::now();
     let mut total = Duration::new(0, 0);
+    let mut update_triggered = false;
 
     event_loop.run(move |event, _, control_flow| {
         match event {
+            Event::NewEvents(StartCause::Init) => {
+                *control_flow = ControlFlow::WaitUntil(Instant::now() + update_frequency);
+            },
+            Event::NewEvents(StartCause::ResumeTimeReached{requested_resume, ..}) => {
+                if running {
+                    // Only do an update when the timer has been reached. We defer running the
+                    // update code until `MainEventsCleared` so that the update can take account
+                    // of user input events.
+                    update_triggered = true;
+
+                    // We schedule the next update relative to the previously scheduled update time,
+                    // rather than Instant::now(). This has a couple benefits:
+                    // - The update timer is immune to drift caused by any delay between the start
+                    //   of the event loop iteration and the time `Instant::now()` is called.
+                    // - If unexpected circumstances cause the update code to get called at a lower
+                    //   frequency than `update_frequency` specifies (say, the OS doesn't schedule
+                    //   the update for 400ms), the scheduler will aggressively schedule updates
+                    //   until we're caught up with real time.
+                    *control_flow = ControlFlow::WaitUntil(requested_resume + update_frequency);
+                    total += update_frequency;
+                }
+            },
             Event::WindowEvent {
                 event: WindowEvent::CloseRequested,
                 ..
@@ -51,9 +73,7 @@ fn main() {
             } => {
                 running = !running;
                 if running {
-                    let now = Instant::now();
-                    last_update = now;
-                    *control_flow = ControlFlow::WaitUntil(now + update_frequency);
+                    *control_flow = ControlFlow::WaitUntil(Instant::now() + update_frequency);
                     println!("- Resuming!");
                 } else {
                     *control_flow = ControlFlow::Wait;
@@ -66,25 +86,18 @@ fn main() {
                 thread::sleep(Duration::from_millis(1));
             }
             Event::MainEventsCleared => {
-                if running {
-                    let dt = last_update.elapsed();
-                    let now = Instant::now();
-                    // We want an update every update_frequency, so calculate the next time before
-                    // we process the update, which takes some amount of time. (If processing the
-                    // update takes more than update_frequency, then the next update will occur
-                    // immediately.)
-                    last_update = now;
-                    *control_flow = ControlFlow::WaitUntil(now + update_frequency);
-
-                    total += dt;
+                if update_triggered {
                     println!(
                         "- Update: +{:?} for a total of {:?}. It's been {:?} since the very start",
-                        dt,
+                        update_frequency,
                         total,
                         started_at.elapsed()
                     );
+
                     // Pretend that updating the application takes about 20ms
                     thread::sleep(Duration::from_millis(20));
+
+                    update_triggered = false;
                 }
                 // Redraw since the application state was updated either by time passing or some
                 // input being handled previously.
```

Let me know if you have any questions about that - it's hard for me to tell what particular things are going to be confusing, since I've spent so much time working with this API!

examples/pauseable.rs
```diff
@@ -0,0 +1,101 @@
...
+fn main() {
+    simple_logger::init().unwrap();
+    let event_loop = EventLoop::new();
+
+    let window = WindowBuilder::new()
+        .with_title("Press space to pause/resume")
+        .build(&event_loop)
+        .unwrap();
+
+    // This is an example of a pauseable game loop. While running == false, the application only
```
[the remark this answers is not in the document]
dabreegster: Done

examples/pauseable.rs
```diff
@@ -0,0 +1,101 @@
...
+                    println!("- Resuming!");
+                } else {
+                    *control_flow = ControlFlow::Wait;
+                    println!("- Pausing");
+                }
+            }
+            Event::WindowEvent { .. } => {
+                println!("- Handling some input");
+                // Pretend that handling the event takes about 1ms
+                thread::sleep(Duration::from_millis(1));
```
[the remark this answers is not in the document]
dabreegster: Removed. I wonder if it'd be useful for winit to optionally warn when this event does take long, to help people who accidentally introduce long blocking steps as a response to some input.

examples/pauseable.rs
dabreegster: Thanks for the detailed fixes! With this patch, the behavior is now a little strange. If you produce input while running, `total` rises too quickly, and after leaving the mouse still again, there's a delay (Seems to be about 100ms to me) before the periodic updates resume. I'll take a closer look tonight. Do we need to handle WaitCancelled, since there's input events interrupting things?

Osspial: What OS are you on? I'm running Windows 10, and I'm not seeing that behavior. It could just be that the additional lines printed to the console on input events are making it seem like `total` is rising too quickly - if you disable all non-update `println`s by commenting out lines `86` and `107`, do you observe the same behavior?

Osspial: Also, we shouldn't have to handle `WaitCancelled`. If the specified timeout has been reached, the event loop should always emit `ResumeTimeReached` even if there are additional events ready to be processed.

dabreegster: (Sorry for the massively delayed response). I'm on Ubuntu 18.04. Commenting the `println`s has no effect. Printing the absolute time since the start is the useful comparison.

Attached a video of the behaviour I'm seeing. As soon as I start wiggling the mouse, the total update time jumps up too quickly. Then after stopping the chain of input events, nothing happens; the periodic updates seem to be gone (or delayed). I press space to pause and resume, then things seem to go back to normal.

winit

Issue #1387, opened 2020-01-11
Event now takes a lifetime


est31:
Since the 0.20.0 release, `Event` as well as `WindowEvent` are now taking lifetimes ([old docs.rs](https://docs.rs/winit/0.20.0-alpha6/winit/event/enum.WindowEvent.html), [new docs.rs](https://docs.rs/winit/0.20.0/winit/event/enum.WindowEvent.html)). The reason for this is the `ScaleFactorChanged` event's `new_inner_size` field. I think this change is bad: First, separation of concerns is violated. Events are supposed to be immutable input to the program, not provide means for the program to react to that input. Second, Event now ceased to be Clone and the lifetime means I have to think about workarounds in glium's example support code: https://github.com/glium/glium/pull/1817.

The PR that has added the lifteime seems to be https://github.com/rust-windowing/winit/pull/895 where it was still called `new_inner_rect_opt`.

Can this feedback maybe be implemented in a different way?

I have great respect for you folks. Thanks for maintaining glutin/winit.

PRs to watch:

* #1456
* #1478
* #2981

goddessfreya:
I think your second point is resolved [here](https://github.com/rust-windowing/winit/pull/1374/files#diff-3d7686001860aa382f8b9c18554cc138R64), right?

kchibisov:
>Events are supposed to be immutable input to the program, not provide means for the program to react to that input

I also agree with you here and share your thoughts. We faced similar issue in `alacritty` during our update to `glutin 0.22.0`, while our "workaround" seems working I don't really like how it looks right now in our code, since we're remapping events with lifetime and transmute everything else. You can take a look on our update PR here https://github.com/alacritty/alacritty/pull/3164. 

>Second, Event now ceased to be Clone and the lifetime means I have to think about workarounds in glium's example support code: glium/glium#1817.

I'm not sure how some parts of `glium` works internally, but maybe you'll get some inspiration from our code for a time being. However if you also thinking about workaround in your downstream library, I find it interesting that the new change accepts as "we need to workaround", rather then a "change makes sense".

Anyway, I'm here to just share some thoughts, since I was fine with the fact that alacritty can't adopt without some unsafe tricks to a new winit, since we're a bit odd in some cases. However if one more downstream client also having troubles then maybe the problem is a bit wider than "alacritty can't adopt without unsafe tricks", as I've initially thought about this lifetime addition.

cc @chrisduerr , since he is aware more than I on our event loop and will likely correct me here.

[...]

PR #1456, opened 2020-02-12
Remove lifetime from `Event` type

tangmi:
- Prototype implementation in response to #1387.
- Only implemented for Windows
- Implemented using `Arc<Mutex<T>>` (I'm unsure of the exact implications here!)
- Tested by running `cargo run --example dpi` and changing the Windows DPI settings while it is running (I'm unsure of the intended behavior, but the state seems to be set correctly!)
- *This is a prototype! There are many questions to answer first!* **Removing the lifetime on `Event` may not even be what we want!**

todo:

- [x] Implement on remaining platforms
  - [x] Windows
  - [ ] iOS
  - [x] macOS
  - [ ] X11
  - [ ] Wayland
  - [ ] Andoid
- [ ] Figure out what to do with `scoped_arc_cell`
  - [ ] (since this is a breaking change): write up an "upgrade" instruction?
  - note: No changes to compatibilities


est31:
Nice PR!

I don't think that performance is *that* important. How often does the DPI change? I'd say it's quite rarely. A `Mutex` should be enough. However, to make changing the implementation easy in the future, you could replace the `Mutex` with an opaque wrapper type that allows reading & writing the `PhysicalSize<u32>` through functions taking `&self`. That wrapper type could then use a `AtomicU64` for storage if desired (`PhysicalSize<u32>` fits into a u64).

Also it would be cool if you could re-add the Clone impls for Event.

tangmi:
I don't thing perf is a big concern at all, DPI can change at most a few times a second (afaik, zooming in/out on the web is the quickest way to change DPI, but that support isn't yet implemented in winit).

I'd like to get @Osspial and @goddessfreya about this change, the key difference between this and the bottom row of https://github.com/rust-windowing/winit/issues/939#issuecomment-505331805 is that this does not expose the `Arc<_>` to the user, disallowing a stray refcount after the event is processed. The performance cost is (hopefully) trivial as I expect `ScaleFactorChanged` events do not happen often and we only incur the extra cost when they do. Please correct me if I'm missing something here!

mitchmindtree:
Thanks for taking the time to address this @tangmi!

I've been updating `conrod` and `nannou` to winit 0.21 and adapting to the new event lifetimes and lack of `Clone` has required some serious refactoring acrobatics and some unfortunate public API changes. I think all of my recent issues are addressed by this PR - especially if we could re-add `Clone` to the `Event` type which these changes seem to allow.

est31:
@mitchmindtree absolutely if this PR doesn't do it, I'll file a PR myself to re-add the Clone impls.

tangmi:
@mitchmindtree My understanding is that `Clone` was removed to disallow users from having a way of modifying the `ScaleFactorChanged` event after the event occurs. If we want to re-introduce it with this PR, we'd need to do something like poison the `ScaleFactorChanged` event's `new_inner_size` so it cannot be modified from a clone.

mitchmindtree:
> My understanding is that Clone was removed to disallow users from having a way of modifying the ScaleFactorChanged event after the event occurs

Users can still do that with this implementation with or without `Clone` by storing the event in a container that lives above the scope of the function submitted to `EventLoop::run`- it was the lifetime that originally prevented this. That said, this safety could still be provided at runtime by something like an `AtomicBool` indicating whether or not the `new_inner_size` handle is still valid. I don't think this should affect an implementation of `Clone`, especially as the contrast in ergonomics is quite significant.

tangmi:
(I wasn’t a part of the original event loops 2.0 discussions so I’m a bit clueless here)

Since EventLoop::run passes a &Event (and Event is not clone), a user would not be able to store it in a larger scope, right? Once it becomes Clone, then they can.

Whether or not Event should be Clone is probably out of the scope of this PR, but it may be worth creating an issue (if there’s not already) discussing the concrete use cases of having it Clone, since there might be a nice solution that even better handles the intended use cases.

tangmi:
Update:

- New design uses a wrapper struct for just the `new_inner_size`
- The wrapper struct is `Clone` (which should allow `Event` to be clone).
- Uses interior mutability as the mechanism to allow a user to modify the value (which could allow `Event` to be passed by `&Event` in `EventLoop::run`).
- Forces each backend to poison the `new_inner_size` wrapper to prevent users from attempting to change the size (if they clone the wrapper)

tangmi:
@goddessfreya: While this PR isn't ready to review, but I would love your thoughts on the concept (replace the lifetime with interior mutability) to see if it's something we'd consider pursuing!

tangmi:
@Osspial no worries at all for the response times! I know we're all busy (and the world is especially weird right now!).

I'll need to take some time myself to re-review this changelist to remember what I was doing and clean it up. If my "todo" list is accurate, I need to finish the remaining platforms, and do some bookkeeping/docs/tests.

tangmi:
@Osspial: turns out most of my changes was just the "interior mut cool thing", which you've greatly improved.

Latest changes:
- rebased onto master
- copy Osspial/scoped_arc_cell into code and modify for winit usage
- remove unused/obsolete code (mostly clone/to_static impls)

I'll try and work my way through the various backends, may need some help with e.g. android (or permission to spam you with github actions emails)

tangmi:
Open questions:
- Should the iOS/macOS event proxies be removed? Do types like this exist in the linux/android backends?
- Untested on the linux/ios/android backends

tangmi:
@Osspial On macOS, `EventProxy` is unconditionally buffered, so we can either pass through the `ScopedArcCell` owner with the proxy or just create a new `ScopedArcCell`, since we're at the point where the user callback runs synchronously (and can resize the window with the user-specified size immediately following).

On Windows, the `send_event` here:

https://github.com/rust-windowing/winit/blob/03335cef851495e0a2843db7019516988740688b/src/platform_impl/windows/event_loop.rs#L1717

may happen synchronously or asynchronously, if the `event_handler` is currently in use. I think even if we pass the `ScopedArcCell` owner with the buffered event, the code that responds to the user-specified size may have already passed if the event ends up being buffered. Thoughts?

TimonPost:
This PR would be very usefull if it could be merged. This event life time is a pain.

alice-i-cecile:
I'm running into this trying to do input playback for Bevy. How can I help move this forward?

madsmtm:
@alice-i-cecile you may be able to use `Event::to_static`. Note that it is likely that that workaround isn't going to work reliably; many parts of `winit` expect that you respond to events somewhat immediately (e.g. sending the event to another thread for later processing is a bad idea).

alice-i-cecile:
Yep, that's the workaround I'm using for now :( It's not great, but it's better than writing my own enum type that has all the variants except one.

tangmi:
@alice-i-cecile this PR has gotten quite stale and probably needs a rewrite given the ~2.5 years that have elapsed and things have certainly changed. It's probably worth bumping #1387 and working with @madsmtm (a current member of rust-windowing) to come up with a design.

alice-i-cecile:
Great, thanks for the heads up. I ended up using a different approach for now (mocking inputs at a higher level of abstraction), but if I end up running into this again I'll take a look there.

madsmtm:
Replaced by https://github.com/rust-windowing/winit/pull/2981

PR #1468, opened 2020-02-14
Map `UserEvent` properly in `Event::to_static`

hecrj:
The current `Event::to_static` returns `None` for user events. This PR makes it return the proper event instead.

This fixes an issue reported in https://github.com/hecrj/iced/issues/188, caused by a panic here:

https://github.com/rust-windowing/winit/blob/ad7d4939a8be2e0d9436d43d0351e2f7599a4237/src/platform_impl/windows/event_loop/runner.rs#L62


murarth:
This needs an entry in the changelog.

hecrj:
@murarth Indeed, my bad. Fixed it.

Issue #1451, opened 2020-02-10
[Wayland] call to inner_size on the same event loop tick when getting Resize event is using old size


kchibisov:
If you call `window.inner_size()` right when you get Resize event, it'll be still using the old size on Wayland. The problem is that inner_size is getting updated right after resize event and executing a bunch user callback.

Matthias-Fauconneau:
my PR also fixes this now.

Issue #1446, opened 2020-02-09
Wayland backend reporting incorrect window size in WindowEvent::Resized event


digitalstreamio:
Tested with version: winit 0.21 

I have an OpenGL app that opens a windowed context with inner size specified to some initial value (800x600).

While running on X11 backend, the size of the window reported by WindowEvent::Resized and window.inner_size() is consistent (800x600).

On Wayland backend, window height reported by Resized event is 30px less than value reported by window.inner_size(). (difference seems to be equal to window decoration height). If I use the values from the Resized event and set OpenGL's viewport, it results with a 30px black bar on top of the window, indicating that value reported by the event is incorrect and does not cover the whole inner part of the window . 

```
                WindowEvent::Resized(size) => {
                    info!("window resized {}x{}", size.width, size.height);
                    let size2 = ctx.platform.windowed_context.window().inner_size();
                    info!("window inner size {}x{}", size2.width, size2.height);
                }
```

Actual:

 Here is reported output.

> INFO [zinc64::ui::main] - window resized 800x570
> INFO [zinc64::ui::main] - window inner size 800x600

Exptected:

> INFO [zinc64::ui::main] - window resized 800x600
> INFO [zinc64::ui::main] - window inner size 800x600

kchibisov:
Could you provide me log by applying the following diff?

```diff
diff --git a/src/platform_impl/linux/wayland/event_loop.rs b/src/platform_impl/linux/wayland/event_loop.rs
index 1ceedf19..eb22c195 100644
--- a/src/platform_impl/linux/wayland/event_loop.rs
+++ b/src/platform_impl/linux/wayland/event_loop.rs
@@ -719,10 +719,12 @@ impl<T> EventLoop<T> {
                         let physical_size = logical_size
                             .to_physical(window.new_dpi.unwrap_or(window.prev_dpi) as f64);

+                        println!("Setting inner size to a user ({},{})", w,h);
                         callback(Event::WindowEvent {
                             window_id,
                             event: WindowEvent::Resized(physical_size),
                         });
+                        println!("Applying inner size initial change ({},{})", w, h);
                         *window.size = (w, h);
                     }
                 }
diff --git a/src/platform_impl/linux/wayland/window.rs b/src/platform_impl/linux/wayland/window.rs
index 03b58b38..65ebd8f2 100644
--- a/src/platform_impl/linux/wayland/window.rs
+++ b/src/platform_impl/linux/wayland/window.rs
[...]

PR #1473, opened 2020-02-17
On Wayland, Hide CSD for fullscreen windows

kchibisov:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - On Wayland, Add HiDPI cursor support
 - On Web, add the ability to query "Light" or "Dark" system theme send `ThemeChanged` on change.
 - Fix `Event::to_static` returning `None` for user events.
+- On Wayland, Hide CSD for fullscreen windows.
 
 # 0.21.0 (2020-02-04)
 
--- a/src/platform_impl/linux/wayland/event_loop.rs
+++ b/src/platform_impl/linux/wayland/event_loop.rs
@@ -39,7 +39,10 @@ use crate::{
     window::{CursorIcon, WindowId as RootWindowId},
 };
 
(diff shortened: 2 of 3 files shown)


kaimast:
Thanks for fixing this! Has this been released yet?

kchibisov:
>Thanks for fixing this! Has this been released yet?

Yes it was in 0.22.0.

kaimast:
Just tested this with `glutin 0.24`/`winit 0.22`.

The CSD is indeed gone but the window context is still smaller than the screen. Essentially I now have a gap at the bottom of the window that shows whatever is behind it.

Not sure if this is a glutin or winit error, but the `inner_size` of the window does not seem to get updated correctly.

kchibisov:
@kaimast `inner_size` is delayed on update on Wayland, that's true. But if you're requesting inner_size of `X` there's no guarantee that you'll actually use it, compositor can give the size of `Y` instead. + make sure that you're properly resizing things and calling `resize` on window context.

kchibisov:
For inner size issue look at https://github.com/rust-windowing/winit/issues/1451 , but for `resize` read through https://github.com/rust-windowing/winit/issues/1446

kaimast:
> @kaimast `inner_size` is delayed on update on Wayland, that's true. But if you're requesting inner_size of `X` there's no guarantee that you'll actually use it, compositor can give the size of `Y` instead. + make sure that you're properly resizing things and calling `resize` on window context.

Sorry, I am actually not using inner_size but the Resize event. I am also calling window_context.resize(). 
I noticed I have the same issue with the fullscreen example in glutin. So I am pretty sure its a bug :(

kchibisov:
please report this issue separately instead of commenting on PR. Also include some example for me to test, since I can't repro it?

kaimast:
I did at #1509

PR #1478, opened 2020-02-21
Implement Clone for 'static events

semtexzv:
Not applicable:

There is no reason why 'static events can't be Clone. This unblocks most use cases for cloning safe events while not changing current design significantly.

Blocking WASM + OpenGL implementation for Amethyst ( See https://github.com/amethyst/amethyst/pull/2040)


est31:
Ideally, we should wait for #1456 and then just add a derive impl.

semtexzv:
I agree, current design is not ideal, but if no other approach is found, this is a way to get Clone safely for events.

src/event.rs
```diff
@@ -807,8 +921,10 @@ pub enum VirtualKeyCode {
     Multiply,
     Mute,
     MyComputer,
-    NavigateForward,  // also called "Prior"
-    NavigateBackward, // also called "Next"
+    NavigateForward,
+    // also called "Prior"
```
filnet: comment looks misplaced now

src/event.rs
```diff
@@ -807,8 +921,10 @@ pub enum VirtualKeyCode {
...
+    NavigateBackward,
+    // also called "Next"
```
filnet: same

azriel91:
Heya, is it okay to get this merged (and perhaps published) before a big `Event` refactor?

Since a refactor would be a breaking change anyway, I don't see any harm in allowing downstream projects to depend on a published version with this change, with the prospect that they have to migrate to the new `Event` structure (which they have to do anyway).

PR #1483, opened 2020-02-25
Remove logical pixel handling from scroll events

### LINKED ISSUE #1406: Inconsistent usage of LogicalPosition vs PhysicalPosition
I notice that sometime between version 0.20.0-alpha5 and 0.20.0, some parts of the event loop started reporting physical positions instead of logical ones: `CursorMoved::position` and `Touch::location` (with differing types; this seems okay). Conversely, `MouseScrollDelta::PixelDelta` still uses a `LogicalPostiion` (didn't check whether there are others).

I'm actually fine with this either way; converting is easy and the types catch any errors; it just seems like an oversight being inconsistent.

Additionally, my lib mostly uses physical positions internally (to avoid fractional positions / or limited GUI scaling), and using a physical position here would reduce the number of places the lib needs to track the DPI factor.

chrisduerr:
This removes the `LogicalPosition` from the `PixelDelta`, since all
other APIs have been switched to use `PhysicalPosition` instead too.

It also seems like `LineDelta` was using both logical and physical
sizes, depending on platform. I've changed this to explicitly take a
`PhysicalPosition` parameter to force a uniform return value.

I'm not entirely sure which platform uses which kind of pixel sizes by
default, so I've made the assumption that both the web and macOS
platforms deliver their events in LogicalPositions.

Fixes #1406.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -37,7 +37,7 @@ use instant::Instant;
 use std::path::PathBuf;
 
 use crate::{
-    dpi::{LogicalPosition, PhysicalPosition, PhysicalSize},
+    dpi::{PhysicalPosition, PhysicalSize},
     platform_impl,
     window::{Theme, WindowId},
 };
@@ -764,7 +764,7 @@ pub enum MouseScrollDelta {
     /// Scroll events are expressed as a PixelDelta if
     /// supported by the device (eg. a touchpad) and
     /// platform.
-    PixelDelta(LogicalPosition<f64>),
+    PixelDelta(PhysicalPosition<f64>),
 }
 
 /// Symbolic name for a keyboard key.
--- a/src/platform_impl/linux/wayland/pointer.rs
+++ b/src/platform_impl/linux/wayland/pointer.rs
@@ -170,14 +170,15 @@ pub fn implement_pointer(
                                     wl_pointer::Axis::HorizontalScroll => x += value as f32,
                                     _ => unreachable!(),
                                 }
+                                let scale_factor = surface::get_dpi_factor(&surface) as f64;
+                                let delta = LogicalPosition::new(x as f64, y as f64)
+                                    .to_physical(scale_factor);
                                 sink.send_window_event(
                                     WindowEvent::MouseWheel {
                                         device_id: crate::event::DeviceId(
                                             crate::platform_impl::DeviceId::Wayland(DeviceId),
                                         ),
-                                        delta: MouseScrollDelta::PixelDelta(
-                                            (x as f64, y as f64).into(),
-                                        ),
+                                        delta: MouseScrollDelta::PixelDelta(delta),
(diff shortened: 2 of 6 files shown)


src/event.rs
```diff
@@ -640,14 +640,14 @@ pub enum MouseScrollDelta {
     ///
     /// Positive values indicate movement forward
     /// (away from the user) or rightwards.
```
nixpulvis:
```suggestion
    /// Positive values indicate upward and rightward movement.
```

chrisduerr: This has nothing to do with this PR, so I don't see why I would touch any of that?

I can't say if the original statement is correct, since it does not really indicate any concrete direction ("the user" is in front of the screen, so I doubt it moves away from him).

nixpulvis: Why not fix the comment while we're here? It'll never get fixed at this rate.

chrisduerr: Because I do not think this comment makes much sense anyways and I do not plan to argue about it in this PR. If you are more invested in bikeshedding about it, there's nothing wrong with opening a separate PR?

I'm fine with changing it in this PR if there's a direct suggestion from winit's maintainers, but I do not wish to invest my time to do anything about this, or have this PR unnecessarily held back because of it.

nixpulvis: @chrisduerr that's fair. I should mark my comments with (no-block).

src/platform_impl/linux/wayland/pointer.rs
```diff
@@ -204,27 +205,30 @@ pub fn implement_pointer(
...
                             if let Some((x, y)) = axis_discrete_buffer {
+                                let delta = LogicalPosition::new(x as f64, y as f64)
+                                    .to_physical(scale_factor);
                                 sink.send_window_event(
                                     WindowEvent::MouseWheel {
                                         device_id: crate::event::DeviceId(
                                             crate::platform_impl::DeviceId::Wayland(DeviceId),
                                         ),
-                                        delta: MouseScrollDelta::LineDelta(x as f32, y as f32),
+                                        delta: MouseScrollDelta::LineDelta(delta),
```
kchibisov: We shouldn't scale `LineDelta` since it's a discrete value under the hood, which does require it. So add a comment about it and pass it right away.

src/event.rs
```diff
@@ -640,14 +640,14 @@ pub enum MouseScrollDelta {
...
-    LineDelta(f32, f32),
+    LineDelta(PhysicalPosition<f64>),
```
Osspial: I don't think `LineDelta` should use the position type. `PhysicalPosition` refers to pixels, but lines aren't pixels.

chrisduerr: I think that makes a lot of sense, yeah. A simple f32 tuple the way it was before should be good.

PR #1489, opened 2020-03-01
macOS: Only check is_repeat to send ReceivedCharacter. Fixes #1488

### LINKED ISSUE #1488: macOS: Repeated key's ReceivedCharacter cancelled by previous key's keyUp event
On macOS only, with some combinations of repeated keys, the action of releasing the first repeated key will cause the remaining repeated key to no longer call `ReceivedCharacter`.

Steps to reproduce:
1. Hold down <kbd>j</kbd> until it repeats
1. While still holding down <kbd>j</kbd>, hold down <kbd>k</kbd> until that repeats. (<kbd>j</kbd>'s `ReceivedCharacter`s will stop, superseded by <kbd>k</kbd>'s `ReceivedCharacter`s, as expected.)
1. Release <kbd>j</kbd> while still holding <kbd>k</kbd>. This causes <kbd>k</kbd>'s `ReceivedCharacter`s to stop.

Expected behavior:
1. Hold down <kbd>j</kbd> until it repeats
1. While still holding down <kbd>j</kbd>, hold down <kbd>k</kbd> until that repeats. (<kbd>j</kbd>'s `ReceivedCharacter`s will stop, superseded by <kbd>k</kbd>'s `ReceivedCharacter`s, as expected.)
1. Release <kbd>j</kbd> while still holding <kbd>k</kbd>. <kbd>k</kbd>'s `ReceivedCharacter`s should continue until <kbd>k</kbd> is released.

This causes much frustration to, say, `vim` and `less` users with "lazy fingers" while trying to scroll in a terminal emulator utilizing winit.

> Note: This happens, strangely, only with *some* key combinations. For example, trying the above steps with <kbd>a</kbd> and <kbd>s</kbd> will *not* reproduce this problem.

This is likely to be the cause of https://github.com/alacritty/alacritty/issues/2788

jayache80:
Checking state.is_key_down is redundant, since we are handling keyDown.
Moreover, state.is_key_down can be affected by a previous repeated key's
keyUp.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - **Breaking:** On Web, remove the `stdweb` backend.
 - Added `Window::focus_window`to bring the window to the front and set input focus.
 - On Wayland and X11, implement `is_maximized` method on `Window`.
+- On macOS, fix issue where `ReceivedCharacter` was not being emitted during some key repeat events.
 
 # 0.25.0 (2021-05-15)
 
--- a/src/platform_impl/macos/view.rs
+++ b/src/platform_impl/macos/view.rs
@@ -55,7 +55,6 @@ pub(super) struct ViewState {
     pub cursor_state: Arc<Mutex<CursorState>>,
     ime_spot: Option<(f64, f64)>,
     raw_characters: Option<String>,
-    is_key_down: bool,
     pub(super) modifiers: ModifiersState,
     tracking_rect: Option<NSInteger>,
 }
@@ -74,7 +73,6 @@ pub fn new_view(ns_window: id) -> (IdRef, Weak<Mutex<CursorState>>) {
         cursor_state,
         ime_spot: None,
         raw_characters: None,
-        is_key_down: false,
         modifiers: Default::default(),
         tracking_rect: None,
     };
@@ -516,7 +514,6 @@ extern "C" fn insert_text(this: &Object, _sel: Sel, string: id, _replacement_ran
         let slice =
             slice::from_raw_parts(characters.UTF8String() as *const c_uchar, characters.len());
         let string = str::from_utf8_unchecked(slice);
-        state.is_key_down = true;
 
         // We don't need this now, but it's here if that changes.
         //let event: id = msg_send![NSApp(), currentEvent];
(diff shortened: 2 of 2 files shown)


jayache80:
#1488

kjmph:
Oh super interesting! This is fixed as well with the proposed pull request #1449.

jayache80:
@kjmph Yes- that pull request is frying bigger fish and rips out the “pass_along” kludge altogether. I’ll try to get to a macOS system to build #1449 to verify. If so, then this pull request should be closed

saysjonathan:
#1449 builds for me when merged into current master [38fcceb](https://github.com/rust-windowing/winit/tree/38fccebe1fbc4226c75d6180e5317bd93c024951).

jayache80:
Yes, I also merged master 38fcceb into #1449 and built alacritty with it. It solves the alacritty key repeat issue alacritty/alacritty#2788 that this pull request (#1489) was created to solve, but #1449 does it in a better way. #1449 would also resolve issue #1488.

This pull request (#1489) should be closed in lieu of #1449.

jayache80:
Well, it's been a while, Alacritty *still* has the key repetition bug that would be fixed by this pull-request. This pull-request would've been superseded by #1449 but alas that has yet to be ironed out. I think this should be reconsidered.

ArturKovacs (changes requested):
Hi @jayache80! You are right, this PR should be merged as the other one is stuck currently. However there are two things:

1. Considering that state.is_key_down was only used for this purpse, please remove it's declaration from `ViewState` and all other places it shows up.
2. Given that it's not at all obvious why it is okay to remove this condition, please add the following explanation to the commit message:

```
`is_key_down` was only set to true when `insertText` was called. Therefore
`is_keydown` actually meant to store whether the most recently pressed key
generated an `insertText` event, at which, winit produces a `ReceivedCharacter`.
The issue is that `insertText` is *not* called for “key repeat”, but winit wants
to send `ReceivedCharacter` events for “key repeat” too. To solve this, the
`is_key_down` variable was then checked in the `key_down` function to determine
whether it was valid to produce repeated `ReceivedCharacter` events during “key
repeat”. However this check is not needed since `ReceivedCharacter` must always
be called if the key event has a non-empty `characters` property.

Furthermore `is_key_down` didn’t actually store whether the previously pressed
character had an `insertText` event, because it was incorrectly set to false on
every “key up”. This meant that if two keys were pressed consecutively then the
first was released, then `is_key_down` was set to false even if the most recent
keypress did actually produce an `insertText`.
```

jayache80:
@ArturKovacs I've rebased the branch off latest master and ripped out `is_key_down` from `ViewState` completely. I've included your explanation in the commit message.

ArturKovacs:
Awesome! I forgot one thing last time: please add an entry to `CHANGELOG.md` along the lines of:
```
- On macOS, fixed not emitting `ReceivedCharacter` during some key repeat events.
```

jayache80:
@ArturKovacs 

> please add an entry to `CHANGELOG.md`

Added an entry to the changelog (amended to the commit)

ArturKovacs:
> Is the title of this pull-request still appropriate, or should I edit the title?

It is somewhat obscure but I'm not sure how it could be improved, so I'll leave this up to you

Issue #883, opened 2019-05-25
MouseInput should have a position


JMS55:
A common case is to get the location on screen a user clicks. The way to do that currently is to create a variable that stores the cursor location, which is updated whenever CursorMoved is received. Then, when a MouseInput event is received, you use whatever that variable is currently set to. I think it would be useful for winit to do this internally, so that the MouseInput event could store a position, making it easier for the user. winit should also update the variable whenever Window::set_cursor_position() is called.

vberger:
It may be important to keep in mind that there are literally edge cases to take into account: when the cursor reaches the border of the monitor, `MouseInput` events can keep being generated while the actual location of the cursor does not change.

More generally, there is the question of the behavior of such a field when the cursor is outside of the window, as there will be platform dependent quirks.

JMS55:
Good points. Another problem I thought of is since it uses the location from the previous event poll, if your program is lagging badly, than the mouse will have moved a big distance relative to the previous poll, so it will seem like your input is lagging even more. Not sure if that's how it works, but it seems logical.

aloucks:
This would convolute the the `MouseInput` event.

There is already `WindowEvent::CursorMoved` (in-window only and possibly accelerated) and `DeviceEvent::MouseMotion` (tracks raw movement even when the cursor leaves the window and should not be accelerated). The former is for emulating a cursor and the later is better suited for camera control.

You would need to pick one of these to attach to `MouseInput` and either option would be wrong because there are use cases for both. The mouse position should remain separate from the button input.

DefecateRainbows:
This takes like 25 lines of code for the user to implement and doesn't fall within the scope of things that winit sets out to do. This feels opinionated; it's really up to the user to decide if they want to use the location data when the mouse is clicked.

JMS55:
I do not believe it convolutes MouseInput, in my opinion it is a fairly basic task and it would help for winit to at least show the user how to track MouseInput locations. If that's in the form of an example instead of a MouseInput field, that's alright, but I think winit would benefit from adding a field. The fact that there are edge cases that need to be accounted for just gives weight to including it in winit and preventing the user from messing it up.

In terms of technical feasibility and api choices:
* I admit I don't know the details of MouseInput that well. Can MouseInput events be generated from clicks outside the window?
* CursorMoved vs MouseMotion: If MouseInput's are only generated inside the window, than it makes sense to only use CursorMoved
[...]

PR #1490, opened 2020-03-01
Report mouse motion before click

### LINKED ISSUE #942: No mouse events when unfocused on macOS
This has been reported to Alacritty in https://github.com/jwilm/alacritty/issues/2475.

It looks like winit does not report mouse movements inside the window if it is currently unfocused. This makes it difficult to track mouse click positions since the mouse can "teleport" into position without being able to track it.

If this is an intentional design decision, I think that there's no way around reporting mouse position on click events, since a click might come in without any previous motion events at all.

This could be resolved with https://github.com/rust-windowing/winit/issues/883, but it might still make sense to report motions above the unfocused window, since I'm pretty sure other platforms do that?

chrisduerr:
This fixes an issue on macOS where a mouse click would be generated,
without ever getting a mouse motion to the position before the click.
This leads to the application thinking the mouse click occurred at a
position other than the actual mouse location.

This happens due to mouse motion above the window not automatically
giving focus to the window, unless it is actually clicked, making it
possible to move the window without motion events.

Fixes #942.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 - Revert On macOS, fix not sending ReceivedCharacter event for specific keys combinations.
 - on macOS, fix incorrect ReceivedCharacter events for some key combinations.
 - **Breaking:** Use `i32` instead of `u32` for position type in `WindowEvent::Moved`.
+- On macOS, a mouse motion event is now generated before every mouse click.
 
 # 0.21.0 (2020-02-04)
 
--- a/src/platform_impl/macos/view.rs
+++ b/src/platform_impl/macos/view.rs
@@ -869,26 +869,32 @@ fn mouse_click(this: &Object, event: id, button: MouseButton, button_state: Elem
 }
 
 extern "C" fn mouse_down(this: &Object, _sel: Sel, event: id) {
+    mouse_motion(this, event);
     mouse_click(this, event, MouseButton::Left, ElementState::Pressed);
 }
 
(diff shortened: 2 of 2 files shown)


SergeevPavel:
Could we also trigger mouse motion before mouse wheel event?

chrisduerr:
I've added it for pressure change and scroll too, since that makes some sense.

chrisduerr:
Test failure on Windows should be unrelated to this macOS PR.

ryanisaacg:
I merged master into this branch which should solve the CI conflicts and get this ready for a merge.

QuentinPerez:
I have the same issue thank you for the patch, I was wondering if we could add x and y in the mouseClick event instead of doing magic by saving the last x,y sent by the last Mouse::CursorMoved ?

ryanisaacg:
@QuentinPerez Do you mean in the external Winit API?

QuentinPerez:
Yes, like https://github.com/rust-windowing/winit/blob/849b8f5dce4d3f93b251259f5323ea9fd43e0c05/src/event.rs#L256 but for https://github.com/rust-windowing/winit/blob/849b8f5dce4d3f93b251259f5323ea9fd43e0c05/src/event.rs#L277, don't know if it's the best way to do things.

ryanisaacg:
@QuentinPerez That discussion is probably best for a meta-issue on the tracker, like [this previous issue.](https://github.com/rust-windowing/winit/issues/883). This PR is fixing an internal bug regarding emitting the `CursorMoved` on macOS, specifically.

PR #1459, opened 2020-02-13
Remove assertions from Windows dark mode code

chrisduerr:
In general, winit should never assert on anything unless it means that
it is impossible to continue the execution of the program. There are
several assertions in the Windows dark mode code where this is not the
case.

Based on surface level inspection, all existing assertions could be
easily replaced with just simple conditional checks, allowing the
execution of the program to proceed with sane default values.

Fixes #1458.
Fixes #1405.

PR #1496, opened 2020-03-06
Fix several crashes on Windows by heavily simplifying the event loop code

Osspial:
- [x] Tested on all platforms changed
- [x] Compilation warnings were addressed
- [x] `cargo fmt` has been run on this branch
- [x] `cargo doc` builds successfully
- [ ] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users

I'll do a more detailed writeup on the rationale behind these changes when I'm more awake, but the TL;DR is that it drastically simplifies Winit's event loop model by always using the modal event loop. This PR's fixes are an alternate approach to the problems described in #1429 that simplifies the event model more than #1461 does, but it could not exist in its current form without @filnet's work and feedback in those threads.

Fixes #1477, fixes #1496.

PR #1461, opened 2020-02-13
On Windows, fix request_redraw() related panics

filnet:
~~Simplifies the message loops and event handling considerably~~
~~Strongly inspired from the X11 platform~~

~~Partially reverts https://github.com/rust-windowing/winit/commit/6a330a2894873d29fbbfdeebfc1a215577213996~~

 These panics were introduced by https://github.com/rust-windowing/winit/commit/6a330a2894873d29fbbfdeebfc1a215577213996

Fixes https://github.com/rust-windowing/winit/issues/1391
Fixes https://github.com/rust-windowing/winit/issues/1400
Fixes https://github.com/rust-windowing/winit/issues/1466
Fixes #1469
Probably fixes other related issues

The commit mentioned above introduced a few critical regressions.
I did quite a bit of forensic (looking at 0.19.1, 0.20.0-alpha1) to understand what was going on around the handling of request_redraw(). I also looked at how other platforms handle it.

~~I tried really hard to fix the issues using the current approach of posting WM_PAINT messages but got nowhere. Too many corner cases.
In the end I decided to go for the X11 approach of using a HashSet to keep track of windows requiring a redraw and things got much simpler.~~

This PR does not fundamentally change how request_redraw() is handled.
It still calls `RedrawWindow` which sends a `WM_PAINT` message. Which means that:
- multiple `WM_PAINT` messages are coalesced (for example when calling request_redraw while resizing or calling request_redraw multiple times),
- `RedrawRequested` events are always sent when handling a `WM_PAINT` message.

I tried to be very thorough in testing:
- tested all control flow modes (Poll, Wait and WaitUntill) in both the non-modal and modal loops
- made sure that redraw requests are properly coalesced (in particular when resizing)
- tested multi window and fullscreen

There are a few untested corner cases. See https://github.com/rust-windowing/winit/issues/1429 for more details.

And always kept the specs in a corner of my mind:
- https://docs.rs/winit/0.21.0/winit/window/struct.Window.html#method.request_redraw
- https://docs.rs/winit/0.21.0/winit/event/index.html

- [x] Tested on all platforms changed
- [x] Compilation warnings were addressed
- [x] `cargo fmt` has been run on this branch
- [ ] `cargo doc` builds successfully
- [x] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users
- [ ] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior
- [ ] Created or updated an example program if it would help users understand this functionality
[...]

PR #1500, opened 2020-03-07
Bump version to 0.22.0

chrisduerr:
There are two PRs I'm aware of that should be relatively trivial to get
merged, which would fix some issues. Other than those, I don't think it
makes sense to wait on anything.

 - Fix Windows crash: https://github.com/rust-windowing/winit/pull/1459
 - Fix macOS mouse reports: https://github.com/rust-windowing/winit/pull/1490

While #1459 seems pretty essential to actually make winit run, #1490 is
much less important and can probably be ignored if there aren't any
resources to merge it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# Unreleased
+# 0.22.0 (2020-03-07)
 
 - On Windows, fix minor timing issue in wait_until_time_or_msg
 - On macOS, fix `set_simple_screen` to remember frame excluding title bar.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "winit"
-version = "0.21.0"
+version = "0.22.0"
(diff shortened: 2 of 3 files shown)


kchibisov:
I think it should be done when some arcane crashes on Windows should be fixed. So I feel like waiting a bit for #1496.

Osspial:
I think we should hold off on #1496 for this particular release, since #1461 addresses the same issues but is a bit more stable at present. I'd like to get #1496 in for the next release after this, but the outward-facing bugs it resolves have been resolved so there's less pressure to get it out the door immediately.

kchibisov:
Hm, maybe we should at least take a look at the following PRs before the release

#1501 - Revert of bad fix.
#1483 - Fixes minor API inconsistency.

chrisduerr:
It's not worth waiting even a second for #1483 tbh. If it is merged by the time this is good to go, that's no problem. But this API inconsistency has existed for so long that it really has no urgency at all.

If winit's latest version is usable on all major platforms, then we can talk about fixing things that would be nice to have.

PR #1503, opened 2020-03-07
Fix freeze when pressing modifier keys on Windows

Imberflur:

Fixes deadlock that appears to be introduced by #1381


Osspial:
Looks good. This shouldn't need a changelog entry, since the bug it fixes hasn't appeared in any of our releases yet.

Thanks for pulling this fix together!

PR #1508, opened 2020-03-10
Implement run_embedded and schedule_modal_fn on Windows

Osspial:

Implements request 2 in https://github.com/rust-windowing/winit/issues/159#issuecomment-511124405 and the API mentioned in https://github.com/rust-windowing/winit/pull/638#issuecomment-418535577. This brings us halfway to where we need to be to allow users to create DAW plugins on Windows, the other half being to implement embedded windows as discussed in #403 and #159.

This is a draft PR because it depends heavily on the work in #1496.


kunalarya:
@Osspial -- this is really neat!

I wanted to share some work I've been doing to get child window support for MacOS. The key things I've added is a `EventSubscriber` object, which looks functionally similar to your `EventLoopEmbedded` class?

In any case, I've plumbed these changes into a glutin branch and was successfully able to create child windows in my DAW host. I should note that I'm not a GUI dev, and I'm sure I'm doing something horrible unsafe on this branch; however, it does sketch out a rough prototype:
https://github.com/rust-windowing/winit/compare/master...kunalarya:child-window-mac-os

I'm totally fine updating my branch to use the same or similar API. What's your feeling on the direction here, re child windows?

Issue #1028, opened 2019-07-08
Missing platforms on docs.rs


felixrabe:
For some reason, https://docs.rs/winit (0.20.0-alpha1) only provides the following platforms in the "Platform" dropdown menu:

- i686-windows-pc-msvc
- x86_64-windows-pc-msvc
- x86_64-unknown-linux-gnu

Obviously, there's more missing, but I dearly miss macOS stuff.

Osspial:
I'm unable to find the docs.rs logs for this. Does anyone know where you'd look to see what's going on here?

ryanisaacg:
The mac logs are appearing now, and docs.rs is using a new architecture iirc. Should we close this because it affects an older alpha release?

simlay:
Having iOS (and probably android) platform docs would also be quite useful as well I think. The mobile ecosystem is slowly growing (https://github.com/hecrj/iced/pull/57)

Osspial:
@ryanisaacg Let's leave this open, since the docs currently aren't generated for mobile or the web.

Does anyone know of any way to have docs.rs generate docs for mobile and the web? I'm not entirely sure that's possible right now.

ryanisaacg:
I don't think we can convince it to? Another option would be a github pages branch on winit's repo with custom-built docs, but that would be a fair bit of effort.

simlay:
Looks like according to https://github.com/rust-lang/docs.rs/issues/343 there are currently only 6 targets supported (none of which are Android, iOS or WASM) for which docs are built. Though, it looks https://github.com/rust-lang/docs.rs/pull/346 adds support for a `extra_targets` so, that in theory would conveniently allow for mobile and web docs when it's landed and deployed.

jyn514:
Also we have a default platform limit of 10, but we can raise that if you have a use-case for more platforms.

simlay:
So, I looked into getting more of the docs built. Currently on docs.rs, the builds are the default (`i686-pc-windows-msvc`, `x86_64-pc-windows-msvc`, `x86_64-unknown-linux-gnu`, and `x86_64-apple-darwin`) which is everything except `i686-unknown-linux-gnu` at least according to the GUI.

Here's the [build log trace](https://gist.github.com/simlay/19e7972ca6ed35301033416c8faba9c3) for `i686-unknown-linux-gnu`. Due to the layers of docker, I'm not really sure what it's missing. 

[...]

PR #1521, opened 2020-04-03
Added more docs.rs targets

simlay:

This PR adds `wasm32-unknown-unknown` to docs.rs and explicitly states which targets are built for docs.rs. I tested this by doing https://github.com/rust-windowing/winit/issues/1028#issuecomment-607873122 and had mixed results with all of our targets as talked about in https://github.com/rust-windowing/winit/issues/1028#issuecomment-608645825.

`i686-unknown-linux-gnu` build is currently broken. Fix requires a PR to https://github.com/rust-lang/crates-build-env.

It doesn't look like docs.rs supports feature flag ops so I just chose `web-sys` as a default feature for all docrs targets.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,9 @@ documentation = "https://docs.rs/winit"
 categories = ["gui"]
 
 [package.metadata.docs.rs]
-features = ["serde"]
+features = ["serde", "web-sys"]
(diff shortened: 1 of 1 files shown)


Osspial:
Looks good. Thanks for pulling this together!

PR #1524, opened 2020-04-06
Make drag and drop optional (fixes OleInitialize failure #1255)

chemicstry:
Since #1260 is stalled and based on discussion there I suggest an alternative fix for the `OleInitialize` problem (#1255). This makes drag and drop optional and allows `winit` to be used together with libraries like `cpal` without drag and drop functionality. I also improved the panic message as previous was really not informative.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 - On Unix, X11 and Wayland are now optional features (enabled by default)
 - On X11, fix deadlock when calling `set_fullscreen_inner`.
 - On Web, prevent the webpage from scrolling when the user is focused on a winit canvas
-
+- On Windows, drag and drop is now optional and must be enabled with `WindowBuilderExtWindows::with_drag_and_drop(true)`.
 - On Wayland, fix deadlock when calling to `set_inner_size` from a callback.
 - On macOS, add `hide__other_applications` to `EventLoopWindowTarget` via existing `EventLoopWindowTargetExtMacOS` trait. `hide_other_applications` will hide other applications by calling `-[NSApplication hideOtherApplications: nil]`.
 - On android added support for `run_return`.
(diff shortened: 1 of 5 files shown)


Dooskington:
Keeping an eye on this, I am patiently waiting to be able to use `cpal` and `winit` together :)

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Unreleased
 
 - On X11, fix `ResumeTimeReached` being fired too early.
+- On Windows, drag and drop is now optional and must be enabled with `WindowBuilderExtWindows::with_drag_and_drop(true)`.
```
Osspial, line 4: Drag and drop should be opt-out, not opt-in, for the reasons I brought up in https://github.com/rust-windowing/winit/pull/1260#pullrequestreview-324846105.

chemicstry:
Fixed as requested

Uriopass:
Any update on this ? Using `cpal` with `winit` is a pain on windows at the moment. :-)

src/platform/windows.rs
```diff
@@ -117,6 +117,12 @@ pub trait WindowBuilderExtWindows {
 
     /// This sets `WS_EX_NOREDIRECTIONBITMAP`.
     fn with_no_redirection_bitmap(self, flag: bool) -> WindowBuilder;
+
+    /// Enables or disables drag and drop support (enabled by default). Will interfere with other crates
+    /// that use multi-threaded COM API (`CoInitializeEx` with `COINIT_MULTITHREADED` instead of
+    /// `COINIT_APARTMENTTHREADED`) on the same thread.
+    /// See https://docs.microsoft.com/en-us/windows/win32/api/objbase/nf-objbase-coinitialize#remarks for more information.
```
Osspial, line 124: One more note that's occurred to me since the initial review, and that should make it into the documentation:

Even with `with_drag_and_drop` set to `false`, Winit can still call `CoInitializeEx(COINIT_APARTMENTTHREADED)`. Right now that only happens if the user fullscreens a window (see the [`com_initialized`](https://github.com/rust-windowing/winit/blob/878c179761cbc536ad24906486430ed2bb1d6092/src/platform_impl/windows/window.rs#L833-L835) function and the functions that call it), but I can't make any guarantees about whether or not it'll be called elsewhere in Winit's codebase in the future. If the user needs to initialize COM with `COINIT_MULTITHREADED` they should do it before calling any Winit functions, just to be safe.

chemicstry:
Can this be re-reviewed?

PR #1538, opened 2020-04-17
wayland: rework scale factor handling

Matthias-Fauconneau:
wayland: rework scale factor handling

Fixes some hidpi use cases on wayland :
- initialize with primary scale factor (avoid flicker)
- also call Resize on scale factor changes
 apps usually resize their swap chains there and don't handle the explict scalefactorchanged.
 As discussed with kchibisov, calling this on any scale change even if it happens to yield same physical size (e.g move window from 4K laptop to 4K TV), as a defense in depth to try to trigger any application UI/font scale redraw (even when they don't explicitly handle scalefactorchanged).
This might make applications do an identical swap chain resize as well if they don't check, but that's okay.
- minor fix: inner_size within Resize callback (not hdpi related)

Fixes Iced

### DIFF
--- a/src/platform_impl/linux/wayland/event_loop.rs
+++ b/src/platform_impl/linux/wayland/event_loop.rs
@@ -736,11 +736,11 @@ impl<T> EventLoop<T> {
                         let physical_size = logical_size
                             .to_physical(window.new_dpi.unwrap_or(window.prev_dpi) as f64);
 
+                        *window.size = (w, h);
                         callback(Event::WindowEvent {
                             window_id,
                             event: WindowEvent::Resized(physical_size),
                         });
-                        *window.size = (w, h);
                     }
                 }
 
--- a/src/platform_impl/linux/wayland/event_loop.rs
+++ b/src/platform_impl/linux/wayland/event_loop.rs
@@ -716,11 +716,11 @@ impl<T> EventLoop<T> {
                 crate::window::WindowId(crate::platform_impl::WindowId::Wayland(window.wid));
 
             // Update window logical .size field (for callbacks using .inner_size)
-            let (old_size, mut logical_size) = {
+            let (old_logical_size, mut logical_size) = {
                 let mut window_size = window.size.lock().unwrap();
-                let old_size = *window_size;
-                *window_size = window.new_size.unwrap_or(old_size);
-                (old_size, *window_size)
+                let old_logical_size = *window_size;
+                *window_size = window.new_size.unwrap_or(old_logical_size);
+                (old_logical_size, *window_size)
             };
 
             if let Some(scale_factor) = window.new_scale_factor {
@@ -767,7 +767,7 @@ impl<T> EventLoop<T> {
                 }
                 // Don't send resize event downstream if the new logical size and scale is identical to the
                 // current one
-                if logical_size != old_size || window.new_scale_factor.is_some() {
+                if logical_size != old_logical_size || window.new_scale_factor.is_some() {
                     let physical_size = LogicalSize::<f64>::from(logical_size).to_physical(
                         window.new_scale_factor.unwrap_or(window.prev_scale_factor) as f64,
                     );
--- a/src/platform_impl/linux/wayland/event_loop.rs
+++ b/src/platform_impl/linux/wayland/event_loop.rs
@@ -731,11 +731,12 @@ impl<T> EventLoop<T> {
                     .update_scale_factor(scale_factor as u32);
                 let new_logical_size = {
                     let scale_factor = scale_factor as f64;
-                    let mut physical_size = LogicalSize::<f64>::from(logical_size).to_physical(scale_factor);
+                    let mut physical_size =
+                        LogicalSize::<f64>::from(logical_size).to_physical(scale_factor);
                     callback(Event::WindowEvent {
                         window_id,
                         event: WindowEvent::ScaleFactorChanged {
-                            scale_factor: scale_factor,
+                            scale_factor,
                             new_inner_size: &mut physical_size,
                         },
                     });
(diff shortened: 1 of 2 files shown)


kchibisov:
It seems like this it is this bug  #1451 , right? I completely forget about this issue in winit, thx.

Matthias-Fauconneau:
Indeed. Could have saved some time if this had gotten fixed right away. Can't see any reason why you'd want the old size :/

kchibisov:
I don't see any, but it was opened as a followed up to some other bug, which was closed for some reason, and I forgot about this issue, since I never use `inner_size`.

Matthias-Fauconneau:
Okay, I'm not familiar with winit, already had a hard time enough trying to
figure out why the whole code is so complicated and what is the expected
behaviour across platforms, this works for me for now, so I'll let you take
someone else take over if there is a more proper fix to be figured out :).

kchibisov:
I'm not sure what is working here, but on your PR I get missmatch when running `glutin` window example, so it's not a fix.

```
INNER SIZE PhysicalSize { width: 1278, height: 1280 }
PHysical size PhysicalSize { width: 1280, height: 1280 }
```

Matthias-Fauconneau:
You are right, it doesn't. I must have messed up my setup and testing
a binary which still had the ScaleFactor handler by mistake ><

src/platform_impl/linux/wayland/window.rs
```diff
@@ -59,15 +59,25 @@ impl Window {
...
+            window_store
+                .lock()
+                .unwrap()
+                .scale_factor_change(&surface, scale_factor);
+            surface.set_buffer_scale(scale_factor);
         });
 
-        let dpi = get_dpi_factor(&surface) as f64;
+        let scale_factor = {
+            if let Some(output) = get_outputs(&surface).last() {
```
kchibisov, line 71: This is incorrect, initial scale factor on Wayland is `1`and we should trust compositor events, don't guess it by monitor, this won't work. Right now I create a wrong window size on startup, which is larger then my output, because I'm drawing 8k window on 4k screen with your PR. For more https://gitlab.freedesktop.org/wayland/wayland/issues/133

src/platform_impl/linux/wayland/event_loop.rs
```diff
@@ -714,8 +714,43 @@ impl<T> EventLoop<T> {
...
+            let (old_size, mut logical_size) = {
+                let mut window_size = window.size.lock().unwrap();
+                let old_size = *window_size;
+                *window_size = window.new_size.unwrap_or(old_size);
+                (old_size, *window_size)
+            };
+
+            if let Some(scale_factor) = window.new_scale_factor {
+                // Update cursor scale factor
+                self.cursor_manager
```
kchibisov, line 728: Have you checked that lock will be released?

src/platform_impl/linux/wayland/event_loop.rs
```diff
@@ -714,8 +714,43 @@ impl<T> EventLoop<T> {
...
                 crate::window::WindowId(crate::platform_impl::WindowId::Wayland(window.wid));
-            if let Some(frame) = window.frame {
-                if let Some((w, h)) = window.newsize {
+
+            // Update window logical .size field (for callbacks using .inner_size)
+            let (old_size, mut logical_size) = {
+                let mut window_size = window.size.lock().unwrap();
+                let old_size = *window_size;
+                *window_size = window.new_size.unwrap_or(old_size);
+                (old_size, *window_size)
```
kchibisov, line 723: It should be `s/old_size/old_logical_size/` or `s/window_size/new_size/` for consistency.

src/platform_impl/linux/wayland/window.rs
```diff
@@ -512,26 +523,25 @@ impl WindowStore {
         F: FnMut(WindowStoreForEach<'_>),
     {
         for window in &mut self.windows {
+            let prev_scale_factor = window.current_scale_factor;
+            if let Some(scale_factor) = window.new_scale_factor {
+                window.current_scale_factor = scale_factor;
```
kchibisov, line 528: I'm not really a fan that you're updating scale factor right here, but sending new event in event loop. However I feel like it's already complex to handle such things correctly in a callback, so add a comment at least explain that scale factor change is the first event.

src/platform_impl/linux/wayland/event_loop.rs
Matthias-Fauconneau: Is there a way the guard could escape ?

kchibisov: I mean the problem is knowing exactly the time lock will be dropped, I'm just ensuring that it'll be drop in the right time. I feel like lock will be acquired until the end of scope, which is after `callback`? At least I seen issues similar to that already.

src/platform_impl/linux/wayland/window.rs
Matthias-Fauconneau: I just moved it from after to before the callback.
I don't know what is the split between window/event_loop for and the criteria for what should be on either side ?

src/platform_impl/linux/wayland/event_loop.rs
Matthias-Fauconneau: The update is scoped in a block so that the mutex is unlocked there when the mutex guard drops out of the block scope.

src/platform_impl/linux/wayland/window.rs
kchibisov: The goal is to not go out of sync with the library user, right now it's fine, and knowing that backend will be changed anyway let's just keep it.

src/platform_impl/linux/wayland/event_loop.rs
kchibisov: yeah, it's fine. agree.

src/platform_impl/linux/wayland/window.rs
Matthias-Fauconneau: Okay, not clear if/how there is a way to avoid drawing a buffer with the wrong scale.
The get_dpi_factor is misleading since it use surface enter to compute dpi.
That's why I thought we are trying to get a better guess but failing.
Changed the init to make it clear we use an initial 1x scale factor.

kchibisov: >Okay, not clear if/how there is a way to avoid drawing a buffer with the wrong scale.

I've posted you a link, which is telling that it's impossible due to protocols.

>Changed the init to make it clear we use an initial 1x scale factor.

Please, return to old logic, since it was a correct logic and more clear, just hard coding `1` isn't an option. You can add a comment explaining that most of the time it'll be `1` due to how protocols are designed now, and that we get proper scale factor after `enter`.

Matthias-Fauconneau: I feel like if there is a single output we can't go wrong using it.
Reverted to scale_factor=1 constant variable as requested.

vberger:
Thanks!

PR #1542, opened 2020-04-20
[Windows] Allow Windows ALT-Tab behaviour for fullscreen windows

### LINKED ISSUE #1540: [Windows 10] Fullscreen does not interact well with alt-tab
When alt-tabbing to another window while a winit window is fullscreen (either kind) the winit window keeps rendering on top.

## Steps to reproduce
* Checkout latest winit, ie d5609729cc5c23a1d74c8c25d3af9c3bc7512a6f
* `cargo run --example fullscreen`
* Select either **1** or **2**
* Select monitor _ie 0_
* Try to alt-tab into another window.

downstream: https://github.com/big-ab-games/robo-instructus/issues/235

antonmarsden:
Fullscreen windows were forced to be ALWAYS_ON_TOP, meaning that ALT-Tab behaviour did not work as expected. This change enables ALT-Tab behaviour by default. It works really well for windowed borderless mode.

In fullscreen exclusive mode, the behaviour is less than ideal if using a different resolution, i.e., the user will see their desktop in the fullscreen window's resolution when switching to a desktop window. However, note that explicit use of set_always_on_top(true) can be used to revert to the previous ALT-Tab behaviour.

Fixes #1540 


antonmarsden:
> Thanks for pulling this PR together!
> 
> I think this fix is fine as-is for borderless fullscreen windows, but for exclusive fullscreen windows we should enable the `ALWAYS_ON_TOP` flag while the window is focused. Once that's implemented, this should be good to merge.

@Osspial  This appears to  be the current behaviour, at least on my Windows 10 machine, i.e.,
- if I ALT-tab to the fullscreen window it comes to the foreground and the taskbar disappears
- if I click in the fullscreen window (when it's in the background) it comes to the foreground and the taskbar disappears.

Did you have a different experience? Are there any other test cases I haven't considered?

bernastoc:
@antonmarsden I think what @Osspial means is that your fix should only apply to fullscreen windows of variant Borderless, not Exclusive (https://docs.rs/winit/0.22.1/winit/window/enum.Fullscreen.html).

antonmarsden:
@bernastoc That wasn't quite my intepretation. It seems that @Osspial is suggesting that ALWAYS_ON_TOP may need to be true _while the window is focused_. I don't mind working on a solution, but would need a use case to work from - I cannot find any issues with the fix, but I may have missed something.

kchibisov:
>@bernastoc That wasn't quite my intepretation. It seems that @Osspial is suggesting that ALWAYS_ON_TOP may need to be true while the window is focused. I don't mind working on a solution, but would need a use case to work from - I cannot find any issues with the fix, but I may have missed something.

no, it clearly states that for `Fullscreen::Exclusive` it must be always on top, and for `Fullscreen::Borderless` it should be like in your PR. The use case is that exclusive fullscreen performs a modeset, and so showing other apps above it isn't something that could work, since it's usually done by system's compositor, meaning that you should exit exclusive fullscreen.

Boscop:
I'm on Win 8.1.
I just updated to glium 0.28 and Alt-Tabbing away from a borderless fullscreened window still doesn't work.
The window chooser overlay appears to change to the new window **as if** it's actually switching windows, and while I keep Alt pressed, the new window is shown, but when I release Alt, it's back to the fullscreen winit/glutin window.
Any idea why? :)

antonmarsden:
> I'm on Win 8.1.
> I just updated to glium 0.28 and Alt-Tabbing away from a borderless fullscreened window still doesn't work.
> The window chooser overlay appears to change to the new window **as if** it's actually switching windows, and while I keep Alt pressed, the new window is shown, but when I release Alt, it's back to the fullscreen winit/glutin window.
> Any idea why? :)

Perhaps discuss in the #1740 ticket, given that this pull never made the cut?

alexheretic:
It should be fixed now just not released yet, try the latest commit.

Boscop:
@alexheretic Thanks. I just tried with the latest commit (`rev = "793da7d"`) but it's still the same behavior:
When trying to Alt-Tab towards my glutin window, while I keep Alt pressed (after pressing TAB enough times to select it), my glutin window is shown in front but as soon as I release the Alt key, the window that was previously in front it back in front and my glutin window is back at its previous "z-index" in the window order: When I then try to Alt-Tab to it again, I have to press TAB the same number of times again to select it. (And it again disappears when releasing Alt.)

PR #1544, opened 2020-04-21
Sctk8

Matthias-Fauconneau:
Initial draft of a potential port of wayland backend to smithay-client-toolkit ^0.8.
Publishing this draft pull request for inspiration not ready for review.
Also feature gates x11.


vberger:
Hi @Matthias-Fauconneau, and thanks for taking the effort of tackling this task which is much needed!

I am the one who wrote a large part of this code, so let me give you some advice regarding the update of winit:

I just released wayland-rs 0.26, calloop 0.6 and SCTK 0.9, you should probably aim directly for them as they don't introduce a lot of breaking changes compared to the previous version and they drop the `mio` dependency.

In addition to the tedious but mechanical updates linked to the new APIs, SCTK 0.8 introduced a significant change that will directly impact winit:  `sctk::window::Window` is no longer `Send`. The fact that it was before was actually unsound.

This will imply some deep changes in winit's code, as currently its `Window` (which need to be `Send`) own the sctk window. This will thus have to change. My first idea was to centralize the codebase much more around `EventLoop`, and make the `Window` into thin handles that just forward their commands to the `EventLoop` (via channels or shared buffers).

This of course means a significant refactor of the Wayland backend, but it should at the same time allow to significantly simplify it.

Now, if knowing that you still want to take care of this, don't hesitate to ping me here or in the winit/glutin gitter channel (I'm Levans there) if you have any question!

Matthias-Fauconneau:
Hi!

Thanks for the detailed information.
Thanks for your work on Wayland.

I'm not familiar with Send. What prevents it from being Send ?
Looks like some of that EventLoop/Window split is kind of being done with
Option<> signals ><.

I've put this on hold for now as I'm pursuing another solution which would
suit my need as well (doing a direct SCTK backend in iced)
I'm currently trying to forge the most elegant event loop using async
(smol@stjepang)

Maybe some of the work I do in iced_sctk can be easily translated back.
I might come back to this if there is significant interest in an uptodate
winit_sctk backend (maybe with smol as well ;).
But I'd be even happier to see someone else do it >< :P

[https://github.com/hecrj/iced/pull/311]

vberger:
> I'm not familiar with Send. Is it because Filter is not send ? Why ? to take Rcs ? would Arcs fix this or there are deeper issues ?

There is a deeper issue: when a new wayland object is created, it must be done from the same thread as the one hosting the event queue that will manage it, as the internal logic of event queues is not thread-safe. This is an internal constraint of the system `libwayland-client`, and my Rust wrappers cannot do much about that. This is to manage that that I introduced the `Main<_>` and `Attached<_>` adapters in wayland-client.

SCTK's `Window`, to draw its decorations, need to create `wl_buffer`s. It also interacts with the `wl_seat` to manage the interactivity of its decorations. As a result it must be on the same thread as the one hosting the event queue.

Matthias-Fauconneau:
Ok, I'm not clear on the issue.
My understanding after a bit of spelunking, please correct me :
VK_KHR_surface provides platform-agnostic swapchain management.
It requires the implementation (e.g mesa) to peek at the wayland socket.
libwayland-client is the platform-aware interface to provide that.

For more sublime elegance, it would be nice to abstract this interface.
Ideally simply a shared abstraction with mesa.
In the meantime, maybe xdg-foreign could be used to spit libwayland-client
out of the Rust socket/process.

Regardless, is there any use case for a Send wl_buffer ?
Why is libwayland-client not supporting it ? or is this only when using it
through wayland-rs ?
VK_KHR_surface seems to allow for arbitrary multithread multiqueue choas

vberger:
No the issue is not that `wl_buffer` is not `Send`, it is. The issue is that creating a `wl_buffer` must be done on the same thread that the one hosting its `wl_event_queue`, otherwise the creation is racy and could result in lost events or a corrupted protocol state.

Now, Mesa internally uses its own `wl_event_queue` (that it creates from the `wl_display` you provide it at initialization), meaning there is no issue from that side. This is generally the recommended way to share a Wayland connection between different independent libraries: each one has its own `wl_event_queue` managing its objects.

Matthias-Fauconneau:
Okay, isn't that already handled by the Main/Attached logic ?

vberger:
It is, that is why the `Main<_>` and `Attached<_>` handles are `!Send`.

And given SCTK's `Window` needs to keep an `Attached<WlShmPool>` in order to create new buffers to draw the decorations, and well as manipulate `Attached<WlSeat>` to process pointer events to make these borders interactive to move and resize, it thus cannot be `Send`.

Matthias-Fauconneau:
@vberger Would be nice if you could have a look how it looks now ?
I'm hanging right now on :
```
unresolved import `super::event_loop::Env`
  --> winit/src/platform_impl/linux/wayland/window.rs:32:55
32 | use super::{Update, event_loop::{DispatchData, State, Env}, EventLoopWindowTarget, MonitorHandle, WindowId};
   |                                                       ^^^ no `Env` in `platform_impl::platform::wayland::event_loop`
```
default_environment! should be pub struct Env though. I don't know if I'm missing something obvious, a macro hygiene thing or something.

I'm switching back to direct iced_sctk for now in the meantime (since even with an updated winit, it will also have value for iced layer-shell as discussed)

vberger:
> default_environment! should be pub struct Env though. I don't know if I'm missing something obvious, a macro hygiene thing or something.

I think this might be related with this other error:

```
error: cannot find macro `default_environment` in this scope
   --> src/platform_impl/linux/wayland/event_loop.rs:25:1
    |
25  | default_environment!(Env, desktop,
    | ^^^^^^^^^^^^^^^^^^^ help: a macro with a similar name exists: `init_default_environment`
    |
```

Matthias-Fauconneau:
Ah, i should remember to try without stopping on first error sometimes 🙄

Matthias-Fauconneau:
I'm not sure how to test now.
glutin_examples --example window works.
There is an HiDPI bug somewhere in that test. Not sure if that was also before.
Need to check if it's in winit, glutin or that example.

kchibisov:
try to run alacritty some other apps using winit.

fengalin:
@Matthias-Fauconneau thanks for tackling this. I tried playing with your branch but got error compiling.

There seems to be a left-over from a merge:

```
error: expected expression, found `>>`
   --> src/platform_impl/linux/wayland/window.rs:222:3
    |
222 | >>>>>>> a1b03e4... wayland: Rewrite for smithay-client-toolkit 0.8 (#1544)
    |   ^^ expected expression
```
`Handle::with` is missing, and I think there was an issue merging because `struct WindowStoreForEach` is declared in the middle of `impl Handle { .. }`:

```
error[E0599]: no method named `with` found for reference `&platform_impl::platform::wayland::window::Handle` in the current scope
   --> src/platform_impl/linux/wayland/window.rs:189:58
    |
189 |     pub fn inner_size(&self) -> PhysicalSize<u32> { self.with(|window| LogicalSize::<f64>::from(window.size).to_physical(self.scale_factor() as f64)) }
    |                                                          ^^^^ method not found in `&platform_impl::platform::wayland::window::Handle`
```
IMHO, this PR should not rely on nightly features:

```
error[E0554]: `#![feature]` may not be used on the stable release channel
   --> src/lib.rs:135:31
    |
135 | #![cfg_attr(target_os="linux",feature(trait_alias,vec_remove_item,bindings_after_at))]
```

Matthias-Fauconneau:
I fixed this with the second force push.
Yes, the nightly features can be workaround eventually. I find it easier to develop with at first. :)

kchibisov:
The maximum version this PR can use is 1.41.0. I can't review such big PR now (likely won't be able until July), but please, avoid changes outside of Wayland backend, otherwise I'll ask to revert them from this PR anyway, you can send them later, if you really want to, but I won't review new feature flags additions/changes outside of Wayland backend, etc here. Just saying in general.

Also, ensure that you're not breaking winit thread safity guarantees and that you're still using Wayland only with `dlopen` approach.

src/platform_impl/linux/wayland/window.rs
```diff
@@ -465,94 +186,44 @@ pub struct WindowStoreForEach<'a> {
...
-                replace(&mut *window.need_refresh.lock().unwrap(), false),
-                replace(&mut *window.need_frame_refresh.lock().unwrap(), false),
-                make_wid(&window.surface),
-                frame,
-            );
-        }
+    pub fn set_cursor_visible(&self, visible: bool) {
+        //self.cursor_visible = visible; self.set(Cursor(if self.cursor_visible { self.cursor } else { "" }))
+        self.set(Cursor(if visible { "left_ptr" } else { "" })) // Assume default icon
+>>>>>>> a1b03e4... wayland: Rewrite for smithay-client-toolkit 0.8 (#1544)
```
fengalin, line 222: > I fixed this with the second force push.

Problem still present in the remote branch AFAICT.

Matthias-Fauconneau:
Sorry, should be fixed now. I had done commit --amend instead of -a
--amend. So I guess before it just changed the timestamp and was happy to
force push that change ><

returntrip:
I am trying to package `winit` for Fedora. I am using sctk 0.9. I guess when this is commited I should not see any of these errors?

```
error[E0432]: unresolved import `smithay_client_toolkit::pointer`
  --> src/platform_impl/linux/wayland/event_loop.rs:23:29
   |
23 | use smithay_client_toolkit::pointer::{AutoPointer, AutoThemer};
   |                             ^^^^^^^ could not find `pointer` in `smithay_client_toolkit`

error[E0432]: unresolved imports `smithay_client_toolkit::output::OutputMgr`, `smithay_client_toolkit::Environment`
  --> src/platform_impl/linux/wayland/event_loop.rs:48:5
   |
48 |     output::OutputMgr,
   |     ^^^^^^^^^^^^^^^^^ no `OutputMgr` in `output`
...
53 |     Environment,
   |     ^^^^^^^^^^^
   |     |
   |     no `Environment` in the root
   |     help: a similar name exists in the module: `environment`

error[E0432]: unresolved imports `smithay_client_toolkit::keyboard`, `smithay_client_toolkit::keyboard`
 --> src/platform_impl/linux/wayland/keyboard.rs:6:9
  |
5 |     keyboard::{
  |     ^^^^^^^^ could not find `keyboard` in `smithay_client_toolkit`
6 |         self, map_keyboard_auto_with_repeat, Event as KbEvent, KeyRepeatEvent, KeyRepeatKind,
  |         ^^^^ no `keyboard` in the root

error[E0432]: unresolved import `smithay_client_toolkit::keyboard`
   --> src/platform_impl/linux/wayland/keyboard.rs:220:33
    |
220 |     use smithay_client_toolkit::keyboard::keysyms;
    |                                 ^^^^^^^^ could not find `keyboard` in `smithay_client_toolkit`

error[E0432]: unresolved imports `smithay_client_toolkit::output::OutputMgr`, `smithay_client_toolkit::surface::get_dpi_factor`, `smithay_client_toolkit::surface::get_outputs`, `smithay_client_toolkit::window::ConceptFrame`, `smithay_client_toolkit::window::Theme`
  --> src/platform_impl/linux/wayland/window.rs:21:5
   |
21 |     output::OutputMgr,
   |     ^^^^^^^^^^^^^^^^^ no `OutputMgr` in `output`
...
26 |     surface::{get_dpi_factor, get_outputs},
   |               ^^^^^^^^^^^^^^  ^^^^^^^^^^^ no `get_outputs` in `surface`
   |               |
   |               no `get_dpi_factor` in `surface`
27 |     window::{ConceptFrame, Event as WEvent, State as WState, Theme, Window as SWindow},
   |              ^^^^^^^^^^^^ no `ConceptFrame` in `window`      ^^^^^ no `Theme` in `window`

error[E0432]: unresolved import `smithay_client_toolkit::window::Theme`
 --> src/platform/unix.rs:5:70
  |
5 | use smithay_client_toolkit::window::{ButtonState as SCTKButtonState, Theme as SCTKTheme};
  |                                                                      ^^^^^^^^^^^^^^^^^^ no `Theme` in `window`

error[E0432]: unresolved import `instant`
  --> src/event.rs:36:5
   |
36 | use instant::Instant;
   |     ^^^^^^^ use of undeclared type or module `instant`

error[E0432]: unresolved import `instant`
  --> src/event_loop.rs:12:5
   |
12 | use instant::Instant;
   |     ^^^^^^^ use of undeclared type or module `instant`

error[E0603]: module `surface` is private
  --> src/platform_impl/linux/wayland/pointer.rs:16:29
   |
16 | use smithay_client_toolkit::surface;
   |                             ^^^^^^^ private module
   |
note: the module `surface` is defined here
  --> /usr/share/cargo/registry/smithay-client-toolkit-0.9.1/src/lib.rs:52:1
   |
52 | mod surface;
   | ^^^^^^^^^^^^

error[E0603]: module `surface` is private
  --> src/platform_impl/linux/wayland/touch.rs:8:29
   |
8  | use smithay_client_toolkit::surface;
   |                             ^^^^^^^ private module
   |
note: the module `surface` is defined here
  --> /usr/share/cargo/registry/smithay-client-toolkit-0.9.1/src/lib.rs:52:1
   |
52 | mod surface;
   | ^^^^^^^^^^^^

error[E0603]: module `surface` is private
  --> src/platform_impl/linux/wayland/window.rs:26:5
   |
26 |     surface::{get_dpi_factor, get_outputs},
   |     ^^^^^^^ private module
   |
note: the module `surface` is defined here
  --> /usr/share/cargo/registry/smithay-client-toolkit-0.9.1/src/lib.rs:52:1
   |
52 | mod surface;
   | ^^^^^^^^^^^^

error: aborting due to 11 previous errors

Some errors have detailed explanations: E0432, E0603.
For more information about an error, try `rustc --explain E0432`.
error: could not compile `winit`.
```

kchibisov:
How fedora is packaging rust? I'd assume it's trying to have only one crate at the time, which is nearly impossible in rust ecosystem now. Please don't ship downstream apps with this branch, it wasn't tested, reviewed, etc.

returntrip:
> How fedora is packaging rust? I'd assume it's trying to have only one crate at the time, which is nearly impossible in rust ecosystem now.

I am new to packaging,  for what I understand, for Fedora, the crates are packaged individually/separately and they are packaged at their latest version. It is also possible to create a tarball with all the crates but this is not the preferred way. I have not attempted to create the tarball yet but I might do, since I am not able to package winit > glutin > alacritty.

> Please don't ship downstream apps with this branch, it wasn't tested, reviewed, etc.

Of course I wouldn't, hence my question about this PR fixing my packaging issue :).

kchibisov:
>I am new to packaging, for what I understand, for Fedora, the crates are packaged individually/separately and they are packaged at their latest version. It is also possible to create a tarball with all the crates but this is not the preferred way. I have not attempted to create the tarball yet but I might do, since I am not able to package winit > glutin > alacritty.

If the goal is to package alacritty, you may actually use older SCTK, I'm not even sure that there's any other apps that it being shipped somewhere with winit + glutin + sctk combo. Vendoring all crates on the other side is a +- good approach and is what most distros are doing.

kchibisov:
@Matthias-Fauconneau 

Are you going to continue the work on that PR or I can pick this task of updating sctk to new versions myself (I should have plenty of time now to do so)?

In general it requires a rebase + cleaning of pointless comments.

Likely clippy related things should be removed, since those should be handled separately, otherwise it's very hard to review code.

Also have you been able to test things out?

Matthias-Fauconneau:
You're very welcome to finish this.

(I'm doing my current project with sctk directly).

IIRC last time I checked the glutin window example was working.

kchibisov:
Thanks for putting this PR, it helped me with the rework in general

I'm closing this in favor of #1653, since it's basically done and waiting on SCTK/calloop fixes, on which I'm currently working on. Once I'll address them, I'll give it source level refactoring and documentation update, once it'll be done it should be good to land, since I've been running it for some time in alacritty without much issues.

PR #1551, opened 2020-04-30
iOS: convert touch positions to physical

francesca64:
Looks like this got missed whenever the DPI revamp happened. Touch works great on iOS with this change, though! 

Tested on iPhone XR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 - On Windows, fix `WindowBuilder::with_maximized` being ignored.
+- On iOS, touch positions are now properly converted to physical pixels.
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Unreleased
 
 - On Windows, fix `WindowBuilder::with_maximized` being ignored.
+- On iOS, touch positions are now properly converted to physical pixels.
```
Gordon-F, line 4: Looks like breaking changes.

francesca64: I'm merely fixing a violation of our existing API guarantees. If anyone's relying on the current behavior, they're using the API incorrectly.

PR #1581, opened 2020-05-28
Macos fullscreen & dialog support with `run_return`

VZout:
Previously the `run_return` example would freeze if you switch to fullscreen. This fixes it.
It is not a clean solution but if anybody knows more about mac than I do let me know if there is a better way.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 
 - On Wayland, fix deadlock when calling to `set_inner_size` from a callback.
 - On macOS, add `hide__other_applications` to `EventLoopWindowTarget` via existing `EventLoopWindowTargetExtMacOS` trait. `hide_other_applications` will hide other applications by calling `-[NSApplication hideOtherApplications: nil]`.
+- On MacOS, Fixed fullscreen and dialog support for `run_return`.
 
 # 0.22.2 (2020-05-16)
 
--- a/src/platform_impl/macos/app_state.rs
+++ b/src/platform_impl/macos/app_state.rs
@@ -223,7 +223,7 @@ impl Handler {
     }
 }
 
-pub static TRANSITION_FULLSCREEN_MODE: AtomicBool = AtomicBool::new(false);
+pub static INTERRUPT_EVENT_LOOP_EXIT: AtomicBool = AtomicBool::new(false);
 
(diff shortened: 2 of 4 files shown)


VZout:
Rewrote my approach since I noticed I caused a bug with the control flow.

VZout:
Also added experimental dialog/child window support tested with https://github.com/EmbarkStudios/nfd2.

VZout:
@kchibisov @francesca64  You two originally authored the `run_return` implementation for Macos. What do you guys think about this issue?

kchibisov:
I don't have native macOS machine and don't know much about macOS. The run_return impl is a copypasta of run impl. I'm not sure If I ever seen a freeze from run_return, since we're using it for so much time in alacritty, and no one ever reported it. Maybe I don't understand the conditions where it should freeze?

VZout:
@kchibisov So its 2 issues. 1 is running the `window_run_return` sample and pressing the green fullscreen button on mac. It completely freezes the app. The second is that opening dialogs using `nfd2` also freezes the app. 

This PR solves both but not in a very clean way.

francesca64:
@VZout while I wrote the macOS EL2 backend, I never included a `run_return` implementation, since I felt it wouldn't be idiomatic on macOS and that there'd likely be complications.

My perspective might be poisoned from having to write so many workarounds in my life, but I honestly don't think this looks _that_ unclean. While a canonical solution could exist, it'd likely involve a considerable amount of research into what other windowing implementations are doing.

Jasper-Bekkers:
> @VZout while I wrote the macOS EL2 backend, I never included a run_return implementation, since I felt it wouldn't be idiomatic on macOS and that there'd likely be complications.

For various reasons we can't have winit really take control of our application loop like that. It would be a welcome addition to have `run_return` available on all platforms that would support it.

I also don't think this code is super unclean, so I'd love to see this get merged into mainline.

francesca64:
@Jasper-Bekkers 
> It would be a welcome addition to have run_return available on all platforms that would support it.

I believe that's already the current story.

@vbogaevsky are you around for an additional review?

VZout:
I've fixed a minor issue. A colleague noticed that maocs dialogs were not being closes properly and always exited with the canceled event.

Issue #774, opened 2019-01-26
Could Wayland support be optional?


neonmoe:
Could the Wayland part of winit be made into its own feature? I'm asking this because it bothers me that the Wayland parts make up a considerable part of the dependencies of this crate, yet it's so wip that I currently manually force x11 in all my projects which use winit. I'm ready to contribute myself as well, I'd just like to hear if this is in any way feasible.

vberger:
I personally think it would make sense to have both `x11` and `wayland` be cargo features (both enabled by default). Apps that are only sensible in a wayland-only or x11-only context (because they use platform specific logic) but still want to use winit for the convenience it provides would be users of such flags.

Though, as I understand your initial concern is not that. What do you mean that "wayland support is so wip" exactly? A lot of progress has been done in the recent releases.

neonmoe:
It seems the last time I tested, I was on an old version of winit! Very sorry about that, it seems that the problems I was having (resizing and the window decorations being weird) have indeed been fixed.

That said, I think that splitting x11 and wayland into features would indeed be a good idea, for the reasons provided, and because I find joy in minimizing dependencies, being able to have different builds for x11 and wayland instead of bundling them up in one sounds very good.

vberger:
This would be pretty easy to implement overall, it's mainly a matter of scattering a few `#[cfg(...)]` in the right places. I suggest adding the two cargo features `platform_x11` and `platform_wayland`, which would be in the default feature set, and would only have an effect when compiling for unix platforms.

neonmoe:
I had a go at this today, and I have a working version of the `platform_x11` feature now working (ie. it can be disabled and nothing breaks, except for a few warnings I didn't clean up yet).

However, I would like some guidance on a few points:
1. It seems that target-specific features are somewhat WIP (https://github.com/rust-lang/cargo/issues/1197), so afaik, it isn't possible to make a specific target's dependencies optional, and then enable them as default features for that target. Of course, it's still possible to just make the `platform_x11` and `platform_wayland` features, and just leave the dependencies in. Removing the then-useless dependencies based on features should probably have it's own issue after this one is resolved, which would have to wait until Cargo is capable of this.
2. Glutin is very coupled to winit in this area, so it needs similar features. Those features should always be enabled in sync with winit (if winit has `platform_x11` enabled and glutin doesn't, glutin doesn't compile), and I'm not sure how that should be done. Just documentation (if so, where?), or is there a way to tell this to Cargo?
[...]

PR #1597, opened 2020-06-06
Make wayland or X11 optional

ogoffart:
As suggested in https://github.com/rust-windowing/winit/issues/774
Wayland adds lots of dependencies when building 

I only tested on Linux/X11.  I also checked it compiles with wayland alone

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 # Unreleased
+- On Unix, X11 and Wayland are now optional features (enabled by default)
 - On X11, fix deadlock when calling `set_fullscreen_inner`.
 - On Web, prevent the webpage from scrolling when the user is focused on a winit canvas
 
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,21 @@
+#[cfg(all(
+    any(
+        target_os = "linux",
+        target_os = "dragonfly",
+        target_os = "freebsd",
+        target_os = "netbsd",
+        target_os = "openbsd"
+    ),
+    not(feature = "x11"),
+    not(feature = "wayland")
+))]
+compile_error!("Please select a feature to build for unix: `x11`, `wayland`");
+
+#[cfg(all(
+    target_arch = "wasm32",
+    not(feature = "web-sys"),
+    not(feature = "stdweb")
+))]
+compile_error!("Please select a feature to build for web: `web-sys`, `stdweb`");
+
+fn main() {}
--- a/build.rs
+++ b/build.rs
@@ -9,13 +9,13 @@
     not(feature = "x11"),
     not(feature = "wayland")
 ))]
-compile_error!("Please select a feature to build for unix: `x11`, `wayland`");
+compile_error!("at least one of the \"x11\"/\"wayland\" features must be enabled");
 
 #[cfg(all(
     target_arch = "wasm32",
     not(feature = "web-sys"),
     not(feature = "stdweb")
 ))]
-compile_error!("Please select a feature to build for web: `web-sys`, `stdweb`");
+compile_error!("at least one of the \"web-sys\"/\"stdweb\" features must be enabled");
 
 fn main() {}
--- a/src/platform/unix.rs
+++ b/src/platform/unix.rs
@@ -1,7 +1,24 @@
-#![cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "netbsd", target_os = "openbsd"))]
+#![cfg(any(
+    target_os = "linux",
+    target_os = "dragonfly",
+    target_os = "freebsd",
+    target_os = "netbsd",
+    target_os = "openbsd"
+))]
+#![cfg_attr(
+    not(all(feature = "x11", feature = "wayland")),
+    allow(
+        unused_imports,
+        unused_variables,
+        unused_mut,
+        unreachable_patterns,
+        irrefutable_let_patterns
+    )
+)]
 
 use std::{os::raw, ptr, sync::Arc};
 
+#[cfg(feature = "wayland")]
 use smithay_client_toolkit::window::{ButtonState as SCTKButtonState, Theme as SCTKTheme};
 
 use crate::{
@@ -11,16 +28,19 @@ use crate::{
     window::{Window, WindowBuilder},
 };
 
+#[cfg(feature = "x11")]
+use crate::platform_impl::x11::{ffi::XVisualInfo, XConnection};
 use crate::platform_impl::{
-    x11::{ffi::XVisualInfo, XConnection},
     EventLoop as LinuxEventLoop, EventLoopWindowTarget as LinuxEventLoopWindowTarget,
     Window as LinuxWindow,
 };
 
 // TODO: stupid hack so that glutin can do its work
+#[cfg(feature = "x11")]
 #[doc(hidden)]
 pub use crate::platform_impl::x11;
 
+#[cfg(feature = "x11")]
 pub use crate::platform_impl::{x11::util::WindowType as XWindowType, XNotSupported};
 
 /// Additional methods on `EventLoopWindowTarget` that are specific to Unix.
@@ -32,6 +52,7 @@ pub trait EventLoopWindowTargetExtUnix {
     fn is_x11(&self) -> bool;
 
     #[doc(hidden)]
+    #[cfg(feature = "x11")]
     fn xlib_xconnection(&self) -> Option<Arc<XConnection>>;
 
     /// Returns a pointer to the `wl_display` object of wayland that is used by this
@@ -56,6 +77,7 @@ impl<T> EventLoopWindowTargetExtUnix for EventLoopWindowTarget<T> {
 
     #[inline]
     #[doc(hidden)]
+    #[cfg(feature = "x11")]
     fn xlib_xconnection(&self) -> Option<Arc<XConnection>> {
         match self.p {
             LinuxEventLoopWindowTarget::X(ref e) => Some(e.x_connection().clone()),
@@ -66,6 +88,7 @@ impl<T> EventLoopWindowTargetExtUnix for EventLoopWindowTarget<T> {
     #[inline]
     fn wayland_display(&self) -> Option<*mut raw::c_void> {
         match self.p {
+            #[cfg(feature = "wayland")]
             LinuxEventLoopWindowTarget::Wayland(ref p) => {
                 Some(p.display().get_display_ptr() as *mut _)
             }
@@ -76,6 +99,7 @@ impl<T> EventLoopWindowTargetExtUnix for EventLoopWindowTarget<T> {
 
 /// Additional methods on `EventLoop` that are specific to Unix.
 pub trait EventLoopExtUnix {
+    #[cfg(feature = "x11")]
     /// Builds a new `EventLoop` that is forced to use X11.
     ///
     /// # Panics
@@ -86,6 +110,7 @@ pub trait EventLoopExtUnix {
     where
         Self: Sized;
 
+    #[cfg(feature = "wayland")]
     /// Builds a new `EventLoop` that is forced to use Wayland.
     ///
     /// # Panics
@@ -104,6 +129,7 @@ pub trait EventLoopExtUnix {
     where
         Self: Sized;
 
+    #[cfg(feature = "x11")]
     /// Builds a new X11 `EventLoop` on any thread.
     ///
     /// This method bypasses the cross-platform compatibility requirement
@@ -112,6 +138,7 @@ pub trait EventLoopExtUnix {
     where
         Self: Sized;
 
+    #[cfg(feature = "wayland")]
     /// Builds a new Wayland `EventLoop` on any thread.
     ///
     /// This method bypasses the cross-platform compatibility requirement
@@ -134,11 +161,13 @@ impl<T> EventLoopExtUnix for EventLoop<T> {
         wrap_ev(LinuxEventLoop::new_any_thread())
     }
 
+    #[cfg(feature = "x11")]
     #[inline]
     fn new_x11_any_thread() -> Result<Self, XNotSupported> {
         LinuxEventLoop::new_x11_any_thread().map(wrap_ev)
     }
 
+    #[cfg(feature = "wayland")]
     #[inline]
     fn new_wayland_any_thread() -> Self {
         wrap_ev(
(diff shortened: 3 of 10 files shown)


src/platform/unix.rs
```diff
@@ -76,6 +99,7 @@ impl<T> EventLoopWindowTargetExtUnix for EventLoopWindowTarget<T> {
 
 /// Additional methods on `EventLoop` that are specific to Unix.
 pub trait EventLoopExtUnix {
+    #[cfg(feature = "x11")]
```
murarth, line 102: This may be a matter of personal preference, but I feel that `#[cfg]` attributes such as this one should go below the doc comment to improve readability.

src/platform/unix.rs
```diff
@@ -66,6 +88,7 @@ impl<T> EventLoopWindowTargetExtUnix for EventLoopWindowTarget<T> {
     #[inline]
     fn wayland_display(&self) -> Option<*mut raw::c_void> {
```
murarth, line 89: For the sake of consistency, I think backend-specific methods should always be marked with `#[cfg(...)]`; not just when the method signature contains a backend-specific type.

kchibisov:
I general, warnings must be fixed and not silenced with `allow` unless it's impossible to fix them, which is quite rare tbqh.

src/platform_impl/linux/mod.rs
```diff
@@ -48,161 +71,192 @@ pub struct PlatformSpecificWindowBuilderAttributes {
...
+    #[cfg(feature = "x11")]
     X(x11::DeviceId),
+    #[cfg(feature = "wayland")]
     Wayland(wayland::DeviceId),
+    #[cfg(all(not(feature = "x11"), not(feature = "wayland")))]
+    Dummy,
 }
 
 impl DeviceId {
     pub unsafe fn dummy() -> Self {
```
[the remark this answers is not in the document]
kchibisov: If both x11 and wayland backends are disabled on unix, the compilation should be aborted I guess on build.rs stage, what do you think? That's the thing we're currently doing in alacritty, since we do expose those features as well.

There could be a potential in a future one could want `Dummy`, but the goal is to always get the best code now, so it seems like aborting when none are presented is a way to go.

src/platform/unix.rs
```diff
@@ -1,7 +1,24 @@
...
+))]
+#![cfg_attr(
+    not(all(feature = "x11", feature = "wayland")),
+    allow(
+        unused_imports,
+        unused_variables,
+        unused_mut,
+        unreachable_patterns,
+        irrefutable_let_patterns
+    )
```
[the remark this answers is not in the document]
ogoffart: The problem is that trying to fix these warning requires lots of #[cfg] gymnastic, in a lot of locations. My personal opinion is that the code looks better without it.
Please note that these warning are not unconditionally disabled. They are only disabled if one back-end is disabled, which is not the default.
If you really insist i can try to eliminate warnings.

src/platform_impl/linux/mod.rs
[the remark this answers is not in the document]
ogoffart: I disabled the compilation with a compile_error and removed the Dummy.

src/platform/unix.rs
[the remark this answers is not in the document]
kchibisov: Not silencing them makes you get good warnings when you have only one feature enabled, which actually helps, so adding more cfg seems a way to go tbqh.

src/platform_impl/linux/mod.rs
```diff
@@ -1,12 +1,21 @@
 #![cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "netbsd", target_os = "openbsd"))]
 
-use std::{collections::VecDeque, env, ffi::CStr, fmt, mem::MaybeUninit, os::raw::*, sync::Arc};
+#[cfg(all(not(feature = "x11"), not(feature = "wayland")))]
+compile_error!("Please select a feature to build for web: `w11`, `wayland`");
```
murarth, line 4: `to build for web`? I don't understand this. Is this a mistake?

Also, you've written `w11` here instead of `x11`.

src/platform_impl/linux/mod.rs
```diff
@@ -34,12 +55,14 @@ const BACKEND_PREFERENCE_ENV_VAR: &str = "WINIT_UNIX_BACKEND";
 
 #[derive(Clone)]
 pub struct PlatformSpecificWindowBuilderAttributes {
```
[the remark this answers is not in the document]
murarth: The `#[cfg(...)]` attributes added to the methods appear correct, but you've forgotten to add `#[cfg]` to the corresponding fields.

src/platform_impl/linux/mod.rs
```diff
@@ -34,175 +45,199 @@ const BACKEND_PREFERENCE_ENV_VAR: &str = "WINIT_UNIX_BACKEND";
...
     pub static ref X11_BACKEND: Mutex<Result<Arc<XConnection>, XNotSupported>> =
         Mutex::new(XConnection::new(Some(x_error_callback)).map(Arc::new));
 }
 
 #[derive(Debug, Clone)]
+#[allow(dead_code)]
 pub enum OsError {
+    #[cfg(feature = "x11")]
     XError(XError),
     XMisc(&'static str),
```
murarth, line 92: `#[cfg]` should be added to the `XMisc` variant, as well. I think that should resolve the `dead_code` warning here.

src/platform_impl/linux/mod.rs
ogoffart: bad copy-paste

ogoffart:
Ping?

Maybe there should also be some CI configuration testing the build with either x11 or wayland disabled.

kchibisov:
I'm in favor of adding CI jobs for those targets, so go ahead.

murarth:
I'm not sure if GitHub would accept a non-member changing the CI files (for security reasons), so let's leave that out of this PR and I'll make a separate PR for new CI jobs after this is merged.

src/platform_impl/linux/mod.rs
```diff
@@ -34,175 +51,213 @@ const BACKEND_PREFERENCE_ENV_VAR: &str = "WINIT_UNIX_BACKEND";
...
     XError(XError),
+    #[cfg(feature = "x11")]
     XMisc(&'static str),
 }
 
 impl fmt::Display for OsError {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+    fn fmt(&self, _f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+        #[cfg(not(feature = "x11"))]
+        panic!();
```
murarth, line 116: It should be possible to remove this `panic!()` and simply mark each `match` arm below with `#[cfg(feature = "x11")]`. It may be counter-intuitive, but the construct `match foo {}` is valid for an `enum` value with no defined variants.

ogoffart: I thought the same, but it did not work.
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=482ef496084d4ac4ae03268489e6108f

ogoffart: Actually, i managed to make it work by dereferencing the value

murarth:
@kchibisov Does this look good to you?

src/platform/unix.rs
```diff
@@ -1,37 +1,49 @@
 #![cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "netbsd", target_os = "openbsd"))]
 
-use std::{os::raw, ptr, sync::Arc};
+use std::os::raw;
```
kchibisov, line 4: remove blank line.

src/platform/unix.rs
```diff
@@ -1,37 +1,49 @@
...
 use crate::{
-    dpi::Size,
     event_loop::{EventLoop, EventLoopWindowTarget},
     monitor::MonitorHandle,
     window::{Window, WindowBuilder},
 };
 
+#[cfg(feature = "x11")]
+use crate::dpi::Size;
+
```
kchibisov, line 19: remove blank like here as well.

src/platform/unix.rs
```diff
@@ -1,37 +1,49 @@
...
 
+#[cfg(feature = "x11")]
 pub use crate::platform_impl::{x11::util::WindowType as XWindowType, XNotSupported};
 
 /// Additional methods on `EventLoopWindowTarget` that are specific to Unix.
 pub trait EventLoopWindowTargetExtUnix {
     /// True if the `EventLoopWindowTarget` uses Wayland.
+    #[cfg(feature = "wayland")]
     fn is_wayland(&self) -> bool;
     ///
```
kchibisov, line 40: While we're at it, could you remove this `///` thing and just have a blank line in between.

src/platform/unix.rs
```diff
@@ -221,73 +257,91 @@ pub trait WindowExtUnix {
...
             LinuxWindow::X(ref w) => Some(w.xlib_xconnection()),
+            #[cfg(feature = "wayland")]
             _ => None,
         }
     }
 
     #[inline]
+    #[cfg(feature = "x11")]
     fn set_urgent(&self, is_urgent: bool) {
+        #[allow(irrefutable_let_patterns)]
```
kchibisov, line 303: This bumps minimum rust version to `1.43.0`, please, remove it for now, otherwise it's nearly impossible to ship winit apps, like alacritty. You are free to add a comment that it should be added in a future back.

src/platform_impl/linux/mod.rs
```diff
@@ -1,12 +1,27 @@
...
+#![cfg(any(
+    target_os = "linux",
+    target_os = "dragonfly",
+    target_os = "freebsd",
+    target_os = "netbsd",
+    target_os = "openbsd"
+))]
+
+#[cfg(all(not(feature = "x11"), not(feature = "wayland")))]
+compile_error!("Please select a feature to build for unix: `x11`, `wayland`");
```
kchibisov, line 10: Please, do that in build.rs where it should belong, otherwise it's hard to notice such errors, since a lot of things errors out in the same time + it's more clear, when it's in build.rs

src/platform/unix.rs
ogoffart: I assume you meant 1.33. But i'll change it.

src/platform_impl/linux/mod.rs
ogoffart: I added a build.rs with that check, and also the check for the web features

src/platform/unix.rs
kchibisov: No, I mean exactly 1.43, since complication is failing on 1.43.

src/platform/unix.rs
```diff
@@ -1,13 +1,11 @@
 #![cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "netbsd", target_os = "openbsd"))]
 
+#[cfg(feature = "wayland")]
+use smithay_client_toolkit::window::{ButtonState as SCTKButtonState, Theme as SCTKTheme};
 use std::os::raw;
```
kchibisov, line 5: The import of `smithay_client_toolkit` was fine. The idea was to group `std` imports, so move that one back.

build.rs
```diff
@@ -0,0 +1,21 @@
...
+    not(feature = "wayland")
+))]
+compile_error!("Please select a feature to build for unix: `x11`, `wayland`");
+
+#[cfg(all(
+    target_arch = "wasm32",
+    not(feature = "web-sys"),
+    not(feature = "stdweb")
+))]
+compile_error!("Please select a feature to build for web: `web-sys`, `stdweb`");
```
kchibisov, line 19:
```suggestion
compile_error!("at least one of the \"web-sys\"/\"stdweb\" features must be enabled");
```

build.rs
```diff
@@ -0,0 +1,21 @@
...
+        target_os = "linux",
+        target_os = "dragonfly",
+        target_os = "freebsd",
+        target_os = "netbsd",
+        target_os = "openbsd"
+    ),
+    not(feature = "x11"),
+    not(feature = "wayland")
+))]
+compile_error!("Please select a feature to build for unix: `x11`, `wayland`");
```
kchibisov, line 12:
```suggestion
compile_error!("at least one of the \"x11\"/\"wayland\" features must be enabled");
```

src/platform/unix.rs
```diff
@@ -394,6 +466,7 @@ impl MonitorHandleExtUnix for MonitorHandle {
     }
 }
 
+#[cfg(feature = "wayland")]
 /// Wrapper for implementing SCTK's theme trait.
```
kchibisov, line 470: comment should be above `cfg`.

src/platform/unix.rs
```diff
@@ -148,11 +170,13 @@ impl<T> EventLoopExtUnix for EventLoop<T> {
...
     }
 
+    #[cfg(feature = "x11")]
     #[inline]
     fn new_x11() -> Result<Self, XNotSupported> {
         LinuxEventLoop::new_x11().map(wrap_ev)
     }
 
+    #[cfg(feature = "wayland")]
     #[inline]
```
kchibisov, line 180: swap `#inline` and `#cfg` to follow the rest of the code in other files, apply to all similar things in that file.

src/platform/unix.rs
```diff
@@ -1,37 +1,46 @@
...
+use crate::platform_impl::x11::{ffi::XVisualInfo, XConnection};
 use crate::platform_impl::{
-    x11::{ffi::XVisualInfo, XConnection},
     EventLoop as LinuxEventLoop, EventLoopWindowTarget as LinuxEventLoopWindowTarget,
     Window as LinuxWindow,
 };
 
 // TODO: stupid hack so that glutin can do its work
+#[cfg(feature = "x11")]
 #[doc(hidden)]
```
kchibisov, line 27: swap `doc` and `cfg`.

CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 # Unreleased
+- On Unix, X11 and Wayland are now optional features (enabled by default)
```
kchibisov, line 2: in general `macOS` is unix(and macOS even has Xorg), so maybe `On Unix non macOS`. You should likely update it as well on that PR.

murarth: I think it's fine the way it is. It may be technically incorrect, but "Unix" is used elsewhere in winit docs to mean "Unix except for MacOS." I think users reading the changelog will understand that X11 and Wayland are not part of the Mac winit platform.

kchibisov: I kind of agree here tbqh, since winit already using not entirely accurate words, so let's just keep it the way it is.

PR #1604, opened 2020-06-12
Android run return

### LINKED ISSUE #1603: [Android] Run Return for Android
Currently run return is only available for Desktop operating systems. But at Embark Studios we require `run_return` for android as well.

VZout:

Fixes #1603

### DIFF
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -13,6 +13,7 @@ use ndk::{
 };
 use ndk_glue::{Event, Rect};
 use std::{
+    cell::Cell,
     collections::VecDeque,
     sync::{Arc, Mutex, RwLock},
     time::{Duration, Instant},
@@ -44,6 +45,23 @@ fn poll(poll: Poll) -> Option<EventSource> {
 pub struct EventLoop<T: 'static> {
     window_target: event_loop::EventLoopWindowTarget<T>,
     user_queue: Arc<Mutex<VecDeque<T>>>,
+    cf: Cell<ControlFlow>,
+    first_event: Option<EventSource>,
+    start_cause: event::StartCause,
+    looper: ThreadLooper,
+    running: bool,
+}
+
+macro_rules! call_event_handler {
+    ( $event_handler:expr, $window_target:expr, $cf:expr, $event:expr ) => {
+        {
+            if $cf != ControlFlow::Exit {
+                $event_handler($event, $window_target, &mut $cf);
+            } else {
+                $event_handler($event, $window_target, &mut ControlFlow::Exit);
+            }
+        }
+    }
 }
 
 impl<T: 'static> EventLoop<T> {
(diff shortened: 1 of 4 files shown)


src/platform_impl/windows/event_loop/runner.rs
```diff
@@ -250,7 +250,8 @@ impl<T> EventLoopRunner<T> {
         loop {
             // We do this instead of using a `while let` loop because if we use a `while let`
             // loop the reference returned `borrow_mut()` doesn't get dropped until the end
-            // of the loop's body and attempts to add events to the event buffer while in
```
filnet, line 253: seems unrelated

VZout: Good catch!

VZout:
Opened the PR. CI appears to be failing due to a unrelated issue.

francesca64:
Thanks for contributing again!

@dvc94ch are you fine with being pinged to review things like this? You're not listed on the [table](https://github.com/rust-windowing/winit/wiki/Testers-and-Contributors), but you seem to be the current expert. Let me know if I shouldn't pester you!

src/platform_impl/android/mod.rs
```diff
@@ -44,6 +44,20 @@ fn poll(poll: Poll) -> Option<EventSource> {
...
+}
+
+macro_rules! call_event_handler {
+    ( $event_handler:expr, $window_target:expr, $cf:expr, $event:expr ) => {{
+        if $cf != ControlFlow::Exit {
+            $event_handler($event, $window_target, &mut $cf);
+        } else {
+            $event_handler($event, $window_target, &mut ControlFlow::Exit);
+        }
+    }};
```
dvc94ch, line 60: Can you explain what this macro is for? The if else block seems redundant

dvc94ch:
@francesca64 it's fine, I'm not watching the repo, feel free to cc me on Android stuff.

src/platform_impl/android/mod.rs
VZout: Its pretty much copied from the windows implementation.

if the user sets the control flow at `MainEventsCleared` but later changes the control flow to `wait` in the `RedrawRequested` event it won't exit anymore. This only allows you to change the control flow if the control flow is not exit.

This happens for example every single event with the `window_run_return` example which always sets the `*control_flow = ControlFlow::Wait;` back to wait even if the control flow is exit already. See here: 
https://github.com/rust-windowing/winit/blob/master/examples/window_run_return.rs#L31

dvc94ch: Oh I see, you're passing a mutable reference, my bad

src/platform_impl/android/mod.rs
```diff
@@ -160,50 +203,66 @@ impl<T: 'static> EventLoop<T> {
...
                             self.window_target(),
-                            &mut cf,
+                            control_flow,
+                            event::Event::UserEvent(event)
                         );
                     }
                 }
-                None => {}
+                None => {
+                    control_flow = ControlFlow::Exit;
```
msiglreith, line 215: This causes my current application to restart in a loop.
Not sure what's the intention here as `first_event` is initialized as `None` on `EventLoop` creation which causes this match to always end up in the branch imo

VZout: Are you using `run_return` directly or `return`?

While the control flow is set to exit there the first time it is not necessarily permanent. There is a chance for the user to change it later to something else when `MainEventsCleared` or `RedrawRequested` is emitted. Than the first event is set to something other than none. Also the next time we enter the loop `first_event` won't (unless the poll returns None) be None anymore because it is assigned to something else down below.

VZout: Ah I see why this would be problematic when using `run`.

VZout: This works with the winit `window` example which uses `run` because the control flow is always set to ` *control_flow = ControlFlow::Wait;` by the user.

msiglreith: > Are you using run_return directly or return

`run`

> While the control flow is set to exit there the first time it is not necessarily permanent. There is a chance for the user to change it later to something else when MainEventsCleared or RedrawRequested is emitted.

That's not the case as far as I can see. In particular, that's caused by the way `call_event_handler` macro is implemented (https://github.com/VZout/winit/blob/6ce0fa2f09f94ecfad6ac3c37732c6c1f39f2401/src/platform_impl/android/mod.rs#L58). It creates a local variable which will throw away the value we set the control flow in our event handler.

> This works with the winit window example which uses run because the control flow is always set to *control_flow = ControlFlow::Wait; by the user.

Maybe, my phone works different, but I also overwrite it to `ControlFlow::Wait` at the beginning.

VZout: Hm your right. The call_event_handler would prevent that. I remember testing this with `run` and it working but maybe I messed something up.

PR #292, opened 2017-09-21
Add mouse event capturing when click-dragging out of a win32 window

Osspial:
This PR updates the win32 backend to capture mouse input when a user clicks and hold a mouse button in a window, allowing movement and other mouse events to be sent to a window without the cursor necessarily hovering over said window.

This feature is quite useful in GUI applications, where users may click-drag something and end up with the mouse outside the window. An example can be seen in Firefox or Chrome, by dragging the scroll bar and moving the mouse outside the window - the bar can still be moved, even though the mouse is outside the window and it wouldn't be receiving events had it not explicitly captured the mouse.

PR #1607, opened 2020-06-15
macOS: Support click-dragging out of a window

### LINKED ISSUE #1599: [macos] Click-dragging out of a window doesn't work on macOS
How to reproduce:
- run "control_flow" example on macOS
- press LMB inside the window area
- move the cursor outside of a window.

Expected: CursorMoved events keep coming
Real: CursorMoved disappear when the cursor leaves a window area

It works fine on Windows and Linux (checked on X11), but not on macOS.

ozkriff:
This is my naive attempt to fix #1599. It seems to work, though I've only checked on a remote macOS machine.

Fixes #1599 (_"[macos] Click-dragging out of a window doesn't work on macOS"_) 
Related to #292 (_"Add mouse event capturing when click-dragging out of a win32 window"_)

### DIFF
--- a/src/platform_impl/macos/view.rs
+++ b/src/platform_impl/macos/view.rs
@@ -910,13 +910,12 @@ fn mouse_motion(this: &Object, event: id) {
         let view_point = view.convertPoint_fromView_(window_point, nil);
         let view_rect = NSView::frame(view);
 
-        let mouse_buttons_down: NSInteger = msg_send![class!(NSEvent), pressedMouseButtons];
-
         if view_point.x.is_sign_negative()
             || view_point.y.is_sign_negative()
             || view_point.x > view_rect.size.width
             || view_point.y > view_rect.size.height
         {
+            let mouse_buttons_down: NSInteger = msg_send![class!(NSEvent), pressedMouseButtons];
             if mouse_buttons_down == 0 {
                 // Point is outside of the client area (view) and no buttons are pressed
                 return;
--- a/src/platform_impl/macos/view.rs
+++ b/src/platform_impl/macos/view.rs
@@ -58,7 +58,6 @@ pub(super) struct ViewState {
     is_key_down: bool,
     pub(super) modifiers: ModifiersState,
     tracking_rect: Option<NSInteger>,
-    mouse_buttons_down: u32,
 }
 
 impl ViewState {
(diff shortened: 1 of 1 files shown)


vitvakatu:
I can confirm the fix is working on native macOS Catalina

ozkriff:
Not sure why the "CI / Tests (stable, x86_64-apple-ios, macos-latest) (pull_request)" job was auto-canceled. I don't have the access to restart it.

src/platform_impl/macos/view.rs
```diff
@@ -77,6 +78,7 @@ pub fn new_view(ns_window: id) -> (IdRef, Weak<Mutex<CursorState>>) {
         is_key_down: false,
         modifiers: Default::default(),
         tracking_rect: None,
+        mouse_buttons_down: 0,
```
chrisduerr, line 81: Mouse events provide the `pressedMouseButtons` field. Have you considered using that instead of implementing tracking yourself?

ozkriff:
@chrisduerr yeah, pressedMouseButtons seems to work fine, thanks. Updated the PR.

src/platform_impl/macos/view.rs
```diff
@@ -910,13 +910,17 @@ fn mouse_motion(this: &Object, event: id) {
         let view_point = view.convertPoint_fromView_(window_point, nil);
         let view_rect = NSView::frame(view);
 
+        let mouse_buttons_down: NSInteger = msg_send![class!(NSEvent), pressedMouseButtons];
```
chrisduerr, line 913: Why put this outside of the conditional? Seems like a waste to do this when you don't need to?

chrisduerr: You can of course keep this outside and just add `|| mouse_button_down == 0`, which would reduce the level of nesting. Perhaps that was the original intention of putting it here?

ozkriff: Moved the variable inside the if body.

It's possible to merge both conditions, but the result is (subjectively) harder to read because of all the brackets.

chrisduerr: Since there's also no clean way to inline the msg_send! directly, meaning that this variation also has a slight performance advantage in the general case of sending events, I agree that it's the best choice.

PR #1616, opened 2020-07-03
Move available_monitors and primary_monitor to EventLoopWindowTarget

### LINKED ISSUE #1566: `available_monitors` on EventLoopWindowTarget
Issue: You can't list monitors unless you already have a window open and/or you're outside the event loop. I've a project where I want to spawn a window on each monitor periodically, then close them. The problem is that `available_monitors` is implemented only on EventLoop, not on the subset found inside the `run` closure.

I currently have a workaround where I spawn a window that I hide for the duration of the program and use it only to list monitors. However, I find this hacky at best and I think there could be a better solution.

Proposition: move `available_monitors` to EventLoopWindowTarget, so that it's available inside the `run` closure. I believe if it's possible on window inside the event loop, it should also be possible.

I don't have experience with the codebase, but if it is relatively beginner-friendly and with a bit of guidance I should be able to roll out a PR. Please let me know if that seems a good option.

If there is something that I overlooked, please let me know.

AdminXVII:

I only have Linux machines, so the other platforms where not tested. This should not change behaviour though.

Fix #1566

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- On all platforms, `available_monitors` and `primary_monitor` are now on `EventLoopWindowTarget` rather than `EventLoop` to list monitors event in the event loop.
 - On Unix, X11 and Wayland are now optional features (enabled by default)
 - On X11, fix deadlock when calling `set_fullscreen_inner`.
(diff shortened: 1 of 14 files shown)


Osspial:
Looks good to me. Thanks for pulling this together!

PR #1617, opened 2020-07-04
Add custom cursor icon support on Windows

Osspial:


kchibisov:
The PR mentions windows, however you're touching other backends and exposing a new API to some sort. So what's the buffer pixel on Windows(RGBA/ARGB)? On Wayland for example the buffer is in `ARGB`, so it could mean that for custom cursor loading I should change it a bit. Also, maybe we should check what cursor buffer format is on majority platforms and just use it? I'd be interested myself in `ARGB`, but I guess the most used should be picked?

Osspial:
I needed to touch the other platforms because this PR makes `CursorIcon` non-`Copy` and non-`Hash`, and adds a new variant that the other platforms weren't handling. Right now, I'm having the other platforms treat custom cursors as the default cursor, which seems like the least intrusive way of ignoring the `Custom` variant.

The public API surface area for this is fairly small - we're just adding a `Custom(Icon)` variant to `CursorIcon`, and adding a `new_with_hot_spot` method to `Icon`. `Icon`'s pixel buffer was already publicly formatted as RGBA, and this PR doesn't change that. Each individual backend is expected to convert that to the pixel format it needs, and on Windows, that's BGRA.

> I also can add wayland implementation for such thing, I guess, but it's a bit obscure. And has some edges like scale factor and what user should provide. I'd assume that all HiDPI handling is up to user here, and it should just ensure that buffer is scaled properly.

@kchibisov That'd be great, thanks! On Windows, the OS doesn't do any automatic DPI scaling for the custom cursor, so the user would have to manually upload an appropriately-scaled version of the cursor when the scale factor changes. I'm personally okay with that approach, but it should definitely be documented.

kchibisov:
note, I'll send Wayland impl later, since it'll actually require some rework of Wayland backend before doing so. Just saying so won't end up waiting on something.

Osspial:
I've thought some more about how icon scaling works, and I've changed my mind somewhat. I still don't think that Winit should do any implicit icon scaling beyond what the OS does, but it should provide an API for attaching multiple images at different scales for a single logical `Icon`. This is for a couple of reasons:

- Windows `ICO` files and icon resources include multiple icon sizes in a single file, and if an `Icon` struct only contains an icon at a single scale (as it does today), users have to manually reload those icons when the scale changes. That's super counterintuitive.
- The OS can use multiple window icon sizes at the same time. The current model doesn't support that usecase.
- Windows changes the cursor size when the cursor changes monitor, not when the underlying window changes the scale factor. Expecting users to monitor the exact position of the cursor in a multiple-monitor setup to determine the cursor size is unreasonable, especially since not all backends give access to that information.

I'll update this PR to expose that API and implement it on Windows.

kchibisov:
I'll just add that on Wayland client draws the cursor itself with the size and scale it wants. winit shouldn't care about size part on it, since sctk does that already for it, so only scale. Since scaling is done by clients, they should just set buffer scale and attach the proper buffer size (just physical size thing).

So multiple icons is softly required thing on Wayland (you can in theory draw with some big scale, and rely on compositor downscaling).

>Windows changes the cursor size when the cursor changes monitor, not when the underlying window changes the scale factor. Expecting users to monitor the exact position of the cursor in a multiple-monitor setup to determine the cursor size is unreasonable, especially since not all backends give access to that information.

Same for Wayland, but since cursor is managed by users, and it's just a normal surface, you get scale changes events for it separately, and so can react and pick proper cursor for that scale, so not a big issue. So if winit want's some consistency across the platforms the cursor DPI scaling should be entirely done by winit, and user should just provide some set of images to pick from.

>The OS can use multiple window icon sizes at the same time. The current model doesn't support that usecase.

Not sure how it's related to cursor related things?

Osspial:
> Not sure how it's related to cursor related things?

The current API uses the same type for custom window icons and custom cursor icons, so adding support in the public API for this helps improve both the cursor implementation and the window icon implementation.

src/icon.rs
```diff
@@ -12,108 +15,121 @@ pub(crate) struct Pixel {
...
+            PhysicalPosition::new(size.width / 2, size.height / 2),
+        )
+    }
+
+    /// Creates a `RgbaIcon` from 32bpp RGBA data, with a defined cursor hot spot. The hot spot is
+    /// the exact pixel in the icon image where the cursor clicking point is, and is ignored when
+    /// the icon is used as a window icon.
+    ///
+    /// ## Panics
+    /// Panics if the length of `rgba` must be divisible by 4, or if `width * height` doesn't equal
```
kchibisov, line 99:
```suggestion
    /// Panics if the length of `rgba` is not divisible by 4, or if `width * height` doesn't equal
```

src/icon.rs
```diff
@@ -12,108 +15,121 @@ pub(crate) struct Pixel {
...
-#[derive(Debug, Clone, PartialEq, Eq)]
-pub(crate) struct RgbaIcon {
-    pub(crate) rgba: Vec<u8>,
-    pub(crate) width: u32,
-    pub(crate) height: u32,
+#[derive(Clone, PartialEq, Eq, Hash)]
+pub struct RgbaIcon<I: Deref<Target = [u8]>> {
+    pub(crate) rgba: I,
+    pub(crate) size: PhysicalSize<u32>,
+    pub(crate) hot_spot: PhysicalPosition<u32>,
```
kchibisov, line 22: I don't really like that we have a cursor and normal icons under the same `RgbaIcon`, in general, we can name it like `RgbaBuffer`, since the intent to have `RgbaBuffer` and have `Icon` and a new type for cursor icons, which will use that `RgbaBuffer` under the hood?

It's just strange that we have a field that shouldn't be used in any ways in certain scenarios, but in other cases it's fine to use them. Such issues should be solved on type system level.

src/icon.rs
```diff
@@ -125,11 +141,61 @@ impl fmt::Debug for Icon {
 impl Icon {
     /// Creates an `Icon` from 32bpp RGBA data.
     ///
-    /// The length of `rgba` must be divisible by 4, and `width * height` must equal
-    /// `rgba.len() / 4`. Otherwise, this will return a `BadIcon` error.
-    pub fn from_rgba(rgba: Vec<u8>, width: u32, height: u32) -> Result<Self, BadIcon> {
+    /// ## Panics
+    /// Panics if the length of `rgba` must be divisible by 4, or if `width * height` doesn't equal
```
kchibisov, line 145:
```suggestion
    /// Panics if the length of `rgba` is not divisible by 4, or if `width * height` doesn't equal
```

src/icon.rs
```diff
@@ -125,11 +141,61 @@ impl fmt::Debug for Icon {
...
+            inner: PlatformIcon::from_rgba(rgba.into(), size)?,
+        })
+    }
+
+    /// Creates an `Icon` from 32bpp RGBA data, with a defined cursor hot spot. The hot spot is
+    /// the exact pixel in the icon image where the cursor clicking point is, and is ignored when
+    /// the icon is used as a window icon.
+    ///
+    /// ## Panics
+    /// Panics if the length of `rgba` must be divisible by 4, or if `width * height` doesn't equal
```
kchibisov, line 158:
```suggestion
    /// Panics if the length of `rgba` is not divisible by 4, or if `width * height` doesn't equal
```

src/icon.rs
```diff
@@ -125,11 +141,61 @@ impl fmt::Debug for Icon {
...
+    ///
+    /// `get_icon` will be lazily called for a particular icon size whenever the window manager
+    /// needs an icon of that size. The `PhysicalSize<u32>` parameter specifies the window manager's
+    /// suggested icon size for a particular scale factor, and will always be a square. The `f64`
+    /// parameter specifies the scale factor that the window manager is requesting the icon with.
+    /// `get_icon` will only be called once for any given suggested icon size.
+    ///
+    /// If `get_icon` returns `Err(e)` for a given size, Winit will invoke `warn!` on the returned
+    /// error and will try to retrieve a differently-sized icon from `get_icon`.
+    pub fn from_rgba_fn<F, B>(mut get_icon: F) -> Result<Self, std::io::Error>
```
kchibisov, line 180: I don't really like that the error is `std::io::Error` here, since there's no io involved, the only thing that could happen is that cursor isn't found? In that case, having a clear type or even using `Option` should be better? The option applies to the closure you're using as well, so if something is not present it'll be `None`, not some random io error.

src/platform_impl/linux/x11/util/cursor.rs
```diff
@@ -1,13 +1,20 @@
 use crate::window::CursorIcon;
+use std::mem;
 
 use super::*;
 
 impl XConnection {
     pub fn set_cursor_icon(&self, window: ffi::Window, cursor: Option<CursorIcon>) {
+        // ignore custom cursors for now
+        let cursor = match cursor {
+            Some(CursorIcon::Custom(_)) => Some(CursorIcon::default()),
```
kchibisov, line 10: What is the point of filtering out cursor here, if you can safely do that in `get_cursor`? You have already and arm for that in a match, just load `left_ptr` as you do on `web` and `Wayland` now?

Osspial: Good point. That's been corrected.

src/icon.rs
Osspial: `io::Error` is used as the standard OS error type in Rust, even in cases where the error isn't directly related to IO (see [`io::Error::last_os_error()`](https://doc.rust-lang.org/std/io/struct.Error.html#method.last_os_error)). Still, it's not relevant for this function since `get_icon` is supposed to be called lazily, so I've changed the function to return `Self` instead of a `Result`.

src/icon.rs
Osspial: That's a decent point. I've updated the API to split the custom window type and custom cursor type into two parts, though it's not documented yet. Could you take a look at it?

anlumo:
This change removes the `Send` trait from `winit::window::CursorIcon` for Windows, which is kinda a big deal. Is there a way to get it back?

```
    = help: within `winit::platform_impl::platform::icon::RaiiIcon`, the trait `std::marker::Send` is not implemented for `*mut winapi::shared::windef::HICON__`
    = note: required because it appears within the type `winit::platform_impl::platform::icon::RaiiIcon`
    = note: required because of the requirements on the impl of `std::marker::Sync` for `std::sync::Arc<winit::platform_impl::platform::icon::RaiiIcon>`
    = note: required because it appears within the type `winit::platform_impl::platform::icon::CustomCursorIcon`
    = note: required because it appears within the type `winit::icon::CustomCursorIcon`
    = note: required because it appears within the type `winit::window::CursorIcon`
```

Osspial:
> This change removes the `Send` trait from `winit::window::CursorIcon` for Windows, which is kinda a big deal. Is there a way to get it back?

Yep! That was just an oversight - the implementation was written with `Send` and `Sync` in mind, but I forgot to mark the relevant types with those traits. That's fixed now.

daxpedda:
Closing due to inactivity.

Tracked in #3005.

PR #1621, opened 2020-07-09
Fix Windows transparency behavior to support fully-opaque regions

jimporter:

This PR is primarily here to fix a downstream bug (https://github.com/alacritty/alacritty/issues/1927). Testing the transparency feature out on Windows (both via `examples/transparency.rs` and with a build of alacritty), I was unable to reproduce the issue described in the old comment in this code. I suppose it's possible that this only manifests on some configurations though (for completeness, I'm testing on a Windows 10 environment, though I cross-compiled things on a Linux box using GNU tools). To be on the safe side, it might make sense for other people with access to a Windows environment to double check that this doesn't regress anything.

In addition, I removed a couple of redundant bits in the Windows API calls (see the description below).

---
*Original commit message follows:*

This patch removes an unneeded workaround for transparent windows on the Windows platform. In addition, it simplifies a couple of related API calls:

* Remove the `CreateRectRgn` call, since we want the entire window's region to have blur behind it, and `DwnEnableBlurBehindWindow` does that by default.
* Remove the `color_key` for `SetLayeredWindowAttributes`, since it's not used (we're not passing `winuser::LWA_COLORKEY` to the flags).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 - On Android, calling `WindowEvent::Focused` now works properly instead of always returning false. 
+- On Windows, fix bug preventing windows with transparency enabled from having fully-opaque regions.
 
 # 0.23.0 (2020-10-02)
 
--- a/src/platform_impl/windows/window.rs
+++ b/src/platform_impl/windows/window.rs
@@ -23,7 +23,6 @@ use winapi::{
         ole2,
         oleidl::LPDROPTARGET,
         shobjidl_core::{CLSID_TaskbarList, ITaskbarList2},
-        wingdi::{CreateRectRgn, DeleteObject},
         winnt::LPCWSTR,
         winuser,
     },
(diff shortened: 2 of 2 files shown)


jimporter:
@Osspial Does this look ok? (Note: I rebased this onto the latest master revision and Android tests are failing now, but looking at the CI logs, this is happening for other builds too and doesn't have anything to do with my patch, which is Windows-only anyway.)

jimporter:
@Osspial, @kchibisov: I rebased this again onto the latest master and the tests look good again. Just let me know if there's anything I need to change so that this can merge. Thanks!

msiglreith (commented):
Works fine for me, tested with an OpenGL application - 1.0 and non-0.0 showed the desired results.
Could you updated to latest master and adjust the changelog entry please?

jimporter:
@msiglreith Ok, I updated my branch and moved the relnote to the right section (and rephrased it a bit so it's less hard to understand). Thanks for taking a look!

msiglreith (approved):
Nice, thanks!

Issue #1497, opened 2020-03-06
Tracking issue for CompositionEvent support


murarth:
This issue tracks the implementation of IME composition events on each platform, an API initially proposed in #1293. PRs implementing this event should be made against the `composition-event` branch, which will be merged into `master` once all implementations are complete.

* [x] Public API implemented (#1404)
* [ ] Implemented on all platforms
  * [ ] Windows
  * [ ] Mac OS
  * [ ] iOS
  * [x] Linux X11 (#1404)
  * [ ] Linux Wayland
  * [ ] Web
* [ ] Document `CompositionEvent`
* [ ] Add an entry to `CHANGELOG.md`

chrisduerr:
As a consumer of winit I'd recommend against naming these events `CompositionEvent`. Applications that deal with display servers will already know that term in connection with window compositing and it could easily be confused with something related to that.

It also does not fall in line with the existing winit naming of functions like `set_ime_position`.

I think it would be much better to have a more self descriptive name for this event that cannot be confused with anything else related to the winit project. I'd strongly suggest that `IME` should at least be part of the name.

pickfire:
Oh, I thought `CompositionEvent` means some kind of compose key thing, I never used those and don't even know how to use compose key, I had only used `IME` to input characters not in the keyboard.

garasubo:
@pickfire This event naming comes from web specification (Ref: https://developer.mozilla.org/en-US/docs/Web/API/Element/compositionstart_event). And IME will create those composition events.
But as @chrisduerr is pointed out, I agree that calling those events as "composition" is confusing.

garasubo:
Proposed the rename these events to `IMEPreeditEvent` in #1622

kaiwk:
For mac, maybe we can use this API? [composedString](https://developer.apple.com/documentation/objectivec/nsobject/1385416-composedstring?language=objc)

kchibisov:
I've decided to take a look on the proposed events wrt compositing, since I've been working on IME handling in winit for Wayland recently.

Right now I can see that the current API that is being proposed looks more like

```rust
pub enum CompositionEvent {
    CompositionStart(String),
    CompositionUpdate(String, usize),
    CompositionEnd(String),
}
```

If I were modelling it from Wayland side it'll be more like.
```rust
pub enum CompositionEvent {
    /// IME got enabled on the surface you're on.
    CompositionEnabled
[...]

PR #1622, opened 2020-07-10
rename CompostionEvent to IMEPreeditEvent

garasubo:
Ref: https://github.com/rust-windowing/winit/issues/1497#issuecomment-609060877

To avoid confusion from users, rename `CompositionEvent` to `IMEPreeditEvent` to make it clear that those events are related to IME.
Also, added documents

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -271,7 +271,7 @@ pub enum WindowEvent<'a> {
     ModifiersChanged(ModifiersState),
 
     /// An event from IME
-    IME(IMEEvent),
+    IME(IME),
 
     /// The cursor has moved on the window.
     CursorMoved {
@@ -627,23 +627,25 @@ pub struct KeyboardInput {
     pub modifiers: ModifiersState,
 }
 
+/// Describes an event from input method.
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-pub enum IMEEvent {
-    /// The user enables IME
+pub enum IME {
+    /// Notifies when the IME was enabled.
     Enabled,
 
-    /// The user updates preedit status on IME.
+    /// Notifies when a new composing text should be set at the cursor position.
     ///
-    /// The value represents a pair of the preedit string and the cursor begin position and end position.
-    /// When the cursor is hidden, both value will be None
+    /// The value represents a pair of the preedit string and the cursor begin position and end
+    /// position. When both indices are `None`, the cursor should be hidden.
+    ///
     /// The cursor position is byte-wise indexed.
     Preedit(String, Option<usize>, Option<usize>),
 
-    /// The user completes the current IME session with the value.
+    /// Notifies when text should be inserted into the editor widget.
     Commit(String),
 
-    /// The user disables IME
+    /// Notifies when the IME was disabled.
     Disabled,
 }
 
--- a/src/event.rs
+++ b/src/event.rs
@@ -632,17 +632,17 @@ pub struct KeyboardInput {
 pub enum IMEEvent {
     /// The user enables IME
     Enabled,
-    /// Preedit session is going to begin with the value
-    PreeditStart(String),
 
     /// The user updates preedit status on IME.
     ///
-    /// The value represents a pair of the preedit string and the cursor position.
+    /// The value represents a pair of the preedit string and the cursor begin position and end position.
+    /// When the cursor is hidden, both value will be None
     /// The cursor position is byte-wise indexed.
     Preedit(String, Option<usize>, Option<usize>),
 
     /// The user completes the current IME session with the value.
     Commit(String),
+
     /// The user disables IME
     Disabled,
 }
--- a/src/event.rs
+++ b/src/event.rs
@@ -271,7 +271,7 @@ pub enum WindowEvent<'a> {
     ModifiersChanged(ModifiersState),
 
     /// An event from IME
-    Composition(CompositionEvent),
+    IME(IMEEvent),
 
     /// The cursor has moved on the window.
     CursorMoved {
@@ -379,7 +379,7 @@ impl Clone for WindowEvent<'static> {
                 input: *input,
                 is_synthetic: *is_synthetic,
             },
-            Composition(composition) => Composition(composition.clone()),
+            IME(preedit_state) => IME(preedit_state.clone()),
             ModifiersChanged(modifiers) => ModifiersChanged(modifiers.clone()),
             #[allow(deprecated)]
             CursorMoved {
@@ -471,7 +471,7 @@ impl<'a> WindowEvent<'a> {
                 is_synthetic,
             }),
             ModifiersChanged(modifiers) => Some(ModifiersChanged(modifiers)),
-            Composition(event) => Some(Composition(event)),
+            IME(event) => Some(IME(event)),
             #[allow(deprecated)]
             CursorMoved {
                 device_id,
@@ -629,10 +629,22 @@ pub struct KeyboardInput {
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-pub enum CompositionEvent {
-    CompositionStart(String),
-    CompositionUpdate(String, usize),
-    CompositionEnd(String),
+pub enum IMEEvent {
+    /// The user enables IME
+    Enabled,
+    /// Preedit session is going to begin with the value
+    PreeditStart(String),
+
+    /// The user updates preedit status on IME.
+    ///
+    /// The value represents a pair of the preedit string and the cursor position.
+    /// The cursor position is byte-wise indexed.
+    Preedit(String, Option<usize>, Option<usize>),
+
+    /// The user completes the current IME session with the value.
+    Commit(String),
+    /// The user disables IME
+    Disabled,
 }
 
 /// Describes touch-screen input state.
--- a/src/platform_impl/linux/x11/event_processor.rs
+++ b/src/platform_impl/linux/x11/event_processor.rs
@@ -16,8 +16,8 @@ use crate::platform_impl::platform::x11::ime::{ImeEvent, ImeEventReceiver};
 use crate::{
     dpi::{PhysicalPosition, PhysicalSize},
     event::{
-        CompositionEvent, DeviceEvent, ElementState, Event, KeyboardInput, ModifiersState,
-        TouchPhase, WindowEvent,
+        DeviceEvent, ElementState, Event, KeyboardInput, ModifiersState, IMEEvent, TouchPhase,
+        WindowEvent,
     },
     event_loop::EventLoopWindowTarget as RootELW,
 };
@@ -1238,23 +1238,25 @@ impl<T: 'static> EventProcessor<T> {
         }
         match self.ime_event_receiver.try_recv() {
             Ok((window, event)) => match event {
+                ImeEvent::Enabled => {
+                    callback(Event::WindowEvent {
+                        window_id: mkwid(window),
+                        event: WindowEvent::IME(IMEEvent::Enabled),
+                    });
+                }
                 ImeEvent::Start => {
                     self.is_composing = true;
                     self.composed_text = None;
                     callback(Event::WindowEvent {
                         window_id: mkwid(window),
-                        event: WindowEvent::Composition(CompositionEvent::CompositionStart(
-                            "".to_owned(),
-                        )),
+                        event: WindowEvent::IME(IMEEvent::Preedit("".to_owned(), None, None)),
                     });
                 }
                 ImeEvent::Update(text, position) => {
                     if self.is_composing {
                         callback(Event::WindowEvent {
                             window_id: mkwid(window),
-                            event: WindowEvent::Composition(CompositionEvent::CompositionUpdate(
-                                text, position,
-                            )),
+                            event: WindowEvent::IME(IMEEvent::Preedit(text, Some(position), None)),
                         });
                     }
                 }
@@ -1262,7 +1264,7 @@ impl<T: 'static> EventProcessor<T> {
                     self.is_composing = false;
                     callback(Event::WindowEvent {
                         window_id: mkwid(window),
-                        event: WindowEvent::Composition(CompositionEvent::CompositionEnd(
+                        event: WindowEvent::IME(IMEEvent::Commit(
                             self.composed_text.take().unwrap_or("".to_owned()),
                         )),
                     });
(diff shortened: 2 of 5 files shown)


garasubo:
@Osspial it makes sense for me. Will rename them.

garasubo:
@Osspial Could you check my change? I renamed them. One CI check was failed, but I think this is not related to this change.

src/event.rs
```diff
@@ -629,10 +629,18 @@ pub struct KeyboardInput {
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-pub enum CompositionEvent {
-    CompositionStart(String),
-    CompositionUpdate(String, usize),
-    CompositionEnd(String),
+pub enum PreeditState {
```
kchibisov, line 632: I think we can name it a bit differently, since it's not only about preedit state, since you have `commit`, etc. I'd also like to see adoption of 
```
pub enum CompositionEvent {
    CompositionEnabled,
    CompositionPreedit(String, Option<usize>, Option<usize>),
    CompositionCommit(String),
    CompositionDisabled,
}
```

That I've mentioned on #1497 if you don't have any objects, unfortunately I don't really know much about X11, so it's a bit hard for me to update its code myself.

Also we can name rename `PreeditState` to `IMEEvent`, and have

```rust
pub enum IMEEvent {
    Enabled,
    Preedit(String, Option<usize>, Option<usize>),
    Commit(String),
    Disabled,
}
```

So I guess Start and Update should be transformed into `Preedit`, and End should be transformed into Commit.

Enabled and Disabled should be implemented, and I guess you have that information from X server, at least I can see something like that in xim docs.

kchibisov (changes requested):
I've rebased composition-event branch.

I also think that we can apply the proposed changes(about event restructure) from the issue I've mentioned in that PR, right now we only have X11 backend, so that's why I think it should be done here.

If you have any questions or concerns let me know.

src/event.rs
garasubo: I think you miss Start (or maybe PreeditStart?) event. Other than that, I think it looks fine.

kchibisov: >I think you miss Start (or maybe PreeditStart?) event. Other than that, I think it looks fine.

Could you explain why `PreeditStart` is required? on Wayland for example everything starts with just `Preedit`.

garasubo: In Wep API, CompositionStart exist in the specification. And in x11, every IME session starts with preedit start event. So, I believe something to tell us when a new IME session starts is necessary.

kchibisov: By start you mean that input method got open or just new input? Since Wayland has a concept of 'Enabled', when user uses its IME, like they opened 'fcitx' or 'ibus' and focused your app. But there's no notion of 'Start', and everything is communicated via 'Preedit' and 'Commit'.

My question was about start, since it looked like `Preedit("", None, None)` to me.

kchibisov: My main concern is that I can't really send anything about on `PreeditStart` on Wayland, since I don't have a `Start` at all, I can synthesize a `PreeditStart("")`, but I don't feel like it'll make sense, so I was wondering  if we can send `Preedit("", None, None)` instead of start on those platforms.

src/event.rs
```diff
@@ -623,6 +627,26 @@ pub struct KeyboardInput {
     pub modifiers: ModifiersState,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum IMEEvent {
+    /// The user enables IME
+    Enabled,
```
kchibisov, line 634: Does X11 have an event to tell you when you actually have IME running or how to you obtain that information? Also does it tell you when disable happens, since if X11 doesn't have those events they should be issued at some points, since users may only handle IME things after you've actually got those, since IME is user initiated, I guess.

Or it's not the case for X11 and it just always works?

src/event.rs
garasubo: > By start you mean that input method got open or just new input?

I mean start a input method editor got open. For example, the user input "きょうは"(Enter)"いい"（Enter）"てんきですね"(Enter), then the start event will happen just before the user start input k of "きょうは", "i" of "いい", and "ｔ" of "てんきですね".
It is different concept from enable event. The enable event will happen when the user press the key to switch IME mode (e.g. "Zenkaku/Hankaku" button in JP keyboard).

> My main concern is that I can't really send anything about on PreeditStart on Wayland

In X11 and Web API allows to start with some texts. For example, I have a committed text "きょうは", and I want to convert it into Kanji after commit, then I can select this string and press key to convert, then the new session will start with "きょうは".
If you have Japanese IME, you  can see those behaviors in MDN doc page
https://developer.mozilla.org/en-US/docs/Web/API/Element/compositionstart_event

kchibisov: >In X11 and Web API allows to start with some texts. For example, I have a committed text "きょうは", and I want to convert it into Kanji after commit, then I can select this string and press key to convert, then the new session will start with "きょうは".

Yeah, I'll just get `Preedit` for that on Wayland, not start, and my primary concern why can't we merge Start and `Preedit`, since it's literally doesn't matter what reports initial text, right? Or you want different handling for `Start("Hello")` and `Preedit("Hello", None, None)`, I'd assume for both cases you shouldn't insert text into widget, right? Since it's a job of `Commit`.

src/event.rs
garasubo: Oh, sorry. This is WIP commit. I think we can get those status with XIMPreeditStateNotifyCallback, so will implement that and test it

src/event.rs
garasubo: > Or you want different handling for Start("Hello") and Preedit("Hello", None, None)

So, your suggestion is the application can detect the session start by checking the second argument is empty?
I feel this way is not intuitive.
I think start event in IME is common since this is defined not only in X11, also Web API and other famous framework, like gtk. So, I prefer to having a separate member to tell the session start.

kchibisov: >So, your suggestion is the application can detect the session start by checking the second argument is empty?

But why do you need to detect session start at all? The start is when you've got any IME event.

I mean, I don't understand how end users will handle `Start` differently to `Preedit`? Could you give an example where handling of those 2 are different and it makes sense? I get that it's common API for X11 and Web, but it really feels like someone decided to add redundant event that you can mirror with other events.

garasubo: Hmm, I don't have an idea of specific usage for Start event for now, but Start event looks relatively common, so I'm afraid that there is some needs.
But if there is no special use case, it might be OK to combine them.

garasubo:
I tried to implement Enable/Disable event, but it seems difficult in X11.
I assumed that we could implement them with `XNPreeditStateNotifyCallback`, but it doesn't work in my environment when I switched the IME.
I investigated some other projects, and found that in Vim, they don't implement since they couldn't find a good way in X11.
Ref: https://groups.google.com/g/vim_dev/c/9j76Z8hd23E?pli=1

@kchibisov do you think we should give up enable/disable event only on X11?

kchibisov:
>do you think we should give up enable/disable event only on X11?

Have you looked into something like that gtk on that? It seems strange that there's no indication that IME got opened, though, we can just always send enable when creating IME object and send `Disabled`on close or never send `Disabled` if it's too complicated to implement now. The point is to send `Enabled`, if it won't be optimal we won't lose anything, since right now downstream assumes that IME is always enabled.

The thing is that we must send it, but what we use to indicate that could be something not really optimal, if implementing it is something that complicated.

garasubo:
Yeah, GTK doesn't support enable events either (and won't support X11 at all in GTK4). In X11, we don't destory IME object basically, so the current solution would be never sending `Disable` event and send `Enable` event at the beginning.

kchibisov:
>so the current solution would be never sending Disable event and send Enable event at the beginning.

Looks fine to me, I guess.

kchibisov:
I've looked into things again, and I think we can go for now with `Disabled`/`Enabled` events tbqh, so just strip them form your impl and we should be good to go.

garasubo:
Sorry for late response. I think I can update the change in this weekend.

kchibisov:
Since you've kept enable and removed disable is that correct that there's no way to know that something got disabled? Also, would you mind rebasing the branch?

garasubo:
> Is that correct that there's no way to know that something got disabled?

Yes. Also, actually, it doesn't know IME enabled. It just sends enable event when it creates the context object, as we discussed before.

> Also, would you mind rebasing the branch?

Sure. Will do it later

garasubo:
Squash and rebased

src/event.rs
```diff
@@ -629,10 +629,22 @@ pub struct KeyboardInput {
...
+    PreeditStart(String),
+
+    /// The user updates preedit status on IME.
+    ///
+    /// The value represents a pair of the preedit string and the cursor position.
+    /// The cursor position is byte-wise indexed.
+    Preedit(String, Option<usize>, Option<usize>),
+
+    /// The user completes the current IME session with the value.
+    Commit(String),
```
kchibisov, line 645: New line here.

src/event.rs
```diff
@@ -629,10 +629,22 @@ pub struct KeyboardInput {
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-pub enum CompositionEvent {
-    CompositionStart(String),
-    CompositionUpdate(String, usize),
-    CompositionEnd(String),
+pub enum IMEEvent {
+    /// The user enables IME
+    Enabled,
```
kchibisov, line 634: New line here.

src/event.rs
```diff
@@ -629,10 +629,22 @@ pub struct KeyboardInput {
...
+    /// Preedit session is going to begin with the value
+    PreeditStart(String),
+
+    /// The user updates preedit status on IME.
+    ///
+    /// The value represents a pair of the preedit string and the cursor position.
+    /// The cursor position is byte-wise indexed.
```
kchibisov, line 641: We should document that the first argument is a cursor begin and second is cursor end. Also if your cursor being and end are equal you should send both of them the same, so in your code you should always send Some(position), Some(position)?

Also, both `None` should represent a hidden cursor, we should state it as well.

src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -1238,31 +1238,33 @@ impl<T: 'static> EventProcessor<T> {
...
                     });
                 }
                 ImeEvent::Update(text, position) => {
                     if self.is_composing {
                         callback(Event::WindowEvent {
                             window_id: mkwid(window),
-                            event: WindowEvent::Composition(CompositionEvent::CompositionUpdate(
-                                text, position,
-                            )),
+                            event: WindowEvent::IME(IMEEvent::Preedit(text, Some(position), None)),
```
kchibisov, line 1259: Also, is there a way to inform that cursor is hidden, thus shouldn't be drawn, if there's such thing we should inform about that via sending `None` and `None`? 

```suggestion
                            event: WindowEvent::IME(IMEEvent::Preedit(text, Some(position), Some(position))),
```

kchibisov (changes requested):
You should also fix formatting. After that it should be good to go, I think.

src/event.rs
garasubo: Added

src/event.rs
garasubo: Added

src/event.rs
garasubo: Added those description

src/platform_impl/linux/x11/event_processor.rs
garasubo: Fixed.

In X11, I cannot find any event related to hidden cursor.

src/platform_impl/linux/x11/ime/context.rs
```diff
@@ -89,7 +89,7 @@ extern "C" fn preedit_draw_callback(
             .collect()
     };
     let mut old_text_tail = client_data.text.split_off(chg_range.end);
-    client_data.text.split_off(chg_range.start);
+    let _ = client_data.text.split_off(chg_range.start);
```
kchibisov, line 92: This is ignore is intentional right?

kchibisov (changes requested):
@garasubo I've decided to rename the event one more time, and call it `IME`, since event suffix seems redundant here. I've also slightly updated the docs, are you fine with proposed changes?

kaiwk:
> I tried to implement Enable/Disable event, but it seems difficult in X11.
> I assumed that we could implement them with `XNPreeditStateNotifyCallback`, but it doesn't work in my environment when I switched the IME.
> I investigated some other projects, and found that in Vim, they don't implement since they couldn't find a good way in X11.
> Ref: https://groups.google.com/g/vim_dev/c/9j76Z8hd23E?pli=1
> 
> @kchibisov do you think we should give up enable/disable event only on X11?

From this [guide](https://developer.mozilla.org/en-US/docs/Mozilla/IME_handling_guide):

> If IME is available on focused elements, we call that state "enabled". If IME is not fully available(i.e., user cannot enable IME), we call this state "disabled".

Could we just simply think that when winit window gets focused, send `Enabled` and when window loses focus, send `Disabled`? I didn't find anything similar on Mac, except [NSInputContext](https://developer.apple.com/documentation/appkit/nstextinputcontext/1534378-activate?language=objc) has `activate/deactivate`, but it is invoked by the system only.

kchibisov:
>Could we just simply think that when winit window gets focused, send Enabled and when window loses focus, send Disabled?

The problem is on Wayland it works differently, the system tells you that IME enabled only when you get `Enabled` event from the system, it's not related to a focus, and can be triggered without focus change, we can model that on X11 around a focus events if it's preferred.

So I wanted to expose this, since if you don't have text-input you won't ever get Enabled, thus there's no point in requesting ime updates all the time.

src/platform_impl/linux/x11/ime/context.rs
garasubo: Yes

garasubo:
@kchibisov Thanks. LGTM.

kchibisov (approved):

src/event.rs
kaiwk: > I mean, I don't understand how end users will handle Start differently to Preedit? Could you give an example where handling of those 2 are different and it makes sense?

A example I met before is, I need to know whether I should handle `ReceiveCharacter` or not. When I get `PreeditStart`, I can set a flag like `isPreediting`, if `isPreediting` is true, I will ignore `ReceiveCharacter` event.

(However, we can do it in winit though, when preediting, we don't trigger `ReceiveCharacter`.)

I think it may be useful when user need to do stuff depend on IME state.

kchibisov: If we do preedit we don't need to send receivedchar, since the data is being send via preedit.

kaiwk: I'm not sure, but there may be other events in the similar situation. For example, user may also want to ignore `KeyboardInput` when do preedit. We can't take into account all the cases I suppose.

PR #1624, opened 2020-07-13
android: don't panic if ANativeWindow handle is not available

msiglreith:
**Change:** Return null-ptr instead of panic if the ANativeWindow handle is not available when requesting the RawWindowHandle

`ash-window` provides functionality for enumerating the required extensions based on the RawWindowHandle which would be happen before Instance creation in Vulkan. At this point, usually the ANativeWindow will not be available, which causes a panic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - On Windows, `set_ime_position` is now a no-op instead of a runtime crash.
 - On Android, `set_fullscreen` is now a no-op instead of a runtime crash.
 - On iOS and Android, `set_inner_size` is now a no-op instead of a runtime crash.
+- On Android, the native window handle returned by `raw_window_handle` might be null instead of panicking
 - **Breaking:** On Web, `set_cursor_position` and `set_cursor_grab` will now always return an error.
 - **Breaking:** `PixelDelta` scroll events now return a `PhysicalPosition`.
 
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -473,10 +473,18 @@ impl Window {
     pub fn set_cursor_visible(&self, _: bool) {}
 
     pub fn raw_window_handle(&self) -> raw_window_handle::RawWindowHandle {
+        // The native window is only available between `Resume` and `Suspend` events, otherwise it will be null.
+        // The returned handle may differ between calls after receiving a `Resume` event.
+
         let a_native_window = if let Some(native_window) = ndk_glue::native_window().as_ref() {
             unsafe { native_window.ptr().as_mut() as *mut _ as *mut _ }
         } else {
-            panic!("native window null");
+            // The handle to the native window is not available at the time of the call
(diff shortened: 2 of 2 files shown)


msiglreith:
cc @dvc94ch as you seem to be Android maintainer 😃

dvc94ch:
I've discussed this previously, people keep asking about it. In winit the Resumed/Suspended events are emitted. Only between these events is it ok to use the ANativeWindow

msiglreith:
Yes, I also added it as code comment in the PR. This is rather about returning a nullptr instead of panicking in the case, which is imo nicer.

dvc94ch:
A nill pointer can cause undefined behavior if it is for example dereferenced. Panicking is safe. What do you expect to do with the null pointer? It's incorrect usage of the api, which means it's a programmer error. There isn't anything you can do to recover and just looping until it returns not null is not something I'd want to encourage

msiglreith:
The use case is for WSI initialization on Vulkan.
For bootstrapping the user is required to specify the required WSI extensions on a per platform technology basis. This process , typically, happens at [program startup](https://github.com/norse-rs/ash-window/blob/master/examples/winit.rs#L19-L29). Ideally, this could be deferred directly from the Handle **type** returned when retrieving the `RawWindowHandle`, as some platforms support multiple different paths (android: https://github.com/norse-rs/ash-window/blob/master/src/lib.rs#L159). At this point I'm not really interested in the content only the type information. I would like to avoid just hardcoding the information based on a compiler switch but it's also okay if that results in being option.

msiglreith:
@francesca64 by chance you have some thoughts on this?

francesca64:
@msiglreith I definitely don't want to pass null pointers around, since I'm 100% sure it'd end up being a footgun for someone. It'd be less problematic to have Android's `RawWindwHandle` variant contain `Option<NonNull<_>>`, but I share @dvc94ch's sentiment that calling `raw_window_handle` in that case is a developer error, and letting people try to handle that at runtime doesn't make sense.

For your specific use-case, it'd likely be ideal for us to create a new enum that has all the variants of `RawWindowHandle` but without any data. It looks like that's exactly what's needed in the `enumerate_required_extensions` function you linked to. It'd be really easy to do (besides thinking of a name), but I'd want to hear an argument for why that'd be beneficial for any platform other than Android.

msiglreith:
Thanks for the feedback!

> For your specific use-case, it'd likely be ideal for us to create a new enum that has all the variants of RawWindowHandle but without any data. It looks like that's exactly what's needed in the enumerate_required_extensions function you linked to.

yes, I will raise that in the raw-window-handle issue tracker as it should be part of this crate rather.

> It'd be really easy to do (besides thinking of a name), but I'd want to hear an argument for why that'd be beneficial for any platform other than Android.

So far the only platform which profits from it would be Android up to my knowledge, which might be not too much of a driving force, I agree. Might need to adjust `ash-window` api or implementation depending on what the options/constraints are.

Closing it for now as this doesn't seem to be the right way forward 😸

PR #1625, opened 2020-07-16
Fix Window platform support documentation

chrisduerr:
This resolves various problems with the documentation about platform
support on the Window struct.

It also completely removes pointless runtime errors in favor of
consistent no-ops on all platforms that do not support a feature.


chrisduerr:
I've completely reworked this PR to include more than just `set_ime_position`. The entire `Window` struct should now have proper platform documentation.

PR #1630, opened 2020-07-23
Web: Use mouse events instead of pointer events if the latter isn't supported

josh65536:
Pointer events aren't supported on Safari yet, so this pull request uses mouse events when pointer events aren't supported on the web.


josh65536:
I'm not experienced with stdweb, so I'm not as confident about fixing the problem for that backend.

Issue #1636, opened 2020-07-28
[Android] EventLoop exists instantly


msiglreith:
The EventLoop hits an `ControlFlow::Exit` path instantly when starting execution on latest master

See more details in this thread: https://github.com/rust-windowing/winit/pull/1604#discussion_r453812286

PR #1638, opened 2020-07-29
android: fix event loop polling

msiglreith:
- Poll looper for `ControlFlow::Poll`, currently it won't progress in this case
- Don't exit the EventLoop when no new event has been polled from the Looper, which causes the program to instantly exit

fixes https://github.com/rust-windowing/winit/issues/1636

cc @VZout 

### DIFF
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -211,9 +211,7 @@ impl<T: 'static> EventLoop<T> {
                         );
                     }
                 }
-                None => {
-                    control_flow = ControlFlow::Exit;
-                }
+                None => {}
             }
 
             call_event_handler!(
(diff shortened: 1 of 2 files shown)


src/platform_impl/android/mod.rs
```diff
@@ -258,6 +256,11 @@ impl<T: 'static> EventLoop<T> {
                     break 'event_loop;
                 }
                 ControlFlow::Poll => {
+                    self.first_event = poll(
```
JasperDeSutter, line 259: Not polling here was an oversight in the initial implementation. It is indeed needed to support `ControlFlow::Poll`.

Jasper-Bekkers:
cc @MarijnS95

francesca64:
I'd like a CHANGELOG entry for this, since it's helpful way to document what bugs have been fixed in a given release.

msiglreith:
Added the `ControlFlow::Poll` fix to the CHANGELOG, the other fix only relates to an issue on current master not affecting any release.

PR #1644, opened 2020-08-05
🍏 iOS threading, initial work

Jasper-Bekkers:
The way our event loop works on iOS is that we call `run` and process all the windowing events on another thread (we send them over an mpsc channel). However, on iOS a lot of things need to happen on the main thread as documented / asserted in winit.

This follows along with some of the macOS code and uses Grand Central Dispatch to move the execution of some of these code blocks back to the main thread.

This doesn't remove all the asserts, nor does it fix all the cases, however I've moved over enough for our internal use-case.

 - [x] cargo fmt has been run on this branch
 - [x] cargo doc builds successfully
 - [ ] Added an entry to CHANGELOG.md if knowledge of this change could be valuable to users
 - [ ] Updated feature matrix, if new features were added or implemented

### DIFF
--- a/src/platform_impl/ios/window.rs
+++ b/src/platform_impl/ios/window.rs
@@ -315,7 +315,7 @@ impl Deref for Window {
     type Target = Inner;
 
     fn deref(&self) -> &Inner {
-       &self.inner
+        &self.inner
     }
 }
 
@@ -550,7 +550,7 @@ impl Inner {
         Queue::main().exec_sync(move || {
             let window: id = *window;
             let bounds: CGRect = msg_send![window, bounds];
-            
+
             if app_state::os_capabilities().safe_area {
                 let safe_area: UIEdgeInsets = msg_send![window, safeAreaInsets];
                 let safe_bounds = CGRect {
--- a/src/platform_impl/ios/window.rs
+++ b/src/platform_impl/ios/window.rs
@@ -4,6 +4,7 @@ use std::{
     ops::{Deref, DerefMut},
 };
 
+use dispatch::Queue;
 use objc::runtime::{Class, Object, BOOL, NO, YES};
 
 use crate::{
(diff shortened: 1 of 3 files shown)


Jasper-Bekkers:
@Osspial since you did the initial implementation, could you review this PR or redirect me somebody that can?

francesca64:
@Jasper-Bekkers I wrote the macOS EL2 backend, GCD included. The iOS EL2 backend was written by my very close co-worker @mtak-, so I'm the girl you're looking for!

Thanks so much for making this PR! It's fantastic to see more love being given to winit's mobile story. I'll probably find time to review this at some point this week (I've been a bit busier than usual).

Always feel free to ping me for anything iOS or Android related! I'm kinda sorta supposed to keep these things nice as part of my job.

Jasper-Bekkers:
Will do @francesca64, good to know I can ping you about this since I'll probably have more patches incoming over time as our use and experience with iOS continues.

francesca64:
This'll need a CHANGELOG entry, since I think it's a cool improvement to announce.

I hadn't noticed that the GCD usage in the macOS backend was refactored a while ago; to clarify, I wrote the raw FFI code that was there before that. I don't want it to seem like I took credit for something I didn't do!

Anyway, the code quality looks good! However, when `outer_size` calls `screen_frame` I'm hitting `EXC_BREAKPOINT` at the `dispatch_sync_f` within `exec_sync`'s implementation. Are you able to reproduce this? The app I'm testing with does all windowing on the main thread, if that makes a difference.

Also, I'm curious about the choice to use `exec_sync` over `exec_async`, which is used in the macOS backend.

Jasper-Bekkers:
>This'll need a CHANGELOG entry, since I think it's a cool improvement to announce.

Will do.

> Anyway, the code quality looks good! However, when outer_size calls screen_frame I'm hitting EXC_BREAKPOINT at the dispatch_sync_f within exec_sync's implementation. Are you able to reproduce this? The app I'm testing with does all windowing on the main thread, if that makes a difference.

I don't have an app like that so I have missed that during testing, I can set up a small repro to take a look at that when I get back to this next week.

> Also, I'm curious about the choice to use exec_sync over exec_async, which is used in the macOS backend.

`exec_async` doesn't allow for return values

src/platform_impl/ios/window.rs
```diff
@@ -493,86 +497,104 @@ impl Inner {
...
     }
 }
 
 impl Inner {
-    // requires main thread
+    // executed on main thread
     unsafe fn screen_frame(&self) -> CGRect {
-        self.to_screen_space(msg_send![self.window, bounds])
+        let window = self.window;
+        Queue::main().exec_sync(move || {
```
michaelkirk, line 511: > The way our event loop works on iOS is that we call run and process all the windowing events on another thread (we send them over an mpsc channel). However, on iOS a lot of things need to happen on the main thread as documented / asserted in winit.

It seems like this would deadlock for anyone calling these methods who is *already* on the main thread. Is it reasonable to say that users of this lib can't call these methods from the main thread. That is surprising to me.

A better solution for your situation might be to hoist these dispatches to your client code.

Or am I misunderstanding?

Jasper-Bekkers:
I've left this open for a while hoping I'd get back to it, but over the last 2 years I didn't find the time to get back to this. Closing to cleanup my open PR lists. 

If anyone drives by and is willing to pick this up feel free, but for me it's so low on the prio list I won't be able to get to it any time soon. When I do, I'll either re-open this or start a new PR entirely since it's probably diverged way too much.

madsmtm:
I hadn't seen this issue, but I have plans to fix this myself in https://github.com/rust-windowing/winit/issues/2464 in the near-ish future, so don't worry about it

PR #1652, opened 2020-08-14
Fix HiDPI vs. set_cursor_icon for web

michaelkirk:
PhysicalSize is recorded as canvas.size, whereas LogicalSize is stored
as canvas.style.size.

The previous cursor behavior on stdweb clobbered all style - thus losing
the LogicalSize.

- [x] Tested on all platforms changed (tested with stdweb and web_sys example here: https://github.com/michaelkirk/winit_example)
- [x] Compilation warnings were addressed (removed some "unused import" warning that predate my change)
- [] Created or updated an example program if it would help users understand this functionality
- [] Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented

Here are some before/after screencasts from the app where I encountered this bug. It's hard to make sense of looping GIFs - the video starts when my cursor is on the right DOM-inspector pane, then the canvas is made much larger as my mouse enters the canvas because our app sets the cursor icon (to "auto" so it's an invisible cursor change, but enough to trigger the bug).

You can see that as soon as we edit the cursor, we lose the style info for the canvas, thus lose it's "logical" sizing, relying on the "physical" sizing.

**before**
before mov

**after**
after mov

### DIFF
--- a/src/platform_impl/web/window.rs
+++ b/src/platform_impl/web/window.rs
@@ -165,8 +165,7 @@ impl Window {
             CursorIcon::RowResize => "row-resize",
         };
         *self.previous_pointer.borrow_mut() = text;
-        self.canvas
-            .set_attribute("style", &format!("cursor: {}", text));
+        backend::set_canvas_style_property(self.canvas.raw(), "cursor", text);
     }
 
     #[inline]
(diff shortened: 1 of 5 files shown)


src/platform_impl/web/window.rs
```diff
@@ -165,8 +165,7 @@ impl Window {
             CursorIcon::RowResize => "row-resize",
         };
         *self.previous_pointer.borrow_mut() = text;
-        self.canvas
-            .set_attribute("style", &format!("cursor: {}", text));
+        backend::set_canvas_style_property(self.canvas.raw(), "cursor", text);
```
michaelkirk, line 168: here's the bug - clobbering everything in style.

PR #1664, opened 2020-08-19
Fix crash on NetBSD

chrisduerr:
The `_lwp_self` function cannot be used to reliably determine the main
thread, see
https://github.com/alacritty/alacritty/issues/2631#issuecomment-676723289.

It might always be equal to the PID, but it's certainly not always 1
when the thread is the main thread.

However, Rust's built in `Thread::id` and `Thread::name` function will
always return `ThreadId(1)` and `Some("main")`. Since converting the
thread's ID to a number is not supported on stable Rust, checking that
the thread is labeled `Some("main")` seems like the most reliable
option. It should also be a good fallback in general.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 - On macOS, add `NSWindow.hasShadow` support.
 - **Breaking:** On Web, `set_cursor_position` and `set_cursor_grab` will now always return an error.
 - **Breaking:** `PixelDelta` scroll events now return a `PhysicalPosition`.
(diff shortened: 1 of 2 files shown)


tnn2:
Hi! I have confirmed that this change solves the problem we observed on NetBSD.

kchibisov (approved):
Well, it's not something that 'reliable' as well, but we don't have much options either for now, unless rust let us compare tid. The `_lwp_self` function was checked, and it indeed returns random things other than `1`, so I'm inclined to merge this PR as is, knowing that it fixes issues for real netbsd users.

PR #1666, opened 2020-08-20
Ability to force a theme on Windows

VZout:
Added the `WindowBuilderExtWindows::with_theme` function to allow forcing a theme. In case you always want to have a dark theme no matter the system settings.


repi:
Any feedback about this or merging it in?

chrisduerr:
I don't think there's any Windows maintainer capable of reviewing this.

src/platform_impl/windows/window_state.rs
```diff
@@ -32,6 +32,7 @@ pub struct WindowState {
     pub modifiers_state: ModifiersState,
     pub fullscreen: Option<Fullscreen>,
     pub is_dark_mode: bool,
+    pub forced_theme: Option<Theme>,
```
filnet, line 35: would it make sense to get rid of is_dark_mode and use only theme ?
probably a breaking change.

VZout: So forcing a theme is added alongside the normal functionality of using the windows theme settings.
Also a window is not always use the preferred theme on old windows versions.

A approach could be extending the theme enum to be something like:
````rust
enum {
  OS_DETERMINED,
  FORCED_DARK,
  FORCED_LIGHT,
}
````
and `OS_DETERMINED` would be the default.


However this does not get rid of `is_dark` because using dark style can fail and it is used to check the OS preference.

**Review by msiglreith — APPROVED** [review]
Tested on win10:
- `None`: Correctly shows the theme selected in the OS
- `Some(theme)`: Correctly overrides the theme set by the user
- `None` + theme switch: Correctly switches to OS theme
- `Some(theme)` + theme swtich: Correctly keeps the forced theme

Good to go imo if there are no further code related concerns

File: CHANGELOG.md [hunk 2]
```diff
@@ -7,6 +7,7 @@
 - On Web, prevent the webpage from scrolling when the user is focused on a winit canvas
 - On Web, calling `window.set_cursor_icon` no longer breaks HiDPI scaling
 - On Windows, drag and drop is now optional and must be enabled with `WindowBuilderExtWindows::with_drag_and_drop(true)`.
+- On Windows, force themes now with `WindowBuilderExtWindows::with_theme(Theme::)`.
```
kchibisov, line 10: The location is wrong, please rebase that PR properly.

I'd also word it like

```suggestion
- On Windows, add `WindowBuilderExtWindows::with_theme(Theme::)` to force specific theme.
```

FEATURES.md
```diff
@@ -117,7 +117,7 @@ If your PR makes notable changes to Winit's features, please update this section
 * Setting the taskbar icon
 * Setting the parent window
 * `WS_EX_NOREDIRECTIONBITMAP` support
-* Theme the title bar according to Windows 10 Dark Mode setting
+* Theme the title bar according to Windows 10 Dark Mode setting or force a theme
```
kchibisov, line 120: I don't like `a`, since you're forcing some specific theme and not current one, etc. If you have other ideas on how to show the intent more clear, let me know.

```suggestion
* Theme the title bar according to Windows 10 Dark Mode setting or force preferred theme
```

src/platform/windows.rs
```diff
@@ -125,6 +125,9 @@ pub trait WindowBuilderExtWindows {
     /// If you need COM API with `COINIT_MULTITHREADED` you must initialize it before calling any winit functions.
     /// See https://docs.microsoft.com/en-us/windows/win32/api/objbase/nf-objbase-coinitialize#remarks for more information.
     fn with_drag_and_drop(self, flag: bool) -> WindowBuilder;
+
+    /// Forces either Dark or Light theme. By default the system settings are used.
```
kchibisov, line 129: Just by looking on that function, I feel like it'll make sense to accept `Option<Theme>`, meaning that providing `None` will use system's preferred `Theme`. That seems more ergonomic and can interact better with downstream configurations, since otherwise they should avoid calling that method, what do you think?

src/platform_impl/windows/mod.rs
```diff
@@ -13,13 +13,15 @@ pub use self::icon::WinIcon as PlatformIcon;
...
 use crate::icon::Icon;
+use crate::window::Theme;
 
 #[derive(Clone)]
 pub struct PlatformSpecificWindowBuilderAttributes {
     pub parent: Option<HWND>,
     pub taskbar_icon: Option<Icon>,
     pub no_redirection_bitmap: bool,
     pub drag_and_drop: bool,
+    pub forced_theme: Option<Theme>,
```
kchibisov, line 24: I don't really like the word `force` here, maybe we should use `prefer` and in some other places?

```suggestion
    pub preferred_theme: Option<Theme>,
```

kchibisov (changes requested):
I don't have windows machine to test it, but I'd do some source level review, since I have some style questions.

src/platform_impl/windows/dark_mode.rs
```diff
@@ -72,9 +74,13 @@ lazy_static! {
...
 /// Returns true if dark mode was set, false if not.
-pub fn try_dark_mode(hwnd: HWND) -> bool {
+pub fn try_dark_mode(hwnd: HWND, preferred_theme: Theme) -> bool {
     if *DARK_MODE_SUPPORTED {
-        let is_dark_mode = should_use_dark_mode();
+        let is_dark_mode: bool = if preferred_theme != Theme::System {
+            preferred_theme == Theme::Dark
+        } else {
+            should_use_dark_mode()
+        };
```
kchibisov, line 83: Could you rewrite it with `let is_dark_mode = match preffered_theme`?

src/window.rs
```diff
@@ -865,6 +865,7 @@ pub enum Fullscreen {
 
 #[derive(Clone, Debug, PartialEq)]
 pub enum Theme {
+    System,
```
kchibisov, line 868: I don't like that change, and still prefer the use of `Option`, since we use `Theme` in `ThemeChanged` event, and so if the user  gets `System` it is confusing, since they don't know how to adapt to it. And even if winit can't send `System` users are still have to do something about it, so let's go back to `Option` approach to avoid that confusion.

kchibisov: Note, you can internally map `None` to `System`, but you shouldn't expose that API to the user, since users don't have a way to handle `System` variant.

src/platform_impl/windows/event_loop.rs
```diff
@@ -1871,20 +1871,27 @@ unsafe extern "system" fn public_window_callback<T: 'static>(
...
+                let mut window_state = subclass_input.window_state.lock();
+                let changed = window_state.is_dark_mode != is_dark_mode;
 
-                window_state.is_dark_mode = is_dark_mode;
-                mem::drop(window_state);
-                subclass_input.send_event(Event::WindowEvent {
-                    window_id: RootWindowId(WindowId(window)),
-                    event: ThemeChanged(theme),
-                });
+                if changed {
```
kchibisov, line 1881: Could you 'inline' changed thing?

kchibisov (changes requested):

src/platform_impl/windows/event_loop.rs
VZout: Done! Also implemented your other feedback.

src/platform_impl/windows/dark_mode.rs
```diff
@@ -72,9 +74,12 @@ lazy_static! {
 
 /// Attempt to set dark mode on a window, if necessary.
 /// Returns true if dark mode was set, false if not.
-pub fn try_dark_mode(hwnd: HWND) -> bool {
+pub fn try_dark_mode(hwnd: HWND, preferred_theme: Option<Theme>) -> bool {
     if *DARK_MODE_SUPPORTED {
-        let is_dark_mode = should_use_dark_mode();
+        let is_dark_mode: bool = match preferred_theme {
```
filnet, line 79: do you really need to specify the bool type ?

src/platform_impl/windows/event_loop.rs
```diff
@@ -1871,20 +1871,26 @@ unsafe extern "system" fn public_window_callback<T: 'static>(
         winuser::WM_SETTINGCHANGE => {
             use crate::event::WindowEvent::ThemeChanged;
 
-            let is_dark_mode = try_dark_mode(window);
-            let mut window_state = subclass_input.window_state.lock();
-            let changed = window_state.is_dark_mode != is_dark_mode;
+            let preferred_theme = subclass_input.window_state.lock().preferred_theme.clone();
```
filnet, line 1874: not sure you need to clone (it's an enum)

src/platform_impl/windows/window.rs
```diff
@@ -726,14 +726,15 @@ unsafe fn init<T: 'static>(
     // If the system theme is dark, we need to set the window theme now
     // before we update the window flags (and possibly show the
     // window for the first time).
-    let dark_mode = try_dark_mode(real_window.0);
+    let dark_mode = try_dark_mode(real_window.0, pl_attribs.preferred_theme.clone());
```
filnet, line 729: is clone() needed ?

src/platform/windows.rs
```diff
@@ -125,6 +125,9 @@ pub trait WindowBuilderExtWindows {
     /// If you need COM API with `COINIT_MULTITHREADED` you must initialize it before calling any winit functions.
     /// See https://docs.microsoft.com/en-us/windows/win32/api/objbase/nf-objbase-coinitialize#remarks for more information.
     fn with_drag_and_drop(self, flag: bool) -> WindowBuilder;
+
+    /// Forces a theme or use the system settings if None is provided.
```
kchibisov, line 129:
```suggestion
    /// Forces a theme or uses the system settings if `None` was provided.
```

src/platform_impl/windows/window_state.rs
```diff
@@ -32,6 +32,7 @@ pub struct WindowState {
     pub modifiers_state: ModifiersState,
     pub fullscreen: Option<Fullscreen>,
     pub is_dark_mode: bool,
```
kchibisov, line 34: I feel like for the more code simplicity we can use `Theme` instead of `bool` here, so `try_dark_mode` will also return a `Theme`, and then you'd have less mapping to `bool -> Theme` , issuing a new `ThemeChanged` when `current_theme != new_theme`?

kchibisov (changes requested):
That looks better and I feel like we're getting closer! Thanks for your work so  far.

src/platform_impl/windows/window_state.rs
VZout: Yeah I like this changed. Implemented it!

src/platform_impl/windows/dark_mode.rs
```diff
@@ -72,9 +74,12 @@ lazy_static! {
 
 /// Attempt to set dark mode on a window, if necessary.
 /// Returns true if dark mode was set, false if not.
```
kchibisov, line 76: The comment isn't not correct anymore, we now return `Theme`.

src/platform_impl/windows/dark_mode.rs
```diff
@@ -72,9 +74,12 @@ lazy_static! {
...
-pub fn try_dark_mode(hwnd: HWND) -> bool {
+pub fn try_dark_mode(hwnd: HWND, preferred_theme: Option<Theme>) -> Theme {
```
kchibisov, line 77: I'd actually rename this function, since it it's not about dark mode anymore, it's just about trying to get a theme, however when I see try, I'm expecting `Result`, so I'm not sure if this is a good name. Maybe `pick_theme` or something like that, if you have better ideas let me know.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Unreleased
 
 - On Android, calling `WindowEvent::Focused` now works properly instead of always returning false. 
+- On Windows, add `WindowBuilderExtWindows::with_theme(Theme::)` to set a preferred theme.
```
kchibisov, line 4: Should append into the changelog, and not insert between. Also, there's no need to provide argument in changelog, just a name.

```suggestion
- On Windows, add `WindowBuilderExtWindows::with_theme` to set a preferred theme.
```

kchibisov (changes requested):
You should address compilation warnings, besides that it should be good to go.

src/platform_impl/windows/dark_mode.rs
VZout: if your preferred theme is dark it could theoretically fail on old windows versions. In that case it would return `Theme::Light`.
So I think `try` should be included in the name. `try_select_theme`, `try_pick_theme` or `try_theme` 🤷 

I don't think it should necessarily return a result. I could change it to a result but it would make the calling code less nice IMO. (It would return a result, if the result is ok the current theme is the preferred theme if the result is a error than the theme is light. Which works but the caller of `try_dark_mode` needs to understand that the standard theme is light so I prefer to hide it away into the `try_` function)

CHANGELOG.md
VZout: 👍 commited

src/platform_impl/windows/dark_mode.rs
VZout: fixed

kchibisov (approved):
Needs rebase and should be good to go.

src/platform/windows.rs
```diff
@@ -82,7 +82,7 @@ pub trait WindowExtWindows {
     fn set_taskbar_icon(&self, taskbar_icon: Option<Icon>);
 
     /// Whether the system theme is currently Windows 10's "Dark Mode".
```
msiglreith, line 84: Documentation doesn't match anymore

src/platform/windows.rs
```diff
@@ -82,7 +82,7 @@ pub trait WindowExtWindows {
...
-    fn is_dark_mode(&self) -> bool;
+    fn theme(&self) -> Theme;
```
msiglreith, line 85: This seems like a breaking change, so it should be listed in the changelog

VZout: Changed the comment

src/platform/windows.rs
VZout: Changed the comment

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Unreleased
 
+- ***Breaking:** On Windows, Renamed `WindowBuilderExtWindows`'s is_dark_mode` to `theme`.
```
msiglreith, line 3:
```suggestion
- **Breaking:** On Windows, Renamed `WindowBuilderExtWindows`'s `is_dark_mode` to `theme`.
```
(see https://github.com/rust-windowing/winit/blob/1ef38f04127d3399b792aefcbcb233cf390ab5d7/CHANGELOG.md)

PR #1667, opened 2020-08-20
Add option to blur background on macOS

joboet:
Related to #538 

This PR adds `with_blurred_background` to `WindowBuilderExtMacOS` which makes the view semi-transparent and blurs the background of the window using a `NSVisualEffectView`.

I think this feature could conflict with other cross-platform implementations, but as there is no standard for blurred backgrounds on Wayland or the web, these features would need to be platform-gated anyway.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- On macOS, added `with_blurred_background` to make the window semi-transparent and blur the background.
 - On iOS, fixed support for the "Debug View Heirarchy" feature in Xcode.
 - On all platforms, `available_monitors` and `primary_monitor` are now on `EventLoopWindowTarget` rather than `EventLoop` to list monitors event in the event loop.
 - On Unix, X11 and Wayland are now optional features (enabled by default)
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -126,6 +126,7 @@ If your PR makes notable changes to Winit's features, please update this section
 * Hidden titlebar
 * Hidden titlebar buttons
 * Full-size content view
+* Blurred background
 
 ### Unix
 * Window urgency
--- a/src/platform/macos.rs
+++ b/src/platform/macos.rs
@@ -148,6 +148,8 @@ pub trait WindowBuilderExtMacOS {
     fn with_resize_increments(self, increments: LogicalSize<f64>) -> WindowBuilder;
     fn with_disallow_hidpi(self, disallow_hidpi: bool) -> WindowBuilder;
     fn with_has_shadow(self, has_shadow: bool) -> WindowBuilder;
+    /// Blurs the background.
+    fn with_blurred_background(self, blurred_background: bool) -> WindowBuilder;
 }
 
 impl WindowBuilderExtMacOS for WindowBuilder {
@@ -211,6 +213,12 @@ impl WindowBuilderExtMacOS for WindowBuilder {
     #[inline]
     fn with_has_shadow(mut self, has_shadow: bool) -> WindowBuilder {
         self.platform_specific.has_shadow = has_shadow;
+        self
+    }
+
+    #[inline]
+    fn with_blurred_background(mut self, blurred_background: bool) -> WindowBuilder {
+        self.platform_specific.blurred_background = blurred_background;
         self
     }
 }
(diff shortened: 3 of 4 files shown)


kchibisov:
Can this be triggered during runtime? I think like it could be done? Limiting it to be enabled only on startup seems a bit lacking for what its worth.

> but as there is no standard for blurred backgrounds on Wayland

That's true, I'll ask if there's something not widely supported but still a known way to do things.

on X11 there's no way to do it either, it just some compositors are doing it somehow. You can do something for KDE, I guess, but that's it.

Also, can you control the blur effect or it's just a generic blur? If so it'll be nice to provide some control over that to the users.

CoffeeBlend:
Yes, I suppose this could be done at runtime, you can disable the blur by setting the state property.
The blur effect is determined by macOS, there is a way to add a custom filter with CoreGraphics, but that is a bit more complicated and I'm not sure whether it's allowed on the App Store.

By the way, please don't merge this, the view doesn't resize with the window yet.

kchibisov:
>By the way, please don't merge this, the view doesn't resize with the window yet.

I have no intent in merging non trivial code on a platforms I can't test, don't worry. But you should open as draft or mark as [WIP] in a title when you open things that you don't want being merged (you've updated that already, but saying for the future).

Also, I'd like to wait for other backends maintainers to chime in and tell how the blur is working on their platforms ( I personally have no idea how it could work on Windows).

cc @Osspial for blur on Windows.
cc @murarth whether winit should set DE specific things like `_KDE_NET_WM_BLUR_BEHIND_REGION` if the window blur was requested.

kchibisov:
As for the Wayland, there's a https://github.com/KDE/kwayland/blob/master/src/client/protocols/blur.xml, which should be pretty straightforward to implement inside of winit( I can submit impl). In general it works in a way that you can request to blur some **parts** of the window, so if other platforms have such concept we may accept an array of rects (PhysicalPosition, PhysicalSize things) to set blur.

kchibisov:
Oh, and you may actually be interested in how the previous effort was going https://github.com/rust-windowing/winit/pull/568 , it had macOS bits working at least.

CoffeeBlend:
Thanks for the tip 😉. The resizing works now. 
About the activation at runtime, IMHO my current design fits better into the style of the transparancy option, having one availiable at runtime and the other not is a bit incoherent. There is however an option in NSVisualEffectView to always blur, even when the window is out of focus, maybe having that availiable to the user is better? At the moment I'm just choosing the default.

kchibisov:
>About the activation at runtime, IMHO my current design fits better into the style of the transparancy option

The thing is that transparency option doesn't mean that you will or won't have it, and users may draw with alpha 1, so the window will be opaque, however they'll be able to go for opacity only for certain regions, and for those regions they may want to have blur. Requesting blur without ever making blurring the window(user just doesn't  trigger a state, where something transparent will be shown) seems strange, and the system may interpret it in a weird way.

kchibisov:
>even when the window is out of focus

Not sure about that one, but I guess on macOS everyone is blurring even unfocused windows, so that could make a bit more sense?

CoffeeBlend:
> Not sure about that one, but I guess on macOS everyone is blurring even unfocused windows, so that could make a bit more sense?

No, Finder for example only blurs when in focus. But I guess every app is handling that differently…

kchibisov:
>No, Finder for example only blurs when in focus. But I guess every app is handling that differently…

Hm, fair enough. That part of when to blur seems macOS specific and could be passed a special window creation attribute for macOS, what do you think?

CoffeeBlend:
Sure, that sounds fine.
About the activation, the filter allows a mask image to be set, so we can activate the blur for specific zones only. We could implement an API similar to the KWayland protocol (with regions of blur) and have the mask image set to those regions. That would work on both macOS and Linux.

kchibisov:
>About the activation, the filter allows a mask image to be set, so we can activate the blur for specific zones only. We could implement an API similar to the KWayland protocol (with regions of blur) and have the mask image set to those regions. That would work on both macOS and Linux.

Yeah, it'll be nice to have such control exposed. However I'd like to see how platforms like Windows are working here, I'd assume that they use something very similar, but I can't say anything about it, since I'm not using it.

joboet:
Confused Gandalf: I have no memory of this PR

Wow, this PR is old 😄! I noticed `set_blur` got added a while ago, I'll go ahead and close this...

PR #1669, opened 2020-08-21
add composition event support for Mac

kaiwk:
   - [x] Tested on Mac

see: #1497


kchibisov:
I'll take a look on it once we figure out renaming and preferred event structure on #1622, if you have any concerns or questions you can chime in.

kchibisov:
We've updated the composition branch, so if you can update your PR based on that it'll be nice.

kchibisov:
Note, you'd need to send IME::Enable at some point, though, given that both X11 and macOS doesn't have such information I may end up removing this even entirely.

kaiwk:
Not sure why CI failed.

Issue #793, opened 2019-02-13
`get_current_monitor()` panics on Wayland.


icefoxen:
See: https://github.com/ggez/ggez/issues/579

Basically, this function does a `some_vec.last().unwrap()` of a `Vec` that can be null.  The vec appears to be acquired from `sctk::surface::get_outputs()` here: https://github.com/Smithay/client-toolkit/blob/2c28ada5c7dd067ac1245382bc371cc933d4c416/src/surface.rs#L109 .  As far as I can tell, the `outputs` vec is created empty and only has things added to it on `SurfaceUserData::enter()`, but I don't know enough about the Wayland protocol to understand what's actually going on there.

vberger:
Welp.

If this `Vec` is empty, it means that the surface is currently not being displayed on any monitor, likely because the method was called early in the init process and the window is not currently displayed yet.

I'm not sure what would be the best thing to do here, given the API winit enforces. Would it make sens to take advantage of the pending breaking changes to change the prototype of `get_current_motinor()` to return an `Option<MonitorId>` ? (or even to return a `Vec<MonitorId>` tbh)

Osspial:
I'd be fine with either returning `Option<MonitorId>` or `Vec<MonitorId>`.

If we go with `Vec<MonitorId>` we should consider if we define an order for the returned monitors (say, order it by the area of intersection of the window of each monitor). That should be possible on win32, but I don't know what the story on that looks like.

vberger:
> order it by the area of intersection of the window of each monitor

That would not be possible for wayland. The only order that can be implemented there would be « the order in which the window entered the monitors », which is hardly useful.

Osspial:
Thanks, good to know. Is there any way to find out a window's primary monitor on Wayland? I ask because on systems where different monitors have different DPIs, it can be useful to know which monitor the system considers to be the main monitor for DPI scaling, which I'd like to support if possible (on Windows that's done through the [`MonitorFromWindow`](https://docs.microsoft.com/en-us/windows/desktop/api/winuser/nf-winuser-monitorfromwindow) function). If we can expose that alongside a list of all monitors the window overlaps with that would be awesome.

vberger:
There is no concept of "primary monitor" in wayland. Regarding DPI, what happens is that the client app actually tells the server that it is drawing its contents using a given dpi factor, and the server takes care of upscaling or downscaling if that does not match the dpi factor of the monitor.

[...]

PR #1670, opened 2020-08-21
Rework Fullscreen::Borderless enum variant

kchibisov:
This changes `Fullscreen::Borderless` enum variant from
`Fullscreen::Borderless(MonitorHandle)` to
`Fullscreen::Borderless(Option<MonitorHandle>)`. Providing
`None` to it will result in picking system's preferred monitor.

This is an implementation of proposal from https://github.com/rust-windowing/winit/issues/793#issuecomment-676820743 . I can only speak for Wayland implementation here. So `macOS`,`ios`,`X11`, and `Windows`, implementations are interested and likely wrong or inefficient to do what this variant should do.

cc for checking implementations
@francesca64 (not sure if you're active checking macOS/ios stuff now, let me know if you don't want to take a look)
@Osspial (windows) The implementation I've provided needs fixing for sure, since I have a hard time understand how windows even works in winit.
@murarth  (X11)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@
 - On Web, `WindowEvent::Resized` is now emitted when `Window::set_inner_size` is called.
 - **Breaking:** `Fullscreen` enum now uses `Option<Borderless>` instead of `Borderless` to pick system preferred monitor.
 - **Breaking:** `Fullscreen` enum now uses `Borderless<Option<MonitorHandle>>` instead of `Borderless<MonitorHandle>` to pick system's preferred monitor.
+- **Breaking:** `Fullscreen` enum now uses `Borderless<Option<MonitorHandle>>` instead of `Borderless<MonitorHandle>` to allow picking the current monitor.
 
 # 0.22.2 (2020-05-16)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@
 - On Web, the event handler closure passed to `EventLoop::run` now gets dropped after the event loop is destroyed.
 - **Breaking:** On Web, the canvas element associated to a `Window` is no longer removed from the DOM when the `Window` is dropped.
 - On Web, `WindowEvent::Resized` is now emitted when `Window::set_inner_size` is called.
+- **Breaking:** `Fullscreen` enum now uses `Option<Borderless>` instead of `Borderless` to pick system preferred monitor.
 
 # 0.22.2 (2020-05-16)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@
 - **Breaking:** On Web, the canvas element associated to a `Window` is no longer removed from the DOM when the `Window` is dropped.
 - On Web, `WindowEvent::Resized` is now emitted when `Window::set_inner_size` is called.
 - **Breaking:** `Fullscreen` enum now uses `Option<Borderless>` instead of `Borderless` to pick system preferred monitor.
+- **Breaking:** `Fullscreen` enum now uses `Borderless<Option<MonitorHandle>>` instead of `Borderless<MonitorHandle>` to pick system's preferred monitor.
 
 # 0.22.2 (2020-05-16)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,9 +41,7 @@
 - On Web, the event handler closure passed to `EventLoop::run` now gets dropped after the event loop is destroyed.
 - **Breaking:** On Web, the canvas element associated to a `Window` is no longer removed from the DOM when the `Window` is dropped.
 - On Web, `WindowEvent::Resized` is now emitted when `Window::set_inner_size` is called.
-- **Breaking:** `Fullscreen` enum now uses `Option<Borderless>` instead of `Borderless` to pick system preferred monitor.
-- **Breaking:** `Fullscreen` enum now uses `Borderless<Option<MonitorHandle>>` instead of `Borderless<MonitorHandle>` to pick system's preferred monitor.
-- **Breaking:** `Fullscreen` enum now uses `Borderless<Option<MonitorHandle>>` instead of `Borderless<MonitorHandle>` to allow picking the current monitor.
+- **Breaking:** `Fullscreen` enum now uses `Borderless(Option<MonitorHandle>)` instead of `Borderless(MonitorHandle)` to allow picking the current monitor.
 
 # 0.22.2 (2020-05-16)
 
--- a/examples/window_debug.rs
+++ b/examples/window_debug.rs
@@ -21,6 +21,7 @@ fn main() {
     eprintln!("debugging keys:");
     eprintln!("  (E) Enter exclusive fullscreen");
     eprintln!("  (F) Toggle borderless fullscreen");
+    eprintln!("  (P) Toggle borderless fullscreen on system's preffered monitor");
     eprintln!("  (M) Toggle minimized");
     eprintln!("  (Q) Quit event loop");
     eprintln!("  (V) Toggle visibility");
@@ -85,8 +86,15 @@ fn main() {
                         if window.fullscreen().is_some() {
                             window.set_fullscreen(None);
                         } else {
-                            let monitor = window.current_monitor().unwrap();
+                            let monitor = window.current_monitor();
                             window.set_fullscreen(Some(Fullscreen::Borderless(monitor)));
+                        }
+                    }
+                    VirtualKeyCode::P => {
+                        if window.fullscreen().is_some() {
+                            window.set_fullscreen(None);
+                        } else {
+                            window.set_fullscreen(Some(Fullscreen::Borderless(None)));
                         }
                     }
                     VirtualKeyCode::M => {
--- a/src/platform_impl/ios/view.rs
+++ b/src/platform_impl/ios/view.rs
@@ -519,7 +519,15 @@ pub unsafe fn create_window(
             msg_send![window, setScreen:video_mode.monitor().ui_screen()]
         }
         Some(Fullscreen::Borderless(ref monitor)) => {
-            msg_send![window, setScreen:monitor.ui_screen()]
+            let uiscreen: id = match monitor {
+                Some(ref monitor) => monitor.ui_screen() as id,
+                None => {
+                    let uiscreen: id = msg_send![window, screen];
+                    uiscreen
+                }
+            };
+
+            msg_send![window, setScreen: uiscreen]
         }
         None => (),
     }
--- a/src/platform_impl/ios/view.rs
+++ b/src/platform_impl/ios/view.rs
@@ -519,8 +519,8 @@ pub unsafe fn create_window(
             msg_send![window, setScreen:video_mode.monitor().ui_screen()]
         }
         Some(Fullscreen::Borderless(ref monitor)) => {
-            let uiscreen: id = match monitor {
-                Some(ref monitor) => monitor.ui_screen() as id,
+            let uiscreen: id = match &monitor {
+                Some(monitor) => monitor.ui_screen() as id,
                 None => {
                     let uiscreen: id = msg_send![window, screen];
                     uiscreen
--- a/src/platform_impl/ios/window.rs
+++ b/src/platform_impl/ios/window.rs
@@ -196,7 +196,8 @@ impl Inner {
                     let () = msg_send![uiscreen, setCurrentMode: video_mode.video_mode.screen_mode];
                     uiscreen
                 }
-                Some(Fullscreen::Borderless(monitor)) => monitor.ui_screen() as id,
+                Some(Fullscreen::Borderless(Some(monitor))) => monitor.ui_screen() as id,
+                Some(Fullscreen::Borderless(None)) => self.current_monitor().ui_screen() as id,
                 None => {
                     warn!("`Window::set_fullscreen(None)` ignored on iOS");
                     return;
@@ -235,7 +236,7 @@ impl Inner {
                 && screen_space_bounds.size.width == screen_bounds.size.width
                 && screen_space_bounds.size.height == screen_bounds.size.height
             {
-                Some(Fullscreen::Borderless(monitor))
+                Some(Fullscreen::Borderless(Some(monitor)))
             } else {
                 None
             }
@@ -353,8 +354,10 @@ impl Window {
                 Some(Fullscreen::Exclusive(ref video_mode)) => {
                     video_mode.video_mode.monitor.ui_screen() as id
                 }
-                Some(Fullscreen::Borderless(ref monitor)) => monitor.ui_screen() as id,
-                None => monitor::main_uiscreen().ui_screen(),
+                Some(Fullscreen::Borderless(Some(ref monitor))) => monitor.inner.ui_screen(),
+                Some(Fullscreen::Borderless(None)) | None => {
+                    monitor::main_uiscreen().ui_screen() as id
+                }
             };
 
             let screen_bounds: CGRect = msg_send![screen, bounds];
--- a/src/platform_impl/ios/window.rs
+++ b/src/platform_impl/ios/window.rs
@@ -196,10 +196,9 @@ impl Inner {
                     let () = msg_send![uiscreen, setCurrentMode: video_mode.video_mode.screen_mode];
                     uiscreen
                 }
-                Some(Fullscreen::Borderless(Some(monitor))) => monitor.ui_screen() as id,
-                Some(Fullscreen::Borderless(None)) => {
-                    self.current_monitor_inner().ui_screen() as id
-                }
+                Some(Fullscreen::Borderless(monitor)) => monitor
+                    .unwrap_or_else(|| self.current_monitor_inner())
+                    .ui_screen() as id,
                 None => {
                     warn!("`Window::set_fullscreen(None)` ignored on iOS");
                     return;
--- a/src/platform_impl/ios/window.rs
+++ b/src/platform_impl/ios/window.rs
@@ -197,7 +197,7 @@ impl Inner {
                     uiscreen
                 }
                 Some(Fullscreen::Borderless(Some(monitor))) => monitor.ui_screen() as id,
-                Some(Fullscreen::Borderless(None)) => self.current_monitor().ui_screen() as id,
+                Some(Fullscreen::Borderless(None)) => self.current_monitor_inner().ui_screen() as id,
                 None => {
                     warn!("`Window::set_fullscreen(None)` ignored on iOS");
                     return;
--- a/src/platform_impl/ios/window.rs
+++ b/src/platform_impl/ios/window.rs
@@ -197,7 +197,9 @@ impl Inner {
                     uiscreen
                 }
                 Some(Fullscreen::Borderless(Some(monitor))) => monitor.ui_screen() as id,
-                Some(Fullscreen::Borderless(None)) => self.current_monitor_inner().ui_screen() as id,
+                Some(Fullscreen::Borderless(None)) => {
+                    self.current_monitor_inner().ui_screen() as id
+                }
                 None => {
                     warn!("`Window::set_fullscreen(None)` ignored on iOS");
                     return;
--- a/src/platform_impl/linux/wayland/window.rs
+++ b/src/platform_impl/linux/wayland/window.rs
@@ -155,14 +155,14 @@ impl Window {
                 panic!("Wayland doesn't support exclusive fullscreen")
             }
             Some(Fullscreen::Borderless(monitor)) => {
-                let monitor = if let Some(RootMonitorHandle {
-                    inner: PlatformMonitorHandle::Wayland(monitor),
-                }) = monitor
-                {
-                    Some(monitor.proxy)
-                } else {
-                    None
-                };
+                let monitor = monitor.map(|monitor| {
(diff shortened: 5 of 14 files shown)


src/platform_impl/windows/event_loop.rs
```diff
@@ -851,7 +851,8 @@ unsafe extern "system" fn public_window_callback<T: 'static>(
                     winuser::MonitorFromRect(&new_rect, winuser::MONITOR_DEFAULTTONULL);
                 match fullscreen {
                     Fullscreen::Borderless(ref mut fullscreen_monitor) => {
-                        if new_monitor != fullscreen_monitor.inner.hmonitor()
+                        // FIXME This should be properly implemented.
```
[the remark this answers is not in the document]
kchibisov: Done

kchibisov:
I've implemented parts on all platforms and tested at least X11 and Wayland. The impl for macOS, ios, and windows are implemented in conservative way by mapping `None` to `current_monitor` early on.

CHANGELOG.md
```diff
@@ -33,6 +33,7 @@
 - On macOS, updated core-* dependencies and cocoa.
 - Bump `parking_lot` to 0.11
 - On Android, bump `ndk`, `ndk-sys` and `ndk-glue` to 0.2. Checkout the new ndk-glue main proc attribute.
+- **Breaking:** `Fullscreen` enum now uses `Option<Borderless>` instead of `Borderless` to pick system preferred monitor.
```
chrisduerr, line 36: `Borderless(Option<MonitorHandle>)`, not `Option<Borderless>`, right?

examples/window_debug.rs
```diff
@@ -85,10 +86,17 @@ fn main() {
                         if window.fullscreen().is_some() {
                             window.set_fullscreen(None);
                         } else {
-                            let monitor = window.current_monitor().unwrap();
+                            let monitor = window.current_monitor();
                             window.set_fullscreen(Some(Fullscreen::Borderless(monitor)));
```
chrisduerr, line 90: Isn't this the same as Borderless(None)?

src/platform_impl/ios/view.rs
```diff
@@ -519,7 +519,15 @@ pub unsafe fn create_window(
             msg_send![window, setScreen:video_mode.monitor().ui_screen()]
         }
         Some(Fullscreen::Borderless(ref monitor)) => {
-            msg_send![window, setScreen:monitor.ui_screen()]
+            let uiscreen: id = match monitor {
+                Some(ref monitor) => monitor.ui_screen() as id,
+                None => {
+                    let uiscreen: id = msg_send![window, screen];
+                    uiscreen
```
chrisduerr, line 526: Type inference issues I'd assume?

src/platform_impl/ios/view.rs
```diff
@@ -519,7 +519,15 @@ pub unsafe fn create_window(
             msg_send![window, setScreen:video_mode.monitor().ui_screen()]
         }
         Some(Fullscreen::Borderless(ref monitor)) => {
-            msg_send![window, setScreen:monitor.ui_screen()]
+            let uiscreen: id = match monitor {
+                Some(ref monitor) => monitor.ui_screen() as id,
```
chrisduerr, line 523: Usually should use `match &monitor` instead of `Some(ref monitor)`. There's a surprising amount of `ref`s here, given that they're almost never necessary in Rust anymore.

src/platform_impl/ios/window.rs
```diff
@@ -196,7 +196,10 @@ impl Inner {
                     let () = msg_send![uiscreen, setCurrentMode: video_mode.video_mode.screen_mode];
                     uiscreen
                 }
-                Some(Fullscreen::Borderless(monitor)) => monitor.ui_screen() as id,
+                Some(Fullscreen::Borderless(Some(monitor))) => monitor.ui_screen() as id,
+                Some(Fullscreen::Borderless(None)) => {
+                    self.current_monitor_inner().ui_screen() as id
+                }
```
chrisduerr, line 202: Could in theory combine these and do something like `monitor.unwrap_or_else(|_| self.current_monitor_inner()).ui_screen() as id` maybe?

src/platform_impl/ios/window.rs
```diff
@@ -353,8 +356,10 @@ impl Window {
                 Some(Fullscreen::Exclusive(ref video_mode)) => {
                     video_mode.video_mode.monitor.ui_screen() as id
                 }
-                Some(Fullscreen::Borderless(ref monitor)) => monitor.ui_screen() as id,
-                None => monitor::main_uiscreen().ui_screen(),
+                Some(Fullscreen::Borderless(Some(ref monitor))) => monitor.inner.ui_screen(),
+                Some(Fullscreen::Borderless(None)) | None => {
+                    monitor::main_uiscreen().ui_screen() as id
+                }
```
chrisduerr, line 362: I'd assume this is because we don't have any monitor yet? Should this be documented maybe that it's not something supported at startup?

src/platform_impl/linux/wayland/window.rs
```diff
@@ -154,11 +154,18 @@ impl Window {
...
-            Some(Fullscreen::Borderless(_)) => unreachable!(),
+            Some(Fullscreen::Borderless(monitor)) => {
+                let monitor = if let Some(RootMonitorHandle {
+                    inner: PlatformMonitorHandle::Wayland(monitor),
+                }) = monitor
+                {
+                    Some(monitor.proxy)
+                } else {
+                    None
+                };
```
chrisduerr, line 165: This looks suspiciously `map`y.

src/platform_impl/linux/wayland/window.rs
```diff
@@ -348,16 +357,18 @@ impl Window {
...
-                    .set_fullscreen(Some(&monitor_id.proxy));
+            Some(Fullscreen::Borderless(monitor)) => {
+                let monitor = if let Some(RootMonitorHandle {
+                    inner: PlatformMonitorHandle::Wayland(monitor),
+                }) = monitor
+                {
+                    Some(monitor.proxy)
+                } else {
+                    None
+                };
```
chrisduerr, line 368: This also looks very `map`y.

src/platform_impl/linux/x11/window.rs
```diff
@@ -585,6 +585,16 @@ impl UnownedWindow {
 
     fn set_fullscreen_inner(&self, fullscreen: Option<Fullscreen>) -> Option<util::Flusher<'_>> {
         let mut shared_state_lock = self.shared_state.lock();
+        // Map Borderless(`None`) to current monitor.
```
chrisduerr, line 588:
```suggestion
        let mut shared_state_lock = self.shared_state.lock();
        
        // Map Borderless(`None`) to current monitor.
```

I'd just give it a touch of space.

src/platform_impl/linux/x11/window.rs
```diff
@@ -585,6 +585,16 @@ impl UnownedWindow {
...
+        let fullscreen = match fullscreen {
+            Some(Fullscreen::Borderless(None)) => {
+                let current_monitor = Some(RootMonitorHandle {
+                    inner: PlatformMonitorHandle::X(self.current_monitor()),
+                });
+                Some(Fullscreen::Borderless(current_monitor))
+            }
+            _ => fullscreen,
+        };
```
chrisduerr, line 597: Could also make it a `mut fullscreen` in parameters and use `if let Some` here, right? I think that might come out a couple lines shorter.

src/platform_impl/linux/x11/window.rs
```diff
@@ -649,10 +659,12 @@ impl UnownedWindow {
...
-                    }) => (Some(video_mode), video_mode.monitor.as_ref().unwrap()),
-                    Fullscreen::Borderless(RootMonitorHandle {
-                        inner: PlatformMonitorHandle::X(ref monitor),
-                    }) => (None, monitor),
+                    }) => (Some(video_mode), video_mode.monitor.clone().unwrap()),
+                    Fullscreen::Borderless(Some(RootMonitorHandle {
+                        inner: PlatformMonitorHandle::X(monitor),
+                    })) => (None, monitor),
+                    // We've mapped `None` in borderless earlier.
+                    Fullscreen::Borderless(None) => unreachable!(),
```
chrisduerr, line 667: This looks a bit unfortunate to me. Why does it need to be remapped earlier instead of handled here? So comparison to the existing state works? I don't really see where the monitor is needed in the old/new comparison.

src/platform_impl/windows/event_loop.rs
```diff
@@ -853,7 +853,9 @@ unsafe extern "system" fn public_window_callback<T: 'static>(
                     winuser::MonitorFromRect(&new_rect, winuser::MONITOR_DEFAULTTONULL);
                 match fullscreen {
                     Fullscreen::Borderless(ref mut fullscreen_monitor) => {
-                        if new_monitor != fullscreen_monitor.inner.hmonitor()
+                        // Unwrap is safe, since we explicitly map `None` to current monitor
+                        // when we're setting fullscreen state.
+                        if new_monitor != fullscreen_monitor.as_ref().unwrap().inner.hmonitor()
```
chrisduerr, line 858: This just seems ripe for failure. I don't like the mapping things and then making assumptions later structure.

src/window.rs
```diff
@@ -624,6 +624,10 @@ impl Window {
 
     /// Gets the window's current fullscreen state.
     ///
+    /// If `None` in `Fullscreen::Borderless` was returned it means
+    /// that the window is in a `Fullscreen` state, however there's
+    /// no monitor information available.
```
chrisduerr, line 629: This seems a bit problematic, because we seem to conflate the concepts of "no monitor information available" and "current / primary monitor". Like a user doesn't request fullscreen on no monitor, that doesn't make sense.

src/window.rs
```diff
@@ -850,10 +854,15 @@ impl Default for CursorIcon {
     }
 }
 
+/// Fullscreen mode to be set on certain output.
+///
+/// Providing `None` to `Borderless` will result in fullscreen
+/// being set on a system preferred monitor. In general it'll be
+/// the current monitor.
```
chrisduerr, line 861: I'm not sure if I would say something about the "system preferred monitor" here, it could be easily confused with the primary monitor. Which is never used here afaics?

chrisduerr (changes requested):

CHANGELOG.md
kchibisov: indeed.

examples/window_debug.rs
kchibisov: Well, yeah, but that's purely for testing. Could be different on Wayland in certain cases though.

src/platform_impl/ios/view.rs
kchibisov: Yeah, it's really strange how it worked, but returning `msg_send!` wasn't working. I'm not sure that there's a CI log for that, but it was somewhere.

src/platform_impl/ios/view.rs
kchibisov: that's a pretty old rust iirc, so that's why it's using ref, I've just mirrored existing code.

src/platform_impl/ios/window.rs
kchibisov: It's basically that we're asking for a current monitor. However if you look at ios it doesn't really have a fullscreen in a way you'd know. It just always sets screen and it's always fullscreened, that branch just picks the 'preffered monitor', which is the current one here. Not sure whether such thing should be documented, since it's there for a long time.

src/window.rs
kchibisov: I mean what should I describe here then? It can clearly be returned with `None` on GNOME(Wayland) :P

src/window.rs
kchibisov: Yeah, we can reword it to be current monitor?

src/platform_impl/windows/event_loop.rs
kchibisov: Well, you can't store it with `None`, so you always store it unwraped, one more thing that you can't access monitor here.

src/platform_impl/linux/wayland/window.rs
kchibisov: The reason why there's no `map` its' because there's an `X` variant on `PlatformMonitorHandle` so I need `if let` anyway. And that `X` variant is presented based on feature, so in one case you must put `unreachable!` there, so I decided to you that, since shorter.

src/platform_impl/linux/wayland/window.rs
kchibisov: The reason why there's no `map` its' because there's an `X` variant on `PlatformMonitorHandle` so I need `if let` anyway. And that `X` variant is presented based on feature, so in one case you must put `unreachable!` there, so I decided to you that, since shorter.

src/platform_impl/linux/x11/window.rs
kchibisov: I'll just match it later on instead of expanding early on

src/platform_impl/windows/event_loop.rs
kchibisov: I'll ask for current monitor, since it should be available here.

kchibisov: Oh, nvm, it's acually important to store those things unwrapped and not in Option for fullscreen. So how we should approach that? I'm fine with `unwrap` here tbqh.

src/platform_impl/ios/window.rs
kchibisov: Yeah, it should work.

src/platform_impl/windows/window.rs
```diff
@@ -393,6 +393,17 @@ impl Window {
...
+        let fullscreen = match fullscreen {
+            Some(Fullscreen::Borderless(None)) => {
+                let current_monitor = Some(RootMonitorHandle {
+                    inner: monitor::current_monitor(window.0),
+                });
+
+                Some(Fullscreen::Borderless(current_monitor))
+            }
+            _ => fullscreen,
+        };
```
chrisduerr, line 405: Why can't this be stored with None as Borderless? Do you want to make it possible to go from fullscreen on one monitor to fullscreen on another using None as parameter? When would you ever want to do that?

chrisduerr (changes requested):

src/platform_impl/windows/window.rs
kchibisov: >Why can't this be stored with None as Borderless? Do you want to make it possible to go from fullscreen on one monitor to fullscreen on another using None as parameter? When would you ever want to do that?

it's due to monitor logic in event_loop.rs when you can move fullscreren window across the monitor, since I bet the current monitor after move is a new monitor, so you won't be able to update things properly. We may always update monitor there if the current one is `None` though.

src/platform_impl/linux/x11/window.rs
```diff
@@ -649,10 +649,11 @@ impl UnownedWindow {
...
                         video_mode: PlatformVideoMode::X(ref video_mode),
-                    }) => (Some(video_mode), video_mode.monitor.as_ref().unwrap()),
-                    Fullscreen::Borderless(RootMonitorHandle {
-                        inner: PlatformMonitorHandle::X(ref monitor),
-                    }) => (None, monitor),
+                    }) => (Some(video_mode), video_mode.monitor.clone().unwrap()),
+                    Fullscreen::Borderless(Some(RootMonitorHandle {
+                        inner: PlatformMonitorHandle::X(monitor),
+                    })) => (None, monitor),
+                    Fullscreen::Borderless(None) => (None, self.current_monitor()),
```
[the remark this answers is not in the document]
kchibisov: >I'd assume you changed this to a owned instead of a reference because Rust doesn't allow you to return &self.current_monitor() due to lifetime guarantees here?

Yes.

kchibisov:
@chrisduerr I've updated PR, let me know if it's fine now.

src/platform_impl/windows/event_loop.rs
```diff
@@ -853,8 +853,14 @@ unsafe extern "system" fn public_window_callback<T: 'static>(
...
-                            && new_monitor != ptr::null_mut()
+                        if new_monitor == ptr::null_mut() {
+                            return 0;
+                        }
+
+                        if fullscreen_monitor
+                            .as_ref()
+                            .map(|monitor| new_monitor != monitor.inner.hmonitor())
+                            .unwrap_or(true)
                         {
```
chrisduerr, line 864: Any reason why these aren't combined anymore? Seems strange to have one thing as a gate in the front and the other that does the same thing as a dedicated block.

src/window.rs
```diff
@@ -628,6 +628,8 @@ impl Window {
     ///
     /// - **iOS:** Can only be called on the main thread.
     /// - **Android:** Will always return `None`.
+    /// - **Wayland** Can return `Borderless(None)` in some scenarios
+    ///
```
chrisduerr, line 632:
```suggestion
    /// - **Wayland** Can return `Borderless(None)` in some scenarios.
    ///
```

I personally am not a huge fan of the "in some scenarios". Can we be more concrete here?

src/window.rs
```diff
@@ -850,10 +852,14 @@ impl Default for CursorIcon {
     }
 }
 
+/// Fullscreen mode to be set on certain output.
```
chrisduerr, line 855:
```suggestion
/// Fullscreen modes.
```

The certain output stuff isn't really true since it might not be a specific output and Exclusive takes a VideoMode instead?

src/window.rs
```diff
@@ -850,10 +852,14 @@ impl Default for CursorIcon {
...
+///
+/// Providing `None` to `Borderless` will result in fullscreen
+/// being set on a current monitor.
```
chrisduerr, line 858: Maybe this is better as a doc comment on the `Borderless` enum variant?

chrisduerr (changes requested):

src/platform_impl/linux/wayland/window.rs
```diff
@@ -154,11 +154,17 @@ impl Window {
...
-            #[cfg(feature = "x11")]
-            Some(Fullscreen::Borderless(_)) => unreachable!(),
+            Some(Fullscreen::Borderless(monitor)) => {
+                let monitor = monitor
+                    .map(|RootMonitorHandle { inner: monitor }| match monitor {
+                        PlatformMonitorHandle::Wayland(monitor) => Some(monitor.proxy),
+                        #[cfg(feature = "x11")]
+                        PlatformMonitorHandle::X(_) => None,
+                    })
+                    .flatten();
```
chrisduerr, line 164:
```suggestion
                let monitor = monitor
                    .and_then(|RootMonitorHandle { inner: monitor }| match monitor {
                        PlatformMonitorHandle::Wayland(monitor) => Some(monitor.proxy),
                        #[cfg(feature = "x11")]
                        PlatformMonitorHandle::X(_) => None,
                    });
```

src/platform_impl/linux/wayland/window.rs
```diff
@@ -348,16 +356,17 @@ impl Window {
...
-                    .unwrap()
-                    .set_fullscreen(Some(&monitor_id.proxy));
+            Some(Fullscreen::Borderless(monitor)) => {
+                let monitor = monitor
+                    .map(|RootMonitorHandle { inner: monitor }| match monitor {
+                        PlatformMonitorHandle::Wayland(monitor) => Some(monitor.proxy),
+                        #[cfg(feature = "x11")]
+                        PlatformMonitorHandle::X(_) => None,
+                    })
+                    .flatten();
```
chrisduerr, line 366:
```suggestion
                let monitor = monitor
                    .and_then(|RootMonitorHandle { inner: monitor }| match monitor {
                        PlatformMonitorHandle::Wayland(monitor) => Some(monitor.proxy),
                        #[cfg(feature = "x11")]
                        PlatformMonitorHandle::X(_) => None,
                    });
```

src/window.rs
```diff
@@ -628,6 +628,8 @@ impl Window {
     ///
     /// - **iOS:** Can only be called on the main thread.
     /// - **Android:** Will always return `None`.
+    /// - **Wayland:** Can return `Borderless(None)` when no monitors are presented.
```
chrisduerr, line 631: What do you mean by "no monitors are presented"? Like no monitors are *present*? Usually a monitor does the presenting and isn't presented, right?

src/window.rs
```diff
@@ -628,6 +628,8 @@ impl Window {
...
+    ///
```
chrisduerr, line 632:
```suggestion
```

src/window.rs
```diff
@@ -850,10 +852,14 @@ impl Default for CursorIcon {
...
 }
 
+/// Fullscreen modes.
 #[derive(Clone, Debug, PartialEq)]
 pub enum Fullscreen {
     Exclusive(VideoMode),
-    Borderless(MonitorHandle),
+
+    /// Providing `None` to `Borderless` will result in fullscreen
+    /// being set on a current monitor.
```
chrisduerr, line 861:
```suggestion
    /// Providing `None` to `Borderless` will fullscreen on the current monitor.
```

I think this is a bit better. Since "fullscreen is set on..." doesn't really sound right. And either way it should be "**the** current monitor".

chrisduerr (changes requested):

src/window.rs
kchibisov: It means that user powered off every monitor it had, however our window was fullscreened when the user done that, so it's fullscreened, but there are no outputs.

chrisduerr (approved):
LGTM

PR #1671, opened 2020-08-21
Make current_monitor return `Option<MonitorHandle>` 

### LINKED ISSUE #793: `get_current_monitor()` panics on Wayland.

kchibisov:
On certain platfroms window couldn't be on any monitor
resulting in failures of `current_monitor` function on such,
since they can't return any monitor at all.

Such issue was happening on Wayland, since the window
isn't on any monitor, unless the user has drawn something into it.

Returning `Option<MonitorHandle>` will give an ability to
handle such situations gracefully by properly indicating that
there's no current monitor.

Fixes #793.

If anyone wants to submit more specific platform implementations feel free to post a patch or open PR against my branch.

cc @murarth (X11)
cc @francesca64 (macOS/ios)
cc @Osspial (Windows)

### DIFF
--- a/src/platform_impl/ios/window.rs
+++ b/src/platform_impl/ios/window.rs
@@ -224,7 +224,7 @@ impl Inner {
 
     pub fn fullscreen(&self) -> Option<Fullscreen> {
         unsafe {
-            let monitor = self.current_monitor();
+            let monitor = self.current_monitor_inner();
             let uiscreen = monitor.inner.ui_screen();
             let screen_space_bounds = self.screen_frame();
             let screen_bounds: CGRect = msg_send![uiscreen, bounds];
@@ -258,13 +258,18 @@ impl Inner {
         warn!("`Window::set_ime_position` is ignored on iOS")
     }
 
-    pub fn current_monitor(&self) -> RootMonitorHandle {
+    // Allow directly accessing the current monitor internally without unwrapping.
+    fn current_monitor_inner(&self) -> RootMonitorHandle {
         unsafe {
             let uiscreen: id = msg_send![self.window, screen];
             RootMonitorHandle {
                 inner: MonitorHandle::retained_new(uiscreen),
             }
         }
+    }
+
+    pub fn current_monitor(&self) -> Option<RootMonitorHandle> {
+        Some(self.current_monitor_inner())
     }
 
     pub fn available_monitors(&self) -> VecDeque<MonitorHandle> {
--- a/src/platform_impl/linux/wayland/window.rs
+++ b/src/platform_impl/linux/wayland/window.rs
@@ -334,8 +334,10 @@ impl Window {
 
     pub fn fullscreen(&self) -> Option<Fullscreen> {
         if *(self.fullscreen.lock().unwrap()) {
+            // If the monitor cannot be determined, we cannot report any fullscreen mode.
+            let current_monitor = self.current_monitor()?;
             Some(Fullscreen::Borderless(RootMonitorHandle {
-                inner: PlatformMonitorHandle::Wayland(self.current_monitor()),
+                inner: PlatformMonitorHandle::Wayland(current_monitor),
             }))
         } else {
             None
@@ -396,12 +398,12 @@ impl Window {
         &self.surface
     }
 
-    pub fn current_monitor(&self) -> MonitorHandle {
-        let output = get_outputs(&self.surface).last().unwrap().clone();
-        MonitorHandle {
+    pub fn current_monitor(&self) -> Option<MonitorHandle> {
+        let output = get_outputs(&self.surface).last()?.clone();
+        Some(MonitorHandle {
             proxy: output,
             mgr: self.outputs.clone(),
-        }
+        })
     }
 
     pub fn available_monitors(&self) -> VecDeque<MonitorHandle> {
--- a/src/platform_impl/macos/window.rs
+++ b/src/platform_impl/macos/window.rs
@@ -972,7 +972,8 @@ impl UnownedWindow {
     }
 
     #[inline]
-    pub fn current_monitor(&self) -> RootMonitorHandle {
+    // Allow directly accessing the current monitor internally without unwrapping.
+    pub(crate) fn current_monitor_inner(&self) -> RootMonitorHandle {
         unsafe {
             let screen: id = msg_send![*self.ns_window, screen];
             let desc = NSScreen::deviceDescription(screen);
@@ -983,6 +984,11 @@ impl UnownedWindow {
                 inner: MonitorHandle::new(display_id),
             }
         }
+    }
+
+    #[inline]
+    pub fn current_monitor(&self) -> Option<RootMonitorHandle> {
+        Some(self.current_monitor_inner())
     }
 
     #[inline]
--- a/src/platform_impl/web/window.rs
+++ b/src/platform_impl/web/window.rs
@@ -201,7 +201,7 @@ impl Window {
     #[inline]
     pub fn fullscreen(&self) -> Option<Fullscreen> {
         if self.canvas.is_fullscreen() {
-            Some(Fullscreen::Borderless(self.current_monitor()))
+            Some(Fullscreen::Borderless(self.current_monitor_inner()))
         } else {
             None
         }
@@ -237,10 +237,16 @@ impl Window {
     }
 
     #[inline]
-    pub fn current_monitor(&self) -> RootMH {
+    // Allow directly accessing the current monitor internally without unwrapping.
+    fn current_monitor_inner(&self) -> RootMH {
         RootMH {
             inner: monitor::Handle,
         }
+    }
+
+    #[inline]
+    pub fn current_monitor(&self) -> Option<RootMH> {
+        Some(self.current_monitor_inner())
     }
 
     #[inline]
--- a/src/window.rs
+++ b/src/window.rs
@@ -736,13 +736,15 @@ impl Window {
 
 /// Monitor info functions.
 impl Window {
-    /// Returns the monitor on which the window currently resides
+    /// Returns the monitor on which the window currently resides.
+    ///
+    /// Returns `None` if current monitor can't be detected.
     ///
     /// ## Platform-specific
     ///
     /// **iOS:** Can only be called on the main thread.
     #[inline]
-    pub fn current_monitor(&self) -> MonitorHandle {
+    pub fn current_monitor(&self) -> Option<MonitorHandle> {
         self.window.current_monitor()
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,9 +25,10 @@
 - **Breaking:** On Web, `set_cursor_position` and `set_cursor_grab` will now always return an error.
 - **Breaking:** `PixelDelta` scroll events now return a `PhysicalPosition`.
 - On NetBSD, fixed crash due to incorrect detection of the main thread.
-- **Breaking:** The virtual key code `Subtract` has been renamed to `NumpadSubtract`
-- **Breaking:** On X11, `-` key is mapped to the `Minus` virtual key code, instead of `Subtract`
+- **Breaking:** The virtual key code `Subtract` has been renamed to `NumpadSubtract`.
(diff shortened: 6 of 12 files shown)


src/window.rs
```diff
@@ -736,13 +736,17 @@ impl Window {
 
 /// Monitor info functions.
 impl Window {
-    /// Returns the monitor on which the window currently resides
+    /// Returns the monitor on which the window currently resides.
+    ///
+    /// If `None` was returned it could mean that window isn't presented on any monitor.
+    /// However you must not treat it as it's minimized, obscured, or localted on some
+    /// virtual desktop.
```
chrisduerr, line 743: I'd personally rephrase this, since the wording is awkward and it's not particularly useful. Particularly the "**was** returned" and "**could** mean" seems very strange.

I'm also not sure what the note about not treating it as minimized/obscured/localted(?)/virtual desktop is about? Are you trying to say that people should still render to it?

kchibisov: >I'm also not sure what the note about not treating it as minimized/obscured/localted(?)/virtual desktop is about? Are you trying to say that people should still render to it?

I'm saying that users must not use that to detect that window is minimized, obscured, or something in their render code to make a decision whether to render or not.

chrisduerr: You mean because when people never render there would never be a monitor? Seems a bit backwards to phrase it like that.

kchibisov: No, I mean that they shouldn't do something like that.

```rust
loop {

  // Reading events. 

  if window.current_monitor.is_none() {
    continue;
  }

  render();
}
```

chrisduerr: Why not? Assuming that it has been rendered once before?

kchibisov: Well, if we're talking about Wayland a canonical way to do that is frame callbacks, since you might be asked to render your window when minimized. So purely confusing. I can remove this doc all together, saying that on Wayland it can return None.

chrisduerr: I think that would be better, yeah. I feel like this documentation was made in good faith, but will likely just lead to confusion.

kchibisov: I've updated the docs, let me know what you think.

src/platform_impl/ios/window.rs
```diff
@@ -224,7 +224,7 @@ impl Inner {
 
     pub fn fullscreen(&self) -> Option<Fullscreen> {
         unsafe {
-            let monitor = self.current_monitor();
+            let monitor = self.current_monitor().unwrap();
```
chrisduerr, line 227: This is mostly a sidenote, but with this way of using the current_monitor API internally we cannot ensure that the typesystem will warn us when things change in the future.

The alternative would be to have an internal `current_monitor` function that returns just `RootMonitorHandle` and then call that in the public `current_monitor` function. That way we can call the internal one here and avoid the `unwrap`.

src/platform_impl/linux/wayland/window.rs
```diff
@@ -335,7 +335,7 @@ impl Window {
     pub fn fullscreen(&self) -> Option<Fullscreen> {
         if *(self.fullscreen.lock().unwrap()) {
             Some(Fullscreen::Borderless(RootMonitorHandle {
-                inner: PlatformMonitorHandle::Wayland(self.current_monitor()),
+                inner: PlatformMonitorHandle::Wayland(self.current_monitor().unwrap()),
```
chrisduerr, line 338: This is Wayland, right? What if no monitor is connected? The point of making current_monitor an option is because it might be None on Wayland, right?

chrisduerr (changes requested):

src/platform_impl/linux/wayland/window.rs
kchibisov: If you're in a fullscreen you must be presented on some output, however the unwrap part should be solved here https://github.com/rust-windowing/winit/pull/1670.

If nothing is connected, the app is fullscreened, and user calls to that function it'll likely crash, but it would crash earlier as well. There's no way to solve  that unless I rework the API in #1670.

src/platform_impl/ios/window.rs
kchibisov: >This is mostly a sidenote, but with this way of using the current_monitor API internally we cannot ensure that the typesystem will warn us when things change in the future.

>The alternative would be to have an internal current_monitor function that returns just RootMonitorHandle and then call that in the public current_monitor function. That way we can call the internal one here and avoid the unwrap.

Do you mean to handle it via `cfg`'s in `src/window.rs`, I guess it could work?

src/platform_impl/linux/wayland/window.rs
chrisduerr: I mean there definitely is a way. It could just return None. That would actually be less code than unwrapping.

src/platform_impl/ios/window.rs
chrisduerr: No, I mean having an `inner_current_monitor` (or whatever you want to call it) and then have the body of `current_monitor` just be `Some(self.inner_current_monitor())`. So if you just need the monitor internally, you can call `inner_current_monitor` without having to unwrap anything.

src/platform_impl/linux/wayland/window.rs
kchibisov: yeah, but it'll be miss informing the users on a state, since you're in a fullscreen state actually, so I don't really like returning `None`.

src/platform_impl/ios/window.rs
kchibisov: Ah, gotcha, it makes some sense to do it that way.

src/platform_impl/linux/wayland/window.rs
chrisduerr: Internally we think that the user is in a fullscreen state, but Wayland is telling us that we're not. I think I'd rather trust Wayland here than potentially crashing? It just seems like it would be the desired behavior, even in case things do go wrong.

With stuff like this it's also always easy to have one end (like the fullscreen lock) messed up due to refactorings and then have another place suddenly crash. I don't really see the **advantage** of unwrapping.

kchibisov: >Internally we think that the user is in a fullscreen state, but Wayland is telling us that we're not. I think I'd rather trust Wayland here than potentially crashing? It just seems like it would be the desired behavior, even in case things do go wrong.

No, it's not like that.

Wayland says that you **are** fullscreened, but it doesn't know on which monitor, since none are presented. And you can actually ask it to un fullscreen yourself or request fullscreen by passing `0` or calling `unset_fullscreen` respectively. So it's not that we think internally, we're saying that we're in fullscreen for sure, since Wayland told us so, but we can't say where, since every out disappeared.

The `self.fullscreen` mutex is set from the Wayland event in a callback handler, not from any of user events. If you do `set_fullscreen` on Wayland you're not touching this variable, it'll be touched when compositor tell you that you're in fullscreen that.

So performing an unwrap here is correct, since you know that you're in fullscreen, but you can't tell where. Returning `None` will lie to a user that you're not fullscreened, when in reality you're fullscreened.

>I think I'd rather trust Wayland here than potentially crashing?

Yes, that's what I'm doing. Wayland tells me that I'm fullscreened, but it doesn't know where, since the surface left all outputs. And since I can't expose that I'm just fullscreened and I must provide some output I crash, since there's no way to expose that unless #1670. 

>With stuff like this it's also always easy to have one end (like the fullscreen lock) messed up due to refactorings and then have another place suddenly crash. I don't really see the advantage of unwrapping.

You can't mess up this lock since you can only read it in winit, the set is done in handling of informing that your window actually went fullscreen. If you mess up the set of that mutex, you're already doing things incorrectly.

chrisduerr: > So performing an unwrap here is correct, since you know that you're in fullscreen, but you can't tell where. 

How is performing an unwrap correct? It's never correct to crash.

> Yes, that's what I'm doing. Wayland tells me that I'm fullscreened, but it doesn't know where, since the surface left all outputs. And since I can't expose that I'm just fullscreened and I must provide some output I crash, since there's no way to expose that unless #1670.

Crashing is **not** a reasonable response to being unable to determine which output is currently in fullscreen. If that could ever happen, crashing is completely unacceptable.

kchibisov: And what should we do? lie to the users? postpone  PR? we're already crashing on master, so I don't see any issue with it for now, knowing that PR solving that, is already up(even in a WIP state though).

If you want to report incorrect value to a user, it's likely will be fine, but it's also wrong.

chrisduerr: "Lying" to the user is the much better alternative, yes. If we don't have a decent value to give to the user and the user cannot handle it either, it's much better to return an incorrect value that will allow the error to correct itself in the future rather than just crashing.

There's something that is more wrong than reporting an incorrect value and that's just straight-up crashing in a scenario where we have no good reason to crash.

src/platform_impl/linux/wayland/window.rs
```diff
@@ -334,8 +334,12 @@ impl Window {
 
     pub fn fullscreen(&self) -> Option<Fullscreen> {
         if *(self.fullscreen.lock().unwrap()) {
+            // We're reporting incorrect value to a user to prevent a crash. In general
+            // we can't do much about it unless we'd be able to indicate that we're in a
+            // fullscreen state, but we don't know on which output.
```
chrisduerr, line 339: This seems very verbose.

```suggestion
            // If the monitor cannot be determined, we cannot report any fullscreen mode.
```

Seems like this has the same amount of information, in significantly less words? I don't see much point in the second sentence of your documentation.

kchibisov: That's some sort of meta `TODO`, I'm fine with shortening it, since I won't forget to properly rewrite it once #1670 will be ready.

src/platform_impl/ios/window.rs
```diff
@@ -258,7 +258,7 @@ impl Inner {
         warn!("`Window::set_ime_position` is ignored on iOS")
     }
 
-    pub fn current_monitor(&self) -> RootMonitorHandle {
+    fn current_monitor_inner(&self) -> RootMonitorHandle {
```
chrisduerr, line 261:
```suggestion
    // Allow directly accessing the current monitor internally without unwrapping.
    fn current_monitor_inner(&self) -> RootMonitorHandle {
```

src/platform_impl/macos/window.rs
```diff
@@ -972,7 +972,7 @@ impl UnownedWindow {
     }
 
     #[inline]
-    pub fn current_monitor(&self) -> RootMonitorHandle {
+    pub(crate) fn current_monitor_inner(&self) -> RootMonitorHandle {
```
chrisduerr, line 975:
```suggestion
    // Allow directly accessing the current monitor internally without unwrapping.
    pub(crate) fn current_monitor_inner(&self) -> RootMonitorHandle {
```

src/platform_impl/web/window.rs
```diff
@@ -237,12 +237,17 @@ impl Window {
     }
 
     #[inline]
-    pub fn current_monitor(&self) -> RootMH {
+    fn current_monitor_inner(&self) -> RootMH {
```
chrisduerr, line 240:
```suggestion
    // Allow directly accessing the current monitor internally without unwrapping.
    fn current_monitor_inner(&self) -> RootMH {
```

chrisduerr (changes requested):
After these documentation suggestions, everything should be good as far as I'm concerned.

chrisduerr (approved):

PR #1672, opened 2020-08-22
Impl. mouse capturing on web target

### LINKED ISSUE #1660: Mouse/pointer not captured when dragging out of the canvas area on web target
When dragging from inside the canvas to outside, the `winit` window stops receiving mouse events.

Possible fixes:

- When `PointerEvent` is available, one can use [`canvas.setPointerCapture`](https://developer.mozilla.org/en-US/docs/Web/API/Element/setPointerCapture) on `pointerdown` to capture the pointer.
- When using the `MouseEvent` fallback, the `mouse*` events should be attached to the window and the event handling code has to manually track the mouse capturing state. There is an [implementation in `kiss3d`](https://github.com/sebcrozet/kiss3d/pull/207/files) ([`web-sys` version](https://github.com/sebcrozet/kiss3d/blob/765f49c73d428c871125cbaca940a0320dd291a3/src/window/webgl_canvas.rs#L143) that can serve as an example.

alvinhochun:

### Tasks:

- [x] Implement for `web-sys` using `PointerEvent`
- [x] Implement for `web-sys` using `MouseEvent` (Frankly it would be much simpler if `winit` just requires users to use a [`PointerEvent` polyfill](https://github.com/jquery/PEP) if they want to support browsers with only `MouseEvent`.)
- [x] Implement for `stdweb` using `PointerEvent`
- [ ] Implement for `stdweb` using `MouseEvent` (low priority, might be skipped since `MouseEvent` support isn't even implemented in the `stdweb` backend)

Closes #1660

I have also refactored the pointer/mouse handling code a bit for the `web-sys` target due to it getting quite messy, and it also helps when adding touch event handling later.

The code had been changed for both `stdweb` and `web-sys`, but I have only tested with `web-sys`. Though technically they both use the same web API so their behaviour should not be any different.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 - On Android, fix `ControlFlow::Poll` not polling the Android event queue.
 - On macOS, add `NSWindow.hasShadow` support.
 - On Web, fix vertical mouse wheel scrolling being inverted.
+- On Web, implement mouse capturing for click-dragging out of the canvas.
 - **Breaking:** On Web, `set_cursor_position` and `set_cursor_grab` will now always return an error.
 - **Breaking:** `PixelDelta` scroll events now return a `PhysicalPosition`.
 - On NetBSD, fixed crash due to incorrect detection of the main thread.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -97,6 +97,7 @@ version = "0.3.22"
 optional = true
 features = [
     'console',
+    "AddEventListenerOptions",
     'CssStyleDeclaration',
     'BeforeUnloadEvent',
     'Document',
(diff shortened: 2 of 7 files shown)


alvinhochun:
Hi @ryanisaacg can you please review this PR?

ryanisaacg:
Sure, I'll review it this weekend. (Btw, in the future feel free to use the Github review request tool. I don't think Github notifies me when a PR goes from draft -> ready for review.)

chrisduerr:
@ryanisaacg I don't think you can request reviews unless you're a member of rust-windowing.

ryanisaacg:
That is an odd design choice on the part of Github. Feel free to ping me for reviews instead, I suppose!

PR #1679, opened 2020-08-25
Add prefix byte for extended scancodes on Windows

vickles:
Currently on Windows only the low byte of scancodes is passed back through the `KeyboardInput{ scancode }` field. This causes scancode collisions between different keys. 

Some examples that I've run up against in my game project:

* Both HOME and NumPad 7 map to `0x47`. HOME should be `0xE0 47`, and NumPad 7 should be `0x47`.
* Both UP and NumPad 8 map to `0x48`. UP should be `0xE0 48`, and NumPad 8 should be `0x48`.

I've been referencing this Microsoft doc: [Keyboard Scan Code Specification](https://download.microsoft.com/download/1/6/1/161ba512-40e2-4cc9-843a-923143f3456c/scancode.doc). It seems to contains the full breakdown of how scancodes are defined on windows.

This change prepends the prefix byte to scancodes which have the extended bit set. It allows me to distinguish between keys properly by scancode, and aligns them with the codes in that spec doc.


src/platform_impl/windows/event.rs
```diff
@@ -388,8 +388,8 @@ pub fn process_key_params(
     wparam: WPARAM,
     lparam: LPARAM,
 ) -> Option<(ScanCode, Option<VirtualKeyCode>)> {
-    let scancode = ((lparam >> 16) & 0xff) as UINT;
     let extended = (lparam & 0x01000000) != 0;
+    let scancode = if extended { 0xE000 } else { 0x0000 } | ((lparam >> 16) & 0xff) as UINT;
     handle_extended_keys(wparam as _, scancode, extended)
```
chrisduerr, line 393: Seems like the method `handle_extended_keys` exists specifically to handle extended keys. So it might be appropriate to move this inside of it?

vickles: :+1: I've moved it in to `handle_extended_keys`

kchibisov:
@vickles just fyi, since you rebase that patch from time to time, we haven't forgot about it, but since neither @chrisduerr nor me maintaining Windows, we're not that fast to take actions when it comes to it. However if it'll take too long for Windows maintainer(s) to appear we'll likely go back to it ourselves, even though we don't use Windows in particular.

vickles:
No problem. I'm updating my branch both to make sure it stays mergeable here, and also because I'm using that branch on my fork as the source of winit in my project

src/platform_impl/windows/event.rs
```diff
@@ -343,6 +343,7 @@ pub fn handle_extended_keys(
     extended: bool,
 ) -> Option<(c_int, UINT)> {
     // Welcome to hell https://blog.molecular-matters.com/2011/09/05/properly-handling-keyboard-input/
+    scancode = if extended { 0xE000 } else { 0x0000 } | scancode;
```
msiglreith, line 346: This affects the PAUSE/SCROLL logic below as it doesn't filter the additional event caused by RAWINPUT:

PR:
```
DeviceEvent { device_id: DeviceId(DeviceId(65599)), event: Key(KeyboardInput { scancode: 57373, state: Released, virtual_keycode: Some(Pause), modifiers: (empty) }) }
WindowEvent { window_id: WindowId(WindowId(0x75065e)), event: KeyboardInput { device_id: DeviceId(DeviceId(0)), input: KeyboardInput { scancode: 69, state: Released, virtual_keycode: Some(Pause), modifiers: (empty) }, is_synthetic: false } }
DeviceEvent { device_id: DeviceId(DeviceId(65599)), event: Key(KeyboardInput { scancode: 69, state: Released, virtual_keycode: Some(Pause), modifiers: (empty) }) }
```

master branch:
```
WindowEvent { window_id: WindowId(WindowId(0xd06cc)), event: KeyboardInput { device_id:  DeviceId(DeviceId(0)), input: KeyboardInput { scancode: 69, state: Released, virtual_keycode: Some(Pause), modifiers: (empty) }, is_synthetic: false } }
DeviceEvent { device_id: DeviceId(DeviceId(65599)), event: Key(KeyboardInput { scancode: 69, state: Released, virtual_keycode: Some(Pause), modifiers: (empty) }) }
```

msiglreith (changes requested):

src/platform_impl/windows/event.rs
vickles: I've updated this logic. I tested master and found it was reporting the same scancode and virtual key for scancode and numlock. The updated version should handle them both as separate entities better. I also tested scroll lock, pause and numlock with a variety of modifiers and it seems to be looking happy now.

msiglreith (approved):

PR #1686, opened 2020-08-28
Unify Minus/Subtract virtual keycodes

chrisduerr:
On all platforms other than Linux/X11, the Subtract key was uniformly
used only for the numpad. To make this cross-platform compatible, the
`-` key will now map to `Minus` on X11 instead of `Subtract`.

Since people have been confused about the difference between `Minus` and
`Subtract` in the past, the `Subtract` key has also been renamed to
`NumpadSubtract`. This is a breaking change that might be annoying to
downstream since there's no direct improvement, but it should help new
users in the future. Alternatively this could just be documented, rather
than explicitly mentioning the numpad in the name.


chrisduerr (commented):
Note that I've only explicitly tested the new `Minus` binding on Linux/X11. I'm assuming that the other existing bindings are correct, based on the online documentation available for each platform.

CHANGELOG.md
```diff
@@ -22,6 +22,8 @@
 - **Breaking:** On Web, `set_cursor_position` and `set_cursor_grab` will now always return an error.
 - **Breaking:** `PixelDelta` scroll events now return a `PhysicalPosition`.
 - On NetBSD, fixed crash due to incorrect detection of the main thread.
+- **Breaking:** The virtual key code `Subtract` has been renamed to `NumpadSubtract`
+- **Breaking:** On Linux, `-` key is mapped to the `Minus` virtual key code, instead of `Subtract`
```
kchibisov, line 26: It should be `On X11`. There's no `linux` in particular when it comes to winit, and Wayland was fine.

kchibisov (changes requested):

kchibisov (approved):

PR #1687, opened 2020-08-28
Move `#[cfg(...)]` -based compile errors from build.rs to lib.rs

Waridley:
#[cfg(...)] attributes don't work as expected from build.rs -- they refer to the configuration of the *host* system which the build.rs script will be running on. If you attempt to cross-compile to a different target, the crate may attempt to compile when it shouldn't, or fail when it should work.

The other way to check this would be to read Cargo environment variables in build.rs, as described in https://github.com/rust-lang/cargo/issues/4302
However, since this crate does nothing else in build.rs anway, I figured it made sense to just move the existing configuration checks to lib.rs


ogoffart:
Note that the checks are already there in the platform_impl:

https://github.com/rust-windowing/winit/blob/658a9a4ea8b50c5279042aa2a23f27fc583bb878/src/platform_impl/linux/mod.rs#L9-L10

https://github.com/rust-windowing/winit/blob/658a9a4ea8b50c5279042aa2a23f27fc583bb878/src/platform_impl/web/mod.rs#L34-L35

Having them in platform_impl instead of lib.rs avoid repeating the #[cfg] for the platform.
I believe one location is src/ is enough and the check is not required in lib.rs.

for reference, I was asked to add a build.rs there: https://github.com/rust-windowing/winit/pull/1597#discussion_r439762817

Waridley:
Oh, true, I didn't notice those. Well, the current build.rs makes it impossible to build for WASM on Linux... So I guess either the build.rs just needs deleted entirely, or we have to go with the environment variable option... (Personally I think it's not worth it since the `compile_error!` macro works fine and the environment variables are poorly documented and seem to have even changed since the issue I linked. But I'm totally willing to get it working if that's what the maintainers want.)

.gitignore
```diff
@@ -2,6 +2,7 @@ Cargo.lock
 target/
 rls/
 .vscode/
+.idea/
```
kchibisov, line 5: I'm not sure why it should be added. In general, I don't like the entries in that file, so let's not add any extra.

For such things you have a global gitignore file, so please remove it from here.

build.rs
```diff
@@ -1,21 +0,0 @@
...
-compile_error!("at least one of the \"x11\"/\"wayland\" features must be enabled");
-
-#[cfg(all(
-    target_arch = "wasm32",
-    not(feature = "web-sys"),
-    not(feature = "stdweb")
-))]
-compile_error!("at least one of the \"web-sys\"/\"stdweb\" features must be enabled");
-
-fn main() {}
```
kchibisov, line 21: Is it correct  that it fails in particular in cross build from macOS/Windows to Linux/BSD? Everything winit is using coming from `dlopen`, so there shouldn't be any issues wrt those checks if you build from any os presented in those cfgs to other os presented in those cfgs.

kchibisov (changes requested):
>The other way to check this would be to read Cargo environment variables in build.rs, as described in rust-lang/cargo#4302
However, since this crate does nothing else in build.rs anway, I figured it made sense to just move the existing configuration checks to lib.rs

I'd rather keep anything build related in a build.rs, not in lib.rs somewhere, so if you can do that, please, go ahead.

build.rs
ogoffart: I believe the problem is compiling for WASM with  default-features=false.
The build script will be built with the linux target, but none of the x11 or wayland feature will be selected.

In fact, the build script should not fail at compile time, but it should fail at run time, and use the std::env::vars to get the current target.

kchibisov: Yeah, I agree. builds.rs should check a target we're compiling for and then throw an error. The main point is that I don't want to put those checks in lib.rs, since I'd like all such stuff be in a one place.

.gitignore
chrisduerr: Yeah I agree with that. The existing .vscode files are already kinda useless. People should maintain their global gitignores instead of having to add this to every single project.

Waridley: Yeah, I kinda forgot I had added that when making the PR. Only even did so because every other project I've ever contributed to had it, and `.vscode/` was already there. Also sometimes for for personal projects I don't ignore `.idea/`, but that's of questionable value anyway.

build.rs
Waridley: Oh, yeah, slight correction: it's impossible to build for WASM from Linux *iff* using `default-features = false` and not enabling x11 or wayland. So, it's possible to work around, but still needs fixed because this is both confusing to users and wastes compile time when compiling to WASM.

I'm noticing several places that `compile_error!()` is still used, and redundant `#cfg[]` checks in several places. Would you rather them all be removed in the same commit if I make sure `build.rs` provides the correct checks, or just fix build.rs first and worry about cleanup later?

murarth:
@Waridley What's the status of this PR? Is there something preventing this from being merged?

Waridley:
@murarth I was mostly hoping for a reply to my question in the above conversation:

> I'm noticing several places that `compile_error!()` is still used, and redundant `#cfg[]` checks in several places. Would you rather them all be removed in the same commit if I make sure `build.rs` provides the correct checks, or just fix build.rs first and worry about cleanup later?

But I think I will just worry about `build.rs` for now; it'd be a lot more work for me to go through all the code and remove the cfg checks from the code itself and make sure that doesn't break anything.

Waridley:
Oh, I've also not been able to get it to build with the `stdweb` feature... which is deprecated, but still not technically removed. But it fails on the current master as well, so it's not my changes. I'm not sure it's even a problem with winit at all, unless I'm missing something. At any rate, building with `web-sys` without X11/Wayland is now working, so I'll mark this as ready for review.

murarth:
This needs a changelog entry. Otherwise, I'm ready to approve it for merge.

build.rs
```diff
@@ -1,21 +1,33 @@
...
-        target_os = "freebsd",
-        target_os = "netbsd",
-        target_os = "openbsd"
-    ),
-    not(feature = "x11"),
-    not(feature = "wayland")
-))]
-compile_error!("at least one of the \"x11\"/\"wayland\" features must be enabled");
+fn main() {
+    use std::env::var;
```
kchibisov, line 2: Move that one line above main.

kchibisov (changes requested):
As it was mentioned it requires a changelog entry.

chrisduerr (commented):
I personally find the build.rs file for this rather suboptimal. This introduces a lot of additional points of failure and chances for misuse. The early returns used here for example could create big issues with future build.rs changes and the cargo environment variable checking is fragile (which you can see by the error handling necessary).

Generally the `build.rs` file is provided by cargo to run things that are prerequisites for the build process. While it makes sense to put things here to fail as early as possible, this isn't really a prerequisite for everything else. So putting it in a different place should work fine too.

So I think in general this would be simpler and easier to understand by moving it into lib.rs and thus it would hopefully be better for future maintenance.

kchibisov:
>So I think in general this would be simpler and easier to understand by moving it into lib.rs and thus it would hopefully be better for future maintenance.

I agree, that's actually looks pretty messy already.

murarth:
Taking @chrisduerr's critique into account, I'm requesting the following changes:

* The code should be refactored to remove early returns.
* `CARGO_CFG_TARGET_OS` is a [documented env var](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts), so its presence can be assumed. The value should be `.unwrap()`d rather than issuing a warning for something we expect never to occur.
* A short macro should be defined to clean up the `CARGO_FEATURE_<feature>` checks; e.g. `has_feature!("X11")`, expanding to the `var()` expression.

However, I disagree with this point:

> Generally the `build.rs` file is provided by cargo to run things that are prerequisites for the build process. While it makes sense to put things here to fail as early as possible, this isn't really a prerequisite for everything else. So putting it in a different place should work fine too.
> 
> So I think in general this would be simpler and easier to understand by moving it into lib.rs and thus it would hopefully be better for future maintenance.

These checks were moved into `build.rs` because rustc does not stop compilation when a `compile_error!` invocation is encountered. That is, if a user compiles without one of the necessary features, the clear and concise error message provided by `compile_error!` may be obscured by the many other errors generated about missing names. Putting the check in `build.rs` means the user sees only the error about the missing feature and can easily correct it.

ogoffart:
> These checks were moved into build.rs because rustc does not stop compilation when a compile_error! invocation is encountered. That is, if a user compiles without one of the necessary features, the clear and concise error message provided by compile_error! may be obscured by the many other errors generated about missing names. Putting the check in build.rs means the user sees only the error about the missing feature and can easily correct it.

There are other ways to solve this issue, For example putting all the module declarations in `lib.rs` in something like if-cfg statement to not compile any code if the features are not there

I personally think that the build script is a bit overkill as it adds compilation time (compile/link/run of the build script).

kchibisov:
>I personally think that the build script is a bit overkill as it adds compilation time (compile/link/run of the build script).

The main use case for those checks is downstream crates, if you end up shipping crate to crates-io that doesn't build you have way more serious issues than having winit failing to build, also when winit is dependency, it'll abort compilation earlier than you reach compiling of your own crate, so I don't see any issues with putting those checks in llib.rs.

We did something like that in alacritty to fix cross for example

So I'm still in favor of having them in lib.rs, since it's actually simpler and easier to maintain. The use case with 'a lot of error messages' is artificial, that can come up when you're working on winit itself, but I guess as a winit developer you're aware that you need some of this feature enabled, no?

chrisduerr:
> A short macro should be defined to clean up the CARGO_FEATURE_<feature> checks; e.g. has_feature!("X11"), expanding to the var() expression.

I'd like to point out that macros add significant complexity which I would find questionable for something that is fundamentally so simple. So I don't think hiding the complexity behind macros is a good idea.

murarth:
The `cfg-if` approach seems like a promising alternative to the `build.rs` solution. Are there any drawbacks to this?

chrisduerr:
Not sure what you mean by `cfg-if`? Are you talking about the crate?

ogoffart:
Something like this in lib.rs

```rust
cfg_if!{
    if #[cfg(all(any(unix, ...), not(any(feature="wayland", feature="x11"))] {
          compile_error!("you must enable x11 or waymand feature")
     } else if #[cfg(all(wasm..., not(any(feature="stdweb"), feature="websys"))]
          compile_error!("you must enable stdweb or websys")
     } else {
         // the current content of lib.rs
         mod foo;  mod bar; ...
     }
}
```

chrisduerr:
That seems just... worse? Why would you want to introduce additional complexity like that?

Waridley:
I think the first main benefit `cfg-if` provides is being able to use `else` instead of constantly repeating the same conditions but wrapped in `not()`

murarth:
I want to resolve this issue, but we're having trouble coming to a consensus on the fix. What I'm seeing is, there are actually two issues that we're trying to resolve with this PR. The main issue, that one can't build for WASM on Linux without providing one of the irrelevant X11/Wayland features; and the secondary issue, that a feature-less build will fail in a less-than-graceful manner.

So, here's what I propose:

* We set aside the secondary issue for now and revisit it at a later time, if it becomes a greater problem (e.g. if users submit issues about the potentially confusing error output).
* We address the main issue by removing `build.rs` entirely and making no other changes. The `compile_error!`s for Linux and Web will stay in their respective `platform_impl` modules.

Any PRs to improve the understandability of the code or the readability of compile error output can be submitted later and considered at that time.

murarth (approved):

kchibisov (approved):

PR #1694, opened 2020-09-04
Make `primary_monitor` return `Option<MonitorHandle>`

### LINKED ISSUE #1683: Make `primary_monitor` return Option
Wayland doesn't have a concept of primary output in particular, so returning some random monitor isn't something user may actually want. Returning `None` will eliminate a `crash` that we have now when there's no output available, and also will make it pair it nicely with https://github.com/rust-windowing/winit/pull/1670, to make request fullscreen on a compositor preferred monitor, which could be primary.

On Wayland `primary_monitor` function will always return `None`, on other platforms it continue to return whatever it's returning now, but wrapped in `Some`.

kchibisov:
Certain platforms like Wayland don't have a concept of
primary Monitor in particular. To indicate that
`primary_monitor` will return `None` as well as in cases
where the primary monitor can't be detected.

Fixes #1683.

### DIFF
--- a/examples/video_modes.rs
+++ b/examples/video_modes.rs
@@ -3,7 +3,13 @@ use winit::event_loop::EventLoop;
 fn main() {
     simple_logger::init().unwrap();
     let event_loop = EventLoop::new();
-    let monitor = event_loop.primary_monitor();
+    let monitor = match event_loop.primary_monitor() {
+        Some(monitor) => monitor,
+        None => {
+            println!("No primary monitor detected.");
+            return;
+        }
+    };
 
     println!("Listing available video modes:");
 
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -178,11 +178,15 @@ impl<T> EventLoopWindowTarget<T> {
     }
 
     /// Returns the primary monitor of the system.
+    ///
+    /// Returns `None` if it can't identify any monitor as a primary one.
+    ///
+    /// ## Platform-specific
+    ///
+    /// **Wayland:** Always returns `None`.
     #[inline]
-    pub fn primary_monitor(&self) -> MonitorHandle {
-        MonitorHandle {
-            inner: self.p.primary_monitor(),
-        }
+    pub fn primary_monitor(&self) -> Option<MonitorHandle> {
+        self.p.primary_monitor()
     }
 }
 
--- a/src/platform_impl/ios/event_loop.rs
+++ b/src/platform_impl/ios/event_loop.rs
@@ -13,6 +13,7 @@ use crate::{
     event_loop::{
         ControlFlow, EventLoopClosed, EventLoopWindowTarget as RootEventLoopWindowTarget,
     },
+    monitor::MonitorHandle as RootMonitorHandle,
     platform::ios::Idiom,
 };
 
@@ -56,9 +57,11 @@ impl<T: 'static> EventLoopWindowTarget<T> {
         unsafe { monitor::uiscreens() }
     }
 
-    pub fn primary_monitor(&self) -> MonitorHandle {
+    pub fn primary_monitor(&self) -> Option<RootMonitorHandle> {
         // guaranteed to be on main thread
-        unsafe { monitor::main_uiscreen() }
+        let monitor = unsafe { monitor::main_uiscreen() };
+
+        Some(RootMonitorHandle { inner: monitor })
     }
 }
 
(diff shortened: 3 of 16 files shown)


examples/video_modes.rs
```diff
@@ -3,7 +3,13 @@ use winit::event_loop::EventLoop;
...
     simple_logger::init().unwrap();
     let event_loop = EventLoop::new();
-    let monitor = event_loop.primary_monitor();
+    let monitor = match event_loop.primary_monitor() {
+        Some(monitor) => monitor,
+        None => {
+            println!("No primary monitor detected.");
+            return;
+        }
+    };
```
chrisduerr, line 12: Probably easier to just unwrap/expect here?

src/platform_impl/ios/event_loop.rs
```diff
@@ -56,9 +57,10 @@ impl<T: 'static> EventLoopWindowTarget<T> {
         unsafe { monitor::uiscreens() }
     }
 
-    pub fn primary_monitor(&self) -> MonitorHandle {
+    pub fn primary_monitor(&self) -> Option<RootMonitorHandle> {
         // guaranteed to be on main thread
-        unsafe { monitor::main_uiscreen() }
+        let inner = unsafe { monitor::main_uiscreen() };
+        Some(RootMonitorHandle { inner })
```
chrisduerr, line 63:
```suggestion
        let monitor = unsafe { monitor::main_uiscreen() };
        Some(RootMonitorHandle { inner: monitor })
```

I think this naming would be slightly better and prevent confusions with potential future refactorings? But it doesn't matter that much, especially since using `inner` makes the return shorter.

src/platform_impl/ios/window.rs
```diff
@@ -271,8 +271,9 @@ impl Inner {
         unsafe { monitor::uiscreens() }
     }
 
-    pub fn primary_monitor(&self) -> MonitorHandle {
-        unsafe { monitor::main_uiscreen() }
+    pub fn primary_monitor(&self) -> Option<RootMonitorHandle> {
+        let handle = unsafe { monitor::main_uiscreen() };
+        Some(RootMonitorHandle { inner: handle })
```
chrisduerr, line 276: Would probably make sense to at least be consistent with the event_loop.rs variable naming here.

src/platform_impl/linux/mod.rs
```diff
@@ -450,8 +450,18 @@ impl Window {
...
-    pub fn primary_monitor(&self) -> MonitorHandle {
-        x11_or_wayland!(match self; Window(window) => window.primary_monitor(); as MonitorHandle)
+    pub fn primary_monitor(&self) -> Option<RootMonitorHandle> {
+        match self {
+            #[cfg(feature = "x11")]
+            &Window::X(ref window) => {
+                let primary_monitor = window.primary_monitor();
+                Some(RootMonitorHandle {
+                    inner: MonitorHandle::X(primary_monitor),
+                })
```
chrisduerr, line 460:
```suggestion
                let primary_monitor = MonitorHandle::X(window.primary_monitor());
                Some(RootMonitorHandle { inner: primary_monitor })
```

Just a potential style improvement suggestion.

src/event_loop.rs
```diff
@@ -178,11 +178,11 @@ impl<T> EventLoopWindowTarget<T> {
     }
 
     /// Returns the primary monitor of the system.
+    ///
+    /// Returns `None` if it can't identify any monitor as a primary one.
```
chrisduerr, line 182: Shouldn't the documentation here match the one on Window?

chrisduerr (changes requested):

examples/video_modes.rs
kchibisov: Maybe, but you can run this on Wayland without any issue and got your `no primary monitor detected`, since this example doesn't require window in particular, I don't really like panic, since the intention is to show that it's **normal** that primary monitor isn't available.

src/event_loop.rs
kchibisov: Well, for Wayland bits - yes, but for `ios` there's no need to, since it's always called on the main thread due to event loop limitations, since it's always on a mean thread already.

examples/video_modes.rs
chrisduerr: Sure, certainly not unreasonable.

chrisduerr (approved):

Issue #1695, opened 2020-09-04
Horizontal scrolling backwards on mac


michaelkirk:
My application has vertical scrolling working well, but horizontal scrolling is backwards for macos. I'm not 100% sure that this is a bug winit, but here's what I know:

There are two kinds of MouseScrollDelta: LineDelta and PixelDelta. Which one is used depends on some combination of the platform and device.

For vertical scrolling, the sign of macos's PixelDelta matches the sign of wayland and x11's LineDelta.
For horizontal scrolling, the sign of macos's PixelDelta is **opposite** the sign of wayland and x11's LineDelta.

More specifically:

macos (uses PixelDelta):
  negative: swipe up, **swipe left** (scrollbar moves down, right, respectively - aka "natural scrolling")
  positive: swipe down, swipe right (scrollbar moves up, left, respectively - aka "natural scrolling")

linux-x11 (via Parallels VM on same mac) (uses LineDelta):
  negative: swipe up, **swipe right**
  positive: swipe down swipe left 
                                                        
linux-wayland (via Parallels VM on same mac) (uses LineDelta): 
  negative: swipe up, swipe right
  positive: swipe down swipe left

It seems like two possibilities:

1. This is working as intended, and it's just a discrepancy between PixelDelta vs LineDelta in general. We should clarify in the docs.

-or-

2. This is a mac specific bug in winit and we need to `* -1.0` the horizontal PixelDelta to be consistent with other platforms.

michaelkirk:
One thing which would help me debug, would be if someone running Wayland with PixelOffset scrolling (I think most modern trackpads will use PixelOffset on Linux?), could try my debug branch here: https://github.com/michaelkirk/winit/tree/mkirk/h-scroll

**Steps to help debug**:

```
git clone https://github.com/michaelkirk/winit.git
cd winit
git checkout mkirk/h-scroll
cargo build --example window_debug
cargo run --example window_debug 2> stderr_out
```

**Then**:

1. move your mouse over the window that launched
2. click the window
3. swipe **up** on your trackpad a few times (being careful not to swipe in other directions)
4. swipe **left** on your track pad a few times  (being careful not to swipe in other directions)
5. close the window and attach the output from stderr_out to this issue

![scroll_debug mov](https://user-images.githubusercontent.com/217057/92280680-d2686a80-eeae-11ea-8a98-923dc067e50f.gif)

kchibisov:
[...]

PR #1696, opened 2020-09-04
invert h-scroll direction for macos to be consistent with other platforms.

michaelkirk:
Fixes https://github.com/rust-windowing/winit/issues/1695

---

- [none] Compilation warnings were addressed
- [-] Created or updated an example program if it would help users understand this functionality
- [n/a] Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 - On Windows, drag and drop is now optional and must be enabled with `WindowBuilderExtWindows::with_drag_and_drop(true)`.
 - On Wayland, fix deadlock when calling to `set_inner_size` from a callback.
 - On macOS, add `hide__other_applications` to `EventLoopWindowTarget` via existing `EventLoopWindowTargetExtMacOS` trait. `hide_other_applications` will hide other applications by calling `-[NSApplication hideOtherApplications: nil]`.
-- On macOS - fix inverted horizontal scroll
+- On macOS, fix inverted horizontal scroll.
 - On android added support for `run_return`.
 - On MacOS, Fixed fullscreen and dialog support for `run_return`.
 - On Windows, fix bug where we'd try to emit `MainEventsCleared` events during nested win32 event loops.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - On Windows, drag and drop is now optional and must be enabled with `WindowBuilderExtWindows::with_drag_and_drop(true)`.
 - On Wayland, fix deadlock when calling to `set_inner_size` from a callback.
 - On macOS, add `hide__other_applications` to `EventLoopWindowTarget` via existing `EventLoopWindowTargetExtMacOS` trait. `hide_other_applications` will hide other applications by calling `-[NSApplication hideOtherApplications: nil]`.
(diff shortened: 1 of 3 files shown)


src/platform_impl/macos/view.rs
```diff
@@ -1003,7 +1003,8 @@ extern "C" fn scroll_wheel(this: &Object, _sel: Sel, event: id) {
         let state = &mut *(state_ptr as *mut ViewState);
 
         let delta = {
-            let (x, y) = (event.scrollingDeltaX(), event.scrollingDeltaY());
+            // macos horizontal sign convention is the inverse of winit
```
kchibisov, line 1006:
```suggestion
            // macOS horizontal sign convention is the inverse of winit.
```

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 - On Windows, drag and drop is now optional and must be enabled with `WindowBuilderExtWindows::with_drag_and_drop(true)`.
 - On Wayland, fix deadlock when calling to `set_inner_size` from a callback.
 - On macOS, add `hide__other_applications` to `EventLoopWindowTarget` via existing `EventLoopWindowTargetExtMacOS` trait. `hide_other_applications` will hide other applications by calling `-[NSApplication hideOtherApplications: nil]`.
+- On macOS - fix inverted horizontal scroll
```
kchibisov, line 12: Could you 'push back' the new entry instead of inserting in the middle of a section?

Also reformat it to match other entries here.

```suggestion
- On macOS, fix inverted horizontal scroll.
```

src/event.rs
```diff
@@ -748,22 +748,30 @@ pub enum MouseButton {
     Other(u8),
 }
 
-/// Describes a difference in the mouse scroll wheel state.
+/// A measured distance as the user scrolls.
+///
+/// Many modern compositors will express touchpad drags and tablet surface touches in units of
```
kchibisov, line 753: could we discard tablets information from here? winit doesn't handle those. I'm not sure about the word compositors here, I'd call it systems, or something like that, since some systems like X11 doesn't have. I'd also remove 'modern' part.

In general, I'd just wrote what devices could produce those events, but I feel like it should be above enum variants, not the enum itself, so basically like it was before, so you'll have a cause of event attached to event.

michaelkirk:
Hmmm... CI failure seems spurious:

```
error: could not download file from 'https://static.rust-lang.org/rustup/release-stable.toml' to 'C:\Users\RUNNER~1\AppData\Local\Temp\rustup-updateMIpKpw\release-stable.toml'
error: caused by: failed to make network request
```

kchibisov (changes requested):

src/event.rs
michaelkirk: I'm not sure I understand some of your suggestions. Since nothing in my PR behaviorally changes how these enums should be used, I'll just revert the doc changes.

michaelkirk:
> Could you 'push back' the new entry instead of inserting in the middle of a section?

Done! Could you please take another look @kchibisov?

kchibisov (approved):

PR #1703, opened 2020-09-11
Fix view frame in portrait when starting iOS app in landscape

MichaelHills:
- ~[ ] Created or updated an example program if it would help users understand this functionality~
- ~[ ] Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented~

I tested this in simulator + my iPhone 6 using Bevy https://github.com/bevyengine/bevy/issues/87

Might be nice to have at least 1 other person test this as well? Directly on winit rather than on bevy.

This is what the problem looks like. If you start up in portrait mode then everything is fine, auto-rotations all work well. If you start up in landscope mode, its permanently broken even if you rotate.

image

image


src/platform_impl/ios/view.rs
```diff
@@ -134,6 +134,18 @@ unsafe fn get_view_class(root_view_class: &'static Class) -> &'static Class {
                     height: screen_frame.size.height as f64,
                 }
                 .to_physical(scale_factor.into());
+
+                // If the app is started in landscape, the view frame and window bounds can be mismatched.
+                // The view frame will be in portrait and the window bounds in landscape. So apply the
+                // window bounds to the view frame to make it consistent.
```
francesca64, line 140: Thanks for adding an explanation! I always appreciate these a lot.

src/platform_impl/ios/view.rs
```diff
@@ -134,6 +134,18 @@ unsafe fn get_view_class(root_view_class: &'static Class) -> &'static Class {
...
+                let view_frame: CGRect = msg_send![object, frame];
+                let window_bounds: CGRect = msg_send![window, bounds];
+                if view_frame.size.width != window_bounds.size.width
+                    || view_frame.size.height != window_bounds.size.height
```
francesca64, line 144: There shouldn't be a problem with deriving `PartialEq` for `CGSize`, since we define it ourselves in the `ffi` module.

src/platform_impl/ios/view.rs
```diff
@@ -134,6 +134,18 @@ unsafe fn get_view_class(root_view_class: &'static Class) -> &'static Class {
...
+                {
+                    let () = msg_send!(object, setFrame: window_bounds);
```
francesca64, line 146:
```suggestion
                    let () = msg_send![object, setFrame:window_bounds];
```
I admit that parentheses look cuter than square brackets, but square brackets are the convention we've been using for this macro.

(The use of spaces after colons in `msg_send` has creeped into the codebase, but the Apple documentation doesn't use a space after the colon, so I'm enforcing that.)

francesca64 (changes requested):
Thanks for the PR, and for giving some love to iOS support!

I maintain winit on mobile; I don't watch this repo (winit generates a ton of notifications), but I'll always come if pinged!

I'll test this out on an iPhone XR and report back.

src/platform_impl/ios/view.rs
MichaelHills: Added `PartialEq` and just compared size. To be honest I don't know if I need to compare the `origin` as well. I haven't actively coded iOS for quite some number of years now. At least in the case of `winit` I think it doesn't really matter, we just have everything at `(0, 0)` anyway right?

src/platform_impl/ios/view.rs
MichaelHills: I didn't even notice I had it different, fixed. 👍

src/platform_impl/ios/view.rs
```diff
@@ -123,17 +123,26 @@ unsafe fn get_view_class(root_view_class: &'static Class) -> &'static Class {
 
                 let window: id = msg_send![object, window];
                 assert!(!window.is_null());
-                let bounds: CGRect = msg_send![window, bounds];
+                let window_bounds: CGRect = msg_send![window, bounds];
                 let screen: id = msg_send![window, screen];
                 let screen_space: id = msg_send![screen, coordinateSpace];
                 let screen_frame: CGRect =
-                    msg_send![object, convertRect:bounds toCoordinateSpace:screen_space];
+                    msg_send![object, convertRect:window_bounds.clone() toCoordinateSpace:screen_space];
```
MichaelHills, line 130: Since we already have the window bounds in scope, I decided to just clone here to avoid the borrow and re-use it down below. I'm new to using rust so still figuring my way around the borrow checker.

MichaelHills:
@francesca64 I have been testing via Bevy, but I think it would be better if I had a pure winit project setup for testing. That way I could this PR and any other issues I'm having. I looked over the winit examples but couldn't really find a drop-in sample for iOS. I tried `window.rs` but it's just a black screen which wouldn't really reproduce the issue I'm seeing in Bevy that this PR is about.

Can you recommend any of the examples or do you happen to have a suitable winit example for iOS? Just something like displaying an image would be enough. This would also be useful for me to try to produce a pure winit crash repro for https://github.com/rust-windowing/winit/issues/1705

src/platform_impl/ios/view.rs
francesca64: It doesn't look like anything guarantees a zero origin (see Discussion section on [`bounds`](https://developer.apple.com/documentation/uikit/uiview/1622580-bounds?language=objc) & [`frame`](https://developer.apple.com/documentation/uikit/uiview/1622621-frame?language=objc)), so we should probably just derive `PartialEq` for `CGRect` and compare the rects.

(I don't know off the top of my head if our particular usage leads to the origin always being zero, but that could end up changing anyway, so comparing the rects is the safest bet.)

src/platform_impl/ios/view.rs
francesca64: `CGRect` should probably just derive `Copy`, since it's just 4 floats.

francesca64:
@MichaelHills I actually have something super cool for that exact purpose, but I probably won't be able to get it to you until tomorrow.

src/platform_impl/ios/view.rs
MichaelHills: I've added `PartialEq` to `CGPoint` and `CGRect` and now compare the rects.

src/platform_impl/ios/view.rs
MichaelHills: Thanks updated. I'm new to rust so when I saw `Clone` on `CGRect` I thought that meant `Copy` and was wondering why I still needed to clone to avoid the borrow. Now I know the difference between `Copy` vs `Clone`. 👍

MichaelHills:
No worries take your time. I think the code is good now but I'll feel better if I can repro this PR directly on `winit`. :)

src/platform_impl/ios/view.rs
MichaelHills: I've actually changed this back to have the space because the PR build failed. I ran `cargo fmt` and re-pushed.

src/platform_impl/ios/view.rs
MichaelHills: Just realised I've said I'm new to rust twice in the same comment thread. Did I mention that I'm n...

francesca64:
So, I made this thing called [cargo-mobile](https://github.com/BrainiumLLC/cargo-mobile), which you can use to generate a minimal winit program that you can run on device. Hopefully the instructions in the README are easy to follow, but I'll answer any questions.

If you want something closer to "something like displaying an image", the `wgpu` template pack will generate a minimal wgpu program that renders a triangle... but swapping out the winit version being used might be cumbersome, since it's using an older (forked!) version of wgpu. So, it's potentially easier to just work off of the `winit` template pack, but I'll also be updating the `wgpu` template pack next week.

MichaelHills:
Nice! This is really cool. I managed to get it building the winit and wgpu examples but had to do the following:

- `cargo mobile init` (crashes because I don't have android set up)
- `cargo apple init` (to finish the apple setup)

The reason I had to run `cargo mobile init` first is because it doesn't copy the template code over if you just run `cargo apple init` by itself and `xcodegen` complains about `../../src` not existing.

The examples run on macOS, but unfortunately it crashes on my phone.

```
(lldb) bt
* thread #1, stop reason = signal SIGABRT
  * frame #0: 0x00000001009e2418 dyld`__abort_with_payload + 8
    frame #1: 0x00000001009e1a58 dyld`abort_with_payload_wrapper_internal + 100
    frame #2: 0x00000001009e1a88 dyld`abort_with_payload + 12
    frame #3: 0x00000001009a48f8 dyld`dyld::halt(char const*) + 304
    frame #4: 0x00000001009a76e4 dyld`dyld::_main(macho_header const*, unsigned long, int, char const**, char const**, char const**, unsigned long*) + 5396
    frame #5: 0x00000001009a1044 dyld`_dyld_start + 68
```

MichaelHills:
I ended up running the wgpu triangle example in my manual xcode project that I was using for bevy, and re-produced the issue and confirmed the fix. 👍 

image
image

MichaelHills:
Oh and for `cargo-mobile`, I think it would be great to be able to deploy to the simulator where the dev cycle is a little bit faster.

francesca64:
Thanks for trying it out! I'll fix all those problems soon (we're pushing to officially release it in about a month). With regards to the crash, are you using Rust 1.46.0?

francesca64 (approved):
I tested this on my iPhone XR and everything looks good! Thanks so much.

MichaelHills:
I'm on rust 1.48.0 nightly. I could try rolling back to 1.45.2 to see if it resolves the issue?

francesca64:
That'd be a big help; I was hitting what appears to be the same issue when I was on 1.46.0, and downgrading to 1.45.2 fixed it for me. You'll also need to `cargo clean` first, unfortunately... having a second person experience this would be very helpful for when I inevitably have to report this upstream.

MichaelHills:
I updated cargo mobile then switched back to 1.45.2, winit and wgpu templates run fine. Back to 1.48.0 nightly, get the crash again on any of the xamples but the bevy example works in the simulator.

Also while building for aarch64-apple-ios I get a huge amount of spam of this
```'+cyclone' is not a recognized feature for this target (ignoring feature)```

**Comment by francesca64** [comment]
Thanks for trying that! It sounds like I can bisect this more by comparing our build processes.

The spam was an upstream problem, which fortunately doesn't seem to impact the build at all: https://github.com/rust-lang/rust/issues/74427#issuecomment-666752947

**Comment by MichaelHills** [comment]
When I link the static lib `.a` to my bevy Xcode project it runs fine. I brought up Xcode and tried to run the cargo mobile project and got this:

```
dyld: Library not loaded: /Users/mikeh/repos/winit-ios12/target/aarch64-apple-ios/debug/deps/libwinit_ios12.dylib
  Referenced from: /var/containers/Bundle/Application/CC19CF5E-FFA7-4038-AFC2-5A71A52221FD/winit-ios12.app/winit-ios12
  Reason: image not found
```

This seems to line up with the backtrace I posted earlier referencing `dyld`. I don't know why it's trying to use a dylib. Anyway hopefully that's enough information to help you narrow it down.

PR #1707, opened 2020-09-17
Fix examples build by bumping simple_logger to v1.9

alvinhochun:
Fix regression from #1702. `simple_logger::SimpleLogger` is not public in `simple-logger` < v1.9.


kchibisov (approved):
It’s not a regression, since libs don’t have lock file and your lock file were preventing tests to succeed, but it’s good to explicitly require that change. ( you may notice that ci was green there )

PR #1712, opened 2020-09-19
Deprecate the stdweb backend

ryanisaacg:


src/platform_impl/web/mod.rs
```diff
@@ -29,6 +29,10 @@ mod backend;
 
 #[cfg(feature = "stdweb")]
 #[path = "stdweb/mod.rs"]
+#[cfg_attr(
+    feature = "stdweb",
+    deprecated(since = "0.23.0", note = "Please migrate to web-sys over stdweb")
+)]
```
chrisduerr, line 35: Wouldn't it be easier to put this at the top of the backend module?

ryanisaacg: 🤷, I'm not sure it changes much either way.

chrisduerr: You wouldn't need the feature check? Just seems cleaner to me.

ryanisaacg: Good point, thanks 👍

chrisduerr (approved):

PR #1713, opened 2020-09-19
Prefix numpad virtual key codes with Numpad

kchibisov:
This commit is a follow up to a2db4c0a320aafc10d240c432fe5ef4e4d84629d
to make it clear which virtual key codes are located on numpad.


CHANGELOG.md
```diff
@@ -25,14 +25,14 @@
...
 - On NetBSD, fixed crash due to incorrect detection of the main thread.
-- **Breaking:** The virtual key code `Subtract` has been renamed to `NumpadSubtract`.
 - **Breaking:** On X11, `-` key is mapped to the `Minus` virtual key code, instead of `Subtract`.
 - On macOS, fix inverted horizontal scroll.
 - **Breaking:** `current_monitor` now returns `Option<MonitorHandle>`.
 - **Breaking:** `primary_monitor` now returns `Option<MonitorHandle>`.
 - On macOS, updated core-* dependencies and cocoa.
 - Bump `parking_lot` to 0.11
 - On Android, bump `ndk`, `ndk-sys` and `ndk-glue` to 0.2. Checkout the new ndk-glue main proc attribute.
+- **Breaking:** The virtual key code `Add`, `Multiply`, `Divide`, `Decimal`, and `Subtract` now prefixed with `Numpad`.
```
kchibisov, line 35: I've decided to squash those into one entry, since it was a lot of text.

src/platform_impl/linux/wayland/keyboard.rs
```diff
@@ -305,7 +305,7 @@ fn keysym_to_vkey(keysym: u32) -> Option<VirtualKeyCode> {
         // misc
         // => Some(VirtualKeyCode::AbntC1),
         // => Some(VirtualKeyCode::AbntC2),
-        keysyms::XKB_KEY_plus => Some(VirtualKeyCode::Add),
+        // keysyms::XKB_KEY_plus => Some(VirtualKeyCode::Add),
```
chrisduerr, line 308: Why remove this? People might be relying on this, right? If anything, we should add another virtual keycode for these.

chrisduerr (changes requested):

src/platform_impl/linux/wayland/keyboard.rs
kchibisov: Well, the input model in winit is unclear, and you can't add those on other platforms anyway(except X11), so I've decided to remove it to avoid further confusion.

Don't get me wrong, I've tried adding new codes initially, but it turned out that those will be only on X11/Wayland, so I've reverted it.

chrisduerr: Is having keycodes for just X11/Wayland a problem? That seems better than removing it completely.

kchibisov: Kind of, since you're relying on those being accessed via shift + equals right now, so if you add special keys for those you change that implication, meaning that macOS/Windows/web not supporting those is a bug.

I can add those again that's not a big deal.

chrisduerr: Just doesn't sound great to remove features like that. Especially since people might be relying on it.

kchibisov: Ok, I've added those.

CHANGELOG.md
chrisduerr: This sentence doesn't make much sense, there seems to be missing a verb. 

```suggestion
- **Breaking:** Prefixed virtual key codes `Add`, `Multiply`, `Divide`, `Decimal`, and `Subtract` with `Numpad`.
```

chrisduerr (changes requested):

chrisduerr (approved):

PR #1715, opened 2020-09-20
Add cleanup code to web backend, mostly web-sys

### LINKED ISSUE #1688: Web backend does not handle ControlFlow::Exit nicely
1. Bug: The code in `runner::Shared::handle_event` does not properly enforce that `ControlFlow::Exit` is sticky. It only checks for `self.is_closed()` and does not check the actual user-assigned `control` out parameter. (I'm fixing this in a PR, just including this here for completeness.)
2. After the event loop has ended, the canvas is not removed from the DOM or blanked and the event handlers are not removed.

alvinhochun:

(copied from changelog:)

- On Web (web-sys only), the `Event::LoopDestroyed` event is correctly emitted when leaving the page.
- On Web ~~(web-sys only)~~, the `WindowEvent::Destroyed` event now gets emitted when a `Window` is dropped.
- On Web (web-sys only), the event listeners are now removed when a `Window` is dropped or when the event loop is destroyed.
- On Web ~~(web-sys only)~~, the event handler closure passed to `EventLoop::run` now gets dropped after the event loop is destroyed.
- **Breaking:** On Web, the canvas element associated to a `Window` is no longer removed from the DOM when the `Window` is dropped.

Closes #1688

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,11 @@
 - On Android, bump `ndk`, `ndk-sys` and `ndk-glue` to 0.2. Checkout the new ndk-glue main proc attribute.
 - **Breaking:** Prefixed virtual key codes `Add`, `Multiply`, `Divide`, `Decimal`, and `Subtract` with `Numpad`.
 - Added `Asterisk` and `Plus` virtual key codes.
+- On Web (web-sys only), the `Event::LoopDestroyed` event is correctly emitted when leaving the page.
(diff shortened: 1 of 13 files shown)


alvinhochun:
cc @ryanisaacg

ryanisaacg (approved):
Looks good, though I think the compile errors mean you have to stub out some methods for stdweb.

alvinhochun:
I even got the changelog wrong somehow. Now fixed.

Issue #1718, opened 2020-09-22
Moved event should convert into `PhysicalLocation` on Mac?


kaiwk:
If I understand properly, we should convert position in physical as `emit_resize_event` does?

https://github.com/rust-windowing/winit/blob/71e3d254226b47d8cbdb1d9432f11e4cdc659a46/src/platform_impl/macos/window_delegate.rs#L100-L117

chrisduerr:
I'm not quite sure what you're asking here. The `Moved` event already emits a `PhysicalPosition`. Are you requesting some kind of change, or looking for support?

kchibisov:
I feel like the issue that there's no conversion, if so please send patch.

kaiwk:
> I feel like the issue that there's no conversion

that's what i mean, `(x, y).into()` use `x` and `y` directly as physical unit, but seems they are in logical actually. I'm not quite sure though, the patch did fix my problem.

francesca64:
Closing, since this was fixed in #1722

PR #1722, opened 2020-09-23
Convert location to physical on Mac when `emit_move_event`

kaiwk:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,7 @@
 - **Breaking:** On Web, the canvas element associated to a `Window` is no longer removed from the DOM when the `Window` is dropped.
 - On Web, `WindowEvent::Resized` is now emitted when `Window::set_inner_size` is called.
 - **Breaking:** `Fullscreen` enum now uses `Borderless(Option<MonitorHandle>)` instead of `Borderless(MonitorHandle)` to allow picking the current monitor.
-- On MacOS, convert location to physical unit in `emit_move_event`.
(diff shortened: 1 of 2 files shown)


kaiwk:
see #1718

chrisduerr (changes requested):
Could you add a changelog entry? Other than that it's good to go.

CHANGELOG.md
```diff
@@ -42,6 +42,7 @@
 - **Breaking:** On Web, the canvas element associated to a `Window` is no longer removed from the DOM when the `Window` is dropped.
 - On Web, `WindowEvent::Resized` is now emitted when `Window::set_inner_size` is called.
 - **Breaking:** `Fullscreen` enum now uses `Borderless(Option<MonitorHandle>)` instead of `Borderless(MonitorHandle)` to allow picking the current monitor.
+- On MacOS, convert location to physical unit in `emit_move_event`.
```
kchibisov, line 45:
```suggestion
- On macOS, fix `WindowEvent::Moved` coordinates when scale factor isn't 1.
```

kchibisov (changes requested):

CHANGELOG.md
chrisduerr: I think it's worth it to describe the error, rather than just saying they're fixed when the scale factor isn't 1.

```suggestion
- On MacOS, fix `WindowEvent::Moved` ignoring the scale factor.
```

chrisduerr (approved):

PR #1724, opened 2020-09-23
Fix missing ModifiersChanged event from start

### LINKED ISSUE #1563: Missing modifiers changed event on X11
It seems like winit sometimes fails to send modifiers changed event on X11 during startup, you can see that in this logs, you release `Alt`, but you don't get `ModifiersChanged` event at all. I've seen this issue a bunch of times, with interaction with dmenu(not on my system though) afaics, however I don't have logs for it.

[alacritty.log](https://github.com/rust-windowing/winit/files/4605459/alacritty.log)

nixpulvis:
This patch updates the X11 `FocusIn` and `FocusOut` event processor to properly track the synthetic events produced.

fixes #1563

### DIFF
--- a/src/platform_impl/linux/x11/event_processor.rs
+++ b/src/platform_impl/linux/x11/event_processor.rs
@@ -923,7 +923,7 @@ impl<T: 'static> EventProcessor<T> {
                                 },
                             });
 
-                            // Issue key press events for all pressed keys.
+                            // Issue key press events for all pressed keys
                             Self::handle_pressed_keys(
                                 &wt,
                                 window_id,
--- a/src/platform_impl/linux/x11/event_processor.rs
+++ b/src/platform_impl/linux/x11/event_processor.rs
@@ -1243,7 +1243,12 @@ impl<T: 'static> EventProcessor<T> {
         let modifiers = device_mod_state.modifiers();
 
         // Update modifiers state and emit key events based on which keys are currently pressed.
-        for keycode in wt.xconn.query_keymap().into_iter().filter(|k| *k >= KEYCODE_OFFSET) {
+        for keycode in wt
+            .xconn
+            .query_keymap()
+            .into_iter()
+            .filter(|k| *k >= KEYCODE_OFFSET)
+        {
             let scancode = (keycode - KEYCODE_OFFSET) as u32;
             let keysym = wt.xconn.keycode_to_keysym(keycode);
             let virtual_keycode = events::keysym_to_element(keysym as c_uint);
--- a/src/platform_impl/linux/x11/event_processor.rs
+++ b/src/platform_impl/linux/x11/event_processor.rs
@@ -920,10 +920,13 @@ impl<T: 'static> EventProcessor<T> {
                                 },
                             });
 
-                            // Issue key press events for all pressed keys
-                            self.handle_pressed_keys(
+                            // Issue key press events for all pressed keys.
+                            Self::handle_pressed_keys(
+                                &wt,
                                 window_id,
                                 ElementState::Pressed,
+                                &self.mod_keymap,
+                                &mut self.device_mod_state,
                                 &mut callback,
                             );
                         }
@@ -942,9 +945,12 @@ impl<T: 'static> EventProcessor<T> {
                             let window_id = mkwid(xev.event);
 
                             // Issue key release events for all pressed keys
-                            self.handle_pressed_keys(
+                            Self::handle_pressed_keys(
+                                &wt,
                                 window_id,
                                 ElementState::Released,
+                                &self.mod_keymap,
+                                &mut self.device_mod_state,
                                 &mut callback,
                             );
 
@@ -1221,29 +1227,32 @@ impl<T: 'static> EventProcessor<T> {
     }
 
     fn handle_pressed_keys<F>(
-        &self,
+        wt: &super::EventLoopWindowTarget<T>,
         window_id: crate::window::WindowId,
         state: ElementState,
+        mod_keymap: &ModifierKeymap,
+        device_mod_state: &mut ModifierKeyState,
         callback: &mut F,
     ) where
         F: FnMut(Event<'_, T>),
     {
-        let wt = get_xtarget(&self.target);
-
         let device_id = mkdid(util::VIRTUAL_CORE_KEYBOARD);
-        let modifiers = self.device_mod_state.modifiers();
+        let modifiers = device_mod_state.modifiers();
 
-        // Get the set of keys currently pressed and apply Key events to each
-        let keys = wt.xconn.query_keymap();
-
-        for keycode in &keys {
-            if keycode < 8 {
-                continue;
-            }
-
+        // Get the set of keys currently pressed and apply Key events to each and also updating
+        // the pressed modifiers.
+        for keycode in wt.xconn.query_keymap().into_iter().filter(|k| *k >= 8) {
             let scancode = (keycode - 8) as u32;
             let keysym = wt.xconn.keycode_to_keysym(keycode);
             let virtual_keycode = events::keysym_to_element(keysym as c_uint);
+
+            if let Some(modifier) = mod_keymap.get_modifier(keycode as ffi::KeyCode) {
+                device_mod_state.key_event(
+                    ElementState::Pressed,
+                    keycode as ffi::KeyCode,
+                    modifier,
+                );
+            }
 
             #[allow(deprecated)]
             callback(Event::WindowEvent {
--- a/src/platform_impl/linux/x11/event_processor.rs
+++ b/src/platform_impl/linux/x11/event_processor.rs
@@ -20,6 +20,9 @@ use crate::{
     event_loop::EventLoopWindowTarget as RootELW,
 };
 
+/// The X11 documentation states: "Keycodes lie in the inclusive range [8,255]".
+const KEYCODE_OFFSET: usize = 8;
+
 pub(super) struct EventProcessor<T: 'static> {
     pub(super) dnd: Dnd,
     pub(super) ime_receiver: ImeReceiver,
@@ -565,7 +568,7 @@ impl<T: 'static> EventProcessor<T> {
                 // When a compose sequence or IME pre-edit is finished, it ends in a KeyPress with
                 // a keycode of 0.
                 if keycode != 0 {
-                    let scancode = keycode - 8;
+                    let scancode = keycode - KEYCODE_OFFSET;
                     let keysym = wt.xconn.lookup_keysym(xkev);
                     let virtual_keycode = events::keysym_to_element(keysym as c_uint);
 
@@ -1087,10 +1090,10 @@ impl<T: 'static> EventProcessor<T> {
 
                         let device_id = mkdid(xev.sourceid);
                         let keycode = xev.detail;
-                        if keycode < 8 {
+                        if keycode < KEYCODE_OFFSET {
                             return;
                         }
-                        let scancode = (keycode - 8) as u32;
+                        let scancode = (keycode - KEYCODE_OFFSET) as u32;
                         let keysym = wt.xconn.keycode_to_keysym(keycode as ffi::KeyCode);
                         let virtual_keycode = events::keysym_to_element(keysym as c_uint);
                         let modifiers = self.device_mod_state.modifiers();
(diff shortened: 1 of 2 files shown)


nixpulvis:
Ping @murarth, since I believe you worked on this last.

murarth:
`update_modifiers!` is invoked for motion and button events simply because X11 happens to provide the information and we want to pass it on to the user as soon as the change is observed.

Modifier state is not updated for raw {button,key,motion} events because X11 does not provide modifier state in the `XIRawEvent` struct.

src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -615,6 +615,7 @@ impl<T: 'static> EventProcessor<T> {
                 } else {
                     return;
                 };
+                let xkev: &mut ffi::XKeyEvent = xev.as_mut();
```
murarth, line 618: This is an invalid cast. `xev` is an `XGenericEvent` instance in this case.

nixpulvis: OK, I think this explains the strange behavior I was seeing. Joke's on me for trusting the type system within `x11_dl`.

That said, I'm back to being very confused as to how to track updates to the `device_mod_state` correctly within this event handler.

nixpulvis:
I've updated the PR description and rebased the code with what I now believe to be a correct fix. I realize there are probably refactoring to this overall logic that could be done. We can decide if that's needed after we are happy with the logic.

murarth:
@kchibisov Can you confirm that this fixes #1563?

kchibisov:
>@kchibisov Can you confirm that this fixes #1563?

I can't since I'm not the one having the issue, @nixpulvis is the one who is having that issue actually, and I've forwarded its bug report from alacritty.

nixpulvis:
@murarth yes, I can confirm this fixes the issue I was having.

In essence, I often launch `alacritty` with <kbd>LOGO</kbd> + <kbd>`</kbd>, which can easily spawn fast enough that I'm still holding <kbd>LOGO</kbd> during initialization. This causes a ` ModifiersChanged(LOGO)` event to be received, but I never received a corresponding `ModifiersChanged((empty))` on release.

However, now I'm seeing a bug in this patch. The `ModifiersChanged((empty))` comes in too early. It shouldn't be sent on `MappingNotify`, it should be sent on the release. So something is still wrong. Sorry for the confusion.

murarth:
@nixpulvis I think this new issue stems from `device_mod_state` tracking key press states, but failing to recognize when a key is already pressed at startup (or on `FocusIn`). I think if you get the current keymap (using `XConnection::query_keymap`) and reset `device_mod_state` based on that when either `FocusIn` or `MappingNotify` are received, it should resolve the issue.

I hope I'm being clear with that explanation. If not, I can write up a commit and add it to this PR.

kchibisov:
>I think this new issue stems from device_mod_state tracking key press states, but failing to recognize when a key is already pressed at startup (or on FocusIn). I think if you get the current keymap (using XConnection::query_keymap) and reset device_mod_state based on that when either FocusIn or MappingNotify are received, it should resolve the issue.

Yeah, that's what we concluded with @nixpulvis yesterday, so you'd need to process all pressed keys and fill pressed mods in `ModifierKeyState`, so pressed mods will be accurate. That should likely be done when getting you're getting `FocusIn` event.

winit also should likely not issue any `ModifiersChanged` events if the `ModifierKeymap` is empty, since it logically doesn't make any sense to do so, since you don't know where modifiers are, but since mods are on events it may be fine to always send mods update even before you've got a mapping.

nixpulvis:
I've quickly updated the code to test this approach. It seems like this does in fact resolve the issue properly this time. :tada: 

Now the question is just a matter of any potential code cleanup we can think of. I'm personally struggling with this a bit, because my instinct would be to probably try and clean up too much other unrelated stuff in here. Generally speaking I find it hard to parse through the logic of the key handling in general.

kchibisov:
@nixpulvis I've cleaned up your patch a bit, I guess it should also work?

```diff
diff --git a/CHANGELOG.md b/CHANGELOG.md
index da2e8e34..19539829 100644
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@
 - On Web, `WindowEvent::Resized` is now emitted when `Window::set_inner_size` is called.
 - **Breaking:** `Fullscreen` enum now uses `Borderless(Option<MonitorHandle>)` instead of `Borderless(MonitorHandle)` to allow picking the current monitor.
 - On MacOS, fix `WindowEvent::Moved` ignoring the scale factor.
+- On X11, fix incorrect modifiers state on startup.

 # 0.22.2 (2020-05-16)

diff --git a/src/platform_impl/linux/x11/event_processor.rs b/src/platform_impl/linux/x11/event_processor.rs
index 976c17b6..5c771ce8 100644
--- a/src/platform_impl/linux/x11/event_processor.rs
+++ b/src/platform_impl/linux/x11/event_processor.rs
@@ -920,10 +920,13 @@ impl<T: 'static> EventProcessor<T> {
                                 },
                             });

-                            // Issue key press events for all pressed keys
-                            self.handle_pressed_keys(
+                            // Issue key press events for all pressed keys.
+                            Self::handle_pressed_keys(
+                                &wt,
                                 window_id,
                                 ElementState::Pressed,
+                                &self.mod_keymap,
+                                &mut self.device_mod_state,
                                 &mut callback,
                             );
                         }
@@ -942,9 +945,12 @@ impl<T: 'static> EventProcessor<T> {
                             let window_id = mkwid(xev.event);

                             // Issue key release events for all pressed keys
-                            self.handle_pressed_keys(
+                            Self::handle_pressed_keys(
+                                &wt,
                                 window_id,
                                 ElementState::Released,
+                                &self.mod_keymap,
+                                &mut self.device_mod_state,
                                 &mut callback,
                             );

@@ -1221,30 +1227,33 @@ impl<T: 'static> EventProcessor<T> {
     }

     fn handle_pressed_keys<F>(
-        &self,
+        wt: &super::EventLoopWindowTarget<T>,
         window_id: crate::window::WindowId,
         state: ElementState,
+        mod_keymap: &ModifierKeymap,
+        device_mod_state: &mut ModifierKeyState,
         callback: &mut F,
     ) where
         F: FnMut(Event<'_, T>),
     {
-        let wt = get_xtarget(&self.target);
-
         let device_id = mkdid(util::VIRTUAL_CORE_KEYBOARD);
-        let modifiers = self.device_mod_state.modifiers();
-
-        // Get the set of keys currently pressed and apply Key events to each
-        let keys = wt.xconn.query_keymap();
-
-        for keycode in &keys {
-            if keycode < 8 {
-                continue;
-            }
+        let modifiers = device_mod_state.modifiers();

+        // Get the set of keys currently pressed and apply Key events to each and also updating
+        // the pressed modifiers.
+        for keycode in wt.xconn.query_keymap().into_iter().filter(|k| *k >= 8) {
             let scancode = (keycode - 8) as u32;
             let keysym = wt.xconn.keycode_to_keysym(keycode);
             let virtual_keycode = events::keysym_to_element(keysym as c_uint);

+            if let Some(modifier) = mod_keymap.get_modifier(keycode as ffi::KeyCode) {
+                device_mod_state.key_event(
+                    ElementState::Pressed,
+                    keycode as ffi::KeyCode,
+                    modifier,
+                );
+            }
+
             #[allow(deprecated)]
             callback(Event::WindowEvent {
                 window_id,

```

src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -1221,30 +1227,33 @@ impl<T: 'static> EventProcessor<T> {
...
+        // the pressed modifiers.
+        for keycode in wt.xconn.query_keymap().into_iter().filter(|k| *k >= 8) {
             let scancode = (keycode - 8) as u32;
             let keysym = wt.xconn.keycode_to_keysym(keycode);
             let virtual_keycode = events::keysym_to_element(keysym as c_uint);
 
+            if let Some(modifier) = mod_keymap.get_modifier(keycode as ffi::KeyCode) {
+                device_mod_state.key_event(
+                    ElementState::Pressed,
+                    keycode as ffi::KeyCode,
```
murarth, line 1252: The two `as ffi::KeyCode` casts here are unnecessary. The iterator type is explicitly `ffi::KeyCode`.

Also `ElementState::Pressed` should instead be the `state` argument passed to the function, as this function is called at two different points. When focus is gained, we're simulating new keys being pressed. When focus is lost, we are simulating all keys being released.

nixpulvis: Thank for catching that. Updated.

murarth (approved):

src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -920,10 +920,13 @@ impl<T: 'static> EventProcessor<T> {
                                 },
                             });
 
-                            // Issue key press events for all pressed keys
-                            self.handle_pressed_keys(
+                            // Issue key press events for all pressed keys.
```
chrisduerr, line 923: Based on the other comments in this diff, seems like there shouldn't be punctuation at the end?

src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -1221,30 +1227,29 @@ impl<T: 'static> EventProcessor<T> {
...
-
-        for keycode in &keys {
-            if keycode < 8 {
-                continue;
-            }
+        let modifiers = device_mod_state.modifiers();
 
+        // Get the set of keys currently pressed and apply Key events to each and also updating
+        // the pressed modifiers.
+        for keycode in wt.xconn.query_keymap().into_iter().filter(|k| *k >= 8) {
```
chrisduerr, line 1244: This isn't different from the existing code, but I'd like to know why keycodes below 8 are filtered? Just seems kinda arbitrary without documentation.

src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -1221,30 +1227,29 @@ impl<T: 'static> EventProcessor<T> {
...
             let scancode = (keycode - 8) as u32;
             let keysym = wt.xconn.keycode_to_keysym(keycode);
             let virtual_keycode = events::keysym_to_element(keysym as c_uint);
 
+            if let Some(modifier) = mod_keymap.get_modifier(keycode as ffi::KeyCode) {
+                device_mod_state.key_event(state, keycode, modifier);
+            }
```
chrisduerr, line 1251: Without looking at the rest of the code, I'd assume this is the place where the bugfix actually happens and the modifiers state tracking is fixed? So `key_event` function on the `device_mod_state` is responsible for updating the modifiers?

chrisduerr (changes requested):

src/platform_impl/linux/x11/event_processor.rs
nixpulvis: LOL, I managed to completely skip reading that. I was even wondering why it was showing up in the diff.

src/platform_impl/linux/x11/event_processor.rs
nixpulvis: Correct.

Without this the modifier state change needed from the synthetic key events triggered by `FocusIn` never happened.

src/platform_impl/linux/x11/event_processor.rs
nixpulvis: Running `xmodmap -pk` gives me a somewhat unsatisfying answer:

```
There are 7 KeySyms per KeyCode; KeyCodes range from 8 to 255.

    KeyCode	Keysym (Keysym)	...
    Value  	Value   (Name) 	...

      8    	
      9    	0xff1b (Escape)	0x0000 (NoSymbol)	0xff1b (Escape)	
...
```

nixpulvis: "Keycodes lie in the inclusive range [8,255]" [[1]]( https://www.x.org/releases/X11R7.7/doc/xproto/x11protocol.html#Keyboards)

nixpulvis: Here's a potential explanation: http://who-t.blogspot.com/2014/12/why-255-keycode-limit-in-x-is-real.html?showComment=1417626693719#c1258137695853791876

src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -1221,30 +1230,33 @@ impl<T: 'static> EventProcessor<T> {
...
+        let modifiers = device_mod_state.modifiers();
 
-        for keycode in &keys {
-            if keycode < 8 {
-                continue;
-            }
-
-            let scancode = (keycode - 8) as u32;
+        // Get the set of keys currently pressed and apply Key events to each and also updating
+        // the pressed modifiers.
```
chrisduerr, line 1246:
```suggestion
        // Update modifiers state and emit key events based on which keys are currently pressed.
```

Shorter and a little easier to read I think. The `and also updating` part sounds a bit strange.

chrisduerr (changes requested):

src/platform_impl/linux/x11/event_processor.rs
kchibisov: Yeah, I agree. Since I was the one who wrote that part, I guess it's fine if I commit it.

chrisduerr (approved):

PR #1726, opened 2020-09-24
Update readme info regarding WebAssembly and web target

alvinhochun:
The readme contained really outdated info which was probably written back when Winit still used Emscripten. This update reflects the current web support more accurately.

([Rendered](https://github.com/rust-windowing/winit/blob/64100ec157f621632c206a37dd6303297898879c/README.md#webassembly))

cc @ryanisaacg


ryanisaacg (approved):

ryanisaacg:
I'm not sure why the status checks aren't running, but I can't merge until they pass.

alvinhochun:
I rebased and force-pushed but it looks like the checks are still stuck. Let's wait for a bit then I guess.

alvinhochun:
I think the checks aren't running because this PR only changes the readme, which doesn't match the path filter specified in the github workflow?

ryanisaacg:
That makes sense but puts us in a bit of a pickle, because I can't merge anything that doesn't pass the checks and I also can't change the branch protection rules.

ryanisaacg:
cc @francesca64, would you be able to merge this?

francesca64:
@ryanisaacg I'll gladly invoke the dark power.

For a long-term solution, it doesn't look like branch protection rules are sophisticated enough to be conditional on what files changed / etc... we have control over which checks are required, but that's about it. So we should probably either lift the requirements, or change the checks, unfortunately. Maybe so that the checks still run on non-.rs files, but are just no-ops?

PR #1729, opened 2020-10-02
Release 0.23.0

kchibisov:


alvinhochun:
Since the changelog is so long, do you think the items should be reordered for a bit and grouped by platform?

chrisduerr:
> Since the changelog is so long, do you think the items should be reordered for a bit and grouped by platform?

I think it would be nice to make a general policy about that, so it doesn't have to be done retroactively. But this might get quite complicated when you have stuff like things that affects only linux/macOS for example. But you can always try to group things as reasonably as possible.

kchibisov:
I don't think that reordering the chagnelog is something we should do here. If we want 'ordered' changelog we should open an issue and agree on some format.

chrisduerr:
> I don't think that reordering the chagnelog is something we should do here. If we want 'ordered' changelog we should open an issue and agree on some format.

This certainly shouldn't be done on a whim, but agreed upon by everyone. But generally there is no problem with reordering changelogs in the release bump given that a format has been agreed upon.

PR #1733, opened 2020-10-08
Added WindowHasFocus and WindowLostFocus events to Android

maxded:
Ran into the issue were `WindowEvent::Focused` would always return false on Android, which was caused by unhandled events  from`ndk-glue`. This is a simple PR that solves that issue, updated the changelog as well!

Tested locally on my Android device and works as expected.


VZout (approved):
Looks good! This was an oversight in my initial implementation.

msiglreith (approved):
Thanks!

PR #1736, opened 2020-10-09
Fix failing assertion on start-up with Safari

### LINKED ISSUE #1734: winit 0.23: Failing assertion with Safari on macOS
After upgrading from 0.22.1 to 0.23, we're running into a failing assertion on "start-up":

```
panicked at 'assertion failed: `(left == right)`
  left: `false`,
 right: `true`', /Users/simon/.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.23.0/src/platform_impl/web/web_sys/scaling.rs:68:13

Stack:



__wbg_new_59cb74e423758ede@http://localhost:8080/app.bundle.js:689:132
wasm-stub@[wasm code]
<?>.wasm-function[console_error_panic_hook::Error::new::hdb4d49a68a543c24]@[wasm code]
<?>.wasm-function[console_error_panic_hook::hook_impl::h9716bce768f17344]@[wasm code]
<?>.wasm-function[console_error_panic_hook::hook::h586abbba1151be8d]@[wasm code]
<?>.wasm-function[core::ops::function::Fn::call::hba1e174aa40ce9a8]@[wasm code]
<?>.wasm-function[std::panicking::rust_panic_with_hook::hc3afc981c0e7d6a8]@[wasm code]
<?>.wasm-function[std::panicking::begin_panic_handler::{{closure}}::hefb02c9f4c7b9e72]@[wasm code]
<?>.wasm-function[std::sys_common::backtrace::__rust_end_short_backtrace::h3f57de3f451b589e]@[wasm code]
<?>.wasm-function[rust_begin_unwind]@[wasm code]
<?>.wasm-function[std::panicking::begin_panic_fmt::h5802585e40202418]@[wasm code]
<?>.wasm-function[winit::platform_impl::platform::backend::scaling::ScaleChangeDetectorInternal::create_mql::hfb13c54c98c691ff]@[wasm code]
<?>.wasm-function[winit::platform_impl::platform::backend::scaling::ScaleChangeDetectorInternal::new::h395d43e8ecfcf1b1]@[wasm code]
<?>.wasm-function[winit::platform_impl::platform::backend::scaling::ScaleChangeDetector::new::hcb035cd5601fe1e5]@[wasm code]
<?>.wasm-function[winit::platform_impl::platform::event_loop::runner::Shared<T>::set_on_scale_change::h9deb11feb67a1f5c]@[wasm code]
<?>.wasm-function[winit::platform_impl::platform::event_loop::window_target::WindowTarget<T>::run::hc04244059ec485b5]@[wasm code]
<?>.wasm-function[winit::platform_impl::platform::event_loop::EventLoop<T>::run::h3f6931dc2635136a]@[wasm code]
<?>.wasm-function[winit::event_loop::EventLoop<T>::run::h910affad9a270080]@[wasm code]
...
```

That's the `assert_eq` in `create_mql`:

```rust
    fn create_mql(
        closure: Closure<dyn FnMut(MediaQueryListEvent)>,
    ) -> Option<MediaQueryListHandle> {
        let current_scale = super::scale_factor();
        // This media query initially matches the current `devicePixelRatio`.
        // We add 0.0001 to the lower and upper bounds such that it won't fail
        // due to floating point precision limitations.
        let media_query = format!(
            "(min-resolution: {:.4}dppx) and (max-resolution: {:.4}dppx)",
            current_scale - 0.0001,
            current_scale + 0.0001,
        );
        let mql = MediaQueryListHandle::new(&media_query, closure);
        if let Some(mql) = &mql {
            assert_eq!(mql.mql().matches(), true);
        }
        mql
    }
```

I added debug output and the media query that's formatted is this:

`(min-resolution: 1.9999dppx) and (max-resolution: 2.0001dppx)`

Indeed, if I take any web site in Safari, open the web inspector, open the console and run the following snippet the `matches` property is also false:

```js
window.matchMedia("(min-resolution: 1.9999dppx) and (max-resolution: 2.0001dppx)").matches
false = $2
```

This issue does not appear with Chrome on macOS.

tronical:
The initial media query that's used to watch for device pixel ratio
changes should match. Unfortunately it doesn't with Safari and the
corresponding assertion fails. This is because resolution is not a
supported support property for queries. As a workaround, this patch
extends the query to optionally match for the webkit specific DPR
property directly.

This fixes #1734

### DIFF
--- a/src/platform_impl/web/web_sys/scaling.rs
+++ b/src/platform_impl/web/web_sys/scaling.rs
@@ -59,9 +59,9 @@ impl ScaleChangeDetectorInternal {
         // We add 0.0001 to the lower and upper bounds such that it won't fail
         // due to floating point precision limitations.
         let media_query = format!(
-            "(min-resolution: {:.4}dppx) and (max-resolution: {:.4}dppx)",
-            current_scale - 0.0001,
-            current_scale + 0.0001,
+            "(min-resolution: {min_scale:.4}dppx) and (max-resolution: {max_scale:.4}dppx),
+             (-webkit-min-device-pixel-ratio: {min_scale:.4}) and (-webkit-max-device-pixel-ratio: {max_scale:.4})",
+            min_scale = current_scale - 0.0001, max_scale= current_scale + 0.0001,
         );
         let mql = MediaQueryListHandle::new(&media_query, closure);
         if let Some(mql) = &mql {


src/platform_impl/web/web_sys/scaling.rs
```diff
@@ -58,10 +58,12 @@ impl ScaleChangeDetectorInternal {
...
         // due to floating point precision limitations.
+        let min_scale = current_scale - 0.0001;
+        let max_scale = current_scale + 0.0001;
         let media_query = format!(
-            "(min-resolution: {:.4}dppx) and (max-resolution: {:.4}dppx)",
-            current_scale - 0.0001,
-            current_scale + 0.0001,
+            "(min-resolution: {:.4}dppx) and (max-resolution: {:.4}dppx),
+             (-webkit-min-device-pixel-ratio: {:.4}) and (-webkit-max-device-pixel-ratio: {:.4})",
+            min_scale, max_scale, min_scale, max_scale,
```
ryanisaacg, line 66: Rather than repeating `min_scale` and `max_scale`, could you use the named format mode? e.g. `{min_scale:.4}`. It should make this a little clearer

ryanisaacg (changes requested):
Thanks for the quick fix! Just one comment to address and we'll be good to merge.

src/platform_impl/web/web_sys/scaling.rs
tronical: Ah yes, excellent point, that's much cleaner.

ryanisaacg (approved):

ryanisaacg:
Thanks so much!

alvinhochun:
Ouch, I did not see that they are not supported by Webkit. Sorry about that. (I don't have a system running macOS to test with.)

I guess it should've been a check to return `None` instead of an assert, so that it can fail gracefully as it is not a critical error. But if it didn't panic, you won't even notice it not working!

By the way, don't you think this is worth mentioning in the changelog?

Issue #1643, opened 2020-08-04
Rename platform::desktop::EventLoopExtDesktop


msiglreith:
With android support for `EventLoopExtDesktop` the naming doesn't fit anymore. Maybe just drop the `Desktop` suffix or exchange it with `Poll`?

VZout:
Yep I agree, I should have renamed it when I implemented `run_return` for android. Feel free to make a quick PR 😄

PR #1738, opened 2020-10-10
Rename desktop eventloop extensions to run_return extension

msiglreith:

Closed https://github.com/rust-windowing/winit/issues/1643

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -35,10 +35,10 @@
 //! entire program terminates.
 //!
 //! Winit no longer uses a `EventLoop::poll_events() -> impl Iterator<Event>`-based event loop
-//! model, since that can't be implemented properly on web and mobile platforms and works poorly on
-//! most desktop platforms. However, this model can be re-implemented to an extent on desktops with
-//! [`EventLoopExtDesktop::run_return`]. See that method's documentation for more reasons about why
-//! it's discouraged, beyond mobile/web compatibility reasons.
+//! model, since that can't be implemented properly on some platforms (e.g web, iOS) and works poorly on
+//! most other platforms. However, this model can be re-implemented to an extent with
+//! [`EventLoopExtRunReturn::run_return`]. See that method's documentation for more reasons about why
+//! it's discouraged, beyond compatibility reasons.
 //!
 //!
 //! ```no_run
@@ -109,7 +109,7 @@
 //! window visible only once you're ready to render into it.
 //!
 //! [`EventLoop`]: event_loop::EventLoop
-//! [`EventLoopExtDesktop::run_return`]: ./platform/desktop/trait.EventLoopExtDesktop.html#tymethod.run_return
+//! [`EventLoopExtRunReturn::run_return`]: ./platform/run_return/trait.EventLoopExtRunReturn.html#tymethod.run_return
 //! [`EventLoop::new()`]: event_loop::EventLoop::new
 //! [event_loop_run]: event_loop::EventLoop::run
 //! [`ControlFlow`]: event_loop::ControlFlow
--- a/src/platform/mod.rs
+++ b/src/platform/mod.rs
@@ -11,7 +11,7 @@
 //!
 //! And the following platform-specific module:
 //!
-//! - `desktop` (available on `windows`, `unix`, and `macos`)
+//! - `run_return` (available on `windows`, `unix`, `macos`, and `android`)
 //!
 //! However only the module corresponding to the platform you're compiling to will be available.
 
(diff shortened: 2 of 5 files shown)


src/platform/mod.rs
```diff
@@ -11,7 +11,7 @@
 //!
 //! And the following platform-specific module:
 //!
-//! - `desktop` (available on `windows`, `unix`, and `macos`)
+//! - `run_return` (available on `windows`, `unix`, `macos` and `android`)
```
kchibisov, line 14:
```suggestion
//! - `run_return` (available on `windows`, `unix`, `macos`, and `android`)
```

src/lib.rs
```diff
@@ -35,10 +35,10 @@
 //! entire program terminates.
 //!
 //! Winit no longer uses a `EventLoop::poll_events() -> impl Iterator<Event>`-based event loop
-//! model, since that can't be implemented properly on web and mobile platforms and works poorly on
-//! most desktop platforms. However, this model can be re-implemented to an extent on desktops with
-//! [`EventLoopExtDesktop::run_return`]. See that method's documentation for more reasons about why
-//! it's discouraged, beyond mobile/web compatibility reasons.
+//! model, since that can't be implemented properly on some platforms (e.g web) and works poorly on
```
kchibisov, line 38: I think we should mention `ios` here, since it's the only reason why it's like that. I'm not sure in what way though.

```suggestion
//! model, since that can't be implemented properly on some platforms (e.g web, ios) and works poorly on
```

kchibisov (changes requested):
I think it makes sense to name it like that, since the whole purpose of that trait is to provide run return anyway.

kchibisov (approved):

kchibisov:
I'd like to wait a bit with merging this, since the change is minor, but it's breaking, so no need to merge that 'early'.

src/lib.rs
```diff
@@ -35,10 +35,10 @@
 //! entire program terminates.
 //!
 //! Winit no longer uses a `EventLoop::poll_events() -> impl Iterator<Event>`-based event loop
-//! model, since that can't be implemented properly on web and mobile platforms and works poorly on
-//! most desktop platforms. However, this model can be re-implemented to an extent on desktops with
-//! [`EventLoopExtDesktop::run_return`]. See that method's documentation for more reasons about why
-//! it's discouraged, beyond mobile/web compatibility reasons.
+//! model, since that can't be implemented properly on some platforms (e.g web, ios) and works poorly on
```
francesca64, line 38:
```suggestion
//! model, since that can't be implemented properly on some platforms (e.g web, iOS) and works poorly on
```

francesca64 (approved):
I'm also in favor of this 👍

francesca64:
@kchibisov do we want to move forward on this yet, or were you thinking we should hold off on making the next release breaking?

aloucks:
> I'd like to wait a bit with merging this, since the change is minor, but it's breaking, so no need to merge that 'early'.

Perhaps you could add a type alias for the old trait name and mark it `#[deprecated]`.

kchibisov:
>Perhaps you could add a type alias for the old trait name and mark it #[deprecated].

I mean it seems like the next release will be already breaking, due to some other PR, so I'm not sure that we need to add deprecation for anything, knowing that there's no real purpose for it, and it'll add just complexity.

kchibisov:
Also, this PR requires a rebase, and I guess we can merge it, since some breaking change was already merged afaics.

chrisduerr (commented):
I don't want to come in here and start bikeshedding on something that is already approved, but I've been thinking about this for a while.

Is `EventLoopExtRunReturn` really the best name? The function is called `run_return`, sure, but it seems weird to name a trait based on the name of the function it provides.

The trait name should ideally describe what it does, or what it should be used for. And I'm not sure it's great for that. Generally I feel like the name `run_return` is not very descriptive?

Maybe something like `EventLoopExtNonBlocking` would be better, though I don't love this either.

If that's just me, I'm fine with just going for `RunReturn`, but I was curious if other people share this sentiment and might have alternative suggestions that are better choices for method names.

kchibisov:
>Maybe something like EventLoopExtNonBlocking would be better, though I don't love this either.

run also non blocking though, it just doesn't return, however run return is actually returning.

chrisduerr:
> run also non blocking though, it just doesn't return, however run return is actually returning.

Yeah, that's fair, which is why I don't like that name either. If you see no problem with `RunReturn` I wouldn't block the merge of this either. Since the function is named this way anyways, it should be fine to move on with this PR either way.

Senre (approved):

PR #1743, opened 2020-10-17
Feature gate more dependencies

kchibisov:
Wayland backend is self contained and only requires
smithay-client-toolkit to work, thus feature gate the
rest of the deps that are only for X11.


chrisduerr (approved):
As long as CI is happy, everything should be good.

aloucks:
This reintroduces an issue where the same dependency is specified multiple times (e.g. `parking_lot`). This created issues in the past where the dependency was updated in one place, but not in the others. This was fixed in #884.

kchibisov:
>This reintroduces an issue where the same dependency is specified multiple times (e.g. parking_lot). This created issues in the past where the dependency was updated in one place, but not in the others. This was fixed in #884.

It's not an issue if someone can at least look at the code when they are bumping things, if they just blindly merge nothing can be done a tthis point.

chrisduerr:
Also you say that, but this PR removed the duplicate `objc` dependency which did actually use different versions. So it basically just doesn't change anything.

PR #1746, opened 2020-10-22
[Windows] Fix use after free during window destruction

### LINKED ISSUE #1745: [Windows] Use after free in public_window_callback during window destruction
**Description**

I am able to reproduce stable crash condition, when, with fast window creation-deletion cycle, program crashes with exit code.
image
Crash is happening inside `winit::platform_impl::platform::event_loop::public_window_callback()` when trying to get `subclass_input.event_loop_runner.thread_msg_target()`.
image
That happens because `subclass_input_ptr` points to corrupted memory.
Normal `public_window_callback` call:
```rust
subclass_input_ptr as usize = 2523725522720
subclass_input.event_loop_runner.as_ref() as *const _ as usize = 2523636609216
```
Call before crash:
```rust
subclass_input_ptr as usize = 2523725522720
subclass_input.event_loop_runner.as_ref() as *const _ as usize = 21
```
i.e. `EventLoopRunnerShared` inside `SubclassInput` is `Rc` pointer that points to memory address 21, which is obviously not valid.

This happens during window destruction, inside of `winuser::DestroyWindow()`
image

So, inside `public_window_callback` we receive message `DESTROY_MSG_ID` and call `winuser::DestroyWindow` which sends its own message, which invoke `public_window_callback` again, but its state is already invalid.
What's the message? It's [WM_NCDESTROY](https://docs.microsoft.com/en-us/windows/win32/winmsg/wm-ncdestroy).

**Context**
It was found during testing of `Drop` implementation for `SwapChain` and `Surface` in `wgpu-rs` crate.
More interesting thing is that this bug is only reproducible with Vulkan backend. Dx12 backend doesn't crash.
While Dx12 backend is also receiving `WM_NCDESTROY`, but Rc pointer still points to valid location.
At first, i though that's probably because of a different order of `WM_DESTROY` and `WM_NCDESTROY` events,
But winapi docs is stating that:
```
The WM_NCDESTROY message is sent after the child windows have been destroyed. In contrast, WM_DESTROY is sent before the child windows are destroyed.
```
So, it just that when `WM_NCDESTROY ` event is processed in `public_window_callback` pointer `SubclassInput` is always points to already freed memory, it's just not overwritten yet by something else.

**Proposed solution**
Right now `SubclassInput` is destroyed right after `WM_DESTROY` event
```rust
        winuser::WM_DESTROY => {
            use crate::event::WindowEvent::Destroyed;
            ole2::RevokeDragDrop(window);
            subclass_input.send_event(Event::WindowEvent {
                window_id: RootWindowId(WindowId(window)),
                event: Destroyed,
            });
            subclass_input.event_loop_runner.remove_window(window);

            drop(subclass_input);
            Box::from_raw(subclass_input_ptr as *mut SubclassInput<T>);
            0
        }
```
If winapi reference is right, we should do that after `WM_NCDESTROY` event:
```rust
        winuser::WM_DESTROY => {
            use crate::event::WindowEvent::Destroyed;
            ole2::RevokeDragDrop(window);
            subclass_input.send_event(Event::WindowEvent {
                window_id: RootWindowId(WindowId(window)),
                event: Destroyed,
            });
            subclass_input.event_loop_runner.remove_window(window);
            0
        }

        winuser::WM_NCDESTROY => {
            drop(subclass_input);
            Box::from_raw(subclass_input_ptr as *mut SubclassInput<T>);
            0
        }
```
At least, this fixes crash for me. Before this fix, in a tight loop, test app was crashing after 1-3 seconds. Now it works fine for minutes.

qthree:
Fixes #1745.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - **Breaking:** On Windows, include prefix byte in scancodes.
 - On Wayland, fix window not being resizeable when using `with_min_inner_size` in `WindowBuilder`.
 - On Unix, fix cross-compiling to wasm32 without enabling X11 or Wayland.
+- On Windows, fix use after free crash during window destruction.
 
 # 0.23.0 (2020-10-02)
 
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -810,7 +810,10 @@ unsafe extern "system" fn public_window_callback<T: 'static>(
                 event: Destroyed,
             });
             subclass_input.event_loop_runner.remove_window(window);
+            0
+        }
 
+        winuser::WM_NCDESTROY => {
             drop(subclass_input);
             Box::from_raw(subclass_input_ptr as *mut SubclassInput<T>);
             0


kchibisov (changes requested):
This change requires CHANGELOG entry

msiglreith (approved):

kchibisov (approved):
Style seems ok.

PR #1747, opened 2020-10-27
Fix WindowEvent::ReceivedCharacter on web

### LINKED ISSUE #1741: wasm: WindowEvent::CharacterReceived is not sent anymore
Since 0.23 WindowEvent::CharacterReceived is not sent anymore. I tracked it down to commit 6cfddfea21319ef7e28f4a6dc16e119ab033f550 at https://github.com/rust-windowing/winit/blob/5a78fe33e84023ca4c89cc30ebeebac0fe89e8f0/src/platform_impl/web/web_sys/canvas.rs#L160 where the `prevent_default()` call results in the `keypress` event to be never received on the canvas element.

This is happening at least with Safari and Chrome on macOS.

tronical:
The event was never sent to the application because of the unconditional
preventDefault()
call on keydown.

Fixes #1741

### DIFF
--- a/src/platform_impl/web/stdweb/canvas.rs
+++ b/src/platform_impl/web/stdweb/canvas.rs
@@ -10,9 +10,9 @@ use stdweb::js;
 use stdweb::traits::IPointerEvent;
 use stdweb::unstable::TryInto;
 use stdweb::web::event::{
-    BlurEvent, ConcreteEvent, FocusEvent, FullscreenChangeEvent, IEvent, KeyDownEvent,
-    KeyPressEvent, KeyUpEvent, MouseWheelEvent, PointerDownEvent, PointerMoveEvent,
-    PointerOutEvent, PointerOverEvent, PointerUpEvent,
+    BlurEvent, ConcreteEvent, FocusEvent, FullscreenChangeEvent, IEvent, IKeyboardEvent,
+    KeyDownEvent, KeyPressEvent, KeyUpEvent, ModifierKey, MouseWheelEvent, PointerDownEvent,
+    PointerMoveEvent, PointerOutEvent, PointerOverEvent, PointerUpEvent,
 };
 use stdweb::web::html_element::CanvasElement;
 use stdweb::web::{document, EventListenerHandle, IElement, IEventTarget, IHtmlElement};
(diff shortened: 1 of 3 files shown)


ryanisaacg:
Looks good! Before we merge, we have to address the compile error on stdweb. Importing `stdweb::web::event::IKeyboardEvent` should do the trick.

tronical:
Thank you! That did the trick :)

src/platform_impl/web/stdweb/canvas.rs
```diff
@@ -122,7 +122,11 @@ impl Canvas {
         F: 'static + FnMut(ScanCode, Option<VirtualKeyCode>, ModifiersState),
     {
         self.on_keyboard_release = Some(self.add_user_event(move |event: KeyUpEvent| {
-            event.prevent_default();
+            let event_key = &event.key();
+            let is_key_string = event_key.len() == 1 || !event_key.is_ascii();
+            if !is_key_string {
+                event.prevent_default();
+            }
```
alvinhochun, line 129: Shouldn't this be put inside `on_keyboard_press` instead of `on_keyboard_release`?

alvinhochun (changes requested):
See inline review. Also, I think the code deserves at least a short comment to describe what it does and why it is needed.

alvinhochun:
Sorry, this doesn't block normal shortcut keys (e.g. Ctrl+R), but it does block function keys (e.g. F5). It does appear that we still need to check for modifiers. This *should* work, but I'm not entirely sure if it does what people usually expects AltGr to do:

```rust
let is_shortcut_modifiers = (event.ctrl_key() || event.alt_key()) && !event.get_modifier_state("AltGr");
if !is_key_string || is_shortcut_modifiers {
    event.prevent_default();
}
```

Also, I believe you need to add `preventDefault` to the `keypress` listener to stop Space or Shift+Space from scrolling the page.

src/platform_impl/web/stdweb/canvas.rs
tronical: Ouch, yes, of course. Thanks for spotting!

tronical:
> Sorry, this doesn't block normal shortcut keys (e.g. Ctrl+R), but it does block function keys (e.g. F5). It does appear that we still need to check for modifiers. This should work, but I'm not entirely sure if it does what people usually expects AltGr to do:
>
> let is_shortcut_modifiers = (event.ctrl_key() || event.alt_key()) && !event.get_modifier_state("AltGr");
> if !is_key_string || is_shortcut_modifiers {
>     event.prevent_default();
> }

Done. Also added the missing comment, you're totally right that the handling needs it.

> Also, I believe you need to add preventDefault to the keypress listener to stop Space or Shift+Space from scrolling the page.

I'm not sure I understand, can you elaborate?

tronical:
The remaining CI failure looks like an unrelated network problem to me.

alvinhochun:
> > Also, I believe you need to add preventDefault to the keypress listener to stop Space or Shift+Space from scrolling the page.
> 
> I'm not sure I understand, can you elaborate?

Open https://codepen.io/alvinhochun/pen/yLJpwGX, focus on the canvas and try pressing Space. You will find that the page gets scrolled down. Uncomment `e.preventDefault();` in the `keypress` listener and you will find that this behaviour is blocked.

tronical:
> Open https://codepen.io/alvinhochun/pen/yLJpwGX, focus on the canvas and try pressing Space. You will find that the page gets scrolled down. Uncomment e.preventDefault(); in the keypress listener and you will find that this behaviour is blocked.

Got it, thanks!

Added this in a follow-up commit on top (and fixed changelog grammar in initial commit).

alvinhochun (approved):
Looks good to me!

PR #1749, opened 2020-10-29
Issue 1588

### LINKED ISSUE #1588: [Android] `raw_window_handle` usage requirements are inobvious
On android, when creating a window and then (pretty much) immediately requesting it's raw window handle, winit will panic at [src/platform_impl/android/mod.rs:418](https://github.com/rust-windowing/winit/blob/master/src/platform_impl/android/mod.rs#L418): `native window null`.

This is because the code will call into [android-ndk-rs:ndk-glue/src/lib.rs:68](https://github.com/rust-windowing/android-ndk-rs/blob/867d2344d9c3aca3ce3fc45af98adb81c8ec1912/ndk-glue/src/lib.rs#L50), which then tries to read the native window, which is not yet set because it's [callback](https://github.com/rust-windowing/android-ndk-rs/blob/867d2344d9c3aca3ce3fc45af98adb81c8ec1912/ndk-glue/src/lib.rs#L230) has not arrived yet.

I can currently reliably reproduce this issue on my Google Pixel 4.

I'm not sure whether this issue should be in the winit repo or the [android-ndk-rs](https://github.com/rust-windowing/android-ndk-rs) repo, feel free to transfer it.

genuinebyte:

Looking to resolve #1588. I'm not 100% on the language used so let me know what you think. Always happy to make changes if needed.

### DIFF
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -505,7 +505,7 @@ impl Window {
         let a_native_window = if let Some(native_window) = ndk_glue::native_window().as_ref() {
             unsafe { native_window.ptr().as_mut() as *mut _ as *mut _ }
         } else {
-            panic!("native window null");
+            panic!("The native window is null. Are you in between Event::Resumed and Event::Suspended?");
         };
         let mut handle = raw_window_handle::android::AndroidHandle::empty();
         handle.a_native_window = a_native_window;
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -505,7 +505,7 @@ impl Window {
         let a_native_window = if let Some(native_window) = ndk_glue::native_window().as_ref() {
             unsafe { native_window.ptr().as_mut() as *mut _ as *mut _ }
         } else {
-            panic!("The native window is null. Are you in between Event::Resumed and Event::Suspended?");
+            panic!("Cannot get the native window, it's null and will always be null before Event::Resumed and after Event::Suspended. Make sure you only call this function between those events.");
         };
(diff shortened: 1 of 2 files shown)


src/platform_impl/android/mod.rs
```diff
@@ -505,7 +505,7 @@ impl Window {
         let a_native_window = if let Some(native_window) = ndk_glue::native_window().as_ref() {
             unsafe { native_window.ptr().as_mut() as *mut _ as *mut _ }
         } else {
-            panic!("native window null");
+            panic!("The native window is null. Are you in between Event::Resumed and Event::Suspended?");
```
francesca64, line 508: It's totally fine to make this even more descriptive, since this message is really our first line of defense against user confusion. Currently, it doesn't really outline that this can only be called between those events, so expanding on that would be great!

francesca64 (changes requested):
Thanks so much for addressing this! This should be a big step forward for Android backend usability.

In the future, could you use a more descriptive PR title? Instead of watching this repo, I just skim the issue and PR list every day for new things related to mobile, so descriptive titles help a lot with that. Additionally, you can always ping me on anything mobile-related!

src/platform_impl/android/mod.rs
genuinebyte: I hope it is now all informing!

genuinebyte:
> Thanks so much for addressing this! This should be a big step forward for Android backend usability.
No problem! I'm trying to see what I can do to help get Android running on the main winit/glutin crates, so I had a look at the Android issues and this seemed easy enough.

I'll try to use better titles in the future and I'll be sure to ping you!

francesca64 (approved):
Looks great, thanks!

PR #1756, opened 2020-11-05
X11: Fix `request_redraw` not waking the event loop

### LINKED ISSUE #1755: Requesting redraw on X11 does not wake up the event loop
It seems like while requesting the redraw stores the need for a redraw in a shared variable, it never actually makes an attempt to wake up the event loop. This means that this function only delivers the requested event when users already get other events.

@murarth are you aware of this? If not, is there an easy way to wake up the event loop on X11?

Somewhat related: https://github.com/rust-windowing/glutin/issues/1335

murarth:
Fixes #1755

### DIFF
--- a/src/platform_impl/linux/x11/mod.rs
+++ b/src/platform_impl/linux/x11/mod.rs
@@ -309,9 +309,11 @@ impl<T: 'static> EventLoop<T> {
             }
             // Empty the redraw requests
             {
-                while let Ok(wid) = self.redraw_channel.try_recv() {
+                while let Ok(window_id) = self.redraw_channel.try_recv() {
                     sticky_exit_callback(
-                        Event::RedrawRequested(crate::window::WindowId(super::WindowId::X(wid))),
+                        Event::RedrawRequested(crate::window::WindowId(super::WindowId::X(
+                            window_id,
+                        ))),
                         &self.target,
                         &mut control_flow,
                         &mut callback,
--- a/src/platform_impl/linux/x11/mod.rs
+++ b/src/platform_impl/linux/x11/mod.rs
@@ -24,7 +24,7 @@ pub use self::{
 
 use std::{
     cell::RefCell,
-    collections::{HashMap, HashSet},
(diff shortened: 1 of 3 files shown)


murarth:
@chrisduerr Can you confirm this fixes the issue?

src/platform_impl/linux/x11/mod.rs
```diff
@@ -301,10 +309,7 @@ impl<T: 'static> EventLoop<T> {
             }
             // Empty the redraw requests
             {
-                // Release the lock to prevent deadlock
-                let windows: Vec<_> = wt.pending_redraws.lock().unwrap().drain().collect();
-
-                for wid in windows {
+                while let Ok(wid) = self.redraw_channel.try_recv() {
```
chrisduerr, line 312: Can we name this `window_id`? I feel like we don't gain anything by using a cryptic `wid` identifier here.

chrisduerr (changes requested):
I can indeed confirm that this does fix the `request_redraw` bug.

src/platform_impl/linux/x11/mod.rs
```diff
@@ -309,9 +309,11 @@ impl<T: 'static> EventLoop<T> {
             }
             // Empty the redraw requests
             {
-                while let Ok(wid) = self.redraw_channel.try_recv() {
+                while let Ok(window_id) = self.redraw_channel.try_recv() {
                     sticky_exit_callback(
-                        Event::RedrawRequested(crate::window::WindowId(super::WindowId::X(wid))),
+                        Event::RedrawRequested(crate::window::WindowId(super::WindowId::X(
+                            window_id,
+                        ))),
```
chrisduerr, line 316: Oh now I know why you used `wid`. :smiley: 

Generally I'd recommend to extract things into variables when ugly wrapping like this is introduced, using full paths from the crate root also certainly doesn't help with making things shorter.

That's up to you as the maintainer of the X11 backend though.

chrisduerr (approved):

PR #1758, opened 2020-11-07
X11: Fix multiple RedrawRequested events per event loop iteration

murarth:
Fixup to #1756, which could generate multiple `RedrawRequested` events for the same window id in a single event loop iteration.

### DIFF
--- a/src/platform_impl/linux/x11/mod.rs
+++ b/src/platform_impl/linux/x11/mod.rs
@@ -309,19 +309,20 @@ impl<T: 'static> EventLoop<T> {
             }
             // Empty the redraw requests
             {
-                // Set of windows that have already received `RedrawRequested` this iteration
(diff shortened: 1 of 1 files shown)


chrisduerr (approved):

murarth:
I realized a couple of issues with the previous fix:

* Multiple calls to `request_redraw` for the same window id would result in multiple instances of `RedrawRequested` for that window id in the same event loop iteration.
* If an application calls `request_redraw(id)` while handling `RedrawRequested(id)`, it would cause an infinite loop as winit continues processing each new window id.

This PR solves both these issues. The only remaining vulnerability of this solution is that if an application thread were to continually fill the `request_redraw` channel, winit would never stop processing these incoming window ids. However, I feel that no reasonable application should do such a thing, so it's a vulnerability that can be overlooked.

PR #1759, opened 2020-11-08
Allow listening to file open events on macos

### LINKED ISSUE #1751: Double-click opening a file on macOS
When a file is associated with an app on macOS, the [`application:openFile:`](https://developer.apple.com/documentation/appkit/nsapplicationdelegate/1428612-application?language=objc) or the [`application:openFiles:`](https://developer.apple.com/documentation/appkit/nsapplicationdelegate/1428742-application?language=objc) function gets called instead of passing the target filepath as an argument. The application must handle theese calls otherwise the os will display an error message to the user that the appilcation doesn't support opening the file type.

Winit currently doesn't seem to handle either of these and there doesn't seem to be any other way of making a winit application able to handle opening a file for when the user double-clicks on an associated file.

ArturKovacs:
Closes #1751 

I did not check the example program point because I would indeed like to create an example for this but that would need to be bundled with `cargo bundle` and a script would be required to patch the `Info.plist` file in the bundle to allow associating certain file-types with the app. I could create an example program in a separate repo where all this is handled. But I'm not sure how that would be or if that should be referenced from this repo.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 - Added `Window::drag_window`. Implemented on Windows, macOS, X11 and Wayland.
 - On X11, bump `mio` to 0.7.
 - On macOS, emit `RedrawRequested` events immediately while the window is being resized.
+- On macOS, add `EventLoopWindowTargetExtMacOS::set_file_open_callback`.
 
 # 0.24.0 (2020-12-09)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@
 - **Breaking**: `MouseButton::Other` now uses `u16`.
 - On macOS, add `set_open_files_callback`.
 - On macOS, implement `Event::OpenFiles` to allow opening associated files.
+- **Breaking** On all platfoms, add `Event::OpenFiles` and implement it on macOS.
 
 # 0.23.0 (2020-10-02)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,7 @@
 - On Android, support multi-touch.
 - On Wayland, extra mouse buttons are not dropped anymore.
 - **Breaking**: `MouseButton::Other` now uses `u16`.
+- On macOS, add `set_open_files_callback`.
 
 # 0.23.0 (2020-10-02)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,7 @@
 - On macOS, add `set_open_files_callback`.
 - On macOS, implement `Event::OpenFiles` to allow opening associated files.
 - **Breaking** On all platfoms, add `Event::OpenFiles` and implement it on macOS.
+- **Breaking** New `Event::OpenFiles` Event, for macOS "open with..." support.
 
 # 0.23.0 (2020-10-02)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - On Windows, change the default window size (1024x768) to match the default on other desktop platforms (800x600).
 - On Windows, fix bug causing mouse capture to not be released.
 - On Windows, fix fullscreen not preserving minimized/maximized state.
+- On macOS, add `EventLoopExtMacOS::set_file_open_callback`.
 - On Android, unimplemented events are marked as unhandled on the native event loop.
 - On Windows, added `WindowBuilderExtWindows::with_menu` to set a custom menu at window creation time.
 - On Android, bump `ndk` and `ndk-glue` to 0.3: use predefined constants for event `ident`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@
 - Added `Window::drag_window`. Implemented on Windows, macOS, X11 and Wayland.
 - On X11, bump `mio` to 0.7.
 - On macOS, emit `RedrawRequested` events immediately while the window is being resized.
-- On macOS, add `EventLoopWindowTargetExtMacOS::set_file_open_callback`.
+- On macOS, add `EventLoopExtMacOS::set_file_open_callback`.
 
 # 0.24.0 (2020-12-09)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@
 - On Wayland, extra mouse buttons are not dropped anymore.
 - **Breaking**: `MouseButton::Other` now uses `u16`.
 - On macOS, add `set_open_files_callback`.
+- On macOS, implement `Event::OpenFiles` to allow opening associated files.
 
 # 0.23.0 (2020-10-02)
 
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -127,7 +127,6 @@ If your PR makes notable changes to Winit's features, please update this section
 * Hidden titlebar
 * Hidden titlebar buttons
 * Full-size content view
-* Forward file-open evnets
 
 ### Unix
 * Window urgency
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -127,6 +127,7 @@ If your PR makes notable changes to Winit's features, please update this section
 * Hidden titlebar
 * Hidden titlebar buttons
 * Full-size content view
+* Forward file-open evnets
 
 ### Unix
 * Window urgency
--- a/src/event.rs
+++ b/src/event.rs
@@ -67,14 +67,15 @@ pub enum Event<'a, T: 'static> {
         event: DeviceEvent,
     },
 
-    /// Emitted on macOS to let the application know that a file or files are requested to be opened.
+    /// Emitted to let the application know that a file or files are requested to be opened.
     ///
-    /// This for example happens when the user double-clicks on a file in Finder and the file's type
-    /// is associated with this application. More precisely this event is emitted when either of
-    /// `application:openFile:` or `application:openFiles:` is called on the application delegate.
+    /// This for example happens when the user double-clicks on a file in the file browser
+    /// and the file's type is associated with this application.
     ///
-    /// Note that to allow associating file types with an application, said application must have
-    /// an appropriate `Info.plist` file in the application bundle defining `CFBundleDocumentTypes`.
+    /// ## Platform-specific
+    /// 
+    /// This is only implemented on **macOS**. (Other systems usually provide the path as an
+    /// argument. See: `std::env::args`.)
     OpenFiles(Vec<PathBuf>),
 
     /// Emitted when an event is sent from [`EventLoopProxy::send_event`](crate::event_loop::EventLoopProxy::send_event)
--- a/src/event.rs
+++ b/src/event.rs
@@ -68,7 +68,7 @@ pub enum Event<'a, T: 'static> {
     },
 
     /// Emitted on macOS to let the application know that a file or files are requested to be opened.
-    /// 
+    ///
     /// This for example happens when the user double-clicks on a file in Finder and the file's type
     /// is associated with this application. (Other systems usually pass the path to the files as
     /// arguments to the program. See `std::env::args()`.)
--- a/src/event.rs
+++ b/src/event.rs
@@ -67,17 +67,6 @@ pub enum Event<'a, T: 'static> {
         event: DeviceEvent,
     },
 
-    /// Emitted to let the application know that a file or files are requested to be opened.
-    ///
-    /// This for example happens when the user double-clicks on a file in the file browser
-    /// and the file's type is associated with this application.
-    ///
-    /// ## Platform-specific
-    ///
-    /// This is only implemented on **macOS**. (Other systems usually provide the path as an
-    /// argument. See: `std::env::args`.)
-    OpenFiles(Vec<PathBuf>),
-
     /// Emitted when an event is sent from [`EventLoopProxy::send_event`](crate::event_loop::EventLoopProxy::send_event)
     UserEvent(T),
 
@@ -142,7 +131,6 @@ impl<T: Clone> Clone for Event<'static, T> {
                 device_id: *device_id,
                 event: event.clone(),
             },
-            OpenFiles(files) => OpenFiles(files.clone()),
             NewEvents(cause) => NewEvents(cause.clone()),
             MainEventsCleared => MainEventsCleared,
             RedrawRequested(wid) => RedrawRequested(*wid),
@@ -161,7 +149,6 @@ impl<'a, T> Event<'a, T> {
             UserEvent(_) => Err(self),
             WindowEvent { window_id, event } => Ok(WindowEvent { window_id, event }),
             DeviceEvent { device_id, event } => Ok(DeviceEvent { device_id, event }),
-            OpenFiles(files) => Ok(OpenFiles(files)),
             NewEvents(cause) => Ok(NewEvents(cause)),
             MainEventsCleared => Ok(MainEventsCleared),
             RedrawRequested(wid) => Ok(RedrawRequested(wid)),
@@ -182,7 +169,6 @@ impl<'a, T> Event<'a, T> {
                 .map(|event| WindowEvent { window_id, event }),
             UserEvent(event) => Some(UserEvent(event)),
             DeviceEvent { device_id, event } => Some(DeviceEvent { device_id, event }),
-            OpenFiles(files) => Some(OpenFiles(files)),
             NewEvents(cause) => Some(NewEvents(cause)),
             MainEventsCleared => Some(MainEventsCleared),
             RedrawRequested(wid) => Some(RedrawRequested(wid)),
--- a/src/event.rs
+++ b/src/event.rs
@@ -67,6 +67,21 @@ pub enum Event<'a, T: 'static> {
         event: DeviceEvent,
     },
 
+    /// Emitted on macOS to let the application know that a file or files are requested to be opened.
+    /// 
+    /// This for example happens when the user double-clicks on a file in Finder and the file's type
+    /// is associated with this application. (Other systems usually pass the path to the files as
+    /// arguments to the program. See `std::env::args()`.)
+    ///
+    /// More precisely this event is emmited when either of `application:openFile:` or
+    /// `application:openFiles:` is called on the application delegate. Winit always responds to
+    /// the OS as if the operation was succesful regardles of wheter this event was handled or
+    /// wheter the files could indeed be opened.
+    ///
+    /// Note that to allow associating file types with an application, said application must have
+    /// an appropriate `Info.plist` file in the application bundle defining `CFBundleDocumentTypes`.
+    OpenFiles(Vec<PathBuf>),
+
     /// Emitted when an event is sent from [`EventLoopProxy::send_event`](crate::event_loop::EventLoopProxy::send_event)
     UserEvent(T),
 
@@ -131,6 +146,7 @@ impl<T: Clone> Clone for Event<'static, T> {
                 device_id: *device_id,
                 event: event.clone(),
             },
+            OpenFiles(files) => OpenFiles(files.clone()),
             NewEvents(cause) => NewEvents(cause.clone()),
             MainEventsCleared => MainEventsCleared,
             RedrawRequested(wid) => RedrawRequested(*wid),
@@ -149,6 +165,7 @@ impl<'a, T> Event<'a, T> {
             UserEvent(_) => Err(self),
             WindowEvent { window_id, event } => Ok(WindowEvent { window_id, event }),
             DeviceEvent { device_id, event } => Ok(DeviceEvent { device_id, event }),
+            OpenFiles(files) => Ok(OpenFiles(files)),
             NewEvents(cause) => Ok(NewEvents(cause)),
             MainEventsCleared => Ok(MainEventsCleared),
             RedrawRequested(wid) => Ok(RedrawRequested(wid)),
@@ -169,6 +186,7 @@ impl<'a, T> Event<'a, T> {
                 .map(|event| WindowEvent { window_id, event }),
             UserEvent(event) => Some(UserEvent(event)),
             DeviceEvent { device_id, event } => Some(DeviceEvent { device_id, event }),
+            OpenFiles(files) => Some(OpenFiles(files)),
             NewEvents(cause) => Some(NewEvents(cause)),
             MainEventsCleared => Some(MainEventsCleared),
             RedrawRequested(wid) => Some(RedrawRequested(wid)),
--- a/src/event.rs
+++ b/src/event.rs
@@ -73,7 +73,7 @@ pub enum Event<'a, T: 'static> {
     /// and the file's type is associated with this application.
     ///
     /// ## Platform-specific
-    /// 
+    ///
     /// This is only implemented on **macOS**. (Other systems usually provide the path as an
     /// argument. See: `std::env::args`.)
     OpenFiles(Vec<PathBuf>),
--- a/src/event.rs
+++ b/src/event.rs
@@ -70,13 +70,8 @@ pub enum Event<'a, T: 'static> {
     /// Emitted on macOS to let the application know that a file or files are requested to be opened.
     ///
     /// This for example happens when the user double-clicks on a file in Finder and the file's type
-    /// is associated with this application. (Other systems usually pass the path to the files as
-    /// arguments to the program. See `std::env::args()`.)
-    ///
-    /// More precisely this event is emmited when either of `application:openFile:` or
-    /// `application:openFiles:` is called on the application delegate. Winit always responds to
-    /// the OS as if the operation was succesful regardles of wheter this event was handled or
-    /// wheter the files could indeed be opened.
+    /// is associated with this application. More precisely this event is emitted when either of
+    /// `application:openFile:` or `application:openFiles:` is called on the application delegate.
     ///
     /// Note that to allow associating file types with an application, said application must have
     /// an appropriate `Info.plist` file in the application bundle defining `CFBundleDocumentTypes`.
--- a/src/platform/macos.rs
+++ b/src/platform/macos.rs
@@ -1,6 +1,6 @@
 #![cfg(target_os = "macos")]
 
-use std::{os::raw::c_void, path::PathBuf};
+use std::os::raw::c_void;
 
 use crate::{
     dpi::LogicalSize,
@@ -211,21 +211,6 @@ pub trait EventLoopWindowTargetExtMacOS {
     fn hide_application(&self);
     /// Hide the other applications. In most applications this is typically triggered with Command+Option-H.
     fn hide_other_applications(&self);
-
-    /// Registers or removes the callback that gets called when the user double-clicks
-    /// a file in Finder whose type is associated with this application. More specifically the
-    /// callback gets called when either `application:openFile:` or `application:openFiles:` gets
-    /// called on the application delegate.
-    ///
-    /// Winit defines both `application:openFile:` and `application:openFiles:` and it will respond
-    /// to the OS as if the operation was succesful regardles of wheter there is a callback registered
-    /// and regardless of whether the callback could indeed open the files.
-    ///
-    /// Note that to allow associating file types with an application, said application must have
-    /// an appropriate `Info.plist` file in the application bundle defining `CFBundleDocumentTypes`.
-    ///
-    /// Set the calback to `None` to prevent it from getting called. By default it is `None`.
-    fn set_open_files_callback<F: FnMut(&[PathBuf]) + 'static>(&self, callback: Option<F>);
 }
 
 impl<T> EventLoopWindowTargetExtMacOS for EventLoopWindowTarget<T> {
@@ -239,13 +224,5 @@ impl<T> EventLoopWindowTargetExtMacOS for EventLoopWindowTarget<T> {
         let cls = objc::runtime::Class::get("NSApplication").unwrap();
         let app: cocoa::base::id = unsafe { msg_send![cls, sharedApplication] };
         unsafe { msg_send![app, hideOtherApplications: 0] }
-    }
-
-    fn set_open_files_callback<F>(&self, callback: Option<F>)
-    where
-        F: FnMut(&[PathBuf]) + 'static,
-    {
-        let mut borrowed = self.p.open_files_callback.borrow_mut();
-        *borrowed = callback.map(|cb| Box::new(cb) as Box<dyn FnMut(&[PathBuf])>);
     }
 }
--- a/src/platform/macos.rs
+++ b/src/platform/macos.rs
@@ -2,6 +2,8 @@
 
 use std::{os::raw::c_void, path::PathBuf};
 
+use cocoa::{base::nil, foundation::NSAutoreleasePool};
+
 use crate::{
     dpi::LogicalSize,
     event_loop::EventLoopWindowTarget,
@@ -231,9 +233,10 @@ pub trait EventLoopWindowTargetExtMacOS {
     ///
     /// Other systems usually provide the path as a program argument.
     /// See: `std::env::args`.
-    fn set_file_open_callback<F>(&self, callback: Option<F>)
-    where
-        F: FnMut(Vec<PathBuf>) -> FileOpenResult + Send + 'static;
+    fn set_file_open_callback(
+        &self,
+        callback: Option<Box<dyn FnMut(Vec<PathBuf>) -> FileOpenResult + Send + 'static>>,
+    );
 }
 
 impl<T> EventLoopWindowTargetExtMacOS for EventLoopWindowTarget<T> {
@@ -249,10 +252,25 @@ impl<T> EventLoopWindowTargetExtMacOS for EventLoopWindowTarget<T> {
         unsafe { msg_send![app, hideOtherApplications: 0] }
     }
 
-    fn set_file_open_callback<F>(&self, callback: Option<F>)
-    where
-        F: FnMut(Vec<PathBuf>) -> FileOpenResult + Send + 'static,
-    {
-        AppState::set_file_open_callback(callback);
+    fn set_file_open_callback(
+        &self,
+        callback: Option<Box<dyn FnMut(Vec<PathBuf>) -> FileOpenResult + Send + 'static>>,
+    ) {
+        let has_callback = callback.is_some();
+        let should_change_delegate = AppState::set_file_open_callback(callback);
+        if should_change_delegate {
+            unsafe {
+                let cls = class!(NSApplication);
+                let app: cocoa::base::id = msg_send![cls, sharedApplication];
+                let delegate = if has_callback {
+                    *self.p.delegate_with_file_open
+                } else {
+                    *self.p.delegate
+                };
+                let pool = NSAutoreleasePool::new(nil);
+                let () = msg_send![app, setDelegate: delegate];
+                let () = msg_send![pool, drain];
+            }
+        }
     }
 }
--- a/src/platform/macos.rs
+++ b/src/platform/macos.rs
@@ -1,11 +1,12 @@
 #![cfg(target_os = "macos")]
 
-use std::os::raw::c_void;
+use std::{os::raw::c_void, path::PathBuf};
 
 use crate::{
     dpi::LogicalSize,
     event_loop::EventLoopWindowTarget,
     monitor::MonitorHandle,
+    platform_impl::AppState,
     window::{Window, WindowBuilder},
 };
 
@@ -205,12 +206,34 @@ impl MonitorHandleExtMacOS for MonitorHandle {
     }
 }
 
+#[derive(Debug, PartialEq, Eq, Copy, Clone)]
+pub enum FileOpenResult {
+    Success,
+    Cancel,
+    Failure,
+}
+
 /// Additional methods on `EventLoopWindowTarget` that are specific to macOS.
 pub trait EventLoopWindowTargetExtMacOS {
     /// Hide the entire application. In most applications this is typically triggered with Command-H.
     fn hide_application(&self);
     /// Hide the other applications. In most applications this is typically triggered with Command+Option-H.
     fn hide_other_applications(&self);
+
+    /// Set or remove the callback which gets called when a file or files are requested to
+    /// be opened.
+    ///
+    /// This for example happens when the user double-clicks on a file in Finder and the
+    /// file's type is associated with this application.
+    ///
+    /// The callback function receives the absolute path to the target files as an argument and it should
+    /// indicate the success of the operation in its return value. See: `application:openFiles:` in AppKit.
+    ///
+    /// Other systems usually provide the path as a program argument.
+    /// See: `std::env::args`.
+    fn set_file_open_callback<F>(&self, callback: Option<F>)
+    where
+        F: FnMut(Vec<PathBuf>) -> FileOpenResult + Send + 'static;
 }
 
 impl<T> EventLoopWindowTargetExtMacOS for EventLoopWindowTarget<T> {
@@ -224,5 +247,12 @@ impl<T> EventLoopWindowTargetExtMacOS for EventLoopWindowTarget<T> {
         let cls = objc::runtime::Class::get("NSApplication").unwrap();
         let app: cocoa::base::id = unsafe { msg_send![cls, sharedApplication] };
         unsafe { msg_send![app, hideOtherApplications: 0] }
+    }
+
+    fn set_file_open_callback<F>(&self, callback: Option<F>)
+    where
+        F: FnMut(Vec<PathBuf>) -> FileOpenResult + Send + 'static,
+    {
+        AppState::set_file_open_callback(callback);
     }
 }
--- a/src/platform/macos.rs
+++ b/src/platform/macos.rs
@@ -6,7 +6,7 @@ use cocoa::{base::nil, foundation::NSAutoreleasePool};
 
 use crate::{
     dpi::LogicalSize,
-    event_loop::EventLoopWindowTarget,
+    event_loop::{EventLoop, EventLoopWindowTarget},
     monitor::MonitorHandle,
     platform_impl::AppState,
     window::{Window, WindowBuilder},
@@ -215,13 +215,7 @@ pub enum FileOpenResult {
     Failure,
 }
 
-/// Additional methods on `EventLoopWindowTarget` that are specific to macOS.
-pub trait EventLoopWindowTargetExtMacOS {
-    /// Hide the entire application. In most applications this is typically triggered with Command-H.
-    fn hide_application(&self);
-    /// Hide the other applications. In most applications this is typically triggered with Command+Option-H.
-    fn hide_other_applications(&self);
-
+pub trait EventLoopExtMacOS {
     /// Set or remove the callback which gets called when a file or files are requested to
     /// be opened.
     ///
@@ -239,19 +233,7 @@ pub trait EventLoopWindowTargetExtMacOS {
     );
 }
 
-impl<T> EventLoopWindowTargetExtMacOS for EventLoopWindowTarget<T> {
-    fn hide_application(&self) {
-        let cls = objc::runtime::Class::get("NSApplication").unwrap();
-        let app: cocoa::base::id = unsafe { msg_send![cls, sharedApplication] };
-        unsafe { msg_send![app, hide: 0] }
-    }
-
-    fn hide_other_applications(&self) {
-        let cls = objc::runtime::Class::get("NSApplication").unwrap();
-        let app: cocoa::base::id = unsafe { msg_send![cls, sharedApplication] };
-        unsafe { msg_send![app, hideOtherApplications: 0] }
-    }
-
+impl<T> EventLoopExtMacOS for EventLoop<T> {
     fn set_file_open_callback(
         &self,
         callback: Option<Box<dyn FnMut(Vec<PathBuf>) -> FileOpenResult + Send + 'static>>,
@@ -272,5 +254,27 @@ impl<T> EventLoopWindowTargetExtMacOS for EventLoopWindowTarget<T> {
                 let () = msg_send![pool, drain];
             }
         }
+    }
+}
+
+/// Additional methods on `EventLoopWindowTarget` that are specific to macOS.
+pub trait EventLoopWindowTargetExtMacOS {
+    /// Hide the entire application. In most applications this is typically triggered with Command-H.
+    fn hide_application(&self);
+    /// Hide the other applications. In most applications this is typically triggered with Command+Option-H.
+    fn hide_other_applications(&self);
+}
+
+impl<T> EventLoopWindowTargetExtMacOS for EventLoopWindowTarget<T> {
+    fn hide_application(&self) {
+        let cls = objc::runtime::Class::get("NSApplication").unwrap();
+        let app: cocoa::base::id = unsafe { msg_send![cls, sharedApplication] };
+        unsafe { msg_send![app, hide: 0] }
+    }
+
+    fn hide_other_applications(&self) {
+        let cls = objc::runtime::Class::get("NSApplication").unwrap();
+        let app: cocoa::base::id = unsafe { msg_send![cls, sharedApplication] };
+        unsafe { msg_send![app, hideOtherApplications: 0] }
     }
 }
--- a/src/platform/macos.rs
+++ b/src/platform/macos.rs
@@ -1,6 +1,6 @@
 #![cfg(target_os = "macos")]
 
-use std::os::raw::c_void;
+use std::{os::raw::c_void, path::PathBuf};
 
 use crate::{
     dpi::LogicalSize,
@@ -211,6 +211,21 @@ pub trait EventLoopWindowTargetExtMacOS {
     fn hide_application(&self);
     /// Hide the other applications. In most applications this is typically triggered with Command+Option-H.
     fn hide_other_applications(&self);
+
+    /// Registers or removes the callback that gets called when the user double-clicks
+    /// a file in Finder whose type is associated with this application. More specifically the
+    /// callback gets called when either `application:openFile:` or `application:openFiles:` gets
+    /// called on the application delegate.
+    ///
+    /// Winit defines both `application:openFile:` and `application:openFiles:` and it will respond
+    /// to the OS as if the operation was succesful regardles of wheter there is a callback registered
+    /// and regardless of whether the callback could indeed open the files.
+    ///
+    /// Note that to allow associating file types with an application, said application must have
+    /// an appropriate `Info.plist` file in the application bundle defining `CFBundleDocumentTypes`.
+    ///
+    /// Set the calback to `None` to prevent it from getting called. By default it is `None`.
+    fn set_open_files_callback<F: FnMut(&[PathBuf]) + 'static>(&self, callback: Option<F>);
 }
 
 impl<T> EventLoopWindowTargetExtMacOS for EventLoopWindowTarget<T> {
@@ -224,5 +239,13 @@ impl<T> EventLoopWindowTargetExtMacOS for EventLoopWindowTarget<T> {
         let cls = objc::runtime::Class::get("NSApplication").unwrap();
         let app: cocoa::base::id = unsafe { msg_send![cls, sharedApplication] };
         unsafe { msg_send![app, hideOtherApplications: 0] }
+    }
+
+    fn set_open_files_callback<F>(&self, callback: Option<F>)
+    where
+        F: FnMut(&[PathBuf]) + 'static,
+    {
+        let mut borrowed = self.p.open_files_callback.borrow_mut();
+        *borrowed = callback.map(|cb| Box::new(cb) as Box<dyn FnMut(&[PathBuf])>);
     }
 }
--- a/src/platform_impl/macos/app_delegate.rs
+++ b/src/platform_impl/macos/app_delegate.rs
@@ -62,12 +62,10 @@ lazy_static! {
 }
 
 /// Copies the contents of the ns string into a `String` which gets returned.
-fn ns_string_to_rust(ns_string: id) -> String {
-    unsafe {
-        let slice = slice::from_raw_parts(ns_string.UTF8String() as *mut u8, ns_string.len());
-        let string = str::from_utf8_unchecked(slice);
-        string.to_owned()
-    }
+unsafe fn ns_string_to_rust(ns_string: id) -> String {
+    let slice = slice::from_raw_parts(ns_string.UTF8String() as *mut u8, ns_string.len());
+    let string = str::from_utf8_unchecked(slice);
+    string.to_owned()
 }
 
 extern "C" fn new(class: &Class, _: Sel) -> id {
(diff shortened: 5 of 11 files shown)


ArturKovacs:
I created an example repo at https://github.com/ArturKovacs/winit_open_file

chrisduerr (commented):
I personally think that adding callbacks to a library that is usually based on an event loop it ships itself makes things extremely annoying to use downstream.

The state management required to integrate with the rest of the application would likely just require locking whenever the shared state is accessed, all of which each downstream user is responsible for themselves.

ArturKovacs:
That's a fair assesment. I made this implementation based on the discussion at #1751. Would you suggest using a platform specific event API instead? Something like:

```rust
enum Event {
   WindowEvent {...}
    // All the other events events 

    PlatformSpecific(PlatformSpecificEvent)
}

enum MacOsEvent {
    ...
}
impl PlatformSpecificEventMacOsExt for PlatformSpecificEvent {
    fn into_macos_event(self) -> MacOsEvent {...}
}
```

chrisduerr:
Events are optional anyways. So it's probably easiest to just emit an event and document clearly when it is emitted.

ArturKovacs:
I agree but @alvinhochun seemed to be against adding a variant to `Event` which only gets triggered on one platform.

ArturKovacs:
@chrisduerr if you are in executive position I'll just implement this as an event so that this can be closed. Otherwise I'm in a rather uncomfortable situation here as I'm willing to go with either approach but I don't know what I'm supposed to do to get this resolved. I would like to get some guidance for what should be the next step from here.

chrisduerr:
> @chrisduerr if you are in executive position

I am not. I am not the maintainer of the macOS backend. That would officially be @vbogaevsky I believe, I'm not sure if he's still around. Though if he's not around, I'm probably as good as it's gonna get.

ArturKovacs:
I just updated the implementation so that now an application level `Event` is sent. I also updated the example at https://github.com/ArturKovacs/winit_open_file

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 - On Wayland, the keypad arrow keys are now recognized.
 - **Breaking** Rename `desktop::EventLoopExtDesktop` to `run_return::EventLoopExtRunReturn`.
 - On Wayland, default font size in CSD increased from 11 to 17.
+- **Breaking** On all platfoms, add `Event::OpenFiles` and implement it on macOS.
```
chrisduerr, line 16:
```suggestion
- **Breaking** New `Event::OpenFiles` Event, for macOS "open with..." support.
```

FEATURES.md
```diff
@@ -126,6 +126,7 @@ If your PR makes notable changes to Winit's features, please update this section
 * Hidden titlebar
 * Hidden titlebar buttons
 * Full-size content view
+* Forward file-open evnets
```
chrisduerr, line 129: I'm not really sure what the goal of the FEATURES.md file is, but this seems a bit out of place here. There's also a typo.

src/event.rs
```diff
@@ -67,6 +67,21 @@ pub enum Event<'a, T: 'static> {
...
+    /// is associated with this application. (Other systems usually pass the path to the files as
+    /// arguments to the program. See `std::env::args()`.)
+    ///
+    /// More precisely this event is emmited when either of `application:openFile:` or
+    /// `application:openFiles:` is called on the application delegate. Winit always responds to
+    /// the OS as if the operation was succesful regardles of wheter this event was handled or
+    /// wheter the files could indeed be opened.
+    ///
+    /// Note that to allow associating file types with an application, said application must have
+    /// an appropriate `Info.plist` file in the application bundle defining `CFBundleDocumentTypes`.
```
chrisduerr, line 82: This is **a lot** of text for such a simple feature. We should aim at being concise with documentation to make sure people are actually able to read it.

I'd advise you to try and shorten it a bit yourself, try to focus on what a user of winit would care about, instead of documenting how the implementation inside of winit works. There's also a few typos, so a quick spellcheck is probably more useful than hoping a reviewer is able to spot everything.

If you need help, please let me know and I can try to make some suggestions.

src/platform_impl/macos/app_delegate.rs
```diff
@@ -36,10 +38,35 @@ lazy_static! {
...
+    let utf8_ptr = unsafe { ns_string.UTF8String() } as *mut u8;
+    let utf8_slice = unsafe { std::slice::from_raw_parts(utf8_ptr, utf8_len) };
+    let mut utf8_vec = Vec::<u8>::with_capacity(utf8_len);
+    unsafe {
+        utf8_vec.set_len(utf8_len);
+    }
+    utf8_vec.copy_from_slice(utf8_slice);
+
+    unsafe { String::from_utf8_unchecked(utf8_vec) }
+}
```
chrisduerr, line 68: This seems like a very convoluted way to convert between a UTF8String and a Rust String. Chances are this also isn't the first time this had to be done in winit, so I **strongly** suggest to look at functions like `get_characters` or `insert_text`.

src/platform_impl/macos/app_delegate.rs
```diff
@@ -79,3 +106,45 @@ extern "C" fn did_resign_active(this: &Object, _: Sel, _: id) {
...
+    let mut filenames_vec = Vec::with_capacity(1);
+    let string = ns_string_to_rust(filename);
+    filenames_vec.push(string.into());
+    let event = Event::OpenFiles(filenames_vec);
+    AppState::queue_event(EventWrapper::StaticEvent(event));
+    trace!("Completed `application:openFile:`");
+
+    // Return true to indicate to the OS that the file type is supported.
+    // (If the filetype turns out not to be supported, it's the application's
+    // responsibility to inform the user)
```
chrisduerr, line 121: Why is this singling out the filetype? There are so many other reasons why I file could potentially not be opened. It might be gone, the user might not have permissions to open it, or someone else might already be busy with it.

This is also not documentation for the user, this is internal documentation. So we should document here *why* we chose to always return `YES` instead of actually giving it a proper value.

src/platform_impl/macos/app_delegate.rs
```diff
@@ -79,3 +106,45 @@ extern "C" fn did_resign_active(this: &Object, _: Sel, _: id) {
     }
     trace!("Completed `applicationDidResignActive`");
 }
+
+extern "C" fn application_open_file(_this: &Object, _: Sel, _sender: id, filename: id) -> BOOL {
+    trace!("Triggered `application:openFile:`");
+    let mut filenames_vec = Vec::with_capacity(1);
+    let string = ns_string_to_rust(filename);
+    filenames_vec.push(string.into());
+    let event = Event::OpenFiles(filenames_vec);
```
chrisduerr, line 115:
```suggestion
    let string = ns_string_to_rust(filename);
    let event = Event::OpenFiles(vec![string.into()]);
```

src/platform_impl/macos/app_delegate.rs
```diff
@@ -79,3 +106,45 @@ extern "C" fn did_resign_active(this: &Object, _: Sel, _: id) {
...
+    trace!("Completed `application:openFile:`");
+
+    // Return true to indicate to the OS that the file type is supported.
+    // (If the filetype turns out not to be supported, it's the application's
+    // responsibility to inform the user)
+    YES
+}
+
+extern "C" fn application_open_files(_this: &Object, _: Sel, _sender: id, filenames: id) {
+    use cocoa::foundation::NSArray;
```
chrisduerr, line 126: Please put imports at the top of the file.

src/platform_impl/macos/app_delegate.rs
```diff
@@ -79,3 +106,45 @@ extern "C" fn did_resign_active(this: &Object, _: Sel, _: id) {
...
+
+    #[allow(non_upper_case_globals)]
+    const NSApplicationDelegateReplySuccess: i32 = 0;
```
chrisduerr, line 129: Please put this constant at the top of the file. Someone else might want to use it and we don't want to redefine it a billion times.

Also there's a reason why you got a warning; please fix it instead of just silencing it.

src/platform_impl/macos/app_delegate.rs
```diff
@@ -79,3 +106,45 @@ extern "C" fn did_resign_active(this: &Object, _: Sel, _: id) {
...
+    }
+    let event = Event::OpenFiles(filenames_vec);
+    AppState::queue_event(EventWrapper::StaticEvent(event));
+
+    let cls = objc::runtime::Class::get("NSApplication").unwrap();
+    let app: cocoa::base::id = unsafe { msg_send![cls, sharedApplication] };
+    // Indicate to the OS that the file types are supported.
+    // (If a filetype turns out not to be supported, it's the application's
+    // responsibility to inform the user)
+    unsafe { msg_send![app, replyToOpenOrPrint: NSApplicationDelegateReplySuccess] }
```
chrisduerr, line 147: What happens if we don't do that here. Do you know why this doesn't just return a boolean? That seems a bit inconsistent.

The documentation says that delegates "should" invoke this, not that they must.

chrisduerr (changes requested):

src/platform_impl/macos/app_delegate.rs
ArturKovacs: > Please put this constant at the top of the file.

Right away.

> Also there's a reason why you got a warning; please fix it instead of just silencing it.

I understand your concern but I feel like the tone was a bit uncalled for. The reason why I silenced the warning was that this constant gets to have the exact name that it was given in appkit. This is consistent with how types and functions are exposed by the `cocoa` crate.

src/platform_impl/macos/app_delegate.rs
ArturKovacs: The reason why I originally put this here is that I thought that omitting this would make the OS display a message saying that the application cannot open files in the "..." format. As shown on the image [here](https://stackoverflow.com/questions/37018964/on-mac-os-x-opening-my-document-by-double-clicking-at-the-document-fails-if-my-a). But it seems that as long as the `openFile` and `openFiles` functions are implemented this message isn't displayed even if those functions indicate failure.

Although I think that the two functions as defined by winit should have identical behavior so if `openFile` returns `YES` then `openFiles` should also indicate success. The reasoning is that even if this is not what the user expects it's probably better if the behavior is consistent as opposed to "it sometimes works but sometimes it doesn't".

FEATURES.md
ArturKovacs: Yeah, I wasn't sure about this either so I'm removing it.

src/platform_impl/macos/app_delegate.rs
chrisduerr: If macOS assumes success by default, the behavior is identical though. That's why I'm asking if it makes any difference. If it doesn't there's no difference between the two.

ArturKovacs: I went through the docs again in hopes of finding the answer to what's the effect of indicating success or failure. But I did not find the answer to that.

I then did some tests with a minimal Objective-C application selecting one or two files on my desktop and using the "right-click" context menu to open them with the test application. First of all, only `openFiles` was called even if I had a single file selected. I tried not calling the `replyToOpenOrPrint` function, and I tried all combinations of calling it on the `sharedApplication` and the `sender` with both success and failure. I could not see any difference on the GUI or within Console. In other words I couldn't see any difference at all.

I still believe that whatever we choose `openFile` to do, the same behavior should apply to `openFiles`.

What do you reckon, what should these functions indicate?

chrisduerr: Thanks for looking into this, that seems like really useful information. It sounds like it is mostly a recommendation then, based on the assumption that the application is not a toolkit and instead handles everything by itself.

> I still believe that whatever we choose openFile to do, the same behavior should apply to openFiles.

You can't return a boolean though, so that simply isn't an option. Trying to draw any parallels between the two like returning a boolean is the same as executing some callback is pointless. One does not stand in any relation to the other really.

Sure, we shouldn't return FALSE from one and return the callback with a success state, but if the callback is by default true, then returning TRUE should be fine.

Based on what I've seen, it seems like it *should* work fine to not execute that callback, however this might require some amount of trust that things aren't going to change in the future. I'm not sure that I trust in that. So the sensible and robust solution is probably to do what you're doing right now and keep it.

src/event.rs
```diff
@@ -67,6 +67,16 @@ pub enum Event<'a, T: 'static> {
         event: DeviceEvent,
     },
 
+    /// Emitted on macOS to let the application know that a file or files are requested to be opened.
```
chrisduerr, line 70: I think this first line should be generic and not platform specific. Mentioning that this is only supported on macOS should go in a platform-specific notice at the bottom. At least that's the way things are usually documented in winit.

src/event.rs
```diff
@@ -67,6 +67,16 @@ pub enum Event<'a, T: 'static> {
...
+    ///
+    /// This for example happens when the user double-clicks on a file in Finder and the file's type
+    /// is associated with this application. More precisely this event is emitted when either of
+    /// `application:openFile:` or `application:openFiles:` is called on the application delegate.
+    ///
+    /// Note that to allow associating file types with an application, said application must have
+    /// an appropriate `Info.plist` file in the application bundle defining `CFBundleDocumentTypes`.
+    OpenFiles(Vec<PathBuf>),
```
chrisduerr, line 78: I just noticed that this is not a device or window event. That's probably going to cause significant headache downstream. Though I'm not entirely sure what to do about it. I don't think it makes sense to attach this to any device or window.

src/event.rs
```diff
@@ -67,6 +67,16 @@ pub enum Event<'a, T: 'static> {
         event: DeviceEvent,
     },
 
+    /// Emitted on macOS to let the application know that a file or files are requested to be opened.
+    ///
+    /// This for example happens when the user double-clicks on a file in Finder and the file's type
+    /// is associated with this application. More precisely this event is emitted when either of
+    /// `application:openFile:` or `application:openFiles:` is called on the application delegate.
```
chrisduerr, line 74: I don't see any reason to mention `openFile` or `openFiles` here.

src/event.rs
```diff
@@ -67,6 +67,16 @@ pub enum Event<'a, T: 'static> {
...
+    ///
+    /// Note that to allow associating file types with an application, said application must have
+    /// an appropriate `Info.plist` file in the application bundle defining `CFBundleDocumentTypes`.
```
chrisduerr, line 77: Is this related to the `OpenFiles` event? Isn't that just documentation for how to get macOS to allow you to open files with your application?

chrisduerr (changes requested):

src/event.rs
ArturKovacs: I think it is related because I don't know any other scenario for when this event is relevant than picking the application from the file browser to open a file with, and in that case the application must have an appropriate `Info.plist`, otherwise it cannot be selected. It took much longer for me than I would have liked to figure out all the stuff needed for this to work so I thought I'll save some time for others by giving a pointer for what to look up.

ArturKovacs: Just to be clear I'm not completely against removing it, I'm just describing why we might consider keeping it.

chrisduerr: I just don't think we should be documenting things that are largely unrelated here. This is only the `OpenFiles` event documentation, not macOS documentation. If there's some official docs you want to link, maybe. But explaining a larger setup process to get to the point of receiving the event seems out of scope.

src/platform_impl/macos/app_delegate.rs
ArturKovacs: Awesome possum

src/event.rs
ArturKovacs: Just out of curiosity, why does this cause headache downstream? The only thing I can think of is having to add in a match arm about once per application. Why is that a cause of significant headache?

src/event.rs
ArturKovacs: Alright, I removed it.

src/event.rs
chrisduerr: All events are either WindowEvents or DeviceEvents. This would be the first event that doesn't come from winit that is a "root event".

ArturKovacs:
It seems to me that I addressed all the feedback. Is there anything I'm missing?

chrisduerr (commented):
> It seems to me that I addressed all the feedback. Is there anything I'm missing?

I'm not really a fan of how this is done in macOS and the resulting implementation in winit. So I personally don't really feel like it's going to be very beneficial to merge it.

ArturKovacs:
To be honest I'm not really a fan either of how this is done in macOS, but if this is not exposed by winit, how should winit applications open associated files on macOS? I personally don't see any other way, and denying this feature would mean that no image viewer, video player, text editor, etc. could be properly made for macOS using winit.

If this PR is not that way to solve the issue, that's fine, but I would be very sad if this wouldn't be solved at all.

chrisduerr:
>  how should winit applications open associated files on macOS

Some things, especially those that do not integrate at all with any other platforms (like this one), are better provided separately outside of winit.

ArturKovacs:
Okay, that's reasonable if said feature can be implemented outside of winit. I don't see how this could be done outside of winit while using winit for the rest. Please describe how that would be done.

ArturKovacs:
I wouldn't want to involve other contributors but we seem to be completely stuck with something that I believe is quite important for macOS applications. @francesca64 and @ryanisaacg you both seem to have the skills and the authority required to help find a solution here, so I'd like to ask you to help.

ryanisaacg:
Sorry, I've never done any macOS-related development. I just happen to have a mac that I can compile code on, but I don't know anything about mac-specific programming.

alvinhochun:
It seems that the problem here is the lack of a generic way for users to handle native events that aren't handled by Winit. Since Winit is in control of the `NSApplication` delegates, outside code simply can't intercept the macOS-specific `openFiles` event and others, so this function cannot be implemented externally. (I'm just assuming this is the case.)

If this is true, then Winit should will need to expose these native events for users to intercept them, or perhaps there can be a way to allow the user to set their own `NSApplication` delegates. Otherwise users won't be able to use Winit to build applications that relies on such platform-specific events.

I don't know macOS application programming at all so I can't really give any more constructive comments.

src/platform_impl/macos/app_delegate.rs
francesca64: @chrisduerr you're being rough throughout this PR, and it looks like you ignored @ArturKovacs here instead of apologizing. Even if that wasn't your intention, it's still how @ArturKovacs feels, and that's something important to address.

I know open source can be stressful, so please don't push yourself too hard.

src/platform_impl/macos/app_delegate.rs
```diff
@@ -1,10 +1,20 @@
...
-    runtime::{Class, Object, Sel},
+    runtime::{Class, Object, Sel, BOOL, YES},
 };
-use std::os::raw::c_void;
+use std::{os::raw::c_void, slice, str};
+
+/// This constant represents the one with an identical name in appkit. See:
+/// https://developer.apple.com/documentation/appkit/nsapplicationdelegatereply/nsapplicationdelegatereplysuccess?language=occ
+#[allow(non_upper_case_globals)]
+const NSApplicationDelegateReplySuccess: i32 = 0;
```
francesca64, line 17: The [`ffi`](https://github.com/rust-windowing/winit/blob/master/src/platform_impl/macos/ffi.rs) module is our designated dumping ground for constants and such.

src/platform_impl/macos/app_delegate.rs
```diff
@@ -79,3 +107,43 @@ extern "C" fn did_resign_active(this: &Object, _: Sel, _: id) {
     }
     trace!("Completed `applicationDidResignActive`");
 }
+
+extern "C" fn application_open_file(_this: &Object, _: Sel, _sender: id, filename: id) -> BOOL {
+    trace!("Triggered `application:openFile:`");
```
francesca64, line 112: Thanks for carrying on my proud legacy of having stuffed this module full of `trace` logging!

francesca64:
I actually wrote the macOS EL2 backend originally, even though I'm not involved in it much anymore. Feel free to ping me on macOS stuff until we find a new maintainer for it!

Speaking of, @ArturKovacs, would you be interested in maintaining the macOS backend? You seem very thoughtful, and that's the main qualification (besides having time to burn).

The implementation looks reasonable (I can give a proper review later), so it's just the design that's contentious here, wrt platform uniformity? If we don't believe this event will be used on other platforms, we could just add a macOS `EventLoopExt` method that gives you a receiver for this. The extension traits are basically made for covering things we can't provide a consistent cross-platform API for, and I'd have an easy time accepting this PR if we took that route.

While there's merit in exploring ways that problems like this could be solved outside of winit, I don't think that's necessary in this case. The implementation is small, self-contained, and straightforward, and seems like a normal thing for winit to cover.

chrisduerr:
> While there's merit in exploring ways that problems like this could be solved outside of winit, I don't think that's necessary in this case. The implementation is small, self-contained, and straightforward, and seems like a normal thing for winit to cover.

Does it? It's not related to windowing at all, so it seems far more out of scope than many other topics that have been denied.

francesca64:
@chrisduerr please address my comment above: https://github.com/rust-windowing/winit/pull/1759#discussion_r542742125

src/platform_impl/macos/app_delegate.rs
ArturKovacs: In an attempt to help resolving this, I'd like to add that I'm not upset with you @chrisduerr. I did feel like your approach was not very considerate but I'm willing to move on.

ArturKovacs:
> Speaking of, @ArturKovacs, would you be interested in maintaining the macOS backend?

This suprised me and made me very excited. I would definitely be interested, yes. 

> we could just add a macOS `EventLoopExt` method that gives you a receiver for this. 

I would be fine with this. When you say receiver, are you referring to `mpsc::channel` or just as in some way of informing the client?

francesca64:
Awesome! You've been memberized. And yeah, `mpsc::channel`, if that sounds good to you.

chrisduerr:
I think the idea proposed by @alvinhochun makes much more sense for winit. Worrying mainly about windowing needs and allowing applications to hook into the platform specific APIs themselves when necessary.

ArturKovacs:
The part that's not clear to me about using a channel is how would we avoid leaking memory if many of these "events" end up being unhandled. We cannot guarantee that the receiver will be pulled from.

In the first commmit of this PR I had an implementation which used the callback approach proposed by @alvinhochun but I didn't make an `EventLoopExt`, I instead used the `EventLoopWindowTargetExtMacOS` just because that already existed, and because I don't really know what the difference is. More importantly I made it such that the file open callback gets called on the same thread where the main event handling callback gets called, and this happens while processing all the other events. If we decide to use a callback it seems like a reasonable alternative to require that callback is `Send` and call it synchronously from `application:openFiles` and `application:openFile`. This would allow the user to decide how they want to indicate the success of the operation by, let's say a boolean return value.

Admittedly this would require some method by which the closure gets transferred to the application delegate. Although I have to say I'm a bit confused about what's running on separate threads because I would think that the main thread, the event loop callback thread, and the application delegate thread are the same.

francesca64:
Ah, that's a super good point. We could just only create the channel if the user calls the method to get the receiver, or are you thinking about the case where the user only pulls from it infrequently?

Either way, I think I actually prefer the sound of the callback approach, since your point about indicating success sounds like something we definitely want.

> because I don't really know what the difference is

`EventLoopExt` is typically used for adding different constructors (which is probably what we'd use if we wanted channels), whereas `EventLoopWindowTargetExt` gives you methods you can call from within `run`/`run_return`, which sounds like the right choice for using callbacks.

> Although I have to say I'm a bit confused about what's running on separate threads

Hey, me too! I once observed that the methods on our view were run asynchronously, which was scary... I don't really have a strong mental model of what happens once we cross the barrier into ObjC/AppKit land.

francesca64:
@chrisduerr please don't participate in design discussion until after addressing us in https://github.com/rust-windowing/winit/pull/1759#discussion_r544595255. I want to assume good faith, but it's difficult to do that if you won't communicate.

chrisduerr:
I'm sorry, but it's difficult to assume "good faith" from people just trying to stir up drama. I will not be participating in that.

But it doesn't seem like quality is any priority with winit's current leadership, which is quite unfortunate.

ArturKovacs:
I’ve implemented the callback method again, but I’m observing a strange behaviour. When I double click on a file in Finder there’s a delay of about 5 seconds between the application startup and the window appearing. If I just run the application directly, this delay isn’t there.

My first guess was that my implementation of the `openFiles` function was somehow blocking execution. But according to the Console log of the traces, this is not the case. The delay can be seen after the ``Completed `applicationDidFinishLaunching` `` message. Also the `SecTaskLoadEntitlements failed error=22` message always appears there when opening an associated file.

<details>

<summary>Click for the log</summary>

```
19/12/20 16:41:49,000 winit_open[0]: hello world
19/12/20 16:41:49,307 launchservicesd[75]: SecTaskLoadEntitlements failed error=22
19/12/20 16:41:49,318 launchservicesd[75]: SecTaskLoadEntitlements failed error=22
19/12/20 16:41:49,000 winit_open[0]: Triggered `viewDidMoveToWindow`
19/12/20 16:41:49,000 winit_open[0]: Completed `viewDidMoveToWindow`
19/12/20 16:41:49,000 winit_open[0]: Triggered `validAttributesForMarkedText`
19/12/20 16:41:49,000 winit_open[0]: Completed `validAttributesForMarkedText`
19/12/20 16:41:49,000 winit_open[0]: Triggered `validAttributesForMarkedText`
19/12/20 16:41:49,000 winit_open[0]: Completed `validAttributesForMarkedText`
19/12/20 16:41:49,000 winit_open[0]: Triggered `validAttributesForMarkedText`
19/12/20 16:41:49,000 winit_open[0]: Completed `validAttributesForMarkedText`
19/12/20 16:41:49,000 winit_open[0]: Locked shared state in `set_fullscreen`
19/12/20 16:41:49,000 winit_open[0]: Unlocked shared state in `set_fullscreen`
19/12/20 16:41:49,000 winit_open[0]: Triggered `windowDidBecomeKey:`
19/12/20 16:41:49,000 winit_open[0]: Completed `windowDidBecomeKey:`
19/12/20 16:41:49,471 appleeventsd[49]: SecTaskLoadEntitlements failed error=22
19/12/20 16:41:49,000 winit_open[0]: Triggered `validAttributesForMarkedText`
19/12/20 16:41:49,000 winit_open[0]: Completed `validAttributesForMarkedText`
19/12/20 16:41:49,000 winit_open[0]: Triggered `validAttributesForMarkedText`
19/12/20 16:41:49,000 winit_open[0]: Completed `validAttributesForMarkedText`
19/12/20 16:41:49,000 winit_open[0]: Triggered `validAttributesForMarkedText`
19/12/20 16:41:49,000 winit_open[0]: Completed `validAttributesForMarkedText`
19/12/20 16:41:49,000 winit_open[0]: Triggered `applicationDidBecomeActive`
19/12/20 16:41:49,000 winit_open[0]: Completed `applicationDidBecomeActive`
19/12/20 16:41:49,000 winit_open[0]: Triggered `application:openFiles:`
19/12/20 16:41:49,000 winit_open[0]: Completed `application:openFiles:`
19/12/20 16:41:49,000 winit_open[0]: Triggered `applicationDidFinishLaunching`
19/12/20 16:41:49,000 winit_open[0]: Completed `applicationDidFinishLaunching`
19/12/20 16:41:54,441 launchservicesd[75]: SecTaskLoadEntitlements failed error=22
19/12/20 16:41:54,442 launchservicesd[75]: SecTaskLoadEntitlements failed error=22
19/12/20 16:41:54,000 winit_open[0]: Triggered `mouseEntered`
19/12/20 16:41:54,000 winit_open[0]: Completed `mouseEntered`
19/12/20 16:41:55,000 winit_open[0]: Triggered `activationHackMouseMoved`
19/12/20 16:41:55,000 winit_open[0]: Completed `activationHackMouseMoved`
19/12/20 16:41:55,000 winit_open[0]: Triggered `mouseExited`
19/12/20 16:41:55,000 winit_open[0]: Completed `mouseExited`
19/12/20 16:41:57,000 winit_open[0]: Triggered `windowDidResignKey:`
19/12/20 16:41:57,000 winit_open[0]: Completed `windowDidResignKey:`
19/12/20 16:41:57,000 winit_open[0]: Triggered `applicationDidResignActive`
19/12/20 16:41:57,000 winit_open[0]: Completed `applicationDidResignActive`
19/12/20 16:43:25,000 winit_open[0]: Triggered `windowDidBecomeKey:`
19/12/20 16:43:25,000 winit_open[0]: Completed `windowDidBecomeKey:`
19/12/20 16:43:25,000 winit_open[0]: Triggered `applicationDidBecomeActive`
19/12/20 16:43:25,000 winit_open[0]: Completed `applicationDidBecomeActive`
19/12/20 16:43:26,000 winit_open[0]: Triggered `mouseEntered`
19/12/20 16:43:26,000 winit_open[0]: Completed `mouseEntered`
19/12/20 16:43:26,000 winit_open[0]: Triggered `mouseExited`
19/12/20 16:43:26,000 winit_open[0]: Completed `mouseExited`
19/12/20 16:43:27,000 winit_open[0]: Triggered `windowShouldClose:`
19/12/20 16:43:27,000 winit_open[0]: Completed `windowShouldClose:`
```

</details>

The log has considerably different contents when there’s no issue during startup (starting the application directly, using the `.app`).

<details>

<summary>Click for the log</summary>

```
20/12/20 14:31:27,000 winit_open[0]: hello world
20/12/20 14:31:27,317 launchservicesd[75]: SecTaskLoadEntitlements failed error=22
20/12/20 14:31:27,322 launchservicesd[75]: SecTaskLoadEntitlements failed error=22
20/12/20 14:31:27,000 winit_open[0]: Triggered `viewDidMoveToWindow`
20/12/20 14:31:27,000 winit_open[0]: Completed `viewDidMoveToWindow`
20/12/20 14:31:27,000 winit_open[0]: Triggered `validAttributesForMarkedText`
20/12/20 14:31:27,000 winit_open[0]: Completed `validAttributesForMarkedText`
20/12/20 14:31:27,000 winit_open[0]: Triggered `validAttributesForMarkedText`
20/12/20 14:31:27,000 winit_open[0]: Completed `validAttributesForMarkedText`
20/12/20 14:31:27,000 winit_open[0]: Triggered `validAttributesForMarkedText`
20/12/20 14:31:27,000 winit_open[0]: Completed `validAttributesForMarkedText`
20/12/20 14:31:27,000 winit_open[0]: Locked shared state in `set_fullscreen`
20/12/20 14:31:27,000 winit_open[0]: Unlocked shared state in `set_fullscreen`
20/12/20 14:31:27,000 winit_open[0]: Triggered `windowDidBecomeKey:`
20/12/20 14:31:27,000 winit_open[0]: Completed `windowDidBecomeKey:`
20/12/20 14:31:27,376 appleeventsd[49]: SecTaskLoadEntitlements failed error=22
20/12/20 14:31:27,000 winit_open[0]: Triggered `validAttributesForMarkedText`
20/12/20 14:31:27,000 winit_open[0]: Completed `validAttributesForMarkedText`
20/12/20 14:31:27,000 winit_open[0]: Triggered `validAttributesForMarkedText`
20/12/20 14:31:27,000 winit_open[0]: Completed `validAttributesForMarkedText`
20/12/20 14:31:27,000 winit_open[0]: Triggered `validAttributesForMarkedText`
20/12/20 14:31:27,000 winit_open[0]: Completed `validAttributesForMarkedText`
20/12/20 14:31:27,000 winit_open[0]: Triggered `applicationDidBecomeActive`
20/12/20 14:31:27,000 winit_open[0]: Completed `applicationDidBecomeActive`
20/12/20 14:31:27,000 winit_open[0]: Triggered `applicationDidFinishLaunching`
20/12/20 14:31:27,000 winit_open[0]: Completed `applicationDidFinishLaunching`
20/12/20 14:31:28,000 winit_open[0]: Triggered `keyDown`
20/12/20 14:31:28,000 winit_open[0]: Triggered `hasMarkedText`
20/12/20 14:31:28,000 winit_open[0]: Completed `hasMarkedText`
20/12/20 14:31:28,000 winit_open[0]: Triggered `hasMarkedText`
20/12/20 14:31:28,000 winit_open[0]: Completed `hasMarkedText`
20/12/20 14:31:28,000 winit_open[0]: Triggered `doCommandBySelector`
20/12/20 14:31:28,000 winit_open[0]: Completed `doCommandBySelector`
20/12/20 14:31:28,000 winit_open[0]: Completed `keyDown`
```

</details>

Using the Instrument app I recorded a trace where I could roughly locate the wait interval. The trace includes system calls and the time spent in them.

Screen Shot 2020-12-20 at 14 42 51

Here are the stack traces and durations for those long blocking invocations of these system calls:

First - 4.84 sec:

```
   0 libsystem_kernel.dylib mach_msg_trap
   1 libsystem_kernel.dylib mach_msg
   2 CoreFoundation __CFRunLoopServiceMachPort
   3 CoreFoundation __CFRunLoopRun
   4 CoreFoundation CFRunLoopRunSpecific
   5 AppKit _NSEventThread
   6 libsystem_pthread.dylib _pthread_body
   7 libsystem_pthread.dylib _pthread_start
   8 libsystem_pthread.dylib thread_start
```

Second (main thread) - 4.52 sec:

```
   0 libsystem_kernel.dylib mach_msg_trap
   1 libsystem_kernel.dylib mach_msg
   2 CoreFoundation __CFRunLoopServiceMachPort
   3 CoreFoundation __CFRunLoopRun
   4 CoreFoundation CFRunLoopRunSpecific
   5 HIToolbox RunCurrentEventLoopInMode
   6 HIToolbox ReceiveNextEventCommon
   7 HIToolbox _BlockUntilNextEventMatchingListInModeWithFilter
   8 AppKit _DPSNextEvent
   9 AppKit -[NSApplication _nextEventMatchingEventMask:untilDate:inMode:dequeue:]
  10 AppKit -[NSApplication run]
  11 winit_open winit::platform_impl::platform::event_loop::EventLoop$LT$T$GT$::run::he2b3290e18b06902
  12 winit_open winit::event_loop::EventLoop$LT$T$GT$::run::h89caa778cd324342
  13 winit_open winit_open::main::h990a67d0184d3864
  14 winit_open std::sys_common::backtrace::__rust_begin_short_backtrace::h8f105761de818030
  15 winit_open std::rt::lang_start::_$u7b$$u7b$closure$u7d$$u7d$::h5238b75dd9633a1d (.llvm.3847561433247225113)
  16 winit_open std::rt::lang_start_internal::hd0c760d47f593c0a
  17 winit_open main
  18 libdyld.dylib start
```

Last - 3.85 sec:

```
   0 libsystem_kernel.dylib kevent_qos
   1 libdispatch.dylib _dispatch_mgr_invoke
   2 libdispatch.dylib _dispatch_mgr_thread
```

At this point I'm guessing the answer may lie somewhere in how winit's macos backend works. Do you have an idea about this @francesca64?

ArturKovacs:
Following [this](https://github.com/sveinbjornt/Platypus/issues/110) thread it really seems that the delay I experienced was unrelated to winit. The window now appears instantly if I open an associated file from Finder. For me the fix was two part. First I ran
```
defaults write io.github.arturkovacs.winit_open NSQuitAlwaysKeepsWindows -bool false
```
But after this, the delay was still present so then I removed the saved state folder using `AppCleaner` as it was shown in the thread above. This resolved it for me.

ArturKovacs:
I pushed a new commit which implements the callback design and I updated the example to use this latest commit. Example: https://github.com/ArturKovacs/winit_open_file

src/platform_impl/macos/app_delegate.rs
```diff
@@ -36,10 +48,28 @@ lazy_static! {
...
+            sel!(application:openFiles:),
+            application_open_files as extern "C" fn(&Object, Sel, id, id),
+        );
+
         AppDelegateClass(decl.register())
     };
 }
 
+/// Copies the contents of the ns string into a `String` which gets returned.
+fn ns_string_to_rust(ns_string: id) -> String {
```
francesca64, line 65:
```suggestion
unsafe fn ns_string_to_rust(ns_string: id) -> String {
```
I feel marking the function itself as `unsafe` better communicates that this function is unsafe to call/use (versus being safe to use but containing unsafe implementation details)

Now, you might've noticed that this module is actually full of functions that have an inner `unsafe` block, seemingly contradicting my advice! That's because marking those `extern` functions as `unsafe` would change their signature, which AFAIK wouldn't play nice with `add_method`.

src/platform_impl/macos/app_state.rs
```diff
@@ -248,6 +251,14 @@ impl AppState {
         }));
     }
 
+    pub fn set_file_open_callback<F>(callback: Option<F>)
+    where
+        F: FnMut(Vec<PathBuf>) -> FileOpenResult + Send + 'static,
+    {
+        *HANDLER.file_open_callback.lock().unwrap() =
+            callback.map(|c| Box::new(c) as Box<dyn FnMut(Vec<PathBuf>) -> FileOpenResult + Send>);
```
francesca64, line 259:
```suggestion
            callback.map(|c| Box::new(c) as _);
```
I believe you can elide some or all of the type after `as` here, though I leave it at your discretion if that's better or worse for readability.

src/platform_impl/macos/app_delegate.rs
```diff
@@ -79,3 +109,36 @@ extern "C" fn did_resign_active(this: &Object, _: Sel, _: id) {
...
+    } else {
+        NO
+    }
+}
+
+extern "C" fn application_open_files(_this: &Object, _: Sel, sender: id, filenames: id) {
+    trace!("Triggered `application:openFiles:`");
+    let filenames_len = unsafe { filenames.count() };
+    let mut filenames_vec = Vec::with_capacity(filenames_len as usize);
+    for i in 0..filenames_len {
```
francesca64, line 130:
```rust
let filenames_vec = (0..filenames_len)
    .map(|i| {
        let filename = unsafe { filenames.objectAtIndex(i) };
        ns_string_to_rust(filename).into()
    })
   .collect();
```
IIRC this has the benefits of using `with_capacity` as well as being slightly more concise. Though, iterators still don't seem to be used a ton throughout winit, which is definitely partially my fault (...to be frank, at the peak of my activity, I didn't really know how to use iterators!)

In any case, I'll leave that up to preference as well!

francesca64 (changes requested):
Sorry for the delay! This looks great, but I've thrown a few small nitpicks your way for good measure.

src/platform_impl/macos/app_delegate.rs
ArturKovacs: > I feel marking the function itself as unsafe better communicates that this function is unsafe to call/use

You are completely right

> marking those `extern` functions as `unsafe` would change their signature

Oh I had no idea about this. A quick internet search didn't yield any information; where could I learn more about this?

src/platform_impl/macos/app_state.rs
ArturKovacs: Yes, this works!

src/platform_impl/macos/app_delegate.rs
ArturKovacs: What you suggest seems more idiomatic, so I'll go with that.

ArturKovacs:
> Sorry for the delay!

No worries! I think I addressed all the feedback. 😊

src/platform_impl/macos/app_state.rs
```diff
@@ -318,6 +329,27 @@ impl AppState {
         HANDLER.events().append(&mut wrappers);
     }
 
+    pub fn open_files(paths: Vec<PathBuf>) -> FileOpenResult {
+        let result = catch_unwind(|| {
```
madsmtm, line 333: Good call to make the external callback panic safe!

src/platform_impl/macos/app_delegate.rs
```diff
@@ -36,10 +48,26 @@ lazy_static! {
...
+            sel!(application:openFiles:),
+            application_open_files as extern "C" fn(&Object, Sel, id, id),
+        );
+
         AppDelegateClass(decl.register())
     };
 }
 
+/// Copies the contents of the ns string into a `String` which gets returned.
+unsafe fn ns_string_to_rust(ns_string: id) -> String {
```
madsmtm, line 65: Maybe move this to `macos/util/mod.rs`, next to `ns_string_id_ref` - they're kinda polar opposites, and both useful to have in general!

src/platform_impl/macos/app_state.rs
```diff
@@ -318,6 +329,27 @@ impl AppState {
...
+                (callback)(paths)
+            } else {
+                FileOpenResult::Failure
+            }
+        });
+        match result {
+            Ok(result) => result,
+            Err(e) => {
+                error!(
+                    "Paniced when trying to execute open files callback: {:?}",
```
madsmtm, line 345: Paniced -> Panicked

src/platform_impl/macos/app_delegate.rs
```diff
@@ -36,10 +48,26 @@ lazy_static! {
...
 
+        decl.add_method(
+            sel!(application:openFile:),
+            application_open_file as extern "C" fn(&Object, Sel, id, id) -> BOOL,
+        );
+        decl.add_method(
+            sel!(application:openFiles:),
+            application_open_files as extern "C" fn(&Object, Sel, id, id),
+        );
+
```
madsmtm, line 59: Did you verify that the default behaviour, when not implementing these on the delegate, matches the behaviour we have now with `event_loop.set_file_open_callback(None)`?

madsmtm (changes requested):
Annoyingly, NSApplication doesn't emit a "Notification" for this that one could just register to receive, so we kinda have to handle this in `winit` somehow!

So I agree that we should find a better way to expose these things to the user, I don't think this will be the first case where we'll need to do this!

But for now, this provides us with an implementation that works and allows people to build something, and simply by being in the codebase, it very clearly illustrates the need for a better solution!

madsmtm:
Also, I would say your `winit_open_file` example is flawed; most applications should either open a new window for every file:
```rust
let window_target = event_loop.window_target();
event_loop.set_file_open_callback(Some(|paths: Vec<PathBuf>| {
    for path in paths.iter() {
        let filename = path.as_os_str().to_owned().into_string().unwrap();
        let title = format!("> Opened: '{}'", filename);
        WindowBuilder::new().with_title(title).build(&event_loop).unwrap();
    }
    FileOpenResult::Success
}));
```
Or open a single window with all the files (like Preview.app does):
```rust
let window_target = event_loop.window_target();
event_loop.set_file_open_callback(Some(|paths: Vec<PathBuf>| {
    let mut title = "> Opened: ".into();
    for path in paths.iter() {
        let filename = path.as_os_str().to_owned().into_string().unwrap();
        title.push_str(&filename);
    }
    WindowBuilder::new().with_title(title).build(&event_loop).unwrap();
    FileOpenResult::Success
}));
```

(Terrible example code, untested!)

ArturKovacs:
Thank you @madsmtm. I will address your feedback properly at a later time but I just wanted to note that it's probably a good idea to merge #1853 before this so that this can also use the "application stopping user callback panic".

src/platform_impl/macos/app_delegate.rs
ArturKovacs: I can confirm that it does *not* match.

When these are not implemented, the following message is shown when attempting to open a file.

winit_open - no delegate function defined

This message isn't shown when the functions are implemented on the delegate, regardless of whether the functions indicate success or failure. I'm not sure how or if we want to deal with this. An argument of leaving it as it is, is that the `Info.plist` file must declare that the application supports opening certain file types for this to work anyway. IIRC the application cannot be selected from the "open with" list in case the `Info.plist` doesn't declare a matching file type.

ArturKovacs:
I also realized that `event_loop.set_file_open_callback(None)` does not complie because the compiler cannot infer the type for `Option<F>`.

I'm wondering if we should require the boxed type here that we later convert to anyways. Like this

```rust
fn set_file_open_callback(
    &self,
    callback: Option<Box<dyn FnMut(Vec<PathBuf>) -> FileOpenResult + Send + 'static>>
);
```

src/platform_impl/macos/app_delegate.rs
francesca64: > where could I learn more about this?

I don't really know where you can find more about this, but marking one of those `extern` functions as `unsafe extern` causes it to no longer type-check when passed to `add_method`.

francesca64:
Ah, that's always fun. That revised signature seems like the only solution, so it looks good to me!

I'll also be sure to look at #1853!

src/platform/macos.rs
```diff
@@ -205,12 +206,34 @@ impl MonitorHandleExtMacOS for MonitorHandle {
...
+    ///
+    /// This for example happens when the user double-clicks on a file in Finder and the
+    /// file's type is associated with this application.
+    ///
+    /// The callback function receives the absolute path to the target files as an argument and it should
+    /// indicate the success of the operation in its return value. See: `application:openFiles:` in AppKit.
+    ///
+    /// Other systems usually provide the path as a program argument.
+    /// See: `std::env::args`.
+    fn set_file_open_callback<F>(&self, callback: Option<F>)
```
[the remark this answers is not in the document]
francesca64: Great catch!

I'm seconding the creation of `EventLoopExtMacOS`, though it seems like it'd be useful to have the method on both?

src/platform_impl/macos/app_delegate.rs
francesca64: It's probably more correct/expected to preserve the old behavior, though I imagine that'd be quite difficult to do? I don't think this is too big a deal either way.

Maybe it'd be helpful for `open_files` to emit a log message when it's triggered but the callback is `None`?

src/platform/macos.rs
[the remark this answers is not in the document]
ArturKovacs: > I'm fairly certain this should be set on EventLoopExtMacOS

I like the intention to communicate that this should be called before calling `run`, but I agree with @francesca64 in that this should also be callable while the event loop is running. So I believe it should be implemented on `EventLoopWindowTarget`.

However I don't see why it would be beneficial to have it on both since `EventLoop` implements `Deref<Target=EventLoopWindowTarget>` and as such, all functions implemented on `EventLoopWindowTarget` can be invoked on `EventLoop` as well.

src/platform_impl/macos/app_delegate.rs
ArturKovacs: In my latest commit I changed the implementation such that the default behavior (i.e. when the callback is set to `None`) is identical to not having the delegate methods implemented.

src/platform_impl/macos/event_loop.rs
```diff
@@ -13,25 +13,35 @@ use crate::{
...
+        },
     },
 };
 
 pub struct EventLoopWindowTarget<T: 'static> {
     pub sender: mpsc::Sender<T>, // this is only here to be cloned elsewhere
     pub receiver: mpsc::Receiver<T>,
+    pub delegate: IdRef,
+    pub delegate_with_file_open: IdRef,
 }
```
ArturKovacs, line 34: I wasn't entirely sure about adding the application delegates here, but this was the only reasonable way I could make these accessible at `set_file_open_callback()`.

src/platform/macos.rs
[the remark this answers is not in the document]
madsmtm: I don't really see when you would call `set_file_open_callback`, since . Can you give an example of where this would be useful?

If not, I would probably argue it should be `new_with_file_open_callback` or similar which you call on creation of the EventLoop (or maybe have an `EventLoopBuilder` at some point in the future), such that you can't change the callback afterwards. This would probably also allow getting rid of a `Mutex` in the `Handler` later on.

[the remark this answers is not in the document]
madsmtm: It would also mean you wouldn't need to create and store two delegate instances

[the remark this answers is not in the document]
ArturKovacs: I have to admit that your arguments are really compelling. I'll wait a bit so that @francesca64 has time to respond, but otherwise I'm inclined to create that `new_with_file_open_callback` and remove `set_file_open_callback`.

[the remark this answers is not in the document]
francesca64: Seconding only having `new_with_file_open_callback`; it sounds like a concise and correct API!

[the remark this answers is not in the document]
ArturKovacs: I hit a problem. The window needs an `EventLoop` when it's constructed. So `new_with_file_open_callback` must be called before the window exists, which leads to a very awkward solution if one desires to use anything depending on the event loop in the callback. The code below is the nicest approach I could come up with, and it's a mess. The `Mutex<Option<>>` part of the type wouldn't be needed if this could be called after the window was created, so that the closure could directly take a clone of `Arc<Window>`.
```rust
let window_ref: Arc<Mutex<Option<Window>>> = Default::default();
let event_loop;
{
    let window_ref = window_ref.clone();
    event_loop = EventLoop::<()>::new_with_file_open_callback(move |paths: Vec<PathBuf>| {
        let guard = window_ref.lock().unwrap();
        let window = match &*guard {
            Some(window) => window,
            None => return FileOpenResult::Failure,
        };
        for path in paths.iter() {
            open_file(window, path.as_ref());
        }
        FileOpenResult::Success
    });
}
``` 

Probably the best is to have a `set_file_open_callback` on the `EventLoop`.

ArturKovacs:
The odd delay during startup keeps coming up. It's what I described earlier here: https://github.com/rust-windowing/winit/pull/1759#issuecomment-748612938

I'm suspecting this has to do with restoring windows because the "savedState" folder - that I need to delete to temporarily solve the issue - seems to contain information required to restore the windows. Although this is just speculation. In an attempt to set the [`restorable`](https://developer.apple.com/documentation/appkit/nswindow/1526255-restorable?language=objc) property to `NO`, I tried `msg_send[*ns_window, setRestorable: NO]` ) in the `create_window` function, but it didn't have an effect. I also tried overriding `restoreWindowWithIdentifier` on our NSApplication class, such that always fails, but that resulted in getting an error message window on every startup.

This could possibly be related to the fact that I'm running OS X 10.11 which is not supported anymore.

src/platform/macos.rs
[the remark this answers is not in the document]
madsmtm: Hmm, I think I would argue that this behaviour is unidiomatic MacOS behaviour; one should almost always create new windows from the given files.

In the odd cases where you want to modify previously created windows, it is, as you've demonstrated, still possible, albeit requires a `Mutex` (which makes sense, since we're throwing a `Mutex` out in our `Handler`).

Though yeah, it dawns to me that we can't even access the event loop in `new_with_file_open_callback`...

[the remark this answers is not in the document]
madsmtm: Also really dislike that we have to create two different delegates just for this functionality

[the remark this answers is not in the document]
madsmtm: But I think I can accept it being like this for now (maybe with a few comments about what we've talked about here), at a later point `winit` really needs better callback support!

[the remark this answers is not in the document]
ArturKovacs: > I think I can accept it being like this for now

Just to be clear, are you saying you'd be okay with having `set_file_open_callback` on the `EventLoop`?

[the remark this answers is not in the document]
madsmtm: Yup, sorry for being unclear

francesca64:
Sorry, I totally missed your comment about the delay before!

I wouldn't expect using an older macOS version to be the problem. Does this still happen even if you comment out everything in `application_open_file`?

ArturKovacs:
I tried commenting-out both `application_open_file` and `application_open_file`. The delay is still there. Also the delay happens before either of these functions is called, which can be seen from the logs shown in https://github.com/rust-windowing/winit/pull/1759#issuecomment-748612938

ArturKovacs:
I think, @madsmtm, you would make an excellent maintainer for macOS. I'm not planning to step down but there's quite a lot in the backlog, and we could really speed things up if you had the ability to approve PRs. Plus the more, the merrier anyways. Would you be interested? What do you think @francesca64?

madsmtm:
Thanks for the suggestion(!); however, I have lot of things in my own backlog, and can't take on further roles or responsibilities right now.

Maybe in a month or so my situation will have settled a bit more, I'll gladly reconsider it then!

ArturKovacs:
I rebased these changes on top of the master and I don't experience the delay at startup anymore.

francesca64:
Weird! There's only 3 commits that could've caused that, right? It might be worth bisecting to verify that it's actually fixed and not just an issue that comes and goes.

ArturKovacs:
I'm assuming you mean these three:
629cd86c7cc674c40c1f20b992bdd13784f357dd
ffe2143d14b1c0877f6d6fb32c2f4896d30162fe
4192d04a53202c199f94d1b7d883a34c9ad09272

Although these were also added between creating this PR and rebasing:
ba704c4eb49638da7516250d26adee35ee0bc9a4
0487876826819936f52f26d192627bd552867886
0d634a0061224a4accb51b031d1a85320d690a98

I'm not sure how this could be bisected. If I understand correctly, all changes in this PR are on top of 629cd86c7cc674c40c1f20b992bdd13784f357dd, so when checking out a commit before this PR, I can't handle the "file open event". The best idea I had so far is to squash these changes on a local repo and rebase that on top of each of those commits and for each, test if the delay is there.

ArturKovacs:
I still have a copy of the not-rebased repo locally if that helps. I'm not sure what to do here. I guess the question is how do we become reasonably sure that the delay is fixed. (The difficulty is that it may be spurious.) @madsmtm and @francesca64 this might be strange to ask but have you tested it locally? Has the delay never come up for you?

ArturKovacs:
I learned that I've been using the word "spurious" incorrectly 😬

I seem to remember that @madsmtm you mentioned somewhere that you intend to spend some time with this. (Maybe I only dreamt that :D) I just want to add that it would be very helpful because due to this, the macOS release of [emulsion](https://github.com/ArturKovacs/emulsion) is really lacking.

Korne127:
Hey; will this still be merged at some point?
I read through this whole conversation and it seems like there has been a consensus to implement this, but it has now been unchanged for two years.
I don't know if it has just been forgotten and overshadowed by other things or if there are reasons why it's in this state and won't be merged (for now).
But I can just say that it looks like much work has been put into this and I'd definitely appreciate this feature to be implemented and file open events to be possible on macOS.

raphamorim:
> Hey; will this still be merged at some point?
I read through this whole conversation and it seems like there has been a consensus to implement this, but it has now been unchanged for two years.
I don't know if it has just been forgotten and overshadowed by other things or if there are reasons why it's in this state and won't be merged (for now).
But I can just say that it looks like much work has been put into this and I'd definitely appreciate this feature to be implemented and file open events to be possible on macOS.

+ 1

And if needs with help testing and updating this PR, I would be glad to take it as well cc @ArturKovacs

madsmtm:
I've posted some information on how I think the design of this feature should be in https://github.com/rust-windowing/winit/issues/1751#issuecomment-1623169493. This does not entirely resolve usability concerns, though I think those are less important than getting something that at least works.

I'm truly sorry for never getting this in, but I think the approach I've suggested there should be a good way forward (and could also set precedent for how we solve similar issues).

I'll close this PR for now though, since a lot of the discussion + code is out of date, and I want to make space for a new implementation.

PR #1761, opened 2020-11-10
[RFC] event: Unwrap user event in map_nonuser_event Err()

MarijnS95:
:warning:  **WARNING** :warning: : This is a _breaking_ API change!

Currently separating an `Event<T>` into a user event `T` and `Event<()>` is not ergonomic: when `map_nonuser_event` returns `Err()` the caller "knows" this can only contain `UserEvent(T)`, yet has to write additional unpacking logic to extract the event, with an unnecessary `unreachable!()`.

Solve this by returning `T` directly in the Err case, instead of `Event<'a, T>`.

- [x] Compilation warnings were addressed (N/A: there were no compiler warnings in the first place?)
  - How about introducing a user event to `multithreaded.rs`? That is the exact use case we need this for: handle user events (from the thread) in the event loop, and send nonuser events into the thread.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - **Breaking:** On Web, remove the `stdweb` backend.
 - Added `Window::focus_window`to bring the window to the front and set input focus.
 - On Wayland and X11, implement `is_maximized` method on `Window`.
+- **Breaking:** `map_nonuser_event` now returns the unwrapped user event `T` in `Err()`.
 
 # 0.25.0 (2021-05-15)
 
--- a/src/event.rs
+++ b/src/event.rs
@@ -143,10 +143,10 @@ impl<T: Clone> Clone for Event<'static, T> {
 }
 
 impl<'a, T> Event<'a, T> {
-    pub fn map_nonuser_event<U>(self) -> Result<Event<'a, U>, Event<'a, T>> {
+    pub fn map_nonuser_event<U>(self) -> Result<Event<'a, U>, T> {
         use self::Event::*;
         match self {
-            UserEvent(_) => Err(self),
+            UserEvent(user_event) => Err(user_event),
             WindowEvent { window_id, event } => Ok(WindowEvent { window_id, event }),
             DeviceEvent { device_id, event } => Ok(DeviceEvent { device_id, event }),
             NewEvents(cause) => Ok(NewEvents(cause)),


chrisduerr (changes requested):
Why would anyone use `map_nonuser_event` to separate `Event<T>` into `Event<()>` and `T`? The purpose is to map `Event<T>` to `Event<()>`, why would that ever return the user event when you're clearly trying to get rid of it?

MarijnS95:
> when you're clearly trying to get rid of it?

@chrisduerr  Who says that I'm trying to get rid of them? I want to separate these events out and `Result` is clearly used here to give me that, for lack of an `Either` type.

> The purpose is to map `Event<T>` to `Event<()>`, why would that ever return the user event

It... already does? Just not in a very friendly way?

As explained my usecase is identical to what is going on in `multithreaded.rs`. Right now I have to write a panicking abomination like:

```rust
if let winit::event::Event::UserEvent(thread_event) = event {
    // Received message from thread
} else if let Some(event) = event.to_static() {
    match event.map_nonuser_event() {
        Ok(event) => {
            // Send event to thread
            let _ = event_send.send(event);
        }
        Err(_) => unreachable!(),
    }
}
```
Or
```rust
match event.map_nonuser_event() {
    Ok(event) => {
        // Send event to thread
        if let Some(event) = event.to_static() {
            let _ = event_send.send(event);
        }
    }
    Err(user_event) => match user_event {
        winit::event::Event::UserEvent(thread_event) => {
            // Received message from thread
        }
        _ => unreachable!(),
    }
}
```
Or any other variation of `match`/`if let` you can come up with.

With this change I can get rid of that unnecessary `unreachable!()` case and simply write:
```rust
match event.map_nonuser_event() {
    Ok(event) => {
        // Send event to thread
        if let Some(event) = event.to_static() {
            let _ = event_send.send(event);
        }
    }
    Err(thread_event) => match user_event {
        // Received message from thread
    }
}
```

chrisduerr:
> As explained my usecase is identical to what is going on in multithreaded.rs

Which doesn't use `map_nonuser_event` event though?

It seems like your problem is fundamentally with the combination of `to_static` and `map_nonuser_event`? Can you provide a minimal working example?

MarijnS95:
> Which doesn't use `map_nonuser_event` event though?

If it were to use user events, then it'd need it to send an `Event<'static, ()>` into the thread. Hence my request in the PR description to add such an event to this particular example.

> It seems like your problem is fundamentally with the combination of `to_static` and `map_nonuser_event`? Can you provide a minimal working example?

There is not really a "problem" that this PR attempts to tackle, merely making the resulting `Err()` value from `map_nonuser_event` more readily usable. What do you expect to see in a minimal working example that is not covered by the code samples above?

madsmtm:
This is redundant after https://github.com/rust-windowing/winit/pull/3694 (part of `v0.31`), as `UserEvent` has been removed entirely in favour of a wake-up mechanism instead.

PR #1764, opened 2020-11-12
Added request_user_attention to Window

maxded:
Moved macOS' `WindowExt::request_user_attention` and X11's `WindowsExt::set_urgent` methods to `Window::request_user_attention`, implementing it for macOS, X11, and Windows. 


chrisduerr (changes requested):
What's the point of calling `request_user_attention` over requesting urgency on X11?

maxded:
Completely overlooked `set_urgent` on X11, my bad. Replaced X11's `set_urgent` with `request_user_attention`, which now also takes an Option to be able to "turn off" the requesting of the user's attention.

src/window.rs
```diff
@@ -874,3 +896,15 @@ pub enum Theme {
...
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum RequestUserAttentionType {
+    Critical,
+    Informational,
+}
+
+impl Default for RequestUserAttentionType {
+    fn default() -> Self {
+        RequestUserAttentionType::Critical
```
kchibisov, line 908: What's the point of using `Critical` by default, I think `Informational` should be default, since it's more relaxed

src/window.rs
```diff
@@ -682,6 +682,28 @@ impl Window {
...
+    /// - **MacOS:** `RequestUserAttentionType::Critical` bounces the dock icon until the
+    /// application is in focus. `RequestUserAttentionType::Informational` bounces the   
+    /// dock icon once.
+    ///
+    /// - **Windows:** `RequestUserAttentionType::Critical` flashes both the window and the
+    /// taskbar button until the application is in focus. `RequestUserAttentionType::Informational`
+    /// flashes the taskbar button until the application is in focus.
+    ///
+    /// - **X11:** `RequestUserAttentionType::Critical` and `RequestUserAttentionType::Informational`
+    /// provide the exact same result, which is determined by the WM.
```
kchibisov, line 702: We should move enum variants description to the corresponding enum variants , and not list it on the method.

src/window.rs
```diff
@@ -682,6 +682,28 @@ impl Window {
     pub fn set_ime_position<P: Into<Position>>(&self, position: P) {
         self.window.set_ime_position(position.into())
     }
+
+    /// Requests user attention to the window, this has no effect if the application
+    /// is already focused.
```
kchibisov, line 687: You should document what `None` does, also should we do that via `None` or via special method, like `unset_user_attention` or similar.

I wonder what others think about it, @chrisduerr , @alvinhochun , @francesca64 ?

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Unreleased
 
+- Added `request_user_attention` method to `Window`.
+- **Breaking:** On macOS, removed `WindowExt::request_user_attention`, use `Window::request_user_attention`.  
+- **Breaking:** On X11, removed `WindowExt::set_urgent`, use `Window::request_user_attention`.
```
kchibisov, line 5: Could you append changes and not add them to the head, thx?

src/platform_impl/macos/window.rs
```diff
@@ -977,6 +981,33 @@ impl UnownedWindow {
...
+            match request_type {
+                Some(request_type) => {
+                    shared_state.request_user_attention_id =
+                        Some(NSApp().requestUserAttention_(match request_type {
+                            RequestUserAttentionType::Critical => {
+                                NSRequestUserAttentionType::NSCriticalRequest
+                            }
+                            RequestUserAttentionType::Informational => {
+                                NSRequestUserAttentionType::NSInformationalRequest
+                            }
```
kchibisov, line 998: Could you `map` request_type to the corresponding platform variants and not have inline match?

src/platform_impl/windows/window.rs
```diff
@@ -621,6 +621,41 @@ impl Window {
...
+                Some(req_type) => winuser::FLASHWINFO {
+                    cbSize: mem::size_of::<winuser::FLASHWINFO>() as UINT,
+                    hwnd: window.0,
+                    dwFlags: match req_type {
+                        RequestUserAttentionType::Critical => {
+                            winuser::FLASHW_ALL | winuser::FLASHW_TIMERNOFG
+                        }
+                        RequestUserAttentionType::Informational => {
+                            winuser::FLASHW_TRAY | winuser::FLASHW_TIMERNOFG
+                        }
```
kchibisov, line 639: Could your request to a platform specific variants instead of inlining it here? You should do that
for `uCount` and `dwFlags` for both `None` and `Some` pathes.

kchibisov (changes requested):
In general, I'm +1 on the proposed changes and it should simplify handling downstream.

alvinhochun:
I'm thinking that "un-requesting" user attention seems like a rare use case -- how often would an application want get attention from the user, and then somehow change its mind to not want user attention? So perhaps winit shouldn't even need an API to do this.

That said, I'm not sure how the "urgency hint" works on X11. It looks like it must be explicitly unset by the application, which is kind of not what I would expect having been used to the behaviour on Windows. On Windows, the flashing initiated by `FlashWindowEx` is automatically ended when the user bring the window in focus. (AFAIK if you call `FlashWindowEx` with `FLASHW_STOP`, you can stop the flashing but the taskbar button stays highlighted, meaning you wouldn't really be able to "un-request" user attention.)

On the surface, it looks like that implementing `request_user_attention` on X11 with the "urgency hint" may not be very appropriate.

kchibisov:
>I'm thinking that "un-requesting" user attention seems like a rare use case -- how often would an application want get attention from the user, and then somehow change its mind to not want user attention? So perhaps winit shouldn't even need an API to do this.

I think the purpose is to unrequest once user focuses the app back, since from my memory it gets triggered faster than wm decides that something isn't urgent anymore. Also for multi window application you can only unset attention when only focusing specific window in the app?

Also, I don't see anything wrong with modeling X11's urgency via new `request_user_attention`, since it'll make API more crossplatform friendly downstream, instead of having a bunch of cfg's around, like we have now in e.g. alacritty for macOS/X11.

cc @murarth  wrt X11 parts and whether it makes sense to set urgency via `request_user_attention`.

kchibisov:
Also, I think alacritty will be a client that will unset urgency on demand in some cases, it can get external request to disable urgency(e.g. it got escape sequence to disable urgency), thus it should somehow communicate it to the wm, so there's a use case for it.

alvinhochun:
Here's the thing, what should happen if the user don't un-request user attention after the window is focused and then the user switch to another window? On both Windows and macOS the user attention would already be automatically un-requested by the system, but on X11 it would continue to have the urgent hint and keep showing up as urgent depending on the WM. This seems to be a very notable discrepancy.

In my opinion, if `request_user_attention` is made to be a cross-platform API, it should try its best to align the behaviour on different platforms. This means either automatically unsetting the urgent hint on window focus for X11, or re-request user attention on window unfocus for Windows and macOS if the user code did not un-request user attention.

> Also, I think alacritty will be a client that will unset urgency on demand in some cases, it can get external request to disable urgency(e.g. it got escape sequence to disable urgency), thus it should somehow communicate it to the wm, so there's a use case for it.

If it is the case, then I guess it does makes sense to provide the ability to un-request user attention.

src/platform_impl/android/mod.rs
```diff
@@ -485,6 +485,8 @@ impl Window {
 
     pub fn set_ime_position(&self, _position: Position) {}
 
+    pub fn request_user_attention(&self, _request_type: Option<window::RequestUserAttentionType>) {}
```
filnet, line 488: you could add the same warning as found in the ios case

src/platform_impl/linux/mod.rs
```diff
@@ -418,6 +418,16 @@ impl Window {
         x11_or_wayland!(match self; Window(w) => w.set_ime_position(position))
     }
 
+    #[inline]
+    pub fn request_user_attention(&self, _request_type: Option<RequestUserAttentionType>) {
+        match self {
+            #[cfg(feature = "x11")]
+            &Window::X(ref w) => w.request_user_attention(_request_type),
+            #[cfg(feature = "wayland")]
+            _ => (),
```
filnet, line 427: add same warning as in the ios case ?

src/platform_impl/web/window.rs
```diff
@@ -268,6 +270,11 @@ impl Window {
         // Currently a no-op as it does not seem there is good support for this on web
     }
 
+    #[inline]
+    pub fn request_user_attention(&self, _request_type: Option<RequestUserAttentionType>) {
+        // Currently an intentional no-op
```
filnet, line 275: warning ?

kchibisov: I think it's done this way due to consistency with the backend other handlings of such things? I don't see a reason to diverge from existing backend code style...

maxded: Jup, I stayed consistent with the different backends, which might look a bit weird in this PR but overall is probably better.

maxded:
Instead of using the urgency hint on X11 we could also directly send the `NET_WM_STATE_DEMANDS_ATTENTION` event. 

From the spec: "_NET_WM_STATE_DEMANDS_ATTENTION indicates that some action in or with the window happened. For example, it may be set by the Window Manager if the window requested activation but the Window Manager refused it, or the application may set it if it finished some work. This state may be set by both the Client and the Window Manager. It should be unset by the Window Manager when it decides the window got the required attention (usually, that it got activated)."

https://specifications.freedesktop.org/wm-spec/wm-spec-1.4.html

Here it becomes the WM responsibility to "unset" requiring the user attention. Not very knowledgeable about X11 so I could be wrong here :)

src/window.rs
maxded: I tried to stay consistent with other APIs like `set_window_icon`, which when providing `None` calls an `unset` method internally. It's both consistent and more ergonomic than providing a separate `unset_user_attention` I believe.

kchibisov: I think we can go with `None` then.

src/platform_impl/windows/window.rs
maxded: How do you mean exactly?

kchibisov:
```rust
    let dwFlags = request_type
        .map(|ty| match ty {
            RequestUserAttentionType::Critical => winuser::FLASHW_ALL | winuser::FLASHW_TIMERNOFG,
            RequestUserAttentionType::Informational => {
                winuser::FLASHW_TRAY | winuser::FLASHW_TIMERNOFG
            }
        })
        .unwarp_or(winuser::FLASHW_STOP);
```

chrisduerr:
> Also, I think alacritty will be a client that will unset urgency on demand in some cases, it can get external request to disable urgency(e.g. it got escape sequence to disable urgency), thus it should somehow communicate it to the wm, so there's a use case for it.

It doesn't need to retract urgency requested previously though. So this can be 100% handled in Alacritty and there's absolutely no need for an API that can retract urgency in Alacritty.

> Here's the thing, what should happen if the user don't un-request user attention after the window is focused and then the user switch to another window? On both Windows and macOS the user attention would already be automatically un-requested by the system, but on X11 it would continue to have the urgent hint and keep showing up as urgent depending on the WM. This seems to be a very notable discrepancy.

Are you sure about that? Urgency is usually unset by the window manager as soon as an application is focused. Have you actually tested this? I'd be very curious which window manager doesn't unset urgency by itself.

> Instead of using the urgency hint on X11 we could also directly send the NET_WM_STATE_DEMANDS_ATTENTION event.

Urgency hints as defined by ICCCM should be more widely supported than the EWMH spec.

kchibisov:
>Are you sure about that? Urgency is usually unset by the window manager as soon as an application is focused. Have you actually tested this? I'd be very curious which window manager doesn't unset urgency by itself.

You can try and test it in alacritty, but we do set `set_urgent(false)`.

chrisduerr:
> You can try and test it in alacritty, but we do set set_urgent(false).

Yeah, but it's useless. At least on i3 urgency is unset by the window manager anyways. Removing this in Alacritty doesn't change anything about its behavior.

kchibisov:
>Yeah, but it's useless. At least on i3 urgency is unset by the window manager anyways. Removing this in Alacritty doesn't change anything about its behavior.

And you've tested it, right?

chrisduerr:
> And you've tested it, right?

Of course, I've verified it again just to make sure.

kchibisov:
Just by looking into X11 ICCCM docs related to urgency, there's no notion that wm must unset urgency, thus the call unsetting it is required.

chrisduerr:
> Just by looking into X11 ICCCM docs related to urgency, there's no notion that wm must unset urgency, thus the call unsetting it is required.

That is correct and it is generally done this way (which is why Alacritty does it). Though that does not change anything about the fact that it's not really necessary for Alacritty. Just doing it on focus automatically internally would be perfectly fine for our usecases.

src/window.rs
maxded: Good point changed it!

src/window.rs
maxded: Moved it all to the enum variants of `RequestUserAttentionType` :)

CHANGELOG.md
maxded: Oh whoops, my bad. Changed it.

kchibisov:
As I said earlier, there's no guarantee that the urgency hint will be unset by wm( or I can't really find anything related to that). Thus we should have a way to unset it somehow, providing `None` to do so could work fine for now. We may document that `None` doesn't work on macOS, it's fine to do so that way.

Also, I'm still not sure how it'll work on Wayland in a future(if they'll ever come to agreement on what to do with urgency, etc) https://gitlab.freedesktop.org/wayland/wayland-protocols/-/issues/31

src/platform_impl/linux/x11/window.rs
```diff
@@ -1306,6 +1289,21 @@ impl UnownedWindow {
         self.set_ime_position_physical(x, y);
     }
 
+    #[inline]
+    pub fn request_user_attention(&self, request_type: RequestUserAttentionType) {
```
kchibisov, line 1293: You should address warnings.

maxded:
Alright, I'll set it back!

kchibisov (changes requested):
Also see above.

src/window.rs
```diff
@@ -682,6 +682,19 @@ impl Window {
     pub fn set_ime_position<P: Into<Position>>(&self, position: P) {
         self.window.set_ime_position(position.into())
     }
+
+    /// Requests user attention to the window, this has no effect if the application
+    /// is already focused. Providing `None` will unset the requesting of user attention.
```
kchibisov, line 687:
```suggestion
    /// is already focused. Providing `None` will unset the request for user attention.
```

src/window.rs
```diff
@@ -682,6 +682,19 @@ impl Window {
...
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **iOS / Android / Web / Wayland:** Unsupported.
+    ///
+    /// On macOS, providing `None` has no effect.
```
kchibisov, line 693: Could you update platform specific section style so it'll look like the one you wrote for `RequestUserAttentionType` events, so it should be without a new line between `ios`/`macOS` and should be '- **MacOS:** `None` has no effect'.

kchibisov (changes requested):

kchibisov:
@msiglreith Since you're doing windows stuff iirc, would you mind checking the windows implementation?

src/window.rs
msiglreith, line 687: It should be clear to the user if they are expected to call the unset option, but this seems to be already in discussion.

src/platform_impl/windows/window.rs
```diff
@@ -621,6 +621,33 @@ impl Window {
...
 
+    #[inline]
+    pub fn request_user_attention(&self, request_type: Option<RequestUserAttentionType>) {
+        let window = self.window.clone();
+
+        self.thread_executor.execute_in_thread(move || unsafe {
+            let (flags, count) = request_type
+                .map(|ty| match ty {
+                    RequestUserAttentionType::Critical => {
+                        (winuser::FLASHW_ALL | winuser::FLASHW_TIMERNOFG, u32::MAX)
```
msiglreith, line 632: Why is it `u32::MAX` here and `0` in the other cases? Based on the passed flags there should to be no difference in the behavior imo.

msiglreith (commented):
I'm generally in favor of this, thanks!
A few points to address in my eyes.

Windows 10 test results:
- Critical, Informational and None working as expected if not in foreground
- Critical case slightly unexpected result if the window is already in focus as it flashes the window caption once. This kinda conflicts with the documentation as it explicitly states having no influence

src/platform_impl/windows/window.rs
maxded: The difference between `Critical` and `Informational` on Windows is the continuous flashing of the window itself when `Critical`. `Informational` only flashes the taskbar button. The number of times the window flashes is `u32::MAX`.

src/window.rs
maxded: Added extra documentation regarding this!

src/platform_impl/windows/window.rs
msiglreith: nvm I somehow confused it with `dwTimeout`

msiglreith (approved):

src/window.rs
```diff
@@ -874,3 +891,23 @@ pub enum Theme {
...
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum RequestUserAttentionType {
+    /// ## Platform-specific
+    /// - **MacOS:** Bounces the dock icon until the application is in focus.
+    /// - **Windows:** Flashes both the window and the taskbar button until the application is in focus.
+    /// - **X11:** Sets the WM's `XUrgencyHint`. No distinction between `Critical` and `Informational`.
+    Critical,
+    /// ## Platform-specific
+    /// - **MacOS:** Bounces the dock icon once.
```
kchibisov, line 903:
```suggestion
    /// - **macOS:** Bounces the dock icon once.
```

src/window.rs
```diff
@@ -874,3 +891,23 @@ pub enum Theme {
     Light,
     Dark,
 }
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum RequestUserAttentionType {
+    /// ## Platform-specific
+    /// - **MacOS:** Bounces the dock icon until the application is in focus.
```
kchibisov, line 898:
```suggestion
    /// - **macOS:** Bounces the dock icon until the application is in focus.
```

kchibisov (changes requested):
Besides this one more comment, it's fine from my side.

I'll keep this open for a week so other maintainers can chime in if they want and discuss the proposed changes. If no complain will be raised I'll merge this PR.

@chrisduerr I know that you like ranting about docs, so your feedback wrt them is welcome.

src/platform_impl/linux/x11/window.rs
```diff
@@ -1306,6 +1289,23 @@ impl UnownedWindow {
         self.set_ime_position_physical(x, y);
     }
 
+    #[inline]
+    pub fn request_user_attention(&self, request_type: Option<RequestUserAttentionType>) {
```
chrisduerr, line 1293: I'd personally say that `RequestUserAttentionType` is incorrectly named. It should be `UserAttentionType`, or `UserAttentionRequestType`, the former probably being a bit better.

macOS already uses this, so one could argue it's easier to just keep things the same. But since this introduces breakage in this area anyways, it might be worth evaluating a rename.

src/window.rs
```diff
@@ -682,6 +682,23 @@ impl Window {
     pub fn set_ime_position<P: Into<Position>>(&self, position: P) {
         self.window.set_ime_position(position.into())
     }
+
+    /// Requests user attention to the window, this has no effect if the application
+    /// is already focused. Providing `None` will unset the request for user attention.
+    ///
+    /// Unsetting the request for user attention is automatically done by the WM when
+    /// the window receives focus.
```
chrisduerr, line 690: The specification is not able to guarantee this, so we should not do so either. If we state anything, we should state the opposite.

src/window.rs
```diff
@@ -682,6 +682,23 @@ impl Window {
     pub fn set_ime_position<P: Into<Position>>(&self, position: P) {
         self.window.set_ime_position(position.into())
     }
+
+    /// Requests user attention to the window, this has no effect if the application
+    /// is already focused. Providing `None` will unset the request for user attention.
```
chrisduerr, line 687: What does 'user attention' mean? It's probably useful to explain how this might manifest for the user to explain to downstream when they can use it.

src/window.rs
```diff
@@ -874,3 +891,23 @@ pub enum Theme {
...
+pub enum RequestUserAttentionType {
+    /// ## Platform-specific
+    /// - **macOS:** Bounces the dock icon until the application is in focus.
+    /// - **Windows:** Flashes both the window and the taskbar button until the application is in focus.
+    /// - **X11:** Sets the WM's `XUrgencyHint`. No distinction between `Critical` and `Informational`.
+    Critical,
+    /// ## Platform-specific
+    /// - **macOS:** Bounces the dock icon once.
+    /// - **Windows:** Flashes the taskbar button until the application is in focus.
+    /// - **X11:** Sets the WM's `XUrgencyHint`. No distinction between `Critical` and `Informational`.
```
chrisduerr, line 905: That's a lot of text I have to read just to find out that most of these are no different on X11. Might be useful to add a doc comment on the enum itself and describe that there?

chrisduerr (changes requested):

src/window.rs
maxded: Both Windows and macOS do give this guarantee though only X11 doesn't. That's why I put a platform-specific comment on X11 "warning" the user we can't give this guarantee.

src/window.rs
maxded: Agreed! Moved it to the enum itself instead of the variants.

src/window.rs
maxded: Added a reference to `UserAttentionType` as that explains it quite well. Is that ok?

src/platform_impl/linux/x11/window.rs
maxded: Think `UserAttentionType` is best as well :) Just copied it over from macOS when I started on this for consistency.

src/window.rs
chrisduerr: Pointing out that it isn't done on all platforms is much more useful though. It tells you that it might be done on some, but can't be taken for granted.

src/window.rs
```diff
@@ -682,6 +682,26 @@ impl Window {
...
     }
+
+    /// Requests user attention to the window, this has no effect if the application
+    /// is already focused. How requesting for user attention manifests is platform dependent,
+    /// see `UserAttentionType` for details.
+    ///
+    /// Providing `None` will unset the request for user attention.
+    ///
+    /// Unsetting the request for user attention might not be done automatically by the WM
+    /// when the window receives input.
```
chrisduerr, line 693:
```suggestion
    /// Providing `None` will unset the request for user attention. Unsetting the request for user attention might not be done automatically by the WM
    /// when the window receives input.
```

I think these can be combined, just to have text that is a bit easier to just read, rather than a huge list of disconnected points.

src/window.rs
```diff
@@ -682,6 +682,26 @@ impl Window {
...
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **iOS / Android / Web / Wayland:** Unsupported.
+    /// - **macOS:** `None` has no effect.
+    /// - **X11:** Automatic unsetting the request for user attention when the window  
+    /// receives focus might not be handled by some WMs.
```
chrisduerr, line 700: Might be easier to just state this as a fact. Like "Attention requests must be manually cleared". That way, people are automatically encouraged to do it right on X11.

chrisduerr (changes requested):

chrisduerr (approved):

kchibisov (approved):

PR #1767, opened 2020-11-16
Added focus_window to Window

### LINKED ISSUE #1750: Feature: bring window to top
In some applications, it is desirable to bring a window to the foreground without having it forced to be always on top. Right now this can be "achieved" through a hacky solution: `window.set_always_on_top(true)`, wait for an arbitrary duration for the execution thread to finish, and then call `window.set_always_on_top(false)`. This can of course fail as there is no telling if the winit's thread execution finished before calling `window.set_always_on_top(false)`. 

Would love to see a window API call along the lines of `window.bring_window_to_top()`. Let me know what you think!

maxded:
Fixes #1750

`Window::focus_window` brings the window to the front and sets input focus to that window. As described in the documentation, this method steals focus from other applications and should be used with care. Even though this method can cause quite a disruptive user experience I do believe winit should provide the option to users of this crate as valid use cases exist. Once #1764 has been merged, documentation could be expanded to redirect users to `request_user_attention` for a less disruptive way of getting attention to a window.

Decided to call it `focus_window` but the name could be changed for more clarity. Some APIs, like `set_minimized`, have a boolean parameter for toggling the, in this case, "minimized" state. For `focus_window` I decided not to do that as "unfocusing" is quite strange, like, where does the focus go? Also deliberately made the choice to "ignore" `focus_window` on minimized and not visible windows for improved user experience.

Don't have a Linux X11 machine at my disposal so the implementation might/is probably incorrect.

Tested on Windows and macOS.

### DIFF
--- a/src/platform_impl/linux/x11/window.rs
+++ b/src/platform_impl/linux/x11/window.rs
@@ -1309,7 +1309,6 @@ impl UnownedWindow {
     fn set_focus_inner(&self) -> util::Flusher<'_> {
         unsafe {
             let atom = self.xconn.get_atom_unchecked(b"_NET_ACTIVE_WINDOW\0");
-            
             self.xconn.send_client_msg(
                 self.xwindow,
                 self.root,
@@ -1322,9 +1321,23 @@ impl UnownedWindow {
 
     #[inline]
     pub fn focus_window(&self) {
-        self.focus_window_inner()
-            .flush()
-            .expect("Failed to focus window");
+        let state_atom = self.xconn.get_atom_unchecked(b"WM_STATE\0");
+        let is_minimized = if let Ok(state) =
+            self.xconn
+                .get_property(self.xwindow, state_atom, state_atom)
+        {
+            state.contains(&ffi::IconicState)
+        };
+        let is_visible = match self.shared_state.lock().visibility {
+            Visibility::Yes => true,
+            Visibility::YesWait | Visibility::No => false,
+        };
+
+        if is_visible && !is_minimized {
+            self.focus_window_inner()
+                .flush()
+                .expect("Failed to focus window");
+        }
     }
 
     #[inline]
--- a/src/platform_impl/linux/x11/window.rs
+++ b/src/platform_impl/linux/x11/window.rs
@@ -1306,6 +1306,27 @@ impl UnownedWindow {
         self.set_ime_position_physical(x, y);
     }
 
+    fn set_focus_inner(&self) -> util::Flusher<'_> {
+        unsafe {
+            let atom = self.xconn.get_atom_unchecked(b"_NET_ACTIVE_WINDOW\0");
+            
+            self.xconn.send_client_msg(
+                self.xwindow,
+                self.root,
+                atom,
+                Some(ffi::SubstructureRedirectMask | ffi::SubstructureNotifyMask),
+                [1, ffi::CurrentTime as c_long, 0, 0, 0],
+            )
+        }
+    }
+
+    #[inline]
+    pub fn focus_window(&self) {
+        self.focus_window_inner()
+            .flush()
+            .expect("Failed to focus window");
+    }
+
     #[inline]
     pub fn id(&self) -> WindowId {
         WindowId(self.xwindow)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - On X11, fix `Window::request_redraw` not waking the event loop.
 - On Wayland, the keypad arrow keys are now recognized.
 - **Breaking** Rename `desktop::EventLoopExtDesktop` to `run_return::EventLoopExtRunReturn`.
+- Added `Window::focus_window`to bring the window to the front and set input focus.
 
 # 0.23.0 (2020-10-02)
 
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -485,6 +485,8 @@ impl Window {
 
     pub fn set_ime_position(&self, _position: Position) {}
 
+    pub fn focus_window(&self) {}
+
     pub fn set_cursor_icon(&self, _: window::CursorIcon) {}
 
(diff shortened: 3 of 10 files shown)


chrisduerr:
> Even though this method can cause quite a disruptive user experience I do believe winit should provide the option to users of this crate as valid use cases exist.

You're saying this, but you're not providing any of those valid usecases. I'd be curious what those would be, especially since this isn't even possible on all platforms.

Generally I think it's pretty useless for winit to be the arbiter of what is morally right or wrong, but winit should definitely be the place to decide if something has a valid usecase, or if it is just a useless API that will lock people into a design that will cause failure when trying to make it work for other platforms.

So I personally think this should be very carefully questioned before being added. Only because it's easy to add, doesn't mean it's a good idea.

src/platform_impl/linux/x11/window.rs
```diff
@@ -1306,6 +1306,40 @@ impl UnownedWindow {
...
+            let atom = self.xconn.get_atom_unchecked(b"_NET_ACTIVE_WINDOW\0");
+            self.xconn.send_client_msg(
+                self.xwindow,
+                self.root,
+                atom,
+                Some(ffi::SubstructureRedirectMask | ffi::SubstructureNotifyMask),
+                [1, ffi::CurrentTime as c_long, 0, 0, 0],
+            )
+        }
+    }
```
chrisduerr, line 1320: This code is dead? If the goal was to call this with the `focus_window_inner` call, I don't see why it wouldn't just be inlined?

chrisduerr (changes requested):

maxded:
One use case would be opening links to your application. For example, Discord provides links users can click on to join a Discord server. The user expects, when clicking the link, to be brought to that server by opening/bringing Discord to the front and show the newly joined server. In these use cases, the user knows that it will be redirected to the application and focus stealing is ok. 

Instead of focusing the window when clicking a link, Discord could have gone the `request_user_attention` route but that just introduces another step the user has to take, which they, in most cases, don't want. (The link has clear intention).

> So I personally think this should be very carefully questioned before being added. Only because it's easy to add, doesn't mean it's a good idea.

Agreed! Part of the reasoning behind this PR is to spark discussion :)

chrisduerr:
> One use case would be opening links to your application. For example, Discord provides links users can click on to join a Discord server. The user expects, when clicking the link, to be brought to that server by opening/bringing Discord to the front and show the newly joined server. In these use cases, the user knows that it will be redirected to the application and focus stealing is ok.

Shouldn't that be handled by whatever mechanism decides which application to open a link with? I suppose that's kinda fair if an application is just re-executed or receives some event without getting focus assigned.

kchibisov:
I'm actually not sure how it should be done on X11 for raising window there's a https://specifications.freedesktop.org/startup-notification-spec/startup-notification-latest.txt , and for Wayland there will be similar protocol (it's under development now https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/50).

artursapek:
This would be useful for applications that can handle custom URL schemes - an already-running application that handles a link click from a web browser could bring itself to the foreground to indicate to the user that that is where the link click was handled.

repi:
Since we (Embark) created this PR we've been using it in production now for 6 months on Windows, Linux, Mac x86 and Mac aarch64 without any problems, specifically use it as one of our apps handles URLs and needs to bring the app window into focus when clicking a link. And has been working fine and reliable.

What would be needed to make some progress on this and get it in? Would be nice to get off our fork of winit

ArturKovacs:
I tried this on macOS with the following code. I run it with `cargo run`, quickly click on another window after the winit window shows up, but after "Focusing" gets printed, nothing seems to change, the winit window remains in the background.

```rust
use winit::{
    dpi::LogicalSize,
    event::{Event, WindowEvent},
    window::WindowBuilder,
};
fn main() {
    // event_loop.set_activation_policy(ActivationPolicy::Accessory);
        .with_title("my fantastic window ")
        .with_inner_size(LogicalSize::new(200, 50))
        .build(&event_loop)
        .unwrap();
    let start = std::time::Instant::now();
    let mut focused = false;
        if start.elapsed().as_secs_f32() > 4.0 && !focused {
            println!("Focusing");
            window.focus_window();
            focused = true;
        }
        match event {
                _ => (),
            },
            _ => (),
        }
    });
}

```

I tried merging master into this, and also tried calling `activateIgnoringOtherApps_` right after `makeKeyAndOrderFront_`, with no success.

Note that there's another PR, #1868 that implements this in a slightly different way. I could not get that to work either, so it's probably just my system. With that, the code seems fine so I'm willing to approve this for macOS. @msiglreith at this point I'm willing to approve both of these PRs, although I would request a change for the other PR. I could also test this for X11, and I think @maroider can test for X11 as well. So if you don't mind I'd like to ask you to decide which one we should go with.

msiglreith:
@ArturKovacs Ah sorry, I missed your question at the end somehow. For the Windows impl I'm slightly favoring this one, but the other one is fine as well. If someone is willing to approve the x11 parts we can go with this one as it's more complete/cross-platform.

ArturKovacs:
Yeah, I'll take responsibility for reviewing the X11 implementation for this. I fixed the compiler errors and then tested on X11. It works as expected. @maxded do you mind if I push my fixes along with some minor changes in X11 to this branch?

maxded:
@ArturKovacs go ahead!

ArturKovacs:
I can't push here, the branch name is not a link so I don't know what's the remote for it. (I pushed into a new `window_focus` branch in this repo, but that's not very helpful) Please help :D

Screenshot from 2021-05-17 18-39-22

maxded:
@ArturKovacs that is my bad, I think I've accidentally removed the branch some time ago. Should we continue this feature/PR from your new branch or should I revive mine and go from there?

ArturKovacs:
I'd just like to continue under this PR if that's possible, but otherwise it really doesn't make a difference I think.

PR #1615, opened 2020-06-30
On Windows, fix bug where we'd try to emit `MainEventsCleared` events during nested win32 event loops

Osspial:
- [x] Tested on all platforms changed
- [x] Compilation warnings were addressed
- [x] `cargo fmt` has been run on this branch
- [x] `cargo doc` builds successfully
- [x] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users

Fixes #1587 and #1602. Fixes #1611 .

PR #1769, opened 2020-11-19
Windows: Delayed Message Boxes Fix.

VZout:
Message boxes did not appear when opening them. The sound did play but they never were made visible unless the system triggers a redraw (for example when you press `alt`). The solution to this is to add the default `WM_PAINT` behavior back in. (`DefWindowProc` does this if you don't override `WM_PAINT`)

(This issue appeard in #1615)

### DIFF
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -1974,7 +1974,7 @@ unsafe extern "system" fn thread_event_target_callback<T: 'static>(
             }
 
             // Default WM_PAINT behaviour. This makes sure modals and popups are shown immediatly when opening them.
-            let mut ps: winuser::PAINTSTRUCT = std::mem::zeroed();
+            let mut ps = std::mem::MaybeUninit::<winuser::PAINTSTRUCT>::zeroed().assume_init();
             winuser::BeginPaint(window, &mut ps as *mut _);
             winuser::EndPaint(window, &mut ps as *mut _);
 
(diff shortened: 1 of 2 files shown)


maroider:
I ran ahead and tried your changes and it seems this fixes the problem on my machine (latest Win10).
This also seems to fix another problem I observed with message boxes and filed dialogs parented to the Winit window where instead of flashing, the child window would toggle the color of its title bar every time I clicked the Winit window.

VZout:
CI now fails on docs after I updated the changelog...

Seems to be unrelated though: `failed to load source for dependency bitflags`

src/platform_impl/windows/event_loop.rs
```diff
@@ -1973,6 +1973,11 @@ unsafe extern "system" fn thread_event_target_callback<T: 'static>(
                 }
             }
 
+            // Default WM_PAINT behaviour. This makes sure modals and popups are shown immediatly when opening them.
```
msiglreith, line 1976: I think this could be replaced then with just calling
```rust
commctrl::DefSubclassProc(window, msg, wparam, lparam)
```

Message boxes are still working correctly then.

msiglreith (commented):
thanks for looking into this!

src/platform_impl/windows/event_loop.rs
VZout: Did that in the first place but thought ya'll might prefer to call beginpain and endpaint instead.
Changed it to `commctrl::DefSubclassProc(window, msg, wparam, lparam)` in my latest commit

maroider: One thing that does bother me about this is that it's not clear why `DefSubclassProc` (or `BeginPaint` + `EndPaint` for that matter) isn't being called in the first place. It has apparently been this way since at least #638 (Event Loop 2.0).

PR #1771, opened 2020-11-21
On Wayland, increase default font size in CSD

kchibisov:
This commit increased default font size from 11 to 17 making it
identical to the one SCTK is using under the hood, since it's more
readable.


kchibisov:
@JMS55 since you've commented a lot in the past wrt CSD, does changing to 17 looks fine or we want slightly different size?

JMS55:
I think it's a good idea!

PR #1776, opened 2020-11-26
Android: support multi-touch

maxded:
Small PR enabling multi-touch on Android :) 

Also rearranged an entry in the changelog as it was added to the top instead of the bottom.

### DIFF
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -1,4 +1,4 @@
-#![cfg(target_os = "android")]
+//#![cfg(target_os = "android")]
 
 use crate::{
     dpi::{PhysicalPosition, PhysicalSize, Position, Size},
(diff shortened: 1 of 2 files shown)


src/platform_impl/android/mod.rs
```diff
@@ -1,4 +1,4 @@
-#![cfg(target_os = "android")]
+//#![cfg(target_os = "android")]
```
msiglreith, line 1: unintentional change?

src/platform_impl/android/mod.rs
```diff
@@ -181,37 +181,45 @@ impl<T: 'static> EventLoop<T> {
...
+                                                y: pointer.y() as _,
                                             };
-                                            call_event_handler!(
-                                                event_handler,
-                                                self.window_target(),
-                                                control_flow,
-                                                event
-                                            );
+
+                                            if let Some(phase) = phase {
```
msiglreith, line 203: minor: could move the if condition out of the loop

msiglreith (commented):
thanks!

src/platform_impl/android/mod.rs
maxded: Oh whoops :)

maxded:
Done, should be good to go now!

msiglreith (approved):

PR #1780, opened 2020-11-29
Windows: Changed thread_event_target_callback's WM_DESTROY to WM_NCDESTROY

VZout:
Our application was not exiting gracefully but crashed on `platform_impl/windows/event_loop.rs:1927`. 
`WM_DESTROY` destroys `subclass_input` but `WM_NCDESTOY` is send after `WM_DESTROY` ([See MS docs](https://docs.microsoft.com/en-us/windows/win32/winmsg/wm-ncdestroy)) which causes `subclass_input.event_loop_runner.clone();` to fail.

Examples don't repro, are the they just lucky that the memory didn't change?
Creating a minimal test case from our project would be quite a lot of work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- On Windows, fix applications not exiting gracefully.
 - On Android, calling `WindowEvent::Focused` now works properly instead of always returning false. 
 - On Windows, fix alt-tab behaviour by removing borderless fullscreen "always on top" flag.
 - On Windows, fix bug preventing windows with transparency enabled from having fully-opaque regions.
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -1939,7 +1939,7 @@ unsafe extern "system" fn thread_event_target_callback<T: 'static>(
     // the closure to catch_unwind directly so that the match body indendation wouldn't change and
     // the git blame and history would be preserved.
     let callback = || match msg {
-        winuser::WM_DESTROY => {
+        winuser::WM_NCDESTROY => {
             Box::from_raw(subclass_input);
             drop(subclass_input);
             0


VZout:
Actually seems exactly the same as #1745 but `thread_event_target_callback` instead of `public_window_callback`

src/platform_impl/windows/event_loop.rs
```diff
@@ -1939,7 +1939,7 @@ unsafe extern "system" fn thread_event_target_callback<T: 'static>(
     // the closure to catch_unwind directly so that the match body indendation wouldn't change and
     // the git blame and history would be preserved.
     let callback = || match msg {
-        winuser::WM_DESTROY => {
+        winuser::WM_NCDESTROY => {
```
chrisduerr, line 1942: The documentation says `WM_DESTROY` is for destruction of the main window, while `WM_NCDESTROY` is for child windows.

Looking at the docs, it seems like winit should be using `WM_DESTROY`, what makes you believe that `WM_NCDESTROY` is the better choice?

chrisduerr (changes requested):

src/platform_impl/windows/event_loop.rs
VZout: Because `WM_NCDESTROY` is also emitted if there are no child windows I believe.

chrisduerr: But `WM_NCDESTROY` being emitted doesn't seem sufficient as a reason to use it? That just seems like a coincidental fix.

maroider: ###### I'm (obviously) not @VZout, but...
I think you've misunderstood the docs.
When a window receives `WM_DESTROY`, three things are promised:

1. The window is being destroyed.
2. Any child windows still exist (but will soon be destroyed).
3. (implicitly) A `WM_NCDESTROY` message will be dispatched soon.

For `WM_NCDESTROY`, the following should hold true:

1. Any child windows are now destroyed
2. [MSDN](https://docs.microsoft.com/en-us/windows/win32/winmsg/wm-ncdestroy#remarks) "This message frees any memory internally allocated for the window."

This would suggest that the window can't receive any further messages after `WM_NCDESTROY` since Windows has presumably invalidated the window's handle, message queue, and other associated state (see [this SO answer](https://stackoverflow.com/a/41362844)).

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 
+- On Windows, fix applications not exiting gracefully.
```
maroider, line 3: I think it's worth mentioning the use-after-free (and perhaps how window destruction triggers it) here.

src/platform_impl/windows/event_loop.rs
VZout: Your answer was more detailed anyway ♥

CHANGELOG.md
VZout: How about:
"On Windows, fix applications not exiting gracefully due to `thread_event_target_callback` accessing corrupted memory."
or a more details:
"On Windows, fix applications not exiting gracefully due to `WM_DESTROY` events destroying `subclass_input` but `WM_NCDESTOY` is send after `WM_DESTROY` where the former uses `subclass_input."

VZout: changed it to my first suggestion :)

chrisduerr (commented):
I still find it questionable that memory is freed based on an event which *hopefully* is the last one received, that seems very unsafe to me.

But I have no knowledge of the Windows backend and do not plan to change that.

maroider:
It is unfortunate that this isn't documented all that well, but [it's probably safe to do it this way](https://devblogs.microsoft.com/oldnewthing/20050726-00/?p=34803), provided that no other [funny buisness](https://devblogs.microsoft.com/oldnewthing/20050727-16/?p=34793) occurs.

chrisduerr:
> it's probably safe to do it this way,

I'd say "probable safe" is not good enough personally.

As I've stated previously I'm not familiar with the Windows winit internals, but this error seems to be triggered by trying to handle events after we've already decided that the lifetime was over. Would it not be possible to store this information internally and prevent further events from being processed even if they would come in?

maroider:
It should be possible to replace the window's "window procedure" (event handler function) with a no-op function after the window receives `WM_NCDESTROY`, which would make it safe to deallocate `subclass_input` regardless of what messages are received afterwards.

msiglreith (approved):
IMO it's fine to move forward with this change - more sophisticated cleanup can be done as followup as well. This would include removing the subclass, which is supposed to be done in  `WM_NCDESTROY` anyway due to reverse ordering.

PR #1783, opened 2020-11-30
Android: Improved multi-touch

MarnixKuijs:
In the previous implementation of multi-touch there was a issue. When having two fingers on the screen and you would release one it would result in ended events being generated for all fingers even though not all fingers were lifted. This implementation will generate started and ended events for the right finger id's. Furthermore this implementation will generate a cancelled event with different defaults. These defaults should be according to spec since the user shouldn't preform any action it would normally perform: "_You will not receive any more points in it. You should treat this as an up event, but not perform any action that you normally would_" according to the [ndk documentation](https://developer.android.com/ndk/reference/group/input#anonymous-enum-32).

### DIFF
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -166,61 +166,61 @@ Legend:
 - ❓: Unknown status
 
 ### Windowing
-|Feature                          |Windows|MacOS   |Linux x11   |Linux Wayland  |Android|iOS    |WASM      |
-|-------------------------------- | ----- | ----   | -------    | -----------   | ----- | ----- | -------- |
-|Window initialization            |✔️     |✔️     |▢[#5]      |✔️             |▢[#33]|▢[#33] |✔️        |
-|Providing pointer to init OpenGL |✔️     |✔️     |✔️         |✔️             |✔️     |✔️    |**N/A**|
-|Providing pointer to init Vulkan |✔️     |✔️     |✔️         |✔️             |✔️     |❓     |**N/A**|
-|Window decorations               |✔️     |✔️     |✔️         |▢[#306]        |**N/A**|**N/A**|**N/A**|
-|Window decorations toggle        |✔️     |✔️     |✔️         |✔️             |**N/A**|**N/A**|**N/A**|
-|Window resizing                  |✔️     |▢[#219]|✔️         |▢[#306]        |**N/A**|**N/A**|✔️        |
-|Window resize increments         |❌     |❌     |❌         |❌             |❌    |❌     |**N/A**|
-|Window transparency              |✔️     |✔️     |✔️         |✔️             |**N/A**|**N/A**|N/A        |
-|Window maximization              |✔️     |✔️     |✔️         |✔️             |**N/A**|**N/A**|**N/A**|
-|Window maximization toggle       |✔️     |✔️     |✔️         |✔️             |**N/A**|**N/A**|**N/A**|
-|Window minimization              |✔️     |✔️     |✔️         |✔️             |**N/A**|**N/A**|**N/A**|
-|Fullscreen                       |✔️     |✔️     |✔️         |✔️             |**N/A**|✔️     |✔️        |
-|Fullscreen toggle                |✔️     |✔️     |✔️         |✔️             |**N/A**|✔️     |✔️        |
-|Exclusive fullscreen             |✔️     |✔️     |✔️         |**N/A**         |❌    |✔️     |**N/A**|
-|HiDPI support                    |✔️     |✔️     |✔️         |✔️             |▢[#721]|✔️    |✔️ \*1|
-|Popup windows                    |❌     |❌     |❌         |❌             |❌    |❌     |**N/A**|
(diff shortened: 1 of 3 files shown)


msiglreith:
Thanks, but I'm not totally sure what's the motivation behind this one. Please add a small description. I assume it refers to the default values for cancelled touch events?

MarnixKuijs:
Updated the description to provide some detail.

MarnixKuijs:
Converted back to draft to make some extra changes

MarnixKuijs:
Seems one check failed due to network error and not my code changes.

MarnixKuijs:
PR should be ready now. Cancelled events are generated for every pointer now. This seems to be the better approach, since it is more clear to the user that the moment a cancelled event is generated its generated for all fingers instead of a non existing one.

src/platform_impl/android/mod.rs
```diff
@@ -180,47 +180,117 @@ impl<T: 'static> EventLoop<T> {
...
+                                                        x: pointer.x() as _,
+                                                        y: pointer.y() as _,
+                                                    };
+
+                                                    let event = event::Event::WindowEvent {
+                                                        window_id,
+                                                        event: event::WindowEvent::Touch(
+                                                            event::Touch {
+                                                                device_id,
+                                                                phase: event::TouchPhase::Moved,
```
msiglreith, line 277: should be `Cancelled`

src/platform_impl/android/mod.rs
```diff
@@ -180,47 +180,117 @@ impl<T: 'static> EventLoop<T> {
...
+                                                        self.window_target(),
+                                                        control_flow,
+                                                        event
+                                                    );
+                                                }
+                                            }
+                                            MotionAction::Cancel => {
+                                                for pointer in motion_event.pointers() {
+                                                    let location = PhysicalPosition {
+                                                        x: pointer.x() as _,
```
msiglreith, line 268: you previously had `0.0` here for defaults - was it intential to move back to the pointer values?

msiglreith (changes requested):
Looks good overall, a few points are open and I would be also interested a bit in compatibility with other backends (ie iOS) regarding semantics.

- If possible is there a way to shrink the code a bit as the branches often only differ in small details?
- Do you know by chance how the cancelled case is handled on iOS? Single event vs multiple events, default values, etc?

cc @francesca64

FEATURES.md
```diff
@@ -166,61 +166,61 @@ Legend:
...
-|Window transparency              |✔️     |✔️     |✔️         |✔️             |**N/A**|**N/A**|N/A        |
-|Window maximization              |✔️     |✔️     |✔️         |✔️             |**N/A**|**N/A**|**N/A**|
-|Window maximization toggle       |✔️     |✔️     |✔️         |✔️             |**N/A**|**N/A**|**N/A**|
-|Window minimization              |✔️     |✔️     |✔️         |✔️             |**N/A**|**N/A**|**N/A**|
-|Fullscreen                       |✔️     |✔️     |✔️         |✔️             |**N/A**|✔️     |✔️        |
-|Fullscreen toggle                |✔️     |✔️     |✔️         |✔️             |**N/A**|✔️     |✔️        |
-|Exclusive fullscreen             |✔️     |✔️     |✔️         |**N/A**         |❌    |✔️     |**N/A**|
-|HiDPI support                    |✔️     |✔️     |✔️         |✔️             |▢[#721]|✔️    |✔️ \*1|
-|Popup windows                    |❌     |❌     |❌         |❌             |❌    |❌     |**N/A**|
+| Feature                          | Windows | MacOS   | Linux x11 | Linux Wayland | Android | iOS     | WASM    |
```
francesca64, line 169: Are these formatting changes intentional?

MarnixKuijs: No they aren't I think my vscode decided to do that automatically. I will change it back

src/platform_impl/android/mod.rs
MarnixKuijs: Yes it is intentional from my side, but I have to be honest I'm unsure what the right approach is when it comes to cancelled events.

msiglreith: looking through the iOS implementation, the current proposal seems to match quite well

MarnixKuijs:
Did some clean up to reduce code bloat and changed the feature matrix back to its original formatting. I guess I should leave cancelled how it is?

msiglreith (approved):

francesca64 (changes requested):
Thanks for the PR! The semantics look good. Could you add a CHANGELOG entry, so users know about this cool change?

MarnixKuijs:
So there already was a mention of Android multi-touch in the change log and since it was still in unreleased I just added some detail instead of making an extra entry.

francesca64:
Sorry, I didn't realize there was already an unreleased entry for multi-touch! I think it'd actually be better not to touch the CHANGELOG then, since the issue in question was never released.

MarnixKuijs:
Ooh okay, let me undo my change.

francesca64 (approved):
Looks great! We'll merge once CI finishes.

PR #1784, opened 2020-12-04
Windows: Preserve minimized/maximized state in fullscreen

### LINKED ISSUE #1765: Windows: Fullscreen doesn't preserve minimized/maximized state
## Version information

* winit 0.23.0
* Windows 10

## Steps to reproduce the problem

1. Run the [`fullscreen`](https://github.com/rust-windowing/winit/blob/master/examples/fullscreen.rs) example:
```sh
cargo run --example fullscreen
```
2. Maximize the window.
3. Enter fullscreen by pressing <kbd>F</kbd>.
4. Exit fullscreen by pressing <kbd>F</kbd> again.

After exiting fullscreen, the window is no longer maximized, and it covers the whole screen.

## Expected behavior

After exiting fullscreen, the window is *still* maximized, and unmaximizing it brings it to the initial size and position before it was maximized.

This is how many programs with fullscreen functionality behave (e.g. Chrome, VLC).

relrelb:
By lining up `Window::set_fullscreen` with Chromium's [`FullscreenHandler::SetFullscreenImpl`](https://chromium.googlesource.com/chromium/src/+/40351207e40888d64d3e02711287144ede7b4454/ui/views/win/fullscreen_handler.cc#55).

**EDIT**: `Window::set_fullscreen` is now based on https://devblogs.microsoft.com/oldnewthing/20100412-00/?p=14353, as it seems to be much more clean and simple compared to the Chromium way. I would also like to receive some feedback regarding the chosen method.

Fixes #1765.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 
-- On Windows, fix fullscreen doesn't preserve minimized/maximized state.
+- On Windows, fix fullscreen not preserving minimized/maximized state.
 
 # 0.24.0 (2020-12-09)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,12 @@
+# Unreleased
+
+- On Windows, fix fullscreen doesn't preserve minimized/maximized state.
+
 # 0.24.0 (2020-12-09)
 
 - On Windows, fix applications not exiting gracefully due to thread_event_target_callback accessing corrupted memory.
 - On Windows, implement `Window::set_ime_position`.
 - **Breaking:** On Windows, Renamed `WindowBuilderExtWindows`'s `is_dark_mode` to `theme`.
-- On Windows, fix fullscreen doesn't preserve minimized/maximized state.
 - **Breaking:** On Windows, renamed `WindowBuilderExtWindows::is_dark_mode` to `theme`.
 - On Windows, add `WindowBuilderExtWindows::with_theme` to set a preferred theme.
 - On Windows, fix bug causing message boxes to appear delayed.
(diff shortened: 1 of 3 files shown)


src/platform_impl/windows/window.rs
```diff
@@ -487,21 +473,22 @@ impl Window {
                 winuser::PeekMessageW(&mut msg, ptr::null_mut(), 0, 0, 0);
             }
 
-            // Update window style
-            WindowState::set_window_flags(window_state_lock, window.0, |f| {
```
msiglreith, line 491: I think it would be ideal to keep this part here and adopt the `set_window_flags` implementation by basically moving all the parts below into it. So we change the window style only in a single place and it's also the only usage of the `FULLSCREEN` window flags

msiglreith (changes requested):
Hi 👋 
thanks for looking into this and setting up a PR!

I haven't tested it locally, but will do it soon. For now a few style related change requests.
Also, thanks for the changelog cleanup (:

src/platform_impl/windows/window.rs
```diff
@@ -523,34 +518,17 @@ impl Window {
                             position.1,
                             size.0 as i32,
                             size.1 as i32,
-                            winuser::SWP_ASYNCWINDOWPOS | winuser::SWP_NOZORDER,
+                            winuser::SWP_NOOWNERZORDER | winuser::SWP_FRAMECHANGED,
```
relrelb, line 521: I'm not exactly sure what to put here, I just aligned it with https://devblogs.microsoft.com/oldnewthing/20100412-00/?p=14353.

msiglreith: I think the previous flags make sense here, would prefer to keep them

CHANGELOG.md
```diff
@@ -1,9 +1,12 @@
+# Unreleased
+
+- On Windows, fix fullscreen doesn't preserve minimized/maximized state.
```
maroider, line 3: Nit: "fix fullscreen not preserving" is more grammatically correct than "fix fullscreen doesn't preserve".

msiglreith (approved):

PR #1787, opened 2020-12-05
On Wayland, don't drop extra mouse buttons

kchibisov:
Since Wayland mouse buttons are generally u16, we can only somehow
translate them into u8 used by winit. Also, they're coming from
linux/input-event-codes.h directly, thus translating them into
something cross-platform is cumbersome, thus just passing them
downstream as 'button - BTN_LEFT' for now.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ raw-window-handle = "0.3"
 bitflags = "1"
 
 [dev-dependencies]
-image = "0.23"
+image = "0.23.12"
 simple_logger = "1.9"
 
 [target.'cfg(target_os = "android")'.dependencies]
--- a/src/platform_impl/linux/wayland/seat/pointer/handlers.rs
+++ b/src/platform_impl/linux/wayland/seat/pointer/handlers.rs
@@ -17,6 +17,11 @@ use crate::platform_impl::wayland::{self, DeviceId};
 
 use super::{PointerData, WinitPointer};
 
+// These values are comming from <linux/input-event-codes.h>.
+const BTN_LEFT: u32 = 0x110;
+const BTN_RIGHT: u32 = 0x111;
+const BTN_MIDDLE: u32 = 0x112;
+
 #[inline]
 pub(super) fn handle_pointer(
     pointer: ThemedPointer,
@@ -153,11 +158,10 @@ pub(super) fn handle_pointer(
             };
 
             let button = match button {
-                0x110 => MouseButton::Left,
-                0x111 => MouseButton::Right,
-                0x112 => MouseButton::Middle,
-                // TODO - figure out the translation.
-                _ => return,
+                BTN_LEFT => MouseButton::Left,
+                BTN_RIGHT => MouseButton::Right,
+                BTN_MIDDLE => MouseButton::Middle,
+                button => MouseButton::Other(button as u16),
             };
 
             event_sink.push_window_event(
--- a/src/platform_impl/linux/x11/event_processor.rs
+++ b/src/platform_impl/linux/x11/event_processor.rs
@@ -709,7 +709,7 @@ impl<T: 'static> EventProcessor<T> {
                                 event: MouseInput {
                                     device_id,
                                     state,
-                                    button: Other(x as u8),
+                                    button: Other(x as u16),
                                     modifiers,
                                 },
                             }),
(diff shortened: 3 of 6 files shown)


src/platform_impl/linux/wayland/seat/pointer/handlers.rs
```diff
@@ -153,11 +158,10 @@ pub(super) fn handle_pointer(
...
             let button = match button {
-                0x110 => MouseButton::Left,
-                0x111 => MouseButton::Right,
-                0x112 => MouseButton::Middle,
-                // TODO - figure out the translation.
-                _ => return,
+                BTN_LEFT => MouseButton::Left,
+                BTN_RIGHT => MouseButton::Right,
+                BTN_MIDDLE => MouseButton::Middle,
+                button => MouseButton::Other((button - BTN_LEFT) as u8),
```
chrisduerr, line 164: Can there be any buttons that are smaller `0x110`, or why does it not start at `0`? If so, that would cause troubles, right?

kchibisov: I think it could happen but I'm not sure how to handle that without changing winit's `Other` type, since button is u16, not u8. And I have no clue if kernel uses certain things for mouse buttons and not e.g. numpad.

kchibisov: But if it matters, I took this logic from glfw, and they subtract left mouse button here if it matters. In theory I can convert everything to u16 and it'll 100% work.

chrisduerr: I mean at the very least it shouldn't cause an overflow like it's currently doing. Reporting an incorrect button seems much worse than reporting no buttons at all and I wouldn't want to crash in debug mode.

Unfortunately the button values seem kinda odd to me, usually I would have thought a right shift or bitwise ops could have removed all pointless bits, but this `0x110` value seems a bit arbitrary. Unless the button value is always the last 4 bits of `button`.

kchibisov: Just look at the include file I've mentioned on the issue, they are really arbitrary.

chrisduerr: That definitely looks like those buttons are a thing though. So we either have to filter them or remap them I suppose. But we shouldn't rely on overflow.

kchibisov: I'll just bump things to u16.

Cargo.toml
```diff
@@ -33,7 +33,7 @@ raw-window-handle = "0.3"
 bitflags = "1"
 
 [dev-dependencies]
-image = "0.23"
+image = "0.23.12"
```
kchibisov, line 36: Doesn't build with earlier version of image afaics.

src/platform_impl/windows/event_loop.rs
```diff
@@ -1307,7 +1307,7 @@ unsafe extern "system" fn public_window_callback<T: 'static>(
                 event: MouseInput {
                     device_id: DEVICE_ID,
                     state: Pressed,
-                    button: Other(xbutton as u8),
+                    button: Other(xbutton),
```
kchibisov, line 1310: Seems like it was also incorrectly converted here.

src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -709,7 +709,7 @@ impl<T: 'static> EventProcessor<T> {
                                 event: MouseInput {
                                     device_id,
                                     state,
-                                    button: Other(x as u8),
+                                    button: Other(x as u16),
```
kchibisov, line 712: I'm not sure how sane X11 is, but I think it's fine to use u16, given that kernel uses u16 as well.

Cargo.toml
chrisduerr: This is one of the reasons why I would **always** encourage specifying the full versions. Only because it builds with 0.23.X, doesn't mean it builds with 0.23.0. So I'd rather have it a bit higher than necessary than too low.

src/platform_impl/linux/x11/event_processor.rs
chrisduerr: What is the type natively? usize?

chrisduerr (approved):

kchibisov:
I've looked into kernel docs and it seems like the the only mouse buttons starting from BTN_MISC up to BTN_TASK, so we can subtract `0x100` for them if it matters, but we can keep u16 for now, since it's more clear.

PR #1797, opened 2020-12-11
On Windows, fix bug causing mouse capture to not be released.

### LINKED ISSUE #1796: Windows: Mouse capture not released in some cases
To enter the incorrect state:

1 - Run one of the window examples.
2 - Press a mouse button in the window content.
3 - Press ALT+TAB to focus another window.
4 - Release the mouse button.
5 - Focus the example window again.
6 - Press and release a mouse button in the window content.

After these steps it is not possible to click the window close button because the mouse is captured.

SamRodri:

Fixes #1796

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+- On Windows, fix bug causing mouse capture to not be released.
+
 # 0.24.0 (2020-12-09)
 
 - On Windows, fix applications not exiting gracefully due to thread_event_target_callback accessing corrupted memory.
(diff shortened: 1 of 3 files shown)


maroider:
I tried your changes, and they seem to fix the issue.

msiglreith (approved):

maroider:
I must admit that I only tested whether it fixed #1796 without checking for regressions. I've now discovered that doing the following seems to cause a deadlock:
1. Press down one mouse button over the window
2. Press down another mouse button over the window, without releasing the first one

PR #1798, opened 2020-12-11
[Windows] Improve handling of window destruction

### LINKED ISSUE #1795: Windows: Improve application teardown
We should ensure that we don't access dropped value (ie subclass input) after receiving the destruction event messages.
In particular, this includes removing the subclass from the window.

maroider:

This is a follow-up to #1780 and #1746.

Fixes #1795.

### DIFF
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -4,6 +4,7 @@ mod runner;
 
 use parking_lot::Mutex;
 use std::{
+    cell::Cell,
     collections::VecDeque,
     marker::PhantomData,
     mem, panic, ptr,
@@ -86,6 +87,8 @@ pub(crate) struct SubclassInput<T: 'static> {
     pub window_state: Arc<Mutex<WindowState>>,
     pub event_loop_runner: EventLoopRunnerShared<T>,
     pub file_drop_handler: Option<FileDropHandler>,
+    pub subclass_removed: Cell<bool>,
+    pub recurse_depth: Cell<u32>,
 }
 
 impl<T> SubclassInput<T> {
@@ -616,6 +619,17 @@ fn subclass_event_target_window<T>(
     }
 }
 
+fn remove_event_target_window_subclass<T: 'static>(window: HWND) {
+    let removal_result = unsafe {
+        commctrl::RemoveWindowSubclass(
+            window,
+            Some(thread_event_target_callback::<T>),
+            THREAD_EVENT_TARGET_SUBCLASS_ID,
+        )
+    };
+    assert_eq!(removal_result, 1);
+}
+
 /// Capture mouse input, allowing `window` to receive mouse events when the cursor is outside of
 /// the window.
 unsafe fn capture_mouse(window: HWND, window_state: &mut WindowState) {
@@ -646,6 +660,17 @@ pub(crate) fn subclass_window<T>(window: HWND, subclass_input: SubclassInput<T>)
         )
     };
     assert_eq!(subclass_result, 1);
+}
+
+fn remove_window_subclass<T: 'static>(window: HWND) {
+    let removal_result = unsafe {
+        commctrl::RemoveWindowSubclass(
+            window,
+            Some(public_window_callback::<T>),
+            WINDOW_SUBCLASS_ID,
+        )
+    };
+    assert_eq!(removal_result, 1);
 }
 
 fn normalize_pointer_pressure(pressure: u32) -> Option<Force> {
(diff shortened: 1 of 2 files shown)


src/platform_impl/windows/event_loop.rs
```diff
@@ -815,7 +840,10 @@ unsafe extern "system" fn public_window_callback<T: 'static>(
 
         winuser::WM_NCDESTROY => {
             drop(subclass_input);
-            Box::from_raw(subclass_input_ptr as *mut SubclassInput<T>);
+            remove_window_subclass(
+                window,
+                Box::from_raw(subclass_input_ptr as *mut SubclassInput<T>),
+            );
```
chrisduerr, line 846: Dropping subclass input first and then trying to do something with a pointer to it seems strange to me.

src/platform_impl/windows/event_loop.rs
```diff
@@ -648,6 +662,17 @@ pub(crate) fn subclass_window<T>(window: HWND, subclass_input: SubclassInput<T>)
...
 
+fn remove_window_subclass<T>(window: HWND, _subclass_input: Box<SubclassInput<T>>) {
+    let removal_result = unsafe {
+        commctrl::RemoveWindowSubclass(
+            window,
+            Some(public_window_callback::<T>),
+            WINDOW_SUBCLASS_ID,
+        )
+    };
+    assert_eq!(removal_result, 1);
```
chrisduerr, line 673: I'm not sure how I feel about these assertions. I feel like the most common way for this to fail is if the subclass is already gone and I'm not sure if we should tear down the entire application in that case.

With applications running multiple windows this could lead to a crash of the entire application, assuming this subclass info is added separately for each window (I'm not sure about that?).

chrisduerr (changes requested):

src/platform_impl/windows/event_loop.rs
maroider: The `subclass_input` that's dropped has a type of `&SubclassInput<T>` (introduced [here](https://github.com/rust-windowing/winit/blob/3e9681e8d94db0e72750b884fdb4ab0b3ab44d75/src/platform_impl/windows/event_loop.rs#L781)). Dropping it is necessary to avoid UB.

src/platform_impl/windows/event_loop.rs
maroider: The subclass is registered individually for every window, so I don't see how it could be "already gone" when `event_loop` is the only module with enough information (the callback function pointer and subclass ID) to do so.

src/platform_impl/windows/event_loop.rs
maroider: > references are copy types

Ah, I had completely forgotten, but that then makes me wonder why it was ever there in the first place.

maroider: I've dug into the code some more, and it seems like there isn't anything guaranteeing that the event target window gets destroyed after all other windows. This should be fixable by temporarily cloning and holding onto the `Rc` until we don't need it.
 The slightly more ambiguous issue is to do with pointer aliasing. It feels like having a `&` reference to something while also having a `*mut` pointer to it is insta-UB, but I'd have to consult with someone more knowledgeable on this.

src/platform_impl/windows/event_loop.rs
maroider: Additionally, it seems like the only "always safe" place to deregister subclasses is during window destruction.

maroider:
The code in question now relies on `Box::from_raw` and `Box::into_raw` rather than casting to references, as previously suggested.
Dropping `subclass_input` has been moved to the very end of the callback functions.

I realized as I was writing this that the potential panics due to the added `assert`s is no longer covered by `catch_unwind`. Admittedly, if a panic does occur, then that means that one of the assumptions we're making to maintain memory safety has been violated.

maroider:
Panicky bits are now inside `catch_unwind` again.

src/platform_impl/windows/event_loop.rs
```diff
@@ -2148,5 +2184,14 @@ unsafe extern "system" fn thread_event_target_callback<T: 'static>(
         _ => commctrl::DefSubclassProc(window, msg, wparam, lparam),
     };
 
-    runner.catch_unwind(callback).unwrap_or(-1)
+    let result = subclass_input
+        .event_loop_runner
+        .catch_unwind(callback)
+        .unwrap_or(-1);
+    if ncdestroy {
+        remove_event_target_window_subclass(window, subclass_input);
```
msiglreith, line 2192: I'm not really happy with that solution tbh - unfortunately the topic is quite complex to get right due to the recursive call of the window proc (e.g DestroyWindow will invoke the callback again for th NCDESTROY case etc) and this approach makes it harder to maintain imo.
To be fair I currently don't have a clear opinion on the best way either. I played a bit around with local dereferencing but it doesn't feel robust either: https://github.com/msiglreith/winit/commit/9fd2ceea6763f4db60d52ecb62d2c85daaff6d49

maroider: Would returning `(LRESULT, Option<Box<SubclassInput<T>>>)` from the `callback` closure and calling `Box::into_raw` in the `(_, Some(_))` case be better?

maroider: **Or** (assuming that I've understood the recursive nature of `WndProc` correctly): The `subclass_input: SubclassInput<T>` could be changed to be `Rc<SubclassInput<T>>` which always has one "fake" reference. At the beginning of each call to `*_callback` we'd deref the `subclass_input_ptr` and clone the `Rc`, which would ensure that we never drop the `subclass_input` before all calls to the callback are done. `WM_NCDESTROY` would then deregister the subclass and remove the "fake" reference, in that order.

maroider:
`public_window_callback` now tracks how deep the recursion is, and only drops the `subclass_input` when we're certain that there is no recursion. `thread_event_target_callback` does not seem to get called recursively at the moment, so I've omitted tracking it for now. Tracking recursion in `thread_event_target_callback` may be advisable, but I'm not 100% certain of the necessity of it.

The window subclass removal happens immediately after matching `WM_NCDESTROY`, since doing it later causes it to unconditionally fail in `public_window_callback`.

src/platform_impl/windows/event_loop.rs
maroider: >  The slightly more ambiguous issue is to do with pointer aliasing. It feels like having a `&` reference to something while also having a `*mut` pointer to it is insta-UB, but I'll have to consult with someone more knowledgeable on this before reaching a conclusion.

I talked with someone on the rust community discord, and it seems like this is indeed UB.

msiglreith (approved):
Thanks!

PR #1807, opened 2020-12-23
[Windows] Set the cursor icon when the cursor first enters a window

### LINKED ISSUE #1682: Windows spinning cursor on startup
Try  running the hello-triangle example in release mode on Windows to reproduce.

Not only is the initial startup time of the application uncomfortably long, but the blue spinning circle appears.
Even when the app finishes loading it persists. It only vanishes once you start moving the cursor.

maroider:

I'm not sure if this is the best way to address #1682 since I'm not familiar with how Windows handles cursors WRT windows, and I plan on looking further into this before converting from a draft PR to a normal PR.

Fixes #1682

### DIFF
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -20,7 +20,7 @@ use winapi::shared::basetsd::{DWORD_PTR, UINT_PTR};
 
 use winapi::{
     shared::{
-        minwindef::{BOOL, DWORD, HIWORD, INT, LOWORD, LPARAM, LRESULT, UINT, WPARAM},
+        minwindef::{BOOL, DWORD, HIWORD, INT, LOWORD, LPARAM, LRESULT, UINT, WORD, WPARAM},
         windef::{HWND, POINT, RECT},
         windowsx, winerror,
     },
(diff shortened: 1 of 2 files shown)


maroider:
It turns out that the first attempt at a fix made it so the "resize" cursor icons never appeared when hovering the cursor over the window borders.

maroider:
It turns out that `WM_SETCURSOR` arrives with hit-test results in `lParam`, so we can use that instead of relying on other messages.

I'm mildly concerned about ordering guarantees here, since users may want to set a cursor and have it applied immediately when the cursor enters the window. It may actually be fine in practice, but I don't really know.

src/platform_impl/windows/event_loop.rs
```diff
@@ -1664,11 +1664,7 @@ unsafe fn public_window_callback_inner<T: 'static>(
         winuser::WM_SETCURSOR => {
             let set_cursor_to = {
                 let window_state = subclass_input.window_state.lock();
-                if window_state
-                    .mouse
-                    .cursor_flags()
-                    .contains(CursorFlags::IN_WINDOW)
-                {
+                if (lparam & 0xFFFF) == winuser::HTCLIENT {
```
msiglreith, line 1667: `(lparam & 0xFFFF)` -> `LOWORD(lparam)` 
Also, wrapping the condition into a separate let statement and/or short command regarding hit-testing here would be great

msiglreith (changes requested):
Thanks!

src/platform_impl/windows/event_loop.rs
maroider: > command

I assume you meant to write "comment" here.

maroider:
Never mind that push. I trusted rust-analyzer a bit too much while I simultaneously did something rust-analyzer didn't support.

msiglreith (approved):

Issue #720, opened 2018-11-20
Drag & Drop meta-issue


Osspial:
This is mostly a stub issue to keep track of progress on improvements to the drag/drop API:
- Dropping lists of files
- Dropping non-file data
  - Dropping multiple types of data at once (i.e. receiving unformatted text and a formatted, HTML version of that text)
  - Potential datatypes to support:
    - Text
    - HTML
    - Image
    - Audio

I did some research into platform-specific clipboard APIs, and as far as I can tell those are the datatypes that we can probably support across all platforms. I'm hesitant to support audio data sine there seems to be way more variance in how platforms support that than the other datatypes, but from what I've seen most platforms expose it in one way or another.

maroider:
A couple of relevant issues have cropped up since this issue was created:
- The position of the thing being dragged over the window isn't delivered to users (#1550)
- Drag & Drop isn't implemented on Wayland (#1881)

janus-culsans:
I think out of all these issues #1550 would be the best starting point, also given that it has the most upvotes.
Essentially this boils down to allowing the cursor position to be read while the mouse is dragging something over the window.
This is completely impossible currently.

kchibisov:
The mouse coordinates are sent during dnd in dnd events? You clearly know where file is dropped. It's just not attached to the event itself and is sent separately.

janus-culsans:
Why not just tell how it is sent? 😄 
`CursorMoved` is not getting sent, unless bevy uses some very old version of winit or somehow breaks it.
Do you mean `DeviceEvent::MouseMotion`? Whatever it is, more people beside me don't seem to know/get that

kchibisov:
@janus-culsans The cursor position should be broadcasted during the drag and drop. It could be that a particular platform isn't doing so which could indicate a bug. The events are coming not from the drag and drop events, but from `Cursor` window events, so maybe on a platforms you're testing it's lacking.

goolord:
it would be nice to support both drag & drop as a target AND as a source

janus-culsans:
@kchibisov if the unsupported platform is MSWindows, okay?

kchibisov:
@janus-culsans are you planning to implement something mentioned here?

janus-culsans:
@kchibisov I have no knowledge of your internals and no time left, otherwise I would have probably understood your other comments and taken a shot. 
I'm sorry, it's just frustrating.

[...]

PR #1808, opened 2020-12-25
Add file drag and drop support on wasm

ShadowMitia:

Hello!

I'm trying to add drag and drop for wasm.
I don't know if I should open an issue for this as well, let me know if I should.

I've arrived to a sort of PoC for it on this branch but I have several issues.

If you want to try out this branch, just run the "web" example and watch for any of the `DroppedFile`, `HoveredFile`, `HoveredFileCanceled` in the `WindowEvent`.

First of all I seem to have a bug for `HoveredFiles` because when getting the files in the underlying objects, the file list is always 0. I need to investigate that some more.

Currently the `DroppedFile`, `HoveredFile` only return the filenames (not the full path), mainly because of HTML5 security reasons. And I don't really know how to move from here.

I see two ways to solve this :

- Change `DroppedFile`, `HoveredFile` for wasm, which return some new type which will be the file data directly (or allows for easy access, without having to deal with HTML stuff).
- Keep the interface, but have another object that allows to access the files, and use the filename in `DroppedFile`, `HoveredFile` to gain access to the correct file. This sounds more reaosnable, but I have currrently no idea how to do this.

I'll look into this some more, but if anyone has any ideas or feedbacks I would appreciate it!

Related issue : https://github.com/rust-windowing/winit/issues/720


ShadowMitia:
I've arrived to a POC. I've thought about it and I realised that this needs to be as simple as possible, and give the user the choice of managing what to do with the file, just like with the regular interface. It still needs a different interface in wasm, because we're dealing with the file directly. So I've chosen to return a `web_sys::File` directly, and then have the user manage the event however they seem. I've got an example running in the `web.rs` example.

daxpedda:
I'm going to close this as it's pretty old and needs to be done from scratch anyway.

Would love to see support for this, happy to review a new PR!

Issue #1763, opened 2020-11-11
Request_redraw() call from another thread doesn’t produce RedrawRequested event.


VincentFTS:
On MacOS, when Window::request_redraw() is called from another thread, Event::RedrawRequested is not emitted until another event is.
To reproduce this, just launch request_redraw_threaded example.

VincentFTS:
FYI : not reproduced with X11, but reproduced with web-sys.

kchibisov:
I think some backends don't wake up for whatever reason, and it was fixed like recently for X11.

I know that it wake ups on Wayland, so I'll save time testing.

VincentFTS:
I hope this whatever reason is whatever easy to fix !!

mr1sunshine:
I can confirm that this is still an issue on the current `master`. I can work it out with having `Poll` control flow, but it consumes a lot of resources unfortunately. I'm just wondering if there are any estimations on it?

ArturKovacs:
Resolved by #1812

PR #1812, opened 2020-12-31
macOS: Wake up run loop immediately when request redraw is received

mr1sunshine:
We allow to have RunLoop running only on the main thread. Which means if
we call Window::request_redraw() from other the thread then we have to
wait until some other event arrives on the main thread. That situation
is even worse when we have ControlFlow set to the `Wait` mode then user
will not ever render anything.

The goal of this PR is a try to fix macOS-related part of the issue #1763 I was not sure if it is a good idea to wake up RunLoop directly from the AppState, so please correct me with the right place to add this logic if needed.

- [x] Tested on all platforms changed (only macOS is affected)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Added `is_maximized` method to `Window`.
 - On Windows, fix bug where clicking the decoration bar would make the cursor blink.
 - On Windows, fix bug causing newly created windows to erroneously display the "wait" (spinning) cursor.
+- On macOS, wake up the event loop immediately when a redraw is requested.
 - On Windows, change the default window size (1024x768) to match the default on other desktop platforms (800x600).
 - On Windows, fix bug causing mouse capture to not be released.
 - On Windows, fix fullscreen not preserving minimized/maximized state.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 - On Windows, change the default window size (1024x768) to match the default on other desktop platforms (800x600).
 - On Windows, fix bug causing mouse capture to not be released.
 - On Windows, fix fullscreen not preserving minimized/maximized state.
+- macOS: Wake up run loop immediately when request redraw is received.
```
maroider, line 6:
```suggestion
- On macOS, wake up the run loop immediately when a redraw is requested.
```
Nit: Shouldn't this added at the top rather than at the bottom?

mr1sunshine: @maroider fixed, thanks!

ArturKovacs (approved):
Awesome, thanks a lot. I'll merge this as soon as you resolve the merge conflicts.

mr1sunshine:
@ArturKovacs should be good now, please proceed with merge.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - Added `is_maximized` method to `Window`.
 - On Windows, fix bug where clicking the decoration bar would make the cursor blink.
 - On Windows, fix bug causing newly created windows to erroneously display the "wait" (spinning) cursor.
+- On macOS, wake up the run loop immediately when a redraw is requested.
```
maroider, line 9:
```suggestion
- On macOS, wake up the event loop immediately when a redraw is requested.
```

mr1sunshine: @maroider Fixed, thanks.

Issue #1723, opened 2020-09-23
Android redraw requests aren't implemented


glalonde:
https://github.com/rust-windowing/winit/blob/c9558c5f0e8f5ccb7b250a59c46c0f7a41c2416a/src/platform_impl/android/mod.rs#L408

For now, implementing redraw in MainEventsCleared works. But I'm not sure what's supposed to go on here.

francesca64:
It looks like it was just punted:

https://github.com/rust-windowing/winit/blob/be2e17d605d1780bba15cd7fa4a98ef4917cedf4/src/platform_impl/android/mod.rs#L408-L410

I haven't seen any discussion indicating that it'd be thorny to implement; it's fairly simple on most platforms, albeit very platform-specific. A PR would be very welcome!

francesca64:
@dvc94ch do you remember anything about this?

dvc94ch:
Not sure what it's useful for. I guess setting an AtomicBool and calling looper.wake() should do the trick?

rmyhovych:
@francesca64 I've been having an issue with this in my Android application, where the RedrawRequested event was never called and my scene would stay static after 2 frames, so I made my own small function that fixed it for me:

`    pub fn request_redraw() {
        match ndk_glue::native_window().as_ref() {
            Some(native_window) => {
                let a_native_window: *mut ndk_sys::ANativeWindow = native_window.ptr().as_ptr();
                let a_native_activity: *mut ndk_sys::ANativeActivity =
                    ndk_glue::native_activity().ptr().as_ptr();
                unsafe {
                    match (*(*a_native_activity).callbacks).onNativeWindowRedrawNeeded {
                        Some(callback) => callback(a_native_activity, a_native_window),
                        None => (),
                    };
                };
            }
            None => (),
        }
    }
`

I'm very new to Rust and Native Android development, is there anything wrong with my way of doing?

PR #1822, opened 2021-01-09
Android: add request_redraw based on ndk_glue

semtexzv:

Fixes https://github.com/rust-windowing/winit/issues/1723, just picked up the code from there, seems to work fine

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 # Unreleased
-
+- On Android, implemented `request_redraw`, fixing issue where iced would not render properly
 - On Windows, fix bug causing newly created windows to erroneously display the "wait" (spinning) cursor.
 - On Windows, change the default window size (1024x768) to match the default on other desktop platforms (800x600).
 - On Windows, fix bug causing mouse capture to not be released.
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -448,7 +448,20 @@ impl Window {
     }
 
     pub fn request_redraw(&self) {
-        // TODO
+        match ndk_glue::native_window().as_ref() {
+            Some(native_window) => {
+                let a_native_window: *mut ndk_sys::ANativeWindow = native_window.ptr().as_ptr();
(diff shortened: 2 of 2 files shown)


MarijnS95:
While this might work it is not all that elegant to go through the unsafe, native `onNativeWindowRedrawNeeded` callback that is really only supposed to be called by Android, but happens to be publicly available to users of `ndk-glue`.

Instead we might want to make the `fn wake(Event)` function public to push the event safely in through there (this is where the above native call ends up). Then however I don't think we want to make that function public, nor is this going to help distinguish where the event came from (may we ever need that).

Two options remain:
1. Use `looper.wake()` and somehow finagle the "redraw requested' state somewhere, either in an atomic bool as suggested by @dvc94ch, maybe a more generic atomic integer/queue that can keep multiple events may we need it in the future, or stuff it into the existing `user_queue` that is already read when the event loop receives `EventSource::User` (perhaps by wrapping its items in another enum);
2. Create a pipe (see `ndk-glue` code), attach that to the looper on a new `ident`, add a new `EventSource` that is returned from `fn poll()` when an event is retrieved on that pipe, then read that event from the pipe and set `redraw = true`.

Let's discuss with the maintainers here to find the best solution before deciding anything :smile:. CC @msiglreith and @francesca64

msiglreith:
I would prefer the 1st option suggested by @MarijnS95 

> Use looper.wake() and somehow finagle the "redraw requested' state somewhere, either in an atomic bool as suggested by @dvc94ch, maybe a more generic atomic integer/queue that can keep multiple events may we need it in the future, or stuff it into the existing user_queue that is already read when the event loop receives EventSource::User (perhaps by wrapping its items in another enum);

Wrapping the current user event in an enum and a RedrawRequest would probably be straightforward as the parts are already in place. Immediately emitting the redraw request after the MainEventsCleared call seems more tricky, but could be a followup imo.

msiglreith:
Thanks, https://github.com/rust-windowing/winit/pull/1953 landed which covers the functionality from this PR

PR #1826, opened 2021-01-13
android: Use event identifier instead of userdata pointer

MarijnS95:
ndk-glue currently sets both the `ident` field and user-data pointer to `0` or `1` for the event pipe and input queue respectively, to tell these sources apart. While it works to reinterpret this `data` pointer as integer identifier it shouldn't be abused for that, in particular when one may wish to provide extra information with an event in the future; then the `data` field is used as pointer (or abused as abstract value) for that.

---

These constants (`0` and `1`) are internal to `ndk-glue` and not publicly documented anywhere (afaik). I believe they should at least be exposed as constants, or perhaps abstracted behind a `poll` function internal to `ndk-glue`. Consequently we might want to warn API users when they try use one of these "reserved `ident`s" (but only on the thread looper set up by `ndk-glue`s `init()` function)?

CC @dvc94ch @msiglreith 

### DIFF
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -30,7 +30,7 @@ enum EventSource {
 
 fn poll(poll: Poll) -> Option<EventSource> {
     match poll {
-        Poll::Event { data, .. } => match data as usize {
+        Poll::Event { ident, .. } => match ident {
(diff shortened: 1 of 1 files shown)


dvc94ch (approved):
LGTM, thanks for all your work on android

PR #1830, opened 2021-01-18
 On Windows, fix deadlock caused by mouse capture

maroider:
The issue was caused by calling SetCapture on a window which already
had the capture. The WM_CAPTURECHANGED handler assumed that it would
only run if the capture was lost, but that wasn't the case. This made
the handler to try to lock the window state mutex while it was already
locked.

The issue was introduced in #1797 (932cbe4).

### DIFF
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -1394,8 +1394,13 @@ unsafe fn public_window_callback_inner<T: 'static>(
         }
 
         winuser::WM_CAPTURECHANGED => {
-            // window lost mouse capture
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 
+- On Windows, fix deadlock caused by pressing multiple mouse buttons down at once.
```
msiglreith, line 3: As the issues only persists in the unreleased master there is no need for a changelog entry

maroider: Ok, will remove.

PR #1833, opened 2021-01-23
Derive more things

LoganDark:
Having issues where LogicalPosition<f32> has no Default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 - On Windows, added `WindowBuilderExtWindows::with_owner_window` to allow creating popup windows.
 - On Windows, added `WindowExtWindows::set_enable` to allow creating modal popup windows.
 - On macOS, emit `RedrawRequested` events immediately while the window is being resized.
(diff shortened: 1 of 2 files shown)


LoganDark:
bruh the windows check failed because of a network error, oof

ArturKovacs:
`Default` and `Hash` seem reasonable to derive, but `Eq` should be handled like this:
```rust
// Only implement `Eq` for `LogicalPosition`s where the inner type is `Eq`
impl<P: Eq> Eq for LogicalPosition<P> {}
```

maroider:
Isn't that what `#[derive()]` does anyway?

ArturKovacs:
> Isn't that what `#[derive()]` does anyway?

I clearly misunderstood the documentation. Yeah these can all just be derive.

Although please add a changelog entry along the lines of
```
- Implement `Default`, `Hash`, and `Eq` for `LogicalPosition`, `PhysicalPosition`, `LogicalSize`, and `PhysicalSize`.
```

LoganDark:
@ArturKovacs Done

PR #1840, opened 2021-01-26
Add dragging window with cursor feature

### LINKED ISSUE #221: Consider options for redefining locations where the user can drag, resize, close, etc. a window
This is most likely low priority but a nice-to-have for making custom styled windows using borderless windows, and still allowing the user to drag a custom rendered title bar. On windows this is implemented using `WM_NCHITTEST`, which allows the application to define custom areas that can do these operations.

daxpedda:

---

Partially fixes #221.

~~Still working on the other platforms, see https://github.com/rust-windowing/winit/issues/221#issuecomment-767690279 for details.
Would really appreciate somebody looking at the Wayland implementation, any bikeshedding is welcome too.~~
~~I've successfully tried all implementations, currently stuck at trying to fix the Windows implementation deadlock.~~

`cargo run --example drag_window`

~~Blocked on https://github.com/Smithay/client-toolkit/pull/174.~~
~~Blocked on [`smithay-client-toolkit`](https://crates.io/crates/smithay-client-toolkit) releasing a new version.~~

~~Wayland implementation at #1854.~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,6 @@
 - On Android, bump `ndk` and `ndk-glue` to 0.3: use predefined constants for event `ident`.
 - On Windows, fixed `WindowEvent::ThemeChanged` not properly firing and fixed `Window::theme` returning the wrong theme.
 - On Web, added support for `DeviceEvent::MouseMotion` to listen for relative mouse movements.
-- Added `Window::set_drag_window`. Implemented on Windows, macOS and X11.
 - Added `Window::begin_drag`. Implemented on Windows, macOS and X11.
 
 # 0.24.0 (2020-12-09)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 - On Android, bump `ndk` and `ndk-glue` to 0.3: use predefined constants for event `ident`.
 - On Windows, fixed `WindowEvent::ThemeChanged` not properly firing and fixed `Window::theme` returning the wrong theme.
 - On Web, added support for `DeviceEvent::MouseMotion` to listen for relative mouse movements.
-- Added `Window::drag_window`. Implemented on Windows, macOS and X11.
+- Added `Window::drag_window`. Implemented on Windows, macOS, X11 and Wayland.
 
 # 0.24.0 (2020-12-09)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,6 @@
 - On Android, bump `ndk` and `ndk-glue` to 0.3: use predefined constants for event `ident`.
 - On Windows, fixed `WindowEvent::ThemeChanged` not properly firing and fixed `Window::theme` returning the wrong theme.
 - On Web, added support for `DeviceEvent::MouseMotion` to listen for relative mouse movements.
-- Added `Window::set_drag_window`. Implemented on Windows, macOS, X11 and Wayland.
 - Added `Window::set_drag_window`. Implemented on Windows, macOS and X11.
 - Added `Window::begin_drag`. Implemented on Windows, macOS and X11.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 - On Android, bump `ndk` and `ndk-glue` to 0.3: use predefined constants for event `ident`.
 - On Windows, fixed `WindowEvent::ThemeChanged` not properly firing and fixed `Window::theme` returning the wrong theme.
 - On Web, added support for `DeviceEvent::MouseMotion` to listen for relative mouse movements.
-- Added `Window::begin_drag`. Implemented on Windows, macOS and X11.
+- Added `Window::drag_window`. Implemented on Windows, macOS and X11.
 
 # 0.24.0 (2020-12-09)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - On Web, added support for `DeviceEvent::MouseMotion` to listen for relative mouse movements.
 - Added `Window::set_drag_window`. Implemented on Windows, macOS, X11 and Wayland.
 - Added `Window::set_drag_window`. Implemented on Windows, macOS and X11.
+- Added `Window::begin_drag`. Implemented on Windows, macOS and X11.
 
 # 0.24.0 (2020-12-09)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - On Windows, fixed `WindowEvent::ThemeChanged` not properly firing and fixed `Window::theme` returning the wrong theme.
 - On Web, added support for `DeviceEvent::MouseMotion` to listen for relative mouse movements.
 - Added `Window::set_drag_window`. Implemented on Windows, macOS, X11 and Wayland.
+- Added `Window::set_drag_window`. Implemented on Windows, macOS and X11.
 
 # 0.24.0 (2020-12-09)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - On Android, bump `ndk` and `ndk-glue` to 0.3: use predefined constants for event `ident`.
 - On Windows, fixed `WindowEvent::ThemeChanged` not properly firing and fixed `Window::theme` returning the wrong theme.
 - On Web, added support for `DeviceEvent::MouseMotion` to listen for relative mouse movements.
+- Added `Window::set_drag_window`. Implemented on Windows, macOS, X11 and Wayland.
 
 # 0.24.0 (2020-12-09)
 
--- a/examples/drag_window.rs
+++ b/examples/drag_window.rs
@@ -35,7 +35,7 @@ fn main() {
                     &window_1
                 };
 
-                window.set_drag_window().unwrap()
+                window.begin_drag().unwrap()
             }
             WindowEvent::CursorEntered { .. } => {
                 let (drag_target, other) = if (window_id == window_1.id() && switched)
--- a/examples/drag_window.rs
+++ b/examples/drag_window.rs
@@ -37,6 +37,17 @@ fn main() {
 
                 window.set_drag_window().unwrap()
             }
+            WindowEvent::CursorEntered { .. } => {
+                let (drag_target, other) = if (window_id == window_1.id() && switched)
+                    || (window_id == window_2.id() && !switched)
+                {
+                    (&window_2, &window_1)
+                } else {
+                    (&window_1, &window_2)
+                };
+                drag_target.set_title("drag target");
+                other.set_title("winit window");
+            }
             WindowEvent::KeyboardInput {
                 input:
                     KeyboardInput {
--- a/examples/drag_window.rs
+++ b/examples/drag_window.rs
@@ -4,7 +4,7 @@ use winit::{
         ElementState, Event, KeyboardInput, MouseButton, StartCause, VirtualKeyCode, WindowEvent,
     },
     event_loop::{ControlFlow, EventLoop},
-    window::WindowBuilder,
+    window::{Window, WindowBuilder, WindowId},
 };
 
 fn main() {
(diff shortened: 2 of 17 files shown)


maroider:
Could you use a more descriptive PR title?

daxpedda:
> Could you use a more descriptive PR title?

That was an accident :/, can't think of a better formulation, suggestions welcome.

maroider:
I don't think the current API is a good fit for the Windows backend.

For a given mouse action, the `WM_NCHITTEST` message will, in my experience, arrive before any other mouse events. If we try to set a flag when the user calls `Window::set_drag_window` and then modify the next `WM_NCHITTEST` result, we run the risk of introducing some unfortunate edge cases. If the window spawns beneath the user's cursor, and the user doesn't move the cursor before pressing a mouse button, then the drag won't be initiated at all.

Note that I haven't tested this, but there was a similar issue I had to solve in #1807.

daxpedda:
@maroider I've used [`WM_NCLBUTTONDOWN`](https://docs.microsoft.com/en-us/windows/win32/inputdev/wm-nclbuttondown) instead, does this resolve your concern?

src/platform_impl/windows/window.rs
```diff
@@ -356,6 +356,21 @@ impl Window {
         Ok(())
     }
 
+    #[inline]
+    pub fn set_drag_window(&self) -> Result<(), ExternalError> {
+        unsafe {
+            winuser::ReleaseCapture();
```
daxpedda, line 362: Should this set `window_state.mouse.capture_count` to `0`?

maroider: You should test for regressions for #1830, #1797, and #1796.

You should also check what mouse capture looks like (on current `master`) when you do the following:
1. Press a (non-left) mouse button down with the cursor in the client area of the window and don't release it.
2. Move the cursor to the title bar.
3. Initiate a drag by pressing the left mouse button down.
4. Release the left moues button.
5. Move the mouse cursor outside the window.
6. Release the other mouse button.

That's (probably) the behaviour you want to match.

maroider: You should also make sure nothing strange happens if `Window::set_drag_window` is called without any mouse buttons being pressed.

daxpedda: Thank you, that helped a lot, everything seems to be in order except one big thing:
indeed this implementation causes a deadlock, it actually always happens, as soon as one starts to drag the window, it stops all incoming events, until the mouse button is released again.

I tried to figure it out but failed, any help would be appreciated.

maroider: It's not really a deadlock. You're blocking the thread with `SendMessageW`, which will block until the message has been processed. Using `PostMessageW` instead seems to work fine.

One interesting "feature" I discovered is that you can initiate the drag when the cursor is outside the window by doing the following:
1. Press a non-left mouse button down inside the window.
2. Move the cursor outside the window.
3. Press the left mouse button.

I'm not sure if this is desirable.

daxpedda:
I wanna thank @maroider and @vberger, they were a lot of help with this PR.

If no more concerns are raised I believe the implementation is complete, we are just waiting for an update from [`smithay-client-toolkit`](https://crates.io/crates/smithay-client-toolkit).

src/platform_impl/windows/window.rs
maroider: In the second case, you have try dragging twice. Once inside the window, and once outside.

daxpedda: Alright, I reproduced it. Should we document this in your opinion?

maroider: This outside-the-window-initiated dragging isn't something I see much sense in keeping around. I'd prefer to have a check for whether the mouse is inside the bounds of the window or not, since that's what I'd assume is expected on other platforms.

daxpedda: So your comment prompted me to check how other platforms handle this, apparently this is pretty consistent except for Wayland.
This outside-the-window-initiated dragging makes it possible to drag a window from another window, so I think it's best to keep it like that and note that this doesn't work on Wayland, I will also add an example for this.

daxpedda:
Wayland implementation split off to #1854.
This is now ready to be merged.

examples/drag_window.rs
```diff
@@ -0,0 +1,56 @@
...
+                let window = if (window_id == window_1.id() && switched)
+                    || (window_id == window_2.id() && !switched)
+                {
+                    &window_2
+                } else {
+                    &window_1
+                };
+
+                window.set_drag_window().unwrap()
+            }
```
maroider, line 39:
```suggestion
            }
            WindowEvent::CursorEntered { .. } => {
                let (drag_target, other) = if (window_id == window_1.id() && switched)
                    || (window_id == window_2.id() && !switched)
                {
                    (&window_2, &window_1)
                } else {
                    (&window_1, &window_2)
                };
                drag_target.set_title("drag target");
                other.set_title("winit window");
            }
```

I had to insert the above to make sense of what pressing <kbd>X</kbd> here actually does.

maroider (commented):
Bikeshed: I don't feel like `set_drag_window` is a particularly good name. The other `set_*` methods on `Window` either manipulate some piece of retained state, or aren't quite as restricted in when they can be called successfully. Perhaps `begin_drag` would be better.

examples/drag_window.rs
daxpedda: Good idea.

daxpedda:
`being_drag` sounds better indeed, on it.

daxpedda:
Requesting a review from @maroider, @ArturKovacs, @murarth and @vberger please.
Thank you!

CHANGELOG.md
```diff
@@ -12,6 +12,8 @@
 - On Android, bump `ndk` and `ndk-glue` to 0.3: use predefined constants for event `ident`.
 - On Windows, fixed `WindowEvent::ThemeChanged` not properly firing and fixed `Window::theme` returning the wrong theme.
 - On Web, added support for `DeviceEvent::MouseMotion` to listen for relative mouse movements.
+- Added `Window::set_drag_window`. Implemented on Windows, macOS and X11.
+- Added `Window::begin_drag`. Implemented on Windows, macOS and X11.
```
maroider, line 16:
```suggestion
- Added `Window::begin_drag`. Implemented on Windows, macOS and X11.
```
I guess this is an oversight.

maroider (approved):
I'll admit that I'm not a huge fan of the Windows implementation, but that's mostly because it conflicts with how I thought the message queue was supposed to work.

As I recall from when I last tested this, it works great, despite my initial concern.

In any case, if you want a review from the _actual_ Windows maintainer, then that's @msiglreith.

CHANGELOG.md
daxpedda: Yes, thank you.

daxpedda:
Requesting a review from @msiglreith.

msiglreith (approved):
LGTM, played a bit around with it and didn't notice any issues (Windows 10). The only minor part is that it still works even if the event receiving window differs from the drag window, but not sure if it's worth worrying about.

(Sth to discuss might be the method name as it sounds like drag-and-drop related - maybe `drag_window` would fit better?)

daxpedda:
> (Sth to discuss might be the method name as it sounds like drag-and-drop related - maybe `drag_window` would fit better?)

Agreed.

I intend to merge the Wayland implementation, which needs to upgrade SCTK to 0.13, this adds `libfreetype-dev` and `libexpat-dev` as a dependency on Linux, is that acceptable or should I keep it in a seperate PR for now?

vberger:
> I intend to merge the Wayland implementation, which needs to upgrade SCTK to 0.13, this adds libfreetype-dev and libexpat-dev as a dependency on Linux, is that acceptable or should I keep it in a seperate PR for now?

These are actually transitive dependencies of the new dependency on `libfontconfig`, which is used to find a font for rendering the text on decorations. I don't expect to find many systems that don't have fontconfig. (And, of course, the `-dev` dependencies are only build-time dependencies, not runtime).

ArturKovacs:
This mostly looks good regarding the macOS implementation. There is something strange I experienced though. Start the example, bring another application to the front, and while the other application is in the front, click the close button on one of the winit windows. They won't close, not even if I bring them back to the front after this.

I put some `println`s into `src/platform_impl/macos/app_state.rs` within the block that starts with
```rust
if HANDLER.should_exit() {
```

Turns out that somehow the application gets stuck in a state where neither of the two windows is the "main window". The source of this dialog handling code is #1581. I didn't dig deeper, I don't really have an idea what causes this or if it's normal.

daxpedda:
@ArturKovacs I am assuming this has nothing to do with `drag_window` but that the example is using multiple windows?

ArturKovacs:
That's what I'm suspecting as well, but I couldn't reproduce it with the `multithreaded` example.

daxpedda:
@vberger Do you know what the builds issue is? Apparently `expat` doesn't like the 32-bit C++ compiler GitHub Actions is offering?

vberger:
An option would be to just install fontconfig in the CI environment, and avoid having the -sys crates build their own versions of the system libs they are associated with (which is apparently something servo likes to do?).

daxpedda:
So do I have to add a new Winit crate feature that activates `servo-fontconfig`'s `force_system_lib` to fix this then? I can't figure out a simpler/cleaner way.

vberger:
Ah wait, looks like the lib was already installed. Looks like the issue I had with my CI: is pkgconfig installed in the CI environment? Without it the build scripts cannot find the libraries.

daxpedda:
> is pkgconfig installed in the CI environment? Without it the build scripts cannot find the libraries.

Definitely installed: https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#installed-apt-packages.

vberger:
Hmm, given the error log says `/usr/bin/ld: cannot find -lstdc++`, I guess some part of the multilib c++ toolchain is missing. Maybe try installing both `gcc-multilib` and `g++-multilib`, rather than just the former?

daxpedda:
I don't even know what to do with this error message now.
@vberger I'm willing to try new ideas if you have any, but I would rather spend my time making a PR for `rust_fontconfig`, would you accept one?

vberger:
Hmm, I guess servo-fontconfig-sys is unhappy because it cannot find an i686 version of libfontconfig, and then tries to build its own which fails horribly...

I myself am not opposed to using `rust_fontconfig` as long as it works correctly, but I don't know if the winit maintainers would be okay with using such a new and not battle-tested crate.

An other option would be to completely disable writing the title in SCTK fallback decorations, I guess. Having text rendering in SCTK always struck me as quite an heavy dependency for so little gain.

daxpedda:
> ..., but I don't know if the winit maintainers would be okay with using such a new and not battle-tested crate.

Should I ping the rest of the Linux maintainers for this?

> An other option would be to completely disable writing the title in SCTK fallback decorations, I guess. Having text rendering in SCTK always struck me as quite an heavy dependency for so little gain.

Personally, I agree with that, do we need to aquire consensus for that, or can you decide that?

vberger:
@daxpedda To avoid blocking this further on unrelated stuff, I've released version 0.12.3 which only contains the change you need on top of 0.12.2, so you can just use that. All the stuff about fontconfig is unrelated an should be tackled elsewhere.

daxpedda:
Alright, thats great. Thanks!

vberger (approved):
The wayland part looks good.

murarth (approved):

daxpedda:
> That's what I'm suspecting as well, but I couldn't reproduce it with the `multithreaded` example.

@ArturKovacs I got my hands on a Apple machine just now and tried it, I wasn't able to reproduce the issue :disappointed:.
I am still unsure what this could have to do with the `drag_window` implementation, but seeing that I don't see how I will fix this, I would rather table the MacOS implementation.

Potentially we could also just document not to use this function with multiple windows on MacOS?

ArturKovacs (approved):
Never mind, I could reproduce the faulty behavior on the `master` branch, so it's in fact completely unrelated. This can be merged as far as I'm concerned.

ArturKovacs:
Considering that there hasn't been Windows related changes since @msiglreith approved, it seems that everyone required approved these changes. I'm merging this now.

PR #1841, opened 2021-01-26
Fix HINSTANCE returned by raw_window_handle on 64 bit Windows

madsmtm:
  - Tested with `x86_64-pc-windows-gnu`, but can't get my build system to produce a 32bit version, so can't test that :/
  - I didn't add it to the changelog since this is a very small fix.

We previously retrieved the value using `GetWindowLongW`, but [the documentation](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getwindowlongw) explicity says that `GetWindowLongPtrW` should be used for pointers and handles (which is what we're fetching).

### DIFF
--- a/src/platform_impl/windows/window.rs
+++ b/src/platform_impl/windows/window.rs
@@ -278,7 +278,7 @@ impl Window {
 
     #[inline]
     pub fn hinstance(&self) -> HINSTANCE {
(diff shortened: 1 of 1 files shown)


msiglreith (approved):
Thanks!

PR #1842, opened 2021-01-26
Add ability to assign a menu when creating a window on Windows

madsmtm:

Added `WindowBuilderExtWindows::with_menu`.

The primary reason for doing it here instead of just assigning a menu later on with `winapi::um::winuser::SetMenu` is to avoid that `with_inner_size` breaks, since the menu takes up some space that is not taken into account, which means the user would have to use `set_inner_size` again after adding the menu.
This, I'm guessing, could cause a jarring effect for the user (that the window resizes when starting).

Unresolved:
- [x] ~Child windows cannot have a menu, so perhaps this should be represented typewise?~

### DIFF
--- a/src/platform_impl/windows/mod.rs
+++ b/src/platform_impl/windows/mod.rs
@@ -17,7 +17,6 @@ use crate::window::Theme;
 
 #[derive(Clone)]
 pub struct PlatformSpecificWindowBuilderAttributes {
-    // TODO: parent and menu are mutually exclusive; a child window cannot have a menu!
     pub parent: Option<HWND>,
     pub menu: Option<HMENU>,
     pub taskbar_icon: Option<Icon>,
--- a/src/platform_impl/windows/mod.rs
+++ b/src/platform_impl/windows/mod.rs
@@ -1,6 +1,6 @@
 #![cfg(target_os = "windows")]
 
-use winapi::{self, shared::windef::HWND};
+use winapi::{self, shared::windef::HMENU, shared::windef::HWND};
 
 pub use self::{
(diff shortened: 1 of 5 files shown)


src/platform_impl/windows/mod.rs
```diff
@@ -17,7 +17,9 @@ use crate::window::Theme;
 
 #[derive(Clone)]
 pub struct PlatformSpecificWindowBuilderAttributes {
+    // TODO: parent and menu are mutually exclusive; a child window cannot have a menu!
```
msiglreith, line 20: This should be added the `with_menu` documentationa and I would say a `warn` during the creation if that case is hit would be enough

msiglreith (commented):
Thanks! LGTM

madsmtm:
Done. Should I do a rebase?

It also occurs to me, both `with_parent_window` and `with_menu` are probably unsafe, since you could safely pass just about any pointer to these, and bad things would probably happen then?

Also, `with_win32_menu` is maybe a better name? Though not so important.

msiglreith:
> Done. Should I do a rebase?

Thanks - no, we squash-merge here :D

> It also occurs to me, both with_parent_window and with_menu are probably unsafe, since you could safely pass just about any pointer to these, and bad things would probably happen then?

Agree, but I think that's a different issue right now, I would like to handle this separately. There is also an open issue regarding exposing typedefs like `HWND` or `HMENU` in public API without exporting.

> Also, with_win32_menu is maybe a better name? Though not so important.

The current name looks totally fine to me!

msiglreith (approved):

PR #1843, opened 2021-01-27
IOS ScreenEdge Corrected Default

### LINKED ISSUE #1613: [iOS] Crashed by *** -[__NSArrayM insertObject:atIndex:]: object cannot be nil
If I tap on the moment it starts up on iOS, the following error occurs and the app finished.

```
2020-06-28 20:11:01.732058+0900 <project-name>[58982:6919187] *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '*** -[__NSArrayM insertObject:atIndex:]: object cannot be nil'
*** First throw call stack:
(0x1c4642a48 0x1c4369fa4 0x1c4698360 0x1c46979f4 0x1c451a120 0x1c82ba31c 0x1c82ba4f8 0x1c82ba71c 0x1c82bb460 0x1c82af4c0 0x1c82aec38 0x1c82ae9fc 0x1c8704374 0x1c86e0a9c 0x1c8758c20 0x1c875b17c 0x1c8753ff0 0x1c45c0a00 0x1c45c0958 0x1c45c00f0 0x1c45bb23c 0x1c45baadc 0x1ce55b328 0x1c86c863c 0x104d324f4 0x104d5f45c 0x104d0f71c 0x104be78e4 0x104be7930 0x104be7a94 0x104be7b7c 0x104be79f0 0x104be7870 0x104be7c68 0x104be790c 0x104be7848 0x1c4444360)
libc++abi.dylib: terminating with uncaught exception of type NSException
(lldb) 
```

Xcode wrote the following stack trace for me.
スクリーンショット 2020-06-28 20 34 18

スクリーンショット 2020-06-28 20 34 31

```
UIKitCore`UIApplicationMain:
    0x1c86c7eac <+0>:    sub    sp, sp, #0xa0             ; =0xa0 
    0x1c86c7eb0 <+4>:    stp    x28, x27, [sp, #0x40]
    0x1c86c7eb4 <+8>:    stp    x26, x25, [sp, #0x50]
    0x1c86c7eb8 <+12>:   stp    x24, x23, [sp, #0x60]
    0x1c86c7ebc <+16>:   stp    x22, x21, [sp, #0x70]
    0x1c86c7ec0 <+20>:   stp    x20, x19, [sp, #0x80]
    0x1c86c7ec4 <+24>:   stp    x29, x30, [sp, #0x90]
    〜 omitted 〜
    0x1c86c8610 <+1892>: bl     0x1c441e9d8               ; kdebug_trace
    0x1c86c8614 <+1896>: mov    x0, x23
    0x1c86c8618 <+1900>: bl     0x1c4384130               ; objc_release
    0x1c86c861c <+1904>: ldr    x0, [sp, #0x28]
    0x1c86c8620 <+1908>: bl     0x1c4384130               ; objc_release
    0x1c86c8624 <+1912>: ldr    x0, [sp, #0x30]
    0x1c86c8628 <+1916>: bl     0x1c8c03944               ; symbol stub for: -[UIDocumentSharingController _setMailSubject:]
    0x1c86c862c <+1920>: ldr    x0, [x20, #0xd08]
    0x1c86c8630 <+1924>: adrp   x8, 213832
    0x1c86c8634 <+1928>: add    x1, x8, #0x378            ; =0x378 
    0x1c86c8638 <+1932>: bl     0x1c4366080               ; objc_msgSend
->  0x1c86c863c <+1936>: ldr    x0, [sp, #0x38]
    0x1c86c8640 <+1940>: bl     0x1c4384130               ; objc_release
    0x1c86c8644 <+1944>: mov    x0, x22
    0x1c86c8648 <+1948>: bl     0x1c4384130               ; objc_release
```

My codes
https://github.com/inoutch/tearchan/blob/03edc923569011a0c8e10ac97f1e0e77e5a4574d/src/core/engine.rs#L67-L93

Inexplicably, if I wait for a few seconds after app starting and tap later, it doesn't finish, and it also get the touch event successfully. Also, I use gfx-rs in addition to winit-rs and do present surface or not, an error may or may not occur (This is not related to gfx-rs, and to timing issue?)

### LINKED ISSUE #1705: Crash in iOS when swiping up for the control center
On iOS 12.4.8 on my iPhone 6, using Bevy on winit, if I swipe up for the control center immediately after app load I get a crash. If I first touch the screen to fire some touch events before trying to open the control center, then there is no crash and everything works as normal.

I'm trying to reproduce this directly on winit, but haven't been able to yet. Thought I'd post first anyway to see if this sounds familiar. I've googled various parts of the stack trace with no success.

It's really bizarre because the crash is inside iOS libraries. Perhaps the gesture recogniser is registered on the winit-defined UIView / UIViewController and some interaction there is the problem. Unfortunately without any source to the stack trace I don't know what this NSArray is. I suspect it's just accumulating touch events as part of the "delay touch near edge of screen" logic.

```
Exception	NSException *	"*** -[__NSArrayM insertObject:atIndex:]: object cannot be nil"	0x0000000282d89110

* thread #1, queue = 'com.apple.main-thread', stop reason = breakpoint 1.1
    frame #0: 0x00000001bd3239c0 libobjc.A.dylib`objc_exception_throw
    frame #1: 0x00000001be0c4bec CoreFoundation`_CFThrowFormattedException + 112
    frame #2: 0x00000001be0358ac CoreFoundation`-[__NSArrayM insertObject:atIndex:] + 1212
    frame #3: 0x00000001ea565868 UIKitCore`-[UIGestureRecognizer _delayTouch:forEvent:] + 216
    frame #4: 0x00000001ea5813c0 UIKitCore`-[_UISystemGestureGateGestureRecognizer _delayTouch:forEvent:] + 140
    frame #5: 0x00000001ea565a68 UIKitCore`-[UIGestureRecognizer _delayTouchesForEvent:inPhase:] + 220
    frame #6: 0x00000001ea565c98 UIKitCore`-[UIGestureRecognizer _delayTouchesForEventIfNeeded:] + 100
    frame #7: 0x00000001ea566a9c UIKitCore`-[UIGestureRecognizer _updateGestureWithEvent:buttonEvent:] + 504
    frame #8: 0x00000001ea55ac78 UIKitCore`_UIGestureEnvironmentUpdate + 2180
    frame #9: 0x00000001ea55a3a8 UIKitCore`-[UIGestureEnvironment _deliverEvent:toGestureRecognizers:usingBlock:] + 384
    frame #10: 0x00000001ea55a188 UIKitCore`-[UIGestureEnvironment _updateForEvent:window:] + 204
    frame #11: 0x00000001ea9727d0 UIKitCore`-[UIWindow sendEvent:] + 3112
    frame #12: 0x00000001ea95285c UIKitCore`-[UIApplication sendEvent:] + 340
    frame #13: 0x00000001eaa189d4 UIKitCore`__dispatchPreprocessedEventFromEventQueue + 1768
    frame #14: 0x00000001eaa1b100 UIKitCore`__handleEventQueueInternal + 4828
    frame #15: 0x00000001eaa14330 UIKitCore`__handleHIDEventFetcherDrain + 152
    frame #16: 0x00000001be0dcf1c CoreFoundation`__CFRUNLOOP_IS_CALLING_OUT_TO_A_SOURCE0_PERFORM_FUNCTION__ + 24
    frame #17: 0x00000001be0dce9c CoreFoundation`__CFRunLoopDoSource0 + 88
    frame #18: 0x00000001be0dc784 CoreFoundation`__CFRunLoopDoSources0 + 176
    frame #19: 0x00000001be0d76c0 CoreFoundation`__CFRunLoopRun + 1004
    frame #20: 0x00000001be0d6fb4 CoreFoundation`CFRunLoopRunSpecific + 436
    frame #21: 0x00000001c02d879c GraphicsServices`GSEventRunModal + 104
    frame #22: 0x00000001ea938c38 UIKitCore`UIApplicationMain + 212
  * frame #23: 0x0000000101dc0910 Callisto`cart_tmp_winit::platform_impl::platform::event_loop::EventLoop$LT$T$GT$::run::h2a3bed959aaaa659(self=EventLoop<()> @ 0x000000016ef50bb0, event_handler=closure-1 @ 0x000000016ef50bd0) at event_loop.rs:116:13
    frame #24: 0x0000000101de5b4c Callisto`cart_tmp_winit::event_loop::EventLoop$LT$T$GT$::run::h8c8ef82fc7ef7463(self=<unavailable>, event_handler=<unavailable>) at event_loop.rs:149:9
    frame #25: 0x0000000101de1724 Callisto`bevy_winit::run::h215a617795758f71(event_loop=<unavailable>, event_handler=<unavailable>) at lib.rs:43:5
    frame #26: 0x0000000101dd1d28 Callisto`bevy_winit::winit_runner::h90a304e554acabdc(app=App @ 0x000000016ef51c28) at lib.rs:244:9
    frame #27: 0x0000000101dda810 Callisto`core::ops::function::Fn::call::ha1caefc47bb26836((null)=0x0000000000000001, (null)=(bevy_app::app::App) @ 0x000000016ef51c28) at function.rs:70:5
    frame #28: 0x0000000102de6698 Callisto`_$LT$alloc..boxed..Box$LT$F$GT$$u20$as$u20$core..ops..function..Fn$LT$A$GT$$GT$::call::h88bacb7a918cef10(self=0x000000016ef52628, args=(bevy_app::app::App) @ 0x000000016ef51f58) at boxed.rs:1056:9
    frame #29: 0x0000000102de8ca8 Callisto`bevy_app::app::App::run::h0e84ac1eed74c29a(self=App @ 0x000000016ef52ee8) at app.rs:80:9
    frame #30: 0x0000000102dcf734 Callisto`bevy_app::app_builder::AppBuilder::run::hed1780d15d7a0af0(self=0x000000016ef53558) at app_builder.rs:45:9
    frame #31: 0x0000000100f06808 Callisto`bevy_main at lib.rs:48:5
    frame #32: 0x0000000100eb3d5c Callisto`main at main.swift:17:1
    frame #33: 0x00000001bdb9a8e0 libdyld.dylib`start + 4
```

bigmstone:
With IOS' PlatformSpeccificWindowBuilderAttributes the
preferred_screen_edges_deferring_system_gestures value was previously
set to the default (u8 default). This was an invalid value and caused
different crashes under screen edge events. Changing this value to a
valid value from the ScreenEdge enum fixes these issues.

Fixes #1705 and fixes #1613

### DIFF
--- a/src/platform_impl/ios/window.rs
+++ b/src/platform_impl/ios/window.rs
@@ -648,7 +648,7 @@ impl Default for PlatformSpecificWindowBuilderAttributes {
             valid_orientations: Default::default(),
             prefers_home_indicator_hidden: false,
             prefers_status_bar_hidden: false,
-            preferred_screen_edges_deferring_system_gestures: Default::default(),
+            preferred_screen_edges_deferring_system_gestures: ScreenEdge::ALL,
         }
     }
 }


bigmstone:
I have this as draft while I get more 👀 on it. It's a fairly low risk change for the repo since it only affects IOS and has a pretty obvious/easy to understand outcome. I have tested it on iPhone 12 pro or whatever they're calling it these days with iOS 14.3 and soon to be 14.4. I can't test on simulator on my mac since it's M1 and that workflow is broken right now (last I put any effort into it) w/ Rust and cargo-lips. If anyone else can test that would be great. If not I can take this out of draft if maintainers are 👍 for that.

MichaelHills:
Nice, works for me my iPhone 11 on Bevy. Changing it to `ScreenEdge::NONE` works too. Why is the derived default invalid?

bigmstone:
I am somewhat forced to guess why since we don't have the source for the portion of code that triggers the issue. It's entirely possible that this is a bug on Apple's side. Before I take this out of draft I need to create a blank view purely in objective-c or swift with `UIRectEdge` set to 0. Also my description/title in the PR actually needs to be changed a bit. Since the default value of u8 is 0 it's still a valid value of `UIRectEdge`. The Apple code is opaque -- so I haven't found a clear understanding yet of why this value results in an index offset in what appears to be they system gesture handler. Happy to have some extra context there if anyone has it.

bigmstone:
Thinking through this a bit more I think this PR is only masking the problem. Not solving it. Without this change I notice that if I touch on a non-edge then later edge touches are handled correctly. As though something was uninitialized and that first touch initializes things. In my search for a solution to this problem I looked a bit at the touch handler but didn't see anything obvious that stood out as being initialized there. This change is essentially making all of the initial touches (even those on the edge) handled by winit first. Then deferred to the system. Making my hypothesis that something is uninitialized stronger.

Anyone have thoughts on what on the winit side is either uninitialized or a call we make to initialize something when a touch is handled?

bigmstone:
@MichaelHills changing to `NONE` doesn't work on my end. I overlooked that part of your comment somehow at first. It fails in the same way. Can you double check that on your end?

MichaelHills:
I'll try to reproduce the original problem again tonight using `Default::default()` and verify

nmpribeiro:
This PR fixes my iOS crashes whenever I touch my app.

madsmtm:
`Default::default()` returns `ScreenEdge::NONE`, see [the relevant `bitflags` documentation](https://docs.rs/bitflags/1.2.1/bitflags/#default) together with the [the `ScreenEdge` definition](https://github.com/rust-windowing/winit/blob/master/src/platform/ios.rs#L303-L317) - so it's _not_ because it's an invalid value!

So as you said, this PR is only masking the problem, but nice find, it gets us closer to the root of the problem!

[The documentation](https://developer.apple.com/documentation/uikit/uiviewcontroller/2887512-preferredscreenedgesdeferringsys?language=objc) on `preferredScreenEdgesDeferringSystemGestures` says:
> Whenever possible, you should allow the system gestures to take precedence. However, immersive apps can use this property to allow app-defined gestures to take precedence over the system gestures.

So I'm weary against changing the default to making the application take precedence, but if we can't find the root cause it could be a temporary solution until we do?

bigmstone:
I ended up discovering the same docs back when I was looking at this too and reached the same conclusion -- the default behavior is what you'd expect "out of the box" from winit or similar. I wasn't able to find a clear smoking gun on this one. Ended up just using the release version of winit with the config for ScreenEdge set. I'm of the mind that I should close this PR. The discussion is likely useful to help whomever ends up solving it, but it should be linked in the open issues tracking this.

Issue #141, opened 2017-02-14
Handle panics in the user callbacks


tomaka:
Add `panic::catch_unwind` around the callback calls.

mitchmindtree:
I think this will only apply to certain user-callback calls, and will probably require being implemented uniquely by each backend.

For example in the macOS backend most user-callback calls occur on the main rust thread, while a certain few (i.e. `Resized`, `Focused`) are called by objective-C `NSWindowDelegate` callbacks. As a result probably only the delegate callback calls need to be wrapped in `catch_unwind`.

@tomaka In this case, what should be done once the unwind has been caught? I'm thinking that either:

1. The error could be stored somewhere so that the `panic!` can be re-invoked once execution is passed back to the main rust thread. One problem with this is that if a `panic!` occurs during window resizing, the `panic!`ing user callback will be called again and again until the resize has ended and only *then* will the `panic!` occur on the main rust thread.
2. Some objective-C equivalent to `panic!` is called. Is this even safe though? It would probably be weird for the user to sometimes get rust `panic!`s and other times get objective-C panics. For this reason I'm probably leaning towards 1 despite the resizing issue.

tasogare3710:
I encountered a need to handle errors at the caller of EventsLoop::poll_events/run_forever.

I think that I have something to do with this issue. 

Rather than using panic::catch_unwind, I think it is a good solution to add the try_poll_events/try_run_forever method and return the error value to the caller and handle it appropriately.

mtak-:
After EventLoop 2.0, there will be no returning from `run` on certain platforms. On those platforms, I'm pretty sure this means there will be no good time to `resume_unwind`.

One idea is to catch the panic and pass it into the callback. Something like changing the callback to take a `Result<Event<T>, Panicked>` parameter instead of an `Event<T>`.

```rs
struct Panicked(Box<Any + Send + 'static>);
```

Osspial:
What platforms does this still need to be implemented on? This has been fixed on Windows,

I'd like to split this into separate issues for each platform, for improved tracking.

vberger:
What is the exact need?

With evl2, the x11 and Wayland backend do not invoke the user callback from withing a system callback, so panics simply unwind the rust app without crossing FFI frontier. Is anything ore needed?

Osspial:
For Linux, no. This may be necessary on iOS, Android, macOS, and the web, though.

madsmtm:
[...]

PR #1853, opened 2021-02-06
Fix for closure-captured values not being dropped on panic

### LINKED ISSUE #1850: Closure-captured values aren't dropped on panic
This is similar to #1058, but unlike a normal termination, I'm talking about panics. A panic inside the closure is something that can happen in a user application, and it would generally expect the involved objects to be dropped in order to do safe cleanup. This is blocking https://github.com/gfx-rs/wgpu/issues/861

ArturKovacs:

Fixes #1850

This moves the ownership of the callback-closure to the `EventLoop`. The colusre is now contained within an `Rc` and the global handler only has a `Weak` reference to it, in order to ensure that the closure gets dropped with the `EventLoop`. Furthermore this catches panic unwinding, coming from the user callback to prevent it from unwinding into foreign code. The application is then requested to stop and the caught unwinding is resumed at a point where the callstack doesn't contain foreign code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - On Android, unimplemented events are marked as unhandled on the native event loop.
 - On Windows, added `WindowBuilderExtWindows::with_menu` to set a custom menu at window creation time.
 - On Android, bump `ndk` and `ndk-glue` to 0.3: use predefined constants for event `ident`.
+- On macOS, fix event-closure-captured objects not being dropped on panic.
 
 # 0.24.0 (2020-12-09)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 - On Android, unimplemented events are marked as unhandled on the native event loop.
 - On Windows, added `WindowBuilderExtWindows::with_menu` to set a custom menu at window creation time.
 - On Android, bump `ndk` and `ndk-glue` to 0.3: use predefined constants for event `ident`.
-- On macOS, fix event-closure-captured objects not being dropped on panic.
+- On macOS, fix objects captured by the event loop closure not being dropped on panic.
 
 # 0.24.0 (2020-12-09)
 
--- a/src/platform_impl/macos/app_state.rs
+++ b/src/platform_impl/macos/app_state.rs
@@ -249,7 +249,6 @@ pub static INTERRUPT_EVENT_LOOP_EXIT: AtomicBool = AtomicBool::new(false);
 pub enum AppState {}
 
 impl AppState {
-    // This function extends lifetime of `callback` to 'static as its side effect
     pub fn set_callback<T>(
         callback: Weak<Mutex<Box<dyn FnMut(Event<'_, T>, &RootWindowTarget<T>, &mut ControlFlow)>>>,
         window_target: Rc<RootWindowTarget<T>>,
--- a/src/platform_impl/macos/app_state.rs
+++ b/src/platform_impl/macos/app_state.rs
@@ -26,6 +26,7 @@ use crate::{
     event_loop::{ControlFlow, EventLoopWindowTarget as RootWindowTarget},
     platform_impl::platform::{
         event::{EventProxy, EventWrapper},
+        event_loop::CURRENT_PANIC,
         observer::EventLoopWaker,
         util::{IdRef, Never},
         window::get_window_id,
@@ -285,6 +286,9 @@ impl AppState {
     }
 
     pub fn wakeup() {
+        if CURRENT_PANIC.lock().unwrap().is_some() {
+            return;
+        }
         if !HANDLER.is_ready() {
             return;
         }
@@ -338,6 +342,9 @@ impl AppState {
     }
 
     pub fn cleared() {
+        if CURRENT_PANIC.lock().unwrap().is_some() {
+            return;
+        }
         if !HANDLER.is_ready() {
             return;
         }
--- a/src/platform_impl/macos/app_state.rs
+++ b/src/platform_impl/macos/app_state.rs
@@ -1,12 +1,13 @@
 use std::{
+    cell::{RefCell, RefMut},
     collections::VecDeque,
     fmt::{self, Debug},
     hint::unreachable_unchecked,
     mem,
-    rc::Rc,
+    rc::{Rc, Weak},
     sync::{
         atomic::{AtomicBool, Ordering},
-        Mutex, MutexGuard, Weak,
+        Mutex, MutexGuard,
     },
     time::Instant,
 };
@@ -52,7 +53,7 @@ pub trait EventHandler: Debug {
 }
 
 struct EventLoopHandler<T: 'static> {
-    callback: Weak<Mutex<Box<dyn FnMut(Event<'_, T>, &RootWindowTarget<T>, &mut ControlFlow)>>>,
+    callback: Weak<RefCell<Box<dyn FnMut(Event<'_, T>, &RootWindowTarget<T>, &mut ControlFlow)>>>,
     will_exit: bool,
     window_target: Rc<RootWindowTarget<T>>,
 }
@@ -62,11 +63,11 @@ impl<T> EventLoopHandler<T> {
     where
         F: FnOnce(
             &mut EventLoopHandler<T>,
-            MutexGuard<'_, Box<dyn FnMut(Event<'_, T>, &RootWindowTarget<T>, &mut ControlFlow)>>,
+            RefMut<'_, Box<dyn FnMut(Event<'_, T>, &RootWindowTarget<T>, &mut ControlFlow)>>,
         ),
     {
         if let Some(callback) = self.callback.upgrade() {
-            let callback = callback.lock().unwrap();
+            let callback = callback.borrow_mut();
             (f)(self, callback);
         } else {
             panic!(
@@ -254,7 +255,9 @@ pub enum AppState {}
 
 impl AppState {
     pub fn set_callback<T>(
-        callback: Weak<Mutex<Box<dyn FnMut(Event<'_, T>, &RootWindowTarget<T>, &mut ControlFlow)>>>,
+        callback: Weak<
+            RefCell<Box<dyn FnMut(Event<'_, T>, &RootWindowTarget<T>, &mut ControlFlow)>>,
+        >,
         window_target: Rc<RootWindowTarget<T>>,
     ) {
         *HANDLER.callback.lock().unwrap() = Some(Box::new(EventLoopHandler {
--- a/src/platform_impl/macos/app_state.rs
+++ b/src/platform_impl/macos/app_state.rs
@@ -24,7 +24,7 @@ use crate::{
     event_loop::{ControlFlow, EventLoopWindowTarget as RootWindowTarget},
     platform_impl::platform::{
         event::{EventProxy, EventWrapper},
-        event_loop::{post_dummy_event, CURRENT_PANIC},
+        event_loop::{post_dummy_event, PanicInfo},
         observer::EventLoopWaker,
         util::{IdRef, Never},
         window::get_window_id,
@@ -250,7 +250,7 @@ impl Handler {
 
 pub static INTERRUPT_EVENT_LOOP_EXIT: AtomicBool = AtomicBool::new(false);
 
-pub enum AppState {}
+pub(crate) enum AppState {}
 
 impl AppState {
     pub fn set_callback<T>(
@@ -281,8 +281,9 @@ impl AppState {
         HANDLER.set_in_callback(false);
     }
 
-    pub fn wakeup() {
-        if CURRENT_PANIC.lock().unwrap().is_some() || !HANDLER.is_ready() {
+    pub fn wakeup(panic_info: Weak<PanicInfo>) {
+        let panic_info = panic_info.upgrade().unwrap();
+        if panic_info.is_panicking() || !HANDLER.is_ready() {
             return;
         }
         let start = HANDLER.get_start_time().unwrap();
@@ -334,8 +335,9 @@ impl AppState {
         HANDLER.events().append(&mut wrappers);
     }
 
-    pub fn cleared() {
-        if CURRENT_PANIC.lock().unwrap().is_some() || !HANDLER.is_ready() {
+    pub fn cleared(panic_info: Weak<PanicInfo>) {
+        let panic_info = panic_info.upgrade().unwrap();
+        if panic_info.is_panicking() || !HANDLER.is_ready() {
             return;
         }
         if !HANDLER.get_in_callback() {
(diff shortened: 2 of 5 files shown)


ArturKovacs:
@francesca64 do you mind giving this a look? Also I don't want to necessarily always ask you to be the "other maintainer" when reviewing macOS PRs but I don't know of any other maintainer who has experience with macOS.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 - On Android, unimplemented events are marked as unhandled on the native event loop.
 - On Windows, added `WindowBuilderExtWindows::with_menu` to set a custom menu at window creation time.
 - On Android, bump `ndk` and `ndk-glue` to 0.3: use predefined constants for event `ident`.
+- On macOS, fix event-closure-captured objects not being dropped on panic.
```
maroider, line 11:
```suggestion
- On macOS, fix objects captured by the event loop closure not being dropped on panic.
```
I'm not too sure if this reads better to everyone, but it does to me.

maroider (changes requested):
Hello!
I come bearing nit-picks!

---

On a more serious note: one thing that does worry me here is whether it is possible to panic across FFI boundaries from the callback.

CHANGELOG.md
ArturKovacs: Yes, this seems easier to understand.

src/platform_impl/macos/app_state.rs
```diff
@@ -68,23 +68,43 @@ impl<T> Debug for EventLoopHandler<T> {
...
+        if let Some(callback) = self.callback.upgrade() {
+            let mut callback = callback.lock().unwrap();
+            (callback)(event.userify(), &self.window_target, control_flow);
+            self.will_exit |= *control_flow == ControlFlow::Exit;
+            if self.will_exit {
+                *control_flow = ControlFlow::Exit;
+            }
+        } else {
+            // Logging an error instead of panicing because might happen while the
+            // application is already panicing.
```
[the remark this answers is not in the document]
ArturKovacs: Thanks!

src/platform_impl/macos/event_loop.rs
```diff
@@ -50,6 +55,16 @@ impl<T: 'static> EventLoopWindowTarget<T> {
...
     window_target: Rc<RootWindowTarget<T>>,
+
+    /// We make sure that the callback closure is dropped during a panic
+    /// by making the event loop own it.
+    ///
+    /// Every other reference should be a Weak reference which is only upgraded
+    /// into a strong reference in order to call the callback but then the
+    /// strong reference should be dropped as soon as possible.
+    callback:
+        Option<Arc<Mutex<Box<dyn FnMut(Event<'_, T>, &RootWindowTarget<T>, &mut ControlFlow)>>>>,
```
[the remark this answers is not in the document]
ArturKovacs: My assumption was that this has to be `Arc<Mutex>` because the callback is exposed through the global `HANDLER` object. But apparently the borrow checker is still happy if this is just `Rc<RefCell>` which is a complete mystery to me and it makes be a bit worried.

ArturKovacs:
> I come bearing nit-picks!

Welcome here :D

> On a more serious note: one thing that does worry me here is whether it is possible to panic across FFI boundaries from the callback.

Unfortunately it is. The stack trace for events on macOS is approximately
```
- "event handler callback"
- app_state::EventLoopHandler::handle_nonuser_event
- app_state::Handler::handle_nonuser_event
- app_state::AppState::cleared
- observer::control_flow_end_handler
- "foreign code"
...
- event_loop::EventLoop::run
```

Although when I tested, it seemed to work as expected but considering that it's undefined it's not exactly a solution. As you noted earlier, this ties back into #141. I'll give this a bit more thought.

maroider:
The Windows backend's `EventLoopRunner` may serve as a source of inspiration.

maroider:
Is there anything stopping you from storing it in the `EventLoop`?

src/platform_impl/macos/event_loop.rs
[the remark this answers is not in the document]
francesca64: I _vaguely_ remember there being something particular about `HANDLER` only being accessed on one thread, but this was over 2 years ago and I frankly don't have a mental model of this anymore.

I optimistically had a reason for putting non-thready things in `EventLoopHandler`, and apparently I also reviewed a PR that furthered that decision: #1108 

I really don't know anymore, though.

src/platform_impl/macos/event_loop.rs
```diff
@@ -114,6 +164,49 @@ impl<T> EventLoop<T> {
...
+            }
+            unsafe {
+                let app_class = class!(NSApplication);
+                let app: id = msg_send![app_class, sharedApplication];
+                let _: () = msg_send![app, stop: nil];
+
+                // Posting an dummy event to get stop to take effect immediately.
+                // See: https://stackoverflow.com/questions/48041279/stopping-the-nsapplication-main-event-loop/48064752#48064752
+                let event_class = class!(NSEvent);
+                let dummy_event: id = msg_send![
```
francesca64, line 191: It looks like there's code duplication here with the dummy event in `AppState::cleared`, which would be great to consolidate!

src/platform_impl/macos/event_loop.rs
```diff
@@ -114,6 +164,49 @@ impl<T> EventLoop<T> {
...
+            {
+                let mut panic = CURRENT_PANIC.lock().unwrap();
+                if panic.is_none() {
+                    *panic = Some(e);
+                }
+            }
+            unsafe {
+                let app_class = class!(NSApplication);
+                let app: id = msg_send![app_class, sharedApplication];
+                let _: () = msg_send![app, stop: nil];
```
francesca64, line 186: I know this isn't consistently used yet, but this is equivalent:
```suggestion
                let () = msg_send![app, stop: nil];
```

src/platform_impl/macos/event_loop.rs
```diff
@@ -50,6 +65,16 @@ impl<T: 'static> EventLoopWindowTarget<T> {
...
 pub struct EventLoop<T: 'static> {
     window_target: Rc<RootWindowTarget<T>>,
+
+    /// We make sure that the callback closure is dropped during a panic
+    /// by making the event loop own it.
+    ///
+    /// Every other reference should be a Weak reference which is only upgraded
+    /// into a strong reference in order to call the callback but then the
+    /// strong reference should be dropped as soon as possible.
+    _callback:
```
francesca64, line 75: What's the motive for prefixing this with an underscore? It looks like this field is accessed below.

francesca64 (changes requested):
Looks great! I won't be able to test it out until next week, though.

Thanks so much for fixing this!

> I don't want to necessarily always ask you to be the "other maintainer" when reviewing macOS PRs but I don't know of any other maintainer who has experience with macOS.

No worries! You can always ping me, but know that I don't have a ton of bandwidth for winit, so you'll get to hear me apologize for delays a bunch!

I'm not aware of anyone else around who's familiar with macOS, though @kchibisov has reviewed macOS PRs in the past. In any case, feel free to let me know if you see anyone around who would be good to make a Member too!

src/platform_impl/macos/event_loop.rs
ArturKovacs: The nightly compiler emits a warning, because this is never read. https://github.com/rust-windowing/winit/pull/1853/checks?check_run_id=1892520990#step:9:63

ArturKovacs:
> Is there anything stopping you from storing it in the `EventLoop`?

I moved it into the `EventLoop`. To ensure that the conversion from the raw pointer (at `observer::control_flow_end_handler`) into a rust object is safe, I used `Rc` and `Weak` here as well. Earlier we had a "Mutex locking" which we replaced with upgrading a `Weak` to an `Rc`. Considering that the Mutex (in theory) never experienced contention, I believe that this improvement is negligible regarding runtime speed. But now it's done so there you go.

Although there's one thing I'm entirely sure about. I had to use `AssertUnwindSafe` on the `Weak` because the standard library doesn't have a blanket implementation for `UnwindSafe` for `Weak<T>` even if T implements `RefUnwindSafe`. Whereas `Rc` does have this blanket implementation so I'm a bit confused.

> I'm not aware of anyone else around who's familiar with macOS, though @kchibisov has reviewed macOS PRs in the past.

Thanks!

> In any case, feel free to let me know if you see anyone around who would be good to make a Member too!

Will do!

src/platform_impl/macos/event_loop.rs
francesca64: Ah, gotcha! Thanks for improving my understanding of the rules around that.

ArturKovacs:
@francesca64 Let me know approximately when you are available, and then I'll fix the conflicts so that you can review this.

src/platform_impl/macos/observer.rs
```diff
@@ -152,31 +192,41 @@ impl RunLoop {
...
             handler,
-            ptr::null_mut(),
+            context,
         );
         CFRunLoopAddObserver(self.0, observer, kCFRunLoopCommonModes);
     }
 }
 
-pub fn setup_control_flow_observers() {
+pub(crate) fn setup_control_flow_observers(panic_info: Weak<PanicInfo>) {
```
francesca64, line 209: This function isn't visible outside of `platform_impl::macos` either way, right? I'm not sure if adding `(crate)` makes that more explicit or if it makes the overall module structure and their relative visibilities less clear.

(This comment applies to other things this PR marks `pub(crate)` too)

src/platform_impl/macos/event_loop.rs
```diff
@@ -114,6 +191,54 @@ impl<T> EventLoop<T> {
...
+            {
+                let panic_info = panic_info.upgrade().unwrap();
+                panic_info.set_panic(e);
+            }
+            unsafe {
+                let app_class = class!(NSApplication);
+                let app: id = msg_send![app_class, sharedApplication];
+                let () = msg_send![app, stop: nil];
+
+                // Posting an dummy event to get stop to take effect immediately.
```
francesca64, line 233: Using backticks makes it easier not to read "stop" as a verb:
```suggestion
                // Posting a dummy event to get `stop` to take effect immediately.
```

src/platform_impl/macos/app_state.rs
```diff
@@ -265,8 +281,9 @@ impl AppState {
         HANDLER.set_in_callback(false);
     }
 
-    pub fn wakeup() {
-        if !HANDLER.is_ready() {
+    pub fn wakeup(panic_info: Weak<PanicInfo>) {
+        let panic_info = panic_info.upgrade().unwrap();
```
francesca64, line 285: If these `unwrap`s are more-or-less unreachable, then making them `expect`s that indicate that this panicking is a developer error / invariant violation would be great.

src/platform_impl/macos/event_loop.rs
```diff
@@ -23,6 +33,38 @@ use crate::{
...
 
+//pub type PanicData = Option<Box<dyn Any + Send + 'static>>;
+
+#[derive(Default)]
+pub(crate) struct PanicInfo {
+    inner: Cell<Option<Box<dyn Any + Send + 'static>>>,
+}
+// ------------------------------------
+// WARNING:
+// This struct is UnwindSafe if `get_mut` is never called on the inner data.
```
francesca64, line 44: If I'm reading this right, we're safe from this as long as no `get_mut` usages are in that `PanicInfo` impl block below, or is the scope larger than that?

src/platform_impl/macos/event_loop.rs
```diff
@@ -98,14 +152,37 @@ impl<T> EventLoop<T> {
...
+        // they passed to callback will actually outlive it, some apps just can't move
+        // everything to event loop, so this is something that they should care about.
+        let callback = unsafe {
+            mem::transmute::<
+                Rc<RefCell<dyn FnMut(Event<'_, T>, &RootWindowTarget<T>, &mut ControlFlow)>>,
+                Rc<RefCell<dyn FnMut(Event<'_, T>, &RootWindowTarget<T>, &mut ControlFlow)>>,
+            >(Rc::new(RefCell::new(callback)))
+        };
+
+        self._callback = Some(callback.clone());
```
francesca64, line 166: Consistency:
```suggestion
        self._callback = Some(Rc::clone(&callback));
```

francesca64 (changes requested):
I've got some deadlines looming, so I won't be around much this month. I'm taking today to catch up on my github backlog though, so sorry for reviewing this before giving you a chance to fix conflicts/etc.

src/platform_impl/macos/observer.rs
```diff
@@ -103,41 +124,60 @@ pub struct CFRunLoopSourceContext {
     pub perform: Option<extern "C" fn(*mut c_void)>,
 }
 
+fn control_flow_handler<F>(panic_info: *mut c_void, f: F)
```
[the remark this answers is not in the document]
ArturKovacs: It damn well does :D

src/platform_impl/macos/observer.rs
```diff
@@ -1,6 +1,18 @@
-use std::{self, os::raw::*, ptr, time::Instant};
+use core::panic;
```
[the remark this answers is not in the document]
ArturKovacs: It undeniably seems that way :D

src/platform_impl/macos/observer.rs
```diff
@@ -152,31 +192,41 @@ impl RunLoop {
...
         CFRunLoopAddObserver(self.0, observer, kCFRunLoopCommonModes);
     }
 }
 
-pub fn setup_control_flow_observers() {
+pub(crate) fn setup_control_flow_observers(panic_info: Weak<PanicInfo>) {
     unsafe {
+        let mut context = CFRunLoopObserverContext {
+            info: Weak::into_raw(panic_info) as *mut _,
+            version: 0,
```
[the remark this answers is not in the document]
ArturKovacs: Yeah putting the documentation there makes sense.

src/platform_impl/macos/observer.rs
ArturKovacs: As far as I understand, if these are just `pub` then they can still be reached from outside the crate. This also limits the optimizations that can be done when compiling the crate.

But I don't feel strongly about this, so I'm changing them to `pub`.

src/platform_impl/macos/event_loop.rs
ArturKovacs: I think you are reading this right. Of course, the rest of the module can also directly access the private `inner` so that's why I choose the specific wording. Although I can see that it's a strange way to say what I'm trying to say here, so I just made an attempt at improving this.

francesca64: This reads very clearly now, thanks!

francesca64 (approved):
Everything looks awesome! Thanks again for fixing this.

(I also saw the reddit thread you made about `Weak`'s unwind safety; you have great initiative!)

ArturKovacs:
> (I also saw the reddit thread you made about `Weak`'s unwind safety; you have great initiative!)

Oh stop it, you 😊

PR #1858, opened 2021-02-10
Add ability to hook/wrap an EventLoop.

lachlansneff:
I've been thinking about how to get wgpu to integrate into the winit event loop (specifically to avoid needing users to continually call `device.poll` in order for gpu buffers to actually get mapped).

Anyhow, there were two ways of doing this:

1. Add some special casing to winit for wgpu that will call `device.poll` once per frame.
2. Add generalized hooking ability to winit.

The second option seemed preferable to me. Another issue is that, if `control_flow` is set to `Wait`, the event handler will only be called if the user interacts. That's not good enough in this case, so the hook wraps the user-specified event handler so it can detect if the handler set the control flow to `Wait` and change it to `WaitUntil` or similar instead.

The main things added are the `Hook` trait and `EventLoop::set_hook`.

`EventLoop::set_hook` actually consumes the `EventLoop` and returns a new type of `EventLoop<T, H>`, where `H: Hook<T>`.

### DIFF
--- a/examples/hook.rs
+++ b/examples/hook.rs
@@ -1,12 +1,21 @@
 use simple_logger::SimpleLogger;
-use winit::{event::{ElementState, Event, WindowEvent}, event_loop::{ControlFlow, EventLoop, EventLoopWindowTarget, Hook}, window::WindowBuilder};
+use winit::{
+    event::{ElementState, Event, WindowEvent},
+    event_loop::{ControlFlow, EventLoop, EventLoopWindowTarget, Hook},
+    window::WindowBuilder,
+};
 
 struct ExampleHook;
 
 impl<T: std::fmt::Debug> Hook<T> for ExampleHook {
-    fn run<F>(&mut self, handler: F, event: Event<'_, T>, target: &EventLoopWindowTarget<T>, control_flow: &mut ControlFlow)
-    where
-        F: FnOnce(Event<'_, T>, &EventLoopWindowTarget<T>, &mut ControlFlow)
+    fn run<F>(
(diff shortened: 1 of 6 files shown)


ArturKovacs:
I see why this would be necessary. I don't quite understand the situation maybe because I've never used wgpu but it seems to me that this sort of thing is not supposed to be managed by winit. I believe you could for example structure you code like this to achieve the same effect:
```rust
struct MyHook {
    // Whatever state you need to store for the hook
}
impl MyHook {
    fn got_event<T, F>(
        &mut self,
        event: Event<'_, T>,
        target: &EventLoopWindowTarget<T>,
        control_flow: &mut ControlFlow
        handler: F,
    ) {
        // do before handler stuff
        handler(event, target, control_flow);
        // do after handler stuff
    }
}
fn main() {
        .build(&event_loop)
        .unwrap();
    let mut hook = MyHook {};
    event_loop.run(move |event, target, control_flow| {
        hook.got_event(event, target, control_flow, |event, _ control_flow| {
            println!("in handler: {:?}", event);
            match event {
                ...
            }
        });
    });
}
```

maroider:
> specifically to avoid needing users to continually call `device.poll` in order for gpu buffers to actually get mapped

Is this still a problem, actually? I believe I've read that this limitation has been fixed or something like that?

madsmtm:
We're working on alternative solutions to this in https://github.com/rust-windowing/winit/issues/3432 and https://github.com/rust-windowing/winit/issues/3474, this PR is now obsolete.

PR #1868, opened 2021-02-25
Add window focus extension for macos and windows

xiaopengli89:
I add window focus api on platform macos and windows.

### DIFF
--- a/src/platform_impl/macos/window.rs
+++ b/src/platform_impl/macos/window.rs
@@ -1155,14 +1155,14 @@ impl WindowExtMacOS for UnownedWindow {
     #[inline]
     fn focus(&self) {
         unsafe {
-            self.ns_window.makeKeyWindow();
+            util::make_key_window_async(*self.ns_window);
         }
     }
 
     #[inline]
     fn focus_and_front(&self) {
         unsafe {
-            self.ns_window.makeKeyAndOrderFront_(nil);
+            util::make_key_and_order_front_async(*self.ns_window);
         }
     }
 }
--- a/src/platform_impl/macos/window.rs
+++ b/src/platform_impl/macos/window.rs
@@ -1151,6 +1151,20 @@ impl WindowExtMacOS for UnownedWindow {
                 .setHasShadow_(if has_shadow { YES } else { NO })
         }
     }
+
+    #[inline]
+    fn focus(&self) {
+        unsafe {
+            self.ns_window.makeKeyWindow();
+        }
+    }
+
+    #[inline]
+    fn focus_and_front(&self) {
+        unsafe {
+            self.ns_window.makeKeyAndOrderFront_(nil);
+        }
+    }
 }
 
 impl Drop for UnownedWindow {
--- a/src/platform/macos.rs
+++ b/src/platform/macos.rs
@@ -38,6 +38,12 @@ pub trait WindowExtMacOS {
 
     /// Sets whether or not the window has shadow.
     fn set_has_shadow(&self, has_shadow: bool);
+
+    /// Focus the window
+    fn focus(&self);
+
+    /// Focus the window, and moves the window to the front of the screen list
+    fn focus_and_front(&self);
 }
 
 impl WindowExtMacOS for Window {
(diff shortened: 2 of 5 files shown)


msiglreith:
Thanks, related PR: https://github.com/rust-windowing/winit/pull/1767

artursapek:
It would be nice to get some form of this merged. Being able to bring a window to the front is a powerful option (to be used responsibly)

msiglreith (changes requested):
The given implementation doesn't really work for my test case on Windows:

Setting `focus_and_font` when receiving a device key, will focus and immediately unfocus the window again. It's still in the same z position and doesn't receive keyboard input. The other PR behaves more in line here with its implementation. But I haven't looked into why..

I don't know about the Mac parts but I'm more inclined to moving forward with the other one.

xiaopengli89:
I replaced the `winuser::SetFocus` function with `force_window_active`, and it works.

src/platform_impl/macos/window.rs
```diff
@@ -1151,6 +1151,20 @@ impl WindowExtMacOS for UnownedWindow {
...
+            self.ns_window.makeKeyWindow();
+        }
+    }
+
+    #[inline]
+    fn focus_and_front(&self) {
+        unsafe {
+            self.ns_window.makeKeyAndOrderFront_(nil);
+        }
+    }
```
madsmtm, line 1167: There's already a way to call this on MacOS, it's done by using `window.set_visible(true)` - so unsure what the benefits are of this?

src/platform_impl/macos/window.rs
```diff
@@ -1151,6 +1151,20 @@ impl WindowExtMacOS for UnownedWindow {
                 .setHasShadow_(if has_shadow { YES } else { NO })
         }
     }
+
+    #[inline]
+    fn focus(&self) {
+        unsafe {
+            self.ns_window.makeKeyWindow();
```
madsmtm, line 1158: Like most things on `NSWindow`, this can probably only be run on the main thread, see the related `macos::util::async::make_key_and_order_front_async`

madsmtm (changes requested):
Don't know about the Windows API, but I don' really see the purpose of the MacOS API - one of the functions are already possible, and the other is only useful if you'd want the window, which must be visible, to receive keyboard events, but without bringing it forwards - I can only imagine very few cases where that would be required, and in those cases you probably need a lot more control than `winit` gives currently.

src/platform_impl/macos/window.rs
xiaopengli89: Thanks for your reviews. You'are right, I just found that `window.set_visible(true)` is doing the same thing in this case, the purpose of this api is to focus a specific window and bring it front, may be used in multiple window situations. The`set_visible` function name just like to set the window's visibility as it does on windows platform.

src/platform_impl/macos/window.rs
xiaopengli89: Great, I will fix it according to your suggestion.

ArturKovacs:
I see two reasons #1767 is attractive 
1, that also provides an implementation for X11
2, @msiglreith prefers the Windows implementation of #1767 compared to this

I may be wrong, but I thought that @murarth, the X11 maintainer is not around. In that case we shouldn't try to pursue a PR that changes X11 because it won't be merged in the foreseeable future. And regarding the Windows implementation, this PR could take the Windows implementation from #1767.

All in all I think we should keep moving with this and again this should just use some code from #1767, if needed. What do you think @msiglreith?

msiglreith:
fine for me, I haven't yet tried the latest changes but will do so soon.

msiglreith:
Tested it and works fine in general. I would be interested what's the behavior on other platforms (ie MacOS) in case the window is minimized or invisible. In both cases input focus is grabbed on Windows but the window isn't shown.

ArturKovacs:
@msiglreith I have a feeling you haven't seen this comment: https://github.com/rust-windowing/winit/pull/1767#issuecomment-835373637

msiglreith:
https://github.com/rust-windowing/winit/pull/1944 got approved and merged. I hope this covers this use case as well, thanks nonethless for the work!

ArturKovacs:
Yes! Thank you for your work @xiaopengli89 it definitely helped to get momentum behind this feature!

xiaopengli89:
That's great!

PR #1874, opened 2021-03-02
Fix seg-fault when using without a window

### LINKED ISSUE #1869: (macOS) Trying to exit the event loop causes it to throw an exception and start polling infinitely instead
Here is a cool `main.rs`:

```rust
use std::time::{Instant, Duration};
use winit::event_loop::{EventLoop, ControlFlow};
use winit::window::WindowBuilder;

fn main() {
	let start = Instant::now();
	let eloop = EventLoop::new();

	// make macOS show the crash exception dialog by telling it that this is a GUI application
	let window = WindowBuilder::new().build(&eloop);
	drop(window);

	eloop.run(move |event, _, flow| {
		println!("{:?}", &event);
		*flow = ControlFlow::WaitUntil(Instant::now() + Duration::from_secs_f32(0.01));
		if (Instant::now() - start).as_secs_f64() > 1.0 {
			*flow = ControlFlow::Exit;
		}
	})
}
```

What's cool is that it shows an exception dialog and the loop doesn't stop when I tell it to! It actually starts polling which is _definitely_ not what I asked for. The exception dialog doesn't show if I just abort the process after setting `*flow` but obviously that isn't very safe and nothing gets a chance to clean up.

Hey, did you know that if you change `WaitUntil` to `Poll` then sometimes you'll get an unrecognized selector crash?

... what is going on here? xD

(no offense towards winit, this is honestly the first bug I've ever encountered in the most recent version)

ArturKovacs:

Fixes #1869

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- On macOS, fix segmentation fault after dropping the main window.
 - On Android, `InputEvent::KeyEvent` is partially implemented providing the key scancode.
 - Added `is_maximized` method to `Window`.
 - On Windows, fix bug where clicking the decoration bar would make the cursor blink.
--- a/src/platform_impl/macos/app_state.rs
+++ b/src/platform_impl/macos/app_state.rs
@@ -341,9 +341,7 @@ impl AppState {
             unsafe {
                 let app: id = NSApp();
                 let windows: id = msg_send![app, windows];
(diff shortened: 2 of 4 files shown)


ArturKovacs:
@kchibisov Please let me know if you are available to review this.

ArturKovacs:
Considering that kchibisov is currently not available, @francesca64 could you take a look at this?

francesca64 (approved):
Thanks a ton for catching this; looks great!

PR #1876, opened 2021-03-07
Only try publishing when a version tag is pushed

ArturKovacs:

Currently the publish workflow attempts to publish every time a change in the `Cargo.toml` is pushed to the `master`. The publish fails if the version already exists but the failure happens only after compiling winit, wasting compute time and sending a false "publish failed" notification to the maintainer who made the merge. I understand that it's not a "serious" problem but the fix is very simple, so I think it's worth applying.

Just tagging a few maintainers for a review:  @msiglreith, @vberger, @murarth

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -2,8 +2,8 @@ name: Publish
 
 on:
   push:
-    branches: [master]
-    paths: "Cargo.toml"
+    tags:
+    - 'v*' # Push events to matching v*, i.e. v1.0, v20.15.10
 
 jobs:
   Publish:


msiglreith (approved):

vberger (approved):

ArturKovacs:
Interesting... the tests won't start and GitHub won't let me merge. I'll close and open this again.

ArturKovacs:
I'm guessing it's a privilege thing. @francesca64 if you find this useful, could you merge this?

maroider:
It seems like `publish.yml` isn't [among the files](https://github.com/rust-windowing/winit/blob/889258f538010d4fa222e5e897fc0affd389dccf/.github/workflows/ci.yml#L3-L8) which trigger a CI run.

francesca64:
If I'm understanding correctly, this would change our release workflow from what it is now:

1. contributor bumps version in PR, which we merge
2. publish automatically happens
(1-to-1 correspondence between master branch versioning and crates.io)
3. we optimistically remember to tag a release on github

to this:
 
1. contributor bumps version in PR, which we merge
2. we optimistically remember to tag a release on github
3. if so, publish automatically happens
(1-to-1 correspondence between github releases and crates.io)

The existing workflow has been in place since before I even joined the project, IIRC, so it'd kinda be a big change. That said, this change would effectively guarantee that we reliably tag releases, which is probably a big enough boon to justify the friction of changing our procedure.

(This also has the side benefit of facilitating making "long-term support" releases of older minor versions, though I think we've only had to do that once.)

ArturKovacs:
I think your understanding is correct.

francesca64 (approved):

PR #1879, opened 2021-03-10
Pen tablet support on X11

tasgon:

This is based on @DorianRudolph's work for adding pen tablet support on Linux on X11. The only thing I changed was to have it send `Touch` events in a similar manner to how pen tablet input is handled on Windows. There are quite a few things to address before accepting this PR, but I'd like to get feedback on whether such a change would be accepted before cleaning it up.

### DIFF
--- a/src/platform_impl/linux/x11/event_processor.rs
+++ b/src/platform_impl/linux/x11/event_processor.rs
@@ -748,18 +748,20 @@ impl<T: 'static> EventProcessor<T> {
                                 };
 
                                 let pressure = get_value(stylus.pressure) / stylus.pressure.max;
-                                let tilt_x = if let Some(vi) = stylus.tilt_x { get_value(vi) } else { 0. };
-                                let tilt_y = if let Some(vi) = stylus.tilt_y { get_value(vi) } else { 0. };
+                                //let tilt_x = if let Some(vi) = stylus.tilt_x { get_value(vi) } else { 0. };
+                                //let tilt_y = if let Some(vi) = stylus.tilt_y { get_value(vi) } else { 0. };
+
+                                let touch = Touch {
+                                    device_id,
+                                    phase: TouchPhase::Moved,
+                                    location: position,
+                                    force: Some(crate::event::Force::Normalized(pressure)),
+                                    id: xev.sourceid as u64
+                                };
 
                                 callback(Event::WindowEvent {
                                     window_id,
-                                    event: StylusMoved {
-                                        device_id,
-                                        position,
-                                        pressure,
-                                        tilt_x,
-                                        tilt_y
-                                    }
+                                    event: crate::event::WindowEvent::Touch(touch),
                                 });
                             } else {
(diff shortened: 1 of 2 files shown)


codecnotsupported:
I'm curious as well what's blocking this merge. 
I gave it a shot and it has a few problems.
For one, it doesn't work with my Wacom tablet. 
Two it still needs formatting. 
And three, probably needs to get updated for the current version of Winit.

If you really need tablet support, I THINK WindowEvent::AxisMotion kinda already reports this. albeit not in pixel coordinate format. But in tablet space coordinate space.

irrenhaus:
`WindowEvent::AxisMotion` might work. I can read the pressure data from there and just rely on the mouse emulation of the X11 Wacom driver to get the actual events like position & button presses.
My main problem going with this approach when I tried it was: How do I know which `WindowEvent::AxisMotion` is the correct one for my Wacom tablet? There seems to be no way of actually verifying the source of the event (please enlighten me if there is one!), so I'd have to rely on sketchy mappings like "does the event have the correct number of axis".

maroider:
> If you really need tablet support, I THINK WindowEvent::AxisMotion supports tablets, albeit not in pixel coordinate format. But in tablet space coordinate space.

I honestly have no idea what `WindowEvent::AxisMotion` is intended for.

As for what's blocking this PR in particular (barring any discussion on the proposed API): We don't have an X11 maintianer right now, and no-one else in the @rust-windowing org has stepped up to review and merge this PR.

codecnotsupported:
> > If you really need tablet support, I THINK WindowEvent::AxisMotion supports tablets, albeit not in pixel coordinate format. But in tablet space coordinate space.
> 
> I honestly have no idea what `WindowEvent::AxisMotion` is intended for, and I'm surprised that you can get tablet input through it.
> 
> As for what's blocking this PR in particular (barring any discussion on the proposed API): We don't have an X11 maintainer right now, and no-one else in the @rust-windowing org has stepped up to review and merge this PR.


I think `WindowEvent::AxisMotion` is used for (_but not exclusive to_) controller-/joy- sticks.

> `WindowEvent::AxisMotion` might work. I can read the pressure data from there and just rely on the mouse emulation of the X11 Wacom driver to get the actual events like position & button presses. My main problem going with this approach when I tried it was: How do I know which `WindowEvent::AxisMotion` is the correct one for my Wacom tablet? There seems to be no way of actually verifying the source of the event (please enlighten me if there is one!), so I'd have to rely on sketchy mappings like "does the event have the correct number of axis".

Winit can differentiate between device id just fine.
How to to actually use DeviceId I find confusing as well.
You can't construct deviceId since it's platform dependent and platform_impl is private. 
It's also unclear to me how it traces back to the platform specific implementation to which may help in deducing what the device is from the DeviceId. But I digress.

mahkoh:
Fyi in https://github.com/maroider/winit/pull/6 I've added a platform extension that allows the XInput id to be retrieved from DeviceId.

PR #1883, opened 2021-03-14
feat: add interface for changing window's z-level

z4122:

On macOS, added `set_window_level` and `get_window_level` to set and get window's z-level.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
 - On macOS, initialize the Menu Bar with minimal defaults. (Can be prevented using `enable_default_menu_creation`)
 - On macOS, change the default behavior for first click when the window was unfocused. Now the window becomes focused and then emits a `MouseInput` event on a "first mouse click".
 - Implement mint (math interoperability standard types) conversions (under feature flag `mint`).
+- On macOS, added `set_level` and `get_level` to set and get window's z-level.
 
 # 0.24.0 (2020-12-09)
 
--- a/src/platform/macos.rs
+++ b/src/platform/macos.rs
@@ -6,7 +6,7 @@ use crate::{
     dpi::LogicalSize,
     event_loop::{EventLoop, EventLoopWindowTarget},
     monitor::MonitorHandle,
-    platform_impl::get_aux_state_mut,
+    platform_impl::{get_aux_state_mut, util},
     window::{Window, WindowBuilder},
 };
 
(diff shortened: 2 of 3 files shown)


src/platform/macos.rs
```diff
@@ -38,6 +38,12 @@ pub trait WindowExtMacOS {
 
     /// Sets whether or not the window has shadow.
     fn set_has_shadow(&self, has_shadow: bool);
+
+    /// Get window z-level
+    fn get_window_level(&self) -> i64;
+
+    /// Set window z-level
+    fn set_window_level(&self, level: i64);
```
ArturKovacs, line 46: These should return/take an `isize` instead of an `i64`, because an `NSInteger` is 32 bits wide when compiled to a 32 bit architecture.

Furthermore I think these should just be called `get_level` and `set_level`; I don't think that adding "window" provides any information here. I think the documentation could include a link to the appropriate apple developers link, like
https://developer.apple.com/documentation/appkit/nswindow/1419511-level?language=objc

ArturKovacs (changes requested):
It looks pretty good, except for those few things I noted.

ArturKovacs:
It just occurred to me that maybe there's some merit in exploring a cross platform API for this functionality.

For example, on Windows, there's: [`SetWindowPos`](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setwindowpos)
And on X11, there is a group of functions something similar (e.g. `XRaiseWindow`, `XLowerWindow`) see: https://tronche.com/gui/x/xlib/window/stacking-order.html

CHANGELOG.md
```diff
@@ -16,7 +16,7 @@
 - On Web, added support for `DeviceEvent::MouseMotion` to listen for relative mouse movements.
 - Added `Window::drag_window`. Implemented on Windows, macOS, X11 and Wayland.
 - On X11, bump `mio` to 0.7.
-
+- On macOS, added `set_window_level` and `get_window_level` to set and get window's z-level.
```
ArturKovacs, line 19: There should be an empty line after the last line here.

src/platform/macos.rs
```diff
@@ -70,6 +76,16 @@ impl WindowExtMacOS for Window {
...
+
+    #[inline]
+    fn get_window_level(&self) -> i64 {
+        self.window.get_window_level()
+    }
+
+    #[inline]
+    fn set_window_level(&self, level: i64) {
+        self.window.set_window_level(level)
+    }
```
ArturKovacs, line 88: This should instead use the `set_level_async` in `util/async.rs` to make this call thread-safe.

ArturKovacs (changes requested):

madsmtm:
`set_window_level` was added in https://github.com/rust-windowing/winit/pull/2534, querying is tracked in https://github.com/rust-windowing/winit/issues/2602.

Issue #1434, opened 2020-02-03
Allow users to click through the window


Ayplow:
For overlays that draw over other windows, it's important that the user is able to interact with the applications "below" the overlay. Some programs that make use of this are [f.lux](https://justgetflux.com/) and [Discord's Overlay](https://discordapp.com).

Somewhere to start with an implementation:
- [Windows' `WS_EX_TRANSPARENT` window style](https://www.codeproject.com/Articles/12365/How-to-make-a-transparent-and-click-through-window)
- [X11's Nonrectangular Window Shape Extension Protocol](https://www.x.org/releases/X11R7.7/doc/xextproto/shape.html)
- [NSWindow's `ignoresMouseEvents` flag](https://developer.apple.com/documentation/appkit/nswindow/1419354-ignoresmouseevents)

samhedin:
I would love this feature. On windows it, as you say, just seems to be a matter of setting `WS_EX_TRANSPARENT`. It's already used in some form here, with the use of `winuser::WS_EX_TRANSPARENT` https://docs.rs/winit/0.22.1/x86_64-pc-windows-msvc/src/winit/platform_impl/windows/event_loop.rs.html
I have no idea how to start working on this, but I would be interested in putting in some effort if someone would point me in the right direction.

samhedin:
I tried my hand at this and from what it looks to me, it shouldn't be worse than setting the bit (for windows) in ´windows/window_state.rs´. As a test, I added
`        style_ex |= WS_EX_COMPOSITED;
        style_ex |= WS_EX_TRANSPARENT;
`
to the bottom of the function `to_window_styles` as well as commenting out `| winuser::CS_OWNDC` in `window.rs` as per the note in https://docs.microsoft.com/en-us/windows/win32/winmsg/extended-window-styles

This did not work, neither did any combination of these settings work. I might be flailing around a bit randomly - but most resources I see (including the one linked in op) seem to say that this should be all it takes. I might be missing something basic here.

kchibisov:
not sure if it'll work on Wayland, but there's no standard protocol for overlays either now on it. Though, I wonder if winit should support overlays, since they usually use special set of protocols on each platform IIRC.

kazzix14:
I also would love this future. think [this](https://www.x.org/releases/X11R7.7/doc/compositeproto/compositeproto.txt) can be used?
> 3.2 Composite Overlay Window
Version 0.3 of the protocol adds the Composite Overlay Window,

[...]

PR #1884, opened 2021-03-16
feat: add set_ignore_cursor_events

z4122:

On macOS and Windows, added `set_ignore_cursor_events` to make window catch or ignore mouse events.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
 - On macOS, initialize the Menu Bar with minimal defaults. (Can be prevented using `enable_default_menu_creation`)
 - On macOS, change the default behavior for first click when the window was unfocused. Now the window becomes focused and then emits a `MouseInput` event on a "first mouse click".
 - Implement mint (math interoperability standard types) conversions (under feature flag `mint`).
+- On macOS and Windows, added `set_ignore_mouse_events` to let the window ignore mouse events.
 
 # 0.24.0 (2020-12-09)
 
--- a/src/platform_impl/macos/util/async.rs
+++ b/src/platform_impl/macos/util/async.rs
@@ -10,7 +10,7 @@ use cocoa::{
 };
 use dispatch::Queue;
 use objc::rc::autoreleasepool;
-use objc::runtime::NO;
+use objc::runtime::{NO, YES};
 
 use crate::{
     dpi::LogicalSize,
@@ -85,6 +85,14 @@ pub unsafe fn set_level_async(ns_window: id, level: ffi::NSWindowLevel) {
     let ns_window = MainThreadSafe(ns_window);
     Queue::main().exec_async(move || {
         ns_window.setLevel_(level as _);
+    });
+}
+
+// `setIgnoresMouseEvents_:` isn't thread-safe, and fails silently.
+pub unsafe fn set_ignore_mouse_events(ns_window: id, ignore: bool) {
+    let ns_window = MainThreadSafe(ns_window);
+    Queue::main().exec_async(move || {
+        ns_window.setIgnoresMouseEvents_(if ignore { YES } else { NO });
     });
 }
 
--- a/src/platform_impl/windows/window_state.rs
+++ b/src/platform_impl/windows/window_state.rs
@@ -86,6 +86,8 @@ bitflags! {
 
         const MINIMIZED = 1 << 12;
 
+        const IGNORE_CURSOR_EVENT = 1 << 14;
+
         const EXCLUSIVE_FULLSCREEN_OR_MASK = WindowFlags::ALWAYS_ON_TOP.bits;
         const NO_DECORATIONS_AND_MASK = !WindowFlags::RESIZABLE.bits;
         const INVISIBLE_AND_MASK = !WindowFlags::MAXIMIZED.bits;
@@ -222,6 +224,9 @@ impl WindowFlags {
         }
         if self.contains(WindowFlags::MAXIMIZED) {
             style |= WS_MAXIMIZE;
+        }
+        if self.contains(WindowFlags::IGNORE_CURSOR_EVENT) {
+            style_ex |= WS_EX_TRANSPARENT | WS_EX_LAYERED;
         }
 
         style |= WS_CLIPSIBLINGS | WS_CLIPCHILDREN | WS_SYSMENU;
--- a/src/window.rs
+++ b/src/window.rs
@@ -817,6 +817,20 @@ impl Window {
     pub fn drag_window(&self) -> Result<(), ExternalError> {
         self.window.drag_window()
     }
+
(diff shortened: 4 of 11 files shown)


src/window.rs
```diff
@@ -780,6 +780,20 @@ impl Window {
...
     }
+
+    /// Modifies whether the window catches cursor events.
+    ///
+    /// If `false`, this will catch the cursor events. If `true`, this will ignore the cursor.
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **:** The cursor is only hidden within the confines of the window.
+    /// - **:** The cursor is only hidden within the confines of the window.
```
maroider, line 791:
```suggestion
```
Is this something you meant to remove, or what?

examples/cursor_grab.rs
```diff
@@ -36,6 +36,7 @@ fn main() {
                         Escape => *control_flow = ControlFlow::Exit,
                         G => window.set_cursor_grab(!modifiers.shift()).unwrap(),
                         H => window.set_cursor_visible(modifiers.shift()),
+                        I => window.set_ignore_cursor_events(!modifiers.shift()).unwrap(),
```
maroider, line 39: I'm not sure if this quite fits in with the `cursor_grab` example, but that's mostly a naming thing for me.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 - On Web, added support for `DeviceEvent::MouseMotion` to listen for relative mouse movements.
 - Added `Window::drag_window`. Implemented on Windows, macOS, X11 and Wayland.
 - On X11, bump `mio` to 0.7.
+- On macOS and Windows, added `set_ignore_cursor_events` to make window catch or ignore mouse events.
```
maroider, line 19:
```suggestion
- On macOS and Windows, added `set_ignore_cursor_events` to let the window ignore mouse events.
```

src/platform_impl/windows/window_state.rs
```diff
@@ -219,6 +221,9 @@ impl WindowFlags {
         if self.contains(WindowFlags::MAXIMIZED) {
             style |= WS_MAXIMIZE;
         }
+        if self.contains(WindowFlags::IGNORE_CURSOR_EVENT) {
+            style_ex |= WS_EX_TRANSPARENT | WS_EX_LAYERED;
```
maroider, line 225: In #1815, we decided to remove the `WS_EX_LAYERED` style as it is documented as being incompatible with the CS_OWNDC window class style.
When I test stuff (not this PR, mind you) on my own machine it seems to work fine, but the documentation nevertheless says it is incompatible.

maroider (commented):
I'd like to test this later, but here's some feedback anyway.

cc @ArturKovacs, @msiglreith

ArturKovacs:
Please elaborate on what's the benefit of adding this. To me it seems that this is identical to simply not processing mouse events coming from the event loop.

z4122:
> Please elaborate on what's the benefit of adding this. To me it seems that this is identical to simply not processing mouse events coming from the event loop.

We can use this feature to create a transparent floating window uppon some other windows. Then we can draw something or just show something on this window at the same time don't effect user's normal operations. It is useful in situation like screen sharing.

maroider:
A concrete example of a program utilizing something like this is [ScreenToGif](https://www.screentogif.com/how-to-use). I couldn't see if they're using this exact technique by quickly searching through their repository.

There's actually an issue open for this already: #1434.

As noted in #1434, it isn't clear if Winit should explicitly support "overlays", but I think allowing click-through windows should be fairly inoffensive. An X11 implementation in Winit should probably be possible, although I wouldn't block this PR on it.

src/platform_impl/windows/window_state.rs
msiglreith: Good point.. maybe we can get rid of the `CS_OWNDC` part as it's mainly/only for OpenGL and I believe we can get around it. Will play around with this one a bit in the next days.

--------------

Would be nice to have reference link added here regarding hit-test behavior (https://docs.microsoft.com/de-de/windows/win32/winmsg/window-features?redirectedfrom=MSDN#layered-windows)

src/window.rs
```diff
@@ -780,6 +780,18 @@ impl Window {
...
+
+    /// Modifies whether the window catches cursor events.
+    ///
+    /// If `false`, this will catch the cursor events. If `true`, this will ignore the cursor.
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **iOS / Android / Web / X11 / Wayland:** Unsupported.
+    #[inline]
+    pub fn set_ignore_cursor_events(&self, ignore: bool) -> Result<(), ExternalError> {
```
msiglreith, line 792: A bit of bikeshedding but I would prefer some affirmative variant of the method name as `set_ignore_cursor_events(false)` might be hard to read. My suggestion would be `set_cursor_hittest`.

src/platform_impl/macos/util/async.rs
```diff
@@ -88,6 +88,13 @@ pub unsafe fn set_level_async(ns_window: id, level: ffi::NSWindowLevel) {
     });
 }
 
+pub unsafe fn set_window_ignore_cursor_events(ns_window: id, ignore: bool) {
```
ArturKovacs, line 91: I believe this should be called `set_ignore_cursor_events_async`. The "window" is removed because it doesn't add any information and the "async" is added telling the caller that this change will not have taken effect by the time the function returns.

ArturKovacs (changes requested):

coderedart:
Just wanted to mention that Qt uses the Qt::WindowTransparentForInput as a widget attribute and I believe it is a much more intuitive name.  Or something like input passthrough.

ArturKovacs:
We are starting to go deep into bikeshedding territory; it's a scary place where I don't want to be. @z4122 please rename `set_ignore_mouse_events` to `set_cursor_hittest`, and also rename `WindowFlags::IGNORE_CURSOR_EVENT` to `CURSOR_HITTEST`, and everything else that's related and I missed.

(WindowTransparentForInput in Qt seems to ignore both mouse and keyboard events but this PR only ignores mouse events)

coderedart:
Qt::WA_TransparentForMouseEvents exists too.

maroider:
Coming back to this PR after a couple of months is a bit disorienting, but I believe we're currently blocked on:
- [ ] Deciding on a name for the `Window` method
- [ ] An entry in the feature matrix
- [ ] #1893/#1910

chengyuhui:
Is there some way I could help to get this feature merged? I am targeting Windows in particular, although a cross-platform solution is also desired (don't know if we'll ever able to implement this on Linux).

kchibisov:
Closing in favor of https://github.com/rust-windowing/winit/pull/2232.

PR #1887, opened 2021-03-17
feat: catch mouse moved when window blured

z4122:

On macOS, replace tracking_rect with tracking_area to catch mouse moved when window blured.

In macOS 10.5 and later, [tracking_area](https://developer.apple.com/documentation/appkit/nsview/1483333-trackingareas) provide a greater range of functionality than [tracking_rect](https://developer.apple.com/documentation/appkit/nsview/1483668-addtrackingrect)


ArturKovacs:
I'm going to risk pointing out the obvious, but this does not compile.

CHANGELOG.md
```diff
@@ -16,7 +16,7 @@
 - On Web, added support for `DeviceEvent::MouseMotion` to listen for relative mouse movements.
 - Added `Window::drag_window`. Implemented on Windows, macOS, X11 and Wayland.
 - On X11, bump `mio` to 0.7.
-
+- On macOS, used tracking_area to replace tracking_rect to catch mouse moved when window blured.
```
ArturKovacs, line 19: There should be a line break after the last line of the list here.

src/platform_impl/macos/view.rs
```diff
@@ -325,19 +329,22 @@ extern "C" fn frame_did_change(this: &Object, _sel: Sel, _event: id) {
         let state_ptr: *mut c_void = *this.get_ivar("winitState");
         let state = &mut *(state_ptr as *mut ViewState);
 
-        if let Some(tracking_rect) = state.tracking_rect.take() {
-            let _: () = msg_send![this, removeTrackingRect: tracking_rect];
+        if let Some(tracking_area) = state.tracking_area.take() {
+            let _: () = msg_send![this, removeTrackingArea: tracking_area];
```
ArturKovacs, line 333: Since you are allocation a new tracking area later, the old tracking area here should be freed by sending a [`release`](https://developer.apple.com/documentation/objectivec/1418956-nsobject/1571957-release) message to it.

ArturKovacs (changes requested):
It's pretty nice, there are just two small things that need to be addressed.

madsmtm:
Closing in favour of https://github.com/rust-windowing/winit/pull/4514.

PR #1888, opened 2021-03-18
New keyboard API for the web

maroider:

This implements #753 for WASM.

## TODOs

- [x] Decide if a `stdweb` backend is worthwhile, since it's deprecated.
    - [x] ~~Implement a `stdweb` backend if need be.~~
- [ ] A bunch of `TODO` comments in the code.
- [ ] Debug why my [example](https://gist.github.com/maroider/6c8bcff3642a07d111ed85bf077a4c8a/fb8176a26081567f5a29c427a734caf46aff4b42) program sometimes freezes with certain input sequences.


src/keyboard.rs
```diff
@@ -659,6 +664,409 @@ pub enum KeyCode {
     F35,
 }
 
+impl KeyCode {
+    pub fn from_key_code_attribute_value(kcav: &str) -> Self {
```
ArturKovacs, line 668: Considering that this is web-specific functionality, I think this `impl` along with the other web-specific conversions added to this file should be somewhere under the `platform/web` (or `platform_impl/web`) module.

src/keyboard.rs
```diff
@@ -1393,6 +1801,635 @@ pub enum Key<'a> {
...
+            "NumLock" => Key::NumLock,
+            "ScrollLock" => Key::ScrollLock,
+            "Shift" => Key::Shift,
+            "Symbol" => Key::Symbol,
+            "SymbolLock" => Key::SymbolLock,
+            "Hyper" => Key::Hyper,
+            "Meta" => Key::Super,
+            "Enter" => Key::Enter,
+            "Tab" => Key::Tab,
+            "Space" => Key::Space,
```
ArturKovacs, line 1825: Oops! Space is encoded as `" "` on the web. See https://www.w3.org/TR/uievents-key/#keys-whitespace

ArturKovacs (changes requested):

src/keyboard.rs
maroider: I guess that's what happens when I do bulk edits.

ArturKovacs:
For the record: I'll test and properly review this once the merge conflicts are resolved.

src/keyboard.rs
maroider: Done

maroider:
Welp, I messed up the merge more times than I'd like to admit, but the merge conflicts should be resolved now.

src/platform_impl/web/web_sys/event.rs
```diff
@@ -61,188 +62,59 @@ pub fn mouse_scroll_delta(event: &WheelEvent) -> Option<MouseScrollDelta> {
...
+        2 => KeyLocation::Right,
+        3 => KeyLocation::Numpad,
+        // TODO: Is this reasonable to do?
+        _ => KeyLocation::Standard,
+    }
+}
+
+pub fn key_repeat(event: &KeyboardEvent) -> bool {
+    event.repeat()
 }
```
ArturKovacs, line 102: Nitpick: it seems that this function just adds unnecessary code.

src/platform_impl/web/web_sys/event.rs
```diff
@@ -61,188 +62,59 @@ pub fn mouse_scroll_delta(event: &WheelEvent) -> Option<MouseScrollDelta> {
...
+
+// TODO: What should be done about `KeyboardEvent.isComposing`?
```
ArturKovacs, line 104: In that case, the keyboard event should probably be ignored for this PR. And there should be a `ReceivedImeText` emitted for [`compositionend`](https://developer.mozilla.org/en-US/docs/Web/API/Element/compositionend_event).

ArturKovacs (changes requested):
Coming along nicely

ArturKovacs:
Furthermore, <kbd>Space</kbd>, <kbd>Tab</kbd>, and <kbd>Enter</kbd> emit `None` for the text field; they should emit a string. (I believe Enter should be `"\r"`)

src/keyboard.rs
```diff
@@ -176,10 +176,11 @@ pub enum NativeKeyCode {
     Windows(u16),
     MacOS(u32),
     XKB(u32),
+    Web(&'static str),
```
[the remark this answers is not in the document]
maroider: I think this can also be fixed by using the [`#[serde(deserialize_with = "func")]`](https://serde.rs/field-attrs.html#deserialize_with) attribute, but a `String` would also be fine, as this should only happen very rarely, and I don't expect that people would want to match on this value under normal circumstances.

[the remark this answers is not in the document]
maroider: Actually, using `String` would make `NativeKeyCode` no longer `Copy`.

src/platform_impl/web/keyboard.rs
```diff
@@ -0,0 +1,521 @@
...
+            "F28" => KeyCode::F28,
+            "F29" => KeyCode::F29,
+            "F30" => KeyCode::F30,
+            "F31" => KeyCode::F31,
+            "F32" => KeyCode::F32,
+            "F33" => KeyCode::F33,
+            "F34" => KeyCode::F34,
+            "F35" => KeyCode::F35,
+            // TODO: Fix unbounded leak
+            string @ _ => KeyCode::Unidentified(NativeKeyCode::Web(Box::leak(
```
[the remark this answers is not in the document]
maroider: Yep, it is.

src/platform_impl/web/event_loop/window_target.rs
```diff
@@ -68,48 +71,55 @@ impl<T> WindowTarget<T> {
         });
 
         let runner = self.runner.clone();
-        canvas.on_keyboard_press(move |scancode, virtual_keycode, modifiers| {
+        canvas.on_keyboard_press(move |physical_key, logical_key, text, location, repeat| {
```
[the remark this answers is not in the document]
maroider: > I'm not sure if we should drop support for stdweb just yet although

To this end, I've opened up #1941.

src/keyboard.rs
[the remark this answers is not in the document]
maroider: I've just removed the contents of the `Web` variant for now.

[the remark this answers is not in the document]
ArturKovacs: After thinking about it a bit, I still think that it should be a `String`. Loosing the `Copy` trait is a bit unfortunate but it still seems to be the most reasonable thing to do, especially because as you said, this type should be rarely used.

src/platform_impl/web/web_sys/canvas.rs
```diff
@@ -170,32 +173,34 @@ impl Canvas {
                     event.prevent_default();
```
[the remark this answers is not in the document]
maroider: I completely agree.

src/platform_impl/web/web_sys/event.rs
maroider: I recon it's a leftover from some previous state.

src/platform_impl/web/event_loop/window_target.rs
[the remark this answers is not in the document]
maroider: We'll just ignore this for now, since #1941 has been merged.

src/platform_impl/web/web_sys/event.rs
```diff
@@ -61,188 +62,59 @@ pub fn mouse_scroll_delta(event: &WheelEvent) -> Option<MouseScrollDelta> {
...
-    m.set(ModifiersState::LOGO, event.meta_key());
-    m
+pub fn key_text(event: &KeyboardEvent) -> Option<&'static str> {
+    let key = event.key();
+    if let Key::Character(text) = Key::from_key_attribute_value(&key) {
+        // TODO: Fix unbounded leak
+        Some(Box::leak(String::from(text).into_boxed_str()))
+    } else {
+        None
+    }
```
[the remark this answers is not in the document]
maroider: I copied over what I used in the Linux PR, which is about what you described. I'm a bit uncertain of how mutexes work on WASM, but it seemed to run fine when I tested it.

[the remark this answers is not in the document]
maroider: @ryanisaacg, how do mutexes work on WASM?

maroider:
> Furthermore, Space, Tab, and Enter emit `None` for the text field; they should emit a string. (I believe Enter should be `"\r"`)

This should be fixed now.

What remains now is [a single TODO comment](https://github.com/rust-windowing/winit/blob/c75b53653905696e8db27098b5ae24d652599d61/src/platform_impl/web/web_sys/event.rs#L122) as well as debugging the strange freezing.

src/platform_impl/web/web_sys/event.rs
```diff
@@ -69,23 +69,41 @@ pub fn key_code(event: &KeyboardEvent) -> KeyCode {
...
-        }
-        Key::Tab => Some(Box::leak(String::from("\t").into_boxed_str())),
-        Key::Enter => Some(Box::leak(String::from("\r").into_boxed_str())),
-        Key::Space => Some(Box::leak(String::from(" ").into_boxed_str())),
+    let key = Key::from_key_attribute_value(&key);
+    match &key {
+        Key::Character(text) => Some(*text),
+        Key::Tab => Some("\t"),
+        Key::Enter => Some("\r"),
+        Key::Space => Some(" "),
```
ArturKovacs, line 83: Note that the `Key` has the `to_text` method which does pretty much this.

Woyten:
@maroider Is there any update on this topic? I would like to see #2092 fixed. I could do the fixing myself but I am not sure whether my fix would be accepted and/or is in conflict with this PR.

src/platform_impl/web/web_sys/event.rs
```diff
@@ -61,188 +62,78 @@ pub fn mouse_scroll_delta(event: &WheelEvent) -> Option<MouseScrollDelta> {
...
-        "WebStop" => VirtualKeyCode::WebStop,
-        "Yen" => VirtualKeyCode::Yen,
-        _ => return None,
-    })
+fn cached_string<S: AsRef<str>>(string: S) -> &'static str {
+    use std::collections::HashSet;
+    use std::sync::Mutex;
+
+    lazy_static::lazy_static! {
+        static ref STRING_CACHE: Mutex<HashSet<&'static str>> = Mutex::new(HashSet::new());
```
Liamolucko, line 94: The main thread isn't allowed to block on the web, so a `Mutex` probably shouldn't be used here. A `thread_local!` should work.

src/platform_impl/web/web_sys/event.rs
```diff
@@ -61,188 +62,78 @@ pub fn mouse_scroll_delta(event: &WheelEvent) -> Option<MouseScrollDelta> {
...
+        _ => ModifiersState::empty(),
+    }
 }
+
+// pub fn codepoint(event: &KeyboardEvent) -> char {
+//     // `event.key()` always returns a non-empty `String`. Therefore, this should
+//     // never panic.
+//     // https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key
+//     event.key().chars().next().unwrap()
+// }
```
Liamolucko, line 139: This should probably be removed.

src/platform_impl/web/web_sys/event.rs
Liamolucko: To be fair, I don't think the event loop can be run outside of the main thread so this probably won't be a problem in practice, but that's even more of a reason to use a thread local.

erwanvivien:
Is there any update on this and how can we contribute if we need to ?

Issue #725, opened 2018-11-22
Can't resize window without decorations


CryZe:
When disabling decorations, the window can't be resized anymore, even with `.with_resizable(true)` specified. Maximizing and other operations don't seem to work either.

francesca64:
Thanks for reporting this! I've successfully reproduced it.

This repo looks like a good exploration of how to fix this: https://github.com/melak47/BorderlessWindow

@Osspial, do you want to tackle this, or should I?

Osspial:
It's not immediately clear what the best solution is for fixing this. The reason borderless windows can't be resized on Windows is that Windows, by default, shows resize handles on the window borders. The way melak47/BorderlessWindow works is that it reworks the resize detection code so that the resize handles appear inside the window's client area, and I'm not comfortable with automatically taking over an amount the edges of Winit windows to enable resize handles.

In the past, people have brought up having some sort of API for creating resize borders in the client area of borderless windows. @francesca64 What does the situation look like for resizing borderless windows on other platforms? If most/all of them have this behavior I think we should look into what API changes we'd need to support this.

If most other platforms *do* let you resize borderless windows, we could set it up so that you can drag on the window shadow to handle resizing. This is what Windows 10 does, but it's inconsistent with Windows 7 and may not work on Windows 7 machines with Aero disabled (since windows with that configuration don't have a window shadow).

FrankenApps:
I think the best solution would be to implement this the same way it is done in Electron.

i509VCB:
I've tested this with X11 and Wayland in KDE and a window there is not resizable without decorations even when explicitly specifying `with_resizable(true)`.

Cannot test in GNOME or other DEs sadly.

garasubo:
I've tested this with X11 in GNOME Ubuntu, and it is not resizable either when `with_decorations(false)`

i509VCB:
I noticed this option in KDE, specifically in plasma settings, however enabling the option to add a handle to resize windows with no border does change behavior from winit X11 or wayland.

![Screenshot_20210514_165953](https://user-images.githubusercontent.com/30619168/118335679-b6b66900-b4ff-11eb-821a-bd531c7f805d.png)


Uhh who would we ask here to change the labels on the issue so this is recognized as multi-platform?

maroider:
> Uhh who would we ask here to change the labels on the issue so this is recognized as multi-platform?

:)

---

[...]

PR #1891, opened 2021-03-21
fix(windows): fix aero-snap of borderless window

### LINKED ISSUE #1548: No window snapping without decorations on Windows
It seems like window snapping does not work on Windows when decorations are disabled. This was tested using the Windows shortcuts, since the decorations aren't available to drag them of course.

This was originally reported to Alacritty in https://github.com/alacritty/alacritty/issues/3602.

amrbashir:

Fixes #1548.

**Note**: This PR is merely a port of the cpp code in this [repo](https://github.com/melak47/BorderlessWindow/)

**Note**: The window isn't borderless when first created and needs to be maximized at least once or resized or anything that redraws the window before it can work properly. This however not a bug in my implementation, it is a bug in the subclass handler, you can reproduce it by adding `println!("ANYTHING")` inside `WM_NCCREATE`. It will never be called which means there is a gap before the subclass handler gets attached.


src/platform_impl/windows/event_loop.rs
```diff
@@ -1961,6 +1961,106 @@ unsafe fn public_window_callback_inner<T: 'static>(
...
+                        params.rgrc[0] = monitor_info.rcWork;
+                    }
+                }
+                0 // must return a value here, other wise the window won't be borderless (without decorations)
+            } else {
+                commctrl::DefSubclassProc(window, msg, wparam, lparam)
+            }
+        }
+
+        winuser::WM_NCHITTEST => {
```
maroider, line 1982: I'm not opposed to enabling resizing on borderless windows in principle, but there should be some thought as to how this is exposed to users of Winit, as well as how this might be implemented on other platforms. It's actually not clear to me whether the resizing area is on the window's "shadow", or if it intrudes on the client area.

See #725 and #221 for more info.

src/platform_impl/windows/window_state.rs
```diff
@@ -197,6 +197,8 @@ impl WindowFlags {
         if self.contains(WindowFlags::DECORATIONS) {
             style |= WS_CAPTION | WS_MINIMIZEBOX | WS_BORDER;
             style_ex = WS_EX_WINDOWEDGE;
+        } else {
+            style |= WS_CAPTION | WS_MAXIMIZEBOX | WS_MINIMIZEBOX | WS_SIZEBOX;
```
maroider, line 201: Do you absolutely need `WS_MINIMIZEBOX` here?

maroider (commented):
I haven't had the chance to test this, as I'm currently in Linux-land, implementing #753.

As discussed [on Discord](https://discord.com/channels/273534239310479360/583054410670669833/822947960857034783) yesterday, some messages can be delivered to the window procedure during the call to `CreateWindow`, and `WM_NCCREATE` is presumably among them. I'm going to look into fixing this, but that's going to have to sit on the back-burner while I work on #753.

examples/drag_window.rs
```diff
@@ -11,7 +11,10 @@ fn main() {
     SimpleLogger::new().init().unwrap();
     let event_loop = EventLoop::new();
 
-    let window_1 = WindowBuilder::new().build(&event_loop).unwrap();
+    let window_1 = WindowBuilder::new()
+        .with_decorations(false)
+        .build(&event_loop)
+        .unwrap();
```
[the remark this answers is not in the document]
amrbashir: Sure

src/platform_impl/windows/window_state.rs
amrbashir: Yeah it is needed, Otherwise the window can't be minimized using <kbd>Win</kbd> + <kbd>Arrow Down</kbd>

amrbashir:
It is ok for this PR to wait until the subclass handler issue is fixed, In fact you can mark the PR as draft until then.

examples/drag_window.rs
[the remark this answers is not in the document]
maroider: Actually, I think it may be better to use the `window_debug` example for this. I can't quite believe I forgot about it.

src/platform_impl/windows/event_loop.rs
amrbashir: It intrudes in the first few pixels in the client area. It is calculated by the cursor position relative to the window edge + the border size. we can even replace the border size with custom value or remove it.

I see some valid concerns about where the resize handler should appear.

And I don't know about other platforms tbh but I think macos does let you resize broderless windows.

amrbashir:
I found a workaround, we can resize the window with one pixel then revert it back after attaching the subclass.

After this line:
https://github.com/rust-windowing/winit/blob/e8cdf8b092ed3ed2c1dae95606664a30168d4028/src/platform_impl/windows/window.rs#L122

we can do something like this
```rust
event_loop::subclass_window(win.window.0, subclass_input);

let win_flags = win.window_state.lock().window_flags();
if !win_flags.contains(WindowFlags::DECORATIONS) {
	let original = win.inner_size();
	win.set_inner_size(
		PhysicalSize::new(original.width + 1, original.height).into(),
	);
    win.set_inner_size(original.into());
}
```

src/platform_impl/windows/event_loop.rs
maroider: I really would prefer if custom resize areas were done in a separate PR, so that this one can be focused only on smoothing over a Windows quirk.

maroider:
I don't particularly like the hack you've proposed, but @msiglreith is the Windows maintainer in the end.

I'd prefer to move from using a subclass to using `public_window_callback` as a window procedure, but I'm also not entirely comfortable with that given that the original reason for using a subclass doesn't seem to be documented anywhere.

Unfortunately, I can't give you a precise estimate of when I'll get on with fixing this, but I hope to look into this during the next 2-4 weeks.

src/platform_impl/windows/event_loop.rs
amrbashir: Yeah, That makes sense, I can make this PR focus only on Fixing the aero-snap but it would still require either the workaround I purposed below or a fix for the subclass issue.

examples/window_debug.rs
```diff
@@ -26,9 +26,11 @@ fn main() {
     eprintln!("  (Q) Quit event loop");
     eprintln!("  (V) Toggle visibility");
     eprintln!("  (X) Toggle maximized");
+    eprintln!("  (B) Toggle Borderless");
```
filnet, line 29: Borderless should start with a lowercase b to be consistent with other entries.

src/platform_impl/windows/event_loop.rs
```diff
@@ -1961,6 +1961,24 @@ unsafe fn public_window_callback_inner<T: 'static>(
             commctrl::DefSubclassProc(window, msg, wparam, lparam)
         }
 
+        winuser::WM_NCCALCSIZE => {
+            let win_flags = subclass_input.window_state.lock().window_flags();
+
+            if !win_flags.contains(WindowFlags::DECORATIONS) {
+                // adjust the maximized borderless window fill the work area rectangle of the display monitor
```
filnet, line 1968: "to fill" ?

src/platform_impl/windows/event_loop.rs
```diff
@@ -1961,6 +1961,24 @@ unsafe fn public_window_callback_inner<T: 'static>(
...
+                if util::is_maximized(window) {
+                    let monitor = monitor::current_monitor(window);
+                    if let Ok(monitor_info) = monitor::get_monitor_info(monitor.hmonitor()) {
+                        let params = &mut *(lparam as *mut winuser::NCCALCSIZE_PARAMS);
+                        params.rgrc[0] = monitor_info.rcWork;
+                    }
+                }
+                0 // must return a value here, other wise the window won't be borderless (without decorations)
```
filnet, line 1976: "otherwise" not "other wise"

src/platform_impl/windows/util.rs
```diff
@@ -196,6 +196,15 @@ pub fn is_focused(window: HWND) -> bool {
     window == unsafe { winuser::GetActiveWindow() }
 }
 
+pub fn is_maximized(window: HWND) -> bool {
+    unsafe {
+        let mut placement: winuser::WINDOWPLACEMENT = mem::zeroed();
+        placement.length = mem::size_of::<winuser::WINDOWPLACEMENT>() as u32;
+        winuser::GetWindowPlacement(window, &mut placement);
+        return placement.showCmd == winuser::SW_MAXIMIZE as u32;
```
filnet, line 204: no need for "return" ?

amrbashir: not needed indeed, that was just my poor rust skills

amrbashir:
This workaround isn't enough too because the transparency seems to be broken too because of the subclass issue. 

To confirm that transparency works well with a normal window procedure I cloned this repo https://github.com/melak47/BorderlessWindow/ and added the same transparency code that winit uses, and it works as expected along with the aero-snap and resizing

amrbashir:
I will mark this PR as ready for review, since #1933 got merged and fixed the issues I had.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - Added `Window::focus_window`to bring the window to the front and set input focus.
 - On Wayland and X11, implement `is_maximized` method on `Window`.
 - On macOS, fix issue where `ReceivedCharacter` was not being emitted during some key repeat events.
+- On Windows, fix aero-snap for borderless(undecorated) window.
```
maroider, line 9:
```suggestion
- On Windows, fix aero-snap for borderless (undecorated) windows.
```

src/platform_impl/windows/window_state.rs
```diff
@@ -192,13 +189,17 @@ impl WindowFlags {
 
         let (mut style, mut style_ex) = (0, 0);
 
+        style |= WS_CLIPSIBLINGS | WS_CLIPCHILDREN | WS_SYSMENU | WS_CAPTION | WS_MINIMIZEBOX;
+        style_ex |= WS_EX_ACCEPTFILES;
```
maroider, line 193: Nit: Is there a reason that you've moved this up here? It made the changes *slightly* harder to follow.

maroider (commented):
This seems mostly "good to go" except for one thing:
If you run the `window` example with `.with_decorations(false)` and then use <kbd>Win</kbd> + "the left arrow key" such that the window snaps to the border between two monitors, then the window will spill onto the adjacent monitor. I haven't been able to reproduce this bug with other monitor configurations.

src/platform_impl/windows/window_state.rs
amrbashir: Since these styles are always added, it felt more intuitive to me to have them before all the checks. I don't have strong opinion about it, I can revert it if you want.

amrbashir:
> This seems mostly "good to go" except for one thing:
> If you run the `window` example with `.with_decorations(false)` and then use Win + "the left arrow key" such that the window snaps to the border between two monitors, then the window will spill onto the adjacent monitor. I haven't been able to reproduce this bug with other monitor configurations. melak47/BorderlessWindow does not exhibit this behaviour.

Unfortunately, I don't have other monitors to test this behavior.

src/platform_impl/windows/window_state.rs
maroider: When you put it like that, it does sound fairly reasonable.

src/platform_impl/windows/window_state.rs
```diff
@@ -192,13 +189,17 @@ impl WindowFlags {
...
         if self.contains(WindowFlags::RESIZABLE) {
             style |= WS_SIZEBOX | WS_MAXIMIZEBOX;
         }
         if self.contains(WindowFlags::DECORATIONS) {
-            style |= WS_CAPTION | WS_MINIMIZEBOX | WS_BORDER;
-            style_ex = WS_EX_WINDOWEDGE;
+            style |= WS_BORDER;
+            style_ex |= WS_EX_WINDOWEDGE;
         }
+
```
maroider, line 202: Nit: What's the deal with this blank line?

maroider:
> Unfortunately, I don't have other monitors to test this behavior.

Welp, I guess I'll dig through stuff and find the difference.

src/platform_impl/windows/window_state.rs
amrbashir: that was by mistake, will remove it.

maroider:
I have managed to reproduce the bug on melak47/BorderlessWindow now. It doesn't work on the initial aero-snap, but if I maximize the window and then use aero-snap shortcuts, then it'll spill over like how winit's window does. This also lets me confirm that it's a problem in both the left and right directions.

I also did a quick test of this "maximized to left or right aero-snap" scenario in a couple of apps, and only a couple of them seem to handle it correctly.
- Windows 10 Calculator, Notepad, File Explorer, Edge: Vertical size is off
- Chrome, Krita, Cheat Engine, Git GUI, LibreOffice Calc, Zenmap, VLC Media Player, WinDirStat: Vertical size is off
- VSCode: Same as winit
- Firefox, Thunderbird: Same as winit, but the "spill" is somehow smaller
- Audacity, WizTree: Get it right

Audacity, Cheat Engine, Zenmap, and WinDirStat seem to use the "classic" win32 GUI framework, so perhaps there's something to be learned by peeking at Audacity's win32 implementation.

amrbashir:
@maroider thanks for your effort and sorry I can't help much with that bug but currently there is a bigger bug, transparent isn't working with this PR as I mention [above](https://github.com/rust-windowing/winit/pull/1891#issuecomment-829156827)

maroider:
So I've poked at this a bit in the week that's has passed, and I can't quite seem to figure out how to do it correctly.

At its core, the "bad window size" issue seems to stem from how we make the client area fill the non-client area of the window, which also includes the (7px thick) window shadow (+ the 1px thick border). I've had some minor success trying to rectify this, but I haven't landed on anything that can be considered consistently not buggy.

Further complicating matters, all attempts to adjust the size of a maximized window in `WM_NCCALCSIZE` seem to make the window *no longer maximized* in the eyes of Windows. The window may be covering the entire monitor, but there's some internal window state that gets thrown out of whack when we adjust the size as the PR already does, as well as in my own attempts to fix it.

I tried peeking at what Audacity does, but I couldn't get the Audacity to compile, so I was unable to really *test* what I thought some piece of code was doing. Audacity uses wxWidgets, which I did manage to get to compile, although the provided sample programs exhibit the typical "vertical size is off" behaviour that I've now come to expect in most Windows applications. This suggests that it's *not* some special handling of a particular window message that does the trick, but rather some "after-the-fact" readjustment. This is in line with the behaviour I've observed, where Audacity seemingly has a "correction" step after an aero-snap.

Today, I was relaxing by trying out the demo for Ex-Zodiac, and I discovered some strange behaviour when fiddling with a window size setting. The game somehow managed to convince Windows that the window was maximized even *after* resizing the window to a smaller size. The window decorations kept the "restore" (or whatever it's called) button, even when the window *visibly* wasn't maximized. The game uses the Godot game engine, so I suspect that Godot's source code may contain a hint at a potential solution.

As for the transparency issue you've mentioned, @amrbashir, that doesn't *really* annoy me quite as much as the aero-snap size bug(s), so I probably won't be spending much—if any—energy on looking into a solution until *after* the aero-snap size bug(s) is(/are) fixed. I suggest that you check what glutin's `transparent` example looks like in combination with this PR, instead of just "adding a button" (which I presume is taken from the traditional win32 GUI framework).

amrbashir:
> As for the transparency issue you've mentioned, @amrbashir, that doesn't _really_ annoy me quite as much as the aero-snap size bug(s), so I probably won't be spending much—if any—energy on looking into a solution until _after_ the aero-snap size bug(s) is(/are) fixed. I suggest that you check what glutin's `transparent` example looks like in combination with this PR, instead of just "adding a button" (which I presume is taken from the traditional win32 GUI framework).

I just tried glutin with this PR, I can see a white flash for a split second until glutin draws a triangle on the window, but with winit v0.25, I don't see this white flash.

maroider:
How does it behave when you point glutin at winit's `master`?

amrbashir:
Interesting, so with 63ad47a7bf2d9e9b853c3eeb1cc71d61887ffe31 commit, the window is created but I can't see it nor anything glutin draws to it. 
It works fine with 27e654834327ea775b9b75423bc3c5f318ddffd5

maroider:
Well, that's concerning 😓. I guess I'll have to have a look at that ASAP.

maroider:
Welp, I've managed to reproduce the issues you've encountered, with a few caveats:
- The "initial flash of white" is a problem with regular windows, even without #1933
- Regular windows experience no problems other than the above

I didn't test stuff exhaustively, though, as I'm dead tired right now.
In any case, I've opened #1980 to track this regression.

maroider:
I believe I've found a workable solution to the "window spill-over "by using parts of [this SO answer](https://stackoverflow.com/a/39735058/14806876) (thanks, @MrGibus, for linking me this). It doesn't *quite* get rid of all of it, but I think this is about as correct as other applications seem to get it.

The slight vertical size error still persists, but I believe this is an error in some internal win32 function. The error goes away when I use a stub handler for `WM_NCPAINT` that always returns `0` and does nothing else.

I did try to see if I could link to an internal function I found (`ThemeDefWindowProcW` from `UxTheme.dll`) by stepping through what `DefWindowProcW` called, but I didn't really try further once linking failed (maybe I could hook the function somehow). In any case, using internal APIs like that is probably not what we want to do, so even if it did work immediately, I wouldn't really want to push that to `master`.

It might be possible to make borderless applications correct in this regard by using this `WM_NCPAINT` technique, but it's not entirely clear to me if that would work, and I haven't investigated it further.

There's also apparently a bug with the `window_debug` example where I have to press <kbd>B</kbd> twice in order to initially make the window borderless.

Bonus fact: If you use a `WM_NCPAINT` handler like the following, you'll get yourself some "Vista/Windows 7"-style window decorations after toggling decorations off and on again.

```rust
winuser::WM_NCPAINT => {
    if userdata
        .window_state
        .lock()
        .window_flags()
        .contains(WindowFlags::DECORATIONS)
    {
        winuser::DefWindowProcW(window, msg, wparam, lparam)
    } else {
        0
    }
}
```

amrbashir:
I guess there is nothing but BUGS in this PR and I am tempted to close it tbh.
```rs
// bug: each call adds an extra 16px to width and around 40px to height
// expected: nothing changes as we are just sitting the window size to the same current size
window.set_inner_size(window.inner_size());
window.set_inner_size(window.inner_size());
window.set_inner_size(window.inner_size());
```
It doesn't happen on master btw.

maroider:
> I guess there is nothing but BUGS in this PR

I guess that's what we get for abusing Windows in this way 😅. However, I *think* we can turn this PR into something usable, it just might take some more months before we get there (largely due to how sporadically I work on things).

amrbashir:
> I guess there is nothing but BUGS in this PR and I am tempted to close it tbh.
> 
> ```rust
> // bug: each call adds an extra 16px to width and around 40px to height
> // expected: nothing changes as we are just sitting the window size to the same current size
> window.set_inner_size(window.inner_size());
> window.set_inner_size(window.inner_size());
> window.set_inner_size(window.inner_size());
> ```
> 
> It doesn't happen on master btw and it happens in this PR even before @maroider spill-over fix.

I think I found the solution to this bug from this [SO answer](https://stackoverflow.com/a/21614755)

```rs
let dpi = hwnd_dpi(window);
let titlebar_height = winuser::GetSystemMetricsForDpi(winuser::SM_CYSIZE, dpi)
	+ winuser::GetSystemMetricsForDpi(winuser::SM_CXPADDEDBORDER, dpi) * 2
	+ winuser::GetSystemMetricsForDpi(winuser::SM_CYBORDER, dpi);

let rect = adjust_window_rect(
	window,
    RECT {
		top: 0,
        left: 0,
        bottom: (y) as LONG,
        right: x as LONG,
	},
)
.expect("adjust_window_rect failed");
```

amrbashir:
I'd like to make some progress in this PR if possible and last time I checked, there was two major issues with this PR:
- Undecorated windows now have shadows and resizing handles. While I think the resizing handles is Okay, The shadows are not because it will mess up with transparent windows. So, I am not sure what would be the way forward with this.
- `set_inner_size()` overflows the desired size. (Should be fixed in (4c0c760), but need someone else to confirm and review if it is the desired approach). Also the fix (4c0c760) might not work if we removed the spillover fix (4618897) by @maroider (haven't tested this scenario).

aevyrie:
> * Undecorated windows now have shadows and resizing handles. While I think the resizing handles is Okay, The shadows are not because it will mess up with transparent windows. So, I am not sure what would be the way forward with this. Remove the spillover fix?

Forgive my ignorance. Is it possible for your work in this PR to be configurable? Fixing the issue addressed in this PR would effectively unblock using winit for custom decorated windows, and making this bugfix opt-in (or auto disabled) for users of transparent windows could be a way of preventing a regression.

amrbashir:
Unfortunately this PR is not done yet and there is currently no way to manage whether the shadow is present or not. It is not the goal of this PR.

amrbashir:
I made another attempt at fixing the size overflow in [`00ee199`](https://github.com/rust-windowing/winit/pull/1891/commits/00ee199171fa0fe530ecce689cf0b6679009b976). It is inspired by the hit test function in this microsoft [guide](https://docs.microsoft.com/en-us/windows/win32/dwm/customframe#appendix-c-hittestnca-function). It works by removing the `WS_CAPTION` and `WS_SIZEBOX` from the window styles of the undecorated window before passing it to `AdjustWindowRect*` functions. It works correctly without the spillover fix from [`4618897`](https://github.com/rust-windowing/winit/pull/1891/commits/4618897d4dd7c6643830495520bc4f269410ef09). When used with the spillover fix, the window starts to shrink instead of getting bigger.

For now, I commented out the spillover fix. I don't know if it is worth to keep it anyways. What do you think @maroider.

amrbashir:
superseded by #2419

PR #1892, opened 2021-03-21
Hold on to Android NativeWindow lock to prevent races

maciejgodek:
The current Android implementation of winit aims to guarantee the validity of pointers handed out by `Window::raw_window_handle()` between a `Resumed` event and a matching `Suspended` event. I don't think this guarantee is actually upheld.

The design of the `ndk_glue` crate requires that users hold on to a read lock on the `NativeWindow` handle received from `ndk_glue::native_window()` while they intend to use it, including through implicitly dependent objects such as `RawWindowHandle` (or `EGLSurface`, and so on). Otherwise, Android may destroy the underlying resource before dependent resources are cleaned up, leading to a race condition and UB.

Unfortunately, due to a subtle bug, such user behavior would cause a deadlock in the current version of `ndk_glue`. This should no longer be the case once rust-windowing/android-ndk-rs#117 is resolved.

From that point on, in order to ensure validity of the raw pointers to Android's `ANativeWindow` handed out by `Window::raw_window_handle()`, winit's Android `EventLoop` should hold on to a read lock on the `NativeWindow` returned by `ndk_glue::native_window()` between receiving the `WindowCreated` event and the matching `WindowDestroyed` event from `ndk_glue`'s event pipe.

Other designs are possible, of course, but as long as raw pointers are being handed out, some similar protection will be necessary.

﻿Depends on: https://github.com/rust-windowing/android-ndk-rs/pull/134

### DIFF
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -13,6 +13,7 @@ use ndk::{
     native_window::NativeWindow,
 };
 use ndk_glue::{Event, Rect};
+use raw_window_handle::RawWindowHandle;
 use std::{
     collections::VecDeque,
     sync::{Arc, Mutex, RwLock, RwLockReadGuard},
@@ -72,6 +73,7 @@ impl<T: 'static> EventLoop<T> {
         Self {
             window_target: event_loop::EventLoopWindowTarget {
                 p: EventLoopWindowTarget {
+                    raw_window_handle: Default::default(),
                     _marker: std::marker::PhantomData,
                 },
                 _marker: std::marker::PhantomData,
@@ -117,6 +119,9 @@ impl<T: 'static> EventLoop<T> {
                         let native_window_lock = ndk_glue::native_window();
                         // The window could have gone away before we got the message
                         if native_window_lock.is_some() {
+                            self.window_target
+                                .p
+                                .update_native_window(native_window_lock.as_ref());
                             self.native_window_lock = Some(native_window_lock);
                             call_event_handler!(
                                 event_handler,
@@ -140,6 +145,7 @@ impl<T: 'static> EventLoop<T> {
                                 event::Event::Suspended
                             );
                         }
+                        self.window_target.p.update_native_window(None);
                     }
                     Event::Pause => self.running = false,
                     Event::Resume => self.running = true,
@@ -418,10 +424,19 @@ impl<T> Clone for EventLoopProxy<T> {
 }
 
 pub struct EventLoopWindowTarget<T: 'static> {
+    raw_window_handle: Arc<Mutex<Option<RawWindowHandle>>>,
     _marker: std::marker::PhantomData<T>,
 }
 
 impl<T: 'static> EventLoopWindowTarget<T> {
+    fn update_native_window(&self, native_window: Option<&NativeWindow>) {
+        *self.raw_window_handle.lock().unwrap() = native_window.map(|native_window| {
+            let mut handle = raw_window_handle::android::AndroidHandle::empty();
+            handle.a_native_window = unsafe { native_window.ptr().as_mut() as *mut _ as *mut _ };
+            RawWindowHandle::Android(handle)
+        });
+    }
+
     pub fn primary_monitor(&self) -> Option<monitor::MonitorHandle> {
         Some(monitor::MonitorHandle {
             inner: MonitorHandle,
@@ -456,16 +471,20 @@ impl DeviceId {
 #[derive(Clone, Copy, Debug, Default, Eq, PartialEq)]
 pub struct PlatformSpecificWindowBuilderAttributes;
 
-pub struct Window;
+pub struct Window {
+    raw_window_handle: Arc<Mutex<Option<RawWindowHandle>>>,
+}
 
 impl Window {
     pub fn new<T: 'static>(
-        _el: &EventLoopWindowTarget<T>,
+        el: &EventLoopWindowTarget<T>,
         _window_attrs: window::WindowAttributes,
         _: PlatformSpecificWindowBuilderAttributes,
     ) -> Result<Self, error::OsError> {
         // FIXME this ignores requested window attributes
-        Ok(Self)
+        Ok(Self {
+            raw_window_handle: Arc::clone(&el.raw_window_handle),
+        })
     }
 
     pub fn id(&self) -> WindowId {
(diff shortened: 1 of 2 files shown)


src/platform_impl/android/mod.rs
```diff
@@ -43,6 +44,7 @@ fn poll(poll: Poll) -> Option<EventSource> {
 
 pub struct EventLoop<T: 'static> {
     window_target: event_loop::EventLoopWindowTarget<T>,
+    native_window_lock: Option<RwLockReadGuard<'static, Option<NativeWindow>>>,
```
MarijnS95, line 47:
```suggestion
    /// Hold a readlock on the native window between `Resumed` and `Suspended` due to/because [..]
    native_window_lock: Option<RwLockReadGuard<'static, Option<NativeWindow>>>,
```

This file is quite lacking documentation wise, perhaps this little intricacy is a great place to start changing that!

MarijnS95 (approved):
That's pretty much how I had it in mind! Let's get a `Depends on: https://github.com/rust-windowing/android-ndk-rs/pull/134` in the PR description and pump out a new `ndk-glue` release quick so that the version can be bumped and this can make its way in, too.

Thanks!

EDIT:

> Note that this commit depends on recent fixes to `ndk_glue`

Feel free to (or rather, please) link to this PR directly from the commit for future reference :).

src/platform_impl/android/mod.rs
maciejgodek: Sure!

src/platform_impl/android/mod.rs
```diff
@@ -43,6 +44,10 @@ fn poll(poll: Poll) -> Option<EventSource> {
 
 pub struct EventLoop<T: 'static> {
     window_target: event_loop::EventLoopWindowTarget<T>,
+    // This read guard will be held between each pair of `Resumed` and `Suspended`
+    // events to ensure that `ndk_glue` does not release the `NativeWindow` prematurely,
+    // invalidating raw handles obtained by the user through `Window::raw_window_handle()`
+    native_window_lock: Option<RwLockReadGuard<'static, Option<NativeWindow>>>,
```
MarijnS95, line 50: You should use triple slashes (`///`) here to turn this into a documentation comment, like the suggestion described. That way it shows up in `rustdoc` output and editors/IDEs with `rustdoc` support like `rust-analyzer` :grimacing: !

MarijnS95: More specifically you can (and should!) convert all item references to intra-doc-links:

```suggestion
    /// This read guard will be held between each pair of
    /// [`event::Event::Resumed`] and [`event::Event::Suspended`] events to
    /// ensure that [`ndk_glue`] does not release the [`NativeWindow`]
    /// prematurely, invalidating raw handles obtained by the user through
    /// [`Window::raw_window_handle()`]
    native_window_lock: Option<RwLockReadGuard<'static, Option<NativeWindow>>>,
```

If you find the full path too long you can also use `` [`Resumed`](event::Event::Resumed) ``.

src/platform_impl/android/mod.rs
```diff
@@ -106,22 +112,32 @@ impl<T: 'static> EventLoop<T> {
             match self.first_event.take() {
                 Some(EventSource::Callback) => match ndk_glue::poll_events().unwrap() {
                     Event::WindowCreated => {
-                        call_event_handler!(
-                            event_handler,
-                            self.window_target(),
-                            control_flow,
-                            event::Event::Resumed
-                        );
+                        let native_window_lock = ndk_glue::native_window();
```
MarijnS95, line 115: There are two calls remaining to `ndk_glue::native_window()`. IMO these should use `self.native_window_lock` and panic if it was `None` (implying it was called outside a `Resumed` - `Suspended` "block") instead of trying to grab the window on their own regard.

That might require some refactoring as `fn raw_window_handle()` only has access to  `struct Window` which is currently empty, with `EventLoopWindowTarget` supposedly tying them together. I'm not sure what the desired flow is here, the other platform implementations should be able to help out here.

MarijnS95 (changes requested):

src/platform_impl/android/mod.rs
maciejgodek: I tried something like that first. My attempt ran into `RwLockReadGuard` being `!Send` and I gave that approach up. Will give it another try though.

maciejgodek: Ok, I came up with something. Let me know what you think.

dvc94ch (approved):
Great work @maciejgodek and @MarijnS95 . Thanks for thinking through the subleties of the api and coming up with a simple solution.

src/platform_impl/android/mod.rs
MarijnS95: I would have preferred to keep the entire thing in one place without two mutexes/locks. It looks like other backends like Wayland store modifyable data in `EventLoopWindowTarget` through a `RefCell`:

https://github.com/rust-windowing/winit/blob/86748fbc681ea070bc37e90a569c192555b402cd/src/platform_impl/linux/wayland/event_loop/mod.rs#L50-L51

That's where I'd put `native_window_lock` instead. Perhaps in a similar `*State` struct and/or with a `type` alias because it is getting quite long :sweat_smile: 

Maybe the winit-Android developers have a better overview of where to place (mutable) data shared between the `EventLoop` and `Window`; @msiglreith @dvc94ch?

maciejgodek: I would prefer that too, but I assumed that `Window` needs to be `Send`, which complicates things. I'll take another look at the other backends and see if I can glean anything more.

maciejgodek: Desktop platforms seem to make sure `Window` is `Send` and synchronize state with the event loop (?). It's not clear how important this is on Android, but if it is, then the fact that you can't put `ReadLockReadGuard` behind an `Arc<Mutex<T>>` becomes a problem.
https://github.com/rust-windowing/winit/blob/86748fbc681ea070bc37e90a569c192555b402cd/src/platform_impl/linux/wayland/window/mod.rs#L46 https://github.com/rust-windowing/winit/blob/86748fbc681ea070bc37e90a569c192555b402cd/src/platform_impl/linux/x11/window.rs#L105

maciejgodek: @MarijnS95 Still no idea how to solve this without extra synchronization. Some input from a maintainer might be useful.

MarijnS95: Unfortunately I have not had the time to look at any alternatives either, seems like wrapping our state (either the readlock or raw window handle) inside yet another mutex is the only option - given `Send` limits this might only be possible with the window handle which has to be unconditionally retrieved and stored like this PR does currently.

It would have also been nice if the lock could be transposed to remove the `Option` (only want to hold the lock if it contains a valid window) but that is not possible either.

PR #1898, opened 2021-03-29
Fix CI warnings

maroider:

I'm not too sure about silencing the `unused_code` warnings, as they only appear on nightly.


maroider:
Welp, WASM CI seems to be borked somehow. I'm not seeing any errors when building locally.

src/platform_impl/linux/mod.rs
```diff
@@ -613,7 +613,7 @@ impl<T: 'static> EventLoop<T> {
             "Failed to initialize any backend! Wayland status: {:?} X11 status: {:?}",
             wayland_err, x11_err,
         );
-        panic!(err_string);
+        std::panic::panic_any(err_string);
```
ArturKovacs, line 616: Considering that panic uses the same syntax as `format!`, which is used above, I believe this could be and probably should be written as
```rust
        panic!(
            "Failed to initialize any backend! Wayland status: {:?} X11 status: {:?}",
            wayland_err, x11_err,
        );
```

src/platform_impl/linux/x11/util/atom.rs
```diff
@@ -32,7 +32,7 @@ impl XConnection {
                     name,
                     self.check_errors(),
                 );
-                panic!(msg);
+                std::panic::panic_any(msg);
```
ArturKovacs, line 35: Same as before - `panic!`, `format!`

PR #1902, opened 2021-04-06
Don't call NSApplication.finishLaunching on window creation

madsmtm:
  - MacOS 10.14.6

`NSApplication.finishLaunching` is called internally `NSApplication.run`, and is generally not something user code should call - I couldn't find the reasoning behind this being there in the first place, git blame reveals c38110cac from 2014, so probably a piece of legacy code.

Removing this fixes creating new windows when you have assigned a main menu to the application.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- On macOS, fix creating new windows when the application has a main menu.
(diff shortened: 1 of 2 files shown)


ArturKovacs:
I tested this on 10.11 and it seems to be Oll Korrect.

PR #1903, opened 2021-04-06
MacOS: Only activate after the application has finished launching

madsmtm:
  - MacOS 10.14.6

This fixes the main menu not responding until you refocus, at least from what I can tell - though we might have to do something similar to https://github.com/linebender/druid/pull/994 to fix it on Catalina.

Additionally, activating the application when it has launched means we don't need the activation hack from https://github.com/rust-windowing/winit/pull/1318 any longer - though I would like @francesca64 to give their input here, it's very likely that I'm wrong on this point!


ArturKovacs:
I tested this (I was moving my mouse while running `cargo run --example multiwindow`) and only one of the windows show up, so it seems this unfortunately does not replace the activation hack.

madsmtm:
Oops, forgot to reply to this, but the discrepancy was indeed nicely spotted! I too will have to spend some more time on this / on #1918 at some point :wink:

madsmtm:
So, I found yet another hacky way to activate the windows, so this now works on the `multiwindow` example - however, the hack is much less involved, it only fixes the case where windows are created before `applicationDidFinishLaunching` and everything else works without it.

It should be noted that it could be implemented differently: By tracking windows created before `applicationDidFinishLaunching` instead of querying information about the windows later on. The chosen approach seemed simpler to me, but it might have some downsides e.g. if more complex things that are technically also windows also are created in this time frame

ArturKovacs:
I don't want to jinx it, but @madsmtm, I think you are a genius. It works perfectly.

ArturKovacs:
I somewhat hastily merged this, but @adamnemecek could you test if this works for you?

madsmtm:
Well, it shouldn't work for them (and everyone else on Catalina or above), it's missing https://github.com/rust-windowing/winit/pull/1922, right?

ArturKovacs:
All that #1922 does is that it allows you to specify an activation policy other than `NSApplicationActivationPolicyRegular`. So if I'm not mistaken than if this doesn't work on 10.15, then #1922 won't work either.

@casperstorm would mind testing this to see if the default menu can be opened right after startup, on the master branch?

madsmtm:
The vital thing is, it also moves the setting of the activation policy to `applicationDidFinishLaunching`, where previously it was called before, and apparently that causes problems on Catalina and above.

casperstorm:
> @casperstorm would mind testing this to see if the default menu can be opened right after startup, on the master branch?

It indeed does open right after startup on 10.15.7.

ArturKovacs:
Awesome, thank youuu 😊

PR #1905, opened 2021-04-13
Correct the false documentation about macOS dpi

### LINKED ISSUE #1828: macOS allows configuring the scaling factor, but the docs say otherwise
As of winit 0.24.0, the docs state:

> **macOS:** "retina displays" have a scale factor of 2.0. Otherwise, the scale factor is 1.0. Intermediate scale factors are never used. It's possible for any display to use that 2.0 scale factor, given the use of the command line.

This is false. macOS allows you to configure the scale factor:

Display preferences

It can be between 1 and 2, or even larger than 2. I believe their newer machines (>=2016) use scale factors under `2.0` by default, but my only experience with them is in the Apple store, so someone may have just manually changed that setting.

For the record, my display uses 2x scaling because I respect my pixels, but those who don't might be using non-`2.0` scale factors.

ArturKovacs:
Closes #1828 

### DIFF
--- a/src/dpi.rs
+++ b/src/dpi.rs
@@ -69,9 +69,10 @@
 //!   selection of "nice" scale factors, i.e. 1.0, 1.25, 1.5... on Windows 7, the scale factor is
 //!   global and changing it requires logging out. See [this article][windows_1] for technical
 //!   details.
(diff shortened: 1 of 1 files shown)


ArturKovacs:
@LoganDark please review this.

LoganDark:
> @LoganDark please review this.

LGTM

Issue #1893, opened 2021-03-24
Windows: Check removal of CS_OWNDC style


msiglreith:
Setting this style for OpenGL would be nice to have but comes with some limitations when using in combination with window styles, resulting in issues to provide functionality.

PR #1910, opened 2021-04-20
Windows: Remove owned DC context per window

msiglreith:
The flag is not required for OpenGL and comes with several limitations when used with other style flags.

Tested with `glutin` and `raw-gl-context` against sample applications (inlc multi window setups with shared contexts) on Windows 10. Didn't affect context handling or rendering in these scenarios.

Fixes https://github.com/rust-windowing/winit/issues/1893

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- On Windows, remove internally unique DC per window.
 - macOS: Remove the need to call `set_ime_position` after moving the window.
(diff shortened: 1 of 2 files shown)


maroider:
One thing I'm very curious about is whether the `CS_OWNDC` class style has any measurable impact on performance.
Unfortunately, I don't have any experience with benchmarking OpenGL, so I don't feel comfortable testing it myself.

msiglreith:
Totally forgot about this one! Tried with a simple example (`grr` + `raw-gl-context`) with some gpu timers and any difference seems to go down in noise

PR #1901, opened 2021-04-04
Mac: Redraw immediately to prevent shaking on window resize

xiaopengli89:
Fixes #1837, gfx-rs/wgpu-rs#536, rust-windowing/glutin#1340, servo/webrender#1640

PR #853, opened 2019-04-25
[WIP] Macos Evlp2

ZeGentzy:
I kind of fucked up and had to manually back port changes. Git is hard. It will all get squashed in the end, so doesn't matter.

I think this is correct, or maybe not. All I know is it compiles :)

PR #1911, opened 2021-04-20
macos: Handle resizes synchronously

scoopr:
Allows for a lot better resize behaviour on macos, when allowed
to redraw the frame within the resize event callback.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 - On Windows, added `WindowExtWindows::set_enable` to allow creating modal popup windows.
 - On macOS, emit `RedrawRequested` events immediately while the window is being resized.
 - Implement `Default`, `Hash`, and `Eq` for `LogicalPosition`, `PhysicalPosition`, `LogicalSize`, and `PhysicalSize`.
+- On macOS, emit `Resized` events immediately from the callback.
 
 # 0.24.0 (2020-12-09)
 
--- a/src/platform_impl/macos/app_state.rs
+++ b/src/platform_impl/macos/app_state.rs
@@ -331,6 +331,10 @@ impl AppState {
         HANDLER.handle_nonuser_event(EventWrapper::StaticEvent(Event::RedrawRequested(window_id)));
     }
 
+    pub fn handle_event(wrapper: EventWrapper) {
+        HANDLER.handle_nonuser_event(wrapper);
+    }
+
     pub fn queue_event(wrapper: EventWrapper) {
         if !unsafe { msg_send![class!(NSThread), isMainThread] } {
             panic!("Event queued from different thread: {:#?}", wrapper);
--- a/src/platform_impl/macos/window_delegate.rs
+++ b/src/platform_impl/macos/window_delegate.rs
@@ -97,12 +97,18 @@ impl WindowDelegateState {
         AppState::queue_event(wrapper);
     }
 
-    pub fn emit_resize_event(&mut self) {
+    pub fn handle_resize_event(&mut self) {
         let rect = unsafe { NSView::frame(*self.ns_view) };
         let scale_factor = self.get_scale_factor();
         let logical_size = LogicalSize::new(rect.size.width as f64, rect.size.height as f64);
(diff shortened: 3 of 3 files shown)


scoopr:
I thought #1901 would of fixed this, but it doesn't do anything when the view has a metal or opengl CALayer attached, as `display` won't be called, as far as I could understand.

scoopr:
Hmm, on further testing, this fixes especially the case when I actually render in the `Resized` event, in conjunction with drawing on `MainEventsCleared`. 

So I'm not sure, but is the OS resizing the window, winit giving `MainEventsCleared`, and after that the `Resized` event?

But in any case, I believe rendering within the `Resized` event handler is somewhat required if I want to enable and use `presentsWithTransaction` with the `CAMetalLayer`

I'm curious is all the queuing of the events all that necessary overall though.

ArturKovacs:
I think that the biggest problem -that #1901 did not address- is that now you get the redraw events before the resized events, which is almost certainly unexpected by developers. Note that it's advised to render in the [`RedrawRequested`](https://docs.rs/winit/0.24.0/winit/event/enum.Event.html#variant.RedrawRequested) event; in which case #1901 does fix the flickering/blank screen issue.

However I do believe that there's merit in solving the problem of "getting the redraw events before the resized events" and this PR does solve that. Although I am somewhat uncomfortable with synchronously dispatching some events while others are queued. I have to admit that I don't really understand why we queue events in the first place rather than just immediately dispatching all of them (I tried getting to the bottom of this but all I found was #853 which didn't help enlightening me). 

Maybe @francesca64 knows why events are queued and only dispatched in the `AppState::cleared` function. That would probably help in deciding if it's okay to synchronously dispatch the resized event from the "windowDidResize".

scoopr:
I'm totally fine with changing all the callbacks to non-queueing, thats just changing the `emit_event` to not queue but handle it directly.

it seems Francesca did set it to send resizes directly in https://github.com/francesca64/winit/commit/b75073a5b2a8d65ab8806a00ffee390752255c8c but then later in https://github.com/rust-windowing/winit/pull/1173 it was then deferred again. Unfortunately it seems Aleksi has removed his github account.

In there he both defers the event and change a runloop mode.

I guess I didn't double-check that the other events come through, but with a quick check, key events are not coming through while resizing, but thats not too unexpected. But I did get continuous rendering of the app still (and with  maineventscleared), so should be fine on that front.

ArturKovacs:
It would be to the very least surprising if there wouldn't be a good reason why events are queued instead of dispatched immediately. Unfortunately I don't know what that reason is. As another attempt I'm tagging @goddessfreya to help here. Otherwise I can't think of a better way of finding the reason, than to try to make every event dispatched immediately.

francesca64:
@scoopr thanks for doing the archaeology on that!

@ArturKovacs neither Osspial nor me have been able to communicate with Freya over either email or Discord in the past year or so, and her website's been gone for just as long, so all we can do is hope that she's alright. On the bright side, she likely wouldn't know about this, since #853 was her politely rebasing my macOS EL2 work that I'd sorta kinda abandoned. 

For more historical background, I based the macOS `AppState` on the struct of the same name in the iOS backend. This is because the latter was written by my work teammate @mtak-. He queued events, so I did too, since it made sense to me at the time.

As for resize in particular... we apparently discussed issues I found with queuing resizes, but since 99% of the discussion happened in calls, that's all the info left now. It's possible he remembers why it was important to queue events in the first place, but he's presently on leave, so I'm not sure when I'd be able to ask him.

ArturKovacs:
First of all thank you @francesca64 for clarifying the situation with Freya. I wish her the best.

And thanks for describing the history of how we came to queue events, it is definitely helpful.

I'll probably spend some time with this in the near future as I think it's an important issue, but in any case I'm looking forward to hearing from @mtak- because even if the immediate dispatch would seem to work, I'd have the feeling that we are missing something that @mtak- knew about.

madsmtm:
See https://github.com/rust-windowing/winit/issues/2640#issuecomment-2284333496, I plan to fix this myself, but it'll take a while.

Issue #1768, opened 2020-11-17
[web_sys] Standard web hotkeys are prevented


ostwilkens:
When I'm on the web, I expect to be able to use Ctrl+R, F5, F12 etc. 
I can understand a developer might want to hijack some keys, but that can easily be done using javascript in the host document. 
The way it is now, it is not possible to get the default behaviour (working F5) using winit. 

#1576 should be reverted.

ostwilkens:
For an immersive game, you could argue that preventing these hotkeys is correct. 
The scope of winit however, is both apps and games. For a web app, you expect these keys to work as normal.

alvinhochun:
Making it an option makes more sense than reverting the behaviour completely.

I'm curious though: How would a webapp make use of `winit`? Implementing a web UI using `winit` in its current state will result in significant accessibility and usability issues that it doesn't seem to make much sense outside of games.

ostwilkens:
Yes, an option would be fine :-)

The reason I mentioned web apps is this quote from features.md: 
> ## Winit Scope
> Winit aims to expose an interface that abstracts over window creation and input handling, and can be used to create both **games and applications**.

There are kinds of apps that are best implemented in a game engine. Visualization apps often require good rendering performance. 
While the current tools may not be production ready for this usecase, I do believe the web will be moving more and more to compiled wasm apps, allowing the developer full control over the experience. 
I can absolutely see a future filled with flashy, interactive web applications using winit. 

My personal usecase is a full-viewport game using bevy_webgl2. (think agar.io).

Maximkaaa:
> I'm curious though: How would a webapp make use of `winit`? Implementing a web UI using `winit` in its current state will result in significant accessibility and usability issues that it doesn't seem to make much sense outside of games.

In my case I'm developing a cross-platform GIS library (drawing maps), and web is one of the targets. The consumers of the library are mainly JS applications creating web interface, and using rust-driven map just a black box component. For these applications I definitely don't want `F5` or `Ctrl+Shift+I` to be prevented. I actually have an issue for that in my repo: https://github.com/Everpoint/yoda/issues/3

But even for games, disabling `F5`, `F11` or `F12` in browsers is very dubious decision.

A1Liu:
It would be nice if this were fixed.

[...]

PR #1919, opened 2021-04-28
android: ignore volume keys

alula:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- On Android, ignoring volume up/down key events since most applications should not override the default behavior.
 - On macOS, fix creating new windows when the application has a main menu.
 - On Windows, fix fractional deltas for mouse wheel device events.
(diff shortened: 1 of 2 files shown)


msiglreith (approved):
Thanks, fine for me

cc @MarijnS95 to check that you don't require it by chance

MarijnS95:
I miss proper justification why this is disabled. The main problem appears to be that `winit` marks every key as "handled" regardless of what the application does with it.

What if someone built an audio application with Winit and listens for volume keys for some reason or another, and suddenly doesn't receive them anymore?

Besides, we're at 65 characters of indentation now, this is getting beyond maintainable.

maroider:
> The main problem appears to be that winit marks every key as "handled" regardless of what the application does with it.

It seems like there's a need to expose the ability to set the "handled" state of certain events, and not just on Android (see #1768).

torokati44:
> It seems like there's a need to expose the ability to set the "handled" state of certain events, and not just on Android

I definitely agree, but until that, as a stopgap, I'd love to see this merged!

PR #1920, opened 2021-04-29
MacOS: Use `initialFirstResponder` instead of `makeFirstResponder`

madsmtm:
Minor nit, as recommended by [the documentation](https://developer.apple.com/documentation/appkit/nswindow/1419366-makefirstresponder?language=objc).

This changes the point at which `becomeFirstResponder` (a notification we don't use) is emitted to later on when `makeKeyAndOrderFront`/`makeKeyWindow` is called, which I think is more correct behaviour.

    - MacOS 10.14.6 Mojave


ArturKovacs:
Awesome, thanks!

PR #1922, opened 2021-04-29
Require setting the activation policy on the event loop

ArturKovacs:
The activation policy could previously be specified through the window builder. This was incorrect because the activation policy is an application-wide property, not window-specific.

### DIFF
--- a/src/platform/macos.rs
+++ b/src/platform/macos.rs
@@ -4,8 +4,9 @@ use std::os::raw::c_void;
 
 use crate::{
     dpi::LogicalSize,
-    event_loop::EventLoopWindowTarget,
+    event_loop::{EventLoop, EventLoopWindowTarget},
     monitor::MonitorHandle,
+    platform_impl::get_aux_state_mut,
     window::{Window, WindowBuilder},
 };
 
@@ -100,8 +101,6 @@ impl Default for ActivationPolicy {
 ///  - `with_titlebar_buttons_hidden`
 ///  - `with_fullsize_content_view`
 pub trait WindowBuilderExtMacOS {
-    /// Sets the activation policy for the window being built.
-    fn with_activation_policy(self, activation_policy: ActivationPolicy) -> WindowBuilder;
     /// Enables click-and-drag behavior for the entire window, not just the titlebar.
     fn with_movable_by_window_background(self, movable_by_window_background: bool)
         -> WindowBuilder;
@@ -122,12 +121,6 @@ pub trait WindowBuilderExtMacOS {
 }
 
 impl WindowBuilderExtMacOS for WindowBuilder {
-    #[inline]
-    fn with_activation_policy(mut self, activation_policy: ActivationPolicy) -> WindowBuilder {
-        self.platform_specific.activation_policy = activation_policy;
-        self
-    }
-
     #[inline]
     fn with_movable_by_window_background(
         mut self,
@@ -183,6 +176,20 @@ impl WindowBuilderExtMacOS for WindowBuilder {
     fn with_has_shadow(mut self, has_shadow: bool) -> WindowBuilder {
         self.platform_specific.has_shadow = has_shadow;
         self
+    }
+}
+
+pub trait EventLoopExtMacOS {
+    /// Sets the activation policy for the application.
+    /// It is set to `NSApplicationActivationPolicyRegular` by default.
+    fn set_activation_policy(&mut self, activation_policy: ActivationPolicy);
+}
+impl<T> EventLoopExtMacOS for EventLoop<T> {
+    #[inline]
+    fn set_activation_policy(&mut self, activation_policy: ActivationPolicy) {
+        unsafe {
+            get_aux_state_mut(&**self.event_loop.delegate).activation_policy = activation_policy;
+        }
     }
 }
 
--- a/src/platform/macos.rs
+++ b/src/platform/macos.rs
@@ -180,8 +180,11 @@ impl WindowBuilderExtMacOS for WindowBuilder {
 }
 
 pub trait EventLoopExtMacOS {
-    /// Sets the activation policy for the application.
-    /// It is set to `NSApplicationActivationPolicyRegular` by default.
+    /// Sets the activation policy for the application. It is set to
+    /// `NSApplicationActivationPolicyRegular` by default.
+    ///
+    /// This function only takes effect if it's called before calling [`run`](crate::event_loop::EventLoop::run) or
+    /// [`run_return`](crate::platform::run_return::EventLoopExtRunReturn::run_return)
     fn set_activation_policy(&mut self, activation_policy: ActivationPolicy);
 }
 impl<T> EventLoopExtMacOS for EventLoop<T> {
--- a/src/platform_impl/macos/app_delegate.rs
+++ b/src/platform_impl/macos/app_delegate.rs
@@ -13,9 +13,9 @@ use std::{
 static AUX_DELEGATE_STATE_NAME: &str = "auxState";
 
 pub struct AuxDelegateState {
-    /// We store this value in order to be able defer setting the activation policy until
+    /// We store this value in order to be able to defer setting the activation policy until
     /// after the app has finished launching. If the activation policy is set earlier, the
-    /// menubar is unresponsive for example.
+    /// menubar is initially unresponsive on macOS 10.15 for example.
     pub activation_policy: ActivationPolicy,
 }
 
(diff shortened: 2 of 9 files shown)


ArturKovacs:
@madsmtm please review this.

src/platform_impl/macos/window.rs
```diff
@@ -359,11 +338,11 @@ impl UnownedWindow {
...
 
-        create_app(pl_attribs.activation_policy).ok_or_else(|| {
+        let ns_app = unsafe { NSApp() };
+        if ns_app == nil {
             unsafe { pool.drain() };
-            os_error!(OsError::CreationError("Couldn't create `NSApplication`"))
-        })?;
-
+            return Err(os_error!(OsError::CreationError("`NSApp()` returned nil")));
+        }
```
madsmtm, line 345: I don't think there's a reason for calling `NSApp` here any longer?

src/platform_impl/macos/app_delegate.rs
```diff
@@ -1,9 +1,23 @@
...
     runtime::{Class, Object, Sel},
 };
+use std::{
+    cell::{RefCell, RefMut},
+    os::raw::c_void,
+};
+
+static AUX_DELEGATE_STATE_NAME: &str = "auxState";
+
+pub struct AuxDelegateState {
```
madsmtm, line 15: I think I would have just put this into the existing `Handler`/`HANDLER` state at some point, for example during `EventLoop::run_return` - but this works as well, so I don't have a strong preference here

src/platform/macos.rs
```diff
@@ -186,6 +179,20 @@ impl WindowBuilderExtMacOS for WindowBuilder {
     }
 }
 
+pub trait EventLoopExtMacOS {
+    /// Sets the activation policy for the application.
+    /// It is set to `NSApplicationActivationPolicyRegular` by default.
+    fn set_activation_policy(&mut self, activation_policy: ActivationPolicy);
```
madsmtm, line 185: It should be documented that this can only be called before `EventLoop::run`.

Or, alternatively, it should be as a `with_activation_policy_and_user_event` creation method (and in the future, if the pattern is common enough, we could create `EventLoopBuilder`)

src/platform_impl/macos/app_delegate.rs
```diff
@@ -1,9 +1,23 @@
...
+    /// We store this value in order to be able defer setting the activation policy until
+    /// after the app has finished launching. If the activation policy is set earlier, the
+    /// menubar is unresponsive for example.
```
madsmtm, line 18: Add note about Catalina, e.g.:
```suggestion
    /// menubar, for example, is initially unresponsive on Catalina and above.
```

src/platform_impl/macos/app_delegate.rs
```diff
@@ -14,17 +28,51 @@ lazy_static! {
...
 
         AppDelegateClass(decl.register())
     };
 }
 
-extern "C" fn did_finish_launching(_: &Object, _: Sel, _: id) {
+/// Safety: Assumes that Object is an instance of APP_DELEGATE_CLASS
+pub unsafe fn get_aux_state_mut(this: &Object) -> RefMut<'_, AuxDelegateState> {
+    let ptr: *mut c_void = *this.get_ivar(AUX_DELEGATE_STATE_NAME);
+    // Watch out that this needs to be the correct type
```
madsmtm, line 47: Can't you just set `ptr` to the correct type above? (If not, there's probably something in Rust's type conversion that I don't understand, bear with me)

src/platform_impl/macos/app_state.rs
```diff
@@ -457,3 +464,18 @@ unsafe fn window_activation_hack(ns_app: id) {
...
+        let ns_app = NSApp();
+        // We need to delay setting the activation policy and activating the app
+        // until we have the main menu all set up. Otherwise the menu won't be interactable.
+        let act_pol = get_aux_state_mut(app_delegate).activation_policy;
+        ns_app.setActivationPolicy_(match act_pol {
+            ActivationPolicy::Regular => NSApplicationActivationPolicyRegular,
+            ActivationPolicy::Accessory => NSApplicationActivationPolicyAccessory,
+            ActivationPolicy::Prohibited => NSApplicationActivationPolicyProhibited,
+        });
+        let () = msg_send![ns_app, activateIgnoringOtherApps: YES];
```
madsmtm, line 479: This is already called in `AppState::launched` - and should first be called after `window_activation_hack` (I would think that yields better behaviour, though maybe it doesn't make a difference)

src/platform/macos.rs
ArturKovacs: Will do!

src/platform_impl/macos/app_delegate.rs
ArturKovacs: I'm not a big fan of the`HANDLER` global; everything has to be Atomic or Mutex there.

src/platform_impl/macos/app_delegate.rs
ArturKovacs: This is not true, it's also unresponsive on my 10.11. Is there a version where it's responsive from the start? If there is this could be worded like "the menubar is unresponsive on some systems"

src/platform_impl/macos/app_delegate.rs
ArturKovacs: The ivar itself is `*mut c_void` so I guess this better exposes what is happening.

But as far as I know, I could just write `let ptr: *mut RefCell<AuxDelegateState> = ...` but I'm not a 100% sure and I'm also not sure if the ivar has to be a void pointer instead of a pointer with the actual type.

src/platform_impl/macos/app_delegate.rs
```diff
@@ -14,17 +28,51 @@ lazy_static! {
...
+    (*(ptr as *mut RefCell<AuxDelegateState>)).borrow_mut()
+}
+
+extern "C" fn new(class: &Class, _: Sel) -> id {
+    unsafe {
+        let this: id = msg_send![class, alloc];
+        let this: id = msg_send![this, init];
```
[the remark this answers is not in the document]
ArturKovacs: I may be incorrect but wouldn't that only allocate memory for the superclass? If so that's incorrect, because this class has additional fields compared to the superclass.

src/platform_impl/macos/app_state.rs
ArturKovacs: You are correct, I accidentally left it there.

src/platform_impl/macos/window.rs
ArturKovacs: Thanks for pointing this out I absolutely did not realize that.

src/platform_impl/macos/app_delegate.rs
madsmtm: On the current `master`, that is, after #1903, the menu is responsive on my machine, running MacOS 10.14.6 Mojave - and I thought it was on yours as well? But if I'm mistaken there, then you're right, `the menubar is unresponsive on some systems` is better.

src/platform_impl/macos/app_delegate.rs
madsmtm: Fair enough, let's keep it as-is, this we at least know is correct

src/platform_impl/macos/app_state.rs
```diff
@@ -457,3 +464,17 @@ unsafe fn window_activation_hack(ns_app: id) {
         }
     }
 }
+fn apply_activation_policy(app_delegate: &Object) {
+    unsafe {
+        use cocoa::appkit::NSApplicationActivationPolicy::*;
+        let ns_app = NSApp();
+        // We need to delay setting the activation policy and activating the app
+        // until we have the main menu all set up. Otherwise the menu won't be interactable.
```
madsmtm, line 472:
```suggestion
        // until `applicationDidFinishLaunching` has been called. Otherwise the
        // menu bar won't be interactable.
```

I mean, we call this before `menu::initialize`, so I think this new wording is more correct?

madsmtm (commented):
Your changes looks fine, now it's just comments :wink:

src/platform_impl/macos/app_state.rs
ArturKovacs: You are absolutely right.

src/platform_impl/macos/app_delegate.rs
ArturKovacs: Oh. Now I understand what I was confused about. Yes, after #1903 everything works on my system (10.11), but now you enlightened me that that's not the case on `10.15` and `11`.

emilk:
This created a regression: https://github.com/rust-windowing/winit/issues/2051

PR #1931, opened 2021-05-07
Fix randr dpi with reported 1mm size displays

Chiheisen:

Some monitors have broken / wrong EDID like the Odyssey G9, which results in randr reporting them as size of 1mm x 1mm.
xrandr output: `DP-2 connected primary 5120x1440+0+1440 (normal left inverted right x axis y axis) 1mm x 1mm`

The current sanity check only looks for 0mm x 0mm, PR just extends the check to 0/1 instead of just 0. 
There theoretically could be displays with a size of 1x1, but I'm not aware of such displays existing (/ being useful).

### DIFF
--- a/src/platform_impl/linux/x11/util/randr.rs
+++ b/src/platform_impl/linux/x11/util/randr.rs
@@ -18,8 +18,8 @@ pub fn calc_dpi_factor(
     (width_mm, height_mm): (u64, u64),
 ) -> f64 {
     // See http://xpra.org/trac/ticket/728 for more information.
-    if width_mm == 0 || height_mm == 0 {
-        warn!("XRandR reported that the display's 0mm in size, which is certifiably insane");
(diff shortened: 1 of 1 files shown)


kchibisov:
Nah, it's not really a fix, they are not exactly 1x1 or 0x1, they could be 17x22mm(reported by nvidia drivers sometimes). Which also gives insane DPI. In general, I'd clamp all dpi values higher than 35 to 1(well 20 would be already enough, but 35 is better).

You can explain this number by picking some phone screen, e.g. 5 inches, and some insane resolution for it, like 8k/16k, than everything that is higher than that is likely garbage from the drivers.

kchibisov:
I was already doing that, but it faced some skepticism, though, it doesn't matter know

https://github.com/rust-windowing/winit/pull/1403

kchibisov:
Fixed upstream in b4774861db2f765ff016b1a3370ebf03debbd30d

PR #1938, opened 2021-05-14
Use windows-rs instead of winapi-rs

clemenswasser:
We should probably adopt [windows-rs](https://github.com/microsoft/windows-rs) since they are now the official bindings to Windows for Rust coming directly from Microsoft.

This PR is currently blocked by microsoft/windows-rs#81
This functionality is required in `src/platform_impl/windows/drop_handler.rs`.
Because of that this PR does not build and I can not test the changes I made.
Therefore I probably broke many things.


maroider:
I'm very sceptical of adopting `windows-rs` before the whole "prebuilt-bindings as a crate" thing (microsoft/windows-rs#432) is figured out, as build times seem to be impacted considerably, which is already something people are dissatisfied with Winit for. Cross-compiling from Linux to Windows is also broken in `windows-rs` (microsoft/windows-rs#638, which is blocked on rust-lang/rust#58713), and I'm fairly certain that some users of Winit cross-compile from Linux to Windows.

clemenswasser:
> I'm very sceptical of adopting `windows-rs` before the whole "prebuilt-bindings as a crate" thing ([microsoft/windows-rs#432](https://github.com/microsoft/windows-rs/issues/432)) is figured out, as build times seem to be impacted considerably, which is already something people are dissatisfied with Winit for. Cross-compiling from Linux to Windows is also broken in `windows-rs` ([microsoft/windows-rs#638](https://github.com/microsoft/windows-rs/issues/638), which is blocked on [rust-lang/rust#58713](https://github.com/rust-lang/rust/issues/58713)), and I'm fairly certain that some users of Winit cross-compile from Linux to Windows.

Agree with all of your points. I wanted to make a draft PR sooner than later to discuss theses kinds of problems.
I would also like to wait a bit more to be a bit more flexible with regards to windows-rs development.

madsmtm:
Another point, the direct dependency [`parking_lot`](https://github.com/Amanieu/parking_lot) uses `winapi`, so we would end up with both `winapi` and `windows-rs` in our dependency tree

MarijnS95:
@clemenswasser Why did this get closed, especially now that the prebuilt API crate has just been released? GNU cross-compilation support is being worked on next, and while it might definitely be a good idea to let it all stabilize for a little while and evaluate major pain points with the new windows-rs crate, now seems a better time than ever to be on top of this one.

clemenswasser:
> @clemenswasser Why did this get closed, especially now that the prebuilt API crate has just been released? GNU cross-compilation support is being worked on next, and while it might definitely be a good idea to let it all stabilize for a little while and evaluate major pain points with the new windows-rs crate, now seems a better time than ever to be on top of this one.

Yes. I know. I am working on it. I closed this, because it is quite outdated and it's easier to start fresh.

MarijnS95:
@clemenswasser Glad to hear that, good to know and thank you for still working on this! Otherwise I might have started doing the same in parallel.

maroider:
Since no-one else linked the new PR, then I guess I'll do so: #2057

PR #1939, opened 2021-05-15
Release 0.25.0

### LINKED ISSUE #1926: v0.25.0 release
The list of changes under "Unreleased" is approximately as long as the list under 0.24.0

It seems reasonable to make a new release at this point.

Is there anyone opposing to this? @francesca64 @maroider @msiglreith @murarth @kchibisov @madsmtm

ArturKovacs:
I can't push directly to the master, so it's a PR. This might mean that @francesca64, you will need to push the tag to trigger the publishing workflow. (It might also be easier if you make the commit yourself and just close this PR)

Closes #1926

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# Unreleased
+# 0.25.0 (2021-05-15)
 
 - **Breaking:** On macOS, replace `WindowBuilderExtMacOS::with_activation_policy` with `EventLoopExtMacOS::set_activation_policy`
 - On macOS, wait with activating the application until the application has initialized.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "winit"
-version = "0.24.0"
+version = "0.25.0"
 authors = ["The winit contributors", "Pierre Krieger <pierre.krieger1708@gmail.com>"]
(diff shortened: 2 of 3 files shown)


ArturKovacs:
I pushed the tag `v0.25.0`

francesca64:
Great stuff! If I'm not mistaken, [creating a release through GitHub's UI](https://github.com/rust-windowing/winit/releases) is probably the easiest way to create a tag in the future. That has the side-effect of creating a 1-to-1 correspondence of GitHub releases to crates.io releases, too!

ArturKovacs:
Now I see what you were talking about. I created the tag at some weird merge commit which isn't even on the master branch here. I knew something is going to go wrong so I double checked and triple checked everything and it still went wrong. It's annoying to say the least, but let's hope it's not going to cause an issue for anyone.

francesca64:
@ArturKovacs it's okay! These things happen, and it's such a minor divergence that it shouldn't really ever be an issue for anyone.

@madsmtm that would no longer perfectly reflect the state of history when 0.25.0 was released on crates.io, though if the contents are the same, that distinction might not be observable. If that's the case, then I'm personally fine with either decision.

ArturKovacs:
Alright, let's leave everything as it is. If this causes an issue for someone then we should address the problem.

Issue #1662, opened 2020-08-18
Deprecate stdweb support?


ryanisaacg:
We're currently on the hook for maintaining two different web backends, and I'm not sure how much sense that makes going forward.

Supporting both `stdweb` and `web-sys` made sense when the web backend was created, because they were both actively developed projects with no clear 'better' choice. I'm not sure it makes sense going forward, given [stdweb](https://github.com/koute/stdweb) hasn't had a commit merged in around 10 months, including various PRs that have been piling up.

Additionally, contributions from downstream users (the lifeblood of the web backend, now that I'm too busy to actively develop it) don't always support `stdweb` (see #1659, #1652)

I see two possible paths forward:
- Continue to merge PRs that exclusively fix or improve the `web-sys` backend, letting `stdweb` fall out of feature parity
- Make a hard break and remove `stdweb` support in the next breaking change

francesca64:
Looks like we're not the only ones: https://github.com/sebcrozet/kiss3d/pull/241

I'd favor outright removing it, unless you believe it'd be very upsetting to users. I don't think we should have support for things we don't actually support, since then we'd have to document "this is the worse backend that we care less about, so you probably shouldn't use it", which would really raise the question of why it's included in the first place.

alvinhochun:
(In reply to https://github.com/rust-windowing/winit/issues/1704#issuecomment-694326955)
> deprecating it now and waiting to remove it until 0.24.0 would be the conservative approach. However, if you believe that fairly few people are currently using the `stdweb` backend and/or if migrating is easy, then I think we should just be aggressive and remove it now.

I can't tell exactly how many people are using the `stdweb` backend, but [searching for `winit` + `stdweb` on github](https://github.com/search?q=filename%3ACargo.toml+winit+stdweb&type=Code) yields 24 results, while [searching for `winit` + `web-sys`](https://github.com/search?q=filename%3ACargo.toml+winit+web-sys&type=Code) yields 72 results. This does not account for any projects not publicly accessible on github, and the search results does not indicate actual usage.

I scanned through the search result and it seems like [A/B Street](https://github.com/dabreegster/abstreet) is the only project on github that actually is using `winit` with `stdweb`. I think we should at least ask for their opinion.

I wouldn't say migrating from `stdweb` to `web-sys` is easy. While `stdweb` and `web-sys` can be used together, not being able to use `cargo-web` may be a significant downside to some.

alvinhochun:
[...]

PR #1941, opened 2021-05-16
Remove support for `stdweb`

maroider:

We have now had two releases of Winit (`0.24` and `0.25`) with the `stdweb` backed marked as deprecated (see #1662 and #1712). Removing it now seems like a reasonable thing to do at this point.

There is still some documentation to update, but this is mostly what I'd imagine would be required to remove `stdweb` support from Winit.

I do have one thing I'd like to get some opinions on: Should there still be a `web-sys` feature, or should it be unconditionally built now? I don't think there are any viable active alternatives to `web-sys` at the moment, although I know [@kettle11](https://github.com/kettle11) is working on [something](https://github.com/kettle11/kwasm) vaguely in that direction.

I know you've stepped down from maintaining Winit, @ryanisaacg, but I'd appreciate it if you had any comments. Similarly, I'd also appreciate any comments from @alvinhochun.

cc @ArturKovacs @francesca64 @msiglreith, as you guys are the remaining maintainers


alvinhochun:
I'm fine with removing `stdweb` support.

I'm not sure about leaving the feature flag in or not. It would be nice to be able to easily add support for another web binding crate in the future, but as long as there isn't an alternative now it may be more of a burden.

ryanisaacg:
Personally, I think it's unlikely there will be a new binding crate incompatible with the web-sys interface; I'm not aware of any that exist, other than stdweb (which predates it, is unmaintained, and is generally inferior). My recommendation would be to jettison all of the feature flags, and have the web target specifically and only be compatible with web-sys.

This is also a breaking change either way, and should be marked as such.

(However, take this with a grain of salt. I bet on stdweb and was wrong; maybe this bet is wrong too?)

msiglreith:
I'm fine as well, including removing the feature flags.
Regarding `wasm-binding`, maybe there will be needs for compatibility with similar libraries (building on top of web-sys) in the future, but it might be possible to get rid of the dependency as well I guess.

ArturKovacs:
I think a sufficient number of experts have spoken and I don't have a horse in this race, so I'd say let's move on with this, and also remove the `web-sys` feature flag in addition to the `stdweb` one.

francesca64:
I also agree with what's been said.

zicklag:
Sounds good to me. :+1:

ArturKovacs:
@zicklag 😄 that's exactly the kind of energy we need here.

maroider:
CI seems to be mostly happy now, and I think I've made all the required changes to documentation and the like. I'd like to run a sanity-check with a real program later, but this PR should otherwise be ready to be merged.

madsmtm (approved):

ArturKovacs (approved):

dsluijk (approved):

ArturKovacs:
I'd say this can be merged, but I think @francesca64 only you wield the power to do so.

msiglreith:
It looks like there are some outdated checks leftover in the branch settings which block the merge process:
grafik

msiglreith:
@francesca64 Could you look at the checks above by chance? I assume it requires some changes in the admin settings.

ArturKovacs:
I guess I should have written this earlier, but note that this was also brought up here https://github.com/rust-windowing/winit/pull/1952#issuecomment-850223936

PR #1944, opened 2021-05-18
Implemented `focus_window` 

### LINKED ISSUE #1750: Feature: bring window to top

maxded:
Fixes #1750

`Window::focus_window` brings the window to the front and sets input focus to that window. As described in the documentation, this method steals focus from other applications and should be used with care. Even though this method can cause quite a disruptive user experience I do believe winit should provide the option to users of this crate as valid use cases exist. Once #1764 has been merged, documentation could be expanded to redirect users to `request_user_attention` for a less disruptive way of getting attention to a window.

Decided to call it `focus_window` but the name could be changed for more clarity. Some APIs, like `set_minimized`, have a boolean parameter for toggling the, in this case, "minimized" state. For `focus_window` I decided not to do that as "unfocusing" is quite strange, like, where does the focus go? Also deliberately made the choice to "ignore" `focus_window` on minimized and not visible windows for improved user experience.

Don't have a Linux X11 machine at my disposal so the implementation might/is probably incorrect.

- [x] Tested on Windows
- [x] Tested on macOS  
- [ ] Tested on X11


maxded:
This is a continuation on #1767, which I've accidentally removed and was unable to recover. @ArturKovacs, you should be able to push your changes now. Sorry for the delay!

ArturKovacs:
Not a problem, these things happen especially with PRs that are left open for long. I'll fix the conflicts as well while I'm at it

src/platform_impl/macos/window.rs
```diff
@@ -269,6 +269,7 @@ lazy_static! {
 #[derive(Default)]
 pub struct SharedState {
     pub resizable: bool,
+    pub visible: bool,
```
msiglreith, line 272: this gets default initialized to `false` it seems

ArturKovacs: Yes you are right! After fixing this, it works for me! I just jumped to assuming that it was my system at fault after the fact that the other PR didn't work either.

ArturKovacs: I just pushed the fix; it seemed easiest to do so.

ArturKovacs:
Actually I gave it another thought and realized that it's probably best if the visibility is queried through a system API instead of trying to keep track of it ourselves.

There are basically two things that free CI services struggle with:
1, Cryptocurrency mining
2, My committing habits

ArturKovacs (approved):
Approving for macOS and X11

msiglreith (approved):
Thanks! Works fine on Windows as well 👍

PR #1951, opened 2021-05-27
Document how `main_thread_id` for Windows works

i509VCB:
- [ ] Tested on all platforms changed (N/A, no source level changes)

I was looking around winit one day and I found this method with no documentation. Upon looking further into the topic of getting the main thread id on Windows, it is more complex than the few LOC with no documentation.
Therefore I decided to add the documentation so a random onlooker is not confused when they see this method in the future.

I think I got the general idea documented, ironing out grammar can be done before merging :)

I think this one is addressed to you @msiglreith since you are listed as the maintainer.

### DIFF
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -267,17 +267,19 @@ impl<T> EventLoopWindowTarget<T> {
 
 /// Returns the id of the main thread.
 ///
-/// Windows has no real API to check if the current executing thread is the "main thread", unlike macOS.
+/// Windows has no real API to check if the current executing thread is the "main thread", unlike
+/// macOS.
 ///
-/// Windows will let us look up the current thread's id, but there's no API that lets us check what the
-/// id of the main thread is. We would somehow need to get the main thread's id before a developer
-/// could spin off any other threads inside of the main entrypoint in order to emulate the capabilities
-/// of other platforms.
+/// Windows will let us look up the current thread's id, but there's no API that lets us check what
+/// the id of the main thread is. We would somehow need to get the main thread's id before a
+/// developer could spin off any other threads inside of the main entrypoint in order to emulate the
+/// capabilities of other platforms.
 ///
 /// We can get the id of the main thread by using CRT initialization. CRT initialization can be used
-/// to setup global state within a program. The OS will call a list of function pointers which assign
-/// values to a static variable. To have get a hold of the main thread id, we need to place our
-/// function pointer inside of the `.CRT$XCU` section so it is called before the main entrypoint.
+/// to setup global state within a program. The OS will call a list of function pointers which
+/// assign values to a static variable. To have get a hold of the main thread id, we need to place
+/// our function pointer inside of the `.CRT$XCU` section so it is called before the main
+/// entrypoint.
 ///
 /// Full details of CRT initialization can be found here:
 /// https://docs.microsoft.com/en-us/cpp/c-runtime-library/crt-initialization?view=msvc-160
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -270,15 +270,14 @@ impl<T> EventLoopWindowTarget<T> {
 /// Windows has no real API to check if the current executing thread is the "main thread", unlike macOS.
 ///
 /// Windows will let us look up the current thread's id, but there's no API that lets us check what the
-/// id of the main thread is.
-/// We would somehow need to get the main thread's id before a developer could spin off any other
-/// threads inside of the main entrypoint in order to emulate the capabilities of other platforms.
+/// id of the main thread is. We would somehow need to get the main thread's id before a developer
+/// could spin off any other threads inside of the main entrypoint in order to emulate the capabilities
+/// of other platforms.
 ///
-/// We can get the id of the main thread by using CRT initialization.
-/// CRT initialization can be used to setup global state within a program.
-/// The OS will call a list of function pointers which assign a values to a static fields.
-/// To have get a hold of the main thread id, we need to place our function pointer inside
-/// of the `.CRT$XCU` section so it is called before the main entrypoint.
+/// We can get the id of the main thread by using CRT initialization. CRT initialization can be used
+/// to setup global state within a program. The OS will call a list of function pointers which assign
+/// values to a static fields. To have get a hold of the main thread id, we need to place our
+/// function pointer inside of the `.CRT$XCU` section so it is called before the main entrypoint.
 ///
 /// Full details of CRT initialization can be found here:
 /// https://docs.microsoft.com/en-us/cpp/c-runtime-library/crt-initialization?view=msvc-160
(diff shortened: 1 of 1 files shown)


msiglreith (approved):
Thanks a lot!

src/platform_impl/windows/event_loop.rs
```diff
@@ -265,10 +265,37 @@ impl<T> EventLoopWindowTarget<T> {
     }
 }
 
+/// Returns the id of the main thread.
+///
+/// Windows has no real API to check if the current executing thread is the "main thread" unlike macOS.
+/// Unix-like OSes have some APIs or semantics that exist regarding whether the main thread is in use or
+/// if the main thread id equals the pid like on Linux.
```
maroider, line 272: Could you clarify a bit here: is there a Unix API to check for the main thread, or is it guaranteed that "main thread id = pid"?

src/platform_impl/windows/event_loop.rs
```diff
@@ -265,10 +265,37 @@ impl<T> EventLoopWindowTarget<T> {
     }
 }
 
+/// Returns the id of the main thread.
+///
+/// Windows has no real API to check if the current executing thread is the "main thread" unlike macOS.
```
maroider, line 270:
```suggestion
/// Windows has no real API to check if the current executing thread is the "main thread", unlike macOS.
```

src/platform_impl/windows/event_loop.rs
```diff
@@ -265,10 +265,37 @@ impl<T> EventLoopWindowTarget<T> {
...
+/// Unix-like OSes have some APIs or semantics that exist regarding whether the main thread is in use or
+/// if the main thread id equals the pid like on Linux.
+///
+/// Windows will let us look up the current thread's id but this leaves us with the question of what
+/// is the main thread's id?
```
maroider, line 275:
```suggestion
/// Windows will let us look up the current thread's id, but there's no API that lets us check what the
/// id of the main thread is.
```

src/platform_impl/windows/event_loop.rs
```diff
@@ -265,10 +265,37 @@ impl<T> EventLoopWindowTarget<T> {
...
+/// We would somehow need to get the main thread's id before a developer could spin off any other
+/// threads inside of the main entrypoint in order to emulate the capabilities of other platforms.
+///
+/// We can get the id of the main thread by using CRT initialization.
+/// CRT initialization can be used to setup global state within a program by calling a list of
+/// function pointers which should assign a value to a static field.
+/// To have the field's value set to the main thread id, we need to place our function pointer inside
+/// of the section `.CRT$XCU` so it is called before the main entrypoint.
```
maroider, line 283:
```suggestion
/// To have get a hold of the main thread id, we need to place our function pointer inside
/// of the `.CRT$XCU` section so it is called before the main entrypoint.
```

I'm not sure what you mean by "To have the field's value set".

src/platform_impl/windows/event_loop.rs
```diff
@@ -265,10 +265,37 @@ impl<T> EventLoopWindowTarget<T> {
...
+/// if the main thread id equals the pid like on Linux.
+///
+/// Windows will let us look up the current thread's id but this leaves us with the question of what
+/// is the main thread's id?
+/// We would somehow need to get the main thread's id before a developer could spin off any other
+/// threads inside of the main entrypoint in order to emulate the capabilities of other platforms.
+///
+/// We can get the id of the main thread by using CRT initialization.
+/// CRT initialization can be used to setup global state within a program by calling a list of
+/// function pointers which should assign a value to a static field.
```
maroider, line 281: This makes it sound like we're the ones calling a list of function pointers.

src/platform_impl/windows/event_loop.rs
```diff
@@ -265,10 +265,37 @@ impl<T> EventLoopWindowTarget<T> {
...
+/// To have the field's value set to the main thread id, we need to place our function pointer inside
+/// of the section `.CRT$XCU` so it is called before the main entrypoint.
+///
+/// Full details of CRT initialization can be found here:
+/// https://docs.microsoft.com/en-us/cpp/c-runtime-library/crt-initialization?view=msvc-160
 fn main_thread_id() -> DWORD {
     static mut MAIN_THREAD_ID: DWORD = 0;
+
+    /// Function pointer used in CRT initialization section to set the above static field's value.
```
maroider, line 290: Does this doc comment actually work as a doc comment? As in: if you build the docs with `--document-private-items`, does this appear on `INIT_MAIN_THREAD_ID`?

maroider (changes requested):
You'll probably want to re-flow some (doc) comments as I didn't check the line widths.

src/platform_impl/windows/event_loop.rs
i509VCB: Decided to omit the Unix example since each unix has it's own way of checking for the main thread.

src/platform_impl/windows/event_loop.rs
i509VCB: I don't even see `INIT_MAIN_THREAD_ID` when running with that option you mentioned.

src/platform_impl/windows/event_loop.rs
i509VCB: OS calls the function pointers. New docs there should make it more clear.

i509VCB:
Regarding reflow of doc comments, I generally will do 120ish. Let me know if you want it at a strict 120 or if going a little over is fine.

src/platform_impl/windows/event_loop.rs
```diff
@@ -265,10 +265,34 @@ impl<T> EventLoopWindowTarget<T> {
...
+/// Windows has no real API to check if the current executing thread is the "main thread", unlike macOS.
+///
+/// Windows will let us look up the current thread's id, but there's no API that lets us check what the
+/// id of the main thread is. We would somehow need to get the main thread's id before a developer
+/// could spin off any other threads inside of the main entrypoint in order to emulate the capabilities
+/// of other platforms.
+///
+/// We can get the id of the main thread by using CRT initialization. CRT initialization can be used
+/// to setup global state within a program. The OS will call a list of function pointers which assign
+/// values to a static fields. To have get a hold of the main thread id, we need to place our
```
maroider, line 279:
```suggestion
/// values to a static variable. To have get a hold of the main thread id, we need to place our
```
not: The ID of the main thread is being assigned to a static variable, not a field of a struct.

maroider (changes requested):

maroider:
> Regarding reflow of doc comments, I generally will do 120ish. Let me know if you want it at a strict 120 or if going a little over is fine.

I'm not particularly great at sticking to line width limits, but when I do, I prefer being strict.

src/platform_impl/windows/event_loop.rs
maroider: Does it show as a doc comment if you use your IDE's on-hover hint?

i509VCB: yes 
Screenshot_20210528_112624

maroider: In that case, I'm fine with leaving it as-is.

maroider (approved):
LGTM

PR #1952, opened 2021-05-27
X11: Retry Interrupted event polls once

### LINKED ISSUE #1947: Panic with `ControlFlow::Wait`when computer goes to sleep on Linux/X11
Hello,

I encountered a panic within `winit` when putting my closing and re-opening the lid on my laptop (i.e. when sleep mode is entered) while executing the `winit` (version `0.24`+) application:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 4, kind: Interrupted, message: "Interrupted system call" }', /home/jfrimmel/.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.25.0/src/platform_impl/linux/x11/mod.rs:360:54
```
The platform is Linux with only the `x11` feature enabled. I haven't checked other platforms.

# Reproduction
```bash
$ cargo new winit-sleep-linux
$ cd winit-sleep-linux
$ echo 'winit = { version = "0.25", default-features = false, features=["x11"]}' >> Cargo.toml
$ nano src/main.rs # insert code below
$ cargo run # close the lid/enter sleep otherwise; wait a short time; open the lid/leave sleep mode
```
The content of `src/main.rs` is very simle and reduced, it does not even contain a window:
```rust
fn main() {
    let event_loop = winit::event_loop::EventLoop::new();
    event_loop.run(|_, _, control_flow| *control_flow = winit::event_loop::ControlFlow::Wait);
}
```
# Backtrace
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 4, kind: Interrupted, message: "Interrupted system call" }', /home/jfrimmel/.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.25.0/src/platform_impl/linux/x11/mod.rs:360:54
stack backtrace:
   0: rust_begin_unwind
             at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/std/src/panicking.rs:493:5
   1: core::panicking::panic_fmt
             at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/core/src/panicking.rs:92:14
   2: core::option::expect_none_failed
             at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/core/src/option.rs:1329:5
   3: core::result::Result<T,E>::unwrap
             at /home/jfrimmel/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/result.rs:1037:23
   4: winit::platform_impl::platform::x11::EventLoop<T>::run_return
             at /home/jfrimmel/.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.25.0/src/platform_impl/linux/x11/mod.rs:360:17
   5: winit::platform_impl::platform::x11::EventLoop<T>::run
             at /home/jfrimmel/.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.25.0/src/platform_impl/linux/x11/mod.rs:385:9
   6: winit::platform_impl::platform::EventLoop<T>::run
             at /home/jfrimmel/.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.25.0/src/platform_impl/linux/mod.rs:662:56
   7: winit::event_loop::EventLoop<T>::run
             at /home/jfrimmel/.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.25.0/src/event_loop.rs:154:9
   8: winit_sleep_linux::main
             at ./src/main.rs:3:5
   9: core::ops::function::FnOnce::call_once
             at /home/jfrimmel/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:227:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

jfrimmel:
As the description of `std::io::ErrorKind::Interrupted` states:
> Interrupted operations can typically be retried.

If there is a poll, which fails for being interrupted, it may succeed on a retry. Therefore this commit retries such an interrupted operation once. The background is, that entering sleep mode will cause interrupted events, which leads to panics on wake-up. This commit handles that case gracefully.

Fixes #1947.

- [x] Tested on all platforms changed (only X11 backend affected)
- [x] Created or updated an example program if it would help users understand this functionality (not relevant for this bug-fix)
- [x] Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented (none)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - On macOS, drop the event callback before exiting.
 - On Android, implement `Window::request_redraw`
+- On X11, prevent a panic on interrupted system calls.
 - **Breaking:** On Web, remove the `stdweb` backend.
 - Added `Window::focus_window`to bring the window to the front and set input focus.
 - On Wayland and X11, implement `is_maximized` method on `Window`.
--- a/src/platform_impl/linux/x11/mod.rs
+++ b/src/platform_impl/linux/x11/mod.rs
@@ -357,7 +357,14 @@ impl<T: 'static> EventLoop<T> {
             // If the XConnection already contains buffered events, we don't
             // need to wait for data on the socket.
             if !self.event_processor.poll() {
-                self.poll.poll(&mut events, timeout).unwrap();
+                self.poll
+                    .poll(&mut events, timeout)
+                    .or_else(|e| match e.kind() {
+                        // retry the operation once, e.g. for sleeping
+                        std::io::ErrorKind::Interrupted => self.poll.poll(&mut events, timeout),
+                        _ => Err(e),
+                    })
+                    .unwrap();
                 events.clear();
             }
 


jfrimmel:
I don't know, if such a bugfix should be noted in the changelog or documentation. If the answer is yes, I'd be happy to amend this PR.

ArturKovacs:
Thanks for the PR! Yeah add a changelog entry, but this seems to be good otherwise.

ArturKovacs:
@francesca64, two web related tests that don't exist anymore are still being required to pass; this prevents PRs from getting merged. The stable test that replaced these is
```
stable, wasm32-unknown-unknown, windows-latest, web
```

If I'm not mistaken you can change this under Settings / Branches / Rules / "Require status checks to pass before merging"

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 - **Breaking:** On Web, remove the `stdweb` backend.
 - Added `Window::focus_window`to bring the window to the front and set input focus.
+- Prevent a panic on interrupted system calls on the `x11` backend.
```
maroider, line 5:
```suggestion
- On X11, prevent a panic on interrupted system calls.
```
Could you move this to the top of the changelog?

maroider (changes requested):

CHANGELOG.md
jfrimmel: Done 😉

maroider (approved):
LGTM

ArturKovacs:
I cancelled the most recent workflow, because it will stall on the non-existing checks anyways.

jfrimmel:
Hi, is there anything missing (except for the CI run)? Or should only the CI be re-started?

ArturKovacs:
The PR is all good. We are just waiting for Francesca to update the repo settings, and then we can re-start the ci checks.

maroider:
We're waiting on @francesca64 to change some branch protection rules. We changed a couple CI targets recently, and the branch protection rules seem to be incompatible now.

francesca64:
I've updated the branch rules! You should be good to go.

maroider:
You may want to rebase @jfrimmel, as `CHANGELOG.md` has conflicts now.

jfrimmel:
I've rebased onto the current `master` (i.e. there are no actual code changes)

hcsch:
Looping until the polling succeeds might be a better choice. I recently had an issue with the polling code as it is in v0.25 (and the current master) where debugging a multi-threaded application causes the code to panic due to an interrupt. This happens because the debugger stops all threads when a breakpoint is hit, even outside of the thread the event loop is running in. This seems to reliably interrupts syscalls.

I tried my MCVE from #1972 with your changed code that re-polls once in case of an interrupt and was still able to get a panic due to an unwrapped interrupt error. Admittedly it was a lot harder and probably would happen rarely if ever in the wild, but I'd say it is better to rule it out entirely.

Something like the following should work (I checked that with my example and couldn't get it to panic due to an interrupt, which makes sense):
```rust
if !self.event_processor.poll() {
    while let Err(e) = self.poll.poll(&mut events, timeout) {
        match e.kind() {
            std::io::ErrorKind::Interrupted => (), // retry polling
            _ => Err(e).expect("polling failed irrecoverably"),
        }
    }
    events.clear();
}
```

ArturKovacs:
@hcsch convinced me that it would indeed be better to poll until it returns with something else than "Interrupted". I would also add a debug log message so that if somehow an application gets stuck in an endless polling loop, the developer would know why that happens. Like so:

```rust
if !self.event_processor.poll() {
    while let Err(e) = self.poll.poll(&mut events, timeout) {
        match e.kind() {
            // retry polling
            std::io::ErrorKind::Interrupted => debug!("Poll returned 'Interrupted'. Retrying."),
            _ => Err(e).expect("polling failed irrecoverably"),
        }
    }
    events.clear();
}
```

hcsch:
It seems that with e9d5b2007a5b7cb70a514dad6812fbc22f2f619c the code won't panic anymore and will retry in the next iteration of the main loop (unlike with what was discussed here, where it would retry immediately). This issue should therefore maybe also be closed (as #1972 was with this commit).

PR #1954, opened 2021-05-29
Drop the event callback before exiting on macOS

### LINKED ISSUE #1948: Drop callback on exit on macOS
As far as I can tell the event callback, and hence nothing moved to it gets dropped on application exit. It should be dropped, just like on Windows.

ArturKovacs:
Closes #1948 

### DIFF
--- a/src/platform_impl/macos/event_loop.rs
+++ b/src/platform_impl/macos/event_loop.rs
@@ -141,7 +141,6 @@ impl<T> EventLoop<T> {
         F: 'static + FnMut(Event<'_, T>, &RootWindowTarget<T>, &mut ControlFlow),
     {
         self.run_return(callback);
-        drop(self._callback.take());
         process::exit(0);
     }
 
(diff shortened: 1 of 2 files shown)


src/platform_impl/macos/event_loop.rs
```diff
@@ -141,6 +141,8 @@ impl<T> EventLoop<T> {
         F: 'static + FnMut(Event<'_, T>, &RootWindowTarget<T>, &mut ControlFlow),
     {
         self.run_return(callback);
+        // Drop the callback before exiting
+        self._callback = None;
```
maroider, line 145:
```suggestion
        drop(self._callback.take());
```
This should invalidate the need for a comment

src/platform_impl/macos/event_loop.rs
```diff
@@ -141,6 +141,7 @@ impl<T> EventLoop<T> {
         F: 'static + FnMut(Event<'_, T>, &RootWindowTarget<T>, &mut ControlFlow),
     {
         self.run_return(callback);
+        drop(self._callback.take());
```
maroider, line 144: Upon further though, I think you should move this to the end of `run_return`, as the closure passed to `run_return` should also be dropped before returning.

ArturKovacs: Donezo

maroider (approved):

madsmtm (approved):

PR #1957, opened 2021-06-05
Update `raw-window-handle` to `0.4.1`

madsmtm:
Ready for when https://github.com/rust-windowing/raw-window-handle/pull/70 gets merged, and a stable version published.

### DIFF
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -595,9 +595,9 @@ impl Window {
         } else {
             panic!("Cannot get the native window, it's null and will always be null before Event::Resumed and after Event::Suspended. Make sure you only call this function between those events.");
         };
-        let mut handle = raw_window_handle::android::AndroidHandle::empty();
+        let mut handle = raw_window_handle::AndroidNdkHandle::empty();
         handle.a_native_window = a_native_window;
-        raw_window_handle::RawWindowHandle::Android(handle)
+        raw_window_handle::RawWindowHandle::AndroidNdk(handle)
     }
 
     pub fn config(&self) -> Configuration {
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -12,6 +12,7 @@ use ndk::{
     looper::{ForeignLooper, Poll, ThreadLooper},
 };
 use ndk_glue::{Event, Rect};
+use raw_window_handle::{AndroidNdkHandle, RawWindowHandle};
 use std::{
     collections::VecDeque,
     sync::{Arc, Mutex, RwLock},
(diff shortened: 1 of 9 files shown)


src/platform_impl/android/mod.rs
```diff
@@ -587,15 +588,14 @@ impl Window {
...
     }
 
-    pub fn raw_window_handle(&self) -> raw_window_handle::RawWindowHandle {
-        let a_native_window = if let Some(native_window) = ndk_glue::native_window().as_ref() {
+    pub fn raw_window_handle(&self) -> RawWindowHandle {
+        let mut handle = AndroidNDKHandle::empty();
+        handle.a_native_window = if let Some(native_window) = ndk_glue::native_window().as_ref() {
             unsafe { native_window.ptr().as_mut() as *mut _ as *mut _ }
         } else {
             panic!("Cannot get the native window, it's null and will always be null before Event::Resumed and after Event::Suspended. Make sure you only call this function between those events.");
```
msiglreith, line 596: looks like we will never get rid of this one, sad

madsmtm:
This should be ready now, but I don't think it should be included in the `0.26.0` release, since that will prevent `winit` `0.26.0` from being used with crates that are using `raw-window-handle` version `0.3`. We probably need to do collaborate on an ecosystem-wide rollout

madsmtm:
See https://github.com/rust-windowing/raw-window-handle/pull/74 for an upgrade plan

kchibisov:
@madsmtm Could you provide an update on this PR? I can't keep an eye on all the bits in ecosystem right now, and I don't want to hold `v0.26.0` release of winit.

madsmtm:
This is currently waiting on `raw-window-handle` `v0.3.4` being released, I suspect it will take a day, maybe three. See https://github.com/rust-windowing/raw-window-handle/pull/79.

If it takes longer due to unforeseen circumstances, I'm fine with just releasing `winit` `v0.26.0` as-is.

a1phyr:
`raw-window-handle` `v0.3.4` has just been released, so this can go forward.

madsmtm:
Yup, this is ready to be merged.

I've tested this with a small personal project using [`bevy`](https://github.com/bevyengine/bevy), they use the `raw-window-handle` traits from `v0.3`, and it works fine with this PR.

Again, just for clarification, because of the blanket `impl<T: new::HasRawWindowHandle> old::HasRawWindowHandle for T` in `v0.3.4`, by implementing the new trait we still implement the old one. Hence this is not a breaking change! (Users might have to do a `cargo update` though, not entirely sure about how the cargo caching works).

src/platform_impl/android/mod.rs
```diff
@@ -589,15 +590,14 @@ impl Window {
...
-    pub fn raw_window_handle(&self) -> raw_window_handle::RawWindowHandle {
-        let a_native_window = if let Some(native_window) = ndk_glue::native_window().as_ref() {
-            unsafe { native_window.ptr().as_mut() as *mut _ as *mut _ }
+    pub fn raw_window_handle(&self) -> RawWindowHandle {
+        let mut handle = AndroidNdkHandle::empty();
+        if let Some(native_window) = ndk_glue::native_window().as_ref() {
+            handle.a_native_window = unsafe { native_window.ptr().as_mut() as *mut _ as *mut _ }
         } else {
-            panic!("Cannot get the native window, it's null and will always be null before Event::Resumed and after Event::Suspended. Make sure you only call this function between those events.");
+            log::warn!("Cannot get the native window while the application is suspended (between Event::Suspended and Event::Resumed).");
```
msiglreith, line 598: Sorry I think this was not clear from my original comment. `panic!` was the consensus on an earlier discussion allowing to return a null-ptr. At a certain point in `raw-window-handle` the handle was behind a `Option` with the intention to clearer annotate that it may be null, but this got reverted again.

The conclusion here is that this should still be panic if the handle is not available.

madsmtm: I agree with what `@Lokathor` said here: https://github.com/rust-windowing/raw-window-handle/pull/70#issuecomment-855303860; we should just return a NULL pointer, downstream code should check whether `a_native_window` is NULL or not.

Remember, the NULL check has to happen on all platforms! E.g. a `HasRawWindowHandle` provider might legally return `RawWindowHandle::AppKit(AppKitHandle::empty())`, and consumer libraries _must_ be able to handle that (even if they handle it by just panicking)!

madsmtm: Of course, you're our Android maintainer, so if you still disagree, I'll remove the commit (and maybe make a separate PR for it :wink:)

msiglreith: I would prefer to keep the change out for this PR. I'm personally not opposed to it tho (as I tried to change it in the past as well heh - latest conclusion: https://github.com/rust-windowing/winit/pull/1624#issuecomment-673097014)

madsmtm: I hadn't seen the linked PR, sorry for stirring it up again.

I've removed the commit, this PR should be ready for real now :wink:

msiglreith (approved):

PR #1961, opened 2021-06-12
Retry XGrabPointer a couple of times (up to 500ms) if it failes

Visse:
If you try to grab the cursor right after you created the window you get the error:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os(OsError { line: 1266, file: "/home/visse/repos/winit/src/platform_impl/linux/x11/window.rs", error: XMisc("Cursor could not be grabbed: grab location not viewable") })', src/main.rs:11:34
```

Minimal repro:
```rust
use winit::{event_loop::EventLoop, window::WindowBuilder};

fn main() {

        .build(&event_loop)
        .unwrap();

    window.set_cursor_grab(true).unwrap();
}
```

Checking the sdl implementation, it looks like they are solving it by retrying it a couple of times:
https://github.com/libsdl-org/SDL/blob/e65a6583201ee1a9a3bc3064655b186fc16ef719/src/video/x11/SDL_x11window.c#L1611-L1619

This PR simply implements the same scheme and retries XGrabPointer up to 10 times with a slight delay between each, for a total delay up to 500ms.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - **Breaking:** On Web, remove the `stdweb` backend.
 - Added `Window::focus_window`to bring the window to the front and set input focus.
 - On Wayland and X11, implement `is_maximized` method on `Window`.
+- On X11 retry cursor grab on failure (up to 10 times / 500ms)
 
 # 0.25.0 (2021-05-15)
 
--- a/src/platform_impl/linux/x11/window.rs
+++ b/src/platform_impl/linux/x11/window.rs
@@ -1225,31 +1225,41 @@ impl UnownedWindow {
             (self.xconn.xlib.XUngrabPointer)(self.xconn.display, ffi::CurrentTime);
         }
         let result = if grab {
-            let result = unsafe {
-                (self.xconn.xlib.XGrabPointer)(
-                    self.xconn.display,
-                    self.xwindow,
-                    ffi::True,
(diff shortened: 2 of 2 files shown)


Visse:
Is there something I can do to speed up merging? Or is something missing? I'm not too familiar with the process.

ArturKovacs (changes requested):
This is awesome!

Depending on the application, this delay may be significant, so I recommend adding a warning about this to the documentation of the `set_cursor_grab` function. If you have any better idea how we can let the developers know about this, I'm open to other suggestions.

dhardy:
Would a viable alternative be to wait until the first cursor event is received and *then* request the cursor grab? It shouldn't make any difference to the user except that if the cursor isn't initially over the window it won't be captured until it is (which seems fine?). That way you're never blocking.

parasyte:
Related: #2080

rib:
The SDL solution to this looks like a hack really. The X11 backend could be updated to track the 'mapped' state of the window to account for the asynchronous nature of creating X windows. The code for cursor grabs would then need to have a "pending" state that's used in case the window is not currently mapped. Whenever winit gets notified of a window being mapped then it would check for any pending grabs an then apply them, which should then succeed.

notgull:
Once #3122 is merged, I can update this PR to wait for events using the new userspace event queue introduced in that PR.

PR #1963, opened 2021-06-12
Emit ScaleFactorChanged event on monitor reconnect

### LINKED ISSUE #1962: X11 not sending ScaleFactorChange on monitor reconnect
This has been reported to Alacritty in https://github.com/alacritty/alacritty/issues/5213.

It seems like when all available monitors get disconnected, there's a ScaleFactorChanged event with a factor of 1, which isn't too unexpected when there's no monitor to determine the DPI appropriately.

The issue however is that when the monitor is connected again, there does not seem to be another ScaleFactorChanged event that informs the client about the new scale factor.

oxalica:
Fixes #1962 
When disconnect the only monitor, scale factor is reset to 1.0. We need
to set it back when the monitor is reconnected.

We previously assume current window must be on an existing monitor, but
that's not true in case of reconnecting the only one monitor.

  - Tested on X11, KDE, via alacritty. (Fixes https://github.com/alacritty/alacritty/issues/5213)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@
 - **Breaking:** On Wayland, Theme trait and its support types are dropped.
 - On Wayland, bump `smithay-client-toolkit` to 0.15.
 - On Wayland, implement `request_user_attention` with `xdg_activation_v1`.
-
+- On X11, emit missing `WindowEvent::ScaleFactorChanged` when the only monitor gets reconnected.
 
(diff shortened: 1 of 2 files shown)


ArturKovacs (approved):
Great! Just fix the conflict with the changelog, and I'll merge this.

oxalica:
Rebased.

ArturKovacs (approved):

oxalica:
The CI failure is unrelated.

PR #1976, opened 2021-07-12
Remove libc dependency on non-linux platforms

maroider:
- ~~Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior~~
- ~~Created or updated an example program if it would help users understand this functionality~~
- ~~Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented~~

---

We don't really need to bring in all of `libc` when all we're using is `libc::c_void`, which is just an alias for `core::ffi::c_void` after `libc = "0.2.0"` anyway.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,6 @@ wayland = ["wayland-client", "sctk"]
 [dependencies]
 instant = { version = "0.1", features = ["wasm-bindgen"] }
 lazy_static = "1"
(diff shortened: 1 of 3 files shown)


ArturKovacs (approved):
Awesoe possum!

madsmtm (approved):

msiglreith (approved):

PR #1977, opened 2021-07-12
Prevent ghost window from showing up on taskbar

### LINKED ISSUE #1265: Winit creating invisible Windows on Windows
It seems like there's an issue where winit is spawning invisible windows on Windows after some period of time. They seem to come from the `Winit Thread Event Target`. It seems like in the worst case this can actually lead to the process getting killed as a result of it.

A crash seems to cause the following error, however since this is likely a result, not the cause, I'd treat it with some caution.

```
panicked at 'PostMessage failed; is the messages queue full?', C:\Users\travis.cargo\registry\src\github.com-1ecc6299db9ec823\winit-0.19.1\src\platform\windows\events_loop.rs:302:13
```

This has been reported to Alacritty in https://github.com/jwilm/alacritty/issues/2656, however it has not been confirmed with the winit rework.

@leandrosansilva if you could try the latest Alacritty master and report back here if that also is broken, that would be very helpful.

Osspial:
Closes #1265. The only way to reproduce this I've found is to wait around for several hours, and after I added this code I haven't seen the issue crop up at all, so I'm fairly sure the underlying bug is fixed. 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - **Breaking:** On Web, remove the `stdweb` backend.
 - Added `Window::focus_window`to bring the window to the front and set input focus.
 - On Wayland and X11, implement `is_maximized` method on `Window`.
+- On Windows, prevent ghost window from showing up in taskbar after several hours of use.
 
 # 0.25.0 (2021-05-15)
 
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -599,6 +599,15 @@ fn create_event_target_window() -> HWND {
             winuser::WS_EX_NOACTIVATE
                 | winuser::WS_EX_TRANSPARENT
                 | winuser::WS_EX_LAYERED
+                // WS_EX_TOOLWINDOW prevents this window from ever showing up in the taskbar, which
+                // we want to avoid. If you remove this style, this window won't show up in the
+                // taskbar *initially*, but will eventually magically appear in the taskbar.
+                //
+                // Now, I wasn't able to figure out what *causes* the window to eventually show up
+                // in the taskbar, since it does so without warning, after several hours, with no
+                // obvious code-visible state changes to this window. I'd guess that some win32
+                // state change gets triggered at some point, but I'd need to be able to debug the
+                // windows shell directly to figure that out and I can't do that.
                 | winuser::WS_EX_TOOLWINDOW,
             THREAD_EVENT_TARGET_WINDOW_CLASS.as_ptr(),
             ptr::null_mut(),
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -596,7 +596,10 @@ lazy_static! {
 fn create_event_target_window() -> HWND {
     unsafe {
         let window = winuser::CreateWindowExW(
-            winuser::WS_EX_NOACTIVATE | winuser::WS_EX_TRANSPARENT | winuser::WS_EX_LAYERED | winuser::WS_EX_TOOLWINDOW,
+            winuser::WS_EX_NOACTIVATE
+                | winuser::WS_EX_TRANSPARENT
+                | winuser::WS_EX_LAYERED
+                | winuser::WS_EX_TOOLWINDOW,
             THREAD_EVENT_TARGET_WINDOW_CLASS.as_ptr(),
             ptr::null_mut(),
             0,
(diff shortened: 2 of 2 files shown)


maroider:
Funny that you'd request a review right now 😄. I just started checking out your changes.

I can't say I've actually encountered the bug described in #1265, but I'm going to leave a window from `master` and a window from this PR open for a couple of hours and see if anything interesting happens.

maroider:
I let it sit overnight, but I didn't encounter the bug at all.

maroider:
I'd really prefer to get an independent confirmation that this PR is indeed a fix.
And to that end: cc @chris-morgan

chris-morgan:
Sorry, I can’t confirm it one way or the other as I’m on a new laptop with Arch Linux now, and currently in a different country from the laptop and display where I knew some tricks that would *normally* reproduce it before long (apart from the whole Zoom screen sharing thing that is almost certainly connected), which involved related to adding and removing monitors at different pixel densities, sleeping, waking up, *&c.* I never found just waiting to trigger anything.

But the change outlined here sounds about right, and if it solves the Zoom thing, I think it’ll *probably* solve the other as well. I say go for it, and reassess it later if it turns out not to have solved all of the problem. The change is unlikely to be in any way *harmful*.

maroider (approved):
I think we can go ahead with this PR as @chris-morgan suggested, although there are some conflicts that need to be resolved now.

maroider:
If any among the "windows testers gang" can repro the bug and the fix, then that would be lovely :heart:, but perhaps not strictly necessary.

cc @msiglreith @ryanisaacg @notriddle @0xpr03 @arthrp @FliegendeWurst @chelmich

0xpr03:
I'll run fix & repro over night, let's see

0xpr03:
Err wait, what exactly should I run to verify this ? Alacritty old for the crash maybe and what for the fix ?

msiglreith:
One way to reproduce it consistently for me was to manually restart the Windows Explorer process 😄
With the proposed PR this case at least is covered!

0xpr03:
grafik
Yep, can reproduce it this way.

1. kill explorer.exe
2. start a new instance

And as you can see the fixed version doesn't spawn a new window

madsmtm:
Well, we now have more reproduction info so maybe @Osspial would update the code comment to include this?

0xpr03:
Update: Simply letting it run over night, with one monitor shutting off over night didn't repro this. So the explorer kill-switch seems to be the only reliable way of reproducing this, though without a crash, which makes it pretty hard to test for.

src/platform_impl/windows/event_loop.rs
```diff
@@ -596,7 +596,19 @@ lazy_static! {
...
+                | winuser::WS_EX_LAYERED
+                // WS_EX_TOOLWINDOW prevents this window from ever showing up in the taskbar, which
+                // we want to avoid. If you remove this style, this window won't show up in the
+                // taskbar *initially*, but will eventually magically appear in the taskbar.
+                //
+                // Now, I wasn't able to figure out what *causes* the window to eventually show up
+                // in the taskbar, since it does so without warning, after several hours, with no
+                // obvious code-visible state changes to this window. I'd guess that some win32
+                // state change gets triggered at some point, but I'd need to be able to debug the
+                // windows shell directly to figure that out and I can't do that.
```
maroider, line 610:
```suggestion
                // WS_EX_TOOLWINDOW prevents this window from ever showing up in the taskbar, which
                // we want to avoid. If you remove this style, this window won't show up in the
                // taskbar *initially*, but it can show up at some later point. This can sometimes
                // happen on its own after several hours have passed, although this has proven
                // difficult to reproduce. Alternatively, it can be manually triggered by killing
                // `explorer.exe` and then starting the process back up.
                // It is unclear why the bug is triggered by waiting for several hours.
```

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - **Breaking:** On Web, remove the `stdweb` backend.
 - Added `Window::focus_window`to bring the window to the front and set input focus.
 - On Wayland and X11, implement `is_maximized` method on `Window`.
+- On Windows, prevent ghost window from showing up in taskbar after several hours of use.
```
maroider, line 8:
```suggestion
- On Windows, prevent ghost window from showing up in the taskbar after either several hours of use or restarting `explorer.exe`.
```

maroider (changes requested):
Restarting `explorer.exe` triggers the bug for me, and the fix also works for me.

msiglreith (approved):
Thanks for the work I think this is good to go. I applied the suggestions from @maroider 
Also thanks for the tests and reviews!

PR #1979, opened 2021-07-27
Add composition event on macOS

komi1230:

This change is to enable to use IME.

Before:

https://user-images.githubusercontent.com/27967726/127104794-4469f128-565b-4b56-a5a5-5a327f89a5da.mov

After:

https://user-images.githubusercontent.com/27967726/127105174-c3b703be-68f8-4354-830c-05f8e178ce27.mov

### DIFF
--- a/src/platform_impl/macos/view.rs
+++ b/src/platform_impl/macos/view.rs
@@ -29,7 +29,7 @@ use crate::{
             scancode_to_keycode, EventWrapper,
         },
         ffi::*,
-        util::{self, IdRef},
+        util::{self, ns_string_char_count, IdRef},
         window::get_window_id,
         DEVICE_ID,
     },
@@ -59,7 +59,7 @@ pub(super) struct ViewState {
     tracking_rect: Option<NSInteger>,
     is_ime_activated: bool,
     is_preediting: bool,
-    marked_text: String,
+    marked_text: id,
 }
 
 impl ViewState {
@@ -71,6 +71,8 @@ impl ViewState {
 pub fn new_view(ns_window: id) -> (IdRef, Weak<Mutex<CursorState>>) {
     let cursor_state = Default::default();
     let cursor_access = Arc::downgrade(&cursor_state);
+    let marked_text =
+        unsafe { <id as NSMutableAttributedString>::init(NSMutableAttributedString::alloc(nil)) };
     let state = ViewState {
         ns_window,
         cursor_state,
@@ -80,7 +82,7 @@ pub fn new_view(ns_window: id) -> (IdRef, Weak<Mutex<CursorState>>) {
         tracking_rect: None,
         is_ime_activated: false,
         is_preediting: false,
-        marked_text: String::new(),
+        marked_text,
     };
     unsafe {
         // This is free'd in `dealloc`
@@ -267,7 +269,6 @@ lazy_static! {
             accepts_first_mouse as extern "C" fn(&Object, Sel, id) -> BOOL,
         );
         decl.add_ivar::<*mut c_void>("winitState");
-        decl.add_ivar::<id>("markedText");
         let protocol = Protocol::get("NSTextInputClient").unwrap();
         decl.add_protocol(&protocol);
         ViewClass(decl.register())
@@ -277,9 +278,9 @@ lazy_static! {
 extern "C" fn dealloc(this: &Object, _sel: Sel) {
     unsafe {
         let state: *mut c_void = *this.get_ivar("winitState");
-        let marked_text: id = *this.get_ivar("markedText");
-        let _: () = msg_send![marked_text, release];
-        Box::from_raw(state as *mut ViewState);
+        let state = state as *mut ViewState;
+        let _: () = msg_send![(*state).marked_text, release];
+        Box::from_raw(state);
     }
 }
 
@@ -288,9 +289,6 @@ extern "C" fn init_with_winit(this: &Object, _sel: Sel, state: *mut c_void) -> i
         let this: id = msg_send![this, init];
         if this != nil {
             (*this).set_ivar("winitState", state);
-            let marked_text =
-                <id as NSMutableAttributedString>::init(NSMutableAttributedString::alloc(nil));
-            (*this).set_ivar("markedText", marked_text);
             let _: () = msg_send![this, setPostsFrameChangedNotifications: YES];
 
             let notification_center: &Object =
@@ -397,17 +395,20 @@ extern "C" fn reset_cursor_rects(this: &Object, _sel: Sel) {
 extern "C" fn has_marked_text(this: &Object, _sel: Sel) -> BOOL {
     unsafe {
         trace!("Triggered `hasMarkedText`");
-        let marked_text: id = *this.get_ivar("markedText");
+        let state_ptr: *mut c_void = *this.get_ivar("winitState");
+        let state = &mut *(state_ptr as *mut ViewState);
+        let retval = (state.marked_text.length() > 0) as BOOL;
         trace!("Completed `hasMarkedText`");
-        (marked_text.length() > 0) as BOOL
+        retval
     }
 }
 
 extern "C" fn marked_range(this: &Object, _sel: Sel) -> NSRange {
     unsafe {
         trace!("Triggered `markedRange`");
-        let marked_text: id = *this.get_ivar("markedText");
-        let length = marked_text.length();
+        let state_ptr: *mut c_void = *this.get_ivar("winitState");
+        let state = &mut *(state_ptr as *mut ViewState);
+        let length = state.marked_text.length();
         trace!("Completed `markedRange`");
         if length > 0 {
             NSRange::new(0, length - 1)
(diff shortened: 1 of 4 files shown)


komi1230:
hmm....
This change has no effect on Android.
I'm not sure why CI fails.

maroider:
I think we might need to update our Android and WASM CI, so you don't need to do anything yourself (unless you'd like to fix our CI for us :D).

madsmtm:
Is it possible for you to provide a minimal example, so that we can reproduce without having to build Alacritty?

Additionally, maybe you could also provide an explanation of why this is necessary / why it works, because that is not immediately obvious to me ;)

komi1230:
Thank you for fixing CI !

And sorry for less explanation.
This PR is related to https://github.com/rust-windowing/winit/issues/1497.

It seems Winit has backend and frontend.
Backend catches events from OS (mouse or keyboard, for example) and passes over them to frontend, and also frontend catches events from backend and arranges them into common structs or types.

Backend for each OS corresponds to `platform_impl` dir.

So far, in macOS, keyboard event didn't work fine for Japanese language.
macOS has three types of keyboard event: `insertText`, `setMarkedText`, `doCommandBySelector`.
`insertText` is for English alphabet, this is easy, and `doCommandBySelector` is for keyboard shortcuts, not inserting some character (Ctrl-c, for example).
And `setMarkedText` is for other languages.
This is really tough point of IME...

This is a topic of linguistics.
For example, Japanese language has some formats: Kanji, Hiragana and Katakana, (maybe you know 😋 )
When we input Kanji, we have to start from Hiragana by typing English alphabet at first and convert it into Kanji, like `sakura` into `さくら` into `桜`.
In the step of converting Hiragana into Kanji, there is a challenge.
Hiragana can be pronounced in only one way, but a couple of Kanji characters have the same pronounce, as `さくら` can be `桜` or `佐倉`.
So when we input Kanji, we have to select which Kanji to input.

Let's get back to a topic of IME API.
In selecting Kanji, characters to insert are still not decided. In macOS, this undecided character is called `markedText`. 
When we have decided which Kanji to input, the selected character is handled in `insertText` from `setMarkedText`.
So inputting Kanji is inserted by `insertText`.
But this is a problem. We don't know which Kanji we are selecting.
So far, `insertText` was the only function to insert characters but `setMarkedText` should be another function to display undecided character.

In this PR, I implemented this architecture by adding event-push part to `setMarkedText`.
So you can find which Kanji we are selecting is visible and decided Kanji is properly inserted and displayed.

Thank you for reading long description.

Ref:
- [macOS's keyboard input architecture](https://developer.apple.com/library/archive/documentation/TextFonts/Conceptual/CocoaTextArchitecture/TextEditing/TextEditing.html#//apple_ref/doc/uid/TP40009459-CH3-SW3)

komi1230:
I'm sorry I don't have a minimal example for this.
But you can reproduce this fixed Alacritty in this repo.
https://github.com/komi1230/tmp

ArturKovacs:
Correct me if I'm wrong but it seems that this implementation does the following:
- Send the currently selected IME text to the application by sending a `ReceivedCharacter` for each character in the text.
- When the selecteed IME text changes, winit sends as many `DELETE` characters as long the previously selected text was, expecting the application to remove all those characters from the screen. Then winit sends the characters of the newly selected text again.

I have to admit, this is a kinda smart way of dealing with IME without changing winit's API. I would actually be willing to merge this, considering that the discussion at #1497 got stuck almost a year ago. We can add a nice IME API later.

With that said, I haven't tested this PR so I'm going to give a proper review later, I think, but in the meantime @madsmtm feel free to review this.

ArturKovacs:
Here's a minimal test application:

<details>

<summary>
Click for the code
</summary>

```rust
use log::LevelFilter;
use winit::{
    event::{Event, WindowEvent},
    window::WindowBuilder,
};

fn main() {
    SimpleLogger::new().with_level(LevelFilter::Debug).init().unwrap();

        .build(&event_loop)
        .unwrap();
    
        match event {
            Event::WindowEvent {
                window_id,
            Event::WindowEvent {
                event: WindowEvent::ReceivedCharacter(ch),
                ..
            } => {
                println!("recv ch: {:?}", ch);
            }
            Event::MainEventsCleared => {
                window.request_redraw();
            }
            _ => (),
        }
    });
}

```

</details>

src/platform_impl/macos/view.rs
```diff
@@ -257,8 +260,16 @@ lazy_static! {
             sel!(acceptsFirstMouse:),
             accepts_first_mouse as extern "C" fn(&Object, Sel, id) -> BOOL,
         );
+        decl.add_method(
+            sel!(clearMarkedText),
+            clear_marked_text as extern "C" fn(&mut Object, Sel),
+        );
```
ArturKovacs, line 266: As far as I can tell, it would be simpler if `clear_marked_text` would *not* be a member function. It could instead be a regular rust function, in which case you wouldn't need to add it as a method here, and he function could be defined like this:
```rust
fn clear_marked_text(this: &mut Object) {
    ...
}
```

src/platform_impl/macos/view.rs
```diff
@@ -433,22 +450,57 @@ extern "C" fn set_marked_text(
...
     trace!("Completed `setMarkedText`");
 }
 
 extern "C" fn unmark_text(this: &Object, _sel: Sel) {
     trace!("Triggered `unmarkText`");
     unsafe {
-        let marked_text: id = *this.get_ivar("markedText");
-        let mutable_string = marked_text.mutableString();
-        let _: () = msg_send![mutable_string, setString:""];
+        let _: () = msg_send![this, clearMarkedText];
```
ArturKovacs, line 489: If `clearMarkedText` is not a member, you could call it by `clear_marked_text(this)`

src/platform_impl/macos/view.rs
```diff
@@ -686,6 +768,23 @@ extern "C" fn key_down(this: &Object, _sel: Sel, event: id) {
...
+            if is_ime_activated && is_arrow_or_space_key(virtual_keycode.unwrap()) {
+                (*this).set_ivar("isPreediting", true);
+                let marked_text_ref: &mut id = this.get_mut_ivar("markedText");
+                let composed_string = marked_text_ref.clone().string();
+                let slice = slice::from_raw_parts(
+                    composed_string.UTF8String() as *const c_uchar,
+                    composed_string.len(),
+                );
+                let composed_string = str::from_utf8_unchecked(slice);
+                delete_marked_text(state, composed_string.chars().count() as i32);
```
ArturKovacs, line 782: What is the reasoning behind deleting the IME text here? As far as I understand, we always delete the text in `set_marked_text` before sending the new one.

src/platform_impl/macos/view.rs
```diff
@@ -686,6 +768,23 @@ extern "C" fn key_down(this: &Object, _sel: Sel, event: id) {
...
+            }
+
+            // Clear them here so that we can know whether they have changed afterwards.
+            let _: () = msg_send![this, clearMarkedText];
+
```
ArturKovacs, line 787: I don't understand the comment here. Could you describe this differently? (Maybe this is related to the fact that `interpretKeyEvents` below, may invoke `set_marked_text` or `insert_text`. Regardless it's not clear what this comment means)

ArturKovacs (changes requested):

src/platform_impl/macos/view.rs
komi1230: That sounds good !

src/platform_impl/macos/view.rs
```diff
@@ -257,8 +260,16 @@ lazy_static! {
...
+        decl.add_method(
+            sel!(clearMarkedText),
+            clear_marked_text as extern "C" fn(&mut Object, Sel),
+        );
         decl.add_ivar::<*mut c_void>("winitState");
         decl.add_ivar::<id>("markedText");
+        decl.add_ivar::<bool>("isIMEActivated");
+        decl.add_ivar::<bool>("isPreediting");
+        decl.add_ivar::<i32>("currentCursorPosition");
+        decl.add_ivar::<i32>("previousCursorPosition");
```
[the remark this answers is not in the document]
komi1230: By storing state information in ViewState, it got simpler.
Thanks !

src/platform_impl/macos/view.rs
komi1230: That's true.
I misunderstood some points while debugging....

komi1230:
@ArturKovacs Thank you for reviewing ! I fixed parts you pointed out. Can you check again ?

src/platform_impl/macos/view.rs
```diff
@@ -637,7 +699,7 @@ fn update_potentially_stale_modifiers(state: &mut ViewState, event: id) {
     }
 }
 
-extern "C" fn key_down(this: &Object, _sel: Sel, event: id) {
+extern "C" fn key_down(this: &mut Object, _sel: Sel, event: id) {
```
ArturKovacs, line 702: The `this` parameter doesn't have to be a `mut` anymore.

src/platform_impl/macos/view.rs
```diff
@@ -433,22 +454,44 @@ extern "C" fn set_marked_text(
...
         *marked_text_ref = marked_text;
+
+        let composed_string = marked_text_ref.clone().string();
+        let slice = slice::from_raw_parts(
+            composed_string.UTF8String() as *const c_uchar,
+            composed_string.len(),
+        );
+        let composed_string = str::from_utf8_unchecked(slice);
+
+        state.marked_text = composed_string.to_string();
```
ArturKovacs, line 465: Now there's a `state.marked_text` as well as a `markedText` field. They seem to store the same text. I have a specific idea of resolving this. Would you mind if I'd added changes to this PR?

If you would not like me to add commits to this PR, that's completely okay, but then please remove either `state.marked_text` or `markedText`. Preferrably we would only have the `NSMutableAttributedString` stored within the `ViewState`.

ArturKovacs (changes requested):

src/platform_impl/macos/view.rs
komi1230: Thanks !!

src/platform_impl/macos/view.rs
komi1230: Welcome your patch !
I'm also confused there are duplicated information for `markedText` and `state.marked_text`.
`markedText` is referred in `setMarkedText` but I'm not sure how to remove this because I'm not familiar with Obj-c and type system between Obj-c and Rust.

I appreciate your help.

ArturKovacs:
It should be fine, but please test as thoroughly as you can, because I may have broken something.

komi1230:
This works fine and looks really simple !
Thanks !
This seems ready to merge !

komi1230:
https://github.com/rust-windowing/winit/pull/1669 is duplicated against this PR.
It may be no problem to close.

ArturKovacs:
Oh no. I forgot about the changelog. Maybe I'll make another PR just for updating the changelog

garasubo:
@komi1230 If you are interested in #1497, please help implementing the APIs for composition events. The APIs itself are defined and implemented for X11 and Wayland, but no implementation for other platforms.

komi1230:
Sure! When I tackle this issue, I refered your implementation for X11 and Wayland. I'd like to return the favor. I'll also try.

PR #1981, opened 2021-07-30
Load "hand2" and "hand1" cursor icons for Wayland CursorIcon::Hand

AidanGG:

As per the X11 implementation https://github.com/rust-windowing/winit/blob/1c4d6e7613c3a3870cecb4cfa0eecc97409d45ff/src/platform_impl/linux/x11/util/cursor.rs#L84 the Wayland implementation should load cursors `hand2` and `hand1`, since some cursor themes such as Breeze do not ship an icon named `hand`. Without this change, Alacritty falls back to the default Adwaita hand icon when setting Breeze as the cursor theme, since Adwaita symlinks `hand` to `hand2`.


kchibisov (changes requested):
Should add changelog entry, other than that it looks fine.

AidanGG:
Done.

kchibisov (approved):

PR #1987, opened 2021-08-06
Fix X11 memory leak and remove mio-misc

### LINKED ISSUE #1984: Memory leak in EventLoop
On version 0.25.0 and the current git version when taking the example from the documentation summarized below, the application continues to climb in memory usage. That does not happen in version 0.24.0

Compiled in release for target x86_64_unknown_linux_gnu on X11

```rust
use winit::{event::*, event_loop::*, window::WindowBuilder};

fn main() {

        match event {
            Event::WindowEvent { event: WindowEvent::CloseRequested, window_id } if window_id == window.id() => {
                *control_flow = ControlFlow::Exit
            }
            Event::MainEventsCleared => {
                window.request_redraw();
            }
            Event::RedrawRequested(_) => {}
            _ => (),
        }
    });
}
```

Davester47:
I also fixed a couple of clippy lints in the files relevant to the bug, but some of them I was afraid to touch. As far as I could tell, mio-misc was causing the bug, so I removed it. It was hardly used, and easily replaced. I tested it against the example in #1984 and the leak is gone.

    I didn't see many bug fixes in the Changelog, so I didn't add it. No api changes were made.
    No user-facing changes.

Fixes #1984

### DIFF
--- a/src/platform_impl/linux/x11/mod.rs
+++ b/src/platform_impl/linux/x11/mod.rs
@@ -192,10 +192,7 @@ impl<T: 'static> EventLoop<T> {
             .register(&mut SourceFd(&xconn.x11_fd), X_TOKEN, Interest::READABLE)
             .unwrap();
 
-        //let (user_sender, user_channel) = channel(queue.clone(), NotificationId::gen_next());
         let (user_sender, user_channel) = std::sync::mpsc::channel();
-
-        //let (redraw_sender, redraw_channel) = channel(queue, NotificationId::gen_next());
         let (redraw_sender, redraw_channel) = std::sync::mpsc::channel();
 
         let target = Rc::new(RootELW {
@@ -442,18 +439,10 @@ impl<T> EventLoopWindowTarget<T> {
 
 impl<T: 'static> EventLoopProxy<T> {
     pub fn send_event(&self, event: T) -> Result<(), EventLoopClosed<T>> {
-        match self
-            .user_sender
+        self.user_sender
             .send(event)
             .map_err(|e| EventLoopClosed(e.0))
-        {
-            Ok(_) => {
-                // Can't send the event as an error, so panic
-                self.waker.wake().unwrap();
-                Ok(())
-            }
-            Err(e) => Err(EventLoopClosed(e.0)),
-        }
+            .map(|_| self.waker.wake().unwrap())
     }
 }
 
--- a/src/platform_impl/linux/x11/mod.rs
+++ b/src/platform_impl/linux/x11/mod.rs
@@ -412,8 +412,8 @@ impl<T: 'static> EventLoop<T> {
                             super::WindowId::X(wid),
                         )) = event
                         {
-                            wt.redraw_sender.waker.wake().unwrap();
                             wt.redraw_sender.sender.send(wid).unwrap();
+                            wt.redraw_sender.waker.wake().unwrap();
                         } else {
                             callback(event, window_target, control_flow);
                         }
@@ -442,12 +442,18 @@ impl<T> EventLoopWindowTarget<T> {
 
 impl<T: 'static> EventLoopProxy<T> {
     pub fn send_event(&self, event: T) -> Result<(), EventLoopClosed<T>> {
-        if self.waker.wake().is_err() {
-            return Err(EventLoopClosed(event));
-        }
-        self.user_sender
+        match self
+            .user_sender
             .send(event)
             .map_err(|e| EventLoopClosed(e.0))
+        {
+            Ok(_) => {
+                // Can't send the event as an error, so panic
+                self.waker.wake().unwrap();
+                Ok(())
+            }
+            Err(e) => Err(EventLoopClosed(e.0)),
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/platform_impl/linux/x11/mod.rs
```diff
@@ -407,7 +412,8 @@ impl<T: 'static> EventLoop<T> {
                             super::WindowId::X(wid),
                         )) = event
                         {
-                            wt.redraw_sender.send(wid).unwrap();
+                            wt.redraw_sender.waker.wake().unwrap();
+                            wt.redraw_sender.sender.send(wid).unwrap();
```
madsmtm, line 416: Shouldn't we send the event first, and only afterwards wake the polling thread? Or is there something here that I'm missing?

This goes for all three places where `Waker::wake()` is used.

Davester47: Yeah, that would make more sense. I'll make the change.

src/platform_impl/linux/x11/mod.rs
```diff
@@ -436,13 +442,18 @@ impl<T> EventLoopWindowTarget<T> {
...
+            .send(event)
+            .map_err(|e| EventLoopClosed(e.0))
+        {
+            Ok(_) => {
+                // Can't send the event as an error, so panic
+                self.waker.wake().unwrap();
+                Ok(())
+            }
+            Err(e) => Err(EventLoopClosed(e.0)),
+        }
```
madsmtm, line 456: I think you can use `Result::and_then` here instead? Or maybe that requires [RFC 2229](https://github.com/rust-lang/rfcs/blob/master/text/2229-capture-disjoint-fields.md)?

src/platform_impl/linux/x11/mod.rs
```diff
@@ -185,15 +187,16 @@ impl<T: 'static> EventLoop<T> {
...
             .register(&mut SourceFd(&xconn.x11_fd), X_TOKEN, Interest::READABLE)
             .unwrap();
 
-        let (user_sender, user_channel) = channel(queue.clone(), NotificationId::gen_next());
+        //let (user_sender, user_channel) = channel(queue.clone(), NotificationId::gen_next());
+        let (user_sender, user_channel) = std::sync::mpsc::channel();
 
-        let (redraw_sender, redraw_channel) = channel(queue, NotificationId::gen_next());
+        //let (redraw_sender, redraw_channel) = channel(queue, NotificationId::gen_next());
+        let (redraw_sender, redraw_channel) = std::sync::mpsc::channel();
```
madsmtm, line 199: Don't need these comments here, the old code is preserved by Git 😉

madsmtm (commented):
Looks fine to me, but haven't tested it and am not very experienced in using `mio`, so can't say for sure.

src/platform_impl/linux/x11/mod.rs
Davester47: I tried to follow your advice and use `Result::and_then` but clippy yelled at me and said I should use `Result::map` instead. I don't think that RFC 2229 applies here.

src/platform_impl/linux/x11/mod.rs
Davester47: Whoops, that one slipped past me. Thanks for pointing it out!

Davester47:
This is my first time using mio, so I'm not very experienced here either. Should I squash all my commits into one?

Davester47:
What's the process for getting this merged? I don't see any maintainers for X11 in https://github.com/rust-windowing/winit/wiki/Testers-and-Contributors, but is there someone else that can review my PR? IMO, this should get merged somewhat soon, since this bug is [affecting](#1987) [people](https://github.com/lilith645/Maat-Graphics/commit/175357fc4f6aefc43e07602215c00cbe4ce0433a). Is there anything I can do to help the process along?

maroider (commented):
> What's the process for getting this merged?

I quite frankly don't know.

Ideally, someone interested will come along and become a maintainer (or the previous X11 maintainer might become active again), but there's no telling how long that would take.

src/platform_impl/linux/x11/mod.rs
```diff
@@ -64,6 +58,11 @@ use crate::{
 const X_TOKEN: Token = Token(0);
 const USER_REDRAW_TOKEN: Token = Token(1);
 
+pub(self) struct WakeSender<T> {
+    pub sender: Sender<T>,
+    pub waker: Arc<Waker>,
+}
```
ArturKovacs, line 64: I believe that removing all the `pub` keywords from here would give the same accessibility. (Non-pub fields are accessible to the entire module)

ArturKovacs (changes requested):

ArturKovacs:
I'm willing to approve this, once the merge conflicts get resolved.

src/platform_impl/linux/x11/mod.rs
Davester47: You were right, the `pub` keywords aren't necessary. I have removed them.

Davester47:
Do you want me to squash my changes?

maroider:
There's no need for that. This repo is set up to only allow "squash and merge" merging from the web UI.

ArturKovacs (approved):

ArturKovacs:
@Davester47 don't change this PR now, but generally the best is to have a separate branch on your repo for the changes. Then just push new commits to that branch, no squashing, no force-pushing. This is the best for the review process as well because it allows us to see only those changes that you made since the last review.

Issue #1967, opened 2021-06-16
Window decorations on Gnome/Wayland


last-partizan:
Hello.

Recently neovide started to use wayland for rendering, and it uses those ugly decorations mentioned in #306 

I see winit switched to using client-toolkit for rendering, and it makes no attempt to give any visual quality, but acts only as fallback.

- https://github.com/Smithay/client-toolkit/issues/162#issuecomment-798955486

So, where is ther right place to fix those fallback decorations?

I created [discussion at discourse.gnome.org](https://discourse.gnome.org/t/proper-way-to-render-window-decorations-on-wayland-for-non-gtk-apps/6734) and someone suggested that winit should use [libdecor](https://gitlab.gnome.org/jadahl/libdecoration) to implement gtk decorations.

Could this be implemented?

By the way, is there some way to ask winit not to use wayland, even if it's available? That would fix issue with decorations as workaround, until it's resolved.

last-partizan:
https://github.com/Kethku/neovide/issues/728

Linking this issue.

alexheretic:
> By the way, is there some way to ask winit not to use wayland, even if it's available? That would fix issue with decorations as workaround, until it's resolved.

You can force xwayland usage with env var `WINIT_UNIX_BACKEND=x11`, see [docs](https://docs.rs/winit/0.25.0/winit/event_loop/struct.EventLoop.html#method.new).

maroider:
Using `libdecor` as a fallback sounds interesting, but it may be better to have such an integration implemented upstream in SCTK.

cc @vberger

vberger:
I purposefully will not support proper pretty decorations directly in SCTK, because that would require SCTK itself to depend on a whole font rendering engine to just to write text in the titlebar, and decorations look and feel is a complex matter on which everyone has its own opinion.

This is why SCTK provides [a trait](https://docs.rs/smithay-client-toolkit/0.14.0/smithay_client_toolkit/window/trait.Frame.html) that can be implemented to create an draw decorations as desired. This trait could be implemented in winit, or as a standalone crate. I just don't plan to have any font rendering in SCTK, that is way out of its scope.

Implementing the use of `libdecor` when available would, however, be in scope for SCTK.

vberger:
[...]

PR #1990, opened 2021-08-09
Update smithay-client-toolkit to 'v0.15.0'

kchibisov:
This commit also drops 'Theme' trait with its support types
in favor of 'FallbackFrame' meaning that winit will use some
predefined frame for the time being, since porting 'ConceptFrame'
will require adding font rendering librarires right into winit,
which is not desired.

### DIFF
--- a/src/platform_impl/linux/wayland/event_loop/mod.rs
+++ b/src/platform_impl/linux/wayland/event_loop/mod.rs
@@ -1,6 +1,7 @@
 use std::cell::RefCell;
 use std::collections::HashMap;
 use std::error::Error;
+use std::io::Result as IOResult;
 use std::process;
 use std::rc::Rc;
 use std::time::{Duration, Instant};
@@ -34,6 +35,8 @@ pub use state::WinitState;
 
 use sink::EventSink;
 
+type WinitDispatcher = calloop::Dispatcher<'static, WaylandSource, WinitState>;
+
 pub struct EventLoopWindowTarget<T> {
     /// Wayland display.
     pub display: Display,
@@ -42,7 +45,7 @@ pub struct EventLoopWindowTarget<T> {
     pub env: Environment<WinitEnv>,
 
     /// Event loop handle.
-    pub event_loop_handle: calloop::LoopHandle<WinitState>,
+    pub event_loop_handle: calloop::LoopHandle<'static, WinitState>,
 
     /// Output manager.
     pub output_manager: OutputManager,
@@ -50,8 +53,8 @@ pub struct EventLoopWindowTarget<T> {
     /// State that we share across callbacks.
     pub state: RefCell<WinitState>,
 
-    /// Wayland source.
-    pub wayland_source: Rc<calloop::Source<WaylandSource>>,
+    /// Dispatcher of Wayland events.
+    pub wayland_dispatcher: WinitDispatcher,
 
     /// A proxy to wake up event loop.
     pub event_loop_awakener: calloop::ping::Ping,
@@ -70,7 +73,7 @@ pub struct EventLoopWindowTarget<T> {
 
 pub struct EventLoop<T: 'static> {
     /// Event loop.
-    event_loop: calloop::EventLoop<WinitState>,
+    event_loop: calloop::EventLoop<'static, WinitState>,
 
     /// Wayland display.
     display: Display,
@@ -81,8 +84,8 @@ pub struct EventLoop<T: 'static> {
     /// Sender of user events.
     user_events_sender: calloop::channel::Sender<T>,
 
-    /// Wayland source of events.
-    wayland_source: Rc<calloop::Source<WaylandSource>>,
+    /// Dispatcher of Wayland events.
+    pub wayland_dispatcher: WinitDispatcher,
 
     /// Window target.
     window_target: RootEventLoopWindowTarget<T>,
@@ -102,7 +105,7 @@ impl<T: 'static> EventLoop<T> {
         let env = Environment::new(&display_proxy, &mut event_queue, WinitEnv::new())?;
 
         // Create event loop.
-        let event_loop = calloop::EventLoop::<WinitState>::new()?;
+        let event_loop = calloop::EventLoop::<'static, WinitState>::try_new()?;
         // Build windowing features.
         let windowing_features = WindowingFeatures::new(&env);
 
@@ -116,8 +119,21 @@ impl<T: 'static> EventLoop<T> {
         let output_manager = OutputManager::new(&env);
 
         // A source of events that we plug into our event loop.
-        let wayland_source = WaylandSource::new(event_queue).quick_insert(event_loop.handle())?;
-        let wayland_source = Rc::new(wayland_source);
+        let wayland_source = WaylandSource::new(event_queue);
+        let wayland_dispatcher = calloop::Dispatcher::new(wayland_source, |_, queue, ddata| {
+            queue.dispatch_pending(ddata, |event, object, _| {
+                panic!(
+                    "[calloop] Encountered an orphan event: {}@{} : {}",
+                    event.interface,
+                    object.as_ref().id(),
+                    event.name
+                );
+            })
+        });
+
+        let _wayland_source_dispatcher = event_loop
+            .handle()
+            .register_dispatcher(wayland_dispatcher.clone())?;
 
         // A source of user events.
         let pending_user_events = Rc::new(RefCell::new(Vec::new()));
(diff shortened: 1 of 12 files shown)


kchibisov:
Since sctk dropped `ConceptFrame` support new decorations are looking pretty awkward, however considering  that other libraries (sdl,glfw) provide poor looking decorations, since everything else is 'complexity & bringing font rendering engines' I don't think that we should port old sctk code at all.

What winit can do is to provide abstractions to work with decorations surfaces and provide buffers to attach to them.

kchibisov:
Added @vberger wrt calloop and decorations update.

maroider:
Relevant issues:
#1967
Smithay/client-toolkit#196

src/platform_impl/linux/wayland/event_loop/mod.rs
```diff
@@ -34,6 +35,9 @@ pub use state::WinitState;
 
 use sink::EventSink;
 
+type WinitDispatcher =
+    calloop::Dispatcher<WaylandSource, fn((), &mut EventQueue, &mut WinitState) -> IOResult<u32>>;
```
vberger, line 39: The second type parameter of `calloop::Dispatcher` is supposed to be the `Data` type of the calloop event loop, not the type of the closure (which is internally boxed). So tbh I'm surprised this type event works?

kchibisov: I'm not sure how to use it then. Assuming that I'd use `()` for `Data` I'm getting.

```
   --> src/platform_impl/linux/wayland/event_loop/mod.rs:192:33
    |
134 |           let wayland_dispatcher = calloop::Dispatcher::new(wayland_source, |_, queue, ddata| {
    |  ___________________________________________________________________________-
135 | |             queue.dispatch_pending(ddata, |event, object, _| {
136 | |                 panic!(
137 | |                     "[calloop] Encountered an orphan event: {}@{} : {}",
...   |
142 | |             })
143 | |         });
    | |_________- the found closure
...
192 |               wayland_dispatcher: wayland_dispatcher.clone(),
    |                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `()`, found closure
    |
    = note: expected struct `RefCell<Dispatcher<_, ()>>`
               found struct `RefCell<Dispatcher<_, [closure@src/platform_impl/linux/wayland/event_loop/mod.rs:134:75: 143:10]>>`

```
So it's saying that the type I'm providing when building dispatcher is closure and the result is Dispatcher<WaylandSource, closure>.

If you have an example how to use it it'll be nice.

vberger: Oh, right, my bad. I didn't realize that SCTK still depended on calloop 0.7... This part of calloop's API has been changed in 0.8...

src/platform_impl/linux/wayland/event_loop/mod.rs
```diff
@@ -116,8 +119,21 @@ impl<T: 'static> EventLoop<T> {
         let output_manager = OutputManager::new(&env);
 
         // A source of events that we plug into our event loop.
-        let wayland_source = WaylandSource::new(event_queue).quick_insert(event_loop.handle())?;
-        let wayland_source = Rc::new(wayland_source);
+        let wayland_source = WaylandSource::new(event_queue);
+        let wayland_dispatcher = calloop::Dispatcher::new(wayland_source, |_, queue, ddata| {
```
chrisduerr, line 123: What's `ddata`? Does the name have some significance or is this just typoed?

src/platform_impl/linux/wayland/event_loop/mod.rs
```diff
@@ -116,8 +119,21 @@ impl<T: 'static> EventLoop<T> {
...
+            queue.dispatch_pending(ddata, |event, object, _| {
+                panic!(
+                    "[calloop] Encountered an orphan event: {}@{} : {}",
+                    event.interface,
+                    object.as_ref().id(),
+                    event.name
+                );
```
chrisduerr, line 130: I'd be curious what this means exactly, I'd assume this happens when a callback isn't defined for an event? Is there a need to panic here?

src/platform_impl/linux/wayland/event_loop/mod.rs
```diff
@@ -403,16 +419,17 @@ impl<T: 'static> EventLoop<T> {
...
-                let source = self.wayland_source.clone();
-                let dispatched = handle.with_source(&source, |wayland_source| {
-                    let queue = wayland_source.queue();
-                    self.with_state(|state| {
-                        queue.dispatch_pending(state, |_, _, _| unimplemented!())
-                    })
-                });
+                let mut wayland_source = self.wayland_dispatcher.as_source_mut();
+                let queue = wayland_source.queue();
+                let state = match &mut self.window_target.p {
```
chrisduerr, line 424: Fortunately you haven't introduced it in this PR, but I have to say `p` isn't the greatest field name.

src/platform_impl/linux/wayland/event_loop/mod.rs
```diff
@@ -403,16 +419,17 @@ impl<T: 'static> EventLoop<T> {
...
+                    crate::platform_impl::EventLoopWindowTarget::Wayland(ref mut window_target) => {
```
chrisduerr, line 425: Maybe import part of this? A fully qualified path like this seems somewhat unusual. And is the `ref mut` necessary? Would have expected this to work just because of the `&mut self.window_target.p`.

src/platform_impl/linux/wayland/event_loop/mod.rs
```diff
@@ -403,16 +419,17 @@ impl<T: 'static> EventLoop<T> {
...
+                        window_target.state.get_mut()
+                    }
+                    #[cfg(feature = "x11")]
+                    _ => unreachable!(),
+                };
 
-                if let Ok(dispatched) = dispatched {
+                if let Ok(dispatched) = queue.dispatch_pending(state, |_, _, _| unimplemented!()) {
```
chrisduerr, line 432: I know the unimplemented here was already present in the previous code, but care to point out what isn't implemented here?

src/platform_impl/linux/wayland/event_loop/mod.rs
kchibisov: it's basically WinitState here, so I'd rename it, just called it that way initially since you have to cast it with older version of library.

src/platform_impl/linux/wayland/event_loop/mod.rs
kchibisov: It's for the case when wayland object isn't attached to any queue but you get a callback for it, with basically shouldn't happen, and when it's happening it's a bug in winit logic.

src/platform_impl/linux/wayland/event_loop/mod.rs
kchibisov: I know, but it's pretty old name in `event_loop.rs` and any refactoring to such names should be done in separate PRs, since it's touching the entire winit source tree.

src/platform_impl/linux/wayland/event_loop/mod.rs
kchibisov: The same unimplemented as before.

kchibisov:
I've asked you on irc to review this PR wrt dropping fancy client side decorations, since it's affecting alacritty as the end user. I personally don't really care, since I don't like client side decorations as a concept and would use them purely for 'special' clients or debugging, but we have gnome, thus we're affecting them.

src/platform_impl/linux/wayland/event_loop/mod.rs
chrisduerr: Does that bug in winit logic indicate that things will never be working again properly in the future anyway? Otherwise I think a panic isn't ideal.

src/platform_impl/linux/wayland/event_loop/mod.rs
chrisduerr: Definitely, just had to point this out.

chrisduerr:
> I've asked you on irc to review this PR wrt dropping fancy client side decorations, since it's affecting alacritty as the end user.

I was aware of this already since I was around on IRC when @vberger decided that this was the best approach forward. I somewhat nudged him in that direction when persistently complaining about the font handling of existing decorations.

I think for sctk it makes a lot of sense and I'm not sure it makes sense for winit to step in and provide them. I'd say winit probably comes with similar issues as sctk and will likely fail to provide anything other than super basic decorations with awkward font rendering. So shoving it even further downstream at least should make that easier, since font handling is likely present there already (it is in Alacritty at least).

With an infinite maintenance budget I think this should definitely be handled above application level in winit. But unfortunately that's just not the case. So the approach makes sense.

Issue #1988, opened 2021-08-08
WinitPointer::set_cursor fails silently on Wayland


sbosnick:
The `set_pointer` method on `WinitPointer` (in the 
`winit::platform_impl::platform::wayland::seat::pointer` module) fails silently if
`smithay_client_toolkit` is unable to set any of the named cursors that correspond
to the `CursorIcon` value passed in.

This can happen if the `ThemeSpec::System` cursor theme from `smithay_client_toolkit`
is not present on the system (i.e. `XCURSOR_THEME` is set to a non-existent value
or there isn't a theme named `default` on the system).

See alacritty/alacritty#4780 for an example of where this is occurring.

The failure in `set_pointer` is likely not something that `winit` can meaningfully
recover from, but it could log an error (or warning) when this happens. This would
allow users of `winit` to better detect this condition by reviewing their logs.

I propose that `set_pointer` log an error when this condition occurs.

PR #1991, opened 2021-08-11
On Wayland, log error for failure to set cursor

sbosnick:
Fixes: #1988

The inability to set the cursor using any of the named cursor files
likely indicates an error in the system on which we are running.
WinitPointer::set_cursor also does not have any way of returning an
error so just log the error to assist users in diagnosing the problem.

### DIFF
--- a/src/platform_impl/linux/wayland/seat/pointer/mod.rs
+++ b/src/platform_impl/linux/wayland/seat/pointer/mod.rs
@@ -112,7 +112,7 @@ impl WinitPointer {
                 return;
             }
         }
-        error!("Failed to set cursor");
+        warn!("Failed to set cursor");
     }
 
     /// Confine the pointer to a surface.
--- a/src/platform_impl/linux/wayland/seat/pointer/mod.rs
+++ b/src/platform_impl/linux/wayland/seat/pointer/mod.rs
@@ -112,7 +112,7 @@ impl WinitPointer {
                 return;
             }
         }
-        warn!("Failed to set cursor");
+        warn!("Failed to set cursor to {:?}", cursor_icon);
     }
 
     /// Confine the pointer to a surface.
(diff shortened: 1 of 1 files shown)


sbosnick:
I haven't check the last four boxes on the check list for the commit because I haven't done anything related to what is described, but I don't believe this PR needs any of the items that are listed.

src/platform_impl/linux/wayland/seat/pointer/mod.rs
```diff
@@ -109,9 +109,10 @@ impl WinitPointer {
         let serial = Some(self.latest_serial.get());
         for cursor in cursors {
             if self.pointer.set_cursor(cursor, serial).is_ok() {
-                break;
+                return;
             }
         }
+        error!("Failed to set cursor");
```
madsmtm, line 115: Minor nit: This should maybe be a warning instead, since the system can easily continue without the cursor having been set?

madsmtm (approved):

src/platform_impl/linux/wayland/seat/pointer/mod.rs
sbosnick: I have changed the `error!()` to a `warn!()`. This does make it a bit inconsistent with the same error as it is logged by `smithay_client_toolkit::window::concept_frame`, but that will likely be a non-issue if #1990 is merged (as the concept frame will be gone by then).

src/platform_impl/linux/wayland/seat/pointer/mod.rs
```diff
@@ -109,9 +109,10 @@ impl WinitPointer {
         let serial = Some(self.latest_serial.get());
         for cursor in cursors {
             if self.pointer.set_cursor(cursor, serial).is_ok() {
-                break;
+                return;
             }
         }
+        warn!("Failed to set cursor");
```
kchibisov, line 115: I don't think it's informative enough, you should provide the thing you've queried for, e.g. `cursor_icon`.

kchibisov (changes requested):

src/platform_impl/linux/wayland/seat/pointer/mod.rs
sbosnick: I have made this change.

kchibisov (approved):

Issue #1980, opened 2021-07-28
Borderless transparent windows don't appear (#1933 regression)


maroider:
As discussed [here](https://github.com/rust-windowing/winit/pull/1891#issuecomment-887814517), when glutin's `transparent` example is run against the current `master` branch, the window never appears. When run against the `master` branch *before* #1933 was merged, the window *does* appear.

PR #1994, opened 2021-08-11
Fix Window visibility regression

msiglreith:
Fixes https://github.com/rust-windowing/winit/issues/1980

The `visible` attribute was not correctly applied in case the decorations were disabled. Reordering to ensure that a `WM_NCCALCSIZE` (I think that's the desired event) is triggered due to size update. 

CHANGELOG entry omitted as the regression is not in any release.

### DIFF
--- a/src/platform_impl/windows/window.rs
+++ b/src/platform_impl/windows/window.rs
@@ -882,6 +882,11 @@ unsafe fn post_init<T: 'static>(
         thread_executor: event_loop.create_thread_executor(),
     };
 
+    // Set visible before setting the size
+    // to ensure the the attribute is correctly
+    // applied.
+    win.set_visible(attributes.visible);
+
     let dimensions = attributes
         .inner_size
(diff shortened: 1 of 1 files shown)


src/platform_impl/windows/window.rs
```diff
@@ -882,6 +882,11 @@ unsafe fn post_init<T: 'static>(
         thread_executor: event_loop.create_thread_executor(),
     };
 
+    // Set visible before setting the size
+    // to ensure the the attribute is correctly
+    // applied.
```
maroider, line 887: nit: This feels kind of cramped.
```suggestion
    // Set visible before setting the size to ensure the
    // attribute is correctly applied.
```
It could also be on one line, IMO.

maroider (changes requested):
It's unclear to me why we don't/can't make the window invisible immediately upon creation instead of deferring it like this, but this change seems to fix the bug for me.

PR #1999, opened 2021-08-15
On Wayland, implement 'request_user_attention'

kchibisov:
This commit implements 'request_user_attention' on Wayland with
new 'xdg_activation_v1' protocol.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ targets = ["i686-pc-windows-msvc", "x86_64-pc-windows-msvc", "i686-unknown-linux
 [features]
 default = ["x11", "wayland"]
 x11 = ["x11-dl", "mio", "mio-misc", "percent-encoding", "parking_lot"]
-wayland = ["wayland-client", "sctk"]
+wayland = ["wayland-client", "wayland-protocols", "sctk"]
 
 [dependencies]
 instant = { version = "0.1", features = ["wasm-bindgen"] }
@@ -83,7 +83,8 @@ features = [
 ]
 
 [target.'cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "openbsd", target_os = "netbsd"))'.dependencies]
-wayland-client = { version = "0.28", features = [ "dlopen"] , optional = true }
+wayland-client = { version = "0.29", features = [ "dlopen"] , optional = true }
+wayland-protocols = { version = "0.29", features = [ "staging_protocols"], optional = true }
 sctk = { package = "smithay-client-toolkit", version = "0.15.0", optional = true }
 mio = { version = "0.7", features = ["os-ext"], optional = true }
 mio-misc = { version = "1.0", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -83,7 +83,7 @@ features = [
 ]
 
 [target.'cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "openbsd", target_os = "netbsd"))'.dependencies]
-wayland-client = { version = "0.29", features = [ "dlopen"] , optional = true }
+wayland-client = { version = "0.29", features = [ "dlopen"], optional = true }
 wayland-protocols = { version = "0.29", features = [ "staging_protocols"], optional = true }
 sctk = { package = "smithay-client-toolkit", version = "0.15.0", optional = true }
 mio = { version = "0.7", features = ["os-ext"], optional = true }
--- a/src/platform_impl/linux/mod.rs
+++ b/src/platform_impl/linux/mod.rs
@@ -437,12 +437,12 @@ impl Window {
             _ => (),
         }
     }
-    pub fn request_user_attention(&self, _request_type: Option<UserAttentionType>) {
+    pub fn request_user_attention(&self, request_type: Option<UserAttentionType>) {
         match self {
             #[cfg(feature = "x11")]
-            Window::X(ref w) => w.request_user_attention(_request_type),
+            Window::X(ref w) => w.request_user_attention(request_type),
             #[cfg(feature = "wayland")]
-            Window::Wayland(ref w) => w.request_user_attention(_request_type),
+            Window::Wayland(ref w) => w.request_user_attention(request_type),
         }
     }
 
--- a/src/platform_impl/linux/mod.rs
+++ b/src/platform_impl/linux/mod.rs
@@ -442,7 +442,7 @@ impl Window {
             #[cfg(feature = "x11")]
             Window::X(ref w) => w.request_user_attention(_request_type),
             #[cfg(feature = "wayland")]
-            _ => (),
+            Window::Wayland(ref w) => w.request_user_attention(_request_type),
         }
     }
 
--- a/src/platform_impl/linux/wayland/window/mod.rs
+++ b/src/platform_impl/linux/wayland/window/mod.rs
@@ -17,7 +17,7 @@ use crate::platform_impl::{
     MonitorHandle as PlatformMonitorHandle, OsError,
     PlatformSpecificWindowBuilderAttributes as PlatformAttributes,
 };
-use crate::window::{CursorIcon, Fullscreen, WindowAttributes};
+use crate::window::{CursorIcon, Fullscreen, UserAttentionType, WindowAttributes};
 
 use super::env::WindowingFeatures;
 use super::event_loop::WinitState;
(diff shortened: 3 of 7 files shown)


src/platform_impl/linux/mod.rs
```diff
@@ -442,7 +442,7 @@ impl Window {
             #[cfg(feature = "x11")]
             Window::X(ref w) => w.request_user_attention(_request_type),
             #[cfg(feature = "wayland")]
-            _ => (),
+            Window::Wayland(ref w) => w.request_user_attention(_request_type),
```
chrisduerr, line 445: Can remove the `_` now that there's definitely always going to be a usage for `_request_type`.

src/platform_impl/linux/wayland/window/mod.rs
```diff
@@ -457,6 +462,17 @@ impl Window {
         Ok(())
     }
 
+    pub fn request_user_attention(&self, request_type: Option<UserAttentionType>) {
+        if !self.windowing_features.xdg_activation() {
+            warn!("'request_user_attention' isn't supported.");
```
chrisduerr, line 467: I see no other warning that uses punctuation at the end in winit at the moment. So removing that probably makes sense. Backticks also seem to be used exclusively instead of single quotes.

```suggestion
            warn!("`request_user_attention` isn't supported");
```

src/platform_impl/linux/wayland/window/mod.rs
```diff
@@ -457,6 +462,17 @@ impl Window {
...
+            return;
+        }
+
+        let attention_request = WindowRequest::Attention(request_type);
+        self.window_requests.lock().unwrap().push(attention_request);
+        self.event_loop_awakener.ping();
```
chrisduerr, line 473: Just something I've noticed that could be considered for a future refactoring: In this file every single instance of `window_requests.lock().unwrap().push(_)` is followed up by `event_loop_awakener.ping()`. If those two must go together it would probably be a good idea to have a function or something that does this, rather than duplicating it everywhere.

src/platform_impl/linux/wayland/window/shim.rs
```diff
@@ -188,6 +211,48 @@ impl WindowHandle {
...
+
+                let xdg_activation_token = xdg_activation.get_activation_token();
+                let surface = self.window.surface();
+                let window_id = wayland::make_wid(surface);
+                let xdg_activation = xdg_activation.clone();
+                xdg_activation_token.quick_assign(move |xdg_token, event, mut dispatch_data| {
+                    let winit_state = dispatch_data.get::<WinitState>().unwrap();
+                    if let Some(window_handle) = winit_state.window_map.get_mut(&window_id) {
+                        let surface = window_handle.window.surface();
+                        if let xdg_activation_token_v1::Event::Done { token } = event {
```
chrisduerr, line 240: Wouldn't it be better to check for this first, before doing any of the surface acquisition logic? So just a quick `event != Done => return`?

src/platform_impl/linux/wayland/window/shim.rs
```diff
@@ -188,6 +211,48 @@ impl WindowHandle {
...
+                            xdg_activation.activate(token, surface);
+                            *window_handle.xdg_activation_token.borrow_mut() = None;
```
chrisduerr, line 242: I don't quite understand what's going on here. The `activate` call is used to raise urgency according to the protocol. Since the `xdg_activation_token` is used to "un-raise" urgency I'd thus expect this to be set to something. But here you clear it by setting it to `None`. Wouldn't that prevent people from removing urgency or does this do something different from what I'm thinking?

chrisduerr (changes requested):

src/platform_impl/linux/wayland/window/shim.rs
kchibisov: The thing is that token can be used only once, thus it should be dropped right after the use, otherwise it'll result in a protocol error. Unsetting is done by compositor afaics.

kchibisov:
@chrisduerr Should be fine now, I've removed unsetting logic for the time being, but on the other hand we can keep it, since it won't harm anyway.

src/platform_impl/linux/wayland/window/mod.rs
```diff
@@ -530,12 +491,16 @@ impl Window {
             ..WaylandHandle::empty()
         }
     }
+
+    #[inline]
+    fn send_request(&self, request: WindowRequest) {
+        self.window_requests.lock().unwrap().push(request);
+        self.event_loop_awakener.ping();
+    }
```
chrisduerr, line 499: This really cleaned things up more than I even expected. Nice job.

src/platform_impl/linux/wayland/window/shim.rs
```diff
@@ -188,6 +214,52 @@ impl WindowHandle {
...
+
+        // Unsetting is done by compositor.
+        if request_type.is_none() {
+            return;
+        }
+
+        // The token is already present, nothing to set.
+        if self.xdg_activation_token.borrow().is_some() {
+            return;
+        }
```
chrisduerr, line 231: Is this the only reason why the `xdg_activation_token` needs to be stored? From a Rust point of view, I don't see much other use for it, but maybe there's some destructor stuff that requires keeping it alive?

src/platform_impl/linux/wayland/window/shim.rs
```diff
@@ -188,6 +214,52 @@ impl WindowHandle {
...
+    pub fn set_user_attention(&self, request_type: Option<UserAttentionType>) {
+        let xdg_activation = match self.xdg_activation.as_ref() {
+            None => return,
+            Some(xdg_activation) => xdg_activation,
+        };
+
+        // Unsetting is done by compositor.
+        if request_type.is_none() {
+            return;
+        }
```
chrisduerr, line 226: Could in theory combine this with the other match statement, that way only a single `return` would be necessary which might make it easier to read for some. But I personally also like just keeping the conditional logic trivial and separate so this should be fine.

chrisduerr (changes requested):
I haven't tested this myself on sway, but I'd assume if you've tested it with Alacritty there shouldn't be any need to compile sway from source just to verify what you've already tested. If you'd like a second look anyway, please let me know.

src/platform_impl/linux/wayland/window/shim.rs
kchibisov: It's getting destroyed via destroy method in a callback, so it's stored just to be properly cleaned.

chrisduerr: But what is destroyed is the token that is passed in the callback, not the token stored, right? So the storage servers no purpose?

kchibisov: They are the same, but I get what you're talking about. I think we can use `Cell<bool>` for that.

chrisduerr: Yeah, exactly. If the goal is only to not activate multiple times, a boolean is sufficient. Atomic might work?

kchibisov: Nah, you don't need atomic here it's single threaded in all cases since queue isn't sync + send.

kchibisov: Updated.

src/platform_impl/linux/wayland/window/shim.rs
```diff
@@ -188,6 +210,48 @@ impl WindowHandle {
...
     }
 
+    pub fn set_user_attention(&self, request_type: Option<UserAttentionType>) {
+        let xdg_activation = match self.xdg_activation.as_ref() {
+            None => return,
+            Some(xdg_activation) => xdg_activation,
+        };
+
+        //  Unsetting urgency is done by compositor and if the token is already is present we don't
+        //  need to create one more, so performing eaerly return.
```
chrisduerr, line 220: This comment is a bit messed up. I'd also recommend simplifying it a bit. Could just fix the typos though.

```suggestion
        // Urgency is only removed by the compositor and there's no need to raise urgency when it was already raised.
```

src/window.rs
```diff
@@ -755,9 +755,10 @@ impl Window {
     ///
     /// ## Platform-specific
     ///
-    /// - **iOS / Android / Web / Wayland:** Unsupported.
+    /// - **iOS / Android / Web :** Unsupported.
     /// - **macOS:** `None` has no effect.
     /// - **X11:** Requests for user attention must be manually cleared.
+    /// - **Wayland:** Requires `xdg_activation_v1` protocol.
```
chrisduerr, line 761: Maybe should comment that `None` has no effect, similar to macOS?

chrisduerr (changes requested):

chrisduerr (approved):
Code looks good to me.

Cargo.toml
```diff
@@ -83,7 +83,8 @@ features = [
 ]
 
 [target.'cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "openbsd", target_os = "netbsd"))'.dependencies]
-wayland-client = { version = "0.28", features = [ "dlopen"] , optional = true }
+wayland-client = { version = "0.29", features = [ "dlopen"] , optional = true }
```
maroider, line 86:
```suggestion
wayland-client = { version = "0.29", features = [ "dlopen"], optional = true }
```

PR #2006, opened 2021-08-22
Fix transparency on X11

sandmor:
This PR fixes the broke transparency feature on X11.

Note that for this [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md) said that this already worked, but it was code as TODO and only would work if someone provide a custom visual, which is not documented. This PR fixes that, allowing winit to find an appropriate visual for transparency.

This also fix [this issue](https://github.com/gfx-rs/wgpu/issues/687) in wgpu for X11 allowing to create transparent background(I just tried that, I didn't do extensive testing).

### DIFF
--- a/src/platform_impl/linux/x11/window.rs
+++ b/src/platform_impl/linux/x11/window.rs
@@ -8,6 +8,7 @@ use std::{
     ptr, slice,
     sync::Arc,
 };
+use x11_dl::xlib::TrueColor;
 
 use libc;
 use mio_misc::channel::Sender;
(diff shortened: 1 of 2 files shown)


coderedart:
does this mean that we can finally have transparent windows in winit with vulkan/wgpu?

sandmor:
No, this only affects X11(so a specific subset of Linux) I will don't do anything for any other platform furthermore I just tested the hello-triangle example with transparent background using a local clone of wgpu repo with the winit dependency pointing to my fork, and it worked smoothly contrary to before where you only saw a black background where it should be transparent.
As I said this only affects X11, that issue seems present on multiple platforms

src/platform_impl/linux/x11/window.rs
```diff
@@ -180,13 +181,41 @@ impl UnownedWindow {
...
+                if unsafe {
+                    (xconn.xlib.XMatchVisualInfo)(
+                        xconn.display,
+                        screen_id,
+                        32,
+                        TrueColor,
+                        vinfo.as_mut_ptr(),
+                    )
+                } != 0
+                {
```
ArturKovacs, line 198: It might be cleaner to extract this into a variable.

```suggestion
                let vinfo_initialized = unsafe {
                    (xconn.xlib.XMatchVisualInfo)(
                        xconn.display,
                        screen_id,
                        32,
                        TrueColor,
                        vinfo.as_mut_ptr(),
                    ) != 0
                };
                if vinfo_initialized {
```

src/platform_impl/linux/x11/window.rs
```diff
@@ -180,13 +181,41 @@ impl UnownedWindow {
...
+                    let vinfo = unsafe { vinfo.assume_init() };
+                    (vinfo.depth, vinfo.visual)
+                } else {
+                    // There is not a visual that matches the criteria of being `TrueColor` and 32 bits depth,
+                    // and panic or error is not appropiate so just ignore it and continue
+                    (ffi::CopyFromParent, ffi::CopyFromParent as *mut ffi::Visual)
+                }
```
ArturKovacs, line 205: I think it's a good idea to log a debug message here like
```rust
debug!("Could not set transparency, because XMatchVisualInfo returned zero for the required parameters");
```

ArturKovacs (commented):
Thanks a lot for the contribution!

I think there are only two small improvements that could be made, but otherwise this looks ready to be merged.

src/platform_impl/linux/x11/window.rs
sandmor: Addressed, thank you

src/platform_impl/linux/x11/window.rs
sandmor: Done

ArturKovacs (approved):

coderedart:
Thank you for this pr. i have been waiting for wgpu transparency for months. i thought the issue was with wgpu all this time. <3

sandmor:
Is it okay I was building a library to use winit with tiny-skia after realize that wgpu don't support transparency, then it doesn't worked as I expected am I `println` the depth of the window that was using when it was 24 I was to check the winit window creation code and then solved the issue :slightly_smiling_face:

PR #2009, opened 2021-08-22
Fix potential bug

sandmor:
So, after my last PR I realized of at small bug that can happen if we cannot find a visual for transparency. If that is the case the code anyway will attempt to create a `ColorMap` with `ffi::CopyFromParent` which even if I'm not sure what the result would be is not the intended behavior and I think that can possible cause problems.

### DIFF
--- a/src/platform_impl/linux/x11/window.rs
+++ b/src/platform_impl/linux/x11/window.rs
@@ -213,7 +213,7 @@ impl UnownedWindow {
                     let visual = vi.visual;
                     (xconn.xlib.XCreateColormap)(xconn.display, root, visual, ffi::AllocNone)
                 }
-            } else if window_attrs.transparent {
+            } else if visual as c_int != ffi::CopyFromParent {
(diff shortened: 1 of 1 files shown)


maroider:
Last-minute realizations like this are why I prefer to wait a day after approving a PR before merging it.

ArturKovacs:
> Last-minute realizations like this are why I prefer to wait a day after approving a PR before merging it.

You have a point there. Although new conflicts might be introduced during that day which is a bit annoying for the contributor. I don't know what the best strategy is.

ArturKovacs (approved):
Good catch! I missed this one as well.

sandmor:
Thank you

maroider:
> Although new conflicts might be introduced during that day which is a bit annoying for the contributor.

The most frequent source of conflicts seems to be CHANGELOG.md. I *think* putting entries at the top mitigates this, but I'm not entirely sure. I need to run a test at some point.

PR #2014, opened 2021-08-28
Windows: replace struct interface with one from crate

ItoYo16u:
I replaced  the temporary struct with one from crate in windows/dark-mode.rs according to comment.

- []  Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users


maroider:
Thanks for taking the time to contribute.

However, the change you've proposed is already being undertaken as a part of #2014.

PR #2021, opened 2021-09-08
Android: `Window::config` now initially returns accurate data.

ArthurKValladares:

As it currently stands, any data read from the initial config, such as the scale factor, is incorrect. This change makes it so that we can read accurate data from the config without needing `Event::ConfigChanged` to be triggered.


msiglreith (approved):
thanks!

Issue #2022, opened 2021-09-10
Expose occlusion state via NSWindowOccludedState


xacrimon:
On newer MacOS versions and Apple Silicon machines, Metal only runs resource cleanup *when* `NSWindowOccludedState` is set to false. Winit needs to expose this so the user can stop rendering when the window is fully occluded to prevent resource exhaustion.

maroider:
> Metal only runs resource cleanup `NSWindowOccludedState` is set to false

I can only assume you're missing an "if" between "cleanup" and "`NSWindowOccludedState`", and that you're referring to the instance property documented on [this page](https://developer.apple.com/documentation/appkit/nswindow/1419321-occlusionstate). However, that looks like a read-only property, so I don't see how we can even *set* it to `false`.

xacrimon:
@maroider Yeah I made a typo. We shouldn't set it but merely expose it somehow. It should return this on Mac and `true` on other platforms. This allows a user to stop rendering when the window is occluded so they don't exhaust the Metal resources.

maroider:
I guess something like `Window::is_visible()` might work, although this *could* conceivably also be a mac-specific extension (ie. a method on `WindowExtMacOS`).

xacrimon:
Yeah, that call is up to you guys but my take is that `Window::is_visible` is the better alternative due to end user simplicity.

xacrimon:
I can submit a patch to implement whatever you guys decide on.

kchibisov:
We have `Occluded` event for that.

PR #2023, opened 2021-09-12
Expose window visibility state

xacrimon:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 - On Wayland, bump `smithay-client-toolkit` to 0.15.
 - On Wayland, implement `request_user_attention` with `xdg_activation_v1`.
 - On X11, emit missing `WindowEvent::ScaleFactorChanged` when the only monitor gets reconnected.
-- Expose `NSWindowOcclusionState` on macOS through `Window::visible`.
+- On macOS , expose `NSWindowOcclusionState` through `Window::visible`. This method returns `false` on all other platforms.
 
 # 0.25.0 (2021-05-15)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - On Wayland, bump `smithay-client-toolkit` to 0.15.
 - On Wayland, implement `request_user_attention` with `xdg_activation_v1`.
 - On X11, emit missing `WindowEvent::ScaleFactorChanged` when the only monitor gets reconnected.
+- Expose `NSWindowOcclusionState` on macOS through `Window::visible`.
 
 # 0.25.0 (2021-05-15)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 - On Wayland, bump `smithay-client-toolkit` to 0.15.
 - On Wayland, implement `request_user_attention` with `xdg_activation_v1`.
 - On X11, emit missing `WindowEvent::ScaleFactorChanged` when the only monitor gets reconnected.
-- On macOS , expose `NSWindowOcclusionState` through `Window::visible`. This method returns `false` on all other platforms.
+- On macOS , expose `NSWindowOcclusionState` through `Window::visible`. This method returns `true` on all other platforms.
 
 # 0.25.0 (2021-05-15)
 
--- a/src/window.rs
+++ b/src/window.rs
@@ -767,6 +767,18 @@ impl Window {
     pub fn request_user_attention(&self, request_type: Option<UserAttentionType>) {
         self.window.request_user_attention(request_type)
     }
+
+    /// Returns the occlusion state of the current window. The occlusion
+    /// state is either `true` when the window is either hidden or fully covered
+    /// or `false` when some part of the window is visible.
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **iOS / Android / Web / Windows / X11 / Wayland:** Always returns false
+    #[inline]
+    pub fn occluded(&self) -> bool {
+        self.window.occluded()
+    }
 }
 
 /// Cursor functions.
--- a/src/window.rs
+++ b/src/window.rs
@@ -769,7 +769,9 @@ impl Window {
     }
 
     /// Returns the visible state of the window. This is `true` if some part of the window
-    /// is visible on screen and false otherwise.
+    /// is visible on screen and false otherwise. This is useful for pausing rendering
+    /// when the window isn't visible since some platforms may behave strangely if you
+    /// try to draw to an occluded window.
     ///
     /// ## Platform-specific
     ///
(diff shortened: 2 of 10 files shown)


xacrimon:
This issue solves #2022 and provides a functionality for helping users avoid https://github.com/gfx-rs/wgpu/issues/1936

src/window.rs
```diff
@@ -767,6 +767,19 @@ impl Window {
...
     }
+
+    /// Returns the visible state of the window. This is `true` if some part of the window
+    /// is visible on screen and false otherwise. This is useful for pausing rendering
+    /// when the window isn't visible since some platforms may behave strangely if you
+    /// try to draw to an occluded window.
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **iOS / Android / Web / Windows / X11 / Wayland:** Unsupported and always returns false
```
maroider, line 778:
```suggestion
    /// - **iOS / Android / Web / Windows / X11 / Wayland:** Unsupported and always returns `false`
```

src/window.rs
```diff
@@ -767,6 +767,19 @@ impl Window {
...
+    #[inline]
+    pub fn visible(&self) -> bool {
```
maroider, line 780: I'm not sure if this is the best name. The only alternative I can think of is `occluded`, but that inverts the logic.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 - On Wayland, bump `smithay-client-toolkit` to 0.15.
 - On Wayland, implement `request_user_attention` with `xdg_activation_v1`.
 - On X11, emit missing `WindowEvent::ScaleFactorChanged` when the only monitor gets reconnected.
+- Expose `NSWindowOcclusionState` on macOS through `Window::visible`.
```
maroider, line 17:
```suggestion
- On macOS , expose `NSWindowOcclusionState` through `Window::visible`. This method returns `false` on all other platforms.
```

maroider (commented):
cc @madsmtm

src/window.rs
madsmtm: Also, the code returns `true`

src/window.rs
```diff
@@ -767,6 +767,19 @@ impl Window {
     pub fn request_user_attention(&self, request_type: Option<UserAttentionType>) {
         self.window.request_user_attention(request_type)
     }
+
+    /// Returns the visible state of the window. This is `true` if some part of the window
+    /// is visible on screen and false otherwise. This is useful for pausing rendering
+    /// when the window isn't visible since some platforms may behave strangely if you
+    /// try to draw to an occluded window.
```
madsmtm, line 774: Can you give an example of this? E.g. why is it not enough to just check the `visible` property (via. [`-isVisible`](https://developer.apple.com/documentation/appkit/nswindow/1419132-visible?language=objc))

madsmtm (changes requested):
Can this be supported on other platforms? If not, maybe we should just expose it in `WindowExtMacOS` instead?

Edit: Ooops, sorry, missed the linked issue.

src/window.rs
madsmtm: This is intended as a way to see whether we should render to the window, so maybe name it as such: `should_render`?

xacrimon: Yeah that's what it is. We could name it occluded but that makes it confusing since it could be partially occluded or something else. `should_render` could work but it feels weird to name it after the purpose of which it exists and what we think the user will do with it rather than what it actually is. Describing the purpose of it via docs as I did feels like it makes more sense but I am happy to change it. I based the naming on how the rest of the api is named after what it is rather than what it's for. @madsmtm

src/window.rs
xacrimon: fixed

CHANGELOG.md
xacrimon: fixed

src/window.rs
xacrimon: @madsmtm `visible` returns true even when the window in question is fully occluded by other windows, i.e it's not hidden in the window manager. `AnimationCoreKit` on Mac as of some recent Big Sur versions on amd64 (can't confirm this) and the M1 chip stops periodical resource reclamation tasks as a battery saving measure when the window is hidden *or* fully occluded. As described in the linked `wgpu` issue this means that graphics applications may exhaust resources like memory and `MTLDrawable`'s when the window becomes fully occluded, causing an error within Metal.

xacrimon:
@madsmtm I believe Windows also has an API about this since chrome relies on it to pause rendering (unsure about X11/Wayland) but I lack the knowledge and hardware to test it. That said this is not of immediate concern of any other platform than MacOS since all other platforms don't change their resource handling behaviour when the window is fully occluded.

src/window.rs
madsmtm: I can see your reasoning, `visible` is fine to me

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 - On Wayland, bump `smithay-client-toolkit` to 0.15.
 - On Wayland, implement `request_user_attention` with `xdg_activation_v1`.
 - On X11, emit missing `WindowEvent::ScaleFactorChanged` when the only monitor gets reconnected.
+- On macOS , expose `NSWindowOcclusionState` through `Window::visible`. This method returns `false` on all other platforms.
```
madsmtm, line 17:
```suggestion
- On macOS , expose `NSWindowOcclusionState` through `Window::visible`. This method returns `true` on all other platforms.
```

madsmtm:
> I believe Windows also has an API about this

Then I think `Window::visible` is the correct place to put this.

maroider:
I unfortunately don't have any knowledge of the Windows side of this, but I did find [this document](https://github.com/chromium/chromium/blob/master/docs/windows_native_window_occlusion_tracking.md) on the chromium GitHub page, which indicates that tracking occlusion on Windows is a lot more work.

ArturKovacs:
@madsmtm I really think you should be memberized, especially because I'm considering moving away from being a maintainer.

CHANGELOG.md
xacrimon: fixed

xacrimon:
@maroider @madsmtm fixed the last open review, lmk if I need to anything to move this forward

madsmtm:
Bump @xacrimon

kchibisov (changes requested):
I don't like the addition and it won't work on Wayland for sure, since your window doesn't know whether it's visible or not. It seems like the idea is to draw when window is visible, but each OS provides far better mechanism to do so and I'd suggest to leave that to downstream clients, since winit doesn't perform frame scheduling, etc, etc.

If someone rely on `visible` in drawing they'll end up always drawing on platforms like Wayland.

xacrimon:
@kchibisov A similar feature is merged now anyway but we are sort of forced to do this due to Metal disabling GC for nonvisible windows on aarch OSX. There isn't a better way we know of to do it on that platform.

PR #2025, opened 2021-09-16
Add cursor grab for web target

TotalKrill:
- [x] Tested on all platforms changed                                patched winit, tried it in bevy_winit, works
- [x] Compilation warnings were addressed                      cargo build --target wasm32-unknown-unknown


cwfitzgerald:
I have implemented this independent of winit, and enabling pointer grab causes this call to throw a nasty error: https://github.com/rust-windowing/winit/blob/master/src/platform_impl/web/web_sys/canvas/pointer_handler.rs#L82-L84

> scene-viewer.js:347 panicked at 'Failed to set pointer capture: JsValue(InvalidStateError: Failed to execute 'setPointerCapture' on 'Element': InvalidStateError

TotalKrill:
Would not the correct way be to not have an expect in fallible code? 

The other way is if we could accurately predict that the state was actually that the pointer is already locked, that would solve your specific case, but since I do not know which ways that the pointerlock function can fail, its not a very good fix.

I mean, unfortunately in web, elements are shared in an unsafe manner, making it hard to have full control. As you say, you can do JS calls on the element and rust code will not ( i believe ) have a chance to react to it?

cwfitzgerald:
That sounds like a reasonable plan, I'll fork winit tonight and disable the expect to see what happens.

cwfitzgerald:
@TotalKrill I just checked out your branch and disabled the expect there, and everything works as we expect. So with that one change I think this PR is ready to go.

maroider:
> So with that one change I think this PR is ready to go.

\+ resolving the merge conflict for `CHANGELOG.md`

TotalKrill:
Im actually not entirely sure just removing the expect, I rather dislike silent errors

Would a log::error! be okay to add after the set_pointer_capture() call? like this:

``` rust
                let e = canvas.set_pointer_capture(event.pointer_id());
                if !e.is_ok() {
                    log::error("{:?}", e);
                }
```

cwfitzgerald:
The issue with a error is that the error will occur every single time you click the mouse when the cursor is grabbed. That would unnecessarily spam the logs. I don't think there's a situation where this could fail that we really care that it fails.

cwfitzgerald (approved):

src/platform_impl/web/web_sys/canvas/pointer_handler.rs
```diff
@@ -79,9 +79,7 @@ impl PointerHandler {
                     event::mouse_button(&event),
                     event::mouse_modifiers(&event),
                 );
-                canvas
-                    .set_pointer_capture(event.pointer_id())
-                    .expect("Failed to set pointer capture");
+                let _e = canvas.set_pointer_capture(event.pointer_id());
```
madsmtm, line 82: Nit: Document why we're swallowing the error. Something similar to what @cwfitzgerald wrote is fine:

> ... the error will occur every single time you click the mouse when the cursor is grabbed. [...] there's [not] a situation where this could fail that we really care that it fails.

madsmtm (approved):

TotalKrill:
I added a comment there and modified the commit message to be a bit more explicit

src/window.rs
```diff
@@ -799,7 +799,8 @@ impl Window {
     /// ## Platform-specific
     ///
     /// - **macOS:** This locks the cursor in a fixed location, which looks visually awkward.
-    /// - **iOS / Android / Web:** Always returns an [`ExternalError::NotSupported`].
+    /// - **Web:** This uses the pointer lock API
```
maroider, line 802: Are there any quirks of the pointer lock API that warrant the inclusion of this note? To my knowledge, the `## Platform-specific` section is for quirks that are relevant for users of winit, yet I don't see how this is all that useful to know.

src/platform_impl/web/web_sys/canvas/pointer_handler.rs
maroider: Are there any other cases where we expect calling `setPointerCapture` to fail?

src/window.rs
TotalKrill: I can remove that line if that is blocking this PR

src/platform_impl/web/web_sys/canvas/pointer_handler.rs
TotalKrill: I do not know to be honest, so if anyone else has any idea, feel free to suggest them

maroider: So, I looked up [the spec](https://w3c.github.io/pointerevents/#setting-pointer-capture), and it seems like the call will fail if there is no [active pointer](https://w3c.github.io/pointerevents/#dfn-active-pointer) (and it will also fail "If the element is not connected", but that's a weird situation to be in, I think).

src/window.rs
maroider: In light of the failure conditions for [`setPointerCapture`](https://w3c.github.io/pointerevents/#setting-pointer-capture), I believe the following should be documented:
> 3. If the element is not connected [DOM], throw an "InvalidStateError" DOMException.
> 4. If this method is invoked while the element's node document [DOM] has a locked element ([PointerLock] pointerLockElement), throw an "InvalidStateError" DOMException.
> 5. If the pointer is not in the active buttons state or the element's node document is not the active document of the pointer, then terminate these steps.

This text should probably not be copied verbatim.

maroider:
After coming back to this issue, I've come to the conclusion that the PR is fine as-is, although I'd like to test it myself before approving it.

maroider (approved):
Seems to work fine when I tested it.

ArturKovacs:
Note that the chagelog entry will end up under the wrong release after merging.

I suggest merging the latest `master` into this branch and then moving your changelog entry to the topmost section.

maroider:
I rebased the branch on top of master to get rid of the merge conflicts. Hope you don't mind :)

I'll merge this in a bit.

TotalKrill:
I definitely don't mind, in fact I am thankful 👍

PR #2039, opened 2021-10-22
Add cargo audit CI + audit and deps.rs badges

martin-t:

This will allow to automatically detect potential vulnerabilities as part of CI. It also runs periodically once a week because new ones can be discovered as time passes without any changes to winit.

I also added 2 badges to readme
- For cargo audit - I think a separate badge is better because it helps spread awareness of cargo audit and as more crates throughout the ecosystem begin using it, issues will be fixed faster.
- For deps.rs - It helps people notice outdated dependencies faster and often can be used to quickly determine the health / maintenance status of a project.

Currently winit transitively depends on time and chrono which have advisories released for them (but chrono [doesn't yet have a fix](https://github.com/chronotope/chrono/issues/602)), this means CI will fail and that's expected. You might want to merge this PR only after those have been fixed to avoid a constantly failing CI. It's also possible to avoid them by using simple_logger without default features (but then logs in examples will lack timestamps AFAICT).

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -1,0 +1,28 @@
+name: Audit
+
+# This is a separate file so it can have a separate badge in readme
+# and therefore spread awareness of cargo audit a tiny bit.
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+  schedule:
+    # Runs at 15:00 UTC on Fri
+    - cron: "0 15 * * 5"
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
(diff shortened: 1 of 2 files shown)


madsmtm:
We're using `cargo-deny` nowadays, so I don't really see the added value in `cargo-audit`.

I'm unsure of the value of the badges, at least for the second one [it seems like deps.rs doesn't check platform-specific dependencies](https://deps.rs/repo/github/rust-windowing/winit), which makes it kinda useless

madsmtm:
So I think I'll close this PR, feel free to reopen if you want to add just the audit badge

PR #2044, opened 2021-10-31
On Wayland, add wayland-dlopen feature to use dlopen

### LINKED ISSUE #2037: Can wayland `dlopen` feature be optional?
Currently `wayland-client` and `sctk` use `dlopen` feature to link with system wayland library

but I want static version for dependency problem and preferred Rust implementation

kchibisov:
While winit was always using dlopen for opening system libs, it
provides a way now to disable dlopen feature helping with linking
on some targets.

Fixes #2037.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@
 - On Wayland, implement `request_user_attention` with `xdg_activation_v1`.
 - On X11, emit missing `WindowEvent::ScaleFactorChanged` when the only monitor gets reconnected.
 - On X11, if RANDR based scale factor is higher than 20 reset it to 1
-- On Wayland, feature `wayland-dlopen` to use `dlopen` for opening system libraries enabled by default.
+- On Wayland, add an enabled-by-default feature called `wayland-dlopen` so users can opt out of using `dlopen` to load system libraries.
 
 # 0.25.0 (2021-05-15)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - On Wayland, implement `request_user_attention` with `xdg_activation_v1`.
 - On X11, emit missing `WindowEvent::ScaleFactorChanged` when the only monitor gets reconnected.
 - On X11, if RANDR based scale factor is higher than 20 reset it to 1
+- On Wayland, feature `wayland-dlopen` to use `dlopen` for opening system libraries enabled by default.
 
 # 0.25.0 (2021-05-15)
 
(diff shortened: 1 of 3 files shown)


maroider:
It looks like you didn't input the feature array correctly in the CI config.

```
Run cargo $CMD doc --no-deps --target x86_64-unknown-linux-gnu $OPTIONS --features $FEATURES
  cargo $CMD doc --no-deps --target x86_64-unknown-linux-gnu $OPTIONS --features $FEATURES
  shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
  env:
    RUST_BACKTRACE: 1
    CARGO_INCREMENTAL: 0
    RUSTFLAGS: -C debuginfo=0
    OPTIONS: --no-default-features
    FEATURES: ,wayland
    CMD: 
    Updating crates.io index
 Downloading crates ...
  Downloaded scoped-tls v1.0.0
  Downloaded version_check v0.9.3
  Downloaded downcast-rs v1.2.0
[...]
```

However, it is not in any way clear to me whether it wants an inline YAML array, a quoted string, or maybe something else.

kchibisov:
>However, it is not in any way clear to me whether it wants an inline YAML array, a quoted string, or maybe something else.

It wanted a quoted string in the end :- |

Cargo.toml
```diff
@@ -83,9 +84,9 @@ features = [
 ]
 
 [target.'cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "openbsd", target_os = "netbsd"))'.dependencies]
-wayland-client = { version = "0.29", features = [ "dlopen"], optional = true }
+wayland-client = { version = "0.29", default_features = false,  features = ["use_system_lib"], optional = true }
```
maroider, line 87: Are there any default features to disable in the first place?

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 - On Wayland, implement `request_user_attention` with `xdg_activation_v1`.
 - On X11, emit missing `WindowEvent::ScaleFactorChanged` when the only monitor gets reconnected.
 - On X11, if RANDR based scale factor is higher than 20 reset it to 1
+- On Wayland, feature `wayland-dlopen` to use `dlopen` for opening system libraries enabled by default.
```
maroider, line 20:
```suggestion
- On Wayland, add an enabled-by-default feature called `wayland-dlopen` so users can opt out of using `dlopen` to load system libraries.
```
The important change here is that users can opt out, so I think we should call that out explicitly.

maroider (commented):
> It wanted a quoted string in the end :- |

Ah. At least it works now :)

Otherwise, this looks mostly OK, although I'm unable to test it myself since my Wayland setup is broken ATM.

Cargo.toml
kchibisov: `dlopen` itself is a default feature.

maroider (approved):

Issue #1995, opened 2021-08-12
Coordinate Android NDK update with winit 0.27


msiglreith:
https://github.com/rust-windowing/winit/issues?q=is%3Aopen+label%3A%22platform%3A+Android%22+milestone%3A%22Version+0.27%22+

- [ ] https://github.com/rust-windowing/android-ndk-rs/pull/314 Coordinate with the release of `ndk 0.7.0` (breaking changes: https://github.com/rust-windowing/android-ndk-rs/blob/master/ndk/CHANGELOG.md) prior to releasing `winit` (seems like a nice cut-off point);
- [ ] Alongside the update, re-exporting ndk-glue would be helpful under the light of https://github.com/rust-windowing/winit/pull/1993
  - [ ] Consider extending `ndk_context` instead and automagically support a wider range of `ndk-glue` versions?
  - [ ] Alternatively there have been suggestions [TODO: reference needed] of reworking how `ndk-glue` passes its structs to the user and removing `static` globals plus their accessor functions;
- [x] Use `NativeWindow`'s `RawWindowHandle` implementation directly: https://github.com/rust-windowing/android-ndk-rs/pull/274
- [x] https://github.com/rust-windowing/winit/pull/2307
- [x] https://github.com/rust-windowing/android-ndk-rs/pull/282
- [x] https://github.com/rust-windowing/android-ndk-rs/pull/285
- [x] https://github.com/rust-windowing/android-ndk-rs/pull/286
- [x] https://github.com/rust-windowing/android-ndk-rs/pull/287
- [x] https://github.com/rust-windowing/android-ndk-rs/pull/288
- [x] Fix `winit` CI failures on Android: https://github.com/rust-windowing/android-ndk-rs/pull/289
- [x] https://github.com/rust-windowing/android-ndk-rs/pull/290
- [x] https://github.com/rust-windowing/android-ndk-rs/pull/291
- [x] https://github.com/rust-windowing/android-ndk-rs/pull/292
- [x] https://github.com/rust-windowing/android-ndk-rs/pull/293
- [x] https://github.com/rust-windowing/android-ndk-rs/pull/294
- [x] https://github.com/rust-windowing/android-ndk-rs/pull/295
- [x] https://github.com/rust-windowing/android-ndk-rs/pull/296
- [ ] https://github.com/rust-windowing/android-ndk-rs/pull/320

msiglreith:
[...]

PR #2047, opened 2021-11-01
Android: Bump ndk/ndk-glue version

msiglreith:

Tested with a small sample application that it still runs with the updated ndk version.
This is breaking as users need to update their version of `ndk-glue` as well as it internally has a static global variable which should be unique across crates!

https://github.com/rust-windowing/winit/issues/1995

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,7 @@ The `ndk_glue` version needs to match the version used by `winit`. Otherwise, th
 | :---: | :------------------: |
 | 0.24  | `ndk_glue = "0.2.0"` |
 | 0.25  | `ndk_glue = "0.3.0"` |
+| 0.26  | `ndk_glue = "0.4.0"` |
 
 Running on an Android device needs a dynamic system library, add this to Cargo.toml:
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -125,6 +126,6 @@ To ensure compatibility with older MacOS systems, winit links to
 CGDisplayCreateUUIDFromDisplayID through the CoreGraphics framework.
 However, under certain setups this function is only available to be linked
 through the newer ColorSync framework. So, winit provides the
-`WINIT_LINK_COLORSYNC` environment variable which can be set to `1` or `true` 
+`WINIT_LINK_COLORSYNC` environment variable which can be set to `1` or `true`
```
maroider, line 129:
```suggestion
`WINIT_LINK_COLORSYNC` environment variable which can be set to `1` or `true`
```

maroider (approved):
This change seems simple enough, and CI passes, so that's nice.
If there isn't anything wrong with `ndk = "0.4.0"`, then I don't think there's much else to check.

Is the omission of a re-export (as requested in #1993/#1995) intentional?

msiglreith:
> Is the omission of a re-export (as requested in #1993/#1995) intentional?

Right I missed that part. I'm experimenting with it a bit and not sure yet - adding features (e.g `logger`) for transitive depenencies is not supported, requiring us to re-export it somehow (not sure how tho with target specific features) or the user needs respecify the `ndk-glue` dependency with the requested feature themself. Another part is requiring to override the path to the `ndk-glue` to make the macro work (see the `ndk_glue = ".."` part).
```rust
#[cfg_attr(target_os = "android", ndk_glue::main(backtrace = "on", ndk_glue = "winit::platform::android::ndk_glue"))]
fn main() {
   ...
}
```

Overall I would probably keep it out for now. Maybe somehow has a suggestion how properly export it. The above attempts doesn't look so user friendly.

PR #2046, opened 2021-11-01
macOS: titlebar in borderless fullscreen fix

tinaun:
Fixes #1195, but does not implement the user-facing changes talked about in that issue (should we create a new issue for that feature?)

- [x] Tested on all platforms changed
- [x] Compilation warnings were addressed
- [x] `cargo fmt` has been run on this branch
- [x] `cargo doc` builds successfully
- [x] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users
- [ ] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior
- [ ] Created or updated an example program if it would help users understand this functionality
- [x] Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented

PR #2053, opened 2021-11-04
Show the menu bar in borderless fullscreen on macOS

### LINKED ISSUE #1195: Fullscreen Titlebar Won't Show on macOS

irh:
Fixes #1195.

Following on from @cixel's work in #1326, this PR aims to fix the issue where the menu bar doesn't show up when using the borderless fullscreen mode on macOS.

#2046 was opened recently but following some discussion @francesca64 proposed reactivating this branch in a new PR.

There are a couple of edge cases that need to be manually tested, [e.g. this comment](https://github.com/rust-windowing/winit/pull/1326#issuecomment-761823885)
> Running `cargo run --example multithreaded`:
> 1. Press `esc` twice to close the first two windows
> 2. Press `alt`+`F`. The window goes into exclusive fullscreen
> 3. Press `F`. The window seems to change to borderless fullscreen which is indicated by the fact that I can use `cmd`+`tab` to switch between applications. But now the menu bar isn't shown.

Additionally, since I [first looked at this issue](https://github.com/rust-windowing/winit/pull/1326#issuecomment-933350528), I've added a fix for the following path:

Running `cargo run --example multithreaded`:
1. Press `F`. The window goes into borderless fullscreen.
2. Press `alt`+`F`. The window goes into exclusive fullscreen
3. Press `F`. The window returns to borderless fullscreen. At this point the menu bar should appear normal and the green 'resize' button should be active.

I assume that there are further edge cases to consider that I haven't found, the logic here is a bit thorny and it can be hard to keep track of all the options and modes, but for the main cases this seems to be working correctly, at least in my testing.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 - On Wayland, add an enabled-by-default feature called `wayland-dlopen` so users can opt out of using `dlopen` to load system libraries.
 - **Breaking:** On Android, bump `ndk` and `ndk-glue` to 0.4.
 - On Windows, increase wait timer resolution for more accurate timing when using `WaitUntil`.
+- On macOS, fix an issue that prevented the menu bar from showing in borderless fullscreen mode.
 
 # 0.25.0 (2021-05-15)
 
--- a/src/platform_impl/macos/window.rs
+++ b/src/platform_impl/macos/window.rs
@@ -909,6 +909,13 @@ impl UnownedWindow {
                 NSApp().setPresentationOptions_(presentation_options);
 
                 util::restore_display_mode_async(video_mode.monitor().inner.native_identifier());
+
+                // Restore the normal window level following the Borderless fullscreen
+                // `CGShieldingWindowLevel() + 1` hack.
+                let () = msg_send![
+                    *self.ns_window,
+                    setLevel: ffi::NSWindowLevel::NSNormalWindowLevel
+                ];
             },
             _ => INTERRUPT_EVENT_LOOP_EXIT.store(false, Ordering::SeqCst),
         };
--- a/src/platform_impl/macos/window.rs
+++ b/src/platform_impl/macos/window.rs
@@ -785,12 +785,13 @@ impl UnownedWindow {
 
             let mut fade_token = ffi::kCGDisplayFadeReservationInvalidToken;
 
-            unsafe {
-                let app = NSApp();
-                trace!("Locked shared state in `set_fullscreen`");
-
-                let mut shared_state_lock = self.shared_state.lock().unwrap();
-                shared_state_lock.save_presentation_opts = Some(app.presentationOptions_());
+            if matches!(old_fullscreen, Some(Fullscreen::Borderless(_))) {
+                unsafe {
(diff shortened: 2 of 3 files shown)


francesca64:
With the first test case, if I click somewhere else on the menu bar in borderless fullscreen, the menu bar disappears again. This only seems to happen after transitioning from exclusive->borderless, and the menu bar doesn't reappear even if I cycle exclusive->borderless again.

With the second test case, I unreliably experience a substantial delay (sometimes even briefly beach-balling) when transitioning from exclusive->borderless. However, this already happen on `master`, so it's not an issue with this PR.

irh:
> With the first test case, if I click somewhere else on the menu bar in borderless fullscreen, the menu bar disappears again. This only seems to happen after transitioning from exclusive->borderless, and the menu bar doesn't reappear even if I cycle exclusive->borderless again.

Yes I see that too, well spotted 👍 I'll see if I can find a fix.

irh:
OK I can see that I won't be able to find a fix any time soon (I have no clue what the root cause is, other than it likely being related to the hacks involved in exclusive fullscreen), and I'm not going to have more time to dedicate to this until sometime next week. If this needs to be fixed before merging I'd be happy for someone with more knowledge of mac windowing than me to take this over, I'm mostly just pulling on levers I don't really understand here!

francesca64:
@irh completely understandable; windowing is a nightmare, and you've already done so much to help!

I also won't be able to investigate this until next week. If anyone wants to swoop in and try to figure this out before then, they'll be handsomely rewarded with at least 1 heart emoji.

irh:
An update from me, I've got no spare time for at least another couple of weeks while my daughter's quarantining from a Covid close contact, so I'm not going to be able to help with pushing this forward for quite a while.

I've also noticed that the multithreaded example is badly broken on the M1 macbook that I use at home (the wrong resolution is used in exclusive fullscreen, and interaction seems broken).

My thinking is that the thing to do here would be to merge a fix for the missing menu issue so that at least the idiomatic macOS fullscreen mode is working correctly, and then leave the refinement of mode switching as followup work? I fear that getting the mode switching to 100% will be a bit of a thorny challenge that will continue to drag on, and the fix for the primary issue is pretty straightforward, so it seems reasonable to me to treat the two topics separately.

madsmtm (approved):
Agree, I even think we should just merge this as-is since it includes helpful considerations and goes part of the way in fixing the mode-switching issue.

I've opened https://github.com/rust-windowing/winit/issues/2068 to track fixing the mode-switching for later.

francesca64 (approved):
@madsmtm seconded; this PR is strictly an improvement.

@irh thank you for everything you've done, and I wish you and your family the best!

Issue #2060, opened 2021-11-11
Windows: Parent of owned child window not visible


msiglreith:
Declaring one windows as owner will make the owner window invisible which is unintentional and a regression from 0.25.
Seems to be caused by https://github.com/rust-windowing/winit/pull/1933

Minimum example:
```rust
fn main() -> anyhow::Result<()> {
    let event_loop = EventLoop::new();

    let owner = WindowBuilder::new()
        .with_title("owner")
        .build(&event_loop)
        .unwrap();

    let child = WindowBuilder::new()
        .with_title("child")
        .with_owner_window(owner.hwnd() as _)
        .build(&event_loop)
        .unwrap();

    event_loop.run(move |event, _, control_flow| {
        *control_flow = ControlFlow::Wait;
    })
}
```

msiglreith:
hm looks like only the child window receives a `WM_PAINT` event

msiglreith:
Not passing the parent HWND on window creation for owned windows seems to alleviate the issue and show the parent window correctly - doing the initialization in `WM_CREATE` instead of `WM_NCCREATE` as well but that would defeat the purpose of #1933
Unsure what's the underlying issue tho

maroider:
If I've understood the issue correctly, it seems you've run into what Raymond Chen describes [here](https://devblogs.microsoft.com/oldnewthing/20100315-00/?p=14613). That is, a window can optionally have `a parent XOR an owner`.

msiglreith:
@maroider sorry, forget that part - not setting the handle delete the ownership which isn't the goal either 😓

maroider:
I think there were only a couple of things that absolutely *had* to happen in response to `WM_NCCREATE` rather than `WM_CREATE`, like a call to `EnableNonClientDpiScaling` (which strictly speaking doesn't require our window state to be initialized). We don't currently dispatch any events in response to `WM_NCCREATE` and `WM_CREATE`, so it should be feasible to move most of the initialization logic into `WM_CREATE` (with some care).

maroider:
I finally have a Windows 10 VM I can test with, and I managed to reproduce the bug.
Strangely enough, the owner window still appears when I hover over the preview in the taskbar.
I'll get you a review for #2062 shortly.

PR #2062, opened 2021-11-14
Windows: Split window initialization across NCCREATE and CREATE

msiglreith:
Fixes https://github.com/rust-windowing/winit/issues/2060

Doing all the initialization steps inside `WM_NCCREATE` doesn't seem to work out (see issue) - aspects affecting the client area like size appear to require handling inside `WM_CREATE`.

### DIFF
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -38,7 +38,7 @@ use crate::{
     monitor::MonitorHandle as RootMonitorHandle,
     platform_impl::platform::{
         dark_mode::try_theme,
-        dpi::{become_dpi_aware, dpi_to_scale_factor, enable_non_client_dpi_scaling},
+        dpi::{become_dpi_aware, dpi_to_scale_factor},
         drop_handler::FileDropHandler,
         event::{self, handle_extended_keys, process_key_params, vkey_to_winit_vkey},
         monitor::{self, MonitorHandle},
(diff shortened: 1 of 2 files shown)


src/platform_impl/windows/window.rs
```diff
@@ -726,20 +676,184 @@ unsafe impl Send for WindowWrapper {}
...
+        enable_non_client_dpi_scaling(window);
+
+        Window {
+            window: WindowWrapper(window),
+            window_state,
+            thread_executor: self.event_loop.create_thread_executor(),
+        }
+    }
+
+    unsafe fn create_window_data(&self, win: &mut Window) -> event_loop::WindowData<T> {
```
maroider, line 742:
```suggestion
    unsafe fn create_window_data(&self, win: &Window) -> event_loop::WindowData<T> {
```

nit: I don't think there's any reason to actually pass in the `Window` mutably here.

maroider (commented):
This seems to be largely sensible, although I'm not entirely sold on having `on_nccreate` make the call to `SetWindowLongPtrW`.

Furthermore, I tested it in my brand spanking new Windows 10 VM, and the fix seems to work as advertised :)

msiglreith:
Thanks for the review and testing!

> I'm not entirely sold on having on_nccreate make the call to SetWindowLongPtrW

I switched it back to storing it in the event callback

src/platform_impl/windows/event_loop.rs
```diff
@@ -813,23 +813,28 @@ pub(super) unsafe extern "system" fn public_window_callback<T: 'static>(
...
+                Some(userdata) => {
+                    winuser::SetWindowLongPtrW(window, winuser::GWL_USERDATA, userdata as _);
+                    winuser::DefWindowProcW(window, msg, wparam, lparam)
+                }
+                None => -1, // failed to create the window
+            };
+
+            return result;
         }
+        // `userdata` is set in `WM_NCCREATE` appearing before `WM_CREATE`.
```
maroider, line 826:
```suggestion
```
This comment isn't quite as necessary now that the call to `SetWindowLongPtrW` is clearly visible.

maroider (approved):

PR #2067, opened 2021-11-17
Prevent SEGFAULT on X11 composition event handling

kawaemon:
note: I'm opening PR to `composition-event` branch, not master.

I found that `preedit_draw_callback` function dereferences a null pointer when system locale is not properly set. This PR adds null-check so that prevent segfault.


kchibisov:
Sorry for not taking a look at your branch initially. Given that composition-event branch got merged I've send your fix here https://github.com/rust-windowing/winit/pull/2316.

Issue #2010, opened 2021-08-22
The case for callbacks


madsmtm:
Hey, I've been tinkering with (primarily) the macOS backend for a while now, and I feel like I've been bit several times by the callback-y nature of AppKit poorly matching the (almost) data-only `Event`.

So here's my proposal for remedying this.

## Background

Most of the underlying system APIs work by letting the user register callbacks on a window, that then get called when an event happens. See the below table for a complete overview:

| Library | Method                                                        |
| ------- | ------------------------------------------------------------- |
| AppKit  | Multiple callbacks per application/window/view class.         |
| UIKit   | Same as AppKit.                                               |
| Windows | Single callback per window class.                             |
| Web     | Multiple callbacks per DOM element.                           |
| Wayland | Multiple callbacks on "seats" (don't really understand this). |
| X11     | Uses an internal event queue that you poll from.              |
| Android | Same as X11.                                                  |

`winit` then turns this into a variant of `event::Event`, and either calls the event handler directly or buffers it for later (currently buffered events on macOS: `WindowEvent` except `ScaleFactorChanged`, `DeviceEvent` and `UserEvent`).

## The issue

Well, this is honestly a pretty good user-facing API! It's easy to share data between event handlers (since there's only one), and is in general pretty "rusty".

However, it's very much an abstraction, and as with almost any abstraction, we lose some form of control.

One problem is that it's hard for users to know which events are emitted directly / blocking, and hence need to be acted on immediately (like `RedrawRequested`), and which ones are buffered and can be handled at leisure.

Another is that the user can't directly respond to the events unless we break the assumption that `Event`s are just data, see https://github.com/rust-windowing/winit/issues/1387.

Thirdly, on macOS, the behaviour is different depending on whether a callback is registered or not, see https://github.com/rust-windowing/winit/pull/1759#discussion_r571592243. So we can't expose this behaviour as an event, because it would negatively affect users that aren't handling the event.

In general, the design forces you to pay for stuff you don't use, which is against the Rust principle of zero-cost abstractions.

## The actual proposal

I would like to propose that we change the internals of our backends into having callback-style APIs.

We'd still build the much more user-friendly `EventLoop` API on top of that, but we'd have the goal of at some point in the future exposing some of these "lower-level" callbacks to user code.

[...]

PR #2073, opened 2021-11-25
Emit LoopDestroyed on cmd+q

### LINKED ISSUE #1998: `Event::LoopDestroyed` not always emmited
On Mac, when I press Cmd+Q the window closes, but there is no `Event::LoopDestroyed` emitted, so I cannot save my program state.

winit version: 0.25.0, used via glium/glutin.

tinaun:
fixes #1998 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- On macOS, Fix emitting `Event::LoopDestroyed` on CMD+Q.
 - On macOS, fixed an issue where having multiple windows would prevent run_return from ever returning.
 - On Wayland, fix bug where the cursor wouldn't hide in GNOME.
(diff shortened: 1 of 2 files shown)


tinaun:
It can be separate from this PR but I’m wondering if we should make cmd-q a “close requested” event since you can cancel out of it with `applicationShouldTerminate:`

madsmtm:
We should consider the option I proposed here: https://github.com/rust-windowing/winit/issues/1998#issuecomment-979778868.

Otherwise, this PR looks fine (except missing a changelog entry).

> make cmd-q a “close requested” event since you can cancel out of it with applicationShouldTerminate:

`WindowEvent::CloseRequested` exists, so that would require us to call it for every window, which is not that nice; and also, there is really no way to communicate back whether the application should continue terminating, or keep going (see #2010)

emilk:
I'm so glad this is fixed 🥳

PR #2074, opened 2021-11-27
Use `ResizeObserver` to control the size of the canvas

### LINKED ISSUE #1661: Allow canvas size to be controlled by page layout on web target
Currently, the web target uses a fixed window size and it sizes the canvas based on it. It would be nice to instead have a mode that allows the canvas to be laid out on the page with CSS layout like flexbox or CSS grid. The resizing of the element can be detected using `ResizeObsever`, which can also handle pixel-perfect canvas sizing (https://github.com/rust-windowing/winit/pull/1659#issuecomment-675601670). In this mode, the `winit` window should not be resizable by user code.

Liamolucko:

Specifically on Firefox Developer Edition and Safari.

Resolves #1661

I've changed the web backend to set the window size entirely based on the canvas' CSS size using `ResizeObserver`. It uses `device-pixel-content-box` where available, but falls back on `ResizeObserverEntry.contentRect` if it's not supported.

One issue is how to handle the window's initial size. `ResizeObserver` does fire an event for the initial size when created, but it won't run until the task finishes, which means it isn't available for any init code before the event loop starts. I've worked around this by just getting the element's computed size and subtracting padding, and using that to estimate its size for the initial `inner_size` value, but this doesn't work at all if the canvas is created by winit and then added to the DOM. If a canvas isn't in the DOM, its size is reported as 0, so if the canvas is created this way its initial size will always be 0.

At the moment this has to pull in a git version of `web-sys` and enable `cfg(web_sys_unstable_apis)`, so it can't really be shipped. The API does seem to be supported in pretty much all modern browsers, but it's marked unstable since it's a draft.

# To-do
- [x] Wait for `web-sys` to release a new version
- [X] Either wait for `ResizeObserver` to be stabilised or feature gate this somehow

# Unresolved questions
- [ ] Should this be hidden behind a mode, instead of always enabled? Applications which already resize manually using `set_inner_size` should work functionally identically anyway; this should only really affect things which don't touch the size at all.
- [ ] Should setting min/max size set the CSS min/max-width/height properties?
- [ ] Should outer size refer to the canvas' border box? Now that CSS can be freely applied to the canvas, it's possible for it to have padding or borders, in which case the element's outer size might not be the same as its content size.


maxammann:
Looks very cool! Let me know if you need help testing.

rukai:
I tried this branch on my app and my app blows up in wgpu because I'm giving it a window with size of 0x0.
Maybe we should we be respecting https://docs.rs/winit/latest/winit/window/struct.Window.html#method.set_min_inner_size or at least hardcode a minimum size of 1x1.

Liamolucko:
> I tried this branch on my app and my app blows up in wgpu because I'm giving it a window with size of 0x0. Maybe we should we be respecting https://docs.rs/winit/latest/winit/window/struct.Window.html#method.set_min_inner_size or at least hardcode a minimum size of 1x1 ?

That issue was occuring because the window's initial size was being measured in `Window::new`, before the canvas was added to the DOM. At that point, the canvas's size is reported as 0x0. To work around that, I've changed it to measure the canvas's initial size on the first call to `Window::inner_size` instead, if it isn't in the DOM during `Window::new`.

maroider:
> Maybe we should we be respecting https://docs.rs/winit/latest/winit/window/struct.Window.html#method.set_min_inner_size or at least hardcode a minimum size of 1x1?

Getting resize events that tell you the window is `0x0` is expected at present time, so user code generally has to check for it before passing the size on to wgpu anyway.

stephanemagnenat:
Any status update on this PR? The bug it solves is likely to hit pretty much everyone using Rust with GPU on the web (through Bevy or something else), so even an imperfect implementation that relies on some unstable-but-widely-available feature is certainly much better than the status quo.

rukai:
I tested again and it now works perfectly for my application!

Thanks so much for implementing this, I would love to see it merged under a feature flag before ResizeObserver is stabilized.

For anyone else wanting to make use of this, here is the html, css and rust I used to get value out of this.
The result is a winit app that fills all the available horizontal space at a 4/3 aspect ratio.
```html
<div id="render">
    <canvas gets inserted here by rust code>
</div>
```

```css
#render {
    width: 100%;
    aspect-ratio: 4 / 3;
}
```

```rust
        let window = Window::new(&event_loop).unwrap();

        let canvas = window.canvas();
        canvas
            .style()
            .set_css_text("display: block; width: 100%; height: 100%");

        document.get_element_by_id("render").unwrap()
            .append_child(&web_sys::Element::from(canvas))
            .unwrap();
```

maroider:
Given that the spec for the API in question is still a "working draft", I'm a bit hesitant to enable it by default, despite its "wide availablity".

Liamolucko:
I've put the `ResizeObserver` functionality behind a `css-size` feature, so that this can be used without everyone having to enable `cfg(web_sys_unstable_apis)`.

I've noticed a bit of an issue, though - if the scale factor is changed before the event loop starts, it'll panic, since it tries to handle the event synchronously when the event loop isn't running. It used to just not check for them at all before the event loop started, but that's not great either. How do other platforms handle this?

rukai:
From looking at the table here: https://developer.mozilla.org/en-US/docs/Web/API/Resize_Observer_API#browser_compatibility
It looks like all the unstable API's we need have been stabilized?
Can wasm-bindgen be updated accordingly and this PR progressed?

Liamolucko:
It was marked unstable because the spec is a 'Working Draft', which isn't technically a real spec yet. `web-sys` already has plenty of Working Drafts not marked unstable, though (e.g. the [File API](https://www.w3.org/TR/FileAPI/)), so I think it'd be fine to stabilize it.

src/platform_impl/web/web_sys/mod.rs
```diff
@@ -60,35 +60,83 @@ impl WindowExtWebSys for Window {
...
+
+    // Use `getBoundingClientRect` instead of the width and height properties because it doesn't round to the nearest integer.
+    let rect = raw.get_bounding_client_rect();
+    let style = window
+        .get_computed_style(raw)
+        .unwrap()
+        .expect("`getComputedStyle` returned `None`");
+
+    let prop = |name| -> f64 {
+        let value = style.get_property_value(name).unwrap();
```
jwagner, line 86: Might be missing something but what guarantees that the property values are going to be in px?

Liamolucko: I didn't think about this at the time, but after doing some digging it's *almost* guaranteed - the properties will always be in `px` unless the canvas' `display` is `none` or `contents`.

<details>
<summary>Spec-digging:</summary>

First, [getComputedStyle](https://drafts.csswg.org/cssom/#dom-window-getcomputedstyle) defines the values of the properties it returns as such:

> with the value being the [resolved value](https://drafts.csswg.org/cssom/#resolved-value) computed for obj using the style rules associated with doc.

For the padding properties, the resolved value is defined as:

> If the property applies to the element or pseudo-element and the [resolved value](https://drafts.csswg.org/cssom/#resolved-value) of the ['display'](https://drafts.csswg.org/css-display-3/#propdef-display) property is not ['none'](https://drafts.csswg.org/css-display-3/#valdef-display-none) or ['contents'](https://drafts.csswg.org/css-display-3/#valdef-display-contents), then the resolved value is the [used value](https://drafts.csswg.org/css-cascade-5/#used-value). Otherwise the resolved value is the [computed value](https://drafts.csswg.org/css-cascade-5/#computed-value).

The used value is defined as:

> The used value is the result of taking the [computed value](https://drafts.csswg.org/css-cascade-5/#computed-value) and completing any remaining calculations to make it the absolute theoretical value used in the formatting of the document.

This isn't really a formal definition, but I think that this should mean that the used value of the padding is an [absolute length](https://drafts.csswg.org/css-values-4/#absolute-lengths).

No special case is defined for the border width properties, so they fall back to the default:

> The [resolved value](https://drafts.csswg.org/cssom/#resolved-value) is the [computed value](https://drafts.csswg.org/css-cascade-5/#computed-value).

The computed value is defined as:

> The computed value is the result of resolving the [specified value](https://drafts.csswg.org/css-cascade-5/#specified-value) as defined in the “Computed Value” line of the property definition table, generally absolutizing it in preparation for [inheritance](https://drafts.csswg.org/css-cascade-5/#inheritance).

The padding properties are defined [here](https://drafts.csswg.org/css-box/#padding-physical), and their computed values are defined as:

> a computed [\<length-percentage\>](https://drafts.csswg.org/css-values-4/#typedef-length-percentage) value

`<length-percentage>` just means `<length>` or `<percentage>`.

The border width properties are defined [here](https://drafts.csswg.org/css-backgrounds/#the-border-width), and their computed values are defined as:

> absolute length; zero if the border style is [none](https://drafts.csswg.org/css-backgrounds/#valdef-line-style-none) or [hidden](https://drafts.csswg.org/css-backgrounds/#valdef-line-style-hidden)

So, I think both of these properties' resolved values should be absolute lengths as long as the canvas' `display` isn't `none` or `contents`, since:
- the border width's computed value is defined to always be an absolute length.
- as long as the canvas' display isn't `none` or `contents`, the padding's resolved value is its used value, which is absolute.

Then, how these values are converted to strings is defined in [getPropertyValue](https://www.w3.org/TR/cssom/#dom-cssstyledeclaration-getpropertyvalue), which calls into the [serialize a CSS value](https://www.w3.org/TR/cssom/#serialize-a-css-value) to do the actual conversion. `<length>` is defined to be serialized as:

> The <number> component serialized as per <number> followed by the unit in its canonical form as defined in its respective specification.

Both the properties' resolved values were [absolute lengths](https://drafts.csswg.org/css-values-4/#absolute-lengths) when the `display` wasn't `none` or `contents`, and:

> All of the absolute length units are [compatible](https://drafts.csswg.org/css-values-4/#compatible-units), and [px](https://drafts.csswg.org/css-values-4/#px) is their [canonical unit](https://drafts.csswg.org/css-values-4/#canonical-unit).

So, I think the values should always be in `px` when the `display` isn't `none` or `contents`.

When it is, though, the `padding` can be a `<percentage>`, which is serialized as:

> The <number> component serialized as per <number> followed by the literal string "%" (U+0025).

So, this won't work when `display` is `none` or `contents`.

</details>

So, I've changed `prop` to return 0 for padding when `display` is `none` or `contents`, since I don't think it's worth the effort to compute the percentage for a hidden canvas.

Also, removing the `px` by slicing is dodgy, so I've replaced that with a `strip_suffix("px")` and added a bit more context to the panic message.

rukai:
https://github.com/rust-windowing/winit/pull/2216 has merged, so it would be good to rebase/merge to make it easy to test on the new example.

src/platform_impl/web/web_sys/mod.rs
jwagner: I wasn't aware of that. Thanks! :+1:

src/window.rs
```diff
@@ -521,7 +521,7 @@ impl Window {
     ///
     /// - **iOS:** Can only be called on the main thread. Returns the `PhysicalSize` of the window's
     ///   [safe area] in screen space coordinates.
-    /// - **Web:** Returns the size of the canvas element.
+    /// - **Web:** Returns the size of the canvas element's framebuffer.
```
stephanemagnenat, line 524: A comment regarding documentation, `inner_size`'s documentation for web talks about the "framebuffer size", but `set_inner_size` talks about the "content box via the `style` attribute". However, if we have HiDPI (device pixel ratio > 1), the framebuffer size (physical size) will be bigger than the style attribute (logical size). So the documentation looks inconsistency to me, for two functions that look like getter and setter from their names. Or am I missing something?

src/window.rs
```diff
@@ -550,7 +550,7 @@ impl Window {
     /// ## Platform-specific
     ///
     /// - **iOS / Android:** Unsupported.
-    /// - **Web:** Sets the size of the canvas element.
+    /// - **Web:** Sets the size of the canvas element's content box via the `style` attribute.
```
stephanemagnenat, line 553: See my comment above.

marcofugaro:
This is indispensable for any canvas application compiled to wasm, any progress?

src/window.rs
Liamolucko: Sort of. `inner_size` returns the framebuffer size as a typed `PhysicalSize`, which when converted to a `LogicalSize` using `PhysicalSize::to_logical` will be ~equal to the logical size set with `set_inner_size`.

It's true that they're not a proper getter/setter pair though. It's possible for the CSS set by `set_inner_size` to be overridden with `!important`, causing `set_inner_size` to have no effect  what's returned by `inner_size`.

stephanemagnenat: Ah yes, I missed the return type. Thank you for the explanation! An option could be for `set_inner_size` to be called something like `set_inner_logical_size` to account for its parameter to be untyped wrt physical vs logical. But it is of course a tradeoff between naming simplicity and semantic precision...

src/platform_impl/web/web_sys/mod.rs
```diff
@@ -60,35 +60,101 @@ impl WindowExtWebSys for Window {
...
+        .expect("`getComputedStyle` failed")
+        .expect("`getComputedStyle` returned `None`");
+
+    // This also can't fail according to the spec.
+    if style.get_property_value("box-sizing").unwrap() == "border-box" {
+        let prop = |name| -> f64 {
+            let value = style.get_property_value(name).unwrap();
+            // Cut off the last two characters to remove the `px` from the end.
+            value[..value.len() - 2].parse().unwrap()
+        };
```
madsmtm, line 147: Nit: Deduplicate this helper closure into a function (since it's the same as above).

src/platform_impl/web/web_sys/mod.rs
```diff
@@ -60,35 +60,101 @@ impl WindowExtWebSys for Window {
...
+        raw.set_width(physical_size.width);
+        raw.set_height(physical_size.height);
+    }
 
-    raw.set_width(physical_size.width);
-    raw.set_height(physical_size.height);
+    let window = web_sys::window().unwrap();
+    let style = window
+        .get_computed_style(raw)
+        // This can't fail according to the spec; I don't know why web-sys marks it as throwing and having an optional result.
```
madsmtm, line 137: Perhaps we should open an issue in `web-sys` about it, then?

src/platform_impl/web/event_loop/window_target.rs
```diff
@@ -233,35 +229,45 @@ impl<T> EventLoopWindowTarget<T> {
...
+                let new_size = if backend::is_fullscreen(&raw) {
+                    intended_size = old_size;
 
-            backend::set_canvas_size(&raw, Size::Physical(new_size));
-            runner.send_event(Event::WindowEvent {
-                window_id: RootWindowId(id),
-                event: WindowEvent::Resized(new_size),
+                    backend::inner_size(&raw)
+                        // I don't think it's possible for an element to become fullscreen whilst not being in the DOM.
+                        .unwrap()
```
madsmtm, line 254: Nit: Use `expect` with such a message instead

src/platform_impl/web/event_loop/runner.rs
```diff
@@ -289,82 +336,178 @@ impl<T: 'static> Shared<T> {
         }
     }
 
-    pub fn handle_scale_changed(&self, old_scale: f64, new_scale: f64) {
-        // If there aren't any windows, then there is nothing to do here.
-        if self.0.all_canvases.borrow().is_empty() {
+    /// Handle a change in scale factor, without any other information.
+    ///
+    /// `initial` is whether this is coming from the initial `ScaleChangeDetector`, rather than the `ResizeObserver`.
```
madsmtm, line 341: Flags like this is IMHO usually bad practice, perhaps things can be refactored slightly to avoid it? (concretely: I have trouble immediately parsing the `!(initial && observer_will_run)` conditional).

madsmtm (changes requested):
I gave parts of it a cursory look, but this is a huge change and I'm not well-versed enough in web technologies to really tell if things are correct or not.

Sadly, we [currently don't have a maintainer for the web platform](https://github.com/rust-windowing/winit/blob/7341ee80eacde5272cd0cf000d0aba43d983f634/.github/CODEOWNERS#L31-L33), so the best way forwards @marcofugaro and other interested parties, would be to review the changeset with your detail-oriented eyes on (adding a simple "approving" review via. GitHub is also data), and test it in edge-cases (e.g. also when the `css-size` feature is not enabled, or when changing monitors and the like).

That would give me more confidence that the change is correct, and I would be much more willing to merge it then.

toji:
> Given that the spec for the API in question is still a "working draft", I'm a bit hesitant to enable it by default, despite its "wide availablity".

Pointed here from a conversation on Mastodon. I know effectively nothing about Rust but I am the editor of [one](https://www.w3.org/TR/webxr/) or [two](https://www.w3.org/TR/webgpu/) web standards, so I figured I could help clarify this point at least.

A spec being a "Working Draft" doesn't mean much in terms of how stable, widespread, or standardized the specification has become. Ideally the spec authors want their specification to be granted "Candidate Recommendation" (CR) status by the W3C, but that typically isn't even considered until at least two independent browsers have been _shipping_ compatible implementations of the feature for at least a little while. And even the spec moving from Draft to CR status is a formality that has no real bearing on how browsers, developers, or users approach the feature.

What this means for developers is that [CanIUse.com](https://caniuse.com/mdn-api_resizeobserver) is a far more practical reference than the spec documents for determining what is _actually_ shipped and stable. In this specific case ResizeObserver has been ~universally available since Jan 2020, and is definitely not going to be removed from the web platform or see any breaking changes in the foreseeable future. [Chrome usage stats](https://chromestatus.com/metrics/feature/timeline/popularity/2592) show that over 20% of web pages that users load make use of ResizeObserver, and nobody is eager to break 20% of the web, regardless of whether the spec says "Draft" at the top or not.

Hope that helps clear up some of the (admittedly confusing) verbiage around web specs!

mcclure:
It appears that this code has been working for over a year and the ResizeObserver "draft" API has been unchanged for over three years. Caniuse has ResizeObserver working on everything except IE11.

I was hoping to release some wgpu sample code before Chrome's initial release of webgpu next week (?). `--cfg=web_sys_unstable_apis` is not a problem for this application because the whole of webgpu is unstable.
This PR is somewhere between useful and essential for practical webgpu use. I set the version of winit in my sample code's Cargo.toml ( https://github.com/mcclure/rs-hello [2b3303b](https://github.com/mcclure/rs-hello/commit/2b3303ba333d2fdbf0cf89eb76c5faf2eac7c05a) / https://data.runhello.com/j/wgpu/2/) to directly pull this branch from github and as-is the PR works great (although it's too bad because I think this means I'm basing my code on a version of winit older than 0.28.3?). I can't speak to your package's stability policy but is there a way some version of this could get into shipping winit versions? The feature is a very good one.

Thanks!

maroider:
Thanks for the clarification, @toji. If anyone is willing to pick up work on this again (merge conflicts and whatnot), then I'd be happy to see us use `ResizeObserver` by default.

>  I can't speak to your package's stability policy but is there a way some version of this could get into shipping winit versions?

We generally don't back-port changes due to a lack of bandwidth. If I'm not mistaken, a few fixes have been back-ported in the past, but I don't recall the specifics.

daxpedda:
I'm happy to get this PR sorted out and rebased after https://github.com/rust-windowing/winit/pull/2789.

mcclure:
Awesome!

> If I'm not mistaken, some fixes do get back-ported, but I don't recall the criteria for it.

Sorry, I meant "can it get in a future shipping version", not that I needed it compatible with any specific shipping version.

daxpedda:
I think the big problem right now is how to deal with `--cfg=web_sys_unstable_apis`, it's a pretty terrible user experience, though I wouldn't mind still going through with this as long as it's gated behind that instead of introducing a crate feature that requires it.

I made https://github.com/rustwasm/wasm-bindgen/pull/3459 to avoid this whole issue though.

daxpedda:
Replaced by #2859.

PR #2076, opened 2021-11-28
Fix warnings

filnet:

### DIFF
--- a/examples/request_redraw_threaded.rs
+++ b/examples/request_redraw_threaded.rs
@@ -1,14 +1,14 @@
-use std::{thread, time};
-
-use simple_logger::SimpleLogger;
-use winit::{
-    event::{Event, WindowEvent},
-    event_loop::{ControlFlow, EventLoop},
-    window::WindowBuilder,
-};
-
 #[cfg(not(target_arch = "wasm32"))]
 fn main() {
+    use std::{thread, time};
+
+    use simple_logger::SimpleLogger;
+    use winit::{
+        event::{Event, WindowEvent},
+        event_loop::{ControlFlow, EventLoop},
+        window::WindowBuilder,
+    };
+
     SimpleLogger::new().init().unwrap();
     let event_loop = EventLoop::new();
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,7 +50,7 @@ jobs:
     env:
       RUST_BACKTRACE: 1
       CARGO_INCREMENTAL: 0
-      RUSTFLAGS: "-C debuginfo=0"
+      RUSTFLAGS: "-C debuginfo=0 --deny warnings"
       OPTIONS: ${{ matrix.platform.options }}
       FEATURES: ${{ format(',{0}', matrix.platform.features ) }}
       CMD: ${{ matrix.platform.cmd }}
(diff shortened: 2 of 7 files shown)


filnet:
Would it make sense to fail the builds if there are warnings like the on this PR fixes ?

madsmtm:
Thanks for this, these have been bothering me for a while!

> Would it make sense to fail the builds if there are warnings like the ones this PR fixes ?

I would say yes (can easily be done with `RUSTFLAGS: "-D warnings"`)

filnet:
> I would say yes (can easily be done with `RUSTFLAGS: "-D warnings"`)

Where would one do that. There are no CI related files in the winit repo.

maroider:
> There are no CI related files in the winit repo.

GitHub actions looks in `.github/workflows` for workflows, and our CI is in `.github/workflows/ci.yml`.

filnet:
The wasm build fail as expected.

filnet:
I am getting spammed by CI messages about unrelated failed checks whenever I push changes to the branch.
Doesn't make for a good user experience when contributing to winit.

The failed checks should be disabled until properly fixed ?

filnet:
> GitHub actions looks in `.github/workflows` for workflows, and our CI is in `.github/workflows/ci.yml`.

Totally overlooked them ! Thanks.

madsmtm:
Please don't disable the failing `nightly` checks; they're very useful for spotting future incompatibility errors (see #2078), and they don't prevent people from merging the PR. If we disable them, we forget them :wink:

maroider:
> and they don't prevent people from merging the PR

They don't??

Now that you've mentioned it, the "Squash and merge" button does look awfully clickable.

madsmtm:
> The notification spam is unfortunate, but I don't know if we can turn those off.

I don't think we can, though people can control it themselves under [Settings => Notifications => Actions](https://github.com/settings/notifications).

filnet:
Ping.

examples/request_redraw_threaded.rs
```diff
@@ -1,6 +1,9 @@
+#[cfg(not(target_arch = "wasm32"))]
 use std::{thread, time};
 
+#[cfg(not(target_arch = "wasm32"))]
 use simple_logger::SimpleLogger;
+#[cfg(not(target_arch = "wasm32"))]
```
maroider, line 6: This doesn't look all that great, but it's not that big of an issue either.

maroider (approved):

examples/request_redraw_threaded.rs
maroider: > I agree but what's the alternative ?

Realistically, there's the option of moving the imports into the `main` function. There's also the option of using the `xtask` pattern to work around this, but that's more work.

filnet: I moved the imports in the `main` function.

rukai:
Ah, just a heads up that warnings are not stable.
So if you want to fail on warnings you should really pin your rust version, see https://github.com/rukai/mado_rust_library_ci_template/ for more details.

madsmtm:
Well, we run CI on `nightly`, which would surface changes to warnings that would start failing our CI in a way that prevents a merge.

I'm not strictly opposed to pinning our Rust version, though I think it should be for other reasons like guaranteeing a MSRV.

rukai:
Ah sure, if your paying enough attention to the nightly workflow then that would work too.

PR #2078, opened 2021-11-29
Fix `CGDisplayCreateUUIDFromDisplayID` linking (again)

madsmtm:
See previous issue: https://github.com/rust-windowing/winit/pull/1626.

The `cocoa` crate links to AppKit, which made the symbol `CGDisplayCreateUUIDFromDisplayID` from ApplicationServices/ColorSync (which AppKit uses internally) available to us (on macOS 10.8 to 10.13).

However, this does not work on macOS 10.7 (where AppKit does not link to ColorSync internally). Instead of relying on this, we should just link to ApplicationServices directly.

Also, lately something changed in rustc so that the linker's macOS version is 10.7, see https://github.com/rust-lang/rust/issues/91372, which is why the nightly builds were failing.

  - [x] Tested on macOS Mojave 10.14.6, using `MACOSX_DEPLOYMENT_TARGET=X.Y cargo build --example window`.
  - [x] Tested on macOS 10.10
  - [x] And if @Imberflur could test this using their docker setup that would be nice as well!

### DIFF
--- a/src/platform_impl/macos/ffi.rs
+++ b/src/platform_impl/macos/ffi.rs
@@ -163,9 +163,18 @@ pub const IO8BitOverlayPixels: &str = "O8";
 pub type CGWindowLevel = i32;
 pub type CGDisplayModeRef = *mut c_void;
 
+// `CGDisplayCreateUUIDFromDisplayID` comes from the `ColorSync` framework.
+// However, that framework was only introduced "publicly" in macOS 10.13.
+//
+// Since we want to support older versions, we can't link to `ColorSync`
+// directly. Fortunately, it has always been available as a subframework of
+// `ApplicationServices`, see:
+// https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/OSX_Technology_Overview/SystemFrameworks/SystemFrameworks.html#//apple_ref/doc/uid/TP40001067-CH210-BBCFFIEG
+//
+// TODO: Remove the WINIT_LINK_COLORSYNC hack, it is probably not needed.
 #[cfg_attr(
     not(use_colorsync_cgdisplaycreateuuidfromdisplayid),
-    link(name = "CoreGraphics", kind = "framework")
+    link(name = "ApplicationServices", kind = "framework")
 )]
 #[cfg_attr(
     use_colorsync_cgdisplaycreateuuidfromdisplayid,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- Fix linking to the `ColorSync` framework on macOS 10.7, and in newer Rust versions.
 - On Web, implement cursor grabbing through the pointer lock API.
 - On X11, add mappings for numpad comma, numpad enter, numlock and pause.
 - On macOS, fix Pinyin IME input by reverting a change that intended to improve IME.


madsmtm:
Nope, I was wrong, this doesn't fix it

madsmtm:
I've investigated this some more, and I'm pretty sure I've found the correct way of doing it.

@Imberflur and @ArturKovacs, sorry for pinging you before, but this is now ready to be tested.

Also pinging @ryanisaacg, don't know what macOS version you have but the more testers of this the merrier :wink:

ArturKovacs:
Unfortunately I won't have access to a macOS device until January. Feel free to ping me again at that time if this is still open then.

madsmtm:
There is kind of a deadline on this, `winit` will break for macOS users in Rust 1.58, and I would really like to get this fix rolled out a while before that (and perhaps backported to some of the older, yet still quite widely used `winit` versions?)

madsmtm:
To get this merged quicker, I took out the `WINIT_LINK_COLORSYNC` changes, will follow up with those in a separate PR.

src/platform_impl/macos/ffi.rs
```diff
@@ -165,9 +165,18 @@ pub const IO8BitOverlayPixels: &str = "O8";
...
+// Since we want to support older versions, we can't link to `ColorSync`
+// directly. Fortunately, it has always been available as a subframework of
+// `ApplicationServices`, see:
+// https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/OSX_Technology_Overview/SystemFrameworks/SystemFrameworks.html#//apple_ref/doc/uid/TP40001067-CH210-BBCFFIEG
+//
+// TODO: Remove the WINIT_LINK_COLORSYNC hack, it is probably not needed.
 #[cfg_attr(
     not(use_colorsync_cgdisplaycreateuuidfromdisplayid),
-    link(name = "CoreGraphics", kind = "framework")
+    link(name = "ApplicationServices", kind = "framework")
```
madsmtm, line 179: Note that a bit further down in this file we're linking against `CoreGraphics`, so the only change is that we now also link to `ApplicationServices`.

Imberflur:
@madsmtm I tested this with the docker setup and it links fine :+1:  (although we aren't using that setup anymore currently)

icefoxen:
Deferring to new ggez devs: @PSteinhaus , @nobbele and @IGBC .

src/platform_impl/macos/ffi.rs
A-Helberg: Tested on MBP running macOS 12.0.1 and it works 👍

hecrj:
@ArturKovacs Thanks for the heads up!

Indeed, there are a bunch of projects I maintain that have failing CI with the `beta` toolchain. Ideally, a fix should be merged and released before the new `stable` release.

madsmtm:
If we could get this tested by someone with a macOS version < 10.13 that would be nice, but this PR very probably works as-is (the previous behaviour utilized undocumented behaviour on older platforms, while linking to `ApplicationServices` matches what the docs say).

madsmtm:
Relatedly, an update to the people following this issue:

The changes to `rustc` [have been reverted](https://github.com/rust-lang/rust/pull/91870), and the revert will (very likely) be backported to `beta`. So the rush for getting this rolled out before 1.58 is gone (though we should still fix the issue!)

ArturKovacs:
I have macOS 10.15 so it's not that easy for me to test on macOS < 10.13

However given that my machine is from mid 2012, I *think* I can reinstall it to the old OS version. That's a bit intrusive but I can do it if it's reeeealy needed.

hkratz:
I have an old Macbook with 10.10 lying around. Not being familiar with winit what would be the best way to test it? Just `cargo build` / `cargo test`?

madsmtm:
Just something where the linking is triggered. Try:
```sh
cargo clean
MACOSX_DEPLOYMENT_TARGET=10.7 cargo build --example window
cargo clean
cargo build --example window
```

hkratz:
Looks good:
```
$ git remote -v
origin	git@github.com:madsmtm/winit.git (fetch)
origin	git@github.com:madsmtm/winit.git (push)
$ git status
On branch fix-colorsync-linking
Your branch is up to date with 'origin/fix-colorsync-linking'.

nothing to commit, working tree clean
$ sw_vers
ProductName:	Mac OS X
ProductVersion:	10.10.5
BuildVersion:	14F2511
$ rustc --version
rustc 1.57.0 (f1edd0429 2021-11-29)
$ cargo clean
$ cargo build --example window --quiet
$ MACOSX_DEPLOYMENT_TARGET=10.7 cargo build --example window --quiet
$ cargo clean
$ cargo build --example window --quiet
$ 
```

madsmtm:
Wonderful, thanks a lot @hkratz!

Just out of interest, could I get you to do:
```sh
cargo clean
MACOSX_DEPLOYMENT_TARGET=10.7 cargo run --example window
```

And just verify that a window actually opens? I think it's rare that `winit` is actually used on older machines, so it would be nice to hear that working

hkratz:
> ```shell
> cargo clean
> MACOSX_DEPLOYMENT_TARGET=10.7 cargo run --example window
> ```

@madsmtm It shows a small window and produces lots of logging output after adding a workaround for https://github.com/borntyping/rust-simple_logger/issues/43

madsmtm:
Wonderful, thanks!

madsmtm:
Well, I think this has had enough eyes on it

PR #1932, opened 2021-05-07
 New keyboard API for Linux

maroider:
- [ ] Tested on all platforms changed
- [x] Compilation warnings were addressed
- [x] `cargo fmt` has been run on this branch
- [x] `cargo doc` builds successfully
- [ ] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users
- [ ] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior
- [x] ~~Created or updated an example program if it would help users understand this functionality~~
- [ ] Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented

This implements #753 for Linux (I know that issue number by heart now, lol). Much like #1888, this PR is lacking in some areas, but unlike the aforementioned PR, this PR should be usable enough to get some useful feedback and testing done.

# TODOs

- [x] Make IMEs work as expected. I have 0 experience with IMEs, so this part may be completely incorrect.
- [x] Do another pass on `keymap.rs`, as there are things there I suspect can be mapped, and some things that may have been mapped incorrectly.
- [x] Handle keyboard layout changes
- [x] Should Ctrl override any attempts to compose a character (with dead keys)? `gnome-terminal` agrees, but `konsole` disagrees. Should it be configurable instead?
- [x] Ensure control characters only occur when we expect them. This is mostly implemented, but there seems to be a couple edge cases which I'm not inclined to fix today.
- [ ] Make sure it runs fine under multiseat configurations
    - [ ] Wayland
    - [ ] X11
- [x] Various `TODO` comments in the code.
    - [x] `src/platform_impl/linux/common/keymap.rs`
    - [ ] `src/platform_impl/linux/common/xkb_state.rs`
    - [x] `src/platform_impl/linux/x11/event_processor.rs`

cc @garasubo @jamadazi

Issue #1495, opened 2020-03-05
Improving the stability


ArturKovacs:
### Why?

Winit is big. There are a bunch of software, and most importantly, people relying on it but there still have been rather serious regressions introduced recently. The following ones practically render winit programs useless under Windows and XMonad.

#1429
#1461
#1474

The first two both refer to a group of issues that affect the Windows platform. These are present in winit 0.21.0 meaning that everyone who targets Windows must skip this version of winit. But no matter how impactful, those issues still slip through the cracks and may make it into multiple crates the same way they made it into winit, glutin and glium.

The pourpose of this current issue is to nudge the project towards more stable releases.

### How?

This could be done by introducing more automated tests. Some rudimental tests could be done with plain old cargo tests, however I think tests that give keyboard and mouse input would be much more useful. 

Here are a few tools that could help us do that

https://github.com/enigo-rs/enigo
https://pypi.org/project/pynput/
https://pypi.org/project/PyAutoGUI/

None of these seem to support Wayland but I don't know what tha means for us. PyAutoGUI seems to be the most mature but I only made a quick search so there could be better alternatives out there.

Currently the way I imagine it is that each test has two executables:
1. a rust binary that's built using winit 
2. and a separate tester executable that launches the rust program.

The tester program gives the rust app some simulated mouse and keyboard input and the rust program, in turn, gives some output from which the tester determines whether the rust program passed the test.

For all of this to work there needs to be a testing environment that supports gui tests. I can see that we are currently using Travis and Appveyor. Do those support running graphical programs that expect and simulate keyboard and mouse input?

With all this said I would be up for writing some tests and ensuring that they behave properly on Windows at least, but of course I would need help with the rest.

Osspial:
Thanks for opening this issue! I'm definitely interested in the idea of having automated tests - I've definitely noticed that more and more issues are slipping through the cracks of PRs, and that's not a sustainable situation for us.

We actually moved over to GitHub actions for CI recently, so we're not using Travis or Appveyor any more. I don't know if it supports GUI environments, so I've opened #1498 to test things out and get a sense of where that stands.

ArturKovacs:
Thanks a lot for checking the GUI availability in GitHub actions! I started working on this I'll report back when I have something of relevance.

Issue #5, opened 2016-02-27
Add support for XCB


tomaka:

kvark:
Some machines don't support `VK_KHR_xlib_surface` but do support `VK_KHR_xcb_surface`, so this is a compatibility concern:
```
Instance Extensions     count = 3
        VK_KHR_surface                  : extension revision 25
        VK_KHR_xcb_surface              : extension revision  5
        VK_EXT_debug_report             : extension revision  2
```

kvark:
Found this library as a potential workaround: https://github.com/Aatch/rust-xcb

nicokoch:
Just a heads up for anyone searching a workaround.

The intel linux driver does indeed support xlib, but it has to be enabled at compile time.
You can compile the vulkan driver yourself with `VK_KHR_xlib_surface` enabled by enabling 
`-DVK_USE_PLATFORM_XLIB_KHR`.

For anyone using archlinux, I have submitted a feature request to their bug tracker (https://bugs.archlinux.org/task/50160). It looks like they are working on it (or at least are aware of the problem). For other distros, you can try submitting a bug report to their corresponding bug tracker.

kvark:
Thanks for the info @nicokoch

mikedilger:
With xlib I'm experiencing (in vulkan) get_physical_device_surface_formats_khr() doesn't return until the window is focused.  With xcb (in a separate C++ program  - Sascha Willems demos) this doesn't happen, the window draws irrespective of focus.  I can't be sure the reason for this behavioral difference is xlib vs xcb, but it is looking that way, in which case I prefer xcb and +1 this issue.

mikedilger:
Just fyi:
https://crates.io/crates/ash exposes Vulkan's 
https://docs.rs/ash/0.24.1/ash/vk/types/struct.XcbSurfaceCreateInfoKHR.html
which wants not only the connection (`xcb_connection_t`) but also the window (`xcb_window_t`).  I notice that winit already provides the former with `get_xcb_connection()` but not the latter, so searching I found this issue and now I remember xcb isn't really supported yet.

francesca64:
I don't believe adding a full-fledged XCB backend to be a good idea until winit stabilizes more. We have enough of a struggle keeping the existing backends synchronized.

That said, are you sure that using the return of `get_xlib_window` as your `xcb_window_t` doesn't work? I'd imagine it would work.

mikedilger:
[...]

PR #2088, opened 2021-12-04
Add integration tests

mahkoh:
Note: This patch is based on my work in #1932 and cannot currently be merged. Please ignore all but the ~~last~~ two last commit. That is, ignore everything outside the `it` directory.

This patch adds integration tests than run as a github action. Currently only the X11 backend is being tested. It should not be hard to add support for the wayland backend.

You can see them in action here: https://github.com/mahkoh/winit/runs/4417323203

Part of https://github.com/rust-windowing/winit/issues/1495.


rib:
The tricky thing with porting from xlib to xcb is that xlib does more than just wrap the wire protocol, it also includes various utilities, with the most notable here maybe being the XFilterEvent mechanism that supports IME integration.

It doesn't seem ideal that this series ends up removing IME support.

Since xlib is based on xcb and an xcb connection can be got from xlib I would hope that xcb usage could be introduced incrementally where it's particularly beneficial, e.g. to help avoid synchronous round trips.

rib:
It looks like there's loads of really good work buried in this series - though it's a bit surprising how long of a series of changes this is stacked on top of.

The first commit in this series is apparently two years old :/

I was looking at some small changes in the x11 backend and now I'm worried about how much of a huge backlog of x11 changes there are for winit.

kchibisov:
>It doesn't seem ideal that this series ends up removing IME support.

>Since xlib is based on xcb and an xcb connection can be got from xlib I would hope that xcb usage could be introduced incrementally where it's particularly beneficial, e.g. to help avoid synchronous round trips.

IME is broken on X11 with xlib anyway. And the only way to solve it is to move to xcb.

For more see https://gitlab.freedesktop.org/xorg/lib/libx11/-/merge_requests/127 . I'm not sure there's a way to get it MR any further, so moving away from xlib is the only solution for this particular problem.

If we want to solve IME problem with xcb there's https://crates.io/crates/xcb-imdkit and upstream C library is maintained by fctix developers, so it's kind of battle tested.

mahkoh:
> It looks like there's loads of really good work buried in this series - though it's a bit surprising how long of a series of changes this is stacked on top of.
> 
> The first commit in this series is apparently two years old :/

Like I said in the OP, for this PR only the last two commits are relevant. The remaining commits are part of the linked PR.

rib:
> Like I said in the OP, for this PR only the last two commits are relevant. The remaining commits are part of the linked PR.

Yeah, although I'd seen that this issue was only focused on the final patches, I still initially interpreted "This patch is based on my work in #1932" to mean that this series was a super set of #1932, so when I looked over the series of commits here I was under the impression it was the same series under #1932.

Taking another look at the commits for #1932 I do see now that some things like XInput2 enabling are now being done by @maroider so I probably got the wrong impression about what work is outstanding from looking at the commits here.

Is there another issue where you've maybe discussed more about upstreaming the XCB port - I couldn't find anything, so this was the only place I'd seen that work. I just found this old issue which didn't have any link to your work: https://github.com/rust-windowing/winit/issues/5

rib:
> > It doesn't seem ideal that this series ends up removing IME support.
> 
> > Since xlib is based on xcb and an xcb connection can be got from xlib I would hope that xcb usage could be introduced incrementally where it's particularly beneficial, e.g. to help avoid synchronous round trips.
> 
> IME is broken on X11 with xlib anyway. And the only way to solve it is to move to xcb.
> 
> For more see https://gitlab.freedesktop.org/xorg/lib/libx11/-/merge_requests/127 . I'm not sure there's a way to get it MR any further, so moving away from xlib is the only solution for this particular problem.
> 
> If we want to solve IME problem with xcb there's https://crates.io/crates/xcb-imdkit and upstream C library is maintained by fctix developers, so it's kind of battle tested.

Yeah, this makes sense. Tbh, when I was first skimming the patch series it just stood out that the series was removing IME support and I hadn't initially seen much discussion of that so I was concerned this change might potentially be for the wrong reasons (just to remove xlib for the sake of it even if it regresses features).

I'd also recently come across PR https://github.com/rust-windowing/winit/pull/2182 which gave me an initial impression their might be users actively using XIM based IMEs with winit considering going to the effort to fix issues with it. Actually though it looks like it might be more the opposite - they are trying to workaround the way in which key repeats are synthesized with im=ibus, and arguably that problem would also be solved by removing the XIM support from the current backend.

notgull:
@mahkoh Do you have plans to rebase this work on the new XCB-based `x11rb` backend? I think that integration tests in this way would be very valuable. Otherwise I will close this PR for triage purposes.

mahkoh:
I haven't used X11 in 4 years.

PR #96, opened 2016-11-28
Child window

k-brac:
Here is my attempt to implement #93 for windows.

I am not sure it is the way to go (I changed some members visibility).

I have added an example as a proof of concept.

PR #2096, opened 2021-12-14
[X11] Create windows with a parent window

t-sin:
This PR can create windows with a user-specified parent window. It's intended to realize what discussed in issue #159 for X11.

EDITED: ~~For compatibility, this introduces a new method `new_x11_any_thread_with_root()` to create such windows. I considered to change the signature of `new_x11_any_thread()`, it takes a parent window ID, but I cannot decide that so simply create a new method. I want to discuss about this point.~~

The way above was an wrong way. For now this PR adds `with_x11_parent()` method to `WindowBuilder` by importing `winit::platform::unix::WindowBuilderExtUnix`, we can set X11 window ID as a parent with it. It seems good because it's similar way in supporting child window in Windows (#96).

~~By the way this PR is for my VST3 plugin using [vst3-sys](https://github.com/RustAudio/vst3-sys) so I don't have any ideas to test this PR without VST3 codes. I want some advice about testing.~~ This PR adds example code and it works.


t-sin:
This PR cannot create a child window because it fails arround [here](https://github.com/t-sin/winit/blob/3e8e991be994b4a127cf259debf9c5bd1d4409c6/src/platform_impl/linux/x11/util/input.rs#L113). I should consider the xinput2. Please wait.

t-sin:
Some problem when this PR created are resolved. It's ready for review.

PR #2098, opened 2021-12-16
Makes wayland app_id function signature more ergonomic

benbrittain:


kchibisov (changes requested):
Could you squash and rebase it?

kchibisov (approved):
Thanks

PR #2100, opened 2021-12-17
Add exit code to `ControlFlow::Exit`

MultisampledNight:
This adds a value to the `Exit` variant of `ControlFlow`, which describes the exit code when the program shuts down and `EventLoop::run` has to exit the process manually through `std::process::exit` in order to have the function return `!`. Also, this represents a breaking change and would probably fit for the `0.27` release, in case it'll ever happen.

# Background

Previously, `0` was hardcoded as exit value, which caused people having to use either `EventLoop::run_return` using `std::process::exit` afterwards, which would introduce an unneeded overhead for quite common behavior, or `std::process::exit` directly in the closure taken in by `EventLoop::run`, which would not run any `Drop` implementations on structs captured by the closure, in order to have an exit code accompanied by an error in the console or similar. In this PR we introduce a value to `ControlFlow::Exit`, used anyways if the program wants to exit properly, by which the user can directly control the exit value if needed.

# Why a draft?

I'd like to get some review on the mess I created first, such as my way to get the final control flow value on Mac OS by returning it from `AppState::exit`, or any other cases where I messed up the original meaning of the code. Finally, I'm happy about any other criticism or advice, too, I'm not coding in Rust for too long.

  * [X] Linux (both X11 and Wayland)
  * [X] Windows (through Wine)
  * [ ] Mac OS
  * [ ] Android
  * [ ] iOS
  * [ ] wasm32
- [X] ~~Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented~~


examples/exit_code.rs
```diff
@@ -0,0 +1,57 @@
...
+                VirtualKeyCode::Key7 => *flow = ControlFlow::Exit(7),
+                VirtualKeyCode::Key8 => *flow = ControlFlow::Exit(8),
+                VirtualKeyCode::Key9 => *flow = ControlFlow::Exit(9),
+                VirtualKeyCode::Key0 => *flow = ControlFlow::Exit(0),
+                _ => (),
+            },
+            Event::WindowEvent {
+                event: WindowEvent::CloseRequested,
+                ..
+            } => *flow = ControlFlow::Exit(47),
```
filnet, line 53: should be 42 ?

MultisampledNight: No, it's a reference to something else, a game.

filnet: Yep, works too :)

src/event_loop.rs
```diff
@@ -88,7 +88,10 @@ pub enum ControlFlow {
     /// Send a `LoopDestroyed` event and stop the event loop. This variant is *sticky* - once set,
     /// `control_flow` cannot be changed from `Exit`, and any future attempts to do so will result
     /// in the `control_flow` parameter being reset to `Exit`.
-    Exit,
+    /// The contained number will be used as exit code, if the platform supports that (this
+    /// _excludes_ ios, android and wasm32), it's an [`u8`] to prevent some weird surprises as
+    /// described in [`std::process::exit`].
+    Exit(u8),
```
aloucks, line 94: This should be an `i32`. It shouldn't be winit's decision to impose platform limitations. If anything, I would modify this to say that the value will be passed to `std::process:exit` and link to https://doc.rust-lang.org/std/process/fn.exit.html#platform-specific-behavior to describe platform specific behavior.

MultisampledNight: I thought this could avoid nasty surprises with negative values on unix-alikes, since due to two's complement `-47` would end up as `214` (as it did on my machine), but fair enough if that's not winits problem. I'll commit now, but I'm unable to push anything until Tuesday.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Unreleased
 
 - On X11, add mappings for numpad comma, numpad enter, numlock and pause.
+- Add a value to the `Exit` variant of `ControlFlow` which controls the exit code after running.
```
madsmtm, line 4: Please prefix these kind of changes with `**Breaking:**` like below

madsmtm (changes requested):
Nice! I agree that this is functionality that `winit` should expose.

examples/exit_code.rs
```diff
@@ -0,0 +1,57 @@
+use {
```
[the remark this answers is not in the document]
MultisampledNight: Fair enough.

examples/control_flow.rs
```diff
@@ -88,7 +88,7 @@ fn main() {
                     window.request_redraw();
                 }
                 if close_requested {
-                    *control_flow = ControlFlow::Exit;
+                    *control_flow = ControlFlow::Exit(0);
```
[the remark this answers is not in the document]
MultisampledNight: That indeed does make sense, this adds a direct connection to "success" and `0` instead of letting an user who never worked with exit codes in the dark. I'm not sure if this would be better as an associated constant though, a function like that looks it would actually do something more special imo.

src/event_loop.rs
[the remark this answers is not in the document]
MultisampledNight: Variant 2 looks better to me, since variant 1 lets in my eyes

1. fall `From` implementations out-of-date (if a dependency upgrades winit with this, sees `#[non_exhaustive]`, adds a `_ => Self::Unknown` arm, and upgrades winit later on with another new variant in there, then this could be easily forgotten and whoops, suddently we have something that impls `From<ControlFlow>` but doesn't map the new variant correctly).
2. seem `Exit` like it wouldn't exit without any code at all, which isn't true.

src/platform_impl/linux/mod.rs
```diff
@@ -743,8 +743,9 @@ fn sticky_exit_callback<T, F>(
 {
     // make ControlFlow::Exit sticky by providing a dummy
     // control flow reference if it is already Exit.
-    let mut dummy = ControlFlow::Exit;
-    let cf = if *control_flow == ControlFlow::Exit {
+    let mut dummy;
+    let cf = if let ControlFlow::Exit(code) = *control_flow {
+        dummy = ControlFlow::Exit(code);
```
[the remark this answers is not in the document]
MultisampledNight: Well, then I might just use the windows version.

src/event_loop.rs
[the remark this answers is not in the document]
maroider: I prefer the first alternative, as I don't think anyone's really `match`-ing on `ControlFlow`.

>2. seem `Exit` like it would exit without any code at all, which isn't true.

I don't think this is an issue. If someone cares about the exit code, then I think we can trust them to assume that `0` will be used in the absence of any errors (like panics). In any case, we'd want to document exactly what `ControlFlow::Exit` does.

src/platform_impl/windows/event_loop.rs
```diff
@@ -234,7 +236,15 @@ impl<T: 'static> EventLoop<T> {
         unsafe {
             runner.loop_destroyed();
         }
+        let exit_code = if let ControlFlow::Exit(code) = runner.control_flow() {
```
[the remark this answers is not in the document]
MultisampledNight: That's possible for sure, I was... quite lazy when implementing that as I had a hard time understanding all the winapi functions.

src/platform_impl/macos/app_state.rs
```diff
@@ -99,9 +99,8 @@ impl<T> EventHandler for EventLoopHandler<T> {
     fn handle_nonuser_event(&mut self, event: Event<'_, Never>, control_flow: &mut ControlFlow) {
         self.with_callback(|this, mut callback| {
             (callback)(event.userify(), &this.window_target, control_flow);
-            this.will_exit |= *control_flow == ControlFlow::Exit;
-            if this.will_exit {
-                *control_flow = ControlFlow::Exit;
+            if let ControlFlow::Exit(_) = *control_flow {
+                this.will_exit = true;
```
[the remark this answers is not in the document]
MultisampledNight: Indeed. Didn't think about this existing for stickyness.

src/event_loop.rs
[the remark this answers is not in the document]
MultisampledNight: Then I'm okay with the `ExitWithCode` variant co-existing with the `Exit` variant, even though I'm not too sure about `#[non_exhaustive]`. What exact advantage does that add if no one seriously matches `ControlFlow`?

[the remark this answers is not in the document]
maroider: It's mostly to make (potential) future additions be a non-breaking change.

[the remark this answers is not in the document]
madsmtm: > I prefer the first alternative, as I don't think anyone's really `match`-ing on `ControlFlow`.

Then the associated constant would work just as well, in terms of breakage, right?

I think my preference would be the associated constant, since our code won't have to handle two enum variants that does effectively the same thing.

src/platform/run_return.rs
```diff
@@ -33,7 +34,12 @@ pub trait EventLoopExtRunReturn {
     /// underlying OS APIs, which cannot be hidden by `winit` without severe stability repercussions.
     ///
     /// You are strongly encouraged to use `run`, unless the use of this is absolutely necessary.
-    fn run_return<F>(&mut self, event_handler: F)
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **Unix-alikes** (**X11** or **Wayland**): This function returns `1` upon disconnection from
+    ///   the display server.
```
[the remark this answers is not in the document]
MultisampledNight: It might be, but I'm not too sure how to word that.

```rust
    /// - **Unix-alikes** (**X11** or **Wayland**): The function exits with exit code 1 if the display server disconnects.
```
sounds a bit weird imo, since the function seemingly never "exits".

examples/control_flow.rs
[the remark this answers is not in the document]
MultisampledNight: Now with `Exit` being an alias for `ExitWithCode(0)`, I don't know how useful this is. Would a `ExitFailure` (can't think of anything more creative) be more confusing than useful for this?

CHANGELOG.md
MultisampledNight: Done, is the current version fine?

MultisampledNight:
I think I'm done with applying most of the things noted by the reviews, huge thanks to everyone who invested their time into this. Could anyone take an (eventually final) look at this? Take your time I guess, there is no reason to haste.

src/platform_impl/macos/app_state.rs
```diff
@@ -98,25 +97,25 @@ impl<T> Debug for EventLoopHandler<T> {
 impl<T> EventHandler for EventLoopHandler<T> {
     fn handle_nonuser_event(&mut self, event: Event<'_, Never>, control_flow: &mut ControlFlow) {
         self.with_callback(|this, mut callback| {
-            (callback)(event.userify(), &this.window_target, control_flow);
-            this.will_exit |= *control_flow == ControlFlow::Exit;
-            if this.will_exit {
-                *control_flow = ControlFlow::Exit;
+            if let ControlFlow::ExitWithCode(code) = *control_flow {
+                let dummy = &mut ControlFlow::ExitWithCode(code);
```
madsmtm, line 101: Nit: Do you need the intermediate "dummy" value here?

examples/control_flow.rs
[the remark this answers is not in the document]
madsmtm: Nah, I don't think `ExitFailure` or similar is needed

CHANGELOG.md
madsmtm: Maybe make it clearer that because of the added constant, this is in practice unlikely to cause breakage?

madsmtm (approved):
This looks fine, only have minor unimportant nits

src/platform_impl/macos/app_state.rs
MultisampledNight: Not necessarily, no. But putting it inline would cause `rustfmt` to split the callback into multiple lines, which looks even worse imo.

src/event_loop.rs
```diff
@@ -88,7 +88,24 @@ pub enum ControlFlow {
     /// Send a `LoopDestroyed` event and stop the event loop. This variant is *sticky* - once set,
     /// `control_flow` cannot be changed from `Exit`, and any future attempts to do so will result
     /// in the `control_flow` parameter being reset to `Exit`.
```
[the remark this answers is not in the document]
MultisampledNight: It... should, indeed.

src/platform_impl/macos/app_state.rs
madsmtm: Ah, fair. Could we call it `sticky` then?

MultisampledNight: It isn't sticky at all since it only persists for that one callback, but sure, why not.

src/platform/run_return.rs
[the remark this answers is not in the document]
MultisampledNight: Pretty fine, yeah. Thanks!

src/platform_impl/macos/app_state.rs
madsmtm: Nah, true, `dummy` is fine

src/event_loop.rs
```diff
@@ -86,9 +86,29 @@ pub enum ControlFlow {
...
+    /// `control_flow` cannot be changed from `ExitWithCode`, and any future attempts to do so will
+    /// result in the `control_flow` parameter being reset to `ExitWithCode`.
+    ///
+    /// The contained number will be used as exit code. The [`Exit`] constant is a shortcut for this
+    /// with exit code 0.
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **Android**, **iOS**, **WASM**: The supplied exit code is unused.
+    /// - **Unix-alikes** (**Wayland** or **X11**): On some platforms, only the 8 least significant
```
maroider, line 98:
```suggestion
    /// - **Unix**: On most Unix-like platforms, only the 8 least significant
```

I think sticking close to `std`'s documentation is best here.

src/event_loop.rs
```diff
@@ -145,6 +165,11 @@ impl<T> EventLoop<T> {
     ///
     /// Any values not passed to this function will *not* be dropped.
     ///
+    /// ## Platform-specific
+    ///
+    /// - **Unix-alikes** (**X11** or **Wayland**): The program terminates with exit code 1 in case
+    ///   the display server disconnects.
```
maroider, line 171:
```suggestion
    /// - **X11**, **Wayland**: The program terminates with exit code 1 in case
    ///   the display server disconnects.
```

src/event_loop.rs
```diff
@@ -86,9 +86,29 @@ pub enum ControlFlow {
...
-    Exit,
+    /// `control_flow` cannot be changed from `ExitWithCode`, and any future attempts to do so will
+    /// result in the `control_flow` parameter being reset to `ExitWithCode`.
+    ///
+    /// The contained number will be used as exit code. The [`Exit`] constant is a shortcut for this
+    /// with exit code 0.
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **Android**, **iOS**, **WASM**: The supplied exit code is unused.
```
maroider, line 97:
```suggestion
    /// - **Android / iOS / WASM**: The supplied exit code is unused.
```

` / `-separation is the usual way we list a quirk that affects multiple platforms under `Platform-specific` headings.

maroider (commented):
This looks mostly OK to me.

I have no nits implementation-wise.

maroider (approved):

ArturKovacs:
I'm not a fan of the `const Exit`. I think that optimally we should completely remove that sometime in the future, mostly because it doesn't follow the naming convention.

I think we should add the [deprecated](https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute) attribute to the `const Exit` with the note that says something like "use `ExitWithCode(0)` instead".

MultisampledNight:
Generally, I agree with this being a constant falling out of order, which isn't good. But adding a new constant _and_ marking it in the same release as deprecated might seem a bit confusing, thus we'd have to adjust _all_ examples again in order to have them stay "*this* is how you're supposed to use winit".

Would something like this be fine though, in case everyone here agrees with marking it as deprecated?

```rust
#[deprecated = "this is actually a constant now, consider using ControlFlow::ExitWithCode instead which this aliases to"]
const Exit: ... = ...;
```

madsmtm:
I agree with you in general, @ArturKovacs, but we must remember that this is used in literally _every_ `winit` application in existence. While deprecating it is certainly better than removing it, I still think it's overly disruptive (again _every_ `winit` user will see this warning when upgrading) for a feature that is very rarely used.

If we vastly change the design then I'd love to have it removed or renamed to `ControlFlow::EXIT`, but until then I'd like to keep it as-is.

Anyhow, if we want to deprecate it we can do so in another PR, feel free to open one if you feel strongly about this (or just want us to remember doing it at some point).

ArturKovacs:
I don't feel too strongly about this, so feel free to merge this as is.

src/platform_impl/linux/wayland/event_loop/mod.rs
```diff
@@ -434,17 +436,17 @@ impl<T: 'static> EventLoop<T> {
...
 
             match control_flow {
-                ControlFlow::Exit => break,
+                ControlFlow::ExitWithCode(code) => break code,
                 ControlFlow::Poll => {
                     // Non-blocking dispatch.
                     let timeout = Duration::from_millis(0);
                     if self.loop_dispatch(Some(timeout)).is_err() {
-                        break;
+                        break 1;
```
kchibisov, line 449: Please, return last os error code since the only way we could fail due dispatch if something is wrong with polling it.  And I think there's a way to get code from error enum itself, but I'm not sure if it's still there, but it should.

kchibisov (changes requested):

src/platform_impl/linux/wayland/event_loop/mod.rs
MultisampledNight: I'll take a look at it, thanks!

kchibisov: Yeah, it should. Since it's io::Error under the hood, so just getting code from it should work.

MultisampledNight: From what I see in https://doc.rust-lang.org/stable/std/io/struct.Error.html#method.raw_os_error, the underlying OS error code can only be fetched by using `std::io::Error::last_os_error`, whose usages I can neither find in winit (well, on windows, but that's not relevant here) nor in `calloop`s code. Was calling `std::io::Error::last_os_error` to get an error to examine for the exit code in your mind or did you mean something else?

kchibisov: No, I mean that `self.loop_dispatch` is returning `std::io::Result` from which you can get `std::io::Error` and on which you can call `https://doc.rust-lang.org/std/io/struct.Error.html#method.raw_os_error`, since the `Error` it's using under the hood is `nix` Error, but it's being built from `errno` value and its from impl is using `from_raw_os_error` when converting.

MultisampledNight: Fair enough. Thanks for the explanation!

kchibisov: >nor in calloops code, so the error from self.loop_dispatch likely does not contain the error code. 

It does contain it, since that's what `epoll_wait` returns in calloop.

MultisampledNight: There we go, are the changes fine for you?

kchibisov (approved):
Thanks, LGTM.

maroider:
I take it we're ready to merge this now?

MultisampledNight:
Everything's implemented, no objections since half a week, 3 approvals and no outstanding questions (except that deprecated `Exit` constant), I'd say yes.

ArturKovacs (approved):

msiglreith (approved):

madsmtm:
Thanks a lot @MultisampledNight, I know it's taken a while, but that's how it is when modifying internals that each maintainer has to approve, and public API that we all have to agree on 😉 (and I mean, this one has progressed significantly faster than other `winit` changes, so that's a plus)!

PR #2101, opened 2021-12-18
[Windows] Reverse horizonal scroll direction in `WindowEvent::MouseWheel`

### LINKED ISSUE #2099: Horizontal Scrolling Backwards on Windows
**Issue**
Horizonal and vertical scrolling directions do not match.

**Expected Behavior**
Both x/y should behave the same, e.g. both directions are "natural" scrolling on the touchpad.

**Observed Behavior**
Vertical scrolling respects OS and is in the "natural" direction. Horizontal scrolling is inverted and does not follow the "natural" scroll direction. First noticed when using native builds of https://github.com/emilk/egui/issues/356.

**Reproducing the Bug**
I've created a minimal reproducible winit example here: https://github.com/aevyrie/winit_scroll_test.

aevyrie:

Fixes #2099
Addresses part of downstream issue https://github.com/emilk/egui/issues/356

On windows, the horizonal and vertical scrolling directions do not match when running the MRE in #2099. Considering vertical scroll is used in many places already, it seems sensible to swap horizontal scroll deltas on windows, like #1695.

This PR also changes the `mouse_wheel` example to demonstrate scroll behavior. In its current state, it only reports mouse device events, which don't seem to report horizontal scrolling on windows. I've changed this to ` WindowEvent::MouseWheel`, which allows you to test x and y scrolling - which seems to be the intent of the example considering it already allows repositioning the window in x and y. I've also flipped the delta applied to the window, so it moves in the direction the page you are scrolling would move. This makes it easier to verify "natural" scrolling is working as expected.

### DIFF
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -1200,7 +1200,7 @@ unsafe fn public_window_callback_inner<T: 'static>(
 
             let value = (wparam >> 16) as i16;
             let value = value as i32;
-            let value = value as f32 / winuser::WHEEL_DELTA as f32;
+            let value = -value as f32 / winuser::WHEEL_DELTA as f32;
 
             update_modifiers(window, userdata);
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 - On X11, add mappings for numpad comma, numpad enter, numlock and pause.
+- On Windows, reversed horizontal scrolling in `WindowEvent::MouseWheel` to match the direction of vertical scrolling.
 
 # 0.26.0 (2021-12-01)
 
--- a/examples/mouse_wheel.rs
+++ b/examples/mouse_wheel.rs
@@ -1,6 +1,6 @@
 use simple_logger::SimpleLogger;
 use winit::{
-    event::{DeviceEvent, Event, WindowEvent},
+    event::{Event, WindowEvent},
     event_loop::{ControlFlow, EventLoop},
     window::WindowBuilder,
 };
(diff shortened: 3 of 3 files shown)


src/platform_impl/windows/event_loop.rs
```diff
@@ -1200,7 +1200,7 @@ unsafe fn public_window_callback_inner<T: 'static>(
 
             let value = (wparam >> 16) as i16;
             let value = value as i32;
-            let value = value as f32 / winuser::WHEEL_DELTA as f32;
+            let value = -value as f32 / winuser::WHEEL_DELTA as f32;
```
aloucks, line 1203: https://docs.microsoft.com/en-us/windows/win32/api/windowsx/nf-windowsx-get_y_lparam

aevyrie: @aloucks could you expand on the relevance of this please? I'm not familiar with the win32 api.

aevyrie: Is the suggestion to use `lparam` instead of `wparam`?

aloucks: I think I misread part of https://docs.microsoft.com/en-us/windows/win32/inputdev/wm-mousehwheel and thought the negative value might be related. I think you can disregard my previous comment.

Have you checked to see how GLFW treats the mouse wheel direction? I think we should just use that prior art as reference and have the same behavior.

parasyte: GLFW reference: https://github.com/glfw/glfw/blob/53d86c64d709ff52886580d338d9b3b2b1f27266/src/win32_window.c#L935-L941

aevyrie: > GLFW reference: https://github.com/glfw/glfw/blob/53d86c64d709ff52886580d338d9b3b2b1f27266/src/win32_window.c#L935-L941

Awesome, if I read this correctly, than this PR follows GLFW prior art.

emilk:
I've cherry-picked the changes in this PR into https://github.com/rust-windowing/winit/pull/2105, so I suggest we close this PR in favor of that more all-encompassing PR (which also includes updated documentation).

madsmtm:
Agreed, closing in favor of #2105

aevyrie:
Sounds great. Thanks!

PR #2105, opened 2021-12-20
Reverse horizontal scroll direction

emilk:
**EDIT**: previously the PR only changed the direction on Mac, now I aim to changed it for all platforms.

With this PR, positive scroll values indicate that the content (that which is being moved when scrolling) should move RIGHT and DOWN, i.e. following the common GUI coordinates (used by winit elsewhere) where the X axis increases to the right, and the Y axis increases downwards.

With this PR we now match what GLFW does: https://github.com/glfw/glfw/blob/53d86c64d709ff52886580d338d9b3b2b1f27266/src/cocoa_window.m#L614:L627 https://github.com/glfw/glfw/blob/53d86c64d709ff52886580d338d9b3b2b1f27266/src/win32_window.c#L935-L941

This will remove the need for a workaround in my egui crate at https://github.com/emilk/egui/blob/master/egui-winit/src/lib.rs#L457

This is a breaking change to make winit coordinates consistent. It also adds documentation so winit users actually know what the scroll values mean.


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Unreleased
 
 - On X11, add mappings for numpad comma, numpad enter, numlock and pause.
+- On Mac, reversed horizontal scrolling in `WindowEvent::MouseWheel` to match the direction of vertical scrolling.
```
madsmtm, line 4: This is arguably a breaking change, perhaps mark it as such (using `**Breaking:**`) to be safe?

madsmtm (commented):
Linking the previous issue https://github.com/rust-windowing/winit/issues/1695 and PR https://github.com/rust-windowing/winit/pull/1696 that introduced the current behaviour.

Tagging author of that PR @michaelkirk, I don't feel experienced enough to either approve or disapprove this change.

madsmtm:
Sorry, didn't see the linked windows PR #2101. Does it make sense to merge this before that, or should they really be done together?

And do you know how the other platforms fare in this?

michaelkirk:
It was a while ago that I worked on this, so unfortunately most of the details are now lost on me. I'll say though that before this change, the scrolling behavior in my apps matches other system apps (e.g. like Safari).

So that after this change, horizontal scrolling in my apps becomes undesirably reversed. 

<details>
<summary>e.g. if you want to try one of the applications I work on...</summary>
<pre>
git clone https://github.com/a-b-street/abstreet.git
cd abstreet
cargo run --bin widgetry_demo
</pre>

Shrinking the window horizontally narrow enough should allow you to scroll the overflow horizontally:

https://user-images.githubusercontent.com/217057/147866502-eab8cc28-5b16-423f-ba5a-381da42b8186.mp4


</details>

---

So if this change is working for people, there must be something wrong with how I'm handling winit events, which I'm very willing to except, but haven't found my mistake yet.

I'm aware of egui, but I haven't personally used it. I was trying to test out horizontal scrolling in egui, but couldn't get the feature to work. Can someone more familiar with egui point me to an example?

Here's some things I tried to get horizontal scrolling to occur in egui, but none of them seemed to induce horizontal scrollability: https://github.com/michaelkirk/egui/tree/mkirk/hscroll-demo

parasyte:
> I was trying to test out horizontal scrolling in egui, but couldn't get the feature to work. Can someone more familiar with egui point me to an example?

The `egui_demo_app` has a lot of places where horizontal scroll can be used:

- "Code Example" window; shrink the window so the scroll bar appears
- "Context menu"; the plot is scrollable in both directions
- "Plot"; this plot is also scrollable in both directions
- "Widget Gallery"; has another scrollable plot
- "Color" tab; resize the entire app window and you can get a horizontal scrollbar to show here

I just personally tested this, and horizontal scroll works appropriately on Windows 11 (with a tilt wheel) and macOS (trackpad) despite this PR and #2101 because `egui-winit` reverses the horizontal scroll direction on both platforms: https://github.com/emilk/egui/blob/b1fd6a44e82c8cde7fab0b952851d2d2fd785349/egui-winit/src/lib.rs#L456-L463

msiglreith:
As mentioned in the other thread: it seems to me (but I can't test it..) that the current behavior follows the documented coordinate system for the scroll delta values https://docs.rs/winit/latest/winit/event/enum.MouseScrollDelta.html#variant.LineDelta (up-right), while the more common coordinate system appears up-left/down-right which also aligns more with the window coordinates.

If that's the case it should be treated as breaking change and the documentation being updated as well.

emilk:
The common GUI coordinates have the origin (0,0) in the top left corner, and having X and Y increase to the right and down. So if I move my fingers on a trackpad down and to the right, I should get positive deltas. With this PR, I get just that on my Mac (with the default "natural" scrolling, i.e. the same scrolling as on a touch screen).

emilk:
I've updated the example, documentation, and made this a breaking change in the changelog. I've also cherry-picked the commit e9fe74e00da026b89c5447b7747330ab8f6048d9 from https://github.com/rust-windowing/winit/pull/2101

I need volunteers to test the behavior on Linux and web (or instruction on how to test it on web myself).

src/platform_impl/windows/event_loop.rs
```diff
@@ -1200,7 +1200,7 @@ unsafe fn public_window_callback_inner<T: 'static>(
 
             let value = (wparam >> 16) as i16;
             let value = value as i32;
-            let value = value as f32 / winuser::WHEEL_DELTA as f32;
+            let value = -value as f32 / winuser::WHEEL_DELTA as f32;
```
madsmtm, line 1203: Please add a comment, mostly to draw attention to the fact that we're inverting the direction here (that minus is very small, yet quite significant).

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 - On X11, add mappings for numpad comma, numpad enter, numlock and pause.
 - On macOS, fix Pinyin IME input by reverting a change that intended to improve IME.
 - On Windows, fix a crash with transparent windows on Windows 11.
+- **Breaking:**: reverse horizontal scrolling sign in `MouseScrollDelta` to match the direction of vertical scrolling. A positive X value now means moving the content to the right. The meaning of vertical scrolling stays the same: a positive Y values means moving the content down.
```
madsmtm, line 6: Minor nits.

```suggestion
- **Breaking:**: Reverse horizontal scrolling sign in `MouseScrollDelta` to match the direction of vertical scrolling. A positive X value now means moving the content to the right. The meaning of vertical scrolling stays the same: a positive Y value means moving the content down.
```

madsmtm (commented):
`abstreet`'s `widgetry` is handling horizontal and vertical scrolling differently, so the issue is probably on their end (see also https://github.com/a-b-street/abstreet/pull/831).

Tested on macOS, can approve that the `mouse_wheel` example worked weirdly before and now it works as it should.

Agree that we need to ensure this is correct on Linux and Web as well.

madsmtm:
I don't have the setup to test `winit` on the web, but a quick test by comparing the sign of [`WheelEvent.deltaX`](https://developer.mozilla.org/en-US/docs/Web/API/WheelEvent/deltaX) with the sign of `MouseScrollDelta` does suggest that we should be inverting it (like we do with `deltaY` here):

https://github.com/rust-windowing/winit/blob/25ff30ee8c3b07f9bca43ecc60352e3baee8b5d0/src/platform_impl/web/web_sys/event.rs#L51-L52

madsmtm:
@dabreegster tested this in https://github.com/a-b-street/abstreet/pull/831#issuecomment-1011508970, and that suggests that this is also reversed on Linux.

madsmtm:
[Our X11 implementation](https://github.com/rust-windowing/winit/blob/a52f755ce8dfffe53329163364908937c780057d/src/platform_impl/linux/x11/event_processor.rs#L696-L697) is different from [GLFW's](https://github.com/glfw/glfw/blob/53d86c64d709ff52886580d338d9b3b2b1f27266/src/x11_window.c#L1385-L1388).

Our Wayland implementation [here](https://github.com/rust-windowing/winit/blob/a52f755ce8dfffe53329163364908937c780057d/src/platform_impl/linux/wayland/seat/pointer/handlers.rs#L197-L198) and [here](https://github.com/rust-windowing/winit/blob/a52f755ce8dfffe53329163364908937c780057d/src/platform_impl/linux/wayland/seat/pointer/handlers.rs#L220-L221) also differs [here](https://github.com/glfw/glfw/blob/53d86c64d709ff52886580d338d9b3b2b1f27266/src/wl_init.c#L357-L360).

But again, haven't tested any of these, so I might very well be wrong.

emilk:
I've inverted hscroll on all five platforms. I believe they are now all correct and the same.

* Mac: I've tested extensively, and it now matches GLFW
* Windows: no matches GLFW, and matches experiences of egui users
* X11: now matches GLFW
* `web_sys` matches [what I do in egui_web](https://github.com/emilk/egui/blob/master/egui_web/src/lib.rs#L1028:L1029)
* Wayland: untested, but X and Y are now treated consistently, so likely correct

In other words, I think this can be merged!

madsmtm (changes requested):
This is probably missing a change to the X11 backend:
https://github.com/rust-windowing/winit/blob/001fb7ef60efe989b54b07024b660fd07e94afc3/src/platform_impl/linux/x11/event_processor.rs#L780-L783

Otherwise I approve, I'll try to test the WebAssembly backend, but I don't have the setup for testing the Linux implementation, @kchibisov and @maroider?

emilk:
> This is probably missing a change to the X11 backend:

Good catch! I reverted hscroll there too in https://github.com/rust-windowing/winit/pull/2105/commits/7c7953dd1f041a43e292d0b6c34033f4579a4540

examples/mouse_wheel.rs
```diff
@@ -14,29 +14,41 @@ fn main() {
...
 
+    println!(
+        r"
+When using so called 'natural scrolling' (scrolling that acts like on a touch screen), this is what to expect:
+
+Moving your finger downwards on a scroll wheel should make the window move down, and you should see a positive Y scroll value.
+
+When moving fingers on a trackpad down and to the right, you should see positive X and Y deltas, and the window should move down and to the right.
+
+With reverse scrolling, you should the the inverse behavior.
```
msiglreith, line 25:
```suggestion
With reverse scrolling, you should see the inverse behavior.
```

msiglreith (approved):

michaelkirk:
I wanted to check in, since the scope of this PR has changed.

Originally the PR was "invert horizontal scrolling on mac", which by itself would have made winit inconsistent across platforms, which had me perplexed. 

But then windows was inverted, and then X11 and wayland were inverted, and finally web_sys was inverted.

So that with these additional changes, the scope of the PR is now "invert horizontal scrolling in winit" (on every single platform).

A summary of the transformations being done on scroll delta's across platforms:

||before (x,y)| after (x,y) |
|-|-|-|
|wayland|(➕,➖)|(➖,➖)|
|x11|(➕,➖)|(➖,➖)|
|mac|(➖,➕)|(➕,➕)| 
|web_sys|(➕,➖)|(➖,➖)|
|windows|(➕,➖)|(➖,➖)|

It's nice to no longer have mixed signs within a single platform, and the new units arguably do make more sense than the old (positive being "down and to the right").

I just wanted to try to make sure I understood the tradeoff being proposed: The cost of this is a one-time breakage of horizontal scrolling for every existing winit consumer. Right?

madsmtm:
Thank you @michaelkirk for the summary!

Yes, the idea is to change the coordinate-system for scrolling to match the window position coordinate system (explained better by @msiglreith [here](https://github.com/rust-windowing/winit/issues/2099#issuecomment-1003745692)).

And yes, this will break for downstream users, but if they were handling scroll directions properly before (like `aabstreet` is) the fix for them is most likely that they won't have to work with mixed signs for the deltas any more.

maroider:
So if I'm reading this correctly, macOS has a coordinate system that's "+y = up", "+x = left"?

madsmtm:
Yes, macOS _events_ have that coordinate system (macOS _windows_ have "+x = right" and "+y = up", see [this](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/CocoaDrawingGuide/Transforms/Transforms.html#//apple_ref/doc/uid/TP40003290-CH204-SW2)).

Source: The documentation for [`NSEvent -deltaX`](https://developer.apple.com/documentation/appkit/nsevent/1534871-deltax?language=objc) and [`NSEvent -deltaY`](https://developer.apple.com/documentation/appkit/nsevent/1534158-deltay?language=objc) says:

> -1 for [...] right and 1 for [...] left.

> -1 for [...] down and 1 for [...] up.

(So although not stated explicitly in the documentation for `scrollingDeltaX` and `scrollingDeltaY`, `deltaX` and `deltaY` link directly to them, so I think it's safe to assume that they work the same. Also, this matches results from testing)

emilk:
@michaelkirk makes an interesting point: with this PR, winit will report negated scroll directions compared to what native libraries report, on all platforms but on Mac. This could perhaps be a bit surprising, and we could consider flipping this.

From my point of view, the most important thing is that we now have a coordinate system that is the same everywhere in winit (X=right, Y=down), but scroll values can be interpreted two ways: how the _content_ should move, or how the _view_ should move. There is no right or wrong here. In this PR we are committing ourselves to having the scroll direction indicate how the _content_ should move. The main advantage here is that I would consider it less of a breaking change for winit: this PR flips the less-used horizontal scroll direction, instead of the more commonly used vertical scroll direction.

dhardy:
From my point of view changing the direction of the x-axis of `LineDelta` makes sense, but it's just a coordinate system so no big deal. Is `PixelDelta` similarly affected?

emilk:
> From my point of view changing the direction of the x-axis of `LineDelta` makes sense, but it's just a coordinate system so no big deal. Is `PixelDelta` similarly affected?

Yes, this PR flips the X sign for both `LineDelta` and `PixelDelta`. It is all consistent.

msiglreith:
Are there open points in this PR? (cc @madsmtm). Otherwise this looks good to go with the current implementation (according to the Mac axis definition).

msiglreith (approved):
thanks!

PR #2107, opened 2021-12-23
Provide examples for all window position/size setters

rukai:

This API is easy to understand once you've seen an example, but figuring it out from just the type signature is really difficult.
So every method with args of `<S: Into<Size>>` or `<P: Into<Position>>` should have an example demonstrating how to use it.

I considered putting this documentation just once in `Size` or `Position`.
But that would have been strange as we would be documenting how to instantiate a different type on a types documentation page.
And it would be harder to find for the user.

### DIFF
--- a/src/platform/unix.rs
+++ b/src/platform/unix.rs
@@ -345,9 +345,31 @@ pub trait WindowBuilderExtUnix {
     #[cfg(feature = "x11")]
     fn with_gtk_theme_variant(self, variant: String) -> Self;
     /// Build window with resize increment hint. Only implemented on X11.
+    ///
+    /// ```
+    /// # use winit::dpi::{LogicalSize, PhysicalSize};
+    /// # use winit::window::WindowBuilder;
+    /// # use winit::platform::unix::WindowBuilderExtUnix;
+    /// // Specify the size in logical dimensions like this:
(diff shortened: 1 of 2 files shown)


src/platform/unix.rs
```diff
@@ -345,9 +345,31 @@ pub trait WindowBuilderExtUnix {
     #[cfg(feature = "x11")]
     fn with_gtk_theme_variant(self, variant: String) -> Self;
     /// Build window with resize increment hint. Only implemented on X11.
+    ///
+    /// ```no_run
```
madsmtm, line 349: We could actually allow this example (and others) to run in CI, right?

rukai: ah, probably, ill see what happens

rukai: image
answer seems to be no.
I'll readd the no_run

madsmtm: Oh yeah, because of the event loop. But actually I was only thinking of the ones on the `WindowBuilder` (which doesn't require this)

rukai: Oh right, done!

maroider (approved):
I assume this is ready to be merged now?

rukai:
from my perspective yes

madsmtm (approved):
Also from mine, thanks for your work!

Issue #1402, opened 2020-01-17
Consider adding clippy on CI


kchibisov:
Right now clippy(stable branch) on `winit` (X11/Wayland backends) produces 1500 lines of warnings/errors, so I think it makes sense to at least take a look on them, since most of the time they actually make sense.

MrGibus:
Just started working on some things on windows backend and noticed a few pointless shifts ```<< 0```. Ran clippy and it warned about a number of float comparisons and quite a bit of redundant code.
So I'd agree with this.

PR #2108, opened 2021-12-23
Fix clippy warnings

rukai:
- [x] Tested on all platforms changed - Should be no change in functionality.

There are still some lints left, but this should be all the non-controversial ones. The rest can be ignored or fixed in a separate PR.

I decided to just ignore new_without_default as it seems really aggressive to me. Happy to just remove the ignore and leave the decision for a follow up PR.

Make sure to review the diff with whitespace disabled to keep the match changes readable.

### DIFF
--- a/src/platform_impl/linux/x11/ime/callbacks.rs
+++ b/src/platform_impl/linux/x11/ime/callbacks.rs
@@ -62,7 +62,7 @@ pub unsafe fn set_destroy_callback(
     inner: &ImeInner,
 ) -> Result<(), XError> {
     xim_set_callback(
-        &xconn,
+        xconn,
         im,
         ffi::XNDestroyCallback_0.as_ptr() as *const _,
         &inner.destroy_callback as *const _ as *mut _,
@@ -70,6 +70,7 @@ pub unsafe fn set_destroy_callback(
 }
 
 #[derive(Debug)]
+#[allow(clippy::enum_variant_names)]
 enum ReplaceImError {
     MethodOpenFailed(PotentialInputMethods),
     ContextCreationFailed(ImeContextCreationError),
@@ -136,13 +137,17 @@ pub unsafe extern "C" fn xim_instantiate_callback(
     let inner: *mut ImeInner = client_data as _;
     if !inner.is_null() {
         let xconn = &(*inner).xconn;
-        let result = replace_im(inner);
-        if result.is_ok() {
-            let _ = unset_instantiate_callback(xconn, client_data);
-            (*inner).is_fallback = false;
-        } else if result.is_err() && (*inner).is_destroyed {
-            // We have no usable input methods!
-            result.expect("Failed to reopen input method");
+        match replace_im(inner) {
+            Ok(()) => {
+                let _ = unset_instantiate_callback(xconn, client_data);
+                (*inner).is_fallback = false;
+            }
+            Err(err) => {
+                if (*inner).is_destroyed {
+                    // We have no usable input methods!
+                    panic!("Failed to reopen input method: {:?}", err);
+                }
+            }
         }
     }
 }
@@ -163,12 +168,12 @@ pub unsafe extern "C" fn xim_destroy_callback(
         if !(*inner).is_fallback {
             let _ = set_instantiate_callback(xconn, client_data);
             // Attempt to open fallback input method.
-            let result = replace_im(inner);
-            if result.is_ok() {
-                (*inner).is_fallback = true;
-            } else {
-                // We have no usable input methods!
-                result.expect("Failed to open fallback input method");
+            match replace_im(inner) {
+                Ok(()) => (*inner).is_fallback = true,
+                Err(err) => {
+                    // We have no usable input methods!
+                    panic!("Failed to open fallback input method: {:?}", err);
+                }
             }
         }
     }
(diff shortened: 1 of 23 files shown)


madsmtm (changes requested):
Related: https://github.com/rust-windowing/winit/pull/2076.

Quick notes: Agree with most of these, maybe we should enable `clippy` in CI in the future?

However, I think we should discuss the `match` changes in the examples: I think there's value in keeping them as-is, because it provides a nice "template" that users can easily copy-paste and extend on. Same goes for `winit` developers, I think it's useful to be able to easily drop in an extra pattern when debugging an example.

rukai:
I think the match changes make sense but happy to roll them back to keep this PR easy to merge.
Which of these change should I rollback:
* combining 2 match statements into 1
* replacing `match` with `if let`
?

madsmtm:
Honestly, can I choose both? At first glance I meant only the `match -> if let`, but I'm unsure if e.g. `examples/mouse_wheel.rs` got any clearer by combining the `match` statements, so I'd like to discuss that as well / at least have it explicitly changed.

rukai:
yep, both was definitely an option!
All the match changes should be reverted now

src/platform_impl/linux/x11/util/icon.rs
```diff
@@ -27,7 +27,7 @@ impl Icon {
         data.push(rgba_icon.height as Cardinal);
         let pixels = rgba_icon.rgba.as_ptr() as *const Pixel;
         for pixel_index in 0..pixel_count {
-            let pixel = unsafe { &*pixels.offset(pixel_index as isize) };
+            let pixel = unsafe { &*pixels.add(pixel_index) };
```
madsmtm, line 30: This one scared me for a second (thought it was suddenly trying to use integer addition, which is strictly not the same as pointer offset), but `pointer::add` is just a convenience for this exact case

madsmtm (approved):
Looked through it all, it looks fine, nice work!

Would it be possible for you to do the same for the other platforms as well?

src/lib.rs
```diff
@@ -131,6 +131,7 @@
 
 #![deny(rust_2018_idioms)]
 #![deny(rustdoc::broken_intra_doc_links)]
+#![allow(clippy::new_without_default)]
```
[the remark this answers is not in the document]
rukai: Alright, reverted this so each individual case can be investigated.

rukai:
I do actually have access to macos and windows so im happy to do those as well, but lets land this PR first.

kchibisov (changes requested):
Thx for the PR. It would be nice to finally resolve this issue https://github.com/rust-windowing/winit/issues/1402.

To test platforms other than your host I could recommend you doing something like `cargo +nightly clippy  --target=x86_64-pc-windows-msvc --all-targets ` (this works for like every platform winit supports).

rukai:
If you are happy with the fixes in this PR please land them first.

src/platform_impl/linux/x11/ime/callbacks.rs
```diff
@@ -136,13 +137,17 @@ pub unsafe extern "C" fn xim_instantiate_callback(
...
-            result.expect("Failed to reopen input method");
+        match replace_im(inner) {
+            Ok(()) => {
+                let _ = unset_instantiate_callback(xconn, client_data);
+                (*inner).is_fallback = false;
+            }
+            Err(err) => {
+                if (*inner).is_destroyed {
+                    // We have no usable input methods!
+                    panic!("Failed to reopen input method: {:?}", err);
```
maroider, line 148: Debug formatting is probably the correct "direct" translation, but is it the correct formatting option here?

src/platform_impl/linux/x11/window.rs
```diff
@@ -431,7 +433,7 @@ impl UnownedWindow {
 
             // Select XInput2 events
             let mask = {
```
maroider, line 435: Is the extra scope necessary/useful here?

maroider (changes requested):
looks mostly ok to me

src/platform_impl/linux/x11/ime/callbacks.rs
rukai: yes, I think so.
Not sure what else I could do here.

src/platform_impl/linux/x11/util/geometry.rs
```diff
@@ -248,7 +248,7 @@ impl XConnection {
             );
 
             // The list of children isn't used
-            if children != ptr::null_mut() {
+            if children.is_null() {
```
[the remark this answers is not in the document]
rukai: A pretty dangerous oversight, yes.
Good catch.

src/platform_impl/linux/x11/ime/callbacks.rs
maroider: Well, the alternative is "normal" formatting, but it's not that big of a deal either way.

rukai: I would just try it if I were at a computer but I'm pretty sure you can't just normal format an enum unless it implements display

src/platform_impl/linux/x11/window.rs
rukai: Removed the scope

src/platform_impl/linux/x11/ime/callbacks.rs
maroider: Ah, alright, that's my bad for not checking what the error type is.

maroider (approved):
It seems like this should be ready to be merged.

I'm assuming adding clippy to our CI config is intended to be added in a separate PR.

PR #2114, opened 2021-12-28
Update to rust 2021 edition

rukai:

The changes are generated from running:
```
cargo fix --edition
cargo fix --edition --target=x86_64-pc-windows-msvc
cargo fix --edition --target=x86_64-apple-darwin
cargo fmt
```
And then setting the edition to 2021

All of the generated changes are due to the disjoint capture changes https://doc.rust-lang.org/nightly/edition-guide/rust-2021/disjoint-capture-in-closures.html


rukai:
Oh right, lots of platform specific code in winit, ill investigate the issues.

maroider:
Did you run `cargo fix --edition --target=x86_64-pc-windows-msvc` and `cargo fix --edition --target=x86_64-apple-darwin` because CI was failing, or what?

rukai:
yeah CI was failing before I ran that.
Without running those extra commands `cargo fix` doesnt get run on the windows and mac os specific code.

src/platform_impl/windows/window.rs
```diff
@@ -610,26 +621,29 @@ impl Window {
...
-                cbSize: mem::size_of::<winuser::FLASHWINFO>() as UINT,
-                hwnd: window.0,
-                dwFlags: flags,
-                uCount: count,
-                dwTimeout: 0,
-            };
-            winuser::FlashWindowEx(&mut flash_info);
+        self.thread_executor.execute_in_thread(move || {
+            let _ = &window;
+            unsafe {
```
madsmtm, line 626: To have a smaller diff here, please move the `unsafe` back on the line it was before, if that makes sense

madsmtm (changes requested):
Thanks for your work in cleaning these things up!

src/platform_impl/windows/window.rs
rukai: As in `self.thread_executor.execute_in_thread(move || { let _ = &window; unsafe {` ?
I dont think rust fmt would allow that.

madsmtm: I was thinking:
```rust
self.thread_executor.execute_in_thread(move || unsafe {
    let _ = &window;
```

rukai: ah! I didnt think of that, sounds good to me

madsmtm (approved):

madsmtm:
I've independently verified that no `cargo fix --edition` was necessary on the other platforms, and that the feature changes (from the changed feature resolver) are of no concern

Issue #2045, opened 2021-10-31
Provide a more complete WASM example


maroider:
As noted  [here](https://github.com/rust-windowing/winit/issues/1662#issuecomment-698656972) and [here](https://github.com/rust-windowing/winit/issues/1662#issuecomment-698774314), `winit`'s current `web` example falls a bit short of showing off all the steps required to use `winit` in the browser.

PR #2115, opened 2021-12-28
Improve web example

rukai:
closes https://github.com/rust-windowing/winit/issues/2045

Infrastructure taken from wgpu's web example runner https://github.com/gfx-rs/wgpu/wiki/Running-on-the-Web-with-WebGPU-and-WebGL

I kept everything as they did it, but happy to further adjust or cleanup for winit's needs.
I adjusted the existing example logic to insert into a specific div rather than just appended to the body as this allows us to more easily place other html elements around it for testing different scenarios.
We could hardcode `run-wasm-example.rs` to always run the web.rs example but I think it makes sense to keep it flexible so developers can easily convert and test other examples.


README.md
```diff
@@ -82,6 +82,14 @@ For example code using Winit with WebAssembly, check out the [web example]. For
 information on using Rust on WebAssembly, check out the [Rust and WebAssembly
 book].
 
+To run the web example:
+
+1. Ensure you have bash and python 3 installed.
```
maroider, line 87: Requiring `bash` for everyone, even on Windows, feels a bit weird, but I suppose everyone using git has access to "git bash". As for listing python 3 as a hard dependency, well, that seems like a slight lie, as installing `basic-http-server` would seem to circumvent the need for python.

examples/web.rs
```diff
@@ -20,17 +20,33 @@ pub fn main() {
...
 
     event_loop.run(move |event, _, control_flow| {
         *control_flow = ControlFlow::Wait;
 
         #[cfg(target_arch = "wasm32")]
-        log::debug!("{:?}", event);
+        {
+            log::debug!("{:?}", event);
+
+            // Getting access to browser logs requries a lot of setup on mobile devices.
```
maroider, line 37:
```suggestion
            // Getting access to browser logs requires a lot of setup on mobile devices.
```

maroider (commented):
I'm admittedly more of a fan of using `xtask`s over adding shell scripts, but that's not something I'd consider a deal-breaker.

README.md
rukai: tbh I have no idea how well it runs on windows.
I'm just taking from the wgpu docs which seems to suggest it works fine as long as you have bash https://github.com/gfx-rs/wgpu/wiki/Running-on-the-Web-with-WebGPU-and-WebGL

rukai:
I was unfamiliar with the xtask pattern. seems like a pretty good idea to me, I'll investigate using that here.

rukai:
Some thoughts on xtask's:
* Given we only have one task, would we better off making it just `cargo run-wasm-example web` instead of `cargo xtask run-wasm-example web`. That would make things simpler for both the user and developer.
* It is recommended to have your project setup as a workspace when using xtasks, presumably to improve compile times by sharing the target directory. Should winit be refactored into a workspace?

rukai:
Alright should be good to review again.

I'm thinking of pulling most of the run-wasm-example logic into another crate outside of the winit repo.
I would further extend this crate to also support running binary crates not just `--example`s.
That way wgpu can make use of the improvements brought by run-wasm-example and users dont have to copy paste and then modify large chunks of arbitrary logic to get their web builds going.
However a downside of this is less visibility into whats going on inside and if winit needs to tweak something for their personal use that may be more complicated or impossible.

However we could totally land this PR as is and then consider any such changes later on.

There are many advantages brought by writing run-wasm-example in rust.
* cross platform
* 0 external dependencies
* better UX + more robust than anything hacked together with bash/ps/bat
* wasm-bindgen-cli version is always in sync with wasm-bindgen version because `cargo update` updates both of them at the same time

run-wasm/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "run-wasm"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+cargo-run-wasm = { git = "https://github.com/rukai/cargo-run-wasm" } # TODO: use crates.io release before merging
```
maroider, line 9: We technically don't need a crates.io release before using this in an xtask.

rukai: technically yes, may as well do the release before hand though.

rukai:
I ended up pulling the logic into its own external crate and settled on an API im really happy with.
It allows users to easily reuse it for their own purposes, allowing them to use it for running full crates not just running examples.

rukai:
I changed it to use the crates.io release.

However at the last moment I discovered a potential problem with the implementation: https://github.com/rustwasm/wasm-bindgen/issues/2757

rukai:
On the wgpu side we decided to just leave the current implementation as is.
We can reevaluate the implementation if breakage ever occurs.

To be on the safe side I removed `run-wasm` from CI, considering the implementation is just 1 function call it wasnt doing much.
This way if breakage ever occurs winit's CI isnt broken and I will still be notified by the cargo-run-wasm nightly dep check https://github.com/rukai/cargo-run-wasm/blob/main/.github/workflows/nightly_deps_check.yml to promptly fix the issue.

The PR is now ready for review again.

.cargo/config
```diff
@@ -0,0 +1,2 @@
+[alias]
+run-wasm = "run --release --package run-wasm --"
```
maroider, line 2:
```suggestion
run-wasm = "run --package run-wasm --"
```
The cost of compiling in release mode isn't worth the non-observable performance benefit.

maroider (commented):
I have a couple of issues still:
- `.cargo/config` should probably be `.cargo/config.toml`, as per the (now modified) [documentation](https://doc.rust-lang.org/cargo/reference/config.html#hierarchical-structure).
- Why does the runner compile *everything* into its own subdirectory in `target/`?
- It would be nice if the runner could exit gracefully on `Ctrl+C`.

I also found a further refinement on the bevy discord that we could apply here: it turns out that `cargo run` can be configured to do something else other than simply trying to execute a "native executable". We set the [`target.wasm32-unknown-unknown.runner`](https://doc.rust-lang.org/cargo/reference/config.html#targettriplerunner) key to configure this for wasm. The `cargo-run-wasm` might need some modification in order for this to work, however.
```toml
[target.wasm32-unknown-unknown]
runner = ["cargo", "run-wasm"]
```

.cargo/config
maroider: Another nit: the norm for multi-argument commands is to use an array of strings, with one string per argument, so you'd get `run-wasm = ["run", "--package", "run-wasm", "--"]`

rukai:
> Why does the runner compile everything into its own subdirectory in target/?

That is explained here https://github.com/rukai/cargo-run-wasm/blob/0dce868c851306655fd312f3c59454ce788a92e0/src/lib.rs#L97

> It would be nice if the runner could exit gracefully on Ctrl+C

What is ungraceful about the current shutdown?

.cargo/config
rukai: On my machine, just going visually from the time taken for the `Serving web on http://localhost:8000` message to appear on debug mode it takes about ~3s on release it takes ~100ms

That seems like a huge improvement to iterative builds and makes the up front --release build absolutely worth it.

rukai:
> I also found a further refinement on the bevy discord that we could apply here: it turns out that cargo run can be configured to do something else other than simply trying to execute a "native executable". We set the target.wasm32-unknown-unknown.runner key to configure this for wasm. The cargo-run-wasm crate might need some modification in order for this to work, however.

There would definitely be some advantages to this runner approach:
* `cargo run --target wasm-unknown-unknown --example web` is standard and so will fit the expectations of experienced cargo users.
* Doesn't reimplement cargo's `--release`, `--example` etc. arguments so it will always remain consistent with cargo's arguments

However the core reason I think this is a bad idea is:
* The runner approach would prevent the RUSTFLAGS workaround https://github.com/rukai/cargo-run-wasm/blob/0dce868c851306655fd312f3c59454ce788a92e0/src/lib.rs#L97

Another reason against is:
* The runner approach prevents adding cargo-run-wasm specific cli args as the runner only receives the name of the executable to run.
    + In the future we might want to add various cli args such as: enable/disable post-build wasm optimization, enable/disable webpage refresh on rebuild.

I would consider changing to a runner approach in the future if the RUSTFLAGS workaround became unnecessary.

All the feedback for this PR so far should be addressed now.

maroider:
> What is ungraceful about the current shutdown?

It's not that bad tbh, I just don't like seeing the error message I get when I use the one way I have of closing `run-wasm`.

> The runner approach would prevent the RUSTFLAGS workaround https://github.com/rukai/cargo-run-wasm/blob/0dce868c851306655fd312f3c59454ce788a92e0/src/lib.rs#L97

I was unaware of this being an issue (and I use `mold` by default on Linux). It looks to me like this is a problem specifically with using the `RUSTFLAGS` environment variable, and I haven't noticed anything like what's described in the linked issue when using `rustflags` set in *some* `config.toml`.

rukai:
Can you send me the error message?
On my machine when I close with Ctrl-C it closes silently.

Huh weird that you dont get the rebuild issues, I'll investigate.

rukai:
I can reproduce it with just:
image
On winit master branch with `~/.cargo/config.toml` of:
```toml
[build]
[target.x86_64-unknown-linux-gnu]
rustflags = [
  "-C", "link-arg=-fuse-ld=lld",
]
```

maroider:
Ah, I figured out why this wasn't affecting me. Turns out I was still using `mold -run` as a rustc wrapper, and that doesn't have the same issue.

rukai:
That makes sense because mold -run doesn't change rustflags!

examples/web.rs
```diff
@@ -22,15 +22,39 @@ pub fn main() {
...
+        body.append_child(&canvas).unwrap();
+
+        let log_header = document.create_element("h2").unwrap();
+        log_header.set_text_content(Some("Event Log"));
+        body.append_child(&log_header).unwrap();
+
+        let log_list = document.create_element("ul").unwrap();
+        body.append_child(&log_list).unwrap();
+        log_list
+    };
```
maroider, line 36: Maybe we should have this live in a freestanding function.

examples/web.rs
```diff
@@ -22,15 +22,39 @@ pub fn main() {
...
+            if let Event::WindowEvent { event, .. } = &event {
+                let window = web_sys::window().unwrap();
+                let document = window.document().unwrap();
+                let log = document.create_element("li").unwrap();
+                log.set_text_content(Some(&format!("{:?}", event)));
+                log_list
+                    .insert_before(&log, log_list.first_child().as_ref())
+                    .unwrap();
+            }
+        }
```
maroider, line 57: ditto

maroider (commented):
As it stands, this is already an improvement on the status quo. In the interest of getting this merged at all, I think it's OK to merge it mostly as-is.

rukai:
CI failure not from my PR.

maroider:
Ah, `unsafe_op_in_unsafe_fn`. I'll PR a quick fix.

maroider:
#2207 should fix it

rukai:
green!

Issue #1886, opened 2021-03-16
Compiling `glutin` fails with missing `set_suspend_callback`


M2WZ:
I'm on Windows 10 and tried to compile the [egui_template](https://github.com/emilk/egui_template/) for android, but got [a few errors (pastebin)](https://pastebin.com/DnZcF6Np). It are problems with glutin, but after asking on [glutin's freenode](https://webchat.freenode.net/#glutin), someone said the first error 
```
error[E0599]: no method named `set_suspend_callback` found for reference `&EventLoopWindowTarget<T>` in the current scope
--> C:\Users\Me\.cargo\registry\src\github.com-1ecc6299db9ec823\glutin-0.26.0\src\api\android\mod.rs:76:12 
     |
76 |         el.set_suspend_callback(Some(Box::new(move |suspended| {
     |            ^^^^^^^^^^^^^^^^^^^^ method not found in `&EventLoopWindowTarget<T>`
```
is an error from winit, from [an older commit](https://github.com/rust-windowing/winit/commit/b8828105cf6231e249da279851285aecd2970b4b#diff-8a469f9184766b66e49647a75192f001ef549e5f6fcf8c98eb4fc23ce9e87e15L7-L10) and pointed out that the CI from winit doesn't test android since the switch to Github Actions.
Could someone look into this (and does anyone know anything about the other errors)?

msiglreith:
Hi!
`winit` includes Android CI checks (https://github.com/rust-windowing/winit/blob/master/.github/workflows/ci.yml#L42) and I also did some Android dev with wgpu a few weeks ago. It seems more like `glutin` bump the `winit` dependencies at some point which caused issues with the Android platform implementation. See also https://github.com/rust-windowing/glutin/issues/1307, which also includes some more WIP branch/PR.

maroider:
> someone said [...] and pointed out that the CI from winit doesn't test android since the switch to Github Actions.

Hi! That was me.

What was actually written was:

> That first error looks like a regression from b8828105 (Winit), which didn't get caught by CI somehow

> >    Date: Wed May 6 15:27:49 2020 +0200 

> lol, this must have been broken for a while

> https://github.com/rust-windowing/winit/commit/b8828105cf6231e249da279851285aecd2970b4b#diff-8a469f9184766b66e49647a75192f001ef549e5f6fcf8c98eb4fc23ce9e87e15L7-L10

> wow, Glutin hasn't had android CI since the switch to GitHub actions

[...]

PR #2118, opened 2021-12-28
 Compiling on windows for android errors #1886

Kaiser1989:

This merge request adds functionality to implement a working solution for glutin on android.

Everthing starts here (Glutin build failure for android targets)
https://github.com/rust-windowing/glutin/issues/1307

I fixed this, by customizing winit and glutin:
https://github.com/Kaiser1989/winit
https://github.com/Kaiser1989/glutin

Working examples with both fixes can be found here:
https://github.com/Kaiser1989/rust-android-example
https://github.com/Kaiser1989/game-gl

Several messages arrived me, that i should pls try to merge my changes to the original crates, so these projects can stay up-to-date. 

PS: Do not check the single commits, just check the 3 files. There aren't that much changes, mainly adding resume and suspend events, and brining back run_return function.

Best regards
Philipp


src/event_loop.rs
```diff
@@ -154,6 +154,14 @@ impl<T> EventLoop<T> {
         self.event_loop.run(event_handler)
     }
 
+    #[inline]
+    pub fn run_return<F>(&mut self, event_handler: F)
```
madsmtm, line 158: `run_return` is already available on Android with the `platform::run_return::EventLoopExtRunReturn` trait

src/platform_impl/android/mod.rs
```diff
@@ -144,8 +144,24 @@ impl<T: 'static> EventLoop<T> {
...
+                    }
+                    Event::Resume => {
+                        call_event_handler!(
+                            event_handler,
+                            self.window_target(),
+                            control_flow,
+                            event::Event::Resumed
+                        );
+                        self.running = true;
+                    }
```
madsmtm, line 164: Can you please add some information about why this is necessary, maybe with a few links to the Android docs? `WindowDestroyed` already sends `Suspended`, why is that not enough?

src/platform_impl/android/mod.rs
```diff
@@ -169,6 +185,18 @@ impl<T: 'static> EventLoop<T> {
...
+                            window_id: window::WindowId(WindowId),
+                            event: event::WindowEvent::CloseRequested,
+                        };
+                        call_event_handler!(
+                            event_handler,
+                            self.window_target(),
+                            control_flow,
+                            event
+                        );
+                    }
```
madsmtm, line 199: It seems wrong to send `CloseRequested` when the application is destroyed; maybe it should be the `Destroyed` event instead?

madsmtm (changes requested):

madsmtm:
It should be noted, I don't know anything about Android and haven't actually read the linked issues, so maybe you've already answered the comments there.

src/platform_impl/android/mod.rs
Kaiser1989: https://github.com/rust-windowing/glutin/issues/904
https://github.com/rust-windowing/glutin/pull/1313

Check this comment:
https://github.com/rust-windowing/glutin/pull/1313#issuecomment-656477993

Android has some special windows context handling, where context cannot be resused after resuming. Data has to be reinitialized. Therefore we need to handle those events.

Old version just sets flags:
```
Event::Pause => self.running = false,
Event::Resume => self.running = true
```

Kaiser1989: More discussion on this:
https://github.com/rust-windowing/android-ndk-rs/issues/117

src/platform_impl/android/mod.rs
Kaiser1989: If I get everthing correctly, there is no direct destroy in android. There is just a request to destroy the app. It's send to background, to make reopening faster. If there are not enough resources available, the app gets destroyed, but those events are handled by OS.

I will check this again!

src/event_loop.rs
Kaiser1989: I will check this again! Maybe it can be removed.

Kaiser1989:
I removed the duplicated run_return function. You're right, it can be easily aquired by using the trait. Currently the biggest Problem is, that pause and resume need to be handled on android side, graphics data need to be reinitialized. Currently these events are lost.

src/platform_impl/android/mod.rs
Kaiser1989: I checked this. I'm open to change this to WindowEvent::Destroyed. I think it's some sort of interpretation:
Is it asking for destroying or is it force destroying. 
When android receives this event, render context is still alive and can be saved or released gracefully, therefore i thought it as "the window is requested to be closed". You can also see this as forced shutdown.

If this is the only problem, i will change it.

src/platform_impl/android/mod.rs
MarijnS95: I doubt that is correct: https://developer.android.com/guide/components/activities/activity-lifecycle#onpause

src/platform_impl/android/mod.rs
MarijnS95: `Event::Destroy` happens when the entire activity is destroyed (https://developer.android.com/guide/components/activities/activity-lifecycle#ondestroy), theoretically the window should have been destroyed prior, with `Event::WindowDestroyed`?

This is dispatched as a `winit` `Suspended` event though. Android''s Suspend/Pause and Resume events represent something different than winit's `Suspended` / `Resumed` (specific for Android backends iirc).

src/platform_impl/android/mod.rs
Kaiser1989: > I doubt that is correct: https://developer.android.com/guide/components/activities/activity-lifecycle#onpause
 Why? We need to handle pause and resume events on android side. The point is, no one can tell how long the app was paused, maybe it was send to background for several hours... We need to free graphics resources here. If the app resumes, it's possible that original window context is not valid anymore. We cannot keep rendering to that destroyed window.
 
 Just settings flags is not an option.

MarijnS95: The linked docs are pretty clear on this:

> You can also use the onPause() method to release system resources, handles to sensors (like GPS), or any resources that may affect battery life while your activity is paused and the user does not need them. However, as mentioned above in the onResume() section, a Paused activity may still be fully visible if in multi-window mode. As such, you should consider using onStop() instead of onPause() to fully release or adjust UI-related resources and operations to better support multi-window mode.

MarijnS95: In other words: while I do agree it might be nice to forward these events through `winit`, the `Suspended` and `Resumed` events are not right as they serve a different purpose.

`onPause` and `onResume` should result in a [`WindowEvent::Focused`](https://docs.rs/winit/latest/winit/event/enum.WindowEvent.html#variant.Focused).

Perhaps we should rename `Suspended` and `Resumed` to more closely represent the graphics state needing to be forcibly destroyed and recreated, respectively?

Kaiser1989: I agree, maybe I abused these events to get my context handling correctly. The main point is, that we must not lose important android events. We need to be able to react on window changes.
When do resourced need to be created, when do they need to be destroyed.

And yes you are right, it would be much more coherent, if these events would be of type `WindowEvent`. Like i'm handling the resize event:

```
    WindowEvent::Resized(physical_size) => {
        if let Some(device_ctx) = self.device_ctx.as_mut() {
            device_ctx.window_context.resize(physical_size);

            // call resize device callback
            self.runner.resize_device(
                &device_ctx.gl,
                physical_size.width,
                physical_size.height,
            );
        }
    }
    ....
```

src/platform_impl/android/mod.rs
Kaiser1989: It would be best, if we can strictly divide Event and WindowEvent. The questio is, if winit fires all those events correctly, also in android environment. Again, we must not lose any important events for graphics handling.

src/event_loop.rs
Kaiser1989: I removed this.

src/platform_impl/android/mod.rs
Kaiser1989: I checked the android behaviour:
Focus can not be used to create the device, as i get following events when starting the main loop:

- resume
- resize
- focus

There is no window create event or such. The focus event is too late to use it for creation, as resize event already executed.

MarijnS95: > Focus can not be used to create the device

Nor is the android documentation suggesting to. Swapchain creation and deletion "must" be tied to window creation and destroying: you can do it whenever you like, as long as it is created _after_ `WindowCreated` and destroyed before or during `WindowDestroyed`. Applications are free to perform this elsewhere, as long as they synchronize properly with the events (ie. hold on to the lock from `native_window()`, and destroy your resources followed by releasing this lock whenever `WindowDestroyed` is called).

> There is no window create event or such.

At what level are you logging this? Is logging initialized early enough (ie. in `ndk_glue::main`)? If Android doesn't give your application a window there's nothing to render to.
My best guess is "resume" in your list is referring to `event::Event::Resumed`, ie. triggered on `ndk_glue::Event::WindowCreated`?

src/platform_impl/android/mod.rs
MarijnS95: > Again, we must not lose any important events for graphics handling.

As said it'd be great if this event is handled properly in `winit`, but we must not diverge from `winit` patterns too far. It is already a stretch to have `Suspended` and `Resumed` (whether their naming is useful/accurate is still up for debate) as this makes `winit` applications for other platforms not immediately runnable on Android without refactoring swapchain-lifetime logic to adhere to these events.

(Tangent: theoretically it'd be nice if every platform could emit these events for swapchain creation and destruction, making the whole thing more generic. Even better: we provide the `RawWindowHandle` only in the `Resumed` [or otherwise named] (`Window`)`Event` so that crates are forced to handle this event instead of calling `fn raw_window_handle` at will.)

> I think it's some sort of interpretation:
> Is it asking for destroying or is it force destroying.

Regarding this prior question: Whenever this function/event is called, you are _supposed to_ release all resources related to the window (again: the swapchain) and return from this function (`ndk_glue` does that after the lock from `fn native_window()` is released, presuming the caller read the documentation and is already holding on to this lock while they have resources created on the window). In that sense you _could_ keep using and rendering to the swapchain without ever returning from `on_window_destroyed`, but I don't know what Android will do in this case (it's already out of focus and most likely also out of view, so it'll probably result in an ANR, and no new events will arrive).

src/event_loop.rs
MarijnS95: Don't forget to remove it from the changelog too.

While at it, make sure crate users understand what changes to make, or how they can take advantage of the new functionality that is added.

lattice0:
Does this have a future? I also made PRs, but simpler https://github.com/rust-windowing/winit/pull/2297

MarijnS95:
I don't think so. We could pick up some of the review comments in smaller followup steps but this seems to have gone stale.

MarijnS95:
I've picked up this `glutin` on Android issue in hopes of finally being able to resolve and close lots of reports and PRs around the `rust-windowing` repos. This PR is not necessary to make it work, and as discussed above all three event-mappings that it adds are wrong.

If someone wishes to proceed with the suggested renaming of events to be more in-line with Android naming and/or make it more consistent with other platforms, please open a new PR which clearly states that intent. Feel free to also add proper events for `onResume` and `onPause`, which - for the record again - have nothing to do with `winit` `Resumed`/`Suspended` for window creation/destruction.

PR #2119, opened 2021-12-29
Revert "Add composition event on macOS (#1979)"

### LINKED ISSUE #2097: Version 0.26.0 breaks macOS IME
It appears that a recent change has caused IME to no longer work on macOS.

Originally reported to Alacritty in https://github.com/alacritty/alacritty/issues/5669.

@wendajiang could you maybe more detailed description of your issue? A short video of expected/actual would probably be helpful. In your Alacritty report it's not entirely obvious if "ni hao" was what you typed or if that was inserted.

ArturKovacs:
Fixes #2097

This reverts commit 8afeb910bdbdcaeefdcea948098c3379de3f433f.

Reverting because this change made Pinyin input unusable
(only latin characters showed even after selecting the
desired Chinese character)

Furthermore the change itself was pretty much a hack. The change allowed an application to show the preedit text. However winit doesn't yet have an API for this so the implementation simply sent `ReceivedCharacter` events for each character of the preedit text and whenever the text changed it sent 'u\{7f}' (aka DELETE) characters for each UTF32 code unit of the previous preedit text, then sent each character of the new preedit text.

For all these reasons, I believe a revert is warranted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 - On X11, add mappings for numpad comma, numpad enter, numlock and pause.
+- On macOS, fix Pinyin IME input by reverting a change that intended to improve IME.
 
 # 0.26.0 (2021-12-01)
 
--- a/src/platform_impl/macos/util/mod.rs
+++ b/src/platform_impl/macos/util/mod.rs
@@ -3,13 +3,8 @@ mod cursor;
 
 pub use self::{cursor::*, r#async::*};
 
-use std::{
-    cell::Cell,
-    ops::{BitAnd, Deref},
-    rc::Rc,
-};
+use std::ops::{BitAnd, Deref};
 
-use block::ConcreteBlock;
 use cocoa::{
     appkit::{NSApp, NSWindowStyleMask},
     base::{id, nil},
@@ -18,12 +13,8 @@ use cocoa::{
 use core_graphics::display::CGDisplay;
 use objc::runtime::{Class, Object, Sel, BOOL, YES};
 
-use crate::{
-    dpi::LogicalPosition,
-    platform_impl::platform::ffi::{
-        self, NSRange, NSStringEnumerationByComposedCharacterSequences,
-    },
-};
+use crate::dpi::LogicalPosition;
+use crate::platform_impl::platform::ffi;
 
 // Replace with `!` once stable
 #[derive(Debug)]
@@ -111,31 +102,6 @@ pub fn window_position(position: LogicalPosition<f64>) -> NSPoint {
 
 pub unsafe fn ns_string_id_ref(s: &str) -> IdRef {
     IdRef::new(NSString::alloc(nil).init_str(s))
-}
-
-/// Returns the number of characters in a string.
-/// (A single character may consist of multiple UTF-32 code units.
-/// This is possible when long sequences of composing characters are present)
-///
-/// Unsafe because assumes that the `string` is an `NSString` object
-pub unsafe fn ns_string_char_count(string: id) -> usize {
-    let length: NSUInteger = msg_send![string, length];
-    let range = NSRange {
-        location: 0,
-        length,
-    };
-    let char_count = Rc::new(Cell::new(0));
-    let block = {
-        let char_count = char_count.clone();
-        ConcreteBlock::new(move || char_count.set(char_count.get() + 1)).copy()
-    };
-    let block = &*block;
-    let () = msg_send![string,
-        enumerateSubstringsInRange:range
-        options:NSStringEnumerationByComposedCharacterSequences
-        usingBlock:block
-    ];
-    char_count.get()
 }
 
 #[allow(dead_code)] // In case we want to use this function in the future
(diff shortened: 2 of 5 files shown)


ArturKovacs:
With all that said I believe there's a strong need for a proper IME API in winit.

ArturKovacs:
@madsmtm @maroider Do you mind if I merge this?

maroider:
So if I've read this correctly, IME will work *fine*, but will lack the "enhancement" added by #1979?

If so, then I have no issue with reverting it, as having all IMEs work *at all* is more important than hacking in pre-edit text.

ArturKovacs:
> So if I've read this correctly, IME will work _fine_, but will lack the "enhancement" added by #1979?

Yes.

src/platform_impl/macos/util/mod.rs
```diff
@@ -113,31 +104,6 @@ pub unsafe fn ns_string_id_ref(s: &str) -> IdRef {
...
-pub unsafe fn ns_string_char_count(string: id) -> usize {
-    let length: NSUInteger = msg_send![string, length];
-    let range = NSRange {
-        location: 0,
-        length,
-    };
-    let char_count = Rc::new(Cell::new(0));
-    let block = {
-        let char_count = char_count.clone();
-        ConcreteBlock::new(move || char_count.set(char_count.get() + 1)).copy()
```
madsmtm, line 130: This was also UB btw, the block must take the four arguments as described in the [docs](https://developer.apple.com/documentation/foundation/nsstring/1416774-enumeratesubstringsinrange?language=objc) (even if unused)

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Unreleased
 
 - On X11, add mappings for numpad comma, numpad enter, numlock and pause.
+- On macOS, revert a change that inteded to improve IME but caused Pinyin input to stop working. (In other words, this change fixes Pinyin input)
```
madsmtm, line 4: `inteded` is misspelled, and the wording is a bit convoluted. Maybe something along the lines of "fixes Pinyin IME input by reverting ..."

src/platform_impl/macos/view.rs
```diff
@@ -431,62 +423,32 @@ extern "C" fn set_marked_text(
...
 }
 
-extern "C" fn unmark_text(this: &mut Object, _sel: Sel) {
+extern "C" fn unmark_text(this: &Object, _sel: Sel) {
     trace!("Triggered `unmarkText`");
     unsafe {
-        clear_marked_text(this);
+        let marked_text: id = *this.get_ivar("markedText");
+        let mutable_string = marked_text.mutableString();
+        let _: () = msg_send![mutable_string, setString:""];
```
madsmtm, line 445: Not sure whether this or the other method (release + alloc) is better here, but let's just leave it be for now

madsmtm:
Tagging @komi1230 and linking the issue: https://github.com/rust-windowing/winit/pull/1979

madsmtm (approved):
I'm very inexperienced with IME, but I agree that it's more important to fix the regression than for the "smart" rendering to work (to be honest, I never liked that hack, so that may also influence my decision).

komi1230:
I was also inexperienced in IME and #1979 was just a workaround to make Japanese input work fine.

I'm sorry for taking time from you and I appreciate you to spending much cost to fix this.

src/platform_impl/macos/util/mod.rs
ArturKovacs: Thanks for pointing this out, I'll try to remember this for the future.

CHANGELOG.md
ArturKovacs: Should be resolved now

src/platform_impl/macos/view.rs
ArturKovacs: I agree that releasing and allocating a new one would be safer, but the reference to `markedText` seems to be constrainted to the View, so it seems safe enough.

ArturKovacs:
@komi1230 I'm glad you contributed, and your original change added a feature that I also wanted have in winit. Actually I still want this feature but we will need a separate API for IME to do it right.

ArturKovacs:
@madsmtm did I address all your concerns and would you say that this can be merged?

madsmtm:
Yup, go ahead!

komi1230:
As the current status, Japanese input method still doesn't work by merging this change and Chinese input method *accidentally* works fine.
We should add new API to handle pre-edit text as @ArturKovacs says.

I am Alacritty user with Japanese language, so I have big motivation to fix this issue.

In the below code, there is `is_synthetic` field in struct `KeyboardInput` in enum `WindowEvent`.
I looked up how to use this field but I'm not sure.

https://github.com/rust-windowing/winit/blob/1b3b82a3c1369c5248a2e6a251230ba6c615e918/src/event.rs#L251-L264

To implement new API to handle pre-edit text, should we add new field for this ?

ArturKovacs:
> To implement new API to handle pre-edit text, should we add new field for this ?

I don't think that's a good approach. I think we will need a new WindowEvent kind. See the following thread for discussion about this: https://github.com/rust-windowing/winit/issues/1497

Issue #2082, opened 2021-12-03
Transparent windows are completely broken on Windows 11


parasyte:
Running the `transparent` example on Windows 11 (version 21H2, build 22000.348) appears to completely crash something in the OS. My screens flash black for a moment, and all windows appear as just the decoration with empty contents for a second until things settle down. This looks pretty similar to what happens when you update the display drivers, except window positions are not reset (because the driver update temporarily changes the desktop resolution).

Whatever crashes also disables Night Light; I have to re-enable it manually in `Settings > System > Display`. Working on this bug in the middle of the night is _really_ unkind to my eyesight.

`git bisect` points at #1933 as causing the crash.

The `transparent` window itself is _completely invisible_, but it still receives events. I am not sure if this example is supposed to behave this way. It acts the same with `0.25.0` on Windows 11. I don't think it's supposed to do this, considering https://github.com/rust-windowing/winit/pull/1933#issuecomment-873575816 shows a video with this example creating an actual window. Something I don't get at all.

parasyte:
Here's some more info with screenshots to help illustrate the issues.

`0.26.0` and `0.25.0` both show the same completely invisible window behavior. Here's what it looks like with 0.25.0 specifically. I'm hovering over the window in the taskbar to bring up the little invisible window preview:

![winit-transparent-0 25 0](https://user-images.githubusercontent.com/456942/144582184-2ecd2622-e98e-4dfd-8816-9b55b5bdb310.png)

And if I run the same example in `0.24.0`, I do see something resembling a transparent window:

![winit-transparent-0 24 0](https://user-images.githubusercontent.com/456942/144582259-9c80d4e8-e116-4621-98dc-0a431f214839.png)

🤷

msiglreith:
Thanks for the report - checked on Windows 10 and seems to work fine there. Unfortunately, I don't have a Win11 compatible system so I currently can't looks further into this.

msiglreith:
Should be fixed by https://github.com/rust-windowing/winit/pull/2121

parasyte:
Confirmed!

PR #2121, opened 2022-01-01
Fix transparent window crash on Windows 11

yslib:
Maybe the transparent setting in WM_NCCREATE on Windows 11 will cause a block when callling DwmEnableBlureBehindWindow and will crash. Puts them into WM_CREATE and it works. #2082 
I have not found any details about  the behavior change on Windows 11. Perhaps it's could be a Win11 bug.
I have no win8 or earlier enviroment so I can't test them on these platforms. 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 - On X11, add mappings for numpad comma, numpad enter, numlock and pause.
+- On Windows, fix a crash with transparent windows on Windows 11.
 
 # 0.26.0 (2021-12-01)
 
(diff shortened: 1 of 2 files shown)


msiglreith (approved):
Thanks! Tried on Win10 and works fine as well there.

msiglreith:
Could you add a changelog entry please?
Example
```
On Windows, fix a crash with transparent windows on Windows 11.
```

yslib:
> Could you add a changelog entry please? Example
> 
> ```
> On Windows, fix a crash with transparent windows on Windows 11.
> ```

Sorry for forgetting it. It was added.

PR #2126, opened 2022-01-03
Remove entries from PULL_REQUEST_TEMPLATE.md that are now covered by CI

rukai:
- [x] Tested on all platforms changed
- [x] Compilation warnings were addressed
- [x] `cargo fmt` has been run on this branch
- [x] `cargo doc` builds successfully
- [x] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users
- [x] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior
- [x] Created or updated an example program if it would help users understand this functionality
- [x] Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented

The more entries there are, the more likely people are to start just ignoring them all.
By removing the ones already covered by CI then the remaining entries are more likely to be followed.

PR #2130, opened 2022-01-03
Make CI run on changes to non-code files as well

madsmtm:
The repository is configured such that specific checks are required to pass before merging; since CI doesn't run on PRs that don't change code, they can't be merged without administrator intervention.

Currently affecting https://github.com/rust-windowing/winit/pull/2126, and has affected many others in the past.

Also, while this definition is currently correct, we might in the future change something so that changes to other files (like markdown files) really should be run through CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,16 +2,8 @@ name: CI
 
 on:
   pull_request:
-    paths:
-      - '**.rs'
-      - '**.toml'
-      - '.github/workflows/ci.yml'
   push:
     branches: [master]
-    paths:
-      - '**.rs'
-      - '**.toml'
-      - '.github/workflows/ci.yml'
 
 jobs:
   Check_Formatting:


maroider:
Ideally, we'd have CI skip any code-checking steps if no code files were touched, instead of running all the steps all of the time.

rukai (approved):
LGTM
Running CI on all PRs is not that expensive and is far safer as there are a lot of ways to break things outside of changing code.

madsmtm:
> Ideally, we'd have CI skip any code-checking steps if no source code was touched, instead of running all the steps all of the time, but that's also a bit more complicated to set up.

Also not something neither you nor I can fix, I think it's part of the repo settings that only an admin can access?

maroider:
> Also not something neither you nor I can fix, I think it's part of the repo settings that only an admin/"owner" can access?

It's something a more sophisticated CI config could solve.

madsmtm:
Fair enough, I'll gladly admit I don't know enough about GitHub Actions.

maroider (approved):
This, while somewhat wasteful, is probably nicer than having to circumvent the "CI passes" check.

We should probably try to get a more granular CI config set up eventually.

PR #2132, opened 2022-01-04
Set WindowBuilder to must_use

rukai:

Makes sense to be must_use as failing to call finish() would almost certainly indicate a bug.


maroider:
I tend to shy away from using `#[must_use]` for fear of generating a lot of noisy warnings, but I suppose `WindowBuilder` won't be used too often.

kchibisov (approved):

PR #2134, opened 2022-01-04
Make `WindowAttributes` private

madsmtm:
`WindowAttributes` was introduced way back in https://github.com/rust-windowing/glutin/pull/603, and later slowly had it's functionality moved into `WindowBuilder` (created in https://github.com/rust-windowing/glutin/pull/769).

At this point, the only thing you can achieve with `WindowAttributes` is:
- Inspection of the state that `WindowBuilder` is currently in - I can't see a usecase for this? And if we do get one, such ability can be added to `WindowBuilder` directly.
- Resetting `[min_/max_]inner_size` and `position` - Looking at [prior art](https://doc.rust-lang.org/std/process/struct.Command.html), this seems like a thing that struct builders rarely support, though we could do so if desired (making the functions take `Option`s).

The downsides, however, is more API surface that users have to look at, and that we have to support, and that the documentation for `WindowBuilder` is more scattered. I think these downsides outweigh the slight exposed capabilities, and would like to propose we remove this structure from the public API.

  - Not really applicable


madsmtm:
I'm not sure I fully understand, do you want:
```rust
struct WindowBuilder {
    inner_size: Option<Size>,
    resizable: bool,
    ... // Other fields from `WindowAttributes`
}
```
?

Especially the phrase "pod-like data" is confusing me?

aloucks:
> I'm not sure I fully understand, do you want:
> 
> ```rust
> struct WindowBuilder {
>     inner_size: Option<Size>,
>     resizable: bool,
>     ... // Other fields from `WindowAttributes`
> }
> ```

Yes

madsmtm:
Fair enough - I agree, but I'd like to keep it out of this PR to make the changes easier to review. I'll follow up with one to do that (though maybe first after #2137)

daxpedda:
Cc #2613.

Issue #2120, opened 2021-12-30
Platform-specific event callbacks


maroider:
Issues like #1878, #1928, and #1759 have made it clear that there is a need to hook into platform-specific events that winit itself doesn't (and arguably shouldn't) handle. Ideally, we'd give users the ability to hook into these events with minimal effort, and in a way that doesn't interfere with winit (although some would perhaps like to do that as well).

The simplest solution I can think of to this is to add an `EventBuilder` struct where you can register platform-specific callback functions before the event loop is ever entered. This will make it so that these callbacks don't miss any events, like the file opening events in #1759.

Granted, this doesn't lend itself to being a particularly ergonomic API, as any data shared between the platform-specific callbacks and the platform-independent on would have to be behind something like an `Rc<RefCell<T>>` or an `Arc<Mutex<T>>`. I think there *may* exist a solution where sharing such state is easier, but I don't want to spend much energy on that ATM.

Another issue is that this will (by necessity) make some projects (like AccessKit) end up depending directly on winit or have to provide an adapter crate. Maybe it could be possible to pull a `raw-window-handle` and create some cross-windowing-library interface for this, but  I don't think it's ultimately going to be all that feasible.

The Windows backend for this should be fairly trivial, so I'll go and implement that soon-ish, unless there are any major concerns with this proposal.

###### This is on some level related to #2010, but this concerns the public API, rather than the internal one.

maroider:
cc @mwcampbell

maroider:
cc @mwcampbell @fschutt

maroider:
The minimal API for Windows would look something like this:
```rust
impl WindowsEventLoopBuilder for EventLoopBuilder {
    pub fn with_callback<F>(&mut self, event_mask: u32, callback: F)
    where
        F: FnMut(*mut c_void, u32, usize, usize) -> usize,
    {
    }
}
```
Although something should probably be done to let end users correlate the `WindowId` with the `HWND` (the *mut c_void in this case) more easily.

mwcampbell:
Thanks @maroider for working on this.

I'll settle for having a separate `accesskit_winit` crate rather than having any of my existing crates depend directly on winit.

I assume the `*mut c_void` parameter to the callback is an HWND. If so, that looks fine.

ArturKovacs:
The API you proposed for Windows looks totally reasonable. Maybe it would be better to pass in the WindowId instead of the HWND and provide a platform specific extension trait for getting the HWND from the WindowId.

[...]

PR #2137, opened 2022-01-05
Add `EventLoopBuilder`

madsmtm:
The beginning of https://github.com/rust-windowing/winit/issues/2120.

  - [x] macOS
  - [ ] X11
  - [ ] Wayland
  - [ ] Windows

### DIFF
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -10,6 +10,7 @@
 //! [event_loop_proxy]: crate::event_loop::EventLoopProxy
 //! [send_event]: crate::event_loop::EventLoopProxy::send_event
 use instant::Instant;
+use std::marker::PhantomData;
 use std::ops::Deref;
 use std::{error, fmt};
 
@@ -31,7 +32,7 @@ use crate::{event::Event, monitor::MonitorHandle, platform_impl};
 ///
 pub struct EventLoop<T: 'static> {
     pub(crate) event_loop: platform_impl::EventLoop<T>,
-    pub(crate) _marker: ::std::marker::PhantomData<*mut ()>, // Not Send nor Sync
+    pub(crate) _marker: PhantomData<*mut ()>, // Not Send nor Sync
 }
 
 /// Target that associates windows with an `EventLoop`.
@@ -42,7 +43,62 @@ pub struct EventLoop<T: 'static> {
 /// `&EventLoop`.
 pub struct EventLoopWindowTarget<T: 'static> {
     pub(crate) p: platform_impl::EventLoopWindowTarget<T>,
-    pub(crate) _marker: ::std::marker::PhantomData<*mut ()>, // Not Send nor Sync
+    pub(crate) _marker: PhantomData<*mut ()>, // Not Send nor Sync
+}
+
+/// Object that allows building the event loop.
+///
+/// This is used to make specifying options that affect the whole application
+/// easier. But note that constructing multiple event loops is not supported.
+#[derive(Debug, Clone, Default)]
+pub struct EventLoopBuilder<T: 'static> {
+    pub(crate) platform_specific: platform_impl::PlatformSpecificEventLoopAttributes,
+    _p: PhantomData<T>,
+}
+
+impl EventLoopBuilder<()> {
+    /// Start building a new event loop.
+    #[inline]
+    pub fn new() -> Self {
+        Self::with_user_event()
+    }
+}
+
+impl<T> EventLoopBuilder<T> {
+    /// Start building a new event loop, with the given type as the user event
+    /// type.
+    #[inline]
+    pub fn with_user_event() -> Self {
+        Self {
+            platform_specific: Default::default(),
+            _p: PhantomData,
+        }
+    }
+
+    /// Builds a new event loop.
+    ///
+    /// ***For cross-platform compatibility, the `EventLoop` must be created on the main thread.***
+    /// Attempting to create the event loop on a different thread will panic. This restriction isn't
+    /// strictly necessary on all platforms, but is imposed to eliminate any nasty surprises when
+    /// porting to platforms that require it. `EventLoopBuilderExt::any_thread` functions are exposed
+    /// in the relevant `platform` module if the target platform supports creating an event loop on
+    /// any thread.
+    ///
+    /// Usage will result in display backend initialisation, this can be controlled on linux
+    /// using an environment variable `WINIT_UNIX_BACKEND`. Legal values are `x11` and `wayland`.
+    /// If it is not set, winit will try to connect to a wayland connection, and if it fails will
+    /// fallback on x11. If this variable is set with any other value, winit will panic.
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **iOS:** Can only be called on the main thread.
+    #[inline]
+    pub fn build(&mut self) -> EventLoop<T> {
+        EventLoop {
+            event_loop: platform_impl::EventLoop::new(&self.platform_specific),
+            _marker: PhantomData,
+        }
+    }
 }
 
 impl<T> fmt::Debug for EventLoop<T> {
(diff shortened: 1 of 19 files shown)


src/platform_impl/macos/app_delegate.rs
```diff
@@ -13,12 +13,8 @@ use std::{
 static AUX_DELEGATE_STATE_NAME: &str = "auxState";
 
 pub struct AuxDelegateState {
-    /// We store this value in order to be able to defer setting the activation policy until
-    /// after the app has finished launching. If the activation policy is set earlier, the
-    /// menubar is initially unresponsive on macOS 10.15 for example.
```
madsmtm, line 18: This comment was merged into the one in `apply_activation_policy`.

src/platform_impl/macos/app_delegate.rs
```diff
@@ -13,12 +13,8 @@ use std::{
...
     pub activation_policy: ActivationPolicy,
-
-    pub create_default_menu: bool,
+    pub default_menu: bool,
```
madsmtm, line 17: Renamed for consistency

src/platform/unix.rs
```diff
@@ -93,100 +92,42 @@ impl<T> EventLoopWindowTargetExtUnix for EventLoopWindowTarget<T> {
...
-/// Additional methods on `EventLoop` that are specific to Unix.
-pub trait EventLoopExtUnix {
-    /// Builds a new `EventLoop` that is forced to use X11.
-    ///
-    /// # Panics
-    ///
-    /// If called outside the main thread. To initialize an X11 event loop outside
-    /// the main thread, use [`new_x11_any_thread`](#tymethod.new_x11_any_thread).
+/// Additional methods on [`EventLoopBuilder`] that are specific to Unix.
+pub trait EventLoopBuilderExtUnix {
```
kchibisov, line 96: Could we maybe separate `X11` and `Wayland` here? I'd like to see something like that for Wayland, since it'd allow to unsubscribing from events that winit handles but downstream doesn't care (which will require extra protocols, etc, etc). I'm not sure that it'll be ergonomic with just `ExtUnix`.

kchibisov (changes requested):
I'm in favor of adding something like that to give ability to opt-out form specific handling winit is performing on Wayland, like `pointer grabs, extra motions, etc, etc.`, since those are mostly required for games, but certain clients don't need those, so opt-ting out when building event loop would be nice.

However I don't think that `X11` and `Wayland` share the same options, since those are completely different backends and you'd end up with `_wayland` methods and Wayland related structs if you'd continue to use unified struct...

src/platform/unix.rs
madsmtm: I agree that we should consider splitting the traits in `platform::unix` into separate traits, but would you be okay with doing that in a separate PR? I can whip one up if you want.

kchibisov: I mean not all traits, but precisely this one, so we would have proper API? Though, it's not like it'll change a lot, you'd just end up rewriting changelog entry, since you add one thing and remove it right away. I think doing it properly right away is the way to go, and it shouldn't be that hard to implement.

madsmtm: All the trait methods in `platform::unix` (except `MonitorHandleExtUnix::native_id`) are already `cfg`'d to the specific backend, and if we don't intend changing those I would argue we should keep the current design for consistency.

And what about `EventLoopBuilderExtUnix::any_thread`? Should we keep `EventLoopBuilderExtUnix` for that, or should it now be present in both traits?

kchibisov: I've rethought things a bit and I'd suggest to go with the current `Unix` approach, since we'd need this trait anyway to have code composition seamlessly in mixed x11/wayland environments.

So let's keep it for later design.

maroider (commented):
I'm not a massive fan of reflowing (doc) comments to fit what appears to be 80 columns, but that's not a hill I want to die on either. The changelog entries also look weird to me.

Should we mark the changelog entries as `**Breaking: **`?

Otherwise, this looks good.

madsmtm:
Sorry about the reflow, I've changed it back, same with the changelog entries. In lack of this style choice being automatic (https://github.com/rust-lang/rustfmt/issues/3347) I just did it on instinct 🙄.

I've marked the entries as breaking, well spotted!

src/platform_impl/macos/event_loop.rs
```diff
@@ -114,8 +120,23 @@ pub struct EventLoop<T: 'static> {
     _callback: Option<Rc<RefCell<dyn FnMut(Event<'_, T>, &RootWindowTarget<T>, &mut ControlFlow)>>>,
 }
 
+#[derive(Debug, Copy, Clone, PartialEq, Hash)]
+pub(crate) struct PlatformSpecificEventLoopAttributes {
```
ArturKovacs, line 124: The entire `platform_impl` module is non-public so if I understand correctly, then everything within it can be just `pub` and it will only be visible to the crate. It would probably look cleaner that way.

https://github.com/rust-windowing/winit/blob/001fb7ef60efe989b54b07024b660fd07e94afc3/src/lib.rs#L158

ArturKovacs (commented):
I read the macos implementation and I only had the following remark.

src/platform_impl/macos/event_loop.rs
ArturKovacs: You are making a good point. I'll leave it up to you.

msiglreith (commented):
Fine from my side!
The extension methods not using `with_` prefixes feels a bit inconsistent to me with the rest of the existing builder API in `winit`.

madsmtm:
> The extension methods not using `with_` prefixes feels a bit inconsistent to me with the rest of the existing builder API in `winit`.

You're right, I've changed these now

src/platform_impl/macos/event_loop.rs
madsmtm: I'll leave it as is for now at least, because I prefer the explicitness

src/platform/windows.rs
```diff
@@ -9,62 +9,61 @@ use winapi::shared::windef::{HMENU, HWND};
...
-    fn new_dpi_unaware() -> Self
-    where
-        Self: Sized;
+    fn with_any_thread(&mut self, any_thread: bool) -> &mut Self;
 
-    /// Creates a DPI-unaware event loop off of the main thread.
+    /// Whether to enable process-wide DPI awareness.
+    ///
+    /// By default, we will attempt to enable process-wide DPI awareness. If
+    /// that's undesirable, you can create disable it with this function.
```
filnet, line 35: typo: "...you can create disable it..."

src/platform_impl/linux/mod.rs
```diff
@@ -568,13 +591,28 @@ impl<T: 'static> Clone for EventLoopProxy<T> {
 }
 
 impl<T: 'static> EventLoop<T> {
-    pub fn new() -> EventLoop<T> {
-        assert_is_main_thread("new_any_thread");
+    pub(crate) fn new(attributes: &PlatformSpecificEventLoopAttributes) -> Self {
+        if !attributes.any_thread && !is_main_thread() {
+            panic!(
+                "Initializing the event loop outside of the main thread is a significant \
+                 cross-platform compatibility hazard. If you really, absolutely need to create an \
```
filnet, line 598: I would replace the comma by "and" or remove "really".

src/platform_impl/windows/event_loop.rs
```diff
@@ -113,47 +113,43 @@ pub struct EventLoop<T: 'static> {
...
+impl<T: 'static> EventLoop<T> {
+    pub(crate) fn new(attributes: &PlatformSpecificEventLoopAttributes) -> Self {
         let thread_id = unsafe { processthreadsapi::GetCurrentThreadId() };
-        if thread_id != main_thread_id() {
-            panic!(concat!(
+
+        if !attributes.any_thread && thread_id != main_thread_id() {
+            panic!(
                 "Initializing the event loop outside of the main thread is a significant \
                  cross-platform compatibility hazard. If you really, absolutely need to create an \
```
filnet, line 144: same

src/platform/unix.rs
```diff
@@ -93,100 +92,42 @@ impl<T> EventLoopWindowTargetExtUnix for EventLoopWindowTarget<T> {
...
-    where
-        Self: Sized;
-}
-
-fn wrap_ev<T>(event_loop: LinuxEventLoop<T>) -> EventLoop<T> {
-    EventLoop {
-        event_loop,
-        _marker: std::marker::PhantomData,
-    }
+    /// This method bypasses the cross-platform compatibility requirement that
```
filnet, line 107: "This method allows to bypass..." ?

src/platform/macos.rs
```diff
@@ -179,36 +178,63 @@ impl WindowBuilderExtMacOS for WindowBuilder {
...
+    /// Menu creation is enabled by default.
     ///
-    /// The default menu creation is enabled by default.
+    /// # Example
     ///
-    /// This function only takes effect if it's called before calling
-    /// [`run`](crate::event_loop::EventLoop::run) or
-    /// [`run_return`](crate::platform::run_return::EventLoopExtRunReturn::run_return)
-    fn enable_default_menu_creation(&mut self, enable: bool);
+    /// Disable creating a default menubar:
```
filnet, line 210: typo: use period instead of colon ?

src/event_loop.rs
```diff
@@ -42,7 +43,62 @@ pub struct EventLoop<T: 'static> {
 /// `&EventLoop`.
 pub struct EventLoopWindowTarget<T: 'static> {
     pub(crate) p: platform_impl::EventLoopWindowTarget<T>,
-    pub(crate) _marker: ::std::marker::PhantomData<*mut ()>, // Not Send nor Sync
+    pub(crate) _marker: PhantomData<*mut ()>, // Not Send nor Sync
+}
+
+/// Object that allows you to build the event loop.
```
filnet, line 49: avoid "you" ?

src/platform/unix.rs
madsmtm: I went with the same text as `EventLoopBuilderExtWindows::with_any_thread`

madsmtm:
Thanks for the review @filnet!

ArturKovacs (approved):

kchibisov (approved):

kchibisov:
The build failure is unrelated and present on master.

PR #2138, opened 2022-01-05
Fix some invalid `msg_send!` usage

madsmtm:
Change a few types for more correct behaviour (e.g. using `bool` may be undefined behaviour, one should use `BOOL`).

The only critical thing here is the stuff in `unmark_text` (introduced in #2119), that part is certainly undefined behaviour, and will probably cause problems somewhere down the line.

### DIFF
--- a/src/platform_impl/macos/util/mod.rs
+++ b/src/platform_impl/macos/util/mod.rs
@@ -114,7 +114,8 @@ pub unsafe fn app_name() -> Option<id> {
 }
 
 pub unsafe fn superclass<'a>(this: &'a Object) -> &'a Class {
-    msg_send![this, superclass]
+    let superclass: *const Class = msg_send![this, superclass];
+    &*superclass
 }
 
 pub unsafe fn create_input_context(view: id) -> IdRef {
--- a/src/platform_impl/macos/util/mod.rs
+++ b/src/platform_impl/macos/util/mod.rs
@@ -40,10 +40,9 @@ impl IdRef {
         IdRef(inner)
     }
 
-    #[allow(dead_code)]
     pub fn retain(inner: id) -> IdRef {
         if inner != nil {
-            let () = unsafe { msg_send![inner, retain] };
+            let _: id = unsafe { msg_send![inner, retain] };
         }
         IdRef(inner)
     }
@@ -76,10 +75,7 @@ impl Deref for IdRef {
 
 impl Clone for IdRef {
     fn clone(&self) -> IdRef {
-        if self.0 != nil {
-            let _: id = unsafe { msg_send![self.0, retain] };
-        }
-        IdRef(self.0)
+        IdRef::retain(self.0)
     }
 }
 
@@ -118,8 +114,7 @@ pub unsafe fn app_name() -> Option<id> {
 }
 
 pub unsafe fn superclass<'a>(this: &'a Object) -> &'a Class {
-    let superclass: id = msg_send![this, superclass];
-    &*(superclass as *const _)
+    msg_send![this, superclass]
 }
 
 pub unsafe fn create_input_context(view: id) -> IdRef {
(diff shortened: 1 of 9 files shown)


src/platform_impl/macos/view.rs
```diff
@@ -442,7 +442,9 @@ extern "C" fn unmark_text(this: &Object, _sel: Sel) {
     unsafe {
         let marked_text: id = *this.get_ivar("markedText");
         let mutable_string = marked_text.mutableString();
-        let _: () = msg_send![mutable_string, setString:""];
+        let s: id = msg_send![class!(NSString), new];
+        let _: () = msg_send![mutable_string, setString: s];
+        let _: () = msg_send![s, release];
```
ArturKovacs, line 447: You probably know that `let _: () = ` can be written as `let () = `

But I understand that keeping consistency with the old stuff is probably better here.

ArturKovacs (commented):
Thank you for making these improvements.

There is one thing that bugs me. The fact that we have to write

```rust
my_objc_bool != NO
```

I know that it's necessary to compare against NO, but then we have a double negation for every single truthy check.

Here's an alternative:
```rust
trait BoolYeah {
    /// This is true when `self` is not `NO`
    /// 
    /// Note that this different from
    /// `self == YES`
    /// because YES is the value 1 but
    /// a BOOL may be any signed 8bit value
    fn yeah(self) -> bool;
    fn no(self) -> bool;
}
impl BoolYeah for objc::runtime::BOOL {
    #[inline]
    fn yeah(self) -> bool {
        self != 0
    }
    #[inline]
    fn no(self) -> bool {
        self == 0
    }
}
```

So that we can write

```rust
if window_has_focus.yeah() {
```

aloucks:
> `trait BoolYeah`

It would be more clear to have `is_true()` and `is_false()` instead of `yeah()` and `no()`.

ArturKovacs:
Just to throw another option into the mix, what about `truthy()` and `falsy()`? This naming comes from JavaScript.

maroider:
Maybe just a conversion, e.g. `.to_bool()`, would be good enough.

src/platform_impl/macos/util/mod.rs
```diff
@@ -118,8 +114,7 @@ pub unsafe fn app_name() -> Option<id> {
 }
 
 pub unsafe fn superclass<'a>(this: &'a Object) -> &'a Class {
-    let superclass: id = msg_send![this, superclass];
-    &*(superclass as *const _)
+    msg_send![this, superclass]
 }
```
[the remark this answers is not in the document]
madsmtm: `objc::runtime::Class` is equivalent to `objc_class`, not the typedef `Class` you're referring to. Yes, confusing.

Anyhow, I found these issues by using a modified version of [`objc`'s `"verify_message"` feature](https://docs.rs/objc/0.2.7/objc/#message-type-verification); I rewrote this because the `msg_send!` was returning `id`, which does not have the same encoding as `*const Class`.

So this is not a soundness-changing issue (because we convert the `id` to `*const Class` afterwards), though I think the code is clearer this way IMO.

We can also do:
```rust
let superclass: *const Class = msg_send![this, superclass];
&*superclass
```

src/platform_impl/macos/view.rs
madsmtm: Yeah, this was for consistency, though also because I find `let _: () = ` easier to parse.

src/platform_impl/macos/util/mod.rs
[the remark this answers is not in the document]
ArturKovacs: Yes, this one seems much better!

src/platform_impl/macos/view.rs
ArturKovacs: I really like `let () =`. I feel cheeky when I write it down.

ArturKovacs:
Alright, I'm looking forward to moving to your fork of objc then.

madsmtm:
> Alright, I'm looking forward to moving to your fork of objc then.

Thanks! Though intending to move to something else in the unknown future shouldn't be a blocker for adding `BoolYeah` now, just not something I'd want to do in this PR in any case.

ArturKovacs (approved):
Awesome!

PR #1788, opened 2020-12-05
New keyboard API for Windows

ArturKovacs:
- [x] Tested on all platforms changed
- [x] Compilation warnings were addressed
- [x] `cargo fmt` has been run on this branch
- [x] `cargo doc` builds successfully
- [ ] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users
- [ ] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior
- [x] Created or updated an example program if it would help users understand this functionality
- [x] Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented

This contains the Windows implementation of the keyboard API discussed at #753

### Update

As of writing this the API implemented by this PR is as follows. 

<details>

<summary>Click for the code</summary>

-----------------

`ReceivedCharacter(char)` was removed. Its functionality is now covered by `KeyboardInput` and `ReceivedImeText`.

```rust

pub enum WindowEvent {
    /// The user commited an IME string for this window.
    ///
    /// This is a temporary API until #1497 gets completed. See:
    /// https://github.com/rust-windowing/winit/issues/1497
    ReceivedImeText(String),

    KeyboardInput {
        device_id: DeviceId,
        event: KeyEvent,
        is_synthetic: bool,
    },

   // ...
}


/// Contains the platform-native physical key identifier (aka scancode)
#[derive(Debug, Eq, PartialEq, Hash, Serialize, Deserialize)]
pub enum NativeKeyCode {
    Unidentified,
    Windows(u16),
    MacOS(u16),
    XKB(u32),
}

/// Represents the position of a key independent of the
/// currently active layout.
/// 
/// Conforms to https://www.w3.org/TR/uievents-code/
/// 
/// For most keys the name comes from their US-layout equivalent.
#[derive(Debug, Eq, PartialEq, Hash, Serialize, Deserialize)]
#[non_exhaustive]
pub enum KeyCode {
    A,
    B,
    // ...
    ControlLeft,
    ControlRight,
    // ...
    F1,
    F2,
    // ...
    
    /// The native scancode is provided (if available) in order
    /// to allow the user to specify keybindings for keys which
    /// are not defined by this API.
    Unidentified(NativeKeyCode)
}

pub trait KeyCodeExtScancode {
    /// The raw value of the platform specific physical key identifier.
    fn to_scancode(self) -> u32;
    /// Constructs a `KeyCode` from a platform specific physical key identifier.
[...]

PR #2143, opened 2022-01-10
New keyboard rebase on master

ArturKovacs:
This should *not* be merged. This is only intended for allowing review. Once this is accepted, I plan on force-pushing this branch to winit's `new-keyboard` branch.

To make the review process easier, here's the list of the conflicting files
(I only changed these during the rebase)
```
	both modified:   Cargo.toml
	both modified:   examples/multithreaded.rs
	both modified:   src/event.rs
	both modified:   src/platform/windows.rs
	deleted by them: src/platform_impl/windows/event.rs
	both modified:   src/platform_impl/windows/event_loop.rs
	both modified:   src/platform_impl/windows/window.rs
```


src/platform_impl/windows/event_loop.rs
```diff
@@ -1640,68 +1648,23 @@ unsafe fn public_window_callback_inner<T: 'static>(
...
-                        input: KeyboardInput {
-                            scancode,
-                            virtual_keycode,
-                            state: Released,
-                            modifiers: event::get_key_mods(),
-                        },
-                        is_synthetic: true,
-                    },
-                })
-            }
```
ArturKovacs, line 1669: I was a bit hesitant with this. Why did we send a RELEASED event for the pressed keys during WM_SETFOCUS?

src/platform_impl/windows/event_loop.rs
```diff
@@ -1734,17 +1697,12 @@ unsafe fn public_window_callback_inner<T: 'static>(
...
-                None => winuser::DefWindowProcW(window, msg, wparam, lparam),
+                None => result = ProcResult::DefWindowProc,
             }
         }
 
-        winuser::WM_DROPFILES => {
-            // See `FileDropHandler` for implementation.
-            0
-        }
-
```
maroider, line 1747: Why is this being removed again? It's present in #1788 as well.

maroider (commented):
LGTM

src/platform_impl/windows/event_loop.rs
maroider: As I understand it, it's to notify the user that they ought to treat the keys as released when focus is lost.

src/platform_impl/windows/event_loop.rs
```diff
@@ -108,6 +113,18 @@ impl<T> ThreadMsgTargetData<T> {
...
+#[derive(Clone, Copy)]
+pub(crate) enum ProcResult {
+    DefWindowProc,
+    Value(isize),
+}
+impl Default for ProcResult {
+    fn default() -> Self {
+        ProcResult::DefWindowProc
+    }
+}
```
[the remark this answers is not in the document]
ArturKovacs: Yup, this is new, I'm not even sure if I ended up using it, but it seemed sensible.

src/platform_impl/windows/event_loop.rs
ArturKovacs: Because it's handled by this:
https://github.com/rust-windowing/winit/pull/2143/files#diff-edacb93792c7cf2d4cfe6f4ff2259781ed3c13e060fcc381e9cbb30ce481a336R780

src/platform_impl/windows/event_loop.rs
ArturKovacs: But this happened on SETFOCUS not just KILLFOCUS.

maroider: Welp, that's strange. I'd say we don't do anything until someone complains.

ArturKovacs:
The formatting was incorrect, so I ran `cargo fmt` and amended the changes to the commit named "Fix the drag_window example"

maroider:
> The formatting was incorrect, so I ran cargo fmt and amended the changes to the commit named "Fix the drag_window example"

You can do `git commit --fixup HEAD~1` and then `git rebase -i --autosquash HEAD~2` to fix the previous commit.

ArturKovacs:
I already committed the changes so I can't do this fixup thingy. I guess I could simply just squash the two commits.

maroider:
Squashing seems alright to me.

fredizzimo:
Hi,

We have been using these changes in [Neovide](https://github.com/neovide/neovide) for quite some time now, however there has been a long standing bug, which just recently got fixed by my pull request into Neovide's own fork of winit https://github.com/neovide/winit/pull/4.

The issue is that we had some random hangs, which was tracked down to the event handler. There were some PeekMessage calls inside the keypress handling calls, which caused reentrancy, which in turn caused deadlocks. Some of the bug reports were like this https://github.com/neovide/neovide/issues/1010, or this https://github.com/neovide/neovide/issues/1014.

I myself had a more reproducible way of causing the hang, using either a tiling window manager like Nog, or Microsoft Powertoys, and then quickly change the position of the window. It usually did not take many times for it to hang. However, I have since then moved on to a more powerful computer and Windows 11, and I'm really struggling to reproduce the original bug with that system even without my fixes, so it  seems like it affects some systems worse than others. But the bug was still quite frequent for @Kethku, and since the pull request was merged it has no longer happened.

Unfortunately I don't remember the exact details of exactly where it was hanging, since I made the changes a few months ago, and used the changes on my personal machine only, until I sent the pull request to Neovide. But I think my description in https://github.com/neovide/winit/pull/4 is correct.

We were planning to run with these changes for a while still to verify that everything works correctly, and that it really fixes the issue, but now we saw this pull request, and it looks like it's getting closer to get merged, I want to ask what action we should take, since there's a risk that the same bug affects other applications as well. Should I send a pull request against this branch?

maroider:
Thanks for reporting this, holding a lock while calling into `PeekMessageW` does indeed seem like a deadlock hazard.

I think you should PR the patch against `new-keyboard`.

Kethku:
Awesome. I will look into making a PR soon.

Btw we've been using a (now pretty outdated) fork which has the changes from all 3 of the new keyboard branches (windows macos and linux). Is there such a branch in the main winit repository? Would it make sense for us to create one for testing purposes? Such a thing would make it WAY easier for us to continue testing the new functionality which seems ideal before merging this PR into main winit for everyone. What are your thoughts?

maroider:
The plan is for all the other branches to get merged into `new-keyboard` before anything lands in `master`.

Kethku:
Done. PR with the fix is now open against new-keyboard. Let me know if you need anything else from this fix.

Super excited for these changes. We've been using them for about 4 months now and they are a massive improvement.

Kethku:
Hey is there any way I (or other contributors from the neovide project) could help push this forward? Would it help to make PRs from each of the other keyboard branches into new-keyboard and or create a temporary branch which has all of the new keyboard api implementations for testing purposes?

Such a thing could be maintained and recreated (by us) any time the upstream PRs change just so we can get more testing of the new api before merging it into master.

I'm not trying to put any extra time pressure or such on the integration, just looking for ways we could pitch in to help.

archseer:
@maroider I understand there's been a fair bit of burnout over the `new-keyboard` rework but I'd like to echo @Kethku's sentiment: Is there any way for us to pitch in and help with the implementation / rebases / fixes?

It might make sense to merge the remaining web/linux/macOS PRs into `new-keyboard` as is -- even in it's current unfinished form they're already a significant improvement over `master` (and as demonstrated by Neovide, reliable enough to use: https://github.com/neovide/neovide/blob/bb1255d30363d41f66c1fbdf52344b0c3021bd53/Cargo.toml#L42-L43=), this would allow us to use the branch downstream and test out the code.

It would also make it easier to work on remaining items incrementally, rather than delivering everything in one large PR.

maroider:
I've unfortunately been in a bit of a funk since some time in February. I don't know if I'll be able to push through anything in the short term, but I do agree that it's imperative that we get *something* merged into some unified branch. I'll *try* to get back to you within the week with some sort of action plan, but I won't promise anything today.

Kethku:
Take care of yourself. All of this can wait

ArturKovacs:
To add my two cents: as far as I know the only blocking issue is to get IME to work on the X11 implementation. I would say that it must be done before we merge because leaving IME support out would be a significant regression compared to the current state. (ReceivedCharacter sends characters that were inserted through IME)

archseer:
It would help with the cascade of branches that all need to be independently rebased right now:

`rust-windowing/winit:new-keyboard` <- [`maroider:new-keyboard-linux`](https://github.com/maroider/winit/tree/new-keyboard-linux) <- [`mahkoh:x11-nt`](https://github.com/mahkoh/winit/tree/x11-nt) <- [`ArturKovacs/artur-new-keyboard-linux`](https://github.com/ArturKovacs/winit/tree/artur-new-keyboard-linux)

kchibisov:
@archseer the IME for all present platforms is implemented #2243 and could be merged to master pretty soon afaics.

ArturKovacs:
@maroider I merged master into this again. I tested it and everything *seems* to be working.

maroider (approved):

maroider:
I'm closing this now that the rebase has happened

PR #2144, opened 2022-01-10
Add Foreground and Background events for iOS

HackerFoo:

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -76,6 +76,12 @@ pub enum Event<'a, T: 'static> {
     /// Emitted when the application has been resumed.
     Resumed,
 
+    /// Emitted when the application will enter the foreground.
+    Foreground,
+
+    /// Emitted when the application has entered the background.
+    Background,
+
     /// Emitted when all of the event loop's input events have been processed and redraw processing
     /// is about to begin.
     ///
@@ -138,6 +144,8 @@ impl<T: Clone> Clone for Event<'static, T> {
             LoopDestroyed => LoopDestroyed,
             Suspended => Suspended,
             Resumed => Resumed,
+            Foreground => Foreground,
+            Background => Background,
         }
     }
 }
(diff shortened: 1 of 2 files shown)


madsmtm:
Hmm, how does this differ from `WindowEvent::Focused`?

Also, this is something touching the public API, and will need discussion to figure out if it's relevant for at the very least a few platforms (see also the issues on Android: https://github.com/rust-windowing/raw-window-handle/issues/84).

HackerFoo:
`WindowEvent::Focused` is sent in response to [`becomeKey()/resignKey()`](https://developer.apple.com/documentation/uikit/uiwindow), which seems to be for indicating the most recently active window, while `WindowEvent::Foreground/Background` events indicate transitions in an [app's lifecycle](https://developer.apple.com/documentation/uikit/app_and_environment/managing_your_app_s_life_cycle):

iOS App Lifecycle

It should probably be sent at the `onResume()` and `onPause()` transitions for Android:

Android App Lifecycle

The purpose is to allow the app to switch from and to a low power state, rather than for tracking the user's focus. For iOS in particular, an app may not use the GPU while in the background, or it will be killed by the OS.

madsmtm:
The implementation looks good, though I must admit I have a hard time grokking the different lifecycle events - maybe we should have some documentation that directly maps `winit`s lifecycle events to Android/iOS lifecycle events, kinda like you just described, just for all of them?

Anyhow, I think these events belong in `winit`, but it's harder to justify this PR without the Android implementation; could you be persuaded to attempt that?

> For iOS in particular, an app may not use the GPU while in the background, or it will be killed by the OS.

Use-cases like these would be useful to have documented!

HackerFoo:
Sorry, I don't have Android set up right now, and I can't justify putting the time in to properly test it. For someone who has `winit` running on Android, it probably wouldn't take much time to add Android support for these events.

ArthurKValladares:
I will voice support for this PR. I work for a company planning on releasing a production product using winit on iOS, and we need this functionality to be able to properly handle backgrounding/foregrounding.

madsmtm:
Superceded by https://github.com/rust-windowing/winit/pull/2980

Issue #2135, opened 2022-01-05
Ideas for solving changelog conflicts


madsmtm:
Most changes warrant an entry in `CHANGELOG.md`, however, this means that we get `git` conflicts whenever a PR is merged; this is tough on contributors, because they have to keep their PRs "up to date" to have a chance of being merged.

A few ideas for solving this problem, for fun let's react with the corresponding emoji for the solution(s) you like the most:

🎉: The maintainer that merges solves the conflict before merging.

❤️: We add changelog placeholders [like GitLab did in 2015](https://about.gitlab.com/blog/2015/02/10/gitlab-reduced-merge-conflicts-by-90-percent-with-changelog-placeholders/), and users just randomly pick one of these to add their entry to.

🚀: We set up a bot that does the merge when a maintainer says `@winitbot merge`, and in this process it fixes the changelog conflicts, kinda like Rust itself does with [@bors](https://github.com/bors). [This is hard work, not something I want to spearhead FYI]

👀: The PR author adds the changelog entry to the PR description, and the person doing a new release manually collects all of these together.

👎: We keep the status-quo.

Whichever solution we choose, I'll make sure to document it.

EDIT: Added the 👀 proposal

maroider:
> 🎉: The maintainer that merges solves the conflict before merging.

This seems reasonable when people check the "give contributors write access to the branch" checkbox when creating the PR. I don't know how we'd resolve merge conflicts in the absence of this permission.

madsmtm:
Right, we'll document that as well.

madsmtm:
@kchibisov suggested on Matrix that we remove `merge=union` so that old entries don't accidentally get put in the wrong place

ArturKovacs:
> we remove `merge=union` so that old entries don't accidentally get put in the wrong place

As far as I can tell, if we remove `merge=union`, then PRs will just have more conflicts in the changelog, but once "resolving" the conflict the item would still stay under the wrong release if a release is published while the PR is open.

Or would the changelog line in the PR  magically jump to the top after making the release? That doesn't seem likely. What am I missing here?

madsmtm:
The difference is that `merge=union` puts the entry under the wrong header _automatically_, while with the default settings it, a person has to manually resolve the conflict, hopefully in the process discovering that the entry is in the wrong place.

madsmtm:
[...]

PR #2145, opened 2022-01-10
Update changelog guidelines to prevent conflicts from blocking PRs

madsmtm:
Fixes https://github.com/rust-windowing/winit/issues/2135.

I didn't really feel this warranted an addition to the checklist in `PULL_REQUEST_TEMPLATE.md`, though we could do that as well?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+Please only add new entries to the top of this list, right below the `# Unreleased` header.
+
 # Unreleased
 
 - **Breaking:** Rename the `Exit` variant of `ControlFlow` to `ExitWithCode`, which holds a value to control the exit code after running. Add an `Exit` constant which aliases to `ExitWithCode(0)` instead to avoid major breakage. This shouldn't affect most existing programs.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 All notable changes to this project will be documented in this file.
 
-Please only add new entries to the top of this list, right below the `# Unreleased` header.
+Please keep one empty line before and after all headers. (This is required for `git` to produce a conflict when a release is made while a PR is open and the PR's changelog entry would go into the wrong section).
+
+And please only add new entries to the top of this list, right below the `# Unreleased` header.
 
 # Unreleased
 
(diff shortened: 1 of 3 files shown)


msiglreith (approved):

CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+Please only add new entries to the top of this list, right below the `# Unreleased` header.
+
+# Unreleased
```
ArturKovacs, line 7: Also add to this:

"There must always be at least one empty line after all section headers, including `# Unreleased`. This is so that git produces a conflict when a release is made while a PR is open and the PR's changelog entry would go into the wrong section."

ArturKovacs (changes requested):

CHANGELOG.md
madsmtm: I realize my explanation [here](https://github.com/rust-windowing/winit/issues/2135#issuecomment-1008907007) might have been a bit misleading on this point - without the newline(s), the diff would look like:
```diff
@@ -1,5 +1,8 @@
 # Unreleased
 
+# Version X
 - Old entry 1
 - Old entry 2
 - Old entry 3
```

Which would still cause a conflict

ArturKovacs: But not in this case:

```
# Unreleased
# Version

- Old entry 1
- Old entry 2
```

madsmtm: Oh, you're right! I've added something similar to what you wrote now.

CONTRIBUTING.md
```diff
@@ -34,9 +37,25 @@ Once your PR is open, you can ask for review by a maintainer of your platform. W
...
+
+If you believe a new release is warranted, you can make a pull-request with:
+- An updated version number (remember to change the version everywhere it is used).
+- A new section in the changelog (below the `# Unreleased` section).
+
+This gives contributors an opportunity to squeeze in an extra PR or two that they feel is valuable
+enough to warrant blocking the release a little.
+
+Once the PR is merged, a CI job will automatically release the new version. Remember that the release
+date in the changelog must be kept in check with the actual release date.
```
ArturKovacs, line 61: If we want to be pedantic, this is not true. A maintainer has to create a tag in order for the CI to make the release. The easiest way to do this is to make a new release from the GitHub Releases page and create a tag from there.

madsmtm: Heh, I would not say that's "being pedantic"; that's just being correct 😉 (I thought it worked the way I wrote it). I've updated the text now.

ArturKovacs: In that's case it's nice to clear that up. 😁

ArturKovacs (approved):

madsmtm:
Thanks for the review (both of you)!

PR #2146, opened 2022-01-11
Fix windows deadlock

Kethku:

Details about this change found here: https://github.com/rust-windowing/winit/pull/2143#issuecomment-1009408778

TLDR: We have been encountering a deadlock on windows when using a fork of the new keyboard changes in neovide. This change fixes that deadlock


fredizzimo:
I think we can probably remove the changelog, since the bug is specific to the the `new-keyboard` branch.

ArturKovacs:
Big thanks for submitting this fix! This issue went completely under our radar. Although I believe that the fix is incorrect. See my alternative solution for details: https://github.com/ArturKovacs/winit/pull/1

I spent the past two days more or less getting lost in learning about and trying to fight with the Windows event handling API, and I still don't feel confident about my understanding of it, so correct me if I'm wrong.

ArturKovacs:
My solution has been merged into #2143 which should get force pushed to `new-keyboard` in the foreseeable future so I'll close this for now.

PR #2147, opened 2022-01-11
macOS: Composition Event

komi1230:
This is related to #1497.

To make IME work fine in all languages, we should add new event for composition as discussed in #1497 and #2119.
This PR defines enum variant `Composition` and implement it in macOS.

I found some effort in #1404. (Thanks, @garasubo)
But this branch is very old and it's hard to sync with current master branch.

Thanks,

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - **Breaking:** Replaced `EventLoopExtMacOS` with `EventLoopBuilderExtMacOS` (which also has renamed methods).
 - **Breaking:** Replaced `EventLoopExtWindows` with `EventLoopBuilderExtWindows` (which also has renamed methods).
 - **Breaking:** Replaced `EventLoopExtUnix` with `EventLoopBuilderExtUnix` (which also has renamed methods).
+- On Wayland, fix resize and scale factor changes not being propagated properly.
 
 # 0.26.1 (2022-01-05)
 
--- a/src/platform_impl/linux/wayland/event_loop/mod.rs
+++ b/src/platform_impl/linux/wayland/event_loop/mod.rs
@@ -24,7 +24,7 @@ use crate::platform_impl::EventLoopWindowTarget as PlatformEventLoopWindowTarget
 use super::env::{WindowingFeatures, WinitEnv};
 use super::output::OutputManager;
 use super::seat::SeatManager;
-use super::window::shim::{self, WindowUpdate};
+use super::window::shim::{self, WindowRequest, WindowUpdate};
 use super::{DeviceId, WindowId};
 
 mod proxy;
@@ -249,7 +249,8 @@ impl<T: 'static> EventLoop<T> {
             }
 
             // Process 'new' pending updates.
-            self.with_state(|state| {
+            self.with_window_target(|window_target| {
+                let state = window_target.state.get_mut();
                 window_updates.clear();
                 window_updates.extend(
                     state
@@ -261,7 +262,8 @@ impl<T: 'static> EventLoop<T> {
 
             for (window_id, window_update) in window_updates.iter_mut() {
                 if let Some(scale_factor) = window_update.scale_factor.map(|f| f as f64) {
-                    let mut physical_size = self.with_state(|state| {
+                    let mut physical_size = self.with_window_target(|window_target| {
+                        let state = window_target.state.get_mut();
                         let window_handle = state.window_map.get(window_id).unwrap();
                         let mut size = window_handle.size.lock().unwrap();
 
@@ -294,7 +296,8 @@ impl<T: 'static> EventLoop<T> {
                 }
 
                 if let Some(size) = window_update.size.take() {
-                    let physical_size = self.with_state(|state| {
+                    let physical_size = self.with_window_target(|window_target| {
+                        let state = window_target.state.get_mut();
                         let window_handle = state.window_map.get_mut(window_id).unwrap();
                         let mut window_size = window_handle.size.lock().unwrap();
 
@@ -319,6 +322,15 @@ impl<T: 'static> EventLoop<T> {
 
                         // Mark that refresh isn't required, since we've done it right now.
                         window_update.refresh_frame = false;
+
+                        // Queue request for redraw into the next iteration of the loop, since
+                        // resize and scale factor changes are double buffered.
+                        window_handle
+                            .pending_window_requests
+                            .lock()
+                            .unwrap()
+                            .push(WindowRequest::Redraw);
+                        window_target.event_loop_awakener.ping();
 
                         physical_size
                     });
@@ -356,7 +368,8 @@ impl<T: 'static> EventLoop<T> {
             // The purpose of the back buffer and that swap is to not hold borrow_mut when
             // we're doing callback to the user, since we can double borrow if the user decides
             // to create a window in one of those callbacks.
-            self.with_state(|state| {
+            self.with_window_target(|window_target| {
+                let state = window_target.state.get_mut();
                 std::mem::swap(
                     &mut event_sink_back_buffer,
                     &mut state.event_sink.window_events,
@@ -381,7 +394,8 @@ impl<T: 'static> EventLoop<T> {
             for (window_id, window_update) in window_updates.iter() {
                 // Handle refresh of the frame.
                 if window_update.refresh_frame {
-                    self.with_state(|state| {
+                    self.with_window_target(|window_target| {
+                        let state = window_target.state.get_mut();
                         let window_handle = state.window_map.get_mut(window_id).unwrap();
                         window_handle.window.refresh();
                         if !window_update.redraw_requested {
@@ -526,9 +540,9 @@ impl<T: 'static> EventLoop<T> {
         &self.window_target
     }
 
-    fn with_state<U, F: FnOnce(&mut WinitState) -> U>(&mut self, f: F) -> U {
+    fn with_window_target<U, F: FnOnce(&mut EventLoopWindowTarget<T>) -> U>(&mut self, f: F) -> U {
         let state = match &mut self.window_target.p {
-            PlatformEventLoopWindowTarget::Wayland(window_target) => window_target.state.get_mut(),
+            PlatformEventLoopWindowTarget::Wayland(window_target) => window_target,
             #[cfg(feature = "x11")]
             _ => unreachable!(),
         };


kchibisov:
I'm not familiar with macOS, but I can help bringing and finishing IME handling (by resurrecting old branch and as well as pushing Wayland bits) and providing help downstream(adding inline ime into alacritty).

komi1230:
Thanks, @kchibisov. According to the discussion in https://github.com/rust-windowing/winit/issues/1497#issuecomment-699652584, the API design for IME seems to be fixed like [this commit](https://github.com/rust-windowing/winit/pull/2147/commits/55bbd191d1ec8e56ffcda92cbb360b2ce1e8740d).

I know this change is destructive, and this change should be done very carefully. I already found master version of Alacritty 
 would be broken if this was merged. The maintainer of Alacritty already know this. (ref https://github.com/alacritty/alacritty/issues/1101#issuecomment-1013600997)

If you agree with this design and you are willing to implement it for Wayland, you may have to cherry-pick one of commits in this PR. Should I split this PR into defining `CompositionEvent` and its implement of macOS ?

kchibisov:
>If you agree with this design and you are willing to implement it for Wayland, you may have to cherry-pick one of the commits in this PR. Should I split this PR into defining CompositionEvent and its implementation for macOS ?

Just wait until tomorrow so I rebase composition-event and as well alacritty support. You'll just open PR against that branch, and then we rollup all of the platform changes in one patch.

garasubo:
@komi1230 Thank you for working on MacOS side.
@kchibisov I can help the implementation and verification for x11. Let me know if you need support.

And I recently got a windows machine, so I might support windows implementation in near future.

kchibisov:
@komi1230 I've update `composition-event` branch, so please update your branch to make use of API the `composition-event` branch is using right now.

If you have any questions wrt that API or you're thinking that you need to expose something else to make things work properly on macOS feel free share.

komi1230:
Thanks. I updated this PR to correspond to `composition-event` branch.
And I agree with shortening the name of enum variant `Composition` into `IME`. This is more understandable and clear.

I have confirmed this works fine in macOS. I think this PR is ready to merge.

src/platform_impl/macos/view.rs
```diff
@@ -433,6 +460,36 @@ extern "C" fn set_marked_text(
...
+        let (cursor_position, text_offset) = (
+            selected_range.location as usize,
+            selected_range.length as usize,
+        );
+        state.ime_state = Some(ImeState {
+            text: preedit_string.clone(),
+            cursor_start: cursor_position,
+            cursor_end: cursor_position + text_offset,
+        });
+        if state.ime_state.is_none() {
```
kchibisov, line 477: I can't be `None`, since you just assigned `Some` to it. I'm not sure what you're doing here.

src/platform_impl/macos/view.rs
```diff
@@ -504,30 +561,29 @@ extern "C" fn insert_text(this: &Object, _sel: Sel, string: id, _replacement_ran
...
-        let mut events = VecDeque::with_capacity(characters.len());
+        if state.ime_state.is_some() {
+            AppState::queue_event(EventWrapper::StaticEvent(Event::WindowEvent {
+                window_id: WindowId(get_window_id(state.ns_window)),
+                event: WindowEvent::IME(IME::Commit(string.clone())),
+            }));
+            state.ime_state = None;
+            AppState::queue_event(EventWrapper::StaticEvent(Event::WindowEvent {
+                window_id: WindowId(get_window_id(state.ns_window)),
+                event: WindowEvent::IME(IME::Disabled),
```
kchibisov, line 577: The idea behind `IME::Disabled` is when your IME stops completely, so if you had some text in flight you'll abort inserting and also stop providing anything to IME. For example if you have two layouts, one is `en(us)` and the other one is `mozc` input, switching from `en(us)` to `mozc` should trigger `IME::Enabled`, since you don't need IME for `en(us)`, and so switching from `mozc` to `en(us)` should issue `IME::Disabled`, but not when you're editing with IME.

If you don't know when user triggers IME you should likely send IME::Enabled at start and then one on exit of a particular window. Though, I'd assume that there's a way to determine whether there's IME or not on macOS.

kchibisov (changes requested):
I'll leave the rest to @madsmtm , since I'm not a macOS expert(I don't use macOS).

kchibisov:
I've added support for new IME event handling into alacritty alacritty/alacritty#5790 . Though, I'm using my Wayland branch, but just cloning and pointing to your branch in alacritty's `Cargo.toml` should just work.

src/platform_impl/macos/view.rs
```diff
@@ -414,16 +422,35 @@ extern "C" fn selected_range(_this: &Object, _sel: Sel) -> NSRange {
...
+    unsafe {
+        let has_attr = msg_send![string, isKindOfClass: class!(NSAttributedString)];
+        let characters = if has_attr {
+            // This is a *mut NSAttributedString
+            msg_send![string, string]
+        } else {
+            // This is already a *mut NSString
+            string
+        };
+        let foo =
```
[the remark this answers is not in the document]
komi1230: This is a remnant from when I was developing. I will change the name of this variable.
This should be more understandable name.

src/platform_impl/macos/view.rs
komi1230: This was my mistake.

src/platform_impl/macos/view.rs
komi1230: I surveyed how to handle the notification about what IME has been changed, and I found this would be very difficult.

In order to do this, you have to implement protocol `NSTextInputClient` and class `NSTextInputContext`. Here protocol in Objective-C is like trait in Rust. 

About `NSTextInputClient`, you have to implement some functions like `setMarkedText`, `insertText` and `validAttributesForMarkedText` with subclassing `NSView`. This is described in [Apple's official developer guide about NSTextInputClient](https://developer.apple.com/documentation/appkit/nstextinputclient#overview). So now you can handle marked text with already implemented functions.

But about `NSTextInputContext`, this is used to handle the information about what IME has been changed. This class is needed to catch notification about IME. This is still not implemented in [cocoa](https://github.com/servo/core-foundation-rs/tree/master/cocoa).

That's why I only remove `IME::Disabled` in this PR.

komi1230:
Thanks, @kchibisov. I appreciate you for your great work!! Now I can find IME is properly working in my Alacritty.

https://user-images.githubusercontent.com/27967726/149905048-df62dafa-e512-4c91-87b4-de27b626822c.mov

kchibisov:
>Thanks, @kchibisov. I appreciate you for your great work!! Now I can find IME is properly working in my Alacritty.

It seems like cursor position is off on your gif. When the cursor is in the end of string(so while you're typing) it shouldn't be visible, however when you start moving with left and right arrow keys it should draw hollow block cursor. Also, make sure that those offsets are in bytes and not in chars.

src/platform_impl/macos/view.rs
```diff
@@ -414,16 +437,44 @@ extern "C" fn selected_range(_this: &Object, _sel: Sel) -> NSRange {
...
+            msg_send![string, string]
+        } else {
+            // This is already a *mut NSString
+            string
+        };
+        let utf8_sequence =
+            slice::from_raw_parts(characters.UTF8String() as *const c_uchar, characters.len());
+        str::from_utf8_unchecked(utf8_sequence).to_string()
+    }
+}
```
ArturKovacs, line 454: - I think this function should be moved to the `util` module, so that it can be re-used from other places
- The function itself should be marked unsafe, because it assumes that the `string` is either an `NSAttributedString` or an `NSString`. (And the function doesn't check if this assumption is true.)

src/platform_impl/macos/view.rs
```diff
@@ -414,16 +437,44 @@ extern "C" fn selected_range(_this: &Object, _sel: Sel) -> NSRange {
     util::EMPTY_RANGE
 }
 
+fn id_to_string(string: id) -> String {
+    unsafe {
+        let has_attr = msg_send![string, isKindOfClass: class!(NSAttributedString)];
+        let characters = if has_attr {
```
ArturKovacs, line 443: Here the compiler infers that `has_attr` is a `bool` but that's wrong. [`isKindOfClass`](https://developer.apple.com/documentation/objectivec/1418956-nsobject/1418511-iskindofclass/) returns a `BOOL` which is not the same as a `bool`. Important: to check if a `BOOL` is true, you must compare it with 0 or compare it with `NO`.

```suggestion
        let has_attr: BOOL = msg_send![string, isKindOfClass: class!(NSAttributedString)];
        let characters = if has_attr != NO {
```

komi1230:
@kchibisov When I set `cursor_start` as byte-wise length of string, the cursor has been in the end of string and now invisible. In addition to this, I fixed some behavior of IME when repeating key without making side effect on window.
Could you check again ?

src/platform_impl/macos/view.rs
```diff
@@ -433,6 +484,35 @@ extern "C" fn set_marked_text(
             marked_text.initWithString(string);
         };
         *marked_text_ref = marked_text;
+
+        // Update ViewState with new marked text
+        let state_ptr: *mut c_void = *this.get_ivar("winitState");
+        let state = &mut *(state_ptr as *mut ViewState);
+        let preedit_string = id_to_string(string);
+
+        // Notify IME is active if application still doen't know it.
```
ArturKovacs, line 493: Typo
```suggestion
        // Notify IME is active if application still doesn't know it.
```

src/platform_impl/macos/view.rs
```diff
@@ -670,29 +764,37 @@ extern "C" fn key_down(this: &Object, _sel: Sel, event: id) {
...
+        match virtual_keycode.unwrap() {
+            VirtualKeyCode::Back
+            | VirtualKeyCode::Up
+            | VirtualKeyCode::Down
+            | VirtualKeyCode::Left
+            | VirtualKeyCode::Right => {}
+            _ => {
+                AppState::queue_event(EventWrapper::StaticEvent(window_event));
+            }
+        };
```
ArturKovacs, line 777: What kind of side effect are you refferring to?

The application will have to filter keyboard input during IME anyways. For example the application must ignore text input through the keyboard while the IME is active. So I think we should send all keyboard input to the application and let it decide what it wants to do with the keyboard events.

src/platform_impl/macos/view.rs
```diff
@@ -670,29 +764,37 @@ extern "C" fn key_down(this: &Object, _sel: Sel, event: id) {
...
+
+        // Emit `ReceivedCharacter` for key repeats
+        if is_repeat == YES {
```
ArturKovacs, line 780: This is not guaranteed to work. You should compare the `BOOL` with `NO` or compare it with 0.
```suggestion
        if is_repeat != NO {
```

ArturKovacs (changes requested):

src/platform_impl/macos/view.rs
komi1230: Exactly.

src/platform_impl/macos/view.rs
komi1230: I'm not inexperienced on macOS, and I thought this would return `YES`. 
I fixed. Thanks.

src/platform_impl/macos/view.rs
komi1230: Exactly.

src/platform_impl/macos/view.rs
```diff
@@ -286,13 +296,26 @@ extern "C" fn init_with_winit(this: &Object, _sel: Sel, state: *mut c_void) -> i
...
+            ];
+            // About IME change
+            let input_source_did_changed_notification_name = IdRef::new(
+                NSString::alloc(nil)
+                    .init_str("NSTextInputContextKeyboardSelectionDidChangeNotification"),
+            );
+            let _: () = msg_send![
+                notification_center,
+                addObserver: this
+                selector: sel!(initWithClient:)
```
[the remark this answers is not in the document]
komi1230: I'm not familiar with how to use selector, and I want your advice. I imagine this selector would be like a hook on some specified method. So is this better to be hooked on `keyDown` ?
Then this part will be like below.

```rust
let _: () = msg_send![
      notification_center,
      addObserver: this
      selector: sel!(keyDown:)
      name: input_source_did_changed_notification_name
      object: this
  ];
```

src/platform_impl/macos/view.rs
komi1230: You're right. I move this function to `util/` and mark this function unsafe.

src/platform_impl/macos/view.rs
komi1230: I tested this PR with Alacritty, which @kchibisov has developed for inline input (https://github.com/alacritty/alacritty/pull/5790), and I found these keycodes would make effect not only on IME but on window.
For example, I typed `echo こんにちは` and here `こんにちは` is marked text. And before inserting marked text, I push delele key, then only marked text should be deleted but also inserted text before had been deleted.

https://user-images.githubusercontent.com/27967726/150171774-e25f2e6d-f0fa-4a6a-a811-0b0d4d690ecc.mov

komi1230:
@ArturKovacs @kchibisov I fixed some points, and I want your advice for this. If you feel OK about these changes, please merge.

ArturKovacs:
I hope you don't mind that I added some changes. We apparently don't need `NSTextInputContextKeyboardSelectionDidChangeNotification`, so I removed it.

ArturKovacs:
@kchibisov note that I added some important information to the documentation of `IME::Enabled`

src/event.rs
```diff
@@ -636,6 +636,11 @@ pub struct KeyboardInput {
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub enum IME {
     /// Notifies when the IME was enabled.
+    ///
+    /// `KeyboardInput` and `ReceivedCharacter` events may still be sent
+    /// between the `IME::Enabled` and the `IME::Disabled` events.
+    /// The application may need to ignore some of those events while
+    /// the IME is enabled.
```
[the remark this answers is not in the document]
kchibisov: So given that I'm not sure that comment wrt `Enabled` is relevant. Also on some X11 enabled is just send on start and send back on end, so it's mostly about ignoring events during preedit handling. Don't you think?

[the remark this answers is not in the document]
ArturKovacs: Can you describe what `IME::Enabled` and `IME::Disabled` are used for as precisely as you can?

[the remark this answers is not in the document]
ArturKovacs: > so it's mostly about ignoring events during preedit handling

Yes，I think you are right. But I still don't know what `Enabled` and `Disabled` are used for :(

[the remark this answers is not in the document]
kchibisov: The original idea behind those two events were to indicate that the user has activated/deactivated some sort of input method. The `Enabled` would indicate that user should start performing calls to `window.set_ime_position`, the disabled would indicate that all pending IME handling could be safely aborted, since input method is gone.

Without those events the end user will have to always perform `set_ime_position` to keep IME popup in a meaningful place.

I think you're thinking of those events as a `Composition::Started` and `Composition::Ended`, but that's not the case, and those aren't required since `Preedit` and `Commit`/`Disabled` are some sort of indication that text composing started/ended.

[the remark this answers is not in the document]
ArturKovacs: Now I understand. Thank you!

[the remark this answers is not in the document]
ArturKovacs: But we should document that, because I bet that it won't be obvious to the users either.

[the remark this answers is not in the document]
kchibisov: >Yes，I think you are right. But I still don't know what Enabled and Disabled are used for :(

They are just hints to not do any IME related processing when handling input. The thing is that this API is only about event, however there should be methods on a Window to provide a feedback wrt things you're doing, like setting surrounding text, etc, etc.

And if your IME is disabled you don't need to perform a series of `set_ime_position, set_surrounding_text, set_contnets_type` calls.

But the later calls are not here.

We could go completely without `Enabled` and `Disabled` events, and leave it up to winits' backends to optimize calls to mentioned methods when IME isn't present.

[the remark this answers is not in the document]
kchibisov: >But we should document that, because I bet that it won't be obvious to the users either.

Yes, but I'll give a pass on documentation when we open it against master, since it'll likely be soon. We're waiting on X11 to be fixed and on macOS impl to be reviewed, since if backend doesn't support IME or the API isn't implemented it'll still work but via `ReceivedCharacter` API.

ArturKovacs (approved):

ArturKovacs:
@madsmtm I reviewed the IME part of the PR, and now it seems done now. If it's up to me, you can merge it.

@kchibisov is there anything else that needs to be done with this?

@komi1230 As you can see I pushed some commits to this branch. If you don't mind I'll take over this and address the remaining feedback.

komi1230:
I completely agree with your changes and have found this works good in macOS.

kchibisov (changes requested):
I've tested the changes on my macOS machine.

1. The backspace key is triggered twice. always. even on us layout.
2.  I think we shouldn't send backspace and other keys downstream when we're doing composing. If we really want to, we can mark them as synthetic, so the users will still get keys, but will know that those weren't for them.

komi1230:
This is a topic about responsibility boundary between winit and application. I think all keyboard event should be sent to application, and let application decide how to handle all keyboard event. This is possible via IME enum because application know when IME has been enabled. 
Winit is low-level GUI crate so that this should be used as interface to GUI.

ArturKovacs:
I thought about it some more. Let's say that we don't report keyboard events during IME. What happens if I'm using a Chinese keyboard and I want to play a game, where <kbd>W</kbd> from the US layout means "move forward". Winit has no way of knowing that the application is a game, and wants to interpret the raw key events instead of the IME events.

How does this usually work with games? Do chinese users have to switch to a US layout just to play games? What's the situation with a Japanese keyboard? Do you switch to Romaji just for gaming?


<details>

<summary>Order of events have changed but not much else</summary>

Note that the order of the events are now changed, but other than that this PR sends the same events as before, on US layout

On `master`
```
key: KeyboardInput { scancode: 51, state: Pressed, virtual_keycode: Some(Back), modifiers: (empty) }
ch: '\u{7f}' text: 
key: KeyboardInput { scancode: 51, state: Released, virtual_keycode: Some(Back), modifiers: (empty) }
```

This PR:
```
ch: '\u{7f}' text: 
key: KeyboardInput { scancode: 51, state: Pressed, virtual_keycode: Some(Back), modifiers: (empty) }
key: KeyboardInput { scancode: 51, state: Released, virtual_keycode: Some(Back), modifiers: (empty) }
```

</details>

kchibisov:
>I thought about it some more. Let's say that we don't report keyboard events during IME. What happens if I'm using a Chinese keyboard and I want to play a game, where W from the US layout means "move forward". Winit has no way of knowing that the application is a game, and wants to interpret the raw key events instead of the IME events.

I think they just disable IME, I'm not sure if there's a request to disable it on macOS, but on Wayland you have one. Since on Wayland your user won't be able to play the game anyway if it's relying on KeyboardInput, since IME just eats all of it and you get no events at all.

So to accommodate for such case winit needs explicit methods to disable/enable IME. 

>Note that the order of the events are now changed, but other than that this PR sends the same events as before, on US layout.

@ArturKovacs are you sure that you've used latest changes from that PR? I'll recheck, but I think I see both of them in the logs.

kchibisov:
>@ArturKovacs are you sure that you've used latest changes from that PR? I'll recheck, but I think I see both of them in the logs.

Hm, the change of order seems to break things in alacritty, since it relies on suppressing characters from `ReceivedChar` based on the key you've pressed before. And it kind of makes sense to send characters after you've pressed a key.

ArturKovacs:
We can keep the same order as in `master`. I only wanted to have the `IME::Enabled` be before the KeyboardInput, but I didn't intend to change the order of `ReceivedCharacter`

kchibisov:
>We can keep the same order as in master. I only wanted to have the IME::Enabled be before the KeyboardInput, but I didn't intend to change the order of ReceivedCharacter

Yes, please, let's go back to previous order. Since on all platforms it's that way.

komi1230:
Should `IME::Enabled` and `IME::Disabled` be sent in `keyDown` event ?

kchibisov:
>This is a topic about responsibility boundary between winit and application. I think all keyboard event should be sent to application, and let application decide how to handle all keyboard event. This is possible via IME enum because application know when IME has been enabled.

I do understand, but winit is aslo about providing the same experience on all platforms. That's why it sends modifiers as empty if you unfocus the window even if you have something pressed.

So if someone rely on keyboard events being send during composing they're out of luck on Wayland, since on Wayland if you don't want to handle IME you explicitly call for `disable`(well, you don't have to, if you haven't enabled it) after you've got `Enter` event from compositor.

I'd suggest to add `enable_ime`/`disable_ime` methods on a window to enable and disable IME respectively. So if the users want to write a game they just call for `disable` and get events they need.

komi1230:
@ArturKovacs Ssame opinion. I want you to complete this PR ! Go ahead !

ArturKovacs:
So I've been working on this, but I seem to have hit an issue. It seems impossible to disable IME.

A "widget" (NSView) either supports IME (`NSTextInputClient`) or it doesn't and there's no way to enable or disable it after the widget was created. One workaround would be to have two NSView instances and switch between them during runtime. But this is a horrible hack, because then, every operation on the NSView would have to be done on both NSView instances, and I'm not even sure if this would work at all.

I also ran a couple of games with the "Pinyin - Simplified" input on macOS 10.15. The games I tested either didn't have IME input at all or if they had IME input it was enabled during gamplay so any keyboard input brought up the preedit selection box.

So the current state of things seems to be that "if you want IME in your game, you should have it enabled all the time".

So due to all of this, I think we shouldn't even try to implement enabling/disabling IME on macOS unfortunately.

ArturKovacs:
This should be ready @kchibisov; please give it a try.

komi1230:
Actually I have also think about how to change IME and found it's difficult to do it. I found, however, [NSTextInputContext](https://developer.apple.com/documentation/appkit/nstextinputcontext) has `selectedKeyboardInputSource` and `keyboardInputSources` as method. See below image.

スクリーンショット 2022-01-27 11 52 03

As you know, macOS doesn't have direct API to disable IME but you can realize `disable_ime` artificially by setting other input source in `selectedKeyboardInputSource`. For example, in our implementation, we can implement `disable_ime` with like the following code:

```rust
fn disable_ime() {
    let input_source_list: Vec<String> = keyboard_input_sources();
    let mut current_input_source = selected_keyboard_input_source();
    for input_source in input_source_list {
        current_input_source = input_source;

        // Type some key and check if this uses IME
        key_down(some_key_code);
        if has_marked_text() { 
            discard_marked_text();
            return;
        } else {
            return;
        }
    }
}
```

But this approach seems very ugly. Should we give up implementing `disable_ime`?

kchibisov:
@komi1230 

Nah, that's not how it should be done.

@ArturKovacs It seems like KeyboardInput isn't send anymore? I don't get backspace, etc. Also I've noticed that you can't delete last char in preedit. To repro just type anything and try deleting it, you'll notice that event which sets preedit to empty string doesn't happen, however you can't commit that string by pressing `Enter` until you type something. So it seems like IME in winit gets desynced with ime on macOS.

As far as I can see macOS gets keyboard input no matter whether there's IME running or not? But the only thing that gets lost when IME is present is `received_chars` information? Could we somehow get `received chars` when we have IME preedit being done?

I think the thing we can with enable and disable is to guarantee the following:

1. If you've issued `enable` you'll get `IME` events. When IME composing is done you won't ever get any `KeyboardInput` until it commits or preedit string gets empty.
2. If you've issued `disable` you'll always get `KeyboardInput`, but there's no guarantee that you'll continue to get `ReceivedChars` event.

The thing about those statements is that it's enough for games, since you're relying on virtual key codes and not received chars most of the time. And I kind of need this API for Wayland, since you'll likely need a way to toggle IME input, since IME just eats everything you can't even have keybindings during it composing, however it'll be nice for me to have a way to disable it for e.g. Vi mode in alacritty, since you don't type anything into it.

The thing you should consider is that winit is a mid layer, and it can simulate enable/disable IME if platforms doesn't native supports it. Though, you'd have to Also early return in methods like `set_ime_position` or set it to a value where pop-up window won't occur...

What do you think @ArturKovacs ?

ArturKovacs:
Thanks both of you for the input! I think I found a way to implement enabling/disabling IME correctly. I'm currently working on that.

ArturKovacs:
@kchibisov I added the `set_ime_enabled` function to Window, and I think I fixed the issues you found.

As far as the double backspace is concerned: I think this is not caused by this branch. I checked what happens on 0.26.1 and I get the exact same events when I press backspace using the US layout
```
key: KeyboardInput { scancode: 51, state: Pressed, virtual_keycode: Some(Back), modifiers: (empty) }
ch: '\u{7f}'
key: KeyboardInput { scancode: 51, state: Released, virtual_keycode: Some(Back), modifiers: (empty) }
```

src/platform_impl/macos/view.rs
```diff
@@ -50,38 +50,75 @@ impl Default for CursorState {
...
         ns_window,
+        ime_subview: IdRef::new(nil),
         cursor_state,
         ime_spot: None,
         raw_characters: None,
         modifiers: Default::default(),
         tracking_rect: None,
+        ime_state: ImeState::Disabled,
+        input_source,
+        app_enabled_ime: false, // TODO: add example to show how to use IME
```
kchibisov, line 107: I think you've added it?

kchibisov (changes requested):
The thing I've noticed that I don't get empty preedit events anymore. So if start typing something, and then deleting it winit and system IME gets out of sync, since the last event I'm getting from winit is preedit with one cjk char in it, however backspace one more time I can't commit with enter, since system ime is likely empty already, but winit hasn't send me Preedit with empty string, it wasn't like that before, so I think it's something wrong with your handling of preedit.

src/platform_impl/macos/view.rs
```diff
@@ -504,31 +646,25 @@ extern "C" fn insert_text(this: &Object, _sel: Sel, string: id, _replacement_ran
...
 
-        let mut events = VecDeque::with_capacity(characters.len());
-        for character in string.chars().filter(|c| !is_corporate_character(*c)) {
-            events.push_back(EventWrapper::StaticEvent(Event::WindowEvent {
-                window_id: WindowId(get_window_id(state.ns_window)),
-                event: WindowEvent::ReceivedCharacter(character),
-            }));
+        // Marked text will be clear when inserting marked text.
+        if state.is_ime_enabled() && !is_control {
+            state.ime_state = ImeState::WillCommit { text: string };
```
kchibisov, line 662: Maybe just return and don't nest `for` loop?

src/platform_impl/macos/view.rs
```diff
@@ -50,38 +50,75 @@ impl Default for CursorState {
...
+        input_source,
+        app_enabled_ime: false, // TODO: add example to show how to use IME
+        characters_to_send: Vec::with_capacity(8),
     };
     unsafe {
         // This is free'd in `dealloc`
         let state_ptr = Box::into_raw(Box::new(state)) as *mut c_void;
         let ns_view: id = msg_send![VIEW_CLASS.0, alloc];
+        let ime_subview: id = msg_send![IME_SUBVIEW_CLASS.0, alloc];
+        let ime_subview: id = msg_send![ime_subview, initWithWinit: state_ptr];
```
kchibisov, line 115: Would be nice to have a comment why do we need a subview for that?

src/platform_impl/macos/view.rs
ArturKovacs: I don't know what you are referring to. Please explain

src/window.rs
```diff
@@ -800,6 +800,26 @@ impl Window {
...
+    /// When IME is disabled, the window won't receive [`IME`] events, and will
+    /// receive [`KeyboardInput`] events for every keypress instead. With IME
+    /// disabled, the window will also get [`ReceivedCharacter`] events for
+    /// certain keyboard input. Disabling IME is useful for games for example.
+    ///
+    /// [`IME`]: crate::event::WindowEvent::IME
+    /// [`KeyboardInput`]: crate::event::WindowEvent::KeyboardInput
+    /// [`ReceivedCharacter`]: crate::event::WindowEvent::ReceivedCharacter
+    #[inline]
+    pub fn set_ime_enabled(&self, enabled: bool) {
```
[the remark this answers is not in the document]
ArturKovacs: Hmm yeah the phrase "enabled" might cause a confusion with the IME::Enabled event which is mostly unrelated. Alright, let's go with `use_ime`.

src/platform_impl/macos/view.rs
```diff
@@ -509,6 +576,19 @@ extern "C" fn unmark_text(this: &Object, _sel: Sel) {
         let _: () = msg_send![s, release];
         let input_context: id = msg_send![this, inputContext];
         let _: () = msg_send![input_context, discardMarkedText];
+
+        let state_ptr: *mut c_void = *this.get_ivar("winitState");
+        let state = &mut *(state_ptr as *mut ViewState);
+        AppState::queue_event(EventWrapper::StaticEvent(Event::WindowEvent {
+            window_id: WindowId(get_window_id(state.ns_window)),
+            event: WindowEvent::IME(IME::Preedit(
+                String::new(), // empty string
```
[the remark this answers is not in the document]
ArturKovacs: True :D

src/platform_impl/macos/view.rs
ArturKovacs: Indeed

src/window.rs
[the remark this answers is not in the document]
maroider: Neither name seems to convey the intent to me. May I (tentatively) propose `set_ime_wanted`/`set_wants_ime` and/or `set_allow_ime`/`set_ime_allowed`?

src/platform_impl/macos/view.rs
ArturKovacs: I'll add a comment to `IME_SUBVIEW_CLASS`

src/window.rs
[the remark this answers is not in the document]
ArturKovacs: In my opinion, `set_ime_allowed` is the best so far. It's also consistent with the other IME related function, `set_ime_position`

src/platform_impl/macos/view.rs
kchibisov: I mean, that you can early return, and you won't need else branch. So you won't nest for loop increasing  overall indention.

ArturKovacs:
> The thing I've noticed that I don't get empty preedit events anymore.

I'm on 10.15 and I do. Using the `ime_input` example and the Pinyin input, I press <kbd>N</kbd>, <kbd>I</kbd>, <kbd>Backspace</kbd>, <kbd>Backspace</kbd>, and this is what I get:
```
Enabled
Preedit("n", Some(1), Some(1))
Preedit("ni", Some(2), Some(2))
Preedit("n", Some(1), Some(1))
Preedit("", Some(0), Some(0))
```

However there was a bug that it didn't report keyboard input after deleting the preedit text. Now that should be fixed, so please check if it works.

kchibisov (changes requested):
I'm fine with using `set_ime_allowed`, would you mind adding stabs for other platforms which basically logs that not-implemented, so we at least get everything to build?

I'd send Wayland implementation for them once we've done with macOS.

I'll try to look into X11 impl, since it'll be the only blocker for merging IME inside winit.

Besides that, one thing I'd like to known is how macOS handles and expects users to move cursor around in preedit string? I wasn't able to do so with hiragana IME they provide. Would you mind testing other apps like Terminal.app and see what they are trying to do? Maybe they expect to process key events and set surrounding text in IME string...

ArturKovacs:
> I'll try to look into X11 impl, since it'll be the only blocker for merging IME inside winit.

If you want me to, I can also have a stab at the X11 implementation, sice I've been working on that for the new keyboard branch. If I do it, than I would do it using IBus.

kchibisov:
>If you want me to, I can also have a stab at the X11 implementation

That would be really nice, thx.

> sice I've been working on that for the new keyboard branch. If I do it, than I would do it using IBus.

Do you mean testing with IBus or something else? We have X11 implemented in some way on this branch, however it turns out it's not really working.

ArturKovacs:
> Do you mean testing with IBus or something else?

I mean to remove the existing implementation, and create a new implementation that uses IBus to get IME events. I have no idea how to make the existing implementation work, but I have a decent idea of how to use IBus.

kchibisov:
Won't it make us to depend on ibus for X11 IME? Can't IME be done just with what X11 provides?

ArturKovacs:
Why is that a concern? Is there any system currently that uses X11 and doesn't have IBus?

ArturKovacs:
On a different note: what is stopping this from getting merged?

kchibisov:
>Why is that a concern? Is there any system currently that uses X11 and doesn't have IBus?

I never had ibus installed and some use fctix, etc, not ibus.

kchibisov:
> On a different note: what is stopping this from getting merged?

I should test it, which I'll do in an hour.

ArturKovacs:
> I never had ibus installed and some use fctix, etc, not ibus.

In that case, I think I'll try to get the current implementation to work.

madsmtm (changes requested):
I'll try to allocate time for reviewing it this Friday (I'm not well-versed in IME, so doing a bit of research and testing it on my local machine will take some time)

If I haven't responded by then, feel free to merge

src/window.rs
```diff
@@ -800,6 +800,26 @@ impl Window {
...
+    ///
+    /// When IME is not allowed, the window won't receive [`IME`] events, and will
+    /// receive [`KeyboardInput`] events for every keypress instead. Without
+    /// allowing IME, the window will also get [`ReceivedCharacter`] events for
+    /// certain keyboard input. Not allowing IME is useful for games for example.
+    ///
+    /// [`IME`]: crate::event::WindowEvent::IME
+    /// [`KeyboardInput`]: crate::event::WindowEvent::KeyboardInput
+    /// [`ReceivedCharacter`]: crate::event::WindowEvent::ReceivedCharacter
+    #[inline]
```
kchibisov, line 818: We should likely add a note somewhere that IME is disabled by default.

kchibisov (changes requested):
I've tested it on my macOS machine with alacritty and it fails in the following ways:

1. The state of IME isn't persistent. If I call `set_ime_allowed(true)` and change input source ime will stop working.
2. The ime gets randomly locked up. The event loop is continuing to run, but I have neither Keyboard nor IME events. I've tested it with hiragana and you likely can press arrow keys to make it trigger.
3. When IME is disabled I still have IME window, but it's in the left bottom corner of the screen. I'd expect that it doesn't exist.

I think the locking up is the major pita now.

src/platform_impl/macos/util/mod.rs
```diff
@@ -146,3 +147,17 @@ pub unsafe fn toggle_style_mask(window: id, view: id, mask: NSWindowStyleMask, o
...
+    let characters = if has_attr != NO {
+        // This is a *mut NSAttributedString
+        msg_send![string, string]
+    } else {
+        // This is already a *mut NSString
+        string
+    };
+    let utf8_sequence =
+        std::slice::from_raw_parts(characters.UTF8String() as *const c_uchar, characters.len());
+    std::str::from_utf8_unchecked(utf8_sequence).to_string()
```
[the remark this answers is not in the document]
ArturKovacs: Good point. We don't know how `UTF8String` handles strings that are already ill-formed.

I'd suggest using `String::from_utf8_lossy` because panicking in such cases would be uncalled for. In fact we should also rename the function to `id_to_string_lossy` to be explicit about this, and to allow for creating another version of this function which returns a Result in case the conversion failed.

ArturKovacs:
I could reproduce the first two with alacritty, but I haven't found a reliable way to reproduce them. Although I think did find how to reproduce getting double characters:

The double character appearance can be reproduced by having the US layout active, switch to another desktop (with CTRL + Arrow), and switch back. Pressing keys like A, S, D will produce two characters in alacritty. This can then be reverted by having alacritty in focus, switching to Hiragana, pressing A, then switching back to US.
HOWEVER after this, if one presses Backspace, then every keypress will again produce two characters.

My previous commit should fix this, and hopefully some of the other issues as well.

kchibisov:
@ArturKovacs it seems like I can't get alacritty locked anymore, which is good.

However if I enable ime on creation, but start with ABC layout. Toggling layout to hiragana won't make IME work, I'll just continue to receive input as if there's no IME running at all. There's no way for downstream to enable IME in that case, since it doesn't even known that anything changed. So winit should track end enabled IME handling routines for a particular window based on recent call to `set_ime_allowed`. So when user switches layout from ABC to hiragana or something like that IME will work automatically.

ArturKovacs:
> However if I enable ime on creation, but start with ABC layout. Toggling layout to hiragana won't make IME work, I'll just continue to receive input as if there's no IME running at all.

I tried to reproduce this, but I couldn't. I modified the `ime_example` so that `ime_set_enabled` is called immediately after the window is created (before `run`). I had the US layout active when starting the application, then I switched to Hiragana and I received IME events as I should, and the candidate selection box appeared.　Does this do the same for you?

src/window.rs
ArturKovacs: Should be done now

src/platform_impl/macos/util/mod.rs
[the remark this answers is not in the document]
ArturKovacs: This is now addressed

ArturKovacs:
@madsmtm suggested that we might not need the subview if we instead implmenet the `NSTextInputClient` with the original winit view, and then simply only call `interpretKeyEvents` when IME is allowed. This yields a simpler implementation and the behaviour seems to be identical.

Note that both of the solutions require IME to be allowed for dead-key combinations to be forwarded to the application.

src/platform_impl/macos/view.rs
```diff
@@ -98,6 +125,32 @@ pub unsafe fn set_ime_position(ns_view: id, input_context: id, x: f64, y: f64) {
...
 }
 
+pub unsafe fn set_ime_allowed(ns_view: id, ime_allowed: bool) {
+    let state_ptr: *mut c_void = *(*ns_view).get_mut_ivar("winitState");
+    let state = &mut *(state_ptr as *mut ViewState);
+    if state.app_allowed_ime == ime_allowed {
+        return;
+    }
+    state.app_allowed_ime = ime_allowed;
+    if !ime_allowed {
```
kchibisov, line 135: Could you early return here? Also, I think it's more logical to test `state.app_allowed_ime` after you've assigned to it, but it's a matter of preference.

src/platform_impl/macos/view.rs
```diff
@@ -504,68 +613,46 @@ extern "C" fn insert_text(this: &Object, _sel: Sel, string: id, _replacement_ran
         let state_ptr: *mut c_void = *this.get_ivar("winitState");
         let state = &mut *(state_ptr as *mut ViewState);
 
-        let has_attr: BOOL = msg_send![string, isKindOfClass: class!(NSAttributedString)];
-        let characters = if has_attr != NO {
-            // This is a *mut NSAttributedString
-            msg_send![string, string]
+        let string = id_to_string_lossy(string);
+
+        let is_control = if let Some(ch) = string.chars().next() {
```
kchibisov, line 618: This looks like `map_or_else` to me.

src/platform_impl/macos/view.rs
```diff
@@ -504,68 +613,46 @@ extern "C" fn insert_text(this: &Object, _sel: Sel, string: id, _replacement_ran
...
     }
     trace!("Completed `insertText`");
 }
 
 extern "C" fn do_command_by_selector(this: &Object, _sel: Sel, command: Sel) {
-    trace!("Triggered `doCommandBySelector`");
-    // Basically, we're sent this message whenever a keyboard event that doesn't generate a "human readable" character
-    // happens, i.e. newlines, tabs, and Ctrl+C.
+    trace!("Triggered `doCommandBySelector`, with {}", command.name());
+    // Basically, we're sent this message whenever a keyboard event was sent to
```
kchibisov, line 641:
```suggestion
    // Basically, we sent this message whenever a keyboard event was sent to
```

kchibisov (changes requested):
It seems like everything is working for me now. Just some minor commends and should be good from my side.

src/platform_impl/macos/view.rs
ArturKovacs: The original was correct.

"we are sent this message..." means "this message is sent to us..."

Maybe the wording should be changed so that it's more clear.

src/platform_impl/macos/view.rs
ArturKovacs: I do return early. It's 3 lines above this one, at line 132.

kchibisov: I mean that can do for this line as well afaics, since the rest of the function is inside this if.

src/platform_impl/macos/view.rs
```diff
@@ -644,54 +731,64 @@ extern "C" fn key_down(this: &Object, _sel: Sel, event: id) {
...
-            window_id,
-            event: WindowEvent::KeyboardInput {
-                device_id: DEVICE_ID,
-                input: KeyboardInput {
-                    state: ElementState::Pressed,
-                    scancode,
-                    virtual_keycode,
-                    modifiers: event_mods(event),
+        let preedit_related = was_in_preedit || now_in_preedit;
+        if !preedit_related || state.forward_key_to_app || !state.app_allowed_ime {
```
madsmtm, line 768: I _think_ accessing `state.forward_key_to_app` is pretty bad here without a mutex/atomics:
`state: &mut ViewState`, and since it is not passed down to `interpretKeyEvents`, Rust is free to reorder or even elide the access (since it "knows" no-one could have modified `forward_key_to_app` between where it was set to `true`, and here).

madsmtm: Maybe we can get rid of the `forward_key_to_app` variable altogether, and instead do the work inside `doCommandBySelector:`? Also, I can't seem to find a place where `doCommandBySelector:` is triggered while IME is active, can you give me some examples of this?

examples/ime_input.rs
```diff
@@ -0,0 +1,86 @@
...
+            Event::WindowEvent {
+                event: WindowEvent::Moved(new_pos),
+                ..
+            } => {
+                window_pos = new_pos;
+                if may_show_ime {
+                    println!("window moved {:?}", window_pos);
+                    window.set_ime_position(window_pos);
+                }
+            }
```
madsmtm, line 46: It seems cumbersome that the user has to set the IME position every time the window moves - can we change `set_ime_position` so that this is not required?

madsmtm (changes requested):
So I think I've laid my eyes on most of it, in general it looks really good, nice work all of you!

A bit of terminology: The official term for the popup / overlay that allows you to select the desired characters is called the "candidate window", see [the chinese guide](https://support.apple.com/sv-se/guide/chinese-input-method/cim12992/104/mac/12.0) and [the japanese guide](https://support.apple.com/sv-se/guide/japanese-input-method/jpim10262/6.3/mac/12.0) - as @ArturKovacs pointed out to me, the type of candidate window differs between input methods. I think it would be nice to have this term documented somewhere (along with all the other names that IME has).

Another important thing: Creating a new `NSTextInputContext` in `util::create_input_context` is probably wrong, we should just use `NSView -inputContext` (which means `view::set_ime_position` is likely working on the wrong `NSTextInputContext`).

src/platform_impl/macos/view.rs
```diff
@@ -644,54 +731,64 @@ extern "C" fn key_down(this: &Object, _sel: Sel, event: id) {
...
+
+        let input_source = current_input_source();
+        if state.input_source != input_source && state.is_ime_enabled() {
+            state.ime_state = ImeState::Disabled;
+            state.input_source = input_source;
+            AppState::queue_event(EventWrapper::StaticEvent(Event::WindowEvent {
+                window_id: WindowId(get_window_id(state.ns_window)),
+                event: WindowEvent::IME(IME::Disabled),
+            }));
+        }
```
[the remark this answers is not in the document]
madsmtm: @kchibisov could you elaborate on the requirements on these?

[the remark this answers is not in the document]
kchibisov: The purpose is that after you've got disabled you don't need to handle `set_ime_position` at all and if you had something in `preedit` you could safely discard it.

src/platform_impl/macos/view.rs
```diff
@@ -433,6 +501,34 @@ extern "C" fn set_marked_text(
...
+        let cursor_end = preedit_string.len();
+        state.ime_state = ImeState::Preedit;
+
+        // Send WindowEvent for updating marked text
+        AppState::queue_event(EventWrapper::StaticEvent(Event::WindowEvent {
+            window_id: WindowId(get_window_id(state.ns_window)),
+            event: WindowEvent::IME(IME::Preedit(
+                preedit_string,
+                Some(cursor_start),
+                Some(cursor_end),
```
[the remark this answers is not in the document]
kchibisov: The purpose is that the position of your cursor is defined by your IME. For example on Wayland you can move your cursor around in preedit text, however it's done entirely by IME. So cursor position is to indicate where your cursor is currently is and where its end. Since if start != end you likely have a so called marked text(selection). The position of the cursor in preedit determines how it should look in the end.

I still have a question who is responsible to update it on macOS, since I can move cursor on preedit string in terminal.app, but not with winit, since when I press left/rigth nothing changes, however in terminal app it seems like moving.

Maybe the user should set cursor position itself on macOS, but at this information is essential for Wayland IME and is the only way to obtain current cursor position in preedit string, so you can draw cursor in the right place.

Does it clarify it for you?

src/window.rs
```diff
@@ -800,6 +800,32 @@ impl Window {
...
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **macOS:** IME must be enabled to receive text-input where dead-key sequences are combined.
+    ///
+    /// [`IME`]: crate::event::WindowEvent::IME
+    /// [`KeyboardInput`]: crate::event::WindowEvent::KeyboardInput
+    /// [`ReceivedCharacter`]: crate::event::WindowEvent::ReceivedCharacter
+    #[inline]
+    pub fn set_ime_allowed(&self, allowed: bool) {
```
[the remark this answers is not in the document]
kchibisov: Nice catch, I forgot that we should likely add it on a builder.

src/platform_impl/macos/view.rs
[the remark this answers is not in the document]
madsmtm: Hmm, okay, but a `Preedit("", None, None)` could do the same, right?

src/platform_impl/macos/view.rs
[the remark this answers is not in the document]
madsmtm: Hmm, I think downstream (e.g. `alacritty`/`winit`, not AppKit) is expected to render things differently, including the cursor/selected text. At first I thought `firstRectForCharacterRange:actualRange:` could be used to help tracking the currently focused characters, but then again, its value is cached, so probably not.

I actually think the `selectedRange` parameter in `setMarkedText:selectedRange:replacementRange:` could do the trick though.

src/platform_impl/macos/view.rs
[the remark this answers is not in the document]
kchibisov: It means that you could have IME running, but just preedit is empty, so you still call `set_ime_postion` and potentially will call `set_surrounding_text`(if we add this one).

You could send empty preedit if you want though, but your preedit is empty anyway after the last `commit`, since this is how it usually works.

src/platform_impl/macos/view.rs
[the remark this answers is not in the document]
kchibisov: I mean, the current information this API exposes is already enough for alacritty to draw everything. At least on Wayland it looks like in any other app from my testing.

examples/ime_input.rs
ArturKovacs: I used this function incorrectly. The function sets the IME position relative to the window, so this is not needed anyways. (Althought there's a bug that the candidate selection box appears at the wrong place then deleting the preedit text and typeing again makes it appear at the correct place. I should fix that later)

src/platform_impl/macos/view.rs
ArturKovacs: Very well spotted! I believe the easiest fix for us is using a [compiler_fence](https://doc.rust-lang.org/std/sync/atomic/fn.compiler_fence.html) after the `interpretKeyEvents`

> can you give me some examples of this?

With the Pinyin input active, press N, then press backspace, then press the up arrow. The up arrow will trigger `doCommandBySelector`.

src/platform_impl/macos/view.rs
```diff
@@ -50,31 +50,58 @@ impl Default for CursorState {
...
-    raw_characters: Option<String>,
     pub(super) modifiers: ModifiersState,
     tracking_rect: Option<NSInteger>,
+    ime_state: ImeState,
+    input_source: String,
+
+    /// True iff the application wants IME events.
+    ///
+    /// Can be set using `set_ime_allowed`
+    app_allowed_ime: bool,
```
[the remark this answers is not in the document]
ArturKovacs: Done

src/platform_impl/macos/view.rs
```diff
@@ -257,6 +267,53 @@ lazy_static! {
             sel!(acceptsFirstMouse:),
             accepts_first_mouse as extern "C" fn(&Object, Sel, id) -> BOOL,
         );
+
+        // ------------------------------------------------------------------
+        // NSTextInputClient
```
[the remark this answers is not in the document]
ArturKovacs: Done

src/platform_impl/macos/view.rs
```diff
@@ -414,6 +472,13 @@ extern "C" fn selected_range(_this: &Object, _sel: Sel) -> NSRange {
     util::EMPTY_RANGE
 }
 
+unsafe fn current_input_source() -> String {
+    let ns_text_input_context: id = msg_send![class!(NSTextInputContext), alloc];
+    let input_context: &Object = msg_send![ns_text_input_context, init];
```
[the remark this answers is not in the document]
ArturKovacs: Good catch! Done

examples/ime_input.rs
```diff
@@ -0,0 +1,86 @@
...
+    window::WindowBuilder,
+};
+
+fn main() {
+    SimpleLogger::new()
+        .with_level(LevelFilter::Debug)
+        .init()
+        .unwrap();
+
+    println!("Press F2 to toggle IME. See the documentation of `set_ime_allowed` for more info");
```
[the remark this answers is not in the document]
ArturKovacs: I merged them into an example called `ime`

src/platform_impl/macos/view.rs
```diff
@@ -504,68 +613,46 @@ extern "C" fn insert_text(this: &Object, _sel: Sel, string: id, _replacement_ran
...
-            slice::from_raw_parts(characters.UTF8String() as *const c_uchar, characters.len());
-        let string = str::from_utf8_unchecked(slice);
-
         // We don't need this now, but it's here if that changes.
         //let event: id = msg_send![NSApp(), currentEvent];
 
-        let mut events = VecDeque::with_capacity(characters.len());
-        for character in string.chars().filter(|c| !is_corporate_character(*c)) {
-            events.push_back(EventWrapper::StaticEvent(Event::WindowEvent {
+        // Marked text will be clear when inserting marked text.
```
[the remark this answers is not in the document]
ArturKovacs: I believe, this was left here from an earlier state. I removed it.

src/platform_impl/macos/view.rs
ArturKovacs: I changed the wording

src/platform_impl/macos/view.rs
ArturKovacs: I changed it to `map_or`

src/platform_impl/macos/view.rs
ArturKovacs: I don't think it's an improvement but I also don't think it's worth the time arguing about it, so I changed it.

kchibisov:
what's the status on this @madsmtm @ArturKovacs ?

ArturKovacs:
Sorry I'm super busy these days, I'll probably be able to work on it between the 4th and 6th but not sure.

ArturKovacs:
If you ask me I think this is pretty much mergeable as it is. The conflicts will have to be resolved of course, but the other stuff isn't that important in my opinion. It's more important to have this feature and we can tweak it later.

I don't really have enough time to continue working on this these days. Although I can resolve the merge conflicts if you are willing to merge this afterwards, @kchibisov.

kchibisov:
If @madsmtm is fine with that or could resolve raised issues, I could merge it.

ArturKovacs:
I merged the `composition-event` branch and test on macos and it seems to work correctly. However the merge introduced a few compile errors for the Linux implementation. Could you please fix that and push the fix to this PR @kchibisov?

kchibisov:
@ArturKovacs it seems you've merged incorrectly, could you just discard the changes you've made to `x11/ime/context.rs`? After that it should be fine.

I can reset file later if you can't do that now.

ArturKovacs:
I think it's best if you do it.

kchibisov:
Ok, I've rebased and applied the patch on composition-event branch. Thanks.

PR #2149, opened 2022-01-13
Rename internal structs for consistency

madsmtm:
Seeing `Id` in the macOS backend has confused me _many_ times (especially because `Id` has another meaning in Objective-C), this PR is just a generalization of that problem (of structs not being named what they're exported as).


rukai (approved):
All of these renames make sense to me!

msiglreith (approved):

maroider (approved):

PR #2150, opened 2022-01-13
macOS RAII trace guards

madsmtm:
To make having the `trace!` calls more ergonomic.

### DIFF
--- a/src/platform_impl/macos/util/mod.rs
+++ b/src/platform_impl/macos/util/mod.rs
@@ -11,7 +11,7 @@ use cocoa::{
     foundation::{NSPoint, NSRect, NSString, NSUInteger},
 };
 use core_graphics::display::CGDisplay;
-use objc::runtime::{Class, Object, Sel, BOOL, YES};
+use objc::runtime::{Class, Object};
 
 use crate::dpi::LogicalPosition;
 use crate::platform_impl::platform::ffi;
@@ -79,6 +79,25 @@ impl Clone for IdRef {
     }
 }
 
+pub(crate) struct TraceGuard {
+    called_from_fn: &'static str,
+}
+
+impl TraceGuard {
+    #[inline]
+    pub(crate) fn new(called_from_fn: &'static str) -> Self {
+        trace!("Triggered `{}`", called_from_fn);
+        Self { called_from_fn }
+    }
+}
+
+impl Drop for TraceGuard {
+    #[inline]
+    fn drop(&mut self) {
+        trace!("Completed `{}`", self.called_from_fn);
+    }
+}
+
 // For consistency with other platforms, this will...
 // 1. translate the bottom-left window corner into the top-left window corner
 // 2. translate the coordinate from a bottom-left origin coordinate system to a top-left one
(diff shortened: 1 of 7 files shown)


ArturKovacs:
First of all this has been on the back of my mind, glad you are doing it! Although it seems better to me to use a third party crate for this, like [`scopeguard`](https://crates.io/crates/scopeguard)

maroider:
This is actually something [`tracing`](https://lib.rs/crates/tracing) (that one `log`-alternative made by the Tokio folks) has built-in. We could admittedly "rip off" some of their macros, ie. we'd define something like `trace_span`/`trace_scope` as

```rust
macro_rules! trace_scope {
    ($($arg : tt) *) => { /* construct the scope guard here */ }
}
```

ArturKovacs:
@maroider Could show an example of how would that be used? `scopeguard` would be used like this:
```rust
extern "C" fn did_finish_launching(this: &Object, _: Sel, _: id) {
    trace!("Triggered `applicationDidFinishLaunching`");
    defer!{ trace!("Completed `applicationDidFinishLaunching`"); };
    AppState::launched(this);
}
```

maroider:
Suck a macro would conceptually wrap up these macro invocations with one macro invocation.

That is, you'd get
```rust
trace_scope!("`applicationDidFinishLaunching`");
```
instead of
```rust
trace!("Triggered `applicationDidFinishLaunching`");
defer!{ trace!("Completed `applicationDidFinishLaunching`"); };
```

madsmtm:
I'm generally in favour of macros to make life simpler, but I don't think their usage is justified in this case; it would be making these lines a measly 16 characters shorter:
```rust
// From
let _trace = TraceGuard::new("applicationDidFinishLaunching:");
// To
trace_scope!("applicationDidFinishLaunching:");
```
(Since `trace_scope!` would still need the `TraceGuard` struct or similar internally)

Though perhaps there's something I'm missing here?

maroider:
```rust
trace_scope!("applicationDidFinishLaunching:");
```
is a lot less noisy than
```rust
let _trace = TraceGuard::new("applicationDidFinishLaunching:");
```
and generally looks more like "regular" logging.

maroider:
> An argument for a macro based solution is that the filename and line number in the trace will be a lot more helpful.

The [#[track_caller]](`https://doc.rust-lang.org/reference/attributes/codegen.html#the-track_caller-attribute`) can be used to let a function "see" where it is being invoked, provided the calling function isn't *also* annotated with `#[track_caller]`.

ArturKovacs (approved):
I didn't actually execute any code but I read through it and it looks good! I'm really glad you did this @madsmtm

src/platform_impl/macos/util/mod.rs
```diff
@@ -79,6 +79,35 @@ impl Clone for IdRef {
     }
 }
 
+macro_rules! trace_scope {
+    ($s:literal) => {
+        $crate::platform_impl::platform::util::TraceGuard::new(module_path!(), $s)
+    };
+}
```
maroider, line 86: Are you absolutely certain that this will drop the `TraceGuard` at the end of the scope instead of being dropped immediately when used as `trace_scope!();`? I'm pretty sure the `scopeguard` crate uses a `let` binding for this.

madsmtm: No, you're definitely right, I intended to add that but totally forgot!

ArturKovacs: Good catch. Now I actually tried printing some text after the `trace_scope` and the printed text showed up between the Triggered and the Completed traces, so great!

PR #2152, opened 2022-01-13
Added keyboard layout translation for key input in wasm

Maix0:
I had this problem where when running on WASM, the keyboard input was locked to QWERTY.
The change I made only fix the layout for key that output letters (so A, B, C, ...) and is only made because the new Keyboard API is to be released someday.
Other keys could be added but I don't have the knowledge to automate such things (as I need to open a page, test every key on my keyboard and record the `key` value by hand)


kchibisov:
Sorry that you've patched was missed likely due to wasm maintainers being mia in general in winit. We have a more general solution to what you describe in https://github.com/rust-windowing/winit/pull/2662 which will land about now.

PR #2154, opened 2022-01-18
Add IME event support for Wayland

kchibisov:


kchibisov:
If someone is interested in testing it in client the support is implemented in https://github.com/alacritty/alacritty/pull/5790

rano-oss (approved):

rano-oss (changes requested):
Could you bump wayland-rs (protocols/client) to 29.4? There is a use after free bug that has been fixed there. It might make the text gibberish.
Except for that it looks great!

kchibisov:
>Could you bump wayland-rs (protocols/client) to 29.4? There is a use after free bug that has been fixed there. It might make the text gibberish.
Except for that it looks great!

Usually you don't need that, since downstream can do that. Though we could force the usage of 0.29.4 in winit I guess.(alacritty is already using it for example).

PR #2155, opened 2022-01-19
On X11, Fix for repeated event loop iteration when `ControlFlow` was `Wait`

ArturKovacs:


ArturKovacs:
I'll merge this on the 1st of February if no-one gives feedback.

src/platform_impl/linux/x11/mod.rs
```diff
@@ -262,120 +295,166 @@ impl<T: 'static> EventLoop<T> {
...
-        let mut events = Events::with_capacity(8);
-        let mut cause = StartCause::Init;
-
-        let exit_code = loop {
+        struct IterationResult {
+            deadline: Option<Instant>,
+            timeout: Option<Duration>,
+            wait_start: Instant,
+        }
+        fn single_iteration<T, F>(
```
madsmtm, line 303: Can this be a closure instead? I think that would make the code shorter and clearer.

madsmtm (changes requested):
Quick look, implementation looks fine (especially after reading the explanatory comment in the code)

src/platform_impl/linux/x11/mod.rs
ArturKovacs: As far as I know, that's not possible, because the closure would need to hold a mutable reference to things like `self` and `control_flow` but those are also mutably borrowed outside of `single_iteration`, which would produce two mutable borrows at the same time which is not allowed.

madsmtm: Ah right, and disjoint closure capture won't save us because it uses methods on `self`.

Then perhaps as a method instead (I know the diff will be larger, but I think it's better that way)?

ArturKovacs: I don't think that would be better. The two are approximately equally good in my opinion.

The method would still have to take everything as a parameter, and it would still have to spell out the type constraints (`F`) on the callback parameter. But that would also force the `IterationResult` struct to be elevated to module level visibility, which isn't needed with the current implementation.

For this reason, I think it would be a waste of time to change `single_iteration` to be a method.

madsmtm: Fair enough

ashdnazg:
Testing this locally, it seems that only a single `MainEventsCleared` is received after input events, causing the app to effectively only run if keys are pressed or the mouse is moved.

maroider:
Yeah, `ControlFlow::Poll` seems to behave as if it is `ControlFlow::Wait` with this change.

ArturKovacs:
Thanks for the review, guys! It should now be correct.

maroider:
It just occured to me that I didn't check the behavoiur of `WaitUntil`, which I belive might also be affected by this bug.

ArturKovacs:
😅 oops I didn't check either. I'm afraid it doesn't wait, not even after the fix.

maroider:
`WaitUntil` seems to work correctly for me with current `master`.

PR #2157, opened 2022-01-20
Add touchpad magnify and rotate gestures support for macOS

jsatka:
Add touchpad two-finger pinch and rotate gestures support for macOS.

Adds two new events, `WindowEvent::TouchpadMagnify` with positive delta values indicating magnification (zooming in) and negative delta values indicating shrinking (zooming out), and `WindowEvent::TouchpadRotate` with positive delta values indicating rotation to left (counterclockwise) and negative delta values indicating rotation to right (clockwise).

Checklist


jsatka:
Fixed formatting

madsmtm (commented):
Haven't looked much at the implementation, but really I'm not sure I can justify this feature belonging to `winit`? Is it possible to support these gestures on other platforms?

jsatka:
> Is it possible to support these gestures on other platforms?

Judging by documentations it should:
- Wayland supports both: [libinput > Gestures](https://wayland.freedesktop.org/libinput/doc/latest/gestures.html)
- Windows supports magnify only: [Touch gestures for Windows](https://support.microsoft.com/en-us/windows/touch-gestures-for-windows-a9d28305-4818-a5df-4e2b-e5590f850741)

src/platform_impl/macos/view.rs
```diff
@@ -1046,6 +1054,64 @@ extern "C" fn scroll_wheel(this: &Object, _sel: Sel, event: id) {
...
+
+    unsafe {
+        let state_ptr: *mut c_void = *this.get_ivar("winitState");
+        let state = &mut *(state_ptr as *mut ViewState);
+
+        let delta = event.magnification();
+        let phase = match event.phase() {
+            NSEventPhase::NSEventPhaseBegan => TouchPhase::Started,
+            NSEventPhase::NSEventPhaseChanged => TouchPhase::Moved,
+            NSEventPhase::NSEventPhaseEnded => TouchPhase::Ended,
```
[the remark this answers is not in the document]
jsatka: It does make sense adding `NSEventPhaseCancelled => TouchPhase::Cancelled,` mapping indeed. Will include it.

In the context of these gestures the `NSEventPhaseMayBegin` would mean that user has placed two fingers on trackpad, which could be a beginning for multiple types of two-finger gestures. `NSEventPhaseBegan` will always be reported once AppKit detects beginning of specific type of gesture.

dhardy:
> I'm not sure I can justify this feature belonging to winit?

It sounds like integration with regular desktop input facilities to me, given that the desktop platform is responsible for reporting gestures (not raw touchpad input), so I'd say yes, this belongs in Winit.

Would it make more sense to report `WindowEvent::Gesture(device_id, phase, Gesture::Magnify)` etc, given the possibility of adding `Swipe` and `Hold` in the future?

jsatka:
> Would it make more sense to report `WindowEvent::Gesture(device_id, phase, Gesture::Magnify)` etc, given the possibility of adding `Swipe` and `Hold` in the future?

Gesture types may have different requirements for reporting the delta, such as swipe gesture having either a direction as additional parameter, or delta as x and y components. But workarounds do exist; swipes could be split into vertical and horizontal swipe types and while hold gesture doesn't technically have a delta it could just be reported always as zero.

I think it is mainly a matter of opinion, which way of reporting is better. Personally I'd prefer a flat structure, so that as an application developer I can add handling for a single type of gesture without additional layers of match arms. But that's just my two cents.

examples/touchpad_gestures.rs
```diff
@@ -0,0 +1,44 @@
...
+                        println!("Zoomed in {}", delta);
+                    } else {
+                        println!("Zoomed out {}", delta);
+                    }
+                }
+                WindowEvent::TouchpadRotate { delta, .. } => {
+                    if delta > 0.0 {
+                        println!("Rotated left {}", delta);
+                    } else {
+                        println!("Rotated right {}", delta);
```
madsmtm, line 36: Nit:

```suggestion
                        println!("Rotated counterclockwise {}", delta);
                    } else {
                        println!("Rotated clockwise {}", delta);
```

src/event.rs
```diff
@@ -317,6 +317,30 @@ pub enum WindowEvent<'a> {
...
+        delta: f64,
+        phase: TouchPhase,
+    },
+
+    /// Touchpad rotation event with two-finger rotation gesture.
+    ///
+    /// Positive delta values indicate rotation to left (counterclockwise) and
+    /// negative delta values indicate rotation to right (clockwise).
+    ///
+    /// At the moment, only supported on macOS.
```
madsmtm, line 337: Nit: Use the usual `## Platform-specific` header

src/event.rs
```diff
@@ -317,6 +317,30 @@ pub enum WindowEvent<'a> {
...
+    TouchpadMagnify {
+        device_id: DeviceId,
+        delta: f64,
+        phase: TouchPhase,
+    },
+
+    /// Touchpad rotation event with two-finger rotation gesture.
+    ///
+    /// Positive delta values indicate rotation to left (counterclockwise) and
+    /// negative delta values indicate rotation to right (clockwise).
```
madsmtm, line 335: Nit: "counterclockwise" and "clockwise" are the correct and precise terms for this, just always use those.

madsmtm (changes requested):
Fair, I've changed my mind, I agree that we should have this in `winit`, at least for the time being.

I think there is merit in exploring an API for "zoom" and "rotate" in general (e.g. not something specific to touchpads), but not something I want to block this PR further on, that can always be added later.

I don't have permissions to touch your branch, so would you mind merging the latest `master` and fixing the conflicts?

src/platform_impl/macos/view.rs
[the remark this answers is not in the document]
madsmtm: I was going to argue that I'd want the behaviours here to be similar to when scrolling, but it seems like `NSEventPhaseMayBegin` is never actually used in `magnifyWithEvent:` and `rotateWithEvent:`, so yeah, probably fine as it is

jsatka:
Latest `master` is now merged with all checks fixed.

src/event.rs
```diff
@@ -426,6 +426,32 @@ pub enum WindowEvent<'a> {
         modifiers: ModifiersState,
     },
 
+    /// Touchpad magnification event with two-finger pinch gesture.
+    ///
+    /// Positive delta values indicate magnification (zooming in) and
+    /// negative delta values indicate shrinking (zooming out).
+    ///
+    /// At the moment, only supported on macOS.
```
madsmtm, line 434: Nit: Use `## Platform-specific` section

src/platform_impl/macos/view.rs
```diff
@@ -1196,6 +1204,66 @@ extern "C" fn scroll_wheel(this: &Object, _sel: Sel, event: id) {
...
+            },
+        };
+
+        AppState::queue_event(EventWrapper::StaticEvent(window_event));
+    }
+    trace!("Completed `magnifyWithEvent`");
+}
+
+extern "C" fn rotate_with_event(this: &Object, _sel: Sel, event: id) {
+    trace!("Triggered `rotateWithEvent`");
```
madsmtm, line 1238: Nit: Use `trace_scope!` macro (and remove `trace` at the end of this function)

src/platform_impl/macos/view.rs
```diff
@@ -1196,6 +1204,66 @@ extern "C" fn scroll_wheel(this: &Object, _sel: Sel, event: id) {
     }
 }
 
+extern "C" fn magnify_with_event(this: &Object, _sel: Sel, event: id) {
+    trace!("Triggered `magnifyWithEvent`");
```
madsmtm, line 1208: Nit: Use `trace_scope!` macro (and remove `trace` at the end of this function)

madsmtm (commented):
Thanks! Just a few more fixes if you wouldn't mind?

madsmtm (approved):

madsmtm:
No worries!

emilk:
I'm eagerly awaiting this to be published in the next winit release!

PR #2158, opened 2022-01-21
MacOS: Fix tracking of phase changes for mousewheel (scroll) on trackpad

apessino:
I noticed that TouchPhase::Ended events were occasionally not reported when scrolling with two fingers with a trackpad. Logging the events generated by view.rs/scroll_wheel confirmed that, in certain cases such as touching and then letting go quickly after a Start event had been posted, no corresponding Ended event would be posted.

The problem was that only a NSEventPhase::NSEventPhaseEnded generated a corresponding winit event, but MacOS posts NSEventPhase::NSEventPhaseCancelled as well to signal that the tracking has finished.

Adding NSEventPhase::NSEventPhaseCancelled fixes the problem, now all Started events receive a corresponding Ended event, regardless of how the user taps the trackpad.

### DIFF
--- a/src/platform_impl/macos/view.rs
+++ b/src/platform_impl/macos/view.rs
@@ -1016,8 +1016,12 @@ extern "C" fn scroll_wheel(this: &Object, _sel: Sel, event: id) {
             NSEventPhase::NSEventPhaseMayBegin | NSEventPhase::NSEventPhaseBegan => {
                 TouchPhase::Started
             }
-            NSEventPhase::NSEventPhaseEnded => TouchPhase::Ended,
-            _ => TouchPhase::Moved,
+            NSEventPhase::NSEventPhaseEnded | NSEventPhase::NSEventPhaseCancelled => {
+                TouchPhase::Ended
+            }
+            _ => {
+                TouchPhase::Moved
+            }
         };
 
         let device_event = Event::DeviceEvent {


madsmtm:
Thanks!

Would you mind documenting in `WindowEvent::MouseWheel` that a `TouchPhase::Started` may be received twice? (First from `NSEventPhaseMayBegin`, then again from `NSEventPhaseBegan`).
I'm not sure if you can also receive `TouchPhase::Ended` twice, but if so, that should also be documented.

madsmtm:
Maybe the better solution is actually to not react on `NSEventPhaseMayBegin`? See https://github.com/rust-windowing/winit/pull/2157, where we would have to not do so.

apessino:
Hey! Sorry about the delay - I am being pulled in a million directions atm and it'll be a couple of days before I can dedicate some more time to this.

I did see the double Start - the first is sent when the two fingers are placed on the pad, the second once the fingers are moved. The problem is that if you touch and start to move immediately, then MayBegin is *not* sent (I am sure there are reasons, but it seems really silly), so you can't respond to only one of the two "begin" events, unless you track the current state and just not send the duplicated event in the case in which the user puts two fingers on the pad but does not move immediately (MayBegin is sent) and then starts to move the fingers (Began is sent). Very annoying...

Another thing I noticed is that right now Moved events are sent unconditionally because:

```
let phase = match event.phase() {
...
_ => TouchPhase::Moved,
```

This results in Moved events even when there is no touch at all (for example when scrolling with a mouse wheel). I think it would be better to send a TouchedMoved in response to NSEventPhase::NSEventPhaseChanged (it seems to be sent consistently as you drag on the trackpad with two fingers), and have either Touch::phase be an Option<TouchPhase> (None when no actual touch is being processed) or add None to TouchPhase and have that be the result of matching _. I am not sure if this has further implications...

Finally, where would you like me to document the TouchPhase behavior? In the code or is there another place I should refer to? First time contributing... sorry about my ignorance. :)

Thanks!

apessino:
One more change that, hopefully, should make tracking the touchpad on Macs possible to a satisfactory extent with minimal changes to winit (no changes at all visible to the users, in fact).

My original PR here fixed the missing Ended events, but that's still not enough. MacOS also supports "momentum scrolling." The current version of winit makes it impossible to distinguish between touch events and mouse scroll events once the user has lifted the two fingers from the pad. Mouse scroll is still received, but if an application needs to track mouse scrolling and touchpad scrolling separately (as I needed in the example below) there is no way to do so reliably.

At first, I tried faking it in my backend by tracking touch started/ended (that's how I found the Ended problem in the first place) and then delaying the actual end until the scrolling magnitude would go to ~0. 

Turns out that this can't be done properly, because MacOS is inconsistent in sending the scrolling events... it often sends bursts of them at once for a frame, other times it "misses" frames altogether. There is no guarantee that the momentum phase is enqueued right after the touch phase has ended - it might take a couple of events before it is received. 

This means that a 0-magnitude scrolling vector is not a reliable indicator of the process being done, and waiting for frames or scanning the events queue all seem like really bad ideas.

I tried a bunch of things, but at the end the only reliable way to get momentum events is to read the momentum phase - the only problem was to pass the information on from winit with minimal impact to the existing API.

To be clear, the "proper" way to address this would be to expand winit to provide a superset of the various platforms' definitions of touchscreen, gestures and touchpad events. For now, this is what I came up with - it works, nothing needs to be modified and it should not affect existing code at all, since it only affects the touch state information that could not be used reliably before:

```
        // The "momentum phase," if any, has higher priority than touch phase (the two should
        // be mutually exclusive anyhow, which is why the API is rather incoherent). If no momentum
        // phase is recorded (or rather, the started/ended cases of the momentum phase) then we
        // report the touch phase.
        let phase = match event.momentumPhase() {
            NSEventPhase::NSEventPhaseMayBegin | NSEventPhase::NSEventPhaseBegan => {
                TouchPhase::Started
            }
            NSEventPhase::NSEventPhaseEnded | NSEventPhase::NSEventPhaseCancelled => {
                TouchPhase::Ended
            }
            _ => match event.phase() {
                NSEventPhase::NSEventPhaseMayBegin | NSEventPhase::NSEventPhaseBegan => {
                    TouchPhase::Started
                }
                NSEventPhase::NSEventPhaseEnded | NSEventPhase::NSEventPhaseCancelled => {
                    TouchPhase::Ended
                }
                _ => TouchPhase::Moved,
            },
        };
```

What this does is map standard and momentum Started/Ended pairs of scroll events into separate, discrete Started/Ended winit phases. There *might* be a gap with no scrolling data between them, but in practice this is not an issue for tracking, as far as I can tell. It's a limited subset of the state data that could be available, but it's good enough for most practical purposes. In the backend, I see these as two discrete "started/ended" touch sequences, and as long as the user code supports them as such it should be possible to track the scrolling reliably.

For example: I have been working on a cross platform viewer for a custom renderer and I wanted the camera tracking to behave *exactly* like Blender does on both PC and Mac, with either the mouse/keyboard or the touchpad, and without having to ask that the user makes unnecessary configuration decisions. It should work out of the box exactly like Blender does on all platforms. 

With the changes above I was finally able to get it working right, all interactions, velocities, key combinations, etc. are exactly the same as Blender no matter the input device. Here's a quick video showing how the trackpad is handled correctly, including momentum:

https://user-images.githubusercontent.com/8810462/151682260-d8aa427c-f1d9-4287-b942-3a6345b22616.mp4

Please let me know if there are other issues to discuss - and perhaps for the next major revision of winit we can work on expanding the touch/scrolling interface to pass along all available state data.

apessino:
Hi again! Is there something preventing this PR from being accepted? It's working well for me and it has no side effects I can think of. 

Thanks!

LoganDark:
Ended and Cancelled are separate for a reason! Please add a new type to Winit for Cancelled instead of shoving it into Ended :(.

src/platform_impl/macos/view.rs
```diff
@@ -1148,12 +1148,27 @@ extern "C" fn scroll_wheel(this: &Object, _sel: Sel, event: id) {
...
+        // be mutually exclusive anyhow, which is why the API is rather incoherent). If no momentum
+        // phase is recorded (or rather, the started/ended cases of the momentum phase) then we
+        // report the touch phase.
+        let phase = match event.momentumPhase() {
             NSEventPhase::NSEventPhaseMayBegin | NSEventPhase::NSEventPhaseBegan => {
                 TouchPhase::Started
             }
-            NSEventPhase::NSEventPhaseEnded => TouchPhase::Ended,
-            _ => TouchPhase::Moved,
+            NSEventPhase::NSEventPhaseEnded | NSEventPhase::NSEventPhaseCancelled => {
```
madsmtm, line 1160: We have `TouchPhase::Cancelled` now, so maybe we should use that?

madsmtm: On second thought, I don't think that makes sense if we don't have a `TouchPhase::MayBegin` event - a "cancelled" scroll doesn't communicate anything other than an ended scroll in our case.

madsmtm (commented):
Hey @apessino, really sorry for the long delay!

`winit`'s handling of touch is certainly lacking, and this PR does a good job at making the situation better. The double `Started` events are still unfortunate, but that happens without this PR, so it's a strict improvement.

I've merged it now, but I fucked up the merge quite badly since I wanted to add a changelog entry, and ended up not giving you proper `git` attribution, sorry about that!

PR #2168, opened 2022-01-29
Always set timer when polling to avoid slow waking

### LINKED ISSUE #2167: Delay after first poll when resuming [iOS]
After an app is suspended, it can take several seconds to get the next event after `NewEvents`. If this takes too long, a watchdog will kill the app. Oddly, the length of the delay seems to be dependent on how long the app was suspended.

HackerFoo:
This fixes #2167. I suspect that the timer needs to be reset after suspending, but I can't find anything on it yet.

Note that removing this case causes the match to drop down to the other polling case that does the same thing, and calls `self.waker.start()`. So effectively, this change calls `self.waker.start()` (setting the next firing time to `std::f64::MIN`) every time through the loop, not just when the polling mode changes.

I'm submitting this PR because it works, and was a very frustrating bug for me, so it should help others encountering the same thing. I'd be grateful for an explanation of *why* it works, though.

### DIFF
--- a/src/platform_impl/ios/app_state.rs
+++ b/src/platform_impl/ios/app_state.rs
@@ -410,9 +410,6 @@ impl AppState {
 
         let new = self.control_flow;
         match (old, new) {
-            (ControlFlow::Poll, ControlFlow::Poll) => self.set_state(AppStateImpl::PollFinished {
-                waiting_event_handler,
-            }),
             (ControlFlow::Wait, ControlFlow::Wait) => {
                 let start = Instant::now();
                 self.set_state(AppStateImpl::Waiting {


madsmtm (changes requested):
Sorry for the delay!

I took at look at the [equivalent code for macOS](https://github.com/rust-windowing/winit/blob/bf366cb99d33bc2da9b50cdc91ed61cfe1d22110/src/platform_impl/macos/app_state.rs#L443-L449), the logic for when to call the waker is equivalent before this change, so this is probably an issue very specific to something iOS does.

I don't really have a good setup available to test it myself, but I think this is a case of "we don't really have a maintainer for iOS, it works for you, and clearly you put some thought into this issue", so I'm thinking we should just merge it.

Though a changelog entry, and maybe a quick comment in the code about that the `Poll => Poll` transition is also covered by the case further down would be nice!

madsmtm:
Moved to https://github.com/rust-windowing/winit/pull/2979/

PR #2169, opened 2022-01-30
feat: add `Window::is_visible`

amrbashir:

### A bit of context to whom it may concern: 

I am a maintainer at https://github.com/tauri-apps and a while back we forked `winit` as `tao` but recently we decided to move back to depend on `winit` directly but we need [some features ](https://matrix.to/#/!DGpLzJTRzBDTwZiogk:matrix.org/$PucGFtd5aIroCbugTCW7yIFdp0SNWaLugwb_xZwgOXc?via=libera.chat&via=matrix.org&via=community.rs) implemented at `winit` before we can make the transition. 

This PR is a first of multiple upcoming PRs hopefully but I can only provide macOS and Windows implementations.

### DIFF
--- a/src/platform_impl/ios/window.rs
+++ b/src/platform_impl/ios/window.rs
@@ -60,6 +60,11 @@ impl Inner {
         }
     }
 
+    pub fn is_visible(&self) -> Option<bool> {
+        warn!("`Window::is_visible` is ignored on iOS");
+        None
+    }
+
     pub fn request_redraw(&self) {
         unsafe {
             if self.gl_or_metal_backed {
@@ -197,11 +202,6 @@ impl Inner {
     pub fn is_maximized(&self) -> bool {
         warn!("`Window::is_maximized` is ignored on iOS");
         false
-    }
-
-    pub fn is_visible(&self) -> Option<bool> {
-        warn!("`Window::is_visible` is ignored on iOS");
-        None
     }
 
     pub fn set_fullscreen(&self, monitor: Option<Fullscreen>) {
--- a/src/platform_impl/ios/window.rs
+++ b/src/platform_impl/ios/window.rs
@@ -199,6 +199,11 @@ impl Inner {
         false
     }
 
+    pub fn is_visible(&self) -> bool {
+        warn!("`Window::is_visible` is ignored on iOS");
+        true
+    }
+
     pub fn set_fullscreen(&self, monitor: Option<Fullscreen>) {
         unsafe {
             let uiscreen = match monitor {
--- a/src/platform_impl/ios/window.rs
+++ b/src/platform_impl/ios/window.rs
@@ -199,9 +199,9 @@ impl Inner {
         false
     }
 
-    pub fn is_visible(&self) -> bool {
+    pub fn is_visible(&self) -> Option<bool> {
         warn!("`Window::is_visible` is ignored on iOS");
-        true
+        None
     }
 
     pub fn set_fullscreen(&self, monitor: Option<Fullscreen>) {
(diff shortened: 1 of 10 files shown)


src/platform_impl/linux/wayland/window/mod.rs
```diff
@@ -363,6 +363,11 @@ impl Window {
         self.maximized.load(Ordering::Relaxed)
     }
 
+    #[inline]
+    pub fn is_visible(&self) -> bool {
+        true
```
kchibisov, line 368: I'm stronly agains this API in it's current form. If anything it should return `Option<bool>`, since on Wayland you don't known that. And you also shouldn't rely on it to draw in your code, since it's just bad.

kchibisov (changes requested):

src/platform_impl/linux/wayland/window/mod.rs
amrbashir: I don't mind `Option<bool>` but this was just meant as a placeholder because I can only provide implementation of Windows and macOS.

src/platform_impl/ios/window.rs
```diff
@@ -199,6 +199,11 @@ impl Inner {
         false
     }
 
+    pub fn is_visible(&self) -> bool {
+        warn!("`Window::is_visible` is ignored on iOS");
```
kchibisov, line 203: Yeah, just looking at it by using `Option` you can clear state with `None` that visibility can't be determined.

src/window.rs
```diff
@@ -708,7 +708,7 @@ impl Window {
     /// - **Wayland / X11:** Not implemented.
```
kchibisov, line 708: Could add a note about `None` that it means that visibility can't be determined? Also you should likely add a note that it's not recommended way to drive your rendering backend, etc, etc.

Also, on Wayland it's `Unsupported` and not `Not implemented`, since it can't be implemented.

kchibisov (changes requested):

madsmtm:
How does `Window::is_visible` compare to `Window::set_visible`?

madsmtm:
See also previous PR: https://github.com/rust-windowing/winit/pull/2023

A quick look at the differences, that PR uses [`occlusionState`](https://developer.apple.com/documentation/appkit/nswindow/1419321-occlusionstate?language=objc) on macOS, which is probably better if the use-case of `is_visible` is to determine when you need to render.

We used the name `visible` in that PR, but given precedent in `Window::is_maximized`, this is probably better.

amrbashir:
> How does `Window::is_visible` compare to `Window::set_visible`?

It is the getter of the state set by `Window::set_visible`
> See also previous PR: #2023

sorry, forgot to check if there was previous PRs.

> A quick look at the differences, that PR uses [`occlusionState`](https://developer.apple.com/documentation/appkit/nswindow/1419321-occlusionstate?language=objc) on macOS, which is probably better if the use-case of `is_visible` is to determine when you need to render.
> 
> We used the name `Window::visible` in that PR, but given precedent in `Window::is_maximized`, `Window::is_visible` is probably better.

yeah `Window::is_visible` is a better name for this PR implementation while the other PR should be `Window::is_occluded`

madsmtm (approved):
> The use-case (at least what I used it for) is checking whether I should hide or show the window when a hot key is pressed.

Ah yeah, I think you or someone else mentioned that on Matrix, given that I think this PR is fine-as-is, thanks!

madsmtm:
I'll merge your three PRs (this one, https://github.com/rust-windowing/winit/pull/2171 and https://github.com/rust-windowing/winit/pull/2172) in a few days unless someone raises any objections, thanks again!

kchibisov (approved):

PR #2171, opened 2022-01-30
feat: add `Window::is_resizable`

amrbashir:

see #2169 for context

### DIFF
--- a/src/platform_impl/web/window.rs
+++ b/src/platform_impl/web/window.rs
@@ -151,6 +151,10 @@ impl Window {
         // Intentionally a no-op: users can't resize canvas elements
     }
 
+    pub fn is_resizable(&self) -> bool {
+        true
+    }
+
     #[inline]
     pub fn scale_factor(&self) -> f64 {
         super::backend::scale_factor()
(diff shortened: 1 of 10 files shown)


madsmtm (changes requested):
Nice, thanks!

Can we document this in `set_resizable` (I assume these are directly correlated, as in, the value set in `set_resizable` is reflected in `is_resizable`)?

filnet:
Also move `is_resizable` just after the corresponding `set_` function so it matches other `set_` / `is_` couples.
Ideally do the same for the similar PRs you created (visible and decorated)

src/platform_impl/web/window.rs
```diff
@@ -246,6 +246,10 @@ impl Window {
         false
     }
 
+    pub fn is_resizable(&self) -> bool {
```
filnet, line 249: some are `#[inline]` while others not. Is there a logic ?

amrbashir: Just a typo on my end. I just follow the convention of the platforms I didn't implement

amrbashir:
> Nice, thanks!
> 
> Can we document this in `set_resizable` (I assume these are directly correlated, as in, the value set in `set_resizable` is reflected in `is_resizable`)?

Seems like `set_maximized` and `is_maximized` don't link to each other and are just written one after another, should I follow this convention or should I link each setter with its getter (including `set_maximized` and `is_maximized`) ?

madsmtm (approved):
> written one after another

That's good enough for me, thanks!

kchibisov (approved):

PR #2172, opened 2022-01-30
feat: add `Window::is_decorated`

amrbashir:

see #2169 for context

### DIFF
--- a/src/window.rs
+++ b/src/window.rs
@@ -701,17 +701,6 @@ impl Window {
         self.window.is_maximized()
     }
 
-    /// Gets the window's current decorations state.
-    ///
-    /// ## Platform-specific
-    ///
-    /// - **Wayland / X11:** Not implemented.
-    /// - **iOS / Android / Web:** Unsupported.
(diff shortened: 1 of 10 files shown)


src/window.rs
```diff
@@ -701,6 +701,17 @@ impl Window {
         self.window.is_maximized()
     }
 
+    /// Gets the window's current decorations state.
+    ///
```
madsmtm, line 705: Could you document use-cases, and relation to `set_decorations`?

madsmtm (changes requested):
Nice, thanks!

src/window.rs
amrbashir: Why is documenting a use-case necessary? it is simply a getter of a state, and again I just followed the convention of `is_maximized`.

madsmtm: Moving them close together is enough for me 👍

madsmtm (approved):
Thanks!

kchibisov (approved):

PR #2176, opened 2022-02-01
Fix inner_size and inner_position of minimized window

### LINKED ISSUE #2015: Windows: inner_size() reports 0x0 size when window is minimized
Noticed this weird behavior.

If you minimize the window on Windows and then request window size using `inner_size()`, the returned value is `0x0`. I think the window size should not be zero since you can hover over minimized window taskbar icon and still see the preview of the window (it can continue to render).

Basically if you now want to always render (even when minimized) you have to track and maintain your own window size metrics and not rely on `inner_size()`. Seems like a bug maybe, but not sure how other platforms behave when minimized?

kazzix14:
This PR fixes #2015.
I found that `minimized_window.inner_position()` also returns invalid value which is `Ok(PhysicalPosition { x: -32000, y: -32000 })`, so I fixed it too.

### DIFF
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -1084,9 +1084,11 @@ unsafe fn public_window_callback_inner<T: 'static>(
             if wparam == winuser::SC_RESTORE {
                 let mut w = userdata.window_state.lock();
                 w.set_window_flags_in_place(|f| f.set(WindowFlags::MINIMIZED, false));
+                w.saved_inner_rect = None;
             }
             if wparam == winuser::SC_MINIMIZE {
                 let mut w = userdata.window_state.lock();
+                w.saved_inner_rect = Some(util::get_client_rect(&mut *window).unwrap());
                 w.set_window_flags_in_place(|f| f.set(WindowFlags::MINIMIZED, true));
             }
             // Send `WindowEvent::Minimized` here if we decide to implement one
--- a/src/platform_impl/windows/window.rs
+++ b/src/platform_impl/windows/window.rs
@@ -118,6 +118,10 @@ impl Window {
 
     #[inline]
     pub fn inner_position(&self) -> Result<PhysicalPosition<i32>, NotSupportedError> {
+        if let Some(rect) = self.window_state.lock().saved_inner_rect {
+            return Ok(PhysicalPosition::new(rect.left as i32, rect.top as i32));
+        }
+
         let mut position: POINT = unsafe { mem::zeroed() };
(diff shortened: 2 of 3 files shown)


src/platform_impl/windows/window.rs
```diff
@@ -157,10 +161,11 @@ impl Window {
 
     #[inline]
     pub fn inner_size(&self) -> PhysicalSize<u32> {
-        let mut rect: RECT = unsafe { mem::zeroed() };
-        if unsafe { winuser::GetClientRect(self.window.0, &mut rect) } == 0 {
-            panic!("Unexpected GetClientRect failure: please report this error to https://github.com/rust-windowing/winit")
-        }
+        let rect = match self.window_state.lock().saved_inner_rect {
```
filnet, line 164: could use `Option::unwrap_or_else()` on the `saved_inner_rect`.

src/platform_impl/windows/event_loop.rs
```diff
@@ -1084,9 +1084,11 @@ unsafe fn public_window_callback_inner<T: 'static>(
             if wparam == winuser::SC_RESTORE {
                 let mut w = userdata.window_state.lock();
                 w.set_window_flags_in_place(|f| f.set(WindowFlags::MINIMIZED, false));
+                w.saved_inner_rect = None;
```
filnet, line 1087: move one line up just to be safe ?

src/platform_impl/windows/window.rs
```diff
@@ -157,10 +161,12 @@ impl Window {
...
     #[inline]
     pub fn inner_size(&self) -> PhysicalSize<u32> {
-        let mut rect: RECT = unsafe { mem::zeroed() };
-        if unsafe { winuser::GetClientRect(self.window.0, &mut rect) } == 0 {
-            panic!("Unexpected GetClientRect failure: please report this error to https://github.com/rust-windowing/winit")
-        }
+        let rect = self.window_state.lock()
+            .saved_inner_rect
+            .unwrap_or_else( ||
+                util::get_client_rect(self.window.0).expect("Unexpected GetClientRct failure: please report this error to https://github.com/rust-windowing/winit")
```
parasyte, line 167:
```suggestion
                util::get_client_rect(self.window.0).expect("Unexpected GetClientRect failure: please report this error to https://github.com/rust-windowing/winit")
```

parasyte (commented):
With this patch, I should be able to remove several zero-checks from about 30 different projects. 🤞

src/platform_impl/windows/window.rs
kazzix14: Thank you for noticing me typos! I fixed them.

PR #2177, opened 2022-02-01
feat(Windows): add skip taskbar methods

amrbashir:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
-- On Widnows, added `WindowExtWindows::set_skip_taskbar` and `WindowBuilderExtWindows::with_skip_taskbar`.
+- On Windows, added `WindowExtWindows::set_skip_taskbar` and `WindowBuilderExtWindows::with_skip_taskbar`.
 - **Breaking:** Bump `ndk` version to 0.6, ndk-sys to `v0.3`, `ndk-glue` to `0.6`.
 - Remove no longer needed `WINIT_LINK_COLORSYNC` environment variable.
 - **Breaking:** Rename the `Exit` variant of `ControlFlow` to `ExitWithCode`, which holds a value to control the exit code after running. Add an `Exit` constant which aliases to `ExitWithCode(0)` instead to avoid major breakage. This shouldn't affect most existing programs.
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- On Widnows, added `WindowExtWindows::set_skip_taskbar` and `WindowBuilderExtWindows::with_skip_taskbar`.
```
filnet, line 11: typo : `Widnows`

msiglreith:
Hello! Thanks for the PR, sorry for the long silence on this topic.
Would it be possible to cover this feature with the standard windows styles or via a owner (we could even allow to use `EventLoop` as owner)? Both options should allow to hide the taskbar icon. I'm a bit hesitant with the current proposal as it might have side effects with other parts.

amrbashir:
AFAIK, window styles can't achieve this behavior but I am open to suggestion of what window styles can do that.

Using `EventLoop`'s window as owner could work. I haven't tried it though so there might be undesirable effects. I will report back when I test this.

We have been using this in tao for quite some time now, around 6 or 7 months and there was no side effects observed. Even electron is using the same API.
TBH, I don't understand the hesitance in using `ITaskbarList` since it is specifically designed for this exact behavior and winit is already using its sibling `ITaskbarList2`.

msiglreith (commented):
Fair enough, I was looking if there was a cheaper option to achieve this with the existing API but I'm fine with the TaskList API as well. (`WS_EX_TOOLWINDOW` seems to have some side-effects, the event loop owned window works as well but doesn't allow toggling at runtime). The only smaller issue I found was toggling with borderless will result in always showing the taskbar requiring explicit toggling of borderless to fix it again but that seems rather minor overall.

My only point open would be renaming the functions to `set_taskbar` and `with_taskbar`, dropping the `skip` - fits better to the current API imo.

amrbashir:
> The only smaller issue I found was toggling with borderless will result in always showing the taskbar requiring explicit toggling of borderless to fix it again but that seems rather minor overall.

Never noticed this behavior before in `tao`; it might be because we are returning `0` in `WM_NCCALCSIZE` to achieve borderless.

> My only point open would be renaming the functions to `set_taskbar` and `with_taskbar`, dropping the skip - fits better to the current API imo.

The name was straight out taken from electron's `setSkipTaskbar` and gtk `set_skip_taskbar_hint`. I feel like `with_taskbar` and `set_taskbar` doesn't convey the main usage of the function.

msiglreith (approved):

PR #2178, opened 2022-02-02
Wayland: Report unaccelerated mouse deltas in `DeviceEvent::MouseMotion`

lassipulkkinen:
This behavior is consistent with at least the X11 backend, and is implied by the documentation:

> This represents raw, unfiltered physical motion. Not to be confused with WindowEvent::CursorMoved.

### DIFF
--- a/src/platform_impl/linux/wayland/seat/pointer/handlers.rs
+++ b/src/platform_impl/linux/wayland/seat/pointer/handlers.rs
@@ -297,9 +297,17 @@ pub(super) fn handle_pointer(
 
 #[inline]
 pub(super) fn handle_relative_pointer(event: RelativePointerEvent, winit_state: &mut WinitState) {
-    if let RelativePointerEvent::RelativeMotion { dx, dy, .. } = event {
-        winit_state
-            .event_sink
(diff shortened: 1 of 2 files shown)


src/platform_impl/linux/wayland/seat/pointer/handlers.rs
```diff
@@ -297,9 +297,17 @@ pub(super) fn handle_pointer(
...
+        dy_unaccel,
+        ..
+    } = event
+    {
+        winit_state.event_sink.push_device_event(
+            DeviceEvent::MouseMotion {
+                delta: (dx_unaccel, dy_unaccel),
+            },
+            DeviceId,
+        )
```
kchibisov, line 311: Could you refactor it to use `match`? I think it's getting pretty hard to read with the current formatting.

kchibisov (changes requested):
It seems like you're missing changelog entry for that. Other than that it should be good.

lassipulkkinen:
Wasn't sure if the change was significant enough for a changelog entry, I've added one now.

kchibisov (approved):

PR #2180, opened 2022-02-04
macOS `set_ime_position` fixes

madsmtm:
Found while reviewing at #2147.

Previously, you were required to call `set_ime_position` every time the window is moved. I don't know how the other platforms fare in this regard though?

Additionally, we were creating an extra `NSTextInputContext` and called `invalidateCharacterCoordinates` on that instead of on the NSView's actual input context. I should note, in my testing I couldn't verify whether `invalidateCharacterCoordinates` actually made a difference, but that may very well just be a bug in my macOS version (10.14.6).


kchibisov:
>Previously, you were required to call set_ime_position every time the window is moved. I don't know how the other platforms fare in this regard though?

On Wayland the spot is relative to the window and not something on a screen, and since you don't know when your window is being moved it's up to compositor to handle that.

On X11 I think most apps don't do anything here? So IME spot just stays outside of the window. I think the most logical thing to do is to hide ime popup somehow if we know that the window is being moved.

madsmtm:
> On Wayland the spot is relative to the window and not an absolute position on screen

On macOS, the spot is an absolute position on the screen, but this PR makes it closer to Wayland's behaviour (the spot is stored as relative to the window, and only converted to an absolute position on the screen when the OS asks for it).

PR #2182, opened 2022-02-04
Fix X11 KeyboardInput events sending with ibus input method.

ashdnazg:

Previously, KeyboardInput events were sent when they weren't filtered by
the IM. In cases where the IM first filtered an event and later replayed
it, the replayed event was the one to be sent. This caused issues under
the ibus input method and repeated keys. Ibus replays the repeated key
events one by one, which depending on frame rate, can be long after the
key was released.
The solution is to keep track of the time of arrival of the last event
of each keycode. An event is a repeat if its time is not later than the
previous event for the same keycode.
This solution is in use in glfw as well: https://github.com/glfw/glfw/commit/9a3664b

### DIFF
--- a/src/platform_impl/linux/x11/event_processor.rs
+++ b/src/platform_impl/linux/x11/event_processor.rs
@@ -10,7 +10,7 @@ use super::{
     XExtension,
 };
 
-use util::modifiers::{ModifierKeyState, ModifierKeymap};
+use util::modifiers::{Modifier, ModifierKeyState, ModifierKeymap};
 
 use crate::{
     dpi::{PhysicalPosition, PhysicalSize},
@@ -130,27 +130,6 @@ impl<T: 'static> EventProcessor<T> {
             }
         {
             return;
-        }
-
-        // We can't call a `&mut self` method because of the above borrow,
-        // so we use this macro for repeated modifier state updates.
-        macro_rules! update_modifiers {
-            ( $state:expr , $modifier:expr ) => {{
-                match ($state, $modifier) {
-                    (state, modifier) => {
-                        if let Some(modifiers) =
-                            self.device_mod_state.update_state(&state, modifier)
-                        {
-                            if let Some(window_id) = self.active_window {
-                                callback(Event::WindowEvent {
-                                    window_id: mkwid(window_id),
-                                    event: WindowEvent::ModifiersChanged(modifiers),
-                                });
-                            }
-                        }
-                    }
-                }
-            }};
         }
 
         let event_type = xev.get_type();
@@ -572,9 +551,12 @@ impl<T: 'static> EventProcessor<T> {
                     let keysym = wt.xconn.lookup_keysym(xkev);
                     let virtual_keycode = events::keysym_to_element(keysym as c_uint);
 
-                    update_modifiers!(
+                    Self::update_modifiers(
+                        &mut self.device_mod_state,
+                        self.active_window,
                         ModifiersState::from_x11_mask(xkev.state),
-                        self.mod_keymap.get_modifier(xkev.keycode as ffi::KeyCode)
+                        self.mod_keymap.get_modifier(xkev.keycode as ffi::KeyCode),
+                        &mut callback,
                     );
 
                     let modifiers = self.device_mod_state.modifiers();
@@ -645,7 +627,13 @@ impl<T: 'static> EventProcessor<T> {
                         }
 
                         let modifiers = ModifiersState::from_x11(&xev.mods);
-                        update_modifiers!(modifiers, None);
+                        Self::update_modifiers(
+                            &mut self.device_mod_state,
+                            self.active_window,
+                            modifiers,
+                            None,
+                            &mut callback,
+                        );
 
                         let state = if xev.evtype == ffi::XI_ButtonPress {
                             Pressed
@@ -722,7 +710,13 @@ impl<T: 'static> EventProcessor<T> {
                         let new_cursor_pos = (xev.event_x, xev.event_y);
 
                         let modifiers = ModifiersState::from_x11(&xev.mods);
-                        update_modifiers!(modifiers, None);
+                        Self::update_modifiers(
+                            &mut self.device_mod_state,
+                            self.active_window,
+                            modifiers,
+                            None,
+                            &mut callback,
+                        );
 
                         let cursor_moved = self.with_window(xev.event, |window| {
                             let mut shared_state_lock = window.shared_state.lock();
@@ -1222,6 +1216,25 @@ impl<T: 'static> EventProcessor<T> {
 
         if let Ok((window_id, x, y)) = self.ime_receiver.try_recv() {
             wt.ime.borrow_mut().send_xim_spot(window_id, x, y);
+        }
+    }
+
+    fn update_modifiers<F>(
+        device_mod_state: &mut ModifierKeyState,
+        active_window: Option<ffi::Window>,
+        state: ModifiersState,
+        modifier: Option<Modifier>,
+        callback: &mut F,
+    ) where
+        F: FnMut(Event<'_, T>),
+    {
+        if let Some(modifiers) = device_mod_state.update_state(&state, modifier) {
+            if let Some(window_id) = active_window {
+                callback(Event::WindowEvent {
+                    window_id: mkwid(window_id),
+                    event: WindowEvent::ModifiersChanged(modifiers),
+                });
+            }
         }
     }
 
--- a/src/platform_impl/linux/x11/event_processor.rs
+++ b/src/platform_impl/linux/x11/event_processor.rs
@@ -37,6 +37,7 @@ pub(super) struct EventProcessor<T: 'static> {
     pub(super) first_touch: Option<u64>,
     // Currently focused window belonging to this process
     pub(super) active_window: Option<ffi::Window>,
+    pub(super) key_event_times: [ffi::Time; 256],
 }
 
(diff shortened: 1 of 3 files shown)


ashdnazg:
@ArturKovacs you can see the effects in this bug report in Veloren: https://gitlab.com/veloren/veloren/-/issues/877 in particular this comment: https://gitlab.com/veloren/veloren/-/issues/877#note_476757314
You can easily reproduce this on linux/X11:
1) Set your X11 input method to ibus `export XMODIFIERS=@im=ibus`
2) Turn Repeat Keys on
3) Run the `window` example from this branch: https://github.com/Imberflur/winit/tree/event-test
4) Hold any key down for 2 seconds
5) Release the key
6) See how you keep getting key presses.

maroider:
I think the issue is that we lack people who are both experienced with and interested in maintaining the X11 backend.

ashdnazg:
The perils of open source...

rib:
Reviewing some of the outstanding PRs / issues for x11 I also couldn't help but notice this [mammoth series](https://github.com/rust-windowing/winit/pull/2088) which includes full integration of XInput2 and libxkbcommon (definitely a good direction for the x11 backend) as well as a full port to xcb.

What's notable with respect to this issue though is that the above series effectively removes IME support via the XIM protocol, arguing that IMEs could be better supported either higher up in the stack, or via direct ibus support.

On the one hand it doesn't seem ideal that removing XIM integration has the effect of removing at least crude IME support from the x11 backend, but on the other hand _this_ issue effectively demonstrates how XIM is a poor interface for IME integration, and it looks like removing XIM integration (use of XFilterEvents) entirely would potentially be another solution for this bug?

src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -1238,6 +1193,103 @@ impl<T: 'static> EventProcessor<T> {
...
+        let device_id = mkdid(device);
+        let keycode = xkev.keycode;
+
+        // When a compose sequence or IME pre-edit is finished, it ends in a KeyPress with
+        // a keycode of 0.
+        if keycode != 0 {
+            let signed_time = xkev.time as i64;
+            let time_since_previous_event =
+                signed_time.wrapping_sub(key_event_times[keycode as usize] as i64);
+
```
rib, line 1234: this look suspicious - casting the new time and the time that it's being compared with into i64 values will defeat the purpose of using wrapping_sub to account for the 32bit timestamp wrapping around I think?

src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -1238,6 +1193,103 @@ impl<T: 'static> EventProcessor<T> {
...
+            // An event is a replay of a previous event if it:
+            // 1) Isn't filtered by the IM.
+            // 2) Isn't the first time that an event with this keycode arrives.
+            // 3) Has a timestamp earlier or equal to the latest non-replayed event with this keycode.
```
rib, line 1238: could maybe also be good for future readers to have some comments about why it's necessary to filter out these repeat events. This is kind of a hacky workaround really, so it feels like that should be really clearly sign posted in the code.

src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -1238,6 +1193,103 @@ impl<T: 'static> EventProcessor<T> {
...
+            let replayed_event = !filtered_by_im
+                && time_since_previous_event != signed_time
+                && time_since_previous_event <= 0;
+
+            if !replayed_event {
+                key_event_times[keycode as usize] = xkev.time;
```
rib, line 1244: considering the `wrapping_sub` arithmetic earlier, I guess this should be more careful to preserve the 32bit size of the X event timestamp. I guess `ffi::Time` is the same a `time_t` which I think can be 8 bytes on 64bit platforms. Could be cast back again later but still, just seems like it's important to be clear about the size if using wrapping arithmetic.

rib:
At least in terms of the description of the issue given, and the proposed fix this seems like a reasonable workaround. I mean it's kinda terrible to have this kind of arms race with the XIM filtering, but x11 is kinda long in the tooth, and it's not surprising :)

I suppose a question to ask here though is who's responsible for key repeat synthesis. Does winit document that it does key repeat synthesis, or set the expectation that it's left to downstream users to handle key repeat?

Looking around, it seems like this is still a bit of an unanswered question:
https://github.com/rust-windowing/winit/issues/753
so winit doesn't currently seem to have a clear opinion on how key repeats are exposed to apps, and doesn't promise that key repeats aren't delivered by the OS.

see also https://github.com/rust-windowing/winit/issues/310 and https://github.com/PistonDevelopers/piston/issues/1220

I'm wondering about that because arguably the input module is configured by the user to generate key repeats, and the behaviour seen looks like it's technically inline with what the user had configured (except very laggy), even though that's not what was expected / desired within Veloren.

Looking at the patch itself, it looks ok to me besides some questions around the wrapping arithmetic and maybe extra comments about why the filtering is needed. It would be slightly easier to review the changes if `handle_key_events` were split into a function in a separate patch, since there's a decent amount of code that moves in the patch with inline changes too, but I think I followed it. Essentially it uses a heuristic to recognise key repeats that come from an IM filter (based on their timestamp) and discards them - based on the solution in glfw.

Besides the question around how winit is supposed to handle key repeats generally, I do wonder about the other (simpler) option raised by the patch series I linked above, which is that winit could drop it's use of XFilterEvents entirely and consider that mechanism to be out-dated and inadequate?

ArturKovacs:
Is this still relevant, given that #2243 has been merged? @kchibisov might be able to help in that case.

rib:
@ArturKovacs it looks like that work was essentially to integrate more deeply with XIM so that pre-edit events from the XIM server could be translated into platform-independent Winit events.

ArturKovacs:
On my local repo, I merged master into this, but after the merge there was a `WindowEvent::KeyboardInput` sent for the first IME input. I'm guessing that `ImeEvent::Start` is received too late from `ime_event_receiver`, but I don't know why that would happen or what's the right way to fix it.

(I tested on Ubuntu 20.04 LTS with GNOME 3.36.8)

@kchibisov please help, this defeated me.

kchibisov:
@ArturKovacs I'll try to take a look once I have time for that particular issue.

geirsagberg:
I'm facing a similar issue on macOS in Bevy when key repeat rate is set faster than FPS (15ms key repeat, 16.7ms fps), see https://github.com/bevyengine/bevy/issues/5909

madsmtm:
> I'm facing a similar issue on macOS in Bevy when key repeat rate is set faster than FPS (15ms key repeat, 16.7ms fps), see [bevyengine/bevy#5909](https://github.com/bevyengine/bevy/issues/5909)

Please open a new issue for that, it'll get lost in this one

src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -37,6 +37,7 @@ pub(super) struct EventProcessor<T: 'static> {
     pub(super) first_touch: Option<u64>,
     // Currently focused window belonging to this process
     pub(super) active_window: Option<ffi::Window>,
+    pub(super) key_event_times: [ffi::Time; 256],
```
zesterer, line 40: Perhaps `256 - KEYCODE_OFFSET` given that we know the last 8 entries will never be used?

src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -1225,6 +1174,122 @@ impl<T: 'static> EventProcessor<T> {
...
+            };
+
+            for chr in written.chars() {
+                let event = Event::WindowEvent {
+                    window_id,
+                    event: WindowEvent::ReceivedCharacter(chr),
+                };
+                callback(event);
+            }
+        }
```
zesterer, line 1290: This early return feels a little easy to break in a future refactor. Perhaps something a bit less brittle like the following?

```rs
if state == Pressed && !filtered_by_im {
    if let Some(ic) = wt.ime.borrow().get_context(window) {
        let written = wt.xconn.lookup_utf8(ic, xkev);
        
        for chr in written.chars() {
            let event = Event::WindowEvent {
                window_id,
                event: WindowEvent::ReceivedCharacter(chr),
            };
            callback(event);
        }
    }
}
```

zesterer (commented):
Thanks for opening this PR, I'm really excited to see progress made on it: we're had quite a number of users (including myself) bitten by this problem, and it seems like recent changes to the GNOME stack make it much more likely to happen. For these users, this isn't just a minor inconvenience: it makes playing Veloren mutually exclusive with the use of repeat keys, and most afflicted users never even discover that disabling repeat keys 'solves' the problem.

ashdnazg:
@zesterer thanks!
I'll have time to address your comments in a few days.

kchibisov:
Be aware that keyboard patches won't be applied due to https://github.com/rust-windowing/winit/pull/2662 , so you should test on that branch your issues instead.

zesterer:
> Be aware that keyboard patches won't be applied due to #2662 , so you should test on that branch your issues instead.

Do you know whether that branch is likely to fix this problem incidentally, or will this fix likely need porting over?

kchibisov:
>Do you know whether that branch is likely to fix this problem incidentally, or will this fix likely need porting over?

That branch has entirely different input model, so it's hard to say for me.

kchibisov:
Could this issue be rechecked with the latest master, given that all the keyboard input was reworked on X11 to use xkb?

PR #1945, opened 2021-05-20
Add support for `event::Touch` on web targets

dsluijk:
- [ ] Tested on all platforms changed
- [x] Compilation warnings were addressed
- [x] `cargo fmt` has been run on this branch
- [x] `cargo doc` builds successfully
- [x] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users
- [x] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior
- [x] Created or updated an example program if it would help users understand this functionality
- [x] Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented

This pull request adds touch support to web targets by the Pointer events API.

This PR also adds a new datatype to the Touch struct, `PointerType`. This indicates which pointer is being used to generate the touch event. As this has been added to the platform-independent Touch struct, all platforms supporting touch had to be changed slightly. I manually checked all implementations and added a None value to all (non-web) Touch structs, but it is possible I missed one.

Closes #1673 .

Issue #99, opened 2016-12-11
Pen/Tablet Input Support


ghost:
I propose we support pen tablet input, such as on devices like the Surface Pro/Studio, Galaxy Note, and eventually the Apple Pencil. These devices support things like pressure sensitivity and rotation/gyro. I don't think this is necessarily relevant to games (although it definitely can be), but it's definitely relevant to the larger digital 3D community.

tangmi:
There is an issue open on imgui (https://github.com/ocornut/imgui/issues/2372) that has a little documentation and discussion on how to handle pen input across platforms that might be relevant when work on this starts.

DorianRudolph:
I made a preliminary implementation for pen support on X11 here https://github.com/DorianRudolph/winit/tree/stylus

Pen pressure can already be accessed on X11 via the raw axis events, but there you get no information about the range of values.

lehmanju:
@DorianRudolph What is the state of your implementation? Any help needed? I could look into stylus support for Wayland but first we should probably make sure the API exposed in Winit is correct.

kchibisov:
@lehmanju if you're interested in contributing Wayland bits, you may add helper it to https://github.com/smithay/client-toolkit , since it's something that will be required anyway.

kirawi:
A bit confused, but how would you consider it to be not implemented? It looks like it effectively is through [Touch](https://docs.rs/winit/0.23.0/winit/event/struct.Touch.html).

tangmi:
Pen input devices can have additional values, such as pen pressure, tilt, and azimuth that are not currently supported in winit. Additionally (and anecdotally), applications sometimes want to be able to distinguish between touch and pen input, even if most interactions end up being the same.

kirawi:
`Touch` has the `force` field which seems to define pressure sensitivity and tilt. I could see the case for the latter, but wouldn't you still consider that stylus input is mostly supported?

tangmi:
That's interesting... Looks like it's only used for the Apple Pencil: https://github.com/rust-windowing/winit/blob/b9f3d333e41464457f6e42640793bf88b9563727/src/platform_impl/ios/view.rs#L233

I think we'd still need to hook up azimuth to do that cool tilt shading effect Apple showed off when they released the Pencil...

Although even more broadly than that, I still think it's probably worth it to separate it out into a more generalized `Pen` event and implement it for all the platforms.

kirawi:
It looks like Windows does have pressure sensitivity support:
[...]

PR #2188, opened 2022-02-08
Web touch event

### LINKED ISSUE #1673: Implement touch handling on web target(s)
Adding touch event handling shouldn't be too hard, but it is tightly coupled with mouse handling due to the web using a unified Pointer Event model. Here are my thoughts:

- Grabbing touch input from `PointerEvent` is simple enough. However, I would be concerned about correctly mapping the [`pointerType` ](https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/pointerType) :
    - `"mouse"` - obviously should be mapped to `CursorMove`, `CursorEntered`, `CursorLeft` and `MouseInput`. In addition, I believe that only pointers with `isPrimary === true` should be reported.
    - `"touch"` - obviously should be mapped to `Touch`.
    - `"pen"` - this is tricky, as `winit` does not have an event for pen/stylus. I suggest that for now we just map pen input the same way as mouse input until `winit` finally supports pen events properly.
    - Empty strings or unknown values - the same as mouse.
- Supporting browsers without `PointerEvent` - I suggest we do **not** add support for `TouchEvent`, because:
    - `PointerEvent` is getting [widespread support](https://caniuse.com/#feat=pointer).
    - `Adding TouchEvent` handling will likely make the code quite complicated.
    - ~~If the user really care about supporting older browsers, they can use a [JavaScript polyfill](https://github.com/jquery/PEP).~~ Turns out the polyfill does not provide `offsetX`/`offsetY` which `winit` is currently using...
- Should touch input also synthesize mouse input?
    - My idea would be that, since some users may not want to go through the trouble of implementing explicit touch handling there should be an option to allow user code to disable touch support and automatically treat the primary touch input as mouse input. Otherwise, touch input will only generate touch events and not mouse events.
    - However, it seems that this is not an option for other platforms. Perhaps it makes more sense to just provide one behaviour, which is for touch input to only generate touch events and not mouse events.

ryo33:

Closes #1673 

This PR is inherited from inactive PR #1945.

The differences are as follows:

- Correctly calculates the physical position by treating with scale factor ~~and flipping the Y-axis~~.
- Support touch pressure.
- add `touch-action: none` to a canvas to prevent the browser from canceling pointer events.
- Working [demo](https://ryo33.github.io/winit-web-touch-event-test/) (pressure will be reflected in the cursor size).

Source code of the demo is [here](https://github.com/ryo33/winit-web-touch-event-test)

cursor


src/platform_impl/web/web_sys/canvas/pointer_handler.rs
```diff
@@ -105,11 +115,129 @@ impl PointerHandler {
...
+                }
+
+                handler(
+                    event.pointer_id(),
+                    touch_physical_position(&event, &canvas),
+                    Force::Calibrated {
+                        force: event.pressure() as f64,
+                        max_possible_force: 1.0,
+                        altitude_angle: None, // I don't have any idea how to calculate this from tilt_x and tilt_y.
+                    },
```
Liamolucko, line 137: This should probably be `Force::Normalized`, since the pressure ranges from 0 to 1 and there's no 'average' pressure.

ryo33: Thanks, I resolved this.

src/platform_impl/web/event_loop/window_target.rs
```diff
@@ -231,6 +231,62 @@ impl<T> WindowTarget<T> {
...
 
+        let runner = self.runner.clone();
+        canvas.on_touch_move(move |device_id, location, force| {
+            runner.send_event(Event::WindowEvent {
+                window_id: WindowId(id),
+                event: WindowEvent::Touch(Touch {
+                    id: device_id as u64,
+                    device_id: DeviceId(device::Id(device_id)),
+                    phase: TouchPhase::Moved,
+                    force: Some(force), // Todo
```
Liamolucko, line 242: What's this TODO referring to?

src/platform_impl/web/web_sys/canvas/pointer_handler.rs
```diff
@@ -105,11 +115,113 @@ impl PointerHandler {
...
+                if event.pointer_type() != "touch" {
+                    return;
+                }
+
+                handler(
+                    event.pointer_id(),
+                    touch_physical_position(&event, &canvas),
+                    Force::Normalized(event.pressure() as f64),
+                );
+            },
```
Liamolucko, line 198: This closure is duplicated between all the touch events, so it might be a good idea to extract it out into another method.

src/platform_impl/web/web_sys/canvas/pointer_handler.rs
```diff
@@ -105,11 +115,113 @@ impl PointerHandler {
...
+        self.on_touch_up = None;
+        self.on_touch_cancel = None;
+    }
+}
+
+fn touch_physical_position(
+    event: &PointerEvent,
+    canvas: &HtmlCanvasElement,
+) -> PhysicalPosition<f64> {
+    // not window scale factor here because we can modify the scale factor.
```
Liamolucko, line 219: I'm not quite sure what you mean by this.

src/platform_impl/web/event_loop/window_target.rs
ryo33: Thanks. I think this is misplaced but I  should consider the meaning of it once.

ryo33: I removed this

src/platform_impl/web/web_sys/canvas/pointer_handler.rs
ryo33: I resolved this

ryo33:
I resolved the conflicts.

ryo33:
Also I update the demo with latest code

src/platform_impl/web/web_sys/canvas/pointer_handler.rs
Liamolucko: I'd probably phrase it `// Calculate the scale factor every time, since it's possible for the canvas's width and height attributes to get out of sync with its actual HTML size.` (I think that's what you meant?)

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- On Web, add support touch events.
```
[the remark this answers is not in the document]
Liamolucko: I think you accidentally reverted this while resolving the merge conflicts.

src/platform_impl/web/web_sys/canvas/pointer_handler.rs
ryo33: @Liamolucko Yes, yours is fine.

Liamolucko (approved):
LGTM

ryo33:
@ryanisaacg Could you review this and suggest another reviewer for this?

ryanisaacg:
I haven't been involved in any winit development for a few years at least; I'm not a good reviewer any longer. (Besides, it does look like you have an approving review?)

ryo33:
I referred https://github.com/rust-windowing/winit/wiki/Testers-and-Contributors
This list seems to be not maintained.
I need one more review as written in CONTRIBUTING.md. Can someone review and merge it for me?

src/platform_impl/web/web_sys/canvas/pointer_handler.rs
```diff
@@ -105,11 +115,79 @@ impl PointerHandler {
...
+) -> PhysicalPosition<f64> {
+    // Calculate the scale factor every time, since it's possible for the canvas's width and height attributes to get out of sync with its actual HTML size.
+    let dpi_width = canvas.width() as f64 / canvas.offset_width() as f64;
+    let dpi_height = canvas.height() as f64 / canvas.offset_height() as f64;
+    PhysicalPosition {
+        x: event.client_x() as f64 * dpi_width,
+        // Flip the Y axis because canvas's origin is top-left.
+        y: canvas.height() as f64 - event.client_y() as f64 * dpi_height,
     }
 }
```
maroider, line 193: We should probably handle the canvas being resized, but I don't believe we do that anywhere right now (see #1661). However, I'm not sure if this is the correct approach, and I'd prefer such a change to be wholistic rather than touching only one part of the API. Is touch input somehow completely borked if you don't do this?

maroider (commented):
I have to note that I'm not particularly familiar with the web backend, but I'll give this my best shot.

---

The touch events appear to send positions that are relative to the viewport, with a negated y-axis, and bounded by the size of the canvas. This is in contrast to cursor events, which report positions which are relative to the upper-left corner of the canvas.

src/platform_impl/web/web_sys/canvas/pointer_handler.rs
ryo33: There are cases where resolution changes and resizes are not synchronized, and I had that problem when I developed a game https://github.com/Hihaheho/Hug using this changeset. However, this is a rare case and may not need to be resolved in this PR.

ryo33: I'm starting to remember. A case I struggled with and resolved is scale factor change. In the Hug game, I should change the scale factor for iOS devices for performance reasons, but the change is not reflected in `scale_factor`.

ryo33: https://github.com/rust-windowing/winit/blob/master/src/platform_impl/web/web_sys/mod.rs#L79-L82
This is the source of the problem. `scale_factor` of the web platform always returns `device_pixel_ratio`.

johanhelsing:
I tested this out again on my game (using winit through bevy_winit). I get the expected touch events, however I now also get mouse events at the same time, which does not match the behavior on linux at least. Also, looking at the docs for `WindowEvent::MouseInput` I couldn't fint a way to tell whether a mouse button event originated from touch or mouse.

Maybe the best solution would be to add checks to the pointer event handlers and ignore events if `PointerEvent::pointer_type == "touch"`? At least this seemed to do the trick for my game.

like this https://github.com/johanhelsing/winit/commit/736168e29c92bd2ba507cf9a084c3156a0416c74 ?

Thanks for picking this up, btw. Looking forward to using it!

ryo33:
Thank you for testing this and giving me a feedback. I’ll add the checks to the pointer event handlers like your commit when I have a time.

ryo33:
I merged @johanhelsing's branch. thanks!

src/platform_impl/web/web_sys/canvas/pointer_handler.rs
johanhelsing: > I'm starting to remember. A case I struggled with and resolved is scale factor change. In the Hug game, I should change the scale factor for iOS devices for performance reasons, but the change is not reflected in `scale_factor`.

Not sure if I follow... How would you "change the scale factor?". Is there some web API or winit api to override this? Is it if you transform the canvas html element somehow (with css transforms and such?)

@maroider @ryo33  In the interest of getting this merged. How should we move forward? Some options I could think of:

1. Rename `touch_physical_position` to something non-touch specific and also call it form the the pointer events as well?
2. Move the workaround to `web::Window::scale_factor` and check the canvas `offset_width`/height there. Should probably also get rid of `web_sys::scale_factor` or perhaps rename it to `device_pixel_ratio` since it's apparantly misleading?
3. Remove the workaround from this PR, and deal with the issue separately?

maroider: I'd prefer to deal with the issue separately.

Zageron:
Any intent to merge and close this out in the near future? :)

johanhelsing:
> Any intent to merge and close this out in the near future? :)

As far as I understood, what remains to be done is just remove the workaround for the scale factor.

And also fix the conflicts, I guess.

ryo33:
I'll work within this week.

oscrim:
Hello, is this PR still worked on? I could very much use this in my project :D

ryo33:
I am taking some time off for mental reasons. It's a simple fix and I'd like to merge it as soon as someone sends a pull request to my branch.

ryo33:
Thanks! @oscrim 
I’ve merged it.

src/platform_impl/web/web_sys/event.rs
```diff
@@ -246,3 +246,11 @@ pub fn codepoint(event: &KeyboardEvent) -> char {
     // https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key
     event.key().chars().next().unwrap()
 }
+
+pub fn touch_position(event: &PointerEvent) -> LogicalPosition<f64> {
+    LogicalPosition {
+        x: event.client_x() as f64,
+        // Flip the Y axis because canvas's origin is top-left.
```
johanhelsing, line 253: This comment should probably be removed? There is no flipping going on here.

```suggestion
```

johanhelsing (changes requested):
Thanks again for picking this up!

src/platform_impl/web/web_sys/event.rs
ryo33: Thanks, I've restored flipping.

ryo33:
I've restored flipping and updated the [demo](https://ryo33.github.io/winit-web-touch-event-test/).

ryo33:
I've resolved conflicts and failed checks

src/platform_impl/web/web_sys/canvas/pointer_handler.rs
```diff
@@ -99,6 +136,13 @@ impl PointerHandler {
...
+                // touch events are handled separately
+                // handling them here would produce duplicate mouse events, inconsistent with
+                // other platforms.
+                if event.pointer_type() == "touch" {
+                    return;
+                }
+
                 if prevent_default {
                     event.prevent_default();
                 }
```
madsmtm, line 148: This should probably also happen in `on_touch_move`?

madsmtm (changes requested):
Thanks, looks good in general. My iPad is sadly too old to properly test this, so someone else will have to do that.

Implementation-wise, I think I would prefer to only have one callback per event (e.g. only one place where we register something on `"pointermove"`), I find that makes it easier to understand what's going on; but not a blocker for me.

Also, a few things regarding documentation:
- Could I get you to update the documentation for the `Touch` structure / the `WindowEvent::Touch` to indicate which platforms this is supported on?
- Some more documentation on when `CursorMoved` and `MouseMotion` are sent would be nice!
- This is probably a breaking change (normal cursor events are no longer sent)

ryo33:
- [x] have one callback per event
- [x] correct use of `prevent_default`
- [x] change the changelog expression with "breaking change"
- [x] update the documentation for the Touch structure / the WindowEvent::Touch to indicate which platforms this is supported on?
- [x] update the [DEMO](https://ryo33.github.io/winit-web-touch-event-test/)

gilescope:
Tried it. Touch events are coming through for me on wasm in firefox simulator and on firefox on iphone. I don't seem to get TouchpadMagnify gestures coming through - that would be nice to have but not essential (as you can probably compute it from the touch events).

gilescope:
" - Only available on **macOS**." Ok. so that explains that one. If someone writes a nice gesture detection lib then let me know as I really can't see why we can't build something generic above the touch events that works on all OSes.

gilescope:
One weird thing, the y axis seems inverted - needs to be height - y. Top left should be 0,0. egui touch handling and my brain expects this, but it seems that bottom left is currently 0,0.
I'm assuming that oddness is coming from this PR.

src/platform_impl/web/web_sys/event.rs
gilescope: Maybe we should make flipping optional? I don't seem to need it flipped (or have a setting to skip flipping for now until later when we can correctly identify all the exact conditions when we should flip).

ryo33:
Oh, I may misunderstand the treatment of the y-axis in winit.

https://docs.rs/winit/latest/winit/event/enum.WindowEvent.html#variant.CursorMoved
> (x,y) coords in pixels relative to the top-left corner of the window. Because the range of this data is limited by the display area and it may have been transformed by the OS to implement effects such as cursor acceleration, it should not be used to implement non-cursor-like interactions such as 3D camera control.

So we don't need y-axis flipping?

gilescope:
Personally I don’t think so. Have you got a second flip that offsets it in
the web demo? Just wondering why that one works.

mockersf:
> So we don't need y-axis flipping?

I'm trying this PR in Bevy, and I had to flip the Y axis in Bevy for it to work. So I would say that it isn't needed.

donmai-me:
I've tried the demo on my iPad and it works great! However, there are two problems I've encountered. First is that touch events using Apple Pencil don't register. Second is that canceled touch events, such as when switching windows when using four-finger swipes, aren't removed.

The last one may be because the demo's source code doesn't handle canceled touch events.

ryo33:
Apple Pencil emits pointer events with `pointer_type = "pen"`.
Does anyone know how winit handles Apple Pencil on iOS target? If it handles it as touches, we need to handle the "pen".

ryo33:
I'm finally confident that we don't need the y-flip, and it's ready to merge.
I don't think we should support pen events for now because other platforms do not support it yet, as #99.

## Demo improvement (now it's really simple, comprehensive, and easy to understand)

See [demo](https://ryo33.github.io/winit-web-touch-event-test/) and its [source](https://github.com/ryo33/winit-web-touch-event-test/blob/main/src/main.rs).

Changes are:
- Use UI instead of Sprite because bevy's UI and winit share the same coordinate system (so no complex position transformation).
- Handle the touch cancel event (thanks to @donmai-me for reporting).

gilescope:
I note the same on android but I too suspect it is not due to this pr. neet
to run an emulator or steal someone’s phone to check out what’s happening.

peterhirn:
I was curious and fixed the android issue: [demo](https://peterhirn.github.io/winit-web-touch-event-test/), [repo](https://github.com/peterhirn/winit-web-touch-event-test). Everything seems to work as expected. Nice work.

gilescope:
(My android issue was unrelated - was using a texture > 4096 high.)

madsmtm (approved):
I'm sorry, I haven't really been following this issue, and am a bit confused - @ryo33, do you feel this is ready now?

I'd be fine with merging it as-is (once CI passes), it has gone for long enough and while it may not be perfect it's clearly an improvement :)

ryo33:
@madsmtm Yes, I think it’s ready to merge.

madsmtm:
Cool, thanks for your contribution, and sorry again it took so long, it's difficult that we don't have a maintainer for web!

ryo33:
I haven’t figured winit out yet, but I hope I would be a maintainer for web in the future maybe. I’m creating a huge product that depends on winit, and it should work well on web. I may work on keyboard issues when my hands are free.

Issue #1605, opened 2020-06-13
NSGraphicsContext::currentContext not accessible from redraw handler on macOS


jrmuizel:
Calling `msg_send![class![NSGraphicsContext], currentContext]` returns null when called from the winit event loop. I suspect this is because the currentContext is only valid while drawRect is being called.

Not being able to get the currentContext makes winit unusable for drawing on macOS unless  OpenGL or Metal is used.

john01dav:
I'm working on a 2D framebuffer crate as well and it seems that this isn't an issue. The Mac OS code in my crate gets the graphics context in the same way, but there's another [issue](https://github.com/servo/core-foundation-rs/issues/497). Perhaps you can help with this? I'd very much like there to be a working GPU-less 2D framebuffer crate for winit. To address the issue that you had the `RedrawRequested` event handler is called from within `draw_rect`, so as long as you do  your drawing there it should be fine. I found this out by having a debugger check the stack trace when handling this event.

john01dav:
Okay, so it turns out that RedrawRequested in winit isn't always called from draw_rect. Specifically, it isn't when it happens due to a requested repaint. If instead of requesting the repaint from winit, I send `setNeedsDisplay:YES` to the view, then winit calls the RedrawRequested event from draw_rect, which I _thought_ would resolve teh sisue that I'm having with by GPU-less crate, but it did not. Nonetheless, your original issue appears to be solved this way (and a PR to winit to make request_repaint() trigger `setNeedsDisplay: YES` when on AppKit would seem to be in order), maybe you know how to solve mine?

jrmuizel:
Can you push the updated version of `softbuffer` that does this?

john01dav:
> Can you push the updated version of `softbuffer` that does this?

I made the setNeedsDisplay change not in softbuffer but in one of the examples (the animation one, the only one that is broken on mac os as the others just display a static image). I did so in a quick and dirty testing way (I just added the setNeedsDisplay call directly in the example, so it won't compile on anything except Mac OS). Nonetheless, it should be sufficient to demonstrate the issue, so I've uploaded it as a zip here. I didn't want to push such a quick and dirty version to the main repo.

The correct place to put the setNeedsDisplay call is in winit, and I didn't want to dig through it to figure out exactly where and deal with a non-standard version for testing when it might not even work. Of course, if this ever works, I'll submit a PR to winit to add it in.

[...]

PR #2189, opened 2022-02-08
Modified repaint system on Mac OS to use the OS-mediated system instead of doing it entirely in winit to better support non-GPU-accelerated rendering

john01dav:

There has been prior discussion of this issue and how I found out that this is the needed fix:
 - https://github.com/servo/core-foundation-rs/issues/497
 - https://github.com/john01dav/softbuffer/pull/1
 - https://github.com/rust-windowing/winit/issues/1605

The short version of the context for this change is that in working on [softbuffer](https://crates.io/crates/softbuffer), my new crate to allow GPU-less 2D display to a raw window handle window I encountered a bug where once something was shown to a Mac OS window (during a resize) then nothing more could be shown until the window resized. After much digging, it turned out that the cause for this is that winit is not properly notifying Mac OS of the repaints in order to cause it to allow it to push these repaints to the screen. This is because on Mac OS winit handles repaints entirely within itself and does not notify the OS or allow it to mediate them. This pull request changes that by making request_repaint() notify the OS instead of simply tracking it internally. It also removes the internal tracking system as it is no longer used.

My biggest concern with this pull request is that it might break GPU-accelerated rendering. I can't test this myself as the only Mac computer that I have access to is too old to support Metal and this Wgpu. I tried to find some opengl crates with usable examples for testing, but what I found didn't work properly before the change so the results from after the change are not meaningful. **As such, before this is merged it is *vital* that someone who has access to a more modern Mac computer test this with wgpu and other GPU-accelerated rendering systems to ensure that they are not broken**.

### DIFF
--- a/src/platform_impl/macos/app_state.rs
+++ b/src/platform_impl/macos/app_state.rs
@@ -33,7 +33,7 @@ use crate::{
             event::{EventProxy, EventWrapper},
             event_loop::{post_dummy_event, PanicInfo},
             menu,
-            observer::{CFRunLoopGetMain, CFRunLoopWakeUp, EventLoopWaker},
+            observer::EventLoopWaker,
             util::{IdRef, Never},
             window::get_window_id,
         },
@@ -133,7 +133,6 @@ struct Handler {
     start_time: Mutex<Option<Instant>>,
     callback: Mutex<Option<Box<dyn EventHandler>>>,
     pending_events: Mutex<VecDeque<EventWrapper>>,
-    pending_redraw: Mutex<Vec<WindowId>>,
     waker: Mutex<EventLoopWaker>,
 }
 
@@ -143,10 +142,6 @@ unsafe impl Sync for Handler {}
 impl Handler {
     fn events(&self) -> MutexGuard<'_, VecDeque<EventWrapper>> {
         self.pending_events.lock().unwrap()
-    }
-
-    fn redraw(&self) -> MutexGuard<'_, Vec<WindowId>> {
-        self.pending_redraw.lock().unwrap()
     }
 
     fn waker(&self) -> MutexGuard<'_, EventLoopWaker> {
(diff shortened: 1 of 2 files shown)


tinaun:
this currently does break gpu rendering in my wgpu project, unfortunately

kchibisov:
>Change the public API and other platforms' implementations to have a parameter passed to request_redraw to specify whether hardware-accelerated rendering is being used.

In general you either use window in hardware acceleration context or not. So as a solution if macOS requires something else in software acceleration context you could add a window builder option for macOS like `with_hardware_acceleration(bool)`. Enabled by default or `could_hardware_accelerate(bool)`.

john01dav:
> > Change the public API and other platforms' implementations to have a parameter passed to request_redraw to specify whether hardware-accelerated rendering is being used.
> 
> In general you either use window in hardware acceleration context or not. So as a solution if macOS requires something else in software acceleration context you could add a window builder option for macOS like `with_hardware_acceleration(bool)`. Enabled by default or `could_hardware_accelerate(bool)`.

Having a separate API on different platforms seems questionable at best. Ideally, users of softbuffer could just see that the platforms that they care about are supported and then not care about any platform-specific concerns any further than that.

john01dav:
In order to get softbuffer fully working on mac os as quickly as possible, iI've re-added the non-OS-mediated repaint system, and tested against both softbuffer and wgpu and both work. After this is merged, I'll push a softbuffer update warning about not doing anything heavy in RedrawRequested's handler and to cache, due to the double call when not using GPU-accelerated rendering on Mac OS.

john01dav:
@kpreid The OS-mediated system is a no-op when using GPU-based games, so it won't affect that use case.

kpreid:
I understood your description as that all applications, GPU-accelerated or not, would receive _two_ `RedrawRequested` events per frame or per request. If that's not the case, then I have no concerns.

john01dav:
That is indeed not the case. To confirm, only non-GPU-accelerated applications (which presently means only softbuffer to the best of my knowledge) receive two `RedrawRequested` events. GPU-accelerated applications receive only one. 

I just re-read my description and I see now that is very unclear. I'll edit it to make it more clear.

Issue #1634, opened 2020-07-27
DeviceEvent should be optional, causes unneccessary CPU usage


Ciantic:
I use winit for background application for Windows 10. 

Even though not a single window is visible or activated it eats CPU if I move the mouse: Process Explorer shows 0.33% of CPU usage when moving a mouse, but when I stop the mouse it goes to 0%. 

This indicates that it just keeps on listening the device events, even though I don't even need them. It's not ideal for a background application to eat CPU when mouse is moved outside the application's own window.

I don't know is background applications (that occasionally show a Window) even a target for winit, but I surely would want it to be.

Thanks.

---

There is multiple ways to fix this, easiest but not so great is to make an `emit_device_events(bool)` for the EventLoop builder. Another way would be to make them *toggleable* at runtime, so one could toggle DeviceEvents on when necessary (e.g. when the window is activated).

alvinhochun:
Yeah, I don't quite understand why one would *by default* want to receive mouse movements and keyboard events when not in foreground.

Is this behaviour present on other platforms?

chrisduerr:
There's some more discussion in https://github.com/rust-windowing/winit/issues/1775, but it boils down to the same issue. Specifically the mouse events are the problem, not DeviceEvents in general.

Ciantic:
To me the biggest issue is that it winit windows also acts as keylogger, they just keeps listening keyboard events they have no business to.

chrisduerr:
That should be irrelevant. Any application that wants to keylog on this application can, offering the ability to do so changes nothing. It's absolutely trivial to get access to. So the only thing that matters is if the application actually uses this information or not and that has nothing to do with winit providing it.

Ciantic:
I'm pretty sure WM_KEYUP, WM_KEYDOWN, and WM_CHAR is not routed to all applications as I type right now? There is something else going on in winit than that.

garasubo:
In Windows, DeviceEvents are abstraction of Raw Input. And [the document](https://docs.microsoft.com/en-us/windows/win32/inputdev/about-raw-input) said:
> By default, no application receives raw input. To receive raw input from a device, an application must register the device.

So, technically, we can make them as optional. 

In X11, DevicEvents are abstraction of `XI_Raw.*` events. To receive them, an application need to set the mask by calling `XISelectEvents`. So we can make them as optional as well.

[...]

PR #2195, opened 2022-02-16
Add X11 opt-in function for device events

chrisduerr:
Previously on X11, by default all global events were broadcasted to
every Winit application. This unnecessarily drains battery due to
excessive CPU usage when moving the mouse.

To resolve this, device events are now ignored by default and users must
manually opt into it using `EventLoop::set_filter_device_events`.

Fixes (#1634) on Linux.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - Added `Window::set_ime_allowed` supported on desktop platforms.
 - **Breaking:** IME input on desktop platforms won't be received unless it's explicitly allowed via `Window::set_ime_allowed` and new `WindowEvent::Ime` events are handled.
 - On macOS, `WindowEvent::Resized` is now emitted in `frameDidChange` instead of `windowDidResize`.
+- **Breaking:** On X11, device events are now ignored for unfocused windows by default, use `EventLoopWindowTarget::set_device_event_filter` to set the filter level.
 
 # 0.26.1 (2022-01-05)
 
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -286,6 +286,28 @@ impl<T> EventLoopWindowTarget<T> {
     pub fn primary_monitor(&self) -> Option<MonitorHandle> {
         self.p.primary_monitor()
     }
+
+    /// Change [`DeviceEvent`] filter mode.
+    ///
+    /// Since the [`DeviceEvent`] capture can lead to high CPU usage for unfocused windows, winit
+    /// will ignore them by default for unfocused windows on Linux/BSD. This method allows changing
+    /// this filter at runtime to explicitly capture them again.
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **Wayland / Windows / macOS / iOS / Android / Web**: Unsupported.
+    ///
+    /// [`DeviceEvent`]: crate::event::DeviceEvent
+    pub fn set_device_event_filter(&mut self, _filter: DeviceEventFilter) {
+        #[cfg(any(
+            target_os = "linux",
+            target_os = "dragonfly",
+            target_os = "freebsd",
+            target_os = "netbsd",
+            target_os = "openbsd"
+        ))]
+        self.p.set_device_event_filter(_filter);
+    }
 }
 
 /// Used to send custom events to `EventLoop`.
@@ -330,3 +352,20 @@ impl<T> fmt::Display for EventLoopClosed<T> {
 }
 
 impl<T: fmt::Debug> error::Error for EventLoopClosed<T> {}
+
+/// Fiter controlling the propagation of device events.
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Debug)]
+pub enum DeviceEventFilter {
+    /// Always filter out device events.
+    Always,
+    /// Filter out device events while the window is not focused.
+    Unfocused,
+    /// Report all device events regardless of window focus.
+    Never,
+}
+
+impl Default for DeviceEventFilter {
+    fn default() -> Self {
+        Self::Unfocused
+    }
+}
--- a/src/platform_impl/linux/mod.rs
+++ b/src/platform_impl/linux/mod.rs
@@ -30,7 +30,9 @@ use crate::{
     dpi::{PhysicalPosition, PhysicalSize, Position, Size},
     error::{ExternalError, NotSupportedError, OsError as RootOsError},
     event::Event,
-    event_loop::{ControlFlow, EventLoopClosed, EventLoopWindowTarget as RootELW},
+    event_loop::{
+        ControlFlow, DeviceEventFilter, EventLoopClosed, EventLoopWindowTarget as RootELW,
+    },
     icon::Icon,
     monitor::{MonitorHandle as RootMonitorHandle, VideoMode as RootVideoMode},
     window::{CursorIcon, Fullscreen, UserAttentionType, WindowAttributes},
@@ -732,7 +734,7 @@ impl<T: 'static> EventLoop<T> {
     }
 
     pub fn window_target(&self) -> &crate::event_loop::EventLoopWindowTarget<T> {
-        x11_or_wayland!(match self; EventLoop(evl) => evl.window_target())
+        x11_or_wayland!(match self; EventLoop(evlp) => evlp.window_target())
     }
 }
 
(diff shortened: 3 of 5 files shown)


kchibisov (changes requested):
Could you make it part of the event loop builder. And probably a unix extension only for x11? I'm not sure that we'd need that for other platforms. Also, have you verified  that it fixes the underlying issue?

chrisduerr:
> And probably a unix extension only for x11? I'm not sure that we'd need that for other platforms.

Just to clarify this: Even if it was only needed on X11 I'd try to make it available on all platforms just so downstream doesn't have to put it behind a cfg. Unless the functionality ***can't*** be implemented on some platforms, there's no reason not to offer it. In this case it would just be no-op without any issues obviously.

> Also, have you verified that it fixes the underlying issue?

Yup, tested that it works at startup, tested that it works manually activating/deactivating repeatedly. And verified that the CPU usage does actually stay at zero whenever device events are filtered out. So it works like a charm.

This is also why I'm not sure if a builder is the right place for it though: It is possible to change this at runtime without any problem, so why would we force people to only set it at startup? And there's no reason to have both really because you don't get events right after event loop creation anyway so doing it in a separate statement right after `EventLoop::new` doesn't change anything.

So what benefit do we gain by putting it in the builder? As far as I can tell, none.

src/event_loop.rs
```diff
@@ -185,6 +185,28 @@ impl<T> EventLoop<T> {
             event_loop_proxy: self.event_loop.create_proxy(),
         }
     }
+
+    /// Change [`DeviceEvent`] filter mode.
+    ///
+    /// Since the [`DeviceEvent`] capture can lead to high CPU usage for unfocused windows, winit
+    /// will ignore them by default on Linux/BSD. This method allows changing this filter at
```
madsmtm, line 192: Wayland is not implemented yet
```suggestion
    /// will ignore them by default on X11. This method allows changing this filter at
```

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - **Breaking:** Bump `ndk` version to 0.6, ndk-sys to `v0.3`, `ndk-glue` to `0.6`.
 - Remove no longer needed `WINIT_LINK_COLORSYNC` environment variable.
 - **Breaking:** Rename the `Exit` variant of `ControlFlow` to `ExitWithCode`, which holds a value to control the exit code after running. Add an `Exit` constant which aliases to `ExitWithCode(0)` instead to avoid major breakage. This shouldn't affect most existing programs.
+- **Breaking:** On X11, device events are now ignored by default, use `EventLoop::set_filter_device_events` to opt-into device events again.
```
madsmtm, line 16: I'm not sure making it disabled by default is a good idea while implementations for the other platforms don't exist.

src/event_loop.rs
```diff
@@ -185,6 +185,28 @@ impl<T> EventLoop<T> {
...
+    /// runtime to explicitly capture them again.
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **Wayland / Windows / macOS / iOS / Android / Web**: Unsupported.
```
madsmtm, line 197: I think `set_filter_device_events` could make sense on other platforms as well, so maybe `s/Unsupported/Not implemented`?

madsmtm (changes requested):
A benefit of having it on the `EventLoopBuilder` is that we expose slightly less functionality, giving us more freedom further down the line (as an example, a proper macOS implementation possibly requires changing whether a method is implemented or not, and things can definitely be made more efficient).

Also, please ensure the examples that use `DeviceEvent` work after this change.

CHANGELOG.md
kchibisov: It's not really true, we have some device events on Wayland, like pointer motion information which is something used by e.g. games.

src/event_loop.rs
chrisduerr: As far as I understand, `Unsupported` means `Unsupported` by Winit. I've considered they way I should phrase it and chose to go for what most of the other Winit documentation uses. I think especially with documentation consistency matters more here and I don't want to change it everywhere.

src/event_loop.rs
chrisduerr: I've intentionally chosen this wording because they already get ignored on Wayland, simply because there is no other option.

kchibisov (approved):

src/event_loop.rs
```diff
@@ -215,6 +215,28 @@ impl<T> EventLoopWindowTarget<T> {
...
+    ///
+    /// [`DeviceEvent`]: crate::event::DeviceEvent
+    pub fn set_filter_device_events(&self, filter_events: bool) {
```
maroider, line 230: This is going to trigger an `unused_variables` warning.

CHANGELOG.md
maroider: The "correct" default would be to have device events be tied to window focus, which (AIUI) is what we get on Wayland. You'd then opt out of this behaviour on platforms where that's allowed.

kchibisov:
Could probably do the following:

```rust
pub enum DeviceEventFilter {
    /// Always filter device events.
    Always
    /// Only receive device events when the window is being focused.
    Focused
    /// Don't filter device events.
    Never
}
```

And then use `Focused` by default. The naming could change, but we certainly could go beyond a simple boolean.

chrisduerr:
Should now only filter device events for unfocused windows by default.

CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On Wayland, fix polling during consecutive `EventLoop::run_return` invocations.
 - On Windows, fix race issue creating fullscreen windows with `WindowBuilder::with_fullscreen`
 - On Android, `virtual_keycode` for `KeyboardInput` events is now filled in where a suitable match is found.
+- **Breaking:** On X11, device events are now ignored for unfocused windows by default, use `EventLoop::set_device_event_filter` to set the filter level.
```
kchibisov, line 35: Should we mention that it's on `EventLoopWindowTarget`, I think it's correct for that, since the `run` and `run_return` callbacks take target?

chrisduerr: Sure.

kchibisov (approved):

src/event_loop.rs
madsmtm: You're right, the rest of the docs aren't consistent on this point.

src/event_loop.rs
```diff
@@ -251,6 +251,28 @@ impl<T> EventLoopWindowTarget<T> {
     pub fn primary_monitor(&self) -> Option<MonitorHandle> {
         self.p.primary_monitor()
     }
+
+    /// Change [`DeviceEvent`] filter mode.
+    ///
+    /// Since the [`DeviceEvent`] capture can lead to high CPU usage for unfocused windows, winit
+    /// will ignore them by default for unfocused windows on Linux/BSD. This method allows changing
```
madsmtm, line 258: Just checked, unfocused applications do not receive `DeviceEvent`s on macOS.

madsmtm (changes requested):
Re putting this on `EventLoopBuilder` (static configuration before creation) vs. `EventLoop` (dynamically after creation, but before running) vs. `EventLoopWindowTarget` (dynamically while running):

Of course it is _possible_ to change this setting dynamically on macOS, but given that the reason we're even making this change is for efficiency reasons, the question becomes more "would it be somewhat efficient?".

The answer here is a bit more nuanced, and I haven't actually implemented it so I can't say for sure, but I know it would be at least somewhat nontrivial to ensure thread safety of the underlying functions needed like `CFRunLoopAddSource`/`CFRunLoopRemoveSource` / method swizzling. I also do know that only having the possibility to change this once, before the event loop is created, is definitely the most efficient option.

But I guess my objection mostly arises from different values: I want to expose the minimal set of APIs that solves people's problems - if we then find a concrete use-case that need us to expose something more powerful, I'm up for it, but until then the more free to change implementation details we can be, the better.

(Aside: `set_device_event_filter(DeviceEventFilter::Always)` would probably require root access on macOS, so maybe it should return a `Result`?)

kchibisov:
>(Aside: set_device_event_filter(DeviceEventFilter::Always) would probably require root access on macOS, so maybe it should return a Result?)

We can say that it's a no-op. Given that on Wayland you'd need something like that as well + direct libinput access. So I don't think we should return result, just saying that it's not-supported is the way to go, given that running things under root isn't something anyone should do.

Changing on the fly could be required for global hotkeys like options, so if the user added global hotkey it must be changed on the fly.

src/event_loop.rs
```diff
@@ -286,6 +286,28 @@ impl<T> EventLoopWindowTarget<T> {
...
+    /// this filter at runtime to explicitly capture them again.
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **Wayland / Windows / macOS / iOS / Android / Web**: Unsupported.
+    ///
+    /// [`DeviceEvent`]: crate::event::DeviceEvent
+    pub fn set_device_event_filter(&mut self, _filter: DeviceEventFilter) {
```
wusyong, line 301: There isn't a `DerefMut` implementation for `EventLoop` to `EventLoopWindowTarget`.
Do you think we should add it? Or keep it immutable reference only and find workaround.

Issue #2192, opened 2022-02-12
On Windows release events does not get triggered while dragging


feelingnothing:
Windows 10 21H2

```rust
use winit::event::{ButtonId, DeviceEvent, ElementState, Event, MouseButton, WindowEvent};
use winit::event_loop::EventLoop;
use winit::window::WindowBuilder;

fn main() {
    let event_loop = EventLoop::new();
    let window = WindowBuilder::new().build(&event_loop).unwrap();

    event_loop.run(move |event, _, _| if let Event::WindowEvent { event, ..} = event {
        if let WindowEvent::MouseInput { state, button, .. } = event {
            if button != MouseButton::Left { return; }
            match state {
                ElementState::Pressed => {
                    println!("Pressed");
                    let _ = window.drag_window();
                }
                ElementState::Released => {
                    println!("Released");
                }
            }
        }
    });
}
```

MSRP:
Clone https://github.com/emilk/egui.git, and paste into `hello_world` example this piece of code into the update function
```rust
let egui::InnerResponse { response, .. } = egui::TopBottomPanel::new(TopBottomSide::Top, "titlebar").show(&ctx, |ui| {
    if ui.button("Button").clicked() {
        println!("Button Clicked");
    }
});

if response.interact(egui::Sense::drag()).dragged() {
    frame.drag_window();
}
```
Go to `egui-winit/src/lib.rs` and print the mouse event at 227 line and run the example, drag the top area and release the mouse button, then try to click everywhere at the program and you will see that the only events that are coming are pressed, but not released
Sometimes it works, sometimes not, when you change the focus from the window, it fixes itself
![result](https://user-images.githubusercontent.com/47575622/153730101-51aea5a0-ca9b-4944-af11-02ccb3369802.gif)

feelingnothing:
After some debugging I have come to solution to reimplement dragging with naive but for sure working dragging approach with last position stored in `WindowState`, because if we are making the windows think that we are pressing on taskbar title, we do not get release events on left mouse button
https://github.com/rust-windowing/winit/blob/f9643917d32efaf7e8e923c342e60dc593a813ca/src/platform_impl/windows/window.rs#L338-L343

PR #2200, opened 2022-02-19
On Windows, fix left mouse button release event not being sent on drag finish

feelingnothing:
After tricking windows into thinking we are pressing on `HTCAPTION`, we would not get `WM_NCLBUTTONUP`, thus creating unexpected behavior in `winit` dependent projects (https://github.com/emilk/egui/issues/1245#issue-1134534176)
Some debugging after it appears that `wine` have different logic that `winapi` (for context see [this](https://matrix.to/#/!DGpLzJTRzBDTwZiogk:matrix.org/$8hlJG0C6Eiorsj0suXpUr5IOXx0if1nhIfPRHoMwZI8?via=matrix.org&via=libera.chat))

With event tracing we can see that the most suitable event triggered on every drag is `WM_EXITSIZEMOVE`([docs](https://docs.microsoft.com/en-us/windows/win32/winmsg/wm-exitsizemove)):
image

- Fixes: https://github.com/rust-windowing/winit/issues/2192

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - Added `Window::is_decorated`.
 - On X11, fix for repeated event loop iteration when `ControlFlow` was `Wait`
 - On Wayland, report unaccelerated mouse deltas in `DeviceEvent::MouseMotion`.
+- On Windows, fix left mouse button release event not being sent on drag finish
 - **Breaking:** Bump `ndk` version to 0.6, ndk-sys to `v0.3`, `ndk-glue` to `0.6`.
 - Remove no longer needed `WINIT_LINK_COLORSYNC` environment variable.
 - **Breaking:** Rename the `Exit` variant of `ControlFlow` to `ExitWithCode`, which holds a value to control the exit code after running. Add an `Exit` constant which aliases to `ExitWithCode(0)` instead to avoid major breakage. This shouldn't affect most existing programs.
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -886,15 +886,21 @@ unsafe fn public_window_callback_inner<T: 'static>(
         }
 
         winuser::WM_EXITSIZEMOVE => {
-            userdata
-                .window_state
-                .lock()
-                .set_window_flags_in_place(|f| f.remove(WindowFlags::MARKER_IN_SIZE_MOVE));
+            let mut window_state = userdata.window_state.lock();
+            // when tricking windows into thinking that we are pressing on the HTCAPTION,
+            // we would not get WM_NCLBUTTONUP, so we have to check it manually,
+            // and when dragging is done - release the button ourselves
+            if window_state.dragging {
+                window_state.dragging = false;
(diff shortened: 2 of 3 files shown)


emilk:
I'd be interested in seeing this merged! the `drag_window` feature is critical for users that want to draw their own window frame.

msiglreith (changes requested):
Thanks, but this leads to undesired `Released` events if we are actually dragging inside the `HTCAPTION` area.
One idea could be to instead emit a message with `HTCLIENT` instead of `HTCAPTION` on `drag_window`, use it as indicator that it's a user generated event in the event loop, but forwarding `HTCAPTION` to `DefWindowProc` to internally let it do the dragging stuff for us..

maroider:
Can't we just set the flag inside [`Window::drag_window`](https://github.com/rust-windowing/winit/blob/2ae12fb0a0f5c81e5c1ab5c690bbe5a138368ca0/src/platform_impl/windows/window.rs#L339-L360). `Window` should have access to `WindowState`.

feelingnothing:
> Can't we just set the flag inside [`Window::drag_window`](https://github.com/rust-windowing/winit/blob/2ae12fb0a0f5c81e5c1ab5c690bbe5a138368ca0/src/platform_impl/windows/window.rs#L339-L360)? `Window` should have access to `WindowState`.
I mean there is no difference, anyways to have to check for `HTCAPTION` in `WM_NCLBUTTONDOWN`, either we do it in `drag_window` or in event handler doesn't matter. I like how it makes all dragging logic in dragging function, but check for `HTCAPTION` is dragging logic too.

feelingnothing:
> Thanks, but this leads to undesired `Released` events if we are actually dragging inside the `HTCAPTION` area. One idea could be to instead emit a message with `HTCLIENT` instead of `HTCAPTION` on `drag_window`, use it as indicator that it's a user generated event in the event loop, but forwarding `HTCAPTION` to `DefWindowProc` to internally let it do the dragging stuff for us..

I don't think there is a way to `Release` event to be emitted in unwanted situation, considering dragging flag in window state. With event logger from @maroider we can see that `WM_EXITSIZEMOVE` is the most suitable one. The windows api tells that this event is trigged in the end of every move or resize, considering the dragging flag I'm sure nothing out of ordinary would happend

msiglreith:
> Can't we just set the flag inside [Window::drag_window](https://github.com/rust-windowing/winit/blob/2ae12fb0a0f5c81e5c1ab5c690bbe5a138368ca0/src/platform_impl/windows/window.rs#L339-L360)? Window should have access to WindowState.

Right, I like this more under the light `drag_resize_window`. I guess we could even drop then the whole handling `NCLBUTTONDOWN` event.

> I don't think there is a way to Release event to be emitted in unwanted situation, considering dragging flag in window state.

Sorry if I wasn't clear on it. Normal dragging without `drag_window` using the native decorations from Windows will emit the same `HTCAPTION` message which we use for emulating. So in both cases the dragging flag will be set leading to the released event. In the case of native decorations this is not desired as we do not emit the corresponding Pressed event.

maroider:
> I guess we could even drop then the whole handling `NCLBUTTONDOWN` event.

We're handling that to fix an unrelated issue.

Issue #2112, opened 2021-12-27
WindowEvent { Focused } occurs after WindowEvent { MouseInput { Pressed } } on web backend


rukai:
Heres a screenshot from running the web.rs example
![image](https://user-images.githubusercontent.com/5120858/147465782-8ccb44a6-2345-4fb0-8bee-7966c57679b3.png)

The other backends return the window focus first and my application and https://github.com/rukai/winit_input_helper rely on this functionality.
The current handling in the web backend causes the first mouse click event to be lost.

rukai:
Given that any click event should also result in a window focus event, I think the solution might be to generate an extra window focus event here like we already do for CursorMoved.
https://github.com/rust-windowing/winit/blob/7e0c6ee09757663efacd9d7ba7c70ff101ec4e35/src/platform_impl/web/event_loop/window_target.rs#L153+L176

PR #2202, opened 2022-02-20
web: Manually emit focused event on mouse click

rukai:
closes https://github.com/rust-windowing/winit/issues/2112

Reasoning is explained on the issue.


rukai:
Resolved the merge conflict

grovesNL (approved):
Thank you, this seems reasonable to me!

It seems like we could also consider handling the element's `focus` event generally, but I'm not sure we could rely on the event order (e.g. `focus` might fire after `mousedown`).

maroider (approved):
A future refinement on this might be to avoid sending a second `Focused(true)` event when our `on_blur` handler gets run.

rukai:
Thanks!

PR #2205, opened 2022-02-23
On Wayland, fix resize not propagating properly

kchibisov:
On Wayland window size and scaling are double buffered, so winit should
send redraw requested for a client on the next frame as well.


kchibisov:
As an example issue is with glutin multi window example on sway.

without patch

picture

with patch.

picture

PR #2208, opened 2022-02-25
web: Add `EventLoop::spawn`

### LINKED ISSUE #1714: Suggestion: Add a web-only event loop run API that returns to caller
Motivation
---

With `EventLoop::run` it is currently impossible to start the event loop cleanly from either JavaScript or Rust since it throws an exception to not return execution to the caller. Technically one can use JS to catch the exception, but if someone wants to do this in Rust, it would require some glue code. It is also a pretty awful hack.

Moreover, if the event loop is started from `#[wasm_bindgen(start)]`, the JS `Promise` returned from initializing the WASM module will be rejected, making it harder to distinguish between a successful load and a failure in initializing the WASM module. One can compare the exception string, but this is also quite dirty.

Proposal
---

A new web-only extension method should be added to the `EventLoop`. which kick-starts the event loop and returns without throwing an exception. The caller is still required to return execution to the browser for the event loop to actually start handling events.

```rust
trait EventLoopExtWeb {
    type UserEvent;

    fn run_web(self, event_handler: F)
    where
        F: FnMut(
            Event<'_, Self::UserEvent>,
            &EventLoopWindowTarget<Self::UserEvent>,
            &mut ControlFlow,
        );
}
```

Unresolved questions
---

- Does `run_web` need to keep the `EventLoopWindowTarget` alive beyond the function scope and how? In the existing web backend, it only holds `runner::Shared`, which is effectively kept alive by the DOM event listener closures in the form of an `Rc` loop, so dropping `EventLoopWindowTarget` will not actually cause the event loop to be dropped, but it may not be wise to rely on this behaviour.
- Should there be a way for the user to retrieve and reuse the `EventLoop` after it has exited? In the case of `EventLoopExtDesktop::run_return`, the `EventLoop` and its associated `Window`s can be reused easily. However, to do the same on web will require `run_web` to return a `Future` or a `Promise`, which becomes *ready* or *fulfilled* respectively when the `EventLoop` exits.
- Should we allow the user to create multiple `EventLoop`s which can be run simultaneously? Technically there shouldn't be any issues allowing this,

Liamolucko:

This is the same as `EventLoop::run`, but doesn't throw an exception in order to return `!`.

I decided to name it `spawn` rather than `run_web` because I think that's more descriptive, but I'm happy to change it to `run_web`.

Resolves #1714


grovesNL (approved):
This looks great, thank you!

I agree that `spawn` might be a little more descriptive than `run_web`, so that name sounds reasonable to me.

src/platform/web.rs
```diff
@@ -27,3 +31,38 @@ impl WindowBuilderExtWebSys for WindowBuilder {
         self
     }
 }
+
+/// Additional methods on `EventLoop` to spawn the event loop without throwing an exception.
```
maroider, line 35:
```suggestion
/// Additional methods on `EventLoop` that are specific to the web.
```
I suppose the trait is "single-use" right now, but I don't believe we call that out in other single-platform extension traits.

maroider (approved):
I think you just need to resolve the merge conflicts now.

PR #2209, opened 2022-02-26
Add web_aspect_ratio example

rukai:

Adds a new example that creates a web canvas that fills an aspect-ratio'd div
The main point of this example is to demonstrate the functionality provided by https://github.com/rust-windowing/winit/pull/2074
The example can land before or after, but landing it before will make it easier to test #2074

Another benefit of this example is demonstrating how to create a web canvas that fills up the available space while maintaining the same aspect ratio.
Although this is slightly out of scope of winit I believe it would be extremely useful to provide as figuring out what combination of html elements and css configuration I needed to get this to work was surprisingly difficult for me even though I've been dabbling with web technologies for ages.

Pinging @liamolucko for feedback as this relates to your PR.

Currently running `cargo run-wasm --example web_aspect_ratio` will give:
image
With #2074 and running `RUSTFLAGS="--cfg=web_sys_unstable_apis" cargo run-wasm --example web_aspect_ratio --features css-size` will give:
image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,7 +55,6 @@ mint = { version = "0.5.6", optional = true }
 [dev-dependencies]
 image = { version = "0.24.0", default-features = false, features = ["png"] }
 simple_logger = { version = "2.1.0", default_features = false }
-web-sys = { version = "0.3.4", features = ['CanvasRenderingContext2d'] }
 
 [target.'cfg(target_os = "android")'.dependencies]
 # Coordinate the next winit release with android-ndk-rs: https://github.com/rust-windowing/winit/issues/1995
@@ -143,6 +142,7 @@ version = "0.2.45"
 
 [target.'cfg(target_arch = "wasm32")'.dev-dependencies]
 console_log = "0.2"
+web-sys = { version = "0.3.22", features = ['CanvasRenderingContext2d'] }
 
 [workspace]
 members = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,6 +55,7 @@ mint = { version = "0.5.6", optional = true }
 [dev-dependencies]
 image = { version = "0.24.0", default-features = false, features = ["png"] }
 simple_logger = { version = "2.1.0", default_features = false }
+web-sys = { version = "0.3.4", features = ['CanvasRenderingContext2d'] }
 
 [target.'cfg(target_os = "android")'.dependencies]
 # Coordinate the next winit release with android-ndk-rs: https://github.com/rust-windowing/winit/issues/1995
--- a/examples/web_aspect_ratio.rs
+++ b/examples/web_aspect_ratio.rs
@@ -6,7 +6,7 @@ pub fn main() {
 mod wasm {
     use wasm_bindgen::prelude::*;
     use wasm_bindgen::JsCast;
-    use web_sys::{HtmlCanvasElement, HtmlElement};
+    use web_sys::HtmlCanvasElement;
     use winit::{
         dpi::PhysicalSize,
         event::{Event, WindowEvent},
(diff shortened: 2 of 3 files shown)


examples/web_aspect_ratio.rs
```diff
@@ -0,0 +1,107 @@
...
+            .style()
+            .set_css_text("margin: auto; width: 50%; aspect-ratio: 4 / 1;");
+        body.append_child(&parent_div).unwrap();
+
+        // Set a background color for the canvas to make it easier to tell the where the canvas is for debugging purposes.
+        let canvas = window.canvas();
+        canvas
+            .style()
+            .set_css_text("display: block; width: 100%; height: 100%; background-color: crimson;");
+        parent_div.append_child(&canvas).unwrap();
```
Liamolucko, line 77: You shouldn't need to create a parent `<div>`; applying the CSS directly to the canvas works fine.

examples/web_aspect_ratio.rs
```diff
@@ -0,0 +1,107 @@
...
+
+    #[wasm_bindgen(start)]
+    pub fn run() {
+        console_log::init_with_level(log::Level::Debug).expect("error initializing logger");
+        let event_loop = EventLoop::new();
+
+        let window = WindowBuilder::new()
+            .with_title("A fantastic window!")
+            // A small default size is used to better demonstrate issues that come from failing to update the size
+            .with_inner_size(PhysicalSize::new(100, 100))
```
Liamolucko, line 33: This could be a bit misleading to people reading the example, since you almost never want to use `with_inner_size` when sizing using `ResizeObserver`; it works in this case since the canvas' `style` gets overridden below, but that should probably be documented more clearly.

Something like: 'Note: this would normally fix the canvas' width and height to 100px, but since the canvas' `style` gets overridden below in `create_canvas`, it just sets the initial framebuffer size.'

examples/web_aspect_ratio.rs
rukai: oh I thanks I didnt realize that.
I think I prefer to have the div separate in my actual application as it means I can have a generic canvas created by wasm that the html template can configure the size of by setting the css on the parent div.

However in this example it probably makes sense to just keep it simple and do all in one considering we cant control the raw html file anyway.
I've changed the PR to remove the div but am happy to revert if someone else prefers the original way.

Cargo.toml
```diff
@@ -34,6 +34,7 @@ mint = { version = "0.5.6", optional = true }
 [dev-dependencies]
 image = { version = "0.24.0", default-features = false, features = ["png"] }
 simple_logger = "2.1.0"
+web-sys = { version = "0.3.4", features = ['CanvasRenderingContext2d'] }
```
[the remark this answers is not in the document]
rukai: Not sure where I got that version from...
Its functionally identical but I updated to use the same version as the non dev-dep to make it less confusing.

rukai:
I've resolved merge conflicts, would appreciate a review.

Cargo.toml
```diff
@@ -55,6 +55,7 @@ mint = { version = "0.5.6", optional = true }
 [dev-dependencies]
 image = { version = "0.24.0", default-features = false, features = ["png"] }
 simple_logger = { version = "2.1.0", default_features = false }
+web-sys = { version = "0.3.22", features = ['CanvasRenderingContext2d'] }
```
madsmtm, line 58: Can you put this in a `[target.'cfg(target_arch = "wasm32")'.dev-dependencies]` instead?

madsmtm (changes requested):
I'm not sure I understand how it is supposed to look like now, and how it is supposed to look like after the ResizeObserver?

Here's a screenshot of what it looks like to me on Firefox:

screenshot

rukai:
I have addressed your feedback and added screenshots of both to the PR description.

madsmtm (changes requested):
Thanks, that was helpful! Would you mind noting in the example that the behaviour is not as it should be _yet_, but will be after the other PR?

rukai:
done!

madsmtm (approved):
Thanks, that makes it much more understandable

Issue #2123, opened 2022-01-01
Terminal updates after monitor disable & enable cycle broken with #1963


d-e-s-o:
Hi,

Over in https://github.com/alacritty/alacritty/issues/5703 we have been trying to get to the root of a bug that manifests in the alacritty terminal not being redrawn properly after the monitor got disabled and enabled. Since this is a regression, I bisected it to 
125ee0b446eab48dda56390c96b8f1973d45b3e6 (https://github.com/rust-windowing/winit/pull/1963).

Is this a known issue?

maroider:
Thanks for reporting this. As far as I'm aware, this hasn't been reported before.

d-e-s-o:
I played around with the code some. If I make the following change:
```patch
--- src/platform_impl/linux/x11/event_processor.rs
+++ src/platform_impl/linux/x11/event_processor.rs
@@ -1182,7 +1182,7 @@ impl<T: 'static> EventProcessor<T> {
                                                 // If there all monitors are closed before, scale
                                                 // factor would be already changed to 1.0.
                                                 maybe_prev_scale_factor.unwrap_or(1.0),
-                                                new_monitor.scale_factor,
+                                                1.0,
                                                 width,
                                                 height,
                                                 &*window.shared_state.lock(),
```

everything works fine. So I printed out `new_monitor.scale_factor` and it's `1.4166666666666667` (also see the [log](https://github.com/alacritty/alacritty/files/7812481/Alacritty-12601.log) I attached to the aforementioned `alacritty` task). So, if I change the code as follows:
```patch
--- src/platform_impl/linux/x11/event_processor.rs
+++ src/platform_impl/linux/x11/event_processor.rs
@@ -1181,7 +1181,7 @@ impl<T: 'static> EventProcessor<T> {
                                             let (new_width, new_height) = window.adjust_for_dpi(
                                                 // If there all monitors are closed before, scale
                                                 // factor would be already changed to 1.0.
-                                                maybe_prev_scale_factor.unwrap_or(1.0),
+                                                maybe_prev_scale_factor.unwrap_or(1.4166666666666667),
                                                 new_monitor.scale_factor,
                                                 width,
                                                 height,
```
everything works. So it appears that because we are defaulting to a wrong scaling factor of `1.0` we screw up the window update, setting sizes that don't make sense.

d-e-s-o:
[...]

PR #2210, opened 2022-02-26
Fix scale factor calculation when the only monitor is reconnected

d-e-s-o:
The scale factor being sent when the only monitor is disconnected and
reconnected is hard coded to 1.0. That may work by chance, if that's the
scale factor in use currently, but it does not work in the general case.
As a result, clients may end up with wrongly scaled or laid out window
contents after reconnect, as was reported over in
https://github.com/alacritty/alacritty/issues/5703, for example.

The problem was introduced by change 125ee0b, which caused an additional
ScaleFactorChanged event to be sent on monitor reconnect, but got the
scale factor wrong when the only monitor is disconnected and
reconnected.
This change fixes the problem by using the current monitor's scale
factor in this case. The event is still being sent as intended by
125ee0b.

Closes: #2123

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - Added `Window::is_resizable`.
 - Added `Window::is_decorated`.
 - On X11, fix for repeated event loop iteration when `ControlFlow` was `Wait`
+- On X11, fix scale factor calculation when the only monitor is reconnected
 - On Wayland, report unaccelerated mouse deltas in `DeviceEvent::MouseMotion`.
 - **Breaking:** Bump `ndk` version to 0.6, ndk-sys to `v0.3`, `ndk-glue` to `0.6`.
 - Remove no longer needed `WINIT_LINK_COLORSYNC` environment variable.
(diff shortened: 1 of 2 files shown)


kchibisov (approved):

d-e-s-o:
Thanks for the review and approval, @kchibisov. May I ask what the next steps here are?

kchibisov:
>Thanks for the review and approval, @kchibisov. May I ask what the next steps here are?

I'm not the X11 maintainer and don't have ability to test it, but the change sounds good to me. So unless some X11 maintainer or tester reply I'm not sure it'll go farther.

Though, if it stay that way for a while I could just merge it.

Xuanwo:
I'm using X11 right now, is there a quick way to verify this PR works or not?

kchibisov:
I mean you can poweroff/on monitor that is using scale factor != 1. If it picks something other than `1` when powering on, it should work.

d-e-s-o:
> I mean you can poweroff/on monitor that is using scale factor != 1. If it picks something other than `1` when powering on, it should work.

Yep, what Kirill said. Should be the last monitor to be turned off, though, I believe.

Xuanwo:
I tested the behavior, and it works.

I tested with `cargo run --example window` and make the size of the window 1/4 of my display whose scale factor is 2.

- on current main branch: After powering off/on my display, the windows' size changed to the 
 full size of my display.
- on this PR: the size of the window is unchanged.

maroider:
I guess this is simple enough to merge. If no-one protests, I guess I'll merge this.

PR #2211, opened 2022-02-26
Faster builds

madsmtm:
WIP


rukai:
You might want to consider using https://github.com/Swatinem/rust-cache which does all this sort of stuff for you.
Only downside is you'll have to checkin the Cargo.lock file but you should really be doing that anyway considering you have deny warnings enabled.

madsmtm:
Replaced with https://github.com/rust-windowing/winit/pull/2988, where we will also discuss committing `Cargo.lock`.

PR #2213, opened 2022-03-01
feat(Windows): add `with_translate_accel_callback`

amrbashir:

### DIFF
--- a/src/platform/windows.rs
+++ b/src/platform/windows.rs
@@ -15,6 +15,8 @@ use crate::{
     window::{BadIcon, Icon, Theme, Window, WindowBuilder},
 };
 
+pub type TranslateAcceleratorCallback = dyn FnMut(winapi::um::winuser::MSG) -> bool;
+
 /// Additional methods on `EventLoop` that are specific to Windows.
 pub trait EventLoopBuilderExtWindows {
     /// Whether to allow the event loop to be created off of the main thread.
@@ -51,6 +53,49 @@ pub trait EventLoopBuilderExtWindows {
     /// # }
     /// ```
     fn with_dpi_aware(&mut self, dpi_aware: bool) -> &mut Self;
+
+    /// A callback, executed before dispatching a win32 message to the window procedure, where you
+    /// can call [`TranslateAcceleratorW`](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-translateacceleratorw) to
+    /// processes accelerator keys for menu commands.
+    ///
+    /// The callback return value is used as an indicator to whether the message
+    /// should be translated using [`TranslateMessage`](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-translatemessage)
+    ///
+    /// # Example
+    ///
+    /// ```no_run
+    /// thread_local! {
+    ///     static ACCELS:Vec<ACCEL> = {
+    ///         unsafe {
+    ///             vec![ACCEL {
+    ///                 cmd: 554,
+    ///                 key: VkKeyScanW('a' as u16) as _  ,
+    ///                 fVirt: VK_CONTROL as _,
+    ///             }]
+    ///         }
+    ///     }
+    /// }
+    ///
+    /// fn main() {
+    ///     let event_loop = EventLoopBuilder::new()
+    ///         .with_translate_accel_callback(Box::new(|mut m| unsafe {
+    ///             let mut translated = false;
+    ///             ACCELS.with(|accels| {
+    ///                 translated = TranslateAcceleratorW(
+    ///                     m.hwnd,
+    ///                     CreateAcceleratorTableW(accels.as_ptr() as _, 1),
+    ///                     &mut m,
+    ///                 ) == 1;
+    ///             });
+    ///             translated
+    ///         }))
+    ///         .build();
+    /// }
+    /// ```
+    fn with_translate_accel_callback(
+        &mut self,
+        callback: Box<TranslateAcceleratorCallback>,
+    ) -> &mut Self;
 }
 
 impl<T> EventLoopBuilderExtWindows for EventLoopBuilder<T> {
@@ -63,6 +108,15 @@ impl<T> EventLoopBuilderExtWindows for EventLoopBuilder<T> {
     #[inline]
     fn with_dpi_aware(&mut self, dpi_aware: bool) -> &mut Self {
         self.platform_specific.dpi_aware = dpi_aware;
+        self
+    }
+
+    #[inline]
+    fn with_translate_accel_callback(
+        &mut self,
+        callback: Box<TranslateAcceleratorCallback>,
+    ) -> &mut Self {
+        self.platform_specific.translate_accel_callback = Some(callback);
         self
     }
 }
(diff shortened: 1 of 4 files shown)


src/platform/windows.rs
```diff
@@ -51,6 +51,51 @@ pub trait EventLoopBuilderExtWindows {
...
+    ///                     CreateAcceleratorTableW(accels.as_ptr() as _, 1),
+    ///                     &mut m,
+    ///                 ) == 1;
+    ///             });
+    ///             translated
+    ///         }))
+    ///         .build();
+    /// }
+    /// ```
+    fn with_translate_accel_callback(
```
msiglreith, line 95: As the implementation is not tied to to accelerators internally, we could also just give it a more general name like `with_msg_hook`

src/platform_impl/windows/event_loop.rs
```diff
@@ -173,6 +175,7 @@ impl<T: 'static> EventLoop<T> {
                 },
                 _marker: PhantomData,
             },
+            translate_accel_callback: Some(Box::new(|_| false)),
```
msiglreith, line 178: shouldn't this pass `attributes.translate_accel_callback` here?

src/platform_impl/windows/event_loop.rs
```diff
@@ -213,8 +216,15 @@ impl<T: 'static> EventLoop<T> {
                 if 0 == winuser::GetMessageW(&mut msg, ptr::null_mut(), 0, 0) {
                     break 'main 0;
                 }
-                winuser::TranslateMessage(&msg);
-                winuser::DispatchMessageW(&msg);
+
+                let translated = self
+                    .translate_accel_callback
+                    .as_deref()
+                    .unwrap_or(&Box::new(|_| false))(msg);
```
msiglreith, line 223: The fallback callback could be created once on creation of the eventloop avoiding having to reallocate it on each iteration.

src/platform/windows.rs
```diff
@@ -65,6 +110,15 @@ impl<T> EventLoopBuilderExtWindows for EventLoopBuilder<T> {
         self.platform_specific.dpi_aware = dpi_aware;
         self
     }
+
+    #[inline]
+    fn with_translate_accel_callback(
+        &mut self,
+        callback: Box<dyn FnMut(winapi::um::winuser::MSG) -> bool>,
```
msiglreith, line 117: As part of a public facing API `MSG` needs to be re-exported or even better a struct defined locally by `winit`.

msiglreith (changes requested):
thanks, some points would need to be addressed - given the recent switch to `windows-sys` the `winapi` parts require some work as well

src/platform/windows.rs
amrbashir: I like that, this could also let users disable winit's internal Dispatch/Translate and allow them to use their own calls to Dispatch/Translate.

src/platform_impl/windows/event_loop.rs
amrbashir: Right. Silly mistake.

src/platform_impl/windows/event_loop.rs
amrbashir: Great optimization, thanks. I think I can avoid allocating altogether for a slightly uglier code.

src/platform/windows.rs
amrbashir: I guess reexporting is better so it can be used in other win32 apis directly.

amrbashir:
I will be updating the PR to `windows-sys` in a couple of days.

src/platform/windows.rs
```diff
@@ -19,6 +19,8 @@ pub type HMENU = isize;
 pub type HMONITOR = isize;
 /// Instance Handle type used by Win32 API
 pub type HINSTANCE = isize;
+/// Message queue type used by Win32 API.
+pub use windows_sys::Win32::UI::WindowsAndMessaging::MSG;
```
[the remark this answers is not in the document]
amrbashir: I guess `*mut c_void` is better. The closure should be able to mutate the `MSG` passed to it as seen in the example.

maroider (approved):
LGTM

src/platform/windows.rs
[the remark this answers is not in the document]
msiglreith: > The closure should be able to mutate the MSG passed to it as seen in the example.

Is there a reason for using a `*mut` pointer instead of `*const`? I currently don't see a reason for being able to mutate the message as the user can also just create a local copy.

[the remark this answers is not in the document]
amrbashir: I meant that as an argument against defining winit's version of `MSG` because I thought `(Translate|Dispatch)Message` functions mutates the passed `MSG` but after checking the docs, it doesn't so both options are viable. I am leaning towards using `*const c_void` but let me know what you prefer.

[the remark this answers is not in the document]
msiglreith: `*const c_void` is fine for me

src/platform/windows.rs
```diff
@@ -56,6 +56,38 @@ pub trait EventLoopBuilderExtWindows {
...
+    /// builder.with_msg_hook(Box::new(|mut msg|{
+    ///     let msg = msg as *mut MSG;
+    /// #   let accels: Vec<ACCEL> = Vec::new();
+    ///     unsafe {
+    ///         let translated = TranslateAcceleratorW(
+    ///             (*msg).hwnd,
+    ///             CreateAcceleratorTableW(accels.as_ptr() as _, 1),
+    ///             msg,
+    ///         ) == 1;
+    ///         if !translated {
```
msiglreith, line 82: this branch could be removed imo, and just let the winit internal part handle the message - keeping it minimal

src/platform/windows.rs
```diff
@@ -56,6 +56,38 @@ pub trait EventLoopBuilderExtWindows {
...
     /// ```
     fn with_dpi_aware(&mut self, dpi_aware: bool) -> &mut Self;
+
+    /// A callback to be executed before dispatching a win32 message to the window procedure.
+    /// Return true to disable winit's internal message dispatching.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use  windows_sys::Win32::UI::WindowsAndMessaging::{ACCEL, CreateAcceleratorTableW, TranslateAcceleratorW, DispatchMessageW, TranslateMessage, MSG};
```
msiglreith, line 66:
```suggestion
    /// # use windows_sys::Win32::UI::WindowsAndMessaging::{ACCEL, CreateAcceleratorTableW, TranslateAcceleratorW, DispatchMessageW, TranslateMessage, MSG};
```

src/platform/windows.rs
```diff
@@ -56,6 +56,38 @@ pub trait EventLoopBuilderExtWindows {
...
+    ///         ) == 1;
+    ///         if !translated {
+    ///             TranslateMessage(msg);
+    ///             DispatchMessageW(msg);
+    ///         }
+    ///     }
+    ///     true
+    /// }));
+    /// ```
+    fn with_msg_hook(&mut self, callback: Box<dyn FnMut(*mut c_void) -> bool>) -> &mut Self;
```
msiglreith: I think we can also hide the `Box` in the interface by making the funktion generic and box it internally.
```rust
fn with_msg_hook<F>(&mut self, callback: F) where F: FnMut(*const c_void) -> bool + 'static
```

msiglreith (commented):
A few smaller points besides that it should be good to go

msiglreith (approved):
thanks!

Issue #2116, opened 2021-12-28
scrolling is not captured on web mobile


rukai:
On the web on desktop devices, scrolling the mouse while the cursor is over the winit window will not scroll the webpage.
This is good and desirable.

However one finger dragging over the winit window on a mobile device will scroll the webpage.
This is bad, particularly on mobile as the cursor position events reports strange movement due to the page scroll negating the movement of the 'cursor'.

Libraries like three-js handle both of these cases by preventing scrolling.

Possibly it will be fixed by just handling the relevant events in https://github.com/rust-windowing/winit/pull/1945
Or maybe we need to do more.

rukai:
After experimenting with https://github.com/rust-windowing/winit/pull/1945 it doesnt yet solve this problem but it can be adjusted to fix it with the addition of calls to `event.prevent_default()`.

This change is enough to get it working.
![image](https://user-images.githubusercontent.com/5120858/156861039-72a61367-e52a-4dcf-aebc-839ffafae921.png)
We would obviously need to rewrite to not use those `mem::forget`s though.

However thinking about this there are valid use cases for allowing scrolling while the canvas is focused.
You could use a winit for rendering a completely noninteractive animation or you could have a UI that never needs to click and drag. In these cases being able to scroll freely on the web page is more natural for the user.

As such maybe the solution here is to move all the `event.prevent_default()` callbacks behind a `web_disable_scroll()` method.

However allowing scrolling by default is quite problematic as any sort of click and drag functionality ends up completely broken on mobile.
So we would probably be better off defaulting to disabling web scroll and providing a `web_enable_scroll()` method to enable it for the users who know they really want it.

PR #2216, opened 2022-03-05
WindowBuilderExtWebSys::with_prevent_default disables scrolling on both mobile and desktop (previously just desktop)

rukai:
closes https://github.com/rust-windowing/winit/issues/2116

Tested on firefox on my desktop and firefox on my android phone.

### DIFF
--- a/src/platform/web.rs
+++ b/src/platform/web.rs
@@ -19,15 +19,6 @@ pub trait WindowExtWebSys {
 }
 
 pub trait WindowBuilderExtWebSys {
-    /// Enable scrolling of the web page the canvas is in when the canvas is focused.
-    ///
-    /// Scrolling is disabled by default because the scroll input on many mobile devices
-    /// is the same as click and dragging which is a very common input method for many applications.
-    ///
-    /// So only call this method if you know that you will never need to handle mouse wheel inputs
-    /// or click and dragging.
-    fn enable_web_page_scroll(self) -> Self;
-
     fn with_canvas(self, canvas: Option<HtmlCanvasElement>) -> Self;
 
     /// Whether `event.preventDefault` should be automatically called to prevent event propagation
@@ -43,12 +34,6 @@ pub trait WindowBuilderExtWebSys {
 }
 
 impl WindowBuilderExtWebSys for WindowBuilder {
-    fn enable_web_page_scroll(mut self) -> Self {
-        self.platform_specific.enable_web_page_scroll = true;
-
-        self
-    }
-
     fn with_canvas(mut self, canvas: Option<HtmlCanvasElement>) -> Self {
         self.platform_specific.canvas = canvas;
 
--- a/src/platform/web.rs
+++ b/src/platform/web.rs
@@ -19,6 +19,15 @@ pub trait WindowExtWebSys {
 }
 
 pub trait WindowBuilderExtWebSys {
+    /// Enable scrolling of the web page the canvas is in when the canvas is focused.
+    ///
+    /// Scrolling is disabled by default because the scroll input on many mobile devices
+    /// is the same as click and dragging which is a very common input method for many applications.
+    ///
+    /// So only call this method if you know that you will never need to handle mouse wheel inputs
+    /// or click and dragging.
+    fn enable_web_scroll(self) -> Self;
+
     fn with_canvas(self, canvas: Option<HtmlCanvasElement>) -> Self;
 
     /// Whether `event.preventDefault` should be automatically called to prevent event propagation
@@ -34,6 +43,12 @@ pub trait WindowBuilderExtWebSys {
 }
 
 impl WindowBuilderExtWebSys for WindowBuilder {
+    fn enable_web_scroll(mut self) -> Self {
+        self.platform_specific.enable_web_scroll = true;
+
+        self
+    }
+
     fn with_canvas(mut self, canvas: Option<HtmlCanvasElement>) -> Self {
         self.platform_specific.canvas = canvas;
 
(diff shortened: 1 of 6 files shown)


rukai:
Hmmmm not sure how much it matters but should be tagged web not android

src/platform_impl/web/web_sys/canvas.rs
```diff
@@ -72,11 +73,34 @@ impl Canvas {
...
+                }),
+                common.add_event("touchstart", move |event: Event| {
+                    event.prevent_default();
+                }),
+                common.add_event("touchend", move |event: Event| {
+                    event.prevent_default();
+                }),
+                common.add_event("wheel", move |event: Event| {
+                    event.prevent_default();
+                }),
```
madsmtm, line 97: I think I'd rather keep the `event.prevent_default()` calls inside the original event handlers, and then check a boolean in there, instead of registering new event handlers just for that.

src/platform/web.rs
```diff
@@ -17,10 +17,25 @@ pub trait WindowExtWebSys {
...
 
 pub trait WindowBuilderExtWebSys {
+    /// Enable scrolling of the web page the canvas is in when the canvas is focused.
+    ///
+    /// Scrolling is disabled by default because the scroll input on many mobile devices
+    /// is the same as click and dragging which is a very common input method for many applications.
+    ///
+    /// So only call this method if you know that you will never need to handle mouse wheel inputs
+    /// or click and dragging.
+    fn enable_web_scroll(self) -> Self;
```
madsmtm, line 27: Suggestion: `enable_scroll`?

rukai: I think the web there is important, the user needs to know this is a web specific hack.

rukai: Despite the method being on a trait called WindowBuilderExtWebSys when the method is called there is no indication which trait the method comes from.
As such I think its better to make it explicit that this is a web specific method.
I think `enable_web_scroll` does this well enough.
But now I'm thinking we should go a step further and name it `enable_web_page_scroll` to help avoid the misunderstanding that this affects the canvas's ability to receive scroll events.

src/platform_impl/web/web_sys/canvas.rs
rukai: I'm not super familiar with web, what is the concern here?
My guess is performance but I want to check.

rukai: I've resolved merge conflicts, so would appreciate a response here.
But also, when I get time I'll try to just rewrite as suggested, as I'd rather not stall on something small like this.

madsmtm: Sorry, I'd forgotten all about this.

I think my was mostly that I don't know enough about how `preventDefault` actually works to say with certainty whether the fact that it works here is part of the spec, or accidental.

Also, it seems like in the meantime https://github.com/rust-windowing/winit/pull/2365 was merged, doesn't that solve exactly the same problem?

rukai: This PR introduces `prevent_default` calls for many new kinds of events, so its still needed in some form.

However I think you make a good point and we should just put these new `prevent_default`s behind `with_prevent_default` instead of introducing a new `enable_web_scroll`.

I think generally an application falls into wants every prevent_default call or wants none of them.
I believe the few outliers can then manually make their own prevent_default calls directly through web_sys if really needed.
And if winit decides it needs to provide finer grained selection of its prevent_defaults, the API can always be changed later.

Does that sound good to you?

rukai: I went ahead and did that.
The new functionality of this PR is now behind with_prevent_default.

rukai: > I think my was mostly that I don't know enough about how preventDefault actually works to say with certainty whether the fact that it works here is part of the spec, or accidental.

I feel like the only reasonable interpretation of [preventDefault](https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault)  is that if any of the callbacks call prevent_default then the default callback handler is not run.
Additionally I am not duplicating any callbacks, so I'm not even hitting that possible edge case any more.
So I believe my implementation is sound now and will be sound even if additional event handlers are added.

That said if you want me to write it in a different way im all ears.

madsmtm: I still think I would like it all to be handled inside one callback, like `prevent_default` works currently.

rukai: I'm not quite following, these events don't have an existing callback.

madsmtm: Apologies! I got confused because we do have an existing event for `pointermove`, so I would like that one to handled like the rest; but you're right that we don't have existing `touchstart`/`touchend` events, you should add those 

```rust
// Canvas
pub fn on_touch_start(&mut self, prevent_default: bool) {
    self.on_touch_start = Some(self.common.add_event("touchstart", move |event: TouchEvent| {
        if prevent_default {
            event.prevent_default();
        }
    }));
}

// EventLoopWindowTarget::register
canvas.on_touch_start(prevent_default);
canvas.on_touch_end(prevent_default);
```

rukai: Oh my bad, I missed pointermove already had a callback.
I've made the changes as requested and verified it still functions correctly.

madsmtm (approved):
Thanks, that looks good!

I changed the changlog entry to note it as a breaking change, while it doesn't change the API it changes the semantics.

PR #2218, opened 2022-03-15
Always send RedrawEventsCleared

HackerFoo:
For all platforms other than iOS, `RedrawEventsCleared` is sent even if there are no redraw events. This change makes iOS consistent with the other platforms.

This solves a bug on iOS where Bevy relies on this event in https://github.com/bevyengine/bevy/pull/3974


kchibisov (changes requested):
Could you add changelog entry for fixing that?

kchibisov (approved):

PR #2057, opened 2021-11-06
Adopt windows-sys

clemenswasser:
Change from using the unofficial winapi bindings, to using the official windows bindings from Microsoft.

- [x] Tested on all platforms changed
- [x] Compilation warnings were addressed
- [x] `cargo fmt` has been run on this branch
- [x] `cargo doc` builds successfully
- [x] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users
- [ ] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior
- [ ] Created or updated an example program if it would help users understand this functionality
- [ ] Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented

Issue #1816, opened 2021-01-05
Duplicate dependencies


baptistemanson:
Hi all,
I'm a absolute beginner in Rust.

While using the command "cargo tree -d", I saw duplicates on `core-graphics, core-foundation-sys, core-foundation`.
- Am I reading the dups correctly?
- Is it fair to say that updating core-video-sys to more recent libs would shave of around 100kB of dups out of the 300kB of winit?

Thanks for your patience!
```
core-graphics v0.19.2
└── core-video-sys v0.1.4
    └── winit v0.24.0
        └── wgpu v0.1.0 (/Users/inovia/Desktop/rust/wgpu)

core-graphics v0.22.2
├── cocoa v0.24.0
│   └── winit v0.24.0
│       └── wgpu v0.1.0 (/Users/inovia/Desktop/rust/wgpu)
└── winit v0.24.0 (*)
```

madsmtm:
This is not an issue with `winit`, but with `core-video-sys`, they need to release a new version with newer dependencies (last release was 2020-05-06). See https://github.com/LuoZijun/rust-core-video-sys/issues/16 for a similar issue/suggestion.

PR #2220, opened 2022-03-18
Quicker Mac builds by removing dependencies

kettle11:
This PR is not ready to go, but I wanted to create a pull request to have a record of the concept.  It might be useful as a point of reference even if not directly used.

This change removes direct dependencies on the `cocoa`, `core-foundation`, and `core-graphics` crates by copy-pasting the used parts from those libraries directly into `winit`.

On my M1 Mac Air this dropped release build times from ~3.95 seconds to ~2.7 seconds.

No functionality was changed other than one tiny segment of code that I was too lazy to figure out how to port over.

For this to become a real PR it would need a few things:

* More organized copy-pasted code. I didn't have a formed strategy when I started out so the structure is all over the place.
* An evaluation and plan of the ongoing maintenance burden added by this approach.
* A fix for that one spot I was too lazy to port correctly.
* Consideration of how to properly attribute the libraries copied from.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,9 +42,9 @@ ndk-glue = "0.6"
 
 [target.'cfg(any(target_os = "ios", target_os = "macos"))'.dependencies]
 objc = "0.2.7"
+libc = "0.2.64"
 
 [target.'cfg(target_os = "macos")'.dependencies]
-cocoa = "0.24"
 core-foundation = "0.9"
 core-graphics = "0.22"
 dispatch = "0.2.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,14 +45,7 @@ objc = "0.2.7"
 libc = "0.2.64"
 
 [target.'cfg(target_os = "macos")'.dependencies]
-core-foundation = "0.9"
-core-graphics = "0.22"
 dispatch = "0.2.0"
-
-[target.'cfg(target_os = "macos")'.dependencies.core-video-sys]
-version = "0.1.4"
-default_features = false
-features = ["display_link"]
 
 [target.'cfg(target_os = "windows")'.dependencies]
 parking_lot = "0.12"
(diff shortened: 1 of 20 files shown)


madsmtm:
Thanks for the idea! On my machine a clean build goes from ~18s to ~10s, so this would be nice!

However, I don't think the tradeoff is worth it - in part because I think I want to move towards using something else entirely (like [`cacao`](https://github.com/ryanmcgrath/cacao) or something), but mostly because I don't think we can really cope with such a massive added maintenance burden.

(I know I argued strongly for low build times in https://github.com/rust-windowing/winit/pull/2057, but that was when the difference _needlessly_ went from 13 seconds to around 1 minute).

(Also, this would solve https://github.com/rust-windowing/winit/issues/1816).

kettle11:
There's definitely added maintenance burden so it's reasonable to not pick this approach but I don't think it's as massive as it seems initially. 

I say that for a few reasons:
* Many (most?) of the lines in the above PR are commented out and the change could be reduced substantially
* It's probably not *super* common that a new system API is needed, and it's only a slight effort to find new needed bindings and copy-paste them into Winit.
* The code copied is all relatively simple, so it's less likely to regress dramatically. 
* It's also unlikely the underlying binding crates are going to be fixed or updated in ways that leave Winit behind.

I think the biggest cost is that when adding new functionality to Winit that requires new bindings there's a non-trivial amount of friction that makes implementing that new functionality slightly more frustrating. This might slow continuing maintenance.

madsmtm:
I grabbed part of `core-video-sys` in https://github.com/rust-windowing/winit/pull/2326 because it was going stale, but I don't think I'll be replacing `cocoa`, `core-foundation` and `core-graphics` just yet - I have other plans for how to do this (mainly, I want to replace them with a [safe abstraction](https://github.com/madsmtm/objc2)!).

Still, thanks for the PR!

PR #2221, opened 2022-03-21
macOS: Fix emitting `Event::LoopDestroyed` on CMD+Q / pressing Quit in the menu bar

madsmtm:
Fixes https://github.com/rust-windowing/winit/issues/1998

Alternatively we could do https://github.com/rust-windowing/winit/pull/2073 - we _might_ still want to do that, since it makes our cleanup more robust, but I think there is merit in using `stop:` over `terminate:`, since it would allow applications using `run_return` to continue running.

- [x] ~Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior~


tinaun:
I don't know of any macOS apps where using cmd-q or [App] -> Quit doesn't terminate the entire program, however

Issue #2224, opened 2022-03-25
Creating window as fullscreen borderless does not work on Windows


repi:
When one creates a window with `WindowBuilder::with_fullscreen(Some(winit::window::Fullscreen::Borderless(Some(monitor_id)))` it doesn't actually create a borderless fullscreen window, but instead a default sized & positioned smaller window that doesn't cover the entire monitor.

Looks like one may get `Moved(0,0)` and `Resized(w,h)` events corresponding the the borderless fullscreen, but also directly after that get another position & size set. So may be a bug of that being sent incorrectly in the Windows window creation code / fullscreen toggle logic?

## Repro

This is reproducible in the example.

1. Run: `cargo run --example fullscreen`
1. Pick option 2 for borderless
1. Pick 0 for default/main monitor

Result: window is created and has no title bar or frame, but it is not fullscreen.

Only tested on a Windows 11 machine so far, with a single monitor.

repi:
Looks like the `fullscreen` sample here occasionally work.

This run didn't work, here with a log output of the `Resized` event on WM_SIZE:

```sh
$ cargo run --example fullscreen
    Finished dev [unoptimized + debuginfo] target(s) in 0.06s
     Running `target\debug\examples\fullscreen.exe`
Please choose the fullscreen mode: (1) exclusive, (2) borderless: 2
Monitor #0: Some("\\\\.\\DISPLAY1\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}")
Please write the number of the monitor to use: 0
Using Some("\\\\.\\DISPLAY1\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}")
[src\platform_impl\windows\event_loop.rs:1036] physical_size = PhysicalSize {
    width: 2864,
    height: 1104,
}
[src\platform_impl\windows\event_loop.rs:1036] physical_size = PhysicalSize {
    width: 800,
    height: 600,
}
[src\platform_impl\windows\event_loop.rs:1036] physical_size = PhysicalSize {
    width: 816,
    height: 639,
}
[...]

PR #2225, opened 2022-03-25
Fix fullscreen window messaging race on Windows

repi:
This attempts to fix issue that looks like a messaging race of window size/position changes on window creation when using `WindowBuilder::with_fullscreen` on Windows. 

Believe it is from most using `SWP_ASYNCWINDOWPOS` but some internal one such as `ITaskbarList2::MarkFullscreenWindow` not using it causing WM_SIZE (and likely more messages) often arriving in wrong order and making the fullscreen window that was created getting the wrong intermediate size and be stuck with that. 

This slightly rearranges the Windows fullscreen window size/pos setting to be in the absolute end, and avoid redundantly setting inner size or potential maximize state that also generated window resizing.

In local testing on Windows 11 machine this have been working solid both with the `fullscreen` example and with our own larger application that uses winit. But would be good to get confirmation from more devs on Win 10/11.

Resolves: #2224

## Notes


msiglreith (approved):
Thanks, the changes look sensible.
While I couldn't reproduce the initial issue on my Windows 10 machine, it still works fine with this PR (:

repi:
> While I couldn't reproduce the initial issue on my Windows 10 machine, it still works fine with this PR (:

Thanks for trying! Wonder if this is something that has changed with later Windows updates, or if it is just somewhat non-deterministic in its order or based on configuration/setup. Really hard to tell with these type of stuff on Windows

PR #2226, opened 2022-03-26
android: Add mapping from NDK Keycode to VirtualKeyCode

### LINKED ISSUE #1867: Android: Provide keycode for KeyEvents
Key handling has been implemented in https://github.com/rust-windowing/winit/pull/1839
The keycodes are currently left as `None`, but ndk provides access to the underlying codes from the retrieved `key`.

Requires mapping of Android KeyCode to winit's KeyCode.

torokati44:
When working on https://github.com/torokati44/ruffle-android, I discovered that all keypress events arrive with their scancode set to `0`, and their keycode set to `None`. Which I found less than useful.

The [NDK docs for AKeyEvent_getScanCode](https://developer.android.com/ndk/reference/group/input#akeyevent_getscancode) mention that "These values are not reliable and vary from device to device.".
So I guess always setting them to `0` is acceptable from their end.

Tested this on Android 12, with a keyboard connected to my computer, through wired `adb` and https://github.com/Genymobile/scrcpy/, and it seemed to work fine.

AFAICT This fixes #1867.

 - Only Android is affected, and I am testing on it.
 - I have not found any existing documentation that would have to be updated due to this change.
 - No new functionality, only implementing/fixing an existing one.
 - The change is not significant enough to warrant updating the matrix.

### DIFF
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -2,13 +2,14 @@
 
 use crate::{
     dpi::{PhysicalPosition, PhysicalSize, Position, Size},
-    error, event,
+    error,
+    event::{self, VirtualKeyCode},
     event_loop::{self, ControlFlow},
     monitor, window,
 };
 use ndk::{
     configuration::Configuration,
-    event::{InputEvent, KeyAction, MotionAction},
+    event::{InputEvent, KeyAction, Keycode, MotionAction},
     looper::{ForeignLooper, Poll, ThreadLooper},
 };
 use ndk_glue::{Event, Rect};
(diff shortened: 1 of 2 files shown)


torokati44:
I think this is now reviewable.

I have left quite a few `???` comments in there where it wasn't immediately obvious what the mapping should be.
What should I do about those? Delete those comments and accept the current mapping as a whole as "good enough", or remove any mapped keys which may not be perfect match? Or leave the uncertainty in there with those comments?

msiglreith (approved):
thanks! the less clear cases are fine in my eyes

src/platform_impl/android/mod.rs
```diff
@@ -44,6 +45,170 @@ enum EventSource {
...
+        Keycode::Enter => Some(VirtualKeyCode::Return), // not on the Numpad
+        Keycode::Tab => Some(VirtualKeyCode::Tab),
+
+        Keycode::PageUp => Some(VirtualKeyCode::PageUp),
+        Keycode::PageDown => Some(VirtualKeyCode::PageDown),
+        Keycode::MoveHome => Some(VirtualKeyCode::Home),
+        Keycode::MoveEnd => Some(VirtualKeyCode::End),
+        Keycode::Insert => Some(VirtualKeyCode::Insert),
+
+        Keycode::Del => Some(VirtualKeyCode::Back), // Backspace (above Enter)
```
MarijnS95, line 138: Started playing about with these codes, and it seems you haven't implemented a mapping for `KeyCode::Back` which would be needed to hanlde a graceful app exit (or going to the previous page) when the user presses it. The implementation for `VirtualKeyCode::Back` reads:

https://github.com/rust-windowing/winit/blob/f04fa5d54f4ec10cdb6d084deeb79d3e6d27ae67/src/event.rs#L962-L964

So I don't think we should add a new `Back` `VirtualKeyCode` and rename existing uses to `Backspace`, as that's really hard to notice for end-users upgrading their `winit`. Perhaps we can piggyback `NavigateBackward` but android already supports that as `KeyCode::NavigatePrevious`:

```rust
Keycode::Back => Some(VirtualKeyCode::NavigateBackward),
```

How do you think we should go forward?

PR #2228, opened 2022-03-27
Add methods to set ControlFlow

swooster:
Bare-bones `winit` applications must import many different types from `winit`... this PR reduces that by adding methods so `ControlFlow` can be updated without naming its type, hopefully resulting in minor simplifications to the examples.

 **(I've only tested on Windows, but I've run all examples)**
 **(not applicable)**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On Wayland, fix polling during consecutive `EventLoop::run_return` invocations.
 - On Windows, fix race issue creating fullscreen windows with `WindowBuilder::with_fullscreen`
 - On Android, `virtual_keycode` for `KeyboardInput` events is now filled in where a suitable match is found.
+- Added helper methods on `ControlFlow` to set its value.
 
 # 0.26.1 (2022-01-05)
 
--- a/examples/control_flow.rs
+++ b/examples/control_flow.rs
@@ -3,7 +3,7 @@ use std::{thread, time};
 use simple_logger::SimpleLogger;
 use winit::{
     event::{Event, KeyboardInput, WindowEvent},
-    event_loop::{ControlFlow, EventLoop},
+    event_loop::EventLoop,
     window::WindowBuilder,
 };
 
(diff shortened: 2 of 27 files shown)


madsmtm:
Huh, interesting idea! Not sure I'm really in favour, since it sort of hides the fact that you can query the current control flow.

If it's just the imports that bother you, we could consider making a prelude, though of course that has other disadvantages.

Regarding naming, maybe prefix the methods with `set_` or similar?

maroider:
I'm personally mostly OK with the proposed API, mostly because I find that method calls would be slightly more convenient to type out than `*control_flow = ControlFlow::Whatever`. The import isn't really that bothersome to me.

swooster:
I'm happy to prefix everything with `set_`. It now occurs to me that e.g. `control_flow.wait_until(instant);` could be mistaken for something that blocks immediately and `control_flow.set_wait_until(instant);` seems less likely to be misleading.

I'm _extremely_ wary of crate preludes after being subjected to wildcard-import-heavy code at my day job; I prefer to look for ways to avoid needing to do that.

Admittedly, importing `ControlFlow` isn't _that_ big a deal. It's just that after noticing this relatively simple tweak could shave off an import and eliminate redundant mentions of "control flow" from each control flow change, I figured it'd be worth suggesting via a PR.

swooster:
(ok, I've changed the names to prefix everything with `set_`)

maroider (approved):
LGTM

swooster:
(resolved conflicts)

CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On Wayland, fix polling during consecutive `EventLoop::run_return` invocations.
 - On Windows, fix race issue creating fullscreen windows with `WindowBuilder::with_fullscreen`
 - On Android, `virtual_keycode` for `KeyboardInput` events is now filled in where a suitable match is found.
+- Added `ControlFlow::set_poll`, `ControlFlow::set_wait`, `ControlFlow::set_wait_until`, `ControlFlow::set_exit_with_code`, and `ControlFlow::set_exit` methods so typical apps can avoid importing `ControlFlow`.
```
kchibisov, line 35: I don't like how long this is, could be shorter, like
```suggestion
- Added helper methods on `ControlFlow` to set its value.
```

kchibisov (changes requested):

kchibisov (approved):

swooster:
(rebased and incorporated changelog suggestion)

PR #2230, opened 2022-03-27
Doc tweak: Discourage use of WaitUntil to implement VSync

Ralith:
I occasionally observe user confusion regarding the best way to render at the display refresh rate rather than a wasteful unlimited FPS. `WaitUntil` is a poor choice because a loop based on it is unlikely to be in-phase with the refresh interval, putting applications at high risk of erratic frame drops, and may not be sufficiently precise. This added text should help guide people towards the more robust solution.

### DIFF
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -140,6 +140,10 @@ pub enum ControlFlow {
     Wait,
     /// When the current loop iteration finishes, suspend the thread until either another event
     /// arrives or the given time is reached.
+    ///
+    /// Useful for implementing efficient timers. Applications which want to render at the display's
+    /// native refresh rate should instead use `Poll` and the VSync functionality of a graphics API
+    /// to reduce odds of missed frames.
     WaitUntil(Instant),
     /// Send a `LoopDestroyed` event and stop the event loop. This variant is *sticky* - once set,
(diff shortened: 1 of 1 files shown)


src/event_loop.rs
```diff
@@ -140,6 +140,10 @@ pub enum ControlFlow {
     Wait,
     /// When the current loop iteration finishes, suspend the thread until either another event
     /// arrives or the given time is reached.
+    ///
+    /// Useful for implementing efficient timers. Applications which want to render at the display's
+    /// native refresh rate should instead use `Poll` and the VSync functionality of a graphics API
```
mooman219, line 145: Nit: You shouldn't use poll on web to render at the display's refresh rate. WebGL does not have control over vsync of buffer swapping. You should use `requestAnimationFrame` while on the web to request a redraw, and the browser will attempt to align that as appropriate. I believe this aligns to winit's requestRedraw, and redraw request events.

dhardy:
How does this work with frame callbacks? See: https://github.com/gfx-rs/wgpu/issues/2375

msiglreith (approved):
thanks - In particular on Windows the timer is very imprecise and winit also adds quite some overhead currently :(

Issue #2191, opened 2022-02-11
On macOS the frame size could change without window resize


kchibisov:
If you're using multiple windows and do tabbing for them the the of the frame could change without size of the window being changed, leading to winit not issuing resize events for the windows in tab group. So winit should send resize event when the frame changes as well, and not only when the window gets resized.

I'm not familiar with macOS API, however I could say that the frame rect we get in https://github.com/rust-windowing/winit/blob/0e52672f4aea376c3ef8a06b64aedca46a40e9b8/src/platform_impl/macos/view.rs#L324 is absolutely correct. And `window_did_resize` method on window delegate don't get called at all.

@madsmtm Do you known something about it? Seems like something that could be easily fixed?

sshock:
(Note: to test tabbed windows in macOS, make sure to set "Prefer tabs" to `always` in system preferences.)

I was able to confirm using the `multiwindow` example that hitting a key to create a new (tabbed) window triggers two `frameDidChange` and does not trigger `windowDidResize` (and therefore does not emit a `WindowEvent::Resized`).

However, here is the interesting part: if after creating one or more (tabbed) windows I use the mouse to resize the window, then after that point every new (tabbed) window created will trigger all four of these:
* frameDidChange
* frameDidChange
* windowDidResize
* frameDidChange

As for the rect size, I have noticed while testing with `multiwindow` in glutin that when resizing explicitly with the mouse the size is correct, but the one emitted when creating a new (tabbed) window has a height that is a little too big (like it's including the height of the tab).

The good news is it sounds like if we emit resize event from frame_did_change then not only will it fix the problem of it not getting emitted all the time, but it should have the correct rect.

It does seem though that even after this is fixed in some scenarios we will get multiple resize events, some with the wrong rect and some with the correct.  But I guess that is ok as long as the one with the correct one comes last?  (🤞)

kchibisov:
Yeah, the last resize is the only thing that matters here.

madsmtm:
Apologies for now answering this sooner, am satisfied with your resolution

PR #2239, opened 2022-04-04
macOS: Emit resize on frame_did_change

sshock:
* Fixes https://github.com/rust-windowing/winit/issues/2191
* Fixes downstream https://github.com/alacritty/alacritty/issues/5876

- (n/a) Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior
- (n/a) Created or updated an example program if it would help users understand this functionality
- (n/a) Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented

For testing:
* First make sure "Prefer tabs" is set to "always" in macOS system preferences
    * (so new windows will get grouped as tabs).
* Add a debug statements such as `println!("emitting resize event with size = {:?}", size);`
    * Add to `emit_resize_event` method in src/platform_impl/macos/window_delegate.rs.
    * Add to `frame_did_change` method in src/platform_impl/macos/view.rs. (next to new code)
* Run the `multiwindow` example program.
    * Each time you hit a key to create a new window, note that the last resize event emitted has the correct size.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - **Breaking:** Added new `WindowEvent::Ime` supported on desktop platforms.
 - Added `Window::set_ime_allowed` supported on desktop platforms.
 - **Breaking:** IME input on desktop platforms won't be received unless it's explicitly allowed via `Window::set_ime_allowed` and new `WindowEvent::Ime` events are handled.
+- On macOS, `WindowEvent::Resized` is now emitted in `frameDidChange` instead of `windowDidResize`.
 
 # 0.26.1 (2022-01-05)
 
--- a/src/platform_impl/macos/view.rs
+++ b/src/platform_impl/macos/view.rs
@@ -20,7 +20,7 @@ use objc::{
 };
 
 use crate::{
-    dpi::LogicalPosition,
+    dpi::{LogicalPosition, LogicalSize},
     event::{
         DeviceEvent, ElementState, Event, Ime, KeyboardInput, ModifiersState, MouseButton,
         MouseScrollDelta, TouchPhase, VirtualKeyCode, WindowEvent,
@@ -400,8 +400,17 @@ extern "C" fn frame_did_change(this: &Object, _sel: Sel, _event: id) {
             userData:ptr::null_mut::<c_void>()
             assumeInside:NO
         ];
-
         state.tracking_rect = Some(tracking_rect);
+
+        // Emit resize event here rather than from windowDidResize because:
+        // 1. When a new window is created as a tab, the frame size may change without a window resize occurring.
+        // 2. Even when a window resize does occur on a new tabbed window, it contains the wrong size (includes tab height).
+        let logical_size = LogicalSize::new(rect.size.width as f64, rect.size.height as f64);
+        let size = logical_size.to_physical::<u32>(state.get_scale_factor());
+        AppState::queue_event(EventWrapper::StaticEvent(Event::WindowEvent {
+            window_id: WindowId(get_window_id(state.ns_window)),
+            event: WindowEvent::Resized(size),
+        }));
     }
 }
 
--- a/src/platform_impl/macos/window_delegate.rs
+++ b/src/platform_impl/macos/window_delegate.rs
@@ -98,14 +98,6 @@ impl WindowDelegateState {
         AppState::queue_event(wrapper);
     }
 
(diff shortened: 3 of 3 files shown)


src/platform_impl/macos/view.rs
```diff
@@ -317,6 +317,8 @@ extern "C" fn view_did_move_to_window(this: &Object, _sel: Sel) {
             assumeInside:NO
         ];
         state.tracking_rect = Some(tracking_rect);
+
+        // TBD: Do we need to emit a resize event here like we do in frame_did_change?
```
kchibisov, line 321: I'm not sure. But you can check it by looking through logs in various cases and see if it's called during 'tab' related stuff and whether the size is correct. If it always called once, maybe we should resize here and not in frameDidChange. Also, I'd assume that every resize is followed by frameDidChange or prefix by it. So we could remove resize from windowDidResize?

sshock: Ok, I did some testing and yes, the viewDidMoveToWindow is called when creating and deleting (tabbed) window, but it has the wrong size for new tabbed windows.  So it looks like we don't want or need this one.  I'll go ahead and remove the comment.

You are correct that every windowDidResize is prefixed or followed by a frameDidChange.  For normal resizing with the mouse, frameDidChange comes first.  For a new window, if windowDidResize gets called at all, the frameDidChange still comes first, however in the case of a tabbed window, frameDidChange also gets called after and that is the one that has the correct value we need.

I get the feeling that when creating a new tabbed window, macOS behaves as if it first exists without the tab bar (so has a taller height), and then an instant later decides that it is being placed in a tab so then it evokes another frameDidChange with the correct size (that excludes the tab height).

Anyway, regarding windowDidResize I think you're right we can actually remove emitting a resize from there now.  Shall I do that?

kchibisov: >Anyway, regarding windowDidResize I think you're right we can actually remove emitting a resize from there now. Shall I do that?

Yeah, go ahead.

kchibisov (changes requested):
You should also add a changelog entry for that change.

sshock:
> You should also add a changelog entry for that change.

Done!

CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On Wayland, fix polling during consecutive `EventLoop::run_return` invocations.
 - On Windows, fix race issue creating fullscreen windows with `WindowBuilder::with_fullscreen`
 - On Android, `virtual_keycode` for `KeyboardInput` events is now filled in where a suitable match is found.
+- On macOS: `WindowEvent::Resized` is now emitted when `frameDidChange` is called and is no longer emitted when `windowDidResize` is called.
```
kchibisov, line 35:
```suggestion
- On macOS, `WindowEvent::Resized` is now emitted when `frameDidChange` is called and is no longer emitted when `windowDidResize` is called.
```

src/platform_impl/macos/view.rs
```diff
@@ -337,8 +337,14 @@ extern "C" fn frame_did_change(this: &Object, _sel: Sel, _event: id) {
             userData:ptr::null_mut::<c_void>()
             assumeInside:NO
         ];
-
         state.tracking_rect = Some(tracking_rect);
+
+        let logical_size = LogicalSize::new(rect.size.width as f64, rect.size.height as f64);
```
kchibisov, line 342: Could you add an explanation comment why resize is done here, so it'll be easier to understand the motivation behind it?

src/platform_impl/macos/window_delegate.rs
```diff
@@ -286,7 +278,6 @@ extern "C" fn window_will_close(this: &Object, _: Sel, _: id) {
 extern "C" fn window_did_resize(this: &Object, _: Sel, _: id) {
     trace_scope!("windowDidResize:");
     with_state(this, |state| {
-        state.emit_resize_event();
         state.emit_move_event();
```
kchibisov, line 281: Could you add a comment that resize size is reported in `frameDidResize`.

kchibisov (changes requested):

src/platform_impl/macos/view.rs
sshock: done

src/platform_impl/macos/window_delegate.rs
sshock: done

kchibisov (approved):

PR #2243, opened 2022-04-06
Add new `IME` event for desktop platforms

### LINKED ISSUE #2174: keyboard pops up always on phosh and plasma mobile
I've been playing around with just running random winit apps on phosh and plasma mobile on the pine phone, and the keyboard appears to pop up always (i have no text field elements, etc., it's just a graphics canvas basically)

So these systems are running wayland, and I was told in a channel to grep for text_input, e.g.:

```
1234567891011[ 663368.247] wl_registry@2.global(18, "zwp_text_input_manager_v3", 1)
[ 663368.386]  -> wl_registry@2.bind(18, "zwp_text_input_manager_v3", 1, new id [unknown]@7)
[ 663409.541]  -> zwp_text_input_manager_v3@7.get_text_input(new id zwp_text_input_v3@14, wl_seat@12)
[ 663574.530] wl_registry@2.global(18, "zwp_text_input_manager_v3", 1)
[ 663598.612] wl_registry@2.global(18, "zwp_text_input_manager_v3", 1)
[ 663604.224] wl_registry@2.global(18, "zwp_text_input_manager_v3", 1)
[ 663609.518] wl_registry@2.global(18, "zwp_text_input_manager_v3", 1)
[ 663615.969] wl_registry@36.global(18, "zwp_text_input_manager_v3", 1)
[ 663617.496] wl_registry@37.global(18, "zwp_text_input_manager_v3", 1)
[ 663635.486] wl_registry@39.global(18, "zwp_text_input_manager_v3", 1)
[ 663637.125] wl_registry@41.global(18, "zwp_text_input_manager_v3", 1)
```

where gedit, for example, `WAYLAND_DEBUG=1 gedit 2>&1 | grep text_input` has:

```
[2904291.467] wl_registry@2.global(18, "zwp_text_input_manager_v3", 1)
[2905408.074] wl_registry@27.global(18, "zwp_text_input_manager_v3", 1)
[2905408.092]  -> wl_registry@27.bind(18, "zwp_text_input_manager_v3", 1, new id [unknown]@41)
[2905408.115]  -> zwp_text_input_manager_v3@41.get_text_input(new id zwp_text_input_v3@42, wl_seat@20)
[2958948.053] zwp_text_input_v3@42.enter(wl_surface@28)
[2958948.686]  -> zwp_text_input_v3@42.enable()
[2958948.766]  -> zwp_text_input_v3@42.set_surrounding_text("", 0, 0)
[2958948.805]  -> zwp_text_input_v3@42.set_text_change_cause(0)
[2958948.866]  -> zwp_text_input_v3@42.set_content_type(0, 0)
[2958948.898]  -> zwp_text_input_v3@42.set_cursor_rectangle(24, 47, 0, 18)
[2958948.935]  -> zwp_text_input_v3@42.commit()
```

I don't really know anything about wayland, but i'm wondering if this is a winit related issue, since it popping up text seems like something the application itself specifies via some wayland api?

kchibisov:
This commit brings new IME event to account for preedit state of input
method, also adding `Window::set_ime_allowed` to toggle IME input on
the particular window.

This commit implements API as designed in #1497 for Wayland, X11, and
macOS.

Fixes #2174.

Co-authored-by: Artur Kovacs <kovacs.artur.barnabas@gmail.com>
Co-authored-by: Murarth <murarth@gmail.com>
Co-authored-by: Yusuke Kominami <yukke.konan@gmail.com>

--

I've finished implementing X11/Wayland bits. All functionality is supported across X11, Wayland, and macOS. The downstream implementation of new event handling is present in https://github.com/alacritty/alacritty/pull/5790.

The PR handles all the platforms currently having IME input in winit, so there're no blocks left.

I've tested all the implementations with mentioned client and they worked as intedend from my experience, though I haven't payed much attention to macOS one, but at least it works.

cc @chrisduerr @garasubo


kchibisov:
Now it has complete support across desktop platforms, so reviews are welcome.

src/platform_impl/linux/wayland/window/shim.rs
```diff
@@ -304,6 +311,27 @@ impl WindowHandle {
...
+        for text_input in self.text_inputs.iter() {
+            text_input.set_input_allowed(allowed);
+
+            let event = if allowed {
+                WindowEvent::IME(IME::Enabled)
+            } else {
+                WindowEvent::IME(IME::Disabled)
+            };
+
+            event_sink.push_window_event(event, window_id);
```
chrisduerr, line 331: Why is an event generated for every single IME? Shouldn't the event only be generated once?

chrisduerr (changes requested):
Only looked at the Linux implementations, but seems good to me mostly.

src/platform_impl/linux/wayland/window/shim.rs
kchibisov: How so? If text input loses focus on the surface you send disable, since you don't have IME at all, meaning that IME is disabled, and when you get focus back you send enable?

Or are you talking in the context of multiple text-inputs?

chrisduerr: I'm talking in the context of multiple `text-inputs`. That's opaque to the client, right? But this still duplicates the event for all of the inputs. It doesn't seem useful to get the event `X` times? The `text_input` is never used for this, so it's not like you're passing the specific input?

kchibisov: Ah, no, multiple text inputs are kind of broken right now, since I think protocol get clarified wrt that? At least I don't remember a lot of stuff from it, but the thing is that you can't have multiple text inputs over the same surface from what I can see for the same seat.

I'll try updating it according to the spec on Wayland side.

chrisduerr: I mean regardless of if it works properly or not, the code doesn't make any sense? Saying "this is broken anyway" seems like a weak excuse. I'd assume you have it as a `Vec` because it might work in the future? So why not write it properly? Just gotta move things down a couple lines.

kchibisov: For whatever reason I've thought that you're talking about completely different place in handler, but it was in window/shim.rs. I've moved the lines, should be good.

I've also checked the `text-input` protocol, and what winit is doing make sense, since it has one text-input per seat, but the protocol talking about multiple text inputs per seat, so we're good.

kchibisov:
Also, what do you think about naming it `Ime` event instead of `IME`? I think the former is more rusty?

There's also an issue wrt setting surrounding text, but I think it could be done separately, since the original PR is more about event and not about IME helpers, since those helpers are not really required.

maroider:
> Also, what do you think about naming it Ime event instead of IME? I think the former is more rusty?

`Ime` seems more idiomatic than `IME`, yes.

chrisduerr:
Pretty sure clippy even lints against `IME`?

kchibisov:
I've done all renaming, if there won't be any issues raised until 23.04.2022 I'll merge it that day.

src/event.rs
```diff
@@ -627,6 +634,28 @@ pub struct KeyboardInput {
...
+    /// Notifies when the IME was enabled.
+    Enabled,
+
+    /// Notifies when a new composing text should be set at the cursor position.
+    ///
+    /// The value represents a pair of the preedit string and the cursor begin position and end
+    /// position. When both indices are `None`, the cursor should be hidden.
+    ///
+    /// The cursor position is byte-wise indexed.
+    Preedit(String, Option<usize>, Option<usize>),
```
madsmtm, line 650: I think this should be `Preedit(String, Option<(usize, usize)>)`, since there is never a case (as far as I can see) where you know the beginning or the end, but not both.

(Even Wayland's [`Event::PreeditString`](https://docs.rs/wayland-protocols/0.29.4/wayland_protocols/unstable/text_input/v3/client/zwp_text_input_v3/enum.Event.html#variant.PreeditString) seem to indicate that both values are usually positive, except in the hidden case where they're both negative, though I'm not certain here at all!)

src/event.rs
```diff
@@ -627,6 +634,28 @@ pub struct KeyboardInput {
     pub modifiers: ModifiersState,
 }
 
+/// Describes an event from input method.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum Ime {
+    /// Notifies when the IME was enabled.
+    Enabled,
```
madsmtm, line 642: I still don't get why `Ime::Enabled` and `Ime::Disabled` exist, and when the user can expect to receive them - are they just so you know when you don't need to call `Window::set_ime_position`, as an optimization? If so, I feel like there might be a better way to express this (or maybe they're not really warranted).

In any case I'd like for them to be more documented.

src/event.rs
```diff
@@ -627,6 +634,28 @@ pub struct KeyboardInput {
...
+    /// Notifies when a new composing text should be set at the cursor position.
+    ///
+    /// The value represents a pair of the preedit string and the cursor begin position and end
+    /// position. When both indices are `None`, the cursor should be hidden.
+    ///
+    /// The cursor position is byte-wise indexed.
+    Preedit(String, Option<usize>, Option<usize>),
+
+    /// Notifies when text should be inserted into the editor widget.
+    Commit(String),
```
madsmtm, line 653: I suspect it would make sense to merge this into `WindowEvent::ReceivedCharacter`, so clients only need to handle one event when they want to actually insert text into their buffer.

But this is probably a larger effort for lesser benefit, so I won't block on that.

madsmtm (changes requested):
I created https://github.com/rust-windowing/winit/pull/2258 for some work on documentation that I'd like in this.

MacOS implementation looks good enough for me for now, I'll probably clean some things up after we fully revamp the keyboard API, but that's not a blocker for this.

ArturKovacs:
First of all thanks for all your work guys! This is definitely one of the bigger improvements in winit. Although I found a bug.

I don't get Commit events on X11 with the Pinyin input. (Ubuntu 20.04)

I called `set_ime_allowed(true)` before entering the event loop

I type type:
`nihao`

The input selection box shows up correctly and I get the following events
```
ime: Preedit("", None, None)
ime: Preedit("", Some(0), Some(0))
ime: Preedit("", Some(0), Some(0))
ime: Commit("")
ime: Preedit("", None, None)
ime: Preedit("你", Some(3), Some(3))
ime: Preedit("你", Some(3), Some(3))
ime: Preedit("你好", Some(6), Some(6))
ime: Preedit("你哈", Some(6), Some(6))
ime: Preedit("你好", Some(6), Some(6))
```
Then I press <kbd>1</kbd> and the input selection box disappears but I don't get a `Commit` event. In only get the key release

```
regular keypress: KeyboardInput { scancode: 2, state: Released, virtual_keycode: Some(Key1), modifiers: (empty) }
```

The same happens if I press <kbd>Return</kbd> or <kbd>Space</kbd> instead of <kbd>1</kbd> (except of course I get the release event for the released key at the end)

src/event.rs
kchibisov: Yeah, I'll change that.

src/event.rs
kchibisov: Yeah, and the user can disable IME on the fly. Also it's some kind of feedback to `set_ime_allowed`.

src/event.rs
kchibisov: No, it won't. The point is that you should know that you've got a commit from the IME, so you can remove preedit area and insert text at once into the widget. I'm pretty sure everything will change with new keyboard API, but the IME text won't be attached to any keyboard event anyway, so it should be delivered via separate event, like commit, and it's more intuitive for the end user, since that's how every IME API works out there.

kchibisov:
In case someone has documentation suggestions, feel free to add them in #2258.

kchibisov:
@ArturKovacs would you mind looking into `preedit` callbacks in X11 impl and events you're getting?

I can't repro that with default gnome + Pinyin on fedora 35. It works as expected for me.

It shouldn't be that hard to figure it out. If you don't have time to look into it, you should likely tell me what's your setup is so I can try to repro, though it'll take time that way.

src/event.rs
kchibisov: Done.

ArturKovacs:
> I can't repro that with default gnome + Pinyin on fedora 35. It works as expected for me.

That's odd. I added `trace!` macros to the preedeit callback functions and I get "start" and "draw", but I don't get "done".

Maybe we should continue this conversation on Matrix, I could share my screen in a call if you want.

kchibisov:
@ArturKovacs I'll try to look around and see what we can do here. Maybe XIM wants to insert text at different point other than in Done callback, but at least on fedora default IME setup everything works as expected...

kchibisov:
@ArturKovacs would you mind trying latest commit on X11 one more time, it should work now from what I can see.

kchibisov:
Also, should we do anything wrt keyboard input during IME input on X11? Should we not send it at all? What do you think @ArturKovacs ?

ArturKovacs:
It works! You're a genius!

Regarding the keyboard input during preedit, I think we agreed that we never send that.

kchibisov:
>Regarding the keyboard input during preedit, I think we agreed that we never send that.

I've altered that as well, so it should work.

While working with preedit downstream, I've noticed that I'd like to have some sort of indication when preedit gets cleared, so when winit is sending `Preedit(String::new(), None)` event. Maybe we should indicate that preedit got cleared somehow, for example `Preedit(enum { Cleared, Text })` or `Preedit(Option<PreeditText>)` and send `None` when it gets cleared?

madsmtm:
You mean we could do the following:
```rust
enum Ime {
    // ...
    
    Preedit(String, usize, usize),
    PreeditCleared,
    
    // ...
}
```

With a translation of:
```rust
let new = match old {
    Preedit(string, None) => PreeditCleared,
    Preedit(string, Some((begin, end))) => Preedit(string, begin, end),
}
```

That is, `string` in `Preedit(string, None)` is always irrelevant?

kchibisov:
>That is, string in Preedit(string, None) is always irrelevant?

No, when that `string` is `empty`. So `Preedit("a", None)` is fine, but `Preedit("", None)` will translate into `PreeditCleared`.

madsmtm:
And `Preedit("", Some((0, 0))) -> PreeditCleared`?

Anyway, seems to me like we're adding a state that is representable in multiple ways - but then again, I'm fairly certain macOS doesn't really have the concept of `PreeditCleared`, so I probably just don't really understand the purpose of it.

kchibisov:
The purpouse to indicate that there's no preedit, so application can start processing normal keys input again? Certain IMEs are a bot fynamic here and forward regular keyboard input when preedit goes to empty string.

ArturKovacs:
I still don't understand why we would need such an event. Winit should never send keyboard input events during preedit according to what we agreed. In other words: the application should always process KeyboardInput events because they only arrive outside of preedit. Therefore there's no need to indicate that the preedit ended.

kchibisov:
Hm, I guess that's ok then. So besides upcoming changes to docs I don't see anything blocking it.

madsmtm (approved):
API and macOS LGTM 👍

kchibisov:
Given that we had approval from each platform(except X11 due to missing maintainer, but backend was tested and originally submitted from one of them), we should be fine.

Issues that arise will be addressed latter on.

PR #2250, opened 2022-04-09
Unify `with_app_id` and `with_class` methods

### LINKED ISSUE #1739: Unify `with_app_id` and `with_class` on Linux/BSD
Winit currently has two calls that do very similar things, `with_app_id` and `with_class`. Generally I think these are rarely used with different parameters and more commonly the code is just duplicated between both of them. As an abstraction library that shouldn't make downstream care too much about Wayland/X11, I think this gives winit the opportunity to unify these two methods to make things a bit easier to use.

They are of course not exactly the same and depending on the platform the number of arguments is different, but since they should generally overlap this can be resolved by thorough documentation.

Of course even when unified, people could still do x11/Wayland checks on their own if they really want to treat these methods differently.

So the main reason why one wouldn't want to do this is that these aren't commonly set to the same thing. I can only speak for myself and Alacritty where we do indeed set these to the same value, so I'm curious about others here. Are there any downstream users treating these as different things?

kchibisov:
Both APIs are used to set application name. This commit unifies the API
between Wayland and X11, so downstream applications can remove platform
specific code here.

Fixes #1739.

cc @chrisduerr

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On Wayland, fix `TouchPhase::Ended` always reporting the location of the first touch down, unless the compositor
   sent a cancel or frame event.
 - On iOS, send `RedrawEventsCleared` even if there are no redraw events, consistent with other platforms.
+- **Breaking:** Replaced `Window::with_app_id` and `Window::with_class` with `Window::with_name` on `WindowBuilderExtUnix`.
 
 # 0.26.1 (2022-01-05)
 
--- a/src/platform/unix.rs
+++ b/src/platform/unix.rs
@@ -21,7 +21,8 @@ use crate::dpi::Size;
 #[cfg(feature = "x11")]
 use crate::platform_impl::x11::{ffi::XVisualInfo, XConnection};
 use crate::platform_impl::{
-    Backend, EventLoopWindowTarget as LinuxEventLoopWindowTarget, Window as LinuxWindow,
+    ApplicationName, Backend, EventLoopWindowTarget as LinuxEventLoopWindowTarget,
+    Window as LinuxWindow,
 };
 
 // TODO: stupid hack so that glutin can do its work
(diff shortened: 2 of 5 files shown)


src/platform/unix.rs
```diff
@@ -270,21 +271,34 @@ impl WindowExtUnix for Window {
...
     #[cfg(feature = "x11")]
     fn with_x11_screen(self, screen_id: i32) -> Self;
 
-    /// Build window with `WM_CLASS` hint; defaults to the name of the binary. Only relevant on X11.
-    #[cfg(feature = "x11")]
-    fn with_class(self, class: String, instance: String) -> Self;
+    /// Build window with the given `general` and `instance` names.
+    ///
+    /// On Wayland, the `general` name sets an application ID, which should match the `.desktop`
+    /// file destributed with your program. The `instance` is `no-op` on it.
```
chrisduerr, line 281:
```suggestion
    /// file destributed with your program. The `instance` is a `no-op`.
```

chrisduerr (approved):
Other than the minor text refinement, this looks good to me.

CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On Wayland, fix polling during consecutive `EventLoop::run_return` invocations.
 - On Windows, fix race issue creating fullscreen windows with `WindowBuilder::with_fullscreen`
 - On Android, `virtual_keycode` for `KeyboardInput` events is now filled in where a suitable match is found.
+- **Breaking:** Replaced `Window::with_app_id` and `Window::with_class` with `Window::with_name`.
```
madsmtm, line 35: Use `WindowBuilderExtUnix` here to indicate that this change only affects those platforms.

madsmtm (changes requested):
Otherwise approve

madsmtm (approved):

src/platform/unix.rs
```diff
@@ -270,21 +271,34 @@ impl WindowExtUnix for Window {
...
+    ///
+    /// On Wayland, the `general` name sets an application ID, which should match the `.desktop`
+    /// file destributed with your program. The `instance` is a `no-op`.
+    ///
+    /// On X11, the `general` sets general class of `WM_CLASS(STRING)`, while `instance` set the
+    /// instance part of it. The resulted property looks like `WM_CLASS(STRING) = "general", "instance"`.
+    ///
+    /// For details about application ID conventions, see the
+    /// [Desktop Entry Spec](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#desktop-file-id)
+    fn with_name<T: Into<String>>(self, general: T, instance: T) -> Self;
```
maroider, line 288: I'm not a fan of using `T: Into<String>` where `T` appears in multiple parameters.

chrisduerr: Just `impl Into<String>` should be superior tbh.

kchibisov: Yeah, that seems better.

PR #2252, opened 2022-04-09
Unify behavior of `resizable` across platforms

### LINKED ISSUE #2242: Behaviour of resizable is different across platforms
When setting `window.set_resizable(false)` the window is still resizable via `window.set_inner_size()` on macOS and Windows, but not on Linux. I would expect that `window.set_resizable(false)` just disables the manual handles for the mouse on all platforms but keep it possible to change the size from the application itself.

kchibisov:
This makes X11 and Wayland follow Windows and macOS, so the size of the
window could be set even though it has resizable attribute set to false.

Fixes #2242.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ And please only add new entries to the top of this list, right below the `# Unre
   sent a cancel or frame event.
 - On iOS, send `RedrawEventsCleared` even if there are no redraw events, consistent with other platforms.
 - **Breaking:** Replaced `Window::with_app_id` and `Window::with_class` with `Window::with_name` on `WindowBuilderExtUnix`.
+- On Wayland and X11, fix window not resizing with `Window::set_inner_size` after calling `Window:set_resizable(false)`.
 
 # 0.26.1 (2022-01-05)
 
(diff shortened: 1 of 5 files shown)


kchibisov:
@sourcebox would you mind testing this PR and tell whether it resolves your issue? I think it should resize the window now with `set_inner_size`.

sourcebox:
@kchibisov I just tested your PR and it works as expected so far. I'm typically not using `winit` directly but via `egui`. For testing I just modified the `min_max_size` example. Here's the full code:

```rust
use winit::{
    dpi::LogicalSize,
    event::{Event, WindowEvent},
    window::WindowBuilder,
};

fn main() {

    window.set_resizable(false);
    window.set_inner_size(LogicalSize::new(200.0, 200.0));

        println!("{:?}", event);

        match event {
            Event::WindowEvent {
                ..
            } => *control_flow = ControlFlow::Exit,
            Event::WindowEvent {
                event: WindowEvent::KeyboardInput { .. },
                ..
            } => {
                window.set_inner_size(LogicalSize::new(550.0, 550.0));
            }
            _ => (),
        }
    });
}
```
With your patch, `window.set_inner_size()` changes the dimensions accordingly, both on initialization and within the event loop when pressing a key.

PR #2255, opened 2022-04-11
Fix touch position being reported wrongly on `TouchPhase::Ended`

### LINKED ISSUE #1996: [Wayland] touch_points never removed
In `platform_impl/linux/wayland/seat/touch/handlers.rs`, elements of `touch_points` are not removed even when a touch event ends. I think in principle this is a memory leak, but it also means that if a touch is started and ended without any movement in between, the `location` associated to the `Touch` with `TouchPhase::Ended`, is the wrong location!

I think (but have not tested) that this is repaired simply by removing any element with matching `id` after a `TouchEvent::Up`.

tronical:
When all the receive from the compositor is `TouchEvent::Down` and
`TouchEvent::Up` for the same id, we would record the touch position in
the touch_points vector the first time. On `TouchEvent::Up` we'd find it
and report `TouchPhase::Ended` with that location. The next time we
receive `TouchEvent::Down` for the same id, we'd however unconditionally
append a new `TouchPoint` to `inner.touch_points`, with the new
position. On release however we'd find the earlier point and report its
location, which basically means that `TouchPhase::Ended` always and
forever reported the location of the very first touch down event.

Instead, this patch updates an existing touch point location with the
same id on `TouchDown`.

Fixes #1996

### DIFF
--- a/src/platform_impl/linux/wayland/seat/touch/handlers.rs
+++ b/src/platform_impl/linux/wayland/seat/touch/handlers.rs
@@ -44,6 +44,8 @@ pub(super) fn handle_touch(
                 window_id,
             );
 
+            // For `TouchEvent::Up` we don't receive a position, so we're tracking active
+            // touch points. Update either a known touch id or register a new one.
             if let Some(i) = inner.touch_points.iter().position(|p| p.id == id) {
                 inner.touch_points[i].position = position;
             } else {


tronical:
Yes. I ran into this issue with a touch device and wayland and verified that this patch fixes it.

src/platform_impl/linux/wayland/seat/touch/handlers.rs
```diff
@@ -44,9 +44,13 @@ pub(super) fn handle_touch(
                 window_id,
             );
 
-            inner
-                .touch_points
-                .push(TouchPoint::new(surface, position, id));
+            if let Some(i) = inner.touch_points.iter().position(|p| p.id == id) {
```
kchibisov, line 47: Would you mind adding a comment why it's done that way?

kchibisov (changes requested):
Should be good then, though, would be nice to have comment in code wrt that.

kchibisov (approved):

kchibisov:
Thanks.

PR #2257, opened 2022-04-13
detect left and right option keys as modifiers on macos implementation

erenoku:
The left and right alt keys modifiers were previously detected as the same modifier. It created some problems on macos mainly on terminal emulators like alacritty. https://github.com/alacritty/alacritty/issues/62 would be possible to fix with this PR.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -1018,10 +1018,10 @@ bitflags! {
         /// The "alt" key.
         const ALT = 0b100 << 6;
         /// the left "alt" key.
-        #[cfg(target_os = "macos")]
+        /// IMPORTANT: This flag will only be set on macos
         const LALT = 0b010 << 6;
         /// the right "alt" key.
-        #[cfg(target_os = "macos")]
+        /// IMPORTANT: This flag will only be set on macos
         const RALT = 0b001 << 6;
         /// This is the "windows" key on PC and "command" key on Mac.
         const LOGO = 0b100 << 9;
--- a/src/event.rs
+++ b/src/event.rs
@@ -1007,8 +1007,6 @@ bitflags! {
     /// Each flag represents a modifier and is set if this modifier is active.
     #[derive(Default)]
     pub struct ModifiersState: u32 {
-        // left and right modifiers are currently commented out, but we should be able to support
-        // them in a future release
         /// The "shift" key.
         const SHIFT = 0b100;
         // const LSHIFT = 0b010;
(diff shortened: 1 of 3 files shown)


src/event.rs
```diff
@@ -1019,8 +1017,12 @@ bitflags! {
...
         /// The "alt" key.
         const ALT = 0b100 << 6;
-        // const LALT = 0b010 << 6;
-        // const RALT = 0b001 << 6;
+        /// the left "alt" key.
+        #[cfg(target_os = "macos")]
+        const LALT = 0b010 << 6;
+        /// the right "alt" key.
+        #[cfg(target_os = "macos")]
+        const RALT = 0b001 << 6;
```
madsmtm, line 1025: We shouldn't change `ModifiersState` depending on the target (since that makes it quite difficult for downstream to write platform-independent code). Better to document where `LALT` and `RALT` are supported.

```suggestion
        /// the left "alt" key.
        const LALT = 0b010 << 6;
        /// the right "alt" key.
        const RALT = 0b001 << 6;
```

src/platform_impl/macos/event.rs
```diff
@@ -241,8 +241,16 @@ pub fn check_function_keys(string: &str) -> Option<VirtualKeyCode> {
     None
 }
 
+const NS_LEFT_ALTERNATE_KEY_MASK: u64 = 0x000020 | NSEventModifierFlags::NSAlternateKeyMask.bits();
+const NS_RIGHT_ALTERNATE_KEY_MASK: u64 = 0x000040 | NSEventModifierFlags::NSAlternateKeyMask.bits();
```
madsmtm, line 245: Do you have some documentation for these magic values? Whether you do or not, a bit of documentation on them would be nice!

madsmtm (changes requested):
Interesting!

I'm quite out of the loop on your linked alacritty issue, how does this compare to https://github.com/rust-windowing/winit/pull/1449?

erenoku:
I think https://github.com/alacritty/alacritty/issues/62 and https://github.com/rust-windowing/winit/pull/1449 are pretty similar on purpose. They both aim to make it possible to set option as alt.

The reason why differentiating between left and right alt modifier keys are necessary, is that on non English keyboards important character such as ~ or {[]} are typed with a combination of option keys. It is discussed later in the issue I linked https://github.com/alacritty/alacritty/issues/62#issuecomment-422871585. So disabling the standard usage (as option modifier) on both keys will make it impossible for non English keyboard users to type these characters. Other terminal emulators such as Kitty and ITerm have an option to choose which option key to detect as alt.

src/platform_impl/macos/event.rs
erenoku: I couldn't find any official documentation from apple on how these key masks should be formulated but I found [this discussion which includes some useful formulas](https://community.folivora.ai/t/could-you-provide-a-formula-for-determining-keyboard-shortcuts-values/3606) and these key mask values are used in many projects such as SFML [permalink of the code snippet](https://github.com/fishbupt/sfml/blob/86694bcf5850045ff8bc9b6e2eb3874ee40cacec/src/SFML/Window/OSX/SFKeyboardModifiersHelper.mm#L43-L44) and [iTerm](https://github.com/gnachman/iTerm2/blob/14148ffa9510e13a52043e64b552d7bf6c44307c/sources/iTermKeyboardHandler.h#L15-L16).

src/event.rs
erenoku: Fixed

PR #2258, opened 2022-04-17
Improve `IME` documentation

madsmtm:
See #2243.

Please correct me on the specifics here, I may very well be wrong on some of this.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -275,6 +275,8 @@ pub enum WindowEvent<'a> {
 
     /// An event from input method.
     ///
+    /// **Note :** You have to explicitly enable this event using [`Window::set_ime_allowed`].
+    ///
     /// Platform-specific behavior:
     /// - **iOS / Android / Web :** Unsupported.
     Ime(Ime),
@@ -672,11 +674,14 @@ pub struct KeyboardInput {
 /// Ime::Commit("啊不")
 /// ```
 ///
-/// **Note: You have to explicitly enable this event using [`Window::set_ime_allowed`].**
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub enum Ime {
     /// Notifies when the IME was enabled.
+    ///
+    /// After getting this event you could receive [`Preedit`](Self::Preedit) and
+    /// [`Commit`](Self::Commit) events. You should also start performing IME related requests
+    /// like [`Window::set_ime_position`].
     Enabled,
 
     /// Notifies when a new composing text should be set at the cursor position.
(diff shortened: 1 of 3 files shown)


kchibisov (approved):

src/event.rs
```diff
@@ -634,7 +637,42 @@ pub struct KeyboardInput {
     pub modifiers: ModifiersState,
 }
 
-/// Describes an event from input method.
+/// Describes [input method](https://en.wikipedia.org/wiki/Input_method) events.
+///
+/// This is also called a "composition event".
+///
+/// Most keypresses using a latin-like keyboard layout simply generate an [`WindowEvent::ReceivedCharacter`].
```
msiglreith, line 644:
```suggestion
/// Most keypresses using a latin-like keyboard layout simply generate a [`WindowEvent::ReceivedCharacter`].
```

src/event.rs
```diff
@@ -634,7 +637,42 @@ pub struct KeyboardInput {
...
+/// // Press "`" key
+/// Ime::Preedit("`", Some(0), Some(0))
+/// // Press "E" key
+/// Ime::Commit("é")
+/// ```
+///
+/// Additionally, certain input devices are configured to display a candidate box that allow the user to select the
+/// desired character interactively. (To properly position this box, you must use [`Window::set_ime_position`].)
+///
+/// An example of a keyboard layout which uses candiate boxes is pinyin. On a latin keybaord the following event
```
msiglreith, line 660:
```suggestion
/// An example of a keyboard layout which uses candidate boxes is pinyin. On a latin keybaord the following event
```

src/event.rs
```diff
@@ -634,7 +637,42 @@ pub struct KeyboardInput {
...
+///
+/// An example of a keyboard layout which uses candidate boxes is pinyin. On a latin keybaord the following event
+/// sequence could be obtained:
+/// ```ignore
+/// // Press "A" key
+/// Ime::Preedit("a", Some(1), Some(1))
+/// // Press "B" key
+/// Ime::Preedit("a b", Some(3), Some(3))
+/// // Press left arrow key
+/// Ime::Preedit("a b", Some(1), Some(1))
```
madsmtm, line 668: I should note: This part is technically not true, currently our macOS implementation gives `Ime::Preedit("a b", Some(3), Some(3))`. But it _should_ give `Ime::Preedit("a b", Some(1), Some(1))`.

kchibisov: I can ensure you that it'll work like that on Wayland at least, and that's how it should be on macOS as well? If it doesn't work like that it's likely a macOS bug, though, not really a blocking one...

kchibisov:
@madsmtm I've pushed a commit adding more docs about `Enabled`/`Disabled`.

src/event.rs
madsmtm: Yeah, it's a bug, should be possible to fix just has to do some UTF-16 to UTF-8 encoding that I didn't really see how to do easily.

PR #2263, opened 2022-04-24
wayland: Replace fallback CSD with proper one

PolyMeilex:
Hi!
I wrote a simple CSD implementation for Smithay client toolkit, and I think it could be a good default for winit.
Here is winit running with those decorations:
|   |   |
|---|---|
|active|inactive|

Let me also answer questions that will most likely come up:
- `Why not libdecor?`
   - Using libdecor from Rust Wayland ecosystem is nowhere near completion, and in my opinion it will never be. And besides some initial tests with calling it from Rust, no one is actively working on it atm.
   - If I'm wrong and libdecor-rs will be a thing, then we can just drop my impl easily, so it's a win-win scenario.
- `Why  does it look like GTK/Adwaita?`
   - Mutter/Gnome is probably the only big compositor that does not support SSD, so I think it's a good idea to make it fit there, other DEs will render their own SSD anyway, so users of those will be happy anyway.
   - I think it just looks neutral, it's simple and minimal, so I doubt anyone will find this kind of design offensive.
- `Why does winit need a good default?`
  - Not all developers that use winit are well versed in Linux world, let alone know what Wayland is. So I believe that we have to have a good default, so we get a decent look, even if the developer of an app haven't spent time understanding what Wayland is and why it needs CSD.

> DISCLAIMER: Although I'm part of Smithay, this PR is just my private endavor, so nothing I say here should be taken as an opinion of Smithay as a whole, just as a random guy on the internet trying to make winit prettier.

closes: #1967

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,7 +39,7 @@ And please only add new entries to the top of this list, right below the `# Unre
   sent a cancel or frame event.
 - On iOS, send `RedrawEventsCleared` even if there are no redraw events, consistent with other platforms.
 - **Breaking:** Replaced `Window::with_app_id` and `Window::with_class` with `Window::with_name` on `WindowBuilderExtUnix`.
-- On Wayland, fallback CSD was replaced with proper one;
+- On Wayland, fallback CSD was replaced with proper one:
   - `WindowBuilderExtUnix::with_wayland_csd_theme` to set color theme in builder.
   - `WindowExtUnix::wayland_set_csd_theme` to set color theme when creating a window.
   - `WINIT_WAYLAND_CSD_THEME` env variable was added, it can be used to set "dark"/"light" theme in apps that don't expose theme setting.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ And please only add new entries to the top of this list, right below the `# Unre
   sent a cancel or frame event.
 - On iOS, send `RedrawEventsCleared` even if there are no redraw events, consistent with other platforms.
 - **Breaking:** Replaced `Window::with_app_id` and `Window::with_class` with `Window::with_name` on `WindowBuilderExtUnix`.
+- On Wayland, fallback CSD was replaced with proper one.
 - On Wayland and X11, fix window not resizing with `Window::set_inner_size` after calling `Window:set_resizable(false)`.
 - On Windows, fix wrong fullscreen monitors being recognized when handling WM_WINDOWPOSCHANGING messages
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,12 +39,12 @@ And please only add new entries to the top of this list, right below the `# Unre
   sent a cancel or frame event.
 - On iOS, send `RedrawEventsCleared` even if there are no redraw events, consistent with other platforms.
 - **Breaking:** Replaced `Window::with_app_id` and `Window::with_class` with `Window::with_name` on `WindowBuilderExtUnix`.
-- On Wayland, fallback CSD was replaced with proper one.
-  - `WindowBuilderExtUnix::with_wayland_csd_theme` used to set color theme in builder.
-  - `WindowExtUnix::wayland_set_csd_theme` used to change color theme on the fly.
+- On Wayland, fallback CSD was replaced with proper one;
+  - `WindowBuilderExtUnix::with_wayland_csd_theme` to set color theme in builder.
+  - `WindowExtUnix::wayland_set_csd_theme` to set color theme when creating a window.
   - `WINIT_WAYLAND_CSD_THEME` env variable was added, it can be used to set "dark"/"light" theme in apps that don't expose theme setting.
-  - `wayland-csd-adwaita` feature that enables proper CSD.
-  - `wayland-csd-title` feature that enables title rendering, uses FreeType system library.
+  - `wayland-csd-adwaita` feature that enables proper CSD with title rendering using FreeType system library.
+  - `wayland-csd-adwaita-notitle` feature that enables CSD but without title rendering.
 - On Wayland and X11, fix window not resizing with `Window::set_inner_size` after calling `Window:set_resizable(false)`.
 - On Windows, fix wrong fullscreen monitors being recognized when handling WM_WINDOWPOSCHANGING messages
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On Wayland, fallback CSD was replaced with proper one.
   - `WindowBuilderExtUnix::with_wayland_csd_theme` used to set color theme in builder.
   - `WindowExtUnix::wayland_set_csd_theme` used to change color theme on the fly.
+  - `WINIT_WAYLAND_CSD_THEME` env variable was added, it can be used to set "dark"/"light" theme in apps that don't expose theme setting.
   - `wayland-csd-adwaita` feature that enables proper CSD.
   - `wayland-csd-title` feature that enables title rendering, uses FreeType system library.
 - On Wayland and X11, fix window not resizing with `Window::set_inner_size` after calling `Window:set_resizable(false)`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,10 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On iOS, send `RedrawEventsCleared` even if there are no redraw events, consistent with other platforms.
 - **Breaking:** Replaced `Window::with_app_id` and `Window::with_class` with `Window::with_name` on `WindowBuilderExtUnix`.
 - On Wayland, fallback CSD was replaced with proper one.
+  - `WindowBuilderExtUnix::with_wayland_csd_theme` used to set color theme in builder.
+  - `WindowExtUnix::wayland_set_csd_theme` used to change color theme on the fly.
+  - `wayland-csd-adwaita` feature that enables proper CSD.
+  - `wayland-csd-title` feature that enables title rendering, uses FreeType system library.
 - On Wayland and X11, fix window not resizing with `Window::set_inner_size` after calling `Window:set_resizable(false)`.
 - On Windows, fix wrong fullscreen monitors being recognized when handling WM_WINDOWPOSCHANGING messages
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ targets = ["i686-pc-windows-msvc", "x86_64-pc-windows-msvc", "i686-unknown-linux
 [features]
 default = ["x11", "wayland", "wayland-dlopen"]
 x11 = ["x11-dl", "mio", "percent-encoding", "parking_lot"]
-wayland = ["wayland-client", "wayland-protocols", "sctk"]
+wayland = ["wayland-client", "wayland-protocols", "sctk", "sctk-adwaita"]
 wayland-dlopen = ["sctk/dlopen", "wayland-client/dlopen"]
 
 [dependencies]
@@ -90,6 +90,7 @@ features = [
 wayland-client = { version = "0.29", default_features = false,  features = ["use_system_lib"], optional = true }
 wayland-protocols = { version = "0.29", features = [ "staging_protocols"], optional = true }
 sctk = { package = "smithay-client-toolkit", version = "0.15.4", default_features = false, features = ["calloop"],  optional = true }
+sctk-adwaita = { version = "0.3", optional = true }
 mio = { version = "0.8", features = ["os-ext"], optional = true }
 x11-dl = { version = "2.18.5", optional = true }
 percent-encoding = { version = "2.0", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ x11 = ["x11-dl", "mio", "percent-encoding", "parking_lot"]
 wayland = ["wayland-client", "wayland-protocols", "sctk"]
 wayland-dlopen = ["sctk/dlopen", "wayland-client/dlopen"]
 wayland-csd-adwaita = ["sctk-adwaita"]
+wayland-csd-title = ["sctk-adwaita/title"]
 
 [dependencies]
 instant = { version = "0.1", features = ["wasm-bindgen"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,10 +17,11 @@ default-target = "x86_64-unknown-linux-gnu"
 targets = ["i686-pc-windows-msvc", "x86_64-pc-windows-msvc", "i686-unknown-linux-gnu", "x86_64-unknown-linux-gnu", "x86_64-apple-darwin", "wasm32-unknown-unknown"]
 
 [features]
-default = ["x11", "wayland", "wayland-dlopen"]
+default = ["x11", "wayland", "wayland-dlopen", "wayland-csd-adwaita"]
 x11 = ["x11-dl", "mio", "percent-encoding", "parking_lot"]
-wayland = ["wayland-client", "wayland-protocols", "sctk", "sctk-adwaita"]
+wayland = ["wayland-client", "wayland-protocols", "sctk"]
 wayland-dlopen = ["sctk/dlopen", "wayland-client/dlopen"]
+wayland-csd-adwaita = ["sctk-adwaita"]
 
 [dependencies]
 instant = { version = "0.1", features = ["wasm-bindgen"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,8 +21,8 @@ default = ["x11", "wayland", "wayland-dlopen", "wayland-csd-adwaita"]
 x11 = ["x11-dl", "mio", "percent-encoding", "parking_lot"]
 wayland = ["wayland-client", "wayland-protocols", "sctk"]
 wayland-dlopen = ["sctk/dlopen", "wayland-client/dlopen"]
-wayland-csd-adwaita = ["sctk-adwaita"]
-wayland-csd-title = ["sctk-adwaita/title"]
+wayland-csd-adwaita = ["sctk-adwaita", "sctk-adwaita/title"]
+wayland-csd-adwaita-notitle = ["sctk-adwaita"]
 
 [dependencies]
 instant = { version = "0.1", features = ["wasm-bindgen"] }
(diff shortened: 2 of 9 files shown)


kchibisov (changes requested):
I'm not that against the idea behind providing some good client side decorations, however as war as I can tell they're always white? Would be nice to have a dark ones, but that's not really a blocker.

Also, do they support text rendering or something like that? Since that should be supported if we're going to outsource it for some other crate.

PolyMeilex:
> Would be nice to have a dark ones

Dark mode is definitely planned, but my problem is that I don't know how to expose it, if this was just my app I would simply get a dark mode preference from dbus desktop portal, but that would mean bringing a full dbus library to winit, I don't think that would be good for compile times.
So should we maybe just expose dark mode get/set methods to users of winit? (that would be probably Wayland only API)

> Also, do they support text rendering or something like that?

I intentionally omitted that, because I was concerned that compile times may discourage winit from merging this, but if we decide that a few more text related dependencies is fine, I will happily implement it.

kchibisov:
>I intentionally omitted that, because I was concerned that compile times may discourage winit from merging this, but if we decide that a few more text related dependencies is fine, I will happily implement it.

I'd probably put that under the feature flag. You may try `crossfont` or something more rusty.

PolyMeilex:
Title and Dark/Light/Custom themes were added.
Adding text rendering did not change compile times in any significant way, so it's always enabled. I guess I was needlessly worried about that.  
|   |   |
|---|---|
|image|image|

src/platform/unix.rs
```diff
@@ -32,6 +32,9 @@ pub use crate::platform_impl::x11;
 #[cfg(feature = "x11")]
 pub use crate::platform_impl::{x11::util::WindowType as XWindowType, XNotSupported};
 
+#[cfg(feature = "wayland")]
+pub use sctk_adwaita::FrameConfig;
```
kchibisov, line 36: I don't like pub reexports of dependencies.

I can see that you're using it just for `Dark` and `Light` theme variants, could you use https://github.com/rust-windowing/winit/blob/ce890c34551d9fb542f10dcb644d22d382e0c921/src/window.rs#L1100 instead? I think there're also methods to change theme on the `Window` already, they're just not exposed on Wayland/X11.

src/platform_impl/linux/wayland/window/mod.rs
```diff
@@ -105,7 +106,7 @@ impl Window {
         let theme_manager = event_loop_window_target.theme_manager.clone();
         let mut window = event_loop_window_target
             .env
-            .create_window::<FallbackFrame, _>(
```
kchibisov, line 108: I think it would make sense to keep both under the feature flag? Though, we should probably enable adwaita theme by default.

That's pretty straightforward with something like `type WinitFrame = FallbackFrame` and using appropriate type...

src/platform_impl/linux/wayland/window/shim.rs
```diff
@@ -53,6 +54,9 @@ pub enum WindowRequest {
     /// Request decorations change.
     Decorate(bool),
 
+    /// Request decorations change.
+    CsdConfig(FrameConfig),
```
kchibisov, line 58: I'd probably change it to `CsdThemeVariant(crate::window::Theme)`. (Note you should export `Theme`, full path is for clarity here).

kchibisov (changes requested):
That looks good, however I'd guess that you're using some hard coded font? Honestly I'd suggest to use https://github.com/alacritty/crossfont, since it basically a thin wrapper around system fontconfig and freetype. And then request `sans-serif` font.

Though, its API could be not ready, but you can ping wrt it in `#alacritty`.

The problem is that if we want similar text look, we should use freetype for like everything...

I'm not suggesting other font crates, since I haven't used them, they lack automatic fallback, and they are quite heavy compared to crossfont...

src/platform/unix.rs
PolyMeilex: This was done to allow custom themes, as Dark/Light are just two default ones, but nothing stops you from defining your own colors instead of calling dark()/light(), but sure I can just limit it to those two, it will simplify the API.

kchibisov: Maybe I just read it in the wrong way. We had in the past structure to initialize Wayland csd stuff, could probably bring it back in some form, I just don't like reexporting, since it could create unnecessary breaking changes downstream.

PolyMeilex:
It is now ported to crossfont, glyph positioning is kinda hacky because crossfont heavily assumes monospace everywhere, I will redo it once https://github.com/alacritty/crossfont/pull/40 is merged and released. Nonetheless it looks good enough even now.

Any ideas why CI is unhappy with crossfont/FreeType? Both libfreetype-dev and libfontconfig-dev are available.

kchibisov:
@PolyMeilex try adding pkg-config?

CHANGELOG.md
```diff
@@ -39,6 +39,7 @@ And please only add new entries to the top of this list, right below the `# Unre
   sent a cancel or frame event.
 - On iOS, send `RedrawEventsCleared` even if there are no redraw events, consistent with other platforms.
 - **Breaking:** Replaced `Window::with_app_id` and `Window::with_class` with `Window::with_name` on `WindowBuilderExtUnix`.
+- On Wayland, fallback CSD was replaced with proper one.
 - On Wayland and X11, fix window not resizing with `Window::set_inner_size` after calling `Window:set_resizable(false)`.
 - On Windows, fix wrong fullscreen monitors being recognized when handling WM_WINDOWPOSCHANGING messages
```
kchibisov, line 44: Should add a note wrt `with_wayland_csd_theme` and a `wayland-csd-adwaita` feature.

kchibisov (changes requested):
The code looks good, just some minor notes.

kchibisov:
@chrisduerr how you'd prefer handling of theme coloring downstream? Via winit env or exposing some sort of config/env variable for alacritty?

The only way I'm aware of picking the right theme variant is via dbus api, however it's not something I want/we should add.

With the current API it's more like downstream option, since there's a `with_theme` name.

Might remove the method entirely and handle everything via env variable?

kchibisov:
@PolyMeilex have you considered adding different button for maximized when the window is already maximized? I think that's what everyone is doing and could better for ux.

Should probably bump crossfont as well, since proportional font rendering isn't ideal right now.

src/platform_impl/linux/wayland/window/shim.rs
```diff
@@ -418,6 +423,23 @@ pub fn handle_window_requests(winit_state: &mut WinitState) {
                     let window_update = window_updates.get_mut(window_id).unwrap();
                     window_update.refresh_frame = true;
                 }
+                WindowRequest::CsdThemeVariant(theme) => {
```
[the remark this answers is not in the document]
PolyMeilex: Done

src/platform_impl/linux/wayland/window/mod.rs
```diff
@@ -139,6 +144,18 @@ impl Window {
...
             .map_err(|_| os_error!(OsError::WaylandMisc("failed to create window.")))?;
 
+        // Set CSD frame config
+        #[cfg(feature = "sctk-adwaita")]
+        if let Some(theme) = platform_attributes.csd_theme {
+            use sctk_adwaita::FrameConfig;
+
+            let config = match theme {
+                Theme::Light => FrameConfig::light(),
+                Theme::Dark => FrameConfig::dark(),
```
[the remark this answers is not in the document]
PolyMeilex: Done

CHANGELOG.md
PolyMeilex: Done

PolyMeilex:
> @PolyMeilex have you considered adding different button for maximized when the window is already maximized? I think that's what everyone is doing and could better for ux.

Done

> Should probably bump crossfont as well, since proportional font rendering isn't ideal right now.

Done, hacky positioning replaced with proper one that makes use of new crossfont, looks a lot better.

> Might remove the method entirely and handle everything via env variable?

Env variable does not allow for proper dynamic changing of the theme based on dbus, as client can connect to dbus and listen for theme changes and pass those down to winit.

chrisduerr:
> how you'd prefer handling of theme coloring downstream? Via winit env or exposing some sort of config/env variable for alacritty?

Good question. If we were talking about GTK, I'd strongly encourage using an environment variable so at least you know it's guaranteed to work. But with winit being a little more hands-on, I wouldn't be surprised if there was no way to do this for all winit apps (since users are usually unaware of their apps using winit anyway).

I feel like the best choice might be to just load based on variable by default and allow overriding on demand via winit's API?

src/platform_impl/linux/wayland/window/mod.rs
```diff
@@ -139,6 +144,16 @@ impl Window {
             )
             .map_err(|_| os_error!(OsError::WaylandMisc("failed to create window.")))?;
 
+        // Set CSD frame config
+        #[cfg(feature = "sctk-adwaita")]
+        if let Some(theme) = platform_attributes.csd_theme {
```
kchibisov, line 149: Would you mind checking env variable `WINIT_WAYLAND_CSD_THEME` for `dark` or `light` when `None` was found?

You should also add it to the changelog and to the new `with_wayland_csd_theme` method saying that it could be used in case downstream application doesn't want to expose settings for wayland client side decorations to the user on its own.

Should be a minor change.

Cargo.toml
```diff
@@ -17,10 +17,12 @@ default-target = "x86_64-unknown-linux-gnu"
 targets = ["i686-pc-windows-msvc", "x86_64-pc-windows-msvc", "i686-unknown-linux-gnu", "x86_64-unknown-linux-gnu", "x86_64-apple-darwin", "wasm32-unknown-unknown"]
 
 [features]
-default = ["x11", "wayland", "wayland-dlopen"]
+default = ["x11", "wayland", "wayland-dlopen", "wayland-csd-adwaita"]
 x11 = ["x11-dl", "mio", "percent-encoding", "parking_lot"]
 wayland = ["wayland-client", "wayland-protocols", "sctk"]
 wayland-dlopen = ["sctk/dlopen", "wayland-client/dlopen"]
+wayland-csd-adwaita = ["sctk-adwaita"]
+wayland-csd-title = ["sctk-adwaita/title"]
```
kchibisov, line 25: I think we should make `wayland-csd-adwaita` with title and have `wayland-csd-adwaita-notitle` without it. Make `wayland-csd-adwaita` enabled by default and add a proper note in the changelog that it's enabled by default now.

I think it makes sense to have title rendering for the end users by default.

Cargo.toml
```diff
@@ -90,6 +92,7 @@ features = [
 wayland-client = { version = "0.29", default_features = false,  features = ["use_system_lib"], optional = true }
 wayland-protocols = { version = "0.29", features = [ "staging_protocols"], optional = true }
 sctk = { package = "smithay-client-toolkit", version = "0.15.4", default_features = false, features = ["calloop"],  optional = true }
+sctk-adwaita = { version = "0.3", optional = true }
```
kchibisov, line 95: Could you explicitly specify patch version you've used, since proper text rendering isn't on `0.3`, but on some release.

kchibisov (changes requested):
Just some minor comments. Should be good to go after that.

src/platform_impl/linux/wayland/window/mod.rs
PolyMeilex: Done

Cargo.toml
PolyMeilex: Done

Cargo.toml
kchibisov: Bump.

src/platform_impl/linux/wayland/window/mod.rs
```diff
@@ -139,6 +146,28 @@ impl Window {
...
+                    "dark" => {
+                        window.set_frame_config(sctk_adwaita::FrameConfig::dark());
+                    }
+                    "light" => {
+                        window.set_frame_config(sctk_adwaita::FrameConfig::light());
+                    }
+                    _ => {}
+                }
+            }
+        }
```
kchibisov, line 169: Could you refactor all of that into something like

```rust
 #[cfg(feature = "sctk-adwaita")]
{
    let theme = platform_attributes.csd_theme.unwrap_or_else(|| {
        match std::env::var(WAYLAND_CSD_THEME_ENV_VAR).unwrap_or_default("") {
            "dark" => Theme::Dark,
            _ => Theme::Light
         }
    })
    let config = match theme {
         Theme::Light => sctk_adwaita::FrameConfig::light(),
         Theme::Dark => sctk_adwaita::FrameConfig::dark(),
    };
    window.set_frame_cofig(config)
}
```

I think it would be shorter and more readable.

CHANGELOG.md
```diff
@@ -39,6 +39,12 @@ And please only add new entries to the top of this list, right below the `# Unre
   sent a cancel or frame event.
 - On iOS, send `RedrawEventsCleared` even if there are no redraw events, consistent with other platforms.
 - **Breaking:** Replaced `Window::with_app_id` and `Window::with_class` with `Window::with_name` on `WindowBuilderExtUnix`.
+- On Wayland, fallback CSD was replaced with proper one.
+  - `WindowBuilderExtUnix::with_wayland_csd_theme` used to set color theme in builder.
+  - `WindowExtUnix::wayland_set_csd_theme` used to change color theme on the fly.
```
kchibisov, line 44:
```suggestion
  - `WindowExtUnix::wayland_set_csd_theme` to change color theme on the fly.
```

CHANGELOG.md
```diff
@@ -39,6 +39,12 @@ And please only add new entries to the top of this list, right below the `# Unre
   sent a cancel or frame event.
 - On iOS, send `RedrawEventsCleared` even if there are no redraw events, consistent with other platforms.
 - **Breaking:** Replaced `Window::with_app_id` and `Window::with_class` with `Window::with_name` on `WindowBuilderExtUnix`.
+- On Wayland, fallback CSD was replaced with proper one.
+  - `WindowBuilderExtUnix::with_wayland_csd_theme` used to set color theme in builder.
```
kchibisov, line 43:
```suggestion
  - `WindowBuilderExtUnix::with_wayland_csd_theme` to set color theme when creating a window.
```

CHANGELOG.md
```diff
@@ -39,6 +39,12 @@ And please only add new entries to the top of this list, right below the `# Unre
   sent a cancel or frame event.
 - On iOS, send `RedrawEventsCleared` even if there are no redraw events, consistent with other platforms.
 - **Breaking:** Replaced `Window::with_app_id` and `Window::with_class` with `Window::with_name` on `WindowBuilderExtUnix`.
+- On Wayland, fallback CSD was replaced with proper one.
```
kchibisov, line 42:
```suggestion
- On Wayland, fallback CSD was replaced with proper one:
```

CHANGELOG.md
```diff
@@ -39,6 +39,12 @@ And please only add new entries to the top of this list, right below the `# Unre
...
+  - `WindowBuilderExtUnix::with_wayland_csd_theme` used to set color theme in builder.
+  - `WindowExtUnix::wayland_set_csd_theme` used to change color theme on the fly.
+  - `WINIT_WAYLAND_CSD_THEME` env variable was added, it can be used to set "dark"/"light" theme in apps that don't expose theme setting.
+  - `wayland-csd-adwaita` feature that enables proper CSD.
+  - `wayland-csd-title` feature that enables title rendering, uses FreeType system library.
```
kchibisov, line 47: Still need to change those, see comment in `Cargo.toml`.

kchibisov (changes requested):
The CHANGELOG bits are optional, but we're getting close. Thanks for your work so far.

Cargo.toml
kchibisov: Try installing `libexpat1-dev`.

CHANGELOG.md
```diff
@@ -39,6 +39,12 @@ And please only add new entries to the top of this list, right below the `# Unre
   sent a cancel or frame event.
 - On iOS, send `RedrawEventsCleared` even if there are no redraw events, consistent with other platforms.
 - **Breaking:** Replaced `Window::with_app_id` and `Window::with_class` with `Window::with_name` on `WindowBuilderExtUnix`.
+- On Wayland, fallback CSD was replaced with proper one;
```
kchibisov, line 42:
```suggestion
- On Wayland, fallback CSD was replaced with proper one:
```

kchibisov (changes requested):
Could you also apply this
```diff
diff --git a/src/platform_impl/linux/wayland/window/mod.rs b/src/platform_impl/linux/wayland/window/mod.rs
index 7bd98854..601d9c8a 100644
--- a/src/platform_impl/linux/wayland/window/mod.rs
+++ b/src/platform_impl/linux/wayland/window/mod.rs
@@ -158,12 +158,7 @@ impl Window {
                 }
             });
 
-            let config = match theme {
-                Theme::Light => sctk_adwaita::FrameConfig::light(),
-                Theme::Dark => sctk_adwaita::FrameConfig::dark(),
-            };
-
-            window.set_frame_config(config);
+            window.set_frame_config(theme.into());
         }
 
         // Set decorations.
@@ -577,3 +572,13 @@ impl Drop for Window {
         self.send_request(WindowRequest::Close);
     }
 }
+
+#[cfg(feature = "sctk-adwaita")]
+impl From<Theme> for sctk_adwaita::FrameConfig {
+    fn from(theme: Theme) -> Self {
+        match theme {
+            Theme::Light => sctk_adwaita::FrameConfig::light(),
+            Theme::Dark => sctk_adwaita::FrameConfig::dark(),
+        }
+    }
+}
diff --git a/src/platform_impl/linux/wayland/window/shim.rs b/src/platform_impl/linux/wayland/window/shim.rs
index b9defe0e..03d64e63 100644
--- a/src/platform_impl/linux/wayland/window/shim.rs
+++ b/src/platform_impl/linux/wayland/window/shim.rs
@@ -425,11 +425,7 @@ pub fn handle_window_requests(winit_state: &mut WinitState) {
                 }
                 #[cfg(feature = "sctk-adwaita")]
                 WindowRequest::CsdThemeVariant(theme) => {
-                    let config = match theme {
-                        Theme::Light => sctk_adwaita::FrameConfig::light(),
-                        Theme::Dark => sctk_adwaita::FrameConfig::dark(),
-                    };
-                    window_handle.window.set_frame_config(config);
+                    window_handle.window.set_frame_config(theme.into());
 
                     let window_update = window_updates.get_mut(window_id).unwrap();
                     window_update.refresh_frame = true;

```

CHANGELOG.md
PolyMeilex: My bad, fixed

PolyMeilex:
> Could you also apply this

Done

kchibisov (approved):

kchibisov:
Thanks!

PR #2264, opened 2022-04-25
Fix embedded NULs in C wide strings returned from Windows API

exactly-one-kas:
- [ ] ~Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users~
  - This change fixes an error introduced in b222dde83504e402d51fa40a3c78dee40cb2ee5f, which hasn't been included in a release yet

### DIFF
--- a/src/platform_impl/windows/monitor.rs
+++ b/src/platform_impl/windows/monitor.rs
@@ -1,10 +1,7 @@
 use std::{
     collections::{BTreeSet, VecDeque},
-    ffi::OsString,
     hash::Hash,
-    io, mem,
-    os::windows::prelude::OsStringExt,
-    ptr,
+    io, mem, ptr,
 };
 
 use windows_sys::Win32::{
@@ -17,6 +14,7 @@ use windows_sys::Win32::{
     },
 };
 
+use super::util::decode_wide;
 use crate::{
     dpi::{PhysicalPosition, PhysicalSize},
     monitor::{MonitorHandle as RootMonitorHandle, VideoMode as RootVideoMode},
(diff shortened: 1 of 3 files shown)


msiglreith (approved):

madsmtm:
Uhh, actually, wasn't the previous behaviour correct?

I'm fairly certain that truncating at NULL bytes is generally treated as a potential security vulnerability, e.g. someone could create a `MonitorHandle::name` that had the value `"abc\0malicious"` but `winit` users would only see `"abc"`. In this case, not really scary, mostly just talking in principle, were it e.g. a path name or a username or something that you allow passing on elsewhere it would have been more worrisome.

Maybe what you meant was to _strip_ NULL bytes? (Instead of truncating?)

aloucks:
Perhaps an issue should be opened with `windows-sys` if the behavior changed after adopting that library.

msiglreith:
In both cases the returned string from the windows api is null-terminated so I think the proposed changes (actually reverting to the previous behaviour) are fine. For the monitor name we don't even have a length so we would basically hope anything after the name is untouched as we zero it beforehand.

> Maybe what you meant was to strip NULL bytes? (Instead of truncating?)

I don't see the benefit here given the proposed threat model from above. It would be still easy to forge a name resulting in the same identifier imo.

Besides that I don't see any discussion in the windows-sys PR on this topic so it seems rather just slipped through - therefore probably better to discuss it in a separate issue.

exactly-one-kas:
The `windows` crate is "just" a thin wrapper over the Windows API. The `szDevice` field on `MONITORINFOEXW` ist just a C wide-string that's allocated within the struct with a maximum length of `CCHDEVICENAME`. (Likely to save a heap-allocation.)

That array beind embedded in the struct is the only reason that we can safely get a slice with embedded zeroes - `GetMonitorInfo()` behaves as intended.

PR #2266, opened 2022-04-26
Fix assigning the wrong monitor when receiving Windows move events

exactly-one-kas:


maroider:
Can you describe the bug being fixed here in more detail (or link to a relevant issue if I've missed one)?

exactly-one-kas:
When I set a window to be in borderless fullscreen mode for a non-primary monitor and later use `window.fullscreen()` to get the current monitor for the window, it reports the primary monitor despite the window still being on the other monitor.

Windows sends `WM_WINDOWPOSCHANGING` with partial coordinates if for example the window gets resized. In that case x and y are zero and the `SWP_NOMOVE` flags is set.

Without using the current coordinates of the window for `x` and `y` this results in `MonitorFromRect()` returning the primary monitor.

msiglreith (approved):
thanks!

PR #2268, opened 2022-04-26
Support not applying an activation policy

indygreg:
This commit introduces an `ActivationPolicy::None` variant that
enables us to skip the call to `NSApp::setActivationPolicy()` at
application startup time.

I needed this functionality in order to author a system tray only
application - one that has a menu in the system tray but not an icon
in the dock.

If you apply an activation policy, the app icon appears in the dock,
even if your bundle uses `LSUIElement = true` in its Info.plist.
Interestingly, if you run the same executable outside the context
of an app bundle and don't set an activation policy, I can never get
focus to work in text elements. So it looks like apps in the wild
will need to conditionally define the activation policy at run-time
depending on whether they are run from a bundle or not.

Arguably the proper way to model "no activation policy" would be to
use `Option<T>`. I chose not to do this so I didn't break the API.
But if you want me to refactor to use `Option<T>`, I could do that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,8 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On iOS, send `RedrawEventsCleared` even if there are no redraw events, consistent with other platforms.
 - **Breaking:** Replaced `Window::with_app_id` and `Window::with_class` with `Window::with_name` on `WindowBuilderExtUnix`.
 - On Wayland and X11, fix window not resizing with `Window::set_inner_size` after calling `Window:set_resizable(false)`.
+- A new `ActivationPolicy::None` variant can be used to not set an activation policy. This variant may be useful for system tray only applications not wanting a dock icon.
+
 
 # 0.26.1 (2022-01-05)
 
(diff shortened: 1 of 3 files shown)


madsmtm:
Hmm, I'm not sure I understand, it sounds to me like you just want to use `builder.with_activation_policy(ActivationPolicy::Accessory)` (then you wouldn't have to define `LSUIElement = 1` either)?

indygreg:
Wait - does `NSApplicationActivationPolicyAccessory` actually work?! I thought I tried this but couldn't get the system tray to display. Or maybe I didn't try it because I misinterpreted the Apple docs saying it disables the menu bar (which I conflated with the system tray, which often uses `NSMenu`). Give me a few minutes to test this...

indygreg:
Took me a while to get my `egui` based app to be able to call in the new API for controlling activation settings since I had to port things to use unreleased crate versions, but, yes, `ActivationPolicy::Accessory` works as expected!

(I'm new to AppKit development and don't really know what I'm doing yet.)

Sorry for the PR spam: the new APIs do everything I need regarding window activation!

madsmtm:
No problem, glad you figured it out!

Issue #2270, opened 2022-04-27
Optimization and better compatibility with math libraries


Wandalen:
Most math libraries define vectors as `#[repr(C)]`. 
For example popular libs `cgmath` and `nalgebra` do that:
- [cgmath::Vector2< _ >](https://docs.rs/cgmath/0.18.0/cgmath/struct.Vector2.html)
- [nalgebra::Vector2< _ >](https://docs.rs/nalgebra/0.30.1/nalgebra/base/type.Vector2.html)

`winit` does not define such structures as [winit](https://docs.rs/winit/0.26.1/winit/index.html)::[dpi](https://docs.rs/winit/0.26.1/winit/dpi/index.html)::[PhysicalSize](https://docs.rs/winit/0.26.1/winit/dpi/struct.PhysicalSize.html#) as `#[repr(C)]`. If `winit` followed the same pattern, it would let just reinterpret data instead of constructing another vector. That would make code more optimal, `winit` more compatible with math libraries and more comfortable to use. Although, you may argue each pro, nevertheless, I don't see any con.

Can I please open PR?

madsmtm:
I think this is a case of very premature optimization, especially because a normal reinterpretation of the bytes compiles down to nothing, see [this playground link (view the generated assembly in release mode)](https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=4c7b3c0de6ce2fa424f97d57ca10ba6f).

Wandalen:
I am offering only to add `#[repr(C)]` to structures. Why is it necessary? Because most math libs do that for vectors. Definitions of vectors should be compatible among libs. Forget about my use case, I am sure there are more use cases requiring this changes.

Please attach label `type: enhancement`

Wandalen:
@madsmtm as I noted, optimization is not the main reason vectors should have `#[repr(C)]`. Also, optimization not always work as in the over-simplified experiment. If you will embed vector in another structures compiler is much more restricted in what it can do. Also compiler might fail to optimize it in case if you build for web and optimization is set for size. 

And contrary, `#[repr(C)]` does not have any negative effect as a far as I see.

maroider:
I personally don't see much use for it, but you also don't have to ask before opening a PR. For more complex issues (and certain bugfixes), it may be preferable to open an issue first or coordinate through other channels (like our matrix/irc room/channel).

> Also compiler might fail to optimize it in case if you build for web and optimization is set for size.

[...]

PR #2271, opened 2022-04-28
Make defenition of vectors compatible with math libs.

Wandalen:
Adding reprt(C) make definition of vectors of winit compatible with definition of vectors in math libs.
Most math libs define vector with reprt(C).

Enhancment #2270.

I can add all that.


Wandalen:
Is any other changes required?

Issue #1006, opened 2019-07-03
Implement kms backend?


goddessfreya:
Right, so there was some chatter about implementing a kmsmod backend like SDL's. Mouse input could be provided by a library outside of winit that uses evdev.

Such a backend would allow us to run applications in the tty, without the presence of X11/Wayland/DirectFB, which I'm sure would open up a lot of silly possibilities. To use alacritty as an example, such a backend would allow it to replace similar projects like kmscon, which I think would kind of be nifty.

@chrisduerr What's your opinion on running alacritty in the tty?

Osspial:
Could you provide a link to the SDL discussion and what exactly kmsmod is? I can't seem to find either from some quick google searches.

chrisduerr:
> @chrisduerr What's your opinion on running alacritty in the tty?

Running Alacritty itself in the TTY seems a bit pointless tbh, since it builds on GUI and GPU Acceleration at its foundation, but doesn't provide much functionality when it comes to window management (it instead outsources that to the WM).

However, it would be possible to rewrite the renderer of Alacritty for a TTY backend, but at that point it would probably be easier to just skip glutin/winit completely.

However I've never had any interest in replacing my console, so maybe other people have different opinions here. I don't think it wouldn't be fun to play around with, but I would never upstream support for a PR that implements this since it's just a load of mostly unnecessary (to me) complexity.

And I feel like that might be similar for winit/glutin in general. I'd be curious *why* this was requested. Which kind of application should this support and who's the target audience? Since many features of winit/glutin probably don't exist there anyways, would it maybe be better to just write a completely separate toolkit? What can winit/glutin provide here that wouldn't have to be written from scratch?

Also additional platforms of course always mean additional work and maintenance overhead. Testing in CI might be difficult and it is probably even more work than just testing linux with a display server. So who would be willing to maintain it? Initial implementation is one thing, but often maintainership is even harder than the initial support.

Of course these are just my two cents, making guesses about a chat conversation I haven't been part of, but in general I'm not immediately in love with the idea.

ZeGentzy:
> Could you provide a link to the SDL discussion and what exactly kmsmod is? I can't seem to find either from some quick google searches.

Discussion took place on discord. Go to the gamedev channel on the rust community discord and search for `there's one thing winit doesn't do as far as i know that i would like to: run straight out of a terminal`.

https://wiki.archlinux.org/index.php/Kernel_mode_setting

ZeGentzy:
[...]

PR #2272, opened 2022-04-29
Initial kmsdrm support

StratusFearMe21:

## What's this?
This is the start of a kmsdrm backend for winit! Essentially, if wayland and X.org are not available, this PR makes winit create a gbm device and a libinput listener to send the device window events. 

## KMS/DRM
**KMS/DRM** is the system that wayland compositors and X use to draw things to the screen. It's an abstraction over the graphics card essentially. In order to render a framebuffer to the screen, you must
1. Open the global DRM file descriptor
2. Select a connector (monitor)
3. Select a CRTC
4. Add a framebuffer
5. Add a plane
6. Commit all previous changes to the DRM file descriptor

My PR does steps 1-4 and 6. Steps 5 is where the user will come in and initialize a framebuffer for their graphics library of choice.

## Event Handling
KMS/DRM mode does not have a traditional input system, in fact it has none at all. It requires that you grab the raw input file descriptors from `/dev/input` and read those. This is where `libinput` comes in. This is the same library that `smithay` and many wayland compositors use to complete this task. It even provides an event based interface which integrates very well with `winit`. I chose this library because it melds very well with `winit`, and it handles every type of input that `winit` supports out of the box, *however*

- I heard talk in #1006 about using a Rust crate to handle raw input, `libinput` is not written in Rust
- `libinput` supports multi-touch *given* that you are not performing a gesture. In this case `libinput` will produce gesture events and only tell you when the gesture begins, moves, and ends. These gestures do not provide finger ids either.
- The way that my backend is implemented, it produces both `WindowEvent`s and `DeviceEvent`s even if a window has not been initialized.

Also, I should probably mention, there is an invisible *virtual* cursor that my KMS/DRM backend simulates which takes deltas from `libinput` and produces `DeviceEvent`s for the raw mouse deltas and `WindowEvent`s for the virtual location of the cursor on screen.

It it possible to access the gbm device mutex via the `gbm_device()`, the current connector via `gbm_connector()` and the current crtc via the `gbm_crtc()` method in the `EventLoopWindowTargetExtUnix` trait

## Note
- I'm positively stumped on what to do about these warnings
```shell
warning: use of deprecated field `event::KeyboardInput::modifiers`: Deprecat
ed in favor of WindowEvent::ModifiersChanged
   --> src/platform_impl/linux/drm/event_loop.rs:529:88
    |
529 | ...                   }, virtual_keycode: CHAR_MAPPINGS[k as usize], m
odifiers: self.modifiers } , is_synthetic: false }}, &mut ());
    |                                                                      ^
^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(deprecated)]` on by default

warning: associated function is never used: `dummy`
  --> src/platform_impl/linux/drm/mod.rs:49:25
   |
49 |     pub const unsafe fn dummy() -> Self {
   |                         ^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: associated function is never used: `dummy`
   --> src/platform_impl/linux/drm/mod.rs:138:25
    |
138 |     pub const unsafe fn dummy() -> Self {
    |                         ^^^^^

warning: `winit` (lib) generated 3 warnings
```

## In the future
I could add an actual cursor and render it to the screen using the cursor framebuffer.

## Thanks for considering my PR!


maroider:
Wow! This is neat. I'll be sure to look closer at this when I can, but for now, I just have to note that I think your editor's autoformatting bungled up our `CHANGELOG.md` (all instances of `-` for bullet-points have become `*`) and `Cargo.toml` (4-space indentation has become tab-based indentation).

StratusFearMe21:
> Wow! This is neat. I'll be sure to look closer at this when I can, but for now, I just have to note that I think your editor's autoformatting bungled up our `CHANGELOG.md`, `FEATURES.md` (all instances of `-` for bullet-points have become `*`), and `Cargo.toml` (4-space indentation has become tab-based indentation).

Whoops! Sorry about that. I'm using [remark](https://github.com/wooorm/remark) and [taplo](https://github.com/tamasfe/taplo) for my formatting, I'll go back and fix the formatting on the documents you mentioned

maroider:
`Cargo.toml` and `FEATURES.md` both still have what looks like unintentional formatting changes.

StratusFearMe21:
Sorry about that, I fixed the formatting again

StratusFearMe21:
Forget about the `raw-window-handle` issue, I've confirmed myself that gbm is indeed thread-safe. You'll see above the PR I made to raw-window-handle

Drakulix:
One of the maintainers of drm-rs/gbm.rs/input.rs/smithay here, this looks very well put together and I cannot spot any obvious mistakes. (But obviously I don't know a lot about winit internals.)

One nitpick: The mode selection is likely copied from some outdated example code. drm-rs gained the option to read out mode-types recently. Most monitors have a "Preferred" mode, which is likely what winit should select. (https://docs.rs/drm/latest/drm/control/struct.Mode.html#method.mode_type).

Additionally, I think there was an earlier attempt to do something like this for glutin, I think(?): https://github.com/Smithay/gbm.rs/pull/2
I am not in the loop at the great scheme of things here, but if there is any way, I can help with reviewing or adjusting the code of any of these libs, please let me know. I would be glad to support such an endeavour, even if I don't have much time to write code for it myself. Given our wayland-compositor project smithay might also contain some code, that could be useful in the future, I would like to avoid code duplication where possible and rather build a shared foundation, if that is of any interest. I also see it as a net positive, that drm-rs/gbm.rs/input.rs might get some more usage in the future with this implementation.

Thanks for working on this! :+1:

StratusFearMe21:
> One nitpick: The mode selection is likely copied from some outdated example code.
Yeah, guilty. That should be fixed now

StratusFearMe21:
@Drakulix, Is it possible to capture the keyboard with libinput? Currently all the key presses that you make in the window are passed to the tty, and that just doesn't seem ideal.

StratusFearMe21:
Thanks for the input, I'll start with `libseat`

StratusFearMe21:
`libseat` works like a champ, and on top of that integrates flawlessly with `libinput`. Thanks for the tip

StratusFearMe21:
@Drakulix, `libseat` seems to be working great, however, it seems to prevent the app from logging anything to the console. Is there a solution to this?

StratusFearMe21:
@Drakulix, this fork PR is very close to being feature compete. All it needs is a propper cursor, however, no matter what I try, I can't get a cursor buffer/planw working. Is their some reference I can look at for how to make and use the DRM  cursor plane?

Drakulix:
> @Drakulix, this PR is very close to being feature compete. All it needs is a propper cursor, however, no matter what I try, I can't get a cursor buffer/plane working. Is their some reference I can look at for how to make and use the DRM cursor plane?

I don't think we have a single compositor, that makes use of the cursor plane, right now.. (they used to, but not since we switched to the atomic api).

The code in smithay to support this use-case does exist however: https://github.com/Smithay/smithay/blob/master/src/backend/drm/surface/atomic.rs#L870-L917 . It should be enough to just add those to your `AtomicModeReq`.

A thing you might be missing is, that the kernel expects `SRC_W` and `SRC_H` in 16.16 fixed point values: https://github.com/Smithay/smithay/blob/master/src/backend/drm/surface/atomic.rs#L886

Another is that, as far as I know, there is no guarantee, that a dumb-buffer will work for cursor planes (depending on the driver). So you might be better of using gbm for that. (You can request a writable gbm-buffer, so you don't need a GL context to render to it: https://docs.rs/gbm/latest/gbm/struct.BufferObjectFlags.html#associatedconstant.WRITE)

MarijnS95:
One nitpick: I've generally heard of this as DRM/KMS since KMS is afaik part of the DRM subsystem and not the other way around: https://www.kernel.org/doc/html/latest/gpu/drm-kms.html.

As such, would it make sense to `s/kmsdrm/drmkms` or is there some underlying reason? If this has already been documented/discussed, apologies!

StratusFearMe21:
> One nitpick: I've generally heard of this as DRM/KMS since KMS is afaik part of the DRM subsystem and not the other way around: https://www.kernel.org/doc/html/latest/gpu/drm-kms.html.
> 
> As such, would it make sense to `s/kmsdrm/drmkms` or is there some underlying reason? If this has already been documented/discussed, apologies!

The reason I do it like this is because thats how SDL2 does it https://wiki.libsdl.org/FAQUsingSDL. But of course, I'd be happy to `s/kmsdrm/drmkms` if that isnt a good enough reason.

StratusFearMe21:
> Another is that, as far as I know, there is no guarantee, that a dumb-buffer will work for cursor planes (depending on the driver). So you might be better of using gbm for that.

Oof, I just removed gbm as a dependency on this package and made it only required in glutin. I did this because Vulkan uses an entirely different system than GBM for framebuffers and such. I guess its not a big deal that this implementation has no visible cursor

StratusFearMe21:
@Drakulix, `libseat` works very well, however, it seems to supress the app from printing anything. Is there a way to fix this?

Drakulix:
> @Drakulix, `libseat` works very well, however, it seems to supress the app from printing anything. Is there a way to fix this?
> 
> P.S: thanks for being so helpful on this PR.

Not really, sadly. The graphical tty-mode suppresses any log messages getting printed to the tty and the kernel does not keep a buffer of "missed" messages.

StratusFearMe21:
> > @Drakulix, `libseat` works very well, however, it seems to supress the app from printing anything. Is there a way to fix this?
> > P.S: thanks for being so helpful on this PR.
> 
> Not really, sadly. The graphical tty-mode suppresses any log messages getting printed to the tty and the kernel does not keep a buffer of "missed" messages.

I see, thanks for responding so quickly

StratusFearMe21:
I'm pretty confident with this PR at this point. As a functionality test, I ported Alacritty to this PR and now it runs straight from the tty https://github.com/StratusFearMe21/alacritty

StratusFearMe21:
@Drakulix,  If I have an `xkbcommon` context, how do I get the default options for what the keymap should be? If I put empty strings for all the options, it compiles the keymap but it doesn't seem to compile the right one.

StratusFearMe21:
Nevermind, it must be something on my end

StratusFearMe21:
Okay, I fixed the issue, logical error

StratusFearMe21:
@Drakulix, (sorry for *@ing* you so much). I'm still trying to get the cursor buffer working, but it's still throwing me an `InvalidArgument` when I try to commit it. This is the code I'm trying to get working
```rust
use drm::control::Device as ControlDevice;
use drm::Device;

use drm::buffer::DrmFourcc;

use drm::control::ResourceHandle;
use drm::control::{self, atomic, connector, crtc, property, AtomicCommitFlags};
use gbm::BufferObjectFlags;

#[derive(Debug)]
/// A simple wrapper for a device node.
pub struct Card(std::fs::File);

/// Implementing `AsRawFd` is a prerequisite to implementing the traits found
/// in this crate. Here, we are just calling `as_raw_fd()` on the inner File.
impl std::os::unix::io::AsRawFd for Card {
    fn as_raw_fd(&self) -> std::os::unix::io::RawFd {
        self.0.as_raw_fd()
    }
}

/// With `AsRawFd` implemented, we can now implement `drm::Device`.
impl Device for Card {}
impl ControlDevice for Card {}

/// Simple helper methods for opening a `Card`.
impl Card {
    pub fn open(path: &str) -> Self {
        let mut options = std::fs::OpenOptions::new();
        options.read(true);
        options.write(true);
        Card(options.open(path).unwrap())
    }

    pub fn open_global() -> Self {
        Self::open("/dev/dri/card0")
    }
}

fn find_prop_id<T: ResourceHandle>(
    card: &Card,
    handle: T,
    name: &'static str,
) -> Option<property::Handle> {
    let props = card
        .get_properties(handle)
        .expect("Could not get props of connector");
    let (ids, _vals) = props.as_props_and_values();
    ids.iter()
        .find(|&id| {
            let info = card.get_property(*id).unwrap();
            info.name().to_str().map(|x| x == name).unwrap_or(false)
        })
        .cloned()
}

pub fn main() {
    let card_raw = Card::open_global();
    let card = gbm::Device::new(card_raw).unwrap();

    card.set_client_capability(drm::ClientCapability::UniversalPlanes, true)
        .expect("Unable to request UniversalPlanes capability");
    card.set_client_capability(drm::ClientCapability::Atomic, true)
        .expect("Unable to request Atomic capability");
    card.get_driver_capability(drm::DriverCapability::CursorWidth)
        .unwrap();
    card.get_driver_capability(drm::DriverCapability::CursorHeight)
        .unwrap();

    // Load the information.
    let res = card
        .resource_handles()
        .expect("Could not load normal resource ids.");
    let coninfo: Vec<connector::Info> = res
        .connectors()
        .iter()
        .flat_map(|con| card.get_connector(*con))
        .collect();
    let crtcinfo: Vec<crtc::Info> = res
        .crtcs()
        .iter()
        .flat_map(|crtc| card.get_crtc(*crtc))
        .collect();

    // Filter each connector until we find one that's connected.
    let con = coninfo
        .iter()
        .find(|&i| i.state() == connector::State::Connected)
        .expect("No connected connectors");

    // Get the first (usually best) mode
    let &mode = con.modes().get(0).expect("No modes found on connector");

    let (disp_width, disp_height) = mode.size();

    // Find a crtc and FB
    let crtc = crtcinfo.get(0).expect("No crtcs found");

    // Select the pixel format
    let fmt = DrmFourcc::Xrgb8888;

    // Create a DB
    // If buffer resolution is above display resolution, a ENOSPC (not enough GPU memory) error may
    // occur
    let mut db = card
        .create_buffer_object::<()>(
            disp_width.into(),
            disp_height.into(),
            fmt,
            BufferObjectFlags::WRITE | BufferObjectFlags::SCANOUT,
        )
        .expect("Could not create dumb buffer");
    let mut cursor_db = card
        .create_buffer_object::<()>(
            64,
            64,
            fmt,
            BufferObjectFlags::WRITE | BufferObjectFlags::SCANOUT | BufferObjectFlags::CURSOR,
        )
        .expect("Could not create dumb buffer");

    // Map it and grey it out.
    {
        let mut buffer = Vec::new();
        for _ in 0..disp_width {
            for _ in 0..disp_height {
                buffer.push(10);
                buffer.push(10);
                buffer.push(10);
                buffer.push(10);
            }
        }
        db.write(&buffer).unwrap().unwrap();
    }

    {
        let mut buffer = Vec::new();
        for _ in 0..64 {
            for _ in 0..64 {
                buffer.push(255);
                buffer.push(255);
                buffer.push(255);
                buffer.push(255);
            }
        }
        cursor_db.write(&buffer).unwrap().unwrap();
    }

    // Create an FB:
    let fb = card
        .add_framebuffer(&db, 24, 32)
        .expect("Could not create FB");

    let cursor_fb = card
        .add_framebuffer(&db, 24, 32)
        .expect("Could not create FB");

    let planes = card.plane_handles().expect("Could not list planes");
    let (better_planes, compatible_planes): (
        Vec<control::plane::Handle>,
        Vec<control::plane::Handle>,
    ) = planes
        .planes()
        .iter()
        .filter(|&&plane| {
            card.get_plane(plane)
                .map(|plane_info| {
                    let compatible_crtcs = res.filter_crtcs(plane_info.possible_crtcs());
                    compatible_crtcs.contains(&crtc.handle())
                })
                .unwrap_or(false)
        })
        .partition(|&&plane| {
            if let Ok(props) = card.get_properties(plane) {
                let (ids, vals) = props.as_props_and_values();
                for (&id, &val) in ids.iter().zip(vals.iter()) {
                    if let Ok(info) = card.get_property(id) {
                        if info.name().to_str().map(|x| x == "type").unwrap_or(false) {
                            return val == (drm::control::PlaneType::Primary as u32).into();
                        }
                    }
                }
            }
            false
        });
    let plane = *better_planes.get(0).unwrap_or(&compatible_planes[0]);

    let cursor_plane = *compatible_planes
        .iter()
        .find(|&&plane| {
            if let Ok(props) = card.get_properties(plane) {
                let (ids, vals) = props.as_props_and_values();
                for (&id, &val) in ids.iter().zip(vals.iter()) {
                    if let Ok(info) = card.get_property(id) {
                        if info.name().to_str().map(|x| x == "type").unwrap_or(false) {
                            return val == (drm::control::PlaneType::Cursor as u32).into();
                        }
                    }
                }
            }
            false
        })
        .unwrap();

    println!("{:#?}", mode);
    println!("{:#?}", fb);
    println!("{:#?}", cursor_fb);
    println!("{:#?}", db);
    println!("{:#?}", cursor_db);
    println!("{:#?}", plane);
    println!("{:#?}", cursor_plane);

    let mut atomic_req = atomic::AtomicModeReq::new();
    atomic_req.add_property(
        con.handle(),
        find_prop_id(&card, con.handle(), "CRTC_ID").expect("Could not get CRTC_ID"),
        property::Value::CRTC(Some(crtc.handle())),
    );
    let blob = card
        .create_property_blob(&mode)
        .expect("Failed to create blob");
    atomic_req.add_property(
        crtc.handle(),
        find_prop_id(&card, crtc.handle(), "MODE_ID").expect("Could not get MODE_ID"),
        blob,
    );
    atomic_req.add_property(
        crtc.handle(),
        find_prop_id(&card, crtc.handle(), "ACTIVE").expect("Could not get ACTIVE"),
        property::Value::Boolean(true),
    );

    atomic_req.add_property(
        plane,
        find_prop_id(&card, plane, "FB_ID").expect("Could not get FB_ID"),
        property::Value::Framebuffer(Some(fb)),
    );
    atomic_req.add_property(
        plane,
        find_prop_id(&card, plane, "CRTC_ID").expect("Could not get CRTC_ID"),
        property::Value::CRTC(Some(crtc.handle())),
    );
    atomic_req.add_property(
        plane,
        find_prop_id(&card, plane, "SRC_X").expect("Could not get SRC_X"),
        property::Value::UnsignedRange(0),
    );
    atomic_req.add_property(
        plane,
        find_prop_id(&card, plane, "SRC_Y").expect("Could not get SRC_Y"),
        property::Value::UnsignedRange(0),
    );
    atomic_req.add_property(
        plane,
        find_prop_id(&card, plane, "SRC_W").expect("Could not get SRC_W"),
        property::Value::UnsignedRange((mode.size().0 as u64) << 16),
    );
    atomic_req.add_property(
        plane,
        find_prop_id(&card, plane, "SRC_H").expect("Could not get SRC_H"),
        property::Value::UnsignedRange((mode.size().1 as u64) << 16),
    );
    atomic_req.add_property(
        plane,
        find_prop_id(&card, plane, "CRTC_X").expect("Could not get CRTC_X"),
        property::Value::SignedRange(0),
    );
    atomic_req.add_property(
        plane,
        find_prop_id(&card, plane, "CRTC_Y").expect("Could not get CRTC_Y"),
        property::Value::SignedRange(0),
    );
    atomic_req.add_property(
        plane,
        find_prop_id(&card, plane, "CRTC_W").expect("Could not get CRTC_W"),
        property::Value::UnsignedRange(mode.size().0 as u64),
    );
    atomic_req.add_property(
        plane,
        find_prop_id(&card, plane, "CRTC_H").expect("Could not get CRTC_H"),
        property::Value::UnsignedRange(mode.size().1 as u64),
    );

    atomic_req.add_property(
        cursor_plane,
        find_prop_id(&card, cursor_plane, "FB_ID").expect("Could not get FB_ID"),
        property::Value::Framebuffer(Some(cursor_fb)),
    );
    atomic_req.add_property(
        cursor_plane,
        find_prop_id(&card, cursor_plane, "CRTC_ID").expect("Could not get CRTC_ID"),
        property::Value::CRTC(Some(crtc.handle())),
    );
    atomic_req.add_property(
        cursor_plane,
        find_prop_id(&card, cursor_plane, "SRC_X").expect("Could not get SRC_X"),
        property::Value::UnsignedRange(0),
    );
    atomic_req.add_property(
        cursor_plane,
        find_prop_id(&card, cursor_plane, "SRC_Y").expect("Could not get SRC_Y"),
        property::Value::UnsignedRange(0),
    );
    atomic_req.add_property(
        cursor_plane,
        find_prop_id(&card, cursor_plane, "SRC_W").expect("Could not get SRC_W"),
        property::Value::UnsignedRange(64 << 16),
    );
    atomic_req.add_property(
        cursor_plane,
        find_prop_id(&card, cursor_plane, "SRC_H").expect("Could not get SRC_H"),
        property::Value::UnsignedRange(64 << 16),
    );
    atomic_req.add_property(
        cursor_plane,
        find_prop_id(&card, cursor_plane, "CRTC_X").expect("Could not get CRTC_X"),
        property::Value::SignedRange(0),
    );
    atomic_req.add_property(
        cursor_plane,
        find_prop_id(&card, cursor_plane, "CRTC_Y").expect("Could not get CRTC_Y"),
        property::Value::SignedRange(0),
    );
    atomic_req.add_property(
        cursor_plane,
        find_prop_id(&card, cursor_plane, "CRTC_W").expect("Could not get CRTC_W"),
        property::Value::UnsignedRange(64),
    );
    atomic_req.add_property(
        cursor_plane,
        find_prop_id(&card, cursor_plane, "CRTC_H").expect("Could not get CRTC_H"),
        property::Value::UnsignedRange(64),
    );

    card.atomic_commit(AtomicCommitFlags::ALLOW_MODESET, atomic_req)
        .expect("Failed to set mode");

    // Set the crtc
    // On many setups, this requires root access.

    let five_seconds = ::std::time::Duration::from_millis(5000);
    ::std::thread::sleep(five_seconds);

    card.destroy_framebuffer(fb).unwrap();
}
```
I guess what I'm asking is what am I missing?

src/platform_impl/linux/drm/event_loop.rs
```diff
@@ -0,0 +1,1341 @@
...
+                    return Err(crate::error::OsError::new(
+                        line!(),
+                        file!(),
+                        crate::platform_impl::OsError::DrmMisc("Failed to dispatch seat"),
+                    ));
+                }
+            }
+            s
+        };
+        let dev = seat.open_device(&"/dev/dri/card0").map_err(|_| {
```
Drakulix, line 806: There is no guarantee for `card0` being the right one (e.g. on optimus systems). Even the naming is just a linux convention.

A better way would be to select the primary gpu e.g. via udev (see smithay's primary_gpu function for example: https://github.com/Smithay/smithay/blob/master/src/backend/udev.rs#L231-L263).

But really this should be configurable (e.g. via an environment variable).

kchibisov: Yeah, at this rate winit could become kiosk wayland compositor :P 

Just env variable could be fine, I guess.

StratusFearMe21:
> I also gave these changes a try with Slint and it works out of the box. Amazing work @StratusFearMe21!

Yay! I'm glad everything's working!

StratusFearMe21:
Yeah, at this point, I think this is ready to be merged. Im very happy with this PR in its current state.

src/platform_impl/linux/drm/event_loop.rs
```diff
@@ -0,0 +1,1417 @@
...
+        self.context.as_raw_fd()
+    }
+}
+
+impl EventSource for LibinputInputBackend {
+    type Event = crate::event::Event<'static, ()>;
+    type Metadata = ();
+    type Ret = ();
+
+    fn process_events<F>(
```
kchibisov, line 109: This function is barely readable due to overly complex nesting. Certain names are just one character. First of all I'd suggest to extract event handling parts.

reworked object creation, since you're using full pathes all over the place, could try looking into other backends which try aliasing backend stuff with `PlatformDeviceId` and crate public API left as is e.g. `DeviceId`.

This alone should reduce lots of code here and make it more readable.

src/platform_impl/linux/drm/event_loop.rs
```diff
@@ -0,0 +1,1417 @@
...
+
+    /// drm mode
+    pub mode: drm::control::Mode,
+
+    /// drm plane
+    pub plane: drm::control::plane::Handle,
+
+    // /// drm dumbbuffer containing the cursor
+    // pub cursor_buffer: drm::control::framebuffer::Handle,
+    /// drm device
```
kchibisov, line 708: Doc comment went wrong.

src/platform_impl/linux/drm/event_loop.rs
```diff
@@ -0,0 +1,1417 @@
...
+    pub connector: drm::control::connector::Info,
+
+    /// drm crtc
+    pub crtc: drm::control::crtc::Info,
+
+    /// drm mode
+    pub mode: drm::control::Mode,
+
+    /// drm plane
+    pub plane: drm::control::plane::Handle,
```
kchibisov, line 704: All doc comments must start with capital character. So `Drm connector.`, `Drm crtc` and so on.

src/platform_impl/linux/drm/event_loop.rs
```diff
@@ -0,0 +1,1417 @@
...
+                    if let libseat::SeatEvent::Enable = event {
+                        t_active.store(true, std::sync::atomic::Ordering::SeqCst);
+                    }
+                },
+                None,
+            )
+            .map_err(|_| {
+                crate::error::OsError::new(
+                    line!(),
+                    file!(),
```
kchibisov, line 791: I'm not sure that forwarding line and file is any useful? Well it could be while debugging, but it seems to much and not really sensible, since users won't look into code anyway.

src/platform_impl/linux/drm/event_loop.rs
```diff
@@ -0,0 +1,1417 @@
...
+                    crate::platform_impl::OsError::DrmMisc("Failed to open libseat"),
```
kchibisov, line 792: I think error messages start with lower cased letter.

src/platform_impl/linux/drm/event_loop.rs
```diff
@@ -0,0 +1,1417 @@
...
+                )
+            })?;
+
+            while !active.load(std::sync::atomic::Ordering::SeqCst) {
+                if let Err(_) = s.dispatch(-1) {
+                    return Err(crate::error::OsError::new(
+                        line!(),
+                        file!(),
+                        crate::platform_impl::OsError::DrmMisc("Failed to dispatch seat"),
+                    ));
```
kchibisov, line 802: That looks wrong. You basically get error, discard it and send something you've made up instead of forwarding more useful error. Probably will apply to other error reporting when working with system libs.

src/platform_impl/linux/drm/event_loop.rs
```diff
@@ -0,0 +1,1417 @@
...
+            }
+            s
+        };
+        // Safety
+        //
+        // This string value has the same lifetime as the seat in question, and will not be dropped
+        // until the seat is, which is not before `udev_assign_seat` is run.
+        let seat_name = unsafe { std::mem::transmute::<&str, &'static str>(seat.name()) };
+        let card_path = std::env::var("WINIT_DRM_CARD").ok().map_or_else(
+            || {
```
kchibisov, line 813: I'd extracted it into separate function, since it's too much for closure to be readable.

src/platform_impl/linux/drm/event_loop.rs
```diff
@@ -0,0 +1,1417 @@
...
+                                    .unwrap_or_else(|| std::ffi::OsString::from("seat0"))
+                                    == seat_name
+                            })
+                            .flat_map(|device| device.devnode().map(std::path::PathBuf::from))
+                            .next()
+                    })
+                    .ok_or(crate::error::OsError::new(
+                        line!(),
+                        file!(),
+                        crate::platform_impl::OsError::DrmMisc("Failed to find suitable GPU"),
```
kchibisov, line 877: Also, I think it'd make sense for that particular error when we failed to find GPU inform about env variable right away.

src/platform_impl/linux/drm/event_loop.rs
```diff
@@ -0,0 +1,1417 @@
...
+        )
+        .ok_or(crate::error::OsError::new(
+            line!(),
+            file!(),
+            crate::platform_impl::OsError::DrmMisc("Failed to compile XKB keymap"),
+        ))?;
+        let state = xkb::State::new(&keymap);
+        let compose_table = xkb::compose::Table::new_from_locale(
+            &xkb_ctx,
+            std::env::var_os("LC_ALL")
```
kchibisov, line 910: The code is very dense and hard to read, would suggest to logically space it, at least by operation kind. This applies to everything from what I can say in that PR.

src/platform_impl/linux/drm/event_loop.rs
```diff
@@ -0,0 +1,1417 @@
...
+                line!(),
+                file!(),
+                crate::platform_impl::OsError::DrmMisc("Failed to initialize DRM"),
+            )
+        })?;
+        let drm = Card(std::sync::Arc::new(dev.1));
+        let mut input = input::Libinput::new_with_udev(Interface(seat, HashMap::new()));
+        input.udev_assign_seat(seat_name).unwrap();
+        let xkb_ctx = xkb::Context::new(xkb::CONTEXT_NO_FLAGS);
+        let keymap = xkb::Keymap::new_from_names(
```
kchibisov, line 893: Should probably abstract xkb handling.

src/platform_impl/linux/drm/event_loop.rs
```diff
@@ -0,0 +1,1417 @@
...
+        }
+        }
+        false
+        });
+        let plane = *better_planes.get(0).unwrap_or(&compatible_planes[0]);
+        */
+        let (p_better_planes, p_compatible_planes): (
+            Vec<drm::control::plane::Handle>,
+            Vec<drm::control::plane::Handle>,
+        ) = planes
```
kchibisov, line 1053: Could move plane creation to some function.

In general, I think it'd make sense to make code in event loop creation more declarative, and documented wrt steps. So it'll be easier for others to grasp it.

src/platform_impl/linux/drm/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+impl Ord for MonitorHandle {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        self.0.interface_id().cmp(&other.0.interface_id())
+    }
+}
+
+impl MonitorHandle {
+    #[inline]
+    pub fn name(&self) -> Option<String> {
+        Some(String::from("card0"))
```
kchibisov, line 38: I'm not sure from where compositors get display names, but it'd be nice to have those.

I'd probably returned `None`, since it's not like a name right now, and arbitrary, since user could pick `card1`...

src/platform_impl/linux/drm/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+impl VideoMode {
+    #[inline]
+    pub fn size(&self) -> PhysicalSize<u32> {
+        let size = self.0.size();
+        PhysicalSize::new(size.0 as u32, size.1 as u32)
+    }
+
+    #[inline]
+    pub fn bit_depth(&self) -> u16 {
+        32
```
kchibisov, line 84: That also looks broken.

src/platform_impl/linux/drm/window.rs
```diff
@@ -0,0 +1,354 @@
...
+        atomic_req.add_property(
+            event_loop_window_target.plane,
+            find_prop_id(
+                &event_loop_window_target.device,
+                event_loop_window_target.plane,
+                "CRTC_H",
+            )
+            .expect("Could not get CRTC_H"),
+            property::Value::UnsignedRange(event_loop_window_target.mode.size().1 as u64),
+        );
```
kchibisov, line 175: All of those `atomic_req` stuff is looking the same, consider adding a function or macro for them. Will likely work in some other places.

src/platform_impl/linux/drm/window.rs
```diff
@@ -0,0 +1,354 @@
...
+    }
+
+    #[inline]
+    pub fn outer_size(&self) -> PhysicalSize<u32> {
+        self.inner_size()
+    }
+
+    #[inline]
+    pub fn set_inner_size(&self, _size: Size) {
+        // It's technically possible to do this by changing video modes but that seems a little
```
kchibisov, line 237: This shouldn't be done by changing video mode. Just state that it's impossible to do so.

src/platform_impl/linux/drm/window.rs
```diff
@@ -0,0 +1,354 @@
...
+    pub fn is_resizable(&self) -> bool {
+        false
+    }
+
+    #[inline]
+    pub fn set_cursor_icon(&self, _cursor: CursorIcon) {}
+
+    #[inline]
+    pub fn set_cursor_grab(&self, _grab: bool) -> Result<(), ExternalError> {
+        Err(ExternalError::NotSupported(NotSupportedError::new()))
```
kchibisov, line 260: You'd need to update everything in public documentation wrt those restrictions iirc. Docs in e.g. `src/window.rs` have `Platform specific` sections, and `drm` should be documented everywhere it diverges or something isn't supported.

src/platform_impl/linux/drm/window.rs
```diff
@@ -0,0 +1,354 @@
...
+        Some(Fullscreen::Exclusive(crate::monitor::VideoMode {
+            video_mode: crate::platform_impl::VideoMode::Drm(super::VideoMode(
+                self.0,
+                self.1.clone(),
+            )),
+        }))
+    }
+
+    #[inline]
+    pub fn set_fullscreen(&self, _monitor: Option<Fullscreen>) {}
```
kchibisov, line 308: And this call is actually possible, since it's the one that changes video modes with exclusive fullscreen.

src/platform_impl/linux/drm/window.rs
```diff
@@ -0,0 +1,354 @@
...
+            .connectors()
+            .iter()
+            .map(|f| super::MonitorHandle(self.3.get_connector(*f).unwrap()))
+            .collect()
+    }
+
+    #[inline]
+    pub fn raw_window_handle(&self) -> raw_window_handle::DrmHandle {
+        let mut rwh = raw_window_handle::DrmHandle::empty();
+        rwh.fd = self.3.as_raw_fd();
```
kchibisov, line 344: Could it be more descriptive than `self.3`?

src/platform_impl/linux/wayland/event_loop/proxy.rs
```diff
@@ -2,7 +2,7 @@
 
 use std::sync::mpsc::SendError;
 
-use sctk::reexports::calloop::channel::Sender;
+use calloop::channel::Sender;
```
kchibisov, line 5: That's tough, I'm not sure how we should handle such sharing. In general, I'd probably make sharing of `calloop` between drm and wayland backend, so they stay on the same version.

src/platform_impl/linux/xkb_keymap.rs
```diff
@@ -0,0 +1,415 @@
+//! Convert Wayland keys to winit keys.
+#![allow(non_upper_case_globals)]
+
+use crate::event::VirtualKeyCode;
+
+/*
+ * TTY function keys, cleverly chosen to map to ASCII, for convenience of
```
kchibisov, line 7: That looks like yoinked from client-toolkit or something. Could you get rid of C style comments please?

kchibisov (changes requested):
I'm not a drm expert or anything like that, I just went though code style and changes you're bringing to winit, so we'll try keep the code in some kind of good state from the start.

Besides suggestions I've noted. I'd suggest to run clippy and look for issues in your particular backend (clippy fails right now on winit).

Besides that you should make CI green and went through out docs to mention about drm backend in platform supported stuff.

Also, the CI job must be altered to test drm backend as well and it should pass in CI in some way. Though, given that libseat is unlikely to be present on ubuntu we should somehow handle that.

Thanks for your patch, that's really neat.

StratusFearMe21:
Alrighty, thanks for reviewing my PR. I've never made a PR to such a high-profile crate before, so sorry that my formatting is a little inconsistent.  It's a little bit of a disappointment though that DRM can't be a default backend. Anyhoo, I'll begin this refactor tommorrow

src/platform_impl/linux/xkb_keymap.rs
StratusFearMe21: Guilty as charged

kchibisov:
>I was hoping we could do it like how SDL2 does it where it is possible to run apps using the kmsdrm video driver, and whether the app cares or not didn't matter. Anyhoo, I'll begin this refactor tomorrow

My main concert is that it brings lots of dependencies most users not sure that want to have. Also it's impossible to run it on some nvidia systems due to lack of gbm(some has it now though).

Also, certain stuff like `multiwindow` isn't possible on it, so applications should explicitly handle it and tune their stuff wrt that. Trying to run and straight crashing due to how limited kms backend is doesn't sound ideal to me.

Though, maybe other winit maintainers/downstream users have better opinion on that topic.

StratusFearMe21:
GBM Is a superset of the DRM API exclusively for creating and managing framebuffers. Therefore Nvidia systems shouldn't be effected by not using GBM. As for `glutin`, it *does* use GBM for buffer management, and my PR adding that can be found [here](https://github.com/rust-windowing/glutin/pull/1403)

StratusFearMe21:
Yeah, I guess the big weakness in this patch is that it brings in a lot of C dependencies. I agree with you on that.

StratusFearMe21:
As for GBM support for NVidia GPUs, [that shouldn't be a problem](https://www.nvidia.com/download/driverResults.aspx/181274/en-us)

Drakulix:
> As for GBM support on NVidia GPUs, [that shouldn't be a problem](https://www.nvidia.com/download/driverResults.aspx/181274/en-us)

There are a bunch of fairly common cards, that are not supported by 495 anymore and need to use 470-legacy drivers.

So sadly this will remain a problem for the foreseeable future. Still that shouldn't hold this PR back imo.

PolyMeilex:
> Though, given that libseat is unlikely to be present on ubuntu we should somehow handle that.

Just so you know, it is packaged in Ubuntu 22, but it havent't reached its LTS release yet, so I debut it's available in CI as ubuntu-latest as of now. So `libseat-rs` should be easily testable in CI soon'ish. 

Meanwhile, in Smithay we build it manually: https://github.com/Smithay/smithay/blob/d5d8b670246d0da26bfd4fd19c4cbad5ca71bf9d/.github/workflows/ci.yml#L180-L189

src/platform_impl/linux/xkb_keymap.rs
StratusFearMe21: Fixed

Cargo.toml
```diff
@@ -17,36 +17,37 @@ default-target = "x86_64-unknown-linux-gnu"
...
-log = "0.4"
-serde = { version = "1", optional = true, features = ["serde_derive"] }
-raw-window-handle = "0.4.2"
-bitflags = "1"
-mint = { version = "0.5.6", optional = true }
+instant = { version = "0.1.12", features = ["wasm-bindgen"] }
+lazy_static = "1.4.0"
+log = "0.4.17"
+serde = { version = "1.0.137", optional = true, features = ["serde_derive"] }
+raw-window-handle = { git = "https://github.com/rust-windowing/raw-window-handle" }
```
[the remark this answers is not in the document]
StratusFearMe21: Well, the reason I'm using the Git dependency in this PR is because [I made a PR](https://github.com/rust-windowing/raw-window-handle/pull/91) to `raw-window-handle` which was merged but has not been released yet

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- On Linux, initial support has been added for a kmsdrm backend
```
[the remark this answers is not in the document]
StratusFearMe21: Fixed

src/platform/unix.rs
```diff
@@ -25,13 +28,43 @@ use crate::platform_impl::{
...
+        unsafe { std::fs::File::from_raw_fd(*self.0) };
+    }
+}
+
+#[cfg(feature = "kmsdrm")]
+/// With `AsRawFd` implemented, we can now implement `drm::Device`.
+impl Device for Card {}
+#[cfg(feature = "kmsdrm")]
+impl ControlDevice for Card {}
+
```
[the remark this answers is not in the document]
StratusFearMe21: Fixed

src/platform/unix.rs
```diff
@@ -42,6 +75,9 @@ pub trait EventLoopWindowTargetExtUnix {
     #[cfg(feature = "x11")]
     fn is_x11(&self) -> bool;
 
+    #[cfg(feature = "kmsdrm")]
+    fn is_drm(&self) -> bool;
```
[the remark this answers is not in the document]
StratusFearMe21: Fixed, I guess this locks in DRM being a breaking change huh?

src/platform_impl/linux/drm/event_loop.rs
```diff
@@ -0,0 +1,1417 @@
...
+    xkb_keymap: xkb::Keymap,
+    xkb_compose: xkb::compose::State,
+    token: Token,
+    touch_location: PhysicalPosition<f64>,
+    screen_size: (u32, u32),
+    modifiers: ModifiersState,
+    cursor_positon: PhysicalPosition<f64>,
+    timer_handle: calloop::timer::TimerHandle<(KeyboardInput, Option<char>)>,
+    // cursor_plane: drm::control::plane::Handle,
+    // cursor_buffer: drm::control::framebuffer::Handle,
```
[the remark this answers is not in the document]
StratusFearMe21: Fixedd

src/platform_impl/linux/drm/event_loop.rs
StratusFearMe21: Fixed

src/platform_impl/linux/drm/event_loop.rs
StratusFearMe21: Fixed

src/platform_impl/linux/drm/event_loop.rs
StratusFearMe21: Fixed

src/platform_impl/linux/drm/event_loop.rs
StratusFearMe21: Fixed

src/platform_impl/linux/drm/event_loop.rs
StratusFearMe21: Fixed

src/platform_impl/linux/drm/event_loop.rs
StratusFearMe21: fixed

src/platform_impl/linux/drm/event_loop.rs
StratusFearMe21: Fixed

src/platform_impl/linux/drm/event_loop.rs
StratusFearMe21: wdym? Like add another dependency?

Cargo.toml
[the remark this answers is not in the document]
StratusFearMe21: Same scenario with `xkbcommon`

src/platform_impl/linux/drm/mod.rs
StratusFearMe21: Fixed

src/platform_impl/linux/drm/mod.rs
StratusFearMe21: Bit depth is decided by the client (eg. `glutin`), not winit. That might be a problem.

src/platform_impl/linux/drm/window.rs
StratusFearMe21: Fixed

src/platform_impl/linux/drm/window.rs
StratusFearMe21: Fixed

src/platform_impl/linux/drm/event_loop.rs
StratusFearMe21: Fixed

src/platform_impl/linux/drm/window.rs
StratusFearMe21: Fixed

src/platform_impl/linux/drm/event_loop.rs
StratusFearMe21: Fixed

src/platform_impl/linux/wayland/event_loop/proxy.rs
StratusFearMe21: Fixed

src/platform_impl/linux/drm/window.rs
StratusFearMe21: Fixed

StratusFearMe21:
I've finished with all of @kchibisov 's suggestions, *except* for getting the CI to work. Unfortunately, I've never really used GitHub Workflows before and it would be very helpful is someone (@Drakulix @PolyMeilex) could assist me with this.

dhardy:
I don't mind providing the odd hint on CI tooling...

```
Error:   thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: "`\"pkg-config\" \"--libs\" \"--cflags\" \"libudev\"` did not exit successfully: exit status: 1\nerror: could not find system library 'libudev' required by the 'libudev-sys' crate\n\n--- stderr\nPackage libudev was not found in the pkg-config search path.\nPerhaps you should add the directory containing `libudev.pc'\nto the PKG_CONFIG_PATH environment variable\nNo package 'libudev' found\n"', /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/libudev-sys-0.1.4/build.rs:38:41
```

Not uncommon to have minimal dependencies installed by default in CI and install required libs as an extra step. E.g. [Bevy installs libudev-dev here](https://github.com/bevyengine/bevy/blob/main/.github/workflows/ci.yml#L41=).

```
error: `#[doc(hidden)]` is ignored on trait impl items
Error:    --> src/platform/unix.rs:109:5
    |
109 |     #[doc(hidden)]
    |     ^^^^^^^^^^^^^^ help: remove this attribute
```
Should be self-explanatory? There's a few more of these (last failing runner).

StratusFearMe21:
Im not having a problem grasping why the CI is failing (I use Arch BTW). Its that everytime I write a new `ci.yml` it infinitely "Expects task to start" and never runs.

src/platform_impl/linux/kms/event_loop.rs
```diff
@@ -0,0 +1,1461 @@
...
+        )
+        .ok_or_else(|| {
+            OsError::new(
+                line!(),
+                file!(),
+                platform_impl::OsError::KmsMisc("failed to compile XKB keymap"),
+            )
+        })?;
+
+        let state = xkb::State::new(&keymap);
```
kchibisov, line 1022: I do remember that you've asked wrt what should be done with `xkb`. I was saying that you should split code from that file into separate modules, so you'd have input related stuff in input.rs, and so on, since right now this is quite a huge file to grasp.

The logical separation done which is done in wayland/x11 backends could be taken as an example.

Right now my main issue with this particular file that it contains all the logic for everything in winit, but if you split related code into separate files and be more declarative it would be much more helpful for others to read it.

src/platform_impl/linux/kms/event_loop.rs
```diff
@@ -0,0 +1,1461 @@
...
+        let card_path = std::env::var("WINIT_DRM_CARD")
+            .ok()
+            .map_or_else(|| find_card_path(seat_name), |p| Ok(Into::into(p)))?;
+
+        #[cfg(feature = "kms-ext")]
+        // Opening the card using our seat allows us to do so unprivallaged
+        let dev = seat
+            .open_device(&card_path)
+            .map_err(|e| {
+                OsError::new(
```
kchibisov, line 962: There's a macro `os_error!` for that already in `src/error.rs`

kchibisov (changes requested):
Just some general notes.

src/platform_impl/linux/kms/event_loop.rs
```diff
@@ -0,0 +1,1461 @@
...
+        token: Token,
+        mut callback: F,
+    ) -> std::io::Result<PostAction>
+    where
+        F: FnMut(Self::Event, &mut ()) -> Self::Ret,
+    {
+        if token == self.token {
+            self.context.dispatch()?;
+
+            for event in &mut self.context {
```
StratusFearMe21, line 680: I can't borrow `self` as mutable twice, hence why all these are macros

src/platform_impl/linux/kms/event_loop.rs
```diff
@@ -0,0 +1,1461 @@
...
+                match event {
+                    input::Event::Device(ev) => handle_device_event!(ev, callback),
+                    input::Event::Touch(ev) => handle_touch_event!(self, ev, callback),
+                    input::Event::Tablet(ev) => handle_tablet_tool_event!(self, ev, callback),
+                    input::Event::Pointer(ev) => handle_pointer_event!(self, ev, callback),
+                    input::Event::Keyboard(ev) => handle_keyboard_event!(self, ev, callback),
```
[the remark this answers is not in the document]
StratusFearMe21: https://github.com/rust-windowing/winit/pull/2272/files/b5caf34276370ea147db43eb49659d9a65a3b90d#r869652968

src/platform_impl/linux/kms/event_loop.rs
StratusFearMe21: Fixed

src/platform/unix.rs
```diff
@@ -49,24 +55,54 @@ pub trait EventLoopWindowTargetExtUnix {
...
-    fn is_x11(&self) -> bool {
-        !self.p.is_wayland()
+    fn unix_backend(&self) -> Backend {
+        match self.p {
+            #[cfg(feature = "x11")]
+            LinuxEventLoopWindowTarget::X(_) => Backend::X,
+            #[cfg(feature = "wayland")]
+            LinuxEventLoopWindowTarget::Wayland(_) => Backend::Wayland,
+            #[cfg(feature = "kms")]
+            LinuxEventLoopWindowTarget::Kms(_) => Backend::Kms,
```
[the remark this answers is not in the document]
StratusFearMe21: Well, `drm` is also a crate which this crate depends on soooo...

src/platform/unix.rs
```diff
@@ -49,24 +55,54 @@ pub trait EventLoopWindowTargetExtUnix {
...
+
+    /// Returns the current mode of the drm device
+    ///
+    /// Returns `None` if the `EventLoop` doesn't use drm.
+    #[cfg(feature = "kms")]
+    fn drm_mode(&self) -> Option<Mode>;
+
+    /// Returns the primary plane of the drm device
+    ///
+    /// Returns `None` if the `EventLoop` doesn't use drm.
```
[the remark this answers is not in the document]
StratusFearMe21: Well, drm is also a crate which this crate depends on soooo...

src/platform_impl/linux/kms/event_loop.rs
StratusFearMe21: Fixed?

StratusFearMe21:
I'm literally stumped, I'm not knowledgable enough with Github workflows to solve this problem, if someone could fix the `ci.yml` for me, that would be very helpful

madsmtm:
Hey @StratusFearMe21, are you still interested in working on this?

In the spirit of moving this forward, maybe CI could be dropped for now, and we could create an issue to tackle that later?

StratusFearMe21:
Wow! I'm absolutely honored, it would be my pleasure to help maintain winit @madsmtm

madsmtm:
An extra maintainer would be very welcome!

I've sent you an invite to the @rust-windowing organization, for now you'll have write access to `winit` and `glutin`, but not administrator rights; the process for granting you higher permissions is not really existing, [see this](https://github.com/rust-windowing/winit/issues/2319) for a bit of background; but I digress, if you need it, it'll probably happen at some point.

If, when you fix the merge conflicts, could add yourself to `.github/CODEOWNERS` and [the testers and contributors table](https://github.com/rust-windowing/winit/wiki/Testers-and-Contributors), then you'll get pinged when people change the backend.

A few notes on how we work: While we technically have a policy for when things can be merged (2 maintainers must have reviewed), it is rarely followed because of lack of manpower - try to have a few people review your own PRs before merging them, but remember that it is better to merge a PR than have it sitting around forever. Re platforms that no-one support (currently Web), any maintainer may merge PRs here.

Are you on [Matrix / IRC](https://github.com/rust-windowing/winit#contact-us)? Would be nice if you were, I find that's a good way to stay in touch; feel free to ask any questions if you're uncertain about something!

StratusFearMe21:
Okay, so it seems I've left this PR alone for too long. At this point I think I'll have to re-write this PR from Scratch. In the meantime though I'm gonna wait until this PR (https://github.com/rust-windowing/glutin/pull/1435) is merged before doing that.This PR has been in the works for ~5 months now and I think I can do a better job now than I could then anyway. Thanks for understanding

madsmtm:
No worries, I hope you can at least reuse some of the work you've done in this PR!

Cargo.toml
```diff
@@ -17,36 +17,38 @@ default-target = "x86_64-unknown-linux-gnu"
 targets = ["i686-pc-windows-msvc", "x86_64-pc-windows-msvc", "i686-unknown-linux-gnu", "x86_64-unknown-linux-gnu", "x86_64-apple-darwin", "wasm32-unknown-unknown"]
 
 [features]
-default = ["x11", "wayland", "wayland-dlopen"]
+default = ["x11", "wayland", "wayland-dlopen", "kms"]
```
notgull, line 20: IMO `drmkms` should be opt-in rather than opt-out

madsmtm:
@StratusFearMe21 there's some people on Matrix wondering what your plans with this are?

madsmtm:
Thanks for the update! And don't worry, it's fine, we all have a life to tend to outside open source ;). Take the time you need!

mraerino:
I took the liberty of picking up this PR and doing the grunt-work of merging current master into it and making it compile: https://github.com/rust-windowing/winit/pull/2795

daxpedda:
Closing due to inactivity.
See #1006.

PR #2285, opened 2022-05-09
linux: Send scancode instead of keycodes

rofferom:
X11 sends Linux keycodes that only match scancodes on a limited
part of the keys. For example for extended scancodes, the values
don't match.

showkey command can display a mapping:
- https://linux.die.net/man/1/showkey
- https://github.com/legionus/kbd/blob/master/src/showkey.c

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -95,6 +95,7 @@ x11-dl = { version = "2.18.5", optional = true }
 percent-encoding = { version = "2.0", optional = true }
 parking_lot = { version = "0.12.0", optional = true }
 libc = "0.2.64"
+keycode = "0.3.0"
 
 [target.'cfg(target_arch = "wasm32")'.dependencies.web_sys]
 package = "web-sys"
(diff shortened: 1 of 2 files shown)


maroider:
Thanks for taking the time to contribute (and apologies for taking so long to respond). Is this supposed to map Linux's keycodes to USB HID scancodes?

rofferom:
Yes it is, but actually all implementations send "OS keycodes" rather than the raw HID scancode. My patch is incomplete, I'm not event using it anymore. Furthermore, it is a breaking change.

Maybe the correct fix would be to update the documentation of `winit::event::ScanCode`, because the values are more "OS/Hardware-dependent keyboard scan code".

What do you think about that ?

maroider:
> What do you think about that?

Welp, I kind of forgot to reply before now. Basically, I think #2662 is preferable, since it would give you a proper platform-independent value instead of passing through the platform-dependent "scancode/keycode".

kchibisov:
Thanks for your patch, but as it was stated and as it was there for years I think, the keyboard API model proposed in #2662 took care of it.

Issue #2284, opened 2022-05-08
[macOS] EventLoop::run_return with multiple windows does not return


kcking:
Modifying the window_run_return example to spawn multiple windows (but keep a single event loop) results in the event_loop never returning control to the caller. This happens when focusing the non-first window. I have only been testing this on macOS.

```rs
// Limit this example to only compatible platforms.
#[cfg(any(
    target_os = "windows",
    target_os = "macos",
    target_os = "linux",
    target_os = "dragonfly",
    target_os = "freebsd",
    target_os = "netbsd",
    target_os = "openbsd"
))]
fn main() {
    use std::{thread::sleep, time::Duration};

    use simple_logger::SimpleLogger;
    use winit::{
        event::{Event, WindowEvent},
        event_loop::EventLoop,
        platform::run_return::EventLoopExtRunReturn,
        window::{Window, WindowBuilder},
    };
    let mut event_loop = EventLoop::new();

    SimpleLogger::new().init().unwrap();
    let _window = WindowBuilder::new()
        .with_title("A fantastic window!")
        .build(&event_loop)
        .unwrap();

    let mut quit = false;

    let mut frame = 0u128;

    let mut windows = vec![];

    while !quit {
        frame += 1;
        event_loop.run_return(|event, _, control_flow| {
            control_flow.set_wait();

            if let Event::WindowEvent { event, .. } = &event {
                // Print only Window events to reduce noise
                println!("{:?}", event);
            }

            match event {
                Event::WindowEvent {
                    event: WindowEvent::CloseRequested,
                    ..
                } => {
                    quit = true;
                }
                Event::MainEventsCleared => {
                    control_flow.set_exit();
                }
                _ => (),
            }
        });

        if frame % 300 == 0 {
            windows.push(Window::new(&event_loop));
        }

        // Sleep for 1/60 second to simulate rendering
        println!("rendering");
        sleep(Duration::from_millis(16));
    }
}

#[cfg(any(target_os = "ios", target_os = "android", target_arch = "wasm32"))]
fn main() {
    println!("This platform doesn't support run_return.");
}

```

kcking:
With some debugging I found that the cause appears to be the `is_dialog` logic in https://github.com/rust-windowing/winit/blob/f3f6f1008a5891ed0c9e5e8f26568f8f2d49f164/src/platform_impl/macos/app_state.rs#L412

hard-coding this to false fixes the issue for me, but I'm first going to look further into what this logic is for and if there is a general fix.

[...]

PR #2027, opened 2021-09-29
macOS: Fix native file dialogs freezing the event loop

emilk:
Closes https://github.com/rust-windowing/winit/issues/1779

Previously all native dialogs, such as [rfd](https://github.com/PolyMeilex/rfd), would cause the event loop (event_loop.run) to freeze.

All code by @PolyMeilex (at https://github.com/PolyMeilex/winit/commit/e4a0a15fc42496a03b76a633ea9d123209751f39)

I just opened the PR as @PolyMeilex says he doesn't have the time for it right now.

- [x] Tested on all platforms changed
- [x] Compilation warnings were addressed
- [x] `cargo fmt` has been run on this branch
- [x] `cargo doc` builds successfully
- [x] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users
- [ ] ~Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior~
- [ ] ~Created or updated an example program if it would help users understand this functionality~
- [ ] ~Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented~

PR #2292, opened 2022-05-16
[macOS] remove old dialog fix that seems to be superceded by https://github.c…

kcking:
…om/rust-windowing/winit/pull/2027/files

It appears that an [old fix](https://github.com/rust-windowing/winit/pull/1581) for having multiple windows opening (in particular a native file dialog) broke `EventLoop::run_return` on macOS. In particular, this PR prevented the EventLoop from exiting if more than one window was open.

Since then, a [more general fix](https://github.com/rust-windowing/winit/pull/2027) has been released that prevents running the EventLoop re-entrant-ly.

This PR essentially manually reverts #1581, fixing EventLoop::run_return on macOS when multiple windows are open.

  - tested with `rfd` file dialog winit example 
  - also tested multiple windows being open and returning from EventLoop::run_return
~- [ ] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior~
~- [ ] Created or updated an example program if it would help users understand this functionality~
~- [ ] Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented~

This PR fixes https://github.com/rust-windowing/winit/issues/2284

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - **Breaking:** Added new `WindowEvent::Ime` supported on desktop platforms.
 - Added `Window::set_ime_allowed` supported on desktop platforms.
 - **Breaking:** IME input on desktop platforms won't be received unless it's explicitly allowed via `Window::set_ime_allowed` and new `WindowEvent::Ime` events are handled.
+- On macOS, fixed an issue where having multiple windows would prevent run_return from ever returning.
 
 # 0.26.1 (2022-01-05)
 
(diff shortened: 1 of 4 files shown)


madsmtm (commented):
Wow this is a brittle workaround in any case, and could probably even end up being unable to close in some fullscreen transition corner cases - that said, have you tested how this works while performing a fullscreen transition?

And has this been tested using [`tinyfiledialogs-rs`](https://github.com/jdm/tinyfiledialogs-rs) as well (I think they use a different approach from other file dialog libraries)?

kcking:
Thanks for taking a look!

I published my test repo (which originally just had an rfd test) and added tinydialogs and the winit fullscreen example. https://github.com/kcking/filedialogtest

Both of them appear to work fine for me. Let me know if there's anything else you think should be tried.

madsmtm (approved):
I tried the `winit` fullscreen example with more than one window, shutdown seems to work flawlessly now (which it definitely didn't before).

Thanks a lot for investigating and fixing this!

Issue #1199, opened 2019-10-01
Integration with async ecosystem


dylanede:
At the moment, I believe it is unclear how to best integrate futures and asynchronous execution in an application with a winit event loop. Ideally, I would imagine that futures/streams could be created for events (e.g. a stream for window resize events), and some special executor would run on the current thread that internally calls `EventLoop::run`. I am not sure however how to integrate this with, for example, tokio. Perhaps this is all something that could be built in a separate crate on top of winit, though I think this is the best place to discuss it.

The main immediate benefit I see of futur-ising the API is allowing state machines involving user input to be written much more naturally using `async` functions.

ZeGentzy:
I think @Osspial been working on something async related.

Osspial:
I don't think Winit `async`/`await` `async`/`await` support should live in the main repository, since it can be cleanly implemented on top of Winit's existing event loop API, it isn't necessary in all cases, and there's much more room for creating an opinionated API. However, I'd be alright with rust-windowing endorsing an official `async` wrapper for Winit.

Also, I have worked on an async wrapper, although it doesn't work on master quite yet. You can check it out here. https://github.com/osspial/winit-async

ryanisaacg:
One fairly significant papercut I've run into when working on my own async event loop wrapper is that it's very difficult to manage the lifetimes required to create a window during the execution of the Future.

FredrikNoren:
Here's a small example of how to get winit+wgpu+tokio working: https://gist.github.com/FredrikNoren/7c3535b11e99e8fcd8dd3d55f9a934a2

exellian:
> Your example is trivially fixed by adding a `move`, as is done in most of winit's examples.
> 
> ```diff
> - event_loop.run(|event, target, cf| {
> + event_loop.run(move |event, target, cf| {
> ```
> 
> I also can't see a `'static` bound on [Runtime::block_on](https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#method.block_on)'s signature: `pub fn block_on<F: Future>(&self, future: F) -> F::Output`.

Ok thanks you are right Runtime::block_on actually doesn't require 'static lifetime so the borrowing problem is obsolete. But still the problem is that Runtime::block_on blocks the thread and therefore no events in the meantime will be registered. So i don't see that there is a truly async single threaded event_loop possible with the current design. Only if you would use more than one thread and use channels?

ps:
[...]

PR #2294, opened 2022-05-19
Make `EventLoopProxy` `Sync`

Liamolucko:

Previously, `EventLoopProxy` was `Send` but not `Sync`. The only reason for this that I could see was that `mpsc::Sender` wasn't `Sync`; however, the alternative `mpsc::SyncSender` is, so I've switched to using that instead.

It's not named that because it implements `Sync`;  the difference is that `SyncSender` has a fixed-size buffer and will block if its buffer is full, whereas `Sender` will never block. For the size of the buffer, I've just arbitrarily picked 10; I've got no idea what would be optimal.

The reason I think it needs to implement `Sync` is for the sake of #1199. In an executor which polls futures within the event loop, the best way I can think of to implement `Waker` is by sending an event through `EventLoopProxy` to wake up the event loop; however, `Waker` is `Send` and `Sync`, so `EventLoopProxy` also has to be `Send` and `Sync` for a `Waker` implementation to use it.

~~On the web, `EventLoopProxy` is neither `Send` nor `Sync`. This could be fixed, but would be much more complicated and isn't needed for the async use case - on the web, `wasm_bindgen_futures::spawn_local` could just be used instead, so a custom `Waker` implementation isn't needed.~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- `EventLoopProxy` now implements `Sync` as well as `Send`.
+- `EventLoopProxy` now implements `Send` on WebAssembly.
 - Build docs on `docs.rs` for iOS and Android as well.
 - **Breaking:** Removed the `WindowAttributes` struct, since all its functionality is accessible from `WindowBuilder`.
 - Added `WindowBuilder::transparent` getter to check if the user set `transparent` attribute.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -136,8 +136,10 @@ features = [
     'WheelEvent'
 ]
 
-[target.'cfg(target_arch = "wasm32")'.dependencies.wasm-bindgen]
-version = "0.2.45"
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+wasm-bindgen = "0.2.45"
+wasm-bindgen-futures = "0.4.31"
+async-channel = "1.6.1"
 
 [target.'cfg(target_arch = "wasm32")'.dev-dependencies]
 console_log = "0.2"
--- a/src/platform_impl/ios/event_loop.rs
+++ b/src/platform_impl/ios/event_loop.rs
@@ -4,7 +4,7 @@ use std::{
     fmt::{self, Debug},
     marker::PhantomData,
     mem, ptr,
-    sync::mpsc::{self, Receiver, Sender},
+    sync::mpsc::{self, Receiver, SyncSender},
 };
 
 use crate::{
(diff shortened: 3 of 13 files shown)


madsmtm (changes requested):
I agree with this in principle (haven't looked into the implementation yet), however I would like our platforms to not diverge further on what marker traits they implement (since that is a quite hard-to-debug thing).

So maybe we could fix the "web's `EventLoopProxy` isn't `Send`/`Sync`" while we're at it? If not, I don't think `EventLoopProxy` should be those marker traits on any platform!

Liamolucko:
There are two ways I can think of to implement `Send + Sync` for `EventLoopProxy` on the web:

- Pull in `wasm-bindgen-futures` and some async channel library as dependencies, and spawn a task which just waits for incoming `EventLoopProxy` messages and handles them. This is easier, but pulls in a couple extra dependencies.
- Manually implement a thread waker using atomics, and combine that with a `std` channel to send messages. This would basically mean duplicating the thread waker that `wasm-bindgen-futures` uses internally.

Which would you prefer?

Liamolucko:
Okay, I've now implemented `Send + Sync` on the web as well using `wasm-bindgen-futures` and `async-channel`.

maroider:
I'd honestly prefer the "manual" approach, but I don't know how much more work that would be. In either case, the web backend is kind of weird to me in many ways, so I think I'll have to read more of it and then probably consult someone before I feel comfortable with either approach.

Liamolucko:
Using `SyncSender` here is a bit more problematic than I initially thought - if its buffer is full, calling `EventLoopProxy::send_event` will deadlock, which would be quite frustrating to debug.

Also, this isn't as vital for async as I thought - wrapping the  `EventLoopProxy` in a mutex makes it `Sync`, so `EventLoopProxy` itself doesn't have to be.

daxpedda:
I just stumbled on the same issue, but I only need `Send`, not `Sync`, to implement a multi-threaded engine on Wasm that can send `UserEvent`s to the Winit loop.

I would be more then happy to contribute, what would be the current blocker?

daxpedda:
I will make a new PR only for `Send` then, with the goal of having `EventLoopProxy` share the same traits on all platforms.

daxpedda:
This can be closed in favor of https://github.com/rust-windowing/winit/pull/2834.
Ultimately if we want to implement `Sync` for `EventLoopProxy`, we need to do it for all backends, similar to https://github.com/rust-windowing/winit/pull/2749.

We had a brief talk about this on Matrix and decided that this can be done by the user by just wrapping it around a `Mutex`, which we would have to do internally anyway.

This would also be fixed automatically by https://github.com/rust-lang/rust/pull/111087.

madsmtm:
Noting that this has now been fully resolved by https://github.com/rust-windowing/winit/pull/3449.

PR #2301, opened 2022-05-23
[Windows] Avoid GetModuleHandle(NULL)

### LINKED ISSUE #2300: Windows: Avoid GetModuleHandle(NULL)
The library makes extensive use of GetModuleHandle(NULL) to obtain the instance handle. This is discouraged, as it [does not work for targets that are not executables, such as DLLs](https://stackoverflow.com/questions/21718027/getmodulehandlenull-vs-hinstance).

A possible solution for this would be to use the __ImageBase linker pseudovariable, which is intended for static libraries [as described here](https://devblogs.microsoft.com/oldnewthing/20041025-00/?p=37483).

This trick has also been used in the Windows Implementation Library for C++ which has extensive usage:

https://github.com/microsoft/wil/blob/master/include/wil/win32_helpers.h#L513-L514

AronParker:

Fixes #2300.

winit's GetModuleHandle(NULL) [returns an invalid instance handle if it were to be used inside of a DLL or other non-executable target](https://stackoverflow.com/questions/21718027/getmodulehandlenull-vs-hinstance).. Using the [pseudo-variable of the microsoft linker](https://devblogs.microsoft.com/oldnewthing/20041025-00/?p=37483) not only reduces work, but also works reliably inside both DLLs and executables. More info is inside the issue.

The documentation states that this only works with the microsoft linker, I've tested it using both the stable-GNU and stable-MSVC toolchain and it both seems to work, which suggests that on windows the microsoft linker is always used.

### DIFF
--- a/src/platform_impl/windows/util.rs
+++ b/src/platform_impl/windows/util.rs
@@ -220,7 +220,7 @@ pub fn get_instance_handle() -> HINSTANCE {
         static __ImageBase: IMAGE_DOS_HEADER;
     }
 
-    unsafe { &__ImageBase as *const IMAGE_DOS_HEADER as _ }
+    unsafe { &__ImageBase as *const _ as _ }
 }
 
 impl CursorIcon {
--- a/src/platform_impl/windows/util.rs
+++ b/src/platform_impl/windows/util.rs
@@ -12,11 +12,11 @@ use std::{
 use windows_sys::{
     core::{HRESULT, PCWSTR},
     Win32::{
-        Foundation::{BOOL, HWND, RECT, HINSTANCE},
+        Foundation::{BOOL, HINSTANCE, HWND, RECT},
         Graphics::Gdi::{ClientToScreen, InvalidateRgn, HMONITOR},
         System::{
             LibraryLoader::{GetProcAddress, LoadLibraryA},
-            SystemServices::IMAGE_DOS_HEADER
+            SystemServices::IMAGE_DOS_HEADER,
         },
         UI::{
             HiDpi::{DPI_AWARENESS_CONTEXT, MONITOR_DPI_TYPE, PROCESS_DPI_AWARENESS},
--- a/src/platform_impl/windows/util.rs
+++ b/src/platform_impl/windows/util.rs
@@ -12,9 +12,12 @@ use std::{
 use windows_sys::{
     core::{HRESULT, PCWSTR},
     Win32::{
-        Foundation::{BOOL, HWND, RECT},
+        Foundation::{BOOL, HWND, RECT, HINSTANCE},
         Graphics::Gdi::{ClientToScreen, InvalidateRgn, HMONITOR},
-        System::LibraryLoader::{GetProcAddress, LoadLibraryA},
+        System::{
+            LibraryLoader::{GetProcAddress, LoadLibraryA},
+            SystemServices::IMAGE_DOS_HEADER
+        },
         UI::{
             HiDpi::{DPI_AWARENESS_CONTEXT, MONITOR_DPI_TYPE, PROCESS_DPI_AWARENESS},
             Input::KeyboardAndMouse::GetActiveWindow,
@@ -203,6 +206,21 @@ pub fn get_desktop_rect() -> RECT {
 
 pub fn is_focused(window: HWND) -> bool {
     window == unsafe { GetActiveWindow() }
+}
+
+pub fn get_instance_handle() -> HINSTANCE {
+    // Gets the instance handle by taking the address of the
+    // pseudo-variable created by the microsoft linker:
+    // https://devblogs.microsoft.com/oldnewthing/20041025-00/?p=37483
+
+    // This is preferred over GetModuleHandle(NULL) because it also works in DLLs:
+    // https://stackoverflow.com/questions/21718027/getmodulehandlenull-vs-hinstance
+
+    extern "C" {
+        static __ImageBase: IMAGE_DOS_HEADER;
+    }
+
+    unsafe { &__ImageBase as *const IMAGE_DOS_HEADER as _ }
 }
 
 impl CursorIcon {
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -27,10 +27,7 @@ use windows_sys::Win32::{
         RDW_INTERNALPAINT, SC_SCREENSAVE,
     },
     Media::{timeBeginPeriod, timeEndPeriod, timeGetDevCaps, TIMECAPS, TIMERR_NOERROR},
-    System::{
-        LibraryLoader::GetModuleHandleW, Ole::RevokeDragDrop, Threading::GetCurrentThreadId,
-        WindowsProgramming::INFINITE,
-    },
+    System::{Ole::RevokeDragDrop, Threading::GetCurrentThreadId, WindowsProgramming::INFINITE},
     UI::{
         Controls::{HOVER_DEFAULT, WM_MOUSELEAVE},
         Input::{
@@ -647,7 +644,7 @@ fn create_event_target_window<T: 'static>() -> HWND {
             lpfnWndProc: Some(thread_event_target_callback::<T>),
             cbClsExtra: 0,
             cbWndExtra: 0,
-            hInstance: GetModuleHandleW(ptr::null()),
+            hInstance: util::get_instance_handle(),
             hIcon: 0,
             hCursor: 0, // must be null in order for cursor state to work properly
             hbrBackground: 0,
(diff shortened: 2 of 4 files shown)


src/platform_impl/windows/util.rs
```diff
@@ -205,6 +208,21 @@ pub fn is_focused(window: HWND) -> bool {
     window == unsafe { GetActiveWindow() }
 }
 
+pub fn get_instance_handle() -> HINSTANCE {
+    // Gets the instance handle by taking the address of the
+    // pseudo-variable created by the microsoft linker:
+    // https://devblogs.microsoft.com/oldnewthing/20041025-00/?p=37483
```
filnet, line 214: As mentioned, this works when using a Microsoft linker. 
What about MSYS2/Mingw and assimilated environment ?
Will they provide a Microsoft compatible linker ?

AronParker: I've put together a little executable that tests this:

```c
#include <Windows.h>
#include <stdio.h>

extern IMAGE_DOS_HEADER __ImageBase;

int main() {
	HINSTANCE imgb = (HINSTANCE)&__ImageBase;
	HINSTANCE modh = GetModuleHandle(NULL);

	printf("ImageBase: %p\n", imgb);
	printf("GetModuleHandle: %p\n", modh);

	if (imgb == modh) {
		printf("SUCCESS: pointers are equivalent");
	} else {
		fprintf(stderr, "ERROR: pointer mismatch");
	}

	return 0;
}
```

running it on MSYS2 gives me the following output:

```bash
Aron@DESKTOP-6CIB9GL MINGW64 ~
$ gcc test.c -o test

Aron@DESKTOP-6CIB9GL MINGW64 ~
$ ./test
ImageBase: 00007ff783760000
GetModuleHandle: 00007ff783760000
SUCCESS: pointers are equivalent
Aron@DESKTOP-6CIB9GL MINGW64 ~
$ gcc --version
gcc.exe (Rev1, Built by MSYS2 project) 12.1.0
Copyright (C) 2022 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

Sadly my knowledge about windows linkers isn't deep enough, but my gut feeling is that everyone who went through building a windows linker directly or indirectly implements this "feature".

filnet: Looks good then. I was mostly concerned about MSSY2/Mingw...

Maybe we could fallback to the old behavior when required (should there be a way to detect that).
Not simple since we don't even know if such environments exist at all...

filnet: I went ahead and tested a rust equivalent of your test (rustc does not use gcc but llvm/clang afaik).

```rust
use windows_sys::Win32::Foundation::HINSTANCE;
use windows_sys::Win32::System::LibraryLoader::GetModuleHandleW;
use windows_sys::Win32::System::SystemServices::IMAGE_DOS_HEADER;

extern "C" {
    static __ImageBase: IMAGE_DOS_HEADER;
}

fn main() {
    let imgb = unsafe { &__ImageBase as *const IMAGE_DOS_HEADER as HINSTANCE };
    let modh = unsafe { GetModuleHandleW(std::ptr::null()) };

    println!("ImageBase: {}", imgb);
    println!("GetModuleHandle: {}", modh);

    if imgb == modh {
        println!("SUCCESS: pointers are equivalent");
    } else {
        println!("ERROR: pointer mismatch");
    }
}
```

Works on MINGW64 and on CLANG64 (not shown):

```console
Philippe@DESKTOP MINGW64 ~/rust/handle
$ cargo run
   Compiling handle v0.1.0 (C:\msys64\home\Philippe\rust\handle)
    Finished dev [unoptimized + debuginfo] target(s) in 1.82s
     Running `target\debug\handle.exe`
ImageBase: 140698631536640
GetModuleHandle: 140698631536640
SUCCESS: pointers are equivalent
```

AronParker: > Looks good then. I was mostly concerned about MSSY2/Mingw...
> 
> Maybe we could fallback to the old behavior when required (should there be a way to detect that). Not simple since we don't even know if such environments exist at all...

Sadly I believe implementing a fallback would be quite difficult. There are two cases in which this wouldn't work:

1. `__ImageBase` is not defined:

I have never seen this in practice, might be worth to get in touch with some people who know a bit more about linkers on Windows but I've never seen it fail anywhere. As far as I know, there is no way to detect whether FFI externs are defined at compile-time. The good thing is though, should this trick fail clients will be notified by compile-time errors so it would be quickly noticed and fixed.

2. `__ImageBase` address is wrong:

This can only happen if someone redefined the variable. This is disallowed by the standard anyway (as variable names prefixed with two underscores are reserved for the compiler/implementation). I suppose in debug mode one could print a warning if `__ImageBase` diverges from `GetModuleHandle(NULL)`, although in some cases that is to be expected (e.g. when running inside a DLL).

I have a high degree of confidence that this will be functional and an improvement over the current approach, however I am not an expert on Windows linkers so that's why I put it up as a draft for now. Might be worth looking into what other major libraries use this library to increase confidence in using it.

aloucks: There seems to be another technique described in that stackoverflow link that doesn't depend on the linker. Has there been any experimentation with this? (I have not tried it myself)

https://stackoverflow.com/a/56804892

```
HMODULE getThisModuleHandle()
{
    //Returns module handle where this function is running in: EXE or DLL
    HMODULE hModule = NULL;
    ::GetModuleHandleEx(GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS | 
        GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT, 
        (LPCTSTR)getThisModuleHandle, &hModule);

    return hModule;
}
```

maroider: `__ImageBase` has been around for a long time (check out this [blog post](https://devblogs.microsoft.com/oldnewthing/20041025-00/?p=37483) by Raymond Chen from 2004). I'd be surprised if we encountered any issues with it.

AronParker: The `GetModuleHandleEx` approach is clever and would probably work but there is another advantage of the `__ImageBase`: It does not require a syscall, it has the overhead of a global variable access which makes it the most efficient and fastest option.

Should there be a concern about it not working (even though we have yet to find a platform on where this doesn't work on - compatibility layers around windows have existed for decades and as maroider stated it's probably 2 decades old at this point so it's probably stable), we could lock it behind a feature gate. I'd still be in favour of making `__ImageBase` the default, as it is the more cost efficient option. Should it not be working, the `GetModuleHandleEx` approach could be used.

One of the big advantages is that this approach cannot silently fail. If `__ImageBase` is missing there'll be a compiler error, if its address is wrong, window creation will fail due to an invalid `HINSTANCE`. I believe it would be worth it to test this feature and should there be any issues this commit can be instantly reverted and we can get back to the drawing board and figure out a hybrid approach.

maroider (commented):
I think this is mostly fine. I haven't tested it myself, but I think this should be a case of "if it compiles, it works".

src/platform_impl/windows/util.rs
```diff
@@ -205,6 +208,21 @@ pub fn is_focused(window: HWND) -> bool {
...
+
+    // This is preferred over GetModuleHandle(NULL) because it also works in DLLs:
+    // https://stackoverflow.com/questions/21718027/getmodulehandlenull-vs-hinstance
+
+    extern "C" {
+        static __ImageBase: IMAGE_DOS_HEADER;
+    }
+
+    unsafe { &__ImageBase as *const IMAGE_DOS_HEADER as _ }
```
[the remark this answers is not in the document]
AronParker: That gives me the following error:

```
211 | pub fn get_instance_handle() -> HINSTANCE {
    |                                 --------- expected `isize` because of return type
...
223 |     unsafe { <*const _>::cast(&__ImageBase) }
    |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `isize`, found *-ptr```

`windows_sys` redefines `HINSTANCE` to be isize, so I guess you'd need to cast a reference as isize if that works somehow, good idea though!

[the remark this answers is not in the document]
maroider: Ah, right, it's an `isize` and not a pointer. My bad.

Of the options you presented, I think the last one is the "best".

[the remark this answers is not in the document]
AronParker: Fair enough, I pushed a commit to change it.

maroider (approved):

msiglreith (approved):
thanks!

PR #2303, opened 2022-05-26
Reorganize EventLoopBuilder::build() platform documentation.

kpreid:
This PR modifies documentation only; I tested it by viewing local `cargo doc` output.

Since there's a "Platform-specific" header, it makes sense to put the Linux-specific part under it. On the other hand, "Can only be called on the main thread." is true for all platforms, and documented in the preceding text, not just iOS, so there is no reason to call it out for iOS specifically.

A further improvement might be to describe the specific side-effects calling the function could have on non-Linux platforms (rather than just saying "display backend initialisation"), but I don't know of any to mention.

### DIFF
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -84,14 +84,14 @@ impl<T> EventLoopBuilder<T> {
     /// in the relevant `platform` module if the target platform supports creating an event loop on
     /// any thread.
     ///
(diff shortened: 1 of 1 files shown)


maroider (approved):
I don't really have anything to add, but I like the change.

Issue #2293, opened 2022-05-18
Evolution of the Android backend


rib:
Hi,

_Sorry in advance this is a fairly long read :) This this aims to open a discussion as much as it does try to highlight some practical issues I found recently while working with the Android backend. The winit Discussions don't seem super active and since I also had some practical issues to bring up figured I'd post here, but happy to try and split this up if it makes sense..._

I've recently been experimenting with an alternative "glue" layer for building native Rust applications on Android that's based on the `GameActivity` class (in turn based on the widely used `AppCompatActivity` class) that's part of Google's Android Game Development Kit: 

Ref: https://developer.android.com/games/agdk
and https://developer.android.com/games/agdk/integrate-game-activity 

Here's the glue layer I've implemented so far: https://github.com/rib/agdk-rust/tree/main/game-activity

The README there also has a bit more context, including a bit about my motivation for looking at this but essentially `NativeActivity` isn't practically usable in many circumstances (at least not without subclassing in Java to augment with additional functionality, but even then it's impossible to use `AppCompatActivity` which has become a very standard foundational class for Android development that provides numerous back-ported compatibility APIs on Android).

I originally aimed to base the game-activity glue API on [ndk-glue](https://github.com/rust-windowing/android-ndk-rs/tree/master/ndk-glue) but as I made progress I found I needed to diverge in a number of ways from the ndk-glue API, such as for robustly handling synchronization between the java main thread and native thread, and I also chose to define a standard `extern "C" android_main()` entry point ABI for applications that doesn't require any special macros.

So although I originally imagined I might be able to enable winit support simply by being a drop in alternative to `ndk-glue` I ended up working on more substantial winit backend changes to be able to build on this work further.

Just for reference at this point I've published these three minimal Android app examples based on this work, to test creating a winit + wgpu app and a winit-based egui app. I have an experimental Bevy app I've been poking at locally but for now Bevy doesn't really work on Android due to it's lack of handling of lifecycle events.

Please find my example test apps here:
https://github.com/rib/agdk-rust/tree/main/examples

Please find my initial Android backend for Winit here: https://github.com/rib/winit/tree/agdk-game-activity

Now it's a bit tricky to determine the next steps for this work, considering that ideally I'd like to be able to enable this functionality in winit upstream, not just a private branch.

For starters I figured it would be good to share the current status of this work, and then I was thinking I'd list some of the most actionable topics that have come to light in the process of working on this that might help determine some next steps...

# Technical backend issues that were noticed:

[...]

Issue #2185, opened 2022-02-06
Documentation on `Event::{Suspended, Resumed}` is lacking


teryror:
I'm in the early stages of building a desktop application, and trying to figure out which events I need to handle, and how. In general, I'd like the documentation to include an overview of "events to handle to achieve basic behavior end users would expect", but as it is, I'm just looking at the page for the `winit::event::Event` type. While not ideal, that would be fine, however:

The single-sentence descriptions of the `Suspended` and `Resumed` events aren't very helpful. I tried logging when receiving them, but I couldn't get them to fire at all (on Ubuntu 21.10 running X11). I initially expected them in response to minimizing and re-opening the window, respectively, but evidently that's not how they work. Multiple google searches didn't help, so I ended up looking at the `winit` source code. Based on that, it seems these events are only emitted on iOS and Android, but I would like confirmation that this is correct.

There are more questions left open by the documentation - even if I'm right, and these event types don't matter to my use case. Here's what I would assume these events behave like, based on pretty much nothing but their names:

* They strictly alternate, i.e. ignoring all other events, an `Event::Suspended` must be followed by an `Event::Resumed` and vice-versa.
* After the application is `Suspended`, it won't receive any events until it is `Resumed`.
* (Assuming they are indeed exclusive to mobile platforms) `Suspended` fires when the user switches to another app (or the home screen), but not when it is killed. `Resumed` fires when it is reopened later.

It's also unclear when they can even fire relative to the other event types - maybe include them in the "traditional event-handling loop" reference code sample in the documentation for the `winit::event` module? Since I don't care about mobile platforms, I can't be bothered to reverse engineer how their respective platform layers work, but either way, I'd expect all of this to be clearly communicated _somewhere_. (Maybe it is, just in a very non-obvious place?)

Thanks in advance!

rib:
I've recently been working on using winit on Android, (in particular experimenting with an alternative "glue" layer based on the GameActivity class that's part of Google's Android Game Development Kit) and have found there are a few particular details that make application portability to Android tricky right now.

The main inconsistency I've seen with Android is that applications must be prepared to re-create their surface state each time the application is `Resumed` and drop their surface state when it's `Suspended`. This is required due to how the OS actually destroys/creates the native window for the application whenever it's paused.

[...]

PR #2307, opened 2022-05-28
android: Hold `NativeWindow` lock until after notifying the user with `Event::Suspended`

MarijnS95:
This applies https://github.com/rust-windowing/android-ndk-rs/issues/117 on the `winit` side: Android destroys its window/surface as soon as the user returns from [`onNativeWindowDestroyed`], and we "fixed" this on the `ndk-glue` side by sending the `WindowDestroyed` event before locking the window and removing it: this lock has to wait for any user of `ndk-glue` - ie. `winit` - to give up its readlock on the window, which is what we utilize here to give users of `winit` "time" to destroy any resource created on top of a `RawWindowHandle`.

since we can't pass the user a `RawWindowHandle` through the `HasRawWindowHandle` trait we have to document this case explicitly and keep the lock alive on the `winit` side instead.

[`onNativeWindowDestroyed`]: https://developer.android.com/ndk/reference/struct/a-native-activity-callbacks#onnativewindowdestroyed

---

I'm creating this as draft because there are different things to consider: for one I've constantly commented that we should probably pass the `RawWindowHandle` and/or Android `NativeWindow` inside the `Resumed` event _somehow_ **across every platform** for consistency, and it'd be great if this includes the lock to make the user responsible for holding it instead of implicitly unlocking after `Event::Suspended`.  Then someone recently created a whole bunch of very long issues mentioning similar (concurrency) issues and longevity of the Android backend in general - sorry I've yet to catch up to those but there are too many and they're way too long 😁

### DIFF
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -10,8 +10,9 @@ use ndk::{
     configuration::Configuration,
     event::{InputEvent, KeyAction, Keycode, MotionAction},
     looper::{ForeignLooper, Poll, ThreadLooper},
+    native_window::NativeWindow,
 };
-use ndk_glue::{Event, Rect};
+use ndk_glue::{Event, LockReadGuard, Rect};
 use once_cell::sync::Lazy;
 use raw_window_handle::{HasRawWindowHandle, RawWindowHandle};
 
@@ -239,6 +240,7 @@ pub struct EventLoop<T: 'static> {
     start_cause: event::StartCause,
     looper: ThreadLooper,
     running: bool,
+    window_lock: Option<LockReadGuard<NativeWindow>>,
 }
 
 #[derive(Default, Debug, Copy, Clone, PartialEq, Eq, Hash)]
@@ -268,6 +270,7 @@ impl<T: 'static> EventLoop<T> {
             start_cause: event::StartCause::Init,
             looper: ThreadLooper::for_thread().unwrap(),
             running: false,
+            window_lock: None,
         }
     }
 
@@ -300,22 +303,42 @@ impl<T: 'static> EventLoop<T> {
             match self.first_event.take() {
                 Some(EventSource::Callback) => match ndk_glue::poll_events().unwrap() {
                     Event::WindowCreated => {
-                        call_event_handler!(
-                            event_handler,
-                            self.window_target(),
-                            control_flow,
-                            event::Event::Resumed
-                        );
+                        // Acquire a lock on the window to prevent Android from destroying
+                        // it until we've notified and waited for the user in Event::Suspended.
+                        // WARNING: ndk-glue is inherently racy (https://github.com/rust-windowing/winit/issues/2293)
+                        // and may have already received onNativeWindowDestroyed while this thread hasn't yet processed
+                        // the event, and would see a `None` lock+window in that case.
+                        if let Some(next_window_lock) = ndk_glue::native_window() {
+                            assert!(
+                                self.window_lock.replace(next_window_lock).is_none(),
+                                "Received `Event::WindowCreated` while we were already holding a lock"
+                            );
+                            call_event_handler!(
+                                event_handler,
+                                self.window_target(),
+                                control_flow,
+                                event::Event::Resumed
+                            );
+                        } else {
+                            warn!("Received `Event::WindowCreated` while `ndk_glue::native_window()` provides no window");
+                        }
                     }
                     Event::WindowResized => resized = true,
                     Event::WindowRedrawNeeded => redraw = true,
                     Event::WindowDestroyed => {
-                        call_event_handler!(
-                            event_handler,
-                            self.window_target(),
-                            control_flow,
-                            event::Event::Suspended
-                        );
+                        // Release the lock, allowing Android to clean up this surface
+                        // WARNING: See above - if ndk-glue is racy, this event may be called
+                        // without having a `self.window_lock` in place.
+                        if self.window_lock.take().is_some() {
+                            call_event_handler!(
+                                event_handler,
+                                self.window_target(),
+                                control_flow,
+                                event::Event::Suspended
+                            );
+                        } else {
+                            warn!("Received `Event::WindowDestroyed` while we were not holding a window lock");
+                        }
                     }
                     Event::Pause => self.running = false,
                     Event::Resume => self.running = true,
@@ -369,7 +392,7 @@ impl<T: 'static> EventLoop<T> {
                 },
                 Some(EventSource::InputQueue) => {
                     if let Some(input_queue) = ndk_glue::input_queue().as_ref() {
-                        while let Some(event) = input_queue.get_event() {
+                        while let Some(event) = input_queue.get_event().expect("get_event") {
                             if let Some(event) = input_queue.pre_dispatch(event) {
                                 let mut handled = true;
                                 let window_id = window::WindowId(WindowId);
@@ -799,7 +822,7 @@ impl Window {
     }
 
     pub fn raw_window_handle(&self) -> RawWindowHandle {
-        if let Some(native_window) = ndk_glue::native_window().as_ref() {
+        if let Some(native_window) = ndk_glue::native_window() {
             native_window.raw_window_handle()
         } else {
             panic!("Cannot get the native window, it's null and will always be null before Event::Resumed and after Event::Suspended. Make sure you only call this function between those events.");
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- On Android, `ndk-glue`'s `NativeWindow` lock is now held between `Event::Resumed` and `Event::Suspended`.
 - On Web, added `EventLoopExtWebSys` with a `spawn` method to start the event loop without throwing an exception.
 - Added `WindowEvent::Occluded(bool)`, currently implemented on macOS and X11.
 - On X11, fix events for caps lock key not being sent
@@ -64,7 +65,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - Implemented `Eq` for `Fullscreen`, `Theme`, and `UserAttentionType`.
 - **Breaking:** `Window::set_cursor_grab` now accepts `CursorGrabMode` to control grabbing behavior.
 - On Wayland, add support for `Window::set_cursor_position`.
-- Fix on macOS `WindowBuilder::with_disallow_hidpi`, setting true or false by the user no matter the SO default value. 
+- Fix on macOS `WindowBuilder::with_disallow_hidpi`, setting true or false by the user no matter the SO default value.
 - `EventLoopBuilder::build` will now panic when the `EventLoop` is being created more than once.
 - Added `From<u64>` for `WindowId` and `From<WindowId>` for `u64`.
 - Added `MonitorHandle::refresh_rate_millihertz` to get monitor's refresh rate.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,8 +55,8 @@ simple_logger = "2.1.0"
 
 [target.'cfg(target_os = "android")'.dependencies]
 # Coordinate the next winit release with android-ndk-rs: https://github.com/rust-windowing/winit/issues/1995
-ndk = { git = "https://github.com/rust-windowing/android-ndk-rs", rev = "7e33384" }
-ndk-glue = { git = "https://github.com/rust-windowing/android-ndk-rs", rev = "7e33384" }
+ndk = { git = "https://github.com/rust-windowing/android-ndk-rs", rev = "a0c5e99" }
+ndk-glue = { git = "https://github.com/rust-windowing/android-ndk-rs", rev = "a0c5e99" }
 
 [target.'cfg(any(target_os = "ios", target_os = "macos"))'.dependencies]
 objc = "0.2.7"
--- a/src/window.rs
+++ b/src/window.rs
@@ -1039,8 +1039,17 @@ unsafe impl raw_window_handle::HasRawWindowHandle for Window {
     ///
     /// ## Platform-specific
     ///
-    /// - **Android:** Only available after receiving the Resumed event and before Suspended. *If you*
-    /// *try to get the handle outside of that period, this function will panic*!
+    /// ### Android
+    ///
+    /// Only available after receiving [`Event::Resumed`] and before [`Event::Suspended`]. *If you
+    /// try to get the handle outside of that period, this function will panic*!
+    ///
+    /// Make sure to release or destroy any resources created from this `RawWindowHandle` (ie. Vulkan
+    /// or OpenGL surfaces) before returning from [`Event::Suspended`], at which point Android will
+    /// release the underlying window/surface: any subsequent interaction is undefined behavior.
+    ///
+    /// [`Event::Resumed`]: crate::event::Event::Resumed
+    /// [`Event::Suspended`]: crate::event::Event::Suspended
     fn raw_window_handle(&self) -> raw_window_handle::RawWindowHandle {
         self.window.raw_window_handle()
     }


rib:
> then someone recently created a whole bunch of very long issues mentioning similar (concurrency) issues and longevity of the Android backend in general - sorry I've yet to catch up to those but there are too many and they're way too long

hehe, yeah that'll be me, sorry :)

TL;DR = I don't think ndk-glue handles native <-> Java synchronization robustly at the moment.

Ideally I think it could be good for `ndk-glue` to provide a callback based API for polling + dispatching events which can handle synchronization in a way that's similar to the original `android_native_app_glue` that's ships with the Android NDK where the glue can insert "pre" and "post" code around callbacks that let `ndk-glue` encapsulate this kind of synchronization so downstream users don't have to follow this kind of convention to get correct synchronization.

I feel like this is an error-prone design at the moment, that pushes the fiddly burden of handling synchronization to downstream users (winit) while also exposing implementation details that could ideally be encapsulated in the glue layer (now winit will depend on parking_lot because that's how ndk_glue internally locks the window state, and ndk_glue now can't change how it handles synchronization without breaking it's API)

I've tried to solve this with how I designed these [native-activity](https://github.com/rib/agdk-rust/tree/main/native-activity) and [game-activity](https://github.com/rib/agdk-rust/tree/main/game-activity) glue crates, including also handling synchronization for saving app state.

Those concerns aside for a moment though; the changes in this draft look good for being able to address the main synchronization issue I've seen with ndk-glue for window surface destruction.

MarijnS95:
> TL;DR = I don't think ndk-glue handles native <-> Java synchronization robustly at the moment.

TL;DR: I can't but agree.

In principle NDK glue looks like a straightforward, concise wrapper. Using a `pipe` to tie into `looper`, and using locks to synchronize objects created and destroyed in its callbacks. Consumers of `ndk-glue` - in particular `winit` which lives under the same orga - are probably designed to tightly work with/around the design of `ndk-glue`, and hide its implementation details. That didn't really seem to have happened, and the Android backend in `winit` feels like it works "just good enough" and the missing features are only slowly trickling in when users finally need it (ie. it doesn't seem to be used often).

In the end `ndk-glue` could have also provided a (callback) layer on top of the looper, pipe, and locks, instead of having this implemented inside `winit` and any other possible consumer of `ndk-glue`. Combine that with `winit`s "poor" (because desktop differs here) handling of "dynamic" windows that "randomly" (dis)appear and require platform-specific `Suspended` and `Resumed` events and the oddities of the Android API trickle down all the way into user code.  But that's a discussion better left for #2185/#2293.

> so downstream users don't have to follow this kind of convention to get correct synchronization.

For the record I have yet to find a primitive that helps users be aware that they **must** handle a `WindowDestroyed`/`Suspended` event in terms of destroying their Vulkan/OpenGL surfaces, before returning/continuing from the handling of this event.

> now winit will depend on parking_lot because that's how ndk_glue internally locks the window state, and ndk_glue now can't change how it handles synchronization without breaking it's API

`parking_lot` is not necessary for this change. It simply makes the API more convenient by not having an `Option` inside the `ReadLock` that you know will never be `None` after initially checking it (fwiw many other platform backends have a `parking_lot` dependency as well, this locking crate appears to be widely accepted).

> Those concerns aside for a moment though; the changes in this draft look good for being able to address the main synchronization issue I've seen with ndk-glue for window surface destruction.

Yes, I'm surprised we fixed this in `ndk-glue` but the change wasn't yet applied to `winit`... Priorities I guess.

What's your thought: get this in now, and keep discussing the future with your game/native-activity wrappers and see where we end up time-wise? There are currently some tricky issues with `ndk-glue` which'll take time to resolve that I don't have; at the same time your crate may solve a bunch of those implicitly but will also take quite some time/effort to discuss and migrate to? Most importantly updating the user-facing `winit` part to better suit Android instead of feeling and behaving like an afterthought.

MarijnS95:
Addendum: It's funny that I just now started reading #2293 and addressed your main concern with this PR, while I was really doing that after noticing that it needed to be done in `glutin` because `winit` didn't do it yet. The comment you quoted came from another `ndk-glue` - but not `winit` - user who saw the same concerns and I just forgot to update `winit` after fixing it there.

_However_.

I may have done that "on purpose".

In certain cases the caller may invoke `ndk_glue::native_window()`, and themselves be responsible for storing the lock next to their handle on a Vulkan/OpenGL surface, instead of `winit`.

At least this is what I initially thought when voicing these concerns in https://github.com/rust-windowing/glutin/pull/1411#discussion_r883647764, where this thought is valid.

_But then_.

It is more common for applications to use the `HasRawWindowHandle` trait on a `winit::Window`, perhaps even the `glutin` implementation should use that without ever depending on the `ndk_glue` crate directly.

---

Summing that up, this PR is probably the right thing to do after all given that users of `winit` don't use `ndk_glue` hence don't see the lock as that's now hidden behind the `Suspended`/`Resumed` events. Otherwise there's no harm in holding the readlock twice.

rib:
> For the record I have yet to find a primitive that helps users be aware that they **must** handle a `WindowDestroyed`/`Suspended` event in terms of destroying their Vulkan/OpenGL surfaces, before returning/continuing from the handling of this event.

Right, I think this is a big enough portability hazard that I think it needs to be addressed at a higher level than just the Android backend. I don't tend to think there's a simple primitive we can find to solve this within the Android backend.

With issue https://github.com/rust-windowing/winit/pull/2182 and https://github.com/rust-windowing/winit/issues/2293 I think I'd currently like to advocate for Winit supporting resume events consistently across all platforms, and then either just updating all examples and potentially introducing some impediments on other platforms that _strongly_ encourage following a pattern where GPU resources like your context and first window + surface are only lazily allocated once you reach your first resume event.

I recently worked on two downstream projects to investigate the feasibility of this and I was able to get both egui and Bevy running based on this model, in such a way that it also worked to run on Windows following the same model.

E.g. this Egui example, based on Winit, runs on Android and Windows: https://github.com/rib/agdk-rust/blob/main/examples/agdk-egui/src/lib.rs
and here's a proof of concept for Bevy too: https://github.com/rib/bevy/tree/android-deferred-render-init (I had a local Winit branch that generates Resumed events on Windows so the same approach ran on Windows)
I'd say that Bevy was _by far_ the more complex architecture to adapt, but in the end it wasn't too big of a change.

rib:
> `parking_lot` is not necessary for this change. It simply makes the API more convenient by not having an `Option` inside the `ReadLock` that you know will never be `None` after initially checking it (fwiw many other platform backends have a `parking_lot` dependency as well, this locking crate appears to be widely accepted).

Yep, and to be clear, I'm not making any kind of value judgement on parking lot here which is definitely widely used and yeah it makes sense that other backends are also using it. I'm more just concerned with the leaky `ndk-glue` abstraction. The choice of locking and synchronization primitives can ideally be encapsulated within a library. In this case `ndk-glue` now has a public API that imposes a parking_lot dependency on its downstream users and it also means ndk_glue can't simply switch to a different synchronization strategy without breaking its public API, which doesn't seem ideal to me.

rib:
> What's your thought: get this in now, and keep discussing the future with your game/native-activity wrappers and see where we end up time-wise? There are currently some tricky issues with `ndk-glue` which'll take time to resolve that I don't have; at the same time your crate may solve a bunch of those implicitly but will also take quite some time/effort to discuss and migrate to? Most importantly updating the user-facing `winit` part to better suit Android instead of feeling and behaving like an afterthought.

My gut instinct atm is that there's no reason to block a fix like this on any other proposal to potentially swap out `ndk_glue`; this should hopefully fix a legitimate synchronization issue that exists now, and the concerns around leaky API etc don't extend outside of the Android backend so they don't affect users of the Winit API.

Thinking ahead; I'm currently thinking of creating a shim `android_activity` crate (I was considering renaming my `agdk-rust` repo to `android-activity`) that would provide a single place that can allow Rust applications to configure their Activity glue layer. For starters the crate would optionally depend on the `native-activity` and `game-activity` crates and let you pick a glue layer based on a feature. Additional glue layers could be implemented to support different Activity base classes (E.g. there were some ideas to create a `RustActivity` crate suggested here: https://github.com/rust-windowing/android-ndk-rs/issues/266#issuecomment-1105891391).

This would result in a single crate + API for Winit to use for Android glue while giving applications the ability to choose different Activity base classes. Right now my experimental Winit backend conditionally switches between `native-activity` and `game-activity` based on a feature in Winit (since the APIs are compatible it just needs a `#[cfg()]` test around the `use *-activity as ndk_glue` currently), but I was thinking it might be beneficial to push that out to a separate crate in case there are opportunities to share code for different activity classes and to allow for non-winit use cases that could also benefit from a common API.

Also just for reference, I'm not expecting that `android_activity` would have to be a fully encapsulated, lowest-common-denominator API for different activity base classes (since otherwise we wouldn't be able to support any additional Android features in Winit beyond what NativeActivity allows). I was thinking there would be a core API that's common across activity classes (such as `poll_events()` and the input API and APIs for saving state + querying data_dirs) and then the features for selecting the implementation would also be able to expose APIs that may be specific to that activity class.

For the more general challenge with Android being a kind of second-class platform atm then I mainly thinking about trying to normalize lazy graphics state initialization via resumed events across platforms as a starting point, re: https://github.com/rust-windowing/winit/pull/2307#issuecomment-1140827415

rib:
> _But then_.
> 
> It is more common for applications to use the `HasRawWindowHandle` trait on a `winit::Window`, perhaps even the `glutin` implementation should use that without ever depending on the `ndk_glue` crate directly.
> 
> Summing that up, this PR is probably the right thing to do after all given that users of `winit` don't use `ndk_glue` hence don't see the lock as that's now hidden behind the `Suspended`/`Resumed` events. Otherwise there's no harm in holding the readlock twice.

hmm, yeah, I didn't realize glutin was currently dependent on `ndk_glue` directly. Especially considering this kind of locking design in ndk_glue I would think it's probably for the best to at least aim to keep the synchronization responsibilities encapsulated inside Winit, and would guess it'd be best for glutin to be getting it's raw window handles via Winit.

MarijnS95:
> > For the record I have yet to find a primitive that helps users be aware that they **must** handle a `WindowDestroyed`/`Suspended` event in terms of destroying their Vulkan/OpenGL surfaces, before returning/continuing from the handling of this event.
> 
> Right, I think this is a big enough portability hazard that I think it needs to be addressed at a higher level than just the Android backend. I don't tend to think there's a simple primitive we can find to solve this within the Android backend.

In case it wasn't entirely clear I'm specifically looking for a solution that forces the API to be consistent across all platforms, **to replace** the Android-specific after-thought that we have now. In addition something that more clearly pushes the burden of managing the implicit dependency between Vulkan/GL surfaces and a window.

However, I initially limited my thoughts to `RawWindowHandle` in `Resumed` (and removing the getter/trait from `Window`), because `Window` is pretty much an empty shell on Android that I don't look at much. But for true platform agnostic behaviour things presumably need to change across the board.

For one I remember some getter functions on `Window` not behaving correctly prior to `Resumed` (IIRC it was the window size, which isn't surprising) but will need to recheck that before commenting further.

> With issue #2182 and #2293 I think I'd currently like to advocate for Winit supporting resume events consistently across all platforms, and then either just updating all examples and potentially introducing some impediments on other platforms that _strongly_ encourage following a pattern where GPU resources like your context and first window + surface are only lazily allocated once you reach your first resume event.

Exactly this. Slightly more burden for "desktop app" writers, but they'll "thank us" when their app runs flawless on mobile, instead of finding out through obscure errors that they need to handle this differently.

---

> Yep, and to be clear, I'm not making any kind of value judgement on parking lot here which is definitely widely used and yeah it makes sense that other backends are also using it. I'm more just concerned with the leaky `ndk-glue` abstraction. The choice of locking and synchronization primitives can ideally be encapsulated within a library. In this case `ndk-glue` now has a public API that imposes a parking_lot dependency on its downstream users and it also means ndk_glue can't simply switch to a different synchronization strategy without breaking its public API, which doesn't seem ideal to me.

I guess we could newtype-wrap it just to get this issue done and over with, at which point we're free to implement whatever locking abstraction we want.

---

> My gut instinct atm is that there's no reason to block a fix like this on any other proposal to potentially swap out `ndk_glue`; this should hopefully fix a legitimate synchronization issue that exists now, and the concerns around leaky API etc don't extend outside of the Android backend so they don't affect users of the Winit API.
>
> ...
>
> For the more general challenge with Android being a kind of second-class platform atm then I mainly thinking about trying to normalize lazy graphics state initialization via resumed events across platforms as a starting point, re: [#2307 (comment)](https://github.com/rust-windowing/winit/pull/2307#issuecomment-1140827415)

I guess we should talk to the `winit` maintainers for a timeline on the next release. Specifically I'd like to get these changes merged but without a git dependency - though at the same time that forces the next `winit` release to coordinate with an NDK/ndk_glue breaking release as requested in https://github.com/rust-windowing/winit/issues/1995.

On the side, in separate issues that don't fill up this PR, we should discuss how to proceed on making Android a first-class citizen, how to lay out the underlying crates and architecture, etc.

---

> hmm, yeah, I didn't realize glutin was currently dependent on `ndk_glue` directly. Especially considering this kind of locking design in ndk_glue I would think it's probably for the best to at least aim to keep the synchronization responsibilities encapsulated inside Winit, and would guess it'd be best for glutin to be getting it's raw window handles via Winit.

We've already shoved `ndk_glue` out as soon as it came in, since this should be handled through `raw-window-handle` instead: https://github.com/rust-windowing/glutin/pull/1418

Over time (I don't use `glutin`, so progress only happens in down-time) I'm pushing towards working `winit` out the door too: `raw-window-handle` was specifically designed for this, allows for a more generic implementation (just a `match` block) across all platforms, and will allow arbitrary Android `Surface`s (`NativeWindow`s...) to be used as render target (https://github.com/rust-windowing/android-ndk-rs/pull/274).

MarijnS95:
@rib I like the newtype suggestion, see https://github.com/rust-windowing/android-ndk-rs/pull/288. This is now all ready from my side, and probably the last change we can do before pushing out `0.27`. I wish we could do a larger `Suspended`/`Resumed` refactor, but that'll probably be a larger endeavour requiring precise API design.

src/platform_impl/android/mod.rs
```diff
@@ -299,6 +302,17 @@ impl<T: 'static> EventLoop<T> {
             match self.first_event.take() {
                 Some(EventSource::Callback) => match ndk_glue::poll_events().unwrap() {
                     Event::WindowCreated => {
+                        // Acquire a lock on the window to prevent Android from destroying
+                        // it until we've notified and waited for the user in Event::Suspended.
+                        // XXX: Don't send this event if Android already removed the window
+                        // before we managed to handle this event?
```
rib, line 308: Right; ndk-glue's design is still fundamentally unsynchronized so it looks like there is still a race condition here and you could potentially get a panic through no fault of the application.

I could vaguely imagine a situation where you switch to an app and then hit something like 'home' which could immediately suspend the app and if the device was pretty busy with destroying another app and doesn't really have time to think it seems plausible an app could get a window create + terminated delivered before the OS ends up scheduling the rust thread if you're unlucky.

This is still going to be better than the current situation I guess, so maybe still worth landing in the short term.

Maybe this comment should instead highlight that there's still a race condition here?

Maybe it should be possible to "try" and get the native window instead, so the race could be handled more gracefully without a panic?

MarijnS95:
> This applies [rust-windowing/android-ndk-rs#117](https://github.com/rust-windowing/android-ndk-rs/issues/117) on the `winit` side

It is thanks to @rib's comment that links to these PRs, that I found #1892 again which _did end up applying https://github.com/rust-windowing/android-ndk-rs/issues/117 to `winit`_. It wasn't forgotten about, I just forgot about that PR being open and ending up stale...

src/platform_impl/android/mod.rs
MarijnS95: > Maybe this comment should instead

Hence `XXX`.

msiglreith (approved):

maciejgodek:
#1892 author here, glad this got fixed eventually! Apologies for abandoning the PR, my priorities changed for a while. :)

MarijnS95:
@maciejgodek no worries, I think this code even ended up being replaced with a completely different `android-activity` backend now. Feel free to close the PR to that effect though, as there's nothing actionable anymore.

PR #2313, opened 2022-06-05
Migrate from lazy_static to once_cell

james7132:

## Motivation
`lazy_static!`, while a declarative macro, is a macro nonetheless. It can add quite a bit of additional compilation time cost. `once_cell::sync::Lazy` does the same thing with generics, and can be used more flexibly (i.e. non-static lazily initialized values), and has been [proposed to be added to `std`](https://github.com/rust-lang/rust/issues/74465).

I'm trying to reduce the compile time and dependency tree complexity of a dependent project: [bevy](https://bevyengine.org), which is using winit. `lazy_static` and `once_cell` are both in our dependency tree and both end up doing the same thing.

## Solution
Migrate to `once_cell`.

---

I don't think any of the changes here are user visible beyond the dependency in the tree, and should work as-is on all of them, but I could be wrong.

### DIFF
--- a/src/platform_impl/linux/mod.rs
+++ b/src/platform_impl/linux/mod.rs
@@ -14,8 +14,6 @@ use crate::window::Theme;
 #[cfg(feature = "wayland")]
 use std::error::Error;
 
-#[cfg(feature = "x11")]
-use once_cell::sync::Lazy;
 use std::{collections::VecDeque, env, fmt};
 #[cfg(feature = "x11")]
 use std::{ffi::CStr, mem::MaybeUninit, os::raw::*, sync::Arc};
@@ -23,6 +21,8 @@ use std::{ffi::CStr, mem::MaybeUninit, os::raw::*, sync::Arc};
 #[cfg(feature = "x11")]
 use parking_lot::Mutex;
 use raw_window_handle::RawWindowHandle;
+#[cfg(feature = "x11")]
+use once_cell::sync::Lazy;
 
 #[cfg(feature = "x11")]
 pub use self::x11::XNotSupported;
--- a/src/platform_impl/linux/mod.rs
+++ b/src/platform_impl/linux/mod.rs
@@ -19,10 +19,10 @@ use std::{collections::VecDeque, env, fmt};
 use std::{ffi::CStr, mem::MaybeUninit, os::raw::*, sync::Arc};
 
 #[cfg(feature = "x11")]
+use once_cell::sync::Lazy;
+#[cfg(feature = "x11")]
 use parking_lot::Mutex;
 use raw_window_handle::RawWindowHandle;
-#[cfg(feature = "x11")]
-use once_cell::sync::Lazy;
 
 #[cfg(feature = "x11")]
 pub use self::x11::XNotSupported;
--- a/src/platform_impl/linux/mod.rs
+++ b/src/platform_impl/linux/mod.rs
@@ -14,6 +14,8 @@ use crate::window::Theme;
 #[cfg(feature = "wayland")]
 use std::error::Error;
 
+#[cfg(feature = "x11")]
+use once_cell::sync::Lazy;
 use std::{collections::VecDeque, env, fmt};
 #[cfg(feature = "x11")]
 use std::{ffi::CStr, mem::MaybeUninit, os::raw::*, sync::Arc};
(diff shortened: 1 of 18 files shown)


src/platform_impl/linux/x11/util/wm.rs
```diff
@@ -1,12 +1,12 @@
 use parking_lot::Mutex;
 
 use super::*;
+use once_cell::sync::Lazy;
```
filnet, line 4: move it up to be consistent with other similar changes ?

madsmtm (approved):
Sure, if it helps you. This is likely going to be closer to the final `std` API anyhow, and then the switch will be easier when we get there.

src/platform_impl/macos/app_state.rs
```diff
@@ -40,10 +30,18 @@ use crate::{
...
+use cocoa::{
+    appkit::{NSApp, NSApplication, NSWindow},
+    base::{id, nil},
+    foundation::NSSize,
+};
+use objc::{
+    rc::autoreleasepool,
+    runtime::{Object, BOOL, NO, YES},
+};
+use once_cell::sync::Lazy;
```
maroider, line 42: Why did you move this below the `crate`-level imports?

src/platform_impl/macos/window.rs
```diff
@@ -47,6 +35,18 @@ use objc::{
...
+use std::{
+    collections::VecDeque,
+    convert::TryInto,
+    f64, ops,
+    os::raw::c_void,
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        Arc, Mutex, MutexGuard, Weak,
+    },
+};
```
maroider, line 49: ditto

maroider (commented):
Mostly nitpicks from me. I don't quite get why you've grouped "external crate" imports with/above `std` imports. I was under the impression that `std -> extern crate -> local crate` is the "normal" way of grouping imports (or at least quite popular, as it's the only option besides "preserve" in [rustfmt's configuration](https://rust-lang.github.io/rustfmt/?version=v1.4.38&search=#group_imports)).

src/platform_impl/linux/x11/monitor.rs
```diff
@@ -14,13 +14,12 @@ use crate::{
     monitor::{MonitorHandle as RootMonitorHandle, VideoMode as RootVideoMode},
     platform_impl::{MonitorHandle as PlatformMonitorHandle, VideoMode as PlatformVideoMode},
 };
+use once_cell::sync::Lazy;
```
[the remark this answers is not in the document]
kchibisov: It should be with `use parking::lot`, since it's an external crate.

src/platform_impl/linux/x11/ime/input_method.rs
```diff
@@ -1,3 +1,5 @@
+use once_cell::sync::Lazy;
+use parking_lot::Mutex;
```
kchibisov, line 2: Should move below std.

kchibisov, line 2: Should move below std.

src/platform_impl/linux/mod.rs
```diff
@@ -14,6 +14,8 @@ use crate::window::Theme;
 #[cfg(feature = "wayland")]
 use std::error::Error;
 
+#[cfg(feature = "x11")]
+use once_cell::sync::Lazy;
```
kchibisov, line 18: move to `raw_window_handle`

kchibisov (changes requested):
> I don't quite get why you've grouped "external crate" imports with/above std imports. I was under the impression that std -> extern crate -> local crate

I agree with @maroider . external crates should be separate from std. It just winit doesn't have formatting that way across the code base, but it doesn't mean that we should break it even more.

src/platform_impl/macos/app_state.rs
james7132: I ran `cargo fmt --all`, which I think automatically did this. I'll go ahead and undo this.

james7132:
Should have fixed the spurious moves as a result of `cargo fmt --all`, including some of the aforementioned broken ones.

Is there a way to add this via a `rustfmt.toml`? Seems like a trivial thing for any contributor to just run the command and undo all of these.

madsmtm:
> Is there a way to add this via a `rustfmt.toml`? Seems like a trivial thing for any contributor to just run the command and undo all of these.

There will be, it's just not stable yet.

maroider (approved):
Everything seems OK now. Like, @madsmtm, I'm of the opinion that we should move to the `std` API when that stabilizes, and this will make that switch easier.

MarijnS95:
@kchibisov I don't really mind which `rustfmt` to use, given that `stable` seems to work _iff the config option is passed on the cmdline_.

Would you do the honors and PR that `rustfmt` from `alacritty` here?

PR #2314, opened 2022-06-05
macOS Add background color for Window

rappet:
Changing the background color on macOS is most useful for
changing the title bar color, as the content of the window will most likely be overwritten by the user.

- Add `WindowExtMacOS::set_background_color`
- Add `WindowBuilderExtMacOS::with_background_color`

Only changing the background color using new functionality (the title bar is also affected by this a bit):

grafik

Changing the background color and making the title bar transparent:

grafik

### DIFF
--- a/src/platform_impl/macos/window.rs
+++ b/src/platform_impl/macos/window.rs
@@ -74,6 +74,7 @@ pub struct PlatformSpecificWindowBuilderAttributes {
     pub resize_increments: Option<LogicalSize<f64>>,
     pub disallow_hidpi: bool,
     pub has_shadow: bool,
+    pub background_color: Option<(u8, u8, u8, u8)>,
 }
 
 impl Default for PlatformSpecificWindowBuilderAttributes {
@@ -89,6 +90,7 @@ impl Default for PlatformSpecificWindowBuilderAttributes {
             resize_increments: None,
             disallow_hidpi: false,
             has_shadow: true,
+            background_color: None,
         }
     }
 }
@@ -238,6 +240,18 @@ fn create_window(
             if attrs.position.is_none() {
                 ns_window.center();
             }
+
+            if let Some((r, g, b, a)) = pl_attrs.background_color {
+                let color = NSColor::colorWithSRGBRed_green_blue_alpha_(
+                    nil,
+                    r as f64 / 255f64,
+                    g as f64 / 255f64,
+                    b as f64 / 255f64,
+                    a as f64 / 255f64,
+                );
+                ns_window.setBackgroundColor_(color);
+            }
+
             ns_window
         })
     })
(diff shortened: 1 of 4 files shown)


rappet:
One question that is open on my side is, if and how winit should handle colorspaces in the future,
as I did only implement bindings for sRGB in this case and macOS can support other color spaces for this (sRGB, P3, ...).

src/platform_impl/macos/window.rs
```diff
@@ -238,6 +240,18 @@ fn create_window(
...
+
+            if let Some((r, g, b, a)) = pl_attrs.background_color {
+                let color = NSColor::colorWithSRGBRed_green_blue_alpha_(
+                    nil,
+                    r as f64 / 255f64,
+                    g as f64 / 255f64,
+                    b as f64 / 255f64,
+                    a as f64 / 255f64,
+                );
+                ns_window.setBackgroundColor_(color);
```
madsmtm, line 252: This is going to be overridden by a `setBackgroundColor_` call in `UnownedWindow::new` further down in this file, when `window.with_transparent(true)` has been called.

src/platform_impl/macos/window.rs
```diff
@@ -1241,6 +1255,20 @@ impl WindowExtMacOS for UnownedWindow {
...
+    fn set_background_color(&self, red: u8, green: u8, blue: u8, alpha: u8) {
+        autoreleasepool(|| unsafe {
+            let ns_background = NSColor::colorWithSRGBRed_green_blue_alpha_(
+                nil,
+                red as f64 / 255f64,
+                green as f64 / 255f64,
+                blue as f64 / 255f64,
+                alpha as f64 / 255f64,
+            );
+            self.ns_window.setBackgroundColor_(ns_background);
```
madsmtm, line 1269: This is very likely not thread safe, and should always be done on the main thread. As an example, try search for `set_title_async` to see how we do that.

madsmtm (changes requested):
Thanks, this probably makes sense to have.

sRGB is probably fine for now, that's also what our icons are using.

madsmtm:
This can certainly be done by an external crate, just like `set_has_shadow`, `with_title_hidden`, and a lot of the other extra functionality we expose can.

But idk if it makes sense to have in our scope or not, my initial reaction was "yeah sure" because it's quite a small feature.

Matts966:
As a [neovide](https://github.com/neovide/neovide) user, this is one of the most desireble features...

Matts966:
`NSWindow` is greatly abstracted by `winit` and it seems difficult for me to set properties from the outside of `winit` as long as we use `winit`.
Is there any workaround?

kchibisov:
You can access it just fine from anywhere, one way is https://docs.rs/winit/0.27.1/x86_64-apple-darwin/winit/platform/macos/trait.WindowExtMacOS.html#tymethod.ns_window.

Matts966:
I tried to implement in neovide.
Thank you :)

https://github.com/neovide/neovide/pull/1459

kchibisov:
Given that feature could be easily added downstream when need. I'd say it's out of scope for winit.

PR #2315, opened 2022-06-07
Run clippy on CI

### LINKED ISSUE #1402: Consider adding clippy on CI

kchibisov:
Fixes #1402.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,8 @@ jobs:
     - name: Check Formatting
       run: cargo +stable fmt --all -- --check
 
-  Tests:
+  tests:
+    name: Tests
     strategy:
       fail-fast: false
       matrix:
@@ -62,6 +63,7 @@ jobs:
       with:
         rust-version: ${{ matrix.rust_version }}${{ matrix.platform.host }}
         targets: ${{ matrix.platform.target }}
+        components: clippy
 
     - name: Install Linux dependencies
       if: (matrix.platform.os == 'ubuntu-latest')
@@ -93,6 +95,10 @@ jobs:
         !contains(matrix.platform.target, 'wasm32'))
       run: cargo $CMD test --verbose --target ${{ matrix.platform.target }} $OPTIONS --features $FEATURES
 
+    - name: Lint with clippy
+      shell: bash
+      if: (matrix.rust_version != 'nightly') && !contains(matrix.platform.options, '--no-default-features')
+      run: cargo clippy --all-targets --target ${{ matrix.platform.target }} $OPTIONS --features $FEATURES -- -Dwarnings
 
     - name: Build with serde enabled
       shell: bash
--- a/examples/fullscreen.rs
+++ b/examples/fullscreen.rs
@@ -1,3 +1,5 @@
+#![allow(clippy::single_match)]
+
 use std::io::{stdin, stdout, Write};
 
 use simple_logger::SimpleLogger;
--- a/examples/request_redraw_threaded.rs
+++ b/examples/request_redraw_threaded.rs
@@ -1,3 +1,5 @@
+#![allow(clippy::single_match)]
+
 #[cfg(not(target_arch = "wasm32"))]
 fn main() {
     use std::{thread, time};
@@ -28,10 +30,10 @@ fn main() {
         control_flow.set_wait();
 
         match event {
-            Event::WindowEvent { event, .. } => match event {
-                WindowEvent::CloseRequested => control_flow.set_exit(),
-                _ => (),
-            },
+            Event::WindowEvent {
+                event: WindowEvent::CloseRequested,
+                ..
+            } => control_flow.set_exit(),
             Event::RedrawRequested(_) => {
                 println!("\nredrawing!\n");
             }
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -94,6 +94,8 @@ impl<T> EventLoopBuilder<T> {
     ///   will fall back on X11. If this variable is set with any other value, winit will panic.
     #[inline]
     pub fn build(&mut self) -> EventLoop<T> {
+        // Certain platforms accept a mutable reference in their API.
+        #[allow(clippy::unnecessary_mut_passed)]
         EventLoop {
             event_loop: platform_impl::EventLoop::new(&mut self.platform_specific),
             _marker: PhantomData,
@@ -218,6 +220,12 @@ impl EventLoop<()> {
     #[inline]
     pub fn new() -> EventLoop<()> {
         EventLoopBuilder::new().build()
+    }
+}
+
+impl Default for EventLoop<()> {
+    fn default() -> Self {
+        Self::new()
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -132,6 +132,9 @@
 
 #![deny(rust_2018_idioms)]
 #![deny(rustdoc::broken_intra_doc_links)]
+#![deny(clippy::all)]
+#![cfg_attr(feature = "cargo-clippy", deny(warnings))]
+#![allow(clippy::missing_safety_doc)]
 
 #[allow(unused_imports)]
 #[macro_use]
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -240,7 +240,7 @@ pub struct EventLoop<T: 'static> {
     running: bool,
 }
 
-#[derive(Default, Debug, Copy, Clone, PartialEq, Hash)]
+#[derive(Default, Debug, Copy, Clone, PartialEq, Eq, Hash)]
 pub(crate) struct PlatformSpecificEventLoopAttributes {}
 
 macro_rules! call_event_handler {
@@ -846,18 +846,16 @@ impl MonitorHandle {
 
     pub fn video_modes(&self) -> impl Iterator<Item = monitor::VideoMode> {
         let size = self.size().into();
-        let mut v = Vec::new();
         // FIXME this is not the real refresh rate
-        // (it is guarunteed to support 32 bit color though)
-        v.push(monitor::VideoMode {
+        // (it is guaranteed to support 32 bit color though)
+        std::iter::once(monitor::VideoMode {
             video_mode: VideoMode {
                 size,
                 bit_depth: 32,
                 refresh_rate: 60,
                 monitor: self.clone(),
             },
-        });
-        v.into_iter()
+        })
     }
 }
 
--- a/src/platform_impl/ios/mod.rs
+++ b/src/platform_impl/ios/mod.rs
@@ -56,6 +56,7 @@
 //! Also note that app may not receive the LoopDestroyed event if suspended; it might be SIGKILL'ed.
 
 #![cfg(target_os = "ios")]
+#![allow(clippy::let_unit_value)]
 
 // TODO: (mtak-) UIKit requires main thread for virtually all function/method calls. This could be
 // worked around in the future by using GCD (grand central dispatch) and/or caching of values like
@@ -108,9 +109,7 @@ unsafe impl Sync for DeviceId {}
 pub enum OsError {}
 
 impl fmt::Display for OsError {
-    fn fmt(&self, _: &mut fmt::Formatter<'_>) -> fmt::Result {
-        match self {
-            _ => unreachable!(),
-        }
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "os error")
     }
 }
--- a/src/platform_impl/ios/window.rs
+++ b/src/platform_impl/ios/window.rs
@@ -37,9 +37,9 @@ pub struct Inner {
 impl Drop for Inner {
     fn drop(&mut self) {
         unsafe {
-            let () = msg_send![self.view, release];
-            let () = msg_send![self.view_controller, release];
-            let () = msg_send![self.window, release];
+            let _: () = msg_send![self.view, release];
+            let _: () = msg_send![self.view_controller, release];
+            let _: () = msg_send![self.window, release];
         }
     }
 }
@@ -52,10 +52,10 @@ impl Inner {
     pub fn set_visible(&self, visible: bool) {
         match visible {
             true => unsafe {
-                let () = msg_send![self.window, setHidden: NO];
+                let _: () = msg_send![self.window, setHidden: NO];
             },
             false => unsafe {
-                let () = msg_send![self.window, setHidden: YES];
+                let _: () = msg_send![self.window, setHidden: YES];
             },
         }
     }
@@ -78,7 +78,7 @@ impl Inner {
                 // https://developer.apple.com/documentation/uikit/uiview/1622437-setneedsdisplay?language=objc
                 app_state::queue_gl_or_metal_redraw(self.window);
             } else {
-                let () = msg_send![self.view, setNeedsDisplay];
+                let _: () = msg_send![self.view, setNeedsDisplay];
             }
         }
     }
@@ -119,8 +119,8 @@ impl Inner {
                 },
                 size: screen_frame.size,
             };
-            let bounds = self.from_screen_space(new_screen_frame);
-            let () = msg_send![self.window, setBounds: bounds];
+            let bounds = self.rect_from_screen_space(new_screen_frame);
+            let _: () = msg_send![self.window, setBounds: bounds];
         }
     }
 
@@ -218,7 +218,7 @@ impl Inner {
             let uiscreen = match monitor {
                 Some(Fullscreen::Exclusive(video_mode)) => {
                     let uiscreen = video_mode.video_mode.monitor.ui_screen() as id;
-                    let () =
+                    let _: () =
                         msg_send![uiscreen, setCurrentMode: video_mode.video_mode.screen_mode.0];
                     uiscreen
                 }
@@ -234,16 +234,16 @@ impl Inner {
             // this is pretty slow on iOS, so avoid doing it if we can
             let current: id = msg_send![self.window, screen];
             if uiscreen != current {
-                let () = msg_send![self.window, setScreen: uiscreen];
+                let _: () = msg_send![self.window, setScreen: uiscreen];
             }
 
             let bounds: CGRect = msg_send![uiscreen, bounds];
-            let () = msg_send![self.window, setFrame: bounds];
+            let _: () = msg_send![self.window, setFrame: bounds];
 
             // For external displays, we must disable overscan compensation or
             // the displayed image will have giant black bars surrounding it on
             // each side
-            let () = msg_send![
+            let _: () = msg_send![
                 uiscreen,
                 setOverscanCompensation: UIScreenOverscanCompensation::None
             ];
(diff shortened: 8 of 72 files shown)


.github/workflows/ci.yml
```diff
@@ -93,6 +96,10 @@ jobs:
         !contains(matrix.platform.target, 'wasm32'))
       run: cargo $CMD test --verbose --target ${{ matrix.platform.target }} $OPTIONS --features $FEATURES
 
+    - name: Lint with clippy
+      shell: bash
+      if: matrix.rust_version != 'nightly'
+      run: echo $FEATURES && cargo clippy --all-targets --target ${{ matrix.platform.target }} $OPTIONS --features $FEATURES
```
MarijnS95, line 102: Do we need to `echo` what features are available? That should already be part of the `job` title because this is part of the `strategy: matrix:`?

kchibisov: I just added it to test, since I want to understand how to blacklist the run on non-default-features.

src/platform_impl/ios/window.rs
```diff
@@ -452,7 +452,7 @@ impl Window {
             // Like the Windows and macOS backends, we send a `ScaleFactorChanged` and `Resized`
             // event on window creation if the DPI factor != 1.0
             let scale_factor: CGFloat = msg_send![view, contentScaleFactor];
-            let scale_factor: f64 = scale_factor.into();
+            let scale_factor: f64 = scale_factor as _;
```
filnet, line 455: The variable type was removed in a similar expression higher up.

src/platform_impl/linux/wayland/window/shim.rs
```diff
@@ -406,7 +406,8 @@ pub fn handle_window_requests(winit_state: &mut WinitState) {
     // Process the rest of the events.
     for (window_id, window_handle) in window_map.iter_mut() {
         let mut requests = window_handle.pending_window_requests.lock().unwrap();
-        for request in requests.drain(..) {
+        let requests = requests.drain(..);
+        for request in requests {
```
filnet, line 410: is that change necessary ?

kchibisov: Sort of. clippy on nightly suggested it.

src/platform_impl/linux/x11/util/geometry.rs
```diff
@@ -370,16 +370,20 @@ impl XConnection {
...
+            );
             FrameExtentsHeuristic {
                 frame_extents,
                 heuristic_path: UnsupportedNested,
             }
         } else {
             // This is the case for xmonad and dwm, AKA the only WMs tested that supplied a
             // border value. This is convenient, since we can use it to get an accurate frame.
-            let frame_extents = FrameExtents::from_border(border.into());
+            let frame_extents = FrameExtents::from_border(border as c_ulong);
```
filnet, line 386: Which clippy gets triggered here ?

PS: the best is to do one commit per clippy type...

src/platform_impl/macos/util/async.rs
```diff
@@ -170,7 +170,6 @@ pub unsafe fn set_maximized_async(
             let curr_mask = ns_window.styleMask();
             if shared_state_lock.fullscreen.is_some() {
                 // Handle it in window_did_exit_fullscreen
-                return;
```
filnet, line 173: probably some clippy but scary...

src/platform_impl/ios/window.rs
kchibisov: I don't understand what do you want from me here. The problem is that CGFLoat is either f32 or f64, but winit API is f64, so we need to cast it to f64 here. Using `.into()` won't work, since clippy finds it as redundant when `CGFloat` is `f64`, however we can cast it like `as _` and explicitly state that we assign to `f64` value.

src/platform_impl/macos/util/async.rs
kchibisov: Yeah, will refactor it.

kchibisov:
@MarijnS95 Do you happen to know why android target randomly fails from time to time? I think I've seen it a bunch of times here, rerunning do fix the issues.

src/event_loop.rs
```diff
@@ -221,6 +223,12 @@ impl EventLoop<()> {
     }
 }
 
+impl Default for EventLoop<()> {
```
madsmtm, line 226: Nit: Add changelog entry

src/lib.rs
```diff
@@ -132,6 +132,9 @@
 
 #![deny(rust_2018_idioms)]
 #![deny(rustdoc::broken_intra_doc_links)]
+#![deny(clippy::all)]
+#![cfg_attr(feature = "cargo-clippy", deny(warnings))]
```
madsmtm, line 136: How does this work?

examples/fullscreen.rs
```diff
@@ -34,8 +34,8 @@ fn main() {
     event_loop.run(move |event, _, control_flow| {
         control_flow.set_wait();
 
-        match event {
-            Event::WindowEvent { event, .. } => match event {
+        if let Event::WindowEvent { event, .. } = event {
+            match event {
```
madsmtm, line 38: See [my comment on the last Clippy PR](https://github.com/rust-windowing/winit/pull/2108#pullrequestreview-839703098), I still think there is value in making the examples use `match` instead of `if let`.

src/platform_impl/macos/view.rs
```diff
@@ -901,7 +897,7 @@ extern "C" fn insert_tab(this: &Object, _sel: Sel, _sender: id) {
         let first_responder: id = msg_send![window, firstResponder];
         let this_ptr = this as *const _ as *mut _;
         if first_responder == this_ptr {
-            let (): _ = msg_send![window, selectNextKeyView: this];
+            let _: () = msg_send![window, selectNextKeyView: this];
```
madsmtm, line 900: Haha, nice!

src/window.rs
```diff
@@ -1126,15 +1126,15 @@ impl Default for CursorIcon {
 }
 
 /// Fullscreen modes.
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq, Eq)]
```
madsmtm, line 1129: Nit: Add changelog entry

madsmtm (changes requested):
iOS and macOS changes looks fine (I never did like the `let () = msg_send![...]`, so cool that Clippy doesn't either).

src/platform_impl/macos/window_delegate.rs
```diff
@@ -269,7 +269,7 @@ extern "C" fn window_will_close(this: &Object, _: Sel, _: id) {
         autoreleasepool(|| {
             // Since El Capitan, we need to be careful that delegate methods can't
             // be called after the window closes.
-            let () = msg_send![*state.ns_window, setDelegate: nil];
+            let _: () = msg_send![*state.ns_window, setDelegate: nil];
```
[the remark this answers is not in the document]
MarijnS95: Ah I see now that you added `allow(clippy::let_unit_value)` for this, which is why my `clippy` is happy about it. Then you don't need to make this change anymore too :)

examples/request_redraw_threaded.rs
```diff
@@ -28,10 +28,9 @@ fn main() {
         control_flow.set_wait();
 
         match event {
-            Event::WindowEvent { event, .. } => match event {
-                WindowEvent::CloseRequested => control_flow.set_exit(),
-                _ => (),
-            },
+            Event::WindowEvent { event, .. } if event == WindowEvent::CloseRequested => {
```
MarijnS95, line 31: It seems common to handle these as:
```suggestion
            Event::WindowEvent { event: WindowEvent::CloseRequested, .. } => {
```

src/lib.rs
MarijnS95: `clippy` enables the `cargo-clippy` feature on all crates when running through it, automatically denying all `rustc` warnings in this instance. However, this implies warnings from `clippy::all` (that lint group should only include warn/deny lints) and the `#![deny(clippy::all)]` is thus superfluous.

src/platform_impl/android/mod.rs
```diff
@@ -845,18 +845,17 @@ impl MonitorHandle {
 
     pub fn video_modes(&self) -> impl Iterator<Item = monitor::VideoMode> {
         let size = self.size().into();
-        let mut v = Vec::new();
         // FIXME this is not the real refresh rate
         // (it is guarunteed to support 32 bit color though)
```
MarijnS95, line 849: Can you fix the typo here as a drive-by fix :)

src/platform_impl/ios/mod.rs
```diff
@@ -109,8 +110,6 @@ pub enum OsError {}
 
 impl fmt::Display for OsError {
     fn fmt(&self, _: &mut fmt::Formatter<'_>) -> fmt::Result {
-        match self {
-            _ => unreachable!(),
-        }
+        unreachable!()
```
MarijnS95, line 113: `unreachable!()` typically means something the compiler hasn't been able to deduce... This type is `pub`, are we sure no-one is ever able to call `Display` on `OsError`?

src/platform_impl/linux/wayland/window/shim.rs
kchibisov: Yes. I don't quite get as well, but I'm not that against either.

src/lib.rs
kchibisov: deny clippy all means that it'll error on clippy warnings instead of treating them as warnings.

.github/workflows/ci.yml
```diff
@@ -93,6 +95,10 @@ jobs:
         !contains(matrix.platform.target, 'wasm32'))
       run: cargo $CMD test --verbose --target ${{ matrix.platform.target }} $OPTIONS --features $FEATURES
 
+    - name: Lint with clippy
+      shell: bash
+      if: (matrix.rust_version != 'nightly') && !contains(matrix.platform.options, '--no-default-features')
+      run: cargo clippy --all-targets --target ${{ matrix.platform.target }} $OPTIONS --features $FEATURES
```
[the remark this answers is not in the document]
kchibisov: I think we don't want to run clippy for examples actually.

src/platform_impl/windows/ime.rs
```diff
@@ -80,19 +80,18 @@ impl ImeContext {
...
-        let size = ImmGetCompositionStringW(self.himc, gcs_mode, null_mut(), 0);
-        if size < 0 {
-            return None;
-        } else if size == 0 {
-            return Some(Vec::new());
-        }
+        let size = match ImmGetCompositionStringW(self.himc, gcs_mode, null_mut(), 0) {
+            0 => return Some(Vec::new()),
+            size if size < 0 => return None,
+            size => size,
```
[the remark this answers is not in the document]
kchibisov: No, it suggested to refactor into `match`, so did I.

src/platform_impl/web/web_sys/mod.rs
```diff
@@ -98,7 +98,7 @@ pub fn set_canvas_style_property(raw: &HtmlCanvasElement, property: &str, value:
     let style = raw.style();
     style
         .set_property(property, value)
-        .expect(&format!("Failed to set {}", property));
+        .unwrap_or_else(|_| panic!("Failed to set {}", property))
```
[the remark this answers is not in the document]
kchibisov: I just refactored the old code to be the same? At least I've autoapplied clippy suggestion here.

src/platform_impl/macos/window_delegate.rs
```diff
@@ -550,12 +551,12 @@ extern "C" fn window_did_fail_to_enter_fullscreen(this: &Object, _: Sel, _: id)
             shared_state.target_fullscreen = None;
         });
         if state.initial_fullscreen {
-            let _: () = unsafe {
-                msg_send![*state.ns_window,
+            unsafe {
+                let _: () = msg_send![*state.ns_window,
```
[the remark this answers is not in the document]
kchibisov: I've manually updated all the `let ()`, since I hate them. And we have half of the code with `let _: ()` and `let ()`, the first one is more clear, since it clearly states that we bind the type.

src/platform_impl/linux/x11/util/randr.rs
```diff
@@ -164,7 +164,8 @@ impl XConnection {
         (self.xrandr.XRRFreeOutputInfo)(output_info);
         Some((name, scale_factor, modes))
     }
-    pub fn set_crtc_config(&self, crtc_id: RRCrtc, mode_id: RRMode) -> Result<(), ()> {
+
+    pub fn set_crtc_config(&self, crtc_id: RRCrtc, mode_id: RRMode) -> Option<()> {
```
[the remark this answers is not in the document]
kchibisov: Yeah, should probably add one, the point is that this function call literally does nothing in return.

src/platform_impl/ios/mod.rs
kchibisov: Yeah, I'll add some stub.

src/platform_impl/windows/ime.rs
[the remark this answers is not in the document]
kchibisov: Though, we need a value, so we either bind it and match with cmp, or do what I'm doing.

src/platform_impl/web/web_sys/mod.rs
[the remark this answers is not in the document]
kchibisov: I've include it, but there's no display for it, and I don't know anything about web to impl it myself.

.github/workflows/ci.yml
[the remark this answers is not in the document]
MarijnS95: I think we should be consistent and enforce it everywhere, so that the examples are also clean-enough.

src/window.rs
kchibisov: I don't think internal impl requires it? Also I'm not sure that we should add changelog notes for `Eq` impls, since they are not breaking the code and transparent.

src/platform_impl/linux/x11/util/randr.rs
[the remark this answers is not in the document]
MarijnS95: It doesn't mean the same though, the point is that the caller is supposed to check if this function succeeded or not. Return the `status: i32` in the `Err()` case for my part.

src/platform_impl/web/web_sys/mod.rs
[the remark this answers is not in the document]
MarijnS95: > I've include it, but there's no display for it, and I don't know anything about web to impl it myself.

This comment didn't show up before a page reload :grimacing: - expect uses `Debug` not `Display` so you should be able to trivially `: {:?}` add that to the `panic!`.

.github/workflows/ci.yml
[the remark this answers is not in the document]
MarijnS95: In addition there's `--all-targets` here and you fixed+allowed clippy lints in the `examples/`, so this is only sensible?

MarijnS95 (approved):
It looks like all my comments were addressed minus some minor nits which we can continue to discuss, nice job!

.github/workflows/ci.yml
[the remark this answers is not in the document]
kchibisov: I'd add `-D warnings` on CI then, I guess.

Having deny in lib.rs is still easier to understand that you must fix those issues.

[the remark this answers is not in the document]
MarijnS95: Yeah, leave the deny in `lib.rs`, that should hopefully decrease the amount of noise in PRs that would otherwise require lots of cleanup pushes to "make the CI happy".

src/lib.rs
MarijnS95: But that's also what `#![cfg_attr(feature = "cargo-clippy", deny(warnings))]` does :)

kchibisov: I think those about rustc warnings, and not clippy onces? I think it was like that in the past, at least.

src/window.rs
madsmtm: Internal impls don't matter.

It's not a breaking change, but in general I think all API changes should be documented in the changelog, including small ones like this. But I don't have a strong opinion on this, you're free to disagree.

src/platform_impl/ios/mod.rs
```diff
@@ -108,9 +109,7 @@ unsafe impl Sync for DeviceId {}
 pub enum OsError {}
 
 impl fmt::Display for OsError {
-    fn fmt(&self, _: &mut fmt::Formatter<'_>) -> fmt::Result {
-        match self {
-            _ => unreachable!(),
-        }
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "os error.")
```
maroider, line 113: nit: this is hardly a sentence, so the period seems misplaced
```suggestion
        write!(f, "os error")
```

src/platform_impl/web/web_sys/scaling.rs
```diff
@@ -65,13 +65,13 @@ impl ScaleChangeDetectorInternal {
         );
         let mql = MediaQueryListHandle::new(&media_query, closure);
         if let Some(mql) = &mql {
-            assert_eq!(mql.mql().matches(), true);
+            assert!(!mql.mql().matches());
```
maroider, line 68: It *looks* like the logic got accidentally negated here.

src/platform_impl/web/web_sys/scaling.rs
```diff
@@ -65,13 +65,13 @@ impl ScaleChangeDetectorInternal {
...
         }
         mql
     }
 
     fn handler(&mut self, event: MediaQueryListEvent) {
-        assert_eq!(event.matches(), false);
+        assert!(event.matches());
```
maroider, line 74: ditto

.github/workflows/ci.yml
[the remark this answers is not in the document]
maroider: I'm fairly certain that the general advice here is to _not_ deny warnings in `*.rs` files, and instead only do it in CI, as it might break compilation with future versions of `rustc` and/or `clippy`.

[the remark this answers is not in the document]
kchibisov: That's the point, we don't deny warnings for rustc or any related command, just when running clippy. So there's no way the build will break, only clippy could, but it would only mean that CI is broken.

madsmtm (approved):

.github/workflows/ci.yml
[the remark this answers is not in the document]
maroider: Ah, I see. That's clever.

src/window.rs
maroider: A changelog entry would indeed be nice.

maroider (approved):
I don't have any further blocking concerns, although it would be nice to have changelog entries for all the added trait impls on public types.

msiglreith (approved):

kchibisov:
Android failures were unrelated and quite random recently.

src/platform_impl/macos/mod.rs
```diff
@@ -1,4 +1,5 @@
 #![cfg(target_os = "macos")]
+#![allow(clippy::let_unit_value)]
```
madsmtm, line 2: That we have to specify this is actually a regression, I've filed an issue about it here: https://github.com/rust-lang/rust-clippy/issues/8998.

PR #2318, opened 2022-06-09
On Android, use `HasRawWindowHandle` directly from the `ndk` crate

MarijnS95:
The `ndk` crate now implements [`HasRawWindowHandle` directly on `NativeWindow`], relieving the burden on `winit` to carry the same implementation details.

[`HasRawWindowHandle` directly on `NativeWindow`]: https://github.com/rust-windowing/android-ndk-rs/pull/274

---

This PR is also being opened to discuss how we should proceed with up and coming `ndk` (+ `ndk_glue`) changes that are not yet published. I'd like to coordinate the next (which is going to be breaking) release of those crates with `winit` so that we're not unnecessarily creating a desync in the ecosystem (https://github.com/rust-windowing/winit/issues/1995). That'd require Android PRs to remain open until `winit` eventually starts the `0.27` cycle and I decide when and where to make the cut and publish the `ndk`. Alternatives are:

- Make preview/alpha/beta releases for `ndk`+`ndk_glue`;
- Use git references in the `patch` section: this doesn't carry over to user-crates having a `git` reference on `winit`, allowing only local development;
- Temporarily replace these crates with `git` dependencies: this prevents `winit` from making release before the Android crates are released;
- (Repeating what was mentioned above: leaving dependent Android PRs open and merging them in at the last moment, when an `ndk` release is made).

I'm sure similar issues might have come up with other crates, bar the "coordinate a release" bit.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,9 +38,9 @@ image = { version = "0.24.0", default-features = false, features = ["png"] }
 simple_logger = "2.1.0"
 
 [target.'cfg(target_os = "android")'.dependencies]
-ndk = "0.6"
-ndk-sys = "0.3"
-ndk-glue = "0.6"
+# Coordinate the next winit release with android-ndk-rs: https://github.com/rust-windowing/winit/issues/1995
+ndk = { git = "https://github.com/rust-windowing/android-ndk-rs", rev = "7e33384" }
+ndk-glue = { git = "https://github.com/rust-windowing/android-ndk-rs", rev = "7e33384" }
 
 [target.'cfg(any(target_os = "ios", target_os = "macos"))'.dependencies]
 objc = "0.2.7"
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -13,7 +13,7 @@ use ndk::{
 };
 use ndk_glue::{Event, Rect};
 use once_cell::sync::Lazy;
-use raw_window_handle::{AndroidNdkHandle, RawWindowHandle};
+use raw_window_handle::{HasRawWindowHandle, RawWindowHandle};
 
 use crate::{
     dpi::{PhysicalPosition, PhysicalSize, Position, Size},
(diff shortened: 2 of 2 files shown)


madsmtm:
I think the third alternative:
> Temporarily replace these crates with git dependencies: this prevents winit from making release before the Android crates are released

Is my preferred option, since it allows things to be merged (so we don't end up with an even larger PR-backlog than we already have), and we won't accidentally forget to update to the correct version before publishing.

If you want to spend the time on it, preview/alpha/beta releases are also fine.

MarijnS95:
> Is my preferred option, since it allows things to be merged (so we don't end up with an even larger PR-backlog than we already have), and we won't accidentally forget to update to the correct version before publishing.

Lovely, that's my preferred approach as well; `ndk` lives under the same orga anyway so if everything falls apart `winit` maintainers can release it too, but I'll make sure to be around for that :)

If all `winit` maintainers agree I'll change this PR to that, so that we can get it merged.

msiglreith (approved):

MarijnS95:
@msiglreith The CI seems to fail on the dreaded `= note: ld: error: unable to find library -lgcc` again (we have multiple tested workarounds in `cargo-apk`...), and it has been haunting random other PRs recently too... Any idea what's up and why it only appears sometimes?

msiglreith:
@MarijnS95 Might be related to the ndk update of the github runners: https://github.com/actions/virtual-environments/issues/5595
Not sure tho why the v23 workaround for the missing `libgcc` doesn't seem to trigger here /:

MarijnS95:
Oh that's horrible, we did our best to improve the workaround but it's failing for `cargo test --no-build`... Because we only build `--lib`s in principle.

I don't even recall `cargo apk` having support for `test`, it seems the CI is using `cargo apk -- <build/test>` which actually forwards the arguments back to `cargo`, how does that even work :thinking: 

All the other builds succeed, I may have disabled test-building for the same reason in `glutin` :sweat_smile: https://github.com/rust-windowing/glutin/pull/1415/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR103

MarijnS95:
I just tested `cargo apk -- build` locally, and it doesn't invoke `cargo-apk` at all?

MarijnS95:
Turns out this is a special feature of `cargo-apk`, but its implementation got desynced with regular `cargo build`. It _is_ going through `cargo-apk` but doesn't include the workaround (heh, I only implemented that in `fn build()`) and it's not passing a default `--target aarch64-linux-android` either (nor the one for the detected ABI of a connected `adb` device [^1]) despite preparing the NDK for specifically that target triple... Fixing that now.


[^1]: switching behaviour for build/run like that seems fragile at best, perhaps we should mimic `cargo-ndk` and print the current triple that we're building for?

PR #2324, opened 2022-06-10
Build docs on `docs.rs` for iOS and Android as well

madsmtm:
Fixes https://github.com/rust-windowing/winit/issues/1028. `docs.rs` has come a long way since that issue, so the problems they encountered are (probably) no more - at least `ndk` (Android) and `objc2` (iOS + macOS) all successfully build.

Tested with:
```
cargo doc --target aarch64-linux-android
cargo doc --target x86_64-apple-ios
```

We'll have to watch https://docs.rs/crate/winit/latest/builds after the next release, to ensure that everything actually works as expected!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,6 +24,7 @@ jobs:
       matrix:
         rust_version: [stable, nightly]
         platform:
+          # Note: Make sure that we test all the `docs.rs` targets defined in Cargo.toml!
           - { target: x86_64-pc-windows-msvc,   os: windows-latest,  }
           - { target: i686-pc-windows-msvc,     os: windows-latest,  }
           - { target: x86_64-pc-windows-gnu,    os: windows-latest, host: -x86_64-pc-windows-gnu }
@@ -77,7 +78,6 @@ jobs:
 
     - name: Check documentation
       shell: bash
-      if: matrix.platform.target != 'wasm32-unknown-unknown'
       run: cargo $CMD doc --no-deps --target ${{ matrix.platform.target }} $OPTIONS --features $FEATURES
 
     - name: Build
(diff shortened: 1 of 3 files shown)


maroider (approved):

MarijnS95:
@madsmtm it seems that we've run out of luck. The Android docs got built for `0.27.0`: https://docs.rs/winit/0.27.0/aarch64-linux-android/winit/index.html

But are no longer there on `0.28`/`0.29` :disappointed: 

Notice that there are only ever 7 platforms in the dropdown, [on `0.27.0` `i686-unknown-linux-gnu` is missing](https://docs.rs/winit/0.27.0/i686-unknown-linux-gnu/winit/index.html) even though it is definitely in the list in this PR.

https://docs.rs/about/metadata doesn't specify that there's a limit in the number of `targets` one can have :confused:

madsmtm:
They do specify the default limit [here](https://docs.rs/about/builds#hitting-resource-limits), and the project specific ones [here](https://docs.rs/crate/winit/latest/builds), but we shouldn't be hitting that (it's 10, we're only using 8)!

I think it's due to the usual "must specify native or game activity" issue; remember, docs.rs is basically just running the following command, it doesn't know anything about enabling other features.
```console
RUSTDOCFLAGS="--cfg docsrs" cargo +nightly doc --target=aarch64-linux-android --features=rwh_04,rwh_05,rwh_06,serde
```

I'll whip up a PR to fix it by just enabling the native activity when building docs, hold on.

(Ugh, this would be much more of a non-issue if https://github.com/rust-windowing/winit/pull/3076 was implemented).

MarijnS95:
I didn't think about that, but also didn't expect it given that there's no error like that in the docs.rs build logs (or anything hinting that it is compiling for 8 different targets)?

PR #1074, opened 2019-07-28
Always use stable rustfmt for CI.

ZeGentzy:
Signed-off-by: Hal Gentz <zegentzy@protonmail.com>

- [ ] Tested on all platforms changed
- [ ] Compilation warnings were addressed
- [ ] `cargo fmt` has been run on this branch
- [ ] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users
- [ ] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior
- [ ] Created or updated an example program if it would help users understand this functionality
- [ ] Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented

PR #2325, opened 2022-06-10
Refine rustfmt

kchibisov:
That's just a prototype how it should look to apply nightly rustfmt with a bit more involved rules.

There's no intent to do that before the new keyboard API rework.

### DIFF
--- a/examples/control_flow.rs
+++ b/examples/control_flow.rs
@@ -85,14 +85,9 @@ impl ApplicationHandler for ControlFlowDemo {
         match event {
             WindowEvent::CloseRequested => {
                 self.close_requested = true;
-            }
+            },
             WindowEvent::KeyboardInput {
-                event:
-                    KeyEvent {
-                        logical_key: key,
-                        state: ElementState::Pressed,
-                        ..
-                    },
+                event: KeyEvent { logical_key: key, state: ElementState::Pressed, .. },
                 ..
             } => match key.as_ref() {
                 // WARNING: Consider using `key_without_modifiers()` if available on your platform.
@@ -100,29 +95,29 @@ impl ApplicationHandler for ControlFlowDemo {
                 Key::Character("1") => {
                     self.mode = Mode::Wait;
                     warn!("mode: {:?}", self.mode);
-                }
+                },
                 Key::Character("2") => {
                     self.mode = Mode::WaitUntil;
                     warn!("mode: {:?}", self.mode);
-                }
+                },
                 Key::Character("3") => {
                     self.mode = Mode::Poll;
                     warn!("mode: {:?}", self.mode);
-                }
+                },
                 Key::Character("r") => {
                     self.request_redraw = !self.request_redraw;
                     warn!("request_redraw: {}", self.request_redraw);
-                }
+                },
                 Key::Named(NamedKey::Escape) => {
                     self.close_requested = true;
-                }
+                },
                 _ => (),
             },
             WindowEvent::RedrawRequested => {
                 let window = self.window.as_ref().unwrap();
                 window.pre_present_notify();
                 fill::fill_window(window);
-            }
+            },
             _ => (),
         }
     }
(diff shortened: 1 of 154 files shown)


examples/control_flow.rs
```diff
@@ -3,11 +3,9 @@
...
 
 use simple_logger::SimpleLogger;
-use winit::{
-    event::{Event, KeyboardInput, WindowEvent},
-    event_loop::EventLoop,
-    window::WindowBuilder,
-};
+use winit::event::{Event, KeyboardInput, WindowEvent};
+use winit::event_loop::EventLoop;
+use winit::window::WindowBuilder;
```
MarijnS95, line 8: I personally quite like this grouping, are we missing an option?

madsmtm: I dislike it, I find that it makes merge conflicts harder

maroider: I've personally switched to `imports_granularity = "Item"` to make my life easier on that front.

kchibisov: I personally hate the grouping current winit is using, and personally I don't have it in Wayland backend, since rustfmt doesn't enforce it. It just nearly impossible to read it and rebase. Or maybe I've just used to `Module` option.

aloucks:
For what it's worth, `+nightly` was originally used as the formatting check, but caused problems for contributors who weren't using nightly Rust. See also: https://github.com/rust-windowing/winit/pull/1074

madsmtm:
We agreed to do this in today's meeting, though only after a few of the larger in-flight PRs have been merged.

I raised the concern that users would have to run `cargo +nightly fmt`, and that their editor would be set to run the stable `rustfmt` by default. My concerns would be assuaged if we:
1. Added a note about this to `CONTRIBUTING.md`, perhaps noting that you can configure `rust-analyzer.rustfmt.extraArgs` with `["+nightly"]` to make this work by default.
2. Changed the CI step to be called "Check nightly formatting".

madsmtm (approved):
I assume the only thing you've changed is `rustfmt.toml`. Had a look at it, I agree with all the options, though I think you might get trouble with `reorder_impl_items` until https://github.com/madsmtm/objc2/issues/479 is resolved, so might want to wait with that.

kchibisov:
The only thing I could change is comment width to `80`, so it's a bit more compact, but that's about it.

rustfmt.toml
```diff
@@ -1,3 +1,16 @@
...
+match_block_trailing_comma = true
+condense_wildcard_suffixes = true
+use_field_init_shorthand = true
+normalize_doc_attributes = true
+overflow_delimited_expr = true
+imports_granularity = "Module"
+use_small_heuristics = "Max"
+normalize_comments = true
+reorder_impl_items = true
+use_try_shorthand = true
```
daxpedda, line 11: This is deprecated, I don't think we need that in Rustfmt.

rustfmt.toml
```diff
@@ -1,3 +1,16 @@
...
 force_explicit_abi=true
```
daxpedda, line 12: This can be removed, its default is already `true`.

daxpedda (changes requested):
[`unstable_features`](https://rust-lang.github.io/rustfmt/?version=master#unstable_features) is missing.

I would also like to add the following:
- [`error_on_unformatted`](https://rust-lang.github.io/rustfmt/?version=master#error_on_unformatted)
- [`format_macro_bodies`](https://rust-lang.github.io/rustfmt/?version=master#format_macro_bodies)
- [`format_macro_matchers`](https://rust-lang.github.io/rustfmt/?version=master#format_macro_matchers)
- [`hex_literal_case`](https://rust-lang.github.io/rustfmt/?version=master#hex_literal_case)

I'm also against the following rules because they move away from the default Rust style:
- [`match_block_trailing_comma`](https://rust-lang.github.io/rustfmt/?version=master#match_block_trailing_comma)
- [`overflow_delimited_expr`](https://rust-lang.github.io/rustfmt/?version=master#overflow_delimited_expr)
- [`use_small_heuristics`](https://rust-lang.github.io/rustfmt/?version=master#use_small_heuristics)

kchibisov:
And `use_small` is the main reason for me to change it, because the code is way easier to read with it, since it's more compact.

daxpedda (approved):

Issue #1677, opened 2020-08-24
Add `set_cursor_lock` and distinguish between grabbing (confining) cursor and locking cursor


alvinhochun:
Currently there is a `set_cursor_grab` API which is supposed to confine the cursor movement to within the window client area, but sometimes the user may want to lock the cursor to the current position instead (e.g. for camera control). Also, the web platform does not support confining the cursor but does support locking the cursor with the Pointer Lock API. Therefore, I would suggest adding a new API `set_cursor_lock` to lock the cursor to the current position, and fixing the `set_cursor_grab` API to actually confine the cursor on all supported platforms.

A better alternative would be to change `set_cursor_grab` to take an `enum CursorGrabState { Off, Confine, Lock }` so that only one of the two cursor grabbing modes can be active at once.

|                 | set_cursor_grab / Confine                   | set_cursor_lock / Lock                                   |
|-----------------|--------------------------------------------------|----------------------------------------------------------|
| Windows         | Supported, already implemented                   | Supported, unimplemented (could reuse cursor grab impl.) |
| Linux (X11)     | Supported, already implemented                   | Unknown (could just move the mouse after mouse moved?)   |
| Linux (Wayland) | Supported, unimplemented (#1678) | Supported, already implemented for `set_cursor_grab`     |
| macOS           | Supported, unimplemented (#1093)                 | Supported, already implemented for `set_cursor_grab`     |
| iOS             | Not supported                                    | Not supported                                            |
| Android         | Not supported                                    | Not supported                                            |
| Web             | Not supported                                    | Supported, unimplemented (#1674)                         |

kchibisov:
After thinking about this issue I wonder how those functions could relate to each other?

By that I mean if I do `set_cursor_grab(true)`, what the following `set_cursor_lock(true)` should do? Vise versa?

So we can make them

- Mutually exclusive
- Lock > grab, meaning if the cursor is locked, `grab` won't do anything until  `unlock`. And on `unlock` the pending `grab` will apply.

I personally think that the second variant is something more logical, however the first option is also worth considering.

The point here that such behavior must be explicitly defined, so we should agree on that before hand.

alvinhochun:
If we want them to be mutually exclusive, then perhaps the API should be changed to something like this?

```rust
fn set_cursor_grab(&self, grab_mode: CursorGrabMode) -> Result<CursorGrabState, ExternalError>;
fn get_cursor_grab(&self) -> CursorGrabState;

[...]

PR #2329, opened 2022-06-11
Refine `Window::set_cursor_grab` API

kchibisov:
This commit renames `Window::set_cursor_grab` to
`Window::set_cursor_grab_mode`. The new API now accepts enumeration
to control the way cursor grab is performed. The value could be: `lock`,
`confine`, or `none`.

This commit also implements `Window::set_cursor_position` for Wayland,
since it's tied to locked cursor.

Implements API from #1677.

### DIFF
--- a/examples/cursor_grab.rs
+++ b/examples/cursor_grab.rs
@@ -4,7 +4,7 @@ use simple_logger::SimpleLogger;
 use winit::{
     event::{DeviceEvent, ElementState, Event, KeyboardInput, ModifiersState, WindowEvent},
     event_loop::EventLoop,
-    window::WindowBuilder,
+    window::{CursorGrabMode, WindowBuilder},
 };
 
 fn main() {
@@ -36,7 +36,11 @@ fn main() {
                     use winit::event::VirtualKeyCode::*;
                     match key {
                         Escape => control_flow.set_exit(),
-                        G => window.set_cursor_grab(!modifiers.shift()).unwrap(),
+                        G => window
+                            .set_cursor_grab_mode(CursorGrabMode::Confined)
+                            .unwrap(),
+                        L => window.set_cursor_grab_mode(CursorGrabMode::Locked).unwrap(),
+                        A => window.set_cursor_grab_mode(CursorGrabMode::None).unwrap(),
                         H => window.set_cursor_visible(modifiers.shift()),
                         _ => (),
                     }
--- a/examples/cursor_grab.rs
+++ b/examples/cursor_grab.rs
@@ -34,15 +34,23 @@ fn main() {
                     ..
                 } => {
                     use winit::event::VirtualKeyCode::*;
-                    match key {
-                        Escape => control_flow.set_exit(),
-                        G => window
-                            .set_cursor_grab_mode(CursorGrabMode::Confined)
-                            .unwrap(),
-                        L => window.set_cursor_grab_mode(CursorGrabMode::Locked).unwrap(),
-                        A => window.set_cursor_grab_mode(CursorGrabMode::None).unwrap(),
-                        H => window.set_cursor_visible(modifiers.shift()),
-                        _ => (),
+                    let result = match key {
+                        Escape => {
+                            control_flow.set_exit();
+                            Ok(())
+                        }
+                        G => window.set_cursor_grab(CursorGrabMode::Confined),
+                        L => window.set_cursor_grab(CursorGrabMode::Locked),
+                        A => window.set_cursor_grab(CursorGrabMode::None),
+                        H => {
+                            window.set_cursor_visible(modifiers.shift());
+                            Ok(())
+                        }
+                        _ => Ok(()),
+                    };
+
+                    if let Err(err) = result {
+                        println!("error: {}", err);
                     }
                 }
                 WindowEvent::ModifiersChanged(m) => modifiers = m,
--- a/examples/multithreaded.rs
+++ b/examples/multithreaded.rs
@@ -89,13 +89,19 @@ fn main() {
                                 (false, _) => None,
                             }),
                             L if state => {
-                                window.set_cursor_grab_mode(CursorGrabMode::Locked).unwrap()
+                                if let Err(err) = window.set_cursor_grab(CursorGrabMode::Locked) {
+                                    println!("error: {}", err);
+                                }
                             }
-                            G if state => window
-                                .set_cursor_grab_mode(CursorGrabMode::Confined)
-                                .unwrap(),
+                            G if state => {
+                                if let Err(err) = window.set_cursor_grab(CursorGrabMode::Confined) {
+                                    println!("error: {}", err);
+                                }
+                            }
                             G | L if !state => {
-                                window.set_cursor_grab_mode(CursorGrabMode::None).unwrap()
+                                if let Err(err) = window.set_cursor_grab(CursorGrabMode::None) {
+                                    println!("error: {}", err);
+                                }
                             }
                             H => window.set_cursor_visible(!state),
                             I => {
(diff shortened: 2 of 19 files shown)


src/window.rs
```diff
@@ -1035,6 +1027,35 @@ unsafe impl raw_window_handle::HasRawWindowHandle for Window {
...
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **macOS:** Not implemented.
+    /// - ** iOS / Android / Web:** Always returns an [`ExternalError::NotSupported`].
+    Confined,
+
+    /// The cursor is locked inside the window area to the certain position.
+    ///
+    /// There's no guarantee that the cursor will be hidden, you should h
```
madsmtm, line 1050: nit: incomplete

src/window.rs
```diff
@@ -917,18 +917,10 @@ impl Window {
...
-    ///
-    /// ## Platform-specific
-    ///
-    /// - **macOS:** This locks the cursor in a fixed location, which looks visually awkward.
-    /// - **iOS / Android:** Always returns an [`ExternalError::NotSupported`].
+    /// Set grabbing [mode]([`CursorGrabMode`]) on the cursor preventing it from leaving the window.
     #[inline]
-    pub fn set_cursor_grab(&self, grab: bool) -> Result<(), ExternalError> {
-        self.window.set_cursor_grab(grab)
+    pub fn set_cursor_grab_mode(&self, mode: CursorGrabMode) -> Result<(), ExternalError> {
```
madsmtm, line 922: I think the old name `set_cursor_grab` is fine - will also make it easier for users to upgrade, since they will instantly see that they should use an enum instead.

Alternatively, we could call it `set_cursor_constraints` or something?

src/window.rs
```diff
@@ -917,18 +917,10 @@ impl Window {
...
-    /// Grabs the cursor, preventing it from leaving the window.
-    ///
-    /// There's no guarantee that the cursor will be hidden. You should
-    /// hide it by yourself if you want so.
-    ///
-    /// ## Platform-specific
-    ///
-    /// - **macOS:** This locks the cursor in a fixed location, which looks visually awkward.
-    /// - **iOS / Android:** Always returns an [`ExternalError::NotSupported`].
+    /// Set grabbing [mode]([`CursorGrabMode`]) on the cursor preventing it from leaving the window.
```
madsmtm, line 920: Add example of using `.or` to try `Confined`, and fall back to `Locked`. E.g. something like:
```suggestion
    /// Set grabbing [mode]([`CursorGrabMode`]) on the cursor preventing it from leaving the window.
    ///
    /// # Example
    ///
    /// First try confining the cursor, and if that fails, try locking it instead.
    ///
    /// ```no-run
    /// # use winit::event_loop::EventLoop;
    /// # use winit::window::{CursorGrabMode, Window};
    /// # let mut event_loop = EventLoop::new();
    /// # let window = Window::new(&event_loop).unwrap();
    /// window.set_cursor_grab(CursorGrabMode::Confined)
    ///             .or_else(|_e| window.set_cursor_grab(CursorGrabMode::Locked))
    ///             .unwrap();
    /// ```
```

madsmtm (changes requested):
Nice, we should definitely do this!

src/window.rs
maroider: I think keeping the old name is fine

src/window.rs
```diff
@@ -1035,6 +1027,35 @@ unsafe impl raw_window_handle::HasRawWindowHandle for Window {
     }
 }
 
+/// The behavior of cursor grabbing.
+///
+/// To grab cursor [`Window::set_cursor_grab_mode`] should be used.
```
maroider, line 1032: nit: phrasing
```suggestion
/// Use this enum with [`Window::set_cursor_grab_mode`] to grab the cursor.
```

src/window.rs
```diff
@@ -1035,6 +1027,35 @@ unsafe impl raw_window_handle::HasRawWindowHandle for Window {
...
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum CursorGrabMode {
+    /// No grabbing of the cursor is performed.
+    None,
+
+    /// There's no guarantee that the cursor will be hidden. You should hide it by yourself if you
+    /// want so.
```
maroider, line 1040: nit: missing words
```suggestion
    /// There's no guarantee that the cursor will be hidden. You should hide it by yourself if you
    /// want to do so.
```
dropping everything after "want" also works.

examples/cursor_grab.rs
```diff
@@ -36,7 +36,11 @@ fn main() {
                     use winit::event::VirtualKeyCode::*;
                     match key {
                         Escape => control_flow.set_exit(),
-                        G => window.set_cursor_grab(!modifiers.shift()).unwrap(),
+                        G => window
+                            .set_cursor_grab_mode(CursorGrabMode::Confined)
+                            .unwrap(),
+                        L => window.set_cursor_grab_mode(CursorGrabMode::Locked).unwrap(),
+                        A => window.set_cursor_grab_mode(CursorGrabMode::None).unwrap(),
```
maroider, line 43: Is panicking the correct thing to do here? I'd prefer printing errors instead.

examples/multithreaded.rs
```diff
@@ -88,7 +88,15 @@ fn main() {
...
-                            G => window.set_cursor_grab(state).unwrap(),
+                            L if state => {
+                                window.set_cursor_grab_mode(CursorGrabMode::Locked).unwrap()
+                            }
+                            G if state => window
+                                .set_cursor_grab_mode(CursorGrabMode::Confined)
+                                .unwrap(),
+                            G | L if !state => {
+                                window.set_cursor_grab_mode(CursorGrabMode::None).unwrap()
+                            }
```
maroider, line 99: ditto

src/window.rs
```diff
@@ -1035,6 +1027,35 @@ unsafe impl raw_window_handle::HasRawWindowHandle for Window {
...
+    /// - ** iOS / Android / Web:** Always returns an [`ExternalError::NotSupported`].
+    Confined,
+
+    /// The cursor is locked inside the window area to the certain position.
+    ///
+    /// There's no guarantee that the cursor will be hidden, you should h
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **X11 / Windows:** Not implemented.
```
maroider, line 1054: The X11 and Windows backends both always return `ExternalError::NotSupported` in this case. This line makes it seem like they *might* not do so.

src/window.rs
```diff
@@ -1035,6 +1027,35 @@ unsafe impl raw_window_handle::HasRawWindowHandle for Window {
...
+pub enum CursorGrabMode {
+    /// No grabbing of the cursor is performed.
+    None,
+
+    /// There's no guarantee that the cursor will be hidden. You should hide it by yourself if you
+    /// want so.
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **macOS:** Not implemented.
```
maroider, line 1044: The macOS backend always returns `ExternalError::NotSupported` in this case. This line makes it seem like it *might* not do so.

maroider (commented):
I like what you're going for here

src/window.rs
kchibisov: The problem is that on those platforms it can be implemented. But on `ios`, `android`, etc won't ever have something like that.

Though I think we can also say that they always report.

maroider: That's probably an important distincion, yeah.

madsmtm (approved):
Perfect, thanks!

msiglreith (approved):

src/window.rs
maroider: If it's the case that we *may* support this mode on X11 and Windows at some point, then I think it's a good idea to note that they return `ExternalError::NotSupported` for now, but may not do so in the future.

kchibisov: Something like `Not implemented. Always return [ExternalError::NotSupported] for now.` ?

maroider: Slap an "s" at the end of "return", and then that should suffice.

kchibisov: Done.

maroider (approved):

PR #2330, opened 2022-06-11
Add `WindowBuilder::transparent`

kchibisov:
This is required to help hardware accelerated libraries like glutin
that accept WindowBuilder instead of RawWindowHandle, since the api
to access builder properties directly was removed.

Follow up to 44288f6.


kchibisov:
In general certain libraries do accept WindowBuilder instead of RawWindowHandle, so I'm not sure that making things private was right in the first place.

As for now, we should either revert the initial commit that broke glutin or add this getter to accommodate for glutin needs.

Eventually glutin will be refactored to not use winit directly, but rather RawWindowHandle, but that's unlikely to be very soon.

madsmtm (approved):
Of course, sorry it causes breakage for `glutin`!

maroider (commented):
I have no strong feelings one way or the other

PR #2322, opened 2022-06-10
Release 0.27.0 version

kchibisov:
I think it's time to cut a new release.

If there're suggestion on what should made into it let me know.

---

Blocked on:

- [x] #1995
- [x] #2376
- [x] #2378

Post release checklist:

- [x] [`docs.rs` successfully builds](https://docs.rs/crate/winit/latest/builds) after #2324

PR #2331, opened 2022-06-12
Consistently deliver a `Resumed` event on all platforms

rib:
To be more consistent with mobile platforms this updates the Windows,
macOS, Wayland, X11 and Web backends to all emit a Resumed event
immediately after the initial `NewEvents(StartCause::Init)` event.

The documentation for the Resumed event has also been updated to
recommend that applications should wait until they receive the Resumed
event before creating any graphics context and their first window to
help ensure portability.

This consistency makes it possible to write applications that lazily
initialize their graphics state when the application resumes without
any platform-specific knowledge. Previously, applications that wanted
to run on Android needed to have `target_os="android"` specific code
to initialize within the `Resumed` event because that same logic
would never run on non-mobile platforms.

Relates to:
https://github.com/rust-windowing/winit/discussions/2111 - Is it possible to automatically port over a GUI application from Linux/Windows to Android that was written in Rust?
https://github.com/rust-windowing/winit/issues/2185 - Documentation on `Event::{Suspended, Resumed}` is lacking
https://github.com/rust-windowing/winit/pull/2144 - Add Foreground and Background events for iOS
https://github.com/rust-windowing/winit/issues/2293 - Evolution of the Android backend

  - [x] Tested on Windows
  - [x] Tested on X11
  - [x] Tested on Wayland
  - [ ] Tested on macOS
  - [ ] Tested on iOS
  - [x] Tested on Web
  - For reference, this is an example of a winit application that has special case `target_os="android"` handling which could be removed: https://github.com/rib/bluey/blob/7428d9512ecfff22f2e437820f02585bc6ccf1ba/bluey-ui/src/lib.rs#L115

Fixes: #2185

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - **Breaking**, update `raw-window-handle` to `v0.5` and implement `HasRawDisplayHandle` for `Window` and `EventLoopWindowTarget`.
 - On X11, add function `register_xlib_error_hook` into `winit::platform::unix` to subscribe for errors comming from Xlib.
 - On Android, upgrade `ndk` and `ndk-glue` dependencies to the recently released `0.7.0`.
+- All platforms can now be relied on to emit a `Resumed` event. Applications are recommended to lazily initialize graphics state and windows on first resume for portability.
 
 # 0.26.1 (2022-01-05)
 
--- a/src/event.rs
+++ b/src/event.rs
@@ -74,9 +74,107 @@ pub enum Event<'a, T: 'static> {
     UserEvent(T),
 
     /// Emitted when the application has been suspended.
+    ///
+    /// # Portability
+    ///
+    /// Not all platforms support the notion of suspending applications, and there may be no
+    /// technical way to guarantee being able to emit a `Suspended` event if the OS has
+    /// no formal application lifecycle (currently only Android and iOS do). For this reason,
+    /// Winit does not currently try to emit pseudo `Suspended` events before the application
+    /// quits on platforms without an application lifecycle.
+    ///
+    /// Considering that the implementation of `Suspended` and [`Resumed`] events may be internally
+    /// driven by multiple platform-specific events, and that there may be subtle differences across
+    /// platforms with how these internal events are delivered, it's recommended that applications
+    /// be able to gracefully handle redundant (i.e. back-to-back) `Suspended` or [`Resumed`] events.
+    ///
+    /// Also see [`Resumed`] notes.
+    ///
+    /// ## Android
+    ///
+    /// On Android, the `Suspended` event is only sent when the application's associated
+    /// [`SurfaceView`] is destroyed. This is expected to closely correlate with the [`onPause`]
+    /// lifecycle event but there may technically be a discrepancy.
+    ///
+    /// [`onPause`]: https://developer.android.com/reference/android/app/Activity#onPause()
+    ///
+    /// Applications that need to run on Android should assume their [`SurfaceView`] has been
+    /// destroyed, which indirectly invalidates any existing render surfaces that may have been
+    /// created outside of Winit (such as an `EGLSurface`, [`VkSurfaceKHR`] or [`wgpu::Surface`]).
+    ///
+    /// After being `Suspended` on Android applications must drop all render surfaces before
+    /// the event callback completes, which may be re-created when the application is next [`Resumed`].
+    ///
+    /// [`SurfaceView`]: https://developer.android.com/reference/android/view/SurfaceView
+    /// [Activity lifecycle]: https://developer.android.com/guide/components/activities/activity-lifecycle
+    /// [`VkSurfaceKHR`]: https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkSurfaceKHR.html
+    /// [`wgpu::Surface`]: https://docs.rs/wgpu/latest/wgpu/struct.Surface.html
+    ///
+    /// ## iOS
+    ///
+    /// On iOS, the `Suspended` event is currently emitted in response to an
+    /// [`applicationWillResignActive`] callback which means that the application is
+    /// about to transition from the active to inactive state (according to the
+    /// [iOS application lifecycle]).
+    ///
+    /// [`applicationWillResignActive`]: https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1622950-applicationwillresignactive
+    /// [iOS application lifecycle]: https://developer.apple.com/documentation/uikit/app_and_environment/managing_your_app_s_life_cycle
+    ///
+    /// [`Resumed`]: Self::Resumed
     Suspended,
 
     /// Emitted when the application has been resumed.
+    ///
+    /// For consistency, all platforms emit a `Resumed` event even if they don't themselves have a
+    /// formal suspend/resume lifecycle. For systems without a standard suspend/resume lifecycle
+    /// the `Resumed` event is always emitted after the [`NewEvents(StartCause::Init)`][StartCause::Init]
+    /// event.
+    ///
+    /// # Portability
+    ///
+    /// It's recommended that applications should only initialize their graphics context and create
+    /// a window after they have received their first `Resumed` event. Some systems
+    /// (specifically Android) won't allow applications to create a render surface until they are
+    /// resumed.
+    ///
+    /// Considering that the implementation of [`Suspended`] and `Resumed` events may be internally
+    /// driven by multiple platform-specific events, and that there may be subtle differences across
+    /// platforms with how these internal events are delivered, it's recommended that applications
+    /// be able to gracefully handle redundant (i.e. back-to-back) [`Suspended`] or `Resumed` events.
+    ///
+    /// Also see [`Suspended`] notes.
+    ///
+    /// ## Android
+    ///
+    /// On Android, the `Resumed` event is sent when a new [`SurfaceView`] has been created. This is
+    /// expected to closely correlate with the [`onResume`] lifecycle event but there may technically
+    /// be a discrepancy.
+    ///
+    /// [`onResume`]: https://developer.android.com/reference/android/app/Activity#onResume()
+    ///
+    /// Applications that need to run on Android must wait until they have been `Resumed`
+    /// before they will be able to create a render surface (such as an `EGLSurface`,
+    /// [`VkSurfaceKHR`] or [`wgpu::Surface`]) which depend on having a
+    /// [`SurfaceView`]. Applications must also assume that if they are [`Suspended`], then their
+    /// render surfaces are invalid and should be dropped.
+    ///
+    /// Also see [`Suspended`] notes.
+    ///
+    /// [`SurfaceView`]: https://developer.android.com/reference/android/view/SurfaceView
+    /// [Activity lifecycle]: https://developer.android.com/guide/components/activities/activity-lifecycle
+    /// [`VkSurfaceKHR`]: https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkSurfaceKHR.html
+    /// [`wgpu::Surface`]: https://docs.rs/wgpu/latest/wgpu/struct.Surface.html
+    ///
+    /// ## iOS
+    ///
+    /// On iOS, the `Resumed` event is emitted in response to an [`applicationDidBecomeActive`]
+    /// callback which means the application is "active" (according to the
+    /// [iOS application lifecycle]).
+    ///
+    /// [`applicationDidBecomeActive`]: https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1622956-applicationdidbecomeactive
+    /// [iOS application lifecycle]: https://developer.apple.com/documentation/uikit/app_and_environment/managing_your_app_s_life_cycle
+    ///
+    /// [`Suspended`]: Self::Suspended
     Resumed,
 
     /// Emitted when all of the event loop's input events have been processed and redraw processing
--- a/src/platform_impl/web/event_loop/runner.rs
+++ b/src/platform_impl/web/event_loop/runner.rs
@@ -158,8 +158,9 @@ impl<T: 'static> Shared<T> {
     }
 
     pub fn init(&self) {
-        let start_cause = Event::NewEvents(StartCause::Init);
-        self.run_until_cleared(iter::once(start_cause));
+        // NB: For consistency all platforms must emit a 'resumed' event even though web
+        // applications don't themselves have a formal suspend/resume lifecycle.
+        self.run_until_cleared([Event::NewEvents(StartCause::Init), Event::Resumed].into_iter());
     }
 
     // Run the polling logic for the Poll ControlFlow, which involves clearing the queue
--- a/src/platform_impl/linux/wayland/event_loop/mod.rs
+++ b/src/platform_impl/linux/wayland/event_loop/mod.rs
@@ -232,6 +232,10 @@ impl<T: 'static> EventLoop<T> {
             &mut control_flow,
         );
 
+        // NB: For consistency all platforms must emit a 'resumed' event even though Wayland
+        // applications don't themselves have a formal suspend/resume lifecycle.
+        callback(Event::Resumed, &self.window_target, &mut control_flow);
+
         let mut window_updates: Vec<(WindowId, WindowUpdate)> = Vec::new();
         let mut event_sink_back_buffer = Vec::new();
 
--- a/src/platform_impl/linux/x11/mod.rs
+++ b/src/platform_impl/linux/x11/mod.rs
@@ -333,6 +333,17 @@ impl<T: 'static> EventLoop<T> {
                 callback,
             );
 
+            // NB: For consistency all platforms must emit a 'resumed' event even though X11
+            // applications don't themselves have a formal suspend/resume lifecycle.
+            if *cause == StartCause::Init {
+                sticky_exit_callback(
+                    crate::event::Event::Resumed,
+                    &this.target,
+                    control_flow,
+                    callback,
+                );
+            }
+
             // Process all pending events
             this.drain_events(callback, control_flow);
 
--- a/src/platform_impl/macos/app_state.rs
+++ b/src/platform_impl/macos/app_state.rs
@@ -302,6 +302,9 @@ impl AppState {
         HANDLER.handle_nonuser_event(EventWrapper::StaticEvent(Event::NewEvents(
             StartCause::Init,
         )));
+        // NB: For consistency all platforms must emit a 'resumed' event even though macOS
+        // applications don't themselves have a formal suspend/resume lifecycle.
+        HANDLER.handle_nonuser_event(EventWrapper::StaticEvent(Event::Resumed));
         HANDLER.set_in_callback(false);
     }
 
--- a/src/platform_impl/windows/event_loop/runner.rs
+++ b/src/platform_impl/windows/event_loop/runner.rs
@@ -393,6 +393,11 @@ impl<T> EventLoopRunner<T> {
             }
         };
         self.call_event_handler(Event::NewEvents(start_cause));
+        // NB: For consistency all platforms must emit a 'resumed' event even though Windows
+        // applications don't themselves have a formal suspend/resume lifecycle.
+        if init {
+            self.call_event_handler(Event::Resumed);
+        }
         self.dispatch_buffered_events();
         RedrawWindow(self.thread_msg_target, ptr::null(), 0, RDW_INTERNALPAINT);
     }


MarijnS95:
Nit: you don't need to have `cfg(target_os="android")` on the events because they _won't be firing anyway_, but you _do_ absolutely need them on the code that creates / initializes a graphics context on a created window.

src/event.rs
```diff
@@ -77,6 +77,17 @@ pub enum Event<'a, T: 'static> {
...
+    /// formal suspend/resume lifecycle. For systems without a standard suspend/resume lifecycle
+    /// the `Resumed` event is is always emitted after the `NewEvents(StartCause::Init)` event.
+    ///
+    /// # Portability
+    ///
+    /// It's recommended that applications should only initialized their graphics context and create
+    /// a window after they have received their first `Resumed` event. Some systems
+    /// (specifically Android) won't allow applications to create a render surface until they are
+    /// resumed (on Android this also signals that the application has an associated Surface View).
     Resumed,
```
MarijnS95, line 91: Should the docs here also link back to `` [`Self::Suspended`] `` and state that it must be handled in terms of cleaning up the "render surface" again? And repeat that on the `Suspended` docs itself?

src/platform_impl/web/event_loop/runner.rs
```diff
@@ -158,8 +158,11 @@ impl<T: 'static> Shared<T> {
     }
 
     pub fn init(&self) {
-        let start_cause = Event::NewEvents(StartCause::Init);
-        self.run_until_cleared(iter::once(start_cause));
+        // NB: For consistency all platforms must emit a 'resumed' event even though web
+        // applications don't themselves have a formal suspend/resume lifecycle.
+        let init_events =
+            iter::once(Event::NewEvents(StartCause::Init)).chain(iter::once(Event::Resumed));
```
MarijnS95, line 164: Rust nowadays implements `IntoIter` directly on static-sized arrays, right? Meaning you could write this as `[Event::NewEvents(StartCause::Init), Event::Resumed].into_iter()`.

CHANGELOG.md
```diff
@@ -58,6 +58,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - **Breaking:** On X11, device events are now ignored for unfocused windows by default, use `EventLoopWindowTarget::set_device_event_filter` to set the filter level.
 - Implemented `Default` on `EventLoop<()>`.
 - Implemented `Eq` for `Fullscreen`, `Theme`, and `UserAttentionType`.
+- All platforms can now be relied on to emit a `Resumed` event. Applications are recommended to lazily initialize graphics state and windows on first resume for portability.
```
MarijnS95, line 61: I've never dynamically created windows inside the `Resumed` event yet (nor anywhere else), is that a thing?

MarijnS95:
@rib Do you think it's worth renaming these events in light of e.g. #2118? Something closer to `WindowCreated`/`WindowDestroyed`? It doesn't map to Android's `onPause`/`onResume` bit of the cycle anyway.

rib:
> Nit: you don't need to have `cfg(target_os="android")` on the events because they _won't be firing anyway_, but you _do_ absolutely need them on the code that creates / initializes a graphics context on a created window.

:) Nit nit: A few benefits to `cfg(target_os="android")` guarding the `Resumed` events path for Android are:
1. be clear where initialization happens on iOS which also emits `Resumed` events,
2. being forwards compatible. Existing code that's guarded now (such as the linked code) can remain forwards compatible (at least as far as this specific code is concerned) with a future version of Winit that starts to emit Resumed events on all platforms - otherwise other platforms would start to initialize twice if a patch like this lands.

src/event.rs
rib: > Should the docs here also link back to `` [`Self::Suspended`] `` and state that it must be handled in terms of cleaning up the "render surface" again? And repeat that on the `Suspended` docs itself?

To some extent I was thinking of trying to keep some of the Android specific discussion separate.

Although my initial motivation came from wanting to write code that runs on Android and on desktop which this helps with, the change itself isn't really about Android. It's really just a general consistency improvement that means applications can start relying on the Resumed event across all platforms - which apps can take or leave.

That said though I did specifically add the `# Portability` notes, and this is intended to be a small stepping stone towards making the winit API guide towards being portable by default so maybe we should really flesh out the discussion of Android details here. I'm just thinking it could be good to decouple that stuff in case we start iterating on docs that depend on other nuanced details that are Android specific and also potentially subject to change in the run up to 0.27.

In terms of providing a general recommendation that avoids being too Android specific it's tricky to know what to advise right now. On Android even, technically, there's a good chance your app won't blow up if you ignore `Suspended` events and just make sure to create a new surface on the next resume (there won't be any winit redraw events while suspended which is the most likely place where things will go wrong)

src/platform_impl/web/event_loop/runner.rs
rib: Yeah, that'd sounds good - tbh, I think I just made that change on autopilot after I saw that code was already using `.chain()` for linking events but yeah in this case it's kinda silly to wrap two iter::once iters.

src/event.rs
MarijnS95: > To some extent I was thinking of trying to keep some of the Android specific discussion separate.

Oh no, I didn't mention android anywhere :sweat_smile:. If you're generalizing `Resumed` it would be stupid to not generalize `Suspended` too, right? The user doesn't "need to know" that other platforms never invalidate their surfaces.

_However_.

Perhaps this _is_ relevant when a `window` is destroyed, for which a [separate event exists](https://docs.rs/winit/latest/winit/event/enum.WindowEvent.html#variant.Destroyed) :vomiting_face:?

_Piggy-backing on my rename request_:

I think I suggested `WindowEvent::Created` at some point, which could nicely complement this already-existing `WindowEvent::Destroyed`? Would that work out for Android and do away with `Suspended`/`Resumed` completely?

---

> On Android even, technically, there's a good chance your app won't blow up if you ignore `Suspended` events and just make sure to create a new surface on the next resume (there won't be any winit redraw events while suspended which is the most likely place where things will go wrong)

I recently implemented proper `Clone` + `Drop` ownership semantics for `NativeWindow` (which is _just_ a `Surface`, i.e. a producer) and I don't remember whether it's invalid to push contents into it when there's no consumer (i.e. no display).

MarijnS95:
> :) Nit nit

I just think users are wrapping it like that right now, so perhaps this API change might already blow up if they created e.g. a swapchain, and then create it again in `Resumed` :D

MarijnS95:
> > :) Nit nit
> 
> I just think users aren't wrapping it like that right now, so perhaps this API change might already blow up if they created e.g. a swapchain, and then create it again in `Resumed` :D

CHANGELOG.md
rib: Yeah - for sure that's at least is fine with the Android, Windows and X11 backends.

The Egui app I linked above does that, as do a bunch of smaller test apps I have. This Android enabling branch I have for Bevy also does the same: https://github.com/rib/bevy/tree/android-enabling (deferring all render state + window creation until after resumed).

Although the Android backend can technically let you create a (hollow) Window even before we have a surface view that conceptually really doesn't sit well with me so I'd much rather wait until it's possible to create a render surface too before making the winit window.

Thinking out loud for ways of making Winit usage portable by default then another idea we could consider here is that create_window() can be explicitly documented to fail before the application reaches it's first `Resumed` event (and would fail consistently across platforms).

MarijnS95: Yeah, either that or do the `WindowEvent::Created` dance which hands over control of newly created windows from `winit` to the user, and disallow windows to be acquired elsewhere?

Not sure what such an API would look like though, but yes it's technically correct that an Android _window_ is only created after `Resumed` (and I think the functions like `fn size()` on `window` fail one way or another before that) and destroyed again after `Suspend`.

As mentioned above, not sure what `WindowEvent::Destroyed` means in the case of Android, and if we should use that instead of `::Suspended`.

rib:
> @rib Do you think it's worth renaming these events in light of e.g. the massive discussion in #2118? Something closer to `WindowCreated`/`WindowDestroyed`? It doesn't map to Android's `onPause`/`onResume` bit of the cycle anyway.

Something to consider here is that the existing `Suspended` and `Resumed` are also supported on iOS so any renaming of the existing events will affect the iOS backend too.

I think there's probably still some value in having some portable lifecycle events like "suspend" and "resume" that have reasonably practical mappings on Android and iOS - even though there's some nuance to how we define "suspended" on Android. (and as an asside I recall thinking the android backend should probably be mapping "suspended" to the logical OR of 'lifecycle suspended' and 'window destroyed', instead of directly to the 'window destroyed' state).

Since the mapping is a bit fuzzy if you need to handle some platform specific details I tend to think it would also be good if there was a practical way to also notified about Android or iOS specific lifecycle events. Maybe instead of having to expose any platform-specific details there could be some general `PLATFORM_LIFECYCLE_CHANGED` event which just provides the event loop wake up and then assumes apps can access that state easily via a platform-specific API.

What's extra funky here though is that the `window destroyed` / ` created` events are not conceptually lifecycle events on Android.

rib:
> > :) Nit nit
> 
> I just think users weren't wrapping it like that, so perhaps this API change might already blow up if they created e.g. a swapchain, and then create it again in `Resumed` :D
> 
> They certainly don't/didn't _need_ to, which is the word that was used in the description that I'm nitting about 😅 - but they _did_ need to have it around their code _outside_ the event loop that (indirectly) calls into `fn raw_window_handle()` or run into a nice `panic` on Android.

I concede the nit battle :) yeah, existing apps wouldn't necessarily _need_ to `cfg()` guard mobile/android specific `Resumed` code if they rely on the Resumed event not being emitted and just guard the non-mobile code.

Since I've been guarding the android-specific Resumed path in some apps I've written recently (just to give me one fewer thing to consider in case we do end up landing a change like this) then my wording was probably just echoing what I've felt I needed to do to write portable, forwards-compatible winit code.

src/event.rs
rib: yeah, I took it as being android specific because it's the only platform where it could be prudent to destroy any rendering surfaces you have.

I'd prefer to keep any potential reworking of suspend as a separate topic/issue if possible for a number of reasons:

1. We're solving a known portability problem by consistently emitting `Resumed` events - and the same can't really be said for `Suspended`. (apps running on Android may want to destroy render surfaces but it's ok to not do that on other platforms which is different to the `Resumed` situation where we have mutually exclusive paths for initialzing Android vs other platforms)
2. One reasonable interpretation of the resulting behaviour is that most platforms will never actually "suspend" and so it remains reasonable that there is no `Suspended` event, except on mobile platforms that have a well defined lifecycle. Conceptually it's harder to argue that these other platforms don't "resume" though, so I don't see the slight asymmetry as a major concern right now (not in comparison to the benefit of at least making `Resumed` consistent).
3. For non-mobile platforms it may be technically impossible to guarantee that a `Suspended` event will be delivered because the operating doesn't provide the required lifecycle hooks that exist on Android / iOS (e.g. processes can simply be killed via signals)

rib: This discussion is a good example for highlighting why the Android NDK terminology is really unhelpful unfortunately...

What the Android NDK calls a `NativeWindow` is a `SurfaceView` in Java - and even that isn't the clearest name for explaining what the object represents.

Conceptually a SurfaceView on Android is somewhat comparable to a rendering API surface, such as an EGLSurface. A SurfaceView encapsulates the swap chain for rendering to the window (or rather for getting buffers that can be given to Surface Flinger). Android technically even has a Window API in Java which could even be considered the true "native window" but since you don't have to explicitly create a SurfaceView for a Window then it's not a distinction that ever comes up in practice.

The main difference between Android and other OSs, is that other OSs let you create your render surfaces (and allocate/configure a swap chain) manually for a given native window but on Android they effectively create the render surface for you. When you create an EGLSurface or similar on Android there is a bit of a white lie involved because you don't really own the backing of the surface like you would on other OSs. That's the crux of why Android is awkward here you have to be prepared for your EGLSurface (or equivalent) to become invalid. If comparing with other OSs then it's _not_ equivalent to the native window being destroyed - so terminology is important here.

Some consequences of this to consider are:
- Just because the Android NDK tells us that our 'window' has been destroyed / created doesn't mean that we should use the same terminology within the public API of Winit. In particular; if we want to expose the Android NDK "window created" and "window destroyed" events via Winit I would strongly recommend avoiding the word "window" and probably advocate for events like `WindowEvent::SurfaceCreated` `WindowEvent::SurfaceDestroyed` to avoid a lot of confusion with how they related to `WindowEvent::Destroyed`.
- Since Android's `NativeWindow` is actually just a rendering surface then certain questions about a "window" should be answerable without a surface - such as querying width/height which doesn't necessarily have to be done via the SurfaceView (conceptually the same as being able to query the width/height of an X11 window or wl_surface without necessarily looking at the width/height of an EGLSurface)

CHANGELOG.md
rib: > Yeah, either that or do the `WindowEvent::Created` dance which hands over control of newly created windows from `winit` to the user, and disallow windows to be acquired elsewhere?
> 
> Not sure what such an API would look like though, but yes it's technically correct that an Android _window_ is only created after `Resumed` (and I think the functions like `fn size()` on `window` fail one way or another before that) and destroyed again after `Suspend`.
> 
> As mentioned above, not sure what `WindowEvent::Destroyed` means in the case of Android, and if we should use that instead of `::Suspended`.

I think I've addressed how I currently see this in this comment: https://github.com/rust-windowing/winit/pull/2331#discussion_r895666043

Going in this kind of direction I'd probably call your `WindowEvent::Created` `WindowEvent::SurfaceCreated` and the opposite `WindowEvent::SurfaceDestroyed` where it would become much easier to disambiguate from `WindowEvent::Destroyed`.

The challenge with that direction though is that Winit isn't currently responsible of render surfaces - it assumes all OSs can manually allocate surfaces for a window which basically isn't true on Android. That would make it hard know how these events would be in any way cross-platform. On Android it would make sense that these would deliver a handle for the the SurfaceView but there's no equivalent on other platforms - on other platforms it's just left to APIs like wgpu to allocate surfaces and it's assumed that they are fully managed/owned outside Winit.

Technically it's quite tricky, I'd say, to have an abstraction like Winit that doesn't claim any responsibility for render surfaces and for synchronizing swap buffers too so I'd certainly think it's worth questioning if that's the best design, but yeah at this point we're opening a much bigger can of worms, so it probably makes sense to discuss this stuff in a separate issue.

rib:
Ok, with the updates I've pushed I bit the bullet and wrote much more comprehensive documentation for both the `Suspended` and `Resumed` events that provide some general guidance about writing portable applications as well as providing Android and iOS specific information so that developers can understand how these events map to OS-specific lifecycle details.

I had been thinking it could be good to try and deal with documentation improvements separately but while it makes sense to try and some general recommendations about handling these events it's a good opportunity to add more details too.

I've updated the commit message in line with these changes, and some of the comments above and I've now marked the patch as Resolving #2185

I updated the web backend change as suggested by @MarijnS95 above and at least smoke tested the change via `cargo run-wasm --example web`

Although this has provoked some interesting discussions around some of the Android specific thorns I think for now most of that looks like it probably deserves follow up issues + discussions and hopefully doesn't need to block making this incremental change for improved consistency first.

MarijnS95:
> Something to consider here is that the existing `Suspended` and `Resumed` are also supported on iOS so any renaming of the existing events will affect the iOS backend too.

Right, that makes this situation a bit more complicated as I have **zero** experience with iOS. Are these events purely used for window/surface _creation and destruction_ there too?

> I think there's probably still some value in having some portable lifecycle events like "suspend" and "resume" that have reasonably practical mappings on Android and iOS

Correct, hence, my linking to #2118 - that PR author also got thrown off by the unfortunate naming, as you said:

> What's extra funky here though is that the `window destroyed` / ` created` events are not conceptually lifecycle events on Android.

They are not, which is why I think we should have separate events that represent window creation/destruction. Android doesn't ever raise `WindowEvent::Destroyed` which should be used in place of `Suspended`. Then we can have separate events that propagate _actual lifecycle_ events?

---

> Since I've been guarding the android-specific Resumed path in some apps I've written recently (just to give me one fewer thing to consider in case we do end up landing a change like this) then my wording was probably just echoing what I've felt I needed to do to write portable, forwards-compatible winit code.

Exactly, I'll have to go over a bunch of applications and make sure that we're either guarding these or move the initialization for _all platforms_ into `::Resumed`.

----

> Although this has provoked some interesting discussions around some of the Android specific thorns I think for now most of that looks like it probably deserves follow up issues + discussions and hopefully doesn't need to block making this incremental change for improved consistency first.

Fair enough, if your documentation looks good we should get this in as it is clearly an improvement over the status-quo - and separately **before `winit 0.27` please** discuss whether we should rework the events entirely. Repeating what we both seem to agree on: these aren't strictly speaking lifecycle events (not at all, cc #2118), but window created and destroyed events that require the user to perform some resource handling.

MarijnS95:
https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1622956-applicationdidbecomeactive / https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1622950-applicationwillresignactive (these callbacks map to `Resumed`/`Suspended` resp.) mention to throttle up/down framerates, nothing said about needing to create/destroy surfaces.

More fuel for switching these to `WindowEvent::{Created, Destroyed}` on android and pull in `onPause`/`onResume` handling from #2118 that then maps to `Suspended`/`Resumed`?

src/event.rs
MarijnS95: > * and probably advocate for events like `WindowEvent::SurfaceCreated` `WindowEvent::SurfaceDestroyed` to avoid a lot of confusion with how they related to `WindowEvent::Destroyed`.

:+1: Yes, I guess it is better to add separate events for this then. The window (activity, really) technically stays alive but may pause because the user stops interacting with it, and separately from those lifecycle events the backing buffer of our `Surface` _producer_ may become invalid/recreated.

Then later we may map activity destruction to this (cc some thoughts about recently running into multiple `NativeActivity` _instances_ within the same executable process making the whole Rust Android stack fall apart).

CHANGELOG.md
MarijnS95: > On Android it would make sense that these would deliver a handle for the the SurfaceView but there's no equivalent on other platforms

`RawWindowHandle`?

src/event.rs
```diff
@@ -74,9 +74,93 @@ pub enum Event<'a, T: 'static> {
...
+    /// Winit does not currently try to emit pseudo `Suspended` events before the application
+    /// quits on platforms without an application lifecycle.
+    ///
+    /// In case of corner cases with how platform-specific events are mapped to Winit `Suspended`
+    /// and `Resumed` events it's recommended that applications be able to gracefully handle redundant
+    /// (i.e. back to back) `Suspended` or `Resumed` events.
+    ///
+    /// Also see [`Resumed`][Self::Resumed] notes.
+    ///
+    /// # Android
```
MarijnS95, line 92: Since this fits under the `Portability` umbrella, should we make this h2 with two `##`?

src/event.rs
```diff
@@ -74,9 +74,93 @@ pub enum Event<'a, T: 'static> {
...
+    /// On Android specifically, the `Suspended` event relates to two things:
+    /// 1. The application is paused (according to the [Activity lifecycle][android_ls])
+    /// 2. Or the application has had it's associated SurfaceView destroyed (which is expected
+    /// to closely correlate with being paused)
+    ///
+    /// Applications that need to run on Android should assume their SurfaceView has been
+    /// destroyed, which indirectly invalidates any existing render surfaces that may have been
+    /// created outside of Winit (such as an `EGLSurface`, `VkSurfaceKHR` or [`wgpu::Surface`][wgpu_surface]).
+    ///
+    /// [wgpu_surface]: https://docs.rs/wgpu/latest/wgpu/struct.Surface.html
```
MarijnS95, line 103: You can also make this `` [`wgpu::Surface`] `` and remove the `[wgpu_surface]` link target entirely. Same for `android_ls`.

Shall we also add links for `EGLSurface` and (Ash's) `VkSurfaceKHR`?

src/event.rs
```diff
@@ -74,9 +74,93 @@ pub enum Event<'a, T: 'static> {
...
+    ///
+    /// After being `Suspended` on Android it's recommended to immediately drop all render
+    /// surfaces which may be re-created when the application is next `Resumed`
```
MarijnS95, line 106:
```suggestion
    /// surfaces which may be re-created when the application is next `Resumed`.
```

Not sure about "recommended", it's pretty much required if no-one keeps the `NativeWindow` alive (`winit` won't) otherwise leading to UB?

rib:
> > Something to consider here is that the existing `Suspended` and `Resumed` are also supported on iOS so any renaming of the existing events will affect the iOS backend too.
> 
> Right, that makes this situation a bit more complicated as I have **zero** experience with iOS. Are these events purely used for window/surface _creation and destruction_ there too?

It's the other way around - they are purely iOS lifecycle based - and I'm fairly sure that there's no equivalent requirement to destroy and recreate render surfaces on iOS like there is on Android.

With the way they are currently mapped to the iOS lifecycle events though I have a feeling that it could be slightly better to map `Suspended` to iOS's "inactive" state. Currently `Suspended` gets mapped to "about to become inactive _soon_" on iOS. The subtlety here is that on iOS applications aren't required to stop rendering at this point (though they may reduce their frame rate) and I think perhaps Winit's `Suspended` event could be more clearly documented if it was a platform-agnostic way of indicating when rendering is blocked on Android and iOS. On iOS I think that would be when the app really becomes "inactive".

rib:
> > Since I've been guarding the android-specific Resumed path in some apps I've written recently (just to give me one fewer thing to consider in case we do end up landing a change like this) then my wording was probably just echoing what I've felt I needed to do to write portable, forwards-compatible winit code.
> 
> Exactly, I'll have to go over a bunch of applications and make sure that we're either guarding these or move the initialization for _all platforms_ into `::Resumed`.

Yeah, for sure, if we can land this change I'd be interested in seeing some (all ideally) of the examples be updated to demonstrate lazy initialization of graphics state via the `Resumed` event.

Ideally the first-party examples in the Winit repo can try to demonstrate how to write portable code that will run on all supported platforms - even if there are things that could be improved about the API to make it harder to write non-portable code.

Some of the examples here, demonstrate winit applications that can run on both Android and Windows:
https://github.com/rib/agdk-rust/tree/main/examples as well as this bluetooth UI: https://github.com/rib/bluey/tree/main/bluey-ui

It would be good to verify the same lazy initialization via `Resumed` also works as expected on web and iOS.

kchibisov:
>I currently still think it's reasonable to keep the Suspended and Resumed events in Winit and documenting that they delimit when applications are blocked from rendering which is useful/meaningful on Android and iOS.

I think resume if fine, however could we name them `Resume -> CouldStartDrawing`, `Suspend -> Suspend(action)`.

where
```rust
enum SuspendAction {
  ClearGraphicsContenxt,
  None,
  ThrottleDrawing
}
```

That was I think you can map it to anything and it'll be the most portable thing, since desktop platforms will have `None` or `ThrottleDrawing`, android will have `ClearGraphicsContext`, and ios should `ThrottleDrawing`.

The issue when the window should be created is also related to macOS, since the window should e created when the event loop is running, otherwise it could miss on some features iirc.

The problem with `Resumed`, since it's not clear for other platforms, suspend could be left as is.

src/event.rs
rib: > > * and probably advocate for events like `WindowEvent::SurfaceCreated` `WindowEvent::SurfaceDestroyed` to avoid a lot of confusion with how they related to `WindowEvent::Destroyed`.
> 
> 👍 Yes, I guess it is better to add separate events for this then. The window (activity, really) technically stays alive but may pause because the user stops interacting with it, and separately from those lifecycle events the backing buffer of our `Surface` _producer_ may become invalid/recreated.
> 
> Then later we may map activity destruction to this (cc some thoughts about recently running into multiple `NativeActivity` _instances_ within the same executable process making the whole Rust Android stack fall apart).

Worried that we're spawning a lot of side discussions here but yeah, this was something I was unsure about recently with the activity/glue crates I was experimenting with. Actually my first pass would have supported multiple activities but I ended up trying to emulate the global getter API from ndk_glue to made it easier to port some existing code, but I think it would be relatively easy to re-instate support for multiple activities and also enable support for this in Winit too.

The main requirements are:
1. remove any global getters for android app/activity state
2. the main() entrypoint on Android needs to accept an application pointer
3. The application pointer needs to be given to Winit when creating an event loop
4. Downstream winit users need to be able to query the application from the event loop (similar to getting the X11/wayland display)

when I say "application pointer" I essentially mean an opaque struct that encapsulates the activity. E.g. in my `native-activity` glue crate that would correspond to this [`AndroidApp`](https://github.com/rib/agdk-rust/blob/864d782c241aa7e403786774ebf057ec134f93be/native-activity/src/lib.rs#L284) struct / API.

The main tricky thing is then with defining the C ABI entry point (`android_main()`) which wouldn't be quite as ergonomic without the global static API to get the `AndroidApp` since it would have to be passed as a `[transparent]` pointer - but still that might not be too bad.

MarijnS95:
> and documenting that they delimit when applications are blocked from rendering

There's still quite a difference between "hey stop rendering now" and "hey you must recreate your graphics context/surface otherwise things don't work", and I think @kchibisov's suggestion solves that.

> The problem with `Resumed`, since it's not clear for other platforms, suspend could be left as is.

We can have an `enum ResumeAction` too with the antonyms of what you suggested?

src/event.rs
MarijnS95: Yeah we're spawning a side-discussion here. I was doing the same with https://github.com/rust-windowing/android-ndk-rs/pull/223#issuecomment-1039225349 but that branch never reached main - in short because it requires your point 3. i.e. `The application pointer needs to be given to Winit when creating an event loop` which I think is a fair requirement to prevent a lot of mishaps _at compile-time_.

Though my idea was to actually pass `ndk::NativeActivity` around - it's an opaque pointer internally but if your `android_main()` is wrapped by `ndk_glue 0.6` and winit uses `ndk_glue 0.5`. We'd have to discuss this further in light of making `winit` support multiple Android backends concurrently (as well how _and where_ to design and maintain a consistent API, and where to place all the crates - `android-ndk-rs` has my preference of course).

IIRC I also "got stuck" at storing the rest of our Rust state in its `instance` `void *` userdata, but recently found an ancient comment in the `ndk` that it was used by a previous `android_native_app_glue`-something-something crate as well - we just have to make sure that the user can't touch it if we do.

MarijnS95:
@kchibisov Also fine to put that burden on the user. They can just have an `if surface.is_none() { surface = new_surface(Window as dyn RawWindowHandle)` of sorts.

kchibisov:
I mean you don't need it as non-exhaustive. Just 3 possible things.

1. Destroy context.
2. Don't render anymore.
3. Do nothing.

Pretty sure application that's **enough** to be crossplatform here.

MarijnS95:
Shall we move all this discussion over to #2337 then which was specifically opened for this (perhaps reword the title a bit), and just merge this PR as it clearly addresses platform-divergence and a documentation deficit?

MarijnS95:
@rib Yeah exactly my point, this PR is good as it is and better describes the status-quo + handles it more conveniently across platforms, but we definitely have more work to do after that to improve the status-quo instead :)

src/event.rs
rib: yep the short hand wgpu::Surface linking makes sense.

There isn't really a canonical link for `EGLSurface` - the official EGL spec is a Khronos PDF. The Vulkan spec is at least published as HTML so could maybe link to the spec but not sure it's that important for either really.

Not sure about the "Same for `android_ls`" comment - that's a link about the Android activity life cycle so there wouldn't be a shorthand way of making that link?

src/event.rs
rib: yeah I wrote "recommended" In the sense that you could maybe technically avoid doing anything that tries to interact with underlying surface view/`NativeWindow` and you could might get away with that but yeah I might be wrong about that. I'll say it's a "must".

kchibisov:
>Step 1. lets introduce a consistent place where apps can (optionally) initialize render state by having Resumed be consistently emitted on all platforms. (unless there is disagreement on that specifically). It's a minimal change that doesn't solve all portability concerns re: Android but it does have an immediate / practical benefit.

I agree that we should at least have an event when creating a window/graphics context, etc, is safe cross platform. The problem is that I don't like `Resumed` name for it. I'd probably go for `CouldInitializeGraphics`, `CouldStartDrawing` or something like that.

Also, I think certain platforms, at least Wayland has wrong impl of `StartCause::Init` and though `Resume`, since if you call `run_return` twice you'll have `Resume` twice, which is wrong. Those should emit this event only once by prepending those in the queue or by using `Once`.

I can send patch for Wayland/X11 once once we agree on name, since I really don't like resume, because it's not clear what we resume from when we just created an event loop.

src/event.rs
rib: I also added the clarification that the render surfaces should be dropped before the the event callback is complete, since that may related to how the backend implementation synchronizes with Java:

```rust
    /// After being `Suspended` on Android applications must drop all render surfaces before
    /// the event callback completes, which may be re-created when the application is next `Resumed`
```

src/event.rs
MarijnS95: You can make that an `[Android lifecycle]: https...` too.

Yeah, link Vulkan if easy, I tried linking (E)GL in the ndk (see recent doc improvement PRs) and it's a bit of a hit and miss. Only need this if it's easy and makes it more clear to the user what you're talking about / referencing.

src/event.rs
MarijnS95: Yeah I like that wording.

Not sure either, I don't think Vulkan/EGL "acquire" the nativewindow so it _might_ be UB - but it's not if the caller acquires (by means of cloning) a NativeWindow externally. Oh my, we don't need to document that case.

MarijnS95: The main takeaway - I guess - is that a render surface can only be created after Resumed, and that you must do that every time. A render surface (producer) _may_ stay valid after Suspended (re. acquire above) but it'd be useless if android gives you a new, different producer. Then users wonder why their screen stays black :)

rib:
> I mean you don't need it as non-exhaustive. Just 3 possible things.
> 
> 1. Destroy context.
> 2. Don't render anymore.
> 3. Do nothing.
> 
> Pretty sure application that's **enough** to be crossplatform here.



> > Step 1. lets introduce a consistent place where apps can (optionally) initialize render state by having Resumed be consistently emitted on all platforms. (unless there is disagreement on that specifically). It's a minimal change that doesn't solve all portability concerns re: Android but it does have an immediate / practical benefit.
> 
> I agree that we should at least have an event when creating a window/graphics context, etc, is safe cross platform. The problem is that I don't like `Resumed` name for it. I'd probably go for `CouldInitializeGraphics`, `CouldStartDrawing` or something like that.

but that's the existing name and this patch wasn't really looking to bring that into question - esp considering that nothing about the event's semantics are being changed :/


> 
> Also, I think certain platforms, at least Wayland has wrong impl of `StartCause::Init` and though `Resume`, since if you call `run_return` twice you'll have `Resume` twice, which is wrong. Those should emit this event only once by prepending those in the queue or by using `Once`.

The patch consistently applies the rule that `Resumed` will be emitted after `StartCause::Init` on all platforms, including for Wayland which is the main concern for this PR. Although I think I've seen some inconsistency with `StartCause::Init` emission that should be an orthogonal issue, right?

I'd prefer to not unnecessarily block this change on an unrelated issue if possible, please.

> 
> I can send patch for Wayland/X11 once once we agree on name, since I really don't like resume, because it's not clear what we resume from when we just created an event loop.

Yeah, this same question applies on Android and iOS - technically when the app starts you resume "from... nothing" but I don't see any big issue with that - it's still simple to understand, and it's just a limitation of the precision of language at that point. For that concern specifically then arguably the iOS terms of "active" and "inactive" are slightly better because they are just nouns for states instead of describing a transition - but that still feels like a very separate issue from this PR, since this question already exists in Winit as-is today.

The app is still "resuming" in the same sense that an Android or iOS application "resumes" when it first starts - it's just that technically the app will never be suspended on platforms without an application lifecycle.

This PR doesn't introduce the names - although I admit I personally find the names are reasonable from my pov.

My big concern here is that discussing naming can be a notoriously time consuming process which can lead to endless disagreements and going around in circles arguing the semantics of words. English often isn't very precise and names are often a compromise so I really hope it can be kept as a separate topic.

kchibisov:
>The patch consistently applies the rule that Resumed will be emitted after StartCause::Init on all platforms, including for Wayland which is the main concern for this PR. Although I think I've seen some inconsistency with StartCause::Init emission that should be an orthogonal issue, right?

Right, I don't want to look into it here.

>Yeah, this same question applies on Android and iOS - technically when the app starts you resume "from... nothing" but I don't see any big issue with that - it's still simple to understand, and it's just a limitation of the precision of language at that point.

I think we can keep it like that... The issue is that on other platforms it's not clear at all what `Resumed`, since before this event was never sent on those. If you call `run_return` multiple times should it send `Resumed` event again and again? From what I can see with android -> no, it shouldn't, but the end users will unlikely to figure it out without reading much into the docs of that event.

The problem with consistently delivering `Resume` that for android the name was fine, since it has `Suspend`, but for the rest(especially for desktop) keeping the name like that.

I'm not trying to block  this PR or anything, but unifying name for `Resume` when unifying it across the platforms seems like a natural move? Basically what this event maps to `CanCreateWindow` or similar.

Would like to hear from @maroider on that regard.

rib:
> > The patch consistently applies the rule that Resumed will be emitted after StartCause::Init on all platforms, including for Wayland which is the main concern for this PR. Although I think I've seen some inconsistency with StartCause::Init emission that should be an orthogonal issue, right?
> 
> Right, I don't want to look into it here.
> 
> > Yeah, this same question applies on Android and iOS - technically when the app starts you resume "from... nothing" but I don't see any big issue with that - it's still simple to understand, and it's just a limitation of the precision of language at that point.
> 
> I think we can keep it like that... The issue is that on other platforms it's not clear at all what `Resumed`, since before this event was never sent on those. If you call `run_return` multiple times should it send `Resumed` event again and again? From what I can see with android -> no, it shouldn't, but the end users will unlikely to figure it out without reading much into the docs of that event.

Right, this is quite a specific corner case that I honestly never considered. Have you seen that come up before where an application will exit its event loop and then run another one? 

Imagining crafting some kind of test of this case on Android then it probably argue that it would _also_ make sense on Android to send a pseudo `Resumed` event just like other platforms because it's really part of the Winit API guarantees that an applications should be able to put initialization logic in their `Resumed` event handler. So if you made a contrived Android application that could quit its Winit event loop and then run a second one well that second run might fail if the `Resumed` event isn't run.

I think that's tricky to imagine happening in practice on Android, but yeah with the way I have been recently proposing Android tweak it's logic for emitting `Suspended` and `Resumed` events I think it would technically end up being more consistent with e.g. Wayland + X11 because it would (naturally) end up emitting pseudo `Resumed` events in this situation you're thinking of.

rib:
> The problem with consistently delivering `Resume` that for android the name was fine, since it has `Suspend`, but for the rest(especially for desktop) keeping the name like that.

The way I think about it is that 'conceptually any winit app could be suspended (blocked from rendering) but it just so happens (as an implementation detail) that desktop applications will probably never be suspended.'

It's plausible that in some situations, e.g. via some wayland extension protocol the backend could later support suspending apps but in all likelihood they probably won't end up suspended.

> 
> I'm not trying to block this PR or anything, but unifying name for `Resume` when unifying it across the platforms seems like a natural move? Basically what this event maps to `CanCreateWindow` or similar.

My current mental model for what Suspended/Resumed mean is that rendering is disabled or enabled, and I think those can be quite simple semantics to document and understand.

Technically, right now, the semantics are not quite that simple (and they aren't documented as such either) if you look at the platform-specific details but as a follow up to this I think we can make some very small changes to the iOS backend and the Android backend (based on comments above) which will give us that additional consistency.

The idea of initializing rendering state in the resumed event can be seen as a secondary fallout from that first definition because it just becomes logical that apps would only initialize their rendering state at the first point they are told that rendering is enabled.

But yeah, this is where we also start getting into the various ideas about how we can evolve the Winit API so that instead of this being an implied, recommended, location to initialize rendering state then maybe it can somehow become (effectively) impossible to try and initialize render state beforehand - looking for some design that's ergonomic and leads to all Winit apps being portable between mobile, desktop and web.

All this is still up for discussion though via follow up issues - and yeah there's a chance we could do something more drastic with the Suspended/Resumed events in support of that but it's also a much more open-ended discussion so I'm just trying to take a baby step with this first PR.

src/event.rs
rib: Yeah, I'd be interested to pick up a discussion about building on what I found with the `native-activity` and `game-activity` crates.

Those two crates demonstrate that it's possible to define a compatible API that can support both NativeActivity and GameActivity (and theoretically future activities such as a hypothetical `RustActivity`) and I think it would be relatively simple to tweak and remove the static global `AndroidApp` getter to also be able to support multiple activities. I think I even added winit API to pass an `AndroidApp` when creating an event loop in an iteration of that work.

The next step I had roughly been imagining was defining an `android-activity` crate that would effectively be a single point of configuration for what specific Activity your app is uses and it would basically be like a MUX for the real implementation (such as `native-activity` or `game-activity` or other-activity...

Maybe this could be done under `android-ndk-rs` somewhere, or under `rust-windowing`  - though I'd probably see it as something separate from `ndk-glue` at this point (though we could potentially re-purpose `ndk-glue` as that kind of MUX crate.

kchibisov (approved):
I'll be fine with resumed event for now then.

@MarijnS95 do you have anything left here?

maroider:
> I'm not trying to block this PR or anything, but unifying name for `Resume` when unifying it across the platforms seems like a natural move? Basically what this event maps to `CanCreateWindow` or similar.
> 
> Would like to hear from @maroider on that regard.

I don't feel too strongly about the name right now, although I do admit that we could (and perhaps should) come up with a better name than `Resumed`.

---

As for the PR, I think it's *fine* to merge it, but I'm not sure about the overall direction. From what I understand, the main reason we're exposing `Event::Resumed` and `Event::Suspended` is because of this "surface lifetime" tracking, right?

Assuming that's the case, well, I've been thinking a bunch about whether we can move the responsibility for all this "surface lifetime" handling downstream, so that crates like `wgpu` handle it "behind the scenes". I'm not entirely sure if that would be the *correct* direction either, but it seems to me that such an arrangement would be nicer for users. My thinking so far is that this would take the form of some added API surface in `raw-window-handle` (or maybe another "interface" crate).

rib:
I see `Suspended` and `Resumed` as being more general / cross-platform than these surface lifetime issues we have on Android - they are a high-level a reflection of the need to pause and resume rendering for applications as part of some system-defined lifecycle (particularly on mobile OSs).

Android defines an "Activity lifecycle": https://developer.android.com/guide/components/activities/activity-lifecycle
and iOS defines an application lifecycle: https://developer.apple.com/documentation/uikit/app_and_environment/managing_your_app_s_life_cycle

So in that sense they are equally relevant to iOS as they are on Android and iOS does not have any surface lifetime challenges like Android.

Since Winit provides an abstraction for building cross-platform applications it makes sense that this notion that an application may become paused/suspended/inactive (which means that rendering is temporarily blocked) would be exposed in a very general / cross-platform way.

This patch is about extending that notion across all platforms consistently - even though we know in practice we wouldn't expect to see a desktop application get suspended.

In some ways the fact that this also practically helps us deal with the surface lifetime issues on Android in a more portable way is something of a distraction for this PR (I think this improvement in consistency would be a valid change even if it wasn't helping us manage surfaces on Android).

MarijnS95:
> If you call `run_return` multiple times should it send `Resumed` event again and again? From what I can see with android -> no, it shouldn't, but the end users will unlikely to figure it out without reading much into the docs of that event.

No, I don't think so. The selling point for `run_return` seems to be able to jump in and out of "the loop", polling for events for a while and then jumping back to ones own code. We wouldn't fire `Resumed` again just as we wouldn't fire all keyboard/mouse events "recorded" thus far. This is a one-off event that only happens once, I think it's confusing to dispatch it to the user multiple times (every time they "poll" for new events).

Does this PR currently do that in any way?

---

At risk of sounding like a broken record: I don't think we should abstract this very real problem of recreating a graphics surface behind some "oh maybe you want to throttle or stop rendering for a bit" event. It's a platform-specific detail, but one that can't be trivially handled by `winit` but must be handled by the user.

Just in case I created a little demonstration in `glutin` to showcase a black screen when the user doesn't recreate their graphics surface on a subsequent `Suspended` + `Resumed` combination: 
https://github.com/MarijnS95/glutin/commit/android-without-surface-recreation. It seems EGL holds a reference on the `NativeWindow` (Java `Surface`) internally as no faults/errors occur after Android is technically allowed to destroy it, but we're still rendering to it - but that may also just be my platform.

As such it's a real, application breaking "defect" that must be handled by the user with more thought than "oh let's not render" - in that case we could have a flag on `MainEventsCleared` and/or not send `RedrawRequested`, or whatever the recommended way/event is nowadays to perform rendering in.

Sure, I am all for having this properly abstracted away in high-level rendering wrappers like `wgpu` and `glutin` (though not sure how to trivially do that besides telling the user to call some kind of `handle_winit_events(&event)` helper in their loop yet). But how would I deal with that in my Vulkan renderer based plainly on top of Ash? Can I still get access to properly labeled, clear events in my event loop?

---

Regarding `raw-window-handle`: perhaps? Not sure what that API would look like, as we'd have to replace the inner pointer with a new pointer which is odd - and doesn't map cleanly when working with "regular" Android `Surface`s in a non-`NativeActivity`, non-`winit` app (you either have the surface, or you don't - and this is coordinated from Java callbacks). Being able to hold a (`parking_lot` mapped-) lock may help Android in certain scenarios though

---

And repeating myself again: yes, I also feel that this PR is not the right place to address this (hence opened a new issue), but I still think we can do a lot better here. As such I'm happy for this PR to make it in as-is.

---

> But yeah, this is where we also start getting into the various ideas about how we can evolve the Winit API so that instead of this being an implied, recommended, location to initialize rendering state then maybe it can somehow become (effectively) impossible to try and initialize render state beforehand - looking for some design that's ergonomic and leads to all Winit apps being portable between mobile, desktop and web.

:+1: on this!

src/event.rs
```diff
@@ -74,9 +74,94 @@ pub enum Event<'a, T: 'static> {
...
+    ///
+    /// Not all platforms support the notion of suspending applications, and there may be no
+    /// technical way to guarantee being able to emit a `Suspended` event if the OS has
+    /// no formal application lifecycle (currently only Android and iOS do). For this reason
+    /// Winit does not currently try to emit pseudo `Suspended` events before the application
+    /// quits on platforms without an application lifecycle.
+    ///
+    /// In case of corner cases with how platform-specific events are mapped to Winit `Suspended`
+    /// and `Resumed` events it's recommended that applications be able to gracefully handle redundant
+    /// (i.e. back to back) `Suspended` or `Resumed` events.
```
MarijnS95, line 88: Is this currently something we're bothered by?

src/event.rs
```diff
@@ -74,9 +74,94 @@ pub enum Event<'a, T: 'static> {
...
+    ///
+    /// Also see [`Resumed`][Self::Resumed] notes.
+    ///
+    /// ## Android
+    ///
+    /// On Android specifically, the `Suspended` event relates to two things:
+    /// 1. The application is paused (according to the [Activity lifecycle])
+    /// 2. Or the application has had it's associated SurfaceView destroyed (which is expected
```
MarijnS95, line 96: Callback info is useful here, perhaps (at the very least backticks):

```suggestion
    /// 2. Or the application has had it's associated [`SurfaceView`] destroyed (which is expected
```

And then:

```markdown
[`SurfaceView`]: https://developer.android.com/reference/android/view/SurfaceView
```

below.

src/event.rs
```diff
@@ -74,9 +74,94 @@ pub enum Event<'a, T: 'static> {
...
 
     /// Emitted when the application has been suspended.
+    ///
+    /// # Portability
+    ///
+    /// Not all platforms support the notion of suspending applications, and there may be no
+    /// technical way to guarantee being able to emit a `Suspended` event if the OS has
+    /// no formal application lifecycle (currently only Android and iOS do). For this reason
+    /// Winit does not currently try to emit pseudo `Suspended` events before the application
+    /// quits on platforms without an application lifecycle.
```
MarijnS95, line 84: I'm personally a big fan of (over)eagerly applying intradoc-links, not sure if _all_ of these should be referencing to `Self::Suspended` and `Self::Resumed`? Also below.

src/event.rs
MarijnS95, line 96: Also, I think you meant to use `its` possessive here, rather than `has had _it is_ associated SurfaceView...`? In other places it is more clear to write out `it is` over `it's` - I remember certain Khronos repos having a CI to disallow these contractions explicitly :)

src/event.rs
```diff
@@ -74,9 +74,94 @@ pub enum Event<'a, T: 'static> {
...
+    /// Also see [`Resumed`][Self::Resumed] notes.
+    ///
+    /// ## Android
+    ///
+    /// On Android specifically, the `Suspended` event relates to two things:
+    /// 1. The application is paused (according to the [Activity lifecycle])
+    /// 2. Or the application has had it's associated SurfaceView destroyed (which is expected
+    /// to closely correlate with being paused)
+    ///
+    /// Applications that need to run on Android should assume their SurfaceView has been
```
MarijnS95, line 99: Backticks for `SurfaceView`, and make it a link too if comment from above is applied.

src/event.rs
```diff
@@ -74,9 +74,94 @@ pub enum Event<'a, T: 'static> {
...
+    /// destroyed, which indirectly invalidates any existing render surfaces that may have been
+    /// created outside of Winit (such as an `EGLSurface`, [`VkSurfaceKHR`] or [`wgpu::Surface`]).
+    ///
+    /// After being `Suspended` on Android applications must drop all render surfaces before
+    /// the event callback completes, which may be re-created when the application is next `Resumed`.
```
MarijnS95, line 104: This seems to repeat, but also slightly contradict the above paragraph. Has my render surface "already been destroyed/invalidated", or will that happen _after completing the event callback_? Perhaps merge the two and make it clear that it is the second (as long as the lock is held, see my other PR).

src/event.rs
```diff
@@ -74,9 +74,94 @@ pub enum Event<'a, T: 'static> {
...
+    /// [Activity lifecycle]: https://developer.android.com/guide/components/activities/activity-lifecycle
+    /// [`VkSurfaceKHR`]: https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkSurfaceKHR.html
+    /// [`wgpu::Surface`]: https://docs.rs/wgpu/latest/wgpu/struct.Surface.html
+    ///
+    /// ## iOS
+    ///
+    /// On iOS the `Suspended` event is currently emitted in response to an
+    /// [`applicationWillResignActive`] callback which means that the application is
+    /// about to transition from the active to inactive state (according to the
+    /// [iOS application lifecycle])
```
MarijnS95, line 115:
```suggestion
    /// [iOS application lifecycle]).
```

src/event.rs
```diff
@@ -74,9 +74,94 @@ pub enum Event<'a, T: 'static> {
...
 
     /// Emitted when the application has been resumed.
+    ///
+    /// For consistency all platforms emit a `Resumed` event even if they don't themselves have a
+    /// formal suspend/resume lifecycle. For systems without a standard suspend/resume lifecycle
+    /// the `Resumed` event is is always emitted after the `NewEvents(StartCause::Init)` event.
+    ///
+    /// # Portability
+    ///
+    /// It's recommended that applications should only initialized their graphics context and create
```
MarijnS95, line 129:
```suggestion
    /// It's recommended that applications should only initialize their graphics context and create
```

src/event.rs
```diff
@@ -74,9 +74,94 @@ pub enum Event<'a, T: 'static> {
...
+    /// a window after they have received their first `Resumed` event. Some systems
+    /// (specifically Android) won't allow applications to create a render surface until they are
+    /// resumed.
+    ///
+    /// In case of corner cases with how platform-specific events are mapped to Winit `Suspended`
```
MarijnS95, line 134: In case of corner cases?

src/event.rs
```diff
@@ -74,9 +74,94 @@ pub enum Event<'a, T: 'static> {
...
+    ///
+    /// [Activity lifecycle]: https://developer.android.com/guide/components/activities/activity-lifecycle
+    /// [`VkSurfaceKHR`]: https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkSurfaceKHR.html
+    /// [`wgpu::Surface`]: https://docs.rs/wgpu/latest/wgpu/struct.Surface.html
+    ///
+    /// ## iOS
+    ///
+    /// On iOS the `Resumed` event is emitted in response to an [`applicationDidBecomeActive`]
+    /// callback which means the application is "active" (according to the
+    /// [iOS application lifecycle])
```
MarijnS95, line 161:
```suggestion
    /// [iOS application lifecycle]).
```

src/event.rs
```diff
@@ -74,9 +74,94 @@ pub enum Event<'a, T: 'static> {
...
+    /// and `Resumed` events it's recommended that applications be able to gracefully handle redundant
+    /// (i.e. back to back) `Suspended` or `Resumed` events.
+    ///
+    /// Also see [`Suspended`][Self::Suspended] notes.
+    ///
+    /// ## Android
+    ///
+    /// On Android specifically, the `Suspended` event relates to two things:
+    /// 1. The application has resumed (according to the [Activity lifecycle])
+    /// 2. And the application has an associated SurfaceView (which is expected to closely correlate
```
MarijnS95, line 144: Same comments about `SurfaceView`.

src/event.rs
```diff
@@ -74,9 +74,94 @@ pub enum Event<'a, T: 'static> {
...
+    /// with being resumed)
+    ///
+    /// Applications that need to run on Android must wait until they have been `Resumed`
+    /// before they will be able to create a render surface (such as an `EGLSurface`,
+    /// [`VkSurfaceKHR`] or [`wgpu::Surface`]) which depend on having a
+    /// SurfaceView. Applications must also assume that if they are `Suspended` then their
+    /// render surfaces are invalid and should be dropped. (Also see [`Suspended`][Self::Suspended] notes)
```
MarijnS95, line 151:
```suggestion
    /// render surfaces are invalid and should be dropped.
    ///
    /// Also see [`Suspended`][Self::Suspended] notes.
```

To match what you wrote there.

MarijnS95 (changes requested):
Just style nits now :)

src/event.rs
rib: It came up in the discussion of https://github.com/rust-windowing/winit/issues/2185 - so it hopefully helps clarify what applications can (or can not rely on).

Some specific, plausible corner cases were also discussed there, whereby I feel like it would be wise for Winit to not over-promise that there will never be unforseen corner cases (considering that it seems like it shouldn't be a significant burden for applications to be prepared to handle redundant events).

The alternative imho would be to explicitly promise that Winit will never emit redundant events (otherwise the documentation isn't helping clarify what applications can rely on).

In practice I'd expect Winit to strive to avoid any redundant events, but by under promising I think there's a better chance that apps will cope fine if they do hit some kind of corner case.

src/event.rs
rib: The updated docs now consistently link to the other event (i.e. the `Suspended` docs have links to `Resumed` and the `Resumed` docs link back to `Suspended`), and neither links back to itself, it just uses back ticks.

src/event.rs
rib: I ended up taking another pass at rewording this:

```rust
    /// Considering that the implementation of `Suspended` and [`Resumed`] events may be internally
    /// driven by multiple platform-specific events, and there may be subtle differences across
    /// platforms with how these internal events are delivered, it's recommended that applications
    /// be able to gracefully handle redundant (i.e. back-to-back) `Suspended` or [`Resumed`] events.
```

rib:
Okey, my latest doc updates hopefully fix the last nits that were raised 🤞

rib:
NB: it's still expected that there will be some evolution of these `Suspended` / `Resumed` docs.

E.g. my thinking had been to follow up with a change to the iOS backend to only suspend when the application has become inactive (not before it becomes inactive) at which point we will have simpler, more consistent semantics for the `Suspended` state across iOS and Android. At this point we can document that "`Suspended`" means that the application's rendering has been suspended by the OS.

Right now the winit docs just say `Suspended` is "Emitted when the application has been suspended" which isn't very helpful for understanding what being "suspended" actually means - and that's hard to clarify right now with the subtle inconsistencies between Android + iOS.

maroider:
Just checking in since we're working on getting the next release out the door (#2322): is this PR just waiting on approving reviews?

rib:
I think so.

From @MarijnS95:
> this PR is good as it is and better describes the status-quo + handles it more conveniently across platforms,

and:
> Just style nits now :)

@kchibisov also approved the PR, after discussion:
> I'll be fine with resumed event for now then.


In summary I think there's more to do (via separate issues/PRs) with making Suspend/Resumed events consistent between iOS and Android and hopefully the docs will be further updated along with those future changes but for now this should be an incremental improvement for cross platform consistency (between mobile and desktop) and a general improvement in terms of documentation.

maroider (approved):

src/event.rs
```diff
@@ -74,9 +74,105 @@ pub enum Event<'a, T: 'static> {
...
+    /// ## Android
+    ///
+    /// On Android specifically, the `Suspended` event relates to two things:
+    /// 1. The application is paused (according to the [Activity lifecycle])
+    /// 2. Or the application has had its associated [`SurfaceView`] destroyed (which is expected
+    /// to closely correlate with being paused)
+    ///
+    /// Applications that need to run on Android should assume their [`SurfaceView`] has been
+    /// destroyed, which indirectly invalidates any existing render surfaces that may have been
+    /// created outside of Winit (such as an `EGLSurface`, [`VkSurfaceKHR`] or [`wgpu::Surface`]).
```
MarijnS95, line 102: As discussed before I'm not fond of this wording: you first state that it could be _either_ destroyed or just paused, then continue stating that users should assume it has been destroyed. Is it perhaps better to state that it has been destroyed first, followed by mentioning that "this is likely/typically because the application has been paused (according to the [Activity lifecycle])"?

src/event.rs
```diff
@@ -74,9 +74,105 @@ pub enum Event<'a, T: 'static> {
...
+    /// [iOS application lifecycle]: https://developer.apple.com/documentation/uikit/app_and_environment/managing_your_app_s_life_cycle
+    ///
+    /// [`Resumed`]: Self::Resumed
     Suspended,
 
     /// Emitted when the application has been resumed.
+    ///
+    /// For consistency all platforms emit a `Resumed` event even if they don't themselves have a
+    /// formal suspend/resume lifecycle. For systems without a standard suspend/resume lifecycle
+    /// the `Resumed` event is is always emitted after the [`NewEvents(StartCause::Init)`][StartCause::Init]
```
MarijnS95, line 129:
```suggestion
    /// the `Resumed` event is always emitted after the [`NewEvents(StartCause::Init)`][StartCause::Init]
```

src/event.rs
```diff
@@ -74,9 +74,105 @@ pub enum Event<'a, T: 'static> {
...
+    /// platforms with how these internal events are delivered, it's recommended that applications
+    /// be able to gracefully handle redundant (i.e. back-to-back) [`Suspended`] or `Resumed` events.
+    ///
+    /// Also see [`Suspended`] notes.
+    ///
+    /// ## Android
+    ///
+    /// On Android specifically, the [`Suspended`] event relates to two things:
+    /// 1. The application has resumed (according to the [Activity lifecycle])
+    /// 2. And the application has an associated [`SurfaceView`] (which is expected to closely correlate
```
MarijnS95, line 150: Perhaps my comment above on the similar docs for `Suspended` are simply resolved by having an `And` here as first word for `2.` (instead of `Or`), just like you have here.

src/event.rs
```diff
@@ -74,9 +74,105 @@ pub enum Event<'a, T: 'static> {
...
+    /// platforms with how these internal events are delivered, it's recommended that applications
+    /// be able to gracefully handle redundant (i.e. back-to-back) `Suspended` or [`Resumed`] events.
+    ///
+    /// Also see [`Resumed`] notes.
+    ///
+    /// ## Android
+    ///
+    /// On Android specifically, the `Suspended` event relates to two things:
+    /// 1. The application is paused (according to the [Activity lifecycle])
+    /// 2. Or the application has had its associated [`SurfaceView`] destroyed (which is expected
```
MarijnS95, line 97: So, this'd be:
```suggestion
    /// 2. And the application has had its associated [`SurfaceView`] destroyed (which is expected
```

MarijnS95 (approved):
I'm okay with this after fixing the `is is` typo, and bringing up the same lifecycle-paused vs surface destroyed question: which may be simply addressed by `s/Or/And` which has been done for `Resumed` but not `Suspended`?

src/event.rs
rib: The intent here is to let people know how the implementation currently works, which might be important if you're dealing with some Android-specific use case.

The use of "Or" is because that's how the actual implementation work right now, which isn't being changed. There are two things that may lead result in getting a "Suspended" event on Android, and you don't need both (so it's "Or" not "And").

The second part is effectively saying that developers should assume the worst. Since there are two possible causes for the app to be suspended and one of them technically doesn't invalidate their surface view then I think it's best to recommend that applications should behave as if their surface view has been destroyed.

I.e. the first part says exactly how the implementation works; the second part recommends that you always treat it like your surface view was destroyed.

Personally I think the information makes sense and is consistent with itself, so I'm not quite sure how to improve the wording.

Your suggestion seems to effectively remove the information about how it works internally which seems like it could be useful to know for some Android specific use cases.

If I change it to "state that it has been destroyed first" that's not always true, but I do recommend that you act as if it's always true.

src/event.rs
rib: The choice of "Or" above and "And" was _very_ _very_ intentional :-) , those are the words that reflect the actual logic in each case.

I.e.:
- when being suspended you _might_ have had your surface view destroyed (one of two possible reasons to suspend - logical OR).
- when resuming you _definitely_ do have a surface view (there are two factors that must _both_ be satisfied before we resume - logical AND)

rib:
I just pushed an update that resolves the changelog conflict and fixes an "is is" typo that @MarijnS95 pointed out.

I haven't changed the wording about the implementation details that @MarijnS95 has recently commented on while I think the current wording is a correct reflection of the android-specific details.

There might be some better wording we could come up with to clarify further but I'd like to keep the information itself which the last suggested changes wouldn't do imho.

I also figure that this kind of re-wording can happen after the change lands. Even though this does try to improve the state of the docs for the Suspended/Resumed events, I think the technical consistency change is more important.

src/event.rs
rib: Though actually one glaring issue here is that the first line is a copy and paste from the `Suspended` event docs. This should  start with "On Android specifically, the **[`Resumed`]** event relates to two things"

rib: okey, I've tried adding another line to really try and clarify the meaning here as well as fix the copy and paste error.

```diff
--- a/src/event.rs
+++ b/src/event.rs
@@ -97,6 +97,9 @@ pub enum Event<'a, T: 'static> {
     /// 2. Or the application has had its associated [`SurfaceView`] destroyed (which is expected
     /// to closely correlate with being paused)
     ///
+    /// _(I.e. **either** of these may lead to a `Suspended` event, but applications should act
+    /// as if both have happened.)_
+    ///
     /// Applications that need to run on Android should assume their [`SurfaceView`] has been
     /// destroyed, which indirectly invalidates any existing render surfaces that may have been
     /// created outside of Winit (such as an `EGLSurface`, [`VkSurfaceKHR`] or [`wgpu::Surface`]).
@@ -145,11 +148,13 @@ pub enum Event<'a, T: 'static> {
     ///
     /// ## Android
     ///
-    /// On Android specifically, the [`Suspended`] event relates to two things:
+    /// On Android specifically, the `Resumed` event relates to two things:
     /// 1. The application has resumed (according to the [Activity lifecycle])
     /// 2. And the application has an associated [`SurfaceView`] (which is expected to closely correlate
     /// with being resumed)
     ///
+    /// _(I.e. **both** of these must happen before the application gets a `Resumed` event.)_
+    ///
     /// Applications that need to run on Android must wait until they have been `Resumed`
     /// before they will be able to create a render surface (such as an `EGLSurface`,
     /// [`VkSurfaceKHR`] or [`wgpu::Surface`]) which depend on having a
```

rib:
last push tries to add a couple of lines to further clarify the android-specific logic for suspended/resumed events and fixes a copy and paste error that was noticed from @MarijnS95's comments:

```diff
--- a/src/event.rs
+++ b/src/event.rs
@@ -97,6 +97,9 @@ pub enum Event<'a, T: 'static> {
     /// 2. Or the application has had its associated [`SurfaceView`] destroyed (which is expected
     /// to closely correlate with being paused)
     ///
+    /// _(I.e. **either** of these may lead to a `Suspended` event, but applications should act
+    /// as if both have happened.)_
+    ///
     /// Applications that need to run on Android should assume their [`SurfaceView`] has been
     /// destroyed, which indirectly invalidates any existing render surfaces that may have been
     /// created outside of Winit (such as an `EGLSurface`, [`VkSurfaceKHR`] or [`wgpu::Surface`]).
@@ -145,11 +148,13 @@ pub enum Event<'a, T: 'static> {
     ///
     /// ## Android
     ///
-    /// On Android specifically, the [`Suspended`] event relates to two things:
+    /// On Android specifically, the `Resumed` event relates to two things:
     /// 1. The application has resumed (according to the [Activity lifecycle])
     /// 2. And the application has an associated [`SurfaceView`] (which is expected to closely correlate
     /// with being resumed)
     ///
+    /// _(I.e. **both** of these must happen before the application gets a `Resumed` event.)_
+    ///
     /// Applications that need to run on Android must wait until they have been `Resumed`
     /// before they will be able to create a render surface (such as an `EGLSurface`,
     /// [`VkSurfaceKHR`] or [`wgpu::Surface`]) which depend on having a
```

src/event.rs
MarijnS95: > The use of "Or" is because that's how the actual implementation work right now, which isn't being changed

It feels like we've discussed this before, but I've forgot the outcome, so I'll ask again [^1]:

[^1]: and to restate: long walls of text repeating the same thing over and over make this only harder to discuss.

Where exactly is Android's `onPause` callback - resulting in `Event:Pause` - converted into `Event::Suspended` on the Winit side? All I can see - when checked out on this PR - is solely `Event::WindowDestroyed` resulting in `Event::Suspended`. Furthermore, digging down for calls to that callback show only `onSurfaceChanged`/`onSurfaceDestroyed` calls on `android::view::Surface`.

https://cs.android.com/android/platform/superproject/+/master:frameworks/native/include/android/native_activity.h;l=204;bpv=1;bpt=1?q=onNativeWindowDestroyed

rib: Ah, damn, I think I'm recalling how my android-activity branch works, not master, urgh sorry :/

I think the way I described it is how it _should_ be implemented but right now that's not technically the case.

I guess I'll just drop the Android docs for now and instead this info can be added if the implementation is updated (fixed imho).

rib: okey, I've updated the technical descriptions of how the events are implemented currently to say:

> On Android, the `Suspended` event is only sent when the application's associated
> [`SurfaceView`] is destroyed. This is expected to closely correlate with the [`onPause`]
> lifecycle event but there may technically be a discrepancy.

and:
>  On Android the `Resumed` event is sent when a new [`SurfaceView`] has been created. This is
> expected to closely correlate with the [`onResume`] lifecycle event but there may technically
> be a discrepancy.

rib:
After realising that I was actually documenting the Android behaviour of the `Suspend`/`Resumed` events I have on a branch and not on master I've tweaked the technical descriptions for Android again - sorry for the churn.

src/event.rs
```diff
@@ -74,9 +74,107 @@ pub enum Event<'a, T: 'static> {
     UserEvent(T),
 
     /// Emitted when the application has been suspended.
+    ///
+    /// # Portability
+    ///
+    /// Not all platforms support the notion of suspending applications, and there may be no
+    /// technical way to guarantee being able to emit a `Suspended` event if the OS has
+    /// no formal application lifecycle (currently only Android and iOS do). For this reason
```
maroider, line 82:
```suggestion
    /// no formal application lifecycle (currently only Android and iOS do). For this reason,
```

src/event.rs
```diff
@@ -74,9 +74,107 @@ pub enum Event<'a, T: 'static> {
...
+    /// Winit does not currently try to emit pseudo `Suspended` events before the application
+    /// quits on platforms without an application lifecycle.
+    ///
+    /// Considering that the implementation of `Suspended` and [`Resumed`] events may be internally
+    /// driven by multiple platform-specific events, and there may be subtle differences across
```
maroider, line 87:
```suggestion
    /// driven by multiple platform-specific events, and that there may be subtle differences across
```

src/event.rs
```diff
@@ -74,9 +74,107 @@ pub enum Event<'a, T: 'static> {
...
+    /// the event callback completes, which may be re-created when the application is next [`Resumed`].
+    ///
+    /// [`SurfaceView`]: https://developer.android.com/reference/android/view/SurfaceView
+    /// [Activity lifecycle]: https://developer.android.com/guide/components/activities/activity-lifecycle
+    /// [`VkSurfaceKHR`]: https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkSurfaceKHR.html
+    /// [`wgpu::Surface`]: https://docs.rs/wgpu/latest/wgpu/struct.Surface.html
+    ///
+    /// ## iOS
+    ///
+    /// On iOS the `Suspended` event is currently emitted in response to an
```
maroider, line 115:
```suggestion
    /// On iOS, the `Suspended` event is currently emitted in response to an
```

src/event.rs
```diff
@@ -74,9 +74,107 @@ pub enum Event<'a, T: 'static> {
...
+    ///
+    /// [`applicationWillResignActive`]: https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1622950-applicationwillresignactive
+    /// [iOS application lifecycle]: https://developer.apple.com/documentation/uikit/app_and_environment/managing_your_app_s_life_cycle
+    ///
+    /// [`Resumed`]: Self::Resumed
     Suspended,
 
     /// Emitted when the application has been resumed.
+    ///
+    /// For consistency all platforms emit a `Resumed` event even if they don't themselves have a
```
maroider, line 128:
```suggestion
    /// For consistency, all platforms emit a `Resumed` event even if they don't themselves have a
```

src/event.rs
```diff
@@ -74,9 +74,107 @@ pub enum Event<'a, T: 'static> {
...
+    ///
+    /// # Portability
+    ///
+    /// It's recommended that applications should only initialize their graphics context and create
+    /// a window after they have received their first `Resumed` event. Some systems
+    /// (specifically Android) won't allow applications to create a render surface until they are
+    /// resumed.
+    ///
+    /// Considering that the implementation of [`Suspended`] and `Resumed` events may be internally
+    /// driven by multiple platform-specific events, and there may be subtle differences across
```
maroider, line 141:
```suggestion
    /// driven by multiple platform-specific events, and that there may be subtle differences across
```

src/event.rs
```diff
@@ -74,9 +74,107 @@ pub enum Event<'a, T: 'static> {
...
+    /// platforms with how these internal events are delivered, it's recommended that applications
+    /// be able to gracefully handle redundant (i.e. back-to-back) [`Suspended`] or `Resumed` events.
+    ///
+    /// Also see [`Suspended`] notes.
+    ///
+    /// ## Android
+    ///
+    /// On Android the `Resumed` event is sent when a new [`SurfaceView`] has been created. This is
```
maroider, line 149:
```suggestion
    /// On Android, the `Resumed` event is sent when a new [`SurfaceView`] has been created. This is
```

src/event.rs
```diff
@@ -74,9 +74,107 @@ pub enum Event<'a, T: 'static> {
...
+    /// expected to closely correlate with the [`onResume`] lifecycle event but there may technically
+    /// be a discrepancy.
+    ///
+    /// [`onResume`]: https://developer.android.com/reference/android/app/Activity#onResume()
+    ///
+    /// Applications that need to run on Android must wait until they have been `Resumed`
+    /// before they will be able to create a render surface (such as an `EGLSurface`,
+    /// [`VkSurfaceKHR`] or [`wgpu::Surface`]) which depend on having a
+    /// [`SurfaceView`]. Applications must also assume that if they are [`Suspended`] then their
```
maroider, line 158:
```suggestion
    /// [`SurfaceView`]. Applications must also assume that if they are [`Suspended`], then their
```

src/event.rs
```diff
@@ -74,9 +74,107 @@ pub enum Event<'a, T: 'static> {
...
+    /// Also see [`Suspended`] notes.
+    ///
+    /// [`SurfaceView`]: https://developer.android.com/reference/android/view/SurfaceView
+    /// [Activity lifecycle]: https://developer.android.com/guide/components/activities/activity-lifecycle
+    /// [`VkSurfaceKHR`]: https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkSurfaceKHR.html
+    /// [`wgpu::Surface`]: https://docs.rs/wgpu/latest/wgpu/struct.Surface.html
+    ///
+    /// ## iOS
+    ///
+    /// On iOS the `Resumed` event is emitted in response to an [`applicationDidBecomeActive`]
```
maroider, line 170:
```suggestion
    /// On iOS, the `Resumed` event is emitted in response to an [`applicationDidBecomeActive`]
```

maroider (commented):
Just some nits

kchibisov (approved):

maroider (approved):

src/event.rs
MarijnS95: @rib Thanks, that's what I was trying to get across all along. The documentation in its current form is exactly what I had been looking for, good job on that.  I completely agree with ``This is expected to closely correlate with the [`onPause`] lifecycle event but there may technically be a discrepancy.`` as well.

---

Whether or not it is a good idea to generalize these in the future is something we should continue to discuss in https://github.com/rust-windowing/winit/issues/2337. I have some extra feelings on the `OR` and `AND` case for `Suspended` and `Resumed` that I'll share with you there.

PR #2336, opened 2022-06-13
Fix hiding a maximized window On Windows

### LINKED ISSUE #2233: On Windows, hiding a maximized window, unmaximizes instead
### Steps to reproduce:
1. `cargo run --example window_debug`
2. press <kbd>x</kbd> to maximize the window
3. press <kbd>v</kbd> to hide it

### Bug:
The window isn't hidden and isn't responsive. The window actually restored its original unmaximized size instead. The window icon is removed from the taskbar.

### Expected:
The window is hidden, the window icon is removed from the taskbar and the window doesn't restore to its original unmaximized size.

amrbashir:
Fixes #2233

I couldn't find a better way to fix this so I left a comment.


amrbashir:
maybe a better way to go about this would be to introduce a new method `WindowFlags::diff_mask` and add a comment that it should only be called on the diff between `self` and `new`?

msiglreith:
Thanks but I'm not totally happy with introducing more conditionals.
The issue seems to be that the second `ShowWindow(SW_MAXIMIZE)` call shadows the `SW_HIDE` from earlier due to implicit window activation from `SW_MAXIMIZE`. I currently would prefer to:
- Removing the `VISIBILE` masking in `WindowState::mask`
- Splitting the `ShowWindow` to only the `VISIBLE` case on start of the function and move the `!VISIBLE` case down to the end of the function.

amrbashir:
Thank you for the suggestions, it is much better to handle it that way.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- On Windows, Fix hiding a maximized window.
```
filnet, line 11: `Fix` should not be capitalized.

msiglreith (approved):

PR #2339, opened 2022-06-15
macOS disallow_highdpi will set explicity the value to avoid the SO value by default

Nazariglez:
Hi! I just made a simple change to how the `disallow_highdpi`option behaves. 

The default value of `setWantsBestResolutionOpenGLSurface_` was `false` until macos 10.14 and `true` after 10.5, we should set it to `YES` o `NO` to avoid always the default system value in favour of the user's code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - Implemented `Eq` for `Fullscreen`, `Theme`, and `UserAttentionType`.
 - **Breaking:** `Window::set_cursor_grab` now accepts `CursorGrabMode` to control grabbing behavior.
 - On Wayland, add support for `Window::set_cursor_position`.
+- Fix on macOS `WindowBuilder::with_disallow_hidpi`, setting true o false by the user no matter the SO default value. 
 
 # 0.26.1 (2022-01-05)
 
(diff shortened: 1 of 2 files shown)


src/platform_impl/macos/window.rs
```diff
@@ -100,8 +100,13 @@ unsafe fn create_view(
 ) -> Option<(IdRef, Weak<Mutex<CursorState>>)> {
     let (ns_view, cursor_state) = new_view(ns_window);
     ns_view.non_nil().map(|ns_view| {
+        // The default value of `setWantsBestResolutionOpenGLSurface_` was `false` until
+        // macos 10.14 and `true` after 10.5, we should set it to `YES` o `NO` to avoid
+        // always the default system value in favour of the user's code
```
madsmtm, line 105: Nit:

```suggestion
        // The default value of `setWantsBestResolutionOpenGLSurface:` was `false` until
        // macos 10.14 and `true` after 10.15, we should set it to `YES` or `NO` to avoid
        // always the default system value in favour of the user's code
```

madsmtm (commented):
Yeah, seems like a good idea, could you add an entry to the changelog for this?

Nazariglez:
Done!

CHANGELOG.md
```diff
@@ -60,6 +60,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - Implemented `Eq` for `Fullscreen`, `Theme`, and `UserAttentionType`.
 - **Breaking:** `Window::set_cursor_grab` now accepts `CursorGrabMode` to control grabbing behavior.
 - On Wayland, add support for `Window::set_cursor_position`.
+- Fix on macOS `WindowBuilder::with_disallow_hidpi`, setting true o false by the user no matter the SO default value.
```
madsmtm, line 63: Nit: Whoops, again

```suggestion
- Fix on macOS `WindowBuilder::with_disallow_hidpi`, setting true or false by the user no matter the SO default value. 
```

madsmtm (approved):

CHANGELOG.md
Nazariglez: lol, thx!

madsmtm:
Thanks!

PR #2343, opened 2022-06-20
Android: avoid deadlocks while handling UserEvent

Rodrigodd:
Calling `EventLoopProxy::send_event` while handling a UserEvent event would cause a deadlock.

edit: restored original PR template

### DIFF
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -2,7 +2,7 @@
 
 use std::{
     collections::VecDeque,
-    sync::{Arc, Mutex, RwLock},
+    sync::{mpsc, RwLock},
     time::{Duration, Instant},
 };
 
@@ -234,7 +234,8 @@ fn poll(poll: Poll) -> Option<EventSource> {
 
 pub struct EventLoop<T: 'static> {
     window_target: event_loop::EventLoopWindowTarget<T>,
-    user_queue: Arc<Mutex<VecDeque<T>>>,
+    user_events_sender: mpsc::Sender<T>,
+    user_events_receiver: mpsc::Receiver<T>,
     first_event: Option<EventSource>,
     start_cause: event::StartCause,
     looper: ThreadLooper,
@@ -256,6 +257,7 @@ macro_rules! call_event_handler {
 
 impl<T: 'static> EventLoop<T> {
     pub(crate) fn new(_: &PlatformSpecificEventLoopAttributes) -> Self {
+        let (sender, receiver) = mpsc::channel();
         Self {
             window_target: event_loop::EventLoopWindowTarget {
                 p: EventLoopWindowTarget {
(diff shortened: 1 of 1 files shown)


maroider:
Maybe we want a channel instead of an `Arc<Mutex<VecDequeue>>`? It looks like all we're doing is pushing new events to the back and popping old events from the front.

Rodrigodd:
> Maybe we want a channel instead of an Arc<Mutex<VecDequeue>>? It looks like all we're doing is pushing new events to the back and popping old events from the front.

@maroider, yes, replacing it by a `mpsc` like the others backends made more sense.

src/platform_impl/android/mod.rs
```diff
@@ -256,14 +257,16 @@ macro_rules! call_event_handler {
...
         Self {
             window_target: event_loop::EventLoopWindowTarget {
                 p: EventLoopWindowTarget {
                     _marker: std::marker::PhantomData,
                 },
                 _marker: std::marker::PhantomData,
             },
-            user_queue: Default::default(),
+            user_events_sender: sender,
+            user_events_receiver: receiver,
```
maroider, line 269: I'd prefer field init shorthand here.

MarijnS95 (approved):
Looks great, haven't tested it though!

MarijnS95:
Tested now on Android, including recursive `send_event()` calls that lock up without this PR, as stated.

rib:
Just for cross-reference I figured I'd note that I also ended up switching to using an mpsc channel for user events in my updated Android backend based on `android-activity`: https://github.com/rib/winit/tree/android-activity

One notable difference was that I also used the `PeekableReceiver` from the X11 backend that enables the backend to peek and check for any pending user events before blocking and preemtively set a zero timeout in that case.

To some extent that should be redundant since we always have to issue a looper wake() when sending a user event which can be relied upon to trigger a future wakeup if that's issued while the looper is not blocking. On the other hand, something else that was noted in the X11 backend was a tendency to get _excessive_ wake ups which are mitigated by explicitly checking for user events and redraw requests after a wake up - so that it can avoid running the event loop if there's not really anything to do. That's also enabled by using the `PeekableReceiver`.

MarijnS95:
I found an old branch toying around with this too (as I wasn't at all content with the way that PR implemented `request_redraw()` back then). Linking it for reference, just pretty sure that things may be dated/wrong/incomplete otherwise it would have ended up in a PR :grimacing:

https://github.com/MarijnS95/winit/compare/android-internal-event-pipe

MarijnS95:
@maroider I tested it, and I think this is okay - it's way better than the status-quo at least.

Would've personally used a pipe with file descriptor since that better fits the looper structure (which has to be `.wake()`d manually here), but it may also be slightly less performant? I don't really have a preference, we're _planning to™_ replace most of the Android backend anyway.

msiglreith (approved):

maroider:
Are we merging this or not? It looks like we have no blockers...

Issue #2166, opened 2022-01-29
Running a Second EventLoop Creates a Segfault


Chloe-Woahie:
I created an EventLoop, WinitInputHelper, EventLoop, and Window just fine. The event loop runs as well.

```rust
let event_loop = EventLoop::new();
let input = WinitInputHelper::new();
let window = {
    let size = LogicalSize::new(self.width as f64, self.height as f64);
    WindowBuilder::new()
        .with_title("rendering-demo-rs")
        .with_inner_size(size)
        .with_min_inner_size(size)
        .build(&event_loop)
        .unwrap()
};
let pixels = {
    let window_size = window.inner_size();
    let surface_texture =
        SurfaceTexture::new(window_size.width, window_size.height, &window);
    Pixels::new(self.width, self.height, surface_texture).unwrap()
};

event_loop.run(move |event, _, control_flow| {});
```

However, later in the code, I created another EventLoop forgetting that I didn't have a WIndow attached to it. Using event_loop.run() causes a segmentation fault with no other error. It segfaults when I run either event loop.
```rust
let event_loop = EventLoop::new();
self.event_loop.run(move |event, _, control_flow| {});
```
```rust
let event_loop = EventLoop::new();
event_loop.run(move |event, _, control_flow| {});
```
```bash
   Compiling draw_triangles v0.1.0 (/home/fekwek/Git/rust-pixel-buffer/examples/draw_triangles)
    Finished dev [unoptimized + debuginfo] target(s) in 4.26s
     Running `target/debug/draw_triangles`
"Rust Pixel Buffer Example"
Segmentation fault (core dumped)
```

It also segfaults when I create another event loop and attach a window to it.
```rust
let event_loop = EventLoop::new();
        let input = WinitInputHelper::new();
        let window = {
            let size = LogicalSize::new(self.width as f64, self.height as f64);
            WindowBuilder::new()
                .with_title("rendering-demo-rs")
                .with_inner_size(size)
                .with_min_inner_size(size)
                .build(&event_loop)
                .unwrap()
        };
        let pixels = {
            let window_size = window.inner_size();
            let surface_texture =
                SurfaceTexture::new(window_size.width, window_size.height, &window);
            Pixels::new(self.width, self.height, surface_texture).unwrap()
        };
        event_loop.run(move |event, _, control_flow| {});
```

(This is my first pull request so the quality may be bad)

madsmtm:
[...]

PR #2344, opened 2022-06-21
Panic on multiple EventLoop instantiation, document why

MarcusGrass:
- [ ] Tested on all platforms changed *(Only tested on x11, although this change is platform agnostic)*

Since it was extremely easy to implement using existing dependencies I just made a prototype implementation that agressively fixes https://github.com/rust-windowing/winit/issues/2166. 

In the issue discussions were around just documenting it, or making it panic. I'm more for making it panic immediately since the problems can be hard to diagnose if they occur "naturally" such as in the report of a segfault, with other example in that same issue of SendErrors. Getting a segfault and tracking it down to EventLoopBuilder's docs will probably waste a lot of time for users.

I'm not ecstatic about having a static OnceCell to guard against double instantiation, but I can't think of any smoother solution

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - **Breaking:** `Window::set_cursor_grab` now accepts `CursorGrabMode` to control grabbing behavior.
 - On Wayland, add support for `Window::set_cursor_position`.
 - Fix on macOS `WindowBuilder::with_disallow_hidpi`, setting true or false by the user no matter the SO default value. 
+- Panics if EventLoopBuilder is built twice in the same runtime and documented that that action it's unsupported
 
 # 0.26.1 (2022-01-05)
 
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -11,6 +11,7 @@ use instant::Instant;
 use std::marker::PhantomData;
 use std::ops::Deref;
 use std::{error, fmt};
+use once_cell::sync::OnceCell;
 
 use crate::{event::Event, monitor::MonitorHandle, platform_impl};
 
@@ -55,6 +56,8 @@ pub struct EventLoopBuilder<T: 'static> {
     _p: PhantomData<T>,
 }
 
+static EVENT_LOOP_CREATED: OnceCell<()> = OnceCell::new();
+
 impl EventLoopBuilder<()> {
     /// Start building a new event loop.
     #[inline]
@@ -76,8 +79,11 @@ impl<T> EventLoopBuilder<T> {
 
     /// Builds a new event loop.
     ///
-    /// ***For cross-platform compatibility, the [`EventLoop`] must be created on the main thread.***
-    /// Attempting to create the event loop on a different thread will panic. This restriction isn't
+    /// ***For cross-platform compatibility, the [`EventLoop`] must be created on the main thread,
+    /// and only once per application.***
+    ///
+    /// Attempting to create the event loop on a different thread, or multiple event loops in
+    /// the same application, will panic. This restriction isn't
     /// strictly necessary on all platforms, but is imposed to eliminate any nasty surprises when
     /// porting to platforms that require it. `EventLoopBuilderExt::any_thread` functions are exposed
     /// in the relevant [`platform`] module if the target platform supports creating an event loop on
@@ -95,6 +101,9 @@ impl<T> EventLoopBuilder<T> {
     /// [`platform`]: crate::platform
     #[inline]
     pub fn build(&mut self) -> EventLoop<T> {
+        if EVENT_LOOP_CREATED.set(()).is_err() {
+            panic!("Attempted to build an EventLoop twice in the same application which is unsupported.")
+        }
         // Certain platforms accept a mutable reference in their API.
         #[allow(clippy::unnecessary_mut_passed)]
         EventLoop {


src/event_loop.rs
```diff
@@ -55,6 +56,8 @@ pub struct EventLoopBuilder<T: 'static> {
     _p: PhantomData<T>,
 }
 
+static EVENT_LOOP_CREATED: OnceCell<()> = OnceCell::new();
```
kchibisov, line 59: Could you move it inside of the `build` function?

CHANGELOG.md
```diff
@@ -61,6 +61,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - **Breaking:** `Window::set_cursor_grab` now accepts `CursorGrabMode` to control grabbing behavior.
 - On Wayland, add support for `Window::set_cursor_position`.
 - Fix on macOS `WindowBuilder::with_disallow_hidpi`, setting true or false by the user no matter the SO default value. 
+- Panics if EventLoopBuilder is built twice in the same runtime and documented that that action is unsupported
```
kchibisov, line 64:
```suggestion
- `EventLoopBuilder::build` will now panic when the `EventLoop` is being created more than once.
```

kchibisov (changes requested):

maroider (approved):
I don't have much of anything to point out beyond what @kchibisov has already done. I *would* like a better panic message, but I can't seem to formulate something that's both unambiguously better and doesn't remove the "this isn't supported" part right now.

CHANGELOG.md
```diff
@@ -61,6 +61,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - **Breaking:** `Window::set_cursor_grab` now accepts `CursorGrabMode` to control grabbing behavior.
 - On Wayland, add support for `Window::set_cursor_position`.
 - Fix on macOS `WindowBuilder::with_disallow_hidpi`, setting true or false by the user no matter the SO default value. 
+- EventLoopBuilder::build` will now panic when the `EventLoop` is being created more than once.
```
kchibisov, line 64:
```suggestion
- EventLoopBuilder::build` will now panic when the `EventLoop` is being created more than once.
```
```suggestion
- `EventLoopBuilder::build` will now panic when the `EventLoop` is being created more than once.
```

src/event_loop.rs
```diff
@@ -8,6 +8,7 @@
 //! See the root-level documentation for information on how to create and use an event loop to
 //! handle events.
 use instant::Instant;
+use once_cell::sync::OnceCell;
```
kchibisov, line 11: Could you move `instant` and `once_cell` one line below `std`, like
```rust
std::

instant
once_cell
```

src/event_loop.rs
```diff
@@ -95,6 +99,10 @@ impl<T> EventLoopBuilder<T> {
     /// [`platform`]: crate::platform
     #[inline]
     pub fn build(&mut self) -> EventLoop<T> {
+        static EVENT_LOOP_CREATED: OnceCell<()> = OnceCell::new();
+        if EVENT_LOOP_CREATED.set(()).is_err() {
+            panic!("Attempted to build an EventLoop twice in the same application which is unsupported.")
```
kchibisov, line 104:
```suggestion
            panic!("Attempted to build an EventLoop twice in the same application which is unsupported.")
```
```suggestion
            panic!("Creating EventLoop multiple times is not supported.")
```

kchibisov (changes requested):

kchibisov (approved):

MarijnS95:
Just in case anyone comes across this error on Android:

Its framework can launch multiple activities inside the same process. This activity start however pretends to be like a `main()` call that is typically only called once, nor does `ndk-glue` properly support >1 activity running at a given time (everything goes bonkers otherwise). If users create their `EventLoop` inside that fake `main()` function, they'll run into this error at some point.

That's not something caused by this PR though, rather it's highlighting the underlying issues that we have yet to resolve ;)

kchibisov:
We could have platform specific methods to avoid this error, like `new_any_tread` methods available on certain platforms, like Wayland/X11. Something like that could be done for Android.

MarijnS95:
Please no. This is "yet another `ndk-glue` misunderstanding" that has to be corrected rather than swept under the rug. Its design doesn't allow for a use-case that's clearly possible.

When does this happen? Depends on how Android launches the users' activity; on my phone it _at least **relaunches**_ when switching to "pop-up", allowing me to test  resizes (and it took a while to understand why it was completely and utterly broken :grimacing:).

grovesNL:
Should we clear the static on `Event::LoopDestroyed`? There are some potential use cases where it might make sense to recreate the event loop (e.g. wasm module where the process fully exits but the `OnceCell` remains set) or have multiple event loops (e.g. when not using `run` as described in #1714).

kchibisov:
I think some platforms(macOS/ios) will abort if you try to recreate the event loop, the error is for a particular functions to prevent cross platforms issues, however we can provide separate methods that allow event loop recreation.

grovesNL:
@kchibisov Platform-specific extensions would be ok. I just thought it might be cleaner to clear the `OnceCell` in one place if multiple platforms support event loop recreation.

With the current API, is it possible to attempt to recreate the event loop on macOS and iOS without having to go through one of the non-portable paths (e.g. `run_return`)? If we have to use `run_return` in order to recreate the event loop on macOS/iOS anyway, maybe we could document the restriction there.

kchibisov:
`run_return` isn't about recreating the loop, it's not about loop dropping like you do `run_return` and then `std::mem::drop(event_loop)`. Run return basically means that you can call `run_return`, poll events you want into some buffer, exit closure, processes events, render, and call `run_return` again. It's more about polling events on demand, not dropping the event loop.

EventLoop creation and dropping is complicated topic wrt recreation since system libraries to initialize some global state most of the time. On drop it usually persist or something else has non-null state, etc, etc.

So I'm not sure what needs to be documented, in fact I'm not sure there're platforms safely allowing event loop recreation, most of the time it's never tested and if it works you're lucky.

grovesNL:
Yeah I agree with that. I mean you can't normally return from `run`, so in general it's not clear if it would be possible to recreate the event loop after its running without going through a non-portable path (e.g. `run_return`). I was wondering if there's any downside to clearing the `OnceCell` on `Event::LoopDestroyed` for all platforms since the event loop can't usually be recreated when using `run`.

On the web it can be a bit tricky avoiding recreating the event loop (e.g. single page application that uses winit on a single tab). Even though we're using `run`, destroy the event loop, and exit the process, the static `OnceCell` remains set in the wasm module. Normally this works pretty well, but I had to revert this PR in my winit fork as a workaround to avoid the `OnceCell` already being set. It's probably possible to wait for GC to fully unload the wasm module or work around it in other ways.

kchibisov:
>for all platforms since the event loop can't usually be recreated when using run.

Well, for `run` it doesn't make any sense, since `run` terminates the program. If the platform supports multiple `EventLoop::new` I'd suggest to add a platform specific method on `EventLoopBuilder`, like `build_web`, which will avoid checking the variable in the first place.

grovesNL:
Right, #1714 / #2208 will definitely make this easier, but we'd still have the same issue if the `OnceCell` is shared between calls to `spawn`. In other words, if the `OnceCell` isn't cleared during `Event::LoopDestroyed` or when the caller exits, we still won't be able to recreate the event loop.

PR #2346, opened 2022-06-23
Make the `EventLoopWindowTarget` passed to the `EventLoop::run` callback live for `'static`

Liamolucko:
- [ ] ~~Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior~~ this is purely a signature change, so no extra docs are needed on top.

This PR changes the signature of `EventLoop::run` to make the `EventLoopWindowTarget` passed to the callback live for `'static`, and changes the signature of `EventLoop::run_return` to make the `EventLoopWindowTarget` live for as long as the reference to the event loop.

The main motivation for this is writing an async executor on top of winit (ref #1199). Such an executor needs to pass an `EventLoopWindowTarget` to the future to create windows with, which then needs to live for the entirety of the future (i.e., `'static`).

This plus #2294 should make it possible to write an async executor on top of winit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- The `EventLoopWindowTarget` passed to the `EventLoop::run` callback now has a lifetime of `'static`.
+- The `EventLoopWindowTarget` passed to the `EventLoop::run_return` callback now lives for as long as the reference to the event loop.
 - Build docs on `docs.rs` for iOS and Android as well.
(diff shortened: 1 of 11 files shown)


madsmtm (changes requested):
Thanks for the ping, but I'm gonna have to reject this change; we're actively trying to get the codebase in a spot where the `EventLoopWindowTarget` (now `ActiveEventLoop`) is specifically _not_ static, but rather an `&mut` reference to internal data, and which might change upon each iteration.

> The main motivation for this is writing an async executor on top of winit

I believe this may already have been done in [`async-winit`](https://docs.rs/async-winit/latest/async_winit/)?

In any case, the way to solve this would be via. message-passing and by waking the event loop, and letting it process things in there.

PR #2348, opened 2022-06-26
On macOS, fix confirmed character inserted

kchibisov:
When confirming input in e.g. Korean IME or using characters like
`+` winit was sending those twice, once via `Ime::Commit` and the
other one via `ReceivedCharacter`, since those events weren't generating
any `Ime::Preedit` and were forwarded due to `do_command_by_selector`.

### DIFF
--- a/src/platform_impl/macos/view.rs
+++ b/src/platform_impl/macos/view.rs
@@ -54,11 +54,19 @@ impl Default for CursorState {
     }
 }
 
-#[derive(Eq, PartialEq)]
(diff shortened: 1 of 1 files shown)


kchibisov:
Issue was mentioned downstream here https://github.com/alacritty/alacritty/pull/6079#discussion_r903724983 .

Basically winit is sending e.g. `Enter` which is being pressed to confirm `Ime` input. Such behavior is unexpected and proposed solution to simply eat the next character raised right after commit in a single `key_down` processing.

PR #2351, opened 2022-06-29
Implement From<u64> for WindowId and vise-versa

kchibisov:
This should help downstream applications to expose WindowId to the end
users via e.g. IPC to control particular windows in multi window
systems.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On Wayland, add support for `Window::set_cursor_position`.
 - Fix on macOS `WindowBuilder::with_disallow_hidpi`, setting true or false by the user no matter the SO default value. 
 - `EventLoopBuilder::build` will now panic when the `EventLoop` is being created more than once.
+- Added `From<u64>` for `WindowId` and `From<WindowId>` for `u64`.
 
 # 0.26.1 (2022-01-05)
 
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -627,6 +627,18 @@ impl WindowId {
     }
 }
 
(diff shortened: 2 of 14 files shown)


chrisduerr (approved):

maroider:
I'm not sure if casting to and from `u64` is entirely unproblematic here.

Windows' window handles are `usize`. 32-bit Windows installs are admittedly likely quite rare, and I don't believe Windows 11 has a 32-bit edition, but this still feels a bit iffy to me.

It also looks like X11's window IDs are always `u32`s, `X11/Xmd.h` defines `CARD32` as `unsigned int` on 64-bit systems, and `unsigned long` on 32-bit systems, although `x11-dl` has defined `CARD32` as `c_ulong`, which seems like it would be unsound (or at the very least incorrect) to me, as `c_ulong` appears to be 8 bytes long on 64-bit Linux.

kchibisov:
>Windows' window handles are usize. 32-bit Windows installs are admittedly likely quite rare, and I don't believe Windows 11 has a 32-bit edition, but this still feels a bit iffy to me.

That's why we use `u64`, so it can fit any possible value into it? The point is to have ability to convert `WindowId` to something, send it via some IPC mechanism, and then again receive it and create `WindowId` out of it to access something like `HashMap<WindowId, Window>`.

>I'm not sure if casting to and from u64 is entirely unproblematic here.

It should be, since you can't do anything with it given that `WindowId` is an opaque struct for which only checking for identity is stable, it's not like you can build anything out of it nor it's guaranteed to work if you cast it to anything.

chrisduerr:
Yeah at the very least you should be able to convert to `u64`. Converting from it might be more controversial but I don't think it should be that big of a deal either.

MarijnS95 (approved):
For Android, I guess this is okay. Hopefully we can support multiple activities == multiple windows at some point.

PR #2352, opened 2022-06-30
Add `refresh_rate` for `MonitorHandle`

kchibisov:
This also alters `VideoMode::refresh_rate` to return u32 which is a
monitor refresh rate in mHz.

### DIFF
--- a/src/monitor.rs
+++ b/src/monitor.rs
@@ -39,8 +39,8 @@ impl Ord for VideoMode {
         self.monitor().cmp(&other.monitor()).then(
             size.cmp(&other_size)
                 .then(
-                    self.refresh_rate()
-                        .cmp(&other.refresh_rate())
+                    self.refresh_rate_millihertz()
+                        .cmp(&other.refresh_rate_millihertz())
                         .then(self.bit_depth().cmp(&other.bit_depth())),
                 )
                 .reverse(),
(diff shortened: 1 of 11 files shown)


src/monitor.rs
```diff
@@ -68,11 +68,9 @@ impl VideoMode {
         self.video_mode.bit_depth()
     }
 
-    /// Returns the refresh rate of this video mode. **Note**: the returned
-    /// refresh rate is an integer approximation, and you shouldn't rely on this
-    /// value to be exact.
+    /// Returns the refresh rate of this video mode in mHz.
     #[inline]
-    pub fn refresh_rate(&self) -> u16 {
+    pub fn refresh_rate(&self) -> u32 {
```
dhardy, line 73: Is this a "silent break change" (changing the units without changing the API, other than the return type)? It might then be preferable to change the name, e.g. to `refresh_rate_mHz`?

Besides which, what is the purpose of exporting this value? Is it used to schedule frames (probably not the best mechanism?), or only to get a rough idea of monitor performance? In the latter case, this is unnecessary accuracy.

kchibisov: The idea is to have a fallback to schedule frames when vsync extension isn't present and we don't have any better way to do so. Aslo most APIs return it as mHz on Linux which makes sense, since your refresh rate is something like 59.997 most of the time.

maroider: > Is this a "silent break change"

I think that's likely, yeah. Creating a new method and deprecating the old one is probably the way to go.

kchibisov: I think we can safely remove the old one given that there's no need for it anymore? They are the same, just precision got changed.

maroider (approved):

src/monitor.rs
```diff
@@ -68,12 +68,10 @@ impl VideoMode {
...
     }
 
-    /// Returns the refresh rate of this video mode. **Note**: the returned
-    /// refresh rate is an integer approximation, and you shouldn't rely on this
-    /// value to be exact.
+    /// Returns the refresh rate of this video mode in mHz.
     #[inline]
-    pub fn refresh_rate(&self) -> u16 {
-        self.video_mode.refresh_rate()
+    pub fn refresh_rate_mhz(&self) -> u32 {
```
dhardy, line 73: `mhz` could easily be mistaken as `MHz`... sigh. This is a case where I don't (personally) think standard naming conventions should apply.

maroider: Writing it out as `refresh_rate_millihertz` would remove that ambiguity. I'm not that attached to the shortened name myself.

kchibisov: renamed.

PR #2350, opened 2022-06-29
Examples: less redundancy, improve fullscreen

dhardy:
Added [softbuffer](https://crates.io/crates/softbuffer) as a dev-dependency, as discussed in #2246. This is only one new example; I'm not sure it has a lot of utility. (It would be nice to see this used in at least `examples/transparency`, but this didn't work and due to the simple nature of the `softbuffer` crate I opted not to investigate further.)

Improved input handling of `examples/fullscreen` (expecting use of the console for input while in fullscreen mode is not user friendly).

Removed three examples which are redundant with others.

- [ ] Tested on all platforms changed
- [ ] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users
- [ ] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior
- [ ] Created or updated an example program if it would help users understand this functionality
- [ ] Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented

PR #2353, opened 2022-07-01
examples/window: use softbuffer

dhardy:
See also #2350, https://github.com/john01dav/softbuffer/pull/20.

### DIFF
--- a/examples/window.rs
+++ b/examples/window.rs
@@ -1,6 +1,7 @@
 #![allow(clippy::single_match)]
 
 use simple_logger::SimpleLogger;
+use softbuffer::GraphicsContext;
 use winit::{
     event::{Event, WindowEvent},
     event_loop::EventLoop,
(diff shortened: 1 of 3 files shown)


examples/window.rs
```diff
@@ -17,17 +18,29 @@ fn main() {
...
 
         match event {
+            Event::RedrawRequested(window_id) if window_id == graphics_context.window().id() => {
+                let (width, height) = {
+                    let size = graphics_context.window().inner_size();
+                    (size.width, size.height)
+                };
+                let colour = 0xffee36;
+                let mut buffer = Vec::new();
+                buffer.resize(width as usize * height as usize, colour);
```
filnet, line 35: should use `Vec::with_capacity()` to create a properly sized buffer upfront.

kchibisov: If we're talking about optimizing. Doing `vec![color; (width * height) as usize]` is the way to go.

dhardy: @filnet it only allocates once so no difference. @kchibisov is of course neater.

filnet: > @filnet it only allocates once so no difference.

I learned something today.

maroider:
The X11-only CI appears to be failing due to a somehow missing `waland-client` system library. Not sure what's up with that.

dhardy:
`softbuffer` has this dependency:

> wayland-client = {version = "0.29", features = ["use_system_lib"], default_features = false}

I tried adding `libwayland-client0` as a dependency... though since `winit` itself already uses `wayland-client` with this feature I'm doubtful.

dhardy:
No surprise, that didn't help:

> libwayland-client0 is already the newest version (1.18.0-1).

Strangely, its the `x11` build that failed.

Running `cargo test --no-run --verbose --target x86_64-unknown-linux-gnu --no-default-features --features x11` (what the command reduces to) on my machine (running X11) works.

maroider:
> Strangely, its the x11 build that failed.

Indeed. It's quite strange. Could it be that there's some side-effect `softbuffer` inadvertently relies on when we enable our wayland deps?

i509VCB:
Personally I'd be a little skeptical of using softbuffer as an example as it has some bugs. And there is also the maintenance issue as several issues and pull requests have been opened in the last 2 months without any responses.

I do recall we had a `winit-blit` project, but that never really took off.

madsmtm:
We've forked [`swbuf`](https://github.com/rust-windowing/swbuf) under @rust-windowing, so we'll probably want to use that instead - I'll close this for now, esp. also because `swbuf` doesn't support all of `winit`'s platforms yet (Android and iOS missing), but free to reopen if you update it to use that

dhardy:
Good to hear that there's a maintained fork. If you'd rather see `swbuf` support more platforms first then I guess there's no point updating this PR for now.

MarijnS95:
Randomly found this (branch deletion poked the PR), iOS/MacOS are now supported and Android will be soon: https://github.com/rust-windowing/softbuffer/pull/288

Seems like `winit` now already uses `softbuffer` in some examples, but disabled it on these platforms for "some reason" :/

PR #2354, opened 2022-07-01
Add 'Occluded(bool)' window event (currently supported on X11 and Macos)

trimental:

Checking whether a window is occluded is very useful for increasing the idling time of programs, especially when the program involves heavy rendering. You don't want to waste time rendering to a window that can't even be seen by the user. 

MacOS easily provides the occlusion state of windows unlike other platforms. By checking that `ns_window.occlusionState()` contains the `NSWindowOcclusionStateVisible` flag you can check to see if the window is being displayed to the user. This returns false when being occluded partially or if the window can be seen through blurred windows.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
-- Added `Occluded(bool)` window event (current only supported on MacOS and X11) and on macOS, added `WindowExtMacOS::is_occluded`
+- Added `Occluded(bool)` window event (current only supported on MacOS and X11)
 - On X11, fix events for caps lock key not being sent
 - Build docs on `docs.rs` for iOS and Android as well.
 - **Breaking:** Removed the `WindowAttributes` struct, since all its functionality is accessible from `WindowBuilder`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
-- On macOS, added `WindowExtWindows::is_occluded`
+- Added `Occluded(bool)` window event (current only supported on macos) and on macOS, added `WindowExtWindows::is_occluded`
 - On X11, fix events for caps lock key not being sent
 - Build docs on `docs.rs` for iOS and Android as well.
 - **Breaking:** Removed the `WindowAttributes` struct, since all its functionality is accessible from `WindowBuilder`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- On macOS, added `WindowExtWindows::is_occluded`
 - On X11, fix events for caps lock key not being sent
 - Build docs on `docs.rs` for iOS and Android as well.
 - **Breaking:** Removed the `WindowAttributes` struct, since all its functionality is accessible from `WindowBuilder`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
-- Added `Occluded(bool)` window event (current only supported on MacOS and X11)
+- Added `WindowEvent::Occluded(bool)`, currently implemented on macOS and X11.
 - On X11, fix events for caps lock key not being sent
 - Build docs on `docs.rs` for iOS and Android as well.
 - **Breaking:** Removed the `WindowAttributes` struct, since all its functionality is accessible from `WindowBuilder`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
-- Added `Occluded(bool)` window event (current only supported on macos) and on macOS, added `WindowExtWindows::is_occluded`
+- Added `Occluded(bool)` window event (current only supported on macos) and on macOS, added `WindowExtMacOS::is_occluded`
 - On X11, fix events for caps lock key not being sent
 - Build docs on `docs.rs` for iOS and Android as well.
 - **Breaking:** Removed the `WindowAttributes` struct, since all its functionality is accessible from `WindowBuilder`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
-- Added `Occluded(bool)` window event (current only supported on macos) and on macOS, added `WindowExtMacOS::is_occluded`
+- Added `Occluded(bool)` window event (current only supported on MacOS and X11) and on macOS, added `WindowExtMacOS::is_occluded`
 - On X11, fix events for caps lock key not being sent
 - Build docs on `docs.rs` for iOS and Android as well.
 - **Breaking:** Removed the `WindowAttributes` struct, since all its functionality is accessible from `WindowBuilder`.
--- a/src/event.rs
+++ b/src/event.rs
@@ -372,6 +372,14 @@ pub enum WindowEvent<'a> {
     ///
     /// At the moment this is only supported on Windows.
     ThemeChanged(Theme),
+
+    /// The window has been occluded (completely hidden from view).
+    ///
+    /// This is different to window visibility as it depends on whether the window is closed,
+    /// minimised, set invisible, or fully occluded by another window.
+    ///
+    /// At the moment this is only supported on macOS.
+    Occluded(bool),
 }
 
 impl Clone for WindowEvent<'static> {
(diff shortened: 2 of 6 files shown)


kchibisov:
Can macOS notify when the window is occluded or you should constantly poll and check if it's occluded or not?

Assuming that the window is occluded by window X, then user closes window X, how the window behind it understand that it should start drawing. Focus hasn't changed.

I'm not familiar with macOS that much, but I'm not sure how downstream, e.g. alacritty can reliably make use of this API.

trimental:
I haven't looked into whether or not macOS gives an event for when a window becomes occluded, however if you think that this api would be better as an event, I can look into that.

If the window was occluded by window X then `is_occluded() == true` and when window X is closed then `is_occluded() == false`. I've written a gameboy emulator and by adding
```
loop {
if !window.is_occluded() {
    // Insert rendering code here
}
// sleep
}
```
I've dropped cpu usage when occluded to about 2-4% from 10-12%. Sure you could add an event to signify this change instead of `is_occluded()` but downstream would still probably use a boolean that tracks whether the window is occluded. I can't think of what you would do with an occlusion event, unless you wanted to sleep completely until the window reappeared instead of checking during rendering.

trimental:
@kchibisov I've added the `Occluded(bool)` window event for you to look over.

I think perhaps both options could be taken of the event and also `is_occluded()`. The event would make adding future platforms easier, once [wayland adds occlusion detection](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/99).

trimental:
@https://github.com/alacritty/alacritty/pull/6177

src/event.rs
```diff
@@ -372,6 +372,14 @@ pub enum WindowEvent<'a> {
     ///
     /// At the moment this is only supported on Windows.
     ThemeChanged(Theme),
+
+    /// The window has been occluded (completely hidden from view).
+    ///
+    /// This is different to window visibility as it depends on whether the window is closed,
+    /// minimised, set invisible, or fully occluded by another window.
+    ///
+    /// At the moment this is only supported on macOS.
```
maroider, line 381:
```suggestion
    /// ## Platform-specific
    ///
    /// At the moment this is only supported on macOS.
```

We probably want to put platform-specific notes like this under a `## Platform-specific` heading.

src/platform/macos.rs
```diff
@@ -33,6 +33,13 @@ pub trait WindowExtMacOS {
     /// space or taking control over the entire monitor.
     fn set_simple_fullscreen(&self, fullscreen: bool) -> bool;
 
+    // Returns whether or not the window is occluded.
+    //
+    // Occlusion differs from window visibility as an occluded window
+    // can be occluded from the screen surface for reasons such as being closed,
+    // minimised, not set visible, or behind another window.
+    fn is_occluded(&self) -> bool;
```
maroider, line 41: We may want to explicitly recommend using the `Occluded` event here.

maroider (commented):
It seems to me this may conceptually overlap with `Window::is_visible` (which was introduced in #2169), although the specific macOS API in use in this PR appears to be different.

trimental:
@maroider Yes they differ in the fact that `Window::is_visible` is only concerned with whether the window is currently on the screen, even if it is hidden / occluded by another window, therefore if a window isn't explicitly minimised (sort of rare in macOS) then it returns true. To me it seems that `Window::is_occluded` serves the main use case of `Window::is_visible` but just to a better extent (ie you don't want to paint to windows that can't be directly seen)

kchibisov:
>It seems to me that this may conceptually overlap with Window::is_visible

Those are different though, visible is visible or not, it can't control rendering at all, but occluded is much more involved here, since it handles previews and such. Basically `is_visible` is a getter for `set_visible`.

So I'm inclined to have `Occluded` event, since it can be of help, given that it guarantees that you can stop rendering.

src/platform/macos.rs
```diff
@@ -33,6 +33,16 @@ pub trait WindowExtMacOS {
...
     fn set_simple_fullscreen(&self, fullscreen: bool) -> bool;
 
+    /// Returns whether or not the window is occluded.
+    ///
+    /// Occlusion differs from window visibility as an occluded window
+    /// can be occluded from the screen surface for reasons such as being closed,
+    /// minimised, not set visible, or behind another window.
+    ///
+    /// Alternatively the event [`WindowEvent::Occluded`](crate::event::WindowEvent::Occluded)
+    /// can be used to track the occlusion state of the window.
```
maroider, line 43:
```suggestion
    /// Alternatively, the [`WindowEvent::Occluded`] event
    /// can be used to track the occlusion state of the window.
    ///
    /// [`WindowEvent::Occluded`]: crate::event::WindowEvent::Occluded
```
nit: grammar, inline link instead of reference link

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- Added `Occluded(bool)` window event (current only supported on macos) and on macOS, added `WindowExtMacOS::is_occluded`
```
maroider, line 11:
```suggestion
- On macOS, added `WindowEvent::Occluded(bool)` and `WindowExtMacOS::is_occluded`.
```
I think this would be more concise, but I'm not sure to what degree we've used this sort of formulation for sparsely-supported platform-independent APIs in the past.

trimental: I don't know if I agree with this. Saying "On macOS, added `WindowEvent::Occluded(bool)`" in my opinion seems to imply this windows event is only accessible on macOS. Users attempting to match `WindowEvent` will have to handle the `Occluded` event on all platforms (or just `_ =>`) , and other platforms can send this event in the future (ie [wayland](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/99))

trimental:
I've added support for X11 if anyone wants to test that. Seems to work fine for compositors that actually track occlusion such as openbox.

trimental:
Given now two platforms implement the `Occluded` event, I've decided to remove `is_occluded`, as I believe the event makes more sense for platform cross compatibility

kchibisov:
Seems like Windows has similar API, I'll send a can try to send a patch for that as well.

https://docs.microsoft.com/en-us/windows/win32/api/dxgi1_2/nf-dxgi1_2-idxgifactory2-registerocclusionstatuswindow

kchibisov:
Yeah, I also realized that after looking more into it. So I guess we can ignore it for now.

src/event.rs
```diff
@@ -372,6 +372,16 @@ pub enum WindowEvent<'a> {
...
     ThemeChanged(Theme),
+
+    /// The window has been occluded (completely hidden from view).
+    ///
+    /// This is different to window visibility as it depends on whether the window is closed,
+    /// minimised, set invisible, or fully occluded by another window.
+    ///
+    /// ## Platform-specific
+    ///
+    /// At the moment this is only supported on macOS and X11.
```
kchibisov, line 383: Could you describe platform section in the common format we use? like

```
/// Platform-specific behavior:
/// - **iOS / Android / Web :** Unsupported.
```

src/event.rs
```diff
@@ -461,6 +471,7 @@ impl Clone for WindowEvent<'static> {
             ScaleFactorChanged { .. } => {
                 unreachable!("Static event can't be about scale factor changing")
             }
+            Occluded(o) => Occluded(*o),
```
kchibisov, line 474:
```suggestion
            Occluded(occluded) => Occluded(*occluded),
```

src/event.rs
```diff
@@ -546,6 +557,7 @@ impl<'a> WindowEvent<'a> {
             Touch(touch) => Some(Touch(touch)),
             ThemeChanged(theme) => Some(ThemeChanged(theme)),
             ScaleFactorChanged { .. } => None,
+            Occluded(o) => Some(Occluded(o)),
```
kchibisov, line 560:
```suggestion
            Occluded(occluded) => Some(Occluded(occluded)),
```

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- Added `Occluded(bool)` window event (current only supported on MacOS and X11)
```
kchibisov, line 11:
```suggestion
- Added `WindowEvent::Occluded(bool)`, currently implemented on macOS and X11.
```

kchibisov (changes requested):

src/event.rs
trimental: I added wayland into that as well

src/event.rs
```diff
@@ -378,9 +378,8 @@ pub enum WindowEvent<'a> {
     /// This is different to window visibility as it depends on whether the window is closed,
     /// minimised, set invisible, or fully occluded by another window.
     ///
-    /// ## Platform-specific
-    ///
-    /// At the moment this is only supported on macOS and X11.
+    /// Platform-specific behavior:
+    /// - **iOS / Android / Web / Wayland :** Unsupported.
```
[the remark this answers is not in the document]
trimental: 🤦

kchibisov (approved):

trimental:
A heads up that it seems that window occlusion might be broken on the macOS 13 beta. macOS sets the window as not occluded when the app starts up, and will only set it as occluded if the window is minimised. Ie it doesn't consider being occluded by other windows. It would be nice if someone else could check if this is broken for them as well.

PR #2358, opened 2022-07-04
Generate docs.rs documentation with doc_auto_cfg

i509VCB:
- [ ] Tested on all platforms changed (Linux tested at least)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On macOS, added support for `WindowEvent::ThemeChanged`.
 - **Breaking:** Removed `WindowBuilderExtWindows::with_theme` and `WindowBuilderExtWayland::with_wayland_csd_theme` in favour of `WindowBuilder::with_theme`.
 - **Breaking:** Removed `WindowExtWindows::theme` in favour of `Window::theme`.
+- Enabled `doc_auto_cfg` when generating docs on docs.rs for feature labels.
 
 # 0.27.4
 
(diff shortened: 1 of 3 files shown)


MarijnS95 (changes requested):
Thanks for replicating https://github.com/rust-windowing/android-ndk-rs/pull/293 on the `winit` side, I almost forgot :grimacing:

i509VCB:
I guess I forgot this, but I have changed the pull request to use doc_auto_cfg

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On macOS, added support for `WindowEvent::ThemeChanged`.
 - **Breaking:** Removed `WindowBuilderExtWindows::with_theme` and `WindowBuilderExtWayland::with_wayland_csd_theme` in favour of `WindowBuilder::with_theme`.
 - **Breaking:** Removed `WindowExtWindows::theme` in favour of `Window::theme`.
+- Enable `doc_auto_cfg` when generating docs on docs.rs for feature labels.
```
MarijnS95, line 30: Seems `winit` changelog (unlike `glutin` and hopefully also `android-ndk-rs`) uses past tense in their changelog: this should probably be `Enabled`.

src/lib.rs
```diff
@@ -135,6 +135,10 @@
 #![deny(rustdoc::broken_intra_doc_links)]
 #![deny(clippy::all)]
 #![cfg_attr(feature = "cargo-clippy", deny(warnings))]
+// Enable doc_auto_cfg when generating docs.
```
MarijnS95, line 138: The user can read themselves that it is being enabled when the `docsrs` cfg is true, no need to paraphrase the code here.

Why don't you explain what it means/does instead? Otherwise, remove this line altogether.

MarijnS95 (approved):
Thanks!

madsmtm:
Thanks!

Issue #2359, opened 2022-07-04
Using `XI_Raw*`-events for input processing on X11 causes a risk of input-bind clashing between the Application and WM


MarcusGrass:
I wrote a WM and got a [bug report of weird scrolling behaviour using alacritty](https://github.com/MarcusGrass/pgwm/issues/1).

The issue:
Alacritty receives and acts upon input that should be grabbed.

What I believe is the cause:
Alacritty uses glutin which uses winit for input. Winit listens to `XI_Raw*` events which, according to [this ancient changelog](https://www.x.org/releases/X11R7.7/doc/inputproto/XI2proto.txt) are sent regardless of grab state. 
Using `XI_Raw*` means that if any Rust application built on winit has clashing input-bindings with a WM it'll cause a bug.
In this particular case, Mod+scroll to resize a window causes unwanted scrolling inside the window as well.

This isn't really a major bug and I'm guessing that there's a reason to use `XI_Raw*` over the alternatives. But if not, I think it'd be good to not use them to prevent clashes on X11

maroider:
The only "raw" event we use for scrolling is `XI_RawMotion`, but that only feeds into a `DeviceEvent`, which Alacritty (to my knowledge) ignores. I suspect that the issue has something to do with how we use `XI_Motion` for regular scroll events (we also use `XI_ButtonPress` and `XI_ButtonRelease`, but those shouldn't suffer from the "xinput2 ignores grab state for raw events" issue).

MarcusGrass:
I think it's the below code, since the scroll is a ButtonPressEvent on X11, and since winit uses XI_RAW I can't grab it
```Rust 
ffi::XI_RawButtonPress | ffi::XI_RawButtonRelease => {
                        let xev: &ffi::XIRawEvent = unsafe { &*(xev.data as *const _) };
                        if xev.flags & ffi::XIPointerEmulated == 0 {
                            callback(Event::DeviceEvent {
                                device_id: mkdid(xev.deviceid),
                                event: DeviceEvent::Button {
                                    button: xev.detail as u32,
                                    state: match xev.evtype {
                                        ffi::XI_RawButtonPress => Pressed,
                                        ffi::XI_RawButtonRelease => Released,
                                        _ => unreachable!(),
                                    },
                                },
                            });
                        }
                    }
```
![image](https://user-images.githubusercontent.com/34198073/177273721-2e09780a-3588-4a6f-ade0-f68f42bfb5fa.png)
[...]

PR #2362, opened 2022-07-05
Primarily use ButtonPress and ButtonRelease for mouse events

MarcusGrass:

(Sorry for the weird PR spam)

This change makes ButtonPresses like KeyPresses use regular events instead of XI-events which allows the scroll-button to be grabbed, the reason for wanting to do this is stated in https://github.com/rust-windowing/winit/issues/2359.

If I understood what's happening here correctly, and I might not have:
You cannot grab XI_Button events and still get regular Button events. XI uses motion for scrolling and doesn't count scrolling as a button press.
This leads to the problem where you can't grab scroll-keys as they aren't registered as being pressed when scrolling happens.

This change modifies that to make winit use regular ButtonPressEvents for scrolling and stops it listening for XI_Button events.
Interestingly there already was logic implemented for scrolling using buttons, but because XI doesn't count scrolling as buttons that logic was never activated, it's now used in the regular ButtonPress handling instead.

For extension devices not targeting the core-pointer scrolling behaviour should be the same, although that's hard to test.

Tested on x11 and now MouseWheel events only arrive if not grabbed. Additionally tested building glutin and then alacritty on top of this change and the problem is gone.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - Fix on macOS `WindowBuilder::with_disallow_hidpi`, setting true or false by the user no matter the SO default value. 
 - `EventLoopBuilder::build` will now panic when the `EventLoop` is being created more than once.
 - Added `From<u64>` for `WindowId` and `From<WindowId>` for `u64`.
+- On X11, take mouse-events except motion from vanilla X11 instead of xinput2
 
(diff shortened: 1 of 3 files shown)


notgull:
Is there information lost when moving from `XIButtonPress` to `xproto`'s `ButtonPress`? That would be my main worry with this PR. Why can't `xinput` grab the scroll button?

kchibisov:
closing due to lack of activity.

PR #2364, opened 2022-07-06
Remove the deprecated `modifiers` field on several events

maroider:
It looks like the Web backend doesn't emit ModifiersChanged events at all right now. This has to be resolved before including this change upstream. I've elected to leave "modifier-querying" code in-place for now in order to avoid accidentally introducing any regressions.


kchibisov:
done upstream.

PR #2366, opened 2022-07-06
On Wayland send Focused(false) for new window

kchibisov:
On Wayland winit will always get an explicit focused event from the
system and will transfer it downstream. So send focused false to enforce
it.

--

This doesn't affect the end users, since they'll have explicit event and shouldn't make any assumption like if I documented that window is unfocused by default.

### DIFF
--- a/src/platform_impl/linux/wayland/window/mod.rs
+++ b/src/platform_impl/linux/wayland/window/mod.rs
@@ -254,6 +254,12 @@ impl Window {
 
         winit_state.window_map.insert(window_id, window_handle);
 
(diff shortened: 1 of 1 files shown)


kchibisov:
>This probably won't break user code, but it still feels very weird to me to just send a fake event like this.

Well, it's not a fake, since it's not focused. The problem is that downstream clients think that default focus is `true`. And I don't think we can just state, assume that the window isn't focused on startup. While I do like that the window isn't focused on startup statement, I don't think it'll work across all platforms become internally explicit.

Right now it solves issues on Wayland, since on Wayland the window isn't focused by default.

PR #2369, opened 2022-07-09
wayland: Add support for fractional scaling

mahkoh:

---

Depends on https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/143

### DIFF
--- a/src/platform_impl/linux/wayland/event_loop/mod.rs
+++ b/src/platform_impl/linux/wayland/event_loop/mod.rs
@@ -374,10 +374,11 @@ impl<T: 'static> EventLoop<T> {
             });
 
             for (window_id, window_compositor_update) in window_compositor_updates.iter_mut() {
-                if let Some(scale_factor) = window_compositor_update.scale_factor.map(|f| f as f64)
-                {
+                if let Some(scale_factor) = window_compositor_update.scale_factor {
                     let mut physical_size = self.with_state(|state| {
                         let window_handle = state.window_map.get(window_id).unwrap();
+                        *window_handle.scale_factor.lock().unwrap() = scale_factor;
+
                         let mut size = window_handle.size.lock().unwrap();
 
                         // Update the new logical size if it was changed.
@@ -409,6 +410,13 @@ impl<T: 'static> EventLoop<T> {
                 if let Some(size) = window_compositor_update.size.take() {
                     let physical_size = self.with_state(|state| {
                         let window_handle = state.window_map.get_mut(window_id).unwrap();
+
+                        if let Some(vp) = &window_handle.viewport {
+                            // If we have a viewport then we support fractional scaling. As per the
+                            // protocol, we have to set the viewport size of the requested size.
+                            vp.set_destination(size.width as _, size.height as _);
+                        }
+
                         let mut window_size = window_handle.size.lock().unwrap();
 
                         // Always issue resize event on scale factor change.
@@ -419,9 +427,8 @@ impl<T: 'static> EventLoop<T> {
                             None
                         } else {
                             *window_size = size;
-                            let scale_factor =
-                                sctk::get_surface_scale_factor(window_handle.window.surface());
-                            let physical_size = size.to_physical(scale_factor as f64);
+                            let scale_factor = window_handle.scale_factor();
+                            let physical_size = size.to_physical(scale_factor);
                             Some(physical_size)
                         };
 
--- a/src/platform_impl/linux/wayland/window/mod.rs
+++ b/src/platform_impl/linux/wayland/window/mod.rs
@@ -11,12 +11,14 @@ use raw_window_handle::{
     RawDisplayHandle, RawWindowHandle, WaylandDisplayHandle, WaylandWindowHandle,
 };
 use sctk::window::Decorations;
+use wayland_protocols::viewporter::client::wp_viewporter::WpViewporter;
 
 use crate::dpi::{LogicalSize, PhysicalPosition, PhysicalSize, Position, Size};
 use crate::error::{ExternalError, NotSupportedError, OsError as RootOsError};
 use crate::platform_impl::{
-    Fullscreen, MonitorHandle as PlatformMonitorHandle, OsError,
-    PlatformSpecificWindowBuilderAttributes as PlatformAttributes,
+    wayland::protocols::wp_fractional_scale_manager_v1::WpFractionalScaleManagerV1,
+    wayland::protocols::wp_fractional_scale_v1, Fullscreen, MonitorHandle as PlatformMonitorHandle,
+    OsError, PlatformSpecificWindowBuilderAttributes as PlatformAttributes,
 };
 use crate::window::{
     CursorGrabMode, CursorIcon, Theme, UserAttentionType, WindowAttributes, WindowButtons,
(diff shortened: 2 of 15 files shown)


mahkoh:
PS: I've noticed that winit never destroys any wayland objects. I've kept it that way with the new objects I've added.

kchibisov:
@mahkoh winit should destroy the surface via sctk iirc, unless sctk changed that?

I don't see anything wrong in destroying those, but there might be a reason why it's that way.

In general, unless the protocol will be accepted there won't be any action on that PR.

mahkoh:
>@mahkoh winit should destroy the surface via sctk iirc, unless sctk changed that?

I don't believe this applies to the wl_surface itself.

kchibisov:
>I don't believe this applies to the wl_surface itself.

Ah, I think it doesn't destroy it in the Window, since it's basically a handle, yeah, should do that, would you send a patch against master or I should do so?

mahkoh:
Please go ahead. I'll add destructor calls for the new objects I've add in this PR.

kchibisov:
@mahkoh it's being dropped here https://github.com/Smithay/client-toolkit/blob/f03c4b7dea89648dbcba43de80eb14687e0e9da8/src/shell/xdg.rs#L139, so no need for winit to drop it explicitly as well.

mahkoh:
`self.surface` refers to an xdg_surface not to a wl_surface.

kchibisov:
Hm, you're right. I've confused the place, but as well as the surface that is being dropped by sctk, since the one it's dropping is invisible one for decorations. So yeah, we should drop this surface in winit.

mahkoh:
Keep in mind that the wl_surface must be dropped after the xdg_surface. Otherwise a protocol error occurs.

mahkoh:
They continue to be rendered at integer scales (usually the fractional scale rounded up.) In my tests I did not see any problems with that.

kchibisov:
I mean, they should likely scale font if they do font rendering to account for that? That's not the with default sctk decorations, but it's the case with adwaita frame.

In general I'd not take any action further here, unless the protocol get merged and sctk get a release...

mahkoh:
>I mean, they should likely scale font if they do font rendering to account for that?

Yes, they scale the font at the next integer scale.

https://user-images.githubusercontent.com/1882250/179069105-e78bbdfa-7d93-4c0b-bf9b-c9af3ede5515.mp4

>In general I'd not take any action further here, unless the protocol get merged and sctk get a release...

This has nothing to do with sctk. It provides a strict improvement for the content of the window while leaving the frame unchanged.

mahkoh:
There will be at least one breaking change in the protocol: The scale type has been changed from fixed to uint with a denominator of 120. I will try to adapt this PR soon.

kchibisov:
@mahkoh I'd assume stabilization is soon™

mahkoh:
Incidentally working on rebasing this right now.

mahkoh:
The protocol has been accepted.

kchibisov:
@mahkoh I'd wait until the sway merges their stuff wrt fractional scaling so I'd have easier time testing it. Given that implementation isn't in wlroots yet it may take some time, but we're not in the rush anyway.

For now just schedule it for winit 0.28.

mahkoh:
Updated with the final protocol text. The SDL implementation has been merged: https://github.com/libsdl-org/SDL/pull/5906

What is the timeline for 0.28?

mahkoh:
Added the missing destroy calls for the add-on objects. See https://github.com/rust-windowing/winit/pull/2369#issuecomment-1179556061

kchibisov:
>What is the timeline for 0.28?

There's no timeline.

I'd basically wait for sway to get fractional scaling, if it won't and we'll be doing 0.28 I'll go back to it.

oxalica:
> I'd basically wait for sway to get fractional scaling, if it won't and we'll be doing 0.28 I'll go back to it.

Note that KDE/kwin [has already merged](https://invent.kde.org/plasma/kwin/-/merge_requests/2598) the implementation of fractional scaling and it would be available since the next release (5.27). I hope we can get it in soon.

src/platform_impl/linux/wayland/window/mod.rs
```diff
@@ -86,33 +91,64 @@ impl Window {
...
+        let mut viewport = None;
+        let mut scale_factor = 1.0;
+
+        let viewporter = event_loop_window_target.env.get_global::<WpViewporter>();
+        let fractional_scale_manager = event_loop_window_target
             .env
-            .create_surface_with_scale_callback(move |scale, surface, mut dispatch_data| {
+            .get_global::<WpFractionalScaleManagerV1>();
+
+        if let (Some(vp), Some(fsm)) = (viewporter, fractional_scale_manager) {
```
kchibisov, line 104: I think you could use full names other than `vp` and `fsm`.

src/platform_impl/linux/wayland/window/mod.rs
```diff
@@ -86,33 +91,64 @@ impl Window {
...
+                    .unwrap()
+                    .scale_factor = Some(scale as f64 / 120.0);
                 winit_state
                     .window_user_requests
                     .get_mut(&window_id)
                     .unwrap()
                     .refresh_frame = true;
+            });
+            fractional_scale = Some(fs.detach());
+            viewport = Some(vp.get_viewport(&surface).detach());
```
kchibisov, line 123: Maybe we should store them in some common struct like

```rust
struct FractionalScalingState {
  viewporter: Viewporter,
  fractional_scaling: FractionalScaling,
}
```

That should simplify a bunch of `Option` handling, since you have either both `None` or both `Some` at the same time.

src/platform_impl/linux/wayland/window/mod.rs
```diff
@@ -86,33 +91,64 @@ impl Window {
...
+        } else {
+            surface = event_loop_window_target
+                .env
+                .create_surface_with_scale_callback(move |scale, surface, mut dispatch_data| {
+                    let winit_state = dispatch_data.get::<WinitState>().unwrap();
 
-                // Set pending scale factor.
-                window_compositor_update.scale_factor = Some(scale);
+                    // Get the window that received the event.
```
kchibisov, line 130: You can extract all of that into a single free function `fn apply_scale(window_id, scale, dispatch_data)`, so you can use it in both fractional and normal path.

src/platform_impl/linux/wayland/window/mod.rs
kchibisov, line 104: You could do `let surface = if` here.

src/platform_impl/linux/wayland/event_loop/mod.rs
```diff
@@ -409,6 +410,13 @@ impl<T: 'static> EventLoop<T> {
                 if let Some(size) = window_compositor_update.size.take() {
                     let physical_size = self.with_state(|state| {
                         let window_handle = state.window_map.get_mut(window_id).unwrap();
+
+                        if let Some(vp) = &window_handle.viewport {
+                            // If we have a viewport then we support fractional scaling. As per the
+                            // protocol, we have to set the viewport size of the requested size.
```
kchibisov, line 416:
```suggestion
                            // If we have a viewport then we support fractional scaling. As per the
                            // protocol, we have to set the viewport size of the size prior scaling.
```

kchibisov (changes requested):
I haven't tested it, just reviewed based on the protocols specification.

kchibisov:
@mahkoh I could finish this PR myself given that everything is more of a codestyle rant. Just let me know (though I'd push anyway in-before release).

mahkoh:
Sorry I'm busy with work at the moment. If you could finish it that would be great.

kchibisov:
Applied.

Thanks.

PR #2370, opened 2022-07-09
On Wayland, drop wl_surface on window close

kchibisov:

r @mahkoh

### DIFF
--- a/src/platform_impl/linux/wayland/window/shim.rs
+++ b/src/platform_impl/linux/wayland/window/shim.rs
@@ -1,4 +1,5 @@
 use std::cell::Cell;
+use std::mem::ManuallyDrop;
 use std::sync::{Arc, Mutex};
 
 use sctk::reexports::client::protocol::wl_compositor::WlCompositor;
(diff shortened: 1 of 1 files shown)


mahkoh:
You could do this without unsafe by having a `SurfaceHolder` stored in WindowHandle after `window`. `SurfaceHolder` would then destroy the surface when dropped. Your code seems correct.

kchibisov:
@mahkoh I do like it with unsafe, since I'm being certain that way in which order rust will drop stuff, in fact that's what's ManuallyDrop was meant for. If I store after window I wouldn't be that certain.

mahkoh:
You could be certain because struct is specified to drop fields from top to bottom. Anyway, the current version is fine by me.

kchibisov:
>You could be certain because struct is specified to drop fields from top to bottom. Anyway, the current version is fine by me.

Yeah, but then some other folks who'll take a look at the code should be aware that ordering isn't changed, that way it's more clear. Though, it's a matter of opinion.

PR #2371, opened 2022-07-11
Reduce `WM_PAINT` messages of thread target window

### LINKED ISSUE #2367: Intense `WM_PAINT` msg in `thread_event_target_callback`
I encounter a delayed mouse interaction when using webview2 with winit.
Here's the branch to reproduce: https://github.com/tauri-apps/wry/tree/winit-reproduce
Run `cargo r --example hello_world` and press any tab or list constantly will find it gets a few seconds delay.

I found if we reduce[ `RedrawWindow` calls](https://github.com/rust-windowing/winit/blob/master/src/platform_impl/windows/event_loop.rs#L2178) in `thread_event_target_callback`, it can resolve this issue.
Could we just call `RedrawWindow` only when `thread_event_target_callback` need to handle the message instead of calling  `RedrawWindow` no matter what?

wusyong:
Fixes #2367

Since there's no comment on the issue, I would like to open a PR and give it a try. Examples are working and we can still get nonstop messages with `ControlFlow::Poll` if we want.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On Windows, added `WindowExtWindows::set_skip_taskbar` and `WindowBuilderExtWindows::with_skip_taskbar`.
 - On Windows, added `EventLoopBuilderExtWindows::with_msg_hook`.
 - On Windows, remove internally unique DC per window.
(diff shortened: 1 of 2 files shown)


maroider:
I'll take a look at this once I feel I have a better grasp of what exactly this part of the code is supposed to do. I'm afraid to touch it with my current level of understanding.

kchibisov:
Is this still an issue on the latest master?

PR #2374, opened 2022-07-14
windows: Use correct value for mouse wheel delta

grovesNL:

During the switch to windows-sys, it looks like we unintentionally started using `usButtonFlags` instead of `usButtonData` for this event.

Original winapi version: `usButtonData` is used for the delta
https://github.com/rust-windowing/winit/blob/f3f6f1008a5891ed0c9e5e8f26568f8f2d49f164/src/platform_impl/windows/event_loop.rs#L2155-L2164
After switch to windows-sys: `usButtonFlags` is used for the delta
https://github.com/rust-windowing/winit/blob/b222dde83504e402d51fa40a3c78dee40cb2ee5f/src/platform_impl/windows/event_loop.rs#L2153-L2163


msiglreith (approved):
nicely spotted!

PR #2376, opened 2022-07-15
Update raw-window-handle to v0.5.0

kchibisov:
This updates raw-window-handle to v0.5.0. The change is breaking
and removes the old `HasRawWindowHandle` trait implementation,
but that's intentional given that internal structs got renamed.

### DIFF
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -13,6 +13,7 @@ use std::{error, fmt};
 
 use instant::Instant;
 use once_cell::sync::OnceCell;
+use raw_window_handle::{HasRawDisplayHandle, RawDisplayHandle};
 
 use crate::{event::Event, monitor::MonitorHandle, platform_impl};
 
@@ -334,6 +335,13 @@ impl<T> EventLoopWindowTarget<T> {
             target_os = "openbsd"
         ))]
         self.p.set_device_event_filter(_filter);
+    }
+}
+
+unsafe impl<T> HasRawDisplayHandle for EventLoopWindowTarget<T> {
+    /// Returns a [`raw_window_handle::RawDisplayHandle`] for the event loop.
+    fn raw_display_handle(&self) -> RawDisplayHandle {
+        self.p.raw_display_handle()
     }
 }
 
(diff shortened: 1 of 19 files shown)


maroider:
I take it we're aiming to land this in `v0.27.0`?

kchibisov:
>I take it we're aiming to land this in v0.27.0?

Yeah, I think so.

src/event_loop.rs
```diff
@@ -334,6 +334,11 @@ impl<T> EventLoopWindowTarget<T> {
         ))]
         self.p.set_device_event_filter(_filter);
     }
+
+    /// Returns a [`raw_window_handle::RawDisplayHandle`] for the event loop.
+    pub fn raw_display_handle(&self) -> raw_window_handle::RawDisplayHandle {
+        self.p.raw_display_handle()
+    }
```
maroider, line 341: Why did you choose an inherent method instead of using the `HasRawDisplayHandle` trait?

kchibisov: I'm still not sure about those, though I think using trait methods is preferable way, I'd bring them back.

MarijnS95: We "brought them back" for `android-ndk-rs` since nothing in `raw-window-handle` states that these traits are suddenly deprecated (they're for ease of use when passing a window type, though I agree that it's not absolutely necessary since `raw_window_handle::RawDisplayHandle` might just as well be accepted by APIs without having to deal with borrows).

This is public API for the event loop, right? Not sure if every type in `winit` should implement this, for example a single event loop can have multiple windows associated so having the trait/function doesn't make much sense?

kchibisov:
I've updated implementation and now use traits for it. Waiting on ndk to update, so android will start built.

MarijnS95:
> I've updated implementation and now use traits for it. Waiting on ndk to update, so android will start built.

You might get away with bumping the commit hash to https://github.com/rust-windowing/android-ndk-rs/commit/814be0864bd05f6fff616299ce25c075487fa244?

maroider (commented):
I think we should be mostly good to go now.

maroider (approved):

PR #2381, opened 2022-07-19
Bump sctk-adwaita to 0.4.1

### LINKED ISSUE #2373: winit suddenly requires `cmake` dependency
It looks like ever since `sctk-adwaita` got upgraded (and their dependencies got upgraded) this crate is dragging in `cmake`, polluting everybody's build system and setup. Our project has a zero-tolerance policy for `cmake` (or other build time tools since we just want to be able to build trough `cargo` with no additional tools required to be set up).

Dependency tree seems to be:

`sctk-adwaita` ->`crossfont` -> `servo-fontconfig` -> `servo-fontconfig-sys` -> `expat-sys` -> `cmake`

Presumably all of this isn't strictly required and can be disabled?

kchibisov:
This should force the use of system libraries for Fontconfig
and freetype instead of building them with cmake if missing.

Fixes #2373.

### DIFF
--- a/src/platform_impl/macos/app_delegate.rs
+++ b/src/platform_impl/macos/app_delegate.rs
@@ -72,7 +72,7 @@ extern "C" fn dealloc(this: &Object, _: Sel) {
         let state_ptr: *mut c_void = *(this.get_ivar(AUX_DELEGATE_STATE_NAME));
         // As soon as the box is constructed it is immediately dropped, releasing the underlying
         // memory
-        Box::from_raw(state_ptr as *mut RefCell<AuxDelegateState>);
+        drop(Box::from_raw(state_ptr as *mut RefCell<AuxDelegateState>));
     }
 }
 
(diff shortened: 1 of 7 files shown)


src/platform_impl/macos/app_delegate.rs
```diff
@@ -72,7 +73,7 @@ extern "C" fn dealloc(this: &Object, _: Sel) {
         let state_ptr: *mut c_void = *(this.get_ivar(AUX_DELEGATE_STATE_NAME));
         // As soon as the box is constructed it is immediately dropped, releasing the underlying
         // memory
-        Box::from_raw(state_ptr as *mut RefCell<AuxDelegateState>);
+        mem::drop(Box::from_raw(state_ptr as *mut RefCell<AuxDelegateState>));
```
maroider, line 76: Isn't `drop` already included as part of the prelude?

maroider (approved):
I suppose the dropping tweak is fine, albeit unrelated.

PR #2385, opened 2022-07-20
Add method to hook xlib error handler

### LINKED ISSUE #2378: Publicly expose last error received from X11
Unfortunately in multi-threaded environment there's no way to handle errors safely other than setting error handling routine and store the error information in some thread safe structure like we do in winit with mutex.

However the problem with xlib here is that error handling callback is global for the entire process, so if glutin wants to make use of GLX, but doesn't want to depend on winit, it must get the error information out of winit somehow(glutin can call X related functions to perform all it wants, it just has no way to safely obtain errors from X11.

For more one could read https://www.remlab.net/op/xlib.shtml it has a pretty good explanation on that matter.

The thing with winit is that it already exposes the error via `XConnection`, however it's hidden from the end users and is being used internally by some crates that know about it.

So, I'd suggest to document that and expose last error from X11 in some more defined way. Either via proving it in some sort of `Arc<Mutex<>>` or via pointer to some function to call in hope to obtain the error information in a thread safe way.

Note, that XCB doesn't have mentioned issues in general, however GLX and some other APIs do require the use of Xlib unfortunately.

cc @maroider

kchibisov:
This should help glutin to handle errors coming from GLX
and provide multithreading access in a safe way.

Fixes #2378.

### DIFF
--- a/src/platform_impl/linux/mod.rs
+++ b/src/platform_impl/linux/mod.rs
@@ -26,6 +26,8 @@ use raw_window_handle::{RawDisplayHandle, RawWindowHandle};
 pub use self::x11::XNotSupported;
 #[cfg(feature = "x11")]
 use self::x11::{ffi::XVisualInfo, util::WindowType as XWindowType, XConnection, XError};
+#[cfg(feature = "x11")]
+use crate::platform::unix::XlibErrorHook;
 #[cfg(feature = "wayland")]
 use crate::window::Theme;
 use crate::{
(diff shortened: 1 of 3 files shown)


src/platform_impl/linux/mod.rs
```diff
@@ -607,7 +618,10 @@ unsafe extern "C" fn x_error_callback(
             minor_code: (*event).minor_code,
         };
 
-        error!("X11 error: {:#?}", error);
+        // Don't log error.
+        if should_error {
+            error!("X11 error: {:#?}", error);
+        }
```
maroider, line 624: Is logging an error when there are no error hooks not interesting?

kchibisov: It's. I'll fix it.

notgull:
I'd rather it be a `Box<dyn Fn()>` rather than an `fn()`, so that closures can be attached to the functions.

kchibisov:
I thought about, but I'm not sure. Given that it must be send + sync.

maroider:
If we want closures, then I don't think there's necessarily anything wrong with requiring said closures to be `Send + Sync` when the alternative is to use global mutable state. Either way is fine by me.

kchibisov:
I've updated to use closures instead.

Issue #323, opened 2017-10-16
More granular control over decorations on WindowBuilder


sodiumjoe:
Per https://github.com/jwilm/alacritty/issues/686 and https://github.com/jwilm/alacritty/issues/168

iterm2 and Atom both allow customizations of the titlebar. It would be nice to have more granular control over:

* hide titlebar without removing borders and resize controls
* background color
* foreground color

vberger:
How would other platform allow it?

For wayland there is not technical block, as we are drawing the decorations. It's "just" a matter of implementing in in wayland-window.

PaperPrototype:
So has this been implemented yet?

PaperPrototype:
Never mind. https://github.com/rust-windowing/winit/pull/389

PR #2396, opened 2022-07-26
Pen and pressure support for Windows and Android

Atilogit:

### About the TODO for getToolType:

My PR for the ndk repo [#323](https://github.com/rust-windowing/android-ndk-rs/pull/323) needs to go through before this can be implemented.
Using the event source worked reliably on my devices so this workaround should be good for now.

This relates to #99

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- Added support for digital pen information in `WindowEvent::Touch`.
+- On Windows and Android, implemented digital pen information.
+- On Android, implemented touch pressure.
 - On Windows, fix hiding a maximized window.
 - On Android, `ndk-glue`'s `NativeWindow` lock is now held between `Event::Resumed` and `Event::Suspended`.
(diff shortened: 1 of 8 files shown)


Doublonmousse:
I think this functionality would greatly benefit from the use of WinRT either as a supplement of the Win32 calls on Windows to get the pen information or to replace them completely (keeping potentially both with conditional flags for different windows version if version compatibility is too much of an issue otherwise).

Indeed on Windows :
- Pen pressure is limited to 1024 levels with Win32, not with WinRT (returns a f32 with a precision that can be larger, 4096 levels between 0 and 1 if the pen supports it)
- Better support for the barrel button/side buttons. With the Win32 API, no information is available on the status of the pen when it's hovering (unlike linux). With the WinRT API, this information is always available, when the pen is hovering or on the surface. This also means a better support/user experience overall for pens over Win32 (see the issue here https://github.com/microsoft/WindowsAppSDK/issues/233 for reference)
- there are other API/events sent by WinRT over Win32: event for undocking the pen, and haptic feedback but I guess it's way less interesting to support (and probably very windows specific)

I'm testing these changes (POC at this point, I'm only printing out the info) here : https://github.com/Doublonmousse/winit/tree/winit-pen

ActuallyHappening:
Note: This does not implement the double tap functionality of Apple Pens

PR #2397, opened 2022-07-26
ci: manually point ANDROID_NDK_ROOT to latest supplied version

MarijnS95:
It seems the symlink to `ndk-bundle` and this environment variable pointing to it have been removed to prevent the sdkmanager from failing, when finding the SDK setup to be in an "indeterminate" state.  It is now up to the users themselves to install an NDK through that tool or point the right variables to a preinstalled "latest" NDK.

https://github.com/actions/virtual-environments/issues/2689
https://github.com/actions/virtual-environments/issues/5879
https://github.com/actions/virtual-environments/pull/5926

---

Draft because I'm still waiting for confirmation that the removal of this env var is intended, instead of just pointing it to the latest LTS release _without going through a symlink_ to match previous behaviour.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,6 +67,11 @@ jobs:
         targets: ${{ matrix.platform.target }}
         components: clippy
 
+    - name: Setup NDK path
+      shell: bash
+      # "Temporary" workaround until https://github.com/actions/virtual-environments/issues/5879#issuecomment-1195156618
+      # gets looked into.
+      run: echo "ANDROID_NDK_ROOT=$ANDROID_NDK_LATEST_HOME" >> $GITHUB_ENV
     - name: Install Linux dependencies
       if: (matrix.platform.os == 'ubuntu-latest')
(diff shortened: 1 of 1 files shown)


kchibisov:
In general I'm fine with such solution tbqh, could change it whenever we feel like if upstream changes it again.

MarijnS95:
Let's get this in then to unblock everyone as the CI failures are super annoying. We can remove it again when there's another sensible default).

Note that we're targeting latest now instead of LTS, which is going to be NDK r25 (not sure if GH Actions venv has that yet...) until at least 2023: https://github.com/android/ndk/wiki#release-schedule

kchibisov:
Couldn't that be added inside `env` section?

kchibisov (approved):
That's fine. I'm not into github actions anyway.

MarijnS95:
For reference, looks like we can back out of this soon: https://github.com/actions/virtual-environments/issues/5879#issuecomment-1197811704

PR #2400, opened 2022-07-27
Explicitly specify minimum supported rust version

### LINKED ISSUE #1075: Minimum Supported Rust Version
Back in https://github.com/rust-windowing/winit/issues/531 it seems that it was agreed to test against `1.24` in CI, and now [travis](https://github.com/rust-windowing/winit/blob/master/.travis.yml) and [appveyor](https://github.com/rust-windowing/winit/blob/master/appveyor.yml) just test against `stable`.

It would be better for the long term health of the community if foundational crates such as `winit` had a particular version number for minimum rust requirement. Could we move `winit` back in that direction? Since everyone is working on `winit-0.20` being a "big change" with event loop 2.0 and all that, it'd be nice to also use that as a time to transition back to a policy of having a specific minimum rust version and trying to be aware of when that minimum version goes up.

kchibisov:
This should help with distributing apps using winit.

Fixes #1075.

--

Will require a patch bump afterwards unfortunately.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        rust_version: [stable, nightly]
+        rust_version: [1.57.0, stable, nightly]
         platform:
           # Note: Make sure that we test all the `docs.rs` targets defined in Cargo.toml!
           - { target: x86_64-pc-windows-msvc,   os: windows-latest,  }
@@ -103,7 +103,7 @@ jobs:
 
     - name: Lint with clippy
       shell: bash
-      if: (matrix.rust_version != 'nightly') && !contains(matrix.platform.options, '--no-default-features')
+      if: (matrix.rust_version == '1.57.0') && !contains(matrix.platform.options, '--no-default-features')
       run: cargo clippy --all-targets --target ${{ matrix.platform.target }} $OPTIONS --features $FEATURES -- -Dwarnings
 
     - name: Build with serde enabled
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,6 +20,7 @@ your description of the issue as detailed as possible:
 
 When making a code contribution to winit, before opening your pull request, please make sure that:
 
+- your patch builds with Winit's minimal supported rust version - Rust 1.57.0.
 - you tested your modifications on all the platforms impacted, or if not possible detail which platforms
   were not tested, and what should be tested, so that a maintainer or another contributor can test them
 - you updated any relevant documentation in winit
--- a/src/platform/unix.rs
+++ b/src/platform/unix.rs
@@ -299,10 +299,11 @@ impl WindowExtUnix for Window {
     #[inline]
     #[cfg(feature = "wayland")]
     fn wayland_set_csd_theme(&self, theme: Theme) {
+        #[allow(clippy::single_match)]
         match self.window {
             LinuxWindow::Wayland(ref w) => w.set_csd_theme(theme),
             #[cfg(feature = "x11")]
-            _ => {}
+            _ => (),
         }
     }
 
--- a/src/platform_impl/ios/app_state.rs
+++ b/src/platform_impl/ios/app_state.rs
@@ -793,7 +793,7 @@ pub unsafe fn handle_main_events_cleared() {
         return;
     }
     match this.state_mut() {
-        &mut AppStateImpl::ProcessingEvents { .. } => {}
+        AppStateImpl::ProcessingEvents { .. } => {}
         _ => bug!("`ProcessingRedraws` happened unexpectedly"),
     };
     drop(this);
--- a/src/platform_impl/linux/mod.rs
+++ b/src/platform_impl/linux/mod.rs
@@ -587,7 +587,8 @@ impl Window {
 
 /// Hooks for X11 errors.
 #[cfg(feature = "x11")]
-pub(crate) static mut XLIB_ERROR_HOOKS: Mutex<Vec<XlibErrorHook>> = Mutex::new(Vec::new());
+pub(crate) static mut XLIB_ERROR_HOOKS: Lazy<Mutex<Vec<XlibErrorHook>>> =
+    Lazy::new(|| Mutex::new(Vec::new()));
 
 #[cfg(feature = "x11")]
 unsafe extern "C" fn x_error_callback(
@@ -633,7 +634,7 @@ unsafe extern "C" fn x_error_callback(
 
 pub enum EventLoop<T: 'static> {
     #[cfg(feature = "wayland")]
-    Wayland(wayland::EventLoop<T>),
+    Wayland(Box<wayland::EventLoop<T>>),
     #[cfg(feature = "x11")]
     X(x11::EventLoop<T>),
 }
@@ -723,7 +724,7 @@ impl<T: 'static> EventLoop<T> {
 
     #[cfg(feature = "wayland")]
     fn new_wayland_any_thread() -> Result<EventLoop<T>, Box<dyn Error>> {
-        wayland::EventLoop::new().map(EventLoop::Wayland)
+        wayland::EventLoop::new().map(|evlp| EventLoop::Wayland(Box::new(evlp)))
     }
 
     #[cfg(feature = "x11")]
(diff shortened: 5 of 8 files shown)


madsmtm (commented):
I don't mind testing some minimum Rust version in CI (as long as we don't start saying "no, I won't accept this feature because it bumps the MSRV").

Note that we can't guarantee this for downstream users without _all_ of our dependencies also setting the field to the same version or lower!

dhardy:
Want a badge in the README?
```
![Minimum rustc version](https://img.shields.io/badge/rustc-1.57-lightgray.svg)
```

kchibisov:
>Want a badge in the README?

I don't like them, and we have that stated in `Cargo.toml` anyway.

>Note that we can't guarantee this for downstream users without all of our dependencies also setting the field to the same version or lower!

We actually do guarantee, given that CI builds winit now with MSRV and it won't pass otherwise.

madsmtm:
> We actually do guarantee, given that CI builds winit now with MSRV and it won't pass otherwise.

True, but that only says something about a version of Rust that `winit` currently builds with right now, not what it is guaranteed to build with in the future.

As an example, let's take a hypothetical `winit v0.27.1`, which has the semver guarantee "builds with Rust 1.57". This version depends on the crate `instant v0.1`, which doesn't specify a `rust-version`, but it builds on Rust 1.57, so all is fine.
Later on (could theoretically be years from now), the `instant` crate is improved, using some new Rust > 1.57 feature, and the `instant` crate includes this in a patch version, which they can since they don't provide any guarantees about their MSRV.
Aaand suddenly the guarantee that `winit v0.27.1` builds with Rust 1.57 no longer holds!

madsmtm:
My point is, I'm fine with stating what our current MSRV is, but we must take care not to make it part of our SemVer contract, since we can't uphold such a guarantee.

kchibisov:
Yes, but applications can downgrade the dependencies, that's the point. In general version change is a breaking change so library writers should have some sanity.

madsmtm:
> Yes, but applications can downgrade the dependencies, that's the point.

I _think_ I understand the use-case better now, so we just state "there exists dependencies that can satisfy this specific version of `winit`, and can compile on Rust 1.57 - but you'll have to go looking for them yourself, since they may have been updated since".
Distributions can then use this information to say "we can safely specify a loose `winit = "0.27"` requirement, since a bump to `0.27.2` would not increase MSRV", while they'd still have to pin e.g. the `instant = "=0.1.12"` crate since that crate doesn't make any such guarantees.

Is that the idea? Or is there another use-case for distributions that I'm missing?

maroider (commented):
Won't our CI break if one of our dependencies bump their MSRV in a patch release (or minor release for post-`1.0` versions)?

kchibisov:
Yes. And that we can ship alacritty to fedora/ubuntu since those update rust not that often.

kchibisov:
>Won't our CI break if one of our dependencies bumps their MSRV in a patch release (or minor release for post-1.0 versions)?

I'm not entire sure, but we'll see how it works.

In general libraries must not break msrv in a not breaking change without a reason. At some point we can make all of them behave.

I've started that since it broke the ability to ship alacritty to any non rolling release distro, and I haven't noticed it due to not having automatic tests for msrv, etc. (But I was very happy having `Mutex` in `static` without any `Lazy` craft).

CONTRIBUTING.md
```diff
@@ -20,6 +20,7 @@ your description of the issue as detailed as possible:
 
 When making a code contribution to winit, before opening your pull request, please make sure that:
 
+- your patch builds with Winit's minimal supported rust version - Rust 1.57.0.
```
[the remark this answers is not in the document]
kchibisov: It's fine. I don't really like that we have so many checks in template, I tend to not look at them sometimes.

maroider:
> I've started that since it broke the ability to ship alacritty to any non rolling release distro, and I haven't noticed it due to not having automatic tests for msrv, etc.

Do we have some sort of list of distros we care about in this regard, since I assume we then want to keep ourselves compatible with the `rustc` versions they ship?

kchibisov:
>Do we have some sort of list of distros we care about in this regard, since I assume we then want to keep ourselves compatible with the rustc versions they ship?

In general I'd suggest to target latest stable ubuntu rustc version. Every other distro is likely having newer version.

kchibisov:
In general, what would be the resolution here? I think that we can go ahead and keep it that way and see how often it'll break CI for us, but it should be pretty rare case.

CONTRIBUTING.md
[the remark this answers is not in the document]
MarijnS95: Me neither, especially things that are caught by the CI are simply wasteful to have in the checklist. Contributors will likely see it and patch up their PR to satisfy the CI. Anything we can automate is a win IMO.

MarijnS95:
@kchibisov I'd just go ahead with this. Any crate that breaks us / our CI at some point should simply result in a discussion whether the crate in question is at fault (and should realistically be yanked **if** MSRV is considered part of "breaking version bump" semver-wise), or our view of MSRV should be bumped and published. At the very least it won't hurt testing for it now.

.github/workflows/ci.yml
```diff
@@ -103,7 +103,7 @@ jobs:
 
     - name: Lint with clippy
       shell: bash
-      if: (matrix.rust_version != 'nightly') && !contains(matrix.platform.options, '--no-default-features')
+      if: (matrix.rust_version == '1.57.0') && !contains(matrix.platform.options, '--no-default-features')
       run: cargo clippy --all-targets --target ${{ matrix.platform.target }} $OPTIONS --features $FEATURES -- -Dwarnings
```
MarijnS95, line 107: I feel like we should also run `clippy` on latest at least, it typically has a few new lints that we might want to clean up. MSRV should just be an "it builds" kind of test (also across all our platforms).

src/platform/unix.rs
```diff
@@ -299,10 +299,11 @@ impl WindowExtUnix for Window {
     #[inline]
     #[cfg(feature = "wayland")]
     fn wayland_set_csd_theme(&self, theme: Theme) {
+        #[allow(clippy::single_match)]
```
MarijnS95, line 302: Don't the other functions above need this too, which have a seemingly identical `match` block?

src/platform_impl/ios/app_state.rs
```diff
@@ -792,8 +792,8 @@ pub unsafe fn handle_main_events_cleared() {
     if !this.has_launched() {
         return;
     }
-    match this.state_mut() {
-        &mut AppStateImpl::ProcessingEvents { .. } => {}
+    match *this.state_mut() {
+        AppStateImpl::ProcessingEvents { .. } => {}
```
MarijnS95, line 796: Don't think you need either of these, thanks to auto-ref these days?

src/platform_impl/linux/mod.rs
```diff
@@ -633,7 +634,7 @@ unsafe extern "C" fn x_error_callback(
 
 pub enum EventLoop<T: 'static> {
     #[cfg(feature = "wayland")]
-    Wayland(wayland::EventLoop<T>),
+    Wayland(Box<wayland::EventLoop<T>>),
```
MarijnS95, line 637: What is this box about?

.github/workflows/ci.yml
kchibisov: The problem is that those are mutually exclusive. So passing clippy run on stable, could result in error on old stable, and vice-versa.

src/platform/unix.rs
kchibisov: No, because they return, but this does not.

src/platform_impl/ios/app_state.rs
kchibisov: I'm not sure, there must be a reason why clippy suggested it.

src/platform_impl/linux/mod.rs
kchibisov: It complains that the size of the variant is too large.

ajtribick:
Looks like the MSRV imposed by dependencies is now 1.59, due to the MSRV of `time` (imported via `simple_logger`) changing in version 0.3.14 (i.e. they consider it a non-breaking change).

PR #2404, opened 2022-07-30
On X11, fix crash when can't disable ime

### LINKED ISSUE #2402: init fails on x11 (winit 0.27)
On the latest git master and in the recently released 0.27, when I run the window example via: `cargo run --release --example window`, it panics with:
image

0.26 did not have this issue

kchibisov:
Fixes #2402.


rukai:
Init now completes successfully and no longer panics:
image

PR #2409, opened 2022-08-05
Add Windows opt-in function for device events

ajtribick:
   - [ ] Windows <8.1
   - [ ] Windows 8.1
   - [ ] Windows 10 (<21H2)
   - [x] Windows 10 (21H2)
   - [ ] Windows 11
- [x] Created or updated an example program if it would help users understand this functionality (N/A - already included in `window_debug.rs`)
- [x] Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented (N/A)

Addresses (#1634) on Windows.

An attempt at doing for Windows what #2195 did for X11. Device events are by default only listened to on the focused window and need to be manually opted into via `EventLoop::set_device_event_filter`.

This appears to work in a very simple test application on Windows 10 21H2. I didn't see a list of supported Windows versions, in any case I'm not able to test other ones.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 # Unreleased
 
 - On Windows, respect min/max inner sizes when creating the window.
+- **Breaking:** On Windows, device events are now ignored for unfocused windows by default, use `EventLoopWindowTarget::set_device_event_filter` to set the filter level.
 
 # 0.27.1 (2022-07-30)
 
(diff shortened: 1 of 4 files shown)


maroider:
> I didn't see a list of supported Windows versions, in any case I'm not able to test other ones.

We don't really have a list of formally supported Windows versions, and since the last person who was able to test Windows 7 had their setup somehow rendered inoperable, we can only really test Windows 10 and up. I suppose this is a topic worth discussing on its own, but that's probably something best done in a dedicated issue.

msiglreith (approved):

Issue #2020, opened 2021-09-04
Support updating resize increments


necauqua:
As an answer to #135, `WindowBuilderExt::with_resize_increments` was added and implemented where WM supports it (X11/MacOS).

However this is limited, since it only allows you to set those increments at window creation and you cannot change them later.

The example from the first issue still holds, to add resize increments to Alacritty (that still was not done so far, and which I've tried to do myself) you need to set the increments for existing window dynamically - so when the glyph size changes (because user changed the font in the config and it was reloaded) the resize increments can be fixed as well.

As far as I understood both on X11 and on MacOS nothing prevents you from setting the increments for the existing window (and continuing with the example, other terminal emulators are doing just that) any number of times.

maroider:
Feel free to send a PR our way if you're up to implementing this change.

From the documentation pages linked in #135, it does indeed seem entirely reasonable to dynamically alter the resize increments.

PR #2410, opened 2022-08-07
x11: Fix size hints being reset after a DPI change

necauqua:
I actually wanted to start working on #2020 almost a year after I created it, but in the process found this bug that when you move the window to another screen resize increments go poof

### DIFF
--- a/src/platform_impl/linux/x11/window.rs
+++ b/src/platform_impl/linux/x11/window.rs
@@ -371,14 +371,14 @@ impl UnownedWindow {
                     } else {
                         max_inner_size = Some(dimensions.into());
                         min_inner_size = Some(dimensions.into());
-
-                        let mut shared_state = window.shared_state.get_mut();
-                        shared_state.min_inner_size = window_attrs.min_inner_size;
-                        shared_state.max_inner_size = window_attrs.max_inner_size;
-                        shared_state.resize_increments = pl_attribs.resize_increments;
-                        shared_state.base_size = pl_attribs.base_size;
                     }
                 }
+
+                let mut shared_state = window.shared_state.get_mut();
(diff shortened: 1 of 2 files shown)


src/platform_impl/linux/x11/window.rs
```diff
@@ -371,15 +371,15 @@ impl UnownedWindow {
...
-
-                        let mut shared_state = window.shared_state.get_mut();
-                        shared_state.min_inner_size = window_attrs.min_inner_size;
-                        shared_state.max_inner_size = window_attrs.max_inner_size;
-                        shared_state.resize_increments = pl_attribs.resize_increments;
-                        shared_state.base_size = pl_attribs.base_size;
                     }
                 }
 
+                let mut shared_state = window.shared_state.get_mut();
```
kchibisov, line 377: I'm not quite sure how commit message refers to dpi change in the first place. From what I can see they were never called in the first place for a resizable window, which is a bug.

So you must state that in `CHANGELOG.md` and commit message, since what you're telling fixing is not what you've actually fixed.

Something like `On X11, fix min, max, and resize increments not being applied for resizable window.`

kchibisov (changes requested):

src/platform_impl/linux/x11/window.rs
kchibisov: Given how bad our X11 backend is I think I got why it works sometimes now. It basically sets the attributes but not keep them since the window is resizable.

`On X11, fix min, max, and resize increments not persist for resizable window.`

That is a bit more clear, I think.

necauqua: Done

kchibisov (approved):

Issue #2308, opened 2022-05-29
Expose window 'surface' size, distinct from the 'inner' size


rib:
*Edited: to use "surface size" instead of "physical size" as discussed. The edit also tries to clarify the table information.*

Currently the general, go-to API for querying the size of a winit window is `.inner_size()` which has conflicting requirements due to some downstream consumers wanting to know the size to create render surfaces and other downstream consumers wanting to know the safe, inner bounds of content (which could exist within a potentially larger surface).

For lower-level rendering needs, such as for supporting integration with wgpu and or directly with opengl / vulkan etc then what downstream wants to know is the size of the surface they should create, which effectively determines how much memory to allocate for a render target and the width and height of that render target.

_Incidentally, 'physical size' is how the Bevy engine refers to the size it tracks, based on reading the winit `.inner_size()` which is a good example where it's quite clear what semantics they are primarily looking for (since they will pass the size to wgpu to configure render surfaces). In this case Bevy is not conceptually interested in knowing about insets for things like frame decorations or mobile OS safe areas._

Conceptually the `inner_size()`  is 'inner' with respect to the 'outer' size, and the `outer_size()` is primarily applicable to desktop window systems which may have window frames that extend outside the content area of the applications window, and may also be larger than the surface size that's used by the application for rendering. For example on X11 the inner size will technically relate to a separate smaller child window that's parented by the window manager onto a larger frame window.

_Incidentally on Wayland which was designed to try and encourage client-side window decorations and also sandbox clients the core protocol doesn't let you introspect frame details or an outer screen position._

Here's a matrix to try and summarize the varying semantics for the inner/outer and surface sizes across window systems to help show why I think it would be useful to expose the physical size explicitly, to decouple it from the inner_size:

| Window System | Outer Size | Inner Size | Surface Size |
|---|---|---|--|
| X11 | Size of parent frame window owned by the window manager (e.g. with title bar and border around content)  | Size of nested content window (which is usually a child of the frame window) </br></br>This may vary but I want to highlight that on X11 due to it's async nature there are times where it makes sense to optimistically refer to a pending/requested size (so e.g. `.set_inner_size()` followed by `.inner_size()` can return a pending size) but the 'surface' size would be the remote, server-side size | Size of nested content window (which is usually a child of the frame window) </br></br>In some situations the driver may need to reconcile mismatched surface/window sizes due to async resizing of window, e.g. using a top-left gravity and no scaling. | 
[...]

Issue #2306, opened 2022-05-28
Counterintuitive resizability behaviors [on x11]


64kramsystem:
I'm using `winit` (0.26.x) as underlying library of a game, which I'm developing on Linux/x11. On startup, I need to resize the window, and prevent the user from resizing it.

Now, there are two problems:

1. `Window#set_resizable(false)` prevents also programmatic changes (in other words, it will cause subsequent calls to `set_inner_size()` won't have any effect); this is unexpected - I'd expect only user resizing to be prevented, not also programmatic one.

2. Invoking `set_inner_size()` and immediately after `set_resizable(false)`, will cause resizing to be disabled _without_ the window size to be changed; in order to change the size, a pause needs to be taken between the two statements (e.g. stepping through in a debugging session); I actually worked this around by putting `set_resizable()` in a resizing event callback.

It's not clear to me if the behaviors above are intended or they're bugs. If they're intended, I think they should be documented, as they're not obvious (I'd even classify them as counterintuitive).

rib:
I agree that intuitively I'd only expect `set_resizable(false)` to request that the window manager should block the user from resizing the window, and I think that's the intent of the winit API.

The X11 backend does try to specifically account for `set_inner_size()` being called after `set_resizable(false)`, so I don't think the behaviour you're currently seeing is intentional.

This looks like a race condition due to `set_inner_size()` not caching the set size locally as a "pending" inner size that can be referenced by `set_resizable()`. On X11 the winit window's inner size only changes asynchronously, depending on whether the window manager decides to honor the request that's sent when `set_inner_size()` is called. Ideally though `set_resizable()` should be able to see that there is a "pending" (or something similar) inner size that has been requested and when it tries to disable resizing by setting the min/max to be the same in WM_NORMAL_HINTS it should request the pending size, not the
last size that was notified by the server.

For example the order of events can currently look like this:

| timeline of events/actions within winit           | winit's inner_size |
|---------------------------|---------------------|
| \<initial state\>                                               | 200x200|
| set_inner_size(300x300)                           | 200x200|
|  --- Send Resize Window request (300x300)    (i.e. just send async request)            | 200x200       |
|set_resizable(false)                                  | 200x200|
| --- read inner_size = 200x200               | 200x200|
| --- set WM_NORMAL_HINTS min/max = 200x200 | 200x200|
[...]

PR #2411, opened 2022-08-07
Allow changing resize increments after window creation

### LINKED ISSUE #2020: Support updating resize increments

necauqua:
Pretty basic "I hope it works" implementation, only tested on X11, can test on a mac in a few days.

Once this is reviewed->merged->released I hope to finally make Alacritty _not suck_, heh

~The branch is based on the #2410 because I kept moving the test window to the other screen, I'll rebase once it's merged,~ done

~Currently this is a draft to run your CI and to get reviews.~ should be ready

Closes #2020

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- Migrated `WindowBuilderExtUnix::with_resize_increments` to `WindowBuilder`.
+- Added `Window::resize_increments`/`Window::set_resize_increments` to update resize increments at runtime for X11/macOS.
 - macOS/iOS: Use `objc2` instead of `objc` internally.
 - **Breaking:** Bump MSRV from `1.57` to `1.60`.
 - **Breaking:** Split the `platform::unix` module into `platform::x11` and `platform::wayland`. The extension types are similarly renamed.
--- a/examples/window_resize_increments.rs
+++ b/examples/window_resize_increments.rs
@@ -1,0 +1,57 @@
+use log::debug;
+use simple_logger::SimpleLogger;
+use winit::{
+    dpi::LogicalSize,
+    event::{ElementState, Event, KeyboardInput, VirtualKeyCode, WindowEvent},
+    event_loop::EventLoop,
+    window::WindowBuilder,
+};
+
+fn main() {
+    SimpleLogger::new().init().unwrap();
+    let event_loop = EventLoop::new();
+
+    let window = WindowBuilder::new()
+        .with_title("A fantastic window!")
+        .with_inner_size(LogicalSize::new(128.0, 128.0))
+        .with_resize_increments(LogicalSize::new(25.0, 25.0))
+        .build(&event_loop)
+        .unwrap();
+
+    let mut has_increments = true;
+
+    event_loop.run(move |event, _, control_flow| {
+        control_flow.set_wait();
+
+        match event {
+            Event::WindowEvent {
+                event: WindowEvent::CloseRequested,
+                window_id,
+            } if window_id == window.id() => control_flow.set_exit(),
+            Event::WindowEvent {
+                event:
+                    WindowEvent::KeyboardInput {
+                        input:
+                            KeyboardInput {
+                                virtual_keycode: Some(VirtualKeyCode::Space),
+                                state: ElementState::Released,
+                                ..
+                            },
+                        ..
+                    },
+                window_id,
+            } if window_id == window.id() => {
+                has_increments = !has_increments;
+
+                let new_increments = match window.resize_increments() {
+                    Some(_) => None,
+                    None => Some(LogicalSize::new(25.0, 25.0)),
+                };
+                debug!("Had increments: {}", new_increments.is_none());
+                window.set_resize_increments(new_increments);
+            }
+            Event::MainEventsCleared => window.request_redraw(),
+            _ => (),
+        }
+    });
+}
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -757,6 +757,12 @@ impl Window {
 
     pub fn set_max_inner_size(&self, _: Option<Size>) {}
 
+    pub fn resize_increments(&self) -> Option<PhysicalSize<u32>> {
+        None
+    }
+
+    pub fn set_resize_increments(&self, _increments: Option<Size>) {}
+
     pub fn set_title(&self, _title: &str) {}
 
     pub fn set_visible(&self, _visibility: bool) {}
--- a/src/platform_impl/linux/mod.rs
+++ b/src/platform_impl/linux/mod.rs
@@ -95,8 +95,6 @@ pub struct PlatformSpecificWindowBuilderAttributes {
     #[cfg(feature = "x11")]
     pub screen_id: Option<i32>,
     #[cfg(feature = "x11")]
-    pub resize_increments: Option<Size>,
-    #[cfg(feature = "x11")]
     pub base_size: Option<Size>,
     #[cfg(feature = "x11")]
     pub override_redirect: bool,
@@ -116,8 +114,6 @@ impl Default for PlatformSpecificWindowBuilderAttributes {
             visual_infos: None,
             #[cfg(feature = "x11")]
             screen_id: None,
-            #[cfg(feature = "x11")]
-            resize_increments: None,
             #[cfg(feature = "x11")]
             base_size: None,
             #[cfg(feature = "x11")]
@@ -389,6 +385,16 @@ impl Window {
     #[inline]
     pub fn set_max_inner_size(&self, dimensions: Option<Size>) {
         x11_or_wayland!(match self; Window(w) => w.set_max_inner_size(dimensions))
+    }
+
+    #[inline]
+    pub fn resize_increments(&self) -> Option<PhysicalSize<u32>> {
+        x11_or_wayland!(match self; Window(w) => w.resize_increments())
+    }
+
+    #[inline]
+    pub fn set_resize_increments(&self, increments: Option<Size>) {
+        x11_or_wayland!(match self; Window(w) => w.set_resize_increments(increments))
     }
 
     #[inline]
--- a/src/platform_impl/linux/x11/util/hint.rs
+++ b/src/platform_impl/linux/x11/util/hint.rs
@@ -197,11 +197,17 @@ impl<'a> NormalHints<'a> {
     }
 
     pub fn get_position(&self) -> Option<(i32, i32)> {
-        if has_flag(self.size_hints.flags, ffi::PPosition) {
-            Some((self.size_hints.x as i32, self.size_hints.y as i32))
-        } else {
-            None
-        }
+        has_flag(self.size_hints.flags, ffi::PPosition)
+            .then(|| (self.size_hints.x as i32, self.size_hints.y as i32))
+    }
+
+    pub fn get_resize_increments(&self) -> Option<(u32, u32)> {
+        has_flag(self.size_hints.flags, ffi::PResizeInc).then(|| {
+            (
+                self.size_hints.width_inc as u32,
+                self.size_hints.height_inc as u32,
+            )
+        })
     }
 
     pub fn set_position(&mut self, position: Option<(i32, i32)>) {
--- a/src/platform_impl/linux/x11/window.rs
+++ b/src/platform_impl/linux/x11/window.rs
@@ -378,7 +378,7 @@ impl UnownedWindow {
                 let mut shared_state = window.shared_state.get_mut().unwrap();
                 shared_state.min_inner_size = min_inner_size.map(Into::into);
                 shared_state.max_inner_size = max_inner_size.map(Into::into);
-                shared_state.resize_increments = pl_attribs.resize_increments;
+                shared_state.resize_increments = window_attrs.resize_increments;
                 shared_state.base_size = pl_attribs.base_size;
 
                 let mut normal_hints = util::NormalHints::new(xconn);
@@ -387,7 +387,7 @@ impl UnownedWindow {
                 normal_hints.set_min_size(min_inner_size.map(Into::into));
                 normal_hints.set_max_size(max_inner_size.map(Into::into));
                 normal_hints.set_resize_increments(
-                    pl_attribs
+                    window_attrs
                         .resize_increments
                         .map(|size| size.to_physical::<u32>(scale_factor).into()),
                 );
@@ -1170,6 +1170,24 @@ impl UnownedWindow {
         let physical_dimensions =
             dimensions.map(|dimensions| dimensions.to_physical::<u32>(self.scale_factor()).into());
         self.set_max_inner_size_physical(physical_dimensions);
+    }
+
+    #[inline]
+    pub fn resize_increments(&self) -> Option<PhysicalSize<u32>> {
+        self.xconn
+            .get_normal_hints(self.xwindow)
+            .ok()
+            .and_then(|hints| hints.get_resize_increments())
+            .map(Into::into)
+    }
+
+    #[inline]
+    pub fn set_resize_increments(&self, increments: Option<Size>) {
+        self.shared_state_lock().resize_increments = increments;
+        let physical_increments =
+            increments.map(|increments| increments.to_physical::<u32>(self.scale_factor()).into());
+        self.update_normal_hints(|hints| hints.set_resize_increments(physical_increments))
+            .expect("Failed to call `XSetWMNormalHints`");
     }
 
     pub(crate) fn adjust_for_dpi(
--- a/src/platform_impl/macos/window.rs
+++ b/src/platform_impl/macos/window.rs
@@ -242,10 +242,14 @@ fn create_window(
             }
 
             if let Some(increments) = pl_attrs.resize_increments {
-                let (x, y) = (increments.width, increments.height);
-                if x >= 1.0 && y >= 1.0 {
-                    let size = NSSize::new(x as CGFloat, y as CGFloat);
-                    ns_window.setResizeIncrements_(size);
+                let (w, h) = (increments.width, increments.height);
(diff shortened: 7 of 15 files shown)


necauqua:
Still marked as a draft because I did not test this on a mac yet, if you have a mac and are willing to test this you are welcome

kchibisov:
I could test macOs, but not soon. Maybe @madsmtm could check.

examples/window_resize_increments.rs
```diff
@@ -0,0 +1,68 @@
+#[cfg(any(target_os = "macos", all(target_os = "linux", feature = "x11")))]
```
kchibisov, line 1: This is all redundant. Given that Api is there for all the platform.

examples/window_resize_increments.rs
```diff
@@ -0,0 +1,68 @@
...
+    #[cfg(all(target_os = "linux", feature = "x11"))]
+    use winit::platform::unix::WindowBuilderExtUnix;
+
+    SimpleLogger::new().init().unwrap();
+    let event_loop = EventLoop::new();
+
+    let window = WindowBuilder::new()
+        .with_title("A fantastic window!")
+        .with_inner_size(LogicalSize::new(128.0, 128.0))
+        .with_resize_increments(LogicalSize::new(25.0, 25.0))
```
kchibisov, line 23: Could you move this method on `WindowBuilder`? I don't think we should extension gate the builder call, given that we have method for all the platforms now.

examples/window_resize_increments.rs
```diff
@@ -0,0 +1,68 @@
...
+                                virtual_keycode: Some(VirtualKeyCode::Space),
+                                state: ElementState::Released,
+                                ..
+                            },
+                        ..
+                    },
+                window_id,
+            } if window_id == window.id() => {
+                has_increments = !has_increments;
+                log::debug!("Has increments: {}", has_increments);
```
kchibisov, line 51: Could just `use log::debug`.

examples/window_resize_increments.rs
```diff
@@ -0,0 +1,68 @@
...
+                };
+                window.set_resize_increments(increments);
+            }
+            Event::MainEventsCleared => window.request_redraw(),
+            _ => (),
+        }
+    });
+}
+
+#[cfg(not(any(target_os = "macos", all(target_os = "linux", feature = "x11"))))]
```
kchibisov, line 65: No need for that.

src/platform_impl/android/mod.rs
```diff
@@ -757,6 +757,12 @@ impl Window {
 
     pub fn set_max_inner_size(&self, _: Option<Size>) {}
 
+    pub fn resize_increments(&self) -> Option<PhysicalSize<u32>> {
+        None
+    }
+
+    pub fn set_resize_increments(&self, _increments: Option<Size>) {}
```
kchibisov, line 764: Could you add `warn!("Setting resize increments is not supported")`? Though, look at the similar messages we are using, might need lowercase it.

src/platform_impl/linux/mod.rs
```diff
@@ -388,6 +388,26 @@ impl Window {
         x11_or_wayland!(match self; Window(w) => w.set_max_inner_size(dimensions))
     }
 
+    #[inline]
+    pub fn resize_increments(&self) -> Option<PhysicalSize<u32>> {
+        match self {
+            #[cfg(feature = "x11")]
+            Window::X(ref w) => w.resize_increments(),
+            #[cfg(feature = "wayland")]
+            Window::Wayland(_) => None,
```
kchibisov, line 397: Could you add direct calls to Wayland? I'll add support for it later on, so we can at least has a warning for now.

src/platform_impl/linux/mod.rs
```diff
@@ -388,6 +388,26 @@ impl Window {
...
+        }
+    }
+
+    #[inline]
+    pub fn set_resize_increments(&self, _increments: Option<Size>) {
+        match self {
+            #[cfg(feature = "x11")]
+            Window::X(ref w) => w.set_resize_increments(_increments),
+            #[cfg(feature = "wayland")]
+            Window::Wayland(_) => (),
```
kchibisov, line 407: ditto.

src/platform_impl/linux/x11/util/hint.rs
```diff
@@ -245,6 +245,17 @@ impl<'a> NormalHints<'a> {
         }
     }
 
+    pub fn get_resize_increments(&self) -> Option<(u32, u32)> {
+        if has_flag(self.size_hints.flags, ffi::PResizeInc) {
```
kchibisov, line 249: Could use `then`?

src/platform_impl/web/window.rs
```diff
@@ -153,6 +153,16 @@ impl Window {
...
     }
 
+    #[inline]
+    pub fn resize_increments(&self) -> Option<PhysicalSize<u32>> {
+        None
+    }
+
+    #[inline]
+    pub fn set_resize_increments(&self, _increments: Option<Size>) {
+        // Intentionally a no-op: users can't resize canvas elements
```
kchibisov, line 163: ditto about warning in log.

src/platform_impl/windows/window.rs
```diff
@@ -230,6 +230,14 @@ impl Window {
         self.set_inner_size(size.into());
     }
 
+    #[inline]
+    pub fn resize_increments(&self) -> Option<PhysicalSize<u32>> {
+        None
+    }
+
+    #[inline]
+    pub fn set_resize_increments(&self, _increments: Option<Size>) {}
```
kchibisov, line 239: ditto about warning.

src/window.rs
```diff
@@ -604,6 +604,29 @@ impl Window {
...
+        self.window.resize_increments()
+    }
+
+    /// Sets window resize increments.
+    /// Only has an effect on X11 and macOS.
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **macOS:** Increments are converted to logical size and then macOS rounds them to whole numbers.
+    /// - **iOS / Android / Web / Wayland / Windows:** Unsupported.
```
kchibisov, line 624: Could you specify that `Wayland / Windows` are not implemented, since you can do something implement the API on those as well.

src/window.rs
```diff
@@ -604,6 +604,29 @@ impl Window {
...
+    /// ## Platform-specific
+    ///
+    /// - **iOS / Android / Web / Wayland / Windows:** Always returns [`None`].
+    #[inline]
+    pub fn resize_increments(&self) -> Option<PhysicalSize<u32>> {
+        self.window.resize_increments()
+    }
+
+    /// Sets window resize increments.
+    /// Only has an effect on X11 and macOS.
```
kchibisov, line 619: You don't need this note about X11/macOs given that we have a platform specific section for them already.

kchibisov (changes requested):

src/platform_impl/macos/window.rs
```diff
@@ -605,6 +605,28 @@ impl UnownedWindow {
...
+        } else {
+            None
+        }
+    }
+
+    pub fn set_resize_increments(&self, increments: Option<Size>) {
+        let size = increments
+            .map(|increments| {
+                let logical = increments.to_logical::<f64>(self.scale_factor());
+                NSSize::new(logical.width.min(1.0), logical.height.min(1.0))
```
madsmtm, line 622: You probably want:
```suggestion
                NSSize::new(logical.width.max(1.0), logical.height.max(1.0))
```

Since right now, clearing the resize increments works, but setting it again doesn't.

examples/window_resize_increments.rs
```diff
@@ -0,0 +1,68 @@
...
+    let event_loop = EventLoop::new();
+
+    let window = WindowBuilder::new()
+        .with_title("A fantastic window!")
+        .with_inner_size(LogicalSize::new(128.0, 128.0))
+        .with_resize_increments(LogicalSize::new(25.0, 25.0))
+        .build(&event_loop)
+        .unwrap();
+
+    let mut has_increments = true;
```
madsmtm, line 27: Given that we have an API for accessing the resize increments, perhaps we could use that instead of the user needing to store a value here?

src/platform_impl/macos/window.rs
```diff
@@ -249,7 +249,7 @@ fn create_window(
                 let (x, y) = (increments.width, increments.height);
                 if x >= 1.0 && y >= 1.0 {
                     let size = NSSize::new(x as CGFloat, y as CGFloat);
-                    ns_window.setResizeIncrements_(size);
+                    ns_window.setContentResizeIncrements_(size);
```
madsmtm, line 252: Any particular reason you chose [`contentResizeIncrements`](https://developer.apple.com/documentation/appkit/nswindow/1419649-contentresizeincrements?language=objc) over [`resizeIncrements`](https://developer.apple.com/documentation/appkit/nswindow/1419390-resizeincrements?language=objc)?

I'm guessing because it's easier for your use-case?

Perhaps the methods should be called `inner_resize_increments` and `set_inner_resize_increments`? See also https://github.com/rust-windowing/winit/issues/2308#issuecomment-1197248011.

madsmtm (changes requested):
Thanks for the PR, looks good and seems to work fine barring a few things

src/platform_impl/macos/window.rs
rib: In the case of x11 specifically, it's really an "outer" size increment if you want to be more specific, since it's a hint to the window manager which willy typically handle resizing based on interaction with a separate parent frame that is created.

The distinction can even be important at times with x11 too because the outer size may change size out of sync with the inner size. To avoid artifacts window managers utilize a window "gravity" mechanism, and compositors try to hide the latency too. On a related note winit currently has some issues with how it handles this asynchronous behaviour on x11: https://github.com/rust-windowing/winit/issues/2306

rib: I'd maybe avoid being specific, with "outer" vs "inner" here since it may vary between window systems what the increments would be relative too.

madsmtm: I think `resizeIncrements` is "outer" while `contentResizeIncrements` is "inner", so consistency is an argument for sticking with `resizeIncrements`

necauqua: On my desktop (awesomewm) if I set window borders to be super-thick resize increments behaved like 'inner' - and also inner increments make way more sense for the only application of resize increments I can think of - terminal emulators.

That's why I changed it to inner for the macos builder - also I vaguely recall that when I tried to implement this directly in alacritty a year ago the inner ones worked better on mac (that I then had at hands) or something.

rib: actually, yeah I suppose there is the sense that they are logically related to content - considering things like terminals or line-based views are the typical reason to want any resize increments

rib: tbh I think I was also misremembering how the wmhints work for x11 - it's one thing that the window manager is going to drive window resizing based on interaction with the outer frame, but the constraint it is going to be applying while handling a resize is indeed going to be to based on the desired increments for the inner window (as you say @necauqua - terminal emulators are the classic use case)

examples/window_resize_increments.rs
necauqua: Ok that's a personal habit, you can always write `log::something!("msg")` without importing anything and semantically it looks nice. I'll import it here tho

src/platform_impl/web/window.rs
necauqua: In those new dummy methods for other platforms I replicated surrounding comments/warnings/whatever - so that they look consistent with everything else that's already there

src/platform_impl/android/mod.rs
necauqua: In those new dummy methods for other platforms I replicated surrounding comments/warnings/whatever - so that they look consistent with everything else that's already there

necauqua:
Ok so I guess @madsmtm tested this on a mac so I'm undrafting this.
A couple of remaining minor reviews waiting for clarification after my comment and contentResizeIncrements/resizeIncrements debate also waits for a quorum agreement I guess

src/window.rs
```diff
@@ -333,6 +335,23 @@ impl WindowBuilder {
         self
     }
 
+    /// Build window with resize increment hint. Only implemented on X11 and macOS.
```
madsmtm, line 338: Nit: Use `## Platform-specific` section

src/window.rs
```diff
@@ -604,6 +623,29 @@ impl Window {
...
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **iOS / Android / Web / Wayland / Windows:** Always returns [`None`].
+    #[inline]
+    pub fn resize_increments(&self) -> Option<PhysicalSize<u32>> {
+        self.window.resize_increments()
+    }
+
+    /// Sets window resize increments.
```
madsmtm, line 637: Please document use-case (terminal emulators)

src/platform_impl/macos/window.rs
madsmtm: Let's just go with `contentResizeIncrements`, maybe a short comment saying that we use that instead of `resizeIncrements` would be nice.

I think the name `resize_increments` is fine, it sounds like there's basically never a need for "outer" resize increments.

madsmtm (approved):
Works flawlessly on macOS now, thanks!

necauqua:
Hey @kchibisov have you had a chance to see my reply to your comments about logging on unsupported platforms?

Do you insist on adding those logs or does my reasoning make sense to keep it as I did it initially

kchibisov:
@necauqua the thing is that `log` was added after some point in time, so the platforms inherited `// not supported` comments and there wasn't anything to change that. In general we want to log, but for now we can leave it as is for consistency sake.

src/window.rs
kchibisov: Bump.

src/window.rs
kchibisov: Bump.

src/platform_impl/linux/x11/window.rs
```diff
@@ -386,7 +386,7 @@ impl UnownedWindow {
                 normal_hints.set_min_size(min_inner_size.map(Into::into));
                 normal_hints.set_max_size(max_inner_size.map(Into::into));
                 normal_hints.set_resize_increments(
-                    pl_attribs
+                    window_attrs
                         .resize_increments
                         .map(|size| size.to_physical::<u32>(scale_factor).into()),
```
kchibisov, line 391: There's no need for map given that function handles that or the scaling is wrong when calling the `set_resize_increments` here? Could you add a comment about that at least? I do understand that the code was like that before, but it would be nice to ensure that and remove extra convertion, since it looks confusing.

kchibisov (changes requested):

necauqua:
Ok then

Maybe later some separate effort should be put in to update all these comments with log warnings (I'd personally also add a feature to disable that? or better a special log target(s) for those 'not implemented/unsupported' warnings)

src/platform_impl/linux/x11/window.rs
necauqua: You might be confused by the two lines above which look like `normal_hints.set_max_size(max_inner_size.map(Into::into));` - but in fact I modelled `resize_incremets` exactly like min/max_inner_size.

In those lines above e.g. the `min_inner_size` is is defined as `let mut min_inner_size = window_attrs.min_inner_size.map(|size| size.to_physical::<u32>(scale_factor))`, it just needs to be a mut local to update it for the ability to make windows unresizable.

src/window.rs
necauqua: I referenced the setter instead, like every other `with_` does

necauqua:
oh wow I dont think I've anything to do with those fails

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- Added `Window::resize_increments`/`Window::set_resize_increments` to update resize increments at runtime for X11/macOS; Migrated `WindowBuilderExtUnix::with_resize_increments` to `WindowBuilder`.
```
kchibisov, line 11: Could you split this line in two? No need to put two changes on the one line.

src/window.rs
```diff
@@ -333,6 +335,17 @@ impl WindowBuilder {
         self
     }
 
+    /// Build window with resize increment hint.
```
kchibisov, line 338:
```suggestion
    /// Build window with resize increments hint.
```

src/platform_impl/linux/x11/window.rs
kchibisov: I'm not saying that the logic is wrong or anything. I'm trying to say that `set_resize_increments` already does conversion to physical pixels with the `scale_factor` so the logic seems redundant and you can pass `resize_increments` directly?

There's a chance that I don't understand this code or scale factor isn't there.

kchibisov (changes requested):

src/platform_impl/linux/x11/window.rs
necauqua: Here we call `normal_hints.set_resize_increments` which just accepts a pair of i32s, physical, not `window.set_resize_increments` which accepts the `Size` enum and does the conversion - so that on window creation there is only one `xconn.set_normal_hints` call to set all the initial hints at once.

kchibisov (approved):

kchibisov:
Could you rebase. Should be good after that.

PR #2416, opened 2022-08-10
Note the platform versions that `winit` supports

madsmtm:
Currently only macOS and Windows, since I somewhat know how versioning works there. Missing iOS and Android! Would it make sense to do a similar thing for Wayland and X11?

I think we should do something similar for WASM, but unsure how, maybe it should be as verbose as specifying each browser an the version that it is expected to work on?

Spawned by https://github.com/rust-windowing/winit/pull/2409#issuecomment-1207266938.

### DIFF
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -1,10 +1,10 @@
 # Winit Scope
 
 Winit aims to expose an interface that abstracts over window creation and input handling, and can
-be used to create both games and applications. It supports the main graphical platforms:
+be used to create both games and applications. It supports the following main graphical platforms:
 - Desktop
-  - Windows
-  - macOS
+  - Windows 7+ (10+ is tested regularly)
+  - macOS 10.7+ (10.14+ is tested regularly)
   - Unix
     - via X11
     - via Wayland


maroider:
> maybe it should be as verbose as specifying each browser an the version that it is expected to work on?

I think it could make sense to list what browsers we specifically try to support, and since browsers need to be updated on a regular basis, I think it's not that unreasonable to only really try to target the latest stable version (and maybe one or two older versions if there's a desire and bandwidth for it) of each. In practice, I suspect this ends up being mainly Chrome and Firefox, and maybe Safari, but I'm unsure if anyone who touches the web backend with any regularity actually tests anything on Safari.

madsmtm:
Well, also kinda depends on what these support tables are to be used for - reconsidering, I think it makes sense to guarantee support for some specific version (or a general "latest firefox + chrome"), which also means that if a PR doesn't support that version, it's a breaking change.

Of course, if someone reports that `winit` doesn't work on some lower version, we'll do our best to fix that.

kchibisov:
That should be fine as is for now.

Issue #2415, opened 2022-08-10
RFC: Could winit 0.27 optionally support raw_window_handle 0.4 (for compatibility with wgpu 0.13) perhaps?


rib:
The recent introduction of the raw_window_handle 0.5 crate dependency has ended up blocking some projects from being able to update to winit 0.27 while they need to wait for it to permeate through other crates first (e.g. waiting for a wgpu 0.14 release which is estimated to happen around the end of September, ref: https://github.com/gfx-rs/wgpu/pull/2918)

For reference, here is an Egui PR that looked at upgrading to Winit 0.27: https://github.com/emilk/egui/pull/1877 which is blocked on wgpu 0.14.

Similarly a PR for Bevy: https://github.com/bevyengine/bevy/pull/5347 that's also blocked on wgpu 0.14 (though the timing looks reasonable here because they only just released Bevy 0.8 anyway, so waiting for wgpu 0.14 isn't a huge deal)

More generally; considering how central wgpu is within the Rust ecosystem I wonder if Winit could strive to ensure, where possible, that new releases remain compatible with the most recent stable release of wgpu.

As it is I guess there will be quite some lag before developers really start using Winit 0.27 which might also mean that potential issues with a new release may not get discovered until quite late.

Maybe future changes similar to this could be phased in over two releases to avoid having to wait a full release cycle for crates like wgpu before a new version of winit can be adopted.

In this case I suppose it was also just unfortunate timing that wgpu was released so recently, just before winit 0.27 was released.

kchibisov:
If someone has time and dedication to add a feature for old raw window handle, I can review and we can make a new bug fix release for that, not a big deal. But I don't have time to write it up myself. 

In general that would be nice to have some communication in terms of bugs/winit releases, etc, with major consumers. Unfortunately only alacritty is present here and provide in time feedback due to me actively working on all related projects. I could be present on wgpu/whatever bug tracker/channels, but I don't have infinite time, sadly to keep track of everything..

maroider:
I'm very much interested in having better coordination around these sorts of things, and I'd be happy to spend the time required to facilitate this.

[...]

PR #2418, opened 2022-08-10
Implement version 0.4 of the `HasRawWindowHandle` trait

rib:
This makes Winit 0.27 compatible with crates like Wgpu 0.13 that are
using the raw_window_handle v0.4 crate and aren't able to upgrade to 0.5
until they do a new release (since it requires a semver change).

The change is intended to be self-contained (instead of pushing
the details into all the platform_impl backends) since this is only
intended to be a temporary trait implementation for backwards
compatibility that will likely be removed before the next Winit release.

Addresses: #2415

**Update:** I've marked it a tested since Emil left this comment about testing this branch with Egui: https://github.com/emilk/egui/pull/1877#issuecomment-1211035430

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,7 +46,8 @@ instant = { version = "0.1", features = ["wasm-bindgen"] }
 once_cell = "1.12"
 log = "0.4"
 serde = { version = "1", optional = true, features = ["serde_derive"] }
-raw-window-handle = "0.5.0"
+raw_window_handle = { package = "raw-window-handle", version = "0.5" }
+raw_window_handle_04 = { package = "raw-window-handle", version = "0.4" }
 bitflags = "1"
 mint = { version = "0.5.6", optional = true }
 
--- a/src/window.rs
+++ b/src/window.rs
@@ -1066,6 +1066,109 @@ unsafe impl HasRawDisplayHandle for Window {
         self.window.raw_display_handle()
     }
 }
+unsafe impl raw_window_handle_04::HasRawWindowHandle for Window {
+    /// Returns a [`raw_window_handle_04::RawWindowHandle`] for the Window
+    ///
+    /// This provides backwards compatibility for downstream crates that have not yet
+    /// upgraded to `raw_window_handle` version 0.5, such as Wgpu version 0.13.
+    ///
+    /// ## Platform-specific
+    ///
+    /// ### Android
+    ///
+    /// Only available after receiving [`Event::Resumed`] and before [`Event::Suspended`]. *If you
+    /// try to get the handle outside of that period, this function will panic*!
+    ///
+    /// Make sure to release or destroy any resources created from this `RawWindowHandle` (ie. Vulkan
+    /// or OpenGL surfaces) before returning from [`Event::Suspended`], at which point Android will
+    /// release the underlying window/surface: any subsequent interaction is undefined behavior.
+    ///
+    /// [`Event::Resumed`]: crate::event::Event::Resumed
+    /// [`Event::Suspended`]: crate::event::Event::Suspended
+    fn raw_window_handle(&self) -> raw_window_handle_04::RawWindowHandle {
+        use raw_window_handle_04::{
+            AndroidNdkHandle, AppKitHandle, HaikuHandle, OrbitalHandle, UiKitHandle, WaylandHandle,
+            WebHandle, Win32Handle, WinRtHandle, XcbHandle, XlibHandle,
+        };
+
+        // XXX: Ideally this would be encapsulated either through a
+        // compatibility API from raw_window_handle_05 or else within the
+        // backends but since this is only to provide short-term backwards
+        // compatibility, we just handle the full mapping inline here.
+        //
+        // The intention is to remove this trait implementation before Winit
+        // 0.28, once crates have had time to upgrade to raw_window_handle 0.5
+
+        match (self.window.raw_window_handle(), self.window.raw_display_handle()) {
+            (RawWindowHandle::UiKit(window_handle), _) => {
+                let mut handle = UiKitHandle::empty();
+                handle.ui_view = window_handle.ui_view;
+                handle.ui_window = window_handle.ui_window;
+                handle.ui_view_controller = window_handle.ui_view_controller;
+                raw_window_handle_04::RawWindowHandle::UiKit(handle)
+            },
+            (RawWindowHandle::AppKit(window_handle), _) => {
+                let mut handle = AppKitHandle::empty();
+                handle.ns_window = window_handle.ns_window;
+                handle.ns_view = window_handle.ns_view;
+                raw_window_handle_04::RawWindowHandle::AppKit(handle)
+            },
+            (RawWindowHandle::Orbital(window_handle), _) => {
+                let mut handle = OrbitalHandle::empty();
+                handle.window = window_handle.window;
+                raw_window_handle_04::RawWindowHandle::Orbital(handle)
+            },
+            (RawWindowHandle::Xlib(window_handle), RawDisplayHandle::Xlib(display_handle)) => {
+                let mut handle = XlibHandle::empty();
+                handle.display = display_handle.display;
+                handle.window = window_handle.window;
+                handle.visual_id = window_handle.visual_id;
+                raw_window_handle_04::RawWindowHandle::Xlib(handle)
+            },
+            (RawWindowHandle::Xcb(window_handle), RawDisplayHandle::Xcb(display_handle)) => {
+                let mut handle = XcbHandle::empty();
+                handle.connection = display_handle.connection;
+                handle.window = window_handle.window;
+                handle.visual_id = window_handle.visual_id;
+                raw_window_handle_04::RawWindowHandle::Xcb(handle)
+            },
+            (RawWindowHandle::Wayland(window_handle), RawDisplayHandle::Wayland(display_handle)) => {
+                let mut handle = WaylandHandle::empty();
+                handle.display = display_handle.display;
+                handle.surface = window_handle.surface;
+                raw_window_handle_04::RawWindowHandle::Wayland(handle)
+            },
+            (RawWindowHandle::Win32(window_handle), _) => {
+                let mut handle = Win32Handle::empty();
+                handle.hwnd = window_handle.hwnd;
+                handle.hinstance = window_handle.hinstance;
+                raw_window_handle_04::RawWindowHandle::Win32(handle)
+            },
+            (RawWindowHandle::WinRt(window_handle), _) => {
+                let mut handle = WinRtHandle::empty();
+                handle.core_window = window_handle.core_window;
+                raw_window_handle_04::RawWindowHandle::WinRt(handle)
+            },
+            (RawWindowHandle::Web(window_handle), _) => {
+                let mut handle = WebHandle::empty();
+                handle.id = window_handle.id;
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -46,7 +46,8 @@ instant = { version = "0.1", features = ["wasm-bindgen"] }
 once_cell = "1.12"
 log = "0.4"
 serde = { version = "1", optional = true, features = ["serde_derive"] }
-raw-window-handle = "0.5.0"
+raw_window_handle_05 = { package = "raw-window-handle", version = "0.5" }
```
madsmtm, line 49: Nit: Maybe use `raw_window_handle` for the new version, and `raw_window_handle_04` for the old?

madsmtm (commented):
I'm fine with this, unless we find a way to do the semver trick again (I'll have a go at it as well as @maroider).

rib:
> I'm fine with this, unless we find a way to do the semver trick again (I'll have a go at it as well as @maroider).

We could use use this same code to do something similar to the semver trick (the terminology is maybe a bit muddled here though) where we could update the raw_window_handle 0.5 crate to depend on v0.4 so that it can provide a blanket implementation of the 0.4 trait in terms of the new one.

One trade-off there might be that it would be harder to back that out as a temporary change for backwards compatibility, which I think might be more reasonable to do here in Winit.

Cargo.toml
rib: yeah, makes sense

rib: fixed

emilk:
This is a really good idea from my point of view, and would unblock `egui/eframe` updating to winit 0.27 (provided it is followed up by a 0.27.2 patch release)

rib:
For reference, an alternative solution I'm experimenting with is the "semver trick" approach, which updates raw_window_handle 0.4 to implement its trait automatically for anything that implements the the new traits. This would effectively make wgpu compatible with anything built against raw_window_handle 0.5.

This is the blanket implementation I have for raw_window_handle 0.4 currently:
```rust
/// For forwards compatibility this provides a blanket implementation of the `RawWindowHandle`
/// for anything that is based on the newer 0.5 version of the `raw_window_handle` crate and
/// implements both `HasRawWindowHandle` and `HasRawDisplayHandle`
unsafe impl<T: raw_window_handle_05::HasRawWindowHandle + raw_window_handle_05::HasRawDisplayHandle> HasRawWindowHandle for T
{
    fn raw_window_handle(&self) -> RawWindowHandle {
        match raw_window_handle_05::HasRawWindowHandle::raw_window_handle(self) {
            raw_window_handle_05::RawWindowHandle::UiKit(window_handle) => {
                let mut handle = UiKitHandle::empty();
                handle.ui_view = window_handle.ui_view;
                handle.ui_window = window_handle.ui_window;
                handle.ui_view_controller = window_handle.ui_view_controller;
                RawWindowHandle::UiKit(handle)
            },
            raw_window_handle_05::RawWindowHandle::AppKit(window_handle) => {
                let mut handle = AppKitHandle::empty();
                handle.ns_window = window_handle.ns_window;
                handle.ns_view = window_handle.ns_view;
                RawWindowHandle::AppKit(handle)
            },
            raw_window_handle_05::RawWindowHandle::Orbital(window_handle) => {
                let mut handle = OrbitalHandle::empty();
                handle.window = window_handle.window;
                RawWindowHandle::Orbital(handle)
            },
            raw_window_handle_05::RawWindowHandle::Xlib(window_handle) => {
                if let raw_window_handle_05::RawDisplayHandle::Xlib(display_handle) = raw_window_handle_05::HasRawDisplayHandle::raw_display_handle(self) {
                    let mut handle = XlibHandle::empty();
                    handle.display = display_handle.display;
                    handle.window = window_handle.window;
                    handle.visual_id = window_handle.visual_id;
                    RawWindowHandle::Xlib(handle)
                } else {
                    panic!("Unsupported display handle associated with Xlib window");
                }
            },
            raw_window_handle_05::RawWindowHandle::Xcb(window_handle) => {
                if let raw_window_handle_05::RawDisplayHandle::Xcb(display_handle) = raw_window_handle_05::HasRawDisplayHandle::raw_display_handle(self) {
                    let mut handle = XcbHandle::empty();
                    handle.connection = display_handle.connection;
                    handle.window = window_handle.window;
                    handle.visual_id = window_handle.visual_id;
                    RawWindowHandle::Xcb(handle)
                } else {
                    panic!("Unsupported display handle associated with Xcb window");
                }
            },
            raw_window_handle_05::RawWindowHandle::Wayland(window_handle) => {
                if let raw_window_handle_05::RawDisplayHandle::Wayland(display_handle) = raw_window_handle_05::HasRawDisplayHandle::raw_display_handle(self) {
                    let mut handle = WaylandHandle::empty();
                    handle.display = display_handle.display;
                    handle.surface = window_handle.surface;
                    RawWindowHandle::Wayland(handle)
                } else {
                    panic!("Unsupported display handle associated with Xcb window");
                }
            },
            raw_window_handle_05::RawWindowHandle::Win32(window_handle) => {
                let mut handle = Win32Handle::empty();
                handle.hwnd = window_handle.hwnd;
                handle.hinstance = window_handle.hinstance;
                RawWindowHandle::Win32(handle)
            },
            raw_window_handle_05::RawWindowHandle::WinRt(window_handle) => {
                let mut handle = WinRtHandle::empty();
                handle.core_window = window_handle.core_window;
                RawWindowHandle::WinRt(handle)
            },
            raw_window_handle_05::RawWindowHandle::Web(window_handle) => {
                let mut handle = WebHandle::empty();
                handle.id = window_handle.id;
                RawWindowHandle::Web(handle)
            },
            raw_window_handle_05::RawWindowHandle::AndroidNdk(window_handle) => {
                let mut handle = AndroidNdkHandle::empty();
                handle.a_native_window = window_handle.a_native_window;
                RawWindowHandle::AndroidNdk(handle)
            },
            raw_window_handle_05::RawWindowHandle::Haiku(window_handle) => {
                let mut handle = HaikuHandle::empty();
                handle.b_window = window_handle.b_window;
                handle.b_direct_window = window_handle.b_direct_window;
                RawWindowHandle::Haiku(handle)
            },
            _ => panic!("No HasRawWindowHandle version 0.4 backwards compatibility for new Winit window type"),
        }
    }
}
```

The problem with that is that I'm not sure atm how to avoid conflicting with with the other blanket implementation that exists in the raw_window_handle crate:
```rust
unsafe impl<'a, T: HasRawWindowHandle + ?Sized> HasRawWindowHandle for &'a T {
    fn raw_window_handle(&self) -> RawWindowHandle {
        (*self).raw_window_handle()
    }
}
```

I tried adding a `+ Sized` requirement for the new `impl` to try and avoid conflicting with this other implementation but that didn't work. maybe someone more familiar with how rust trait bounds work has a clue for how to let these both exist together?

madsmtm:
Heh, I tried this in https://github.com/rust-windowing/raw-window-handle/pull/97 as well, I don't think it's possible to do (without specialization or other unstable features)

madsmtm:
Yet another alternative: Make version `0.5` not a breaking change (add type-aliases and deprecated fields), though that only delays the issue until when we want to actually remove the deprecated functionality...

rib:
> Yet another alternative: Make version `0.5` not a breaking change (add type-aliases and deprecated fields), though that only delays the issue until when we want to actually remove the deprecated functionality...

My initial approach had been to handle this in 0.5 and depend on the 0.4 crate but the orphan rules (or something like that) mean you can't provide a blanket implementation for the 0.4 trait that's not implemented in the current crate - so then it wouldn't provide compatibility with wgpu that depends on the 0.4 crate.

madsmtm:
> My initial approach had been to handle this in 0.5 and depend on the 0.4 crate but the orphan rules (or something like that) mean you can't provide a blanket implementation for the 0.4 trait that's not implemented in the current crate - so then it wouldn't provide compatibility with wgpu that depends on the 0.4 crate.

Yeah, those blanket traits are really interfering with the semver trick.

rib:
atm I guess I would er towards handling the backwards compatibility in Winit - I think mainly because it seems cleaner to just remove it before releasing 0.28 and it just provides a short-term stop-gap until wgpu updates to 0.5

madsmtm:
I'm fine with either

kchibisov:
Winit should work around that issue, since it'll be solved by either a wgpu or winit release. Given that wgpu is harder to bump we can solve it in winit, since there's a clear demand from the consumers.

raw-window-handle shouldn't be touched at all here, since its api is stable to some point and keeping shim for old api will result in a more breaking change in the future.

rib:
Just for reference Emil mentioned here that they were able to test this branch with Egui and it seemed to work as intended: https://github.com/emilk/egui/pull/1877#issuecomment-1211035430

src/window.rs
```diff
@@ -1066,6 +1066,121 @@ unsafe impl HasRawDisplayHandle for Window {
...
+
+        // XXX: Ideally this would be encapsulated either through a
+        // compatibility API from raw_window_handle_05 or else within the
+        // backends but since this is only to provide short-term backwards
+        // compatibility, we just handle the full mapping inline here.
+        //
+        // The intention is to remove this trait implementation before Winit
+        // 0.28, once crates have had time to upgrade to raw_window_handle 0.5
+
+        match self.window.raw_window_handle() {
```
kchibisov, line 1102: Could you match both `raw_window_handle` and `raw_display_handle`, so we can get rid of `if let` craft?

kchibisov (changes requested):

src/window.rs
rib: I've updated it to match both in a tuple now so e.g. the Xlib case looks like this now:
```rust
           (RawWindowHandle::Xlib(window_handle), RawDisplayHandle::Xlib(display_handle)) => {
                let mut handle = XlibHandle::empty();
                handle.display = display_handle.display;
                handle.window = window_handle.window;
                handle.visual_id = window_handle.visual_id;
                raw_window_handle_04::RawWindowHandle::Xlib(handle)
            },
```

rib:
I just noticed that I needed to rebase my patch on top of a "breaking" change conflict in the changelog which I guess will need to be branched around now to be able to spin a 0.27.2 release with this patch.

That's kinda awkward timing

kchibisov:
I don't think that Windows change is actually a breaking change. Given that missing support for the Api was added. I'll send a patch to correct that.

rib:
> I don't think that Windows change is actually a breaking change. Given that missing support for the Api was added. I'll send a patch to correct that.

ah, would make things easier if that's the case 🤞

maroider:
> I don't think that Windows change is actually a breaking change.

AIUI, it changes the default behavior for device events from "always delivered" to "delivered when application is focused", which I would argue is techinally breaking, although I suspect very few people have actually updated to 0.28.0 (since we introduced the API in said version), so the breakage will likely be minor. The "best practice" is to ignore device events when you don't have the focus anyway, so that is likely also a mitigating factor.

kchibisov:
>AIUI, it changes the default behavior for device events from "always delivered" to "delivered when application is focused", which I would argue is techinally breaking, although I suspect very few people have actually updated to 0.28.0 (since we introduced the API in said version), so the breakage will likely be minor. The "best practice" is to ignore device events when you don't have the focus anyway, so that is likely also a mitigating factor.

The documented default behavior is `Auto`, so if someone relies on `Never`, they should explicitly state that by using a function. So it's not a breaking change, since relying on that means that they rely on missing impl in winit.

Issue #2413, opened 2022-08-09
HitTest event


msiglreith:
Add a `WindowEvent::HitTest(button, x, y, &mut Area)` event which allows the user to specify special interaction areas on button clicks. Area maybe one of the borders, the caption, system buttons, etc. Inspired by the [WM_NCHITTEST](https://docs.microsoft.com/en-us/windows/win32/inputdev/wm-nchittest) event.

### Motivation
- Mostly interesting for windows without decorations where we are usually required to emulate the system behavior.
- On Windows, feedback of the internal `WM_NCHITTEST` seems to be used for certain system behavior, ie. snap layout when hovering the maximize button or screen reader support.
- Unify the `drag_` functions and hopefully improve the ['left button down'-guessing](https://docs.rs/winit/latest/winit/window/struct.Window.html#method.drag_window) situation

maroider:
This is something I admittedly contemplated when `drag_window` was being implemented, but I ultimately didn't bring it up since I wasn't confident it could be meaningfully generalized across platforms.

PR #2419, opened 2022-08-10
On Windows, improve support for undecorated windows

msiglreith:
Alternative PR to https://github.com/rust-windowing/winit/pull/1891

- Add an option to add the drop shadow for borderless windows. This also will result in the compositor to correctly identify the window to draw the rounded corners on Windows 11. Unfortunately shows an black 1px line on top of the window.. overall just sad state of legacy stuff piled up and the compositor isn't very helpful either ):
- Support for Shake, Snap, Alt-Print, etc features which a normal window can do. This is derived by the window styles. To not show the non-client area we just spread the client area to cover it (mostly)

Non-client area hit control https://github.com/rust-windowing/winit/issues/2413 would be needed as next stop to properly emulate resize borders (not in the dropshadow tho..) and other functionality like dragging for the shake feature.

- [x] Tested on all platforms changed: Windows 10, Windows 11 VM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ And please only add new entries to the top of this list, right below the `# Unre
 # Unreleased
 
 - On Windows, respect min/max inner sizes when creating the window.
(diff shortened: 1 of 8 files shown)


msiglreith:
Thanks a lot for testing!

PR #2420, opened 2022-08-11
Add CODEOWNERS file

madsmtm:
This makes it very clear when you're stepping down from the post as a maintainer, and makes it clear for users who is expected to review their PR.

The way it works is [detailed in the docs](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners), but in short, if edits are made to files specified here, the "code owner" will automatically be assigned as the reviewer.
An example: If I submitted a PR that edited `src/event.rs`, `src/platform/windows.rs` and `src/platform_impl/linux/wayland/output.rs`, `@msiglreith` and `@kchibisov` would automatically be marked as reviewers.

PR #2421, opened 2022-08-11
Split `platform::unix` into `platform::x11` and `platform::wayland`

madsmtm:
There were basically no overlap between the platform-specific extensions in these, so it made little sense to keep them in the same file!

Will gladly rename the functions as well, e.g. `wayland_set_csd_theme -> set_csd_theme`?

~Blocked-ish on #2420~

### DIFF
--- a/src/platform/wayland.rs
+++ b/src/platform/wayland.rs
@@ -142,13 +142,14 @@ impl WindowExtWayland for Window {
 
 /// Additional methods on [`WindowBuilder`] that are specific to Wayland.
 pub trait WindowBuilderExtWayland {
-    /// Build window with the given application ID.
+    /// Build window with the given name.
     ///
-    /// This should match the `.desktop` file destributed with your program.
+    /// The `general` name sets an application ID, which should match the `.desktop`
+    /// file destributed with your program. The `instance` is a `no-op`.
     ///
     /// For details about application ID conventions, see the
     /// [Desktop Entry Spec](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#desktop-file-id)
-    fn with_application_id(self, application_id: impl Into<String>) -> Self;
+    fn with_name(self, general: impl Into<String>, instance: impl Into<String>) -> Self;
 
     /// Build window with certain decoration [`Theme`]
     ///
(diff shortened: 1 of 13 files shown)


notgull:
I agree with this decision. Wayland and X11 are very different, so it makes sense to keep them in their own separate modules.

src/platform/wayland.rs
```diff
@@ -0,0 +1,186 @@
...
+
+/// Additional methods on [`WindowBuilder`] that are specific to Wayland.
+pub trait WindowBuilderExtWayland {
+    /// Build window with the given application ID.
+    ///
+    /// This should match the `.desktop` file destributed with your program.
+    ///
+    /// For details about application ID conventions, see the
+    /// [Desktop Entry Spec](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#desktop-file-id)
+    fn with_application_id(self, application_id: impl Into<String>) -> Self;
```
madsmtm, line 151: Previously, this was named `with_name`, like it still is in `WindowBuilderExtX11::with_name`, but I felt this `with_application_id` better captures what the meaning actually is!

kchibisov: This was intentionally and that is to unify X11 and Wayland function to set title. I don't really want to go back to separate methods again. I'm still not sure that I want to split them, given that it doesn't provide that much value.

They where separate before and we had `with_appid` and `with_class`, but there were a request to unify them.

madsmtm: Fair enough, I've changed it back

src/platform/wayland.rs
```diff
@@ -0,0 +1,186 @@
...
+    /// Possible values for env variable are: "dark" and light"
+    fn wayland_set_csd_theme(&self, config: Theme);
+
+    /// Check if the window is ready for drawing
+    ///
+    /// It is a remnant of a previous implementation detail for the
+    /// wayland backend, and is no longer relevant.
+    ///
+    /// Always return `true`.
+    #[deprecated]
```
kchibisov, line 104: Could we remove it entirely?

src/platform/x11.rs
```diff
@@ -225,9 +128,8 @@ pub trait WindowExtUnix {
     fn is_ready(&self) -> bool;
```
kchibisov, line 128: This should be removed as well?

kchibisov (changes requested):

kchibisov (changes requested):
CI.

kchibisov (approved):

PR #2422, opened 2022-08-11
Use `super` instead of `platform_impl::platform` in paths

madsmtm:
Makes it easier to do refactors later on

### DIFF
--- a/src/platform_impl/ios/app_state.rs
+++ b/src/platform_impl/ios/app_state.rs
@@ -12,19 +12,19 @@ use std::{
 use objc::runtime::{BOOL, YES};
 use once_cell::sync::Lazy;
 
+use super::{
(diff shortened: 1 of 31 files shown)


kchibisov:
Honestly, I really don't like `super` path for modules. It's not that clear where things are coming from, etc. They are not that readable and prone to errors when you move files around (e.g. when you move something to a module).

madsmtm:
I agree somewhat, for example I didn't change this on the `wayland` platform, since it would mean doing `super::super::super`. But I think `crate::platform_impl::platform` is just ugly and confusing (since there's not actually a `platform` module in `platform_impl`, it's just an implementation detail of how `platform_impl` works).

Motivation was because I'd like to change these paths in the future, and it would mean a smaller diff when that time comes.

kchibisov:
I don't like super in general.

PR #2427, opened 2022-08-12
Move to `objc2`

madsmtm:
_A sneak-peak at the improvements I've been working on for iOS and macOS._

Move to [`objc2`](https://github.com/madsmtm/objc2), my fork of [`objc`](https://github.com/SSheldon/rust-objc) which adds a lot of things, in particular macros for proper memory management, and a lot of fixed soundness issues.

When done, this should _vastly_ improve the safety and ease-of-maintenance of the iOS and macOS platforms. Opening this PR to have something to refer to, but will split it up into smaller parts when I'm ready to actually get it merged.

  - [x] macOS 10.14
  - [x] iOS simulator (can't get log output here though, so a bit harder)
  - [x] iOS 9.3.6 (iPad mini 1st gen)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,11 +65,8 @@ ndk-glue = "0.7.0"
 objc = { version = "=0.3.0-beta.3", package = "objc2" }
 
 [target.'cfg(target_os = "macos")'.dependencies]
-# Branch: objc2
-# TODO: Use non-git versions before we release
-cocoa = { git = "https://github.com/madsmtm/core-foundation-rs.git", rev = "c770e620ba0766fc1d2a9f83327b0fee905eb5cb" }
-core-foundation = { git = "https://github.com/madsmtm/core-foundation-rs.git", rev = "c770e620ba0766fc1d2a9f83327b0fee905eb5cb" }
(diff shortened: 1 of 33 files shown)


notgull:
Should we enable the "exception" feature? It seems unsound got exceptions to unwind into Rust.

madsmtm:
> Should we enable the "exception" feature? It seems unsound got exceptions to unwind into Rust.

The feature flag is called `catch-all` if you want every `msg_send!` to be wrapped in a `@try/@catch`.

Re soundness: That feature is just as unsound as letting exceptions unwind into Rust, since our message sending is implemented in Rust. `extern "C-unwind"` would need to be stable before it is sound (can be tried with `objc2`'s `unstable-c-unwind` feature flag), but that makes the point moot as letting exceptions just unwind into Rust also becomes sound.

Practically, I haven't enabled it because exceptions are rare, the user-experience approximately the same, and enabling it requires running `clang` which is inconvenient for some users.

madsmtm:
This is very close to where I want it to be, so I'll [start opening smaller PRs](https://github.com/rust-windowing/winit/pull/2452).

madsmtm:
This is finished, the remaining parts are in https://github.com/rust-windowing/winit/pull/2465 and https://github.com/rust-windowing/winit/pull/2462

PR #2428, opened 2022-08-12
iOS: Fix a few instances of UB

madsmtm:
Mostly interesting for 32-bit iOS.

Found while making #2427.


francesca64:
Sorry for the delay! This LGTM, but I'm currently unemployed and therefore no longer have any iOS devices to test on. 

@ArthurKValladares would you mind taking a look at this?

Issue #1550, opened 2020-04-29
Drag&Drop API / HoveredFile, DroppedFile


dhardy:
This API is too limited in my opinion. In general, one needs at least a coordinate relative to the window and to receive updates when this coordinate changes. In my tests on X11, the app does not receive usual cursor or touch events when a file is hovered, hence the window has no way of determining the current coordinate.

Suggestion: add a "source" parameter to `DroppedFile`, `HoveredFile`, and `HoveredFileCancelled`, and continue to deliver cursor and touch events as normal.

Alternatively, include a coordinate and add a `HoveredFileMoved` event.

(I have not investigated platform capabilities and don't have a direct use, but wish to add support to KAS, which must determine the widget under the drop.)

Boscop:
I just spend 2 days trying to figure out why my imgui application thinks I dropped the file at a different location than where my mouse pointer is when I drop the file (when I release the mouse button).

It turned out that it's because winit doesn't send any `CursorMoved` events (nor `CursorEntered`) before `DroppedFile`, only immediately after. That's too late :/
**My application sees the file get dropped at the location where the mouse was when it was last hovered over the window** (where it was before `CursorLeft`), which is totally confusing behavior for the user because
this causes the file to get perceived as being dropped into the wrong "drop area" :(

Ideally, **winit should send `CursorEntered` and `CursorMoved` as soon as the file-dragging cursor enters the window. And even before `HoveredFile` is sent.**

Can you please fix this? :)
This change wouldn't break anything, right?

My use case:
I have different "drop areas" on my imgui app and I need to detect which one receives the dropped file.
This can only work if I get mouse pos events before `DroppedFile`.
(In between `HoveredFile` and `DroppedFile`, I also want to highlight the drop area that the mouse is currently hovering over (the one that would receive the dropped file if the mouse button was released here), so this also requires getting mouse pos events before `DroppedFile`.)

janus-culsans:
This is a major issue, I can't even think of many usecases where it would help to just have an event for "dragged a file over the window", unless the file can only be put into exactly one place.

kvark:
Unfortunately hitting this as well. I'm working on a scene editor. When dropping a model into the scene, I need to show where it's going to land. This is currently impossible since Winit doesn't communicate the cursor position to me during the drag :(

daxpedda:
[...]

PR #2429, opened 2022-08-12
Wayland: support basic Drag&Drop

### LINKED ISSUE #1881: Support drag and drop on wayland
Drag and drop windows events are currently not supported on wayland.

I wonder how it is possible to implement it considering smithay-client-toolkit contains some helper for it: DataDevice.

SludgePhD:

Closes https://github.com/rust-windowing/winit/issues/1881

### DIFF
--- a/src/platform_impl/linux/wayland/event_loop/state.rs
+++ b/src/platform_impl/linux/wayland/event_loop/state.rs
@@ -2,6 +2,8 @@
 
 use std::collections::HashMap;
 
+use wayland_client::Display;
+
 use super::EventSink;
 use crate::platform_impl::wayland::window::shim::{
     WindowCompositorUpdate, WindowHandle, WindowUserRequest,
@@ -28,4 +30,7 @@ pub struct WinitState {
     /// and requests from winit's windows are being forwarded to them either via
     /// `WindowUpdate` or buffer on the associated with it `WindowHandle`.
     pub window_map: HashMap<WindowId, WindowHandle>,
+
+    /// Handle to the display, to flush events when needed.
+    pub display: Display,
 }
--- a/src/platform_impl/linux/wayland/event_loop/state.rs
+++ b/src/platform_impl/linux/wayland/event_loop/state.rs
@@ -2,8 +2,6 @@
 
 use std::collections::HashMap;
 
-use wayland_client::Display;
-
 use super::EventSink;
 use crate::platform_impl::wayland::window::shim::{
     WindowCompositorUpdate, WindowHandle, WindowUserRequest,
@@ -30,7 +28,4 @@ pub struct WinitState {
     /// and requests from winit's windows are being forwarded to them either via
     /// `WindowUpdate` or buffer on the associated with it `WindowHandle`.
     pub window_map: HashMap<WindowId, WindowHandle>,
-
-    /// Handle to the display, to flush events when needed.
-    pub display: Display,
 }
--- a/src/platform_impl/linux/wayland/seat/dnd/handlers.rs
+++ b/src/platform_impl/linux/wayland/seat/dnd/handlers.rs
@@ -1,11 +1,15 @@
 use std::{
-    io::{self, BufRead, BufReader},
+    io::{self, Read},
+    os::unix::prelude::{AsRawFd, RawFd},
     path::PathBuf,
+    str,
 };
 
 use percent_encoding::percent_decode_str;
-use sctk::data_device::{DataOffer, DndEvent};
-use wayland_client::Display;
+use sctk::{
+    data_device::{DataOffer, DndEvent, ReadPipe},
+    reexports::calloop::{generic::Generic, Interest, LoopHandle, Mode, PostAction},
+};
 
 use crate::{
     dpi::PhysicalPosition,
@@ -27,11 +31,10 @@ pub(super) fn handle_dnd(event: DndEvent<'_>, inner: &mut DndInner, winit_state:
             ..
         } => {
             let window_id = make_wid(&surface);
-
-            if let Ok(paths) = parse_offer(&winit_state.display, offer) {
+            inner.window_id = Some(window_id);
+            offer.accept(Some(MIME_TYPE.into()));
+            let _ = parse_offer(&inner.loop_handle, offer, move |paths, winit_state| {
                 if !paths.is_empty() {
-                    offer.accept(Some(MIME_TYPE.into()));
-
                     winit_state.event_sink.push_window_event(
                         WindowEvent::CursorEntered {
                             device_id: crate::event::DeviceId(
@@ -56,21 +59,20 @@ pub(super) fn handle_dnd(event: DndEvent<'_>, inner: &mut DndInner, winit_state:
                             .event_sink
                             .push_window_event(WindowEvent::HoveredFile(path), window_id);
                     }
-                    inner.window_id = Some(window_id);
                 }
-            }
+            });
         }
         DndEvent::Drop { offer: Some(offer) } => {
             if let Some(window_id) = inner.window_id {
                 inner.window_id = None;
 
-                if let Ok(paths) = parse_offer(&winit_state.display, offer) {
+                let _ = parse_offer(&inner.loop_handle, offer, move |paths, winit_state| {
                     for path in paths {
                         winit_state
                             .event_sink
                             .push_window_event(WindowEvent::DroppedFile(path), window_id);
                     }
-                }
+                });
             }
         }
         DndEvent::Leave => {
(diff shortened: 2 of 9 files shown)


SludgePhD:
I've also pushed a commit that emits all appropriate cursor events, which should address https://github.com/rust-windowing/winit/issues/1550 at least on Wayland. Other platforms don't do this yet, so let me know if I should leave it out.

(though one issue I noticed is that some compositors send DnD events when the cursor is slightly outside the window (eg. hovering over its shadow only on KDE), which results in out-of-bounds cursor coordinates being sent to the app.

kchibisov:
Could you test that it doesn't blow up in alacritty for example or when using https://github.com/smithay/smithay-clipboard ? I think gnome doesn't allow to have drag and drop and clipboard by using multiple data device managers. IIRC it uses just the last one, and that's the reason I haven't implemented drag and drop inside winit, since winit should support clipboard.

The work to bring clipboard is #2156 , but I don't have that much time to work on it right now unfortunately, but I hope to go back to it soon...

SludgePhD:
I've tested with Alacritty on KDE Plasma, and both Drag&Drop and clipboard access work fine with this patch applied.

If GNOME doesn't support this properly, that does seem unfortunate, but this PR will at least make apps work on other compositors, and also make apps work everywhere that only need Drag&Drop and no clipboard access.

Do you happen to know if GNOME's behavior is correct here or does the Wayland specification *require* compositors to support multiple data device managers?

kchibisov:
Would suggest to test GNOME, since it's a major compositor and we can't break it even when it's a GNOME bug.

SludgePhD:
Alacritty built with this patch running in a nested GNOME session does not crash or break when dropping a file, but Drag&Drop doesn't do anything (so behavior matches winit master). Clipboard access still works fine.

Amusingly, Konsole in the same nested GNOME session *does* crash when dropping a file from a Dolphin window in the same session. Additionally, closing the nested GNOME shell window made it abort with several assertion failures, so it looks like winit is miles ahead of the typical Linux-on-the-desktop experience already... sadly

kchibisov:
The thing is that it'll work, but if you create another device manager the privious one (in winit here) would stop working. I do want to have drag and drop in winit Wayland, but I'm not sure whether I'll have time to look into it, given I think that clipboard issue should be resolved first, so we can have both at the same time without collapsing.

SludgePhD:
Yeah, I understand (unless Alacritty somehow does something differently?). As far as I can tell, this PR does not regress anything even if applications do use another data manager for accessing the clipboard. Drag&Drop simply won't work, just like it does now.

kchibisov:
I know what alacritty does and I know why the feature wasn't in winit, since I remember the GNOME bug given that it's quite easy to repro it. Besides that it's fair to assume that we can have drag and drop that way for now.

src/platform_impl/linux/wayland/seat/dnd/handlers.rs
```diff
@@ -0,0 +1,136 @@
+use std::{
+    io::{self, BufRead, BufReader},
+    path::PathBuf,
+};
```
kchibisov, line 4: Could you use the style of the wayland module and group by module? Should apply everywhere.

src/platform_impl/linux/wayland/seat/dnd/handlers.rs
```diff
@@ -0,0 +1,136 @@
...
+        _ => {}
+    }
+}
+
+fn parse_offer(display: &Display, offer: &DataOffer) -> io::Result<Vec<PathBuf>> {
+    let can_accept = offer.with_mime_types(|types| types.iter().any(|s| s == MIME_TYPE));
+    if can_accept {
+        // Format: https://www.iana.org/assignments/media-types/text/uri-list
+        let mut paths = Vec::new();
+        let pipe = offer.receive(MIME_TYPE.into())?;
```
kchibisov, line 116: The pipe must be inserted and handled via callop pipe reader and be non blocking when doing so, otherwise it can deadlock.

src/platform_impl/linux/wayland/seat/dnd/handlers.rs
```diff
@@ -0,0 +1,136 @@
...
+        let _ = display.flush();
```
kchibisov, line 117: Not sure why you flush here.

src/platform_impl/linux/wayland/seat/dnd/handlers.rs
```diff
@@ -0,0 +1,136 @@
...
+
+            let decoded = match percent_decode_str(&line).decode_utf8() {
+                Ok(decoded) => decoded,
+                Err(_) => continue,
+            };
+            if let Some(path) = decoded.strip_prefix("file://") {
+                paths.push(PathBuf::from(path));
+            }
+        }
+        Ok(paths)
```
kchibisov, line 132: The entire pipe reading should be calloop drived.

src/platform_impl/linux/wayland/seat/dnd/mod.rs
```diff
@@ -0,0 +1,32 @@
+mod handlers;
```
kchibisov, line 1: `mod` must be bellow all `use`, akin how it's done in other files.

src/platform_impl/linux/wayland/seat/dnd/handlers.rs
```diff
@@ -0,0 +1,136 @@
...
+                        device_id: crate::event::DeviceId(crate::platform_impl::DeviceId::Wayland(
+                            DeviceId,
+                        )),
+                    },
+                    window_id,
+                );
+            }
+        }
+        DndEvent::Motion { x, y, .. } => {
+            if let Some(window_id) = inner.window_id {
```
kchibisov, line 94: I'm not sure why it generates that event, could you elaborate?

src/platform_impl/linux/wayland/seat/dnd/handlers.rs
```diff
@@ -0,0 +1,136 @@
...
+        }
+        DndEvent::Leave => {
+            if let Some(window_id) = inner.window_id {
+                inner.window_id = None;
+
+                winit_state
+                    .event_sink
+                    .push_window_event(WindowEvent::HoveredFileCancelled, window_id);
+                winit_state.event_sink.push_window_event(
+                    WindowEvent::CursorLeft {
```
kchibisov, line 84: Also not sure how it's related, I'm pretty sure `wl_pointer` will send the right events anyway?

src/platform_impl/linux/wayland/seat/dnd/handlers.rs
```diff
@@ -0,0 +1,136 @@
...
+                        WindowEvent::CursorEntered {
+                            device_id: crate::event::DeviceId(
+                                crate::platform_impl::DeviceId::Wayland(DeviceId),
+                            ),
+                        },
+                        window_id,
+                    );
+                    winit_state.event_sink.push_window_event(
+                        WindowEvent::CursorMoved {
+                            device_id: crate::event::DeviceId(
```
kchibisov, line 45: ditto.

src/platform_impl/linux/wayland/event_loop/state.rs
```diff
@@ -28,4 +30,7 @@ pub struct WinitState {
     /// and requests from winit's windows are being forwarded to them either via
     /// `WindowUpdate` or buffer on the associated with it `WindowHandle`.
     pub window_map: HashMap<WindowId, WindowHandle>,
+
+    /// Handle to the display, to flush events when needed.
```
kchibisov, line 34: Not sure whether you really need it? Seems like not.

kchibisov (changes requested):
See the clipboard PR I've linked to you before on how to handle pipe reading with calloop. It could also give an idea how to integrate dnd given that it has data device as well.

src/platform_impl/linux/wayland/seat/dnd/handlers.rs
SludgePhD: This flushes the `receive` command and avoids a deadlock when reading from the pipe. This seemed like a much simpler way to do that than to mess with callbacks, but I see how calloop makes that pretty straightforward.

src/platform_impl/linux/wayland/seat/dnd/handlers.rs
SludgePhD: That doesn't seem to happen at least on KDE and GNOME. Only when no file is being dragged over.

SludgePhD:
Addressed your comments

src/platform_impl/linux/wayland/seat/dnd/handlers.rs
```diff
@@ -0,0 +1,185 @@
...
+            offer: Some(offer),
+            surface,
+            x,
+            y,
+            ..
+        } => {
+            let window_id = make_wid(&surface);
+            inner.window_id = Some(window_id);
+            offer.accept(Some(MIME_TYPE.into()));
+            let _ = parse_offer(&inner.loop_handle, offer, move |paths, winit_state| {
```
kchibisov, line 30: None of these is needed.

kchibisov (changes requested):
Unfortunatelly this doesn't work for me on sway and I think the reason is that serial we pass in `accept` is `zero`.

The issue with sway that there's no `drop` event. The mime types do match, etc.

Given that it's likely broken due to sctk bug I can't merge it, unless there's a way to make it work with e.g. thunar + alacritty on sway without patching gtk. Other applications do work here, but they pass normal serial.

src/platform_impl/linux/wayland/seat/dnd/handlers.rs
SludgePhD: The callback emits the `HoveredFile` event though?

SludgePhD:
sway appears to (correctly) send CRLF-delimited URI lists, while GNOME and KDE only use `\n` as a delimiter. I've pushed a fix for that, which makes things work correctly on sway for me.

kchibisov:
>sway appears to (correctly) send CRLF-delimited URI lists, while GNOME and KDE only use \n as a delimiter. I've pushed a fix for that, which makes things work correctly on sway for me.

My problem that I don't see `done` event at all. It's not like something is crlf or anything. Also sway has nothing to do with crlf, that's what applications set. Could you try sway master branch?

src/platform_impl/linux/wayland/seat/dnd/handlers.rs
kchibisov: Hm, yeah. Sure. I'm not sure why anyone would need it. but ok.

SludgePhD:
Huh, right, that must be thunar's doing then. I'll check if I can repro with sway master.

SludgePhD:
Sorry for the confusion, it reproduces even with the release of sway that I was using, the `\r` issue just distracted me a bit.

I've patched sctk to pass the correct serial to the compositor but it still does not work (https://github.com/Smithay/client-toolkit/commit/bef07ea62c6b1367c2588bb2f3c4a0e31951fd39). The sway debug log is also not very helpful, it only contains this when I try to drag&drop:

```
00:00:59.201 [wlr] [types/data_device/wlr_data_source.c:189] Offering additional MIME type after wl_data_device.set_selection
00:00:59.201 [wlr] [types/data_device/wlr_data_source.c:196] Ignoring duplicate MIME type offer text/uri-list
00:00:59.201 [wlr] [types/data_device/wlr_data_source.c:189] Offering additional MIME type after wl_data_device.set_selection
```

SludgePhD:
Sway also appears to send *multiple* hover enter events in sequence, which end up in the application as multiple `HoveredFile` events without any intervening `HoveredFileCancelled` or `DroppedFile`. I think this might be a sway or wlroots bug?

kchibisov:
But it works for me with any other application, so it's confusing.

valadaptive:
As mentioned in #1881, this should work since the relevant GNOME bug has been fixed. Once #4079 is merged I may take a look at this and #4009, and see if I can get DnD working on Wayland.

kchibisov:
@SludgePhD
@PeakKS 

Given that `gnome` issue seems like was backported for a while, I'm fine with proceeding with the said PR, since it _shouldn't blow up_. Either I will be the one updating _this_ version (it has correct event loop integrations out of all other attempts iirc, and generally, I don't think anything internally changed that much from the time this was opened) or you can pick it up if someone wants so.

kchibisov:
Just in case @tymmesyde due to also submitting wayland dnd PR with slightly a different impl.

SludgePhD:
@kchibisov feel free to take over this PR, I don't currently need this functionality

src/platform_impl/linux/wayland/seat/dnd/handlers.rs
eira-fransham: It's not specified in the Wayland protocol whether `wl_pointer` events are still emitted while a drag operation is in progress, but the existence of `enter`, `leave`, and `motion` on `wl_data_device` (see [protocol docs](https://wayland.app/protocols/wayland#wl_data_device)) appear to imply that the intention is for pointer and data device to be mutually exclusive.

src/platform_impl/linux/wayland/seat/dnd/handlers.rs
eira-fransham: I'm researching cross-platform drag-and-drop for Slint's winit backend, and all platforms separate the cursor position for a drag-and-drop operation from the "regular" cursor position for operations such as hover and click. I don't think this should be handled here, it should be a separate event. Downstream UI frameworks can merge the two kinds of cursor position events if they want, but from the perspective of winit I think it makes more sense to emulate what the platform APIs do. At least in cases where the platform APIs agree, like they do for drag-and-drop cursor position.

ogoffart:
superseded by #4571

Issue #2093, opened 2021-12-13
Separate the concept of "platform" and "GUI toolkit"


madsmtm:
`raw-window-handle` recently changed their API to use the names of the toolkit instead of the platform. E.g. `Windows -> Win32` and `MacOS -> AppKit`.

This is done because the mapping is not 1:1, an example in `winit` is Wayland vs. X11 (X11 can technically be used on MacOS as well), another prominent example is that UIKit applications can be used on MacOS using Mac Catalyst. Additionally, AppKit applications can technically also run under non-macos systems using [GNUStep](http://www.gnustep.org/) (though `winit` likely won't work out-of-the-box for this), and I suspect similar compatibility libraries exist for the other toolkits.

`winit` is only built to handle the most common platform/toolkit combinations, but maybe we should start to change this? It would allow using `winit` in more situations, and would also open up for adding support for choosing Qt or GTK (the fairly popular [`druid-shell`](https://github.com/linebender/druid/tree/master/druid-shell) and the fork [`tao`](https://github.com/tauri-apps/tao) both use GTK, if that's any indication) as one's backing toolkit.

Concretely, we could begin by renaming things:
- Documentation ("Platform-specific" section -> "Toolkit-specific" section)
- Extension traits (`WindowExtIOS` -> `WindowExtUiKit`)
- Directory/file structure externally (e.g. `platform::macos` -> `toolkit::appkit`)
- Directory/file structure internally (e.g. `platform_impl/windows` -> `toolkit_impl/win32`)

In the future we could then expose features to enable each toolkit, with sensible defaults set per platform (see https://github.com/rust-lang/cargo/issues/1197#issuecomment-268203727).

Downsides include teachability (most users probably don't care about the distinction), `git` conflicts (there are a lot of open PRs that would potentially be affected by a rename) and churn (any usage of the `platform` module would have to change).

maroider:
Making platform support be less tied to the compilation target is certainly an intriguing idea, but that would seemingly require rust-lang/cargo#1197 to be implemented instead of being a feature request.

As for renaming things in the docs, well, I'm not sure "GUI toolkit" is the best name here. Wayland certainly doesn't fit the bill. "Platform" seems neutral enough to me, and doesn't necessarily imply that it's *strictly* tied to an OS (although a "platform" is *usually* tied to one OS).

[...]

PR #2430, opened 2022-08-13
Make `winit` completely platform-agnostic

madsmtm:
Builds upon: https://github.com/rust-windowing/winit/pull/2421.
Related: https://github.com/rust-windowing/winit/issues/2093

A great thing about `winit` is how it supports a varied number of different platforms. I would like to make it even easier to add support for new platforms. Possible examples include: Mac Catalyst, GNUStep, X11 on macOS, SwiftUI, GTK, KMS/DRM, WinRT, and whatever new thing that vendors may decide we should start using.

We already have code in `src/platform_impl/linux/mod.rs` to handle the fact that both X11 and Wayland can be present on the same target. While some of the above examples can be bolted on to that existing system, I really feel we need a generic way to handle the cases where more than one platform may be available on a specific target. This PR is an attempt at that.

The basic idea is to take the `x11_or_wayland!` macro, extend it a bit, and then handle delegation to the desired platform at the top level instead (e.g. inside `src/event_loop.rs`, `src/window.rs`, ...).

I haven't implemented this everywhere yet, wanted to get some feedback first, but compare `src/platform_impl/linux/mod.rs` and `src/event_loop.rs` in commit [`e74b801`](https://github.com/rust-windowing/winit/pull/2430/commits/e74b8012b1359cffad81312cd050749e630790be) to get a feel for what this would look like.

In the end, we would end up with the following directory structure:
```
src/platform_impl/
    android/
    ios/
    macos/
    unix/utils.rs // Shared utility functions between Unix platforms, if needed
    wayland/
    web/
    windows/
    x11/
    mod.rs
```

(This would also make Wayland and X11 feel a lot less like second-class citizens).

- [ ] Updated documentation

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -12,14 +12,15 @@
 /src/platform/ios.rs                @francesca64
 /src/platform_impl/ios              @francesca64
 
-# Unix in general
-/src/platform/unix.rs               @kchibisov
+# Unix
 /src/platform_impl/linux/mod.rs     @kchibisov
 
 # Wayland
+/src/platform/wayland.rs            @kchibisov
 /src/platform_impl/linux/wayland    @kchibisov
 
 # X11
+/src/platform/x11.rs                @kchibisov
 /src/platform_impl/linux/x11        @kchibisov
 
 # macOS
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- **Breaking:** Split the `platform::unix` module into `platform::x11` and `platform::wayland`. The extension types are similarly renamed.
 - On macOS, fixed touch phase reporting when scrolling.
 - On X11, fix min, max and resize increment hints not persisting for resizable windows (e.g. on DPI change).
 - On Windows, respect min/max inner sizes when creating the window.
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -9,6 +9,7 @@
 //! handle events.
 use std::marker::PhantomData;
 use std::ops::Deref;
+use std::rc::Rc;
 use std::{error, fmt};
 
 use instant::Instant;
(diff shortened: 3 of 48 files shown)


rockisch:
While looking for WinRT support, I one way or another ended up in this PR, and although I understand there are no current plans to do it, would be nice if the new structure supported future work on it.

With that in mind, based on your initial comment, wouldn't it make more sense to call `windows` folder `win32`, so that in a possible future winit uses `windows` in the same vein `unix` is planned to be used?

On another totally separate note, since this will move quite a lot of files, maybe it would be a good idea to adopt the 'newer module system' for stuff inside `platform_impl`? At the end of the day it's purely a preference choice, but I feel more people are growing used to it than using `mod.rs`s.

thatsofia:
Isn't WinRT/UWP deprecated now? What's the point in renaming things for a dead platform?

rockisch:
> Isn't WinRT/UWP deprecated now? What's the point in renaming things for a dead platform?

...is it? Genuine question, I'm not that into Windows GUI development, so I don't really get what Microsoft's plan is behind the whole 'App SDK' stuff.

It seems like the Windows Store only opened doors for Win32 apps last year tho, so unless they deprecated WinRT in the meantime, I feel like it should still be relevant.

ids1024:
I think UWP is basically deprecated, "WinRT" isn't, and there's work on better integrating it with win32 stuff? As part of Microsoft's "Project Reunion".

But honestly I don't entirely follow exactly what's going on with these APIs or exactly how Microsoft is using all of these terms.

thatsofia:
Yes, many Microsoft employees have said so on Microsoft's own GitHub repos and on Twitter. Unfortunately Microsoft themselves haven't made any actual announcements as far as I know, but it's clear that it's now deprecated. I myself got burned by investing in UWP. 

"Project Reunion" has seemingly nothing to do with UWP and confirms the deprecation of UWP. Project Reunion is all about win32 APIs: WebView2, WinUI 3.0, GameInput.

daxpedda:
We briefly touched this on IRC, but my preference would be to move all backends entirely out of the Winit crate and into separate crates (still a monorepo).

Though I'm not against this PR and don't mind approving it, it's probably pointless if we have consensus on the separate crates idea.

madsmtm:
As stated, we're going with [another idea](https://github.com/rust-windowing/winit/issues/3433).

PR #2438, opened 2022-08-23
Update `sctk-adwaita`

### LINKED ISSUE #2373: winit suddenly requires `cmake` dependency

daxpedda:
`sctk-adwaita` now uses [`ab_glyph`](https://crates.io/crates/ab_glyph) instead of [`crossfont`](https://crates.io/crates/crossfont). As far as I can tell this isn't a breaking change.

I'm not sure if some users would prefer `crossfont` over `ab_glyph`. Alternatively we could add another crate feature for that? I think it would have been better to remove `wayland-csd-adwaita-notitle` and add `wayland-csd-adwaita-ab-glyph` and `wayland-csd-adwaita-crossfont` instead, but it's too late for that now.

Fixes #2373.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On Windows, added `WindowExtWindows::set_undecorated_shadow` and `WindowBuilderExtWindows::with_undecorated_shadow` to draw the drop shadow behind a borderless window.
 - On Windows, fixed default window features (ie snap, animations, shake, etc.) when decorations are disabled.
 - On macOS, add support for two-finger touchpad magnification and rotation gestures with new events `WindowEvent::TouchpadMagnify` and `WindowEvent::TouchpadRotate`.
+- On Wayland, `wayland-csd-adwaita` now uses `ab_glyph` instead of `crossfont` to render the title for decorations.
+- On Wayland, a new `wayland-csd-adwaita-crossfont` feature was added to use `crossfont` instead of `ab_glyph` for decorations.
 
 # 0.27.2 (2022-8-12)
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,6 +39,7 @@ x11 = ["x11-dl", "mio", "percent-encoding", "parking_lot"]
 wayland = ["wayland-client", "wayland-protocols", "sctk"]
 wayland-dlopen = ["sctk/dlopen", "wayland-client/dlopen"]
 wayland-csd-adwaita = ["sctk-adwaita", "sctk-adwaita/ab_glyph"]
+wayland-csd-adwaita-crossfont = ["sctk-adwaita", "sctk-adwaita/crossfont"]
 wayland-csd-adwaita-notitle = ["sctk-adwaita"]
 
 [dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,7 @@ default = ["x11", "wayland", "wayland-dlopen", "wayland-csd-adwaita"]
 x11 = ["x11-dl", "mio", "percent-encoding", "parking_lot"]
 wayland = ["wayland-client", "wayland-protocols", "sctk"]
 wayland-dlopen = ["sctk/dlopen", "wayland-client/dlopen"]
-wayland-csd-adwaita = ["sctk-adwaita", "sctk-adwaita/title"]
+wayland-csd-adwaita = ["sctk-adwaita", "sctk-adwaita/ab_glyph"]
 wayland-csd-adwaita-notitle = ["sctk-adwaita"]
 
 [dependencies]
@@ -105,7 +105,7 @@ features = [
 wayland-client = { version = "0.29.4", default_features = false,  features = ["use_system_lib"], optional = true }
 wayland-protocols = { version = "0.29.4", features = [ "staging_protocols"], optional = true }
 sctk = { package = "smithay-client-toolkit", version = "0.16.0", default_features = false, features = ["calloop"],  optional = true }
-sctk-adwaita = { version = "0.4.1", optional = true }
+sctk-adwaita = { version = "0.5.1", default_features = false, optional = true }
 mio = { version = "0.8", features = ["os-ext"], optional = true }
 x11-dl = { version = "2.18.5", optional = true }
 percent-encoding = { version = "2.0", optional = true }
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,7 +44,6 @@ jobs:
     env:
       RUST_BACKTRACE: 1
       CARGO_INCREMENTAL: 0
-      PKG_CONFIG_ALLOW_CROSS: 1
       RUSTFLAGS: "-C debuginfo=0 --deny warnings"
       OPTIONS: ${{ matrix.platform.options }}
       FEATURES: ${{ format(',{0}', matrix.platform.features ) }}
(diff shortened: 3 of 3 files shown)


kchibisov (changes requested):
We could add as an option, but I'd prefer crossfont to have native looks of the fonts instead of whatever `ab_glyph` is using. I agree that we should change default for that. It also uses some random hardcoded font instead of system font, like crossfont does.

But for default rust windowing library using rust font renderer is likely easier to test and cross build since most of the users don't care.

Also you should update the CI script to remove all the extra deps installing.

Cargo.toml
```diff
@@ -105,7 +105,7 @@ features = [
 wayland-client = { version = "0.29.4", default_features = false,  features = ["use_system_lib"], optional = true }
 wayland-protocols = { version = "0.29.4", features = [ "staging_protocols"], optional = true }
 sctk = { package = "smithay-client-toolkit", version = "0.16.0", default_features = false, features = ["calloop"],  optional = true }
-sctk-adwaita = { version = "0.4.1", optional = true }
+sctk-adwaita = { version = "0.5.0", optional = true }
```
kchibisov, line 108: Please, specify the latest versions.

daxpedda: Why though? What's wrong with users being able to downgrade if they want to?

Anyway, done.

kchibisov: Because it's the latest version you're writing the patch against? So we know that it works, but we don't know that `0.5.0` works. In general you should use the dependency you're writing the patch against.

Cargo.toml
```diff
@@ -38,7 +38,7 @@ default = ["x11", "wayland", "wayland-dlopen", "wayland-csd-adwaita"]
 x11 = ["x11-dl", "mio", "percent-encoding", "parking_lot"]
 wayland = ["wayland-client", "wayland-protocols", "sctk"]
 wayland-dlopen = ["sctk/dlopen", "wayland-client/dlopen"]
-wayland-csd-adwaita = ["sctk-adwaita", "sctk-adwaita/title"]
+wayland-csd-adwaita = ["sctk-adwaita", "sctk-adwaita/ab_glyph"]
```
kchibisov, line 41: There should be a cross font feature as well.

Cargo.toml
daxpedda: I was actually writing the patch against 0.5.0, not 0.5.1. In any case, I really don't have a stack in this, I was going to use the newest version anyway.

Cargo.toml
daxpedda: Done!

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On Windows, added `WindowExtWindows::set_undecorated_shadow` and `WindowBuilderExtWindows::with_undecorated_shadow` to draw the drop shadow behind a borderless window.
 - On Windows, fixed default window features (ie snap, animations, shake, etc.) when decorations are disabled.
 - On macOS, add support for two-finger touchpad magnification and rotation gestures with new events `WindowEvent::TouchpadMagnify` and `WindowEvent::TouchpadRotate`.
+- On Wayland, `wayland-csd-adwaita` now uses `ab_glyph` instead of `crossfont`. A new `wayland-csd-adwaita-crossfont` feature was added to continue using `crossfont` instead.
```
kchibisov, line 14: Could you split those into 2 messages.

kchibisov (changes requested):

CHANGELOG.md
daxpedda: Done.

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On Windows, added `WindowExtWindows::set_undecorated_shadow` and `WindowBuilderExtWindows::with_undecorated_shadow` to draw the drop shadow behind a borderless window.
 - On Windows, fixed default window features (ie snap, animations, shake, etc.) when decorations are disabled.
 - On macOS, add support for two-finger touchpad magnification and rotation gestures with new events `WindowEvent::TouchpadMagnify` and `WindowEvent::TouchpadRotate`.
+- On Wayland, `wayland-csd-adwaita` now uses `ab_glyph` instead of `crossfont` to render the title for fallback CSD.
+- On Wayland, a new `wayland-csd-adwaita-crossfont` feature was added to use `crossfont` instead of `ab_glyph` to render the title for fallback CSD.
```
kchibisov, line 15:
```suggestion
- On Wayland, a new `wayland-csd-adwaita-crossfont` feature was added to use `crossfont` instead of `ab_glyph` for decorations.
```

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On Windows, added `WindowExtWindows::set_undecorated_shadow` and `WindowBuilderExtWindows::with_undecorated_shadow` to draw the drop shadow behind a borderless window.
 - On Windows, fixed default window features (ie snap, animations, shake, etc.) when decorations are disabled.
 - On macOS, add support for two-finger touchpad magnification and rotation gestures with new events `WindowEvent::TouchpadMagnify` and `WindowEvent::TouchpadRotate`.
+- On Wayland, `wayland-csd-adwaita` now uses `ab_glyph` instead of `crossfont` to render the title for fallback CSD.
```
```suggestion
- On Wayland, `wayland-csd-adwaita` now uses `ab_glyph` instead of `crossfont` to render the title for decorations.
```

kchibisov (changes requested):

CHANGELOG.md
daxpedda: Done.

CHANGELOG.md
daxpedda: I think you meant to remove the "to render the title" part too?

kchibisov: Nah, title is fine, since that's how they differ.

kchibisov (approved):
Should be fine that way, I guess.

CHANGELOG.md
daxpedda: Sure, that makes sense.
Done!

Cargo.toml
```diff
@@ -105,7 +106,7 @@ features = [
 wayland-client = { version = "0.29.4", default_features = false,  features = ["use_system_lib"], optional = true }
 wayland-protocols = { version = "0.29.4", features = [ "staging_protocols"], optional = true }
 sctk = { package = "smithay-client-toolkit", version = "0.16.0", default_features = false, features = ["calloop"],  optional = true }
-sctk-adwaita = { version = "0.4.1", optional = true }
+sctk-adwaita = { version = "0.5.1", optional = true }
```
alexheretic, line 109: `ab_glyph` [is sctk-adwaita default](https://github.com/PolyMeilex/sctk-adwaita/blob/master/Cargo.toml#L23) so you'll want to disable that here to allow "notitle" to work & for crossfont users to avoid ab_glyph in the dependency tree.
```suggestion
sctk-adwaita = { version = "0.5.1", default_features = false, optional = true }
```

kchibisov: That makes sense, I had a wrong memory that it's not default. Thx.

kchibisov (changes requested):
Should add `default-features = false` to `sctk-adwaita`.

Cargo.toml
daxpedda: Done!

kchibisov (approved):

PR #2439, opened 2022-08-23
Mark new events as breaking change

kchibisov:
Adding a new enum variant is a breaking change in winit.

@madsmtm it's a breaking change for the reference.


madsmtm:
Sorry, I missed that!

PR #2440, opened 2022-08-23
Wayland: Use sctk-adwaita 0.5.1 auto theme selection

alexheretic:
On Wayland, if not otherwise specified use upstream automatic CSD theme selection.

This is available in [sctk-adwaita 0.5.1](https://github.com/PolyMeilex/sctk-adwaita/blob/master/CHANGELOG.md#051), ~so this PR requires #2438~

- ~[ ] Created or updated an example program if it would help users understand this functionality~
- ~[ ] Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - **Breaking:** On macOS, add support for two-finger touchpad magnification and rotation gestures with new events `WindowEvent::TouchpadMagnify` and `WindowEvent::TouchpadRotate`.
 - On Wayland, `wayland-csd-adwaita` now uses `ab_glyph` instead of `crossfont` to render the title for decorations.
 - On Wayland, a new `wayland-csd-adwaita-crossfont` feature was added to use `crossfont` instead of `ab_glyph` for decorations.
+- On Wayland, if not otherwise specified use upstream automatic CSD theme selection.
 
 # 0.27.2 (2022-8-12)
 
--- a/src/platform_impl/linux/wayland/window/mod.rs
+++ b/src/platform_impl/linux/wayland/window/mod.rs
@@ -170,16 +170,15 @@ impl Window {
             )
             .map_err(|_| os_error!(OsError::WaylandMisc("failed to create window.")))?;
 
+        // Set CSD frame config from theme if specified,
+        // otherwise use upstream automatic selection.
         #[cfg(feature = "sctk-adwaita")]
-        {
-            // Set CSD frame config from theme if specified,
-            // otherwise use upstream automatic selection.
-            if let Some(theme) = platform_attributes
-                .csd_theme
-                .or_else(|| std::env::var(WAYLAND_CSD_THEME_ENV_VAR).try_into().ok())
-            {
-                window.set_frame_config(theme.into());
-            }
+        if let Some(theme) = platform_attributes.csd_theme.or_else(|| {
+            std::env::var(WAYLAND_CSD_THEME_ENV_VAR)
+                .ok()
+                .and_then(|s| s.as_str().try_into().ok())
+        }) {
+            window.set_frame_config(theme.into());
         }
 
         // Set decorations.
@@ -652,12 +651,5 @@ impl TryFrom<&str> for Theme {
         } else {
             Err(())
         }
-    }
-}
-impl TryFrom<Result<String, std::env::VarError>> for Theme {
-    type Error = ();
-
-    fn try_from(env: Result<String, std::env::VarError>) -> Result<Self, Self::Error> {
-        env.map_err(|_| ())?.as_str().try_into()
     }
 }
--- a/src/platform_impl/linux/wayland/window/mod.rs
+++ b/src/platform_impl/linux/wayland/window/mod.rs
@@ -170,18 +170,20 @@ impl Window {
             )
             .map_err(|_| os_error!(OsError::WaylandMisc("failed to create window.")))?;
 
-        // Set CSD frame config
+        // Set CSD frame config from theme if specified
         #[cfg(feature = "sctk-adwaita")]
         {
-            let theme = platform_attributes.csd_theme.unwrap_or_else(|| {
-                let env = std::env::var(WAYLAND_CSD_THEME_ENV_VAR).unwrap_or_default();
-                match env.to_lowercase().as_str() {
-                    "dark" => Theme::Dark,
-                    _ => Theme::Light,
+            let theme = platform_attributes.csd_theme.or_else(|| {
+                match std::env::var(WAYLAND_CSD_THEME_ENV_VAR) {
+                    Ok(t) if t.eq_ignore_ascii_case("dark") => Some(Theme::Dark),
+                    Ok(t) if t.eq_ignore_ascii_case("light") => Some(Theme::Light),
+                    _ => None,
                 }
             });
-
-            window.set_frame_config(theme.into());
+            // If not specified use upstream default (automatically selects theme)
+            if let Some(theme) = theme {
+                window.set_frame_config(theme.into());
+            }
         }
 
         // Set decorations.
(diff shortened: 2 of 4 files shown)


src/platform_impl/linux/wayland/window/mod.rs
```diff
@@ -170,18 +170,20 @@ impl Window {
...
-            let theme = platform_attributes.csd_theme.unwrap_or_else(|| {
-                let env = std::env::var(WAYLAND_CSD_THEME_ENV_VAR).unwrap_or_default();
-                match env.to_lowercase().as_str() {
-                    "dark" => Theme::Dark,
-                    _ => Theme::Light,
+            let theme = platform_attributes.csd_theme.or_else(|| {
+                match std::env::var(WAYLAND_CSD_THEME_ENV_VAR) {
+                    Ok(t) if t.eq_ignore_ascii_case("dark") => Some(Theme::Dark),
+                    Ok(t) if t.eq_ignore_ascii_case("light") => Some(Theme::Light),
+                    _ => None,
```
kchibisov, line 180: I'm not sure why change it that way. You should add explicit `"light"` and return `None` from catch-all.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On Windows, added `WindowExtWindows::set_undecorated_shadow` and `WindowBuilderExtWindows::with_undecorated_shadow` to draw the drop shadow behind a borderless window.
 - On Windows, fixed default window features (ie snap, animations, shake, etc.) when decorations are disabled.
 - On macOS, add support for two-finger touchpad magnification and rotation gestures with new events `WindowEvent::TouchpadMagnify` and `WindowEvent::TouchpadRotate`.
+- On Wayland, if not otherwise specified use upstream automatic CSD theme selection.
```
kchibisov, line 14: I don't think that it matters. It was automatic for the users already? And upstream method assumes that you have portal(which I don't for example).

kchibisov (changes requested):

CHANGELOG.md
alexheretic: Previously it defaulted to "light", in what way was it automatic? 

The new functionality upstream will use system settings to select theme. e.g. for me on gnome it will automatically use "dark" as indicated by dbus. If no dbus / dbus system preference, upstream continues to pick "light". We could add more methods of fetching system theme preference upstream, if they exist.

src/platform_impl/linux/wayland/window/mod.rs
alexheretic: > You should add explicit "light" and return None from catch-all.

Isn't that what is already written? The change allows the `None` path to use upstream defaults, instead of always setting.

I kept existing case insensitivity, just used `eq_ignore_ascii_case` instead of `to_lowercase` for efficiency.

kchibisov: Then you can do something like

```rust
let mut env_var = std::env::var(WAYLAND_CSD_THEME_ENV_VAR).unwrap_or_default();
// Can modify only first 5 chars with `chars` and take, but it shouldn't matter.
env_var.make_ascii_lowercase();
match env_var {
  "dark" => Some(Theme::Dark),
  "light" => Some(Theme::Light),
  _ => None,
}
```

So you'd have 1 conversion and one allocation, unlike 2 conversions like you have now?

They are pretty much the same though. Maybe I just don't like how `Ok(t) if t.eq_ignore_ascii_case("dark")` looks.

CHANGELOG.md
kchibisov: It was automatic in a way that there were no explicitly stated default, so whatever winit was using for the default was up to winit. So in that regard I'm not sure that it should be in `CHANGELOG` at all, given that there's no change in user facing docs compared to previous version or breaking change, since if you were relied on winit using the light theme you were relying on implementation behavior.

We could still keep it like that, I don't really care, just don't want to have overly long changelog in the end.

alexheretic: I put it in the change log because it's a change that may be interesting to winit users. For example, if downstream had been setting the theme themselves because it didn't match the system dark-preference they could read this and see if the automatic selection satisfies their use case.

Isn't this the kind of thing a changelog is for?

kchibisov: I guess it's fine then.

We in alacritty has our own setting that you can control and live update via IPC commands.

src/platform_impl/linux/wayland/window/mod.rs
alexheretic: Your version just seems less direct to me and more code. `eq_ignore_ascii_case` avoids any byte conversion when the len doesn't match so I think the code is 2 comparisons max 1 of which can iterate over max 5 bytes (to compare to "light").

This kind of analysis is almost certainly not worth the effort, though can be interesting. Originally removing the previously existing 2nd allocation (`to_lowercase`) seemed worthwhile since, at the time, it seemed not much effort.

:shrug: Overall not a big deal, I don't understand the dislike of match guards, but feel free to edit it to whatever you like!

kchibisov: Could go with that then. My point is to have something more readable for the code that is cold and never executed, so it's easier to grasp it.

```rust
let env_var = std::env::var(WAYLAND_CSD_THEME_ENV_VAR).unwrap_or_default();
if env_var.eq_ignore_ascii_case("dark") {
    Some(Theme::Dark)
} else if env_var.eq_ignore_ascii_case("light") {
    Some(Theme::Light)
} else {
    None
}
```

kchibisov: Oh, besides that. Sorry for bothering you, but could you add a method like `try_into` for winit `Theme` here, so we can simply write.

```rust
// If not specified use upstream default (automatically selects theme)
if let Some(theme) = platform_attributes.csd_theme.try_into() {
    window.set_frame_config(theme.into());
}
```

alexheretic: I think a `try_into` on `csd_theme` option that also checks an env-var may be a bit surprising. I've refactored to add an use a `impl TryFrom<&str> for Theme` though, wydt?

src/window.rs
```diff
@@ -1282,6 +1282,26 @@ pub enum Theme {
     Dark,
 }
 
+impl TryFrom<&str> for Theme {
```
kchibisov, line 1285: I think it should be `Wayland` backend only for now?

src/platform_impl/linux/wayland/window/mod.rs
```diff
@@ -170,18 +170,19 @@ impl Window {
...
+            // otherwise use upstream automatic selection.
+            if let Some(theme) = platform_attributes.csd_theme.or_else(|| {
+                std::env::var(WAYLAND_CSD_THEME_ENV_VAR)
+                    .ok()?
+                    .as_str()
+                    .try_into()
+                    .ok()
+            }) {
+                window.set_frame_config(theme.into());
+            }
```
kchibisov, line 185: You can shorten it like that, I think. I'm also not sure that you need `as_str` given that there's a `deref` impl.

```suggestion
            if let Ok(theme) = platform_attributes.csd_theme.ok_or_else(|| {
                std::env::var(WAYLAND_CSD_THEME_ENV_VAR).and_then(|var| var.as_str().try_into())
            }) {
                window.set_frame_config(theme.into());
            }
```

kchibisov (changes requested):

src/platform_impl/linux/wayland/window/mod.rs
alexheretic: Ok I've shortened the original commit's 7 extra lines to 30 extra lines now. Hopefully that's short enough now :laughing:

kchibisov (approved):
I've slightly adjusted formatting, should be good to go now. Thanks.

PR #2441, opened 2022-08-24
Add support for theme on macOS, and unify theme methods

keiya01:

## How to test

- `WindowBuilderExtMacOS::with_theme` should force theme even if theme is changed by system preference
- `WindowExtMacOS::theme` should return current theme even if theme is changed.
- `ThemeChanged` event should be happen whenever theme is changed.


src/platform_impl/macos/window_delegate.rs
```diff
@@ -573,3 +597,27 @@ extern "C" fn window_did_change_occlusion_state(this: &Object, _: Sel, _: id) {
...
+extern "C" fn effective_appearance_did_changed_on_main_thread(this: &Object, _: Sel, _: id) {
+    with_state(this, |state| {
+        let theme = get_ns_theme();
+        let current_theme = state.window.upgrade().map(|w| {
+            let mut state = w.shared_state.lock().unwrap();
+            let current_theme = state.current_theme;
+            state.current_theme = theme;
+            current_theme
+        });
+        println!("Change Theme: {:?}", theme);
```
madsmtm, line 617: Remove debug logging.

madsmtm (changes requested):
Thanks, implementation looks good!

Given that we now have three platforms (macOS, Windows and Wayland) that support querying and setting themes, perhaps it would make sense to unify these, and add an example for it; but that can be done later.

madsmtm (approved):
Cool, thanks! (Will merge once CI finishes)

src/platform/macos.rs
```diff
@@ -113,6 +121,12 @@ pub trait WindowBuilderExtMacOS {
     fn with_resize_increments(self, increments: LogicalSize<f64>) -> WindowBuilder;
     fn with_disallow_hidpi(self, disallow_hidpi: bool) -> WindowBuilder;
     fn with_has_shadow(self, has_shadow: bool) -> WindowBuilder;
+    /// Sets a specific theme for the window.
+    ///
+    /// If `None` is provided, the window will use the system theme.
+    ///
+    /// The default is `None`.
+    fn with_theme(self, theme: Option<Theme>) -> WindowBuilder;
```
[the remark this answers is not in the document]
madsmtm: I note this above as well, but decided to just postpone it until someone put in the work to make a PR for it.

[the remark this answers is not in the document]
kchibisov: I mean that you don't need that much skills for that, basically grep all using `Theme` and do move.

src/platform/macos.rs
```diff
@@ -36,6 +36,9 @@ pub trait WindowExtMacOS {
 
     /// Sets whether or not the window has shadow.
     fn set_has_shadow(&self, has_shadow: bool);
+
+    /// Returns the current window theme.
+    fn theme(&self) -> Theme;
```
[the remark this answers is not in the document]
keiya01: > Could we not do that as a macOS extension? We have theme setting on like every platform

If I do this, I need to implement theme method for getting current theme on unix?
I think this feature is not implemented on unix platform.

> And should likely return Option in that case, since we could fail identify it? Like on old macOS.

This should be handled by https://github.com/rust-windowing/winit/pull/2441/files#diff-4803bb2335a17d700335730b9f62828e001faa602b9b894dff572422c737b78aR1364-R1367.

src/platform/macos.rs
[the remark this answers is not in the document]
keiya01: How do we handle wayland and x11 platform? These have different interface about theme.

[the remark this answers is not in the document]
kchibisov: X11 leave unimplemented and reference the gtk variant. Wayland, macOS, and windows should use the same `with_theme`. For the getter you should return the `None` for Wayland and windows if you don't plan to add it.

You don't have to implement anything here, just add `with_theme` on `WindowBuilder` and `theme` on `Window` that uses existing impls from the extensions.

src/window.rs
```diff
@@ -1282,6 +1304,13 @@ pub enum Theme {
     Dark,
 }
 
+impl Default for Theme {
+    #[inline]
+    fn default() -> Self {
+        Self::Light
+    }
+}
+
```
madsmtm, line 1313: I know that while most users use a light theme, I don't actually think a default makes sense here?

src/platform/x11.rs
```diff
@@ -260,6 +260,7 @@ impl WindowBuilderExtX11 for WindowBuilder {
     }
 
     #[inline]
+    #[cfg(feature = "x11")]
```
madsmtm, line 263: Nit: Remove

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- On Windows, added `WindowExtWindows::set_undecorated_shadow` and `WindowBuilderExtWindows::with_undecorated_shadow` to draw the drop shadow behind a borderless window.
+- On Windows, fixed default window features (ie snap, animations, shake, etc.) when decorations are disabled.
+- On macOS, add support for two-finger touchpad magnification and rotation gestures with new events `WindowEvent::TouchpadMagnify` and `WindowEvent::TouchpadRotate`.
```
madsmtm, line 13: Don't worry about the changelog, I'll take care of it before merge

madsmtm (changes requested):
Thanks for that, that is indeed nicer!

src/window.rs
keiya01: I fixed this here.
[fix: remove unnecessary default for theme](https://github.com/rust-windowing/winit/pull/2441/commits/42c7318ca004d8bdda7a73cb3a5cf31fb92836fb)

src/platform/x11.rs
keiya01: [refactor: remove unnecessary cfg](https://github.com/rust-windowing/winit/pull/2441/commits/f565a87b7c89505f4f907be3ec4bf314b641f109)
I fixed this here

msiglreith (approved):
A note in the changelog should be added that the `WindowExtWindows::theme` got replaced by `Window::theme` and `WindowBuilderExtWindows::with_theme` by `WindowBuilder::with_theme`

madsmtm (commented):
@kchibisov does it make sense to merge `WindowBuilderExtWayland::with_wayland_csd_theme` with `WindowBuilder::with_theme`?

kchibisov:
> @kchibisov does it make sense to merge WindowBuilderExtWayland::with_wayland_csd_theme with WindowBuilder::with_theme?

Yes. And we can add a comment that it controls only CSD. That's perfectly fine.

keiya01:
I added comment about CSD!

madsmtm (approved):
I updated the macOS impl since the changes in #2465, and while I was at it, removed the remaining `WindowBuilderExtWayland::with_wayland_csd_theme` and updated the changelog as per requested by @msiglreith

src/window.rs
```diff
@@ -923,6 +942,15 @@ impl Window {
     pub fn request_user_attention(&self, request_type: Option<UserAttentionType>) {
         self.window.request_user_attention(request_type)
     }
+
+    /// Returns the current window theme.
+    /// ## Platform-specific
```
kchibisov, line 947:
```suggestion
    /// Returns the current window theme.
    ///
    /// ## Platform-specific
```

kchibisov (changes requested):
Besides that small doc change it should be good to go, though it requires a rebase.

PR #2443, opened 2022-08-25
Document `WindowEvent::ThemeChanged` OS support

daxpedda:
`WindowEvent::ThemeChanged` is also supported on Web.


madsmtm:
This is already handled in https://github.com/rust-windowing/winit/pull/2441, so I think I'll just go with that PR. Thanks though!

PR #2444, opened 2022-08-26
Android: rework backend to use android_activity crate

### LINKED ISSUE #2299: Android: `request_redraw()` blocks `send_event()`
In android, `Window::request_redraw()` can block `UserEvent`s send by `EvenLoopProxy::send_event` to reach the event loop callback.

I think this happens because of the way that request_redraw is injected in the Looper. In `request_redraw` the request is store in the static `INTERNAL_EVENT`, and the Looper is wakened:

https://github.com/rust-windowing/winit/blob/4dd2b66aaa3a338d7569d108444da609ee675b43/src/platform_impl/android/mod.rs#L676-L679

And in `Poll::Wake`, the request is injected in the Looper, taking priority over `EventSource::User`:

https://github.com/rust-windowing/winit/blob/4dd2b66aaa3a338d7569d108444da609ee675b43/src/platform_impl/android/mod.rs#L220-L226

And `send_event` also wakes up the Looper:

https://github.com/rust-windowing/winit/blob/4dd2b66aaa3a338d7569d108444da609ee675b43/src/platform_impl/android/mod.rs#L583-L587

This means that if both calls to `loop.wake()` creates a single ` Poll::Wake`, `request_event` will always be blocking `send_event` (as happens in my device).

## Reproducing the issue

I can reproduce this bug by adding the following example to the `examples` folder: https://gist.github.com/Rodrigodd/379a24ea39c6bead98a8b5dad5e0753e

And running in android, as instructed in the README.md.
```
// add this to Cargo.toml
[[example]]
name = "android_send_event_block"
crate-type = ["cdylib"]
```
```
cargo apk run --example android_send_event_block
adb logcat RustStdoutStderr:D *:S
```

Reproducible on Android 11.

rib:
This is a follow up on the discussion here: https://github.com/rust-windowing/winit/issues/2293

This updates the Android backend to use the [android_activity crate](https://github.com/rib/android-activity) instead
of [ndk-glue](https://github.com/rust-windowing/android-ndk-rs/tree/master/ndk-glue). This solves a few issues:
1. The backend is agnostic of the application's choice of Activity base class
2. Indirect support for `GameActivity` also implies support for `AppCompatActivity` based applications (not possible with `NativeActivity`)
3. Winit is no longer responsible for handling any Java synchronization details, since these are encapsulated by the design of android_activity
4. The backend no longer depends on global / static getters for state such as the native_window() which puts it in a better position to support running multiple activities within a single Android process.
5. Redraw requests are flagged, not queued, in a way that avoids taking priority over user events (resolves #2299)

Probably the most notable change for applications is that this backend _requires_ the use of the `EventLoopBuilderExtAndroid` trait to be able to associate an `AndroidApp` with an event loop that is being built. This is due `android-activity` and this backend strictly avoiding the need for any global, static state. E.g. an application's entry point for Android may look something like:

```rust
#[cfg(target_os = "android")]
#[no_mangle]
fn android_main(app: AndroidApp) {
    use winit::platform::android::EventLoopBuilderExtAndroid;

    android_logger::init_once(android_logger::Config::default().with_min_level(log::Level::Trace));

    let event_loop = EventLoopBuilder::new().with_android_app(app).build();
    _main(event_loop);
}
```

I feel that it's reasonable to have some platform specific requirements at this point, similar to how there are special requirements for web applications. Theoretically we could consider introducing some thread local state that could still allow for multiple activities running in a single process but it seems good to avoid that if not strictly necessary.

For reference work was recently done to support building Egui, EFrame applications based on `android-activity` and for that we added a general purpose mechanism for applications to configure platform specific EventLoopBuilder options: ref: https://github.com/emilk/egui/commit/fb92434aace14d013a5cdc4683b39f2735d182ef and https://github.com/rib/android-activity/blob/5dab74466c53a764c9921c588f01a38edb2f18bc/examples/agdk-eframe/src/lib.rs#L25 (so I think in practice this requirement is manageable)

Addresses: PR https://github.com/rust-windowing/winit/pull/1892
Addresses: PR https://github.com/rust-windowing/winit/pull/2307
Addresses: PR https://github.com/rust-windowing/winit/pull/2343

Addresses: #2293
Resolves: #2299

- [x] Created or updated an example program if it would help users understand this functionality - the `android-activity` crate itself has [several examples](https://github.com/rib/android-activity/tree/main/examples)

_P.S Apologies that the patch does combine a few inter-related changes while re-working the backend (such as switching to a `sticky_exit_callback` utility function consistent with other backends) and these could potentially be split out if it's making the patch tricky to review for anyone, but otherwise I'm hoping to avoid the extra work at this point to split the patch up while it still seems reasonably OK to review 🤞_

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,6 +42,8 @@ wayland-dlopen = ["sctk/dlopen", "wayland-client/dlopen"]
 wayland-csd-adwaita = ["sctk-adwaita", "sctk-adwaita/ab_glyph"]
 wayland-csd-adwaita-crossfont = ["sctk-adwaita", "sctk-adwaita/crossfont"]
 wayland-csd-adwaita-notitle = ["sctk-adwaita"]
+android-native-activity = [ "android-activity/native-activity" ]
+android-game-activity = [ "android-activity/game-activity" ]
 
 [dependencies]
 instant = { version = "0.1", features = ["wasm-bindgen"] }
@@ -60,7 +62,7 @@ simple_logger = { version = "2.1.0", default_features = false }
 [target.'cfg(target_os = "android")'.dependencies]
 # Coordinate the next winit release with android-ndk-rs: https://github.com/rust-windowing/winit/issues/1995
 ndk = "0.7.0"
-ndk-glue = "0.7.0"
+android-activity = "0.4.0-beta.1"
 
 [target.'cfg(any(target_os = "ios", target_os = "macos"))'.dependencies]
 objc2 = "=0.3.0-beta.3"
--- a/README.md
+++ b/README.md
@@ -99,36 +99,95 @@ book].
 
 #### Android
 
-This library makes use of the [ndk-rs](https://github.com/rust-windowing/android-ndk-rs) crates, refer to that repo for more documentation.
+The Android backend builds on (and exposes types from) the [`ndk`](https://docs.rs/ndk/0.7.0/ndk/) crate.
 
-The `ndk-glue` version needs to match the version used by `winit`. Otherwise, the application will not start correctly as `ndk-glue`'s internal `NativeActivity` static is not the same due to version mismatch.
+Native Android applications need some form of "glue" crate that is responsible
+for defining the main entry point for your Rust application as well as tracking
+various life-cycle events and synchronizing with the main JVM thread.
 
-`winit` compatibility table with `ndk-glue`:
+Winit uses the [android-activity](https://github.com/rib/android-activity) as a
+glue crate (prior to `0.28` it used
+[ndk-glue](https://github.com/rust-windowing/android-ndk-rs/tree/master/ndk-glue).
 
-| winit |       ndk-glue       |
-| :---: | :------------------: |
-| 0.24  | `ndk-glue = "0.2.0"` |
-| 0.25  | `ndk-glue = "0.3.0"` |
-| 0.26  | `ndk-glue = "0.5.0"` |
-| 0.27  | `ndk-glue = "0.7.0"` |
+The version of the glue crate that your application depends on _must_ match the
+version that Winit depends on because the glue crate is responsible for your
+application's main entrypoint. If Cargo resolves multiple versions they will
+clash.
+
+`winit` glue compatibility table:
+
+| winit |       ndk-glue               |
+| :---: | :--------------------------: |
+| 0.28  | `android-activity = "0.4"`   |
+| 0.27  | `ndk-glue = "0.7"`           |
+| 0.26  | `ndk-glue = "0.5"`           |
+| 0.25  | `ndk-glue = "0.3"`           |
+| 0.24  | `ndk-glue = "0.2"`           |
+
+The recommended way to avoid a conflict with the glue version is to avoid explicitly
+depending on the `android-activity` crate, and instead consume the API that
+is re-exported by Winit under `winit::platform::android::activity::*`
 
 Running on an Android device needs a dynamic system library, add this to Cargo.toml:
 
 ```toml
-[[example]]
-name = "request_redraw_threaded"
+[lib]
+name = "main"
 crate-type = ["cdylib"]
 ```
 
-And add this to the example file to add the native activity glue:
+All Android applications are based on an `Activity` subclass and the
+`android-activity` crate is designed to support different choices for this base
+class. Your application _must_ specify the base class it needs via a feature flag:
+
+| Base Class       | Feature Flag      |  Notes  |
+| :--------------: | :---------------: | :-----: |
+| `NativeActivity` | `android-native-activity` | Built-in to Android - it is possible to use without compiling any Java or Kotlin code. Java or Kotlin code may be needed to subclass `NativeActivity` to access some platform features. It does not derive from the [`AndroidAppCompat`] base class.|
+| [`GameActivity`] | `android-game-activity`   | Derives from [`AndroidAppCompat`] which is a defacto standard `Activity` base class that helps support a wider range of Android versions. Requires a build system that can compile Java or Kotlin and fetch Android dependencies from a [Maven repository][agdk_jetpack] (or link with an embedded [release][agdk_releases] of [`GameActivity`]) |
+
+[`GameActivity`]: https://developer.android.com/games/agdk/game-activity
+[`GameTextInput`]: https://developer.android.com/games/agdk/add-support-for-text-input
+[`AndroidAppCompat`]: https://developer.android.com/reference/androidx/appcompat/app/AppCompatActivity
+[agdk_jetpack]: https://developer.android.com/jetpack/androidx/releases/games
+[agdk_releases]: https://developer.android.com/games/agdk/download#agdk-libraries
+[Gradle]: https://developer.android.com/studio/build
+
+For example, add this to Cargo.toml:
+```toml
+winit = { version = "0.28", features = [ "android-native-activity" ] }
+
+[target.'cfg(target_os = "android")'.dependencies]
+android_logger = "0.11.0"
+```
+
+And, for example, define an entry point for your library like this:
 ```rust
-#[cfg_attr(target_os = "android", ndk_glue::main(backtrace = "on"))]
-fn main() {
-    ...
+#[cfg(target_os = "android")]
+use winit::platform::android::activity::AndroidApp;
+
+#[cfg(target_os = "android")]
+#[no_mangle]
+fn android_main(app: AndroidApp) {
+    use winit::platform::android::EventLoopBuilderExtAndroid;
+
+    android_logger::init_once(android_logger::Config::default().with_min_level(log::Level::Trace));
+
+    let event_loop = EventLoopBuilder::with_user_event()
+        .with_android_app(app)
+        .build();
+    _main(event_loop);
 }
 ```
 
-And run the application with `cargo apk run --example request_redraw_threaded`
+For more details, refer to these `android-activity` [example applications](https://github.com/rib/android-activity/tree/main/examples).
+
+##### Converting from `ndk-glue` to `android-activity`
+
+If your application is currently based on `NativeActivity` via the `ndk-glue` crate and building with `cargo apk` then the minimal changes would be:
+1. Remove `ndk-glue` from your `Cargo.toml`
+2. Enable the `"android-native-activity"` feature for Winit: `winit = { version = "0.28", features = [ "android-native-activity" ] }`
+3. Add an `android_main` entrypoint (as above), instead of using the '`[ndk_glue::main]` proc macro from `ndk-macros` (optionally add a dependency on `android_logger` and initialize logging as above).
+4. Pass a clone of the `AndroidApp` that your application receives to Winit when building your event loop (as shown above).
 
 #### MacOS
 
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -97,8 +97,18 @@ impl<T> EventLoopBuilder<T> {
     ///   `WINIT_UNIX_BACKEND`. Legal values are `x11` and `wayland`.
     ///   If it is not set, winit will try to connect to a Wayland connection, and if that fails,
     ///   will fall back on X11. If this variable is set with any other value, winit will panic.
+    /// - **Android:** Must be configured with an `AndroidApp` from `android_main()` by calling
+    ///     [`.with_android_app(app)`] before calling `.build()`.
     ///
     /// [`platform`]: crate::platform
+    #[cfg_attr(
+        target_os = "android",
+        doc = "[`.with_android_app(app)`]: crate::platform::android::EventLoopBuilderExtAndroid::with_android_app"
+    )]
+    #[cfg_attr(
+        not(target_os = "android"),
+        doc = "[`.with_android_app(app)`]: #only-available-on-android"
+    )]
     #[inline]
     pub fn build(&mut self) -> EventLoop<T> {
         static EVENT_LOOP_CREATED: OnceCell<()> = OnceCell::new();
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -2,18 +2,18 @@
 
 use std::{
     collections::VecDeque,
-    sync::{mpsc, RwLock},
+    hash::Hash,
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        mpsc, Arc,
+    },
     time::{Duration, Instant},
 };
 
-use ndk::{
-    configuration::Configuration,
-    event::{InputEvent, KeyAction, Keycode, MotionAction},
-    looper::{ForeignLooper, Poll, ThreadLooper},
-    native_window::NativeWindow,
+use android_activity::input::{InputEvent, KeyAction, Keycode, MotionAction};
+use android_activity::{
+    AndroidApp, AndroidAppWaker, ConfigurationRef, InputStatus, MainEvent, Rect,
 };
-use ndk_glue::{Event, LockReadGuard, Rect};
-use once_cell::sync::Lazy;
 use raw_window_handle::{
     AndroidDisplayHandle, HasRawWindowHandle, RawDisplayHandle, RawWindowHandle,
 };
@@ -22,35 +22,10 @@ use crate::platform_impl::Fullscreen;
 use crate::{
     dpi::{PhysicalPosition, PhysicalSize, Position, Size},
     error,
-    event::{self, VirtualKeyCode},
-    event_loop::{self, ControlFlow},
+    event::{self, StartCause, VirtualKeyCode},
+    event_loop::{self, ControlFlow, EventLoopWindowTarget as RootELW},
     window::{self, CursorGrabMode, Theme},
 };
-
-static CONFIG: Lazy<RwLock<Configuration>> = Lazy::new(|| {
-    RwLock::new(Configuration::from_asset_manager(
-        #[allow(deprecated)] // TODO: rust-windowing/winit#2196
-        &ndk_glue::native_activity().asset_manager(),
-    ))
-});
-// If this is `Some()` a `Poll::Wake` is considered an `EventSource::Internal` with the event
-// contained in the `Option`. The event is moved outside of the `Option` replacing it with a
-// `None`.
-//
-// This allows us to inject event into the event loop without going through `ndk-glue` and
-// calling unsafe function that should only be called by Android.
-static INTERNAL_EVENT: Lazy<RwLock<Option<InternalEvent>>> = Lazy::new(|| RwLock::new(None));
-
-enum InternalEvent {
-    RedrawRequested,
-}
-
-enum EventSource {
-    Callback,
-    InputQueue,
-    User,
-    Internal(InternalEvent),
-}
 
 fn ndk_keycode_to_virtualkeycode(keycode: Keycode) -> Option<event::VirtualKeyCode> {
     match keycode {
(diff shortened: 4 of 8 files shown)


rib:
I'm not sure at the moment what the best way of linking to android specific API from common docs would be (currently causing CI failures)

src/event_loop.rs
```diff
@@ -97,7 +97,10 @@ impl<T> EventLoopBuilder<T> {
     ///   `WINIT_UNIX_BACKEND`. Legal values are `x11` and `wayland`.
     ///   If it is not set, winit will try to connect to a Wayland connection, and if that fails,
     ///   will fall back on X11. If this variable is set with any other value, winit will panic.
+    /// - **Android:** Must be configured with an `AndroidApp` from `android_main()` by calling
+    ///     [`.with_android_app(app)`] before calling `.build()`.
     ///
+    /// [`.with_android_app(app)`]: crate::platform::android::EventLoopBuilderExtAndroid::with_android_app
```
madsmtm, line 103: You can do something like:
```suggestion
    #[cfg_attr(target_os = "android", doc = "[`.with_android_app(app)`]: crate::platform::android::EventLoopBuilderExtAndroid::with_android_app")]
    #[cfg_attr(not(target_os = "android"), doc = "[`.with_android_app(app)`]: #only-available-on-android")]
```

rib: ah, neat, thanks!

rib:
I wasn't aware that Winit supported compiler version 1.57.0 (the cause of the last CI failure) and although I've now updated android-activity to support being compiled with 1.57.0 I also now need to upstream patches for `cargo ndk` which doesn't support building with `1.57.0` :/

I'm wondering if we could bring the MSRV for Winit forwards a little bit? - at least to `1.58.0` which introduced support for format strings that can embed names of local variables without having to pass them as arguments. Personally I use that syntax sugar a lot - and it would also avoid the need to update `cargo ndk` too.

madsmtm:
This is a breaking change, right? In that case, we may be able to bump MSRV, since it'll probably take us 2-3 months before we'll release a new version.

(Also, remember to add a CHANGELOG entry)

torokati44:
With the MSRP bump now in, is this unblocked? I'm super stoked to be able to put the `View` of my `Activity` within a `Layout`! (I understand this will be possible with `GameActivity` now, right?)

rib:
We should be able to unblock this once I get a chance to do a 0.3 release.

I was also taking a pass at trying to add Rust bindings for GameTextInput to the API to support input methods (soft keyboards) but that hasn't gone as smoothly as I was hoping and so I may have to skip that for now (I had been thinking it would be good to cram this into 0.3 ideally)

Off the top of my head I'm not actually sure if you need GameActivity to be able to put the view of your activity in a layout. I would guess that's possible with NativeActivity too. GameActivity derives from AppCompatActivity which NativeActivity does not and there might be something about that that helps with Layout integration perhaps?

I'll try and get a chance soon to update / unblock this PR.

torokati44:
> Off the top of my head I'm not actually sure if you need GameActivity to be able to put the view of your activity in a layout. I would guess that's possible with NativeActivity too.

I tried to do so really hard, but haven't succeeded. This still doesn't rule out that it's possible, but it didn't seem easy at all.

> GameActivity derives from AppCompatActivity which NativeActivity does not and there might be something about that that helps with Layout integration perhaps?

Yes, this is also the conclusion I ended up settling on so far.

> I'll try and get a chance soon to update / unblock this PR.

Yay!

rib:
Okey, so I had a chance to update this yesterday - based on a new release of `android-activity` that will aim to maintain an MSRV in line with winit and with a CHANGLOG and README updates (e.g. summarizing how to update apps using `ndk-glue`).

I started looking at exposing `GameTextInput` support but hit some issues with that. In the mean time though I did make a change to wrap the `InputEvent` enum that's exposed by `android-activity` to help ensure it will be possible to add an IME/Text input event without needing a semver bump.

README.md
```diff
@@ -99,36 +99,77 @@ book].
...
-`winit` compatibility table with `ndk-glue`:
+Winit uses the [android-activity](https://github.com/rib/android-activity) as a glue crate (prior to `0.28` it used [ndk-glue](https://github.com/rust-windowing/android-ndk-rs/tree/master/ndk-glue).
 
-| winit |       ndk-glue       |
-| :---: | :------------------: |
-| 0.24  | `ndk-glue = "0.2.0"` |
-| 0.25  | `ndk-glue = "0.3.0"` |
-| 0.26  | `ndk-glue = "0.5.0"` |
-| 0.27  | `ndk-glue = "0.7.0"` |
+The version of the glue crate that your application depends on _must_ match the version that Winit depends on because the glue crate is responsible for your application's main entrypoint. If Cargo resolves multiple versions they will clash.
```
msiglreith, line 108: Wouldn't it be possible to just re-export the interesting parts from `android-activity` including the `native-activity`/`game-activity` flags? This would remove the requirement for users to have a compatible crate version.

src/platform_impl/android/mod.rs
```diff
@@ -216,383 +189,593 @@ fn ndk_keycode_to_virtualkeycode(keycode: Keycode) -> Option<event::VirtualKeyCo
...
+    pub fn try_recv(&mut self) -> Result<T, mpsc::TryRecvError> {
+        if let Some(first) = self.first.take() {
+            return Ok(first);
+        }
+        self.recv.try_recv()
+    }
+}
+
+#[derive(Clone)]
+struct SharedFlagSetter {
```
msiglreith, line 226: minor: imo this could be just merged into the `SharedFlag`

src/platform_impl/android/mod.rs
```diff
@@ -216,383 +189,593 @@ fn ndk_keycode_to_virtualkeycode(keycode: Keycode) -> Option<event::VirtualKeyCo
...
-                    // contains a Sender it will never disconnect, so no error handling need.
-                    while let Ok(event) = self.user_events_receiver.try_recv() {
-                        call_event_handler!(
-                            event_handler,
-                            self.window_target(),
-                            control_flow,
-                            event::Event::UserEvent(event)
-                        );
-                    }
+                InputEvent::KeyEvent(key) => {
```
msiglreith, line 553: `android-activity` seems to treat all events as handled by the upper layer similar to what `winit` is currently doing. There is an open issue to declare in particular hardware keys as not-handled to allow volume control. We would probably some way to set the `handled` flag when finishing the event.

msiglreith (commented):
thanks for looking into this topic - allowing a broader set of activities seems interesting while keeping compatibility with the base `NativeActivity`. Tested it yesterday a bit with simpler applications.

One point to discuss would probably be organization in general under the light that we would effectively offload larger implementation details into an external dependency requiring syncing for issues and releases.

src/platform_impl/android/mod.rs
torokati44: Definitely agree.

README.md
rib: NB: Android applications are always JVM based and start with a JVM `Activity` which is subclassed to be able to create "native" applications by loading a shared library via `dlopen`/`LoadLibrary`.

The shared library needs to export a standard symbol that the Activity class can find as the entry point for running the native C/C++/Rust code for the application.

The 'glue' layer/crate (e..g `ndk_glue` or `android-activity`) is what provides that symbol which is the entry point for the application, and there can technically only be a single instance of that symbol.

The ABI / details for exactly how that symbol should be defined also depends on the implementation details of the Activity subclass (such as `NativeActivity` vs `GameActivity`) which conflict and mean it's not possible to support both Activity base classes in a single share library.

The state that the glue layer works with is all bootstrapped based on state passed from the JVM Activity into that entrypoint which also means that if you were to somehow get an application to link multiple instances of the glue layer and avoid a symbol clash for the entrypoint then only one of them actually going to passed the state it needs to operate.

Conceptually this is a bit different to shared libraries where multiple version can be completely orthogonal.


Differences in `NativeActivity` and `GameActivity` lead to quite significant differences in the Rust/native glue layer, especially for input handling which currently also leads to using (exposed in the public API) completely different types for input events (for motion and key events) that aren't runtime compatible.

It's sort of comparable to how winit backends aren't designed to be runtime selectable, which is reasonable from the pov that you only need one at a time. For `android-activity` we only need to support one Activity class at a time so it only strives for API compatibility across different Activity base classes - not for dynamic, runtime API + ABI compatibility.

Hope that clarifies a bit the unique challenge with glue crates here.

src/platform_impl/android/mod.rs
rib: Right, technically the NDK InputQueue API, used with `NativeActivity` supports the notion that an application might not handle a key event in case there might be a fallback interpretation but yeah currently the `android-activity` API doesn't expose that notion.

One reason for that was that it's not something that's supported with `GameActivity`, but yeah it could be reasonable to go the other way and allow apps to report events as 'handled' or not and just ignore that with `GameActivity`.

This will require a semver bump, but should be a fairly simple change.

MarijnS95: Better use a larger rework to solve these issues from the get-go instead of having to rework/refactor again

#1919 / #2432 both describe this issue.

> This will require a semver bump, but should be a fairly simple change.

Shouldn't your crate still be under some alpha/dev notion while working our the final API details within this PR?

README.md
MarijnS95: > and there can technically only be a single instance of that symbol.

Such a situation should lead to a linker error?

Fwiw the text quoted here is what I set out to fix with aforementioned work on `ndk-glue`, to deprecate the `ndk-context` crate again (which is holding untyped `static` global pointers), and rely on runtime passing of all the various _typed_ NDK instances/pointers. Any mismatch in crate versions would result in a compile error (unless the user somehow manages to combine them _and_ forget to initialize `winit` with said instance through an extension `trait`).

src/platform_impl/android/mod.rs
rib: > > This updates the Android backend to use the [android_activity crate](https://github.com/rib/android-activity) instead
> > of [ndk-glue](https://github.com/rust-windowing/android-ndk-rs/tree/master/ndk-glue). This solves a few issues:
> 
> I have previously requested to keep `ndk-glue` as a pure-Rust glue implementation atop `NativeActivity` and even offered to finalize a rework branch to match the new API (effectively redoing some of the code that currently lives in `winit`, solving exactly the issues that were reported later) as long as we could set up a shared place to define this interface (or toggle it purely at compile time) yet nothing seems to have been done in this regard, making this an instant no-go from me.
> 
> Thanks @msiglreith for making me aware, as I hadn't seen/expected this to make it into a PR yet.

I recall we had some discussion around wanting a pure rust glue layer which I agree would be nice to have eventually but I think I tried to say I saw that as an implementation detail that we can hopefully tackle separately. I think we are hopefully also in a better position with `android-activity` to consider incrementally re-writing the `android_native_app_glue` code - which is about 650 lines of C code.

I tried to say that I believe that the `android_native_app_glue` was a good design / foundation in the short term which has been battle tested over 10 years (for example handling synchronization with the JVM in a way that is abstracted by the glue layer instead of making the application/middleware on top be involved).

Even though the `NativeActivity` backend compiles a small amount C code, that's transparent to the public API here, so I would hope we don't have to be blocked on needing to rewrite that code in Rust right now if it doesn't really fix something?

In terms of the comments about offering to finalize a rework branch to match the new API I'm not sure what this is referring too, so sorry if we had different ideas in some earlier discussion. I tried to look over https://github.com/rust-windowing/winit/issues/2293 but I don't see this. I see these initial questions from you, which might be what you're referring to?:
> - Where does that API live?
> - Who'll write it? Who'll review and sign off on it?
> - Who'll implement/uprev ndk-glue for it?
> - Who'll make winit use the new API?

and after my first pass at implementing `android-activity` I made a proposal and specifically tried to propose an answer for each of these questions: https://github.com/rust-windowing/winit/issues/2293#issuecomment-1173216806

At this point `android-activity` is intended to be able to support a common API for multiple `Activity` classes, starting with `NativeActivity` (can eventually be pure rust) and `GameActivity` and this was arrived at in an incremental way after my first experiments with just supporting GameActivity.

In particular, in [July](https://github.com/rust-windowing/winit/issues/2293#issuecomment-1173216806) I answered your question "Who'll make winit use the new API" with:
> I think the branch needs to be rebased and re-iterated to tidy a few things up but hopefully this is a useful starting point: https://github.com/rib/winit/tree/android-activity

and I didn't see that there was any objection to my proposal or push back - you even sounded positive about some of the issues that I had tackled with its design, such as removing all global static state?

rib:
> > This updates the Android backend to use the [android_activity crate](https://github.com/rib/android-activity) instead
> > of [ndk-glue](https://github.com/rust-windowing/android-ndk-rs/tree/master/ndk-glue). This solves a few issues:
> 
> I have previously requested to keep `ndk-glue` as a pure-Rust glue implementation atop `NativeActivity` and even offered to finalize a rework branch to match the new API (effectively redoing some of the code that currently lives in `winit`, solving exactly the issues that were reported later) as long as we could set up a shared place to define this interface (or toggle it purely at compile time) yet nothing seems to have been done in this regard, making this an instant no-go from me.
> 
> Thanks @msiglreith for making me aware, as I hadn't seen/expected this to make it into a PR yet.

I recall we had some discussion around wanting a pure rust glue layer which I agree would be nice to have eventually but I think I tried to say I saw that as an implementation detail that we can hopefully tackle separately. I think we are hopefully also in a better position with `android-activity` to consider incrementally re-writing the `android_native_app_glue` code - which is about 650 lines of C code.

I tried to say that I believe that the `android_native_app_glue` was a good design / foundation in the short term which has been battle tested over 10 years (for example handling synchronization with the JVM in a way that is abstracted by the glue layer instead of making the application/middleware on top be involved).

Even though the `NativeActivity` backend compiles a small amount C code, that's transparent to the public API here, so I would hope we don't have to be blocked on needing to rewrite that code in Rust right now if it doesn't really fix something?

In terms of the comments about offering to finalize a rework branch to match the new API I'm not sure what this is referring too, so sorry if we had different ideas in some earlier discussion. I tried to look over https://github.com/rust-windowing/winit/issues/2293 but I don't see this. I see these initial questions from you, which might be what you're referring to?:
> - Where does that API live?
> - Who'll write it? Who'll review and sign off on it?
> - Who'll implement/uprev ndk-glue for it?
> - Who'll make winit use the new API?

and after my first pass at implementing `android-activity` I made a proposal and specifically tried to propose an answer for each of these questions: https://github.com/rust-windowing/winit/issues/2293#issuecomment-1173216806

At this point `android-activity` is intended to be able to support a common API for multiple `Activity` classes, starting with `NativeActivity` (can eventually be pure rust) and `GameActivity` and this was arrived at in an incremental way after my first experiments with just supporting GameActivity.

In particular, in [July](https://github.com/rust-windowing/winit/issues/2293#issuecomment-1173216806) I answered your question "Who'll make winit use the new API" with:
> I think the branch needs to be rebased and re-iterated to tidy a few things up but hopefully this is a useful starting point: https://github.com/rib/winit/tree/android-activity

and I didn't see that there was any objection to my proposal or push back - you even sounded positive about some of the issues that I had tackled with its design, such as removing all global static state?

README.md
msiglreith: @rib thanks for the detailed write up but I believe I failed to get my point across or I still miss some part:
My proposal would be that the user doesn't have to basically know anything about the underlying `android-activity` crate acting as backend for the android implementation.
```toml
[dependencies]
winit = { version = "0.x", features = ["android-native-activity"] } # internally enables the `android-activity/native-activity` feature
```

instead of  
```toml
[dependencies]
winit = "0.x"
android-activity = { version = "0.y", features = ["native-activity"] }
```

Under `platform/android` public api (ie `AndroidApp`, etc) would be re-exported if necessary. I tried it locally with `NativeActivity` and this seems to work in a quick prototype given the entrypoints are specified in the `android-activity` crate with the feature flags passed trough by `winit`

src/platform_impl/android/mod.rs
rib: I've update the API for `input_events` to make it possible to convey whether an application really handled the event or not so it does not let the InputQueue handle fallbacks: https://github.com/rib/android-activity/pull/32

I'll update this branch/PR once that lands, and after making a corresponding 0.4 release.

Thanks for catching this. I realized I'd left a comment about potentially being able to support fallbacks, potentially higher in the stack for consistency across backends (maybe I was thinking Winit at the time) but having the feedback at this level makes sense to have as a base line.

rib: > Shouldn't your crate still be under some alpha/dev notion while working our the final API details within this PR?

I've been dog-fooding the android-activity crate and the winit backend (mainly based on egui usage) for some time now and it's been through a few iterations with an API that I was starting to feel was quite stable / maintainable and so I was actually expecting that the 0.3 release was going to be fine (modulo bug fixes for something that wouldn't affect the API).

Of course, even with the best laid plans we ended up finding a good reason to bump the API here so yeah in retrospect if I'd have used a pre-release tag we could still target 0.3 but bumping to 0.4 isn't really a problem.

Since the intention of the PR is to merge the new backend into master my preference would be for Winit to depend on a stable release of `android-activity`.

Question now is whether I'm optimistic and assume 0.4 should be good to go API-wise, or maybe I'll publish a pre-release this time :)

README.md
rib: My initial instinct here is that this feels like a layering violation - though I can see the appeal of having just one thing specifying a version for `android-activity` (could reduce churn in case of API bumps)

My main concern is that the glue crate fundamentally defines the ABI for the entrypoint for the application and as you noted there is the `AndroidApp` type that's passed to the application's `android_main`. I think it would seem quite odd to me that I'm getting the definition for a type passed to my application `main` from a middleware/framework crate.

The responsibilities to decide what `Activity` class is used and the ABI for the application entry point seem like they should lay directly with the application crate and everything is downstream from that.

Please also keep in mind that android-activity is intended to be usable for native Rust apps that don't necessarily use Winit, For example I'm poking at using it with OpenXR, not using Winit in that case. (So my general mindset is to associate android-activity with the orthogonal definition of the application entrypoint before considering that framework the app is built on)

I'm currently not quite sure how this would work for something like an egui-eframe or Bevy application where Winit is mostly abstracted away from the application (except there is still a requirement to pass that `AndroidApp` in to eframe for it to construct its Winit event loop).

In general if there are apps using Winit behind an abstraction then you'd still have the original requirement for the application crate to have a direct dependency on the android-activity crate - or else those abstractions would also need to re-export `AndroidApp` in a similar way.

I guess I'm not quite sure either way about this idea atm. I can see the benefit that some apps they could just defer to winit to specify the android-activity version but I can't help but find it a bit odd that a key type for the main entry point is coming from a higher-level middleware crate.

rib: > > and there can technically only be a single instance of that symbol.
> 
> Such a situation should lead to a linker error?

It depends:

In the case of `ndk-glue` it assumes you're using a proc macro that actually makes the application crate define the symbol that `NativeActivity` looks for, so there won't be a linker clash and things will only go wrong at runtime because only the version of `ndk-glue` that's linked to the application will get initialized by that entrypoint. If a crate like winit has resolved a different version then when it tries to query some state it will be interacting with an uninitialized version of `ndk-glue`.

In the case of `android-activity` there will generally be a compile-time error before linking because the feature that determines which `Activity` backend to use is unlikely to be specified for additional versions and it's a compile-time error if no Activity backend has been chosen. If you did somehow specify a backend for each version then yep you should still hit a linker error while compiling.

src/platform_impl/android/mod.rs
rib: okey, I've published a 0.4.0-beta with the `input_events` API change to allow apps to report events as handled/unhandled.

I've updated the branch for this PR to report all events as `Handled` for now

MarijnS95:
> I recall we had some discussion around wanting a pure rust glue layer which I agree would be nice to have eventually but I think I tried to say I saw that as an implementation detail that we can hopefully tackle separately. I think we are hopefully also in a better position with `android-activity` to consider incrementally re-writing the `android_native_app_glue` code - which is about 650 lines of C code.

It can be tackled separately (i.e. not conflating this PR) without temporarily forcing your C-only backend on Android + `winit` users?

> I tried to say that I believe that the `android_native_app_glue` was a good design / foundation in the short term which has been battle tested over 10 years (for example handling synchronization with the JVM in a way that is abstracted by the glue layer instead of making the application/middleware on top be involved).
> 
> Even though the `NativeActivity` backend compiles a small amount C code, that's transparent to the public API here, so I would hope we don't have to be blocked on needing to rewrite that code in Rust right now if it doesn't really fix something?

Again, no need to rewrite - it is mostly finalizing an in-progress refactor of `ndk-glue` code, and making sure it provides an implementation for the proposed API.

> In terms of the comments about offering to finalize a rework branch to match the new API I'm not sure what this is referring too, so sorry if we had different ideas in some earlier discussion. I tried to look over #2293 but I don't see this. I see these initial questions from you, which might be what you're referring to?:
> 
> > * Where does that API live?
> > * Who'll write it? Who'll review and sign off on it?
> > * Who'll implement/uprev ndk-glue for it?
> > * Who'll make winit use the new API?
> 
> and after my first pass at implementing `android-activity` I made a proposal and specifically tried to propose an answer for each of these questions: [#2293 (comment)](https://github.com/rust-windowing/winit/issues/2293#issuecomment-1173216806)

Fair enough, let's create a repo under `rust-windowing` that defines the common traits or however you propose the new API to look like, where we subsequently discuss the semantics. I'll make sure `ndk-glue` implements the API and hence serve as a drop-in replacement just like your C Native/GameActivity.

> and I didn't see that there was any objection to my proposal or push back - you even sounded positive about some of the issues that I had tackled with its design, such as removing all global static state?

I was positive about an independent contributor reporting the _exact same_ issues I had on my TODO list, giving some extra incentive to solve them at once by a big `ndk-glue` + `winit` refactor - not by deleting `ndk-glue` altogether.

MarijnS95:
Fwiw it looks like @msiglreith shares similar concerns about offloading larger chunks of the backend to external crates, that'll somehow have to synchronize on issues, PRs and releases: https://github.com/rust-windowing/winit/pull/2444#pullrequestreview-1112555981

rib:
> > I recall we had some discussion around wanting a pure rust glue layer which I agree would be nice to have eventually but I think I tried to say I saw that as an implementation detail that we can hopefully tackle separately. I think we are hopefully also in a better position with `android-activity` to consider incrementally re-writing the `android_native_app_glue` code - which is about 650 lines of C code.
> 
> It can be tackled separately (i.e. not conflating this PR) without temporarily forcing your C-only backend on Android + `winit` users?

Sorry but this seems like a disingenuous criticism at this point. The backend isn't "C-only" by a long shot but it does currently reuse some _very_ well tested C code. In the context of a low-level piece of glue for an operating system, that inherently going to involve some unsafe code (to interact with JNI at the very least) I really don't understand the hang up here. 

It seems pedantic to point out but it's perhaps also worth noting that APIs like `ALooper` and `AInputQueue` (which a pure Rust glue layer, including `ndk-glue`, would use) are also implemented in C/C++, written by the same NDK developers from Google. It seems quite arbitrary to be ok with using `AInputQueue` from Rust but not be ok with `android_native_app_glue`.

Using this C code just saved a small amount of time because the code already existed and has been widely used for a long time. The C code also naturally addressed at least one specific design issue that was noted after I initially looked at using ndk-glue - namely it encapsulates how it synchronizes with Java for saving state and creating/destroying window surfaces.

As I've noted before - this code can be re-written in Rust if we want and since it's fully encapsulated it also won't affect the public API.

As a litmus test for this issue; do you also plan on re-implementing existing NDK libraries that are written in C/C++ out of some concern that all C/C++ code should be avoided on Android? It would technically be possible, and even relatively straight forward, but probably also quite a chore.

> 
> > I tried to say that I believe that the `android_native_app_glue` was a good design / foundation in the short term which has been battle tested over 10 years (for example handling synchronization with the JVM in a way that is abstracted by the glue layer instead of making the application/middleware on top be involved).
> > Even though the `NativeActivity` backend compiles a small amount C code, that's transparent to the public API here, so I would hope we don't have to be blocked on needing to rewrite that code in Rust right now if it doesn't really fix something?
> 
> Again, no need to rewrite - it is mostly finalizing an in-progress refactor of `ndk-glue` code, and making sure it provides an implementation for the proposed API.

As far as I can tell, you're apparently upset that I either didn't submit all my work as a PR against the ndk-glue project or else upset that I didn't wait for you to do the work instead.

Fwiw I originally looked at building on ndk-glue and as I did this work incrementally in the open I've also repeatedly documented various issues I hit.

I also reached out to the ndk-glue project in April (https://github.com/rust-windowing/android-ndk-rs/issues/266) explaining my interest in using GameActivity and the brief feedback at the time was that there was more interest in creating a vendored RustActivity at some point, though there wasn't much of a clear plan to implement that yet.

> 
> > In terms of the comments about offering to finalize a rework branch to match the new API I'm not sure what this is referring too, so sorry if we had different ideas in some earlier discussion. I tried to look over #2293 but I don't see this. I see these initial questions from you, which might be what you're referring to?:
> > > * Where does that API live?
> > > * Who'll write it? Who'll review and sign off on it?
> > > * Who'll implement/uprev ndk-glue for it?
> > > * Who'll make winit use the new API?
> > 
> > 
> > and after my first pass at implementing `android-activity` I made a proposal and specifically tried to propose an answer for each of these questions: [#2293 (comment)](https://github.com/rust-windowing/winit/issues/2293#issuecomment-1173216806)
> 
> Fair enough, let's create a repo under `rust-windowing` that defines the common traits or however you propose the new API to look like, where we subsequently discuss the semantics. I'll make sure `ndk-glue` implements the API and hence serve as a drop-in replacement just like your C Native/GameActivity.

Sorry but this doesn't seem reasonable at this point.

I made and shared the android-activity repo specifically to be able to create a common API that could be Activity agnostic and I invited you to share feedback and contribute. (which you are still welcome to do)

I didn't just come up with a new API and then decide to write an implementation with an intention to make ndk-glue redundant.

The incremental process of designing the common API happened as result of building this project which supports _both_ GameActivity and NativeActivity and I currently see practical benefits to supporting multiple Activities via backends in a single repo instead of prioritizing external glue crates (comparable to in-tree Winit backends vs out-of tree Winit backends).

It's quite frustrating that you've become focused on the fact that the `NativeActivity` backend uses some C code as a way of apparently trying to discredit it. `android-activity` doesn't provide a "C Native" backend - it provides a `NativeActivity` backend. If there is a strong interest in having a pure Rust `NativeActivity` backend then we can look at re-writing that C code.

> 
> > and I didn't see that there was any objection to my proposal or push back - you even sounded positive about some of the issues that I had tackled with its design, such as removing all global static state?
> 
> I was positive about an independent contributor reporting the _exact same_ issues I had on my TODO list, giving some extra incentive to solve them at once by a big `ndk-glue` + `winit` refactor - not by deleting `ndk-glue` altogether.

I rather wish you could see the positive of finding someone else facing the same issues and being willing to work towards solving them - those people are potential collaborators.

It makes sense that android-activity ended up as a separate repo from ndk-glue on multiple levels:
1. It has a different scope: Common API over multiple Activity backends
2. It has a different API and entry point ABI that was specifically arrived at as a result of 1. The implementation of which shares something like 10 lines of code from `ndk-glue` (i.e. a PR would basically just be swapping the project with another project).
3. I'm not a maintainer for ndk-glue and when I shared my interest in supporting GameActivity very early on I got no indication there would be interest in accepting my work upstream.

rib:
For reference here I recently merged this PR for android-activity which fully ports the native-activity backend to Rust: https://github.com/rib/android-activity/pull/35. It has no affect on the public API / design etc but since multiple people have taken some issue with the fact that the backend wasn't pure Rust I figured I may as well just do the port to Rust now to avoid it being a distraction.

README.md
rib: Just to follow up again on this idea of yours @msiglreith... Since this seems like an orthogonal change we could make later; can we maybe agree to re-visit this as a separate issue/PR?

I agree that it could be an ergonomics improvement in some cases to re-export `AndroidApp` via the Winit crate if that means application crates don't need to explicitly declare their own dependence on `android-activity`.

More generally I think the information about the `android-activity` version needing to match what Winit expects is still valid (similar to how it had to match with ndk-glue) and it seems worthwhile to document this. This idea to re-export `AndroidApp` should hopefully just make it less likely that developers will get caught out by this requirement.

src/platform_impl/android/mod.rs
rib: I suppose I'll mark this one as resolved for now, since it was only noted as 'minor'. At the time of writing it was very intentional from my pov to decouple getting/setting at the type level because conceptually some code is only supposed to have the ability to set the flag and not clear it - so this uses the type system to effectively document the intent and enforce that constraint.

As noted we could combine them if we really figure it's overkill but I also had in mind that a solution like this could be used in other backends too later and in that case it'd maybe be better to keep the more explicit design - and it may be considered less overkill as a shared utility (in case the intention to limit the ability to clear the flag isn't always immediately apparent).

rib:
> thanks for looking into this topic - allowing a broader set of activities seems interesting while keeping compatibility with the base `NativeActivity`. Tested it yesterday a bit with simpler applications.
> 
> One point to discuss would probably be organization in general under the light that we would effectively offload larger implementation details into an external dependency requiring syncing for issues and releases.

Thanks @msiglreith for taking the time to test this out and esp on the constructive feedback that led to the input API fix, as well as your suggested ergonomic improvement to re-export `AndroidApp` + activity features via the Winit crate.

Regarding organization in general:

Although I'm open to e.g. moving the repo under the rust-windowing organisation I think my preference would be to wait and see if that would be helpful because I'd prefer to consider organisational changes in the context of a specific challenge/problem.

One minor reservation is that I'd like `android-activity` to be useful as a standalone glue layer without Winit (e.g. I'm experimenting with using it with OpenXR) and the `rust-windowing` organisation mostly revolves around Winit currently.

Github makes it easy to add contributors to the repo without it needing to belong to an organisation so if there are others that are interested in contributing regularly that's easy to address. My preference is to add developers I know based on their PRs/issues/discussion etc.

If there were some reason why I wouldn't be able to contribute to / maintain the project in the future then I'd of course look for a new maintainer or e.g. agree to moving the repo to a neutral organisation such as `rust-windowing` if that's what most people wanted. For example, even though I originally authored this [GPU Top](https://github.com/rib/gputop) project, I was more than happy to pass on maintenance.

In terms of release coordination my hope is that it shouldn't usually be a concern if winit sticks to stable features and release (unless we've specifically discussed coordinating / synchronizing something).

I prefer to always keep the main branch of a project stable such that a release can always be cut from `main`.

If there is some feature that's critical for Winit specifically then I would expect to only land that feature based on a branch of Winit that would validate the design + implementation. That means that by the time the feature would land in android-activity we should already know that it's stable as far as winit is concerned and that feature can immediately be rolled into a new stable release which Winit can depend on.

Does that sound reasonable?

madsmtm:
I think your points wrt. maintenance are valid, and I doubt it'll be a problem in practice - it's essentially the same thing I've done with [`objc2`](https://github.com/madsmtm/objc2), so there's some precedent.

I think you've made some good points in general, but I lack the technical knowledge to actually determine which approach (`android-activity` or `ndk-glue`) is better. I understand that you and @MarijnS95 must be frustrated by this thread, and I would like to thank you both for being very civil.

README.md
rib: I've just updated this branch to include this suggestion.

The `android-activity` API is now re-exported under `winit::platform::android::activity::*` and there are now "android-native-activity" and "android-game-activity" features that set the corresponding features in `android-activity`.

This way it's possible for application crates to avoid explicitly adding an `android-activity` dependency which could lead to a version conflict any time winit bumps its `android-activity` version.

The README reflects this too.

rib:
For reference here; the change I pushed today was to implement @msiglreith's idea of re-exporting the android-activity API so that it's possible for applications to avoid explicitly depending on `android-activity` themselves. This should help avoid version conflicts whenever Winit may need to bump what version of `android-activity` it depends on. 

This also adds "android-native-activity" and "android-game-activity" features for Winit that set the corresponding feature in `android-activity`.

I feel like at this point hopefully all feedback has been addressed?

rib:
just rebased, updated the features table and removed an out-of-date comment

rib:
Since this PR is currently blocked due to requested changes from @MarijnS95 I've re-requested review while I hope all issue have been resolved now.

rib:
Hey all, sorry to poke here but since there hasn't been any follow up here in the last week I wonder if we can agree whether this PR is ready to land?

@msiglreith since you're currently listed as the maintainer for the Android backend maybe you're able to comment?

I think that landing this could help with being able to progress other Android features as well as other cross-cutting features that affect all backends but are awkward to deal with while this big change for the Android backend is pending.

fwiw @dvc94ch who was the original author of `ndk-glue` recently gave a +1 for this work over at https://github.com/rust-windowing/android-ndk-rs/issues/266

dvc94ch (approved):
I tested this PR and it works great. It enables some cool new stuff like building apps with custom and/or multiple activities and I think it's a pragmatic way forward for the rust on android ecosystem.

msiglreith:
sorry for taking so long! I have a week off where I'll try to work on tha backlog.
User-facing interface looks fine to me, thanks for the changes. This would hopefully minimize disruption for users in case of changes.

msiglreith (commented):
Overall, I believe this goes in the right direction. I would merge this once the remaining parts are addressed and there are no further arguments against this step. Given the large change my general opinion on this topic:

pro:

- Compatible with NativeActivity -> no hard requirement on Java
- Reduce user-level API - easier to upgrade versions and less dependencies for the users
- Appears strong demand from users to support more sophisticated features
- C parts reimplemented in Rust based on reference impl, doesn't regress in this place

con: 

- Increases fragmentation of the Rust/Android ecosystem (build systems, windowing and activity handling). Stays compatible with ndk and cargo-apk, which is positive.
- Moving parts of the backend out of the organization causing syncing overhead, as well as larger changes in general comes with some risk 👀

Cargo.toml
```diff
@@ -60,7 +62,7 @@ simple_logger = { version = "2.1.0", default_features = false }
 [target.'cfg(target_os = "android")'.dependencies]
 # Coordinate the next winit release with android-ndk-rs: https://github.com/rust-windowing/winit/issues/1995
 ndk = "0.7.0"
-ndk-glue = "0.7.0"
+android-activity = "0.4.0-beta.1"
```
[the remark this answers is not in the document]
rib: The only thing I've generally been waiting for before releasing 0.4 is the green light that we're good to go here.

My original PR was based on a stable release but with the review here we picked up on the need to iterate the input API.

So just in case we spotted some other last-minute issue I went with a beta release.

I'll probably just roll the 0.4 release today while I'm not aware of any issue with the current API.

src/platform_impl/android/mod.rs
```diff
@@ -216,383 +191,593 @@ fn ndk_keycode_to_virtualkeycode(keycode: Keycode) -> Option<event::VirtualKeyCo
...
+        trace!("Mainloop iteration");
+
+        sticky_exit_callback(
+            event::Event::NewEvents(*cause),
+            self.window_target(),
+            control_flow,
+            callback,
+        );
+
+        //let mut redraw = false;
```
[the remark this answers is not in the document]
rib: yeah, it looks like I forgot to remove this when I re-worked how redraws are queued to have a persistent `pending_redraw` flag for squashing redundant redraw requests. I'll remove this and the other commented out code below too.

dvc94ch:
> Increases fragmentation of the Rust/Android ecosystem (build systems, windowing and activity handling). Stays compatible with ndk and cargo-apk, which is positive.

As far as ecosystem fragmentation goes, xbuild is compatible with both native activity and custom kotlin activities:
```yaml
android:
  gradle: true
  dependencies:
  - 'androidx.appcompat:appcompat:1.4.1'
  - 'androidx.games:games-activity:1.1.0'
  manifest:
    package: co.realfit.agdkegui
    application:
      has_code: true
      theme: '@style/Theme.AppCompat.Light.NoActionBar'
      activities:
      - name: .MainActivity
```
any code in the `kotlin` directory will be compiled into your app. there is `cargo-mobile` which is a different thing all together relying on vendor tools and ide's and `cargo-ndk`, `cargo-apk`, `xbuild` which are similar in aiming to provide a `cargo build` style simplicity. my hope is that this effort and the xbuild effort can lead to less fragmentation in the long run.

dvc94ch:
One remaining question would be, since the effort to port the native-app-glue to rust has been made, why does the game activity not make use of it?

rib:
> One remaining question would be, since the effort to port the native-app-glue to rust has been made, why does the game activity not make use of it?

The `GameActivity` class is not derived from `NativeActivity`, so the Rust port for `NativeActivity` wouldn't work with `GameActivity` as is.

For example` GameActivity` doesn't use the `AInputQueue` abstraction in the NDK.

There would be code that could be borrowed from the port of the glue code for `NativeActivity` (since there is also some similarity to how they both work) but essentially it's a separate piece of work to do a port for `GameActivity` - and there's notably quite a bit more code to port for `GameActivity` (mainly related to input handling).

`GameActivity` is also a specific, third-party `Activity` that is maintained by Google (with Java and C/C++ code that are designed to work together from an upstream project) and so, even though there are some patches to their C++ code it's possible that we might want to re-sync the with upstream changes and support future versions of `GameActivity`.

For a similar end-result my gut feeling is that we should instead plan to look at vendoring / forking `GameActivity` at some point, to create a `RustActivity` with a design that's more like `GameActivity` than `NativeActivity` and port the C/C++ code to Rust for that. It maybe just comes down to semantics but I think that if we effectively do the port of the C/C++ code for `GameActivity` then it will make sense to call it something different.

rib:
About build system compatibility in general - I'm not quite sure how come that was listed as a con really.

As far as I know this backend + `android-activity` should hopefully support all the same build tools as the previous backend / `ndk-glue`.

There will be additional build system requirements for any native app that doesn't use `NativeActivity` because that's the only one that's distributed as part of Android. (There's no glue layer / backend that could change that OS constraint)

It's a notable consideration when deciding whether to use `NativeActivity` or `GameActivity` but it's meant to be a good thing that developers now get that option / choice (I.e. that's supposed to be one of the main 'pro's here, not a 'con')

README.md
```diff
@@ -99,36 +99,93 @@ book].
...
 ```
 
-And add this to the example file to add the native activity glue:
+All Android applications are based on an `Activity` subclass and the
+`android-activity` crate is designed to support different choices for this base
+class. You application must specify the base class it needs via a feature flag:
+
+| Base Class       | Feature Flag      |  Notes  |
+| :--------------: | :---------------: | :-----: |
+| `NativeActivity` | `android-native-activity` | Built-in to Android - making it possible to build some tests/demos without needing to compile any JVM code. Can give a false sense of convenience because it's often not really possible to avoid needing a build system that can compile some JVM code, to at least subclass `NativeActivity` |
```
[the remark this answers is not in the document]
rib: I've taken a pass at tweaking the wording here.

I've tried to remove the language that implies it's only "tests/demos" that can use NativeActivity or assuming anything about the false sense of convenience it has.

The general intent here was to try and highlight a few of the pros/cons of each option that can help someone make a choice.

I didn't say anything about lack of extensibility, from the pov that developers can subclass `NativeActivity` if they need to (so GameActivity / NativeActivity are both similarly extensible) - but did highlight that it may be necessary to subclass `NativeActivity` to access some platform features.

So although it's notably possible to use `NativeActivity` without compiling any Java/Kotlin code, it's hopefully also clear that it may sometimes be necessary to subclass `NativeActivity`.

I'm not super happy with how these notes are currently packed in this table though, since it squashes the feature name and it's awkward to format the notes. Maybe the notes should be split out into some pros/cons afterwards - not quite sure atm.

msiglreith (approved):

msiglreith:
> About build system compatibility in general - I'm not quite sure how come that was listed as a con really.

Ah, I just meant fragmentation in general not specific to this PR as there is like `cargo-mobile`, `cargo-apk`, `cargo-quad-apk`, `xbuild`, `crossbow` and more with slightly different requirements but still feels like it could be reduced 👀

rib:
Whoop! Thanks for merging @msiglreith! (and also the feedback given)

Although I didn't end up making the stable release for android-activity the other day as I had suggested I'll look at that now and open up a PR to bump the dep to a stable 0.4 version.

Hopefully can also start looking at follow up changes / PRs that I've been a bit reluctant to look at while this PR was pending.

PR #2452, opened 2022-08-31
Initial transition to `objc2`

madsmtm:
Part of https://github.com/rust-windowing/winit/pull/2427.

Concrete improvements in `objc2` which can be seen in this PR:
- `msg_send!` can now return `bool` without special handling.
- Types in `msg_send!` are now required to be `Encode`. This helped catching a lot of undefined behaviour, which are fixed in https://github.com/rust-windowing/winit/pull/2138 and https://github.com/rust-windowing/winit/pull/2428.

Checklist:
  - [x] macOS 10.14
  - [x] iOS simulator (can't get log output here though, so a bit harder)
  - [x] iOS 9.3.6 (iPad mini 1st gen)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- macOS/iOS: Use `objc2` instead of `objc` internally.
 - **Breaking:** Bump MSRV from `1.57` to `1.60`.
 - **Breaking:** Split the `platform::unix` module into `platform::x11` and `platform::wayland`. The extension types are similarly renamed.
 - **Breaking:**: Removed deprecated method `platform::unix::WindowExtUnix::is_ready`.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,12 +62,14 @@ ndk = "0.7.0"
 ndk-glue = "0.7.0"
 
 [target.'cfg(any(target_os = "ios", target_os = "macos"))'.dependencies]
-objc = "0.2.7"
+objc = { version = "=0.3.0-beta.3", package = "objc2" }
 
 [target.'cfg(target_os = "macos")'.dependencies]
-cocoa = "0.24"
-core-foundation = "0.9"
-core-graphics = "0.22"
+# Branch: objc2
+# TODO: Use non-git versions before we release
+cocoa = { git = "https://github.com/madsmtm/core-foundation-rs.git", rev = "c770e620ba0766fc1d2a9f83327b0fee905eb5cb" }
+core-foundation = { git = "https://github.com/madsmtm/core-foundation-rs.git", rev = "c770e620ba0766fc1d2a9f83327b0fee905eb5cb" }
+core-graphics = { git = "https://github.com/madsmtm/core-foundation-rs.git", rev = "c770e620ba0766fc1d2a9f83327b0fee905eb5cb" }
 dispatch = "0.2.0"
 
 [target.'cfg(target_os = "windows")'.dependencies.windows-sys]
--- a/src/platform_impl/ios/monitor.rs
+++ b/src/platform_impl/ios/monitor.rs
@@ -4,12 +4,14 @@ use std::{
     ops::{Deref, DerefMut},
 };
 
+use objc::foundation::{NSInteger, NSUInteger};
+
 use crate::{
     dpi::{PhysicalPosition, PhysicalSize},
     monitor::{MonitorHandle as RootMonitorHandle, VideoMode as RootVideoMode},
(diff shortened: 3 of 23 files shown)


Cargo.toml
```diff
@@ -62,12 +62,13 @@ ndk = "0.7.0"
...
+objc = { version = "=0.3.0-beta.2", package = "objc2" }
 
 [target.'cfg(target_os = "macos")'.dependencies]
-cocoa = "0.24"
-core-foundation = "0.9"
-core-graphics = "0.22"
+# Branch: objc2
+cocoa = { git = "https://github.com/madsmtm/core-foundation-rs.git", rev = "8e6c4854de2408ed1e3e614dbe6f9f2762bde7c3" }
+core-foundation = { git = "https://github.com/madsmtm/core-foundation-rs.git", rev = "8e6c4854de2408ed1e3e614dbe6f9f2762bde7c3" }
+core-graphics = { git = "https://github.com/madsmtm/core-foundation-rs.git", rev = "8e6c4854de2408ed1e3e614dbe6f9f2762bde7c3" }
```
madsmtm, line 71: This is only in the interim; once https://github.com/rust-windowing/winit/pull/2427 is fully merged, this will be fixed

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- **Breaking:** Bump MSRV from `1.57` to `1.60`.
```
madsmtm, line 11: Note that `objc2` doesn't have a MSRV policy yet, see https://github.com/madsmtm/objc2/issues/203.

But since we use a pinned dependency, this shouldn't be a problem (yet).

.github/workflows/ci.yml
```diff
@@ -22,7 +22,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        rust_version: [1.57.0, stable, nightly]
+        rust_version: ['1.60.0', stable, nightly]
```
kchibisov, line 25: Not sure if bumping straight to 1.60 worth it? We might want to keep msrv only for linux/bsd though, since there's no such thing really as a system rust installation. But that should be addressed in a separate PR.

And in general every change to the CI should be done in a separate patch.

Cargo.toml
kchibisov, line 71: Could you add `TODO` like we had for android before, so we won't forget to update to non-git versions.

src/platform_impl/ios/monitor.rs
```diff
@@ -175,8 +167,8 @@ impl fmt::Debug for MonitorHandle {
 
 impl MonitorHandle {
     pub fn retained_new(uiscreen: id) -> MonitorHandle {
+        assert_main_thread!("`MonitorHandle` can only be cloned on the main thread on iOS");
```
kchibisov, line 170: This has nothing to do with that PR, but could `libdispatch` used to `drop` from the main thread? So we won't have runtime panics like that? Though I think the issue that you can't use that from the other thread as well?

It just all looks wrong to me. We say that winit window is `Send` and you can get `MonitorHandle` from the `Window`(owning one iirc), but when you drop it you'll panic?

kchibisov (changes requested):
I don't have that much to say, given that I have no clue about the `objc` internals. The changes look good, some minor notes.

.github/workflows/ci.yml
madsmtm: > Not sure if bumping straight to 1.60 worth it? We might want to keep msrv only for linux/bsd though, since there's no such thing really as a system rust installation. But that should be addressed in a separate PR.

Yeah, unsure here too, will mull this over a bit. `objc2` requires 1.60 because of changes to `Cargo.toml`, so unsure how that affects other platforms that don't use `objc2`.

> And in general every change to the CI should be done in a separate patch.

I think it makes sense to do them together, so that you easily know which PR needs a specific MSRV, but I can see why it makes sense to do them separately given that all maintainers might not review my iOS/macOS PRs.

src/platform_impl/ios/monitor.rs
madsmtm: There is _so_ much thread-unsafety in our macOS (and probably also iOS, but less well-versed in that) implementation, really, you have no _idea_!

Anyway, to answer your question: Yes, we could and should use `libdispatch` to do things on the main thread, both here and in a lot of other places. I'm in the process of exploring solutions, have focused on `objc2` stuff first though.

.github/workflows/ci.yml
madsmtm: I moved the bump to https://github.com/rust-windowing/winit/pull/2453.

PR #1882, opened 2021-03-14
feat: add accept_first_mouse for macOS

z4122:
- [x] Tested on all platforms changed
- [x] Compilation warnings were addressed
- [x] `cargo fmt` has been run on this branch
- [x] `cargo doc` builds successfully
- [x] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users
- [x] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior
- [x] Created or updated an example program if it would help users understand this functionality
- [x] Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented

On macOS, change the default behavior for first click when window was unfocused. Now window will become focused and then respond to the click event when mouse clicks.
To keep consistent with windows.

PR #2457, opened 2022-09-02
Accepts first mouse

mgax:
- ~Created or updated an example program if it would help users understand this functionality~

---

Quoting https://github.com/rust-windowing/winit/pull/1882#pullrequestreview-648621241:
> At some point we should consider making it user-configurable, but let's wait until someone comes with a concrete use-case!

The use case is for Neovide (https://github.com/neovide/neovide/issues/1343), so it behaves more like a native MacOS application.


mgax:
CI is failing with a type mismatch between `bool` and `BOOL`. I've [added](https://github.com/rust-windowing/winit/pull/2457/commits/d2268318182aab06f05623098f97c887a65c26ba) `as BOOL` to the return value; I hope it's the right thing to do, and the `unsafe` block isn't hiding anything unsafe :)

madsmtm (commented):
This looks good, thanks!

If you don't mind, I think I'll wait with this 'till https://github.com/rust-windowing/winit/pull/2427 is done, at which point the `bool` stuff should also be easier.

mgax:
> If you don't mind, I think I'll wait with this 'till #2427 is done, at which point the `bool` stuff should also be easier.

I don't mind. But it's sorted out – at least, the CI is happy :)

mgax:
I've updated the branch. The new `#[sel(acceptsFirstMouse:)]` syntax is pretty neat.

I think the tests failed because of GitHub; one just says `Error: Internal Server Error`. I'd re-run it but I don't have permission. @madsmtm can you trigger a re-run? Or should I just push an empty commit?

madsmtm (changes requested):
Apologies, I should have been more clear, I wasn't actually ready with https://github.com/rust-windowing/winit/pull/2427 yet.

Could I get you to merge the latest `master`, and then move `accepts_first_mouse` from `ViewState` to `WinitView` (would do it myself, but I don't have write access to your branch).

mgax:
> Apologies, I should have been more clear, I wasn't actually ready with #2427 yet.

No worries! I'm not used to working with Rust or Objective-C and it's fun to follow along while you refactor the internals :)

I hope I got it right with storing the flag on `SharedState`, that's what other flags seem to be doing.

src/platform_impl/macos/window.rs
```diff
@@ -149,6 +151,7 @@ pub struct SharedState {
     /// bar in exclusive fullscreen but want to restore the original options when
     /// transitioning back to borderless fullscreen.
     save_presentation_opts: Option<NSApplicationPresentationOptions>,
+    accepts_first_mouse: bool,
```
madsmtm, line 154: Actually, since we're accessing this from inside `WinitView`, I would really prefer if this variable could live in `WinitView`.

I know that may be a bit difficult, you'll have to change the initializer to pass it through that - please do tell if you need help with it!

(I'm trying to move away from having one big `SharedState`/`ViewState` struct, and instead just have normal ivars on the objects).

madsmtm (changes requested):

src/platform_impl/macos/window.rs
mgax: > please do tell if you need help with it!

Yes please :)

First of all, what should the variable definition look like? `accepts_first_mouse: IvarDrop<Box<bool>>` perhaps? Is this the right place? https://github.com/rust-windowing/winit/blob/ba49db2cb9932f2ec90ba9466f980b19bb102081/src/platform_impl/macos/view.rs#L135-L139

Then, this `new` function should take an extra `accepts_first_mouse: bool` parameter?
https://github.com/rust-windowing/winit/blob/ba49db2cb9932f2ec90ba9466f980b19bb102081/src/platform_impl/macos/view.rs#L916-L918

It returns a `Id<Self, Shared>`, and I don't understand what that means, but I guess it's a pointer to the new instance; should I store it in a variable, set `accepts_first_mouse` on it, and return it?

Alternatively, would it make sense to add a `set_accepts_first_mouse` method, to be called after `new`? I can't get this to work though; seems it's not possible to get a mutable reference to the view class:

```
error[E0596]: cannot borrow data in dereference of `Id<WinitView, Shared>` as mutable
   --> src/platform_impl/macos/window.rs:361:9
    |
361 |         view.set_accepts_first_mouse(pl_attrs.accepts_first_mouse);
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot borrow as mutable
    |
    = help: trait `DerefMut` is required to modify through a dereference, but it is not implemented for `Id<WinitView, Shared>`
```

---
BTW, the pull request has _Allow edits and access to secrets by maintainers_ checked, so you should be able to push to it, if you want.

madsmtm: > First of all, what should the variable definition look like? `accepts_first_mouse: IvarDrop<Box<bool>>` perhaps?

`accepts_first_mouse: bool` would suffice (the `IvarDrop` is only used for types like `Box` and `Id`, while `bool` doesn't need to be dropped (and implements `Encode`)).

> Is this the right place?
> 
> https://github.com/rust-windowing/winit/blob/ba49db2cb9932f2ec90ba9466f980b19bb102081/src/platform_impl/macos/view.rs#L135-L139

Yup.

> Then, this `new` function should take an extra `accepts_first_mouse: bool` parameter?
> 
> https://github.com/rust-windowing/winit/blob/ba49db2cb9932f2ec90ba9466f980b19bb102081/src/platform_impl/macos/view.rs#L916-L918

Indeed!

> It returns a `Id<Self, Shared>`, and I don't understand what that means, but I guess it's a pointer to the new instance

It's a pointer to the new instance, that will automatically `release` on `Drop`. Think of `Id<T, Shared>` as the Objective-C equivalent of `Arc`, while `Id<T, Owned>` is `Box`.

> should I store it in a variable, set `accepts_first_mouse` on it, and return it?

You should change the `initWithId:` selector (that is just an arbitary name I picked) such that it can accept another argument. For example `initWithWindow:acceptsFirstMouse:`.

> Alternatively, would it make sense to add a `set_accepts_first_mouse` method, to be called after `new`? I can't get this to work though; seems it's not possible to get a mutable reference to the view class:
> 
> ```
> error[E0596]: cannot borrow data in dereference of `Id<WinitView, Shared>` as mutable
>    --> src/platform_impl/macos/window.rs:361:9
>     |
> 361 |         view.set_accepts_first_mouse(pl_attrs.accepts_first_mouse);
>     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot borrow as mutable
>     |
>     = help: trait `DerefMut` is required to modify through a dereference, but it is not implemented for `Id<WinitView, Shared>`
> ```

It could work if you returned `Id<Self, Owned>` above. I'm in the progress of figuring out what we should actually do here, leaning towards having e.g. `accepts_first_mouse: Cell<bool>` if we'd want it to be mutable. See https://github.com/madsmtm/objc2/issues/265 for details.

> BTW, the pull request has _Allow edits and access to secrets by maintainers_ checked, so you should be able to push to it, if you want.

Huh, I tried it before and it didn't work, but maybe I messed something up. If you want to have a go at doing stuff yourself, feel free to, otherwise I can do it (in a few days probably)

mgax: The `initWithWindow:acceptsFirstMouse:` selector feels a bit janky, and I guess `accepts_first_mouse` _should_ be mutable, but TBH I'm not up to messing with Rust <-> ObjC much more :)

src/platform_impl/macos/view.rs
```diff
@@ -185,6 +190,7 @@ declare_class!(
                 }
 
                 this.state.input_source = this.current_input_source();
+                Ivar::write(&mut this.accepts_first_mouse, accepts_first_mouse);
```
madsmtm, line 193: Nit: I'd like the variable to be initialized just below the other `Ivar::write` (so one can easily know that calling methods on `this` is sound)

src/platform_impl/macos/view.rs
```diff
@@ -911,14 +917,21 @@ declare_class!(
         #[sel(acceptsFirstMouse:)]
         fn accepts_first_mouse(&self, _event: &NSEvent) -> bool {
             trace_scope!("acceptsFirstMouse:");
-            self.window().accepts_first_mouse()
+            println!("acceptsFirstMouse = {}", *self.accepts_first_mouse);
```
madsmtm, line 920: Nit: Remove println

src/platform_impl/macos/window.rs
madsmtm: Yeah, I agree that `accepts_first_mouse` should be mutable, again, working on this in https://github.com/madsmtm/objc2/issues/265 ;)

madsmtm (changes requested):
Wonderful, only small nits

src/platform_impl/macos/view.rs
mgax: 🤦‍♀️ whoopsie :)

madsmtm (approved):
Thanks!

madsmtm:
I had to fix the changelog entry beforehand but still didn't seem to have access to the branch, idk. maybe I'm doing something wrong, so ended up merging this manually in 48b843e42d8b807e8147217b666a732db41db31a instead.

mgax:
@madsmtm thanks for seeing this through!

PR #2459, opened 2022-09-02
Remove `WindowBuilderExtIOS::with_root_view_class`

madsmtm:
This was introduced in https://github.com/rust-windowing/winit/pull/609 (CC @mtak-), and while it makes sense, it feels quite unclean and inhibits other improvements (also, it is unsound, but that could be fixed by marking the function as `unsafe`).

Primary motivation is that it prevents us from doing the iOS equivalent of https://github.com/rust-windowing/winit/pull/2458, since `objc2::declare_class!` (intentionally) does not support dynamically declaring classes. If we want this, it should be done in a way that the desired superclass is known at compile-time.

I did a search on GitHub to get a feel for who was using it, seems like it's really only `glutin`, and that will have to do something else after https://github.com/rust-windowing/glutin/pull/1435 anyhow. Libraries like [`wgpu`](https://github.com/gfx-rs/wgpu/blob/a08ea2a255f4e3965827bc5bc0a7e283855ff890/wgpu-hal/src/metal/surface.rs#L128) and [`vulkano-win`](https://github.com/vulkano-rs/vulkano/blob/2e570f8c3fe5bcf50e7c54895b9caac8158b1226/vulkano-win/src/winit.rs#L222) have already found solutions for this, I strongly suspect that `glutin` could do something similar like (but I don't have a setup to verify that atm.):
```rust
let main_layer: Id<CALayer, Shared> = msg_send_id![view, layer];
let new_layer: Id<CALayer, Shared> = msg_send_id![CAEAGLLayer::class(), new];
let _: () = msg_send![main_layer, addSublayer: &*new_layer];
```

Would like your input on this one @francesca64?


francesca64:
@ArthurKValladares does brainstorm still use this?

ArthurKValladares:
> @ArthurKValladares does brainstorm still use this?

We do still use this internally. I can try to carve out some time early next week to explore the alternative solutions linked there used by wgpu and vulkano. We are also already on our own branch, with changes that don't really makes sense to be merged upstream, so I would not consider our internal usage of this feature to be a blocker for the PR.

madsmtm:
Alternative solution would be to allow overriding the return value from `+[UIView layerClass]`.

I'm going to merge this now, though if you find that none of the other solutions are workable, feel free to post an issue about it, then we can more easily discuss what to do.

PR #2460, opened 2022-09-02
Removed once_cell usage where possible now that Mutex::new and RwLock…

vahanbgs:
…:new are const fns


vahanbgs:
Requires Rust 1.63 to compile, that's why CI checks are failing

madsmtm:
Yeah, see https://github.com/rust-windowing/winit/pull/2453 for a bit of discussion.

In short, I expect that we'll first be able to include in around ~6 months when distros have caught up, so some time around February.

kchibisov:
There's no intention of using 1.63 in a near future.

PR #2483, opened 2022-09-09
Simplify internal type construction

madsmtm:
Change things internally such that platform implementations don't have to import `Fullscreen`, `MonitorHandle` and `VideoMode` from the top-level crate.

  - This is really a "it compiles, it works" change, but have tested on macOS and iOS

### DIFF
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -23,7 +23,6 @@ use crate::{
     error,
     event::{self, VirtualKeyCode},
     event_loop::{self, ControlFlow},
-    monitor,
     window::{self, CursorGrabMode},
 };
 
@@ -636,10 +635,8 @@ pub struct EventLoopWindowTarget<T: 'static> {
 }
 
 impl<T: 'static> EventLoopWindowTarget<T> {
-    pub fn primary_monitor(&self) -> Option<monitor::MonitorHandle> {
-        Some(monitor::MonitorHandle {
-            inner: MonitorHandle,
-        })
+    pub fn primary_monitor(&self) -> Option<MonitorHandle> {
+        Some(MonitorHandle)
     }
 
     pub fn available_monitors(&self) -> VecDeque<MonitorHandle> {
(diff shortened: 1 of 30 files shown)


src/platform_impl/android/mod.rs
```diff
@@ -18,12 +18,12 @@ use raw_window_handle::{
     AndroidDisplayHandle, HasRawWindowHandle, RawDisplayHandle, RawWindowHandle,
 };
 
+use super::Fullscreen;
```
kchibisov, line 21: Could you use full patch instead of related via `super`?

src/platform_impl/ios/monitor.rs
```diff
@@ -220,21 +218,22 @@ impl Inner {
...
                 let mode: id = msg_send![available_modes, objectAtIndex: i];
+                // Use Ord impl of RootVideoMode
                 modes.insert(RootVideoMode {
                     video_mode: VideoMode::retained_new(self.uiscreen, mode),
                 });
             }
         }
 
-        modes.into_iter()
+        modes.into_iter().map(|mode| mode.video_mode)
```
kchibisov, line 236: That looks strange. Why do we even need `BTreeSet` in the first place here? To sort them somehow? I doubt other platforms sort anything here? And if sorted order is required can we do that on a higher level instead to force on all platforms instead?

kchibisov (changes requested):
Converting on a higher level seems fine to me.

src/platform_impl/ios/monitor.rs
madsmtm: Windows does it for a reason:
https://github.com/rust-windowing/winit/blob/155f1f972046d77c9b0900db2d0500d89a7ada45/src/platform_impl/windows/monitor.rs#L228-L230

I suspect that iOS just followed along.

madsmtm: Agree that it would be nicer to do the sorting on a higher level, though the deduplication probably makes sense to only do on Windows.

kchibisov: Hm. Maybe that's fine the way it's then.

msiglreith (approved):

kchibisov (approved):

PR #2488, opened 2022-09-09
Remove automatic publish script

kchibisov:
This script is confusing and provides no value especially with release branches and patch fixes.


MarijnS95:
I don't see anything confusing here when it comes to release branches and patch fixes? As long as a tag gets pushed (with the version bumped), it'll auto-release?

MarijnS95:
@madsmtm Same issue for `android-ndk-rs`, is there any way to "fix" this and make it say "The rust-windowing publishers" or something along those lines?

kchibisov:
>I don't see anything confusing here when it comes to release branches and patch fixes?

It just takes zero effort for me to make a release. I can actually add a tag after I've published, so I can ensure that everything is sane in case CI will fail. I don't like anything automatic when it uses exactly one line to run in the end. Also, given that we have no error propagation from the CI since we allow it to fail, there's no way for you to know that something has failed or not.

It already caused issues a bunch of times in glutin and not really a good thing, given that it'll try to make a release for different tag, etc.

The metadata is tight to the `token` of who pushed the crate. So on CI we use the token generated by @francesca64 , so all releases are made by her.

madsmtm (approved):
I don't think it's possible to fix change the crate metadata (unless we created a new GitHub "user" account) - anyhow, it tends to be @kchibisov that does the release PRs anyhow, so if you want to do it manually, let's just

PR #2489, opened 2022-09-09
Add release process

### LINKED ISSUE #2454: Formalize release process
winit should have a reliable not blocking way to handle releases and patch releases.-

We have something like that formally documented in alacritty and I'd suggest that we can port it with adjustments into winit. See https://github.com/alacritty/alacritty/blob/master/CONTRIBUTING.md#release-process.

kchibisov:
This should help maintainers to handle releases and
establish a non-blocking workflow.

Fixes #2454.

--
cc @chrisduerr since this is ported from alacritty and I'm not sure how we should do that wrt `-dev` versions in libraries.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -48,15 +48,25 @@ The current maintainers are listed in the [CODEOWNERS](.github/CODEOWNERS) file.
 
 If you are interested in being pinged when testing is needed for a certain platform, please add yourself to the [Testers and Contributors](https://github.com/rust-windowing/winit/wiki/Testers-and-Contributors) table!
 
-## Making a new release
+## Release process
 
-If you believe a new release is warranted, you can make a pull-request with:
-- An updated version number (remember to change the version everywhere it is used).
-- A new section in the changelog (below the `# Unreleased` section).
+Given that winit is a widely used library we should be able to make a patch
+releases at any time we want without blocking the development of new features.
 
-This gives contributors an opportunity to squeeze in an extra PR or two that they feel is valuable
-enough to warrant blocking the release a little.
+To achieve these goals, a new branch is created for every new release. Releases
+and later patch releases are committed and tagged in this branch.
 
-Once the PR is merged, a maintainer will create a new tag matching the version name (e.g. `v0.26.1`),
-and a CI job will automatically release the new version. Remember that the release date in the
-changelog must be kept in check with the actual release date.
+The exact steps for an exemplary `0.2.0` release might look like this:
+  1. Initially the version on the latest master is `0.1.0`
+  2. A new `v0.2.x` branch is created for the release
+  3. In the branch, the version is bumped to `v0.2.0`
+  4. The new commit in the branch is tagged `v0.2.0`
+  5. The version is pushed to crates.io
+  6. A GitHub release is created for the `v0.2.0` tag
+  7. On master, the version is bumped to `0.2.0` and the CHANGELOG is updated
+  
+When doing a patch release the process is similar:
+  1. Initially the version of the latest release is `0.2.0`
+  2. Checkout the `v0.2.x` branch
+  3. Cherry-pick the required non-breaking changes into the `v0.2.x`
+  4. Follow steps 3-7 of the regular release example


CONTRIBUTING.md
```diff
@@ -48,15 +48,25 @@ The current maintainers are listed in the [CODEOWNERS](.github/CODEOWNERS) file.
...
+and later patch releases are committed and tagged in this branch.
+
+The exact steps for an exemplary `0.2.0` release might look like this:
+  1. Initially the version on the latest master is `0.1.0`
+  2. A new `v0.2.x` branch is created for the release
+  3. In the branch, the version is bumped to `v0.2.0`
+  4. The new commit in the branch is tagged `v0.2.0`
+  5. The version is pushed to crates.io
+  6. A GitHub release is created for the `v0.2.0` tag
+  7. On master, the version is bumped to `0.2.0` and the CHANGELOG is updated
```
madsmtm, line 66: I'd like the commit from which the release was made (the commit hash is present in crates metadata) to be reachable from `master`

madsmtm (commented):
There should probably be a step for remembering to:
- Check that `docs.rs` works
- The changelog date is correct

CONTRIBUTING.md
kchibisov: Reachable in what way? It'll reside on its own branch created from master branch. You can't have it on master.

kchibisov:
>The changelog date is correct

I don't really like the date field since there's no real value in it. You have it on github release page and on crates.io, why do we need it in a changelog that you can screw is questionable.

>Check that docs.rs works

How do you check that? Isn't having `doc` build on CI passing isn't required for that?

madsmtm:
> > The changelog date is correct
> 
> I don't really like the date field since there's no real value in it. You have it on github release page and on crates.io, why do we need it in a changelog that you can screw is questionable.

I'm not opinionated here, if we want to remove the dates then sure.

> > Check that docs.rs works
> 
> How do you check that? Isn't having `doc` build on CI passing isn't required for that?

Just go to https://docs.rs/crate/winit/latest, and make sure it doesn't say there's an error; maybe it's not that important, but I've seen it break sometimes.

CONTRIBUTING.md
madsmtm: Reachable as-in when I've checked out `master`, I can see the tag in `git log`

chrisduerr: Not having the release on `master` is kinda essential to this entire release process.

The point is not having to go for any roundtrips holding up master while making releases.

CONTRIBUTING.md
```diff
@@ -48,15 +48,25 @@ The current maintainers are listed in the [CODEOWNERS](.github/CODEOWNERS) file.
...
+## Release process
+
+Given that winit is a widely used library we should be able to make a patch
+releases at any time we want without blocking the development of new features.
+
+To achieve these goals, a new branch is created for every new release. Releases
+and later patch releases are committed and tagged in this branch.
+
+The exact steps for an exemplary `0.2.0` release might look like this:
+  1. Initially the version on the latest master is `0.1.0`
```
chrisduerr, line 60: I think the `-dev` suffix is useful to indicate that this is a different version from `0.1.0` and you won't ever find the `0.1.0` version on master.

CONTRIBUTING.md
kchibisov: @madsmtm you can `git checkout tag`. But yeah, the entire point is not having releases on master.

CONTRIBUTING.md
kchibisov: The issue is we want to make it possible to use `patch` section, so if we bump to dev it would mean it's a breaking change and it'll prevent `patch` section from working.

Like for alacritty it makes sense, but we're an application. But for library which you sometimes want to have as breaking that could help?

chrisduerr: Yeah I guess that makes some sense, though ultimately any git change *could* be a breaking change but that's every commit.

CONTRIBUTING.md
madsmtm: What I wanted to say was, it would be nice if we could have the first `0.x.0` release be visible from `master`; of course the rest wouldn't be

chrisduerr: That means you won't be able to make any commits to master from RC start to RC end of that `0.x.0` release though. If you have longer release/testing processes, that can significantly hold back features from getting merged.

Imo no PR should ever be held back because you're in the process of making a release, to me that indicates the release process is broken.

CONTRIBUTING.md
kchibisov: Yeah, but you can still can pull and probably fix them somehow. But if you have `-dev` you can't even pull.

CONTRIBUTING.md
kchibisov: We won't have RCs for example and will have a branch for all ongoing releases. So we could tag it, but it's visible anyway, since you branched out from master.

CONTRIBUTING.md
chrisduerr: I mean the version would be updated after release, right? So a breaking commit would then not work with patch anymore either.

But that point is kinda moot because with `-dev` this won't even work for patch versions and once you select that breaking change by picking either `-dev` or major version, you'll have to lock it to a commit anyway unless you want everything to explode constantly.

kchibisov: >But that point is kinda moot because with -dev this won't even work for patch versions and once you select that breaking change by picking either -dev or major version, you'll have to lock it to a commit anyway unless you want everything to explode constantly.

I think the issue we're talking about that if you have winit as reexport from e.g. glutin if you bump to `dev` you won't be able to do any updates (even if you intended to fix them) because cargo will prevent it. The point is to allow picking master due to semver compat and fixing issues on your own.

kchibisov:
I think I'll go like that for now. If non having `dev` will confuse folks, we'll adjust that.

PR #2490, opened 2022-09-09
Add WindowBuilder::with_centered_position for Windows, macOS and x11.

### LINKED ISSUE #2487: Add option to center the window position on desktop.
Feature request.

ar37-rs:
fixes #2487

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ Please keep one empty line before and after all headers. (This is required for `
 And please only add new entries to the top of this list, right below the `# Unreleased` header.
 
 # Unreleased
-
+- Added `WindowBuilder::with_centered_position` for Windows, macOS and x11.
 - Migrated `WindowBuilderExtUnix::with_resize_increments` to `WindowBuilder`.
 - Added `Window::resize_increments`/`Window::set_resize_increments` to update resize increments at runtime for X11/macOS.
 - macOS/iOS: Use `objc2` instead of `objc` internally.
--- a/examples/cursor.rs
+++ b/examples/cursor.rs
@@ -2,6 +2,7 @@
 
 use simple_logger::SimpleLogger;
 use winit::{
+    dpi::LogicalSize,
     event::{ElementState, Event, KeyboardInput, WindowEvent},
     event_loop::EventLoop,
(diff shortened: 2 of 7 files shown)


kchibisov (changes requested):
What's the point of having this inside winit given that you can implement it outside of winit and it couldn't be supported on all the platforms?

You can set position via `with_outer_position` and get monitor information from `event_loop`.

ar37-rs:
@kchibisov it's actually very easy to do that and actually i've done that before. i meant i want the window being centered at initialization (being centered before it shown), so it doesn't look ugly by showing the window position twice: one on random position and then it being centered after it shown on random position.

ar37-rs:
Implement it outside of winit and getting to the know the monitor screen size it's easy on Windows by just get the value out of `extern "C" fn GetSystemMetrics(_: i32) -> i32`, but on Mac, Linux it's not.

kchibisov:
I mean that you have an Api available on event loop for that?

https://docs.rs/winit/0.27.2/winit/event_loop/struct.EventLoop.html#method.primary_monitor

or https://docs.rs/winit/0.27.2/winit/event_loop/struct.EventLoop.html#method.available_monitors

And then you can use `with_outer_position`?

kchibisov:
Isn't monitor size not enough for you, which you can get out of `MonitorHandle` from event loop prior creating the window? You don't need any extra crate for that...

ar37-rs:
Oh i didn't know that before. that's what i need, thank you. now i'm able to center the window  like so: 
```rust
    let inner_size = PhysicalSize::new(800, 600);
    let size = event_loop.primary_monitor().unwrap().size();
    let x = (size.width - inner_size.width) / 2;
    let y = (size.height - inner_size.height) / 2;
    let centered_position = Position::Logical(LogicalPosition::new(x as _, y as _));
        .with_inner_size(inner_size)
        .with_position(centered_position)
        .with_transparent(true)
        .build(&event_loop)
        .unwrap();
```

kchibisov:
You seem to compute `PhysicalPosition` here, but besides that, that's how I'd think it works.

More reliable way could be to create `hidden` window and then compute position based on `current_monitor` and `set_outer_position` for it. But you've got an idea.

Issue #2495, opened 2022-09-17
Windows: Focus event not emitted after minimize


AustinJ235:
## Info:
**Version:** 0.27.3
**Platform:** Windows 11

## Issue:
Upon maximizing the window from tray. `WindowEvent::Focused(true)` is not emitted. `WindowEvent::Focused(false)` is however emitted when the window is minimized to tray.

## Example:
```rust
use winit::{
    event::{Event, WindowEvent},
    event_loop::{ControlFlow, EventLoop},
    window::WindowBuilder,
};

fn main() {
    let event_loop = EventLoop::new();
    let window = WindowBuilder::new().build(&event_loop).unwrap();

    event_loop.run(move |event, _, control_flow| {
        *control_flow = ControlFlow::Wait;

        match event {
            Event::WindowEvent {
                event: WindowEvent::CloseRequested,
                window_id,
            } if window_id == window.id() => *control_flow = ControlFlow::Exit,

            Event::WindowEvent {
                event: WindowEvent::Focused(focused),
                ..
            } => {
                println!("{}", focused);
            },
            
            _ => (),
        }
    });
}
```

killianc3:
it seems that replacing the WM_NCACTIVATE message in the event_loop with WM_ACTIVATE solves the problem. I make a pull request for that: #2496

msiglreith:
fixed by https://github.com/rust-windowing/winit/pull/2496

Issue #2102, opened 2021-12-18
Incorrect focus events for windows created without focus


swooster:
On Windows 10, I've noticed that if `cargo run` takes a while to build my executable and I switch focus to another application, then the focus events sent to my executable are incorrect, preventing it from determining whether it's currently focused.

Here's a minimal reproducible example:
```rust
use std::{thread, time};

use winit::{
    event::{Event, WindowEvent},
    event_loop::{ControlFlow, EventLoop},
    window::Window,
};

fn main() {
    println!("Select another window within 5 seconds.");
    thread::sleep(time::Duration::from_secs(5));

    let event_loop = EventLoop::new();
    let window = Window::new(&event_loop).unwrap();

    event_loop.run(move |event, _event_loop_target, control_flow| match event {
        Event::WindowEvent {
            event: WindowEvent::CloseRequested,
            window_id,
        } if window_id == window.id() => {
            *control_flow = ControlFlow::Exit;
        }
        Event::WindowEvent {
            event: WindowEvent::Focused(focused),
            window_id,
        } if window_id == window.id() => {
            println!("Focused: {}", focused);
            // window.set_cursor_grab(focused).unwrap();
            // window.set_cursor_visible(!focused);
        }
        _ => (),
    });
}
```

The behavior I'd expect is at least one (ideally both) of the following:
* The `winit` window receives a `Focused(true)` event if it's created in a focused state.
* The `winit` window receives a `Focused(false)` event if it's created in an unfocused state.

With `winit` 0.26.0, if you run this and focus another application during the 5 second wait, then the `winit` window will be created in the background, yet the application will receive a `Focused(true)` event. If the application tries to act on this by grabbing the cursor (uncomment the lines near the end), then you end up with a strange situation where the cursor is constrained to the bounds of the `winit` window, despite another application having focus and being in front.

With `winit` 0.20.0 through 0.25.0, if you follow the above steps, then the `winit` window receives no event, even when focused for the first time.

I haven't tested with versions of `winit` older than 0.20.0.

maroider:
So I put a `panic!()` inside our `WM_SETFOCUS` handler, and it looks like having it be invoked while inside the `init` function is causing some trouble. #1933 strikes again.

<details>
<summary>
the backtrace
</summary>


```
thread 'main' panicked at 'explicit panic', C:\Users\Markus Røyset\source-2\github.com\maroider\winit\src\platform_impl\windows\event_loop.rs:1678:13
stack backtrace:
[...]

PR #2496, opened 2022-09-18
fix issue #2495

killianc3:
Simple fix for the window focus problem (#2495) under Windows. 
(no additional test, except the test provided in the problem description)


msiglreith (changes requested):
Thanks but this seems to regress https://github.com/rust-windowing/winit/issues/2102
Adding an explicit path for  `WM_ACTIVATE && WA_CLICKACTIVE` which triggers a focus gain should work.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- Fixed focus event not being triggered on Windows (#2495).
```
msiglreith, line 11: small stylistic change (:
```suggestion
- On Windows, fixed focus event emission on minimize.
```

src/platform_impl/windows/event_loop.rs
```diff
@@ -1901,7 +1901,7 @@ unsafe fn public_window_callback_inner<T: 'static>(
         }
 
         WM_NCACTIVATE => {
-            let is_active = wparam == 1;
+            let is_active = wparam != 0;
```
msiglreith, line 1904: Interesting find, seems to emit `0x20_0000` for some reason (undocumented) - might be some combination of active & minimized. Overall the message behavior seems a bit unintuitive of Windows in this case ..

```suggestion
            let is_active = wparam != false.into();
```

msiglreith (commented):
thanks, yes seems to work! small changes left beside that lgtm

msiglreith (approved):

PR #2501, opened 2022-09-22
Add touchpad magnify gesture support for Windows.

killianc3:
After seeing touchpad gestures implemented on macOS ( #2157 ), I looked for an implementation for Windows.

According to the microsfot documentation, there is a WM_GESTURE message that can be processed to support touchpad gestures. But at first I couldn't get one of these messages, then after searching, it seems that no one has implemented this message yet.

Then, in the microsoft doc, I saw this table https://bit.ly/3LDVdRd which describes how the gestures correspond to the existing messages, so based on that and on #2157, I implemented the zoom gesture.

What do you think about it ?
Thanks in advance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- **Breaking:** On Windows, add support for two-finger touchpad magnification gesture with new event `WindowEvent::TouchpadMagnify`.
 - On Windows, fixed focus event emission on minimize.
 - On MacOS, made `accepts_first_mouse` configurable.
 - Migrated `WindowBuilderExtUnix::with_resize_increments` to `WindowBuilder`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,8 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
-- **Breaking:** On Windows, add support for two-finger touchpad magnification gesture with new event `WindowEvent::TouchpadMagnify`.
+- On Windows, added support for touchpad magnification gesture with `WindowEvent::TouchpadMagnify` event.
+- **Breaking:** On Windows the `WindowEvent::MouseWheel` event is no longer emitted during magnification gesture.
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- **Breaking:** On Windows, add support for two-finger touchpad magnification gesture with new event `WindowEvent::TouchpadMagnify`.
```
madsmtm, line 11: Can you explain here why it's a breaking change (maybe separate the changelog entry in two?)

killianc3: I was thinking that maybe some people have already implemented the magnification gesture in their code by checking if the control key is activated when they receive a MouseWheel event. But with these changes, their code might not work on Windows anymore because MouseWheel will no longer be emitted. 

And I also follow this #2157 which is also a note as a break change.

killianc3:
I was thinking that maybe some people have already implemented the magnification gesture in their code by checking if the control key is activated when they receive a MouseWheel event. But with these changes, their code might not work on Windows anymore because MouseWheel will no longer be emitted.

And I also follow this #2157  wich is also a note as a break change.

madsmtm:
Yeah, I could guess - I was more thinking explaining it _in_ the changelog, so that the user knows "oh, I gotta check for my usage of `MouseWheel`, and see if I do that"

killianc3:
Oh yes, I see, so something like that should be more understandable ?

madsmtm (approved):
Yup, that looks wonderful.

Can't say anything about the implementation, `@msiglreith` will likely do that at some point

msiglreith:
The implementation looks fine but I'm a bit unsure about doing this inside `winit`:
- There seems no clear mapping from mouse wheel values to magnification factor. Mapping from `-1/1` to zoom factors seems more application specific. Support `Direct Manipulation` or other Windows API in the future might lead to inconsistent behavior here.
- Not totally sure about other platforms but such a keybind should be platform-agnostic.

killianc3:
Yes, you're right, after thinking about it a bit, I also think these changes are not very much in line with Winit's goals, my bad.

 Thanks for your time !

Issue #2502, opened 2022-09-23
Transparent example not working


mgalos999:
When I run `cargo run --example transparent` instead of a transparent window I just get a white window with no decorations.

It seems the problem was introduced in this PR:
https://github.com/rust-windowing/winit/pull/2419
If I revert to a commit before that PR was merged, then I see no GUI window showing up at all which I believe is the intended result (I see the icon in the taskbar and it appears in the alt+tab window list).

When I run the transparent example on the latest commit I just get a white window (with the expected icon in the taskbar)

I am running Windows 10.

mgalos999:
Ok I think I have a fix, but I'm not sure if it will really do the job.

I think you just have to insert this in `src/platform_impl/windows.rs`
```
if self.contains(WindowFlags::TRANSPARENT) {
    style_ex |= WS_EX_LAYERED;
}
```

Note you won't want to use the `WS_EX_TRANSPARENT` style since that doesn't actually make the window transparent, it just makes mouse clicks passthrough

msiglreith:
thanks for the report. It should work fine when actually drawing something onto the window surface, which the examples don't do.
With the changes from the PR the screens windows will be completely invisible for me, no matter what I draw on it.

amrbashir:
> It should work fine when actually drawing something onto the window surface

However this would be an overhead for apps that don't need to draw the window surface, for example when using Webview2.

Transparency used to work out of the box for undecorated windows before https://github.com/rust-windowing/winit/pull/1933 . I really don't know what caused the regression or if it can be fixed but it is the reason why I kept [TAO](https://github.com/tauri-apps/tao) (the winit fork for WRY) using subclasses.

This could also be a blocker for us to migrate back to using winit directly.

kchibisov:
I might be not understand how transparency works on windows, but aren't you required to provide a buffer with alpha value in the end?

Is it just blending with black background now if you draw into subview(not sure how that called on windows, but that's my understanding of what happens). You might just clear it with fully transparent color it or do something like.

amrbashir:
Webview2 works a bit differently. I don't remember the correct term for it but it is a sort of a child window which has its own window surface.
So in order to have transparency, we would need to draw the window surface which is just a bit annoying since we don't need to draw the window surface for any other purpose. 

> This could also be a blocker for us to migrate back to using winit directly.

Also this was an overstatement from me, I think I wanted to say it will be a minor setback.

kchibisov:
[...]

PR #2503, opened 2022-09-23
fixed window transparency

mgalos999:
fixes https://github.com/rust-windowing/winit/issues/2502


stefnotch:
This seems to make borderless, transparent windows invisible for me.
(Using winit v0.28.6 as a base)

TeamDman:
> This seems to make borderless, transparent windows invisible for me.
> (Using winit v0.28.6 as a base)

Is that not the expected behaviour for a transparent window?

TeamDman:
can confirm it makes everything transparent, not just the parts that should be transparent lol

(used cargo patch to try this PR in my project, the main window should be where the outline is but it is entirely invisible)

image

danielnehrig:
is there any issues with this pr ? looking forward to that working.

TheAlan404:
bump

Issue #2504, opened 2022-09-28
`ReceivedCharacter` no longer emitted on Windows for Alt+key since PR#2445


ilya-bobyr:
In https://github.com/rust-windowing/winit/pull/2445 a change was made, affecting only Windows, to spot sending `ReceivedCharacter` in response to `WM_SYSCHAR` messages, or `Alt+key` combination.

This creates a discrepancy between Windows and other platforms.  As on other platforms, `ReceivedCharacter` is still sent when `Alt+b` is pressed, for example.

Alacritty is one of the applications that use `winit` and is negatively affected by this change.
See this issue for Alacritty specific discussion: https://github.com/alacritty/alacritty/issues/6356

Was this intentional that `Alt+key` is no longer sending `ReceivedCharacter` and other platforms will be updated similarly, or was it an accident?

msiglreith:
Yes this was kind of intentional based on the Windows documentation and how these `Alt + <Key>` combinations are processed on Windows. In particular, the following quote would speak against emitting a `ReceivedChar` in this scenario:

>The [WM_SYSCHAR](https://learn.microsoft.com/en-us/windows/desktop/menurc/wm-syschar) message indicates a system character. As with [WM_SYSKEYDOWN](https://learn.microsoft.com/en-us/windows/desktop/inputdev/wm-syskeydown), you should generally pass this message directly to [DefWindowProc](https://learn.microsoft.com/en-us/windows/desktop/api/winuser/nf-winuser-defwindowproca). Otherwise, you may interfere with standard system commands. In particular, do not treat WM_SYSCHAR as text that the user has typed.

kchibisov:
I think the right way would be to revert that for now? Given that it's causing portability issues and causing issues downstream? The right fix for that would be a keyboard v2 API?

What do you think @msiglreith ?

msiglreith:
I mean we can forward the `WM_SYSCHAR` stuff but there seems no good solution in the long run even to expose a `ReceivedChar` for these. The option I currently see are:

1. Don't support system keybinds (basically the previous behavior)
2. Support system keybinds and don't emit `ReceivedChar` (e.g `Alt + Space` will open the menu and not print ` `)
3. System keybinds and emit `ReceivedChar` - emitting two different things for one keybind
4. Expose some flag for the user to decide..

kchibisov:
The issue is that we need `ReceivedChar` downstream. It'll all change with keyboard v2, but I'm not sure it's close.

[...]

PR #2505, opened 2022-09-28
Windows, emit `ReceivedCharacter` on system keybinds

msiglreith:
Workaround for https://github.com/rust-windowing/winit/issues/2504 to support keybinding based on `ReceivedCharacter`. I currently don't see a possibility for a clean solution in the short run without API changes (extra fields or keybinding api)

### DIFF
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -65,7 +65,7 @@ use windows_sys::Win32::{
             WM_MOUSEWHEEL, WM_NCACTIVATE, WM_NCCALCSIZE, WM_NCCREATE, WM_NCDESTROY,
             WM_NCLBUTTONDOWN, WM_PAINT, WM_POINTERDOWN, WM_POINTERUP, WM_POINTERUPDATE,
             WM_RBUTTONDOWN, WM_RBUTTONUP, WM_SETCURSOR, WM_SETFOCUS, WM_SETTINGCHANGE, WM_SIZE,
-            WM_SYSCOMMAND, WM_SYSKEYDOWN, WM_SYSKEYUP, WM_TOUCH, WM_WINDOWPOSCHANGED,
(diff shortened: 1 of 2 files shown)


ajtribick:
Depending on whether or not it makes sense to generate `ReceivedCharacter` for things like ALT+Space (which on Windows should open the window menu), it may make sense to generate a `ReceivedCharacter` event in the handler for `WM_MENUCHAR` instead of handling it from `WM_SYSCHAR`.

src/platform_impl/windows/event_loop.rs
```diff
@@ -1217,7 +1217,18 @@ unsafe fn public_window_callback_inner<T: 'static>(
...
             }
 
-            0
+            // todo(msiglreith):
+            //   Ideally, `WM_SYSCHAR` shouldn't emit a `ReceivedChar` event
+            //   indicating user text input. As we lack dedicated support
+            //   accelerators/keybindings these events will be additionally
+            //   emitted for downstream users.
+            //   This means certain key combinations (ie Alt + Space) will
+            //   trigger the default system behavior **and** emit a char event.
```
ilya-bobyr, line 1226: Would it make sense to mention that this is done in order to have consistent behavior across different platforms?

Viewed from the Windows only perspective, it is indeed strange.
But if a cross-platform application wants to react to `Alt+A`, unless `winit` emits the same sequence of events on all platforms for `Alt+A`, the application logic would need to become platform specific.
For applications that want to process `Alt+key` in a uniform manner, it becomes even more work, as they would need to repeat the decoding written above in this branch.

kchibisov (approved):
I've tested it and it seems to provide the behavior that was before. At least I don't see that many difference between them.

kchibisov:
@msiglreith also, it's not about implementing keybindings for us in alacritty, but more about forwarding keys to terminal applications, since those use alt bindings. With keyboard v2 I think we would get rid of that patch anyway.

PR #2507, opened 2022-10-02
Make raw-window-handle 0.4 optional

djkoloski:
The motivation for this PR is to remove multiple crate versions from the dependency graph for the [`clippy::multiple_crate_versions`](https://rust-lang.github.io/rust-clippy/master/#multiple_crate_versions) lint. I saw that raw-window-handle 0.4 support is slated for removal in 0.28, so it's fine if you'd rather wait until then.


kchibisov:
I don't think that it makes sense given that it'll be removed anyway and we can't add patch release with a new feature?

djkoloski:
That's fine, thanks for the consideration!

PR #2509, opened 2022-10-03
Update cargo-run-wasm

rukai:

This PR updates cargo-run-wasm to the latest release.

A common request I have received is to remove the `<h1></h1>` from the page and give easy access to custom css.
To accommodate that I have reworked the API a little and changed the format of the page.
I have used the `body { margin: 0px; }` css which is the recommended default css to use as with `cargo run-wasm` as it gives the wasm app more space to work with by removing the pointless margin.


madsmtm (approved):
Seems reasonable, thanks!

PR #2519, opened 2022-10-18
macos: Send CapsLock keyboard events

rofferom:
On macOS, no `WindowEvent` is sent when CapsLock is pressed.

Actually there is a trick: only the CapsLock state change is sent by macOS, not the state of the keyboard key. The press/release cycle is simulated to fake the presses.

Checklist

### DIFF
--- a/src/platform_impl/macos/event.rs
+++ b/src/platform_impl/macos/event.rs
@@ -222,6 +222,7 @@ pub fn code_to_key(code: KeyCode, scancode: u16) -> Key {
         KeyCode::ShiftRight => Key::Shift,
         KeyCode::AltRight => Key::Alt,
         KeyCode::ControlRight => Key::Control,
+        KeyCode::CapsLock => Key::CapsLock,
 
         KeyCode::NumLock => Key::NumLock,
         KeyCode::AudioVolumeUp => Key::AudioVolumeUp,
@@ -417,6 +418,7 @@ impl KeyCodeExtScancode for KeyCode {
             KeyCode::SuperRight => Some(0x36),
             KeyCode::SuperLeft => Some(0x37),
             KeyCode::ShiftLeft => Some(0x38),
+            KeyCode::CapsLock => Some(0x39),
             KeyCode::AltLeft => Some(0x3a),
             KeyCode::ControlLeft => Some(0x3b),
             KeyCode::ShiftRight => Some(0x3c),
--- a/src/platform_impl/macos/view.rs
+++ b/src/platform_impl/macos/view.rs
@@ -15,8 +15,8 @@ use objc2::{class, declare_class, msg_send, msg_send_id, mutability, sel, ClassT
 
 use super::{
     appkit::{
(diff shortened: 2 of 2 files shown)


madsmtm (commented):
If we were to move forwards with this, I'd probably advocate for caps lock being added to `VirtualKeyCode`.

But, while I see what you're trying to do, and the implementation looks sound from that perspective, I don't think this is the right way to solve it (since it's impossible for the user to detect how long caps lock is being held, making it unusable for many use-cases); instead, we should be hooking into the keypress at another lower level (which I know is possible, though I don't know how...), such that we can register both the press and the release.

rofferom:
About how CapsLock should be exposed, I had a look at [SDL implementation](https://github.com/libsdl-org/SDL/blob/34231f5ba0f47050fe2c6d50d0810903c262c7f1/src/video/cocoa/SDL_cocoawindow.m#L1123): it seems that it is a modifier **and** an event. With the fake Press/Release trick too.

I made some tests with IOKit. With this API it is possible to catch the real key state, but it requires to grants extra permissions to the application in the Preference Pane, which isn't a correct solution.

It seems that it is possible to get events from Quartz, but I don't know the limitations and if it could solve this issue.

kchibisov:
If you're still interested in this, could you rebase or test on the latest master? We were changing the keyboard API, so that's why it was left unnoticed. Now we have a dedicated key for CapsLock and such and generally report the key downstream.

rofferom:
I tried to integrate my commit in the new macOS codebase, but currently it looks like a big hack even if it works. I'm still struggling to understand the `update_modifiers()` logic.

And github closed this PR, I don't know why.... The branch [still exist on my fork](https://github.com/rofferom/winit/commit/55769261d3b9b3cc3d6da9729463786d33d92f73)

eugenesvk:
> instead, we should be hooking into the keypress at another lower level (which I know is possible, though I don't know how...), such that we can register both the press and the release.

as far as I understood from the doc of the dev who made Karabiner Elements https://docs.google.com/presentation/d/1nEaiPUduh1vjks0rDVRTcJaEULbSWWh1tVdG2HF_XSU/edit#slide=id.p you unfortunatley can't hook into those events without accessibility/other requirements, both of such "raw" IOHID methods require them (though would still be nice if Winit could optionally have such methods)

FYI he also has simple code examples on how to use them https://github.com/pqrs-org/osx-event-observer-examples

madsmtm:
> as far as I understood from the doc of the dev who made Karabiner Elements https://docs.google.com/presentation/d/1nEaiPUduh1vjks0rDVRTcJaEULbSWWh1tVdG2HF_XSU/edit#slide=id.p you unfortunatley can't hook into those events without accessibility/other requirements, both of such "raw" IOHID methods require them (though would still be nice if Winit could optionally have such methods)
> 
> FYI he also has simple code examples on how to use them https://github.com/pqrs-org/osx-event-observer-examples

Yeah, I think that matches my understanding these days as well. But thanks for the links!

If someone was willing to file a PR implementing this behind a feature flag or smth, I'd be glad to review it.

PR #2525, opened 2022-10-24
Add `Window::set_content_protected` on macOS and Windows

amrbashir:
	- [x] macOS
	- [X] Windows

Porting https://github.com/tauri-apps/tao/commit/802146fb8692a46185846a64163c174520450c43


src/window.rs
```diff
@@ -952,6 +952,18 @@ impl Window {
     pub fn theme(&self) -> Option<Theme> {
         self.window.theme()
     }
+
+    /// Prevents the window contents from being captured by other apps.
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **iOS / Android / x11 / Wayland / Web:** Unsupported.
+    pub fn set_content_protected(&self, enabled: bool) {
```
kchibisov, line 961: Could name parameter with `_enabled`?

kchibisov (changes requested):
You might want to have builder for that as well?

src/window.rs
amrbashir: right

msiglreith (approved):

madsmtm (commented):
macOS impls looks good, unsure of how to test it though?

I tried recording the window with QuickTime, but that didn't seem to make a difference?

amrbashir:
Looks like `NSWindowSharingNone` doesn't completely prevent all apps from reading the window content https://github.com/electron/electron/issues/14415#issuecomment-419736562
> This is OS API limitation. We use `[NSWindow setSharingType:NSWindowSharingNone]`, which does not prevent QuickTime from capturing the window unfortunately.

src/window.rs
```diff
@@ -953,6 +968,16 @@ impl Window {
         self.window.theme()
     }
 
+    /// Prevents the window contents from being captured by other apps.
+    ///
```
madsmtm, line 972: > > This is OS API limitation. We use `[NSWindow setSharingType:NSWindowSharingNone]`, which does not prevent QuickTime from capturing the window unfortunately.

Fair enough - would you mind documenting that here as well?

madsmtm (approved):
Thanks, macOS impl looks good

src/window.rs
amrbashir: sure

PR #2526, opened 2022-10-24
Add `Window::title` getter on Windows and macOS

amrbashir:
	- [x] macOS 
	- [X] Windows

Porting https://github.com/tauri-apps/tao/commit/c50529b3ee453c73acf36bd7e1cd7c14669951f3

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- On Windows and macOS, Add `Window::title` getter.
 - On Windows, fix focusing menubar when pressing `Alt`.
 - On MacOS, made `accepts_first_mouse` configurable.
 - Migrated `WindowBuilderExtUnix::with_resize_increments` to `WindowBuilder`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
-- On Windows and macOS, Add `Window::title` getter.
+- On Windows and macOS, add `Window::title` to query the current window title.
 - On Windows, fix focusing menubar when pressing `Alt`.
 - On MacOS, made `accepts_first_mouse` configurable.
 - Migrated `WindowBuilderExtUnix::with_resize_increments` to `WindowBuilder`.
--- a/src/platform_impl/windows/window.rs
+++ b/src/platform_impl/windows/window.rs
@@ -701,7 +701,7 @@ impl Window {
     #[inline]
     pub fn title(&self) -> String {
         let len = unsafe { GetWindowTextLengthW(self.window.0) };
-        let mut buf = vec![0; len as usize];
+        let mut buf = vec![0; (len + 1) as usize];
         unsafe { GetWindowTextW(self.window.0, buf.as_mut_ptr(), len) };
         String::from_utf16_lossy(&buf[..(len - 1) as _])
     }
--- a/src/platform_impl/windows/window.rs
+++ b/src/platform_impl/windows/window.rs
@@ -703,7 +703,7 @@ impl Window {
         let len = unsafe { GetWindowTextLengthW(self.window.0) };
         let mut buf = vec![0; (len + 1) as usize];
         unsafe { GetWindowTextW(self.window.0, buf.as_mut_ptr(), len) };
-        String::from_utf16_lossy(&buf[..(len - 1) as _])
+        String::from_utf16_lossy(&buf[..len as _])
     }
 
     #[inline]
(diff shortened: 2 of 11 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- On Windows and macOS, Add `Window::title` getter.
```
```suggestion
- On Windows and macOS, add `Window::title` to query the current window title.
```

src/window.rs
```diff
@@ -952,6 +952,16 @@ impl Window {
     pub fn theme(&self) -> Option<Theme> {
         self.window.theme()
     }
+
+    /// Gets the current title of the window.
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **iOS / Android / x11 / Wayland / Web:** Unsupported.
```
msiglreith, line 960: should add that we return an empty string in this case.

src/platform_impl/windows/window.rs
```diff
@@ -698,6 +698,14 @@ impl Window {
         Some(self.window_state_lock().current_theme)
     }
 
+    #[inline]
+    pub fn title(&self) -> String {
+        let len = unsafe { GetWindowTextLengthW(self.window.0) };
+        let mut buf = vec![0; len as usize];
```
msiglreith, line 704: don't we need to account for the `\0` terminator here? (haven't tested it so far)

amrbashir: yeah you're correct

src/window.rs
amrbashir: Sure

src/platform_impl/windows/window.rs
```diff
@@ -698,6 +698,14 @@ impl Window {
         Some(self.window_state_lock().current_theme)
     }
 
+    #[inline]
+    pub fn title(&self) -> String {
+        let len = unsafe { GetWindowTextLengthW(self.window.0) };
+        let mut buf = vec![0; (len + 1) as usize];
+        unsafe { GetWindowTextW(self.window.0, buf.as_mut_ptr(), len) };
+        String::from_utf16_lossy(&buf[..(len - 1) as _])
```
kchibisov, line 706: Isn't what you're doing is basically `OsString` here? Though, you might be fine with raw buffers.

amrbashir: do you want to the API to return `OsString` instead or you want me to go through `OsString` to `String`

kchibisov: OsString to String.

amrbashir: Wouldn't that result in extra allocation?

kchibisov: I guess it's fine then.

src/platform_impl/windows/window.rs
```diff
@@ -698,6 +698,14 @@ impl Window {
         Some(self.window_state_lock().current_theme)
     }
 
+    #[inline]
+    pub fn title(&self) -> String {
+        let len = unsafe { GetWindowTextLengthW(self.window.0) };
+        let mut buf = vec![0; (len + 1) as usize];
+        unsafe { GetWindowTextW(self.window.0, buf.as_mut_ptr(), len) };
+        String::from_utf16_lossy(&buf[..len as _])
```
msiglreith, line 706: Let's use the `util::decode_wide` function.
I tested it on the `window` example and it dropped the `!` at the end and includes a `\0`.
Below an implementation suggestion which returns the expected string.
```rust
let len = unsafe { GetWindowTextLengthW(self.window.0) } + 1; // incl. NULL terminator
let mut buf = vec![0; len as usize];
unsafe { GetWindowTextW(self.window.0, buf.as_mut_ptr(), len) };
util::decode_wide(&buf).to_string_lossy().to_string()
```

src/window.rs
```diff
@@ -952,6 +952,16 @@ impl Window {
     pub fn theme(&self) -> Option<Theme> {
         self.window.theme()
     }
+
+    /// Gets the current title of the window.
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **iOS / Android / x11 / Wayland / Web:** Unsupported. Always return an empty string.
```
msiglreith, line 960:
```suggestion
    /// - **iOS / Android / x11 / Wayland / Web:** Unsupported. Always returns an empty string.
```

src/platform_impl/windows/window.rs
amrbashir: Doesn't that result in an extra allocation?

msiglreith: fine for me - it's used in other places already and this function shouldn't be called a lot imo

amrbashir: done

madsmtm (approved):
macOS impl looks good, and I tested it and it works fine, thanks!

msiglreith (approved):

kchibisov (approved):

PR #2530, opened 2022-10-26
On Windows, fix icons specified on `WindowBuilder` not taking effect for windows created after the first one

amrbashir:

### DIFF
--- a/src/platform_impl/windows/window.rs
+++ b/src/platform_impl/windows/window.rs
@@ -786,7 +786,8 @@ impl<'a, T: 'static> InitData<'a, T> {
         let window_state = {
             let window_state = WindowState::new(
                 &self.attributes,
-                self.pl_attribs.taskbar_icon.clone(),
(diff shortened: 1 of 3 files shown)


src/platform_impl/windows/window.rs
```diff
@@ -794,7 +794,8 @@ impl<'a, T: 'static> InitData<'a, T> {
         let window_state = {
             let window_state = WindowState::new(
                 &self.attributes,
-                self.pl_attribs.taskbar_icon.clone(),
+                // Will be set later in on_create
+                None,
```
msiglreith, line 798: let's just remove it from the constructor then as this is the only place we call `WindowState::new`

msiglreith (approved):

src/platform_impl/windows/window.rs
amrbashir: sure

PR #2531, opened 2022-10-26
Add `Window::has_focus`

amrbashir:
	- [X] Windows
	- [X] X11
	- [ ] Wayland
	- [x] Web 
	- [ ] Android
	- [ ] iOS
	- [x] macOS

Porting https://github.com/tauri-apps/tao/commit/7d2eeeebb4da15e9aeda9bf17e80ebdf23c95cee

### DIFF
--- a/src/platform_impl/ios/window.rs
+++ b/src/platform_impl/ios/window.rs
@@ -348,6 +348,10 @@ impl Inner {
         None
     }
 
+    pub fn has_focus(&self) -> bool {
+        self.window.isKeyWindow()
+    }
+
     #[inline]
     pub fn set_theme(&self, _theme: Option<Theme>) {
         warn!("`Window::set_theme` is ignored on iOS");
--- a/src/platform_impl/linux/wayland/window/shim.rs
+++ b/src/platform_impl/linux/wayland/window/shim.rs
@@ -1,5 +1,6 @@
 use std::cell::Cell;
 use std::mem::ManuallyDrop;
+use std::sync::atomic::AtomicBool;
 use std::sync::{Arc, Mutex};
 
 use sctk::reexports::client::protocol::wl_compositor::WlCompositor;
@@ -154,6 +155,9 @@ pub struct WindowHandle {
     /// Whether the window is resizable.
     pub is_resizable: Cell<bool>,
 
+    /// Whether the window has keyboard focus.
+    pub has_focus: Arc<AtomicBool>,
+
     /// Allow IME events for that window.
     pub ime_allowed: Cell<bool>,
 
@@ -187,6 +191,7 @@ impl WindowHandle {
         env: &Environment<WinitEnv>,
         window: Window<WinitFrame>,
         size: Arc<Mutex<LogicalSize<u32>>>,
+        has_focus: Arc<AtomicBool>,
         pending_window_requests: Arc<Mutex<Vec<WindowRequest>>>,
     ) -> Self {
         let xdg_activation = env.get_global::<XdgActivationV1>();
@@ -209,6 +214,7 @@ impl WindowHandle {
             attention_requested: Cell::new(false),
             compositor,
             ime_allowed: Cell::new(false),
+            has_focus,
         }
     }
 
--- a/src/platform_impl/linux/x11/window.rs
+++ b/src/platform_impl/linux/x11/window.rs
@@ -55,6 +55,7 @@ pub struct SharedState {
     pub resize_increments: Option<Size>,
     pub base_size: Option<Size>,
     pub visibility: Visibility,
+    pub has_focus: bool,
 }
 
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
@@ -94,6 +95,7 @@ impl SharedState {
             max_inner_size: None,
             resize_increments: None,
             base_size: None,
+            has_focus: true,
         })
     }
 }
@@ -1602,6 +1604,10 @@ impl UnownedWindow {
     }
 
     #[inline]
+    pub fn has_focus(&self) -> bool {
+        self.shared_state_lock().has_focus
+    }
+
     pub fn title(&self) -> String {
         String::new()
     }
--- a/src/window.rs
+++ b/src/window.rs
@@ -1022,6 +1022,16 @@ impl Window {
         self.window.focus_window()
     }
 
+    /// Gets whether the window has keyboard focus.
+    ///
+    /// This queries the same state information as [`WindowEvent::Focused`].
+    ///
+    /// [`WindowEvent::Focused`]: crate::event::WindowEvent::Focused
+    #[inline]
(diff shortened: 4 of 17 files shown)


src/window.rs
```diff
@@ -925,6 +925,17 @@ impl Window {
         self.window.focus_window()
     }
 
+    /// Gets whether the widnow has keyboard focus.
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **X11 / Wayland:** Not implemented
+    /// - **iOS / Android / Web:** Unsupported, always returns `true`.
```
madsmtm, line 933: Nit: Web should probably be "not implemented" instead of "unsupported"

src/window.rs
```diff
@@ -925,6 +925,17 @@ impl Window {
         self.window.focus_window()
     }
 
+    /// Gets whether the widnow has keyboard focus.
```
madsmtm, line 928:
```suggestion
    /// Gets whether the window has keyboard focus.
```

src/window.rs
```diff
@@ -925,6 +925,17 @@ impl Window {
...
+    ///
```
madsmtm, line 929:
```suggestion
    ///
    /// This queries the same state information as [`WindowEvent::Focused`].
    ///
```

madsmtm (approved):
Thanks, looks good, only a few minor nits.

kchibisov:
I think it's dangerous to add this with blanket `true` impl to be honest, since users my start relying on it. Given that you simply want to get the last value of `Focused(changed)` event we should have the Api on all the platform implemented from the start.

amrbashir:
Unfortunately, I don't have much experience with other platforms so I can't implement them, however I may be able to implement x11 if you steer me into the right direction.

kchibisov:
@amrbashir I mean it's not about knowing though. You can see where we sent `Fosused` event and stash its result on a window, it's just rust and doesn't require extra knowledge at all. For example on Wayland you'd need to pass some `Arc<AtomicBool>` around, on X11 can likely something similar or maybe you can query things directly.

There're no platform specifics at all in the end.

amrbashir:
cool, thought there might be a way to query the OS like macOS impl for example.

kchibisov:
Maybe on X11, but be aware that active window isn't keyboard focused window sometimes, since you can hover the window with mouse and it'll be active, but won't have focus.

amrbashir:
alright, x11, wayland and web, are implemented now but I was only able to test x11.

examples/window.rs
```diff
@@ -19,7 +19,7 @@ fn main() {
 
     event_loop.run(move |event, _, control_flow| {
         control_flow.set_wait();
-        println!("{:?}", event);
+        dbg!(window.has_focus());
```
kchibisov, line 22: This should be reverted.

src/platform_impl/linux/x11/window.rs
```diff
@@ -91,6 +92,7 @@ impl SharedState {
             max_inner_size: None,
             resize_increments: None,
             base_size: None,
+            has_focus: true,
```
kchibisov, line 95: Shouldn't this be `false`?

src/platform_impl/linux/wayland/window/shim.rs
```diff
@@ -151,6 +152,9 @@ pub struct WindowHandle {
     /// Whether the window is resizable.
     pub is_resizable: Cell<bool>,
 
+    /// Whether the window is resizable.
```
kchibisov, line 155:
```suggestion
    /// Whether the window has keyboard focus.
```

kchibisov (changes requested):
While there's a way to know whether the window is focus or not on ios/Android, I guess the issue was that it's unclear how to hook into the current handlers, which is sort of understandable.

Would leave that to android and ios maintainers.

src/platform_impl/linux/x11/window.rs
amrbashir: doesn't the window start initially with keyboard focus?

amrbashir:
Android and iOS are now implemented, although I am not sure if a global state on Android is the way to go.

kchibisov (approved):
LGTM.

Would like to hear from android maintainers wrt the change and tracking
focus in a global state. @MarijnS95 or @msiglreith.

Also, you'd need a rebase, I guess.

MarijnS95:
Android _can_ support multiple windows (called `Activity`s) so this _should not_ go into a `static` global.

amrbashir:
> Android can support multiple windows (called `Activitys`) so this should not go into a `static` global.

I am aware of that but glancing over the code, the events are dispatched with a dummy window id and there is no way to differentiate between different activity events.

MarijnS95:
That's unfortunate to hear @amrbashir, I thought @rib's solution in #2444 would implicitly address this; or at least his crate opens the door for multi-window (multi-`Activity` _instances_) setup as that's relevant to some of my use-cases going forward.

Regardless, I thought at least in the past we have a per-window `struct`/storage location where I stored the `RawWindowHandle` lock.
Looks like @rib removed that (probably for good reason, though I have not had the opportunity to look at any of the code in detail yet), and looking back it seems to have been in `EventLoop` which is already a better place than a global `static` (but not per-`Window`, as I remembered), allowing for easier move to a per-window `struct` when we get to implementing this.

src/platform_impl/ios/window.rs
```diff
@@ -348,6 +348,11 @@ impl Inner {
         None
     }
 
+    pub fn has_focus(&self) -> bool {
+        let is_key: Bool = unsafe { msg_send![self.window, isKeyWindow] };
+        is_key.as_bool()
```
madsmtm, line 353: Nit: using raw `bool` works

madsmtm (approved):
Haven't tested the iOS impl, but is very trivial, so accepted from my side. macOS also looks good, thanks for your work @amrbashir!

src/platform_impl/ios/window.rs
amrbashir: oh, good to know.

src/platform_impl/linux/x11/window.rs
madsmtm: On macOS, the window is initially not focused (and `window.has_focus()` returns `false`), but it receives a `WindowEvent::Focused` shortly after launching

amrbashir: This only affects x11, on macOS we query the OS directly. The reason I made x11 start with `true` is because Windows also starts with `true` and I think x11 is very similar to Windows in this matter, but I haven't had the chance to test x11 (I may test it later today).

kchibisov (approved):

kchibisov:
@amrbashir could you rebase it?

amrbashir:
Should be fine now

PR #2534, opened 2022-10-27
Add `WindowLevel` API

amrbashir:
  - [X] Windows
  - [ ] x11
  - [x] macOS

Porting https://github.com/tauri-apps/tao/commit/a2a7b7262cc55e4c6defb79d5f77efce9d7e386d


src/window.rs
```diff
@@ -322,10 +324,23 @@ impl WindowBuilder {
     /// See [`Window::set_always_on_top`] for details.
     #[inline]
     pub fn with_always_on_top(mut self, always_on_top: bool) -> Self {
+        self.window.always_on_bottom = false;
         self.window.always_on_top = always_on_top;
```
kchibisov, line 328: Could you we refactor this to use a single `with` method and use `enum`?

```rust
pub fn with_window_layer_preference(mut self, layer: LayerPreference);

pub enum LayerPreference {
    AlwaysOnTop,
    Normal,
    AlwaysOnBottom,
}
```

Could name it differently I don't like the **layer** in the name, but you should get the idea.

kchibisov (changes requested):

src/window.rs
amrbashir: maybe:
```rs
pub fn with_window_level(mut self, level: WindowLevel) -> Self;

pub enum WindowLevel { 
  AlwaysOnTop,
  Normal,
  AlwaysOnBottom
}
```

kchibisov: Yeah, that's better.

src/window.rs
```diff
@@ -1379,3 +1376,19 @@ impl Default for UserAttentionType {
...
+/// ## Platform-specific
+///
+/// - **iOS / Android / Web / Wayland:** Unsupported.
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum WindowLevel {
+    /// The window will always be below other windows,
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **Windows**: There is no guarantee that the window will be the bottom most but it will try to be
```
msiglreith, line 1389: In what scenario does it cause issues?

src/window.rs
```diff
@@ -1379,3 +1376,19 @@ impl Default for UserAttentionType {
         UserAttentionType::Informational
     }
 }
+
+/// ## Platform-specific
+///
+/// - **iOS / Android / Web / Wayland:** Unsupported.
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum WindowLevel {
```
msiglreith, line 1384: Should add a bit more documentation what exactly is meant with `WindowLevel` and how it affects window ordering in the Z-hierarchy

src/platform_impl/windows/window_state.rs
```diff
@@ -320,6 +321,24 @@ impl WindowFlags {
             }
         }
 
+        if diff.contains(WindowFlags::ALWAYS_ON_BOTTOM) {
+            unsafe {
+                SetWindowPos(
+                    window,
+                    match new.contains(WindowFlags::ALWAYS_ON_BOTTOM) {
+                        true => HWND_BOTTOM,
+                        false => HWND_NOTOPMOST,
```
msiglreith, line 330: This won't work correctly when switching from `AlwaysBottom` to  `AlwaysTop` as the diff will contain both flags.
With the current implementation it will end up in `HWND_NOTOPMOST` (==`WindowLevel::Normal`).

Both cases should be handled in one if condition and we only have one call to `SetWindowPos` which resolves the tri-state.

msiglreith (changes requested):

src/window.rs
amrbashir: I was thinking maybe when two windows try to be the bottom most but this could be the same for any OS and also for `AlwaysOnTop`

src/platform_impl/windows/window_state.rs
amrbashir: Good catch, let me think if [843fd86](https://github.com/rust-windowing/winit/pull/2534/commits/843fd865627ec01c96dfad004a3c2e4f5cf6ed70) is what you have in mind.

src/window.rs
```diff
@@ -1379,3 +1376,17 @@ impl Default for UserAttentionType {
         UserAttentionType::Informational
     }
 }
+
+/// Describes the window order in z-hierarchy.
```
msiglreith, line 1380: Something more descriptive imo - trying to emphasize the grouping aspect of the window levels

```suggestion
/// A window level groups windows with respect to their z-position.
///
/// The relative ordering between windows in different window levels is fixed.
/// The z-order of a window within the same window level may change dynamically on user interaction.
```

src/platform_impl/windows/window_state.rs
```diff
@@ -302,13 +303,19 @@ impl WindowFlags {
             }
         }
 
-        if diff.contains(WindowFlags::ALWAYS_ON_TOP) {
+        if diff.contains(WindowFlags::ALWAYS_ON_TOP) || diff.contains(WindowFlags::ALWAYS_ON_BOTTOM)
```
msiglreith, line 306:
```suggestion
        if diff.intersects(WindowFlags::ALWAYS_ON_TOP | WindowFlags::ALWAYS_ON_BOTTOM)
```

examples/multithreaded.rs
```diff
@@ -66,7 +66,7 @@ fn main() {
                         let state = !modifiers.shift();
                         use VirtualKeyCode::*;
                         match key {
-                            A => window.set_always_on_top(state),
+                            A => window.set_window_level(WindowLevel::AlwaysOnBottom),
```
msiglreith, line 69: maybe `Key1 => AlwaysOnTop`, `Key2 => AlwaysOnBottom` and `Key3 => Normal`
allowing to set all 3 states individually would be good

msiglreith (commented):
thanks, a few smaller suggestions from my side

src/window.rs
amrbashir: thanks, I really didn't know what to write there.

src/platform_impl/windows/event_loop.rs
```diff
@@ -1136,12 +1136,6 @@ unsafe fn public_window_callback_inner<T: 'static>(
                         }
                     }
                 }
-            } else if window_state
-                .window_flags
-                .contains(WindowFlags::ALWAYS_ON_BOTTOM)
-            {
-                let window_pos = &mut *(lparam as *mut WINDOWPOS);
-                window_pos.hwndInsertAfter = HWND_BOTTOM;
```
msiglreith: why was this removed? I thought this is necessary to ensure it stays at the bottom

amrbashir: According to this [SO answer](https://stackoverflow.com/a/65052538), it is but it is kind of an alternative way and the same is not done for always on top so I removed it for consistency.

src/platform_impl/macos/appkit/window.rs
```diff
@@ -267,6 +267,8 @@ use window_level_key::*;
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
 #[repr(isize)]
 pub enum NSWindowLevel {
+    #[doc(alias = "BelowNormalWindowLevel")]
+    BelowNormal = (kCGBaseWindowLevelKey - 1) as _,
```
madsmtm, line 271: I looked into how window levels actually work, and it turns out that what we're doing currently is wrong, see https://github.com/rust-windowing/winit/pull/2545.

I'm fine with the implementation of `BelowNormal` though, but it'll have to be updated to use `kCGNormalWindowLevel - 1` after that PR lands.

src/window.rs
```diff
@@ -1379,3 +1376,20 @@ impl Default for UserAttentionType {
...
+/// ## Platform-specific
+///
+/// - **iOS / Android / Web / Wayland:** Unsupported.
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum WindowLevel {
+    /// The window will always be below other windows,
+    AlwaysOnBottom,
+    Normal,
+    /// The window will always be on top of other windows,
+    AlwaysOnTop,
```
madsmtm, line 1394:
```suggestion
    /// The window will always be below normal windows,
    AlwaysOnBottom,
    Normal,
    /// The window will always be on top of normal windows,
    AlwaysOnTop,
```

Modals, popups and other such windows will still be above on macOS

madsmtm (changes requested):
macOS and iOS looks fine, barring the note.

May I ask what the use-case is for `WindowLevel::AlwaysOnBottom`? (And maybe we could add that to the documentation)

src/window.rs
```diff
@@ -821,14 +821,11 @@ impl Window {
         self.window.is_decorated()
     }
 
-    /// Change whether or not the window will always be on top of other windows.
-    ///
-    /// ## Platform-specific
+    /// Change the window level.
     ///
```
kchibisov, line 825: Could you add a note that it's just a hint and the system could ignore the request? Could put it on a enum variant.

src/window.rs
```diff
@@ -1379,3 +1376,20 @@ impl Default for UserAttentionType {
...
+///
+/// The relative ordering between windows in different window levels is fixed.
+/// The z-order of a window within the same window level may change dynamically on user interaction.
+///
+/// ## Platform-specific
+///
+/// - **iOS / Android / Web / Wayland:** Unsupported.
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum WindowLevel {
+    /// The window will always be below other windows,
```
kchibisov, line 1390: Add dots at the end, not commas.

src/window.rs
```diff
@@ -1379,3 +1376,20 @@ impl Default for UserAttentionType {
...
+    AlwaysOnBottom,
+    Normal,
```
kchibisov, line 1392:
```suggestion
    
    Normal,
```

kchibisov (changes requested):

src/window.rs
```diff
@@ -1389,3 +1388,21 @@ impl Default for UserAttentionType {
...
+/// A window level groups windows with respect to their z-position.
+///
+/// The relative ordering between windows in different window levels is fixed.
+/// The z-order of a window within the same window level may change dynamically on user interaction.
+///
+/// ## Platform-specific
+///
+/// - **iOS / Android / Web / Wayland:** Unsupported.
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum WindowLevel {
```
kchibisov, line 1401: Could you impl `Default` for it and then use `Default::default()` instead of `Normal`?

src/window.rs
```diff
@@ -821,14 +822,12 @@ impl Window {
         self.window.is_decorated()
     }
 
-    /// Change whether or not the window will always be on top of other windows.
-    ///
-    /// ## Platform-specific
+    /// Change the window level.
+    /// This is just a hint to the OS, and the system could ignore it.
```
kchibisov, line 826:
```suggestion
    /// Change the window level.
    ///
    /// This is just a hint to the OS, and the system could ignore it.
```

src/window.rs
```diff
@@ -315,14 +315,15 @@ impl WindowBuilder {
         self
     }
 
-    /// Sets whether or not the window will always be on top of other windows.
+    /// Sets the window level.
+    /// This is just a hint to the OS, and the system could ignore it.
```
kchibisov, line 319:
```suggestion
    /// Sets the window level.
    ///
    /// This is just a hint to the OS, and the system could ignore it.
```

kchibisov (changes requested):

src/platform_impl/macos/appkit/window.rs
madsmtm: I've merged that PR now, so you should be able to rebase now

madsmtm: (Perhaps name the new constant something like `BELOW_NORMAL` to differentiate it from the other ones that are meant to closely mimic the ones available in the system's `NSWindowLevel`)

amrbashir: Sure

src/window.rs
```diff
@@ -1389,3 +1390,27 @@ impl Default for UserAttentionType {
...
+///
+/// ## Platform-specific
+///
+/// - **iOS / Android / Web / Wayland:** Unsupported.
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum WindowLevel {
+    /// The window will always be below normal windows.
+    AlwaysOnBottom,
+
+    Normal,
```
madsmtm, line 1407: Nit:
```suggestion
    /// The default.
    Normal,
```

src/window.rs
```diff
@@ -1389,3 +1390,27 @@ impl Default for UserAttentionType {
...
+/// The relative ordering between windows in different window levels is fixed.
+/// The z-order of a window within the same window level may change dynamically on user interaction.
+///
+/// ## Platform-specific
+///
+/// - **iOS / Android / Web / Wayland:** Unsupported.
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum WindowLevel {
+    /// The window will always be below normal windows.
+    AlwaysOnBottom,
```
madsmtm, line 1405: It is not immediately clear why you would want to use this level, could we add use-cases for this to the documentation?

madsmtm (approved):
Approved from my side, thanks!

src/window.rs
amrbashir: It is hard for me to come up with a good use-case but I think this will be useful for a widgets app.

kchibisov (approved):

kchibisov:
Thanks!

Issue #2412, opened 2022-08-08
Implement VSync source


kchibisov:
This is a common concept in e.g. browsers to have non blocking frame scheduling at the display refresh rate.

It seems like it's possible on all desktop platforms and the web. For the rest we may indicate that vsync source is not supported.

The way to integrate it would be to add `WindowEvent::Frame` which must be send when the new frame is ready to be drawn. That will come from the compositor directly and the users must request those event via `Window::request_frame() -> Result<()>`, where the result will indicate that the event can't be scheduled due to reasons.

The Api that should be used for that on platform are the following.

X11 - Xpresent
macOS - `DisplayLink`
Wayland - frame callbacks
Windows - DWM get composition timings + a timer.
Web - request animation frame.

The Api is essential for Wayland clients, since they must not rely on vsync from the graphics Api.

dhardy:
Sounds good to me, though it would be worth considering how failures of `request_frame` should be handled (e.g. requesting a wake in 16ms?).

kchibisov:
>should be handled (e.g. requesting a wake in 16ms?).

I think it's up to application what to do here. Like it clearly knows how that it failed before hand, so can just fallback to something else. I don't want to add timers to every backend, just to provide some hand holding. At least for now. In general this can only fail on X11 if you don't have `libXpresent` installed, so it's like you should really try to fail...

rib:
Based on discussing this a bit on IRC I think it could be good to highlight one clear sub-problem that came up which is that it could be really helpful to have just _one_ place where applications are allowed to render.

e.g. in the case of Wayland, if it was guaranteed that apps only rendered during `RedrawRequested` event callbacks then the backend could potentially register frame callbacks there to get notifications from the compositor for throttling the application.

(so potentially frame throttling might be possible on Wayland with no new API if it were documented that rendering were limited to `RedrawRequested` events. Similarly this should be possible for X11).

Thinking more generally about high level APIs related to rendering synchronization may open a bigger can of worms, which could be good to try and break down.

One initial concern I have about a `request_frame()` API that seems inspired by `requestAnimationFrame` in browsers is that it looks kind of back-to-front compared to typical presentation APIs that would throttle / synchronize you based on a previously submitted frame.

[...]

PR #2535, opened 2022-10-27
Add frame throttling source

kchibisov:
This commit adds `Window::request_frame_throttling_hint` request and the corresponding `WindowEvent::CanRedrawFrame` event.

The request is intended to be used as an alternative to timer based rendering and vsync extensions.

For now only X11 support is implemented.

The related issue.

https://github.com/rust-windowing/winit/issues/2412

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,8 @@ And please only add new entries to the top of this list, right below the `# Unre
 - **Breaking:** Removed `WindowBuilderExtWindows::with_theme` and `WindowBuilderExtWayland::with_wayland_csd_theme` in favour of `WindowBuilder::with_theme`.
 - **Breaking:** Removed `WindowExtWindows::theme` in favour of `Window::theme`.
 - Enabled `doc_auto_cfg` when generating docs on docs.rs for feature labels.
+- **Breaking:** Added `WindowEvent::CanRedrawFrame` event as a frame throttling hint event to help driving rendering loop.
+- Added `Window::request_frame_throttling_hint` which can be used to request frame throtting hints from the system compositor, currently implemented on X11.
 
 # 0.27.5
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -105,7 +105,7 @@ wayland-protocols = { version = "0.29.4", features = [ "staging_protocols"], opt
 sctk = { package = "smithay-client-toolkit", version = "0.16.0", default_features = false, features = ["calloop"],  optional = true }
 sctk-adwaita = { version = "0.5.1", default_features = false, optional = true }
 mio = { version = "0.8", features = ["os-ext"], optional = true }
-x11-dl = { version = "2.18.5", optional = true }
+x11-dl = { version = "2.20.0", optional = true }
 percent-encoding = { version = "2.0", optional = true }
 libc = "0.2.64"
 
--- a/src/event.rs
+++ b/src/event.rs
@@ -515,6 +515,18 @@ pub enum WindowEvent<'a> {
     /// Platform-specific behavior:
     /// - **iOS / Android / Web / Wayland / Windows:** Unsupported.
     Occluded(bool),
+
+    /// A redraw throttling hint which is sent by the system compositor when the redraw operation
+    /// should be performed. This event is useful when you want contiguous non-blocking drawing
+    /// at the refresh rate of the monitor without using `Vsync` extension and relying on a timer
+    /// based rendering.
+    ///
+    /// The only way to obtain it is by using [`Window::request_frame_throttling_hint`].
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **iOS / Android / Web / Wayland / Windows:** Unsupported.
+    CanRedrawFrame,
 }
 
 impl Clone for WindowEvent<'static> {
@@ -623,6 +635,7 @@ impl Clone for WindowEvent<'static> {
                 unreachable!("Static event can't be about scale factor changing")
             }
             Occluded(occluded) => Occluded(*occluded),
+            CanRedrawFrame => CanRedrawFrame,
         };
     }
 }
@@ -727,6 +740,7 @@ impl<'a> WindowEvent<'a> {
             ThemeChanged(theme) => Some(ThemeChanged(theme)),
             ScaleFactorChanged { .. } => None,
             Occluded(occluded) => Some(Occluded(occluded)),
+            CanRedrawFrame => Some(CanRedrawFrame),
         }
     }
 }
--- a/src/platform_impl/linux/mod.rs
+++ b/src/platform_impl/linux/mod.rs
@@ -529,6 +529,11 @@ impl Window {
     }
 
     #[inline]
+    pub fn request_frame_throttling_hint(&self) -> Result<(), ExternalError> {
+        x11_or_wayland!(match self; Window(w) => w.request_frame_throttling_hint())
+    }
+
+    #[inline]
     pub fn current_monitor(&self) -> Option<MonitorHandle> {
         match self {
             #[cfg(feature = "x11")]
--- a/src/platform_impl/linux/wayland/window/mod.rs
+++ b/src/platform_impl/linux/wayland/window/mod.rs
@@ -367,6 +367,11 @@ impl Window {
     }
 
     #[inline]
+    pub fn request_frame_throttling_hint(&self) -> Result<(), ExternalError> {
+        Err(ExternalError::NotSupported(NotSupportedError::new()))
+    }
+
+    #[inline]
     pub fn outer_size(&self) -> PhysicalSize<u32> {
         self.size
             .lock()
--- a/src/platform_impl/linux/x11/event_processor.rs
+++ b/src/platform_impl/linux/x11/event_processor.rs
@@ -28,6 +28,7 @@ pub(super) struct EventProcessor<T: 'static> {
     pub(super) ime_receiver: ImeReceiver,
     pub(super) ime_event_receiver: ImeEventReceiver,
     pub(super) randr_event_offset: c_int,
+    pub(super) xpresent_event_offset: Option<c_int>,
     pub(super) devices: RefCell<HashMap<DeviceId, Device>>,
     pub(super) xi2ext: XExtension,
     pub(super) target: Rc<RootELW<T>>,
@@ -653,7 +654,31 @@ impl<T: 'static> EventProcessor<T> {
                 } else {
                     return;
                 };
+
                 let xev = &guard.cookie;
+
+                // Handle xpresent.
+                if self.xpresent_event_offset == Some(xev.extension)
+                    && xev.evtype == ffi::PresentCompleteNotify
+                {
+                    let xev: &ffi::XPresentCompleteNotifyEvent =
+                        unsafe { &*(xev.data as *const _) };
+
+                    let window_id = mkwid(xev.window);
+                    if wt.windows.borrow().contains_key(&window_id.0) {
+                        // Remove the event input.
+                        let _ = wt
+                            .xconn
+                            .xpresent_free_input(xev.window, xev.eid as ffi::XID);
+
+                        callback(Event::WindowEvent {
+                            window_id,
+                            event: crate::event::WindowEvent::CanRedrawFrame,
+                        });
+                    }
+                }
+
+                // Handle input.
                 if self.xi2ext.opcode != xev.extension {
                     return;
                 }
@@ -1194,7 +1219,9 @@ impl<T: 'static> EventProcessor<T> {
                         }
                     }
 
-                    _ => {}
+                    _ => {
+                        println!("Here")
+                    }
                 }
             }
             _ => {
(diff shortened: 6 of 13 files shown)


rib:
I'm quite surprised that this is touching the X Present protocol.

The last I saw of that protocol it was a draft that Keithp was working on while still in the Intel graphics team which he left before finishing this work. At that time there was relatively little interest left in the community to work on this kind of things so I'm quite surprised that it got developed further in the end.

The present protocol is something that's more intended to be used at the same level as the DRI2/3 protocols - i.e. it addresses shortcomings we found in DRI3. I would expect it's more applicable for compositors and for clients I'd expect it to be used in the graphics driver (e.g. Mesa),

I would need to check into this more but I'd expect that subscribing to these present complete notifications is only going to work if you happen to be running on a driver that's using the present protocol internally - but you can't just assume that surely?

My initial instinct here would have been to look at these protocols from Owen Taylor here: https://fishsoup.net/misc/wm-spec-synchronization.html. Those protocols were defined during a kind of heyday for Linux X11 graphics stack development in collaboration between Red Hat and Intel and I'd expect them to be most widely supported in compositors (e.g in Gnome Shell / KDE).

I think those protocols are probably a bit more appropriate to use within Winit, outside of the driver.

kchibisov:
>I'm quite surprised that this is touching the X Present protocol.

It's there for a long time actually, since 2015 and used. Though there are pretty few users in the wild probably due to lack of advertising.

>I would need to check into this more but I'd expect that subscribing to these present complete notifications is only going to work if you happen to be running on a driver that's using the present protocol internally - but you can't just assume that surely?

I'm not sure how it works internally, I'd assume compositor support here? I know that it works with XWayland for example out of the box and and is delivered similar to frame callbacks resulting in identical frame rates between frame callbacks and xpresent driven rendering.

The calls could return error and that's intentional, since such information may be missing or not supported by the system.


>My initial instinct here would have been to look at these protocols from Owen Taylor here: https://fishsoup.net/misc/wm-spec-synchronization.html. Those protocols were defined during a kind of heyday for Linux X11 graphics stack development in collaboration between Red Hat and Intel and I'd expect them to be most widely supported in compositors (e.g in Gnome Shell / KDE).

I went with `Xpresent` since at some point of time(in `#wayland`) I was advised to use `Xpresent` to frame callback like rendering. I'm not familiar with `Xsync` but it seems that it could be used similarly.

rib:
There's hopefully some actively maintained reference code for using these protocols somewhere (e.g. maybe in gtk/gdk) but here's some old C code I wrote in the past that uses these protocols, that you might be able to decipher / steal from:

https://github.com/rib/cglib/blob/rig/cglib/winsys/cg-winsys-egl-x11.c

That [cglib](https://github.com/rib/cglib), or the slightly earlier [cogl](https://github.com/rib/cogl) may even have other things that are potentially of interest in this context of Winit and X11 in particular, considering that it was actively developed at the time that these protocols were defined for the Gnome 3 desktop.

Since energy and interest moved on from that, e.g. as folks started working on Wayland then relatively little will have changed with these things since then, and I expect there are some things that are only really documented in the code of these old projects.

Incidentally those projects also supported similar frame synchronization for Wayland which may potentially be helpful to poke through.

rib:
> > I'm quite surprised that this is touching the X Present protocol.
> 
> It's there for a long time actually, since 2015 and used. Though there are pretty few users in the wild probably due to lack of advertising.

yeah, seems so - just quite surprising for me.

It's not as much of a surprise though that there aren't many users - on the one hand I don't think there are many people left that are actively working with x11 at this level but also I think this protocol is only really intended to be used for things like EGL and GLX (e.g. from within Mesa).

For a while keithp was working with Valve, looking to knock x11 into shape for better VR support and I wonder if maybe finishing this off tied into that work at the time.

> 
> > I would need to check into this more but I'd expect that subscribing to these present complete notifications is only going to work if you happen to be running on a driver that's using the present protocol internally - but you can't just assume that surely?
> 
> I'm not sure how it works internally, I'd assume compositor support here? I know that it works with XWayland for example out of the box and and is delivered similar to frame callbacks resulting in identical frame rates between frame callbacks and xpresent driven rendering.

I had a poke around e.g. in mutter (gnome shell) and the xorg xserver to double check this. It's a protocol that's supported in the xorg xserver, not the compositor - which is partly also why I wouldn't think it's a great fit here, since it doesn't really let you sync with your compositor (except if running on something like xwayland). The protocol is essentially a DRI3 extension which is what your EGL or GLX drivers use.

Yeah xwayland integrates with present in the same xorg xserver repo. It generates fake msc events and integrates with the wayland compositor's frame callbacks as I understand it from skimming through the code.

As far as I know DRI3 was never supported by Nvidia and I also guess the situation is the same with this Present extension. At least James Jones did give feedback on the spec but I doubt very much that they implemented it.

> 
> The calls could return error and that's intentional, since such information may be missing or not supported by the system.
> 
> > My initial instinct here would have been to look at these protocols from Owen Taylor here: https://fishsoup.net/misc/wm-spec-synchronization.html. Those protocols were defined during a kind of heyday for Linux X11 graphics stack development in collaboration between Red Hat and Intel and I'd expect them to be most widely supported in compositors (e.g in Gnome Shell / KDE).
> 
> I went with `Xpresent` since at some point of time(in `#wayland`) I was advised to use `Xpresent` to frame callback like rendering. I'm not familiar with `Xsync` but it seems that it could be used similarly.

In the _specific_ case of an x11 application that's known to run on xwayland then Xpresent can kind of do what you're looking for I think, since it would indirectly let the client integrate with the real Wayland compositor's frame callbacks - as a replacement for being driven by the vblank.

That's not consistent though with how it would work for an x11 client running on an x11 compositor - in that case they don't get indirect feedback from the x11 compositor.

If you dig through some of the discussions for the spec then there were ideas to extend it with the concept of "Present redirection" (e.g. see https://lists.x.org/archives/xorg-devel/2013-October/038262.html) which would have better connected x11 clients with an x11 compositor but it looks like those ideas got dropped.

Even in for the assumed xwayland case though it does seems like a notable layering violation for an x11 application to be touching the present protocol outside of the driver - it's more like it should be the basis of an EGL/GLX extension that you could use, besides just using it when calling egl/glxSwapBuffers.

E.g. it looks like the GLX driver may conflict with this; hooking into `PresentComplete` events for `GLX_BufferSwapComplete` 
 https://lists.x.org/archives/xorg-devel/2013-November/039176.html 

The fact that EGL/GLX uses e.g. DRI2/3/Present isn't really something that applications are supposed to know / assume, since it was generally only true for open source drivers, and Nvidia had their own equivalent protocols. Even if assuming running on open source drivers the availability of extensions like these depends on the version of the xserver, graphics drivers and client libraries which shouldn't really be assumed by something like Winit.

The protocols like `_NET_WM_FRAME_DRAWN` linked above are intended to be used outside of the drivers by middleware layers such as Cogl linked above, GDK or SDL and may be a good fit for use in Winit here.

kchibisov:
Thanks for your input here, it's really helpful to me with how to approach X11 interface for that given that docs for it are either lacking or you should clearly know where to look at.

>The fact that EGL/GLX uses e.g. DRI2/3/Present isn't really something that applications are supposed to know / assume, since it was generally only true for open source drivers, and Nvidia had their own equivalent protocols. Even if assuming running on open source drivers the availability of extensions like these depends on the version of the xserver, graphics drivers and client libraries which shouldn't really be assumed by something like Winit.

>The protocols like _NET_WM_FRAME_DRAWN linked above are intended to be used outside of the drivers by middleware layers such as Cogl linked above, GDK or SDL and may be a good fit for use in Winit here.

I'll try to look into it and see how it goes, but from what I see it seems to be similar to what I do with xpresent.

rib:
No worries - I hope it's at least constructive input.

At least for some of this kind of stuff I was partially involved in its design/development (some of these protocols were effectively developed in support of a (failed) mobile OS called moblin/meego/tizen (depending on the year) largely done between Intel (where I was) and Red Hat (since it was also the foundation for Gnome 3).

For things that weren't just discussed internally or just on IRC then it's sometimes possible to dredge history up from freedesktop.org mailing lists, or old gnome / freedesktop.org bugzilla archives for this kind of thing. (and digging through the git log for mesa / xorg xserver etc)

For these protocols in particular I would have been working closely with Owen at the time and might be able to dig up more details/context if we get stuck with something here.

kchibisov:
@rib from what I've read and managed to find wrt `_NET_WM_FRAME_DRAWN` it seems that it's only supported by `mutter` and `gtk`/`cogl`. No other compositor/wm has support for it meaning that all of that is sort of useless, like there? Maybe I understand it wrong though and it's not only about the `_NET_WM_FRAME_DRAWN`. So from what I can see it won't work with compositors liku `picom` or plain X server.

I don't understand why using X present results in like 300 fps with picom and not even approaching monitor refresh rate without compositor. I have a feeling that the issue is that it has something to do with interfering with GLX, like you said?

It kind of feels like X present is sort of special and should be avoided, and Xpresent(or used via the _sync_contol extension, that's strange though) shouldn't be used at all. Maybe some different approach with X present should be taken?

rib:
I would have expected `_NET_WM_FRAME_DRAWN` to also be supported by KDE - maybe we can look into that to try and double check that.

For plain, non-composited X then at that point you would be able to rely on `egl/glxSwapInterval(1)` and `egl/glxSwapBuffers` to throttle but I don't think there would be a good way of getting async feedback, except maybe `GLX_INTEL_swap_event`.

yeah `_NET_WM_FRAME_DRAWN` is only relevant to running with an X compositor.

There certainly no unified solution that covers both composited and non-composited x11 clients unfortunately.

kchibisov:
The log with X present https://paste.rs/3Mv running with `picom` compositor. I don't understand why it sends the same `CompleteNotify` events(I'm not talking about doubled events,  but about events having the same `ust` and `mst`), maybe the issue is that I toggle it, but that shouldn't  be relevant.

The other part of it that XWayland works, but that was sort of expected. From what I understand the X present is used by `OMS_sync_control` thing to get timings.

rib:
Grepping through the KWin repo it indeed looks like they don't support `NET_WM_FRAME_DONE` :/ that's a shame.

Considering that Ubuntu defaults to Gnome Shell I would probably guess that Gnome Shell does basically cover _most_ Linux desktop users (and some of the other popular desktops use mutter or gnome shell under the hood) but kwin is surely also a lot of users too.

Something that some compositors do is to unredirect windows that are fullscreen (i.e. games) which should then be able to synchronize with vblank via `egl/glXSwapBuffers` and that might be one factor for why the lack of support for synchronizing composited clients might go (largely) unnoticed.

Quite a lot of windowed x11 apps may just use software rendering with APIs like cairo and likely just update based on UI interaction, instead of rendering continuously which would also make this less of an issue for a lot of cases.

It's all quite a big mess though unfortunately :-D

kchibisov:
It seems like chromium is using xpresent, but to only get triplet(ust/msc/sbc) and schedules event itself. The way it does that by asking for those events and removing select right away. However with a closer look they say that sometimes the drivers return garbage data so they say that getting vsync params is not available.

>Something that some compositors do is to unredirect windows that are fullscreen (i.e. games) which should then be able to synchronize with vblank via egl/glXSwapBuffers and that might be one factor for why the lack of support for synchronizing composited clients might go (largely) unnoticed.

The problem is that we want to sync to vblank in a non-blocking way. I'm not sure there's a way to do so with glx/x11, etc? On Wayland you simply drive by frame callbacks and call it a day.

rib:
> It seems like chromium is using xpresent, but to only get triplet(ust/msc/sbc) and schedules event itself. The way it does that by asking for those events and removing select right away. However with a closer look they say that sometimes the drivers return garbage data so they say that getting vsync params is not available.

hehe, wow :)

are they maybe using it in a context where they aren't _also_ using egl/glx? - that might be quite reasonable (and then it might just be a separate issue that some drivers have inconsistent definitions of the msc counters)

> 
> > Something that some compositors do is to unredirect windows that are fullscreen (i.e. games) which should then be able to synchronize with vblank via egl/glXSwapBuffers and that might be one factor for why the lack of support for synchronizing composited clients might go (largely) unnoticed.
> 
> The problem is that we want to sync to vblank in a non-blocking way. I'm not sure there's a way to do so with glx/x11, etc? On Wayland you simply drive by frame callbacks and call it a day.

There are the GLX_INTEL_swap_events that are likely to be supported by other open source drivers besides just intel, but almost certainly not Nvidia. If I recall right, that enabled non-composited apps to synchronize without blocking. 

but yeah it's all more of a mix and match approach with lots of special cases.

rib:
As the discussion has helped remind me of the challenge it was dealing with this stuff in the past I figured I'd at least try and break down the a list of cases that may involve different solutions:

- windowed client, x11 compositor, open source drivers
- full screen client, x11 compositor (potentially unredirected), open source drivers
- windowed client, x11 compositor, nvidia drivers
- full screen client, x11 compositor (potentially unredirected), nvidia drivers
- windowed client, xwayland assuming open source drivers (since wayland not generally supported on nvidia)
- full screen client, xwayland (potentially unredirected), assuming O/S drivers
- windowed client, non-composited, open source drivers
- full screen client, non-composited, open source drivers
- windowed client, non-composited, nvidia drivers
- full screen client, non-composited, nvidia drivers

This actually probably needs to be expanded further by considering applications rendering with Open GL[ES] with GLX vs EGL and then Vulkan, and then potentially consider applications that are simply software rendering.

For supporting X11 it may make sense to just try and pick some of the higher-priority, common cases and see if they can be supported, and potentially just leave other cases unsupported/unthrottled.

Unfortunately even with that idea it may be tricky in some cases where the current best solution may currently depend on GLX/EGL which isn't currently in-scope for Winit.

kchibisov:
I've opened a separate branch for that  #2883

PR #2537, opened 2022-10-28
On Windows and macOS, add API to enable/disable window controls

amrbashir:
	- [X] Windows
	- [x] macOS

Porting https://github.com/tauri-apps/tao/commit/a50fd867b3df033dff077f5320b4b7037b04e454

### DIFF
--- a/src/platform_impl/windows/window.rs
+++ b/src/platform_impl/windows/window.rs
@@ -1017,7 +1017,7 @@ where
     // WindowFlags::VISIBLE and MAXIMIZED are set down below after the window has been configured.
     window_flags.set(WindowFlags::RESIZABLE, attributes.resizable);
     // Will be changed later using `window.set_enabled_buttons` but we need to set a default here
-    // so the diffing later can work
+    // so the diffing later can work.
     window_flags.set(WindowFlags::CLOSABLE, true);
 
     let parent = match pl_attribs.parent {
--- a/src/platform_impl/windows/window.rs
+++ b/src/platform_impl/windows/window.rs
@@ -70,7 +70,9 @@ use crate::{
         window_state::{CursorFlags, SavedWindow, WindowFlags, WindowState},
         Fullscreen, Parent, PlatformSpecificWindowBuilderAttributes, WindowId,
     },
-    window::{CursorGrabMode, CursorIcon, Theme, UserAttentionType, WindowAttributes},
+    window::{
+        CursorGrabMode, CursorIcon, Theme, UserAttentionType, WindowAttributes, WindowButtons,
+    },
 };
 
 /// The Win32 implementation of the main `Window` object.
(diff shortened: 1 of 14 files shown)


kchibisov:
I have a general question wrt that. Is this about the decoration thing or you can actually disable `Minimize`/`Maximize`, ect? What the goal for `set_closable`, etc? If that's all about buttons and has nothing to do with the ability to close the window it should be named differently. From now I see that `set_closable(false)` prevents the window close?

amrbashir:
Yeah, this is mainly about disabling the buttons itself so it can't be clicked but the window can still be closed in other ways.

kchibisov:
Then we should explicitly say that it's all about buttons in decorations. You can use `bitflags` setter and getter here, no need for so many methods returning bool. It's also better to manage bitflags than a bunch of bools.

kchibisov (changes requested):
See above.

The bitflags should be used instead.
The api should explicitly mention decorations.

Sorry for that short of review here, don't have that many of time and I'm only capable to reveiw general changes in the API not the windows/macOS code anyway.

amrbashir:
Done, although I don't really like the naming, so I am open to suggestions.

src/window.rs
```diff
@@ -732,6 +745,26 @@ impl Window {
         self.window.is_resizable()
     }
 
+    /// Sets the enabled window buttons.
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **Wayland / X11:** Not implemented.
+    /// - **Web / iOS / Android:** Unsupported.
+    pub fn set_window_buttons(&self, buttons: WindowButtons) {
```
madsmtm, line 754: Naming idea: `[set_]enabled_buttons` or `[set_]visible_buttons` (`window` is at least redundant here)

madsmtm (commented):
Thanks! I agree with the bitflag API.

An example would be nice for testing the functionality?

src/window.rs
amrbashir: `[set_]enabled_buttons` sounds good, thanks.

kchibisov (approved):
Api change is fine. The rest is up to windows/macOS maintainers.

madsmtm (changes requested):
macOS impl looks good, haven't tested it though; would you mind adding an example for it?

amrbashir:
Sure

madsmtm (approved):
Thanks for the example - it enabled me to find a bug which I've pushed a fix for ;)

src/platform_impl/windows/window.rs
```diff
@@ -974,6 +1016,9 @@ where
     window_flags.set(WindowFlags::TRANSPARENT, attributes.transparent);
     // WindowFlags::VISIBLE and MAXIMIZED are set down below after the window has been configured.
     window_flags.set(WindowFlags::RESIZABLE, attributes.resizable);
+    // Will be changed later using `window.set_enabled_buttons` but we need to set a default here
+    // so the diffing later can work
```
kchibisov, line 1020:
```suggestion
    // Will be changed later using `window.set_enabled_buttons` but we need to set a default here
    // so the diffing later can work.
```

src/platform_impl/windows/window_state.rs
```diff
@@ -344,6 +349,21 @@ impl WindowFlags {
...
+                EnableMenuItem(
+                    GetSystemMenu(window, 0),
+                    SC_CLOSE,
+                    MF_BYCOMMAND
+                        | if new.contains(WindowFlags::CLOSABLE) {
+                            MF_ENABLED
+                        } else {
+                            MF_DISABLED
+                        },
+                );
```
kchibisov, line 363: Could we compose this on a separate line, so this is a bit more readable?

kchibisov (changes requested):
Just a few minor notes. Should be fine afterwards.

msiglreith (approved):
Tested successfully on Windows 10
(interesting that Windows seems to automatically adjust the style towards a tool window once minimize and maximize are disabled)

src/platform_impl/windows/window_state.rs
amrbashir: separating it on a new line didn't feel to me as much enhancement so I too the liberty to use `bool::then` in [`4ff0733`](https://github.com/rust-windowing/winit/pull/2537/commits/4ff07333d7a68fdf655996e5962dcc7c8c1c579f)

kchibisov (approved):
Could you rebase? Should be good to go afterwards.

amrbashir:
Done!

PR #2538, opened 2022-10-28
On Windows, Add `HitTest` event

amrbashir:

Ref: #2413


kchibisov:
I have a general question of what is the use for this sort of API? From what I see right now it seems like the user shouldn't care if the X button was clicked, since they have no control over it anyway? Why could you know that the border was pressed or it's windows specific and you want to filter out the events?

amrbashir:
At `tauri` we will only use this event for `tauri-egui` crate so we can make undecorated window resizable using the mouse. There is another way to do it by using the API proposed in https://github.com/rust-windowing/winit/pull/2003 (we are actually doing that in `wry` crate by passing cursor location from js) but it has a disadvantage, you can't support snap layout using it or at least my last attempt was unsuccessful.

kchibisov:
My understanding of undecorated window handling wrt resizes is that you manually decide on them and resize with. So basically #2003.

>you can't support snap layout using it or at least my last attempt was unsuccessful.

I'd assume you mean tiling to regions? Not sure what it means otherwise. From my understanding tiling to edges should be done by the system compositor and not the window. If you don't want to resize past the monitor boundaries you can likely check that during resize, but I'm not sure how you're limited here on windows.

amrbashir:
Snap layout is a Windows 11 feature, see https://learn.microsoft.com/en-us/windows/apps/desktop/modernize/apply-snap-layout-menu  See win32 solution in How do I fix it section. The window manager does handle the snapping but the snap layout UI won't show unless you respond to `WM_NCHT
ITTEST` with `HTMAXBUTTOM`

kchibisov:
>The window manager does handle the snapping but the snap layout UI won't show unless you respond to WM_NCHT ITTEST with HTMAXBUTTOM

Shouldn't winit do this under the hood without exposing the event to the user directly?

amrbashir:
The snap layout will automatically work for decorated window but when decorations is disabled, winit can't know where is the maximize button, maybe there is no custom maximize button at all.

kchibisov:
Can't you create windows specific api for that or it really requires an event? The application could tell winit where the button is or something like that.

amrbashir:
It is possible to add an api to let winit store the button coordinates and size and internally handle that but you'd miss out on free resize and drag handling unless `winit` handles resizing too internally, which iirc wasn't desirable last time I checked.

everdrone:
> It is possible to add an api to let winit store the button coordinates and size and internally handle that but you'd miss out on free resize and drag handling unless `winit` handles resizing too internally, which iirc wasn't desirable last time I checked.

Hi, I'm trying to replicate the behavior used by Electron with `titlebarOverlay` as described [here](https://www.electronjs.org/docs/latest/tutorial/window-customization#window-controls-overlay-macos-windows).

[source code for reference](https://github.com/electron/electron/blob/835ace243c2e655d9d57ab4b4c6f20fa05579615/shell/browser/api/electron_api_browser_window.cc#L345)

This is a pretty handy customization options used by many popular apps and would be amazing to have it implemented here.

galop1n:
Will this be merged ? undecorated window resizing seems to only be doable well though the hit test event !

Jasper-Bekkers:
We're in need of this as well, I've created a rebased version of this PR here: https://github.com/rust-windowing/winit/commit/f034f8bea7ce38244acea5e571f4f1a91d6c13bf

We'd like to see this merged to main as well, as we're drawing custom decorations.

PR #2540, opened 2022-10-31
Bump `windows-sys` to 0.42

Awpteamoose:


msiglreith (approved):
thanks! `parking_lot` seems to also upgrade to 0.42 so it makes sense

Issue #2506, opened 2022-09-30
Provide support Xbox Series X|S based on windows-rs


HeavyRain266:
After series of tests on console's devkit made by friend's studio, It's mainly possible to upstream Xbox as the new supported platform (NO CI)

Implementation is simple, mostly copy-paste windows code under `xbox` module which requires disabling GDI support in windows-rs and removal of related functions and some other stuff.

**NOTE**: I cannot create Pull Request at the moment because I'm waiting for private devkit for my studio @eukaryote-interactive.

madsmtm:
Could you provide some more information about the platform? How is it supported by Rust (target info), links to how to use it in practice (`cargo` tools), example of how the user would use it, does it work with graphics libraries like `wgpu`?

HeavyRain266:
It's pretty much modified Windows based on the same win32 API (used to be UWP), it doesn't require compiler port unlike Switch or PlayStation 5. What you need is, generated target JSON file.

Input is a bit different story. We will have to introduce gamepad support using standard xinput from Windows APIs

From the tests, works perfectly fine with `wgpu`'s DX12 renderer except rare cases where `winapi` links random DLLs which are unsupported there but they have plans for `windows-rs` switch in `d3d12-rs` crate. What you need to create the window is to disable `GDI` feature in `windows-rs` then remove related functions and use something different from same library.

madsmtm:
Hmm, we've long been working on a rewrite of the keyboard event handling, so maybe see the probably a bit outdated branch `new-keyboard` and/or the two keyboard/gamepad issues https://github.com/rust-windowing/winit/issues/1806 and https://github.com/rust-windowing/winit/issues/944, perhaps the problem is already solved by those?

Whether or not that is the case, we could probably accept code to support it. It sounds like there is a lot of duplication with the normal Windows platform, and that this is a rather niche use-case, so perhaps it would be better to have it as a cargo feature?

HeavyRain266:
If compiler licensing team will agree on adding it as new target then it will be much easier than cargo feature. For now I have to dig more into getting device context because using plain win32 for xbox is wild guessing and testing since it's not documented. Microsoft only provides documenter GDKX wrapper which is pre-compiled and window crearion ie single call in main function.

For the time being, I will look into introducing gamepad support through xinput.

HeavyRain266:
[...]

PR #2543, opened 2022-11-01
Initial Xbox Gamepad support

HeavyRain266:
Initial change for #2506 which in the future hopefuly will introduce Xbox platform based on Win32/WinRT APIs which are not under NDA.

> **Note** [`Windows.Gaming.Input`](https://learn.microsoft.com/en-us/uwp/api/windows.gaming.input?view=winrt-22621) API depends on focused window and [`Gamepad`](https://learn.microsoft.com/en-us/uwp/api/windows.gaming.input.gamepad?view=winrt-22621) module provides Xbox specific metadata such as assigned users, which is required for upstream console compatibility.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -829,6 +829,16 @@ pub struct KeyboardInput {
     pub modifiers: ModifiersState,
 }
 
+// Describes a gamepad input event.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct GamepadInput {
+    // Identifies physical button pressed
+    pub button: ButtonId,
+
+    pub state: ElementState,
+}
+
 /// Describes [input method](https://en.wikipedia.org/wiki/Input_method) events.
 ///
 /// This is also called a "composition event".


HeavyRain266:
@msiglreith Could you help me model this feature as windows specific to not interact with other platforms?

(I've been told on matrix to ping you here)

msiglreith:
Have you looked into the old gamepad PR already? https://github.com/rust-windowing/winit/pull/1969
In general it would be nice to keep the user facing API similar if possible but I haven't looked into the new gaming API at all so far besides the comments in the issue

HeavyRain266:
> Have you looked into the old gamepad PR already? https://github.com/rust-windowing/winit/pull/1969
> In general it would be nice to keep the user facing API similar if possible but I haven't looked into the new gaming API at all so far besides the comments in the issue

I have not but someone linked old branch with implementation based on HID instead, thanks for porting this one out.

New API is a bit different from Xinput as of being rewritten in C++ with more support for more kinds of gaming input than gamepads and requipment for focused window. Users are mostly using gilrs instead which doesn't require any changes in winit itself for cross platform gamepad handling. There is some info about [migration](https://learn.microsoft.com/en-us/gaming/gdk/_content/gc/input/porting/input-porting-xinput), could be mapped 1:1 with previous PR but also add API features required for xbox compatibility on top.

HeavyRain266:
Unfortunatelly, I'm closing this for now, as I have no time to rebase it and then start testing the new API. I will get back to it in coming months as it requires deeper knowledge of winit internals and a lot of planning and testing.

PR #2544, opened 2022-11-02
On macOS, fix `WindowEvent::Destroyed` isn't emitted 

wusyong:
It seems `windowWillClose` can only be called when `windowShouldClose` returns true.
So the `WindowEvent::Destroyed` is never called when the window is dropped.
This PR add another queue event when the window is dropped during Drop trait is called.

### DIFF
--- a/src/platform_impl/macos/util/async.rs
+++ b/src/platform_impl/macos/util/async.rs
@@ -8,11 +8,15 @@ use objc2::rc::{autoreleasepool, Id, Shared};
 
 use crate::{
     dpi::LogicalSize,
+    event::{Event, WindowEvent},
     platform_impl::platform::{
+        app_state::AppState,
         appkit::{NSScreen, NSWindow, NSWindowLevel, NSWindowStyleMask},
+        event::EventWrapper,
         ffi,
-        window::{SharedState, SharedStateMutexGuard},
+        window::{SharedState, SharedStateMutexGuard, WinitWindow},
     },
+    window::WindowId,
 };
 
 // Unsafe wrapper type that allows us to dispatch things that aren't Send.
@@ -218,11 +222,17 @@ pub(crate) fn set_title_async(window: &NSWindow, title: String) {
 //
 // ArturKovacs: It's important that this operation keeps the underlying window alive
 // through the `Id` because otherwise it would dereference free'd memory
-pub(crate) fn close_async(window: Id<NSWindow, Shared>) {
+pub(crate) fn close_async(window: Id<WinitWindow, Shared>) {
     let window = MainThreadSafe(window);
     Queue::main().exec_async(move || {
         autoreleasepool(move |_| {
             window.close();
+
+            let event = Event::WindowEvent {
+                window_id: WindowId(window.id()),
+                event: WindowEvent::Destroyed,
+            };
+            AppState::queue_event(EventWrapper::StaticEvent(event));
         });
     });
 }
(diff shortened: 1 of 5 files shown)


src/platform_impl/macos/util/async.rs
```diff
@@ -218,11 +222,17 @@ pub(crate) fn set_title_async(window: &NSWindow, title: String) {
...
     let window = MainThreadSafe(window);
     Queue::main().exec_async(move || {
         autoreleasepool(move |_| {
             window.close();
+
+            let event = Event::WindowEvent {
+                window_id: WindowId(window.id()),
+                event: WindowEvent::Destroyed,
+            };
+            AppState::queue_event(EventWrapper::StaticEvent(event));
```
madsmtm, line 235: Nit: Use `window.emit_event`

madsmtm (changes requested):
Huh, odd that this hasn't been discovered before! Does this mean that our overriding of `windowWillClose:` is effectively useless?

However, I doubt this fixes the problem completely, since the window may be closed before the last reference to it is destroyed (for example in the `window_debug` example). Do you have an idea for how to fix this?

src/platform_impl/macos/util/async.rs
wusyong: Only `WinitWindowDelegate` has this method. Should I add such method to `WinitWindow` too? Or could I just access `WinitWindowDelegate` by `pub(crate)` it.

wusyong:
> Does this mean that our overriding of windowWillClose: is effectively useless?

I couldn't find any other way to trigger `windowWillClose:` than returning true in `windowShouldClose:`.

> However, I doubt this fixes the problem completely, since the window may be closed before the last reference to it is destroyed (for example in the window_debug example). Do you have an idea for how to fix this?

Yeah, `Destroyed` event won't be sent in this case but it's consistent across platforms (at least on x11 and windows). But I assume that's okay if you are already going to exit.  In my use case, I treat `Destroyed` event as a way to gracefully shutdown. I have a collection of window (HasMap for example). `CloseRequested` will drop the window in the collection and `set_exit` if the collection is empty during `Destroyed`.

src/platform_impl/macos/util/async.rs
madsmtm: Just do `pub(crate)`

madsmtm:
> I couldn't find any other way to trigger `windowWillClose:` than returning true in `windowShouldClose:`.

Then we should maybe remove the code in it, along with a note? Though I'm a bit weary about doing so, since it seems like, see this commit from 2018 that added the note about El Capitan https://github.com/goddessfreya/winit/commit/8e9fc01bd6b404f59488b130413f48e4e5f89b0d (from PR https://github.com/rust-windowing/winit/pull/853).

@francesca64 do you remember anything about this?

> Yeah, `Destroyed` event won't be sent in this case but it's consistent across platforms (at least on x11 and windows).

Would you mind documenting this on `Event::Destroyed`, then?

francesca64:
@madsmtm I sadly don't remember, sorry!

madsmtm:
No worries, thanks for the quick response!

src/platform_impl/macos/util/async.rs
wusyong: @madsmtm I got a real weird behavior if I move a shared pointer of `delegate` into `exec_async`, `windowWillClose` will be called somehow!
So I can just call `delegate.window.close()` instead in latest commit. Wdyt?

wusyong: I found the delegate has to be used in `exec_async` somehow, so `windowWillCose` can be invoked.
So I call `delegate.window.close()` instead which should look a little bit cleaner.

madsmtm:
I've discovered that `windowWillClose:` is called if you run `close:` synchronously, which means this should be fixed by https://github.com/rust-windowing/winit/pull/2574

wusyong:
@madsmtm I slightly modified multiwindow example and check `windows` hash map on `Destroyed` event.
It can work based on your PR! I can close this one if you are going to merge #2574.

PR #2545, opened 2022-11-02
macOS: Fix `NSWindowLevel` values

madsmtm:
Fix our handling of `NSWindowLevel` values.

This worked previously because we only used `NSWindowLevel::Normal` and `NSWindowLevel::Floating`, whose values happened to be almost correct anyways.

Discovered while reviewing https://github.com/rust-windowing/winit/pull/2534


kchibisov:
The formatting should be fixed.

Issue #2520, opened 2022-10-18
macOS fullscreen toggle is broken


rofferom:
## Problem description

Mutiple toggle of fullscreen on macOS won't work:
```
.set_fullscreen(Some(window::Fullscreen::Borderless(monitor)))
.set_fullscreen(None)
.set_fullscreen(Some(window::Fullscreen::Borderless(monitor)))
```

The first and the second calls are working, but the last one does nothing.

## Analysis try

After a quick `git bisect`, it seems that #2458 (d67c9281) has broken the feature.

I put some logs to try to track what is happening, it seems to be because `WinitWindowDelegate` isn't called on `windowWillEnterFullscreen`, whereas `windowWillExitFullScreen` is called.

So this state is updated in `windowWillExitFullScreen` handler [full code here](https://github.com/rust-windowing/winit/blob/462bb4d3245d39f10dc3e0706cd6f892b8cd9e74/src/platform_impl/macos/window_delegate.rs#L242) :
```rust
shared_state.in_fullscreen_transition = true;
```

Then, `macos::UnownedWindow::set_fullscreen()` will return here ([full code here](https://github.com/rust-windowing/winit/blob/462bb4d3245d39f10dc3e0706cd6f892b8cd9e74/src/platform_impl/macos/window.rs#L775))
```rust
        if shared_state_lock.in_fullscreen_transition {
            // We can't set fullscreen here.
            // Set fullscreen after transition.
            shared_state_lock.target_fullscreen = Some(fullscreen);
            return;
        }
```

## Infos

* OS version: 12.6 (x64)
* SHA-1 used: 462bb4d3245d39f10dc3e0706cd6f892b8cd9e74

madsmtm:
Good catch! I've put up a fix here: https://github.com/rust-windowing/winit/pull/2546

rofferom:
Nice, thanks !

I took my some time to understand the patch :sweat_smile:

PR #2546, opened 2022-11-02
macOS: Fix overridden fullscreen selectors

madsmtm:
Fixes https://github.com/rust-windowing/winit/issues/2520

These were incorrectly named (`Fullscreen` instead of `FullScreen`, notice the big "S").

  - Introduced in https://github.com/rust-windowing/winit/pull/2458 which has not been released yet, so no need to note this in the changelog.

### DIFF
--- a/src/platform_impl/macos/window_delegate.rs
+++ b/src/platform_impl/macos/window_delegate.rs
@@ -221,9 +221,9 @@ declare_class!(
         }
 
         /// Invoked when before enter fullscreen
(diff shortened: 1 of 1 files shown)


rofferom:
Works on my machine :+1:

madsmtm:
For reference: I'm working on things in [`objc2`](https://github.com/madsmtm/objc2) that could help catching mistakes like these in the future

PR #2547, opened 2022-11-03
Fix panic in current_monitor_inner

xiaopengli89:

```
panic:  uncaught panic: "expected screen" at .cargo/git/checkouts/winit-560a7c8630e332ba/a8d2a7e/src/platform_impl/macos/window.rs:1071:40
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - **Breaking:** Removed `WindowBuilderExtWindows::with_theme` and `WindowBuilderExtWayland::with_wayland_csd_theme` in favour of `WindowBuilder::with_theme`.
 - **Breaking:** Removed `WindowExtWindows::theme` in favour of `Window::theme`.
 - Enabled `doc_auto_cfg` when generating docs on docs.rs for feature labels.
-- macOS: Return `None` instead of panic when getting monitor fails
+- On macOS, fix panic when getting current monitor without any monitor attached.
 
 # 0.27.5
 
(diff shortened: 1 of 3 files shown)


kchibisov (approved):

kchibisov (changes requested):
Would need a changelog entry for that change.

madsmtm (approved):
Thanks!

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - **Breaking:** Removed `WindowBuilderExtWindows::with_theme` and `WindowBuilderExtWayland::with_wayland_csd_theme` in favour of `WindowBuilder::with_theme`.
 - **Breaking:** Removed `WindowExtWindows::theme` in favour of `Window::theme`.
 - Enabled `doc_auto_cfg` when generating docs on docs.rs for feature labels.
+- macOS: Return `None` instead of panic when getting monitor fails
```
kchibisov, line 31:
```suggestion
- On macOS, fix panic when getting current monitor without any monitor attached.
```

kchibisov: I think that's what exactly happening?

src/platform_impl/macos/window.rs
```diff
@@ -804,7 +804,9 @@ impl WinitWindow {
         if let Some(ref fullscreen) = fullscreen {
             let new_screen = match fullscreen {
                 Fullscreen::Borderless(Some(monitor)) => monitor.clone(),
-                Fullscreen::Borderless(None) => self.current_monitor_inner(),
+                Fullscreen::Borderless(None) => {
+                    self.current_monitor_inner().expect("expected screen")
```
kchibisov, line 808: Wait, why do we crash here? Could we just return and no-op?

xiaopengli89: `set_fullscreen` has no return value, the user always assumes it will succeed. I'm not sure if no-op here wil cause a state error.

kchibisov: It won't compositor/system could ignore full-screen request. That's sort of expected. You also don't really know whether it succeeded or not until you query.

xiaopengli89: Ok

kchibisov (approved):

PR #2548, opened 2022-11-04
Add `WindowBuilder::with_parent_window`

wusyong:
Add an extension method similar to `WindowBuilderExtWindows::with_parent_window`.
I'm not too familiar with objc2. Let me know if it's not idiomatic enough.

### DIFF
--- a/examples/child_window.rs
+++ b/examples/child_window.rs
@@ -56,7 +56,7 @@ fn main() {
                     *control_flow = ControlFlow::Exit;
                 }
                 WindowEvent::CursorEntered { device_id: _ } => {
-                    // println when the cursor entered in a window even if the child window is created
+                    // On x11, println when the cursor entered in a window even if the child window is created
                     // by some key inputs.
                     // the child windows are always placed at (0, 0) with size (200, 200) in the parent window,
                     // so we also can see this log when we move the cursor arround (200, 200) in parent window.
--- a/examples/child_window.rs
+++ b/examples/child_window.rs
@@ -6,7 +6,7 @@
 fn main() {
     use std::collections::HashMap;
 
-    use raw_window_handle::{HasRawWindowHandle, RawWindowHandle};
+    use raw_window_handle::HasRawWindowHandle;
     use winit::{
         dpi::{LogicalPosition, LogicalSize, Position},
         event::{ElementState, Event, KeyboardInput, WindowEvent},
@@ -15,10 +15,11 @@ fn main() {
     };
 
     fn spawn_child_window(
-        parent: RawWindowHandle,
+        parent: &Window,
         event_loop: &EventLoopWindowTarget<()>,
         windows: &mut HashMap<WindowId, Window>,
     ) {
+        let parent = parent.raw_window_handle();
         let mut builder = WindowBuilder::new()
             .with_title("child window")
             .with_inner_size(LogicalSize::new(200.0f32, 200.0f32))
@@ -43,7 +44,7 @@ fn main() {
         .build(&event_loop)
         .unwrap();
 
-    let root = parent_window.raw_window_handle();
+    let root = parent_window;
     println!("parent window: {:?})", root);
 
     event_loop.run(move |event: Event<'_, ()>, event_loop, control_flow| {
@@ -70,7 +71,7 @@ fn main() {
                         },
                     ..
                 } => {
-                    spawn_child_window(root, event_loop, &mut windows);
+                    spawn_child_window(&root, event_loop, &mut windows);
                 }
                 _ => (),
             }
--- a/examples/child_window.rs
+++ b/examples/child_window.rs
@@ -1,37 +1,38 @@
-#[cfg(all(target_os = "linux", feature = "x11"))]
-use std::collections::HashMap;
-
-#[cfg(all(target_os = "linux", feature = "x11"))]
-use winit::{
-    dpi::{LogicalPosition, LogicalSize, Position},
-    event::{ElementState, Event, KeyboardInput, WindowEvent},
-    event_loop::{ControlFlow, EventLoop, EventLoopWindowTarget},
-    platform::x11::{WindowBuilderExtX11, WindowExtX11},
-    window::{Window, WindowBuilder, WindowId},
-};
-
-#[cfg(all(target_os = "linux", feature = "x11"))]
-fn spawn_child_window(
-    parent: u32,
-    event_loop: &EventLoopWindowTarget<()>,
-    windows: &mut HashMap<u32, Window>,
-) {
-    let child_window = WindowBuilder::new()
-        .with_parent(WindowId::from(parent as u64))
-        .with_title("child window")
-        .with_inner_size(LogicalSize::new(200.0f32, 200.0f32))
-        .with_position(Position::Logical(LogicalPosition::new(0.0, 0.0)))
-        .with_visible(true)
-        .build(event_loop)
-        .unwrap();
-
-    let id = child_window.xlib_window().unwrap() as u32;
-    windows.insert(id, child_window);
-    println!("child window created with id: {}", id);
-}
-
-#[cfg(all(target_os = "linux", feature = "x11"))]
+#[cfg(any(
+    all(target_os = "linux", feature = "x11"),
+    target_os = "macos",
+    target_os = "windows"
+))]
 fn main() {
+    use std::collections::HashMap;
+
+    use raw_window_handle::{HasRawWindowHandle, RawWindowHandle};
+    use winit::{
+        dpi::{LogicalPosition, LogicalSize, Position},
+        event::{ElementState, Event, KeyboardInput, WindowEvent},
+        event_loop::{ControlFlow, EventLoop, EventLoopWindowTarget},
+        window::{Window, WindowBuilder, WindowId},
+    };
+
+    fn spawn_child_window(
+        parent: RawWindowHandle,
+        event_loop: &EventLoopWindowTarget<()>,
+        windows: &mut HashMap<WindowId, Window>,
+    ) {
+        let child_window = WindowBuilder::new()
+            .with_parent_window(Some(parent))
+            .with_title("child window")
+            .with_inner_size(LogicalSize::new(200.0f32, 200.0f32))
+            .with_position(Position::Logical(LogicalPosition::new(0.0, 0.0)))
+            .with_visible(true)
+            .build(event_loop)
+            .unwrap();
+
+        let id = child_window.id();
+        windows.insert(id, child_window);
+        println!("child window created with id: {:?}", id);
+    }
+
     let mut windows = HashMap::new();
 
     let event_loop: EventLoop<()> = EventLoop::new();
@@ -42,8 +43,8 @@ fn main() {
         .build(&event_loop)
         .unwrap();
 
-    let root = parent_window.xlib_window().unwrap() as u32;
-    println!("parent window id: {})", root);
+    let root = parent_window.raw_window_handle();
+    println!("parent window: {:?})", root);
 
     event_loop.run(move |event: Event<'_, ()>, event_loop, control_flow| {
         *control_flow = ControlFlow::Wait;
@@ -77,7 +78,11 @@ fn main() {
     })
 }
 
-#[cfg(not(all(target_os = "linux", feature = "x11")))]
+#[cfg(not(any(
+    all(target_os = "linux", feature = "x11"),
+    target_os = "macos",
+    target_os = "windows"
+)))]
 fn main() {
-    panic!("This example is supported only on x11.");
+    panic!("This example is supported only on x11, macOS, and Windows.");
 }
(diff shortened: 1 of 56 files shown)


src/platform/macos.rs
```diff
@@ -113,6 +114,8 @@ pub trait WindowBuilderExtMacOS {
     fn with_has_shadow(self, has_shadow: bool) -> WindowBuilder;
     /// Window accepts click-through mouse events.
     fn with_accepts_first_mouse(self, accepts_first_mouse: bool) -> WindowBuilder;
+    /// Sets a parent to the window to be created.
+    fn with_parent_window(self, parent: *mut c_void) -> WindowBuilder;
```
madsmtm, line 118: This is an `unsafe` operation (since we're taking a raw pointer), and it should be marked as such. Would also be nice with a `# Safety` section detailing exactly which guarantees the user must uphold

src/platform_impl/macos/window.rs
```diff
@@ -79,6 +89,7 @@ pub struct PlatformSpecificWindowBuilderAttributes {
     pub disallow_hidpi: bool,
     pub has_shadow: bool,
     pub accepts_first_mouse: bool,
+    pub parent: Parent,
```
madsmtm, line 92: I think I'd rather use `Option<Id<NSWindow, Shared>>` here, and then create the `Id<NSWindow, Shared>` directly inside `WindowBuilderExtMacOS::with_parent_window` by retaining the pointer with `Id::retain` (that way we also know that the window will be valid for the duration we need it to be)

src/platform_impl/macos/window.rs
```diff
@@ -344,6 +356,13 @@ impl WinitWindow {
                     }
                 }
 
+                if let Parent::ChildOf(parent) = pl_attrs.parent {
+                    #[allow(clippy::let_unit_value)]
+                    unsafe {
+                        let () = msg_send![parent as *mut Object, addChildWindow: &*this, ordered: NSWindowOrderingMode::NSWindowAbove];
```
madsmtm, line 362: This could be added as an `unsafe` method on `NSWindow` instead, which would give me more confidence in the safety than this inline usage does

src/platform_impl/macos/window.rs
```diff
@@ -344,6 +356,13 @@ impl WinitWindow {
...
+                    }
+                }
+
```
madsmtm, line 365: Is there a specific reason you placed this code here instead of further down? As-in, does the `addChildWindow:ordered:` call need to happen at a certain point, or can it be done basically anytime?

madsmtm (changes requested):
A few nits on the `objc2` usage, and some change to the `unsafe` stuff.

May I ask, for your use-case, do you only ever need the parent window to be a window already managed by `winit`? Because if so, we could create a safe, much cleaner API with just `fn with_parent_window(self, parent: &Window) -> WindowBuilder`.

src/platform_impl/macos/window.rs
wusyong: It can be further down but I saw every platform specific attributes are handled here.
Where do you think is better?

src/platform_impl/macos/window.rs
wusyong: I can do this but `NSWindow` defined in `appkit` module isn't public.
Can I just re-export them to platform module?

src/platform_impl/macos/window.rs
madsmtm: No then it's fine, just wanted to know ;)

src/platform_impl/macos/window.rs
madsmtm: I think you can just do `pub(crate) parent: Option<Id<NSWindow, Shared>>`?

src/platform/macos.rs
madsmtm: No, I think it's correct that we take a `*mut c_void` pointer (unless your use-case allows it to just be `&Window`, see https://github.com/rust-windowing/winit/pull/2548#pullrequestreview-1189662991).

wusyong: I see. My use-cases needs to cover all NSWindow from others rather than just winit's window I think.
I'll make this `unsafe` and add `#Safety` doc.

kchibisov (changes requested):
It seems like we're basically adding the `with_parent` that we have on X11 already? Maybe we should unify the interfaces and use the `RawWindowHandle` as the argument to them?

I'm against having 2 calls doing the same things on multiple platforms fwiw.

wusyong:
@kchibisov Windows also has [`with_parent_window](https://docs.rs/winit/latest/x86_64-pc-windows-msvc/winit/window/struct.WindowBuilder.html#method.with_parent_window)`. I can unify them if you agree.

kchibisov:
Yes, unifying and using raw-window-handle for that is what I want. Though I'm not sure if we want raw-window-handle or take &Window in builder.

src/platform/windows.rs
```diff
@@ -235,15 +228,9 @@ pub trait WindowBuilderExtWindows {
...
-    #[inline]
-    fn with_parent_window(mut self, parent: HWND) -> WindowBuilder {
-        self.platform_specific.parent = Parent::ChildOf(parent);
-        self
-    }
-
     #[inline]
     fn with_owner_window(mut self, parent: HWND) -> WindowBuilder {
-        self.platform_specific.parent = Parent::OwnedBy(parent);
+        self.platform_specific.parent = Some(parent);
```
msiglreith, line 233: would like to have it renamed to `self.platform_specific.owner` as otherwise we have shared attribute `parent` and platform specific `parent` attribute

src/platform_impl/windows/window.rs
```diff
@@ -1008,22 +1008,24 @@ where
...
-        Parent::OwnedBy(parent) => {
-            window_flags.set(WindowFlags::POPUP, true);
-            Some(parent)
-        }
-        Parent::None => {
-            window_flags.set(WindowFlags::ON_TASKBAR, true);
-            None
+            Some(handle.hwnd as HWND)
         }
+        _ => match pl_attribs.parent {
```
msiglreith, line 1019: `None` instead of `_`

msiglreith (changes requested):

src/window.rs
```diff
@@ -386,6 +388,21 @@ impl WindowBuilder {
...
+    ///
+    /// The default is `None`.
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **Windows** : A child window has the WS_CHILD style and is confined to the client area of its parent window.
+    /// For more information, see <https://docs.microsoft.com/en-us/windows/win32/winmsg/window-features#child-windows>
+    /// - **Android / iOS / Wayland:** Ignored.
+    #[inline]
+    pub fn with_parent_window(mut self, parent_window: Option<RawWindowHandle>) -> Self {
```
maroider, line 401: Since we're taking a `RawWindowHandle` directly here, shouldn't this technically be marked `unsafe`?

kchibisov: Yes, this is true. I was thinking about having `&impl AsRawWindowHandle` as argument, but the issue is that user may want to send window to other thread and then create a new one from the main thread with the parent that you've already sent or your window might be not present.

I'd suggest to mark it `unsafe` then, and say that raw-window-handle must point to the valid window. Enforcing life time is not required here I guess?

kchibisov (changes requested):

src/platform_impl/windows/window.rs
wusyong: There might be other `RawWindowHandle` variant, so I use `_` here.
Do you want me to add more arms like `Some(_)`?

examples/child_window.rs
```diff
@@ -1,37 +1,41 @@
...
+        let child_window = unsafe {
+            WindowBuilder::new()
+                .with_parent_window(Some(parent))
+                .with_title("child window")
+                .with_inner_size(LogicalSize::new(200.0f32, 200.0f32))
+                .with_position(Position::Logical(LogicalPosition::new(0.0, 0.0)))
+                .with_visible(true)
+                .build(event_loop)
+                .unwrap()
+        };
```
madsmtm, line 32: Nit: Shrink `unsafe`

```suggestion
        let mut builder = WindowBuilder::new()
            .with_title("child window")
            .with_inner_size(LogicalSize::new(200.0f32, 200.0f32))
            .with_position(Position::Logical(LogicalPosition::new(0.0, 0.0)))
            .with_visible(true);
        builder = unsafe { builder.with_parent_window(Some(parent)) };
        let child_window = builder.build(event_loop).unwrap();
```

madsmtm (changes requested):

src/platform_impl/windows/window.rs
```diff
@@ -1008,22 +1008,24 @@ where
...
-        Parent::OwnedBy(parent) => {
-            window_flags.set(WindowFlags::POPUP, true);
-            Some(parent)
-        }
-        Parent::None => {
-            window_flags.set(WindowFlags::ON_TASKBAR, true);
-            None
+            Some(handle.hwnd as HWND)
         }
+        _ => match pl_attribs.owner {
```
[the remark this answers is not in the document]
wusyong: Added document in ` WindowBuilderExtWindows::with_owner_window`

src/platform_impl/linux/x11/window.rs
```diff
@@ -121,9 +121,8 @@ impl UnownedWindow {
         pl_attribs: PlatformSpecificWindowBuilderAttributes,
     ) -> Result<UnownedWindow, RootOsError> {
         let xconn = &event_loop.xconn;
-        let root = if let Some(id) = pl_attribs.parent_id {
-            // WindowId is XID under the hood which doesn't exceed u32, so this conversion is lossless
-            u64::from(id) as _
+        let root = if let Some(RawWindowHandle::Xlib(handle)) = window_attrs.parent_window {
+            handle.window
```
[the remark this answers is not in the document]
wusyong: `WindowBuilderExtX11::with_parent` takes `WindowId`. I assume it only handles Xlib?
I only saw Xcb in raw_window_handle_04 usage.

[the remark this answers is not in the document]
maroider: It would probably be fine to accept both. The `window` and `visual_id` are both the same, except Xlib has (likely for historical reasons) these values typed as `c_ulong`, even though I'm fairly certain both values are guaranteed to be 32-bit at the protocol level.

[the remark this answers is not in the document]
wusyong: Alright! I update with a Xcb arm to match.

madsmtm (approved):
Looks good from my side, I also tested it on macOS and i works flawlessly

src/platform_impl/macos/window.rs
```diff
@@ -357,6 +358,29 @@ impl WinitWindow {
...
+                        // SAFETY: Caller ensures the pointer is valid or NULL
+                        let parent: Id<NSWindow, Shared> =
+                            unsafe { Id::retain(handle.ns_window.cast()) }.unwrap_or_else(|| {
+                                // SAFETY: Caller ensures the pointer is valid or NULL
+                                let parent_view: Id<NSView, Shared> =
+                                    unsafe { Id::retain(handle.ns_view.cast()) }
+                                        .expect("raw window handle should be non-empty");
+                                parent_view
+                                    .window()
+                                    .expect("parent view should be installed in a window")
```
kchibisov, line 372: Could you propagate errors instead of doing `except`? You have `os_error` and such for that.

kchibisov (changes requested):
Looks mostly good.

src/platform_impl/macos/window.rs
```diff
@@ -369,6 +370,38 @@ impl WinitWindow {
...
+                                    "parent view should be installed in a window"
+                                ))
+                            })?
+                        }
+                    };
+                // SAFETY: We know that there are no parent -> child -> parent cycles since the only place in `winit`
+                // where we allow making a window a child window is right here, just after it's been created.
+                unsafe { parent.addChildWindow(&this, NSWindowOrderingMode::NSWindowAbove) };
+            }
+            Some(raw) => unreachable!("Invalid raw window handle {raw:?} on macOS"),
```
kchibisov, line 401: Could you use `panic!` here? Other than that should be good to go.

src/platform_impl/macos/window.rs
```diff
@@ -369,6 +370,38 @@ impl WinitWindow {
...
+            None => {}
```
kchibisov, line 402:
```suggestion
            None => (),
```

kchibisov: Unfortunately, we're using stable rustfmt for now and it doesn't reformat such things...

kchibisov (changes requested):

src/window.rs
```diff
@@ -401,6 +403,26 @@ impl WindowBuilder {
...
+    /// The default is `None`.
+    ///
+    /// ## Safety
+    ///
+    /// `parent_window` must be a valid window handle.
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **Windows** : A child window has the WS_CHILD style and is confined
+    /// to the client area of its parent window. For more information, see
```
dhardy, line 417: On X11 the child is also confined to the parent window's area (at least when I last tested).

madsmtm:
Cool, thanks @wusyong!

PR #2550, opened 2022-11-07
On macOS, add documentEdited APIs

amrbashir:

Port of https://github.com/tauri-apps/tao/commit/33fdeab6291d4aef8ea9facb58fe9583f6c1aaf3

### DIFF
--- a/src/platform/macos.rs
+++ b/src/platform/macos.rs
@@ -54,7 +54,7 @@ pub trait WindowExtMacOS {
     fn is_document_edited(&self) -> bool;
 
     /// Put the window in a state which indicates a file save is required.
-    fn set_decoument_edited(&self, edited: bool);
+    fn set_document_edited(&self, edited: bool);
 }
 
 impl WindowExtMacOS for Window {
@@ -94,8 +94,8 @@ impl WindowExtMacOS for Window {
     }
 
     #[inline]
-    fn set_decoument_edited(&self, edited: bool) {
-        self.window.set_decoument_edited(edited)
+    fn set_document_edited(&self, edited: bool) {
+        self.window.set_document_edited(edited)
     }
 }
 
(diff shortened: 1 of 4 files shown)


kchibisov:
How does it work? Is it basically enable/disable prompt and there's no event attached to it that says that user is trying to close the window with edit? Do you get `CloseRequested` in that case?

Is that to prevent force close by the system, since winit app can ignore `CloseRequested` and emulate the dialog thingy?

amrbashir:
I don't use macOS but I think it is just a hint to the user that this window has unsaved changes, see this image 
image

src/platform/macos.rs
```diff
@@ -36,6 +36,12 @@ pub trait WindowExtMacOS {
 
     /// Sets whether or not the window has shadow.
     fn set_has_shadow(&self, has_shadow: bool);
+
+    /// Get the window's edit state.
+    fn is_document_edited(&self) -> bool;
+
+    /// Put the window in a state which indicates a file save is required.
+    fn set_decoument_edited(&self, edited: bool);
```
madsmtm, line 44: Typo
```suggestion
    fn set_document_edited(&self, edited: bool);
```

src/platform/macos.rs
```diff
@@ -36,6 +36,12 @@ pub trait WindowExtMacOS {
 
     /// Sets whether or not the window has shadow.
     fn set_has_shadow(&self, has_shadow: bool);
+
+    /// Get the window's edit state.
```
madsmtm, line 40: Would be nice to document how this relates to `CloseRequested`, something like:
```suggestion
    /// Get the window's edit state.
    ///
    /// # Examples
    ///
    /// ```ignore
    /// WindowEvent::CloseRequested => {
    ///     if window.is_document_edited() {
    ///         // Show the user a save pop-up or similar
    ///     } else {
    ///         // Close the window
    ///         drop(window);
    ///     }
    /// }
    /// ```
```

madsmtm (changes requested):
Tested it, it has no effect on `CloseRequested`, is simply a visual indication (though that would be nice to document!). It looks as follows on macOS 10.14.6:

Normal:
unedited

Edited:
edited

madsmtm (approved):
Thanks, will merge in a few hours once CI catches up

PR #2551, opened 2022-11-07
On macOS,  add `EventLoopBuilderExtMacOS::with_activate_ignoring_other_apps`

amrbashir:

Port of https://github.com/tauri-apps/tao/commit/d2c6a91c4588bab460ef5924b10cfb224c1336c6


madsmtm (changes requested):
Yes please!

I've actually been considering making the default `false` with `debug_assertions` or something similar, since there is actually very little reason to call `activateIgnoringOtherApps(true)` (and it can be very annoying, looking at you Steam!), _except_ when developing it's something that developers have come to somewhat expect from Xcode.

amrbashir:
> I've actually been considering making the default `false` with `debug_assertions`

imo, this should probably be done in user code.

madsmtm (approved):
> imo, this should probably be done in user code.

Yeah probably.

Anyway, thanks again, if you rebase the CI should be fixed

PR #2552, opened 2022-11-07
On macOS, add tabbing APIs

amrbashir:

Port of https://github.com/tauri-apps/tao/commit/8815291e8cf4ac855ae4b9ad93183a27d6da5bb7 and https://github.com/tauri-apps/tao/commit/7c7ce8ab2d838a79ecdf83df00124c418a6a51f6


kchibisov:
Out of curiosity how tauri handles tabs on windows?

amrbashir:
Not sure what you mean, but Windows doesn't have tabs like macOS, each Window is displayed in its own frame.

kchibisov:
I was curious what tauri does on windows for tabs? Do they have some builtin tabs and on macOS it's using native tabs or how it's handled?

amrbashir:
No, we don't handle tabs at all on other platforms. This is purely macOS case, the default behavior is to group related windows into tabs and some users dislike that behavior and wanted an API to change it.

kchibisov:
@amrbashir do you still want this in winit or I could pick up this PR?

amrbashir:
yeah sure, let me fix the conflicts for you

kchibisov:
@amrbashir could we have a way to focus a specific tab and cycle through the tabs? So the default macOS bindings could be supported by winit downstream applications? I'd assume you'd need something like that in tauri regardless given that users likely want to cycle through tabs or use `cmd + [1-9]` hotkey to focus a specific tab?

It would also be nice to have and example for macOS tabbing, so you could open a window and do general hotkeys related stuff (cmd + tab to open a new tab, cmd + ] and cmd + [ to cycle and numbers to focus exact tab).

amrbashir:
The reason why we added these options in Tauri was because the default behavior in macOS is to group windows into tabs and some users disliked that behavior and just needed an API to disable that.

Unfortunately I rarely develop for macOS, and I don't have access to one at all so I can't help with that task. The official documentation on apple site is also pretty lacking but I found this article https://christiantietze.de/posts/2019/07/nswindow-tabbing-multiple-nswindowcontroller/ which can be helpful.

kchibisov:
I'll try to implement that based on your PR, I'm sort of aware what is required given that it was already done in the past by some users.

PR #2553, opened 2022-11-08
Add `Window::set_theme`

amrbashir:
	- [X] Windows
	- [x] macOS 
	- [ ] Wayland 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 - On Windows, macOS, X11 and Wayland, add `Window::set_theme`.
 - **Breaking:** Remove `WindowExtWayland::wayland_set_csd_theme` and `WindowBuilderExtX11::with_gtk_theme_variant`.
+- On Windows, revert window background to an empty brush to avoid white flashes when changing scaling.
 - **Breaking:** Removed `Window::set_always_on_top` and related APIs in favor of `Window::set_window_level`.
 - On Windows, MacOS and X11, add always on bottom APIs.
 - On Windows, fix the value in `MouseButton::Other`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
-- On Windows, macOS , X11 and Wayland, add `Window::set_theme`.
+- On Windows, macOS, X11 and Wayland, add `Window::set_theme`.
 - **Breaking:** Remove `WindowExtWayland::wayland_set_csd_theme` and `WindowBuilderExtX11::with_gtk_theme_variant`.
 - On Windows, fix icons specified on `WindowBuilder` not taking effect for windows created after the first one.
 - On Windows and macOS, add `Window::title` to query the current window title.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- On Windows, fix icons specified on `WindowBuilder` not taking effect for windows created after the first one.
 - On Windows and macOS, add `Window::title` to query the current window title.
 - On Windows, fix focusing menubar when pressing `Alt`.
 - On MacOS, made `accepts_first_mouse` configurable.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,13 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 - On Windows, macOS, X11 and Wayland, add `Window::set_theme`.
 - **Breaking:** Remove `WindowExtWayland::wayland_set_csd_theme` and `WindowBuilderExtX11::with_gtk_theme_variant`.
+- **Breaking:** Removed `Window::set_always_on_top` and related APIs in favor of `Window::set_window_level`.
+- On Windows, MacOS and X11, add always on bottom APIs.
+- On Windows, fix the value in `MouseButton::Other`.
+- On macOS, add `WindowExtMacOS::is_document_edited` and `WindowExtMacOS::set_document_edited` APIs.
+- **Breaking:** Removed `WindowBuilderExtIOS::with_root_view_class`; instead, you should use `[[view layer] addSublayer: ...]` to add an instance of the desired layer class (e.g. `CAEAGLLayer` or `CAMetalLayer`). See `vulkano-win` or `wgpu` for examples of this.
+- On MacOS and Windows, add `Window::set_content_protected`.
+- On MacOS, add `EventLoopBuilderExtMacOS::with_activate_ignoring_other_apps`.
 - On Windows, fix icons specified on `WindowBuilder` not taking effect for windows created after the first one.
 - On Windows and macOS, add `Window::title` to query the current window title.
 - On Windows, fix focusing menubar when pressing `Alt`.
@@ -33,6 +40,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - **Breaking:** Removed `WindowExtWindows::theme` in favour of `Window::theme`.
 - Enabled `doc_auto_cfg` when generating docs on docs.rs for feature labels.
 - On macOS, fix panic when getting current monitor without any monitor attached.
+- **Breaking:** On Android, switched to using [`android-activity`](https://github.com/rib/android-activity) crate as a glue layer instead of [`ndk-glue](https://github.com/rust-windowing/android-ndk-rs/tree/master/ndk-glue). See [README.md#Android](https://github.com/rust-windowing/winit#Android) for more details. ([#2444](https://github.com/rust-windowing/winit/pull/2444))
 
 # 0.27.5
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,8 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
-- On Windows, macOS and Wayland, add `Window::set_theme`.
+- On Windows, macOS , X11 and Wayland, add `Window::set_theme`.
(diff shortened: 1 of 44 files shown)


src/platform_impl/linux/wayland/window/mod.rs
```diff
@@ -620,6 +620,11 @@ impl Window {
         self.event_loop_awakener.ping();
     }
 
+    #[inline]
+    pub fn set_theme(&self, theme: Theme) {
```
kchibisov, line 624: Could you remove `set_csd_theme` and related APIs? Also add a breaking change that it was removed in favor of unified `set_theme` API?

I think other backends might have similar APIs in place(X11's gtk thing is special though), might worth unifying as well.

src/platform_impl/linux/wayland/window/shim.rs
```diff
@@ -537,6 +540,10 @@ pub fn handle_window_requests(winit_state: &mut WinitState) {
                     let event_sink = &mut winit_state.event_sink;
                     event_sink.push_window_event(WindowEvent::Destroyed, *window_id);
                 }
+                WindowRequest::Theme(_theme) => {
+                    #[cfg(feature = "sctk-adwaita")]
+                    window_handle.window.set_frame_config(_theme.into());
```
kchibisov, line 545: Same about CSD here.

src/window.rs
```diff
@@ -943,12 +943,22 @@ impl Window {
         self.window.request_user_attention(request_type)
     }
 
-    /// Returns the current window theme.
+    /// Sets the current window theme.
```
kchibisov, line 946: Could you also port a Wayland section from `set_csd_theme`? Since on Wayland the theme you have is from client side decorations.

kchibisov (changes requested):

src/platform_impl/linux/wayland/window/mod.rs
amrbashir: I took a stab at it but I was not sure if mapping `Theme::Dark` to `dark` and `Theme::Light` to `light` is correct, so let me know if they are wrong.

src/window.rs
```diff
@@ -947,6 +947,9 @@ impl Window {
     ///
     /// ## Platform-specific
     ///
+    /// - **Wayland:** You can also use `WINIT_WAYLAND_CSD_THEME` env variable to set the theme.
+    /// Possible values for env variable are: "dark" and light". When unspecified, a theme is automatically selected.
+    ///
     /// - **iOS / Android / Web / x11:** Unsupported.
```
kchibisov, line 953: Should update x11 docs here as well. Probably if you've decided to touch gtk theme you should also update builder.

amrbashir:
CI is failing because of the dev-dependency `image` requiring a 1.61 MSRV

kchibisov:
>CI is failing because of the dev-dependency image requiring a 1.61 MSRV

Could you pin it for now to a version that is known to work with 1.60 version?

msiglreith:
Given `with_theme` and `theme` support the `None` as system default theme, `set_theme` should support is as well imo

amrbashir:
It was possible on most platforms but:
- x11, I set `_GTK_THEME_VARIANT` to empty string, not sure if that works
- macOS, used the same variant as `Theme::light`, I couldn't find a reliable way to get the system default theme and use it.

src/platform_impl/linux/x11/window.rs
```diff
@@ -564,9 +564,14 @@ impl UnownedWindow {
...
     }
 
-    fn set_gtk_theme_variant(&self, variant: String) -> util::Flusher<'_> {
+    pub fn set_theme_inner(&self, theme: Option<Theme>) -> util::Flusher<'_> {
         let hint_atom = unsafe { self.xconn.get_atom_unchecked(b"_GTK_THEME_VARIANT\0") };
         let utf8_atom = unsafe { self.xconn.get_atom_unchecked(b"UTF8_STRING\0") };
+        let variant = match theme {
+            Some(Theme::Dark) => "dark",
+            Some(Theme::Light) => "light",
+            None => "",
```
kchibisov, line 573: Just set `dark` for `None` I guess? Most desktop environments use dark style by default. The way to get system preference is via side channels which I'd want to see in winit.

kchibisov (changes requested):
Just a minor note, should be good afterwards.

src/platform_impl/linux/x11/window.rs
amrbashir: done, x11 and macOS will now fallback to `Theme::Dark`

src/window.rs
```diff
@@ -943,12 +944,26 @@ impl Window {
...
 
-    /// Returns the current window theme.
+    /// Sets the current window theme. Use `None` to fallback to system default.
     ///
     /// ## Platform-specific
     ///
+    /// - **Wayland:** You can also use `WINIT_WAYLAND_CSD_THEME` env variable to set the theme.
+    /// Possible values for env variable are: "dark" and light". When unspecified, a theme is automatically selected.
+    /// -**x11:** Sets `_GTK_THEME_VARIANT` hint to `dark` or `light`.
+    /// -**x11 / macOS**: `None` will default to  [`Theme::Dark`].
```
madsmtm, line 954: I would expect that `None` on macOS just makes it use the default behaviour (which would be to change theme automatically when the system does)?

madsmtm: Should probably also note that on macOS this is an application global, not actually something to set on the window

madsmtm (changes requested):
Perhaps it would make sense to add `Theme::Auto`, and then just use `Theme` instead of `Option<Theme>`?
Though I guess that depends on what kind of information you desire from `Window::theme` (is it the currently active theme, or the setting)?

src/platform_impl/macos/window.rs
```diff
@@ -1235,13 +1240,13 @@ pub(super) fn get_ns_theme() -> Theme {
...
     let has_theme: bool = unsafe { msg_send![&app, respondsToSelector: sel!(effectiveAppearance)] };
     if has_theme {
         let name = match theme {
-            Theme::Dark => NSString::from_str("NSAppearanceNameDarkAqua"),
-            Theme::Light => NSString::from_str("NSAppearanceNameAqua"),
+            Some(Theme::Dark) | None => NSString::from_str("NSAppearanceNameDarkAqua"),
+            Some(Theme::Light) => NSString::from_str("NSAppearanceNameAqua"),
         };
         let appearance = NSAppearance::appearanceNamed(&name);
         app.setAppearance(&appearance);
```
madsmtm, line 1252: `NSApplication::setAppearance` is wrongly mapped, it should be (note the `Option`):
```rust
#[sel(setAppearance:)]
pub fn setAppearance(&self, appearance: Option<&NSAppearance>);
```

Which would allow you to reset back to the default of not having overridden the theme.

src/window.rs
amrbashir: > Perhaps it would make sense to add `Theme::Auto`, and then just use `Theme` instead of `Option<Theme>`?
> Though I guess that depends on what kind of information one desires from `Window::theme` (is it the currently active theme, or the setting)?

I'd imagine the current active theme.

> I would expect that `None` on macOS just makes it use the default behaviour (which would be to change theme automatically when the system does)?

Yes, but it wasn't clear to me how to do, which you explained below,  thanks.

> Should probably also note that on macOS this is an application global, not actually something to set on the window

Should I remove this method from `Window` for macOS and add it to `EventLoopExtMacOS` instead?

src/platform_impl/macos/window.rs
amrbashir: thanks

src/window.rs
madsmtm: > Should I remove this method from `Window` for macOS and add it to `EventLoopExtMacOS` instead?

No, I think there is value in having a unified API since most users are only going to be having one window anyways. But I think it'd be nice to have documented!

amrbashir: Added the notes. Let me know if you want to me to add `Theme::Auto`, I think it can be acheived for macOS, Windows and Wayland. (not sure about x11, couldn't find anything related to the matter).

madsmtm: Let's just keep using `Option<Theme>`, since it's probably strictly more correct as the return type for `Window::theme`

src/window.rs
```diff
@@ -943,12 +945,28 @@ impl Window {
...
-    /// Returns the current window theme.
+    /// Sets the current window theme. Use `None` to fallback to system default.
     ///
     /// ## Platform-specific
     ///
+    /// - **macOS:** This is an app-wide setting.
+    /// - **Wayland:** You can also use `WINIT_WAYLAND_CSD_THEME` env variable to set the theme.
+    /// Possible values for env variable are: "dark" and light". When unspecified, a theme is automatically selected.
+    /// -**x11:** Sets `_GTK_THEME_VARIANT` hint to `dark` or `light`.
+    /// -**x11**: `None` will default to  [`Theme::Dark`].
```
kchibisov, line 956: Could you merge these two? I don't see a reason to keep them on separate lines.

kchibisov (changes requested):

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- On Windows, macOS , X11 and Wayland, add `Window::set_theme`.
```
```suggestion
- On Windows, macOS, X11 and Wayland, add `Window::set_theme`.
```

msiglreith (approved):

kchibisov (approved):

kchibisov:
>Perhaps it would make sense to add Theme::Auto, and then just use Theme instead of Option<Theme>?
Though I guess that depends on what kind of information one desires from Window::theme (is it the currently active theme, or the setting)?

I think `None` is more common here. We already use `None` as a hint for other apis and it's natural, since you may add `theme` api to get current theme and on some platforms it may result in `None` and when you pass `None` back it'll be persistent.

Though, using `Auto` for theme also makes, sense.

madsmtm (approved):
I went ahead and fixed the result of `Window::theme` on macOS after calling `Window::set_theme`, and improved the `theme` example to showcase the new functionality.

Thanks for your contribution!

PR #2554, opened 2022-11-11
Add smart magnify gesture support for macOS

Jim-Eckerlein:
Add smart magnify gesture support for macOS.

[Apple Developer Documentation](https://developer.apple.com/documentation/appkit/nsresponder/1532984-smartmagnifywithevent)

On a Mac, smart magnification is triggered by a double tap with two fingers and is commonly used to zoom on a certain object (e.g. a paragraph of a PDF) or (sort of like a toggle) to reset any zoom. The gesture is also supported in Safari, Pages, etc.

The event is general enough that its generating gesture is allowed to vary across platforms. It could also be generated by another device.

Unfortunatly, neither [Windows](https://support.microsoft.com/en-us/windows/touch-gestures-for-windows-a9d28305-4818-a5df-4e2b-e5590f850741) nor [Wayland](https://wayland.freedesktop.org/libinput/doc/latest/gestures.html) support this gesture or any other gesture with the same effect.

## Checklist


src/event.rs
```diff
@@ -444,6 +444,17 @@ pub enum WindowEvent<'a> {
...
 
+    /// Smart magnification event.
+    ///
+    /// Generally used to magnify by an automatically calculated factor
+    /// to fit a certain portion of the window currently under the cursor,
+    /// e.g. a paragraph of a PDF, or to reset any magnification.
+    ///
+    /// ## Platform-specific
+    ///
+    /// - Only available on **macOS 10.8** and later.
```
madsmtm, line 455: Maybe state here how to trigger it (like you did in your PR description)

madsmtm (approved):
Implementation looks fine, thanks!

I'm a bit weary adding something so platform specific, but I guess it'll be better to just add it and then maybe remove it again once https://github.com/rust-windowing/winit/issues/2120 finishes.

madsmtm:
Thank you!

PR #2558, opened 2022-11-14
Only build, but don't run tests in MSRV CI

madsmtm:
Since the MSRV of development dependencies can easily be bumped without it affecting the MSRV of the published version of `winit`.

Done because the `image` crate now has an MSRV of 1.61: https://github.com/image-rs/image/commit/11492233cd59fc0e2701ab12cd00607e9bbf125c

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,15 +77,22 @@ jobs:
       shell: bash
       run: cargo $CMD doc --no-deps --target ${{ matrix.platform.target }} $OPTIONS --features $FEATURES --document-private-items
 
+    - name: Build crate
+      shell: bash
+      run: cargo $CMD build --verbose --target ${{ matrix.platform.target }} $OPTIONS --features $FEATURES
+
     - name: Build tests
       shell: bash
+      if: matrix.rust_version != '1.60.0'
       run: cargo $CMD test --no-run --verbose --target ${{ matrix.platform.target }} $OPTIONS --features $FEATURES
+
     - name: Run tests
       shell: bash
       if: (
         !contains(matrix.platform.target, 'android') &&
         !contains(matrix.platform.target, 'ios') &&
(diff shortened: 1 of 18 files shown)


msiglreith:
Is it possible to define th MSRV in an `env` section within the github workflow, so it can be referenced? Having to sync multiple locations on an update doesn't seem good to me

madsmtm:
> Is it possible to define th MSRV in an `env` section within the github workflow, so it can be referenced? Having to sync multiple locations on an update doesn't seem good to me

I found a way without using `env`, though I don't think that's really ideal though.

I'm unsure if `env` works inside the `strategy` key, but I guess I should just try it.

maroider:
We could get the number of explicit mentions of the MSRV down to zero if we'd like, but the way I'd propose to do that is a bit hacky.

Basically, we'd define a step that reads the MSRV from our CARGO.toml, and then makes that available to subsequent steps through `steps.rust-version.outputs.version`.
```yaml
    - id: rust-version
      run: bash -c "echo version=\$(grep rust-version Cargo.toml | grep -P -o \"\d*\.\d*.\d*\")" >> $GITHUB_OUTPUT
```

madsmtm:
Given that we also write our MSRV in READMEs and such elsewhere, I don't think it's worth it. When one wants to do an MSRV bump, all one has to do is grep for the current MSRV

kchibisov:
I don't like the use of `MSRV`. Could we just explicitly use the version? It doesn't save us anything and I like explicit version, since I can `grep` the version and automatically `sed` it.

Besides that the change sounds good to me.

madsmtm:
I agree with @kchibisov that just duplicating the version is fine, given that it's done so all over our documentation as well

madsmtm:
I'll just go with this as it is now, since the failing CI is blocking everything else, feel free to open a new PR if you want to move the MSRV to only be specified in one place

PR #2561, opened 2022-11-21
ci: Specify to build `winit` package when invoking `cargo-apk`

MarijnS95:
Since https://github.com/dvc94ch/cargo-subcommand/pull/23 `cargo-apk` now strictly searches for workspaces first before committing to finding the right package _within said workspace_, and bails when no package was selected since we don't support selecting (building, packaging, running) >1 target currently.

Perhaps it's a bit hash to enforce this on free-form `cargo apk --` invocations, but it is what it is.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,7 +75,7 @@ jobs:
 
     - name: Check documentation
       shell: bash
-      run: cargo $CMD doc --no-deps --target ${{ matrix.platform.target }} $OPTIONS --features $FEATURES --document-private-items
+      run: cargo doc --no-deps --target ${{ matrix.platform.target }} $OPTIONS --features $FEATURES --document-private-items
 
     - name: Build crate
       shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,11 +88,11 @@ jobs:
 
     - name: Run tests
       shell: bash
-      if: (
+      if: >
         !contains(matrix.platform.target, 'android') &&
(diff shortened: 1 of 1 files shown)


MarijnS95:
I'll close this soon, now that original behaviour has been restored _while using `clap`_ in https://github.com/rust-windowing/android-ndk-rs/pull/363 :tada:; `winit` doesn't need to do anything to upgrade to the latest `cargo-apk`.

However, this PR does contain some nice cleanups... :sweat_smile:

madsmtm (approved):
I'm fine with the `options: -p winit` fix

MarijnS95:
Looks like the CI is all green now!

Fwiw it's not strictly a "fix" but more so explicitly specifying what we want `cargo apk` to do; OTOH `cargo` has defined ways in which it selects the package to build when not using `--all`/`--workspace`, but I vaguely recall writing that that logic was vastly complicated and error prone (based on `$PWD`, cmdline flags _and_ default configs in `[workspace]`).

MarijnS95:
Restarting one job, https://github.com/rust-windowing/winit/actions/runs/3533140877/jobs/5928396199 just got stuck for 15 minutes...

MarijnS95:
I keep forgetting that I have merge rights here :grimacing: - merging this in now so that stuff merged to `master` no longer results in commits with failed CI...

MarijnS95:
Fwiw I'm trying to back out of this change:

https://github.com/dvc94ch/cargo-subcommand/pull/25
https://github.com/rust-windowing/android-ndk-rs/issues/371

PR #2564, opened 2022-11-22
On Windows, fix left mouse button release event not being sent after `Window::drag_window`

feelingnothing:
This is a dupe for #2200 because I'm bad at git.
After tricking windows into thinking we are pressing on `HTCAPTION`, we would not get `WM_NCLBUTTONUP`, thus creating unexpected behavior in `winit` dependent projects (https://github.com/emilk/egui/issues/1245#issue-1134534176, https://github.com/bevyengine/bevy/issues/4394, https://github.com/tauri-apps/tauri/issues/3811)
Some debugging after it appears that `wine` have different logic that `winapi` (for context see [this](https://matrix.to/#/!DGpLzJTRzBDTwZiogk:matrix.org/$8hlJG0C6Eiorsj0suXpUr5IOXx0if1nhIfPRHoMwZI8?via=matrix.org&via=libera.chat))

With event tracing we can see that the most suitable event triggered on every drag is `WM_EXITSIZEMOVE`([docs](https://docs.microsoft.com/en-us/windows/win32/winmsg/wm-exitsizemove)):
image

- Fixes: https://github.com/rust-windowing/winit/issues/2192

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- On Windows, fix left mouse button release event not being sent after `Window::drag_window`
 - On Windows, fix icons specified on `WindowBuilder` not taking effect for windows created after the first one.
 - On Windows and macOS, add `Window::title` to query the current window title.
 - On Windows, fix focusing menubar when pressing `Alt`.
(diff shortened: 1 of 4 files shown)


feelingnothing:
Can we get some traction on this PR and issue as a whole. Many projects that utilize this library are under impact from being unable to include custom undecorated window with usable gui.

msiglreith (approved):
Not really happy with emitting some custom messages and tracking these dragging parts internally - fine for me for now. https://github.com/rust-windowing/winit/pull/2538 seems the more proper way to handle it at least on Window in the future

Issue #2563, opened 2022-11-21
Side mouse buttons only recognized on initial input


Satellile:
In Bevy 0.9, I am watching for Mouse5 or Mouse4 to be pressed using MouseButton::Other(32) and MouseButton::Other(64). 

However, it only recognizes the input on the first press, and then proceeds to no longer recognize the input. Left click works fine.

The code I used in Bevy 0.9 to watch for mouse input follows:

    fn watch_for_mouse_input(mouse: Res<Input<MouseButton>>) {
        if mouse.just_pressed(MouseButton::Other(64)) { println!("Detected input 'Mouse5'!"); }
        if mouse.just_pressed(MouseButton::Other(32)) { println!("Detected input 'Mouse4'!"); }
        if mouse.just_pressed(MouseButton::Left)      { println!("Detected input 'Left Click'!"); }
    }

And the output in console (comparing when the input is recognized vs the MouseEventInput stream from Bevy) is attached in this screenshot:
![November 21 1337_WindowsTerminal](https://user-images.githubusercontent.com/118775491/203166520-9ae93d3e-b29b-4409-98b0-e6989ac73778.png)
Notice how after Mouse4 and Mouse5 being recognized once, subsequent presses are no longer recognized.

My apologizes if this is not actually a winit issue; I'm not sure if this should be reported here or in the Bevy github somewhere.

madsmtm:
I suspect this will be fixed by https://github.com/rust-windowing/winit/pull/2565 as well

Issue #2562, opened 2022-11-21
Side mouse buttons are recognized by MouseButton::Other(32) and MouseButton::Other(64) instead of an expected number


Satellile:
I expected the two common side mouse buttons (Mouse4 and Mouse5) to be referenced by MouseButton::Other(1) and MouseButton::Other(2) (or by 4 and 5), but they are currently referenced by MouseButton::Other(32) and MouseButton::Other(64).

I believe this is a relatively recent change (and perhaps unintended?), as I had been referring to them previously using 1 and 2 (in a Bevy 0.8 project).

madsmtm:
Which operating system / display server protocol are you using?

Satellile:
> Which operating system / display server protocol are you using?

I'm on Windows 10 (Education, Version 10.0.19045 Build 19045)

madsmtm:
Good catch, I've opened a PR to fix this in https://github.com/rust-windowing/winit/pull/2565

PR #2565, opened 2022-11-22
Fix value in `MouseButton::Other` on Windows

madsmtm:
Fixes https://github.com/rust-windowing/winit/issues/2562 and probably fixes https://github.com/rust-windowing/winit/issues/2563.

This was a mistake in https://github.com/rust-windowing/winit/pull/2057.

We used `winapi`'s [`GET_XBUTTON_WPARAM`](https://docs.rs/winapi/0.3.9/src/winapi/um/winuser.rs.html#1297-1299) before which is using `HIWORD` instead of `LOWORD`.


madsmtm:
We might want to include this in a patch version to `v0.27`? (@kchibisov)

msiglreith (approved):

PR #2576, opened 2022-11-30
Implemented option_as_alt to allow for ignoring option key characters

ayax79:
Based off the [Fix Alt Modifier on Mac OS PR](https://github.com/rust-windowing/winit/pull/1449).

I have tested by rebuilding alacritty [here](https://github.com/ayax79/alacritty/tree/option_as_alt) and testing to ensure that alt bindings in zellij work correctly. I have also tested that IME works. 


src/platform_impl/macos/view.rs
```diff
@@ -79,6 +80,9 @@ pub(super) struct ViewState {
     /// True if the current key event should be forwarded
     /// to the application, even during IME
     forward_key_to_app: bool,
+
+    raw_characters: Option<String>,
```
kjmph, line 84: Odd, I stripped all this state. I found this and insert_text were causing duplicate characters, and insert_text was unnecessary. I wasn't able to reproduce any previous issues with the improved key_down handling.

ayax79: Going through the thread is a bit of an archaeological dig. It appears that @chrisduerr added it back in the last patch.. if I am getting the timeline right. I was getting duplicate characters too.. I'll poke around with that. I also prefer to avoid state if possible.

chrisduerr: Just fyi I wouldn't put too much faith in my path. I at least don't remember writing it at all and it was mostly an alternative suggestion to begin with.

ayax79: Sounds good.  I think I need to spend some more time debugging, as neither version seems to be working now.

ayax79:
I reverted back to the version that is roughly the same port as @kjmph's branch. I then disabled IME and everything works correctly. This makes sense as all the IME handling was added after the original fix was created. Now to figure out why...

My test case is:
- Recompile alacritty
- Run Zellij
- Alt+N creates a new pane

ayax79:
This call ```unsafe { self.interpretKeyEvents(&events_for_nsview) };``` seems to be the problem.. Looking to see if can build a modified event for it.

ayax79:
This version works and does not break IME (as far I can tell).
To use set option_as_alt to true

CHANGELOG.md
```diff
@@ -4,10 +4,23 @@ All notable changes to this project will be documented in this file.
...
+    /// device's raw character will be placed in event queues with the
+    /// Alt modifier set.
+    /// being processed as received characters. Instead, the input
+    /// device's raw character will be placed in event queues with the
+    /// Alt modifier set.
 And please only add new entries to the top of this list, right below the `# Unreleased` header.
 
 # Unreleased
-
+<<<<<<< HEAD
```
chrisduerr, line 22: This doesn't look right.

chrisduerr (commented):
Are you still working on this patch? If it's done you should probably ping one of the macOS maintainers, I don't think it's on their radar.

CHANGELOG.md
ayax79: sorry, missed when i was updating the branch to master.

ayax79: What's the best way of finding out the mac maintainers?

chrisduerr: https://github.com/rust-windowing/winit/wiki/Testers-and-Contributors

It's @madsmtm.

ayax79: Thanks!

ayax79:
@madsmtm, this is complete minus updating the feature matrix and documentation. I am not sure this is relevant for the feature matrix, and I was unsure if anymore was needed for documentation beyond the documentation on the trait. thoughts?

src/platform/macos.rs
```diff
@@ -55,6 +55,22 @@ pub trait WindowExtMacOS {
 
     /// Put the window in a state which indicates a file save is required.
     fn set_document_edited(&self, edited: bool);
+
+    /// Toggles the `Option` key being interpretted as an `Alt` modifier.
```
madsmtm, line 59: Nit:
```suggestion
    /// Set whether the OPTION key is interpreted as the ALT modifier.
    ///
    /// The default is `false`.
```

src/platform/macos.rs
```diff
@@ -55,6 +55,22 @@ pub trait WindowExtMacOS {
...
+    ///
+    /// This will ignore diacritical marks and accent characters from
+    /// being processed as received characters. Instead, the input
+    /// device's raw character will be placed in event queues with the
+    /// Alt modifier set.
+    fn set_option_as_alt(&self, option_as_alt: bool);
+
+    /// Gets whether the `Option` key sis being interpretted as an `Alt` modifier
```
madsmtm, line 67: Nit:
```suggestion
    /// Whether the OPTION key is interpreted as the ALT modifier.
```

src/platform_impl/macos/window.rs
```diff
@@ -153,6 +155,7 @@ pub struct SharedState {
     /// transitioning back to borderless fullscreen.
     save_presentation_opts: Option<NSApplicationPresentationOptions>,
     pub current_theme: Option<Theme>,
+    pub(crate) ignore_alt_modifier: bool,
```
madsmtm, line 158: Nit: Call the variable `option_as_alt` everywhere

src/platform_impl/macos/appkit/event.rs
```diff
@@ -123,6 +152,15 @@ extern_methods!(
...
         pub fn stage(&self) -> NSInteger;
 
+        #[sel(isARepeat)]
+        pub fn is_a_repeat(&self) -> bool;
+
+        #[sel(windowNumber)]
+        pub fn window_number(&self) -> NSInteger;
+
+        #[sel(timestamp)]
+        pub fn timestamp(&self) -> f64;
```
madsmtm, line 162: Nit: Should return `NSTimeInterval`
```suggestion
        pub fn timestamp(&self) -> NSTimeInterval;
```

examples/window_option_as_alt.rs
```diff
@@ -0,0 +1,69 @@
...
+        control_flow.set_wait();
+        // println!("{:?}", event);
+
+        match event {
+            Event::WindowEvent {
+                event: WindowEvent::CloseRequested,
+                window_id,
+            } if window_id == window.id() => control_flow.set_exit(),
+            Event::WindowEvent { event, .. } => match event {
+                WindowEvent::KeyboardInput {
```
madsmtm, line 40: Nit: React to mouse input instead, since this option modifies how keyboard input works

src/platform/macos.rs
```diff
@@ -55,6 +55,22 @@ pub trait WindowExtMacOS {
 
     /// Put the window in a state which indicates a file save is required.
     fn set_document_edited(&self, edited: bool);
+
+    /// Toggles the `Option` key being interpretted as an `Alt` modifier.
+    ///
+    /// This will ignore diacritical marks and accent characters from
+    /// being processed as received characters. Instead, the input
+    /// device's raw character will be placed in event queues with the
+    /// Alt modifier set.
```
madsmtm, line 64: Reading this, I'm still unsure of what the intended effect is? Maybe you could add a few examples of the desired behaviour, something like this:

```rust
// Press OPTION/ALT key
KeyboardInput { state: Pressed, keycode: Some(Alt), ... }
ModifiersChanged(ALT)

// Press E key
match (ime_allowed, option_as_alt) {
    (true, false) => {
        Ime::Preedit("´", Some((0, 0)))
    }
    (true, true) | (false, true) => {
        KeyboardInput { state: Pressed, keycode: Some(E), .. }
        ReceivedCharacter('e')
        KeyboardInput { state: Released, keycode: Some(E), .. }
    }
    (false, false) => {
        KeyboardInput { state: Pressed, keycode: Some(E), .. }
        KeyboardInput { state: Released, keycode: Some(E), .. }
    }
}

// Release OPTION/ALT key
KeyboardInput { state: Released, keycode: Some(Alt), ... }
ModifiersChanged()

// Press E key again
if ime_allowed && !option_as_alt {
    Ime::Preedit("", None)
    Ime::Commit("é")
} else {
    KeyboardInput { state: Pressed, keycode: Some(E), .. }
    ReceivedCharacter('e')
    KeyboardInput { state: Released, keycode: Some(E), .. }
}
```

(And the same for an example involving e.g. Pinyin)

madsmtm (changes requested):
Thanks for the ping, it was on my radar, I just have a rather long backlog atm.

Haven't looked much at the actual implementation, see my last comment; I'm unsure of what precisely the intended effect is, so I'm unable to ensure it works as expected ;)

src/platform/macos.rs
ayax79: How about the use case and where I have utilized the fix... 

Basically, when enabled, option should NEVER send an extended character, giving the same behavior as the ALT key on other operating systems. I imagine this feature is really only critical in apps like Alacritty, but it is huge for Alacritty. 

I picked up this issue because I use the combination Alacritty and Zellij on linux. I would like to use the same combination on my mac work machine. Zellij is pretty unusable on Alacritty on the mac due to the option key's default behavior of sending extended characters versus just acting as an ALT key. This actually causes a pretty poor experience in a wide range of terminal apps -- anything that requires an ALT key. 

Every other terminal on the mac (Terminal.app, Iterm.app, Kitty) have the ability to disable sending extended characters, like this fix implements. 

I have been using this fix daily for a while now. If you would like try it, my fork of Alacritty where I have utilized this fix is here: https://github.com/ayax79/alacritty/tree/option_as_alt

Here is my alacritty config: https://github.com/ayax79/config/blob/main/alacritty/alacritty.yml
The relevant section is ```option_as_alt: true``` under ```window:```

When enabled if you run zellij, OPT+n should open a new terminal pane. When disabled OPT+n should give you ˜. 

This has been an outstanding issue in alacritty for a long time: https://github.com/alacritty/alacritty/issues/62

examples/window_option_as_alt.rs
ayax79: Updated this whole example to be more useful.

ucabgao:
Any update on this PR? Have been waiting for this feature for a long time as an Alacritty user.

ayax79:
> Any update on this PR? Have been waiting for this feature for a long time as an Alacritty user.

I think we are just waiting for feedback from @madsmtm

src/platform_impl/macos/view.rs
```diff
@@ -514,6 +524,7 @@ declare_class!(
 
             let ime_related = was_in_preedit || now_in_preedit || text_commited;
 
+            // todo - wouldn't self.state.forward_key_to_app always be false?
```
madsmtm, line 527: Nope, `interpretKeyEvents:` calls `doCommandBySelector:` internally, so that may change it

src/platform_impl/macos/view.rs
```diff
@@ -484,7 +485,10 @@ declare_class!(
             }
             let was_in_preedit = self.state.ime_state == ImeState::Preedit;
 
-            let characters = get_characters(event, false);
+            let ev_mods = event_mods(event);
+            let option_as_alt = self.window().is_option_as_alt();
+            let ignore_alt_characters = option_as_alt && ev_mods.alt() && !ev_mods.ctrl();
```
madsmtm, line 490: I think you'd want to ignore it if `ev_mods.logo()` is set as well - at least that's what I think Terminal.app does?

CHANGELOG.md
```diff
@@ -7,7 +7,7 @@ Please keep one empty line before and after all headers. (This is required for `
 And please only add new entries to the top of this list, right below the `# Unreleased` header.
 
 # Unreleased
-
+- On macOS, added `WindowExtMacOS::is_option_as_alt` and `WindowExtMacOS::set_option_as_alt` to ignore diacritical marks and accent characters from being processed as received characters. Instead, the input device's raw character will be placed in event queues with the Alt modifier set.
```
madsmtm, line 10: Nit: space before the new entry
```suggestion

- On macOS, added `WindowExtMacOS::is_option_as_alt` and `WindowExtMacOS::set_option_as_alt` to ignore diacritical marks and accent characters from being processed as received characters. Instead, the input device's raw character will be placed in event queues with the Alt modifier set.
```

madsmtm (changes requested):
What I was mostly worried about was how this interacts with IME, and as far as I can tell it seems like it doesn't negatively impact that at all, so that's nice!

I looked through the implementation, looks good, only a few concerns.

Thanks for your patience!

kchibisov:
I'll add one more note, that such behavior is desired for only e.g. one of the `Option` modifiers keeping the other one as before. Would it be possible to adjust for that, like say **Left as Alt** or **Right as Alt**, so the user could use both at the same time.

madsmtm:
So something like:
```rust
enum OptionAsAlt {
    All,
    OnlyRight,
    OnlyLeft,
    None,
}
```

Though I suspect it will be a bit tricky to implement, since macOS doesn't easily expose which OPTION button is being held down.

kchibisov:
>Though I suspect it will be a bit tricky to implement, since macOS doesn't easily expose which OPTION button is being held down. (So I'd be fine with deferring that).

I think you have that via the raw key event? At least in the past we were able to determine left/right alt?

madsmtm:
Yeah, it's definitely possible to know when the left or the right ALT button has been pressed, but (at least as far as I know) macOS doesn't track that information for us (like it [otherwise does](https://developer.apple.com/documentation/appkit/nsevent/1534405-modifierflags?language=objc)), hence why it might be a bit tricky.

Perhaps we can leave the user to implement it themselves on top of `set_option_as_alt`? Something like:
```rust
let mut set_option_as_lalt = true;

match event {
    WindowEvent::KeyboardInput {
        input: KeyboardInput {
            virtual_keycode: Some(VirtualKeyCode::LAlt),
            state,
            ..
        },
        ..
    } => {
        if set_option_as_lalt {
            window.set_option_as_alt(state == ElementState::Pressed);
        }
    }
    _ => {}
}
```

(Assuming it works? Would be nice to have the example updated with it if so.)

saamalik:
@ayax79 - would be great to get this merged in. Sorry, I can't help much =)

ayax79:
Sorry... I just started a new job and I have been pretty slammed. Hoping to get back to this soon.

kchibisov:
I've taken the liberty to update this branch and add the requested functionality. I've tested on my macOS machine with the example in the repo, and the events are correct for all left/right/both/none variations.

@madsmtm could you take a look?

examples/window_option_as_alt.rs
```diff
@@ -0,0 +1,67 @@
...
+fn main() {
+    let event_loop = EventLoop::new();
+
+    let window = WindowBuilder::new()
+        .with_title("A fantastic window!")
+        .with_inner_size(winit::dpi::LogicalSize::new(128.0, 128.0))
+        .build(&event_loop)
+        .unwrap();
+
+    let mut option_as_alt = OptionAsAlt::None;
```
madsmtm, line 26: Since we have a getter method `WindowExtMacOS::option_as_alt`, let's use that instead of storing the value here

src/platform_impl/macos/appkit/event.rs
```diff
@@ -123,13 +152,34 @@ extern_methods!(
...
         pub fn characters(&self) -> Option<Id<NSString, Shared>> {
             unsafe { msg_send_id![self, characters] }
         }
 
         pub fn charactersIgnoringModifiers(&self) -> Option<Id<NSString, Shared>> {
             unsafe { msg_send_id![self, charactersIgnoringModifiers] }
         }
+
+        pub fn lalt_pressed(&self) -> bool {
+            const NX_DEVICELALTKEYMASK: u32 = 0x00000020;
```
madsmtm, line 173: Nit: This would be very nice to have documented, including a link to the [source](https://github.com/apple-oss-distributions/IOHIDFamily/blob/IOHIDFamily-1892.60.4/IOHIDSystem/IOKit/hidsystem/IOLLEvent.h#L258-L259).

And e.g. put it next to `NSEventModifierFlags`.

src/platform_impl/macos/view.rs
```diff
@@ -463,7 +464,16 @@ declare_class!(
...
 
-            let characters = get_characters(event, false);
+            // Get the characters from the event.
+            let ev_mods = event_mods(event);
+            let ignore_alt_characters = match self.window().option_as_alt() {
+                OptionAsAlt::OnlyLeft if event.lalt_pressed() => true,
+                OptionAsAlt::OnlyRight if event.ralt_pressed() => true,
+                OptionAsAlt::Both if ev_mods.alt() => true,
+                _ => false,
+            } && !ev_mods.ctrl();
```
madsmtm, line 474: Still think we may want to check `!ev_mods.logo()` as well - at least that's what I think Terminal.app does?

madsmtm (approved):
Implementation looks fine, only a few nits wrt. documentation and such

src/platform_impl/macos/view.rs
kchibisov: But the `logo` shouldn't modify input chars at all?

src/platform_impl/macos/appkit/event.rs
kchibisov: You'd be surprised how many of the sources are there for it. It's basically from `CGEventFlags` from what I know or similar. It least this part of the code is also present in the `minifb`.

src/platform_impl/macos/view.rs
kchibisov: I guess it would make sense to have `logo` just for consistency, since otherwise you tell to modify logo.

src/platform_impl/macos/appkit/event.rs
kchibisov: I've added just so it could be found in case someone decides to expand the functionality.

kchibisov:
Thanks, @ayax79

madsmtm:
Perfect, thanks!

chrisduerr:
Yeah this has been a long standing issue on macOS that wasn't *particularly* difficult but simply didn't get resolved because there was no macOS developer interested in stepping up and fixing it. So thanks for stepping up.

kjmph:
Thanks everyone! I can't wait to try this out!

saamalik:
Woot woot! this is awesome! Thanks team!

jkp:
Just pulled and built the latest - am unsure how to actually enable this from the config though?

kchibisov:
Yes, the `alacritty.yml` has the documentation for that stuff.

PR #2581, opened 2022-12-03
Retain `WS_MAXIMZE` when unminmizing a maximized window

amrbashir:

For a reproduction of the bug:
```
git clone https://github.com/amrbashir/winit --branch fix/windows/restoring-minimized-maximized
cd winit
git checkout cf92e881526767db13f3549ac091986205a213f6
cargo run --example window_debug
```
and follow the instructions displayed on the terminal

### DIFF
--- a/src/platform_impl/windows/window_state.rs
+++ b/src/platform_impl/windows/window_state.rs
@@ -110,8 +110,6 @@ bitflags! {
         /// Drop shadow for undecorated windows.
         const MARKER_UNDECORATED_SHADOW = 1 << 17;
 
-        const MAGIC = 1 << 18;
-
         const EXCLUSIVE_FULLSCREEN_OR_MASK = WindowFlags::ALWAYS_ON_TOP.bits;
     }
 }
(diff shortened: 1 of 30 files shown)


msiglreith (commented):
I looked a bit locally into this as I would like to avoid the extra flag.
The style change in `apply_diff` seems to cause the issue with the maximized state - `MINIMIZED` and `MAXIMIZED` are handled beforehands tho using the `ShowWindow` calls. Removing the corresponding flags from the `diff` seems to work as well

amrbashir:
> Removing the corresponding flags from the `diff` seems to work as well

Sorry, not quite following what you mean.

src/platform_impl/windows/window_state.rs
```diff
@@ -114,6 +114,12 @@ bitflags! {
         /// Drop shadow for undecorated windows.
         const MARKER_UNDECORATED_SHADOW = 1 << 20;
 
+        /// When minmizing a maximized Windows, `WM_SIZE` which we use to update the `MAXIMIZED` bit
+        /// is fired with `wparam` set to `SIZE_MINIMIZED` and thus the `MAXIMIZED` bit will be unset
+        /// and when un-minimizing the window, the `MAXIMIZED` bit will still be unset and later on
+        /// in `apply_diff` when `new.to_window_styles()` is called, it will not add `WS_MAXIMIZE` window style.
+        const MARKER_WAS_MAXIMIZED = 1 << 18;
```
maroider, line 121: This conflicts with `IGNORE_CURSOR_EVENT`. Did you rebase without checking that the flag was still non-conflicting?

amrbashir: yeah I forgot to check. Good catch! will fix it now.

msiglreith:
> > Removing the corresponding flags from the `diff` seems to work as well
> 
> Sorry, not quite following what you mean.

I made a commit on your reproduction commit which shows my suggestion: https://github.com/msiglreith/winit/commit/cf7f052a0872a60ac374b7f66a37de2065fb509f

Issue #2568, opened 2022-11-24
Windows maximize size incorrect on Windows without decorations


chrisduerr:
See https://github.com/alacritty/alacritty/issues/6515:

![image](https://user-images.githubusercontent.com/25470747/202878783-3b943f98-53be-4229-a2bd-d3e38d4bf8f0.png)

It seems like the calculation for maximized size is always done based on the window size including decorations.

PR #2584, opened 2022-12-06
On Windows, fix window size for maximized, undecorated windows

msiglreith:
Mostly takes the relevant pieces from https://github.com/rust-windowing/winit/pull/1891 regarding max window handling which haven't been included so far.
Attributed @amrbashir in the commit accordingly, I hope that's fine!

Fixes https://github.com/rust-windowing/winit/issues/2568


src/platform_impl/windows/event_loop.rs
```diff
@@ -978,19 +978,27 @@ unsafe fn public_window_callback_inner<T: 'static>(
...
-            // ahead of the window surface. Currently, there seems no option to achieve this
-            // with the Windows API.
-            if window_flags.contains(WindowFlags::MARKER_UNDECORATED_SHADOW) {
-                let params = &mut *(lparam as *mut NCCALCSIZE_PARAMS);
+            let params = &mut *(lparam as *mut NCCALCSIZE_PARAMS);
+
+            if util::is_maximized(window) {
+                // Limit the window size when maximized to the current monitor.
+                // Otherwise it would include the non-existent decorations.
+                let monitor = monitor::current_monitor(window);
```
amrbashir, line 986: using `monitor::current_monitor` which uses `MonitorFromWindow` and `MONITOR_DEFAULTTONEAREST` is unreliable in this situation, see https://github.com/MicrosoftEdge/WebView2Feedback/issues/2549#issuecomment-1180843867, instead, `MonitorFromRect` should be used

amrbashir (changes requested):
Thank you for the attribution, unfortunately a bug with this code came to light since then, see below.

kchibisov:
@msiglreith is this good to go?

msiglreith:
not yet, I will update it today

Issue #2580, opened 2022-12-01
Possibility of supporting the ability to make a window visible but not active?


fgimian:
Hey there folks, hope all is well 😄 

As per the code at https://github.com/rust-windowing/winit/blob/master/src/platform_impl/windows/window_state.rs#L304, winit will use the `SW_SHOW` flag when the window is set to visible which will both make it visible and activate the window (i.e. switch to it).  [Here](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-showwindow) are the relevant SDK docs.

Since I'm developing a widget application which is really intended as an overlay, I'm using window settings such as these below:

```rust
    let window_builder = WindowBuilder::new()
        .with_title("TotalMix Volume Control")
        .with_always_on_top(true)
        .with_decorations(false)
        .with_skip_taskbar(true)
        .with_drag_and_drop(false)
        .with_resizable(false)
        .with_transparent(true)
...
                // Allow the mouse cursor to pass through the window.
                gl_window.window().set_cursor_hittest(false).unwrap();
...
```

This works great but I'd really like the ability to make the window visible without activating it (using the `SW_SHOWNOACTIVATE`  flag) so that focus is not taken away from whatever window the user is working on.  I've verified this indeed works exactly as expected too.

Do you think this is something we could support in winit?  I'm not sure if this is possible in all OSs but I know that it definitely works on Windows.

Thanks heaps!
Fotis

madsmtm:
Something similar is probably possible on macOS using [`-[NSWindow orderFront:]`](https://developer.apple.com/documentation/appkit/nswindow/1419495-orderfront?language=objc).

Issue #1077, opened 2019-07-29
Focused issue when creating multiple new windows at once.


hardiesoft:
I have noticed different behavior of when the WindowEvent::Focused(bool) event is dispatched, depending whether you are on Windows or MacOS.

I've modified the [multiwindow example here to illustrate these differences:](https://github.com/hardiesoft/winit/blob/focus_on_creation_issue/examples/multiwindow.rs)

On running the example on windows I get the following output:
New window WindowId(WindowId(0x40e7a))
New window WindowId(WindowId(0x30edc))
Lost focus: WindowId(WindowId(0x40e7a))

On macos, I get:
New window WindowId(Id(140494821671056))
New window WindowId(Id(140494818628096))
Got focus: WindowId(Id(140494818628096))

The windows example seems more wrong than the macos example, since we never get an event for the second window that gained focus when the first window lost it.

Depending on your point of view though, the macos example could also be wrong if we're aiming for consistency with the model that each window gains focus when it is created, and loses it when another one is created afterwards.

I'm happy to put together a PR for this issue, but I'd first like to start some discussion around what the correct behavior should be.

As an aside, I also think winit is missing an APIs to programmatically set window focus and to control stacking order of windows.  If it's agreed that those pieces of functionality are in scope for winit, then I'll write up issues for them.

ZeGentzy:
This is unfortunate. I think it would make sense if on all platforms, a lost focus on the old window then a got focus on the new window were _both_ issued, in that order.

@murarth Do you know what's the current behavior on x11?

> As an aside, I also think winit is missing an APIs to programmatically set window focus and to control stacking order of windows. If it's agreed that those pieces of functionality are in scope for winit, then I'll write up issues for them.

I think that would be in scope.

hardiesoft:
I've just tested on x11, and happily it does what you'd expect:

New window WindowId(X(WindowId(48234497)))
New window WindowId(X(WindowId(48234498)))
Got focus: WindowId(X(WindowId(48234497)))
Lost focus: WindowId(X(WindowId(48234497)))
Got focus: WindowId(X(WindowId(48234498)))

ZeGentzy:
What version of winit did you test? Master might have this delay fixed.

hardiesoft:
I was building from master.

murarth:
I'm getting the same result as @hardiesoft for window creation events. However, I'm unable to reproduce any delay in events.

[...]

PR #2585, opened 2022-12-06
Add `WindowBuilder::with_active`

amrbashir:

Porting https://github.com/tauri-apps/tao/commit/e42ff07190c37b6b8a7808900133838247a730df

Fixes https://github.com/rust-windowing/winit/issues/2580

### DIFF
--- a/src/platform_impl/windows/window_state.rs
+++ b/src/platform_impl/windows/window_state.rs
@@ -116,6 +116,8 @@ bitflags! {
         /// Drop shadow for undecorated windows.
         const MARKER_UNDECORATED_SHADOW = 1 << 20;
 
+        const MARKER_NO_ACTIVATE = 1 << 18;
+
         const EXCLUSIVE_FULLSCREEN_OR_MASK = WindowFlags::ALWAYS_ON_TOP.bits;
     }
 }
@@ -306,8 +308,14 @@ impl WindowFlags {
         }
 
         if new.contains(WindowFlags::VISIBLE) {
+            let flag = if self.contains(WindowFlags::MARKER_NO_ACTIVATE) {
+                self.set(WindowFlags::MARKER_NO_ACTIVATE, false);
+                SWP_NOACTIVATE
+            } else {
+                SW_SHOW
+            };
             unsafe {
-                ShowWindow(window, SW_SHOW);
+                ShowWindow(window, flag);
             }
         }
 
--- a/src/platform_impl/windows/window_state.rs
+++ b/src/platform_impl/windows/window_state.rs
@@ -116,7 +116,7 @@ bitflags! {
         /// Drop shadow for undecorated windows.
         const MARKER_UNDECORATED_SHADOW = 1 << 20;
 
-        const MARKER_NO_ACTIVATE = 1 << 18;
+        const MARKER_ACTIVATE = 1 << 18;
 
         const EXCLUSIVE_FULLSCREEN_OR_MASK = WindowFlags::ALWAYS_ON_TOP.bits;
     }
@@ -308,8 +308,8 @@ impl WindowFlags {
         }
 
         if new.contains(WindowFlags::VISIBLE) {
-            let flag = if self.contains(WindowFlags::MARKER_NO_ACTIVATE) {
-                self.set(WindowFlags::MARKER_NO_ACTIVATE, false);
+            let flag = !if self.contains(WindowFlags::MARKER_ACTIVATE) {
+                self.set(WindowFlags::MARKER_ACTIVATE, true);
                 SWP_NOACTIVATE
             } else {
                 SW_SHOW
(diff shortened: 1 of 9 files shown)


kchibisov (changes requested):
I'm fine with the change, however we should solve the focusing on startup issue first. Given that you don't add any new events, I'd assume that you rely on window being focused.

See this issue
https://github.com/rust-windowing/winit/issues/1077

I'd advise to add a synthetic event(Basically send Focused(false)/Focused(true) where it makes sense) on platform where window is focused/unfocused by default.

By default we should also state that the window is unfocused.

amrbashir:
> I'm fine with the change, however we should solve the focusing on startup issue first. Given that you don't add any new events, I'd assume that you rely on window being focused.

I think on all platforms, a newly spawned window always gains focus, and this method mainly exists to stop that behavior which is useful for a widgets/overlay app where you don't necessarily need focus and merely want to show the window on Top with some content.

> I'd advise to add a synthetic event(Basically send Focused(false)/Focused(true) where it makes sense) on platform where window is focused/unfocused by default.

I am okay with synthesizing a Focus event, should that be implemented in another PR or in this one? 

> By default we should also state that the window is unfocused.

I think the opposite is true, no? a window by default gains focus, unless `with_active` is used with `false`?
or did you mean that users should base their logic on the window not having focus and handle the synthesized `Focus` event instead?

amrbashir:
Looking at #1077, using winit's master branch, I get the correct events order on Windows that matches x11 here https://github.com/rust-windowing/winit/issues/1077#issuecomment-516211759.
Unfortunately, I won't be able to test on macOS or Wayland so I think this should be fixed in another PR, is this considered a hard blocker?

kchibisov:
So what I think we should do is to:

1. Say in docs for `Window::new` that **Focus** state of the window will be delivered via explicit event. And say that initial state of the window is **unfocused**...
2. Emit explicitly whether the window is focused or not from the `new` (this is done on Wayland already).

So if the new window is always focused then you send explicit focus, if it's not focused on start or may not be focused you send unfocused, but be careful here, since we don't want to end up unfocused.

@madsmtm do you know the focus situation on macOS on start? I'd assume it's always focused, no? So we'd need an explicit focus.

amrbashir:
If I understand this correctly the flow should be:
1. create window (unfocused)
2. emit `Focused(false)` event
3. create second window (unfocused)
4. emit `Focused(false)` event
5. emit `Focused(true)` event for the first window

kchibisov:
>emit Focused(true) event for the first window

For the one that is focused by the os. But yes, something like that. If you can guarantee that Os will send `Focused(true)` then it's fine, I guess.

amrbashir:
I tested on Windows and according to this comment https://github.com/rust-windowing/winit/issues/1077#issuecomment-516211759 , that seems to be the current behavior and you said that Wayland synthesize this so only macOS is left which I cannot test.

kchibisov:
@madsmtm ping wrt macOS.

madsmtm:
`windowDidBecomeKey:` is run regardless of how the window became focused, so the `Focused(_)` event is correctly tracked on macOS.

The window is assumed to be unfocused by default, so if we need to match the other platforms here, we should emit a `Focused(false)` event once the window is created (a simple `AppState::queue_event(...)` inside `Window::new` should suffice).

madsmtm:
I've pushed a fix that does this now

madsmtm (approved):
This looks fine to me.

(Note that, like many things, it won't work on macOS if done outside the event loop, see [the readme](https://github.com/rust-windowing/winit#macos). But that's fine for now).

src/window.rs
```diff
@@ -403,6 +405,17 @@ impl WindowBuilder {
         self
     }
 
+    /// Whether the window will be initially focused or not.
+    ///
+    /// ## Platform-specific:
+    ///
+    /// **Android / iOS / X11 / Wayland:** Unsupported.
```
kchibisov, line 412:
```suggestion
    /// **Android / iOS / X11 / Wayland / Orbital:** Unsupported.
```

kchibisov:
>I tested on Windows and according to this comment https://github.com/rust-windowing/winit/issues/1077#issuecomment-516211759 for X11 , that seems to be the current behavior and you said that Wayland synthesize this so only macOS is left which I cannot test.

I've retested on X11 and it wasn't like that. Though, I've pushed a fix for that. Be aware that we can't use `CreateNotify` given that for whatever reason Xorg doesn't bother sending it (Or maybe it's an XWayland bug?).

kchibisov (approved):
I've added X11 impl and slightly changed the docs for the `with_active`. Should be good now.

src/platform_impl/windows/window_state.rs
```diff
@@ -116,6 +116,8 @@ bitflags! {
         /// Drop shadow for undecorated windows.
         const MARKER_UNDECORATED_SHADOW = 1 << 20;
 
+        const MARKER_NO_ACTIVATE = 1 << 18;
```
msiglreith, line 119: Should invert the flag as it's easier to read imo and adjust the logic accordingly - `MARKER_ACTIVATE`

PR #2586, opened 2022-12-09
use cfg aliases throught the code base

amrbashir:

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -17,7 +17,6 @@ fn main() {
         os_macos: { target_os = "macos" },
         os_android: { target_os = "android" },
         os_ios: { target_os = "ios" },
-        wasm: { target_arch = "wasm32" },
         x11: { feature = "x11" },
         wayland: { feature = "wayland" },
 
--- a/build.rs
+++ b/build.rs
@@ -22,10 +22,5 @@ fn main() {
         arch_wasm: { target_arch = "wasm32" },
         x11: { feature = "x11" },
         wayland: { feature = "wayland" },
-
-        // dependencies
-        mint : { feature = "mint" },
-        serde : { feature = "serde" },
-        sctk_adwaita : { feature = "sctk-adwaita" },
     }
 }
--- a/build.rs
+++ b/build.rs
@@ -4,7 +4,7 @@ fn main() {
     // Setup cfg aliases
     cfg_aliases! {
         // Platforms
-        linux: {
+        os_linux: {
             any(
                 target_os = "linux",
                 target_os = "dragonfly",
@@ -13,12 +13,13 @@ fn main() {
                 target_os = "openbsd"
             )
         },
-        macos: { target_os = "macos" },
-        android: { target_os = "android" },
-        ios: { target_os = "ios" },
+        os_windows: { target_os = "windows" },
+        os_macos: { target_os = "macos" },
+        os_android: { target_os = "android" },
+        os_ios: { target_os = "ios" },
         wasm: { target_arch = "wasm32" },
-        x11: { all(linux, feature = "x11") },
-        wayland: { all(linux, feature = "wayland") },
+        x11: { feature = "x11" },
+        wayland: { feature = "wayland" },
 
         // dependencies
         mint : { feature = "mint" },
--- a/build.rs
+++ b/build.rs
@@ -5,22 +5,16 @@ fn main() {
     println!("cargo:rerun-if-changed=build.rs");
     // Setup cfg aliases
     cfg_aliases! {
-        // Platforms
-        os_linuxy: {
-            any(
-                target_os = "linux",
-                target_os = "dragonfly",
-                target_os = "freebsd",
-                target_os = "netbsd",
-                target_os = "openbsd"
-            )
-        },
-        os_windows: { target_os = "windows" },
-        os_macos: { target_os = "macos" },
-        os_android: { target_os = "android" },
-        os_ios: { target_os = "ios" },
-        arch_wasm: { target_arch = "wasm32" },
-        x11: { feature = "x11" },
-        wayland: { feature = "wayland" },
+        // Systems.
+        android: { target_os = "android" },
+        wasm: { target_arch = "wasm32" },
+        macos: { target_os = "macos" },
+        ios: { target_os = "ios" },
+        apple: { any(target_os = "ios", target_os = "macos") },
+        free_unix: { all(unix, not(apple), not(android)) },
+
+        // Native displays.
+        x11_platform: { all(feature = "x11", free_unix, not(wasm)) },
+        wayland_platform: { all(feature = "wayland", free_unix, not(wasm)) },
     }
 }
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,28 @@
+use cfg_aliases::cfg_aliases;
+
+fn main() {
+    // Setup cfg aliases
+    cfg_aliases! {
+        // Platforms
+        linux: {
+            any(
+                target_os = "linux",
+                target_os = "dragonfly",
+                target_os = "freebsd",
+                target_os = "netbsd",
+                target_os = "openbsd"
+            )
+        },
+        macos: { target_os = "macos" },
+        android: { target_os = "android" },
+        ios: { target_os = "ios" },
+        wasm: { target_arch = "wasm32" },
+        x11: { all(linux, feature = "x11") },
+        wayland: { all(linux, feature = "wayland") },
+
+        // dependencies
+        mint : { feature = "mint" },
+        serde : { feature = "serde" },
+        sctk_adwaita : { feature = "sctk-adwaita" },
+    }
+}
--- a/build.rs
+++ b/build.rs
@@ -1,6 +1,8 @@
 use cfg_aliases::cfg_aliases;
 
 fn main() {
+    // The script doesn't depend on our code
+    println!("cargo:rerun-if-changed=build.rs");
     // Setup cfg aliases
     cfg_aliases! {
         // Platforms
--- a/build.rs
+++ b/build.rs
@@ -4,7 +4,7 @@ fn main() {
     // Setup cfg aliases
     cfg_aliases! {
         // Platforms
-        os_linux: {
+        os_linuxy: {
             any(
                 target_os = "linux",
                 target_os = "dragonfly",
--- a/build.rs
+++ b/build.rs
@@ -17,6 +17,7 @@ fn main() {
         os_macos: { target_os = "macos" },
         os_android: { target_os = "android" },
         os_ios: { target_os = "ios" },
+        arch_wasm: { target_arch = "wasm32" },
         x11: { feature = "x11" },
         wayland: { feature = "wayland" },
 
--- a/build.rs
+++ b/build.rs
@@ -6,10 +6,11 @@ fn main() {
     // Setup cfg aliases
     cfg_aliases! {
         // Systems.
-        android: { target_os = "android" },
-        wasm: { target_arch = "wasm32" },
-        macos: { target_os = "macos" },
-        ios: { target_os = "ios" },
+        android_platform: { target_os = "android" },
+        wasm_platform: { target_arch = "wasm32" },
+        macos_platform: { target_os = "macos" },
+        ios_platform: { target_os = "ios" },
+        windows_platform: { target_os = "windows" },
         apple: { any(target_os = "ios", target_os = "macos") },
         free_unix: { all(unix, not(apple), not(android)) },
 
--- a/build.rs
+++ b/build.rs
@@ -12,7 +12,7 @@ fn main() {
         ios_platform: { target_os = "ios" },
         windows_platform: { target_os = "windows" },
         apple: { any(target_os = "ios", target_os = "macos") },
-        free_unix: { all(unix, not(apple), not(android)) },
+        free_unix: { all(unix, not(apple), not(android_platform)) },
 
         // Native displays.
         x11_platform: { all(feature = "x11", free_unix, not(wasm)) },
--- a/examples/child_window.rs
+++ b/examples/child_window.rs
@@ -1,7 +1,7 @@
-#[cfg(all(target_os = "linux", feature = "x11"))]
+#[cfg(x11)]
 use std::collections::HashMap;
 
-#[cfg(all(target_os = "linux", feature = "x11"))]
+#[cfg(x11)]
 use winit::{
     dpi::{LogicalPosition, LogicalSize, Position},
     event::{ElementState, Event, KeyboardInput, WindowEvent},
@@ -10,7 +10,7 @@ use winit::{
     window::{Window, WindowBuilder, WindowId},
 };
 
-#[cfg(all(target_os = "linux", feature = "x11"))]
+#[cfg(x11)]
 fn spawn_child_window(
     parent: u32,
     event_loop: &EventLoopWindowTarget<()>,
@@ -30,7 +30,7 @@ fn spawn_child_window(
     println!("child window created with id: {}", id);
 }
 
-#[cfg(all(target_os = "linux", feature = "x11"))]
+#[cfg(x11)]
 fn main() {
     let mut windows = HashMap::new();
 
@@ -77,7 +77,7 @@ fn main() {
     })
 }
 
-#[cfg(not(all(target_os = "linux", feature = "x11")))]
+#[cfg(not(x11))]
 fn main() {
     panic!("This example is supported only on x11.");
 }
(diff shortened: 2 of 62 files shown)


src/platform/mod.rs
```diff
@@ -21,41 +21,14 @@ pub mod android;
...
-    target_os = "windows",
-    target_os = "macos",
-    target_os = "android",
-    target_os = "linux",
-    target_os = "dragonfly",
-    target_os = "freebsd",
-    target_os = "netbsd",
-    target_os = "openbsd"
-))]
+#[cfg(any(windows, x11, wayland, target_os = "macos", target_os = "android"))]
```
maroider, line 33:
```suggestion
#[cfg(any(windows, x11, wayland, macos, android))]
```

src/platform_impl/mod.rs
```diff
@@ -58,15 +58,12 @@ impl From<Fullscreen> for RootFullscreen {
 }
 
 #[cfg(all(
-    not(target_os = "ios"),
-    not(target_os = "windows"),
-    not(target_os = "linux"),
+    not(x11),
+    not(wayland),
```
maroider, line 62:
```suggestion
    not(linux),
```
Let's keep things 1:1. There's a separate error when you don't enable either the `x11` or `wayland` features.

src/window.rs
```diff
@@ -1033,7 +1034,7 @@ impl Window {
     ///
     /// [`NSWindowSharingNone`]: https://developer.apple.com/documentation/appkit/nswindowsharingtype/nswindowsharingnone
     pub fn set_content_protected(&self, _protected: bool) {
-        #[cfg(any(target_os = "macos", target_os = "windows"))]
+        #[cfg(any(target_os = "macos", windows))]
```
maroider, line 1037:
```suggestion
        #[cfg(any(macos, windows))]
```

maroider (changes requested):
I'd prefer if we kept things as close to 1:1 as possible, so I'd prefer if `x11` and `wayland` were only aliases for their respective features, without also implying `linux`. Some `cfg` guards were also re-ordered, and I'd prefer if we kept the existing order (this isn't a huge deal, tbh, just a preference).

On the whole, though, I think this is a good change, and it makes our `cfg` guards a lot less noisy.

src/event_loop.rs
```diff
@@ -345,7 +345,7 @@ impl<T> EventLoopWindowTarget<T> {
             target_os = "freebsd",
             target_os = "netbsd",
             target_os = "openbsd",
-            target_os = "windows"
+            windows
```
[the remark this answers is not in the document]
amrbashir: clippy wasn't happy about it at all and will emit https://rust-lang.github.io/rust-clippy/master/index.html#mismatched_target_os

src/platform/mod.rs
amrbashir: same as above

src/window.rs
amrbashir: same as above, clippy

src/platform_impl/mod.rs
amrbashir: sure but I will have to switch it to be `target_os = "linux"` because of clippy

src/event_loop.rs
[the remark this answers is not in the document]
amrbashir: actually, allowing the clippy lint is a lot annoying, I have to add it in multiple files, and sometimes it is not enough and have to add it on multiple lines. 

Alternative names I considered is, `p_linux`, `p_macos`, `p_android`, `p_ios` and `p_windows`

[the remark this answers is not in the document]
maroider: Maybe `os_whatver`, since it's a bunch of `target_os = "whatever"` aliases?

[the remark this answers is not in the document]
amrbashir: yeah sounds good, should I also add one for `windows`, just for consistency?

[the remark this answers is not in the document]
maroider: Windows is a little odd, but we should do the same for windows, yeah, as I'm under the impression that `cfg(target_os = "windows")` is subtly different from `cfg(windows)`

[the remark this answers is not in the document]
amrbashir: the rust docs suggest they are slightly different, but I couldn't find out any relevant docs as to why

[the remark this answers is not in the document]
maroider: Yeah, it's not particularly clear what the difference is, but since we've been using `target_os` so far, we should just keep using it.

amrbashir:
> Some cfg guards were also re-ordered, and I'd prefer if we kept the existing order (this isn't a huge deal, tbh, just a preference)

Reverted!

maroider:
So, I checked out your changes locally and grepped through for `cfg` guards, and found that you haven't converted some `cfg` guards in the following files:
- `src/platform_impl/android/mod.rs`
- `src/platform_impl/ios/mod.rs`
- `src/platform_impl/linux/wayland/window/shim.rs`

One issue that did occur to me just now is that `os_linux` makes it seem like it is only set when targeting linux, which isn't true. I wouldn't consider it a hard blocker, but we might want to come up with a name that better conveys what `os_linux` actually means. `wasm` is also already a built-in `cfg` option, so we may want to use a `cfg` alias with a different name.

amrbashir:
> So, I checked out your changes locally and grepped through for `cfg` guards, and found that you haven't converted some `cfg` guards in the following files:
> 
> * `src/platform_impl/android/mod.rs`
> * `src/platform_impl/ios/mod.rs`
> * `src/platform_impl/linux/wayland/window/shim.rs`

Thanks, will update them too.

> One issue that did occur to me just now is that `os_linux` makes it seem like it is only set when targeting Linux, which isn't true. I wouldn't consider it a hard blocker, but we might want to come up with a name that better conveys what `os_linux` actually means.

I wanted to name it `os_unix` but that would convey support for macOS, ios and android also.

> `wasm` is also already a built-in `cfg` option, so we may want to use a `cfg` alias with a different name.

Can't I just remove the `wasm` custom alias and use the built-in cfg alias?

amrbashir:
> Can't I just remove the wasm custom alias and use the built-in cfg alias?

Looks like I can't, so I went ahead and changed it `arch_wasm`

maroider:
> I wanted to name it `os_unix` but that would convey support for macOS, ios and android also.

My ideas here are not great (mostly just `os_linuxy`).

> Can't I just remove the wasm custom alias and use the built-in cfg alias?

That's probably fine. As far as I'm aware, there's only really `wasm32` targets for `rustc`. If I were to hazard a guess, `wasm64` support will likely arrive some day, but I don't think we directly rely on anything that would prevent Winit from "just working" on `wasm64`.

maroider:
> Looks like I can't, so I went ahead and changed it `arch_wasm`

Uuh, ok. Why didn't it work?

amrbashir:
I don't know what the built-in `wasm` expands to and couldn't find it out from searching through google results but seems like it doesn't satisfy the check here on Windows 
https://github.com/rust-windowing/winit/blob/89eea64a4aa952ddb6a4612f61075f8ad622e1a9/src/platform_impl/mod.rs

see this run https://github.com/rust-windowing/winit/actions/runs/3664441578/jobs/6194851823

amrbashir:
> My ideas here are not great (mostly just `os_linuxy`).

Sounds acceptable to me for now until a better name is found.

maroider (approved):
I can't think of any further issues beyond what's already been mentioned.

build.rs
```diff
@@ -0,0 +1,29 @@
+use cfg_aliases::cfg_aliases;
+
+fn main() {
```
madsmtm, line 3: Add the following to avoid running the build script on every code change
```suggestion
fn main() {
    // The script doesn't depend on our code
    println!("cargo:rerun-if-changed=build.rs");
```

madsmtm (commented):
Apart from `os_linuxy` I don't think we really gain much from this (other than an increase in compile-times, and an extra indirection where people have to go see what `os_ios` actually means ;) ).

But I'm not that against it, and if you think it would make it easier to integrate GTK, then sure, go for it (though yeah, I still think https://github.com/rust-windowing/winit/pull/2430 is the most "correct" way to go, but eh, that may also just be sunk costs speaking).

amrbashir:
> Apart from `os_linuxy` I don't think we really gain much from this (other than an increase in compile-times, and an extra indirection where people have to go see what `os_ios` actually means ;) ).
> 
> But I'm not that against it, and if you think it would make it easier to integrate GTK, then sure, go for it (though yeah, I still think #2430 is the most "correct" way to go, but eh, that may also just be sunk costs speaking).

TBH, I feel neutral about it, I think we can integrate GTK without it, but it was a requested by @kchibisov

maroider:
> I don't think we really gain much from this (other than an increase in compile-times

Oof, I forgot about compile-times. I don't think this will add more than a couple of seconds, but any regression is unfortunate.

> TBH, I feel neutral about it, I think we can integrate GTK without it, but it was requested by @kchibisov

Can you link that request, just for posterity?

amrbashir:
https://matrix.to/#/!DGpLzJTRzBDTwZiogk:matrix.org/$9t77LEXJ2t3jIs2FAoa0oGUZZQfq1l1un2i6QhJi67w?via=libera.chat&via=matrix.org&via=matrix.fedibird.com

kchibisov:
Then, @amrbashir, since you've asked what's needs to be done to bring the GTK in the tree. The following layout from that PR should be implemented. It looks kind of ok and @madsmtm could likely help you with that, but for a working concept of such a thing you can look again at glutin.

https://github.com/rust-windowing/winit/pull/2430

https://github.com/rust-windowing/glutin/blob/master/glutin/src/lib.rs#L42 - that's how dispatch is done in glutin, but @madsmtm has slightly better version of this macro in their work.

So the `cfg` you're adding right now will be hidden inside the dispatch macros and not even seen in the code other than in module inclusion!.

build.rs
```diff
@@ -0,0 +1,31 @@
...
+        os_linuxy: {
+            any(
+                target_os = "linux",
+                target_os = "dragonfly",
+                target_os = "freebsd",
+                target_os = "netbsd",
+                target_os = "openbsd"
+            )
+        },
+        os_windows: { target_os = "windows" },
```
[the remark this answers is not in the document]
amrbashir: clippy will emit an error for using `windows` along with `target_os`

build.rs
```diff
@@ -0,0 +1,31 @@
...
+        os_macos: { target_os = "macos" },
+        os_android: { target_os = "android" },
+        os_ios: { target_os = "ios" },
```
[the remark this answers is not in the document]
amrbashir: clippy wasn't happy about it at all and will emit [rust-lang.github.io/rust-clippy/master/index.html#mismatched_target_os](https://rust-lang.github.io/rust-clippy/master/index.html#mismatched_target_os)

build.rs
```diff
@@ -0,0 +1,31 @@
...
+        // Platforms
+        os_linuxy: {
+            any(
+                target_os = "linux",
+                target_os = "dragonfly",
+                target_os = "freebsd",
+                target_os = "netbsd",
+                target_os = "openbsd"
+            )
+        },
```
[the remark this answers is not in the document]
amrbashir: is [this](https://github.com/rust-windowing/winit/pull/2586/commits/05f26956e91be7687f5a35fb5529ef7096bf4d05) what you wanted?

examples/child_window.rs
```diff
@@ -77,7 +77,7 @@ fn main() {
     })
 }
 
-#[cfg(not(all(os_linuxy, x11)))]
+#[cfg(not(all(free_unix, x11_platform)))]
```
kchibisov, line 80: x11_platform implies free_unix, so you can just use it. The same for wayland. So you'd have just wayland/x11 in the code, and not free unix. Use free_unix only when we're really aiming for that and not for x11/wayland.

madsmtm (approved):
Oh... I had understood previously that you were against https://github.com/rust-windowing/winit/pull/2430, @kchibisov, which is why I haven't worked on it more?

(Though these two PRs are not exclusive, so still 👍 from me on this one)

examples/window_run_return.rs
```diff
@@ -1,16 +1,7 @@
...
-    target_os = "windows",
-    target_os = "macos",
-    target_os = "linux",
-    target_os = "dragonfly",
-    target_os = "freebsd",
-    target_os = "netbsd",
-    target_os = "openbsd",
-    target_os = "android",
-))]
+#[cfg(any(windows, macos, free_unix, android))]
```
kchibisov, line 4: Would suggest to use `x11` and `wayland` here, since it's about display server, not the free_unix.

src/event_loop.rs
```diff
@@ -339,14 +339,7 @@ impl<T> EventLoopWindowTarget<T> {
...
     pub fn set_device_event_filter(&self, _filter: DeviceEventFilter) {
-        #[cfg(any(
-            target_os = "linux",
-            target_os = "dragonfly",
-            target_os = "freebsd",
-            target_os = "netbsd",
-            target_os = "openbsd",
-            target_os = "windows"
-        ))]
+        #[cfg(any(free_unix, windows))]
```
kchibisov, line 342: x11 or windows.

src/platform/mod.rs
```diff
@@ -15,47 +15,20 @@
 //!
 //! However only the module corresponding to the platform you're compiling to will be available.
 
-#[cfg(target_os = "android")]
+#[cfg(android)]
```
kchibisov, line 18: I wonder why it fails here? If we can't use android like that I'd suggest to change so the failure gets resolved.

src/platform/mod.rs
```diff
@@ -15,47 +15,20 @@
...
-    feature = "x11",
-    any(
-        target_os = "linux",
-        target_os = "dragonfly",
-        target_os = "freebsd",
-        target_os = "netbsd",
-        target_os = "openbsd",
-    )
-))]
+#[cfg(all(x11_platform, free_unix))]
```
kchibisov, line 30: Just X11 here.

src/platform/mod.rs
```diff
@@ -15,47 +15,20 @@
...
-    target_os = "windows",
-    target_os = "macos",
-    target_os = "android",
-    target_os = "linux",
-    target_os = "dragonfly",
-    target_os = "freebsd",
-    target_os = "netbsd",
-    target_os = "openbsd"
-))]
+#[cfg(any(windows, macos, android, free_unix))]
```
kchibisov, line 33: x11 and wayland.

src/platform_impl/linux/mod.rs
```diff
@@ -1,35 +1,29 @@
-#![cfg(any(
-    target_os = "linux",
-    target_os = "dragonfly",
-    target_os = "freebsd",
-    target_os = "netbsd",
-    target_os = "openbsd"
-))]
-
-#[cfg(all(not(feature = "x11"), not(feature = "wayland")))]
+#![cfg(free_unix)]
```
kchibisov, line 1: But this is likely fine like that.

src/platform_impl/linux/wayland/mod.rs
```diff
@@ -1,10 +1,4 @@
-#![cfg(any(
-    target_os = "linux",
-    target_os = "dragonfly",
-    target_os = "freebsd",
-    target_os = "netbsd",
-    target_os = "openbsd"
-))]
+#![cfg(free_unix)]
```
kchibisov, line 1: That's just wayland_platform.

src/platform_impl/linux/x11/mod.rs
```diff
@@ -1,10 +1,4 @@
-#![cfg(any(
-    target_os = "linux",
-    target_os = "dragonfly",
-    target_os = "freebsd",
-    target_os = "netbsd",
-    target_os = "openbsd"
-))]
+#![cfg(free_unix)]
```
kchibisov, line 1: That's just x11_platform

src/platform_impl/mod.rs
```diff
@@ -1,28 +1,22 @@
...
 #[path = "windows/mod.rs"]
 mod platform;
-#[cfg(any(
-    target_os = "linux",
-    target_os = "dragonfly",
-    target_os = "freebsd",
-    target_os = "netbsd",
-    target_os = "openbsd"
-))]
+#[cfg(free_unix)]
```
kchibisov, line 7: any x11/wayland?

kchibisov (changes requested):

src/platform/mod.rs
amrbashir: the solution @maroider came up with was to change the aliases to have `os_*` prefix

kchibisov: Could have platform_android, so it'll look like x11/wayland.

src/event_loop.rs
amrbashir: clippy will complain about unused functions on Wayland, changed it to x11, wayland or Windows

kchibisov (approved):

kchibisov:
Thanks!

PR #2593, opened 2022-12-16
Fix pointer index for multi-touch gesture in Android

haricot:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- On Android, fix pointer index for multi-touch gesture in Android
 - On Windows, retain `WS_MAXIMIZE` window style when un-minimizing a maximized window.
 - On Windows, fix left mouse button release event not being sent after `Window::drag_window`.
 - On macOS, run most actions on the main thread, which is strictly more correct, but might make multithreaded applications block slightly more.
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -506,7 +506,8 @@ impl<T: 'static> EventLoop<T> {
                             | event::TouchPhase::Ended => {
                                 Box::new(
                                     std::iter::once(motion_event.pointer_at_index(
-                                        motion_event.pointer_index(),
+                                        // Handle pointer index of multi-touch gesture
+                                        motion_event.pointer_count() - 1_usize,
                                     ))
                                 )
                             },


haricot:
This change doesn't seem to work like the wasm version of egui.
I put this PR in WIP while testing a solution otherwise I will post an issue.

haricot:
now the multi-touch interaction seems similar to the web version of egui.

haricot:
it works now, sorry for the useless commit, I can redo a clean PR if necessary.

src/platform_impl/android/mod.rs
```diff
@@ -502,15 +502,16 @@ impl<T: 'static> EventLoop<T> {
...
                             dyn Iterator<Item = android_activity::input::Pointer<'_>>,
                         > = match phase {
-                            event::TouchPhase::Started
-                            | event::TouchPhase::Ended => {
+                            event::TouchPhase::Started => {
                                 Box::new(
                                     std::iter::once(motion_event.pointer_at_index(
-                                        motion_event.pointer_index(),
+                                        // In case it could be multi-touch gestures this requires the index of the last pointer.
+                                        motion_event.pointer_count() - 1_usize,
```
rib, line 509: It's not currently clear to me just from the comment why we should be assuming the last pointer is the correct one to reference here. Is it perhaps possible to quote some Android docs that clarify a bit why the index of the last pointer is appropriate here?

As far as I understand it from [these docs](https://developer.android.com/reference/android/view/MotionEvent#overview) the motion event's action code for `Up/PointerUp/Down/PointerDown` events should contain a pointer index (the action and pointer index are combined in a single number initially which need to be separated by masking and shifting) and that's what `.pointer_index()` does (it extracts the pointer index that's encoded in the action code).

_(`Up/Down` events should implicitly relate to pointer index zero)_

The `PointerDown` events seem pretty _likely_ to correspond to the last pointer but the Android docs don't seem to promise that as far as I can tell. The e.g. say "The order in which individual pointers appear within a motion event is undefined".

src/platform_impl/android/mod.rs
```diff
@@ -502,15 +502,16 @@ impl<T: 'static> EventLoop<T> {
...
                                     ))
                                 )
                             },
-                            event::TouchPhase::Moved
+                            event::TouchPhase::Ended
```
rib, line 513: All `Up/PointerUp/Down/PointerDown` events should only relate to a single pointer as far as I know so it doesn't look right for this to move.

Overall there may be multiple active pointers during a multi-touch gesture but the `Ended` action itself should only relate to a single pointer.

Since we combine `Up` and `PointerUp` into `Ended` this change implies to me that if you were to start a multi-touch guesture with 4 fingers and then lift any finger that would result in an `Ended` event for _all_ the fingers/pointers because we'd iterate all active pointers and put them in the `Ended` phase.

rib (commented):
Thanks for looking at this.

I've noticed before that multi touch wasn't working as expected in egui on Android for some reason so there's probably something amiss here but it's not clear to me that these changes are appropriate currently.

Especially the change to emit multiple `Touch` events for a `Up` or `PointerUp` looks very surprising to me currently.

src/platform_impl/android/mod.rs
haricot: it could be the fact that the order of iterator PointersIter is reversed.

there is also this doc about it :
https://developer.android.com/develop/ui/views/touch-and-input/gestures/multi
and/or  unmask  event like:
https://github.com/rust-mobile/android-activity/compare/main...haricot:android-activity:stylus_input

src/platform_impl/android/mod.rs
haricot: Thank you for your clarification, I see this problem now.

CHANGELOG.md
```diff
@@ -909,4 +910,4 @@ _Yanked_
 - Changed the X11 fullscreen code to use `xrandr` instead of `xxf86vm`.
 - Fixed the Wayland backend to produce `Refresh` event after window creation.
 - Changed the `Suspended` event to be outside of `WindowEvent`.
-- Fixed the X11 backend sometimes reporting the wrong virtual key (#273).
+- Fixed the X11 backend sometimes reporting the wrong virtual key (#273).
```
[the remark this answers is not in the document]
haricot: unrelated change

src/platform_impl/android/mod.rs
haricot: Finally I do not think that the order is reversed, this is not a problem with a single touch since there is only one element. but in multi touch it seems to need a particular index.

I wonder if this is necessary to have access to the index  (or we already has with motion_event.action() &  motion_event.pointer_index()) :


[ACTION_POINTER_DOWN](https://developer.android.com/reference/androidx/core/view/MotionEventCompat#ACTION_POINTER_DOWN)—For extra pointers that enter the screen beyond the first. The index of the pointer that just went down can be obtained by using [getActionIndex()](https://developer.android.com/reference/androidx/core/view/MotionEventCompat#getActionIndex(android.view.MotionEvent)).

[ACTION_POINTER_UP](https://developer.android.com/reference/androidx/core/view/MotionEventCompat#ACTION_POINTER_UP)—Sent when a non-primary pointer goes up. The index of the pointer that just went up can be obtained by using [getActionIndex()](https://developer.android.com/reference/androidx/core/view/MotionEventCompat#getActionIndex(android.view.MotionEvent)).

` android-activity/game-activity-csrc/game-activity/native_app_glue/android_native_app_glue.c `


```
#define ACTION_POINTER_DOWN 0x00000005
#define ACTION_POINTER_UP 0x00000006

static bool default_motion_filter(const GameActivityMotionEvent* event) {
    // Ignore any non-touch events.
    // return event->source == SOURCE_TOUCHSCREEN;
    return (event->source == SOURCE_TOUCHSCREEN ||
event-> action == ACTION_POINTER_DOWN
event-> action ==  ACTION_POINTER_UP
);
}
```

PR #2594, opened 2022-12-18
windows: Use strict provenace for pointer casts

notgull:

Rust is planning on phasing out ptr-to-usize and usize-to-ptr conversions in the future, and only allowing them through methods that explicitly preserve provenance. See rust-lang/rust#95228 for more information. As these kinds of casts are most prevalent in the Windows implementation of `winit`, I've implemented this there.

I'm not too familiar with this codebase, so please tell me if there's anything I've missed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,8 +42,8 @@ wayland-dlopen = ["sctk/dlopen", "wayland-client/dlopen"]
 wayland-csd-adwaita = ["sctk-adwaita", "sctk-adwaita/ab_glyph"]
 wayland-csd-adwaita-crossfont = ["sctk-adwaita", "sctk-adwaita/crossfont"]
 wayland-csd-adwaita-notitle = ["sctk-adwaita"]
-android-native-activity = [ "android-activity/native-activity" ]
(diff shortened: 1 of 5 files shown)


notgull:
It looks like the Clippy lints are failing. Some of it is on platforms I didn't touch, so I don't think it's my fault? (Except for the ones about ptr transmutes, that's definitely me). Is there an active issue/PR for this?

madsmtm (commented):
If we want to do this I'd probably rather (temporarily) use the `sptr` crate

notgull:
> If we want to do this I'd probably rather (temporarily) use the `sptr` crate.

Sounds good, I've replaced the hand-rolled implementation with `sptr`.

> But in general I don't really think `winit` is the right place to experiment with this, since we can't actually test with `miri` whether the provenance we're doing is correct.

I think that `strict_provenance` will probably be deployed ecosystem wide once it hits. Pointer casts will likely become warnings. This helps us get ahead of the curve. Besides, I think it makes the code more clear.

notgull:
I don't feel like maintaining this branch. I'll close this until strict provenance becomes more prominent.

PR #2596, opened 2022-12-20
fix `unnecessary cast` lint

amrbashir:

### DIFF
--- a/examples/child_window.rs
+++ b/examples/child_window.rs
@@ -32,7 +32,7 @@ fn spawn_child_window(
 
 #[cfg(all(target_os = "linux", feature = "x11"))]
 fn main() {
-    let mut windows = HashMap::new();
+    let mut windows: HashMap<u64, Window> = HashMap::new();
 
     let event_loop: EventLoop<()> = EventLoop::new();
     let parent_window = WindowBuilder::new()
@@ -69,7 +69,7 @@ fn main() {
                         },
                     ..
                 } => {
-                    spawn_child_window(root, event_loop, &mut windows);
+                    spawn_child_window(root as _, event_loop, &mut windows);
                 }
                 _ => (),
             }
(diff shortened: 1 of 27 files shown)


examples/child_window.rs
```diff
@@ -12,27 +12,27 @@ use winit::{
 
 #[cfg(all(target_os = "linux", feature = "x11"))]
 fn spawn_child_window(
-    parent: u32,
+    parent: u64,
     event_loop: &EventLoopWindowTarget<()>,
-    windows: &mut HashMap<u32, Window>,
+    windows: &mut HashMap<u64, Window>,
```
madsmtm, line 17: Maybe just use `std::raw::c_ulong` here?

src/platform_impl/linux/x11/monitor.rs
```diff
@@ -108,7 +108,11 @@ impl std::hash::Hash for MonitorHandle {
 #[inline]
 pub fn mode_refresh_rate_millihertz(mode: &XRRModeInfo) -> Option<u32> {
     if mode.dotClock > 0 && mode.hTotal > 0 && mode.vTotal > 0 {
-        Some((mode.dotClock as u64 * 1000 / (mode.hTotal as u64 * mode.vTotal as u64)) as u32)
+        #[cfg(target_pointer_width = "64")]
+        let dot_clock = mode.dotClock;
+        #[cfg(target_pointer_width = "32")]
+        let dot_clock = mode.dotClock as u64;
+        Some((dot_clock * 1000 / (mode.hTotal as u64 * mode.vTotal as u64)) as u32)
```
madsmtm, line 115: Maybe just silence the lint here?

madsmtm (changes requested):
This will remove our 32-bit support on macOS and iOS, which, while they may be old and not tested in CI, I'm against it, mostly since an old iPad (`armv7-apple-ios`) is currently the only iOS device that I can test on

madsmtm (approved):
Cool, thanks!

PR #2598, opened 2022-12-20
Replace lazy window message ids with a slimmer version

notgull:

In the current Win32 backend, a `Lazy<u32>` is used to initialize and store `winit`-specific window messages. However, a `Lazy` incurs two extra pointers on top of the word that the `u32` takes up. I have created a new racy implementation using an `AtomicU32` that only uses one additional pointer. This should be a mild improvement over the current setup.

### DIFF
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -646,7 +646,7 @@ impl LazyMessageId {
         // Store the new ID. Since `RegisterWindowMessageA` returns the same value for any given string,
         // the target value will always either be a). `INVALID_ID` or b). the correct ID. Therefore a
         // compare-and-swap operation here (or really any consideration) is never necessary.
-        self.id.store(new_id, Ordering::Relaxed); 
+        self.id.store(new_id, Ordering::Relaxed);
 
         new_id
     }
@@ -671,7 +671,8 @@ static CANCEL_WAIT_UNTIL_MSG_ID: LazyMessageId = LazyMessageId::new("Winit::Canc
 pub static DESTROY_MSG_ID: LazyMessageId = LazyMessageId::new("Winit::DestroyMsg\0");
 // WPARAM is a bool specifying the `WindowFlags::MARKER_RETAIN_STATE_ON_SIZE` flag. See the
 // documentation in the `window_state` module for more information.
-pub static SET_RETAIN_STATE_ON_SIZE_MSG_ID: LazyMessageId = LazyMessageId::new("Winit::SetRetainMaximized\0");
+pub static SET_RETAIN_STATE_ON_SIZE_MSG_ID: LazyMessageId =
+    LazyMessageId::new("Winit::SetRetainMaximized\0");
 static THREAD_EVENT_TARGET_WINDOW_CLASS: Lazy<Vec<u16>> =
     Lazy::new(|| util::encode_wide("Winit Thread Event Target"));
 /// When the taskbar is created, it registers a message with the "TaskbarCreated" string and then broadcasts this message to all top-level windows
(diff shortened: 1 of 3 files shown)


src/platform_impl/windows/event_loop.rs
```diff
@@ -599,40 +601,90 @@ impl<T: 'static> EventLoopProxy<T> {
...
+    /// Get the message ID.
+    pub fn get(&self) -> u32 {
+        // Load the ID.
+        let mut id = self.id.load(Ordering::Relaxed);
+
+        // If the ID is invalid, create one.
+        if id == INVALID_ID {
+            // Register the message.
+            assert!(self.name.ends_with('\0'));
+            let new_id = unsafe { RegisterWindowMessageA(self.name.as_ptr()) };
```
madsmtm, line 634: Nit: Add safety comment

madsmtm (commented):
Perhaps we should instead use `WM_USER + 1`, `WM_USER + 2`, ... as suggested by the documentation of [`RegisterWindowMessageA`](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-registerwindowmessagea)?

notgull:
> Perhaps we should instead use `WM_USER + 1`, `WM_USER + 2`, ... as suggested by the documentation of [`RegisterWindowMessageA`](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-registerwindowmessagea)?

My gut says that, for the purpose of safety, we want to avoid accidentally stepping on the toes of anyone else. The best way of avoiding that is `RegisterWindowMessageA`.

notgull:
@msiglreith Are there any blockers to this being merged?

src/platform_impl/windows/event_loop.rs
```diff
@@ -599,40 +601,91 @@ impl<T: 'static> EventLoopProxy<T> {
...
+                0,
+                "RegisterWindowMessageA returned zero for '{}': {}",
+                self.name,
+                std::io::Error::last_os_error()
+            );
+
+            // Try to store the new ID.
+            let swapped_id = self
+                .id
+                .compare_exchange(id, new_id, Ordering::SeqCst, Ordering::SeqCst)
```
madsmtm, line 649: I don't think sequential consistency is required here, `Relaxed` should be fine.

See [the relevant chapter](https://marabos.nl/atomics/atomics.html#example-racy-init) in Mara Bos' new book about atomics for details.

madsmtm (changes requested):
I think this might be more fitting to introduce into the `windows` crate instead, and then we'd use it from there? In any case, I feel that this is complexity that doesn't really gain us much (not that I'm against performance optimizations, but I suspect it won't really be measurable).

notgull:
> I think this might be more fitting to introduce into the `windows` crate instead, and then we'd use it from there? In any case, I feel that this is complexity that doesn't really gain us much (not that I'm against performance optimizations, but I suspect it won't really be measurable).

The main goal here is to avoid going through `OnceCell` for what is a simple numerical value that can easily be inclined into the atomic. That being said, I understand if the extra complexity isn't warranted, although I don't think it belongs in the `windows` crate.

madsmtm:
👍 I'll leave it up to @msiglreith to determine whether the extra complexity is warranted or not (I only marked this as "changes requested" because of the atomics comment)

src/platform_impl/windows/event_loop.rs
notgull: I replaced it with `AcqRel` and `Acquire`. I'd like to avoid a potential race condition with another event loop.

maroider (changes requested):
This is likely a small improvement, but it would likely be difficult to measure it.
It might also be worth considering renaming the type to `LazyMessageId` or `LazyMsgId`, since `LazyId` is a bit generic.

msiglreith (approved):
thanks, looks pretty compact now!
With the new release out, ready to merge this as well

PR #2599, opened 2022-12-20
Swap assert by debug_assert for recovereable issue. Fixes #2597

### LINKED ISSUE #2597: 'assertion failed: !event.matches()' hits erratically for some users. Should this actually be an assertion?
Hi everybody! 👋

in the app I'm working at, we have an erratic crash since long ago where this assert is being hit randomly, only for some users, upon resizing the browser window of a wasm browser game that uses winit and glutin. We still didn't manage to narrow it down or reproduce it consistently for all kinds of clients, but looking at the code where this assertion is called, it seems to me that it might not be needed.

This is how this error looks (for the sake of Google indexing, just in case someone else finds it):

```
panicked at 'assertion failed: !event.matches()', C:\Users\runneradmin\.cargo\registry\src\github.com-1ecc6299db9ec823\winit-0.27.5\src\platform_impl\web\web_sys\scaling.rs:74:9
```

or, in previous versions:

```
panicked at 'assertion failed: `(left == right)`
  left: `true`,
  right: `false`', C:\Users\runneradmin\.cargo\registry\src\github.com-1ecc6299db9ec823\winit-0.26.1\src\platform_impl\web\web_sys\scaling.rs:74:9
```

I pinged the original author, @alvinhochun, to ask about this ([here](https://github.com/rust-windowing/winit/pull/1690/files#r1053661524)) and it seems that, I quote:

> removing this assert should be perfectly fine because it is recoverable

So that's what I'm going to do. I plan on opening a PR for that if no one objects.

Does this sound familiar to anyone? Any further ideas are welcome 🙂

Maximetinu:
Fixes #2597

- [x] Tested on all platforms changed. **Web**. ✅ 
- [x] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users. **Not relevant enough to go into CHANGELOG.md imo.** ✅ 
- [x] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior. **No documentation to update** ✅
- [x] Created or updated an example program if it would help users understand this functionality **No need** ✅
- [x] Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented **No new features, it's a fix** ✅

### DIFF
--- a/src/platform_impl/web/web_sys/scaling.rs
+++ b/src/platform_impl/web/web_sys/scaling.rs
@@ -70,8 +70,7 @@ impl ScaleChangeDetectorInternal {
         mql
     }
 
-    fn handler(&mut self, event: MediaQueryListEvent) {
-        debug_assert!(!event.matches());
+    fn handler(&mut self, _event: MediaQueryListEvent) {
         let mql = self
             .mql
             .take()
--- a/src/platform_impl/web/web_sys/scaling.rs
+++ b/src/platform_impl/web/web_sys/scaling.rs
@@ -71,7 +71,7 @@ impl ScaleChangeDetectorInternal {
     }
 
     fn handler(&mut self, event: MediaQueryListEvent) {
-        assert!(!event.matches());
+        debug_assert!(!event.matches());
         let mql = self
             .mql
             .take()


madsmtm:
My best guess as to why this happens is that there's some sort of race-condition on the event registration - or maybe the scale-factor just changes enough such that the media query-state changes, while it is still within the specified bounds?

Idk, anyhow, I would rather just remove the assertion altogether

Maximetinu:
Should I remove it then? If so, then I can also remove the event parameter completely, as it won't be used anymore inside that function. What do you think?

madsmtm:
Yeah, go ahead and remove the assert and the parameter.

Maximetinu:
Done! ✅ 

I haven't finally removed the `event: MediaQueryListEvent `parameter, I've just marked it as _unused. It cannot be removed because the `listener` at the `MediaQueryListHandle` needs it to be used by other parts of the code, like the `canvas` to check if the browser is on dark mode or not

madsmtm:
That's fine, thanks!

Maximetinu:
Great, thanks! How often do you release? I'd love to have it as a 0.27.x hotfix, if possible 🙏

madsmtm:
See https://github.com/rust-windowing/winit/pull/2600 - I'd guess that 0.28 will land some time before February.

PR #2600, opened 2022-12-21
Release version 0.28.0

madsmtm:
The changelog is getting larger, and it was indirectly requested in https://github.com/rust-windowing/winit/pull/2157#issuecomment-1360370469 that we do a new release soon; so let's start the discussion of what that would entail.

See also [the milestone](https://github.com/rust-windowing/winit/milestone/8).

Tagging major library consumers, to ask what their timelines on releases are (so that we can somewhat coordinate on them, like, if you're planning a release on a specific date, we could try to release 0.28 shortly before that) (do tell me if I should tag someone else too in the future):
- `bevy`: @alice-i-cecile, @cart
- `egui`/`eframe`: @emilk
- `iced`: @hecrj
- `ggez`: @PSteinhaus

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,7 +45,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - **Breaking:** Removed `WindowBuilderExtWindows::with_theme` and `WindowBuilderExtWayland::with_wayland_csd_theme` in favour of `WindowBuilder::with_theme`.
 - **Breaking:** Removed `WindowExtWindows::theme` in favour of `Window::theme`.
 - Enabled `doc_auto_cfg` when generating docs on docs.rs for feature labels.
-- **Breaking:** On Android, switched to using [`android-activity`](https://github.com/rib/android-activity) crate as a glue layer instead of [`ndk-glue](https://github.com/rust-windowing/android-ndk-rs/tree/master/ndk-glue). See [README.md#Android](https://github.com/rust-windowing/winit#Android) for more details. ([#2444](https://github.com/rust-windowing/winit/pull/2444))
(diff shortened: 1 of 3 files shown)


alice-i-cecile:
Bevy's next release is scheduled for approximately the start of February.

kchibisov:
The way you've created branch is not correct, and you should create the branch. Consult the https://github.com/rust-windowing/winit/blob/master/CONTRIBUTING.md#release-process if you want to do a release, right now it seems like you've opened it against the master branch and has a weird history of commits. It should be just one comment at this point.

Regardless other issues, I'd probably like to solve more issues from the milestone, they are there for a reason. I don't think that there's a strong need for winit.

At least I'd suggest to go through the patches we have from @amrbashir since they are small and some of them like https://github.com/rust-windowing/winit/pull/2585 is a breaking change, since it could change the focused semantics.

As well as including RedoxOs support (cc @jackpot51).

I've updated the milestones and suggest to iterate through out the small patches at least.

I'd also like to bring Wayland fractional scaling patch, but it's hard for me to verify it given that most stuff wrt it is in works, so I'd like to wait for it for a bit, since it's a major change that is long awaited.

--

I would suggest to open issues instead of PRs if you want to discuss, but that's unrelated. Though, we can use it as an issue, given that you've opened PR against wrong branch and you'll have to recreate it anyway :p.

madsmtm:
I agree with everything you said, and yeah, let's continue with using this PR for discussion and then create a new one when we're ready.

I think I'd like to aim for sometime around mid January, so I guess I'll try to make a push for the things in the milestone in the start of that month.

nobbele:
>I think I'd like to aim for sometime around mid January, so I guess I'll try to make a push for the things in the milestone in the new year.

Is there an update to this ETA? The new release of ggez should be ready now.

kchibisov:
Not sure about ETA, but I'd start preparing release soon...

kchibisov:
Closing due to branch being opened against master.

PR #2605, opened 2022-12-26
Fix middle/other mouse buttons on macOS

RomanAkberov:
On macOS, `otherMouseDown:` and `otherMouseUp:` callbacks always result in a `MouseButton::Middle` event.

### DIFF
--- a/src/platform_impl/macos/view.rs
+++ b/src/platform_impl/macos/view.rs
@@ -688,42 +688,42 @@ declare_class!(
         fn mouse_down(&mut self, event: &NSEvent) {
             trace_scope!("mouseDown:");
             self.mouse_motion(event);
-            self.mouse_click(event, MouseButton::Left, ElementState::Pressed);
+            self.mouse_click(event, ElementState::Pressed);
         }
 
         #[sel(mouseUp:)]
         fn mouse_up(&mut self, event: &NSEvent) {
             trace_scope!("mouseUp:");
             self.mouse_motion(event);
-            self.mouse_click(event, MouseButton::Left, ElementState::Released);
+            self.mouse_click(event, ElementState::Released);
         }
 
         #[sel(rightMouseDown:)]
         fn right_mouse_down(&mut self, event: &NSEvent) {
             trace_scope!("rightMouseDown:");
             self.mouse_motion(event);
-            self.mouse_click(event, MouseButton::Right, ElementState::Pressed);
+            self.mouse_click(event, ElementState::Pressed);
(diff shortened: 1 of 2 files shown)


src/platform_impl/macos/view.rs
```diff
@@ -131,6 +131,15 @@ fn retrieve_keycode(event: &NSEvent) -> Option<VirtualKeyCode> {
     })
 }
 
+fn mouse_button(event: &NSEvent) -> MouseButton {
+    match event.buttonNumber() {
+        0 => MouseButton::Left,
+        1 => MouseButton::Right,
```
madsmtm, line 137: Hmm, would these ever happen?

src/platform_impl/macos/view.rs
```diff
@@ -131,6 +131,15 @@ fn retrieve_keycode(event: &NSEvent) -> Option<VirtualKeyCode> {
...
+        2 => MouseButton::Middle,
```
madsmtm, line 138: Is this constant documented anywhere?

madsmtm (changes requested):

madsmtm (changes requested):
Also needs a changelog entry IMHO ;)

src/platform_impl/macos/view.rs
RomanAkberov: Not with the NSOtherMouse... events, but the property is also set for the NSLeftMouse.../NSRightMouse... events. For the no-mouse-button event types, it's always 0, see https://developer.apple.com/documentation/appkit/nsevent/1527828-buttonnumber. Anyway, I think it's better to always convert the value here than to panic or return an Option.

src/platform_impl/macos/view.rs
RomanAkberov: Not in the AppKit docs AFAIK. AppKit seems to completely ignore the relevance of the buttons _other_ than left/right. The closest thing in the Apple docs I was able to find is this constant in the CG/Quartz docs: https://developer.apple.com/documentation/coregraphics/cgmousebutton/kcgmousebuttoncenter. 
At least this is how both glfw and sdl handle it. And it works for my MX Vertical: middle = 2, back = 3, forward = 4.

RomanAkberov:
I've added a changelog entry and some comments btw.

kchibisov (approved):

kchibisov (approved):
I've slightly cleaned up the code. Should be good now.

Thanks!

PR #2610, opened 2022-12-29
Fix `RedrawRequested` not emitted on Wayland in resize

### LINKED ISSUE #2609: wayland: `RedrawRequested` isn't called while resizing window
On Wayland a `RedrawRequested` event isn't emitted until resizing ends. If an application is only drawing (for instance, with softbuffer) in response to `RedrawRequested` events, this is a problem, since the window doesn't visually change in size at all until we commit a buffer with the new size.

This seems to differ from the behavior on X or Windows, where the window does redraw during a resize.

I see `sctk::window::Event::Configure` is being produced throughout the resize, but this doesn't end up producing a `RedrawRequested` event from winit. Not sure exactly what should be changed for this to behave appropriately.

I've been testing this with https://github.com/rust-windowing/softbuffer/blob/master/examples/winit.rs, but removing the `request_redraw()` call for resize events, which I added there to make the example work properly on Wayland.

kchibisov:
Fixes #2609.


ids1024 (approved):
This results in smooth resizing with the softbuffer code I was seeing an issue with.

I was thinking of a similar fix but wasn't so sure where it should set `redraw_requested`.

PR #2612, opened 2022-12-31
Add `Window::set_transparent`

kchibisov:
Provide a hint to system compositor whether the window is opaque or not. Only implemented on macOS and Wayland for now.

### DIFF
--- a/src/platform_impl/linux/wayland/window/mod.rs
+++ b/src/platform_impl/linux/wayland/window/mod.rs
@@ -261,6 +261,9 @@ impl Window {
             window_requests.clone(),
         );
 
+        // Set opaque region.
+        window_handle.set_transparent(attributes.transparent);
+
         // Set resizable state, so we can determine how to handle `Window::set_inner_size`.
         window_handle.is_resizable.set(attributes.resizable);
 
@@ -330,6 +333,11 @@ impl Window {
     #[inline]
     pub fn set_title(&self, title: &str) {
         self.send_request(WindowRequest::Title(title.to_owned()));
+    }
+
+    #[inline]
+    pub fn set_transparent(&self, transparent: bool) {
+        self.send_request(WindowRequest::Transparent(transparent));
     }
 
     #[inline]
--- a/src/window.rs
+++ b/src/window.rs
@@ -306,7 +306,9 @@ impl WindowBuilder {
     /// Sets whether the background of the window should be transparent.
     ///
     /// If this is `true`, writing colors with alpha values different than
-    /// `1.0` will produce a transparent window.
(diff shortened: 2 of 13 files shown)


src/window.rs
```diff
@@ -735,6 +735,23 @@ impl Window {
...
     }
 
+    /// Change the window opacity state.
+    ///
+    /// This is just a hint that may not change anything about
+    /// the window transparency, however doing a missmatch between
+    /// the content of your window and this hint may result in
+    /// visual artifacts.
+    ///
+    /// The default value follows [`WindowBuilder::with_transparent`].
```
madsmtm, line 745: Perhaps update `with_transparent` to point to this documentation instead?

src/platform_impl/linux/wayland/window/mod.rs
```diff
@@ -260,6 +260,9 @@ impl Window {
             window_requests.clone(),
         );
 
+        // Set opaque region.
+        window_handle.set_opaque(!attributes.transparent);
+
```
madsmtm, line 265: Add changelog entry about this + add it to the list of platforms `WindowBuilder::with_transparent` supports

madsmtm (approved):
I think I would prefer `set_transparent`, to keep in line with `with_transparent` (also, "transparent" is immediately understandable to me as a non-native speaker, which "opaque" is, well, more opaque ;) ).

But just a preference, otherwise approved from my side.

src/platform_impl/linux/wayland/window/mod.rs
kchibisov: Do you mean that opaque region is now set on Wayland? I'm not quite sure that it should be stated in a changelog tbh...

Or do you mean something else? Like it shouldn't fix or affect the users in any way, since it's just a hint and the window was considered transparent.

kchibisov:
>think I would prefer set_transparent, to keep in line with with_transparent (also, "transparent" is immediately understandable to me as a non-native speaker, which "opaque" is, well, more opaque ;) ).

I actually like `opaque` more, since it's more common/discoverable then `transparent`. Most systems refer to it as opaque anyway.

kchibisov:
I've decided to name it `set_transparent` otherwise it could indeed be a bit confusing for the end users.

src/platform_impl/linux/wayland/window/mod.rs
madsmtm: If it's just a hint then no problem

kchibisov: Yeah, most apps don't even bother to set it.

PR #2614, opened 2023-01-01
internal: Add libxcb and replace some Xlib functions

notgull:
- [x] Tested on all platforms changed.

This PR replaces portions of the X11 backend. Rather than using `libX11`, which is an outdated and buggy library, it uses `libxcb`, though the `x11rb` crate, to communicate with the X server. This PR significantly reduces the amount of unsafe code found in the backend, For some reason `cargo geiger` crashes on this codebase, so I can't accurately measure the change, but I can guarantee it's significantly lower. In fact, almost all of the unsafe code in the X11 backend has been removed.

I'm aware that prior work on this has been done in #1932 and #2088, but at this point I think it's best to start from scratch.

IME appeared to be broken after I started integration of `x11rb` (as if it wasn't broken anyways). I've deleted it completely for now. We may want to use [`xcb-imdkit`](https://github.com/fcitx/xcb-imdkit) or [`xim-rs`](https://crates.io/crates/xim) to replace it.

Blocked on #2662 for now


ids1024:
Running in a NetBSD VM I see "failed to open library "libxcb.so.1".

Looks like the library is at `/usr/X11R7/lib/libxcb.so.2`.

kchibisov:
It would be nice to make the GLX/EGL work as it was before, and also ensure that it doesn't break other stuff like wgpu/vulkan. Since I think what we have is a mix of xlib and xcb.

What is the situation with error handling here as in which library delivers the errors(Xlib?)?

I'm not that familiar with X11 though.

notgull:
> It would be nice to make the GLX/EGL work as it was before, and also ensure that it doesn't break other stuff like wgpu/vulkan. Since I think what we have is a mix of xlib and xcb.

We should be able to just keep the `Display` pointer around and still pass that on through `as_raw_display_handle`, so nothing should break unless we want it to.

> What is the situation with error handling here as in which library delivers the errors(Xlib?)?

I still keep the `Xlib` error handling infrastructure around. However, `x11rb` translates `libxcb` errors to higher-level errors (see [`ConnectionError`](https://docs.rs/x11rb/latest/x11rb/errors/enum.ConnectionError.html)). When I receive these errors, I usually panic on them, because that's what the backend did before.

notgull:
Okay, I tested this code on:

* Linux
* OpenBSD
* FreeBSD
* DragonflyBSD

NetBSD appears to suffer from an `x11-dl` bug (AltF02/x11-rs#166), and Illumos fails to build (I'll PR that later).

notgull:
What is the current status of this PR? Are there any blockers that I can resolve?

kchibisov:
I need to find time and read through it. Also it should have feature parity with the current master. Have you resolved the IME issues?

notgull:
Not yet. I'll go ahead and add that the next time I have time.

notgull:
Putting this back into the draft phase as this is now blocked on the next `xim-rs` release. Also I should probably implement XKB support.

kchibisov:
We already have XKB support on the keyboard v2 (or new-keyboard) branch. In general you can go without XKB and we can merge it later on, unless you really want xkb now for whatever reason. Some of the XKB stuff is present in wayland backend already for example.

notgull:
> In general you can go without XKB and we can merge it later on, unless you really want xkb now for whatever reason.

Alright, if you say so. I'll replace it with a naïve algorithm for now.

madsmtm:
FYI we just got https://github.com/rust-windowing/winit/pull/2662, which I think we should focus on first - so you might have to do redo some work once that lands @notgull

notgull:
Don't worry, I've been continuously rebasing this branch off of master anyways. As long as it uses `xkbcommon` instead of Xlib's opaque solution it should be easy to integrate anyways.

notgull:
Actually, I'd rather just wait for #2662 to be merged anyhow, since it seems this PR is blocked on that one.

notgull:
Supersedes by #2825

PR #2615, opened 2023-01-03
Rework DnD

amrbashir:
  - [X] Windows
  - [ ] macOS
  - [x] X11


kchibisov:
Could you elaborate why this is needed? Don't you get pointer motion events during DnD? It's just strange that we have 2 sources of pointer events during dnd with that.

kchibisov:
This is also a breaking change.

amrbashir:
> Could you elaborate why this is needed? Don't you get pointer motion events during DnD? It's just strange that we have 2 sources of pointer events during dnd with that.

You get the cursor event only right before the `DroppedFile` event happens but not while `HoveredFile`, also there has been reports that it sometimes fire after the `DroppedFile` (Can't reproduce but see https://github.com/rust-windowing/winit/issues/1550#issuecomment-798955588). Adding the cursor position to these events removes the need to sync with other events.

src/event.rs
```diff
@@ -337,13 +337,21 @@ pub enum WindowEvent<'a> {
     ///
     /// When the user drops multiple files at once, this event will be emitted for each file
     /// separately.
-    DroppedFile(PathBuf),
+    DroppedFile {
+        path: PathBuf,
+        // (x,y) coords in pixels relative to the top-left corner of the window.
```
kchibisov, line 342:
```suggestion
        /// The position of the mouse cursor, similar to [`WindowEvent::CursorMoved`].
```

src/event.rs
```diff
@@ -337,13 +337,21 @@ pub enum WindowEvent<'a> {
...
+    },
 
     /// A file is being hovered over the window.
     ///
     /// When the user hovers multiple files at once, this event will be emitted for each file
     /// separately.
-    HoveredFile(PathBuf),
+    HoveredFile {
+        path: PathBuf,
+        // (x,y) coords in pixels relative to the top-left corner of the window.
```
kchibisov, line 352:
```suggestion
        /// The position of the mouse cursor, similar to [`WindowEvent::CursorMoved`].
```

kchibisov (changes requested):
Just to ensure, have you tested the X11 part of it?

amrbashir:
I haven't tested X11 nor macOS but I can test X11 later today.

amrbashir:
Alright, X11 should be fine now. The returned coordinates were root window coordinates so I had to translate them before dispatching the event.

src/platform_impl/linux/x11/util/geometry.rs
```diff
@@ -156,8 +156,34 @@ impl XConnection {
...
+        unsafe {
+            (self.xlib.XTranslateCoordinates)(
+                self.display,
+                src_w,
+                dest_w,
+                src_x,
+                src_y,
+                &mut coords.x_rel,
+                &mut coords.y_rel,
                 &mut coords.child,
```
kchibisov, line 187: I'm not entirely sure why you need a second method here, given that the only difference is the `(0, 0)`? Though maybe it's fine to have it that way.

amrbashir: the original method was mapping window -> root while I need the opposite (root -> window)

amrbashir: I could keep the new one (because it is more flexible) and remove the old one and migrate its usages to the new one.

kchibisov: I mean that you path both Root and other window, so you can swap them? Would be the other way around.

amrbashir: Sorry, I am not sure I understand what you mean.

kchibisov: I mean that your new call supersedes the existing one, since with it you can express both translations just fine. The old one wasn't working due to inability to pass `src_x` and `src_y` and hardcoded it to zero.

But I guess it's fine the way it is since you don't have remember that you need `(0, 0)` for one of the translations.

src/platform_impl/macos/window_delegate.rs
```diff
@@ -172,9 +172,13 @@ declare_class!(
             let filenames = pb.propertyListForType(unsafe { NSFilenamesPboardType });
             let filenames: Id<NSArray<NSString>, Shared> = unsafe { Id::cast(filenames) };
 
+            let dl: NSPoint = unsafe { msg_send![sender, draggingLocation] };
+            let scale_factor = self.window.scale_factor();
+            let position = LogicalPosition::<f64>::from((dl.x, dl.y)).to_physical(scale_factor);
```
madsmtm, line 177: We're missing a conversion to `winit`'s coordinate system, which has origin in the top-left (e.g. use `util::bottom_left_to_top_left`).

Same below.

madsmtm (changes requested):
I agree that it makes sense to have for `DroppedFile`, but not so sure about `HoveredFile`? It seems like, for that to be useful, you'd have to receive the event continuously (which you at least don't on macOS)

kchibisov:
>but not so sure about HoveredFile? It seems like, for that to be useful, you'd have to receive the event continuously (which you at least don't on macOS)

That would strange, I'd assume you can know that you're over some element while doing drag and drop? Maybe the event is sent via some side channel, so we should plumb it to hovered?

amrbashir:
> but not so sure about HoveredFile? It seems like, for that to be useful, you'd have to receive the event continuously (which you at least don't on macOS)

It is the same behavior on Windows and Linux too, the event is emitted only once upon entering the window.

src/platform_impl/macos/window_delegate.rs
```diff
@@ -172,9 +172,18 @@ declare_class!(
...
             let filenames: Id<NSArray<NSString>, Shared> = unsafe { Id::cast(filenames) };
 
+            let dl: NSPoint = unsafe { msg_send![sender, draggingLocation] };
+            let rect = NSRect {
+                origin: dl,
+                ..Default::default()
+            };
+            let y = util::bottom_left_to_top_left(rect);
+            let scale_factor = self.window.scale_factor();
+            let position = LogicalPosition::<f64>::from((dl.x, y)).to_physical(scale_factor);
```
madsmtm, line 179: Hmm, nope, turns out the `y` coordinate was still wrong. I've pushed a proper fix for this now.

amrbashir: thanks

madsmtm (approved):
macOS impl tested and approved from my side.

> It is the same behavior on Windows and Linux too, the event is emitted only once upon entering the window.

`HoveredFile.position` still doesn't really make sense to me, perhaps we should instead just emit two events `CursorMoved { position } + HoveredFile { path }` (or the other way around, if that's better)?

> Adding the cursor position to these events removes the need to sync with other events.

(Though I do believe you when you say it's useful, I don't use the API myself, so it's not a blocker for me).

amrbashir:
> > It is the same behavior on Windows and Linux too, the event is emitted only once upon entering the window.
> 
> `HoveredFile.position` still doesn't really make sense to me, perhaps we should instead just emit two events `CursorMoved { position } + HoveredFile { path }` (or the other way around, if that's better)?

I agree, when I first encountered the event, I thought it would be triggered continuously until the file is dropped, cancelled or moves out of the window.

I think a better proposal would be to introduce new events, `DragEnter`, `DragHover`, `DragDrop`, `DragCancel` and/or maybe `DragLeave`? On Windows (and on macOS too, I think,) `DragLeave` and `DragCancel` are the same thing, so maybe only One of them should be added.

kchibisov:
>I think a better proposal would be to introduce new events, DragEnter, DragHover, DragDrop, DragCancel and/or maybe DragLeave? On Windows (and on macOS too, I think,) DragLeave and DragCancel are the same thing, so maybe only One of them should be added.

`DragEnter`, `DragHover`, `DragDrop`, and `DrapLeave` sounds good to me. It'll match the Wayland pretty match.

However, I think we might want to send CursorMotion events instead of `DragHover` and remove it? What's even the purpose of it anyway?

@madsmtm do you get any cursor events when moving drag and drop over the window, don't you get CurserMoved?

>(Though I do believe you when you say it's useful, I don't use the API myself, so it's not a blocker for me).

I think what we should do is to plumb the current mouse position inside the `DnD` and suppress the mouse/touch events during dnd? That way the API for it will be unique among the input backend, since both touch and mouse could start DND.

Though, it could be hard to plumb the mouse/touch position inside the dnd, though you can broadcast on such platforms that you're doing DND and emit extra bits...

amrbashir:
> However, I think we might want to send `CursorMotion` events instead of `DragHover` and remove it? What's even the purpose of it anyway?

using `DragHover` you can get the current hover position and then check if it is inside a drop zone or not, if it is, you may highlight the drop zone so the user have a visual indication where to drop the file.

kchibisov:
>using DragHover you can get the current hover position and then check if it is inside a drop zone or not, if it is, you may highlight the drop zone so the user have a visual indication where to drop the file.

Ah, so it pairs with some API to set DND zones? On Wayland most of the time you should be fine with `CursorMoved` and `DragEnter`, so you know that you have DND with the cursor and can highlight drop areas yourself, but you don't define them on a display server level, it's up to you how to handle them.

The thing is that winit should send either CursorMoved or Hovered event, it shouldn't send both at the same time.

amrbashir:
> The thing is that winit should send either CursorMoved or Hovered event, it shouldn't send both at the same time.

why not both? if the OS sends them both, I think winit should pass that through as well

madsmtm:
> do you get any cursor events when moving drag and drop over the window, don't you get CurserMoved?

Nope, you get neither `WindowEvent::CursorMoved` nor `DeviceEvent::MouseMotion`.

But it seems you can [opt-in](https://developer.apple.com/documentation/appkit/nsdraggingdestination/1416049-wantsperiodicdraggingupdates?language=objc) to getting updates about the current drag operation, so one could probably implement a `DragHover` event on macOS.

---

Should we consider postponing this PR to 0.29? It seems like there's still quite some work to do figuring out the desired design (and implementing it, and testing it).

kchibisov:
>why not both? if the OS sends them both, I think winit should pass that through as well

I guess using `both` will also work, I'm just afraid of the potential collisions and such, since DND mouse movements shouldn't update mouse cursor. But I guess on most systems the event data is separate already.

I'm not sure I like `Hovered` as a name, maybe `DragMoved` will be better.

kchibisov:
>Should we consider postponing this PR to 0.29? It seems like there's still quite some work to do figuring out the desired design (and implementing it, and testing it).

We may do so if it'll be the only PR left.

amrbashir:
> I'm not sure I like `Hovered` as a name, maybe `DragMoved` will be better.

I went with `DragOver` since it could be triggered even if the mouse hasn't moved.

dhardy:
Use of `CursorMotion` sounds problematic: what happens if the event source is touch not mouse?

> Ah, so it pairs with some API to set DND zones?

I much prefer the current API (just send the coordinates).

---

The current API seems fine except maybe one thing: it clones a `Vec<PathBuf>` on each cursor motion. Is this sensible over, maybe `dyn AsRef<Path>` or some such?

Also the API should be clear on whether `CursorMotion`, `Touch` etc. may still be received during DnD.

amrbashir:
> The current API seems fine except maybe one thing: it clones a `Vec<PathBuf>` on each cursor motion. Is this sensible over, maybe `dyn AsRef<Path>` or some such?

Using `dyn AsRef<Path>` would require manual implementation for the `Debug` trait and would need to either remove the `PartialEq` derive or implement it manually so I don't think it is worth it.

I initially had `DragOver` only emit the drag location without paths, since `DragEnter` emits the paths and it will be re emitted on `DragDrop` but I added just to show it is possible and if it is desired behavior or not.

dhardy:
> Using dyn AsRef<Path> would require manual implementation ...

Rust doesn't support multi-trait objects (yet) but you can do this:
```rust
pub trait HasPath: Debug + PartialEq + Eq + AsRef<Path> {}
```
But this is somewhat irrelevant in my opinion. Do we want to lock winit into an API which copies a bunch of paths on every mouse movement? Or is it irrelevant since all platforms re-send those on each movement anyway?

amrbashir:
> Do we want to lock winit into an API which copies a bunch of paths on every mouse movement? Or is it irrelevant since all platforms re-send those on each movement anyway?

Only Windows doesn't send it on each movement, x11 and macOS do.

madsmtm:
I think I'd prefer to just leave paths out of `DragOver` - if you need to do something with the paths while they're being dragged you can just store it yourself from `DragEnter`.

And maybe even remove it from `DragDrop`? Can the dragged items change while they're being dragged? (Or should we leave it in for ease-of-use?)

amrbashir:
> And maybe even remove it from `DragDrop`? Can the dragged items change while they're being dragged? (Or should we leave it in for ease-of-use?)

I doubt that it can change but I'd say keep it for ease-of-use

amrbashir:
@kchibisov It is been a while, and now that I have some time to work on this and fix conflicts, do you think this rework should land soon?

geom3trik:
Is this PR still compatible with the recent winit API changes or does it need updating?

kchibisov:
You can open a PR with the new proposed events, with impl only for one backend and see how it works with the recent pointer API rework.

I myself don't have issues with the suggested design, since it's mostly about dropping into _your window_, starting drag is not really possible from winit and it'll require the same work as e.g. clipboard APIs. Though, it's not that much.

amrbashir:
I will see If I can update this PR once again when I have some time. 


For starting a drag operation, there is `drag-rs` which can handle that, see https://github.com/crabnebula-dev/drag-rs/blob/v2/examples/winit/src/main.rs for an example with winit

jomala:
This is so close! Can we close it out, @amrbashir? It looks like at lesat half-a-dozen projects would be grateful!

valadaptive:
I rebased this PR at #4079 (lost the commit history because untangling all the merge commits was infeasible).

Once I fix the X11 implementation, I can test it out across the implemented platforms.

PR #2616, opened 2023-01-03
m: Remove xlib_xconnection from public interface

notgull:

This PR removes the `xlib_xconnection()` function from the `platform::x11` module. This was previously used by `glutin`, but is not used anymore. Removing this lets me make some types in the `platform_impl::x11` module `pub(crate)`, which makes it easier to identify what's being used and what isn't.


kchibisov:
Should add a breaking changelog entry for that, since it was publicly exposed in the end. Other than that looks fine.

kchibisov (approved):

PR #2619, opened 2023-01-05
chore: Alphabetize dependencies

notgull:

Makes it so `cargo add` works properly, and also looks nicer.


madsmtm (approved):
Seems like an issue with `cargo add` if it can't handle it otherwise, but sure, let's do it

PR #2625, opened 2023-01-10
Fix `Window::set_minimized(false)` when window was minimized by `Win+D`

amrbashir:

### DIFF
--- a/src/platform_impl/windows/window.rs
+++ b/src/platform_impl/windows/window.rs
@@ -40,7 +40,7 @@ use windows_sys::Win32::{
         },
         WindowsAndMessaging::{
             CreateWindowExW, FlashWindowEx, GetClientRect, GetCursorPos, GetForegroundWindow,
-            GetSystemMetrics, GetWindowPlacement, GetWindowTextLengthW, GetWindowTextW,
+            GetSystemMetrics, GetWindowPlacement, GetWindowTextLengthW, GetWindowTextW, IsIconic,
             IsWindowVisible, LoadCursorW, PeekMessageW, PostMessageW, RegisterClassExW, SetCursor,
             SetCursorPos, SetForegroundWindow, SetWindowDisplayAffinity, SetWindowPlacement,
             SetWindowPos, SetWindowTextW, CS_HREDRAW, CS_VREDRAW, CW_USEDEFAULT, FLASHWINFO,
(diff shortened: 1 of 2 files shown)


src/platform_impl/windows/window.rs
```diff
@@ -455,8 +455,13 @@ impl Window {
...
+        let is_minimized = self.is_minimized();
+
         self.thread_executor.execute_in_thread(move || {
             let _ = &window;
+            WindowState::set_window_flags_in_place(&mut window_state.lock().unwrap(), |f| {
+                f.set(WindowFlags::MINIMIZED, is_minimized)
+            });
             WindowState::set_window_flags(window_state.lock().unwrap(), window.0, |f| {
                 f.set(WindowFlags::MINIMIZED, minimized)
             });
```
kchibisov, line 467: The code setting the same flag to some bool looks a bit odd. Could you elaborate why 2 calls are in the row? can't you somehow do a singe call based on both minimized and is_minimized? 

If anything, maybe it should be updated from the `event_loop` somehow? I guess the issue is that you failed to update the state from the event_loop code around the minimize, so you try to fix-up the state when asked to changed it. That just looks error prone, once anyone wanted to for-example broadcast the window state changed, since you poll for state changed...

kchibisov (changes requested):

src/platform_impl/windows/window.rs
amrbashir: My bad, I forgot to explain the bug.

Basically, when using `Win + D` hotkey, all windows get minimized (probably because the `Win + D` hotkey is registered by `explorer.exe` and not an OS hotkey) but the relevant win32 Window Message is not sent to these windows (again because of `explorer.exe`) so the state gets out of sync, this is mostly fine, however, one can't call `set_minimized(false)` because the state doesn't know the window is minimized and the diffing wouldn't recognize the change in `MINIMIZED` flag, so the only solution I found was to query the OS for the minimized state and set it before diffing.

kchibisov: So it means that you can't rely on system at all? Meaning that you should poll in `is_minimized`/`set_minimized`/focus changes?

Could that happen to other states, like maximized/full-screen? What happens if you minimize from full-screen/maximize with `Win+D`?

amrbashir: > So it means that you can't rely on system at all? Meaning that you should poll in `is_minimized`/`set_minimized`/focus changes?

Kinda, other apps could minimize your app windows which is the case here. I suspect `explorer.exe` or another app within Windows file registers `Win + D` shortcut and then minimizes all the windows programmatically.

> Could that happen to other states, like maximized/full-screen?

I think this only affects the minimized state since `Win + D` is a common shortcut for Windows, However, as I explained above, other apps could manipulate your windows at any time.

> What happens if you minimize from full-screen/maximize with `Win+D`?

Haven't tested but probably the same behavior.

kchibisov: >Haven't tested but probably the same behavior.

I mean that maximize state will desync that way as well.

amrbashir: As far as I tested, maximize state is unaffected. It will still report `false` when the window is minimized using `Win + D` and will report `true` when it is restored.

kchibisov (approved):

PR #2626, opened 2023-01-11
Add `Window::is_minimized`

amrbashir:
  - [X] Windows
  - [x] macOS 
  - [ ] x11

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
-- Add `Window::is_minimized`
+- Add `Window::is_minimized`.
 - On X11, fix errors handled during `register_xlib_error_hook` invocation bleeding into winit.
 - Add `Window::has_focus`.
 - On Windows, fix `Window::set_minimized(false)` not working for windows minimized by `Win + D` hotkey.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- Add `Window::is_minimized`
 - **Breaking:** On Web, touch input no longer fires `WindowEvent::Cursor*`, `WindowEvent::MouseInput`, or `DeviceEvent::MouseMotion` like other platforms, but instead it fires `WindowEvent::Touch`.
 - **Breaking:** Removed platform specific `WindowBuilder::with_parent` API in favor of `WindowBuilder::with_parent_window`.
 - On Windows, retain `WS_MAXIMIZE` window style when un-minimizing a maximized window.
--- a/src/platform_impl/linux/x11/window.rs
+++ b/src/platform_impl/linux/x11/window.rs
@@ -801,7 +801,7 @@ impl UnownedWindow {
     }
 
     #[inline]
-    pub fn is_minimized(&self) -> bool {
+    pub fn is_minimized(&self) -> Option<bool> {
         let state_atom = unsafe { self.xconn.get_atom_unchecked(b"_NET_WM_STATE\0") };
         let state = self
             .xconn
--- a/src/platform_impl/linux/x11/window.rs
+++ b/src/platform_impl/linux/x11/window.rs
@@ -809,10 +809,7 @@ impl UnownedWindow {
         let hidden_atom = unsafe { self.xconn.get_atom_unchecked(b"_NET_WM_STATE_HIDDEN\0") };
 
         Some(match state {
-            Ok(atoms) => {
-                let hidden = atoms.iter().any(|atom: &ffi::Atom| *atom == hidden_atom);
-                hidden
-            }
+            Ok(atoms) => atoms.iter().any(|atom: &ffi::Atom| *atom == hidden_atom),
             _ => false,
         })
     }
--- a/src/platform_impl/linux/x11/window.rs
+++ b/src/platform_impl/linux/x11/window.rs
@@ -800,6 +800,23 @@ impl UnownedWindow {
         self.xconn.primary_monitor()
     }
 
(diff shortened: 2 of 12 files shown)


src/window.rs
```diff
@@ -822,6 +822,16 @@ impl Window {
         self.window.set_minimized(minimized);
     }
 
+    /// Gets the window's current minimized state.
+    ///
```
kchibisov, line 826: Could you add a note that it shouldn't stop rendering for minimized windows, however you could lower fps for minimize windows.

Not sure how to phrase it better, but you've got an idea.

src/window.rs
```diff
@@ -822,6 +822,16 @@ impl Window {
...
+    /// ## Platform-specific
```
kchibisov, line 827: While I think that `minimize` makes sense, could you return `Option<bool>` instead? On Wayland for example you can request minimize, but you don't know the state, whether you're minimized or not.

The impl for that on Wayland will be plain `None` return.

On other platforms you could also return `None` while they are un-implemented, meaning that you can't determine whether the window is minimized or not.

kchibisov (changes requested):
The change is ok. But would need to return `Option` given that some platforms don't provide a way to know whether the window is minimized or not.

amrbashir:
> The change is ok. But would need to return `Option` given that some platforms don't provide a way to know whether the window is minimized or not.

I am fine with that but I should note that there is other similar functions that return only bools, like `is_maximized` and `is_resizable` while they don't support all platforms and I think it should be consistent.

kchibisov:
It's not about support though. It's about 'you can't possibly know' that you're minimized or not.

With maximized/resizable you surely know such a thing.

amrbashir:
yeah, makes a bit of sense.

src/window.rs
```diff
@@ -822,6 +822,20 @@ impl Window {
         self.window.set_minimized(minimized);
     }
 
+    /// Gets the window's current minimized state.
+    ///
+    /// ## Note
+    ///
+    /// You shouldn't stop rendering for minimized windows, however you could lower the fps.
```
kchibisov, line 829: Could you say what `None` means?

src/window.rs
```diff
@@ -822,6 +822,20 @@ impl Window {
...
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **iOS / Android / Web / Wayland / Orbital:** Unsupported.
```
kchibisov, line 833: Could you say that on `Wayland` it's always `None`.

src/platform_impl/web/window.rs
```diff
@@ -277,6 +277,12 @@ impl Window {
         // Intentionally a no-op, as canvases cannot be 'minimized'
     }
 
+    #[inline]
+    pub fn is_minimized(&self) -> Option<bool> {
+        // Intentionally a no-op, as canvases cannot be 'minimized'
+        None
```
kchibisov, line 283: It means that it's always `Some(false)`? The comment about a `no-op` doesn't make any sense.

src/platform_impl/linux/x11/window.rs
```diff
@@ -800,6 +800,23 @@ impl UnownedWindow {
...
+        let state_atom = unsafe { self.xconn.get_atom_unchecked(b"_NET_WM_STATE\0") };
+        let state = self
+            .xconn
+            .get_property(self.xwindow, state_atom, ffi::XA_ATOM);
+        let hidden_atom = unsafe { self.xconn.get_atom_unchecked(b"_NET_WM_STATE_HIDDEN\0") };
+
+        Some(match state {
+            Ok(atoms) => {
+                let hidden = atoms.iter().any(|atom: &ffi::Atom| *atom == hidden_atom);
+                hidden
```
kchibisov, line 814: You don't need tmp variable here.

kchibisov (changes requested):

src/window.rs
```diff
@@ -841,6 +841,21 @@ impl Window {
         self.window.set_minimized(minimized);
     }
 
+    /// Gets the window's current minimized state.
+    /// `None` will be returned, if the minimized state couldn't be determined.
+    /// ## Note
```
kchibisov, line 846:
```suggestion
    /// Gets the window's current minimized state.
    ///
    /// `None` will be returned, if the minimized state couldn't be determined.
    ///
    /// ## Note
```

src/platform_impl/windows/window.rs
```diff
@@ -475,6 +475,11 @@ impl Window {
         });
     }
 
+    #[inline]
+    pub fn is_minimized(&self) -> Option<bool> {
+        Some(unsafe { IsIconic(self.hwnd()) == 1 })
```
msiglreith, line 480: Closer to the official docs: `If the window is iconic, the return value is nonzero.`
```suggestion
        Some(unsafe { IsIconic(self.hwnd()) != false.into() })
```

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- Add `Window::is_minimized`
```
```suggestion
- Add `Window::is_minimized`.
```

src/platform_impl/windows/window.rs
msiglreith: I'm not too happy with the additional `unwraps` required due to returning an `Option` as it's not immediately clear that it always will be `Some`. Maybe a `is_minimized :: HWND -> bool` variant in `util` and the function above just returns `Some(util::is_minimized(self.hwnd()))`. Internally we would use the `util.:is_minimized` to avoid the `unwrap` calls

amrbashir: yeah sounds better.

madsmtm (approved):
Yeah, looks fine on macOS

kchibisov (approved):

PR #2632, opened 2023-01-14
Properly remove window mouse event listeners

DouglasDwyer:
- [ ] [N/A] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior
- [ ] [N/A] Created or updated an example program if it would help users understand this functionality
- [ ] [N/A] Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented

The web backend for `winit` registers the `onmousedown`, `onmouseup`, and `onmousemove` events on the global Javascript window object. When the `winit` window is destroyed, however, these events are not properly removed. This is because the events are registered with the `capture: true`, but `capture: true` is not passed to the appropriate `removeEventListener` call. This PR updates the web backend to pass the event creation options to the event removal function. This prevents errors in which the Javascript runtime tries to invoke the event listener closures after they are destroyed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - **Breaking:** Remove the unstable `xlib_xconnection()` function from the private interface.
 - Added Orbital support for Redox OS
 - On X11, added `drag_resize_window` method.
+- On web, fix removal of mouse event listeners from the global object upon window distruction.
 
 # 0.27.5
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -129,6 +129,7 @@ features = [
     'DomRect',
     'Element',
     'Event',
+    "EventListenerOptions",
     'EventTarget',
     'FocusEvent',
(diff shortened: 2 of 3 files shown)


maroider:
It looks like you're right about that, but I'm stumped as to how you managed to observe it in the first place.

DouglasDwyer:
@maroider Absolutely! This can be reproduced by creating a Winit window on the web platform, and then dropping it, while allowing the event loop to continue. When you wave your cursor over the browser window after this happens, errors about a destroyed closure should appear in the console.

madsmtm (approved):
I'm fine with merging this now unless @maroider wants to test it first?

maroider:
I had a go at following your instructions, but I couldn't get the error to show up in the console. It's probably fine to merge it as-is, even if I couldn't reproduce the issue, as the MDN docs clearly show we're using `removeEventListener` wrong anyway.

madsmtm:
Thank you @DouglasDwyer!

PR #2638, opened 2023-01-20
[MacOS] Replace NSTrackingRect with NSTrackingArea

The-Minecraft-Scientist:
This PR updates winit's macOS implementation to use NSTrackingArea instead of TrackingRects. These have [more features](https://developer.apple.com/documentation/appkit/nstrackingareaoptions?language=objc) than TrackingRects, which could be used to implement some of the cursor-related features that aren't currently implemented for macOS. 

I'm pretty inexperienced when it comes to working with big codebases, and am happy to hear any advice/requested changes.

- [ X] Tested on all platforms changed
(no user-facing changes)

### DIFF
--- a/src/platform_impl/macos/view.rs
+++ b/src/platform_impl/macos/view.rs
@@ -12,9 +12,9 @@ use objc2::runtime::{Object, Sel};
 use objc2::{class, declare_class, msg_send, msg_send_id, sel, ClassType};
 
 use super::appkit::{
-    NSApp, NSCursor, NSEvent, NSEventModifierFlags, NSEventPhase, NSResponder, NSTrackingRectTag,
-    NSView,
+    NSApp, NSCursor, NSEvent, NSEventModifierFlags, NSEventPhase, NSResponder, NSView,
 };
+use crate::platform_impl::platform::appkit::{NSTrackingArea, NSTrackingAreaOptions};
 use crate::{
     dpi::{LogicalPosition, LogicalSize},
     event::{
@@ -33,7 +33,6 @@ use crate::{
     },
     window::WindowId,
 };
-use crate::platform_impl::platform::appkit::NSTrackingArea;
 
 #[derive(Debug)]
 pub struct CursorState {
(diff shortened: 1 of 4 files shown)


src/platform_impl/macos/view.rs
```diff
@@ -199,25 +199,34 @@ declare_class!(
...
             let rect = self.visibleRect();
-            let tracking_rect = self.add_tracking_rect(rect, false);
-            self.state.tracking_rect = Some(tracking_rect);
+            let tracking_area = self.init_and_add_tracking_area(
+                NSTrackingAreaOptions::NSTrackingMouseEnteredAndExited
+                    | NSTrackingAreaOptions::NSTrackingMouseMoved
+                    | NSTrackingAreaOptions::NSTrackingActiveAlways,
+                rect,
+            );
+            self.state.tracking_area = Some(tracking_area);
```
madsmtm, line 229: Hmm, it sounds like in the documentation that we should instead do this in [`-updateTrackingAreas`](https://developer.apple.com/documentation/appkit/nsview/1483719-updatetrackingareas)?

src/platform_impl/macos/view.rs
```diff
@@ -199,25 +199,34 @@ declare_class!(
...
+                self.removeTrackingArea(&tracking_area);
             }
-
             let rect = self.visibleRect();
-            let tracking_rect = self.add_tracking_rect(rect, false);
-            self.state.tracking_rect = Some(tracking_rect);
+            let tracking_area = self.init_and_add_tracking_area(
+                NSTrackingAreaOptions::NSTrackingMouseEnteredAndExited
+                    | NSTrackingAreaOptions::NSTrackingMouseMoved
+                    | NSTrackingAreaOptions::NSTrackingActiveAlways,
```
madsmtm, line 226: You're using `NSTrackingActiveAlways` here, while using `NSTrackingActiveInActiveApp` above. Is this intentional?

madsmtm (changes requested):
> could be used to implement some of the cursor-related features that aren't currently implemented for macOS.

I assume that you would like to implement those later, then?

> I'm pretty inexperienced when it comes to working with big codebases, and am happy to hear any advice/requested changes.

No worries, things look generally fine

The-Minecraft-Scientist:
I followed all of your suggestions, but I changed this PR to a draft since I plan to implement some missing cursor features before trying to merge.

src/platform_impl/macos/view.rs
The-Minecraft-Scientist: Nope! I had both places set to `NSTrackingActiveAlways` for testing, but I forgot to set them both back to NSTrackingActiveInActiveApp when I pushed this commit

Random-Scientist:
Closing in favor of #4514

Issue #971, opened 2019-06-23
Don't panic :)


felixrabe:
I think there is too much `unwrap` in winit and it could benefit from a healty dose of `Result`s. Thus I'm opening this issue to spark discussion on:

[ ] Is such a change desirable?
[ ] If yes, start a PR converting a handful of currently non-Result-returning functions to return `Result`.
[ ] Decide on when to release this. Seems to make sense to not include this change in 0.20, but maybe 0.21?
[ ] Pretty please use the codename "Don't panic :)" for that release. :)

[Original discussion](https://github.com/rust-windowing/winit/issues/806#issuecomment-504714535).

felixrabe:
Results can always be converted to panics, but not the other way round.

vberger:
I'll vote against blindly trying to remove all panics in winit, for the following reason.

There are imo two kinds of panics:

- "something bad occurred but it might be feasible to recover", these are absolutely worth converting to errors (examples would be "the current platform does not support the requested actions")
- "something was fundamentally broken in the internal state of winit", these panics reveal a bug in winit, and the downstream crate cannot really do anything about it. I believe it makes sense to keep these panics (for example the wayland backend contains a large number of `unwrap`s that fall in this category).

So I'd prefer a guiding rule like: "Don't panic if this error occurring is not a bug and can be handled by the downstream crate".

Osspial:
I largely agree with @vberger here. There *is* one caveat I'd like to make here, though: In the "fundamentally broken" case, we should be using `expect` instead of `unwrap`, so that people seeing the panic can know what's causing it and where to report it.

Still, I like what this issue is trying to accomplish. In terms of concrete steps to take, I'd suggest banning `unwrap`s in favor of `expect` in future PRs, and then going over Winit's codebase to determine what panics are the result of bugs and which should be removed.

felixrabe:
I think we actually agree. I propose sensible conversions of `unwrap` to `Result` where someone originally prototyped the API and it was just not changed since, because it would break the API. Not unrecoverable error conditions. (That's what I meant by "healthy dose", and I'm currently also working my way through "Programming Rust" that contains a very nice section on the semantics of `panic` vs. `Result`.)

winit was started before Rust was even 1.0, and the `?` operator that helps making `Result`s much easier to use has only been added last year, so it makes sense that in the spirit of "let's ship it", many places where `Result` would (now) be preferable were just left to panic instead.

maxjoehnk:
[...]

PR #2646, opened 2023-01-23
Handle error cases in `MonitorHandle`

madsmtm:
Part of https://github.com/rust-windowing/winit/issues/971.

Makes all `MonitorHandle` methods fallible. The same is not done for `VideoMode`, since they are just static blobs of data (rather than something queried dynamically) (but that might change in the future?).

  - [ ] Android
  - [x] iOS
  - [x] macOS
  - [ ] Wayland
  - [ ] Web
  - [ ] Windows
  - [ ] X11
  - [ ] Orbital

### Unresolved

I'm unsure what we should do in the case where some functionality isn't implemented yet? Is the correct thing to return a dummy value? Perhaps it would actually be better if we just panic with an `unimplemented!()`?

Though in any case I think this PR is an improvement over the status quo.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,14 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- Added `monitor::MonitorGone`.
+- **Breaking:** Changed the following methods to return `Result<T, MonitorGone>`, allowing handling errors that occur if the monitor disconnects:
+  - `MonitorHandle::name`
+  - `MonitorHandle::size`
+  - `MonitorHandle::position`
+  - `MonitorHandle::refresh_rate_millihertz`
+  - `MonitorHandle::scale_factor`
+  - `MonitorHandle::video_modes`
 - Add `Window::is_minimized`.
 - On X11, fix errors handled during `register_xlib_error_hook` invocation bleeding into winit.
 - Add `Window::has_focus`.
--- a/examples/fullscreen.rs
+++ b/examples/fullscreen.rs
@@ -25,7 +25,11 @@ fn main() {
     println!("Monitor: {:?}", monitor.name());
 
     let mut mode_index = 0;
-    let mut mode = monitor.video_modes().next().expect("no mode found");
+    let mut mode = monitor
+        .video_modes()
+        .expect("monitor gone")
+        .next()
+        .expect("no mode found");
     println!("Mode: {}", mode);
 
     println!("Keys:");
@@ -78,16 +82,26 @@ fn main() {
                         println!("Monitor: {:?}", monitor.name());
 
                         mode_index = 0;
-                        mode = monitor.video_modes().next().expect("no mode found");
+                        mode = monitor
+                            .video_modes()
+                            .expect("monitor gone")
+                            .next()
+                            .expect("no mode found");
                         println!("Mode: {}", mode);
                     }
                     VirtualKeyCode::M => {
                         mode_index += 1;
-                        if let Some(m) = monitor.video_modes().nth(mode_index) {
+                        if let Some(m) =
+                            monitor.video_modes().expect("monitor gone").nth(mode_index)
+                        {
                             mode = m;
                         } else {
                             mode_index = 0;
-                            mode = monitor.video_modes().next().expect("no mode found");
+                            mode = monitor
+                                .video_modes()
+                                .expect("monitor gone")
+                                .next()
+                                .expect("no mode found");
                         }
                         println!("Mode: {}", mode);
                     }
--- a/examples/monitor_list.rs
+++ b/examples/monitor_list.rs
@@ -1,13 +1,31 @@
 #![allow(clippy::single_match)]
+use std::thread;
+use std::time::Duration;
 
 use simple_logger::SimpleLogger;
-use winit::{event_loop::EventLoop, window::WindowBuilder};
+use winit::event_loop::EventLoop;
 
 fn main() {
     SimpleLogger::new().init().unwrap();
     let event_loop = EventLoop::new();
-    let window = WindowBuilder::new().build(&event_loop).unwrap();
 
-    dbg!(window.available_monitors().collect::<Vec<_>>());
-    dbg!(window.primary_monitor());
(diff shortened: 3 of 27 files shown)


kchibisov:
>I'm unsure what we should do in the case where some functionality isn't implemented yet? Is the correct thing to return a dummy value? Perhaps it would actually be better if we just panic with an unimplemented!()?

That's not an option. Would suggest to return a `Result` instead.

kchibisov (changes requested):
I think we could add a `MonitorData` or something like that and access it within the callback?

Then on the monitor handle itself we could have 2 methods.

```rust
fn is_alive(&self) -> bool;
fn with_data<T, F: Fn(MonitorData) -> T>(&self) -> Result<T>;
```

Then on that data you can access the things you need, that way it'll be less `result` unwraps and such.

madsmtm:
> ```rust
> fn is_alive(&self) -> bool;
> fn with_description<T, F: Fn(MonitorDescription) -> T>(&self) -> Result<T>;
> ```

There is no callback you could write in which `MonitorDescription` will always be available, since the user could remove their monitor at any time - you must do a copy of the underlying data, in which case `-> Result<MonitorDescription>` would work just fine.

If we do that though, then I think I'd rather just rename `MonitorHandle` to `MonitorDescription` altogether.

> > I'm unsure what we should do in the case where some functionality isn't implemented yet? Is the correct thing to return a dummy value? Perhaps it would actually be better if we just panic with an unimplemented!()?
> 
> That's not an option. Would suggest to return a `Result` instead.

I actually think the `unimplemented!()` strategy would be a fairly good option:
- If someone doesn't use the functionality, or don't support the platform it's being used on, there's no problem
- If someone use it on a platform where it's not supported, they must manually `cfg`-gate the functionality
- Which in turn would drive people to actually implement the functionality!

Vs. returning a `Result<T, (Unimplemented | MonitorGone)>`:
- People will likely just `.unwrap_or(...)`, which will coalesce the two very different errors together
- Once the feature _has_ been implemented, we now have an unused error case (and an implementation that wasn't there before could even somewhat be a breaking change).

Of course, for the things that are fundamentally not _possible_ to support, then we should do change the API.

---

In any case, we should consider making video modes match monitors: So either we have `MonitorHandle` and `VideoModeHandle`, or `MonitorDescription` and `VideoMode`.

dhardy:
`MonitorGone` where? `MonitorMissing` better. (Excuse the missing prepositions.)

Unfortunately this clashes with #2658.

> So either we have MonitorHandle and VideoModeHandle, or MonitorDescription and VideoMode.

This sounds like unnecessary complexity for the sake of consistency. That said, returning a `MonitorDescription` would simplify the current API. I guess the cost is unimportant (extra platform requests for details the app may not want anyway)?

The one part of `MonitorHandle` that confuses me: it appears to be both an identifier and a descriptor? In any case, the only real *purpose* of Winit returning monitor details, as far as I can tell, is to support `window.set_fullscreen(..)`.

xiaopengli89:
How to deal with the situation that the handle value is reused, which will cause consistency problems, how about using `std::os::windows::io::BorrowedHandle::try_clone_to_owned`?

daxpedda (changes requested):
Small nits, otherwise LGTM!

kchibisov (changes requested):
I don't like the approach because the problem is that lifetime should be attached to the event loop, where we don't do that. This problem is solved on winit-next where monitor handle has its own ID and you retrieve it from the event loop, thus none of that can happen on all the platforms.

Adding results to every method doesn't sound appealing to me.

madsmtm:
We talked a bit about this in Friday's meeting, the API should stay mostly the same, but instead each backend should cache the required data about the monitor, such that the data can always be retrieved.

I don't think I communicated this properly then, but I think it'd be fine for the backend to keep a reference to the monitor data alive, if that's enough that it can be fetched.

Internally, it would be something like:
```rust
// Backend 1
struct MonitorHandle {
    // Not retrievable from the handle if the monitor is disconnected, so cached here
    cached_data: {
        name: String,
        refresh_rate: u32,
        ...
    },
    // Used when entering fullscreen on the monitor
    handle: *mut MyHandle,
}

// Backend 2
struct MonitorHandle {
    // Used when entering fullscreen on the monitor, and to retrieve data about the monitor
    handle: Arc<MyHandle>,
}
```

daxpedda:
> I don't think I communicated this properly then, but I think it'd be fine for the backend to keep a reference to the monitor data alive, if that's enough that it can be fetched.

I think we should keep it consistent and not do fetching on some backends and on others we cache. Because we will have to add a big note saying "this data is cached and reflects the state when it was retrieved". Adding backend specific exceptions makes this even more unintuitive.

I would still be fine to leaving things as proposed here, with returning a `Result`, as I don't really see a problem with that, but just caching the data seems to be the only compromise we managed to reach.

kchibisov:
In the current state of thins it shouldn't cache, since there's no way to propagate that the cache got invalidated or the data got updated.

daxpedda:
> In the current state of thins it shouldn't cache, since there's no way to propagate that the cache got invalidated or the data got updated.

What changed your mind? Because that is what we agreed on the last time.

kchibisov:
All backends right now don't cache and kind of live update or query the data. I guess I misunderstood it, since I thought that we shouldn't cache in the future.

Right now to remove reload/cache of data in handles it'll require to likely rewrite all of them all together, so this PR will be way bigger than _report errors_.

kchibisov:
Or what do you mean by caching in particular, like _remember last asked value_ or use only POD and update when the user queries the monitor?

daxpedda:
> Or what do you mean by caching in particular, like _remember last asked value_ or use only POD and update when the user queries the monitor?

The idea was to use POD only, no updating. If you want updated values you have to query the monitor from the event loop again. It might be sensible to add an `refresh()` method though.

kchibisov:
>The idea was to use POD only, no updating. If you want updated values you have to query the monitor from the event loop again. It might be sensible to add an refresh() method though.

Yeah, it's just will have to rewrite all the monitor handles code, since everything is doing the opposite. I'm not against that though, just way more work for this patch and it does completely different compared to this patch.

PR #2647, opened 2023-01-23
implement pen for X11

DorianRudolph:
Issue: https://github.com/rust-windowing/winit/issues/99

I ported my previous pen work (https://github.com/rust-windowing/winit/pull/1879) over to the API from Android/Windows PR (https://github.com/rust-windowing/winit/pull/2396).

Though I have a couple of concerns with that API. I'm not sure if it is best to add the pen pressure on top of the normal touch event. I think an application developer would want to treat pens and touches quite differently. For example, you probably don't want a pinch-to-zoom gesture with one finger and the pen. Though I can also see that if the pen is the sole input, you would also treat it as a touch.
It also seems we break the "contract" of the touch phase, because we need to emit move emits between stop and start when the pen is hovering over the screen.
We could also simplify Touch event by putting `pen_state` into `force`. The unsupported values can just be set to a sensible default.
Note, the Apple Pencil 2 also supports hovering and has a button, so the additional `PenState` variable could probably be filled for that as well. Though I'm not an apple developer.

It is also not clear to me what the difference between `SharedState` and putting state into the `EventProcessor` struct directly. Specifically, where should I put the `PenStatus` struct?

I'm also not sure which button to map to `barrel` and `eraser`. My pen has two side buttons, I mapped the lower to the eraser. Maybe if someone's pen has an actual eraser, they could comment to which button X11 maps that.

Lastly, I would also like to have pen support on wayland, but right now seems an inopportune time to implement that since SCTK is currently being rewritten from scratch, so I don't want to spend much effort on adding pen support to the current version, which is used by winit. It also seems like a lot of work to port winit to the new version of SCTK since its API changes significantly.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -976,8 +976,36 @@ pub struct Touch {
     ///
     /// - Only available on **iOS** 9.0+ and **Windows** 8+.
     pub force: Option<Force>,
+
+
+    /// Describes an press with a digital pen. Is `None` if the Touch event was not triggered
+    /// by a digital pen
+    ///
+    /// ## Platform-specific
+    ///
+    /// - Only available on **Android** and **Windows** 8+.
+    pub pen_state: Option<PenState>,
+
     /// Unique identifier of a finger.
     pub id: u64,
+}
+
+/// Describes the physical state of a digital pen
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct PenState {
+    /// The clockwise rotation of the pointer normalized in a range of 0 to 359.
+    /// Defaults to 0 if not available on platform.
+    pub rotation: f64,
+    /// The angle of tilt of the pointer in a range of -90 to +90 for each axis,
+    /// with a positive value indicating a tilt to the right or towards the user.
+    /// Defaults to (0, 0) if not available on platform.
+    pub tilt: (f64, f64),
+    /// The barrel button is pressed.
+    pub barrel: bool,
+    /// The pen is inverted.
+    pub inverted: bool,
+    /// The eraser button is pressed.
+    pub eraser: bool,
 }
 
 /// Describes the force of a touch event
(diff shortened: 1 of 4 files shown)


xyning:
pen data shouldn't be merged with touch imo as it will complicate the touch events much. pen tablets are different from mouses and touchscreens in many ways, and they deserve their own event types. also, events can be easily combined on the app side if developers want to treat pen events as touches

i did some experiments about basic wayland pen support before. it starts to work inside the client area, but the default window decor doesn't respond to pen events :( . seems like a sctk issue, not sure though

src/event.rs
```diff
@@ -976,10 +976,38 @@ pub struct Touch {
...
+/// Describes the physical state of a digital pen
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct PenState {
+    /// The clockwise rotation of the pointer normalized in a range of 0 to 359.
+    /// Defaults to 0 if not available on platform.
+    pub rotation: f64,
+    /// The angle of tilt of the pointer in a range of -90 to +90 for each axis,
+    /// with a positive value indicating a tilt to the right or towards the user.
+    /// Defaults to (0, 0) if not available on platform.
+    pub tilt: (f64, f64),
```
i509VCB, line 1002: These seem to be in degrees, probably best to explicitly mention that.

DorianRudolph:
@Bloeckchengrafik I was not really happy with gluing everything onto the touch event, and I haven't really had the time to work on that since. I do plan on building an app with pen support, but I can't say if and when that will happen.

notgull:
@kchibisov @madsmtm @daxpedda Do we have a plan-of-attack for pen events like this? I think that other platforms have concepts of this type of event as well.

kchibisov:
>Do we have a plan-of-attack for pen events like this?

Not yet, but we'd need to decide whether we emulate internally certain things or we just don't try to use tablet when it's not present.

daxpedda:
Implementing pen events for Web would be simple, everything is already in place to add it.

Issue #2582, opened 2022-12-03
Feature request: Support for Wayland's wlr-layer-shell protocol


Keating950:
I would find it really helpful if winit provided a friendly interface for creating [layer shells](https://wayland.app/protocols/wlr-layer-shell-unstable-v1) in Wayland, perhaps as an extension of WindowBuilder. So far, I haven't been able to find a solution for this in the Rust ecosystem other than by going to the relatively low-level [SCTK](https://github.com/Smithay/client-toolkit). I'm imagining this feature could be gated off in one of the traits in [winit::platform::unix](https://docs.rs/winit/latest/winit/platform/unix/index.html) module.

kchibisov:
We actually could support it given that android works similar to the wayland layer-shell stuff. So I'd keep that open, since we might add support for that.

kchibisov:
Similar as in it destroys window from time to time. The problem is that keyboard input will work strangely and mouse input as well.

I'll keep it open to until ether someone or I will try to add it and we'll see how it goes. Though it might be not added in the end given how weird it is.

TheOnlyMrCat:
sctk 0.16 doesn't provide any abstraction for layer shells, but there is an implementation on master. If I understand correctly, this means this feature is waiting on the release of wayland-rs 0.30, as well as the refactor that update would require within winit code.

kchibisov:
Yeah, I'm waiting for wayland-rs for a long time already.

yavko:
> sctk 0.16 doesn't provide any abstraction for layer shells, but there is an implementation on master. If I understand correctly, this means this feature is waiting on the release of wayland-rs 0.30, as well as the refactor that update would require within winit code.

Once #2676 is merged, this won't be the case anymore, as sctk `0.17` is out, and it does provide the abstractions.

Decodetalkers:
Now sctk 0.17 is merged.. Then what is going on? I cannot wait to use layershell

yavko:
> Now sctk 0.17 is merged.. Then what is going on? I cannot wait to use layershell

No one is working on it, that's what's happening. If you want to use it, then add it yourself. Though if I have time I'll try to learn how winit works and *try* impl it myself

RegenJacob:
Would be great to get [window::WindowLevel](https://docs.rs/winit/latest/winit/window/enum.WindowLevel.html#) on wayland

ids1024:
[...]

Issue #2160, opened 2022-01-23
"Blessed" way to do things out of `winit`s scope? (`winit-extras`)


madsmtm:
There are a lot of things out of `winit`'s scope that are useful (and maybe expected by most people that we provide), here's a few examples I've stumbled across, though there are _many_ more:
- Clipboards (https://github.com/rust-windowing/winit/pull/2156)
- macOS gestures (https://github.com/rust-windowing/winit/pull/2157)
- Maybe popups? (https://github.com/rust-windowing/winit/issues/403)
- Menubars (https://github.com/rust-windowing/winit/issues/427 / https://github.com/rust-windowing/winit/pull/1921)
- macOS file open events (https://github.com/rust-windowing/winit/pull/1759)

The readme already notes this, and directs users to various graphics libraries, but these are not always what the user wants (or their graphics library may not support it).

It [has been suggested](https://github.com/rust-windowing/winit/wiki/Missing-features-provided-by-other-crates#wip-features-omitted-from-winit-10) we make a `winit-extras` crate or similar under the @rust-windowing umbrella (name up for bikeshedding), to house these components; this would help us with:
- Providing examples of usage
- Ensuring our APIs actually allow external crates to hook in to the functionality they need (we're working towards this, see https://github.com/rust-windowing/winit/issues/2120)
- Helps keeping these extra features up to date with the latest `winit` version
- In the cases where we depend on an external library for this, this "blesses" that library, providing a more consistent user experience (currently there are _many_ clipboard libraries out there; it would be nice if the ecosystem could converge on a single one)

The alternative is just to make crates for each of these things, and document them in the wiki (though for some of the smaller features like file open events it seems kinda dumb).

madsmtm:
Third alternative is to simply increase the scope, though that brings other problems (such as not having enough maintainer time to support it and general feature creep) - though creating something else under the @rust-windowing umbrella is kinda similar to increasing the scope...

tinaun:
note that the tauri maintainers have already forked winit to add menus / clipboards etc in [tao](https://docs.rs/tao/latest/tao/)

notgull:
I would, personally, support this. I was actually about to make my own extension crate to do this, but I feel having an official endorsement from @rust-windowing would boost its legitimacy.

parasyte:
[...]

PR #2648, opened 2023-01-24
Add `cursor_position` getter API

amrbashir:

### DIFF
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -15,6 +15,8 @@ use instant::{Duration, Instant};
 use once_cell::sync::OnceCell;
 use raw_window_handle::{HasRawDisplayHandle, RawDisplayHandle};
 
+use crate::dpi::PhysicalPosition;
+use crate::error::ExternalError;
 use crate::{event::Event, monitor::MonitorHandle, platform_impl};
 
 /// Provides a way to retrieve events from the system and from the windows that were registered to
@@ -358,6 +360,16 @@ impl<T> EventLoopWindowTarget<T> {
     pub fn set_device_event_filter(&self, _filter: DeviceEventFilter) {
         #[cfg(any(x11_platform, wayland_platform, windows))]
         self.p.set_device_event_filter(_filter);
+    }
+
+    /// Returns the current cursor position
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **iOS / Android**: Unsupported.
+    #[inline]
+    pub fn cursor_position(&self) -> Result<PhysicalPosition<f64>, ExternalError> {
+        self.p.cursor_position()
     }
 }
 
--- a/src/platform_impl/macos/util/mod.rs
+++ b/src/platform_impl/macos/util/mod.rs
@@ -9,6 +9,8 @@ use objc2::foundation::{CGFloat, NSNotFound, NSPoint, NSRange, NSRect, NSUIntege
 
 use crate::dpi::LogicalPosition;
 
+use super::appkit::NSEvent;
+
 // Replace with `!` once stable
 #[derive(Debug)]
 pub enum Never {}
@@ -62,4 +64,9 @@ pub fn window_position(position: LogicalPosition<f64>) -> NSPoint {
         position.x as CGFloat,
         CGDisplay::main().pixels_high() as CGFloat - position.y as CGFloat,
     )
+}
+
+pub fn cursor_position() -> crate::dpi::PhysicalPosition<f64> {
+    let pt: NSPoint = NSEvent::mouseLocation();
+    (pt.x, pt.y).into()
 }
--- a/src/window.rs
+++ b/src/window.rs
@@ -833,7 +833,7 @@ impl Window {
 
     /// Sets the window to minimized or back
     ///
-    /// ## Platform-specific
+    /// ## Platform-specific1
     ///
     /// - **iOS / Android / Web / Orbital:** Unsupported.
     /// - **Wayland:** Un-minimize is unsupported.
@@ -1230,6 +1230,16 @@ impl Window {
     #[inline]
     pub fn set_cursor_hittest(&self, hittest: bool) -> Result<(), ExternalError> {
         self.window.set_cursor_hittest(hittest)
+    }
+
+    /// Returns the current cursor position.
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **iOS / Android / Wayland / Orbital / Web**: Unsupported, returns `None`.
+    #[inline]
+    pub fn cursor_position(&self) -> Result<PhysicalPosition<f64>, ExternalError> {
+        self.window.cursor_position()
     }
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- Add `Window::cursor_position` and `EventLoopWindowTarget::cursor_position`.
 - Add `Window::is_minimized`.
 - On X11, fix errors handled during `register_xlib_error_hook` invocation bleeding into winit.
 - Add `Window::has_focus`.
--- a/examples/window.rs
+++ b/examples/window.rs
@@ -17,9 +17,11 @@ fn main() {
         .build(&event_loop)
         .unwrap();
 
-    event_loop.run(move |event, _, control_flow| {
+    event_loop.run(move |event, event_loop, control_flow| {
+        dbg!(event_loop.cursor_position().unwrap());
+
         control_flow.set_wait();
-        println!("{:?}", event);
+        // println!("{:?}", event);
 
         match event {
             Event::WindowEvent {
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -836,6 +836,12 @@ impl<T: 'static> EventLoopWindowTarget<T> {
     pub fn raw_display_handle(&self) -> RawDisplayHandle {
         RawDisplayHandle::Android(AndroidDisplayHandle::empty())
     }
+
+    pub fn cursor_position(&self) -> Result<PhysicalPosition<f64>, error::ExternalError> {
+        Err(error::ExternalError::NotSupported(
+            error::NotSupportedError::new(),
+        ))
+    }
 }
 
 #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
@@ -1025,6 +1031,12 @@ impl Window {
     }
 
     pub fn set_cursor_visible(&self, _: bool) {}
+
+    pub fn cursor_position(&self) -> Result<PhysicalPosition<f64>, error::ExternalError> {
+        Err(error::ExternalError::NotSupported(
+            error::NotSupportedError::new(),
+        ))
+    }
 
     pub fn drag_window(&self) -> Result<(), error::ExternalError> {
         Err(error::ExternalError::NotSupported(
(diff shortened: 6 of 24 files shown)


filnet:
What's the motivation for getting the cursor position with the new function over getting it via an event ?

kchibisov:
>    get the cursor location
     get the monitor based on the cursor location

This won't work on macOS/Wayland for sure?

amrbashir:
Wayland yes, macOS should be possible, no?

amrbashir:
for macOS there is no direct API to get a monitor based on a point but we can iterate over available monitors and check if the point is contained within one of them

kchibisov:
>Wayland yes, macOS should be possible, no?

On Wayland certainly not. Since you know only about the cursor over your window. You don't even have control were to open a window, so use-case you're describing is very specific and can be achieved on X11 and maybe Windows.

kchibisov:
And in my experience normal window managers will open a new window where your cursor is if you don't explicitly specify the output you want(like on Wayland it's done by majority).

amrbashir:
yeah mostly you don't need to manually position your window but there is use-cases where you want to control which monitor to spawn the window on, for example: spotlight-like apps, widgets, overlays...etc

kchibisov:
But you already have all in-place for that? You could just report `current_monitor` on the event_loop or such?

ids1024:
> On Wayland certainly not. Since you know only about the cursor over your window. You don't even have control were to open a window, so use-case you're describing is very specific and can be achieved on X11 and maybe Windows.

You should be able to implement  a centered "spotlight-like" app like this on Wayland with layer-shell (in compositors that support it), but yeah, you couldn't implement it this way.

src/event_loop.rs
```diff
@@ -359,6 +361,16 @@ impl<T> EventLoopWindowTarget<T> {
         #[cfg(any(x11_platform, wayland_platform, windows))]
         self.p.set_device_event_filter(_filter);
     }
+
+    /// Returns the current cursor position in screen coordinates.
+    ///
```
madsmtm, line 366: Could you document what the difference between `Window::cursor_position` and `EventLoopWindowTarget::cursor_position` is? Or are they exactly the same? And if they're the same, perhaps we can get by with just one of them?

src/window.rs
```diff
@@ -833,7 +833,7 @@ impl Window {
 
     /// Sets the window to minimized or back
     ///
-    /// ## Platform-specific
+    /// ## Platform-specific1
```
madsmtm, line 836: Nit: Typo

src/window.rs
```diff
@@ -1231,6 +1231,16 @@ impl Window {
     pub fn set_cursor_hittest(&self, hittest: bool) -> Result<(), ExternalError> {
         self.window.set_cursor_hittest(hittest)
     }
+
+    /// Returns the current cursor position in screen coordinates.
```
madsmtm, line 1235: Perhaps it would make sense to convert it to window coordinates first, then?

src/platform_impl/macos/util/mod.rs
```diff
@@ -63,3 +65,8 @@ pub fn window_position(position: LogicalPosition<f64>) -> NSPoint {
         CGDisplay::main().pixels_high() as CGFloat - position.y as CGFloat,
     )
 }
+
+pub fn cursor_position() -> crate::dpi::PhysicalPosition<f64> {
+    let pt: NSPoint = NSEvent::mouseLocation();
```
madsmtm, line 70: Huh, I would have expected this to be main-thread only, but I think it's actually fine

madsmtm (changes requested):
macOS impl approved, unsure about desired API, I'll leave that up for you to decide on.

src/event_loop.rs
amrbashir: yeah, they are the same, and is provided on `Window` for convenience purposes only.

src/platform_impl/macos/util/mod.rs
amrbashir: I actually don't have a macOS to test this, so I have to take your word for it.

src/window.rs
amrbashir: Idk, maybe, but the cursor could be out of the window, should it report (0,0) in that case?

madsmtm: Maybe it should just return an `option`?

In any case, I think we should keep just the one on `EventLoopTarget`, that at least makes it very clear that the position is in screen-coordinates, and not window coordinates (and getting window-relative cursor position is fairly easy using `Window::inner_position`)

filnet:
> This is useful for positioning the window on a particular monitor when creating the window, for example, when making a spotlight-like app, where you want to show the window on the monitor the has the cursor, then I would need to:
> 
>     1. get the cursor location
> 
>     2. get the monitor based on the cursor location
> 
>     3. center the app on the monitor
> 
> 
> This PR is for 1 and I will be opening another PR for 2, later today or tomorrow.



> This is useful for positioning the window on a particular monitor when creating the window, for example, when making a spotlight-like app, where you want to show the window on the monitor the has the cursor, then I would need to:
> 
>     1. get the cursor location
> 
>     2. get the monitor based on the cursor location
> 
>     3. center the app on the monitor
> 
> 
> This PR is for 1 and I will be opening another PR for 2, later today or tomorrow.

May be you should add a feature that does 1+2+3 ?

src/event_loop.rs
filnet: the new function stands out like a sore thumb here...
and it opens a door to many more convenience functions.
do we want to go that down that road.

amrbashir: Well, there is `primary_monitor` and `available_monitors` which are provided on `Window` for convenience too. I just followed the pattern, but I don't have a strong feeling for it and removing it, is fine with me.

amrbashir:
> May be you can keep 1 private for 2 ?

2 has been rejected in https://github.com/rust-windowing/winit/pull/2649 so it has to exposed

ids1024:
As I mentioned earlier, the "spotlight-like app" use case on Wayland would be served by layer shell (https://github.com/rust-windowing/winit/issues/2582). I think what would be ideal for that is if either winit or an extension crate could provide an API that can use layer shell, or position things more manually on X/win32/etc. to provide as portable a solution as possible for this particular use case.

I would generally agree with the authors of the base Wayland protocol that applications placing windows at specific spots on screen is generally a bad practice outside of specialized things like this (though Winit already has `set_outer_position`).

https://github.com/rust-windowing/winit/issues/2160 is relevant, wrt including things in winit vs a separate crate.

src/event_loop.rs
amrbashir: removed it from `Window` as I couldn't think of an immediate use-case for it.

src/window.rs
amrbashir: removed it from `Window` as I couldn't think of an immediate use-case for it.

daxpedda (commented):
We just decided in the meeting that we are happy to move forward with this.
The coordinate system used here should be the same as for all other APIs covering the cursor position.

There was also a concern raised about multiple pointers, which is something that we decided we will handle separately but will most likely include adding a device ID parameter.

src/changelog/unreleased.md
```diff
@@ -41,3 +41,4 @@
 - Add `Window::default_attributes` to get default `WindowAttributes`.
 - `log` has been replaced with `tracing`. The old behavior can be emulated by setting the `log` feature on the `tracing` crate.
 - On Windows, confine cursor to center of window when grabbed and hidden.
+- Add `ActiveEventLoop::cursor_position`
```
daxpedda, line 44:
```suggestion
- Add `ActiveEventLoop::cursor_position`.
```

daxpedda (commented):
LGTM.
Can't review the relevant backends though and the CI needs to pass.

amrbashir:
Seems like a couple of tests are failing on nightly but should be unrelated to this PR

daxpedda (approved):
Thanks!

CI should be fixed after https://github.com/rust-windowing/winit/pull/3604.

src/event_loop.rs
```diff
@@ -490,6 +491,16 @@ impl ActiveEventLoop {
             platform: self.p.owned_display_handle(),
         }
     }
+
+    /// Returns the current cursor position in screen coordinates.
```
madsmtm, line 495: Referencing https://github.com/rust-windowing/winit/pull/2648#pullrequestreview-1939258628.

I think we _thought_ this was on `Window`, and hence expected things to be in window coordinates, just like the coordinates for `set_cursor_position` (such that `window.set_cursor_position(window.cursor_position()?)?` is (roughly) a no-op). But we decided against `Window::cursor_position` in https://github.com/rust-windowing/winit/pull/2648/files#r1090072945, this API is `ActiveEventLoop::cursor_position`.

In that case, I'd expect the coordinates to be the same as those from `Window::outer_position`, is that correct? If so, could you refer to that for discussion about the coordinate space, just like `Window::set_outer_position` does?

madsmtm: (If you think that this interpretation is correct, I'll test the macOS implementation myself to ensure that it matches `Window::outer_position`).

amrbashir: > In that case, I'd expect the coordinates to be the same as those from `Window::outer_position`, is that correct?

Yeah

> If so, could you refer to that for discussion about the coordinate space, just like `Window::set_outer_position` does?

like this https://github.com/amrbashir/winit/commit/1f9a3df3ff39eca403d0963dbee72267fdb29a2b ?

src/event_loop.rs
```diff
@@ -490,6 +491,18 @@ impl ActiveEventLoop {
             platform: self.p.owned_display_handle(),
         }
     }
+
+    /// Returns the current cursor position relative to the top-left hand corner of the desktop.
```
kchibisov, line 495: I think we've agreed that it should be the same as in events, @daxpedda ? 

So it should be relative to Window itself.

If it's not, it should be a separate API call with a special name.

kchibisov (changes requested):

src/event_loop.rs
kchibisov: Oh, nvm, it's on the `ActiveEventLoop`.

I'll look once I have time then.

src/event_loop.rs
```diff
@@ -490,6 +491,18 @@ impl ActiveEventLoop {
...
+    ///
+    /// See [`Window::outer_position`] for more information about the coordinates.
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **iOS / Android / Wayland / Orbital / Web**: Unsupported.
+    #[inline]
+    pub fn cursor_position(&self) -> Result<PhysicalPosition<f64>, ExternalError> {
```
madsmtm, line 503: It confused me that the return type of this is not `PhysicalPosition<i32>` like the return type of `Window::outer_position`, but I guess it matches the `CursorMoved` event - apparently cursors can be positioned on a sub-pixel scale, didn't know that!

madsmtm, line 503: It confused me that the return type of this is not `PhysicalPosition<i32>` like the return type of `Window::outer_position`, but I guess it matches the `CursorMoved` event - apparently cursors can be positioned on a sub-pixel scale, didn't know that!

src/event_loop.rs
madsmtm: Yup.

And I've tested it now, and pushed a fix to return positions with the correct scaling and such (it's still a little broken, but that's a more general issue, see https://github.com/rust-windowing/winit/issues/3615).

madsmtm (approved):

src/event_loop.rs
daxpedda: Yeah, it wouldn't make sense indeed.

Issue #2645, opened 2023-01-23
Window positioning is weird on macOS


madsmtm:
`Window::outer_position` and `MonitorHandle::position` is fundamentally broken on macOS.

Screen coordinates are internally closer to a `LogicalPosition`, but since the return type is `PhysicalPosition`, it is converted using the monitor's scale factor, which is incorrect since it assumes that all monitors have that scale factor.

To illustrate the issue, imagine three monitors configured to be next to each other like this:
```
[           ]    [          ]    [           ]
[ 1 (HiDPI) ]    [     2    ]    [     3     ]
[ 1920x1080 ]    [ 1280x800 ]    [ 2560x1440 ]
[           ]    [          ]    [           ]
```

Then the desired and the actual output of the two methods would then be:

| | Desired return | Internal API returns |
| --- | --- | --- |
| `monitor_3.position().x` | `1920 * 2 + 1280` | `1920 + 1280` |
| `window_on_monitor_3.outer_position().x` | `positon_in_monitor + above` | `positon_in_monitor + above` |

madsmtm:
I can see a few ways of fixing this:
1. Make these methods return `Position` instead
2. Create a cursed helper function that fetches all monitors and computes the actual physical position by constructing a graph of bordering monitors. This requires that monitors are always bordering each other, which might not always hold?
3. Create a new type like `DesktopPosition`, which has conversion methods to/from `Position` with relevant documentation. This would also help separate "position on desktop" from "position in window"
  - Ideally we should probably have `ViewPosition`, `DesktopPosition`, and each of these should have variants `Physical`, `Logical`, `PhysicalDelta` and `LogicalDelta`. But that's probably overkill.

dhardy:
Without knowing anything about the size or scale factors of the monitors in your example above, it's not useful. I *presume* the sizes listed are physical pixels but they could even be logical pixels.

[...]

PR #2649, opened 2023-01-24
Add `monitor_from_point` API

amrbashir:
	- [X] Windows
	- [ ] X11
	- [ ] macOS
	- [ ] wayland
	- [ ] iOS
	- [ ] android

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- Add `Window::monitor_from_point` and `EventLoopWindowTarget::monitor_from_point`.
 - Add `Window::is_minimized`.
 - On X11, fix errors handled during `register_xlib_error_hook` invocation bleeding into winit.
 - Add `Window::has_focus`.
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -15,6 +15,7 @@ use instant::{Duration, Instant};
 use once_cell::sync::OnceCell;
 use raw_window_handle::{HasRawDisplayHandle, RawDisplayHandle};
 
+use crate::dpi::PhysicalPosition;
 use crate::{event::Event, monitor::MonitorHandle, platform_impl};
 
 /// Provides a way to retrieve events from the system and from the windows that were registered to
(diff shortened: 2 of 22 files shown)


jackpot51 (approved):
Orbital code looks good

kchibisov (changes requested):
I don't really like such an API, what's the use case for it? It seems like you wanted to pair it with the cursor position, but I think it's not a good thing to do.

What I'd suggest is to add `current_monitor` on the `EventLoopWindowTarget`, so you can get the currently focused monitor on the system, since that's all that matters and what you want anyway.

That way it'll work naturally with the rest of the `MonitorHandle` APIs. So for example on platforms when you don't know such a thing it'll return `None`, and given that most calls accept that `None` it could be used that way.

amrbashir:
> I don't really like such an API, what's the use case for it? It seems like you wanted to pair it with the cursor position, but I think it's not a good thing to do.

Most times, it will be paired with cursor_position but another use-case for tauri users, is to center the Window on a specific monitor, so the users provide the x,y and we use it to detect the nearest monitor and center the window on that monitor, but why isn't it a good thing?

> What I'd suggest is to add current_monitor on the EventLoopWindowTarget, so you can get the currently focused monitor on the system, since that's all that matters and what you want anyway.

Most cases yeah but there is a chance you may need other monitors.

kchibisov:
>Most times, it will be paired with cursor_position but another use-case for tauri users, is to center the Window on a specific monitor, so the users provide the x,y and we use it to detect the nearest monitor and center the window on that monitor, but why isn't it a good thing?

But you can least monitors and provide user to pick a monitor? Winit allows setting fullscreen on any monitor and you have them available on the event loop.

amrbashir:
> But you can least monitors and provide user to pick a monitor? 
by tauri users, I meant developers, not end-users.

dhardy:
We had a little discussion regarding units across multiple monitors in #2645. Short version: I don't think using `PhysicalPosition` for coordinates across multiple screens makes sense, though that is currently what other APIs use.

kchibisov:
The thing is that coordinate space system is using could be negative, don't match, and so on.

I think that there's no value in such API, given that we should use either `None`(let the system decide) or provide exact monitor (from `available_monitors').

The API with the points will be very limited and will add much complexity given that:

1. Not supported by most display servers.
2. Coordinate space could have negative values (at least on Wayland and it'll require custom protocols, yet you have no information where your cursor is), so the API need some restrictions.

My main issue is that lack of use case for such an API, since if you want a current monitor, right now you pass `None` to all APIs accepting output, and you'll get it. If you want a particular monitor you just pass a particular monitor. The location of outputs relative to each other is not that established concept (It's not even a rectangle shape to begin with).

amrbashir:
> My main issue is that lack of use case for such an API, since if you want a current monitor

I agree with that to an extent, the use case is very much rare so I don't really want to push hard on this since the functionality could be implemented in user code by iterating over the available monitors and comparing its bounds with the point.

PR #2650, opened 2023-01-25
Add support for setting the IME purpose

lukaslihotzki:
[Wayland](https://docs.rs/wayland-protocols/latest/wayland_protocols/wp/text_input/zv3/client/zwp_text_input_v3/enum.ContentPurpose.html) and [macOS](https://developer.apple.com/documentation/appkit/nstextcontenttype) support IME purposes. I have implemented it for Wayland only, because I don't have access to macOS. XIM and Windows IMM do not seem to support it. (Although #2614 removes XIM and the successor may support IME purposes.)

This PR adds the purpose `Password` (directly supported by both Wayland and macOS) and `Terminal` (needed for https://github.com/alacritty/alacritty/issues/6644). I don't know if it makes sense to support more purposes (for example, all that Wayland supports). The `ImePurpose` enum is non exhaustive, so new purposes can be added without breaking change.

- [x] Created or updated an example program if it would help users understand this functionality (added to the `ime` example, although the `ime` example does not work at all on Wayland because no window shows up)
- [x] Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented (IME is not contained in the feature matrix)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On Wayland, support fractional scaling via the wp-fractional-scale protocol.
 - On web, fix removal of mouse event listeners from the global object upon window distruction.
 - Add WindowAttributes getter to WindowBuilder to allow introspection of default values.
+- Added `Window::set_ime_purpose` for setting the IME purpose, currently implemented on Wayland only.
 
 # 0.27.5
 
--- a/src/platform_impl/linux/mod.rs
+++ b/src/platform_impl/linux/mod.rs
@@ -34,8 +34,8 @@ use crate::{
     },
     icon::Icon,
     window::{
-        CursorGrabMode, CursorIcon, ResizeDirection, Theme, UserAttentionType, WindowAttributes,
-        WindowButtons, WindowLevel,
+        CursorGrabMode, CursorIcon, ImePurpose, ResizeDirection, Theme, UserAttentionType,
+        WindowAttributes, WindowButtons, WindowLevel,
     },
 };
 
@@ -517,6 +517,11 @@ impl Window {
     #[inline]
     pub fn set_ime_allowed(&self, allowed: bool) {
         x11_or_wayland!(match self; Window(w) => w.set_ime_allowed(allowed))
+    }
+
+    #[inline]
+    pub fn set_ime_purpose(&self, purpose: ImePurpose) {
+        x11_or_wayland!(match self; Window(w) => w.set_ime_purpose(purpose))
     }
 
     #[inline]
--- a/src/platform_impl/linux/wayland/seat/text_input/handlers.rs
+++ b/src/platform_impl/linux/wayland/seat/text_input/handlers.rs
@@ -9,7 +9,7 @@ use crate::event::{Ime, WindowEvent};
 use crate::platform_impl::wayland;
 use crate::platform_impl::wayland::event_loop::WinitState;
 
-use super::{Preedit, TextInputHandler, TextInputInner};
+use super::{Preedit, TextInputHandler, TextInputInner, ZwpTextInputV3Ext};
 
 #[inline]
 pub(super) fn handle_text_input(
@@ -32,6 +32,7 @@ pub(super) fn handle_text_input(
             // Enable text input on that surface.
             if window_handle.ime_allowed.get() {
                 text_input.enable();
+                text_input.set_content_type_by_purpose(window_handle.ime_purpose.get());
                 text_input.commit();
                 event_sink.push_window_event(WindowEvent::Ime(Ime::Enabled), window_id);
             }
--- a/src/window.rs
+++ b/src/window.rs
@@ -1054,6 +1054,16 @@ impl Window {
         self.window.set_ime_allowed(allowed);
     }
 
+    /// Sets the IME purpose for the window using [`ImePurpose`].
+    ///
+    /// ## Platform-specific
+    ///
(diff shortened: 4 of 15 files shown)


jackpot51 (approved):
Orbital code looks good

CHANGELOG.md
```diff
@@ -62,6 +62,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On macOS, fix the mouse buttons other than left/right/middle being reported as middle.
 - On Wayland, support fractional scaling via the wp-fractional-scale protocol.
 - On web, fix removal of mouse event listeners from the global object upon window distruction.
+- Added support for setting the IME purpose, currently implemented on Wayland only.
```
kchibisov, line 65: Could you write out the exact API that you've added? Like `Window::set_ime_purpose`?

src/platform_impl/linux/mod.rs
```diff
@@ -519,6 +519,16 @@ impl Window {
         x11_or_wayland!(match self; Window(w) => w.set_ime_allowed(allowed))
     }
 
+    #[inline]
+    pub fn set_ime_purpose(&self, _purpose: ImePurpose) {
+        match self {
+            #[cfg(x11_platform)]
+            Window::X(_) => (),
+            #[cfg(wayland_platform)]
+            Window::Wayland(ref w) => w.set_ime_purpose(_purpose),
```
kchibisov, line 528: You should add a blank impl for X11 and use the macro, that way it should be consistent with the rest of the calls.

src/platform_impl/linux/wayland/seat/text_input/handlers.rs
```diff
@@ -31,6 +31,7 @@ pub(super) fn handle_text_input(
 
             // Enable text input on that surface.
             if window_handle.ime_allowed.get() {
+                text_input.set_content_type_by_purpose(window_handle.ime_purpose.get());
                 text_input.enable();
```
kchibisov, line 35: From the protocol:

 `The set_surrounding_text, set_content_type and set_cursor_rectangle requests must follow if the text input supports the necessary functionality.`
 
Could you swap them, just to prevent potential weird behavior? While the updates to protocol are atomic on commit, I'd do that just to ensure.

kchibisov (changes requested):
The API looks fine, though we might want to change IME stuff to do a single apply with some `ImeStateUpdate`, so we can utilize the atomicness of some of the underlying APIs.

src/window.rs
```diff
@@ -1025,6 +1025,17 @@ impl Window {
         self.window.set_ime_allowed(allowed);
     }
 
+    /// Sets the IME purpose for the window
```
kchibisov, line 1028:
```suggestion
    /// Sets the IME purpose for the window.
```

src/window.rs
```diff
@@ -1025,6 +1025,17 @@ impl Window {
...
+    ///
```
kchibisov, line 1029: Should reference the `ImePurpose` like we do in other APIs accepting enums.

src/window.rs
```diff
@@ -1516,3 +1527,21 @@ impl Default for WindowLevel {
         Self::Normal
     }
 }
+
+/// IME purposes.
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+#[non_exhaustive]
+pub enum ImePurpose {
+    /// default input
```
kchibisov, line 1535: Maybe something like that?

```suggestion
    /// No special hints for the IME.
```

src/window.rs
```diff
@@ -1516,3 +1527,21 @@ impl Default for WindowLevel {
...
+    Normal,
+    /// input a password
```
kchibisov, line 1537:
```suggestion
    /// The IME is used for password input.
```

src/window.rs
```diff
@@ -1516,3 +1527,21 @@ impl Default for WindowLevel {
...
+    Password,
+    /// input for a terminal
```
kchibisov, line 1539:
```suggestion
    /// The IME is used to input into the terminal.
    ///
    /// For example that could alter OSK on Wayland to show extra buttons.
```

src/window.rs
```diff
@@ -1516,3 +1527,21 @@ impl Default for WindowLevel {
         Self::Normal
     }
 }
+
+/// IME purposes.
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+#[non_exhaustive]
+pub enum ImePurpose {
```
kchibisov, line 1534: I think we should also document which platform supports which variant exactly.

src/window.rs
```diff
@@ -1516,3 +1527,21 @@ impl Default for WindowLevel {
         Self::Normal
     }
 }
+
+/// IME purposes.
```
kchibisov, line 1531: Would should state that it's more of a hint.

src/window.rs
kchibisov: And also likely add a note that it's a default.

kchibisov (changes requested):

src/window.rs
lukaslihotzki: Currently, only Wayland is implemented, so there are only variants that work there. If other platforms are implemented, this should be documented per variant.

kchibisov: I guess it's fine then, but you should have a section like we have for everything else.

kchibisov (changes requested):

kchibisov (approved):

PR #2652, opened 2023-01-27
Run clippy only on MSRV

kchibisov:


kchibisov:
Unfortunately can't do that, since dev dependencies could use higher version and we want to lint them as well.

notgull:
We could use `cargo hack --no-dev-deps` to circumvent this.

kchibisov:
>We could use cargo hack --no-dev-deps to circumvent this.

But that would require installing in into the CI each time...

PR #2659, opened 2023-01-29
Fix rerun-if-changed emmiting from build.rs

### LINKED ISSUE #2657: On all platforms, winit is rebuilt every time I build
Sorry if this is a known/expected behavior - I could not find any mention of it.

I link my projects to a master-branch-clone of `winit` by relative path, and for a while now I've noticed that doing a `cargo build` on any of my apps causes `winit` and all my dependencies on it to rebuild, even when no changes have been made to any code (and certainly not to `winit` itself).  Doing a `cargo build --verbose` confirms that `winit` is never "fresh" and it recompiles each time.

I tried to understand why but so far I have failed. Perhaps it is unavoidable and related to the custom `build.rs` step? This happens on MacOS and Windows, and it is confirmed by others working on my codebase.

Anyone else seeing this? Any suggestions?

Thanks!

kchibisov:
The docs state that it accepts `PATH`, but not like the env variable. So to make it work each `PATH` should be emmited from each `println!`.

Fixes #2657.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -28,7 +28,9 @@ mod wayland {
 
 fn main() {
     // The script doesn't depend on our code
-    println!("cargo:rerun-if-changed=build.rs:wayland_protocols");
+    println!("cargo:rerun-if-changed=build.rs");
+    println!("cargo:rerun-if-changed=wayland_protocols");
+
     // Setup cfg aliases
     cfg_aliases! {
         // Systems.


madsmtm (approved):

build.rs
```diff
@@ -28,7 +28,9 @@ mod wayland {
 
 fn main() {
     // The script doesn't depend on our code
-    println!("cargo:rerun-if-changed=build.rs:wayland_protocols");
+    println!("cargo:rerun-if-changed=build.rs");
```
oxalica, line 31: This line is unnecessary now.

From [cargo reference](https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargorerun-if-changedpath),
> If the build script inherently does not need to re-run under any circumstance, then emitting cargo:rerun-if-changed=build.rs is a simple way to prevent it from being re-run (otherwise, the default if no rerun-if instructions are emitted is to scan the entire package directory for changes). Cargo automatically handles whether or not the script itself needs to be recompiled, and of course the script will be re-run after it has been recompiled. Otherwise, specifying build.rs is redundant and unnecessary.

kchibisov: >This build.rs line is unnecessary now. It was useful because the behavior is different between "no rerun-if-changed" and "have any rerun-if-changed".

I decided to keep it since the wayland protocol stuff will be removed anyway, and if it wouldn't be you or me updating it, someone could easily miss to put that `build.rs` line back.

I hope that's clear.

Issue #2608, opened 2022-12-28
Bug: Right Mouse Button Release event never gets sent


RobWalt:
A while ago, we discovered some [weird behavior with the bevy engine when compiling to wasm](https://github.com/mvlabat/bevy_egui/issues/122).

When opening the context menu in the browser, things behaved weirdly after that. I opened an issue on the bevy UI library we are using (link above). After some analysis it turned out that the mouse state got stuck since we don't receive a `released` event when pressing the right mouse button. This in turn blocks the user from using the wasm app at all since inputs are all screwed up.

This issue might be slightly related to #2455 and #2475 but it still seemed distinct enough to open a new issue.

I'll try to get a minimal reproducible example running in the near future which only utilizes `winit`.

RobWalt:
To get a minimum reproducing example, just edit the following lines in the `examples/web` project:

```rust
// old, line 78
//log::debug("{:?}", event);

// new
if matches!(event, Event::WindowEvent {
  event: WindowEvent::MouseInput { .. },
  ..
}) {
  log::debug("{:?}", event);
}
```

Then just run the example as described in the general `README.md` via 

`cargo run-wasm --example web`

---

I recorded some testing footage. It contains the following two scenarios:

1. The context menu is opened via right click. No `Released` event is fired until we press the right mouse button anywhere else again.
2. The context menu is opened via right click. Pressing the left mouse button anywhere else fires its `Released` event which kind of screws up the pairity of `Pressed <-> Released` events for one button. I think this scenario is also responsible for the issue initially reported here

![winit](https://user-images.githubusercontent.com/26892280/209943343-8ab43de3-fcf1-4d89-b799-58d32e688772.gif)

Additional **IMPORTANT** notes:

- I noticed that the behavior above mainly happens if the right mouse button is pressed for a certain amount of time (anything longer than a very short click)
- The right mouse button release is correctly fired if I just slightly click the button for a very short time

RobWalt:
It seems like the reason could be related to the order some events are fired. Here is some further analysis:

If we press the button just slightly, we get the following order of events (oldest events are on the bottom):

![image](https://user-images.githubusercontent.com/26892280/209946962-2a880e45-eacc-4f5c-968f-33642f18e8f9.png)

It seems like entering the context menu with the cursor counts as a `CursorLeft` event. If we just click the right mouse button, the cursor hasn't left the window yet, leading to correct handling of the `Release` event.

[...]

Issue #2475, opened 2022-09-07
When one MouseInput button is pressed, it blocks the others (web / wasm)


cybersoulK:
I dig in the code, and i narrowed it down to the use of a "pointer" instead of a "mouse"
When the left click is pressed (fine), and then right click is pressed (missing event), and left is released (missing event), the right release will never be called again

2 downsides
- it blocks the other buttons (including wheel)
- release events will never be called again

so for games it has huge impact, the action that was triggered by the mouse input can get "stuck" (because of no release event), and it does block an action from occurring entirely

Leaving the developer very confused (like me for hours xd)

JAD3N:
A hacky workaround you can use is to remove `PointerEvent` from window if touch/pen events aren't important to you.

daxpedda:
Duplicate of https://github.com/rust-windowing/winit/issues/2455.

PR #2664, opened 2023-01-31
Fixes mouse release on wasm

cybersoulK:
Temporary fix for mouse issues in games in wasm platform: #2475 #2608 
This has worked very well for me for 5 months, and we should use it until we have more elaborate implementations


daxpedda:
Fixed by #2662 as far as I can tell.

PR #2665, opened 2023-01-31
Windows: Fix Alt key press entering menu loop

### LINKED ISSUE #2661: Windows, alt key press hangs the event loop
On Windows, pressing `LAlt` hangs the event loop, the `MainEventsCleared` event only happens after the window is clicked, or any other new window event.

This happens since #2521, I believe the app is entering the "menu loop", even though the app does not have a menu.

SamRodri:

Fixes #2661 by detecting if the window has a native menu before delegating to the modal menu loop.

### DIFF
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -50,10 +50,10 @@ use windows_sys::Win32::{
         },
         WindowsAndMessaging::{
             CreateWindowExW, DefWindowProcW, DestroyWindow, DispatchMessageW, GetCursorPos,
-            GetMessageW, LoadCursorW, MsgWaitForMultipleObjectsEx, PeekMessageW, PostMessageW,
-            PostThreadMessageW, RegisterClassExW, RegisterWindowMessageA, SetCursor, SetWindowPos,
-            TranslateMessage, CREATESTRUCTW, GIDC_ARRIVAL, GIDC_REMOVAL, GWL_STYLE, GWL_USERDATA,
-            HTCAPTION, HTCLIENT, MINMAXINFO, MNC_CLOSE, MSG, MWMO_INPUTAVAILABLE,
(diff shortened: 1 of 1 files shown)


SamRodri:
I'm not sure about this fix, I never used the native menu thing and don't know how to create one to test it, the `GetMenu` does return NULL in my app.

Also I don't understand why Windows starts a menu loop when there is no menu, I'm actually assuming that's whats happening based on Stackoverflow.

SamRodri:
Ok, I tested with a menu created like this:

```rust
unsafe {
    let hwnd = window.hwnd();
    let menu = CreateMenu();
    AppendMenuA(menu, MF_STRING, 0, b"Test".as_ptr());
    SetMenu(hwnd, menu);
}
```

It seems to get the focus.

kchibisov:
r @amrbashir

amrbashir (approved):
I am not quite familiar with the keyboard handling, but I am fine with this change and can't see an immediate regression since winit used to always return `0` for this message.

I'd better wait for @msiglreith confirmation.

msiglreith:
Could you provide some more information or an example to reproduce? I can't reproduce the bug from the issue mentioned right now.

Also, please add some more comments in the code why this line was added.

SamRodri:
Modify the example window  to only print relevant events:

```rust

use winit::{
    event::{Event, WindowEvent},
    event_loop::EventLoop,
    window::WindowBuilder,
};

fn main() {

        .build(&event_loop)
        .unwrap();

        control_flow.set_wait();

        match event {
            Event::WindowEvent { event, window_id } => match event {
                WindowEvent::CloseRequested if window_id == window.id() => control_flow.set_exit(),
                WindowEvent::KeyboardInput { input, .. } => println!("{:?}", input),
                _ => {}
            },
            Event::MainEventsCleared => {
                println!("Cleared");
                window.request_redraw();
            }
            _ => (),
        }
    });
}

```

Run it, it will not print "Cleared" after  pressing and releasing `LAlt`.

Will add comments.

msiglreith (approved):
thanks

PR #2666, opened 2023-02-01
On macOS, fix `Ime::Commit` persisting

kchibisov:
This commit clears the currently marked text on `Ime::Commit`, so normal `ReceivedCharacter` input can continue.

--

@madsmtm I was thinking of using `unmarkText` but I don't really want extra `Preedit` after the `Commit`. Do I need that extra stuff with context that is present in `unmarkText`?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- On macOS, fixed `Ime::Commit` persisting for all input after interacting with `Ime`.
 - On macOS, added `WindowExtMacOS::option_as_alt` and `WindowExtMacOS::set_option_as_alt`.
 - On Windows, fix window size for maximized, undecorated windows.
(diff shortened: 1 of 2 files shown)


madsmtm (commented):
Hmm yeah, I think I can see the issue, but this PR only creates a new issue: E.g. using a US keyboard, if I press <kbd>Option + E</kbd>, a preedit is started with `´`, but if I then press <kbd>B</kbd>, the `´` is committed, but the <kbd>B</kbd> keypress is lost.

I honestly don't know how to fix this, I feel so out of my league dealing with IME :/

kchibisov:
Let me recheck, since for me it was working as it should.

kchibisov:
@madsmtm now it should work, I guess.

madsmtm (approved):
Linking a bit of discussion on Matrix: https://matrix.to/#/!DGpLzJTRzBDTwZiogk:matrix.org/$s7UBzLRXwJTYxEO92KxL_cva5IExvJX-nmRnbXYIII0?via=libera.chat&via=matrix.org&via=mozilla.org

PR #2669, opened 2023-02-02
Update Cargo.toml example to use the current version

tronical:


tronical:
Oops, I see that this is fixed in the tag already - so not needed for master. Sorry for the noise :)

PR #2672, opened 2023-02-04
m(windows): Name the waiter thread

notgull:

It's easier to debug if the waiting thread has an explicit name.

Also, quick question, is there any particular reason why we use the waiter thread instead of just `WM_TIMER`? I'm guessing that `WM_TIMER` doesn't play well with `timeBeginPeriod`.


msiglreith (approved):
thanks

> Also, quick question, is there any particular reason why we use the waiter thread instead of just WM_TIMER? I'm guessing that WM_TIMER doesn't play well with timeBeginPeriod.

I honestly don't know why it's exactly implemented like this, was before my time - might be some conflicts with other parts or just historically

PR #2673, opened 2023-02-05
web: Fix Enter key being handled as character E

johanhelsing:
When with_prevent_default is set to false, browsers will send keypress events for the enter key, with event.key set to Enter.

Ignore it by checking the length of the key string.

- ~~[] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior~~
- ~~[ ] Created or updated an example program if it would help users understand this functionality~~
- ~~[ ] Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented~~

### DIFF
--- a/src/platform_impl/web/web_sys/canvas.rs
+++ b/src/platform_impl/web/web_sys/canvas.rs
@@ -239,6 +239,11 @@ impl Canvas {
                     event.prevent_default();
                 }
 
+                if event.key().len() > 1 {
+                    // ignore keypress for "Enter" key etc.
+                    return;
+                }
+
                 handler(event::codepoint(&event));
             },
         ));
--- a/src/platform_impl/web/web_sys/canvas.rs
+++ b/src/platform_impl/web/web_sys/canvas.rs
@@ -239,7 +239,7 @@ impl Canvas {
                     event.prevent_default();
                 }
 
-                if event.key().len() > 1 {
+                if event.key().chars().nth(1).is_some() {
                     // ignore keypress for "Enter" key etc.
                     return;
                 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- On Web, fix Enter key being handled as the character E if `with_prevent_default` is false.
+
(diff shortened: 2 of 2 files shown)


johanhelsing:
Minimal test case: https://github.com/johanhelsing/winit/pull/new/with-prevent-default-receive-characters

Run the web example and press enter. Observe received character "e" in the log.

src/platform_impl/web/web_sys/canvas.rs
```diff
@@ -239,6 +239,11 @@ impl Canvas {
                     event.prevent_default();
                 }
 
+                if event.key().len() > 1 {
+                    // ignore keypress for "Enter" key etc.
+                    return;
+                }
```
maroider, line 245: Wouldn't this also throw out any "single character" key strings encoded as more than one byte? I suspect the only "correct" way of solving this might be to special-case all the "non-character" key strings specified in the [`uievents-key`](https://w3c.github.io/uievents-key/) spec.

johanhelsing: Maybe? The actual handler just does `event.chars().next().unwrap()`. So perhaps we should just check `event.key.chars().len()` here?

johanhelsing: Now it should support any multi-byte single character strings as well.

If there are multi-*character* strings however, those they will now be ignored, while previously, we would just take the first and ignore the rest.

johanhelsing:
@maroider are you happy with this as is, should I rebase to fix the conflicts with the changelog?

johanhelsing:
@daxpedda, @notgull, @ryanisaacg you are [listed as contributors/experts for wasm](https://github.com/rust-windowing/winit/wiki/Testers-and-Contributors)

Do you either of you have the time to review this, and/or know who I could involve to get this PR moving?

daxpedda:
There is currently a merge freeze until we are done with https://github.com/rust-windowing/winit/pull/2789, I will review this issue if it's still one afterwards.

kchibisov:
Mentioned patch merged, needs re-evaluation.

daxpedda:
This can be closed, with #2662 this isn't an issue anymore, just tested it.

PR #2675, opened 2023-02-06
Run examples in CI

notgull:

This PR sets up the example so that they can be ran in the CI using X11 under `xvfb`, essentially allowing us to run them as integration tests. I accomplish this by spawning a thread that sends a user event to the event loop that makes it time out.

There are ways of running the examples of other platforms, but I figured I'd start with X11.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -119,3 +119,16 @@ jobs:
         !contains(matrix.platform.target, 'redox') &&
         matrix.rust_version != '1.60.0'
       run: cargo $CMD test --verbose --target ${{ matrix.platform.target }} $OPTIONS --features serde,$FEATURES
+
+    # derived from  https://github.com/psychon/x11rb/blob/15153e514a4e3e4c116279f76ab1c3635abc05e5/.github/workflows/CI.yml#L141-L159
+    - name: Run examples
+      shell: bash
+      if: >
+        contains(matrix.platform.target, 'linux') &&
+        contains(matrix.platform.features, 'x11')
+      run: |
+          for example in examples/*.rs; do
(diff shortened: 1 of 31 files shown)


kchibisov:
Is the point to ensure that it starts without crashing? If so it might not be that useful. Though, I'm not even sure how to write tests for stuff like that. Usually for libraries like winit you want to test what events you produce based on server events.

So you can ensure that event ordering, inner values, and such works as expected. From what I can tell, only on Wayland it could be possible, but it'll take so much effort to add meaning that it's not really feasible to maintain(you simply need to implement fake compositor sending fake events).

notgull:
> Is the point to ensure that it starts without crashing? If so it might not be that useful.

I figure that, for examples, it can't hurt to at least double check that everything starts up and shuts down okay. As far as I can tell, there's nothing in the `winit` codebase that checks this on a CI run. 

> Though, I'm not even sure how to write tests for stuff like that.

This would probably require a more complicated testing setup. I don't really want to work on anything more complicated than this before #2614 lands, since Xlib is a pain and a half to successfully integration test.

> From what I can tell, only on Wayland it could be possible, but it'll take so much effort to add meaning that it's not really feasible to maintain(you simply need to implement fake compositor sending fake events).

Hmm, on Wayland, is there not an equivalent to `xvfb` yet?

kchibisov:
>Hmm, on Wayland, is there not an equivalent to xvfb yet?

You can run compositors in headless mode without issues, like sway :p. I was talking about sending specific input from compositor and testing the replies, like wayland-rs is doing, but they have much simpler task.

notgull:
Putting in the effort to rebase this would take more emotional investment than I currently do in this PR.

PR #2676, opened 2023-02-06
Update wayland-rs to 0.30.0

### LINKED ISSUE #725: Can't resize window without decorations

### LINKED ISSUE #1760: "internal error: entered unreachable code" when using Wayland backend
Attempting to create a Window on the Wayland backend on my system results in the following:

```
thread 'main' panicked at 'internal error: entered unreachable code', /home/slabity/.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.23.0/src/platform_impl/linux/wayland/window/mod.rs:218:77
[wayland-client error] A handler for wl_keyboard panicked.
```

Which seems to be coming from this:

```rust
// To make our window usable for drawing right away we must `ack` a `configure`
// from the server, the acking part here is done by SCTK window frame, so we just
// need to sync with server so it'll be done automatically for us.
event_loop_handle.with_source(&wayland_source, |event_queue| {
    let event_queue = event_queue.queue();
    let _ = event_queue.sync_roundtrip(&mut *winit_state, |_, _, _| unreachable!());
});
```

Can anyone explain what this little piece of code is doing and what would cause the it to be reachable?

### LINKED ISSUE #2128: Update wayland-client to 0.30
wayland-client below 0.30 is fundamentally broken in that updating your compositor can cause wayland-client applications to abort immediately on start. This is because wayland-client below 0.30 cannot handle unknown shm formats but compositors are allowed to add new formats at any time.

### LINKED ISSUE #2164: xdg_toplevel surface initialization is not optimal with kwin_wayland
Compositors such as kwin need the client to set the app id, title, size hints, preferred decoration mode, and other things that can influence the size in the first configure event before the surface commit. With applications such as alacritty, it appears like that's not the case.

```
[2699002.481]  -> xdg_wm_base@22.get_xdg_surface(new id xdg_surface@24, wl_surface@21)
[2699002.487]  -> xdg_surface@24.get_toplevel(new id xdg_toplevel@25)
[2699002.492]  -> wl_surface@21.commit()
[2699002.495]  -> xdg_toplevel@25.set_min_size(2, 1)
[2699002.500]  -> xdg_surface@24.set_window_geometry(0, 0, 800, 600)
[2699002.508]  -> wl_compositor@4.create_surface(new id wl_surface@26)
[2699002.511]  -> wl_seat@7.get_pointer(new id wl_pointer@27)
[2699002.517]  -> zxdg_decoration_manager_v1@5.get_toplevel_decoration(new id zxdg_toplevel_decoration_v1@28, xdg_toplevel@25)
[2699002.526]  -> zxdg_toplevel_decoration_v1@28.unset_mode()
[2699002.528]  -> wl_surface@21.commit()
[2699002.531]  -> xdg_toplevel@25.set_min_size(2, 1)
[2699002.535]  -> xdg_toplevel@25.set_max_size(0, 0)
[2699002.539]  -> xdg_toplevel@25.set_app_id("Alacritty")
[2699002.542]  -> xdg_toplevel@25.set_min_size(2, 1)
[2699002.546]  -> xdg_toplevel@25.set_max_size(0, 0)
[2699002.551]  -> xdg_toplevel@25.set_title("Alacritty")
```

The relevant properties are set after the first surface commit, which can result in some kwin features not working as expected when using Alacritty, for example window rules. Another minor issue is that xdg_surface.set_window_geometry() is called without a buffer attached.

With compositors such as kwin, the ideal xdg_toplevel surface initialization looks as follows:

* create the xdg_toplevel surface
* set size hints, app id, call xdg_toplevel.set_maximized() or xdg_toplevel.set_fullscreen() if the window needs to be shown maximized or fullscreen, respectively
* create the xdg_toplevel_decoration if needed and set the preferred decoration mode
* wl_surface_commit()

After that, the compositor will respond with a configure event and the application can start painting frames \o/.

The desired initialization sequence should not be controversial because the client provides the compositor as much as possible information so it can do the right thing. From the xdg-shell protocol

```
      After creating a role-specific object and setting it up, the client must
      perform an initial commit without any buffer attached. The compositor
      will reply with an xdg_surface.configure event. The client must
      acknowledge it and is then allowed to attach a buffer to map the surface.
```

### LINKED ISSUE #2560: winit incorrectly assumes `xdg_surface.configure` event always arrive in one roundtrip.
The spec does not make that guarantee, and states the client is not allowed to attach a buffer until a configure event is received. However winit would start attaching buffers anyway after one roundtrip, whether a configure is received or not.

I only find this out because I have a not-fully-implemented compositor. I assume most of the existing compositors would send out configure immediately.

kchibisov:
This update rewrites the winit's Wayland backend using new wayland-rs 0.30 API. This fixes long standing issue with the forward compatibility of the wayland backend, meaning that future updates to the wayland protocol won't break rust code anymore. like it was before when adding new shm/enum variants into the protocol.

Fixes #2560.
Fixes #2164.
Fixes #2128.
Fixes #1760.
Fixes #725.

--

# TBD:
  - [x] Fractional scaling
  - [x] Client side decorations
  - [x] Testing on compositors other than sway
  - [ ] SCTK release
  - [x] CHANGELOG update
  - [x] Keyboard sometimes blast repeats


halotroop2288 (approved):

kchibisov:
The sctk-adwaita decorations are back now as well. So only waiting for upstream patches to propagate properly.

torokati44:
Can confirm this will fix #2753!

kchibisov:
>EDIT: In fact, the first commit, fe73b10, is enough in itself for it to be fixed.

you know, it's +3K -3K commit in the end, but I know that resize on this branch is sane.

torokati44:
> you know, it's +3K -3K commit in the end

Well, yeah, haha... :sweat_smile: It was just easier to test the first commit in itself with the current set of dependencies - with the final commit of the PR I had issues related to AdwaitaFrame and SCTK.

kchibisov:
>Well, yeah, haha... sweat_smile It was just easier to test the first commit in itself with the current set of dependencies - with the final commit of the PR I had issues related to AdwaitaFrame and SCTK.

Like it doesn't build or something like that? That could be the case, given that I haven't update the branch for a while.

kchibisov:
It should build now for what it matters.

torokati44:
Yes, it works with the latest commit now, thanks!

src/platform_impl/linux/wayland/event_loop/mod.rs
```diff
@@ -62,12 +62,16 @@ impl<T: 'static> EventLoop<T> {
...
         let queue_handle = event_queue.handle();
 
         let event_loop = calloop::EventLoop::<WinitState>::try_new()?;
 
-        let winit_state = WinitState::new(&globals, &queue_handle, event_loop.handle())?;
+        let mut winit_state = WinitState::new(&globals, &queue_handle, event_loop.handle())?;
+
+        // NOTE: we must roundtrip with the server to ensure that the state is initialized
+        // correctly.
+        event_queue.roundtrip(&mut winit_state)?;
```
rib, line 74: Could you consider deferring the `NewEvents(Init)` and `Resumed` events until this round trip complete normally via the event loop?

This kind of synchronous round trip, similar to the round trip in `Window::new` seems like a kind of code smell that doesn't seem like it should be necessary ideally.

It could be good if the comment clarified, more specifically, what response is being waited for and why the event loop would be in an inconsistent state if it doesn't wait here.

kchibisov: A roundtrip is basically a barrier for the server requests. It can't block indefinitely, because the `display.sync` is being sent under the hood.

Some compositors just have issues with "really async startup" here, so it's better to force a synchronization.

The one in window is a bit different though, it's to ensure that the `Window` was configured, given that you can't do anything with unconfigured one.

rib: That makes it sound like it's a workaround for fragile, maybe buggy, compositors.

That sounds potentially fair enough if that's the case but it still seems a shame that the comment doesn't point to a specific, verifiable, issue, like a workaround for a particular compositor, otherwise it sounds like a mysterious workaround that will end up being maintained indefinitely because noone can really verify what it fixes.

If it's more because the backend wants the response to fill in some `EventLoop` state (so it's a winit backend issue that means the event loop is configured asynchronously) (which the comment currently seems to imply) then I'd still wonder whether this wait could happen via the normal event loop, similar to how an `NSApp` gets 'launched' asynchronously, or an Android app resumes asynchronously and `NewEvents(Init)` will only be sent after the event loop has finished being set up asynchronously.

If possible it seems worthwhile to aim for only one standard place (the event loop proper) where the backend deals with I/O and the Wayland backend isn't the only backend that effectively wants to configure the `EventLoop` asynchronously.

kchibisov: >That makes it sound like it's a workaround for fragile, maybe buggy, compositors.

Yes, And I also carry workarounds for GNOME for like 2 years, because it doesn't want fix spec violation. 

>That sounds potentially fair enough if that's the case but it still seems a shame that the comment doesn't point to a specific, verifiable, issue, like a workaround for a particular compositor, otherwise it sounds like a mysterious workaround that will end up being maintained indefinitely because noone can really verify what it fixes.

I'd add more clarifications for the places, where not really obvious, which were ported from the old code.

>If it's more because the backend wants the response to fill in some EventLoop state (so it's a winit backend issue that means the event loop is configured asynchronously) (which the comment currently seems to imply) then I'd still wonder whether this wait could happen via the normal event loop, similar to how an NSApp gets 'launched' asynchronously, or an Android app resumes asynchronously and NewEvents(Init) will only be sent after the event loop has finished being set up asynchronously.

This roundtrip is not for winit, but for compositor state, nothing winit should do here, it's required only on initial loop creation, because you bind globals only once.

>) then I'd still wonder whether this wait could happen via the normal event loop, similar to how an NSApp gets 'launched' asynchronously, or an Android app resumes asynchronously and NewEvents(Init) will only be sent after the event loop has finished being set up asynchronously.

Be aware that it's a short wait (unless your compositor is dead broken) and it's required once per event loop creation, so the `::new` is the only place for it. Nothing else is required here, it can't even send events for the user.

>If possible it seems worthwhile to aim for only one standard place (the event loop proper) where the backend deals with I/O and the Wayland backend isn't the only backend that effectively wants to configure the EventLoop asynchronously.

We don't configure anything in the winit's event loop it's already done. The roundtrip is to syncchorinize with the compositor, so it won't get confused for whatever reason it could get confused, there's no state broken state in winit before/after roundtrip.

The bug it was solving was purely to "compositor for whatever reason still things that the object is not alive", because it's likely a race in it, so I've decided for a more conservative startup, which was even before(but was hidden inside the macros).

PR #2677, opened 2023-02-08
Implement `HasRawDisplayHandle` for `EventLoop`

SludgePhD:

The trait is already implemented for `EventLoopWindowTarget`, which `EventLoop` derefs to, but that is not enough when using an API that takes a `T: HasRawDisplayHandle + 'static` (because the API takes ownership of the display handle, in an attempt to offer a safer API to the user than what crates typically do). Also implementing the trait for `EventLoop` itself solves this problem.

(admittedly the use cases for this are rather limited – consuming the `EventLoop` like this makes it impossible to actually handle events, and so makes windows useless too, but it *is* handy when you need a display handle without actually wanting to draw anything; in my specific use case, in order to open a connection to VA-API)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- Implement `HasRawDisplayHandle` for `EventLoop`.
+
 # 0.28.1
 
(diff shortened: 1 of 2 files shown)


madsmtm (approved):
Sure, this makes sense - thanks for the contribution!

PR #2679, opened 2023-02-08
m(windows): Decentralize event handling

notgull:

In the Windows backend, the handling of event messages is handled by a single, 1300 line long match statement. In the interest of modularization, I've split this handler up into several files. This is done by using David Tolnay's famous `inventory` crate to create a decentralized registry containing all of the plugin handlers, and then collecting that into a hash map.

There may be some minor overhead from casting `WindowData` to `dyn GenericWindowData,` but I doubt it's significant.

~~Blocked on #2662~~ Blocked on #2767


notgull:
Crap, I forgot that `inventory` has an MSRV of 1.62. That's about seven months old, so it should be fine, right?

madsmtm:
> MSRV of 1.62. That's about seven months old

Yeah, that's fine

notgull:
Needs a rework for event loop 3.0

PR #2680, opened 2023-02-09
Add Platform enumeration

dhardy:

Add `enum Platform`, which compiles down to a single variant on most targets (two on Unixes).

Main motivation: make it easy to detect the platform without requiring things like this:
```rust
        #[cfg(any(
            target_os = "linux",
            target_os = "dragonfly",
            target_os = "freebsd",
            target_os = "netbsd",
            target_os = "openbsd"
        ))]
        {
            use winit::platform::unix::EventLoopWindowTargetExtUnix;
            if elwt.is_wayland() {
                // ...
            }
        }
```
Now it's just:
```rust
    if elwt.platform().is_wayland() {
        // ...
    }
```
with the method compile-time reducing to `false` on most platforms.

Disadvantage of this approach: there's a little more to keep in sync with the cfg-foo in `platform_impl`.

Possibly this could be tested in CI, though it'd probably just result in *another* copy of the cfg-foo.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- Add `fn EventLoopWindowTarget::platform() -> Platform`
+
 # 0.28.1
 
 - On Wayland, fix crash when dropping a window in multi-window setup.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,6 +54,7 @@ instant = { version = "0.1", features = ["wasm-bindgen"] }
 log = "0.4"
 mint = { version = "0.5.6", optional = true }
 once_cell = "1.12"
+cfg-if = "1.0.0"
 raw_window_handle = { package = "raw-window-handle", version = "0.5" }
 serde = { version = "1", optional = true, features = ["serde_derive"] }
 
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -15,7 +15,7 @@ use instant::{Duration, Instant};
 use once_cell::sync::OnceCell;
 use raw_window_handle::{HasRawDisplayHandle, RawDisplayHandle};
 
-use crate::{event::Event, monitor::MonitorHandle, platform_impl};
+use crate::{event::Event, monitor::MonitorHandle, platform::Platform, platform_impl};
 
 /// Provides a way to retrieve events from the system and from the windows that were registered to
 /// the events loop.
@@ -358,6 +358,11 @@ impl<T> EventLoopWindowTarget<T> {
     pub fn set_device_event_filter(&self, _filter: DeviceEventFilter) {
         #[cfg(any(x11_platform, wayland_platform, windows))]
         self.p.set_device_event_filter(_filter);
+    }
+
+    /// Return detected platform
+    pub fn platform(&self) -> Platform {
+        self.p.platform()
     }
 }
 
--- a/src/platform/mod.rs
+++ b/src/platform/mod.rs
@@ -41,3 +41,130 @@ pub mod x11;
     orbital_platform
 ))]
 pub mod run_return;
+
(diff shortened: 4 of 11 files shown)


madsmtm (commented):
I don't really see this being that useful? If you're doing something platform-specific, you almost always have to import one of the extension traits anyhow, and then you'll need the cfgs anyhow.

dhardy:
In KAS I have two pieces of Wayland-specific code which don't require any platform-specific imports: (1) to use toolkit decorations by default (because winit doesn't support the KDE Wayland extension for server-side decorations yet) and (2) to use slightly different logic for the initial window size.

The other platforms are here for completeness (I haven't found a need for them yet).

kchibisov:
>because winit doesn't support the KDE Wayland extension for server-side decorations yet

What do you mean by that? We support all available server side decorations protocols is there a 3rd one. Alacritty on KDE is using server side decorations for sure with Wayland.

>to use slightly different logic for the initial window size.

I think I might actually change this, so you won't have to deal with that. Could you point to the original issue/example wrt that, we have more control now wrt internals on Wayland, so I might just fix it while at it?

dhardy:
> We support xdg decorations protocol

Guess I should test better; didn't see this noted anywhere. But how do tell whether server-side decorations are supported?

> I think I might actually change this

But change what? Hint something about platform behaviour in `WindowBuilder`? On X11 the scale factor is known before window creation and size is specified in physical pixels; taking advantage of this allows better control than specifying initial size in logical pixels (required for Wayland). Yeah I know KAS is weird for doing *better* than linear scaling.

kchibisov:
>Guess I should test better; didn't see this noted anywhere. But how do I tell whether server-side decorations are supported?

Winit will use server side decorations by default if they are available, and use client side if they don't or if the server forces into client side.

You can test with `WAYLAND_DEBUG=1` and grepping `xdg_decoration` in your log.

>But change what? Hint something about platform behaviour in WindowBuilder? On X11 the scale factor is known before window creation and size is specified in physical pixels; taking advantage of this allows better control than specifying initial size in logical pixels (required for Wayland). Yeah I know KAS is weird for doing better than linear scaling.

This is not the case for the mainline Wayland anymore though, we know the scale ahead of time with fractional scaling v1 and preferred_buffer_scale events.

dhardy:
> Winit will use server side decorations by default if they are available, and use client side if they don't or if the server forces into client side.

Yeah... the goal was to let my toolkit handle drawing decorations in case server-side ones are not supported (thus avoiding two independent text rendering systems as well as the requirement for cmake).

So *how does my toolkit tell whether server side decorations are available*?

> we know the scale ahead of time with fractional scaling v1 and preferred_buffer_scale events.

1. I still don't see a way to retrieve the scale factor *before* creating a window (except by guessing it will be placed on the first monitor). The [protocol](https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/220) supports a `wl_surface`, which doesn't even exist before window creation.
2. Wayland still doesn't support specifying window size in physical pixels.

kchibisov (changes requested):
I think what you suggest is what `RawDisplayHandle` is for, it's already covers all you enum and you get it the same way. The only issue is that you actually do need an `event_loop_window_target` or `event_loop`, but I don't think it's really an issue?

dhardy:
I would be happy with just moving `is_wayland` and `is_x11` to `EventLoopWindowTarget`. They would need an impl for when the corresponding feature is disabled, but that's a trivial impl: `false`.

Otherwise just close this; no one else appears to see any utility in it.

kchibisov:
@dhardy but you can do what you want with the proposed enum with `matches!(evlp.raw_display_handle, RawDisplayHandle::Wayland(_))` instead, this won't add any cfg and will do exactly what you want?

I'll close this due to suggested options being possible already. Though, the only issue is that you have to import raw-window-handle(but it's in deps tree anyway...).

PR #2687, opened 2023-02-17
Add support for `Window::theme` on the web

tronical:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - Implement `HasRawDisplayHandle` for `EventLoop`.
 - On macOS, set resize increments only for live resizes.
 - On Wayland, fix rare crash on DPI change
+- Web: Added support for `Window::theme`.
 
 # 0.28.1
 
--- a/src/platform_impl/web/window.rs
+++ b/src/platform_impl/web/window.rs
@@ -409,7 +409,20 @@ impl Window {
 
     #[inline]
     pub fn theme(&self) -> Option<Theme> {
-        None
+        web_sys::window()
+            .and_then(|window| {
+                window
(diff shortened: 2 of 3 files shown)


ogoffart:
I think this change is useful as is, so IMHO shouldn't prevent to be merged.  But it could also emit the ThemeChanged event when the media query changes (using for example `MediaQueryList::set_onchange`)

tronical:
> I think this change is useful as is, so IMHO shouldn't prevent to be merged. But it could also emit the ThemeChanged event when the media query changes (using for example `MediaQueryList::set_onchange`)

Winit already implements the part of sending `ThemeChanged` when targeting the web. It's just missing the getter for the current value.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 - Implement `HasRawDisplayHandle` for `EventLoop`.
 - On macOS, set resize increments only for live resizes.
+- Added support for `Window::theme` to support per-window dark/light/system theme configuration on web.
```
madsmtm, line 13: Nit: Write the platform first

```suggestion
- Web: Added support for `Window::theme`.
```

madsmtm (approved):
Only a minor nit with the changelog, otherwise looks fine

tronical:
Good point, applied:). Feel free to choose merge with squash. (Or I can also force push a squash if you prefer)

madsmtm:
Don't have a preference for how you do the commits, they'll be squashed once merged anyhow ;)

I'll need you to fix the conflicting changelog though, since I don't have write access to your fork.

tronical:
Oops, not sure why that permission isn't there. I rebased, squashed, and resolved the conflict.

madsmtm:
Thanks!

Issue #1118, opened 2019-08-24
SimpleFullscreen on macOS does not get risized when moving window across screens


chrisduerr:
This was reported to Alacritty in https://github.com/jwilm/alacritty/issues/2741.

It seems like the simple fullscreen mode on macOS always keeps the window at the same size, even after it is moved between screens.

The expectation here is that when moving the window, it should take all of the space available on the new monitor, instead of sticking to the old size.

Toggling simple fullscreen after the window has been moved fixes this issue.

aleksijuvani:
It looks like the window bounds are only set upon the initial `set_simple_fullscreen` call. If the window is in simple fullscreen mode, we should respond to [`windowDidChangeScreen`](https://developer.apple.com/documentation/appkit/nswindowdelegate/1419267-windowdidchangescreen) and change the window bounds to fit the new screen. It seems weird to me for the user to have to "move" the window to put it on another screen, however. We should change `set_simple_fullscreen` to take an `Option<MonitorHandle>` instead of a boolean. Additionally, we should track screen resolution changes with [`didChangeScreenParametersNotification`](https://developer.apple.com/documentation/appkit/nsapplication/1428749-didchangescreenparametersnotific).

aleksijuvani:
That said, I would strongly prefer applications to not use the simple fullscreen functionality. The simple fullscreen mode prevents the user from using some gestures common to fullscreen applications (i.e. three-finger swipe left or right to go back to the desktop). This is because it is not in fact a fullscreen window, but a regular window with its decorations hidden. It's generally just not what users are expecting from a fullscreen application. I would love to hear about the use-cases for it, because I'm confused as to why anyone would want this behaviour.

mirosval:
@aleksijuvani Hi, I'm the original reporter from the Alacritty bug. I prefer this mode with Alacritty, as opposed to the Native Fullscreen, because its faster and more convenient to use.

1. Native fullscreen is like a separate desktop, it is a little slower to switch to it, even if you have animations disabled. Simple fullscreen is instant
2. I use Hammerspoon window management so that I can quickly resize the terminal to half the screen if I need to copy something from the browser for example
3. When I open iMessage or Slack, it opens on top of the terminal, instead of switching the desktop entirely.

I used the same setup with iTerm2, where I would additionally use the `cmd+u` transparency toggling on occasion to copy something from the browser underneath without switching apps. None of this is possible when you use the native fullscreen.

juanibiapina:
[...]

PR #2691, opened 2023-02-20
[macOS] resize simple fullscreen windows on windowDidChangeScreen

dylanscott:
I'm an Alacritty user and ran into #1118 (via alacritty/alacritty#2741). It's tagged with "good first issue" and someone helpfully left some details on the AppKit side of things so I thought I'd take a crack at it. I'm kind of a Rust newb (and even bigger AppKit/ObjC newb) so happy for any feedback if this should be tackled in a different way.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On Wayland, fix rounding issues when doing resize.
 - On macOS, fix wrong focused state on startup.
 - On Windows, fix crash on setting taskbar when using Visual Studio debugger.
+- On macOS, resize simple fullscreen windows on windowDidChangeScreen events.
 
 # 0.28.1
 
--- a/src/platform_impl/macos/window.rs
+++ b/src/platform_impl/macos/window.rs
@@ -150,7 +150,7 @@ pub struct SharedState {
     pub(crate) target_fullscreen: Option<Option<Fullscreen>>,
     pub maximized: bool,
     pub standard_frame: Option<NSRect>,
-    is_simple_fullscreen: bool,
+    pub(crate) is_simple_fullscreen: bool,
     pub saved_style: Option<NSWindowStyleMask>,
     /// Presentation options saved before entering `set_simple_fullscreen`, and
     /// restored upon exiting it. Also used when transitioning from Borderless to
@@ -1317,6 +1317,10 @@ impl WindowExtMacOS for WinitWindow {
             // Tell our window's state that we're in fullscreen
             shared_state_lock.is_simple_fullscreen = true;
 
+            // Drop shared state lock before calling app.setPresentationOptions, because
+            // it will call our windowDidChangeScreen listener which reacquires the lock
+            drop(shared_state_lock);
+
             // Simulate pre-Lion fullscreen by hiding the dock and menu bar
             let presentation_options =
                 NSApplicationPresentationOptions::NSApplicationPresentationAutoHideDock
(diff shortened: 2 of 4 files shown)


src/platform_impl/macos/window.rs
```diff
@@ -1339,11 +1340,14 @@ impl WindowExtMacOS for WinitWindow {
             self.set_style_mask_sync(new_mask);
             shared_state_lock.is_simple_fullscreen = false;
 
-            if let Some(presentation_opts) = shared_state_lock.save_presentation_opts {
+            let save_presentation_opts = shared_state_lock.save_presentation_opts;
+            let frame = shared_state_lock.saved_standard_frame();
+            drop(shared_state_lock);
```
dylanscott, line 1345: when I first added the `windowDidChangeScreen` listener I ran into an issue where we need to acquire the window shared state mutex to check `is_simple_fullscreen`. But calling `app.setPresentationOptions` (which we also do in the above branch / when toggling simple fullscreen on or off) that results in the `windowDidChangeScreen` listener being called synchronously on the same (main) thread.

I wasn't sure the best way to handle this but there were a few other places in this file where the shared state lock is explicitly dropped like this, presumably for similar reasons.

src/platform_impl/macos/window_delegate.rs
```diff
@@ -409,6 +409,17 @@ declare_class!(
                 self.queue_event(WindowEvent::ThemeChanged(theme));
             }
         }
+
+        #[sel(windowDidChangeScreen:)]
+        fn window_did_change_screen(&self, _: Option<&Object>) {
+            trace_scope!("windowDidChangeScreen:");
+            let shared_state = self.window.lock_shared_state("window_did_change_screen");
+            if shared_state.is_simple_fullscreen {
```
kchibisov, line 417: That way you'll drop the lock right away.

```suggestion
            let is_simple_fullscreen = self.window.lock_shared_state("window_did_change_screen").is_simple_fullscreen;
            if is_simple_fullscreen {
```

src/platform_impl/macos/window.rs
```diff
@@ -1314,6 +1314,7 @@ impl WindowExtMacOS for WinitWindow {
 
             // Tell our window's state that we're in fullscreen
             shared_state_lock.is_simple_fullscreen = true;
+            drop(shared_state_lock);
```
kchibisov, line 1318: Add a comment on why you release the lock here.

src/platform_impl/macos/window.rs
kchibisov, line 1345: Will need a comment explaining what you're trying to do here.

kchibisov (changes requested):
Will also need a `CHANGELOG` entry for that.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On Wayland, fix rare crash on DPI change
 - Web: Added support for `Window::theme`.
 - On Wayland, fix rounding issues when doing resize.
+- On macOS, resize simple fullscreen windows on windowDidChangeScreen events
```
kchibisov, line 16: Could you revert arbitrary changes you've done to this file besides this line? Thx.

kchibisov (changes requested):

CHANGELOG.md
dylanscott: sorry didn't notice my editor did that

kchibisov (approved):

kchibisov:
Thanks.

PR #2693, opened 2023-02-23
feat: Cross-platform popup windows

notgull:

This PR adds a cross-platform way of creating popup windows; basically, what `with_owner_window` does for Windows, but cross platform. This should allow us to create popup windows on all platforms that support them.

I've only implemented Windows and X11 for now, since those are the platforms I'm most familiar with. We should also be able to add some of the other platforms.

~~Blocked on #2614~~

### DIFF
--- a/src/platform_impl/linux/x11/window.rs
+++ b/src/platform_impl/linux/x11/window.rs
@@ -454,6 +454,21 @@ impl UnownedWindow {
             );
             leap!(result).ignore_error();
 
+            // Set the transient parent if we don't have a parent already.
+            if event_loop.root == root {
+                let owner = pl_attribs.owner.map(|owner| match owner {
+                    RawWindowHandle::Xlib(x) => x.window,
+                    RawWindowHandle::Xcb(x) => x.window as u64,
+                    raw => unreachable!("Invalid raw window handle {raw:?} on X11"),
+                });
+
+                if let Some(owner) = owner {
+                    unsafe {
+                        (xconn.xlib.XSetTransientForHint)(xconn.display, window.xwindow, owner);
+                    }
+                }
+            }
+
             // Set visibility (map window)
             if window_attrs.visible {
                 leap!(xconn.xcb_connection().map_window(window.xwindow)).ignore_error();
--- a/src/platform_impl/linux/x11/window.rs
+++ b/src/platform_impl/linux/x11/window.rs
@@ -242,6 +242,17 @@ impl UnownedWindow {
             ),
         };
 
+        // Set the transient parent if we don't have a parent already.
+        let owner = if event_loop.root == root {
+            pl_attribs.owner.map(|owner| match owner {
+                RawWindowHandle::Xlib(x) => x.window,
+                RawWindowHandle::Xcb(x) => x.window as u64,
+                raw => unreachable!("Invalid raw window handle {raw:?} on X11"),
+            })
+        } else {
+            None
+        };
+
         let window_attributes = {
             use xproto::EventMask;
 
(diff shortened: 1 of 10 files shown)


src/platform_impl/linux/x11/window.rs
```diff
@@ -424,6 +424,21 @@ impl UnownedWindow {
...
+            if event_loop.root == root {
+                let owner = pl_attribs.owner.map(|owner| match owner {
+                    RawWindowHandle::Xlib(x) => x.window,
+                    RawWindowHandle::Xcb(x) => x.window as u64,
+                    raw => unreachable!("Invalid raw window handle {raw:?} on X11"),
+                });
+
+                if let Some(owner) = owner {
+                    unsafe {
+                        (xconn.xlib.XSetTransientForHint)(xconn.display, window.xwindow, owner);
```
ids1024, line 437: As I understand, for right click menus, drop downs, etc. `override_redirect` Windows are typically used on X11. This is what `GdkPopover` uses. GDK separately has a method that sets `WM_TRANSIENT_FOR`: https://docs.gtk.org/gdk4/method.Toplevel.set_transient_for.html. On Wayland that corresponds to https://wayland.app/protocols/xdg-shell#xdg_toplevel:request:set_parent.

> Setting the override_redirect attribute True for a window allows a window to be mapped, moved, resized, or its border width changed without the intervention of the window manager. This override is usually done for menus that are frequently mapped and almost immediately unmapped again.
>
> Under properly designed window managers, there is a property you can set to tell the window manager to allow a window to pop up with minimal intervention (XA_WM_TRANSIENT_FOR). This is used for dialog boxes, as described in [12.3.1.4.6 Transient Window Field](https://www.oreilly.com/library/view/xlib-programming-manual/9780596806187/ch12s03.html#transient_window_field).

https://www.oreilly.com/library/view/xlib-programming-manual/9780596806187/ch04s03.html

So in Wayland and X, a "popup" and a "transient toplevel" are different things.

notgull: Thanks for telling me! I'll rework this PR once #2614 lands and I can more easily mold the X11 backend into something that is more conducive to this

ids1024:
So using GDK4's terminology (which is based on X11 and Wayland terminology), "popup surfaces" and "transient toplevels" are both useful features, but separate.

I'm not so sure what naming should be used in the winit APIs. It's not really possible to use terms like "menu" and "popup" without people interpreting them in different ways... Maybe "parent" window, as Wayland calls it, would make sense here. That may make more sense than "transient_for" or "owner". But then, "parent" means something *different* with X11 and Windows, where child windows are contained within the bounds of their parent...

Anyway, a transient toplevel may have window decoration, appear in the taskbar/etc, and otherwise behave mostly like a window without a parent. A popup is different, and on Wayland has to be positioned relative to its parent using [`xdg_positioner`](https://wayland.app/protocols/xdg-shell#xdg_positioner). So while on other platforms a "popup" might be similar to a toplevel but with a couple extra properties set, and the window placed in a particular position, on Wayland popups need a more specialized API. Which probably should be separate from the `Window` type.

notgull:
This is no longer blocked on `x11rb` support.

kchibisov:
While I do want popups inside the winit, I don't think this is the way as of now. I'm thinking of having a separate type and we'd need to put more research into it, which I can't really now.

The thing is that popup is not really a regular window in the end on at least Wayland, and we really want have a way to deal with not really a window, but you can still draw into it type of things.

notgull:
> While I do want popups inside the winit, I don't think this is the way as of now. I'm thinking of having a separate type and we'd need to put more research into it, which I can't really now.

I see, sounds good for now. I'll close until we can decide on a better API.

PR #2696, opened 2023-02-24
On macOS, fix initial focused state

### LINKED ISSUE #2695: Macos fires `WindowEvent::Focused(false)` on `WinitWindow` creation
In this commit: https://github.com/rust-windowing/winit/commit/b457329003e8e8f163565039c48690bb4c6a0d59#diff-4803bb2335a17d700335730b9f62828e001faa602b9b894dff572422c737b78aR480 the author fires the `Focused(false)` event on macos window creation. It seems to be inconsistent with the other platforms. 

In my case on macos I receive first `Focused(true)` and then `Focused(false)` when creating a window. The `Focused(true)` is caused by `window_did_become_key` and then `Focused(false)` is caused by the window init function queuing it when it is created. 

On other platforms it seems like the `focused` boolean is automatically set to false upon creation without firing an event. This seems to create divergent behavior on different platforms as macos has an additional event now. 

For now my fix is the following:

```rust
 WindowEvent::Focused(_focused) => {
        self.focused = winit.has_focus();
}
```

Sending an event on window creation creates unexpected behavior. Evenwhile my window is focused winit just informed me it isn't.

kchibisov:
The synthetic focused event was queued after the real event was send leading to focused issues on startup.

Fixes #2695.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On Wayland, fix rare crash on DPI change
 - Web: Added support for `Window::theme`.
 - On Wayland, fix rounding issues when doing resize.
+- On macOS, fix wrong focused state on startup.
 
 # 0.28.1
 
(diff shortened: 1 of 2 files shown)


kchibisov:
r @mockersf

mockersf (approved):

PR #2703, opened 2023-02-27
macOS: send the unfocus event before making the window key

mockersf:

On macOS, when creating a window after the application creation, it starts as unfocused, receiving a `WindowEvent::Focus(false)` immediately after a `WindowEvent::Focus(true)`.

This is because the event queued to set the default focus (false) was sent after the call to `makeKeyAndOrderFront`. This works fine if both windows are created at the same time at application creation, but not if the second window is created later as a result of something happening in the first window.

This PR moves the default event just before making the window key which fixes the issue for me.


kchibisov:
Thanks, for indirectly testing https://github.com/rust-windowing/winit/pull/2696!

mockersf:
aw sorry for missing your PR

PR #2718, opened 2023-03-06
implement set_cursor_hittest for x11

lunixbochs:

### DIFF
--- a/src/platform_impl/linux/x11/window.rs
+++ b/src/platform_impl/linux/x11/window.rs
@@ -12,8 +12,11 @@ use x11rb::{
     connection::Connection,
     properties::{WmHints, WmHintsState, WmSizeHints, WmSizeHintsSpecification},
     protocol::{
-        randr, xinput,
-        xproto::{self, ConnectionExt as _},
+        randr,
+        shape::SK,
+        xfixes::{ConnectionExt, RegionWrapper},
+        xinput,
+        xproto::{self, ConnectionExt as _, Rectangle},
     },
 };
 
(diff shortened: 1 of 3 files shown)


src/platform_impl/linux/x11/window.rs
```diff
@@ -1449,8 +1449,37 @@ impl UnownedWindow {
...
+                width: size.width as u16,
+                height: size.height as u16,
+            })
+        }
+        unsafe {
+            let region = (self.xconn.xfixes.XFixesCreateRegion)(
+                self.xconn.display,
+                rectangles.as_mut_ptr(),
+                rectangles.len() as i32,
+            );
```
notgull, line 1469: Can this region be cached?

lunixbochs: Where? It has the window size in it. Also at least in my use case, I'm usually just disabling hittest once and leaving it disabled for any given window.

lunixbochs:
this has an issue - if the window is resized, the hittest region becomes invalid, so we need to reapply it

kchibisov:
ping @notgull

notgull (changes requested):
Looks good to me, but this needs to be rebased on `master`. Probably using the [`x11rb` equivalent](https://docs.rs/x11rb/latest/x11rb/protocol/xfixes/fn.set_window_shape_region.html).

lunixbochs:
I rebased and switched to xcb. Could use testing and my app doesn't target winit master atm

notgull (approved):
Thanks!

PR #2720, opened 2023-03-06
On web, allow recreating event loop when using `spawn`

grovesNL:

As described in https://github.com/rust-windowing/winit/pull/2344#issuecomment-1176763837, it's useful to be able to recreate event loops on the web when using `spawn`, since the wasm module remains loaded. e.g., switching between tabs on a single page app and recreating a canvas doesn't currently work without some kind of workaround like this.

This won't change the behavior of `run`, so that will still panic when attempting to create multiple event loops.

I know there are probably still closures leaked when using this approach but don't expect this to be an issue in practice. We can also deal with any leaked closures later. At least anecdotally I've been using this approach for a while now and haven't run into any problems.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- On web, allow event loops to be recreated with `spawn`.
 - Bump MSRV from `1.60` to `1.64`.
 - Fix macOS memory leaks.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
-- On web, allow event loops to be recreated with `spawn`.
+- On Web, allow `EventLoop` to be recreated after using `spawn`.
 - Bump MSRV from `1.60` to `1.64`.
 - Fix macOS memory leaks.
 
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -9,11 +9,10 @@
 //! handle events.
 use std::marker::PhantomData;
 use std::ops::Deref;
-use std::sync::Mutex;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::{error, fmt};
 
 use instant::{Duration, Instant};
-use once_cell::sync::OnceCell;
 use raw_window_handle::{HasRawDisplayHandle, RawDisplayHandle};
 
 use crate::{event::Event, monitor::MonitorHandle, platform_impl};
@@ -67,7 +66,7 @@ impl EventLoopBuilder<()> {
     }
 }
 
-static EVENT_LOOP_CREATED: OnceCell<Mutex<bool>> = OnceCell::new();
+static EVENT_LOOP_CREATED: AtomicBool = AtomicBool::new(false);
 
 impl<T> EventLoopBuilder<T> {
     /// Start building a new event loop, with the given type as the user event
@@ -114,14 +113,8 @@ impl<T> EventLoopBuilder<T> {
     )]
     #[inline]
     pub fn build(&mut self) -> EventLoop<T> {
-        let mut event_loop_created = EVENT_LOOP_CREATED
-            .get_or_init(|| Mutex::new(false))
-            .lock()
-            .unwrap();
-        if *event_loop_created {
+        if EVENT_LOOP_CREATED.swap(true, Ordering::Relaxed) {
             panic!("Creating EventLoop multiple times is not supported.");
-        } else {
-            *event_loop_created = true;
         }
 
         // Certain platforms accept a mutable reference in their API.
@@ -132,11 +125,9 @@ impl<T> EventLoopBuilder<T> {
         }
     }
 
-    #[cfg(target_arch = "wasm32")]
+    #[cfg(wasm_platform)]
     pub(crate) fn allow_event_loop_recreation() {
-        if let Some(event_loop_created) = EVENT_LOOP_CREATED.get() {
-            *event_loop_created.lock().unwrap() = false
-        }
+        EVENT_LOOP_CREATED.store(false, Ordering::Relaxed);
     }
 }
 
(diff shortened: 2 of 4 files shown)


grovesNL:
cc @kchibisov this is the approach you mentioned on https://github.com/rust-windowing/winit/pull/2344#issuecomment-1176763837

kchibisov (changes requested):
I think that's not exactly what I've suggested.

What I meant was more like `.with_allow_multiple_loops(true)` in the new `EventLoopBuilderExtWeb`.

Can't you create multiple event loops at once on the web and then use `spawn`? If you can't and you must use `spawn` before creating the second loop, then it should be static in a comment.

You should also update the docs for the `spawn` that it allows multiple event loop creation.

Though, I guess you can't have 2 loops before spawn, since you need to consume the loop?

grovesNL:
I'd be happy to change the approach here but I'm not sure I follow the approach you described.

In case the use case isn't clear, my use case works like this:
- the wasm module is loaded
- spawn/the event loop is created once and sets the static flag
- some action in the application causes the event loop and canvas are destroyed – basically all state is destroyed except for statics because they're shared at the wasm module level
- some action in the application triggers a new spawn/event loop creation, which fails because of the static flag

The flow works the same way as run/spawn normally would, except I need to clear the static flag that's set at the wasm module level because it prevents run/spawn from being called again. I'm not trying to run multiple event loops simultaneously or similar.

kchibisov:
No, what I mean is could you do something like that on wasm

```rust
let evl1 = EventLoop::new();
let evl2 = EventLoop::new();
// And spawn later on.
```

Right now you won't be able to do so, because you clear only on spawn.

What I suggested is that once you build you pass inside the builder a flag that you must not restrict additional creation, so instead of clearing on `spawn` you won't set the static global at all.

grovesNL:
Sure, we could do that instead. I don't have a strong preference though it feels like we should just handle it internally instead of having it be opt-in on the builder. Is it valid to recreate the event loop multiple times without consuming it in run/spawn?

kchibisov:
>Is it valid to recreate the event loop multiple times without consuming it in run/spawn?

Not on any other platform, buy maybe you can do that on Wasm? That's what I'm asking you, if you can create multiple events loop at the same time on wasm, we should allow that with the extension to the builder, on any other platform it's impossible.

daxpedda:
> That's what I'm asking you, if you can create multiple events loop at the same time on wasm, we should allow that with the extension to the builder

I don't think this is possible in Winit, there are event handlers registered on the window itself, like the scale change observer, which would be overwritten by a new event loop.

It's possible to augment these event handlers to handle multiple event loops, but I don't think that's the goal here.

src/event_loop.rs
```diff
@@ -111,17 +114,30 @@ impl<T> EventLoopBuilder<T> {
...
+
         // Certain platforms accept a mutable reference in their API.
         #[allow(clippy::unnecessary_mut_passed)]
         EventLoop {
             event_loop: platform_impl::EventLoop::new(&mut self.platform_specific),
             _marker: PhantomData,
         }
     }
+
+    #[cfg(target_arch = "wasm32")]
```
kchibisov, line 135:
```suggestion
    #[cfg(wasm_platform)]
```

src/event_loop.rs
```diff
@@ -66,6 +67,8 @@ impl EventLoopBuilder<()> {
     }
 }
 
+static EVENT_LOOP_CREATED: OnceCell<Mutex<bool>> = OnceCell::new();
```
kchibisov, line 70: Shouldn't we refactor that to `AtomicBool` or simply `Mutex` given that it could be changed now? The `Mutex` is const anyway now.

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- On web, allow event loops to be recreated with `spawn`.
```
kchibisov, line 11:
```suggestion
- On Web, allow `EventLoop` to be recreated after using `spawn`.
```

src/platform_impl/web/event_loop/mod.rs
```diff
@@ -44,7 +46,15 @@ impl<T> EventLoop<T> {
         unreachable!();
     }
 
-    pub fn spawn<F>(self, mut event_handler: F)
+    pub fn spawn<F>(self, event_handler: F)
```
kchibisov, line 49: Could you update the `spawn` docs on the `trait` wrt new behavior change?

kchibisov (changes requested):

kchibisov:
>I don't think this is possible in Winit, there are event handlers registered on the window itself, like the scale change observer, which would be overwritten by a new event loop.

Thanks, that's what I wanted to know.

notgull:
Wouldn't it be possible to avoid resetting the event handlers by using `addEventListener` instead of setting them?

grovesNL:
Thank you for the review! I think everything should be addressed now.

I don't think it's worth trying to support multiple event loops right now (at least to solve this immediate issue) since it opens up a lot of other questions about how we might want to extend support for the web in general. For example:
- Should we allow multiple windows (canvases) each having their own event loop or should they share one event loop?
- Are we confident that all of the event handlers/CSS queries are set up correctly to support multiple event loops instead of setting a single event loop as the listener?
- How do multiple event loops interact with web workers and offscreen canvases?

I'd be happy to consider ways we could extend this later on, but the current change feels like a simple enough that we could land initially.

daxpedda (approved):
LGTM.

I think we shouldn't try to support multiple event loops until we know what a use-case for this even looks like, in any case I agree with @grovesNL this is out-of-scope for this PR.

daxpedda:
@grovesNL I'm happy to merge this after a rebase.

daxpedda:
I'm gonna close this as it's rather old now and needs a rebase.
Feel free to reopen after a rebase.

In general this issue is tracked in #2885.

arximboldi:
I really wish this was merged. Any chance @grovesNL to reactivate this?

PR #2721, opened 2023-03-07
Web: Updates the canvas to disable the contextmenu event and double click text selection

### LINKED ISSUE #2608: Bug: Right Mouse Button Release event never gets sent

Dig-Doug:

Fixes #2608.

Text selection: Try double clicking the canvas  on this [bevy example](https://bevyengine.org/examples/ui/text/). You'll see the example code below get selected

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ rls/
 *.js
 #*#
 .DS_Store
+.idea/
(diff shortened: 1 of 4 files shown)


daxpedda:
Would you mind confirming this on master? It should be fixed by #2662.
I couldn't reproduce this anymore.

daxpedda:
Feel free to reopen or comment if this is still an issue.

Issue #2722, opened 2023-03-08
Regression in 0.28 - NSWindow leaks on macOS


lunixbochs:
I bisected to 340f951d10c74169c258a120530b1eadf3a3e9fe, this appears to be caused by the objc2 port @madsmtm 

This caused a worse issue for me downstream than just a memory leak - my windows have strong references to metal layers on them, so after opening/closing enough windows, the system runs out of `IOSurface` or something and metal breaks system-wide.

I reproduced the issue by modifying `examples/window.rs` to the following and running `cargo run --example window` to basically create and drop a window every iteration. With the issue, memory usage grows unbounded. In 0.27.5, memory usage fluctuated but was stable. Xcode debugger suggests the `NSWindow` objects are being leaked.

```rust
#![allow(clippy::single_match)]

use simple_logger::SimpleLogger;
use winit::{
    event::{Event, WindowEvent},
    event_loop::EventLoop,
    window::WindowBuilder,
};

fn main() {
    SimpleLogger::new().init().unwrap();
    let event_loop = EventLoop::new();

    let mut window: Option<winit::window::Window> = None;

    event_loop.run(move |event, window_target, control_flow| {
        control_flow.set_wait();
        println!("{:?}", event);

        match event {
            Event::MainEventsCleared => {
                window.replace(WindowBuilder::new()
                    .with_title("A fantastic window!")
                    .with_inner_size(winit::dpi::LogicalSize::new(128.0, 128.0))
                    .build(&window_target)
                    .unwrap());
            }
            _ => (),
        }
    });
}
```

lunixbochs:
The WinitWindow has a retainCount of one, and the WinitView is holding a reference to it. If I release the WinitView, the window releases on its own.

madsmtm:
Oh yeah, we're now holding a strong reference to the window in `WinitView` (`WinitView::_ns_window`), could probably fix that by changing it to a weak reference.

lunixbochs:
I removed that reference and it successfully released the window but I think it still leaked the view

madsmtm:
Hmm, then it might be stuck in an autoreleasepool - could you try inserting `objc2::rc::autoreleasepool`?

lunixbochs:
With this example code, the WinitView is still not releasing:
```rust
fn main() {
    SimpleLogger::new().init().unwrap();
    let event_loop = EventLoop::new();

    let mut window: Option<winit::window::Window> = None;
    autoreleasepool(|_| {
[...]

PR #2727, opened 2023-03-09
Fix macos memory leaks

madsmtm:
Fixes https://github.com/rust-windowing/winit/issues/2722 without having to do the `objc2` update.

### DIFF
--- a/src/platform_impl/macos/util/mod.rs
+++ b/src/platform_impl/macos/util/mod.rs
@@ -6,6 +6,7 @@ pub(crate) use self::r#async::*;
 
 use core_graphics::display::CGDisplay;
 use objc2::foundation::{CGFloat, NSNotFound, NSPoint, NSRange, NSRect, NSUInteger};
+use objc2::{msg_send, ClassType};
 
 use crate::dpi::LogicalPosition;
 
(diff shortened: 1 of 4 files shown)


lunixbochs:
This crashes for me with the examples/window.rs example, if you drop the window:

Crashes every few runs (dropping one window):
```rust

use winit::{
    event::{Event, WindowEvent},
    event_loop::EventLoop,
    window::WindowBuilder,
};

fn main() {

    let mut window: Option<winit::window::Window> = None;
    window.replace(WindowBuilder::new()
        .build(&event_loop)
        .unwrap());

    event_loop.run(move |event, window_target, control_flow| {
        let event_loop = window_target;
        control_flow.set_wait();
        println!("{event:?}");

        match event {
            Event::MainEventsCleared => {
                window.take();
            }
            _ => (),
        }
    });
}
```

Crashes every run (repeatedly opening and closing a window):
```rust

use winit::{
    event::{Event, WindowEvent},
    event_loop::EventLoop,
    window::WindowBuilder,
};

fn main() {

    let mut window: Option<winit::window::Window> = None;

    event_loop.run(move |event, window_target, control_flow| {
        let event_loop = window_target;
        control_flow.set_wait();
        println!("{event:?}");

        match event {
            Event::MainEventsCleared => {
                window.replace(WindowBuilder::new()
                    .build(&event_loop)
                    .unwrap());
            }
            _ => (),
        }
    });
}
```

madsmtm:
I found out that this approach cannot be used (the `Box` that `DeallocHelper` internally uses is deallocated at the wrong time), so we'll have to do something else instead.

madsmtm:
Replaced by https://github.com/rust-windowing/winit/pull/2739

PR #2731, opened 2023-03-12
Web: Only Use PointerEvent for Touch

JAD3N:
- [x]  Created or updated an example program if it would help users understand this functionality

This builds on top of #2721 by making the mouse handler be the way to handle mouse events instead of pointer. Pointer events are still conditionally used (depending on browser support) for touch events.

By doing this it should resolve issue #2475 as pointer events don't trigger events for mouse buttons pressed while another mouse button is being held. More information can be found [here](https://developer.mozilla.org/en-US/docs/Web/API/Element/pointerdown_event) but the main concern is in the first paragraph:
> For mouse, it is fired when the device transitions from no buttons pressed to at least one button pressed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ And please only add new entries to the top of this list, right below the `# Unre
 # Unreleased
 
 - Bump MSRV from `1.60` to `1.64`.
+- Web: Canvases will now capture the contextmenu if prevent_default is enabled so they don't interfere with right click pointer events.
+- Web: Canvases will now have `user-select: none` to prevent double clicks from selecting text outside the canvas.
+- Web: Only use pointer events for touch typed events.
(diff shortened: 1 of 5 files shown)


JAD3N:
Currently I've left all mouse event handlers in `mouse_handler.rs` but it might be merge that into `canvas.rs` and add a new handler for touch events as `touch_handler.rs` and use that as a fallback if pointer events aren't available.

daxpedda:
This is definitely still a problem even after #2662, I will review this soon.

You don't need to rebase it until I review the approach to this.

daxpedda:
I replaced this by https://github.com/rust-windowing/winit/pull/2838.

> When a pointer button is already pressed, new button presses/releases won't go through `pointerdown`/`pointerup`, but through `pointermove`, [..]

So we shouldn't need to handle both events.

PR #2732, opened 2023-03-12
On Web, wake event loop on `request_redraw()`

### LINKED ISSUE #2019: Request_redraw() call from a task spawned by async_std::task::spawn_local doesn’t produce RedrawRequested event.
Request_redraw() call from a task spawned by async_std::task::spawn_local doesn’t produce RedrawRequested event.
Need to wait for another event like a mouse event to receive the RedrawRequested event.

daxpedda:

Currently, when calling `Window::request_redraw()` from outside the event loop, for example from an async background task, it will not wake up the event loop.
This fixes it by simply calling the same internal function used to wake up the event loop by sending a `UserEvent`, I hope that is correct.

Fixes #2019.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 - Bump MSRV from `1.60` to `1.64`.
 - Fix macOS memory leaks.
+- On Web: fix `Window::request_redraw` not waking the event loop when called from outside the loop.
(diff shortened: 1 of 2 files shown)


daxpedda:
CI fails because of the new Clippy version.
See https://github.com/rust-windowing/winit/pull/2729.

daxpedda:
Restarted after #2729. CI is green now.

PR #2735, opened 2023-03-13
On Web, use `target_family = "wasm"`

daxpedda:

This changes all uses of `target_arch = "wasm32"` to `target_family = "wasm"`, which also covers the `wasm64-unknown-unknown` target.

This was introduced in [Rust v1.54](https://github.com/rust-lang/rust/blob/1.68.0/RELEASES.md#compiler-14).

Not sure if I should add a changelog entry?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -103,7 +103,7 @@ features = [
     "Win32_UI_WindowsAndMessaging",
 ]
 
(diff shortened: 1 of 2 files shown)


kchibisov:
Should we test wasm64 on CI now as well? Right now we're running only wasm32.

daxpedda:
I'm happy to add it to the CI, but it's nightly only and requires `-Zbuild-std`, I don't think we want to signal official support. On the other hand it might be nice to know when it breaks.

Let me know and I will add it to the PR.

kchibisov (approved):
Then it's likely fine.

PR #2723, opened 2023-03-08
Bump MSRV to 1.64

kchibisov:

PR #2737, opened 2023-03-13
On Web, `EventLoopProxy` now implements `Send`

daxpedda:

See #2294 for previous discussion on this.

This adjusts `EventLoopProxy` on Web to be `Send`, like other platforms.
Because the `EventLoop` itself is not `Send`, the way this is solved is by using a channel that receives messages in the background and redirects them to the `EventLoop`, with the help of `wasm_bindgen_futures::spawn_local()`, adding a new dependency, even if fairly minor.

Avoiding the `wasm-bindgen-futures` is kind of impossible because we need async support, but I did build our own async channel with the help of a simple `Mutex`.

I have some additional ideas in mind:
- We could avoid the additional dependency by saying that we only support `Send` when you are using `target_feature = "atomics"` and otherwise fall back to what we were doing until now. This would add a bunch of `cfg`s inside the code.
- If we don't want to import a big crate like [`async-channel`](https://crates.io/crates/async-channel/1.8.0/dependencies) we could at least import [`atomic-waker`](https://crates.io/crates/atomic-waker/1.1.0/dependencies), which has no dependencies and is quite tiny, removing the `Mutex`.
- I adjusted the web example to send custom user events to try this out, but it's written in a way to also be able to compile on other platforms, which adds a lot of `cfg`s everywhere, not sure if I should add it here: 
  <details><summary>Adjusted Example</summary>

  ```diff
  --- a/examples/web.rs
  +++ b/examples/web.rs
  @@ -1,13 +1,53 @@
  #![allow(clippy::single_match)]
  
  +use std::time::Duration;
  +
  +#[cfg(not(wasm_platform))]
  +use std::thread;
  +#[cfg(wasm_platform)]
  +use wasm_bindgen::{closure::Closure, JsCast};
  use winit::{
      event::{Event, WindowEvent},
  -    event_loop::EventLoop,
  +    event_loop::EventLoopBuilder,
      window::WindowBuilder,
  };
  
  +const INTERVAL: Duration = Duration::from_millis(2500);
  +
  +#[derive(Clone, Copy, Debug)]
  +pub struct Counter(u64);
  +
  pub fn main() {
  -    let event_loop = EventLoop::new();
  +    #[cfg(not(wasm_platform))]
  +    simple_logger::SimpleLogger::new().init().unwrap();
  +
  +    let event_loop = EventLoopBuilder::with_user_event().build();
  +    let proxy = event_loop.create_proxy();
  +    let mut counter = Counter(0);
  +
  +    let mut sender = move || {
  +        let result = proxy.send_event(counter);
  +        counter.0 += 1;
  +        result
  +    };
  +
  +    #[cfg(not(wasm_platform))]
  +    thread::spawn(move || loop {
  +        thread::sleep(INTERVAL);
  +        if let Err(_) = sender() {
  +            break;
  +        }
  +    });
  +    #[cfg(wasm_platform)]
  +    let closure = { Closure::<dyn FnMut()>::new(move || sender().unwrap()) };
  +    #[cfg(wasm_platform)]
  +    let handle = web_sys::window()
  +        .unwrap()
  +        .set_interval_with_callback_and_timeout_and_arguments_0(
  +            closure.as_ref().unchecked_ref(),
  +            INTERVAL.as_millis().try_into().unwrap(),
  +        )
  +        .unwrap();
  
      let window = WindowBuilder::new()
          .with_title("A fantastic window!")
  @@ -20,14 +60,25 @@ pub fn main() {
      event_loop.run(move |event, _, control_flow| {
          control_flow.set_wait();
  
  +        log::debug!("{:?}", event);
  +
          #[cfg(wasm_platform)]
  -        wasm::log_event(&log_list, &event);
  +        {
  +            let _closure = &closure;
  +            wasm::log_event(&log_list, &event);
  +        }
  
          match event {
              Event::WindowEvent {
                  event: WindowEvent::CloseRequested,
                  window_id,
  -            } if window_id == window.id() => control_flow.set_exit(),
  +            } if window_id == window.id() => {
  +                #[cfg(wasm_platform)]
  +                web_sys::window()
  +                    .unwrap()
  +                    .clear_interval_with_handle(handle);
  +                control_flow.set_exit();
  +            }
              Event::MainEventsCleared => {
                  window.request_redraw();
              }
  @@ -38,6 +89,7 @@ pub fn main() {
  
  #[cfg(wasm_platform)]
  mod wasm {
  +    use super::Counter;
      use wasm_bindgen::prelude::*;
      use winit::{event::Event, window::Window};
  
  @@ -71,20 +123,22 @@ mod wasm {
          log_list
      }
  
  -    pub fn log_event(log_list: &web_sys::Element, event: &Event<()>) {
  -        log::debug!("{:?}", event);
  -
  +    pub fn log_event(log_list: &web_sys::Element, event: &Event<Counter>) {
          // Getting access to browser logs requires a lot of setup on mobile devices.
          // So we implement this basic logging system into the page to give developers an easy alternative.
          // As a bonus its also kind of handy on desktop.
  -        if let Event::WindowEvent { event, .. } = &event {
  -            let window = web_sys::window().unwrap();
  -            let document = window.document().unwrap();
  -            let log = document.create_element("li").unwrap();
  -            log.set_text_content(Some(&format!("{event:?}")));
  -            log_list
  -                .insert_before(&log, log_list.first_child().as_ref())
  -                .unwrap();
  -        }
  +        let event = match &event {
  +            Event::WindowEvent { event, .. } => format!("{event:?}"),
  +            Event::UserEvent(event) => format!("{event:?}"),
  +            _ => return,
  +        };
  +
  +        let window = web_sys::window().unwrap();
  +        let document = window.document().unwrap();
  +        let log = document.create_element("li").unwrap();
  +        log.set_text_content(Some(&event));
  +        log_list
  +            .insert_before(&log, log_list.first_child().as_ref())
  +            .unwrap();
      }
  }
  ```
  </details>

Partially replaces #2294.


src/platform_impl/web/event_loop/proxy.rs
```diff
@@ -1,26 +1,99 @@
...
+            Err(TryRecvError::Empty) => {
+                *self.waker.lock().unwrap() = Some(cx.waker().clone());
+
+                match self.receiver.try_recv() {
+                    Ok(event) => {
+                        self.waker.lock().unwrap().take();
+                        Poll::Ready(Ok(event))
+                    }
+                    Err(TryRecvError::Empty) => Poll::Pending,
+                    Err(TryRecvError::Disconnected) => Poll::Ready(Err(RecvError)),
```
daxpedda, line 93: I take out the `Waker` again if we aren't waiting anymore, which is great for optimization, but on the other hand we are using a `Mutex` which kinda makes this worse?

I don't take out the `Waker` if we fail, because our listener aborts in that case anyway.

daxpedda: I made a small benchmark with this, turns out it's pretty awful, so I just removed it.

src/platform_impl/web/event_loop/proxy.rs
```diff
@@ -1,26 +1,96 @@
...
+}
+
+impl<T: 'static> Future for AsyncReceiver<T> {
+    type Output = Result<T, RecvError>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        match self.receiver.try_recv() {
+            Ok(event) => Poll::Ready(Ok(event)),
+            Err(TryRecvError::Empty) => {
+                *self.waker.lock().unwrap() = Some(cx.waker().clone());
```
notgull, line 85: You should use `will_wake` here in order to avoid unnecessarily cloning a waker twice.

src/platform_impl/web/event_loop/runner.rs
```diff
@@ -108,7 +113,22 @@ impl<T: 'static> Shared<T> {
...
             scale_change_detector: RefCell::new(None),
             unload_event_handle: RefCell::new(None),
-        }))
+            proxy_sender,
+        }));
+
+        wasm_bindgen_futures::spawn_local({
+            let runner = this.clone();
+            async move {
+                while let Ok(value) = (&mut proxy_receiver).await {
```
notgull, line 122: I think it would be more idiomatic to have an `async fn next()` on the receiver and to poll that instead. Futures aren't really meant to be polled after they're completed.

src/platform_impl/web/event_loop/proxy.rs
```diff
@@ -1,26 +1,96 @@
...
+        waker: Arc::clone(&waker),
+    };
+    let receiver = AsyncReceiver { receiver, waker };
+
+    (sender, receiver)
+}
+
+pub struct AsyncSender<T: 'static> {
+    sender: Sender<T>,
+    waker: Arc<Mutex<Option<Waker>>>,
```
notgull, line 49: Is `Mutex` available on WASM? I always forget what synchronization primitives panic on block for non-atomic WASM or not. Although, it looks like that, in single threaded cases, it shouldn't block. Is this right?

daxpedda: > Is `Mutex` available on WASM?

Yes.

> Although, it looks like that, in single threaded cases, it shouldn't block. Is this right?

For single threaded cases it shouldn't block indeed, but it can for multi-threaded cases. Considering we are just storing a single value and not holding a lock in practice it's not a problem.

I would still like to propose that we don't use a `Mutex` here and use [`AtomicWaker`](https://docs.rs/atomic-waker/1.1.0/atomic_waker/struct.AtomicWaker.html) instead. The crate has [no dependencies](https://crates.io/crates/atomic-waker/1.1.0/dependencies) and is [really small](https://docs.rs/atomic-waker/1.1.0/src/atomic_waker/lib.rs.html).

If the dependency is still a concern, I would like to point to my suggestion above: hiding the `Send` support behind the `+atomics` target feature.

src/platform_impl/web/event_loop/proxy.rs
daxpedda: I didn't know about this! I also noticed that neither `futures` or `atomic-waker` use this this optimization in their `AtomicWaker` implementation. I open an issue here: https://github.com/smol-rs/atomic-waker/issues/11.

In our case this won't be possible because to avoid a racing condition we would have to hold the lock, which can lead to blocking in a multi-threaded environment.

notgull: Couldn't you do:

```rust
{
    let mut waker = self.waker.lock().unwrap();
    if waker.map_or(true, |waker| !waker.will_wake(cx.waker())) {
        *waker = Some(cx.waker().clone());
    }
}
```

in order to drop the lock once the temporary scope ends?

src/platform_impl/web/event_loop/runner.rs
daxpedda: I needed some way to yield, so instead of writing a custom [`yield_now`](https://docs.rs/tokio/latest/tokio/task/fn.yield_now.html) I just used [`poll_fn`](https://doc.rust-lang.org/stable/core/future/fn.poll_fn.html), which was stabilized in v1.64, which we just bumped to just recently anyway https://github.com/rust-windowing/winit/pull/2723.

src/platform_impl/web/event_loop/proxy.rs
daxpedda: I wasn't concerned with how to drop the lock, I was concerned with the fact that we hold it in the first place. The assumption in play here is that if you only assign a value without any checks in-between, optimizations can prevent this to block the main thread. Not really a provable assumption, but it's what we have right now without using `AtomicWaker`.

If the the worker yields between holding the lock and comparing and cloning the `Waker`, we are gonna have a problem.

daxpedda:
CI fails on iOS because of `warning: spurious network error (2 tries remaining): [6] Couldn't resolve host name (Could not resolve host: crates.io)`.

notgull:
> CI fails on iOS because of `warning: spurious network error (2 tries remaining): [6] Couldn't resolve host name (Could not resolve host: crates.io)`.

Yeah, the iOS CI check does that sometimes. It's not your fault

src/platform_impl/web/event_loop/proxy.rs
notgull: I would support using `atomic-waker`, but I'd ask the rest of the maintainers first.

daxpedda:
Closing in favor of #2740.

PR #2738, opened 2023-03-13
Only enable instant/wasm-bindgen when building for wasm32 architecture

gferon:
This is the mirror change of https://github.com/emilk/egui/pull/2808 to fix compilation with `bazel`.


daxpedda:
While we are at it, I would like to suggest removing it completely from native builds.
There are only two places that need a `cfg` guard anyway:
- https://github.com/rust-windowing/winit/blob/77f8e511e965354a4b827cfb1ca197e9e51e6ded/src/event.rs#L37
- https://github.com/rust-windowing/winit/blob/77f8e511e965354a4b827cfb1ca197e9e51e6ded/src/event_loop.rs#L14

Examples can be kept by just adding it to the `dev-dependencies` ... only one example anyway.

PR #2739, opened 2023-03-13
Fix macos memory leaks

madsmtm:
Fixes https://github.com/rust-windowing/winit/issues/2722 by patching `objc2` instead, and using version ranges to allow the patched version to be selected.

If the user has `accesskit_macos` in their dependency tree, this will not work until @mwcampbell uses `objc2 = ">=0.3.0-beta.3, <0.3.0-beta.4"` in that.

  - ~Currently very limited internet, and having a few problems with `cargo update`, so I can't test this fully myself right now. I have tested that the version resolution works though!~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 # Unreleased
 
 - Bump MSRV from `1.60` to `1.64`.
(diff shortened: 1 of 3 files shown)


madsmtm:
Thanks @lunixbochs - I've tested it myself now as well.

@kchibisov: Feel free to include this in the next patch version of `winit`.

PR #2740, opened 2023-03-16
On Web, implement `Send` and `Sync` for `Window` and `EventLoopProxy`

daxpedda:
This is an experiment to let `Window` implement `Send` and `Sync` on Wasm and reuse the same code to implement `Sync` for `EventLoopProxy`.
I took some inspiration from [`MainThreadSafe` on macOS](https://github.com/rust-windowing/winit/blob/77f8e511e965354a4b827cfb1ca197e9e51e6ded/src/platform_impl/macos/util/async.rs).

It's implemented in a way that if you call this from the `Window` it will do nothing special and just call the appropriate method on the `Window`, as before. Therefor, except a single `Rc<Cell<bool>>` to `Arc<AtomicBool>` change, there is no overhead if Wasm is only used single-threaded.

If used in a multi-threaded environment from a worker, it will use a `mpsc::Sender` to do it's work on the main thread. Currently this is implemented by just sending a boxed function to execute, but it could be rewritten to just send a message and a message handler would then do the right thing.

When it's necessary to return a value a `Condvar` and a `Mutex` is used to await the completion, which would block, but as it's not in the window that should be fine.

Most of the small changes here and there are to disallow calls to `web_sys::window()`, as these are the main culprits that will fail if not called from a window. This was necessary because it was quite hard to figure out what needs to be changed, so I added a Clippy `disallowed-methods`, which should hopefully make maintenance of this feature easy. Let me know what you think of this.

Builds on top of #2732 and #2737.

Fixes #2294.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,8 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 - Fix macOS memory leaks.
 - On Web: fix `Window::request_redraw` not waking the event loop when called from outside the loop.
-- On Web, `EventLoopProxy` now implements `Send`.
+- On Web, `EventLoopProxy` now implements `Send` and `Sync`.
+- On Web, `Window` now implements `Send` and `Sync`.
 
 # 0.28.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 # 0.28.3
 
 - Fix macOS memory leaks.
+- On Web: fix `Window::request_redraw` not waking the event loop when called from outside the loop.
 
 # 0.28.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 - Fix macOS memory leaks.
 - On Web: fix `Window::request_redraw` not waking the event loop when called from outside the loop.
+- On Web, `EventLoopProxy` now implements `Send`.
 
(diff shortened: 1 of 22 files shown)


daxpedda:
https://github.com/rust-windowing/winit/blob/fbea75d31f4025825e9bf88b99d7c92e4eb79273/tests/sync_object.rs#L5-L9

Or do you mean more then just this simple test?

kchibisov:
>Or do you mean more then just this simple test?

Ugh, I guess it's fine then. it's rather strange requirement given that you can't clone it...

I know that you must have `Send` and I guess I never seen `Sync` failure on its own.

daxpedda:
> it's rather strange requirement given that you can't clone it...

It can be cloned when used in an `Arc`, but the `Arc` itself requires `Send` and `Sync` to be `Send` itself ([docs](https://doc.rust-lang.org/stable/alloc/sync/struct.Arc.html#impl-Send-for-Arc%3CT%3E)). The only way to get around `Sync` otherwise is to wrap it in a `Mutex`, which unfortunately breaks Wasm when run on the main thread.

At least that's my use-case, @Liamolucko described some additional requirements in #2294.

daxpedda:
So I played around with this a bit more, added some comments, prevented leaking unnecessary `Weak`s and a potential block when cloning.

The only thing left is preventing that one `Weak` wrapped in `ManuallyDrop` from leaking, which I have a pretty good idea how to do, but gonna stop wasting my time on this until the other PR's are merged and this one receives a review.

daxpedda:
(fails because of duplicated `syn` dependencies)

daxpedda:
Closed in favor of #2834.

PR #2741, opened 2023-03-16
Winit version 0.28.3

kchibisov:


kchibisov:
The CI failure due to MSRV, the build was done locally on macOS with `1.60.0` without issues.

PR #2743, opened 2023-03-17
Jmaargh/events docs

jmaargh:
Fixes #27361

I ended up changing more text than I'd originally planned at the root level of the docs. Happy to make large changes/reversions if the maintainers would prefer - but to me this reads as a much better "what I would have wanted to know before starting" than what there was.

I also took the opportunity to fix up some docs I found along the way which I thought could use some love, specifically:
- `map_user_event()`
- `Event::DeviceEvent`


src/lib.rs
```diff
@@ -19,27 +19,60 @@
...
+//! marking the end of the primary loop.
+//!
+//! Some platforms currently run one iteration of this primary loop immediately after the first [`Resumed`] event, but
+//! this is platform-dependent and may change.
+//!
+//! The [`Suspended`] and [`Resumed`] loop-stage events are used on some platforms (typically mobile platforms) to
+//! indicate suspension and resumption of the application by the platform. See the [detailed docs][`Suspended`] for details.
+//!
+//! Once the [`ControlFlow`] has been set to exit (typically with [`ControlFlow::set_exit()`] or [`set_exit_with_code()`][`ControlFlow::set_exit()`])
+//! a final [`LoopDestroyed`] event is dispatched. This is irreversible and a typical place to free all application resources,
```
filnet, line 60: suggest to use `time` instead of `place`.

jmaargh:
Anybody keen to review this? I'd love to get something merged while I have some time on my hands

src/lib.rs
```diff
@@ -19,27 +19,60 @@
...
+//! as the process will be terminated after this.
+//!
+//! This overview should be correct for cross-platform use in most cases, please see the detailed descriptions in the
+//! API docs for the exact behaviour of each event and details of any platform-dependent behaviour. In particular on some
+//! platforms it is possible (though discouraged) to use [`run_return()`][`EventLoopExtRunReturn::run_return`] to prevent
+//! [`run()`][`EventLoop::run()`] from hijacking the process, however this may change the order of events significantly,
+//! so carefully check the API documentation.
+//!
+//! The [`winit_input_helper`](https://crates.io/crates/winit_input_helper) crate can help the mangement of main events
```
madsmtm, line 69: While I generally like suggesting crates to do a specific task, it seems like quite a prominent place to put it? If our event handling is so difficult to understand that we need to direct users to a different crate, perhaps we should adopt that crate's API instead?

Relatedly, I suspect the rendering logic presented in the readme in `winit_input_helper` is wrong.

docs/event-loop.svg
madsmtm, line 1: Nit: I dislike the duplicate "main" loop, perhaps that could be merged into one somehow (not sure how though)

madsmtm, line 1: The suspended/resumed events could be clearer that once you get `Suspended`, the next event is guaranteed to be `Resumed`.

Think state machine, each event has arrows exhaustively pointing toward the possible next events.

madsmtm (changes requested):
Hey @jmaargh, sorry that no-one has responded to your PR! I've tried to give a bit of feedback, but in general this is definitely an improvement, and I'd very much like to get it merged.

src/lib.rs
```diff
@@ -19,27 +19,60 @@
...
+//! current process. All application code is then managed from this closure.
 //!
-//! You can retrieve events by calling [`EventLoop::run`][event_loop_run]. This function will
-//! dispatch events for every [`Window`] that was created with that particular [`EventLoop`], and
-//! will run until the `control_flow` argument given to the closure is set to
-//! [`ControlFlow`]`::`[`ExitWithCode`] (which [`ControlFlow`]`::`[`Exit`] aliases to), at which
-//! point [`Event`]`::`[`LoopDestroyed`] is emitted and the entire program terminates.
+//! A running [`EventLoop`] works by passing [`Event`]s to the closure. The closure handles these events and can control
+//! the behaviour of the loop through setting the [`ControlFlow`] also passed to the closure. With the order illustrated
+//! by the following diagram, showing how the events loop.
```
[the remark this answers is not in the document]
jmaargh: A casualty of lazy editing, good catch.

src/lib.rs
jmaargh: Absolutely right, this wasn't appropriate.

docs/event-loop.svg
jmaargh: I wasn't a big fan initially, but decided it was best to clearly communicate the actual current behaviour, even if it's a little less "tidy" (since this is a result of the actual behaviour being less "tidy", at least when I originally wrote this, apologies if it's changed in the meantime).

The aim of the diagram was supposed to be illustrative (for pedagogical purposes for new users), not necessarily comprehensive. This is not least because **I** don't know every corner case to cover for a comprehensive one.

jmaargh: I've added some text to the prose to cover this explicitly.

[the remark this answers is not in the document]
jmaargh: Happy to change this if this is the consensus. However, I used the standard flowchart shapes for "start" and "end" and paired with commonly understood traffic-light colours for a bit of visual flair.

jmaargh:
@madsmtm @daxpedda Thanks both for the reviews, very helpful comments.

I think I've addressed everything (either with changes or comments). I haven't changed the diagram at all, but can do if still desired.

src/lib.rs
madsmtm: No worries, you were just trying to make the library better for everyone!

madsmtm:
Thanks for the changes!

I think @kchibisov is working on changing how things work in https://github.com/rust-windowing/winit/pull/2896, so I'd like to hear from them and @rib before moving forwards.

rib:
There certainly are some changes in the pipeline that are likely to affect this but, so long as we feel like this isn't immutable then maybe it's ok to land and we can update it when things change.

I just tried opening the diagram in https://app.diagrams.net/ and although I guess it's a proprietary web app it seems simple enough to use.

Main changes I'm hoping to see though would be:

RedrawEventsCleared can probably just be removed, since the concept doesn't really make sense imho if we no longer guarantee an explicit ordering for when we emit redraw events relative to other 'main' events.

With the direction discussed for the pump_events/run_ondemand changes then MainEventsCleared is becoming more like an `AboutToWait` - i.e. it's more like a RunLoop observer event for tracking wake up and wait transitions, and should probably get renamed.

LoopDestroyed should probably be renamed to LoopExiting as was discussed in relation with the run_ondemand changes.

kchibisov:
I'm sort of concerned that the diagram isn't really accurate, given that we have backends delivering `RedrawRequested` regardless, we really can't guarantee the ordering like we mention for some backends, not that we really want to.

I'm also not even sure that we'll keep `MainEventsCleared`, it seems like we do don't guarantee when the `RedrawRequested` arrives, we don't really need it, and `AboutToPoll` should be enhough.

daxpedda:
> I just tried opening the diagram in https://app.diagrams.net/ and although I guess it's a proprietary web app it seems simple enough to use.

I think, if it's not too much work for you @jmaargh, we should shift to [Mermaid](https://mermaid.js.org), which has [native support in GitHub](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-diagrams#creating-mermaid-diagrams), [Rustdoc integration](https://crates.io/crates/aquamarine) and [MdBook integration](https://crates.io/crates/mdbook-mermaid).

It's completely open source and has [an online editor](https://mermaid.live) that's fairly easy to use.

jmaargh:
> I'm sort of concerned that the diagram isn't really accurate, given that we have backends delivering `RedrawRequested` regardless, we really can't guarantee the ordering like we mention for some backends, not that we really want to.

I'm afraid I don't understand how this makes the diagram (or the text) inaccurate? If `RedrawRequested` is always delivered on a platform, then that is covered by the "platform ... requests it" text. All the diagram says is that `RedrawRequested`, when it occurs, will occur after `MainEventsCleared`. Could you explain a bit why you feel these docs need to change?

jmaargh:
> I think, if it's not too much work for you @jmaargh, we should shift to [Mermaid](https://mermaid.js.org), which has [native support in GitHub](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-diagrams#creating-mermaid-diagrams), [Rustdoc integration](https://crates.io/crates/aquamarine) and [MdBook integration](https://crates.io/crates/mdbook-mermaid).

No worries, I'll get that done soon.

jmaargh:
@rib As the behaviour changes, I'm very happy for this documentation to change with it. I'd be very happy to be involved with making those updates if that's helpful (unfortunately I can't guarantee an SLA though!)

Unless there's a definite major change to the API/behaviour that's imminent, my preference would be to get this merged as describing the current behaviour. The docs can always (and should) change as the API does.

PR #2744, opened 2023-03-17
feat: Add a safe method for cross-crate interop to winit

notgull:
- [ ] Tested on all platforms changed (quite a few platforms here)

One of the main complaints with the current Rust GUI ecosystem is that interoperability requires unsafe code. As of now, the `raw-window-handle` crate is used for interoperability; however, it has two disadvantages:

- There is no lifetime parameter attached to the handle, meaning that the window can be closed at any time.
- The display handle and all of its windows can become invalid at any time on Android due to the app being suspended.

`raw-window-handle` probably isn't at the appropriate level of abstraction to be fixing this anyways. However, this has led to some friction in crates that depend on it (gfx-rs/wgpu#1463 and rust-windowing/softbuffer#80 off the top of my head). I created a [`window-handle`](https://github.com/notgull/window-handle) crate in an attempt to solve this. The new features are:

- A lifetime parameter is attached to the handles to prevent them from outliving their sources.
- An `Active` struct is used to access the window handles. `Active` can only exist when the application isn't suspended.

This PR adds the current implementation of [`window-handle`](https://github.com/notgull/window-handle) to this crate and implements its traits. This PR also serves as a request for comments on the [`window-handle`](https://github.com/notgull/window-handle) crate. I'm not familiar with all of the platforms, so it's likely that there's some invariant that I've missed.


Lokathor:
I think that the new crate should also have the Zlib license as a license option the same as the raw-window-handle crate does, but otherwise it seems like an entirely appropriate "one level up" abstraction

notgull:
> I think that the new crate should also have the Zlib license as a license option the same as the raw-window-handle crate does,

Sounds good to me, done!

kchibisov:
I wonder whether raw-window-handle itself should provide such abstraction in addition, simply because it's also a no_std single rwhd dep crate? So we'll have the same crate for such handles which we could stabilize and use?

Managing 2 separate crates for handles could be not that easy, because you'd need to update both at the same time if raw-window-handle break something(like we want to add `Result`, when you can't build the handle).

notgull:
> I wonder whether raw-window-handle itself should provide such abstraction in addition, simply because it's also a no_std single rwhd dep crate? So we'll have the same crate for such handles which we could stabilize and use?
> 
> Managing 2 separate crates for handles could be not that easy, because you'd need to update both at the same time if raw-window-handle break something(like we want to add `Result`, when you can't build the handle).

I would support this, I can file a PR if you want @Lokathor

Lokathor:
I'd be entirely fine with adding this right into the existing crate.

notgull:
CI failures are due to having two versions of `raw-window-handle` in the dependency tree. Should be fixed once the patch is released.

Lokathor:
i can do a 0.5.1 later tonight

notgull:
> i can do a 0.5.1 later tonight

That'd be great. That being said, I'd like to make sure that we aren't too hasty to commit to any design decisions, and give some time to let people respond to rust-windowing/raw-window-handle#111

Lokathor:
Alright, good point.

Since it's just one CI pass out of many, and since this is just a draft, we'll wait for more feedback.

notgull:
Ported to rust-windowing/raw-window-handle#116

notgull:
`raw-window-handle` version `0.5.2` has been published, so we can use that now. I'd say this PR is ready for review now.

notgull:
Apple CI is failing because of https://github.com/madsmtm/objc2/issues/432, and the other CI is failing because of `syn` 2.0.

src/event_loop.rs
```diff
@@ -48,6 +50,17 @@ pub struct EventLoopWindowTarget<T: 'static> {
     pub(crate) _marker: PhantomData<*mut ()>, // Not Send nor Sync
 }
 
+/// Target that provides a handle to the display powering the event loop.
+///
+/// This type allows one to take advantage of the display's capabilities, such as
+/// querying the display's resolution or DPI, without having to create a window. It
+/// implements `HasDisplayHandle`.
+#[derive(Clone)]
+pub struct OwnedDisplayHandle {
```
kchibisov, line 59: I don't really understand the purpose of this type. The `EventLoop` itself is what should have the `HasDisplayHandle` impl, and its lifetime is what should be used. The fact that you're not using this type in examples at all suggests me that there's no clear case when it should be used(at least you should use it inside the examples).

The `EventLoopWindowTarget` must have the same lifetime as the `EventLoop` itself, because it's simply a field on `EventLoop` in all the backends.

The `Clone` is also sort of questionable, because it simply will discard the lifetime attached to it, given that there's only a way to get `&OwnedDisplayHandle` not the owned type on its own?

if that's intent on `lifetime` casting, different method should be used with `unsafe` bit on it(because it's simply `mem::tranmsute` like thingy of donig `'static` lifetime).

kchibisov (changes requested):

src/event_loop.rs
notgull: The reason why this type exists is for the `glutin` case. The scenario is that you need an object that has a display handle before a window is created, since you need to query the display. You can't use `EventLoop`, `&EventLoop` or a `DisplayHandle<'_>` taken from an `EventLoop` because the `EventLoop` is owned/borrowed mutably for event handling. You can't use the `EventLoopWindowTarget` that is provided in the event handler, since it disappears between events, which means it can't be used persistently. Normally I'd use a `Window`, but for `glutin` (outside of Win32) you haven't created a `Window` yet.

Therefore I created this type to fill that hole. Something that implements `HasDisplayHandle` that can be used according to Rust's borrowing system. This way, it can be used in the display position without any unsafe hacks.

If the `Clone` is a deal breaker I can get rid of it.

kchibisov: >actually, it might be necessary for the borrowing rules to check out in this case.

I just think that it's `unsafe`, so you might have different method like `to_static`, which casts away lifetime.

notgull: I have added documentation clarifying the purpose of `OwnedWindowHandle` and why it is safe.

src/platform_impl/linux/x11/mod.rs
```diff
@@ -99,6 +100,11 @@ impl<T> PeekableReceiver<T> {
     }
 }
 
+#[derive(Clone)]
+pub struct OwnedDisplayHandle {
+    xconn: Arc<XConnection>,
```
kchibisov, line 105: you can use `Rc` for it given that it's neither send nor sync.

src/platform_impl/linux/wayland/event_loop/mod.rs
```diff
@@ -601,6 +617,19 @@ impl<T: 'static> EventLoop<T> {
...
+#[derive(Clone)]
+pub struct OwnedDisplayHandle {
+    display: Display,
+}
+
+impl OwnedDisplayHandle {
+    pub fn raw_display_handle(&self) -> RawDisplayHandle {
+        let mut display_handle = WaylandDisplayHandle::empty();
+        display_handle.display = self.display.get_display_ptr() as *mut _;
+        RawDisplayHandle::Wayland(display_handle)
```
kchibisov, line 629: you can now do `into()` instead, though, I'd suggest to update all such places to call into given that you're updating the raw-window-handle crate.

examples/borrowed_window.rs
```diff
@@ -0,0 +1,47 @@
+//! Use borrowed window handles in `winit`.
```
kchibisov, line 1: Could you just print it from the common window example, creating more examples to print just one thing seems sort of redundant.

kchibisov (changes requested):

src/platform_impl/linux/x11/mod.rs
notgull: The object that's being passed around uses an `Arc` because it's being stored in a static variable.

https://github.com/rust-windowing/winit/blob/fbea75d31f4025825e9bf88b99d7c92e4eb79273/src/platform_impl/linux/mod.rs#L118-L120

If we want to transition away from this model, it'd probably need to take place outside of this PR.

src/event_loop.rs
```diff
@@ -48,6 +50,71 @@ pub struct EventLoopWindowTarget<T: 'static> {
     pub(crate) _marker: PhantomData<*mut ()>, // Not Send nor Sync
 }
 
+/// Target that provides a handle to the display powering the event loop.
+///
+/// This type allows one to take advantage of the display's capabilities, such as
+/// querying the display's resolution or DPI, without having to create a window. It
+/// implements [`HasDisplayHandle`].
+///
```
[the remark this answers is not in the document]
notgull: I removed the part about querying. "Capabilities" should be enough to describe what it can be used for.

src/event_loop.rs
```diff
@@ -48,6 +50,71 @@ pub struct EventLoopWindowTarget<T: 'static> {
...
+/// cases, you can use a [`Window`]; however, for certain graphics APIs (like OpenGL)
+/// you need to be able to query the display before creating a window. In this case,
+/// you end up with a "chicken and egg" problem; you need a display handle to create
+/// a window, but you need a window to access the display handle.
+///
+/// ![A chicken](https://i.imgur.com/9a5K2nz.jpg)
+///
+/// <sub>Figure 1: A chicken, representing the window in the above metaphor. Note that
+/// there are no eggs. "[Chicken February 2009-1]" by [Joaquim Alves Gaspar] is licensed
+/// under [CC BY-SA 3.0].</sub>
```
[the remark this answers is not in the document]
notgull: :-(

src/platform_impl/linux/x11/mod.rs
kchibisov: Hm, that's fair.

rib:
Hey, sorry to be late catching up with these changes but I was away for a week and missed the RFC poke on https://github.com/rust-windowing/raw-window-handle/issues/111

I'm a little worried that we may have made quite a significant design change that perhaps wasn't required for Android, while I don't currently know that there was a safety/soundness issue.

Following the bread crumbs I wrote some initial thoughts here: https://github.com/rust-windowing/raw-window-handle/issues/111#issuecomment-1493301738

a quick comment on https://github.com/rust-windowing/raw-window-handle/pull/110#issuecomment-1493321465

and a comment on https://github.com/rust-windowing/raw-window-handle/pull/116

Maybe there are multiple reasons for the changes, and i'm just looking at this from the pov of the Android backend, but still figured I should follow up

notgull:
> Hey, sorry to be late catching up with these changes but I was away for a week and missed the RFC poke on [rust-windowing/raw-window-handle#111](https://github.com/rust-windowing/raw-window-handle/issues/111)

No worries! We're still in the early days of this feature.

I responded to your posts. The overarching theme is that the [docs](https://developer.android.com/ndk/reference/struct/a-native-activity-callbacks#onnativewindowdestroyed) seem to indicate that you cannot access the native window handle after it's been dropped.

> You MUST ensure that you do not touch the window object after returning from this function

src/event_loop.rs
```diff
@@ -48,6 +50,73 @@ pub struct EventLoopWindowTarget<T: 'static> {
...
+/// [`EventLoop`]: EventLoop
+/// [`Window`]: crate::window::Window
+/// [`DisplayHandle`]: raw_window_handle::DisplayHandle
+/// [`EventLoopWindowTarget`]: EventLoopWindowTarget
+/// [`HasDisplayHandle`]: raw_window_handle::HasDisplayHandle
+#[derive(Clone)]
+pub struct OwnedDisplayHandle {
+    pub(crate) p: platform_impl::OwnedDisplayHandle,
+
+    /// Make sure that the display handle is !Send and !Sync.
```
madsmtm, line 106: Huh, are you sure that rustdoc's test system will pick this up?

kchibisov: That's how such things should be tested and are being tested in glutin.

Even normal `cargo test` will run them.

madsmtm: Ah cool, TIL!

notgull:
Superseded by #2943

PR #2747, opened 2023-03-22
Fixes rust-windowing/winit#2597 for good

Maximetinu:
Remove assert that is being randomly hit by a very small number of users

This is a follow up to PR https://github.com/rust-windowing/winit/issues/2597. As explained in it (pls check it for context) there were a couple of asserts in this file that were unnecessary, and were causing some users to crash; while removing the assert works perfectly fine, so they don't seem to be preventing from anything serious.

In that PR, I was wary and changed the smallest amount of code possible (just removed 1 assert, out of 2). But now I've found some other users (still very few, maybe 0.001%) hitting the remaining assert. Now that time has verified that removing the first one didn't hurt and was the right call, I think that removing the remaining one is the way to go, for the same reasons explained in the previous PR, to prevent these crashes from happening anymore.

### DIFF
--- a/src/platform_impl/web/web_sys/scaling.rs
+++ b/src/platform_impl/web/web_sys/scaling.rs
@@ -64,9 +64,7 @@ impl ScaleChangeDetectorInternal {
             min_scale = current_scale - 0.0001, max_scale= current_scale + 0.0001,
         );
         let mql = MediaQueryListHandle::new(&media_query, closure);
-        if let Some(mql) = &mql {
-            assert!(mql.mql().matches());
-        }
+
         mql
     }
 


Maximetinu:
Hello, can someone give a quick review to this one? We need it to be able to use official winit instead of our fork!

cc @madsmtm because you merged the previous one 🙏

kchibisov:
If you know how to fix you can post a fix, it's just will be merged with delay. Though, if it's short and doesn't conflict with KB v2, we could merge.

daxpedda:
@Maximetinu could you do what I suggested above and figure out if this isn't an issue with the offset but just a racing condition? Unfortunately I don't seem to be able to reproduce this on my end.

@kchibisov the fix shouldn't conflict with KB v2, I will make one as soon as we know more.

daxpedda:
Replaced by #2850.

Maximetinu:
Sorry @daxpedda I never answered 🙏 I was never able to reproduce neither, just got reports from users hitting this randomly. I hope #2850 fixes it, thx!

Maximetinu:
Hi @daxpedda , sorry to revive such an old thread, but, months later, I got a way to consistently reproduce this crash: https://github.com/rust-windowing/winit/issues/3579

Or, at least, it's a crash in what seems to be the same assert.

When this happened to me months ago, it was caused by resizing the screen (but not always).

As I say in this new #3579, the same assert can also be hit by opening the print dialog

PR #2749, opened 2023-03-26
Implement `Sync` for `EventLoopProxy` (except on Web)

daxpedda:
This implements `Sync` for `EventLoopProxy` for all platforms except Web, which is done in a separate PR (#2740). I tested this only on Linux and Windows, but I think it's pretty safe to say this should just work.


daxpedda:
CI is currently broken. Partly fixed by https://github.com/rust-windowing/winit/pull/2750.

daxpedda:
Closed. We decided it's kinda pointless because users can just wrap it around a `Mutex` themselves.
For reference, it was also decided that adding `crossbeam-channel` is not worth it.

PR #2750, opened 2023-03-26
Skip `syn` in `cargo-deny`

daxpedda:
Current CI is broken because of only some dependencies updating to `syn` v2, this skips it during `cargo-deny`.

This doesn't completely fix the CI because of https://github.com/madsmtm/objc2/issues/432.


kchibisov:
It's skipped now and the required issues were linked.

PR #2751, opened 2023-03-26
On Web, fix async `Window::set_fullscreen`

daxpedda:
Currently the behavior of `Window::set_fullscreen()` on Web queues the request to the current or next keyboard or mouse event. This doesn't work if `Window::set_fullscreen()` is called asynchronously outside the event loop.

This was probably done to satisfy the [transient activation](https://developer.mozilla.org/en-US/docs/Glossary/Transient_activation) [requirement](https://developer.mozilla.org/en-US/docs/Web/API/Element/requestFullscreen#security) for `requestFullscreen()`. But transient activation doesn't run out during the event loop, it's time based, usually one second.

This PR adjusts `Window::set_fullscreen()` to immediately call `requestFullscreen()`, only on failure a request is queued. This required some custom imports because the API was changed in the spec to return a `Promise` and `web-sys` doesn't reflect that change. Additionally Safari doesn't reflect that change before 16.4 (which came out today (2023-03-28)) too. So both APIs are now supported.

Additionally I noticed that the behavior of `Window::set_fullscreen()` on Web was undocumented, so I added some.

### DIFF
--- a/src/platform_impl/web/web_sys/canvas.rs
+++ b/src/platform_impl/web/web_sys/canvas.rs
@@ -10,8 +10,11 @@ use crate::platform_impl::{OsError, PlatformSpecificWindowBuilderAttributes};
 use std::cell::RefCell;
 use std::rc::Rc;
 
+use js_sys::Promise;
 use smol_str::SmolStr;
-use wasm_bindgen::{closure::Closure, JsCast};
+use wasm_bindgen::prelude::wasm_bindgen;
+use wasm_bindgen::{closure::Closure, JsCast, JsValue};
+use wasm_bindgen_futures::JsFuture;
 use web_sys::{
     AddEventListenerOptions, Event, FocusEvent, HtmlCanvasElement, KeyboardEvent,
     MediaQueryListEvent, MouseEvent, WheelEvent,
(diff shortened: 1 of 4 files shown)


daxpedda:
I decided to go ahead and implement properly handling the return value of `Element.requestFullscreen()`.
It was possible that fullscreen mode could be exited by the user without Winit, for example by pressing the button "Exit" that browsers show when you enter fullscreen, so any click or keyboard input afterwards would trigger the queued fullscreen mode again.

I had to pull in `js-sys` for this, which is not an additional dependency, `web-sys` was already relying on it. Additionally I noted in the OP and in the code on why we needed additional custom imports.

src/platform_impl/web/web_sys/canvas.rs
```diff
@@ -465,7 +468,34 @@ impl Common {
...
+
+        // This should return a `Promise`, but some Safari <v16.4 is not up-to-date with the spec.
+        match raw.request_fullscreen() {
+            Ok(value) if !value.is_undefined() => {
+                let promise: PromiseExt = value.unchecked_into();
+                let wants_fullscreen = self.wants_fullscreen.clone();
+                let cb = Closure::once_into_js(move || *wants_fullscreen.borrow_mut() = true);
+                // We use a custom `catch` function here because we don't
+                // want to deal with an async cleanup of the `Closure`.
+                let _ = promise.catch(&cb);
```
daxpedda, line 494: This workaround of dealing with the `Closure` could be removed if we ever pull in `wasm-bindgen-futures`.

daxpedda: I used `wasm-bindgen-futures` instead.

PR #2755, opened 2023-04-03
Fix potential panic

xiaopengli89:
```
uncaught panic: "attempt to divide by zero" at /Users/X20230112120246_ssea_lrtse_cpfx/.cargo/git/checkouts/winit-c2fdb27092aba5a7/3fd7384/src/platform_impl/macos/monitor.rs:233:19
```

### DIFF
--- a/src/platform_impl/macos/monitor.rs
+++ b/src/platform_impl/macos/monitor.rs
@@ -230,6 +230,10 @@ impl MonitorHandle {
                 return None;
             }
 
+            if time.time_value < 1 {
+                return None;
+            }
+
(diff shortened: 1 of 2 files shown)


src/platform_impl/macos/monitor.rs
```diff
@@ -230,6 +230,10 @@ impl MonitorHandle {
                 return None;
             }
 
+            if time.time_value < 1 {
+                return None;
+            }
+
```
daxpedda, line 236: Can we use [`checked_div()`](https://doc.rust-lang.org/std/primitive.i64.html#method.checked_div) instead?

madsmtm (approved):
Good catch, thanks!

PR #2758, opened 2023-04-04
Fix device description leak

xiaopengli89:
`deviceDescription` should be called in autoreleasepool.

图片


madsmtm (approved):
Yeah, that works, thanks for the contribution!

PR #2764, opened 2023-04-10
Update the `bitflags` dependency

burtonageo:
- [❎] Tested on all platforms changed
- [✅] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users
- [❎] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior
- [❎] Created or updated an example program if it would help users understand this functionality
- [❎] Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented

I used the largest set of `#[derive()]` traits on each `bitflags!` instance, so there should be no/minimal compile errors.
I have tested on windows, iOS, mac, and wasm. I have not yet been able to test on Linux or Android, but I used the widest range of derives on those platforms, so they should hopefully not break.

### DIFF
--- a/src/platform_impl/macos/appkit/event.rs
+++ b/src/platform_impl/macos/appkit/event.rs
@@ -194,7 +194,7 @@ const NX_DEVICERALTKEYMASK: u32 = 0x00000040;
 const NX_DEVICERCTLKEYMASK: u32 = 0x00002000;
 
 bitflags! {
-    #[derive(Clone, Copy, Default, Debug, Eq, Hash, PartialEq)]
+    #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
     pub struct NSEventModifierFlags: NSUInteger {
         const NSAlphaShiftKeyMask                     = 1 << 16;
         const NSShiftKeyMask                          = 1 << 17;
@@ -213,7 +213,7 @@ unsafe impl Encode for NSEventModifierFlags {
 }
 
 bitflags! {
-    #[derive(Clone, Copy, Default, Debug, Eq, Hash, PartialEq)]
+    #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
     pub struct NSEventPhase: NSUInteger {
        const NSEventPhaseNone        = 0;
        const NSEventPhaseBegan       = 0x1 << 0;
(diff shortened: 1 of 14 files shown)


madsmtm (commented):
As far as I can see, the public `ModifiersState` and `WindowButtons` are affected by this change, right? I think the `bitflags!` macro adds some methods to these, could you check that these have not regressed (if they have, this would have to be a breaking change)

kchibisov:
Any update like that should go after #2662

kchibisov:
Could you rebase?

madsmtm:
Should be marked as a breaking change, the implemented traits for the types have changed.

daxpedda (approved):

src/platform_impl/macos/appkit/event.rs
```diff
@@ -212,6 +213,7 @@ unsafe impl Encode for NSEventModifierFlags {
 }
 
 bitflags! {
+    #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
```
kchibisov, line 216: Could you change the order the way it's written? Usually it's `Debug, Clone, Copy, PartialEq, Eq, Hash` The only justified change is `Clone, Copy` and `PartialEq, Eq` when it comes to ordering, because they of `Lesser -> More` order and LSP inserts both of them that way.

```suggestion
    #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
```

daxpedda: > Usually it's `Debug, Clone, Copy, PartialEq, Eq, Hash`

Where does `Default` fit in that order?
> [..], because they of `Lesser -> More` order and LSP inserts both of them that way.

Whats `Lesser` and `More`? What's LSP? Do you mean Rust Analyzer?

kchibisov: Doesn't really matter for default tbh.

kchibisov: >Whats Lesser and More? What's LSP? Do you mean Rust Analyzer?

Lesser -> more is about semantics they give, `Copy` is extension of `Clone`, And `Eq` is extension of `PartialEq` if you look at them.

`LSP` is a language server protocol, so rust-analayzer in that case.

kchibisov (approved):

PR #2765, opened 2023-04-10
Add dependabot support

notgull:
Adds dependabot support to keep dependencies up-to-date. This should make sure things are automatically bumped, in the style of #2764


kchibisov:
I'm not sure I like dependabot that much, so we'll go without it.

In general I prefer to manually check deps of winit and given that we're not a binary, patches/minors are being applied automatically.

PR #2770, opened 2023-04-14
Add MouseButton::{Back, Forward} to mouse input events on Wayland, X11, Windows, macOS and Web

### LINKED ISSUE #1785: Cross platfrom mouse button translation
Right now winit only provides translation for `Left`, `Right`, and `Middle,` mouse buttons and everything else is via `Other`, however each platform defines their own value for other for the same buttons, it'll be nice to agree and extend  `MouseButton` enum to include extra mouse buttons as well.

bbb651:
Fixes #1785. This is a breaking change because MouseButton isn't #[non_exhaustive],
it's supported on wayland, x11, windows, macos and web, all other platforms (orbital doesn't support these buttons at all, android and ios don't have mouse support in winit). 

- [ ] Tested on all platforms changed (currently tested on x11 (xwayland), needs testing on other platforms)
- [x] Created or updated an example program if it would help users understand this functionality (I don't think it's necessary for this kind of change, it's fairly self explanatory)


i509VCB (commented):
Tested with xev and the values for X11 appear to be correct.

src/platform_impl/linux/wayland/seat/pointer/handlers.rs
```diff
@@ -171,6 +175,8 @@ pub(super) fn handle_pointer(
                 BTN_LEFT => MouseButton::Left,
                 BTN_RIGHT => MouseButton::Right,
                 BTN_MIDDLE => MouseButton::Middle,
+                BTN_BACK | BTN_SIDE => MouseButton::Back,
+                BTN_FORWARD | BTN_EXTRA => MouseButton::Forward,
```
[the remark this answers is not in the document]
bbb651: I just checked what events my mouse (Logitech G102) generates with python-evdev and the back button is `BTN_SIDE` and the forward button is `BTN_EXTRA`.

madsmtm (changes requested):
macOS impl looks simple enough, though I can't test it since I don't have a mouse with those buttons.

Maybe you could document the variants a bit more to describe what "forward" and "backward" mouse buttons actually mean (or link to a Wikipedia article or something), I suspect I'm not the only one that don't know how those actually look

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- `MouseButton` now supports `Back` and `Forward` varients, emitted from mouse events on Wayland, X11, Windows, macOS and Web.
```
madsmtm, line 11: Should be marked as breaking

bbb651:
> macOS impl looks simple enough, though I can't test it since I don't have a mouse with those buttons.
> 
> Maybe you could document the variants a bit more to describe what "forward" and "backward" mouse buttons actually mean (or link to a Wikipedia article or something), I suspect I'm not the only one that don't know how those actually look

Looking more into it I think macOS doesn't officially support these buttons, I got them from [this project](https://github.com/archagon/sensible-side-buttons/) which looking at it now it's a hack specifically to address this shortcoming. 

I found [this documentation page about possible mouse buttons](https://developer.apple.com/documentation/coregraphics/cgmousebutton):
> Quartz supports up to 32 mouse buttons. The first three buttons are specified using these three constants. Additional buttons are specified in USB order using the integers 3 to 31.

The question now is if treating `3` and `4` as back and forward is a common enough convention that we should adopt it. After searching for way too long I found that at least [chromium does it](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/renderer_host/input/web_input_event_builders_mac.mm;l=113) so it seems pretty sensible to support it. But yeah it definitely needs a comment explaining this and documentation for users that it's not guaranteed to work across all configurations. I will fix this and the changelog tomorrow.

examples/mouse_buttons.rs
```diff
@@ -0,0 +1,36 @@
+use simple_logger::SimpleLogger;
```
kchibisov, line 1: We don't need example just for that.

CHANGELOG.md
kchibisov: bump

CHANGELOG.md
```diff
@@ -8,7 +8,8 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
-- Bump MSRV from `1.60` to `1.64`.
+- `MouseButton` now supports `Back` and `Forward` varients, emitted from mouse events on Wayland, X11, Windows, macOS and Web. 
+- Bump MSRV from `1.60` to `1.64`. This is a **breaking** change.
```
kchibisov, line 12: Will need a rebase correcting this.

src/event.rs
```diff
@@ -1055,12 +1055,21 @@ pub enum ElementState {
 }
 
 /// Describes a button of a mouse controller.
+///
+/// ## Platform-specific
+///
+/// **macOS:** does not offically define `Back` and `Forward`, but supports 29 generic mouse buttons from 3..=31.
+/// `3` and `4` is commonly used for `Back` and `Forward` respectivly by driver vendor and applications such as browsers,
+/// this follows this convention. Not guaranteed to work with all hardware and configurations.
+/// **orbital:** `Back` and `Forward` are unsupported due to orbital not supporting them.
```
kchibisov, line 1064:
```suggestion
/// **Orbital:** `Back` and `Forward` are unsupported due to orbital not supporting them.
```

src/event.rs
```diff
@@ -1055,12 +1055,21 @@ pub enum ElementState {
 }
 
 /// Describes a button of a mouse controller.
+///
+/// ## Platform-specific
+///
+/// **macOS:** does not offically define `Back` and `Forward`, but supports 29 generic mouse buttons from 3..=31.
+/// `3` and `4` is commonly used for `Back` and `Forward` respectivly by driver vendor and applications such as browsers,
+/// this follows this convention. Not guaranteed to work with all hardware and configurations.
```
kchibisov, line 1063: I think this is overly verbose for the users providing little value. I'd suggest it to add to macOS implementation instead in the `match` you've changed. And here I'd just say that it `might now work with all hardware`.

kchibisov (changes requested):
let me know if you still want to work on that, otherwise I'll push it myself.

src/platform_impl/web/web_sys/event.rs
```diff
@@ -5,10 +5,13 @@ use std::convert::TryInto;
 use web_sys::{HtmlCanvasElement, KeyboardEvent, MouseEvent, PointerEvent, WheelEvent};
 
 pub fn mouse_button(event: &MouseEvent) -> MouseButton {
+    // https://w3c.github.io/uievents/#dom-mouseevent-button
```
daxpedda, line 8:
```suggestion
    // https://www.w3.org/TR/uievents/#dom-mouseevent-button
```
We should link to the W3C instead of the editors draft.

src/platform_impl/web/web_sys/event.rs
```diff
@@ -5,10 +5,13 @@ use std::convert::TryInto;
...
     match event.button() {
         0 => MouseButton::Left,
         1 => MouseButton::Middle,
         2 => MouseButton::Right,
+        3 => MouseButton::Back,
+        4 => MouseButton::Forward,
         i => MouseButton::Other((i - 3).try_into().expect("very large mouse button value")),
```
daxpedda, line 15: This was done a very long time ago, but why is this subtracting 3 unlike other backends?
Can we fix this here while we are at it?

daxpedda (changes requested):
LGTM except the nits.

src/platform_impl/web/web_sys/event.rs
daxpedda: I fixed this in https://github.com/rust-windowing/winit/pull/2838.

daxpedda:
@bbb651 ping me if you want me to take care of the web backend.

bbb651:
@daxpedda I'd appropriate it

CHANGELOG.md
kchibisov: The rebase went wrong here. You've added lines to old release.

examples/mouse_buttons.rs
kchibisov: bump, just remove this example.

src/platform_impl/linux/wayland/seat/pointer/handlers.rs
```diff
@@ -0,0 +1,333 @@
+//! Handlers for the pointers we're using.
+
+use std::cell::RefCell;
```
kchibisov, line 3: This file doesn't exist in the tree, would suggest to simply remove your changes and add them again.

src/platform_impl/web/web_sys/event.rs
```diff
@@ -6,47 +6,15 @@ use smol_str::SmolStr;
 use std::convert::TryInto;
 use web_sys::{HtmlCanvasElement, KeyboardEvent, MouseEvent, PointerEvent, WheelEvent};
 
-bitflags! {
-    #[derive(Debug, Clone, Copy, PartialEq, Eq)]
```
kchibisov, line 10: That removal looks a bit suspicious.

kchibisov (changes requested):
Seems like your rebase went wrong.

src/platform_impl/web/web_sys/event.rs
daxpedda: I'm gonna deal with this in a bit.

daxpedda (approved):
@bbb651 Web backend is done.
(only approving the Web backend here)

madsmtm (approved):
MacOS looks good now

msiglreith (approved):
Windows changes also fine

examples/mouse_buttons.rs
daxpedda: Done.

src/platform_impl/linux/wayland/seat/pointer/handlers.rs
daxpedda: Done, let me know if I messed up something.

kchibisov (approved):

kchibisov:
Thanks!

PR #2777, opened 2023-04-24
Replace mio with calloop in the X11 backend

### LINKED ISSUE #2763: Replace mio with calloop in the X11 backend
Right now, there are two backends for Free Unix systems; one for X11 and one for Wayland. The X11 backend uses `mio` to query for display readiness and query for other things (such as user events), while the Wayland backend uses the `calloop` crate in a very similar way.

It would be nice for dependency reduction if the X11 backend used `calloop` instead. In fact, there is probably a lot of code that can be shared between the backends if both used `calloop`.

notgull:

Closes #2763 by replacing `mio` with `calloop` in the X11 backend.

### DIFF
--- a/src/platform_impl/linux/x11/mod.rs
+++ b/src/platform_impl/linux/x11/mod.rs
@@ -493,7 +493,7 @@ impl<T: 'static> EventLoop<T> {
                     // We don't go straight into executing the event loop iteration, we instead go
                     // to the start of this loop and check again if there's any pending event. We
                     // must do this because during the execution of the iteration we sometimes wake
-                    // the mio waker, and if the waker is already awaken before we call poll(),
+                    // the calloop waker, and if the waker is already awaken before we call poll(),
                     // then poll doesn't block, but it returns immediately. This caused the event
                     // loop to run continuously even if the control_flow was `Wait`
                     continue;
--- a/src/platform_impl/linux/x11/mod.rs
+++ b/src/platform_impl/linux/x11/mod.rs
@@ -19,8 +19,7 @@ pub use self::xdisplay::{XError, XNotSupported};
 
 use calloop::channel::{channel, Channel, Event as ChanResult, Sender};
 use calloop::generic::Generic;
-use calloop::Dispatcher;
-use calloop::EventLoop as Loop;
+use calloop::{Dispatcher, EventLoop as Loop};
 
 use std::{
     cell::{Cell, RefCell},
@@ -212,16 +211,25 @@ impl<T: 'static> EventLoop<T> {
         // Create an event loop.
         let event_loop =
             Loop::<EventLoopState<T>>::try_new().expect("Failed to initialize the event loop");
+        let handle = event_loop.handle();
 
         // Create the X11 event dispatcher.
         let source = X11Source::new(xconn.x11_fd, calloop::Interest::READ, calloop::Mode::Level);
-
-        let handle = event_loop.handle();
         handle
             .insert_source(source, |_, _, _| Ok(calloop::PostAction::Continue))
             .expect("Failed to register the X11 event dispatcher");
 
+        // Create a channel for sending user events.
         let (user_sender, user_channel) = channel();
+        handle
+            .insert_source(user_channel, |ev, _, state| {
+                if let ChanResult::Msg(user) = ev {
+                    state.user_events.push_back(user);
+                }
+            })
+            .expect("Failed to register the user event channel with the event loop");
+
+        // Create a channel for handling redraw requests.
         let (redraw_sender, redraw_channel) = channel();
 
         // Create a dispatcher for the redraw channel such that we can dispatch it independent of the
@@ -232,15 +240,6 @@ impl<T: 'static> EventLoop<T> {
                     state.redraw_events.push_back(window_id);
                 }
             });
-
-        // Register the channels with the event loop.
-        handle
-            .insert_source(user_channel, |ev, _, state| {
-                if let ChanResult::Msg(user) = ev {
-                    state.user_events.push_back(user);
-                }
-            })
-            .expect("Failed to register the user event channel with the event loop");
         handle
             .register_dispatcher(redraw_dispatcher.clone())
             .expect("Failed to register the redraw event channel with the event loop");
--- a/src/platform_impl/linux/x11/mod.rs
+++ b/src/platform_impl/linux/x11/mod.rs
@@ -383,9 +383,20 @@ impl<T: 'static> EventLoop<T> {
                     callback,
                 );
             }
+
+            // Quickly dispatch all pending events to clear channels
+            this.event_loop
+                .dispatch(
+                    Some(Duration::ZERO),
+                    &mut get_xtarget(&this.target).state.borrow_mut(),
+                )
+                .ok();
+
             // Empty the redraw requests
             {
                 let mut windows = HashSet::new();
+
+                // Empty the channel.
 
                 while let Some(window_id) = get_xtarget(&this.target)
                     .state
--- a/src/platform_impl/linux/x11/mod.rs
+++ b/src/platform_impl/linux/x11/mod.rs
@@ -19,22 +19,25 @@ pub use self::xdisplay::{XError, XNotSupported};
 
 use std::{
     cell::{Cell, RefCell},
-    collections::{HashMap, HashSet},
+    collections::{HashMap, HashSet, VecDeque},
     ffi::CStr,
     mem::{self, MaybeUninit},
     ops::Deref,
     os::raw::*,
+    os::unix::io::RawFd,
     ptr,
     rc::Rc,
     slice,
-    sync::mpsc::{Receiver, Sender, TryRecvError},
     sync::{mpsc, Arc, Weak},
     time::{Duration, Instant},
 };
 
 use libc::{self, setlocale, LC_CTYPE};
 
-use mio::{unix::SourceFd, Events, Interest, Poll, Token, Waker};
+use calloop::channel::{channel, Event as ChanResult, Sender};
+use calloop::generic::Generic;
+use calloop::EventLoop as Loop;
+
 use raw_window_handle::{RawDisplayHandle, XlibDisplayHandle};
 
 use self::{
@@ -54,81 +57,46 @@ use crate::{
     window::WindowAttributes,
 };
 
-const X_TOKEN: Token = Token(0);
-const USER_REDRAW_TOKEN: Token = Token(1);
-
-struct WakeSender<T> {
-    sender: Sender<T>,
-    waker: Arc<Waker>,
-}
-
-struct PeekableReceiver<T> {
-    recv: Receiver<T>,
-    first: Option<T>,
-}
-
-impl<T> PeekableReceiver<T> {
-    pub fn from_recv(recv: Receiver<T>) -> Self {
-        Self { recv, first: None }
-    }
-    pub fn has_incoming(&mut self) -> bool {
-        if self.first.is_some() {
-            return true;
-        }
-
-        match self.recv.try_recv() {
-            Ok(v) => {
-                self.first = Some(v);
-                true
-            }
-            Err(TryRecvError::Empty) => false,
-            Err(TryRecvError::Disconnected) => {
-                warn!("Channel was disconnected when checking incoming");
-                false
-            }
-        }
-    }
-    pub fn try_recv(&mut self) -> Result<T, TryRecvError> {
-        if let Some(first) = self.first.take() {
-            return Ok(first);
-        }
-        self.recv.try_recv()
-    }
-}
+type X11Source = Generic<RawFd>;
+//type X11Dispatcher = calloop::Dispatcher<'static, X11Source, EventLoopState>;
 
 pub struct EventLoopWindowTarget<T> {
     xconn: Arc<XConnection>,
+    state: RefCell<EventLoopState<T>>,
     wm_delete_window: ffi::Atom,
     net_wm_ping: ffi::Atom,
     ime_sender: ImeSender,
     root: ffi::Window,
     ime: RefCell<Ime>,
     windows: RefCell<HashMap<WindowId, Weak<UnownedWindow>>>,
-    redraw_sender: WakeSender<WindowId>,
+    redraw_sender: Sender<WindowId>,
     device_events: Cell<DeviceEvents>,
     _marker: ::std::marker::PhantomData<T>,
 }
 
 pub struct EventLoop<T: 'static> {
-    poll: Poll,
-    waker: Arc<Waker>,
+    event_loop: Loop<'static, EventLoopState<T>>,
     event_processor: EventProcessor<T>,
-    redraw_receiver: PeekableReceiver<WindowId>,
-    user_receiver: PeekableReceiver<T>, //waker.wake needs to be called whenever something gets sent
     user_sender: Sender<T>,
     target: Rc<RootELW<T>>,
 }
 
+struct EventLoopState<T> {
+    /// Incoming user events.
+    user_events: VecDeque<T>,
+
+    /// Incoming redraw events.
+    redraw_events: VecDeque<WindowId>,
+}
+
 pub struct EventLoopProxy<T: 'static> {
     user_sender: Sender<T>,
-    waker: Arc<Waker>,
 }
 
 impl<T: 'static> Clone for EventLoopProxy<T> {
     fn clone(&self) -> Self {
         EventLoopProxy {
             user_sender: self.user_sender.clone(),
-            waker: self.waker.clone(),
         }
     }
 }
@@ -237,21 +205,48 @@ impl<T: 'static> EventLoop<T> {
 
         xconn.update_cached_wm_info(root);
 
-        let poll = Poll::new().unwrap();
-        let waker = Arc::new(Waker::new(poll.registry(), USER_REDRAW_TOKEN).unwrap());
+        // Create an event loop.
+        let event_loop =
+            Loop::<EventLoopState<T>>::try_new().expect("Failed to initialize the event loop");
 
-        poll.registry()
-            .register(&mut SourceFd(&xconn.x11_fd), X_TOKEN, Interest::READABLE)
-            .unwrap();
+        // Create the X11 event dispatcher.
+        let source = X11Source::new(xconn.x11_fd, calloop::Interest::READ, calloop::Mode::Edge);
+        let dispatcher =
+            calloop::Dispatcher::new(source, |_, _, _| Ok(calloop::PostAction::Continue));
 
-        let (user_sender, user_channel) = std::sync::mpsc::channel();
-        let (redraw_sender, redraw_channel) = std::sync::mpsc::channel();
+        let handle = event_loop.handle();
+        handle
+            .register_dispatcher(dispatcher.clone())
+            .expect("Failed to register the X11 event dispatcher");
+
+        let (user_sender, user_channel) = channel();
+        let (redraw_sender, redraw_channel) = channel();
+
+        // Register the channels with the event loop.
+        handle
+            .insert_source(user_channel, |ev, _, state| {
+                if let ChanResult::Msg(user) = ev {
+                    state.user_events.push_back(user);
+                }
+            })
+            .expect("Failed to register the user event channel with the event loop");
+        handle
+            .insert_source(redraw_channel, |ev, _, state| {
+                if let ChanResult::Msg(window_id) = ev {
+                    state.redraw_events.push_back(window_id);
+                }
+            })
+            .expect("Failed to register the redraw event channel with the event loop");
 
         let kb_state =
             KbdState::from_x11_xkb(unsafe { (xconn.xlib_xcb.XGetXCBConnection)(xconn.display) })
                 .unwrap();
 
         let window_target = EventLoopWindowTarget {
+            state: RefCell::new(EventLoopState {
+                user_events: VecDeque::new(),
+                redraw_events: VecDeque::new(),
+            }),
             ime,
             root,
             windows: Default::default(),
@@ -260,10 +255,7 @@ impl<T: 'static> EventLoop<T> {
             xconn,
             wm_delete_window,
             net_wm_ping,
-            redraw_sender: WakeSender {
-                sender: redraw_sender, // not used again so no clone
-                waker: waker.clone(),
-            },
+            redraw_sender,
             device_events: Default::default(),
         };
 
@@ -310,11 +302,8 @@ impl<T: 'static> EventLoop<T> {
         event_processor.init_device(ffi::XIAllDevices);
 
         EventLoop {
-            poll,
-            waker,
+            event_loop,
             event_processor,
-            redraw_receiver: PeekableReceiver::from_recv(redraw_channel),
-            user_receiver: PeekableReceiver::from_recv(user_channel),
             user_sender,
             target,
         }
@@ -323,7 +312,6 @@ impl<T: 'static> EventLoop<T> {
     pub fn create_proxy(&self) -> EventLoopProxy<T> {
         EventLoopProxy {
             user_sender: self.user_sender.clone(),
-            waker: self.waker.clone(),
         }
     }
 
@@ -372,7 +360,12 @@ impl<T: 'static> EventLoop<T> {
 
             // Empty the user event buffer
             {
-                while let Ok(event) = this.user_receiver.try_recv() {
+                while let Some(event) = get_xtarget(&this.target)
+                    .state
+                    .borrow_mut()
+                    .user_events
+                    .pop_front()
+                {
                     sticky_exit_callback(
                         crate::event::Event::UserEvent(event),
                         &this.target,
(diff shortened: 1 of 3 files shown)


kchibisov:
Could you rebase?

notgull:
Rebased on the latest master!

src/platform_impl/linux/x11/window.rs
```diff
@@ -27,9 +27,9 @@ use crate::{
 };
 
 use super::{
-    ffi, util, EventLoopWindowTarget, ImeRequest, ImeSender, WakeSender, WindowId, XConnection,
-    XError,
+    ffi, util, EventLoopWindowTarget, ImeRequest, ImeSender, WindowId, XConnection, XError,
 };
+use calloop::channel::Sender;
```
kchibisov, line 32: this should be around the line `13` with other crate uses.

src/platform_impl/linux/x11/mod.rs
```diff
@@ -374,7 +361,12 @@ impl<T: 'static> EventLoop<T> {
 
             // Empty the user event buffer
             {
-                while let Ok(event) = this.user_receiver.try_recv() {
+                while let Some(event) = get_xtarget(&this.target)
+                    .state
+                    .borrow_mut()
```
kchibisov, line 366: I really don't like those `borrow_mut`, they should generally be `get_mut()`, though, due to flowed design of x11 backend it's sort of that way due to `Rc`(which could easily be removed from what I remember).

src/platform_impl/linux/x11/mod.rs
```diff
@@ -392,11 +384,27 @@ impl<T: 'static> EventLoop<T> {
...
+                // Empty the channel.
+
+                while let Some(window_id) = get_xtarget(&this.target)
+                    .state
+                    .borrow_mut()
+                    .redraw_events
+                    .pop_front()
+                {
                     windows.insert(window_id);
                 }
```
kchibisov, line 409: Is the logic to push `redraw` events further here? I'd probably suggest to dispatch just redraw requests if that's the intention, so the further logic with empty checks could be eliminated.

Though, it's a matter of preference, maybe we want to dispatch all of them.

src/platform_impl/linux/x11/mod.rs
```diff
@@ -239,21 +206,48 @@ impl<T: 'static> EventLoop<T> {
...
         xconn.update_cached_wm_info(root);
 
-        let poll = Poll::new().unwrap();
-        let waker = Arc::new(Waker::new(poll.registry(), USER_REDRAW_TOKEN).unwrap());
+        // Create an event loop.
+        let event_loop =
+            Loop::<EventLoopState<T>>::try_new().expect("Failed to initialize the event loop");
+
+        // Create the X11 event dispatcher.
+        let source = X11Source::new(xconn.x11_fd, calloop::Interest::READ, calloop::Mode::Edge);
```
kchibisov, line 214: Do we really want `Edge`? In general `Level` is used, because multiple threads could contest the connection.

kchibisov (changes requested):

src/platform_impl/linux/x11/mod.rs
notgull: Okay, I've moved the `state` out of the window target and into the event loop. This should be fine for now.

src/platform_impl/linux/x11/mod.rs
notgull: Sounds good, I've moved the redraw channel into a `Dispatcher` to make this easier.

src/platform_impl/linux/x11/mod.rs
```diff
@@ -239,15 +211,41 @@ impl<T: 'static> EventLoop<T> {
...
+        handle
+            .insert_source(user_channel, |ev, _, state| {
+                if let ChanResult::Msg(user) = ev {
+                    state.user_events.push_back(user);
+                }
+            })
+            .expect("Failed to register the user event channel with the event loop");
+        handle
+            .register_dispatcher(redraw_dispatcher.clone())
+            .expect("Failed to register the redraw event channel with the event loop");
```
kchibisov, line 248: Could we have source/dispatcher insert with the transfer point created, so `user_sender` -> `insert_source` -> `empty line` -> `redraw_sender` -> `dispatcher`, it's just  looks weird that we create dispatcher, then insert user source, and then insert dispatcher.

kchibisov (changes requested):

kchibisov:
Would need to apply something like 
```
commit 8e25d04442328022ffbb407a6826acc4743ce337
Author: Kirill Chibisov <contact@kchibisov.com>
Date:   Tue May 30 12:51:11 2023 +0300

    Fix comment

diff --git a/src/platform_impl/linux/x11/mod.rs b/src/platform_impl/linux/x11/mod.rs
index 4d7a1010..99f7aa07 100644
--- a/src/platform_impl/linux/x11/mod.rs
+++ b/src/platform_impl/linux/x11/mod.rs
@@ -495,7 +495,7 @@ impl<T: 'static> EventLoop<T> {
                     // We don't go straight into executing the event loop iteration, we instead go
                     // to the start of this loop and check again if there's any pending event. We
                     // must do this because during the execution of the iteration we sometimes wake
-                    // the mio waker, and if the waker is already awaken before we call poll(),
+                    // the waker, and if the waker is already awaken before we call poll(),
                     // then poll doesn't block, but it returns immediately. This caused the event
                     // loop to run continuously even if the control_flow was `Wait`
                     continue;

```

kchibisov (approved):

rib:
This is going to require me to rework the X11 changes for the pump_events PR which is going to be quite awkward.

Personally I think I prefer the use of mio which gives a more direct file descriptor poll API which gives the winit backend more explicit control of when it blocks to poll and when it dispatches events which seems like important control to keep inside winit itself.

The wayland backend seems like a case in point here because it currently has bugs with how it flushes events and polls that seem like they would be easier to solve (imho) if it didn't use calloop and used a more direct polling API, such as mio.

Is there a specific issue that is addressed by switching to use calloop?

I tend to think that it should be a core responsibility of winit itself to provide an event loop and be in full control of exactly when the loop blocks and when it dispatches events so it doesn't generally seem ideal to me that we would layer on an external event loop api like calloop vs handling polling directly in winit.

notgull:
> Is there a specific issue that is addressed by switching to use calloop?

The idea that I'm moving towards with this PR is a more centralized `EventLoop` implementation among both of the current Unix implementations. There would be some `event_loop.rs` file in `src/platform_impl/linux`, and the X11/Wayland/KMSDRM/etc implementations would just register their displays into it. This would de-duplicate a significant amount of code between the two backends.

As I understand it, the current Wayland backed is pretty inexorably tied to `calloop`. If there was a way to decouple Wayland from `calloop` and use `mio` or `polling` or even just the `poll()` syscall directly, I would be fine with it. However, it seems like it would be much easier to just "upgrade" the X11 backend to `calloop` than to go in the other direction.

> Personally I think I prefer the use of mio which gives a more direct file descriptor poll API which gives the winit backend more explicit control of when it blocks to poll and when it dispatches events which seems like important control to keep inside winit itself.

The way that we use `calloop` at the moment in both backends (or at least, as far as my understanding of the Wayland backend goes) is pretty similar to just using it as an event dispatcher. What specifically about `mio` makes it easier to build a `pump_events` API around?

kchibisov:
Be aware that with `mio` the only thing you have is Edge triggered polling and nothing more, I also don't see any issue with calloop given that we really need a way to pass a state around in callbacks, because calloop was designed to handle Wayland model in the first place and fits nicely.

kchibisov:
>I tend to think that it should be a core responsibility of winit itself to provide an event loop and be in full control of exactly when the loop blocks and when it dispatches events so it doesn't generally seem ideal to me that we would layer on an external event loop api like calloop vs handling polling directly in winit.

But you have that control with calloop? You dispatch once you call dispatch and the blocking is a part of it, are you talking about blocking for each event source individually or something like that, I don't really understand what you'd expect? If anything, we'd use `polling` and not `mio`.

rib:
> Be aware that with `mio` the only thing you have is Edge triggered polling and nothing more, I also don't see any issue with calloop given that we really need a way to pass a state around in callbacks, because calloop was designed to handle Wayland model in the first place and fits nicely.

Yeah edge triggered is the lower common denominator though - anything that could be done with level triggered should still possible with edge triggering.

My understanding is that the wayland backend used to be based on mio and I think you've pointed at the use of edge triggering a few times as a reason for why it had issues but I've never really seen anything to suggest edge triggering is incompatible with wayland.

There is a very specific protocol that's documented for allowing multithreaded access to wayland sockets by multiple consumers that ensures you only ever have one thing trying to read. I think level triggering should be perfectly suited to this since the design of the protocol naturally ensures that whoever gets granted read access to the socket is responsible for reading everything and queuing it for all other clients so clients shouldn't ever be dependent on edge triggering as far as I can tell.

The wayland protocol for accessing the socket is a bit more fiddly / specific than for other systems and from the last time I was trying to investigate the flush workaround that exists in the winit backend I was fairly sure the calloop source didn't implement the documented protocol correctly but I also feel that the calloop source abstraction makes it more complex to follow the logic and be confident about that.

On the other hand I feel like I would have been able to more quickly address the need for the flush workaround in the wayland winit backend if the backend were using a more direct poll api, such as `polling` or `mio`.

I don't really feel like calloop is particularly designed for wayland - it looks like it's inspired by the glib event loop except e.g. without source priorities or a way to affect the polling timeout when preparing to poll. The design should be able to support wayland but still the added abstraction does make it more complex to reason about the wayland socket reading protocol.

kchibisov:
>I don't really feel like calloop is particularly designed for wayland - it looks like it's inspired by the glib event loop except e.g. without source priorities or a way to affect the polling timeout when preparing to poll. The design should be able to support wayland but still the added abstraction does make it more complex to reason about the wayland socket reading protocol.

The wayland-rs and calloop was entirely written by the same person after multiple iterations on wayland-rs design. `calloop` is using polling and could probably expose it to some degree. The way reading is done is entirely controlled by you in the end with calloop, you simply write the source for it, calloop does very little special when it comes to that and just ensures things like state passing around and have event loop based timers, which are required for Wayland backend.

rib:
> The idea that I'm moving towards with this PR is a more centralized `EventLoop` implementation among both of the current Unix implementations. There would be some `event_loop.rs` file in `src/platform_impl/linux`, and the X11/Wayland/KMSDRM/etc implementations would just register their displays into it. This would de-duplicate a significant amount of code between the two backends.
> 
> As I understand it, the current Wayland backed is pretty inexorably tied to `calloop`. If there was a way to decouple Wayland from `calloop` and use `mio` or `polling` or even just the `poll()` syscall directly, I would be fine with it. However, it seems like it would be much easier to just "upgrade" the X11 backend to `calloop` than to go in the other direction.

I tend to think it'd be good to go with a lower-level polling abstraction for this.

Calloop is a more high-level event loop in itself (at least that's the way I see it) that adds the notion of sources, similar to a `glib`  main loop, but calloop also isn't a particularly battle tested event loop design and considering its similarity to glib it's also missing features like prioritization (something you also have with runloops which also have similarly pluggable sources).

It makes me a bit uneasy knowing that the Wayland Winit backend currently has to implement a workaround because the calloop wayland event source isn't really working correctly on its own. Even with the current flush workaround I'm fairly sure that the wayland backend is not correctly implementing the cooperative socket reading protocol that ensures that multiple threads can safely share a single wayland socket.

I think this idea of extensible sources for a mainloop can be a good higher-level abstraction for applications but I also think Winit's core responsibility is to provide an event loop that's integrated with windowing systems and shouldn't be layering on top of another, external, higher-level event loop API. If we ever want to expose something similar to calloop/glib/runloop sources to applications then I think those concepts should instead be brought into Winit proper.

I personally think it would be worth looking at switching in the other direction and switching the Wayland backend to `mio` or (`polling` + some form of waker). The wayland backend certainly isn't inexorably tied to `calloop` from a technical pov but I know that @kchibisov has always strongly objected to any suggestion of using `mio`. I found it quite tricky trying to reason about how the existing calloop wayland source was, or was not, implementing the cooperative reading protocol and eventually I didn't really have enough time to spend on that and just gave up and instead maintained the existing flush workaround.

In retrospect I think the control flow would be simpler to follow without the calloop source abstraction and currently Winit's wayland backend doesn't really need to track multiple sources as far as I know - it just needs to poll one wayland socket and needs a way to wake polling up from another thread.

kchibisov:
>In retrospect I think the control flow would be simpler to follow without the calloop source abstraction and currently Winit's wayland backend doesn't really need to track multiple sources as far as I know - it just needs to poll one wayland socket and needs a way to wake polling up from another thread.

We do track multiple sources and create/remove sources on demand because you need to handle keyboard repeat(cursor animations), which is done by timer source in calloop(it builds on epoll/kqueue high resolution timers).

>Calloop is a more high-level event loop in itself (at least that's the way I see it) that adds the notion of sources, similar to a glib main loop, but calloop also isn't a particularly battle tested event loop design and considering its similarity to glib it's also missing features like prioritization (something you also have with runloops which also have similarly pluggable sources).

If something doesn't really work it should be fixed not replaced. The mio is being replaced because we use calloop/polling and mio refused to have level/oneshot modes. We also have polling maintainer working on winit from time to time, so I better stick to what we at least slight control over. Which is true for most dependency in winit tree if look at them(I , at least with Wayland/X11 we sort of control all the crates used by us.

I'm opened to see how `polling` based event loop on Wayland will look like, so if you want to try that then go ahead. It's indeed true that calloop could make it less flexible when inserting other event sources into winit, which is an issue right now. I'd really like to see a way to plug `inotify` and such to winit's event loop so you won't spawn threads for such stuff and could be single threaded, but be aware that the amount of work that should be done by the backend and maintained will be way more and close enough to what calloop does, in my opinion.

rib:
> The way that we use `calloop` at the moment in both backends (or at least, as far as my understanding of the Wayland backend goes) is pretty similar to just using it as an event dispatcher. What specifically about `mio` makes it easier to build a `pump_events` API around?

It's not necessarily a lot more difficult but I don't really like that it feels like the insertion of a black box around the `poll` and I don't like the loss of precise control at the core of the Winit event loop which I fully believe should be Winit's responsibility, not something to be pushed to an external libarary.

When I started working on the pump_events changes I had left the wayland backend to last because I really assumed it was going to be the simplest backend to update after having done the android + x11 backends and was a bit disappointed to first be faced with an awkward workaround for calloop and then an awkward sidequest to try and unravel why that workaround was needed.

It was doubly awkward to investigate the calloop wayland source at the time because there was also a PR I didn't initially know about that was upgrading Winit to a new version of calloop which also involved changes to the external wayland source. So the things I was investigating where also changing under my feet.

I found that all a lot more complex than what I had expected to see in the Wayland backend - enough so that I didn't feel like I had the time left over to follow up on trying to debug and fix the upstream calloop source for wayland.

Overall for `mio` (or maybe `polling`) vs `calloop` I think it's just a question of abstraction and control from my perspective. I would prefer to retain the specific control over exactly when Winit flushes writes and polls file descriptors relative to exactly when it dispatches events. Calloop adds some extra abstraction around all of that which currently doesn't seem necessary within Winit, and for the one backend that currently uses Calloop it seems to also have issues as a result too.

kchibisov:
>Overall for mio (or maybe polling) vs calloop I think it's just a question of abstraction and control from my perspective. I would prefer to retain the specific control over exactly when Winit flushes writes and polls file descriptors relative to exactly when it dispatches events. Calloop adds some extra abstraction around all of that which currently doesn't seem necessary within Winit, and for the one backend that currently uses Calloop it seems to also have issues as a result too.

Sure, than what we should use is `polling` due to trust and control mentioned before, it also uses direct syscalls, so it's libc-less on linux which is nice. 

I'd list what we need to handle in Wayland backend:

1. eventfd/kqueue/pipe based wake up of the event loop (`std::mpsc::channel`).
2. timerfd/kqueue or epoll/kqueue sleep based timers for repeat rate which you can start/stop while dispatching other sources.
3. On demand fd insertion for non-blocking reading/writing for clipboard/dnd interactions

I think that's all actually, if we'd have polling based solution handling that I'd be fine.

rib:
> The mio is being replaced because we use calloop/polling and mio refused to have level/oneshot modes.

hehe, at the moment there's one backend using calloop and one using mio, and so it's surely equally valid to justify moving in the other direction.

I think the edge/level triggering stuff is a distraction - I don't see any reason why mio level triggering would be a problem with the wayland backend - if anything edge triggering looks well suited to how the wayland read protocol is design.

kchibisov:
>hehe, at the moment there's one backend using calloop and one using mio, and so it's surely equally valid to justify moving in the other direction.

calloop is using `polling` under the hood, and @notgull is `polling` maintainer.

rib:
> > hehe, at the moment there's one backend using calloop and one using mio, and so it's surely equally valid to justify moving in the other direction.
> 
> calloop is using `polling` under the hood, and @notgull is `polling` maintainer.

`polling` looks like a preferable level of abstraction within a winit backend to me. It's missing the waker api from mio which is convenient but a manual wake up mechanism wouldn't be too tricky to implement.

kchibisov:
>polling looks like a preferable level of abstraction within a winit backend to me. It's missing the waker api from mio which is convenient but a manual wake up mechanism wouldn't be too tricky to implement.

there's some work on that matter https://github.com/notgull/polling-utils

rib:
> > I think the edge/level triggering stuff is a distraction - I don't see any reason why mio level triggering would be a problem with the wayland backend - if anything edge triggering looks well suited to how the wayland read protocol is design.
> 
> I'm just saying that you have artificial limitations where there shouldn't be.

they do have technical reasons behind only supporting edge triggering for consistency across platforms though, and arguably edge triggering is the lower-level, common denominator choice.

In some cases it might be a bit of extra work to handle edge triggering but from what I've seen with the wayland backend edge triggering should actually be ok, without extra work.

rib:
> I'd list what we need to handle in Wayland backend:
> 
> 1. eventfd/kqueue/pipe based wake up of the event loop (`std::mpsc::channel`).
> 2. timerfd/kqueue or epoll/kqueue sleep based timers for repeat rate which you can start/stop while dispatching other sources.
> 3. On demand fd insertion for non-blocking reading/writing for clipboard/dnd interactions (this is not present in the current code, but it was used in draft PRs around handling that)

Currently it seems like the backend just needs to poll the wayland socket, plus you need a waker api that may use a pipe etc. I'm guessing the backend doesn't really need to use timerfd vs just passing a timeout to epoll. timerfd timeouts aren't going to be any more fancy / high precision than a timeout to epoll they are just a convenient way of turning a timeout into a file descriptor, which probably isn't really needed in the winit backend currently.

The stuff for the clipboard might change the polling requirements but that seems like it can be left as a separate thing to address that can be solved with `mio`, `polling` or `calloop`.

rib:
> > they do have technical reasons behind only supporting edge triggering for consistency across platforms though, and arguably edge triggering is the lower-level, common denominator choice.
> 
> It's true, but you can't deny that `level` is the default on `linux` and you have `oneshot` for convenience. It's just a matter of what is provided for you, `polling` provides `level`/`oneshot` on platforms supporting it.

Yeah but `mio` isn't just for Linux so it's reasonable that they constrain things to try and ensure you get an API that's as consistent as possible across all the platforms supported. As far as I recall there where fiddly/subtle details that where difficult to handle consistently across all platforms (presumably windows makes it difficult to support level triggering).

kchibisov:
I mean calloop solves all of 3 points now.

If you could make it work with `polling` without adding 1K LoC on top that would be nice. The main issue is that you have a delay timer and then repeat timer, but it's sort of manageable to do in some way...

kchibisov:
>Yeah but mio isn't just for Linux so it's reasonable that they constrain things to try and ensure you get an API that's as consistent as possible across all the platforms supported. As far as I recall there where fiddly/subtle details that where difficult to handle consistently across all platforms (presumably windows makes it difficult to support level triggering).

That's true, though, polling is using their own impl of windows event loop stuff. I'm just saying that winit already has platform specific code which won't go away, because display servers are different, mio sort of aims to be used without extra platform specific code. So having polling and being able what we want, because we use platform specific code anyway sounds more natural.

They both do sort of the same thing and it's more of a preference at this point (polling on windows has its own event loop impl not linking to any C-like things, like wepoll).

rib:
> > Yeah but mio isn't just for Linux so it's reasonable that they constrain things to try and ensure you get an API that's as consistent as possible across all the platforms supported. As far as I recall there where fiddly/subtle details that where difficult to handle consistently across all platforms (presumably windows makes it difficult to support level triggering).
> 
> That's true, though, polling is using their own impl of windows event loop stuff. I'm just saying that winit already has platform specific code which won't go away, because display servers are different, mio sort of aims to be used without extra platform specific code. So having polling and being able what we want, because we use platform specific code anyway sounds more natural.
> 
> They both do sort of the same thing and it's more of a preference at this point (polling on windows has its own event loop impl not linking to any C-like things, like wepoll).

Yeah, though this line of thinking is pretty much why I'd be more inclined to aim to use a lower level API - e.g. I don't think we'd use any of these on Windows currently where we need to handle windows messages, and these APIs also don't make sense on macos where we're using runloops.

If we're talking about one OS (Linux) then you basically just wants a safe API over `epoll`, the thinner the better really imho.

notgull:
I'm not sure what the advantages would be of creating a `winit`-internal async runtime. Unless there is a specific advantage it seems like it would be easier to just add whatever functionality is needed to pump events to `calloop`. I am opposed to taking on the technical debt that would be required to build the ad hoc, informally-specified, bug-ridden, slow implementation of half of an `async` runtime rather than just using one that already exists. If it's a matter of control, that sounds more like a `calloop` issue  rather than a `winit` one.

Furthermore, I would be opposed to using `polling-utils` in this crate. Not only is it incomplete for now, it is explicitly the wrong way of doing this kind of thing.

Also, irrelevant note: Windows doesn't actually support edge triggered mode. By default Windows only uses level triggered notifications by default, and `mio` uses a hack to get edge triggered polling to work. This is why `polling` doesn't support edge triggered polling on Windows.

kchibisov:
I'd not block this patch further, given that right now the move is to use calloop, future code could use what it'll be better for it.

rib:
> I'm not sure what the advantages would be of creating a winit-internal async runtime. 

My push back here was about:
1. why are we changing this?
3. please don't change this if there isn't really a strong benefit because it's going to create busy work for me to update my pump events branch.
4. there are technical issues with the use of calloop currently in the wayland backend currently, so out of the different options is it really the best choice?

I wasn't proposing anything related to an async runtime and calloop isn't really what I would refer to as an async runtime either, it's an event loop abstraction like glib, runloop or ALooper.

A cross-platform event loop integrated with the window system is pretty much the core thing that Winit provides so it seems reasonable that platform-specific implementation details would be maintained internal to Winit.

> Unless there is a specific advantage it seems like it would be easier to just add whatever functionality is needed to pump events to calloop

I wasn't saying that there is an incompatibility with my pump events changes (and the the pump events branch supports wayland based on calloop) but in the process of working on that branch I discovered that the wayland backend is over complicated (imho) and has to maintain a workaround for it to work with calloop (nothing to do with pump events).

I also wanted to help address the issue in the wayland backend but I found that calloop (and the fact that calloop was being switched under my feet to a new version at the time) made it more complicated to investigate properly and so I just left the existing workaround in place while it wasn't a high priority for me.

For that reason I had some bias against utilizing calloop in another backend since it apparently isn't working well so far.

>  I am opposed to taking on the technical debt that would be required to build the ad hoc, informally-specified, bug-ridden, slow implementation of half of an async runtime rather than just using one that already exists

I don't think this was being proposed.

rib:
> I could provide implementation for Wayland/X11 on your branch, that's not an issue. As well as sorting out issues arising from calloop along the way. I'm mostly interested in your proposed top-level API and macOS/Windows/Android implementations.

Thanks. I already implemented the backend changes for wayland (the implementation was basically complete, but missing docs)

[Re]-implementing the changes for the x11 backend probably won't be difficult either but I still think it was fair of me to question the need for the change to the x11 backend. As well as pointing out that the use of calloop in the wayland backend does currently have issues.

> 
> Also, if you could provide a short update on what's missing and course of action on your PR that would be nice, so I can tune in and help pushing it forward.

Yeah I think it's mainly just a question of writing some API docs.

We are using a branch of winit at Embark that's based on the pump_events changes on X11, Windows, MacOS and Android and   so it's had a reasonable amount of testing.

There's still a bunch of related things that came up while discussing pump_events that would be good to follow up on to further improve cross-platform consistency and fixup some naming that no longer really makes sense. I had been meaning to create an EventLoop 3.0 umbrella issue to track these things but haven't had enough time recently.

> 
> > I also wanted to help address the issue in the wayland backend but I found that calloop (and the fact that calloop was being switched under my feet to a new version at the time) made it more complicated to investigate properly and so I just left the existing workaround in place while it wasn't a high priority for me.
> 
> I'll look into it as a part of my updates and interior changed to wayland backend inside the winit. I have access to most crates in the chain, so it shouldn't take that much time to sort out the things. I also have example crates doing the multithread access, and so on.

sounds good. last time I was looking I think I convinced myself that the wayland source was fixable without needing an api break for calloop but ran out of energy to really try fixing it.

> 
> > or that reason I had some bias against utilizing calloop in another backend since it apparently isn't working well so far.
> 
> The X11 backend doesn't require all the things used by Wayland backend though, if you look on the implementation for X11, it mostly observes the changes and calls dispatchers(event sources) directly, so no special calloop logic being used.

Yep, tbh though it makes me more confused about why we would be using calloop here, because atm calloop feels like its kind of being misused conceptually in the x11 backend. `dispatch` is essentially being called just to `poll` and not really dispatch x11 events and then after the noop-calloop dispatch the backend does the real dispatching.

The "right" way to use calloop would probably be to create an X11Source that's more like the wayland source which would read x11 events and dispatch callbacks, but unlike the wayland backend there should be no need to have internal buffering of events and the calloop dispatching should directly call the winit event loop closure.

I sort of prefer the current misuse of calloop though - it's effectively just being used as an elaborate poll function.

PR #2778, opened 2023-04-26
Don't change the internal canvas size

### LINKED ISSUE #2733: Web: `transferControlToOffscreen()` breaks calls to canvas
I am using [`transferControlToOffscreen()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/transferControlToOffscreen) to draw in a different thread, which unfortunately invalidates any calls to [`width()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/width) and [`height()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/height).

This was not really an issue until now because I can simply avoid calling [`Window::set_inner_size()`](https://docs.rs/winit/0.28.2/wasm32-unknown-unknown/winit/window/struct.Window.html#method.set_inner_size). But later I discovered that switching to fullscreen and changing the scale factor will trigger a call to `width()` and `height()` by Winit I can't prevent.

I have two proposals to fix this:
- Detect the failure when calling [`HtmlCanvasElement::set_width()`](https://docs.rs/web-sys/0.3.61/web_sys/struct.HtmlCanvasElement.html#method.set_width) and just do nothing in that case. This is simple, but would require introducing custom `wasm-bindgen` imports because `set_width()` is not marked with `#[catch]`. This can't be fixed upstream because it's a breaking change.
- Introduce a new method to [`WindowExtWebSys`](https://docs.rs/winit/0.28.2/wasm32-unknown-unknown/winit/platform/web/trait.WindowExtWebSys.html) to actually call the `transferControlToOffscreen()` for the user, storing a flag somewhere that the canvas is now disconnected.

Happy to make a PR of course.

daxpedda:
See https://github.com/rust-windowing/winit/issues/2733, https://github.com/gfx-rs/wgpu/issues/3620 and https://github.com/gfx-rs/wgpu/pull/3690 for more information.

This is a follow-up to https://github.com/gfx-rs/wgpu/pull/3690. Now renderer are responsible for changing the internal canvas size. Winit will still continue to adjust the **visual** size of the canvas.

Both `wgpu` and `softbuffer` handle this correctly. Happy to check and make PR's to other crates that come to mind.

Fixes #2733.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - Bump MSRV from `1.60` to `1.64`.
 - On macOS, fixed potential panic when getting refresh rate.
 - On macOS, fix crash when calling `Window::set_ime_position` from another thread.
+- On Web, the canvas output bitmap size is no longer adjusted.
 
 # 0.28.3
 
--- a/src/platform_impl/web/web_sys/mod.rs
+++ b/src/platform_impl/web/web_sys/mod.rs
@@ -83,12 +83,7 @@ pub fn scale_factor() -> f64 {
 
 pub fn set_canvas_size(raw: &HtmlCanvasElement, size: Size) {
     let scale_factor = scale_factor();
-
-    let physical_size = size.to_physical::<u32>(scale_factor);
     let logical_size = size.to_logical::<f64>(scale_factor);
-
-    raw.set_width(physical_size.width);
-    raw.set_height(physical_size.height);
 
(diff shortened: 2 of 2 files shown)


kchibisov:
Would suggest to check glutin as well.

daxpedda:
I don't think `glutin` supports Wasm. Couldn't find any reference to anything in the code.
https://github.com/rust-windowing/glutin/issues/1167

kchibisov:
Ah, sorry, I for whatever reason thought that it was about macOS.

PR #2784, opened 2023-05-03
On macOS, fix empty marked text blocking input

### LINKED ISSUE #2775: Empty marked text bugs IME committing (can't type anything)
There is a bug in the latest Alacritty release for MacOS (https://github.com/alacritty/alacritty/issues/6817) where the keyboard input seems to be ignored.

It looks like a regression was introduced with this pull request https://github.com/rust-windowing/winit/pull/2666

When `set_marked_text` is called with an empty string, this `self.hasMarkedText()` condition prevents the IME state from changing to committed.

https://github.com/rust-windowing/winit/blob/249609889029af3e5ec19afbd02464e11a265378/src/platform_impl/macos/view.rs#L431-L436

And that's precisely what happens when I'm using [Fig](https://fig.io/).

When focusing on the terminal window, it triggers `set_marked_text` twice, first with a space `" "` and then with an empty string `""`. After that, I can't type anything.

Screenshot 2023-04-21 at 18 56 57&nbsp;&nbsp;&nbsp;Screenshot 2023-04-21 at 18 56 18

kchibisov:
Fixes #2775.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- On macOS, fix empty marked text blocking regular input.
 - On Wayland, fix maximized startup not taking full size on GNOME.
 - On Wayland, fix initial window size not restored for maximized/fullscreened on startup window.
 - On Wayland, `Window::outer_size` now accounts for **client side** decorations.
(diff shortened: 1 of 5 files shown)


kchibisov:
The patch was tested by https://github.com/alacritty/alacritty/issues/6817#issuecomment-1531887614

PR #2785, opened 2023-05-03
Winit version 0.28.4

kchibisov:


kchibisov:
The MSRV was tested locally by pinning the `raw-window-handle`. The wasm clippy failury is irrelevant, because it's in a top-level code due to internals of wasm. Fixing it breaks other code, and non-automative patches are out of point releases.

PR #2789, opened 2023-05-05
[FOLLOW UP] Keyboard Input 2.0 

kchibisov:
Given that the progress is sort of stalled on https://github.com/rust-windowing/winit/pull/2662 I'm opening this branch to continue the work without messing up with the original branch.

Fixes #1932.
Fixes #1890.
Fixes #1888.


src/platform_impl/web/web_sys/event.rs
```diff
@@ -61,190 +62,71 @@ pub fn mouse_scroll_delta(event: &WheelEvent) -> Option<MouseScrollDelta> {
...
-    m.set(ModifiersState::ALT, event.alt_key());
-    m.set(ModifiersState::LOGO, event.meta_key());
-    m
+pub fn key_location(event: &KeyboardEvent) -> KeyLocation {
+    match event.location() {
+        KeyboardEvent::DOM_KEY_LOCATION_LEFT => KeyLocation::Left,
+        KeyboardEvent::DOM_KEY_LOCATION_RIGHT => KeyLocation::Right,
+        KeyboardEvent::DOM_KEY_LOCATION_NUMPAD => KeyLocation::Numpad,
+        KeyboardEvent::DOM_KEY_LOCATION_STANDARD => KeyLocation::Standard,
+        _ => KeyLocation::Standard,
```
daxpedda, line 116: Continuation of https://github.com/rust-windowing/winit/pull/2662#discussion_r1186274430.
> We don't have unknown, and I don't think we should have it.

This will only be triggered if the specification changes and a new location is introduced, so it's kinda wrong what we are returning here, but I agree.

> We might have `unreachable!()`, but I'm afraid to do so for web, so let's keep the `Standard`, it should be more or less safe.

`unreachable!()` would work right now, but the Web changes a lot.

---

We could at least print an error or warning message with `log`? Does that sound appropriate?

kchibisov: I think that should be fine. We might add `Unknown` for sure.

ArturKovacs: I don't think we can reasonably add `Unknown`. The issue is that even the specification kinda defines `Standard` as "all the rest"
> DOM_KEY_LOCATION_STANDARD
>    The key activation MUST NOT be distinguished as the left or right version of the key, and (other than the NumLock key) did not originate from the numeric keypad (or did not originate with a virtual key corresponding to the numeric keypad).

I don't think we could distinguish between `Standard` and `Unknown` on the desktop implementations (Win/X11/Wayland/macOS).

daxpedda: According to the specification there isn't any other key location, so `unreachable!()` is the *correct* thing to do here. 

But if a new key location is added later, which is not out of the question, it would fail, but assuming `DOM_KEY_LOCATION_STANDARD` in that case is definitely **wrong**. So `KeyboardEvent::Unknown` makes sense here.

I'm currently in favor of **not** adding a `KeyboardEvent::Unknown` and instead just returning `KeyLocation::Standard` and use `log::error!()`. There is no need to burden the other backends only because *it's possible* that another key location is added.
> I don't think we could distinguish between `Standard` and `Unknown` on the desktop implementations (Win/X11/Wayland/macOS).

Another nail in the coffin of `KeyboardEvent::Unknown`.

That said it 's unlikely that a new key location will be added, the only one that was being considered is a gamepad, which was agreed to be moved to a dedicated game pad API: https://github.com/w3c/uievents/issues/79. Which has already happened: https://caniuse.com/gamepad.

---

**Conclusion**
We don't need to worry about this.

daxpedda (approved):
Web is LGTM.

When comparing the output of Web and X11, I noticed that using a dead key on X11 gives this:
`physical_key: Backquote, logical_key: Unidentified(Xkb(0xFE52))`
but on Web it gives this:
`physical_key: Backquote, logical_key: Dead(None)`

Don't know anything about X11, but maybe `Xkb(0xFE52)` should be converted to `Key::Dead`?

src/platform_impl/web/web_sys/event.rs
```diff
@@ -61,190 +62,71 @@ pub fn mouse_scroll_delta(event: &WheelEvent) -> Option<MouseScrollDelta> {
...
+    static STRING_CACHE: Lazy<Mutex<HashSet<&'static str>>> =
+        Lazy::new(|| Mutex::new(HashSet::new()));
+
+    let string = string.as_ref();
+    let mut cache = STRING_CACHE.lock().unwrap();
+    if let Some(string) = cache.get(string) {
+        string
+    } else {
+        // borrowck couldn't quite figure out this one on its own
+        let string: &'static str = Box::leak(String::from(string).into_boxed_str());
```
daxpedda, line 104:
```suggestion
        let string: &'static str = Box::leak(Box::from(string.as_ref()));
```
https://doc.rust-lang.org/std/boxed/struct.Box.html#method.from-9
Just leaving this here until we figure out something else.

kchibisov:
no need in this thing.

PR #2790, opened 2023-05-05
Use `cursor-icon` crate for `CursorIcon`

kchibisov:
This crate is aimed to simplify handling of cursor icon across various crates and be used in the public API.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,12 +44,14 @@ wayland-csd-adwaita-crossfont = ["sctk-adwaita", "sctk-adwaita/crossfont"]
 wayland-csd-adwaita-notitle = ["sctk-adwaita"]
 android-native-activity = ["android-activity/native-activity"]
 android-game-activity = ["android-activity/game-activity"]
+serde = ["dep:serde", "cursor-icon/serde"]
 
 [build-dependencies]
 cfg_aliases = "0.1.1"
 
 [dependencies]
 bitflags = "1"
+cursor-icon = "1.0.0"
 instant = { version = "0.1", features = ["wasm-bindgen"] }
 log = "0.4"
 mint = { version = "0.5.6", optional = true }
--- a/src/platform_impl/linux/x11/util/cursor.rs
+++ b/src/platform_impl/linux/x11/util/cursor.rs
@@ -1,3 +1,5 @@
+use std::ffi::CString;
+
 use crate::window::CursorIcon;
 
 use super::*;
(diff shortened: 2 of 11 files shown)


kchibisov:
r @notgull?

src/window.rs
```diff
@@ -15,6 +15,8 @@ use crate::{
 
 pub use crate::icon::{BadIcon, Icon};
 
+pub type CursorIcon = cursor_icon::CursorIcon;
```
notgull, line 18: This type alias should be documented.

Actually, it might be better to just inline the whole `cursor_icon` crate into the `window` module, like so:

```rust
#[doc(inline)]
pub use cursor_icon::{CursorIcon, ParseError as CursorIconParseError};
```

src/platform_impl/linux/x11/util/cursor.rs
```diff
@@ -45,78 +47,15 @@ impl XConnection {
...
-            CursorIcon::SwResize => load(b"bottom_left_corner\0"),
-            CursorIcon::WResize => load(b"left_side\0"),
-            CursorIcon::EwResize => load(b"h_double_arrow\0"),
-            CursorIcon::NsResize => load(b"v_double_arrow\0"),
-            CursorIcon::NwseResize => loadn(&[b"bd_double_arrow\0", b"size_fdiag\0"]),
-            CursorIcon::NeswResize => loadn(&[b"fd_double_arrow\0", b"size_bdiag\0"]),
-            CursorIcon::ColResize => loadn(&[b"split_h\0", b"h_double_arrow\0"]),
-            CursorIcon::RowResize => loadn(&[b"split_v\0", b"v_double_arrow\0"]),
-
-            CursorIcon::Text => loadn(&[b"text\0", b"xterm\0"]),
```
notgull, line 113: Was this "fallback" mechanism here important? It might be good to have a way to load the backup "xterm" cursor in the event that we fail to load the "text" cursor.

Cargo.toml
```diff
@@ -54,6 +55,7 @@ instant = { version = "0.1", features = ["wasm-bindgen"] }
 log = "0.4"
 mint = { version = "0.5.6", optional = true }
 once_cell = "1.12"
+cursor-icon = "1.0.0"
```
notgull, line 58: Nitpick: This dependency should be moved below `bitflags` but above `instant` so that the dependencies are alphabetized.

notgull (changes requested):

src/platform_impl/linux/x11/util/cursor.rs
kchibisov: I don't really care https://www.freedesktop.org/wiki/Specifications/cursor-spec/ . All modern themes have everything correct. The patch is also trivial and could be solved by renaming icon.

msiglreith (approved):

Issue #33, opened 2016-11-03
Change the way you get a window


tomaka:
Original: https://github.com/tomaka/glutin/issues/175

seivan:
I agree with this, because this is a bit of an issue for iOS. 
I got an issue where I need a game engine that uses Winit to work with the rest of the platform (macOS/iOS). The game engine in question uses Winit to set everything up (which btw is an amazing feature on its own). The problem is this won't work if you want Winit to coexist with existing systems. 

An approach would be to let Winit remain is it is and set everything up for you - but also offer a way to give the lowest common denominator `window` for each Platform. 

For iOS, this could be a `UIWindow`, or a `UIViewController` or even just a `UIView`. Right now, it's a bit overbearing(?) for iOS where Winit sets up a whole application (UIApplicationMain and UIApplicatonDelegate) this work perfect if the goal is that whatever project is intended to remain pure rust and not interact with the rest of the platform. 

My issue here, if the game engine works that way, then it's going to be tricky to integrate with the rest of the platform. If you need to present other views or deal with In App Purchase transactions then the current Winit setup doesn't work.

I am not asking to change the current approach, because there's a lot of value in that. 
What I am suggesting is an approach where Winit (and glutin) can work with creating components (say a UIView) for iOS.

seivan:
I did a small proof of concept of getting Winit to work as a `UViewController` that I later present in an Swift application on iOS and takes touch events.  Removed a lot of code from `iOS/mod.rs` for brevity. 
```rust

pub struct Window {
    _events_queue: Arc<RefCell<VecDeque<Event>>>,
    delegate_state: Box<DelegateState>,
}

unsafe impl Send for Window {}
unsafe impl Sync for Window {}

#[derive(Debug)]
struct DelegateState {
    window: id,
    size: LogicalSize,
    scale: f64,
}

impl DelegateState {
    fn new(window: id, size: LogicalSize, scale: f64) -> DelegateState {
        DelegateState {
            window,
            size,
            scale,
        }
    }
}



pub struct EventsLoop {
    events_queue: Arc<RefCell<VecDeque<Event>>>,
}

#[derive(Clone)]
pub struct EventsLoopProxy;

impl EventsLoop {
    pub fn new() -> EventsLoop {
        // unsafe {
        //     if !msg_send![class!(NSThread), isMainThread] {
        //         panic!("`EventsLoop` can only be created on the main thread on iOS");
        //     }
        // }
        EventsLoop {
[...]

PR #2795, opened 2023-05-07
Initial kmsdrm support [forked + updated]

mraerino:
I forked https://github.com/rust-windowing/winit/pull/2272 and updated it to resolve merge conflicts and so it compiles with the current dependencies.

---


.github/workflows/ci.yml
```diff
@@ -16,7 +16,7 @@ jobs:
         components: rustfmt
     - name: Check Formatting
       run: cargo +stable fmt --all -- --check
-  
+
```
notgull, line 19: Why?

Cargo.toml
```diff
@@ -44,6 +44,8 @@ wayland-csd-adwaita-crossfont = ["sctk-adwaita", "sctk-adwaita/crossfont"]
 wayland-csd-adwaita-notitle = ["sctk-adwaita"]
 android-native-activity = ["android-activity/native-activity"]
 android-game-activity = ["android-activity/game-activity"]
+kms = ["drm", "input", "xkbcommon", "udev", "parking_lot"]
+kms-ext = ["libseat"]
```
notgull, line 48: Is it still necessary for split the feature into `kms` and `kms-ext` like this? I forget what the original rationale for this was.

Cargo.toml
```diff
@@ -115,14 +117,20 @@ wayland-backend = { version = "0.1.0", default_features = false, features = ["cl
 wayland-protocols = { version = "0.30.0", features = [ "staging"], optional = true }
 calloop = "0.10.5"
 x11-dl = { version = "2.18.5", optional = true }
+drm = { version = "0.6.2", optional = true }
+input = { version = "0.7.1", optional = true }
+libseat = { version = "0.1.4", optional = true }
+udev = { version = "0.6.3", optional = true }
+xkbcommon = { version = "0.5.0", optional = true }
```
notgull, line 124: It might be worth it to just use `xkbcommon-dl` instead, since that's what #2789 is using.

src/platform_impl/linux/kms/event_loop.rs
```diff
@@ -0,0 +1,701 @@
...
+    /// Event loop handle.
+    pub event_loop_handle: calloop::LoopHandle<'static, EventSink>,
+
+    pub(crate) event_sink: EventSink,
+
+    /// A proxy to wake up event loop.
+    pub event_loop_awakener: calloop::ping::Ping,
+
+    _marker: std::marker::PhantomData<T>,
+}
```
notgull, line 64: It would be nice if all of the fields were either `pub` or `pub(crate)`, instead of this mix.

src/platform_impl/linux/kms/event_loop.rs
```diff
@@ -0,0 +1,701 @@
...
+    *p_better_planes.get(0).unwrap_or(&p_compatible_planes[0])
+}
+
+fn find_card_path(seat_name: &str) -> Result<PathBuf, error::OsError> {
+    let mut enumerator = Enumerator::new().map_err(|e| {
+        os_error!(OsError::KmsError(format!(
+            "failed to open udev enumerator: {}",
+            e
+        )))
+    })?;
```
notgull, line 155: Might be nice to use a combinator [like we have in `softbuffer`](https://github.com/rust-windowing/softbuffer/blob/ecfae194cebf97ca23e450e4f4541d911fa132a0/src/error.rs#L43-L60) to avoid duplicating this `map_err` block`.

src/platform_impl/linux/kms/event_loop.rs
```diff
@@ -0,0 +1,701 @@
...
+
+    enumerator
+        .scan_devices()
+        .map_err(|e| os_error!(OsError::KmsError(format!("failed to scan devices: {}", e))))?
+        .filter(|device| {
+            let dev_seat_name = device
+                .property_value("ID_SEAT")
+                .map(|x| x.to_os_string())
+                .unwrap_or_else(|| std::ffi::OsString::from("seat0"));
```
notgull, line 178: I would avoid using the fully qualified path here.

src/platform_impl/linux/kms/event_loop.rs
```diff
@@ -0,0 +1,701 @@
...
+            if dev_seat_name == seat_name {
+                if let Ok(Some(pci)) = device.parent_with_subsystem(Path::new("pci")) {
+                    if let Some(id) = pci.attribute_value("boot_vga") {
+                        return id == "1";
+                    }
+                }
+            }
+            false
+        })
+        .flat_map(|device| device.devnode().map(std::path::PathBuf::from))
```
notgull, line 188: I would avoid using the fully qualified path here.

src/platform_impl/linux/kms/event_loop.rs
```diff
@@ -0,0 +1,701 @@
...
+                .unwrap_or_else(|| {
+                    std::env::var_os("LC_CTYPE").unwrap_or_else(|| {
+                        std::env::var_os("LANG").unwrap_or_else(|| std::ffi::OsString::from("C"))
+                    })
+                })
+                .as_os_str(),
+            xkb::compose::COMPILE_NO_FLAGS,
+        )
+        .map_err(|_| {
+            // e ^^^ would return ()
```
notgull, line 341: In that case we can just do:

```rust
.map_err(|()| { ... })
```

src/platform_impl/linux/kms/event_loop.rs
```diff
@@ -0,0 +1,701 @@
...
+        let compose_table = xkb::compose::Table::new_from_locale(
+            &xkb_ctx,
+            // These env variables in Linux are the most likely to contain your locale,
+            // "en_US.UTF-8" for example
+            std::env::var_os("LC_ALL")
+                .unwrap_or_else(|| {
+                    std::env::var_os("LC_CTYPE").unwrap_or_else(|| {
+                        std::env::var_os("LANG").unwrap_or_else(|| std::ffi::OsString::from("C"))
+                    })
+                })
```
notgull, line 336: I would avoid using the fully qualified path here.

src/platform_impl/linux/kms/event_loop.rs
```diff
@@ -0,0 +1,701 @@
...
+
+    pub fn run<F>(mut self, callback: F) -> !
+    where
+        F: FnMut(Event<'_, T>, &event_loop::EventLoopWindowTarget<T>, &mut ControlFlow) + 'static,
+    {
+        let exit_code = self.run_return(callback);
+        std::process::exit(exit_code);
+    }
+
+    pub fn run_return<F>(&mut self, mut callback: F) -> i32
```
notgull, line 499: Since this uses `calloop`, Wayland uses `calloop`, and the X11 backend will eventually use `calloop` as well (#2777), it might be nice to extract some of the common code here into a separate module. Not now, but eventually.

src/platform_impl/linux/kms/event_loop.rs
```diff
@@ -0,0 +1,701 @@
...
+            );
+
+            // Send RedrawEventCleared.
+            sticky_exit_callback(
+                Event::RedrawEventsCleared,
+                &self.window_target,
+                &mut control_flow,
+                &mut callback,
+            );
+        };
```
notgull, line 627: Aren't there supposed to be redraw events handled between `MainEventsCleared` and `RedrawEventsCleared`?

src/platform_impl/linux/kms/event_loop.rs
```diff
@@ -0,0 +1,701 @@
...
+            // we're doing callback to the user, since we can double borrow if the user decides
+            // to create a window in one of those callbacks.
+            self.with_window_target(|window_target| {
+                let state = &mut window_target.event_sink;
+                std::mem::swap::<Vec<Event<'static, ()>>>(&mut event_sink_back_buffer, state);
+            });
+
+            // Handle pending window events.
+            for event in event_sink_back_buffer.drain(..) {
+                let event = event.map_nonuser_event().unwrap();
```
notgull, line 608: Why do we need to map out the user events here?

src/platform_impl/linux/kms/input.rs
```diff
@@ -0,0 +1,716 @@
...
+                    input::Event::Pointer(ev) => handle_pointer_event!(self, ev, callback),
+                    input::Event::Keyboard(ev) => handle_keyboard_event!(self, ev, callback),
+                    _ => {}
+                }
+            }
+        }
+        Ok(PostAction::Continue)
+    }
+
+    fn register(&mut self, poll: &mut Poll, factory: &mut TokenFactory) -> calloop::Result<()> {
```
notgull, line 700: Maybe instead of using `poll` directly, we can wrap the `libinput` handling in the [`calloop::Generic`](https://docs.rs/calloop/latest/calloop/generic/struct.Generic.html) type and use that to handle things?

src/platform_impl/linux/kms/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+/// A simple wrapper for a device node.
+pub struct Card(pub(crate) Arc<i32>);
+
+/// Implementing `AsRawFd` is a prerequisite to implementing the traits found
+/// in this crate. Here, we are just calling `as_raw_fd()` on the inner File.
+impl unix::io::AsRawFd for Card {
+    fn as_raw_fd(&self) -> unix::io::RawFd {
+        *self.0
+    }
+}
```
notgull, line 33: Modern versions of the `drm` crate use `AsFd`, maybe we should use that instead?

src/platform_impl/linux/kms/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+
+impl Drop for Card {
+    fn drop(&mut self) {
+        unsafe { std::fs::File::from_raw_fd(*self.0) };
+    }
+}
```
notgull, line 39: This doesn't look sound to me. Since the `i32` can still be used outside of the `Card` (thanks to `AsRawFd`), couldn't it still be used after a free? It might be better to define `Card` as a wrapper around an `OwnedFd` and then wrap that in an `Arc`.

src/platform_impl/linux/kms/window.rs
```diff
@@ -0,0 +1,488 @@
...
+    // there is only one window
+    0.into()
+}
+
+impl Window {
+    pub fn new<T>(
+        event_loop_window_target: &super::event_loop::EventLoopWindowTarget<T>,
+        _attributes: WindowAttributes,
+        _platform_attributes: platform_impl::PlatformSpecificWindowBuilderAttributes,
+    ) -> Result<Self, error::OsError> {
```
notgull, line 76: Might be worth considering #33 here. Maybe, if a window has already been created, it's better to just return an error instead of returning another window?

src/platform_impl/linux/mod.rs
```diff
@@ -204,13 +226,15 @@ pub enum MonitorHandle {
 /// }
 /// ```
 /// The result can be converted to another enum by adding `; as AnotherEnum`
-macro_rules! x11_or_wayland {
+macro_rules! x11_or_wayland_or_drm {
```
notgull, line 229: Especially if we add more backends in the future, it might be better to change the name of this (maybe `platform_specific!`)?

notgull (changes requested):
Went through this PR. I know most of this wasn't you, but it should still be fixed before merging.

Since we have to wait until after #2789 merges, it might be useful to base on that for now?

src/platform_impl/linux/kms/mod.rs
ids1024: Yeah, this isn't right because `drop` is closing this every time you drop a `Card`, even if you cloned it. The dropping needs to be within the `Arc`.

Winit currently requires Rust 1.64.0, so it be correct and require less code to use `struct Card(Arc<OwnedFd>)`.

flukejones:
I don't mean to be a bother, but is there a timeframe for merging this? It's exactly what I'm looking for to be used with slint and in embedded environment (STM32MP157). I can test this later on.

Cargo.toml
flukejones: `kms-ext` uses (`libseat`)[https://sr.ht/~kennylevinsen/seatd/].

> Seat management takes care of mediating access to shared devices (graphics, input), without requiring the applications needing access to be root.

So this lib links/uses existing methods such as logind to create a user "seat" and allow access to various things. The lib is also a good intermediate step between winit and systemd-logind (plus others).

In our case this would allow anyone using this feature to run via TTY without requiring root. Standard `kms` here does require root access.

Unfortunately this feature is broken in this PR.

```
❯ cargo run --example window --no-default-features --features=kms-ext
   Compiling libseat-sys v0.1.7
   Compiling slog v2.7.0
   Compiling libseat v0.1.7
   Compiling winit v0.28.4 (/home/luke/Projects/winit-drm)
error: The platform you're compiling for is not supported by winit
  --> src/platform_impl/mod.rs:68:1
   |
68 | compile_error!("The platform you're compiling for is not supported by winit");
```

kchibisov:
>I don't mean to be a bother, but is there a timeframe for merging this? It's exactly what I'm looking for to be used with slint and in embedded environment (STM32MP157). I can test this later on.

No timeline for changes like that, you might look into smithay project instead if you want to. It's not only about wayland in the end of the day.

flukejones:
> > I don't mean to be a bother, but is there a timeframe for merging this? It's exactly what I'm looking for to be used with slint and in embedded environment (STM32MP157). I can test this later on.
> 
> No timeline for changes like that, you might look into smithay project instead if you want to. It's not only about wayland in the end of the day.

Yeah been looking at a few options. It would be great to get this in winit as it means a few projects would gain it for free. SDL2 also has a similar DRM method and I've used that for my room4doom (as example).

flukejones:
> No timeline for changes like that, you might look into smithay project instead if you want to. It's not only about wayland in the end of the day.

The reason I wondered about this is because if winit implements KMS backend, then many many other projects gain KMS ability for free. Especially nice for GUI toolkits in embedded applications.

daxpedda:
Closing due to inactivity.
See #1006.

PR #2802, opened 2023-05-13
Winit version 0.28.6

kchibisov:


kchibisov:
Tested msrv locally by downgrading packages.

PR #2806, opened 2023-05-14
Increase test coverage for generic modules

notgull:

This PR adds test cases for the higher-level modules of `winit` (dpi, event, etc). The goal here is to have 100% test coverage for `winit`, or at least as high as we can reasonably reach.

image

WIP


notgull:
Eh, it's about ready now. Baby steps.

src/error.rs
```diff
@@ -106,3 +106,25 @@ impl error::Error for OsError {}
...
+
+#[cfg(test)]
+mod tests {
+    #![allow(clippy::redundant_clone)]
+
+    use super::*;
+
+    // Eat attributes for testing
+    #[test]
+    fn attr_coverage() {
```
madsmtm, line 118: Maybe call this `ensure_fmt_does_not_panic` or similar?

Same for other files.

tests/event.rs
```diff
@@ -0,0 +1,280 @@
+use std::collections::{BTreeSet, HashSet};
```
madsmtm, line 1: In general, I would prefer this to live at the bottom of the module it's testing, but that's just personal preference.

madsmtm (approved):
I think the `clone` stuff may be a bit redundant after https://github.com/rust-windowing/winit/pull/2981?

But more tests = yay, so all approved from my side!

kchibisov:
Besides that it doesn't build, could we have tests in the file they are testing?

daxpedda (approved):
Tests LGTM as well.

Let me know if you want me to fix the Web build, didn't look into why it fails rn.

kchibisov (approved):

PR #2811, opened 2023-05-18
Add more documentation for keyboard types

notgull:

This PR adds documentation for some of the items introduced in #2662


src/event.rs
```diff
@@ -910,10 +910,39 @@ pub struct KeyEvent {
...
+    /// this event is the result of one of those repeats.
     pub repeat: bool,
 
+    /// Platform-specific key event information.
+    ///
+    /// On Windows, Linux and macOS, this type contains the key without modifiers and the text with all
+    /// modifiers applied.
+    ///
+    /// On Android, iOS, Redox and Web, this type is a no-op.
     pub(crate) platform_specific: platform_impl::KeyEventExtra,
```
maroider: I'm not sure if this part is necessarily the most useful since the field is `pub(crate)`, though I suppose it might save someone unfamiliar with this part of the code several minutes of browsing through the various backends.

src/keyboard.rs
```diff
@@ -1577,11 +1588,72 @@ impl Key {
...
+/// [`KeyEvent`]: crate::event::KeyEvent
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub enum KeyLocation {
+    /// The key is in its "normal" location on the keyboard.
+    ///
+    /// For instance, the "1" key above the "Q" key on a QWERTY keyboard will use this location. This
+    /// invariant is also returned when the location of the key cannot be identified.
+    ///
+    /// ![Standard 1 key](https://i.imgur.com/TGjUpSC.png)
```
maroider, line 1610: I'm not sure about using imgur as an image host, given how they've recently decided to [delete "old, unused, and inactive content that is not tied to a user account"](https://arstechnica.com/gadgets/2023/04/hosting-site-imgur-will-remove-explicit-and-anonymous-content-next-month/). I'd prefer to use a light-weight SVG hosted in the repo itself, so that our documentation stays more self-contained.

src/event.rs
notgull: This is primarily a response to [this comment](https://github.com/rust-windowing/winit/pull/2662/commits/a2c16e01ee585866e611c12c829fdafb4e69b4c7#r1106569962).

src/keyboard.rs
notgull: Even after running the photos through `svgo`, each photo is around 450 KB a piece, which would add up to about 2 MB. As a comparison, the total size of the source code is around 2.2 MB in and of itself. Therefore, adding these images to the repo itself would nearly double its size.

If you're fine with that, I'm still willing to do this.

src/event.rs
maroider: I guess that's alright, then.

src/keyboard.rs
maroider: I think we can do better than 450 KB if we use lighter SVG as a base, like [this one](https://commons.wikimedia.org/wiki/File:ANSI_US_QWERTY_(Windows).svg), which weights in at 80 KB. It might also be possible to further reduce the size by using SVG links, though I've never used them myself.

notgull: Yeah, that sounds good. I've changed out for that SVG and include it into the docs.

kchibisov:
Could you resolve conflicts?

src/keyboard.rs
```diff
@@ -0,0 +1,1667 @@
...
+/// [`KeyEvent`]: crate::event::KeyEvent
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum KeyLocation {
+    /// The key is in its "normal" location on the keyboard.
+    ///
+    /// For instance, the "1" key above the "Q" key on a QWERTY keyboard will use this location. This
+    /// invariant is also returned when the location of the key cannot be identified.
+    ///
+    /// ![Standard 1 key](https://raw.githubusercontent.com/rust-windowing/winit/master/images/keyboard_standard_1_key.svg)
```
maroider, line 1610: Since there's no neat solution for rust-lang/rust#32104, I guess linking to it through GitHub works fine for now. My gut tells me we should be linking to a tagged release rather than the `master` branch, but linking to `master` initially is fine. I did find a way to link to the local copy of the SVG first, and then fall back to the GitHub link (roughly ``), but that wouldn't link the local version of the image if Winit's docs are built as a dependency, though I suppose instances where that would matter are exceedingly rare, if nonexistent.

maroider:
While I'm sorely tempted to have one of us look into the SVG `use` element as a means of further reducing the SVG size, I think the size is OK now. This looks good otherwise.

Issue #2807, opened 2023-05-15
winit doesn't work with wayland or X11 on nixos unstable - wayland doesn't show any windows and on X11 window is blank


uzytkownik:
When I try to run `window_debug` or 'web` example with wayland on nixos unstable with gnome-shell no window is shown. X11 shows just black screen that doesn't refresh.

Not sure if this helps but on 2.27.5 the program panics on Wayland:

```
thread 'main' panicked at 'internal error: entered unreachable code', /tmp/winit/src/platform_impl/linux/wayland/window/mod.rs:297:77
```

kchibisov:
Winit examples don't draw anything into the window, so the behavior is expected.

The error you've encountered before is because you're missing `libxkbcommon`, which you can try to avoid by statically linking to it(feature in the config).

winit relies on `dlopen` which is not friendly to nix, so you should set `LD_LIBRARY_PATH` or `patchelf`.

notgull:
> winit relies on `dlopen` which is not friendly to nix, so you should set `LD_LIBRARY_PATH` or `patchelf`.

An option to avoid `dlopen` might be nice in the future

kchibisov:
you can have wayland without dlopen if you disable `wayland-dlopen` feature. Only x11 is lacking.

uzytkownik:
> The error you've encountered before is because you're missing `libxkbcommon`, which you can try to avoid by statically linking to it(feature in the config).
> 

Ok. I was expecting the error message about failing to load as with other. Library I was using 2.27.x so I got confused by errors.

> winit relies on `dlopen` which is not friendly to nix, so you should set `LD_LIBRARY_PATH` or `patchelf`.

I know. Here's my shell.nix:

```nix
{ pkgs ? import <nixpkgs> {} }:
pkgs.mkShell {
  packages = [ pkgs.gdb ];
  buildInputs = [ pkgs.stdenv pkgs.cmake pkgs.pkgconfig pkgs.libxkbcommon ];
  nativeBuildInputs = [ pkgs.fontconfig ];
  shellHook = ''
    export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:${pkgs.wayland}/lib:${pkgs.libxkbcommon}/lib
[...]

PR #2812, opened 2023-05-20
Fill the windows in the examples with a solid color

notgull:
- [ ] Tested on all platforms changed (only tested on X11)

When you run the examples for `winit`, the contents of the window will be undefined depending on the platform. To the uninformed user, this appears to be a bug in some cases. In the past we've gotten quite a few issues because of this (#2807 comes to mind) that aren't issues at all.

This PR fills the windows with a solid gray background in order to make sure that there is *something* inside of the window.

Before (X11):

image

After (X11):

image


kchibisov:
That's true, though there's always glutin with window example, which I use as a testing or alacritty branches which I always open to target latest winit dev, since that's what I use myself as a maintainer to test winit API.

notgull:
I've rebased on the latest master. So far `softbuffer` doesn't support Android (rust-windowing/softbuffer#44) or iOS (rust-windowing/softbuffer#43) yet, so I've just made `fill_window` a no-op on those platforms. The Windows CI errors are due to `winit` using `windows-sys` version 0.45 and should be fixed by #2842

kchibisov (changes requested):
All examples must handle `Resize` events, otherwise Wayland won't fly.

Probably macOS won't fly as well.

examples/util/fill.rs
```diff
@@ -0,0 +1,45 @@
...
+    GC.with(|gc| {
+        // Either get the last context used or create a new one.
+        let mut gc = gc.borrow_mut();
+        let context = gc.entry(window.id()).or_insert_with(|| unsafe {
+            GraphicsContext::new(window, window)
+                .expect("Failed to create a softbuffer graphics context")
+        });
+
+        // Fill a buffer with a solid color.
+        const LIGHT_GRAY: u32 = 0xFFD3D3D3;
```
kchibisov, line 33:
```suggestion
        const LIGHT_GRAY: u32 = 0xFF181818;
```

my eyes.

kchibisov:
>All examples must handle Resize events, otherwise Wayland won't fly.

>Probably macOS won't fly as well.


Actually, you do `Resize`, but the thing is that it's not really resized on Wayland, even when winit is telling the right sizes. I'd assume we'd need a new release of softbuffer?

cc @ids1024

ids1024:
At this point we might as well do another release of softbuffer, from the current master, instead of having things relying on 0.2. It's a breaking release, and it will probably have another breaking release for other things, but that's not really an issue at this stage.

kchibisov:
>At this point we might as well do another release of softbuffer, from the current master, instead of having things relying on 0.2. It's a breaking release, and it will probably have another breaking release for other things, but that's not really an issue at this stage.

That would be nice, given that it's un-usable in the current state on Wayland from what I've seen in this PR. The example on softbuffer master does work fine.

notgull:
Now `cargo-deny` is flagging the `memmap2` dependency. Shouldn't be a real issue.

kchibisov:
Could you add memmap2 into deny and open an issue on client-toolkit(I can do this myself if you don't want to)?

kchibisov (approved):

PR #2815, opened 2023-05-21
Fix window set_theme doc string

### LINKED ISSUE #2814: `set_theme` doc is poorly formatted
Looking at the source for `set_theme`, I see some code running on X11. I'm not sure what it does exactly, but presumably it does something [here is the code](https://docs.rs/winit/latest/src/winit/platform_impl/linux/x11/window.rs.html#572-595).

Yet the doc string for `Window::set_theme` says "x11: unsupported". Actually there is a line about x11, but it's broken. Expect a PR shortly.

nicopap:
Fix #2814


kchibisov:
Could you rebase?

src/window.rs
```diff
@@ -1129,9 +1129,9 @@ impl Window {
     ///
     /// - **macOS:** This is an app-wide setting.
     /// - **Wayland:** You can also use `WINIT_WAYLAND_CSD_THEME` env variable to set the theme.
-    /// Possible values for env variable are: "dark" and light". When unspecified, a theme is automatically selected.
-    /// -**x11:** Sets `_GTK_THEME_VARIANT` hint to `dark` or `light` and if `None` is used, it will default to  [`Theme::Dark`].
-    /// - **iOS / Android / Web / x11 / Orbital:** Unsupported.
+    ///   Possible values for env variable are: "dark" and light". When unspecified, a theme is automatically selected.
+    /// - **x11:** Sets `_GTK_THEME_VARIANT` hint to `dark` or `light` and if `None` is used, it will default to  [`Theme::Dark`].
```
kchibisov, line 1133:
```suggestion
    /// - **X11:** Sets `_GTK_THEME_VARIANT` hint to `dark` or `light` and if `None` is used, it will default to  [`Theme::Dark`].
```

Just noticed that it was lower case....

kchibisov (approved):

Issue #2768, opened 2023-04-12
Adjust the `ModifiersChanged` event handling


kchibisov:
The spec got one more "clarification" lately wrt the keyboard modifiers. Given that our `ModifiersChanged` event is tied to `Focus` state, as in on leave we don't have focus and on regaining focus we have modifiers back on.

For winit's event API it simply means that we should send the `ModifiersChanged` event along side the `PointerEnter` and `PointerLeave` events. I'm not sure whether it's a breaking change, but it's something we should consider once we land the #2662 (Keyboard input v2).

The backends should start sending this event when pointer enters the surface of the window, as well as sending it again on keyboard enter/leave.

The good thing is that we still assume the `empty` state when losing, but then compositor should send the modifiers event again to tell us about the pointer focus. Usually compositors will send
such events related to pointer and keyboard focuses, so we should have internal `seat` focus inside
the Wayland backend.

For more see:

https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/259

kchibisov:
Hm, hard to find compositor that actually does so to test. Will move to 0.31 for now.

PR #2817, opened 2023-05-24
New keyboard various fixes

kchibisov:
The main highlight is modifiers event rework.


src/keyboard.rs
```diff
@@ -71,55 +71,179 @@
 
 use smol_str::SmolStr;
 
+/// Represents the current state of the keyboard modifiers
+///
+/// Each flag represents a modifier and is set if this modifier is active.
+#[derive(Default, Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct ModifiersState {
+    pub(crate) flags: ModifiersStateFlags,
```
kchibisov, line 79: I'm not sure how `PartialEq` and `Eq` should work here. Internally in winit we want to compare by each field, but users might not want such a thing.

ArturKovacs:
I don't use macOS regularly so every time this Option/Alt question comes up, I have to do a small research to remember what this is about.

The way I understand it is the following:

With some keyboard layouts, holding down the Option key and pressing another key allows users to produce special characters. However, some applications have keyboard shortcuts that are bound to Option+another key. With the previous keyboard API, winit only reported the special character and not the original key, which prevented the shortcut from being triggered. This is why we needed `option_as_alt`, which prevented the option modifier from being applied onto the received character. With the new API winit reports both, the special character (`logical_key`) and the original character (`key_without_modifiers`), and therefore there's no need for `option_as_alt`.

Is this correct?

kchibisov:
I think I don't like the current impl of left/right modifiers and they way the struct is ending up used downstream. I'll rewrite it to be as a separate field instead.

kchibisov:
Rewrote the way the `left`/`right` modifiers are passed, should be good now.

kchibisov:
I've also removed the deprecated `modifiers` field.

@daxpedda could you ensure that I haven't broke the web by doing so?

src/keyboard.rs
```diff
@@ -1664,3 +1567,209 @@ pub enum KeyLocation {
...
+    }
+
+    /// The state of the right super key.
+    pub fn rsuper_state(&self) -> ModifiersKeyState {
+        if self.pressed_mods.contains(ModifiersKeys::RCONTROL) {
+            ModifiersKeyState::Pressed
+        } else {
+            ModifiersKeyState::Unknown
+        }
+    }
```
madsmtm, line 1750:
```suggestion
    /// The state of the left super key.
    pub fn lsuper_state(&self) -> ModifiersKeyState {
        if self.pressed_mods.contains(ModifiersKeys::LSUPER) {
            ModifiersKeyState::Pressed
        } else {
            ModifiersKeyState::Unknown
        }
    }

    /// The state of the right super key.
    pub fn rsuper_state(&self) -> ModifiersKeyState {
        if self.pressed_mods.contains(ModifiersKeys::RSUPER) {
            ModifiersKeyState::Pressed
        } else {
            ModifiersKeyState::Unknown
        }
    }
```

src/keyboard.rs
```diff
@@ -1664,3 +1567,209 @@ pub enum KeyLocation {
...
+            m.set(ModifiersState::ALT, alt_key);
+            m.set(ModifiersState::SUPER, super_key);
+            Ok(m)
+        }
+    }
+}
+
+/// Extra information on the current [`ModifiersState`].
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct ModifiersMetadata {
```
madsmtm, line 1670: API-wise I'd prefer:
```rust
#[non_exhaustive]
pub struct Modifiers {
    pub state: ModifiersState,
    pub(crate) pressed_mods: ModifiersKeys,
}

// Along with:
// WindowEvent::ModifiersChanged(Modifiers)
```

And in the future, even getting rid of the `pub` on `state`, in favour of explicit methods on `Modifiers`. But just a preference, the current API works fine.

madsmtm (approved):
MacOS impl looks good, nice to have the `OptionAsAlt` removed, better that it can now be implemented outside of `winit`.

src/keyboard.rs
kchibisov: Nah, I had something like that in fist revision and it's just bad. You simply want to write stuff like `Modifiers::SUPER | Modifiers::SHIFT` and you want to have all the bitflags semantics.

You also want to have `eq` based on just `state` not the `Modifiers` in particular, given that pressed_mods/metadata doesn't provide any end value to the users.

So, the main issue is that you really want all of that `bitflags` semantics, the matadata is just optional field to check on some platforms that you have something pressed. On most it's not possible to tell which key resulted in modifiers change.

kchibisov: Though, maybe I got you wrong and you simply want to have a distinct structure for the event.

Not sure if it'll be any better though, I think it'll be around the same, and `#[non_exhaustive]` is redundant due to presence of metadata.

daxpedda (approved):
Web is fine, though I found that if you hold down both left and right of any modifier key and you release only one, it will report that no modifier keys are active, even though you are holding one down. Though this isn't an issue with this PR, it's already one with https://github.com/rust-windowing/winit/pull/2662, I will fix it there.

So this PR LGTM!

src/keyboard.rs
madsmtm: I just wanted a distinct struct for the event, but again, mostly preference

kchibisov: Updated to that.

daxpedda:
> @daxpedda have you tested the case when you hold the modifier before focusing the browser and simply move the mouse over it?

Discussed on Matrix, not applicable here. See https://github.com/rust-windowing/winit/issues/2768.

kchibisov:
>Discussed on Matrix, not applicable here. See #2768.

I think it was re-iterated and your latest commit addresses all of that.

daxpedda:
# Modifiers on Web

## Changes

- When losing focus, we report no active modifiers with `ModifiersChanged`.
- When regaining focus, we can't determine which modifiers are currently active, but we can deduce it from keyboard and mouse events. If focusing happens to coincide with any of those events, it will show up exactly how it does on native.
- Store `has_focus` in `Cell` instead of `RefCell`, it's just a `bool`.

## Failed Solution 1

Why we don't "track" modifiers over keyboard and mouse events while the window isn't in focus?
The problem starts with some terminology inconsistencies, a Winit window is usually what a OS describes as a window. That is not the case for Web, on Web a window is just a canvas. We could track mouse events and even keyboard events while the canvas isn't in focus, but not when the actual browser window is not in focus. If we track modifiers while the canvas isn't in focus but the browser window is, modifiers could change while the browser window isn't in focus and we would get stuck modifiers.

## Failed Solution 2

Trying to resolve the solution outlined above, we could track modifiers while the canvas isn't in focus but the browser window is, but clear modifiers if the browser window loses focus to avoid getting modifiers stuck when the window loses focus. Unfortunately this doesn't really provide any advantage, because the only way a canvas can regain focus without having lost the browser window focus is with a keyboard or mouse event (unless you do it over scripting, which is becoming increasingly insane).

## Conclusion

I wish I could explain any of this better, but this is all very convoluted and the Web APIs are quite complicated.

The conclusion is that there is really no perfect way to have a correct modifier state when regaining the browser window focus, we will have to wait until we get a keyboard or mouse event.

daxpedda:
Leaving https://github.com/rust-windowing/winit/pull/2817#issuecomment-1564731618 here for future reference.

maroider (commented):
Regarding the left/right modifiers reporting, I personally think the commented out portions of `ModifiersState` can be used in a sensible manner to support left/right modifiers (i.e. `SHIFT = LSHIFT | RSHIFT`, but `SHIFT` can be set regardless of whether we support distinguishing between left and right). The way you've implemented is also serviceable, but I'd like to have a rationale for why this way would be preferred.

---

The wasm changes appear to work as intended, as far as I can tell, so that's nice.

---

Other than that, I only have a couple of nits.

kchibisov:
>Regarding the left/right modifiers reporting, I personally think the commented out portions of ModifiersState can be used in a sensible manner to support left/right modifiers (i.e. SHIFT = LSHIFT | RSHIFT, but SHIFT can be set regardless of whether we support distinguishing between left and right). The way you've implemented is also serviceable, but I'd like to have a rationale for why this way would be preferred.

On some platforms you could have modifier change without any key press at all, so left/right can't be determined by any means. Only macOS supports such a thing and only macOS requires this API anyway, because they have broken modifiers input (Option and not AltGr).

src/event.rs
```diff
@@ -966,6 +937,105 @@ pub struct KeyEvent {
...
+    }
+
+    /// The state of the left shift key.
+    pub fn lshift_state(&self) -> ModifiersKeyState {
+        if self.pressed_mods.contains(ModifiersKeys::LSHIFT) {
+            ModifiersKeyState::Pressed
+        } else {
+            ModifiersKeyState::Unknown
+        }
+    }
```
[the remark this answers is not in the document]
kchibisov: done.

src/platform_impl/macos/appkit/event.rs
```diff
@@ -165,6 +136,26 @@ extern_methods!(
             unsafe { msg_send_id![self, charactersIgnoringModifiers] }
         }
 
+        pub fn lshift_pressed(&self) -> bool {
+            let raw_modifiers = self.modifierFlags().bits() as u32;
+            raw_modifiers & NX_DEVICELSHIFTKEYMASK != 0
+        }
```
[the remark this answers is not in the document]
kchibisov: Don't think it'll make a difference here.

PR #2833, opened 2023-06-01
Add support of using custom image as cursor icon on windows

Icemic:

Adds a `Window::set_cursor_custom_icon` method, which we can pass an `Icon` to. And this icon is sightly different from a icon created by `Icon::from_rgba`, because it lacks of hotspot infomation (default to the center of image in fact), so I add an `Icon::from_rgba_cursor` which has extra hotspot parameters.

A custom cursor icon has higher priority than system cursor once it is set (when processing WM_SETCURSOR). And custom cursor icon will be dropped if `Window::set_cursor_icon` is called again.

This feature is only added to Windows platform for now, and `.cur` or `.ani` format is also not supported for potential cross-platform requirements.

Maybe a common `Cursor` struct should be added instead of reusing `Icon`, but on Windows an icon or cursor is not strictly differentiated and I haven't delved into the api of other platforms yet. So I just make minimal changes.

### DIFF
--- a/examples/cursor.rs
+++ b/examples/cursor.rs
@@ -1,10 +1,13 @@
 #![allow(clippy::single_match)]
 
+use std::path::Path;
+
 use simple_logger::SimpleLogger;
 use winit::{
     event::{ElementState, Event, KeyEvent, WindowEvent},
     event_loop::EventLoop,
-    window::{CursorIcon, WindowBuilder},
+    keyboard::{KeyCode, PhysicalKey},
+    window::{CursorIcon, Icon, WindowBuilder},
 };
 
 #[path = "util/fill.rs"]
@@ -22,6 +25,20 @@ fn main() -> Result<(), impl std::error::Error> {
     event_loop.run(move |event, elwt| {
         if let Event::WindowEvent { event, .. } = event {
             match event {
+                WindowEvent::KeyboardInput {
+                    event:
+                        KeyEvent {
+                            state: ElementState::Pressed,
+                            physical_key: PhysicalKey::Code(KeyCode::KeyC),
+                            ..
+                        },
+                    ..
+                } => {
+                    println!("Setting cursor to custom");
+                    let path: &str = concat!(env!("CARGO_MANIFEST_DIR"), "/examples/icon.png");
+                    let icon = load_icon(Path::new(path));
+                    window.set_cursor_custom_icon(icon.clone());
+                }
                 WindowEvent::KeyboardInput {
                     event:
                         KeyEvent {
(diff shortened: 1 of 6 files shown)


msiglreith (changes requested):
Thanks!

I would prefer a dedicated cusor type, where to user can create cursor with a standard icon or with a custom one. At least on X11 IIRC the cursor and icons are different types and mixing them will lead to issues. This would also help to keep a single entry point for the setting the cursor.

madsmtm (commented):
> I would prefer a dedicated cusor type, where to user can create cursor with a standard icon or with a custom one. At least on X11 IIRC the cursor and icons are different types and mixing them will lead to issues. This would also help to keep a single entry point for the setting the cursor.

Fully agree, custom cursors are also supportable on macOS and web, both of which define `PlatformIcon` as `NoIcon`.

We should also consider how this will interact with the `cursor-icon` crate.

kchibisov:
Yeah, we might add a variant to `cursor-icon` crate with the custom cursor image. I'd assume some `ARGB` variant should be fine.

madsmtm:
`cursor_icon::CursorIcon` is `Copy`, so that may be difficult, and perhaps not desired either? There is value in keeping that crate clean of cludges / additions like this.

Instead, we could do something like this (in `winit`):
```rust
// Bikeshed names
pub struct CustomCursorRGBA { ... }
pub struct CustomCursorURL { ... }

trait CursorIconTrait: Sealed { ... }

impl CursorIconTrait for cursor_icon::CursorIcon { ... }
impl CursorIconTrait for CustomCursorRGBA { ... }
impl CursorIconTrait for CustomCursorURL { ... }

impl Window {
    fn set_cursor_icon(&self, icon: impl CursorIconTrait) { ... }
}

//
// Or alternatively:
//

// Bikeshed name
#[non_exhaustive]
pub enum CursorIconEnum {
    Standard(cursor_icon::CursorIcon),
    CustomRGBA(...),
    // E.g. for web
    CustomURL(Url),
}

impl From<cursor_icon::CursorIcon> for CursorIconEnum { ... }
impl From<...> for CursorIconEnum { ... }
impl From<Url> for CursorIconEnum { ... }

impl Window {
    fn set_cursor_icon(&self, icon: impl Into<CursorIconEnum>) { ... }
}
```

kchibisov:
I could add that on Wayland you simply provide and RGBA buffer because you draw the cursor. The animation is also done by you or by compositor, when using named cursors.

daxpedda:
@Icemic we currently have a fully working implementation in https://github.com/rust-windowing/winit/pull/3218 for all backends which went through quite a bit of design work already.

Feel free to drop a review for the Windows backend :)

madsmtm:
Superceded by https://github.com/rust-windowing/winit/pull/3218

PR #2834, opened 2023-06-01
On Web, implement `Send` and `Sync` where appropriate

### LINKED ISSUE #1518: Web worker support
Since web workers support canvas via `OffscreenCanvas`, would it be possible to make winit run in a web worker?

Main issues are these:
- `OffscreenCanvas` implements `EventTarget` so events are not a problem. Although they will have to be forwarded from the main canvas via whatever worker implementation is used. Probably not a `winit` concern.
- `OffscreenCanvas` is different type than `HtmlCanvasElement` and can not be converted in rust. However, since javascript is an interpreted language, inserting `OffscreenCanvas` instead of `HtmlCanvasElement` from js side does not cause a problem. Although browser throws an error due to missing functions (`style()`, `request_fullscreen()`, `set_attribute()`), adding empty stubs fixes the issue and it works happily. This is a bit hacky, but other option would require replacing `backend::RawCanvasType` with enum, which would require a lot of rewrites.
- A lot of functions use `web_sys::window()` and panic on web worker. Could they instead of panicking return default value? Main offenders are `scale_factor()` and `is_fullscreen()` which both are not essential. However, if they return default value would it make debugging harder for regular canvas?

Maybe there's no point in running `winit` in web worker, because the only usable feature without additional glue is the event loop. But I opened this for discussion, as this is relevant for projects that use `winit` and trying to support wasm.

daxpedda:
This implements `Send` and `Sync` for `Window` and `Send` for `EventLoopProxy` on Wasm to make them implement the same traits on all platforms.
I took some inspiration from [`MainThreadSafe` on macOS](https://github.com/rust-windowing/winit/blob/77f8e511e965354a4b827cfb1ca197e9e51e6ded/src/platform_impl/macos/util/async.rs).

It's implemented in a way that if you call this from the `Window` it will do nothing special and just call the appropriate method on the `Window`, as before. Therefor, except a single `Rc<Cell<bool>>` to `Arc<AtomicBool>` change, there is no overhead if Wasm is only used single-threaded.

If used in a multi-threaded environment from a worker, it will use a `mpsc::Sender` to do it's work on the main thread. Currently this is implemented by just sending a boxed function to execute, but it could be rewritten to just send a message and a message handler would then do the right thing.

When it's necessary to return a value a `Condvar` and a `Mutex` is used to await the completion, which would block, but as it's not in the window, where that would fail, thats fine.

Most of the small changes here and there are to disallow calls to `web_sys::window()`, as these are the main culprits that will fail if not called from a window. This was necessary because it was quite hard to figure out what needs to be changed, so I added a Clippy `disallowed-methods`, which should hopefully make maintenance of this feature easy. Let me know what you think of this.

Fixes #1518.
Replaces #2294.
Replaces #2737.
Replaces #2740.

### DIFF
--- a/src/platform/web.rs
+++ b/src/platform/web.rs
@@ -12,7 +12,8 @@ use crate::window::WindowBuilder;
 use web_sys::HtmlCanvasElement;
 
 pub trait WindowExtWebSys {
-    fn canvas(&self) -> HtmlCanvasElement;
+    /// Only returns the canvas if called from inside the window.
+    fn canvas(&self) -> Option<HtmlCanvasElement>;
 
     /// Whether the browser reports the preferred color scheme to be "dark".
     fn is_dark_mode(&self) -> bool;
--- a/src/platform_impl/web/async.rs
+++ b/src/platform_impl/web/async.rs
@@ -1,0 +1,314 @@
+use atomic_waker::AtomicWaker;
+use once_cell::unsync::Lazy;
+use std::future;
+use std::marker::PhantomData;
+use std::ops::Deref;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::mpsc::{self, Receiver, RecvError, SendError, Sender, TryRecvError};
+use std::sync::{Arc, Condvar, Mutex, RwLock};
+use std::task::Poll;
+use wasm_bindgen::prelude::wasm_bindgen;
+use wasm_bindgen::{JsCast, JsValue};
+
+// Unsafe wrapper type that allows us to use `T` when it's not `Send` from other threads.
+// `value` **must** only be accessed on the main thread.
+pub struct MainThreadSafe<const SYNC: bool, T: 'static, E: 'static> {
+    // We wrap this in an `Arc` to allow it to be safely cloned without accessing the value.
+    // The `RwLock` lets us safely drop in any thread.
+    // The `Option` lets us safely drop `T` only in the main thread, while letting other threads drop `None`.
+    value: Arc<RwLock<Option<T>>>,
+    handler: fn(&RwLock<Option<T>>, E),
+    sender: AsyncSender<E>,
+    // Prevent's `Send` or `Sync` to be automatically implemented.
+    local: PhantomData<*const ()>,
+}
+
+impl<const SYNC: bool, T, E> MainThreadSafe<SYNC, T, E> {
+    thread_local! {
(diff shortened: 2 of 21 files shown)


src/platform_impl/web/async.rs
```diff
@@ -0,0 +1,314 @@
...
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::mpsc::{self, Receiver, RecvError, SendError, Sender, TryRecvError};
+use std::sync::{Arc, Condvar, Mutex, RwLock};
+use std::task::Poll;
+use wasm_bindgen::prelude::wasm_bindgen;
+use wasm_bindgen::{JsCast, JsValue};
+
+// Unsafe wrapper type that allows us to use `T` when it's not `Send` from other threads.
+// `value` **must** only be accessed on the main thread.
+pub struct MainThreadSafe<const SYNC: bool, T: 'static, E: 'static> {
```
madsmtm, line 15: Since you're referencing `MainThreadSafe`, I'll note that I've since made [`MainThreadBound`](https://docs.rs/icrate/0.0.2/src/icrate/Foundation/additions/thread.rs.html#170-308), you may be able to take some inspiration there as well.

src/platform/web.rs
```diff
@@ -12,7 +12,8 @@ use crate::window::WindowBuilder;
 use web_sys::HtmlCanvasElement;
 
 pub trait WindowExtWebSys {
-    fn canvas(&self) -> HtmlCanvasElement;
+    /// Only returns the canvas if called from inside the window.
+    fn canvas(&self) -> Option<HtmlCanvasElement>;
```
madsmtm, line 16: This is a breaking change, and should be noted as such in the changelog.

madsmtm (commented):
I like the clippy lint, seems like a good way to go.

src/platform_impl/web/async.rs
daxpedda: Yeah, I've seen that! I was wondering if this thing deserves it's own crate or not ...

src/platform/web.rs
daxpedda: Ah, that is probably what you meant there too: https://github.com/rust-windowing/winit/pull/2764#issuecomment-1572739699.

madsmtm: Yup

daxpedda: Done.

PR #2836, opened 2023-06-01
Replace `instant` with `web-time`

daxpedda:
This replaces [`instant`](https://crates.io/crates/instant) with [`web-time`](https://crates.io/crates/web-time).
Currently `instant` is unmaintained an has a bunch of issues:
- https://github.com/sebcrozet/instant/issues/21
- Panics in [`Instant::duration_since()`](https://github.com/sebcrozet/instant/blob/78e336db632f1f2f7350e2b2385c23d263178251/src/wasm.rs#L24-L27), [unlike std, which currently saturates](https://doc.rust-lang.org/std/time/struct.Instant.html#method.duration_since).
- https://github.com/sebcrozet/instant/issues/49
- No support for [`target_feature = "atomics"`](https://github.com/daxpedda/web-time/blob/b64662bf7aac4f856d681d0053ee1d47294820d4/src/web/instant.rs#L25-L26).

I also gated the dependency behind the Wasm target.

Replaces #2738.


kchibisov (approved):
ack

PR #2837, opened 2023-06-01
Remove mouse event fallback

daxpedda:
There's currently a bunch of fallback code to the mouse event API, which is only there to support old versions of Safari: v<13.
Safari v13 was released in 2019-09-19, so that's almost 4 years.

Can we safely drop support for this?
Running Safari v13 requires at least MacOS 10.13, which is already unsupported by Apple as of 2019-09, or iOS/iPadOS 13, which again is already unsupported by Apple as of 2020-09.

For reference, this would drop support for:
- Any Mac released before late 2009.
- iPhone 6, released in 2014
- iPad 4, Air 1, Mini 3 (the latest, Mini 3, was released in 2014).


daxpedda:
I'm going to close this for now.
Will reopen when it actually becomes a maintenance burden.

PR #2839, opened 2023-06-02
Ignore pen input on Web

daxpedda:
Currently pen input was treated as mouse input, this PR will ignore pen input.
I believe this also aligns with other backends, but would like to get confirmation.


kchibisov (approved):
acked.

PR #2842, opened 2023-06-03
Port to windows-sys v0.48.0

notgull:

Should fix the `cargo deny` errors we've been getting on #2812

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - **Breaking:** Updated `bitflags` crate version to `2`, which changes the API on exposed types.
 - On Web, handle coalesced pointer events, which increases the resolution of pointer inputs.
 - **Breaking:** On Web, `instant` is now replaced by `web_time`.
+- On Windows, port to `windows-sys` version 0.48.0.
 
 # 0.28.6
 
--- a/src/platform_impl/windows/util.rs
+++ b/src/platform_impl/windows/util.rs
@@ -13,7 +13,7 @@ use once_cell::sync::Lazy;
 use windows_sys::{
     core::{HRESULT, PCWSTR},
     Win32::{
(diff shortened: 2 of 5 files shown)


src/platform_impl/windows/util.rs
```diff
@@ -149,6 +149,9 @@ pub fn is_minimized(window: HWND) -> bool {
     unsafe { IsIconic(window) != false.into() }
 }
 
+#[allow(clippy::upper_case_acronyms)]
+pub type HINSTANCE = isize;
```
kchibisov, line 153: It was said that `HMODULE` is the same as `HINSTANCE` nowadays, the difference was only on 16bit windows, so we likely should use `HMODULE`.

https://learn.microsoft.com/en-us/windows/win32/winprog/windows-data-types

kchibisov (changes requested):
Could likely alias, but probably changing `HINSTANCE` to `HMODULE` should do it.

CHANGELOG.md
```diff
@@ -60,6 +60,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On Web, fix pointer button events not being processed when a buttons is already pressed.
 - **Breaking:** Updated `bitflags` crate version to `2`, which changes the API on exposed types.
 - On Web, handle coalesced pointer events, which increases the resolution of pointer inputs.
+- On Windows, port to `windows-sys` version 0.48.0.
```
kchibisov, line 63: I don't think it's meaningful to users. Though the changelog entries have conflicts.

notgull: Disagree; `windows-sys` version churn has been a frequent enough problem that's been resolved as of version 0.48.0 (at least for the time being). Knowledge that we are using this new version may be valuable to users trying to avoid `windows-sys` version churn.

kchibisov (approved):

msiglreith (approved):

PR #2847, opened 2023-06-04
Web: Pointer Cleanup

daxpedda:
# Changes

- Consistently send `WindowEvent::CursorMoved` before pressing or releasing a mouse button.
- Modifiers are now updated on touch events as well.
- Dropping support for versions of Safari below 13. See #2837 for more details notes.
- Fixed text selection by double-clicking or dragging on the canvas on Chrome (apparently not an issue on Firefox).
- Gaining and losing focus on touch now works properly.
- Add support for the [`pointerrawupdate`](https://www.w3.org/TR/pointerevents3/#the-pointerrawupdate-event) event, to allow getting pointer updates as fast as possible.

# Questions (obsolete)

All questions were already answered, just leaving it here for reference.

## Contextmenu

I found that opening context menus can't be completely prevented on Firefox, because they provide a way to bypass any prevention method: Shift+Rightclick. I feel like Winit could document this somewhere, like at [`WindowBuilder::with_prevent_default()`](https://docs.rs/winit/latest/wasm32-unknown-unknown/winit/window/struct.WindowBuilder.html#method.with_prevent_default).

WDYT?

Some notes on this:
- Firefox bug: https://bugzilla.mozilla.org/show_bug.cgi?id=897379.
- [WHATWG spec](https://github.com/whatwg/html/blob/942a9d5b9c0b50bf5bca275e866bb2319ebdca0c/source):
  > User agents may provide means for bypassing the context menu processing model, ensuring that the user can always access the UA's default context menus. For example, the user agent could handle right-clicks that have the Shift key depressed in such a way that it does not fire the `contextmenu` event and instead always shows the default context menu.
- Was removed in https://github.com/whatwg/html/pull/2742. See https://github.com/whatwg/html/issues/6433 and https://github.com/w3c/uievents/issues/336.

## Touch

The Web standard fires `focus/blur` as soon as touch starts and ends, but on Web we currently ignore this and you never get any `WindowEvent::CursorEntered/Left` on touch. How do other backends handle this?

## High Frequency Pointer Input

I added support for `pointerrawupdate`, which is basically the same as `pointermove` but sends an event as soon as it arrives, instead of coalescing multiple events into one. I believe this aligns more with how other backends work. Does that sound alright?


src/platform_impl/web/web_sys/pointer.rs
```diff
@@ -31,133 +32,152 @@ impl PointerHandler {
...
-                            event::mouse_position(&event).to_physical(super::super::scale_factor()),
+                            event::mouse_position(&event).to_physical(super::scale_factor()),
                             event::mouse_button(&event).expect("no mouse button pressed"),
-                            event::mouse_modifiers(&event),
                         );
-
-                        // Error is swallowed here since the error would occur every time the mouse is
-                        // clicked when the cursor is grabbed, and there is probably not a situation where
-                        // this could fail, that we care if it fails.
-                        let _e = canvas.set_pointer_capture(event.pointer_id());
```
daxpedda, line 136: Hey @ryo33!
Do you still remember why you added this in https://github.com/rust-windowing/winit/pull/2188?

ryo33: https://github.com/rust-windowing/winit/commit/c5c99d2357880792a1f552a83eeb67f074103688
It seems added two years ago in this commit. The commit message says the same as the comment:

> * Remove "expect" from fallible code
>
>    code would crash if handling pointer capture outside of winit on
>    every mouse click
>
>    we swallow the error since we could not think of a case where this
>    would fail in a way that would want to handle that it fails

daxpedda: I was talking about the call to `setPointerCapture()`, not why the error is ignored.
In any case, my bad, I just didn't go back far enough: https://github.com/rust-windowing/winit/pull/1672; which addressed https://github.com/rust-windowing/winit/issues/1660.

Now that makes perfect sense, I should definitely not remove that. Will figure out how exactly to sync this with other backends.

Thanks!

src/platform_impl/web/web_sys/pointer.rs
```diff
@@ -173,34 +198,48 @@ impl PointerHandler {
                     fn has_get_coalesced_events(this: &PointerEventExt) -> JsValue;
                 }
 
-                match event.pointer_type().as_str() {
-                    "touch" => {
-                        if prevent_default {
-                            // prevent scroll on mobile web
-                            event.prevent_default();
-                        }
```
daxpedda, line 181: This was added by https://github.com/rust-windowing/winit/pull/2216, it doesn't seem to do anything anymore. @madsmtm do you still remember why this was added specifically?

src/platform_impl/web/web_sys/canvas.rs
```diff
@@ -147,14 +138,6 @@ impl Canvas {
         }));
     }
 
-    pub fn on_touch_end(&mut self, prevent_default: bool) {
-        self.on_touch_end = Some(self.common.add_event("touchend", move |event: Event| {
-            if prevent_default {
-                event.prevent_default();
-            }
-        }));
-    }
```
daxpedda, line 156: Same as above.
Apparently `preventDefault()` on `touchstart` was enough.

src/platform_impl/web/web_sys/pointer.rs
madsmtm: Nope, don't remember, I didn't (and don't) know enough about the web to question it then. Maybe @rukai knows why/if it was necessary?

kchibisov:
>The Web standard fires focus/blur as soon as touch starts and ends, but on Web we currently ignore this and you never get any WindowEvent::CursorEntered/Left on touch. How do other backends handle this?

`CursorEntered/Left` are only for pointer, touch doesn't really have such concept in my opinion and we don't have a backend doing any of that for touch. With touch you simply press on a window without moving any logical resource, the user can click in any point of the screen, so the only thing matters is point of click and release of the touch slot.

>I added support for pointerrawupdate, which is basically the same as pointermove but sends an event as soon as it arrives, instead of coalescing multiple events into one. I believe this aligns more with how other backends work. Does that sound alright?

yeah, the events are usually send as soon as possible on other backends.

>Modifiers are now updated on touch events as well.

Do they simply attached to the touch events as well on the web? Usually touch input is done without any sort of keyboard input at the same time, because, you know, you use your hands... Though, maybe I don't understand something and a concept of `Touch` + `Shift` exists.

Not sure what other input you've requested on irc.

daxpedda:
> Do they simply attached to the touch events as well on the web?

Yeah, they simply attach, it has nothing to do with the actual touch input. We just use it to figure out what modifiers are active after we regain focus. This works because `PointerEvent` inherits `MouseEvent` ... it's pretty weird.

> Not sure what other input you've requested on irc.

No that's it, no more open questions, thank you very much :+1:!

src/platform_impl/web/web_sys/pointer.rs
rukai: I believe this was a necessary part of https://github.com/rust-windowing/winit/pull/2216#discussion_r962224122

As long as the functionality added in the pr is upheld, then I have no issue with it being removed 
That is, when winit is configured to, scrolling should be disabled on both web and mobile browsers.

daxpedda: Yeah, from what I tried the functionality is upheld.
Thanks!

src/platform_impl/web/web_sys/canvas.rs
daxpedda: Resolved in https://github.com/rust-windowing/winit/pull/2847#discussion_r1217186189.

daxpedda:
Apologies ... I meant to squash this actually.

cybersoulK:
@daxpedda removing Mouse Handler is a huge downgrade for game developers. Currently it's the only decent option for games. Pointer doesn't allow pressing both left and right at the same time, and messes up once they are pressed at the same time, since the release is ommited.

daxpedda:
Indeed, the initial implementation didn't handle chorded buttons. This was addressed in #2838.
If you try the latest master it should work just fine.

Let us know if you encounter any issue!

cybersoulK:
@daxpedda 
i am using winit = { git = "https://github.com/rust-windowing/winit" }
but it seems to have the same behaviour as the previous Pointer. 

Wow i am excited to see it being worked on. let me know what i can do to help.

cybersoulK:
@daxpedda you are right! It works fine with Firefox and safari.

The Issue is only in: Chome (Mac) Version 114.0.5735.106 (Official Build) (arm64)

I am also having a new issue now on firefox (Mac) that mousemove seems to be buggy (my character looks at the ground), should i create a separate issue for each?

cybersoulK:
@daxpedda The 2 issues happen too, when i test on Windows. 

Chrome can't do chorded mouse buttons, firefox has a bug in DeviceEvent::MouseMotion

daxpedda:
So apparently Chrome has some bugs with `pointerrawupdate`:
- https://bugs.chromium.org/p/chromium/issues/detail?id=1289683
  No chorded buttons when the pointer is locked.
- https://bugs.chromium.org/p/chromium/issues/detail?id=1068958
  Wrong positions when in an inline frame.

I think I will rip it out completely until these issues are fixed, otherwise it's kind of unusable.

cybersoulK:
@daxpedda yea that was it! it works perfectly now. Thank you so much, i was dealing with this for about a year.

firefox deltas still seems buggy, as of the current master branch

daxpedda:
> firefox deltas still seems buggy, as of the current master branch

This is fixed in #2871, but I will actually split off the fix.

PR #2852, opened 2023-06-05
Add WASM CustomEvent Example

### LINKED ISSUE #2816: Add WASM CustomEvent Example
The current CustomEvent example in the `./examples` folder doesn't show how to do it via WASM.

https://github.com/rust-windowing/winit/blob/master/examples/custom_events.rs

Could we add an example like this + a basic HTML file to show off how to fire custom events from the JS side into the WASM winit event loop?

https://github.com/rust-windowing/winit/discussions/2061#discussioncomment-5813125

stehyn:
The current CustomEvent example in the ./examples folder doesn't show how to do it via WASM.

Atm I'm just printing to console, however we could reuse the `examples/web.rs` code to append to a HTML list. But I wanted to keep the example a bit more streamline.

Also I added some extra CSS to centre align it. The button will be placed on the right of the canvas otherwise. lmk if it's uggo and I'll revert it.

image

Closes #2816
Answers https://github.com/rust-windowing/winit/discussions/2061#discussioncomment-5813125

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -150,7 +150,6 @@ features = [
     'PointerEvent',
     'Window',
     'WheelEvent',
-    "HtmlScriptElement",
 ]
 
 [target.'cfg(target_family = "wasm")'.dev-dependencies.web_sys]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -149,7 +149,8 @@ features = [
     'Node',
     'PointerEvent',
     'Window',
-    'WheelEvent'
+    'WheelEvent',
+    "HtmlScriptElement",
 ]
 
 [target.'cfg(target_family = "wasm")'.dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -153,6 +153,11 @@ features = [
     "HtmlScriptElement",
 ]
 
+[target.'cfg(target_family = "wasm")'.dev-dependencies.web_sys]
+package = "web-sys"
+version = "0.3.22"
+features = ['HtmlScriptElement']
+
 [target.'cfg(target_family = "wasm")'.dependencies]
 js-sys = "0.3"
 wasm-bindgen = "0.2.45"
--- a/examples/wasm_custom_event.rs
+++ b/examples/wasm_custom_event.rs
@@ -92,7 +92,7 @@ mod wasm {
         let body = document.body().unwrap();
 
         // Set a background color for the canvas to make it easier to tell where the canvas is for debugging purposes.
-        canvas.style().set_property("background-color", "crimson");
+        canvas.style().set_property("background-color", "crimson").unwrap();
         body.append_child(&canvas).unwrap();
 
         // Create script element
--- a/examples/wasm_custom_event.rs
+++ b/examples/wasm_custom_event.rs
@@ -1,88 +1,89 @@
-#![allow(clippy::single_match)]
-
-use std::cell::RefCell;
-use wasm_bindgen::prelude::*;
-use winit::{
-    event::{Event, WindowEvent},
-    event_loop::{EventLoop, EventLoopBuilder, EventLoopProxy},
-    window::WindowBuilder,
-};
-
-#[derive(Debug, Clone, Copy)]
-pub enum CustomEvent {
-    Add { a: u32, b: u32 },
-}
-
-thread_local! {
-    pub static EVENT_LOOP_PROXY: RefCell<Option<EventLoopProxy<CustomEvent>>> = RefCell::new(None);
-}
-
-fn add(a: u32, b: u32) -> u32 {
-    a + b
-}
-
-#[cfg_attr(target_arch = "wasm32", wasm_bindgen(js_name = "fireAdd"))]
-pub fn fire_add() {
-    EVENT_LOOP_PROXY.with(|proxy| {
-        if let Some(event_loop_proxy) = proxy.borrow().as_ref() {
-            event_loop_proxy
-                .send_event(CustomEvent::Add { a: 1, b: 2 })
-                .ok();
-        }
-    });
-}
-
+#[cfg(not(wasm_platform))]
 pub fn main() {
-    let event_loop: EventLoop<CustomEvent> =
-        EventLoopBuilder::<CustomEvent>::with_user_event().build();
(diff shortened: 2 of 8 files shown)


Cargo.toml
```diff
@@ -149,7 +149,8 @@ features = [
     'Node',
     'PointerEvent',
     'Window',
-    'WheelEvent'
+    'WheelEvent',
+    "HtmlScriptElement",
```
daxpedda, line 153: This should go into `dev-dependencies`.
Also apparently we are using `'` instead of `"`.

examples/wasm_custom_event.rs
```diff
@@ -0,0 +1,130 @@
...
+    use wasm_bindgen::prelude::*;
+    use wasm_bindgen::JsCast;
+    use web_sys::HtmlScriptElement;
+    use winit::event::Event;
+    use winit::event::WindowEvent;
+    use winit::event_loop::EventLoop;
+    use winit::event_loop::EventLoopBuilder;
+    use winit::event_loop::EventLoopProxy;
+    use winit::window::Window;
+    use winit::window::WindowBuilder;
```
daxpedda, line 27:
```suggestion
    use winit::event::{Event, WindowEvent};
    use winit::event_loop::{EventLoop, EventLoopBuilder, EventLoopProxy};
    use winit::window::{Window, WindowBuilder};
```

examples/wasm_custom_event.rs
```diff
@@ -0,0 +1,130 @@
...
+
+    // Because EventLoopProxy is not Send, we need to wrap it in a RefCell and use thread_local!
+    thread_local! {
+        pub static EVENT_LOOP_PROXY: RefCell<Option<EventLoopProxy<CustomEvent>>> = RefCell::new(None);
+    }
```
daxpedda, line 32: Since #2834 it is actually!
In any case it's still appropriate to use `thread_local!` if we aren't doing anything with web workers.
We should also probably use `once_cell::unsync::OnceCell` instead of `RefCell`.

examples/wasm_custom_event.rs
```diff
@@ -0,0 +1,130 @@
...
+
+    // Function to be called from JS
+    fn wasm_call() -> u32 {
+        42
+    }
+
+    #[derive(Debug, Clone, Copy)]
+    pub enum CustomEvent {
+        WasmCall,
+    }
```
daxpedda, line 42: Let's make a slightly more useful example. Instead of just printing 42 we could e.g. use `Window::set_inner_size()`, which is a valid use-case and would also show how to pass parameters. This would require two input fields for width and height and a button to send.

examples/wasm_custom_event.rs
```diff
@@ -0,0 +1,130 @@
...
+        insert_canvas(&window);
+
+        event_loop.run(move |event, _, control_flow| {
+            control_flow.set_wait();
+
+            match event {
+                Event::WindowEvent {
+                    event: WindowEvent::CloseRequested,
+                    window_id,
+                } if window_id == window.id() => control_flow.set_exit(),
```
daxpedda, line 69: We can skip that here, it's not a thing in Wasm.

examples/wasm_custom_event.rs
```diff
@@ -0,0 +1,130 @@
...
+        use winit::platform::web::WindowExtWebSys;
+
+        let canvas = window.canvas();
+
+        let window = web_sys::window().unwrap();
+        let document = window.document().unwrap();
+        let body = document.body().unwrap();
+
+        // Set a background color for the canvas to make it easier to tell where the canvas is for debugging purposes.
+        canvas.style().set_css_text("background-color: crimson;");
```
daxpedda, line 102:
```suggestion
        canvas.style().set_property("background-color", "crimson");
```

daxpedda (changes requested):
@rukai I guess I didn't really think my last comment through (https://github.com/rust-windowing/winit/issues/2816#issuecomment-1575828967) , but you can't really show the JS in the generated HTML file because it's inserted during runtime, or is there something `cargo-run-wasm` can do about that I'm not aware of?

Additionally, it would also be really nice to not have to build a bunch of HTML through JS or `web-sys`. Again, none of this can actually be shown to the user because it's all added during runtime.

rukai:
> @rukai I guess I didn't really think my last comment through (https://github.com/rust-windowing/winit/issues/2816#issuecomment-1575828967) , but you can't really show the JS in the generated HTML file because it's inserted during runtime, or is there something cargo-run-wasm can do about that I'm not aware of?

Oh uh... true. :sweat_smile: 
Scratch that idea then.

Next best thing I can think of would be just manually creating an unused html file with the appropriate javascript in it :/

daxpedda:
> Next best thing I can think of would be just manually creating an unused html file with the appropriate javascript in it :/

Would you accept a feature request extending the API of `cargo-run-wasm` with something like `with_html_body()`?

rukai:
Yep, I think we do need this feature, I've raised https://github.com/rukai/cargo-run-wasm/issues/35
Not sure when I'll get to implementing it.

stevenhuyn:
I could consider looking into adding that feature

examples/wasm_custom_event.rs
stevenhuyn: I am not certain want to make the comment now. Not super familiar with web workers. (js is single threaded unless you use web-workers is my understanding (?))

examples/wasm_custom_event.rs
stevenhuyn: Skip this match arm?

```rust
Event::WindowEvent {
    window_id,
} if window_id == window.id() => control_flow.set_exit(),
```

examples/wasm_custom_event.rs
stevenhuyn: Awaiting if the cargo-wasm-runner will go through first. https://github.com/rukai/cargo-run-wasm/pull/36

daxpedda:
@stevenhuyn you should keep the conversation you are having in https://github.com/rust-windowing/winit/issues/2816#issuecomment-1577864096 here!
But yeah, I guess that's the best we can do right now. I will take a closer look at it during review.

stevenhuyn:
Ah thought the issue is the ground truth, although I guess implementation details should be in this PR.

PR #2853, opened 2023-06-05
Document that Winit controls the canvas size

### LINKED ISSUE #1491: How to handle window size, and cursor position on web targets?
I've ran into some pretty major issues when trying to handle window (aka: canvas) size on web targets. First i can't get window resize events, because CSS is quietly resizing the window to fit the screen, which means I need to call web-sys and see if it was resized every time before drawing, but now this means the cursors position is incorrect, because a canvas can have a higher or lower resolution then the actual window.

daxpedda:
Would appreciate somebody taking a second look.

Fixes #1491.


daxpedda:
This won't be necessary anymore, already working on https://github.com/rust-windowing/winit/issues/1661.

PR #2854, opened 2023-06-05
Introduce Dependabot

daxpedda:
See https://docs.github.com/en/code-security/dependabot.

Hopefully this will help us with keeping our dependencies up-to-date.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+
+updates:
+  - package-ecosystem: cargo
+    directory: /
+    schedule:
+      interval: daily
+
+  - package-ecosystem: github-actions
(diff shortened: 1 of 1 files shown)


kchibisov:
I think there's a patch with it already. I initially said that I'm not a big fan, given that we maintain all the deps anyway, so we know when to bump them and how.

daxpedda:
Ah, sorry: https://github.com/rust-windowing/winit/pull/2765.
I think it doesn't hurt to have it and it helps with reminding, especially unmaintained backends or dev-dependencies sometimes aren't updated.

Feel free to close if this doesn't convince you.

daxpedda:
I'm gonna close this considering nobody else chimed in.

PR #2858, opened 2023-06-07
Fix `Window::set_inner_size()`

daxpedda:
So since #2778 properly decoupling internal and visual canvas size, we should make things consistent.

### Before (#2778)

Using `Window::set_inner_size()` would apply physical pixels to the internal canvas size and logical pixel to the visual canvas size. The logical pixel would be applied to the canvas size as physical pixels by setting CSS width and height as `px` values.

E.g.: `Window::set_inner_size(PhysicalSize::new(500, 500))` with a scale factor of 2. Would apply 500x500px to the the internal canvas size and 250x250px to the visual size. `Window::inner_size()` would report a size of 500x500px.

In the meanwhile the reported size by `Window::inner_size()` would be the 

### Currently (after #2778)

Same as before, but it would only apply the visual size, internal size wouldn't be touched at all.

### After (this PR)

`Window::set_inner_size()` will apply the correct size to the canvas. E.g. `Window::set_inner_size(PhysicalSize::new(500, 500))` will apply 500x500px to the visual size of the canvas, no matter the current scale factor.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,7 +54,8 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On Wayland, drop `WINIT_WAYLAND_CSD_THEME` variable.
 - Implement `PartialOrd` and `Ord` on types in the `dpi` module.
 - **Breaking:** Bump MSRV from `1.60` to `1.64`.
-- **Breaking:** On Web, the canvas output bitmap size is no longer adjusted.
+- **Breaking:** On Web, `Window::(set_)inner_size()` will return/change the visual canvas size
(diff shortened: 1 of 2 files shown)


Liamolucko:
I'm against this change - I think the old behaviour was correct.

A CSS "pixel" is not a physical pixel, it's a logical pixel. On a screen with a scale factor of 2, one CSS pixel represents a 2x2 grid of physical pixels. So, setting the CSS size of the canvas to the logical size of the window was correct.

daxpedda:
Reverted in #2861.

PR #2859, opened 2023-06-07
Implement `ResizeObserver`

### LINKED ISSUE #1661: Allow canvas size to be controlled by page layout on web target

### LINKED ISSUE #2864: On Web, don't scale the canvas on scale factor changes
The [documentation says about `WindowEvent::ScaleFactorChanged::new_inner_size`](https://docs.rs/winit/0.28.6/winit/event/enum.WindowEvent.html#variant.ScaleFactorChanged):
> By default, this will contain the size suggested by the OS, but it can be changed to any value.

This isn't really possible on Web, because the Web doesn't suggest anything. After some IRC discussion we established that to be consistent with other backends, we should do whats usually "expected" here and scale the canvas with the scale factor, which is how it works currently anyway.

But this is often not desirable, because users usually want to control the size of the canvas through CSS and don't want Winit to overwrite it, which will be made possible by #2859. But changing the scale factor, will currently **by default** overwrite the any user defined CSS values, unless the user changes `WindowEvent::ScaleFactorChanged::new_inner_size` back to it's old value, which was taken into account by #2860.

I'm suggesting we should lean into the "suggested by the OS" argument, which in this case would be whatever CSS says is the new value, not overwriting any values set through CSS **by default**.

daxpedda:
### Accuracy

We have to different forms of accuracy here, which I will name "pixel perfect" and "precise".

"Pixel perfect" information about the size of the canvas is really important, see this [blog post](https://web.dev/device-pixel-content-box) for more details, which includes a [nice visual demo](https://device-pixel-content-box.glitch.me) of the problems that can arise without such information.
But to summarize, the user absolutely requires this information to determine at what resolution to render to the canvas, otherwise they will get visual artifacts. This is important because we calculate physical pixels by converting CSS pixels with the scale factor, if the scale factor is fractional result, we have to **round**, which may or may not match the actual size in physical pixels on the users screen.
This is only a problem with fractional scale factors or if the canvas is positioned with fractional CSS pixel values.
In any case, the only way to accurately get information about the actual physical pixel size of the canvas is `devicePixelContentBoxSize`.

"Precise" describes the fact that some APIs report accurate CSS pixel values, while others just don't return fractional values. We want precise values to be able to accurately convert to physical pixels with the scale factor. Again, this is only a problem with fractional scale factors or if the canvas is positioned with fractional CSS pixel values.

### Consistency between `Window::set_inner_size()`, `Window::inner_size()` and `WindowEvent::Resized`

Preferably the size set by `set_inner_size()` should be the same as the one returned by `inner_size()` and the one received by the `WindowEvent::Resized` event. This is difficult because a lot of the APIs are inconsistent with each other or require imprecise conversions.

To clarify why conversions are relevant here: there is no API to set the size by physical pixels in the Web, you can only set the size of elements in logical pixel units (plenty of other unit types, but they aren't relevant here, see [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Values_and_Units#distance_units)).

`devicePixelContentBoxSize` can never be accurately converted to or from, because it reports pixel perfect physical sizes, any conversion can only round results. But being a pixel off might be an acceptable tradeoff.

`getBoundingClientRect()` respects `scale`/`transform`, which would require significant effort on our side to calculate conversions to or from. See the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function) for some examples what `transform` can do. So we will either have to document that using `scale`/`transform` might significantly alter the output or just refrain from using this API.

### Sync `Window::set_inner_size()` and `Window::inner_size()`

Unfortunately `ResizeObserver` cannot be queried, we receive it's values only through callbacks. So we will have to decide if we would rather give the user inaccurate information through `Window::inner_size()` until they receive a `WindowEvent::Resized` or just not update `Window::inner_size()` until we get accurate values, which the user will be notified of by `WindowEvent::Resized`.

## API Overview

**API:** Name of the API.
**queryable:** Can be queried at any point in time, in contrast to having to wait for a callback.
**`scale`/`transform`:** Will adjust the reported size depending on any `scale` or `transform` properties.
**precise:** Reports CSS pixel values without rounding to the nearest integer.
**pixel perfect:** Reports the size of the element with exact physical pixels.

| API | queryable | `scale`/`transform` | precise | pixel perfect |
|---|:---:|:---:|:---:|:---:|
| [`devicePixelContentBoxSize`](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserverEntry/devicePixelContentBoxSize) | 🯀 | 🯀 | 🯀 | ✅ |
| [`contentRect`](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserverEntry/contentRect)[^1] | 🯀 | 🯀 | ✅ | 🯀 |
| [`getBoundingClientRect()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect) | ✅ | ✅ | ✅ | 🯀 |
| [`getComputedStyle()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle) | ✅ | 🯀 | ✅ | 🯀 |
| [`scrollWidth/Height`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollWidth)[^2] | ✅ | 🯀 | 🯀 | 🯀 |

[^1]: It is marked for deprecation in the future, but is the same as [`contentBoxSize`](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserverEntry/contentBoxSize). Used here because it's the only option in `ResizeObserverEntry` that is supported by Safari.
[^2]: There is also [`clientWidth/Height`](https://developer.mozilla.org/en-US/docs/Web/API/Element/clientWidth), which doesn't include content hidden by a scrollbar and [`offsetWidth/Height`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetWidth), which is similar to `clientWidth/Height` but includes borders.

## API Analysis

We definitely want to use `devicePixelContentBoxSize` somehow. If we don't make it the value returned by `Window::inner_size()` and `WindowEvent::Resized`, we could at least expose it through `WindowExtWebSys`.

If we really want `Window::set_inner_size()` and `Window::inner_size()`/`WindowEvent::Resized` to be consistent, we can use `contentRect`.

Making sure that `Window::inner_size()` is able to return the correct/same value right after using `Window::set_inner_size()` is problematic. `getBoundingClientRect()` respects `scale`/`transform` and `scrollWidth/Height` is imprecise.

## Conclusion

I think using `devicePixelContentBoxSize` has the proper tradeoffs for Winit. We get pixel perfect sizes and I find the tradeoffs acceptable.

We can treat `Window::set_inner_size()` as a request, so it isn't fulfilled until `WindowEvent::Resized` is received. It being "just a request", I can further extrapolate on saying that it's fine if the received size by `Window::inner_size()`/`WindowEvent::Resized` isn't always the same as the one passed into `Window::set_inner_size()`, which will commonly be the same anyway or at most a pixel off.

### Changes

- Removed fullscreen detection completely, as it was only changing the size of the canvas.
- We now don't change the size of the canvas at all unless the user uses `Window::set_inner_size()` or changes `WindowEvent::ScaleFactorChanged::new_inner_size`.
- `Window::inner_size()` now always returns the precise physical pixel size of the canvas, although not in sync with `Window::set_inner_size()`, which now is treated as a request only, see #2868.
- Scale factor changes will always be reported on the Web before resize events, so we wouldn't get the current size of the canvas after a scale factor change. This PR introduces a mechanic to create oneshot `ResizeObserver`s, that report the size back when needed. Therefore the resize and scale factor handlers are now merged, as they are closely intertwined.
- Safari doesn't support `devicePixelContentBoxSize`, which had to be accounted for when scale factor changes, as that could affect the physical size without being reported by `contentRect`, the fallback to `devicePixelContentBoxSize`.
- Introduced a new `EventWrapper` type, that allows synchronous event execution without requiring global handlers. Now that resize and scale factor event handlers are merged, they can't be global, but have to be made per canvas.
- To account for Safari not supporting the initial event in `ResizeObserver` if the size of the element is zero, we had to do some internal size tracking to inhibit duplicate `Resized` events.

### New scale factor flow

When we notify the resize handler, it will re-observer the canvas with the `ResizeObserver` to get the initial event and get the size from there, if the browser supports `devicePixelContentBoxSize`. If the browser doesn't, it will just use `getComputedStyle()` to calculate the inner size of the canvas without any callbacks involved.

1. Scale factor change was detected.
2. Notify the resize handler.
3. When we receive the new current size, we send `ScaleFactorChanged` to the user.
   - If the size didn't change from before and the user didn't change `new_inner_size`, we do nothing.
   - If the size changed and the user didn't change `new_inner_size`, we send a `Resized` event.
   - If the user changes `new_inner_size`, we notify the resize handler again and report `Resized` from there.

---

Fixes #1661.
Fixes #2864.
Replaces #2074.


Liamolucko:
> ... resulting in:
> 
> * Scale factor is not zoom level.
> * There is no API to query the zoom level.
> * `devicePixelContentBoxSize` always reports exactly the physical size in pixels an element occupies on the screen, ignoring scale and zoom, but not pinch to zoom.
> * Converting between any size and `devicePixelContentBoxSize` by using the scale factor would work with zoom, but would be wrong without zoom, e.g. a mobile device.
>
> Therefor there is no way to set `devicePixelContentBoxSize` on an element and there is no way to calculate a conversion reliably.

I don't understand why that prevents converting between `devicePixelContentBoxSize` and the CSS size of the canvas. My understanding is that:

- Zooming with ctrl +/- properly updates both `devicePixelRatio` and `devicePixelContentBoxSize` so that you can calculate the regular size from `devicePixelContentBoxSize` (except in Safari, actually, where `devicePixelRatio` is unchanged, but it doesn't have `devicePixelContentBoxSize` anyway so we can ignore that for now).
- Pinch-to-zoom affects neither `devicePixelRatio` nor `devicePixelContentBoxSize`, and so pretty much has no effect on the page - it literally just zooms into the screen without touching anything else. I've tested that this is the case in desktop Firefox and Chrome, but I haven't been able to test it on mobile - is that where the problem comes from?

So, I think it should be perfectly possible to convert between `devicePixelContentBoxSize` and the CSS size using `devicePixelRatio`. Am I missing something?

mcclure:
In my testing #2074 seems to work perfectly well with "normal" ctrl-+ browser zoom.

src/platform_impl/web/web_sys/canvas.rs
```diff
@@ -73,22 +75,16 @@ impl Canvas {
...
+            super::set_canvas_size(&window, &canvas, size);
+        }
 
-        let canvas = Canvas {
+        Ok(Canvas {
             common: Common {
                 window,
                 raw: canvas,
-                size: Rc::new(Cell::new(size)),
+                size: Rc::default(),
```
Liamolucko, line 87: Just to double-check: you're aware that doing this means that `Window::inner_size` will return `(0, 0)` until the first `ResizeObserver` callback happens, usually after `EventLoop::run` is called? I'm not necessarily against that change but it should probably be documented, since it's a pretty important caveat to know about.

This used to be a bigger problem because `wgpu` would panic when you passed it a canvas with a framebuffer size of 0, but that's no longer a issue since winit no longer sets that. That's why I'm now fine for it to work this way.

src/platform_impl/web/event_loop/runner.rs
```diff
@@ -348,16 +347,11 @@ impl<T: 'static> Shared<T> {
...
-                backend::set_canvas_size(&canvas, crate::dpi::Size::Physical(new_size));
-                self.handle_single_event_sync(
-                    Event::WindowEvent {
-                        window_id: id,
-                        event: crate::event::WindowEvent::Resized(new_size),
-                    },
-                    &mut control,
-                );
+                let canvas = &canvas.borrow();
+                backend::set_canvas_size(canvas.window(), canvas.raw(), new_size.cast());
```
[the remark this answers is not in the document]
daxpedda: Currently this is the intention of `ScaleFactorChanged::new_inner_size`. But I did address this issue in #2860, if users don't want Winit to change or overwrite a size set by CSS, they have to change `new_inner_size` to the old value.

I agree that this is not desirable on Web, but I'm not sure this can be changed with the current API without introducing inconsistent behavior between backends. There might be room for some design work to allow for this, I was thinking of changing `new_inner_size` from `&mut PhysicalSize<u32>` to `&mut Option<PhysicalSize<u32>>`. In any case, this is off-topic for this PR.

An alternative we could apply with how things currently work in Winit, is to set the `new_inner_size` suggested by `ScaleFactorChanged` to one that doesn't actually change anything, ergo doesn't scale. The argument would be that `new_inner_size` is "suggested by the OS", we could argue that Web makes the suggestion to not change the size.

Would like some input from other maintainers about this topic.

src/platform_impl/web/event_loop/runner.rs
```diff
@@ -348,16 +347,11 @@ impl<T: 'static> Shared<T> {
                 &mut control,
             );
 
-            // Then we resize the canvas to the new size and send a `Resized` event:
+            // Then we resize the canvas to the new size, a new
+            // `Resized` event will be sent by the `ResizeObserver`:
```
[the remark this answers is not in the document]
daxpedda: > This isn't necessarily true. When the scale factor changes, it's possible for the logical size to change but the physical size to stay the same (e.g., if the canvas takes up 100% of the window and the user zooms in), in which case the `ResizeObserver` won't trigger because it's only watching the physical size. This is why my implementation used two `ResizeObserver`s, one for each size.

I didn't understand why this was necessary in #2074, we already get a notification when the scale factor changes, how does this help us here?

The comment is correct, **if** the physical size of the canvas actually changes, we will get a notification, as far as I'm aware we don't really care if it's logical size changes.
> Note that you can't get the same `ResizeObserver` to observe both sizes of an element, when you tell it to observe the same element twice it cancels the first observation.

This is good to know, thank you.

src/platform_impl/web/web_sys/canvas.rs
daxpedda: I was aware. Though this isn't new behavior to Winit, just new to Web. So users of `wgpu` had to check for zero-sized windows already (I encountered this myself as well on X11). Though looking at the documentation I can see no mention of it anywhere, so it would probably be good to document this somewhere.

This is probably off-topic for this PR, but we might want to consider a fix for this, like returning `None` if the Window doesn't have a size yet. This would be good for Web as well because if the canvas isn't inserted into the DOM it doesn't have a size either.

src/platform_impl/web/event_loop/runner.rs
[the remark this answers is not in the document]
Liamolucko: Oops, I completely forgot that `new_inner_size` existed. The current situation of requiring that you manually set it to the old value indeed isn't ideal, though. 

The ideal solution in my opinion would be to find out what the new size is as determined by CSS, report that in `new_inner_size`, and then only set `width`/`height` if it was changed. That's what I implemented in #2074.

> An alternative we could apply with how things currently work in Winit, is to set the `new_inner_size` suggested by `ScaleFactorChanged` to one that doesn't actually change anything, ergo doesn't scale. The argument would be that `new_inner_size` is "suggested by the OS", we could argue that Web makes the suggestion to not change the size.

That's basically what I'm suggesting, except that it needs be the value that doesn't change the _new_ size of the canvas, not the size before the scale factor changed.

src/platform_impl/web/event_loop/runner.rs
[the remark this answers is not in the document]
Liamolucko: > The comment is correct, **if** the physical size of the canvas actually changes, we will get a notification, as far as I'm aware we don't really care if it's logical size changes.

That's true, I guess you don't really need to be notified if only the logical size changes. I think I was just playing it safe in #2074 and making sure that an event got sent any time it could possibly be warranted.

When you're observing the logical size of the canvas, though, it's also possible to miss a change in the _physical_ size. For instance, if the canvas has a fixed `width`/`height` set in CSS, zooming in will change the physical size but not the logical size, and so it won't get picked up. So, you probably need to add some code to `scale_factor_changed` which checks for that situation, and manually dispatches a `Resized` event if the `ResizeObserver` isn't going to trigger.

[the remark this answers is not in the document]
daxpedda: > When you're observing the logical size of the canvas, though, it's also possible to miss a change in the _physical_ size.

You mean on Safari I assume? Yeah, good point.

src/platform_impl/web/web_sys/mod.rs
```diff
@@ -52,38 +53,13 @@ impl WindowExtWebSys for Window {
...
-    let logical_size = new_size.to_logical::<f64>(scale_factor);
-    set_canvas_style_property(canvas.raw(), "width", &format!("{}px", logical_size.width));
-    set_canvas_style_property(
-        canvas.raw(),
-        "height",
-        &format!("{}px", logical_size.height),
-    );
+pub fn set_canvas_size(raw: &HtmlCanvasElement, new_size: LogicalSize<f64>) {
+    set_canvas_style_property(raw, "width", &format!("{}px", new_size.width));
+    set_canvas_style_property(raw, "height", &format!("{}px", new_size.height));
```
Liamolucko, line 62: Why'd you get rid of all the logic for dealing with padding?

daxpedda: I just reverted the last commit, I will re-add that logic, keep in mind this is still a draft!
Don't let me discourage you from leaving any comments though, your input is appreciated :heart:!

src/platform_impl/web/web_sys/canvas.rs
daxpedda: Tracking this in https://github.com/rust-windowing/winit/issues/2863.

src/platform_impl/web/event_loop/runner.rs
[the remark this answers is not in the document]
daxpedda: Tracked in https://github.com/rust-windowing/winit/issues/2864.

src/platform_impl/web/web_sys/mod.rs
daxpedda: Added back in https://github.com/rust-windowing/winit/pull/2859/commits/4a9b1a5612f461366a19aef076af0d48761b8eff.

src/platform_impl/web/event_loop/runner.rs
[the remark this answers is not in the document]
daxpedda: Done, I also addressed #2864 while I'm at it.

daxpedda:
I rewrote large parts of the OP because of a misconception I had about what exactly CSS pixels represent. Thank you @Liamolucko for pointing it out to me in https://github.com/rust-windowing/winit/pull/2858#issuecomment-1581820877.

> So, I think it should be perfectly possible to convert between `devicePixelContentBoxSize` and the CSS size using `devicePixelRatio`. Am I missing something?

Even after I rewrote the OP and cleared up any misconceptions I had, it's actually not possible to perfectly convert between `devicePixelContentBoxSize` and the CSS size. It might or might not be a off by a pixel, which may have some very unfortunate consequences as pointed out in the OP, see this [demo] for example.

So I've reworked the PR three times now, any review is appreciated. In the meantime we will be waiting for a new version of `web-sys` anyway.

src/platform_impl/web/web_sys/resize_scaling.rs
```diff
@@ -0,0 +1,288 @@
...
+                    let size = Self::process_entry(&this.window, &this.canvas, entries);
+                    this.resize_enabled.set(true);
+
+                    handler(&mut this, size)
+                }
+
+                observer.disconnect();
+            });
+        Self::create_observer(&self.canvas, &observer_closure);
+    }
```
Liamolucko, line 164: Hah, I actually had the same idea overnight and was about to suggest it, great minds think alike I guess. 👍 

One thing to be wary of though is that `ResizeObserver` isn't 100% guaranteed to call the callback with an initial resize - the way it's spec'd is that it's initialised with a last reported size of `(0, 0)`, and so if the canvas has a size of `(0, 0)` the callback won't fire.

It turns out Safari is the only browser that actually follows the spec here, though - Firefox and Chrome will always fire an initial resize. Specifically, Safari seems to not fire an initial resize for anything below about 0.018px.

You could probably also avoid having to create a temporary `ResizeObserver` here by calling `unobserve` then `observe` again on the canvas with the existing `ResizeObserver`, which will reset everything and fire another initial resize. (This is another place where browsers deviate from the spec - `observer.observe` is supposed to internally call `observer.unobserve` if the element's already being observed, but this didn't work in Firefox, Chrome nor Safari when I actually tried it.)

daxpedda: > It turns out Safari is the only browser that actually follows the spec here, though - Firefox and Chrome will always fire an initial resize. Specifically, Safari seems to not fire an initial resize for anything below about 0.018px.

Well that's really bad, because then the associated `WindowEvent`s would just never fire. In addition to the closure not being cleaned up. Right now it's at least not holding any `Rc`s, only `Weak`s, so it shouldn't prevent other things from being cleaned up. I will do some research.

> You could probably also avoid having to create a temporary `ResizeObserver` here by calling `unobserve` then `observe` again on the canvas with the existing `ResizeObserver`, which will reset everything and fire another initial resize.

That's good to know! I will look into it.

daxpedda: > It turns out Safari is the only browser that actually follows the spec here, though - Firefox and Chrome will always fire an initial resize. Specifically, Safari seems to not fire an initial resize for anything below about 0.018px.

For reference: https://www.w3.org/TR/resize-observer/#intro
> Observation will fire when observation starts if Element is [being rendered](https://html.spec.whatwg.org/#being-rendered), and Element’s size is not 0,0.

But apparently this has changed in https://github.com/w3c/csswg-drafts/issues/3664 and now observers are required to fire an event no matter the size.
This is also tested by [WPT (test 16)](https://wpt.fyi/results/resize-observer/observe.html?run_id=6263261576560640&run_id=5091417146261504&run_id=5146116692115456&run_id=5116970087743488&run_id=5204317525508096) and works correctly in all browsers except Safari. It's fixed in the new Safari preview though.

I can't come up with anything better then:
1. Use some sort of timeout, which isn't ideal but even worse on Safari because it doesn't support `requestIdleCallback()` either.
2. Detect really small sizes beforehand and don't use oneshot `ResizeObserver`s and instead use `scrollWidth/Height` or `getBoundingClientRect()`.

Leaning more towards proposal 2. I will probably create a tracking issue for all the nonsense we have to work around because old devices can't update their Safari which requires us to continue to support old versions.

src/platform_impl/web/event_loop/window_target.rs
```diff
@@ -558,6 +525,33 @@ impl<T> EventLoopWindowTarget<T> {
...
+                }
+            },
+            {
+                let runner = self.runner.clone();
+
+                move |size| {
+                    RefCell::borrow(&canvas_clone).set_inner_size(size);
+                    runner.send_event(Event::WindowEvent {
+                        window_id: RootWindowId(id),
+                        event: WindowEvent::Resized(size),
```
daxpedda, line 549: Apparently changing the writing mode can cause the `ResizeObserver` to fire if `devicePixelContentBoxSize` is used even if the size hasn't changed.
So we actually have to check if the size changed before firing the event, which means we have to be careful about when we actually use `set_inner_size()`, otherwise we can't do the check properly.

mcclure:
Can I ask some questions about using this from an end-user perspective?

My usual way of interacting with Winit in browser is I create a window corresponding to a canvas (either full screen or in a layout) and then I draw to its entire extent with WebGL or WebGPU.

Naively, if the users uses CTRL-+ or CTRL-MINUS on desktop browser I would expect to continue to be told my width and height in actual screen pixels. But naively, if the user is on *mobile*, and they pinch to zoom in, I think I'd expect to be treated like an image and for my output to simply be rescaled. If in that circumstance you still "do the right thing" and continue to report my updating pixel size continuously, that's a nice surprise. I guess that means I can just keep presenting ever-higher resolution versions of my imagery. But it creates a problem.

If the user continues to zoom in on my window, very easily my window could wind up being several times the size of the screen. Mobile screens are already very high resolution. So resource costs to draw a frame could become very high, especially if I am doing multipass rendering (IE allocating an offscreen texture the size of the window— now potentially much larger than the screen), because I'm rendering lots of screen that is in fact offscreen.

Besides just tracking WindowEvent::Resized, is there any hope of being informed what the "visible area" of my window is in this scenario? (I'm not sure if this is something CSS/JS even expose.) That way I could draw only in the visible part of the screen. Alternately, can I get the screen size (I think browser tech calls this the "viewport") in winit, so I can cap out there?

Systemcluster:
> Besides just tracking WindowEvent::Resized, is there any hope of being informed what the "visible area" of my window is in this scenario? (I'm not sure if this is something CSS/JS even expose.) 

You can use [`IntersectionObserver`](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) for intersection with the viewport, but I think that is outside of the scope of this PR.

src/platform_impl/web/web_sys/resize_scaling.rs
daxpedda: > I can't come up with anything better then:
> 
>     1. Use some sort of timeout, which isn't ideal but even worse on Safari because it doesn't support `requestIdleCallback()` either.
> 
>     2. Detect really small sizes beforehand and don't use oneshot `ResizeObserver`s and instead use `scrollWidth/Height` or `getBoundingClientRect()`.

I remembered that I completely forgot about `getComputedStyle()` ... I still have to see how consistent it is with the other APIs, but that would be the perfect solution

> > You could probably also avoid having to create a temporary `ResizeObserver` here by calling `unobserve` then `observe` again on the canvas with the existing `ResizeObserver`, which will reset everything and fire another initial resize.
> 
> That's good to know! I will look into it.

I also realized that this could significantly simplify the implementation and I could also split resize and scale change again.

daxpedda:
> Naively, if the users uses CTRL-+ or CTRL-MINUS on desktop browser I would expect to continue to be told my width and height in actual screen pixels.

Yes, that's what this PR does.

> But naively, if the user is on _mobile_, and they pinch to zoom in, I think I'd expect to be treated like an image and for my output to simply be rescaled.

That is how it currently works in Winit already, pinch-to-zoom has no effect on Winit at all.

> If (as in this patch) in that circumstance you instead "do the right thing" and continue to report my updating pixel size continuously, that's a nice surprise— I guess that means I can just keep presenting ever-higher resolution versions of my imagery.

There isn't a reliable way to detect pinch-to-zoom on Web. You could detect when two pointers are moving away or towards each other and use a bunch of hacks to calculate the zoom factor, but this isn't something really supported or reliable.

> Besides just tracking WindowEvent::Resized, is there any hope of being informed what the "visible area" of my window is in this scenario? (I'm not sure if this is something CSS/JS even expose.) That way I could draw only in the visible part of the screen.

Winit has `WindowEvent::Occluded`, but that only tell you if it's completely occluded or not and isn't implemented on Web. Giving more information would require separate design work to cover other backends. Though getting this information will be complicated and inaccurate on Web.

> Alternately, can I get the screen size (I think browser tech calls this the "viewport") in winit, so I can cap out there?

I'm not sure what you mean with that, that's only useful if the canvas is bigger then the window, which sounds niche to me, but no, we don't have an API that returns that. In any case, sounds like an extended version of `Occluded` covers the use case here.

src/platform_impl/web/web_sys/resize_scaling.rs
daxpedda: Done!

> I also realized that this could significantly simplify the implementation and I could also split resize and scale change again.

Unfortunately this didn't help with that.

src/platform_impl/web/event_loop/window_target.rs
daxpedda: Done in https://github.com/rust-windowing/winit/pull/2859/commits/0eb1d3ece3f2b4fad8f0cb5285c377e433f08712.

daxpedda:
I had to redo the logic again to account for Safari not sending an initial resize event if the canvas has a size of zero.

The PR is ready again. Still waiting for a new `web-sys` release though.

Liamolucko:
> I guess css-size is just always on now?

Yep.

> - `window.canvas()` now returns an Option<HtmlCanvasElement> instead of an HtmlCanvasElement. I assume this is due to upgrading winit rather than the PR?

Yep, that was already the case before this PR.

src/platform_impl/web/web_sys/mod.rs
```diff
@@ -52,38 +52,51 @@ impl WindowExtWebSys for Window {
...
+/// or is not a CSS property that represents a size in pixel.
+pub fn style_size_property(style: &CssStyleDeclaration, property: &str) -> f64 {
+    let prop = style
+        .get_property_value(property)
+        .expect("Found invalid property");
+    prop.strip_suffix("px")
+        .expect("Element was not inserted into the DOM or is not a size in pixel")
+        .parse()
+        .expect("CSS property is not a size in pixel")
 }
```
Liamolucko, line 100: `getPropertyValue` doesn't always return a value in `px` - it _usually_ does, but in the special case where `display` is `none` or `contents` it doesn't.

I found this out in https://github.com/rust-windowing/winit/pull/2074/files#r961531983, the summary of which is that `border-*-width` properties always return values in `px`, but the `padding-*` properties might return percentages when `display` is `none` or `contents`.

I looked into how `width` and `height` work here as well, and they work the same way as `padding-*` - when `display` is `none` or `contents`, they might return percentages, or keywords like `min-content`. But otherwise they'll always return a value in `px`.

The solution to this in `ResizeScaleInternal::notify` is pretty simple: just return (0, 0), since that's what `ResizeObserver` reports for elements with `display: none` or `display: contents`. (side-note: I'm not sure if this is actually defined in the spec, which just says to return the element's content area. Content area's a property of the element's box, and I think elements with `display: none` don't have a box, so I'm not sure where that's supposed to come from.)

For `set_canvas_size` it's a bit trickier. We could just do nothing, but I think it'd make more sense to try and make it so that if/when `display` gets set back to something other than `none` or `contents`, it'll have that size. But we can't really do that properly without access to the size of the element's padding and borders.

That also opens up another problem - if the element's padding and/or borders are defined as relative sizes, they can change, meaning that the actual inner size of the canvas can drift from what it was set to if the window changes size.

One solution would be to also set `box-sizing: content-box`, and then not worry about padding and borders. The problem with that is that it also affects `min/max-width/height`, so if those are set it could mess with their meaning. One way to solve that would be to also apply `min-width: unset; max-width: unset; ...` to the canvas, since you could argue that `set_inner_size` is supposed to 100% guarantee that it'll have that size and capping it to a min/max size in CSS would conflict with that; but that might be reaching too far.

daxpedda: I'm a bit close to crying.

If we detect `display: none` (or `contents`, which resolves to `none` in the case of the canvas anyway), returning zero seems sensible to me.

> For set_canvas_size it's a bit trickier. We could just do nothing, but I think it'd make more sense to try and make it so that if/when `display` gets set back to something other than `none` or `contents`, it'll have that size.

We should do nothing in that case, `Window::set_inner_size()` should become `request_inner_size()` in the future hopefully anyway, see #2868.

> That also opens up another problem - if the element's padding and/or borders are defined as relative sizes, they can change, meaning that the actual inner size of the canvas can drift from what it was set to if the window changes size.

This is starting to get a bit overkill and we have to draw the line somewhere. As far as I decided in #2859, we won't be setting any CSS properties at all unless requested by `Window::set_inner_size()` (or by `WindowEvent::ScaleFactorChanged::new_inner_size`). Setting the window size is only a request, which we will fulfill to the best of our abilities (including calculating border and padding sizes), but the size we report back might be slightly different, e.g. a pixel off if we are using `devicePixelContentBoxSize`.

The thing we really care about is reporting `WindowEvent::Resized` if the physical size changed and report that as accurate as possible, which I believe is still intact. Even if the element's padding or borders are relative, `ResizeObserver` will still pick up any changes and report them to us.

---

Thank you for the careful review so far @Liamolucko, it was really really helpful.

daxpedda: I addressed this as explained above.

daxpedda:
With the latest `web-sys` release this is now ready to be merged. I will let this stew a little bit longer and see if we discover any issues or edge-cases.

I requested a review from @Liamolucko, if you want to review this please take your time, your input was invaluable so far :heart:.

Liamolucko (approved):
All looks good to me!

> Your input was invaluable so far, thank you ❤️.

No worries, I'm just happy to see this finally coming to fruition. Thank you also for all your work on this! ❤️

PR #2862, opened 2023-06-09
Fix for linux aarch64 where c_char is not i8

Imberflur:
- [x] Tested on all platforms changed (I only tested that `cargo c --target aarch64-unknown-linux-gnu` produces no errors)

### DIFF
--- a/src/platform_impl/linux/common/xkb_state.rs
+++ b/src/platform_impl/linux/common/xkb_state.rs
@@ -154,7 +154,7 @@ impl KbdState {
     /// `xkb_state_key_get_utf8`.
     fn make_string_with<F>(&mut self, mut f: F) -> Option<SmolStr>
     where
-        F: FnMut(*mut i8, usize) -> i32,
+        F: FnMut(*mut c_char, usize) -> i32,
     {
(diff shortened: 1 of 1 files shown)


src/platform_impl/linux/common/xkb_state.rs
```diff
@@ -154,7 +154,7 @@ impl KbdState {
     /// `xkb_state_key_get_utf8`.
     fn make_string_with<F>(&mut self, mut f: F) -> Option<SmolStr>
     where
-        F: FnMut(*mut i8, usize) -> i32,
+        F: FnMut(*mut core::ffi::c_char, usize) -> i32,
```
kchibisov, line 157: could you do full export at the top of the file?

kchibisov (changes requested):

src/platform_impl/linux/common/xkb_state.rs
Imberflur: Its already imported, I just didn't check lol :sweat_smile: 

Just changed this to `c_char`

kchibisov (approved):

PR #2869, opened 2023-06-10
Removed the pesky reference for WindowEvent::ScaleFactorChanged

atomicf32:
Disclaimer: I have no idea if it was important that there was a reference for new_inner_size, I just think it is pointless.

I have not tested on linux nor mac os, but it seems to work on windows.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- Made ScaleFactorChanged::new_inner_size not a reference, therefore allowing Event to no longer need a lifetime annotation.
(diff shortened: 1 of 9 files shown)


kchibisov:
See this https://github.com/rust-windowing/winit/pull/1456

--

There was a demand to have a way to adjust sizes during scale factor changes. I personally think that this event must be entirely removed in a favor of merging it with resize, but this reference is needed for now... I don't remember the exact discussion from where it came from, but it's not that hard to find on the issue tracker.

atomicf32:
@kchibisov Sorry if this is obvious, but couldn't they just call to the window's methods to resize it?

kchibisov:
@atomicf32 the methods are async.

PR #2871, opened 2023-06-11
On Web, implement `DeviceEvent`s

### LINKED ISSUE #2036: Most Device events do not work on the web.
I found that the [`DeviceEvent`s](https://docs.rs/winit/0.25.0/winit/event/enum.DeviceEvent.html) `Button`, `Key` and `MouseWheel` do not work in the browser. The only one that fires (as long as the mouse pointer is above the canvas) is `MouseMotion`.

I am not sure, if this is of any importance, but note that I create the `window` with:
```rust
        .with_title("Mouse Wheel events")
        .with_canvas(Some(canvas))
        .build(&event_loop)
        .unwrap();
```

I have a little repro of this over [here](https://github.com/FrankenApps/winit-wasm-events-repro). 

Just check the browser console and compare it with the console output when running on native.

daxpedda:
### Tracking

- [x] `MouseMotion`
- [x] `Motion`
- [x] `MouseWheel`
- [x] `Button`
- [x] `Key`
- [ ] `Text` (not implemented on any backend)

### Changes

This PR intends to implements all `DeviceEvent`s. Additionally `DeviceEvent`s are only emitted according to the setting in `EventLoopWindowTarget::listen_device_events()`.

---

Fixes #2036.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,7 +83,7 @@ And please only add new entries to the top of this list, right below the `# Unre
   the canvas size will be reported through `WindowEvent::Resized`.
 - On Web, respect `EventLoopWindowTarget::listen_device_events()` settings.
 - On Web, fix `DeviceEvent::MouseMotion` only being emitted for each canvas instead of the whole window.
-- On Web, add `DeviceEvent::Motion` and `DeviceEvent::MouseWheel` support.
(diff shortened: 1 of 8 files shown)


daxpedda:
Just hit this bug on Mozilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1753724.
So basically I broke `DeviceEvent::MouseMotion` in Firefox in #2840 and didn't notice.

Will figure out a fix.

PR #2873, opened 2023-06-11
Resume propagation of events

daxpedda:
I'm not entirely sure why `stopPropagation()` is being used, maybe because `preventDefault()` wasn't used back then. In any case, I don't think we want to prevent parents to receive events if they want to.

The call to `cancelBubble` doesn't do anything, it's a getter.

EDIT: See https://github.com/rust-windowing/winit/pull/2873#issuecomment-1586362882, I decided to keep `stopPropagation()` for now.

### DIFF
--- a/src/platform_impl/web/web_sys/canvas.rs
+++ b/src/platform_impl/web/web_sys/canvas.rs
@@ -378,16 +378,10 @@ impl Common {
         E: 'static + AsRef<web_sys::Event> + wasm_bindgen::convert::FromWasmAbi,
         F: 'static + FnMut(E),
     {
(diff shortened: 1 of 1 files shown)


daxpedda:
> It's also blocking `DeviceEvent`s from working correctly, you stop receiving them when the events are triggered in the canvas. See #2871.

I actually had to fix it in a different way. The web backend can't send events from multiple handlers at the same time, it has to restart the event loop every time. So I'm trying as much as possible to coalesce as many events as possible.

So for now `stopPropagation()` is actually helping #2871, otherwise I would have to check if the mouse is over the canvas or not (which shouldn't be too hard). We can eventually remove it when somebody actually complains.

Issue #1228, opened 2019-10-17
Allow changing X11 DPI at runtime


chrisduerr:
This has been reported to Alacritty in https://github.com/jwilm/alacritty/issues/2886.

It seems like currently winit does not allow to modify the DPI at runtime, which I can definitely see as desirable for some users. In Alacritty it would be possible to just change the font size, but I feel like it would still be worth it to just change one thing and the whole system properly rescales.

frebib:
Looks related to https://github.com/rust-windowing/winit/issues/1377

From testing with alacritty, I can't seem to invoke any `ScaleFactorChanged` events, presumably due to the above issue (reproduce with `alacritty --print-events 2>&1 | grep -ie scale` and change monitor DPI or move window to another display with different DPI).

At least in X11, I'm leveraging the `xsettings` behaviour built into GTK to resize at runtime. https://www.freedesktop.org/wiki/Specifications/XSettingsRegistry/
It's definitely not the most scalable solution, but it is at least one source of `dpi changed` events

frebib:
I spent some time digging into this. I managed to cobble together something that works, but is definitely sub-optimal. This is the first Rust I've ever "written" so even getting this to work was a challenge. It uses the existing code to reload monitors and trigger `ScaleFactorChanged` events which is completely overkill for the little work that is actually required. I think this needs restructuring such that `monitor`s can update their own DPI and propagate the events down to the windows, or something like that. I didn't even want to attempt to work out the lifetime/ownership headaches around that. If anyone feels inclined, or motivated to, I'm happy for this to be adapted in a cleaner way so it can be merged. That being said, I'll be using this patch until something better comes along as it will probably work well enough for my changing-dpi needs.

https://github.com/frebib/winit/commit/74d5070c91ff8150c7630dd3b80d66a14cbe5fb6

rhelmot:
I'm interested in working on this, but I have no familiarity with winit code. Could someone on the winit team give me a high-level explanation of what's unacceptable about the above patch so I can start hacking on it?

frebib:
Bump? Anyone able to help out with this patch? I've been using https://github.com/frebib/winit/commit/74d5070c91ff8150c7630dd3b80d66a14cbe5fb6 for the last 2 years with zero issues, although it seems to stop working after https://github.com/rust-windowing/winit/commit/c916eb613734b389d7add7efec7f698ca739bae2

frebib:
[...]

PR #2874, opened 2023-06-11
x11: reload dpi on PropertyChangeEvent

frebib:
This change allows X11 windows to receive ScaleFactorChanged events when using Xft.dpi via the root window PropertyChange event.

Subscribe to PropertyChange events on the root window, and reload all available monitors when an XA_RESOURCE_MANAGER change is received. This is a very heavy handed approach to this problem, but it was the least intrustive way I could find to make it work with my limited skill. Load XResources directly via XGetTextProperty on every get_xft_dpi call as XResourceManagerString returns stale cached data and does not reload for the lifetime of the window (?)

Fixes: https://github.com/rust-windowing/winit/issues/1228

---

Full disclosure- I have no idea what I'm doing. This has worked for me for the past few years with no issues, but the code is probably bad. I've never written Rust in my life, but I figured this was a starting point

### DIFF
--- a/src/platform_impl/linux/x11/event_processor.rs
+++ b/src/platform_impl/linux/x11/event_processor.rs
@@ -1102,6 +1102,17 @@ impl<T: 'static> EventProcessor<T> {
                     _ => {}
                 }
             }
+
+            ffi::PropertyNotify => {
+                let xev: &ffi::XPropertyEvent = xev.as_ref();
+
+                // XA_RESOURCE_MANAGER received from the root window indicates that xresources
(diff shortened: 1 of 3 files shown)


notgull:
I can rewrite this for the new XCB backend if needed.

frebib:
@notgull that would be amazing if you could 😍 this was only ever a hack anyway

kchibisov:
@notgull beep

notgull:
My bad, I forgot to close this after I merged #3222

PR #2879, opened 2023-06-13
On Web, send modifiers on key events first

daxpedda:
I just reordered the events on key presses and releases, `ModifiersChanged` is sent first.
This is consistent with pointer events.


kchibisov:
The ordering wrt modifiers for key events was explicitly like that, because that's how it should be, otherwise you can't encode `Shift` with `Shift` modifier, known as double shift reliably.

PR #1519, opened 2020-03-27
Implement `requestAnimationFrame` for web

chemicstry:
- [x] Tested on all platforms changed
- [x] Compilation warnings were addressed
- [x] `cargo fmt` has been run on this branch
- [x] `cargo doc` builds successfully
- [x] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users
- [x] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior
- [ ] Created or updated an example program if it would help users understand this functionality
- [ ] Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented

Implementation based on discussion in #1493. I did not implement `requestPostAnimationFrame` because it's not documented properly yet and does not exist in `stdweb` or `web_sys` either.

Issue #1305, opened 2019-12-03
Web backend sends RedrawRequested too often


ryanisaacg:
An expected user pattern with the new API is to request redraw after every `EventsCleared`.

On web, the event handlers are run instantly as events come in, which means there can be many more `EventsCleared` than the user expects.

I can think of two ways to address this:
- Don't run the event handlers as often, and buffer up events with a minimum time between processing them. This adds latency where it isn't necessarily needed, so it may not be ideal.
- Only run `RedrawRequested` events on a timer determined by the browser `requestAnimationFrame` API, which does slightly break the API contract.

I'm not sure which of these mitigations is better, so I would welcome discussion.

Osspial:
Why is this a problem? This behavior is also exhibited on Windows since there's pretty much no overhead between the platform event dispatcher and the Winit event closure, but I haven't found it to be an issue because real applications take long enough to process events (due to vsync, rendering, world updates, etc.) that events naturally get buffered up. Even if downstream users have unusually lightweight applications that can reach extreme framerates, that shouldn't pose any functional issues in their applications.

ryanisaacg:
My understanding from a few user reports is that the redraws 'build up' over time, consistently producing more work than is handled.

goddessfreya:
Won't redraw buildup be resolved by `redraw-requested-v2`?

ryanisaacg:
It might be; once it lands on master, we can see if that fixes the problem.

Osspial:
Do web platforms automatically aggregate duplicate `RedrawRequested` events into a single event?

ryanisaacg:
Yes, if you request redraw multiple times before another `RedrawRequested` it should only produce one event

tangmi:
@ryanisaacg I agree that `RedrawRequested` should only fire when `requestAnimationFrame` runs, but not necessarily every frame (so that non-game applications that don't change often can save battery life/cpu cycles). What do you think of a flag that producers of `RedrawRequested` can set/check to kick off a single `requestAnimationFrame` that is cleared when the `requestAnimationFrame` callback finishes? This would effectively debounce `RedrawRequested` events and let the browser  service the "request" when it best thinks it should. A downside of this is that the visuals could always be a frame behind (but I think that might always be the case if visuals are always handled in `requestAnimationFrame`'s callback...?)

OvermindDL1:
> (but I think that might always be the case if visuals are always handled in requestAnimationFrame's callback...?)

[...]

Issue #2863, opened 2023-06-09
Let `Window::inner_size()` return `None` if the Window isn't "ready" yet


daxpedda:
I've heard this come up many times before, but when using Winit to render stuff, users often have to check for the window size not to be zero.

I've encountered this before in X11, but I don't know the story behind this on other backends. For Wasm, it's easy to hit this problem at application startup, where the canvas isn't even inserted yet into the DOM, where it makes perfect sense to return `None` in `Window::inner_size()`. See https://github.com/rust-windowing/winit/pull/2859#discussion_r1222452529 which spawned this issue.

kchibisov:
This is more general issue of async window startup, we have similar problem on Wayland, which is solved right now by dispatching event loop and waiting for the events to complete.

@rib do you happen to remember what the conclusion for stuff like that was? I think we need a way for window to be created async.

My initial idea would be to have normal constructor for that which will return not a window, but `WindowId`, as it'll reserve the slot for window in the backends internal map with unique ids, so the user will have a target for events and a place in the hash map, and the actual window will be received as part of some event so they insert into vacant slot in the map/whatever they use?

mcclure:
Just to note, I filed [a dupe of this bug](https://github.com/rust-windowing/winit/issues/2967) in 2023, and I am still experiencing the problem (in the form of [my program often being unable to run in wasm if I don't include protective code for the nonsense case of a 0,0 window](https://data.runhello.com/j/wgpu/special/stagingbelt/4/)) as of 0.30.4.

PR #2880, opened 2023-06-13
Use `Window.requestIdleCallback()`

daxpedda:
This PR replaces the use of `Window.requestAnimationFrame` with `Window.requestIdleCallback`, which is in line with what `ControlFlow::Poll` is supposed to be:
> When the current loop iteration finishes, immediately begin a new iteration regardless of whether or not new events are available to process.

This is going to make `ControlFlow::Poll` very fast, which is much closer to how other backends work, but it is also gonna consume way more power, which may or may not be desired considering this is Web and running in the browser.

I think a lot of people are using `RedrawRequested` as a VSync source, which it is not as far as I understand. See #2412 for that.

Reverts #1519.
Related #1305.

### DIFF
--- a/src/platform_impl/web/web_sys/timeout.rs
+++ b/src/platform_impl/web/web_sys/timeout.rs
@@ -1,8 +1,11 @@
+use once_cell::unsync::OnceCell;
 use std::cell::Cell;
 use std::rc::Rc;
 use std::time::Duration;
 use wasm_bindgen::closure::Closure;
+use wasm_bindgen::prelude::wasm_bindgen;
 use wasm_bindgen::JsCast;
+use wasm_bindgen::JsValue;
 
 #[derive(Debug)]
 pub struct Timeout {
@@ -40,16 +43,21 @@ impl Drop for Timeout {
 }
 
 #[derive(Debug)]
-pub struct AnimationFrameRequest {
+pub struct IdleCallback {
     window: web_sys::Window,
-    handle: i32,
-    // track callback state, because `cancelAnimationFrame` is slow
+    handle: Handle,
     fired: Rc<Cell<bool>>,
     _closure: Closure<dyn FnMut()>,
 }
 
-impl AnimationFrameRequest {
-    pub fn new<F>(window: web_sys::Window, mut f: F) -> AnimationFrameRequest
+#[derive(Clone, Copy, Debug)]
+enum Handle {
+    IdleCallback(u32),
+    Timeout(i32),
+}
+
+impl IdleCallback {
+    pub fn new<F>(window: web_sys::Window, mut f: F) -> IdleCallback
     where
         F: 'static + FnMut(),
     {
@@ -60,11 +68,21 @@ impl AnimationFrameRequest {
             f();
         }) as Box<dyn FnMut()>);
 
-        let handle = window
-            .request_animation_frame(closure.as_ref().unchecked_ref())
-            .expect("Failed to request animation frame");
+        let handle = if has_idle_callback_support(&window) {
+            Handle::IdleCallback(
+                window
+                    .request_idle_callback(closure.as_ref().unchecked_ref())
+                    .expect("Failed to request idle callback"),
+            )
+        } else {
+            Handle::Timeout(
+                window
+                    .set_timeout_with_callback(closure.as_ref().unchecked_ref())
+                    .expect("Failed to set timeout"),
+            )
+        };
 
-        AnimationFrameRequest {
+        IdleCallback {
             window,
             handle,
             fired,
@@ -73,12 +91,34 @@ impl AnimationFrameRequest {
     }
 }
 
-impl Drop for AnimationFrameRequest {
+impl Drop for IdleCallback {
     fn drop(&mut self) {
         if !(*self.fired).get() {
-            self.window
-                .cancel_animation_frame(self.handle)
-                .expect("Failed to cancel animation frame");
+            match self.handle {
+                Handle::IdleCallback(handle) => self.window.cancel_idle_callback(handle),
+                Handle::Timeout(handle) => self.window.clear_timeout_with_handle(handle),
+            }
         }
     }
+}
+
+fn has_idle_callback_support(window: &web_sys::Window) -> bool {
+    thread_local! {
+        static IDLE_CALLBACK_SUPPORT: OnceCell<bool> = OnceCell::new();
+    }
+
+    IDLE_CALLBACK_SUPPORT.with(|support| {
+        *support.get_or_init(|| {
+            #[wasm_bindgen]
+            extern "C" {
+                type IdleCallbackSupport;
+
+                #[wasm_bindgen(method, getter, js_name = requestIdleCallback)]
+                fn has_request_idle_callback(this: &IdleCallbackSupport) -> JsValue;
+            }
+
+            let support: &IdleCallbackSupport = window.unchecked_ref();
+            !support.has_request_idle_callback().is_undefined()
+        })
+    })
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On Web, fix the bfcache by not using the `beforeunload` event.
 - On Web, fix scale factor resize suggestion always overwriting the canvas size.
 - On macOS, fix crash when dropping `Window`.
+- On Web, use `Window.requestIdleCallback()` for `ControlFlow::Poll` when available.
 
 # 0.28.6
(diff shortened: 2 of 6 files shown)


src/platform_impl/web/web_sys/timeout.rs
```diff
@@ -73,12 +91,34 @@ impl AnimationFrameRequest {
...
+                Handle::Timeout(handle) => self.window.clear_timeout_with_handle(handle),
+            }
         }
     }
 }
+
+fn has_idle_callback_support(window: &web_sys::Window) -> bool {
+    thread_local! {
+        static IDLE_CALLBACK_SUPPORT: OnceCell<bool> = OnceCell::new();
+    }
```
notgull, line 108: The thread local here might be more expensive than just an atomic variable.

notgull (commented):
This kind of thing might want to be configurable. As in, in the event loop constructor for web, add an option for whether or not to use this or the animation frame.

src/platform_impl/web/web_sys/timeout.rs
daxpedda: What, why? I've been using this a lot until now.

daxpedda:
> This kind of thing might want to be configurable. As in, in the event loop constructor for web, add an option for whether or not to use this or the animation frame.

I would rather not, if I understood the API correctly so far, that would push users towards using `RedrawRequest` as a VSync source, which not only is wrong, but it doesn't actually work, because any registered event will wake up the event loop anyway.

kchibisov:
>I would rather not, that would push users towards using RedrawRequest as a VSync source, which not only is wrong, but it doesn't actually work, because any registered event will wake up the event loop anyway.

Yeah, the vsync source is `requestAnimationFrame`, which doesn't map to any present event we have in winit yet, but the frame hint mentioned on #2412 should be it.

src/platform_impl/web/web_sys/timeout.rs
daxpedda: I learned a lot going down this rabbit hole, things I honestly didn't want to know.

In any case, in this case `sync::OnceCell` is more expensive then the `unsync` version.

cybersoulK:
@daxpedda 

The current master i have error in console:

panicked at 'Error in Surface::configure: Validation Error

Caused by:
    Both `Surface` width and height must be non-zero. Wait to recreate the `Surface` until the window has non-zero area.
    
 Might be because of requestIdleCallback starts before the canvas is spawned.
 My canvas is spawned by winit itself, and not manually.

daxpedda:
You always have to check if the window size isn't zero yourself, Winit doesn't offer any guarantees in this regard, this is an issue with other backends as well.

We are tracking solutions for this here: https://github.com/rust-windowing/winit/issues/2863. But it will require some further design work.

GloopShlugger:
I think this change causes the loop to pause due to inactivity (or spuriously) on chrome based browsers, or at least the game loop keeps stopping until there's input / or randomly when embedded in itch.io for example

kchibisov:
I think that's expected, you should use `request_redraw` if you want to wake up. This will happen as well other platforms.

However which control flow mode are you using? Is it `Poll`? @daxpedda can we have some sort of busy loop with `Poll`?

GloopShlugger:
i am using poll and i tried requesting redraws but that did not change anything -- it is for a typical gameloop

kchibisov:
Well, you'd have to wait for https://github.com/rust-windowing/winit/pull/2896 merge, so the `requestAnimationFrame` will be used under the hood.

GloopShlugger:
i see thank you, for now setting controlflow to wait for a zero duration seems to do it

daxpedda:
> I think this change causes the loop to pause due to inactivity (or spuriously) on chrome based browsers, or at least the game loop keeps stopping until there's input / or randomly when embedded in itch.io for example

`Window.requestIdleCallback()` should pretty much be as fast as possible.
FYI there isn't really an alternative, we were using `Window.requestAnimationFrame()` beforehand which is definitely slower and you can't just have a busy loop on the main thread as that would just block the whole page.

@GloopShlugger how long are these pauses? Do you have to do anything to induce them? I can't seem to reproduce this on Chrome nor Firefox.

GloopShlugger:
i have found it to only happen on chrome-based browsers embedded inside an iframe like itch.io, i have also seen it happen with just the dev console open. The way i detected it is by printing every frame and after a while (which has not been a consistent amount) it just stops updating, with or without input. It also on startup seems to run the eventloop once, and then wait for input again, but it has to be specific events like mouse clicks, just moving it sometimes also seems to trigger a rerun, at which point it works again for a bit until it just stops.

I feel like this is 10000% a browser implementation choice like some sort of powersaving measure chrome does, i may be able to provide an itch.io build if required

daxpedda:
That is unfortunate indeed. I always knew `Window.requestIdleCallback()` isn't ideal, but it's turning out to be more of a problem then I thought it would be.

The proper solution is to use the [Prioritized Task Scheduling API](https://wicg.github.io/scheduling-apis), unfortunately this isn't available in Firefox/Safari yet.

I had another idea that might fix this issue:
The problem with `setTimeout()` is that if you call it from another nested `setTimeout()` callback, the there will be a minimum delay of 4ms. We could get around this by queuing the next `setTimeout()` not from inside the callback, but from a [`queueMicroTask()`](https://developer.mozilla.org/en-US/docs/Web/API/queueMicrotask).

I will make a PR and ping you.

GloopShlugger:
interesting i'll definitelly try it, i did end up changing the wait to be reasonable like 5ms in the loop. I do think it would likely end up being fine once / if the update and draw split happens. The other easy short term solution is just forking winit and changing it back to requestanimationframe which isn't that big of a change

daxpedda:
In https://github.com/rust-windowing/winit/pull/2896 we changed `Window::request_redraw()` to use `Window.requestAnimationFrame()`, so if you end up forking it you probably want to change it to `setTimeout()` instead.
But you could just use `ControlFlow::WaitUntil(Instant::now())`, which would end up being the same.

GloopShlugger:
oh my god it landed and i didn't even notice i'll try that first and see if it changes anything

GloopShlugger:
I tried out the newest master and with `controlflow::poll` and requesting redraws both in `AboutToWait` and `RedrawRequested` still get stuck on chrome, but if you use controlflow::wait and `RedrawRequested` it does vsync correctly now

a link with the issue:
https://bongli.itch.io/aldi-ball?secret=tuSVDBz7N4PbBwsr2dYghVupWJA

you can see firefox just runs it immediatly but chrome based browsers get stuck at first it's very random when and if it does get stuck again but it happens even during input
opening the devtools does seem to make it happen more often

GloopShlugger:
@daxpedda it does get stuck immediately and the first frame does not get drawn on chrome until you press something. `request_redraw` too gets stopped after a short amount of time. (easiest to trigger with developer console opened) 
this only happens with set_pool, it will be interesting to see if moving away from `RequestIdleCallback`, I'm very unfamiliar with the winit codebase so i sadly can't help much. But the issue vanishes completely if you use a wait.
Thank you for looking into it so much

daxpedda:
> it does get stuck immediately and the first frame does not get drawn on chrome until you press something. `request_redraw` too gets stopped after a short amount of time.

Are you sure that `Window::request_redraw()` is actually called? It might be that the event you call it from is being throttled.
Basically if you call `Window.request_redraw()` you should always get the `RequestRedraw` event, unless `Window.requestAnimationFrame` is being throttled which I don't believe, so there would be some deeper Winit bug in that case.

GloopShlugger:
I'm calling it from inside the redraw requested event, it works perfectly fine in all other cases, there was no difference between about to wait and that, or both, the event loop just seemingly stops

daxpedda:
Would you mind running a simple JS script in the background to figure out whether this is really a Winit bug or not?
```js
const fun = () => {
    console.log('rAF callback')
    requestAnimationFrame(fun)
}
requestAnimationFrame(fun)
```

GloopShlugger:
when this runs in the background the issue is fixed :,)

daxpedda:
???
I guess this is a bug in Chrome then?
Or maybe it's throttling algorithm is just very weird.

GloopShlugger:
i will try the PR the moment you ping me and i see it, I hope there's some solution that works with winits model, thank you again for looking into it -- dancing in the pits of web api hell

daxpedda:
I spent a lot of time going through the implementation, but I think there is just something weird going with Chrome, I can't find anything wrong with Winit.

If you ever figure out a minimal reproducable example please hit me up, it's hard to debug this without being able to try it.

In any case, see #3044.

PR #2881, opened 2023-06-13
Revert "On Web, send modifiers on key events first"

daxpedda:
Reverts rust-windowing/winit#2879.
Pressing a Shift key and receiving a Shift key modifier beforehand is strange, so it was right the first time.
This aligns with Wayland as well.


kchibisov (approved):
I think most backends in winit do so, because we explicitly changed that. I'm not sure that it's documented or held good internally.

PR #2883, opened 2023-06-16
Add `FrameThrottled` event and a way to request it

### LINKED ISSUE #2412: Implement VSync source

kchibisov:
Add a way to request a frame throttling hint from the windowing system. This is essential for platforms like Wayland where you don't have real vsync in OpenGL/Vulkan and you're advised to use 'frame callbacks' to throttle your rendering.

`Window::request_frame_throttling_hint` is a an end point for users to request `FrameThrottled` event, which returns error when it can't be done, so the users could fallback to their own scheduling strategies.

The API doesn't try to address all sort of frame presentation cases, like how much the frame took, etc. Such cases could be covered later, with different events in addition to the throttling hint.

For now implemented only for Wayland.

Fixes #2412.

--

cc @daxpedda for web.

### DIFF
--- a/src/window.rs
+++ b/src/window.rs
@@ -1127,6 +1127,26 @@ impl Window {
         self.window.request_user_attention(request_type)
     }
 
+    /// Request a frame throttling hint from the windowing system.
+    ///
+    /// The error is returned when frame couldn't be requested and the user should use their
+    /// fallback renderer scheduling logic, like using timer based on monitor refresh rate or
+    /// graphics API extensions (VSYNC).
+    ///
+    /// For more see [`FrameThrottled`] event.
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **Wayland:**  Uses frame callbacks. The user must perform drawing operation resulting in
+    ///                 `wl_surface.commit`(eglSwapBuffers, etc) after issueing a request.
+    /// - ** macOS / Windows / iOS / Android / Web / X11 / Orbital:** Not supported.
+    ///
+    /// [`FrameThrottled`]: crate::event::WindowEvent::FrameThrottled.
+    #[inline]
+    pub fn request_frame_throttling_hint(&self) -> Result<(), NotSupportedError> {
+        self.window.request_frame_throttling_hint()
+    }
+
     /// Sets the current window theme. Use `None` to fallback to system default.
     ///
     /// ## Platform-specific
--- a/src/window.rs
+++ b/src/window.rs
@@ -1137,8 +1137,8 @@ impl Window {
     ///
     /// ## Platform-specific
     ///
-    /// - **Wayland:**  Uses frame callbacks. The user must perform drawing operation resulting in
-    ///                 `wl_surface.commit`(eglSwapBuffers, etc) after issueing a request.
+    /// - **Wayland:**  `wl_surface.commit` must be called for this event to fire.
+    ///                 This is usually done by drawing operations, e.g. `eglSwapBuffers`.
     /// - ** macOS / Windows / iOS / Android / X11 / Orbital:** Not supported.
     ///
     /// [`FrameThrottled`]: crate::event::WindowEvent::FrameThrottled.
--- a/src/window.rs
+++ b/src/window.rs
@@ -1139,7 +1139,7 @@ impl Window {
     ///
     /// - **Wayland:**  Uses frame callbacks. The user must perform drawing operation resulting in
     ///                 `wl_surface.commit`(eglSwapBuffers, etc) after issueing a request.
-    /// - ** macOS / Windows / iOS / Android / Web / X11 / Orbital:** Not supported.
+    /// - ** macOS / Windows / iOS / Android / X11 / Orbital:** Not supported.
     ///
     /// [`FrameThrottled`]: crate::event::WindowEvent::FrameThrottled.
     #[inline]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- **Breaking:** Add `WindowEvent::FrameThrottled` event and `Window::request_frame_throttling_hint` to request it.
 - On Android, changed default behavior of Android to ignore volume keys letting the operating system handle them.
 - On Android, added `EventLoopBuilderExtAndroid::handle_volume_keys` to indicate that the application will handle the volume keys manually.
 - **Breaking:** Rename `DeviceEventFilter` to `DeviceEvents` reversing the behavior of variants.
--- a/src/event.rs
+++ b/src/event.rs
@@ -366,6 +366,14 @@ pub enum WindowEvent<'a> {
     /// hovered.
     HoveredFileCancelled,
 
+    /// An event from windowing system when it is a good time to start drawing a new frame.
+    ///
+    /// This is useful for throttling redrawing operation and driving animations. The event
+    /// is delivered in reaction to [`Window::request_frame_throttling_hint`].
+    ///
+    /// For platform specifics see [`Window::request_frame_throttling_hint`] request.
+    FrameThrottled,
+
     /// The window gained or lost focus.
     ///
     /// The parameter is true if the window has gained focus, and false if it has lost focus.
@@ -553,6 +561,7 @@ impl Clone for WindowEvent<'static> {
             Moved(pos) => Moved(*pos),
             CloseRequested => CloseRequested,
             Destroyed => Destroyed,
+            FrameThrottled => FrameThrottled,
             DroppedFile(file) => DroppedFile(file.clone()),
             HoveredFile(file) => HoveredFile(file.clone()),
             HoveredFileCancelled => HoveredFileCancelled,
@@ -656,6 +665,7 @@ impl<'a> WindowEvent<'a> {
             Moved(position) => Some(Moved(position)),
             CloseRequested => Some(CloseRequested),
             Destroyed => Some(Destroyed),
+            FrameThrottled => Some(FrameThrottled),
             DroppedFile(file) => Some(DroppedFile(file)),
             HoveredFile(file) => Some(HoveredFile(file)),
             HoveredFileCancelled => Some(HoveredFileCancelled),
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -808,6 +808,10 @@ impl Window {
         self.redraw_requester.request_redraw()
     }
 
+    pub fn request_frame_throttling_hint(&self) -> Result<(), error::NotSupportedError> {
+        Err(error::NotSupportedError::new())
+    }
+
     pub fn inner_position(&self) -> Result<PhysicalPosition<i32>, error::NotSupportedError> {
         Err(error::NotSupportedError::new())
     }
--- a/src/platform_impl/ios/window.rs
+++ b/src/platform_impl/ios/window.rs
@@ -75,6 +75,10 @@ impl Inner {
         }
     }
 
+    pub fn request_frame_throttling_hint(&self) -> Result<(), NotSupportedError> {
+        Err(NotSupportedError::new())
+    }
+
     pub fn inner_position(&self) -> Result<PhysicalPosition<i32>, NotSupportedError> {
         unsafe {
             let safe_area = self.safe_area_screen_space();
--- a/src/platform_impl/linux/mod.rs
+++ b/src/platform_impl/linux/mod.rs
@@ -541,12 +541,11 @@ impl Window {
         }
     }
     pub fn request_user_attention(&self, request_type: Option<UserAttentionType>) {
-        match self {
-            #[cfg(x11_platform)]
-            Window::X(ref w) => w.request_user_attention(request_type),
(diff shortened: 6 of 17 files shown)


kchibisov:
I'm not sure that I like this naming, but I want to make it clear that it's about throttling and not a `Vblank` timing in particular, given that minimized windows on Web/Wayland might not ever get this event.

daxpedda (approved):
Web is done.

src/window.rs
```diff
@@ -1127,6 +1127,26 @@ impl Window {
...
+    /// The error is returned when frame couldn't be requested and the user should use their
+    /// fallback renderer scheduling logic, like using timer based on monitor refresh rate or
+    /// graphics API extensions (VSYNC).
+    ///
+    /// For more see [`FrameThrottled`] event.
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **Wayland:**  Uses frame callbacks. The user must perform drawing operation resulting in
+    ///                 `wl_surface.commit`(eglSwapBuffers, etc) after issueing a request.
```
kchibisov, line 1141: @daxpedda does web needs a drawing operation or something like that afterwards like Wayland? Maybe we should state web specifics as well here, like what is used and how it's done?

daxpedda: > does web needs a drawing operation or something like that afterwards like Wayland?

No, nothing needs to be done for Web. If you don't draw everything will continue like before, the current buffer won't be invalidated or anything like that.

> Maybe we should state web specifics as well here, like what is used and how it's done?

I think I would prefer to even remove it for Wayland unless there is something that **needs** to be known by the user. "how it's done" should remain an implementation detail. We haven't done it anywhere else as well unless really necessary.

kchibisov: The thing is that the callback won't be requested if the user doesn't commit, commit is done implicitly by swap buffers, vulkan stuff, or softbuffer. I'm just trying to communicate that commit is not done by the winit itself, because doing so could cause glitches.

daxpedda: That does sounds like something the user needs to know indeed.

By the way, this isn't what I understood from that comment. To me it reads like you must perform a commit, when you request the frame, but it's not specified what happens if you don't. But from what you just wrote, I am understanding that if you don't commit you won't get a `FrameThrottled` event.

```suggestion
    /// - **Wayland:**  `wl_surface.commit` must be called for this event to fire.
    ///                 This is usually done by drawing operations, e.g. `eglSwapBuffers`.
```
Something like that? I have no clue about the terminology btw.

kchibisov: If you don't `commit` you won't get a callback. It'll just be in a pending state, because the surface states are applied atomically on commit.

rib:
We already have `RedrawRequested` events which could be used for this.

Currently `RedrawRequested` is emitted inconsistently and that's a problem in itself which would be good to address.

This is already problematic because on macOS we emit RedrawRequested via drawRect _and_ after MainEventsCleared and applications just don't really know when they should render.

I would really like us to emit `RedrawRequested` something like this:

1. On Web RedrawRequested can be driven by requestAnimationFrame
2. On Wayland RedrawRequested can be driven by frame callbacks
3. On macOS RedrawRequested can be driven by drawRect
4. On Windows RedrawRequested can be driven by WM_PAINT

There shouldn't be any need to opt into frame throttling and this shouldn't require any new API, we can instead address the backend inconsistencies we have so that we can rely on the existing API.

kchibisov:
>Currently RedrawRequested is emitted inconsistently and that's a problem in itself which would be good to address.

The issue is that some clients don't want to draw once they got the throttling event, like alacritty, for example, when nothing changes. However `RedrawRequested` is used to indicate that OS really wants a buffer from you back, so regardless of whatever state you have you must submit a buffer back. The `FrameThrottled` event could throttle the Window state updates, so we get perfect frames, but we can't replace one with another without a rework of the `RedrawRequested`.

rib:
> > Currently RedrawRequested is emitted inconsistently and that's a problem in itself which would be good to address.
> 
> The issue is that some clients don't want to draw once they got the throttling event, like alacritty, for example, when nothing changes. However `RedrawRequested` is used to indicate that OS really wants a buffer from you back, so regardless of whatever state you have you must submit a buffer back. The `FrameThrottled` event could throttle the Window state updates, so we get perfect frames, but we can't replace one with another without a rework of the `RedrawRequested`.

I'd say those semantics for Winit are too poorly defined / inconsistent right now and there is wiggle room to document clear semantics as part of any work to fix how RedrawRequested is emitted consistently..

In terms of the window system being able to demand that you redraw one of the things that's missing for that is being able to report expose rectangles. If the window system really requires a window to be redrawn then that'll likely either be because it's a legacy system, like X11 non-composited responding to Expose events, or e.g. on some systems because the window got resized.

Those cases that _require_ the window to be redrawn should be honored by alacritty.

Those special cases are the exception not the rule for when RedrawRequested should be emitted though.

Most of the time RedrawRequested would follow a redraw_request for a Window.

Maybe I should have been clearer about that in the illustration.

I didn't meant to say that `RedrawRequested` should happen _unconditionally_ / continuously for each of those per-OS mechanism. RedrawRequested should happen after the application has requested a redraw but e.g. triggered via requestAnimationFrame on Web or a frame callback on Wayland etc.

There may be times that it's triggered for the (exceptional) window-system specific cases that you're referring too but normally it would only trigger if alacritty specifically requested it - so either way alacritty should be fine to redraw whenever it fires and it can still handle incremental redrawing correctly. If nothing has changed alacritty wouldn't request a redraw.

kchibisov:
>In terms of the window system being able to demand that you redraw one of the things that's missing for that is being able to report expose rectangles. If the window system really requires a window to be redrawn then that'll likely either be because it's a legacy system, like X11 non-composited responding to Expose events, or e.g. on some systems because the window got resized.

There's a protocol in works for Wayland to request a new buffer to handle GPU resents, so the compositor could re upload stuff. Or, for example, when client side decorations change the state, and they require a commit from the primary surface, so the sync frame perfect rendering could be achieved. In this case `RedrawRequested` is required, though, it could still be throttled by the `frame` callbacks.

>Those cases that require the window to be redrawn should be honored by alacritty


That's true, we do request a redraw, but sometimes we winit asks us to redraw to sync with its internal stuff, like subsurfaces.

>Most of the time RedrawRequested would follow a redraw_request for a Window

That's true, but the event has established certain behavior, we can't really change the semantics, because it might not be followed.

>I didn't meant to say that RedrawRequested should happen unconditionally / continuously for each of those per-OS mechanism. RedrawRequested should happen after the application has requested a redraw but e.g. triggered via requestAnimationFrame on Web or a frame callback on Wayland etc.

>There may be times that it's triggered for the (exceptional) window-system specific cases that you're referring too but normally it would only trigger if alacritty specifically requested it - so either way alacritty should be fine to redraw whenever it fires and it can still handle incremental redrawing correctly. If nothing has changed alacritty wouldn't request a redraw.


That's good, but how we'd handle macOS/Windows now? How the event should behave on them? Should we throttle by the timer, how should we indicate the fallback? What should we do about sync requests, like resize on macOS? I agree that it's messy right now, but I don't see how we can merge them without removing `RedrawRequsted` entirely. If you have a suggestion how it should work I'd like to hear it.

rib:
> > In terms of the window system being able to demand that you redraw one of the things that's missing for that is being able to report expose rectangles. If the window system really requires a window to be redrawn then that'll likely either be because it's a legacy system, like X11 non-composited responding to Expose events, or e.g. on some systems because the window got resized.
> 
> There's a protocol in works for Wayland to request a new buffer to handle GPU resents, so the compositor could re upload stuff. Or, for example, when client side decorations change the state, and they require a commit from the primary surface, so the sync frame perfect rendering could be achieved. In this case `RedrawRequested` is required, though, it could still be throttled by the `frame` callbacks.

yeah, that's ok - I'm not saying there aren't reasons for the window system to request redraws, there are.

In my view though there's no reason for these to require a completely separate mechansim for notifying the app when to redraw.

The app should be able to drive redraws off of one event, and it's a separate detail that sometimes that will be triggered for window system specific reasons and most of the time it will just be triggered because the app requested it.

kchibisov:
>The app should be able to drive redraws off of one event, and it's a separate detail that sometimes that will be triggered for window system specific reasons and most of the time it will just be triggered because the app requested it.

But it simply means that in one case the user must do a redraw and in other we want to draw right away? If what you want is `RedrawRequested(Reason::User/System)`, then we could try to merge them, but what should we do with timer based rendering in such case? Should we indicate that scheduling is instant somehow?

rib:
> But it simply means that in one case the user must do a redraw and in other we want to draw right away?

This feels like a detail we're not seeing eye to eye on. I'm not sure what this means.

From the app pov it shouldn't care about `Reason::User/System`, so no I'm not suggesting that needs to be exposed in the event.

I'm trying to not complicate the discussion but I _do_ think that `RedrawRequested` should include a redraw region (either a single rectangle or a list of rectangles) but I think we can just ignore that as a separate topic for now.

E.g on macOS _all_ redraw events should happen via a `drawRect` callback. So I don't see any special cases like "in one case the user must do a redraw and in other we want to draw right away"

It's possible that macOS might call `drawRect` when we didn't request it and so the app might get a `RedrawRequested` because macOS requires the app to redraw but otherwise it will get a `drawRect` because the winit backend requested a `drawRect` callback in response to the application requesting a redraw.

Similarly for Windows. It's possible the app might sometimes get a `WM_PAINT` because of a window resize that could trigger a `RedrawRequested` event, but otherwise we should just get `WM_PAINT` events because the backend has asked for a `WM_PAINT` in response to the application requesting a redraw.

In all cases here though the application can just get a `RedrawRequested` event and it doesn't need to care whether it's a result of something like a window resize or a result of a request_redraw.

If the app isn't e.g. resizing it's window to trigger extra `drawRect` or `WM_PAINT` events then it will only be receiving `RedrawRequested` events after it requested to redraw, but these will still come off the back of `drawRect` or `WM_PAINT` (it's just that the backend will make a special request for these).

kchibisov:
@daxpedda I think that could work https://github.com/rust-windowing/winit/issues/2412#issuecomment-1596030990 just fine, I'm just not sure if it'll really keep at the monitor frame rate on anything other than Wayland, and how to deal with X11, but we can basically throttle transparently for the user, given that if they want to draw as fast as they can, they don't need any `request_redraw` in the first place.

daxpedda:
> [..], I'm just not sure if it'll really keep at the monitor frame rate [..]

> [..], but we can basically throttle transparently for the user, [..]

I don't think we should. This is a bigger issue that we shouldn't be handling here. Drawing at the monitor frame rate requires some sort of frame pacing, which isn't what we are doing here. Even the whole frame throttling thing is just "frame pacing provided by the system", it's not accurate and far from ideal.

> [..], given that if they want to draw as fast as they can, they don't need any `request_redraw` in the first place.

I thought they did, e.g. MacOS requires you to draw inside `drawRect`.

If this isn't the case and users can draw as fast as they can without `Window::request_redraw()`, then we could certainly remove it, in which case forget my arguments above and we could simplify more things as well.

kchibisov:
>I don't think we should. This is a bigger issue that we shouldn't be handling here. Drawing at the monitor frame rate requires some sort of frame pacing, which isn't what we are doing here. Even the whole frame throttling thing is just "frame pacing provided by the system", it's not accurate and far from ideal.

But I don't see an issue here? For example on Wayland we simply request the callbacks behind the user, and once they do `request_redraw` we align the delivery with the requested frame callback or send it back immediately. That will do exactly what macOS with `drawRect` does from what I can see, and we could do the same for web, and Windows.

I'd wait for @rib input here, but I don't see an issue with such approach.

If we don't want the throttling merged with `RedrawRequested`, we'd have some sort of timer based renderer, but we generally want the users to draw on `RedrawRequested`, or when they don't want to do so due to their scheduling, they could simply ignore the event.

Be aware that we don't force the re-drawings on users, they ask for them and we use windowing system mechanisms to 'start rendering at the right time'.

daxpedda:
After some IRC talk and clearing up some of the misconceptions I had, I arrived at the following conclusion:

There's two use-cases/scenarios Winit needs to handle:
- Frame-throttling.
- System notifies the user that the **MUST** redraw.

Instant redrawing doesn't require any interaction with Winit. To the API I would suggest is:
- We only need one event, it needs a field to tell the user if it's "forced" by the system or it comes from a frame throttling request.
- `Window` only needs one method, requesting a frame throttling event.

In the future we might also want to offer more time-based approaches to frame throttling, which I alluded to here: https://github.com/rust-windowing/winit/issues/2412#issuecomment-1595671021. This is unrelated to this PR though.

PR #2887, opened 2023-06-19
Bring `OptionAsAlt` back for macOS

kchibisov:
The correct handling of this setting requires to change the events we're getting from the macOS on the fly and call `interpretKeyEvents`, which could affect handling of the next events, meaning that we can't provide them on `KeyEvent`.

--

This is basically old option we had with winit `0.28.0`.

cc @fredizzimo if you want to try it with neovide.

### DIFF
--- a/src/platform/macos.rs
+++ b/src/platform/macos.rs
@@ -56,6 +56,17 @@ pub trait WindowExtMacOS {
 
     /// Put the window in a state which indicates a file save is required.
     fn set_document_edited(&self, edited: bool);
+
+    /// Set option as alt behavior as described in [`OptionAsAlt`].
+    ///
+    /// This will ignore diacritical marks and accent characters from
+    /// being processed as received characters. Instead, the input
+    /// device's raw character will be placed in event queues with the
+    /// Alt modifier set.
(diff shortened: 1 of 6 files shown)


src/platform/macos.rs
```diff
@@ -140,6 +161,10 @@ pub trait WindowBuilderExtMacOS {
     fn with_has_shadow(self, has_shadow: bool) -> WindowBuilder;
     /// Window accepts click-through mouse events.
     fn with_accepts_first_mouse(self, accepts_first_mouse: bool) -> WindowBuilder;
+    /// Set whether the `OptionAsAlt` key is interpreted as the `Alt` modifier.
```
madsmtm, line 164:
```suggestion
    /// Set how the <kbd>Option</kbd> keys are interpreted as the `Alt` modifier.
```

madsmtm (approved):
Only gave it a cursory review since I assume you've basically just copied the old implementation. Looks fine

src/platform/macos.rs
kchibisov: Yeah, I just reverted the bits used before + synchronized the handling of `key_up`/`key_down`.

PR #2889, opened 2023-06-20
Provide a way to set cursor area for IME cursor

### LINKED ISSUE #2886: Provide a way to set cursor area for `set_ime_position`
This is usually used to mark area which shouldn't be obscured by the system pop.

kchibisov:
Rename `Window::set_ime_position` to `Window::set_ime_cursor_area` adding a way to create cursor exclusive zone.

Fixes #2886.

cc @fredizzimo

--

I have no clue how to add it correctly on X11 given that what exactly its area arguments do is obscured.

### DIFF
--- a/src/platform_impl/macos/view.rs
+++ b/src/platform_impl/macos/view.rs
@@ -121,6 +121,7 @@ fn get_left_modifier_code(key: &Key) -> KeyCode {
 pub(super) struct ViewState {
     pub cursor_state: Mutex<CursorState>,
     ime_position: LogicalPosition<f64>,
+    ime_size: LogicalSize<f64>,
     pub(super) modifiers: Modifiers,
     phys_modifiers: HashMap<Key, ModLocationMask>,
     tracking_rect: Option<NSTrackingRectTag>,
@@ -166,6 +167,7 @@ declare_class!(
                 let state = ViewState {
                     cursor_state: Default::default(),
                     ime_position: LogicalPosition::new(0.0, 0.0),
+                    ime_size: Default::default(),
                     modifiers: Default::default(),
                     phys_modifiers: Default::default(),
                     tracking_rect: None,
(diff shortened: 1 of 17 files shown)


src/window.rs
```diff
@@ -1017,37 +1017,44 @@ impl Window {
...
     }
 
-    /// Sets location of IME candidate box in client area coordinates relative to the top left.
+    /// Set the IME cursor editing area, where the `position` is the top left corner of that area
+    /// and `size` is the size of this area starting from the position.
     ///
-    /// This is the window / popup / overlay that allows you to select the desired characters.
-    /// The look of this box may differ between input devices, even on the same platform.
+    /// The windowing system could place a candidate box close to that area, but try to not obscure
+    /// the specified area.
```
daxpedda, line 1024: What I'm reading here is that `size` specifies an area, that the candidate box will **avoid**.
I assume what's meant here is that the candidate box should try to **stay inside** the specified area?

kchibisov: No, it'll try to not obscure this area, simply so you can see what you're editing.

daxpedda: Ah, I see!

> Set the IME cursor editing area, where the `position` is the top left corner of that area and `size` is the size of this area starting from the position.

That line was confusing to me then, because it's not defined what the IME cursor editing area is, sounds to me like it's the candidate box.

kchibisov: I've added some clarification to that.

src/platform_impl/macos/view.rs
```diff
@@ -415,11 +417,8 @@ declare_class!(
             let base_y = (content_rect.origin.y + content_rect.size.height) as f64;
             let x = base_x + self.state.ime_position.x;
             let y = base_y - self.state.ime_position.y;
-            // This is not ideal: We _should_ return a different position based on
-            // the currently selected character (which varies depending on the type
-            // and size of the character), but in the current `winit` API there is
-            // no way to express this. Same goes for the `NSSize`.
```
madsmtm, line 421: Nit: While I think this PR improves the situation, by "currently selected character" I think I actually meant the `range` parameter passed to this function.

madsmtm (approved):
macOS impl looks good aside from a minor nit, thanks for doing this work!

src/platform_impl/macos/view.rs
kchibisov: Hm, it's a bit weird that it's like that then, but I think our macOS backend doesn't even allow to move within the preedit string from what I remember, given that each time I test it the input is stuck on the last character, so I guess it's a separate issue, like usually the spot is set around the cursor in the editor and these character range is sort of redundant given that we manage it manually.

madsmtm: I suspect the improper handling here is the reason the input is stuck at the end. But yeah, definitely a separate issue, and this PR does improve the situation.

PR #2896, opened 2023-06-22
Throttle `RedrawRequested`

### LINKED ISSUE #2412: Implement VSync source

kchibisov:
Fixes #2412.

- [ ] macOS (broken drawRect, DisplayLink?)
- [ ] Windows (WM_PAINT?)
- [x] Wayland (frame callbacks)
- [x] web (requestAnimationFrame)
- [ ] fallback (X11, redox) (timer based with global timer seed)

The docs are also not updated and it would be nice to have software example showing FPS it's rendering it right now.

--

Wayland done and it works.

### DIFF
--- a/src/window.rs
+++ b/src/window.rs
@@ -548,6 +548,43 @@ impl Window {
         self.window.request_redraw()
     }
 
+    /// Notify the windowing system that you're before presenting to the window.
+    ///
+    /// You should call this event after you've done drawing operations, but before you submit
+    /// the buffer to the display or commit your drawings. Doing so will help winit to properly
+    /// schedule and do assumptions about its internal state. For example, it could properly
+    /// throttle [`Event::RedrawRequested`].
+    ///
+    /// ## Example
+    ///
+    /// This example illustrates how it looks with OpenGL, but it applies to other graphics
+    /// APIs and software rendering.
+    ///
+    /// ```no_run
+    /// # use winit::event_loop::EventLoop;
+    /// # use winit::window::Window;
+    /// # let mut event_loop = EventLoop::new();
+    /// # let window = Window::new(&event_loop).unwrap();
+    /// # fn swap_buffers() {}
+    /// // Do the actual drawing with OpenGL.
+    ///
+    /// // Notify winit that we're about to submit buffer to the windowing system.
+    /// window.pre_present_notify();
+    ///
+    /// // Sumbit buffer to the windowing system.
+    /// swap_buffers();
+    /// ```
+    ///
+    /// ## Platform-specific
+    ///
+    /// **Wayland:** - schedules a frame callback to throttle [`Event::RedrawRequested`].
+    ///
+    /// [`Event::RedrawRequested`]: crate::event::Event::RedrawRequested
+    #[inline]
+    pub fn pre_present_notify(&self) {
+        self.window.pre_present_notify();
+    }
+
     /// Reset the dead key state of the keyboard.
     ///
     /// This is useful when a dead key is bound to trigger an action. Then
--- a/src/window.rs
+++ b/src/window.rs
@@ -524,11 +524,12 @@ impl Window {
         self.window.scale_factor()
     }
 
-    /// Requests a future [`Event::RedrawRequested`] event to be emitted in a way that is
-    /// synchronized and / or throttled by the windowing system.
+    /// Queues a [`Event::RedrawRequested`] event to be emitted that aligns with the windowing
+    /// system drawing loop.
     ///
     /// This is the **strongly encouraged** method of redrawing windows, as it can integrate with
-    /// OS-requested redraws (e.g. when a window gets resized).
+    /// OS-requested redraws (e.g. when a window gets resized). To improve the event delivery
+    /// consider using [`Window::pre_present_notify`] as described in docs.
     ///
     /// Applications should always aim to redraw whenever they receive a `RedrawRequested` event.
     ///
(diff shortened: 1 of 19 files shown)


kchibisov:
That's basically what I'm using for testing now https://github.com/kchibisov/glutin/tree/test-winit.

Similar changes could be done to `example/window`(it'll need a way to `request_redraw`.

rib:
I think it could be good to focus on one platform at a time with updating how RedrawRequested is throttled.

So maybe just look at the Wayland changes.

I have essentially implemented some of this for Windows in my pump_events branch, because I found it overly convoluted to try and maintain the old logic that was buffering redraw requests to be dispatched after MainEventsCleared.

I think it could also be good to add the present_notification API via a completely separate PR.

Conceptually I would see the present_notification API as way for Winit to get feedback about when the application renders which might be useful for a variety of use cases but in the short term it happens it provides a solution for letting the Wayland backend use frame callbacks.

So it's only indirectly related to throttling - it's needed for throttling on Wayland because otherwise it can't use frame callbacks but that feels like an implementation detail for the Wayland backend.

The stubbed macos change here also e.g. doesn't consider that `setNeedsDisplayInRect` can only be called from the main thread I'm fairly sure, so there still needs to be a queuing mechanism that goes via the main thread that could potentially call `setNeedsDisplayInRect`.

kchibisov:
>I think it could also be good to add the present_notification API via a completely separate PR.

I'll do 2 commits in this PR, because otherwise we'll add an API without a real use.

>The stubbed macos change here also e.g. doesn't consider that setNeedsDisplayInRect can only be called from the main thread I'm fairly sure, so there still needs to be a queuing mechanism that goes via the main thread that could potentially call setNeedsDisplayInRect.

I just tested whether they work at all, but they don't, I'll revert them all together, because `CVDisplayLink` should be used.


I think I'll focus mostly on docs in this PR.

rib:
> > I think it could also be good to add the present_notification API via a completely separate PR.
> 
> I'll do 2 commits in this PR, because otherwise we'll add an API without a real use.

yeah, makes sense

> I think I'll focus mostly on docs in this PR.

Just to note that I at least made some updates to docs in the pump_events branch.

Maybe you could base on top of my pump_events branch to avoid conflicts?

kchibisov:
>Maybe you could base on top of my pump_events branch to avoid conflicts?

I've added just some basic docs, we could improve them with your work ofc, they shouldn't overlap from what I could say.

kchibisov:
This is ready for review wrt docs.

@daxpedda if you want to send `requestAnimationFrame` you could do it after that patch, should be simply with your previous patch.

rib:
I think it would be best to rename `on_present_notify` to something like `pre_present_notify` because it's important to be very clear that that this callback needs to happen _before_ the application actually presents.

Just going by the name I think I would expect to call `on_present_notify` _after_ presenting.

Especially for the initial Wayland use case where the backend wants an opportunity to insert a frame callback before the `surface.commit()` - this ordering is important.

kchibisov:
I've renamed the method. @rib is this good for now?

src/window.rs
```diff
@@ -528,18 +528,24 @@ impl Window {
...
+    /// consider using [`Window::pre_present_notify`] as described in docs.
     ///
     /// This function can cause `RedrawRequested` events to be emitted after [`Event::MainEventsCleared`]
     /// but before `Event::NewEvents` if called in the following circumstances:
     /// * While processing `MainEventsCleared`.
     /// * While processing a `RedrawRequested` event that was sent during `MainEventsCleared` or any
     ///   directly subsequent `RedrawRequested` event.
     ///
+    /// However the event may not arrived on the next event loop, winit will try to align the
+    /// receiving the evente with the windowing system drawing loop.
```
daxpedda, line 541: I'm not sure what this sentence is trying to say.

daxpedda (changes requested):
Web backend implemented and approved.
Small question about the documentation.

src/window.rs
kchibisov: We have a concept that if you request a redraw before the `MainEventsCleared` you'll get the `RedrawRequsted` on the current loop iteration after it, I'm trying to say that it's no longer guaranteed.

daxpedda: Isn't that implicit now that it's bound to the windowing system drawing loop?

kchibisov: Well, we need to say that and that's what this sentence is trying to do.

Though, maybe better wording here is required.

daxpedda: I think this is the part we ought to change as well:
https://github.com/rust-windowing/winit/blob/ec5adcc83ec6eb12a03582809964fc1e7da28699/src/window.rs#L527-L528

```diff
-    /// Emits a [`Event::RedrawRequested`] event in the associated event loop after all OS
-    /// events have been processed by the event loop.
+    /// Queues a [`Event::RedrawRequested`] event to be emitted that aligns with the windowing
+    /// system drawing loop.
```

```suggestion
    /// However as the event aligns with the windowing system drawing loop, it may not arrive in
    /// the next event loop.
```

English isn't my first language, but I hope that's good enough.

daxpedda (approved):

kchibisov:
@daxpedda could you recheck that I've rebased correctly web?

rib:
One thing I was wanted to double check here @kchibisov is whether you have tested how apps behave if they omit the `pre_present_notify` call.

It would be good to be clear on whether this is now a hard requirement or whether the backend can reliably only enable throttling once it has seen that the application has called `pre_present_notify` at least once.

kchibisov:
>It would be good to be clear on whether this is now a hard requirement or whether the backend can reliably only enable throttling once it has seen that the application has called pre_present_notify at least once.

They will get instant redraw, so sounds fine to me, that's what they get before.

daxpedda (approved):
Web is fine, thanks @kchibisov.

PR #2899, opened 2023-06-22
Window: Keep window maximized [was: Do not needlessly call `set_inner_size` when setting min/max size]

hunger:
`set_inner_size` removes the maximized flag from a window (which makes sense IMHO).

So the window leaves maximized state needlessly, even when calls to `set_min_size_inner` and `set_max_size_inner` did not require any size change and the window could stay maximised just fine.

- [ ] ~Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users~


kchibisov:
Hm, but winit explicitly states in the docs that it's expected that the state will change.

kchibisov:
Ah, sorry, I somehow managed to read your description wrongly, nvm then.

msiglreith:
Thanks!
Have you verified that the min/max size are correctly respected with these changes? Calling `inner_size` will just invoke `GetClientRect` which doesn't trigger `WM_GETMINMAXINFO` where we currently clamp the size according to the set min/max bounds. Because of it I assume with this changes we will never update the size as both invocations of `inner_size` will return the same.

msiglreith (approved):

PR #2907, opened 2023-06-25
Don't allow event loops to run in parallel

daxpedda:
Currently multiple event loops can be run at the same time when using `spawn()`, which was not the intention of #2897 and isn't documented as such.

Follow up on #2897.
Cc @grovesNL.

### DIFF
--- a/src/platform_impl/web/event_loop/mod.rs
+++ b/src/platform_impl/web/event_loop/mod.rs
@@ -8,9 +8,7 @@ pub use self::window_target::EventLoopWindowTarget;
 
 use super::{backend, device, window};
 use crate::event::Event;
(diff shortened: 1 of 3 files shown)


grovesNL (approved):
Thank you for catching and fixing this oversight! This works fine for my use case.

arximboldi:
I wish this wan't merged. Why is this here? It makes things unnecessarily complicated on the user end. It should be perfectly fine to have multiple event loops running in parallel, running with multiple canvases?

daxpedda:
Allowing multiple event loops at the same time will collide with future changes, e.g. https://github.com/rust-windowing/winit/issues/3311.

Whats the motivation behind this?

PR #2911, opened 2023-06-26
Consider `RedrawEventsCleared` to be a redraw event

arendjr:
The `is_redraw()` method is used to determine when non-redraw events are received while processing redraws, and to log a warning when that happens. I noticed in a project of mine this resulted in endless spamming of warnings in the console, at least when running in the iOS simulator. Even so, I could not find any actual issues related to this.

By considering the `RedrawEventsCleared` event to also be a redraw event, the warnings stopped, while other behavior is left unmodified.

### DIFF
--- a/src/platform_impl/ios/app_state.rs
+++ b/src/platform_impl/ios/app_state.rs
@@ -59,7 +59,7 @@ enum UserCallbackTransitionResult<'a> {
 
 impl Event<'static, Never> {
     fn is_redraw(&self) -> bool {
-        matches!(self, Event::RedrawRequested(_))
+        matches!(self, Event::RedrawRequested(_) | Event::RedrawEventsCleared)
     }
 }
 


kchibisov:
The `RedrawEventsCleared` event will be gone, you should draw in response to `RedrawRequested` and request a redraw if you want to draw.

arendjr:
Oh sorry, I should have clarified I'm using Bevy, so I'm not directly listening to the events myself. Do you happen to know if they've already incorporated that change?

arendjr:
Maybe I misunderstood. When you said it would be gone, I thought you meant you were deprecating it, but I think you simply meant it would not be there to listen to?

From what I see in the code, the event is generated in this file, but then the code that handles it (where it calls `is_redraw()`) throws warnings. Maybe I’m misunderstanding, but it seems a local problem to me.

kchibisov:
I meant that the event will be removed entirely, so yes, you won't be able to listen to `RedrawEventsCleared`, there will be different event (`AboutToPoll`), but with a different meaning and implications.

Anyway, I'm not an ios maintainer to judge on the local problem in the backend, it's just a change looks weird because `RedrawEventsCleared` is never meant to be `redraw` event.

arendjr:
Ah, I see.

> it's just a change looks weird because RedrawEventsCleared is never meant to be redraw event.

Yeah, I do agree. The `is_redraw()` is private though, so if it helps I wouldn't mind renaming it to `is_redraw_related()` or something like that. But not sure how useful that is now, if it's going to be removed anyway...

rib:
Just to link them, here's the PR that's looking to remove `RedrawEventsCleared`: https://github.com/rust-windowing/winit/pull/2976

madsmtm:
Which now got merged, so I'll close this

Issue #2261, opened 2022-04-21
The web backend probably shouldn't be using the unload event


Liamolucko:
The unload event is currently used by the web backend to emit `Event::LoopDestroyed`. However, [it has several problems](https://developer.mozilla.org/en-US/docs/Web/API/Window/unload_event#usage_notes); it isn't guaranteed to be called, and it prevents the web browser from caching the page in case the user goes back to it.

I think that winit shouldn't use the `unload` event to emit `Event::LoopDestroyed`, and should instead document that it isn't guaranteed to emit it when the web page is closed.

daxpedda:
The [`pagehide`](https://developer.mozilla.org/en-US/docs/Web/API/Window/pagehide_event#usage_notes) event should be used instead. Still unreliable and as said should be documented as such.

PR #2912, opened 2023-06-26
On Web, map bfcache load/unload to suspend/resume

fornwall:

I'm not certain about this being the right way, so see it as a start of discussion.

In #2261, the web backend was changed to use the `pagehide` event instead of `unload`, to allow the bfcache to be used.

See https://web.dev/bfcache/ for more information about the bfcache (back/forward cache), and https://developer.chrome.com/docs/devtools/application/back-forward-cache/ for how to test it in chrome. Summary from the first resource:

> bfcache is an in-memory cache that stores a complete snapshot of a page (including the JavaScript heap) as the user is navigating away

The problem with winit and the bfcache is that the event loop is destroyed in `pagehide` - which happens before the snapshot is put in the bfcache. So when a user navigates back to the page running a winit program it's no longer responding.

This can be seen on e.g. the web example program (`cargo run-wasm --example web`) - after restoring the page from the bfcache the event log is no longer updated.

This PR changes `pagehide` handling to only destroy the event loop if [persisted](https://developer.mozilla.org/en-US/docs/Web/API/PageTransitionEvent/persisted) is `false`. If persisted is `true`, we now emit a `Suspended` event.

It also adds a corresponding listener for `pageshow` with `persisted` being `true`, to emit a `Resumed` event in that case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,7 +76,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On Web, use the correct canvas size when calculating the new size during scale factor change,
   instead of using the output bitmap size.
 - On Web, scale factor and dark mode detection are now more robust.
-- On Web, fix the bfcache by not using the `beforeunload` event.
+- On Web, fix the bfcache by not using the `beforeunload` event and map bfcache loading/unloading to `Suspended`/`Resumed` events.
 - On Web, fix scale factor resize suggestion always overwriting the canvas size.
 - On macOS, fix crash when dropping `Window`.
 - On Web, use `Window.requestIdleCallback()` for `ControlFlow::Poll` when available.
--- a/src/event.rs
+++ b/src/event.rs
@@ -125,6 +125,18 @@ pub enum Event<'a, T: 'static> {
     /// [`applicationWillResignActive`]: https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1622950-applicationwillresignactive
     /// [iOS application lifecycle]: https://developer.apple.com/documentation/uikit/app_and_environment/managing_your_app_s_life_cycle
     ///
+    /// ## Web
+    ///
+    /// On Web, the `Suspended` event is emitted in response to a [`pagehide`] event
+    /// with the property [`persisted`] being true, which means that the page is being
+    /// put in the [´bfcache`] (back/forward cache) - an in-memory cache that stores a
+    /// complete snapshot of a page (including the JavaScript heap) as the user is
+    /// navigating away.
+    ///
+    /// [`pagehide`]: https://developer.mozilla.org/en-US/docs/Web/API/Window/pagehide_event
+    /// [`persisted`]: https://developer.mozilla.org/en-US/docs/Web/API/PageTransitionEvent/persisted
+    /// [`bfcache`]: https://web.dev/bfcache/
+    ///
     /// [`Resumed`]: Self::Resumed
     Suspended,
 
@@ -178,6 +190,18 @@ pub enum Event<'a, T: 'static> {
     ///
     /// [`applicationDidBecomeActive`]: https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1622956-applicationdidbecomeactive
     /// [iOS application lifecycle]: https://developer.apple.com/documentation/uikit/app_and_environment/managing_your_app_s_life_cycle
+    ///
+    /// ## Web
+    ///
+    /// On Web, the `Resumed` event is emitted in response to a [`pageshow`] event
+    /// with the property [`persisted`] being true, which means that the page is being
+    /// restored from the [´bfcache`] (back/forward cache) - an in-memory cache that
+    /// stores a complete snapshot of a page (including the JavaScript heap) as the
+    /// user is navigating away.
+    ///
+    /// [`pageshow`]: https://developer.mozilla.org/en-US/docs/Web/API/Window/pageshow_event
+    /// [`persisted`]: https://developer.mozilla.org/en-US/docs/Web/API/PageTransitionEvent/persisted
+    /// [`bfcache`]: https://web.dev/bfcache/
     ///
     /// [`Suspended`]: Self::Suspended
     Resumed,
--- a/src/platform_impl/web/event_loop/runner.rs
+++ b/src/platform_impl/web/event_loop/runner.rs
@@ -18,7 +18,7 @@ use std::{
     rc::{Rc, Weak},
 };
 use wasm_bindgen::prelude::Closure;
-use web_sys::{KeyboardEvent, PointerEvent, WheelEvent};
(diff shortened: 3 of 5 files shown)


CHANGELOG.md
```diff
@@ -88,6 +88,7 @@ And please only add new entries to the top of this list, right below the `# Unre
   `DeviceEvent::Key` support.
 - **Breaking** `MouseButton` now supports `Back` and `Forward` variants, emitted from mouse events
   on Wayland, X11, Windows, macOS and Web.
+- On Web, do not destroy the event loop on `pagehide` events.
```
daxpedda, line 91: No need to add a new changelog entry, we didn't have a release since the last change to this. We should fold any more details into the last changelog entry about this.
https://github.com/rust-windowing/winit/blob/f9f42fbc4753933158c2090a2948f9aa1776ca37/CHANGELOG.md?plain=1#L79

daxpedda (changes requested):
I think what we should do here is look at [`PageTransitionEvent.persisted`](https://developer.mozilla.org/en-US/docs/Web/API/PageTransitionEvent/persisted).

If it's `true` we should send `Suspended` and also hook up the `pageshow` event and send a `Resumed` event.
If it's `false` we can resume to destroy the event loop.

CHANGELOG.md
fornwall: Thanks, updated this!

fornwall:
> I think what we should do here is look at [PageTransitionEvent.persisted](https://developer.mozilla.org/en-US/docs/Web/API/PageTransitionEvent/persisted).

> If it's true we should send Suspended and also hook up the pageshow event with a Resumed event. If it's false we can resume to destroy the event loop.

Thanks @daxpedda! I updated the PR according to this, could you have a look?

Specifically, is it correct to just `send_event(Event::Suspended)` and `send_event(Event::Resumed)`, or is something else required for proper lifecycle management?

src/platform_impl/web/event_loop/runner.rs
```diff
@@ -183,11 +183,24 @@ impl<T: 'static> Shared<T> {
...
+                }
+            },
+            move |event: PageTransitionEvent| {
+                if event.persisted() {
+                    show_runner.send_event(Event::Suspended);
+                } else {
+                    close_instance.handle_unload();
+                }
+            },
+        ));
```
daxpedda, line 203:
```suggestion
        *self.0.page_transition_event_handle.borrow_mut() = Some(backend::on_page_transition(
            self.window(),
            {
                let runner = self.clone();
                move |event: PageTransitionEvent| {
                    if event.persisted() {
                        runner.send_event(Event::Resumed);
                    }
                }
            },
            {
                let runner = self.clone();
                move |event: PageTransitionEvent| {
                    if event.persisted() {
                        runner.send_event(Event::Suspended);
                    } else {
                        runner.handle_unload();
                    }
                }
            },
        ));
```

src/platform_impl/web/event_loop/runner.rs
```diff
@@ -183,11 +183,24 @@ impl<T: 'static> Shared<T> {
...
-        *self.0.unload_event_handle.borrow_mut() =
-            Some(backend::on_unload(self.window(), move || {
-                close_instance.handle_unload()
-            }));
+        *self.0.page_transition_event_handle.borrow_mut() = Some(backend::on_page_transition(
+            self.window(),
+            move |event: PageTransitionEvent| {
+                if event.persisted() {
+                    hide_runner.send_event(Event::Resumed);
+                }
```
daxpedda, line 194: This should never happen unless it was persisted anyway.
We should turn this into a `debug_assert!` even.

src/event.rs
```diff
@@ -179,6 +191,18 @@ pub enum Event<'a, T: 'static> {
     /// [`applicationDidBecomeActive`]: https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1622956-applicationdidbecomeactive
     /// [iOS application lifecycle]: https://developer.apple.com/documentation/uikit/app_and_environment/managing_your_app_s_life_cycle
     ///
+    /// ## Web
+    ///
+    /// On Web, the `Resumed` event is emitted in response to a [`pageshow`] event
+    /// with the property [`persisted`] being true, which means that the page is being
+    /// restored from the [´bfcache`] (back/forward cache) - an in-memory cache that
+    /// stores a complete snapshot of a page (including the JavaScript heap) as the
+    /// user is navigating away.
```
daxpedda, line 200: As pointed out before, mentioning that `persisted` has to be true should be pointless here.

daxpedda (changes requested):
Great stuff, thank you!
Couple of nits.

Lifecycle management seems fine. Though I'm trying to figure out if it's necessary for us to actively ignore a `request_redraw()` after a `Suspend` event.

src/platform_impl/web/event_loop/runner.rs
fornwall: Thanks - applied this change!

src/platform_impl/web/event_loop/runner.rs
fornwall: A `pageshow` event with `persisted` being `true` is received on "normal" page loads - that is, when loading the page directly, not from the bfcache.

I made a small logger of `pageshow` and `pagehide` events here: https://fornwall.net/bfcache

Loading the page directly, then leaving it (putting it in the bfache), and navigating back (restoring it from the bfcache) shows the three events:

```js
// (1) Normal page load:
{"persisted":false,"type":"pageshow"}

// (2) Leaving the page, putting it in the bfache:
{"persisted":true,"type":"pagehide"}

// (3) Going back to the page, restoring it from the bfcache:
{"persisted":true,"type":"pageshow"}
```

src/event.rs
fornwall: See [the comment above](https://github.com/rust-windowing/winit/pull/2912#discussion_r1245099735)

fornwall:
@daxpedda Thanks again for the quick review!

I applied your suggested code change, and answered your comments about `pageshow` events happening initially with `persisted == false`. Let me know what you think!

src/platform_impl/web/event_loop/runner.rs
fornwall: > It is true that it could fire as part of the initial page load, but I couldn't get it to fire from Winit (when I removed the if check). By any chance would you know why?

@daxpedda A, good point! I guess this comes from the Winit wasm module being loaded async, so once we setup the event listener the initial `pageshow` has already fired?

But as you say, we should probably leave this code as it is now - it's more of an implementation detail that we are not catching the initial `pageshow` when loaded async, and in the future more wasm code might be loaded on the main thread (see https://chromestatus.com/feature/5099433642950656), so winit might actually receive the initial non-persisted `pageshow` event.

daxpedda: Thank you for digging into this.

daxpedda (approved):

PR #2918, opened 2023-07-01
Make KeyCodes Ord

StarStarJ:
I'd like to be able to use the KeyCodes as a reproducable chain, so sorting would be nice to have

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- Implement `PartialOrd` and `Ord` for `KeyCode` and `NativeKeyCode`.
+
 # 0.29.0-beta.0
 
 - On Web, allow event loops to be recreated with `spawn`.
--- a/src/keyboard.rs
+++ b/src/keyboard.rs
@@ -82,7 +82,7 @@ use smol_str::SmolStr;
 ///
 /// - Correctly match key press and release events.
 /// - On non-web platforms, support assigning keybinds to virtually any key through a UI.
-#[derive(Clone, Copy, PartialEq, Eq, Hash)]
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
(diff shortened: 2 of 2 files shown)


kchibisov:
Hm, the issue is that I'm not sure how much sense it makes, could you give a more concrete example, as in rust code for that?

It's a bit weird to sort keycodes when we don't have any numbers directly attached to them and we can reorder them at any point in time.

StarStarJ:
I basically have a BTreeSet of pressed keycodes

And a structure that represents a list of keys one has to press in order to trigger a specific event

so lets say the structure contains:
A => DoThat

B,A => DoThis
B has a potential list of follow up keycodes

C,B,A => DoAnother
C has a potential list of follow up keycodes, b as well

C,D,A => DoYetAnother
D is another keyboard in the list of C's follow up keycodes

The actual order does not matter, because the structure and the input keycodes are both sorted.
The key presses are evaluated every frame

I hope this example is good enough for you. I'm not a native english speaker, so it's not so easy to explain it for me

kchibisov:
Ah, so you basically require any ordering when pressing, not specific, since commonly you'd do a `Vec<KeyCode>` and then `index` to track current progress if you want to press, e.g. `gg` or `abc`.

Maybe it's fine to have `Ord` since it's for native keys.

kchibisov (approved):

kchibisov:
Could you add a `CHANGELOG` entry saying the types you've changed implement `Ord`/`PartialOrd`?

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- Add `PartialOrd` and `Ord` to `KeyCode` and `NativeKeyCode` to allow them to be compared, e.g. to store them into a `BTreeSet`
```
kchibisov, line 11: No need to have that many details.

```suggestion
- Implement `PartialOrd` and `Ord` for `KeyCode` and `NativeKeyCode`.
```

kchibisov (approved):

kchibisov:
Thanks!

PR #2924, opened 2023-07-06
Remove `EventLoopExtIOS::idiom` and `ios::Idiom`

madsmtm:
Introduced in #871 with unclear motivation.

@mtak- could you provide context for why Winit has this? Is it just for convenience?

### DIFF
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -154,7 +154,6 @@ If your PR makes notable changes to Winit's features, please update this section
 * Home indicator visibility
 * Status bar visibility and style
 * Deferring system gestures
-* Getting the device idiom
 * Getting the preferred video mode
 
 ### Web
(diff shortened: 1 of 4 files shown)


madsmtm:
I've decided to do this now, since the equivalent functionality is now easily accessible via. `objc2-ui-kit`.

daxpedda (approved):
> I've decided to pursue this now, since the equivalent functionality is now easily accessible via. `objc2-ui-kit`.

Pursue it how? Did you decide to leave it in Winit after all?

Just my 2¢: I do agree that this probably doesn't have a place in Winit, especially while there isn't any cross-platform API here.

madsmtm:
> Pursue it how? Did you decide to leave it in Winit after all?

I meant, I opened this PR a year ago, and decided to leave it in Winit until this day today, since it is only now that I have a good alternative to recommend users.

Issue #2334, opened 2022-06-13
Broadcast widnow mode changes


kchibisov:
When the widnow changes state, like it goes fullscreen, maximized, etc, winit should broadcast window state. I'd suggest to bundle it with resize events, and probably rename it.

```rust
pub struct WindowAttrs {
  maximized: bool,
  fullscreen: bool,
  minimized: bool,
}
```

```rust
enum WindowEvent {
  Configure(size: PhysicalSize<u32>, attrs: WindowAttrs)
}
```

This api is essential to know when the window gets fullscreened and disable things like opacity or change buttons if they're drawn by the window. Also the app could change layout in fullscreen (hide status bar, like browser).

BillyDM:
I also need this since I need to be able to detect when the window is minimized/un-minimized for my audio plugin host engine.

PR #2926, opened 2023-07-07
Brodcast window attributes

casperstorm:
Fixes https://github.com/rust-windowing/winit/issues/2334.

This PR adds a new `WindowEvent::WindowAttribute` which tells if the window is in fullscreen, minimize or normal mode.
Currently, macOS and Windows has been implemented, since that's the two platforms I am able to develop on.


src/event.rs
```diff
@@ -353,9 +353,27 @@ pub enum StartCause {
...
+}
+
 /// Describes an event from a [`Window`].
 #[derive(Debug, PartialEq)]
 pub enum WindowEvent<'a> {
+    /// An event which represents a change in the attribute of a [`Window`].
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **iOS / Android / Web / Orbital, X11, Wayland:** Unsupported.
```
casperstorm, line 374: For X11 I would assume we can use `UnmapNotify `and `MapNotify ` for minimized state, however I can't help work on it since I don't have a Linux machine at hand.

src/platform_impl/windows/event_loop.rs
```diff
@@ -1383,7 +1396,6 @@ unsafe fn public_window_callback_inner<T: 'static>(
                 let mut w = userdata.window_state_lock();
                 w.set_window_flags_in_place(|f| f.set(WindowFlags::MINIMIZED, true));
             }
-            // Send `WindowEvent::Minimized` here if we decide to implement one
```
casperstorm, line 1386: Using `WM_SIZE` instead, since sending a Minimized even here would only trigger for window header buttons, and not a shortcut.

kchibisov (changes requested):
In general, such information should be a part of the `Resized` event, since it's a part of the state change and resize will be generated as well.

I'm not sure if we should expose `minimize` for that, maybe we should, idk.

Also, could we maybe use something like that https://wayland.app/protocols/xdg-shell#xdg_toplevel:enum:state ? Or at least map some of them. Are tiled states exposed on Windows for example? What about macOS and how it all maps to it?

The new `Resized` event should be called `Configure` I guess to follow the `Wayland` naming since it sorts of makes sense and have a `size` just as a part of it's field and other field is the `bitflags` with the state.

We may have minimized there as well.

casperstorm:
> In general, such information should be a part of the `Resized` event, since it's a part of the state change and resize will be generated as well.
> 
> I'm not sure if we should expose `minimize` for that, maybe we should, idk.
> 
> Also, could we maybe use something like that https://wayland.app/protocols/xdg-shell#xdg_toplevel:enum:state ? Or at least map some of them. Are tiled states exposed on Windows for example? What about macOS and how it all maps to it?
> 
> The new `Resized` event should be called `Configure` I guess to follow the `Wayland` naming since it sorts of makes sense and have a `size` just as a part of it's field and other field is the `bitflags` with the state.
> 
> We may have minimized there as well.

We can rename it `Attribute` to `Configuration` and include `Resized(size)` and then do:

```rust
pub enum Configuration {
    /// The window is in normal mode
    Normal,
    /// The window is in fullscreen mode
    Fullscreen,
    /// The window is minimized
    Minimized,
    /// The size of the window was resized. Contains the client area's new dimensions.
    Resized(PhysicalSize<u32>),
}
```

As for tiling; to my knowledge there's no easy way to get tiling information. On macOS if you tile left or right, the full-screen selector is triggered.

kchibisov:
It should be


```rust
pub struct Configure {
    size: PhysicalSize<u32>,
    state: WindowState, // bitflags, TILED_LEFT, RIGHT, etc, | MAXIMIZED | FULLSCREEN | MINIMIZED
}
```

I could send this API change myself and you could open PR against my branch then.

kchibisov:
Could you send your patches against https://github.com/rust-windowing/winit/pull/2929 ?

casperstorm:
> Could you send your patches against https://github.com/rust-windowing/winit/pull/2929 ?

Yes will look at this when back from vacation in a couple of days.

src/platform_impl/macos/window_delegate.rs
```diff
@@ -305,6 +305,28 @@ declare_class!(
             options
         }
 
+        /// Invoked when miniaturize
+        #[sel(windowDidMiniaturize:)]
+        fn window_did_miniaturize(&mut self, _: Option<&Object>) {
+            trace_scope!("windowDidMiniaturize:");
+
+            self.queue_event(WindowEvent::Configured {
+                size: Default::default(),
```
casperstorm, line 314: What do you want here as size, @kchibisov?

kchibisov: You could send the current size.

src/platform_impl/macos/window_delegate.rs
```diff
@@ -305,6 +305,42 @@ declare_class!(
...
+
+            if let Some(screen) = self.window.screen() {
+                let rect = screen.frame();
+                let logical_size =
+                    LogicalSize::new(rect.size.width as f64, rect.size.height as f64);
+                let size = logical_size.to_physical::<u32>(self.window.scale_factor());
+
+                self.queue_event(WindowEvent::Configured {
+                    size,
+                    state: WindowState::MINIMIZED,
```
kchibisov, line 339: That looks wrong, because the window went away from being minimized.

src/platform_impl/macos/window_delegate.rs
```diff
@@ -332,6 +380,18 @@ declare_class!(
...
+                let rect = screen.frame();
+                let logical_size =
+                    LogicalSize::new(rect.size.width as f64, rect.size.height as f64);
+                let size = logical_size.to_physical::<u32>(self.window.scale_factor());
+
+                self.queue_event(WindowEvent::Configured {
+                    size,
+                    state: Default::default(),
+                });
+            }
```
kchibisov, line 394: In general, won't it all trigger the frame changes or other events? Can't we simply get the all the stiles of `zoomed`, `fullscreen` inside the resize handle or macOS won't resize to the same sizes? Do we have, maybe, some state changes hooks?

src/platform_impl/windows/event_loop.rs
```diff
@@ -1239,27 +1239,38 @@ unsafe fn public_window_callback_inner<T: 'static>(
...
-                {
+                if !window_flags.contains(WindowFlags::MARKER_RETAIN_STATE_ON_SIZE) {
                     let maximized = wparam == SIZE_MAXIMIZED as usize;
                     w.set_window_flags_in_place(|f| f.set(WindowFlags::MAXIMIZED, maximized));
+
+                    if w.fullscreen.is_some() {
+                        state = event::WindowState::FULLSCREEN;
+                    } else if window_flags.contains(WindowFlags::MINIMIZED) {
+                        state = event::WindowState::MINIMIZED;
+                    }
```
kchibisov, line 1262: Seems like `maximized` is left missing.

kchibisov (changes requested):
The maximized is missing on both `macOS` and `Windows`

src/platform_impl/macos/window_delegate.rs
casperstorm: We should use the delegate methods now they are there, imo. That's cleanest imo.

PR #2930, opened 2023-07-08
Rename `Window::set_inner_size` to `Window::request_inner_size`

### LINKED ISSUE #2868: Return the applied size for synchronious `set_inner_size`
The `Window::set_inner_size(Size)` is an async request with the semantic of asking the OS to resize, such request results in a response with the `Resized` event from the OS, however it's not the case on at least Wayland, because the buffer is managed by the client, and the intent of doing `set_inner_size` could mean that you want to draw the window with the sizes you want. However on async platforms you can't really do that, because you have to wait for OS to reply back.

What we could do is the following:

```rust
#[must_use]
fn set_inner_size(Size) -> Option<PhysicalSize>
```

Where `Some` will be returned when the value was applied immediately so the users can use it and `None` when the request went to OS for waiting.

The key point here, is that some clients might want to draw with the sizes they want to set on the current frame, so what they could do with such API is the following

```rust
let new_size = window.set_inner_size(PhysicalSize(200, 200)).unwrap_or_else(|| window.inner_size());
```

So when they want to draw with the sizes they set, they'll not unconditionally assume that what they've used was applied right away.

In addition, the `set_inner_size` should state explicitly that semantic of `Size` under the hood is not being taken into the account and the value which fits better for the backend is used.

kchibisov:
Some systems could resize the window immediately and we'd rather inform the users right away if that was the case, so they could create e.g. EGLSurface without waiting for resize, which is really important for Wayland.

Fixes #2868.

### DIFF
--- a/src/window.rs
+++ b/src/window.rs
@@ -188,7 +188,7 @@ impl WindowBuilder {
     ///
     /// If this is not set, some platform-specific dimensions will be used.
     ///
-    /// See [`Window::set_inner_size`] for details.
+    /// See [`Window::request_inner_size`] for details.
     #[inline]
     pub fn with_inner_size<S: Into<Size>>(mut self, size: S) -> Self {
         self.window.inner_size = Some(size.into());
@@ -652,10 +652,21 @@ impl Window {
         self.window.inner_size()
     }
 
-    /// Modifies the inner size of the window.
+    /// Request the new size for the window.
+    ///
+    /// On platforms where the size is entirely controlled by the user the
+    /// applied size will be returned immediatelly, resize event in such case
+    /// may not be generated.
+    ///
+    /// On platforms where resizing is disallowed by the windowing system, the current
+    /// inner size is returned immidiatelly, and the user one is ignored.
+    ///
+    /// When `None` is returned, it means that the request went to the display system,
+    /// and the actual size will be delivered later with the [`WindowEvent::Resized`].
     ///
     /// See [`Window::inner_size`] for more information about the values.
-    /// This automatically un-maximizes the window if it's maximized.
+    ///
+    /// The request could automatically un-maximize the window if it's maximized.
     ///
     /// ```no_run
     /// # use winit::dpi::{LogicalSize, PhysicalSize};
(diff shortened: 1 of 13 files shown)


kchibisov:
Implementation on some platforms may not represent the actual state of things, I'd suggest to ensure at least `macOS`, `Windows`, `Web` implementations.

notgull (approved):

src/window.rs
```diff
@@ -652,10 +652,21 @@ impl Window {
...
 
-    /// Modifies the inner size of the window.
+    /// Request the new size for the window.
+    ///
+    /// On platforms where the size is entirely controlled by the user the
+    /// applied size will be returned immediatelly, resize event in such case
+    /// may not be generated.
+    ///
+    /// On platforms where resizing is disallowed by the windowing system, the current
+    /// inner size is returned immidiatelly, and the user one is ignored.
```
daxpedda, line 662:
```suggestion
    /// inner size is returned immediately, and the user one is ignored.
```

daxpedda:
Should a `Resized` event be triggered after a `Window::request_inner_size()` even if the size didn't change because e.g. rounding or max-size reached?

kchibisov:
>Should a Resized event be triggered after a Window::request_inner_size() even if the size didn't change because e.g. rounding or max-size reached?

it's up to backend though. If server won't sent a new size it's up to it. Nothing we can do here to be fair.

daxpedda:
> Nothing we can do here to be fair.

On Web we can just request it.

daxpedda (approved):
Web is correct like this, nothing more to be done here except the ongoing discussion about triggering the event if nothing changed.

src/window.rs
```diff
@@ -652,10 +652,21 @@ impl Window {
         self.window.inner_size()
     }
 
-    /// Modifies the inner size of the window.
+    /// Request the new size for the window.
+    ///
+    /// On platforms where the size is entirely controlled by the user the
+    /// applied size will be returned immediatelly, resize event in such case
```
msiglreith, line 658:
```suggestion
    /// applied size will be returned immediately, resize event in such case
```

msiglreith (approved):

PR #2937, opened 2023-07-08
On Web, support Fullscreen API on Safari

### LINKED ISSUE #2935: Fullscreen API compatibility with Safari
Currently we are using [`Element.requestFullscreen()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/requestFullscreen), [`Document.exitFullscreen()`](https://developer.mozilla.org/en-US/docs/Web/API/Document/exitFullscreen), [`Document.fullscreenElement`](https://developer.mozilla.org/en-US/docs/Web/API/Document/fullscreenElement) and the [`fullscreenchange`](https://developer.mozilla.org/en-US/docs/Web/API/Element/fullscreenchange_event) event (in #2929) to access the Fullscreen API.

However the whole API is prefixed on Desktop Safari <16.4 and on iOS Safari (only iPad, on iPhones the whole API isn't supported), which we currently don't take into account.

I will fix this after #2929.

daxpedda:
Builds on top of #2929.

Fixes #2935.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 # Unreleased
 
 - Implement `PartialOrd` and `Ord` for `KeyCode` and `NativeKeyCode`.
+- **Breaking:** Rename `WindowEvent::Resized` to `WindowEvent::Configured` and expose active window states.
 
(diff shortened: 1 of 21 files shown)


daxpedda:
Replaced by #2948.

PR #2938, opened 2023-07-09
Improve detection of key repeats on X11

Imberflur:
Instead of a single `bool` indicating that a key press has occurred and no key has been released since then, we store the scancode of the last pressed key (if it is a key that repeats when held). This fixes a bug where pressing a new key while one is already held down will be flagged as a repeat even though it is obviously not a repeat.

Thanks to @kchibisov for guidance :heart:

### DIFF
--- a/src/platform_impl/linux/x11/event_processor.rs
+++ b/src/platform_impl/linux/x11/event_processor.rs
@@ -31,8 +31,11 @@ pub(super) struct EventProcessor<T: 'static> {
     pub(super) kb_state: KbdState,
     // Number of touch events currently in progress
     pub(super) num_touch: u32,
-    // Whether we've got a key release for the key press.
-    pub(super) got_key_release: bool,
+    // This is the last pressed key that is repeatable (if it hasn't been
+    // released).
+    //
+    // Used to detect key repeats.
+    pub(super) held_key_press: Option<u32>,
     pub(super) first_touch: Option<u64>,
     // Currently focused window belonging to this process
     pub(super) active_window: Option<ffi::Window>,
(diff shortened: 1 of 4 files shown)


kchibisov:
Thanks

Could you apply something like that? I think that should be equivalent to your code.

```diff
diff --git a/src/platform_impl/linux/x11/event_processor.rs b/src/platform_impl/linux/x11/event_processor.rs
index 36dc209f..1f80fbdd 100644
--- a/src/platform_impl/linux/x11/event_processor.rs
+++ b/src/platform_impl/linux/x11/event_processor.rs
@@ -551,6 +551,8 @@ impl<T: 'static> EventProcessor<T> {
                 let window_id = mkwid(window);
                 let device_id = mkdid(util::VIRTUAL_CORE_KEYBOARD);
 
+                let keycode = xkev.keycode as _;
+
                 // Update state to track key repeats and determine whether this key was a repeat.
                 //
                 // Note, when a key is held before focusing on this window the first
@@ -561,24 +563,20 @@ impl<T: 'static> EventProcessor<T> {
                 // Only keys that can repeat should change the held_key_press state since a
                 // continuously held repeatable key may continue repeating after the press of a
                 // non-repeatable key.
-                let repeat = if self.kb_state.key_repeats(xkev.keycode) {
-                    let was_held = self.held_key_press == Some(xkev.keycode);
-
+                let repeat = if self.kb_state.key_repeats(keycode) {
                     if ty == ffi::KeyPress {
-                        self.held_key_press = Some(xkev.keycode);
+                        let was_held = self.held_key_press == Some(keycode);
+                        self.held_key_press = Some(keycode);
                         was_held
                     } else {
-                        if was_held {
-                            self.held_key_press = None;
-                        }
+                        // Key release will only be send for original key press.
+                        self.held_key_press = None;
                         false
                     }
                 } else {
                     false
                 };
 
-                let keycode = xkev.keycode as _;
-
                 let state = if ty == ffi::KeyPress {
                     ElementState::Pressed
                 } else {

```

Imberflur:
I can move `keycode` up, since it is being cast to a u32 which is potentially a different integer type I will probably also change the field type of `held_key_press`.

I already almost removed the `was_held` check when releasing the key but realized that previously pressed keys that aren't the last pressed may be released at that point which iirc doesn't cancel the current repeat (will double check that though). So we need the check that the key being released was the last one pressed. I think that check could benefit from a comment.

Imberflur:
Updated with changes.

Imberflur:
Renamed `was_held` to `is_latest_held` to improve clarity

src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -548,13 +551,39 @@ impl<T: 'static> EventProcessor<T> {
...
+
+                    if ty == ffi::KeyPress {
+                        self.held_key_press = Some(keycode);
+                        is_latest_held
+                    } else {
+                        // Check that the released key is the latest repeatable key that has been
+                        // pressed, since repeats will continue for the latest key press if a
+                        // different previously pressed key is released.
+                        if is_latest_held {
+                            self.held_key_press = None;
```
kchibisov, line 576: Ah, yeah, so this is when releasing other key during repeat...

kchibisov (approved):

PR #2941, opened 2023-07-09
Improve Web specific documentation for various APIs 

daxpedda:
Just documenting some Web specific behavior to `Window::(set_)inner_size`.

Cc #2859.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -430,6 +430,14 @@ pub enum WindowEvent<'a> {
     Ime(Ime),
 
     /// The cursor has moved on the window.
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **Web:** Doesn't take into account CSS [`border`], [`padding`], or [`transform`].
+    ///
+    /// [`border`]: https://developer.mozilla.org/en-US/docs/Web/CSS/border
+    /// [`padding`]: https://developer.mozilla.org/en-US/docs/Web/CSS/padding
+    /// [`transform`]: https://developer.mozilla.org/en-US/docs/Web/CSS/transform
     CursorMoved {
         device_id: DeviceId,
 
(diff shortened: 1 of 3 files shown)


src/window.rs
```diff
@@ -644,7 +644,7 @@ impl Window {
     ///
     /// - **iOS:** Can only be called on the main thread. Returns the `PhysicalSize` of the window's
     ///   [safe area] in screen space coordinates.
-    /// - **Web:** Returns the size of the canvas element.
+    /// - **Web:** Returns the size of the canvas element. Doesn't account for `transform` or `scale`.
```
kchibisov, line 647: But that's sounds wrong then, because `PhysicalSize` must account for scale, as in it's the size after all the scaling.

Not sure that it should be the size after all the transform though, because transform is difficult to handle if it matters.

daxpedda: Just to clarify we are talking about CSS [`scale`](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/scale), not scale factor.

Would love to account for both, but unfortunately that's impossible because `ResizeObserver` doesn't account for it, so we can't fire a `Resized` (`Configured`) event when the `scale` changes. Same for `transform`, which is even more complicated because that can get really tough to calculate.

Additionally `transform` can lead to non square shaped canvases, which we can't really represent in Winit and I don't think we should try to.

daxpedda: I added a clarification that this is about CSS and not scale factor.

daxpedda: See https://github.com/w3c/csswg-drafts/issues/3554#issuecomment-494199650, which commented on this issue.

kchibisov: Is it correct that you can simply pass this size to wgpu or some other rendering stuff without any adjustment and you'll get the correct rendering without rescaling by the browser?

kchibisov: >Just to clarify we are talking about CSS scale, not scale factor.

Then you should explicitly state that you're talking about CSS and not `Window::scale_factor`.

daxpedda: > Is it correct that you can simply pass this size to wgpu or some other rendering stuff without any adjustment and you'll get the correct rendering without rescaling by the browser?

Unless your are using these CSS properties, yes.

> Then you should explicitly state that you're talking about CSS and not `Window::scale_factor`.

I just discovered that CSS `scale` is a sub-property of `transform`, so I will remove the mention of it completely now.

daxpedda: Done, I would appreciate another look.

src/event.rs
```diff
@@ -84,7 +84,7 @@ pub enum Event<'a, T: 'static> {
     ///
     /// Not all platforms support the notion of suspending applications, and there may be no
     /// technical way to guarantee being able to emit a `Suspended` event if the OS has
-    /// no formal application lifecycle (currently only Android and iOS do). For this reason,
+    /// no formal application lifecycle (currently only Android, iOS and Web do). For this reason,
```
kchibisov, line 87:
```suggestion
    /// no formal application lifecycle (currently only Android, iOS, and Web do). For this reason,
```

src/event.rs
```diff
@@ -1148,7 +1148,7 @@ pub struct Touch {
     ///
     /// ## Platform-specific
     ///
-    /// - Only available on **iOS** 9.0+ and **Windows** 8+.
+    /// - Only available on **iOS** 9.0+, **Windows** 8+ and **Web**.
```
kchibisov, line 1151:
```suggestion
    /// - Only available on **iOS** 9.0+, **Windows** 8+, and **Web**.
```

kchibisov (changes requested):
other than that should be fine.

daxpedda:
Done, merging blocked because of the requested change :rofl:.

kchibisov (approved):

PR #2943, opened 2023-07-09
Port to use raw-window-handle v0.6

notgull:

cc rust-windowing/raw-window-handle#125

Testing out the new release before it becomes official.

Tests will fail until I also port `softbuffer` to the new version.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,7 +55,8 @@ cursor-icon = "1.0.0"
 log = "0.4"
 mint = { version = "0.5.6", optional = true }
 once_cell = "1.12"
-raw_window_handle = { package = "raw-window-handle", version = "0.5", features = ["std"] }
+raw_window_handle = { package = "raw-window-handle", git = "https://github.com/rust-windowing/raw-window-handle.git", features = ["std"] }
+raw_window_handle_05 = { package = "raw-window-handle", version = "0.5" }
 serde = { version = "1", optional = true, features = ["serde_derive"] }
 smol_str = "0.2.0"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,7 +65,7 @@ image = { version = "0.24.0", default-features = false, features = ["png"] }
 simple_logger = { version = "2.1.0", default_features = false }
 
 [target.'cfg(not(any(target_os = "android", target_os = "ios")))'.dev-dependencies]
-softbuffer = "0.3.0"
+softbuffer = { git = "https://github.com/rust-windowing/softbuffer.git", branch = "notgull/rwh-v0.6" }
 
 [target.'cfg(target_os = "android")'.dependencies]
 # Coordinate the next winit release with android-ndk-rs: https://github.com/rust-windowing/winit/issues/1995
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,7 +55,7 @@ cursor-icon = "1.0.0"
 log = "0.4"
 mint = { version = "0.5.6", optional = true }
 once_cell = "1.12"
-raw_window_handle = { package = "raw-window-handle", git = "https://github.com/rust-windowing/raw-window-handle.git", features = ["std"] }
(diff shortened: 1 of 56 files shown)


daxpedda (approved):
Web part approved.

examples/child_window.rs
```diff
@@ -37,12 +38,14 @@ fn main() {
     let mut windows = HashMap::new();
 
     let event_loop: EventLoop<()> = EventLoop::new();
-    let parent_window = WindowBuilder::new()
-        .with_title("parent window")
-        .with_position(Position::Logical(LogicalPosition::new(0.0, 0.0)))
-        .with_inner_size(LogicalSize::new(640.0f32, 480.0f32))
-        .build(&event_loop)
-        .unwrap();
+    let parent_window = Rc::new(
```
kchibisov, line 41: What's the reason we have `Rc` now all over the place for the builder?

notgull: This is so the `softbuffer` contexts can hold a reference to the window/display objects. Essentially the new system prevents the issue we saw earlier where Wayland would crash if the `softbuffer` contexts outlived the event loop. The flip side of this is that you need to have a defined type that the `softbuffer` can hold on to (like an `&'a Window` or, in this case, an `Rc<Window>`) so that it can keep the window handle alive.

kchibisov: Then if you're fixing this particular issue you should remove the `ManuallyDrop` otherwise I don't see what it tries to do so?

src/event_loop.rs
```diff
@@ -370,12 +391,82 @@ impl<T> EventLoopWindowTarget<T> {
...
+/// windows. This comes in play, for example, for OpenGL setup. However, it is impossible to hold
+/// a reliable reference to the [`EventLoop`] or an [`EventLoopWindowTarget`] while the event loop
+/// is running. This type aims to fill that gap by being a generic-free, cheaply clonable handle
+/// representing the display server.
+///
+/// If you already have a [`Window`], you can get a display handle from it instead of this.
+///
+/// [`Window`]: crate::window::Window
+#[derive(Clone)]
+pub struct OwnedDisplayHandle {
```
madsmtm, line 444: I'm uncertain if this is the best design, and if it would maybe make sense to have this in the `raw-window-handle` crate instead? Though I guess the `Drop` impl can't really be done in that crate, so probably not :/

madsmtm (changes requested):
MacOS and iOS impls approved.

Cargo.toml
```diff
@@ -55,7 +55,8 @@ cursor-icon = "1.0.0"
 log = "0.4"
 mint = { version = "0.5.6", optional = true }
 once_cell = "1.12"
-raw_window_handle = { package = "raw-window-handle", version = "0.5" }
+raw_window_handle = { package = "raw-window-handle", version = "0.6", features = ["std"] }
+raw_window_handle_05 = { package = "raw-window-handle", version = "0.5" }
```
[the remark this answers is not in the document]
notgull: Maybe that would be a good idea, but IMO it's out of scope for this PR.

src/event_loop.rs
notgull: It's unfortunately necessary for tying the lifetimes of things that need the raw display (like OpenGL) when there isn't a window available yet. As the `EventLoop` must be held exclusively to be polled, we need another type to represent the actual underlying display that can be passed to OpenGL/wgpu/etc.

"Display" is used here as a term for the connector between the client application and the window manager server, like `xcb_connection` or `wl_display`. I'm open to better names.

madsmtm: Undestood - I've asked follow-up questions on Matrix

Cargo.toml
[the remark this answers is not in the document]
madsmtm: I've moved this proposal to https://github.com/rust-windowing/winit/pull/3075

madsmtm:
I'd argue this is blocked on https://github.com/rust-windowing/winit/pull/3075

notgull:
Obsoleted by #3126

PR #2956, opened 2023-07-12
On macOS, add tabbing APIs

kchibisov:
This should let the users control macOS tabbing and allow to create windows in tab.

### DIFF
--- a/src/platform/macos.rs
+++ b/src/platform/macos.rs
@@ -1,4 +1,4 @@
-use std::os::raw::c_void;
+use std::{num::NonZeroUsize, os::raw::c_void};
 
 use objc2::rc::Id;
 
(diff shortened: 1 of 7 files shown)


src/platform/macos.rs
```diff
@@ -38,6 +38,33 @@ pub trait WindowExtMacOS {
     /// Sets whether or not the window has shadow.
     fn set_has_shadow(&self, has_shadow: bool);
 
+    /// Sets whether the system can automatically organize windows into tabs.
+    ///
+    /// <https://developer.apple.com/documentation/appkit/nswindow/1646657-allowsautomaticwindowtabbing>
+    fn set_allows_automatic_window_tabbing(&self, enabled: bool);
+
+    /// Returns whether the system can automatically organize windows into tabs.
+    fn allows_automatic_window_tabbing(&self) -> bool;
```
amrbashir, line 47: I never noticed before but modifying the automatic tabbing behaviour is actually app-wide and not specific to one window, should it be implemented on `EventLoopWindowTargetExtMacOS` and `EventLoopBuilderExtMacOS` instead?

kchibisov: Hm, yeah, I haven't thought about it. We could move it all, I guess. I just got confused because it was a static method on a window.

kchibisov: Moved

Samasaur1:
Would it be possible to add a `select_last_tab` function as well? That's the default behavior of Cmd-9 on macOS.

kchibisov:
@Samasaur1 hm, I think it wasn't doing that for me in Terminal.app, so I've assumed that it was something else. I could whether it was really the case and an api to query amount of windows being tabbed.

Samasaur1:
@kchibisov maybe it's not systemwide? I know it's the behavior in browsers.

I'm making a PR right now to add this functionality.

kchibisov:
I think you don't have to actually, we could do it with the current API.

kchibisov:
Ah, no, we can't because we need to follow the no-op stuff. Could you add an API to query number of tabs from the window?

kchibisov:
Also, maybe we should do tab start with zero instead of one, like we have now, but we can discuss that on the follow up PR.

Samasaur1:
OK, I've gotten held up on the new PR because this one doesn't actually work for me? Running `cargo run --example window_tabbing` and pressing `t` opens a new window instead of a new tab, unless "Prefer tabs when opening documents" is set to "Always" — which is the same behavior as before the PR.

I've tested this on multiple machines and the behavior is the same on all of them

Samasaur1:
That did not work, but I think I have found the problem!

Each `NSWindow` has [a `tabbingMode` property](https://developer.apple.com/documentation/appkit/nswindow/1644729-tabbingmode?language=objc). By default, this is set to `NSWindowTabbingModeAutomatic`, which uses the system `userTabbingPreference` to determine tabbing behavior — which would explain why me changing "Prefer tabs when opening documents" showed the correct behavior. We should set the value of this property to `NSWindowTabbingModePreferred` for each window before showing it, which will cause windows with the same tabbing identifier to correctly tab together.

I'm going to test this to ensure my hypothesis is correct — will report back

Samasaur1:
@kchibisov I was correct, and have fixed the issue. See the linked pull request.

PR #2958, opened 2023-07-13
Fix new tab creation on macOS

Samasaur1:
Set tabbingMode on all macOS windows so that new tabs are correctly created as tabs, rather than separate windows.

See https://github.com/rust-windowing/winit/pull/2956#issuecomment-1634647001


kchibisov:
if we disable automatic tabbing will it still tab?

kchibisov:
Thanks!

PR #2965, opened 2023-07-18
Fix window decoration check method for Gnome environment (using libadwaita)

citorva:

This PR makes a little change in the code in order to fix the method which check if the window have decorations by inverting the return value of the function.

This issue was detected when using the [window_resize_increments](https://github.com/rust-windowing/winit/blob/master/examples/window_resize_increments.rs) example on my Arch Linux Gnome desktop.

Additionally, this PR adds a feature on this example allowing to move the window using the system behaviors when borders are disabled


kchibisov (approved):
This technically needs a changelog entry, but given that it'll be removed during release I don't really mind...

PR #2971, opened 2023-07-22
Correctly detect that we don't support Emscripten

### LINKED ISSUE #2970: Building `wasm32-unknown-emscripten` fails since it includes both wasm and linux platform
Worked out of the box in 0.27.5, but in 0.28.0 it seems that the config checks in https://github.com/rust-windowing/winit/blob/master/src/platform_impl/mod.rs#L9 incorrectly detect `wasm32-unknown-emscripten` as a linux platform, causing compilation failure

See https://github.com/gfx-rs/wgpu/actions/runs/5618255769/job/15223561268?pr=3950#step:9:55

daxpedda:
Winit fails compilation on unsupported targets. But we didn't account for Emscripten until now, it was recognized as a Wasm and Unix platform.

This correctly account for Emscripten.

Fixes #2970.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -8,12 +8,12 @@ fn main() {
     cfg_aliases! {
         // Systems.
         android_platform: { target_os = "android" },
-        wasm_platform: { target_family = "wasm" },
+        wasm_platform: { all(target_family = "wasm", not(target_os = "emscripten")) },
         macos_platform: { target_os = "macos" },
         ios_platform: { target_os = "ios" },
         windows_platform: { target_os = "windows" },
         apple: { any(target_os = "ios", target_os = "macos") },
-        free_unix: { all(unix, not(apple), not(android_platform)) },
+        free_unix: { all(unix, not(apple), not(android_platform), not(target_os = "emscripten")) },
         redox: { target_os = "redox" },
 
         // Native displays.


build.rs
```diff
@@ -8,12 +8,12 @@ fn main() {
...
         // Systems.
         android_platform: { target_os = "android" },
-        wasm_platform: { target_family = "wasm" },
+        wasm_platform: { all(target_family = "wasm", not(target_os = "emscripten")) },
         macos_platform: { target_os = "macos" },
         ios_platform: { target_os = "ios" },
         windows_platform: { target_os = "windows" },
         apple: { any(target_os = "ios", target_os = "macos") },
-        free_unix: { all(unix, not(apple), not(android_platform)) },
+        free_unix: { all(unix, not(apple), not(android_platform), not(target_os = "emscripten")) },
```
kchibisov, line 16: Could we just `not(wasm_platform)` here?

daxpedda: Emscripten is now also excluded from `wasm_platform`, so that wouldn't be accounted for here.
(I'm not sure why Emscripten has `unix` to begin with anyway ...)

kchibisov: Ah, yeah, it'll double negate.

kchibisov (approved):

Issue #2900, opened 2023-06-22
EventLoop 3.0 Changes


rib:
Considering the changes being made for `run()-> Result<>`, `pump_events`, `run_ondemand`, and addressing inconsistent ordering guarantees for events like `MainEventsCleared` and `RedrawRequested`, as well as removing and renaming several core events, then it feels like a similar level of change as happened for EventLoop 2.0.

The details are a bit scattered though, and although some things have been agreed as part of lengthy issue discussions we still need to track those ideas at least as a reminder as to what is left to do.

I'm going to be lite on details here, but a lot of the background discussion probably happened here https://github.com/rust-windowing/winit/issues/2706 and maybe a few other issues we can link back here too.

- [x] Replace `run() ->!` with `run() -> Result` as our portable run API available on _all_ platforms (https://github.com/rust-windowing/winit/pull/2767)
- [x] Replace `run_return()` with `pump_events()` and `run_ondemand()` (for Windows, macOS, Linux + Android) (https://github.com/rust-windowing/winit/pull/2767)
- [ ] Maybe add `run_noreturn() ->!` extension API for iOS
- [x] Rename `LoopDestroyed` `LoopExiting` (https://github.com/rust-windowing/winit/pull/2975)
- [x] Remove `RedrawEventsCleared` (https://github.com/rust-windowing/winit/pull/2976)
- [x] Rename `MainEventsCleared` -> ~~`AboutToPoll`~~ `AboutToWait` (Conceptually `NewEvents` and `AboutToWait` can be seen as a way to observe the underlying blocking/polling of the event loop - similar to registering an observer with a macOS `RunLoop`) (https://github.com/rust-windowing/winit/pull/2976)
- [x] Document that `RedrawRequested` is no longer guaranteed to be delivered in order with respect to any other event - it is platform specific 
- [x] Make it clearer that applications should redraw any time they get a `RedrawRequested` callback (and they shouldn't defer the redraw outside of the winit callback).
- [ ] Make it clearer that `Resumed` events are delivered on all platforms (not just mobile) and `Resumed` events are a good place to initialize an application because it's the first point where it's possible to start creating Windows across all platforms.
- [x] Remove lifetime from the `Event`
- [ ] Aim to implement `RedrawRequested` throttling by default where there's a suitable mechanism provided by the window system
  - [x] By `WM_PAINT` on Windows
  - [ ] Either with `drawRect` or display link on macOS (same for iOS?)
  - [x] By frame callbacks on Wayland
[...]

PR #2973, opened 2023-07-22
Add `Event::try_into_static` and `Event::try_clone`

### LINKED ISSUE #1968: Variant of `Event::to_static` that takes self by reference
Currently `Event::to_static` and `WindowEvent::to_static` takes self by value. This is inconvenient because static variants of these  enums are required to clone them. And so if one needs to clone these enums for any reason one also ends up needing to throw away the `WindowEvent::ScaleFactorChanged` event.

I propose instead to either change `to_static` to take a `&self` or alternatively add an additional method `to_static_cloned`.

birktj:
I have added `Event::try_into_static` to enable converting an event to a static one, but still be able to keep a potential `ScaleFactorChanged` event.

I have also added `Event::try_clone` which is probably more ergonomic to use in most cases.

In the same change I have renamed `Event::to_static` to `Event::into_static` as suggested in the linked issue to be more consistent with the overall rust ecosystem. 

This PR fixes #1968

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On Web, remove unnecessary `Window::is_dark_mode()`, which was replaced with `Window::theme()`.
 - On Web, add `WindowBuilderExtWebSys::with_append()` to append the canvas element to the web page on creation.
 - On Windows, add `drag_resize_window` method support.
+- **Breaking:** Rename `WindowEvent`/`Event::to_static` to `WindowEvent`/`Event::into_static`.
 
 # 0.29.0-beta.0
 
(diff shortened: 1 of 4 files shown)


kchibisov:
Thank for your patch.

I think the real resolution is to remove the lifetime entirely, which I planned to do myself, once we merge the #2900 PR.

daxpedda:
What's the point of a separate method here anyway, I think we should just change the original method to return a `Result`.


> I think the real resolution is to remove the lifetime entirely, which I planned to do myself, once we merge the #2900 PR.

@kchibisov I wouldn't mind merging this until we get to it, otherwise we should just close this imo.

kchibisov:
@daxpedda I don't see a reason to merge this, given that I'll instantly resolve it once the #2900 is merged.

daxpedda:
Closing this then.

PR #2979, opened 2023-07-26
iOS: Always set timer when polling to avoid slow waking

### LINKED ISSUE #2167: Delay after first poll when resuming [iOS]

mockersf:
Fixes #2167

Adopted from #2168 to hope to get it ready before the next release.


mockersf:
As @HackerFoo, this works for me on iOS to avoid crashes.

madsmtm (approved):
Thanks for following up on this!

PR #2980, opened 2023-07-27
Add Occluded events based on foreground/background on iOS and MemoryWarning on iOS/Android

mockersf:
Adopted from #2144, and updated.

From https://github.com/rust-windowing/winit/pull/2144#issuecomment-1129990322 I'm not confident on adding events `Foreground` and `Background` to Android, they seem to be equivalent to `Suspend` / `Resume` where iOS split that in two steps.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -208,30 +208,6 @@ pub enum Event<T: 'static> {
     /// [`Suspended`]: Self::Suspended
     Resumed,
 
-    /// Emitted when the application will enter the foreground.
-    ///
-    /// ## iOS
-    ///
-    /// On iOS, the `Foreground` event is emitted in response to an [`applicationWillEnterForeground`]
-    /// callback which means the application should start preparing its data (according to the
-    /// [iOS application lifecycle]).
-    ///
-    /// [`applicationWillEnterForeground`]: https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1623076-applicationwillenterforeground
-    /// [iOS application lifecycle]: https://developer.apple.com/documentation/uikit/app_and_environment/managing_your_app_s_life_cycle
-    Foreground,
-
-    /// Emitted when the application has entered the background.
-    ///
-    /// ## iOS
-    ///
-    /// On iOS, the `Background` event is emitted in response to an [`applicationDidEnterBackground`]
-    /// callback which means the application should free resources (according to the
-    /// [iOS application lifecycle]).
-    ///
-    /// [`applicationDidEnterBackground`]: https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1622997-applicationdidenterbackground
-    /// [iOS application lifecycle]: https://developer.apple.com/documentation/uikit/app_and_environment/managing_your_app_s_life_cycle
-    Background,
-
     /// Emitted when the event loop is about to block and wait for new events.
     ///
     /// Most applications shouldn't need to hook into this event since there is no real relationship
@@ -284,8 +260,6 @@ impl<T> Event<T> {
             LoopExiting => Ok(LoopExiting),
             Suspended => Ok(Suspended),
             Resumed => Ok(Resumed),
-            Foreground => Ok(Foreground),
-            Background => Ok(Background),
             MemoryWarning => Ok(MemoryWarning),
         }
     }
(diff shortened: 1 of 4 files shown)


madsmtm:
That's really nice, thank you!

From my reading, to be maximally cross-platform compatible, it sounds like we should send an artificial `Background` event just before `Suspend` on Android, and an artificial `Foreground` just after `Resumed` (on all platforms, actually)?

madsmtm:
Hmm, actually it sounds like `applicationDidEnterBackground:` matches the semantics of the `Suspend` event on Android better than `applicationWillResignActive`?

madsmtm:
@rib: Is there an event on Android that requests the application to pause things (e.g. pause the game) and go into a low-power state, without necessarily suspending _everything_?

mockersf:
> From my reading, to be maximally cross-platform compatible, it sounds like we should send an artificial `Background` event just before `Suspend` on Android, and an artificial `Foreground` just after `Resumed` (on all platforms, actually)?

Is it OK if I add those artificial events on other platforms, with a TODO to move them to the correct events on Android for now?

daxpedda:
Web should probably use these for the `pagehide` and `pageshow` events, currently it uses `Occluded` instead.
Should we also emit `Occluded` with `Foreground` and `Background`?
Should this be folded into the new `WindowState` (https://github.com/rust-windowing/winit/pull/2929)?

kchibisov:
If the whole intent is to stop rendering based on foreground/background `Occluded` sounds like a better choice.

daxpedda:
My suggestion was to use **both**, `Occluded` and `Foreground`/`Background`.

Maybe I'm misunderstanding `Foreground`/`Background`, my assumption was that if something is in the background it's not visible, ergo occluded.

kchibisov:
I mean, what's the reason to add them, if `Occluded` could indicate `Foreground`/`Background`?

kchibisov:
@daxpedda it's simply a hint saying that you could stop rendering.

daxpedda:
> I mean, what's the reason to add them, if `Occluded` could indicate `Foreground`/`Background`?

> @daxpedda it's simply a hint saying that you could stop rendering.

I agree, that's why I am proposing to send both, because `Occluded` should not indicate anything else then to stop rendering.

kchibisov:
@daxpedda I'm just arguing against adding special events for `ios` where we already can use existing `Occluded` event.

mockersf:
`Occluded` is a `WindowEvent`, while `Foreground`/`Background` are application events. When dealing with them, you're not bound to any specific window, and putting them as `WindowEvent`would (in my opinion) break some of their meaning.

kchibisov:
>Occluded is a WindowEvent, while Foreground/Background are application events. When dealing with them, you're not bound to any specific window, and putting them as WindowEventwould (in my opinion) break some of their meaning.

>That said, that limit on iOS at least is quite blurry as it's not possible to have more than one window, so it may be ok to mix those there...

I mean, as long as window is not destroyed it's fine to use `Occluded` for that.

mockersf:
Moved to `WindowEvent::Occluded`

src/platform_impl/ios/view.rs
```diff
@@ -525,10 +525,46 @@ declare_class!(
...
+                        &*ptr
+                    };
+                    events.push(EventWrapper::StaticEvent(Event::WindowEvent {
+                        window_id: RootWindowId(window.id()),
+                        event: WindowEvent::Occluded(true),
+                    }));
+                }
+            }
+            let mtm = MainThreadMarker::new().unwrap();
+            app_state::handle_nonuser_events(mtm, events);
```
kchibisov, line 566: Could you extract the code into common function? The only difference is `true/false` to `Occluded`, really.

src/event.rs
```diff
@@ -225,6 +225,26 @@ pub enum Event<T: 'static> {
...
+    ///
+    /// ## iOS
+    ///
+    /// On iOS, the `MemoryWarning` event is emitted in response to an [`applicationDidReceiveMemoryWarning`]
+    /// callback. The application must free as much memory as possible or risk being terminated, see
+    /// [how to respond to memory warnings].
+    ///
+    /// [`applicationDidReceiveMemoryWarning`]: https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1623063-applicationdidreceivememorywarni
+    /// [how to respond to memory warnings]: https://developer.apple.com/documentation/uikit/app_and_environment/managing_your_app_s_life_cycle/responding_to_memory_warnings
+    MemoryWarning,
```
kchibisov, line 247: Should do a proper `## Platform-specific` section with a separate android/ios entries in it, saying that the rest(wayland, macOS) is not supported.

src/event.rs
```diff
@@ -533,8 +554,19 @@ pub enum WindowEvent {
     ///
     /// Platform-specific behavior:
     ///
+    /// ## iOS
```
kchibisov, line 557: Could you change the `Platform-specific behavior:` to `## Platform specific` and use `### iOS` while at it?

kchibisov (changes requested):

src/event.rs
mockersf: used `## Platform-specific` to be in line with the others

kchibisov (approved):

PR #2981, opened 2023-07-28
Remove `lifetime` from the `Event`

### LINKED ISSUE #1387: Event now takes a lifetime

kchibisov:
Lifetimes don't work nicely when dealing with multithreaded environments in the current design of the existing winit's event handling model, so remove it in favor of `Weak` fences passed to client, so they could try to update the size.

Fixes #1387.

--

The changes are uncontroversial and basically removed one concept with another. If backend maintainers think that they can do better now due to event being without a lifetime they should do that in the follow up PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ And please only add new entries to the top of this list, right below the `# Unre
 # Unreleased
 
 - On Windows, added `WindowBuilderExtWindows::with_class_name` to customize the internal class name.
+- **Breaking:** Remove lifetime parameter from `Event` and `WindowEvent`.
+- **Breaking:** `ScaleFactorChanged` now contains a writer instead of a reference to update inner size.
 - On iOS, always wake the event loop when transitioning from `ControlFlow::Poll` to `ControlFlow::Poll`.
 - **Breaking:** `ActivationTokenDone` event which could be requested with the new `startup_notify` module, see its docs for more.
 - On Wayland, make double clicking and moving the CSD frame more reliable.
--- a/examples/child_window.rs
+++ b/examples/child_window.rs
@@ -46,7 +46,7 @@ fn main() -> Result<(), impl std::error::Error> {
 
     println!("parent window: {parent_window:?})");
 
-    event_loop.run(move |event: Event<'_, ()>, event_loop, control_flow| {
+    event_loop.run(move |event: Event<()>, event_loop, control_flow| {
         *control_flow = ControlFlow::Wait;
 
         if let Event::WindowEvent { event, window_id } = event {
--- a/src/error.rs
+++ b/src/error.rs
@@ -8,6 +8,8 @@ use crate::platform_impl;
 pub enum ExternalError {
     /// The operation is not supported by the backend.
     NotSupported(NotSupportedError),
+    /// The operation was ignored.
+    Ignored,
     /// The OS cannot perform the operation.
     Os(OsError),
 }
@@ -74,6 +76,7 @@ impl fmt::Display for ExternalError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
         match self {
             ExternalError::NotSupported(e) => e.fmt(f),
+            ExternalError::Ignored => write!(f, "Operation was ignored"),
             ExternalError::Os(e) => e.fmt(f),
         }
     }
--- a/src/event.rs
+++ b/src/event.rs
@@ -33,13 +33,16 @@
 //!
 //! [`EventLoop::run(...)`]: crate::event_loop::EventLoop::run
 //! [`ControlFlow::WaitUntil`]: crate::event_loop::ControlFlow::WaitUntil
-use smol_str::SmolStr;
 use std::path::PathBuf;
+use std::sync::{Mutex, Weak};
 #[cfg(not(wasm_platform))]
 use std::time::Instant;
+
+use smol_str::SmolStr;
 #[cfg(wasm_platform)]
 use web_time::Instant;
 
+use crate::error::ExternalError;
 #[cfg(doc)]
 use crate::window::Window;
 use crate::{
@@ -53,8 +56,8 @@ use crate::{
 /// Describes a generic event.
 ///
 /// See the module-level docs for more information on the event loop manages each event.
-#[derive(Debug, PartialEq)]
-pub enum Event<'a, T: 'static> {
+#[derive(Debug, Clone, PartialEq)]
+pub enum Event<T: 'static> {
     /// Emitted when new events arrive from the OS to be processed.
     ///
     /// This event type is useful as a place to put code that should be done before you start
@@ -66,7 +69,7 @@ pub enum Event<'a, T: 'static> {
     /// Emitted when the OS sends an event to a winit window.
     WindowEvent {
         window_id: WindowId,
-        event: WindowEvent<'a>,
+        event: WindowEvent,
     },
 
     /// Emitted when the OS sends an event to a device.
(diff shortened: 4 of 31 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- **Breaking:** Remove `PartialEq` implementation from the event.
+- **Breaking:** Remove `lifetime` from the event.
```
madsmtm, line 12:
```suggestion
- **Breaking:** Remove `PartialEq` implementation from `Event` and `WindowEvent`.
- **Breaking:** Remove lifetime parameter from `Event` and `WindowEvent`.
```

src/event.rs
```diff
@@ -590,7 +548,9 @@ pub enum WindowEvent<'a> {
     /// For more information about DPI in general, see the [`dpi`](crate::dpi) module.
     ScaleFactorChanged {
         scale_factor: f64,
-        new_inner_size: &'a mut PhysicalSize<u32>,
+        /// The pointer could be upgraded successfully only from the callback for this event,
+        /// otherwise it'll always fail to do so.
+        new_inner_size: Weak<Mutex<PhysicalSize<u32>>>,
```
madsmtm, line 553: Missing changelog entry on this.

examples/child_window.rs
```diff
@@ -46,7 +46,7 @@ fn main() -> Result<(), impl std::error::Error> {
 
     println!("parent window: {parent_window:?})");
 
-    event_loop.run(move |event: Event<'_, ()>, event_loop, control_flow| {
+    event_loop.run(move |event: Event<()>, event_loop, control_flow| {
```
madsmtm, line 49: Do we have no examples that use `ScaleFactorChanged`?

That seems like an oversight, and makes it hard for us to judge how the API will be used by users.

src/event.rs
madsmtm, line 553: Might make sense to wrap this in a struct, to hide the specific `Weak` and `Mutex` implementation that we use?

src/event.rs
```diff
@@ -54,8 +56,8 @@ use crate::{
 /// Describes a generic event.
 ///
 /// See the module-level docs for more information on the event loop manages each event.
-#[derive(Debug, PartialEq)]
-pub enum Event<'a, T: 'static> {
+#[derive(Debug)]
```
madsmtm, line 59:
```suggestion
#[derive(Debug, Clone)]
```

madsmtm (changes requested):
I wouldn't say it's an "uncontroversial" change, see e.g. https://github.com/rust-windowing/winit/issues/939 where alternatives to a lifetime was discussed.

In any case, while I agree that the lifetime is not ideal from a user's perspective, it _does_ act as sort of "speedbump" for handling events outside of the event loop, which, even though most events don't have a lifetime, is generally not a good idea.

I.e. I think we should consider this change in that light as well, e.g. it's just not just about "what can users now do more easily", but also "what pitfalls will they now fall into more easily".

Approved the iOS and macOS impls.

kchibisov:
>I.e. I think we should consider this change in that light as well, e.g. it's just not just about "what can users now do more easily", but also "what pitfalls will they now fall into more easily".

In the current state of things this design is bad, because we tie event to the types it shouldn't be tied to. Yes this is sort of bad, but the lifetime is still there and we provided all the things to do event handling outside the event loop before with `Event::to_static`.

What I think we should change is to split the event callback handler into small ones, so we can have immediate actions encoded in a different way, like with the return type on the callback or some other handler structs. Events are just data in the end, they shouldn't be used to perform actions.

So the current design is bad, and we already have workaround for it, so it's just justifies its removal.

kchibisov:
>So I fear that in the meantime, more people will be incentivized to handle events outside the callback, which will make it even harder for them to transition to a more fully callback-based model.

They already do that with `Event::to_static`.

examples/child_window.rs
kchibisov: I guess the main case is to write when changing to some size the user wants so you don't have a window jumping in sizes, but not sure that we really want to demonstrate it.

src/error.rs
```diff
@@ -74,6 +76,7 @@ impl fmt::Display for ExternalError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
         match self {
             ExternalError::NotSupported(e) => e.fmt(f),
+            ExternalError::Ignored => write!(f, "Operation was ignored"),
```
madsmtm, line 79: Nit:

```suggestion
            ExternalError::Ignored => write!(f, "operation was ignored"),
```

src/event.rs
```diff
@@ -1301,3 +1062,36 @@ pub enum MouseScrollDelta {
     /// and move the content right and down (to reveal more things left and up).
     PixelDelta(PhysicalPosition<f64>),
 }
+
+/// Handle to synchroniously change the size of the window from the
+/// [`WindowEvent`].
+#[derive(Debug, Clone)]
+pub struct InnerSizeWriter {
```
madsmtm, line 1069: Perhaps this makes sense to put in the `dpi` module?

madsmtm (approved):

src/event.rs
kchibisov: It's a part of the event, I'm not sure it makes sense in other module.

src/error.rs
kchibisov: We're using upper case letters in this file for display issues, so if we want to change we need to do that all at once.

src/platform_impl/android/mod.rs
```diff
@@ -996,8 +1000,8 @@ pub struct MonitorHandle {
     app: AndroidApp,
 }
 impl PartialOrd for MonitorHandle {
-    fn partial_cmp(&self, _other: &Self) -> Option<std::cmp::Ordering> {
-        Some(std::cmp::Ordering::Equal)
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        Some(self.cmp(other))
     }
```
madsmtm, line 1005: This seems like an unintended change?

madsmtm:
I'll note down [@Osspial's positive attitude](https://github.com/rust-windowing/winit/pull/1456#pullrequestreview-412978566) towards the previous PR, that basically resolves my remaining concerns for why we shouldn't do this.

src/platform_impl/android/mod.rs
kchibisov: That was a clippy error that I just fixed transparently.

madsmtm: But it may be wrong now? Or at least the `Ord` implementation should also be fixed then.

kchibisov: Have you seen the change? It's an equivalent transformation, it was always `Eq` before, now I call `cmp` which is also always `Eq`. The `cmp` is a method on `Ord`.

madsmtm: Oh sorry, for some reason I had gotten it into my head that it was calling `PartialEq` on the `AndroidApp` field...

Issue #2464, opened 2022-09-02
macOS/iOS main thread safety


madsmtm:
macOS and iOS have the pervasive concept of "main thread safety", which essentially just means that every UI operation has to happen on the main thread. We have some support for this in `macos::util::async`, but this has been developed more on a case-by-case basis, which is not the way to do thread safety; remember, it is not just certain APIs that are thread unsafe, and the rest are fine, _every_ API is thread unsafe unless documented otherwise!

TODO
- [ ] `objc2`: Allow `Cell` in instance variables so that we don't need to use `&mut self`
- [ ] Make some mechanism so that every call on `WinitWindow` requires running on the main thread first.
  
  Idea (not very pretty, but would work pretty well):
  ```rust
  struct MainThreadOnly<T>(T);
  
  impl<T> MainThreadOnly<T> {
      fn new():
      
      fn run<R: Send>(f: impl FnOnce(T) -> R + Send) -> R {
          ...
      }
  }
  
  struct Window {
      window: MainThreadOnly<Id<WinitWindow, Shared>>,
  }
  
  impl Window {
      pub fn is_maximized(&self) -> bool {
          self.window.run(|window| {
              window.is_zoomed()
          })
      }

      fn set_maximized(&self, maximized: bool) {
          self.window.run(|window| {
              window.set_maximized(maximized);
          })
      }
  }
  ```
- [ ] ...

PR #2985, opened 2023-07-29
Improve thread safety

madsmtm:
Improve thread safety on macOS and iOS.

@daxpedda: This is using a generic mechanism that unifies it with the web backend - I think it's the right way forward, although it might be slightly less efficient (since it has to do the main thread check a bit more often).

Part of https://github.com/rust-windowing/winit/issues/2464.

  - [x] MacOS
  - [ ] Web
  - [ ] IOS


src/platform_impl/web/window.rs
```diff
@@ -60,42 +60,47 @@ impl Window {
...
+    pub(crate) fn maybe_queue_on_main(&self, f: impl FnOnce(&Inner) + Send + 'static) {
+        self.inner.dispatch(f)
+    }
+
+    // TODO: Find a way to make this non-static
+    pub(crate) fn maybe_wait_on_main<R: Send + 'static>(
+        &self,
+        f: impl FnOnce(&Inner) -> R + Send + 'static,
+    ) -> R {
+        self.inner.queue(f)
```
madsmtm, line 93: I am unsure if these two are the best abstractions?

Although slightly less efficient, maybe we should just always use `maybe_wait_on_main`, since that is the only real way to ensure that an action completed, and to propagate panics? I fear that the "fire and forget" style of programming that `maybe_queue_on_main` is will only lead to problems down the road.

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- Make iOS windows usable from other threads.
+- Refactor web backend thread safety slightly.
```
daxpedda, line 12: I think we tend to not document internal changes if they don't affect the user. Not sure though.

daxpedda (approved):
Wow, that's a nice idea.

Web looks good, the improvements you made are great.

src/platform_impl/web/window.rs
daxpedda: I don't mind not propagating the panics into the calling thread.

At least on Web we only panic if our assumptions are wrong, which should never happen unless Winit or the browser has a bug. Generally speaking panicking on Web right now is problematic, it should be considered unrecoverable until the exception handling proposal is implemented (which was recently done, but only for `no_std`, which we don't support for Web).

I do agree though that this isn't ideal, but I would like to avoid blocking the thread when we can, so I would argue it's worth it. Though we could at least add some documentation noting this behavior somewhere. I suggest we add it to the `Window` type docs, where we should have noted that some of these function can block in a multi-threaded environment anyway.

src/platform_impl/web/window.rs
```diff
@@ -60,42 +59,47 @@ impl Window {
...
+        })
+    }
 
-        Ok(window)
+    pub(crate) fn maybe_queue_on_main(&self, f: impl FnOnce(&Inner) + Send + 'static) {
+        self.inner.dispatch(f)
+    }
+
+    // TODO: Find a way to make this non-static
+    pub(crate) fn maybe_wait_on_main<R: Send + 'static>(
```
daxpedda, line 88: This should be straightforward to fix, just add some unsafe to modify the lifetime. Let me know and I can add it.

madsmtm: Please do, I'm not entirely confident that I won't cause unsoundness here.

daxpedda: Hm, I just realized that this will cause `MainThreadSafe::new()` and `Channel::new()` to be `unsafe`, because the `handler` could do something funny with `E`, which isn't actually `'static`.

Is it actually necessary to remove `'static` here?

daxpedda: Forget I said anything.

src/platform_impl/web/window.rs
madsmtm: I've added some (kinda vague) docs about this to `Window`, so that the user is aware that things might block, but not enough that we can't change internally which things block.

Please proof-read it!

src/window.rs
```diff
@@ -20,6 +20,18 @@ pub use cursor_icon::{CursorIcon, ParseError as CursorIconParseError};
 
 /// Represents a window.
 ///
+///
+/// # Threading
+///
+/// This is `Send + Sync`, meaning that it can be freely used from other
+/// threads.
+///
+/// However, some platforms (macOS, web and iOS) only allow user interface
```
daxpedda, line 29:
```suggestion
/// However, some platforms (macOS, Web and iOS) only allow user interface
```

kchibisov (approved):
I've decided to not add Wayland part as of now.

PR #2989, opened 2023-07-31
Merge stable, nightly and MSRV tests into a single CI job

madsmtm:
Shoul improve CI speed, since we have a limited number of concurrent runners, and because I've opted to do less work in those configurations.

One downside is that we can't really do the "not required" on nightly/MSRV builds any more, every check will now have to be required.

Builds upon https://github.com/rust-windowing/winit/pull/2988.


madsmtm:
Hmm, it's like roughly the same speed, if perhaps a bit slower, since even though we're taking up less total compute time, we can also use less concurrency.

PR #2991, opened 2023-07-31
On X11, remove the now-unrefrenced events.rs source file

MarijnS95:
#2662 renamed `VirtualKeyCode` to `Key` yet references to the former type still exist in `src/platform_impl/linux/x11/events.rs`.  As it turns out the `mod events;` in `x11/mod.rs` was removed in the same PR, but the file accidentally stuck around without being referenced anywhere else.


kchibisov:
That's quite a dead code.

kchibisov (approved):

Issue #1823, opened 2021-01-10
SoftInput (Software Keyboard) support


semtexzv:
On android, the application should be able to request showing & hiding of soft input.

This is available through ndk_glue for android. 

But this strangely feels like functionality that should be attached to window.
eg. for desktop we have cursor grabbing, for mobile, having methods for showing keyboard seems similar.

torokati44:
Yeah, this would be really useful. Especially since so far I have been completely unable to pop up the virtual keyboard on `winit`'s `NativeActivity`.
I does not appear when adding`android:windowSoftInputMode="stateAlwaysVisible|adjustResize"` to the `<activity>` entry in `AndroidManifest.xml`, nor when calling `ndk_glue::native_activity().show_soft_input(true);` from Rust.

Fancyflame:
So... does any guy have any ideas? `ndk_glue::native_activity()` is deprecated, it says using `ndk_context::android_context().vm()` instead, but what next? I'm a noob at Android programming.

msiglreith:
Not sure what's up with the NDK function but here is an example going trough jni: https://github.com/msiglreith/pixels-android/blob/master/src/lib.rs#L30 which shows the keyboard when tapping the screen.
It uses current `main` branch as virtual keycode have just been added recently for Android, otherwise it will just return empty ):

Fancyflame:
This example can work on my device. But do you mean this method has a compatibility problem?

msiglreith:
Sry, I referred to the Android winit platform implementation not Android itself: https://github.com/rust-windowing/winit/pull/2226
Should work with older Android versions as well

Fancyflame:
Thanks. That's okay if it only needs update winit to the latest version. 😃
To be honest, I cannot understand what you wrote due to my lack of knowledge about Android and Java.... Can I copy your implementation to my project? I'll mark your repo.

Fancyflame:
@msiglreith Then, how to hide it?

msiglreith:
> Can I copy your implementation to my project?

Feel free. I will probably look into opening a PR for this in the next days if there are no further issues with the JNI implementation.

> Then, how to hide it?

I pushed a small change in the meantime to the repository from above which implements it (hides when tapping the screen again).

Fancyflame:
@msiglreith Thanks for your example. I'm sorry for bothering you again, but... I have another two questions: 
[...]

PR #2993, opened 2023-08-01
Add android ime support

lucasmerlin:
This adds Support for opening the Soft Keyboard on android and passing through `TextInputState` events to the application. 
TextInputState contains the currently edited text as string, the current selection / cursor range and a optional compose region. 
This enables applications to support text input on android including autocomplete and autocorrect. 
While the TextInputState would currently be exclusive to android, I think it should be possible to implement the same api on iOS so there is a unified api for soft keyboards.

I implemented this as a new event because I think the current Ime event wouldn't work well for mobile text input, but it might also be possible to make it work with some tweaks. There would at least need to be some way to set the current text field content, so android has some context on what to predict for autocomplete.

I implemented this api for egui here: https://github.com/lucasmerlin/egui/commit/d47677a3d49a016a176b12a15a1f6beceff94530

<details><summary>See it in action</summary>
<p>

https://user-images.githubusercontent.com/8009393/251972889-5ac5299f-16d2-429e-899c-6d1e8d31987d.mp4

</p>
</details> 

A disadvantage of this api is, that it'd probably be difficult if not impossible to reliably edit rich text with this. The alternative would be to implement some kind of shared abstraction over [InputConnection](https://developer.android.com/reference/android/view/inputmethod/InputConnection) and [UITextInput](https://developer.apple.com/documentation/uikit/uitextinput) but these apis are a lot more complicated and for android it'd be difficult to implement it with the game-activity approach.

I'm bad at naming things, so there is probably a better name for the event than `TextInputState`.

Some other things missing to allow a great keyboard experience on mobile devices:
- [ ] Inset events / some way to query the current insets, so one can make sure the text field is not covered by the keyboard
- [ ] Support setting keyboard flags, for e.g. number / password / email input, enabling / disabling autocorrect, toggling capitalization for sentences

Part of https://github.com/rust-windowing/winit/issues/1823.

---

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -886,19 +886,21 @@ pub enum Ime {
     /// Notifies when text should be inserted into the editor widget.
     ///
     /// Right before this event winit will send empty [`Self::Preedit`] event.
-    Commit(String),
-
-    /// Notifies when the complete text should be replaced.
-    /// This event should be used in combination with
-    /// [`Window::set_ime_surrounding_text`] to set the initial text of the
-    /// currently selected input field.
-    ///
-    /// ## Platform-specific
-    /// This will only be fired on **Android**.
-    Replace {
-        text: String,
-        selection: (usize, usize),
+    Commit {
+        content: String,
+        /// If selection is Some, the selection / cursor position should be updated after
+        /// placing the `content`.
+        selection: Option<(usize, usize)>,
+        /// If compose_region is Some, the compose region should be updated after replacing the
+        /// text.
         compose_region: Option<(usize, usize)>,
+    },
+
+    /// Notifies when the text around the cursor should be deleted.
+    /// If `before_length` and `after_length` are [usize::MAX], the entire text should be deleted.
+    DeleteSurroundingText {
+        before_length: usize,
+        after_length: usize,
     },
 
     /// Notifies when the IME was disabled.
--- a/src/event.rs
+++ b/src/event.rs
@@ -378,8 +378,6 @@ pub enum WindowEvent {
     /// - **iOS / Android / Web / Orbital:** Unsupported.
     Ime(Ime),
 
-    TextInputState(TextInputState),
-
     /// The cursor has moved on the window.
     ///
     /// ## Platform-specific
@@ -890,6 +888,19 @@ pub enum Ime {
     /// Right before this event winit will send empty [`Self::Preedit`] event.
     Commit(String),
 
+    /// Notifies when the complete text should be replaced.
+    /// This event should be used in combination with
+    /// [`Window::set_ime_surrounding_text`] to set the initial text of the
+    /// currently selected input field.
+    ///
+    /// ## Platform-specific
+    /// This will only be fired on **Android**.
+    Replace {
+        text: String,
+        selection: (usize, usize),
+        compose_region: Option<(usize, usize)>,
+    },
+
     /// Notifies when the IME was disabled.
     ///
     /// After receiving this event you won't get any more [`Preedit`](Self::Preedit) or
@@ -1096,28 +1107,4 @@ impl PartialEq for InnerSizeWriter {
     fn eq(&self, other: &Self) -> bool {
         self.new_inner_size.as_ptr() == other.new_inner_size.as_ptr()
     }
-}
-
-/// This struct holds a span within a region of text from `start` (inclusive) to
-/// `end` (exclusive).
-///
-/// An empty span or cursor position is specified with `start == end`.
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
-#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-pub struct TextSpan {
-    /// The start of the span (inclusive)
-    pub start: usize,
-
-    /// The end of the span (exclusive)
-    pub end: usize,
-}
-
-#[derive(Debug, Clone, PartialEq, Eq)]
-#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-pub struct TextInputState {
-    pub text: String,
-    /// A selection defined on the text.
-    pub selection: TextSpan,
-    /// A composing region defined on the text.
-    pub compose_region: Option<TextSpan>,
 }
--- a/src/event.rs
+++ b/src/event.rs
@@ -378,6 +378,8 @@ pub enum WindowEvent {
     /// - **iOS / Android / Web / Orbital:** Unsupported.
     Ime(Ime),
 
+    TextInputState(TextInputState),
+
     /// The cursor has moved on the window.
     ///
     /// ## Platform-specific
@@ -1094,4 +1096,28 @@ impl PartialEq for InnerSizeWriter {
     fn eq(&self, other: &Self) -> bool {
         self.new_inner_size.as_ptr() == other.new_inner_size.as_ptr()
     }
+}
+
+/// This struct holds a span within a region of text from `start` (inclusive) to
+/// `end` (exclusive).
+///
+/// An empty span or cursor position is specified with `start == end`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct TextSpan {
+    /// The start of the span (inclusive)
+    pub start: usize,
+
+    /// The end of the span (exclusive)
+    pub end: usize,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct TextInputState {
+    pub text: String,
+    /// A selection defined on the text.
+    pub selection: TextSpan,
+    /// A composing region defined on the text.
+    pub compose_region: Option<TextSpan>,
 }
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -470,22 +470,33 @@ impl<T: 'static> EventLoop<T> {
                     }
                 }
                 InputEvent::TextEvent(ime_state) => {
-                    let event = event::Event::WindowEvent {
-                        window_id: window::WindowId(WindowId),
-                        event: event::WindowEvent::Ime(
-                            event::Ime::Replace {
-                                text: ime_state.text.to_string(),
-                                selection: (ime_state.selection.start, ime_state.selection.end),
-                                compose_region: ime_state.compose_region.map(|region| (region.start, region.end))
-                            }
-                        )
-                    };
-                    sticky_exit_callback(
-                        event,
-                        self.window_target(),
-                        control_flow,
-                        callback
-                    );
+                    let events = [
+                        // Send a preedit event so the application knows to expect a commit event
+                        event::Ime::Preedit("".to_string(), None),
+                        // Delete all of the current text
+                        event::Ime::DeleteSurroundingText {
+                            before_length: usize::MAX,
+                            after_length: usize::MAX,
+                        },
+                        // Replace the previously deleted text with our updated text, and set the cursor and compose region
+                        event::Ime::Commit {
+                            content: ime_state.text.to_string(),
+                            selection: Some((ime_state.selection.start, ime_state.selection.end)),
+                            compose_region: ime_state.compose_region.map(|region| (region.start, region.end)),
+                        }
+                    ];
+
+                    events.into_iter().for_each(|event| {
+                        sticky_exit_callback(
+                            event::Event::WindowEvent {
+                                window_id: window::WindowId(WindowId),
+                                event: event::WindowEvent::Ime(event),
+                            },
+                            self.window_target(),
+                            control_flow,
+                            callback,
+                        );
+                    });
                 }
                 _ => {
                     warn!("Unknown android_activity input event {event:?}")
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -472,19 +472,11 @@ impl<T: 'static> EventLoop<T> {
                 InputEvent::TextEvent(ime_state) => {
                     let event = event::Event::WindowEvent {
                         window_id: window::WindowId(WindowId),
-                        event: event::WindowEvent::TextInputState(
-                            TextInputState {
-                                text: ime_state.text.to_owned(),
-                                selection: TextSpan {
-                                    start: ime_state.selection.start,
-                                    end: ime_state.selection.end,
-                                },
-                                compose_region: ime_state
-                                    .compose_region
-                                    .map(|region| TextSpan {
-                                        start: region.start,
-                                        end: region.end,
-                                    }),
+                        event: event::WindowEvent::Ime(
+                            event::Ime::Replace {
+                                text: ime_state.text.to_string(),
+                                selection: (ime_state.selection.start, ime_state.selection.end),
+                                compose_region: ime_state.compose_region.map(|region| (region.start, region.end))
                             }
                         )
                     };
@@ -927,30 +919,26 @@ impl Window {
 
     pub fn set_ime_cursor_area(&self, _position: Position, _size: Size) {}
 
-    pub fn set_ime_allowed(&self, _allowed: bool) {}
+    pub fn set_ime_allowed(&self, allowed: bool) {
+        if allowed {
+            self.app.show_soft_input(true);
+        } else {
+            self.app.hide_soft_input(true);
+        }
+    }
 
     pub fn set_ime_purpose(&self, _purpose: ImePurpose) {}
 
-    pub fn begin_ime_input(&self) {
-        self.app.show_soft_input(true);
-    }
-
-    pub fn end_ime_input(&self) {
-        self.app.hide_soft_input(true);
-    }
-
-    pub fn set_text_input_state(&self, state: TextInputState) {
-        self.app.set_text_input_state(android_activity::input::TextInputState {
-            text: state.text,
-            selection: android_activity::input::TextSpan {
-                start: state.selection.start,
-                end: state.selection.end,
-            },
-            compose_region: state.compose_region.map(|region| android_activity::input::TextSpan {
-                start: region.start,
-                end: region.end,
-            }),
-        });
+    pub fn set_ime_surrounding_text(&self, text: String, selection: (usize, usize)) {
+        self.app
+            .set_text_input_state(android_activity::input::TextInputState {
+                text,
+                selection: android_activity::input::TextSpan {
+                    start: selection.0,
+                    end: selection.1,
+                },
+                compose_region: None,
+            });
     }
 
     pub fn focus_window(&self) {}
@@ -1035,7 +1023,6 @@ impl Window {
 pub struct OsError;
 
 use std::fmt::{self, Display, Formatter};
-use crate::event::{TextInputState, TextSpan};
 
 impl Display for OsError {
     fn fmt(&self, fmt: &mut Formatter<'_>) -> Result<(), fmt::Error> {
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -469,6 +469,32 @@ impl<T: 'static> EventLoop<T> {
                         }
                     }
                 }
+                InputEvent::TextEvent(ime_state) => {
+                    let event = event::Event::WindowEvent {
+                        window_id: window::WindowId(WindowId),
+                        event: event::WindowEvent::TextInputState(
+                            TextInputState {
+                                text: ime_state.text.to_owned(),
+                                selection: TextSpan {
+                                    start: ime_state.selection.start,
+                                    end: ime_state.selection.end,
+                                },
+                                compose_region: ime_state
+                                    .compose_region
+                                    .map(|region| TextSpan {
+                                        start: region.start,
+                                        end: region.end,
+                                    }),
+                            }
+                        )
+                    };
+                    sticky_exit_callback(
+                        event,
+                        self.window_target(),
+                        control_flow,
+                        callback
+                    );
+                }
                 _ => {
                     warn!("Unknown android_activity input event {event:?}")
                 }
@@ -905,6 +931,28 @@ impl Window {
 
     pub fn set_ime_purpose(&self, _purpose: ImePurpose) {}
 
+    pub fn begin_ime_input(&self) {
+        self.app.show_soft_input(true);
+    }
+
+    pub fn end_ime_input(&self) {
+        self.app.hide_soft_input(true);
+    }
+
+    pub fn set_text_input_state(&self, state: TextInputState) {
+        self.app.set_text_input_state(android_activity::input::TextInputState {
+            text: state.text,
+            selection: android_activity::input::TextSpan {
+                start: state.selection.start,
+                end: state.selection.end,
+            },
+            compose_region: state.compose_region.map(|region| android_activity::input::TextSpan {
+                start: region.start,
+                end: region.end,
+            }),
+        });
+    }
+
     pub fn focus_window(&self) {}
 
     pub fn request_user_attention(&self, _request_type: Option<window::UserAttentionType>) {}
@@ -987,6 +1035,8 @@ impl Window {
 pub struct OsError;
 
 use std::fmt::{self, Display, Formatter};
+use crate::event::{TextInputState, TextSpan};
+
 impl Display for OsError {
     fn fmt(&self, fmt: &mut Formatter<'_>) -> Result<(), fmt::Error> {
         write!(fmt, "Android OS Error")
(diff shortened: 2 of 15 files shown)


src/window.rs
```diff
@@ -1111,6 +1112,23 @@ impl Window {
...
+
+    /// Hides the IME input (soft keyboard).
+    #[inline]
+    pub fn end_ime_input(&self) {
+        self.window.end_ime_input();
+    }
+
+    pub fn set_text_input_state(&self, state: TextInputState) {
+        self.window.set_text_input_state(state);
+    }
```
kchibisov, line 1130: Could you get a bit more details on what this particular function tries to do? In general, that's true that we'd nee a way to set an IME state, like position of the ime, surrounding text, etc.

I think what we probably should do is to have a `set_ime_state(state: Option<ImeState>)` where you can set that information + position of the input method, content hint, and similar.

src/event.rs
```diff
@@ -1095,3 +1097,27 @@ impl PartialEq for InnerSizeWriter {
...
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct TextInputState {
+    pub text: String,
+    /// A selection defined on the text.
+    pub selection: TextSpan,
+    /// A composing region defined on the text.
+    pub compose_region: Option<TextSpan>,
+}
```
kchibisov, line 1123: That all looks exactly like `Preedit` event, so better to remove it.

src/event.rs
```diff
@@ -378,6 +378,8 @@ pub enum WindowEvent {
     /// - **iOS / Android / Web / Orbital:** Unsupported.
     Ime(Ime),
 
+    TextInputState(TextInputState),
+
```
kchibisov, line 382: No need for a separate event like that, you should improve the `Ime` event.

kchibisov (changes requested):
Thanks for your work on that, even though I'm not familiar with android, but I do use IME with winit and would like to see it improved.

I don't like that we have a separate event for no real reason and that we have IME separate with your approach, which shouldn't be the case.

Could you look at https://wayland.app/protocols/text-input-unstable-v3#zwp_text_input_v3:request:set_surrounding_text and https://wayland.app/protocols/text-input-unstable-v3#zwp_text_input_v3:event:preedit_string ?

I think it's pretty much what you've added, except for the `compose` key? We could extent the `Preedit` API if it matters or just adjust `IME` API to better account for Android here.

In general, if you'd need help with other platforms I could implement some of them (Wayland/macOS/X11/and maybe Windows).

lucasmerlin:
Thanks for the review @kchibisov!
I see how having an additional api is not ideal. I'll try to get it working with the existing Ime event.

kchibisov:
>I see how having an additional api is not ideal. I'll try to get it working with the existing Ime event.

I'm just saying that nothing stops us from extending the existing event, especially given that what you've added looks similar to it.

src/window.rs
```diff
@@ -1111,6 +1112,23 @@ impl Window {
...
+    #[inline]
+    pub fn begin_ime_input(&self) {
+        self.window.begin_ime_input();
+    }
+
+    /// Hides the IME input (soft keyboard).
+    #[inline]
+    pub fn end_ime_input(&self) {
+        self.window.end_ime_input();
+    }
```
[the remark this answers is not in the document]
lucasmerlin: Ah, egui just calls set_ime_allowed once on application start, I think that confused me. I've changed this in my egui branch and removed the begin_ime_input and end_ime_input functions

src/window.rs
lucasmerlin: I've replaced this with a set_ime_surrounding_text function similar to the one from wayland you sent.

kchibisov: Does android have a notion of atomic state changes? Because wayland has for example, and using some big `InputState` would be preferred for it, but I can refactor it myself later.

lucasmerlin:
I've updated my code and added a Ime::Replace event. It looks pretty similar to the Preedit event but works differently: When this event is called the whole text (the one that was previously set with set_surrounding_text) should be replaced with the updated text.

I can't use the Preedit and Commit events, since the android game text input has no such thing as a commit. Whenever the user enters text, we get the complete updated text and a compose and selection region. One could maybe estimate whether a commit has happened based on the existence of a compose region and by diffing the updated text but I don't think that'd work reliably.

src/window.rs
lucasmerlin: I'm not sure what you mean, but I'd guess not. The android game text input library this relies on is quite simple. The idea is that this is called once when a text field is first focused and when the selection gets updated by the user.

kchibisov:
>I've updated my code and added a Ime::Replace event. It looks pretty similar to the Preedit event but works differently: When this event is called the whole text (the one that was previously set with set_surrounding_text) should be replaced with the updated text.

On Wayland we have a similar concept, it's called https://wayland.app/protocols/text-input-unstable-v3#zwp_text_input_v3:event:delete_surrounding_text .

Besides, users can't rely on the replace event alone, we need `Commit` event when the input is done into the editor. Maybe you could adapt the technique from the Wayland protocol, which looks similar to what you're doing? Read the https://wayland.app/protocols/text-input-unstable-v3#zwp_text_input_v3:event:done it should provide an algorithm to process events, which we can adopt on Android.

The key thing here is that `Commit` is mandatory for the work cross platform, Preedit is not that useful, but you can mark the `selection` with it.

Basically you should send `Commit` each time the text is written into the widget, if each input implies text being written into the keyboard widget, then you should send commit for each update. The users on the other side, should reply with `set_surrounding_text` in response to `Commit`, so the `Replace`/delete surrounding text could work. Some clients for example, can't work with surrounding text, because they forward the input into some side channel (think terminal emulator).

However, maybe Android should `Commit` the text when the keyboard is being hidden or something like that?

lucasmerlin:
I cloud in theory do something like this, basically diff the current and previous TextInputState to guess whether there should be a preedit or commit event: 
```rust
let events = if let Some(compose_region) = &ime_state.compose_region {
    vec![event::Ime::Preedit(ime_state.text[compose_region.start..compose_region.end].to_string(), Some((compose_region.start, compose_region.end)))]
} else {
    if let Some(previous_text_input_state) = &self.previous_text_input_state {
        if let Some(prev_compose_region) = &previous_text_input_state.compose_region {
            vec![
                event::Ime::Preedit("".to_string(), None),
                event::Ime::Commit(previous_text_input_state.text[prev_compose_region.start..prev_compose_region.end].to_string()),
            ]
        } else {
            vec![]
        }
    } else {
        vec![]
    }
};
```
But I don't think this would ever work reliably, no matter how many edgecases are handled.

That's why I added the Replace event, which would replace the Commit event for android. If the application doesn't handle the Replace event, it won't have keyboard input support on android. Maybe Replace is a bad name, I guess it could be called CommitReplace or something like that. Since this event would likely be android specific, maybe it could also be a android exclusive platform extension, if there is such a thing for events? 

> However, maybe Android should `Commit` the text when the keyboard is being hidden or something like that?

That would be an idea. While typing the whole text could be sent as a Preedit. This would mean the whole text is highlighted while typing though, instead of the current compose region. Also when the keyboard is hidden the text field has already lost focus, so it would probably not get the final commit event and the text would be dismissed? 

> Some clients for example, can't work with surrounding text, because they forward the input into some side channel (think terminal emulator).

I see how that would be a problem. I think if you don't set any keyboard flags(disable autocomplete/autocorrect), you should reliably be able to get the last typed character from the end of the string in the Ime::Replace event. By comparing the length you could see if the user pressed backspace. But that would also be a horrible hack. I have no better idea though.

kchibisov:
The issue with the `Replace` alone is that it's not clear what it tries to do, because this event doesn't map into any other platform directly. The idea behind `Replace` is likely fine, but what if don't set surrounding text? Could we just get a regular input and such?

`Preedit` is optional, only `Commit` is mandatory, because that's what is being inserted in the end. Surely you have some logic in `egui` to determine when the text is inserted? Do you just count the last state of the `Replace` as it? You should have some heuristic for that, I guess.

kchibisov:
Hm, maybe we could do that differently, could you compute the difference to replace based on the surrounding text provided by user, send replace event telling it to `Delete` the text from the start/end of the cursor, and then you send a `Commit` event resulting in effectively what the `Replace` does?

The only issue I think is the selection or something like that, but it probably should be transparent to the user?

It just looks like the `Replace` event you're proposing is basically what Wayland spec suggests (2-4 points)?

>1. Replace existing preedit string with the cursor. 2. Delete requested surrounding text. 3. Insert commit string with the cursor at its end. 4. Calculate surrounding text to send. 5. Insert new preedit text in cursor position. 6. Place cursor inside preedit text.

src/platform_impl/android/mod.rs
```diff
@@ -901,10 +919,28 @@ impl Window {
...
     pub fn set_ime_cursor_area(&self, _position: Position, _size: Size) {}
 
-    pub fn set_ime_allowed(&self, _allowed: bool) {}
+    pub fn set_ime_allowed(&self, allowed: bool) {
+        if allowed {
+            self.app.show_soft_input(true);
+        } else {
+            self.app.hide_soft_input(true);
+        }
+    }
```
kchibisov, line 928: Could you send this change separately?

lucasmerlin: I could make a separate PR with this change, it won't be very useful without the rest of these changes though, since android unfortunately doesn't send key events for the soft keyboard (See the first note [here](https://developer.android.com/develop/ui/views/touch-and-input/keyboard-input/commands)). So the keyboard would open but nothing will happen when the user starts typing.

kchibisov: But I think it does so with the game activity and recent changes merged here https://github.com/rust-windowing/winit/pull/3004 , no?

lucasmerlin: As far as I understand, those changes are only relevant to Hardware Keyboards (e.g. usb keyboards plugged in your phone), but maybe I'm missing something. @rib can you clarify?

rib: The unicode mapping support from #3004 is also for soft keyboards which also generate KeyEvents, though I do tend to think of IME vs KeyEvents handling as two entirely separate input handling modes (and that's the case for other platforms too) and I'm not sure that it'd be ideal to make this a common API for those separate modes.

Overloading `set_ime_allowed` for showing the keyboard in situations where you are happy to handle `KeyEvents` instead of IME events seems like it could lead to non-portable code because calling this (in the same app) for other platforms will enable IME handling / events which might not be desirable.

It seems like ideally there should be a separate API for being able to show/hide mobile soft keyboards which would have no effect on platforms that don't need a virtual keyboard (but they may otherwise have IME support that the app is not trying to enable).

kchibisov: Well, on Wayland it's tied to the `set_ime_allowed` as well, and it's tied to that on the protocol level though. It's not written in the spec, but that's the defacto how it works. I guess if you don't set surrounding text you won't get any sort of ime-ish input, no?

kchibisov: @rib you enable IME when you write text, on Android you can't really write text without a soft keyboard. The same applies for others, if you write text you generally want an IME enabled, unless you're using the keyboard input where any sort of IME input is disallowed.

Like for example a game using soft keyboard must provide me a way to input IME, because it's likely some field and what if I want to write some script?

On desktop you just sometimes want to remove IME when you actually game, but not chat, which is not the case on Android because you're using on screen controls.

On macOS for example not enabling IME won't let you type in some cases anything meaningful.

Also, enabling IME doesn't mean that you'll receive only IME events, you could still receive regular keys, it's just you indicate that you can receive extra IME events and you can provide a feedback for IME when needed.

lucasmerlin: > is also for soft keyboards which also generate KeyEvents

I just tried this, and none of the soft keyboards I have on my phone triggered any key events, only the TextEvent is triggered.

lucasmerlin:
> Hm, maybe we could do that differently, could you compute the difference to replace based on the surrounding text provided by user, send replace event telling it to Delete the text from the start/end of the cursor, and then you send a Commit event resulting in effectively what the Replace does?

Yeah, something like that could work. I could use the Delete Event to delete the current text and use commit to replace it with the newly received text. Then we would need a new event to update the cursor and compose position though, or else the cursor would always be shown at the end of the text. I could give this a go. In the end, this solution would still require a non standard event (update cursor position, haven't found anything like that in the wayland events) to be added though, so I'm not sure if this is better than the replace solution. 

I tried to implement a diffing algorithm, but I haven't gotten it to work. Since the android keyboard could in theory randomly replace text independent of the cursor position, I feel like a solution that uses diffing would always have some edge cases and bugs. 

@rib do you maybe have an idea how this could be solved? I think you originally also tried to use the existing ime events?

kchibisov:
We can commit the cursor along side the commit, it's not a big deal.

You also don't have to diff actually, you could just delete everything that was before and commit new, so it's exactly like replace, you just send 2 events.

rib:
Sorry for the delay following up here.

I do generally agree with the direction of trying to work with the Winit Ime events if we can make it work (based on adding a DeleteSurroundingText event).

It's unfortunate that the GameActivity interface we get doesn't give us more fine grained events but the way Android's InputConnection works isn't totally dissimilar to other IME protocols so I fee like we should try to avoid leaking the current limitations of GameActivity if at all possible.

It would be nice too if it's not necessary for a toolkit like Egui to have to treat Android as a special case when it comes to input method handling.

>> is also for soft keyboards which also generate KeyEvents
>
> I just tried this, and none of the soft keyboards I have on my phone triggered any key events, only the TextEvent is triggered.

Curious that you see _no_ KeyEvents at all.

I certainly see KeyEvents for at least basic virtual keyboard input (essentially just ascii characters).

We currently rely on these events in our engine while we haven't hooked up IME support yet but it seems like it may really depend on the implementation of different virtual keyboards as to what keys might generate KeyEvents.

I had actually thought more keys would be supported so we could rely on KeyEvents as a stop gap for a little longer but yeah it looks like it's not really something that can be relied on - especially if some virtual keyboards in fact don't emit any KeyEvents :/

src/platform_impl/android/mod.rs
```diff
@@ -469,6 +469,35 @@ impl<T: 'static> EventLoop<T> {
...
                 }
+                InputEvent::TextEvent(ime_state) => {
+                    let events = [
+                        // Send a preedit event so the application knows to expect a commit event
+                        event::Ime::Preedit("".to_string(), None),
+                        // Delete all of the current text
+                        event::Ime::DeleteSurroundingText {
+                            before_length: usize::MAX,
+                            after_length: usize::MAX,
+                        },
```
rib, line 480: Deleting surrounding text with usize::MAX might be a foot gun at some point.

IME edits are relative so the current cursor because conceptually they could be used within a large body of text (e.g. think 10k word Google Doc) so I'm a bit concerned that asking the toolkit to delete like this could potentially result in it deleting someone's larger document unless the toolkit additionally tracks the bounds of the surrounding context it has give Winit and treats that like a sandbox (but that's not currently clearly specified afik).

Maybe we can reliably track the size of the current text so we can more precisely delete just the text from the last update?

src/event.rs
```diff
@@ -886,7 +886,22 @@ pub enum Ime {
...
+        /// text.
+        compose_region: Option<(usize, usize)>,
+    },
+
+    /// Notifies when the text around the cursor should be deleted.
+    /// If `before_length` and `after_length` are [usize::MAX], the entire text should be deleted.
+    DeleteSurroundingText {
+        before_length: usize,
+        after_length: usize,
+    },
```
rib, line 904: Now that we're exposing a selection region it seems like we need to clarify what the before/after are relative to.

On Android the before would be relative to the start of the selection and the after is relative to the end of the selection.

The Winit docs don't currently say whether we're following the Android notion that the "cursor" is just a special case of a selection where start == end.

With that model though we also need to consider how we will delete / update the selection region?

With the InputConnection API there is a separate setSelection that would make it possible to set the selection start==end for being able to delete all the text.

I'm currently thinking we might need to add a `SetSelection` event for being able to just update the start/end indices of the selection.

src/platform_impl/android/mod.rs
kchibisov: I agree with that, doing delete all is sort of weird. Given that it's all about the feedback loop with `setSurroundingText` request, maybe we can base everything on it, since
IME shouldn't touch anything outside of the user set surrounding text, no?

src/event.rs
kchibisov: >With the InputConnection API there is a separate setSelection that would make it possible to set the selection start==end for being able to delete all the text.

>I'm currently thinking we might need to add a SetSelection event for being able to just update the start/end indices of the selection.


The selection should be mapped to the concept of the cursor, so it's effectively the same as in preedit. The `Commit` could set the cursor as well, but I guess the virtual keyboard could move the cursor on its own? Yeah, it  would be fine to have a separate method to advance the selection like you suggest, but we need to make it general on what selection is and probably encapsulate it into a separate type `ImeSelection` or something.

src/event.rs
```diff
@@ -886,7 +886,22 @@ pub enum Ime {
...
+        /// If compose_region is Some, the compose region should be updated after replacing the
+        /// text.
+        compose_region: Option<(usize, usize)>,
+    },
+
+    /// Notifies when the text around the cursor should be deleted.
+    /// If `before_length` and `after_length` are [usize::MAX], the entire text should be deleted.
+    DeleteSurroundingText {
+        before_length: usize,
+        after_length: usize,
```
kchibisov, line 903: We must specify the unit in which those are, because it's not clear whether it's a byte or unicode codepoint based indecies. Some API we have already is in bytes, not code points. So logically this API should be in bytes as well. Like for example Wayland uses bytes for that as well, and we can't really convert to code points in wayland case, because we simply don't know anything about the text around.

Maybe we should have a variant like `enum Amount { Byte(usize), Codepoint(usize) }`, so the client can deal with all of that?

src/event.rs
```diff
@@ -886,7 +886,22 @@ pub enum Ime {
     /// Notifies when text should be inserted into the editor widget.
     ///
     /// Right before this event winit will send empty [`Self::Preedit`] event.
-    Commit(String),
+    Commit {
+        content: String,
+        /// If selection is Some, the selection / cursor position should be updated after
+        /// placing the `content`.
+        selection: Option<(usize, usize)>,
```
kchibisov, line 893: Probably having a separate `Selection` type at this point unified with `Preedit` would be the way forward. Like the selection is used throughout the docs consistently, we should also decide on metrics we use and be explicit, I think on android it's unicode codepoints, but everything else we have is bytes?

kchibisov (changes requested):
Besides we should also communicate the `Enabled/Disabled` state somehow, maybe we should just sent `Enabled` initially and call it a day?

src/event.rs
rib: It feels odd to expose the offsets as bytes. I can sort of see why the wayland protocol would use bytes but forcing toolkits like Egui to have to somehow deal with byte offsets within Rust Strings sounds like it would be horrible to deal with.

rib: The wayland backend should be able to map the byte offsets to the corresponding utf8 codepoint and can somehow treat it as an error / undefined behaviour if you ever get a byte offset into the middle of a utf8 character.

kchibisov: @rib it can't do so for delete surrounding text. Unless we stash surrounding text around. The rust string a sliced with bytes though, so it works pretty much ok.

src/event.rs
```diff
@@ -886,7 +886,22 @@ pub enum Ime {
...
+        /// If compose_region is Some, the compose region should be updated after replacing the
+        /// text.
+        compose_region: Option<(usize, usize)>,
```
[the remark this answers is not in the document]
lucasmerlin: Android shows a underline under the word thats currently being edited. That is the compose region. But since no other platform has this (even iOS doesn't seem to have something like this) I think we could also remove it.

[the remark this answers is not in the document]
kchibisov: Others usually have a preedit and the update is like pending to be inserted.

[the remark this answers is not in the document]
rib: I think Android's compose region is just different terminology for the "preedit" string.

The preedit string is the transient text that would be shown that may preview a tentative completion e.g. for kanji.

I think there are essentially three high level spans of text:
1. the "surrounding" text
2. the "preedit" text (or "compose region" for Android)
3. the "selection" or "cursor" (empty selection)

I think the main thing that's changing here is that we need to introduce the notion that there is surrounding text which Winit didn't previously expose - though this isn't something unique to Android either.

Android has a slightly more general decoupling of the selection/cursor and the preedit string - while other IMEs might only track a cursor (can be considered an empty selection) or the cursor/selection will always be a subset of the preedit string.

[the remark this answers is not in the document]
rib: In the context of the Winit public API we should probably rename "compose_region" here to "preedit_region" or "preedit_span" perhaps.

Conceptually this would be committing some surrounding text and also updating the span for the selection and the span for the preedit string in one event.

xorgy:
Hey @lucasmerlin just wanted to let you know that I am resurrecting this work. I'm going to split off a version without the API changes first. I'll try to make sure the commits retain your attribution. :+ )

lucasmerlin:
Awesome, good luck @xorgy!
I'm happy to help / review, and I'm still interested in landing this. Let me know if you need access to my repo so you can push to the branch of this PR, or if there is some other way to do this. Otherwise just open a new one, and I can close this one with a notice.

jb55:
I tried to follow the discussion, what's the next steps here? I would be happy to help since I am currently using it for my android egui app.

lucasmerlin:
@jb55 Maybe a you could try to get basic text entry working with winit's existing api. There wouldn't be autocomplete support but it would be a first step and maybe enough for some apps. 

Also the people from linebender are very interested in android keyboard support, theres a long discussion here: https://xi.zulipchat.com/#narrow/stream/351333-glazier/topic/Comparisons.20against.20Winit 

I think @xorgy also looked into this but I'm not sure if he made any progress

kchibisov:
The API could be extended, it's just certain aspects should be shared. Like the actual _insert to widget_ part should be shared, etc, so so users will write some kind of cross platform code.

ardocrat:
FYI with current implementation on Android you can not get input from non-english languages as keycodes, at Java side you can get them at `Activity::dispatchKeyEvent` callback, `characters` field of `KeyEvent`. Ideally will be to provide already converted text (also with pressed shift) at input event, like Android made for non-english language.

xorgy:
I have this now rebased
https://github.com/rust-windowing/winit/compare/master...xorgy:winit:android_ime_support

But Android 11 (API 30) and later make it more painful to get the soft keyboard to show up, as others have discovered.

xorgy:
So I have managed to get the soft keyboard to show up, but the only way I was able to do it was by calling into the WindowInsetsController with JNI. The good news for #1823 is that we are getting keyboard events from the IME;
the mixed news is that it doesn't show any complex IMEs currently (won't switch to them). It might be a matter of providing the surrounding text and other things.

lucasmerlin:
Closing this in favor of #3787 by @xorgy, which has a better approach

xorgy:
And just to clarify, #3787 only opens the soft keyboard. Real IME on Android is looking like it will require specific support from the `Activity`, or at very least, the introduction of a View subclass (either when building the Activity, or possibly injected with runtime dex loading).

MarijnS95:
> or at very least, the introduction of a View subclass (either when building the Activity, or possibly injected with runtime dex loading).

Fyi this is actively beeing prototyped and worked on. Will be in `xbuild` first but can easily be ported to `cargo-apk` too.

kchibisov:
Nothing stops winit form using the `xbuild` btw, it's just we had issues with it once we tried to switch the CI.

MarijnS95:
I don't think `winit` must absolutely migrate to `xbuild` right now, just saying that I might be working on some solutions to make embedding Java with otherwise-native apps a bit nicer and easier :)

rib:
> Closing this in favor of #3787 by @xorgy, which has a better approach

I'm not sure this PR should have been closed, while it looks like #3787 was about mainly about showing the soft keyboard but not about generally adding IME support.

This PR still has useful discussion that was a bit more tricky to track down since this was closed.

That said though, I've just opened a different draft PR, that goes for a more constrained approach to enabling support for Winit's existing Ime events, even though they aren't a perfect match for Android: https://github.com/rust-windowing/winit/pull/4451

I have only tested on winit 0.30 so far but it works to enable text input with Egui (0.33).

nicoburns:
I can confirm that on Winit `0.31-beta.2` the soft keyboard correctly opens and closes in response `set_ime_allowed`, but at least on my setup (I suspect it may be keyboard-app dependent), typing on the keyboard doesn't actually generate `WindowEvent`s. Tested on Samsung S23 with Android 16 and `GameActivity` (with `NativeActivity` the software keyboard does not show at all).

Issue #2913, opened 2023-06-27
Change with_parent_window or re-export RawWindowHandle


ecton:
I've been working on a little wrapper for winit, and I originally thought I would be able to avoid depending on raw-window-handle by giving users access to the winit Window which implements `RawWindowHandle`.

My wrapper I'm building is multi-threaded, and I can't send a winit WindowBuilder across threads due to some x11 types not being Send. So, I've recreated my own WindowBuilder, and in an effort to support `with_parent_window`, I realized I cannot actually `WindowBuilder::with_parent_window` providing it a `winit::window::Window`. Instead, I must explicitly use the `RawWindowHandle` trait. This forces my crate to have two dependencies, instead of just winit. This is obviously a very minor issue, but part of me wishes I didn't have to add this dependency.

In my opinion, offering a public API that can't be used without first adding another dependency is a friction point that should be avoided for end users of crates. I would suggest two solutions to this friction point:

- Re-export `RawWindowHandle` so that users of winit can call `raw_window_handle()` on `winit::window::Window` without adding an entry to their Cargo.toml.
- Change `with_parent_window` to accept a generic parameter that implements `RawWindowHandle`, allowing winit to acquire the handle via the trait rather than requiring the user to call the trait directly.

Thank you for the wonderful library.

kchibisov:
>My wrapper I'm building is multi-threaded, and I can't send a winit WindowBuilder across threads due to some x11 types not being Send. So, I've recreated my own WindowBuilder, and in an effort to support with_parent_window, I realized I cannot actually WindowBuilder::with_parent_window providing it a winit::window::Window. Instead, I must explicitly use the RawWindowHandle trait. This forces my crate to have two dependencies, instead of just winit. This is obviously a very minor issue, but part of me wishes I didn't have to add this dependency.

Aside from the issue, but why you'd want to send a `WindowBuilder` across the threads if you can only create a window on the thread event loop is on? Do you send requests with `WindowBuilder` from other threads to build a window on the main thread?

ecton:
> Do you send requests with WindowBuilder from other threads to build a window on the main thread?

Yes. The wrapper I'm writing creates a one-thread-per-window model, which I'm fond of, as no single window can block the event loop or each other.

ecton:
Correct. Or more specifically, an Arc-clone of it, because the event loop thread maintains a collection of open windows.

daxpedda:
[...]

PR #2998, opened 2023-08-02
Export smol_str; impl Ord for Key

### LINKED ISSUE #2996: Constructing Key::Character(_) without importing Smolstr
Currently to construct a `Key::Character(_)` I have to:
1. Import `Smolstr`
2. Do e.g. `Key::Character(SmolStr::new_inline("/"))`

Since Winit includes `Smolstr` in its public API, I recommend that it be publically re-export (e.g. `winit::keyboard::Smolstr`) so that users do not have to add a dependency on (the correct version of) this crate too.

Further, I suggest adding convenience constructors. Some ideas:

-`fn Key::new_char(c: char) -> Option<Self>` — note that control characters are not allowed
- `fn Key::new_str(s: impl AsRef<str>) -> Option<Self>`
- `fn Key::new_str_inline(s: &str) -> Option<Self>` — uses `SmolStr::new_inline`

Since `Key::Character` is not a `char` maybe it has the wrong name? (I bet there's already far too much discussion on that topic.)

dhardy:

Publically export smol_str. Closes #2996 

Implement `Ord` for `Key`, `NativeKey`. Motivation: I have a mapping from `Key` to X that I serialise to a file. This is normally stored in a `HashMap`, but sorted before serialisation to ensure the file does not constantly mutate needlessly (and make direct edits to the file more user friendly).

I *could* add a note to `Key` mentioning that ordering is arbitrary and not guaranteed to remain stable across releases, but this seems like fluff.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On macOS, add tabbing APIs on `WindowExtMacOS` and `EventLoopWindowTargetExtMacOS`.
 - **Breaking:** Rename `Window::set_inner_size` to `Window::request_inner_size` and indicate if the size was applied immediately.
 - On X11, fix false positive flagging of key repeats when pressing different keys with no release between presses.
-- Implement `PartialOrd` and `Ord` for `KeyCode` and `NativeKeyCode`.
+- Implement `PartialOrd` and `Ord` for `Key`, `KeyCode`, `NativeKey` and `NativeKeyCode`.
 - On Web, implement `WindowEvent::Occluded`.
 - On Web, fix touch location to be as accurate as mouse position.
 - On Web, account for CSS `padding`, `border`, and `margin` when getting or setting the canvas position.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - **Breaking:** Rename `Window::set_inner_size` to `Window::request_inner_size` and indicate if the size was applied immediately.
 - On X11, fix false positive flagging of key repeats when pressing different keys with no release between presses.
 - Implement `PartialOrd` and `Ord` for `Key`, `KeyCode`, `NativeKey` and `NativeKeyCode`.
+- Add `ElementState::is_pressed`.
 - On Web, implement `WindowEvent::Occluded`.
 - On Web, fix touch location to be as accurate as mouse position.
 - On Web, account for CSS `padding`, `border`, and `margin` when getting or setting the canvas position.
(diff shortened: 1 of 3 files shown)


dhardy:
Added a convenience method:
```rust
impl ElementState {
    /// True if `self == Pressed`
    pub fn is_pressed(self) -> bool {
        self == ElementState::Pressed
    }
}
````

File: CHANGELOG.md [hunk 1]
```diff
@@ -17,7 +17,8 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On macOS, add tabbing APIs on `WindowExtMacOS` and `EventLoopWindowTargetExtMacOS`.
 - **Breaking:** Rename `Window::set_inner_size` to `Window::request_inner_size` and indicate if the size was applied immediately.
 - On X11, fix false positive flagging of key repeats when pressing different keys with no release between presses.
-- Implement `PartialOrd` and `Ord` for `KeyCode` and `NativeKeyCode`.
+- Implement `PartialOrd` and `Ord` for `Key`, `KeyCode`, `NativeKey` and `NativeKeyCode`.
```
kchibisov, line 20:
```suggestion
- Implement `PartialOrd` and `Ord` for `Key`, `KeyCode`, `NativeKey`, and `NativeKeyCode`.
```

src/keyboard.rs
```diff
@@ -69,7 +69,7 @@
 //
 // --------- END OF W3C SHORT NOTICE ---------------------------------------------------------------
 
-use smol_str::SmolStr;
+pub use smol_str::SmolStr;
```
kchibisov, line 72: Could you inline docs?

src/event.rs
```diff
@@ -1019,6 +1019,13 @@ pub enum ElementState {
     Released,
 }
 
+impl ElementState {
+    /// True if `self == Pressed`
```
kchibisov, line 1023:
```suggestion
    /// True if `self == Pressed`.
```

kchibisov (changes requested):

src/keyboard.rs
dhardy: Looks fine to me when generated locally

kchibisov: Maybe it works due to the `pub use` and not because it's an alias from the `SmolStr`, I know that docs are not inlined in some cases.

kchibisov (approved):

src/keyboard.rs
dhardy: Half the time it seems like rustdoc completely ignores inline/no_inline doc attributes; I could never figure it out.

daxpedda:
Cc #2913.
I'm in favor of having a policy that extern types should always be re-exported, if not their whole library.

Issue #336, opened 2017-11-01
A unified pointer model is probably a good idea


chris-morgan:
I note `Touch`, but it’s rather limited. My Surface Book, for example, has a pressure-sensitive pen, which neither `Touch` nor `Mouse` can model; and the latest such pens support tilt as well.

I suspect it’s worth looking into something like the Pointer Events API which browsers have all comparatively recently taken up.

marcianx:
I'd also like to note https://github.com/tomaka/glutin/issues/916 which I was asked to forward to here:

> The Touch event does not include a touch radius or any other obvious way to distinguish finger and stylus touch events. This is particularly useful for palm rejection on large screens when using a stylus. Could such support be added?

tangmi:
Would a precursor to this be getting pen/tablet support first (https://github.com/rust-windowing/winit/issues/99)?

Osspial:
@tangmi I think it goes the other way - getting richer touch support will lay the ground for adding pen/tablet events. We've slowly been making progress towards richer touch events, via recent touch pressure PRs for [iOS](https://github.com/rust-windowing/winit/pull/1090) and [Windows](https://github.com/rust-windowing/winit/pull/1134), and I expect that once more cross-platform support for rich touch events has been added pen/tablet support won't be too much of a technical or API jump.

> It seems like there needs to be a decision on whether to keep Mouse/Touch/Pen separate or unify them under a "PointerEvent" type.

That's... an intriguing idea, actually. I'd support unifying pointer events into a single type, especially since that'll improve usability in the naive case where an application is designed for just one input type and given input of a format it didn't necessarily expect.

chris-morgan:
The Pointer Events API which I suggested as a model is *all about* unified event types. If you’re not familiar with it, please do read up on it before implementing anything in Winit.

The web still does `click` events for touch and pen, `mousedown` *et al* for pen events, and scrolling/zooming for touch events (adjustable via the `touch-action` CSS property). Naive users should use those things rather than `pointerdown` *et al.* This is a sane model for more than just the web’s backwards compatibility, because touch should generally act differently from a mouse or pen.

tangmi:
@Osspial I wanted to try pen input on web through winit, so I've hacked together basic support for pointer events in `winit-legacy` (diff here: https://github.com/rust-windowing/winit/compare/2e11615...tangmi:winit-legacy-2).

Some observations that may already be known:
[...]

Issue #2963, opened 2023-07-17
Extend touchpad magnify and rotate gestures support to Web


megaboy101:
Hey all,

I recently ran into an issue where I wanted to add two-finger pinch/zoom to a webassembly-compiled Bevy app I was building. 

However it looks like magnify and rotate gesture recognition is currently limited to MacOS (#2157).

I'm not too familiar with Winit's internals, but I'd be interested in adding a webassembly implementation to support these gesture events if possible.

daxpedda:
Hm, I'm not sure we should, it's not supported but we could polyfill it of course. See the [MDN documentation on how to polyfill this](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events/Pinch_zoom_gestures).

Safari has actually a [dedicated API](https://developer.apple.com/documentation/webkitjs/gestureevent) for it, which we could definitely implement.

@kchibisov do we intend to polyfill this for other targets too if they don't have official support for it? I saw in #2157 that Wayland and Windows support it, does X11 support it?

daxpedda:
Alright, let's just go ahead with this then.

@megaboy101 PR's are welcome, I'm happy to review one. Feel free to ping if you need any guidance!

Issue #2942, opened 2023-07-09
Touch input contact size


daxpedda:
I noticed that it has been really helpful to determine the contact size of a touch input by using [`PointerEvent.width/height`](https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/width) to improve accuracy.

I'm not sure about other platforms, but a quick look at the Windows API revealed [`POINTER_TOUCH_INFO.rcContact`](https://learn.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-pointer_touch_info).
In any case, like `force`, this could be wrapped in an `Option`.

kchibisov:
Wayland should have it as well https://wayland.app/protocols/wayland#wl_touch:event:shape

kchibisov:
The thing is that device may not support reporting the shape of the input, so I'd assume it must be an `Option`.

kchibisov:
Also, it's an ellipse and not a rectangle on Wayland, so it's a bit weird here.

daxpedda:
After some digging into this lately, all these APIs are meant to represent ellipsis.

- Web should move all `WindowEvent::Touch` to use [touch events](https://developer.mozilla.org/en-US/docs/Web/API/Touch_events) instead of [pointer events](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events). These expose much richer information ([`Touch.radiusX/Y`](https://developer.mozilla.org/en-US/docs/Web/API/Touch/radiusX) and [`Touch.rotationAngle`](https://developer.mozilla.org/en-US/docs/Web/API/Touch/rotationAngle)) and align much more with other backends. They also clearly state that the contact points are ellipsis.
- Wayland has [`wl_touch::shape`](https://wayland.app/protocols/wayland#wl_touch:event:shape) and [`wl_touch::orientation`](https://wayland.app/protocols/wayland#wl_touch:event:orientation) describing the ellipsis.
- Windows has [`POINTER_TOUCH_INFO.rcContact` and `POINTER_TOUCH_INFO.orientation`](https://learn.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-pointer_touch_info), which are supposed to represent an ellipsis as well, but is just not documented as such. [Chromium treats this information as an ellipsis as well](https://source.chromium.org/chromium/chromium/src/+/ae99ae281ffc2fe19024f86c5c64939efc1b2e87:ui/views/win/hwnd_message_handler.cc;l=3370-3379).

[...]

PR #3001, opened 2023-08-04
Pointer API Sketch

0x182d4454fb211940:
- [ ] Tested on all platforms changed (No. Only wayland tested)
- [ ] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users (No)
- [ ] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior (No)
- [ ] Created or updated an example program if it would help users understand this functionality (No)
- [ ] Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented (No)

This is a very rough sketch of what a pointer-based API could look like (based on #336). Personally I'm interested in moving winit towards an API that can support pen input, which is why I'm interested in this, and I'm willing to help implement it. But I'd like to make sure the WindowEvent API matches what the winit team wants and what would be useful and ergonomic for the user, so please let me know your thoughts!

It's worth noting that by heading in this direction we will be breaking apps which already use cursor and touch events. Would it be better to just add a "pen" event like with touch?

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -378,35 +378,10 @@ pub enum WindowEvent {
     /// - **iOS / Android / Web / Orbital:** Unsupported.
     Ime(Ime),
 
-    PointerMoved {
+    Pointer {
         device_id: DeviceId,
-        source: PointerSource,
-        location: PhysicalPosition<f64>,
-        force: Option<Force>,
-    },
-
-    PointerInput {
-        device_id: DeviceId,
-        source: PointerSource,
-        state: ElementState,
-        button: Option<MouseButton>,
-        location: Option<PhysicalPosition<f64>>,
-        force: Option<Force>,
-    },
-
-    PointerEntered {
-        device_id: DeviceId,
-        source: PointerSource,
-    },
-
-    PointerLeft {
-        device_id: DeviceId,
-        source: PointerSource,
-    },
-
-    PointerCancelled {
-        device_id: DeviceId,
-        source: PointerSource,
+        pointer_id: PointerId,
+        event: PointerEvent,
     },
 
     // /// An mouse button press has been received.
@@ -417,7 +392,7 @@ pub enum WindowEvent {
     // },
     /// A mouse wheel movement or touchpad scroll occurred.
     MouseWheel {
-        device_id: DeviceId,
+        device_id: DeviceId,g
         delta: MouseScrollDelta,
         phase: TouchPhase,
     },
@@ -875,21 +850,45 @@ pub enum Ime {
     Disabled,
 }
 
-#[derive(Debug, Hash, PartialEq, Eq, Clone, Copy)]
-#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-pub enum PointerSource {
-    Cursor,
-    Touch { finger: u64 },
-}
-
-/// Describes touch-screen input state.
-#[derive(Debug, Hash, PartialEq, Eq, Clone, Copy)]
+#[derive(Debug, PartialEq, Clone, Copy)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub enum TouchPhase {
     Started,
     Moved,
     Ended,
     Cancelled,
+}
+
+#[derive(Debug, PartialEq, Clone, Copy)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum PointerEvent {
+    Created,
+    Destroyed,
+    Entered,
+    Left,
+    UpdateForce(Force),
+    UpdateTilt(Tilt),
+    UpdateAngle(f64),
+    Moved(PhysicalPosition<f64>),
+    Button {
+        button: PointerButton,
+        state: ElementState,
+    },
+    MotionCancelled,
+}
+
+#[derive(Debug, Hash, PartialEq, Eq, Clone, Copy)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum PointerId {
+    Cursor,
+    Touch { finger: u64 },
+}
+
+#[derive(Debug, PartialEq, Clone, Copy)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct Tilt {
+    angle_x: f64,
+    angle_y: f64,
 }
 
 /// Describes the force of a touch event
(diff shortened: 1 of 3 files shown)


daxpedda:
I don't think we need all these different types, I think having one enum that can either be `Mouse`/`Touch`/`Pen` with the appropriate data attached to all events should work.

Generally speaking I'm in favor of unifying all three into one event though, it also fits well with the Pointer API on Web.

0x182d4454fb211940:
The reason I went with the extra types is some states don't make sense. For example, a `PointerInput` will never have both `PointerSource::Touch { .. }` and `button: Some(_)`. But I guess there's no need to be overly correct, especially if it's easier to use.

daxpedda:
> The reason I went with the extra types is some states don't make sense. For example, a `PointerInput` will never have both `PointerSource::Touch { .. }` and `button: Some(_)`. But I guess there's no need to be overly correct, especially if it's easier to use.

Yeah, you are right, maybe these types made sense after all? Let's wait for more input.
I would prefer types to represent mutually exclusive stuff indeed, so your previous implementation actually makes more sense to me.

kchibisov:
There was an attempt to do it here https://github.com/rust-windowing/winit/pull/1374 before.

daxpedda:
I don't think I like splitting up all the information in separate events, but it makes the mess with types easier.

0x182d4454fb211940:
Some thoughts and questions while designing this version:

* In #1374, what is `PointerContactArea`? Is that necessary for a first implementation?
* I renamed twist to angle (perhaps rotation would be better?) since I found twist a bit of strange name. What name would be best? Should that value be newtyped?
* On Wayland at least, I know the tablet api sends a "frame" event once all information (pressure, tilt, position) has been sent, this might be a useful thing to add but I don't know how well support would work across platforms. (Thinking about it I'd expect it to work, either they send the event in parts like wayland, and then they sort of have to say when it ends, or they send the event as a whole, and then we can just send a frame event when all the information has been sent).
* The internals of PointerId will probably want to be hidden as in #1374.
* We might want to send a "Capabilties" enum alongside the "created" event or just after. I don't know what that would look like cross-platform, but I suppose we could simulate it by reporting capabilities as the info comes in. Would this even be useful?
* If you squint your eyes, this api seems to be capable of expressing macOS touchpad events too... could scrolling also become a pointer api?
* As you say @daxpedda it is unfortunate that all the info gets split. I think it would probably be quite easy to make some kind of struct which keeps track of the state that comes out of these events, but maybe that's what winit should be doing in the first place. Should this be a kind of "middle api", which the user can opt in to? (so we first unify all platforms in this simple event system, then express more abstract events, closer to what winit currently does)

(I didn't bother updating the wayland implementation, there's not really any point right now)

daxpedda:
> * In [Experiment with an improved event design #1374](https://github.com/rust-windowing/winit/pull/1374), what is `PointerContactArea`? Is that necessary for a first implementation?

I think that's #2942, so no, it's not necessary for a first implementation.
> * I renamed twist to angle (perhaps rotation would be better?) since I found twist a bit of strange name. What name would be best? Should that value be newtyped?

We don't currently support twist/angle/rotation, so I would argue we can leave this for another PR.
> * On Wayland at least, I know the tablet api sends a "frame" event once all information (pressure, tilt, position) has been sent, this might be a useful thing to add but I don't know how well support would work across platforms. (Thinking about it I'd expect it to work, either they send the event in parts like wayland, and then they sort of have to say when it ends, or they send the event as a whole, and then we can just send a frame event when all the information has been sent).

I can't really comment on that, I only maintain the Web backend where this doesn't happen, but I think we can figure that out when we get to the implementation phase.
> * The internals of PointerId will probably want to be hidden as in [Experiment with an improved event design #1374](https://github.com/rust-windowing/winit/pull/1374).

in #1374 `PointerId` was what we call `DeviceId` now. So I don't think they are comparable. It might be interesting to explore if we can can fold this two types together, but I don't particularly see the point of that.
> * We might want to send a "Capabilties" enum alongside the "created" event or just after. I don't know what that would look like cross-platform, but I suppose we could simulate it by reporting capabilities as the info comes in. Would this even be useful?

Depends on what exactly this would entail but it doesn't sound very appealing to me, in any case, I think this is a separate problem we can figure out outside this PR.
> * If you squint your eyes, this api seems to be capable of expressing macOS touchpad events too (rotation, pressure, could add an event for magnification)... as well as this could scrolling become part of the pointer api?

Probably not a bad idea, again though, imo we can figure this out later, cc #2963.
> * As you say @daxpedda it is unfortunate that all the info gets split. I think it would probably be quite easy to make some kind of struct which keeps track of the state that comes out of these events, but maybe that's what winit should be doing in the first place. Should this be a kind of "middle api", which the user can opt in to? (so we first unify all platforms in this simple event system, then express more abstract events, closer to what winit currently does)

I think this is really over-complicating it. Maintainers just have to make a decision here.

@kchibisov I would strongly prefer to send as much information as possible in a single event instead of splitting it up like #1374 did. So I prefer the original implementation of @0x182d4454fb211940: https://github.com/rust-windowing/winit/pull/3001/commits/71125f948c47cf5f23d47c21d641bbf68d908d35. Hopefully we can clean this up a bit to reduce the amount of types, but that's what I would currently prefer, WDYT?
> (I didn't bother updating the wayland implementation, there's not really any point right now)

Don't waste your time on it, we can figure it out after we finish the design phase.

0x182d4454fb211940:
So in the meantime I've made a rough fork of winit and bevy to start working on my project. What worked there has given me another potential idea: splitting the data from the event.

```rust

    control_flow.set_wait();

    match event {
        Event::WindowEvent {
            event: WindowEvent::PointerDown(PointerId::Pen(id)),
            ..
        } => {
            println!("pen is down, with pressure of {}", window.pen(id).pressure.unwrap_or(f64::NAN));
        },
        Event::WindowEvent {
            event: WindowEvent::PointerDown(PointerId::Mouse(id)),
            ..
        } if window.mouse(id).just_pressed(MouseButton::Left) => {
            println!("you just clicked at {:?}", window.mouse(id).position());
        },
        Event::WindowEvent {
            event: WindowEvent::PointerUp(id),
            ..
        } => {
            println!("pointer {id:?} just lost contact with the screen at {:?}", window.pointer(id).position());
        },
        /* snip */
    }
});
```
(example adapted from docs.rs, so it doesn't match the master branch)

This is quite radically different from how winit currently does things, and I'm not sure if this is better (it would either require something which pre-processes events or some kind of cross-module interior mutability). But it does solve the problem of representing attached data correctly (in order to access the extra data, you need to use the id with the correct type), and would also mean that someone who wants to check if the user clicked in a certain area, they could just access the mouses position alongside the event. I've found a similar (integrated with bevy) approach to work quite well for my project.

kchibisov:
See #3876.

PR #3003, opened 2023-08-04
On Wayland, reduce amount of spurious wakeups

kchibisov:
Mark it as breaking, since some clients relied on that behavior, simply because dispatching clients queue always woke up a winit, meaning that they won't be able to use user events for this sake.

### DIFF
--- a/src/platform_impl/linux/wayland/state.rs
+++ b/src/platform_impl/linux/wayland/state.rs
@@ -1,5 +1,6 @@
 use std::cell::RefCell;
 use std::error::Error;
+use std::sync::atomic::Ordering;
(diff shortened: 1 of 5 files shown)


src/platform_impl/linux/wayland/state.rs
```diff
@@ -167,6 +172,8 @@ impl WinitState {
             monitors: Arc::new(Mutex::new(monitors)),
             events_sink: EventSink::new(),
             loop_handle,
+            // Make it true by default.
+            disptached_events: true,
```
madsmtm, line 176: Nit: Typo
```suggestion
            dispatched_events: true,
```

PR #3004, opened 2023-08-04
Android: Support unicode character mapping + dead keys

rib:
Up until now the Android backend has been directly mapping key codes which essentially just represent the "physical" cap of the key (quoted since this also related to virtual keyboards).

Since we didn't account for any meta keys either it meant the backend only supported a 1:1 mapping from key codes, which only covers a tiny subset of characters. For example you couldn't type a colon since there's no keycode for that and we didn't try and map Shift+Semicolon into a colon character.

This has been tricky to support because the `NativeActivity` class doesn't have direct access to the Java `KeyEvent` object which exposes a more convenient `getUnicodeChar` API.

It is now possible to query a `KeyCharcterMap` for the device associated with a `KeyEvent` via the `AndroidApp::device_key_character_map` API which provides a binding to the SDK `KeyCharacterMap` API in Java:

 https://developer.android.com/reference/android/view/KeyCharacterMap

This is effectively what `getUnicodeChar` is implemented based on and is a bit more general purpose.

`KeyCharacterMap` lets us map a key_code + meta_state from a `KeyEvent` into either a unicode character or dead key accent that can be combined with the following key. This mapping is done based on the user's chosen layout for the keyboard.

To enable support for key character maps the
`AndroidApp::input_events()` API was replaced by
`AndroidApp::input_events_iter()` which returns a (lending) iterator for events. This was changed because the previous design made it difficult to allow other AndroidApp APIs to be used while iterating events (mainly because AndroidApp held a lock over the backend during iteration)

-- 

_Note: this currently depends on a git branch of android-activity https://github.com/rust-mobile/android-activity/tree/rib/pr/input-api-rework-with-key-character-maps which I'm aiming to land soon in preparation for a 0.5 release of android-activity in time for Winit 0.29_


src/platform_impl/android/keycodes.rs
```diff
@@ -156,324 +159,393 @@ pub fn to_physical_keycode(keycode: Keycode) -> KeyCode {
...
+    app: &AndroidApp,
+    key_event: &KeyEvent<'_>,
+    combining_accent: &mut Option<char>,
+) -> Option<KeyMapChar> {
+    let device_id = key_event.device_id();
+
+    let key_map = match app.device_key_character_map(device_id) {
+        Ok(key_map) => key_map,
+        Err(err) => {
+            log::error!("Failed to look up `KeyCharacterMap` for device {device_id}: {err:?}");
```
kchibisov, line 177: Do we really want `error!` and not `warn!` here?

src/platform_impl/android/keycodes.rs
```diff
@@ -156,324 +159,393 @@ pub fn to_physical_keycode(keycode: Keycode) -> KeyCode {
...
+            }
+            Some(KeyMapChar::CombiningAccent(accent))
+        }
+        Ok(KeyMapChar::None) => {
+            // Leave any combining_accent state in tact (seems to match how other
+            // Android apps work)
+            None
+        }
+        Err(err) => {
+            log::error!("KeyEvent: Failed to get key map character: {err:?}");
```
kchibisov, line 216: Probably same about it here.

src/platform_impl/android/keycodes.rs
```diff
@@ -156,324 +159,393 @@ pub fn to_physical_keycode(keycode: Keycode) -> KeyCode {
...
+    match key_map.get(key_event.key_code(), key_event.meta_state()) {
+        Ok(KeyMapChar::Unicode(unicode)) => {
+            // Only do dead key combining on key down
+            if key_event.action() == KeyAction::Down {
+                let combined_unicode = if let Some(accent) = combining_accent {
+                    match key_map.get_dead_key(*accent, unicode) {
+                        Ok(Some(key)) => Some(key),
+                        Ok(None) => None,
+                        Err(err) => {
+                            log::error!("KeyEvent: Failed to combine 'dead key' accent '{accent}' with '{unicode}': {err:?}");
```
kchibisov, line 191: Same about `warn!`.

src/platform_impl/android/keycodes.rs
```diff
@@ -156,324 +159,393 @@ pub fn to_physical_keycode(keycode: Keycode) -> KeyCode {
...
+        Err(err) => {
+            log::error!("Failed to look up `KeyCharacterMap` for device {device_id}: {err:?}");
+            return None;
+        }
+    };
+
+    match key_map.get(key_event.key_code(), key_event.meta_state()) {
+        Ok(KeyMapChar::Unicode(unicode)) => {
+            // Only do dead key combining on key down
+            if key_event.action() == KeyAction::Down {
```
kchibisov, line 185: Could you move `if` to match arm to reduce nesting?

src/platform_impl/android/keycodes.rs
```diff
@@ -156,324 +159,393 @@ pub fn to_physical_keycode(keycode: Keycode) -> KeyCode {
...
+                let combined_unicode = if let Some(accent) = combining_accent {
+                    match key_map.get_dead_key(*accent, unicode) {
```
kchibisov, line 187: I think you can combine both `if let` and `match` with the help of `map`/`then`/`mar_or`.

src/platform_impl/android/keycodes.rs
```diff
@@ -156,324 +159,393 @@ pub fn to_physical_keycode(keycode: Keycode) -> KeyCode {
...
-        ProfileSwitch => Key::Unidentified(native),
+    let native = NativeKey::Android(keycode.into());
+
+    match key_char {
+        Some(KeyMapChar::Unicode(c)) => {
+            Key::Character(smol_str::SmolStr::from_iter([c].into_iter()))
+        }
+        Some(KeyMapChar::CombiningAccent(c)) => Key::Dead(Some(c)),
+        None | Some(KeyMapChar::None) => {
+            match keycode {
```
kchibisov, line 234: Could you move `match` into `=>`  with an extra bracket?

src/platform_impl/android/keycodes.rs
```diff
@@ -156,324 +159,393 @@ pub fn to_physical_keycode(keycode: Keycode) -> KeyCode {
...
+                ProgYellow => Key::ColorF2Yellow,
+                ProgBlue => Key::ColorF3Blue,
+                AppSwitch => Key::AppSwitch,
+                Button1 => Key::Unidentified(native),
+                Button2 => Key::Unidentified(native),
+                Button3 => Key::Unidentified(native),
+                Button4 => Key::Unidentified(native),
+                Button5 => Key::Unidentified(native),
+                Button6 => Key::Unidentified(native),
+                Button7 => Key::Unidentified(native),
```
kchibisov, line 452: Could you move all `Unidentified` stuff to the bottom of the `match`?

src/platform_impl/android/mod.rs
```diff
@@ -353,128 +354,143 @@ impl<T: 'static> EventLoop<T> {
...
+        // Process input events
+        match android_app.input_events_iter() {
+            Ok(mut input_iter) => {
+                loop {
+                    let read_event = input_iter.next(|event| {
+                        let mut input_status = InputStatus::Handled;
+                        match event {
+                            InputEvent::MotionEvent(motion_event) => {
+                                let window_id = window::WindowId(WindowId);
+                                let device_id = event::DeviceId(DeviceId);
```
kchibisov, line 370: This loop looks completely insane when it comes to indention, I'm pretty sure it could be logically formatted into something more readable, but the way it's now is barely readable.

src/platform_impl/android/mod.rs
```diff
@@ -353,128 +354,143 @@ impl<T: 'static> EventLoop<T> {
...
-                                )
-                            },
-                            event::TouchPhase::Moved
-                            | event::TouchPhase::Cancelled => {
-                                Box::new(motion_event.pointers())
+        // Process input events
+        match android_app.input_events_iter() {
+            Ok(mut input_iter) => {
+                loop {
+                    let read_event = input_iter.next(|event| {
```
kchibisov, line 365: You can merge all of that by mapping an iterator `map` and then doing `for event in <your_iterator>`.

kchibisov (changes requested):

src/platform_impl/android/keycodes.rs
rib: yeah warn is probably fine

src/platform_impl/android/mod.rs
rib: yeah, I'm not liking the indentation either :/

I at least factored out the code for the character mapping but yeah this really sucks atm.

I can poke at it.

src/platform_impl/android/keycodes.rs
rib: guess so

src/platform_impl/android/mod.rs
rib: The input event iterator isn't a standard Rust Iterator because it's a lending iterator which is largely why the ergonomics are a bit awkward here.

I've factored out a `handle_input_event` function though and that helps a lot with making this readable.

src/platform_impl/android/keycodes.rs
rib: really though - this is just an existing match that's been indented, this is busy work that's not really important surely?

src/platform_impl/android/keycodes.rs
rib: without the extra bracket I guess you mean - can do that

src/platform_impl/android/keycodes.rs
kchibisov: dunno, your editor should allow that to do in a couple keystrokes. I can format it myself under 5-10s.

rib: it's also notable that they are currently in numerical order and some of them are logically related to other supported keys that are adjacent. I think we should leave these ordered as they are.

src/platform_impl/android/keycodes.rs
kchibisov: yes, without.

src/platform_impl/android/keycodes.rs
kchibisov: Maybe, it's just hard to grasp what really is mapped and what's not. If the thing is not mapped it doesn't really matter, so keeping it separately in the end is more pleasant to read.

src/platform_impl/android/mod.rs
kchibisov: If that's really an issue just do early `let input_iter = match android_app ... ` with early return to simply remove nesting. The giga closure could probably be extracted as well.

src/platform_impl/android/keycodes.rs
rib: We aren't really indented much here and it's more convenient for placing the comment this way.

I tried it without the indent but personally I find that much less pleasant for scanning the arms of the match due to the braces of the if statement.

More importantly though cargo fmt agrees and puts the brace back :)

src/platform_impl/android/keycodes.rs
rib: It doesn't really seem like it to me. The outer if let can't be done with a map since it's mapping the Some and None both to Some.

Maybe there's some elaborate combinator version of this code but it's not standing out for me and I think it's reasonable enough as it is.

src/platform_impl/android/mod.rs
rib: I moved the hander to a separate function. 

allowing early return would have been very extremely undesirable in the middle of `single_iteration()`

kchibisov (approved):
Should be good after you fix formatting.

rib:
For reference here, I've landed the corresponding support in android-activity `main` (after also reverting back to a rust-version of 1.64) so the updated Cargo.toml now pulls android-activity from `main`.

I could potentially do a 0.5 beta release if needs be, to avoid the temporary git dependency.

One small tweak that was made before I merged the android-activity API was to rename `get_dead_key` to `get_dead_char` since that's what the corresponding Android SDK is called.

PR #3006, opened 2023-08-05
Remove 'static requirement on `run`

kchibisov:
There's no need to force the static on the users, given that internally some backends were not using static in the first place.

- [-] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users

### DIFF
--- a/src/platform_impl/web/event_loop/mod.rs
+++ b/src/platform_impl/web/event_loop/mod.rs
@@ -29,11 +29,23 @@ impl<T> EventLoop<T> {
         }
     }
 
-    pub fn run<F>(self, event_handler: F) -> !
+    pub fn run<F>(self, mut event_handler: F) -> !
     where
         F: FnMut(Event<T>, &RootEventLoopWindowTarget<T>, &mut ControlFlow),
     {
-        self.spawn_inner(event_handler, false);
+        let target = RootEventLoopWindowTarget {
+            p: self.elw.p.clone(),
+            _marker: PhantomData,
+        };
+
+        // SAFETY: Don't use `move` to make sure we leak the `event_handler` and `target`.
+        let handler: Box<dyn FnMut(_, _)> =
+            Box::new(|event, flow| event_handler(event, &target, flow));
+        // SAFETY: The `transmute` is necessary because `run()` requires `'static`. This is safe
+        // because this function will never return and all resources not cleaned up by the point we
+        // `throw` will leak, making this actually `'static`.
+        let handler = unsafe { std::mem::transmute(handler) };
+        self.elw.p.run(handler, false);
 
         // Throw an exception to break out of Rust execution and use unreachable to tell the
         // compiler this function won't return, giving it a return type of '!'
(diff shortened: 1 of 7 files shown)


src/platform_impl/web/event_loop/mod.rs
```diff
@@ -31,7 +31,7 @@ impl<T> EventLoop<T> {
 
     pub fn run<F>(self, event_handler: F) -> !
     where
-        F: 'static + FnMut(Event<T>, &RootEventLoopWindowTarget<T>, &mut ControlFlow),
```
notgull, line 34: I think this would be unsound on Web, since `run` technically returns immediately, meaning that any references held in this closure would be left dangling.

notgull (commented):
This would work for `async-winit`. In fact, it would work very well, since we could remove the `'static` bound on the future in [`block_on`](https://docs.rs/async-winit/latest/async_winit/event_loop/struct.EventLoop.html#method.block_on).

However, it looks like this might be unsound on web to my uninformed eye,

kchibisov:
@daxpedda @madsmtm @rib 

I guess we can't do that for a reason that some backend no-return basically. However, maybe we should just give up and say that users should use `spawn`/`run` on `web`/`ios` respectively?

src/platform_impl/web/event_loop/mod.rs
daxpedda: On Web `run()` doesn't return immediately, it interrupts the function without unwinding, so the closure passed into `run()` leaks and therefore this should work just fine.

madsmtm:
I think the macOS backend is currently unsound if `'static` is not set, since we may not be clearing the callback consistently.

Though that should be fixable.

Re if the backend doesn't ever return, then not having `'static` is still fine.

kchibisov:
@madsmtm the thing is that we return, but the `API` consumes the `EventLoop` so the user can't rerun things, so `EventLoop` itself is dropped and users can't create it again.

@daxpedda I think I've missed something?

@madsmtm I was mostly curios about ios.

daxpedda:
This should work fine for Web.
Rust actually just added support for the exception handling proposal, but only with `no_std`, which `wasm-bindgen` doesn't support. But when it actually hits `run()` on Web will be even weirder then before and actually breaks without `'static`. I can add a check to make `run()` not available if exception handling is enabled.

> @daxpedda I think I've missed something? Because it seems like other code is bound to `'static`?

I can fix that.

madsmtm (commented):
I'm not sure I understand you, but I was trying to say that the change should be sound on iOS (even though it doesn't compile as-is).

daxpedda:
> > @daxpedda I think I've missed something? Because it seems like other code is bound to `'static`?
> 
> I can fix that.

Required an `unsafe` conversion, but works.

kchibisov:
> Required an unsafe conversion, but works.

Yeah, that's what I was afraid of, so I left it as is, unless a backend maintainer approves.

kchibisov:
@madsmtm you've replied about macOS initially though, not ios though. I guess your reply about ios was indirect though.

Does it matter for macOS how you cleanup, if you basically drop the `EventLoop`, like `'static` shouldn't matter once you literally call `drop` on the entire loop in the end of `run` invocation?

daxpedda (approved):
For Web this works, but probably not useful and I would argue makes `run()` more and more problematic.

We can have another discussion about whether we should leave `run()` on Web or not in a different place and time.

kchibisov:
I guess I'll do the beta 2 with it like that, and then we could probably gather some feedback whether we'd really need it to keep that way.

PR #3007, opened 2023-08-05
Make with_x11_visual take ID instead of a pointer

### LINKED ISSUE #2746: `WindowBuilderX11Ext::with_x11_visual` is unsound
As mentioned in https://github.com/rust-windowing/winit/pull/106#discussion_r246866630 , the function just read from a raw pointer, and the user could pass any dangling raw pointer. 
The function should be marked as unsafe, and should document what are the condition of validity of the pointer and for how long it should be valid.

notgull:
At the moment, the with_x11_visual function takes a pointer and immediately dereferences it to get the visual info inside. As it is safe to pass a null pointer to this function, it is unsound. This commit replaces the pointer parameter with a visual ID, and then uses that ID to look up the actual visual under
the X11 setup. As this is what was already practically happening before, this change shouldn't cause any performance downgrades.

This is a breaking change, but it's done in the name of soundness so it should be okay. It should be trivial for end users to accommodate it, as it's just a matter of getting the visual ID from the pointer to the visual before passing it in.

Closes #2746

### DIFF
--- a/src/platform_impl/linux/x11/window.rs
+++ b/src/platform_impl/linux/x11/window.rs
@@ -214,30 +214,44 @@ impl UnownedWindow {
             None => xconn.default_screen_index() as c_int,
         };
 
+        // An iterator over all of the visuals combined with their depths.
+        let mut all_visuals = xconn
+            .xcb_connection()
+            .setup()
+            .roots
+            .iter()
+            .flat_map(|root| &root.allowed_depths)
+            .flat_map(|depth| {
+                depth
+                    .visuals
+                    .iter()
+                    .map(move |visual| (visual, depth.depth))
+            });
+
         // creating
-        let (visual, depth, require_colormap) = match pl_attribs.visual_infos {
-            Some(vi) => (vi.visualid as _, vi.depth as _, false),
+        let (visualtype, depth, require_colormap) = match pl_attribs.visual_id {
+            Some(vi) => {
+                // Find this specific visual.
+                let (visualtype, depth) = all_visuals
+                    .find(|(visual, _)| visual.visual_id == vi)
+                    .ok_or_else(|| os_error!(OsError::XError(X11Error::NoSuchVisual(vi).into())))?;
+
+                (Some(visualtype), depth, true)
+            }
             None if window_attrs.transparent => {
                 // Find a suitable visual, true color with 32 bits of depth.
-                xconn.xcb_connection().setup().roots
-                    .iter()
-                    .flat_map(|root| &root.allowed_depths)
-                    .filter(|depth| depth.depth == 32)
-                    .flat_map(|depth| depth.visuals.iter().map(move |visual| (visual, depth.depth)))
+                all_visuals
                     .find(|(visual, _)| {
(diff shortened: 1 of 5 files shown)


src/platform_impl/linux/x11/window.rs
```diff
@@ -214,30 +214,44 @@ impl UnownedWindow {
...
+                    }, |(visual, depth)| (Some(visual), depth, true))
             }
-            _ => (
-                x11rb::COPY_FROM_PARENT as _,
-                x11rb::COPY_FROM_PARENT as _,
-                false,
-            ),
+            _ => (None, x11rb::COPY_FROM_PARENT as _, false),
         };
+        let visual = visualtype.map_or(x11rb::COPY_FROM_PARENT, |v| v.visual_id);
```
kchibisov, line 254: Move this to the place it's used, you can inline it inside the `match`.

src/platform_impl/linux/x11/window.rs
```diff
@@ -214,30 +214,44 @@ impl UnownedWindow {
...
+                all_visuals
                     .find(|(visual, _)| {
                         visual.class == xproto::VisualClass::TRUE_COLOR
                     })
                     .map_or_else(|| {
                         debug!("Could not set transparency, because XMatchVisualInfo returned zero for the required parameters");
-                        (x11rb::COPY_FROM_PARENT as _, x11rb::COPY_FROM_PARENT as _, false)
-                    }, |(visual, depth)| (visual.visual_id, depth, true))
+                        (None as _, x11rb::COPY_FROM_PARENT as _, false)
+                    }, |(visual, depth)| (Some(visual), depth, true))
```
kchibisov, line 250: That looks like `find_map().unwrap_or_else()`.

kchibisov (changes requested):
Looks good otherwise.

Could we maybe set `visual` while at it in `raw-window-handle`?

src/platform_impl/linux/x11/window.rs
notgull: It's also used in `create_window`.

kchibisov: This is impressive how messy this all is, given that colormap code is also using that visual stuff.

notgull: Yes, this backend need a serious refactoring.

src/platform_impl/linux/x11/window.rs
```diff
@@ -214,30 +214,48 @@ impl UnownedWindow {
...
-                    .flat_map(|depth| depth.visuals.iter().map(move |visual| (visual, depth.depth)))
-                    .find(|(visual, _)| {
-                        visual.class == xproto::VisualClass::TRUE_COLOR
+                all_visuals
+                    .find_map(|(visual, depth)| {
+                        if visual.class == xproto::VisualClass::TRUE_COLOR {
+                            Some((Some(visual), depth, true))
+                        } else {
+                            None
+                        }
```
kchibisov, line 249: That looks like `(visual.class ==  xproto::VisualClass::TRUE_COLOR).then_some)`.

kchibisov (approved):

PR #3009, opened 2023-08-05
On X11, set visual_id in raw-window-handle

### LINKED ISSUE #2681: x11: `visual_id` is always zero in provided `XlibWindowHandle`
The implementation of `raw_window_handle` in `linux/x11/window.rs` only fills the `window` field of the `XlibWindowHandle`. It would be nice if it also provided the `visual_id`.

kchibisov:
Fixes #2681.

--

r @notgull

### DIFF
--- a/src/platform_impl/linux/x11/window.rs
+++ b/src/platform_impl/linux/x11/window.rs
@@ -113,6 +113,7 @@ unsafe impl Sync for UnownedWindow {}
 pub(crate) struct UnownedWindow {
     pub(crate) xconn: Arc<XConnection>, // never changes
     xwindow: xproto::Window,            // never changes
+    visual: u32,                        // never changes
     root: xproto::Window,               // never changes
     screen_id: i32,                     // never changes
     cursor: Mutex<CursorIcon>,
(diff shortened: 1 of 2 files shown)


src/platform_impl/linux/x11/window.rs
```diff
@@ -323,6 +324,7 @@ impl UnownedWindow {
         let mut window = UnownedWindow {
             xconn: Arc::clone(xconn),
             xwindow: xwindow as xproto::Window,
+            visual,
```
notgull, line 327: IIRC `visual` can be `x11rb::COPY_FROM_PARENT`, which isn't a valid visual ID; it's just a special value passed to `create_window` and other functions to indicate that the real visual should be copied from the parent. You need to use `get_window_attributes` to get the real visual ID if you don't know it already.

notgull (changes requested):

src/platform_impl/linux/x11/window.rs
```diff
@@ -319,10 +320,19 @@ impl UnownedWindow {
             wid
         };
 
+        if visual == x11rb::COPY_FROM_PARENT {
+            visual = leap!(leap!(xconn
+                .xcb_connection()
+                .get_window_attributes(xwindow as xproto::Window))
+            .reply())
+            .visual;
+        }
```
notgull, line 329: I would add a comment to this block explaining why you're doing this.

notgull (approved):
Looks good to me!

notgull (approved):

PR #3011, opened 2023-08-06
Propagate error from `EventLoop` creation

kchibisov:
Inner panics could make it hard to trouble shoot the issues and for some users it's not desirable.

The inner panics were left only when they are used to `assert!` during development.

This reverts commit 9f91bc413fe20618bd7090829832bb074aab15c3 which reverted the original patch which was merged without a proper review.

### DIFF
--- a/examples/child_window.rs
+++ b/examples/child_window.rs
@@ -36,7 +36,7 @@ fn main() -> Result<(), impl std::error::Error> {
 
     let mut windows = HashMap::new();
 
-    let event_loop: EventLoop<()> = EventLoop::new();
+    let event_loop: EventLoop<()> = EventLoop::new().unwrap();
     let parent_window = WindowBuilder::new()
         .with_title("parent window")
         .with_position(Position::Logical(LogicalPosition::new(0.0, 0.0)))
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -18,7 +18,7 @@ use std::time::{Duration, Instant};
 #[cfg(wasm_platform)]
 use web_time::{Duration, Instant};
 
-use crate::error::RunLoopError;
+use crate::error::EventLoopError;
 use crate::{event::Event, monitor::MonitorHandle, platform_impl};
 
 /// Provides a way to retrieve events from the system and from the windows that were registered to
@@ -88,23 +88,23 @@ impl<T> EventLoopBuilder<T> {
     /// ***For cross-platform compatibility, the [`EventLoop`] must be created on the main thread,
     /// and only once per application.***
     ///
-    /// Attempting to create the event loop on a different thread, or multiple event loops in
-    /// the same application, will panic. This restriction isn't
-    /// strictly necessary on all platforms, but is imposed to eliminate any nasty surprises when
-    /// porting to platforms that require it. `EventLoopBuilderExt::any_thread` functions are exposed
-    /// in the relevant [`platform`] module if the target platform supports creating an event loop on
-    /// any thread.
-    ///
     /// Calling this function will result in display backend initialisation.
+    ///
+    /// ## Panics
+    ///
+    /// Attempting to create the event loop off the main thread will panic. This
+    /// restriction isn't strictly necessary on all platforms, but is imposed to
+    /// eliminate any nasty surprises when porting to platforms that require it.
+    /// `EventLoopBuilderExt::any_thread` functions are exposed in the relevant
+    /// [`platform`] module if the target platform supports creating an event
+    /// loop on any thread.
     ///
     /// ## Platform-specific
     ///
-    /// - **Linux:** Backend type can be controlled using an environment variable
-    ///   `WINIT_UNIX_BACKEND`. Legal values are `x11` and `wayland`.
-    ///   If it is not set, winit will try to connect to a Wayland connection, and if that fails,
-    ///   will fall back on X11. If this variable is set with any other value, winit will panic.
-    /// - **Android:** Must be configured with an `AndroidApp` from `android_main()` by calling
-    ///     [`.with_android_app(app)`] before calling `.build()`.
+    /// - **Wayland/X11:** to prevent running under `Wayland` or `X11` unset `WAYLAND_DISPLAY`
+    ///                    or `DISPLAY` respectively when building the event loop.
+    /// - **Android:** must be configured with an `AndroidApp` from `android_main()` by calling
+    ///     [`.with_android_app(app)`] before calling `.build()`, otherwise it'll panic.
     ///
     /// [`platform`]: crate::platform
     #[cfg_attr(
@@ -116,17 +116,17 @@ impl<T> EventLoopBuilder<T> {
         doc = "[`.with_android_app(app)`]: #only-available-on-android"
     )]
     #[inline]
-    pub fn build(&mut self) -> EventLoop<T> {
+    pub fn build(&mut self) -> Result<EventLoop<T>, EventLoopError> {
         if EVENT_LOOP_CREATED.swap(true, Ordering::Relaxed) {
-            panic!("Creating EventLoop multiple times is not supported.");
+            return Err(EventLoopError::RecreationAttempt);
         }
 
         // Certain platforms accept a mutable reference in their API.
         #[allow(clippy::unnecessary_mut_passed)]
-        EventLoop {
-            event_loop: platform_impl::EventLoop::new(&mut self.platform_specific),
+        Ok(EventLoop {
+            event_loop: platform_impl::EventLoop::new(&mut self.platform_specific)?,
             _marker: PhantomData,
-        }
+        })
     }
 
     #[cfg(wasm_platform)]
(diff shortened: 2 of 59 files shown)


src/event_loop.rs
```diff
@@ -116,17 +113,17 @@ impl<T> EventLoopBuilder<T> {
         doc = "[`.with_android_app(app)`]: #only-available-on-android"
     )]
     #[inline]
-    pub fn build(&mut self) -> EventLoop<T> {
+    pub fn build(&mut self) -> Result<EventLoop<T>, EventLoopError> {
         if EVENT_LOOP_CREATED.swap(true, Ordering::Relaxed) {
-            panic!("Creating EventLoop multiple times is not supported.");
+            return Err(EventLoopError::AlreadyRunning);
```
madsmtm, line 118: Is this really recoverable? When would you ever try to build an event loop where one already exists, and not have that fail?

src/event_loop.rs
```diff
@@ -88,21 +88,18 @@ impl<T> EventLoopBuilder<T> {
...
-    /// porting to platforms that require it. `EventLoopBuilderExt::any_thread` functions are exposed
-    /// in the relevant [`platform`] module if the target platform supports creating an event loop on
-    /// any thread.
+    /// Attempting to create the event loop on a different thread, or multiple
+    /// event loops in the same application, will panic. This restriction
+    /// isn't strictly necessary on all platforms, but is imposed to eliminate
+    /// any nasty surprises when porting to platforms that require it.
+    /// `EventLoopBuilderExt::any_thread` functions are exposed in the relevant
+    /// [`platform`] module if the target platform supports creating an event
+    /// loop on any thread.
```
madsmtm, line 97: I'd suggest using standard sections headers, something like this:
```suggestion
    /// # Panics
    ///
    /// Attempting to create the event loop on a different thread, or multiple
    /// event loops in the same application, will panic. This restriction
    /// isn't strictly necessary on all platforms, but is imposed to eliminate
    /// any nasty surprises when porting to platforms that require it.
    /// `EventLoopBuilderExt::any_thread` functions are exposed in the relevant
    /// [`platform`] module if the target platform supports creating an event
    /// loop on any thread.
```

examples/child_window.rs
```diff
@@ -36,7 +36,7 @@ fn main() -> Result<(), impl std::error::Error> {
 
     let mut windows = HashMap::new();
 
-    let event_loop: EventLoop<()> = EventLoop::new();
+    let event_loop: EventLoop<()> = EventLoop::new().unwrap();
```
madsmtm, line 39: I think we should use `?` in the examples instead, generally looks prettier.

src/lib.rs
```diff
@@ -7,7 +7,7 @@
 //!
 //! ```no_run
 //! use winit::event_loop::EventLoop;
-//! let event_loop = EventLoop::new();
+//! let event_loop = EventLoop::new().unwrap();
```
madsmtm, line 10: `?` in doctests are possible if you put `# Ok::<winit::error::EventLoopError>(())` at the end

madsmtm (changes requested):

src/event_loop.rs
kchibisov: Hm, yeah, we could assert, I guess?

examples/child_window.rs
kchibisov: Well, we use `unwrap` for lots of other things as well in examples, making examples prettier is for the separate PR, I'd say.

src/lib.rs
kchibisov: That's the same as with the rest, we'd have to change for everything at once.

src/event_loop.rs
notgull: I think this should return an error, as there's no other real way to tell if the event loop has otherwise been created.

kchibisov: That was my initial thinking that you should indicate that there's a loop running, but maybe not fail hard.

The thing were we should fail hard is when `main` thread is not used, because it's assertion basically on API misuse.

kchibisov: Well, on the other side it's a user controllable action, because winit tracks its own function calls, to not allow creating event loop again on macOS, because it'll break and we don't really want to support that.

daxpedda (approved):
Web doesn't currently use this feature. It will currently panic if not called from the main thread or if we aren't in a browser.

AFAIK other backends may also panic when the event loop is not created in the main thread, so the Web backend can be adjusted when we decide to change that behavior.

kchibisov:
>AFAIK other backends may also panic when the event loop is not created in the main thread, so the Web backend can be adjusted when we decide to change that behavior.

That's expected, it's a logical error, so we want to catch them early on.

src/event_loop.rs
kchibisov: The issue here is that it's a logical error of the programmer, not an error coming from the outside world or interacting with it. It's just whether you've created the event loop by winit means or not.

src/platform_impl/linux/mod.rs
```diff
@@ -754,65 +738,26 @@ impl<T: 'static> EventLoop<T> {
...
-                    #[cfg(wayland_platform)]
-                    return EventLoop::new_wayland_any_thread()
-                        .expect("Failed to initialize Wayland backend");
-                    #[cfg(not(wayland_platform))]
-                    panic!("wayland feature is not enabled");
-                }
-                _ => panic!(
-                    "Unknown environment variable value for {BACKEND_PREFERENCE_ENV_VAR}, try one of `x11`,`wayland`",
-                ),
-            }
```
notgull, line 789: Is there any reason why you're getting rid of this mechanism for selecting the Linux backend? According to the users that I've spoken with it's a useful tool.

src/platform_impl/linux/wayland/event_loop/mod.rs
```diff
@@ -101,32 +117,52 @@ impl<T: 'static> EventLoop<T> {
...
+                    user_events_channel,
+                    move |event, _, winit_state: &mut WinitState| {
+                        if let calloop::channel::Event::Msg(msg) = event {
+                            winit_state.dispatched_events = true;
+                            pending_user_events_clone.borrow_mut().push(msg);
+                        }
+                    },
+                )
+                .map_err(|error| error.error),
+            WaylandError::Calloop
```
notgull, line 144: For big expressions like this, I usually prefer to leave it outside of the macro in order to preserver proper formatting. Like so:

```rust
let res = / *... */;
map_err!(res, WaylandError::Calloop)
```

notgull (commented):
Should "not running on the main thread" be a type of recoverable error as well?

Aside from my questions, looks good to me!

src/platform_impl/linux/mod.rs
kchibisov: You have a standard mechanism of unsetting `WAYLAND_DISPLAY`, the mechanism is error prone, because if you use other crates, it obviously won't work for them.

The real way to do is unset `WAYLAND_DISPLAY` or `DISPLAY`.

notgull: I suppose so, but as this is a user-facing change it should be explicitly mentioned in the `CHANGELOG.md`, alongside the alternative you gave.

kchibisov:
>Should "not running on the main thread" be a type of recoverable error as well?

it's an API misuse given thare's an API to do that on any thread. API misuse is usually an assert.

notgull (commented):
> it's an API misuse given thare's an API to do that on any thread. API misuse is usually an assert.

IMO it should only be an `assert` if there's a reasonable `if` statement that can be used to avoid hitting the assert. For instance it should be possible for a user to call `if winit::is_main_thread()`, or something.

kchibisov:
>IMO it should only be an assert if there's a reasonable if statement that can be used to avoid hitting the assert. For instance it should be possible for a user to call if winit::is_main_thread(), or something.

I'm not sure why winit should provide that, it's done by the means of `stdlib` or `libc`, it's not a special main thread.

notgull:
> I'm not sure why winit should provide that, it's done by the means of `stdlib` or `libc`, it's not a special main thread.

As far as I know, there is no way to reliably tell if a thread is the "main" thread or not. I guess I could create a crate to tell.

kchibisov:
@notgull we not need a crate, winit already checks for that.

notgull:
> @notgull we not need a crate, winit already checks for that.

Even if `winit` checks for that, how should crates using `winit` check for it, in order to avoid errors? This especially matters for `async-winit`, where I'd like to be able to check for this.

kchibisov:
>Even if winit checks for that, how should crates using winit check for it, in order to avoid errors? This especially matters for async-winit, where I'd like to be able to check for this.

Unless a user explicitly created a `std::thread` and tried to create an event loop on it, they are on the main thread.

notgull:
> Unless a user explicitly created a `std::thread` and tried to create an event loop on it, they are on the main thread.

In library code, you don't know whether or not the user has spawned an `std::thread` or not. I guess the real question is: do we have to worry about this function being called from library code?

kchibisov:
I mean, library creating a winit event loop inside its internal routines and not exposing it to the user sounds a bit weird. And what will you do with the error anyway? You can't do anything about it because it's a hard error you can't do about, so you'll effectively disable the winit entirely, because user is misusing the API.

notgull (approved):
I suppose so. That's the last real issue I had.

PR #3017, opened 2023-08-08
Add a way to embed the X11 window into another

### LINKED ISSUE #1149: XEmbed support
X11 supports embedding windows into each other, which is something that has been requested by a few Alacritty users. I'd say this is probably not something very commonly used and rather niche, so I was curious if winit had interest for supporting it.

There already is a PR for adding this on the Alacritty side (https://github.com/jwilm/alacritty/pull/2786), however it seems to me that I should at least inquire if winit is interested in it. That way we could potentially implement it for more than just Alacritty itself.

Since this is an X11 thing only and it is extremely niche, I feel like it might conflict with winit's simpler approach to things, so maybe the recommended solution would be an external library? Though in that case it's probably going to stay in Alacritty only for now.

notgull:

Closes #1149


kchibisov:
If you want to test this, there's a `tabbed` application doing just xembed (that's the reason we have in alacritty xembed support for example).

notgull:
> If you want to test this, there's a `tabbed` application doing just xembed (that's the reason we have in alacritty xembed support for example).

Oh good, I was worried that I would have to write my own harness. In that case it's ready.

kchibisov (approved):
Seems to work just fine with alacritty

Tested-by: Kirill Chibisov <contact@kchibisov.com>

PR #3021, opened 2023-08-11
Fix touch force for Apple Pencil

lucasmerlin:
This sets the touch force property when the touch event was made with an apple pencil.


kchibisov (approved):

kchibisov:
Thanks!

PR #3024, opened 2023-08-13
Reexport raw-window-handle in window module

kchibisov:
We use raw-window-handle extensive in the public API as well as we force the users to use it to get some essential data for interop, thus reexport it.

Fixes: #2913.

### DIFF
--- a/src/window.rs
+++ b/src/window.rs
@@ -18,6 +18,9 @@ pub use crate::icon::{BadIcon, Icon};
 #[doc(inline)]
 pub use cursor_icon::{CursorIcon, ParseError as CursorIconParseError};
 
+#[doc(inline)]
+pub use raw_window_handle;
+
 /// Represents a window.
 ///
 /// # Example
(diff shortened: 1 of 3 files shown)


src/window.rs
```diff
@@ -18,6 +18,9 @@ pub use crate::icon::{BadIcon, Icon};
 #[doc(inline)]
 pub use cursor_icon::{CursorIcon, ParseError as CursorIconParseError};
 
+#[doc(inline)]
+pub use raw_window_handle;
```
notgull, line 22: Is there any reason for the `doc(inline)`? I think we should emphasize that it is an external crate.

notgull (approved):
Looks good to me!

src/window.rs
kchibisov: mostly to easily read docs. We could move it to `winit::reexports` similar, to what sctk does.

notgull: I see, that makes sense. If we start re-exporting a lot of crates we can do that in the future, but for now I'm fine with this.

PR #3034, opened 2023-08-17
Fix invalid hmonitor panic

xiaopengli89:
```
"called `Result::unwrap()` on an `Err` value: Os { code: 1461, kind: Uncategorized, message: \"无效监视器句柄。\" }" at C:\Users\***\.cargo\git\checkouts\winit-c2fdb27092aba5a7\92592ec\src\platform_impl\windows\monitor.rs:212:51
```


kchibisov:
I guess it could affect other calls here as well?

And I guess it happens basically because no lifetime is attached to the monitor handle, so you stash it somewhere and it's getting sort of dead? I'm pretty sure we've discussed something like that before in https://github.com/rust-windowing/winit/pull/2646 . Not sure if you've read it.

PR #3036, opened 2023-08-20
Fix inconsistency in mouse button device events, add hwheel device event on Windows

valaphee:

While working with device events, I noticed that there was an inconsistency in the mouse button device events between Windows/X11 and for example web, because web uses the same ids/order as the MouseButton enum, and Windows/X11 are using the X11 ids, and hwheel device event was ignored on Windows.

Mouse button device events are now using the same order as the MouseButton enum, and I also added hwheel device events for Windows.


notgull (changes requested):
If the documentation [here](https://www.x.org/releases/current/doc/inputproto/XI2proto.txt) is correct, it says that `detail` might not correspond 1:1 with the mouse press.

> detail: The button number, keycode or touch ID

In addition to the fact that my understanding is that the "raw" device events are intended to correspond 1:1 with the underlying driver data (this isn't possible on Web), so modifying it shouldn't happen. As in, for device events, you really shouldn't be trying to interpret the data directly unless you have a pretty good guess for what the device is.

The windows code looks good to me.

valaphee:
Thanks for the input, I removed the remapping of mouse button ids in X11. Both platforms now return "unmodified" device events.

kchibisov (approved):

kchibisov:
@valaphee could you rebase?

notgull (approved):
Thanks!

Issue #3005, opened 2023-08-04
Custom cursor icons


valaphee:
At the moment there is no way to use custom cursor icons (except doing it manually).

All major platforms should support custom cursors even on web.

Loading cursor icons would work the same way as other icons on Windows (LoadImageW), but would require a new method in Window or not using "cursor_icon" and adding a new variant "Custom" to CursorIcon.

madsmtm:
Three open PRs on this:
- https://github.com/rust-windowing/winit/pull/1617
- https://github.com/rust-windowing/winit/pull/2117
- https://github.com/rust-windowing/winit/pull/2833

Biggest problem is how to design the desired API for this.

kchibisov:
I guess we could do something like 

```
pub enum CursorIcon {
    Named(CursorIcon), // from crate
    Raw(Buffer),
}
```

The issue is that on Wayland you'd need a scaling, but we can accept physical and say that we'll scale based on display resolution?

valaphee:
What was the reason why CursorIcon resides in its own crate? I found this pr https://github.com/rust-windowing/winit/commit/bd9cc2a9da7de62bf2cdd5d220f8c3635c3d82cc but I don't think its really worth being in its own crate.

kchibisov:
@valaphee it's used by multiple crates(not only winit) and just encodes w3c icons

Friz64:
#3218 has been merged 🎉

PR #3039, opened 2023-08-22
Custom cursor icon support

valaphee:

One possible solution for https://github.com/rust-windowing/winit/issues/3005, adds a way to set custom icons in winit using the existing `Window::set_cursor_icon` method.

At the moment this is only a draft, because not all major platforms are implemented/tested
- [ ] iOS
- [ ] X11
- [ ] Wayland
- [ ] MacOS
- [x] Web
- [x] Windows

I can't really test for Apple platforms because I don't own Apple devices.


kchibisov (changes requested):
I'm not sure that reusing `Icon` is a good idea, how does it map to the cursor icon used on the platforms? I'm interested wrt macOS/Web/Windows, since X11/Wayland use just argb/xrgb buffers for all of that.

Also, without hotspot it's all kind of useless.

kchibisov:
Also, see https://github.com/rust-windowing/winit/pull/2833/files

valaphee:
For web: you would have to use css and the "cursor" property, where you can specify a url, which can also be an base64 string, containing the png encoded cursor, but this is already how it works, its just a question of encoding the icon (rgba) to png and packing it into base64, (or ico because there its possible to specify the hotspot)

For iOS: NSCursor has an instance method called initWithImage with NSImage and hotSpot, this would suggest Icon being of type NSImage, but cursor icon doesn't seem to implemented there? `debug!("`Window::set_cursor_icon` ignored on iOS")`

Good point, didn't thought about hotspot, I would add it to CursorIcon::Custom

daxpedda (changes requested):
This isn't really feasible for Web like this.

Firstly `PlatformIcon` has to be split, because it's currently used for both cursor and window, in the case of Web window needs to continue using `NoIcon` but cursor now requires `RgbaIcon`.

Secondly adding the `image` dependency is overkill. I would argue that Web should support links before it starts supporting raw RGBA in the first place anyway. If we really want to support RGBA on Web then we should either use the `png` crate directly or use the WebAPI to do the conversion: `ImageData` -> `ImageBitmap` -> `Canvas` -> data URL.

Adding platform-specific formats shouldn't be a problem as they go through methods on `Icon` anyway.

valaphee:
I definitely agree with the image crate being overkill, and I'll split up Icon (as normal icons also don't need the hotspot property).

Using urls could definitely be an option, and for RGBA, I'll take a look into utilizing the web api.

valaphee:
Split-up Icon and CustomCursorIcon pub import, and switched to png crate, web also doesn't use RgbaImage and instead just the url repr.

src/icon.rs
```diff
@@ -123,7 +139,36 @@ impl Icon {
...
+
+impl fmt::Debug for CustomCursorIcon {
+    fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+        fmt::Debug::fmt(&self.inner, formatter)
+    }
+}
+
+impl CustomCursorIcon {
+    pub fn from_rgba(rgba: Vec<u8>, width: u32, height: u32) -> Result<Self, BadIcon> {
+        Self::from_rgba_with_hotspot(rgba, width, height, width / 2, height / 2)
```
daxpedda, line 160:
```suggestion
        Self::from_rgba_with_hotspot(rgba, width, height, 0, 0)
```
Shouldn't the default be the top-left corner?

src/platform_impl/web/icon.rs
daxpedda, line 1: I believe this is the kinda stuff that goes into `src/platform/web.rs`.

src/platform_impl/web/icon.rs
```diff
@@ -0,0 +1,60 @@
...
+    }
+}
+
+#[derive(Clone)]
+pub struct WebIcon {
+    pub(crate) inner: Cow<'static, str>,
+}
+
+impl WebIcon {
+    pub fn from_url(url: &str, hotspot_x: u32, hotspot_y: u32) -> Self {
```
daxpedda, line 23: Can we add a `with_hotspot()` alternative here?

src/platform_impl/web/mod.rs
```diff
@@ -36,6 +37,7 @@ pub(crate) use self::event_loop::{
 pub use self::monitor::{MonitorHandle, VideoMode};
 pub use self::window::{PlatformSpecificWindowBuilderAttributes, Window, WindowId};
 
+pub(crate) use self::icon::WebIcon as PlatformCustomCursorIcon;
 pub(crate) use self::keyboard::KeyEventExtra;
 pub(crate) use crate::icon::NoIcon as PlatformIcon;
```
daxpedda, line 42:
```suggestion
```

src/platform_impl/web/icon.rs
valaphee: The platform-specific implementation at least for Windows is in the `platform_impl` module, and they are exposed in `platform`, by extending Icon

Basically just copied it from there

madsmtm:
Superseded by https://github.com/rust-windowing/winit/pull/3218

PR #3041, opened 2023-08-24
Fix recent CI failures

madsmtm:
We really need to work on making our CI more stable against upstream updates like this, but oh well, let's just fix it for now.


daxpedda:
> Kind of weird to request a review and then merge right after.

The review requests are automatic because of code ownership.

rdrpenguin04:
Clippy on MSRV isn't always useful; a lot of the lints have been improved in the several versions since.

madsmtm:
> > Kind of weird to request a review and then merge right after.
> 
> The review requests are automatic because of code ownership.

Yeah sorry about that, I wanted to unassign you before I did so, but it seems I forgot

PR #3045, opened 2023-08-25
Make iOS fully thread safe

madsmtm:
Part of #2464.

Now, `MonitorHandle` and `VideoMode` are usable from any thread.

Also did a few other cleanups, in preparation for making `EventLoopWindowTarget` independent of the user event type, to fix the macOS and iOS part of https://github.com/rust-windowing/winit/issues/3053.

  - [x] macOS
  - [x] iOS

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- Make iOS `MonitorHandle` and `VideoMode` usable from other threads.
 - Fix window size sometimes being invalid when resizing on macOS.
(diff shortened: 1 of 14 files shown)


madsmtm:
CI [fails installing `cargo-apk` in MSRV](https://github.com/rust-windowing/winit/actions/runs/5991562597/job/16250028134?pr=3045), notifying @rib @MarijnS95, going to ignore it here.

kchibisov:
@madsmtm it's just broken cargo, don't think about it.

kchibisov:
I've fixed it already thoug.h

MarijnS95:
Looks like a transitive dependency that bumped their MSRV eagerly.

PR #3048, opened 2023-08-25
Use a transparent color for the transparent example

stefnotch:

The transparency example used a solid fill color, which seems a tad odd. This fixes this by using a semi-transparent gray. However, we should wait for softbuffer to support transparency. https://github.com/rust-windowing/softbuffer/issues/17
I only have a Windows machine, so I only managed to test the example on Windows.

The example can be tested with `cargo run --example transparent`

### DIFF
--- a/examples/transparent.rs
+++ b/examples/transparent.rs
@@ -32,7 +32,7 @@ fn main() -> Result<(), impl std::error::Error> {
                 ..
             } => control_flow.set_exit(),
             Event::RedrawRequested(_) => {
-                fill::fill_window(&window);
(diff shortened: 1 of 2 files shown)


stefnotch:
I'd love some advice on how to fix the CI.

madsmtm:
> I'd love some advice on how to fix the CI.

Maybe just slap a `#[allow(dead_code)]` on `fill_window`?

kchibisov:
Allowing dead code sounds fine to me.

madsmtm:
I'm going to close this, as `softbuffer` doesn't support transparency yet. We can revisit this once that happens.

PR #3056, opened 2023-08-27
Move `ControlFlow` to `EventLoopWindowTarget`

### LINKED ISSUE #3042: Move `ControlFlow` as methods on `EventLoopWindowTarget`

daxpedda:
This PR makes the following changes:
- Removes `&mut ControlFlow` from `run()` and adds a setter and getter to `EventLoopWindowTarget` instead.
- Exit was split off and now is `exit()` and `is_exit()` on `EventLoopWindowTarget`.
- User-defined exit codes were removed.

Fixes #3042.

### DIFF
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -175,7 +175,7 @@ pub enum ControlFlow {
 }
 
 impl ControlFlow {
-    /// Sets this to wait until a timeout has expired.
+    /// Creates a [`ControlFlow`] that waits until a timeout has expired.
     ///
     /// In most cases, this is set to [`WaitUntil`]. However, if the timeout overflows, it is
     /// instead set to [`Wait`].
@@ -320,9 +320,9 @@ impl<T> EventLoopWindowTarget<T> {
         self.p.control_flow()
     }
 
-    /// Send a [`LoopExiting`] event and stop the event loop.
+    /// This exits the event loop.
     ///
-    /// [`LoopExiting`]: Event::LoopExiting
+    /// See [`LoopExiting`](Event::LoopExiting).
     pub fn exit(&self) {
         self.p.exit()
     }
@@ -330,8 +330,8 @@ impl<T> EventLoopWindowTarget<T> {
     /// Returns if the [`EventLoop`] is about to stop.
     ///
     /// See [`exit()`](Self::exit).
-    pub fn is_exit(&self) -> bool {
-        self.p.is_exit()
+    pub fn exiting(&self) -> bool {
+        self.p.exiting()
     }
 }
 
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -147,20 +147,12 @@ impl<T> fmt::Debug for EventLoopWindowTarget<T> {
     }
 }
 
-/// Set by the user callback given to the [`EventLoop::run`] method.
+/// Set through [`EventLoopWindowTarget::set_control_flow()`].
 ///
 /// Indicates the desired behavior of the event loop after [`Event::AboutToWait`] is emitted.
 ///
 /// Defaults to [`Poll`].
 ///
-/// ## Persistency
-///
-/// Almost every change is persistent between multiple calls to the event loop closure within a
-/// given run loop. The only exception to this is [`ExitWithCode`] which, once set, cannot be unset.
-/// Changes are **not** persistent between multiple calls to `run_ondemand` - issuing a new call will
-/// reset the control flow to [`Poll`].
-///
-/// [`ExitWithCode`]: Self::ExitWithCode
 /// [`Poll`]: Self::Poll
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
 pub enum ControlFlow {
@@ -180,54 +172,9 @@ pub enum ControlFlow {
     ///
     /// [`Poll`]: Self::Poll
     WaitUntil(Instant),
-
-    /// Send a [`LoopExiting`] event and stop the event loop. This variant is *sticky* - once set,
-    /// `control_flow` cannot be changed from `ExitWithCode`, and any future attempts to do so will
-    /// result in the `control_flow` parameter being reset to `ExitWithCode`.
-    ///
-    /// The contained number will be used as exit code. The [`Exit`] constant is a shortcut for this
-    /// with exit code 0.
-    ///
-    /// ## Platform-specific
-    ///
-    /// - **Android / iOS / Web:** The supplied exit code is unused.
-    /// - **Unix:** On most Unix-like platforms, only the 8 least significant bits will be used,
-    ///   which can cause surprises with negative exit values (`-42` would end up as `214`). See
-    ///   [`std::process::exit`].
-    ///
-    /// [`LoopExiting`]: Event::LoopExiting
-    /// [`Exit`]: ControlFlow::Exit
-    ExitWithCode(i32),
 }
 
 impl ControlFlow {
-    /// Alias for [`ExitWithCode`]`(0)`.
-    ///
-    /// [`ExitWithCode`]: Self::ExitWithCode
-    #[allow(non_upper_case_globals)]
-    pub const Exit: Self = Self::ExitWithCode(0);
-
-    /// Sets this to [`Poll`].
-    ///
-    /// [`Poll`]: Self::Poll
-    pub fn set_poll(&mut self) {
-        *self = Self::Poll;
-    }
-
-    /// Sets this to [`Wait`].
-    ///
-    /// [`Wait`]: Self::Wait
-    pub fn set_wait(&mut self) {
-        *self = Self::Wait;
-    }
-
-    /// Sets this to [`WaitUntil`]`(instant)`.
-    ///
-    /// [`WaitUntil`]: Self::WaitUntil
-    pub fn set_wait_until(&mut self, instant: Instant) {
-        *self = Self::WaitUntil(instant);
-    }
-
     /// Sets this to wait until a timeout has expired.
     ///
     /// In most cases, this is set to [`WaitUntil`]. However, if the timeout overflows, it is
@@ -235,25 +182,11 @@ impl ControlFlow {
     ///
     /// [`WaitUntil`]: Self::WaitUntil
     /// [`Wait`]: Self::Wait
-    pub fn set_wait_timeout(&mut self, timeout: Duration) {
+    pub fn wait_duration(timeout: Duration) -> Self {
         match Instant::now().checked_add(timeout) {
-            Some(instant) => self.set_wait_until(instant),
-            None => self.set_wait(),
+            Some(instant) => Self::WaitUntil(instant),
+            None => Self::Wait,
         }
-    }
-
-    /// Sets this to [`ExitWithCode`]`(code)`.
-    ///
-    /// [`ExitWithCode`]: Self::ExitWithCode
-    pub fn set_exit_with_code(&mut self, code: i32) {
-        *self = Self::ExitWithCode(code);
-    }
-
-    /// Sets this to [`Exit`].
-    ///
-    /// [`Exit`]: Self::Exit
-    pub fn set_exit(&mut self) {
-        *self = Self::Exit;
     }
 }
 
@@ -286,8 +219,7 @@ impl<T> EventLoop<T> {
     /// Since the closure is `'static`, it must be a `move` closure if it needs to
     /// access any data from the calling context.
     ///
-    /// See the [`ControlFlow`] docs for information on how changes to `&mut ControlFlow` impact the
-    /// event loop's behavior.
+    /// See the [`set_control_flow()`] docs on how to change the event loop's behavior.
     ///
     /// ## Platform-specific
     ///
@@ -306,12 +238,12 @@ impl<T> EventLoop<T> {
     ///
     ///   This function won't be available with `target_feature = "exception-handling"`.
     ///
-    /// [`ControlFlow`]: crate::event_loop::ControlFlow
+    /// [`set_control_flow()`]: EventLoopWindowTarget::set_control_flow
     #[inline]
     #[cfg(not(all(wasm_platform, target_feature = "exception-handling")))]
     pub fn run<F>(self, event_handler: F) -> Result<(), EventLoopError>
     where
-        F: FnMut(Event<T>, &EventLoopWindowTarget<T>, &mut ControlFlow),
+        F: FnMut(Event<T>, &EventLoopWindowTarget<T>),
     {
         self.event_loop.run(event_handler)
     }
@@ -376,6 +308,30 @@ impl<T> EventLoopWindowTarget<T> {
     /// [`DeviceEvent`]: crate::event::DeviceEvent
     pub fn listen_device_events(&self, allowed: DeviceEvents) {
         self.p.listen_device_events(allowed);
+    }
+
+    /// Sets the [`ControlFlow`].
+    pub fn set_control_flow(&self, control_flow: ControlFlow) {
+        self.p.set_control_flow(control_flow)
+    }
+
+    /// Gets the current [`ControlFlow`].
+    pub fn control_flow(&self) -> ControlFlow {
+        self.p.control_flow()
+    }
+
+    /// Send a [`LoopExiting`] event and stop the event loop.
+    ///
+    /// [`LoopExiting`]: Event::LoopExiting
+    pub fn exit(&self) {
+        self.p.exit()
+    }
+
+    /// Returns if the [`EventLoop`] is about to stop.
+    ///
+    /// See [`exit()`](Self::exit).
+    pub fn is_exit(&self) -> bool {
+        self.p.is_exit()
     }
 }
 
--- a/src/platform/pump_events.rs
+++ b/src/platform/pump_events.rs
@@ -2,7 +2,7 @@ use std::time::Duration;
 
 use crate::{
     event::Event,
-    event_loop::{ControlFlow, EventLoop, EventLoopWindowTarget},
+    event_loop::{EventLoop, EventLoopWindowTarget},
 };
 
 /// The return status for `pump_events`
@@ -63,7 +63,7 @@ pub trait EventLoopExtPumpEvents {
     ///
     ///     'main: loop {
     ///         let timeout = Some(Duration::ZERO);
-    ///         let status = event_loop.pump_events(timeout, |event, _, control_flow| {
+    ///         let status = event_loop.pump_events(timeout, |event, elwt| {
     /// #            if let Event::WindowEvent { event, .. } = &event {
     /// #                // Print only Window events to reduce noise
     /// #                println!("{event:?}");
@@ -73,7 +73,7 @@ pub trait EventLoopExtPumpEvents {
     ///                 Event::WindowEvent {
     ///                     event: WindowEvent::CloseRequested,
     ///                     window_id,
-    ///                 } if window_id == window.id() => control_flow.set_exit(),
+    ///                 } if window_id == window.id() => elwt.exit(),
     ///                 Event::AboutToWait => {
     ///                     window.request_redraw();
     ///                 }
@@ -174,7 +174,7 @@ pub trait EventLoopExtPumpEvents {
     ///   callback.
     fn pump_events<F>(&mut self, timeout: Option<Duration>, event_handler: F) -> PumpStatus
     where
-        F: FnMut(Event<Self::UserEvent>, &EventLoopWindowTarget<Self::UserEvent>, &mut ControlFlow);
+        F: FnMut(Event<Self::UserEvent>, &EventLoopWindowTarget<Self::UserEvent>);
 }
 
 impl<T> EventLoopExtPumpEvents for EventLoop<T> {
@@ -182,7 +182,7 @@ impl<T> EventLoopExtPumpEvents for EventLoop<T> {
 
     fn pump_events<F>(&mut self, timeout: Option<Duration>, event_handler: F) -> PumpStatus
     where
-        F: FnMut(Event<Self::UserEvent>, &EventLoopWindowTarget<Self::UserEvent>, &mut ControlFlow),
+        F: FnMut(Event<Self::UserEvent>, &EventLoopWindowTarget<Self::UserEvent>),
     {
         self.event_loop.pump_events(timeout, event_handler)
     }
--- a/src/platform_impl/linux/mod.rs
+++ b/src/platform_impl/linux/mod.rs
@@ -858,8 +858,8 @@ impl<T> EventLoopWindowTarget<T> {
         x11_or_wayland!(match self; Self(evlp) => evlp.exit())
     }
 
-    pub(crate) fn is_exit(&self) -> bool {
-        x11_or_wayland!(match self; Self(evlp) => evlp.is_exit())
+    pub(crate) fn exiting(&self) -> bool {
+        x11_or_wayland!(match self; Self(evlp) => evlp.exiting())
     }
 
     fn set_exit_code(&self, code: i32) {
--- a/src/platform_impl/linux/mod.rs
+++ b/src/platform_impl/linux/mod.rs
@@ -19,7 +19,7 @@ use crate::platform::x11::XlibErrorHook;
 use crate::{
     dpi::{PhysicalPosition, PhysicalSize, Position, Size},
     error::{EventLoopError, ExternalError, NotSupportedError, OsError as RootOsError},
-    event::{Event, KeyEvent},
+    event::KeyEvent,
     event_loop::{
         AsyncRequestSerial, ControlFlow, DeviceEvents, EventLoopClosed,
         EventLoopWindowTarget as RootELW,
@@ -767,21 +767,21 @@ impl<T: 'static> EventLoop<T> {
 
     pub fn run<F>(mut self, callback: F) -> Result<(), EventLoopError>
     where
-        F: FnMut(crate::event::Event<T>, &RootELW<T>, &mut ControlFlow),
+        F: FnMut(crate::event::Event<T>, &RootELW<T>),
     {
         self.run_ondemand(callback)
     }
 
     pub fn run_ondemand<F>(&mut self, callback: F) -> Result<(), EventLoopError>
     where
-        F: FnMut(crate::event::Event<T>, &RootELW<T>, &mut ControlFlow),
+        F: FnMut(crate::event::Event<T>, &RootELW<T>),
     {
         x11_or_wayland!(match self; EventLoop(evlp) => evlp.run_ondemand(callback))
     }
 
     pub fn pump_events<F>(&mut self, timeout: Option<Duration>, callback: F) -> PumpStatus
     where
-        F: FnMut(crate::event::Event<T>, &RootELW<T>, &mut ControlFlow),
+        F: FnMut(crate::event::Event<T>, &RootELW<T>),
     {
         x11_or_wayland!(match self; EventLoop(evlp) => evlp.pump_events(timeout, callback))
     }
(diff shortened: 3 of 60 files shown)


src/platform_impl/macos/app_state.rs
```diff
@@ -16,18 +16,21 @@ use icrate::Foundation::{is_main_thread, NSSize};
...
 use crate::{
     dpi::PhysicalSize,
     event::{Event, InnerSizeWriter, StartCause, WindowEvent},
-    event_loop::{ControlFlow, EventLoopWindowTarget as RootWindowTarget},
+    event_loop::EventLoopWindowTarget as RootWindowTarget,
     window::WindowId,
 };
 
-static HANDLER: Lazy<Handler> = Lazy::new(Default::default);
+pub(crate) static HANDLER: Lazy<Handler> = Lazy::new(Default::default);
```
madsmtm, line 33: Nit: So far we've only exposed this to other modules through `AppState`, I think I'd like to keep it that way (makes it easier to refactor later on).

madsmtm: (So add a `set_control_flow` method to `AppState` instead)

src/platform_impl/macos/app_state.rs
```diff
@@ -85,27 +85,35 @@ impl<T> Debug for EventLoopHandler<T> {
...
             } else {
-                (callback)(event.userify(), &this.window_target, control_flow);
+                None
+            };
+
+            (callback)(event.userify(), &this.window_target);
+
+            if let Some(code) = is_exit {
+                HANDLER.set_control_flow(ControlFlow::ExitWithCode(code))
             }
```
madsmtm, line 100: Nit: I think it makes sense to move this check to `set_control_flow` instead, that way we're more sure it can never get reset after being set to `ExitWithCode`.

src/event_loop.rs
```diff
@@ -378,6 +262,72 @@ impl<T> EventLoopWindowTarget<T> {
...
+    /// [`LoopExiting`]: Event::LoopExiting
+    /// [`set_exit()`]: Self::set_exit
+    pub fn set_exit_with_code(&self, code: i32) {
+        self.p.set_exit_with_code(code)
+    }
+
+    /// Alias for [`set_exit_with_code()`]`(0)`.
+    ///
+    /// [`set_exit_with_code()`]: Self::set_exit_with_code
+    pub fn set_exit(&self) {
```
madsmtm, line 328: I think this one makes sense to call `exit` instead of `set_exit`; even though it internally works like the others, it's much more understandable as an "action" rather than setting some state.

src/event_loop.rs
```diff
@@ -378,6 +262,72 @@ impl<T> EventLoopWindowTarget<T> {
...
+    /// Useful for implementing efficient timers. Applications which want to render at the display's
+    /// native refresh rate should instead use [`set_poll()`] and the VSync functionality of a graphics API
+    /// to reduce odds of missed frames.
+    ///
+    /// [`set_poll()`]: Self::set_poll
+    pub fn set_wait_until(&self, instant: Instant) {
+        self.p.set_wait_until(instant)
+    }
+
+    /// Sets this to wait until a timeout has expired.
```
notgull, line 291: "this" => "the event loop"

src/event_loop.rs
```diff
@@ -378,6 +262,72 @@ impl<T> EventLoopWindowTarget<T> {
...
+    /// ## Platform-specific
+    ///
+    /// - **Android / iOS / Web:** The supplied exit code is unused.
+    /// - **Unix:** On most Unix-like platforms, only the 8 least significant bits will be used,
+    ///   which can cause surprises with negative exit values (`-42` would end up as `214`). See
+    ///   [`std::process::exit`].
+    ///
+    /// [`LoopExiting`]: Event::LoopExiting
+    /// [`set_exit()`]: Self::set_exit
+    pub fn set_exit_with_code(&self, code: i32) {
```
madsmtm, line 321: Now that `EventLoop::run` returns (on most platforms), do we actually need this? I get that it's convenient, but it's not really cross-platform, and applications that want to return a proper status code can just do so.

madsmtm (approved):

src/platform_impl/macos/app_state.rs
daxpedda: Good idea!

notgull (commented):
Looks good from a brief look

src/event_loop.rs
notgull: Yeah, now that I think about it the whole "exit code" thing might not be necessary now that `EventLoop` isn't the root loop of the application.

daxpedda: Agreed, right now it actually just returns an error code in `EventLoopError::ExitFailure(code)`, that sounds pretty useless to me.

src/event_loop.rs
daxpedda: I'm gonna leave this to later until we have a consensus if we want getters or not.

daxpedda:
@madsmtm could you check the iOS implementation?
I couldn't really figure out what to do with the state changes, maybe everything works as is, I don't really get what exactly is going on there :sweat_smile:.

madsmtm:
> I don't really get what exactly is going on there

Honestly me neither, been meaning to rewrite it for a while now. I'll give it a short test drive, I suspect the situation is going to be somewhat improved (previously, there were certain events for which setting the control flow was ignored).

src/event_loop.rs
daxpedda: Done.

src/platform_impl/macos/app_state.rs
daxpedda: Done.

src/platform_impl/macos/app_state.rs
daxpedda: Done.

madsmtm (approved):
MacOS and iOS impl approved

src/event_loop.rs
dhardy: Did exit codes just vanish? Not that I need them.

daxpedda: Yes.
See https://github.com/rust-windowing/winit/pull/3056#discussion_r1306728448.

src/event_loop.rs
daxpedda: Done.

daxpedda:
I still need to go over the documentation again.
And we still need to figure out if we want getters. I'm in favor of at least adding a getter for exit. WDYT?

dhardy:
Personally I like the previous model: it's possible to `match control_flow { ... }`, and it's clear what is going on.

Now: if I call `elwt.set_poll()` then `elwt.set_wait_until(instant)`, does the latter override or does poll take priority?

I would at least like getters for `poll`, `exit` status (sufficient to adapt my current usage).

daxpedda:
> Personally I like the previous model: it's possible to `match control_flow { ... }`, and it's clear what is going on.

I think the Winit API in general is quite inconsistent with that, there are plenty of things that have setters but no getters. Events as well, some events can be queried with, others can't.

I would argue that the API would be better if all for any setter and I don't think it hurts to be able to query any event instead of having to record events in case 
> Now: if I call `elwt.set_poll()` then `elwt.set_wait_until(instant)`, does the latter override or does poll take priority?

Why would anything "take priority"? `set_exit()` is the only one that can't be overwritten which is documented appropriately.

daxpedda:
Just some thought I wanna drop here:

The original motivation was to add platform-specific control flow variants. My thinking was to remove the enum in favor of setters to be able to add platform-specific methods through extension traits, which is the way Winit has continued to add platform-specific functionality until now.

I was quite frustrated with the fact that we removed the `ControlFlow` type, because I think it makes for a very clear API compared to any getters or setters we could possibly introduce: an enum clearly states that it can only be one of those variants, this way the code documents itself.

In addition, this is a problem not specific to `ControlFlow`, we have already added event variants that are platform-specific that don't make any sense to be there in a cross-platform API. The extension trait model we are currently using just doesn't work with enums.

If we want to continue using the extension trait model for platform-specific features, I came up with the following idea that we could use for enums:
```rust
enum Event {
    EventA,
    EventB(String),
    EventC {
        data: u16,
    },
    Platform(PlatformEvent),
}

impl PlatformEventExtWayland for PlatformEvent {
    fn event(&self) -> WaylandEvent {
        self.0
    }
}

// Not really needed, just a neat idea.
impl PlatformEventExtWindows for PlatformEvent {
    fn event(&self) -> ! {
        match self.0 { }
    }
}
```

An alternative of course could be to simply mark the enum `#[non_exhausive]` and add `cfg`-gated variants.

---

Personally I would be in favor of the first solution. We then could add a `(set_)control_flow()` method to `EventLoopWindowTarget` that returns/takes the `ControlFlow` type. Keeping the setters for convenience might still be a good idea. In any case, can be done in a follow-up PR.

WDYT?

src/platform/pump_events.rs
```diff
@@ -2,7 +2,7 @@ use std::time::Duration;
 
 use crate::{
     event::Event,
-    event_loop::{ControlFlow, EventLoop, EventLoopWindowTarget},
+    event_loop::{EventLoop, EventLoopWindowTarget},
 };
 
 /// The return status for `pump_events`
```
madsmtm, line 8: `PumpStatus` also no longer need the exit code

daxpedda: I was just about to write an issue for that. It's still used by Linux and Windows to return error codes. Will post in a couple.

daxpedda: https://github.com/rust-windowing/winit/issues/3059.

madsmtm:
I think this PR is still an improvement, I'd very much prefer to get rid of `&mut ControlFlow`, since it allows us more flexibility in how we store the information.

Whether the exact user-facing API uses `elwt.[set_]control_flow(...)` or distinct getter/setters is a little irrelevant to me, although I prefer the methods because of the easy extensibility (and slightly looser semantics, which is nice for platform implementers ;)) ).

I will say, I don't think `Exit` has a place in the enum, it's really a separate thing. If you want to have an enum, I think it should be something like `enum NextEvent { Poll, Wait, WaitUntil(Instant) }`.

rib:
I've never been a fan of Winit's `&mut ControlFlow` design due to how it only allows for one thing to implement timers based on `WaitUntil`, which in turn means you can't have portable timers unless a framework on top of winit provides timers, which I find far from ideal for a portable event loop abstraction. (I.e. right now it's not possible to write a utility crate that only depends on Winit and can schedule timers that won't conflict without how the application updates the `control_flow` state).

Timers are such a basic building block for applications so it's always seemed odd to me that Winit doesn't support them.

I'd love to see Winit provide a portable timer API which would replace the main use of `ControlFlow` (`Poll` is an special case of a zero timeout and `Wait` is a special case of an indefinite timeout because there are no timers), and then there could be some separate `EventLoop/EventLoopWindowTarget` API for requesting to exit the loop.

kchibisov:
I'm not sure winit should do that directly though. Maybe in a separate lib, because you can clearly implement timers based on `WaitUntil` which alacritty does for example.

dhardy:
> Why would anything "take priority"? set_exit() is the only one that can't be overwritten which is documented appropriately.

This is an example of why control-flow is *not* a series of setters/getters: `set_wait_until(..)` and `set_poll` do not represent independent properties.

> An alternative of course could be to simply mark the enum #[non_exhausive] and add cfg-gated variants.

`Event` should probably be `#[non_exhaustive]` anyway, just so that Winit v1.1 can add new events.

On the whole I am not a fan of the platform-extension-trait model. For an app targetting only one platform the model is fine, but that's mostly not why you'd use Winit anyway. A cross-platform API should require as little platform-specific code to use as possible. If, say, `Event::PinchGesture` is only generated on MacOS but exists everywhere, then removing unused handlers should be an optimisation problem only (i.e. the app should be able to just write a cross-platform handler for `Event::PinchGesture`). Adding `#[cfg(macos_platform)]` to this isn't a big deal if we must have a minimal API on each platform but is not necessary. (Further motivation for this is that if, later, `PinchGesture` were supported on Wayland, the only change needed for an app to enable it there too would be a one-line `cfg`, and only then if a `cfg` was used.)

rib:
> I'm not sure winit should do that directly though. Maybe in a separate lib, because you can clearly implement timers based on `WaitUntil` which alacritty does for example.

To my mind it _really_ needs to be Winit that supports timers otherwise they aren't portable.

Even if you had a library that supported timers based on something like `control_flow` nothing is able to assume that that library is usable because nothing guarantees that every app based on Winit will use that common library.

Portability for timers is only really practical / usable when they are supported by Winit directly.

Handled in external libraries just means every app / framework has to come up with their own timer abstraction and although the could potentially use a utility crate for that, that in itself isn't really that beneficial (portability is more useful here imho than being able to share a small amount of utility code).

notgull:
I'd like to point out that `async-winit` has a [`Timer`](https://docs.rs/async-winit/latest/async_winit/struct.Timer.html) primitive built on `winit`'s `set_wait_until` functionality.

I like the idea of having a `ControlFlow` enum still, but just having it be get/set on the ELWT.

dhardy:
@rib the issue I see there is that a "portable timer" is only really useful when a `Wake` event can be delivered to the thing requesting being woken. If multiple pieces of code interact with an event-loop directly then how does the main event handler know what to do with `ResumeTimeReached`?

The way I've handled it in Kas is by storing (roughly) a `Vec<(Instant, Id)>`, sorted by time, then sending `Event::TimerUpdate` to the accompanying `Id`. If Winit were to store a list of wake-times it could attach a `u64` or something, but the loop would still not know what to do with `ResumeTimeReached`.

rib:
> @rib the issue I see there is that a "portable timer" is only really useful when a Wake event can be delivered to the thing requesting being woken.

If Winit supported orthogonal timers then there would be some kind of handle for a timer that you create and you'd get an event like `TimerElapsed(timer_handle)` that tells you when a specific timer handle has elapsed its time.

src/event_loop.rs
daxpedda: I guess we will figure out getters in #3064.
Currently my hope is to be able to add back the enum and just expose a setter and getter for the enum.

In the meantime I have renamed `set_exit()` to `exit()`.

src/platform_impl/linux/mod.rs
```diff
@@ -882,22 +881,44 @@ impl<T> EventLoopWindowTarget<T> {
...
-        callback(evt, target, control_flow)
+#[derive(Clone, Copy)]
+pub(crate) enum ControlFlow {
+    Poll,
+    Wait,
+    WaitUntil(Instant),
+    ExitWithCode(i32),
+}
+
+impl Default for ControlFlow {
```
kchibisov, line 918: Given that you clearly share it across the backends could you put it in a common place, so all backends can use? Maybe we should create a `common` in `platform_impl` and have ` `common/event_loop.rs` there? So such stuff is shared?

src/platform_impl/macos/app_state.rs
```diff
@@ -291,6 +276,15 @@ impl Handler {
         *self.control_flow.lock().unwrap()
     }
 
+    fn set_control_flow(&self, new_control_flow: ControlFlow) {
+        let mut control_flow = self.control_flow.lock().unwrap();
+
+        if let ControlFlow::Exit = *control_flow {
+        } else {
+            *control_flow = new_control_flow
+        }
```
kchibisov, line 285: Could you rewrite such things to simply use `!matches!` or `!=`?

kchibisov (changes requested):
In general it cleans up code a lot, so looks good.

Could you refactor common code out and also not use `if let` with empty `then`? Just use `!matches` or `!=`, it just looks weird when the main logic is inside the else.

kchibisov:
Besides, I'd suggest to add a getter, just because this code sort of had getter before and not adding it now could break some clients and make them not work at all...

src/platform_impl/linux/mod.rs
daxpedda: This is on purpose. The original motivation here is to allow platform-specific control flow, which I would implement right after for Web.

src/event_loop.rs
kchibisov: You can't really have the `getters` and `setters` on the enum, they are part of the `EventLoop`. The getter on the `EventLoopWindowTarget` should just return a `ControlFlow`, we could also unify the way `ControlFlow` is being set, so it's just a method taking a `ControlFlow` in.

We can't have a separate methods on the `enum`, simply because we need to maintain state for the `exit` thing, but as it was said, `Exit` is a separate concept, and likely doesn't need to be merged in the `ControlFlow` exposed to the user, and event internally it should likely be just a bool `exiting` stored on the `EventLoopWindowTarget` or `EventLoop` directly.

daxpedda: > You can't really have the `getters` and `setters` on the enum, they are part of the `EventLoop`. The getter on the `EventLoopWindowTarget` should just return a `ControlFlow`, we could also unify the way `ControlFlow` is being set, so it's just a method taking a `ControlFlow` in.

That is what I meant!

src/event_loop.rs
```diff
@@ -378,6 +262,54 @@ impl<T> EventLoopWindowTarget<T> {
...
+    }
+
+    /// Send a [`LoopExiting`] event and stop the event loop. This variant is *sticky* - once set,
+    /// it cannot be changed, and any future attempts to do so will
+    /// result in it being reset to it's previous state.
+    ///
+    /// [`LoopExiting`]: Event::LoopExiting
+    pub fn exit(&self) {
+        self.p.exit()
+    }
```
kchibisov, line 312: There's no variant being set or sticky, just say simply that `request loop to exit, which will result in `LoopExiting`, etc, etc`. If we separate `ControlFlow` from `exit` stuff it'll sound sort of naturally.

src/platform_impl/linux/mod.rs
kchibisov: Hm, but the thing is that you could have a `#[cfg]` guard on the variant as well, so I don't see it being an issue especially if we merge the way control flow is set into a single method?

src/event_loop.rs
daxpedda: I have to make a second pass over the documentation again, will fix it then.

src/platform_impl/linux/mod.rs
daxpedda: See #3064, this was what I was discussing there. I'm happy to change the PR if we can reach a consensus now, I would very much prefer having a single getter and setter method.

src/event_loop.rs
```diff
@@ -378,6 +262,54 @@ impl<T> EventLoopWindowTarget<T> {
...
+            None => self.set_wait(),
+        }
+    }
+
+    /// Send a [`LoopExiting`] event and stop the event loop. This variant is *sticky* - once set,
+    /// it cannot be changed, and any future attempts to do so will
+    /// result in it being reset to it's previous state.
+    ///
+    /// [`LoopExiting`]: Event::LoopExiting
+    pub fn exit(&self) {
```
kchibisov, line 310: Also, given that users may want to propagate the error somehow, maybe we should allow them to pass `payload` with the `exit` somehow? Likely via the means of dynamic dispatch or straight a `Result<(), Box<dyn Error>>`? But it could be handled in a separate PR, it's just a pain point right now to propagate your error outside of the event loop `run` methods.

src/platform_impl/macos/app_state.rs
daxpedda: Thanks, done!

src/event_loop.rs
daxpedda: Yeah I would prefer to figure that out in a separate PR, assuming the discussion in #3059 hasn't concluded the issue already.

daxpedda:
I decided to redo this PR tomorrow. I was hoping to clear up #3064 first, but there doesn't seem to be a clear consensus yet and I don't want to delay this PR any further considering how much code it touches.

Planned changes:
- Reintroduce the `ControlFlow` enum with `non_exhaustive` and `cfg` guards for platform-specific variants.
- Single getter and setter on `EventLoopWindowTarget`.
- Split off `Exit`.

dhardy:
Then as [above](https://github.com/rust-windowing/winit/pull/3056#issuecomment-1695566561) I'd suggest having a getter and setter on `elwt` instead of passing the user a `&mut ControlFlow`.

src/platform_impl/macos/app_state.rs
```diff
@@ -654,7 +651,7 @@ impl AppState {
         let wait_timeout = HANDLER.wait_timeout(); // configured by pump_events
         let app_timeout = match HANDLER.control_flow() {
             ControlFlow::Wait => None,
-            ControlFlow::Poll | ControlFlow::ExitWithCode(_) => Some(Instant::now()),
```
daxpedda, line 657: So I think this is handled a couple of lines above?
```rust
if HANDLER.is_exit() {
    Self::stop();
}
```
But have no clue actually.

src/platform_impl/linux/mod.rs
daxpedda: This is done.

src/event_loop.rs
```diff
@@ -378,6 +262,54 @@ impl<T> EventLoopWindowTarget<T> {
...
+    /// instead uses [`set_wait()`].
+    ///
+    /// [`set_wait_until()`]: Self::set_wait_until
+    /// [`set_wait()`]: Self::set_wait
+    pub fn set_wait_timeout(&self, timeout: Duration) {
+        match Instant::now().checked_add(timeout) {
+            Some(instant) => self.set_wait_until(instant),
+            None => self.set_wait(),
+        }
+    }
```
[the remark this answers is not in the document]
daxpedda: Done.

src/event_loop.rs
daxpedda: Done.

daxpedda:
This is ready now. I'm gonna update OP in a moment.

There was a bunch of this code in various backends:
```rust
match ... {
    ...
    ControlFlow::ExitWithCode(_) => unreachable!(),
}
```
I just removed them because I assume that they were there to exhaust `ControlFlow` variants, should I add an `assert!()` or something instead?

src/event_loop.rs
```diff
@@ -180,52 +172,14 @@ pub enum ControlFlow {
...
-    pub fn set_wait(&mut self) {
-        *self = Self::Wait;
-    }
-
     /// Sets this to [`WaitUntil`]`(instant)`.
     ///
     /// [`WaitUntil`]: Self::WaitUntil
-    pub fn set_wait_until(&mut self, instant: Instant) {
-        *self = Self::WaitUntil(instant);
+    pub fn new_wait_until(instant: Instant) -> Self {
```
dhardy, line 181: This method is useless: users can already write `ControlFlow::WaitUntil(instant)`

src/event_loop.rs
```diff
@@ -235,26 +189,12 @@ impl ControlFlow {
     ///
     /// [`WaitUntil`]: Self::WaitUntil
     /// [`Wait`]: Self::Wait
-    pub fn set_wait_timeout(&mut self, timeout: Duration) {
+    pub fn new_wait_timeout(timeout: Duration) -> Self {
```
dhardy, line 192: Rename to `wait_duration` ?

dhardy (commented):
Looks good other than the nits

src/event_loop.rs
daxpedda: Ah, good point!

src/event_loop.rs
daxpedda: Hm, I think I agree.

src/event_loop.rs
daxpedda: Done.

src/event_loop.rs
daxpedda: Done.

src/event_loop.rs
```diff
@@ -180,81 +172,22 @@ pub enum ControlFlow {
...
     /// [`Wait`]: Self::Wait
-    pub fn set_wait_timeout(&mut self, timeout: Duration) {
+    pub fn wait_duration(timeout: Duration) -> Self {
         match Instant::now().checked_add(timeout) {
-            Some(instant) => self.set_wait_until(instant),
-            None => self.set_wait(),
+            Some(instant) => Self::WaitUntil(instant),
+            None => Self::Wait,
         }
     }
```
kchibisov, line 190: The function changed, however it doesn't explain what it wants to do.

src/event_loop.rs
```diff
@@ -377,6 +309,30 @@ impl<T> EventLoopWindowTarget<T> {
...
+
+    /// Gets the current [`ControlFlow`].
+    pub fn control_flow(&self) -> ControlFlow {
+        self.p.control_flow()
+    }
+
+    /// Send a [`LoopExiting`] event and stop the event loop.
+    ///
+    /// [`LoopExiting`]: Event::LoopExiting
+    pub fn exit(&self) {
```
kchibisov, line 326: The issue here is that we don't actually send anything, we basically say that we want to exit event loop.

src/event_loop.rs
```diff
@@ -377,6 +309,30 @@ impl<T> EventLoopWindowTarget<T> {
...
+        self.p.exit()
+    }
+
+    /// Returns if the [`EventLoop`] is about to stop.
+    ///
+    /// See [`exit()`](Self::exit).
+    pub fn is_exit(&self) -> bool {
+        self.p.is_exit()
+    }
```
kchibisov, line 335: `is_exit` is a pretty bad name for internal state (event loop is not a door in the end to be marked as exit). `exiting` should be better, since it reflects the `state` not what it is.

It's also a bit complicated when it comes to deref impl on the `EventLoop`, since you can sort of exit it (in case of ondemand/pump_events), but you can rerun it. Do we really need the getter for the exit status in the end?

src/platform_impl/linux/wayland/event_loop/mod.rs
```diff
@@ -225,35 +222,29 @@ impl<T: 'static> EventLoop<T> {
...
             self.loop_running = false;
 
-            let mut dummy = self.control_flow;
-            sticky_exit_callback(
-                Event::LoopExiting,
-                self.window_target(),
-                &mut dummy,
-                &mut callback,
-            );
+            callback(Event::LoopExiting, self.window_target());
```
kchibisov, line 247: This `with_code` and `exit_code` is weird given that we basically return an error and can just forward error, I think the `ExitWithCode` should just be entirely removed out of it and the error must be propagated via the means of `Result`. There's an issue on how users should get their error out of the event loop, but they can clearly just store to external variable as of now.

kchibisov (changes requested):

src/event_loop.rs
dhardy: So `terminate` and `is_terminating` are better names?

src/event_loop.rs
daxpedda: Did it change? Or do you just mean the documentation has to be adjusted to not say "set"?

src/platform_impl/linux/wayland/event_loop/mod.rs
daxpedda: This has already been done like you described. `ControlFlow::ExitWithCode` was removed, `with_code()` and `exit_code()` is internal stuff to some backends that where using `ControlFlow::ExitWithCode` before to pass down non-user error codes.

Cc #3059.

daxpedda: I've added an additional notice to the changelog about the removal of `ExitWithCode`.

src/event_loop.rs
daxpedda: I changed it to `exiting()`.

> It's also a bit complicated when it comes to deref impl on the `EventLoop`, since you can sort of exit it (in case of ondemand/pump_events), but you can rerun it. Do we really need the getter for the exit status in the end?

`exiting()` returns false after the `EventLoop` is reset, so shouldn't that be fine?

src/event_loop.rs
daxpedda: I assume the problem here was the documentation, which I adjusted.

src/event_loop.rs
daxpedda: I adjusted the documentation a bit.

src/event_loop.rs
kchibisov: I guess.

src/event_loop.rs
kchibisov: We have an event called `LoopExiting`, not `LoopTerminating`, so using the word terminate will be inconsistent with the event names.

src/platform_impl/linux/wayland/event_loop/mod.rs
kchibisov: Hm, I guess I just got confused why we even use code, when we can just forward error, like the code arises from the error simply by using `1`, but we already have an error in place which we can bubble up?

kchibisov: Yeah, it covers it.

kchibisov (approved):

PR #3058, opened 2023-08-28
Allow the user to force X11 under Wayland

### LINKED ISSUE #3057: Forcing X11 doesn't work
Since #3011 forcing X11 with `with_x11` doesn't work.

The problem is here:
https://github.com/rust-windowing/winit/blob/7541220a419f0e383251c37be574fef9fd5827e4/src/platform_impl/linux/mod.rs#L754C1-L764C10
```rust
#[cfg(wayland_platform)]
if attributes.forced_backend == Some(Backend::Wayland)
    || env::var("WAYLAND_DISPLAY").is_ok()
{
    return EventLoop::new_wayland_any_thread().map_err(Into::into);
}

#[cfg(x11_platform)]
if attributes.forced_backend == Some(Backend::X) || env::var("DISPLAY").is_ok() {
    return Ok(EventLoop::new_x11_any_thread().unwrap());
}
```

If `WAYLAND_DISPLAY` is set, X never gets the chance to force itself.

notgull:
Closes #3057 by making it so the forced backend is considered over all else.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On Web, remove any fullscreen requests from the queue when an external fullscreen activation was detected.
 - On Wayland, fix `TouchPhase::Canceled` being sent for moved events.
 - Mark `startup_notify` unsafe functions as safe.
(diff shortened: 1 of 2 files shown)


kchibisov (changes requested):
I think it needs a CHANGELOG because it's a regression in beta.

kchibisov (approved):

PR #3060, opened 2023-08-28
Support KDE Wayland blur protocol

sh7dm:
Related to #538

- :x:  Tested on all platforms changed
  - [x] Wayland (KWin)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - **Breaking:** Change default `ControlFlow` from `Poll` to `Wait`.
 - **Breaking:** remove `DeviceEvent::Text`.
 - On Android, fix `DeviceId` to contain device id's.
-- Added initial blur support, currently limited to KDE Wayland
+- Add `Window::set_blur` to request a blur behind the window; implemented on Wayland for now.
 
 # 0.29.1-beta
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - **Breaking:** Change default `ControlFlow` from `Poll` to `Wait`.
 - **Breaking:** remove `DeviceEvent::Text`.
 - On Android, fix `DeviceId` to contain device id's.
+- Added initial blur support, currently limited to KDE Wayland
 
 # 0.29.1-beta
 
--- a/src/platform_impl/linux/wayland/state.rs
+++ b/src/platform_impl/linux/wayland/state.rs
@@ -31,6 +31,7 @@ use super::seat::{
     PointerConstraintsState, RelativePointerState, TextInputState, WinitPointerData,
     WinitPointerDataExt, WinitSeatState,
 };
+use super::types::kwin_blur::KWinBlurManager;
 use super::types::wp_fractional_scaling::FractionalScalingManager;
 use super::types::wp_viewporter::ViewporterState;
 use super::types::xdg_activation::XdgActivationState;
@@ -103,6 +104,9 @@ pub struct WinitState {
     /// Fractional scaling manager.
     pub fractional_scaling_manager: Option<FractionalScalingManager>,
 
+    /// KWin blur manager.
+    pub kwin_blur_manager: Option<KWinBlurManager>,
+
     /// Loop handle to re-register event sources, such as keyboard repeat.
     pub loop_handle: LoopHandle<'static, Self>,
 
@@ -161,6 +165,7 @@ impl WinitState {
             window_events_sink: Default::default(),
             viewporter_state,
             fractional_scaling_manager,
+            kwin_blur_manager: KWinBlurManager::new(globals, queue_handle).ok(),
 
             seats,
             text_input_state: TextInputState::new(globals, queue_handle).ok(),
--- a/src/platform_impl/linux/wayland/types/kwin_blur.rs
+++ b/src/platform_impl/linux/wayland/types/kwin_blur.rs
@@ -1,0 +1,70 @@
+//! Handling of KDE-compatible blur.
+
+use sctk::reexports::client::globals::{BindError, GlobalList};
+use sctk::reexports::client::protocol::wl_surface::WlSurface;
+use sctk::reexports::client::Dispatch;
+use sctk::reexports::client::{delegate_dispatch, Connection, Proxy, QueueHandle};
+use wayland_protocols_plasma::blur::client::{
+    org_kde_kwin_blur::OrgKdeKwinBlur, org_kde_kwin_blur_manager::OrgKdeKwinBlurManager,
+};
+
+use sctk::globals::GlobalData;
+
+use crate::platform_impl::wayland::state::WinitState;
+
+/// KWin blur manager.
+#[derive(Debug, Clone)]
+pub struct KWinBlurManager {
+    manager: OrgKdeKwinBlurManager,
+}
+
+impl KWinBlurManager {
+    pub fn new(
+        globals: &GlobalList,
+        queue_handle: &QueueHandle<WinitState>,
+    ) -> Result<Self, BindError> {
+        let manager = globals.bind(queue_handle, 1..=1, GlobalData)?;
+        Ok(Self { manager })
+    }
+
+    pub fn blur(
+        &self,
+        surface: &WlSurface,
+        queue_handle: &QueueHandle<WinitState>,
+    ) -> OrgKdeKwinBlur {
+        self.manager.create(surface, queue_handle, ())
+    }
+
+    pub fn unset(&self, surface: &WlSurface) {
+        self.manager.unset(surface)
+    }
+}
+
+impl Dispatch<OrgKdeKwinBlurManager, GlobalData, WinitState> for KWinBlurManager {
+    fn event(
+        _: &mut WinitState,
+        _: &OrgKdeKwinBlurManager,
+        _: <OrgKdeKwinBlurManager as Proxy>::Event,
+        _: &GlobalData,
+        _: &Connection,
+        _: &QueueHandle<WinitState>,
+    ) {
+        unreachable!("no events defined for org_kde_kwin_blur_manager");
+    }
+}
+
+impl Dispatch<OrgKdeKwinBlur, (), WinitState> for KWinBlurManager {
+    fn event(
+        _: &mut WinitState,
+        _: &OrgKdeKwinBlur,
+        _: <OrgKdeKwinBlur as Proxy>::Event,
+        _: &(),
+        _: &Connection,
+        _: &QueueHandle<WinitState>,
+    ) {
+        unreachable!("no events defined for org_kde_kwin_blur");
+    }
+}
+
+delegate_dispatch!(WinitState: [OrgKdeKwinBlurManager: GlobalData] => KWinBlurManager);
+delegate_dispatch!(WinitState: [OrgKdeKwinBlur: ()] => KWinBlurManager);
--- a/src/platform_impl/linux/wayland/window/mod.rs
+++ b/src/platform_impl/linux/wayland/window/mod.rs
@@ -131,6 +131,8 @@ impl Window {
         // Set transparency hint.
         window_state.set_transparent(attributes.transparent);
 
+        window_state.set_blur(attributes.blur);
+
         // Set the decorations hint.
         window_state.set_decorate(attributes.decorations);
 
@@ -407,6 +409,11 @@ impl Window {
     #[inline]
     pub fn scale_factor(&self) -> f64 {
         self.window_state.lock().unwrap().scale_factor()
+    }
+
+    #[inline]
+    pub fn set_blur(&self, blur: bool) {
+        self.window_state.lock().unwrap().set_blur(blur);
     }
 
     #[inline]
--- a/src/platform_impl/linux/wayland/window/state.rs
+++ b/src/platform_impl/linux/wayland/window/state.rs
@@ -4,7 +4,7 @@ use std::mem::ManuallyDrop;
 use std::num::NonZeroU32;
 use std::sync::{Arc, Weak};
 
-use log::warn;
+use log::{info, warn};
 
 use sctk::reexports::client::protocol::wl_seat::WlSeat;
 use sctk::reexports::client::protocol::wl_shm::WlShm;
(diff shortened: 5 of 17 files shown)


sh7dm:
Other platform codeowners can likely skip reviewing as there're only stub functions.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On Web, never return a `MonitorHandle`.
 - **Breaking:** Move `Event::RedrawRequested` to `WindowEvent::RedrawRequested`.
 - On macOS, fix crash in `window.set_minimized(false)`.
+- Added initial blur support, currently limited to KDE Wayland
```
kchibisov, line 17: You should just name a method on the Window which will be used to do that.

src/platform_impl/linux/wayland/state.rs
```diff
@@ -144,6 +148,11 @@ impl WinitState {
                 (None, None)
             };
 
+        let kwin_blur_manager = match KWinBlurManager::new(globals, queue_handle) {
+            Ok(kbm) => Some(kbm),
+            _ => None,
+        };
+
```
kchibisov, line 155: Inline it into the `Self` and just do `.ok()` instead of `match`.

src/platform_impl/linux/wayland/types/kwin_blur.rs
```diff
@@ -0,0 +1,68 @@
...
+impl Dispatch<OrgKdeKwinBlurManager, GlobalData, WinitState> for KWinBlurManager {
+    fn event(
+        _: &mut WinitState,
+        _: &OrgKdeKwinBlurManager,
+        _: <OrgKdeKwinBlurManager as Proxy>::Event,
+        _: &GlobalData,
+        _: &Connection,
+        _: &QueueHandle<WinitState>,
+    ) {
+    }
```
kchibisov, line 52: If they don't have events, could you put `unreachable!` saying so?

src/platform_impl/linux/wayland/types/kwin_blur.rs
```diff
@@ -0,0 +1,68 @@
...
+
+impl Dispatch<OrgKdeKwinBlur, (), WinitState> for KWinBlurManager {
+    fn event(
+        _: &mut WinitState,
+        _: &OrgKdeKwinBlur,
+        _: <OrgKdeKwinBlur as Proxy>::Event,
+        _: &(),
+        _: &Connection,
+        _: &QueueHandle<WinitState>,
+    ) {
```
kchibisov, line 63: ditto.

src/platform_impl/linux/wayland/window/state.rs
```diff
@@ -157,8 +161,14 @@ impl WindowState {
             .fractional_scaling_manager
             .as_ref()
             .map(|fsm| fsm.fractional_scaling(window.wl_surface(), queue_handle));
+        let blur = winit_state
+            .kwin_blur_manager
+            .as_ref()
+            .map(|kbm| kbm.blur(window.wl_surface(), queue_handle));
```
kchibisov, line 167: What's the reason it's done like that? Can't we create an object when we actually will set a `blur`?

src/platform_impl/linux/wayland/window/state.rs
```diff
@@ -840,6 +850,20 @@ impl WindowState {
...
+    pub fn set_blur(&mut self, blurred: bool) {
+        if blurred {
+            if let Some(blur) = &self.blur {
+                blur.commit();
+            }
+        } else {
+            if let Some(blur_manager) = &self.blur_manager {
+                blur_manager.unset(self.window.wl_surface());
+            }
+        }
```
kchibisov, line 864: We'd need a `debug`/`warn` when trying to set blur and it's not working.

src/platform_impl/linux/x11/window.rs
```diff
@@ -1017,6 +1017,10 @@ impl UnownedWindow {
     #[inline]
     pub fn set_transparent(&self, _transparent: bool) {}
 
+    // TODO
```
kchibisov, line 1020:
```suggestion
```

kchibisov: In general, such stuff is better tracked in the issues.

src/platform_impl/macos/window.rs
```diff
@@ -596,6 +596,9 @@ impl WinitWindow {
         self.setOpaque(!transparent)
     }
 
+    // TODO
```
kchibisov, line 599: ditto.

src/platform_impl/windows/window.rs
```diff
@@ -127,6 +127,9 @@ impl Window {
 
     pub fn set_transparent(&self, _transparent: bool) {}
 
+    // TODO
```
kchibisov, line 130: ditto.

src/platform_impl/linux/wayland/window/mod.rs
```diff
@@ -409,6 +411,11 @@ impl Window {
         self.window_state.lock().unwrap().scale_factor()
     }
 
+    #[inline]
+    pub fn set_blur(&self, blur: bool) {
+        self.window_state.lock().unwrap().set_blur(blur);
```
kchibisov, line 416: Should we try to sync it with opacity? For example, opaque window can't have blur behind it, so it might be better to not event try to set it, but just stash that blur is sort of enabled and then have a way to reload both at the same time?

src/platform_impl/linux/wayland/window/state.rs
```diff
@@ -840,6 +850,20 @@ impl WindowState {
         }
     }
 
+    /// Make window background blurred
+    #[inline]
+    pub fn set_blur(&mut self, blurred: bool) {
+        if blurred {
```
kchibisov, line 856: Is it really fine to do it like that? I think we need to update only when blur changes the state and have initial state as false? Or we could simply destroy the `blur` object?

src/window.rs
```diff
@@ -884,6 +895,12 @@ impl Window {
             .maybe_queue_on_main(move |w| w.set_transparent(transparent))
     }
 
+    /// Change the window blur state.
```
kchibisov, line 898: You should document the platform specifics as done in other methods.

src/window.rs
```diff
@@ -344,6 +346,15 @@ impl WindowBuilder {
         self
     }
 
+    /// Sets whether the background of the window should be blurred by the system.
+    ///
+    /// The default is `false`.
```
kchibisov, line 351: Ditto about platform specifics, but you can point to `Window::set_blur` from here saying that platform specific docs are there.

kchibisov (changes requested):
It's really hard to review the Wayland part, because the protocol is not even documented with regards to what it expects and whether some stuff requires a parent surface commit afterwards.

src/platform_impl/linux/wayland/window/mod.rs
sh7dm: We can, probably that will reduce GPU use in case compositor won't have that heuristic itself.

src/platform_impl/linux/wayland/state.rs
```diff
@@ -143,6 +147,7 @@ impl WinitState {
             } else {
                 (None, None)
             };
+        let kwin_blur_manager = KWinBlurManager::new(globals, queue_handle).ok();
```
kchibisov, line 150: You can inline that on like 169.

src/platform_impl/linux/wayland/types/kwin_blur.rs
```diff
@@ -0,0 +1,70 @@
...
+        unreachable!("no events defined for org_kde_kwin_blur_manager");
```
kchibisov, line 65: it's a different object, not `blur_manager`.

src/platform_impl/linux/wayland/window/state.rs
```diff
@@ -840,6 +848,35 @@ impl WindowState {
...
+                blur_manager.unset(self.window.wl_surface());
+            } else {
+                info!("Blur manager unavailable, unable to change blur")
+            }
+            if let Some(blur) = &self.blur {
+                blur.release();
+                self.blur = None;
+            }
+        }
+        self.is_blurred = blurred;
```
kchibisov, line 877: `self.is_blurred` is redundant, you can define that by `self.blur.is_some()`.

src/platform_impl/linux/wayland/window/state.rs
```diff
@@ -840,6 +848,35 @@ impl WindowState {
...
+                self.blur = self
+                    .blur_manager
+                    .as_ref()
+                    .map(|kbm| kbm.blur(self.window.wl_surface(), &self.queue_handle));
+            }
+            if let Some(blur) = &self.blur {
+                blur.commit();
+            } else {
+                info!("Blur manager unavailable, unable to change blur")
+            }
```
kchibisov, line 865: This looks like `unwrap_or_else`.

src/platform_impl/linux/wayland/window/state.rs
```diff
@@ -840,6 +848,35 @@ impl WindowState {
...
+        } else if !blurred && self.is_blurred {
+            if let Some(blur_manager) = &self.blur_manager {
+                blur_manager.unset(self.window.wl_surface());
+            } else {
+                info!("Blur manager unavailable, unable to change blur")
+            }
+            if let Some(blur) = &self.blur {
```
kchibisov, line 872: Could also move the `blue_manager` logic there and use `unwrap`, since it can't happen that you have `blur`, but not manager.

```suggestion
            if let Some(blur) = self.blur.take() {
```

src/window.rs
kchibisov: bump

CHANGELOG.md
kchibisov:
```suggestion
- Add `Window::set_blur` to request a blur behind the window; implemented on Wayland for now.
```

kchibisov (changes requested):

src/platform_impl/linux/wayland/types/kwin_blur.rs
sh7dm: Will correct the message

src/platform_impl/linux/wayland/window/state.rs
sh7dm: Not really, the else part doesn't give a KwinBlur instance

src/platform_impl/linux/wayland/window/mod.rs
sh7dm: Drawback: we'll have to also sync it with status in WindowBuilder. Might be not really worth it

sh7dm: @kchibisov please re-review

src/platform_impl/linux/wayland/window/state.rs
```diff
@@ -840,6 +847,28 @@ impl WindowState {
...
+    /// Make window background blurred
+    #[inline]
+    pub fn set_blur(&mut self, blurred: bool) {
+        if blurred && self.blur.is_none() {
+            self.blur = self
+                .blur_manager
+                .as_ref()
+                .map(|kbm| kbm.blur(self.window.wl_surface(), &self.queue_handle));
+            if let Some(blur) = &self.blur {
+                blur.commit();
```
kchibisov, line 859: Not setting the region makes it to blue the entire window, right?

kchibisov, line 859: You could do:

```rust
if let Some(blur_manager) = self.blur_manager.as_ref() {
   let blur = blur_manager.blur(self.window.wl_surface(), &self.queue_handle);
   blur.commit();
   self.blur = Some(blur);
} else {
   info!("Blur manager unavailable, unable to change blur")
}
```

so it's a bit more clear on what it's doing.

src/platform_impl/linux/wayland/window/state.rs
```diff
@@ -130,6 +132,8 @@ pub struct WindowState {
 
     viewport: Option<WpViewport>,
     fractional_scale: Option<WpFractionalScaleV1>,
+    blur: Option<OrgKdeKwinBlur>,
```
kchibisov, line 135: We should call a `release` inside the `Drop` impl if the object is present.

kchibisov (changes requested):
This and rebase and it should be good to go.

src/platform_impl/linux/wayland/window/state.rs
sh7dm: Does that for me, also should be like this per docs and other usage examples

src/platform_impl/linux/wayland/window/state.rs
```diff
@@ -901,6 +928,10 @@ impl Drop for WindowState {
         }
 
         surface.destroy();
+
+        if (self.blur.is_some()) {
+            self.blur.take().unwrap().release();
+        }
```
kchibisov, line 934: And do that before the `surface.destroy`.

```suggestion
        if let Some(blur) = self.blur.take() {
            blur.release();
        }
```

src/window.rs
kchibisov: It's actually not resolved, since you don't link.

src/window.rs
```diff
@@ -343,6 +345,16 @@ impl WindowBuilder {
         self
     }
 
+    /// Sets whether the background of the window should be blurred by the system.
+    /// Platform specifics described for Window::set_blur apply here
+    ///
+    /// The default is `false`.
```
kchibisov, line 351:
```suggestion
    /// Sets whether the background of the window should be blurred by the system.
    ///
    /// The default is `false`.
    ///
    /// See [`Window::set_blur`] for details.
```

kchibisov:
These 2 suggestions and it should be good to go.

sh7dm:
Please re-check as I wasn't able to test fully now (building Alacritty with these patches)

kchibisov:
Try this branch https://github.com/alacritty/alacritty/pull/7220 ? I don't use KDE so I just review based on how you use the protocol.

sh7dm:
Will do later, maybe tomorrow

sh7dm:
Working with [this test patch](https://tpaste.us/mgbv) on top of [update-winit-final](https://github.com/kchibisov/alacritty/tree/update-winit-final). Toggling blur works fine

kchibisov (approved):

kchibisov:
Thanks!

src/platform_impl/linux/wayland/window/state.rs
HigherOrderLogic: Hi, I'm trying to implement the `ext-background-effect` protocol in `winit`.

However, I notice that `winit` doesnt call the `set_region` method and just blur the whole window(?). `ext-background effect` has its corresponding method called `set_blur_region`. Do I have to call this or just follow the existing implementation and just dont?

kchibisov: We blur entire surface, yes, but not window, since decorations are excluded.

HigherOrderLogic: Is there any plan for per-region blur in the future?

Issue #3053, opened 2023-08-27
Make `EventLoopWindowTarget<T>` independent of the user event type `T`


madsmtm:
Requested in https://github.com/rust-windowing/winit/issues/2900#issuecomment-1694578554.

Should be possible from an API perspective, though might be a bit tricky to change in the backends.

Platforms that just use `PhantomData<T>` in the type (and thus don't need the actual user event type):
- [x] Wayland
- [x] X11
- [x] iOS
- [x] Android
- [x] macOS
- [x] Orbital: https://github.com/rust-windowing/winit/pull/3055
- [x] Web: https://github.com/rust-windowing/winit/pull/3081
- [x] Windows: https://github.com/rust-windowing/winit/pull/3061

Finish:
- [ ] Remove from top-level type: https://github.com/rust-windowing/winit/pull/3298

notgull:
This would be nice for https://github.com/rust-windowing/winit/pull/3014. The generic type makes it difficult to use the distributed function pointer strategy for event handling that I use in that PR.

nerditation:
initially, in [this comment](https://github.com/rust-windowing/winit/issues/2900#issuecomment-1694701585), I thought the Windows backend is doable but might take some efforts.

after I played a little bit with the Windows backend, it turns out way easier than I thought. I managed to get rid of the generic type T in `platform_imp::EventLoopWindowTarget<T>` and replace it with a `PhantomData` (just to keep the API). surprizingly, the change is relatively minor. here's a patch file I created:

[elwt-windows.patch](https://github.com/rust-windowing/winit/files/12451722/elwt-windows.patch)

there's two commits in the patch, because I was expecting the changes would be very large and I want to make check points after each small step. I did a quick `cargo test --all-features` on my machine and it seems nothing is broken (I hope). I also checked some of the examples, all seems fine.

madsmtm:
Nice! Could you put up a Pull Request? That makes it easier to review and discuss the changes

nerditation:
> Could you put up a Pull Request?

sure, #3061 is created. this is slight different from the patch I posted above, even fewer changes to the existing code, I also added more comments explaining why I did it this way. it's intentionally to be minimum as I'm not very familiar with the code base, but can serve as a start point.

daxpedda:
Unfortunately this isn't going to be easy on Web, but I have started working on it now.

kchibisov:
[...]

PR #3061, opened 2023-08-28
Windows: Make `EventLoopWindowTarget` independent of UserEvent type

nerditation:
the `EventLoopWindowTarget` is needed for window creation. conceptually, only `EventLoop` and `EventLoopProxy` need to be parameterized, and all other parts of the backend should be agnostic about the user event type, parallel to how `Event<T>` is parameterized, but `WindowEvent` is not.

this change removes the dependency on the type of user events from the `EventLoopWindowTarget` for the Windows backend, but keep a phantom data to keep the API intact. to achieve this, I moved the `Receiver` end of the mpsc channel from `ThreadMsgTargetData` into `EventLoop` itself, so the `UserEvent` is only passed between `EventLoop` and `EventLoopProxy`, all other part of the backend just use unit type as a placeholder for user events.

it's similar to the macos backend where an erased `EventHandler` trait object is used so all component except `EventLoop` and `EventLoopProxy` need to be parameterized. however `EventLoop` of the Windows backend already use an `Box<dyn FnMut>` to wrap the user provided event handler callback, so no need for an dedicated trait object, I just modified the wrapper to replace the placeholder user event with real value pulled from the channel. I find this is the approach which need minimum change to be made to existing code. but it does the job and could serve as a starting point to future Windows backend re-works.

### DIFF
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -207,7 +207,7 @@ impl<T: 'static> EventLoop<T> {
             become_dpi_aware();
         }
 
-        let thread_msg_target = create_event_target_window::<T>();
+        let thread_msg_target = create_event_target_window();
 
         let runner_shared = Rc::new(EventLoopRunner::new(thread_msg_target));
 
@@ -834,7 +834,7 @@ static THREAD_EVENT_TARGET_WINDOW_CLASS: Lazy<Vec<u16>> =
 /// <https://docs.microsoft.com/en-us/windows/win32/shell/taskbar#taskbar-creation-notification>
 pub static TASKBAR_CREATED: LazyMessageId = LazyMessageId::new("TaskbarCreated\0");
 
-fn create_event_target_window<T: 'static>() -> HWND {
+fn create_event_target_window() -> HWND {
     use windows_sys::Win32::UI::WindowsAndMessaging::CS_HREDRAW;
     use windows_sys::Win32::UI::WindowsAndMessaging::CS_VREDRAW;
     unsafe {
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -182,10 +182,7 @@ pub struct EventLoopWindowTarget<T: 'static> {
     thread_id: u32,
     thread_msg_target: HWND,
     pub(crate) runner_shared: EventLoopRunnerShared<UserEventPlaceholder>,
-    // TODO
-    // eventually should be removed after all the backends refactored,
-    // but for now should this be invariant or contra-variant to T?
-    _marker: PhantomData<*mut T>,
+    _marker: PhantomData<T>,
 }
 
 impl<T: 'static> EventLoop<T> {
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -98,17 +98,38 @@ use self::runner::RunnerState;
 
 use super::{window::set_skip_taskbar, SelectedCursor};
 
-pub(crate) struct WindowData<T: 'static> {
+/// some backends like macos uses an uninhabited `Never` type,
+/// on windows, `UserEvent`s are also dispatched through the
+/// WNDPROC callback, and due to the re-entrant nature of the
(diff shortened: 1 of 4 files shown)


src/platform_impl/windows/event_loop.rs
```diff
@@ -197,7 +218,11 @@ impl Default for PlatformSpecificEventLoopAttributes {
 pub struct EventLoopWindowTarget<T: 'static> {
     thread_id: u32,
     thread_msg_target: HWND,
-    pub(crate) runner_shared: EventLoopRunnerShared<T>,
+    pub(crate) runner_shared: EventLoopRunnerShared<UserEventPlaceholder>,
+    // TODO
+    // eventually should be removed after all the backends refactored,
+    // but for now should this be invariant or contra-variant to T?
+    _marker: PhantomData<*mut T>,
```
madsmtm, line 225: The correct variance and such is already ensured by the top-level `EventLoopWindowTarget`, just use `PhantomData<T>` here.

src/platform_impl/windows/event_loop.rs
```diff
@@ -135,17 +135,38 @@ static GET_POINTER_TOUCH_INFO: Lazy<Option<GetPointerTouchInfo>> =
...
+// `EventLoopRunnerShared<UserEventPlaceholder>` so we can get rid
+// of the generic parameter T in types which don't depend on T.
+// this is the approach which requires minimum changes to current
+// backend implementation. it should be considered transitional
+// and should be refactored and cleaned up eventually, I hope.
+
+pub(crate) struct WindowData {
     pub window_state: Arc<Mutex<WindowState>>,
-    pub event_loop_runner: EventLoopRunnerShared<T>,
+    pub event_loop_runner: EventLoopRunnerShared<UserEventPlaceholder>,
```
madsmtm, line 161: It seems like every use of `EventLoopRunner` is now parametrized by `UserEventPlaceholder`, maybe you can move the generic into `EventLoopRunner::event_handler`? But unsure about this, might be better to keep it minimal.

src/platform_impl/windows/event_loop.rs
```diff
@@ -135,17 +135,38 @@ static GET_POINTER_TOUCH_INFO: Lazy<Option<GetPointerTouchInfo>> =
...
+/// callback, recursively delivered events must be queued in a
+/// buffer, the current implementation put this queue in
+/// `EventLoopRunner`, which is shared between the event pumping
+/// loop and the callback. because it's hard to decide from the
+/// outside whether a event needs to be buffered, I decided not
+/// use `Event<Never>` for the shared runner state, but use unit
+/// as a placeholder so user events can be buffered as usual,
+/// the real `UserEvent` is pulled from the mpsc channel directly
+/// when the placeholder event is delivered to the event handler
+type UserEventPlaceholder = ();
```
madsmtm, line 150: I think it's preferable to use `struct UserEventPlaceholder;`, so that it's very obvious when we're dealing with a placeholder.

madsmtm (commented):
I'm not sure about the rest, indeed there are room for future improvement, but the idea looks fine to my eyes for now. Would like a Windows expert to have a look first though.

src/platform_impl/windows/event_loop.rs
nerditation: thanks for pointing out. will do.

src/platform_impl/windows/event_loop.rs
nerditation: in fact that's the first step in my initial proof of concept attempt, to remove the generic parameter T on `EventLoopRunner`. then not only all the uses of `EventLoopRunner` need to change, so does the implementation in file `platform_impl/windows/event_loop/runner.rs`, and a lot more diff lines were produced.

I decided to make code change at a minimal just to be safe.

src/platform_impl/windows/event_loop.rs
nerditation: good call.

madsmtm:
@msiglreith @maroider last call for getting a review here, otherwise I'll merge this soon, I'd like to progress on https://github.com/rust-windowing/winit/issues/3053.

PR #3066, opened 2023-08-30
Add direct `builder` methods

madsmtm:
Add methods `EventLoop::builder` and `Window::builder`, which allows creating the related builders without having to import them. See the examples for how things look cleaner.

Unsure about whether we should deprecate the old methods or not?


kchibisov:
What's the motivation behind those changes though, besides import? I'm not sure I really get the motivation behind all of that.

madsmtm:
Only motivation is to reduce the number of imports, and to hopefully make it easier to see at a glance which type you're constructing.

kchibisov:
But it's not like you can't request a builder on your own and it's not like we don't have `default`, so users will still be able to import/construct it?

madsmtm:
Definitely true, and there's no other point to deprecating the old methods other than keeping a slightly cleaner API

daxpedda (approved):
Thanks @madsmtm, the stuff you've done lately improves the API a lot!

kchibisov (approved):

kchibisov:
Needs rebase.

PR #3067, opened 2023-08-30
Refactor errors to be more specific

madsmtm:
Move errors out of the generic `error` module, and into the module where they're used.

This should help the user with actually handling the errors (instead of just propagating them), and with seeing what kind of errors a method can return (more work is still needed in this area, but this is a first step towards that).

The public changes are:
- `error::ExternalError::Ignored` -> `event::RequestIgnored`.
- `error::EventLoopError` is split into:
  - `event_loop::EventLoopCreationError` (with less variants).
  - `event_loop::EventLoopRunError` (with less variants).
- `platform::startup_notify::EventLoopExtStartupNotify::request_activation_token` now errors with `platform::startup_notify::ActivationTokenNotFound` instead of `error::NotSupportedError`.
- `error::ExternalError` -> `window::WindowError`.
- `error::NotSupportedError` -> `window::NotSupportedError`.

A few documents that guide decision making here:
- [Error Handling in Rust](https://nrc.github.io/error-docs/intro.html)
- [Designing error types in Rust](https://mmapped.blog/posts/12-rust-error-handling)
- [`std::error`](https://doc.rust-lang.org/nightly/std/error/index.html)

Checklist:
  - Unsure how much information is relevant here.

Builds upon #3066, DO NOT MERGE YET!

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -26,27 +26,6 @@ pub struct OsError {
     error: platform_impl::OsError,
 }
 
-/// A general error that may occur while running the Winit event loop
-#[derive(Debug)]
-pub enum EventLoopError {
-    /// The operation is not supported by the backend.
-    NotSupported(NotSupportedError),
-    /// The OS cannot perform the operation.
-    Os(OsError),
-    /// The event loop can't be re-run while it's already running
-    AlreadyRunning,
-    /// The event loop can't be re-created.
-    RecreationAttempt,
-    /// Application has exit with an error status.
-    ExitFailure(i32),
-}
-
-impl From<OsError> for EventLoopError {
-    fn from(value: OsError) -> Self {
-        Self::Os(value)
-    }
-}
-
 impl NotSupportedError {
     #[inline]
     #[allow(dead_code)]
@@ -99,22 +78,9 @@ impl fmt::Display for NotSupportedError {
     }
 }
 
-impl fmt::Display for EventLoopError {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
-        match self {
-            EventLoopError::AlreadyRunning => write!(f, "EventLoop is already running"),
-            EventLoopError::RecreationAttempt => write!(f, "EventLoop can't be recreated"),
-            EventLoopError::NotSupported(e) => e.fmt(f),
-            EventLoopError::Os(e) => e.fmt(f),
-            EventLoopError::ExitFailure(status) => write!(f, "Exit Failure: {status}"),
-        }
-    }
-}
-
 impl error::Error for OsError {}
 impl error::Error for ExternalError {}
 impl error::Error for NotSupportedError {}
-impl error::Error for EventLoopError {}
 
 #[cfg(test)]
 mod tests {
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,3 +1,4 @@
+use std::panic::Location;
 use std::{error, fmt};
 
 use crate::platform_impl;
@@ -5,31 +6,30 @@ use crate::platform_impl;
 /// The error type for when the OS cannot perform the requested operation.
 #[derive(Debug)]
 pub struct OsError {
-    line: u32,
-    file: &'static str,
+    location: &'static Location<'static>,
     error: platform_impl::OsError,
 }
 
 impl OsError {
     #[allow(dead_code)]
-    pub(crate) fn new(line: u32, file: &'static str, error: platform_impl::OsError) -> OsError {
-        OsError { line, file, error }
+    #[track_caller] // Allows `Location::caller` to work properly
+    pub(crate) fn new(error: platform_impl::OsError) -> OsError {
+        OsError {
+            location: Location::caller(),
+            error,
+        }
     }
-}
-
-#[allow(unused_macros)]
-macro_rules! os_error {
-    ($error:expr) => {{
-        crate::error::OsError::new(line!(), file!(), $error)
-    }};
 }
 
 impl fmt::Display for OsError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
-        f.pad(&format!(
+        write!(
+            f,
             "os error at {}:{}: {}",
-            self.file, self.line, self.error
-        ))
+            self.location.file(),
+            self.location.line(),
+            self.error
+        )
     }
 }
 
--- a/src/error.rs
+++ b/src/error.rs
@@ -8,8 +8,6 @@ use crate::platform_impl;
 pub enum ExternalError {
     /// The operation is not supported by the backend.
     NotSupported(NotSupportedError),
-    /// The operation was ignored.
-    Ignored,
     /// The OS cannot perform the operation.
     Os(OsError),
 }
@@ -84,7 +82,6 @@ impl fmt::Display for ExternalError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
         match self {
             ExternalError::NotSupported(e) => e.fmt(f),
-            ExternalError::Ignored => write!(f, "Operation was ignored"),
             ExternalError::Os(e) => e.fmt(f),
         }
     }
--- a/src/error.rs
+++ b/src/error.rs
@@ -2,36 +2,12 @@ use std::{error, fmt};
 
 use crate::platform_impl;
 
-// TODO: Rename
-/// An error that may be generated when requesting Winit state
-#[derive(Debug)]
-pub enum ExternalError {
-    /// The operation is not supported by the backend.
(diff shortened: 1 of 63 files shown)


src/event_loop.rs
```diff
@@ -426,6 +427,65 @@ impl<T: 'static> fmt::Debug for EventLoopProxy<T> {
...
+    RecreationAttempt,
+    /// The OS cannot perform the operation.
+    Os(OsError),
+}
+
+impl fmt::Display for EventLoopCreationError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+        match self {
+            Self::RecreationAttempt => write!(f, "tried to recreate EventLoop"),
+            Self::Os(e) => write!(f, "{e}"),
```
notgull, line 444: Generally, it compiles faster if you instead write:

```rust
fmt::Display::fmt(e, f)
```

Goes for the rest of these kinds of errors as well.

src/platform/startup_notify.rs
```diff
@@ -33,6 +34,28 @@ const X11_VAR: &str = "DESKTOP_STARTUP_ID";
...
 
+/// The error type for when the requested operation is not supported by the backend.
+#[derive(Debug)]
+pub struct ActivationTokenNotFound {
+    _marker: (),
+}
+
+impl fmt::Display for ActivationTokenNotFound {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+        write!(f, "the requested operation is not supported by Winit")
```
notgull, line 45: Use `f.write_str` here for the same reason

src/event_loop.rs
madsmtm: Compilation speed in the minor like this is not really a concern IMO, and I prefer the consistency (e.g. you write it very similar as for `Self::RecreationAttempt` and `Self::Os`).

But if you insist, then sure, I don't really care either way.

src/event.rs
```diff
@@ -1068,6 +1068,23 @@ pub enum MouseScrollDelta {
...
 }
 
+/// The request to change the inner size synchronously was ignored.
+///
+/// See [`InnerSizeWriter::request_inner_size`] for details.
+#[derive(Debug)] // Explicitly not other traits, in case we want to extend it in the future
+#[non_exhaustive]
+pub struct RequestIgnored {
+    _priv: (),
+}
```
daxpedda, line 1078: Why do we need `non_exhaustive` if it has a private field?

src/event.rs
```diff
@@ -1081,16 +1098,24 @@ impl InnerSizeWriter {
...
 
-    /// Try to request inner size which will be set synchroniously on the window.
+    /// Try to request a new inner size which will be set synchronously on the
+    /// window.
+    ///
+    ///
+    /// # Errors
+    ///
+    /// This method returns an error when the request to change the inner size
+    /// of the window was ignored - likely because you tried to change the
```
daxpedda, line 1108: I don't think "likely" is necessary, I believe this is the only way this can fail.

daxpedda (approved):
Absolutely love this change, it's a great improvement imo!

Only reviewed the cross-platform stuff and Web, I think Web errors in general need an overhaul so I'm gonna leave things as they are.

src/event.rs
madsmtm: Oh yeah, we don't. I'll remove that

src/event_loop.rs
madsmtm: I rewrote this one to give a more specific error message instead

madsmtm:
> It's also a good idea to extract the "could not find any backends" error in `linux/mod.rs` into its own variant.

Fully agreed, ideally I'd get rid of the generic `OsError`, but we have to start somewhere.

src/platform/startup_notify.rs
madsmtm: This is being tracked upstream in https://github.com/rust-lang/rust/issues/99012, and I believe that's the better way to go (e.g. write the "best" Rust code you can, and let the compiler figure out the details (within limits of course)).

Besides, the compilation performance impact here is negligible compared to so many other things in this crate.

src/error.rs
```diff
@@ -1,142 +1,36 @@
+use std::panic::Location;
```
kchibisov, line 1: What's the point of keeping this module then in the end? The `OsError` sounds like it should be in the platform module, since it depends on the platform. We could maybe name it differently, like `PlatformError`.

src/event.rs
```diff
@@ -1068,6 +1068,22 @@ pub enum MouseScrollDelta {
...
+    _priv: (),
+}
+
+impl fmt::Display for RequestIgnored {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+        f.write_str("the request was ignored")
+    }
+}
+
+impl std::error::Error for RequestIgnored {}
```
kchibisov, line 1085: I think nothing stops from doing a proper import at the top.

src/event.rs
```diff
@@ -1068,6 +1068,22 @@ pub enum MouseScrollDelta {
...
+#[derive(Debug)] // Explicitly not other traits, in case we want to extend it in the future
+pub struct RequestIgnored {
+    _priv: (),
+}
+
+impl fmt::Display for RequestIgnored {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+        f.write_str("the request was ignored")
+    }
+}
```
kchibisov, line 1083: What's the reason this error is here and not on the `window` for example?

src/event_loop.rs
```diff
@@ -426,6 +427,65 @@ impl<T: 'static> fmt::Debug for EventLoopProxy<T> {
...
+/// A general error that may occur while running the Winit event loop
+#[derive(Debug)]
+#[non_exhaustive]
+pub enum EventLoopRunError {
+    /// The event loop can't be re-run while it's already running
+    AlreadyRunning,
+    /// The OS cannot perform the operation.
+    Os(OsError),
+    /// Application has exit with an error status.
+    ExitFailure(i32),
```
kchibisov, line 467: This is likely gone with the recent rework.

src/platform/startup_notify.rs
```diff
@@ -33,6 +34,28 @@ const X11_VAR: &str = "DESKTOP_STARTUP_ID";
 /// The variable which is used mostly on Wayland.
 const WAYLAND_VAR: &str = "XDG_ACTIVATION_TOKEN";
 
+/// The error type for when the requested operation is not supported by the backend.
+#[derive(Debug)]
+pub struct ActivationTokenNotFound {
+    _marker: (),
+}
+
+impl fmt::Display for ActivationTokenNotFound {
```
kchibisov, line 43: Why it was changed to something that has nothing to do with the activation token. It can't be not found. It's either supported or not supported, which is why it was `NotSupportedError`.

src/window.rs
```diff
@@ -1511,6 +1511,49 @@ unsafe impl HasRawDisplayHandle for Window {
...
+    /// The operation is not supported by the backend.
+    NotSupported(NotSupportedError),
+    /// The OS cannot perform the operation.
+    Os(OsError),
+}
+
+impl fmt::Display for WindowError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+        match self {
+            Self::NotSupported(e) => write!(f, "window operation failed: {e}"),
```
kchibisov, line 1527: Could just name it a `RequestError`. I'm also not a fan of `TODO` items in top level stuff, could just open an issue.

kchibisov (changes requested):

src/platform_impl/android/mod.rs
```diff
@@ -19,17 +19,21 @@ use raw_window_handle::{
...
     event::{self, InnerSizeWriter, StartCause},
-    event_loop::{self, ControlFlow, EventLoopWindowTarget as RootELW},
+    event_loop::{
+        self, ControlFlow, EventLoopCreationError, EventLoopRunError,
+        EventLoopWindowTarget as RootELW,
+    },
     platform::pump_events::PumpStatus,
     window::{
-        self, CursorGrabMode, ImePurpose, ResizeDirection, Theme, WindowButtons, WindowLevel,
+        self, CursorGrabMode, ImePurpose, NotSupportedError, ResizeDirection, Theme, WindowButtons,
```
MarijnS95, line 33: Is it just me who thinks these `self` imports are funky?

MarijnS95 (approved):
Ack'ing the Android changes, good stuff!

kchibisov:
ping @madsmtm 

what should we do with this now?

madsmtm:
> what should we do with this now?

I still believe this is a change that should be done, but it'll need work, so I'll close this PR, and try to open new, smaller scoped ones for these error changes.

PR #3070, opened 2023-08-30
Windows: Add `#[deny(unsafe_op_in_unsafe_fn)]`

madsmtm:
Although this does increase the verbosity, it also greatly helps in auditing which parts of the codebase is actually `unsafe`, and thus need extra scrutiny. My plan is to add it to the entirety of `winit`.

[The lint](https://doc.rust-lang.org/stable/nightly-rustc/rustc_lint/builtin/static.UNSAFE_OP_IN_UNSAFE_FN.html) is even stated in [the documentation for the keyword `unsafe`](https://doc.rust-lang.org/std/keyword.unsafe.html), and the desire is to make it warn by default at some point in the far future.

Related: What I'm seeing from applying this lint is that a _lot_ of the Windows backend is `unsafe`, I think we should consider working on improving that.

### DIFF
--- a/src/platform_impl/windows/dpi.rs
+++ b/src/platform_impl/windows/dpi.rs
@@ -78,36 +78,36 @@ pub fn dpi_to_scale_factor(dpi: u32) -> f64 {
 }
 
 pub unsafe fn hwnd_dpi(hwnd: HWND) -> u32 {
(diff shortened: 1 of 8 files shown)


msiglreith (approved):

kchibisov:
@madsmtm will need a rebase.

Issue #2903, opened 2023-06-24
Encode event lifecycle in the type-system


madsmtm:
We have a lot of invariants in how our events are delivered, but these are only vaguely documented. Examples of things that commonly go wrong:
- Initializing windows outside `NewEvents(Init)`/`Resumed`
- Not suspending / resetting the draw buffers on `Suspended`

To remedy this situation, I propose we change the `run` method(s) to use a trait object instead of just a closure.

Initial draft for what typical usage will look like (this will need a lot of tweaking, especially with regards to `ControlFlow`, and possibly we should consider a `SimpleApplication` trait that handles some of this for you):

```rust
struct MyApp {
    window: Window,
}

struct SuspendedApp {
    // ...
}

impl Application for MyApp {
    type SuspendedState = SuspendedApp;

    fn init(event_loop: &EventLoopWindowTarget) -> Self {
        let window = WindowBuilder::new().build(event_loop);
        Self { window }
    }

    fn suspend(self) -> Self::SuspendedState {
        todo!()
    }

    fn resume(state: Self::SuspendedState, event_loop: &EventLoopWindowTarget) -> Self {
        todo!()
    }

    fn on_event(
        &mut self,
        event: Event<'_>,
        event_loop: &EvenLoopWindowTarget,
        control_flow: &mut ControlFlow,
    ) {
         // The usual event handling in here
    }
}

impl Drop for MyApp { ... } // LoopDestroyed

fn main() {
    let event_loop = EventLoop::new();
    event_loop.run::<MyApp>();
}
```

The closure we have now is really nice though, we loose the easiness offered by that, but this is strictly more correct, which I think is important!

Again, this will need a lot of tweaking, opening as an issue first to gather opinions.

Relates to https://github.com/rust-windowing/winit/issues/2900, and many, many more.

notgull:
I generally like this idea, especially since the `Application` trait could probably be implemented on top of a closure to provide compatibility.

kchibisov:
@madsmtm so you do you want more or less the API Wayland backend has internally, you can look at state and e.g pointer handlers to get a grasp on how it works? Since what you suggest is exactly how the API behaves on Wayland, it could simplify stuff for the wayland backend for sure, for example, but it could be hard for other backends like Windows and X11 where they have just one closure to handle all of that.

kchibisov:
cc @rib since it sort of interacts with the way we do event loop APIs and I'd like to hear what you think about all of that.

[...]

PR #3073, opened 2023-08-30
Add `ApplicationHandler` trait

madsmtm:
Part of https://github.com/rust-windowing/winit/issues/2903.

Builds upon https://github.com/rust-windowing/winit/pull/3056, so see the last commit(s) for the changes in this PR.

Add an `ApplicationHandler<T>` trait that replaces the `Event<T>` enum (should probably be kept around for at least one release, but be marked deprecated) by encoding these events as callbacks on a user-specified mutable application object instead.
This allows us to fix many outstanding issues with the current event loop design, including ensuring that windows are only created after initialization, and helping users to properly suspended/resume their application (also here: recreating surfaces and such).

TODO:
- [ ] Figure out what to do about timing.
- [x] Experiment with returning a closure on suspend, instead of having to manually specify the suspended data.
  - https://github.com/rust-windowing/winit/pull/3074
- [ ] Should we do application initialization differently?
- [ ] How exactly should pump_events work?
- [ ] The user currently has to type a lot compared to the old closure, can we make things even nicer?
- [ ] How do we do platform-specific extensions?
- [ ] Naming things.
- [ ] Test on all platforms that the events are actually delivered in the order we want.
- [ ] Changelog entry and update examples.
- [ ] ...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On Web, enable event propagation and let `DeviceEvent`s appear after `WindowEvent`s.
 - On Web, take all transient activations on the canvas and window into account to queue a fullscreen request.
 - On Web, remove any fullscreen requests from the queue when an external fullscreen activation was detected.
+- **Breaking:** Removed `ControlFlow`, can now be set with `set_poll()`, `set_wait()`, .. on `EventLoopWindowTarget`.
 
 # 0.29.1-beta
 
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ another library.
 ```rust
 use winit::{
     event::{Event, WindowEvent},
-    event_loop::{ControlFlow, EventLoop},
+    event_loop::EventLoop,
     window::WindowBuilder,
 };
 
@@ -43,14 +43,14 @@ fn main() {
     let event_loop = EventLoop::new();
     let window = WindowBuilder::new().build(&event_loop).unwrap();
 
-    event_loop.run(move |event, _, control_flow| {
-        *control_flow = ControlFlow::Wait;
+    event_loop.run(move |event, elwt| {
+        elwt.set_wait();
 
         match event {
             Event::WindowEvent {
                 event: WindowEvent::CloseRequested,
                 window_id,
-            } if window_id == window.id() => *control_flow = ControlFlow::Exit,
+            } if window_id == window.id() => elwt.exit(),
             _ => (),
         }
     });
--- a/examples/child_window.rs
+++ b/examples/child_window.rs
@@ -9,7 +9,7 @@ fn main() -> Result<(), impl std::error::Error> {
     use winit::{
         dpi::{LogicalPosition, LogicalSize, Position},
         event::{ElementState, Event, KeyEvent, WindowEvent},
-        event_loop::{ControlFlow, EventLoop, EventLoopWindowTarget},
+        event_loop::{EventLoop, EventLoopWindowTarget},
         window::raw_window_handle::HasRawWindowHandle,
         window::{Window, WindowBuilder, WindowId},
     };
@@ -46,14 +46,14 @@ fn main() -> Result<(), impl std::error::Error> {
 
     println!("parent window: {parent_window:?})");
 
-    event_loop.run(move |event: Event<()>, event_loop, control_flow| {
-        *control_flow = ControlFlow::Wait;
+    event_loop.run(move |event: Event<()>, elwt| {
+        elwt.set_wait();
 
         if let Event::WindowEvent { event, window_id } = event {
             match event {
                 WindowEvent::CloseRequested => {
                     windows.clear();
-                    *control_flow = ControlFlow::Exit;
+                    elwt.exit();
                 }
                 WindowEvent::CursorEntered { device_id: _ } => {
                     // On x11, println when the cursor entered in a window even if the child window is created
@@ -70,7 +70,7 @@ fn main() -> Result<(), impl std::error::Error> {
                         },
                     ..
                 } => {
-                    spawn_child_window(&parent_window, event_loop, &mut windows);
+                    spawn_child_window(&parent_window, elwt, &mut windows);
                 }
                 WindowEvent::RedrawRequested => {
                     if let Some(window) = windows.get(&window_id) {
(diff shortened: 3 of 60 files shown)


daxpedda:
Sorry about the delay in #3056, will try to finish it up today.

notgull:
Now that I see it in real life, I'm not that sure about this one. It's still possible for users to just keep the `Window` around regardless of what the type system says. So it's a handful of extra complexity that may not provide strong enough guarantees.

madsmtm:
For my part, this PR is mostly about ensuring that the application is initialized before you try to create windows, which solves sooo many issues on macOS and iOS, and unblocks a lot of my other work (it's really hard to make sure that e.g. fullscreen-ing works correctly when you basically have to work with the window in two states). This would also be possible with other solutions, e.g. nested closures in `EventLoop::run`, but that would still be ugly.

> It's still possible for users to just keep the `Window` around

True, but there are ways to fix that; one approach could be, drum roll, you guessed it, lifetimes!

Queue the most general example I could conjure: 
```rust
mod winit {
    pub struct Window<'running> { ... }

    impl<'running> Window<'running> {
        pub fn new(elwt: &'running EventLoopWindowTarget) -> Self { ... }
    }

    pub trait ApplicationHandler<'suspended: 'running, 'running, T = ()> {
        type Suspended: 'suspended;

        // Note that binding the elwt here to `'running` is non-trivial, because then we
        // have to allow it to escape. But should be doable, in one form or another.
        fn resume(suspended: Self::Suspended, elwt: &'running EventLoopWindowTarget) -> Self;

        fn suspend(self) -> Self::Suspended;
    }
}

struct State<'suspended> { ... }

struct App<'suspended, 'running> {
    elwt: &'running EventLoopWindowTarget,
    window: Window<'running>,
    state: State<'suspended>,
}

impl<'suspended: 'running, 'running> ApplicationHandler<'suspended, 'running> for App<'suspended, 'running> {
    type Suspended = State<'suspended>;

    fn resume(
        state: Self::Suspended,
        elwt: &'running EventLoopWindowTarget,
    ) -> Self {
        Self {
            elwt,
            window: Window::new(elwt),
            state,
        }
    }

    fn suspend(self) -> Self::Suspended {
        self.state
    }
}
```

But whether or not we decide to do that, I still think moving to a trait like this is the way forwards.

(Again, the above is still _possible_ using closures, but it becomes a nested mess. One closure is nice, three nested closures is quite enough thank you very much).

dhardy:
The trait in the PR looks usable, but I think I'd just set `type Suspended = Self` and continue using [this](https://github.com/kas-gui/kas/blob/master/crates/kas-core/src/shell/window.rs#L29):
```rust
struct WindowData<S: WindowSurface> {
    window: winit::window::Window,
    #[cfg(all(wayland_platform, feature = "clipboard"))]
    wayland_clipboard: Option<smithay_clipboard::Clipboard>,
    surface: S,
    // ...
}

/// Per-window data
pub struct Window<A: AppData, S: WindowSurface> {
    widget: kas::Window<A>,
    ev_state: EventState,
    // ...
    window: Option<WindowData<S>>,
}
```
For any app with multiple windows this strategy probably makes more sense than using a separate `Suspended` type.

Regarding the lifetimes idea, I don't think it enforces anything (unless you are deliberately *not* passing `elwt` into other methods, thus forcing users to store a copy in their app state). Regardless, it looks more painful than useful, so please don't.

madsmtm:
> The trait in the PR looks usable, but I think I'd just set `type Suspended = Self` and continue using [this](https://github.com/kas-gui/kas/blob/master/crates/kas-core/src/shell/window.rs#L29):
> 
> ```rust
> struct WindowData<S: WindowSurface> {
>     window: winit::window::Window,
>     #[cfg(all(wayland_platform, feature = "clipboard"))]
>     wayland_clipboard: Option<smithay_clipboard::Clipboard>,
>     surface: S,
>     // ...
> }
> 
> /// Per-window data
> pub struct Window<A: AppData, S: WindowSurface> {
>     widget: kas::Window<A>,
>     ev_state: EventState,
>     // ...
>     window: Option<WindowData<S>>,
> }
> ```
> 
> For any app with multiple windows this strategy probably makes more sense than using a separate `Suspended` type.

The idea was that `winit` would allow you to know statically that a window has graphics attached; e.g. for your case, it'd look something like this (and you'd avoid the `Option` around the window data):
```rust
struct WindowData<S: WindowSurface{
    window: winit::window::Window,
    #[cfg(all(wayland_platform, feature = "clipboard"))]
    wayland_clipboard: Option<smithay_clipboard::Clipboard>,
    surface: S,
    // ...
}

pub struct Window<A: AppData, S: WindowSurface{
    widget: kas::Window<A>,
    ev_state: EventState,
    // ...
}

struct App {
    windows: Vec<(Window<A, S>, WindowData<S>)>,
}

struct Suspended {
    windows: Vec<Window<A, S>>,
}
```

> Regarding the lifetimes idea, I don't think it enforces anything (unless you are deliberately _not_ passing `elwt` into other methods, thus forcing users to store a copy in their app state). Regardless, it looks more painful than useful, so please don't.

No I definitely agree, we'd want something that was actually understandable and usable, the idea I sent was just to prove the concept (if you replace the `Suspended` type with `App<'suspended, 'running>;`, and fix the `suspend/resume` methods, it _does_ fail to compile, because the `Suspended` type must be longer than `'suspended`, which in turn must be longer than `'running`).
I've edited the comment to make that slightly more clear.

---

But again, I'm definitely on the pragmatic side here, fixing the `Suspend`/`Resume` stuff is not that important to me atm, but there are things this trait would be more useful for moving forwards. So I'm fine with ripping it out, at least in the first iteration.

dhardy:
> (if you replace the Suspended type with App<'suspended, 'running>;, and fix the suspend/resume methods, it does fail to compile, because the Suspended type must be longer than 'suspended, which in turn must be longer than 'running).

This works with your `App`, but it's still possible to get around the restriction by using only a single lifetime: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3d8728a864a3f9f119d77388592d2473

madsmtm:
> it's still possible to get around the restriction by using only a single lifetime

Oh yeah... I suspect there's probably still a way to do it, but let's handle that when we get there.

madsmtm:
Replaced by https://github.com/rust-windowing/winit/pull/3517

PR #3074, opened 2023-08-31
Alternative suspension mechanism for `ApplicationHandler` using closures

madsmtm:
Alternative way of doing the suspension in https://github.com/rust-windowing/winit/pull/3073.


madsmtm:
Requires `#![impl_trait_in_assoc_type]`, so in any case not viable yet, just opened this PR to show how it'd look

PR #3075, opened 2023-08-31
Change how our `raw-window-handle` dependency is exposed

madsmtm:
`raw-window-handle` has yet to reach v1.0, and I doubt it will do so in the near future. This causes many problems to downstream users, as they must upgrade their `winit` version in lockstep with their `wgpu` (or similar) version.
In the past, we've been able to do the semver trick in `raw-window-handle`, but with the current API that's no longer possible.

So, until `raw-window-handle` becomes stable, I propose we make dependency on any version of `raw-window-handle` optional, and behind it's own feature flag, to make it possible for the user to select the version that they want.

This allows us to provide a much better support strategy, where instead of immediately removing support for older versions of `raw-window-handle` and forcing users to upgrade their rendering library, we instead slowly deprecate, perhaps over several releases if deemed necessary.

TODO:
- [x] Finish the implementation.
- [ ] Add an entry to `CHANGELOG.md`
- [ ] Figure out how to do examples now

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,9 +55,31 @@ cursor-icon = "1.0.0"
 log = "0.4"
 mint = { version = "0.5.6", optional = true }
 once_cell = "1.12"
-raw_window_handle = { package = "raw-window-handle", version = "0.5", features = ["std"] }
 serde = { version = "1", optional = true, features = ["serde_derive"] }
 smol_str = "0.2.0"
+
+# Enable support for different versions of `raw-window-handle`.
+#
+# Old ones will be deprecated (and removed in major versions) slowly as
+# `winit` moves towards newer releases of `raw-window-handle`.
+#
+# Note that this whole thing is only necessary because `raw-window-handle` is
+# still unstable, once it reaches a stable v1.0, the need for these features
+# will disappear.
+
+# Enable version 0.4 support by using the "rwh-0-4" feature.
+[dependencies.rwh-0-4]
+package = "raw-window-handle"
+version = "0.4"
+features = ["alloc"]
+optional = true
+
+# Enable version 0.5 support by using the "rwh-0-5" feature.
+[dependencies.rwh-0-5]
+package = "raw-window-handle"
+version = "0.5"
+features = ["std"]
+optional = true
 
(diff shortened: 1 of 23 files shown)


notgull (approved):
I like the idea.

madsmtm:
> This would be great! I'm currently stuck on rwh 0.4 due to using an older version of `wgpu`.

Thanks for the input, an extra data point to consider that maybe we should consider re-adding v0.4 support, in a deprecated form, and at least for a little while longer.

May I ask why you're stuck on the older `wgpu`?

daxpedda (commented):
LGTM, let me know if you want me to do the Web implementation.

madsmtm:
> let me know if you want me to do the Web implementation.

No thanks, I got it.

I would like some input on how we should do our examples, now that they require the `rwh-0-5` feature to be enabled?

We could specify `required-features = ["rwh-0-5"]` for every example, but that means that users have to run with `cargo run --example=my_example --features=rwh-0-5`, which is a bit annoying. The other option is as I've hackily implemented, to stub out drawing to the window unless requested.

daxpedda:
> We could specify `required-features = ["rwh-0-5"]` for every example, but that means that users have to run with `cargo run --example=my_example --features=rwh-0-5`, which is a bit annoying.

```toml
[dev-dependencies]
winit = { path = "", default-features = false, features = ["rwh-0-5"] }
```
Is what I usually do. But that means you can't execute the examples without `rwh-0-5`, which might be fine?

kchibisov:
You could add recent raw window handle into the default feature set, which is what it should be, the previous versions should be enabled on demand.

madsmtm:
> You could add recent raw window handle into the default feature set, which is what it should be, the previous versions should be enabled on demand.

The problem is that the current version could easily change many times while the version of `winit` stays the same.

But I guess it'd be much more pragmatic.

kchibisov:
@madsmtm feature set change is a breaking change. The default feature should be the one of the latest non-breaking release. it's not like you just specify `raw-window-handle = "*"`.

kchibisov:
I think this is all implemented now?

Issue #3072, opened 2023-08-30
macos: create unfocused window


rparrett:
Using macos 13.4.1.
Testing with the main branch.

I'd like to create a single window that begins its life visible but unfocused, never stealing focus.

`with_active(false)` seems like it should accomplish this, but with that, the new window still appears on top of every other window, reacts to key presses, etc.

Even with `.with_window_level(winit::window::WindowLevel::AlwaysOnBottom)` and `.with_visible(false)`, the window seems to take focus away from the currently focused window, interrupting typing (but doesn't seem to react to keypresses itself).

### Repro:

```
sleep 5 && cargo run
# now before the window opens, switch back to a text editor
# and attempt to type some things without being interrupted by the new window stealing focus
```

```rust
use simple_logger::SimpleLogger;
use winit::{
    event::{Event, WindowEvent},
    event_loop::EventLoop,
    window::WindowBuilder,
};

fn main() {
    SimpleLogger::new().init().unwrap();

    let event_loop = EventLoop::new();

    let window = WindowBuilder::new()
        .with_active(false)
        .build(&event_loop)
        .unwrap();

    event_loop.run(move |event, _, control_flow| {
        control_flow.set_wait();

        match event {
            Event::WindowEvent {
                event: WindowEvent::CloseRequested,
                window_id,
            } if window_id == window.id() => control_flow.set_exit(),
            _ => (),
        }
    });
}
```

madsmtm:
Have you looked through [the platform-specific extensions in `winit::platform::macos`](https://docs.rs/winit/0.28.6/x86_64-apple-darwin/winit/platform/index.html)? I'm not sure exactly what you want, but it might be possible with a combination of the options in there (in particular see `EventLoopBuilderExtMacOS`, it allows you to set the activation policy and such).

rparrett:
Thanks, I hadn't seen that. Unfortunately even with
```rust
let event_loop = EventLoopBuilder::new()
    .with_activation_policy(ActivationPolicy::Accessory)
    .build();
```
The new window steals focus.

I realize that my report is a bit all-over-the-place. To clarify, the behavior I'm after is:

- have terminal open
- cargo run
- new window appears behind the terminal
- focus was never stolen from the terminal

And in every other way, I would like for it to be normal window. Any mention of the status bar is me not realizing this is a separate concept from focus in macos.

[...]

PR #3076, opened 2023-09-01
Display all platform-specific documentation on docs.rs

madsmtm:
This required a bit of refactoring regarding platform privacy, but all of it for the better.


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- **Breaking:** No longer export `platform::x11::XNotSupported`.
```
madsmtm, line 11: @notgull I couldn't see a reason why this was exported, so I removed it, but maybe you can shed some light on it?

src/lib.rs
```diff
@@ -139,7 +139,8 @@
 #![deny(clippy::all)]
 #![cfg_attr(feature = "cargo-clippy", deny(warnings))]
 // Doc feature labels can be tested locally by running RUSTDOCFLAGS="--cfg=docsrs" cargo +nightly doc
-#![cfg_attr(docsrs, feature(doc_auto_cfg))]
+#![cfg_attr(docsrs, feature(doc_auto_cfg, doc_cfg_hide))]
+#![cfg_attr(docsrs, doc(cfg_hide(doc)))]
```
daxpedda, line 143:
```suggestion
#![cfg_attr(docsrs, feature(doc_auto_cfg, doc_cfg_hide), doc(cfg_hide(doc)))]
```

daxpedda (approved):
I feel like this could become annoying to maintain, but generally a really good change for users!

src/lib.rs
madsmtm: Oh nice!

kchibisov:
I don't quite get the motivation behind this, it could be really confusing to show it all the time, given that you should pay attention to metadata, like where the function actually works.

I guess the only argument pro this is that it's easier to find docs for other targets, but I'd expect when you work with cross platform code you open docs for the target you want, since while winit can do such a thing with `doc` most stuff in projects doesn't have docs and internal stuff is obviously missed from the docs.

madsmtm:
## Motivation

You get to see immediately what functionality is available, without having to select on docs.rs the platform you're currently working on.

This would probably have gone a long way towards the user [here](https://github.com/rust-windowing/winit/issues/3072#issuecomment-1699854226) finding a solution for their problem quicker.

The way the platform module looks on docs.rs now:
now

After this change:
after

kchibisov:
I mean, if they do cross dev, they should open docs with the correct target in the end. I'm just afraid to maintain that and work with such docs because I'm sort of really used to how it works based on target and I use the docs based on target.

The docs.rs even has a field to show docs per target, which sort of implies that they work for this target, but if we always show all the docs you can't really rely on your target picking.

Like for example I don't even want to see docs for macOS when writing code for linux, like why would I even have to read them and in my search, if they are just irrelevant for me?

madsmtm:
Hmm, I definitely get your points! I think the best answer to that would be that the `rustdoc` team has implemented this feature _because_ it's useful for users. I actually thought I could find a lot of examples of this, but it's quite hard since a lot of Rust code is cross-platform. (I also found examples of crates that don't do this, so take it with a grain of salt). But at least two examples:
- [`std::os`](https://doc.rust-lang.org/stable/std/os/index.html)
- [`tokio::net`](https://docs.rs/tokio/1.32.0/tokio/net/index.html#modules) (and the rest of `tokio`)

Myself, I've often wished `libc` did this, I find it really difficult to know the exhaustive list of platforms e.g. [`libc::pthread_main_np`](https://docs.rs/libc/0.2.146/x86_64-apple-darwin/libc/fn.pthread_main_np.html) is available on.

(If you're developing locally, you could always just do `cargo doc --open`, at least for now it'll only show docs for your own target/feature combinations. It's in any case always the best option, since it will show it for _exactly_ your target, instead of just a target that's "close enough" (e.g. for `winit` I'd have to pick `x86_64-apple-darwin` on docs.rs even though I'm currently using `aarch64-apple-darwin`)).

kchibisov:
@madsmtm you can just pass `--target` to like anything and get the docs you want. If something. It's a bit weird when it comes to discoverability, indeed, but it's not like if we change something the ecosystem will change.

Also, have you looked into how rust does the docs in stdlib? They have `#[doc(cfg(windows))]` stuff and it somehow shows, but tells me explicitly that the thing is only available with windows right when I try to look into stuff. Though, maybe rust just builds docs differently?

daxpedda:
I think "I'm not used to seeing the other platforms in the documentation" should be trumped by "increased discovery of platform-specific APIs in a cross-platform library".

About the maintenance issue: @madsmtm has done good effort in putting this together and it seems quite straightforward to me. The way Winit is structured lends itself incredibly well to this system, because we have separated the files so neatly.

I'm still strongly in favor of this change. But I acknowledge this comes with a maintenance burden, which I belief considering the benefits and the current structure of Winit is negligible.

Cc https://github.com/rust-lang/rust/issues/1998.

kchibisov:
@madsmtm the issue I have after this change, is that it looks like this:

picture

So I have no information whatsoever to determine what is available for me and what is not. I was just using `cargo doc` and that's what I use daily to see the docs.

kchibisov:
Yeah, you need nightly for that and `doc_auto_cfg`, so the change is not applicable for us.

MarijnS95:
Looking forward to seeing this, especially in light of #3178, though hope it doesn't conflict too badly :sweat_smile:

madsmtm:
> shouldn't we be able to only expose all other platforms when we also require `doc_auto_cfg`? E.g. with `cfg(docsrs)`?

Apologies for the delay, I've done so now.

src/platform/web.rs
daxpedda, line 1: I can't seem to leave a comment at `WindowBuilderExtWebSys::with_canvas()`, but adding the following on the bottom would be nice:
```rust
    #[cfg_attr(
        not(wasm_platform),
        doc = "",
        doc = "[`HtmlCanvasElement`]: #only-available-on-wasm"
    )]
```

src/platform/android.rs
```diff
@@ -89,5 +89,16 @@ pub mod activity {
...
+
+    #[cfg(not(android_platform))]
+    #[doc(hidden)]
+    pub struct Rect(());
+    #[cfg(not(android_platform))]
+    #[doc(hidden)]
+    pub struct ConfigurationRef(());
+    #[cfg(not(android_platform))]
+    #[doc(hidden)]
+    pub struct AndroidApp(());
```
daxpedda, line 103:
```suggestion
    pub struct Rect;
    #[cfg(not(android_platform))]
    #[doc(hidden)]
    pub struct ConfigurationRef;
    #[cfg(not(android_platform))]
    #[doc(hidden)]
    pub struct AndroidApp;
```

daxpedda (changes requested):
Thank you for doing this, this is so cool!

src/platform/web.rs
```diff
@@ -36,8 +36,13 @@ use crate::window::CustomCursor;
 use crate::window::{Window, WindowBuilder};
 use crate::SendSyncWrapper;
 
+#[cfg(wasm_platform)]
 use web_sys::HtmlCanvasElement;
 
+#[cfg(not(wasm_platform))]
+#[doc(hidden)]
+pub struct HtmlCanvasElement(());
```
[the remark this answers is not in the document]
madsmtm: This was actually intentional, since I wanted to make sure it was unconstructable. But I'm fine with either, it's private anyway, and won't work even if someone compiled with `--cfg=docsrs`.

src/lib.rs
```diff
@@ -154,7 +154,11 @@
 #![deny(unsafe_op_in_unsafe_fn)]
 #![cfg_attr(feature = "cargo-clippy", deny(warnings))]
 // Doc feature labels can be tested locally by running RUSTDOCFLAGS="--cfg=docsrs" cargo +nightly doc
-#![cfg_attr(docsrs, feature(doc_auto_cfg))]
+#![cfg_attr(
+    docsrs,
+    feature(doc_auto_cfg, doc_cfg_hide),
+    doc(cfg_hide(doc, docsrs))
```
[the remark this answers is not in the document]
madsmtm: Right, not any more. I'd still suggest to keep it, since we might in the future, but I'm fine with either.

src/platform/web.rs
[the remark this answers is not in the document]
daxpedda: I see, the only way I'm swinging more towards removing it is because I know I will forget why later and leaving a comment there seems overkill :sweat_smile:

But I'm fine with either.

daxpedda (approved):
LGTM!

PR #3077, opened 2023-09-01
Mark `startup_notify` unsafe functions as safe

kchibisov:
They are safe, since they use the rust `std::env` stuff. Making them safe lets downstream to determine that `std::env` is used and not the `libc` env manipulation routines, which are unsafe.


notgull (approved):
I was the one to originally add this in. On a reread of the relevant docs I'm now fine with this, although our documentation should mention that it uses `env::unset_var`.

kchibisov:
I'm not sure that we should mention internal details, since you should expect `std` stuff by default.

PR #3082, opened 2023-09-03
Web Async Rework

daxpedda:
I tried splitting up the commits in a way that makes sense to review for others but I also for my own sanity.

This basically doesn't use a channel anymore but a simple `AtomicUsize` to measure the amount of user events sent to appropriately wake up the event loop instead of letting the event loop restart for every user event. Also drops the `std::mpsc` channel which should improve the "performance" as well.

I added a bunch of related improvements on top as well.

Follow-up to #3081.


src/platform_impl/web/async/channel.rs
```diff
@@ -0,0 +1,101 @@
...
+}
+
+pub struct AsyncSender<T> {
+    // We need to wrap it into a `Mutex` to make it `Sync`. So the sender can't
+    // be accessed on the main thread, as it could block. Additionally we need
+    // to wrap it in an `Arc` to make it clonable on the main thread without
+    // having to block.
+    sender: Arc<Mutex<Sender<T>>>,
+    closed: Arc<AtomicBool>,
+    waker: Arc<AtomicWaker>,
```
notgull, line 35: Instead of three separate allocations, you should put all of this into one allocation.

daxpedda: Yeah, good point, I should probably do this across the board.

notgull:
Once https://github.com/smol-rs/async-channel/pull/59 is merged, would you be able to replace this functionality with `async-channel`?

daxpedda:
Only the `channel` module. I assume it would also be an improvement over wrapping `std::mpsc` into async.

I could look into gating this stuff behind `target_feature = "atomics"` so most users don't have to pay the price of these dependencies, admittedly very small dependencies.

src/platform_impl/web/async/channel.rs
daxpedda: Done.

daxpedda:
This is ready now, but I'm gonna let it simmer for a while.

PR #3083, opened 2023-09-03
iOS: Add configuration for preferred status bar style

arendjr:


src/platform/ios.rs
```diff
@@ -74,6 +75,16 @@ pub trait WindowExtIOS {
...
+
+    /// Sets whether the [`Window`] prefers the status bar hidden.
+    ///
+    /// The default is system-defined.
+    ///
+    /// This changes the value returned by
+    /// [`-[UIViewController preferredStatusBarStyle]`](https://developer.apple.com/documentation/uikit/uiviewcontroller/1621416-preferredstatusbarstyle?language=objc),
+    /// and then calls
+    /// [`-[UIViewController setNeedsStatusBarAppearanceUpdate]`](https://developer.apple.com/documentation/uikit/uiviewcontroller/1621354-setneedsstatusbarappearanceupdat?language=objc).
+    fn set_preferred_status_bar_style(&self, status_bar_style: UIStatusBarStyle);
```
madsmtm, line 87: Since `UIStatusBarStyle` is private, this will not actually be usable.

Create a new public enum in this file instead.

src/platform/ios.rs
```diff
@@ -74,6 +75,16 @@ pub trait WindowExtIOS {
...
     fn set_prefers_status_bar_hidden(&self, hidden: bool);
+
+    /// Sets whether the [`Window`] prefers the status bar hidden.
+    ///
+    /// The default is system-defined.
+    ///
+    /// This changes the value returned by
+    /// [`-[UIViewController preferredStatusBarStyle]`](https://developer.apple.com/documentation/uikit/uiviewcontroller/1621416-preferredstatusbarstyle?language=objc),
+    /// and then calls
+    /// [`-[UIViewController setNeedsStatusBarAppearanceUpdate]`](https://developer.apple.com/documentation/uikit/uiviewcontroller/1621354-setneedsstatusbarappearanceupdat?language=objc).
```
madsmtm, line 86: Nit: I think this should more be stated in a sort of "you, the user, don't need to call `setNeedsStatusBarAppearanceUpdate`" way.

madsmtm (changes requested):
Yeah, this makes sense to have, thanks!

src/platform_impl/ios/uikit/status_bar.rs
```diff
@@ -0,0 +1,16 @@
...
+
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+#[allow(dead_code)]
+#[repr(isize)]
+pub enum UIStatusBarStyle {
+    #[default]
+    Default = 0,
+    LightContent = 1,
+    DarkContent = 3,
+}
```
kchibisov, line 12: Could we just use `with_theme` that we have, it's exactly the same `light`/`dark` thing?

madsmtm: Agreed

arendjr: I tried following this suggestion, but as I'm implementing it, I'm starting to think it may not be a good idea after all. The thing is, the status bar style doesn't define a *theme*, it merely defines the color of the text and the icons in the status bar. The actual theme, including what is displayed in the background of the status bar, is still determined fully by the app developer. But if they use a light background (what you may call a "light theme"), they should choose `DarkContent`, because it needs to contrast. And similarly "dark themes" require using light content.

Of course I could just reverse the meaning in code, so that `Theme::Dark` maps to `LightContent` and `Theme::Light` to `DarkContent`, but it feels confusing at best.

Alternatively, I could also *not* revert the meaning and specify that `Theme::Light` means light content, but that's confusing too since I feel it would invert the meaning of "theme" compared to other platforms.

So given that there's already the iOS-specific `set_prefers_status_bar_hidden()` I'm now thinking it's better to just leave `set_preferred_status_bar_style()` right next to it, so they're easier to find and use together, and hopefully it'll be clearer to just use the iOS-specific terminology, since app developers usually need to configure it elsewhere in their `.plist` as well, where the same terminology is used.

kchibisov: theme defines a theme not controlled by your app directly. On macOS it's the style of the header bar.

arendjr: Right, but that means the situation is not really comparable to iOS, right? On iOS the background of the status bar is controlled directly by the app, while `set_preferred_status_bar_style()` controls only the foreground color of the text and icons in the status bar.

src/platform/ios.rs
```diff
@@ -270,3 +308,17 @@ bitflags! {
             | ScreenEdge::BOTTOM.bits() | ScreenEdge::RIGHT.bits();
     }
 }
+
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+#[allow(dead_code)]
+#[repr(isize)]
+pub enum UIStatusBarStyle {
```
madsmtm, line 315: The convention in this file is to call this `StatusBarStyle`, without the `UI` prefix.

Also, I'd prefer to not have a `repr` on this, instead we should have an internal `UIStatusBarStyle` that this maps to.

src/platform/ios.rs
```diff
@@ -270,3 +308,17 @@ bitflags! {
...
+    #[default]
+    Default = 0,
+    LightContent = 1,
+    DarkContent = 3,
+}
+
+unsafe impl Encode for UIStatusBarStyle {
```
madsmtm, line 322: This impl exposes the version of our dependency on `objc2` to the outside world; please move it to an internal struct.

madsmtm (changes requested):

src/platform_impl/ios/uikit/status_bar.rs
kchibisov: We could probably somehow make it clear with the docs? For example, when you set a `with_theme(Dark)` it'll be white text, and the other way black text. Like I really don't want to have more options, and in general applications do match the content. If they want to make a contrast, they surely can achieve so.

Just tell in the docs explicitly what this variable is actually doing.

madsmtm: Actually, I think `with_theme` should be using `overrideUserInterfaceStyle`, what I was agreeing to was to reuse the `Theme` struct.

Could someone with a newer iOS device than I check if setting that also changes the status bar style? If not, then I believe the current approach (add a specific option to control the preferred status bar style) is fine.

arendjr: I agree it would make sense if `with_theme()` was implemented to use `overrideUserInterfaceStyle()`. There the `Theme` struct makes sense indeed.

But regardless of whether the theme affects the *default* status bar look (which it doesn't appear to always do, see for instance: https://stackoverflow.com/questions/70518085/how-to-change-status-bar-colour-whenever-i-overrideuserinterfacestyle), it wouldn't be sufficient for all situations. For instance, imagine an app that uses the light theme, but displays a dark background in the upper part of the screen. Such an app would still want to explicitly override the status bar style, so I see the two functions more as orthogonal to one another.

src/platform/ios.rs
```diff
@@ -270,3 +307,12 @@ bitflags! {
             | ScreenEdge::BOTTOM.bits() | ScreenEdge::RIGHT.bits();
     }
 }
+
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+#[allow(dead_code)]
```
madsmtm, line 312: Nit: The type is public, so no need for this
```suggestion
```

madsmtm (approved):
Perfect, thanks!

mockersf (approved):

PR #3086, opened 2023-09-04
Use xbuild instead of cargo apk in the CI

notgull:
In addition to cargo apk being deprecated, xbuild also has support for ios.


xbuild-target/README.md
```diff
@@ -0,0 +1,4 @@
+# xbuild-target
+
+This is a crate that can be built via `x build` in order to simplify CI testing.
```
madsmtm, line 3: `xbuild` does not support building examples?

madsmtm (commented):
Huh, I didn't know it was deprecated. Do you have a reference on why?

notgull:
> Huh, I didn't know it was deprecated. Do you have a reference on why?

Yes, `cargo apk` is deprecated, see https://github.com/rust-mobile/cargo-apk/issues/31#issuecomment-1696820576

xbuild-target/README.md
notgull: Unfortunately it does not.

notgull:
CI is taking a while to be queued. What's that about?

madsmtm:
See https://github.com/rust-windowing/winit/actions/runs/6077445578

notgull:
The CI error is #3903

kchibisov:
r @rib @MarijnS95

notgull:
@MarijnS95 Any chance you can review this in the near future?

xbuild-target/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "xbuild-target"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies.winit]
+version = "0.29.1-beta"
+path = ".."
+features = ["android-game-activity"]
+
```
MarijnS95, line 10: Double trailing newline?

xbuild-target/README.md
MarijnS95: Mind if I get around to "quickly" hacking that in (you'd have to install `xbuild` from the `master` branch in CI) to get this going, and save you the temporary example?

xbuild-target/src/main.rs
```diff
@@ -0,0 +1,6 @@
+fn main() {
+    winit::event_loop::EventLoop::new()
+        .unwrap()
+        .run(|_, _| todo!())
```
MarijnS95, line 4: A real implementation is probably provided by the example that was being compiled before? I'd rather test that, _including_ the part that forwards `android_activity::AndroidApp` into `winit`.

.github/workflows/ci.yml
```diff
@@ -35,19 +35,20 @@ jobs:
...
-            platform: { name: 'Android', target: aarch64-linux-android, os: ubuntu-latest, options: '--package=winit --features=android-native-activity', cmd: 'apk --' }
+            platform: { name: 'Android', target: aarch64-linux-android, os: ubuntu-latest, options: '--package=winit --features=android-native-activity' }
+          - toolchain: 'stable'
+            platform: { name: 'Android', target: aarch64-linux-android, os: ubuntu-latest, options: '--package=winit --features=android-native-activity' }
+          - toolchain: 'nightly'
+            platform: { name: 'Android', target: aarch64-linux-android, os: ubuntu-latest, options: '--package=winit --features=android-native-activity' }
+
 
     env:
       # Set more verbose terminal output
```
MarijnS95, line 54: Android is the only platform utilizing `matrix.platform.cmd` and `$CMD`, let's remove it from this list.

.github/workflows/ci.yml
```diff
@@ -139,6 +141,7 @@ jobs:
     - name: Build tests with serde enabled
       if: >
         !contains(matrix.platform.target, 'redox') &&
+        !contains(matrix.platform.target, 'android') &&
         matrix.toolchain != '1.65.0'
       run: cargo $CMD test --no-run $OPTIONS --features serde
```
MarijnS95, line 146: Stray `$CMD` (are there more?).

.github/workflows/ci.yml
```diff
@@ -87,41 +88,42 @@ jobs:
...
-      with:
-        path: ~/.cargo/bin/cargo-apk
-        # Change this key if we update the required cargo-apk version
-        key: cargo-apk-v0-9-7
-
-    - uses: dtolnay/rust-toolchain@master
-      if: contains(matrix.platform.target, 'android') && (steps.cargo-apk-cache.outputs.cache-hit != 'true')
+    - name: Install xbuild
+      uses: taiki-e/install-action@v2
+      if: contains(matrix.platform.target, 'android') || contains(matrix.platform.target, 'ios')
```
MarijnS95, line 93: Already prepping for using `xbuild` for `ios`?

.github/workflows/ci.yml
```diff
@@ -87,41 +88,42 @@ jobs:
...
+
     - name: Check documentation
       run: cargo doc --no-deps $OPTIONS --document-private-items
 
-    - name: Build crate
-      run: cargo $CMD build $OPTIONS
+    - name: Build crate 
+      run: cargo build $OPTIONS
+
+    - name: Package crate for Android
```
MarijnS95, line 113: Nit (same for iOS below): "build & package"

xbuild-target/README.md
```diff
@@ -0,0 +1,4 @@
...
+
```
MarijnS95, line 4: Nit: double trailing newline.

MarijnS95:
@notgull sure, done. I've been out of the country for quite some time, only got to you `r?` now.

xbuild-target/README.md
notgull: `install-action` can only install released versions of projects.

notgull:
I'm not getting the linker error on my machine

.github/workflows/ci.yml
notgull: It's already being used

xbuild-target/src/main.rs
notgull: I added an actual Android winit app, but it looks like there is a linker error that doesn't occur on my machine.

.github/workflows/ci.yml
```diff
@@ -87,41 +87,43 @@ jobs:
...
 
-    - name: Build crate
-      run: cargo $CMD build $OPTIONS
+    - name: Build crate 
+      run: cargo build $OPTIONS
+
+    - name: Build and package crate for Android
+      if: contains(matrix.platform.target, 'android')
+      # TODO: Figure out why this build fails on Github Actions
+      run: x build -p ios-xbuild-target --platform android --arch arm64
```
kchibisov, line 115: How does it work? Does it simply abort now or it builds, but not with `android` target?

android-xbuild-target/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+version = "0.1.0"
+edition = "2021"
+
+[lib]
+crate_type = ["cdylib"]
+
+[dependencies.winit]
+version = "0.29.1-beta"
+path = ".."
+features = ["android-game-activity"]
```
kchibisov, line 12: I think we always tested `native` one on the CI?

android-xbuild-target/README.md
```diff
@@ -0,0 +1,3 @@
+# android-xbuild-target
+
+This is a crate that can be built via `x build` in order to simplify CI testing.
```
kchibisov, line 3: Maybe we should create a common `xbuild` dir and put these 2 crates there with common docs.

android-xbuild-target/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "android-xbuild-target"
+version = "0.1.0"
```
kchibisov, line 3: disable `publish`.

kchibisov (changes requested):

android-xbuild-target/Cargo.toml
MarijnS95: Testing `native` should prevent issues with having to compile native code :+1:

xbuild/ios-target/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "ios-xbuild-target"
```
MarijnS95, line 2: Nit: given the directory structure I'd name these `xbuild-ios-target`. After all, `xbuild` is targeting an architecture, the architecture isn't targeting `xbuild`.

MarijnS95 (approved):
LGTM, just unfortunate that we cannot have `xbuild` target examples yet :/

android-xbuild-target/Cargo.toml
kchibisov: I think we changed that to `native` and it still failed, but we figured in `#winit` with @notgull what is going on. `xbuild` just doesn't pull the platform and we miss the `libandroid` and `liblog` on the CI.

kchibisov: You can look into https://github.com/rust-windowing/winit/pull/3149 and I got further with xbuild errors.

MarijnS95: The `xbuild` design intends to work more-or-less out of the box (when `clang` is installed), so it's quite strange to see this.

kchibisov: Well, the main issue was that it was passing `ld` linker, which is a `ld.bfd` which can't cross-link, unlike `lld`. And then we had issues with finding `libandroid`, because I think it used the system clang & lld and they don't know about libraries which `xbuild` could bring.

kchibisov: e.g. https://github.com/rust-mobile/xbuild/issues/148

MarijnS95: > And then we had issues with finding `libandroid`, because I think it used the system clang & lld and they don't know about libraries which `xbuild` could bring.

Might that be a `rustflags` precedence issue?

kchibisov: Could be, but I'm not familiar with `xbuild`. I think we tried without `RUSTFLAGS` at all initially. You can open PR based on the one I've linked and try to fix CI if you're familiar with how `xbuild` (I'm not, I can just decide based on what I see in the logs).

notgull:
Closing for now as this is blocked on https://github.com/rust-mobile/xbuild/issues/158

PR #3091, opened 2023-09-09
Fix WaitUntil on X11

Riateche:
It seems that WaitUntil is currently broken on X11: no events are received when the timer expires, unless there is an actual window event. This issue seems to be introduced in c47d08. This PR adds an additional condition to the spurious wake up check, recognizing ResumeTimeReached as a valid cause to wake up.

### DIFF
--- a/src/platform_impl/linux/x11/mod.rs
+++ b/src/platform_impl/linux/x11/mod.rs
@@ -505,18 +505,6 @@ impl<T: 'static> EventLoop<T> {
             return;
         }
 
-        // False positive / spurious wake ups could lead to us spamming
-        // redundant iterations of the event loop with no new events to
-        // dispatch.
-        //
-        // If there's no readable event source then we just double check if we
-        // have any pending `_receiver` events and if not we return without
-        // running a loop iteration.
-        // If we don't have any pending `_receiver`
-        if !self.has_pending() && !self.state.x11_readiness.readable {
-            return;
-        }
-
         // NB: `StartCause::Init` is handled as a special case and doesn't need
         // to be considered here
         let cause = match self.control_flow() {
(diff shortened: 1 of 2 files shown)


notgull (approved):
Looks good to me!

kchibisov (changes requested):
Could you add a changelog entry for this one?

StarStarJ:
I am currently testing the new changes and noticed this bug too. However in my case I was using poll
```rs
event_loop.set_control_flow(winit::event_loop::ControlFlow::Poll)
```

despite the description, the loop did not run the whole time

Additionally to this patch i added a
```rs
| StartCause::Poll
```

to the `matches!` call
which seem to have fixed it.

I only tested this with x11. So I don't know if this affects other backends too

src/platform_impl/linux/x11/mod.rs
```diff
@@ -540,6 +528,21 @@ impl<T: 'static> EventLoop<T> {
...
+        // If there's no readable event source then we just double check if we
+        // have any pending `_receiver` events and if not we return without
+        // running a loop iteration.
+        // If we don't have any pending `_receiver`
+        if !self.has_pending()
+            && !self.state.x11_readiness.readable
+            && !matches!(&cause, StartCause::ResumeTimeReached { .. })
+        {
+            return;
+        }
```
kchibisov, line 544:
```suggestion
        // False positive / spurious wake ups could lead to us spamming
        // redundant iterations of the event loop with no new events to
        // dispatch.
        //
        // If there's no readable event source then we just double check if we
        // have any pending `_receiver` events and if not we return without
        // running a loop iteration.
        // If we don't have any pending `_receiver`
        if !self.has_pending() && !matches!(&cause, StartCause::ResumeTimeReached { .. } | StartCause::Poll)
        {
            return;
        }
```

kchibisov: I think it should be like that, actually. The readiness check seems arbitrary given that we have other sources of events, and X11 readiness has nothing to do with them.

Riateche:
I've added a changelog entry.

kchibisov (approved):

kchibisov:
Thank you!

Issue #3093, opened 2023-09-10
Update core graphics to v0.23


notgull:
This is causing `cargo deny` breakage in CI, as `winit` currently uses v0.22 while `softbuffer` uses v0.23.

PR #3097, opened 2023-09-15
Ignore duplicate dependencies in CI

daxpedda:
I added duplicate dependencies to the ignore list that are breaking the CI.

See https://github.com/servo/core-foundation-rs/issues/634 and https://github.com/rust-windowing/winit/issues/3093.


kchibisov (approved):

kchibisov:
Ah, I guess that's why I got confused why we black list duplicates. It's because our tests have them, but the actual winit doesn't have them...

PR #3098, opened 2023-09-15
Install `cargo-apk` with the stable toolchain

daxpedda:
I noticed that the `cargo-apk` installation fails because it's currently broken on nightly, but even if we want to test Android on nightly, we require `cargo-apk` to be installed with the nightly toolchain.

Also removed caching nightly and stable `cargo-apk` separately, the output is the same now after all.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,12 +64,6 @@ jobs:
     steps:
     - uses: actions/checkout@v3
 
-    - uses: dtolnay/rust-toolchain@master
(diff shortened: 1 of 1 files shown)


notgull:
`cargo-apk` is deprecated, we should be using `xbuild` instead, see #3086

kchibisov (approved):
The xbuild should be handled separately and shouldn't block the CI fix.

Especially given that I'd like to know on xbuild from the android maintainers in the first place.

Issue #3105, opened 2023-09-19
Massive stutters in firefox in 0.29.0-beta, 0.29.1-beta and master (games)


cybersoulK:
I spent the last several hours debugging this.
My game Normal behaviour in firefox is 120 fps stable
 
 I finally troubleshooted that 0.29.0-beta was causing issues in comparison with what i had previously: ed1425416813491380ce351816ab3f513c892058. 
Then i tested all of the versions including the current master 399edf57106aef4c36198c9513beb8ffb1aff349.

The master is by far the worst. 0.29.0-beta and 0.29.1-beta was 120 fps 90% of the time and then stutters down to 20 fps. The master stays at 25fps most of the time and occasionally freezes for 1 Second!!

Here is a video of the issue in current master: https://youtu.be/VCMHpYqanMg

I tested with the latest stable firefox.

kchibisov:
Given that we have no idea what you're doing, I'd suggest to at least try to ask for `redraw_requested` on the window. There were also some different polling modes added for the `web`.

Maybe you just do `Poll` and draw, but we have no idea.

cybersoulK:
i did window.redraw_requested at the end of RedrawRequested and inside AboutToWait.
is there something that i am missing?

daxpedda:
> i tried window.redraw_requested at the end of RedrawRequested, and also tried it inside AboutToWait.
> is there something that i am missing?

It really doesn't matter where you call `redraw_requested()` from.

> I also don't touch ControlFlow at all, it says it's Poll by default

Using `Poll` in Firefox will currently just not work. But Winit can't do anything about that, it's just an issue with Firefox: `requestAnimationFrame()` will get throttled under pressure ([Bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1427603)).

The reason it was working before is `Poll` was using rAF, which was incorrect. If you really need `Poll` in Firefox, consider using `PollStrategy::IdleCallback` from #3084, as @kchibisov already mentioned.

@kchibisov I would like to know how `Poll` is used in practice on other backends: because it just loops as fast as possible it will basically consume 100% CPU. For that reason I've never used `Poll` in practice except when testing stuff in Winit.

---

In any case, I will close this as it isn't really actionable from our side. @cybersoulK we are also guessing at your problem here because we don't have any information. If you can put together something reproducable it would be ideal.

kchibisov:
[...]

PR #3106, opened 2023-09-20
Make `ControlFlow::Wait` the default

daxpedda:
After some discussion in https://github.com/rust-windowing/winit/issues/3105 I was asking myself why `Poll` is the default. So I wanted to know what everybody thinks about switching the default to `Wait`.

I will make a follow-up to try and improve the documentation around our recommendations on which `ControlFlow` to choose, but wanted to keep it separate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On Web, add `EventLoopWindowTargetExtWebSys` and `PollStrategy`, which allows to set different strategies for `ControlFlow::Poll`. By default the Prioritized Task Scheduling API is used, but an option to use `Window.requestIdleCallback` is available as well. Both use `setTimeout()`, with a trick to circumvent throttling to 4ms, as a fallback.
 - Implement `PartialOrd` and `Ord` for `MouseButton`.
 - On X11, fix event loop not waking up on `ControlFlow::Poll` and `ControlFlow::WaitUntil`.
+- **Breaking:** Change default `ControlFlow` to `Wait`.
 
 # 0.29.1-beta
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On Web, add `EventLoopWindowTargetExtWebSys` and `PollStrategy`, which allows to set different strategies for `ControlFlow::Poll`. By default the Prioritized Task Scheduling API is used, but an option to use `Window.requestIdleCallback` is available as well. Both use `setTimeout()`, with a trick to circumvent throttling to 4ms, as a fallback.
 - Implement `PartialOrd` and `Ord` for `MouseButton`.
 - On X11, fix event loop not waking up on `ControlFlow::Poll` and `ControlFlow::WaitUntil`.
-- **Breaking:** Change default `ControlFlow` to `Wait`.
+- **Breaking:** Change default `ControlFlow` from `Poll` to `Wait`.
 
 # 0.29.1-beta
 
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -151,9 +151,9 @@ impl<T> fmt::Debug for EventLoopWindowTarget<T> {
 ///
 /// Indicates the desired behavior of the event loop after [`Event::AboutToWait`] is emitted.
 ///
-/// Defaults to [`Poll`].
+/// Defaults to [`Wait`].
 ///
-/// [`Poll`]: Self::Poll
+/// [`Wait`]: Self::Wait
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
 pub enum ControlFlow {
     /// When the current loop iteration finishes, immediately begin a new iteration regardless of
@@ -193,7 +193,7 @@ impl ControlFlow {
 impl Default for ControlFlow {
     #[inline(always)]
     fn default() -> Self {
-        Self::Poll
+        Self::Wait
     }
 }
 
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -154,13 +154,14 @@ impl<T> fmt::Debug for EventLoopWindowTarget<T> {
 /// Defaults to [`Wait`].
 ///
 /// [`Wait`]: Self::Wait
-#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[derive(Copy, Clone, Debug, Default, PartialEq, Eq)]
 pub enum ControlFlow {
     /// When the current loop iteration finishes, immediately begin a new iteration regardless of
     /// whether or not new events are available to process.
     Poll,
 
     /// When the current loop iteration finishes, suspend the thread until another event arrives.
+    #[default]
     Wait,
 
     /// When the current loop iteration finishes, suspend the thread until either another event
@@ -187,13 +188,6 @@ impl ControlFlow {
             Some(instant) => Self::WaitUntil(instant),
             None => Self::Wait,
         }
-    }
-}
-
-impl Default for ControlFlow {
-    #[inline(always)]
-    fn default() -> Self {
-        Self::Wait
     }
 }
 
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ another library.
 ```rust
 use winit::{
     event::{Event, WindowEvent},
-    event_loop::{ControlFlow, EventLoop},
+    event_loop::EventLoop,
     window::WindowBuilder,
 };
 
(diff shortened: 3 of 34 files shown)


src/platform_impl/ios/app_state.rs
```diff
@@ -234,7 +234,7 @@ impl AppState {
         };
         self.set_state(AppStateImpl::ProcessingEvents {
             event_handler,
-            active_control_flow: ControlFlow::Poll,
```
daxpedda, line 237: @madsmtm again not entirely sure what's going on here, if you don't mind having a second look I would appreciate that.

src/platform_impl/linux/wayland/event_loop/mod.rs
```diff
@@ -230,7 +230,7 @@ impl<T: 'static> EventLoop<T> {
             // Reset the internal state for the loop as we start running to
             // ensure consistent behaviour in case the loop runs and exits more
             // than once.
-            self.set_control_flow(ControlFlow::Poll);
```
daxpedda, line 233: @kchibisov is there something else I have to take into account here?

kchibisov, line 233: You can keep the `Poll` here, I think motivation is to do the first loop iteration non-blocking, because if you see the code, it won't get reset anymore? We just need to change the user default, but if the code really hardcodes the poll it's likely only a one time action.

src/platform_impl/linux/x11/mod.rs
```diff
@@ -442,7 +442,7 @@ impl<T: 'static> EventLoop<T> {
             // Reset the internal state for the loop as we start running to
             // ensure consistent behaviour in case the loop runs and exits more
             // than once.
-            self.set_control_flow(ControlFlow::Poll);
```
kchibisov, line 445: Same here, the reason is to kick start the first entrance to the loop, the rest should be fine.

src/platform_impl/ios/app_state.rs
```diff
@@ -525,8 +525,8 @@ pub fn did_finish_launching(mtm: MainThreadMarker) {
 
     // start waking up the event loop now!
     bug_assert!(
-        this.control_flow == ControlFlow::Poll,
-        "unexpectedly not setup to `Poll` on launch!"
+        this.control_flow == ControlFlow::default(),
+        "unexpectedly not setup to `ControlFlow::default()` on launch!"
```
kchibisov, line 529: Not sure about this one as well.

src/platform_impl/ios/app_state.rs
```diff
@@ -234,7 +234,7 @@ impl AppState {
         };
         self.set_state(AppStateImpl::ProcessingEvents {
             event_handler,
-            active_control_flow: ControlFlow::Poll,
+            active_control_flow: ControlFlow::default(),
```
daxpedda, line 237: @madsmtm again not entirely sure what's going on here, if you don't mind having a second look I would appreciate that.

src/platform_impl/linux/wayland/event_loop/mod.rs
```diff
@@ -230,7 +230,7 @@ impl<T: 'static> EventLoop<T> {
...
+            self.set_control_flow(ControlFlow::default());
```
daxpedda, line 233: @kchibisov is there something else I have to take into account here?

kchibisov (changes requested):
I think the issue we should consider is that initial iteration, should still be with `Poll`, however anything further is with the default behavior. The reason behind is that you technically have events, and the event is the `StartCause`, but with `Wait` you'll wait for the actual events coming from the user or the windowing system, which might not be desirable for the first iteration.

What we should do, is to just switch the user default, but we shouldn't change the way kick start of the loop works, since we always have an event (`StartCause::Init`) and we simply need a wakeup, which `Wait` internally in the backends might not deliver it.

I'd also suggest to emphasize on the `Poll` internally, for the first loop iteration, like it's done in Wayland/X11, so you have more predictable startup, which should be fine across the backends.

daxpedda:
> I think the issue we should consider is that initial iteration, should still be with `Poll`, however anything further is with the default behavior. The reason behind is that you technically have events, and the event is the `StartCause`, but with `Wait` you'll wait for the actual events coming from the user or the windowing system, which might not be desirable for the first iteration.

I'm not following. If `StartCause` is there, shouldn't that kickstart the initial event loop? Or are you saying that some backends just don't work correctly if you start with `ControlFlow::Wait`?

kchibisov:
@daxpedda yes, that's what I'm saying that they could go to sleep in such cases for the first iteration, because the event by its nature, is synthetic, so it can act like it won't wake up the event loop. I think Wayland/X11 backends will need more tuning than what you've done, but there's no issue to keeping the `Poll` for the **first** iteration internally, since it makes the most sense.

In practice, this won't change anything for the end users, since the default polling is still `Wait`, so after the first iteration.

daxpedda:
> In practice, this won't change anything for the end users, since the default polling is still `Wait`, so after the first iteration.

I assume I still have to put it back to `ControlFlow::default()` after the first iteration?
Another question would be if the user uses `EventLoopWindowTarget::control_flow()` would they not see that the `ControlFlow` is `Poll`?

kchibisov:
>I assume I still have to put it back to ControlFlow::default() after the first iteration?
Another question would be if the user uses EventLoopWindowTarget::control_flow() would they not see that the ControlFlow is Poll?

Maybe it's not clear what I suggested, but I'll clarify:

1. When first iteration happens, the internal behavior is if the `Poll` was used or it could be something else, if the backend can guarantee that it'll properly wakeup for the first iteration. The change I've suggested is a conservative approach to that, since I know that it won't break anything internally in the backends.
2. After the first iteration, the behavior set by the user is what should be used, so in
our case, the `EventLoopWindowTarget::control_flow()`. The default is `Wait`, but we shouldn't use the `ControlFlow::default()` in the backends code, we should use the value used by the users, and when the backends requires special polling for a short period of time to remain correctness it should switch to it. But we must not use the `::default()` instead of the `Poll`.
3. Users always see the default as a `Wait` or what **they set**, if the backend changes the way it polls internally for correctness reasons, is up to the backend, it was like that even before.

What you confuse is the policy with the behavior, users are setting the polling policy, how the backend follows that policy is up to the backend. It could use different polling internally, but it should look like the `Wait` behavior for the end user.

kchibisov:
So, how the change should look is that you simply change the default on the `ControlFlow`, and then ensure that the backends don't rely on that default being `Poll`, like ios. Instead they should simply specify the `Poll` when they need it.

daxpedda:
After some IRC conversations we figured out that this was already an issue before this PR.
See https://github.com/rust-windowing/winit/pull/3107.

src/platform_impl/ios/app_state.rs
daxpedda: Moved to https://github.com/rust-windowing/winit/pull/3107, as it seems that this is an issue unrelated to this PR.

src/platform_impl/linux/x11/mod.rs
daxpedda: Moved to https://github.com/rust-windowing/winit/pull/3107, as it seems that this is an issue unrelated to this PR.

src/platform_impl/ios/app_state.rs
daxpedda: Moved to https://github.com/rust-windowing/winit/pull/3107, as it seems that this is an issue unrelated to this PR.

src/event_loop.rs
```diff
@@ -193,7 +193,7 @@ impl ControlFlow {
 impl Default for ControlFlow {
     #[inline(always)]
     fn default() -> Self {
-        Self::Poll
+        Self::Wait
```
kchibisov, line 196: Could you annotate the default on the enum instead of implementing it?

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On Web, add `EventLoopWindowTargetExtWebSys` and `PollStrategy`, which allows to set different strategies for `ControlFlow::Poll`. By default the Prioritized Task Scheduling API is used, but an option to use `Window.requestIdleCallback` is available as well. Both use `setTimeout()`, with a trick to circumvent throttling to 4ms, as a fallback.
 - Implement `PartialOrd` and `Ord` for `MouseButton`.
 - On X11, fix event loop not waking up on `ControlFlow::Poll` and `ControlFlow::WaitUntil`.
+- **Breaking:** Change default `ControlFlow` to `Wait`.
```
kchibisov, line 28:
```suggestion
- **Breaking:** Change default `ControlFlow` from `Poll` to `Wait`.
```

kchibisov (changes requested):

src/event_loop.rs
daxpedda: Done.

CHANGELOG.md
daxpedda: Done.

kchibisov (approved):

PR #3107, opened 2023-09-20
Fix reset to `Poll` after the event loop starts

daxpedda:
Discovered in https://github.com/rust-windowing/winit/pull/3106.

@madsmtm I suspect that the same bug affects iOS:
https://github.com/rust-windowing/winit/blob/c8b685ddbc0ac7d2a122ae47d6ecea968c1ecae5/src/platform_impl/ios/app_state.rs#L237


src/lib.rs
```diff
@@ -50,15 +50,16 @@
...
 //!
+//! // ControlFlow::Poll continuously runs the event loop, even if the OS hasn't
+//! // dispatched any events. This is ideal for games and similar applications.
+//! event_loop.set_control_flow(ControlFlow::Poll);
+//!
+//! // ControlFlow::Wait pauses the event loop if no events are available to process.
+//! // This is ideal for non-game applications that only update in response to user
+//! // input, and uses significantly less power/CPU time than ControlFlow::Poll.
+//! event_loop.set_control_flow(ControlFlow::Wait);
+//!
```
kchibisov, line 61: This section is a bit weird, it's like we have a code, which sets control flows and explains what they do, but it just one discards the other?

kchibisov (changes requested):

src/lib.rs
daxpedda: This documentation has to be improved anyway. But I didn't add it in this PR, I just moved it to reflect the fact that we can change this now *before* `run()`.

We should tackle this in a different PR.

kchibisov: I'd just simply drop it entirely, but could be done in a separate PR.

kchibisov (approved):

Issue #3108, opened 2023-09-20
Add a way to request/show window menu


kchibisov:
Some systems like Wayland/Windows provide a way to ask compositor to show window menu, users can't control the content of such menu, and usually they contain compositor specific actions, like drag/maximize/etc.

Winit already shows the menu on Wayland, but you need a titlebar for that, however, when the titlebar is hidden, users can't request a menu, for example, when they draw title bar on their own.

The api should probably be like that

```
/// The system is free to ignore the request.
Window::show_window_menu(&self);
```

daxpedda:
I removed the Wasm label, as this isn't a think on Web.

kchibisov:
@daxpedda out of curiosity, it's done by the browser, right? 

Also, how the web apps changing right click popup works? For example, what youtube does when you right click on the video, it does show its custom popup, but how it's done?

daxpedda:
Oh, I thought you were talking about the window menu **bar**, but you were just talking about the context menu. Sorry, not sure what the right terms here are.

> @daxpedda out of curiosity, it's done by the browser, right? And users can't ask browser to show browser default right click menu.

By default the default browser context menu is always shown on right-click. Currently we just suppress the default context menu, controlling  that suppression wouldn't be a problem at all.

> Also, how the web apps changing right click popup works? For example, what youtube does when you right click on the video, it does show its custom popup, but how it's done?

There are no custom context menus in that sense, there is only the default one and you have no control over it.
E.g. what YouTube does is just suppress the default one and add an element with CSS styling and stuff on right-click.

kchibisov:
Ah, yeah, we need this right click menu thing, so it seems like it's applicable for web?

daxpedda:
What is actually the intention of this API exactly? Web can just *allow* the context menu to be shown when using right-click. It can't actually open it on demand.

kchibisov:
Ah, ok. Then it's different on the web.

The intent is to show a compositor menu when you click on decorations. So users can pick compositor action.

It looks like

![1](https://github.com/rust-windowing/winit/assets/27620401/c35e482c-fcdb-40c9-8494-61b38802b640)

daxpedda:
So you aren't talking about just a regular context menu, but specifically the context menu related to decorations/the title bar?

Why did you talk about the right-click menu here: https://github.com/rust-windowing/winit/issues/3108#issuecomment-1727865169; was that unrelated?

kchibisov:
[...]

PR #3109, opened 2023-09-21
Added the show system menu method to window

YouKnow-sys:
This PR add support for showing system menu with calling `show_system_menu` method on `Window`.
this can be useful when we are drawing our custom titlebar and want to keep the system menu as well.
currently only supported on windows.

related to: #3108 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,9 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On Web, add `EventLoopWindowTargetExtWebSys` and `PollStrategy`, which allows to set different strategies for `ControlFlow::Poll`. By default the Prioritized Task Scheduling API is used, but an option to use `Window.requestIdleCallback` is available as well. Both use `setTimeout()`, with a trick to circumvent throttling to 4ms, as a fallback.
 - Implement `PartialOrd` and `Ord` for `MouseButton`.
 - On X11, fix event loop not waking up on `ControlFlow::Poll` and `ControlFlow::WaitUntil`.
-- Add `Window::show_window_menu()` which shows the system menu at the mouse position. Currently only implemented on Windows.
+- **Breaking:** Change default `ControlFlow` from `Poll` to `Wait`.
+- **Breaking:** remove `DeviceEvent::Text`.
+- Add `Window::show_window_menu` which shows the system menu at a specified position or the current cursor position. Currently only implemented on Windows.
 
 # 0.29.1-beta
 
@@ -89,6 +91,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - **Breaking:** `with_x11_visual` now takes the visual ID instead of the bare pointer.
 - On X11, add a `with_embedded_parent_window` function to the window builder to allow embedding a window into another window.
 - On iOS, add force data to touch events when using the Apple Pencil.
+- On Android, add force data to touch events.
 
 # 0.29.0-beta.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On Web, add `EventLoopWindowTargetExtWebSys` and `PollStrategy`, which allows to set different strategies for `ControlFlow::Poll`. By default the Prioritized Task Scheduling API is used, but an option to use `Window.requestIdleCallback` is available as well. Both use `setTimeout()`, with a trick to circumvent throttling to 4ms, as a fallback.
 - Implement `PartialOrd` and `Ord` for `MouseButton`.
 - On X11, fix event loop not waking up on `ControlFlow::Poll` and `ControlFlow::WaitUntil`.
+- added the new `show_system_menu` to `Window` this method can be showed to show system menu even if decoration is disabled. currently only supported on windows.
 
 # 0.29.1-beta
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - **Breaking:** Moved `ControlFlow::Exit` to `EventLoopWindowTarget::exit()` and `EventLoopWindowTarget::exiting()` and removed `ControlFlow::ExitWithCode(_)` entirely.
 - On Web, add `EventLoopWindowTargetExtWebSys` and `PollStrategy`, which allows to set different strategies for `ControlFlow::Poll`. By default the Prioritized Task Scheduling API is used, but an option to use `Window.requestIdleCallback` is available as well. Both use `setTimeout()`, with a trick to circumvent throttling to 4ms, as a fallback.
 - Implement `PartialOrd` and `Ord` for `MouseButton`.
+- On X11, fix event loop not waking up on `ControlFlow::Poll` and `ControlFlow::WaitUntil`.
 
 # 0.29.1-beta
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On Web, add `EventLoopWindowTargetExtWebSys` and `PollStrategy`, which allows to set different strategies for `ControlFlow::Poll`. By default the Prioritized Task Scheduling API is used, but an option to use `Window.requestIdleCallback` is available as well. Both use `setTimeout()`, with a trick to circumvent throttling to 4ms, as a fallback.
 - Implement `PartialOrd` and `Ord` for `MouseButton`.
 - On X11, fix event loop not waking up on `ControlFlow::Poll` and `ControlFlow::WaitUntil`.
-- added the new `show_system_menu` to `Window` this method can be showed to show system menu even if decoration is disabled. currently only supported on windows.
+- Add `Window::show_system_menu()` which shows the system menu at the mouse position. Currently only implemented on Windows.
 
 # 0.29.1-beta
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On Web, add `EventLoopWindowTargetExtWebSys` and `PollStrategy`, which allows to set different strategies for `ControlFlow::Poll`. By default the Prioritized Task Scheduling API is used, but an option to use `Window.requestIdleCallback` is available as well. Both use `setTimeout()`, with a trick to circumvent throttling to 4ms, as a fallback.
 - Implement `PartialOrd` and `Ord` for `MouseButton`.
 - On X11, fix event loop not waking up on `ControlFlow::Poll` and `ControlFlow::WaitUntil`.
-- Add `Window::show_system_menu()` which shows the system menu at the mouse position. Currently only implemented on Windows.
+- Add `Window::show_window_menu()` which shows the system menu at the mouse position. Currently only implemented on Windows.
 
 # 0.29.1-beta
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,8 @@ And please only add new entries to the top of this list, right below the `# Unre
 
 # Unreleased
 
+- Renamed `EventLoopExtRunOnDemand` / `run_ondemand` to `EventLoopExtRunOnDemand` / `run_on_demand`.
 - Make iOS `MonitorHandle` and `VideoMode` usable from other threads.
-- Fix window size sometimes being invalid when resizing on macOS.
 - On Web, `ControlFlow::WaitUntil` now uses the Prioritized Task Scheduling API. `setTimeout()`, with a trick to circumvent throttling to 4ms, is used as a fallback.
 - On Web, never return a `MonitorHandle`.
 - **Breaking:** Move `Event::RedrawRequested` to `WindowEvent::RedrawRequested`.
@@ -27,7 +27,8 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On X11, fix event loop not waking up on `ControlFlow::Poll` and `ControlFlow::WaitUntil`.
 - **Breaking:** Change default `ControlFlow` from `Poll` to `Wait`.
 - **Breaking:** remove `DeviceEvent::Text`.
-- Add `Window::show_window_menu` which shows the system menu at a specified position or the current cursor position. Currently only implemented on Windows.
+- On Android, fix `DeviceId` to contain device id's.
+- Add `Window::show_window_menu` to request a titlebar/system menu; implemented on Windows for now. 
 
 # 0.29.1-beta
 
@@ -175,6 +176,10 @@ And please only add new entries to the top of this list, right below the `# Unre
   `DeviceEvent::Key` support.
 - **Breaking** `MouseButton` now supports `Back` and `Forward` variants, emitted from mouse events
   on Wayland, X11, Windows, macOS and Web.
+
+# 0.28.7
+
+- Fix window size sometimes being invalid when resizing on macOS 14 Sonoma.
 
 # 0.28.6
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - **Breaking:** Moved `ControlFlow` to `EventLoopWindowTarget::set_control_flow()` and `EventLoopWindowTarget::control_flow()`.
 - **Breaking:** Moved `ControlFlow::Exit` to `EventLoopWindowTarget::exit()` and `EventLoopWindowTarget::exiting()` and removed `ControlFlow::ExitWithCode(_)` entirely.
 - On Web, add `EventLoopWindowTargetExtWebSys` and `PollStrategy`, which allows to set different strategies for `ControlFlow::Poll`. By default the Prioritized Task Scheduling API is used, but an option to use `Window.requestIdleCallback` is available as well. Both use `setTimeout()`, with a trick to circumvent throttling to 4ms, as a fallback.
+- Implement `PartialOrd` and `Ord` for `MouseButton`.
 
 # 0.29.1-beta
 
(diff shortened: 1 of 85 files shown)


CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On Web, add `EventLoopWindowTargetExtWebSys` and `PollStrategy`, which allows to set different strategies for `ControlFlow::Poll`. By default the Prioritized Task Scheduling API is used, but an option to use `Window.requestIdleCallback` is available as well. Both use `setTimeout()`, with a trick to circumvent throttling to 4ms, as a fallback.
 - Implement `PartialOrd` and `Ord` for `MouseButton`.
 - On X11, fix event loop not waking up on `ControlFlow::Poll` and `ControlFlow::WaitUntil`.
+- added the new `show_system_menu` to `Window` this method can be showed to show system menu even if decoration is disabled. currently only supported on windows.
```
daxpedda, line 28:
```suggestion
- Add `Window::show_system_menu()` which shows the system menu at the mouse position. Currently only implemented on Windows.
```

src/window.rs
```diff
@@ -1396,6 +1396,18 @@ impl Window {
...
     }
 
+    /// Show system menu on current mouse position.
+    ///
+    /// There's no guarantee that calling this method show the system menu.
+    /// user can cancel opening of it before it even open.
+    ///
+    /// ## Platform-specific
+    ///
+    /// calling this method **won't** do anything on any platform other then windows.
```
daxpedda, line 1406:
```suggestion
    /// **Android / iOS / macOS / Orbital / Wayland / Web / X11:** Unsupported.
```

notgull:
This function might be a better fit for [`menubar`](https://github.com/madsmtm/menubar).

src/window.rs
```diff
@@ -1396,6 +1396,18 @@ impl Window {
             .maybe_wait_on_main(|w| w.drag_resize_window(direction))
     }
 
+    /// Show system menu on current mouse position.
+    ///
+    /// There's no guarantee that calling this method show the system menu.
+    /// user can cancel opening of it before it even open.
```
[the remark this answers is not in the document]
daxpedda: If I understood you correctly, it might be good for users to know that this action only works properly if connected to a mouse click, similar to [`Window::drag_window()`](https://docs.rs/winit/0.28.6/winit/window/struct.Window.html#method.drag_window).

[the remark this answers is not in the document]
daxpedda: No, this isn't necessary.

I remember it was necessary in `Window::drag_window()` because it was required by the API in some backends. If it's not required for `Window::show_system_menu()` we shouldn't add any unnecessary constraints.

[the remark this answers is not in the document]
daxpedda: > one thing that we can do is allow passing the position of menu to the `show_system_menu` so user can choose where the menu should open... it can be a Option, so if None we use mouse position

It doesn't sound that important to me, but I have no objections if you want to add that.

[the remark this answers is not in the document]
daxpedda: Personally I'm happy to merge like this, but if you want to add more text please feel free to do so!

Maybe we also want to explain what the "system menu" is actually, or maybe just add a link to [Wikipedia](https://en.wikipedia.org/wiki/Common_menus_in_Microsoft_Windows#System_menu). Again, up to you really.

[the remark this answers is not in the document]
daxpedda: That would be much appreciated.

kchibisov:
>This function might be a better fit for [menubar](https://github.com/madsmtm/menubar).

This has nothing to do with the menubars, this is a system menu on the right click on decorations, which is used to show a compositor menu. Winit already shows such a thing, it just, we don't expose the API winit uses for it, which is required for the
user defined client side decorations.

src/window.rs
```diff
@@ -1396,6 +1396,15 @@ impl Window {
             .maybe_wait_on_main(|w| w.drag_resize_window(direction))
     }
 
+    /// Show system menu at current mouse position.
+    ///
+    /// ## Platform-specific
+    ///
+    /// **Android / iOS / macOS / Orbital / Wayland / Web / X11:** Unsupported.
+    pub fn show_system_menu(&self) {
```
kchibisov, line 1404: I think we shouldn't use mouse position for this, it seems like we can use arbitrary position based on your implementation, so just passing `Position` as argument should be good.

That way we can remove the mouse based wording, and say that we request to show window menu at the given `position`.

Also, could we name it `show_window_menu` and also describe what it is and how it usually looks like and document the use of?

kchibisov (changes requested):
Given that there was a mention of the `mouse` or similar happening, I'd say that wayland wants that action to be associated with some input event, like in reaction to mouse click, keyboard button press, touch press, etc.

We should say that the compositor is free to ignore such request though.

src/window.rs
daxpedda: > Given that there was a mention of the `mouse` or similar happening, I'd say that wayland wants that action to be associated with some input event, like in reaction to mouse click, keyboard button press, touch press, etc.

> I think we shouldn't use mouse position for this, it seems like we can use arbitrary position based on your implementation, so just passing `Position` as argument should be good.
> 
> That way we can remove the mouse based wording, and say that we request to show window menu at the given `position`.

Just to clarify: if we add a position parameter we don't need to associate it with some input event, right?

> We should say that the compositor is free to ignore such request though.

It would be nice if we can figure out the details here. When the compositor ignores this and platform-specific behavior if it differs.

> Also, could we name it `show_window_menu` [..]

Is there actually some defined term for this? I found the term "system menu" for Windows, do Wayland or X11 have their own terms?

See https://github.com/rust-windowing/winit/pull/3109#discussion_r1334347502 for previous discussion.

kchibisov: @daxpedda I mean, it's still based on the event, as in, you have to do it, but the position could be arbitrary and not attached to anything. But that's what wayland wants for that, because you have a concept of serial events, so you just must pass a serial from the latest button event.

>It would be nice if we can figure out the details here. When the compositor ignores this and platform-specific behavior if it differs.

it's a compositor menu in the first place, I'm pretty sure no-one ever guarantees that they will show the menu in the end, and it's transparent for the window/winit (at least on Wayland). I'm sure it's similar on windows, since I don't think user is the one closing it, it's done for him.

>Is there actually some defined term for this? I found the term "system menu" for Windows, do Wayland or X11 have their own terms?

https://wayland.app/protocols/xdg-shell#xdg_toplevel:request:show_window_menu 

That's a wayland API for that.

daxpedda: > @daxpedda I mean, it's still based on the event, as in, you have to do it, but the position could be arbitrary and not attached to anything. But that's what wayland wants for that, because you have a concept of serial events, so you just must pass a serial from the latest button event.

If I understand that correctly, you are saying that you can pass an arbitrary position, but you need the serial from the last button event. Does it matter how long ago this button even was as long as it's the last one? Is this the case for windows as well @YouKnow-sys?

> it's a compositor menu in the first place, I'm pretty sure no-one ever guarantees that they will show the menu in the end, and it's transparent for the window/winit (at least on Wayland). I'm sure it's similar on windows, since I don't think user is the one closing it, it's done for him.

I'm not entirely sure if we can equate closing with not showing, but maybe you mean something else? It would be nice if can actually document this in clear terms, like **in which case** does it not show, is it random? @YouKnow-sys is this the case for Windows as well?

If we just don't know the answer to this we can just write "no guarantee" as @YouKnow-sys had before.

kchibisov: >Does it matter how long ago this button even was as long as it's the last one? Is this the case for windows as well

Compositor defined, not specified, just not long ago.

@daxpedda I mean both, system compositor is in-charge of the menu, it can either not show at all or close without user input.

No guarantee means that it simply may not be showed to the user and winit has no control over such menu.

Wayland docs say the following for example:

>There are no guarantees as to what menu items the window menu contains, or even if a window menu will be drawn at all.

daxpedda: Alright then. Would be nice to confirm exactly whats going on in Windows to figure out if this is platform-specific or not.

kchibisov: I don't think this menu thing differs that much tbh, it would be nice to confirm, but just so you know, it's not a **new** thing, we already show it iirc. It just we don't have a way to show it when there're no decorations at all, but default system decorations usually embed such a thing.

daxpedda: I don't think we should mention it's not guaranteed if that's not the case for all platforms. It should go into platform-specifics in that case, which isn't necessary right now because we only have the Windows implementation.

Winit doesn't offer a way to remove `WS_SYSMENU`, right? Because then I don't think we should mention it at all, it becomes a internal detail at this point.

Thank you for verifying this @YouKnow-sys! I will make some suggestions so you don't have to guess around what we want here!

daxpedda: About the mouse position: personally I don't particularly see the point of supplying one because I assume that implementations will only want to show this menu at the mouse position anyway. So I'm fine with how it is right now, but I'm not against adding an optional position.

I think **requiring** a position is problematic in Winit right now because we are missing https://github.com/rust-windowing/winit/issues/883 or https://github.com/rust-windowing/winit/pull/2648, which would require users to save the last known position somewhere. Not that this is a problem, I just don't know what the motivation would be to do this in the first place.

src/window.rs
```diff
@@ -1396,6 +1396,19 @@ impl Window {
             .maybe_wait_on_main(|w| w.drag_resize_window(direction))
     }
 
+    /// Show window menu at current mouse position.
+    ///
+    /// There's no guarantee that calling this method show the window menu.
+    ///
```
daxpedda, line 1402:
```suggestion
```
As we discussed, this isn't the case for Windows, other platforms should add documentation for this in "Platform-specific"s when the implementation is added.

src/window.rs
```diff
@@ -1396,6 +1396,19 @@ impl Window {
             .maybe_wait_on_main(|w| w.drag_resize_window(direction))
     }
 
+    /// Show window menu at current mouse position.
+    ///
+    /// There's no guarantee that calling this method show the window menu.
+    /// 
+    /// ## Platform-specific
+    /// - **Windows:** Window need to have `WS_SYSMENU` window style to be able to show the [`window menu`].
+    /// - **Android / iOS / macOS / Orbital / Wayland / Web / X11:** Unsupported.
```
daxpedda, line 1405:
```suggestion
    /// **Android / iOS / macOS / Orbital / Wayland / Web / X11:** Unsupported.
```
If Winit doesn't have an API to remove `WS_SYSMENU`, I think this would become an implementation detail. If I understand this correctly, right now, showing the window menu will always work on Windows.

src/window.rs
```diff
@@ -1396,6 +1396,19 @@ impl Window {
...
+    /// 
+    /// [`window menu`]: https://en.wikipedia.org/wiki/Common_menus_in_Microsoft_Windows#System_menu
```
daxpedda, line 1407:
```suggestion
    /// 
    /// [`window menu`]: https://en.wikipedia.org/wiki/Common_menus_in_Microsoft_Windows#System_menu
```
I guess we can skip the link, I can't really come up with wording that makes it useful to incorporate it.

src/window.rs
```diff
@@ -1396,6 +1396,19 @@ impl Window {
             .maybe_wait_on_main(|w| w.drag_resize_window(direction))
     }
 
+    /// Show window menu at current mouse position.
+    ///
```
daxpedda, line 1400:
```suggestion
    ///
    /// This is the context menu that is normally shown when interacting with
    /// the title bar. This is useful when implementing custom decorations.
```
@kchibisov is this enough? Or do you want do add more information here?

src/window.rs
daxpedda: > Although we can still get the window handle and modify the flags ourselves using win32 API, so maybe it does worth having in doc (so user don't accidentally remove this flag)

I'm not familiar with the Windows backend, but couldn't adding or removing flags break a whole lot anyway? At that point I would argue we *maybe* want to add some documentation where you can get the handle, telling users that with this handle they can do all sort of shenanigans and break Winit completely.

daxpedda: > although Im not sure what should the position type be... should it be a tuple of `u16` `(u16, u16)`?

`Into<Position>`, like in [`Window::set_cursor_position()`](https://docs.rs/winit/0.29.1-beta/winit/window/struct.Window.html#method.set_cursor_position). It might not work with an `Option` though, in which case just use [`Position`](https://docs.rs/winit/0.29.1-beta/winit/dpi/enum.Position.html).

src/window.rs
daxpedda: There is currently nothing linking to it, so if you want to keep it you have to use it somewhere.

src/window.rs
```diff
@@ -1396,6 +1396,18 @@ impl Window {
...
+    /// This is the context menu that is normally shown when interacting with
+    /// the title bar. This is useful when implementing custom decorations.
+    /// ## Platform-specific
```
daxpedda, line 1403:
```suggestion
    ///
    /// ## Platform-specific
```

src/window.rs
daxpedda: > `TrackPopupMenu` in win32 expect the position to be in screen coordinates, how can exactly get screen coordinates from `Position`?

You will have to do some kind of calculation. Just looking at [`Window::set_cursor_position()`](https://github.com/rust-windowing/winit/blob/2422ea39d0e97fd43698391f84d9300cd169d8cd/src/platform_impl/windows/window.rs#L415-L418), it's using [`ClientToScreen`](https://docs.rs/windows-sys/0.48.0/windows_sys/Win32/Graphics/Gdi/fn.ClientToScreen.html), which according to the [docs](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-clienttoscreen) seems to be what you are looking for.

> also the thing about using `Into<Position>` is if we use just None now we have to use type annotations as well, because then rust wont know what the type is... is there any better option?

> It might not work with an `Option` though, in which case just use [`Position`](https://docs.rs/winit/0.29.1-beta/winit/dpi/enum.Position.html).

That's what I meant here, just use `Option<Position>` instead of `Option<impl Into<Position>>`.

src/window.rs
```diff
@@ -1396,6 +1396,25 @@ impl Window {
             .maybe_wait_on_main(|w| w.drag_resize_window(direction))
     }
 
+    /// Show window menu at a specified position or the current cursor position.
+    ///
+    /// This is the context menu that is normally shown when interacting with
+    /// the title bar. This is useful when implementing custom decorations.
+    ///
+    /// if `position` is `None` the current cursor position will be used.
```
daxpedda, line 1404:
```suggestion
    /// If `position` is `None` the current cursor position will be used.
```

src/window.rs
```diff
@@ -1396,6 +1396,25 @@ impl Window {
...
+    ///
+    /// ### what is a window menu?
+    /// - windows: [`window menu`]
+    ///
```
daxpedda, line 1408:
```suggestion
```
That's not very structured documentation.

src/window.rs
```diff
@@ -1396,6 +1396,25 @@ impl Window {
             .maybe_wait_on_main(|w| w.drag_resize_window(direction))
     }
 
+    /// Show window menu at a specified position or the current cursor position.
```
daxpedda, line 1399:
```suggestion
    /// Show [window menu] at a specified position or the current cursor position.
```
If you really want to add the link, just do it here.

src/window.rs
```diff
@@ -1396,6 +1396,25 @@ impl Window {
...
+    ///
+    /// if `position` is `None` the current cursor position will be used.
+    ///
+    /// ### what is a window menu?
+    /// - windows: [`window menu`]
+    ///
+    /// ## Platform-specific
+    /// **Android / iOS / macOS / Orbital / Wayland / Web / X11:** Unsupported.
+    ///
+    /// [`window menu`]: https://en.wikipedia.org/wiki/Common_menus_in_Microsoft_Windows#System_menu
```
daxpedda, line 1412:
```suggestion
    /// [window menu]: https://en.wikipedia.org/wiki/Common_menus_in_Microsoft_Windows#System_menu
```

daxpedda:
> I think now its mostly done?

Couple of nits on the documentation, otherwise LGTM from me.
I'm not familiar with the Windows backend, so somebody else has to review that.

daxpedda:
Could you rebase for CI to run?

msiglreith (approved):
Windows implementation lgtm

src/window.rs
```diff
@@ -1394,6 +1394,22 @@ impl Window {
             .maybe_wait_on_main(|w| w.drag_resize_window(direction))
     }
 
+    /// Show [window menu] at a specified position or the current cursor position.
+    ///
+    /// This is the context menu that is normally shown when interacting with
+    /// the title bar. This is useful when implementing custom decorations.
+    ///
+    /// If `position` is `None` the current cursor position will be used.
```
kchibisov, line 1402: I don't like this, `the current cursor position` will be used, winit doesn't track it, and tracking it in the `Window` on some platforms, is not really fun thing to do. I'd just suggest to use `Position` since the clients must track the position already to actually handle the button clicks.

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On X11, fix event loop not waking up on `ControlFlow::Poll` and `ControlFlow::WaitUntil`.
 - **Breaking:** Change default `ControlFlow` from `Poll` to `Wait`.
 - **Breaking:** remove `DeviceEvent::Text`.
+- Add `Window::show_window_menu` which shows the system menu at a specified position or the current cursor position. Currently only implemented on Windows.
```
kchibisov, line 30: it's very long, something like that should be enough. 

```suggestion
- Add `Window::show_window_menu` to request a titlebar/system menu; implemented on Windows for now. 
```

kchibisov (changes requested):
looks mostly good.

CHANGELOG.md
kchibisov: a bit more clear in some cases, that it's the one attached to titlebar. One could think about macOS's app menu integrated into their bar, which is not really what it is.

kchibisov (approved):

PR #3110, opened 2023-09-21
Doc: Update Usage in README.md to adjust to new version

rikkaka:
I found that the usage example in README.md is for a past version of `winit`. Since the version displayed in README.md of master branch is `0.29.1-beta`, I'm trying to edit the example to make it suit this version.
Note: There is also some discussion whether the doc should be removed entirely. But before that I believe it's better to correct the mistake at first because it may confuse some people.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,25 +35,23 @@ another library.
 ```rust
 use winit::{
     event::{Event, WindowEvent},
-    event_loop::{ControlFlow, EventLoop},
+    event_loop::EventLoop,
     window::WindowBuilder,
 };
 
 fn main() {
-    let event_loop = EventLoop::new();
+    let event_loop = EventLoop::new().unwrap();
     let window = WindowBuilder::new().build(&event_loop).unwrap();
 
-    event_loop.run(move |event, elwt| {
-        elwt.set_control_flow(ControlFlow::Wait);
-
+    event_loop.run(move |event, _elwt, cf| {
         match event {
             Event::WindowEvent {
                 event: WindowEvent::CloseRequested,
                 window_id,
-            } if window_id == window.id() => elwt.exit(),
-            _ => (),
-        }
-    });
+            } if window_id == window.id() => cf.set_exit(),
+                _ => (),
+            }
+    }).unwrap();
 }
 ```
 
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ Winit is designed to be a low-level brick in a hierarchy of libraries. Consequen
 show something on the window you need to use the platform-specific getters provided by winit, or
 another library.
 
-```rust, ignore
+```rust, no_run
 use winit::{
     event::{Event, WindowEvent},
     event_loop::EventLoop,
@@ -189,7 +189,7 @@ android_logger = "0.11.0"
 ```
 
 And, for example, define an entry point for your library like this:
-```rust
+```rust, no_run
 #[cfg(target_os = "android")]
 use winit::platform::android::activity::AndroidApp;
 
--- a/README.md
+++ b/README.md
@@ -39,19 +39,23 @@ use winit::{
     window::WindowBuilder,
 };
 
-fn main() {
+fn main() -> Result<(), Box<dyn std::error::Error>> {
     let event_loop = EventLoop::new().unwrap();
-    let window = WindowBuilder::new().build(&event_loop).unwrap();
+    let window = WindowBuilder::new().build(&event_loop)?;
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -35,25 +35,23 @@ another library.
...
 use winit::{
     event::{Event, WindowEvent},
-    event_loop::{ControlFlow, EventLoop},
+    event_loop::EventLoop,
     window::WindowBuilder,
 };
 
 fn main() {
-    let event_loop = EventLoop::new();
+    let event_loop = EventLoop::new().unwrap();
```
madsmtm, line 43: Nit: Use `fn main() -> Result<(), Box<dyn std::error::Error>>` and `?` instead.

README.md
```diff
@@ -35,25 +35,23 @@ another library.
 ```rust
```
madsmtm, line 35: I think if you add the following to `lib.rs` (and add `ignore here`), then the example will be checked by CI, and this sort of oversight won't happen again.
```rust
#[cfg(doctest)]
#[doc = include_str!("../README.md")]
extern "C" {}
```

madsmtm (changes requested):

README.md
```diff
@@ -32,28 +32,30 @@ Winit is designed to be a low-level brick in a hierarchy of libraries. Consequen
...
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let event_loop = EventLoop::new().unwrap();
+    let window = WindowBuilder::new().build(&event_loop)?;
 
-        match event {
+    event_loop
+        .run(move |event, _elwt, cf| match event {
+            Event::RedrawRequested(_) => {
+                println!("Redraw requested");
+            }
```
kchibisov, line 50: What's the reason to revert this part? It was correct before, the `run` closure has only 2 arguments passed and it was all correct. I think the only issue is the `EventLoop::new()` missing a change to call `?`.

README.md
```diff
@@ -32,28 +32,30 @@ Winit is designed to be a low-level brick in a hierarchy of libraries. Consequen
 show something on the window you need to use the platform-specific getters provided by winit, or
 another library.
 
-```rust
+```rust, ignore
```
kchibisov, line 35: By adding `ignore` you disable the checks on the CI, so it's pointless to test with `C` trick in `lib.rs`. Remove the `ignore` and address the CI errors.

README.md
kchibisov: Hm, maybe we should remove the docs then entirely. The README could only reflect the branch it's on, otherwise it's impossible to validate the docs in the first place. We do have examples and the doc comment in the source code in lib.rs.

Tracking the example manually won't work, since it already failed. And any automated testing could be done only for example which we compile.

README.md
madsmtm: I think that's partly my fault; it should've been `no_run`

src/lib.rs
```diff
@@ -159,3 +159,7 @@ mod platform_impl;
 pub mod window;
 
 pub mod platform;
+
+#[cfg(doctest)]
+#[doc = include_str!("../README.md")]
+extern "C" {}
```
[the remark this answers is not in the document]
kchibisov: We have example in the crate level docs actually, I'm not sure we need it in the `README`.

README.md
```diff
@@ -32,28 +32,30 @@ Winit is designed to be a low-level brick in a hierarchy of libraries. Consequen
...
+    event_loop::EventLoop,
     window::WindowBuilder,
 };
 
-fn main() {
-    let event_loop = EventLoop::new();
-    let window = WindowBuilder::new().build(&event_loop).unwrap();
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let event_loop = EventLoop::new().unwrap();
+    let window = WindowBuilder::new().build(&event_loop)?;
```
kchibisov, line 45: Yeah, we should just remove this example in the README, there's no value and we have on in the `lib.rs`, so it's visible on `docs.rs` which is probably more important.

kchibisov (changes requested):

README.md
madsmtm: There has not really been consensus whether that entirely the best idea, not that I know of at least, see e.g. [this question I asked on the Rust API guidelines repo almost two years ago](https://github.com/rust-lang/api-guidelines/discussions/254).

That said, I agree, it'd also be my preference to consolidate the documentation to be on docs.rs.

kchibisov:
The README should be fairly accurate.

PR #3114, opened 2023-09-24
Android: Make DeviceId contain device id's

fornwall:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On X11, fix event loop not waking up on `ControlFlow::Poll` and `ControlFlow::WaitUntil`.
 - **Breaking:** Change default `ControlFlow` from `Poll` to `Wait`.
 - **Breaking:** remove `DeviceEvent::Text`.
+- On Android, fix `DeviceId` to contain device id's.
 
 # 0.29.1-beta
 
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -387,7 +387,7 @@ impl<T: 'static> EventLoop<T> {
         match event {
             InputEvent::MotionEvent(motion_event) => {
                 let window_id = window::WindowId(WindowId);
-                let device_id = event::DeviceId(DeviceId);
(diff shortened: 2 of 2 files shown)


src/platform_impl/android/mod.rs
```diff
@@ -739,11 +739,11 @@ impl From<u64> for WindowId {
 }
 
 #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
-pub struct DeviceId;
+pub struct DeviceId(i32);
 
 impl DeviceId {
     pub const fn dummy() -> Self {
-        DeviceId
+        DeviceId(std::i32::MIN)
```
rib, line 746: it doesn't matter much but I think we could use zero here.

We don't have to worry about whether the ID overlaps with a real device ID since this `dummy()` API is specifically documented to only be applicable for testing:

```rust
    /// The only guarantee made about the return value of this function is that
    /// it will always be equal to itself and to future values returned by this function.
    /// No other guarantees are made. This may be equal to a real `DeviceId`.
```

rib (approved):
Overall this looks good to me, thanks!

Marking it as approved from my pov and happy to leave it up to you about whether to switch the `DeviceId` to be zero

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On X11, fix event loop not waking up on `ControlFlow::Poll` and `ControlFlow::WaitUntil`.
 - **Breaking:** Change default `ControlFlow` from `Poll` to `Wait`.
 - **Breaking:** remove `DeviceEvent::Text`.
+- On Android, fix `DeviceId` to contain device id:s.
```
rib, line 30:
```suggestion
- On Android, fix `DeviceId` to contain device id's.
```

src/platform_impl/android/mod.rs
fornwall: I think `0` makes sense - have made that change in a squashed commit!

CHANGELOG.md
fornwall: Thanks! Have made that change and squashed commits now.

PR #3115, opened 2023-09-25
Fix transparet windows in X11

macmv:
It looks like 584aab4cd0da167f18e3907edfa8981885d7f92c broke transparent windows in X11 (at least on my computer). I fixed it, and updated the docs of `set_transparent` to note that transparent windows do work in X11.

  - I tested this on my computer, running X11, which is the only platform affected.
  - This is a fix of a regression, I didn't add anything to the changelog.
  - `examples/transparent.rs` still works to show this.
  - No features were changed.

### DIFF
--- a/src/platform_impl/linux/x11/window.rs
+++ b/src/platform_impl/linux/x11/window.rs
@@ -242,7 +242,7 @@ impl UnownedWindow {
                 // Find a suitable visual, true color with 32 bits of depth.
                 all_visuals
                     .find_map(|(visual, depth)| {
-                        (visual.class == xproto::VisualClass::TRUE_COLOR)
+                        (depth == 32 && visual.class == xproto::VisualClass::TRUE_COLOR)
                             .then_some((Some(visual), depth, true))
                     })
                     .unwrap_or_else(|| {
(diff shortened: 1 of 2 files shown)


src/platform_impl/linux/x11/window.rs
```diff
@@ -242,7 +242,7 @@ impl UnownedWindow {
                 // Find a suitable visual, true color with 32 bits of depth.
                 all_visuals
                     .find_map(|(visual, depth)| {
-                        (visual.class == xproto::VisualClass::TRUE_COLOR)
+                        (depth == 32 && visual.class == xproto::VisualClass::TRUE_COLOR)
```
kchibisov, line 245: @notgull why do we even search it like that and not by checking the alphamask?

notgull: "Try to find a 32 bit True Color visual" was [how it was done before](https://github.com/rust-windowing/winit/blob/ff0ce9d065431c93e1d7e5d88bb8cfb15f7f2f60/src/platform_impl/linux/x11/window.rs#L203-L230), and I just ported the logic to `x11rb`. So the real question is "why *not* check the alphamask", which we probably should be doing.

notgull (approved):
Looks good to me!

Issue #2876, opened 2023-06-12
Wrong size when resizing window on MacOS Sonoma (Beta)


zmarlon:
When I resize my window on MacOS Sonoma, I get the following window size as my new size: 4294967295x4294967295.
This does not look like valid values to me.

I'm not sure if it's a macOS or winit bug at this point, but I still thought I'd open an issue in case others encounter the same problem.

marko-lazic:
Experienced a similar issue after upgrading to the macOS Sonoma developer beta in my Bevy 0.10.1 app.

```
2023-06-20T20:48:50.569511Z  INFO bevy_winit::system: Creating new window "Bevy App" (0v0)
2023-06-20T20:48:50.598642Z  INFO bevy_diagnostic::system_information_diagnostics_plugin::internal: SystemInfo { os: "MacOS 14.0 ", kernel: "23.0.0", cpu: "Apple M1 Pro", core_count: "10", memory: "16.0 GiB" }
2023-06-20 22:48:50.615 ccc[12532:905455] CAMetalLayer ignoring invalid setDrawableSize width=4294967295.000000 height=4294967295.000000
2023-06-20T20:48:50.616179Z ERROR wgpu::backend::direct: Handling wgpu errors as fatal by default    
thread 'Compute Task Pool (1)' panicked at 'wgpu error: Validation Error

Caused by:
    In Device::create_texture
      note: label = `main_texture_a`
    Dimension X value 4294967295 exceeds the limit of 16384
```

I'm looking for a workaround at this point.

marko-lazic:
As a temporary workaround I run my app with WindowMode::BorderlessFullscreen as opposed to default Windowed.

T-Damer:
Yup @weykon workaround works for me, interesting why it can't handle windowed mode 🤔

rdinkel:
It occurred after already being on sonoma beta 2 but later updating the command line tools and xcode to Command_Line_Tools_for_Xcode_15_beta_2 and xocde beta 2

madsmtm:
It would be nice to get a `winit`+`wgpu` reproducer, a bit hard to tell if this is a Bevy-specific issue or not?

rdinkel:
I used to try out this example https://github.com/StarArawn/bevy_ecs_tilemap/blob/main/examples/mouse_to_tile.rs of this lib https://github.com/StarArawn/bevy_ecs_tilemap/tree/main

OS:  MacOS 14 beta 2 on M1
[...]

PR #3118, opened 2023-09-27
Winit version 0.28.7

madsmtm:
Backport fix for https://github.com/rust-windowing/winit/issues/2876.

CC @kchibisov


madsmtm:
Nightly CI is failing because it's picking up a newer `rustc` than was tested with back then, and MSRV CI is failing because it's using a newer `raw-window-handle`

kchibisov:
The release was discussed on the `#matrix` channel if it matters.

PR #3121, opened 2023-09-29
X11: Add `#[deny(unsafe_op_in_unsafe_fn)]`

madsmtm:
Follow-up to https://github.com/rust-windowing/winit/pull/3070, add `#[deny(unsafe_op_in_unsafe_fn)]` on the remaining platform X11, such that we can now enable it for the entirety of `winit`.

CC @notgull since this may impact your x11rb work?

### DIFF
--- a/src/platform_impl/windows/mod.rs
+++ b/src/platform_impl/windows/mod.rs
@@ -1,5 +1,4 @@
 #![cfg(windows_platform)]
-#![deny(unsafe_op_in_unsafe_fn)]
 
 use smol_str::SmolStr;
 use windows_sys::Win32::{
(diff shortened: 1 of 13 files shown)


src/platform_impl/windows/mod.rs
```diff
@@ -1,5 +1,4 @@
 #![cfg(windows_platform)]
-#![deny(unsafe_op_in_unsafe_fn)]
```
notgull, line 2: Huh?

notgull (approved):
There's some overlap but it shouldn't be too hard to rebase over.

src/platform_impl/windows/mod.rs
madsmtm: It's done globally in `src/lib.rs` now

PR #3122, opened 2023-09-30
Use x11rb for event handling

notgull:

Switches out the previous event handling logic for x11rb-based logic. The last large hurdle before getting rid of Xlib. As part of the process, also gets rid of Xlib's IME and replaces it with the `xim` crate.

Draft because I have not tested this in the slightest.


notgull:
Event handling functionality looks sound to me. I haven't tested IME yet.

notgull:
Tested IME on Fedora 38

notgull:
@riey Can you review this and make sure I'm using XIM right?

Riey:
@notgull Is there an example code to check preedit and position?

notgull:
I'm not sure what you're asking for, but here is the code responsible for handling pre-edits:

https://github.com/rust-windowing/winit/blob/46bb9e5844913542820029931f9ab1a1da636f1c/src/platform_impl/linux/x11/ime.rs#L648-L767

kchibisov:
you can also, always try to do things with `alacritty` if you want to test with _typing_ experience.

Riey:
LGTM.

src/platform_impl/linux/x11/xdisplay.rs
```diff
@@ -91,6 +91,14 @@ impl XConnection {
             conn.map_err(|e| XNotSupported::XcbConversionError(Arc::new(WrapConnectError(e))))?
         };
 
+        // Make sure Xlib knows XCB is handling events.
+        unsafe {
+            (xlib_xcb.XSetEventQueueOwner)(
+                display,
+                x11_dl::xlib_xcb::XEventQueueOwner::XCBOwnsEventQueue,
+            );
+        }
```
psychon, line 100: Since the commit message for this one says

> I don't know what this actually does, but it can't hurt, right?

"Normally", all events and errors go through Xlib. It does some gymnastics internally to ensure that before the reply for a request is handled, everything that happens before is done: Events are in its internal event queue and errors where passed to the error handler. This has the effect of removing all events/errors from XCB's queue.

Thus, normally you cannot use `xcb_poll/wait_for_event()` when Xlib is involved, since Xlib might already have "eaten" an event and put it into its own queue.

The call you are doing here just tells Xlib "do not mess with the event queue, kthxbye". It still will use the flag `XCB_REQUEST_CHECKED` flags when sending requests to that any errors caused "through" Xlib are reported to Xlib via `xcb_wait_for_reply()` and do not end up in XCB's event queue, but it will no longer call `xcb_poll_for_event()` or `xcb_wait_for_event()`.

src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -669,710 +621,685 @@ impl<T: 'static> EventProcessor<T> {
...
-            let mut ime = wt.ime.borrow_mut();
-            match request {
-                ImeRequest::Position(window_id, x, y) => {
-                    ime.send_xim_spot(window_id, x, y);
-                }
-                ImeRequest::Allow(window_id, allowed) => {
-                    ime.set_ime_allowed(window_id, allowed);
-                }
+            X11Event::RandrNotify(_) => {
+                self.process_dpi_change(&mut callback);
```
psychon, line 1222: I came here for another reason, but then noticed this bit. I think this is wrong.

The old code does:
```
if event_type == self.randr_event_offset as c_int {
```
Imagine a `+ 0` in there and check that "event zero" from randr is `RRScreenChangeNotify`: https://docs.rs/x11-dl/latest/x11_dl/xrandr/constant.RRScreenChangeNotify.html

RandrNotify is event 1: https://docs.rs/x11-dl/latest/x11_dl/xrandr/constant.RRNotify.html

psychon (changes requested):

notgull:
Closing as per chat discussion

PR #3126, opened 2023-10-01
Update to new raw-window-handle strategy

notgull:

Essentially a combination of #3075 and #2943. This PR adds explicit support for `raw-window-handle` v0.6 through the `rwh_06` feature, which is enabled by default. It also adds optional support for `raw-window-handle` v0.5 and `raw-window-handle` v0.4, which is exposed through features that are not exposed by default.

This also adds a lifetime to `WindowAttributes`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ categories = ["gui"]
 rust-version = "1.65.0"
 
 [package.metadata.docs.rs]
-features = ["serde"]
+features = ["rwh_04", "rwh_05", "rwh_06", "serde"]
 default-target = "x86_64-unknown-linux-gnu"
 # These are all tested in CI
 targets = [
@@ -35,7 +35,7 @@ targets = [
 rustdoc-args = ["--cfg", "docsrs"]
 
 [features]
-default = ["x11", "wayland", "wayland-dlopen", "wayland-csd-adwaita"]
+default = ["rwh_06", "x11", "wayland", "wayland-dlopen", "wayland-csd-adwaita"]
 x11 = ["x11-dl", "bytemuck", "percent-encoding", "xkbcommon-dl/x11", "x11rb"]
 wayland = ["wayland-client", "wayland-backend", "wayland-protocols", "wayland-protocols-plasma", "sctk", "fnv", "memmap2"]
 wayland-dlopen = ["wayland-backend/dlopen"]
@@ -55,7 +55,9 @@ cursor-icon = "1.0.0"
 log = "0.4"
 mint = { version = "0.5.6", optional = true }
 once_cell = "1.12"
-raw_window_handle = { package = "raw-window-handle", version = "0.5", features = ["std"] }
+rwh_04 = { package = "raw-window-handle", version = "0.4", optional = true }
+rwh_05 = { package = "raw-window-handle", version = "0.5", features = ["std"], optional = true }
+rwh_06 = { package = "raw-window-handle", version = "0.6", features = ["std"], optional = true }
 serde = { version = "1", optional = true, features = ["serde_derive"] }
 smol_str = "0.2.0"
 
--- a/examples/child_window.rs
+++ b/examples/child_window.rs
@@ -26,14 +26,13 @@ fn main() -> Result<(), impl std::error::Error> {
         windows: &mut HashMap<WindowId, Window>,
     ) {
         let parent = parent.window_handle().unwrap();
-        let mut builder = WindowBuilder::new()
+        let child_window = WindowBuilder::new()
             .with_title("child window")
             .with_inner_size(LogicalSize::new(200.0f32, 200.0f32))
             .with_position(Position::Logical(LogicalPosition::new(0.0, 0.0)))
-            .with_visible(true);
-        // `with_parent_window` is unsafe. Parent window must be a valid window.
-        builder = unsafe { builder.with_parent_window(Some(parent)) };
-        let child_window = builder.build(event_loop).unwrap();
+            .with_visible(true)
+            .with_parent_window(Some(parent))
+            .build(event_loop).unwrap();
 
         let id = child_window.id();
         windows.insert(id, child_window);
--- a/examples/child_window.rs
+++ b/examples/child_window.rs
@@ -1,8 +1,14 @@
-#[cfg(any(x11_platform, macos_platform, windows_platform))]
+#[cfg(all(
+    feature = "rwh_06",
+    any(x11_platform, macos_platform, windows_platform)
+))]
 #[path = "util/fill.rs"]
 mod fill;
 
-#[cfg(any(x11_platform, macos_platform, windows_platform))]
+#[cfg(all(
+    feature = "rwh_06",
+    any(x11_platform, macos_platform, windows_platform)
+))]
 fn main() -> Result<(), impl std::error::Error> {
     use std::collections::HashMap;
 
@@ -10,7 +16,7 @@ fn main() -> Result<(), impl std::error::Error> {
         dpi::{LogicalPosition, LogicalSize, Position},
         event::{ElementState, Event, KeyEvent, WindowEvent},
         event_loop::{EventLoop, EventLoopWindowTarget},
-        window::raw_window_handle::HasRawWindowHandle,
+        raw_window_handle::HasWindowHandle,
         window::{Window, WindowBuilder, WindowId},
     };
 
@@ -19,7 +25,7 @@ fn main() -> Result<(), impl std::error::Error> {
         event_loop: &EventLoopWindowTarget<()>,
         windows: &mut HashMap<WindowId, Window>,
     ) {
-        let parent = parent.raw_window_handle();
+        let parent = parent.window_handle().unwrap();
         let mut builder = WindowBuilder::new()
             .with_title("child window")
             .with_inner_size(LogicalSize::new(200.0f32, 200.0f32))
@@ -81,7 +87,10 @@ fn main() -> Result<(), impl std::error::Error> {
     })
 }
 
-#[cfg(not(any(x11_platform, macos_platform, windows_platform)))]
+#[cfg(not(all(
+    feature = "rwh_06",
+    any(x11_platform, macos_platform, windows_platform)
+)))]
 fn main() {
     panic!("This example is supported only on x11, macOS, and Windows.");
 }
--- a/examples/child_window.rs
+++ b/examples/child_window.rs
@@ -9,6 +9,7 @@ mod fill;
     feature = "rwh_06",
     any(x11_platform, macos_platform, windows_platform)
 ))]
+#[allow(deprecated)]
 fn main() -> Result<(), impl std::error::Error> {
     use std::collections::HashMap;
 
@@ -16,7 +17,7 @@ fn main() -> Result<(), impl std::error::Error> {
         dpi::{LogicalPosition, LogicalSize, Position},
         event::{ElementState, Event, KeyEvent, WindowEvent},
         event_loop::{EventLoop, EventLoopWindowTarget},
-        raw_window_handle::HasWindowHandle,
+        raw_window_handle::HasRawWindowHandle,
         window::{Window, WindowBuilder, WindowId},
     };
 
@@ -25,15 +26,15 @@ fn main() -> Result<(), impl std::error::Error> {
         event_loop: &EventLoopWindowTarget<()>,
         windows: &mut HashMap<WindowId, Window>,
     ) {
-        let parent = parent.window_handle().unwrap();
-        let child_window = WindowBuilder::new()
+        let parent = parent.raw_window_handle().unwrap();
+        let mut builder = WindowBuilder::new()
             .with_title("child window")
             .with_inner_size(LogicalSize::new(200.0f32, 200.0f32))
             .with_position(Position::Logical(LogicalPosition::new(0.0, 0.0)))
-            .with_visible(true)
-            .with_parent_window(Some(parent))
-            .build(event_loop)
-            .unwrap();
+            .with_visible(true);
+        // `with_parent_window` is unsafe. Parent window must be a valid window.
+        builder = unsafe { builder.with_parent_window(Some(parent)) };
+        let child_window = builder.build(event_loop).unwrap();
 
         let id = child_window.id();
         windows.insert(id, child_window);
@@ -92,5 +93,5 @@ fn main() -> Result<(), impl std::error::Error> {
     any(x11_platform, macos_platform, windows_platform)
 )))]
 fn main() {
-    panic!("This example is supported only on x11, macOS, and Windows.");
+    panic!("This example is supported only on x11, macOS, and Windows, with the `rwh_06` feature enabled.");
 }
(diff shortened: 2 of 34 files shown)


daxpedda (approved):
Web is LGTM.

src/lib.rs
```diff
@@ -138,6 +138,16 @@
...
 
+#[cfg(all(not(feature = "rwh_06"), not(feature = "rwh_05"), feature = "rwh_04"))]
+#[doc(inline)]
+pub use rwh_04 as raw_window_handle;
+#[cfg(all(not(feature = "rwh_06"), feature = "rwh_05"))]
+#[doc(inline)]
+pub use rwh_05 as raw_window_handle;
+#[cfg(feature = "rwh_06")]
+#[doc(inline)]
+pub use rwh_06 as raw_window_handle;
```
madsmtm, line 149: I'm not sure this is a good idea wrt. stability, e.g. a dependency using `rwh_05` would be very surprised to see this change under them when someone enables `rwh_06`. I'd rather only expose this when `rwh_06` is enabled:
```suggestion
#[cfg(feature = "rwh_06")]
pub use rwh_06 as raw_window_handle;
```

examples/child_window.rs
```diff
@@ -81,7 +87,10 @@ fn main() -> Result<(), impl std::error::Error> {
     })
 }
 
-#[cfg(not(any(x11_platform, macos_platform, windows_platform)))]
+#[cfg(not(all(
+    feature = "rwh_06",
+    any(x11_platform, macos_platform, windows_platform)
+)))]
 fn main() {
     panic!("This example is supported only on x11, macOS, and Windows.");
```
madsmtm, line 95: Nit:

```suggestion
    panic!("This example is supported only on x11, macOS, and Windows, with the `rwh_06` feature enabled.");
```

src/platform/windows.rs
```diff
@@ -161,9 +161,10 @@ impl WindowExtWindows for Window {
 }
 
 /// Additional methods on `WindowBuilder` that are specific to Windows.
+#[allow(rustdoc::broken_intra_doc_links)]
 pub trait WindowBuilderExtWindows {
     /// Set an owner to the window to be created. Can be used to create a dialog box, for example.
-    /// This only works when [`WindowBuilder::with_parent_window`] isn't called or set to `None`.
+    /// This only works when [`Self::with_parent_window`] isn't called or set to `None`.
```
madsmtm, line 167: Nit: The documentation reads nicer with the former

src/platform_impl/macos/window.rs
```diff
@@ -452,29 +448,28 @@ impl WinitWindow {
...
+                                    "raw window handle should be non-empty"
+                                )))
+                            }
+                        };
+                    parent_view.window().ok_or_else(|| {
+                        os_error!(OsError::CreationError(
+                            "parent view should be installed in a window"
+                        ))
+                    })?
                 };
```
madsmtm, line 471: Nit: Could be simplified to something like this instead:
```suggestion
                // SAFETY: The safety contract of WindowHandle<'_> ensures the pointer is valid
                // Unwrap is fine, since the pointer comes from `NonNull`.
                let parent_view: Id<NSView> = unsafe { Id::retain(handle.ns_view.as_ptr().cast()) }.unwrap();
                let parent = parent_view.window().ok_or_else(|| {
                    os_error!(OsError::CreationError(
                        "parent view should be installed in a window"
                    ))
                })?;
```

src/platform_impl/macos/window.rs
```diff
@@ -1347,17 +1342,49 @@ impl WinitWindow {
...
+    }
+
+    #[cfg(feature = "rwh_06")]
+    #[inline]
+    pub fn raw_window_handle_rwh_06(&self) -> Result<rwh_06::RawWindowHandle, rwh_06::HandleError> {
+        let window_handle = rwh_06::AppKitWindowHandle::new(unsafe {
+            // SAFETY: ns_view is never null.
+            let ptr = Id::as_ptr(&self.contentView()) as *mut _;
+            std::ptr::NonNull::new_unchecked(ptr)
+        });
```
madsmtm, line 1376: Nit:
```suggestion
        let window_handle = rwh_06::AppKitWindowHandle::new({
            let ptr = Id::as_ptr(&self.contentView()) as *mut _;
            // SAFETY: `Id<T>` is never null.
            unsafe { std::ptr::NonNull::new_unchecked(ptr) }
        });
```

madsmtm (changes requested):
Thanks for following up on this @notgull!

macOS and iOS impl approved, have opened some discussion on the API though that I would like to see resolved.

src/window.rs
```diff
@@ -157,13 +151,15 @@ pub struct WindowAttributes {
     pub resize_increments: Option<Size>,
     pub content_protected: bool,
     pub window_level: WindowLevel,
-    pub parent_window: Option<RawWindowHandle>,
+    #[cfg(feature = "rwh_06")]
+    pub parent_window: Option<rwh_06::WindowHandle<'a>>,
```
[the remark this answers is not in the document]
kchibisov: While I don't have time to review this yet, I want to comment on this particular matter. We basically need to account for the use case where you send a `WindowBuilder` across the threads.

However our current builder also can't be `Send`. Maybe we should somehow separate the `!Send` part out of the structure? This will also move the lifetime to some other struct.

We need to avoid the lifetime on the builder, but we also need to avoid `rwhnd` and `!Send` all together. There was an issue about that somewhere, but I don't remember where...

Cargo.toml
```diff
@@ -13,7 +13,7 @@ categories = ["gui"]
 rust-version = "1.65.0"
 
 [package.metadata.docs.rs]
-features = ["serde"]
+features = ["rwh_04", "rwh_05", "rwh_06", "serde"]
```
kchibisov, line 16: Could we drop `04` while at it?

examples/util/fill.rs
```diff
@@ -7,9 +7,11 @@
 //! The `softbuffer` crate is used, largely because of its ease of use. `glutin` or `wgpu` could
 //! also be used to fill the window buffer, but they are more complicated to use.
 
+// TODO: Once softbuffer uses rwh_06, use that instead.
+
```
kchibisov, line 11: I don't think that we really need that TODO here.

kchibisov (changes requested):
Could you `expect` instead of doing `unchecked`? I'd rather not have more unsafe and have an actual assert that it got a `Null` for whatever reason.

It'll basically never happen, but the cost of checking `null` in a function you rarely call is non-existent.

src/window.rs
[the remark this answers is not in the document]
madsmtm: On macOS (and iOS), you could create a `Send`/`Sync` handle with something like:
```rust
struct OwnedWindowHandle {
    view: MainThreadBound<Id<NSView>>,
}

impl OwnedWindowHandle {
    fn from_handle(handle: rwh_06::WindowHandle<'_>) -> Self {
        let view = match handle.raw_window_handle() {
            // SAFETY: The safety contract of WindowHandle<'_> ensures the pointer is valid.
            //
            // `NSView` may only be used on the main thread, but that is ensured below
            // by putting it in a MainThreadBound.
            RawWindowHandle::AppKit(handle) => unsafe { Id::retain(handle.ns_view.as_ptr().cast()) }.unwrap();
            handle => panic!("Invalid window handle {handle:?} on macOS"),
        };
        // SAFETY: NSView is safe to Drop on the main thread.
        let mtm = unsafe { MainThreadMarker::new_unchecked() };
        let view = MainThreadBound::new(view, mtm);
        Self { view }
    }
}
```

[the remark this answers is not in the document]
kchibisov: No, I mean, could we simply split the `WindowBuilder` in a way that once you pass the `with_parent` it becomes not `Send`, but by default it's `Send`? I don't want a send/sync handle, I just want most parts of the builder being send and have a way to make it into the non-send version?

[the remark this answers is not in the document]
kchibisov: Like
```
let builder = WindowBuilder::new().with_title("Foo"); // This you can send.
let not_send = builder.with_parent(Some(parent)); // This you can't send anymore.
```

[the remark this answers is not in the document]
kchibisov: Well, it simply means that you mark rwhnd as safe to send + sync, while the real one is not. It's true that Wayland one is send + sync, but from a users point of view what I suggest is zero changes. it's just our internals will have a bit more code, but nothing really bad.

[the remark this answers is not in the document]
kchibisov: And RawWindowHandle is just one example, you can literally add anything, like raw X11 stuff to build a window, some random pointers, etc. You can solve the issue for one type of things, but not for other.

Also, the reason I suggested, is because there's a desire to send window builders between threads, but finish building on other threads.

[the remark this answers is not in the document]
madsmtm: By the complication in API surface, I mean that your suggestion means we must introduce a new `ThreadUnsafeWindowBuilder`, which will be confusing for users for little benefit in my opinion.

> raw X11 stuff to build a window, some random pointers

Seems like we should make those thread safe as well, then.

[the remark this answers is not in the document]
kchibisov: Maybe you haven't done that, but `WindowBuilder<T = ()>` and then `with_parent(self) -> WindowBuilder<T = NotSendUdata>` will work just as fine.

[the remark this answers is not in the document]
kchibisov: I can do this all myself in a separate patch though.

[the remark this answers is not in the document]
madsmtm: That'd work, but still be a complication of the API surface for a benefit we should be able to get another way.

(I was also sorta hoping that we could transition `WindowBuilder` to using `&mut self` at some point, in which case your proposed solution would be harder to work with).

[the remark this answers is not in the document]
madsmtm: > I can do this all myself in a separate patch though.

Yeah, in any case, let's keep this focused on the now added lifetime parameter.

@notgull: Maybe we should wait on changing that for this PR, and just keep the `RawWindowHandle` in there? That'd mean `with_parent_window` is still `unsafe`, but that's just the status quo, so not that bad.

Cargo.toml
notgull: `raw-window-handle` v0.4 is still in use by some users, so I think we should keep it for now.

src/window.rs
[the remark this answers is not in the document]
notgull: I've removed this change for now.

madsmtm (approved):

TornaxO7:
@notgull maybe not needed, but just in case, I fixed the CI build in [this PR](https://github.com/rust-windowing/winit/pull/3129)

src/window.rs
```diff
@@ -1463,41 +1512,71 @@ impl Window {
...
     }
 }
 
-unsafe impl HasRawDisplayHandle for Window {
+#[cfg(feature = "rwh_05")]
+unsafe impl rwh_05::HasRawDisplayHandle for Window {
     /// Returns a [`raw_window_handle::RawDisplayHandle`] used by the [`EventLoop`] that
     /// created a window.
     ///
     /// [`EventLoop`]: crate::event_loop::EventLoop
```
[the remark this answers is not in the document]
TornaxO7: you mean, that this should also be added to the implementation of `HasDisplayHandle` for `v0.6`?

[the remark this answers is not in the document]
madsmtm: Yup

kchibisov:
@TornaxO7 comments are getting resolved over time. This particular one was removed. See https://github.com/rust-windowing/winit/pull/3126/commits/9e15334c7ca3b51f08e31c2063cf463c3bb26e11

src/platform_impl/android/mod.rs
```diff
@@ -938,16 +947,51 @@ impl Window {
...
+    pub fn raw_display_handle_rwh_05(&self) -> rwh_05::RawDisplayHandle {
+        rwh_05::RawDisplayHandle::Android(rwh_05::AndroidDisplayHandle::empty())
+    }
+
+    #[cfg(feature = "rwh_06")]
+    pub fn raw_window_handle_rwh_06(&self) -> Result<rwh_06::RawWindowHandle, rwh_06::HandleError> {
+        if let Some(native_window) = self.app.native_window().as_ref() {
+            let handle = rwh_06::AndroidNdkWindowHandle::new(native_window.ptr().cast());
+            Ok(rwh_06::RawWindowHandle::AndroidNdk(handle))
+        } else {
```
MarijnS95, line 982: I'm implementing this in the NDK, as part of version bumps.

MarijnS95 (changes requested):

src/platform_impl/android/mod.rs
```diff
@@ -938,16 +947,51 @@ impl Window {
...
 
-    pub fn raw_display_handle(&self) -> RawDisplayHandle {
-        RawDisplayHandle::Android(AndroidDisplayHandle::empty())
+    #[cfg(feature = "rwh_05")]
+    pub fn raw_display_handle_rwh_05(&self) -> rwh_05::RawDisplayHandle {
+        rwh_05::RawDisplayHandle::Android(rwh_05::AndroidDisplayHandle::empty())
+    }
+
+    #[cfg(feature = "rwh_06")]
+    pub fn raw_window_handle_rwh_06(&self) -> Result<rwh_06::RawWindowHandle, rwh_06::HandleError> {
```
MarijnS95, line 978: Question while implementing the `ndk` portion for this: implementing the `Has**RAW**WindowHandle` trait is deprecated and there are blanket implementations when `HasWindowHandle` is implemented. Should this function also return a non-raw handle (in terms of that implementation when it is released)?

src/platform_impl/android/mod.rs
MarijnS95: Done in https://github.com/rust-mobile/ndk/pull/434.

src/platform_impl/android/mod.rs
notgull: I considered this, but it would add an additional `unsafe { WIndowHandle::borrow_raw }` call for each backend. I could go either way.

MarijnS95: Right. At least for Android I'm willing to have that `unsafe { borrow_raw() }` inside the `ndk` with appropriate `// SAFETY` docs, so you could call through to that. But it'd have to be consistent across all the backends...

src/platform_impl/android/mod.rs
TornaxO7: May I ask what that means? Does anything has to do be done here or can this be marked as resolved?

MarijnS95: This means that once I publish `ndk 0.8.0`, `winit` should upgrade to it (it's on `-beta.0` now) and reuse the traits provided there.

notgull: Does this block this PR? I think that this should be merged in its current state and then we can upgrade to the latest `ndk` later.

MarijnS95: It does not and I can patch this when submitting the final `ndk` bump inside `winit`.

src/platform_impl/android/mod.rs
MarijnS95: Did you consider whether or not the safe borrow handles should be implemented by `winit`?

notgull: For the time being it leads to a lot more `unsafe` throughout the codebase. So I think that we shouldn't for now.

MarijnS95 (approved):
Approving the android changes, but still have a question about borrowed handles.

kchibisov (approved):

PR #3130, opened 2023-10-04
support pinch, double tap and rotation gestures on iOS

mockersf:

Support existing macOS gestures on iOS

- renamed `TouchpadMagnify` to `PinchGesture`, `SmartMagnify` to `DoubleTapGesture` and `TouchpadRotate` to `RotationGesture`. Those were supported only on macOS, when adding iOS support it feels better to have more generic names
- gesture recognition is not active by default on iOS, and is enabled gesture by gesture by the user. this is because enabling gesture recognition may block some touch events while the OS is trying to recognise a gesture, and may not always be what the user want. it's also possible to disable gesture recognition

`DoubleTapGesture` is a discrete gesture (one event once detection is finished) that can be configured. `PinchGesture` and `RotationGesture` are continuous gestures (events for each incremental update) with data.


alice-i-cecile:
I really like the renames: those are much clearer, especially in the context of things like games. Gestures being off-by-default makes sense, but that will need to be aggressively documented.

The continuous / discrete distinction is useful; could that be included in module-level docs somewhere?

src/event.rs
```diff
@@ -477,8 +477,8 @@ pub enum WindowEvent {
     ///
     /// ## Platform-specific
     ///
-    /// - Only available on **macOS 10.8** and later.
-    SmartMagnify { device_id: DeviceId },
+    /// - Only available on **macOS 10.8** and later, and **iOS**.
+    DoubleTapGesture { device_id: DeviceId },
 
     /// Touchpad rotation event with two-finger rotation gesture.
```
madsmtm, line 483: These docs also need updating;

src/event.rs
```diff
@@ -477,8 +477,8 @@ pub enum WindowEvent {
     ///
     /// ## Platform-specific
     ///
-    /// - Only available on **macOS 10.8** and later.
-    SmartMagnify { device_id: DeviceId },
+    /// - Only available on **macOS 10.8** and later, and **iOS**.
+    DoubleTapGesture { device_id: DeviceId },
```
madsmtm, line 481: Docs on this also needs to be updated.

src/event.rs
```diff
@@ -447,15 +447,15 @@ pub enum WindowEvent {
         button: MouseButton,
     },
 
-    /// Touchpad magnification event with two-finger pinch gesture.
+    /// Two-finger pinch gesture, often use for magnification.
```
madsmtm, line 450: Nit:
```suggestion
    /// Two-finger pinch gesture, often used for magnification.
```

examples/touchpad_gestures.rs
```diff
@@ -17,23 +17,23 @@ fn main() -> Result<(), impl std::error::Error> {
         .build(&event_loop)
         .unwrap();
 
-    println!("Only supported on macOS at the moment.");
+    println!("Only supported on macOS/iOS at the moment.");
```
madsmtm, line 20: This example should enable the events.

src/platform/ios.rs
```diff
@@ -85,6 +85,21 @@ pub trait WindowExtIOS {
...
+
+    /// Sets whether the [`Window`] should recognize double tap gestures.
+    ///
+    /// The default is to not recognize gestures.
+    fn recognize_doubletap_gesture(&self, should_recognize: bool);
+
+    /// Sets whether the [`Window`] should recognize rotation gestures.
+    ///
+    /// The default is to not recognize gestures.
+    fn recognize_rotation_gesture(&self, should_recognize: bool);
```
madsmtm, line 102: That the events being off by default should be documented in the `WindowEvent` enum.

src/platform_impl/ios/uikit/gesture_recognizer.rs
```diff
@@ -0,0 +1,141 @@
+use crate::platform_impl::platform::view::WinitView;
```
madsmtm, line 1: Please keep this out of this module, use `super::UIView` instead.

src/platform_impl/ios/uikit/gesture_recognizer.rs
```diff
@@ -0,0 +1,141 @@
...
+
+unsafe impl Encode for UIGestureRecognizer {
+    const ENCODING: Encoding = Encoding::Object;
+}
+
+// https://developer.apple.com/documentation/uikit/uigesturerecognizer/state
+#[derive(Debug)]
+#[allow(dead_code)]
+#[repr(isize)]
+pub enum UIGestureRecognizerState {
```
madsmtm, line 37: Nit: Use a newtype that wraps `NSInteger`, similar to how it's done with `UIScreenOverscanCompensation` in `uikit/screen.rs`

src/platform_impl/ios/uikit/gesture_recognizer.rs
```diff
@@ -0,0 +1,141 @@
...
+    }
+);
+
+extern_methods!(
+    unsafe impl UIGestureRecognizer {
+        #[method(state)]
+        pub fn state(&self) -> UIGestureRecognizerState;
+
+        #[method(initWithTarget:action:)]
+        pub fn initWithTarget(&self, target: WinitView, action: Sel);
```
madsmtm, line 25: Unsafe because the selector must match the correct method.

```suggestion
        pub unsafe fn initWithTarget_action(&self, target: &AnyObject, action: Sel);
```

Same goes for similar methods on the subclasses.

src/event.rs
```diff
@@ -447,15 +447,15 @@ pub enum WindowEvent {
...
     ///
     /// Positive delta values indicate magnification (zooming in) and
     /// negative delta values indicate shrinking (zooming out).
     ///
     /// ## Platform-specific
     ///
-    /// - Only available on **macOS**.
-    TouchpadMagnify {
+    /// - Only available on **macOS** and **iOS**.
+    PinchGesture {
```
jpedrick, line 458: Would it make sense to add this?
```
#[deprecated(note="use")]
type TypeAlias TouchpadMagnify = PinchGesture;
```

Then macos targets can still compile with warnings?

src/platform_impl/ios/view.rs
```diff
@@ -159,6 +169,86 @@ declare_class!(
...
+            let mtm = MainThreadMarker::new().unwrap();
+            app_state::handle_nonuser_event(mtm, gesture_event);
+        }
+
+        #[method(doubleTapGesture:)]
+        fn double_tap_gesture(&self, recognizer: &UITapGestureRecognizer) {
+            let window = self.window().unwrap();
+            let uiscreen = window.screen();
+
+            if matches!(recognizer.state(), UIGestureRecognizerState::Ended) {
```
madsmtm, line 208: Nit: Just compare for equality.

```suggestion
            if recognizer.state() == UIGestureRecognizerState::Ended {
```

src/platform/ios.rs
madsmtm: Also, do you have a use-case for controlling this after window creation time? If not, perhaps these methods should be on `WindowBuilderExtIOS` instead?

src/platform_impl/ios/view.rs
```diff
@@ -197,6 +296,43 @@ impl WinitView {
...
+        }
+    }
+
+    pub(crate) fn recognize_doubletap_gesture(&self, should_recognize: bool) {
+        if should_recognize && self.ivars().doubletap_gesture_recognizer.borrow().is_none() {
+            let tap = UITapGestureRecognizer::init_with_target(self, sel!(doubleTapGesture:));
+            unsafe {
+                let _: () = msg_send![&tap, setNumberOfTapsRequired: 2_u64];
+                let _: () = msg_send![&tap, setNumberOfTouchesRequired: 1_u64];
+            }
```
madsmtm, line 315: Use the methods you've defined in the `uikit` module:
```suggestion
            tap.setNumberOfTapsRequired(2);
            tap.setNumberOfTouchesRequired(1);
```

src/platform_impl/ios/view.rs
```diff
@@ -197,6 +296,43 @@ impl WinitView {
...
+            self.addGestureRecognizer(tap.as_super());
```
madsmtm, line 316: Nit: Usually, you'll want to avoid using `as_super`; let `Deref` do the work instead:
```suggestion
            self.addGestureRecognizer(&tap);
```

src/platform_impl/ios/view.rs
```diff
@@ -197,6 +296,43 @@ impl WinitView {
...
     }
 
+    pub(crate) fn recognize_pinch_gesture(&self, should_recognize: bool) {
+        if should_recognize && self.ivars().pinch_gesture_recognizer.borrow().is_none() {
+            let pinch = UIPinchGestureRecognizer::init_with_target(self, sel!(pinchGesture:));
+            self.addGestureRecognizer(pinch.as_super());
+            self.ivars().pinch_gesture_recognizer.replace(Some(pinch));
+        } else if let Some(recognizer) = self.ivars().pinch_gesture_recognizer.take() {
+            self.removeGestureRecognizer(&recognizer);
+        }
```
madsmtm, line 306: I think the logic here is wrong? If I have `should_recognize = true` and a recognizer is already added, then this will remove it.

src/event.rs
mockersf: that's not possible on enum variants

madsmtm (changes requested):
Thanks @mockersf, and apologies for not getting to this sooner! I don't know what you could've done better to push me, the continous updating of the branch worked pretty well to keep it on my radar, I've just been a bit hesitant to review iOS additions (since I'm less familiar with those than macOS).

kchibisov:
Maybe we should unify the naming and the meaning of the gestures an probably accumulate them under the same enum? Also, touch screen gestures are different to touchpad ones, usually, but not that much.

For example on Wayland I have swipe, hold, and pinch gestures https://wayland.app/protocols/pointer-gestures-unstable-v1

kchibisov:
I'm also not sure how we should handle the touchpad gestures, since it could be weird to have them on ios, but not on desktop platforms, since touchpad gesture recognition is usually done based on the raw events (unlike touchpad gestures) in the clients, thus winit should be able to map them, but I understand that you generally want to run iOS builtin gesture recognition stuff. Maybe winit should handle touchpad gestures with the help of some lib on linux, but it's a question for the future.

Also, I'd suggest to straight rename events without adding aliases, etc, and probably rename to follow the terminology used by wayland since it's more generic and you already did that for pinch.

What do you also think about having the device type attached along these events, so you can identified whether it was touchpad or touchscreen or you generally don't have such information on ios/macOS available?

madsmtm:
> What do you also think about having the device type attached along these events, so you can identified whether it was touchpad or touchscreen or you generally don't have such information on ios/macOS available?

The `DeviceId` is passed, although for some reason that contains `UIScreen` on iOS? That seems wrong...

In general, I think macOS/iOS tries to hide away which device caused a specific event, at least I can't seem to find any info in the docs on how you'd determine it.

madsmtm:
I took a closer look, I think it could be supported on macOS with https://developer.apple.com/documentation/appkit/nsevent/1530014-deviceid?language=objc, but there is no such thing in UIKit.

kchibisov:
`DeviceId` is quite useless on its own, I was thinking about

```rust
enum GestureSource {
    TouchPad,
    TouchScreen,
}
```

But generally, I'm not sure that these things differ much...

src/platform_impl/ios/view.rs
```diff
@@ -159,6 +169,86 @@ declare_class!(
...
+                UIGestureRecognizerState::Cancelled | UIGestureRecognizerState::Failed => {
+                    TouchPhase::Cancelled
+                }
+                state => panic!("unexpected recognizer state: {:?}", state),
+            };
+
+            let gesture_event = EventWrapper::StaticEvent(Event::WindowEvent {
+                window_id: RootWindowId(window.id()),
+                event: WindowEvent::PinchGesture {
+                    device_id: RootDeviceId(DeviceId {
```
madsmtm, line 191: Apologies for the churn, but you'll also have to update this after https://github.com/rust-windowing/winit/pull/3402.

src/platform/ios.rs
mockersf: > do you have a use-case for controlling this after window creation time?

I think so, enabling gesture recognition delays basic touch a bit, that seemed not desirable when it's not needed. Coming from Bevy and games, I probably don't need gestures on my menu.

src/platform_impl/ios/uikit/gesture_recognizer.rs
mockersf: marked it as unsafe, but not sure what do you mean by that, it's the only one in all of the iOS impl

mockersf:
> Also, I'd suggest to straight rename events without adding aliases, etc, and probably rename to follow the terminology used by wayland since it's more generic and you already did that for pinch.

Wayland doesn't seem to have equivalent for rotation and double tap, so that's it. It has hold and swipe, which are called long press and swipe on iOS, I think "long press" is clearer than "hold" but both work for me.

kchibisov:
I've pointed wayland gestures in the context of existing macOS gestures since you've renamed a bunch. These are for touchpad, not the touchscreen, touchscreen gestures are client side, so it's irrelevant.

src/platform_impl/ios/uikit/gesture_recognizer.rs
```diff
@@ -0,0 +1,147 @@
...
+
+unsafe impl Encode for UIPinchGestureRecognizer {
+    const ENCODING: Encoding = Encoding::Object;
+}
+
+impl UIPinchGestureRecognizer {
+    pub(crate) fn init_with_target(target: &UIView, action: Sel) -> Id<Self> {
+        unsafe { msg_send_id![Self::alloc(), initWithTarget: target, action: action] }
+    }
+}
```
madsmtm, line 85: I'd prefer it like this, as it'll make it easier for me to replace with `icrate`.

```suggestion
extern_methods!(
    unsafe impl UIPinchGestureRecognizer {
        #[method(scale)]
        pub fn scale(&self) -> CGFloat;

        #[method(velocity)]
        pub fn velocity(&self) -> CGFloat;
        
        #[method_id(initWithTarget:action:)]
        pub unsafe fn initWithTarget_action(&self, target: &AnyObject, action: Sel) -> Id<Self>;
    }
);

unsafe impl Encode for UIPinchGestureRecognizer {
    const ENCODING: Encoding = Encoding::Object;
}
```

Same goes for the others.

src/platform_impl/ios/view.rs
```diff
@@ -179,14 +260,23 @@ extern_methods!(
     }
 );
 
+unsafe impl Encode for WinitView {
+    const ENCODING: Encoding = Encoding::Object;
+}
```
madsmtm, line 265: This is wrong, and should be removed.

```suggestion
```

src/platform_impl/ios/view.rs
```diff
@@ -197,6 +287,43 @@ impl WinitView {
...
+            tap.setNumberOfTouchesRequired(1);
+            self.addGestureRecognizer(&tap);
+            self.ivars().doubletap_gesture_recognizer.replace(Some(tap));
+        } else if let Some(recognizer) = self.ivars().doubletap_gesture_recognizer.take() {
+            self.removeGestureRecognizer(&recognizer);
+        }
+    }
+
+    pub(crate) fn recognize_rotation_gesture(&self, should_recognize: bool) {
+        if should_recognize && self.ivars().rotation_gesture_recognizer.borrow().is_none() {
```
madsmtm, line 315: Logic also wrong here

src/platform_impl/ios/uikit/gesture_recognizer.rs
```diff
@@ -0,0 +1,147 @@
...
+    }
+);
+
+extern_methods!(
+    unsafe impl UIGestureRecognizer {
+        #[method(state)]
+        pub fn state(&self) -> UIGestureRecognizerState;
+
+        #[method(initWithTarget:action:)]
+        pub unsafe fn initWithTarget_action(&self, target: &AnyObject, action: Sel);
```
madsmtm, line 29: Nit: This is unused, and could be removed.

madsmtm:
Tested this on my iPad, and fixed a few bugs in the process. The absolute scale for the pinch gesture doesn't match between macOS and iOS, but I'm not sure how to make them, so let's punt on that for now.

kchibisov:
> The absolute scale for the pinch gesture doesn't match between macOS and iOS,

Could you elaborate on what you mean here? Can't we translate into physical sizes of some sort or it's just something else like values are just different and you can't really translate anything?

madsmtm (approved):

madsmtm:
> Could you elaborate on what you mean here? Can't we translate into physical sizes of some sort or it's just something else like values are just different and you can't really translate anything?

Oops, merged before I saw your message.

I'm not entirely sure, I guess it's kinda like the `MouseScrollDelta` event, there isn't really one scale that "makes sense", it kinda depends on what you're doing.

kchibisov:
Delta stuff depends, since there's one discrete and there's one pixel delta, if it's pixel it's scaled, if it's discrete it's not scaled and passed as is, but we have separate types to encode such situation.

mockersf:
thanks for testing on your device and fixing those bugs!

madsmtm:
> Delta stuff depends, since there's one discrete and there's one pixel delta, if it's pixel it's scaled, if it's discrete it's not scaled and passed as is, but we have separate types to encode such situation.

I guess the most correct thing would be to have an enum like `enum Pinch { OnScreen(f32), TouchPad(f32) }`, with the `OnScreen` value corresponding to pixels on the screen, and `TouchPad` being arbitrary values; but eh, people are probably gonna have to do something platform-specific here anyhow.

@mockersf when you do get to supporting this in Bevy, feel free to reach out if the API is inconvenient, or you find a way to represent the data correctly, then we can change it in Winit.

PR #3136, opened 2023-10-10
Make default `WindowBuilder` `Send` and `Sync`

kchibisov:
The default window builder is neither `Send` nor `Sync` due to use of `RawWindowHandle` internally. While we could ensure to make the handle `Send + Sync` within our use, we can't guarantee that the same will hold true in the future.

Thus add a generic to split the builder into `MT-safe` and not `MT-safe` versions.

This could also help with the lifetimes, since default could be bound by static lifetime.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -163,3 +163,18 @@ mod platform_impl;
 pub mod window;
 
 pub mod platform;
+
+/// Wrapper for objects which winit will access on the main thread so they are effectively `Send`
+/// and `Sync`, since they always excute on a single thread.
+///
+/// # Safety
+///
+/// Winit can run only one event loop at the time and the event loop itself is tied to some thread.
+/// The objects could be send across the threads, but once passed to winit, they execute on the
+/// mean thread if the platform demands it. Thus marking such objects as `Send + Sync` is safe.
+#[doc(hidden)]
+#[derive(Clone, Debug)]
+pub(crate) struct SendSyncWrapper<T>(pub(crate) T);
+
+unsafe impl<T> Send for SendSyncWrapper<T> {}
+unsafe impl<T> Sync for SendSyncWrapper<T> {}
--- a/src/window.rs
+++ b/src/window.rs
@@ -6,7 +6,7 @@ use crate::{
     error::{ExternalError, NotSupportedError, OsError},
     event_loop::EventLoopWindowTarget,
     monitor::{MonitorHandle, VideoMode},
-    platform_impl,
+    platform_impl, SendSyncWrapper,
 };
 
 pub use crate::icon::{BadIcon, Icon};
@@ -141,7 +141,6 @@ pub struct WindowAttributes {
     pub resizable: bool,
     pub enabled_buttons: WindowButtons,
     pub title: String,
-    pub fullscreen: Option<Fullscreen>,
     pub maximized: bool,
     pub visible: bool,
     pub transparent: bool,
@@ -152,9 +151,10 @@ pub struct WindowAttributes {
     pub resize_increments: Option<Size>,
     pub content_protected: bool,
     pub window_level: WindowLevel,
-    #[cfg(feature = "rwh_06")]
-    pub parent_window: Option<rwh_06::RawWindowHandle>,
     pub active: bool,
+    #[cfg(feature = "rwh_06")]
+    pub(crate) parent_window: SendSyncWrapper<Option<rwh_06::RawWindowHandle>>,
+    pub(crate) fullscreen: SendSyncWrapper<Option<Fullscreen>>,
 }
 
 impl Default for WindowAttributes {
@@ -169,7 +169,7 @@ impl Default for WindowAttributes {
             enabled_buttons: WindowButtons::all(),
             title: "winit window".to_owned(),
             maximized: false,
-            fullscreen: None,
+            fullscreen: SendSyncWrapper(None),
             visible: true,
             transparent: false,
             blur: false,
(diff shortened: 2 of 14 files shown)


madsmtm (changes requested):
I disagree that this is the right way to solve this issue, window handles _are_ thread safe on macOS/iOS, since we guarantee to never use them outside the main thread.

Rather, we should wrap the window handle in something that implements `Send + Sync` (or maybe we should figure it out in https://github.com/rust-windowing/raw-window-handle/issues/85 instead).

kchibisov:
@madsmtm the issue is not only a window handle as it was said. You can see the CI logs and discover, that all the Web can't be Send, what should we do about it?

kchibisov:
Like, surely, we could keep the `WindowAttributes` as `Send` + `Sync`, but I don't think we can easily guarantee the same about the platform stuff. Also, it's not clear how we should deal with the lifetime, but I don't think it really matters as long as we can convert to `'static' bound?

madsmtm:
I believe the solution is the same on the web: Since the window handle can only be created and accessed from the main thread, then it doesn't matter if the handle was sent between threads (or service workers) in between those accesses. @daxpedda?

> As well as monitors on macOS.

I think it's the same thing here, monitors can be `Send + Sync` on macOS if I just refactor them a bit.

daxpedda (changes requested):
I'm against introducing this generic, we already have solutions in place for this kind of issue in MacOS and Web, we just need to apply them here as well as @madsmtm already suggested.

For Web, we just need to wrap the innards into `MainThreadSafe`, that's it. I'm happy to figure it out once `NotSendSyncWindowAttributes` is removed.

src/lib.rs
```diff
@@ -160,3 +160,18 @@ mod platform_impl;
...
+/// and `Sync`, since they always excute on a single thread.
+///
+/// # Safety
+///
+/// Winit can run only one event loop at the time and the event loop itself is tied to some thread.
+/// The objects could be send across the threads, but once passed to winit, they execute on the
+/// mean thread if the platform demands it. Thus marking such objects as `Send + Sync` is safe.
+#[doc(hidden)]
+#[derive(Clone, Debug)]
+pub struct SendSyncWrapper<T>(pub T);
```
kchibisov, line 174: Not sure whether this object should be public at all, the only reason it's public is because we allow access to `WindowAttributes` and also building them with struct builder.

We could make it crate private and just provide a `get`/`set for `parent` on WindowAttributes?

kchibisov (commented):
I'm only not sure what to do with the `video` mode stuff, Like it's not send/sync in general, but given that macOS will allow creating it only on the main thread, and then winit will use it on the main thread, I think just marking `Fullscreen` in builder should be fine.

There probably better solutions, but searching for macOS docs on what is MT-safe and what is not is really hard.

src/window.rs
```diff
@@ -159,7 +159,7 @@ pub struct WindowAttributes {
     pub resize_increments: Option<Size>,
     pub content_protected: bool,
     pub window_level: WindowLevel,
-    pub parent_window: Option<RawWindowHandle>,
+    pub parent_window: SendSyncWrapper<Option<RawWindowHandle>>,
```
daxpedda, line 162: I think we should either move this out of `WindowAttributes` into `PlatformSpecificWindowBuilderAttributes` or mark it as private. This would also solve the issue with the visibility of `SendSyncWrapper` pointed out above.

src/window.rs
```diff
@@ -478,7 +478,7 @@ impl WindowBuilder {
     /// - **Android / iOS / Wayland / Web:** Unsupported.
     #[inline]
     pub unsafe fn with_parent_window(mut self, parent_window: Option<RawWindowHandle>) -> Self {
```
daxpedda, line 480: If `RawWindowHandle` isn't thread-safe how is this supposed to work?
I guess it's `unsafe`, so it's fine, but maybe panicking would be better here.

Actually, this might be better suited for another PR, but as pointed out above, maybe this should be moved into `PlatformSpecificWindowBuilderAttributes` and call a backend-specific function and handle it there, including panicking if we aren't on the main thread.

src/window.rs
```diff
@@ -1522,10 +1522,8 @@ unsafe impl HasRawWindowHandle for Window {
     /// [`Event::Resumed`]: crate::event::Event::Resumed
     /// [`Event::Suspended`]: crate::event::Event::Suspended
     fn raw_window_handle(&self) -> RawWindowHandle {
-        struct Wrapper(RawWindowHandle);
-        unsafe impl Send for Wrapper {}
         self.window
-            .maybe_wait_on_main(|w| Wrapper(w.raw_window_handle()))
```
daxpedda, line 1528: I asked this in IRC, but I'm confused about this part.
How can we return `RawWindowHandle` here if it isn't `Send`? Shouldn't we panic if this isn't the main thread? As far as I can tell on every backend we just return it without a second thought, at least on Web this is invalid.

daxpedda (changes requested):

src/window.rs
kchibisov: You can do whatever you want with raw-window-handle, since it's just a pointer. However you can access it only in a specific manner most of the time, which is thread safe for winit.

src/window.rs
kchibisov: You just set a data, winit can only read this data on the thread event loop is on.

src/window.rs
daxpedda: Alternatively, we could just wrap `WindowAttributes` in `SendSyncWrapper` ...

daxpedda:
This is broken on Web and shouldn't be `Send + Sync` there.

kchibisov:
Exactly, we discussed that, it's always used on the thread event loop is on.

Issue #3132, opened 2023-10-07
Window class usage is not enforced


Zoxc:
On Windows registering the window class is not checked for errors. This means we can end up creating a window with the wrong class from some other code or another version of `winit`.

In the case of another version of `winit` this can cause crashes as `public_window_callback` expects `lpCreateParams` to contain a pointer to a `InitData` from the same `winit` version.

PR #3137, opened 2023-10-11
Use an unique window class name per winit version

Zoxc:
This uses an unique window class name per winit version and checks for errors when registering class names to avoid unsafety.

Fixes https://github.com/rust-windowing/winit/issues/3132

### DIFF
--- a/src/platform_impl/windows/mod.rs
+++ b/src/platform_impl/windows/mod.rs
@@ -30,7 +30,7 @@ pub struct PlatformSpecificWindowBuilderAttributes {
     pub no_redirection_bitmap: bool,
     pub drag_and_drop: bool,
     pub skip_taskbar: bool,
-    pub class_name: String,
+    pub class_name: Option<String>,
     pub decoration_shadow: bool,
 }
 
@@ -43,7 +43,7 @@ impl Default for PlatformSpecificWindowBuilderAttributes {
             no_redirection_bitmap: false,
             drag_and_drop: true,
             skip_taskbar: false,
-            class_name: "Window Class".to_string(),
+            class_name: None,
             decoration_shadow: false,
         }
     }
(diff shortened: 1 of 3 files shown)


madsmtm:
Having multiple instances of `winit` in the same application is not really supported.

Can you explain why you're hitting this issue? Maybe including [`CARGO_PKG_VERSION`](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates) in the default name would be the better fix?

In any case, I do agree that we should do the error handling better.

Zoxc:
It's easy to end up with multiple different `winit` versions due to frequent incompatible releases, that's how I noticed the issue.

`CARGO_PKG_VERSION` is not guaranteed to be unique, unlike static addresses.

madsmtm:
Hmm, I guess I just don't see how you'd end up with _running_ two versions of `winit` in the same binary?

Zoxc:
I'm running `puffin_egui` in a separate thread from the main GUI app.

Zoxc:
The event loop window class had the same issue, so I've changed that too.

notgull (commented):
@Zoxc Have you tested this?

Zoxc:
Yeah. I get multiple windows with multiple winit versions instead of a crash.

src/platform_impl/windows/mod.rs
```diff
@@ -30,7 +30,7 @@ pub struct PlatformSpecificWindowBuilderAttributes {
     pub no_redirection_bitmap: bool,
     pub drag_and_drop: bool,
     pub skip_taskbar: bool,
-    pub class_name: String,
```
madsmtm, line 33: Why is it not enough to use this existing functionality to override the class name?

Zoxc: I'm not looking to change the class name (or the code that directly uses winit).

Also the result of registering the window class has to be checked to ensure memory safety.

PR #3138, opened 2023-10-12
Update SCTK to 0.18.0

### LINKED ISSUE #2927: Support `Occluded` event for Wayland
It's now possible with `xdg_shell` v6.

kchibisov:
The update is pretty minor, however we support now `WindowEvent::Occluded` when xdg-shell v6 is available.

It also adds support for `Window::show_window_menu`.

Fixes #2927.

--

Blocked on `sctk-adwaita` release.


src/platform_impl/linux/x11/mod.rs
```diff
@@ -269,7 +269,8 @@ impl<T: 'static> EventLoop<T> {
 
         // Create the X11 event dispatcher.
         let source = X11Source::new(
-            xconn.xcb_connection().as_raw_fd(),
+            // SAFETY: xcb owns the FD and outlives the source.
```
kchibisov, line 272: r @notgull . You could probably do something smarter than that, but I just did a conservative change.

ids1024: I guess `XKBConnection` doesn't implement `AsFd`? It probably should.

Then it could be contained in the `Generic` directly. After wrapping with ref-counting if another reference is needed outside the source.

kchibisov: well, upstream only given `c_int` as a fd iirc.

ids1024: https://github.com/psychon/x11rb/pull/889 adds an `AsFd` implementation here.

notgull: This works for now, until the PR above is published.

PR #3142, opened 2023-10-13
Rwh send sync

madsmtm:
Alternative to https://github.com/rust-windowing/winit/pull/3136. Builds upon https://github.com/rust-windowing/winit/pull/3126 to remove the `unsafe` from `WindowBuilder::with_parent_window`, by taking the lifetime inherent in that, and extending it on the platforms where it is relevant.

TODO:
- [ ] Merge the aforementioned PR.
- [ ] Test on all platforms changed
- [ ] Add entry to `CHANGELOG.md`


kchibisov:
>Alternative to https://github.com/rust-windowing/winit/pull/3136.

This is not an alternative since it doesn't achieve what said PR achieves.

However doing something like that is fine.

kchibisov:
After looking into it, I don't think it's a good idea as of now to accept owned handler if winit can only provide `Raw` handles.

This update should be a part of the winit safe handles work.

daxpedda:
That should actually be good to go now, right?

daxpedda:
After looking at this a bit more closely, I question how thread safety is achieved here exactly. `WindowBuilder` is `Send`, but `with_parent()` can be called on any thread, but `WindowHandle` isn't `Send`. So a `WindowHandle` that has been passed into `with_parent()` that would be valid on a thread that is not the main thread, is suddenly being used in the main thread.

madsmtm:
> thread safety

I think this works on platforms with a "main-thread only" concept, unsure about platforms where a handle may be bound to a specific thread (Windows?)

My reasoning for main thread only being safe on macOS/iOS:
- `WindowHandle<'_>` is `!Send + !Sync`, because it is bound to the main thread.
- If we get that in `with_parent_window`, we hence know we're running on the main thread (because of the yet-to-be-documented contract on `WindowHandle<'_>`).
- This we put it into `OwnedWindowHandle`, which contains a `MainThreadBound`, meaning that we can only ever access the field off the main thread.
- So now, even though `WindowBuilder` is moved, we're guaranteed to only access the parent window on the main thread.

daxpedda:
If the only valid thread is the main thread, I agree there is no problem, because `WindowBuilder` only gets consumed in the main thread, which is checked by Winit.

Remaining problems:
- How to correctly implement `OwnedWindowHandle` on all platforms if at all possible, alternatively we have to discuss adding a lifetime to `WindowBuilder`.
- Determine thread safety for all platforms. I think the whole `!Send + !Sync` but valid on the main thread should be documented in `raw-window-handle` as well.

PR #3145, opened 2023-10-14
Upgrade to `ndk 0.8`, `ndk-sys 0.5` and `android-activity 0.5` releases

MarijnS95:
Depends on #3126, https://github.com/rust-mobile/android-activity/pull/128

CC @notgull, @kchibisov

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,6 +45,9 @@ wayland-csd-adwaita-notitle = ["sctk-adwaita"]
 android-native-activity = ["android-activity/native-activity"]
 android-game-activity = ["android-activity/game-activity"]
 serde = ["dep:serde", "cursor-icon/serde", "smol_str/serde"]
+rwh_04 = ["dep:rwh_04", "ndk/rwh_04"]
+rwh_05 = ["dep:rwh_05", "ndk/rwh_05"]
+rwh_06 = ["dep:rwh_06", "ndk/rwh_06"]
 
 [build-dependencies]
 cfg_aliases = "0.1.1"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -69,10 +72,11 @@ simple_logger = { version = "2.1.0", default_features = false }
 softbuffer = "0.3.0"
 
 [target.'cfg(target_os = "android")'.dependencies]
-# Coordinate the next winit release android-activity 0.5 release
-android-activity = "0.5.0-beta.1"
-ndk = "0.8.0-beta.0"
-ndk-sys = "0.5.0-beta.0"
+# https://github.com/rust-mobile/android-activity/pull/128
+# android-activity = "0.5.0"
+android-activity = { git = "https://github.com/MarijnS95/android-activity", rev = "720e5a3" }
```
notgull, line 77: Since we're using a git dependency, should this PR be marked as draft for now?

notgull (approved):
Looks good to me!

Cargo.toml
kchibisov: Yeah, git dependencies are not allowed for stable releases.

MarijnS95: It is already marked as `Depends on`, and I'd rather push forward on getting a release out some time soon.

@rib if you can, the NDK crate is now ready and `android-activity` can go out of beta too.

kchibisov: @MarijnS95 distributions don't package crates having `git` dependencies, even for non-native platforms. I can't do a release with the `git` dependencies.

Maybe sending email to @rib will work if they are missing.

MarijnS95: @kchibisov we are thinking and saying the same.

I'm just not the release manager for `android-activity`, and @rib doesn't seem to have the `cargo publish` step automated in the CI with a token in GitHub secrets so I won't be able publish a release in his name either way.

Regardless, it hasn't been 24h yet since I published the `ndk` and it's still weekend so I wouldn't call it missing, just unfortunate timing.

MarijnS95: Anyway, I've converted this PR to "draft" as that seems to be the main point of contention (...?).

kchibisov: >@kchibisov we are thinking and saying the same.

Ah, release out of **android activity**, not winit. I've opened an issue on their tracker. At least it seems like issues are being read quite often.

MarijnS95: I'm sure he's fully aware of the release schedule, just unfortunate timing on my end and it's quite a long trail (release `raw-window-handle` -> `ndk(-sys)`, `android-activity`, `winit`, all owned and maintained by different people).

notgull (approved):

kchibisov (approved):

PR #3148, opened 2023-10-15
Implement `AsFd`/`AsRawFd` for `EventLoop<T>`

kchibisov:
This should help other crates to integrate winit's event loop into their bigger event loop without adding an extra thread.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On iOS, send events `WindowEvent::Occluded(false)`, `WindowEvent::Occluded(true)` when application enters/leaves foreground.
 - **Breaking** add `Event::MemoryWarning`; implemented on iOS/Android.
 - On Wayland, support `Occluded` event with xdg-shell v6
+- Implement `AsFd`/`AsRawFd` for `EventLoop<T>` on X11 and Wayland.
 
 # 0.29.1-beta
 
--- a/src/platform_impl/linux/x11/mod.rs
+++ b/src/platform_impl/linux/x11/mod.rs
@@ -32,7 +32,7 @@ use std::{
     ops::Deref,
     os::{
         raw::*,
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -34,6 +34,7 @@ And please only add new entries to the top of this list, right below the `# Unre
 - On iOS, send events `WindowEvent::Occluded(false)`, `WindowEvent::Occluded(true)` when application enters/leaves foreground.
 - **Breaking** add `Event::MemoryWarning`; implemented on iOS/Android.
 - On Wayland, support `Occluded` event with xdg-shell v6
+- Implement `AsFd`/`AsRawFd` for `EventLoop<T>` on X11 and Wayland.
```
notgull, line 37: Maybe it would be nice to implement this for `EventLoopWindowTarget` as well?

src/platform_impl/linux/x11/mod.rs
```diff
@@ -658,6 +658,18 @@ impl<T: 'static> EventLoop<T> {
     }
 }
 
+impl<T> AsFd for EventLoop<T> {
+    fn as_fd(&self) -> BorrowedFd<'_> {
+        self.event_loop.as_fd()
```
notgull, line 663: This will fail to compile on systems without `epoll` and `kqueue`, like AIX. I'm not sure if we support any of these systems but it should be written down somewhere.

kchibisov: calloop doesn't support them, so no need to write down.

CHANGELOG.md
kchibisov: I'm not sure that it worth it, basically you want to get `fd` outside the `run` closure, since inside it it's weird, since your event loop is external?

Like it doesn't make any sense why anyone would do that...

notgull: Alright, we can put a pin in that for now

notgull (approved):
LGTM

Issue #3153, opened 2023-10-16
`CursorEntered` and `CursorLeft` events not getting sent if we leave the windows while holding the mouse button down in Windows


YouKnow-sys:
If we press one of mouse button (lets say LB) while inside the client area of the window and then proceed to move the cursor out of the window we no longer receive the `CursorLeft` event, and if we keep holding the mouse button down and then go back to window we wont get `CursorEntered` event either.
this happen because we call `SetCapture` method when user click in client area, and this will lead to windows to continue tracking mouse position and then just keep sending `WM_MOUSEMOVE` event until we release the button (more information in [this url](https://stackoverflow.com/a/2548865))
there are workarounds like manually handle this case like what link above suggested.
I'll send a PR soon with the mentioned fix.

PR #3154, opened 2023-10-16
Fixed `CursorEntered` and `CursorLeft` events not getting sent when holding a mouse button down

YouKnow-sys:
This PR fix the problem mentioned in https://github.com/rust-windowing/winit/issues/3153


src/platform_impl/windows/event_loop.rs
```diff
@@ -1438,39 +1438,63 @@ unsafe fn public_window_callback_inner<T: 'static>(
...
+                        .set_cursor_flags(window, |f| f.set(CursorFlags::IN_WINDOW, false))
+                        .ok();
+
+                    userdata.send_event(Event::WindowEvent {
+                        window_id: RootWindowId(WindowId(window)),
+                        event: CursorLeft {
+                            device_id: DEVICE_ID,
+                        },
+                    });
+                }
```
kchibisov, line 1495: Could you extract all of this mess into a function accepting: window, x, y, position and returning something like `PointerMoveKind { Move, Enter, Leave }` so you can then do something like that
```rust
match get_pointer_move_kind(window, position, x, y) {
    PointerMoveKind::Leave => {
        // Cursor left handling
    }
    PointerMoveKind::Enter => {
       // Cursor enter handling
    }
    PointerMoveKind::Move => (),
}

**Review by kchibisov — CHANGES_REQUESTED** [review]

File: src/platform_impl/windows/event_loop.rs [hunk 2]
```diff
@@ -914,6 +914,41 @@ fn normalize_pointer_pressure(pressure: u32) -> Option<Force> {
...
+enum PointerMoveKind {
+    /// Pointer enterd to the window.
+    Enter,
+    /// Pointer leaved the window client area.
+    Leave,
+    /// Pointer is inside the window or `GetClientRect` failed.
+    None,
+}
+
+unsafe fn get_pointer_move_kind(
```
kchibisov, line 926: But the function itself is safe. Also move it to the bottom of the file with the `PointerMoveKind` thing.

kchibisov (changes requested):
Also missing a changelog entry.

src/platform_impl/windows/event_loop.rs
kchibisov: `unsafe` functions in rust serve a different purpose usually, for example, when you need to hold some lifetime, etc invariant when you call a function, which usually is tracked by rust.

In this case though, we know that it as a whole is safe, since the only call uses `HWND` is error checked.

kchibisov (approved):

PR #3155, opened 2023-10-16
Android: map unknown `Keycodes` `Key::Unidentified`

rib:
Now that the `android_activity::input::Keycodes` enum is `#[non_exhaustive]` but can also preserve the SDK integer value for any keycodes that might not be known about at compile time, this forwards any unknown keycode as `Key::Unidentified`.

_(Required to be compatible with the `android-activity` 0.5.0 release)_


kchibisov:
@rib maybe you should send your patch against https://github.com/rust-windowing/winit/pull/3145 along the `android-activity 0.5`? You can post a diff there and I just merge the update.

The CI won't pass with such a change as of now.

rib:
> @rib maybe you should send your patch against #3145 along the `android-activity 0.5`? You can post a diff there and I just merge the update.
> 
> The CI won't pass with such a change as of now.

Okey, yeah, makes sense. I've squashed this change into https://github.com/rust-windowing/winit/pull/3145

PR #3158, opened 2023-10-17
On macOS, fix tabGroup misuse

kchibisov:
The property is marked as `Weak`, however we used strong `Id`.

Links: https://github.com/alacritty/alacritty/issues/7249

### DIFF
--- a/src/platform_impl/macos/appkit/tab_group.rs
+++ b/src/platform_impl/macos/appkit/tab_group.rs
@@ -23,7 +23,7 @@ extern_methods!(
         pub fn selectPreviousTab(&self);
 
         #[method_id(windows)]
-        pub fn tabbedWindows(&self) -> Id<NSArray<NSWindow>>;
+        pub fn tabbedWindows(&self) -> Option<Id<NSArray<NSWindow>>>;
 
         #[method(setSelectedWindow:)]
         pub fn setSelectedWindow(&self, window: &NSWindow);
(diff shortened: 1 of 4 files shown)


src/platform_impl/macos/appkit/tab_group.rs
```diff
@@ -23,7 +23,7 @@ extern_methods!(
         pub fn selectPreviousTab(&self);
 
         #[method_id(windows)]
-        pub fn tabbedWindows(&self) -> Id<NSArray<NSWindow>>;
+        pub fn tabbedWindows(&self) -> Option<Id<NSArray<NSWindow>>>;
```
kchibisov, line 26: This is nullable.

src/platform_impl/macos/appkit/window.rs
```diff
@@ -218,7 +218,7 @@ extern_methods!(
         pub(crate) fn tabbingIdentifier(&self) -> Id<NSString>;
 
         #[method_id(tabGroup)]
-        pub(crate) fn tabGroup(&self) -> Id<NSWindowTabGroup>;
+        pub(crate) fn tabGroup(&self) -> Option<Id<NSWindowTabGroup>>;
```
kchibisov, line 221: This is `weak`.

kchibisov:
Handling is done the same as other APIs already do that, so it's either will fork just fine or they are all broken anyway.

PR #3167, opened 2023-10-20
Fix doc links

dhardy:


kchibisov (approved):

dhardy:
BTW are we supposed to import HasRawWindowHandle directly from its own crate now? That is inconvenient.

kchibisov:
>BTW are we supposed to import HasRawWindowHandle directly from its own crate now? That is inconvenient.

We have `re-export` but it's a current default re-export. I'm not sure whether we want to have all 3 raw-window-handles re-exported, I'd rather use only the current one.

kchibisov:
https://github.com/rust-windowing/winit/blob/master/src/lib.rs#L143

PR #3172, opened 2023-10-22
On Windows, fix deadlock during `Cursor{Enter,Leave}`

### LINKED ISSUE #3171: Windows: Deadlock reading window state during some window events
Using winit 0.29.2 this code never prints "[..] after state read" for `CursorEntered`:

```rust
use winit::{event_loop::EventLoop, window::WindowBuilder, event::Event};

fn main() -> Result<(), impl std::error::Error> {
    let event_loop = EventLoop::new().unwrap();

    event_loop.run(move |event, _| {
        if let Event::WindowEvent { event, .. } = event {
            println!("{event:?} before state read");
            let _s = window.is_maximized();
            println!("{event:?} after state read");
        }
    })
}
```

kchibisov:
The lock was not released when calling back to the user.

Fixes #3171.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ Unreleased` header.
 
 # Unreleased
 
+- On Windows, fix deadlock when accessing the state during `Cursor{Enter,Leave}`.
+
 # 0.29.2
 
 - **Breaking:** Bump MSRV from `1.60` to `1.65`.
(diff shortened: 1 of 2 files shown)


SamRodri (commented):
@kchibisov this causes another deadlock because the lock is not dropped for PointerMoveKind::None.

PR #3178, opened 2023-10-24
On wasm, provide intradoc-link for `spawn()` function in `EventLoop` docs

MarijnS95:
For https://github.com/rust-windowing/winit/commit/0d366ffbdaabab567f595b38710e8d72f150b514#r130681051

And on any other platform, emit a footnote explaining that this function is only available on wasm.

EDIT: Apologies, it looks like `gh pr create` pushed this through the main repo instead of my fork. Hence there's no `marijn/` prefix on the branchname.

### DIFF
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -224,14 +224,22 @@ impl<T> EventLoop<T> {
     ///   (that Rust doesn't see) that will also mean that the rest of the function is never executed
     ///   and any values not passed to this function will *not* be dropped.
     ///
-    ///   Web applications are recommended to use `spawn()` instead of `run()` to avoid the need
+    ///   Web applications are recommended to use
+    #[cfg_attr(
+        wasm_platform,
+        doc = "[`EventLoopExtWebSys::spawn()`][crate::platform::web::EventLoopExtWebSys::spawn()]"
+    )]
+    #[cfg_attr(not(wasm_platform), doc = "`EventLoopExtWebSys::spawn()`")]
+    ///   [^1] instead of [`run()`] to avoid the need
     ///   for the Javascript exception trick, and to make it clearer that the event loop runs
     ///   asynchronously (via the browser's own, internal, event loop) and doesn't block the
     ///   current thread of execution like it does on other platforms.
     ///
     ///   This function won't be available with `target_feature = "exception-handling"`.
     ///
-    /// [`set_control_flow()`]: EventLoopWindowTarget::set_control_flow
+    /// [`set_control_flow()`]: EventLoopWindowTarget::set_control_flow()
+    /// [`run()`]: Self::run()
+    /// [^1]: `EventLoopExtWebSys::spawn()` is only available on WASM.
     #[inline]
     #[cfg(not(all(wasm_platform, target_feature = "exception-handling")))]
     pub fn run<F>(self, event_handler: F) -> Result<(), EventLoopError>
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -26,16 +26,36 @@
 //! window or a key getting pressed while the window is focused. Devices can generate
 //! [`DeviceEvent`]s, which contain unfiltered event data that isn't specific to a certain window.
 //! Some user activity, like mouse movement, can generate both a [`WindowEvent`] *and* a
-//! [`DeviceEvent`]. You can also create and handle your own custom [`UserEvent`]s, if desired.
+//! [`DeviceEvent`]. You can also create and handle your own custom [`Event::UserEvent`]s, if desired.
 //!
-//! You can retrieve events by calling [`EventLoop::run`][event_loop_run]. This function will
+//! You can retrieve events by calling [`EventLoop::run()`]. This function will
 //! dispatch events for every [`Window`] that was created with that particular [`EventLoop`], and
-//! will run until [`exit()`] is used, at which point [`Event`]`::`[`LoopExiting`].
+//! will run until [`exit()`] is used, at which point [`Event::LoopExiting`].
 //!
 //! Winit no longer uses a `EventLoop::poll_events() -> impl Iterator<Event>`-based event loop
 //! model, since that can't be implemented properly on some platforms (e.g web, iOS) and works poorly on
 //! most other platforms. However, this model can be re-implemented to an extent with
-//! [`EventLoopExtPumpEvents::pump_events`]. See that method's documentation for more reasons about why
+#![cfg_attr(
+    any(
+        windows_platform,
+        macos_platform,
+        android_platform,
+        x11_platform,
+        wayland_platform
+    ),
+    doc = "[`EventLoopExtPumpEvents::pump_events()`][platform::pump_events::EventLoopExtPumpEvents::pump_events()]"
+)]
+#![cfg_attr(
+    not(any(
+        windows_platform,
+        macos_platform,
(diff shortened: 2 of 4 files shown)


src/event_loop.rs
```diff
@@ -224,14 +224,26 @@ impl<T> EventLoop<T> {
...
     ///   This function won't be available with `target_feature = "exception-handling"`.
     ///
-    /// [`set_control_flow()`]: EventLoopWindowTarget::set_control_flow
+    #[cfg_attr(
+        not(wasm_platform),
+        doc = "[^1]: `spawn()` is only available on `wasm` platforms."
+    )]
+    ///
+    /// [`set_control_flow()`]: EventLoopWindowTarget::set_control_flow()
+    /// [`run()`]: Self::run()
```
MarijnS95, line 246: Not super happy with this, because it'll fail to link on `not(all(wasm_platform, target_feature = "exception-handling"))`.

src/lib.rs
```diff
@@ -121,11 +121,10 @@
 //! [window_builder_new]: window::WindowBuilder::new
 //! [window_builder_build]: window::WindowBuilder::build
 //! [window_id_fn]: window::Window::id
-//! [`Event`]: event::Event
-//! [`WindowEvent`]: event::WindowEvent
+//! [`Event::WindowEvent`]: event::WindowEvent
 //! [`DeviceEvent`]: event::DeviceEvent
 //! [`UserEvent`]: event::Event::UserEvent
-//! [`LoopExiting`]: event::Event::LoopExiting
+//! [`Event::LoopExiting`]: event::Event::LoopExiting
```
MarijnS95, line 127: I don't want this PR to turn into a "let's fix _all_ `winit` intradoc links", but a few more changes are still needed to make this consistent.

No clue what the point is with manual relative `.html` links that may not resolve when certain features are not turned on... Pointing the user to a 404 page :(

src/event_loop.rs
daxpedda: `target_feature = "exception-handling"` isn't stable, which is probably why it didn't work for you.

MarijnS95: @daxpedda reading this again, I missed `target_` and assumed it was just `feature =` (from the `[features]` list in `Cargo.toml`).

I think the feature must have been _en_abled (running nightly rustdoc for `--cfg docsrs` => `doc_cfg`), as it's the only way to cause `not(all(true, true)) = not(true) = false` and turn _off_ `fn run()`?

daxpedda: That is weird, `target_feature = "exception-handling"` should not automatically have been enabled by running nightly rustdoc.

src/event_loop.rs
```diff
@@ -224,14 +224,26 @@ impl<T> EventLoop<T> {
     ///   (that Rust doesn't see) that will also mean that the rest of the function is never executed
     ///   and any values not passed to this function will *not* be dropped.
     ///
-    ///   Web applications are recommended to use `spawn()` instead of `run()` to avoid the need
+    ///   Web applications are recommended to use
+    #[cfg_attr(
+        wasm_platform,
+        doc = "[`EventLoopExtWebSys::spawn()`][crate::platform::web::EventLoopExtWebSys::spawn()]"
```
daxpedda, line 230:
```suggestion
        doc = "[`EventLoopExtWebSys::spawn()`](crate::platform::web::EventLoopExtWebSys::spawn())"
```
Is the convention we are using as far as I'm aware.

src/lib.rs
```diff
@@ -26,16 +26,36 @@
...
-//! [`EventLoopExtPumpEvents::pump_events`]. See that method's documentation for more reasons about why
+#![cfg_attr(
+    any(
+        windows_platform,
+        macos_platform,
+        android_platform,
+        x11_platform,
+        wayland_platform
+    ),
+    doc = "[`EventLoopExtPumpEvents::pump_events()`][platform::pump_events::EventLoopExtPumpEvents::pump_events()]"
```
daxpedda, line 46:
```suggestion
    doc = "[`EventLoopExtPumpEvents::pump_events()`](platform::pump_events::EventLoopExtPumpEvents::pump_events())"
```

src/platform/web.rs
```diff
@@ -109,8 +109,8 @@ pub trait EventLoopExtWebSys {
 
     /// Initializes the winit event loop.
     ///
-    /// Unlike `run`, this returns immediately, and doesn't throw an exception in order to
-    /// satisfy its `!` return type.
+    /// Unlike `run()`, this returns immediately, and doesn't throw an exception in order to
```
daxpedda, line 112: Can we link here to run as well?

daxpedda (changes requested):

src/lib.rs
daxpedda: This is probably a problem for another time, but these methods are deprecated anyway.

src/event_loop.rs
MarijnS95: I'm fairly confident that `rustdoc` defines `[foo][bar]` as extension to link to symbols (in addition to a label name): https://rust-lang.github.io/rfcs/1946-intra-rustdoc-links.html#additions-to-the-documentation-syntax

Hence there must either be a label definition (e.g. `/// [bar]: ...`) _or_ `bar` must be a valid Rust path in scope.

On the contrary, `[foo](bar)` is defined as `bar` being a URL, so there are cases where an invalid Rust path is (were?) treated as link, resulting in no errors at build-time and a useless/broken link for the user.

---

I am referencing to the past here because I **do** now get build-time errors for these, but am still wary of edge-cases. And while `[]()` has precedence, `[][]` is definitely used in `winit` as well.

src/platform/web.rs
MarijnS95: I might have not done that because of the above, but sure. Will need to add more rigorous checks for `exception-handling` anyway.

MarijnS95: Done. And in light of #3076 I have changed the footnotes to _always_ be emitted, so that readers on WASM _also_ get the same information (right in-place) that the reference is only available on their platform (with a specific feature).

daxpedda (approved):

src/event_loop.rs
daxpedda: > The point is that it supports that _in addition to URLs_.

Absolutely, I understand, it's why this totally makes sense to me!
Thank you for the explanation!

MarijnS95: Little spark; found it: https://github.com/gtk-rs/gir/pull/1162

> Rustdoc treats `[foo]` as item path, and by the same rule `[fancy text][foo]` too. Unlike `[]()`, which generates a dead link if the right-hand side is a valid Rust item path but doesn't exist in scope (common and unfortunately unavoidable with docs for C code auto-ported to Rust), the `[]` and `[][]` do not generate a link at all.

And there are more details in the whole chain of linked issues and PRs:

https://github.com/gtk-rs/gir/issues/1140#issuecomment-842131324
https://github.com/gtk-rs/gir/pull/1150

daxpedda:
Just rebased so we can merge.

src/event_loop.rs
MarijnS95: Fair enough, I think I replaced those locally with the symbol rather than going through a named link so had a few more, but abandoned that change for the sake of keeping this PR "small".

PR #3181, opened 2023-10-25
redox: Update `redox_syscall` dependency

ids1024:


kchibisov (changes requested):
You need to update `orbclient` to a version having `0.4` of syscall, that's the only reason I haven't updated the `syscall` myself.

PR #3189, opened 2023-10-25
On Wayland, improve initial user size handling

### LINKED ISSUE #3187: Delay `size` convertion on Wayland until we get scaling for initial window
This should be possible now with the updated wording to `xdg_shell`, new `wl_compositor`, and `wp_fractional_scale`.

However, we can't do anything like that about logical sizes for `min/max` inner sizes.

kchibisov:
Keep the user provided size in the original values and convert only when we're getting a `configure` event. On some compositors will have a scale available, so it'll work, however with some we'll still have old 'pick 1` as default.

Also configure_bounds when compositor tells the user to pick the size, that will ensure that initial `with_inner_size` won't grow beyond the working area.

Fixes #3187.

@dhardy can you test this? It should probably make the `PhysicalSize` work on at least some systems and if you pass the big `with_inner_size` winit will constrain to the the size compositor wants.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,10 @@ Unreleased` header.
 
 # Unreleased
 
+- On Wayland, apply correct scale to `PhysicalSize` passed in `WindowBuilder::with_inner_size` when possible.
 - On Wayland, fix `RedrawRequsted` being always sent without decorations and `sctk-adwaita` feature.
+- On Wayland, ignore resize requests when the window is fully tiled.
+- On Wayland, use `configure_bounds` to constrain `with_inner_size` when compositor wants users to pick size.
 - On Windows, fix deadlock when accessing the state during `Cursor{Enter,Leave}`.
 - On macOS, fix deadlock when entering a nested event loop from an event handler.
 
--- a/src/platform_impl/linux/wayland/window/state.rs
+++ b/src/platform_impl/linux/wayland/window/state.rs
@@ -28,7 +28,7 @@ use sctk::shm::Shm;
 use sctk::subcompositor::SubcompositorState;
 use wayland_protocols_plasma::blur::client::org_kde_kwin_blur::OrgKdeKwinBlur;
 
-use crate::dpi::{LogicalPosition, LogicalSize};
+use crate::dpi::{LogicalPosition, LogicalSize, PhysicalSize, Size};
 use crate::error::{ExternalError, NotSupportedError};
 use crate::event::WindowEvent;
 use crate::platform_impl::wayland::event_loop::sink::EventSink;
@@ -133,6 +133,10 @@ pub struct WindowState {
     /// sends `None` for the new size in the configure.
     stateless_size: LogicalSize<u32>,
 
+    /// Initial window size provided by the user. Removed on the first
+    /// configure.
+    initial_size: Option<Size>,
+
     /// The state of the frame callback.
     frame_callback_state: FrameCallbackState,
 
@@ -153,7 +157,7 @@ impl WindowState {
         connection: Connection,
         queue_handle: &QueueHandle<WinitState>,
         winit_state: &WinitState,
-        size: LogicalSize<u32>,
+        initial_size: Size,
         window: Window,
         theme: Option<Theme>,
     ) -> Self {
@@ -194,8 +198,9 @@ impl WindowState {
             resizable: true,
             scale_factor: 1.,
             shm: winit_state.shm.wl_shm().clone(),
-            size,
-            stateless_size: size,
+            size: initial_size.to_logical(1.),
+            stateless_size: initial_size.to_logical(1.),
+            initial_size: Some(initial_size),
             text_inputs: Vec::new(),
             theme,
             title: String::default(),
@@ -253,6 +258,14 @@ impl WindowState {
         subcompositor: &Arc<SubcompositorState>,
         event_sink: &mut EventSink,
     ) -> LogicalSize<u32> {
+        // NOTE: when using fractional scaling or wl_compositor@v6 the scaling
+        // should be delivered before the first configure, thus apply it to
+        // properly scale the physical sizes provided by the users.
+        if let Some(initial_size) = self.initial_size.take() {
+            self.size = initial_size.to_logical(self.scale_factor());
+            self.stateless_size = self.size;
+        }
+
         if configure.decoration_mode == DecorationMode::Client
             && self.frame.is_none()
             && !self.csd_fails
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ Unreleased` header.
 # Unreleased
 
 - On Windows, fix deadlock when accessing the state during `Cursor{Enter,Leave}`.
+- On Wayland, apply correct scale to `PhysicalSize` passed in `WindowBuilder::with_inner_size` when possible.
+- On Wayland, use `configure_bounds` to constrain `with_inner_size` when compositor wants users to pick size.
```
YaLTeR, line 16: I think that there should still be a way to ignore the `configure_bounds` hint. I.e. if I'm making a game and want to spawn it at an exact resolution for testing, even if larger than the working area, I should be able to do that.

kchibisov: The constraint is the same as in GTK btw. Games do create in fullscreen though, so all of that is not applied to them, since fullscreen has size in configure.

YaLTeR: > The constraint is the same as in GTK btw

I argued with GTK devs about this too, I think it's wrong there also.

> Games do create in fullscreen though, so all of that is not applied to them, since fullscreen has size in configure.

Not necessarily. Windowed mode in games is very much a thing, there's even "borderless window" mode in many games.

kchibisov: I still don't know a single game which can open large than working area. And anyway, the game can resize itself to the size it wants it to be.

src/platform_impl/linux/wayland/window/state.rs
```diff
@@ -297,30 +310,46 @@ impl WindowState {
...
+        if constrain {
+            let bounds = self.inner_size_bounds(&configure);
+            new_size.width = bounds
+                .0
+                .map(|bound_w| new_size.width.min(bound_w.get()))
+                .unwrap_or(new_size.width);
+            new_size.width = bounds
+                .1
+                .map(|bound_h| new_size.height.min(bound_h.get()))
+                .unwrap_or(new_size.height);
```
dhardy, line 350: That should be `new_size.height = ...`. Explains why my windows are square...

dhardy:
@kchibisov this does fix initial window sizing using physical-size as input for me (kwin / Wayland), yes. (The height, that is.)

Ultimately though, should winit support constructing windows with physical size on Wayland at all? It turns out my code still needs adaptation since initial sizing currently uses the discovered "monitor scale value" of 2 not the window value (1.5)... there is no way to correctly compute the window size in physical pixels before constructing the window.

I can of course refactor the code to construct a window with some initial size (like 1x1) then resize after discovering the scale factor, but this still means that there is zero point supporting initial size specification in physical pixels.

kchibisov:
>I can of course refactor the code to construct a window with some initial size (like 1x1) then resize after discovering the scale factor, but this still means that there is zero point supporting initial size specification in physical pixels.

I mean, usually you just use logical pixels so the window size is the same regardless of what you have. physical size is more for games, not for UIs. We can't just restrict the use of `PhysicalSize` in winit on Wayland, since it's a cross platform toolkit and physical size still works.

Maybe winit should just use logical everywhere for setting sizes, since it's the most reliable thing anyway, but use Physical for drawing, because that's what matters for drawing.

kchibisov:
Also, resizing on Wayland is really cheap, since it's just assign `width` and `height`, the actual resizes are on the user, so you can just create window with the size you want (`LogicalSize`) and then resize again.

kchibisov:
The point is that `PhysicalSize` is mostly for games, they don't care about scale at all, they need this raw physical size data. They also has in-game UI to scale when possible.

GUIs on the other side should model everything in a logical pixels, so they don't need to know anything about `scaling` at all, since the layout is in logical. However when drawing, they could probably need to know about scaling to increase the size.

macOS for example is also logical pixels, but it's just scale of 2 most of the time.

dhardy:
> Maybe winit should just use logical everywhere for setting sizes, since it's the most reliable thing anyway, but use Physical for drawing, because that's what matters for drawing.

Winit should avoid portability hazards or at least document them. There are plenty of cases where platform-specific behaviour *is* already documented in the API, so possibly that's the best option.

It doesn't matter so much if certain conversions must be done by the user, *so long as there is a clear documented method of doing so*. In other words, if, say, window sizing used only logical pixels, but the application wanted a precise size in physical pixels, the application could convert that to logical even though on platforms like X11 it would need to be converted back to physical. `winit::dpi::Size` already uses `f64` for logical sizes so loss-less conversion should be possible.

Caveat: this type of conversion only makes sense when the scale factor is reliably known. On some platforms this requires constructing a window first. It would be nice if other platforms had a method on `EventLoopWindowTarget` like `fn single_scale_factor(&self) -> Option<f64>` which returns `None` on platforms like Wayland which are not restricted to a single scale factor.

dhardy:
> GUIs on the other side should model everything in a logical pixels, so they don't need to know anything about scaling at all, since the layout is in logical.

This is an assumption I've been trying to break for years because linear scaling is not optimal.

kchibisov:
>This is an assumption I've been trying to break for years because linear scaling is not optimal

Well, you can look what QT is doing, it can scale to any scale without any issue.

dhardy:
Err... are you telling me that it uses linear scaling (which obviously *works* even if sometimes sub-optimal) or that it uses non-linear scaling (thus breaking your argument)?

kchibisov:
It has non-linear, but I think it just uses physical pixels and not rely on scaling factor you pick. You can decide on the window sizes based as per-cent of monitor. I'm not familiar with `QT` but restricting physical sizes is not that hard, you also don't need any scaling for it at all, you just use it when you try to scale your internal UI, but the canvas sizes are all physical.

I'm just saying that there're approaches, and only few require you to know scaling before creating the window, and you can always resize.

I know that I'll change all the logic to create window with the next release, so it could not an issue in the future, but I'm still thinking on how to approach everything correctly.

dhardy:
Aha. Yes, that's my approach, it just means that the initial scaled size is more an estimate than the target. But that's fine I guess (and generally the estimate works well enough anyway).

If Winit only supports logical sizes in windows it forces the user to only convert when a scale factor is available, which helps avoid portability hazards.

I think toolkits like mine and Qt may have to use two paths for optimal cross-platform support:

- Where the scale factor is known in advance, calculate the layout and size once and construct the window with the correct initial size. An API like `EventLoopWindowTarget::single_scale_factor` (see above) helps here.
- Otherwise, construct the window then recompute layout and resize after the scale factor is known (or at least recompute layout at the correct size in physical pixels).

dhardy:
> I think toolkits like mine and Qt may have to use two paths for optimal cross-platform support:

But given that X11 is dying and Windows now supports per-window scale-factors, the first case above is not really worth supporting.

PR #3191, opened 2023-10-25
Prefer `[foo][bar]` over `[foo](bar)` for intra-doc links

daxpedda:
Follow-up to a discussion here: https://github.com/rust-windowing/winit/pull/3178#discussion_r1371280634.

The logic here is that `[foo](bar)` can be used for inline links or intra-doc links, but `[foo][bar]` can be used for labels or intra-doc links, so if no label is present, it can only be a intra-doc link, making this less ambiguous and more robust to error check.

### DIFF
--- a/src/application.rs
+++ b/src/application.rs
@@ -11,7 +11,7 @@ pub trait ApplicationHandler<T: 'static = ()> {
     /// This is a useful place to put code that should be done before you start processing
     /// events, such as updating frame timing information for benchmarking or checking the
     /// [`StartCause`] to see if a timer set by
-    /// [`ControlFlow::WaitUntil`](crate::event_loop::ControlFlow::WaitUntil) has elapsed.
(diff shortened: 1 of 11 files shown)


MarijnS95 (approved):

notgull (approved):
Looks good to me!

daxpedda:
For reference: still waiting for some feedback in https://github.com/rust-lang/rust/pull/117178.

daxpedda:
Apparently the conclusion was that there is no preference: https://github.com/rust-lang/rust/pull/121021.

I also double checked and AFAICT Rustdoc will properly generate an error regardless. So the only difference is that `[]()` will generate a dead link but `[][]` will not, but its not really an issue because CI should always catch it.

Personally I would still prefer `[foo][bar]` over `[foo](bar)` despite there not being any official recommendation by Rustdoc but us being consistent in Winit. So if we have consensus I would still like to move forward with this.

madsmtm (approved):
Am fine with either

PR #3195, opened 2023-10-27
Clarify `scale_factor` docs

### LINKED ISSUE #3183: MonitorHandle::scale_factor does not use fractional factor on Wayland
I see `MonitorHandle::scale_factor` reporting 2 while `Window::scale_factor` correctly reports 1.5 on Wayland.

The docs of `Window::scale_factor` note that it uses `wp-fractional-scale` protocol if available, otherwise returning an integer factor. The docs of `MonitorHandle::scale_factor` do not mention any such limitation — is this just a doc issue?

BTW the `apple_2` [link](https://github.com/rust-windowing/winit/blob/master/src/dpi.rs#L102) is broken.

kchibisov:
Wayland scales each window individually, thus make it clear. Also recommend against using the `MonitorHandle::scale_factor`.

Fixes #3183.

### DIFF
--- a/src/dpi.rs
+++ b/src/dpi.rs
@@ -82,7 +82,8 @@
 //!   If `WINIT_X11_SCALE_FACTOR` is set to `randr`, it'll ignore the `Xft.dpi` field and use the
 //!   XRandR scaling method. Generally speaking, you should try to configure the standard system
 //!   variables to do what you want before resorting to `WINIT_X11_SCALE_FACTOR`.
-//! - **Wayland:** Scale factor is suggested by the the compositor.
+//! - **Wayland:** Scale factor is suggested by the compositor for each window individually. The
+//!   monitor scale factor may differ from the window scale factor.
 //! - **iOS:** Scale factors are set by Apple to the value that best suits the device, and range
 //!   from `1.0` to `3.0`. See [this article][apple_1] and [this article][apple_2] for more
 //!   information.
--- a/src/monitor.rs
+++ b/src/monitor.rs
@@ -138,14 +138,18 @@ impl MonitorHandle {
         self.inner.refresh_rate_millihertz()
     }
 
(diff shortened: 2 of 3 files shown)


dhardy:
My thoughts:

- Windows, X11 and Orbital backends all use physical pixels to size windows. Given that, *if* there is a way to reliably determine the scale factor on these platforms before constructing the window, it would be useful to have a method on the ELWT to get that scale factor (as an `Option` which is `None` where not available).
- `MonitorHandle::scale_factor` is completely useless? The only other potential use I can think of is to try to set up a multi-screen view for a game/virtual world, but for that you really need the precise size and position instead of just a "scale factor", and that's generally not available (xrandr supports at least physical size, but the reported value is often wrong).

kchibisov:
>Windows, X11 and Orbital backends all use physical pixels to size windows. Given that, if there is a way to reliably determine the scale factor on these platforms before constructing the window, it would be useful to have a method on the ELWT to get that scale factor (as an Option which is None where not available).

They can try to get factor from the monitors on the event loop if they really want to, but you don't know where the window is located before creating it, so it doesn't work for these platform as well.

>MonitorHandle::scale_factor is completely useless? The only other potential use I can think of is to try to set up a multi-screen view for a game/virtual world, but for that you really need the precise size and position instead of just a "scale factor", and that's generally not available (xrandr supports at least physical size, but the reported value is often wrong).

Yes, it's useless API.

>In part it comes down to whether Winit is primarily a portable API or just a cross-platform compatibilty layer (i.e. how much platform-specific code users need on top of Winit).

If they use just `Window::scale_factor` and create with logical pixels, then zero.

dhardy:
> If they use just Window::scale_factor and create with logical pixels, then zero.

This would work better if it's possible to create windows without a size at all.

kchibisov:
>This would work better if it's possible to create windows without a size at all.

Yes, but winit can't do something like that with a reliable system to `callback to user to get values from it`.

Issue #1578, opened 2020-05-24
Events for when window is maximized/minimized


sunjay:
There is currently no `WindowEvent` for when a user clicks the maximized/minimized buttons. It would be great if some could be added since there is currently no way to detect that otherwise. You can set the maximized state using `set_maximized`, but you have no way of knowing when the user does that themselves. You could probably hack around it using the `Resized` event or something, but I think it makes more sense for that to have its own variant in the `WindowEvent` enum.

I'm not sure if something similar makes sense for the window fullscreen state, but an event should be added for that too if it is possible for the user to change that themselves.

kchibisov:
moving to winit, since it's something it should do, if it'll be accepted.

repi:
We would want this as well, to be able to persist a setting that the app window was maximized so we can start up with that state

rhysd:
I want `Minimized` event and `Restored` event for my practical use case.

I'm using WebView2 for my application and WebView2 provides a way to reduce memory consumption on inactive state.

- https://github.com/MicrosoftEdge/WebView2Feedback/blob/main/specs/MemoryUsageTargetLevel.md
- https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2.memoryusagetargetlevel?view=webview2-dotnet-1.0.2045.28

To determine when application goes inactive and when application goes back to active state, I need to receive the events. When `Minimized` is received, I can determine the application is inactive and when `Restored` is received, I can determine the application is back to active.

PR #3200, opened 2023-10-30
Implement `WindowEvent::Minimized` and `WindowEvent::Restored` for Windows

rhysd:
  - I only tested this change on Windows since the implementation affects only Windows

This PR partially implements #1578 by adding new window events `WindowEvent::Minimized` and `WindowEvent::Restored`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ Unreleased` header.
 
 # Unreleased
 
+- On Windows, implement `WindowEvent::Minimized` and `WindowEvent::Restored`.
+
 # 0.29.3
 
 - On Wayland, apply correct scale to `PhysicalSize` passed in `WindowBuilder::with_inner_size` when possible.
--- a/examples/util/fill.rs
+++ b/examples/util/fill.rs
@@ -53,6 +53,13 @@ pub(super) fn fill_window(window: &Window) {
     }
 
     GC.with(|gc| {
+        let size = window.inner_size();
+        let (Some(width), Some(height)) =
+            (NonZeroU32::new(size.width), NonZeroU32::new(size.height))
+        else {
+            return;
+        };
+
         // Either get the last context used or create a new one.
         let mut gc = gc.borrow_mut();
         let surface = gc
@@ -61,13 +68,9 @@ pub(super) fn fill_window(window: &Window) {
 
         // Fill a buffer with a solid color.
         const DARK_GRAY: u32 = 0xFF181818;
-        let size = window.inner_size();
 
         surface
-            .resize(
-                NonZeroU32::new(size.width).expect("Width must be greater than zero"),
-                NonZeroU32::new(size.height).expect("Height must be greater than zero"),
-            )
+            .resize(width, height)
             .expect("Failed to resize the softbuffer surface");
 
         let mut buffer = surface
--- a/src/event.rs
+++ b/src/event.rs
@@ -591,6 +591,20 @@ pub enum WindowEvent {
     /// Winit will aggregate duplicate redraw requests into a single event, to
     /// help avoid duplicating rendering work.
     RedrawRequested,
+
+    /// Emitted when a window is minimized.
+    ///
+    /// ## Platform-specific
+    ///
+    /// - Only available on **Windows**.
+    Minimized,
+
+    /// Emitted when a window is restored from minimized state.
+    ///
+    /// ## Platform-specific
+    ///
+    /// - Only available on **Windows**.
+    Restored,
 }
 
 /// Identifier of an input device.
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -964,6 +964,19 @@ unsafe fn lose_active_focus<T>(window: HWND, userdata: &WindowData<T>) {
     });
 }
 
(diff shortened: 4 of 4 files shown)


daxpedda:
See #2929. I'm not sure how many attempts this is now.

rhysd:
I haven't noticed the PR though I tried searching PRs with some words. I have two questions about the PR:

- How does it cover the `Restored` event in this PR?
- Is the PR active? There has been no comment for 4 months.

daxpedda:
> - How does it cover the `Restored` event in this PR?

The `WindowState` tells you if the `Window` is minimized or not. You will receive a `WindowState` every time anything changes.

> - Is the PR active? There has been no comment for 4 months.

Apparently not. I was just leaving it here for reference. E.g. the way you implemented it is not how we used to intend to implement it. I would prefer to go down the route of something more comprehensive like `WindowState` instead of creating more and more events. But I also don't want to hold development because we don't have the manpower to finish up PRs more comprehensive PRs like that.

---

Let's just wait and see what other maintainers have to say about this.
Imo we should just proceed here, if/when this merges, I will also make a PR to add Web support. We can always re-discuss merging this events into something like `WindowState` later, instead of having stale PRs.

kchibisov:
I think adding minimized events is not really a good idea. We already have `Occluded` if it matters and it should  be used to manipulate rendering instead. Minimized is just misleading most of the time and it doesn't mean that you shouldn't draw anymore.

daxpedda:
> I think adding minimized events is not really a good idea. We already have `Occluded` if it matters and it should be used to manipulate rendering instead.

We had a similar discussion before here: https://github.com/rust-windowing/winit/pull/2980.

I still don't understand why this remains a bad idea. I completely agree that `Occluded` should be used to figure out rendering, but these two things are not the same. If a window is minimized or not is completely separate information that might still be useful for users to know.
E.g. some programs might stop running when they are minimized, but will continue running if they are just in the background.
> Minimized is just misleading most of the time and it doesn't mean that you shouldn't draw anymore.

Would you mind elaborating on that?

kchibisov:
>Would you mind elaborating on that?

Minimize doesn't mean that you don't have to render, because compositor could show a preview. Like the preview you have when however over application in windows bar.

daxpedda:
> > Would you mind elaborating on that?
> 
> Minimize doesn't mean that you don't have to render, because compositor could show a preview. Like the preview you have when however over application in windows bar.

Yeah, that's a more complicated problem though and honestly makes information like that even more important. If we don't intend to send `Occluded` when a window is minimized (which I find needs some discussion), it is still good to know as a user that the window is minimized, e.g. pause the game, render at lower FPS.

kchibisov:
>Yeah, that's a more complicated problem though and honestly makes information like that even more important.
If we don't intend to send Occluded when a window is minimized (which I find would need some discussion), it is still good to know as a user that the window is minimized, e.g. pause the game, render at lower FPS.
If we do want to send Occluded, users might still want to know if the window is minimized for the reasons pointed out above.

The thing is that compositor can send `Occluded` for minimize as well, but then it'll send `Occluded(false)` when you try to preview.

daxpedda:
> The thing is that compositor can send `Occluded` for minimize as well, but then it'll send `Occluded(false)` when you try to preview.

Ah alright, nice!

kchibisov:
The thing is that only macOS/Wayland has this event delivered reliably iirc. On Web it should be as well. And that's the only event you can trust to pause rendering.

daxpedda:
> The thing is that only macOS/Wayland has this event delivered reliably iirc. On Web it should be as well. And that's the only event you can trust to pause rendering.

Sure, I mean we can add this kind of information in the documentation. I'm surprised this doesn't work reliably on Windows (though I haven't used Windows in a very long time now).

Pausing rendering should remain in `Occluded` anyway, even if we send a `Minimized` event, we should document that this doesn't mean the user should pause rendering, we should send an `Occluded` event for that separately.

kchibisov:
>even if we send a Minimized event, we should document that this doesn't mean the user should pause rendering

yeah, but we have this sort of information on `is_minimized` already, you just don't have a notification when you're getting minimized, though you can still sort of check during event loop handling whether you're minimized or not.

I just don't want to add more events...

rhysd:
@daxpedda @kchibisov Thank you for the further comments. I understood the situation and the `Occluded` event sounds sufficient for my use case. I'll wait on the PR.

kchibisov:
`Occluded` works on web.

kchibisov:
Ah, no, you were adding for `Windows`. Windows doesn't have a notion of `Occluded`...

rhysd:
Ah, I see. Then I think I need to wait for #2929.

rhysd:
I may not get the discussion well. Should I modify this branch as follows?

- send `Occluded(true)` on `wparam == SC_MINIMIZE`
- send `Occluded(false)` on `wparam == SC_RESTORE`

kchibisov:
No, you shouldn't. `Minimize != Occluded`. 

You need something like that https://learn.microsoft.com/en-us/windows/win32/api/dxgi1_2/nf-dxgi1_2-idxgifactory2-registerocclusionstatusevent or https://learn.microsoft.com/en-us/windows/win32/direct3ddxgi/waiting-when-occluded

PR #3203, opened 2023-10-31
🐉 Support window dragging from outside the main thread

Jasper-Bekkers:
On Windows, at least, `ReleaseCapture` is documented to release the capture of the mouse on the current thread. In our application we were calling `drag_window` on another thread because we've separated the event loop out to a different thread.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ Unreleased` header.
 
 # Unreleased
 
+- On Windows, fix so `drag_window` and `drag_resize_window` can be called from another thread.
+
 # 0.29.3
 
 - On Wayland, apply correct scale to `PhysicalSize` passed in `WindowBuilder::with_inner_size` when possible.
--- a/src/platform_impl/windows/window.rs
+++ b/src/platform_impl/windows/window.rs
@@ -472,27 +472,41 @@ impl Window {
     }
 
     unsafe fn handle_os_dragging(&self, wparam: WPARAM) {
-        let points = {
-            let mut pos = unsafe { mem::zeroed() };
-            unsafe { GetCursorPos(&mut pos) };
-            pos
-        };
-        let points = POINTS {
-            x: points.x as i16,
-            y: points.y as i16,
-        };
-        unsafe { ReleaseCapture() };
+        let window = self.window.clone();
+        let window_state = self.window_state.clone();
 
-        self.window_state_lock().dragging = true;
+        self.thread_executor.execute_in_thread(move || {
+            {
+                let mut guard = window_state.lock().unwrap();
+                if !guard.dragging {
+                    guard.dragging = true;
+                } else {
+                    return;
+                }
+            }
 
-        unsafe {
-            PostMessageW(
-                self.hwnd(),
-                WM_NCLBUTTONDOWN,
-                wparam,
-                &points as *const _ as LPARAM,
-            )
-        };
+            let points = {
+                let mut pos = unsafe { mem::zeroed() };
+                unsafe { GetCursorPos(&mut pos) };
+                pos
+            };
+            let points = POINTS {
+                x: points.x as i16,
+                y: points.y as i16,
+            };
+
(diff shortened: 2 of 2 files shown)


src/platform_impl/windows/window.rs
```diff
@@ -472,27 +472,39 @@ impl Window {
...
 
-        unsafe {
-            PostMessageW(
-                self.hwnd(),
-                WM_NCLBUTTONDOWN,
-                wparam,
-                &points as *const _ as LPARAM,
-            )
-        };
+        let (tx, rx) = channel::<()>();
```
kchibisov, line 477: Do we really need to do a `roundtrip`?

Jasper-Bekkers: The function was blocking before, I kept it as-is, seemed sensible.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ Unreleased` header.
 - On Wayland, use `configure_bounds` to constrain `with_inner_size` when compositor wants users to pick size.
 - On Windows, fix deadlock when accessing the state during `Cursor{Enter,Leave}`.
 - On Windows, add support for `Window::set_transparent`.
+- On Windows, fix so `drag_window` and `drag_resize_window` can be called from another thread.
```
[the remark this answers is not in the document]
Jasper-Bekkers: Nice catch.

src/platform_impl/windows/window.rs
kchibisov: yeah, it's just other functions are non-blocking doing the same things, in general, I think windows backend lacks the policy when doing so, so I'd prefer to either we change all functions to be blocking or we just use what we have now and don't block.

Wayland for example is also non-blocking.

src/platform_impl/windows/window.rs
```diff
@@ -472,27 +472,40 @@ impl Window {
...
-        };
-        unsafe { ReleaseCapture() };
+        {
+            let mut guard = self.window_state_lock();
+            if !guard.dragging {
+                guard.dragging = true;
+            } else {
+                return;
+            }
+        }
```
kchibisov, line 482: Hm, but we're not really dragging here, since we start drag on the main thread. The other code uses `window` and passes `Arc` to state if it matters (see `ime` stuff), I'd suggest to follow it.

To deny drag I think you want to check inside the callback as well, since the callbacks are executed sequentially on the main thread, thus it should correctly suppress when the drag is started.

Jasper-Bekkers: @kchibisov Could you point me a little more concretely? This code was setting this `dragging` boolean before (and I wanted to avoid setting it several times in a row / triggering the callback lots of times and re-starting the drag event).

It's also the only case in which this `dragging` boolean is used from what I can tell, so probably it's just misnamed?

Jasper-Bekkers: Are you suggesting moving this logic into the callback?

kchibisov: yes, because `main loop` has a logic to run on the `dragging`, so you should set the variable from the main loop as well the same as the old code did. Just ensure that it's not dragging, I guess.

So, move and check everything inside the callback.

Jasper-Bekkers: I see what you mean; I had missed that logic. Updated the code to reflect this :-).

src/platform_impl/windows/window.rs
```diff
@@ -472,27 +472,41 @@ impl Window {
...
-            let mut pos = unsafe { mem::zeroed() };
-            unsafe { GetCursorPos(&mut pos) };
-            pos
-        };
-        let points = POINTS {
-            x: points.x as i16,
-            y: points.y as i16,
-        };
-        unsafe { ReleaseCapture() };
+        let hwnd = self.hwnd();
```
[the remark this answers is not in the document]
Jasper-Bekkers: Updated to reflect this, makes sense to keep this consistent.

msiglreith (approved):

PR #3205, opened 2023-10-31
Fix typo in pre_present_notify docs

DevJac:
Fix typo and other small grammar corrections

### DIFF
--- a/src/window.rs
+++ b/src/window.rs
@@ -602,7 +602,7 @@ impl Window {
 
     /// Notify the windowing system before presenting to the window.
     ///
-    /// You should call this event after you've completed drawing operations, but before you submit
+    /// You should call this event after your drawing operations, but before you submit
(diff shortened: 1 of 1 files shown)


src/window.rs
```diff
@@ -600,11 +600,11 @@ impl Window {
         self.window.maybe_queue_on_main(|w| w.request_redraw())
     }
 
-    /// Notify the windowing system that you're before presenting to the window.
+    /// Notify the windowing system before presenting to the window.
     ///
-    /// You should call this event after you've done drawing operations, but before you submit
+    /// You should call this event after you've completed drawing operations, but before you submit
```
kchibisov, line 605: completed is a bit weird since you could think that you need to complete them actually, but drawing is async in general and you don't need to wait.

But I don't really care.

kchibisov (approved):

kchibisov:
Thanks.

Next time request a review on github if you want me to re-look into changes.

PR #3212, opened 2023-11-06
Add option to enable/disable `WS_CLIPCHILDREN` window style

amrbashir:
While working on https://github.com/tauri-apps/wry/pull/1041 which allows adding the webview as a child window, I noticed that when setting the webview to be transparent, the portion of the parent behind it doesn't render properly and that was because of `WS_CLIPCHILDREN`


kchibisov (approved):

notgull:
Why was `WS_CLIPCHILDREN` added in the first place? I'm at the gym right now so I can't do an in depth blame.

msiglreith (changes requested):
This will break child rendering in non-special cases. The parent window can now overdraw the child window.

amrbashir:
Would it make more sense to make this configurable? and what should the default behavior be?

kchibisov:
If that will be the resolution, I'd probably suggest to put this thing on _hold_, until better days/0.30.0.

msiglreith:
Thanks, `WS_CLIPCHILDREN` should be enabled by default.

amrbashir:
alright, should be good to go now.

msiglreith (approved):

Issue #3215, opened 2023-11-07
Windows: set_control_flow has no effect from AboutToWait event


ogoffart:
Consider this modified version of the `timer.rs` example:

```
#![allow(clippy::single_match)]

use std::time::Duration;

use simple_logger::SimpleLogger;
use winit::{
    event::{Event, WindowEvent},
    event_loop::{ControlFlow, EventLoop},
    window::WindowBuilder,
};

#[path = "util/fill.rs"]
mod fill;

fn main() -> Result<(), impl std::error::Error> {
    SimpleLogger::new().init().unwrap();
    let event_loop = EventLoop::new().unwrap();

    let window = WindowBuilder::new()
        .with_title("A fantastic window!")
        .build(&event_loop)
        .unwrap();

    event_loop.run(move |event, elwt| {
        println!("{event:?}");

        match event {
            // New iteration: reset the control flow to Wait as we don't know if we will need to wait
            Event::NewEvents(_) => elwt.set_control_flow(ControlFlow::Wait),
            // ... some other event change internal data structures ...
            Event::AboutToWait => {
                // About to wait:  query internal data structure to know how long we need to sleep
                elwt.set_control_flow(ControlFlow::wait_duration(Duration::from_secs(1)));
            }

            Event::WindowEvent {
                event: WindowEvent::CloseRequested,
                ..
            } => elwt.exit(),
            Event::WindowEvent {
                event: WindowEvent::RedrawRequested,
                ..
            } => {
                fill::fill_window(&window);
            }
            _ => (),
        }
    })
}
```

The gist here is that we set ControlFlow::Wait from NewEvents.  
In our application some events such as mouse or keyboard event may start timers in a timer or actions. So at the end of the iteration, in AboutToWait,  we set the control flow to the computed waiting time.

That works on Linux: we get new event every second from that example.
But on Windows, this blocks forever unless you move the mouse or something.

I've tried debugging this quickly, and it seems like the `wait_and_dispatch_message` function queries the control_flow here:
https://github.com/rust-windowing/winit/blob/c89e6df758652443db9d0b6e1e2af2b35cd3af8e/src/platform_impl/windows/event_loop.rs#L363
And then sends the AboutToWait  shortly after via
https://github.com/rust-windowing/winit/blob/c89e6df758652443db9d0b6e1e2af2b35cd3af8e/src/platform_impl/windows/event_loop.rs#L416

I believe this line should be moved before the line 363. Is that right?

[...]

PR #3216, opened 2023-11-07
Windows: move prepare_wait before querying the control_flow

ogoffart:
In case the AboutToWait event sets the control flow to another value

Fixes https://github.com/rust-windowing/winit/issues/3215

### DIFF
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -356,19 +356,6 @@ impl<T: 'static> EventLoop<T> {
 
     /// Wait for one message and dispatch it, optionally with a timeout
     fn wait_and_dispatch_message(&mut self, timeout: Option<Duration>) {
-        let start = Instant::now();
-
-        let runner = &self.window_target.p.runner_shared;
-
-        let control_flow_timeout = match runner.control_flow() {
-            ControlFlow::Wait => None,
-            ControlFlow::Poll => Some(Duration::ZERO),
-            ControlFlow::WaitUntil(wait_deadline) => {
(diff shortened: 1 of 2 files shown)


src/platform_impl/windows/event_loop.rs
```diff
@@ -360,6 +360,17 @@ impl<T: 'static> EventLoop<T> {
...
+        // observer that will dispatch AboutToWait when about to wait for
+        // events, and NewEvents after the RunLoop wakes up.
+        //
+        // We emulate similar behaviour by treating `GetMessage` as our wait
+        // point and wake up point (when it returns) and we drain all other
+        // pending messages via `PeekMessage` until we come back to "wait" via
+        // `GetMessage`
+        //
+        runner.prepare_wait();
+
```
kchibisov, line 373: I don't quite like the way it's structured in this function. Could you instead move the `timeout` calculation to the place it's actually being used? So basically revert your code and move the `let control_flow_timeout` part.

kchibisov (changes requested):

kchibisov (approved):

src/platform_impl/windows/event_loop.rs
ogoffart: done.  Also moved the computation of start nearer to where it is used. This is not strictly equivalent as time can pass while prepare_wait is being called. But i think this is fine.

kchibisov:
You've tested that it fixes your issue, right?

ogoffart:
> You've tested that it fixes your issue, right?

Yes

msiglreith (approved):

Issue #3198, opened 2023-10-28
Xlib must die


notgull:
Xlib is the legacy strategy for interacting with the X server. It has since been replaced by libxcb, which is lighter and has fewer soundless holes. [`x11rb`](https://github.com/psychon/x11rb) wraps around `libxcb` and reimplements the protocol parsing logic in safe Rust, further reducing the number of places where unsound logic can occur.

List of places in `winit` where Xlib still lurks:

- Event handling and IME (fixed in https://github.com/rust-windowing/winit/pull/3122)
- Cursor handling code (could use https://github.com/esposm03/xcursor-rs instead)
- `SelectXkbEvents` (cc https://github.com/rust-windowing/winit/pull/3087)
- [Keymap querying](https://github.com/rust-windowing/winit/blob/3c9f9da19e8d4ac5d19828717e5370af7a2f6c12/src/platform_impl/linux/x11/util/keys.rs#L69)

Discussion questions:

- I want to be able to replace `x11-dl` with [`tiny-xlib`](https://github.com/notgull/tiny-xlib), is this possible?
- Do we want to add a feature to remove `Xlib` entirely and just use XCB?
- Going further, do we want to add a feature to remove these two external libraries from our X backend altogether and just use `x11rb`'s pure-Rust backend?
  - Is this also available for Wayland?

kchibisov:
>Cursor handling code

isn't cursor being drawn on the server? Or we must submit a `Xcursor` theme to the server? How does it all work? I thought you just send names, it's not wayland with its client side cursors in the end of the day.

>I want to be able to replace x11-dl with [tiny-xlib](https://github.com/notgull/tiny-xlib), is this possible?

doesn't matter, but we don't need x11-dl that hard.

>Do we want to add a feature to remove Xlib entirely and just use XCB?

yes

>Is this also available for Wayland?

no.

notgull:
> > Is this also available for Wayland?
> 
> no.

Can you elaborate on this? I know that `wayland-rs` has a pure-Rust implementation available.

kchibisov:
>Can you elaborate on this? I know that wayland-rs has a pure-Rust implementation available.

any API wants a `wl_proxy` from libwayland, so unless you write a EGL/Vulkan platform to use whatever wayland-rs provides, this won't change.

bjorn3:
[...]

Issue #2875, opened 2023-06-11
Browser support tracking


daxpedda:
## [`Element.requestFullscreen`](https://developer.mozilla.org/en-US/docs/Web/API/Element/requestFullscreen) returns `Promise`

[Source](https://github.com/rust-windowing/winit/blob/fc046add78f676118886ee6ab00684913a2895e1/src/platform_impl/web/web_sys/canvas.rs#L426-L439)
Relevant PR: #2751
[MDN Compatibility Table](https://developer.mozilla.org/en-US/docs/Web/API/Element/requestFullscreen#browser_compatibility)
[Caniuse](https://caniuse.com/mdn-api_element_requestfullscreen_returns_promise)

Requires Safari v16.4.
See [Apple's documentation for supported devices](https://support.apple.com/en-us/HT213411).

## [`MediaQueryList`](https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList) inherits `EventTarget`

[Source](https://github.com/rust-windowing/winit/blob/fc046add78f676118886ee6ab00684913a2895e1/src/platform_impl/web/web_sys/media_query_handle.rs#L23-L28)
Relevant PR: #2850.
[MDN Compatibility Table](https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList#browser_compatibility)
[Caniuse](https://caniuse.com/mdn-api_mediaquerylist_change_event)

Requires Safari v14.
See [Apple's documentation for supported devices](https://support.apple.com/en-us/HT211348).

## [`resolution`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/resolution) media feature

[Source](https://github.com/rust-windowing/winit/blob/fc046add78f676118886ee6ab00684913a2895e1/src/platform_impl/web/web_sys/scaling.rs#L56-L61)
[MDN Compatibility Table](https://github.com/rust-windowing/winit/blob/fc046add78f676118886ee6ab00684913a2895e1/src/platform_impl/web/web_sys/scaling.rs#L56-L61)
[Caniuse](https://caniuse.com/css-media-resolution)

Requires Safari v16.
See [Apple's documentation for supported devices](https://support.apple.com/en-us/HT213411).

## [`PointerEvent.getCoalescedEvents()`](https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/getCoalescedEvents) 

[...]

PR #3218, opened 2023-11-10
Custom cursor images for all desktop platforms

eero-lehtinen:
Issue: #3005

Hello.

There seems to be many PRs relating to this issue, but they don't include all platforms and for some reason lost steam. This PR again tries to make this feature happen, and does it for all desktop platforms (x11, wayland, macos, windows, web).

## Use case
I think the best user of this feature and the reason I'm doing this is Bevy and game engines in general. There non laggy hardware cursors with custom images are very important. Game devs also like their PNGs so supporting platform native cursor files is not that important, but I guess could be added too.

## UPDATE: API
Instead of the things below, we have chosen to make the simplest possible API, where we have a single function for setting the cursor, and it internally recreates whatever is needed to display the cursor every time.

## API 1
At first I tried to make a very simple api (in branch [custom-cursors](https://github.com/rust-windowing/winit/compare/master...eero-lehtinen:winit:custom-cursors?expand=1)) where the user "registers" cursors by calling a window method with a rgba vector and a u64 key. Then when the user want's to change the cursor, they would just use this key. This was simple and efficient because the window itself would handle making and deleting the cursors and switching between them.

This API didn't seem very rusty. I thought that there was a simpler api hiding in there so I tried to make another one, which is in this PR.

## API 2
This PR contains a (simpler?) api that allows users to create their own platform specific cursors that are held by smart pointers and automatically deallocated on drop. This way cursors can be swapped by just setting a single smart pointer with no registration step. **Update:** I moved this implementation to [custom-cursors2-old](https://github.com/rust-windowing/winit/compare/master...eero-lehtinen:winit:custom-cursors2-old?expand=1).

This works wonderfully with windows, macos and web, but linux is really annoying. First of all it seems that all linux cursors would need to support both x11 and wayland, because there is no way to know what kind of event loop they will be inserted into. Secondly there seems to be no way to use smart pointers with wayland, because it needs to have a reference to the wayland connection. So now on wayland we just use an rgba vector so the wayland buffer can be recreated every time the user want's to swap cursors. Linux really was a lot simpler and more efficient with API 1.

## API Conclusion for Me
So in the end I think that API 1 was better because it's just more efficient and simpler on linux. We should probably make the CursorImage struct platform dependent so it could be extended by different platforms. Also maybe we could use some kind of better handle compared to just u64? 

What do you guys think about these apis?

I chose this branch (API 2) to show my thought process, but we could also continue the other branch.

## Scaling
Currently screen scaling isn't handled. Different platforms have different interactions.

- web: seems to upscale cursor automatically, kinda annoying
- x11: no scaling (meaning that if screen scale in increased, the cursor becomes smaller)
- wayland: no scaling. We own the buffer so upscaling is easy, but results in a blurry cursor
- windows: no scaling
- macos: my machine has no option for screen scaling so couldn't test :(

So in general we could solve this just by accepting an array of different sized images that could be used for different scales. Dunno what to do about the web and it's automatic scaling though.

## Other Issues
- I guess memory gets leaked in API 1 for most platforms when the window is dropped and and only on wayland with API 2, but should be pretty easily fixable. **Update**: should be fixed.
- I'm not sure if I'm allowed to destroy wl_buffers in the way I currently do it because according to [doc](https://wayland-client-d.dpldocs.info/wayland.client.protocol.wl_surface_attach.html) we would need to wait for some `wl_buffer.release` event, but I don't know much about that. Also seems to work without issues in practice.

Even if this PR doesn't make it, at least here I show how to create cursors for all platforms :).

## Checklist


src/platform_impl/web/custom_cursor.rs
```diff
@@ -0,0 +1,56 @@
...
+impl WebCustomCursor {
+    pub fn from_rgba(
+        rgba: Vec<u8>,
+        width: u32,
+        height: u32,
+        hotspot_x: u32,
+        hotspot_y: u32,
+    ) -> Result<Self, BadCursor> {
+        #[allow(clippy::disallowed_methods)]
+        let cursor_icon_canvas = web_sys::window()
```
daxpedda, line 22: These functions are disallowed for a reason (the reason is also stated in the warning).
Please pass these through appropriately.

src/platform_impl/web/custom_cursor.rs
daxpedda, line 1: This file should probably be split into an interfaces here and it's implementation in `web_sys`, but considering that this whole split should be removed, it's fine.

src/platform_impl/web/custom_cursor.rs
```diff
@@ -0,0 +1,56 @@
...
+
+        let context = cursor_icon_canvas
+            .get_context("2d")
+            .unwrap()
+            .unwrap()
+            .dyn_into::<CanvasRenderingContext2d>()
+            .unwrap();
+
+        let image_data =
+            web_sys::ImageData::new_with_u8_clamped_array_and_sh(Clamped(&rgba), width, height);
```
daxpedda, line 44: This is not supported when the backing memory is a `SharedArrayBuffer`, which is e.g. mandatory when using `atomics`.

See https://github.com/rust-windowing/softbuffer/blob/ab7688e2ed2e2eca51b3c4e1863a5bd7fe85800e/src/web.rs#L196-L223 for some inspiration how to handle this correctly.

src/platform_impl/web/custom_cursor.rs
```diff
@@ -0,0 +1,56 @@
...
+
+        context
+            .put_image_data(&image_data.unwrap(), 0.0, 0.0)
+            .unwrap();
+
+        let data_url = cursor_icon_canvas.to_data_url().unwrap();
```
daxpedda, line 50: The data URL should be stored in a wrapper, so we can call `URL.revokeObjectURL()` when we don't need it anymore.

src/platform_impl/web/web_sys/mod.rs
```diff
@@ -171,6 +171,11 @@ pub fn set_canvas_style_property(raw: &HtmlCanvasElement, property: &str, value:
         .unwrap_or_else(|err| panic!("error: {err:?}\nFailed to set {property}"))
 }
 
+pub fn get_canvas_style_property(raw: &HtmlCanvasElement, property: &str) -> Option<String> {
+    let style = raw.style();
+    style.get_property_value(property).ok()
+}
```
daxpedda, line 177: We don't do this anymore because we cache the `CssStyleDeclaration`.

daxpedda (changes requested):

src/platform_impl/web/window.rs
```diff
@@ -195,10 +208,26 @@ impl Inner {
 
     #[inline]
     pub fn set_cursor_icon(&self, cursor: CursorIcon) {
-        *self.previous_pointer.borrow_mut() = cursor.name();
+        *self.selected_cursor.borrow_mut() = SelectedCursor::BuiltIn(cursor);
+        if let Some(s) = backend::get_canvas_style_property(self.canvas.borrow().raw(), "cursor") {
+            if s == "none" {
+                return;
+            }
```
daxpedda, line 215: This is new behavior, which should probably be documented.
Could you explain the thought behind it though? It seems fine to me to overwrite any set cursors if the user wants to. It's explicit after all.

eero-lehtinen: I think the issue was that setting a new cursor was overriding the hidden status, which is different from other platforms. It could be another issue though because it was broken before this pr.

daxpedda: Ah, if that's the case we should probably use internal information to determine that instead of reading the CSS style. But preferably just leave it out of this PR and open a separate issue.

src/platform_impl/web/custom_cursor.rs
eero-lehtinen: I would really love to have an unified constructor for all platforms. Passing in extra stuffy only for web would be really annoying. If we need to do that, then the interface seems really inconvinient for web in addition to linux.

daxpedda: I don't think this necessarily changes the public API, it just complicates the API internally.
You will have to split construction in two: one is the public part, which doesn't do much apart from verifying sizes and such. The other part actually builds the cursor and data URL when `CustomCursor` is eventually passed to `Window::set_custom_cursor()`.

The way it's done right now is already an issue exactly because of what the warning was added to prevent: any user calling `CustomCursor` outside the main thread will experience a panic completely circumventing the wire implementation we have in place exactly for these kind of reasons.

eero-lehtinen: So what you are basically saying is that canvas images can only be created in the window struct. Like here in the other branch. https://github.com/eero-lehtinen/winit/blob/55e2cbb4d569d948f964e66b71e52ecf6ef56725/src/platform_impl/web/window.rs#L225. The reason I originally made it that way was that it seemed too complicated to to let users create cursors, which seems to be confirmed now.

daxpedda: `HTMLCanvasElement` can only be created in the Web Window, that is the reason why we disallow creating a Web `Window` to prevent issues like that. It can be done without the Winit `Window` of course, but you will have to implement thread detection logic.

I think the way forward for now is indeed what you have done in this other branch.

As a follow-up PR we can do the following:
- Create some sort of shared `Global` struct to de-duplicate thread detection logic. Use that to actually move the `Canvas` instantiation to `CustomCursor::from_rgba()`.
- Use `OffscreenCanvas` when available, which we need to do anyway when we are in a thread.
- (Bonus) Add a platform-specific extension to create custom icons by URL.

daxpedda:
@eero-lehtinen let me know if you need any help for the Web backend!

eero-lehtinen:
> @eero-lehtinen let me know if you need any help for the Web backend!

Thank you. I will continue working on this today and this week so I'll ask when things come up.

src/platform_impl/web/custom_cursor.rs
eero-lehtinen: Code moved to src/platform_impl/web/custom_cursor.rs. Now is called safely from the Window struct and uses document saved in it.

I also added the possibility to create url icons.

src/platform_impl/web/custom_cursor.rs
eero-lehtinen: If it's fine, I guess we can resolve this.

src/platform_impl/web/custom_cursor.rs
eero-lehtinen: I'm not sure if we have feature flags like that, so I changed it to always do the extra copies.

Also I don't understand what happens here so I basically copy pasted the reference code.

src/platform_impl/web/custom_cursor.rs
eero-lehtinen: I added this and it seems to work.

src/platform_impl/web/web_sys/mod.rs
eero-lehtinen: Yeah removed the whole thing because it was used for hidden detection and we decided to not fix that in this pr.

src/cursor.rs
```diff
@@ -0,0 +1,163 @@
...
+
+use crate::{icon::PIXEL_SIZE, platform_impl::PlatformCustomCursor};
+
+#[derive(Debug, Clone)]
+pub struct CustomCursor {
+    pub(crate) inner: PlatformCustomCursor,
+}
+
+impl CustomCursor {
+    pub fn from_image(image: CursorImage) -> Self {
```
daxpedda, line 14: Could we not just implement `fn from_rgba()` here directly?
No reason two introduce multiple types if it's not necessary.

src/window.rs
```diff
@@ -1346,6 +1347,18 @@ impl Window {
...
+    pub fn register_custom_cursor_icon(&self, key: u64, cursor: CustomCursor) {
+        self.window
+            .maybe_queue_on_main(move |w| w.register_custom_cursor_icon(key, cursor))
+    }
+
+    #[inline]
+    pub fn set_custom_cursor_icon(&self, key: u64) {
+        self.window
+            .maybe_queue_on_main(move |w| w.set_custom_cursor_icon(key))
+    }
```
daxpedda, line 1360: I see what you have done here. Definitely didn't think about this problem in our earlier discussion.
It does make sense though that switching between cursors needs to be supported somehow to prevent any unnecessary overhead.

This probably goes back to having to fully build a cursor in `CustomCursor` would alleviate this problem entirely, which is what you had before but had to rework because Web.

I think we should figure out a way without having to resort to IDs. The solution here is probably to revert to `CustomCursor` being fully built by the user so using `set_custom_cursor_icon()` has no overhead. For Web, the solution here is probably to just return an error when we are in the wrong thread (this will only affect old versions of Safari, because other browsers support `OffscreenCanvas` and can do this in any thread).

I apologize for the confusion here, I didn't look at all the changes you had in https://github.com/rust-windowing/winit/pull/3218#discussion_r1389988046 and didn't see that you actually changed the public API.

src/platform_impl/web/custom_cursor.rs
daxpedda: It's just a `target_feature` flag, it has nothing to do with crate features, so it can just work.
I'm afraid having to always do these copies is undesirable.

Let me worry about this, I can fix it later before merging!

src/platform_impl/web/custom_cursor.rs
```diff
@@ -0,0 +1,103 @@
...
+            type ImageDataExt;
+            #[wasm_bindgen(catch, constructor, js_class = ImageData)]
+            fn new(array: Uint8ClampedArray, sw: u32) -> Result<ImageDataExt, JsValue>;
+        }
+        let array = Uint8Array::new_with_length(image.rgba.len() as u32);
+        array.copy_from(&image.rgba);
+        let array = Uint8ClampedArray::new(&array);
+        let image_data = ImageDataExt::new(array, image.width)
+            .map(JsValue::from)
+            .map(ImageData::unchecked_from_js)
```
daxpedda, line 76: We should also add the `ImageData` crate feature to our `web-sys` dependency.

daxpedda (changes requested):
I think Web looks fine so far.

src/window.rs
daxpedda, line 1360: If I understand correctly API 2 didn't work out well because of the following:
- `CustomCursor` needs to know what backend it would be targeting for Linux (X11 or Wayland).
- Wayland needs access to the connection.
- Additionally now Web can't even create a `CustomCursor` without a `Window`. (potential solution discussed in https://github.com/rust-windowing/winit/pull/3218#discussion_r1390013504)

So I was thinking: could we just move `CustomCursor` creation to be a method on `Window`? E.g. `Window::create_custom_cursor(...) -> Result<CustomCursor>`, that way we can always determine the backend and Wayland can get all the access it needs. Also I was wondering if we could move this up to `EventLoopWindowTarget` instead of `Window`. Seems to me that a `CustomCursor` should work on any `Window`, at least on Web that would be nice.

That should do away with the ID system. WDYT?

---

Please don't do any changes yet, we need some consensus first before you do any unnecessary work.

src/cursor.rs
daxpedda: You should probably hold off on doing anything here as well if we want to use this as proposed in https://github.com/rust-windowing/winit/pull/3218#discussion_r1395537962.

src/platform_impl/web/custom_cursor.rs
eero-lehtinen: How does it build an work without issues then? But sure if it's needed

src/window.rs
eero-lehtinen: Yeah the current thing uses `maybe_queue_on_main` because I just naively used the same kind of code that was around. It would be a lot nicer to be able to actually return things and that could solve needing clunky ids. Though would threading issues persist when we aren't using this queue?

eero-lehtinen: Also dropping wayland cursors seems pretty hard if we don't have a global for the connection.

eero-lehtinen: It would be nice to reuse cursors between windows, so I have no issues with moving it to the `EventLoopWindowTarget`. I don't know that much about winit architecture but maybe that would work nicely in practice too.

src/cursor.rs
```diff
@@ -0,0 +1,163 @@
...
+pub(crate) enum SelectedCursor {
+    BuiltIn(CursorIcon),
+    Custom(u64),
+}
+
+impl Default for SelectedCursor {
+    fn default() -> Self {
+        SelectedCursor::BuiltIn(Default::default())
+    }
+}
```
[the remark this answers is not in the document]
eero-lehtinen: Doesn't work because 
```
the `#[default]` attribute may only be used on unit enum variants
consider a manual implementation of `Default`
```

src/platform_impl/web/custom_cursor.rs
daxpedda: Probably because some dependency enables it. But if that dependency changes we will be affected, so we don't want to rely on that.

src/cursor.rs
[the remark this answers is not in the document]
daxpedda: Aye, thanks for trying!

src/window.rs
daxpedda: > Yeah the current thing uses `maybe_queue_on_main` because I just naively used the same kind of code that was around. It would be a lot nicer to be able to actually return things and that could solve needing clunky ids.

I think you are just looking for `maybe_wait_on_main()`, which is like `maybe_queue_on_main()`, but can actually return stuff.

> Also dropping wayland cursors seems pretty hard if we don't have a global for the connection.

Just spitballing here, but I was thinking that we could do some fun trickery with `Weak`, on `Drop` we can then turn that `Weak` into whatever it's holding and then properly deallocate that. That way if the event loop is destroyed before the user drops the cursor we can still deallocate internally.

In any case, we will probably have to evaluate some different ideas here and do some experiments.
Let me know what you think!

> It would be nice to reuse cursors between windows, so I have no issues with moving it to the `EventLoopWindowTarget`. I don't know that much about winit architecture but maybe that would work nicely in practice too.

From Winit's perspective this should work, unless it's somehow connected to specific windows on some backend. `EventLoopWindowTarget` is also not `Send + Sync`, so that might simplify some things.

daxpedda:
@eero-lehtinen would appreciate a rebase!

eero-lehtinen:
@daxpedda Rebased

src/window.rs
eero-lehtinen: > I think you are just looking for `maybe_wait_on_main()`, which is like `maybe_queue_on_main()`, but can actually return stuff.

Seems to be what I needed from the start.

> Just spitballing here: but I was thinking that we could do some fun trickery with `Weak`, on `Drop` we can then turn that `Weak` into whatever it's holding and then properly deallocate that. That way if the event loop is destroyed before the user drops the cursor we can still deallocate internally.

I don't understand how that helps, you need to explain more. If we keep `Weak`s to cursors internally, then we can deallocate when event loop is dropped, but that would leave invalid cursors around. Also wouldn't help the problem of deallocating cursors dropped by the user when the event loops still exists.

> In any case, we will probably have to evaluate some different ideas here and do some experiments.
Let me know what you think!

A handle system could also work. In bevy, handles send drop events through a channel so the underlying storage knows to delete stuff ([StrongHandle](https://github.com/bevyengine/bevy/blob/bc9e159b2659435c4ca39eebd9b97396486161c2/crates/bevy_asset/src/handle.rs#L94-L101C2)). Though might be too annoying if its only really needed for wayland.

kchibisov (changes requested):
I appreciate your enthusiasm on this issue and thank you for the PR.

However I don't think I'll merge this approach in the current state, especially given how everything is more complex and inconsistent. But it's not like I'll do a release either even if you add those changes (unless someone really asks me to do that).

Like the API should be simple

```rust
pub fn set_cursor(Window, icon: CursorIcon { NamedCursorIcon, Custom }
```

And that's really all it should be, everything else is on the downstream when dealing with that. Yes it can do everything on its own just about fine. No special changes will be required for that on _any platform_ no fighting with Wayland, because it already does that.

I'll also write Wayland part myself, since clearly you never wrote wayland stuff.

Besides, when writing a massive patches, please, do ask in `#winit` or on github issue, if we'd like to cope through review now, because otherwise your patch could stall until a _good times_.

It doesn't mean that the code will be discarded, because internally you're doing pretty much everything what should be done, it'll just be ported to a different API or you can port to the one I've suggested, where you just _set cursor_ and don't do any registration, etc, because just setting it is more than enough (yes, client will just must reset it if they fill so).

daxpedda:
The main problem seems to be how to handle multiple cursors without having to re-initialize them every time. Unfortunately to implement some `Drop` mechanic it seems to require a connection handler not only on both Linux backends. Web is pretty straightforward fortunately, but that is only one backend (the one I reviewed).

@eero-lehtinen I believe following what @kchibisov has said here, this PR should be refactored into just a simple function on `Window` that sets the cursor and that's it. Handling multiple cursors with proper `Drop` mechanic is a topic for a future PR.

---

Some thoughts on that future PR: I think probably the best thing we can do here is as we discussed above: creating the cursor would require `EventLoopWindowTarget`. The `Drop` mechanic could be implemented in two ways for backends that require it:
- The resulting cursor would hold a `Weak` to the internal connection and table and on `Drop` delete the cursor from there.
- Send an internal event that can schedule deleting the cursor, which would be more like garbage collection.

After we finish and merge the first PR I will open an issue where we can discuss the details of that and if it's desirable in the first place.

kchibisov:
>Handling multiple cursors with proper Drop mechanic is a topic for a future PR.

If you do that it'll work properly on Wayland, since wayland already draws custom cursors, they are just hidden, but it already does everything the same. The same with X11. So everything is properly dropped, etc in the way I said that because it's already how it works.

For the future I wouldn't abuse event loop for it, I'd just callback to client to provide a new shape when DPI changes, and clients should cache them somehow.

src/window.rs
daxpedda: > Edit: Thought I guess if we give out `Weak`s, keep `Arc`s, and require destruction through a method on the event loop, things could work out, though inconveniently.

That is what I meant! It's the other way around you interpreted it :)

> A handle system could also work. In bevy, handles send drop events through a channel so the underlying storage knows to delete stuff ([StrongHandle](https://github.com/bevyengine/bevy/blob/bc9e159b2659435c4ca39eebd9b97396486161c2/crates/bevy_asset/src/handle.rs#L94-L101C2)). Though might be too annoying if its only really needed for wayland.

That is indeed the only second solution I can come up as well. In any case, as discussed below, we should move that to a follow-up.

eero-lehtinen:
@kchibisov 
> I'll also write Wayland part myself, since clearly you never wrote wayland stuff.

It's true, I don't know much about wayland. I will do the best baseline implementation I can and you guys can edit it to be the way you like.
 
> Besides, when writing a massive patches, please, do ask in `#winit` or on github issue, if we'd like to cope through review now, because otherwise your patch could stall until a _good times_.

Sorry. I just wanted a patch for myself and thought to post it here to see what you think. Anyway that didn't even work out because Bevy doesn't use 29 yet. I would have to port it to 28 to benefit from it right now.
 
> It doesn't mean that the code will be discarded, because internally you're doing pretty much everything what should be done, it'll just be ported to a different API or you can port to the one I've suggested, where you just _set cursor_ and don't do any registration, etc, because just setting it is more than enough (yes, client will just must reset it if they fill so).

I can port it to be that way. So basically always just a single function that always reallocates the cursor. My current solution tries to be more efficient but I guess creating a cursor even every frame won't be that taxing and will be simpler.

> You can look at pointer impl in it and on wayland-cursor crate.

Thanks for the pointers. I think I was already looking at that when gobbling my solution together.

kchibisov:
>My current solution tries to be more efficient but I guess creating a cursor even every frame won't be that taxing and will be simpler.

You don't need to change cursor unless you change it every frame to something new, like when you're animating it. Once you set it, it's just being used and set back by winit when needed.

eero-lehtinen:
> > My current solution tries to be more efficient but I guess creating a cursor even every frame won't be that taxing and will be simpler.
> 
> You don't need to change cursor unless you change it every frame to something new, like when you're animating it. Once you set it, it's just being used and set back by winit when needed.

Some games might want to change it every frame or at least call the function to make sure if there is no way to query what the current cursor is.

kchibisov:
Once you set the cursor it's winit job to make it the same. Programs doing stupid things should get what they ask for.

eero-lehtinen:
I don't think we disagree. The use case of changing custom cursors often exists and could be optimized for, but not in this pr.

kchibisov:
I mean, you surely can try to optimize, but you don't need to expose any of that to the user. Like you can't drop cursors easily on all platforms (you can't on Wayland, you either append or purge all of them).

src/cursor.rs
```diff
@@ -0,0 +1,171 @@
+use core::fmt;
+use std::{error::Error, sync::Arc};
+
+use crate::{icon::PIXEL_SIZE, platform_impl::PlatformCustomCursor};
+
+#[derive(Debug, Clone)]
+pub struct CustomCursor {
```
daxpedda, line 7: Please add a `Send + Sync` test for `CustomCursor` in https://github.com/rust-windowing/winit/blob/7bed5eecfdcbde16e5619fd137f0229e8e7e8ed4/tests/send_objects.rs and https://github.com/rust-windowing/winit/blob/7bed5eecfdcbde16e5619fd137f0229e8e7e8ed4/tests/sync_object.rs.

src/cursor.rs
```diff
@@ -0,0 +1,171 @@
...
+    pub(crate) inner: Arc<PlatformCustomCursor>,
+}
+
+impl PartialEq for CustomCursor {
+    fn eq(&self, other: &Self) -> bool {
+        Arc::ptr_eq(&self.inner, &other.inner)
+    }
+}
+
+impl Eq for CustomCursor {}
```
daxpedda, line 17: That can be added by the `derive` macro, no?

eero-lehtinen: PartialEq is special because it compares pointers so I don't think so. But now I noticed that I don't even ever compare these, I probably had some idea that didn't work. Do we even want Eq?

daxpedda: We do, it can be cloned and is user facing, so definitely helpful.
`PartialEq` needs a manual implementation yes, but `Eq` doesn't.

src/cursor.rs
```diff
@@ -0,0 +1,171 @@
...
+#[derive(Debug, Clone)]
+pub struct CustomCursor {
+    pub(crate) inner: Arc<PlatformCustomCursor>,
+}
+
+impl PartialEq for CustomCursor {
+    fn eq(&self, other: &Self) -> bool {
+        Arc::ptr_eq(&self.inner, &other.inner)
+    }
+}
```
daxpedda, line 15: Why did we need this anyway? Doesn't the underlying type support `Partial(Eq)` already?

src/cursor.rs
```diff
@@ -0,0 +1,171 @@
+use core::fmt;
+use std::{error::Error, sync::Arc};
+
+use crate::{icon::PIXEL_SIZE, platform_impl::PlatformCustomCursor};
+
+#[derive(Debug, Clone)]
```
daxpedda, line 6: Is `Clone` actually necessary? We could just take a reference in `Window::set_custom_cursor()`.

src/cursor.rs
eero-lehtinen: Implementing `PartialEq` manually and `Eq` with `derive` together doesn't build. So it seems we need that impl.

src/cursor.rs
```diff
@@ -0,0 +1,171 @@
...
+pub struct CustomCursor {
+    pub(crate) inner: Arc<PlatformCustomCursor>,
```
daxpedda, line 8: If we remove `Clone` I think the `Arc` isn't necessary either. Backends which need it, like Web, could add an `Arc` to their `PlatformCustomCursor`.

src/cursor.rs
daxpedda: Ah, I see, because it does some internal asserts over `PlatformCustomCursor`.
Alright then!

src/cursor.rs
eero-lehtinen: Now `PlatformCustomCursor` includes the rgba buffer, so I don't like copying it every time the user want's to set the same cursor multiple times just like in my example. `Clone` is necessary and it would be nice for it to be cheap.

eero-lehtinen: But I guess the user could also put it in an arc themselves. So whatever works for me.

daxpedda: See https://github.com/rust-windowing/winit/pull/3218#discussion_r1397911756.
My thought was if `Window::set_custom_cursor()` takes a reference, users don't have to clone and can keep it.

The assumption here is that backends actually only need a reference, right?

src/cursor.rs
eero-lehtinen: It says `error[E0521]: borrowed data escapes outside of method` so I think arc or something similar might be required.

src/cursor.rs
eero-lehtinen: My comment on the thread above indicates that arc is still needed.

src/cursor.rs
daxpedda: Try using `maybe_wait_on_main()` instead.

daxpedda: On second thought, it might be better to keep the `Arc` but still take a reference. So just take a reference from the user but then clone the underlying `Arc` inside `Window::set_custom_cursor()`, we can decide to remove the `Arc` later if we still think that's better.

src/cursor.rs
daxpedda: For documentation purpose so we can decide that later:
if we use `maybe_wait_on_main()` in `Window::set_custom_cursor()` we can remove the `Arc` if we want to. But I don't see a reason to do that right now considering `maybe_wait_on_main()` is strictly worse then `maybe_queue_on_main()` for the user.

src/cursor.rs
eero-lehtinen: The default Eq would compare the rgba buffers or whatever was inside. I think I needed this pointer comparison at some point but don't anymore. So I dunno if we want this anymore.

daxpedda: We should be able to do it via `derive` if the underlying type supports it now. Imo public types should always implement `Partial(Eq)` if possible.

src/cursor.rs
eero-lehtinen: Sound good

src/cursor.rs
eero-lehtinen: Done

src/cursor.rs
eero-lehtinen: Added normal PartialEq and Eq

src/cursor.rs
```diff
@@ -0,0 +1,163 @@
...
+}
+
+impl Error for BadImage {
+    fn source(&self) -> Option<&(dyn Error + 'static)> {
+        Some(self)
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct CursorImage {
```
daxpedda, line 115: Why is there `ImageCustomCursor` **and** `CursorImage`? `ImageCustomCursor` seems to be just a wrapper over `CursorImage` with no additional benefits. Could this not be just one type?

eero-lehtinen: Yeah it doesn't do anything

daxpedda:
@eero-lehtinen addressed the Web stuff here: https://github.com/eero-lehtinen/winit/pull/1.

eero-lehtinen:
@daxpedda Thank you very much. Looks great.

eero-lehtinen:
@kchibisov 

> Also, for wayland stuff, just do regular sw rendering pipeline expected by it. Look at SCTK examples how the draw software buffers and just create a Shm pool. Like this https://github.com/Smithay/client-toolkit/blob/master/examples/simple_window.rs#L139 should be fine, and then share this pool with every window, like it's already done internally with SCTK.

Now I updated the Wayland implementation to use a `SlotPool`. It's not shared with all windows because I'm not sure how to share it. Would it need a mutex or something because I can't clone it like the shm is cloned?

Earlier you said that you could write the Wayland part yourself. If this current thing isn't correct, that might be easier than trying to guide me.

src/cursor.rs
```diff
@@ -0,0 +1,150 @@
...
+        Ok(Self)
+    }
+}
+
+#[derive(Debug)]
+/// An error produced when using [`Icon::from_rgba`] with invalid arguments.
+pub enum BadImage {
+    /// Produced when the length of the `rgba` argument isn't divisible by 4, thus `rgba` can't be
+    /// safely interpreted as 32bpp RGBA pixels.
+    ByteCountNotDivisibleBy4 { byte_count: usize },
```
madsmtm, line 49: Consideration (that you may have already talked about, if so, I would like to see the reasoning be stated in a code comment):

`softbuffer` [uses](https://docs.rs/softbuffer/latest/softbuffer/struct.Surface.html#method.fetch) `u32` instead of `u8` in its buffer, exactly so that errors like this are entirely avoided.

src/cursor.rs
```diff
@@ -0,0 +1,150 @@
...
+            ),
+        }
+    }
+}
+
+impl Error for BadImage {
+    fn source(&self) -> Option<&(dyn Error + 'static)> {
+        Some(self)
+    }
+}
```
madsmtm, line 97: Fairly sure this is incorrect, and would lead to an infinite loop for consumers.

```suggestion
impl Error for BadImage {}
```

madsmtm (approved):
I haven't followed this enough to comment much on the surface API, but the macOS implementation by itself looks fine.

src/cursor.rs
eero-lehtinen: It doesn't have reasoning other than that I copied it from `src/icon.rs` [link](https://github.com/rust-windowing/winit/blob/7bed5eecfdcbde16e5619fd137f0229e8e7e8ed4/src/icon.rs#L20C26-L20C26).

I'm fine with using u32, but what about endianess? Currently it is using big endian RGBA pixels, which work directly only in web and macos. Windows, X11 and Wayland want little endian ARGB pixels and need extra processing currently.

src/cursor.rs
eero-lehtinen: I copied this too from [src/icon.rs](https://github.com/rust-windowing/winit/blob/7bed5eecfdcbde16e5619fd137f0229e8e7e8ed4/src/icon.rs#L52). So if that is incorrect then both should be.

madsmtm: Well, that type _does_ have "bad" in it's name, so I guess it makes sense that it's behaving badly.

Joking aside, yeah, I'm pretty sure that's wrong as well.

eero-lehtinen: Yeah it's clearly wrong now that I checked Rust docs.

examples/custom_cursors.rs
```diff
@@ -0,0 +1,93 @@
...
+                }
+                KeyCode::KeyS => {
+                    log::debug!("Setting cursor icon to default");
+                    window.set_cursor_icon(Default::default());
+                }
+                KeyCode::KeyD => {
+                    cursor_visible = !cursor_visible;
+                    log::debug!("Setting cursor visibility to {:?}", cursor_visible);
+                    window.set_cursor_visible(cursor_visible);
+                }
```
kchibisov, line 75: We generally use virtual keys and not scancodes. I have no idea where the scancode `S` is on my keyboard for example.

src/cursor.rs
kchibisov: @eero-lehtinen endianess is explicitly stated in the underlying pixel formats. On Wayland it's always little endian, regardless of used arch, maybe the same on X11.

I'd suggest to follow softbuffer approach to buffers here, since it's known to work and using u32 avoids a lot of errors.

The users need to decode images anyway, and they need to ensure that they have alpha properly in them, like they can't just pass data as is.

src/cursor.rs
```diff
@@ -0,0 +1,146 @@
...
+/// Platforms that don't support cursors will export this as `PlatformCustomCursor`.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) struct NoCustomCursor;
+
+#[allow(dead_code)]
+impl NoCustomCursor {
+    pub fn from_rgba(
+        rgba: Vec<u8>,
+        width: u32,
+        height: u32,
```
kchibisov, line 35: Do we really need a `width` and `height`? I'd just use `stride: u32`, since you can internally derive the rest from it.

src/platform_impl/linux/wayland/window/state.rs
```diff
@@ -702,6 +709,42 @@ impl WindowState {
...
+
+            let serial = pointer
+                .pointer()
+                .data::<WinitPointerData>()
+                .and_then(|data| data.pointer_data().latest_enter_serial())
+                .unwrap();
+
+            pointer
+                .pointer()
+                .set_cursor(serial, Some(surface), cursor.hotspot_x, cursor.hotspot_y);
```
kchibisov, line 744: hotspot is surface local and not buffer coordinates, you need to convert.

src/platform_impl/windows/icon.rs
```diff
@@ -160,3 +163,93 @@ pub fn unset_for_window(hwnd: HWND, icon_type: IconType) {
...
+
+    pub fn new(image: &CursorImage) -> Result<Self, io::Error> {
+        let mut bgra = image.rgba.clone();
+        bgra.chunks_exact_mut(4).for_each(|chunk| chunk.swap(0, 2));
+
+        let w = image.width as i32;
+        let h = image.height as i32;
+
+        let handle = unsafe {
+            let mask_bits: Vec<u8> = vec![0xff; ((((w + 15) >> 4) << 1) * h) as usize];
```
kchibisov, line 196: Could you specify the reasoning behind the encoding used here.

src/platform_impl/windows/icon.rs
```diff
@@ -160,3 +163,93 @@ pub fn unset_for_window(hwnd: HWND, icon_type: IconType) {
...
+            let hbm_mask = CreateBitmap(w, h, 1, 1, mask_bits.as_ptr() as *const _);
+            if hbm_mask == 0 {
+                return Err(io::Error::last_os_error());
+            }
+
+            let hdc_screen = GetDC(0);
+            if hdc_screen == 0 {
+                return Err(io::Error::last_os_error());
+            }
```
kchibisov, line 205: Don't we have DC around on the window already for stuff like that?

src/platform_impl/windows/icon.rs
```diff
@@ -160,3 +163,93 @@ pub fn unset_for_window(hwnd: HWND, icon_type: IconType) {
...
+
+#[derive(Debug)]
+struct RaiiCursor {
+    handle: HCURSOR,
+}
+
+#[derive(Clone, Debug)]
+pub struct WinCursor {
+    inner: Arc<RaiiCursor>,
+}
```
kchibisov, line 175: You don't have to nest like that, could just have a single `drop` which checks the ref counting stuff.

src/platform_impl/linux/x11/util/cursor.rs
```diff
@@ -1,12 +1,64 @@
-use std::ffi::CString;
-use std::iter;
+use core::slice;
```
kchibisov, line 1:
```suggestion
use std::slice;
```

src/platform_impl/linux/wayland/window/state.rs
```diff
@@ -197,6 +200,7 @@ impl WindowState {
             resizable: true,
             scale_factor: 1.,
             shm: winit_state.shm.wl_shm().clone(),
+            pool: SlotPool::new(4, &winit_state.shm).unwrap(),
```
kchibisov, line 203: The pool must be shared for all the windows and instantiated if and only if we actually try to set the custom cursor.

src/platform_impl/linux/wayland/types/cursor.rs
```diff
@@ -0,0 +1,72 @@
+use std::sync::Arc;
+
+use cursor_icon::CursorIcon;
+use rustix::fd::OwnedFd;
+use sctk::shm::slot::{Buffer, SlotPool};
+use wayland_backend::client::ObjectData;
+use wayland_client::protocol::wl_shm::Format;
```
kchibisov, line 7: We use `sctk` reexports and not `wayland_client/backend` directly.

src/platform_impl/linux/wayland/types/cursor.rs
```diff
@@ -0,0 +1,72 @@
...
+
+impl ObjectData for IgnoreObjectData {
+    fn event(
+        self: Arc<Self>,
+        _: &wayland_client::backend::Backend,
+        _: wayland_client::backend::protocol::Message<wayland_client::backend::ObjectId, OwnedFd>,
+    ) -> Option<Arc<dyn ObjectData>> {
+        None
+    }
+    fn destroyed(&self, _: wayland_client::backend::ObjectId) {}
```
kchibisov, line 71: Is this dead?

kchibisov (changes requested):

kchibisov:
>Now I updated the Wayland implementation to use a SlotPool. It's not shared with all windows because I'm not sure how to share it. Would it need a mutex or something because I can't clone it like the shm is cloned?

Yeah, something like that. You can always look at how regular cursors are done on Wayland. They are _all_ client side, so we already have custom cursors in that sense on Wayland.

>Earlier you said that you could write the Wayland part yourself. If this current thing isn't correct, that might be easier than trying to guide me.

It's not like you're that far from the solution and the _how to do it_ is already present in the regular cursor code, because Wayland used client side rendering for cursors originally. Just following the SCTK code should be more than enough here, they also use some sort of `pool` under the hood.

src/cursor.rs
daxpedda: I'm not particularly convinced that `softbuffer` has done the right thing here. Even though endianess seems to be defined for Wayland it isn't clear in other backends/OS's (AFAIK). I wrote up some of my thoughts here: https://github.com/rust-windowing/softbuffer/issues/109.

I'm strongly in favor of leaving it as is, as the same API is already used in `Icon`. We can change the API in a follow-up issue/PR where we can have that discussion and change both and not leave them differently.

eero-lehtinen: I would like to leave it as is. I already made it and it already works :D. 

The change would be annoying, because x11 is the only one that accepts u32, so all others would need to be manually converted to u8. This change would remove this single user facing error but not much else, I think. But I'll leave this up to you to choose.

src/cursor.rs
eero-lehtinen: Not really required, but works as a sanity check.

eero-lehtinen: But also I don't really care about it so I can change it.

kchibisov: `stride` and `len` of the supplied buffer will result in the exact same behavior but with one less argument, though, `Icon` uses the `height` and `width`, so we could keep them as is.

examples/custom_cursors.rs
eero-lehtinen: Changed to logical number keys

src/platform_impl/windows/icon.rs
eero-lehtinen: Directly from win [docs](https://learn.microsoft.com/en-us/windows/win32/api/wingdi/nf-wingdi-createbitmap#parameters). I can add a comment as a reminder.

eero-lehtinen: It's pretty weird that it does ">> 4" and "<< 1" but I just believe them

src/platform_impl/windows/icon.rs
eero-lehtinen: If we do I can't find it.

src/platform_impl/linux/wayland/types/cursor.rs
eero-lehtinen: Yes

eero-lehtinen:
I just noticed that I wasn't releasing everything on windows so I fixed that. Windows should be done overall.

src/platform_impl/windows/icon.rs
eero-lehtinen: True, unneeded indirection that was copied from `WinIcon`, but I simplified it anyway.

src/platform_impl/linux/wayland/window/state.rs
eero-lehtinen: Fixed

src/platform_impl/linux/wayland/window/state.rs
eero-lehtinen: Now it is shared but I was unable to instantiate it lazily. The window doesn't have access to the actual `Shm` that is needed to create the pool. It has access to `WlShm` but that isn't enough. 

Also looking at SCTK code doesn't help because they do it with "raw" memfds and shms kind of like I did before [link](https://github.com/Smithay/wayland-rs/blob/fdd88c4fa2f5c5abaa3044f62c4498f5f7b7dbd4/wayland-cursor/src/lib.rs#L125-L164). SlotPool seems more convenient though so we should continue to use that. It also seems to be able to actually reuse memory while `wayland-cursor` doesn't care about that because there are so few named cursor types in existence.

src/cursor.rs
eero-lehtinen: I know, but I guess we are keeping it then.

src/platform_impl/linux/wayland/types/cursor.rs
eero-lehtinen: Fixed

src/platform_impl/linux/x11/util/cursor.rs
eero-lehtinen: Fixed

src/platform_impl/linux/wayland/state.rs
```diff
@@ -148,13 +152,17 @@ impl WinitState {
                 (None, None)
             };
 
+        let shm = Shm::bind(globals, queue_handle).map_err(WaylandError::Bind)?;
```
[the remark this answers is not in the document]
kchibisov: Besides, I wonder if can just use `Lazy` here and move the `Arc<GlobalList>` and a `QueueHandle` inside it, so once you call it it'll `bind` the global instead of carrying it around. That way it should be more ergonomic? Also, I think you can `move` the `GlobalList` into it, but it could be tricky.

In general, I'd suggest to try with `Lazy` since it'll simply code and it'll be more clear.

kchibisov (changes requested):
Also, where have you got these icons? What license they use, etc? Have you drawn them yourself just for winit?

If they were provided by some other license(usually creative commons) we should state that explicitly.

kchibisov:
r @notgull for X11.

src/platform_impl/windows/icon.rs
```diff
@@ -160,3 +165,87 @@ pub fn unset_for_window(hwnd: HWND, icon_type: IconType) {
...
+            }
+
+            Ok(Self::from_handle(handle))
+        }
+    }
+}
+
+impl Drop for WinCursor {
+    fn drop(&mut self) {
+        unsafe { DestroyCursor(self.as_raw_handle()) };
```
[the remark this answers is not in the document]
eero-lehtinen: Oops, yeah it's completely wrong. Wrapper was kinda necessary.

src/platform_impl/linux/wayland/window/state.rs
```diff
@@ -702,6 +711,55 @@ impl WindowState {
...
+        if self.cursor_visible {
+            self.apply_custom_cursor(&cursor);
+        }
+
+        self.selected_cursor = SelectedCursor::Custom(cursor);
+    }
+
+    pub fn apply_custom_cursor(&self, cursor: &CustomCursorInternal) {
+        self.apply_on_poiner(|pointer, _| {
+            // Adapted from https://github.com/Smithay/client-toolkit/blob/6e4b95c967b1e6893fcb596db8888b16962c302a/src/seat/pointer/mod.rs#L452-L496
```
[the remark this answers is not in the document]
eero-lehtinen: Sure

eero-lehtinen:
> Also, where have you got these icons? What license they use, etc? Have you drawn them yourself just for winit?
> 
> If they were provided by some other license(usually creative commons) we should state that explicitly.

I drew them myself with just for this. Should we still somehow state that?

kchibisov:
>I drew them myself with just for this. Should we still somehow state that?

That sounds fine then, I just wanted to ensure whether we need a separate license for them.

src/window.rs
```diff
@@ -1346,6 +1347,13 @@ impl Window {
             .maybe_queue_on_main(move |w| w.set_cursor_icon(cursor))
     }
 
+    #[inline]
+    pub fn set_custom_cursor(&self, cursor: &CustomCursor) {
```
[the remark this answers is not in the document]
eero-lehtinen: I think I added docs for all public items now.

src/cursor.rs
```diff
@@ -0,0 +1,170 @@
...
+/// };
+/// ```
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct CustomCursor {
+    pub(crate) inner: Arc<PlatformCustomCursor>,
+}
+
+impl CustomCursor {
+    pub fn from_rgba(
+        rgba: Vec<u8>,
```
notgull, line 37: Nit: use `impl Into<Vec<u8>>` here instead, so that byte slices can be used here as well.

src/platform_impl/linux/x11/util/cursor.rs
```diff
@@ -1,12 +1,63 @@
...
 
 use x11rb::connection::Connection;
 
-use crate::window::CursorIcon;
+use crate::{cursor::CursorImage, window::CursorIcon};
 
 use super::*;
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct CustomCursorInternal(ffi::Cursor);
```
notgull, line 10: At the moment we're working on phasing out Xlib (#3198). So I would prefer using `x11rb`'s utilities instead. It shouldn't be too complicated:

- Use [`create_pixmap`](https://docs.rs/x11rb/latest/x11rb/protocol/xproto/fn.create_pixmap.html) then [`put_image`](https://docs.rs/x11rb/latest/x11rb/protocol/xproto/fn.put_image.html) to create a pixmap with the cursor's contents.
- Use [`create_cursor`](https://docs.rs/x11rb/latest/x11rb/protocol/xproto/fn.create_cursor.html) to turn that pixmap into a cursor.
- [Change the window attributes](https://docs.rs/x11rb/latest/x11rb/protocol/xproto/struct.ChangeWindowAttributesAux.html#method.cursor) to use this cursor.

src/platform_impl/linux/x11/util/cursor.rs
```diff
@@ -1,12 +1,63 @@
...
+            Self(cursor)
+        }
+    }
+
+    pub unsafe fn destroy(&self, xlib: &ffi::Xlib, display: *mut ffi::Display) {
+        unsafe {
+            (xlib.XFreeCursor)(display, self.0);
+        }
+    }
+}
```
notgull, line 47: Rather than having an unsafe drop function like this, I would prefer for the `CustomCursorInternal` type to clone the `Arc<XConnection>` and use that in a safer `drop` implementation.

src/platform_impl/web/cursor.rs
```diff
@@ -0,0 +1,278 @@
...
+}
+
+impl SelectedCursor {
+    pub fn set_style(&self, style: &Style) {
+        let value = match self {
+            SelectedCursor::Named(icon) => icon.name(),
+            SelectedCursor::Url(url) => url,
+            SelectedCursor::Image(image) => {
+                let image = image.borrow();
+                let value = match image.deref() {
```
notgull, line 90:
```suggestion
                let value = match &**image {
```

src/cursor.rs
```diff
@@ -0,0 +1,170 @@
...
+        }
+
+        if hotspot_x >= width || hotspot_y >= height {
+            return Err(BadImage::HotspotOutOfBounds {
+                width,
+                height,
+                hotspot_x,
+                hotspot_y,
+            });
+        }
```
notgull, line 160: Add tests to make sure proper errors are returned for bad inputs like these.

notgull (changes requested):

src/platform_impl/web/cursor.rs
daxpedda: I very much prefer using `deref()` instead of `&**`.
Why the change?

notgull: Manual `deref()` is a code smell; there is a Clippy lint against it. It's much better to deref the data using the operator.

src/platform_impl/linux/wayland/state.rs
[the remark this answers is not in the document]
eero-lehtinen: I don't get it. How do I bind it exacly?

I can't really create an `Shm` like this. Doesn't it need `impl ShmHandler`? And we anyway have an `Shm` already.
```
let globals = Arc::new(globals);
let globals_clone = globals.clone();
let queue_handle_clone = queue_handle.clone();
let pool = Lazy::new(Box::new(move || {
    let shm = Shm::bind(&globals_clone, &queue_handle_clone).unwrap();
    Arc::new(Mutex::new(SlotPool::new(2, &shm).unwrap()))
}) as _);
```

src/platform/web.rs
```diff
@@ -200,3 +202,20 @@ pub enum PollStrategy {
...
+        Self {
+            inner: PlatformCustomCursor::Url {
+                url,
+                hotspot_x,
+                hotspot_y,
+            }
+            .into(),
+        }
+    }
+}
```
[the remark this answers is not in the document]
eero-lehtinen: Done

src/cursor.rs
```diff
@@ -0,0 +1,170 @@
...
+///     CustomCursor::from_url("http://localhost:3000/cursor.png", 0, 0).unwrap()
+/// };
+/// ```
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct CustomCursor {
+    pub(crate) inner: Arc<PlatformCustomCursor>,
+}
+
+impl CustomCursor {
+    pub fn from_rgba(
```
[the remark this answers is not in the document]
eero-lehtinen: Thanks for noticing.

I moved the relevant parts of the struct documentation here and added a little extra, but I don't think I can add much more constructive information to that. 

I also think the struct level example is plenty and I think it helps people find the ext traits, so I wouldn't move them elsewhere.

src/cursor.rs
```diff
@@ -0,0 +1,170 @@
...
+        hotspot_x: u32,
+        hotspot_y: u32,
+    ) -> Result<Self, BadImage> {
+        if rgba.len() % PIXEL_SIZE != 0 {
+            return Err(BadImage::ByteCountNotDivisibleBy4 {
+                byte_count: rgba.len(),
+            });
+        }
+        let pixel_count = rgba.len() / PIXEL_SIZE;
+        if pixel_count != (width * height) as usize {
```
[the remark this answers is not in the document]
eero-lehtinen: Converting both to u64 seems safe as well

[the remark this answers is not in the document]
eero-lehtinen: Though I guess in multiple places when creating the actual platform cursors we have these possibly wrapping multiplications, so those could be patched up.

src/platform_impl/web/cursor.rs
eero-lehtinen: I would say who cares because it's only enabled with pedantic and its reasoning isn't too convincing. https://rust-lang.github.io/rust-clippy/master/#/explicit_deref_methods

src/cursor.rs
eero-lehtinen: Fixed

src/cursor.rs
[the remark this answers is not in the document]
eero-lehtinen: Or a significantly easier way would be to just have a sensible upper limit of like 4k images. That way we could also report errors here directly instead of silently doing nothing when the platform specific cursor are created.

src/platform_impl/linux/wayland/state.rs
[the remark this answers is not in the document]
kchibisov: Hm, yeah, we need `Shm` already, sorry. My main issue is that I don't want to create a pool until it really needs to be created. I guess it doesn't work like that with global rebind.

Maybe keep it as is(without any Lazy stuff) and I'll look into it myself.

src/platform_impl/web/cursor.rs
daxpedda: > Manual `deref()` is a code smell; there is a Clippy lint against it. It's much better to deref the data using the operator.

Could you elaborate why it is a code smell? The Clippy lint is "pedantic" for a reason, imo it should just be "restriction".

src/platform_impl/linux/x11/util/cursor.rs
eero-lehtinen: I tried to do it with create_cursor, but it seems to actually only be able to show cursors with two colors instead of a full rgba image https://www.x.org/releases/X11R7.7/doc/xproto/x11protocol.html#requests:CreateCursor. And I couldn't find an equivalent to the XcursorImageCreate in x11rb.

src/platform_impl/linux/x11/util/cursor.rs
eero-lehtinen: Done.

eero-lehtinen:
I just did some testing on maximum cursor sizes. 
- Web has a limit of 128x128. 
- At least on my setup, X11 worked with 256x256, but had artifacts with anything larger. 
- Wayland, Windows and macOS had no issues even with 4096x4096. Though on Windows, creating the cursor took like 2 secs every time :) when swapping.

kchibisov:
>At least on my setup, X11 worked with 256x256, but had artifacts with anything larger.

this all boils down to hardware cursors and the plane sizes. You could use `drm_info` to get the value for your setup when under Xorg. Some drivers have issues with hardware cursors in general though.

src/cursor.rs
eero-lehtinen: I think that testing this wouldn't give that much value, as the code and the number comparisons are simple enough. Testing for the overflows could be nice, but I was thinking of other solutions for that in comments above.

src/cursor.rs
```diff
@@ -0,0 +1,190 @@
...
+/// Use a custom image as a cursor (mouse pointer).
+///
+/// ## Platform-specific
+///
+/// - **Web**
+///     Browsers have limits on cursor sizes. On Chromium, Firefox, and Safari, the maximum
+///     size in 128x128.
+///
+///     On Firefox, [a bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1336764) causes
+///     the cursor to not change until the mouse is moved.
```
daxpedda, line 15: Actually the specification mandates the cursor to be resized instead of limited: https://www.w3.org/TR/css-ui-3/#ref-for-concrete-object-size:
> If an operating system is **incapable** of rendering a cursor above a given size, cursors larger than that size must be shrunk to within the OS-supported size bounds, while maintaining the cursor image’s intrinsic ratio, if any.

I don't believe Safari actually limits cursor size to 128x128, because I've seen some bug reports of blurry cursors when they are large, but I'm unable to test this myself.
Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1253573
Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=1204322

I don't think we are generally in the habit of documenting browser bugs inside the API documentation. I don't know if we should change that, but we shouldn't do that in this PR. See https://github.com/rust-windowing/winit/issues/2875.

```suggestion
/// **Web**: Some browsers have limits on cursor sizes typically at 128x128.
```

Despite what I said earlier, I think we should still document at least that, because it's a pretty notable bug compared to what we've found so far in the past.

eero-lehtinen: Okay thank you

eero-lehtinen: But the cursors definitely weren't resized when over 128px, it just showed the default cursor.

src/platform/web.rs
```diff
@@ -200,3 +202,23 @@ pub enum PollStrategy {
     #[default]
     Scheduler,
 }
+
+pub trait CustomCursorExtWebSys {
+    /// Creates a new cursor from a URL pointing to an image.
+    /// It uses the [url css function](https://developer.mozilla.org/en-US/docs/Web/CSS/url),
+    /// so the same image formats and paths are supported.
```
daxpedda, line 209: While we are at it, this might be a bit misleading, even though the specification mandates this, no browser actually supports all image formats they should be supporting.

Again, this is actually a bug, but maybe we could write something like this:
```suggestion
    /// It uses the [url css function](https://developer.mozilla.org/en-US/docs/Web/CSS/url),
    /// but browser support for image formats is inconsistent. Using [PNG] is recommended.
    ///
    /// [PNG]: https://en.wikipedia.org/wiki/PNG
```

eero-lehtinen: That seems nicer.

src/cursor.rs
daxpedda: Do you mean on Safari?

eero-lehtinen: I tested on windows, x11 and wayland with chrome and firefox. All of then showed default cursor with over 128px. The exact same happened on macos with safari.

eero-lehtinen: Or I think my browsers ran with xwayland but anyway.

daxpedda: Yeah Chrome and Firefox are known, I posted the tracking issues of each browser above.
I'm surprised about Safari, FWIW I couldn't find a bug report at WebKit.

Thank you for testing!

eero-lehtinen: And I think the blurriness might be because the cursor gets upscaled when using screen scaling. Maybe some mac machines have huge scalings on by default.

daxpedda: Ah, that explains it.
Btw, if you have the motivation you could write a bug report at https://bugs.webkit.org similar to https://bugzilla.mozilla.org/show_bug.cgi?id=1253573. Maybe mention the spec as well.

I will actually go ahead and mention the spec in the Firefox issue.

eero-lehtinen: I dunno if I care that much :D

eero-lehtinen:
I did some more testing with maximum sizes:
- X11: doesn't accept widths or heights larger than 32767, currently we have a manual panic to hande that.
- Wayland: crashed when I tried to make 1x100mil sized cursor.
- Mac: didn't crash or anything but cursor went invisible at 1x10mil
- Windows: went invisible at 1x100k, became visible again at 1x10mil, so maybe it wraps already internally somehow. Didn't crash.
-  Web: just was invisible when any size was over 128, didn't crash or throw errors.

So what should we do? At least we should not crash and instead show a warning on x11 when too large. Maybe also limit all sizes to u16 to prevent issues on other platforms?

kchibisov:
Can't we clamp to window size?

eero-lehtinen:
We could, but wouldn't it be weird to have different behaviour on different screens? And we would have to clamp the image data somehow too so all lines aren't offset.

kchibisov:
Well, we can clamp to 2048x2048, unless you can find a hardware plane bigger than that I doubt we should care.

eero-lehtinen:
Yeah I was thinking that from_rgba just doesn't accept anything bigger than some sensible max like 2k.

daxpedda:
I don't think clamping is a good idea, we can just return an error if we want to validate some maximum size.

Returning an error seems like the sane way to go about this imo, unless we want to get into the business of scaling images.

kchibisov:
You can't validate some stuff because you don't know whether you'll crash or not beforehand, it's just a pointless complexity for no compelling reason.

daxpedda:
> You can't validate some stuff because you don't know whether you'll crash or not beforehand, it's just a pointless complexity for no compelling reason.

I don't understand, if you want to clamp to a specific size, you could validate to the same size, how is that more complex?

kchibisov:
>I don't understand, if you want to clamp to a specific size, you could validate to the same size, how is that more complex?

Do you want to validate to 2048x2048? Then it's likely fine that way.

daxpedda:
> Do you want to validate to 2048x2048? Then it's likely fine that way.

Indeed!
Actual validation would increase the scope of this PR, which I would like to avoid.

src/platform_impl/web/cursor.rs
notgull: Fair enough; it's your backend, so go ahead.

src/platform_impl/linux/x11/util/cursor.rs
notgull: It probably requires some subtlety; this is how `XcursorImageCreate` handles it under the hood, after all. I've been meaning to remove the `Xcursor` bits in a future PR. So this is fine for now.

notgull (approved):
LGTM on the X11 part. There are some things to fix, but those are probably better suited for other PRs.

daxpedda (approved):
Web approved!

daxpedda:
I noticed that since Firefox 120 the cursor now updates without moving the mouse. @eero-lehtinen would you mind confirming this?

Also addressed https://github.com/eero-lehtinen/winit/pull/1#issuecomment-1821835929 in https://github.com/eero-lehtinen/winit/pull/2.

eero-lehtinen:
@daxpedda 
With Firefox 0.120.0, without [eero-lehtinen#2](https://github.com/eero-lehtinen/winit/pull/2), it didn't update for me without moving the mouse. 

With that PR, it updated but very slowly. Like 2x per second compared full refresh rate (or maybe a little less because of async)  when moving the mouse.

So kind of better but not really fixed.

And that PR nicely fixes my issues in Chrome.

eero-lehtinen:
Ok I did more testing and this seems to be completely fixed from the Firefox side, because it updated at full speed when using `CustomCursor::from_url`. I guess creating the cursor just takes too long for it to be shown in the same frame or something.

daxpedda:
Yeah, I will figure out what needs to be done in the follow-up to enable caching. Web is probably the worst offender here.

src/platform_impl/windows/event_loop.rs
```diff
@@ -2004,26 +2004,23 @@ unsafe fn public_window_callback_inner<T: 'static>(
...
-            };
-
-            match set_cursor_to {
-                Some(cursor) => {
-                    let cursor = unsafe { LoadCursorW(0, util::to_windows_cursor(cursor)) };
-                    unsafe { SetCursor(cursor) };
-                    result = ProcResult::Value(0);
-                }
-                None => result = ProcResult::DefWindowProc(wparam),
+            let window_state = userdata.window_state_lock();
```
msiglreith, line 2007: we should keep locks to a minimum, in particular not covering a call to `DefWindowProc(wparam)` as it might be re-entrant

eero-lehtinen:
What is the status on this? I have rebased but otherwise I haven't done any changes in the last 2 weeks. I'm still ready continue if needed.

kchibisov:
@eero-lehtinen I've fixed some style issue, but it should be good.

kchibisov (approved):

src/cursor.rs
```diff
@@ -0,0 +1,198 @@
...
+    pub(crate) inner: Arc<PlatformCustomCursor>,
+}
+
+impl CustomCursor {
+    /// Creates a new cursor from an rgba buffer.
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **Web:** Setting cursor could be delayed due to use of data URLs, which are async by
+    ///     nature.
```
daxpedda, line 44:
```suggestion
    /// - **Web:** Setting cursor could be delayed due to the creation of `Blob` objects,
    ///   which are async by nature.
```
This actually has nothing to do with data URLs, which we aren't even using in this case.
(this was wrong already before kchibisov improved the language)

kchibisov: Thanks, I found it weird as well, but given that you reviewed I thought that I'll just reformulate.

msiglreith (approved):

daxpedda:
@eero-lehtinen thank you for your patience in seeing this through! Great work!

eero-lehtinen:
Thanks for the help!

PR #3222, opened 2023-11-12
Reload DPI on PropertyChange

### LINKED ISSUE #1228: Allow changing X11 DPI at runtime

notgull:
Supersedes #2874, fixes #1228

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Unreleased` header.
 # Unreleased
 
 - On X11, check common alternative cursor names when loading cursor.
+- On X11, reload the DPI after a property change event.
 - On Windows, fix so `drag_window` and `drag_resize_window` can be called from another thread.
 - On Windows, fix `set_control_flow` in `AboutToWait` not being taken in account.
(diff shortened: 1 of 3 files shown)


frebib:
This is amazing! Thank you 😍
I don't know if it's easy to do, but the cursor size should probably update at the same. Once this is merged, I'll test it and raise a feature request depending on how it goes

notgull:
> I don't know if it's easy to do, but the cursor size should probably update at the same.

How would this be done?

kchibisov:
>I don't know if it's easy to do, but the cursor size should probably update at the same. Once this is merged, I'll test it and raise a feature request depending on how it goes

Cursors are server side on X11, so it's up to it to resize the cursor, not winit.

notgull:
The error here appears to be spurious

kchibisov (approved):

PR #3224, opened 2023-11-13
Fix crash when running iPad build on macOS

arendjr:
Newer versions of macOS have a feature that allows them to run iOS apps in an iPad-like form factor. Unfortunately, my game built with Bevy and winit used to [immediately crash](https://github.com/Couch-Chilis/sudoku-pi/issues/12) when trying this out. This PR fixes the crashes.

The changes in `event_loop.rs` were necessary to get the app to start, while the changes in `app_state.rs` allow the app to also exit cleanly without crashing when pressing Cmd+Q.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Unreleased` header.
 
 # Unreleased
 
+- Fix crash when running iOS app on macOS.
 - On X11, check common alternative cursor names when loading cursor.
 - On Windows, fix so `drag_window` and `drag_resize_window` can be called from another thread.
 - On Windows, fix `set_control_flow` in `AboutToWait` not being taken in account.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ Unreleased` header.
 
 # Unreleased
 
+- Fix crash when running iOS app on macOS.
```
madsmtm, line 14: Nit: I think this is called running under "mac catalyst"?

madsmtm (approved):
Thanks!

CHANGELOG.md
arendjr: To be honest, I’m not even sure 😅 It sounds like Mac Catalyst is a way of building Mac apps from the same codebase as used for an iPad build. But that requires adding a Mac Catalyst target which I had not done. I just ran the iPad build directly on Mac. Maybe Apple also considers that part of the Catalyst

PR #3225, opened 2023-11-14
Draft: Support open urls macos

raphamorim:
# Draft (open to discussion)

Opened this draft PR but haven't properly tested yet (will do during this week in a terminal emulator I work on), just wanted to confirm is something winit would like to add before proceed

Related https://github.com/raphamorim/rio/issues/107 

# TODO

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -251,6 +251,13 @@ pub enum Event<T: 'static> {
     ///
     /// - **macOS / Wayland / Windows / Orbital:** Unsupported.
     MemoryWarning,
+
+    /// Emitted when the app is open by external resources, like opening a file or deeplink.
+    ///
+    /// ## Others
+    ///
+    /// - **Android / Wayland / Windows / Orbital:** Unsupported.
+    Opened { urls: Vec<url::Url> },
 }
 
 impl<T> Event<T> {
@@ -261,6 +268,7 @@ impl<T> Event<T> {
             UserEvent(_) => Err(self),
             WindowEvent { window_id, event } => Ok(WindowEvent { window_id, event }),
             DeviceEvent { device_id, event } => Ok(DeviceEvent { device_id, event }),
+            Opened { urls } => Ok(Opened { urls }),
             NewEvents(cause) => Ok(NewEvents(cause)),
             AboutToWait => Ok(AboutToWait),
             LoopExiting => Ok(LoopExiting),
--- a/src/platform_impl/macos/app_delegate.rs
+++ b/src/platform_impl/macos/app_delegate.rs
@@ -1,3 +1,4 @@
+use std::ffi::CStr;
 use std::ptr::NonNull;
 
 use icrate::Foundation::NSObject;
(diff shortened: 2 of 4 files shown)


src/event.rs
```diff
@@ -251,6 +251,13 @@ pub enum Event<T: 'static> {
     ///
     /// - **macOS / Wayland / Windows / Orbital:** Unsupported.
     MemoryWarning,
+
+    /// Emitted when the app is open by external resources, like opening a file or deeplink.
+    ///
+    /// ## Others
+    ///
+    /// - **Android / Wayland / Windows / Orbital:** Unsupported.
```
daxpedda, line 259:
```suggestion
    /// - **Android / Wayland / Windows / Orbital / Web:** Unsupported.
```

src/platform_impl/macos/app_delegate.rs
```diff
@@ -58,6 +68,20 @@ declare_class!(
             // TODO: Notify every window that it will be destroyed, like done in iOS?
             AppState::internal_exit();
         }
+
+        #[method(applicationOpenUrls:)]
+        fn application_open_urls(&self, urls: Url) -> () {
```
madsmtm, line 73: This method has to match that in [`icrate::Foundation::AppKit::NSApplicationDelegate::application_openURLs`](https://docs.rs/icrate/latest/icrate/AppKit/trait.NSApplicationDelegate.html#method.application_openURLs) exactly, otherwise it just plain won't work.

```suggestion
        #[method(application:openUrls:)]
        fn application_openURLs(&self, application: &NSApplication, urls: &NSArray<NSURL>) {
```

madsmtm (commented):
I'm out of the loop on this, but I remember in https://github.com/rust-windowing/winit/pull/1759 that just _implementing_ similar methods causes trouble, as now suddenly the application launched differently.

Before we merge this, we'll have to check if that is also the case with `application:openUrls:`.

madsmtm:
> wanted to confirm is something winit would like to add before proceed

Thanks!

I'm not fundamentally opposed to the addition, but I will say that this reminds me too much of the aforementioned issue, which is honestly mentally taxing, so I probably won't really spend much time on this.

Apologies for the early letdown :/

raphamorim:
> I'm not fundamentally opposed to the addition, but I will say that this reminds me too much of the aforementioned issue, which is honestly mentally taxing, so I probably won't really spend much time on this.
> Apologies for the early letdown :/

I understand, thanks for the honesty! Then I think might be better close this PR 🙏

rib:
Just for reference here - I happened to also implement support for `application:openUrls` in Winit very recently (except currently against winit 0.28 which we're still based on) along with an `OpenURL` event that is dispatched in response.

I'd be very interested in having something like that upstream at some point.

Currently our macOS application depends on some Apple Script trickery as a workaround for not being hooked into `application:openURLs` but it would be nicer if we could get an event via Winit instead.

One consideration is that `application:openURLs` was add in 10.13 where as  FEATURES.md currently says winit aims to support macOS 10.7+ - though I'd guess it's probably OK to just not support the event on earlier version instead of adding complexity for those versions.

Especially when we look to porting to Winit 0.29 I'd be very interested in re-visiting this hopefully.

PR #3236, opened 2023-11-22
Make Android docs build on docs.rs

madsmtm:
As discussed in https://github.com/rust-windowing/winit/pull/2324#issuecomment-1823536187.

### DIFF
--- a/src/platform/android.rs
+++ b/src/platform/android.rs
@@ -84,5 +84,10 @@ impl<T> EventLoopBuilderExtAndroid for EventLoopBuilder<T> {
 /// use winit::platform::android::activity::AndroidApp;
 /// ```
 pub mod activity {
+    // We enable the `"native-activity"` feature just so that we can build the
+    // docs, but it'll be very confusing for users to see the docs with that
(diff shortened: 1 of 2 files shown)


src/platform/android.rs
```diff
@@ -84,5 +84,10 @@ impl<T> EventLoopBuilderExtAndroid for EventLoopBuilder<T> {
 /// use winit::platform::android::activity::AndroidApp;
 /// ```
 pub mod activity {
+    // We enable the `"native-activity"` feature just so that we can build the
+    // docs, but it'll be very confusing for users to see the docs with that
+    // feature enabled, so we avoid inlining it so that they're forced to view
+    // it on the crate's own docs.rs page.
+    #[doc(no_inline)]
```
MarijnS95, line 91: Didn't even know about this, that's cool!

MarijnS95 (approved):

MarijnS95:
Added this to the `0.29.4` milestone for backporting.

madsmtm:
[Seems like it worked](https://docs.rs/winit/0.29.4/aarch64-linux-android/winit/index.html)

PR #3239, opened 2023-11-24
Winit version 0.29.4

kchibisov:


daxpedda:
> That's the only patch I've excluded #3219 , since I'm not quite sure about it. cc @daxpedda

That can be skipped, it has no effect on functionality.

PR #3241, opened 2023-11-25
Taskbar progress

valaphee:

This adds a simple way to set the taskbar progress on Windows, there is still no way to set the state to error or paused, and don't know if it should be added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ Unreleased` header.
 
 # Unreleased
 
+- On Windows, add `WindowExt::set_taskbar_progress`.
+
 # 0.29.4
 
 - Fix crash when running iOS app on macOS.
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -125,6 +125,7 @@ If your PR makes notable changes to Winit's features, please update this section
 * Setting a menu bar
 * `WS_EX_NOREDIRECTIONBITMAP` support
 * Theme the title bar according to Windows 10 Dark Mode setting or set a preferred theme
+* Setting the taskbar progress
 
 ### macOS
 * Window activation policy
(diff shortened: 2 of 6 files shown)


daxpedda:
Dropping some stuff I found while searching for cross-platform compatibility:
- [`taskbar_interface`](https://github.com/sandmor/taskbar_interface) - Rust crate that can do this externally without the help of Winit. Supports Windows and Linux.
- X11 can be supported through `libxapps` with `_NET_WM_XAPP_PROGRESS`. Actual DE support is spotty.
- There is a [Unity Launcher API](https://wiki.ubuntu.com/Unity/LauncherAPI#Progress) for this, which e.g. KDE supports. I assume this isn't something we can support in Winit.
- MacOS has the ability to draw whatever in the taskbar, so there are some implementations that simply draw a progressbar. See e.g. [DockProgressBar](https://github.com/hokein/DockProgressBar).

valaphee:
Ah didn't know that this is not only a Windows thing, will look into it

daxpedda:
> Ah didn't know that this is not only a Windows thing, will look into it

Let's do it in follow-ups though, because this here is uncontroversial while the others are not.
You should get the other backends approved by the corresponding backend maintainers first before putting in any significant work.

kchibisov:
Such stuff generally has nothing to do with windowing which is all winit about. That's the reason external APIs exist.

On Linux there's no standard interface for that, the same goes to tray. It's usually all dbus, because again, nothing to do with windowing and tied to specific DE, so not winit job.

notgull:
Agreed with @kchibisov. If it has nothing to do with windowing, it doesn't belong in `winit`.

valaphee:
Just out of curiosity, the taskbar buttons have to be handled in WndProc https://github.com/microsoft/Windows-classic-samples/blob/main/Samples/Win7Samples/winui/shell/appshellintegration/TaskbarThumbnailToolbar/ThumbnailToolbar.cpp

daxpedda:
> Such stuff generally has nothing to do with windowing which is all winit about. That's the reason external APIs exist.
> 
> On Linux there's no standard interface for that, the same goes to tray. It's usually all dbus, because again, nothing to do with windowing and tied to specific DE, so not winit job.

I assume this was meant for Linux only? Windows still seems in scope to me considering we are handling plenty of similar taskbar related things there already.

I would consider MacOS not in scope as well because it doesn't actually offer an API to do that.

PR #3243, opened 2023-11-27
Remove erroneous `must_use` from `Window::request_inner_size()`

daxpedda:
`Window::request_inner_size()` has side effects, so being `must_use` doesn't make sense.

### DIFF
--- a/src/window.rs
+++ b/src/window.rs
@@ -754,7 +754,7 @@ impl Window {
     /// may not be generated.
     ///
     /// On platforms where resizing is disallowed by the windowing system, the current
-    /// inner size is returned immidiatelly, and the user one is ignored.
+    /// inner size is returned immediately, and the user one is ignored.
     ///
     /// When `None` is returned, it means that the request went to the display system,
     /// and the actual size will be delivered later with the [`WindowEvent::Resized`].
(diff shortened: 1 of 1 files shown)


daxpedda:
> On wayland it won't ever generate a resize event so if you don't use you have no idea that your window is of the wrong geometry, which could lead to protocol errors.

What do you mean with protocol errors in this context?

kchibisov:
If you `request_inner_size` and don't actually honor the size you requested, some compositors could just decide to do bad things to your window.

Like the `#[must_use]` has a reason indicating that it's a way to communicate **Resized** event instead, because the `Resized` event won't ever arrive on Wayland.

daxpedda:
> If you `request_inner_size` and don't actually honor the size you requested, some compositors could just decide to do bad things to your window.

Understood.

> Like the `#[must_use]` has a reason indicating that it's a way to communicate **Resized** event instead, because the `Resized` event won't ever arrive on Wayland.

This was my question here: https://github.com/rust-windowing/winit/pull/3243#discussion_r1406043391.
Can we not make it so a `Resized` event is always sent as long as the size changed no matter the backend? E.g. if Wayland doesn't send an event can we simulate it?

kchibisov:
>This was my question here: https://github.com/rust-windowing/winit/pull/3243#discussion_r1406043391.
Can we not make it so a Resized event is always sent as long as the size changed no matter the backend? E.g. if Wayland doesn't send an event can we simulate it?

Not with the current design, I won't add side channels for **instant** actions to make Wayland suck because of the bad design.

In general, the size is controlled by user on Wayland, there's no concept of resize event at all on Wayland, with the exception for `Maximized` size and `Fullscreen` (compositor is in right to kill you here if you don't obey).

daxpedda:
Alright!

PR #3254, opened 2023-12-01
fix refresh_rate_millihertz on macOS

xiaopengli89:

When the monitor refresh rate changes, `CVDisplayLinkGetNominalOutputVideoRefreshPeriod` still returns the previous value unless the process is restarted. I don't know the reason, but `CGDisplayModeGetRefreshRate` can return the correct result.

### DIFF
--- a/src/platform_impl/macos/monitor.rs
+++ b/src/platform_impl/macos/monitor.rs
@@ -219,10 +219,9 @@ impl MonitorHandle {
     pub fn refresh_rate_millihertz(&self) -> Option<u32> {
         unsafe {
             let current_display_mode = NativeDisplayMode(CGDisplayCopyDisplayMode(self.0) as _);
-            let refresh_rate =
-                ffi::CGDisplayModeGetRefreshRate(current_display_mode.0).round() as u32;
-            if refresh_rate > 0 {
-                return Some(refresh_rate * 1000);
+            let refresh_rate = ffi::CGDisplayModeGetRefreshRate(current_display_mode.0);
+            if refresh_rate > 0.0 {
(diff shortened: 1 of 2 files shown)


src/platform_impl/macos/monitor.rs
```diff
@@ -216,6 +218,13 @@ impl MonitorHandle {
 
     pub fn refresh_rate_millihertz(&self) -> Option<u32> {
         unsafe {
+            let current_display_mode = NativeDisplayMode(CGDisplayCopyDisplayMode(self.0) as _);
+            let refresh_rate =
+                ffi::CGDisplayModeGetRefreshRate(current_display_mode.0).round() as u32;
+            if refresh_rate > 0 {
+                return Some(refresh_rate * 1000);
+            }
```
kchibisov, line 226: What's the point to have the rest of the code? As a fallback?

src/platform_impl/macos/monitor.rs
```diff
@@ -216,6 +218,13 @@ impl MonitorHandle {
 
     pub fn refresh_rate_millihertz(&self) -> Option<u32> {
         unsafe {
+            let current_display_mode = NativeDisplayMode(CGDisplayCopyDisplayMode(self.0) as _);
+            let refresh_rate =
+                ffi::CGDisplayModeGetRefreshRate(current_display_mode.0).round() as u32;
```
kchibisov, line 223: You should round after conversion to mHz.

kchibisov (changes requested):
Also, a changelog entry.

madsmtm:
Thanks!

PR #3257, opened 2023-12-05
Add DWMWA_SYSTEMBACKDROP_TYPE support on Windows

Dubzer:

With this code, it will be possible to apply the [Acrylic](https://learn.microsoft.com/en-us/windows/apps/design/style/acrylic) and the [Mica](https://learn.microsoft.com/en-us/windows/apps/design/style/mica) backdrop material effect

Tested on Windows 11 build 22631.2715, using new API in the Alacritty source code. 

image


Dubzer:
@msiglreith can you maybe take a look? 👀

msiglreith (approved):
Thanks, lgtm - unable to test it as I don't have Windows 11 installed, on older builds it returns an unsupported parameter error but doesn't cause any issues.

Keeping it on hold (similar to other PRs with backend specific extensions) as there is a larger API change planned for the next version https://github.com/rust-windowing/winit/issues/3367

madsmtm:
> larger API change planned for the next version https://github.com/rust-windowing/winit/issues/3367

I don't think you should block on that, we still need to do a lot more discussion before we're prepared to do the API rework.

Dubzer:
@msiglreith would you like for me to resolve merge conflicts?

msiglreith:
@Dubzer yes, please!

Dubzer:
> @Dubzer yes, please!

Done

MarijnS95:
@Dubzer since you tested this inside Alacritty, did you ever open a PR to enable it? Perhaps behind `[window] blur`?

Dubzer:
@MarijnS95 actually I didn't. I waited for it to update the winit version and then forgot lol. It should be pretty easy to add such a flag if you want to contribute it yourself

MarijnS95:
@Dubzer the plan is to back this platform extension out entirely "lol". There's an API-agnostic `set_blur()` that's currently only implemented on MacOS, which should be generalized and use DWM SBD implemented here (or one of the many other Windows fallbacks on older versions... Thanks Windows). Then we can finally close #538 for real.

Dubzer:
@MarijnS95 Sorry, I'm out of context and not really sure about the current plans for set_blur. 
The reason I made SBD as a platform extension is because DWM on modern Windows versions does not implement the background blur effect. Instead, currently there are two different effects: Acrylic and Mica (also Mica Alt variation)

Acrylic is similar to blur in that it shows what's behind the window on the z-axis. But it's still a different, more complex effect. Here's a breakdown of it from the [design docs](https://learn.microsoft.com/en-us/windows/apps/design/style/acrylic):

image

And Mica doesn't show what's behind a window at all. It blends the desktop background into the window background:

image

Given all these factors, I'm not sure how you want to combine DWM SBD with the set_blur API without creating unexpected results for the user and keeping all the material selection options available

PR #3260, opened 2023-12-09
Update to x11rb 0.13.0

psychon:
The only breaking change is that x11rb no longer reports an error when querying the `WmSizeHints` of a window that does not have this property set. For this reason, the return type of `WmSizeHintsCookie::reply()` changed from `Result<WmSizeHints, SomeError>` to
`Result<Option<WmSizeHints>, SomeError>`.

In `update_normal_hints()`, previously a cryptic error would be reported to the caller. Instead, this now uses `unwrap_or_default()` to get a `WmSizeHints`. All fields of `WmSizeHints` are `Option`s, so this produces an empty object.

`resize_increments()` queries a value from the window and returns an Option. Previously, the error for "missing property" was turned into None via `.ok()`. This commit adds a call to `flatten()` to also turn "property not set" into None.

Finally, `request_user_attention()` queries a window's `WmHints` property and updates one field of it. The code already uses `unwrap_or_default()` to deal with missing properties, so just a call to `flatten()` is needed to merge "missing property" and "error while querying" into one.

Other changes in x11rb do not seem to affect this crate.

x11rb's MSRV increased from 1.56 to 1.63, which is still below the MSRV of this crate, which is 1.65.

- [ ] Tested on all platforms changed (all I did was `cargo run --example window_debug` and `cursor` and each of those opened a window; dunno if that counts as "tested")

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -166,7 +166,7 @@ wayland-client = { version = "0.31.1", optional = true }
 wayland-protocols = { version = "0.31.0", features = [ "staging"], optional = true }
 wayland-protocols-plasma = { version = "0.2.0", features = [ "client" ], optional = true }
 x11-dl = { version = "2.18.5", optional = true }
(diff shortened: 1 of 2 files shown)


notgull (approved):
Thanks!

PR #3261, opened 2023-12-13
On Windows, fix `set_fullscreen` early return for `Fullscreen::Borderless(None)`

amrbashir:

### DIFF
--- a/src/platform_impl/windows/window.rs
+++ b/src/platform_impl/windows/window.rs
@@ -686,9 +686,19 @@ impl Window {
 
         let mut window_state_lock = window_state.lock().unwrap();
         let old_fullscreen = window_state_lock.fullscreen.clone();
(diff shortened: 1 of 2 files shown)


src/platform_impl/windows/window.rs
```diff
@@ -686,9 +686,19 @@ impl Window {
 
         let mut window_state_lock = window_state.lock().unwrap();
         let old_fullscreen = window_state_lock.fullscreen.clone();
-        if window_state_lock.fullscreen == fullscreen {
-            return;
+
+        match (&old_fullscreen, &fullscreen) {
+            // Return if we already in the same fullscreen mode
```
filnet, line 691: "Return if we already are ..."

msiglreith (approved):

amrbashir:
done

PR #3264, opened 2023-12-14
On Wayland, make `wl_subcompositor` protocol optional

MarijnS95:
This protocol is only used for (optional) Client Side Decorations (where) the compositor still takes the burden of compositing various window parts together, via subsurfaces that all belong to a single window.

If this core protocol is not available, as is the case on gamescope, disable CSD.

Related:

https://github.com/Smithay/client-toolkit/issues/210
https://github.com/alacritty/alacritty/issues/5709

### DIFF
--- a/src/platform_impl/linux/wayland/window/state.rs
+++ b/src/platform_impl/linux/wayland/window/state.rs
@@ -254,7 +254,7 @@ impl WindowState {
         &mut self,
         configure: WindowConfigure,
         shm: &Shm,
-        subcompositor: &Arc<SubcompositorState>,
+        subcompositor: &Option<Arc<SubcompositorState>>,
         event_sink: &mut EventSink,
     ) -> LogicalSize<u32> {
         // NOTE: when using fractional scaling or wl_compositor@v6 the scaling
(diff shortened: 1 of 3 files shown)


src/platform_impl/linux/wayland/window/state.rs
```diff
@@ -269,24 +269,26 @@ impl WindowState {
...
-                    frame.set_title(&self.title);
-                    frame.set_scaling_factor(self.scale_factor);
-                    // Hide the frame if we were asked to not decorate.
-                    frame.set_hidden(!self.decorate);
-                    self.frame = Some(frame);
-                }
-                Err(err) => {
-                    warn!("Failed to create client side decorations frame: {err}");
-                    self.csd_fails = true;
+            if let Some(subcompositor) = subcompositor {
```
kchibisov, line 272: Can we embed the check for subcompositor into the `if`, so we can unwrap in the code below reducing the nesting?

I just don't like the way we add more nesting to this, where it was already hard...

kchibisov (changes requested):

src/platform_impl/linux/wayland/window/state.rs
MarijnS95: Alternative is of course an `if subcompositor.is_some()` and unwrapping in the body...

kchibisov (approved):

PR #3270, opened 2023-12-17
On Web, use the new `WebCanvasWindowHandle`

daxpedda:
Now we don't have to set `data-raw-handle` on the canvas anymore.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ Unreleased` header.
 - On macOS, add services menu.
 - **Breaking:** On Web, remove queuing fullscreen request in absence of transient activation.
 - On Web, fix setting cursor icon overriding cursor visibility.
+- **Breaking:** On Web, return `RawWindowHandle::WebCanvas` instead of `RawWindowHandle::Web`.
 
 # 0.29.5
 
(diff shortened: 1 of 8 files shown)


daxpedda:
Unfortunately `WebCanvasWindowHandle` is supposed to be `!Send`, for good reasons. Currently asking in IRC how other platforms are doing this.

daxpedda:
So I went ahead and made `Window::window_handle()` return `HandleError::Unavailable` when not on the main thread. To do this I had to change the call to `raw_window_handle_rwh_06()` from `Window` because it was using `maybe_wait_on_main()`.

@madsmtm I didn't actually change anything in MacOS/iOS behavior because it still calls `maybe_wait_on_main()` with the `SendSyncWrapper` there, but maybe you want to change that to return `HandleError::Unavailable` as well?

madsmtm:
> maybe you want to change that to return `HandleError::Unavailable` as well?

Sure, see https://github.com/rust-windowing/winit/pull/3288.

daxpedda:
Alright, let me go ahead and merge this then.

madsmtm:
Also: Returning `WebCanvasWindowHandle` instead of `WebWindowHandle`, while it shouldn't be a breaking change, in practice it will be, since a lot of users will assume that it returns `WebWindowHandle`, and panic if it doesn't.

daxpedda:
> But returning `WebCanvasWindowHandle` instead of `WebWindowHandle`, while it shouldn't be a breaking change, in practice it will be, since a lot of users will assume that it returns `WebWindowHandle`, and panic if it doesn't.

Hm, alright.

PR #3271, opened 2023-12-18
windows set ime area not work

wjian23:
I encounter  the Api of  set_ime_cursor_area not work on windows with Sogou Ime when i develop a new rust GUI framework Ribir(https://github.com/RibirX/Ribir). I try to fix it here, please take a look。


src/platform_impl/windows/ime.rs
```diff
@@ -124,16 +124,24 @@ impl ImeContext {
...
+            bottom: y + height,
         };
         let candidate_form = CANDIDATEFORM {
             dwIndex: 0,
             dwStyle: CFS_EXCLUDE,
             ptCurrentPos: POINT { x, y },
             rcArea: rc_area,
         };
+        let composition_form = COMPOSITIONFORM {
+            dwStyle: CFS_EXCLUDE,
```
kchibisov, line 136: There's no such option for the form.

kchibisov (changes requested):
Sounds like IME is just bugged, from what I can read on https://learn.microsoft.com/en-us/windows/win32/api/imm/ns-imm-candidateform `rc_area` must be excluded, what you've changed uses wrong style and solely about positioning, but not about exclusion.

Maybe the rectangle passed is not correct, but `COMPOSITION_FORM` shouldn't be required. I'd also suggest to test other IMEs.

src/platform_impl/windows/ime.rs
wjian23: yeah according to the official document should use CFS_POINT here. i'll take a fix

src/platform_impl/windows/ime.rs
```diff
@@ -124,16 +124,24 @@ impl ImeContext {
...
+        let composition_form = COMPOSITIONFORM {
+            dwStyle: CFS_POINT,
+            ptCurrentPos: POINT { x, y: y + height },
+            rcArea: rc_area,
+        };
 
-        unsafe { ImmSetCandidateWindow(self.himc, &candidate_form) };
+        unsafe {
+            ImmSetCompositionWindow(self.himc, &composition_form);
+            ImmSetCandidateWindow(self.himc, &candidate_form);
```
kchibisov, line 143: Why the order is like that and not other way around? Could we set after creating data for the request we want to perform and not doing just set-all in the end of function?

Also, if you want me to re-review you should request it.

wjian23: @kchibisov the order doesn't matter here. Create and set will be more nature, i have modify based on your comments

kchibisov (approved):

kchibisov:
Thanks!

PR #3285, opened 2023-12-22
Fix run_on_demand exiting on consequent call

### LINKED ISSUE #3284: `EventLoop::run_on_demand` stuck in "exiting" when rerun
After finishing a `EventLoop::run_on_demand` run by calling `elwt.exit()` subsequent runs immediately exit.

On subsequent `run_on_demand` calls `elwt.exiting()` is `true`. I would expect exiting to be false until `exit()` is called again.

I've reproduced this issue on Linux & Macos.

## Steps to reproduce
* Run `cargo run --example window_on_demand`
* Exit window
* It hangs forever because the example never actually calls `elwt.exit()`... This is an unrelated bug in the example.

## Steps to reproduce (2)
* Fix example to call `elwt.exit()` on `CloseRequested`.
```diff
diff --git a/examples/window_on_demand.rs b/examples/window_on_demand.rs
index ca9a448a..9945319c 100644
--- a/examples/window_on_demand.rs
+++ b/examples/window_on_demand.rs
@@ -41,6 +41,7 @@ fn main() -> Result<(), impl std::error::Error> {
                     } if window.id() == window_id => {
                         println!("--------------------------------------------------------- Window {idx} CloseRequested");
                         app.window = None;
+                        elwt.exit();
                     }
                     Event::AboutToWait => window.request_redraw(),
                     Event::WindowEvent {

```
* Run `cargo run --example window_on_demand`
* Exit window
* After 5 seconds a new window is created and immediately exited without input
```
...
--------------------------------------------------------- Window 1 CloseRequested
Run 1: AboutToWait
Run 1: LoopExiting
--------------------------------------------------------- Finished first loop
--------------------------------------------------------- Waiting 5 seconds
Run 2: NewEvents(Init)
Run 2: Resumed
DEBUG [sctk_adwaita::buttons] Ignoring "appmenu" button
Run 2: WindowEvent { window_id: WindowId(WindowId(94098374278864)), event: Resized(PhysicalSize { width: 128, height: 128 }) }
Run 2: WindowEvent { window_id: WindowId(WindowId(94098374278864)), event: RedrawRequested }
Run 2: AboutToWait
Run 2: LoopExiting
--------------------------------------------------------- Finished second loop
```

kchibisov:
Fixes #3284.

### DIFF
--- a/src/platform/run_on_demand.rs
+++ b/src/platform/run_on_demand.rs
@@ -76,6 +76,14 @@ impl<T> EventLoopExtRunOnDemand for EventLoop<T> {
     where
         F: FnMut(Event<Self::UserEvent>, &EventLoopWindowTarget<Self::UserEvent>),
     {
+        self.event_loop.window_target().clear_exit();
         self.event_loop.run_on_demand(event_handler)
+    }
+}
+
+impl<T> EventLoopWindowTarget<T> {
+    /// Clear exit status.
+    pub(crate) fn clear_exit(&self) {
+        self.p.clear_exit()
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ Unreleased` header.
 - **Breaking:** On Web, macOS and iOS, return `HandleError::Unavailable` when a window handle is not available.
 - **Breaking:** Bump MSRV from `1.65` to `1.70`.
 - On Wayland, fix `WindowEvent::Destroyed` not being delivered after destroying window.
+- Fix `EventLoopExtRunOnDemand::run_on_demand` not working for consequent invocation
 
 # 0.29.5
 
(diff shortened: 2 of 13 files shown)


notgull (approved):
LGTM to me for X11

src/platform/run_on_demand.rs
```diff
@@ -76,6 +76,14 @@ impl<T> EventLoopExtRunOnDemand for EventLoop<T> {
     where
         F: FnMut(Event<Self::UserEvent>, &EventLoopWindowTarget<Self::UserEvent>),
     {
+        self.event_loop.window_target().clear_exit();
```
madsmtm, line 79: Nit: I think I'd prefer to be able to refactor things internally, maybe each platform should call this internally instead? Though I recognize that for some platforms, that'd mean setting the flag unnecessarily in `run` as well, so IDK.

kchibisov: it's the same on every platform though and they all have the same ways of doing so.

madsmtm: Yeah, but I think I might change that for macOS in the future ;)

Anyway, it's private API, so doesn't really matter.

kchibisov: Well, the point is to make backend maintainers aware that they need to do such a thing.

daxpedda (commented):
This makes sense to me, I can see how this was broken before without "clearing" the exit state.
Not sure about the Wayland stuff, I'm not really familiar with the backend.

Thank you for fixing this @kchibisov. Let me know if there is something else left broken I should take a look at.

Cc #3056, the PR which caused the breakage.

PR #3288, opened 2023-12-22
Window handle: Return an error when not on main thread on macOS and iOS

madsmtm:
Follow-up to https://github.com/rust-windowing/winit/pull/3270.

### DIFF
--- a/src/window.rs
+++ b/src/window.rs
@@ -1532,7 +1532,8 @@ impl rwh_06::HasWindowHandle for Window {
     fn window_handle(&self) -> Result<rwh_06::WindowHandle<'_>, rwh_06::HandleError> {
         let raw = self.window.raw_window_handle_rwh_06()?;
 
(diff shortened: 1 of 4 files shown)


daxpedda:
#3270 was merged!

src/window.rs
```diff
@@ -1532,7 +1532,8 @@ impl rwh_06::HasWindowHandle for Window {
     fn window_handle(&self) -> Result<rwh_06::WindowHandle<'_>, rwh_06::HandleError> {
         let raw = self.window.raw_window_handle_rwh_06()?;
 
-        // SAFETY: The window handle will never be deallocated while the window is alive.
+        // SAFETY: The window handle will never be deallocated while the window is alive,
+        // and the main thread safety requirements are upheld internally by each platform.
```
madsmtm, line 1536: I'm intending to change the docs for `raw-window-handle` to make it clear that this is what's required

PR #3291, opened 2023-12-23
MacOS: cache custom cursor

daxpedda:
This moves the creation of `NSCursor` to `CustomCursorBuilder::build()`.

Note: no public API changes.

PR #3292, opened 2023-12-23
Custom cursor improvements

daxpedda:
- Move `PlatformCustomCursorBuilder::build()` to `PlatformCustomCursor` so platforms can implement their own caching if they want to. See #3291.
- Move `Arc` in `CustomCursor` to platform implementations so platforms who have their own reference counting can use  that. E.g. MacOS with [`Id`](https://docs.rs/objc2/0.5.0/objc2/rc/struct.Id.html).

No public API or backend changes.

### DIFF
--- a/src/cursor.rs
+++ b/src/cursor.rs
@@ -80,7 +80,7 @@ impl CustomCursor {
         hotspot_y: u16,
     ) -> Result<CustomCursorBuilder, BadImage> {
         Ok(CustomCursorBuilder {
-            inner: PlatformCustomCursor::from_rgba(
+            inner: PlatformCustomCursorBuilder::from_rgba(
                 rgba.into(),
                 width,
                 height,
@@ -102,7 +102,7 @@ pub struct CustomCursorBuilder {
 impl CustomCursorBuilder {
     pub fn build<T>(self, window_target: &EventLoopWindowTarget<T>) -> CustomCursor {
         CustomCursor {
-            inner: self.inner.build(&window_target.p),
+            inner: PlatformCustomCursor::build(self.inner, &window_target.p),
         }
     }
 }
--- a/src/cursor.rs
+++ b/src/cursor.rs
@@ -46,24 +46,11 @@ const PIXEL_SIZE: usize = 4;
 /// let window = Window::new(&event_loop).unwrap();
 /// window.set_custom_cursor(&custom_cursor);
 /// ```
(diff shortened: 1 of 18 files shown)


src/cursor.rs
```diff
@@ -165,8 +152,53 @@ impl fmt::Display for BadImage {
 
 impl Error for BadImage {}
 
-/// Platforms export this directly as `PlatformCustomCursor` if they need to only work with images.
-#[derive(Debug, Clone, PartialEq, Eq)]
+/// Platforms export this directly as `PlatformCustomCursorBuilder` if they need to only work with images.
+#[derive(Debug)]
+pub struct OnlyCursorImageBuilder(pub CursorImage);
```
madsmtm, line 157: Shouldn't the inner field be `pub(crate)`?

```suggestion
pub struct OnlyCursorImageBuilder(pub(crate) CursorImage);
```

src/cursor.rs
```diff
@@ -165,8 +152,53 @@ impl fmt::Display for BadImage {
...
+        hotspot_x: u16,
+        hotspot_y: u16,
+    ) -> Result<Self, BadImage> {
+        CursorImage::from_rgba(rgba, width, height, hotspot_x, hotspot_y).map(Self)
+    }
+}
+
+/// Platforms export this directly as `PlatformCustomCursor` if they don't implement caching.
+#[derive(Debug, Clone)]
+pub struct OnlyCursorImage(pub Arc<CursorImage>);
```
madsmtm, line 174: Same as above.

```suggestion
pub struct OnlyCursorImage(pub(crate) Arc<CursorImage>);
```

madsmtm (commented):
> Move `Arc` in `CustomCursor` to platform implementations

Thanks, I thought about that too!

src/cursor.rs
daxpedda: `OnlyCursorImageBuilder` isn't actually exposed through the library.
I would generally prefer if we applied [`unreachable_pub`](https://doc.rust-lang.org/1.74.0/rustc/lints/listing/allowed-by-default.html#unreachable-pub) on the whole library.

But to avoid confusion I will mark `OnlyCursorImageBuilder` itself as `pub(crate)`.

daxpedda: Done.

madsmtm (approved):

PR #3293, opened 2023-12-23
Windows: cache custom cursor

daxpedda:
We might want to let `CustomCursorBuilder::build()` be able to return an error, but that's a discussion for the future.
For now, the behavior doesn't actually change, because before this PR we didn't return an error from `Window::set_custom_cursor()` as well.

### DIFF
--- a/src/platform_impl/windows/icon.rs
+++ b/src/platform_impl/windows/icon.rs
@@ -17,9 +17,12 @@ use windows_sys::{
 };
 
 use crate::icon::*;
-use crate::{cursor::CursorImage, dpi::PhysicalSize};
+use crate::{
+    cursor::{CursorImage, OnlyCursorImageBuilder},
+    dpi::PhysicalSize,
+};
 
-use super::util;
+use super::{util, EventLoopWindowTarget};
 
 impl Pixel {
     fn convert_to_bgra(&mut self) {
@@ -169,7 +172,7 @@ pub fn unset_for_window(hwnd: HWND, icon_type: IconType) {
 #[derive(Debug, Clone)]
 pub enum SelectedCursor {
     Named(CursorIcon),
-    Custom(WinCursor),
+    Custom(Arc<RaiiCursor>),
 }
 
 impl Default for SelectedCursor {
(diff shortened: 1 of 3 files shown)


src/platform_impl/windows/icon.rs
```diff
@@ -239,13 +233,26 @@ impl WinCursor {
...
+
+    pub(crate) fn build<T>(
+        cursor: OnlyCursorImageBuilder,
+        _: &EventLoopWindowTarget<T>,
+    ) -> WinCursor {
+        match WinCursor::new(&cursor.0) {
+            Ok(cursor) => cursor,
+            Err(err) => {
+                warn!("Failed to create custom cursor: {err}");
+                WinCursor::Failed
```
madsmtm, line 248: Seeing this, I think the API should actually be something like:
```rust
impl CustomCursor {
    pub fn from_rgba(
        rgba: impl Into<Vec<u8>>,
        width: u16,
        height: u16,
        hotspot_x: u16,
        hotspot_y: u16,
    ) -> CustomCursorBuilder;
}

impl CustomCursorBuilder {
    pub fn build<T>(self, window_target: &EventLoopWindowTarget<T>) -> Result<CustomCursor, BadImage>;
}
```

But that can be done later.

daxpedda: Agreed, will do that in a follow-up.

src/platform_impl/windows/icon.rs
```diff
@@ -239,13 +233,26 @@ impl WinCursor {
...
 
-            Ok(Self::from_handle(handle))
+            Ok(Self::Cursor(Arc::new(RaiiCursor { handle })))
+        }
+    }
+
+    pub(crate) fn build<T>(
+        cursor: OnlyCursorImageBuilder,
+        _: &EventLoopWindowTarget<T>,
+    ) -> WinCursor {
```
msiglreith, line 243: minor: I would prefer replacing the instances of `WinCursor` with `Self`

msiglreith (approved):

src/platform_impl/windows/icon.rs
daxpedda: Done.

PR #3295, opened 2023-12-23
(Try to) simplify x11's available_monitors()

psychon:
There are two parts to this.

First, `*monitors_lock` is an `Option<Vec<...>>`. The code here wants to clone the vector so that a copy can be returned to the caller. The way to do this was to use `.as_ref()` to get `Option<&Vec<...>>` so that .cloned() becomes available to produce a clone. In this commit, I am replacing this with a straight `.clone()` call.

Next, there was a `.map(Ok)` to turn `Option<Vec<...>>` into `Option<Result<Vec<...>>`. This resulted either in `None` or `Some(Ok(...))`. `Some(Err(...))` was impossible. Then, `.or_else()` was used to turn the `None` case into `Some(...)`, too, so that while "`None`" is still possible at the type level, it cannot actually occur any more. In the end, `unwrap()` was used to get rid of the outer `Option`.

This code confused me. This second case is actually why I started looking at this. Originally, I only wanted to understand the code.

In my new version, I am using `.ok_or()` to turn `None` into `Err(())` and `Some(...)` into `Ok(...)`. Thus, at this point I actually get rid of the `Option` and have a `Result`, but with the wrong error case. Next, `.or_else()` is used to replace the error with the actual querying of data from the screen. Thus, no final `unwrap()` is necessary any more.

This code is better in the metric "less lines of code", but it still seems complicated to me. Thus, in the end I also added two comments that hopefully make this easier to understand.

  - I added two calls to `event_loop.available_monitors().enumerate().for_each(|x| println!("{x:?"));` to some example. The first such line tests the "cache is empty" case while the second one reaches the "cache is filled" case. The output seems okay.
  - Nothing should actually changes
   - Nothing should actually changes
  - Nothing should actually changes
  - Nothing should actually changes

### DIFF
--- a/src/platform_impl/linux/x11/monitor.rs
+++ b/src/platform_impl/linux/x11/monitor.rs
@@ -284,22 +284,16 @@ impl XConnection {
 
     pub fn available_monitors(&self) -> Result<Vec<MonitorHandle>, X11Error> {
         let mut monitors_lock = self.monitor_handles.lock().unwrap();
-        (*monitors_lock)
-            .as_ref()
-            .cloned()
-            .map(Ok)
-            .or_else(|| {
-                self.query_monitor_list()
-                    .map(|mon_list| {
-                        let monitors = Some(mon_list);
-                        if !DISABLE_MONITOR_LIST_CACHING {
-                            (*monitors_lock) = monitors.clone();
-                        }
-                        monitors
-                    })
-                    .transpose()
-            })
-            .unwrap()
+        match *monitors_lock {
+            Some(ref monitors) => Ok(monitors.clone()),
+            None => {
+                let monitors = self.query_monitor_list()?;
+                if !DISABLE_MONITOR_LIST_CACHING {
+                    *monitors_lock = Some(monitors.clone());
+                }
+                Ok(monitors)
+            }
+        }
     }
 
     #[inline]


src/platform_impl/linux/x11/monitor.rs
```diff
@@ -284,22 +284,19 @@ impl XConnection {
...
-                    .map(|mon_list| {
-                        let monitors = Some(mon_list);
-                        if !DISABLE_MONITOR_LIST_CACHING {
-                            (*monitors_lock) = monitors.clone();
-                        }
-                        monitors
-                    })
-                    .transpose()
+            .clone()
+            .ok_or(())
```
notgull, line 290: Not sure if `ok_or(())` is the cleanest way of handling this combinator

psychon: I'm open for suggestions. I looked at https://doc.rust-lang.org/std/option/enum.Option.html searching for `-> Result` and the options are `ok_or`, `ok_or_else`, `transpose` (not really applicable here), `Debug::fmt` (not applicable), `TryFrom` and `TryInto` blanket implementations (which do not do anything useful here).

So, the alternative that I see is a hand-written `match` (which is actually a big win for readability, I think).
```diff
diff --git a/src/platform_impl/linux/x11/monitor.rs b/src/platform_impl/linux/x11/monitor.rs
index e56f54b5..a44e5458 100644
--- a/src/platform_impl/linux/x11/monitor.rs
+++ b/src/platform_impl/linux/x11/monitor.rs
@@ -284,19 +284,16 @@ impl XConnection {
 
     pub fn available_monitors(&self) -> Result<Vec<MonitorHandle>, X11Error> {
         let mut monitors_lock = self.monitor_handles.lock().unwrap();
-        // Copy the cache contents if they are non-empty (Some) producing Result<_, ()>
-        (*monitors_lock)
-            .clone()
-            .ok_or(())
-            // If the cache was empty, actually send a query
-            .or_else(|()| {
-                self.query_monitor_list().map(|mon_list| {
-                    if !DISABLE_MONITOR_LIST_CACHING {
-                        (*monitors_lock) = Some(mon_list.clone());
-                    }
-                    mon_list
-                })
-            })
+        match *monitors_lock {
+            Some(ref monitors) => Ok(monitors.clone()),
+            None => {
+                let monitors = self.query_monitor_list()?;
+                if !DISABLE_MONITOR_LIST_CACHING {
+                    *monitors_lock = Some(monitors.clone());
+                }
+                Ok(monitors)
+            }
+        }
     }
 
     #[inline]
```

Whatya think? (This patch makes the code even three lines shorter...)

kchibisov (approved):

PR #3297, opened 2023-12-24
Properly implement Debug for `Window` and `EventLoop` types

madsmtm:
Properly implement `Debug` for `EventLoop`, `EventLoopBuilder` and `EventLoopProxy`, and require it as a supertrait of `Window` and `ActiveEventLoop`.

It is especially useful for users to be able to know that `Window` is `Debug` (this allows them to easily make their `AppState` `Debug` too, which they could before https://github.com/rust-windowing/winit/pull/3857).

### DIFF
--- a/examples/child_window.rs
+++ b/examples/child_window.rs
@@ -13,6 +13,7 @@ fn main() -> Result<(), impl std::error::Error> {
     #[path = "util/fill.rs"]
     mod fill;
 
+    #[derive(Debug)]
     struct WindowData {
         window: Box<dyn Window>,
         color: u32,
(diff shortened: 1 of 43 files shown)


daxpedda:
This actually highlights a couple of problems we currently have:
- The reason why we put `SendSyncWrapper` on `Fullscreen` inside of `WindowAttributes` is because on MacOS it doesn't implement `Sync`, maybe we should (can we?) fix that. Because at least right now it's exposed as sync even when it shouldn't.
- `SendSyncWrapper` should actually not implement `Debug`, because this means the inner type, which isn't actually `Send + Sync` might get accessed on the wrong thread, e.g. as in `WindowAttributes`. Same with `Clone`.
- `RawWindowHandle` (in `WindowBuilder::with_parent()`) is fine because it's `unsafe` anyway.

daxpedda:
Addressed this in #3303.

PR #3298, opened 2023-12-24
Remove generic parameter `T` from `EventLoopWindowTarget`

madsmtm:
Fixes https://github.com/rust-windowing/winit/issues/3053.

~Blocked on https://github.com/rust-windowing/winit/pull/3061.~

- [ ] Tested on all platforms changed
- [x] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users
- [x] Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior
- [x] Created or updated an example program if it would help users understand this functionality
- [x] Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented

PR #3299, opened 2023-12-24
Replace `EventLoopWindowTarget` with `ActiveEventLoop`

madsmtm:
This is done for several reasons:
- To in the future allow restricting window creation to `ActiveEventLoop` (which resolves a bunch of long-standing issues on macOS and iOS).
- To avoid the `Deref` on `EventLoop`, which is discouraged by the [C-DEREF](https://rust-lang.github.io/api-guidelines/predictability.html#c-deref) guideline.
- To allow further flexibility in each platform, as we are now no longer required to store `crate::event_loop::EventLoopWindowTarget` in the platform's own `EventLoop`.
  - At some point we could perhaps even change `ActiveEventLoop` to contain `&mut`.

~Blocked on https://github.com/rust-windowing/winit/pull/3298.~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,9 @@ Unreleased` header.
 
 # Unreleased
 
-- **Breaking:** Removed unnecessary generic parameter `T` from `EventLoopWindowTarget`.
+- **Breaking:** Removed `EventLoopWindowTarget<T>`, and replaced it with `ActiveEventLoop<'app>`, which allows Winit to clearly define a boundary between what is possible when the application is not running, and when it is.
+- Added `MaybeActiveEventLoop<'app>` trait to help be generic over `EventLoop<T>` and `ActiveEventLoop<'app>`.
+- Added `ActiveEventLoopExtWayland` and `ActiveEventLoopExtX11` to mimic the similar traits on the `EventLoop`.
 - On Windows, macOS, X11, Wayland and Web, implement setting images as cursors. See the `custom_cursors.rs` example.
   - **Breaking:** Remove `Window::set_cursor_icon`
   - Add `WindowBuilder::with_cursor` and `Window::set_cursor` which takes a `CursorIcon` or `CustomCursor`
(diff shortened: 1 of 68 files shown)


daxpedda:
Let me know if I can help out here, happy to do the rebase.

madsmtm:
Unsure the `MaybeActiveEventLoop` is the best design, but not sure what else we could do?

daxpedda:
> * `MaybeActiveEventLoop` being the best design, but not sure what else we could do?

Do we even need traits like that? It's not even less code for us and I'm not sure what users gain out of this.
> * `CustomCursorBuilder::build`, which kind of event loop should it take?

I don't know what the lowest common denominator is, but for Web either event loop is fine.

madsmtm:
> `MaybeActiveEventLoop` being the best design, but not sure what else we could do?
>
> Do we even need traits like that? It's not even less code for us and I'm not sure what users gain out of this.

It's mostly meant as an intermediary, until we feel confident in disabling user's ability to create windows from outside the running event loop.

But before I feel that we can make that change, it must be _easy_ for them to do the correct thing, and that basically requires https://github.com/rust-windowing/winit/issues/2903.

daxpedda:
> > > `MaybeActiveEventLoop` being the best design, but not sure what else we could do?
> >
> > Do we even need traits like that? It's not even less code for us and I'm not sure what users gain out of this.
> 
> It's mostly meant as an intermediary, until we feel confident in disabling user's ability to create windows from outside the running event loop.

What I meant to say is that we can just implement these things as methods on `EventLoop` and `ActiveEventLoop` directly, they don't have to live in a trait. Whats the use-case exactly for being generic over these two?

madsmtm:
> Whats the use-case exactly for being generic over these two?

None other than keeping `Window::new(event_loop)` working. If we don't want that, but are fine with `event_loop.create_window(window_builder)`, then there is no need for the trait.

daxpedda:
> > Whats the use-case exactly for being generic over these two?
> 
> None other than keeping `Window::new(event_loop)` working. If we don't want that, but are fine with `event_loop.create_window(window_builder)`, then there is no need for the trait.

Right ... sorry ... brain-fart. Now it makes sense :sweat_smile:!
I'm fine with keeping the trait then, I don't mind removing it either but we can do that in a follow-up.

daxpedda:
Replaced with https://github.com/rust-windowing/winit/pull/3447.

PR #3300, opened 2023-12-24
Specify rwh_05 min version 0.5.2

### LINKED ISSUE #3223: Change dependency "raw-window-handle" to exclude the feature "std"
When using feature "rwh_05", a dependency cannot be resolved. Specifically, the dependency "raw-window-handle" version ^0.5 (locked to 0.5.0) does not have a feature flag "std". This feature is only introduced in version 0.5.2, but an attempt for its use is being made for version 0.5.0.

line 62 in winit/Cargo.toml:
rwh_05 = { package = "raw-window-handle", version = "0.5", features = ["std"], optional = true }`

This is fixable by either removing the constraint of using feature "std", or by changing the raw-window-handle version to at least 0.5.2.

alexheretic:
Correct rwh_05 minimum version to support "std" feature. Fixes some scenarios where downstream bins don't `cargo update` and try to build with rwh_05 `0.5.0` or `0.5.1` which both lack the "std" feature required in the winit manifest.

Resolves #3223

- ~Tested on all platforms changed~ rwh_05 version already tested
- ~Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users~ n/a
- ~Updated documentation to reflect any user-facing changes, including notes of platform-specific behavior~ n/a
- ~Created or updated an example program if it would help users understand this functionality~ n/a
- ~Updated [feature matrix](https://github.com/rust-windowing/winit/blob/master/FEATURES.md), if new features were added or implemented~ n/a

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,7 +66,7 @@ log = "0.4"
 mint = { version = "0.5.6", optional = true }
 once_cell = "1.12"
 rwh_04 = { package = "raw-window-handle", version = "0.4", optional = true }
-rwh_05 = { package = "raw-window-handle", version = "0.5", features = ["std"], optional = true }
+rwh_05 = { package = "raw-window-handle", version = "0.5.2", features = ["std"], optional = true }
 rwh_06 = { package = "raw-window-handle", version = "0.6", features = ["std"], optional = true }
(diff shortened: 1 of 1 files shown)


notgull (approved):
Thanks!

Issue #3258, opened 2023-12-06
Windows: `MonitorHandle` can become invalid leading to panics


daxpedda:
The `MonitorHandle` on Windows holds a  [`HMONITOR`](https://learn.microsoft.com/en-us/windows/win32/gdi/hmonitor-and-the-device-context), which can become invalid if e.g. the monitor is disconnected. Or to be more precise, the Windows documentation recommends to listen to `WM_DISPLAYCHANGE`, which can potentially invalidate any `HMONITOR`.

All of the `MonitorHandle` methods call the [`GetMonitorInfo`](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getmonitorinfoa) function and just `unwrap()`, which would cause a panic if a `HMONITOR` has become invalid.

See #3255, which only fixed `MonitorHandle::video_modes()`.

PR #3303, opened 2023-12-24
Remove unsound `SendSyncWrapper`

daxpedda:
I noticed that `SendSyncWrapper` caused some real issues here and there. First #3270 and #3288, then #3297.

Also noticed that we simply used `SendSyncWrapper` on `Fullscreen` inside `WindowBuilder`, but there is a getter that exposes this information to the user, bypassing the `!Sync` on `Fullscreen` on MacOS, depending if this `!Sync` was correct or not, that may have been unsound.

Additionally on Web there was an issue where dropping `WindowBuilder` outside the main thread could cause it to try and drop `HtmlCanvasElement`, which is unsound as well (not entirely sure if that is correct, but it's at least semantically correct).

Additionally implementing `Debug` on `SendSyncWrapper` was unsound as well, because `Debug` could access the inner value off the only thread it was intended to be accessed on. Which was the case for `WindowBuilder` on Web with `HtmlCanvasElement` and the custom wrapper I used for `CustomCursor`.

All in all I would say `SendSyncWrapper` is a footgun and apart from the `RawWindowHandle` v0.4 and v0.5 we didn't even really need it.

### DIFF
--- a/src/platform_impl/macos/monitor.rs
+++ b/src/platform_impl/macos/monitor.rs
@@ -1,6 +1,10 @@
 #![allow(clippy::unnecessary_cast)]
 
-use std::{collections::VecDeque, fmt};
+use std::{
+    collections::VecDeque,
+    fmt,
+    sync::{Arc, Mutex},
+};
 
 use core_foundation::{
     array::{CFArrayGetCount, CFArrayGetValueAtIndex},
@@ -23,7 +27,7 @@ pub struct VideoMode {
     bit_depth: u16,
     refresh_rate_millihertz: u32,
     pub(crate) monitor: MonitorHandle,
-    pub(crate) native_mode: NativeDisplayMode,
+    pub(crate) native_mode: Arc<Mutex<NativeDisplayMode>>,
 }
 
 impl PartialEq for VideoMode {
@@ -290,7 +294,7 @@ impl MonitorHandle {
                     refresh_rate_millihertz,
                     bit_depth,
                     monitor: monitor.clone(),
-                    native_mode: NativeDisplayMode(mode),
+                    native_mode: Arc::new(Mutex::new(NativeDisplayMode(mode))),
                 }
             })
         }
--- a/src/platform_impl/macos/monitor.rs
+++ b/src/platform_impl/macos/monitor.rs
@@ -1,10 +1,6 @@
 #![allow(clippy::unnecessary_cast)]
 
-use std::{
-    collections::VecDeque,
-    fmt,
-    sync::{Arc, Mutex},
-};
+use std::{collections::VecDeque, fmt};
 
 use core_foundation::{
     array::{CFArrayGetCount, CFArrayGetValueAtIndex},
@@ -27,7 +23,7 @@ pub struct VideoMode {
     bit_depth: u16,
     refresh_rate_millihertz: u32,
     pub(crate) monitor: MonitorHandle,
-    pub(crate) native_mode: Arc<Mutex<NativeDisplayMode>>,
+    pub(crate) native_mode: NativeDisplayMode,
 }
 
 impl PartialEq for VideoMode {
@@ -64,6 +60,7 @@ impl std::fmt::Debug for VideoMode {
 pub struct NativeDisplayMode(pub ffi::CGDisplayModeRef);
 
 unsafe impl Send for NativeDisplayMode {}
+unsafe impl Sync for NativeDisplayMode {}
 
 impl Drop for NativeDisplayMode {
     fn drop(&mut self) {
@@ -294,7 +291,7 @@ impl MonitorHandle {
                     refresh_rate_millihertz,
                     bit_depth,
                     monitor: monitor.clone(),
-                    native_mode: Arc::new(Mutex::new(NativeDisplayMode(mode))),
+                    native_mode: NativeDisplayMode(mode),
                 }
             })
         }
--- a/src/platform_impl/macos/window.rs
+++ b/src/platform_impl/macos/window.rs
@@ -3,7 +3,6 @@
 use std::collections::VecDeque;
 use std::f64;
 use std::ops;
-use std::sync::PoisonError;
 use std::sync::{Mutex, MutexGuard};
 
 use crate::{
@@ -114,13 +113,9 @@ impl Window {
     pub(crate) fn raw_display_handle_rwh_06(
         &self,
     ) -> Result<rwh_06::RawDisplayHandle, rwh_06::HandleError> {
-        struct UnsafeSendWrapper<T>(T);
-
-        unsafe impl<T> Send for UnsafeSendWrapper<T> {}
-
-        Ok(self
-            .maybe_wait_on_main(|w| UnsafeSendWrapper(w.raw_display_handle_rwh_06()))
-            .0)
+        Ok(rwh_06::RawDisplayHandle::AppKit(
+            rwh_06::AppKitDisplayHandle::new(),
+        ))
     }
 }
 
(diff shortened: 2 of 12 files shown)


src/platform_impl/macos/monitor.rs
```diff
@@ -23,7 +27,7 @@ pub struct VideoMode {
     bit_depth: u16,
     refresh_rate_millihertz: u32,
     pub(crate) monitor: MonitorHandle,
-    pub(crate) native_mode: NativeDisplayMode,
+    pub(crate) native_mode: Arc<Mutex<NativeDisplayMode>>,
```
daxpedda, line 30: This how I made this `Sync` and wrapped in in `Arc` to make it `Clone` as well.
Let me know what you think.

src/platform_impl/macos/window.rs
```diff
@@ -105,6 +106,20 @@ impl Window {
...
+        &self,
+    ) -> Result<rwh_06::RawDisplayHandle, rwh_06::HandleError> {
+        struct UnsafeSendWrapper<T>(T);
+
+        unsafe impl<T> Send for UnsafeSendWrapper<T> {}
+
+        Ok(self
+            .maybe_wait_on_main(|w| UnsafeSendWrapper(w.raw_display_handle_rwh_06()))
+            .0)
+    }
```
daxpedda, line 122: Like in #3288 this might have been incorrect to begin with.

kchibisov:
>Additionally implementing Debug on SendSyncWrapper was unsound as well, because Debug could access the inner value off the only thread it was intended to be accessed on. Which was the case for WindowBuilder on Web with HtmlCanvasElement and the custom wrapper I used for CustomCursor.

reading pointer value is not unsound by any means, if you deref raw pointers during `Debug` it's already concerning.

>Also noticed that we simply used SendSyncWrapper on Fullscreen inside WindowBuilder, but there is a getter that exposes this information to the user, bypassing the !Sync on Fullscreen on MacOS, depending if this !Sync was correct or not, that may have been unsound.

you can't do anything with it outside the main thread, and if you can do then builder has nothing to do with it, we have the same API throughout the code base on a Send + Sync types with the exact same values.

daxpedda:
> reading pointer value is not unsound by any means, if you deref raw pointers during `Debug` it's already concerning.

I think you are talking about `RawWindowHandle`. `HtmlCanvasElement` is not a pointer.
But that is the footgun I was talking about, we were using `SendSyncWrapper` on types that aren't a pointer.

> you can't do anything with it outside the main thread, and if you can do then builder has nothing to do with it, we have the same API throughout the code base on a Send + Sync types with the exact same values.

I think there is a misunderstanding here.
`WindowBuilder` is `Send + Sync`, the public API allows you to extract the `Fullscreen` type from the `WindowBuilder` and access it. Now the user can do this over multiple threads at the same time, because `WindowBuilder` is `Send + Sync`, and `Fullscreen` grants access to a `VideoMode`, which has a bunch of methods that users could call on a type in parallel that isn't `Sync`.

If this isn't an issue, then `VideoMode` should be `Sync`, making `Fullscreen` `Sync` automatically.

kchibisov:
>I think there is a misunderstanding here.
WindowBuilder is Send + Sync, the public API allows you to extract the Fullscreen type from the WindowBuilder and access it. Now the user can do this over multiple threads at the same time, because WindowBuilder is Send + Sync, and Fullscreen grants access to a VideoMode, which has a bunch of methods that users could call on a type in parallel that isn't Sync.

>If this isn't an issue, then VideoMode should be Sync, making Fullscreen Sync automatically.

Window is `Send + Sync` as well and uses the same API, if macOS is bugged macOS should be fixed, nothing.

daxpedda:
> Window is `Send + Sync` as well and uses the same API, if macOS is bugged macOS should be fixed, nothing.

This isn't the same thing, `Window` doesn't return references to `Fullscreen`, but owned values, the resulting `Fullscreen` is then `!Sync`, so no violations there.

I hope the problem is apparent now, `SendSynWrapper` is just a footgun because it's not simple to reason about these things, especially when we have so many very different backends.

src/platform_impl/macos/window.rs
madsmtm: I mean, the display handle is empty on macOS, so it probably doesn't really matter?

src/window.rs
```diff
@@ -181,22 +181,32 @@ impl Default for WindowAttributes {
...
 
+/// Wrapper for [`rwh_06::RawWindowHandle`] for [`WindowAttributes::parent_window`].
+///
+/// # Safety
+///
+/// The user has to account for that when using [`WindowBuilder::with_parent_window()`],
+/// which is `unsafe`.
+#[derive(Debug, Clone)]
+#[cfg(feature = "rwh_06")]
+pub(crate) struct SendSyncRawWindowHandle(pub(crate) rwh_06::RawWindowHandle);
```
madsmtm, line 198: See also https://github.com/rust-windowing/winit/pull/3142 for improvements here.

madsmtm (commented):
Flyby review, unsure what's the correct approach here.

src/platform_impl/macos/monitor.rs
madsmtm: I'd really like to address https://github.com/rust-windowing/winit/pull/2646, since that would give me a clearer picture of what's expected from `VideoMode`: Is it just a static blob of data? If so, then we should perhaps remove the `native_mode` field? If not, then perhaps nothing should be cached, and it'd be more clear what to do here.

In any case, I think `CGDisplayModeRef` (and that means `NativeDisplayMode`) may actually be `Send + Sync` natively, by the argument that you can access global instances (e.g. `CGDisplay::main().display_mode()`) from any thread, and that the data is immutable (at least from the view of the user).

daxpedda: Don't we need some handle to live in `VideoMode` to be able to spawn the window in specific monitors and such? In any case, I'm pretty much clueless about this. Currently we have a couple of types that should be connected to the `EventLoop` but aren't, e.g. like `VideoMode` and `CustomCursor`, so they both need the same solution.

There is a connection to #3317 here, which also requires some connection to `EventLoop`, one proposal is to introduce a lifetime, which could be introduced to `VideoMode` as well.

I guess if it's possible to make this just a static blob of data, it would be by far the best solution in my opinion.

In any case, in the meantime, I will just implement `Sync` on `NativeDisplayMode` and await your confirmation.

src/platform_impl/macos/window.rs
daxpedda: Ah, right, I can simplify this whole thing then.

daxpedda:
I split out the Web part into #3320.
So now all that needs to be done here is figure out `Sync` for `VideoMode` for MacOS.

src/platform_impl/macos/monitor.rs
madsmtm: > need some handle to live in `VideoMode` to be able to spawn the window in specific monitors and such? 

Right. Maybe we should rename it `VideoModeHandle` to reflect that it isn't just static data?

> lifetime on `VideoMode`

Not really required for macOS, since the stuff is `'static` and able to live after `EventLoop` has been dropped.

> implement `Sync` on `NativeDisplayMode` and await your confirmation.

I'll accept that as "sound enough" for now.

daxpedda: > > need some handle to live in `VideoMode` to be able to spawn the window in specific monitors and such?
> 
> Right. Maybe we should rename it `VideoModeHandle` to reflect that it isn't just static data?

I think it's a good idea! I started tracking this here: #3258.
Will actually make a PR afterwards.

> > implement `Sync` on `NativeDisplayMode` and await your confirmation.
> 
> I'll accept that as "sound enough" for now.

Would you mind giving your approval to the PR then? Just to make it official :sweat_smile:.

madsmtm (approved):

PR #3307, opened 2023-12-24
On Web, add `Window::(set_)prevent_default()`

daxpedda:
Current controlling if `Event.preventDefault()` is called or not could only be changed in `WindowBuilder`.
This PR introduces `Window::set_prevent_default()`, so now it can be dynamically adjusted.


notgull (approved):
Looks alright to me

Issue #3306, opened 2023-12-24
Custom cursor tracking issue


daxpedda:
Custom cursors are still missing some refinements.

#### Known Issues
- [x] On Web, correctly handle multiple windows waiting on the same cursor: #3321.
- [x] On Wayland, account for alpha: #3360.
- [ ] Figure out how to prevent `CustomCursor`s to outlive the event loop. This is similar to #3317 and #3258.

#### Improvements
- [x] Add back `Window::set_cursor_icon()` as a deprecated method: #3329.
- [x] Merge icons and cursors in one type and one method: https://github.com/rust-windowing/winit/pull/3308.
- [x] Add a cursor getter/setter on `WindowBuilder`: #3319.
- [ ] Add a cursor getter on `Window`.
- [ ] Make sure implementations check if the new cursor being set isn't the same already set.
  - [x] MacOS: #3324.
  - [ ] Wayland
  - [x] Web: #3321.
  - [ ] Windows
  - [ ] X11
- [ ] Return a `Result` at `CustomCursorBuilder::build()`, AFAIK only Windows needs it.
- [ ] On Windows, extension method to load from a resource, like in [`Icon::from_resource()`](https://docs.rs/winit/0.29.6/x86_64-pc-windows-msvc/winit/platform/windows/trait.IconExtWindows.html#tymethod.from_resource).
- [x] On Web, `async fn build()` extension method to wait until the cursor has successfully finished loading: #3384.

#### Caching
- [x] MacOS: #3291
- [ ] Wayland
- [x] Web: #3276
- [x] Windows: #3293
- [x] X11: #3366

#### Animation
- [x] Web: #3384
- [ ] Windows: support ANI files (loaded from memory or otherwise).
- [ ] MacOS: Use `NSTimer`? (idea from [Firefox](https://source.chromium.org/chromium/chromium/src/+/25cbdb930951c1b8b7863351c45c340fcc95fcc4:ui/base/cocoa/cursor_utils.mm;l=275-285;bpv=0;bpt=0))
- [ ] X11: `XcursorAnimate`

kchibisov:
On Wayland you animate the cursor yourself and you generally supply the Xcursor formatted cursor.

So you should have a bunch of _frames_ and delays between the frames.

kchibisov:
That's how it looks.

https://www.x.org/releases/X11R7.7/doc/man/man3/Xcursor.3.xhtml

So basically  the custom cursor is

```rust
struct CustomCursor {
[...]

PR #3308, opened 2023-12-24
Merge `Window::set_cursor_icon()` and `Window::set_custom_cursor()`

daxpedda:
This merges the two methods `Window::set_cursor_icon()` and `Window::set_custom_cursor()`.
The underlying implementations were not changed at all!

This simplifies the API a bit so we don't have two methods that override each other, in the future this will allow a getter to be implemented when Linux implements caching, see #3306.

In a follow-up I will implement a custom cursor setting in `WindowBuilder`, the merged type allows `WindowAttributes` to easily expose the currently set cursor.

### DIFF
--- a/src/window.rs
+++ b/src/window.rs
@@ -9,7 +9,7 @@ use crate::{
     platform_impl, SendSyncWrapper,
 };
 
-pub use crate::cursor::{BadImage, CustomCursor, CustomCursorBuilder, MAX_CURSOR_SIZE};
+pub use crate::cursor::{BadImage, Cursor, CustomCursor, CustomCursorBuilder, MAX_CURSOR_SIZE};
 pub use crate::icon::{BadIcon, Icon};
 
 #[doc(inline)]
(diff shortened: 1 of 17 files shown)


notgull (approved):
LGTM to me for X11

src/window.rs
```diff
@@ -1336,28 +1336,15 @@ impl Window {
...
     /// Modifies the cursor icon of the window.
-    /// Overwrites cursors set in [`Window::set_custom_cursor`].
     ///
     /// ## Platform-specific
     ///
     /// - **iOS / Android / Orbital:** Unsupported.
     #[inline]
-    pub fn set_cursor_icon(&self, cursor: CursorIcon) {
+    pub fn set_cursor(&self, cursor: impl Into<Cursor>) {
+        let cursor = cursor.into().clone();
```
madsmtm, line 1345: Nit:

```suggestion
        let cursor = cursor.into();
```

madsmtm (approved):

src/window.rs
daxpedda: Right, thank you!

PR #3309, opened 2023-12-24
Reload Xft database on DPI change

### LINKED ISSUE #1228: Allow changing X11 DPI at runtime

notgull:
Closes #1228


daxpedda:
Can't really review it, but pretty sick!

kchibisov (approved):

frebib:
This still doesn't work (at least as far as I'm able to test). I'm testing with Alacritty 0.13.0.

With my old patch, Alacritty logged
```
[22.110655316s] [INFO ] [alacritty] Font size changed to Size(18) with scale factor of 1
[22.110708195s] [INFO ] [alacritty] Cell size: 7 x 16
[22.121738413s] [INFO ] [alacritty] Padding: 22 x 22
[22.121769169s] [INFO ] [alacritty] Width: 728, Height: 808
```
when I ran `xrdb -merge <<< "Xft.dpi: 120"` (note I'm also running xsettingsd as well as running `xrandr --dpi 120`

but with the current upstream (this PR) code I only see these lines logged
```
[9.607223003s] [INFO ] [alacritty] Padding: 22 x 22
[9.607228274s] [INFO ] [alacritty] Width: 910, Height: 1010
```
with the same test.

I don't think the X resources db is actually being reloaded here?

kchibisov:
@frebib have you tried 0.13.1?

frebib:
Yeah, just tried. Same issue.

`dpi 96`:
```
[0.095471876s] [DEBUG] [alacritty_terminal] New num_cols is 125 and num_lines is 61
[0.097583912s] [INFO ] [alacritty] Padding: 22 x 22
[0.097604573s] [INFO ] [alacritty] Width: 920, Height: 1024
```
`dpi 144`:
```
[11.309989434s] [DEBUG] [alacritty_terminal] New num_cols is 122 and num_lines is 59
[11.310289646s] [INFO ] [alacritty] Padding: 22 x 22
[11.310302752s] [INFO ] [alacritty] Width: 900, Height: 996
```
The slight change in width/height/lines/cols is because the window shrinks slightly due to increased padding in the window manager at higher DPIs

PR #3310, opened 2023-12-24
Remove `extern crate` statements

daxpedda:
This was phased out in the 2018 edition for better hygiene, see [The Rust Edition Guide](https://doc.rust-lang.org/edition-guide/rust-2018/path-changes.html#no-more-extern-crate).


notgull (approved):
LGTM to me for X11.

We should figure out if `log::warn!(...)` or if `use log::warn; warn!(...)` is preferred for winit.

madsmtm (approved):
I'm fine with whatever here

daxpedda:
> We should figure out if `log::warn!(...)` or if `use log::warn; warn!(...)` is preferred for winit.

I generally went with the rule that when there is more then one call I will add a `use` statement.

PR #3315, opened 2023-12-24
Document `Window` `Drop` behavior

daxpedda:
Would appreciate any improvement on the wording.

### DIFF
--- a/src/window.rs
+++ b/src/window.rs
@@ -17,6 +17,7 @@ pub use cursor_icon::{CursorIcon, ParseError as CursorIconParseError};
 
 /// Represents a window.
 ///
+/// The window is closed when dropped.
 ///
 /// # Threading
 ///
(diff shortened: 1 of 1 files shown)


src/window.rs
```diff
@@ -62,6 +67,9 @@ impl fmt::Debug for Window {
     }
 }
 
+/// This will close the [`Window`].
+///
+/// See [`Window`] for more details.
```
notgull, line 72: I think the general trend is to put documentation on the `drop()` method instead of the trait implementation.

src/window.rs
```diff
@@ -17,6 +17,7 @@ pub use cursor_icon::{CursorIcon, ParseError as CursorIconParseError};
 
 /// Represents a window.
 ///
+/// To close the [`Window`] it can be dropped.
```
madsmtm, line 20:
```suggestion
/// The window is closed when dropped.
```

Maybe?

src/window.rs
daxpedda: Uh, true, but you aren't supposed to actually call `Drop::drop()`? Looking at `std` I couldn't find any `Drop` implementation documented.

Will go ahead with moving the documentation to the method.

src/window.rs
daxpedda: Thanks!

daxpedda:
@kchibisov why was this moved to 0.31?
I think we can merge this now?

kchibisov:
I thought that it was more the lifecycle and how we should probably retain stuff? But if you want the said wording it's fine as is.

PR #3324, opened 2023-12-26
MacOS: check if cursor changed before applying

daxpedda:
This just adds a check to make sure the cursor actually changed before applying it.

Addresses #3306.


daxpedda:
I was just thinking, that when we implement `Window::cursor()`, we can actually move this check into `Window::set_cursor()` directly.

Let me know if you would rather wait for that.

madsmtm (approved):
> Let me know if you would rather wait for that.

Nah, I think it's fine like this

PR #3328, opened 2023-12-26
Rename `VideoMode` to `VideoModeHandle`

daxpedda:
The point here is to reflect that `VideoMode` is not a static blob of information, but a handle that dynamically accesses that information. This is in line with our other `MonitorHandle` type.
Addresses #3258.

Cc @madsmtm.

### DIFF
--- a/src/monitor.rs
+++ b/src/monitor.rs
@@ -10,28 +10,32 @@ use crate::{
     platform_impl,
 };
 
+/// Deprecated! Use `VideoModeHandle` instead.
+#[deprecated = "Renamed to `VideoModeHandle`"]
+pub type VideoMode = VideoModeHandle;
+
 /// Describes a fullscreen video mode of a monitor.
 ///
 /// Can be acquired with [`MonitorHandle::video_modes`].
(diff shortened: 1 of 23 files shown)


src/monitor.rs
```diff
@@ -14,24 +14,24 @@ use crate::{
 ///
 /// Can be acquired with [`MonitorHandle::video_modes`].
 #[derive(Clone, PartialEq, Eq, Hash)]
-pub struct VideoMode {
-    pub(crate) video_mode: platform_impl::VideoMode,
+pub struct VideoModeHandle {
+    pub(crate) video_mode: platform_impl::VideoModeHandle,
 }
```
madsmtm, line 20: Maybe add:
```rust
#[deprecated = "Renamed to `VideoModeHandle`"]
type VideoMode = VideoModeHandle;
```

And keep it around for a release cycle?

madsmtm (approved):

src/monitor.rs
daxpedda: I don't mind.

daxpedda: Done.

PR #3347, opened 2023-12-31
Bump web-sys to 0.3.66

YatesRocks:

Compiles, cargo test returns no errors


daxpedda:
What is the motivation behind this?
This is just supposed to be the minimal version required, when you depend on Winit you can just use `cargo update`.

PR #3353, opened 2024-01-03
Android: allow EventLoop recreation

### LINKED ISSUE #3325: Android `RecreationAttempt` on Activity reopen
When an android app calls `event_loop.exit()` the application exits, but if the application is opened again without closing it in the "recent apps" the EventLoop creation will fail with a `RecreationAttempt` error.

## Steps to reproduce
- clone https://github.com/SnowyCoder/winit-bug-reopen-example
- run in an android phone
- press back button
- reopen app

## Reported error
```called `Result::unwrap()` on an `Err` value: RecreationAttempt```

## Repo explaination
The repo linked in the "Steps to reproduce" section is a modified version of the one linked in the README android section.
I just updated the dependencies and linked `BrowserBack` to `event_loop.exit()` (that sems to be the logical key invoked when the android back button is pressed).

## Comment
The general problem seems to be that `android_main` can be called more than once without unloading and reloading the native library (ex. if we use an application with two interacting activities, say LoginActivity and GameActivity).

SnowyCoder:

I've added this as the default behavior because it's what is expected in android when you call `event_loop.exit()` and reopening the activity, is it better suited for a sort of `EventLoopBuilderExtAndroid`?

Closes #3325

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Unreleased` header.
 
 # Unreleased
 
+- On Android, allow EventLoop recreation after destruction
 - On X11 and Wayland, fix arrow up on keypad reported as `ArrowLeft`.
 - On Windows, macOS, X11, Wayland and Web, implement setting images as cursors. See the `custom_cursors.rs` example.
   - **Breaking:** Remove `Window::set_cursor_icon`
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -106,6 +106,9 @@ impl<T> EventLoopBuilder<T> {
     ///                    or `DISPLAY` respectively when building the event loop.
     /// - **Android:** must be configured with an `AndroidApp` from `android_main()` by calling
     ///     [`.with_android_app(app)`] before calling `.build()`, otherwise it'll panic.
+    ///     Due to android platform-specific behaviour, it's possible to recreate the EventLoop
+    ///     after the previous one has been destroyed, this is to prevent bugs when multiple
+    ///     activities are used.
     ///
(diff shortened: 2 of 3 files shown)


MarijnS95 (changes requested):
Simply allowing recreation hides a poor design decision to not fix how `winit` and its auxiliary crates mismanage Android's `Activity` Lifecycle, even in the probably close to ~4th refactor of the Android backend with the introduction of the `android-activity` crate. Despite multiple requests to tackle this problem from the get-go, which is expected to have far-stretching effects on the rest of the implementation, it has been ignored on every refactor.

The issue and a simplified solution (tl;dr: move towards handling the `Activity` lifecycle appropriately) is better described in https://github.com/rust-windowing/winit/issues/3325#issuecomment-1879742860.

Since this PR proposes a workaround in the opposite direction by papering over the issue rather than solving towards it, and above all fails to explain `"Due to android platform-specific behaviour"` what the actual problem is _nor any of the drawbacks that this workaround is causing_, I think we should close this PR and work towards implementing my proposed changes from the bottom up.

daxpedda:
Agreed.

SnowyCoder:
Sorry for the trouble, I created this Pull Request to propose a possible solution before discussing the proper solution and before receiving the mentioned comment.
I no longer have the time required to help write a real solution.

MarijnS95:
No worries at all, we're just closing this (as said) to communicate that this is not the direction we want nor are planning to head into.

We don't need any help writing a solution, changes for that are already in progress by the respective winit and/or Android maintainers.

PR #3359, opened 2024-01-04
Deploy master docs to GitHub Pages

daxpedda:
This deploys documentation generated from the master branch to GitHub Pages.
I also added the extra couple of lines needed to [scrape examples](https://doc.rust-lang.org/1.75.0/rustdoc/scraped-examples.html), which should also be visible on docs.rs.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,50 @@
+name: Docs
+
+on:
+  push:
+    branches: [master]
+
+jobs:
+  docs:
+    name: Documentation
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}winit
+
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: read
+      pages: write
+      id-token: write
+    
+    steps:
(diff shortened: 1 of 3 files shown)


madsmtm:
I fear this may confuse users on which docs they need to look at (which is almost always the published one on `docs.rs`), so unless you have some grander plan with this, I think I'd rather not do this?

daxpedda:
> I fear this may confuse users on which docs they need to look at (which is almost always the published one on `docs.rs`), so unless you have some grander plan with this, I think I'd rather not do this?

The only place this is exposed to anybody is a badge on the README saying "master docs" right next to the badge saying "docs", I could improve the "docs" badge saying "crates.io docs". I find this very clear.

But if this is still not clear enough, I can just bury it somewhere else in the README saying "here are the docs to current master branch [URL]".

The purpose here is to increase the visibility of the documentation we are currently working on and exactly how it looks like, especially after #3076.

notgull (approved):
LGTM to me. This is done in a couple of other repos whose names escape me at the moment. Since we have a few users using the master branch of winit this is probably helpful

Issue #3078, opened 2023-09-02
CMD+Shift is not reported correctly on macOS


fredizzimo:
The corresponding bug in Neovide is https://github.com/neovide/neovide/issues/1237

I have been wanting report this bug for quite some time now, since it has caused a pretty bad regression in Neovide, and has a big impact for some users. But unfortunately, I don't own a Mac, so it has been difficult to gather the information I need to report it.

The issue is that CMD + character reports the same text as CMD + Shift + character, the text always contains the base character. This does not play well with Vim, which assumes that you map the resulting characters, not a key + modifiers. We do know from the modifiers that shift is pressed, but that information can't really be used without knowing the keymap, it's possible to turn standard alpha keys to uppercase, but in combination with the number keys for example, we don't have any idea what character to send to Neovim.

Here's a trace captured through the Neovide logging, where `CMD+shift+d` was pressed.
```
TRACE [neovide::window::keyboard_manager] KeyEvent {
    physical_key: SuperLeft,
    logical_key: Super,
    text: None,
    location: Left,
    state: Pressed,
    repeat: false,
    platform_specific: KeyEventExtra {
        text_with_all_modifiers: None,
        key_without_modifiers: Super,
    },
}
TRACE [neovide::redraw_scheduler] Next frame queued
TRACE [neovide::window::keyboard_manager] KeyEvent {
    physical_key: ShiftLeft,
    logical_key: Shift,
    text: None,
    location: Left,
    state: Pressed,
    repeat: false,
    platform_specific: KeyEventExtra {
        text_with_all_modifiers: None,
        key_without_modifiers: Shift,
    },
}
TRACE [neovide::window::keyboard_manager] Modifiers { state: ModifiersState(SHIFT | SUPER), pressed_mods: ModifiersKeys(LSHIFT | LSUPER) }
TRACE [neovide::window::keyboard_manager] KeyEvent {
    physical_key: KeyD,
    logical_key: Character(
        "d",
    ),
    text: Some(
        "d",
    ),
    location: Standard,
    state: Pressed,
    repeat: false,
    platform_specific: KeyEventExtra {
        text_with_all_modifiers: Some(
            "d",
        ),
        key_without_modifiers: Character(
            "d",
        ),
    },
}
```
[...]

PR #3361, opened 2024-01-05
On macOS, report correct logical key when Ctrl or Cmd is pressed

emilk:
* Closes https://github.com/rust-windowing/winit/issues/3078

### DIFF
--- a/src/platform_impl/macos/event.rs
+++ b/src/platform_impl/macos/event.rs
@@ -28,6 +28,7 @@ pub struct KeyEventExtra {
     pub key_without_modifiers: Key,
 }
 
+/// Ignores ALL modifiers.
 pub fn get_modifierless_char(scancode: u16) -> Key {
     let mut string = [0; 16];
     let input_source;
@@ -86,7 +87,7 @@ pub fn get_modifierless_char(scancode: u16) -> Key {
     Key::Character(SmolStr::new(chars))
 }
 
-// Ignores all modifiers except for shift
+// Ignores all modifiers except for SHIFT (yes, even ALT is ignored).
 fn get_logical_key_char(ns_event: &NSEvent, modifierless_chars: &str) -> Key {
     let string = unsafe { ns_event.charactersIgnoringModifiers() }
         .map(|s| s.to_string())
@@ -147,7 +148,7 @@ pub(crate) fn create_key_event(
 
     let key_from_code = code_to_key(physical_key, scancode);
     let (logical_key, key_without_modifiers) = if matches!(key_from_code, Key::Unidentified(_)) {
-        // `key_without_modifiers` ignores all modifiers except for shift.
+        // `get_modifierless_char/key_without_modifiers` ignores ALL modifiers.
         let key_without_modifiers = get_modifierless_char(scancode);
 
         let modifiers = unsafe { ns_event.modifierFlags() };
@@ -160,11 +161,16 @@ pub(crate) fn create_key_event(
             // presses alt+8, the logical key should be "{"
             // Also not checking if this is a release event because then this issue would
             // still affect the key release.
-            Some(text) if !has_ctrl && !has_cmd => Key::Character(text.clone()),
+            Some(text) if !has_ctrl && !has_cmd => {
+                // Character heeding both SHIFT and ALT.
+                Key::Character(text.clone())
+            }
 
             _ => match key_without_modifiers.as_ref() {
+                // Character heeding just SHIFT, ignoring ALT.
                 Key::Character(ch) => get_logical_key_char(ns_event, ch),
-                // Don't try to get text for events which likely don't have it.
+
+                // Character ignoring ALL modifiers.
                 _ => key_without_modifiers.clone(),
             },
         };
--- a/src/platform_impl/macos/event.rs
+++ b/src/platform_impl/macos/event.rs
@@ -115,18 +115,12 @@ pub(crate) fn create_key_event(
     let scancode = unsafe { ns_event.keyCode() };
     let mut physical_key = key_override.unwrap_or_else(|| scancode_to_physicalkey(scancode as u32));
 
-    // NOTE:
-    // The logical key should be the key resulting from applying all modifiers.
-    // Consider these cases:
-    // 1) Pressing the A key: logical key should be "a"
-    // 2) Pressing SHIFT A: logical key should be "A"
-    // 3) Pressing CTRL SHIFT A: logical key should also be "A"
-    // `ns_event.characters()` gets 3) wrong, returning "a" instead of "A".
-    // In fact it gets it wrong anytime CTRL or CMD is pressed.
-    // `ns_event.charactersIgnoringModifiers()` gets everything right,
-    // because it ignores all modifiers except for shift.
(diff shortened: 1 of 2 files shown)


src/platform_impl/macos/event.rs
```diff
@@ -116,7 +116,16 @@ pub(crate) fn create_key_event(
...
+        // The logical key should always be the key resulting from applying all modifiers.
+        // Consider these cases:
+        // 1) Pressing the A key: logical key should be "a"
+        // 2) Pressing SHIFT A: logical key should be "A"
+        // 3) Pressing CTRL SHIFT A: logical key should also be "A"
+        // `ns_event.characters()` gets 3) wrong, returning "a" instead of "A".
+        // In fact it gets it wrong anytime CTRL or CMD is pressed.
+        // `ns_event.charactersIgnoringModifiers()` gets everything right.
+        // See https://github.com/rust-windowing/winit/issues/3078
+        let characters = unsafe { ns_event.charactersIgnoringModifiers() }
```
kchibisov, line 128: But this doesn't make any sense though? Shift is a modifier, character ignoring modifier is just `a`, if it returns `A` it means that it accounts for shift, thus the function doesn't ignore the modifier?

This also changes the logic how `Alt+e` works and now it's always like with `OptionAsAlt`, I think. Maybe it's not bad, but it's a breaking change in behavior.

emilk: I agree that the name `charactersIgnoringModifiers` makes no sense for how it works (I didn't name it!), I'm just reporting how it actually works on macOS Sonoma

kchibisov: Hm, but Apple docs claim that they should remove the mods, like what I said is from their docs, mostly, and I know that the `Alt` behavior will change a lot, though maybe for better.

kchibisov: Yeah, but we have a `key_without_modifiers` just for that... Though, the key_without_modifier is a weird in a sense that it removes the iso shift levels on linux, so you have to use the logical key stuff.

Also, what you get in `key_without_modifier` on macOS when you press `cmd + shift`?

emilk: `charactersIgnoringModifiers` ignores all modifiers _except for shift_: https://developer.apple.com/documentation/appkit/nsevent/1524605-charactersignoringmodifiers

emilk: …but reading the rest of the code in `create_key_event` I see a lot of convoluted logic that I don't quite follow

src/platform_impl/macos/event.rs
```diff
@@ -113,6 +114,20 @@ pub(crate) fn create_key_event(
...
+    // Consider these cases:
+    // 1) Pressing the A key: logical key should be "a"
+    // 2) Pressing SHIFT A: logical key should be "A"
+    // 3) Pressing CTRL SHIFT A: logical key should also be "A"
+    // `ns_event.characters()` gets 3) wrong, returning "a" instead of "A".
+    // In fact it gets it wrong anytime CTRL or CMD is pressed.
+    // `ns_event.charactersIgnoringModifiers()` gets everything right,
+    // because it ignores all modifiers except for shift.
+    // https://developer.apple.com/documentation/appkit/nsevent/1524605-charactersignoringmodifiers
+    // https://github.com/rust-windowing/winit/issues/3078
```
kchibisov, line 128: Is this really necessary now? It at least should be moved to relevant place if it really is.

src/platform_impl/macos/event.rs
```diff
@@ -86,6 +86,7 @@ pub fn get_modifierless_char(scancode: u16) -> Key {
     Key::Character(SmolStr::new(chars))
 }
 
+// Ignores all modifiers except for shift
```
kchibisov, line 89: Don't think it's needed or accurate, since Alt is still applied.

src/platform_impl/macos/event.rs
```diff
@@ -132,18 +147,21 @@ pub(crate) fn create_key_event(
 
     let key_from_code = code_to_key(physical_key, scancode);
     let (logical_key, key_without_modifiers) = if matches!(key_from_code, Key::Unidentified(_)) {
+        // `key_without_modifiers` ignores all modifiers except for shift.
```
kchibisov, line 150: it should ignore all including shift, and I think it does that.

kchibisov (changes requested):

src/platform_impl/macos/event.rs
emilk: I want to document somewhere the different gotchas that one should test before modifying the function again. Do you have a better place to suggest for that?

src/platform_impl/macos/event.rs
emilk: True, my bad!

src/platform_impl/macos/event.rs
emilk: No, this `get_logical_key_char` and `charactersIgnoringModifiers` only heeds the shift key. See https://developer.apple.com/documentation/appkit/nsevent/1524605-charactersignoringmodifiers

I also verified this with `dbg!`

emilk:
Comments addressed, and there is a changelog entry

src/platform_impl/macos/event.rs
emilk: I shortened it down

src/platform_impl/macos/event.rs
```diff
@@ -113,6 +115,14 @@ pub(crate) fn create_key_event(
...
     let mut physical_key = key_override.unwrap_or_else(|| scancode_to_physicalkey(scancode as u32));
 
+    // NOTE: The logical key should heed both SHIFT and ALT if possible.
+    // For instance:
+    // * Pressing the A key: logical key should be "a"
+    // * Pressing SHIFT A: logical key should be "A"
+    // * Pressing CTRL SHIFT A: logical key should also be "A"
+    // This is not easy to tease out of `NSEvent`, but we do our best.
+
+    // `text_with_all_modifiers` is unreliable when ctrl or cmd is pressed
```
kchibisov, line 125: This looks out of place here, it matters when you do `!has_crtl && !has_cmd` and it already has similar comment.

kchibisov (approved):

emilk:
Thanks for the review!

emilk:
I think this would be a good candidate to include in the 0.29.7 patch release

kchibisov:
I already released it in 0.29.9

kchibisov:
It seems like it regressed this https://github.com/alacritty/alacritty/issues/7742

where command is mandatory to produce the correct binding.

jabwd:
This might be a useless continuation on this post, but this change fixed it for me:
https://github.com/alacritty/alacritty/compare/master...jabwd:alacritty:fix/dvorak-qwerty

While trying to fix it in the fork (because this was the literal only reason I stopped using alacritty) I noticed Zed's implementation for this fix is much more complicated, but they don't use winit so I'm not sure if its relevant for alacritty.

Its possible this change should be made within winit, to explicitly support dvorak+qwerty but I feel like it'd pollute their api since it has ways of getting the correct info.

PR #3366, opened 2024-01-05
X11: cache custom cursors

daxpedda:
Addresses #3306.

Wayland isn't really affected, just needed some slight refactoring to account for different backends.

### DIFF
--- a/src/platform_impl/linux/wayland/window/state.rs
+++ b/src/platform_impl/linux/wayland/window/state.rs
@@ -28,7 +28,7 @@ use sctk::shm::Shm;
 use sctk::subcompositor::SubcompositorState;
 use wayland_protocols_plasma::blur::client::org_kde_kwin_blur::OrgKdeKwinBlur;
 
-use crate::cursor::CursorImage;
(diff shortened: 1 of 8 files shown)


src/platform_impl/linux/wayland/window/state.rs
```diff
@@ -726,10 +726,22 @@ impl WindowState {
...
-    pub(crate) fn set_custom_cursor(&mut self, cursor: &CursorImage) {
+    pub(crate) fn set_custom_cursor(&mut self, cursor: RootCustomCursor) {
+        let cursor = match cursor {
+            RootCustomCursor {
+                inner: PlatformCustomCursor::Wayland(cursor),
+            } => cursor.0,
+            RootCustomCursor {
+                inner: PlatformCustomCursor::X11(_),
+            } => {
+                error!("passed a X11 cursor to Wayland backend");
```
daxpedda, line 737: I could change those to `unreachable!`, because you can't recreate event loops on those platforms, right?

notgull (approved):
Looks good to me

PR #3371, opened 2024-01-06
Web: support Firefox `privacy.resistFingerprinting`

### LINKED ISSUE #3345: media query doesn't match, Firefox 122.0b3
WASM,  Firefox Developer Edition 122.0b3 (64-bit)

````
panicked at /home/mike/.cargo/registry/src/index.crates.io-6f17d22bba15001f/winit-0.29.7/src/platform_impl/web/web_sys/resize_scaling.rs:129:9:
created media query doesn't match, 1 != 1
````

Clearly these are floats very close to 1. I don't know why they don't match.  Same code works fine on Firefox 121.0.  I don't know which version of the browser this started with because I'm coming in fresh with a new project.

If it's a bug in firefox, just let me know.

### PR DESCRIPTION
**Pull request by daxpedda**
This only changes an `assert!` to a `debug_assert!` avoiding a Firefox bug.
No functional changes.

Fixes #3345.

### DIFF
--- a/src/platform_impl/web/web_sys/resize_scaling.rs
+++ b/src/platform_impl/web/web_sys/resize_scaling.rs
@@ -127,7 +127,7 @@ impl ResizeScaleInternal {
              (-webkit-device-pixel-ratio: {current_scale})",
         );
         let mql = MediaQueryListHandle::new(window, &media_query, closure);
-        assert!(
+        debug_assert!(
             mql.mql().matches(),
             "created media query doesn't match, {current_scale} != {}",
             super::scale_factor(window)

### CODE REVIEW

File: src/platform_impl/web/web_sys/resize_scaling.rs [hunk 1]
```diff
@@ -127,7 +127,7 @@ impl ResizeScaleInternal {
              (-webkit-device-pixel-ratio: {current_scale})",
         );
         let mql = MediaQueryListHandle::new(window, &media_query, closure);
-        assert!(
+        debug_assert!(
```
notgull, line 130: Doesn't this mean that the bug is still here on Debug mode?

daxpedda: Yes, I should probably remove the changelog entry now that I think about it.

For context: it's not really a problem, because `privacy.resistFingerprinting` isn't stable in Firefox and has to be enabled through `about:config`. This is in addition to being a Firefox bug, Winit does everything right here.

PR #3378, opened 2024-01-09
iOS: fix `request_redraw` from `RedrawRequested`

daxpedda:
This fixes calling `Window::request_redraw()` from `WindowEvent:RedrawRequested` causing panics.
Instead it is queued for the next event loop, which is awoken instantly as currently iOS doesn't support #2412.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ Unreleased` header.
 - **Breaking:** Rename `VideoMode` to `VideoModeHandle` to represent that it doesn't hold static data.
 - **Breaking:** No longer export `platform::x11::XNotSupported`.
 - **Breaking:** Renamed `platform::x11::XWindowType` to `platform::x11::WindowType`.
+- On iOS, fix calling `Window::request_redraw()` from `WindowEvent:RedrawRequested`.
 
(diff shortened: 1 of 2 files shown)


madsmtm (approved):
Hmm, I tried to reproduce the issue, and I don't seem to be able to?

But in any case, this does look more correct.

daxpedda:
Does it actually cause a `RedrawRequested` event? Because a user on Matrix shared some screenshots where it didn't show up.
(I will also still have to remove the warning)

madsmtm:
Hmm, nope, it doesn't (tested with WGPU, it probably does if you use other rendering backends).

mockersf:
With Bevy (so wgpu too), this doesn't fix calling `request_redraw` from a `RedrawRequested` event. It doesn't panic, it just does nothing

daxpedda:
> With Bevy (so wgpu too), this doesn't fix calling `request_redraw` from a `RedrawRequested` event. It doesn't panic, it just does nothing

Thank you for testing it!
I guess I'm gonna go ahead and close this as I can't really fix/develop it as I can't actually test and see what I'm doing.

I left an issue at #3406.

PR #3385, opened 2024-01-12
Windows: Set fullscreen or maximized after CreateWindowEx

kazatsuyu:
Resizing a window in WM_CREATE would update the size of the normal state, even if it was maximized or in full screen.  
Therefore, I changed it so that full-screening and maximizing are done after CreateWindowEx.


kchibisov:
The whole purpose is to have properly set initial size, so once you un-maximize/fullscreen you fallback to the size you've passed in the `WindowBuilder`?

kazatsuyu:
Yes, without this modification, windows created with `with_maximized(true)` will be displayed on the entire screen even when unmaximized.  
If `with_maximized` and `with_position` are used together, the behavior is even stranger: the window will return to its normal size after being maximized once. This is due to `set_outer_position` being called after maximization, anyway this is fixed by this PR.

src/platform_impl/windows/window.rs
```diff
@@ -1398,7 +1387,7 @@ unsafe fn init(
     let menu = attributes.platform_specific.menu;
     let mut initdata = InitData {
         event_loop,
-        attributes,
+        attributes: attributes.clone(),
```
kchibisov, line 1390: Hm, could you stash the fullscreen/maximized and then use them instead of cloning the attributes state?

kchibisov (changes requested):

src/platform_impl/windows/window.rs
kazatsuyu: I fixed this.

kchibisov (approved):

Issue #3367, opened 2024-01-06
Winit API redesign


kchibisov:
Quick links: [Discussion summary](https://github.com/rust-windowing/winit/issues/3367#issuecomment-1889741433) and [current implementation plan](https://github.com/rust-windowing/winit/issues/3367#issuecomment-1883667741).

### We need feedback for events and immediate reactions

The current API doesn't really work when you must get some values from the users
in reply to some events, because users has a clear choice to not do so. In some
cases it could lead to undesired behavior and even crashes. An example of such
thing could be:

  - the dead key API (it goes via globals now).
  - https://github.com/rust-windowing/winit/issues/2921
  - https://github.com/rust-windowing/winit/issues/3349
  - https://github.com/rust-windowing/winit/issues/2304 (require saying that
  you've handled, otherwise you'll block everything
  - https://github.com/rust-windowing/winit/issues/3192 (you need to ask during
  window creation)
  - Clipboard APIs (we need to pick mime-type), the same with DND.
  - Live resizes (when you can adjust the size during resize).

The getting feedback part could be solved vi abuse of `Mutex` but it's not like
you can force users into doing correct cross-platform code, which actually
behaves the same.

There are more issues due to current massive callback approach, I just picked
few of them.

### Monolithic design 

In the current form winit is monolithic meaning that if the users want to use
winit they must include the entire crate introducing a lot of dependencies
and increase in compile time.

This is not great since GUIs don't really need all the winit or all the
backends. We also have a situation where you use winit or die, which is also
not great, so having a crate which provides mostly types could benefit.

Other issue of monolithic design is that winit is growing on its backends and
there's a demand to support niche platforms. Unfortunately, we can't provide
good support for all of them with the current maintainers we have. And some
platforms we can't event test reliably (e.g. redox). To solve this
winit should allow writing out-of-tree backends.

### Interior mutability mess

Winit historically marks the `Window` as `Send + Sync`, while it's actually good
to write multi threaded code, internally it all goes back to event loop thread,
this is true for Windows, Wayland, Web, iOS.

If you even tried contributing into early winit it was event worse on interior
mutabilty and the use of `Mutex`, which were not need in `99%` of the time.

### Objects actually have lifetime

`Window` and other objects created while event loop is running actually has
lifetime and generally can't be used when the event loop is paused.

This issue is pretty clear with the `run_on_demand` sort of APIs where we must
destroy everything between the runs.

[...]

PR #3386, opened 2024-01-12
Begin transition to a trait-based system (`ApplicationHandler`, `WindowHandler`)

madsmtm:
Part of https://github.com/rust-windowing/winit/issues/3367. Blocked on https://github.com/rust-windowing/winit/pull/3299, since I want to keep the active handle being passed around as simple as possible.

Add a trait `ApplicationHandler<T>` which replaces the closure `impl FnMut(Event<T>, ActiveEventLoop<'_>)` matching on `Event<T>`s. This trait is fully compatible with the closure, and all backends and most examples still use the old event design. But it should be possible to slowly migrate things from here.

Note that _this will not be the final design of the trait_, `exit` will probably be replaced by `Drop` (you shouldn't need to access the `ActiveEventLoop<'_>` from there anyhow, and it's more "Rusty"), we'll probably introduce a closure at the beginning instead of `init`, there are events that should be changed to have return types, and so on.
But we need to keep these APIs around mostly as-is, so that we can convert between the trait and the closure, at least until we've gathered experience, and are ready to remove the closure-based API.

TODO:
- [ ] Gather feedback.
- [ ] Finish implementation (I should probably add a `WindowHandler` in this PR too).
- [ ] Prepare at least one backend to migrate to this, to see that it is possible.
- [ ] Add changelog entry and docs.
- [ ] Update all examples (can probably be done in another PR).

### DIFF
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -48,9 +48,8 @@ pub struct EventLoop<T: 'static> {
 /// your callback. [`EventLoop`] will coerce into this type (`impl<T> Deref for
 /// EventLoop<T>`), so functions that take this as a parameter can also take
 /// `&EventLoop`.
-pub struct EventLoopWindowTarget<T: 'static> {
-    pub(crate) p: platform_impl::EventLoopWindowTarget<T>,
-    pub(crate) _marker: PhantomData<*mut T>, // Not Send nor Sync + invariant over T
+pub struct EventLoopWindowTarget {
+    pub(crate) p: platform_impl::EventLoopWindowTarget,
 }
 
 /// Object that allows building the event loop.
@@ -142,7 +141,7 @@ impl<T> fmt::Debug for EventLoop<T> {
     }
 }
 
-impl<T> fmt::Debug for EventLoopWindowTarget<T> {
+impl fmt::Debug for EventLoopWindowTarget {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.pad("EventLoopWindowTarget { .. }")
     }
@@ -244,7 +243,7 @@ impl<T> EventLoop<T> {
     #[cfg(not(all(wasm_platform, target_feature = "exception-handling")))]
     pub fn run<F>(self, event_handler: F) -> Result<(), EventLoopError>
     where
-        F: FnMut(Event<T>, &EventLoopWindowTarget<T>),
+        F: FnMut(Event<T>, &EventLoopWindowTarget),
     {
         self.event_loop.run(event_handler)
     }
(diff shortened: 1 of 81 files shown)


kchibisov (changes requested):
I don't see why I should merge dead from the start code...

madsmtm:
Concretely, because we can ship this to users in `v0.30.0`, and allow them to upgrade to the trait-based approach with the old event-based approach being deprecated, and allow them to give feedback if there's something that didn't work for them.

And _then_ we can work on removing `Event<T>` and associated helpers.

(But I get that [we differ in views](https://github.com/rust-windowing/winit/issues/3367#issuecomment-1884072060) on this).

kchibisov:
I don't work for users to have more work upgrading to incomplete API and then to complete one, I _as downstream project maintainer_ won't waste my time on such mess.

madsmtm:
I don't see your issue as a downstream user? `ApplicationHandler` matches the `Application` trait which you've outlined in [`winit-next`'s `application.rs`](https://github.com/rust-windowing/winit-next/blob/8bcf7d91a6f44853a4795444a661959704704751/winit-core/src/application.rs), which is the primary API change that users are going to have to deal with?

Regarding `ApplicationWindow`, I will add that to this PR as well, though the only thing that would really be different in this initial phase is that `close_requested` won't return anything yet (and that should be a minor thing to upgrade to).

kchibisov:
The thing about my API is that it's not complete because I was asked to wait, more things will change, like surely you can port some of that work, but I can't really deal with return values on Wayland because the don't make any sense to me since everything is queued, so the main motivation to return values is completely lost.

And the whole return is blocked for me because `Window` is `Send` + `Sync`, so I have API users can't use and I can't use, if I can't use it and it solves nothing why should I bother to upgrade Wayland backend and alacritty to it.

kchibisov:
The same issue with neovide since I can't solve their sync issues unless I just do the full design and `&dyn` stuff.

src/event_loop.rs
```diff
@@ -305,12 +308,17 @@ impl<T> EventLoop<T> {
     /// [^1]: `EventLoopExtWebSys::spawn()` is only available on WASM.
     #[inline]
     #[cfg(not(all(wasm_platform, target_feature = "exception-handling")))]
-    pub fn run<F>(self, mut event_handler: F) -> Result<(), EventLoopError>
+    pub fn run<F>(self, handler: F) -> Result<(), EventLoopError>
```
daxpedda, line 311: I believe we could do the following:
```rust
impl<T: FnMut(Event<T>, ActiveEventLoop<'_>)> ApplicationHandler for T { ... }
```
Then we don't need to introduce an additional `EventLoop::run_with()` function.

madsmtm: I think I tried that, but it wasn't backwards compatible since type inference no longer kicks in, you'd have to specify the type of the closure arguments.

E.g. `event_loop.run(|event, active| { ... })` would have to be `event_loop.run(|event: Event, active: ActiveEventLoop<'_>| { ... })`.

kchibisov:
Superseded by #3517.

PR #3387, opened 2024-01-12
Example of migrating backends to `ApplicationHandler`

madsmtm:
See https://github.com/rust-windowing/winit/pull/3386.

I stubled into a roadblock for this, as there are several backends which are currently queuing events. This means they either need some sort of event structure themselves if we're to remove `Event<T>`, or will have to be rewritten to not queue events (preferred).

Backends that are currently queuing events: macOS, iOS, web, Windows, Wayland.

I can only speak for macOS and iOS, but I'm pretty sure those already prevent re-entrancy for most events, so migrating to handling events immediately will be the strictly better solution.

In any case, this PR showcases the migration of the Redox backend, so it should be technologically possible.

### DIFF
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -20,7 +20,7 @@ use web_time::{Duration, Instant};
 
 use crate::error::EventLoopError;
 use crate::event::Event;
-use crate::event_helper::{map_event, MapEventHelper};
+use crate::event_helper::MapEventHelper;
 use crate::handler::ApplicationHandler;
 use crate::{monitor::MonitorHandle, platform_impl};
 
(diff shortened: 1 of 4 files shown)


kchibisov:
yeah, I can't write like that on Wayland at all until I change other API. The issue is that `Window` creation must be made async before I can do so, because I do dispatch events and I can get events during creation of the window for other windows.

madsmtm:
Right, I missed the `EventSink` when I reviewed which backends are queuing events.

madsmtm:
If it's not really feasible to rewrite the Wayland backend to utilize this trait until we do other work, then... Don't! Use the `crate::event_helper::map_event` approach like I've done for macOS currently.

I'm also only talking about removing the `Event<T>` from public API; it could stay in private APIs for backends that need it to queue certain events, until we slowly get rid of unnecessary queuing.

kchibisov:
@madsmtm I may dispatch events from the window creation, like I'd need to change entire window creation process to be `async` like I have on my branch.

PR #3397, opened 2024-01-15
Update platform and core maintainers

kchibisov:


kchibisov:
cc @msiglreith @daxpedda @notgull

.github/CODEOWNERS
```diff
@@ -1,8 +1,7 @@
 # Core maintainers:
```
madsmtm, line 1: This section was meant as a "this person has administrative access to the `rust-windowing` repo", so maybe we should just rename it to that (the concept of "core maintainers" doesn't really make sense).

kchibisov: Hm, that's fair, they have access to all the repos though...

kchibisov: Maybe it's fine to do that that way, since core kind of makes sense, like owner access to all repos just happened to match with these people since they maintain more than project usually.

notgull (approved):
I'm fine with being on the call list for X11

.github/CODEOWNERS
madsmtm: Oh, didn't see that you've change the `rust-windowing` org structure to match this.

```suggestion
# Core maintainers (ping these for administrative tasks in `rust-windowing`):
```

kchibisov: I'm not sure that we should have it here? It's not that discoverable. In general people should ping us on matrix. I've used core maintainers for the people who do public API now.

madsmtm: Fair, perhaps we should just remove that section altogether.

kchibisov: I think we should, and then when the backends will split we'll just add core stuff to get notified for winit-core and specific backends will be separate.

msiglreith:
fine with me, I unfortunately don't have much time due to work

.github/CODEOWNERS
daxpedda, line 1: Web still says "no maintainer", could we remove that while we are at it?

daxpedda, line 1: Would also like to propose @MarijnS95 for Android?
I'm not sure what the history is here.

daxpedda (approved):

.github/CODEOWNERS
kchibisov: They should explicitly opt-in, the current maintainer is kind of accurate.

MarijnS95: Does opting in imply that I can finally get notifications for changes to the Android backend - maybe even for issues tagged as such?

daxpedda: You would get auto-assigned, I think that shows up in notifications?
But in all seriousness: it would imply some level of responsibility of course :smiley:.

kchibisov: Issues not, but you'll be pinged for each change to the sources in android backend.

MarijnS95: Won't be a great start if I'm gone for all of February then. I'll probably be too busy working on other (boilerplate) crates, maybe @rib wants to pick this up while also being the `android-activity` developer/maintainer?

Doesn't mean that I don't still dream of the various window/activity improvements we discussed last week.

MarijnS95: Feel free to sign me up then. I'd like to stay in the loop more of what is happening in the Android backend.

PR #3400, opened 2024-01-15
Remove `WindowBuilder` in favour of `WindowAttributes`

madsmtm:
Blocked on https://github.com/rust-windowing/winit/pull/3318.

Rough draft, the plan is to move towards:
```rust
let attrs = WindowAttributes::new().with_title("MyTitle");
let window = event_loop.create_window(attrs).unwrap();
```

I think in this case it makes sense to remove the builder, because it really _is_ just attributes; but in a case like `EventLoopBuilder`, I think it still makes sense to keep that as a builder.

### DIFF
--- a/examples/custom_cursors.rs
+++ b/examples/custom_cursors.rs
@@ -36,7 +36,7 @@ fn main() -> Result<(), impl std::error::Error> {
     let builder = WindowBuilder::new().with_title("A fantastic window!");
     #[cfg(wasm_platform)]
     let builder = {
-        use winit::platform::web::WindowBuilderExtWebSys;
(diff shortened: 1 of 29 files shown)


kchibisov:
I think I have it ported on winit-next the way you suggested, however there's an issue with extensibility of it in the future. Like with dynamic backend we can't really account for everything ahead of time, so I was thinking that maybe we should create a `trait` with defaults?

But in general having the same type is fine by me. it's called a builder because of the pattern with the `build`, but really it's just a bunch of attributes and a method to set them.

daxpedda:
Closing in favor https://github.com/rust-windowing/winit/pull/3447.

PR #3401, opened 2024-01-15
First attempt at using UIRotationGestureRecognizer

jpedrick:
This is a first attempt at adding a UIKit GestureRecognizer to the ios platform. It is absolutely untested except for compiling and not causing segfaults in the examples. Posting for @madsmtm to provide early feedback.


madsmtm:
Like we discussed on Matrix, this is already being worked on in https://github.com/rust-windowing/winit/pull/3130.

PR #3408, opened 2024-01-18
On Windows, apply `ScaleFactorChanged` new size if different than OS

amrbashir:
This fixes an issue when setting the position of the window on a new monitor and immediately maximizing it

```rs
window.set_outer_position::<PhysicalPosition<u32>>((2000, 200).into());
window.set_maximized(true);
```

Due to the nature of the event loop, the requested position and maximization state will apply correctly but due to the fact that the new position is a different monitor, a `ScaleFactorChanged` is emitted afterwards to the evenloop and a new size is set while the window is still maximized which results in a window that has `WS_MAXIMZE` window style but doesn't cover the whole monitor.

Screenshot:
image


notgull (approved):
Thanks!

Poke @Stefyn, could I convince you to take a look at this?

msiglreith:
small note: as I have a 2nd monitor now I planned to review it at the weekend

msiglreith (changes requested):
Thanks, but I don't like the additional state tracking and logic for handling a specific edge case. Other cases like move + `request_inner_size` will also see similar inconsistent behavior due to retroactively applying outdated size changes.

I would prefer to only apply changes from a buffer `ScaleFactorChanged` if the user explicitly updated the size. Therefore just call `set_size` when `new size != old size` as the original size change has already been applied.

amrbashir:
Hopefully this is what you meant.

CHANGELOG.md
```diff
@@ -47,6 +47,7 @@ Unreleased` header.
 - On X11, fix deadlock when adjusting DPI and resizing at the same time.
 - On Wayland, fix `Focused(false)` being send when other seats still have window focused.
 - On Wayland, fix `Window::set_{min,max}_inner_size` not always applied.
+- On Windows, apply `ScaleFactorChanged` changes only if user new size is different than what the OS reporeted, which fixes and issue when moving a window to another monitor and immediately maximizing it, previously resulted in a maximized window but is not expanded to the full working area of the monitor.
```
msiglreith, line 50:
```suggestion
- On Windows, fix inconsistent resizing behavior with multi-monitor setups when repositioning outside the event loop.
```

kchibisov:
Next time could you _please_ properly format commits before merging?

PR #3410, opened 2024-01-19
On Windows, Remove `WS_CAPTION`, `WS_BORDER` and `WS_EX_WINDOWEDGE` styles for child windows

### LINKED ISSUE #2754: Child window without decorations not working on Win 10
Setting decorations to false doesn't work on child windows on Win 10, the decorations are still visible.

This might be related to other decoration issues that appeared after https://github.com/rust-windowing/winit/pull/2419 

I don't have much experience with the winit code base, but as far as I can see WS_CAPTION and WS_BORDER are always added to the window style and never removed (except in adjust_rect(...) that is not initially called).

Maybe they can only be added if decorations are activated in the first place?

Thanks

amrbashir:
closes #2754


msiglreith (changes requested):
Thanks but this should only affect undecorated windows

amrbashir:
Should be good to go now

msiglreith (approved):

ArthurKlausHoff:
Why WS_CAPTION, WS_BORDER and WS_EX_WINDOWEDGE are only removed from child windows?

This won't change anything if a non-child window doesn't have decorations, they're still broken.

PR #3412, opened 2024-01-20
On macOS, fix incorrect IME cursor rect origin

uiryuu:
`window.set_ime_cursor_area` requires a position from the top left corner according to the documentation. However, the NSRect's origin is from bottom left. The y coordinate should be deducted by height of the `ime_size` to become the bottom-left corner. The current implementation will cause the IME window to overlap with the text (if the client implements correctly according to the winit documentation).

I see refactors in the master branch. I'm not sure whether or not this issue still exists on the master, so I only tried to fix this on the `v0.29.x` branch.

I'm very new to winit, so if I get something horribly wrong, please point it out.

Don't know whether I should add something in the changelog for this; if it's needed please tell me.

- [ ] Tested on all platforms changed (only effects macOS, so didn't test for other platforms)

### DIFF
--- a/src/platform_impl/macos/view.rs
+++ b/src/platform_impl/macos/view.rs
@@ -410,9 +410,9 @@ declare_class!(
             let content_rect = window.contentRectForFrameRect(window.frame());
             let base_x = content_rect.origin.x as f64;
             let base_y = (content_rect.origin.y + content_rect.size.height) as f64;
-            let x = base_x + self.state.ime_position.get().x;
-            let y = base_y - self.state.ime_position.get().y;
             let LogicalSize { width, height } = self.state.ime_size.get();
+            let x = base_x + self.state.ime_position.get().x;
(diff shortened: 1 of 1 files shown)


kchibisov:
Always open patches against the master branch.

uiryuu:
@kchibisov Are you suggesting me make a PR and try to fix it for the master branch?

kchibisov:
Yes, you should fix it for master, and then it's my job to backport the fix into release branch.

uiryuu:
I just confirmed that the issue has been fixed in the master branch. The refactor has fixed the issue.

madsmtm:
> I just confirmed that the issue has been fixed in the master branch. The refactor has fixed the issue.

Yeah, this was done in #3302.

uiryuu:
So how should we deal with this PR then

kchibisov:
I'll back port it manually when we'll be doing 0.29.11

Issue #3419, opened 2024-01-24
Android: Instructions pointed to from README.md are outdated


hrydgard:
The readme points to android-activity [example applications](https://github.com/rib/android-activity/tree/main/examples) , which are for some older version.

madsmtm:
Still an issue: https://github.com/rust-windowing/winit/pull/3420#issuecomment-1908907735

CC @rib, I don't know what's correct to do on Android.

hrydgard:
Well, I did figure it out (use EventLoopBuilder, pull in platform::android::EventLoopBuilderExtAndroid) by mashing together bits and pieces from here and there. But I was also surprised that the current example framework in the wgpu repo doesn't do this, so won't work on Android as far as I can tell.

https://github.com/hrydgard/wgpu-winit-android-example/blob/main/lib/src/lib.rs#L18

kchibisov:
glutin example is always kept up to date and works and yes wgpu doesn't have android support in examples iirc.

hrydgard:
Ah, look at that:

https://github.com/rust-windowing/glutin/blob/master/glutin_examples/examples/android.rs#L8

No idea how I would know to go look there from the winit repo though - googling for `winit 0.29 android init` certainly didn't :)

kchibisov:
It's actually weird that we don't have an example in winit that tries to run something actually on android, even though it'll fail to show any window because softbuffer doesn't work with android yet.

hrydgard:
Well, it would still show a window, just no contents.  

(Off-topic, but I think on Android softbuffer should just be a small snippet of GL code, for max compatibility, despite that going against the softbuffer goal of not requiring a GPU. Android will fall back to a software GL implementation if you manage to get it running on a device without one. I guess the other possibility is some JNI talking to the Android java bitmap API, but the code certainly wouldn't be prettier.)

PR #3420, opened 2024-01-24
Fix Android examples link in README

madsmtm:
Fixes https://github.com/rust-windowing/winit/issues/3419


hrydgard:
Thanks, there's actually some more brokenness in the readme around there. For example, 

> Add an android_main entrypoint (as above),

> Pass a clone of the AndroidApp that your application receives to Winit when building your event loop (as shown above).

But it isn't shown above in either case.

PR #3421, opened 2024-01-24
Ios gesture detection

jpedrick:


src/platform_impl/ios/app_state.rs
```diff
@@ -612,9 +612,9 @@ pub(crate) fn handle_nonuser_events<I: IntoIterator<Item = EventWrapper>>(
         match wrapper {
             EventWrapper::StaticEvent(event) => {
                 if !processing_redraws && event.is_redraw() {
-                    log::info!("processing `RedrawRequested` during the main event loop");
```
jpedrick, line 615: Made this a `trace` just for debugging. For some reason this is very noisy with bevy.

src/platform_impl/ios/uikit/press.rs
```diff
@@ -0,0 +1,95 @@
...
+
+        #[method(key)]
+        pub fn key(&self) -> Option<&UIKey>;
+
+        #[method(type)]
+        pub fn type_(&self) -> UIPressType;
+
+        #[cfg(feature = "i_dont_know_how_to_make_this_work")]
+        #[method_id(gestureRecognizers)]
+        pub fn gesture_recognizers(&self) -> Id<NSArray<&UIGestureRecognizer>>;
```
madsmtm, line 36: Like this ;)
```suggestion
        pub fn gesture_recognizers(&self) -> Id<NSArray<UIGestureRecognizer>>;
```

madsmtm: Also, the docs says this is an optional:
```suggestion
        pub fn gesture_recognizers(&self) -> Option<Id<NSArray<UIGestureRecognizer>>>;
```

src/platform_impl/ios/uikit/press.rs
```diff
@@ -0,0 +1,95 @@
...
+    unsafe impl ClassType for UIPress {
+        type Super = NSObject;
+        type Mutability = mutability::InteriorMutable;
+    }
+);
+
+extern_methods!(
+    unsafe impl UIPress {
+        #[method(locationInView:)]
+        pub fn locationInView(&self, view: Option<&UIView>) -> CGPoint;
```
madsmtm, line 26: I don't think this method exists on `UIPress`.

src/platform_impl/ios/uikit/press.rs
```diff
@@ -0,0 +1,95 @@
...
+
+        #[method(key)]
+        pub fn key(&self) -> Option<&UIKey>;
```
madsmtm, line 29: Methods that return classes should use `Id` and `method_id` (otherwise the pointer may be destroyed while we're unaware).

```suggestion
        #[method_id(key)]
        pub fn key(&self) -> Option<Id<UIKey>>;
```

Same goes for `responder` and `window` methods below.

src/platform_impl/ios/uikit/press.rs
```diff
@@ -0,0 +1,95 @@
...
+
+    TypePageUp = 30,
+    TypePageDown = 31,
+}
+
+#[derive(Debug, PartialEq, Eq)]
+#[allow(dead_code)]
+#[repr(isize)]
+pub enum UIPressPhase {
+    PhaseBegan,      // whenever a button press begins.
```
madsmtm, line 82: Nit: Use ordinary doc comments:
```suggestion
    /// whenever a button press begins.
    PhaseBegan,
```

src/platform_impl/ios/uikit/press.rs
```diff
@@ -0,0 +1,95 @@
...
+
+        #[method(timestamp)]
+        pub fn timestamp(&self) -> NSTimeInterval;
+    }
+);
+
+#[derive(Debug, PartialEq, Eq)]
+#[allow(dead_code)]
+#[repr(isize)]
+pub enum UIPressType {
```
madsmtm, line 64: Nit: Try not to use `enum`s, but rather a struct with `const`s - see `UIScreenOverscanCompensation` for an example.

The reason is that the `enum` approach is unsound, and may break in future versions of iOS.

madsmtm (changes requested):
Did a quick review of `uikit/view.rs`, perhaps try to use the suggestions from there.

I'd also recommend looking at the source code of `icrate`, it is usually a fairly good source for how `extern_methods!` should be implemented.

Also, you should probably rebase this onto `master`.

PR #3424, opened 2024-01-25
Add `EventLoopProxy::waker(self) -> Waker`

madsmtm:
As part of https://github.com/rust-windowing/winit/issues/3367, we're considering removing the generic on `Event<T>`, `EventLoop<T>`, `EventLoopProxy<T>` and the callback. By extension, we'd be getting rid of `Event::UserEvent(T)`.

One approach to doing so would be to allow converting `EventLoopProxy` to the standard library's [`Waker`](https://doc.rust-lang.org/std/task/struct.Waker.html) type, which allows the user to wake a task (in this case the event loop) from any thread; then they can implement what they need themselves on top of that, using the standard library's [`mpsc::channel`](https://doc.rust-lang.org/std/sync/mpsc/fn.channel.html), [`mpsc::sync_channel`](https://doc.rust-lang.org/std/sync/mpsc/fn.sync_channel.html) or likewise, depending on their needs.

This is likely not a perfect solution for wakeups, we still need to properly figure out how to do timers; but this can get us some of the way there.

Blocked on https://github.com/rust-windowing/winit/pull/3449, we need to ensure that `Sync` is actually sound.

CC @notgull, I think this may be slightly useful for `async`?

Questions:
- [ ] Do we need a `UserWoken` event, or is `NewEvents` is enough?
- [x] Should we hide the `Waker` behind a custom type like `EventLoopWaker`, in case we want to expose further functionality in the future?
  - We keep `EventLoopProxy`, and allow that to be converted to `Waker`.
- [ ] Is `Waker` too heavy-weight for this? Can we just use a function `wake(&self)` on `EventLoopProxy`? Or maybe both?
- [ ] Does the user need the ability to handle the case where the event loop has been destroyed? Or is that handled fine enough through other means, like in `mpsc`?

TODO:
- [ ] Finish implementation on all platforms.
- [ ] Properly test `Waker` implementations with updated example.
- [ ] Add an entry to `CHANGELOG.md`.
- [ ] Update documentation.

### DIFF
--- a/winit-common/src/core_foundation/event_loop_proxy.rs
+++ b/winit-common/src/core_foundation/event_loop_proxy.rs
@@ -1,9 +1,12 @@
 use std::os::raw::c_void;
+use std::ptr::NonNull;
 use std::sync::Arc;
+use std::task::{RawWaker, RawWakerVTable, Waker};
 
 use objc2::MainThreadMarker;
 use objc2_core_foundation::{
     kCFRunLoopCommonModes, CFIndex, CFRetained, CFRunLoop, CFRunLoopSource, CFRunLoopSourceContext,
+    Type,
 };
 use winit_core::event_loop::EventLoopProxyProvider;
 
@@ -118,5 +121,55 @@ impl EventLoopProxyProvider for EventLoopProxy {
         // This is required since we may be (probably are) running on a different thread, and the
         // main loop may be sleeping (and `CFRunLoopSourceSignal` won't wake it).
         self.main_loop.wake_up();
+    }
+
+    fn waker(&self) -> Waker {
+        const VTABLE: RawWakerVTable =
+            RawWakerVTable::new(clone_waker, wake, wake_by_ref, drop_waker);
+
+        unsafe fn clone_waker(data: *const ()) -> RawWaker {
+            // SAFETY: The poiner came from `CFRunLoopSource` and is valid and non-null.
+            let source = unsafe { &*data.cast::<CFRunLoopSource>() };
+
+            // Increment reference count.
+            let source = source.retain();
+
+            // Pass ownership to the raw waker.
+            let data: *const CFRunLoopSource = CFRetained::into_raw(source).as_ptr();
+            RawWaker::new(data.cast(), &VTABLE)
+        }
+
+        unsafe fn wake(data: *const ()) {
+            unsafe { wake_by_ref(data) };
+            unsafe { drop_waker(data) };
+        }
+
+        unsafe fn wake_by_ref(data: *const ()) {
+            // SAFETY: The poiner came from `CFRunLoopSource` and is valid and non-null.
+            let source = unsafe { &*data.cast::<CFRunLoopSource>() };
+
+            // Signal the source, which ends up later invoking `perform` on the main thread.
+            //
+            // Multiple signals in quick succession are automatically coalesced into a single
+            // signal.
+            source.signal();
+
+            let main_loop = CFRunLoop::main().unwrap();
+            // Let the main thread know there's a new event.
+            //
+            // This is required since we may be (probably are) running on a different thread, and
+            // the main loop may be sleeping (and `CFRunLoopSourceSignal` won't wake it).
+            main_loop.wake_up();
+        }
+
+        unsafe fn drop_waker(data: *const ()) {
+            let source = data.cast::<CFRunLoopSource>().cast_mut();
+            // SAFETY: The poiner came from `CFRunLoopSource` and is valid and non-null.
+            // We take ownership of a retain count here.
+            let _source = unsafe { CFRetained::from_raw(NonNull::new_unchecked(source)) };
+        }
+
+        let data: *const CFRunLoopSource = CFRetained::into_raw(self.source.clone()).as_ptr();
+        unsafe { Waker::from_raw(RawWaker::new(data.cast(), &VTABLE)) }
     }
 }
--- a/winit-core/src/event_loop/mod.rs
+++ b/winit-core/src/event_loop/mod.rs
@@ -4,6 +4,7 @@ pub mod run_on_demand;
 use std::fmt::{self, Debug};
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::Arc;
+use std::task::Waker;
 use std::time::Duration;
 
 use rwh_06::{DisplayHandle, HandleError, HasDisplayHandle};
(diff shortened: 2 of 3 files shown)


kchibisov:
>One approach to doing so would be to use the standard library's [Waker](https://doc.rust-lang.org/std/task/struct.Waker.html) type instead, which allows the user to wake a task (in this case the event loop) from any thread; then they can implement something similar to EventLoopProxy themselves on top of that, using the standard library's [mpsc::channel](https://doc.rust-lang.org/std/sync/mpsc/fn.channel.html), [mpsc::sync_channel](https://doc.rust-lang.org/std/sync/mpsc/fn.sync_channel.html) or likewise, depending on their needs.

Can we use a separate type which may return a `Waker`? We also need an event indicating the wakeup, so users don't have to poll it. I already have something like that in the `winit-next`, though, it's done not via `Waker`.

Although, maybe our own type will be better in general, unless `Waker` has some better integration with the async stuff? Like having timers on it could probably make sense, and we can't really extend the timer stuff on the waker itself, but we can on our own type.

src/platform_impl/windows/event_loop.rs
```diff
@@ -753,29 +756,56 @@ impl EventLoopThreadExecutor {
...
         Self {
-            target_window: self.target_window,
+            hwnd: self.hwnd,
             event_send: self.event_send.clone(),
         }
     }
 }
 
+fn wake_with_hwnd(hwnd: HWND) {
+    unsafe { PostMessageW(hwnd, USER_EVENT_MSG_ID.get(), 0, 0) };
```
msiglreith, line 773: posting `USER_EVENT_MSG_ID` without sending a corresponding message will cause a panic

examples/custom_events.rs
```diff
@@ -18,30 +20,35 @@ fn main() -> Result<(), impl std::error::Error> {
...
+                break;
+            }
+            waker.wake_by_ref();
         }
     });
 
     event_loop.run(move |event, elwt| match event {
-        Event::UserEvent(event) => println!("user event: {event:?}"),
+        Event::NewEvents(_) => {
+            for event in receiver.iter() {
```
msiglreith, line 48: blocking inside `NewEvents` feels not great, I would suggest `try_iter` here

kchibisov: I'd prefer to have separate event for all of that tbf, so you kind of know that you woke up due to your event.

madsmtm: > I would suggest `try_iter` here

I've changed, thanks!

> I'd prefer to have separate event for all of that tbf, so you kind of know that you woke up due to your event.

I think a `UserWoken` event is a bad idea, for that to be useful we'd have to dispatch it multiple times, once per invocation of `wake`.

This would mean that a valid chain of events is:
```
NewEvents(WaitCancelled { .. })
UserWoken
UserWoken
UserWoken
AboutToWait
```

Which seems counter-intuitive.

I'd be more inclined to adding it as something like `StartCause::UserWoken`, but that has its own set of problems.

kchibisov: No, you invoke it only once, so you don't always poll your events if you don't have anything. Doing in `init` forces to do the checking every time, doing in separate event is the better way to put it.

The wakeup must be single per event loop. The WakeUp itself must be squashed. Basically how it's done in winit-next.

The point of being useful is that you could have complex sources of events and some may require locking to poll, so having a dedicated event sounds better, it also separates the logic and makes it more clear.

madsmtm: > Doing in `init` forces to do the checking every time

Fair point.

My primary fear is that I don't want people to do:
```rust
Event::UserWoken => {
    let event = receiver.try_recv().unwrap();
    // Handle event
}
```

As they may then miss events, and I think that is more easily avoided if the wakeup is bundled into `NewEvents`.

kchibisov: I don't understand what events are they going to miss. The order is `NewEvents`, and then `Wakeup` event, they are explicitly aware of such thing. The thing is that it also doesn't really matter, since they can poll in `NewEvents` and `Wakeup` event, because `NewEvents` and `Wakeup` are coming together, it's just a way to optimize/clean things a bit, given that winit already handled the part that you have new event.

madsmtm: I don't want three calls to `waker.wake_by_ref()` to be guaranteed to result in three `Wakeup`/`UserWoken` events, but I fear the user might _assume_ that's how it works, and design their code around that assumption.

One solution to this would be to fold the event into `NewEvents`, as I'm proposing. Another is better documentation.

> it also doesn't really matter

No this is definitely nitpicking.

src/event_loop.rs
```diff
@@ -462,9 +472,31 @@ impl<T: 'static> EventLoopProxy<T> {
...
+    }
+}
+
+// Note: This does not have the generic from `EventLoopProxy<T>`, since this
+// is the new API that shouldn't need it.
+impl From<EventLoopProxy> for Waker {
+    fn from(proxy: EventLoopProxy) -> Self {
+        proxy.waker()
+    }
 }
```
dhardy, line 500: I'm not sure this is wanted — eventually `EventLoopProxy` will (presumably) just be replaced with a `Waker`.

In fact, maybe just deprecate `EventLoopProxy` and `create_proxy` now, replacing with `EventLoop::create_waker`?

dhardy (commented):
I support this change: it makes winit simpler (at least at the API level).

Also, even though I currently use `UserEvent`, half the reason for that is just to create a `Waker` over it.

src/event_loop.rs
kchibisov: I want to reuse proxy in the future for some internal stuff in the future around `Send`/`Sync` and also, not be limited on the `Waker` API, since it's rather limited. But yeah, in the current state they are mostly the same.

nicoburns:
We're doing this with custom code in Dioxus. It would definitely be nice to have in Winit.

dhardy:
> * Does the user need the ability to handle the case where the event loop has been destroyed?

No: the only purpose of waker is to awaken the event-loop. The current changes to winit (since v0.30) already force sending of user payloads through something like an `mpsc` channel, and this is fine in my opinion.

FYI I updated Kas to use winit last week in this PR: https://github.com/kas-gui/kas/pull/583. You can also see [my own `Waker` implementation there](https://github.com/kas-gui/kas/blob/3a49aa636d6000f9aaed097eb4b99e8bb646e6db/crates/kas-core/src/runner/runner.rs#L160). So providing a waker in winit is not essential, but preferable considering how much `unsafe` code is required.

PR #2767, opened 2023-04-12
Re-work event loop run APIs: adds `run -> Result<>`, `run_ondemand` and `pump_events`

rib:
Overall this re-works the APIs for how an `EventLoop` is run across all backends, based on the recent _loong_ discussion in #2706

The changes let us address several open issues and cover these use-cases, with varying portability caveats:

1. A portable `run()` API that consumes the `EventLoop` and runs the loop on the calling thread until the app exits. This can be supported across _all_ platforms and compared to the previous `run() -> !` API is now able to return a `Result` status on all platforms except iOS and Web. Fixes: #2709
2. A less portable `run_ondmand()` API that covers the use case in #2431 where applications need to be able to re-run a Winit application multiple times against a persistent `EventLoop`. This doesn't allow `Window` state to carry across separate runs of the loop, but does allow orthogonal re-instantiations of a gui application.

   Available On: Windows, Linux, MacOS, Android
   Incompatible with iOS, Web
   Fixes: #2431
3. A less portable (and on MacOS, likely less-optimal) `pump_events()` API that covers the use case in #2706 and allows a Winit event loop to be embedded within an external `loop {}`. Applications call `pump_events()` once per iteration of their own external loop to dispatch all pending Winit events, without blocking the external loop.

   Available On: Windows, Linux, MacOS, Android
   Incompatible With: iOS, Web
   Fixes: #2706

Each method of running the loop will behave consistently in terms of how `NewEvents(Init)`, `Resumed` and `LoopDestroyed` events are dispatched (so portable application code wouldn't necessarily need to have any awareness of which method of running the loop was being used)

In particular by moving away from `run() -> !` we stop calling `std::process::exit()` internally as a means to kill the process without returning which means it's possible to return an exit status and applications can return from their `main()` function normally. This also fixes Android support where an Activity runs in a thread but we can't assume to have full ownership of the process (other services could be running in separate threads).

`run_return` has been removed, and the overlapping use cases that `run_return` previously partially aimed to support have been split across `run_ondemand` and `pump_events`.

To help test the changes this adds:
```
  examples/window_ondemand
  examples/window_pump_events
  examples/window_pump_events_rfd
```

The last _rfd example, tests the interaction between the `rfd` crate and using `pump_events` to embed Winit within an external event loop, and confirms that #2752 is also fixed by this change.

Additionally all examples have generally been updated so that `main()` returns a `Result` from `run()`

Fixes: #2709
Fixes: #2706
Fixes: #2431
Fixes: #2752

[...]

PR #3425, opened 2024-01-25
Remove `EventLoopError::AlreadyRunning`

madsmtm:
This is already prevented by the type-system, and as such it doesn't make sense to have an error case for this.

This was added by @rib in https://github.com/rust-windowing/winit/pull/2767, maybe you can shed some light on if this was an oversight added out of an abundance of caution, or if you think my reasoning is wrong here, and the error actually can happen?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ Unreleased` header.
 - on Windows: add `with_border_color`, `with_title_background_color`, `with_title_text_color` and `with_corner_preference`
 - On Windows, Remove `WS_CAPTION`, `WS_BORDER` and `WS_EX_WINDOWEDGE` styles for child windows without decorations.
 - On Windows, fixed a race condition when sending an event through the loop proxy.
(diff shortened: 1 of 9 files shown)


MarijnS95 (approved):
From a quick view this makes sense. The `EventLoop` cannot be mutably borrowed more than once, and there's no way to `Copy`/`Clone` one to get multiple mutable references either.

kchibisov (approved):

kchibisov:
I don't think that it's an issue actually? Like you can probably start with pump events for one iteration, but then continue with `run` or `run_on_demand`?

kchibisov:
At least it should kind of work now.

rib:
> Like you can probably start with pump events for one iteration, but then continue with run or run_on_demand?

Yeah I think it's just that you'd get ambiguous semantics to try and do something like that, e.g. considering that you probably won't get a `StartCause::Init` event.

Since it's a weird case, I think my preference at the time was to throw an error instead of having to think about what semantics to promise users.

PR #3430, opened 2024-01-27
Document when our team meetings happen, and link to the meeting notes

madsmtm:


README.md
```diff
@@ -19,10 +19,9 @@ For features _outside_ the scope of winit, see [Are we GUI Yet?](https://arewegu
 
 ## Contact Us
 
-Join us in any of these:
+Join us in our [![Matrix](https://img.shields.io/badge/Matrix-%23rust--windowing%3Amatrix.org-blueviolet.svg)](https://matrix.to/#/#rust-windowing:matrix.org) room. If you don't get an answer there, try [![Libera.Chat](https://img.shields.io/badge/libera.chat-%23winit-red.svg)](https://web.libera.chat/#winit).
```
madsmtm, line 22: @kchibisov: Why do we have both a Matrix and the Libera chat? Personally, I'm basically never active on the latter, but does it see use? Or should we deprecate it?

kchibisov: it's because the bridge died, I used it all the time and still present there, maybe if bridge resurrects.

madsmtm: Ah yeah, the IRC bridge, I thought we just sorta gave up on IRC when that happened. But if you're still active on Libera.Chat, and if people use it, then I'm fine with keeping the link.

daxpedda (approved):

README.md
daxpedda: Would prefer to remove it because not all maintainers are there, but all maintainers are on Matrix.

kchibisov: yeah, but some people can't be bothered to create matrix account to ask question, so I'm not sure how to _solve_ that part. With irc you just ask and go, though, it's kind of not that active right now.

README.md
```diff
@@ -19,10 +19,9 @@ For features _outside_ the scope of winit, see [Are we GUI Yet?](https://arewegu
...
 
-[![Matrix](https://img.shields.io/badge/Matrix-%23rust--windowing%3Amatrix.org-blueviolet.svg)](https://matrix.to/#/#rust-windowing:matrix.org)
-[![Libera.Chat](https://img.shields.io/badge/libera.chat-%23winit-red.svg)](https://web.libera.chat/#winit)
+The maintainers have a meeting every friday at UTC 14. The meeting notes can be found [here](https://hackmd.io/@winit-meetings).
```
kchibisov, line 24: Maybe we should have a matrix channel with meeting notifications, etc?

madsmtm: In my opinion, only if we actually decide to use that for calls as well, otherwise let's just keep it on Discord.

kchibisov (approved):

PR #3437, opened 2024-01-28
Fix swapped instance and general class names on X11

ulrikdem:
This fixes a bug introduced in commit d7ec899d, and brings the documentation in line with the implementation and the ICCCM standard. The commit messages contain more details.


notgull (approved):
Thanks!

Issue #3360, opened 2024-01-04
On Wayland, custom cursors don't respect alpha


eero-lehtinen:
I tested this gradient image
![gradient](https://github.com/rust-windowing/winit/assets/7013023/a451ed58-b49a-4ae8-b0fb-08ebaa2ef8a6)
as the cursor. It shows up perfectly fine on X11, but on Wayland, it shows up as just a red block with full opacity.

I don't really see anything wrong with the buffer creation. I'm not able to figure out what is wrong.

Here is my branch where I added the image to the example, but didn't get further https://github.com/eero-lehtinen/winit/tree/custom-cursor-gradient.

kchibisov:
See https://github.com/rust-windowing/winit/pull/3445

PR #3445, opened 2024-01-30
On Wayland, pre-multiply alpha for custom cursor

kchibisov:
Fixes: #3360

--

@daxpedda  I guess we should document whether alpha is premultiplied or not? The assumption is that it's probably not premultiplied?


daxpedda:
Yes, it should not be pre-multiplied and you are correct, we should most definitely be documented.

Web already presumes it's not pre-multiplied:
https://github.com/rust-windowing/winit/blob/df8805c0d27ca3172f4fa1a6bc1bf9709c4c8d0b/src/platform_impl/web/cursor.rs#L602

kchibisov:
I guess that should work?

src/platform/web.rs
```diff
@@ -248,6 +248,8 @@ pub trait CustomCursorExtWebSys {
     /// It uses the [url css function](https://developer.mozilla.org/en-US/docs/Web/CSS/url),
     /// but browser support for image formats is inconsistent. Using [PNG] is recommended.
     ///
+    /// The alpha channel is assumed to be **not** premultiplied.
+    ///
```
daxpedda, line 252:
```suggestion
```
Not relevant here, because we don't parse the image, we pass it straight on as a URL to CSS.

daxpedda (changes requested):
Otherwise LGTM.

src/platform/web.rs
kchibisov: Hm, but you set that alpha is not premultiplied or you just pass as is when it's loaded like that?

daxpedda: When it's loaded from a URL it bypasses that part of the code indeed.

daxpedda (approved):
Doc changes approved!

src/platform_impl/linux/wayland/types/cursor.rs
```diff
@@ -37,12 +37,15 @@ impl CustomCursor {
...
+        for (canvas_chunk, rgba) in canvas.chunks_exact_mut(4).zip(image.rgba.chunks_exact(4)) {
+            // Alpha in buffer is premultiplied.
+            let alpha = rgba[3] as f32 / 255.;
+            let r = (rgba[0] as f32 * alpha) as u32;
+            let g = (rgba[1] as f32 * alpha) as u32;
+            let b = (rgba[2] as f32 * alpha) as u32;
+            let color = ((rgba[3] as u32) << 24) + (r << 16) + (g << 8) + b;
+            let array: &mut [u8; 4] = canvas_chunk.try_into().unwrap();
+            *array = color.to_le_bytes();
         }
```
notgull, line 49: I would add a comment here that this should be SIMD-optimized once the standard library exposes the proper primitives.

notgull (approved):

src/platform_impl/linux/wayland/types/cursor.rs
kchibisov: compilers are good in figuring out such loops, just pointless complexity.

PR #3447, opened 2024-01-31
Deprecate window creation with stale event loop

### LINKED ISSUE #3399: Merge all examples showing one feature into a single example
Having a lot of small examples doesn't make much sense, since they don't show much value and hard to maintain, so having a one big example show casing a lot of features could be better for both users and maintainers, since we can manually test how features work with each other.

Examples that show different event loop API or platform specific stuff could be left as is.

kchibisov:
Creating window when event loop is not running generally doesn't work,
since a bunch of events and sync OS requests can't be processed. This
is also an issue on e.g. Android, since window can't be created outside
event loop easily.

Thus deprecate the window creation when event loop is not running,
as well as other resource creation to running event loop.

Given that all the examples use the bad pattern of creating the window
when event loop is not running and also most example existence is
questionable, since they show single thing and the majority of their
code is window/event loop initialization, they wore merged into
a single example 'window.rs' example that showcases very simple
application using winit.

Fixes #3399.

Replaces https://github.com/rust-windowing/winit/pull/3299 and https://github.com/rust-windowing/winit/pull/3400.

- [x] Tested on all platforms changed // Kind of.

--

@daxpedda could you wire the web example into `window.rs` example somehow? We maybe also want android there as well, @MarijnS95 , just without drawing, I guess?


kchibisov:
That's based on https://github.com/rust-windowing/winit/pull/3299

In general, I have the following questions:

1. Should we add `ActiveEventLoop` getter for `pump_events` trait? The API is kind of weird and requiring it to be normal doesn't sound right.
2. Maybe some types like `CustomCursor` should accept `EventLoop` instead?

notgull (approved):
>     1. Should we add `ActiveEventLoop` getter for `pump_events` trait? The API is kind of weird and requiring it to be normal doesn't sound right.

I don't think this works out.

>     2. Maybe some types like `CustomCursor` should accept `EventLoop` instead?

Maybe we should keep the `EventLoopWindowTarget` type around for this reason.

>     3. Do we need threaded rendering example...? softbuffer is a bit tricky around threads.

I'd agree.

>     4. Do we need something else in window example? I've added pretty much everything, though I don't print just all events, since it's a bit noisy.

Nah.

src/window.rs
```diff
@@ -1398,15 +1399,14 @@ impl Window {
     ///
     /// ```no_run
     /// # use winit::dpi::{LogicalPosition, PhysicalPosition};
-    /// # use winit::event_loop::EventLoop;
     /// # use winit::window::Window;
-    /// # let mut event_loop = EventLoop::new().unwrap();
-    /// # let window = Window::new(&event_loop).unwrap();
+    /// # fn scope(window: &Window) {
```
madsmtm, line 1403: Maybe?
```suggestion
    /// # let window: &Window = unimplemented!();
```
Since the examples are `no_run` anyways.

madsmtm (commented):
I think https://github.com/rust-windowing/winit/pull/3400 is the better way forwards for this, as it allows us to still support the "create event loop while it's not running" use-case, but allows us to mark it as deprecated for a release cycle.

kchibisov:
>"create event loop while it's not running" 

Do you mean `Window` here? I'm not so sure about it, since I'd rather force users here to more correct design and allow us to deal with soundness stuff more nicely. The issue is though, that `pump_events` is a bit out of place and I'm not sure how to enforce lifetimes for it, etc. Like the users are more likely to drop the event loop when using it all together which could lead to issues all over the place. But I guess `pump_events` stuff should be discussed separately and probably marked as `unsafe` in the first place?

src/window.rs
kchibisov: I'm not sure about it since all the code in examples will be kind of _dead_ with such pattern.

kchibisov:
>I think https://github.com/rust-windowing/winit/pull/3400 is the better way forwards for this

I've applied it for now the way you've suggested. Though, I haven't touched the `CustomCursor` for example.

kchibisov:
>Maybe we should keep the EventLoopWindowTarget type around for this reason.

I think the way forward would be to move `CustomCursor` into the `EventLoop` instead of getting the `&EventLoopWindowTarget`.

src/window.rs
madsmtm: Yeah, on the other hand, if someone forgets `no_run`, then the example will still mistakenly run, it'll just do nothing.

madsmtm:
Proposal for custom cursors we discussed today:
```rust
impl EventLoop {
    fn create_cursor(...);
}

impl ActiveEventLoop {
    fn create_cursor(...);
}
```

examples/window.rs
```diff
@@ -1,43 +1,980 @@
...
+            WindowEvent::TouchpadPressure { .. }
+            | WindowEvent::HoveredFileCancelled
+            | WindowEvent::KeyboardInput { .. }
+            | WindowEvent::CursorEntered { .. }
+            | WindowEvent::AxisMotion { .. }
+            | WindowEvent::DroppedFile(_)
+            | WindowEvent::HoveredFile(_)
+            | WindowEvent::Destroyed
+            | WindowEvent::Touch(_)
+            | WindowEvent::Moved(_) => (),
```
madsmtm, line 370: Nit: I'm not sure we should encourage exhaustive matching? It makes it more of a breaking change to add a variant.
```suggestion
            _ => {},
```

examples/child_window.rs
```diff
@@ -13,52 +6,53 @@ mod fill;
...
     let mut windows = HashMap::new();
 
     let event_loop: EventLoop<()> = EventLoop::new().unwrap();
-    let parent_window = Window::builder()
-        .with_title("parent window")
-        .with_position(Position::Logical(LogicalPosition::new(0.0, 0.0)))
-        .with_inner_size(LogicalSize::new(640.0f32, 480.0f32))
-        .build(&event_loop)
-        .unwrap();
+    let mut parent_window_id = unsafe { WindowId::dummy() };
```
madsmtm, line 36: Nit: Let's try to avoid `unsafe` in examples when we can.

```suggestion
    let mut parent_window_id = None;
```

CHANGELOG.md
```diff
@@ -11,15 +11,20 @@ Unreleased` header.
 
 # Unreleased
 
+- **Breaking:** Deprecate `EventLoop::create_window`.
+- **Breaking:** Move `Window::new` to `ActiveEventLoop::create_window` and `EventLoop::create_window`
```
madsmtm, line 15: Nit:

```suggestion
- **Breaking:** Move `Window::new` to `ActiveEventLoop::create_window` and `EventLoop::create_window`. The latter is deprecated, you should instead be creating your window inside the `NewEvents(StartCause::Init)` or `Resumed` events.

    ```rust
    // Before
    let window = Window::new(&event_loop);
    event_loop.run(|event, event_loop| {
        match event {
            // Handle events
        }
    });
    
    // After
    let mut window = None;
    event_loop.run(|event, event_loop| {
        match event {
            Event::Resumed => {
                window = Some(event_loop.create_window(Window::attributes()));
            }
            Event::Suspended => {
                window = None;
            }
            // .. Handle events
        }
    });
    ```
```

examples/pump_events.rs
```diff
@@ -25,38 +25,39 @@ fn main() -> std::process::ExitCode {
...
+                Event::WindowEvent { event, .. } => {
+                    let window = window.as_ref().unwrap();
+                    match event {
+                        WindowEvent::CloseRequested => event_loop.exit(),
+                        WindowEvent::RedrawRequested => fill::fill_window(window),
+                        _ => (),
+                    }
+                }
+                Event::AboutToWait => {
+                    window.as_ref().unwrap().request_redraw();
```
madsmtm, line 53: Nit:

```suggestion
                    window.unwrap().request_redraw();
```

src/event_loop.rs
```diff
@@ -45,11 +45,9 @@ pub struct EventLoop<T: 'static> {
 /// Target that associates windows with an [`EventLoop`].
 ///
 /// This type exists to allow you to create new windows while Winit executes
-/// your callback. [`EventLoop`] will coerce into this type (`impl<T> Deref for
-/// EventLoop<T>`), so functions that take this as a parameter can also take
-/// `&EventLoop`.
-pub struct EventLoopWindowTarget {
-    pub(crate) p: platform_impl::EventLoopWindowTarget,
+/// your callback.
+pub struct ActiveEventLoop {
```
madsmtm, line 49: There was a reason that I included a lifetime here in https://github.com/rust-windowing/winit/pull/3299; this was done so that it'd be easier to change to a mutable reference in the future.

But I'm fine with not doing it yet, and maybe we should just be transitioning to a mutable reference directly.

src/window.rs
```diff
@@ -168,7 +156,7 @@ pub struct WindowAttributes {
     pub fullscreen: Option<Fullscreen>,
     // Platform-specific configuration.
     #[allow(dead_code)]
-    pub(crate) platform_specific: PlatformSpecificWindowBuilderAttributes,
+    pub(crate) platform_specific: PlatformSpecificWindowAttributesAttributes,
```
madsmtm, line 159: Nit: This has two `Attributes` in it.

```suggestion
    pub(crate) platform_specific: PlatformSpecificWindowAttributes,
```

src/window.rs
```diff
@@ -530,50 +511,18 @@ impl WindowBuilder {
...
-        Window::builder().build(event_loop)
-    }
-
-    /// Create a new [`WindowBuilder`] which allows modifying the window's attributes before creation.
+    /// Create a new [`WindowAttributes`] which allows modifying the window's attributes before
+    /// creation.
     #[inline]
-    pub fn builder() -> WindowBuilder {
-        WindowBuilder::default()
+    pub fn attributes() -> WindowAttributes {
```
madsmtm, line 524: I'd prefer this name as it more clearly says what it is.

```suggestion
    pub fn default_attributes() -> WindowAttributes {
```

src/window.rs
```diff
@@ -123,24 +129,6 @@ impl From<u64> for WindowId {
     }
 }
```
madsmtm, line 131: Let's add a deprecated alias here.

```suggestion

#[deprecated = "renamed to `WindowAttributes`"]
pub type WindowBuilder = WindowAttributes;

```

madsmtm (approved):
I like the combined example!

> Should we add `ActiveEventLoop` getter for `pump_events` trait? The API is kind of weird and requiring it to be normal doesn't sound right.

I'm very unsure, but let's punt on it for now, as users still have the option of using `EventLoop::create_window`.

> Do we need threaded rendering example...?

Would be kinda nice at some point, but it can definitely wait.

madsmtm:
@daxpedda I've opened https://github.com/rust-windowing/winit/issues/3473 for making the big example work on the web.

examples/window.rs
kchibisov: it's mostly to show things we have, don't see why not?

src/event_loop.rs
kchibisov: I mean the 0.31 will just take `&mut` everywhere, but from the user side they won't need much to change, I think. Also, I can't really do lifetime unless we agree to wrap window target into some struct carrying `&'a mut EventLoopWindowTarget`.

kchibisov:
I'd like to merge it after the https://github.com/rust-windowing/winit/pull/3471 , so it'll be easier for me to do patch release, since #3471 will go to 0.29...

src/window.rs
kchibisov: I'd still probably go with `fn` for now.

examples/pump_events.rs
kchibisov: you can't do it like that.

CHANGELOG.md
kchibisov: See https://github.com/rust-windowing/winit/issues/3477

Would probably send it separately though.

src/window.rs
madsmtm: True, it's not just a rename, though it does provide much the same functionality, so having an alias is more discoverable for people that don't read the changelog...

But not a hill I want to die on.

CHANGELOG.md
madsmtm: > Do we really want that in the CHANGELOG?

Well, migrations are only really relevant for a particular release, so I think the changelog is the _perfect_ place to put those.

examples/window.rs
madsmtm: I think I wanted to make `WindowEvent` be `#[non_exhaustive]`, but since we don't have that, I guess it doesn't matter that much.

src/window.rs
kchibisov: I mean, you still link to them from the changelog, I just want this thing to be present in rustdoc as well, since the changelog itself has a lot of things.

madsmtm: Well, there's `#[doc(hidden)]` for that. So maybe:

```suggestion

#[deprecated = "functionality moved to `WindowAttributes`"]
#[doc(hidden)]
pub type WindowBuilder = WindowAttributes;

```

examples/window.rs
kchibisov: Given that `WindowEvent` will be likely gone or drastically reworked in 0.31 marking it non_exhaustive sounds like not really a great thing for 0.30.

src/window.rs
kchibisov: Hm, I commented in the wrong thread, but I don't think it really matters.

The thing is that we removed the _core_ method, so you kind of discover the `WindowAttributes` one way or another.

madsmtm: Fair enough, I see that it doesn't matter much.

jackpot51 (approved):
Orbital changes are good

madsmtm:
@daxpedda I believe this is waiting mostly on your review

src/window.rs
daxpedda: What happened here?

daxpedda (approved):
Wow, that's actually a lot of changes in one PR.
LGTM!

src/window.rs
madsmtm: I think maybe a rebase made GitHub close my comments automatically, and then I just assumed they were fixed?

kchibisov: the branch is out of dated, so I resolved them but not pushed yet.

madsmtm (approved):

src/window.rs
```diff
@@ -218,26 +206,19 @@ unsafe impl Send for SendSyncRawWindowHandle {}
...
-    pub fn parent_window(&self) -> Option<&rwh_06::RawWindowHandle> {
-        self.parent_window.as_ref().map(|handle| &handle.0)
-    }
-}
-
-impl WindowBuilder {
     /// Initializes a new builder with default values.
     #[inline]
-    #[deprecated = "use `Window::builder` instead"]
+    #[deprecated = "use `Window::default_attributes` instead"]
```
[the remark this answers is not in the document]
kchibisov: Isn't the point to suggest Window type to avoid import in the first place? Otherwise I'd rather remove the thing all together and say the users to just `WindowAttributes::default` or something.

examples/child_window.rs
```diff
@@ -75,23 +67,28 @@ fn main() -> Result<(), impl std::error::Error> {
...
     })
 }
 
-#[cfg(not(all(
+#[cfg(all(
     feature = "rwh_06",
-    any(x11_platform, macos_platform, windows_platform)
-)))]
+    not(any(x11_platform, macos_platform, windows_platform))
+))]
```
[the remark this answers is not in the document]
kchibisov: Idk, I don't like negations.

src/window.rs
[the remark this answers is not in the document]
madsmtm: Yeah, I'm fine with just suggesting to use `Window::default_attributes`

Issue #3448, opened 2024-02-01
It is a foot-gun that `EventLoopProxy` is `!Sync` on macOS


gretchenfrage:
Hi all, thank you for your consideration. Someone previously created #3220 on this subject, but then closed their own issue when they found that there actually is an obstacle to making `EventLoopProxy` `Sync` on macOS. However, I'm creating this issue for further discussion because I think it is unfortunate that the API is as such.

`EventLoopProxy` being `Sync` only on non-macOS platforms makes it so that one can write code with winit that doesn't seem to be using any platform-specific features but nonetheless fails to compile on macOS while compiling fine on other platforms, which I think goes against winit's generally very nice goals and behaviors in terms of cross-platform abstraction.

Some possible ways I can imagine dealing with this include:

- Make `EventLoopProxy` always `!Sync`
- Make `EventLoopProxy` `Sync` on macOS by wrapping its internals in a mutex
- Make `EventLoopProxy` always `!Sync`, but create a platform-specific `SyncEventLoopProxy` type which is like `EventLoopProxy` but `Sync`, and which `EventLoopProxy` can be converted into through a platform specific-API
- Create a non-default feature flag which makes `EventLoopProxy` `Sync` and document that that feature flag can only be enabled on certain platforms

madsmtm:
I mean, `EventLoopProxy` is `Clone`, so you can always just clone it just before sharing it with another thread.

That said, the underlying `CFRunLoopSourceRef` is considered being marked `Send + Sync` in https://github.com/servo/core-foundation-rs/pull/649, and I'm pretty sure that's correct, the `CFRunLoopSourceRef` is immutable, so should be safe for us to just add an `unsafe impl<T: Send> impl Sync for `EventLoopProxy<T> {}`, at least on macOS.

madsmtm:
Seems like this will need some work on the web too, @daxpedda?

PR #3449, opened 2024-02-01
Make `EventLoopProxy` `Sync`

madsmtm:
Fixes https://github.com/rust-windowing/winit/issues/3448.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Unreleased` header.
 - On X11, don't require XIM to run.
 - On X11, fix xkb state not being updated correctly sometimes leading to wrong input.
 - Fix compatibility with 32-bit platforms without 64-bit atomics.
+- Implement `Sync` for `EventLoopProxy<T: Send>`.
 - On X11, fix swapped instance and general class names.
 - **Breaking:** Removed unnecessary generic parameter `T` from `EventLoopWindowTarget`.
 - On Windows, macOS, X11, Wayland and Web, implement setting images as cursors. See the `custom_cursors.rs` example.
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -253,7 +253,8 @@ impl<T> EventLoop<T> {
         self.event_loop.run(event_handler)
     }
 
-    /// Creates an [`EventLoopProxy`] that can be used to dispatch user events to the main event loop.
+    /// Creates an [`EventLoopProxy`] that can be used to dispatch user events
+    /// to the main event loop, possibly from another thread.
     pub fn create_proxy(&self) -> EventLoopProxy<T> {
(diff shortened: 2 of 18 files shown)


kchibisov:
Maybe we should push the `Waker` PR forward instead, since it should be `Send` + `Sync`?

daxpedda (approved):
This is great!

daxpedda:
> Maybe we should push the `Waker` PR forward instead, since it should be `Send` + `Sync`?

I don't think it hurts to merge this in the meantime.

tests/sync_object.rs
```diff
@@ -1,6 +1,15 @@
 #[allow(dead_code)]
 fn needs_sync<T: Sync>() {}
 
+#[test]
+fn event_loop_proxy_send() {
+    #[allow(dead_code)]
+    fn is_send<T: 'static + Send>() {
+        // ensures that `winit::EventLoopProxy` implements `Send`
+        needs_sync::<winit::event_loop::EventLoopProxy<T>>();
+    }
```
kchibisov, line 10: This should just use `needs_sync`, because we test `Send` in other test.

src/event_loop.rs
```diff
@@ -449,6 +450,8 @@ unsafe impl rwh_05::HasRawDisplayHandle for OwnedDisplayHandle {
 }
 
 /// Used to send custom events to [`EventLoop`].
+///
+/// This type is thread-safe.
```
kchibisov, line 454: That's already clear from `Send` + `Sync` contraints.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ Unreleased` header.
 
 # Unreleased
 
+- Make `EventLoopProxy<T: Send>` be `Sync`.
```
```suggestion
- Implement `Sync` for `EventLoopProxy<T: Send>`.
```

kchibisov (changes requested):

tests/sync_object.rs
daxpedda: This doesn't test if `EventLoopProxy` implements `Send`, it tests if `EventLoopProxy` implement `Sync` if `T` is `Send`.

kchibisov: But you can move the `T: Send` in the test itself?

daxpedda: I don't think you can use generics on `#[test]`. It would be a compile error anyway so we don't *really* needs tests in this sense.

I think the way it's right now is just fine, but maybe adding a comment would help.

madsmtm: Not sure what I should write as a comment, but I've made the exiting comment text a bit clearer

kchibisov (approved):

notgull (approved):

PR #3450, opened 2024-02-02
windows: HID raw input for gamepads and joysticks

valaphee:
This PRs goal is to rudimentary gamepad and joystick support through device events (Motion and Button) and therefore don't require button/axis mapping as they are device dependent.

### DIFF
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -3,8 +3,8 @@
 mod runner;
 
 use std::{
-    cell::Cell,
-    collections::VecDeque,
+    cell::{Cell, RefCell},
+    collections::{HashMap, VecDeque},
     ffi::c_void,
     marker::PhantomData,
     mem, panic, ptr,
@@ -20,7 +20,7 @@ use std::{
 use once_cell::sync::Lazy;
 
 use windows_sys::Win32::{
-    Devices::HumanInterfaceDevice::MOUSE_MOVE_RELATIVE,
+    Devices::HumanInterfaceDevice::{HidP_GetUsagesEx, HidP_GetUsageValue, HidP_Input, MOUSE_MOVE_RELATIVE},
     Foundation::{HWND, LPARAM, LRESULT, POINT, RECT, WPARAM},
     Graphics::Gdi::{
         GetMonitorInfoW, MonitorFromRect, MonitorFromWindow, RedrawWindow, ScreenToClient,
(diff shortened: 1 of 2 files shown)


valaphee:
It might be worth to add some way to of identification in the device added message, as at the moment its not possible to know if its a keyboard, mouse, joystick, etc. and maybe what product was attached.

daxpedda:
We have discussed this in todays meeting and we still need to do more research for some platforms, but the current discussion revolves around the fact that we need a common type for device IDs for all backends.

In general this will need some more design work first!

valaphee:
This PR just extends on the already existing capabilities. A proper gamepad and joystick API should definitely have stable device ids (for hotplug), way to control rumble, or more device specific, leds etc. and map the buttons (which requires definitions), but idk if this should be in winit.

The better approach would be to use the most high-level api on the platform, like on Windows just use WGI (which is also better for Xbox support, and HID report parsing is sometimes inaccurate), X11 or Wayland would still require custom mapping, Web theoretically not.

daxpedda (changes requested):
So right now we came to the conclusion that all backends (except Windows) can do this without any help from Winit at all.

@msiglreith would be great if you could clarify if that applies to Windows as well.

msiglreith:
@daxpedda should be possible on Windows as well, just need to create an additional hidden window same as the current EventLoop implementation.

daxpedda:
> @daxpedda should be possible on Windows as well, just need to create an additional hidden window same as the current EventLoop implementation.

I would consider that if this doesn't work without creating an empty hidden window, it should indeed live in Winit, My assumption here is that this would otherwise introduce not insignificant overhead and would probably need a separate thread?

So this leaves us with three ways to move forward here:
1. Cross-platform event.
2. Platform-specific event for Windows only (and maybe for X11 temporarily as well).
3. Push towards external integration with the event loop by the user.

I'm currently leaning towards 1.

valaphee:
Adding RawInput also just extends the already existing infrastructure, that's why I found it a good idea to add it here.

Even though it's also not that flexible, because with raw input, you can also use device specific features, as you recv and send raw hid reports.
The same functionality can also be achieved with https://github.com/ruabmbua/hidapi-rs without significant overhead.

This API is also not that useful on its own as some devices have different analog ranges, completely different button mappings, etc. (that's why SDL2 mappings exist)
But DeviceEvents are expected to be raw.

daxpedda:
> Either way, gamepad handling is not really a requirement as external crates just work fine. But it would make sense to integrate into existing apis, like emitting DeviceEvents and building a proper gamepad api, etc. (with mapping)

Our current stance has been to leave it to external implementations as long as its possible.
So I don't believe we are going to go further then just emitting HID addition/removal here.

Frankkkkk:
Hi @daxpedda , Would it be possible to show a simple example (or point to a repo that uses them) of HID events as an external impl.? That would be awesome; thanks!

daxpedda:
The suggestion here was to implement e.g. a gamepad API externally, not HID events, which remains the single motivation to continue this in Winit.

Apologies for the confusion!

Issue #2831, opened 2023-05-31
web: prevent_default depending on keys


Vrixyz:
I'm contributing on bevy_egui, comparing its functionalities with eframe.

eframe has a fine control over which keys it wants to prevent its defaults or not, I think this could be a good addition to winit:
https://github.com/emilk/egui/blob/307565efa55158cfa6b82d2e8fdc4c4914b954ed/crates/eframe/src/web/events.rs#L101

Vrixyz:
I suggest changing the current exposed `prevent_default: bool` to something like `enum{Always, Check(Fn<KeyEvent, bool>), Never})` ; not really sure of the impacts though.

daxpedda:
Sounds reasonable to me.
I was trying to figure out if we can skip `preventDefault` completely and rely on fine-grained events, unfortunately that's not possible, there isn't an event for every command: e.g. "F12", opening the developer console.

PR's welcome.

daxpedda:
I think one thing we could do for sure is to detect copy/paste/cut shortcuts and not call `preventDefault()` by default. I believe that's the main use-case currently for something like this anyway.

ddfisher:
I'm looking at implementing this, but wanted to check in first.

My current thinking is to take a closure instead of adding an enum (i.e. change `prevent_default` from `bool` to `Fn(web_sys::Event) -> bool`). It'd still be straightforward for folks who want to have it always be true or false (`|_| true` is still pretty concise) and IMO adds a bit less complexity overall.

A couple drawbacks:
- This will have to be a (relatively minor) breaking change, as we wouldn't be able to support the `WindowExtWebSys::prevent_default` bool getter.  I don't see any way around at least modifying the return value of this function if we want to support more fine-grained control here, though.  My inclination is to just remove it, if that seems reasonable?
- There are a variety of event types which could call this closure, so users will need to use `wasm_bindgen::dyn_ref` (or other members of `JsCast`) to act on event types they care about. I'd love to make this more ergonomic, but I don't see a great way of doing that.

How does this sound as a way forward?

daxpedda:
As pointed out in https://github.com/rust-windowing/winit/issues/2831#issuecomment-1575535287:
> I think one thing we could do for sure is to detect copy/paste/cut shortcuts and not block those by default. I believe that's the main use-case currently for something like this anyway.

I'm not sure what other use-cases are out there, my first impression is that this would satisfy the majority.
Hence: whats your use-case :blush:?

[...]

PR #3451, opened 2024-02-02
Add docs about how to get more fine-grained control of event.preventDefault() calls on web

ddfisher:
The example is a bit verbose, unfortunately, but I'm hoping it's relatively clear and it's also quite close to being directly useful.

See #2831 for more context.

### DIFF
--- a/src/platform/web.rs
+++ b/src/platform/web.rs
@@ -65,8 +65,76 @@ pub trait WindowExtWebSys {
     /// For example, by default using the mouse wheel would cause the page to scroll, enabling this
     /// would prevent that.
     ///
+    /// Specifically, this will call `event.preventDefault()` for the following event types:
+    /// `touchstart`, `wheel`, `contextmenu`, `pointerdown`, `pointermove`, `keyup`, and `keydown`.
+    ///
+    /// For fine-grained control over which events call `event.preventDefault()`, call
+    /// `Window::set_prevent_default(false)` and add an event listener on the canvas for each of
+    /// the above event types that selectively calls `event.preventDefault()`.
+    ///
     /// Some events are impossible to prevent. E.g. Firefox allows to access the native browser
     /// context menu with Shift+Rightclick.
+    ///
+    /// # Example
+    /// This example calls `event.preventDefault()` for all relevant events except for ctrl-c,
+    /// ctrl-x, and ctrl-v.
+    /// ```
+    /// # fn add_prevent_default_listeners(canvas: HtmlCanvasElement) {
+    /// # use wasm_bindgen::closure::Closure;
+    /// # use wasm_bindgen::JsCast;
+    /// let events_types_to_fully_prevent_default =
+    ///     ["touchstart", "wheel", "contextmenu", "pointermove"];
+    /// let key_events_to_partially_prevent_default = ["keyup", "keydown"];
+    /// let pointer_events_to_focus_and_prevent_default = ["pointerdown"];
+    ///
+    /// for event_type in events_types_to_fully_prevent_default.into_iter() {
+    ///     let prevent_default_listener =
+    ///         Closure::<dyn FnMut(_)>::new(move |event: web_sys::Event| {
+    ///             event.prevent_default();
+    ///         });
+    ///
+    ///     let _ = canvas.add_event_listener_with_callback(
+    ///         event_type,
+    ///         prevent_default_listener.as_ref().unchecked_ref(),
+    ///     );
+    ///     prevent_default_listener.forget();
+    /// }
+    ///
+    /// for event_type in pointer_events_to_focus_and_prevent_default.into_iter() {
+    ///     let stored_canvas = canvas.clone();
+    ///     let prevent_default_listener =
+    ///         Closure::<dyn FnMut(_)>::new(move |event: web_sys::Event| {
+    ///             event.prevent_default();
+    ///             let _ = stored_canvas.focus();
+    ///         });
+    ///
+    ///     let _ = canvas.add_event_listener_with_callback(
+    ///         event_type,
+    ///         prevent_default_listener.as_ref().unchecked_ref(),
+    ///     );
+    ///     prevent_default_listener.forget();
+    /// }
+    ///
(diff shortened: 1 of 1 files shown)


src/platform/web.rs
```diff
@@ -65,8 +65,76 @@ pub trait WindowExtWebSys {
     /// For example, by default using the mouse wheel would cause the page to scroll, enabling this
     /// would prevent that.
     ///
+    /// Specifically, this will call `event.preventDefault()` for the following event types:
+    /// `touchstart`, `wheel`, `contextmenu`, `pointerdown`, `pointermove`, `keyup`, and `keydown`.
```
daxpedda, line 69: This should mention that `Event.preventDefault()` is only called for `pointermove` in the case of a [chorded button interaction](https://www.w3.org/TR/pointerevents3/#chorded-button-interactions).

I know this is getting even more into the weeds, but it might be important ...

src/platform/web.rs
```diff
@@ -65,8 +65,76 @@ pub trait WindowExtWebSys {
...
+    ///
+    /// For fine-grained control over which events call `event.preventDefault()`, call
+    /// `Window::set_prevent_default(false)` and add an event listener on the canvas for each of
+    /// the above event types that selectively calls `event.preventDefault()`.
```
daxpedda, line 73:
```suggestion
    /// For fine-grained control over which events call `event.preventDefault()`, call
    /// `Window::set_prevent_default(false)` and add event listeners on the canvas that
    /// selectively call `event.preventDefault()`.
```
Users don't really have to address all these events.

src/platform/web.rs
```diff
@@ -65,8 +65,76 @@ pub trait WindowExtWebSys {
...
+    ///
     /// Some events are impossible to prevent. E.g. Firefox allows to access the native browser
     /// context menu with Shift+Rightclick.
+    ///
+    /// # Example
+    /// This example calls `event.preventDefault()` for all relevant events except for ctrl-c,
+    /// ctrl-x, and ctrl-v.
+    /// ```
+    /// # fn add_prevent_default_listeners(canvas: HtmlCanvasElement) {
```
daxpedda, line 82: It would be good to actually make a full example with `EventLoop::new()` and such.
Will be adding cross-compilation testing for rustdoc into the CI soon.

Doesn't need much, just create a `Window`, use [`WindowBuilder::with_append()`](https://docs.rs/winit/0.29.10/wasm32-unknown-unknown/winit/window/struct.WindowBuilder.html#method.with_append) and get the canvas.

src/platform/web.rs
```diff
@@ -65,8 +65,76 @@ pub trait WindowExtWebSys {
...
+    ///     let prevent_default_listener =
+    ///         Closure::<dyn FnMut(_)>::new(move |event: web_sys::Event| {
+    ///             event.prevent_default();
+    ///         });
+    ///
+    ///     let _ = canvas.add_event_listener_with_callback(
+    ///         event_type,
+    ///         prevent_default_listener.as_ref().unchecked_ref(),
+    ///     );
+    ///     prevent_default_listener.forget();
```
daxpedda, line 100:
```suggestion
    ///     prevent_default_listener.into_js_value();
```
See [`Closure::into_js_value()`](https://docs.rs/wasm-bindgen/0.2.90/wasm_bindgen/closure/struct.Closure.html#method.into_js_value).

src/platform/web.rs
```diff
@@ -65,8 +65,76 @@ pub trait WindowExtWebSys {
...
+    ///
+    /// # Example
+    /// This example calls `event.preventDefault()` for all relevant events except for ctrl-c,
+    /// ctrl-x, and ctrl-v.
+    /// ```
+    /// # fn add_prevent_default_listeners(canvas: HtmlCanvasElement) {
+    /// # use wasm_bindgen::closure::Closure;
+    /// # use wasm_bindgen::JsCast;
+    /// let events_types_to_fully_prevent_default =
+    ///     ["touchstart", "wheel", "contextmenu", "pointermove"];
```
daxpedda, line 86: While we are it, we might want to handle the chorded button interaction for `pointermove`.

src/platform/web.rs
```diff
@@ -65,8 +65,76 @@ pub trait WindowExtWebSys {
...
+    /// For fine-grained control over which events call `event.preventDefault()`, call
+    /// `Window::set_prevent_default(false)` and add an event listener on the canvas for each of
+    /// the above event types that selectively calls `event.preventDefault()`.
+    ///
     /// Some events are impossible to prevent. E.g. Firefox allows to access the native browser
     /// context menu with Shift+Rightclick.
+    ///
+    /// # Example
+    /// This example calls `event.preventDefault()` for all relevant events except for ctrl-c,
+    /// ctrl-x, and ctrl-v.
```
daxpedda, line 80: I think we should pick a different example because I consider the whole preventing clipboard interactions a Winit bug.

E.g. `ctrl-p` for printing? Maybe also hook up [`beforeprint`](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeprint_event) just to make the example complete.

daxpedda (changes requested):
Thank you, this is great!

Let me know if you need any help with anything.

src/platform/web.rs
ddfisher: Oh, interesting -- thanks!

src/platform/web.rs
ddfisher: Oh, interesting -- that's somewhat subtle. Changed!

src/platform/web.rs
ddfisher: Done!

src/platform/web.rs
ddfisher: Switched to `ctrl-p`.

src/platform/web.rs
ddfisher: Done!

ddfisher:
Let me know if this looks good now! Sorry for the slow turnaround here!

daxpedda (changes requested):

Issue #3433, opened 2024-01-27
Split Winit up into multiple crates


madsmtm:
Part of https://github.com/rust-windowing/winit/issues/3367, opening to discuss separately.

Winit is designed around a single crate with a specific set of backends, which is great for users that can use that, but [there is a need](https://github.com/rust-windowing/winit/pull/2430) for other backends that are not implemented in Winit today, like GTK. Additionally, the backend-specific extensions that we do have usually constrain some other part of the design.

So it would probably be useful to try to split Winit into multiple smaller crates (though still in the same repository), to ease doing this work. The top-level `winit` crate would still remain.

Along with this, we'd want a way to write out-of-tree backends, and allow the user to integrate them into their application. This could probably be solved by introducing internal / "backend" traits, and keep a `dyn` in structures inside the `winit` crate.

A rough implementation plan could be:
- [x] Create a `winit-core` crate, which contains the common types that all backends use (like `dpi` and keyboard types)
- [x] Make traits in `winit-core`, that implements the desired functionality, and move each backend to also implement those traits.
- [x] Move the main `winit` crate to an API that is not `cfg`-based, but `dyn Trait`-based.
- [x] Introduce a `winit-common` crate to have shared functionality between the crates, which will be intended for winit internal usage and help writing new backends.
- [x] Move the backends out into separate crates, which all depend on `winit-core`, but still have the main `winit` crate "merge" these backends with `cfg`s, not with a trait.

@kchibisov: I've tried to fairly faithfully reproduce what I felt was the important points from https://github.com/rust-windowing/winit/issues/3367, but please edit this issue description with your own.

EDIT: Implementation history:
- Refactor to `dyn` traits:
  - https://github.com/rust-windowing/winit/pull/4163
  - https://github.com/rust-windowing/winit/pull/4060
  - https://github.com/rust-windowing/winit/pull/3927
  - https://github.com/rust-windowing/winit/pull/3857
  - https://github.com/rust-windowing/winit/pull/4182
  - https://github.com/rust-windowing/winit/pull/4221
- Refactorings:
  - https://github.com/rust-windowing/winit/pull/3864
  - https://github.com/rust-windowing/winit/pull/3902
  - https://github.com/rust-windowing/winit/pull/3941
  - https://github.com/rust-windowing/winit/pull/4160
  - https://github.com/rust-windowing/winit/pull/4225
  - https://github.com/rust-windowing/winit/pull/4228
[...]

Issue #3064, opened 2023-08-29
Platform-Specific Enum Variants


daxpedda:
From https://github.com/rust-windowing/winit/pull/3056#issuecomment-1695542094:

If we want to continue using the extension trait model for platform-specific features, I came up with the following idea that we could use for enums:
```rust
enum Event {
    EventA,
    EventB(String),
    EventC {
        data: u16,
    },
    Platform(PlatformEvent),
}

impl PlatformEventExtWayland for PlatformEvent {
    fn event(&self) -> WaylandEvent {
        self.0
    }
}

// Not really needed, just a neat idea.
impl PlatformEventExtWindows for PlatformEvent {
    fn event(&self) -> ! {
        match self.0 { }
    }
}
```

Alternatively we can always mark the enum `#[non_exhausive]` and add `cfg`-gated variants.

---

For consistency with the current API I would like to go with the first solution. The first thing I would like to apply this to is to re-introduce the `ControlFlow` enum after #3056.

daxpedda:
From https://github.com/rust-windowing/winit/pull/3056#issuecomment-1695566561:
> I will say, I don't think `Exit` has a place in the enum, it's really a separate thing. If you want to have an enum, I think it should be something like `enum NextEvent { Poll, Wait, WaitUntil(Instant) }`.

So when we get to re-introduce the `ControlFlow` enum I would like to specifically tackle this as well.

daxpedda:
Obviously there will always be APIs that should be behind a `cfg` guard because they don't make sense in a cross-platform API, e.g. [`EventLoopBuilder::with_msg_hook()`](https://docs.rs/winit/0.29.1-beta/x86_64-pc-windows-msvc/winit/event_loop/struct.EventLoopBuilder.html#method.with_msg_hook).

I agree with the overall sentiment, but this is a much bigger issue. Generally speaking I prefer keeping the API consistent over having half of the API follow a different pattern/rule. That said, if we can get consensus right here and now to remove the platform-extension-trait model I'm happy to go with that as well.

@dhardy if you like to start the discussion in a new issue that would be great. I'm happy to do the implementation, but it's a lot of work to collect, read and answer throughout the process.

kchibisov:
I'm not sure why it has a separate issue, if enum is not a part of event, which a `ControlFlow` is, simply mark it as `non_exhaustive` and add make some variants `cfg` based.

[...]

PR #3455, opened 2024-02-05
Split out platform-independent types into a new winit-core crate

### LINKED ISSUE #2759: Split parts of winit into separate crates
I recently stumbled upon #1298, which feels like a pretty good idea to me. For some of the types in this crate, they could probably be split out into a `winit-types` crate (preferably developed as a separate crate in the `winit` repo) that avoid linking to backends.

It looks like, originally, `winit-types` and `glutin-x11-sym` (for X11 types) were planned to be split into other crates. I agree with `winit-types`, and I can also agree with splitting global X11 resources into another crate, since there are global resources that need a single authoritative source to be managed. The rationale is similar to that of the [`signal-hook-registry`](https://docs.rs/signal-hook-registry/latest/signal_hook_registry/). However,  there's no harm in having several X11 shared libraries lying around following AltF02/x11-rs#165

I think that `winit-types` and a hypothetical `winit-xlib` crate would be good candidates to be split off of `winit`. Is there any other room for splitting here? It feels like anything further would require us to slice and dice the `platform_impl` module, which just sounds like a pain. On top of that, the more we split off, the larger the dependency tree gets, and we have to consider if the gains in compartmentalization are worth it.

Would probably need to wait until after #2662

notgull:
This PR implements the first item of #3433 by splitting the platform-independent types from `winit` into a new crate: `winit-core`. It contains mostly types, traits and structures from `winit` that are platform independent. As part of this commit, I made a handful of changes to certain types (most notably `Event`) to make them more platform-independent.

A brief summary of the changes made:

- `dpi` and `keyboard` modules are copied over wholesale.
- The errors in the `error` module that don't involve `OsError` are copied as well.
- Some types from the `window` module are copied verbatim. `WindowId` is now a newtype over `u64` instead of a platform-specific value. `ActivationToken` is copied and given methods to make it useful.
- `Event` is given a new generic parameter to describe platform-specific key information. `InnerSizeWriter` is given methods to allow for instantiation. Aside from this, most things from the `event` module are copied verbatim.
- `AsyncRequestSerial` and `ControlFlow` are copied from `event_loop`. `AsyncRequestSerial::get` is made public.

Aside from this, the biggest changes are just moving everything to `winit-core` and then re-exporting it.

I've formatted this PR as a series of changes in order to make reviewing easier. @rust-windowing/winit Please review this PR quickly. It will be difficult to rebase on top of new changes and it'll take me the whole day to remake it.

Closes #2759

  - [x] Ran `cargo check` on all platforms to make sure it builds there.
  - [ ] Did not actually run example programs on all platforms.


notgull:
@jackpot51 Any insight into why the Redox CI is failing with `__errno not found` errors?

jackpot51:
@notgull not sure, it may be trying to compile with the host compiler instead of a cross compiler?

notgull:
It appears that the only failures are the Redox failure (is that new?) and what appears to be a bug in `cargo-apk`. Both of which are unrelated to this PR.

notgull:
No, the Redox failure is new. It looks like Redox can't compile `run-wasm`.

kchibisov:
Why it compiles it in the first place?

notgull:
> Why it compiles it in the first place?

Because with the new workspace layout it tries to compile everything instead of just `winit`

winit/src/platform_impl/ios/mod.rs
```diff
@@ -68,35 +68,22 @@ mod window;
...
 };
 pub(crate) use crate::cursor::NoCustomCursor as PlatformCustomCursor;
 pub(crate) use crate::cursor::NoCustomCursor as PlatformCustomCursorBuilder;
 pub(crate) use crate::icon::NoIcon as PlatformIcon;
 pub(crate) use crate::platform_impl::Fullscreen;
 
-/// There is no way to detect which device that performed a certain event in
-/// UIKit (i.e. you can't differentiate between different external keyboards,
-/// or wether it was the main touchscreen, assistive technologies, or some
-/// other pointer device that caused a touch event).
```
madsmtm, line 89: I'd like to retain this comment in some shape or form, it is nice to know that device IDs will never be supported on iOS.

winit-core/src/event.rs
```diff
@@ -683,7 +665,7 @@ pub struct RawKeyEvent {
 
 /// Describes a keyboard input targeting a window.
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
-pub struct KeyEvent {
+pub struct KeyEvent<Extra> {
```
madsmtm, line 668: I dislike this design, and think we need to discuss further how to do this sort of thing. See also https://github.com/rust-windowing/winit/issues/3064.

Cargo.toml
```diff
@@ -1,265 +1,16 @@
...
-    # iOS
-    "x86_64-apple-ios",
-    # Android
-    "aarch64-linux-android",
-    # Web
-    "wasm32-unknown-unknown",
+[workspace]
+members = [
+    "run-wasm",
+    "winit",
```
madsmtm, line 4: I think it's a mistake to move `winit` to a subdirectory, at least yet, and it also complicates this PR unnecessarily. There is especially value in having the top-level `examples` directory. Let's use the [`include`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-exclude-and-include-fields) field instead to specify which files should be included in the `winit `package.

madsmtm (changes requested):

PR #3458, opened 2024-02-07
Fix compatibility with platforms without AtomicU64

### LINKED ISSUE #3456: AtomicU64 not available on all platforms
https://github.com/rust-windowing/winit/pull/2955 added the use of `AtomicU64`, however as the [portability](https://doc.rust-lang.org/std/sync/atomic/index.html#portability) docs note

> PowerPC and MIPS platforms with 32-bit pointers do not have `AtomicU64` or `AtomicI64` types.

Looks like this is also the case for armv5te.

Now, these are all tier 2 or 3 platforms (from rustc perspective), so I'd understand if that's not a priority.  Just wanted to raise the issue, since the did regress from previous winit releases.

kchibisov:
Fixes #3456.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Unreleased` header.
 
 # Unreleased
 
+- Fix compatibility with 32-bit platforms without 64-bit atomics.
 - On X11, fix swapped instance and general class names.
 - **Breaking:** Removed unnecessary generic parameter `T` from `EventLoopWindowTarget`.
 - On Windows, macOS, X11, Wayland and Web, implement setting images as cursors. See the `custom_cursors.rs` example.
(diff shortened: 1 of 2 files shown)


kchibisov:
@daxpedda is it something relevant for web as well or web is a bit special here?

notgull (changes requested):
On a 32-bit computer it's reasonable to overflow this value in a matter of milliseconds, depending on the hardware. This will likely cause subtle bugs.

I would recommend using [`portable-atomic`](https://docs.rs/portable-atomic/latest/portable_atomic/) here, unless you think the possibility of overflow is worth the risk.

kchibisov:
>On a 32-bit computer it's reasonable to overflow this value in a matter of milliseconds, depending on the hardware. This will likely cause subtle bugs.

overflow is _desired_ though, it's even written like that in the comment. Also, unsigned numbers don't overflow, they `wrap around` generally and their behavior is deterministic here.

Like the end goal is to just do exactly that.

kchibisov:
The worst bug user can have with this thing is that they'll have duplicate requests if they issued `u32::MAX + 1` at the same time, which means that winit has that amount of events in flight, which will likely not really our issue anymore at this point.

notgull (approved):
Fair enough, then

daxpedda:
> @daxpedda is it something relevant for web as well or web is a bit special here?

No, luckily Wasm supports `AtomicU64` just fine.

PR #3460, opened 2024-02-07
Account for WAYLAND_SOCKET when detecting Wayland

### LINKED ISSUE #3459: Wayland availability detection does not check for WAYLAND_SOCKET
Wayland clients have two methods of connecting to the compositor. If the environment variable WAYLAND_DISPLAY is set, they can connect() to the Unix socket at a path indicated by WAYLAND_DISPLAY. A much rarer case is when the environment variable  WAYLAND_SOCKET is set; then the compositor will have provided one end of a socket pair through file descriptor inheritance, and  the value of WAYLAND_SOCKET will indicate the file descriptor ID.

Currently, `winit` when autodetecting which backend to use only checks for WAYLAND_DISPLAY, not WAYLAND_SOCKET. (The code that does this check is in [src/platform_impl/linux/mod.rs](https://github.com/rust-windowing/winit/blob/4d4d6e5052877d9578bac73f52eaf323e9089998/src/platform_impl/linux/mod.rs#L752C13-L754C35) .) It should consider the Wayland backend to be usable if either of WAYLAND_DISPLAY or WAYLAND_SOCKET is available.

To test this, run a program using winit under a compositor or tool that provides a Wayland connection using WAYLAND_SOCKET. (The way I ran into this issue was with the command [`waypipe`](https://gitlab.freedesktop.org/mstoeckl/waypipe/)` --oneshot ssh localhost `[`alacritty`](https://github.com/alacritty/alacritty).)

kchibisov:
Fixes #3459.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ Unreleased` header.
 - On Wayland, fix `Focused(false)` being send when other seats still have window focused.
 - On Wayland, fix `Window::set_{min,max}_inner_size` not always applied.
 - On Windows, fix inconsistent resizing behavior with multi-monitor setups when repositioning outside the event loop.
+- On Wayland, fix `WAYLAND_SOCKET` not used when detecting platform.
 
 # 0.29.10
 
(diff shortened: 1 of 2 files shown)


notgull (approved):
Looks good to me!

PR #3461, opened 2024-02-07
orbital: improve keyboard handling

jackpot51:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,8 @@ Unreleased` header.
 - On Wayland, fix `Window::set_{min,max}_inner_size` not always applied.
 - On Windows, fix inconsistent resizing behavior with multi-monitor setups when repositioning outside the event loop.
 - On Wayland, fix `WAYLAND_SOCKET` not used when detecting platform.
+- On Orbital, fix `logical_key` and `text` not reported in `KeyEvent`.
+- On Orbital, implement `KeyEventExtModifierSupplement`.
 
 # 0.29.10
 
--- a/src/platform_impl/orbital/event_loop.rs
+++ b/src/platform_impl/orbital/event_loop.rs
@@ -12,6 +12,7 @@ use orbclient::{
     ButtonEvent, EventOption, FocusEvent, HoverEvent, KeyEvent, MouseEvent, MoveEvent, QuitEvent,
     ResizeEvent, ScrollEvent, TextInputEvent,
 };
+use smol_str::SmolStr;
 
 use crate::{
     error::EventLoopError,
@@ -154,6 +155,21 @@ struct EventState {
 }
 
 impl EventState {
+    fn character_all_modifiers(&self, character: char) -> char {
+        // Modify character if Ctrl is pressed
+        if self.keyboard.contains(KeyboardModifierState::LCTRL)
+            || self.keyboard.contains(KeyboardModifierState::RCTRL)
+        {
+            if ('a'..='z').contains(&character) {
+                return ((character as u8 - b'a') + 1) as char;
+            }
+            //TODO: more control key variants?
+        }
+
+        // Return character as-is if no special handling required
+        character
+    }
+
     fn key(&mut self, key: PhysicalKey, pressed: bool) {
         let code = match key {
             PhysicalKey::Code(code) => code,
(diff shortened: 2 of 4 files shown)


src/platform_impl/orbital/event_loop.rs
```diff
@@ -333,39 +333,50 @@ impl<T: 'static> EventLoop<T> {
...
+                } else {
+                    let mut character_buf = [0u8; 4];
+                    let character_str = character.encode_utf8(&mut character_buf);
+                    (
+                        Key::Character(character_str.into()),
+                        if pressed {
+                            Some(character_str.into())
+                        } else {
+                            None
                         },
```
kchibisov, line 352: Extract to `let` and convert to `into` before the tuple.

src/platform_impl/orbital/event_loop.rs
```diff
@@ -333,39 +333,50 @@ impl<T: 'static> EventLoop<T> {
...
+                    event: event::WindowEvent::KeyboardInput {
+                        device_id: event::DeviceId(DeviceId),
+                        event: event::KeyEvent {
+                            logical_key,
+                            physical_key,
+                            location: KeyLocation::Standard,
+                            state: element_state(pressed),
+                            repeat: false,
+                            text,
+                            platform_specific: KeyEventExtra {},
```
kchibisov, line 368: if it's not implemented you can `Default::default`?

src/platform_impl/orbital/mod.rs
```diff
@@ -261,3 +266,17 @@ impl VideoMode {
...
+        //TODO: support modifiers changing text on orbital
+        self.text.as_deref()
+    }
+
+    #[inline]
+    fn key_without_modifiers(&self) -> Key {
+        //TODO: support modifiers changing key on orbital
+        self.logical_key.clone()
+    }
+}
```
kchibisov, line 282: If it's the same there's no need to implement it.

kchibisov (changes requested):
You should also add a CHANGELOG message and explicitly specify what actually got improved.

src/platform_impl/orbital/mod.rs
jackpot51: This is for users of that trait (iced) that require it to be implemented. Since Redox doesn't supply different text when control is pressed it mostly does the right thing to provide the original event text and key.

jackpot51:
I'm not sure why this got closed, because I didn't do it.

jackpot51:
I updated the changelog.

src/platform_impl/orbital/event_loop.rs
jackpot51: Perhaps, unless it is considered a requirement to support text_with_all_modifiers and key_without_modifiers correctly before this is merged. Then, I will use this to pass those around like the Linux implementation does.

src/platform_impl/orbital/mod.rs
kchibisov: How `shift` affects the key? For example `shift+4` should be `$`, however key without modifier is `4`, thus to have some sort of platform parity it should at least emit lowercase latter and handle such cases correctly? 

Or the input is not there yet, so it doesn't really matter?

kchibisov: Besides, iced should have a fallback path, since generally you don't have such information, on ios/android, etc. Though, maybe working there is not a priority for them.

jackpot51: I will take some time to implement it. Iced doesn't care about Android or IOS at this time.

kchibisov: I mean, the whole point of key_without_modifier is that you may want original unshifted key and you design your bindings around those keys, so if redox won't do that, it's literally not different than having a `cfg` and asking for the right key.

I'm not completely against having such impls for redox, because the platform is _special_, however it'll likely yield not really working result.

Key in mind that `ctrl+c` stuff with `text_with_all_modifiers` should likely generate proper C0 codes.

src/platform_impl/orbital/event_loop.rs
jackpot51: I'm now using this like the Linux platform does.

src/platform_impl/orbital/mod.rs
jackpot51: I've now implemented these for orbital.

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ Unreleased` header.
 
 # Unreleased
 
+- On orbital, report `logical_key` and `text` in `KeyEvent`
+- On orbital, stub support for `KeyEventExtModifierSupplement`
```
kchibisov: It's not `stub` anymore I guess? If so just something like

```
On orbital, implement `KeyEventExtModifierSupplement`.
```

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ Unreleased` header.
 
 # Unreleased
 
+- On orbital, report `logical_key` and `text` in `KeyEvent`
```
```suggestion
- On orbital, fix `logical_key` and `text` not reported in `KeyEvent`.
```

kchibisov (changes requested):
I've just noticed. You should open your patch against the `master`, not against the release branch.


--

I guess one day orbital will just use libxkbcommon or something?

src/platform_impl/orbital/mod.rs
```diff
@@ -1,10 +1,16 @@
 #![cfg(target_os = "redox")]
 
+use smol_str::SmolStr;
```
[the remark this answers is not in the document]
kchibisov: bump.

CHANGELOG.md
```diff
@@ -49,6 +49,8 @@ Unreleased` header.
 - On Wayland, fix `Window::set_{min,max}_inner_size` not always applied.
 - On Windows, fix inconsistent resizing behavior with multi-monitor setups when repositioning outside the event loop.
 - On Wayland, fix `WAYLAND_SOCKET` not used when detecting platform.
+- On orbital, fix `logical_key` and `text` not reported in `KeyEvent`
+- On orbital, implement `KeyEventExtModifierSupplement`
```
kchibisov, line 53: Just so they are consistent with all the other entries.

```suggestion
- On Orbital, fix `logical_key` and `text` not reported in `KeyEvent`.
- On Orbital, implement `KeyEventExtModifierSupplement`.
```

kchibisov (dismissed):

CHANGELOG.md
kchibisov: Please, implement the suggestions the way they were suggested or _consistent formatting_ or apply them directly with the github UI if you can't be bothered with all of that.

jackpot51:
Updated to fix clippy lint from https://github.com/rust-windowing/winit/actions/runs/7821504741/job/21338589355?pr=3461

kchibisov (approved):

PR #3471, opened 2024-02-09
On X11, cleanup event processor implementation

kchibisov:

--

I basically moved the code around, both commits should build and show what I've tried to achieve.

### DIFF
--- a/src/platform_impl/linux/x11/event_processor.rs
+++ b/src/platform_impl/linux/x11/event_processor.rs
@@ -1,147 +1,65 @@
-use std::{
-    cell::{Cell, RefCell},
-    collections::HashMap,
-    os::raw::{c_char, c_int, c_long, c_ulong},
-    rc::Rc,
-    slice,
-    sync::{Arc, Mutex},
-};
+use std::cell::{Cell, RefCell};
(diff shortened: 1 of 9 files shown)


src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -1503,6 +333,1225 @@ impl EventProcessor {
...
+        if xev.data.get_long(0) as xproto::Atom == wt.wm_delete_window {
+            let event = Event::WindowEvent {
+                window_id,
+                event: WindowEvent::CloseRequested,
+            };
+            callback(&self.target, event);
+            return;
+        }
+
+        if xev.data.get_long(0) as xproto::Atom == wt.net_wm_ping {
```
notgull, line 364: Nit: We should use `else if` instead of `if`.

notgull (approved):
Looks good to me! Seems like we've mostly moved code around.

src/platform_impl/linux/x11/event_processor.rs
kchibisov: it just was insanely long else if. in general i'd probably rewrite this to something else, but not right now.

PR #3478, opened 2024-02-09
Move some of our documentation to docs.rs

madsmtm:
Part of https://github.com/rust-windowing/winit/issues/3431.

Fixes https://github.com/rust-windowing/winit/issues/3440.

Issue #3431, opened 2024-01-27
Consolidate documentation on docs.rs


madsmtm:
There is [no clear guidelines in the Rust community](https://github.com/rust-lang/api-guidelines/discussions/254) on what goes in the README, and what goes on docs.rs. I read [Ed Page's blog post on `winnow`](https://epage.github.io/blog/2023/02/winnow-toml-edit-combine-nom/#usage) a while back, and the idea of moving everything to docs.rs really resonates with me. I'll copy the core points below.

> When exploring documentation options in clap, we found:
>
> - A lot of users want one place to go
> - Github is a poor place to store documentation because people naturally look at main and not the version pertinent to them

So a proposal: Let's move the scattered docs we do have from `README.md` and `FEATURES.md` to docs.rs.

madsmtm:
Related proposal, since we're cleaning things up anyhow: Remove the [Testers and Contributors table](https://github.com/rust-windowing/winit/wiki/Testers-and-Contributors) (and disable the wiki), the table is outdated, and I've at least personally never used it myself (though of course it's basically empty for the backends I maintain, so I guess I wouldn't really have a say here).

daxpedda:
I agree that the Wiki should be removed.

PR #3479, opened 2024-02-09
Remove feature descriptions in `FEATURES.md`

madsmtm:
Builds upon https://github.com/rust-windowing/winit/pull/3478. Related to https://github.com/rust-windowing/winit/issues/3431.

I went through all of the text in `FEATURES.md`, and cross-referenced it with our code, and as I suspected, the information in the code is much more detailed and up to date. So I propose we remove most of the contents in `FEATURES.md`, as it is difficult to keep up to date, and I don't really see the value provided.

Only thing I'm unsure about is the following three points that are yet unimplemented, and as such doesn't have a clear reference in our code:

> - **Popup / modal windows**: Windows can be created relative to the client area of other windows, and parent windows can be disabled in favor of popup windows. This feature also guarantees that popup windows get drawn above their owner.
> - **Gamepad/Joystick events**: Capturing input from gamepads and joysticks.
> - **Device movement events**: Capturing input from the device gyroscope and accelerometer.


daxpedda (commented):
I think this is the right way to go.
Will take a closer look when its ready.

kchibisov:
bump

FEATURES.md
daxpedda, line 1: I feel like this file is now quite small and its contents should either live in `README.md`, `CONTRIBUTING.md` or be removed entirely.

daxpedda (approved):
LGTM.

My small nit can be addressed in a follow-up.

daxpedda:
@madsmtm do you still want to move on with this?

notgull (approved):

PR #3480, opened 2024-02-09
On X11, don't require XSETTINGS

kchibisov:
We could fail to setup property watcher and fail to start, thus don't require XSETTINGS to work.

Fixes: df8805c0 (On X11, reload DPI on _XSETTINGS_SETTINGS)


notgull (approved):
Looks good to me!

PR #3481, opened 2024-02-09
Add more specific error `RequestIgnored`

madsmtm:
Split out from https://github.com/rust-windowing/winit/pull/3067.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -32,6 +32,8 @@
 //!
 //! [`EventLoop::run(...)`]: crate::event_loop::EventLoop::run
 //! [`ControlFlow::WaitUntil`]: crate::event_loop::ControlFlow::WaitUntil
+use std::error::Error;
+use std::fmt;
 use std::path::PathBuf;
 use std::sync::{Mutex, Weak};
 #[cfg(not(web_platform))]
@@ -43,7 +45,6 @@ use smol_str::SmolStr;
 #[cfg(web_platform)]
 use web_time::Instant;
 
-use crate::error::ExternalError;
 #[cfg(doc)]
 use crate::window::Window;
 use crate::{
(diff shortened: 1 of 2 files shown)


src/event.rs
```diff
@@ -1146,6 +1154,22 @@ impl PartialEq for InnerSizeWriter {
     }
 }
 
+/// The request to change the inner size synchronously was ignored.
+///
+/// See [`InnerSizeWriter::request_inner_size`] for details.
+#[derive(Debug, Clone)] // Explicitly not other traits, in case we want to extend it in the future
+pub struct RequestIgnored {
```
madsmtm, line 1161: @kchibisov asked in https://github.com/rust-windowing/winit/pull/3067#discussion_r1318159213:

> What's the reason this error is here and not on the `window` for example?

Because it's specific to `InnerSizeWriter`, which is also in this module.

src/event.rs
```diff
@@ -1146,6 +1154,22 @@ impl PartialEq for InnerSizeWriter {
...
+    _priv: (),
+}
+
+impl fmt::Display for RequestIgnored {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+        f.write_str("the request was ignored")
```
daxpedda, line 1167: Could we put the reason here as well, instead of just the documentation?

daxpedda (approved):

kchibisov (changes requested):
It's not really specific to this particular situation, sometimes it happens on X11 when setting something, e.g. locking.

In general we should have something like
```
error WindowError {
    NotSupported,
    RequestIgnored,
    // etc...
}
```

We can leave it for 0.31, given that the a lot of weird hack like this one will be gone and it'll be more clear what to put into errors, once folks look into what their backends can do.

PR #3484, opened 2024-02-10
Orbital buildout

jackpot51:
There are quite a few stubs on Orbital that can be implemented. These were tested in Redox alongside some additions to Orbital to support them.

### DIFF
--- a/src/platform_impl/orbital/window.rs
+++ b/src/platform_impl/orbital/window.rs
@@ -13,7 +13,8 @@ use crate::{
 };
 
 use super::{
-    EventLoopWindowTarget, MonitorHandle, RedoxSocket, TimeSocket, WindowId, WindowProperties,
+    EventLoopWindowTarget, MonitorHandle, OsError, RedoxSocket, TimeSocket, WindowId,
+    WindowProperties,
 };
 
 // These values match the values uses in the `window_new` function in orbital:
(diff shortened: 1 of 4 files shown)


jackpot51:
Tested and ready for review.

src/platform_impl/orbital/window.rs
```diff
@@ -359,30 +414,58 @@ impl Window {
...
+            window::CursorGrabMode::None => (false, false),
+            window::CursorGrabMode::Confined => (true, false),
+            window::CursorGrabMode::Locked => (true, true),
+        };
+        self.window_socket
+            .write(format!("M,G,{}", if grab { 1 } else { 0 }).as_bytes())
+            .map_err(|err| error::ExternalError::Os(os_error!(OsError::new(err))))?;
+        self.window_socket
+            .write(format!("M,R,{}", if relative { 1 } else { 0 }).as_bytes())
+            .map_err(|err| error::ExternalError::Os(os_error!(OsError::new(err))))?;
```
kchibisov, line 432: What relative means here? Locked is the cursor that stays at the same place and not moved. Confined is the one that clamps to window boundaries. The same as in pointer contraints protocol on wayland.

src/platform_impl/orbital/window.rs
```diff
@@ -298,11 +333,15 @@ impl Window {
     }
 
     #[inline]
-    pub fn set_maximized(&self, _maximized: bool) {}
+    pub fn set_maximized(&self, maximized: bool) {
+        self.set_flag(ORBITAL_FLAG_MAXIMIZED, maximized)
+            .expect("failed to set maximized")
```
kchibisov, line 339: I'd suggest to ignore, we shouldn't crash due to any of that.

src/platform_impl/orbital/window.rs
```diff
@@ -298,11 +333,15 @@ impl Window {
...
+    }
 
     #[inline]
     pub fn is_maximized(&self) -> bool {
-        false
+        self.get_flag(ORBITAL_FLAG_MAXIMIZED)
+            .expect("failed to get maximized")
```
kchibisov, line 345: Can default to `false` if you can't get anything for whatever reason.

kchibisov (changes requested):
That's _too much_ except for innocent requests, though I guess the reason they are like that is because you can fail because socket got closed?

src/platform_impl/orbital/window.rs
jackpot51: Relative means that the cursor will not move and relative mouse events will be sent instead of absolute ones.

jackpot51: I am converting back to a draft because the moue relative event is not handled yet, so I want to handle that properly.

kchibisov: aye, but you don't need _both_ working at the same time, you can just tell that something is not supported.

jackpot51: Latest commit emits `DeviceEvent::MouseMotion` for relative mouse events, so this will work properly.

kchibisov (approved):

Issue #3388, opened 2024-01-13
Ctrl sometimes gets "stuck" on X11


fredizzimo:
This was reported in

https://github.com/neovide/neovide/issues/2228

The [neovide_rCURRENT.log](https://github.com/neovide/neovide/files/13800892/neovide_rCURRENT.log) is huge, by contains all key presses reported by Winit, and the modifier states. Search for `KeyEvent` and `ModifiersState` or filter by `[neovide::window::keyboard_manager]`

The key up event is received, but the modifier state is not updated. The modifiers are reported strange in other ways as well, for example shortly after, when super is pressed it alternates between super and no modifiers state.

I suspect this is some kind of timing issue, since it only happens on certain file types, and Neovide does not have any file type specific handling.

kchibisov:
We pass modifiers from the xkb event. So there's either event or no event, but could be logic to send modifier lazily not work.

fredizzimo:
@kchibisov, we just got a report of a very similar issue, but it was reported to happen on Wayland as well and we don't have any logs yet, so it might be a different but similar issue.

* https://github.com/neovide/neovide/issues/2358

But is it possible that this could affect Wayland too?

kchibisov:
No, wayland having thing like that is not our bug.

fredizzimo:
Thank you, I asked for that information as well. I will open a new issue if I can determine that it's a winit issue, and otherwise just resolve the issue in Neovide.

Issue #3085, opened 2023-09-04
Discussion: Start committing our lockfile


madsmtm:
In [a recent blog post](https://blog.rust-lang.org/2023/08/29/committing-lockfiles.html), the Cargo Team outlined some reasons why they've changed the default for `cargo new` to not exclude the lockfile from version control.

I think `winit` should follow that direction, and start committing our `Cargo.lock`.

This would help greatly with our CI setup, both simplifying it in terms of caching, as well as making it break much less often due to changes in our dependencies.

On the other hand, perhaps it's a good thing that our CI breaks on changes in dependencies, it helps them discover unintended changes - hence why this is a discussion.

daxpedda:
I think Rust has good guidelines on why to or not to commit `Cargo.lock`. https://doc.rust-lang.org/nightly/cargo/faq.html#why-have-cargolock-in-version-control.

My opinion on this is that it's only useful for helping debug issues during development, but ultimately we should make sure that we always support the newest dependency versions as that is what users will get when they use `winit`.

Though I don't see a particular upside for this in Winit, I'm strongly against committing the `Cargo.lock` file unless we also commit to using Dependabot (or anything similar) and continuously update it. Otherwise we are just not testing what users also experience. In addition I would prefer, like mentioned in the blog post, to also use `cargo update` in the CI (in addition to testing without `cargo update`), but that would break the motivation @madsmtm lined out above.

kchibisov:
We could use a lock file only for CI, but not actually publish it, there's nothing wrong with doing something like that. I also like that they talk about keeping msrv compat, yet we have `cargo install` ignoring `Cargo.lock` by default...

Also, I think Cargo.lock is actually ignored from the deps.

notgull:
Yes, I'm fine with committing `Cargo.lock`.

daxpedda:
@kchibisov

> It's a default behavior for libraries now even.

The default that has changed is that its not ignored in `.gitignore`, the default is not to just commit it. See [the Rust blog post about this change](https://blog.rust-lang.org/2023/08/29/committing-lockfiles.html):
> Since there isn't a universal answer to these situations, we felt it was best to leave the choice to developers and give them information they need in making a decision.

> It only solves issues, like having _working bisect_.

I'm not against committing `Cargo.lock`, I'm saying we should **also** check latest versions.
[...]

PR #3486, opened 2024-02-11
On X11, use events modifiers to detect state

kchibisov:
While there's a separate event to deliver modifiers for keyboard, unfortunately, it's not even remotely reflects the modifiers state.

Thus use events along side regular modifier updates to correctly detect the state. Also, apply the modifiers from the regular key event by converting their state to xkb modifiers state.

Links: https://github.com/alacritty/alacritty/issues/7549
Closes: #3388


kchibisov:
@daxpedda @madsmtm @notgull do you have a suggestion what to do with the CI?

daxpedda:
This happens because of missing MSRV support in Cargo dependency selection: https://github.com/rust-lang/cargo/issues/9930.

One way to go here is just to fix the dependency version in the CI with `cargo update -p ahash --precise version`.
Alternatively we can also follow [Cargo's recommendation](https://doc.rust-lang.org/cargo/guide/continuous-integration.html#verifying-rust-version) and use [`cargo-hack`](https://crates.io/crates/cargo-hack).

kchibisov:
Can't we have a `Cargo.lock` with precise versions? And then just exclude it from publishing.

daxpedda:
> Can't we have a `Cargo.lock` with precise versions? And then just exclude it from publishing.

I'm very strongly against committing a `Cargo.lock` file **unless** the issues I mentioned in https://github.com/rust-windowing/winit/issues/3085 would be addressed as well.

kchibisov:
>I'm very strongly against committing a Cargo.lock file unless the issues I mentioned in https://github.com/rust-windowing/winit/issues/3085 would be addressed as well.

It's a default behavior for libraries now even. But I don't see what issue it creates and why you need dependabot for that. It only solves issues, like having _working bisect_. 


It's also worth saying that I'd rather test min version than latest deps, since if semver breaks it's not our issue, and if winit can't work correctly without version `0.8.N+1` we should force it in `Cargo.toml`, otherwise you have broken build which you don't notice because you pull latest and things magically work.

kchibisov:
I've used the `--precise` for now.

src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -1527,42 +1548,148 @@ impl EventProcessor {
...
+        }
+        if let Some(mod3) = mods_indices.mod3.filter(|_| ModMask::M3.intersects(state)) {
+            depressed |= 1 << mod3;
+        }
+        if let Some(logo) = mods_indices.logo.filter(|_| ModMask::M4.intersects(state)) {
+            depressed |= 1 << logo;
+        }
+        if let Some(mod5) = mods_indices.mod5.filter(|_| ModMask::M5.intersects(state)) {
+            depressed |= 1 << mod5;
+        }
```
notgull, line 1688: This code is repetitive and could be deduplicated with a macro

notgull (approved):
Minor quibbles but I'd rather get this merged and fix the quibbles later.

PR #3490, opened 2024-02-12
On X11, don't require XIM to be present

kchibisov:
In general, we may want to use xinput v2 for keyboard input in such cases, so we have compose going, but for now just don't crash if there's no XIM.

--

We can probably wire in such case the xinput v2 key input, thus will have the working dead keys, however I'd fix the crash first.

### DIFF
--- a/src/platform_impl/linux/x11/event_processor.rs
+++ b/src/platform_impl/linux/x11/event_processor.rs
@@ -74,7 +74,11 @@ impl EventProcessor {
 
         // Handle IME requests.
         while let Ok(request) = self.ime_receiver.try_recv() {
-            let ime = window_target.ime.get_mut();
+            let ime = match window_target.ime.as_mut() {
+                Some(ime) => ime,
+                None => continue,
+            };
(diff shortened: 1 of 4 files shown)


src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -74,7 +74,11 @@ impl EventProcessor {
 
         // Handle IME requests.
         while let Ok(request) = self.ime_receiver.try_recv() {
-            let ime = window_target.ime.get_mut();
+            let ime = match window_target.ime.as_mut() {
+                Some(ime) => ime,
+                None => continue,
```
notgull, line 79: Nit: I think `break` would be more appropriate here.

notgull (approved):
Looks good to me!

src/platform_impl/linux/x11/event_processor.rs
kchibisov: We should drain until the error to not increase the queue over time.

PR #3499, opened 2024-02-18
feat: `Event::Reopen` support for macOS

I-Info:

## Description

This event is typically used to handle dock icon clicking event on macOS and  uses [`applicationShouldHandleReopen:hasVisibleWindows:`] callback from AppKit Framework. 

## Use case

It will make it possible to keep application running in backgound when all windows are closed, and create new window after dock icon clicked (application reopen callback).

[`applicationShouldHandleReopen:hasVisibleWindows:`]: https://developer.apple.com/documentation/appkit/nsapplicationdelegate/1428638-applicationshouldhandlereopen


src/event.rs
```diff
@@ -253,6 +253,25 @@ pub enum Event<T: 'static> {
...
+    ///
+    /// ## Platform-specific
+    ///
+    /// ### macOS
+    ///
+    /// On macOS, the `Reopen` event is emitted in response to an [`applicationShouldHandleReopen`]
+    /// callback, which is typically called when the user clicks on the application's icon in the Dock.
+    /// The application should create a new window if it doesn't have any, or bring the
+    /// existing windows to the front. `has_visible_windows` is `true` if the application has
+    /// at least one visible window.
```
madsmtm, line 267: Maybe document it more in Winit terms?

```suggestion
    /// Usually, the user would expect you to create a new window if there isn't any.
```

src/event.rs
```diff
@@ -253,6 +253,25 @@ pub enum Event<T: 'static> {
...
+    ///
+    /// [`applicationShouldHandleReopen`]: https://developer.apple.com/documentation/appkit/nsapplicationdelegate/1428638-applicationshouldhandlereopen
+    ///
+    /// ### Others
+    ///
+    /// - **Android / iOS / Web / Wayland / Windows / Orbital:** Unsupported.
+    Reopen { has_visible_windows: bool },
```
madsmtm, line 274: The user already knows how many windows they have, I don't really see the purpose of this flag?

```suggestion
    Reopen,
```

src/event.rs
```diff
@@ -253,6 +253,25 @@ pub enum Event<T: 'static> {
...
     MemoryWarning,
+
+    /// Emitted when the application has received a reopen request from OS.
+    ///
+    /// ## Platform-specific
+    ///
+    /// ### macOS
+    ///
+    /// On macOS, the `Reopen` event is emitted in response to an [`applicationShouldHandleReopen`]
+    /// callback, which is typically called when the user clicks on the application's icon in the Dock.
```
madsmtm, line 264: Copied more precise text from the docs.

```suggestion
    /// callback, which is usually called whenever the Finder reactivates an already running
    /// application because the user double-clicked it again or used the dock to activate it.
```

examples/reopen.rs
```diff
@@ -0,0 +1,61 @@
+#![allow(clippy::single_match)]
```
madsmtm, line 1: I think I'll wait 'till after https://github.com/rust-windowing/winit/pull/3447, as that reworks how our examples are handled.

madsmtm (changes requested):
I can see this being useful, thanks!

examples/reopen.rs
```diff
@@ -0,0 +1,57 @@
+// TODO: merge examples 
+// https://github.com/rust-windowing/winit/pull/3447
```
I-Info, line 2: - [ ] Merge this example after #3447

kchibisov (changes requested):
nack for 0.30, should be good for 0.31 with traits.

madsmtm (approved):
> nack for 0.30, should be good for 0.31 with traits.

I'm not going to block this on 0.31, it's a simple change that we can do now, and generally enables better behaviour on macOS.

kchibisov:
And what should we do with all of that on other platforms? The event is macOS specific, but it bleeds into other backend and people will demand its handling, while in reality such event doesn't exist for them.

You should never trade quality for short term wins, this is just straight reduces overall quality because the event is not tied to window in particular.

madsmtm:
> And what should we do with all of that on other platforms? The event is macOS specific, but it bleeds into other backend and people will demand its handling, while in reality such event doesn't exist for them.

I think the event is well documented as "Unsupported", which we already have precedent for, I don't see the issue with adding another one.

In the _end_ it won't really matter, since we're going to be changing things for 0.31 anyways. But like I said, in the months until that happens, this is a useful feature.

daxpedda (approved):
I nominated this specifically because I wanted to discuss this "blocking it because its platform-specific" issue in general, but we didn't get to it unfortunately.

Personally I strongly disagree that this somehow reduces the quality of the Winit API, we already have a big amount of events that are not cross-platform, this doesn't change anything.

That said, I'm glad we are trying to fix/improve this issue in v0.31.

kchibisov:
All `Event` stuff is cross platform and has special semantics. `WindowEvent` stuff has specifics, but not the `Event`.

As I said, the solution is already here even for 0.30, if you want I can write it myself for macOS.

madsmtm:
Instead of exposing this through Winit, https://github.com/rust-windowing/winit/pull/3758 is going to allow listening to `NSApplicationDelegate` events outside of Winit using `objc2::declare_class!`, Objective-C, Swift, or similar.

Issue #3500, opened 2024-02-18
X11: Mouse motion events sometimes contain garbage


valaphee:
While printing out events on X11, I noticed that sometimes (not always), the motion events are invalid:
```
DeviceEvent { device_id: DeviceId(X(DeviceId(2))), event: Motion { axis: 0, value: -1.0 } }
DeviceEvent { device_id: DeviceId(X(DeviceId(2))), event: Motion { axis: 1, value: 7.1e-322 } }
DeviceEvent { device_id: DeviceId(X(DeviceId(2))), event: MouseMotion { delta: (-1.0, 7.1e-322) } }
```
Only one axis gets reported and the second axis is just an unrelated value (most likely smth not belonging to the event).
```
for i in 0..xev.valuators.mask_len * 8 {
    if !xinput2::XIMaskIsSet(mask, i) {
        continue;
    }
    let x = unsafe { *value };
```

The easiest way I found to reproduce this is just moving the cursor between two monitors. It might also just be a x11 problem.

I'm running natively latest Arch Linux with Gnome using X11.

notgull:
`7.1e-322` is really close to zero, which makes me think this is an error in interpreting the data from the X11 server.

Can you reproduce this issue while running `xtrace`, then attach the logs?

valaphee:
XIRawEvent raw_values are used instead of XIRawEvent valuators values: https://github.com/rust-windowing/winit/blob/e61a7320a277ecfdafd978ea0276df2f0c64aeda/src/platform_impl/linux/x11/event_processor.rs#L1418

Replacing raw values with values yields way higher precision, (raw values only indicate if a value is zero, positive or negative at least with a mouse device) and doesn't has this problem.

I'll try to get a trace.

PR #3501, opened 2024-02-18
X11: small tweaks for input handling

valaphee:
While trying out gamepad device events for X11 and Wayland, I noticed some issues:
- https://github.com/rust-windowing/winit/issues/3500
- deviceId is useless, as it always is either the value of the virtual keyboard or virtual mouse, even though the real deviceId is known and supplied via sourceid, which is only used for raw keyboard events.

This PRs goal is to address both issues.

### DIFF
--- a/src/platform_impl/linux/x11/event_processor.rs
+++ b/src/platform_impl/linux/x11/event_processor.rs
@@ -980,7 +980,7 @@ impl EventProcessor {
     {
         let wt = Self::window_target(&self.target);
         let window_id = mkwid(event.event as xproto::Window);
-        let device_id = mkdid(event.deviceid as xinput::DeviceId);
+        let device_id = mkdid(event.sourceid as xinput::DeviceId);
 
         // Set the timestamp.
         wt.xconn.set_timestamp(event.time as xproto::Timestamp);
@@ -1053,7 +1053,7 @@ impl EventProcessor {
         // Set the timestamp.
         wt.xconn.set_timestamp(event.time as xproto::Timestamp);
 
-        let device_id = mkdid(event.deviceid as xinput::DeviceId);
+        let device_id = mkdid(event.sourceid as xinput::DeviceId);
         let window = event.event as xproto::Window;
         let window_id = mkwid(window);
         let new_cursor_pos = (event.event_x, event.event_y);
@@ -1146,7 +1146,7 @@ impl EventProcessor {
 
         let window = event.event as xproto::Window;
         let window_id = mkwid(window);
-        let device_id = mkdid(event.deviceid as xinput::DeviceId);
+        let device_id = mkdid(event.sourceid as xinput::DeviceId);
 
         if let Some(all_info) = DeviceInfo::get(&wt.xconn, super::ALL_DEVICES.into()) {
             let mut devices = self.devices.borrow_mut();
@@ -1202,7 +1202,7 @@ impl EventProcessor {
             let event = Event::WindowEvent {
                 window_id: mkwid(window),
                 event: WindowEvent::CursorLeft {
-                    device_id: mkdid(event.deviceid as xinput::DeviceId),
+                    device_id: mkdid(event.sourceid as xinput::DeviceId),
                 },
             };
             callback(&self.target, event);
@@ -1265,7 +1265,7 @@ impl EventProcessor {
         let pointer_id = self
             .devices
             .borrow()
-            .get(&DeviceId(xev.deviceid as xinput::DeviceId))
+            .get(&DeviceId(xev.sourceid as xinput::DeviceId))
             .map(|device| device.attachment)
             .unwrap_or(2);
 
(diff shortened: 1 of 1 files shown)


kchibisov:
>Wayland, I noticed some issues:

I'm not sure you can get access to HID on wayland and open thing yourself, most just ship udev rules, but that's just wrong, there's a protocol in works for things like that, but that's about it.

On X11, I guess you pass xid and user needs special access for thing to work as well?

src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -980,7 +980,7 @@ impl EventProcessor {
     {
         let wt = Self::window_target(&self.target);
         let window_id = mkwid(event.event as xproto::Window);
-        let device_id = mkdid(event.deviceid as xinput::DeviceId);
+        let device_id = mkdid(event.sourceid as xinput::DeviceId);
```
notgull, line 983: For all of these usages it should be indicated in comments that `sourceid` is only supported since Xinput v2.1.

src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -1411,11 +1411,11 @@ impl EventProcessor {
         // Set the timestamp.
         wt.xconn.set_timestamp(xev.time as xproto::Timestamp);
 
-        let did = mkdid(xev.deviceid as xinput::DeviceId);
+        let device_id = mkdid(xev.sourceid as xinput::DeviceId);
 
         let mask =
             unsafe { slice::from_raw_parts(xev.valuators.mask, xev.valuators.mask_len as usize) };
-        let mut value = xev.raw_values;
+        let mut value = xev.valuators.values;
```
notgull, line 1418: What is the difference between `raw_values` and `valuators.values`? [This document](https://www.x.org/releases/X11R7.7/doc/inputproto/XI2proto.txt) says this:

> A RawEvent provides the information provided by the driver to the
client. RawEvent provides both the raw data as supplied by the driver and
transformed data as used in the server. Transformations include, but are
not limited to, axis clipping and acceleration.

I agree with the idea that we should be using these values instead of the raw driver values. However I'm not entirely sure what "axis clipping" is. Google doesn't have many results and all I see is [this release](https://answers.launchpad.net/ubuntu/+source/xorg-server/2:1.3.0.0.dfsg-12ubuntu3):

> 231_In_dmxBackendMouGetInfo_initialize_the_info-> minval_maxval_arrays_to_the_size_of_the_backend_display.patch:
      Fixes potential issue in X input where axis clipping code in
      GetPointerEvents() constrains the pointer's coordinate range to a
      max of 0, causing the mouse to not move.

Eventually I found [this commit](https://gitlab.freedesktop.org/xorg/xserver/-/commit/0aaf28e5633a59563b89a2e42d19fabc84adc3ed), which appears to be somehow clipping the mouse valuators to the monitor's size. So I guess that's what is happening here; the raw values don't keep the monitor border in mind. Hence why it gives us garbage every now and then.

src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -1448,15 +1448,15 @@ impl EventProcessor {
...
         if mouse_delta != (0.0, 0.0) {
             let event = Event::DeviceEvent {
-                device_id: did,
+                device_id,
                 event: DeviceEvent::MouseMotion { delta: mouse_delta },
             };
             callback(&self.target, event);
         }
 
         if scroll_delta != (0.0, 0.0) {
```
notgull, line 1457:
```suggestion
        if scroll_delta.0.abs() > 0.0001 || scroll_delta.1.abs() > 0.0001 {
```

Ditto.

notgull (changes requested):

src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -1053,7 +1053,7 @@ impl EventProcessor {
         // Set the timestamp.
         wt.xconn.set_timestamp(event.time as xproto::Timestamp);
 
-        let device_id = mkdid(event.deviceid as xinput::DeviceId);
+        let device_id = mkdid(event.sourceid as xinput::DeviceId);
```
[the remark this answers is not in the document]
valaphee: Basically figured this out by just connecting some mice. For raw key events sourceid was always used (at least the last 5 years) https://github.com/rust-windowing/winit/blob/master/src/platform_impl/linux/x11/event_processor.rs#L1481 for some reason.

I'll add some comments, describing why deviceid is not used.

src/platform_impl/linux/x11/event_processor.rs
valaphee: I don't know if its actually the raw input, as my pointing device reports different values when looking at the raw hid reports. raw_values just seem to be either -1.0, 0.0 and 1.0, it doesn't make a difference how fast its moving, which is not usable.

But its also not ideal if it is affected by acceleration, as that's the main reason for using raw input.

valaphee:
> > Wayland, I noticed some issues:
> 
> I'm not sure you can get access to HID on wayland and open thing yourself, most just ship udev rules, but that's just wrong, there's a protocol in works for things like that, but that's about it.
> 
> On X11, I guess you pass xid and user needs special access for thing to work as well?

Gilrs is using udev for both Wayland and X11. But theoretically it could also be done though XI2 https://www.x.org/releases/current/doc/man/man4/joystick.4.xhtml.

valaphee:
Switched raw mouse input it back to using raw_values, as otherwise it would be useless, and https://github.com/rust-windowing/winit/issues/3500 should not occur in games, where its normally used, at least when they don't utilize multiple screens on X11.

PR #3502, opened 2024-02-19
fix: set current size in Canvas::create based on inner_size attribute

nicksenger:
This PR updates `Canvas::create` to set the `current_size` based on `attr.inner_size` if present. This allows the window size to be used immediately on web.

- [x ] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Unreleased` header.
 
 # Unreleased
 
+- Set current size when creating canvas with inner size
 - On X11, don't require XIM to run.
 - On X11, fix xkb state not being updated correctly sometimes leading to wrong input.
 - Fix compatibility with 32-bit platforms without 64-bit atomics.
--- a/src/platform_impl/web/web_sys/canvas.rs
+++ b/src/platform_impl/web/web_sys/canvas.rs
@@ -120,6 +120,10 @@ impl Canvas {
         };
 
         if let Some(size) = attr.inner_size {
+            common
+                .current_size
+                .set(size.to_physical(super::scale_factor(&common.window)));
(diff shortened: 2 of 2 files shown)


daxpedda:
The reason why the size isn't immediately available is because there is no way to currently retrieve the actual size of the canvas in pixel synchronously.

For context:
We currently use [`ResizeObserver`s `devicePixelContentBoxSize`](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserverEntry/devicePixelContentBoxSize#browser_compatibility) to get the actual size of the canvas in pixel. AFAIK this is the only Web API that can do this accurately.

Unfortunately this API is event-based, so we receive the size of the canvas in an event, so until we receive this event we don't know the actual size of the canvas.

We have decided that this is currently the best course of action unless we want to introduce a larger API surface which can tell the user if the size returned by `Window::inner_size()` is currently accurate or not, which we feel is a contra-productive direction to go in.

You can see the history and research on this here: https://github.com/rust-windowing/winit/pull/2859.
The solution to this problem we currently have in mind is here: https://github.com/rust-windowing/winit/issues/2863.

---

Your current implementation changes the size reported by `Window::inner_size()` at the start to an inaccurate one, which could lead to weird behavior as seen in [this demo](https://device-pixel-content-box.glitch.me).

With that in mind I will close the PR.

Please let me know if you have any suggestions on how to tackle this problem other then https://github.com/rust-windowing/winit/issues/2863!

nicksenger:
Thanks a lot for the clarification! I figured this was just a weird corner-case, but seems like its been given a fair bit of thought already.

PR #3507, opened 2024-02-21
On X11, force resend modifiers when focus changes

kchibisov:
Given that `ModifiersChanged` is a window event, it means that clients may track it for each window individually, thus not sending it between focus changes may result in modifiers getting desynced on the consumer side.

### DIFF
--- a/src/platform_impl/linux/x11/event_processor.rs
+++ b/src/platform_impl/linux/x11/event_processor.rs
@@ -11,9 +11,10 @@ use x11_dl::xinput2::{
 use x11_dl::xlib::{
     self, Display as XDisplay, Window as XWindow, XAnyEvent, XClientMessageEvent, XConfigureEvent,
     XDestroyWindowEvent, XEvent, XExposeEvent, XKeyEvent, XMapEvent, XPropertyEvent,
-    XReparentEvent, XSelectionEvent, XVisibilityEvent, XkbAnyEvent,
+    XReparentEvent, XSelectionEvent, XVisibilityEvent, XkbAnyEvent, XkbStateRec,
 };
 use x11rb::protocol::xinput;
+use x11rb::protocol::xkb::ID as XkbId;
 use x11rb::protocol::xproto::{self, ConnectionExt as _, ModMask};
 use x11rb::x11_utils::ExtensionInformation;
 use x11rb::x11_utils::Serialize;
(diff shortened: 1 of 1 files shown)


kchibisov:
The web msrv CI is unfixable, because it's not in the lockfile unless you start actually build docs for web.

cc @daxpedda

kchibisov:
Hm, so cargo just gives bogus error message when the crate is yanked. Forcing 3.14 worked.

src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -1692,25 +1772,21 @@ impl EventProcessor {
...
-
-        if self.modifiers.replace(modifiers) != modifiers {
-            callback(
-                &self.target,
-                Event::WindowEvent {
-                    window_id,
-                    event: WindowEvent::ModifiersChanged(self.modifiers.get().into()),
-                },
-            );
+        if self.modifiers.replace(modifiers) != modifiers || force {
```
notgull, line 1784: Nit: Put `force` first to help with branch prediction.

notgull (approved):
Fantastic as always!

src/platform_impl/linux/x11/event_processor.rs
kchibisov: No, it'll be wrong that way since `force and mods != old_mods` won't update the modifiers, maybe I should write it in a comment to make it more clear?

kchibisov: Updated with the comment explicitly pointing that out.

notgull (approved):
Alright, looks good to me

paulcacheux:
Hello @kchibisov this PR is breaking the build of winit on linux because:
- the dependency on x11-dl is on version 2.18.5 https://github.com/rust-windowing/winit/blob/96172693fe0a25c6bc0f0497c9d04429bd17d2cb/Cargo.toml#L190
- there is no `XkbStateRec` in v2.18.5 (only 2.19.0 and up) https://docs.rs/x11-dl/2.18.5/x11_dl/xlib/index.html

How could we fix this ? Does bumping the dependency requirement sounds good to you ? I can do a PR for that

kchibisov:
@paulcacheux `cargo update -p x11-dl` in your app, no need to patch winit for that.

To track min version stuff there's #3085

paulcacheux:
Isn't that an issue that winit is not compatible with the minimal version it requires in it's cargo.toml file ? Of course I can update on my app, but any other app that just installs winit will it this issue no ?

paulcacheux:
Basically my point is that everything was working well, dependabot [opens a PR to bump winit](https://github.com/paulcacheux/GameboyEmulator/pull/35) and it doesn't build anymore. It feels like a free breaking change for not much to win

kchibisov:
> It feels like a free breaking change for not much to win

It's not a breaking change even, that's not how it works. you can update x11-dl without us bumping, if your dependabot can't update x11-dl it's slacking at work.

Surely you can fix the version inside the `Cargo.toml`, but that's the only change I'll accept.

PR #3510, opened 2024-02-22
On X11, replay modifiers consumed by XIM

kchibisov:

### DIFF
--- a/src/platform_impl/linux/x11/event_processor.rs
+++ b/src/platform_impl/linux/x11/event_processor.rs
@@ -1,5 +1,5 @@
 use std::cell::{Cell, RefCell};
-use std::collections::HashMap;
+use std::collections::{HashMap, VecDeque};
 use std::os::raw::{c_char, c_int, c_long, c_ulong};
 use std::slice;
 use std::sync::{Arc, Mutex};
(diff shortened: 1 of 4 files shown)


kchibisov:
There're multiple approach to that, we can (in theory) check if the event is a modifier and handle it before `XIM` and then just discard it once it's arrives from XIM. But this can probably desync us, so the solution is just emit a bit more, though it shouldn't really matter, it's not like we can do better here.

Modifiers will basically go like `Shift(xkb) -> empty(core, replay) -> key(xim forwarded) -> Shift(restoring)`. Not doing so will result in e.g. `Shift + shift` bindings not working, since you can't detect such thing at all.

src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -41,6 +41,9 @@ use crate::platform_impl::x11::{
 /// The X11 documentation states: "Keycodes lie in the inclusive range `[8, 255]`".
 const KEYCODE_OFFSET: u8 = 8;
 
+/// The maximum amount of X modifiers to replay.
+const MAX_MOD_REPLAY_LEN: usize = 32;
```
notgull, line 45:
```suggestion
pub(super) const MAX_MOD_REPLAY_LEN: usize = 32;
```

src/platform_impl/linux/x11/mod.rs
```diff
@@ -322,6 +325,8 @@ impl<T: 'static> EventLoop<T> {
             ime_receiver,
             ime_event_receiver,
             xi2ext,
+            xfiltered_modifiers: Default::default(),
```
notgull, line 328:
```suggestion
            xfiltered_modifiers: std::collections::VecDeque::with_capacity(event_processor::MAX_MOD_REPLAY_LEN),
```

notgull (approved):
Minor nit

Issue #3513, opened 2024-02-23
Provide a clear distinction between the major version and the latest stable version of the documentation.


kacperpaczos:
Ideas:
* change doc to stabe doc
* inform that: The stable and development versions of the documentation are not compatible.

kacperpaczos:
![image](https://github.com/rust-windowing/winit/assets/23406555/ea2983fb-6d80-4f21-9ebc-86bfb2be89c1)

PR #3514, opened 2024-02-23
Provide a clear distinction between the major version and the latest …

kacperpaczos:
…stable version of the documentation
#3513 

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,6 @@
 # winit - Cross-platform window creation and management in Rust
 
 [![Crates.io](https://img.shields.io/crates/v/winit.svg)](https://crates.io/crates/winit)
-[![Docs.rs](https://docs.rs/winit/badge.svg)](https://docs.rs/winit)
-[![Master Docs](https://img.shields.io/github/actions/workflow/status/rust-windowing/winit/docs.yml?branch=master&label=master%20docs
-)](https://rust-windowing.github.io/winit/winit/index.html)
 [![CI Status](https://github.com/rust-windowing/winit/workflows/CI/badge.svg)](https://github.com/rust-windowing/winit/actions)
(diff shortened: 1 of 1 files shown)


kacperpaczos:
image

kchibisov (changes requested):
I'm not sure why change like that, it's already pretty clear given that one thing says `docs` and other says `master` docs.

Just moving it a bit won't do anything, so I don't see why the change needed.

madsmtm:
See also previous discussion in https://github.com/rust-windowing/winit/pull/3359.

I agree that we need to do something here, I am leaning more towards putting the master docs link somewhere without a badge.

README.md
```diff
@@ -1,17 +1,18 @@
...
 
 ```toml
 [dependencies]
 winit = "0.29.10"
 ```
 
-## [Documentation](https://docs.rs/winit)
+## Documentation
+
+[![Stable123Docs](https://docs.rs/winit/badge.svg)](https://docs.rs/winit)
```
daxpedda, line 13:
```suggestion
[![Stable Docs](https://img.shields.io/docsrs/winit?label=stable%20docs)](https://docs.rs/winit)
```

daxpedda (changes requested):
I think the badge to the stable docs should stay at the top of the README, this remains useful.

> I agree that we need to do something here, I am leaning more towards putting the master docs link somewhere without a badge, or maybe prefixing it with `UNSTABLE`, with an orange font?

I think removing the badge to the master documentation is a good idea, expanding the `## Documentation` header and adding the link there should work.

Alternatively, we could just put a warning like the one in the screenshot on top of the master docs (visible inside the documentation not in the README), which I've seen somewhere before ...

PR #3518, opened 2024-02-23
Split `dpi` module out into a separate crate

madsmtm:
Part of https://github.com/rust-windowing/winit/issues/3433.

This PR is mostly about the documentation update, the actual API changes that Tauri needs can be done in subsequent PRs.

CC @amrbashir

### DIFF
--- a/dpi/Cargo.toml
+++ b/dpi/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "dpi"
 version = "0.0.0"
-description = "Cross-platform window creation library."
+description = "Types for handling UI scaling"
 keywords = ["DPI", "HiDPI", "scale-factor"]
 categories = ["gui"]
 rust-version.workspace = true
--- a/dpi/Cargo.toml
+++ b/dpi/Cargo.toml
@@ -1,0 +1,39 @@
+[package]
+name = "dpi"
+version = "0.0.0"
+description = "Cross-platform window creation library."
+keywords = ["DPI", "HiDPI", "scale-factor"]
+categories = ["gui"]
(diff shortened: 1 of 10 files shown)


amrbashir (approved):
Thank you, this is exactly what I was looking for. I will do a small PR for the single unit types afterwards.

dpi/Cargo.toml
```diff
@@ -0,0 +1,39 @@
+[package]
+name = "dpi"
```
kchibisov, line 2: what's the point to name it dpi and not `winit-dpi`, since it's tailored towards the dpi needs of winit?

madsmtm: Mostly because the name `dpi` was free, but also because the crate isn't really _fundamentally_ tied to Winit.

dpi/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+version = "0.0.0"
+description = "Cross-platform window creation library."
```
Friz64, line 4: Needs updating.

src/lib.rs
```diff
@@ -177,7 +196,9 @@
 #[cfg(feature = "rwh_06")]
 pub use rwh_06 as raw_window_handle;
 
-pub mod dpi;
+// Re-export DPI types so that users don't have to put it in Cargo.toml.
+pub use dpi;
```
notgull, line 200:
```suggestion
#[doc(inline)]
pub use dpi;
```

For cohesion

notgull (approved):

notgull (approved):

daxpedda (approved):

PR #3520, opened 2024-02-24
Remove once_cell dependency

james7132:

A follow up to #2313. Removes the `once_cell` dependency, instead using `std::sync::OnceLock` and a minimal polyfill for `std::sync::LazyLock`, which may be stablized soon (see https://github.com/rust-lang/rust/pull/121377).

Should it be named `LazyLock` to match up with the proposed API? I kept it as `Lazy` for now.

This should not require a bump in MSRV, as `OnceLock` was stabilized in 1.70, which this crate is using.

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -2,31 +2,26 @@
 // Replace with std::sync::LazyLock when https://github.com/rust-lang/rust/issues/109736 is stablized.
 
 use std::ops::Deref;
-use std::sync::Mutex;
 use std::sync::OnceLock;
 
-pub(crate) struct Lazy<T, F = fn() -> T> {
+pub(crate) struct Lazy<T> {
     cell: OnceLock<T>,
-    init: Mutex<Option<F>>,
+    init: fn() -> T,
 }
 
-impl<T, F: FnOnce() -> T> Lazy<T, F> {
-    pub const fn new(f: F) -> Self {
+impl<T> Lazy<T> {
+    pub const fn new(f: fn() -> T) -> Self {
         Self {
             cell: OnceLock::new(),
-            init: Mutex::new(Some(f)),
+            init: f,
         }
     }
 }
 
-impl<T, F: FnOnce() -> T> Deref for Lazy<T, F> {
+impl<T> Deref for Lazy<T> {
     type Target = T;
     #[inline]
     fn deref(&self) -> &'_ T {
-        self.cell
-            .get_or_init(|| match self.init.lock().unwrap().take() {
-                Some(f) => f(),
-                None => unreachable!(),
-            })
+        self.cell.get_or_init(self.init)
     }
 }
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -1,22 +1,20 @@
 // A poly-fill for `lazy_cell`
 // Replace with std::sync::LazyLock when https://github.com/rust-lang/rust/issues/109736 is stablized.
 
-use std::cell::Cell;
 use std::ops::Deref;
+use std::sync::Mutex;
 use std::sync::OnceLock;
 
 pub(crate) struct Lazy<T, F = fn() -> T> {
     cell: OnceLock<T>,
-    init: Cell<Option<F>>,
+    init: Mutex<Option<F>>,
 }
-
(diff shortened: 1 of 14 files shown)


james7132:
Didn't realize the type was in the public interface of the crate. Is this still acceptable as is?

daxpedda (commented):
I'm not convinced there is sufficient motivation to do this as long as:
- `once_cell` is one of the most popular crates out there, e.g. our dependencies on Linux use it anyway, so removing it in Winit doesn't actually get rid of the dependency (on Linux).
- Our manual implementation requires `unsafe`.
- `once_cell` is a fairly small and well reviewed dependency.

It doesn't affect Web, so its up to the other maintainers.

daxpedda:
> Didn't realize the type was in the public interface of the crate. Is this still acceptable as is?

I assume you mean because of all the CI failures? These things are not actually publicly exposed, they should rather be `pub(crate)` or `pub(super)`.

james7132:
> Our manual implementation requires `unsafe`.

This really only is needed if the poly-fill is kept. The actual implementation isn't too bad if we replace every `Lazy` with a `OnceLock`.

> I assume you mean because of all the CI failures? These things are not actually publicly exposed, they should rather be pub(crate) or pub(super).

Would it be OK to turn those into `pub(crate)`/`pub(super)` in this PR?

---

With that said, I'm happy to close this out for now and revisit this when `LazyLock` gets stablized, though probably undesirable until the MSRV bump is acceptable.

madsmtm:
I'd rather go with converting our current use of `once_cell` to the stable `OnceLock`, since although it's minor, I don't want to add further `unsafe` to Winit if it can be avoided.

src/utils.rs
```diff
@@ -0,0 +1,33 @@
...
+use std::cell::Cell;
+use std::ops::Deref;
+use std::sync::OnceLock;
+
+pub(crate) struct Lazy<T, F = fn() -> T> {
+    cell: OnceLock<T>,
+    init: Cell<Option<F>>,
+}
+
+unsafe impl<T, F> Sync for Lazy<T, F> {}
```
notgull, line 13: If it were an `Fn() -> T + Sync` this line of unsafe would not be needed.

notgull (changes requested):

notgull:
Agreed with moving away from OnceCell.

src/utils.rs
james7132:
```
error[E0277]: `std::cell::Cell<Option<fn() -> HashSet<KeyCode>>>` cannot be shared between threads safely
  --> src\platform_impl\windows\keyboard_layout.rs:85:25
   |
85 | static NUMPAD_KEYCODES: Lazy<HashSet<KeyCode>> = Lazy::new(|| {
   |                         ^^^^^^^^^^^^^^^^^^^^^^ `std::cell::Cell<Option<fn() -> HashSet<KeyCode>>>` cannot be shared between threads safely
   |
   = help: within `Lazy<HashSet<KeyCode>>`, the trait `Sync` is not implemented for `std::cell::Cell<Option<fn() -> HashSet<KeyCode>>>`
   = note: if you want to do aliasing and mutation between multiple threads, use `std::sync::RwLock`
note: required because it appears within the type `Lazy<HashSet<KeyCode>>`
  --> src\utils.rs:8:19
   |
8  | pub(crate) struct Lazy<T, F: Fn() -> T + Sync = fn() -> T> {
   |                   ^^^^
   = note: shared static variables must have a type that implements `Sync`
```
The issue primarily stems from the Cell on the inside, but said cell is only accessed if and only if the OnceLock has not been initialized, so the unsafe impl is indeed still sound so long as T and F is Send.

james7132: One way around this is just to use `Mutex` in place of the `Cell`, and eat the initialization cost once.

james7132: Nevermind, I misunderstood your comment. Switched to using a function pointer instead of a generic over potential closures.

james7132:
I eliminated the unsafe at @notgull's suggestion.

daxpedda (approved):
Cool!

madsmtm (approved):

notgull (approved):
Thanks!

PR #3523, opened 2024-02-26
Filter out tiny mouse events

### LINKED ISSUE #3500: X11: Mouse motion events sometimes contain garbage

notgull:
Usually, if mouse events are equal to (0, 0) we filter them out.
However, if the event is very close to zero it will still be given to
the user. In some cases this can be caused by bad float math on the X11
server side.

I fix this by refusing to forward events where both of their deltas fall
below a certain threshold.

Closes #3500

### DIFF
--- a/src/platform_impl/linux/x11/event_processor.rs
+++ b/src/platform_impl/linux/x11/event_processor.rs
@@ -1518,8 +1518,8 @@ impl EventProcessor {
         let mask =
             unsafe { slice::from_raw_parts(xev.valuators.mask, xev.valuators.mask_len as usize) };
         let mut value = xev.raw_values;
-        let mut mouse_delta = (0.0, 0.0);
-        let mut scroll_delta = (0.0, 0.0);
+        let mut mouse_delta = util::Delta::default();
+        let mut scroll_delta = util::Delta::default();
         for i in 0..xev.valuators.mask_len * 8 {
             if !xinput2::XIMaskIsSet(mask, i) {
                 continue;
@@ -1529,10 +1529,10 @@ impl EventProcessor {
             // We assume that every XInput2 device with analog axes is a pointing device emitting
             // relative coordinates.
             match i {
-                0 => mouse_delta.0 = x,
-                1 => mouse_delta.1 = x,
-                2 => scroll_delta.0 = x as f32,
-                3 => scroll_delta.1 = x as f32,
+                0 => mouse_delta.set_x(x),
+                1 => mouse_delta.set_y(x),
+                2 => scroll_delta.set_x(x as f32),
+                3 => scroll_delta.set_y(x as f32),
                 _ => {}
             }
 
@@ -1548,7 +1548,7 @@ impl EventProcessor {
             value = unsafe { value.offset(1) };
         }
 
-        if mouse_delta != (0.0, 0.0) {
+        if let Some(mouse_delta) = mouse_delta.consume() {
             let event = Event::DeviceEvent {
                 device_id: did,
                 event: DeviceEvent::MouseMotion { delta: mouse_delta },
@@ -1556,7 +1556,7 @@ impl EventProcessor {
             callback(&self.target, event);
         }
 
-        if scroll_delta != (0.0, 0.0) {
+        if let Some(scroll_delta) = scroll_delta.consume() {
             let event = Event::DeviceEvent {
                 device_id: did,
                 event: DeviceEvent::MouseWheel {
--- a/src/platform_impl/linux/x11/util/mouse.rs
+++ b/src/platform_impl/linux/x11/util/mouse.rs
@@ -1,0 +1,52 @@
+//! Utilities for handling mouse events.
+
+/// Recorded mouse delta designed to filter out noise.
+pub struct Delta<T> {
+    x: T,
+    y: T,
+}
+
+impl<T: Default> Default for Delta<T> {
+    fn default() -> Self {
+        Self {
+            x: Default::default(),
+            y: Default::default(),
+        }
+    }
+}
+
+impl<T: Default> Delta<T> {
+    pub(crate) fn set_x(&mut self, x: T) {
+        self.x = x;
+    }
+
+    pub(crate) fn set_y(&mut self, y: T) {
+        self.y = y;
+    }
+}
+
+macro_rules! consume {
+    ($this:expr, $ty:ty) => {{
+        let this = $this;
+        let (x, y) = match (this.x.abs() < <$ty>::EPSILON, this.y.abs() < <$ty>::EPSILON) {
+            (true, true) => return None,
+            (true, false) => (this.x, 0.0),
+            (false, true) => (0.0, this.y),
+            (false, false) => (this.x, this.y),
+        };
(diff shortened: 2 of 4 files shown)


src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -41,6 +41,9 @@ use crate::platform_impl::x11::{
 /// The X11 documentation states: "Keycodes lie in the inclusive range `[8, 255]`".
 const KEYCODE_OFFSET: u8 = 8;
 
+/// Minimum delta for a device event to be considered.
+const NOTGULL_THRESHOLD: f32 = 4.0e-8;
```
madsmtm, line 45: I should insert my username in more source code! 😁

(I've tried hard with the `mtm: MainThreadMarker`, but while `mtm` is my initials, nothing beats just spelling out the whole thing)

madsmtm (approved):

src/platform_impl/linux/x11/event_processor.rs
kchibisov: Should be renamed to something useful. Also it should be explained why the given constant was picked in the first place, otherwise it _may_ not work for some cases.

There's also a case where you move the mouse really slowly, thus filtering may not actually move the mouse where it should or accumulate in error.

src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -1464,15 +1467,17 @@ impl EventProcessor {
             value = unsafe { value.offset(1) };
         }
 
-        if mouse_delta != (0.0, 0.0) {
+        if mouse_delta.0.abs() as f32 >= NOTGULL_THRESHOLD
+            || mouse_delta.1.abs() as f32 >= NOTGULL_THRESHOLD
```
kchibisov, line 1471: I don't quite like this logic in-particular. Could you instead check if the given value is small enough and null it? Then just send the thing if something is not null, since those low values don't make much sense in the first place.

kchibisov (changes requested):

valaphee:
This wouldn't fix the underlying issue. Interestingly, I didn't had any issues with Bevy for example, but its also not a regression.

Didn't use thresholds because normally raw mouse input events should not be less than 1.0

notgull:
> This wouldn't fix the underlying issue. Interestingly, I didn't had any issues with Bevy for example, but its also not a regression.
> 
> Didn't want to use thresholds because normally raw mouse input events should not be less than 1.0

I think the underlying issue is that the underlying driver is making an error somewhere in the floating point calculation and sending us a very tiny number instead of zero. `7.1e-222` is far less than the epsilon on all platforms, so for all intents and purposes it might as well be zero.

src/platform_impl/linux/x11/event_processor.rs
notgull: I've elected to use the float epsilon instead. Since float comparison breaks down past the epsilon, for all intentions and purposes any number less than the epsilon is zero.

valaphee:
Then this would definitely be a valid fix, but the value is always near zero, even if there is movement

src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -1889,6 +1889,59 @@ impl EventProcessor {
     }
 }
 
+/// Recorded mouse delta designed to filter out noise.
+struct Delta<T> {
+    inner: Option<(T, T)>,
+}
```
kchibisov, line 1895: Put it somewhere else, we have entire directly with called `util`(the most useless name for directly btw) just laying around for stuff like that.

kchibisov (changes requested):

notgull:
> Then this would definitely be a valid fix, but the value is always near zero, even if there is movement, that's why I don't think it's a floating point error.

My understanding of Xinput's raw mouse events is that they are always multiples of 1. So we shouldn't be seeing any floating point numbers anyways. So it's fine for us to assume that a value less than the epsilon is, for all intents and purposes, supposed to be zero.

src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -1889,6 +1889,59 @@ impl EventProcessor {
     }
 }
 
+/// Recorded mouse delta designed to filter out noise.
+struct Delta<T> {
+    inner: Option<(T, T)>,
```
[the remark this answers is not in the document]
notgull: Good point, fixed

src/platform_impl/linux/x11/event_processor.rs
notgull: Moved to `util/mouse.rs`

src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -1464,15 +1464,15 @@ impl EventProcessor {
...
+        if let Some(mouse_delta) = mouse_delta.consume() {
             let event = Event::DeviceEvent {
                 device_id: did,
                 event: DeviceEvent::MouseMotion { delta: mouse_delta },
             };
             callback(&self.target, event);
         }
 
-        if scroll_delta != (0.0, 0.0) {
+        if let Some(scroll_delta) = scroll_delta.consume() {
```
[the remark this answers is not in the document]
notgull: Good point, fixed

kchibisov (approved):

src/platform_impl/linux/x11/util/mouse.rs
```diff
@@ -0,0 +1,52 @@
...
+    }
+}
+
+macro_rules! consume {
+    ($this:expr, $ty:ty) => {{
+        let this = $this;
+        let (x, y) = match (this.x.abs() < <$ty>::EPSILON, this.y.abs() < <$ty>::EPSILON) {
+            (true, true) => return None,
+            (true, false) => (this.x, 0.0),
+            (false, true) => (0.0, this.y),
```
irate-devil, line 34: Hey yall :) The code here is mistakenly returning the axis that's lower than the epsilon.
It's the likely cause of this issue in Bevy: https://github.com/bevyengine/bevy/issues/12198.
```suggestion
            (false, true) => (this.x, 0.0),
            (true, false) => (0.0, this.y),
```

Issue #3525, opened 2024-02-26
Build error on nightly web


notgull:
https://github.com/rust-windowing/winit/actions/runs/8043172792/job/21964856803?pr=3523

PR #3526, opened 2024-02-26
Fix nightly CI

daxpedda:
Fixes https://github.com/rust-windowing/winit/issues/3525.

Currently nightly fails because of FPs introduced by the new `non_local_definitions` lint.
FPs are tracked in https://github.com/rust-lang/rust/issues/121621.


kchibisov (approved):

daxpedda:
After some more testing and reading I'm not exactly confident if this is an FP or not, but I will keep an eye on it and eventually update it.

Issue #2395, opened 2022-07-25
Split max/min sizes constraints


amrbashir:
I think it is a valid use-case to only constrain the window width or height.

I think an implementation of this would be similar to `LogicalSize` and `PhysicalSize`
> I couldn't think of a better names so these are subject to change ofc
```rs
#[derive(Debug, Copy, Clone, Eq, PartialEq, Default, Hash)]
struct LogicalUnit<P>(P);

#[derive(Debug, Copy, Clone, Eq, PartialEq, Default, Hash)]
struct PhysicalUnit<P>(P);

enum Unit {
    Physical(PhysicalUnit<u32>),
    Logical(LogicalUnit<f64>),
}


impl WindowBuilder {
  fn with_max_inner_size<S: Into<Unit>>(width: Option<U>, height: Option<U>);
}
```

madsmtm:
The implementation could build upon https://github.com/rust-windowing/winit/pull/2148

PR #3529, opened 2024-02-26
Add `LogicalPoint`, `PhysicalPoint` and `Point`

amrbashir:
part of #2395

### DIFF
--- a/dpi/CHANGELOG.md
+++ b/dpi/CHANGELOG.md
@@ -1,0 +1,14 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+Please keep one empty line before and after all headers. (This is required for
+`git` to produce a conflict when a release is made while a PR is open and the
+PR's changelog entry would go into the wrong section).
+
+And please only add new entries to the top of this list, right below the `#
+Unreleased` header.
+
+# Unreleased
+
+- Add `LogicalUnit`, `PhysicalUnit` and `PixelUnit` types and related functions.
--- a/dpi/src/lib.rs
+++ b/dpi/src/lib.rs
@@ -35,8 +35,8 @@
 //!
 //! ### Position and Size types
 //!
-//! The [`PhysicalPosition`] / [`PhysicalSize`] types correspond with the actual pixels on the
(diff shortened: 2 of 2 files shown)


daxpedda:
Looking at #2395 I don't really follow what the motivation here is exactly.
Is there some previous related discussion?

daxpedda:
My question is specifically how does `Logical/PhysicalPoint` relate to that? Why can that not be done with the types we have now?

daxpedda:
Ah, I see, thank you!

kchibisov (changes requested):
Should probably update the docs in `### Position and Size types` with the point stuff.

kchibisov:
Besides.

@amrbashir given that you had an interest in such crate in the first place, how you deal with e.g. rounding? For example in winit we have special rounding rules on Wayland, because it's rounding is a bit special, but maybe we should define such things in this crate as well, so the `to_logical/to_physical` will account for rounding and make the default like it was before?

amrbashir:
We didn't give it much thought, since our windows and macOS backend were forked from `winit` and the current behavior of `to_logical/to_physical` have been working fine for us.

amrbashir:
I just realized that the naming `Point` might be confusing when used with width/height as it conveys position than a dimension, maybe it should be renamed to `LogicalUnit`, `PhysicalUnit` and `Unit` or `PixelUnit` ?

kchibisov:
>I just realized that the naming Point might be confusing when used with width/height as it conveys position than a dimension, maybe it should be renamed to LogicalUnit, PhysicalUnit and Unit or PixelUnit ?

Maybe `LogicalUnit`/`PhysicalUnit` is better.

kchibisov (approved):

dpi/CHANGELOG.md
daxpedda, line 1: I guess we are fine with not also moving this changelog to docs.rs like in #3496?

daxpedda (approved):

dpi/CHANGELOG.md
kchibisov: Yeah, it's not that important.

notgull (approved):

PR #3538, opened 2024-03-01
Add documentation example of ignoring key repeats

madsmtm:
Fixes https://github.com/rust-windowing/winit/discussions/3373.


src/event.rs
```diff
@@ -781,6 +781,31 @@ pub struct KeyEvent {
...
+    /// use winit::event::{WindowEvent, KeyEvent, ElementState};
+    /// use winit::keyboard::{KeyCode, PhysicalKey};
+    /// # let window_event = WindowEvent::RedrawRequested; // To make the example compile
+    /// match window_event {
+    ///     WindowEvent::KeyboardInput {
+    ///         event: KeyEvent {
+    ///             physical_key: PhysicalKey::Code(KeyCode::KeyW),
+    ///             state: ElementState::Pressed,
+    ///             repeat: false,
+    ///             ..
```
kchibisov, line 800: could you include `synthetic: false` here as well?

madsmtm: Hmm, do we actually want to recommend that? If I'm writing a 2D game, I'd like to know when the user is pressing `W`, so that I can move the player upwards; in that case, I'd probably _want_ to handle synthetic events too?

kchibisov: I mean they are not represent real keyboard input and I'd rather either move them somewhere or remove all together.

madsmtm: I've opened https://github.com/rust-windowing/winit/issues/3543 for that.

daxpedda (approved):

Issue #3512, opened 2024-02-23
All of the examples are missing


aloucks:
As stated all of the examples are gone. There's one mega example, but it's less helpful when trying to figure out a specific concept.

madsmtm:
Yeah, we merged them in https://github.com/rust-windowing/winit/pull/3447, since we thought it important to give a fuller example of what a real application might look like. What specific concept are you struggling with?

aloucks:
It's more that they offered an index for features and how to solve specific problems. Now the end users have to go through that whole mega example to find the thing they are looking for. Reading and understanding 1000 lines of unfamiliar code is much more involved than 50-100 lines of highly a specific example. Having an index of small examples is much easier from a user perspective.

kchibisov:
I still don't understand, you go and read methods on the `Window` itself, because that's what most of them did and what got removed. Examples on how to use completely different things like pump_events, child_windows are still there.

Like if the struggle is with search within a file I'm not really sure how to help with that. In general really small examples should go into the method/event docs, since that's where they should be.

The example won't change anyway, since it's the only way to have actually useful bug reports, since you can test with that example pretty much everything relatively easy and compare.

madsmtm:
I've reopened this to allow further discussion of this (CC https://github.com/rust-windowing/winit/pull/4136).

kchibisov:
my stance is the same, if example shows a specific pattern of usage of winit in its own (run_on_demand, some drawing pattern) it deserves its own example, otherwise it should go to the documentation method.

Maintaining hundreds of examples at this state is not a good option. One could move e.g. platform specific examples into its own example, like `windows_window.rs`, etc, so platform specific patterns are shown in isolation, but that's about it.

PR #3542, opened 2024-03-01
Refactor main example

madsmtm:
Follow-up to https://github.com/rust-windowing/winit/pull/3447, since I didn't really review the example back then. See each commit for details.

Should also resolve some of the concerns raised in https://github.com/rust-windowing/winit/issues/3512. @aloucks, does this make it better for you? (Especially the rename from `window` -> `full`).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -257,7 +257,7 @@ web-sys = { version = "0.3.22", features = ['CanvasRenderingContext2d'] }
 
 [[example]]
 doc-scrape-examples = true
-name = "window"
+name = "full"
 
 [workspace]
 resolver = "2"
--- a/examples/full.rs
+++ b/examples/full.rs
@@ -1,4 +1,7 @@
-//! Simple winit application.
+//! An example showcasing various functionality that Winit has.
+//!
+//! Often used by Winit developers to test certain functionality, may be a bit
+//! verbose.
 
 use std::collections::HashMap;
 use std::error::Error;
--- a/examples/monitors.rs
+++ b/examples/monitors.rs
@@ -1,0 +1,62 @@
+use std::error::Error;
+use winit::{
+    event::{Event, StartCause},
+    event_loop::{ActiveEventLoop, EventLoop},
+};
+
+fn main() -> Result<(), Box<dyn Error>> {
+    let event_loop = EventLoop::new()?;
+
+    Ok(event_loop.run(|event, event_loop| match event {
+        Event::NewEvents(StartCause::Init) => {
+            dump_monitors(event_loop);
+            event_loop.exit()
+        }
+        _ => {}
+    })?)
+}
+
+fn dump_monitors(event_loop: &ActiveEventLoop) {
+    println!("Monitors information");
+    let primary_monitor = event_loop.primary_monitor();
+    for monitor in event_loop.available_monitors() {
+        let intro = if primary_monitor.as_ref() == Some(&monitor) {
+            "Primary monitor"
+        } else {
+            "Monitor"
+        };
+
+        if let Some(name) = monitor.name() {
+            println!("{intro}: {name}");
+        } else {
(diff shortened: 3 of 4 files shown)


kchibisov (changes requested):
This is just a giant mess now.

madsmtm:
> This is just a giant mess now.

Which commits in particular are you dissatisfied with?

This isn't code like I'd normally write it, but I believe it's better this way as example code, since it's easier to jump in to.

kchibisov:
I don't like to advise to write garbage code, it's re

1edf4f123885a12590eef3c2b3e7faef8e6cb772 (makes the code less readable)
c6dfe620ff093ed1d10dc10d7c7799f34d8500a8 (will start format like garbage iirc)
21fecf2cb23e7be5a76f42ff5cde3ec26d233888 (even worse code which is hard to follow, code is less dense now)
926fb051de17ec6cac291e69a8e87c0e127bebda (there always should be a window example)

madsmtm:
> I don't like to advise to write or suggest garbage code
>
> [1edf4f1](https://github.com/rust-windowing/winit/commit/1edf4f123885a12590eef3c2b3e7faef8e6cb772) (makes the code less readable)

I disagree, I think it becomes much more readable, _especially_ when you're new to the codebase, and don't know what things mean. More specifically, what is it that you dislike? Is it the lack of encapsulation (`Application` is accessing `WindowState`'s fields directly) or the length of the `handle_action` function?

Remember, we also have to optimize for people without `rust-analyzer` set up, most people are going to be viewing this on GitHub.

> [c6dfe62](https://github.com/rust-windowing/winit/commit/c6dfe620ff093ed1d10dc10d7c7799f34d8500a8) (will start format like garbage iirc)

The functionality is unchanged in that commit, the `Display` impl was just forwarding to `Debug`.

> [21fecf2](https://github.com/rust-windowing/winit/commit/21fecf2cb23e7be5a76f42ff5cde3ec26d233888) (even worse code which is hard to follow, code is less dense now)

I can agree that this one may be dubious, though I do still think it improves first-time readability.

> [926fb05](https://github.com/rust-windowing/winit/commit/926fb051de17ec6cac291e69a8e87c0e127bebda) (there always should be a window example)

Then let's add another example called `window`, that instead does the bare minimum?

kchibisov:
>Remember, we also have to optimize for people without rust-analyzer set up, most people are going to be viewing this on GitHub.

github has code navigation for more than a year. I don't even use rust analyzer to navigate any of that as well.

>Then let's add another example called window, that instead does the bare minimum?

Then you should do that in the first place.

>I can agree that this one may be dubious, though I do still think it improves first-time readability.

No, no one cares about giant functions on how to pick location or how all the cursors are named in the example, it's just noise and doesn't show how to do things. 

I'd agree to inline the window WindowState creation and remove its new, and split monitors stuff, but that's about it. I don't care about rename as long as we have `window.rs` example. The new example should be named `app.rs`

And no, this won't solve the linked issue because it's completely irrelevant to what you did and how it should be solved.

daxpedda (approved):
> Then let's add another example called `window`, that instead does the bare minimum?

I think that would be good indeed.

---

I do agree that this is not usually how you write code, but this isn't supposed to be a good Rust code example, its supposed to be a good example on how to use Winit (obviously there is a spectrum here). In this sense I believe these changes are an improvement.

Though it is true that GitHub has code navigation, scrolling through the example is definitely easier then jumping back and forth between functions inside of GitHub.

nixpulvis:
As someone who is still somewhat unfamiliar with winit, I find this example pretty readable and helpful. The only functions that seem complex are `Application::new` and `Application::create_window`, everything else seems like reasonable boilerplate for a full fledged application.

I do think having a more minimal example would help people looking to build quickly from the ground up though, which is usually what I look to do when learning a new library.

PR #3549, opened 2024-03-01
Rework CONTRIBUTING guidelines

kchibisov:
Give better recommendations for contributors on how to propose start the work, write commit messages, and handle review by explicitly stating expectations, however make it clear that maintainers will help you if you have issues.

Write down the guidelines for maintainers on how they should handle contributions and interact with other maintainers to help maintaining good code quality level by listing maintainers responsibilities and providing suggestions on how to handle various situations.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,57 +1,169 @@
-# Winit Contributing Guidelines
+# Contribution Guidelines
 
-## Scope
-[See `FEATURES.md`](./FEATURES.md). When requesting or implementing a new Winit feature, you should
-consider whether or not it's directly related to window creation or input handling. If it isn't, it
-may be worth creating a separate crate that extends Winit's API to add that functionality.
+This document contains guidelines for contributing code to winit. It has to be
+followed in order for your patch to be approved and applied.
 
+## Contributing
 
-## Reporting an issue
+Anyone can contribute to winit, however given that it's a cross platform
+windowing toolkit getting certain changes incorporated upstream could be
+challenging.
 
-When reporting an issue, in order to help the maintainers understand what the problem is, please make
-your description of the issue as detailed as possible:
+To save your time it's wise to check already opened [pull requests][prs] and
+[issues][issues]. In general, bug fixes and missing implementations are always
+accepted, however new API proposals should go into the issue first. When in
+doubt contact us on [matrix][matrix] or via opening an issue.
 
-- if it is a bug, please provide a clear explanation of what happens, what should happen, and how to
-  reproduce the issue, ideally by providing a minimal program exhibiting the problem
-- if it is a feature request, please provide a clear argumentation about why you believe this feature
-  should be supported by winit
+### Submitting your work and handling review
 
-## Making a pull request
+All patches have to be sent on Github as [pull requests][prs]. To simplify your
+life during review it's recommended to check the "give contributors write access
+to the branch" checkbox.
 
-When making a code contribution to winit, before opening your pull request, please make sure that:
+#### Creating commits
 
-- your patch builds with Winit's minimal supported rust version - Rust 1.70.
-- you tested your modifications on all the platforms impacted, or if not possible, detail which platforms
-  were not tested, and what should be tested, so that a maintainer or another contributor can test them
-- you updated any relevant documentation in winit
-- you left comments in your code explaining any part that is not straightforward, so that the
-  maintainers and future contributors don't have to try to guess what your code is supposed to do
-- your PR adds an entry to the changelog file if the introduced change is relevant to winit users.
+When creating a commit, follow these general rules:
+  - Try to limit the first line (title) of the commit message to 50 characters,
+    but not more than 60 characters.
+  - The commit starts with an uppercase latter and everything else must be
+    lowercase, unless you quote a symbol, like `Window`.
+  - Use the body of the commit message to actually explain what your patch does
+    and why it is useful. Even if your patch is a one line fix, the description
+    is not limited in length and may span over multiple paragraphs. Use proper
+    English syntax, grammar and punctuation. The body is generally wrapped
+    at 72 characters.
+  - If you're a downstream user and your fixing your issue, consider to link it
+    like `Links: URL` in the end of the commit message.
+  - Try to address only one issue/topic per commit.
+  - Prefer describing your changes in imperative mood, e.g. *"make xyzzy do
+    frotz"* instead of *"[This patch] makes xyzzy do frotz"* or *"[I] changed
+    xyzzy to do frotz"*, as if you are giving orders to the codebase to change
+    its behaviour.
+  - When fixing an issue from https://github.com/rust-windowing/winit/issues,
+    ensure to have it include in the end of the commit. See other commits
+    on how it's done.
+  - When in doubt, follow the format and layout of the recent existing commits.
 
-  You needn't worry about the added entry causing conflicts, the maintainer that merges the PR will
-  handle those for you when merging (see below).
-- if your PR affects the platform compatibility of one or more features or adds another feature, the
-  relevant sections in [`FEATURES.md`](https://github.com/rust-windowing/winit/blob/master/FEATURES.md#features)
-  should be updated.
+The maintainers will fixup your commit message when applying the patch, but it's
+still recommended to follow the rules above to make the life of maintainers
+easier.
 
-Once your PR is open, you can ask for a review by a maintainer of your platform. Winit's merging policy
-is that a PR must be approved by at least two maintainers of winit before being merged, including
-at least a maintainer of the platform (a maintainer making a PR themselves counts as approving it).
+If you want to communicate something to reviewers, but it doesn't belong to
+the commit message, write that under `--` on PR body. Your PR could look like:
 
-Once your PR is deemed ready, the merging maintainer will take care of resolving conflicts in
-`CHANGELOG.md` (but you must resolve other conflicts yourself). Doing this requires that you check the
-"give contributors write access to the branch" checkbox when creating the PR.
+  ```
+  TITLE (matches commit title)
+
+  Body (matches commit body)
+
+  --
+
+  Your arbitrary message goes here.
+
+  Various checkboxes PR template required you to fill.
+  ```
+
+#### Handling review
+
+During the review process certain events could require an action from your side,
+to communication more efficient the common patterns and reactions are described
+below.
+
+_Event:_ The CI fails to build, but it looks like not your fault. Not
+communicating so, could result into maintainers not looking into your patch,
+unless they CI that CI pass.\
+_Desired behavior:_ Write a message saying roughly the following "The CI failure
+is unrelated", so maintainers will fix it for you.
+
+_Event:_ Collaborator requested review on your PR.\
+_Desired behavior:_ Once you address the request, you _should_ re-request review
+with the github's UI. If you don't agree with what maintainer suggested, you
+should object that and re-request the review. That will indicate that the
+_ball_ is on maintainer's side.
+
+_Event:_ You've opened a PR, but maintainer shortly after commented that they
+want to work on that _themselves_.\
+_Desired behavior:_ Discusses with maintainer their plans if they were not
+outlined in the initial response, because such response means that they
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -1,57 +1,169 @@
...
+### Submitting your work and handling review
+
+All patches have to be sent on Github as [pull requests][prs]. To simplify your
+life during review it's recommended to check the "give contributors write access
+to the branch" checkbox.
+
+#### Creating commits
+
+When creating a commit, follow these general rules:
+  - Try to limit the first line (title) of the commit message to 50 characters,
```
notgull, line 26: I think that a link here to the Git standards this is taken from would be appreciated.

CONTRIBUTING.md
```diff
@@ -1,57 +1,169 @@
...
+    must preserve attribution with `Co-authored-by`, `Suggested-by`, or keep
+    the original commiter.
+  - Rebase their work for them when massive changes to winit codebase were
+    introduced.
+
+When reviewing code of other maintainers all of the above is on the maintainer
+who submitted the patch. Interested maintainers could help push the work over
+the finish line, but teaching other maintainers should be preferred.
+
+When contributor is _regular_ in winit, the maintainer should slowly start
```
notgull, line 146: This paragraph seems unnecessary and overly restrictive, so some context would be appreciated.

notgull (changes requested):
It should also be mentioned here that Markdown files should be wrapped to 80 characters.

CONTRIBUTING.md
kchibisov: I mean, it's a regular process of evolving. It's present in mostly every project I worked in.

CONTRIBUTING.md
kchibisov: 50 characters is not that hard to google, but the source is a bit vague on that. I'd add that multiple editors (nvim, helix) do that sort of indication by default. The 72 chars body is from kernel and in general a lot of inspired by kernel devs, since they know how to write commits. 

https://www.kernel.org/doc/html/latest/process/submitting-patches.html#describe-your-changes

kchibisov:
>It should also be mentioned here that Markdown files should be wrapped to 80 characters.

I was thinking on adding `.editorconfig` for that.

notgull:
> I was thinking on adding `.editorconfig` for that.

While this would be a good solution I would prefer it if it was stated explicitly here.

CONTRIBUTING.md
notgull: I guess what I'm asking is, what bad behavior is this intended to stop? I feel like this is a solution in search of a problem.

CONTRIBUTING.md
notgull: Having this linked in this document would be nice

daxpedda (changes requested):
I think in general this looks good, I would probably like to proof read it but otherwise I believe this is a great improvement.

---

Personally I'm not in favor of documenting changes in commit messages. Documenting changes in PRs has to be done anyway, as it is required to actually get anything merged, otherwise nobody would know whats going on. So to me this is duplicating that work.

I do believe that documenting changes in commit messages is an improvement, but it comes with significant drawbacks. Most notably its a lot more work and according to the new guidelines, that work has to be taken on by maintainers for contributors. Those tradeoffs ought to be weighed and discussed.

Additionally, I would like to point out that this isn't a particularly popular way to do things anymore, e.g. apart from Linux and security related projects I've not seen it often in the Rust open source ecosystem. I believe that today, in general, documenting changes has moved from commit messages to PR's and issues, which I personally don't mind.

I also believe that squashing is fine as long as its not worth to preserve the commit history, which it often isn't in Winit, especially for smaller changes (apart from the fact that commit history is still preserved in the PR itself).

Good commit messages are good though and have minimal tradeoff, I don't see a reason to not have good commit messages. But I do think that even 60 characters is a very traditional limit that doesn't really make sense with modern screens and UIs anymore. Again, we should discuss the tradeoffs here.

I want to clarify that I do actually believe that clearly documenting changes in the commit message is always going to be an improvement. But every improvement has tradeoffs that we should weigh and discuss.

Imo, the Rust open source ecosystem in general doesn't require documenting changes in commit messages (not even the Rust project itself), probably because its a much newer ecosystem, therefor I don't see why we should be different here considering the tradeoffs.

kchibisov:
>I do believe that documenting changes in commit messages is an improvement, but it comes with significant drawbacks. Most notably its a lot more work and according to the new guidelines, that work has to be taken on by maintainers for contributors. Those tradeoffs ought to be weighed and discussed.

I already do that myself for each PR I merge on the span of 3 years. I don't mind continue doing so, it usually takes like a few minutes. Keep in mind that it's recommended to write correct messages, but to not demotivate contributors, I don't mind fixing them myself.

>I also believe that squashing is fine as long as its not worth to preserve the commit history, which it often isn't in Winit, especially for smaller changes (apart from the fact that commit history is still preserved in the PR itself).

Squashing is fine, the thing is that this option can't be used reliably by some parties, once there will be indication that we can write commit messages so things look consistent, I'd bring it back.

>Good commit messages are good though and have minimal tradeoff, I don't see a reason to not have good commit messages. But I do think that even 60 characters is a very traditional limit that doesn't really make sense with modern screens and UIs anymore. Again, we should discuss the tradeoffs here.

Well, it's the topic line, brevity is more important. If you need more than you likely need to split commit into multiples.

>Imo, the Rust open source ecosystem in general doesn't require documenting changes in commit messages (not even the Rust project itself), probably because its a much newer ecosystem, therefor I don't see why we should be different here considering the tradeoffs.

That sounds like their issue and the fallout of using bors. Though, people generally document why they make a particular change. The thing is that you don't really document, but rather express motivation for change, the documentation should belong in code.

CONTRIBUTING.md
```diff
@@ -1,57 +1,169 @@
...
+### Contributions handling
+
+The maintainer must ensure that the external contributions meet the winit's
+quality standards. If it's not, it **is maintainer's responsibility** to bring
+it on par, which includes:
+
+  - Decline the patch if it tries to achieve short term goals instead of solving
+    the problem in general. In such case maintainer must communicate that **as
+    early as possible**, so contributors don't spend their time on something
+    that won't be accepted in the end.
```
madsmtm, line 115: I acknowledge that this may lead to decreased code quality, but I believe that requiring the problem to be solved in general is the wrong approach for Winit. While the project is a few years old, it's still young enough that we don't yet know how to do everything! And the absolute best way to get knowledge is to let people experiment with the code!

As an example, https://github.com/rust-windowing/winit/pull/3499 is solving a short-term goal, but I believe it'll also help guide how the API will have to look like in the future.

Another example is the file open APIs on macOS, which got rejected partly because of technical difficulties in making it work right, but also because we wanted to find some solution to https://github.com/rust-windowing/winit/issues/2120, which has been open for over two years with only recently a solution on the horizon.

My opinion is that we need to get all of the sub-par code into Winit, so that we can iterate on it in there instead of in a PR that will have to be rebased over an over (keyboard v2).

madsmtm (changes requested):
Heh, I've been stressing out the whole week because I wanted to write down some of the things we talked about last meeting so that we could continue the discussion, and bam, I should've just looked on GH, you already did it ;)

So thank you, it will certainly help us in aligning our goals!

CONTRIBUTING.md
kchibisov: >I acknowledge that this may lead to decreased code quality, but I believe that requiring the problem to be solved in general is the wrong approach for Winit. While the project is a few years old, it's still young enough that we don't yet know how to do everything! And the absolute best way to get knowledge is to let people experiment with the code!

it's like 10 years old at this point and it was inside glutin at some point.

>As an example, https://github.com/rust-windowing/winit/pull/3499 is solving a short-term goal, but I believe it'll also help guide how the API will have to look like in the future.

This is not how the API will look though. The trait extension is _fine_, but it was not proposed there.

>Another example is the file open APIs on macOS, which got rejected partly because of technical difficulties in making it work right, but also because we wanted to find some solution to https://github.com/rust-windowing/winit/issues/2120, which has been open for over two years with only recently a solution on the horizon.

Yes, the solution is merged and can be used to implement the said API already.

>My opinion is that we need to get all of the sub-par code into Winit, so that we can iterate on it in there instead of in a PR that will have to be rebased over an over (keyboard v2).

No, the subpar code will stay with us forever, that's how it works unfortunately. Once you land it it'll be pretty much like that for a long time. There're exceptions, but it's how it is, unfortunately. You should never justify not maintainable code.

CONTRIBUTING.md
```diff
@@ -1,57 +1,169 @@
...
+### Contributing
+
+When submitting a patch maintainer should follow the general contributing
+guidelines, however all soft rules (e.g `Try to`), become strict.
+
+To make life simpler for other maintainers it's suggested to create your branch
+under the project repository instead of your own fork. The naming scheme is
+`github_user_name/branch_name`. Doing so will make your work easier to rebase
+for other maintainers when you're absent.
+
```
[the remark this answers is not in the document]
kchibisov: I'm not sure it does? It requires you to be collaborator.

CONTRIBUTING.md
```diff
@@ -1,57 +1,169 @@
...
+When reviewing code of other maintainers all of the above is on the maintainer
+who submitted the patch. Interested maintainers could help push the work over
+the finish line, but teaching other maintainers should be preferred.
+
+When contributor is _regular_ in winit, the maintainer should slowly start
+requiring contributor to match *maintainer* quality standards when writing
+patches and commit messages.
+
+### Contributing
```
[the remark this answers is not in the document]
kchibisov: If you have a better name sure, but it's a subsection in maintainers section extending on the contributors section.

CONTRIBUTING.md
```diff
@@ -1,57 +1,169 @@
...
+    ensure to have it include in the end of the commit. See other commits
+    on how it's done.
+  - When in doubt, follow the format and layout of the recent existing commits.
+
+The maintainers will fixup your commit message when applying the patch, but it's
+still recommended to follow the rules above to make the life of maintainers
+easier.
+
+If you want to communicate something to reviewers, but it doesn't belong to
+the commit message, write that under `--` on PR body. Your PR could look like:
```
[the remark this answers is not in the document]
kchibisov: No the `--` is a standard symbol for this. It's automatically added by the git itself when you author a patch for submission the _traditional_ way.

CONTRIBUTING.md
```diff
@@ -1,57 +1,169 @@
...
+    but not more than 60 characters.
+  - The commit starts with an uppercase latter and everything else must be
+    lowercase, unless you quote a symbol, like `Window`.
+  - Use the body of the commit message to actually explain what your patch does
+    and why it is useful. Even if your patch is a one line fix, the description
+    is not limited in length and may span over multiple paragraphs. Use proper
+    English syntax, grammar and punctuation. The body is generally wrapped
+    at 72 characters.
+  - If you're a downstream user and your fixing your issue, consider to link it
+    like `Links: URL` in the end of the commit message.
```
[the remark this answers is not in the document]
kchibisov: I know, but it's a problem of github in particular. Having correct links helps a lot when looking why the patch was send in the first place.

CONTRIBUTING.md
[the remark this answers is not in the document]
madsmtm: But you're writing that in the GitHub PR body, not in Git?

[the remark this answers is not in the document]
kchibisov: yeah, that's true. I'm just really used to that notation and it's quite common.

CONTRIBUTING.md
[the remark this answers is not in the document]
madsmtm: There's a whole set of different GitHub "roles", but some things do require admin permissions, take a look at https://github.com/rust-windowing/winit/settings/access.

Organization-wide, there's "Owner" and "Member", and only the latter can add others to the org.

In any case, I mostly wanted to codify our previous decision about discussing administrative changes with others before doing them.

daxpedda (changes requested):
We decided to split off any commit message rules into a separate PR and merge everything else now, after all nits are resolved.

PR #3550, opened 2024-03-02
Add a code of conduct

notgull:
As previously discussed in today's meeting, this commit adds a code
of conduct to winit. I've elected to re-use the Rust project's Code
of Conduct, as it is also in use by other popular Rust projects and I
agree with its intentions. The goal is to set forward a set of explicit
expectations for how discourse in the rust-windowing project should be
conducted.

I've deliberately left out a mechanism for enforcing this code of
conduct. My hope is that, by the act of setting these expectations,
contributors will voluntarily follow them. In addition I have no
interest in being an internet janitor. If this becomes necessary in the
future we can look into it later.

rust-windowing collaborators, please read the linked Code of Conduct.
Not only for expectations of behavior, but also for expectations of how
it will be enforced should it come to it.

----

cc @rust-windowing/winit

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,6 @@
+# Code of Conduct
+
+The `rust-windowing` project adheres to the [Rust Code of Conduct]. This
+describes the minimum behavior expected from all contributors.
+
+[Rust Code of Conduct]: https://www.rust-lang.org/policies/code-of-conduct


kchibisov (approved):
it's a bit loose on how it declares things, but I guess it's ok.

daxpedda (approved):

madsmtm (approved):

PR #3554, opened 2024-03-02
Winit version 0.29.13

daxpedda:


kchibisov:
Is it that critical or we can offset it a bit?

daxpedda:
> Is it that critical or we can offset it a bit?

Its a crash, so I would argue its critical. Originally found in Bevy https://github.com/bevyengine/bevy/issues/12126.
Would prefer to get it out at least today.

kchibisov (changes requested):
Reword the commit to match other release commit names, we also keep the `Unreleased` if you look how other releases are done.

kchibisov (approved):

PR #3559, opened 2024-03-02
Fix incorrect delta filtering

### LINKED ISSUE #3558: Mouse events regression in 0.29.12+ on X11
Testing new the Bevy version 0.13 on Linux I encountered a mouse-look camera regression. I tracked it down to release 0.29.12 of `winit`.
On both videos I am trying to move my mouse horizontally and the only difference is `winit` version. 

winit 0.29.12
https://github.com/rust-windowing/winit/assets/160391/4962b5ef-d626-4464-8911-70332a6c4cb3
winit 0.29.11
https://github.com/rust-windowing/winit/assets/160391/83ca6774-8234-4981-986e-8f3e616d96d9

In the first one horizontal movement is very slow and generally mouse moves in a diagonal motion. In the second it works as expected.
I think it may have been caused by https://github.com/rust-windowing/winit/commit/3d4c53459a7e9a616df9673610a2e6bb48205e4f

Likely related to https://github.com/bevyengine/bevy/issues/12198

notgull:
This fixes the incorrect delta filtering. Due to a bug in the filtering
code, it would incorrectly filter out the correct mouse delta rather
than the one below zero. This fixes the order of events.

----

Closes #3558

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ Unreleased` header.
 - **Breaking:** Removed `EventLoopBuilder::with_user_event`, the functionality is now available in `EventLoop::with_user_event`.
 - Add `Window::default_attributes` to get default `WindowAttributes`.
 - `log` has been replaced with `tracing`. The old behavior can be emulated by setting the `log` feature on the `tracing` crate.
+- On X11, fix a bug where some mouse events would be unexpectedly filtered out.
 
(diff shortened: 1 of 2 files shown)


kchibisov (changes requested):
Need a changelog entry for that, since it was backported, so it's a _change_.

kchibisov (approved):

PR #3568, opened 2024-03-05
Bump `x11-dl` to v2.19.1

daxpedda:
This bumps the minimal version of `x11-dl` required from v2.18.5 to v2.19.1, which is required by the changes made in #3507.
Thank you @paulcacheux for point this out.

Addresses https://github.com/rust-windowing/winit/pull/3507#issuecomment-1973316798.


kchibisov:
@daxpedda could we commit the lock file along the way, because otherwise it's impossible to track things like that?

daxpedda:
Committing the lockfile as we discussed in https://github.com/rust-windowing/winit/issues/3085#issuecomment-1961733657 requires a bit more work, we should probably do it separately.

kchibisov (approved):

paulcacheux:
Thanks !

Issue #2260, opened 2022-04-21
How to popup virtual keyboard on mobile device?


Fancyflame:
I'm sorry but I couldn't find any API related to this feature.

madsmtm:
See also https://github.com/rust-windowing/winit/issues/1823, this is not yet implemented

daxpedda:
Closing in favor of #1823.

Issue #3506, opened 2024-02-20
Support subsurface/subview?


madsmtm:
The idea to support "subviews", i.e. a sort of limited `Window` that can be nested inside of `Window`, has floated around a few times, so I wanted to open an issue to discuss it.

A `Subview` would have the following platform equivalent (feel free to edit this):
- iOS: `UIView`
- macOS: `NSView`
- Wayland: ?
- Web: `HTMLCanvasElement`
- X11: ?
- Windows: `HWND` (I think?)
- Android: ?

Related: https://github.com/rust-windowing/winit/issues/696.
See also discussion in https://github.com/rust-windowing/winit/issues/3311.

madsmtm:
Personally, I think we should declare this as out of scope of Winit.

At least on macOS and iOS such subviews wouldn't really be useful on their own, we'd have to start allowing, and that's a slippery slope.

kchibisov:
Keep in mind that this is already present in a form of `child_window`, but the only reason it worked is because of `XID` and `HWND` being the same for window. However the API doesn't match what subviews should do, since obviously they can't fullscreen and they have a special way to receive events and resize.

So I guess this issue is more about removing hacky child window stuff and doing _proper_ subview.

>Personally, I think we should declare this as out of scope of Winit.

it's already in winit and they are very much required to do complex scene compositing/video players, etc. The same as mixing renderers.

ids1024:
> then we're basically becoming GTK

Having *some* way to do subsurfaces/subviews is pretty important for creating a general purpose toolkit like GTK on top of winit. So it seems important unless being a backend for a UI toolkit, or a video player is considered out of scope for winit. Or being used to create a video player, etc.

If it's not supported *in* winit, there would need to be some acceptable way at least to use platform-specific code to implement this on top of winit. Not sure how practical this would be currently on all of these platforms.

Actually using subsurfaces for things like hardware decoded video will inevitably be platform-specific... at least in the absence of libraries that help with that.

Is it possible to represent all of these kind of subviews as a `raw-window-handle` currently, or would we need changes there too?

kchibisov:
Basic subviews don't require anything special. We may have option to create special purpose subveiws, but in general the end goal is to have a way to a separate render target within a window to have video players, overlays drawn with software rendering, just transparent surfaces to dim main scene, etc.

And on macOS maybe some views that affect compositing which you can pass via the platform specific attributes.

madsmtm:
[...]

PR #3571, opened 2024-03-05
Soft keyboard input for iOS

simlay:

Related to #2260.


src/platform/ios.rs
```diff
@@ -100,6 +100,7 @@ pub trait WindowExtIOS {
     ///
     /// The default is to not recognize gestures.
     fn recognize_rotation_gesture(&self, should_recognize: bool);
+    fn set_keyboard_visible(&self, visible: bool);
```
kchibisov, line 103: We already have `set_ime_allowed`, so wire to it for now.

src/platform_impl/ios/uikit/text_field.rs
```diff
@@ -0,0 +1,56 @@
...
+    unsafe impl ClassType for UITextView {
+        #[inherits(NSObject)]
+        type Super = UIView;
+        type Mutability = mutability::InteriorMutable;
+    }
+);
+extern_methods!(
+    unsafe impl UITextView {
+        #[method(text)]
+        pub fn text(&self) -> &NSString;
```
madsmtm, line 21: Lifetime is bound to autorelease pool otherwise, which may be smaller than the lifetime of the UITextView.

```suggestion
        #[method_id(text)]
        pub fn text(&self) -> Id<NSString>;
```

src/platform_impl/ios/uikit/text_field.rs
```diff
@@ -0,0 +1,56 @@
...
+
+        #[method(setText:)]
+        pub fn setText(&self, text: &NSString);
+
+        #[method_id(@__retain_semantics Other delegate)]
```
madsmtm, line 26: `@__retain_semantics` is a compilation-time unstable optimization that's only intended to be used by `icrate`. It works fine without it.

```suggestion
        #[method_id(delegate)]
```

src/platform_impl/ios/text_field.rs
```diff
@@ -0,0 +1,149 @@
...
+    unsafe impl WinitTextField { }
+);
+extern_methods!(
+    unsafe impl WinitTextField {
+        // These are methods from UIResponder
+        #[method(becomeFirstResponder)]
+        pub fn focus(&self) -> bool;
+
+        #[method(resignFirstResponder)]
+        pub fn unfocus(&self) -> bool;
```
madsmtm, line 53: Should be moved to an `extern_methods!` on `UIResponder`.

src/platform_impl/ios/uikit/text_field.rs
```diff
@@ -0,0 +1,56 @@
...
+use objc2::rc::Id;
+use objc2::runtime::{NSObjectProtocol, ProtocolObject};
+use objc2::{extern_class, extern_methods, extern_protocol, mutability, ClassType, ProtocolType};
+
+extern_class!(
+    #[derive(Debug, PartialEq, Eq, Hash)]
+    pub(crate) struct UITextView;
+
+    unsafe impl ClassType for UITextView {
+        #[inherits(NSObject)]
```
madsmtm, line 13: Gives you access to more functionality on `UITextView`.

```suggestion
        #[inherits(UIResponder, NSObject)]
```

src/platform_impl/ios/text_field.rs
```diff
@@ -0,0 +1,149 @@
...
+
+        fn window(&self) -> Option<Id<WinitUIWindow>> {
+            unsafe { msg_send_id![self, window] }
+        }
```
madsmtm, line 57: I'd rather have this on `UIView` and return `UIWindow`, and then when the `WinitUIWindow` is needed, you'd use the method on `UIView`, but with an `Id::cast` to turn it into `WinitUIWindow`.

src/platform_impl/ios/text_field.rs
```diff
@@ -0,0 +1,149 @@
...
+    platform_impl::platform::DEVICE_ID,
+    window::{WindowAttributes, WindowId as RootWindowId},
+};
+
+pub struct WinitTextFieldState {
+    delegate: RefCell<Id<WinitTextFieldDelegate>>,
+}
+
+declare_class!(
+    pub(crate) struct WinitTextField;
```
madsmtm, line 31: I suspect you don't need to subclass `UITextField`, i.e. this class is unnecessary as long as you make sure to store the `WinitTextFieldDelegate` somewhere else (e.g. on `WinitView`, as done currently).

madsmtm (commented):
Have added a few comments wrt. `objc2` stuff.

Regarding the implementation, I'm unsure if this is the right approach? Adding a hidden view like this seems like it could get into a lot of trouble with event focus?

There's no other way to control the software keyboard?

(I suspect this will be easier to solve after something like https://github.com/rust-windowing/winit/issues/3506, but I'd be fine with this until then).

simlay:
> Regarding the implementation, I'm unsure if this is the right approach? Adding a hidden view like this seems like it could get into a lot of trouble with event focus?

After some research and thought, I don't think so but I'm not sure what the least-bad-way to do it is. I wrote about this in a chat but felt like it needed to on this issue.

I decided to look at how Flutter does it and how SwiftUI probably does it.

For Flutter: I created a simple flutter app with just a `TextField`, loaded it into my iOS simulator inspected the UI heirarchy in xcode, and it showed `FlutterTextInputView` which is [defined in the depths of flutter/engine repo](https://github.com/flutter/engine/blob/8a51e124fbf168295a1b05f8d66459bfb29ae8a9/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.h#L131-L174). It's unclear how one puts various UIViews (from the apple uikit framework) mixed in with skia rendering parts.

For SwiftUI, I created a simple view with a `TextView` and it's a `TextEditorTextView` in the UI hierarchy. This has some sub parts which are UIViews. From the relatively few results on a github search for `TextEditorTextView`, the one that fits is `SwiftUI.TextEditorTextView : UITextView` which say they're generated/reverse engineered with [RuntimeBrowser](https://github.com/nst/RuntimeBrowser/). So, it seems that `UITextView` mixed in similar to flutter.

I'd look into how [xamrin](https://dotnet.microsoft.com/en-us/apps/xamarin) does text input but given how old it is, I'm pretty sure it's using a `UITextView` or the like as well.

lucasmerlin:
In my fork of winit I implemented iOS keyboard input by just implementing UIKeyInput for the WinitView. This works pretty well, but it doesn't support autocomplete / autocorrect, for that to work we would need to implement UiTextInput as well. 

Here are the relevant changes:

https://github.com/lucasmerlin/winit/commit/0eec676386054911812434db7efbb5f7ae0ced55#diff-6540857eb4ce8f18ff73ad5a24513873e5572acb1ca867b055a462bdf8ea9e23

If this seems like a better way, I could open a PR

madsmtm:
@simlay: Yeah, why are we not just using [`UITextInput` and the related protocols](https://developer.apple.com/documentation/uikit/uitextinput?language=objc)? It really seems like `UITextView`/`UITextField` is the wrong approach here, since they also _display_ the text itself, which we in Winit want to leave as a task to the user.

simlay:
> @simlay: Yeah, why are we not just using [`UITextInput` and the related protocols](https://developer.apple.com/documentation/uikit/uitextinput?language=objc)? It really seems like `UITextView`/`UITextField` is the wrong approach here, since they also _display_ the text itself, which we in Winit want to leave as a task to the user.

Oh that was just because I was thinking this was how other frameworks did it. I think someone on matrix pointed out that react-native, flutter et al implement the `UITextInput` protocols.

> If this seems like a better way, I could open a PR

@lucasmerlin your changes seem more of the correct path forward. Like you said, the autocomplete part is tough. Does implementing the `UITextInput` protocols enable [moving the cursor of the input](https://www.youtube.com/watch?v=YmK6IJTM5PQ)?

lucasmerlin:
@simlay 
> Does implementing the UITextInput protocols enable [moving the cursor of the input](https://www.youtube.com/watch?v=YmK6IJTM5PQ)?
No Idea, but looking at the docs, I think that's what the [floating cursor apis](https://developer.apple.com/documentation/uikit/uitextinput#2858067) might be for? I have no idea how this would be handled by winit though. 

> @lucasmerlin your changes seem more of the correct path forward.
Feel free to fork / modify / copy the code from my fork if you want to keep working on this. I think it might make sense to open a PR that implements UiKeyInput with basic text input support (which should be fairly straightforward based on my changes) and then open another one that implements UiTextInput afterwards.

madsmtm:
I'm going to go with https://github.com/rust-windowing/winit/pull/3823 first, and then we can work from that to use the more feature-ful protocols to provide even better keyboard support. But thanks for starting on this work Sebastian!

PR #3576, opened 2024-03-09
Add DWMWA_CLOAK support on Windows

Neo-Zhixing:

Tested on Windows 11 build 22621.3155

### DIFF
--- a/src/platform/windows.rs
+++ b/src/platform/windows.rs
@@ -248,6 +248,11 @@ pub trait WindowExtWindows {
     ///
     /// Supported starting with Windows 11 Build 22000.
     fn set_corner_preference(&self, preference: CornerPreference);
+
+    /// Cloaks the window such that it is not visible to the user. The window is still composed by DWM.
+    ///
+    /// Not supported on Windows 7 and earlier.
+    fn set_cloaked(&self, cloaked: bool);
 }
 
 impl WindowExtWindows for Window {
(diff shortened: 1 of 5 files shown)


src/platform/windows.rs
```diff
@@ -248,6 +248,11 @@ pub trait WindowExtWindows {
     ///
     /// Supported starting with Windows 11 Build 22000.
     fn set_corner_preference(&self, preference: CornerPreference);
+
+    /// Cloaks the window such that it is not visible to the user. The window is still composed by DWM.
+    ///
+    /// Not supported on Windows 7 and earlier.
+    fn set_cloaked(&self, cloaked: bool);
```
kchibisov, line 255: That sounds like content protected? Or what it means here? Like it's not visible, but you can screen capture it? Like overlay, etc?

Neo-Zhixing: This description was copy pasted from the Windows docs, but basically `DWMWA_CLOAK` makes a window invisible.

This is explained in detail here: https://devblogs.microsoft.com/oldnewthing/20200302-00/?p=103507

> The shell has a concept called cloaking, in which a window is given all the trappings of visibility, without actually being presented to the user. As far as the app can tell, it seems to be visible: It still has the WS_VISIBLE window style, its coordinates are still within the bounds of the monitor, it still gets WM_PAINT messages, it has a non-empty clipping region, all the ways a traditional Win32 app has of detecting whether it is on screen say, “Yup, you’re on screen. Everything is just fine!”

From the user's perspective, an invisible window will not show up on the taskbar, but a cloaked window will.

riverar: Small nit: A cloaked window will not appear on Taskbar. Cloaking is particularly useful when you want a window to complete its UI layout, sizing, etc. without flickering in front of the user.

PR #3580, opened 2024-03-12
bugfix(dpi): `PhysicalUnit::to_logical` computation

kchibisov:
The conversion of PhysicalUnit was wrongly computed and the tests were not running on the CI for the dpi crate, thus it was missed, even though the test was correctly failing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -145,6 +145,13 @@ jobs:
     - name: Build crate
       run: cargo $CMD build $OPTIONS
 
+    # Test only on Linux x86_64, so we avoid spending unnecessary CI hours.
+    - name: Test dpi crate
+      if: >
+        contains(matrix.platform.name, 'Linux 64bit') &&
+        matrix.toolchain != '1.70.0'
+      run: cargo test -p dpi
+
     - name: Build tests
       if: >
         !contains(matrix.platform.target, 'redox') &&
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -145,6 +145,12 @@ jobs:
     - name: Build crate
       run: cargo $CMD build $OPTIONS
 
+    - name: Test dpi crate
+      if: >
+        contains(matrix.platform.target, 'darwin') &&
```
daxpedda, line 150: Why only on MacOS?

kchibisov: No platform specific code.

daxpedda: Why are we testing on MacOS and not on Linux?

kchibisov: it has only one job on CI, and linux has like 4.

notgull: It does seem kind of weird to test on macOS (where CI hours are expensive) and not on Linux (where CI is cheap). Maybe rig it so it only runs on Linux with both features enabled?

daxpedda: Would also be in favor of that.

kchibisov: >macOS (where CI hours are expensive) and not on Linux (where CI is cheap).

it's run anyway. I don't see any issue with that. it's also not like we're paying for any of that and 1 extra step is like nothing.

Like I can make it more complex (because you need way more checking), but why?

Also, if you want, you can change it yourself.

notgull: > Also, if you want, you can change it yourself.

Fixed!

daxpedda (approved):

PR #3586, opened 2024-03-13
Derive `Debug, Copy, Clone, PartialEq, Serialize, Deserialize` for `PixelUnit`

amrbashir:

---

Somehow I missed deriving those and didn't know until I tried replacing the `dpi` module in `tao` with this crate so I added a test to ensure these types will always be `Copy`.

If it is not too much to ask, could another release be made after merge?

### DIFF
--- a/dpi/src/lib.rs
+++ b/dpi/src/lib.rs
@@ -314,6 +314,8 @@ impl<P: Pixel> From<PhysicalUnit<P>> for f64 {
 }
 
 /// A pixel unit that's either physical or logical.
+#[derive(Debug, Copy, Clone, PartialEq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub enum PixelUnit {
     Physical(PhysicalUnit<i32>),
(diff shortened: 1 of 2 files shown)


dpi/src/lib.rs
```diff
@@ -1205,4 +1207,21 @@ mod tests {
...
+        is_copy::<PixelUnit>();
+
+        is_copy::<LogicalSize<i32>>();
+        is_copy::<PhysicalSize<f64>>();
+        is_copy::<Size>();
+
+        is_copy::<LogicalPosition<i32>>();
+        is_copy::<PhysicalPosition<f64>>();
+        is_copy::<Position>();
+    }
```
notgull, line 1226: Can you add a similar test for `PartialEq`?

amrbashir: Done

madsmtm (approved):

daxpedda (approved):

kchibisov (approved):

PR #3590, opened 2024-03-15
chore(changelog): use 'keep a changelog' format

kchibisov:
Links: https://keepachangelog.com/en/1.0.0/

--

I was thinking maybe we can try to _link_ the changed APIs, since it should be possible now?

I'm also not really liking the way the platform changes look. Maybe use a subsection for them? And for toplevel changes write with a `implemented on X, Y, and Z`?

### DIFF
--- a/src/changelog/unreleased.md
+++ b/src/changelog/unreleased.md
@@ -1,44 +1,91 @@
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
+
+The sections should follow the order `Added`, `Changed`, `Deprecated`,
+`Removed`, and `Fixed`.
+
+Platform specific changed should be added to the end of the section and grouped
+by platform name. Common API additions should have `, implemented` at the end
+for platforms where the API was initially implemented. See the following example
+on how to add them:
+
+```md
+### Added
+
+- Add `Window::turbo()`, implemented on X11, Wayland, and Web.
+- On X11, add `Window::some_rare_api`.
+- On X11, add `Window::even_more_rare_api`.
+- On Wayland, add `Window::common_api`.
+- On Windows, add `Window::some_rare_api`.
+```
+
 ## Unreleased
 
-- Deprecate `EventLoop::run` in favor of `EventLoop::run_app`.
-- Deprecate `EventLoopExtRunOnDemand::run_on_demand` in favor of `EventLoop::run_app_on_demand`.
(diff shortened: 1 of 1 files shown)


notgull (approved):
Looks good to me!

daxpedda (changes requested):
These have become pretty hard to read.

I do agree that now that we have headers for each change type, we shouldn't use the word "add" or "removed" in every entry, but it should still remain readable.

The main issue imo is starting sentences with "On Windows, ...", but then missing the verb, its really odd to read.

My suggestion would be to change the whole "On Windows, ..." to "**Windows: **".
Not a native speaker though, @notgull any thoughts?

kchibisov:
>My suggestion would be to change the whole "On Windows, ..." to "Windows: ".
Not a native speaker though, @notgull any thoughts?

Yeah, I don't like it as well as I said.

notgull:
I agree with adding the indicator before the actual addition.

kchibisov:
The alternative could be putting in the end of the sentence, which also makes it correct, or create sub sections, which is more clean.

daxpedda:
I'm in favor of sub sections.

notgull:
I'm fine with whichever option other maintainers prefer.

madsmtm (commented):
From a user perspective, I think it's more interesting to see (in order, and roughly in the groups that I think are relevant):
- Large changes or reworks.
- Minor additions, changes, removals or deprecations.
- Changed platform-specific behaviour, with one section per platform.

Re grammar: For projects that use Keep A Changelog, I appreciate when they still add the header name (or something more descriptive) to each entry (e.g. `- Added services menu` over `- Services menu`). Changelogs are for humans, they don't have to be the absolute minimal they can be.

But don't have that strong of a preference here.

daxpedda (approved):
LGTM!

I agree with @madsmtm that this is for humans and therefor should be "human-friendly" and not minimal.
Lets wait for our meeting before merging.

kchibisov:
I've added a separate commit as alternative way to structure changelog.

daxpedda (changes requested):
Now that we have sentences I think a dot at the end would be appropriate.

I missed the last meeting, did you decide to remove subsections or what happened here?

kchibisov:
@madsmtm suggested that we should write more correct sentences and subsections ones were harder to read, especially if you want to do inline migrations next to relevant change.

go ahead with fixing formatting as you like.

madsmtm (approved):
Fine with this now.

PR #3593, opened 2024-03-17
On Windows, only return the window handle if it is on the correct thread

notgull:
On Windows, it is generally unsafe to use the HWND outside of the thread
that it originates from. In reality, the HWND is an index into a
thread-local table, so using it outside of the GUI thread can result in
another window being used instead, following by code unsoundness. This
is why the WindowHandle type is !Send. However, Window is Send and Sync,
which means we have to account for this.

Thus far the best solution seems to be to check if we are not in the GUI
thread. If we aren't, refuse the return the window handle.

----

### DIFF
--- a/src/platform_impl/windows/window.rs
+++ b/src/platform_impl/windows/window.rs
@@ -365,7 +365,9 @@ impl Window {
 
     #[cfg(feature = "rwh_06")]
     #[inline]
-    pub fn raw_window_handle_rwh_06(&self) -> Result<rwh_06::RawWindowHandle, rwh_06::HandleError> {
+    pub unsafe fn rwh_06_no_thread_check(
+        &self,
+    ) -> Result<rwh_06::RawWindowHandle, rwh_06::HandleError> {
         let mut window_handle = rwh_06::Win32WindowHandle::new(unsafe {
             // SAFETY: Handle will never be zero.
             std::num::NonZeroIsize::new_unchecked(self.window)
@@ -373,6 +375,20 @@ impl Window {
         let hinstance = unsafe { super::get_window_long(self.hwnd(), GWLP_HINSTANCE) };
         window_handle.hinstance = std::num::NonZeroIsize::new(hinstance);
         Ok(rwh_06::RawWindowHandle::Win32(window_handle))
+    }
+
+    #[cfg(feature = "rwh_06")]
+    #[inline]
+    pub fn raw_window_handle_rwh_06(&self) -> Result<rwh_06::RawWindowHandle, rwh_06::HandleError> {
+        // TODO: Write a test once integration framework is ready to ensure that it holds.
+        // If we aren't in the GUI thread, we can't return the window.
+        if !self.thread_executor.in_event_loop_thread() {
+            tracing::error!("tried to access window handle outside of the main thread");
+            return Err(rwh_06::HandleError::Unavailable);
+        }
+
+        // SAFETY: We are on the correct thread.
+        unsafe { self.rwh_06_no_thread_check() }
     }
 
     #[cfg(feature = "rwh_06")]
--- a/src/changelog/unreleased.md
+++ b/src/changelog/unreleased.md
@@ -64,6 +64,7 @@ changelog entry.
 - On Windows, add `with_title_text_color`, and `with_corner_preference` on
   `WindowAttributesExtWindows`.
 - On Windows, implement resize increments.
+- On Windows, add `AnyThread` API to access window handle off the main thread.
 
 ### Changed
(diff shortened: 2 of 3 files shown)


kchibisov:
How does it behave with e.g. EGL where windows are `Send`(by the spec)? I guess it just creates and then the surface is being send to other thread, and the HWND itself is not touched during that?

src/platform_impl/windows/window.rs
```diff
@@ -382,6 +382,12 @@ impl Window {
     #[cfg(feature = "rwh_06")]
     #[inline]
     pub fn raw_window_handle_rwh_06(&self) -> Result<rwh_06::RawWindowHandle, rwh_06::HandleError> {
+        // If we aren't in the GUI thread, we can't return the window.
```
kchibisov, line 385: This should link to docs on why it's like that.

future_integration_tests/issue_3593.rs
```diff
@@ -0,0 +1,24 @@
+use winit::event_loop::EventLoop;
+use winit::window::Window;
+use rwh_06::HasWindowHandle;
+
+fn issue_3593() {
+    // We should not be able to get a window handle off of the main thread.
+    let event_loop = EventLoop::new().unwrap();
+    event_loop.run(|_, elwt| {
```
kchibisov, line 8: Don't use deprecated API.

kchibisov (changes requested):

msiglreith:
> On Windows, it is generally unsafe to use the HWND outside of the thread
that it originates from. In reality, the HWND is an index into a
thread-local table, so using it outside of the GUI thread can result in
another window being used instead, following by code unsoundness.

Is there a reference for this somewhere?

future_integration_tests/issue_3593.rs
notgull: Is there a better way to create a closure application ad-hoc?

notgull:
Generally windows are thread unsafe unless otherwise specified. For instance, [Raymond Chen](https://devblogs.microsoft.com/oldnewthing/20051010-09/?p=33843) says:

> Informally, one says that the thread “owns” the window. Messages are dispatched to a window procedure only on the thread that owns it, and generally speaking, modifications to a window should be made only from the thread that owns it. 

So `HWND`s are `!Send` but `Sync`... except not really, since the main thread can set values and cause race conditions in the higher level threads.

From the `raw-window-handle` docs:

> Users can safely assume that non-null/0 fields are valid handles, and it is up to the implementer of this trait to ensure that condition is upheld.

I'm saying that the definition of "valid" is "valid for the current context." As `HWND`s are largely useless outside of the thread that they originate from, it is invalid to issue these handles outside of the origin thread.

madsmtm:
CC https://github.com/rust-windowing/raw-window-handle/pull/154, where `Win32WindowHandle` will be marked `!Send` and `!Sync` (in the next breaking version of `raw-window-handle`).

kchibisov:
I'd really like some confirmation from `win32` docs/devs. Since the link is 20 years old, so maybe something changed or there's alternative.

notgull:
@Stehfyn Can you comment here?

msiglreith:
I can see `HWND` or a wrapper of it being `!Send`/`!Sync` in similar fashion as `ptr`.
I'm not agreeing on the thread-local argument, given one could also query it from another process for example and interact with it as it's rather an index into an OS managed array. Therefore, I also feel like we should artifically forbid access to it from another thread.

kchibisov:
I mean how OpenGL/Vulkan then operate where every surface resource  is `Send` between threads, and they build it from the `HWND`?

notgull:
> I mean how OpenGL/Vulkan then operate where every surface resource is `Send` between threads (by spec), and they build it from the `HWND`?

I've added a couple of escape hatches to allow for thread-safe usage. However for the default usage we shouldn't use too many threads.

future_integration_tests/issue_3593.rs
madsmtm: Not really, no. This would have to be done with `ApplicationHandler`:
```rust
struct App;
impl ApplicationHandler for App {
    fn new_events(&mut self, event_loop: &ActiveEventLoop, _cause: StartCause) {
        // Test code in here
    }
}

fn issue_3593() {
    let event_loop = EventLoop::new().unwrap();
    event_loop.run_app(App).unwrap();
}
```

I guess for the purpose of integration tests, we should probably have some sort of mechanism to do this more easily.

kchibisov (approved):
Given it just adds functionally, it should be good to go.

PR #3595, opened 2024-03-18
bugfix(x11): Fix window position calculation overflow during dragging

### LINKED ISSUE #3594: window_drag_resize fullscreen window crashing
### Description

The full-screen window was removed by sliding out to the left off the screen, then stretched from the bottom right corner, and then it crashed. like #3245 

window: PhysicalPosition { x: 0, y: 0 }, pointer: 2614 0
[2024-03-18T02:01:43Z WARN  wgpu_hal::vulkan] Suboptimal present of frame 0
window: PhysicalPosition { x: 0, y: 163 }, pointer: 2823 29
window: PhysicalPosition { x: 0, y: 143 }, pointer: 2801 22
window: PhysicalPosition { x: 0, y: 133 }, pointer: 2938 23
window: PhysicalPosition { x: -568, y: 135 }, pointer: 3836 1922
thread 'main' panicked at src/platform_impl/linux/x11/window.rs:1748:21:
attempt to add with overflow

### OS and window mananger

OS: Manjaro Linux x86_64
Resolution: 3840x2160
DE: Plasma 5.27.11 
WM: KWin

### Winit version

0.29.11

wannacu:
Closes #3594

### DIFF
--- a/src/platform_impl/linux/x11/window.rs
+++ b/src/platform_impl/linux/x11/window.rs
@@ -1789,8 +1789,8 @@ impl UnownedWindow {
                         | xproto::EventMask::SUBSTRUCTURE_NOTIFY,
                 ),
                 [
-                    (window.x as u32 + xinput_fp1616_to_float(pointer.win_x) as u32),
-                    (window.y as u32 + xinput_fp1616_to_float(pointer.win_y) as u32),
+                    (window.x + xinput_fp1616_to_float(pointer.win_x) as i32) as u32,
+                    (window.y + xinput_fp1616_to_float(pointer.win_y) as i32) as u32,
                     action.try_into().unwrap(),
                     1, // Button 1
                     1,


notgull (commented):
Is this tested?

src/platform_impl/linux/x11/window.rs
```diff
@@ -1789,8 +1789,8 @@ impl UnownedWindow {
                         | xproto::EventMask::SUBSTRUCTURE_NOTIFY,
                 ),
                 [
-                    (window.x as u32 + xinput_fp1616_to_float(pointer.win_x) as u32),
-                    (window.y as u32 + xinput_fp1616_to_float(pointer.win_y) as u32),
+                    (window.x + xinput_fp1616_to_float(pointer.win_x) as i32) as u32,
+                    (window.y + xinput_fp1616_to_float(pointer.win_y) as i32) as u32,
```
kchibisov, line 1793: Given that you can repro that, what is _negative_ here? Since I don't see how it can be done otherwise? If it's negative, why the message is `u32`? I guess it's `i32` but encoded as `u32`?

notgull: Yeah, this code doesn't make much sense to me. How does it work now when it didn't work before?

kchibisov: It did overflow when doing math operation because something was negative, then it just was interpreted as `u32`, but `as` just changes interpretation, it doesn't really do operations which have overflow checking.

psychon:
I managed to reproduce this issue (see #3594 for how, but it is quite easy) and then came up with the exact same patch. This calculates the position of the cursor in the root window and that will always be non-negative.

kchibisov (approved):
I'm not sure whether there should be a changelog, since I think in release it'll do things more or less correctly.

notgull (approved):

notgull:
Please rebase on the latest `master` so we can merge

PR #3597, opened 2024-03-19
Add UIGestureRecognizerDelegate and PanGestureRecogniser plus default…

jpedrick:
This PR adds:
* UIGestureRecognizerDelegate -- allows multiple simultaneous gesture recognition
  * Implements the default gesture recognizer delegate for iOS to recognize all gestures simultaneously
  * WinitView `- gestureRecognizer:shouldRecognizeSimultaneouslyWithGestureRecognizer:` -> true
* UIPanGestureRecognizer
* Modifies UIRotationGestureRecognizer handing code to work identically to macOS rotation(sends change in delta instead of velocity)
* Modifies UIPinchGestureRecognizer handing code to work identically to macOS magnify(sends change in delta instead of velocity)


src/changelog/unreleased.md
```diff
@@ -41,3 +41,4 @@
 - Add `Window::default_attributes` to get default `WindowAttributes`.
 - `log` has been replaced with `tracing`. The old behavior can be emulated by setting the `log` feature on the `tracing` crate.
 - On Windows, confine cursor to center of window when grabbed and hidden.
+- On iOS: For IOS/UIKit Support UIGestureRecognizerDelegate for simultaneous gesture input
```
madsmtm, line 44: Could you clean up the PR title, PR description and this changelog entry, being a bit more detailed on what this adds? Perhaps in more generic terms ("Add `PanGesture` window event and support it on iOS").

src/event.rs
```diff
@@ -317,6 +317,22 @@ pub enum WindowEvent {
...
+    ///
+    /// - Only available on **iOS**.
+    /// - On iOS, not recognized by default. It must be enabled when needed.
+    PanGesture {
+        device_id: DeviceId,
+        ///
+        /// This value may be NaN.
+        delta: PhysicalPosition<f32>,
+        velocity: PhysicalPosition<f32>,
         phase: TouchPhase,
```
madsmtm, line 336: Needs doc comments. Perhaps use part of the comment that you've written in the example?

src/event.rs
```diff
@@ -317,6 +317,22 @@ pub enum WindowEvent {
         ///
         /// This value may be NaN.
         delta: f64,
+        velocity: f32,
```
madsmtm, line 320: Needs doc comment, esp. to explain the difference between `delta` and `velocity` in this context.

src/event.rs
```diff
@@ -352,6 +368,7 @@ pub enum WindowEvent {
     RotationGesture {
         device_id: DeviceId,
         delta: f32,
+        velocity: f32,
```
madsmtm, line 371: Needs docs.

src/event.rs
```diff
@@ -1089,12 +1106,14 @@ mod tests {
                 with_window_event(PinchGesture {
                     device_id: did,
                     delta: 0.0,
+                    velocity: 0.0,
                     phase: event::TouchPhase::Started,
                 });
                 with_window_event(DoubleTapGesture { device_id: did });
```
madsmtm, line 1112: Nit: Add test of `PanGesture`

src/platform_impl/ios/uikit/gesture_recognizer.rs
```diff
@@ -19,6 +23,14 @@ extern_methods!(
     unsafe impl UIGestureRecognizer {
         #[method(state)]
         pub fn state(&self) -> UIGestureRecognizerState;
+
+        /// (delegate)[https://developer.apple.com/documentation/uikit/uigesturerecognizer/1624207-delegate?language=objc]
+        /// @property(nullable, nonatomic, weak) id<UIGestureRecognizerDelegate> delegate;
+        #[method(setDelegate:)]
+        pub fn set_delegate(&self, delegate: &ProtocolObject<dyn UIGestureRecognizerDelegate>);
```
madsmtm, line 30: Nit: Use `icrate` naming scheme (method is called `setDelegate`).

Same goes for the rest of this file.

src/platform_impl/ios/view.rs
```diff
@@ -238,6 +279,9 @@ declare_class!(
     }
 );
 
+unsafe impl NSObjectProtocol for WinitView {}
+unsafe impl UIGestureRecognizerDelegate for WinitView {}
```
madsmtm, line 283: Implement these inside of `declare_class!`, and move the implementation for `gestureRecognizer:shouldRecognizeSimultaneouslyWithGestureRecognizer:` inside of the impl for `UIGestureRecognizerDelegate`.

src/platform_impl/ios/view.rs
```diff
@@ -222,12 +230,45 @@ declare_class!(
             };
 
             // Flip the velocity to match macOS.
```
madsmtm, line 232: This is no longer done?

src/platform_impl/ios/uikit/gesture_recognizer.rs
```diff
@@ -119,3 +131,44 @@ extern_methods!(
...
+        pub fn set_translation_in_view(&self, translation: CGPoint, view: &UIView);
+
+        #[method(velocityInView:)]
+        pub fn velocity_in_view(&self, view: &UIView) -> CGPoint;
+    }
+);
+
+unsafe impl Encode for UIPanGestureRecognizer {
+    const ENCODING: Encoding = Encoding::Object;
+}
```
madsmtm, line 161: This should be unnecessary

madsmtm (changes requested):
Gave it a quick look, will maybe take a further look later

src/platform_impl/ios/view.rs
jpedrick: Actually, after testing some more, I'm bringing it back along with some additional translation work to make iOS pinch/rotation events match macOS

src/changelog/unreleased.md
madsmtm: Nit: Perhaps I'd rather have this merged with the "on iOS, add detection support for `PinchGesture`, `DoubleTapGesture` and `RotationGesture`" entry further above in this file; for users, it's not really relevant what modifications were made, since the release has yet to be cut.

src/event.rs
```diff
@@ -317,6 +317,25 @@ pub enum WindowEvent {
...
+    ///
+    /// - Only available on **iOS**.
+    /// - On iOS, not recognized by default. It must be enabled when needed.
+    PanGesture {
+        device_id: DeviceId,
+        /// Change in pixels of pan gesture from last update.
+        delta: PhysicalPosition<f32>,
+        /// Velocity of pan (delta/time?). Only available on **iOS**
+        /// https://developer.apple.com/documentation/uikit/uipangesturerecognizer/1621209-velocityinview?language=objc
+        velocity: PhysicalPosition<f32>,
```
madsmtm, line 338: As a developer using Winit, I'd still have no idea how to use this value. Maybe an example is in order? Or maybe you could update the example to use it in some (meaningful) way?

src/platform_impl/ios/uikit/gesture_recognizer.rs
```diff
@@ -119,3 +131,40 @@ extern_methods!(
...
+    pub(crate) unsafe trait UIGestureRecognizerDelegate: NSObjectProtocol {}
+
+    unsafe impl ProtocolType for dyn UIGestureRecognizerDelegate {
+        const NAME: &'static str = "UIGestureRecognizerDelegate";
+    }
+);
+
+unsafe impl Encode for dyn UIGestureRecognizerDelegate {
+    const ENCODING: Encoding = Encoding::Object;
+}
```
madsmtm, line 170: Should not be needed (and is incorrect).

src/platform_impl/ios/view.rs
```diff
@@ -26,6 +27,12 @@ pub struct WinitViewState {
     pinch_gesture_recognizer: RefCell<Option<Id<UIPinchGestureRecognizer>>>,
     doubletap_gesture_recognizer: RefCell<Option<Id<UITapGestureRecognizer>>>,
     rotation_gesture_recognizer: RefCell<Option<Id<UIRotationGestureRecognizer>>>,
+    pan_gesture_recognizer: RefCell<Option<Id<UIPanGestureRecognizer>>>,
+
+    // for iOS delta references the start of the Gesture
+    rotation_last_delta: RefCell<Option<f64>>,
+    pinch_last_delta: RefCell<Option<f64>>,
+    pan_last_delta: RefCell<Option<CGPoint>>,
```
madsmtm, line 35: Nit: Use `CGFloat` instead of `f64` (to support 32-bit iOS, which is currently the only way for me to test this on a real device).

Also, use `Cell` instead of `RefCell` here, that's both easier to work with, and more performant (but only possible because `Option<CGFloat>` is `Copy`).

And finally, maybe avoid the `Option` and just use `0.0`/`CGPoint::new(0.0, 0.0)` as the default value?

src/platform_impl/ios/view.rs
```diff
@@ -211,23 +235,105 @@ declare_class!(
...
-            // Flip the velocity to match macOS.
-            let delta = -recognizer.velocity() as _;
+
+            // Make delta and velocity negative to match macos, convert to degrees
             let gesture_event = EventWrapper::StaticEvent(Event::WindowEvent {
                 window_id: RootWindowId(window.id()),
                 event: WindowEvent::RotationGesture {
                     device_id: DEVICE_ID,
-                    delta,
+                    delta: -delta as f32 * 180.0 / std::f32::consts::PI,
```
madsmtm, line 274: Nit:

```suggestion
                    delta: (-delta).to_degrees(),
```

src/platform_impl/ios/view.rs
```diff
@@ -291,12 +412,28 @@ impl WinitView {
         }
     }
 
+    pub(crate) fn recognize_pan_gesture(&self, should_recognize: bool) {
+        if should_recognize {
+            if self.ivars().pan_gesture_recognizer.borrow().is_none() {
+                let pan: Id<UIPanGestureRecognizer> = unsafe {
+                    msg_send_id![UIPanGestureRecognizer::alloc(), initWithTarget: self, action: sel!(panGesture:)]
+                };
```
madsmtm, line 420: You've documented the event as a "two-finger pan gesture", maybe I'm misunderstanding, but would it make sense to set `minimumNumberOfTouches` and `maximumNumberOfTouches` to 2?

madsmtm (commented):
I like what you've done with storing the previous values to calculate the deltas, that makes a lot of sense to me.

src/event.rs
jpedrick: I believe it's used to trigger things like "inertial" scrolling. On most Apple UIs a fast pan will cause the document to continue scrolling with decaying velocity after the pan event is over.

I think it could be useful to game developers, for example a low velocity pan could be used to make a character walk, but a high velocity pan could make them jump. I don't particularly have an immediate use for it and I don't really want to develop a toy example. :)

Here's an example that implements dynamic scrolling by manually calculating the scroll velocity:
https://macresearch.org/dynamics-scrolling/

This API gives it for free, so I figured it saves some developers some effort.

src/platform_impl/ios/view.rs
jpedrick: @madsmtm sure I can change to CGFloat. I wonder though, there are many `f64`s and `u64`s in `src/event.rs`, so somehow those don't break your 32-bit iOS device?

I don't mind changing the `Option`s to zeros, but as a Rust noob my understanding was that `Option` is preferable for representing null values?

src/platform_impl/ios/view.rs
jpedrick: I agree, for the swipe gesture I needed to set min/max number of touches to disambiguate swiping from panning as well. I'll make it configurable on the recognize_pan_gesture function.

src/event.rs
jpedrick: I removed velocity from the events to simplify the PR. @madsmtm

src/platform_impl/ios/view.rs
madsmtm: 32-bit devices can _use_ `f64`/`u64`, it's just less performant. The issue is more that `CGFloat` is defined as `f32` on 32-bit devices, and `f64` on 64-bit devices, and it's what Apple's APIs return, so you need to either use `CGFloat` too, or do a cast to ensure that the compiler doesn't complain.

I think you can test that it compiles yourself with `cargo +nightly check -Zbuild-std --target=armv7s-apple-ios`.

src/platform_impl/ios/view.rs
```diff
@@ -167,12 +174,24 @@ declare_class!(
...
+                }
+                UIGestureRecognizerState::Ended => {
+                    let last_scale: f64 = self.ivars().pinch_last_delta.get();
+                    self.ivars().pinch_last_delta.set(0.0);
+                    (TouchPhase::Moved, recognizer.scale() - last_scale)
+                }
                 UIGestureRecognizerState::Cancelled | UIGestureRecognizerState::Failed => {
-                    TouchPhase::Cancelled
+                    self.ivars().rotation_last_delta.set(0.0);
+                    (TouchPhase::Cancelled, -recognizer.scale())
```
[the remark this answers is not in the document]
jpedrick: My thought is that since the event is canceled, the 'scale' would be undone by negating it.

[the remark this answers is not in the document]
madsmtm: Negating here probably makes sense then, thanks

madsmtm:
Thanks @jpedrick, and apologies that I left this behind for like a month

jpedrick:
> Thanks @jpedrick, and apologies that I left this behind for like a month

No worries, glad we got it merged!

Issue #3516, opened 2024-02-23
Sequence of mouse movements seems out of order when using trackpad on macOS


dmbfm:
I'm developing a drawing application and I noticed a weird artifact with mouse positions when drawing with a trackpad on macOS (I'm using a macbook M1 pro on macOS 14.3.1, and using winit 0.29.10). 

What I do is each frame I collect all of the incoming positions `CursorMoved` window event and store them in a vec. Then when it comes to draw the frame I connect these points with lines. 

When using a mouse everything works fine, but when using a trackpad I noticed that I got a lot of weird loops and spikes coming out of the drawn lines. For instance:

<img width="739" alt="Screenshot 2024-02-23 at 09 30 12" src="https://github.com/rust-windowing/winit/assets/8743306/28ba47ea-4bdd-45bd-9e7f-72ca82a5e43f">

In the image above different colors refer to different frames. Notice that the blue line does a sort of 'V' shape, so it appears the mouse locations are coming in out of order. 

By commenting out this line, https://github.com/rust-windowing/winit/blob/a127bd6f6684909852000cdf32469bac3b611887/src/platform_impl/macos/view.rs#L738
which then blocks the `pressureChange` event from emitting `CursorMoved` events, this artifacts go away and everything works correctly. 

It seems like the position from `pressureChange` is not equivalent from regular mouse positions, or maybe macOS does not process them in the same order, causing these loops to appear.

madsmtm:
Huh, in general, I find it odd the amount of places we're calling `mouse_motion`; I'd imagine they only really need calling inside `mouseMoved:`, `mouseDragged:`, `rightMouseDragged:` and `otherMouseDragged:`, it should be redundant everywhere else.

madsmtm:
This was added in https://github.com/rust-windowing/winit/pull/1490, but I can't seem to reproduce the original issue, it's possible that issue has been fixed in macOS itself in the meantime, and the workaround is unnecessary in newer versions?

madsmtm:
I tried debug-logging the data in the event sent to `pressureChangeWithEvent:`, but from my findings, it appears to be executed after `mouseMoved:`, and always with the same data as `mouseMoved:`. So while perhaps a bit inefficient, because we emit the same event twice, it doesn't seem like it should cause the issues you're seeing?

Can you provide a minimal example where you can reproduce the jittery behaviour? Maybe by trying one of the examples in this repo (likely `window`), click-drag your mouse across the window, and see if the output is jittery?

[...]

PR #3601, opened 2024-03-21
Consistently emit extra mouse motion events

madsmtm:
In particular, we don't want to emit those events inside of `pressureChangeWithEvent:`, since the mouse motion value is sometimes outdated.

Additionally, we want to ensure the events have been emitted during other gestures.

Fixes https://github.com/rust-windowing/winit/issues/3516.


kchibisov (approved):
I don't think we make much guarantees, but that should be fine.

Issue #3485, opened 2024-02-11
The window returns the original size after resize on Wayland


alberand:
Hi,
Thank you for your work on this great project. 

I did the following change to the `examples/resizable.rs`
```
diff --git a/examples/resizable.rs b/examples/resizable.rs
index 3bfc4b59..33ed87e9 100644
--- a/examples/resizable.rs
+++ b/examples/resizable.rs
@@ -43,6 +43,7 @@ fn main() -> Result<(), impl std::error::Error> {
                     resizable = !resizable;
                     println!("Resizable: {resizable}");
                     window.set_resizable(resizable);
+                    let _ = window.request_inner_size(LogicalSize::new(500.0, 500.0));
                 }
                 WindowEvent::RedrawRequested => {
                     fill::fill_window(&window);
```
Pressing **space** causes the window to resize to the  [500.0, 500.0] size. However, this window stays in this *new* size until it's unfocused. After focus is lost, the window gets back to its original size `LogicalSize::new(600.0, 300.0)`.

I debugged this a little, and it looks like that `winit` correctly handles `request_inner_size()` and resizes the window. However, the moment when `WindowState.configure()` is called (unfocus event), the `WindowConfigure.new_size` is still set to [600, 300] and `self.inner_size()` is [500, 500]. Therefore, the window gets resized back to the `new_size` [600, 300].

By doing switch to fullscreen and back to normal before unfocusing window, fixes the problem and window stays in the new requested size.

*My speculation is that `request_inner_size()` doesn't change the frame size? Or it's not properly communicated to wayland server?*

Platform:
Wayland Version: 1.22.0
Sway Version: 1.8.1
winit latest origin/master 4d4d6e5052877d9578bac73f52eaf323e9089998

On **space**:
```
Resizable: true
[ 104934.034]  -> xdg_toplevel@24.set_min_size(600, 300)
[ 104934.045]  -> xdg_toplevel@24.set_max_size(600, 300)
[ 104934.064]  -> wl_compositor@3.create_region(new id wl_region@33)
[ 104934.080]  -> wl_region@33.add(0, 0, 2147483647, 2147483647)
[ 104934.092]  -> wl_surface@22.set_opaque_region(wl_region@33)
[ 104934.098]  -> wl_region@33.destroy()
[ 104934.111]  -> xdg_surface@23.set_window_geometry(0, 0, 500, 500)
[ 104934.118]  -> wl_surface@22.commit()
[...]

PR #3602, opened 2024-03-22
Introduce ignore_size_suggestions() for Wayland

alberand:
On Wayland, size is entirely controlled by the user, however winit obeys to the resizes window manager gives us. On tiling window managers, such as Sway, this make it impossible to control window size programmatically with request_inner_size() as they don't always care about windows original size.

This patch introduces window.without_size_suggestions(bool) which tells window to ignore sizes suggested by window managers. Exceptions are when window is switched to fullscreen, maximized or tiled mode.

This is currently implemented only for Wayland.

More on roots of this https://github.com/rust-windowing/winit/issues/3485

- [x] Tested on all platforms changed (Wayland only)

### DIFF
--- a/examples/programmatical_resize.rs
+++ b/examples/programmatical_resize.rs
@@ -1,0 +1,56 @@
+#![allow(clippy::single_match)]
+
+use simple_logger::SimpleLogger;
+use winit::{
+    dpi::LogicalSize,
+    event::{ElementState, Event, KeyEvent, WindowEvent},
+    event_loop::EventLoop,
+    keyboard::{KeyCode, PhysicalKey},
+    window::Window,
+};
+
+#[path = "util/fill.rs"]
+mod fill;
+
+fn main() -> Result<(), impl std::error::Error> {
+    SimpleLogger::new().init().unwrap();
+    let event_loop = EventLoop::new().unwrap();
+
+    let min_size = LogicalSize::new(400.0, 200.0);
+    let max_size = LogicalSize::new(800.0, 400.0);
+    let mut size = min_size;
+
+    let window = Window::builder()
+        .with_title("Hit space to toggle size.")
+        .with_inner_size(size)
+        .with_min_inner_size(min_size)
+        .with_max_inner_size(max_size)
+        .without_size_suggestions(true)
+        .build(&event_loop)
+        .unwrap();
+
+    event_loop.run(move |event, elwt| {
+        if let Event::WindowEvent { event, .. } = event {
+            match event {
+                WindowEvent::CloseRequested => elwt.exit(),
+                WindowEvent::KeyboardInput {
+                    event:
+                        KeyEvent {
+                            physical_key: PhysicalKey::Code(KeyCode::Space),
+                            state: ElementState::Released,
+                            ..
+                        },
+                    ..
+                } => {
+                    size = if size == min_size { max_size } else { min_size };
+                    println!("New size: {:?}", size);
+                    let _ = window.request_inner_size(size);
+                }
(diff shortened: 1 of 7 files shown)


alberand:
Ops, I see that I will need to add more interface implementation here. Before continuing, I would like to ask if that's something what you would be interested in merging

kchibisov:
Just for you, sway will truncate you if you won't obey, and there's a proposed wording that you should use what compositor suggests for tiling.

So I'm not really a fun of that given that you just want a broken behavior and having window truncated and increasing load is really not what you want.

I'd suggest to consider to make your window float or use min/max size constraints.

kchibisov (changes requested):
See https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/258

And there's already a wording for maximized (there's also a discussion to make them less strict), but until that is resolved I'm against such changes..

alberand:
> Just for you, sway will truncate you if you won't obey, and there's a proposed wording that you should use what compositor suggests for tiling.
> 
> So I'm not really a fun of that given that you just want a broken behavior and having window truncated and increasing load is really not what you want.
> 
> I'd suggest to consider to make your window float or use min/max size constraints.

This resizing works only for floating windows. Tiled one are not affected. Not sure why you say it's broken behavior. 

What I'm trying to achieve is that request_inner_size(size) on floating window would resize the window and then stay in this new size until maximized/fullscreen/tiled/requested new size. I see it as expected behavior.

kchibisov:
`request_inner_size` already works like that....

alberand:
Not on Wayland Sway as I reported in that issue.

kchibisov:
See https://github.com/rust-windowing/winit/blob/master/src/platform_impl/linux/wayland/window/state.rs#L645 it checks for state on the latest configure, floating window is `stateless`, thus it'll resize and the new size will be returned from the `request_inner_size`, this will include the geometry changes.

kchibisov:
>Not on Wayland Sway as I reported in that issue.

IIRC sway just wants to resize you back, I'd suggest to fix the issue with sway, since it clearly shouldn't do so, also it wasn't doing so for me last time I checked. 

In general, you can do it yourself in the `Resized` even handle by just changing the size, so no need to patch winit for just that.

alberand:
Handling `Resized` event also doesn't work for me, as Sway will resize it back on the first `configure()` call.

I did this change for winit as I thought it could be handy for any other not well-behaved WM.

No worries, thanks for your time!

kchibisov:
Yeah, but I'm saying that you can ignore what sway tells you in `Resized` though, and just call `request_inner_size` and it'll just use what you want? Also, I'd be fine if we wire the `resizable` to `stateless` thing on Wayland, since you're really not resizable in `maximized`/`fullscreen`/`tiled`, thus when getting `Resized` you can resize back when you know that you're likely _float_.

But it's a separate patch.

kchibisov:
Also, you can adjust the logic of `resizable` to do what you did, since if compositor tries to send you size that is not equal to what you set it's likely doing something _weird_. So you can wire that instead of new API and it should be a bit more _expected_, I guess? 

Though, there's an issue with the maximize/fullscreen for it, which we may still obey...

alberand:
Yeah, make sense, thanks, I will probably go with creating a downstream patch with `resizable` (maybe also try to look into sway :))

kchibisov:
I mean one resizable patch will need in winit for sure, since the size is not _ignored_, we just assume that compositor behaves. I just don't like a _separate API_ for all of that to just ignore , but if it's tied to min/max size it's debatable who is right (compositor or client).

PR #3604, opened 2024-03-24
Silence clippy

notgull:
It appears that clippy's dead code detection has gotten better. This
commit fixes winit's code to match.

### DIFF
--- a/src/platform_impl/web/event_loop/state.rs
+++ b/src/platform_impl/web/event_loop/state.rs
@@ -6,7 +6,7 @@ use web_time::Instant;
 pub enum State {
     Init,
     WaitUntil {
-        timeout: backend::Schedule,
+        _timeout: backend::Schedule,
         start: Instant,
         end: Instant,
     },
@@ -14,7 +14,7 @@ pub enum State {
         start: Instant,
     },
     Poll {
-        request: backend::Schedule,
+        _request: backend::Schedule,
     },
     Exit,
 }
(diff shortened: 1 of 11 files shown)


kchibisov (changes requested):
I'd say that we shouldn't add that many `#[allow(dead_code)]` statements. Probably better to use other cfgs if we already adding one.

notgull:
@madsmtm The iOS failure in CI looks spurious to me, am I correct?

notgull:
> I'd say that we shouldn't add that many `#[allow(dead_code)]` statements. Probably better to use other cfgs if we already adding one.

In many cases we can't use `cfg(...)` guards to get around this issue, like for error types.

madsmtm:
> The iOS failure in CI looks spurious to me, am I correct?

I'd guess so

src/platform_impl/web/event_loop/state.rs
```diff
@@ -6,6 +6,7 @@ use web_time::Instant;
 pub enum State {
     Init,
     WaitUntil {
+        #[allow(dead_code)]
         timeout: backend::Schedule,
```
daxpedda, line 10:
```suggestion
        _timeout: backend::Schedule,
```
This is stored here for its drop behavior.

src/platform_impl/web/event_loop/state.rs
```diff
@@ -14,6 +15,7 @@ pub enum State {
         start: Instant,
     },
     Poll {
+        #[allow(dead_code)]
         request: backend::Schedule,
```
daxpedda, line 19:
```suggestion
        _request: backend::Schedule,
```

daxpedda (changes requested):

kchibisov (approved):

daxpedda (approved):

PR #3605, opened 2024-03-25
windows: remove github URLs from binaries

raggi:
There are really stupid AV rules out there which cause almost any program that contains github URLs to become marked as malware.

While those rules are spurious, they're years old, and AV vendors have a very poor reputation at appropriately following up with these problems. I've already mailed Defender about this weeks ago and they've yet to respond.

Removing these strings from the panic data present in the binary prevents binaries linking the dependency from being spuriously marked as Trojan:Win32/Wacatac.B!ml.

It might be better to use an URL shortener or similar, but I suspect that if someone actually hits these rare cases, googling the path will be more than sufficient to get someone where they need to be.

I have tested this change by producing a release build prior to this change, uploading it somewhere, downloading via firefox -> Defender fires. Apply patch, release build, upload, download, run -> only smart screen, no Defender.

I also have a version of this patch against the 0.29 branch here: https://github.com/raggi/winit/tree/raggi/remove-urls-0.29

I'm happy to update the patch to include changelog, but wanted to send it for feedback first, given the novelty of the change.

### DIFF
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -1245,7 +1245,7 @@ unsafe fn public_window_callback_inner(
 
                 let new_rect = if window_pos.flags & NOMOVE_OR_NOSIZE != 0 {
                     let cur_rect = util::WindowArea::Outer.get_rect(window)
-                        .expect("Unexpected GetWindowRect failure; please report this error to https://github.com/rust-windowing/winit");
+                        .expect("Unexpected GetWindowRect failure; please report this error to rust-windowing/winit");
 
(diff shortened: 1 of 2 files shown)


kchibisov (approved):

madsmtm (approved):
Also fine with this change.

I will say though, unless you add some sort of check for this in CI, we're bound to get another GH URL added in the future.

kchibisov:
These URLs were in winit for like a bunch of years, and I don't remember anyone adding URLs ever.

The URLs in comments are fine, just the code ones are not, but this is really stupid lint, because you can link other resources like that.

kchibisov:
Thanks.

PR #3609, opened 2024-03-28
On Windows, improve the performance for high polling rate mouse.

benkyoujouzu:
When using a high polling rate mouse on Windows, fast mouse motions result in numerous WM_NCHITTEST messages being generated. Processing each of these messages via DefWindowProc leads to noticeable frame rate drops across various applications (bevyengine/bevy#12679).

Given that winit doesn't utilize the return value from DefWindowProc for WM_NCHITTEST, we can bypass invoking DefWindowsProc for this message, thereby substantially boosting performance for high polling rate mice


benkyoujouzu:
It appears that the frame rate drop isn't caused by `DefWindowProcW` on `WM_NCHITTEST`, but rather by a non-zero return value of `public_window_callback` on the `WM_NCHITTEST` event. To illustrate, consider the following handling of `WM_NCHITTEST` in `public_window_callback_inner`, which also resolves the frame rate drop issue:

```
        WM_NCHITTEST => {
            let _ = unsafe { DefWindowProcW(window, msg, wparam, lparam) };
            result = ProcResult::Value(0);
        }
```

However, I'm uncertain about how the return value of `public_window_callback` is processed. Can someone help me with this?

PR #3622, opened 2024-04-06
More concise `Debug` output for `WindowId`.

waywardmonkeys:
Since this is a nested struct, it was printing
`"WindowId(WindowId(2834882))"` when just printing `"WindowId(2834882)"` would be sufficient.

### DIFF
--- a/src/changelog/unreleased.md
+++ b/src/changelog/unreleased.md
@@ -249,3 +249,4 @@ changelog entry.
 - On Web, fix setting cursor icon overriding cursor visibility.
 - On Windows, fix cursor not confined to center of window when grabbed and hidden.
 - On macOS, fix sequence of mouse events being out of order when dragging on the trackpad.
+- Debug formatting of `WindowId` is more concise.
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
I did try this out on macOS but don't have other platforms handy at the moment. I didn't think this was the sort of thing that you'd add a test for, but could do so.

madsmtm (changes requested):
Hmm, I've considered doing this myself, and it's a problem with the `Debug` outputs for all of the other public structs as well, but honestly, I'd rather just use the `derive`d version of this, it more accurately represents what the code is like.

notgull (approved):
I take the opposite position; we should avoid cluttering up the debug output.

waywardmonkeys:
@madsmtm You marked this as needing a change, but I'm not clear what the change would be (especially as you indicated you didn't want it).

kchibisov (approved):
Fine by me.

Though, I'd suggest to check other top-level types, since there're plenty of them which look like that.

In the long term, the nesting will go away entirely.

madsmtm (approved):
No code changes, I just didn't really want this. But if others are fine with it, then I won't block.

daxpedda (approved):

Issue #135, opened 2017-02-11
Support resize increments


siiptuo:
Basically this means restricting window size to multiples of some defined size. [X11](https://tronche.com/gui/x/xlib/ICC/client-to-window-manager/wm-normal-hints.html) and [OS X](https://developer.apple.com/reference/appkit/nswindow/1419390-resizeincrements) provide this feature. Windows doesn't seem to have support for this but it could be emulated with additional code. Not sure about Wayland.

This feature is probably only useful for terminal emulators like [Alacritty](https://github.com/jwilm/alacritty/issues/388). So is it worth to be added to the library?

tomaka:
It should be possible already to set both the min and max dimensions to the same size.

siiptuo:
At least for me, setting min and max dimensions to the same size simply creates an unresizable window. The desired behaviour would be the following: when increment size is for instance 20x10, allowed window sizes are 20x10, 40x10, 20x20, 40x20 etc.

francesca64:
I'd be more inclined to add this to `WinodwBuilderExt` than to try to emulate the feature on Windows, since that sounds like it could get messy easily.

Do you know if X11's resize increments actually work? It might sound like a silly question, but it wouldn't surprise me if WM support is poor.

siiptuo:
I'm unable to find any list, but based on my experience, resize increments are set by many X11 terminal emulators and supported by many (stacking) window managers.

francesca64:
Alright, thanks. This looks really easy to implement, so I can knock this out for both platforms after I finish the X11 refactor I'm working on.

siiptuo:
Sound great!

One related property that could also be supported is *base size*. It works something like this: `window_width = base_width + x * width_inc`. This is useful if the window has a toolbar or padding. Unfortunately only X11 seems to support it.

francesca64:
X11 implementation on this branch: https://github.com/francesca64/winit/tree/x11-resize-inc

I added `with_resize_increments(width_inc: i32, height_inc: i32)` and `with_base_size(base_width: i32, base_height: i32)` to `WindowBuilderExt`. If you don't specify base size, it defaults to the window dimensions.

siiptuo:
Thanks! This works as it should.

However base size should not default to the window dimensions because this basically sets the minimum size. Instead base size should not be set at all by default.

francesca64:
Gotcha, I've corrected that. #494

You can expect this to be in the next release, though I still need to implement this on macOS.

[...]

PR #3623, opened 2024-04-07
Windows: implement resize increments

tetrahedron341:
- [x] Tested on all platforms changed (windows)

Implements #135 on windows


madsmtm:
CC @Cthutu, does this work for you?

Cthutu:
I managed to emulate it with resizing during a resize event (which surprisingly worked well).  But I will try this when I get a chance.

msiglreith (approved):
thanks looks good, didn't notice any issues during testing

PR #3624, opened 2024-04-09
wayland: bump sctk-adwaita to 0.9.0

kchibisov:
This is a breaking change, because the system versions of the libraries used by sctk-adwaita were changed. Such changes cascade through all the deps, so all libraries using winit MUST do a breaking change.

--

Suggestions for how to handle such cascade breaking changes are much appreciated. For now I can only _warn_ in the changelog that you **must** do a breaking change as well to not break builds.


madsmtm (commented):
Yeah, I don't think there's really any other option with forcing a breaking change than what you're doing here

kchibisov:
The point is that it _cascades_ into downstream.

kchibisov:
I've reworded the changelog entry to be more clear on what this actually does.

PR #3625, opened 2024-04-10
wayland: fix CSD decorations glitch when closing

kchibisov:
In rare cases destroying subsurfaces before the main surface could result in a frame where the window is still shown, but decorations got hidden, right before the window itself disappears.

--

cc @Yalter

### DIFF
--- a/src/platform_impl/linux/wayland/window/state.rs
+++ b/src/platform_impl/linux/wayland/window/state.rs
@@ -57,9 +57,6 @@ pub struct WindowState {
     /// The connection to Wayland server.
     pub connection: Connection,
 
-    /// The window frame, which is created from the configure request.
-    frame: Option<WinitFrame>,
-
     /// The `Shm` to set cursor.
     pub shm: WlShm,
 
@@ -155,6 +152,13 @@ pub struct WindowState {
 
     /// The underlying SCTK window.
     pub window: Window,
+
+    // NOTE: The spec says that destroying parent(`window` in our case), will unmap the
+    // subsurfaces. Thus to achieve atomic unmap of the client, drop the decorations
+    // frame after the `window` is dropped. To achieve that we rely on rust's struct
+    // field drop order guarantees.
+    /// The window frame, which is created from the configure request.
+    frame: Option<WinitFrame>,
 }
 
 impl WindowState {
--- a/src/changelog/unreleased.md
+++ b/src/changelog/unreleased.md
@@ -253,3 +253,4 @@ changelog entry.
 - On Web, fix setting cursor icon overriding cursor visibility.
 - On Windows, fix cursor not confined to center of window when grabbed and hidden.
 - On macOS, fix sequence of mouse events being out of order when dragging on the trackpad.
+- On Wayland, fix decoration glitch on close with some compositors


YaLTeR:
Before:
```
[3332284.840] xdg_toplevel@30.close()
[3332284.981]  -> wl_buffer@52.destroy()
[3332285.000]  -> wl_shm_pool@51.destroy()
[3332285.323]  -> wl_buffer@54.destroy()
[3332285.335]  -> wl_shm_pool@53.destroy()
[3332285.787]  -> wl_subsurface@33.destroy()
[3332285.805]  -> wl_surface@32.destroy()
[3332285.824]  -> wl_subsurface@35.destroy()
[3332285.832]  -> wl_surface@34.destroy()
[3332285.846]  -> wl_subsurface@37.destroy()
[3332285.854]  -> wl_surface@36.destroy()
[3332285.869]  -> wl_subsurface@39.destroy()
[3332285.877]  -> wl_surface@38.destroy()
[3332285.886]  -> wl_subsurface@41.destroy()
[3332285.894]  -> wl_surface@40.destroy()
[3332285.904]  -> wl_shm_pool@31.destroy()
[3332286.139]  -> xdg_toplevel@30.destroy()
[3332286.155]  -> xdg_surface@29.destroy()
```
After:
```
[3386250.117] xdg_toplevel@30.close()
[3386250.238]  -> wl_buffer@52.destroy()
[3386250.256]  -> wl_shm_pool@51.destroy()
[3386250.450]  -> wl_buffer@54.destroy()
[3386250.457]  -> wl_shm_pool@53.destroy()
[3386250.784]  -> xdg_toplevel@30.destroy()
[3386250.799]  -> xdg_surface@29.destroy()
[3386250.810]  -> wl_surface@28.destroy()
[3386250.824]  -> wl_subsurface@33.destroy()
[3386250.830]  -> wl_surface@32.destroy()
[3386250.840]  -> wl_subsurface@35.destroy()
[3386250.846]  -> wl_surface@34.destroy()
[3386250.852]  -> wl_subsurface@37.destroy()
[3386250.857]  -> wl_surface@36.destroy()
[3386250.867]  -> wl_subsurface@39.destroy()
[3386250.872]  -> wl_surface@38.destroy()
[3386250.881]  -> wl_subsurface@41.destroy()
[3386250.888]  -> wl_surface@40.destroy()
[3386250.894]  -> wl_shm_pool@31.destroy()
```
Seems to fix it I guess.

YaLTeR:
Fixes it in a compositor too.

src/platform_impl/linux/wayland/window/state.rs
```diff
@@ -155,6 +152,12 @@ pub struct WindowState {
 
     /// The underlying SCTK window.
     pub window: Window,
+
+    // NOTE: The spec says that destroying parent(`window` in our case), will unmap the
+    // subsurfaces. Thus to achieve atomic unmap of the client, drop the decorations
+    // frame after the `window` is dropped.
+    /// The window frame, which is created from the configure request.
+    frame: Option<WinitFrame>,
```
notgull, line 160: Instead of moving it around, why not just add this to the `drop` method for this structure?

```rust
fn drop(&mut self) {
    self.frame = None;
}
```

notgull (changes requested):

src/platform_impl/linux/wayland/window/state.rs
kchibisov: You need a manual drop, to drop the `window` first.

notgull: Then we should wrap the `window` in an `Option` as well. I think it is fragile to rely on the field ordering here, as I believe the Rust team has said that it is not stable

kchibisov: Then you have to wrap all the objects like that.

kchibisov: Drop order is explicitly defined here, so I don't see a reason to use it. https://doc.rust-lang.org/reference/destructors.html

madsmtm: Zero-cost option would be to use `ManuallyDrop`, but the standard library [explicitly recommends against that](https://doc.rust-lang.org/std/mem/struct.ManuallyDrop.html#manuallydrop-and-drop-order), so I think that relying on drop order is the cleaner option here; it's subtle, but it's documented, safe and stable.

madsmtm (approved):

src/platform_impl/linux/wayland/window/state.rs
daxpedda: I think drop order is something that can easily be missed, so at an absolute minimum "drop order" should specifically be mentioned in the comment.

daxpedda (changes requested):

src/platform_impl/linux/wayland/window/state.rs
kchibisov: Should I really document how every bit in rust works, though? If it has a `NOTE:` comment it's already raising attention to at least look into `blame` on why it's like that.

daxpedda: I think a few words for a pretty unknown and non-obvious part of the specification is appropriate here.

notgull: Yeah drop order is probably the issue

notgull (approved):
Ack

daxpedda (approved):

PR #3634, opened 2024-04-17
Update `objc2` crates

madsmtm:
Changes relevant to Winit:
- `icrate` has been deprecated in favour of separate crates per framework, in our case `objc2-foundation` and `objc2-app-kit` (and in the future `objc2-ui-kit` on iOS).
- Moved `MainThreadMarker::run_on_main` to free-standing function `run_on_main`.
- Changed how features work, this should result in less code that we need to compile.
- Enums are now real structs instead of type-aliases and free constants.
  - I'm aware that some of the new enum structs have duplication in their names the renaming algorithm was just a quick first approximation.
  - Also aware that we really ought to implement `Bit*` operations on them, tracking that in https://github.com/madsmtm/objc2/issues/602.

If possible, then I'd like to get this in before `v0.30`, as I'd like to move users away from `icrate` as quickly as possible (it's harder for me to release fixes to that crate).


src/platform_impl/macos/view.rs
```diff
@@ -371,9 +374,13 @@ declare_class!(
             _actual_range: *mut NSRange,
         ) -> NSRect {
             trace_scope!("firstRectForCharacterRange:actualRange:");
-            let rect = dbg!(NSRect::new(self.ivars().ime_position.get(), self.ivars().ime_size.get()));
+            let rect = dbg!(NSRect::new(
```
kchibisov, line 377: Could you drop the `dbg` from here?

kchibisov (approved):
Looks good as long as generated code is the same.

Small nit, but it's an artifact from some other PR.

madsmtm:
> Looks good as long as generated code is the same.

Well... The generated code is different, the full diff can be viewed [here](https://github.com/madsmtm/objc2-generated/compare/v0.1.0...v0.2.0). But the _functionality_ is the same.

PR #3636, opened 2024-04-18
bugfix(android): Allow `Volume*` keys to be passed to the user

MarijnS95:
It seems that discussions around #2748 (cc @Hoodad) being a poor workaround that should rather be implemented properly/differently drowned out the fact that the workaround was implemented incorrectly to begin with.  By having the `if` inside the match arm, rather than /on/ the match arm the `keycode =>` path is /never/ used for `Volume*` keys and this event is thus never passed on to the user, defeating the purpose of calling `handle_volume_keys()`.  Only the `Handled`/`Unhandled` state is affected by the `ignore_volume_keys` workaround, as set by `EventLoopBuilderExtAndroid::handle_volume_keys()`.

Fix this by moving the `if` workaround on the match arm, so that disabling it via `EventLoopBuilderExtAndroid::handle_volume_keys()` causes `Volume*` events to be delivered to user code once again.

### DIFF
--- a/src/changelog/unreleased.md
+++ b/src/changelog/unreleased.md
@@ -255,3 +255,4 @@ changelog entry.
 - On Windows, fix cursor not confined to center of window when grabbed and hidden.
 - On macOS, fix sequence of mouse events being out of order when dragging on the trackpad.
 - On Wayland, fix decoration glitch on close with some compositors.
+- On Android, fix a regression introduced in #2748 to allow volume key events to be received again.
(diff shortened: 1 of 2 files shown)


madsmtm (approved):

kchibisov (changes requested):
Missing a changelog entry.

Hoodad:
Took me a while to see, but yes this should allow the users to react to input 👍 Thanks for the fix!

kchibisov (approved):

PR #3637, opened 2024-04-18
Re-introduce Web examples

### LINKED ISSUE #3473: Merge web example into `window` example
Follow-up to https://github.com/rust-windowing/winit/pull/3447. Also, remember to update the platform-specific documentation on how you run examples on the web.

daxpedda:
This adds back the custom cursor stuff to the examples.
The only other example I didn't port was the scale factor one, which I didn't feel make a whole lot sense after #2859.
The `web` example only showed events inside the window, which we don't need anymore if we print them properly in the console, which I've done now.

As part of showing thing properly in the console, I've switched to using `tracing` instead of `println!` in the relevant examples.

I didn't find any mention of how to run examples on any target, e.g. Android or iOS.
So I feel we either have to add docs for all of these targets or none. Ergo I'm also tending towards removing our `run-wasm` dependency.
WDYT?

Fixes #3473.


examples/window.rs
```diff
@@ -670,9 +695,58 @@ impl WindowState {
...
+                custom_cursors[1].clone(),
+                event_loop.create_custom_cursor(CustomCursor::from_url(
+                    format!(
+                        "https://picsum.photos/128?random={}",
+                        URL_COUNTER.fetch_add(1, Ordering::Relaxed)
+                    ),
+                    64,
+                    64,
+                )),
+            ],
```
kchibisov, line 739: Seems like it's better to not inline it?

examples/window.rs
```diff
@@ -670,9 +695,58 @@ impl WindowState {
...
+        use winit::platform::web::CustomCursorExtWebSys;
+
+        let cursor = CustomCursor::from_animation(
+            Duration::from_secs(3),
+            vec![
+                custom_cursors[0].clone(),
+                custom_cursors[1].clone(),
+                event_loop.create_custom_cursor(CustomCursor::from_url(
+                    format!(
+                        "https://picsum.photos/128?random={}",
```
kchibisov, line 733: Could we use url to github for that? Like we have cursors laying around in the repo already in png formats, so they should be fine.

run-wasm/src/main.rs
```diff
@@ -1,3 +1,3 @@
 fn main() {
-    cargo_run_wasm::run_wasm_with_css("body { margin: 0px; }");
+    cargo_run_wasm::run_wasm_with_css(include_str!("style.css"));
```
kchibisov, line 2: Could probably just inline the `css` into this file, so it's a bit more easier to grasp.

examples/window.rs
```diff
@@ -983,6 +1063,14 @@ const KEY_BINDINGS: &[Binding<&'static str>] = &[
     // C.
     Binding::new("C", ModifiersState::CONTROL, Action::NextCursor),
     Binding::new("C", ModifiersState::ALT, Action::NextCustomCursor),
+    #[cfg(web_platform)]
+    Binding::new("C", ModifiersState::SHIFT, Action::UrlCustomCursor),
```
kchibisov, line 1067: We generally don't use `SHIFT` as you can see, so you can type more or less normally.

examples/window.rs
```diff
@@ -983,6 +1063,14 @@ const KEY_BINDINGS: &[Binding<&'static str>] = &[
...
+    #[cfg(web_platform)]
+    Binding::new(
+        "C",
+        ModifiersState::CONTROL.union(ModifiersState::SHIFT),
```
kchibisov, line 1071: If you want to combine, use `|`, it's more clear.

examples/window.rs
```diff
@@ -670,9 +695,58 @@ impl WindowState {
...
+        use winit::platform::web::CustomCursorExtWebSys;
+
+        let cursor = CustomCursor::from_url(
+            format!(
+                "https://picsum.photos/128?random={}",
+                URL_COUNTER.fetch_add(1, Ordering::Relaxed)
+            ),
+            64,
+            64,
+        );
```
kchibisov, line 710: That thing looks like duplicated in the `animation_cursor` as well, so should likely extract it.

examples/util/tracing.rs
```diff
@@ -0,0 +1,29 @@
...
+        tracing_subscriber::registry()
+            .with(
+                tracing_subscriber::fmt::layer()
+                    .with_ansi(false)
+                    .without_time()
+                    .with_writer(tracing_web::MakeWebConsoleWriter::new()),
+            )
+            .init();
+    }
+}
```
kchibisov, line 29: Could you `cfg` the `init` itself?

examples/window.rs
```diff
@@ -1,13 +1,18 @@
 //! Simple winit application.
 
+#[path = "util/tracing.rs"]
+mod tracing;
+
+use ::tracing::{error, info};
```
kchibisov, line 6: move to external crates.

examples/window.rs
```diff
@@ -1,13 +1,18 @@
 //! Simple winit application.
 
+#[path = "util/tracing.rs"]
+mod tracing;
```
kchibisov, line 4: move under all the `use`

kchibisov (changes requested):

examples/window.rs
```diff
@@ -501,6 +523,9 @@ struct WindowState {
     cursor_hidden: bool,
 }
 
+#[cfg(web_platform)]
+static URL_COUNTER: AtomicU64 = AtomicU64::new(0);
```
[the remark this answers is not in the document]
daxpedda: I did this on purpose because its not cross-platform, but I don't mind putting it at the top.

examples/window.rs
daxpedda: Well the point is to always have it be a different picture, so you can see what happens when its not loading immediately.

run-wasm/src/main.rs
daxpedda: I un-inlined it on purpose so it can be properly formatted and color coded by an IDE, otherwise its just an ugly string.

examples/window.rs
daxpedda: Any alternative suggestion?
I didn't want to move away from the "C" key.

examples/window.rs
daxpedda: Unfortunately that's not possible in a `const` context.

examples/util/tracing.rs
daxpedda: Sounds good!

examples/window.rs
daxpedda: Done.

examples/window.rs
daxpedda: Done.

examples/window.rs
kchibisov: `ALT | SHIFT` should be fine, just don't use plain `SHIFT`

examples/window.rs
kchibisov: I mean, having a function that returns you the thing is not a big of a deal, so that way the URL stays the same for both places for sure. It's less code in the end.

run-wasm/src/main.rs
daxpedda: I'm unsure what the disadvantage is of a separate file here, so the syntax highlighting seems like a worthwhile advantage to me.

examples/window.rs
daxpedda: Any alternative suggestion to the plain `SHIFT`?

run-wasm/src/main.rs
daxpedda: That seems like a very fine trade-off to me.

examples/window.rs
daxpedda: `CONTROL` is taken.
`SUPER` is probably not something we want here, browsers will react funny.
I took `ALT`.

kchibisov (approved):

madsmtm (approved):
> I'm also tending towards removing our `run-wasm` dependency.

CC @rukai

daxpedda:
> > I'm also tending towards removing our `run-wasm` dependency.

I just went ahead and did that, we can consider re-introducing it if we decide we want it after all.

rukai:
For the web example drawing events to the dom: that was valuable for testing the examples on mobile devices which do not have a console accessible in their browsers.

For run-wasm, if there's an alternative way you want users to run wasm examples then by all means swap to that. But currently there is no documented way to run wasm examples so I'm quite confused.

daxpedda:
> For the web example drawing events to the dom: that was valuable for testing the examples on mobile devices which do not have a console accessible in their browsers.

Good point!

> For run-wasm, if there's an alternative way you want users to run wasm examples then by all means swap to that. But currently there is no documented way to run wasm examples so I'm quite confused.
> This all or nothing approach to documentation is really strange, it doesn't have to be perfect but surely we should support the major platforms well with documentation?

My original comment:
> So I feel we either have to add docs for all of these targets or none. Ergo I'm also tending towards removing our `run-wasm` dependency.

Agreed, this is stupid. I think this was just my first thought: if I'm gonna start adding documentation, where is the documentation for the other targets?

`run-wasm` is not required to run the Wasm examples, there are plenty of Rust Wasm runners out there that can be configured globally, e.g. I'm using [`wasm-server-runner`](https://crates.io/crates/wasm-server-runner). So I don't need any additional work at my local Winit repo to run the examples.

I'm aware of the advantages `run-wasm` offers in comparison to other runners and I don't think Winit really needs them. This is why I decided to remove `run-wasm`, in addition to it not being documented.

Though I'm happy to re-add them upon popular request.

Apologies for not originally stating my full reasons in the OP.

rukai:
All good.
I agree that cargo-run-wasm requires a certain level of integration with the repo that isn't ideal. I think the solution here is to just point users towards wasm-server-runner in the readme.

daxpedda:
Would stick with keeping it in the README, pending input from the other maintainers.

Happy to look at a PR!

madsmtm:
I think I'd prefer `winit::platform::web`, and then perhaps instead link more prominently to that from `lib.rs`

daxpedda:
> I think I'd prefer it in `winit::platform::web`, and then perhaps instead link more prominently to that from `lib.rs`

I think it definitely needs something visible from the repo, when you clone the repo you don't automatically run `cargo doc`, where usually "repo documentation" isn't found.

madsmtm:
I'd also be fine with a `examples/README.md` with instructions specific to the examples

PR #3639, opened 2024-04-18
windows: bump windows-sys to 0.52

kchibisov:

### DIFF
--- a/src/platform_impl/windows/dpi.rs
+++ b/src/platform_impl/windows/dpi.rs
@@ -105,7 +105,7 @@ pub unsafe fn hwnd_dpi(hwnd: HWND) -> u32 {
         if unsafe { IsProcessDPIAware() } != false.into() {
             // If the process is DPI aware, then scaling must be handled by the application using
             // this DPI value.
-            unsafe { GetDeviceCaps(hdc, LOGPIXELSX) as u32 }
+            unsafe { GetDeviceCaps(hdc, LOGPIXELSX as i32) as u32 }
         } else {
             // If the process is DPI unaware, then scaling is performed by the OS; we thus return
             // 96 (scale factor 1.0) to prevent the window from being re-scaled by both the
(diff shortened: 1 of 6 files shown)


madsmtm (approved):
Annoying that we have to do the `as u32` in a few places, but I think that's just how the Win32 API is, e.g. [`DwmSetWindowAttribute`](https://learn.microsoft.com/en-us/windows/win32/api/dwmapi/nf-dwmapi-dwmsetwindowattribute) takes a `DWORD` so there isn't much that `windows-rs` can do here.

kchibisov:
>Additionally, HGLOBAL changed from isize to *mut c_void, I assume that's intentional too, at least that's what I'm getting from reading https://github.com/microsoft/windows-rs/issues/1643#issuecomment-1081896209.

Yeah, it's just I'm not sure if all of that is _sound_ now. Like I understood that it's like that.

notgull (approved):

src/platform_impl/windows/dpi.rs
```diff
@@ -105,7 +105,7 @@ pub unsafe fn hwnd_dpi(hwnd: HWND) -> u32 {
         if unsafe { IsProcessDPIAware() } != false.into() {
             // If the process is DPI aware, then scaling must be handled by the application using
             // this DPI value.
-            unsafe { GetDeviceCaps(hdc, LOGPIXELSX) as u32 }
+            unsafe { GetDeviceCaps(hdc, LOGPIXELSX as i32) as u32 }
```
MarijnS95, line 108: We should ask for upstream clarification about these. At least in the `windows` crate where these are newtyped, the underlying ABI is "kept in sync" (IIRC by casting in the wrapper implementation).

MarijnS95 (approved):

src/platform_impl/windows/dpi.rs
kchibisov: Given that casting small u32 to positive i32 have the same binary repr, it's safe to do it here.

In general, a lot of casts seems to be that way because sometimes they accept `enum` but `enum` itself is `i32` and not `u32`, and so on.

You can still ask though.

MarijnS95: Not saying that it's unsafe, just that it's inconvenient after these Windows crates already have all the type information they need; looks like it's all reserved for the high-level `windows` crate though.

PR #3641, opened 2024-04-18
Update dev-dependencies

daxpedda:
Update `image` and `softbuffer`.

### DIFF
--- a/examples/window.rs
+++ b/examples/window.rs
@@ -7,11 +7,12 @@ use std::fmt::Debug;
 use std::mem;
 #[cfg(not(any(android_platform, ios_platform)))]
 use std::num::NonZeroU32;
+use std::sync::Arc;
 
 use ::tracing::{error, info};
 use cursor_icon::CursorIcon;
 #[cfg(not(any(android_platform, ios_platform)))]
-use rwh_05::HasRawDisplayHandle;
+use rwh_06::{DisplayHandle, HasDisplayHandle};
 #[cfg(not(any(android_platform, ios_platform)))]
 use softbuffer::{Context, Surface};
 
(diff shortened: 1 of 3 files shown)


examples/window.rs
```diff
@@ -83,14 +84,21 @@ struct Application {
...
     fn new<T>(event_loop: &EventLoop<T>) -> Self {
         // SAFETY: we drop the context right before the event loop is stopped, thus making it safe.
         #[cfg(not(any(android_platform, ios_platform)))]
-        let context = Some(unsafe { Context::from_raw(event_loop.raw_display_handle()).unwrap() });
+        let context = Some(
+            Context::new(unsafe {
+                std::mem::transmute::<DisplayHandle<'_>, DisplayHandle<'static>>(
+                    event_loop.display_handle().unwrap(),
+                )
+            })
```
daxpedda, line 99: We could get rid of this last `unsafe` by just creating a `Context` for each `Window`.

kchibisov: we can probably create a context before the `state`, since the `loop` is ready?

I don't like context per window given that it creates a lot of resources, like sync queues, etc.

daxpedda: > we can probably create a context before the `state`, since the `loop` is ready?

No, because `run_app` consumes `EventLoop`.

> I don't like context per window given that it creates a lot of resources, like sync queues, etc.

Lets just leave it as is then. I guess in the future it would be nice to introduce some lifetime similar to [`std::thread::scope`](https://doc.rust-lang.org/1.77.2/std/thread/fn.scope.html) to fix issues like that.

kchibisov: `Arc` solves it if it matters.

daxpedda: Solves what exactly and how?

[`ApplicationHandler::window_event()`](https://rust-windowing.github.io/winit/winit/application/trait.ApplicationHandler.html#tymethod.window_event) gives you a `&ActiveEventLoop`, so the lifetime in `DisplayHandle` is bound to that `&` in `ActiveEventLoop`.

kchibisov: In the way we've discussed before for `WindowHandle` but just make a `DisplayHandle` like that as well.

But yeah, we can not solve it right now and it's _fine_.

madsmtm (approved):

PR #3649, opened 2024-04-25
Add remote notifications app delegates

extrawurst:
I need some input here :)

I am not sure about the way to bubble up the events. For now i added a new one entirely but that feels wrong. do you think it should rather be a DeviceEvent cause it is for sure not a window event, right?

Wtf, something went wrong with building this on top of `0.29.x`


extrawurst:
see https://github.com/rust-windowing/winit/pull/3650

PR #3651, opened 2024-04-26
Fix clippy::legacy_numeric_consts lints.

waywardmonkeys:
Prefer to use the associated constants on the types rather than the legacy items like `std::<type>::MAX` or functions like `max_value`. These legacy items are marked to indicate that they will be deprecated in a future version of Rust.

### DIFF
--- a/src/platform_impl/ios/app_state.rs
+++ b/src/platform_impl/ios/app_state.rs
@@ -753,7 +753,7 @@ impl EventLoopWaker {
             // future, but that gets changed to fire immediately in did_finish_launching
             let timer = CFRunLoopTimerCreate(
                 ptr::null_mut(),
-                std::f64::MAX,
+                f64::MAX,
                 0.000_000_1,
                 0,
                 0,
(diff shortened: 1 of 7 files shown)


kchibisov:
I think this lint is not enabled by clippy by default or yet? We'd need it breaking on CI to catch stuff like that in the future and it should be on stable clippy as well.

waywardmonkeys:
> I think this lint is not enabled by clippy by default or yet? We'd need it breaking on CI to catch stuff like that in the future and it should be on stable clippy as well.

It is coming. It is present in clippy but not enabled by default in `stable`, but is enabled in `nightly` and should be making its way to `stable` ...

I've rebased the changes to fix the merge conflict and force pushed it.

madsmtm (approved):

kchibisov:
I'm just saying that without automatic way to check for that there's no way we ensure that it'll stay the same.

In general, if it'll come to stable, we'll fix our CI with these.

kchibisov (changes requested):
See above, I generally don't mind, but it's a style change which one can easily reentroduce without a lint in place, so should be done when lint is _ready_.

waywardmonkeys:
So I should close this and not waste any more time?

kchibisov:
Yes, because the lint is not stable yet and once it's stable we'll fix it along other new lints.

PR #3653, opened 2024-04-26
On Android, bump to `ndk 0.9.0` and `android-activity 0.6.0`

MarijnS95:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -102,9 +102,8 @@ softbuffer = { version = "0.4.0", default-features = false, features = [
 ] }
 
 [target.'cfg(target_os = "android")'.dependencies]
-android-activity = "0.5.0"
-ndk = { version = "0.8.0", default-features = false }
-ndk-sys = "0.5.0"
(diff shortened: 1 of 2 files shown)


src/platform/android.rs
```diff
@@ -19,6 +19,7 @@
 //!
 //! | winit |       ndk-glue               |
 //! | :---: | :--------------------------: |
+//! | 0.30  | `android-activity = "0.6"`   |
```
kchibisov, line 22: Could you drop changes to this file, because I usually do them along the release bump, since there's no `0.30.0` version yet.

kchibisov (changes requested):

src/platform/android.rs
MarijnS95: Do you want me to drop this entirely, or insert a placeholder so that android-activity 0.6 is at least mentioned?

kchibisov: just drop all the thing.

kchibisov (approved):

PR #3654, opened 2024-04-26
chore: ensure that `.cargo` config is not published

kchibisov:
Just in case, so the correct changelog will be rendered when pulling the crate from the crates.io as archive and trying to build it.


madsmtm (approved):
I'm not sure this is necessary, but sure

PR #3655, opened 2024-04-26
Winit version 0.30.0

kchibisov:
--

Should be fine after these:

- [x] https://github.com/rust-windowing/winit/pull/3653
- [x] https://github.com/rust-windowing/winit/pull/3654
- [x] https://github.com/rust-windowing/winit/pull/3597


src/changelog/unreleased.md
MarijnS95, line 1: It's a tad unfortunate that we loose `git blame` info on this file, but it's probably not used very often anyway.

MarijnS95 (approved):

src/changelog/unreleased.md
kchibisov: It was kind of broken even before when I was moving changelog entries around.

linkmauve:
Reading the changelog I wrote #3656, so that the new gestures support more than just two platforms. I’ve found that the unit of the rotation gesture isn’t specified, the Wayland protocol exposes it in degrees but I don’t know what Apple does, it would be useful to specify that before the release.

kchibisov:
The gestures are present for a very long time, like 2-3 years, it's just they were renamed a bit this release, so while specifying them is good, they were always meant for macOS only and the metrics are the same as in all other releases.

src/changelog/unreleased.md
madsmtm: Would it help if you did it in two commits, and then rebase merge the PR?

linkmauve:
Ah you’re right, they were already present in the previous release, but the unit is still unspecified [in the documentation](https://docs.rs/winit/latest/winit/event/enum.WindowEvent.html#variant.TouchpadRotate), so I still don’t know what it’s meant to be even on macOS. Or do you mean it’s a unit/event specific to macOS and shouldn’t be tried to be matched on other windowing systems?

src/changelog/unreleased.md
kchibisov: It doesn't matter how it's done, because it's moved anyway.

daxpedda (approved):

madsmtm:
The unit of rotation is documented in https://github.com/rust-windowing/winit/pull/3597 to be degrees.

madsmtm (approved):

PR #3656, opened 2024-04-27
wayland: Add support for PinchGesture and RotationGesture

linkmauve:
These two events are synthesized from the same event from the same protocol, [zwp_pointer_gestures_v1](https://wayland.app/protocols/pointer-gestures-unstable-v1), and will always come together.

I’ve left over the third value, which is a `dx`/`dy` from the center of the gesture, it could be exposed as a 2D scroll but I feel like this should be a different event than the normal `AxisMotion`.

The documentation doesn’t indicate the unit of the rotation, so I’ve left it as degrees.  I don’t have any Apple OS where I could test whether that event is also in degrees or in radians there.

The other two gestures supported by this protocol, a multi-finger swipe, and a multi-finger hold, aren’t currently matched with any event in winit, would it make sense to expose them?

I’ve only tested on GNOME, but I expect other compositors to expose the same protocol eventually.

The last two items aren’t needed, as they were already added for the other two platforms which support this feature.

### DIFF
--- a/src/platform_impl/linux/wayland/seat/pointer/mod.rs
+++ b/src/platform_impl/linux/wayland/seat/pointer/mod.rs
@@ -14,6 +14,8 @@ use sctk::reexports::protocols::wp::pointer_constraints::zv1::client::zwp_locked
 use sctk::reexports::protocols::wp::cursor_shape::v1::client::wp_cursor_shape_device_v1::WpCursorShapeDeviceV1;
 use sctk::reexports::protocols::wp::cursor_shape::v1::client::wp_cursor_shape_manager_v1::WpCursorShapeManagerV1;
 use sctk::reexports::protocols::wp::pointer_constraints::zv1::client::zwp_pointer_constraints_v1::{Lifetime, ZwpPointerConstraintsV1};
+use sctk::reexports::protocols::wp::pointer_gestures::zv1::client::zwp_pointer_gestures_v1::ZwpPointerGesturesV1;
+use sctk::reexports::protocols::wp::pointer_gestures::zv1::client::zwp_pointer_gesture_pinch_v1::{ZwpPointerGesturePinchV1, Event};
 use sctk::reexports::client::globals::{BindError, GlobalList};
 use sctk::reexports::csd_frame::FrameClick;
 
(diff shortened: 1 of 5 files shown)


madsmtm:
The rotation delta is in degrees on macOS, and that's what the iOS impl tries to match as well.

(Though we should perhaps change it to radians, that's the morally superior unit).

linkmauve:
Thanks!  Now that this other PR is merged, I’ve also implemented `PanGesture` during the pinch.

src/platform_impl/linux/wayland/seat/pointer/mod.rs
```diff
@@ -467,9 +469,107 @@ impl Dispatch<WpCursorShapeManagerV1, GlobalData, WinitState> for SeatState {
...
+                };
+
+                let pan_delta = PhysicalPosition::new(0., 0.);
+                state.window_id = wayland::make_wid(parent_surface);
+                state.previous_scale = 1.;
+
+                (TouchPhase::Started, pan_delta, 0., 0.)
+            },
+            Event::Update { time: _, dx, dy, scale, rotation } => {
+                let pan_delta = PhysicalPosition::new(dx as f32, dy as f32);
```
kchibisov, line 536: It's in `surface` coordinates, but `PhysicalPosition` is buffer, you should convert from logical to physical with the function like in `wl_pointer` events above.

src/platform_impl/linux/wayland/seat/pointer/mod.rs
```diff
@@ -467,9 +469,107 @@ impl Dispatch<WpCursorShapeManagerV1, GlobalData, WinitState> for SeatState {
...
+                let scale_delta = scale - state.previous_scale;
+                state.previous_scale = scale;
+                (TouchPhase::Moved, pan_delta, scale_delta, -rotation as f32)
+            },
+            Event::End { time: _, serial: _, cancelled } => {
+                let pan_delta = PhysicalPosition::new(0., 0.);
+                state.previous_scale = 1.;
+                (
+                    if cancelled == 0 { TouchPhase::Ended } else { TouchPhase::Cancelled },
```
kchibisov, line 545: Define it on a separate line, so it'll end up more compact.

src/platform_impl/linux/wayland/seat/pointer/mod.rs
```diff
@@ -467,9 +469,107 @@ impl Dispatch<WpCursorShapeManagerV1, GlobalData, WinitState> for SeatState {
...
+                    0.,
+                    0.,
+                )
+            },
+            _ => unreachable!("Unknown event {event:?}"),
+        };
+        state.events_sink.push_window_event(
+            WindowEvent::PanGesture { device_id, delta: pan_delta, phase },
+            state.window_id,
+        );
```
kchibisov, line 556: Isn't `Pan` basically a swipe?

@madsmtm do you know about it? Also, it's strange that we don't have amount of fingers in `Pan` gesture if it's so.

src/platform_impl/linux/wayland/seat/pointer/mod.rs
```diff
@@ -467,9 +469,107 @@ impl Dispatch<WpCursorShapeManagerV1, GlobalData, WinitState> for SeatState {
...
+        state.events_sink.push_window_event(
+            WindowEvent::PinchGesture { device_id, delta: scale_delta, phase },
+            state.window_id,
+        );
+        state.events_sink.push_window_event(
+            WindowEvent::RotationGesture { device_id, delta: rotation_delta, phase },
+            state.window_id,
+        );
```
kchibisov, line 564: You should only send updates if there was a change compared to previous state, so if there was not pinch, but just rotation, only rotation should be send.

src/platform_impl/linux/wayland/state.rs
```diff
@@ -100,6 +100,15 @@ pub struct WinitState {
     /// Pointer constraints to handle pointer locking and confining.
     pub pointer_constraints: Option<Arc<PointerConstraintsState>>,
 
+    /// Pointer gestures to handle swipe, pinch and hold.
+    pub pointer_gestures: Option<PointerGesturesState>,
+
+    /// XXX: Is this really meant to stay here?
+    pub window_id: WindowId,
```
kchibisov, line 107: No, it should be like with keyboard focus, on user data.

src/platform_impl/linux/wayland/state.rs
```diff
@@ -100,6 +100,15 @@ pub struct WinitState {
...
+
+    /// XXX: Is this really meant to stay here?
+    pub previous_scale: f64,
```
kchibisov, line 110: No, should be on user data as well.

kchibisov (changes requested):
Besides that I have no idea what `pan` gesture is and that our docs are a bit confusing about it and when to use.

It seems like it's just a movement within the gesture disregarding amount of fingers, etc?

src/platform_impl/linux/wayland/seat/pointer/mod.rs
madsmtm: The terminology is probably different elsewhere, but I think a pan gesture ([docs](https://developer.apple.com/documentation/uikit/touches_presses_and_gestures/handling_uikit_gestures/handling_pan_gestures?language=objc)) is similar, but distinct from a swipe gesture ([docs](https://developer.apple.com/documentation/uikit/touches_presses_and_gestures/handling_uikit_gestures/handling_swipe_gestures?language=objc)); the pan is more general (e.g. moving a map around), a swipe is usually short and only in one direction (e.g. dismissing a pop-up).

It would make sense to have the number of fingers in the event.

kchibisov:
Done in https://github.com/rust-windowing/winit/pull/4338

PR #3658, opened 2024-04-27
Bump version on master

kchibisov:
This commit does not represent a release and only synchronizes CHANGELOG from the latest release.


madsmtm (approved):
Approved, though the git history seems weird to me, I'd assume that when doing a breaking release we would base the `0.30.x` branch off of `master`?

kchibisov:
>Approved, though the git history seems weird to me, I'd assume that when doing a breaking release we would base the 0.30.x branch off of master?

new 0.31.x branch will be created. Everything is according to our `CONTRIBUTING` guidelines.

PR #3660, opened 2024-04-27
Reexport older versions of raw-window-handle

linkmauve:
When the user decides to use an older version of raw-window-handle, through the `rwh_04` or `rwh_05` features, it makes sense to reexport the crate so they don’t have to depend on it manually and can instead use `winit::raw_window_handle`.

### DIFF
--- a/src/changelog/unreleased.md
+++ b/src/changelog/unreleased.md
@@ -40,6 +40,10 @@ changelog entry.
 
 ## Unreleased
 
+### Added
+
+- Reexport `raw-window-handle` versions 0.4 and 0.5 as `raw_window_handle_04` and `raw_window_handle_05`.
+
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


madsmtm (approved):
Yeah, makes sense to me (not sure about the changelog categorization, is this a fix or an addition?)

src/changelog/unreleased.md
```diff
@@ -39,3 +39,7 @@ The migration guide could reference other migration examples in the current
 changelog entry.
 
 ## Unreleased
+
+### Fixed
```
kchibisov, line 43:
```suggestion
### Added
```

kchibisov (changes requested):

kchibisov (approved):
Renamed to avoid breaking change.

Issue #3670, opened 2024-05-01
Reintroduce ability to query use of X11 or Wayland as backend on raw `EventLoop`


vorporeal:
### Description

Pre 0.30, `EventLoopWindowTargetExtX11` enabled users to invoke `is_x11` on an `EventLoop`, as an `EventLoop` could be coerced into an `EventLoopWindowTarget`.

With the 0.30 API changes that replaced `EventLoopWindowTarget` with `ActiveEventLoop`, there is no longer a way to perform this check without doing something indirect/awkward like doing a match on `event_loop.owned_display_handle().raw_display_handle()`.

This functionality was used in our application to determine whether to invoke `winit::platform::x11::register_xlib_error_hook` before starting the event loop.  While we could set up our error hook callback as part of the init event, we'd prefer to do it before starting the event loop for maximum "coverage".

### Relevant platforms

X11

kchibisov:
Yeah, probably was just forgotten. Using `raw-window-handle` and `matches!` should be fine as well.

Though, in the future such identity may go away anyway when the backends could be plugged/etc.

PR #3672, opened 2024-05-02
Support is_x11 and is_wayland on EventLoop

dhardy:

---

Winit v0.30 does not allow testing `is_wayland` on an `EventLoop` (but only on an `ActiveEventLoop`), though this appears to be possible. This is an issue for me.

The trait names `ActiveEventLoopExtWayland` and `ActiveEventLoopExtX11` are now inaccurate (may cause issues with future additional functionality?), but this is the most straightforward solution.

Fixes https://github.com/rust-windowing/winit/issues/3670

### DIFF
--- a/src/changelog/unreleased.md
+++ b/src/changelog/unreleased.md
@@ -12,7 +12,7 @@ on how to add them:
 ### Added
 
 - Add `Window::turbo()`, implemented on X11, Wayland, and Web.
-- Implement `ActiveEventLoopExtWayland` and `ActiveEventLoopExtX11` for `EventLoop`.
+- Add traits `EventLoopExtWayland` and `EventLoopExtX11`, providing methods `is_wayland` and `is_x11` on `EventLoop`.
 - On X11, add `Window::some_rare_api`.
 - On X11, add `Window::even_more_rare_api`.
 - On Wayland, add `Window::common_api`.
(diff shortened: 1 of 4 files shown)


kchibisov (changes requested):
`EventLoop` is not `Active` it should have its own extensions.

dhardy:
New traits then (a copy of these other than the name)?

kchibisov:
Yes. Which is unfortunate, but it will prevent errors if someone tries to add API for the `EventLoop` which requires `EventLoop` to be actually running.

notgull (approved):

src/changelog/unreleased.md
```diff
@@ -12,6 +12,7 @@ on how to add them:
 ### Added
 
 - Add `Window::turbo()`, implemented on X11, Wayland, and Web.
+- Implement `ActiveEventLoopExtWayland` and `ActiveEventLoopExtX11` for `EventLoop`.
```
kchibisov, line 15: This is not correct anymore. It should mention the new traits.

kchibisov (changes requested):

src/changelog/unreleased.md
dhardy: Updated

kchibisov (approved):

PR #3677, opened 2024-05-03
Reduce usage of direct `msg_send!`

madsmtm:
`msg_send!` is more error-prone than using the generated bindings. Most of the methods that we're using now were not previously available in `objc2`, but they are now, so let's use them!


src/platform_impl/ios/window.rs
```diff
@@ -505,12 +507,9 @@ impl Window {
...
 
-        let gl_or_metal_backed = unsafe {
-            let layer_class = WinitView::layerClass();
-            let is_metal = msg_send![layer_class, isSubclassOfClass: class!(CAMetalLayer)];
-            let is_gl = msg_send![layer_class, isSubclassOfClass: class!(CAEAGLLayer)];
-            is_metal || is_gl
-        };
+        let is_metal = view.isKindOfClass(class!(CAMetalLayer));
+        let is_gl = view.isKindOfClass(class!(CAEAGLLayer));
+        let gl_or_metal_backed = is_metal || is_gl;
```
kchibisov, line 512: You can write all of that on the same line.

kchibisov (approved):

Issue #3671, opened 2024-05-01
SSH X11 forwarding seemed to be unsupported


ztlwk:
### Description

Attempting to initialize event loop with ssh x11 forwarding would end up with `NotSupportedError`.

### OS and window mananger

OS: Debian 12
Window manager: kwin

### Winit version

0.30.0

notgull:
Are there any logs in the `tracing` output?

ztlwk:
No, only some stack trace:
```
$ RUST_BACKTRACE=1 DISPLAY=localhost:10 RUST_LOG=trace ./winit-app-example
[2024-05-02T04:02:03Z TRACE winit_app_example] Starting App...
thread 'main' panicked at src/main.rs:34:58:
called `Result::unwrap()` on an `Err` value: NotSupported(NotSupportedError)
stack backtrace:
   0: rust_begin_unwind
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/std/src/panicking.rs:647:5
   1: core::panicking::panic_fmt
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/core/src/panicking.rs:72:14
   2: core::result::unwrap_failed
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/core/src/result.rs:1649:5
   3: core::result::Result<T,E>::unwrap
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/core/src/result.rs:1073:23
   4: winit_app_example::main
             at /home/smtbook/winit-app-example/src/main.rs:34:19
   5: core::ops::function::FnOnce::call_once
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```
FYI, `glxgears` worked properly.

ztlwk:
No, only some stack trace:
```
$ RUST_BACKTRACE=1 DISPLAY=localhost:10 RUST_LOG=trace ./winit-app-example
[2024-05-02T04:02:03Z TRACE winit_app_example] Starting App...
thread 'main' panicked at src/main.rs:34:58:
called `Result::unwrap()` on an `Err` value: NotSupported(NotSupportedError)
stack backtrace:
   0: rust_begin_unwind
[...]

PR #3679, opened 2024-05-04
Add more specific error messages for X11 failed initialization

notgull:
This ensures that, when X11 initialization fails, we know exactly why.

cc #3671

Signed-off-by: John Nunley <dev@notgull.net>

### DIFF
--- a/src/changelog/unreleased.md
+++ b/src/changelog/unreleased.md
@@ -43,6 +43,8 @@ changelog entry.
 ### Added
 
 - Reexport `raw-window-handle` versions 0.4 and 0.5 as `raw_window_handle_04` and `raw_window_handle_05`.
+- On X11, a log message will be emitted to describe why the display failed to
+  initialize if it did.
 
 ### Fixed
 
--- a/src/changelog/unreleased.md
+++ b/src/changelog/unreleased.md
@@ -43,8 +43,8 @@ changelog entry.
 ### Added
 
 - Reexport `raw-window-handle` versions 0.4 and 0.5 as `raw_window_handle_04` and `raw_window_handle_05`.
-- On X11, a log message will be emitted to describe why the display failed to
-  initialize if it did.
+- On X11, the "NotSupportedError" type now has extra information if the display
+  failed to initialize. 
 
 ### Fixed
 
--- a/src/error.rs
+++ b/src/error.rs
@@ -17,7 +17,17 @@ pub enum ExternalError {
 /// The error type for when the requested operation is not supported by the backend.
 #[derive(Clone)]
 pub struct NotSupportedError {
-    _marker: (),
+    inner: NotSupportedRepr,
+}
+
+#[derive(Clone)]
+enum NotSupportedRepr {
+    /// Unclassified.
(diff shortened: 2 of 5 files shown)


src/changelog/unreleased.md
```diff
@@ -39,3 +39,8 @@ The migration guide could reference other migration examples in the current
 changelog entry.
 
 ## Unreleased
+
+### Added
+
+- On X11, a log message will be emitted to describe why the display failed to
+  initialize if it did.
```
kchibisov, line 46: This is not relevant for the users, so drop it.

src/platform_impl/linux/x11/xdisplay.rs
```diff
@@ -65,13 +65,28 @@ unsafe impl Sync for XConnection {}
...
+/// Annotate a result with a `tracing` message.
+trait ResultExt {
+    /// Attach a message to this result.
+    fn context(self, msg: &str) -> Self;
+}
+
+impl<T, E: core::fmt::Debug> ResultExt for Result<T, E> {
+    fn context(self, msg: &str) -> Self {
+        self.map_err(|err| {
+            tracing::error!("{}: {:?}", msg, err);
```
kchibisov, line 77: Could you probably do `use tracing::error!`?

kchibisov (changes requested):
In general, I think the better approach is to add proper annotation for the errors themselves.

Not sure how visible the tracing stuff would be when you actually _boil_ from winit and likely exit, etc.

notgull:
> In general, I think the better approach is to add proper annotation for the errors themselves.

Agreed, but this would require a breaking change.

kchibisov:
But X11 error is not a public API, so you'd just change the internal layout?

notgull:
Huh, I thought it was? Okay then, this can be fixed

kchibisov:
>Huh, I thought it was? Okay then, this can be fixed

Everything inside the `platform_impl` is private, only `platform` is public.

src/error.rs
```diff
@@ -17,7 +17,17 @@ pub enum ExternalError {
 /// The error type for when the requested operation is not supported by the backend.
 #[derive(Clone)]
 pub struct NotSupportedError {
-    _marker: (),
+    inner: NotSupportedRepr,
```
kchibisov, line 20: Shouldn't we do it uniformly with the rest of the errors? Since we generally point to `platform_impl` from here?

Though, in the current state of things it doesn't matter that much.

notgull: That would add a lot of additional code for little benefit

kchibisov:
I'm still not entirely sure what to do with this and I think it'll add more _boilerplate_ when splitting backends...

So, I'll keep it around for now because the change is rather small internally and provides no practical value, since all the info available in e.g. strace if you debug things like failed to load dynamic library.

kchibisov:
Since you've brought it in #3883, why not route everything into `OsError`, should be pretty much conservative and provide all the things needed?

You should also just forward the `EventLoopError` from the X11 backend like Wayland does.

PR #3682, opened 2024-05-05
Enable `rustc-check-cfg`

madsmtm:


build.rs
```diff
@@ -21,4 +21,22 @@ fn main() {
         wayland_platform: { all(feature = "wayland", free_unix, not(redox)) },
         orbital_platform: { redox },
     }
+
+    #[cfg(feature = "unstable-check-cfg")]
```
kchibisov, line 25: I wonder if you can just check for nightly?

daxpedda:
Its just nightly for now, but when Cargo v1.80 releases we would need a permanent solution anyway. Couldn't we just always emit those instead of gating it behind an unstable feature or does this break anything?

madsmtm:
> Its just nightly for now, but when Cargo v1.80 releases we would need a permanent solution anyway. Couldn't we just always emit those instead of gating it behind an unstable feature or does this break anything?

Hmm, it produces a warning when using v1.70:
> warning: cargo:rustc-check-cfg requires -Zcheck-cfg=output flag

I thought it would do more, but seems like the fix is actually just that simple.

daxpedda:
It actually found a bug, this is great!
https://github.com/rust-windowing/winit/actions/runs/8959437002/job/24604828704?pr=3682

daxpedda:
I see you filed a bug as well: https://github.com/katharostech/cfg_aliases/issues/7.

madsmtm:
> I see you filed a bug as well: [katharostech/cfg_aliases#7](https://github.com/katharostech/cfg_aliases/issues/7).

Heh, oops

kchibisov (approved):

PR #3683, opened 2024-05-05
event_loop: remove deprecated `run` APIs

kchibisov:
The APIs are not well suited for the `&dyn ApplicationHandler` model and
`Box<dyn EventLoop>` structure, thus remove them.

### DIFF
--- a/src/platform/web.rs
+++ b/src/platform/web.rs
@@ -55,8 +55,7 @@ use web_sys::HtmlCanvasElement;
 
 use crate::application::ApplicationHandler;
 use crate::cursor::CustomCursorSource;
-use crate::event::Event;
-use crate::event_loop::{self, ActiveEventLoop, EventLoop};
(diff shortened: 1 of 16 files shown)


madsmtm (approved):
👍. Will follow-up with a PR to change the macOS and iOS backends to not need the shim any more.

notgull (approved):
ACK

src/platform/web.rs
```diff
@@ -157,6 +156,8 @@ impl WindowAttributesExtWebSys for WindowAttributes {
 /// Additional methods on `EventLoop` that are specific to the web.
 pub trait EventLoopExtWebSys {
     /// A type provided by the user that can be passed through `Event::UserEvent`.
```
daxpedda, line 158:
```suggestion
    /// A type provided by the user that can be passed through [`Event::UserEvent`].
```

daxpedda (approved):

Issue #3668, opened 2024-05-01
winit 0.30 panics on exit on macOS


matanui159:
### Description

Using this very simple setup that just creates a single window and exits on close:
```rs
use winit::{application::ApplicationHandler, event::WindowEvent, event_loop::{EventLoop, ActiveEventLoop}, window::{Window, WindowId}};

struct App {
    window: Option<Window>,
}

impl App {
    fn new() -> Self {
        Self {
            window: None
        }
    }
}

impl ApplicationHandler for App {
    fn resumed(&mut self, event_loop: &ActiveEventLoop) {
        let window = event_loop.create_window(Default::default()).unwrap();
        self.window = Some(window);
    }

    fn window_event(
        &mut self,
        event_loop: &ActiveEventLoop,
        _: WindowId,
        event: WindowEvent,
    ) {
        match event {
            WindowEvent::CloseRequested => event_loop.exit(),
            _ => {},
        }
    }
}

fn main() {
    let event_loop = EventLoop::new().unwrap();
    let mut app = App::new();
    event_loop.run_app(&mut app).unwrap();
}
```

At exit it will panic with the following error:
```
thread 'main' panicked at [...]/winit-0.30.0/src/platform_impl/macos/app_delegate.rs:147:39:
a delegate was not configured on the application
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

While this isn't the biggest issue since it was already exiting, it prevents doing any checks/cleanup after the event loop has run. Using similar code for 0.29 doesn't cause any issues, and neither does using `run_app_on_demand`.

### macOS version

```shell
ProductName:            macOS
ProductVersion:         14.4.1
BuildVersion:           23E224
```

### Winit version

0.30.0

madsmtm:
Thanks for the reproducer. Enabling a full backtrace reveals the issue to happen when dropping the `Window`, inside `windowWillClose:`.

This will be fixed by https://github.com/rust-windowing/winit/pull/3684, in the meantime, you can prevent this issue by dropping the window yourself inside `exiting` or the `CloseRequested` event:
```rust
WindowEvent::CloseRequested => {
    let _ = self.window.take();
    event_loop.exit();
}
```

PR #3684, opened 2024-05-05
Keep a reference to ApplicationDelegate in NSWindowDelegate and NSView

madsmtm:
The delegate is only weakly referenced by NSApplication, so getting it from there may fail if the event loop has been dropped.

Fixes https://github.com/rust-windowing/winit/issues/3668.

### DIFF
--- a/src/changelog/unreleased.md
+++ b/src/changelog/unreleased.md
@@ -39,3 +39,7 @@ The migration guide could reference other migration examples in the current
 changelog entry.
 
 ## Unreleased
+
+### Fixed
+
+- On macOS, fix panic on exit when dropping windows outside the event loop.
--- a/src/platform_impl/macos/app_delegate.rs
+++ b/src/platform_impl/macos/app_delegate.rs
@@ -52,6 +52,8 @@ pub(super) struct State {
     wait_timeout: Cell<Option<Instant>>,
     pending_events: RefCell<VecDeque<QueuedEvent>>,
     pending_redraw: RefCell<Vec<WindowId>>,
+    // NOTE: This is strongly referenced by our `NSWindowDelegate` and our `NSView` subclass, and
+    // as such should be careful to not add fields that, in turn, strongly reference those.
 }
 
 declare_class!(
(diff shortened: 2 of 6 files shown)


src/changelog/unreleased.md
```diff
@@ -39,3 +39,6 @@ The migration guide could reference other migration examples in the current
 changelog entry.
 
 ## Unreleased
+
+### Fixed
+- On macOS, fix panic on exit when dropping windows outside the event loop.
```
kchibisov, line 44:
```suggestion
### Fixed

- On macOS, fix panic on exit when dropping windows outside the event loop.
```

src/platform_impl/macos/app_delegate.rs
```diff
@@ -52,6 +52,8 @@ pub(super) struct State {
     wait_timeout: Cell<Option<Instant>>,
     pending_events: RefCell<VecDeque<QueuedEvent>>,
     pending_redraw: RefCell<Vec<WindowId>>,
+    // Note: This is strongly referenced by our `NSWindowDelegate` and our `NSView` subclass, and
```
kchibisov, line 55:
```suggestion
    // NOTE: This is strongly referenced by our `NSWindowDelegate` and our `NSView` subclass, and
```

src/platform_impl/macos/window_delegate.rs
```diff
@@ -74,6 +74,9 @@ impl Default for PlatformSpecificWindowAttributes {
 
 #[derive(Debug)]
 pub(crate) struct State {
+    /// Strong reference to the global application state.
+    app_delegate: Id<ApplicationDelegate>,
```
kchibisov, line 78: Does it matter in which order it's dropped?

kchibisov (changes requested):

src/platform_impl/macos/window_delegate.rs
madsmtm: Nope

kchibisov (approved):

PR #3697, opened 2024-05-19
Replace pointer dereference with read_unaligned()

CinTheDev:

**Reason for Pull Request:** On Raspberry Pi, using the Rust crate eframe caused the program to crash on mouse movement. The Backtrace lead to this specific line of code, and the exact error was a "misaligned pointer dereference: address must be a multiple of 0x8 but is xxxx"

The edit has been tested with the Raspberry Pi, which works now.

Technical information:

- Raspberry Pi 3 B+
- Raspbian OS Bullseye 32-Bit
- Using the Waveshare 3.5 (A) display connected via GPIO
- Using <https://github.com/goodtft/LCD-show> as display drivers

### DIFF
--- a/src/platform_impl/linux/x11/event_processor.rs
+++ b/src/platform_impl/linux/x11/event_processor.rs
@@ -1424,7 +1424,7 @@ impl EventProcessor {
             if !xinput2::XIMaskIsSet(mask, i) {
                 continue;
             }
-            let x = unsafe { *value };
+            let x = unsafe { value.read_unaligned() };
 
             // We assume that every XInput2 device with analog axes is a pointing device emitting
             // relative coordinates.


notgull:
Please run the program with xtrace and report the result.

notgull (approved):
Verified that nothing else appears to be broken on Xfce

CinTheDev:
Here is the output of xtrace, I ran it once without the fix (where the program crashes). The second time with the fix (program does not crash; I closed it after a few seconds).

Note: Here I'm using Version 0.29 of winit instead of master, with the fix applied. The fix itself remains unchanged.

[xtrace-nofix.log](https://github.com/rust-windowing/winit/files/15387941/xtrace-nofix.log)
[xtrace-fix.log](https://github.com/rust-windowing/winit/files/15387942/xtrace-fix.log)

Issue #3703, opened 2024-05-22
doc: pre_present_notify feature is not documented in FEATURES.md


filnet:
So it is hard to know on which platforms it is supported.

kchibisov:
It should say so in the docs instead like all other APIs.

`FEATURES.md` should be removed anyway since it serves no real purpose anymore.

filnet:
There is documentation https://github.com/rust-windowing/winit/blob/fff6788c12826e15631f32c1176ff679a3b1983c/src/window.rs#L633 but it is not clear which platforms actually support it. Reads like all platforms do.

kchibisov:
it's only wayland, it should be documented that nothing other than wayland supports it.

filnet:
Note that when creating a PR you are asked if `FEATURES.md` was updated...

PR #3707, opened 2024-05-26
doc: make it explicit that `pre_present_notify` is supported on Wayland only

filnet:
fixes https://github.com/rust-windowing/winit/issues/3703


src/window.rs
```diff
@@ -632,7 +632,8 @@ impl Window {
     ///
     /// ## Platform-specific
     ///
-    /// **Wayland:** - schedules a frame callback to throttle [`WindowEvent::RedrawRequested`].
+    /// - **Android / iOS / X11 / Web / Windows:** Unsupported.
```
filnet, line 635: Should it say "Does nothing", "Not implemented" or "Unsupported" ?

madsmtm: I think "Does nothing"

kchibisov: `Unsupported` is fine, because we can add something in there later. 

you should also mention all the platforms we support.

filnet: > you should also mention all the platforms we support.

`/// - **Wayland:** Schedules a frame callback to throttle [`WindowEvent::RedrawRequested`].`

This is not enough ?

kchibisov: You miss `Orbital` and `macOS` in `Unsupported` list.

kchibisov (approved):

kchibisov:
Thanks

PR #3709, opened 2024-05-27
Implement ApplicationHandler for `&mut A` and `Box<A>`

madsmtm:
I'm not sure if there was a reason these implementations didn't exist before, but they're nice to have.

A future PR could change the signature of `run_app` from taking `&mut A` to just taking `A` (which, in turn, would allow us to make `Drop` work better).


src/application.rs
```diff
@@ -223,3 +223,60 @@ pub trait ApplicationHandler<T: 'static = ()> {
         let _ = event_loop;
     }
 }
+
+impl<A: ApplicationHandler<T>, T: 'static> ApplicationHandler<T> for &mut A {
```
notgull, line 227: Could we add a `?Sized` bound here to enable dynamic dispatch?

notgull (changes requested):
I think we can also implement for `Box<A>`.

madsmtm:
> I think we can also implement for `Box<A>`.

You're right, have done so now.

src/changelog/unreleased.md
```diff
@@ -44,6 +44,7 @@ changelog entry.
 ### Added
 
 - Reexport `raw-window-handle` versions 0.4 and 0.5 as `raw_window_handle_04` and `raw_window_handle_05`.
+- Implement `ApplicationHandler` for `&mut` references and boxes to something that implements `ApplicationHandler`.
```
notgull, line 47: Nit: boxes => heap allocations

notgull (approved):

Issue #3206, opened 2023-11-01
Emit `Suspended` on every platform before `LoopDestroyed` for surface-cleanup feature parity?


MarijnS95:
As we have done in #2331 by emitting `Resumed` on every platform somewhere after the loop starts (for platforms that don't have an explicit point/callback where this should happen) to consistently implement surface creation, we should likely do the same with `Suspended` on loop shutdown for platforms that don't have an explicit callback here. This will allow implementers to perform surface destruction in a single consistent place too.

I'm currently improving a random `winit` sample to utilize these events, and the implementation becomes clumsy when having surface destruction in both `LoopExiting` and `Suspended`. It might even conflict with `WindowEvent::Destroyed`.

On Android I'm not yet sure if triggering a loop exit would terminate the Android app in such a way that surface destruction (== `Suspended`) is raised prior to `LoopExiting` anyway.

---

Side-note: Android supports multiple windows, and once the backend crates handle that properly perhaps `Resumed` needs to be moved to a `WindowEvent`, and `Suspended` will become `Destroyed`?

notgull:
I'm fine with this

daxpedda:
This affects Web as well, I'm in favor of this!

> Side-note: Android supports multiple windows, and once the backend crates handle that properly perhaps `Resumed` needs to be moved to a `WindowEvent`, and `Suspended` will become `Destroyed`?

Can't talk about other backends, but at least on Web this is weird because a `Window` doesn't represent an actual window, but just a canvas (see #696, which might attempt to change this, though I'm not sure how this affects the logic of all this for other backends, e.g. Android). So a single `Window` can't be independently suspended.

MarijnS95:
On Android a `winit::Window` should - theoretically when we fix everything - represent an `Activity`, which really is analogous to a desktop window. These typically have only one drawable surface for native apps/games [^1].

When one goes out of view and is paused (because activities are mostly fullscreen) their backing surface could be `Suspended`, so that might not map to Web then. But we could still start with consistently emitting the event "somewhere" before `LoopExit`.

One thing I haven't really looked into, but which we'll need to is loop consistency. For example on Android, do we use the proper APIs to destroy the app and receive a `NativeWindowDestroyed` (resulting in `winit::Event::Suspended`) before exiting the loop, or should we emulate it if the window is still there?

[...]

PR #3710, opened 2024-05-27
Verify event order

madsmtm:
The order that application handler methods are called in relation to each other is somewhat badly documented, and difficult to know for sure if correct. To remedy this, I've implemented a check that's enabled with debug assertions, and which logs if events are emitted in an unexpected order.

I went with using debug assertions, since I don't know of a better way to ensure that this will be tested, but there's of course a cost here in that if we don't catch all the issues, then downstream users will see these errors.

TODO:
- [ ] Discuss whether the implemented check is the correct one.
- [ ] Write better error message, with an explanation of where a user should report if they see it.
- [ ] Check that this is the current behaviour on all platforms.
  - [x] macOS
  - [ ] iOS
  - [ ] Android
  - [ ] Windows
  - [ ] X11
  - [ ] Wayland
  - [ ] Web

Once done, also fixes https://github.com/rust-windowing/winit/issues/3206.


kchibisov (changes requested):
The general idea is good, though, we'd need something to run all of that on CI in a crossplatform way.

But as for `Suspended` event, I'm still not sure whether we should emit it like that. Probably it won't hurt, but we must ensure that it's semantically the same place on all the platforms, right now I don't think it's the case.

src/event_loop.rs
```diff
@@ -209,6 +209,110 @@ impl EventLoop<()> {
...
+        ) {
+            self.state.expect(State::Running);
+            self.inner.device_event(event_loop, device_id, event);
+        }
+
+        fn memory_warning(&mut self, event_loop: &ActiveEventLoop) {
+            // TODO: What states are allowed when receiving this?
+            self.inner.memory_warning(event_loop);
+        }
+    }
```
kchibisov, line 311: I think it's just a generic event which should be _moved_ somewhere else, like to a separate trait. The `ApplicationHandler` today should represent the lifecycle of the event loop in the future, and real events should be on some other traits.

Issue #3094, opened 2023-09-10
Add cursor locking support for X11 and Windows


Imberflur:
As noted in https://github.com/rust-windowing/winit/issues/1677 locking the cursor can be useful for cases like camera control. `winit` does not currently provide locking for Windows and X11 but the information in that issue indicates that it should be possible for Windows and is maybe(?) possible for X11.

GiantBlargg:
Hacking around it isn't too bad. This is what I've been using on a recent project of mine. Haven't tested windows yet but x11(XWayland) seems to work.
```rust
#[derive(Default)]
struct MouseGrabber {
	last_pos: PhysicalPosition<f64>,
	manual_lock: bool,
}

impl MouseGrabber {
	// Call this on every WindowEvent::CursorMoved to update the last known position
	fn cursor_moved(&mut self, window: &Window, pos: PhysicalPosition<f64>) {
		if self.manual_lock {
			window.set_cursor_position(self.last_pos).unwrap();
		} else {
			self.last_pos = pos;
		}
	}
	// Call this with true to lock and false to unlock
	fn grab(&mut self, window: &Window, grab: bool) {
		if grab {
			if window.set_cursor_grab(CursorGrabMode::Locked).is_err() {
				window.set_cursor_grab(CursorGrabMode::Confined).unwrap();
				self.manual_lock = true;
			}
		} else {
			self.manual_lock = false;
			window.set_cursor_grab(CursorGrabMode::None).unwrap();
		}
		window.set_cursor_visible(!grab);
	}
}
```
This puts the mouse back to its last unlocked position, but it could probably be made simpler if you just wanted to lock to the center of the window.

Imberflur:
@GiantBlargg thanks for sharing a workaround!

kchibisov:
You can also submit this upstream,  just saying. It's not like it'll be any different....

ten3roberts:
Is there any reason or hurdles to implementing cursor locking in winit itself, rather than using this workaround?

dbartussek:
I'd also like to know if there is a reason why this is not implemented on Windows?

It seems very easy to implement by just taking the current cursor position and setting the cursor clipping window to just this position: https://github.com/dbartussek/winit/commit/dbd7cbdfc09252a7bfb1fda95e6607b2040c1ba7
Is there some hidden pitfall here I'm not seeing?

kchibisov:
Windows part done.

gavrix32:
[...]

PR #3711, opened 2024-05-27
Implemented CursorGrabMode::Locked for Windows

dbartussek:
This change fixes the Windows half of issue https://github.com/rust-windowing/winit/issues/3094

    I'm not sure where (if at all) to put this in the changelog. Should I add it as a "Changed" or "Added" entry?

### DIFF
--- a/src/platform_impl/windows/window.rs
+++ b/src/platform_impl/windows/window.rs
@@ -428,14 +428,6 @@ impl Window {
 
     #[inline]
     pub fn set_cursor_grab(&self, mode: CursorGrabMode) -> Result<(), ExternalError> {
-        let confine = match mode {
-            CursorGrabMode::None => false,
-            CursorGrabMode::Confined => true,
-            CursorGrabMode::Locked => {
-                return Err(ExternalError::NotSupported(NotSupportedError::new()))
-            },
-        };
-
         let window = self.window;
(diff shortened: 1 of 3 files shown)


src/platform_impl/windows/window.rs
```diff
@@ -446,7 +438,22 @@ impl Window {
...
+                        CursorGrabMode::Confined => {
+                            f.set(CursorFlags::GRABBED, true);
+                            f.set(CursorFlags::GRABBED_LOCKED, false);
+                        },
+                        CursorGrabMode::Locked => {
+                            f.set(CursorFlags::GRABBED, true);
+                            f.set(CursorFlags::GRABBED_LOCKED, true);
+                        },
+                    };
+                })
```
filnet, line 456: i suggest something along these lines:
```
let (grabbed, locked) = match mode {
	CursorGrabMode::None => (false, false),
	CursorGrabMode::Confined => (true, false),
	CursorGrabMode::Locked => (true, true),
};
f.set(CursorFlags::GRABBED, grabbed);
f.set(CursorFlags::GRABBED_LOCKED, locked);
```

oworope:
Thanks! I hope this gets merged as soon as possible 👏

madsmtm:
Replaced by https://github.com/rust-windowing/winit/pull/4177

PR #3713, opened 2024-05-28
Add Event::OpenFiles

jim-ec:
Emit an event when the application is opened by clicking on files in Finder or by dropping files on the Dock icon. This is macOS specific, as other OSs usually communicate that information through `std::env::args()`.

I am not sure with the current implementation, i.e. maybe the `OpenFiles` event should be defined in a macOS specific event enum type?

This basically implements #1751.

To test the event, a application bundle is required:

```
Winit.App
- Contents
  - Info.plist
  - MacOS
    - <executable>
```

The `Info.plist` should look like this:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
    <key>CFBundleExecutable</key>
    <string>
        <!-- Name of executable -->
    </string>
</dict>
</plist>
```

Also, you need to implement `fn ApplicationHandler::open_files(&mut self, files: Vec<PathBuf>)`.

- [ ] Updated [feature matrix](https://github.com/rust-windowing/winit

### DIFF
--- a/src/application.rs
+++ b/src/application.rs
@@ -1,5 +1,7 @@
 //! End user application handling.
 
+use std::path::PathBuf;
+
 use crate::event::{DeviceEvent, DeviceId, StartCause, WindowEvent};
 use crate::event_loop::ActiveEventLoop;
 use crate::window::WindowId;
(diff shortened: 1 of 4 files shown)


kchibisov (changes requested):
It should be on its own trait as it was requested multiple times. The `Event` will be gone soon, so no need to add anything into it.

madsmtm:
Thanks for the PR, though I'm going to resolve this in a different way, see https://github.com/rust-windowing/winit/issues/1751#issuecomment-2186560639.

PR #3716, opened 2024-06-01
v0.30.1

notgull:
### Added

- Reexport `raw-window-handle` versions 0.4 and 0.5 as `raw_window_handle_04` and `raw_window_handle_05`.
- Implement `ApplicationHandler` for `&mut` references and heap allocations to something that implements `ApplicationHandler`.

### Fixed

- On macOS, fix window dragging glitches when dragging across a monitor boundary with different scale factor.
- On macOS, fix the range in `Ime::Preedit`.


src/changelog/unreleased.md
```diff
@@ -12,6 +12,7 @@ on how to add them:
 ### Added
 
 - Add `Window::turbo()`, implemented on X11, Wayland, and Web.
+- Add traits `EventLoopExtWayland` and `EventLoopExtX11`, providing methods `is_wayland` and `is_x11` on `EventLoop`.
```
madsmtm, line 15: I don't think this goes here, that's the example documentation

madsmtm (approved):
A bit of a large diff (which is totally my own fault), but looks fine, thanks for doing this!

src/changelog/unreleased.md
notgull: Good catch, fixed!

kchibisov:
Want to figure https://github.com/rust-windowing/winit/pull/3722 out and then go back to this one.

notgull:
Looks like #3722 is blocked by debate. As I plan to do a weekly release cadence we can just pick it up in the next release.

kchibisov:
Cherry picked recent macOS queue fixes.

kchibisov (approved):

PR #3720, opened 2024-06-06
macOS/iOS: Various refactorings in application state

madsmtm:
I'm preparing to get rid of our application delegate in favour of registering notification observers, to do so I'm renaming `app_delegate.rs` to `app_state.rs`, and moving the functionality out of the Objective-C method into a normal method.

Additionally, `AppState` previously implemented `Default`, but really, this was a hack done because someone (probably myself) was too lazy to write out the full initialization in `AppDelegate::new`.

This is (deliberately) a non-functional change.

### DIFF
--- a/src/platform_impl/ios/app_state.rs
+++ b/src/platform_impl/ios/app_state.rs
@@ -20,7 +20,7 @@ use objc2_foundation::{
     CGRect, CGSize, MainThreadMarker, NSInteger, NSObjectProtocol, NSOperatingSystemVersion,
     NSProcessInfo,
 };
-use objc2_ui_kit::{UICoordinateSpace, UIView};
+use objc2_ui_kit::{UIApplication, UICoordinateSpace, UIView, UIWindow};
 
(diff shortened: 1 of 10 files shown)


src/platform_impl/ios/app_state.rs
```diff
@@ -662,6 +662,28 @@ fn handle_user_events(mtm: MainThreadMarker) {
     }
 }
 
+pub(crate) fn send_occluded_event_for_all_windows(application: &UIApplication, occluded: bool) {
+    let mtm = MainThreadMarker::from(application);
+
+    let mut events = Vec::new();
+    #[allow(deprecated)]
+    for window in application.windows().iter() {
+        if window.is_kind_of::<WinitUIWindow>() {
```
kchibisov, line 671: move this to `filter`

madsmtm: I'd rather not at this time, it's a safety condition for the `unsafe` just below, and moving it to `filter` makes it seem like they're less connected.

PR #3722, opened 2024-06-06
macOS: fix opacity handling

kchibisov:
Not using `NSColor::clearColor()` results in Quartz thinking that the window is not transparent at all, which results in artifacts.

However, not setting the `windowBackgroundColor` in `Window::set_transparent` results in border not properly rendered.

Fixes: 94664ff6876cd (Don't set the background color)

--

This is kind of a revert, given that https://github.com/alacritty/alacritty/pull/7965 not really doing anything and the window is not really tranparent as it should (it kind of is, but the value is different to what it should be and Quartz glitches behind the window with damage tracking).

Though, Quartz seems to generally glitch with e.g. shadows if you reload opacity with these changes as well. Maybe we should do something completely different, but I only care to fix the regression.

### DIFF
--- a/src/platform_impl/macos/window_delegate.rs
+++ b/src/platform_impl/macos/window_delegate.rs
@@ -10,7 +10,7 @@ use objc2::runtime::{AnyObject, ProtocolObject};
 use objc2::{declare_class, msg_send_id, mutability, sel, ClassType, DeclaredClass};
 use objc2_app_kit::{
     NSAppKitVersionNumber, NSAppKitVersionNumber10_12, NSAppearance, NSApplication,
-    NSApplicationPresentationOptions, NSBackingStoreType, NSDraggingDestination,
+    NSApplicationPresentationOptions, NSBackingStoreType, NSColor, NSDraggingDestination,
     NSFilenamesPboardType, NSPasteboard, NSRequestUserAttentionType, NSScreen, NSView,
     NSWindowButton, NSWindowDelegate, NSWindowFullScreenButton, NSWindowLevel,
     NSWindowOcclusionState, NSWindowOrderingMode, NSWindowSharingType, NSWindowStyleMask,
@@ -613,6 +613,8 @@ fn new_window(
 
         if attrs.transparent {
             window.setOpaque(false);
+            // See `set_transparent` for details on why we do this.
+            window.setBackgroundColor(unsafe { Some(&NSColor::clearColor()) });
         }
 
         // register for drag and drop operations.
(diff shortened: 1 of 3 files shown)


madsmtm (changes requested):
I'm not against this per-say, but I think we should experiment and finish this in Alacritty _first_, and only once we know what the desired behaviour is, then we can consider doing the change in Winit.

kchibisov:
I mean, that breaks not only alacritty, but basically everything that wants to put something transparent on screen, like glutin example, etc, etc.

kchibisov:
I'd also note, that it won't align with how **all** platforms work in winit, so not setting proper transparency sounds like we're not trying to do reliable cross platform behavior.

src/platform_impl/macos/window_delegate.rs
```diff
@@ -821,7 +824,15 @@ impl WindowDelegate {
...
 
     pub fn set_transparent(&self, transparent: bool) {
-        self.window().setOpaque(!transparent)
+        self.window().setOpaque(!transparent);
+
+        let color = if transparent {
+            unsafe { NSColor::clearColor() }
+        } else {
+            unsafe { NSColor::windowBackgroundColor() }
+        };
```
madsmtm, line 833: Better document why `setOpaque` is not enough.

```suggestion
        // AppKit draws the window with a background color by default, which is usually really
        // nice, but gets in the way when we want to allow the contents of the window to be
        // transparent, as in that case, the transparent contents will just be drawn on top of
        // the background color. As such, to allow the window to be transparent, we must also set
        // the background color to one with an empty alpha channel.
        let color = if transparent {
            // Let's use `clearColor`, the alpha channel is 0.0 on that.
            unsafe { NSColor::clearColor() }
        } else {
            // Otherwise, reset to the default `windowBackgroundColor`.
            unsafe { NSColor::windowBackgroundColor() }
        };
```

src/platform_impl/macos/window_delegate.rs
```diff
@@ -821,7 +824,15 @@ impl WindowDelegate {
...
+
+        self.window().setBackgroundColor(Some(color).as_deref());
```
madsmtm, line 835: Nit: Slightly clearer.

```suggestion
        self.window().setBackgroundColor(Some(&color));
```

madsmtm (approved):
Sorry it took me a while, I went and tested this with the following rendering methods:
- Raw `CALayer` (softbuffer with a fix to make alpha channels work there).
- OpenGL (`glutin`).
- Metal (`metal-rs` with `layer.set_opaque(false)`).
- `CGContext` (`cacao`).
- `NSView`'s positioned with different background colors (`cacao`).

And indeed for all of them, the correct behaviour is to set the background color to `clearColor`, `setOpaque` didn't seem to even do anything (!).

Also noting for the future: `-[NSWindow setAlphaValue:]` is similar to what we're doing here, but distinct in that it sets the opacity of the entire window, including decorations / title bar, and this cannot be "overridden" by rendering something with full opacity.

Approved after comments are resolved.

kchibisov:
>And indeed for all of them, the correct behaviour is to set the background color to clearColor, setOpaque didn't seem to even do anything (!).

`setOpaque` is a _hint_ for compositor and nothing more, IIRC. Which means that handling it reduces the load on the compositor, but that's about it. It doesn't actually check whether something is opaque/not.

>Also noting for the future: -[NSWindow setAlphaValue:] is similar to what we're doing here, but distinct in that it sets the opacity of the entire window, including decorations / title bar, and this cannot be "overridden" by rendering something with full opacity.

This is not desired in most cases, because it makes the text transparent as well. So should just leave it.

src/platform_impl/macos/window_delegate.rs
```diff
@@ -821,7 +823,20 @@ impl WindowDelegate {
     }
 
     pub fn set_transparent(&self, transparent: bool) {
-        self.window().setOpaque(!transparent)
+        self.window().setOpaque(!transparent);
```
madsmtm, line 826: > `setOpaque` is a hint for compositor and nothing more, IIRC. Which means that handling it reduces the load on the compositor, but that's about it. It doesn't actually check whether something is opaque/not.

Maybe document that here?

PR #3728, opened 2024-06-11
Web: queue `EventLoopProxy::send_event()` to microtask when triggered from outside the event loop

### LINKED ISSUE #3715: `EventLoopProxy::send_event` executes user events immediately in microtasks
### Description

My understanding is that [`EventLoopProxy::send_event`](https://docs.rs/winit/latest/winit/event_loop/struct.EventLoopProxy.html#method.send_event) is intended for dispatching events to be run during the next run of the event loop. However, when called during a microtask on web (e.g. via `wasm_bindgen_futures::spawn_local`), the event handler is called immediately.  This can cause borrow issues with app-level `RefCell`s.  My sense is that this isn't the intended behavior -- as far as I can tell, this never happens when using `send_event` in an async context on Linux, and was the cause of some surprising crashes that took a while to track down.

I have a simple repro for this issue here: https://github.com/ddfisher/winit-example/tree/main

In short, if we have [this setup](https://github.com/ddfisher/winit-example/blob/main/app/src/main.rs#L31-L61):
```rust
    fn window_event(&mut self, event_loop: &ActiveEventLoop, _id: WindowId, event: WindowEvent) {
        match event {
            [...]
            WindowEvent::MouseInput { state, .. } => {
                if state.is_pressed() {
                    let proxy_copy = self.proxy.clone();
                    wasm_bindgen_futures::spawn_local(async move {
                        console::log!("Sending event.");
                        proxy_copy
                            .send_event(CustomEvent::CustomEvent)
                            .expect("Send event failed");
                        console::log!("Returned from event send.");
                    })
                }
            }
        }
    }

    fn user_event(&mut self, _event_loop: &ActiveEventLoop, event: CustomEvent) {
        match event {
            CustomEvent::CustomEvent => {
                console::log!("Event handled.");
            }
        }
    }
```
... and click on the canvas, we see:
```
Sending event.
Event handled.
Returned from event send.
```

I think this may be related to this code, which makes it look like this behavior is purposeful:
https://github.com/rust-windowing/winit/blob/2d1382f7d669d068535f22e10b953048d6bd596f/src/platform_impl/web/event_loop/runner.rs#L472-L493

Is this the intended behavior?

### Tested browsers

Firefox, Chrome

### Tested devices

M3 Macbook Pro

### Winit version

0.30

daxpedda:
This also checks if its coming from a thread or not, in which case it doesn't queue it.

I also noticed some legacy code that deferred to `ControlFlow::Poll` when receiving events instead of waking up the event loop immediately, which I removed, so now events should come in faster regardless of `PollStrategy`.

Fixes #3715.

### DIFF
--- a/src/changelog/unreleased.md
+++ b/src/changelog/unreleased.md
@@ -40,6 +40,11 @@ changelog entry.
 
 ## Unreleased
 
+### Fixed
+
+- On Web, fix `EventLoopProxy::send_event()` triggering event loop immediately
+  when not called from inside the event loop. Now queues a microtask instead.
+
 ### Removed
 
 - Remove `EventLoop::run`.
--- a/src/changelog/unreleased.md
+++ b/src/changelog/unreleased.md
@@ -40,6 +40,11 @@ changelog entry.
 
 ## Unreleased
 
+### Changed
+
+- On Web, let events wake up event loop immediately when using
+  `ControlFlow::Poll`.
+
 ### Fixed
 
 - On Web, fix `EventLoopProxy::send_event()` triggering event loop immediately
--- a/src/platform_impl/web/async/waker.rs
+++ b/src/platform_impl/web/async/waker.rs
@@ -1,4 +1,5 @@
 use std::future;
+use std::num::NonZeroUsize;
 use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use std::sync::Arc;
 use std::task::Poll;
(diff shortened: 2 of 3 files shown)


madsmtm (approved):
Don't understand what's happening here, just gonna leave an approving review so that you can merge it since it only touches the web backend.

A few questions though: Why are we not using `wasm_bindgen_futures` internally ourselves? And why does `web_sys` not expose `queueMicrotask`?

daxpedda:
> A few questions though: Why are we not using `wasm_bindgen_futures` internally ourselves?

We are in a few places.
While it is true that `wasm_bindgen_futures::spawn_local()` actually uses `queueMicrotask()` internally, it does also do way more then just that, which we don't need in this case.

> And why does `web_sys` not expose `queueMicrotask`?

Just missing, I will make a PR.

madsmtm:
> Just missing, I will make a PR.

Don't sweat it, we do this sort of thing all the time in Winit, just curious.

daxpedda:
> > Just missing, I will make a PR.
> 
> Don't sweat it, we do this sort of thing all the time in Winit, just curious.

Done in https://github.com/rustwasm/wasm-bindgen/pull/3981.

PR #3732, opened 2024-06-15
v0.30.2

notgull:
### Changed

- On Web, let events wake up event loop immediately when using
  `ControlFlow::Poll`.

### Fixed

- On Web, fix `EventLoopProxy::send_event()` triggering event loop immediately
  when not called from inside the event loop. Now queues a microtask instead.
- On Web, stop overwriting default cursor with `CursorIcon::Default`.
- On Web, prevent crash when using `InnerSizeWriter::request_inner_size()`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -142,6 +142,7 @@ features = [
     "NSApplication",
     "NSBitmapImageRep",
     "NSButton",
+    "NSColor",
     "NSControl",
     "NSCursor",
     "NSDragging",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "winit"
-version = "0.30.1"
+version = "0.30.2"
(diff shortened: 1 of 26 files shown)


notgull:
See #3731 for the failing clippy lints

daxpedda (changes requested):
The https://github.com/rust-windowing/winit/pull/3728/commits/858b84e0ce1bc0675847376e4e7910ec214a4646 commit has to be excluded, as this is is technically a breaking change.

I can do it as well tonight.

Thanks for doing this @notgull!

kchibisov:
@daxpedda check that I've reverted the web correctly, given that there's no dedicated commit for the said _breaking_ change.

kchibisov (approved):

daxpedda (approved):
Thanks @kchibisov, Web looks good!

PR #3736, opened 2024-06-17
macOS: Set the theme on the `NSWindow`, instead of application-wide

madsmtm:
This new implementation uses:
- The NSAppearanceCustomization protocol for retrieving the appearance of the window, instead of using the application-wide `-[NSApplication effectiveAppearance]`.
- Key-Value observing for observing the `effectiveAppearance` to compute the `ThemeChanged` event, instead of using the undocumented `AppleInterfaceThemeChangedNotification` notification.

This also fixes `WindowBuilder::with_theme` not having any effect (fixes https://github.com/alacritty/alacritty/issues/8046), and the conversion between `Theme` and `NSAppearance` is made a bit more robust.

### DIFF
--- a/examples/window.rs
+++ b/examples/window.rs
@@ -212,6 +212,12 @@ impl Application {
             Action::PrintHelp => self.print_help(),
             #[cfg(macos_platform)]
             Action::CycleOptionAsAlt => window.cycle_option_as_alt(),
+            Action::SetTheme(theme) => {
+                window.window.set_theme(theme);
+                // Get the resulting current theme to draw with
+                let actual_theme = theme.or_else(|| window.window.theme()).unwrap_or(Theme::Dark);
+                window.set_draw_theme(actual_theme);
+            },
             #[cfg(macos_platform)]
             Action::CreateNewTab => {
                 let tab_id = window.window.tabbing_identifier();
@@ -334,7 +340,7 @@ impl ApplicationHandler<UserEvent> for Application {
             },
             WindowEvent::ThemeChanged(theme) => {
                 info!("Theme changed to {theme:?}");
-                window.set_theme(theme);
+                window.set_draw_theme(theme);
             },
             WindowEvent::RedrawRequested => {
                 if let Err(err) = window.draw() {
@@ -733,8 +739,8 @@ impl WindowState {
         self.window.request_redraw();
     }
 
-    /// Change the theme.
-    fn set_theme(&mut self, theme: Theme) {
+    /// Change the theme that things are drawn in.
+    fn set_draw_theme(&mut self, theme: Theme) {
         self.theme = theme;
         self.window.request_redraw();
     }
@@ -884,6 +890,7 @@ enum Action {
     ShowWindowMenu,
     #[cfg(macos_platform)]
     CycleOptionAsAlt,
+    SetTheme(Option<Theme>),
     #[cfg(macos_platform)]
     CreateNewTab,
     RequestResize,
@@ -915,6 +922,9 @@ impl Action {
             Action::ShowWindowMenu => "Show window menu",
             #[cfg(macos_platform)]
             Action::CycleOptionAsAlt => "Cycle option as alt mode",
+            Action::SetTheme(None) => "Change to the system theme",
+            Action::SetTheme(Some(Theme::Light)) => "Change to a light theme",
+            Action::SetTheme(Some(Theme::Dark)) => "Change to a dark theme",
             #[cfg(macos_platform)]
             Action::CreateNewTab => "Create new tab",
             Action::RequestResize => "Request a resize",
@@ -1059,6 +1069,10 @@ const KEY_BINDINGS: &[Binding<&'static str>] = &[
         Action::AnimationCustomCursor,
     ),
     Binding::new("Z", ModifiersState::CONTROL, Action::ToggleCursorVisibility),
+    // K.
+    Binding::new("K", ModifiersState::empty(), Action::SetTheme(None)),
+    Binding::new("K", ModifiersState::SUPER, Action::SetTheme(Some(Theme::Light))),
+    Binding::new("K", ModifiersState::CONTROL, Action::SetTheme(Some(Theme::Dark))),
     #[cfg(macos_platform)]
     Binding::new("T", ModifiersState::SUPER, Action::CreateNewTab),
     #[cfg(macos_platform)]
--- a/src/changelog/unreleased.md
+++ b/src/changelog/unreleased.md
@@ -45,6 +45,7 @@ changelog entry.
 - On Web, let events wake up event loop immediately when using
   `ControlFlow::Poll`.
 - Bump MSRV from `1.70` to `1.73`.
+- On macOS, set the window theme on the `NSWindow` instead of application-wide.
 
 ### Removed
 
@@ -55,3 +56,4 @@ changelog entry.
 ### Fixed
 
 - On X11, build on arm platforms.
+- On macOS, fixed `WindowBuilder::with_theme` not having any effect on the window.
--- a/src/event.rs
+++ b/src/event.rs
@@ -389,6 +389,8 @@ pub enum WindowEvent {
     /// Applications might wish to react to this to change the theme of the content of the window
     /// when the system changes the window theme.
     ///
+    /// This only reports a change if the window theme was not overridden by [`Window::set_theme`].
+    ///
     /// ## Platform-specific
     ///
     /// - **iOS / Android / X11 / Wayland / Orbital:** Unsupported.
--- a/src/platform_impl/macos/window_delegate.rs
+++ b/src/platform_impl/macos/window_delegate.rs
@@ -1,6 +1,8 @@
 #![allow(clippy::unnecessary_cast)]
 use std::cell::{Cell, RefCell};
 use std::collections::VecDeque;
+use std::ffi::c_void;
+use std::ptr;
 use std::sync::{Arc, Mutex};
 
 use core_graphics::display::{CGDisplay, CGPoint};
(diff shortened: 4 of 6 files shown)


src/event.rs
```diff
@@ -389,8 +389,11 @@ pub enum WindowEvent {
     /// Applications might wish to react to this to change the theme of the content of the window
     /// when the system changes the window theme.
     ///
+    /// This only reports a change if the window theme was not overridden by the application.
+    ///
     /// ## Platform-specific
     ///
+    /// - **Web:** Currently reports a change whether the window theme was overridden or not.
```
madsmtm, line 396: I am unsure if this is the desired behaviour here?

I think the Windows implementation checks the value set by the user, and doesn't emit an event in that case, and it seems like the Web backend doesn't do such a check? @daxpedda?

daxpedda: > I think the Windows implementation checks the value set by the user, and doesn't emit an event in that case, and it seems like the Web backend doesn't do such a check?

Web doesn't have a way to set the theme, ergo `Window::set_theme()` and `WindowAttributes::with_theme()` are no-op on Web. So I think your proposed changes are fine except the platform-specific addition to Web, as there is no way to overwrite the theme on Web.

> In reality, the `ThemeChanged` event is really an application-wide thing, even on Windows (the [`WM_SETTINGCHANGE`](https://learn.microsoft.com/en-us/windows/win32/winmsg/wm-settingchange) event is documented to be system-wide), so perhaps the correct solution would be to move it to `ApplicationHandler`?

This sounds good to me.

examples/window.rs
```diff
@@ -1059,6 +1069,10 @@ const KEY_BINDINGS: &[Binding<&'static str>] = &[
         Action::AnimationCustomCursor,
     ),
     Binding::new("Z", ModifiersState::CONTROL, Action::ToggleCursorVisibility),
+    // T
+    Binding::new("T", ModifiersState::empty(), Action::SetTheme(None)),
+    Binding::new("T", ModifiersState::SHIFT, Action::SetTheme(Some(Theme::Light))),
+    Binding::new("T", ModifiersState::CONTROL, Action::SetTheme(Some(Theme::Dark))),
```
kchibisov, line 1075: `T` is a bit unfortunate, because it's used for macOS tabs, and just `Shift+T` is also unfortunate, because it'll block regular input of `T`.

src/platform_impl/macos/window_delegate.rs
```diff
@@ -1787,34 +1823,43 @@ impl WindowExtMacOS for WindowDelegate {
...
-            NSString::from_str("NSAppearanceNameDarkAqua"),
-        ]))
-        .unwrap();
-    match &*name.to_string() {
-        "NSAppearanceNameDarkAqua" => Theme::Dark,
-        _ => Theme::Light,
+fn dark_appearance_name() -> Retained<NSString> {
+    // Don't use the static `NSAppearanceNameDarkAqua` to allow linking on macOS < 10.14
+    NSString::from_str("NSAppearanceNameDarkAqua")
+}
```
kchibisov, line 1829: What's the reason for a separate function?

src/platform_impl/macos/window_delegate.rs
```diff
@@ -1787,34 +1823,43 @@ impl WindowExtMacOS for WindowDelegate {
...
-                Theme::Dark => NSString::from_str("NSAppearanceNameDarkAqua"),
-                Theme::Light => NSString::from_str("NSAppearanceNameAqua"),
-            };
-            NSAppearance::appearanceNamed(&name).unwrap()
-        });
-        app.setAppearance(appearance.as_ref().map(|a| a.as_ref()));
+fn theme_to_appearance(theme: Option<Theme>) -> Option<Retained<NSAppearance>> {
+    if let Some(theme) = theme {
+        let appearance = match theme {
+            Theme::Light => unsafe { NSAppearance::appearanceNamed(NSAppearanceNameAqua) },
```
kchibisov, line 1852: Will it work on old macOS versions, because before it was using string here.

src/platform_impl/macos/window_delegate.rs
```diff
@@ -1787,34 +1823,43 @@ impl WindowExtMacOS for WindowDelegate {
...
-            let name = match t {
-                Theme::Dark => NSString::from_str("NSAppearanceNameDarkAqua"),
-                Theme::Light => NSString::from_str("NSAppearanceNameAqua"),
-            };
-            NSAppearance::appearanceNamed(&name).unwrap()
-        });
-        app.setAppearance(appearance.as_ref().map(|a| a.as_ref()));
+fn theme_to_appearance(theme: Option<Theme>) -> Option<Retained<NSAppearance>> {
+    if let Some(theme) = theme {
+        let appearance = match theme {
```
kchibisov, line 1851: You can use `?` operator to reduce nesting in this function.

src/platform_impl/macos/window_delegate.rs
```diff
@@ -419,32 +421,62 @@ declare_class!(
...
+        #[method(observeValueForKeyPath:ofObject:change:context:)]
+        fn observe_value(
+            &self,
+            key_path: Option<&NSString>,
+            _object: Option<&AnyObject>,
+            change: Option<&NSDictionary<NSKeyValueChangeKey, AnyObject>>,
+            _context: *mut c_void,
+        ) {
+            trace_scope!("observeValueForKeyPath:ofObject:change:context:");
+            if key_path == Some(ns_string!("effectiveAppearance")) {
```
kchibisov, line 435: Given that entire function is inside the `if` probably should just early return.

src/platform_impl/macos/window_delegate.rs
```diff
@@ -419,32 +421,62 @@ declare_class!(
...
+                let old = change.get(unsafe { NSKeyValueChangeOldKey }).expect("requested change dictionary did not contain `NSKeyValueChangeOldKey`");
+                let new = change.get(unsafe { NSKeyValueChangeNewKey }).expect("requested change dictionary did not contain `NSKeyValueChangeNewKey`");
+
+                // SAFETY: The value of `effectiveAppearance` is `NSAppearance`
+                let old: *const AnyObject = old;
+                let old: *const NSAppearance = old.cast();
+                let old: &NSAppearance = unsafe { &*old };
+                let new: *const AnyObject = new;
+                let new: *const NSAppearance = new.cast();
+                let new: &NSAppearance = unsafe { &*new };
```
kchibisov, line 446: I'm pretty sure these could be written on just a single line, so one for `old` and `new`

src/platform_impl/macos/window_delegate.rs
```diff
@@ -1787,34 +1823,43 @@ impl WindowExtMacOS for WindowDelegate {
...
-            NSString::from_str("NSAppearanceNameAqua"),
-            NSString::from_str("NSAppearanceNameDarkAqua"),
-        ]))
-        .unwrap();
-    match &*name.to_string() {
-        "NSAppearanceNameDarkAqua" => Theme::Dark,
-        _ => Theme::Light,
+fn dark_appearance_name() -> Retained<NSString> {
+    // Don't use the static `NSAppearanceNameDarkAqua` to allow linking on macOS < 10.14
+    NSString::from_str("NSAppearanceNameDarkAqua")
```
kchibisov, line 1828: You should likely use `ns_string!` macro here as well?

kchibisov (changes requested):

src/platform_impl/macos/window_delegate.rs
madsmtm: So that I didn't have to write the comment twice

src/platform_impl/macos/window_delegate.rs
madsmtm: Yeah, it's available since 10.9, see [the docs](https://developer.apple.com/documentation/appkit/nsappearancenameaqua).

src/platform_impl/macos/window_delegate.rs
madsmtm: At some point in the future we might get more `key_path`s that we're observing here, and then it makes sense to handle each in their own `if` - that's the whole reason why I'm checking the keypath, in reality we don't have to right now, since we're only observing one.

src/platform_impl/macos/window_delegate.rs
madsmtm: Yeah, but I like the explicitness of doing it this way, and it avoids using `as`.

Also, it's a mental push for me to get `downcast` to work in `objc2`

examples/window.rs
madsmtm: Do you have a better suggestion?

kchibisov: Any other letter which is not over used? The help is printed anyway, so which letter it is doesn't really matter.

src/changelog/unreleased.md
```diff
@@ -44,9 +44,14 @@ changelog entry.
 
 - On Web, let events wake up event loop immediately when using
   `ControlFlow::Poll`.
+- On macOS, set the window theme on the `NSWindow` instead of application-wide.
```
kchibisov, line 47: I'd consider that this is a _fix_, given that it was not behaving as on any other platform, though, it's a change in behavior which could be excluded from the point release... Not sure about it.

src/event.rs
```diff
@@ -389,6 +389,8 @@ pub enum WindowEvent {
     /// Applications might wish to react to this to change the theme of the content of the window
     /// when the system changes the window theme.
     ///
+    /// This only reports a change if the window theme was not overridden by the application.
+    ///
```
kchibisov, line 393: I think the previous sentence says the same thing. `system changes the window theme` is not a user changes the theme with `Window::set_theme`.

src/platform_impl/macos/window_delegate.rs
kchibisov: In the future, I'd prefer each if case like that to be in its own _function/method_, but for now, I'd use early return, because it's the best thing you can do _now_.

kchibisov (changes requested):

src/event.rs
madsmtm: I've changed this to reference `Window::set_theme` specifically instead

kchibisov (approved):

daxpedda:
See #3837.

PR #3739, opened 2024-06-18
Web: implement `WaitUntilStrategy`

### LINKED ISSUE #3738: Web: alternative strategy for `ControlFlow::WaitUntil`
Currently `ControlFlow::WaitUntil` uses [`Scheduler.postTask()`](https://developer.mozilla.org/en-US/docs/Web/API/Scheduler/postTask) and falls back to [`setTimeout()`](https://developer.mozilla.org/en-US/docs/Web/API/setTimeout).
Both strategies currently don't throttle on any browser as long as the tab is focused.
However when the tab is not focused it does throttle significantly (currently tested in Firefox and Chrome).

Currently the only known way to prevent throttling timers when the tab is not focused is to set up a timer in a worker, which doesn't throttle, and then wake up the main thread through that.
Obviously we shouldn't by default try and trick the browser and prevent it from doing its job, which in this case is throttling to preserve CPU and power consumption, but we can give users a way to opt-in to such methods.

The best I can come up with is to introduce another type analogous to [`PollStrategy`](https://docs.rs/winit/0.30.2/winit/platform/web/enum.PollStrategy.html), e.g. `WaitUntilStrategy`.
Implementation should be quite straightforward.

daxpedda:
This implements an alternative scheduling strategy for `ControlFlow::WaitUntil` on Web to prevent throttling when the tab is not focused by using a worker, which commonly browsers don't throttle regardless of tab focus.

I also added another step to the CI to add JS minification, let me know if you want me to remove it.

Fixes #3738.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -226,3 +226,19 @@ jobs:
         command: check
         log-level: error
         arguments: --all-features --target ${{ matrix.platform.target }}
+
+  swc:
+    name: Minimize JavaScript
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Install SWC
+        run: sudo npm i -g @swc/cli
+      - name: Run SWC
+        run: |
+          swc src/platform_impl/web/web_sys/worker.js -o src/platform_impl/web/web_sys/worker.min.js
+      - name: Check for diff
+        run: |
+          [[ -z $(git status -s) ]]
--- a/src/platform_impl/web/web_sys/worker.js
+++ b/src/platform_impl/web/web_sys/worker.js
@@ -1,0 +1,10 @@
+onmessage = event => {
+    const [port, timeout] = event.data
+    const f = () => port.postMessage(undefined)
+
+    if ('scheduler' in this) {
+        scheduler.postTask(f, { delay: timeout })
+    } else {
+        setTimeout(f, timeout)
+    }
+}
--- a/.gitignore
+++ b/.gitignore
@@ -3,8 +3,5 @@ target/
 rls/
 .vscode/
 *~
-*.wasm
-*.ts
-*.js
 #*#
 .DS_Store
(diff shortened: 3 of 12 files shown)


src/platform_impl/web/web_sys/worker.js
```diff
@@ -0,0 +1,10 @@
+onmessage = event => {
```
madsmtm, line 1: Hmm, I'm confused as to why this is in JavaScript? Is it not possible to write the worker in Rust? Or is it because it'd require a cumbersome `build.rs` step or similar to first compile the Rust code to WASM, and then embed that WASM into the final application?

(I'm totally fine with this being JS either way, mostly just curious).

.github/workflows/ci.yml
```diff
@@ -226,3 +226,19 @@ jobs:
         command: check
         log-level: error
         arguments: --all-features --target ${{ matrix.platform.target }}
+
+  swc:
+    name: SWC
```
madsmtm, line 231: Maybe (as an outsider, I had to look up what SWC is).
```suggestion
    name: Check JavaScript
```

madsmtm (approved):

src/platform_impl/web/web_sys/worker.js
daxpedda: Unfortunately its currently impossible to spawn a worker without at least some JS.
Even the current plan of the [Shared Everything Threads Proposal](https://github.com/WebAssembly/shared-everything-threads) requires a polyfill written in JS to make it work.

In any case, a lot of functionality currently requires JS anyway, its just that it is something `wasm-bindgen` usually takes care of. However spawning workers unfortunately isn't one of those things.

I was briefly considering waiting until I release my `web-thread` crate, but realized that I would require a custom JS solution for single-threaded Wasm anyway, so this works for both.

.github/workflows/ci.yml
daxpedda: Good point!

PR #3744, opened 2024-06-20
Winit version 0.30.3

kchibisov:
--

Need to fix build for arm.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -226,3 +226,19 @@ jobs:
         command: check
         log-level: error
         arguments: --all-features --target ${{ matrix.platform.target }}
+
+  swc:
+    name: Minimize JavaScript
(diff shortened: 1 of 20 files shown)


notgull (approved):

daxpedda (approved):

kchibisov:
@daxpedda please, next time you'll push like that, preserve the correct history, as in the commits you push should be in the same relevant order to the commits here. The actual release commit must be the last one.

In the end, when creating the release CHANGELOG entry, all the entries must be present in `unreleased.md` changelog (this is documented in CONTRIBUTING.md, since you apply patches first and then move entries). This is required to not overly modify the original commits, because if someone else will add one more commit, the rebase will be start requiring more and more manual work. Thanks!

daxpedda:
Apologies!

PR #3745, opened 2024-06-21
Bump version on master

kchibisov:
This commit does not represent a release and only synchronizes CHANGELOG from the latest release.

### DIFF
--- a/src/changelog/v0.30.md
+++ b/src/changelog/v0.30.md
@@ -1,3 +1,24 @@
+## 0.30.3
+
+### Added
+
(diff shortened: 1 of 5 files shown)


src/changelog/v0.30.md
```diff
@@ -1,3 +1,24 @@
+## 0.30.3
```
daxpedda, line 1: These entries should be removed from `unreleased.md`, right?

kchibisov: Yes, they should.

PR #3750, opened 2024-06-23
Use taiki-e/checkout-action

notgull:
taiki-e/checkout-action has a few advantages over actions/checkout,
such as:

- It is written in Bash rather than Node.js
- It does not have frequent breaking changes, reducing churn


madsmtm (approved):
Yeah, I can see the benefit, and we already depend on `taiki-e/install-action`, so this doesn't increase the attack vector / the number of people we depend on.

Issue #3735, opened 2024-06-16
winit/src/platform_impl/linux/x11/ime/context.rs


Y-substance:
### Description

[{
	"resource": "/root/.cargo/registry/src/mirrors.ustc.edu.cn-4affec411d11e50f/winit-0.30.2/src/platform_impl/linux/x11/ime/context.rs",
	"owner": "rustc",
	"code": {
		"value": "Click for full compiler diagnostic",
		"target": {
			"$mid": 1,
			"path": "/diagnostic message [0]",
			"scheme": "rust-analyzer-diagnostics-view",
			"query": "0",
			"fragment": "file:///root/.cargo/registry/src/mirrors.ustc.edu.cn-4affec411d11e50f/winit-0.30.2/src/platform_impl/linux/x11/ime/context.rs"
		}
	},
	"severity": 8,
	"message": "mismatched types\nexpected fn pointer `unsafe extern \"C\" fn(_, *mut u8, *mut u8)`   found fn pointer `unsafe extern \"C\" fn(_, *mut i8, *mut i8)`",
	"source": "rustc",
	"startLineNumber": 161,
	"startColumn": 13,
	"endLineNumber": 163,
	"endColumn": 14
}]

### OS and window mananger

Kali Linux

### Winit version

0.30.2

steev:
https://github.com/rust-windowing/winit/commit/078b46720bf8acb341bdec966296060395ec8e2b#diff-229cef84165211905c3f58d901f77e801dca2b9e8e20ce4a55a2ba75ba423909 is the commit change that breaks it

kchibisov:
I'm pretty sure it's a build failure? Or does it actually build? Which target is that? I'd assume aarch64?

steev:
It is a build failure, yes, it happens here on aarch64-unknown-linux-gnu; I haven't tested any other targets.

kchibisov:
it's just strange that the report looks like something from security tool, but in reality, it's just a regular build failure.

steev:
I thought the same thing :D  I checked because it says Kali Linux, and wanted to make sure it wasn't something we are doing.

For what it's worth, the error when compiling alacritty is:

```
error[E0308]: mismatched types
   --> /home/steev/.cargo/registry/src/index.crates.io-6f17d22bba15001f/winit-0.30.2/src/platform_impl/linux/x11/ime/context.rs:161:13
    |
[...]

PR #3751, opened 2024-06-23
Add atypical targets to CI

notgull:
This adds some Tier-2 and Tier-3 targets to CI, so we can detect
breakage in these platforms early.

cc #3735

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -202,6 +202,130 @@ jobs:
           ~/.cargo/git/db/
         key: cargo-${{ matrix.toolchain }}-${{ matrix.platform.name }}-${{ hashFiles('Cargo.lock') }}
 
+  tier2:
+    name: Check winit for Tier 2 Targets
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        toolchain: [stable, nightly, '1.73']
+
+    steps:
+    - uses: actions/checkout@v4
+
+    - uses: dtolnay/rust-toolchain@master
+      with:
+        toolchain: ${{ matrix.toolchain }}
+
+    - name: Use specific dependency versions for Rust 1.63 compatibility.
+      if: matrix.rust == '1.63'
+      run: |
+        cargo update --package=dashmap --precise 5.4.0
+        cargo update --package=regex --precise=1.9.0
+        cargo update --package=half --precise=2.2.1
+
+    - run: >
+        rustup target add
+        x86_64-unknown-linux-musl
+        x86_64-unknown-linux-gnux32
+        x86_64-linux-android
+        i686-linux-android
+        x86_64-unknown-freebsd
+        x86_64-unknown-netbsd
+        x86_64-unknown-illumos
+        i686-unknown-linux-musl
+        riscv64gc-unknown-linux-gnu
+        aarch64-unknown-linux-gnu
+        aarch64-unknown-linux-musl
+        powerpc64le-unknown-linux-gnu
+        armv5te-unknown-linux-gnueabi
+        s390x-unknown-linux-gnu
+        arm-linux-androideabi
+        sparc64-unknown-linux-gnu
+        sparcv9-sun-solaris
+
+    - run: cargo check --release -vv --all-targets --target=x86_64-unknown-linux-musl
+    - run: cargo check --release -vv --all-targets --target=x86_64-unknown-linux-musl --no-default-features --features x11
+    - run: cargo check --release -vv --all-targets --target=x86_64-unknown-linux-musl --no-default-features --features wayland
+    - run: cargo check --release -vv --all-targets --target=x86_64-unknown-linux-gnux32
+    - run: cargo check --release -vv --all-targets --target=x86_64-unknown-linux-gnux32 --no-default-features --features x11
+    - run: cargo check --release -vv --all-targets --target=x86_64-unknown-linux-gnux32 --no-default-features --features wayland
+    # TODO: Figure out best way to check Android targets.
+    # - run: cargo check --release -vv --all-targets --target=x86_64-linux-android --features android-native-activity
+    # - run: cargo check --release -vv --all-targets --target=x86_64-linux-android --features android-game-activity
+    # - run: cargo check --release -vv --all-targets --target=i686-linux-android --features android-native-activity
+    # - run: cargo check --release -vv --all-targets --target=i686-linux-android --features android-game-activity
+    - run: cargo check --release -vv --all-targets --target=x86_64-unknown-freebsd
+    - run: cargo check --release -vv --all-targets --target=x86_64-unknown-freebsd --no-default-features --features x11
+    - run: cargo check --release -vv --all-targets --target=x86_64-unknown-freebsd --no-default-features --features wayland
+    - run: cargo check --release -vv --all-targets --target=x86_64-unknown-netbsd
+    - run: cargo check --release -vv --all-targets --target=x86_64-unknown-netbsd --no-default-features --features x11
+    - run: cargo check --release -vv --all-targets --target=x86_64-unknown-netbsd --no-default-features --features wayland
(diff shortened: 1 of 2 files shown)


madsmtm (commented):
Hmm, not sure about this, our CI is already quite big, I'd rather cut down on it. Though I guess that doing it this way (only running `cargo check`, and doing it in a single "job") helps somewhat.

I'll note that Cargo supports compiling for multiple targets simultaneously (pass `--target` multiple times), might be worthwhile to check if that's faster (vs. the debugging experience if one target fails, of course).

notgull:
> Hmm, not sure about this, our CI is already quite big, I'd rather cut down on it. Though I guess that doing it this way (only running `cargo check`, and doing it in a single "job") helps somewhat.

The tradeoff is that otherwise we are completely unaware of such CI failures until they're reported to us. I figure that doing a `check` here is a good compromise between "no checks" and, say, using `qemu` to run a full test suite.

> I'll note that Cargo supports compiling for multiple targets simultaneously (pass `--target` multiple times), might be worthwhile to check if that's faster (vs. the debugging experience if one target fails, of course).

The issue is that we also want to test various feature combinations as well, but at least this will narrow it down to three calls.

kchibisov (changes requested):
I'd suggest to add only platforms which are only actually _different_, as in, you should cover possibilities and not all the cases.

For example, `c_char` is either `i8` or `u8`, so you need some target where it's `u8` (arm). You can also try to spot other differences, like where int/float, etc are different, but I don't really think you'll find anything other than `c_char` flip.

kchibisov:
As for musl, we need a musl based image, I guess.

notgull:
> I'd suggest to add only platforms which are only actually _different_, as in, you should cover possibilities and not all the cases.

The issue is that platforms may eventually become different. For instance FreeBSD regularly has breaking changes where fundamental types change. So in order to insulate us from unexpected breakage we need to cast a pretty wide net.

I'm glad to reduce the permutations here, but I think that the net this PR casts is just wide enough to capture all the permutations we care about (e.g. `x86_64` vs `aarch64` vs `s390x`, the various BSDs). So I would appreciate guidance on which cases we are fine with breaking for users.

kchibisov:
> The issue is that platforms may eventually become different

If the platform does so, then all the C stack has the issues as well. In general, I'd advice to have some sort of a lint that detects that the _matching type alias_ is used, and not just the _original type_. Because the issue you're trying to prevent was caused by the fact that `typedef` was not used, but rather the original type.

notgull:
> In general, I'd advice to have some sort of a lint that detects that the _matching type alias_ is used, and not just the _original type_.

Does this lint exist in the current version of Rust? I'm not sure this works, and the only real way to check this is to actually call `cargo check`.

kchibisov:
There's no way and no indications in clippy I've also opened an issue https://github.com/rust-lang/rust-clippy/issues/12988 to at least suggest correct types and not expanded.

kchibisov:
For now, I'd suggest to just test `arm` build, because it covers the `c_char` difference.

notgull:
> For now, I'd suggest to just test `arm` build, because it covers the `c_char` difference.

I've added the other builds for the following reasons:

- musl, because it's a popular libc target and allows us to support Alpine Linux.
- gnux32, which is used frequently in multiarch scenarios.
- FreeBSD/NetBSD/OpenBSD as they are popular unix flavors.
- riscv, as it is a less used architecture but still used frequently in some applications (i.e. riscv computers are gaining more popularity)
- s390x/sparc64 as a representative sample for big-endian systems.

kchibisov:
To make things work on alpine, you should target alpine in particular, since every musl setup has their own quirks. I know that winit works just fine on alpine though.

I'm not sure how often x32 is even used, I think only gentoo has it as an option and that's about it.

FreeBSD and other BSDs should be tested on CI, not their target, because it's not really representative IIRC.

riscv has nothing special when it comes to _testing_.

s390x is a mainframe IIRC and I'm not sure that winit should target those, since they likely don't have graphics in the common sense. Especially given that you exclude Solaris.

I'd just suggest to add `arm64` on CI and if you really want to have a separate CI (maybe powered by sr.ht) with FreeBSD, alpine. I can register the runner if you really want to.

.github/workflows/ci.yml
daxpedda, line 1: Its actually more performant to split these into multiple steps, or use a matrix in this case, I believe.

daxpedda (approved):
Personally I don't mind having the CI run a bit longer. I think there are still some low-hanging fruit to improve the total time, so hopefully I will get to that soon.

I think it would also be nice to document these targets as "T2" somewhere as well.

kchibisov:
The point is that they add nothing valuable other than build CI time, only arm64 is sufficient to test 99% of the edge cases, because of different char.

daxpedda:
> The point is that they add nothing valuable other than build CI time, only arm64 is sufficient to test 99% of the edge cases, because of different char.

I believe that this is adding "nothing valuable" has already been disputed by:
> > I'd suggest to add only platforms which are only actually _different_, as in, you should cover possibilities and not all the cases.
> 
> The issue is that platforms may eventually become different. For instance FreeBSD regularly has breaking changes where fundamental types change. So in order to insulate us from unexpected breakage we need to cast a pretty wide net.

While I agree that this is pretty messed up, if this is true, then checking these platforms does indeed provide some value, as long as we don't have a lint in place which would mitigate this issue otherwise (or some other way to prevent it).

I think the true discussion that should be happening here is to compare the added CI time vs the benefit. Though I agree that the benefits are quite low.

@notgull I think we can better discuss this if you would try to parallelize the checks as I noted in my review (or find out that I was wrong about that being better) and compare the runtime to our current CI and tell us how it went.

notgull:
> I think the true discussion that should be happening here is to compare the added CI time vs the benefit. Though I agree that the benefits are quite low.

I'd argue not being blindsided by build failures is a pretty big benefit.
 
> @notgull I think we can better discuss this if you would try to parallelize the checks as I noted in my review (or find out that I was wrong about that being better) and compare the runtime to our current CI and tell us how it went.

There isn't much of a point; checks are pretty fast and all of the jobs complete way before the mainline test jobs anyways.

.github/workflows/ci.yml
```diff
@@ -202,6 +202,124 @@ jobs:
...
+        aarch64-unknown-linux-musl
+        powerpc64le-unknown-linux-gnu
+        armv5te-unknown-linux-gnueabi
+        s390x-unknown-linux-gnu
+        arm-linux-androideabi
+        sparc64-unknown-linux-gnu
+        sparcv9-sun-solaris
+
+    - run: cargo check --release -vv --all-targets --target=x86_64-unknown-linux-musl
+    - run: cargo check --release -vv --all-targets --target=x86_64-unknown-linux-musl --no-default-features --features x11
```
daxpedda, line 249: - Why are we using `--release`?
- The output of `-vv` seems to make the outout quiet hard to parse to me.
- Why `--all-targets`? What exactly are we testing in this case?

daxpedda (changes requested):
> > I think the true discussion that should be happening here is to compare the added CI time vs the benefit. Though I agree that the benefits are quite low.
> 
> I'd argue not being blindsided by build failures is a pretty big benefit.

I believe that @kchibisov made a good point that the only thing we are checking are bunch of type permutations, which should be covered by much fewer targets. If you disagree or think we are missing something here, it would be great if you elaborate so we can agree on the value provided here.
> > @notgull I think we can better discuss this if you would try to parallelize the checks as I noted in my review (or find out that I was wrong about that being better) and compare the runtime to our current CI and tell us how it went.
> 
> There isn't much of a point; checks are pretty fast and all of the jobs complete way before the mainline test jobs anyways.

Looking at the last run the new checks not only lasted the longest, but also ran the last. So I'm a bit confused about your statement.

From the timings it seems to me that parallelizing this step should speed up the overall time significantly.

Considering that all the data is here, I would appreciate if we manage to use it to make the conversation more constructive.

notgull:
Needs rework

PR #3754, opened 2024-06-23
Remove internal `ActiveEventLoop::clear_exit`

madsmtm:
How to store and clear the internal state should be internal to the implementation on each backend.

I'm doing this because I want to refactor the macOS backend to not need to track this state manually in the future, which I so far _think_ should be possible: `NSApplication` itself already (roughly) tracks this in [`-[NSApplication isRunning]`](https://developer.apple.com/documentation/appkit/nsapplication/1428759-running?language=objc).

### DIFF
--- a/src/platform_impl/linux/mod.rs
+++ b/src/platform_impl/linux/mod.rs
@@ -904,7 +904,7 @@ impl ActiveEventLoop {
         x11_or_wayland!(match self; Self(evlp) => evlp.control_flow())
     }
 
-    pub(crate) fn clear_exit(&self) {
+    fn clear_exit(&self) {
         x11_or_wayland!(match self; Self(evlp) => evlp.clear_exit())
     }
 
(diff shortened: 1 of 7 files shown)


notgull (approved):
ACK for X11 and Windows

kchibisov (approved):

daxpedda (approved):

src/platform_impl/linux/mod.rs
```diff
@@ -904,7 +904,8 @@ impl ActiveEventLoop {
         x11_or_wayland!(match self; Self(evlp) => evlp.control_flow())
     }
 
-    pub(crate) fn clear_exit(&self) {
+    #[allow(dead_code)]
+    fn clear_exit(&self) {
         x11_or_wayland!(match self; Self(evlp) => evlp.clear_exit())
     }
```
kchibisov, line 910: If it's dead, then just remove it.

kchibisov (changes requested):

src/platform_impl/linux/mod.rs
madsmtm: It's not dead, it's called by `self.window_target.p.clear_exit();` on Wayland, similar to `set_exit_code` and `exit_code`.

kchibisov: I changed both platforms to do the same and removed `dead_code`.

kchibisov (approved):

PR #3756, opened 2024-06-24
Move iOS and macOS implementations into new `apple` module

madsmtm:
Move iOS and macOS implementations to a shared folder called `apple`, to allow us to reduce the code-duplication between these platforms in the future.

The folder structure is now:
- `src/platform_impl/apple/`
  - `appkit/`
  - `uikit/`
  - `example_shared_file.rs`
  - `mod.rs`

Part of https://github.com/rust-windowing/winit/issues/1029.

I considered calling the folder `darwin`, as that's the name of the underlying core OS that all of these platforms share, but the implementation here in Winit is really Apple-specific, i.e. even if Apple created a new OS different from Darwin, they would likely still port Foundation and UIKit to that, and the module name would still fit.


notgull (approved):
Nice! I hope this paves the way for further deduplication.

daxpedda (approved):

ActuallyHappening:
I'm a bit of a noob, is the `uikit` for iOS and `appkit` for macOS?

madsmtm:
Yup, but UIKit is also for tvOS, watchOS, visionOS and Mac Catalyst.

Issue #2674, opened 2023-02-05
Is there a way to access macos events or raw Object ?


ta3pks:
I was trying to implement openWith support for my app and on macos this is done using the events by overwriting the methods under NsApplication (thats my understanding of it) [here](https://developer.apple.com/documentation/appkit/nsapplicationdelegate)

Is there a way to tap into these events somehow ?

madsmtm:
There is work in progress on the generic case in https://github.com/rust-windowing/winit/issues/2120, but no clear solution has emerged yet.

madsmtm:
After https://github.com/rust-windowing/winit/pull/3758, listening to `NSApplicationDelegate` events will be possible by using `objc2::declare_class!`, Objective-C, Swift, or similar. [The WIP docs for `winit::platform::macos`](https://rust-windowing.github.io/winit/winit/platform/macos/index.html) should contain a rough example of how to do this once the PR is merged. Feel free to open an issue, probably preferably on [the `objc2` repo](https://github.com/madsmtm/objc2), if you're having trouble with doing so.

Issue #1281, opened 2019-11-20
Consider implementing Lifecycle on controller or view.


seivan:
[These](https://github.com/rust-windowing/winit/blob/master/src/platform_impl/ios/view.rs#L565-L585) could be made as notifications on either the view or the controller.

Which would make it useful in case you want to build  a Window from Rust on an existing `UIApplicationDelegate`

Necessary for https://github.com/rust-windowing/winit/issues/802

maxammann:
This will be a requirement for the maplibre-rs project. We do not want to render maps fullscreen but in separate views.

PR #3758, opened 2024-06-24
Allow the user to register the application delegate on macOS and iOS

madsmtm:
Register for event-loop lifecycle events using the `NSNotification` API instead of application delegates. This gives the user full control over the application delegate, which opens several doors for customization that were previously closed. 

In particular, this fixes https://github.com/rust-windowing/winit/issues/1751, closes https://github.com/rust-windowing/winit/pull/3713, fixes https://github.com/rust-windowing/winit/issues/2674, fixes https://github.com/rust-windowing/winit/pull/3499, fixes https://github.com/rust-windowing/winit/pull/3650 and fixes https://github.com/rust-windowing/winit/pull/2141, all by allowing the user to override `[NS|UI]ApplicationDelegate` themselves outside Winit.

Also fixes https://github.com/rust-windowing/winit/issues/1281 by listening on notifications instead of using the application delegate.

Resolves part of the macOS side of https://github.com/rust-windowing/winit/issues/2120.

- Tested on all platforms changed
  - [x] macOS
  - [x] iOS


daxpedda (approved):
I'm starting to understand the MacOS backend quite a bit now, the cleanup you have done so far is pretty amazing.

src/platform_impl/apple/appkit/app_state.rs
```diff
@@ -47,44 +46,21 @@ pub(super) struct AppState {
...
-
-        #[method(applicationWillTerminate:)]
-        fn app_will_terminate(&self, notification: &NSNotification) {
-            self.will_terminate(notification)
-        }
-    }
-);
+// TODO(madsmtm): Use `MainThreadBound` once that is possible in `static`s.
+thread_local! {
+    static GLOBAL: OnceCell<Rc<AppState>> = const { OnceCell::new() };
```
kchibisov, line 51: What's the reason to `Rc` thread local global? I'd assume the main reason is that the user creates it in `Rc` and shares the pointer with winit, thus it's correctly ref-counted.

And you're using global here just for simplicity?

kchibisov (approved):
I'm not that familiar with observers, but it looks ok-ish? I know that it works by appkit basically signalling you when something happens and you can register a lot of observers.

Also does it affect performance/input latency in any aspect or it's all the same because macOS is just giant message passing infra already and dispatching observer is not slower than dispatching regular methods?

madsmtm:
> Also does it affect performance/input latency in any aspect

I looked into it:

On AppKit, when calling `-[NSApplication setDelegate:]`, most of the delegate's `NSApplicationDelegate` methods (at least the ones we use) are internally registered as observer methods for the notifications we use here. So the performance impact of this PR is basically zero.

On UIKit, the delegate method is called directly, while the notification is emitted immediately afterwards. I'd suspect the performance impact of this PR to be very minimal, as UIKit still creates the `NSNotification` object even when there are no observers for it, but I haven't measured it.

In any case, you don't need to be concerned with buffering of notifications or other such shenanigans, the notifications are emitted at (basically) the same time as the methods are called.

src/platform_impl/apple/appkit/app_state.rs
madsmtm: > `Rc`

I haven't really decided how I want to pass the state around yet, so I just went with something that worked, and matched the semantics of what we were doing before (`Retained<ApplicationDelegate>` is basically the same as `Rc<AppState>`).

> thread local

That was to avoid a bit of `unsafe` until I've released a version of `objc2-foundation` with a `const` `MainThreadBound::new`. But I've worked around it with a helper for now instead.

> global

I _think_ that the only piece of state that has to _actually_ be global (in Winit) is the `ApplicationHandler`, since we have to access that from our overwritten `WinitApplication` when (hackily) dispatching device events. But if I find a different way to implement that, then I might be able to remove this final piece of global state in Winit's AppKit backend.

Korne127:
@kchibisov Just to make sure you didn't miss it

kchibisov:
@Korne127 miss what? I approved the PR long time ago...

Korne127:
Sorry if I'm wrong, to me it just looked like you were expected to give feedback to the new changes / address Mads' answer to your question.

kchibisov:
I've asked for feedback just to confirm my assumption, otherwise I'd _request changes_.

nicoburns:
This looks absolutely fantastic. This would solve a lot of the problems I've had with Winit on MacOS. Giving the application control of the AppDelegate is excellent on it's own. But I'm particular excited that this approach seems to open up the possibility of modular libraries that independently hook into apple system events!

madsmtm:
The holdup for this PR has not been Kirill at all, it's because I was on vacation ;)

Korne127:
Woohoo! Awesome :D
Thanks for all your work on this, and I hope you had a great vacation! :)

madsmtm:
I backported the iOS parts of this to v0.30.10, the macOS parts are more extensive, and will have to wait until v0.31.0.

CC @extrawurst, this should allow what we talked about on Discord with v0.30.10.

pomali:
> I backported the iOS parts of this to v0.30.10, the macOS parts are more extensive, and will have to wait until v0.31.0.

Is there any way I can help with macOS parts?

madsmtm:
Problem is that it depends on a bunch of other refactoring PRs I've done, and backporting those (correctly) is kind of a nightmare.

I know how useful this feature would be to downstream projects, but I can't really justify that time usage, I think my time's better spent getting v0.31 out the door.

Though I'd probably accept a PR that cherry-picked the required stuff against the `v0.30.x` branch, if you want to do it?

Issue #3714, opened 2024-05-28
iOS: Processing non `RedrawRequested` event after the main event loop: AboutToWait


marcpabst:
### Description

Stuck on winit `0.29.15` for now, while waiting for the new API to take form. Running the event loop on iOS causes lots of warnings ````WARNING winit.platform_impl.platform.app_state 2024-05-28 23:23:27,897 app_state.rs:617 processing non `RedrawRequested` event after the main event loop: AboutToWait````. I'm also seeing sporadic slow downs.

I'm running winit through `CPython`/`Pyo3`, butt looks like there is a similar problem on using bevy on iPhone: https://github.com/bevyengine/bevy/discussions/12934

### Device and iOS version

iOS 15.0.2 (19A404)
iPad Pro (11-inch) (2nd generation)

### Winit version

0.29.15

marcpabst:
Wuick follow-up (as I know this is hard to debug without code) - would it potentially help to just move to winit `0.30.*`? Have there been enough changes to event handling that this issue might just disappear on its own?

madsmtm:
> would it potentially help to just move to winit `0.30.*`

Nope, it's an old issue, the application state tracking is broken on newer iOS versions.

ActuallyHappening:
I am still getting this issue, with many noisy logs being printed every frame.
In the end I simply forked this from the point where the `crates.io` release was build, see here: https://github.com/ActuallyHappening/winit/tree/remove-noisy-logs

marcpabst:
Do you think this is actually impacting permance or is log spam?

madsmtm:
No idea, the backend is much older than my involvement in Winit, and haven't been updated for a long while. It's on my list to fix this, but said list is long!

ActuallyHappening:
I might be wrong, but I don't think that specific log event is still around in the `main` branch / `0.30` release

marcpabst:
Right, that inspires hope. I basically just noticed that while Metal is not the most reliably backend to begin with (in terms of dropped frames), I saw massive slow-downs on iPadOS that I wasn't seeing on MacOS.

madsmtm:
> I might be wrong, but I don't think that specific log event is still around in the `main` branch / `0.30` release

It is:
https://github.com/rust-windowing/winit/blob/82d9bbe5598c8605a752641a477261de01340aed/src/platform_impl/apple/uikit/app_state.rs#L549

ActuallyHappening:
[...]

PR #3762, opened 2024-06-26
add: Hacky `cfg` `winit_ignore_noisy_logs_unstable`

ActuallyHappening:

Helps address #3714.
Preferably the underlying issue could be fixed, but this is enough to satisfy my use case

### DIFF
--- a/src/platform_impl/ios/app_state.rs
+++ b/src/platform_impl/ios/app_state.rs
@@ -591,9 +591,15 @@ pub(crate) fn handle_nonuser_events<I: IntoIterator<Item = EventWrapper>>(
             match wrapper {
                 EventWrapper::StaticEvent(event) => {
                     if !processing_redraws && event.is_redraw() {
-                        tracing::info!("processing `RedrawRequested` during the main event loop");
+                        #[cfg(not(winit_ignore_noisy_logs_unstable))]
+                        tracing::info!(
+                            note = "See <https://github.com/rust-windowing/winit/issues/3714>",
+                            "processing `RedrawRequested` during the main event loop"
+                        );
                     } else if processing_redraws && !event.is_redraw() {
+                        #[cfg(not(winit_ignore_noisy_logs_unstable))]
                         tracing::warn!(
+                            note = "See <https://github.com/rust-windowing/winit/issues/3714>",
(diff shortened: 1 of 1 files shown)


daxpedda:
Considering we're using `tracing` now, Shouldn't this be filterable on the user level?

(can look into it when I get home)

ActuallyHappening:
You can do it yes, although it is not easily possible. You can't use an `EnvFilter` string like `info,[{message="very annoying log message here"}]=error` because the `tracing::warn!("very annoying log message here")` is matched by the `info` first.
If it is possible to filter away this on the user side, I would very much like to know! But I've read the `EnvFilter` docs (https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html) a few times and haven't noticed anything :(

daxpedda:
So it should be possible with using [Metadata filtering](https://docs.rs/tracing-subscriber/0.3.18/tracing_subscriber/layer/index.html#enabling-interest). There's a bunch of examples in the documentation using [`with_filter()`](https://docs.rs/tracing-subscriber/0.3.18/tracing_subscriber/layer/trait.Layer.html#method.with_filter).

You should also be able to simply [turn off all warnings in the desired module](https://docs.rs/tracing-subscriber/0.3.18/tracing_subscriber/filter/targets/struct.Targets.html#examples), which is much more blunt.

ActuallyHappening:
> So it should be possible with using [Metadata filtering](https://docs.rs/tracing-subscriber/0.3.18/tracing_subscriber/layer/index.html#enabling-interest). There's a bunch of examples in the documentation using [`with_filter()`](https://docs.rs/tracing-subscriber/0.3.18/tracing_subscriber/layer/trait.Layer.html#method.with_filter).
> 
> You should also be able to simply [turn off all warnings in the desired module](https://docs.rs/tracing-subscriber/0.3.18/tracing_subscriber/filter/targets/struct.Targets.html#examples), which is much more blunt.

Ahh thank you very much! This is very useful and should hopefully solve my problem. Using `bevy`'s [`LogPlugin`](https://docs.rs/bevy/latest/bevy/log/struct.LogPlugin.html#structfield.update_subscriber) should be able to do this.

Issue #3759, opened 2024-06-25
Allow handling of `iOS` Pen double taps / squeezes


ActuallyHappening:
### Description

I am trying to implement a 100% Rust note taking application for my iPad which can integrate with my Apple Pencil, which supports double tapping the tip of the pencil. This functionality is **critical** for apps to handle using Apple Pencil, and I would very much like this functionality to be added to `bevy` somehow since that is the library I am building on top of.
I believe adding this as a `platform_impl/iOS` platform feature would be sufficient, since I can `#[cfg(feature = "iOS")]` any code depending on it away anyway.

I have no experience with `objc2`, `icrate`, `UIKit` or even `winit` itself yet, and would definitely like some pointers (pun) to get me started.

### Relevant platforms

iOS

daxpedda:
Isn't this covered by #99?
Or why open a separate issue?

ActuallyHappening:
> Isn't this covered by #99? Or why open a separate issue?

Yes this is a subset, I've already read #99 but it looked far broader, with "eventually Apple Pencil" though I only really care about Apple Pencil for the moment. Support for other pens is a bonus, and would be nice, but Apple Pencil for my specific use case is the only hardware I have available at the moment.

I have an idea how to implement this, using a delegate (https://developer.apple.com/documentation/applepencil/handling-double-taps-from-apple-pencil) and somehow integrating with the `objc2` (https://docs.rs/objc2/0.5.2/objc2/macro.declare_class.html) using a protocol/trait implementation like:
```rust
	// Not sure if this is all we need to do
        unsafe impl UIPencilInteractionDelegate for WinitViewController {
    #[method(pencilInteraction:didReceiveTap:)]
    fn pencilInteraction_didReceiveTap(
        &self,
        interaction: &UIPencilInteraction,
        tap: &UIPencilInteractionTap,
    ) {
        tracing::error!(
            "Pencil Double Tap, interaction: {:?}, tap: {:?}",
            interaction,
            tap
        );
    }
```

How do I propagate this as an event? I'll work that out soon.
How can you patch this into bevy? I'll work that out soon.

I don't have a PR, but my GitHub fork has quite a few `testing-*` branches

daxpedda:
I'm going to close this in favor of #99 then.
We can re-open this if for some reason this feature is still missing after #99 is closed.

PR #3763, opened 2024-06-28
Winit apple pen

ActuallyHappening:
Implements #3759
Also see #99

I split the implementation and documentation into seperate commits, because I don't know if for the branch I am targeting the way I have updated the change-log is valid and expect to be corrected. (I've written the change under the heading `0.30.4` even though that version hasn't been released yet, since the `unreleased.md` was empty on the `v0.30.x` branch but populated on the `master` branch)

I have tested these changes on my iPad and they work exactly as expected. In an ideal world, I would like this to be released on `crates.io` in a way that `bevy` could integrate so I could use it in my `bevy` application ([tracking issue](https://github.com/bevyengine/bevy/issues/14060)), which is why I was working off the `v0.30.x` branch so I could seamlessly [patch](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-patch-section).

Also, the public API I have chosen seems reasonable to me but I am totally open to it being tweaked or re-worked. But I would ideally want this functionality exposed in some way without me having to maintain my own fork as `bevy` releases roll around. I have intentionally included [`#[non_exhaustive]`](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute) in places that I or other developers may add or improve features to. Notably, this PR only handles pen double tap events since that is all I can physically test with my current hardware, but newer Apple Pencil models support [squeeze actions](https://developer.apple.com/documentation/applepencil/handling-squeezes-from-apple-pencil#) that I know (theoretically) how to implement but haven't. The API intentionally allows for this addition later.
The public API I added is also intended to be extendable by other platform implementations, but again I can't test them physically so haven't included here. Each section has `## Platform Specific \n - iOS only` documentation for that reason.

This is my first contribution to `winit` (and my first major feature contribution in general using Open Source) so please tell me if I've done anything wrong! And thanks so much for this awesome crate, contributing to it has taught me a lot about [objc2](https://docs.rs/objc2/latest/objc2/macro.declare_class.html), UIKit and iOS development!


kchibisov:
You must always target the master. Though, you could still provide with a _backport_ patch once the master implementation is available.

kchibisov:
Oh, also, you can not target 0.30 branch, because you're adding breaking changes, thus they can not be a part of the patch release.

ActuallyHappening:
OK, so if I understand correctly this change since it adds new API surface and so can't are released as a patch version 0.30.4.
I'll have to make a new PR I presume adding this functionality to the `master` branch then, and I think I'll change the naming to `PenSpecialAction`, see #99

Issue #3695, opened 2024-05-11
Possible solution to keep raw window handle vaild to user


ztlwk:
### Description

Here's the idea:
* Require the user to implement a "bridge" type, which owns the corresponding surface, and gets dropped and recreated by winit when being suspended in Android.
* Make the "bridge" type include the window recreation logic, so winit can recreate the window and own it, which assured the users to have a "window" in every update.

daxpedda:
See https://github.com/rust-windowing/winit/issues/3699.
Cc @MarijnS95.

MarijnS95:
Seen this while reading #3699, and I'm more inclined to pursue that issue. For one, this issue doesn't describe a **problem statement**, so it's hard to understand what kind of issue or design problem the author intends this proposed design for.

Maybe this is somewhat loosely related to https://github.com/rust-windowing/winit/issues/2337?

daxpedda:
I think the goal by OP is that we somehow avoid having to re-create `RawWindowHandle`.
You indicated in the last meeting that this is possible, because that's how it was in the past, could you clarify the details here?

So my understanding is that `Window` survives just fine.
`RawWindowHandle` becomes invalid but not unsafe, e.g. rendering just goes into a black hole but doesn't segfault.

How much exactly of the graphics context is invalidated, how about GPU buffers?
Is it possible for Winit to drape over this fact and "patch" the `RawWindowHandle` so we can pretend everything's continuing to work? So just notifying the user that the app is now in the background and no rendering will be visible and after resumption everything continues to work just fine.

ztlwk:
Sorry for the vague description, I have edited it to make it more clear.

MarijnS95:
> So my understanding is that `Window` survives just fine.

On Android, if we consider `Window` to be analogous to an `Activity` (of which we can hackily only have _just one_ currently), yes. The `NativeWindow`/`Surface` in Android is "just" its temporary backing buffer (or well, buffer queue / producer end).

> `RawWindowHandle` becomes invalid but not unsafe, e.g. rendering just goes into a black hole but doesn't segfault.

https://github.com/rust-windowing/raw-window-handle/issues/111#issuecomment-1493408521

> How much exactly of the graphics context is invalidated, how about GPU buffers?

Just the `Surface` and `Swapchain` in Vulkan terms. In GL your context becomes "uncurrent" (because it can only be current when there's a render target, i.e. a `Surface`, as far as I understood) but the UBOs should continue to exist?

---

[...]

PR #3765, opened 2024-06-28
Implement `ApplicationHandler::create|destroy_surfaces()`

### LINKED ISSUE #3699: Cross-platform inconsistency in `ApplicationHandler::resumed()`
Currently Android requires the surface to be recreated in `resumed()`, but this isn't the case for Web and iOS (I think), where the surface does not need to be recreated on `resumed()`.

I'm unsure what the solution here might be, maybe it would be best to split this into two different methods, or maybe we can tell the user somehow that the `Window` is invalid now.

Related: https://github.com/rust-windowing/winit/issues/3695.

daxpedda:
`ApplicationHandler::resumed/suspended()`s meaning has become unclear after iOS and Web implemented them.
For iOS and Web they specifically mean suspending and resuming the application.
For Android it specifically meant that the render surface will become invalid when suspended and must be recreated when resumed.

This PR introduces `ApplicationHandler::create|destroy_surfaces()`, which take the place of `ApplicationHandler::resumed/suspended()`. iOS and Web are now the only ones implementing `ApplicationHandler::resumed/suspended()`.

Fixes #3699.


madsmtm (approved):
I feel like there's still a _lot_ to do here to make this correct, and a lot more to make it ergonomic, but splitting these events from each other is a great first step!

src/application.rs
```diff
@@ -18,69 +18,80 @@ pub trait ApplicationHandler {
...
+    ///
+    /// [suspended]: Self::destroy_surfaces
+    /// [`SurfaceView`]: https://developer.android.com/reference/android/view/SurfaceView
+    /// [Activity lifecycle]: https://developer.android.com/guide/components/activities/activity-lifecycle
+    /// [`VkSurfaceKHR`]: https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkSurfaceKHR.html
+    /// [`wgpu::Surface`]: https://docs.rs/wgpu/latest/wgpu/struct.Surface.html
+    ///
+    /// [`create_surfaces()`]: Self::create_surfaces
+    /// [`destroy_surfaces()`]: Self::destroy_surfaces
+    fn create_surfaces(&mut self, event_loop: &ActiveEventLoop);
```
kchibisov, line 94: I'm not sure I entirely like this method, but the alternative is to store windows in winit and indicate the creation with `WindowId`.

Though, unless the event loop itself could be less backend agnostic I don't think it'll happen.

kchibisov (changes requested):
I don't really like that it called `create_surfaces` and not `can_create_surfaces`, since the first name kind of forces you to create them.

destroy is fine, because it should _force_ you.

src/application.rs
daxpedda: This isn't about `Window` anymore, but only about `RawWindowHandle`.
The hope is that in the future we won't need this anymore if #3695 is resolved, but yeah, otherwise I can't come up with a better idea.

src/changelog/unreleased.md
```diff
@@ -52,6 +52,14 @@ changelog entry.
   to send specific data to be processed on the main thread.
 - Changed `EventLoopProxy::send_event` to `EventLoopProxy::wake_up`, it now
   only wakes up the loop.
+- `ApplicationHandler::create|destroy_surfaces()` was split off from
+  `ApplicationHandler::resumed/suspended()`.
+
+  `ApplicationHandler::create_surfaces()` should, for portability reasons to
+  Android, be the only place to create render surfaces.
```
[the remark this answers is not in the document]
daxpedda: What do you mean exactly?
Couple of lines below it the new behavior of `resumed/suspended()` is explained.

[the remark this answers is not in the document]
kchibisov: I'd probably write it more clear that the `resume` logic should be moved to the new function and that `resumed` serves now completely different purpose, because before we advertised it as a place to create a surfaces.

Alternative is to rename resumed to something else, so users don't have to read into the changelog a lot about why their app is not working anymore.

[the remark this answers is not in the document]
daxpedda: > I'd probably write it more clear that the `resume` logic should be moved to the new function and that `resumed` serves now completely different purpose, because before we advertised it as a place to create a surfaces.

That's not really correct though, the current `resume()` functionality isn't new, its just split off. So users can't just move their `resume()` logic to `can_create_surfaces()`, they have to actually just split it off, like it says in the changelog already.

I'm happy to review a suggestion on what you have in mind though.

> Alternative is to rename resumed to something else, so users don't have to read into the changelog a lot about why their app is not working anymore.

I don't think we have to worry about that because it will fail to build anyway unless `can_create_surfaces()` is implemented. Hopefully at that point users should be looking at the documentation to implement it.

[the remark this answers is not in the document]
kchibisov: hm, that's a good point, probably fine then.

kchibisov (approved):

examples/pump_events.rs
```diff
@@ -22,7 +22,7 @@ fn main() -> std::process::ExitCode {
     }
 
     impl ApplicationHandler for PumpDemo {
-        fn resumed(&mut self, event_loop: &ActiveEventLoop) {
+        fn can_create_surfaces(&mut self, event_loop: &ActiveEventLoop) {
             let window_attributes = Window::default_attributes().with_title("A fantastic window!");
             self.window = Some(event_loop.create_window(window_attributes).unwrap());
```
MarijnS95, line 27: Technically window creation is something that happens _before_ a surface can be created; this changed API just represents the async nature of it (and IIRC @kchibisov said it works like that on Wayland as well). Perhaps our examples shouldn't demonstrate creating window(s) here.

(Especially if we figure out how to make `winit::Window <=> android::Activity`, where there could be multiple `Activity`s contemporarily)

src/application.rs
```diff
@@ -18,69 +18,80 @@ pub trait ApplicationHandler {
...
+    /// been created. This is expected to closely correlate with the [`onResume`] lifecycle
+    /// event but there may technically be a discrepancy.
+    ///
+    /// [`onResume`]: https://developer.android.com/reference/android/app/Activity#onResume()
+    ///
+    /// Applications that need to run on Android must wait until they have been "resumed" before
+    /// they will be able to create a render surface (such as an `EGLSurface`, [`VkSurfaceKHR`]
+    /// or [`wgpu::Surface`]) which depend on having a [`SurfaceView`]. Applications must also
+    /// assume that if they are [suspended], then their render surfaces are invalid and should
+    /// be dropped.
```
MarijnS95, line 84: Now that the wording is fixed, I don't think this makes any sense: your render surface becomes invalid when Android says so; don't assume this when another event is received.

src/application.rs
```diff
@@ -208,25 +219,47 @@ pub trait ApplicationHandler {
...
     ///
-    /// Also see [`Resumed`] notes.
+    /// [`applicationWillResignActive`]: https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1622950-applicationwillresignactive
+    /// [iOS application lifecycle]: https://developer.apple.com/documentation/uikit/app_and_environment/managing_your_app_s_life_cycle
+    ///
+    /// ### Web
     ///
-    /// ## Android
+    /// On Web, the [`suspended()`] method is called in response to a [`pagehide`] event if the
+    /// page is being restored from the [`bfcache`] (back/forward cache) - an in-memory cache that
```
MarijnS95, line 238: I think you meant: `if the page is being moved/stored in(to) the bfcache`?

(EDIT: the original docs said: `the page is being put in the bfcache`)

src/application.rs
```diff
@@ -236,38 +269,24 @@ pub trait ApplicationHandler {
     /// destroyed, which indirectly invalidates any existing render surfaces that may have been
     /// created outside of Winit (such as an `EGLSurface`, [`VkSurfaceKHR`] or [`wgpu::Surface`]).
     ///
-    /// After being `Suspended` on Android applications must drop all render surfaces before
+    /// After being [suspended] on Android applications must drop all render surfaces before
```
MarijnS95, line 272: Same here, we probably shouldn't maintain the `suspended` wording for simplicity and to reduce complexity. Let the users' `Surface`, `Swapchain` etc lifetimes purely depend on the two `surface()` callbacks introduced here, and nothing else.

src/changelog/unreleased.md
```diff
@@ -57,6 +57,14 @@ changelog entry.
 - Changed `EventLoopProxy::send_event` to `EventLoopProxy::wake_up`, it now
   only wakes up the loop.
 - On Web, slightly improve accuracy of `DeviceEvent::MouseMotion`.
+- `ApplicationHandler::create|destroy_surfaces()` was split off from
```
MarijnS95, line 60: `can_create`

MarijnS95: Also, since this is plural, I think on Android we'd receive one callback per `Activity`, that its backing `Surface`/`NativeWindow` was created.

(And this may/will happen at different times depending on which of them is visible)

src/event.rs
```diff
@@ -85,6 +85,11 @@ pub(crate) enum Event {
     /// [`ApplicationHandler::suspended`]: crate::application::ApplicationHandler::suspended
     Suspended,
 
+    /// See [`ApplicationHandler::can_create_surfaces`] for details.
+    ///
+    /// [`ApplicationHandler::can_create_surfaces`]: crate::application::ApplicationHandler::can_create_surfaces
```
MarijnS95, line 90: `()` in the intradoc link? Seems we don't do that often enough :/

src/lib.rs
```diff
@@ -52,7 +52,7 @@
 //! }
 //!
 //! impl ApplicationHandler for App {
-//!     fn resumed(&mut self, event_loop: &ActiveEventLoop) {
+//!     fn can_create_surfaces(&mut self, event_loop: &ActiveEventLoop) {
 //!         self.window = Some(event_loop.create_window(Window::default_attributes()).unwrap());
```
MarijnS95, line 56: This also advocates creating a window _after its render surface became available_?

src/application.rs
```diff
@@ -18,69 +18,80 @@ pub trait ApplicationHandler {
...
+    /// [`pageshow`]: https://developer.mozilla.org/en-US/docs/Web/API/Window/pageshow_event
+    /// [`bfcache`]: https://web.dev/bfcache/
     ///
-    /// ## iOS
+    /// ### Others
     ///
-    /// On iOS, the `Resumed` event is emitted in response to an [`applicationDidBecomeActive`]
-    /// callback which means the application is "active" (according to the
-    /// [iOS application lifecycle]).
+    /// **Android / macOS / Orbital / Wayland / Windows / X11:** Unsupported.
```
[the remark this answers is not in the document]
daxpedda: I made one here https://github.com/rust-windowing/winit/issues/3779.

src/application.rs
```diff
@@ -18,69 +18,80 @@ pub trait ApplicationHandler {
...
+    ///
+    /// Applications should be able to gracefully handle back-to-back [`can_create_surfaces()`] and
+    /// [`destroy_surfaces()`] calls.
+    ///
+    /// ## Platform-specific
+    ///
+    /// ### Android
+    ///
+    /// On Android, the [`can_create_surfaces()`] method is called when a new [`SurfaceView`] has
+    /// been created. This is expected to closely correlate with the [`onResume`] lifecycle
```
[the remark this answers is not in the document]
daxpedda: I tried to change as little as possible, but generally speaking I would actually like to remove information like this entirely.

In the follow-up PR I will do that for Web as well, but I really think we should tell users what to expect and not explain how we implemented anything unless its really relevant.

examples/pump_events.rs
daxpedda: Good point, I will address this in the follow-up PR.

src/application.rs
```diff
@@ -18,69 +18,80 @@ pub trait ApplicationHandler {
...
+    /// event but there may technically be a discrepancy.
+    ///
+    /// [`onResume`]: https://developer.android.com/reference/android/app/Activity#onResume()
+    ///
+    /// Applications that need to run on Android must wait until they have been "resumed" before
```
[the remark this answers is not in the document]
daxpedda: I was actually referring to Android "resumed" and "suspended" in this section.
According to your earlier comment in https://github.com/rust-windowing/winit/pull/3765#discussion_r1661534343 this is still wrong though?

src/application.rs
daxpedda: I'm not following, what do you mean with "another event"?
The word "assume" should probably be removed here?

src/application.rs
daxpedda: Ah, thanks!

src/changelog/unreleased.md
daxpedda: I think we can revisit the whole plural thing when multiple activities become actually possible.

src/lib.rs
daxpedda: Yeah, same as https://github.com/rust-windowing/winit/pull/3765#discussion_r1661536569.

daxpedda (commented):
I apologize for not letting you review this before merging first!

Considering I was trying not to change anything about the documentation or the implementation outside of what we discussed, I assumed this should be pretty straightforward. Clearly I was wrong!

src/application.rs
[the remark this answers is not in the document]
MarijnS95: Yeah, please remove this. It's not really based on anything, clearly wrong, and irrelevant for the users' metal model nor expected programming interface to deal with this.

src/application.rs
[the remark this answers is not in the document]
MarijnS95: Yeah it sets the precedent that surface create/destroy is related to suspend/resume events (either Android or the old `winit` ones you removed here), which it is not.

src/application.rs
MarijnS95: It should just be removed altogether:

1. Android will raise an event when your surface is destroyed;
2. Hence when you destroy it in `suspended()` too, you now have to add complicated extra logic to make sure it's only destroyed in one of the two paths;
3. When an Android app is suspended in the [`onPause()`](https://developer.android.com/guide/components/activities/activity-lifecycle#onpause) sense (pending how #3779 pans out), the app _might still be visible_ and should have a render surface...

src/application.rs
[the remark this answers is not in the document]
MarijnS95: Will be fixed in #3786, marking as resolved.

src/application.rs
MarijnS95: Will be fixed in https://github.com/rust-windowing/winit/pull/3786, marking as resolved.

src/application.rs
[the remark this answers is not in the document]
MarijnS95: And now that I read this again: the `SurfaceView` isn't created, it already exists, but its `SurfaceHolder` creates a new `Surface`. Let's patch that too.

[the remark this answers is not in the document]
MarijnS95: I'm removing mentions of `SurfaceView` entirely. It's not at all relevant to the end user (only `NativeWindow` is, and optionally `SurfaceHolder` for how the callback is implemented on the Java end). Specifically, `NativeActivity` *replaces* the view root of the app by consuming the root `Surface` from the **`Window`** directly, which happens to be implemented using the same `SurfaceHolder` callbacks and subsumes the traditional view tree:

https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/core/java/android/app/NativeActivity.java;l=133;drc=5d123b67756dffcfdebdb936ab2de2b29c799321

(That is not to say, we can still create a hierarchy of sub-surfaces, parented to each other, which is available natively and also what a Java/Kotlin/XML view tree could do if multiple `SurfaceView` elements are nested)

examples/pump_events.rs
MarijnS95: @daxpedda did you ever make a followup for this?

src/application.rs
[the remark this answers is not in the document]
MarijnS95: Addressed in #3786.

src/application.rs
MarijnS95: Addressed in #3786 by replacing `suspended` with `onStart`/`onStop` and merely direct users to the `destroy_surfaces()` callback which is the only valid and consistent place.

src/application.rs
MarijnS95: Addressed in #3786.

examples/pump_events.rs
daxpedda: I don't believe so.

PR #3766, opened 2024-06-28
Web: Remove some unnecessary compatibility code

daxpedda:
Apparently `MouseEvent.movementX/Y` should not be used anymore.
Replacing it with `screenX/Y` simplifies the code, removes unnecessary conversion and increases accuracy.

Addresses https://github.com/rust-windowing/winit/issues/2875.

### DIFF
--- a/src/platform_impl/web/event_loop/runner.rs
+++ b/src/platform_impl/web/event_loop/runner.rs
@@ -234,7 +234,7 @@ impl Shared {
         ));
 
         let runner = self.clone();
-        let window = self.window().clone();
+        let mut delta = backend::event::MouseDelta::new();
         *self.0.on_mouse_move.borrow_mut() = Some(EventListenerHandle::new(
             self.window().clone(),
             "pointermove",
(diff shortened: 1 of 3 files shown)


src/platform_impl/web/event_loop/runner.rs
```diff
@@ -234,64 +234,74 @@ impl Shared {
...
-        let window = self.window().clone();
         *self.0.on_mouse_move.borrow_mut() = Some(EventListenerHandle::new(
             self.window().clone(),
             "pointermove",
-            Closure::new(move |event: PointerEvent| {
-                if !runner.device_events() {
-                    return;
-                }
+            Closure::new({
+                let mut delta = backend::event::MouseDelta::new();
```
madsmtm, line 241: Hmm, maybe you can keep this out where `let window` was before (around line 237), to reduce the diff noise?

src/platform_impl/web/event_loop/runner.rs
```diff
@@ -234,64 +234,74 @@ impl Shared {
...
 
-                    let y_motion = (delta.y != 0.0).then_some(Event::DeviceEvent {
-                        device_id,
-                        event: DeviceEvent::Motion { axis: 1, value: delta.y },
-                    });
+                            if delta.x == 0 && delta.y == 0 {
+                                return None.into_iter().chain(None).chain(None);
+                            }
 
-                    x_motion.into_iter().chain(y_motion).chain(iter::once(Event::DeviceEvent {
```
madsmtm, line 290: Why get rid of `iter::once`?

madsmtm (approved):
Was there a reason why we didn't use `screen[X|Y]` before?

src/platform_impl/web/event_loop/runner.rs
daxpedda: Good point, no reason to scope the variable like this anyway.

src/platform_impl/web/event_loop/runner.rs
daxpedda: Because the type has to match with [this line](https://github.com/daxpedda/winit/blob/bbf16bc6ee66320ab1353c2cf46b9a6e922f107b/src/platform_impl/web/event_loop/runner.rs#L280).
So I had to change it to `Option` to use `None` above so I can deliver an empty iterator.

daxpedda:
> Was there a reason why we didn't use `screen[X|Y]` before?

[`movementX/Y`](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/movementX) is what we actually want here, but apparently the spec says:
https://w3c.github.io/pointerlock/#dom-mouseevent-movementx
> `movementX` and `movementY` must be zero for all mouse events except `mousemove`.

We moved from mouse to pointer events (`mousemove` -> `pointermove`), so that doesn't really work out for us.
It worked until now because all browsers just don't follow the spec.

Issue #3406, opened 2024-01-17
iOS calling `Window::request_redraw()` inside `WindowEvent::RedrawRequested`


daxpedda:
Calling `Window::request_redraw()` inside `WindowEvent::RedrawRequested` currently panics on iOS.
We should not only not panic, but also correctly schedule the next `WindowEvent::RedrawRequested`.
This is related to #2412.

maxammann:
I think I'm hitting the same issue on iOS. The event loop is not progressing. When callling `request_redraw` on the `RedrawRequested`, no other events are progressed afterwards.

PWhiddy:
Any update on this? When I try to run an application on ios, I cannot seem to get request_redraw to trigger RedrawRequested repeatedly.

PWhiddy:
for anyone experiencing the same issue as me - ios not doing anything with request_redraw calls while handling a request redraw event, this was my solution: call request_redraw inside about_to_wait instead!
I only was able to learn about this because of https://github.com/bevyengine/bevy/pull/11245.
This would be super helpful to mention somewhere in the documentation!

kchibisov:
it's clearly a bug that must be fixed though, and IIRC it's kind of fixed now @madsmtm ?

evrimoztamur:
Experiencing the exact same issue, calling `request_redraw()` in any case besides `WindowEvent::RedrawRequested` seems to work.

`about_to_wait` does not seem to run at 60 FPS, on the other hand.

PWhiddy:
@evrimoztamur 
Someone was able to suggest a workaround here that worked great for me!

https://github.com/rust-windowing/winit/issues/4204

evrimoztamur:
Thanks! I'm using that right now, but I still wonder why there's platform-inconsistent behavior right now.

PWhiddy:
I can only assume this aspect of iOS may not have been anticipated in the original design. Hopefully eventually there is a unified API 🙏

PR #3767, opened 2024-06-28
Only run observers in the default run loop mode

madsmtm:
AppKit/UIKit has a concept of "run loops", which are effectively a queue of events that are repeatedly consumed from, and when empty, the consuming thread is `std::thread::park`ed until new events appear.

Apple has locked the exact details of these down quite tightly, especially on iOS, which is part of the reason why everything must be in a callback there instead of the original `pump_events` API, but effectively this is the basis of our event loop.

We are still given the option to listen to certain events, however! And this is how we implement the `new_events`/`about_to_wait` events, as well as the `proxy_wake_up` event.

One somewhat odd thing about run loops is that they support being run re-entrantly / in a nested fashion! In this way, the system takes events out of the queue that the nested caller requests, and keeps the rest of the events queued up until the nested caller returns and the outer caller can now process the queued events. Internally, it looks roughly like the following:

```rust
'outer loop {
    match run_loop.get_event_matching(_) {
        ClickOnCornerOfWindow => {
            'inner loop {
                match run_loop.get_event_matching(ReleaseClick | MouseMoved) {
                    ReleaseClick => break 'inner,
                    MouseMoved => window.resized(),
                    _ => unreachable!(),
                }
            }
        }
        ClickOnCloseButton => {
            break 'outer;
        }
        _ => {} // Other events
    }
}
```

Notably, both calls to the fictional "get_event_matching" can end up parking the thread when there is no more work to be done. Now, what this means for us is that we're given basically two options for how to emit `new_events`/`about_to_wait`/`proxy_wake_up`:

1. Emit it from inside each "get_event_matching" (`kCFRunLoopCommonModes`).
2. Emit it only from the outer event loop's "get_event_matching" (`kCFRunLoopDefaultMode`).

Nested run loops like these are used notably in the following instances:
A. When opening e.g. a file dialog with the `rfd` crate.
B. When clicking the corner of the window and dragging to live-resize it.

I am unsure what the desired behaviour is for `new_events`/`about_to_wait`/`proxy_wake_up`? In case A, the file dialog is spawned from inside the event loop itself, and as such we _cannot_ emit the event (since that would be re-entrant). But in case B, we might want to emit the `new_events`/`about_to_wait`/`proxy_wake_up`, even though that is not technically what is happening with our outer event loop.

I felt like the most consistent answer would be that we always use `kCFRunLoopDefaultMode` when originally writing the PR, though then again, maybe resizing should be considered an implementation detail here, and we should just paper over it (like we somewhat currently do)? CC @daxpedda @kchibisov, I'd like input on this! Would the user still expect `new_events`/`about_to_wait` to be emitted when live-resizing a window, even though the application is still somewhat "alive" at this stage?

(In any case, I will first have to [fix calling `request_redraw` inside of `RedrawRequested`](https://github.com/rust-windowing/winit/issues/3406), since the approach of calling it inside `about_to_wait` would break resizing with this PR in its current form).

TODO:
- [ ] Figure out what is actually the desired behaviour here.
- [ ] Test.
- [ ] Add entry to the changelog.
- [ ] Update internal documentation.
- [ ] Update public documentation (esp. also `pump_events`) with my learnings.
- [ ] Write a proper commit message.

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -12,4 +12,6 @@ disallowed-methods = [
     { path = "web_sys::Document::fullscreen_element", reason = "Doesn't account for compatibility with Safari" },
     { path = "objc2_app_kit::NSView::visibleRect", reason = "We expose a render target to the user, and visibility is not really relevant to that (and can break if you don't use the rectangle position as well). Use `frame` instead." },
     { path = "objc2_app_kit::NSWindow::setFrameTopLeftPoint", reason = "Not sufficient when working with Winit's coordinate system, use `flip_window_screen_coordinates` instead" },
+    { path = "core_foundation::runloop::kCFRunLoopCommonModes", reason = "Using the common modes includes the modal panel mode, which we usually want to avoid" },
+    { path = "objc2_foundation::NSRunLoopCommonModes", reason = "Using the common modes includes the modal panel mode, which we usually want to avoid" }
 ]
--- a/src/platform_impl/apple/appkit/app_state.rs
+++ b/src/platform_impl/apple/appkit/app_state.rs
@@ -323,8 +323,7 @@ impl ApplicationDelegate {
             .upgrade()
             .expect("The panic info must exist here. This failure indicates a developer error.");
 
-        // Return when in event handler due to https://github.com/rust-windowing/winit/issues/1779
-        if panic_info.is_panicking() || !self.ivars().event_handler.ready() || !self.is_running() {
+        if panic_info.is_panicking() || !self.is_running() {
             return;
         }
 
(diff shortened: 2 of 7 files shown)


mirsella:
I tested this on an iPhone 11 running iOS 26.6 with Winit `e0100de`, Bevy 0.19, and wgpu 29.

After five minutes in the background, the current default-mode observers produced a severe FPS drop during resume:

- `31 / 3 / 60 FPS`
- `31 / 4 / 59 FPS`

One run also had a 1.94-second gap between redraw callbacks.

Changing the three control-flow observers to `kCFRunLoopCommonModes` improved both runs:

- `33 / 19 / 59 FPS`
- `29 / 18 / 60 FPS`

To address the reentrancy concern from this PR, I added a guard that skips observer state transitions while Winit is already inside `AppStateImpl::InUserCallback`. The guarded version also avoided the 3 to 4 FPS stall:

- `13 / 19 / 49 / 60 FPS`
- `43 / 32 / 59 FPS`

A native `UIActivityViewController` share sheet also opened and closed normally with the guard.

It looks like default mode alone can delay foreground recovery, while common modes need a reentrancy guard. Would this guarded approach make sense here, or should foreground transitions use a separate mechanism?

PR #3771, opened 2024-07-01
Implement smooth resizing on X11 with _NET_WM_SYNC_REQUEST

### LINKED ISSUE #2153: Feature request: X11 Clean Resizing (aka support _NET_WM_SYNC_REQUEST)
I am trying to write a desktop application that uses Winit for windowing, with Vulkan for rendering (though that should be irrelevant).

A limitation I am facing is that on linux, when the window is resized the image appears to jigger (most notably when reszing from top left).
This occurs because x11 completes the resize before getting a chance to draw a new frame of the new size.

<details>
  <summary>A Gif (Note the background flickers a bit)</summary>

jitter

</details>

(My code for this comes from running the sandbox example [here](https://gitlab.com/CyanPhoton/positronium))

On Windows this is solved (properly?) by blocking the event loop when receiving a WindowEvent::Resized event until a separate thread has created a new VkSwapchain and completed drawing a new frame.

This same approach does not work on x11 due to it's apparent asynchronous behavior, instead while the event loop is blocked the window continues to resize. It is possible to recreate the behavior of Windows on x11 as noted in [this](https://www.qt.io/blog/2009/06/10/smooth-and-solid-resizing-on-x11) Qt blog post with corresponding [commit](https://github.com/qt/qt/commit/2312b121131774a84d01854b3e47d3d2a035a6a1).

As far as I know, without this feature it's just not really possible to create an application with smooth resizing on x11.

For an API, possibly just a new option when building a window like `.with_force_block_resize(bool)` to make it so that on x11 (maybe other platforms too) blocking WindowEvent::Resized events for that window, blocks the window from resizing to mimic the Windows behavior. If this is exposed on Windows then `false` can't really revert the behaviour hence the use of `force` in the function name. This would probably be best like this as an opt, as just like the Qt blog notes, if you block for too long it can cause the resizing to seem unresponsive.

Speykious:

Resolves #2153.

I only tested this on KDE so far, this probably needs more tests on other DEs and WMs.

~~One thing I'm unsure about is whether all the widely used DEs and WMs support the sync API. I know KDE and Gnome definitely do, and it seems that awesomewm also does, but I don't know if some other important ones don't. Searching around, I saw that [Qt preemptively checks if the sync extension is supported](https://github.com/qt/qtbase/blob/bcfa0102e5910d99263ebb73515cc9d35b818b5c/src/plugins/platforms/xcb/qxcbwindow.cpp#L420) through a `hasXSync` call, but I couldn't find a way to do that with `x11rb` so right now there's just no checks...~~ Done through suggestion.

~~On KDE, there's also the subtly weird behavior of the window when resizing a certain way that I mentioned in [this C gist](https://gist.github.com/Speykious/58911fb1c029e17f982b6f76ff8003f5) that I shared in #2153, but it looks more like a KDE bug to me than a misuse of the API.~~ ok it actually definitely is a KDE bug because it happens to all windows of all applications regardless of whether they support `_NET_WM_SYNC_REQUEST`.

### DIFF
--- a/src/platform_impl/linux/x11/event_processor.rs
+++ b/src/platform_impl/linux/x11/event_processor.rs
@@ -438,11 +438,18 @@ impl EventProcessor {
                 None => return,
             };
 
-            #[allow(clippy::identity_op)]
-            let lo = (bytemuck::cast::<c_long, c_ulong>(xev.data.get_long(2)) & 0xffffffff) as u32;
-            #[allow(clippy::identity_op)]
-            let hi = (bytemuck::cast::<c_long, c_ulong>(xev.data.get_long(3)) & 0xffffffff) as u32;
-            let hi = bytemuck::cast::<u32, i32>(hi);
+
+            #[cfg(target_pointer_width = "32")]
+            let (lo, hi) = (
+                bytemuck::cast::<c_long, u32>(xev.data.get_long(2)),
+                xev.data.get_long(3),
+            );
+
+            #[cfg(not(target_pointer_width = "32"))]
+            let (lo, hi) = (
+                (xev.data.get_long(2) & 0xffffffff) as u32,
+                bytemuck::cast::<u32, i32>((xev.data.get_long(3) & 0xffffffff) as u32),
+            );
 
             wt.xconn
                 .xcb_connection()
--- a/src/platform_impl/linux/x11/event_processor.rs
+++ b/src/platform_impl/linux/x11/event_processor.rs
@@ -438,8 +438,11 @@ impl EventProcessor {
                 None => return,
             };
 
-            let lo = (xev.data.get_long(2) & 0xffffffff) as u32;
-            let hi = (xev.data.get_long(3) & 0x0fffffff) as i32;
+            #[allow(clippy::identity_op)]
+            let lo = (bytemuck::cast::<c_long, c_ulong>(xev.data.get_long(2)) & 0xffffffff) as u32;
+            #[allow(clippy::identity_op)]
+            let hi = (bytemuck::cast::<c_long, c_ulong>(xev.data.get_long(3)) & 0xffffffff) as u32;
+            let hi = bytemuck::cast::<u32, i32>(hi);
 
             wt.xconn
                 .xcb_connection()
--- a/src/platform_impl/linux/x11/event_processor.rs
+++ b/src/platform_impl/linux/x11/event_processor.rs
@@ -434,16 +434,13 @@ impl EventProcessor {
             let sync_counter_id = match self
                 .with_window(xev.window as xproto::Window, |window| window.sync_counter_id())
             {
-                Some(sync_counter_id) => sync_counter_id,
-                None => return,
+                Some(Some(sync_counter_id)) => sync_counter_id.get(),
+                _ => return,
             };
 
-
             #[cfg(target_pointer_width = "32")]
-            let (lo, hi) = (
-                bytemuck::cast::<c_long, u32>(xev.data.get_long(2)),
-                xev.data.get_long(3),
-            );
+            let (lo, hi) =
+                (bytemuck::cast::<c_long, u32>(xev.data.get_long(2)), xev.data.get_long(3));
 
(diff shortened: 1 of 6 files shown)


src/platform_impl/linux/x11/window.rs
```diff
@@ -469,21 +472,40 @@ impl UnownedWindow {
                 leap!(window.set_icon_inner(icon.inner)).ignore_error();
             }
 
-            // Opt into handling window close
+            // Opt into handling window close and resize synchronization
```
notgull, line 475: Is there a better source of documentation on resize synchronization?

src/platform_impl/linux/x11/window.rs
```diff
@@ -117,6 +118,7 @@ pub struct UnownedWindow {
     root: xproto::Window,               // never changes
     #[allow(dead_code)]
     screen_id: i32, // never changes
+    sync_counter_id: u32,               // never changes
```
notgull, line 121: This should be `Option<NonZeroU32>` to indicate that zero means an absence of this particular feature, or `NonZeroU32` to indicate that it will never be zero.

src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -429,6 +430,27 @@ impl EventProcessor {
...
+            {
+                Some(sync_counter_id) => sync_counter_id,
+                None => return,
+            };
+
+            #[allow(clippy::identity_op)]
+            let lo = (bytemuck::cast::<c_long, c_ulong>(xev.data.get_long(2)) & 0xffffffff) as u32;
+            #[allow(clippy::identity_op)]
+            let hi = (bytemuck::cast::<c_long, c_ulong>(xev.data.get_long(3)) & 0xffffffff) as u32;
+            let hi = bytemuck::cast::<u32, i32>(hi);
```
notgull, line 445: Do we lose anything by doing `as` instead of `bytemuck::cast`?

notgull (changes requested):
> [Qt preemptively checks if the sync extension is supported](https://github.com/qt/qtbase/blob/bcfa0102e5910d99263ebb73515cc9d35b818b5c/src/plugins/platforms/xcb/qxcbwindow.cpp#L420)

We can do this here. Do [`extension_information`](https://docs.rs/x11rb/latest/x11rb/connection/trait.RequestConnection.html#tymethod.extension_information) with [`sync::X11_EXTENSION_NAME`](https://docs.rs/x11rb/latest/x11rb/protocol/sync/constant.X11_EXTENSION_NAME.html) and check for `None`.

I'll test this later once I have more time.

src/platform_impl/linux/x11/window.rs
Speykious: There's https://www.x.org/releases/X11R7.7/doc/xextproto/sync.html and https://fishsoup.net/misc/wm-spec-synchronization.html, but outside of that there's pretty much just direct implementations like Qt's.

src/platform_impl/linux/x11/event_processor.rs
Speykious: The data's format is 32 bit numbers. On 64 bit Linux we could just not skip the `c_ulong` conversion entirely, but on 32 bit Linux the `c_long` and `c_ulong` types are 32 bit, so the last bit being the sign becomes a problem. So this is done to make sure both 64 bit and 32 bit interpret the data correctly.

Alternatively I could make 2 separate codepaths, one for 32 bit and another for 64 bit. Might be easier to read?

src/platform_impl/linux/x11/window.rs
Speykious: Done! Made sense with querying the extension.

src/platform_impl/linux/x11/window.rs
Speykious: Do I maybe add these links to this comment?

notgull (approved):
Tested this on my machine and it appears to work.

Please rebase on `master` and I will merge.

PR #3774, opened 2024-07-03
wayland: ignore events for dead objects

kchibisov:
Nothing wrong will happen if we ignore events when compositor is at wrong, at least crashing because compositor is just _wrong_ probably is not a great option.

Links: https://github.com/alacritty/alacritty/issues/8065

--

Whatever, the other part of this "fix" will require https://github.com/Smithay/client-toolkit/pull/461 , though, it could be patch bumped by the library users.


src/platform_impl/linux/wayland/seat/keyboard/mod.rs
```diff
@@ -32,7 +32,13 @@ impl Dispatch<WlKeyboard, KeyboardData, WinitState> for WinitState {
...
         let seat_state = match state.seats.get_mut(&data.seat.id()) {
-            Some(seat_state) => seat_state,
+            Some(seat_state) => {
+                if seat_state.keyboard_state.is_none() {
+                    warn!("Received keyboard event {event:?} without keyboard");
+                    return;
+                }
+                seat_state
+            },
             None => return,
```
notgull, line 42: Should we also warn on this branch?

src/platform_impl/linux/wayland/seat/pointer/mod.rs
```diff
@@ -42,6 +44,10 @@ impl PointerHandler for WinitState {
     ) {
         let seat = pointer.winit_data().seat();
         let seat_state = self.seats.get(&seat.id()).unwrap();
+        if seat_state.pointer.is_none() {
+            warn!("Received pointer event without pointer");
+            return;
+        }
```
notgull, line 50: Instead of doing `is_none`, we should use a `match` here and out the result into a `let pointer` variable. That way we don't have to `unwrap` below.

notgull (changes requested):

src/platform_impl/linux/wayland/seat/pointer/mod.rs
kchibisov: I refactored that way because I've assumed that it'll fail due to borrowing somehow, however it seems to work.

src/platform_impl/linux/wayland/seat/keyboard/mod.rs
kchibisov: I've made it a bit more broken compositor proof, though, keep in mind that the old code is like that for like ~7 years, it's just some 4chan devs decided to yolo recently.

notgull (approved):

PR #3776, opened 2024-07-05
add `available_monitors` method to `EventLoop`

umut-sahin:

Hi, I need to access the list of monitors, to set up some state, before the app runs. Also, I'm in a Bevy project, so Bevy defines the app, and I cannot alter its behavior. As a result, I cannot easily get a handle to `ActiveEventLoop`, hence I cannot get the list of monitors.

With this change, list of available monitors is accessible directly in the event loop, and I can easily get the list of monitors before the app is run.

### DIFF
--- a/src/changelog/unreleased.md
+++ b/src/changelog/unreleased.md
@@ -43,6 +43,7 @@ changelog entry.
 ### Added
 
 - Add `ActiveEventLoop::create_proxy()`.
+- Add `EventLoop::available_monitors()`.
 
 ### Changed
 
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -299,6 +299,11 @@ impl EventLoop {
     pub fn create_custom_cursor(&self, custom_cursor: CustomCursorSource) -> CustomCursor {
         self.event_loop.window_target().p.create_custom_cursor(custom_cursor)
     }
+
+    /// Returns the list of all the monitors available on the system.
+    pub fn available_monitors(&self) -> impl Iterator<Item = MonitorHandle> {
+        self.event_loop.window_target().available_monitors()
+    }
 }
 
 #[cfg(feature = "rwh_06")]


madsmtm (approved):
Getting such information in AppKit/UIKit before the application is initialized is _probably_ fine, so I'm good with this change.

kchibisov (changes requested):
On Wayland it's generally _async_, so such information could be wrong. And it's generally not recommended to resort to it.

I'm not that against having this API as _deprecated_ back so I can backport to `0.30.x`, because it was obviously forgotten, but as for `0.31` it'll likely go away again, though you'd likely have better APIs for monitors in general.

daxpedda:
> On Wayland it's generally async, so such information could be wrong. And it's generally not recommended to resort to it.

@kchibisov could you elaborate?
It seems to me that monitor data could always be wrong because it can change during runtime. What exactly is wrong with it at this state?

kchibisov:
>It seems to me that monitor data could always be wrong because it can change during runtime. What exactly is wrong with it at this state?

The main difference is that when the event loop is running you're less likely to get a bug due to data being different because compositor tends to give some _grace_ time for you to work with potentially broken data. However when it's paused and you try to work with this data you can generally crash because it's not valid on the compositor side.

From compositor side it basically informed you that data is wrong and started a timer (like 10 seconds, or whatever it decides) when it'll still accept wrong data, however if your event loop is paused you're not aware of such changes and it could result in arbitrary crash, because winit side of things will think that the data is still sort of valid, but on compositor it's dead and was marked long ago as invalid.

The only real solution to all of that is to have the event loop always running when working with the objects which just plug/unplug, so you naturally follow the grace period.

It also means that you generally shouldn't block for too long(10-30 seconds in most compositors iirc) to avoid that.

This particular issue caused crashes when doing vsync with loop running.

--

Probably all of that is just a Wayland issue here, but I really don't see why not init all of that when the event loop is running, so you're less likely to hit/deal with all of that, especially given that in multi monitor setup your output information is pretty much irrelevant since you should live it all up to system compositor on where to open your application.

daxpedda:
> Probably all of that is just a Wayland issue here, but I really don't see why not init all of that when the event loop is running, so you're less likely to hit/deal with all of that

I agree if this causes issues in Wayland that can't be circumvented we should not do this.
There should be very low motivation to expose this information at this level.

daxpedda:
We discussed in the meeting and we established that this won't work on Wayland.
In that light: closing.

PR #3783, opened 2024-07-10
Introduce `FingerId`

daxpedda:
This replaces `Touch::id` with `Touch::finger_id` and introduces a new type `FingerId` instead.

Additionally on Web it exposes [`PointerEvent.isPrimary`](https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/isPrimary) which tells the user if the finger is the primary/first one in a multi-touch interaction.
I also went ahead and implemented this for Windows.

@MarijnS95 pointed out how this could be implemented for Android in https://github.com/rust-windowing/winit/pull/3783#issuecomment-2227447405.
But other OSs would have to be polyfilled.

We could potentially expose this directly to the user without a platform specific extension trait if we implement enough platforms.

**Followup**
Entry in #2875: Firefox emits wrong primary fingers ([Bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1909197)).

### DIFF
--- a/src/platform_impl/web/device.rs
+++ b/src/platform_impl/web/device.rs
@@ -1,8 +1,0 @@
-#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub struct DeviceId(pub i32);
-
-impl DeviceId {
-    pub const fn dummy() -> Self {
-        Self(0)
-    }
-}
--- a/src/changelog/unreleased.md
+++ b/src/changelog/unreleased.md
@@ -56,6 +56,9 @@ changelog entry.
   well.
 - On Android, add `{Active,}EventLoopExtAndroid::android_app()` to access the app used to create the loop.
 - Add `ActiveEventLoop::system_theme()`, returning the current system theme.
+- Add `Touch::finger_id` with a new type `FingerId`.
+- On Web and Windows, add `FingerIdExt*::is_primary()`, exposing a way to determine
+  the primary finger in a multi-touch interaction.
 
 ### Changed
 
@@ -110,6 +113,7 @@ changelog entry.
   v0.5 support. v0.6 remains in place and is enabled by default.
 - Remove `DeviceEvent::Added` and `DeviceEvent::Removed`.
 - Remove `DeviceEvent::Motion` and `WindowEvent::AxisMotion`.
+- Remove `Touch::id` in favor of `Touch::finger_id`.
 
 ### Fixed
 
--- a/src/event.rs
+++ b/src/event.rs
@@ -447,6 +447,26 @@ impl DeviceId {
     }
 }
 
+/// Identifier of a finger in a touch event.
+///
+/// Whenever a touch event is received it contains a `FingerId` which uniquely identifies the finger
+/// used for the current interaction.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct FingerId(pub(crate) platform_impl::FingerId);
+
+impl FingerId {
+    /// Returns a dummy id, useful for unit testing.
+    ///
+    /// # Notes
+    ///
+    /// The only guarantee made about the return value of this function is that
+    /// it will always be equal to itself and to future values returned by this function.
+    /// No other guarantees are made. This may be equal to a real `FingerId`.
+    pub const fn dummy() -> Self {
+        FingerId(platform_impl::FingerId::dummy())
+    }
+}
+
 /// Represents raw hardware events that are not associated with any particular window.
 ///
 /// Useful for interactions that diverge significantly from a conventional 2D GUI, such as 3D camera
@@ -846,7 +866,7 @@ pub struct Touch {
     ///   [android documentation](https://developer.android.com/reference/android/view/MotionEvent#AXIS_PRESSURE).
     pub force: Option<Force>,
     /// Unique identifier of a finger.
-    pub id: u64,
+    pub finger_id: FingerId,
 }
 
 /// Describes the force of a touch event
@@ -1012,6 +1032,7 @@ mod tests {
             #[allow(unused_mut)]
             let mut x = $closure;
             let did = event::DeviceId::dummy();
+            let fid = event::FingerId::dummy();
 
             #[allow(deprecated)]
             {
@@ -1075,7 +1096,7 @@ mod tests {
                     device_id: did,
                     phase: event::TouchPhase::Started,
                     location: (0.0, 0.0).into(),
-                    id: 0,
+                    finger_id: fid,
                     force: Some(event::Force::Normalized(0.0)),
                 }));
                 with_window_event(ThemeChanged(crate::window::Theme::Light));
@@ -1133,6 +1154,7 @@ mod tests {
         let _ = event::StartCause::Init.clone();
 
         let did = crate::event::DeviceId::dummy().clone();
+        let fid = crate::event::FingerId::dummy().clone();
         HashSet::new().insert(did);
         let mut set = [did, did, did];
         set.sort_unstable();
@@ -1148,7 +1170,7 @@ mod tests {
             device_id: did,
             phase: event::TouchPhase::Started,
             location: (0.0, 0.0).into(),
-            id: 0,
+            finger_id: fid,
             force: Some(event::Force::Normalized(0.0)),
         }
         .clone();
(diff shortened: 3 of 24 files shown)


src/platform_impl/web/device.rs
```diff
@@ -1,8 +1,26 @@
+use web_sys::PointerEvent;
+
 #[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub struct DeviceId(pub i32);
+pub struct DeviceId {
```
madsmtm, line 4: This semantically changes `DeviceId` from just being an "identifier", to being a "container", i.e. it should now probably be named `DeviceData`.

The change I'd rather like to see is to change it to a "handle", as I believe it would be useful for e.g. the gamepad work or querying other device information. Ideally, I think that should be done in a separate PR first (i.e. a PR to rename `DeviceId` to `DeviceHandle`, and perhaps store pointers instead of just a `usize` derived from the pointer). Such a change also affects https://github.com/rust-windowing/winit/pull/3784, as now we suddenly have to do extra work to handle possibly-invalid device identifiers (whereas before it was just UB).

kchibisov:
What's the end goal/motivation with it if you mind sharing? Like it's really not clear how you'd use that aside from touch related stuff?

src/platform_impl/web/device.rs
daxpedda: Unfortunately I'm clueless about how exactly this works on other platforms, on Web the device ID has no data associated with it and can't be used for anything else then comparing it to other device IDs. So a device handle would have use on Web.

The `primary` data I'm adding here is actually associated with the event and not with the device ID, but I can't really add it to the event without cross-platform support.

Fortunately the device ID on Web is not used outside those events so it all makes sense.

daxpedda:
> What's the end goal/motivation with it if you mind sharing? Like it's really not clear how you'd use that aside from touch related stuff?

Its only used for touch and multi pointer situations.

I'm considering implementing this cross-platform for touch only and close this.

daxpedda:
So after looking into how exactly browsers implement [`isPrimary`](https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/isPrimary) I discovered that its basically a polyfill where they figure out what the first touch was by tracking finger IDs between touch start and end.

Windows is the only OS that supports detecting it via the [`TOUCHEVENTF_PRIMARY`](https://learn.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-touchinput) and [`POINTER_FLAG_PRIMARY`](https://learn.microsoft.com/en-us/windows/win32/inputmsg/pointer-flags-contants) flag.

So I see the following options:
1. Change [`Touch::id`](https://docs.rs/winit/0.30.3/winit/event/struct.Touch.html#structfield.id) to `FingerId` and implement platforms traits on it.
2. Add `Touch::primary` and polyfill it for all platforms except Web and Windows.
3. Add extension traits on [`Touch`](https://docs.rs/winit/0.30.3/winit/event/struct.Touch.html), which will require us to `Default` or `Touch::new()` if we want to keep it user constructible.

Personally I'm preferring to go with solution 1.
WDYT?
Cc @kchibisov, @madsmtm, @MarijnS95, @notgull.

kchibisov:
But if the primary is the first click couldn't it be entirely user detectable? I also don't really understand how one would use it?

>Its only used for touch and multi pointer situations.

But how it's supposed to be utilized by the user and what they'll do with it? Is it some web limitation preventing it detecting multitouch correctly? because you should generally understand the order of touch events, etc?

Like in JS world I'd assume that users can just track amount of fingers pressed at the given time or something.

daxpedda:
> But if the primary is the first click couldn't it be entirely user detectable?

Yes and no, its not easy to reliably detect on Web, e.g. starting touch point could be outside the canvas etc.



> But how it's supposed to be utilized by the user and what they'll do with it? Is it some web limitation preventing it detecting multitouch correctly? because you should generally understand the order of touch events, etc?
> 
> Like in JS world I'd assume that users can just track amount of fingers pressed at the given time or something.

kchibisov:
>Its only relevant in scenarios where you want to track the primary finger in a multi-touch interaction. This can be common in games, where the primary finger has a certain function that all other fingers don't have, e.g. movement vs shooting.


@madsmtm @MarijnS95  do you happen to know if something like that exists on ios/android? I'd assume users detect all of that themselves?

In general, I don't really mind having solution 1.

MarijnS95:
I don't exactly understand what this "primary" pointer is in Web; is it just the first pointer/touch that started?

In that case, on Android, you could keep track of the first [`pointer_id()`](https://developer.android.com/ndk/reference/group/input#amotionevent_getpointerid) that went down.

There's also the mention for `AMOTION_EVENT_ACTION_POINTER_DOWN` and `AMOTION_EVENT_ACTION_POINTER_UP` that a _non-primary_ pointer has gone down/up. Probably the fist finger triggers `AMOTION_EVENT_ACTION_BUTTON_PRESS`/`AMOTION_EVENT_ACTION_BUTTON_RELEASE` instead?

---

> Its only relevant in scenarios where you want to track the primary finger in a multi-touch interaction. This can be common in games, where the primary finger has a certain function that all other fingers don't have, e.g. movement vs shooting.

On an unrelated note that sounds very confusing: the first touch controls movement, the second controls aim? I always thought that it was position-based: left half of the screen is moving, right half is panning the camera?

In any case it's fully trackable by the user at least.

daxpedda:
> I don't exactly understand what this "primary" pointer is in Web; is it just the first pointer/touch that started?

Yes.

> On an unrelated note that sounds very confusing: the first touch controls movement, the second controls aim? I always thought that it was position-based: left half of the screen is moving, right half is panning the camera?

I have very little experience with games like that, so I can't say much to that.
For our own use-case, where its not position-based, we tracked it ourselves at start, but because we had some controls outside the canvas (for various reasons not in our control), it interfered sometimes.
So we needed to use `isPrimary`.

MarijnS95:
On Android an app is usually fullscreen but it doesn't need to be - we should test what happens when a motion gesture enters and exit the app similar to your canvas example!

src/platform_impl/web/device.rs
daxpedda: Solved by going with a completely different route introducing a new type `FingerId`.

daxpedda:
I went ahead and implemented `FingerId`, ergo solution 1.

MarijnS95:
> In that case, on Android, you could keep track of the first [`pointer_id()`](https://developer.android.com/ndk/reference/group/input#amotionevent_getpointerid) that went down.
> 
> There's also the mention for `AMOTION_EVENT_ACTION_POINTER_DOWN` and `AMOTION_EVENT_ACTION_POINTER_UP` that a _non-primary_ pointer has gone down/up. Probably the fist finger triggers `AMOTION_EVENT_ACTION_DOWN`/`AMOTION_EVENT_ACTION_UP` instead?

After testing this in winit, the results are somewhat surprising regarding this `action()` enum discrepancy:

- On first finger down, a `Down` event is emitted with `pointer_id()` set to `0`;
- On second finger down, a `PointerDown` event is emitted, as expected for a non-primary. The ID is set to `1`;
- When the first finger is removed, which used to be the primary finger, a `PointerUp` is emitted, surprisingly not primary. This event carries `pointer_id() = 0`;
- Subsequent moves for the "second finger" keep being emitted with `pointer_id() = 1` as expected;
- Finally, removing the second finger emits an `Up` event.

Note that if, instead of removing the second finger, another finger is added, non-primary `PointerUp`/`PointerDown` events are still emitted, but `pointer_id() = 0` is reused.

---

I don't know exactly what kind of API `winit` wants here, but if we want to track the first finger in a series of touches, you'll do something like this:

- On `Down` (primary), store its `pointer_id()`;
- On `(Pointer)Up`, reset this value if it is equal to `pointer_id()`;
- `is_primary()` returns `self.primary_pointer == Some(pointer_id())`.

Meaning that no motion is primary anymore, if the primary finger was removed after some point in time (while a touch is still ongoing).

daxpedda:
> EDIT: Since this is a platform extension, there's nothing to `winit` to "want" if the current platform doesn't support it. No need to polyfill, we're better off providing users the actual info that Android provides which is `pointer_id()` (already in there).

I agree!

MarijnS95 (approved):
Looks good on the Android side (very minimal)!

If anyone needs it we could expose whether this is the first ("primary") finger going down or the last (""primary""?) finger to go up. This is also easy for the user to track by keeping a rolling count of how many up and down events they have seen though.

(Which could be a little cheaper if we create a separate extension keyed off the difference between `Down` and `PointerDown`, and `(Pointer)Up` respectively, but would have to store a `bool` for it anyways)

PR #3784, opened 2024-07-10
Make `DeviceId/WindowId::dummy()` safe

daxpedda:
I'm sure this had some historical context, but none of these IDs are passed into anything anymore.
I included this in the Winit 0.30.4 milestone because this isn't a breaking change.

Let me know if this is somehow a bad idea.

### DIFF
--- a/src/changelog/unreleased.md
+++ b/src/changelog/unreleased.md
@@ -68,6 +68,7 @@ changelog entry.
 - Change signature of `EventLoop::run_app`, `EventLoopExtPumpEvents::pump_app_events` and
   `EventLoopExtRunOnDemand::run_app_on_demand` to accept a `impl ApplicationHandler` directly,
   instead of requiring a `&mut` reference to it.
+- `DeviceId::dummy()` and `WindowId::dummy()` are no longer marked `unsafe`.
 
 ### Removed
 
--- a/src/event.rs
+++ b/src/event.rs
@@ -440,16 +440,13 @@ pub struct DeviceId(pub(crate) platform_impl::DeviceId);
 impl DeviceId {
     /// Returns a dummy id, useful for unit testing.
     ///
-    /// # Safety
+    /// # Notes
     ///
     /// The only guarantee made about the return value of this function is that
     /// it will always be equal to itself and to future values returned by this function.
     /// No other guarantees are made. This may be equal to a real `DeviceId`.
-    ///
-    /// **Passing this into a winit function will result in undefined behavior.**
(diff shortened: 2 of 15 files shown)


notgull:
Looks like @madsmtm asked this a while back [here](https://users.rust-lang.org/t/breaking-change-to-remove-unsafe-from-a-function/80987). The verdict is that it isn't a breaking change.

For me, the question is "is there any way in which calling DeviceId::dummy() can put the program in an invalid state"? As far as I know, there isn't. In fact I think it's guaranteed for most backends that `dummy` does not correspond to a real window.

notgull (approved):

madsmtm (approved):
I suspect it was previously `unsafe` because the `WindowId` is storing a `HWND` on Windows, and may have been passing that to window functions in the past? Or is this still done somewhere? If so, then we might have to change that first.

In any case, this is approved for the AppKit/UIKit platforms.

> Looks like @madsmtm asked this a while back

And got it put [into the documentation](https://doc.rust-lang.org/cargo/reference/semver.html#fn-unsafe-safe) 💪.

kchibisov (approved):
It should be fine given that APIs relying on `WindowId` are unsafe, but it doesn't make a `dummy` constructor unsafe because it can not ensure validity anyway, so the whole object is _safe_ to write but unsafe to use since it doesn't retain anything.

So, if the API relies on Device/Window id, because they represent something it should be _unsafe_ on its own, but not the constructor to build such object, because you can not do anything bad when building since it's pod.

src/changelog/unreleased.md
```diff
@@ -65,6 +65,7 @@ changelog entry.
   `ApplicationHandler::resumed/suspended()` are now only emitted by iOS and Web
   and now signify actually resuming/suspending the application.
 - Rename `platform::web::*ExtWebSys` to `*ExtWeb`.
+- `DeviceId::dummy()` and `WindowId::dummy()` are now not `unsafe`.
```
kchibisov, line 68: Probably this is a bit better than double negation.

```suggestion
- Mark `DeviceId::dummy()` and `WindowId::dummy()` as safe.
```

daxpedda: I intentionally avoided the word "safe" because its not 100% clear that we are talking about removing Rust `unsafe`.

```suggestion
- `DeviceId::dummy()` and `WindowId::dummy()` are no longer marked `unsafe`.
```
Doesn't remove the double negation, but I think using the verb "mark" makes it much easier to understand.

PR #3785, opened 2024-07-10
Web: use the word "Web" correctly and consistently

daxpedda:
Removing any public mentions of `web-sys`, because its currently the only backend we support anyway, so this is a historical leftover when we were still supporting `stdweb`.

About the casing of "Web", apparently OED (Oxford English Dictionary) says its capital unless used in a compound word, e.g. "I am surfing the Web." vs "I am a web user.". Unfortunately I can't link to the OED because its behind a paywall. I didn't want to spend time actually figuring out what exactly are compound words or aren't, so I just opted to use capital casing everywhere.


kchibisov (changes requested):
The `WebSys` should go away from the docs as well?

Also, it's not clear why `docsrs` was added.

daxpedda:
> The `WebSys` should go away from the docs as well?

CI was failing because of that, fixed it now.

> Also, it's not clear why `docsrs` was added.

Because when `cfg(docsrs)` is enabled it can be linked, that's the whole purpose of this `cfg`, to avoid linking to it when its not present.

daxpedda:
> But it'll fail when building for docsrs because `Web` is not present?

But Web is present for `docsrs`?
See #3076 if you don't remember.
Or if you have a way to reproduce a error/warning please share it.

> Like I don't see what was wrong with the old `cfg` check without `docsrs`.

This is just an improvement.

kchibisov (approved):
Ah, forgot that we check docsrs, should be good.

Issue #3779, opened 2024-07-08
Cross-platform consistency for `ApplicationHandler::suspended/resumed()`


daxpedda:
Unfortunately the behavior between Android, iOS and Web doesn't match exactly, so this will require some research and figuring out exactly how we want to expose this to the user.

### Web

Web currently implements this via [`pagehide`](https://developer.mozilla.org/en-US/docs/Web/API/Window/pagehide_event) and [`pageshow`](https://developer.mozilla.org/en-US/docs/Web/API/Window/pageshow_event). Which for `suspended()` means the application is being navigated away from and is being stored in the [B/F Cache](https://web.dev/bfcache) and for `resumed()` means that the page has been restored.

Its important to note that between `suspended()` and `restored()` no user code can be executed and getting `restored()` isn't actually promised.

There is also the [Page Lifecycle API](https://developer.chrome.com/docs/web-platform/page-lifecycle-api), currently only implemented in Chrome, which lets us also detect if the application was frozen by the browser.

### iOS

@madsmtm please feel free to work out this section and correct me.

[`applicationDidEnterBackground`](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1622997-applicationdidenterbackground?language=objc) could be used to detect when iOS is telling the application to stop doing anything. This doesn't mean that any code execution will be suspended immediately but "probably soon".

This however is currently not implemented correctly and instead iOS will call `ApplicationHandler::suspended()` when the application is running in the background via [`applicationWillResignActive`](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1622950-applicationwillresignactive?language=objc), which should probably be `WindowEvent::Focused(false)` instead.

Presumably its also guaranteed that the application will be woken up before being actually closed (?).

### Android

@MarijnS95 your input would be greatly appreciated here.

Android does currently not implement `ApplicationHandler::suspended/resumed()`

MarijnS95:
@daxpedda just in case; you want context after #3765, i.e. knowing that we moved _surface lifetime_ away from the `suspended() and `resumed()` events that Android was previously relying on?

From the look and sound of it you'll want `onStop` and `onStart` from the activity lifecycle: https://developer.android.com/guide/components/activities/activity-lifecycle

daxpedda:
From the documentation I can't seem to be able to definitely say that all code execution is halted after `onStop`, it seems more similar to iOS in that regard, that execution will stop "soon". Am I getting this right?

[...]

PR #3786, opened 2024-07-10
android: Forward `suspended()` and `resumed()` events and patch up platform-specific documentation

MarijnS95:
Addresses #3779

Key them off of `onStop()` and `onStart()` which seems to match the other backends most closely.  These [Android Activity lifecycle] events denote when the application is visible on-screen, and recommend that any heavy lifting for startup and shutdown happens here, as the application may be demoted to the background and later shut down entirely unless the user navigates back to it.

EDIT: Also addresses some of the late review comments that were posted on https://github.com/rust-windowing/winit/pull/3765.

[Android Activity lifecycle]: https://developer.android.com/guide/components/activities/activity-lifecycle

### DIFF
--- a/src/application.rs
+++ b/src/application.rs
@@ -41,11 +41,21 @@ pub trait ApplicationHandler {
     /// [`pageshow`]: https://developer.mozilla.org/en-US/docs/Web/API/Window/pageshow_event
     /// [`bfcache`]: https://web.dev/bfcache/
     ///
+    /// ### Android
+    ///
+    /// On Android, the [`resumed()`] method is called when the `Activity` is (again, if after a
+    /// prior [`suspended()`]) being displayed to the user. This is a good place to begin drawing
+    /// visual elements, running animations, etc. It is driven by Android's [`onStart()`] method.
+    ///
+    /// [`onStart()`]: https://developer.android.com/reference/android/app/Activity#onStart()
+    ///
     /// ### Others
     ///
-    /// **Android / macOS / Orbital / Wayland / Windows / X11:** Unsupported.
+    /// **macOS / Orbital / Wayland / Windows / X11:** Unsupported.
     ///
-    /// [`resumed()`]: Self::resumed
+    /// [`resumed()`]: Self::resumed()
+    /// [`suspended()`]: Self::suspended()
+    /// [`exiting()`]: Self::exiting()
     fn resumed(&mut self, event_loop: &dyn ActiveEventLoop) {
         let _ = event_loop;
     }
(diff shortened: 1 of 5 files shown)


madsmtm (approved):
Nice, feel free to merge once you think it's ready!

daxpedda:
(I changed OP to not fix #3779, it just addresses it)

MarijnS95:
> (I changed OP to not fix #3779, it just addresses it)

I thought this was fixing the issue entirely but it seems there's still more to address for Web/iOS, and maybe together with more Android changes like #3897?

daxpedda:
> @daxpedda let me know if you can review this, it should address all except one (duplicated) comment from [#3765 (review)](https://github.com/rust-windowing/winit/pull/3765#pullrequestreview-2152235928).

I took the whole day today to do some catch up on my open source responsibilities, so if I don't manage to do it today feel free to go ahead.

src/application.rs
```diff
@@ -41,11 +41,21 @@ pub trait ApplicationHandler {
     /// [`pageshow`]: https://developer.mozilla.org/en-US/docs/Web/API/Window/pageshow_event
     /// [`bfcache`]: https://web.dev/bfcache/
     ///
+    /// ### Android
+    ///
+    /// On Android, the [`resumed()`] method is called when the `Activity` is (again, if after a
+    /// prior [`suspended()`]) being displayed to the user.  This is a good place to begin drawing
+    /// visual elements, running animations, etc.  It is driven by Android's [`onStart()`] method.
```
daxpedda, line 48:
```suggestion
    /// visual elements, running animations, etc. It is driven by Android's [`onStart()`] method.
```

src/application.rs
```diff
@@ -71,26 +81,24 @@ pub trait ApplicationHandler {
...
+    /// [`NativeWindow`]: https://developer.android.com/ndk/reference/group/a-native-window
+    /// [`Surface`]: https://developer.android.com/reference/android/view/Surface
+    /// [`onStart`]: https://developer.android.com/reference/android/app/Activity#onStart()
     /// [`VkSurfaceKHR`]: https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkSurfaceKHR.html
     /// [`wgpu::Surface`]: https://docs.rs/wgpu/latest/wgpu/struct.Surface.html
     ///
-    /// [`can_create_surfaces()`]: Self::can_create_surfaces
-    /// [`destroy_surfaces()`]: Self::destroy_surfaces
+    /// [`can_create_surfaces()`]: Self::can_create_surfaces()
+    /// [`destroy_surfaces()`]: Self::destroy_surfaces()
```
daxpedda, line 101: Not that it matters, but why was adding the `()` on the right side necessary?
I would prefer some consensus we can apply to everything, personally I have no preference either way. Currently I believe we mostly omit `()` unless its user visible.

src/application.rs
```diff
@@ -235,18 +243,30 @@ pub trait ApplicationHandler {
     /// ### Web
     ///
     /// On Web, the [`suspended()`] method is called in response to a [`pagehide`] event if the
-    /// page is being restored from the [`bfcache`] (back/forward cache) - an in-memory cache that
+    /// page is being moved/stored in the [`bfcache`] (back/forward cache) - an in-memory cache that
```
daxpedda, line 246:
```suggestion
    /// page is being stored in the [`bfcache`] (back/forward cache) - an in-memory cache that
```
Should be enough I believe.

src/application.rs
```diff
@@ -235,18 +243,30 @@ pub trait ApplicationHandler {
...
     /// stores a complete snapshot of a page (including the JavaScript heap) as the user is
     /// navigating away.
     ///
     /// [`pagehide`]: https://developer.mozilla.org/en-US/docs/Web/API/Window/pagehide_event
     /// [`bfcache`]: https://web.dev/bfcache/
     ///
+    /// ### Android
+    ///
+    /// On Android, the [`suspended()`] method is called when the `Activity` is no longer visible to
+    /// the user.  This is a good place to stop refreshing UI, running animations and other visual
```
daxpedda, line 256:
```suggestion
    /// the user. This is a good place to stop refreshing UI, running animations and other visual
```

src/application.rs
```diff
@@ -235,18 +243,30 @@ pub trait ApplicationHandler {
...
+    /// things.  It is driven by Android's [`onStop()`] method.
```
daxpedda, line 257:
```suggestion
    /// things. It is driven by Android's [`onStop()`] method.
```

src/application.rs
```diff
@@ -259,33 +279,31 @@ pub trait ApplicationHandler {
...
     ///
-    /// Applications that need to run on Android should assume their [`SurfaceView`] has been
+    /// Applications that need to run on Android should assume their [`NativeWindow`] has been
     /// destroyed, which indirectly invalidates any existing render surfaces that may have been
     /// created outside of Winit (such as an `EGLSurface`, [`VkSurfaceKHR`] or [`wgpu::Surface`]).
     ///
-    /// After being [suspended] on Android applications must drop all render surfaces before
-    /// the event callback completes, which may be re-created when the application is next
-    /// [resumed].
+    /// When receiving [`destroy_surfaces()`] Android applications must drop all render surfaces
```
daxpedda, line 291:
```suggestion
    /// When receiving [`destroy_surfaces()`] Android applications should drop all render surfaces
```
I don't remember the exact conclusion we came to, I believe if the surface isn't destroyed its fine, but it will just be useless.
This is important if the surface is stored in a different thread, where implementing a whole blocking mechanism to finish destroying the surface before the callback finishes might undesirable.

daxpedda (approved):
Great stuff!

src/application.rs
MarijnS95: Fair enough, separating individual sentences with a double-space is a habit that stuck with me since Linux kernel development days. A regex for `\.  ` shows that this only ever happened in two other places, so I've removed all these.

src/application.rs
MarijnS95: Not necessary, but it should make `rustdoc` complain if the function is ever replaced with a field; see this as "type safety" for doc comments.

Unfortunately, if there's ever ambiguity when _both_ a field and function exist, both `Self::destroy_surfaces` and `Self::destroy_surfaces()` link to the function and a disambiguator for that was only added last month: https://github.com/rust-lang/rust/pull/130587

daxpedda (approved):

MarijnS95:
Thanks for all the reviews!

PR #3787, opened 2024-07-10
Android: Show keyboard with WindowInsetsController.

xorgy:

Should resolve the Android part of #1823 but it's not real IME.

### DIFF
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -12,6 +12,7 @@ use android_activity::input::{InputEvent, KeyAction, Keycode, MotionAction};
 use android_activity::{
     AndroidApp, AndroidAppWaker, ConfigurationRef, InputStatus, MainEvent, Rect,
 };
+use jni::{objects::JObject, JavaVM};
 use tracing::{debug, trace, warn};
 
 use crate::application::ApplicationHandler;
@@ -757,7 +758,6 @@ fn show_hide_keyboard_fallible(app: AndroidApp, show: bool) -> Result<(), jni::e
     // After Android R, it is no longer possible to show the soft keyboard
     // with `showSoftInput` alone.
     // Here we use `WindowInsetsController`, which is the other way.
-    use jni::{objects::JObject, JavaVM};
     let vm = unsafe { JavaVM::from_raw(app.vm_as_ptr() as _)? };
     let activity = unsafe { JObject::from_raw(app.activity_as_ptr() as _) };
     let mut env = vm.attach_current_thread()?;


src/changelog/unreleased.md
```diff
@@ -53,3 +53,7 @@ changelog entry.
 ### Fixed
 
 - On macOS, fix panic on exit when dropping windows outside the event loop.
+
+### Changed
+
+- On Android, summon and dismiss the soft keyboard with `set_ime_allowed`
```
kchibisov, line 59: it's Added. Should probably be reworded once moved to added.

xorgy:
I'm a little unsatisfied that this is activated with `set_ime_allowed`, but doesn't actually cause any Ime events to happen... but AFAIK that is the behavior on Wayland when you have a simple soft keyboard as well.

src/platform_impl/android/mod.rs
```diff
@@ -747,9 +753,35 @@ impl DeviceId {
...
+        .l()
+        .unwrap();
+    let window_insets_types = env.find_class("android/view/WindowInsets$Type").unwrap();
+    let ime_type =
+        env.call_static_method(&window_insets_types, "ime", "()I", &[]).unwrap().i().unwrap();
+    let _ = if show {
+        env.call_method(&wic, "show", "(I)V", &[ime_type.into()])
+    } else {
+        env.call_method(&wic, "hide", "(I)V", &[ime_type.into()])
+    };
```
kchibisov, line 778: Would be nice to have some docs on why it's done like that, also, do we really need that many unwraps here?

kchibisov (changes requested):
Is it just shown or some input _works_?

kchibisov:
>I'm a little unsatisfied that this is activated with set_ime_allowed, but doesn't actually cause any Ime events to happen... but AFAIK that is the behavior on Wayland when you have a simple soft keyboard as well.

Well, enabling IME doesn't mean that you'll have IME input, because you still enable it for `us` layout, because you don't actually know whether it'll happen or not. It's more of a hint that you're doing text input. Disabling it on the other side is used when you want just hotkeys to work, like in games, but you still toggle it back when you type in game chat or something, so it seems logical, I'd say, maybe just name is not ideal.

src/platform_impl/android/mod.rs
xorgy: Sure, I'll both check into whether there is a more ergonomic API available, and document why it has to be done this way. I think that the `jni` APIs are meant to be used with the `?` operator, so making the function fallible may work.

xorgy:
> Is it just shown or some input _works_?

It produces key events, and you can type with it (including with some alternate layouts). It only works for keyboards that produce these and are not IME-only (e.g. US, US Dvorak, MX-es Gboard, etc.).

xorgy:
@mwcampbell is going to get an accesskit_winit branch ready for winit master (with the user events change) and I'll bring the branch up to date with that and test it again. I'll have your comments addressed by then. :+ )

kchibisov:
>branch ready for winit master

I'll backport this to 0.30.x branch, because it's an addition and doesn't break anything, so if you can test with only 0.30.x in your setup it's fine, though you can test with glutin's android example and see the typing in the console or something or whether it pops when calling.

podusowski:
Does hiding the keyboard work as expected for you? I have ported it to 0.29 to test it with my winit/egui stack and it crashes with:
```
07-14 14:27:00.856 15745 15802 E widnet : winit::platform_impl::platform: Showing or hiding the soft keyboard failed: JavaException
```

```
07-14 14:27:00.867 15745 15802 F al.widnet.debu: java_vm_ext.cc:591] JNI DETECTED ERROR IN APPLICATION: JNI GetObjectClass called with pending exception android.view.ViewRootImpl$CalledFromWrongThreadException: Only the original thread that created a view hierarchy can touch its views.
07-14 14:27:00.867 15745 15802 F al.widnet.debu: java_vm_ext.cc:591] at void android.view.ViewRootImpl.checkThread() (ViewRootImpl.java:9728)
07-14 14:27:00.867 15745 15802 F al.widnet.debu: java_vm_ext.cc:591] at void android.view.ViewRootImpl.requestLayout() (ViewRootImpl.java:1883)
07-14 14:27:00.867 15745 15802 F al.widnet.debu: java_vm_ext.cc:591] at void android.view.ViewRootImpl.notifyInsetsChanged() (ViewRootImpl.java:1864)
07-14 14:27:00.867 15745 15802 F al.widnet.debu: java_vm_ext.cc:591] at void android.view.ViewRootInsetsControllerHost.notifyInsetsChanged() (ViewRootInsetsControllerHost.java:55)
07-14 14:27:00.867 15745 15802 F al.widnet.debu: java_vm_ext.cc:591] at void android.view.InsetsController.notifyVisibilityChanged() (InsetsController.java:1307)
07-14 14:27:00.867 15745 15802 F al.widnet.debu: java_vm_ext.cc:591] at void android.view.InsetsSourceConsumer.setRequestedVisible(boolean) (InsetsSourceConsumer.java:387)
07-14 14:27:00.867 15745 15802 F al.widnet.debu: java_vm_ext.cc:591] at void android.view.InsetsSourceConsumer.hide() (InsetsSourceConsumer.java:229)
07-14 14:27:00.867 15745 15802 F al.widnet.debu: java_vm_ext.cc:591] at void android.view.ImeInsetsSourceConsumer.hide() (ImeInsetsSourceConsumer.java:68)
07-14 14:27:00.867 15745 15802 F al.widnet.debu: java_vm_ext.cc:591] at void android.view.ImeInsetsSourceConsumer.hide(boolean, int) (ImeInsetsSourceConsumer.java:74)
07-14 14:27:00.867 15745 15802 F al.widnet.debu: java_vm_ext.cc:591] at void android.view.InsetsController.hideDirectly(int, boolean, int, boolean) (InsetsController.java:1426)
07-14 14:27:00.867 15745 15802 F al.widnet.debu: java_vm_ext.cc:591] at void android.view.InsetsController.controlAnimationUnchecked(int, android.os.CancellationSignal, android.view.WindowInsetsAnimationControlListener, android.graphics.Rect, boolean, long, android.view.animation.Interpolator, int, int, boolean) (InsetsController.java:1092)
07-14 14:27:00.867 15745 15802 F al.widnet.debu: java_vm_ext.cc:591] at void android.view.InsetsController.applyAnimation(int, boolean, boolean, boolean) (InsetsController.java:1409)
07-14 14:27:00.867 15745 15802 F al.widnet.debu: java_vm_ext.cc:591] at void android.view.InsetsController.applyAnimation(int, boolean, boolean) (InsetsController.java:1390)
07-14 14:27:00.867 15745 15802 F al.widnet.debu: java_vm_ext.cc:591] at void android.view.InsetsController.hide(int, boolean) (InsetsController.java:958)
07-14 14:27:00.867 15745 15802 F al.widnet.debu: java_vm_ext.cc:591] at void android.view.InsetsController.hide(int) (InsetsController.java:933)
07-14 14:27:00.867 15745 15802 F al.widnet.debu: java_vm_ext.cc:591]
07-14 14:27:00.867 15745 15802 F al.widnet.debu: java_vm_ext.cc:591] in call to GetObjectClass
```

Also,  typing in egui does now work, because egui [expects `KeyEvent::text` to be filled](https://github.com/emilk/egui/blob/67eb29001dcb13ee54ca3681bb790f07501be341/crates/egui-winit/src/lib.rs#L775), which is not in the [android impl](https://github.com/rust-windowing/winit/blob/ee3ab33a7c7b72983bffc494ef8a46cc3aa4f698/src/platform_impl/android/mod.rs#L402). But I guess it's a different story.

Anyway, great work on pushing the keyboard input forward.

xorgy:
@podusowski seems threading related, by the message you copied here.
Could you
 - Say what version of Android are you running on
 - Share the application or a skeleton of it? (so that I can take a look)
 
 Cheers

xorgy:
As for the `text` field being filled, Masonry handles `Key::Character` which is why it's working for us. Maybe it makes sense to do some mapping (at least copy the contents of `Key::Character` when it is there).

kchibisov:
Probably can just feel the text with the `logical_key`? Since generally the input is done via `text` and not `logical_key`, because you have compose and such on desktop platforms which is routed via the `text` field.

kchibisov:
>seems threading related, by the message you copied here

Do you by any chance somehow create the `EventLoop` of the main thread? Like inside the `std::thread` or something like that? It looks like `Android` backend doesn't have a check to ensure the thread it's running on, so if you manually speculate with threads it won't notice that.

@MarijnS95 do you happen to know whether android is required to use the main thread? And in general, I think we should have a way to ensure that it's running on the main thread. We can just use the same linux code with `gettid == getpid` I guess?

podusowski:
> Do you by any chance somehow create the `EventLoop` of the main thread? Like inside the `std::thread` or something like that?

@kchibisov No, I don't think so. It's a pretty typical eframe/egui app. I'll chop out some demo once I find some spare time, because currently it's on a private repo.

> Probably can just feel the text with the logical_key?

Yeah, it works nice.

> Could you say what version of Android are you running on

@xorgy It's 12. Running on Mi Note 10 Lite.

lucasmerlin:
Great work! Does this work with NativeActivity and GameActivity?

xorgy:
> Great work! Does this work with NativeActivity and GameActivity?

I've tested it with `NativeActivity`, but since it goes directly to the `Window` it should work with any activity.

podusowski:
I've created a demo based on winit 0.29 and egui, showing a crash here:

https://github.com/podusowski/winit/tree/pr3787-with-egui-crash

it's in `demo_android`. To run it, call

```
make -C demo_android run-on-device
```

I haven't got time to dig into it so no idea from me yet why this is happening.

@lucasmerlin It's `GameActivity`, so I guess it's a second yes. :)

xorgy:
@kchibisov 
> ... do you happen to know whether android is required to use the main thread? And in general, I think we should have a way to ensure that it's running on the main thread. We can just use the same linux code with `gettid == getpid` I guess?

You figure the same check should be added? Would you want to set up an any_thread type thing for Android as well?

kchibisov:
>You figure the same check should be added? Just panic when you start the event loop off the main thread? I think this is still an issue if set_ime_allowed is called off the main thread.

I'm not familiar with how android does things, like it probably won't hurt, but if it always spawns some thread for you which is not the same as `pid` it could break a lot of things, so I generally won't try to change anything, until I know that it's safe.

kchibisov:
@podusowski just in case, you don't have anything duplicated from Android side of things, like `jni`, etc?

I can't really help with android stuff myself, just checking common mistakes.

podusowski:
I have checked the process/thread calls the function and they are the same in both:

```
07-20 21:55:50.830 19863 19945 I widnet  : winit::platform_impl::platform: show_hide_keyboard(true) pid: ThreadId(2) tid: 19863 -> Ok(())
07-20 21:55:52.048 19863 19945 I widnet  : winit::platform_impl::platform: show_hide_keyboard(false) pid: ThreadId(2) tid: 19863 -> Err(JavaException)
```

No idea how it relates to "thread that created a view hierarchy", maybe the whole thing runs on a different, at least in `GameActivity`.

Anyway, I did a test it with `runOnUiThread` and I do not see an exception anymore:
```
    public void showKeyboard() {
        runOnUiThread(new Runnable() {
            @Override
            public void run() {
                Window window = getWindow();
                WindowInsetsController wic = window.getInsetsController();
                wic.show(WindowInsets.Type.ime());
            }
        });
    }
```

I don't even know how to start moving this to jni though.

xorgy:
@podusowski 

> I don't even know how to start moving this to jni though.

I think it would involve injecting Dalvik bytecode, or need to be built into the `Activity`, though there may be another way.

It is fascinating that showing it succeeds on your end, but hiding it fails. I'll have a think on how something similar to what you're suggesting could be done with plain JNI.

kchibisov:
I wonder how C++ stuff solves it, like in QT, since they likely need to workaround it and have a way to run things on the UI thread. I'm not really surprised though that something like that was needed, since macOS/ios/windows do require the exact same thing...

MarijnS95:
> @MarijnS95 do you happen to know whether android is required to use the main thread? And in general, I think we should have a way to ensure that it's running on the main thread. We can just use the same linux code with `gettid == getpid` I guess?

For UI stuff it's been a long time ago, but it seems `runOnUiThread()` was already proposed which exists "exactly" for this purpose?

The thread is spawned off the "main thread for that `Activity`" [by `android-activity`](https://github.com/rust-mobile/android-activity/blob/0d299300f4120821ae1fcaaf0276129c512c2c96/android-activity/src/native_activity/glue.rs#L853), because we "have to" return quickly from that `onCreate` callback. In contrast, inside the users' Rust code, the `Looper` (i.e. your winit loop) is ran.

At some point I've been wondering if it makes more sense to expose these lower level details to apps and `winit` specifically, instead of hiding this thread creation inside `android-activity`. That probably goes paired with untangling the current `onCreate()` != `main()` mess.

podusowski:
@MarijnS95 You mean like exposing looper attached to the main thread (assuming I understand this stuff correctly) or something in `AndroidApp` or move the creation of this thread outside of `android-activity` completely?

podusowski:
I've made some experiments and came up with something like this:

In `android-activity/game-activity-csrc/game-activity/native_app_glue/android_native_app_glue.c`, just before the app thread is created, I saved a pointer to the looper attached to the main thread:

```cpp
android_app->main_thread_looper = ALooper_forThread();
```

Then, I used this looper to run a Rust callback on the main thread like this:

```rust
fn show_hide_keyboard_fallible(app: AndroidApp, show: bool) -> Result<(), jni::errors::Error> {
    let looper = unsafe { ForeignLooper::from_ptr(app.main_thread_looper()) };

    let (mut tx, mut rx) = pipe_channel::channel::<String>();
    looper.add_fd_with_callback(
        unsafe { BorrowedFd::borrow_raw(rx.as_raw_fd()) },
        ndk::looper::FdEvent::INPUT,
        |fd, _| {
            let s = rx.recv().unwrap();

            log::info!(
                "got keyboard show/hide event in pid: {:?} tid: {}: {s}",
                std::thread::current().id(),
                std::process::id()
            );

            // After Android R, it is no longer possible to show the soft keyboard
            // with `showSoftInput` alone.
            // Here we use `WindowInsetsController`, which is the other way.
            let vm = unsafe { JavaVM::from_raw(app.vm_as_ptr() as _).unwrap() };
            let activity = unsafe { JObject::from_raw(app.activity_as_ptr() as _) };
            let mut env = vm.attach_current_thread().unwrap();
            let window = env
                .call_method(&activity, "getWindow", "()Landroid/view/Window;", &[])
                .unwrap()
                .l()
                .unwrap();
            let wic = env
                .call_method(
                    window,
                    "getInsetsController",
                    "()Landroid/view/WindowInsetsController;",
                    &[],
                )
                .unwrap()
                .l()
                .unwrap();
            let window_insets_types = env.find_class("android/view/WindowInsets$Type").unwrap();
            let ime_type = env
                .call_static_method(&window_insets_types, "ime", "()I", &[])
                .unwrap()
                .i()
                .unwrap();
            env.call_method(&wic, &s, "(I)V", &[ime_type.into()])
                .unwrap()
                .v();

            false
        },
    );

    std::thread::sleep(Duration::from_secs(1));

    log::info!(
        "sending keyboard show/hide event from pid: {:?} tid: {}",
        std::thread::current().id(),
        std::process::id()
    );

    if show {
        tx.send("show".to_string());
    } else {
        tx.send("hide".to_string());
    }

    std::thread::sleep(Duration::from_secs(5));

    Ok(())
}
```

Now, obviously this has these `unwraps` and `sleeps` (not to drop everything before the callback runs) which must go away, and I had to patch `android_native_app_glue.c`, but at least I can confirm that it works.

Zoxc:
Why does this not use `android-activity`'s methods to show/hide the keyboard? It will then start to deliver text events, even if it's from the rather limited `GameTextInput` API.

xorgy:
@Zoxc 
> Why does this not use `android-activity`'s methods to show/hide the keyboard? It will then start to deliver text events, even if it's from the rather limited `GameTextInput` API.

On NativeActivity at least, it's a no-op because of a platform change in Android 11.

Zoxc:
Still that seems like something that should be fixed with `android-activity`s `NativeActivity` handling?

podusowski:
> Still that seems like something that should be fixed with `android-activity`s `NativeActivity` handling?

I think either way, some changes are needed in `android-activity` crate.

In:

https://github.com/rust-windowing/winit/pull/3787#issuecomment-2254135276

`show_hide_keyboard_fallible` is garbage, but there seem to be no other way of doing something on the main thread than attaching to its looper. At least this is my understanding, based on a nearly non-existing experience in writing stuff for Android.

xorgy:
> `show_hide_keyboard_fallible` is garbage, but there seem to be no other way of doing something on the main thread than attaching to its looper. At least this is my understanding, based on a nearly non-existing experience in writing stuff for Android.

Well, that's true in some sense, but it's also not that hard to initiate the call from the main thread. The way we do this in Masonry is to have things like this activated by a queue that is processed in the main thread.

I am working on a new prebuilt activity (kinda like the proposed `RustActivity`) that sets up IME and AccessKit, but is otherwise similar to `NativeActivity`, and that will be part of our long term solution.

MarijnS95:
> Should be fixed with [...] as in, this should be in `android-activity`

Yes. If there's any hack required to get this working that's specific to `NativeActivity` and/or doesn't translate to `GameActivity`, this should be abstracted away inside `android-activity`. Perhaps also to keep some JNI workarounds out of `winit`, at least for now.

That should also more cleanly allow you to attach a callback to the "main UI thread" `Looper` (which could use a `pipe` to handle a variety of events).

---

> I am working on a new prebuilt activity (kinda like the proposed `RustActivity`) that sets up IME and AccessKit, but is otherwise similar to `NativeActivity`, and that will be part of our long term solution.

This is going to be contributed to `android-activity` and reuse existing primitives and abstractions right? And for embedding (perhaps prebuilt) Java `.class`/`.jar` files into existing build-tools like `xbuild`/`cargo-apk` via the proposed `println!("xbuild:java-lib=path/to/My.class")` in `build.rs` for those tools to merge into a single `classes.dex` via `d8` (as somewhat suggested with https://github.com/MarijnS95/android-support and other references to it)?

xorgy:
My intention is to contribute it to android-activity, and for it to reuse as much as possible there. The other thing is I want it to work before there's any interesting build infrastructure added, so while I'm excited for your work on that, I'm starting with the basics (as I described above).

MarijnS95:
>  The other thing is I want it to work before there's any interesting build infrastructure added, so while I'm excited for your work on that, I'm starting with the basics (as I described above).

Ah I missed that but am glad to hear it; I've got some ideas and things in-progress and I think we can just meet in the middle in the end 🎉

jb55:
What's the status of this?

xorgy:
@jb55 

> What's the status of this?

Primary issue is that because this interacts with the `View`, it must be called on the thread that creates the `View` (which is the thread where the events are processed); some toolkit users don't trigger effects like this from the the event loop thread, which causes it to crash. This is resolvable a number of ways but interest was lost.

MarijnS95:
> it must be called on the thread that creates the `View` (which is the thread where the events are processed)

Even with `android-activity`, the process/thread that it sets up to call `fn main()` where `winit` runs the `Looper` and should be receiving input events from the `InputQueue`, is a separate thread from the initial Android/JNI thread where `Activity::onCreate()` called `ANativeActivity_onCreate()` in native code, right?

That JNI thread should be the one where the view was created, before `NativeActivity` (on the Java side) stole away the `Surface` for the window -> is there even a `View` there or is it just a `Window`?

DougAnderson444:
> @jb55
> 
> > What's the status of this?
> 
> This is resolvable a number of ways but interest was lost.

There is definitely still interest! ❤️

Could those resolution ways that you are referring to be listed so we can make a plan? 

I'm new to this issue, but would love to see a plan for this go through.

DougAnderson444:
> This is resolvable a number of ways

@xorgy which way is the most viable option in your view?

chanmaoganda:
waiting for further works

PR #3795, opened 2024-07-17
Replace `WindowId` `From/Into` `u64` with `WindowId::into_raw()` and `from_raw()`

daxpedda:
As there shouldn't really be a need for a user-constructible `WindowId`, I assume this was used previously because `WindowId` was passed as a parameter?

~~I introduced `WindowId::to_u64()` as a replacement, but I'm not sure if this has any use case.~~
After the last meeting we decided to add conversion methods called `from_raw()` and `into_raw()`.

### DIFF
--- a/src/changelog/unreleased.md
+++ b/src/changelog/unreleased.md
@@ -62,6 +62,7 @@ changelog entry.
 - Implement `Clone`, `Copy`, `Debug`, `Deserialize`, `Eq`, `Hash`, `Ord`, `PartialEq`, `PartialOrd`
   and `Serialize` on many types.
 - Add `MonitorHandle::current_video_mode()`.
+- Add `WindowId::into_raw()` and `from_raw()`.
 
 ### Changed
 
@@ -127,6 +128,8 @@ changelog entry.
 - Remove `MonitorHandle::size()` and `refresh_rate_millihertz()` in favor of
   `MonitorHandle::current_video_mode()`.
 - On Android, remove all `MonitorHandle` support instead of emitting false data.
+- Remove `impl From<u64> for WindowId` and `impl From<WindowId> for u64`. Replaced with
+  `WindowId::into_raw()` and `from_raw()`.
 
 ### Fixed
 
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -673,16 +673,12 @@ impl WindowId {
     pub const fn dummy() -> Self {
         WindowId
     }
-}
 
-impl From<WindowId> for u64 {
-    fn from(_: WindowId) -> Self {
+    pub const fn into_raw(self) -> u64 {
         0
     }
-}
 
-impl From<u64> for WindowId {
-    fn from(_: u64) -> Self {
+    pub const fn from_raw(_id: u64) -> Self {
         Self
     }
 }
--- a/src/window.rs
+++ b/src/window.rs
@@ -93,10 +93,16 @@ impl WindowId {
         WindowId(platform_impl::WindowId::dummy())
     }
 
+    /// Convert the `WindowId` into the underlying integer.
+    ///
+    /// This is useful if you need to pass the ID across an FFI boundary, or store it in an atomic.
     pub const fn into_raw(self) -> u64 {
         self.0.into_raw()
     }
 
+    /// Construct a `WindowId` from the underlying integer.
+    ///
+    /// This should only be called with integers returned from [`WindowId::into_raw`].
     pub const fn from_raw(id: u64) -> Self {
         Self(platform_impl::WindowId::from_raw(id))
     }
--- a/src/window.rs
+++ b/src/window.rs
@@ -92,23 +92,19 @@ impl WindowId {
     pub const fn dummy() -> Self {
         WindowId(platform_impl::WindowId::dummy())
     }
+
+    pub fn into_raw(self) -> u64 {
+        self.0.into_raw()
+    }
+
+    pub const fn from_raw(id: u64) -> Self {
+        Self(platform_impl::WindowId::from_raw(id))
+    }
 }
 
 impl fmt::Debug for WindowId {
     fn fmt(&self, fmtr: &mut fmt::Formatter<'_>) -> fmt::Result {
         self.0.fmt(fmtr)
-    }
-}
-
-impl From<WindowId> for u64 {
-    fn from(window_id: WindowId) -> Self {
-        window_id.0.into()
-    }
-}
-
-impl From<u64> for WindowId {
-    fn from(raw_id: u64) -> Self {
-        Self(raw_id.into())
     }
 }
 
--- a/src/window.rs
+++ b/src/window.rs
@@ -93,7 +93,7 @@ impl WindowId {
         WindowId(platform_impl::WindowId::dummy())
     }
 
-    pub fn into_raw(self) -> u64 {
+    pub const fn into_raw(self) -> u64 {
         self.0.into_raw()
     }
 
--- a/src/platform_impl/apple/appkit/window.rs
+++ b/src/platform_impl/apple/appkit/window.rs
@@ -76,17 +76,13 @@ impl WindowId {
     pub const fn dummy() -> Self {
         Self(0)
     }
-}
 
-impl From<WindowId> for u64 {
-    fn from(window_id: WindowId) -> Self {
-        window_id.0 as u64
+    pub const fn into_raw(self) -> u64 {
+        self.0 as u64
     }
-}
 
-impl From<u64> for WindowId {
-    fn from(raw_id: u64) -> Self {
-        Self(raw_id as usize)
+    pub const fn from_raw(id: u64) -> Self {
+        Self(id as usize)
     }
 }
 
--- a/src/platform_impl/apple/uikit/window.rs
+++ b/src/platform_impl/apple/uikit/window.rs
@@ -100,7 +100,7 @@ impl WinitUIWindow {
     }
 
     pub(crate) fn id(&self) -> WindowId {
-        (self as *const Self as usize as u64).into()
+        WindowId { window: (self as *const Self as usize) as _ }
     }
 }
 
@@ -679,17 +679,13 @@ impl WindowId {
     pub const fn dummy() -> Self {
         WindowId { window: std::ptr::null_mut() }
     }
-}
 
-impl From<WindowId> for u64 {
-    fn from(window_id: WindowId) -> Self {
-        window_id.window as u64
+    pub fn into_raw(self) -> u64 {
+        self.window as u64
     }
-}
 
-impl From<u64> for WindowId {
-    fn from(raw_id: u64) -> Self {
-        Self { window: raw_id as _ }
+    pub const fn from_raw(id: u64) -> Self {
+        Self { window: id as *mut WinitUIWindow }
     }
 }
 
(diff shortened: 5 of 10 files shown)


notgull (dismissed):
I think having an easy converstion to/from `u64` is nice because it lets you easily store the `u64` in FFI.

kchibisov (changes requested):
Both are used in alarcitty's IPC.

daxpedda:
@kchibisov @notgull would you mind elaborating on those things?
I have no clue about why Alacritty needs this for IPC or what storing in FFI means exactly.

kchibisov:
We set a `WINDOW_ID` in the env variable and then user can use this `WINDOW_ID` of some window to target it from other window with IPC command (`alacritty msg config --window-id WINDOW_ID font.size=10`, the point here is that `WindowId` should be built and passed to `HashMap<WindowId, WindowContext>` when parsing it from the json payload on the socket.

If you remove the `from` we probably can use `u64` instead(not really nice, since you lose type safety), but we can not do anything without a `to` for example.

daxpedda:
> We set a `WINDOW_ID` in the env variable and then user can use this `WINDOW_ID` of some window to target it from other window with IPC command (`alacritty msg config --window-id WINDOW_ID font.size=10`, the point here is that `WindowId` should be built and passed to `HashMap<WindowId, WindowContext>` when parsing it from the json payload on the socket.
> 
> If you remove the `from` we probably can use `u64` instead(not really nice, since you lose type safety), but we can not do anything without a `to` for example.

Personally I would consider this still as a good tradeoff.
The `to_u64()` method makes absolute sense to me, but if we allow arbitrary construction from a `u64` we just throw type safety out the window (no pun intended). If `WindowId` is supposed to represent something, we should make sure it does.
These are just my 2¢, so its all very subjective.

But if this is too contentious, which I would totally understand, I would suggest to at least move away from the `From` implementations to dedicated methods. Same reasons as above: to make it clear to the user that this is an explicit conversion that is not natural; it doesn't produce necessarily valid `WindowId`s.

kchibisov:
Given that `WindowId` is not tied to the window's lifetime it's not really type safe in any form, so limiting won't provide much benefit here and given that we don't accept `WindowId` in any form I probably won't change that for now.

On the other side of things I was thinking to make `WindowId` opaque with a method to get some kind of `id()` and use the rest to encode some metadata, etc.

daxpedda:
> Given that `WindowId` is not tied to the window's lifetime it's not really type safe in any form, ...

Its type safe in the sense that you know a `WindowId` is actually a window ID. Otherwise we could play around with `u64`s directly as well.

nicoburns:
Constructing `WindowId` will be needed for 3rd party backend implementations.

daxpedda:
> Constructing `WindowId` will be needed for 3rd party backend implementations.

This is a separate matter that we need to address globally.
We need all types to be constructible for any backend to work, not just 3rd party, because all backends will be in their own crate. But we don't want direct `winit` users to see the internals or constructors of all these types.

nicoburns:
Another use case for this would be to store the ID as an opaque type somewhere that doesn't otherwise need to depend on Winit. With the functionality I can pass my library code a u64 and get it to pass it back to me when it generates events (e.g. enable IME). It never needs to know that it's a Winit ID, and if I want to integrate it with another system (not winit) then I can.

One pattern I've seen for this that I quite liked is from the slotmap crate which calls these methods `as_ffi` and `from_ffi` which really makes it clear that creating the newtype from u64 is only supported if the u64 was created from the newtype in the first place (https://docs.rs/slotmap/latest/slotmap/struct.KeyData.html#method.as_ffi) without actually removing any functionality.

daxpedda:
> One pattern I've seen for this that I quite liked is from the slotmap crate which calls these methods `as_ffi` and `from_ffi` which really makes it clear that creating the newtype from u64 is only supported if the u64 was created from the newtype in the first place (https://docs.rs/slotmap/latest/slotmap/struct.KeyData.html#method.as_ffi) without actually removing any functionality.

We discussed this in the meeting and are planning to go down exactly this route!

src/platform_impl/android/mod.rs
```diff
@@ -673,16 +673,12 @@ impl WindowId {
...
-impl From<WindowId> for u64 {
-    fn from(_: WindowId) -> Self {
+    pub fn into_raw(self) -> u64 {
         0
     }
-}
 
-impl From<u64> for WindowId {
-    fn from(_: u64) -> Self {
+    pub fn from_raw(_id: u64) -> Self {
```
kchibisov, line 681: Can we mark them as `const`?

src/changelog/unreleased.md
```diff
@@ -62,6 +62,8 @@ changelog entry.
 - Implement `Clone`, `Copy`, `Debug`, `Deserialize`, `Eq`, `Hash`, `Ord`, `PartialEq`, `PartialOrd`
   and `Serialize` on many types.
 - Add `MonitorHandle::current_video_mode()`.
+- On Web, implement `Error` for `platform::web::CustomCursorError`.
```
kchibisov, line 65: This looks like a rebase issue

src/changelog/unreleased.md
```diff
@@ -131,3 +133,5 @@ changelog entry.
 ### Fixed
 
 - On Orbital, `MonitorHandle::name()` now returns `None` instead of a dummy name.
+- Remove `impl From<u64> for WindowId` and `impl From<WindowId> for u64`. Replaced with
```
kchibisov, line 136: that's remove, not fixed.

kchibisov (changes requested):

src/changelog/unreleased.md
daxpedda: Also a rebasing issue!

src/platform_impl/android/mod.rs
daxpedda: `from_raw()` yes, but `into_raw()` not unfortunately, MacOS does pointer casting here.

src/window.rs
```diff
@@ -92,23 +92,19 @@ impl WindowId {
     pub const fn dummy() -> Self {
         WindowId(platform_impl::WindowId::dummy())
     }
-}
 
-impl fmt::Debug for WindowId {
-    fn fmt(&self, fmtr: &mut fmt::Formatter<'_>) -> fmt::Result {
-        self.0.fmt(fmtr)
+    pub fn into_raw(self) -> u64 {
```
madsmtm, line 96: Now that we're making this change, would it make sense to use `*const ()` instead of `u64`? That'd carry the provenance of the ID, which, to be fair, it isn't actually important for the comparisons that we're making, but might allow this to be `const` (I think?).

But maybe I'm thinking too much of removing `WindowId` once we have `WindowHandle`, because I think it would be a requirement that we use a pointer there.

madsmtm (commented):
> `from_ffi` / `to_ffi`

To my eyes, this looks a bit like these can be used in system-level FFI, i.e. as `HWND` on Windows and `*const NSWindow` on macOS (which is _not_ the intention). I think `from_raw` / `into_raw` are better.

src/platform_impl/android/mod.rs
madsmtm: Hmm, we could just make `WindowId` contain `u64` instead of a pointer or whatever it's doing now?

daxpedda: Feel free to push to this branch if you want to do it in this PR!

src/window.rs
daxpedda: It might indeed be worth a discussion removing `WindowId` entirely, but the use-cases mentioned above have to be covered somehow. (I still didn't understand the use-cases above, a matter for another time)

daxpedda: > Nit/thought: Now that we're making this change, would it make sense to use `*const ()` instead of `u64`?

I forgot to answer this particular part:

- Android: implements no `WindowId` to being with.
- iOS: is a pointer.
- MacOS: is a pointer but is converted to a `u64` before being stored in `WindowId`.
- Orbital: is actually a pointer but is converted to a `u64` before being stored in `WindowId`.
- Wayland: is a pointer but is converted to a `u64` before being stored in `WindowId`.
- Web: implements a counter as a `WindowId`, but could use a pointer to the canvas like in `RawWindowHandle`.
- X11: is an integer.

So as far as I can tell X11 is the only problem.
We could simply return a pointer as well and document that it is just an integer in reality.

I guess we can do that in a follow-up.

madsmtm: Yeah, I was kinda rambling here

madsmtm: I'm still not entirely sure where these are useful, maybe we could add that as documentation? But not a blocker for this PR, just a nicety

madsmtm (approved):

src/platform_impl/android/mod.rs
madsmtm: Done now

src/window.rs
daxpedda: From what I understood in the comments above they are used to have an identifier to be able to communicate with outside of Winit, e.g. through environment variables and other crates.
While this doesn't really convince me to have this API in place, I don't properly understand the use-cases either.

But I agree we should leave documentation here, but I will let others do this who actually understand the use-cases.

src/window.rs
```diff
@@ -92,23 +92,19 @@ impl WindowId {
...
 }
 
-impl From<u64> for WindowId {
-    fn from(raw_id: u64) -> Self {
-        Self(raw_id.into())
+impl fmt::Debug for WindowId {
+    fn fmt(&self, fmtr: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.0.fmt(fmtr)
     }
 }
```
kchibisov, line 110: Can't this be derived at this point?

kchibisov (approved):

src/window.rs
daxpedda: This is done on a bunch of times I assume to improve the `Debug` output, I believe it turns `WindowId(WindowId(0))` to `WindowId(0)`.

src/window.rs
nicoburns: @madsmtm One of the main use cases for `WindowId` in general is to:
- Call `window.id()` to get an `WindowId`
- Store it for later use
- And then later pass it back into Winit

This PR allows you to roundtrip a `WindowId` into a built-in type and back into a `WindowId`. Which is useful if the place you want to store the `WindowId` doesn't otherwise depend on Winit: it can then just store a u64 (which it treats as opaque), which gets converted back into a `WindowId` by an outer wrapper layer that does know about Winit.

Basically it allows for separation of concerns.

daxpedda: This is a concept I understand.
But I'm interested in an actual use-case that this facilitates and some reason why Winit has to help out there.

The concept could be argued for any type really, e.g. it could be said about `DeviceId`, `FingerId`, `CustomCursor` or `MonitorHandle`. I hope my point of view is understandable!

I apologize for any naivety on my part here, definitely wish I was more aware of the ecosystem Winit is used and consumed in.

src/window.rs
madsmtm: See https://github.com/rust-windowing/winit/pull/3622

src/window.rs
madsmtm: I've added a little bit of documentation to these types now, feel free to say so if you believe it should say something else.

madsmtm:
@notgull I'm going to go ahead and merge this even though you have "requested changes", because your concern has been resolved.

PR #3796, opened 2024-07-17
Implement common Std traits on many types

daxpedda:
Notable implementations:
- `DeviceId`: `Default`, uses `DeviceId::dumy()`.
- `ActiveEventLoop`, `EventLoop`, `EventLoopProxy`: refrained from implementing anymore then `Debug` because they should be unique.
- `EventLoopBuilder`: `Eq`, `Hash` and `PartialEq`.
- `InnerSizeWriter`: refrained from implementing `Hash`.
- `OwnedDisplayHandle`: `Eq` and `PartialEq`.
- `KeyLocation`, `ModifiersKeyState`: `Ord` and `PartialOrd`.
- `AnyThread`: `Clone`.
- `Window`: `Eq`, `Hash` and `PartialEq` via `WindowId`.

Replaced `Debug` implementations using `f.pad("Type { .. }")` with `f.debug_struct("Type").finish_non_exhaustive()`.

Let me know if anything is too contentious and I can remove it.

Follows [C-COMMON-TRAITS](https://rust-lang.github.io/api-guidelines/interoperability.html#types-eagerly-implement-common-traits-c-common-traits).

### DIFF
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -52,7 +52,7 @@ pub struct EventLoop {
 /// easier. But note that constructing multiple event loops is not supported.
 ///
 /// This can be created using [`EventLoop::builder`].
-#[derive(Default)]
+#[derive(Default, PartialEq, Eq, Hash)]
 pub struct EventLoopBuilder {
     pub(crate) platform_specific: platform_impl::PlatformSpecificEventLoopAttributes,
 }
@@ -115,9 +115,15 @@ impl EventLoopBuilder {
     }
 }
 
+impl fmt::Debug for EventLoopBuilder {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("EventLoopBuilder").finish_non_exhaustive()
+    }
+}
+
 impl fmt::Debug for EventLoop {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.pad("EventLoop { .. }")
+        f.debug_struct("EventLoop").finish_non_exhaustive()
     }
 }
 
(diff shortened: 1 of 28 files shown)


src/keyboard.rs
```diff
@@ -1755,51 +1758,3 @@ bitflags! {
...
-                ModifiersStateSerialize::deserialize(deserializer)?;
-            let mut m = ModifiersState::empty();
-            m.set(ModifiersState::SHIFT, shift_key);
-            m.set(ModifiersState::CONTROL, control_key);
-            m.set(ModifiersState::ALT, alt_key);
-            m.set(ModifiersState::SUPER, super_key);
-            Ok(m)
-        }
-    }
-}
```
kchibisov, line 1805: That's a breaking change.

kchibisov (changes requested):

src/keyboard.rs
kchibisov: We could probably just use the bitflags here, but we should document properly that it's a breaking change.

daxpedda: Done.

kchibisov (approved):

src/event_loop.rs
```diff
@@ -486,7 +492,7 @@ unsafe impl rwh_05::HasRawDisplayHandle for ActiveEventLoop {
 ///
 /// - A zero-sized type that is likely optimized out.
 /// - A reference-counted pointer to the underlying type.
-#[derive(Clone)]
+#[derive(Clone, PartialEq, Eq)]
```
notgull, line 495: What's the user story for this one?

notgull (approved):
Looks mostly fine to me

src/event_loop.rs
daxpedda: Ha!
I would honestly not be able to come up with one.

Happy to remove it if you don't like it!

PR #3801, opened 2024-07-19
Web: Implement `MonitorHandle`

daxpedda:
~~This is based on and requires #3805.~~

This implements `MonitorHandle` on Web by using the [Window Management API](https://developer.mozilla.org/en-US/docs/Web/API/Window_Management_API), which is currently only implemented on Chrome.

This gets a bit complicated, because to access the API the user has to explicitly be asked to give permission, on top of most calls being asynchronous. So to solve this I implemented a couple of things:
- When the `EventLoop` is created, the browser is queried, in the background if the user has previously given permission to use the Window Management API.
- When the event loop is started, `EventLoop::run()`, it is delayed until this asynchronous query has finished. Which should really only be one tick.
- If the user has previously given permission to use the Window Management API, we can use it.
- If not, we fallback to [`Window.screen`](https://developer.mozilla.org/en-US/docs/Web/API/Window/screen), which is implemented on all browsers. The "downside" is that it contains much less information about the current monitor and no information about primary or other monitors.

Cool things:
- At none of these steps the user is actually asked to give permission. To explicitly ask the user for permission, `ActiveEventLoop::request_detailed_monitor_permission()` has to be used.
- None of the `Future`s have to polled to completion, dropping them is fine. They are only there to receive errors or allow the user to "know when something has finished doing its thing".
- `MonitorHandle`s created with the Window Management API can be targeted with `Window::set_fullscreen()`.
- If the user somehow grants the site permission to use the Window Management API externally without going through Winit, we detect this and automatically upgrade the backend.
- Everything that can be cached is cached to minimize any JS FFI calls. E.g. multiple calls to `ActiveEventLoop::request_detailed_monitor_permission()` do nothing, instead they share the result to minimize any JS overhead.

Not cool things:
- There is no way to "upgrade" `MonitorHandle`s to use the new API if they were created using the fallback.
- `VideoModeHandle` is basically useless, its only there to expose `bit_depth()`.
- Using `MonitorHandle` with `Window::set_fullscreen()` only makes sense if it was created using the Window Management API, otherwise it doesn't work.
- Currently the only way for users to receive errors is to pull the returned `Future`s to completion.
- The whole screen orientation API has no cross-platform implementation, so its all implemented via extension traits.

---

Follow-up:
- Add stuff to #2875.
- Return `Option` from `MonitorHandle::scale_factor()`.
- Return `Option` from `MonitorHandle::position()`.
- Rename `Window::set_fullscreen()` to `Window::request_fullscreen()` and maybe return an error. If not introduce an Web-only async version of it to allow users to actually handle failure.
- Add current `bit_depth()` to `MonitorHandle` and remove `VideoModeHandle` from Web again. (#3802)
- Open tracking issue for cross-platform monitor events.
- Open tracking issue for cross-platform screen orientation getters, locking and events.


daxpedda:
I extracted some unrelated changes to #3805 and based this PR on it.

examples/window.rs
```diff
@@ -160,7 +172,23 @@ impl Application {
         Ok(window_id)
     }
 
-    fn handle_action(&mut self, event_loop: &ActiveEventLoop, window_id: WindowId, action: Action) {
+    fn handle_action(&mut self, event_loop: &ActiveEventLoop, action: Action) {
+        match action {
+            Action::PrintHelp => self.print_help(),
```
kchibisov, line 177: I'm pretty sure it won't work here, since it's triggered from the user input.

I'd generally create a separate channel to handle dumping monitors/wakeup on web, and make it web specific if you really need those channels.

examples/window.rs
```diff
@@ -1069,6 +1128,7 @@ const KEY_BINDINGS: &[Binding<&'static str>] = &[
     Binding::new("T", ModifiersState::SUPER, Action::CreateNewTab),
     #[cfg(macos_platform)]
     Binding::new("O", ModifiersState::CONTROL, Action::CycleOptionAsAlt),
+    Binding::new("S", ModifiersState::empty(), Action::Message),
```
kchibisov, line 1131: You don't need to bind it since it does nothing.

examples/window.rs
```diff
@@ -1041,6 +1099,7 @@ const KEY_BINDINGS: &[Binding<&'static str>] = &[
     Binding::new("R", ModifiersState::CONTROL, Action::ToggleResizable),
     Binding::new("R", ModifiersState::ALT, Action::RequestResize),
     // M.
+    Binding::new("M", ModifiersState::empty(), Action::DumpMonitors),
```
kchibisov, line 1102: We generally don't bind with `::empty` because you won't be able to do regular typing. Though, it seems like a few were missed and we have `::empty` with `SetTheme` probably should change that as well to use `ALT`.

src/monitor.rs
```diff
@@ -121,6 +134,15 @@ impl MonitorHandle {
 
     /// Returns the top-left corner position of the monitor relative to the larger full
     /// screen area.
+    ///
+    /// ## Platform-specific
+    ///
+    /// **Web:** Always returns `{ x: 0, y: 0 }` without
```
kchibisov, line 140: Maybe we should say that we return `::default`?

src/platform/web.rs
```diff
@@ -206,6 +213,21 @@ pub trait EventLoopExtWeb {
     ///
     /// [`ControlFlow::WaitUntil`]: crate::event_loop::ControlFlow::WaitUntil
     fn wait_until_strategy(&self) -> WaitUntilStrategy;
+
+    /// Returns if the users device has multiple screens.
+    ///
+    /// Browsers might always return [`false`] to reduce fingerprinting.
+    fn has_multiple_screens(&self) -> Result<bool, NotSupportedError>;
```
kchibisov, line 232: Do we really need this call if we have a call to list outputs? Though, maybe web is special.

src/platform/web.rs
```diff
@@ -436,3 +503,192 @@ impl Display for CustomCursorError {
...
+    }
+}
+
+/// Screen orientation data.
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+pub struct OrientationData {
+    /// The orientation.
+    pub orientation: Orientation,
+    /// [`true`] if the [`orientation`](Self::orientation) is flipped upside down.
+    pub flipped: bool,
```
kchibisov, line 645: I'd just merge the `flipped` into `Orientation`

src/platform/web.rs
```diff
@@ -436,3 +503,192 @@ impl Display for CustomCursorError {
...
+        /// - [`false`]: User is locked to upside down landscape mode.
+        flipped: Option<bool>,
+    },
+    /// User is locked to portrait mode.
+    Portrait {
+        /// - [`None`]: User is locked to both upright or upside down portrait mode.
+        /// - [`false`]: User is locked to upright portrait mode.
+        /// - [`false`]: User is locked to upside down portrait mode.
+        flipped: Option<bool>,
+    },
```
kchibisov, line 683: Would use `Orientation` with all the flipped variants added as extension to this.

src/platform_impl/web/async/wrapper.rs
```diff
@@ -81,3 +83,29 @@ impl<V, S: Clone + Send, E> Clone for Wrapper<V, S, E> {
         }
     }
 }
+
+impl<V, S: Clone + Send, E> Eq for Wrapper<V, S, E> {}
```
kchibisov, line 87: Generally `Eq`/`Ord` go after their `Partial` impls.

src/platform_impl/web/async/dispatcher.rs
```diff
@@ -9,6 +12,44 @@ pub struct Dispatcher<T: 'static>(Wrapper<T, Arc<Sender<Closure<T>>>, Closure<T>
...
+    }
+}
+
+impl<T> Debug for Dispatcher<T> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Dispatcher").finish_non_exhaustive()
+    }
+}
+
+impl<T> Eq for Dispatcher<T> {}
```
kchibisov, line 27: Can't you just derive all of that?

src/platform_impl/web/monitor.rs
```diff
@@ -1,53 +1,798 @@
...
+                                        });
+                                    }
+                                }
+                            }),
+                        );
+                        State::Permission { permission, _handle: handle }
+                    };
+
+                    *runner.monitor().state.borrow_mut() = state;
+                    runner.start_delayed();
```
kchibisov, line 439: All of that is just a massive indention level and the function gets really long, should probably split, use early returns, etc to maintain the readability.

src/platform_impl/web/monitor.rs
```diff
@@ -1,53 +1,798 @@
...
+                                });
+
+                                Poll::Pending
+                            })
+                        },
+                    }
+                } else {
+                    Poll::Ready(Ok(()))
+                }
+            },
```
kchibisov, line 673: another massive indention function.

src/platform_impl/web/web_sys/fullscreen.rs
```diff
@@ -19,21 +26,65 @@ pub fn request_fullscreen(document: &Document, canvas: &HtmlCanvasElement) {
...
+                    error!(
+                        "Selecting a specific screen for fullscreen mode requires a detailed \
+                         screen. See `MonitorHandleExtWeb::is_detailed()`."
+                    )
+                }
+            } else {
+                error!(
+                    "Fullscreen mode selecting a specific screen is not supported by this browser"
+                )
+            }
```
kchibisov, line 77: Would probably early return the outer else's error.

examples/window.rs
```diff
@@ -43,26 +46,34 @@ fn main() -> Result<(), Box<dyn Error>> {
...
-        // Wake up the `event_loop` once every second and dispatch a custom event
-        // from a different thread.
-        info!("Starting to send user event every second");
-        loop {
-            _event_loop_proxy.wake_up();
-            std::thread::sleep(std::time::Duration::from_secs(1));
+    std::thread::spawn({
+        let event_loop_proxy = event_loop.create_proxy();
+        let sender = sender.clone();
+        move || {
```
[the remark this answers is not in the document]
daxpedda: I just use it to prevent further nesting.
But I agree its unusual.

Will change it.

examples/window.rs
```diff
@@ -217,6 +244,29 @@ impl Application {
...
+                    let proxy = event_loop.create_proxy();
+                    let sender = self.sender.clone();
+                    wasm_bindgen_futures::spawn_local(async move {
+                        if let Err(error) = future.await {
+                            error!("{error}")
+                        }
+
+                        let _ = sender.send(Action::DumpMonitors);
+                        proxy.wake_up();
+                    });
```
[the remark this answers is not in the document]
daxpedda: > What's the reason behind all of that and why web can not do what everyone else is doing here?

No other backend needs permission, so this is really impossible to do cross-platform the same cross-platform.
> I understand the permission part, but I don't get why we're spawning a future?

You need to know when the permission has been given. How else do you propose doing this?

examples/window.rs
daxpedda: > I'm pretty sure it won't work here, since it's triggered from the user input.

How is this related to user input?
I can confirm it does actually work.

> I'd generally create a separate channel to handle dumping monitors/wakeup on web, and make it web specific if you really need those channels.

This already has a use-case on native:
I added triggering a proxy wake up with "S".

Seeing your other comments I think you should just try this yourself to see how it works and what it does.

examples/window.rs
daxpedda: It does, see https://github.com/rust-windowing/winit/pull/3801#discussion_r1688280503.

examples/window.rs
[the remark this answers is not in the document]
kchibisov: Generally I'd have an extension just for web based on trait that calls back to you once you got the permission. You can even do that for all permissions you want for whatever reason, not just monitor.

In the example, I'd probably just ask unconditionally during startup, though it could annoy or ask and just dumb anyway, since it'll be much simpler...

src/platform/web.rs
daxpedda: Yes, this is useful to determine if users should even ask for permission.
Asking for permission is an involved process that requires user input through the browser, so its best avoided if not necessary.

examples/window.rs
kchibisov: >This already has a use-case on native:
I added triggering a proxy wake up with "S".

In this case, could you split the actions properly?

src/platform/web.rs
daxpedda: This would require four variants, e.g.
```
enum Orientation {
    Landscape,
    LandscapeFlipped,
    Portrait,
    PortraitFlipped,
}
```
Adding natural would add two more, so six, all with very similar names.
I think this makes it harder to extract information and be more concise about each field and name.

So I consider the current implementation better.

src/platform/web.rs
daxpedda: This wouldn't work, because it wouldn't cover `Portrait { flipped: None }` for example.

src/platform_impl/web/async/wrapper.rs
daxpedda: I generally do this alphabetically in the Web backend.
So I would like to stick to that.

I can make a follow-up PR where I align all `derive`s and implementations in the whole crate to do `PartialEq/PartialOrd` before `Eq/Ord` and generally have the same order throughout the crate if you like.

src/platform_impl/web/async/dispatcher.rs
daxpedda: No, because then it depends on `T`.

examples/window.rs
daxpedda: Done, also changed the message binding.

examples/window.rs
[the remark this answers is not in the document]
daxpedda: > Generally I'd have an extension just for web based on trait that calls back to you once you got the permission. You can even do that for all permissions you want for whatever reason, not just monitor.

Uhm, you mean in the future, because right now that's not possible, right?

> In the example, I'd probably just ask unconditionally during startup, though it could annoy or ask and just dump anyway, since it'll be much simpler...

Asking for permission at startup isn't really an option, that would be an awful example.
But I can dump anyway, I think that's a good idea!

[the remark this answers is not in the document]
daxpedda: Hm, on second that I find this actually useful as an example for how to let users deal with the async APIs on Web.
So I think removing this is a mistake, especially for the example.

I will split it up though, to make it easier to read.

src/platform/web.rs
kchibisov: I was thinking to probably make the type more general in this stage, given that on Wayland I have the following `https://wayland.app/protocols/wayland#wl_output:enum:transform`

So I think `flipped` could also have different meaning...

I'd assume on Web it's just rotated and not really flipped?

examples/window.rs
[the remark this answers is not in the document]
kchibisov: >Uhm, you mean in the future, because right now that's not possible, right?

yeah...

examples/window.rs
daxpedda: Done, I also remove the `_ => ...` shortcut and renamed the function, so the flow should be easier to understand now.

src/monitor.rs
daxpedda: Done.

src/platform/web.rs
daxpedda: No rotating would be between "rotating between landscape and potrait".
Flipped means vertical axis indeed, but that can apply to both landscape and portrait.

I don't know how to interpret the Wayland data, because it doesn't actually contain `Orientation`, it only contains if its flipped and in which degree, but not if its landscape or portrait.
I guess this can be inferred by comparing the width and height?

src/platform_impl/web/event_loop/runner.rs
```diff
@@ -199,17 +219,42 @@ impl Shared {
...
+            } else {
+                *runner = RunnerEnum::Running(Runner::new(event_handler));
+                true
+            }
+        };
+
+        if start {
+            self.init();
+            self.set_listener();
         }
```
kchibisov, line 238: Just looking into it again, can't we just inline the `if start` into the branch with `true`? Or do you implicitly rely on `runner` borrow being dropped, if so I'd probably document that?

kchibisov (changes requested):
+ already present comments about nesting, since the code is really hard to follow...

src/platform_impl/web/async/wrapper.rs
kchibisov: I'd just suggest to do that for the new stuff since it's just pointless, unless we have a formatter support.

I won't block on the current order either, just noticed that it's usually other way around (you can also `derive(Eq)` if you have `impl PartialEq`).

src/platform_impl/web/event_loop/runner.rs
daxpedda: Yes, `runner` has to be dropped, that's the reason for this whole block.

src/platform_impl/web/async/wrapper.rs
daxpedda: No unfortunately no formatter support ...
I will make a PR only for Web then, because Web follow alphabetical order so far.

src/platform_impl/web/event_loop/runner.rs
daxpedda: I removed the whole block by just calling `drop(runner)`.

src/platform_impl/web/web_sys/fullscreen.rs
daxpedda: Done.

src/platform_impl/web/monitor.rs
daxpedda: Refactored.

src/platform/web.rs
daxpedda: We figured out that Wayland transform is probably not the same as screen orientation.

kchibisov (approved):

PR #3802, opened 2024-07-19
Add `MonitorHandle::current_video_mode()`

daxpedda:
After some discussion we decided its more useful to encode current monitor information into `VideoModeHandle` then move more information to `MonitorHandle`. To this end this PR implements `MonitorHandle::current_video_mode()`.

Additionally `MonitorHandle::size()` and `refresh_rate_millihertz()` were removed, because it logically fits into video modes and not the monitor itself.

To improve the output, this PR also changes `VideoModeHandle::refresh_rate_millihertz()` to return a `Option<NonZeroU32>` and `size()` and `bit_depth()` an `Option`.

Follow up to #3801.

### DIFF
--- a/src/changelog/unreleased.md
+++ b/src/changelog/unreleased.md
@@ -122,6 +122,7 @@ changelog entry.
 - Remove `Touch::id` in favor of `Touch::finger_id`.
 - Remove `MonitorHandle::size()` and `refresh_rate_millihertz()` in favor of
   `MonitorHandle::current_video_mode()`.
+- On Android, remove all `MonitorHandle` support instead of emitting false data.
 
 ### Fixed
 
--- a/src/changelog/unreleased.md
+++ b/src/changelog/unreleased.md
@@ -95,6 +95,7 @@ changelog entry.
   accelerated, if the browser supports it.
 - `(Active)EventLoop::create_custom_cursor()` now returns a `Result<CustomCursor, ExternalError>`.
 - Changed how `ModifiersState` is serialized by Serde.
+- `VideoModeHandle::refresh_rate_millihertz()` now returns an `Option`.
 
 ### Removed
 
@@ -118,6 +119,8 @@ changelog entry.
 - Remove `DeviceEvent::Added` and `DeviceEvent::Removed`.
 - Remove `DeviceEvent::Motion` and `WindowEvent::AxisMotion`.
 - Remove `Touch::id` in favor of `Touch::finger_id`.
+- Remove `MonitorHandle::size()` and `refresh_rate_millihertz()` in favor of
+  `MonitorHandle::current_video_mode()`.
 
 ### Fixed
 
--- a/src/changelog/unreleased.md
+++ b/src/changelog/unreleased.md
@@ -95,7 +95,8 @@ changelog entry.
   accelerated, if the browser supports it.
 - `(Active)EventLoop::create_custom_cursor()` now returns a `Result<CustomCursor, ExternalError>`.
 - Changed how `ModifiersState` is serialized by Serde.
-- `VideoModeHandle::refresh_rate_millihertz()` now returns an `Option`.
+- `VideoModeHandle::refresh_rate_millihertz()` and `bit_depth()` now return a `Option<NonZero*>`.
+- `MonitorHandle::position()` now returns an `Option`.
 
 ### Removed
 
@@ -127,3 +128,4 @@ changelog entry.
 - On Web, pen events are now routed through to `WindowEvent::Cursor*`.
 - On macOS, fix panic when releasing not available monitor.
 - On MacOS, return the system theme in `Window::theme()` if no theme override is set.
+- On Orbital, `MonitorHandle::name()` now returns `None` instead of a dummy name.
--- a/src/changelog/unreleased.md
+++ b/src/changelog/unreleased.md
@@ -61,6 +61,7 @@ changelog entry.
   the primary finger in a multi-touch interaction.
 - Implement `Clone`, `Copy`, `Debug`, `Deserialize`, `Eq`, `Hash`, `Ord`, `PartialEq`, `PartialOrd`
   and `Serialize` on many types.
+- Add `MonitorHandle::current_video_mode()`.
 
 ### Changed
 
--- a/src/monitor.rs
+++ b/src/monitor.rs
@@ -64,12 +64,8 @@ impl VideoModeHandle {
     }
 
     /// Returns the refresh rate of this video mode in mHz.
-    ///
-    /// ## Platform-specific
-    ///
-    /// **Web:** Always returns `0`.
     #[inline]
-    pub fn refresh_rate_millihertz(&self) -> u32 {
+    pub fn refresh_rate_millihertz(&self) -> Option<u32> {
         self.video_mode.refresh_rate_millihertz()
     }
 
@@ -85,10 +81,10 @@ impl std::fmt::Display for VideoModeHandle {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         write!(
             f,
-            "{}x{} @ {} mHz ({} bpp)",
+            "{}x{} {}({} bpp)",
             self.size().width,
             self.size().height,
-            self.refresh_rate_millihertz(),
+            self.refresh_rate_millihertz().map(|rate| format!("@ {rate} mHz ")).unwrap_or_default(),
             self.bit_depth()
         )
     }
@@ -139,12 +135,6 @@ impl MonitorHandle {
         self.inner.name()
     }
 
-    /// Returns the monitor's resolution.
-    #[inline]
-    pub fn size(&self) -> PhysicalSize<u32> {
-        self.inner.size()
-    }
-
     /// Returns the top-left corner position of the monitor relative to the larger full
     /// screen area.
     ///
@@ -159,22 +149,6 @@ impl MonitorHandle {
     #[inline]
     pub fn position(&self) -> PhysicalPosition<i32> {
         self.inner.position()
-    }
-
-    /// The monitor refresh rate used by the system.
-    ///
-    /// Return `Some` if succeed, or `None` if failed, which usually happens when the monitor
-    /// the window is on is removed.
-    ///
-    /// When using exclusive fullscreen, the refresh rate of the [`VideoModeHandle`] that was
-    /// used to enter fullscreen should be used instead.
-    ///
-    /// ## Platform-specific
-    ///
-    /// **Web:** Always returns [`None`].
-    #[inline]
-    pub fn refresh_rate_millihertz(&self) -> Option<u32> {
-        self.inner.refresh_rate_millihertz()
     }
 
     /// Returns the scale factor of the underlying monitor. To map logical pixels to physical
--- a/src/monitor.rs
+++ b/src/monitor.rs
@@ -5,6 +5,8 @@
 //! methods, which return an iterator of [`MonitorHandle`]:
 //! - [`ActiveEventLoop::available_monitors`][crate::event_loop::ActiveEventLoop::available_monitors].
 //! - [`Window::available_monitors`][crate::window::Window::available_monitors].
+use std::num::{NonZeroU16, NonZeroU32};
+
 use crate::dpi::{PhysicalPosition, PhysicalSize};
 use crate::platform_impl;
 
@@ -44,7 +46,10 @@ impl Ord for VideoModeHandle {
 }
 
 impl VideoModeHandle {
-    /// Returns the resolution of this video mode.
+    /// Returns the resolution of this video mode. This **must not** be used to create your
+    /// rendering surface. Use [`Window::inner_size()`] instead.
+    ///
+    /// [`Window::inner_size()`]: crate::window::Window::inner_size
     #[inline]
     pub fn size(&self) -> PhysicalSize<u32> {
         self.video_mode.size()
@@ -53,19 +58,14 @@ impl VideoModeHandle {
     /// Returns the bit depth of this video mode, as in how many bits you have
     /// available per color. This is generally 24 bits or 32 bits on modern
     /// systems, depending on whether the alpha channel is counted or not.
-    ///
-    /// ## Platform-specific
-    ///
-    /// - **Wayland / Orbital:** Always returns 32.
-    /// - **iOS:** Always returns 32.
     #[inline]
-    pub fn bit_depth(&self) -> u16 {
+    pub fn bit_depth(&self) -> Option<NonZeroU16> {
         self.video_mode.bit_depth()
     }
 
     /// Returns the refresh rate of this video mode in mHz.
     #[inline]
-    pub fn refresh_rate_millihertz(&self) -> Option<u32> {
+    pub fn refresh_rate_millihertz(&self) -> Option<NonZeroU32> {
         self.video_mode.refresh_rate_millihertz()
     }
 
@@ -81,11 +81,11 @@ impl std::fmt::Display for VideoModeHandle {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         write!(
             f,
-            "{}x{} {}({} bpp)",
+            "{}x{} {}{}",
             self.size().width,
             self.size().height,
             self.refresh_rate_millihertz().map(|rate| format!("@ {rate} mHz ")).unwrap_or_default(),
-            self.bit_depth()
+            self.bit_depth().map(|bit_depth| format!("({bit_depth} bpp)")).unwrap_or_default(),
         )
     }
 }
@@ -112,9 +112,15 @@ impl std::fmt::Display for VideoModeHandle {
 /// to check.
 ///
 /// [`Window`]: crate::window::Window
-#[derive(Debug, Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
+#[derive(Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
 pub struct MonitorHandle {
     pub(crate) inner: platform_impl::MonitorHandle,
+}
+
+impl std::fmt::Debug for MonitorHandle {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.inner.fmt(f)
+    }
 }
 
 impl MonitorHandle {
@@ -140,14 +146,14 @@ impl MonitorHandle {
     ///
     /// ## Platform-specific
     ///
-    /// **Web:** Always returns [`Default`] without
+    /// **Web:** Always returns [`None`] without
     #[cfg_attr(
         any(web_platform, docsrs),
         doc = "[detailed monitor permissions][crate::platform::web::ActiveEventLoopExtWeb::request_detailed_monitor_permission]."
     )]
     #[cfg_attr(not(any(web_platform, docsrs)), doc = "detailed monitor permissions.")]
     #[inline]
-    pub fn position(&self) -> PhysicalPosition<i32> {
+    pub fn position(&self) -> Option<PhysicalPosition<i32>> {
         self.inner.position()
     }
 
--- a/src/monitor.rs
+++ b/src/monitor.rs
@@ -201,6 +201,12 @@ impl MonitorHandle {
         self.inner.scale_factor()
     }
 
+    /// Returns the currently active video mode of this monitor.
+    #[inline]
+    pub fn current_video_mode(&self) -> Option<VideoModeHandle> {
+        self.inner.current_video_mode().map(|video_mode| VideoModeHandle { video_mode })
+    }
+
     /// Returns all fullscreen video modes supported by this monitor.
     #[inline]
     pub fn video_modes(&self) -> impl Iterator<Item = VideoModeHandle> {
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -2,6 +2,7 @@ use std::any::Any;
 use std::cell::Cell;
 use std::collections::VecDeque;
 use std::hash::Hash;
+use std::num::{NonZeroU16, NonZeroU32};
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Mutex};
 use std::time::{Duration, Instant};
@@ -980,8 +981,8 @@ impl MonitorHandle {
         Some("Android Device".to_owned())
     }
 
-    pub fn position(&self) -> PhysicalPosition<i32> {
-        (0, 0).into()
+    pub fn position(&self) -> Option<PhysicalPosition<i32>> {
+        None
     }
 
     pub fn scale_factor(&self) -> f64 {
@@ -989,12 +990,7 @@ impl MonitorHandle {
     }
 
     pub fn current_video_mode(&self) -> Option<VideoModeHandle> {
-        Some(VideoModeHandle {
-            size: screen_size(&self.app),
-            // FIXME: it is guaranteed to support 32 bit color though
-            bit_depth: 32,
-            monitor: self.clone(),
-        })
+        Some(VideoModeHandle { size: screen_size(&self.app), monitor: self.clone() })
     }
 
     pub fn video_modes(&self) -> impl Iterator<Item = VideoModeHandle> {
@@ -1005,7 +1001,6 @@ impl MonitorHandle {
 #[derive(Clone, Debug, Eq, Hash, PartialEq)]
 pub struct VideoModeHandle {
     size: PhysicalSize<u32>,
-    bit_depth: u16,
     monitor: MonitorHandle,
 }
 
@@ -1014,11 +1009,11 @@ impl VideoModeHandle {
         self.size
     }
 
-    pub fn bit_depth(&self) -> u16 {
-        self.bit_depth
+    pub fn bit_depth(&self) -> Option<NonZeroU16> {
+        None
     }
 
-    pub fn refresh_rate_millihertz(&self) -> Option<u32> {
+    pub fn refresh_rate_millihertz(&self) -> Option<NonZeroU32> {
         None
     }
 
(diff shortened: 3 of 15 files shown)


src/platform_impl/apple/appkit/monitor.rs
```diff
@@ -237,6 +237,21 @@ impl MonitorHandle {
...
+    pub fn bit_depth(&self) -> u16 {
+        run_on_main(|mtm| {
+            match self.ns_screen(mtm) {
+                Some(screen) => unsafe {
+                    (NSBitsPerPixelFromDepth(screen.depth())
+                        * screen
+                            .colorSpace()
+                            .map(|color_space| color_space.numberOfColorComponents())
+                            .unwrap_or(4)) as u16
+                },
```
daxpedda, line 249: Please help @madsmtm :cold_sweat:.
Not sure why this is unsafe or if there is a better way to do this.
I guess I could also just assume that the number of color components is 4?

src/platform_impl/android/mod.rs
```diff
@@ -1012,6 +1012,10 @@ impl MonitorHandle {
         self.app.config().density().map(|dpi| dpi as f64 / 160.0).unwrap_or(1.0)
     }
 
+    pub fn bit_depth(&self) -> u16 {
+        32
```
MarijnS95, line 1016: On Android this seems hard. We could ask for the current format but various APIs can always change it.

I've seen though that after changing the Surface/window format, APIs like EGL suddenly return just the configs matching that exact format.

daxpedda: > We could ask for the current format but various APIs can always change it.

I think it should be clear to users that the data only reflects the state at the time it was queried.

In the future it would be nice if we could have some events that reflect a change in available monitors or their settings. Its definitely on my checklist for Web.

> On Android this seems hard.

I'm not too worried about it in this PR because `VideoModeHandle` on Android already always returns 32, so this definitely doesn't change anything about the current state of things.

kchibisov:
The `bit_depth` is more of a property of the active video format, so if I'd change the way handle works I'd probably just keep a handle and an option to get the active `VideoMode` for stuff like that.

The only information which is not video mode related is basically monitor name/position, etc, depth is not it.

daxpedda:
> The `bit_depth` is more of a property of the active video format, so if I'd change the way handle works I'd probably just keep a handle and an option to get the active `VideoMode` for stuff like that.
> 
> The only information which is not video mode related is basically monitor name/position, etc, depth is not it.

That's a good point. That would solve the issue for Web as well.
Will look into it and adapt the PR.

daxpedda:
> The `bit_depth` is more of a property of the active video format, so if I'd change the way handle works I'd probably just keep a handle and an option to get the active `VideoMode` for stuff like that.
> 
> The only information which is not video mode related is basically monitor name/position, etc, depth is not it.

@kchibisov I started working on it right now and realized that this argument invalidates `MonitorHandle` currently having size and refresh rate. Is there some story behind it or should we remove it in favor of `current_video_mode()` exposing the current size and refresh rate?

kchibisov:
>Is there some story behind it or should we remove it in favor of current_video_mode() exposing the current size and refresh rate?

I think it was always like that, which doesn't make much sense. Probably it was like that for convenience or something. Just route everything into `current_video_mode`.

src/platform_impl/apple/appkit/monitor.rs
daxpedda: Resolved by the PR completely changing.

daxpedda:
> > Is there some story behind it or should we remove it in favor of current_video_mode() exposing the current size and refresh rate?
> 
> I think it was always like that, which doesn't make much sense. Probably it was like that for convenience or something. Just route everything into `current_video_mode`.

Done.

src/changelog/unreleased.md
```diff
@@ -70,6 +71,7 @@ changelog entry.
 - Change signature of `EventLoop::run_app`, `EventLoopExtPumpEvents::pump_app_events` and
   `EventLoopExtRunOnDemand::run_app_on_demand` to accept a `impl ApplicationHandler` directly,
   instead of requiring a `&mut` reference to it.
+- `VideoModeHandle::refresh_rate_millihertz()` now returns an `Option`.
```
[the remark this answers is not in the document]
daxpedda: I like it!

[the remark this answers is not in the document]
daxpedda: Done.

kchibisov (approved):

src/platform_impl/android/mod.rs
```diff
@@ -1012,46 +1003,49 @@ impl MonitorHandle {
...
     }
 
     pub fn bit_depth(&self) -> u16 {
         self.bit_depth
     }
 
-    pub fn refresh_rate_millihertz(&self) -> u32 {
-        self.refresh_rate_millihertz
+    pub fn refresh_rate_millihertz(&self) -> Option<NonZeroU32> {
+        None
```
MarijnS95, line 1037: Would you mind adding a TODO that we should get this via either https://developer.android.com/reference/android/view/Display#getRefreshRate() (requires JNI, but works all the way back to API level 1) or via https://developer.android.com/ndk/reference/group/choreographer#achoreographer_registerrefreshratecallback (only available since API level 30, related to https://developer.android.com/media/optimize/performance/frame-rate).

src/platform_impl/android/mod.rs
```diff
@@ -1012,46 +1003,49 @@ impl MonitorHandle {
...
-        let size = self.size().into();
-        // FIXME this is not the real refresh rate
-        // (it is guaranteed to support 32 bit color though)
-        std::iter::once(VideoModeHandle {
-            size,
+    pub fn current_video_mode(&self) -> Option<VideoModeHandle> {
+        Some(VideoModeHandle {
+            size: screen_size(&self.app),
+            // FIXME: it is guaranteed to support 32 bit color though
             bit_depth: 32,
```
MarijnS95, line 1010: They seem to have [deprecated the ability to get the pixel format](https://developer.android.com/reference/android/view/Display#getPixelFormat()) long ago, unfortunately.

There's no point querying the format of the `Surface` since:
1. There won't be a `Surface` until the `Resumed` event;
2. The format typically starts out with `RGB565` for some arcane reason;
3. The user can freely change this pixel format and the compositor will Deal With It™.

Maybe this needs to be `Option` as well?

src/platform_impl/android/mod.rs
```diff
@@ -1012,46 +1003,49 @@ impl MonitorHandle {
...
-            refresh_rate_millihertz: 60000,
             monitor: self.clone(),
         })
     }
+
+    pub fn video_modes(&self) -> impl Iterator<Item = VideoModeHandle> {
+        self.current_video_mode().into_iter()
```
MarijnS95, line 1016: Leaving this here just in case: https://developer.android.com/reference/android/view/Display.Mode

src/platform_impl/android/mod.rs
MarijnS95: Also curious what we should do in the case of VRR.

src/platform_impl/android/mod.rs
```diff
@@ -1012,46 +1003,49 @@ impl MonitorHandle {
...
+        None
     }
 
     pub fn monitor(&self) -> MonitorHandle {
         self.monitor.clone()
     }
 }
+
+fn screen_size(app: &AndroidApp) -> PhysicalSize<u32> {
+    if let Some(native_window) = app.native_window() {
```
MarijnS95, line 1046: Perhaps we could pull this from the `configuration` instead, but it's all DPI-dependent:

https://docs.rs/ndk/latest/ndk/configuration/struct.Configuration.html#method.screen_width_dp

MarijnS95 (approved):
Approving the Android side given that all the weirdness in there (due to a lack of API in the NDK - we could call into Java for this though) was already there and is only moving in this patch but not fundamentally changing.

src/platform_impl/android/mod.rs
daxpedda: > 1. There won't be a `Surface` until the `Resumed` event;

Why would that make it pointless? Do you mean because when you have a surface you use your graphics API to determine your pixel format anyway?

In that case the API would be useless on all platforms wouldn't it?
Personally I don't know who would use that, so I would actually be in favor of removing it wholesale if we come to a consensus.

> Maybe this needs to be `Option` as well?

I think we should change it to `None` if its not supported.

My assumption was always that the main use-case here is to check for HDR support, which Android would support with [`android.view.Display.getSupportedHdrTypes()`](https://developer.android.com/reference/android/view/Display.Mode#getSupportedHdrTypes()). But that API should probably look different and again, not useful to handle in Winit. E.g., we discussed querying pixel format support in Softbuffer already https://github.com/rust-windowing/softbuffer/issues/98 and Wgpu obviously already supports that.

---

If you don't mind I would leave it like this here and make a follow up PR discussing any further changes.
WDYT in any case about all this?

src/platform_impl/android/mod.rs
daxpedda: Let me know if you want me to add any comment for that.

src/platform_impl/android/mod.rs
daxpedda: Added it as a TODO comment.

src/platform_impl/android/mod.rs
daxpedda: Done!

> Also curious what we should do in the case of VRR.

Ha!
We would need to change the API for that, lol.

src/platform_impl/android/mod.rs
MarijnS95: > > 1. There won't be a `Surface` until the `Resumed` event;
> 
> Why would that make it pointless? Do you mean because when you have a surface you use your graphics API to determine your pixel format anyway?

Pointless _on Android_, since I think you can query the video mode (and most apps probably will?) before a "surface" appears?

The pixel format of a surface can and will almost always be disjoint from the bit depth of the video mode on a physical panel/connection because of compositors (though it's probably beneficial to keep them in sync depending on what the scanout hardware is capable of).

> In that case the API would be useless on all platforms wouldn't it? Personally I don't know who would use that, so I would actually be in favor of removing it wholesale if we come to a consensus.

I think all platforms with a list of monitors and their current mode can perfectly benefit from this API. A pixel format may be more helpful over a plain bit depth though, but that's bikeshedding at this point.

> > Maybe this needs to be `Option` as well?
> 
> I think we should change it to `None` if its not supported.

Exactly: if this function can return `Option`, its value will always be `None` on Android.

> My assumption was always that the main use-case here is to check for HDR support, which Android would support with [`android.view.Display.getSupportedHdrTypes()`](https://developer.android.com/reference/android/view/Display.Mode#getSupportedHdrTypes()). 

You can't do that with bit-depth alone, e.g. RGBA8888 and R10G10B10A2 both use 32 bits. Both could be used for a HDR _color space_ but the former likely has visible banding.

> But that API should probably look different and again, not useful to handle in Winit. 

Yeah, HDR is much, _much_ more complicated than that. Better leave it to a users' Vulkan/Dxgi/(E)GL implementation.

Though if a platform provides additional HDR-related details for the current/negotiated _mode_ for a monitor via its native API, it _might_ be nice to expose that on the mode handle.
(things like HDR metadata, VRR, color space from the mode; you have that in Windows and DRM at least)

> E.g., we discussed querying pixel format support in Softbuffer already https://github.com/rust-windowing/softbuffer/issues/98 and Wgpu obviously already supports that.

Seems like that should be disconnected from winit, yes.

> If you don't mind I would leave it like this here and make a follow up PR discussing any further changes. WDYT in any case about all this?

:+1:

src/platform_impl/android/mod.rs
MarijnS95: I think I/we should just play with that idea, but maybe for now let the size of the mode always return `None` or something, to disconnect it from a possible surface?

src/platform_impl/android/mod.rs
daxpedda: > The pixel format of a surface can and will almost always be disjoint from the bit depth of the video mode on a physical panel/connection because of compositors (though it's probably beneficial to keep them in sync depending on what the scanout hardware is capable of).

...

> > In that case the API would be useless on all platforms wouldn't it? Personally I don't know who would use that, so I would actually be in favor of removing it wholesale if we come to a consensus.
> 
> I think all platforms with a list of monitors and their current mode can perfectly benefit from this API. A pixel format may be more helpful over a plain bit depth though, but that's bikeshedding at this point.

Well that would still only make sense if the bit depth has anything to do with the pixel format when you start drawing. Your earlier statement seems to say that this isn't a guarantee?

> > > Maybe this needs to be `Option` as well?
> > 
> > 
> > I think we should change it to `None` if its not supported.
> 
> Exactly: if this function can return `Option`, its value will always be `None` on Android.

Yeah, that's fine. Other platforms don't support querying it as well.

---

It seems this discussion largely boils down to bit depth vs pixel format. In Winit, the bit depth is only useful when you *choose* a video mode handle, which you can via `Window::set_fullscreen()` (not on Android though). But this all implies that the bit depth has some connection to the pixel format, obviously the bit depth doesn't contain all the information of a pixel format, but it does make a difference. E.g. on a 6-bit video mode you can't use 8-bit color pixel format.

In reality of course, the choice is usually between 8-bit (99.9%) and 10-bit per color. No clue how HDR plays a role here, but as we said this would require a different API.

Seems all pointless on Android though if you can't use that information in `Window::set_fullscreen()`, so this basically boils down to use just exposing device data.

---

Conclusion: make it an `Option`.
(will do in a follow-up)

src/platform_impl/android/mod.rs
daxpedda: AFAIK `VideoModeHandle::size()` shouldn't be used to create surfaces anyway, they should just tell you the the resolution in physical pixels of that video mode.

I will add that to the documentation.
Is that enough or is the information exposed here just wrong?

MarijnS95: Exactly that as well. The surface will barely ever be the fill resolution of the screen when statusbars and navbars are removed from the screen.

src/platform_impl/android/mod.rs
daxpedda: We solved this by returning `None` here now.
See https://github.com/rust-windowing/winit/pull/3802#discussion_r1690481503 for more context.

src/platform_impl/android/mod.rs
daxpedda: I've done it in this PR after all.

src/platform_impl/android/mod.rs
daxpedda: I've changed `VideoModeHandle::size()` to return an `Option` now.

daxpedda:
Re-requested reviews because I changed `VideoModeHandle::size()` and `bit_depth()` to return an `Option` now.
Some backends obviously make use of that now instead of returning default values.

src/platform_impl/android/mod.rs
MarijnS95: > It seems this discussion largely boils down to bit depth vs pixel format.

Can't help when discussing how to implement this function, we look at where to pull the necessary information from, and eventually what the API even means and how it is useful. Let me have a look at `Window::set_fullscreen()` and how it utilizes the `bit_depth()` specifically.

> In reality of course, the choice is usually between 8-bit (99.9%) and 10-bit per color.

Again, it's hard to tell those apart if you have `R10G10B10A2` versus `R8G8B8A8` or `R8G8B8X8` which are both 32-bits. We'd need to return the _per-channel_ bit depth in that case?

daxpedda: > > In reality of course, the choice is usually between 8-bit (99.9%) and 10-bit per color.
> 
> Again, it's hard to tell those apart if you have `R10G10B10A2` versus `R8G8B8A8` or `R8G8B8X8` which are both 32-bits. We'd need to return the _per-channel_ bit depth in that case?

It depends if those are part of video modes. If they are pixel formats then I don't think the are relevant to Winit.

src/platform_impl/android/mod.rs
```diff
@@ -768,7 +767,7 @@ impl Window {
     }
 
     pub fn outer_size(&self) -> PhysicalSize<u32> {
-        MonitorHandle::new(self.app.clone()).size()
+        screen_size(&self.app)
```
MarijnS95, line 770: Maybe we should rename this helper to `surface_size()` or inline it entirely?

src/platform_impl/android/mod.rs
```diff
@@ -993,65 +992,61 @@ impl MonitorHandle {
...
-    bit_depth: u16,
-    refresh_rate_millihertz: u32,
     monitor: MonitorHandle,
 }
 
 impl VideoModeHandle {
-    pub fn size(&self) -> PhysicalSize<u32> {
-        self.size.into()
+    pub fn size(&self) -> Option<PhysicalSize<u32>> {
+        None
```
MarijnS95, line 1023: As discussed we're contemplating whether to make it `Option` if we have a chance of falling back to JNI to implement this function - and as a consequence never return `None` on any platform.

MarijnS95 (approved):

src/platform_impl/android/mod.rs
MarijnS95: The only two relevant platforms for this function appear to be Windows and X11, where Windows has it defined here:

https://learn.microsoft.com/en-us/windows/win32/api/wingdi/ns-wingdi-devmodea

Seems to stem from an era where we had less than 32 bits per pixel to define a color (or palette?) and not even a single clue about the pixel _format_ (and related color space).

Let's postpone this discussion now that we can return `None` everywhere this function makes no sense.

src/platform_impl/android/mod.rs
daxpedda: Couldn't inline it because its called in two places.
But renamed!

src/platform_impl/android/mod.rs
daxpedda: Done.

src/platform_impl/android/mod.rs
```diff
@@ -988,70 +983,169 @@ impl Ord for MonitorHandle {
...
+
+    pub fn name(&self) -> Option<String> {
+        let jvm = ffi::jvm(&self.app);
+        let mut env = ffi::env(&jvm);
+
+        Some(
+            ffi::context(&mut env)
+                .display_manager(&mut env)
+                .display(&mut env, self.id)
+                .unwrap()
```
daxpedda, line 1036: This could panic, but is a known issue on all backends.
See https://github.com/rust-windowing/winit/issues/3258.

src/monitor.rs
```diff
@@ -57,23 +62,14 @@ impl VideoModeHandle {
...
     /// available per color. This is generally 24 bits or 32 bits on modern
     /// systems, depending on whether the alpha channel is counted or not.
-    ///
-    /// ## Platform-specific
-    ///
-    /// - **Wayland / Orbital:** Always returns 32.
-    /// - **iOS:** Always returns 32.
     #[inline]
-    pub fn bit_depth(&self) -> u16 {
+    pub fn bit_depth(&self) -> Option<u16> {
```
kchibisov, line 66: Let's change it to `NonZero` as well I guess, can't be zero anyway.

src/platform_impl/android/ffi.rs
daxpedda, line 1: @MarijnS95 let me know if you think anything in here might actually panic.

kchibisov (approved):
I'd probably separate android impl into separate commit and rebase 2 commits in the end, but up to you.

src/platform_impl/android/mod.rs
kchibisov: Well, if you function returns `Option` you can easily avoid panic.

daxpedda: I will make a follow-up PR where I finally tackle #3258.

kchibisov: The thing is that monitor data, such as name, etc, is _static_ more or less. The position could change, indeed, but generally unless we make all of that into events and lifetime I'm not sure we can do much.

daxpedda: Good point.
I'm happy to cache the name, but again, lets discuss it in a follow-up PR.

daxpedda:
> I'd probably separate android impl into separate commit and rebase 2 commits in the end, but up to you.

I did change it up a bit to make it easier to review.
Thanks for the suggestion!

MarijnS95:
Without going through the details, it looks like the Android implementation is quite overdone, now with JNI. Even if we discussed this before, I'd like to revise my statement.

When I proposed looking at Android functions and eventually calling them through JNI, it was meant as a hint that there appear to be matching APIs, and could eventually be used _if there is a pressing need for it_.
Using JNI also beats the user preference for a "native" activity (even though that itself is driven by some Java code, the underlying _implementations_ that you talk against are all native once they've left the `Activity` lifecycle).

---

After all, having multiple displays on Android is extremely unlikely, and even for the current display, what is the user going to use this information for? Anything involving dimensions, pixel formats, refresh rate or color spaces will _always end up going through the `Surface`_. I think Wayland and specifically its color management protocol are opting for the same thing: everything goes through the surface, and the app doesn't and shouldn't really need to know about the display layout.

---

Summary: even if we could (as you have already shown): are you sure we want to start having and managing this blob of JNI "noise" in the Android backend? For likely zero actual use-case?

daxpedda:
> Summary: even if we could (as you have already shown): are you sure we want to start having and managing this blob of JNI "noise" in the Android backend? For likely zero actual use-case?

AFAIC I was only doing this to prevent having an `Option` on this method and learning a bit of JNI on the way.
I'm happy to remove the whole thing **but** Android should then return no `MonitorHandle` at all to prevent this issue, exactly like it was for Web before #3801.

On the other hand I've already implemented it and we will need some JNI sooner or later anyway (I assume?). I don't know if a multi-monitor setup on Android would even work with the current API (not talking about the implementation) or if it actually needs a dedicated platform-specific API (which I would assume).

Let me know how you would like to proceed.

MarijnS95:
Besides the effort you've put in, my preference goes out to not returning this information to the user at all. In case someone requests it we can always aim to re-land your commit from this PR (let's track it somewhere just in case) or keep it around as a reference for users desiring to implement this themselves?

I am not thinking of much that needs JNI now, except maybe the PRs that are ramping up towards IME support (whether that belongs in our out `winit`, I don't know)?

Either way, I've always desired to create an ["`android-support`"](https://github.com/MarijnS95/android-support) kind of crate (under the `rust-mobile` umbrella) that abstracts away the JNI code (and could use precompiled Java/Kotlin classes for the real nasty stuff that is just too horrible to write in Rust via JNI). Details aside, we could move your JNI code there already?

kchibisov:
I'd note that if `Android` stuff is controversial, we can split it into separate PR, since the first 3 commits should be good to go.

daxpedda:
> Besides the effort you've put in, my preference goes out to not returning this information to the user at all. In case someone requests it we can always aim to re-land your commit from this PR (let's track it somewhere just in case) or keep it around as a reference for users desiring to implement this themselves?

I'm unsure why we would remove it in the first place if we would be willing to put it back in?
But maybe we just don't know yet *if* and how we want JNI in Winit yet?

> Either way, I've always desired to create an ["`android-support`"](https://github.com/MarijnS95/android-support) kind of crate (under the `rust-mobile` umbrella) that abstracts away the JNI code (and could use precompiled Java/Kotlin classes for the real nasty stuff that is just too horrible to write in Rust via JNI). Details aside, we could move your JNI code there already?

Yes, lets discuss it on Matrix further!

> I'd note that if `Android` stuff is controversial, we can split it into separate PR, since the first 3 commits should be good to go.

I'm gonna do the removal in a couple of hours or so.
But we should wait for somebody to test/properly review the MacOS changes.

daxpedda:
@MarijnS95 PR is ready for your approval.

MarijnS95:
> I'm unsure why we would remove it in the first place if we would be willing to put it back in?
> But maybe we just don't know _if_ and how we want JNI in Winit yet?

Yeah, this mostly. I don't want to commit to maintaining JNI inside `winit` now, especially given that there doesn't appear to be any usefulness in exposing Android displays and modes via `winit`. If someone comes up with a tangible use-case, we can re-evaluate, perhaps by having your code in a separate crate that's more generically usable by the Android Rust ecosystem (as proposed above, kind of like a sibling to the `ndk` crate but for Java-only interfaces), and call that from `winit` instead :+1:

src/platform_impl/android/mod.rs
```diff
@@ -972,86 +965,60 @@ impl Display for OsError {
...
-        if let Some(native_window) = self.app.native_window() {
-            PhysicalSize::new(native_window.width() as _, native_window.height() as _)
-        } else {
-            PhysicalSize::new(0, 0)
-        }
-    }
-
-    pub fn position(&self) -> PhysicalPosition<i32> {
-        (0, 0).into()
+    pub fn position(&self) -> Option<PhysicalPosition<i32>> {
```
MarijnS95, line 976: Just checking: was this made `Option` because of Android?

src/changelog/unreleased.md
```diff
@@ -101,8 +104,12 @@ changelog entry.
   This feature was incomplete, and the equivalent functionality can be trivially achieved outside
   of `winit` using `objc2-ui-kit` and calling `UIDevice::currentDevice().userInterfaceIdiom()`.
 - On Web, remove unused `platform::web::CustomCursorError::Animation`.
+- Remove `MonitorHandle::size()` and `refresh_rate_millihertz()` in favor of
+  `MonitorHandle::current_video_mode()`.
+- On Android, remove all `MonitorHandle` support instead of emitting false data.
```
MarijnS95, line 109: "and `VideoModeHandle`", or is that implicit to `MonitorHandle`?

MarijnS95 (approved):
Android looks good!

src/changelog/unreleased.md
daxpedda: Yes, because without a `MonitorHandle` there is no way to acquire a `VideoModeHandle`.

src/platform_impl/android/mod.rs
daxpedda: No, Web is the only one that actually needs it so far.
Windows doesn't count because its only using `None` when the handle becomes invalid, which we have to solve differently in the future anyway.

PR #3803, opened 2024-07-20
Web: use raw data in `DeviceEvent::MouseMotion` when using `CursorGrabMode::Confined`

### LINKED ISSUE #3547: Web: use raw mouse input when locking pointer
Currently when using `Window::set_cursor_grab(CursorGrabMode::Locked)` on Web mouse movement will contain mouse acceleration. AFAIU other backends send raw mouse events with mouse acceleration disabled, so Web should follow suit.

This would use [`Element.requestPointerLock` with the `unadjustedMovement` options](https://developer.mozilla.org/en-US/docs/Web/API/Element/requestPointerLock#unadjustedmovement), which is notable only supported by Chrome and even there not on Linux.

daxpedda:
This uses [`Element.requestPointerLock` with the `unadjustedMovement` options](https://developer.mozilla.org/en-US/docs/Web/API/Element/requestPointerLock#unadjustedmovement), which is notable only supported by Chrome and even there not on Linux.

I also updated the documentation to properly reflect the current state.

Needs entry in #2875.
Fixes #3547.

### DIFF
--- a/src/changelog/unreleased.md
+++ b/src/changelog/unreleased.md
@@ -44,6 +44,9 @@ changelog entry.
 
 - Add `ActiveEventLoop::create_proxy()`.
 - On Web, implement `Error` for `platform::web::CustomCursorError`.
+- On Web, add `ActiveEventLoopExtWeb::is_cursor_lock_raw()` to determine if
+  `DeviceEvent::MouseMotion` is returning raw data, not OS accelerated, when using
+  `CursorGrabMode::Locked`.
 
 ### Changed
 
@@ -71,6 +74,8 @@ changelog entry.
   `EventLoopExtRunOnDemand::run_app_on_demand` to accept a `impl ApplicationHandler` directly,
   instead of requiring a `&mut` reference to it.
 - On Web, `Window::canvas()` now returns a reference.
+- On Web, `CursorGrabMode::Locked` now lets `DeviceEvent::MouseMotion` return raw data, not OS
+  accelerated, if the browser supports it.
 
 ### Removed
 
--- a/src/event.rs
+++ b/src/event.rs
@@ -467,6 +467,22 @@ pub enum DeviceEvent {
     ///
     /// This represents raw, unfiltered physical motion. Not to be confused with
     /// [`WindowEvent::CursorMoved`].
+    ///
+    /// ## Platform-specific
+    ///
+    /// **Web:** Only returns raw data, not OS accelerated, if [`CursorGrabMode::Locked`] is used
+    /// and browser support is available, see
+    #[cfg_attr(
+        any(web_platform, docsrs),
+        doc = "[`ActiveEventLoopExtWeb::is_cursor_lock_raw()`][crate::platform::web::ActiveEventLoopExtWeb::is_cursor_lock_raw()]."
+    )]
+    #[cfg_attr(
+        not(any(web_platform, docsrs)),
+        doc = "`ActiveEventLoopExtWeb::is_cursor_lock_raw()`."
+    )]
+    ///
+    #[rustfmt::skip]
+    /// [`CursorGrabMode::Locked`]: crate::window::CursorGrabMode::Locked
     MouseMotion {
         /// (x, y) change in position in unspecified units.
         ///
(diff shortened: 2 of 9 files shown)


src/changelog/unreleased.md
```diff
@@ -44,6 +44,9 @@ changelog entry.
 
 - Add `ActiveEventLoop::create_proxy()`.
 - On Web, implement `Error` for `platform::web::CustomCursorError`.
+- On Web, add `ActiveEventLoopExtWeb::is_cursor_lock_raw()` to determine if
+  `DeviceEvent::MouseMotion` is returning raw data, not OS accelerated, when using
+  `CursorGrabMode::Confined`.
```
kchibisov, line 49: it's `Locked`.

src/changelog/unreleased.md
```diff
@@ -69,6 +72,8 @@ changelog entry.
 - Change signature of `EventLoop::run_app`, `EventLoopExtPumpEvents::pump_app_events` and
   `EventLoopExtRunOnDemand::run_app_on_demand` to accept a `impl ApplicationHandler` directly,
   instead of requiring a `&mut` reference to it.
+- On Web, `CursorGrabMode::Confined` now lets `DeviceEvent::MouseMotion` return raw data, not OS
```
kchibisov, line 75: locked.

src/event.rs
```diff
@@ -467,6 +467,21 @@ pub enum DeviceEvent {
     ///
     /// This represents raw, unfiltered physical motion. Not to be confused with
     /// [`WindowEvent::CursorMoved`].
+    ///
+    /// ## Platform-specific
+    ///
+    /// **Web:** Always returns OS accelerated data unless [`CursorGrabMode::Confined`] is used depending on browser support, see
```
kchibisov, line 473: locked, and delta is unaccelerated.

src/window.rs
```diff
@@ -1478,6 +1478,23 @@ impl Window {
     ///     .unwrap();
     /// # }
     /// ```
+    ///
+    /// ## Platform-specific
+    ///
+    /// **Web:** Always returns an [`ExternalError::NotSupported`] with
+    /// [`CursorGrabMode::Confined`]. [`DeviceEvent::MouseMotion`] might return OS accelerated data
```
kchibisov, line 1485: locked.

kchibisov (changes requested):
I'd assume that you mean `Locked` pointer, which is the one that stays in place.

Also, could you change the cursor position when the cursor is locked, since it's other API that usually works with locked cursor?

And I'd assume there's no way to confine cursor at all on web?

daxpedda:
> Also, could you change the cursor position when the cursor is locked, since it's other API that usually works with locked cursor?

You can't change the cursor position on Web. What other API?
I'm not sure what you mean with that question.

> And I'd assume there's no way to confine cursor at all on web?

Not AFAIK.

kchibisov:
>You can't change the cursor position on Web. What other API?
I'm not sure what you mean with that question.

Locked cursor is usually works in pair with `set_cursor_position` API, since otherwise it's sort of useless, unless you completely hide the cursor and draw everything yourself, which could also be an option.

On Wayland, for example, where manipulating with cursor is also prohibited, the _locked pointer_ has its own method to set its position, but it only works when you have the cursor lock present.

It's not really mandatory, it's just changes how you write things basically.

src/event.rs
daxpedda: It says "accelerated data **unless**".
I removed the negation which makes it easier to understand though.

Thanks!

src/window.rs
daxpedda: This one is actually correct!

daxpedda:
> Locked cursor is usually works in pair with `set_cursor_position` API, since otherwise it's sort of useless, unless you completely hide the cursor and draw everything yourself, which could also be an option.

Ah I see!
Yes, on Web the cursor is hidden, so its not a problem.
The position of the cursor remains the same as the position it was when it was locked.

Is the cursor not hidden in other backends when locked?
If not I should document this for Web.

kchibisov:
>Is the cursor not hidden in other backends when locked?
If not I should document this for Web.

it's generally not hidden and users tend to hide it themselves.

daxpedda:
> > Is the cursor not hidden in other backends when locked?
> > If not I should document this for Web.
> 
> it's generally not hidden and users tend to hide it themselves.

Ah, `CursorGrabMode::Locked` already says:
> There's no guarantee that the cursor will be hidden. You should hide it by yourself if you want to do so.

I removed some duplicate documentation now that I see `CursorGrabMode` includes it already.

kchibisov (approved):

PR #3806, opened 2024-07-20
Rename refresh rate

daxpedda:
While working on #3802 I wondered about `VideoModeHandle::refresh_rate_millihertz()`s really long name.
WDYT about renaming it to `refresh_rate()`?
Alternatively `refresh_rate_mh()`?

Requires #3802.

### DIFF
--- a/examples/window.rs
+++ b/examples/window.rs
@@ -236,15 +236,17 @@ impl Application {
                 info!("{intro}: [no name]");
             }
 
-            let PhysicalSize { width, height } = monitor.size();
-            info!(
-                "  Current mode: {width}x{height}{}",
-                if let Some(m_hz) = monitor.refresh_rate_millihertz() {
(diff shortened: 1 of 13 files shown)


kchibisov:
I think the reason it's named like that is because we wanted to change the precision of it. Maybe we should return `Duration` though and call it `refresh_rate`?

daxpedda:
> Maybe we should return `Duration` though and call it `refresh_rate`?

This actually made me look into the definition of the word "Refresh Rate" and while I couldn't find anything better quick, [Wikipedia says](https://en.wikipedia.org/wiki/Refresh_rate):
> ... the number of times per second that a raster-based display device displays a new image.

So it would make sense to use Hz for this. The fact that we renamed it makes a lot of sense now.

Reading through https://github.com/rust-windowing/winit/pull/2352 also convinced me that the way it is right now is better then what I'm proposing here.

PR #3809, opened 2024-07-22
macOS: fix building with `feature = "rwh_04"`

daxpedda:
I noticed while working on MacOS that I had some build error in `rwh_04::HasRawWindowHandle`, but always thought there was just something wrong with my cross-compile setup. I believe the issue was caused in https://github.com/rust-windowing/winit/pull/3391, where all calls to `contentView()` were renamed while this one was forgotten.

This should have been caught by the CI, but we don't check these features in CI. Luckily we may not want to actually fix the CI and instead remove this compatibility layer?

If this patch is correct, it should also be broken in [v0.30](https://github.com/rust-windowing/winit/blob/v0.30.4/src/platform_impl/macos/window_delegate.rs#L1611-L1618).

Cc @madsmtm.

### DIFF
--- a/src/changelog/unreleased.md
+++ b/src/changelog/unreleased.md
@@ -82,3 +82,7 @@ changelog entry.
   This feature was incomplete, and the equivalent functionality can be trivially achieved outside
   of `winit` using `objc2-ui-kit` and calling `UIDevice::currentDevice().userInterfaceIdiom()`.
 - On Web, remove unused `platform::web::CustomCursorError::Animation`.
+
+### Fixed
+
+- On MacOS, fix building with `feature = "rwh_04"`.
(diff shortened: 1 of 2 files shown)


kchibisov (approved):

kchibisov:
We'd have to fix and backport into 0.30.0, but I think we can just remove the 0.4.0 and maybe even 0.5.0 handle.

I don't think there're many stuff around with those and you can kind of convert between them yourself if you really need them back.

kchibisov:
>If this patch is correct, it should also be broken in [v0.30](https://github.com/rust-windowing/winit/blob/v0.30.4/src/platform_impl/macos/window_delegate.rs#L1611-L1618).

0.4 is the same as 0.5, so it's correct.

PR #3810, opened 2024-07-22
Pen/Stylus support

### LINKED ISSUE #99: Pen/Tablet Input Support

daxpedda:
WIP!

Fixes #99.


src/event.rs
```diff
@@ -907,6 +908,228 @@ pub type AxisId = u32;
...
+    Unknown {
+        /// Name of the cursor type.
+        name: Option<SmolStr>,
+        state: Option<ToolState>,
+    },
+}
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+#[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
+pub struct ToolState {
```
daxpedda, line 929: I would be glad for an alternative to the term "Tool" or "State".
Originally got the idea from [Linux](https://github.com/torvalds/linux/blob/933069701c1b507825b514317d4edd5d3fd9d417/include/uapi/linux/input-event-codes.h#L402-L406), because we need a term that applies to all sorts of different tools.
But honestly I think "Pen" would have been fine as well?

Maybe "Stylus" would fit all the possible tools:
- Pen
- Eraser
- Brush
- Pencil
- Airbrush

src/event.rs
```diff
@@ -907,6 +908,228 @@ pub type AxisId = u32;
...
+    Mouse,
+    Pen(ToolState),
+    Eraser(ToolState),
+    /// Unknown cursor type. E.g. cursor was moved by a voice-recognition software or a keyboard
+    /// interaction.
+    Unknown {
+        /// Name of the cursor type.
+        name: Option<SmolStr>,
+        state: Option<ToolState>,
+    },
```
daxpedda, line 924: I believe this to be useful, but let me know if you disagree.

src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -1377,6 +1381,7 @@ impl EventProcessor {
                     event: WindowEvent::CursorMoved {
                         device_id: mkdid(util::VIRTUAL_CORE_POINTER),
                         position: location.cast(),
+                        r#type: CursorType::Mouse,
                     },
                 };
                 callback(&self.target, event);
```
daxpedda, line 1387: > Mouse cursor position changes when touch events are received.
> Only the first concurrently active touch ID moves the mouse cursor.

This is a bit confusing to me, especially because I believe it doesn't map to other backends. Why do we tell the user that the mouse moves if the cursor moves with touch input?
Mouse input is cursor movement, but not all cursor movement is mouse input.

I'm really not sure what the purpose of this is.

src/event.rs
```diff
@@ -262,12 +266,12 @@ pub enum WindowEvent {
     /// [`transform`]: https://developer.mozilla.org/en-US/docs/Web/CSS/transform
     CursorLeft { device_id: DeviceId },
 
+    /// A cursor button press has been received.
+    CursorInput { device_id: DeviceId, state: ElementState, button: CursorButton },
```
daxpedda, line 270: At this point we should probably rename "Cursor" to "Pointer".
"Cursor" commonly refers to the visual representation of an input device on the screen.
But "Pointer" is a broader category that can include any device and is not specific to its visual representation on the screen.

I couldn't really find anything concrete on this online.
WDYT?

src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -1121,7 +1125,7 @@ impl EventProcessor {
 
             let event = Event::WindowEvent {
                 window_id,
-                event: WindowEvent::CursorMoved { device_id, position },
+                event: WindowEvent::CursorMoved { device_id, position, r#type: CursorType::Mouse },
```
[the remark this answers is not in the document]
daxpedda: Okay I figured this one out as well.
After digging around in Chromium for a while I found [how they determined if this was a stylus or not](https://source.chromium.org/chromium/chromium/src/+/c85f08af47c25bc5eb4a7f81c0ac0620d8d767ca:ui/events/devices/x11/device_data_manager_x11.cc;l=198-200).

After some digging, I also found that [Glazier was doing something quite similar](https://github.com/linebender/glazier/blob/8d253c3d4ad12c1dd491895bf3eb8ef1728d01b6/src/backend/x11/pointer.rs#L146-L157), in addition to detecting erasers (in a hacky way, but it works). There is a lot to learn from them when actually implementing the backends for Pens/Stylus's.

Additionally, after some more digging into X11, I found that there is an alternative to figuring out the type: [`XListInputDevices`](https://www.x.org/releases/current/doc/man/man3/XListInputDevices.3.xhtml). That's right, using the old XInput version and now way to query individual devices. This exposes what they call "Device Types", which unfortunately don't include pens, but they let you recognize the drawing tablet,  which works as well.

I found that [Chromium was doing something similar for touchpads, but not for pens](https://source.chromium.org/chromium/chromium/src/+/0a212cf3cbde2f8639dc3238dfe4762969bd170d:ui/events/devices/x11/device_data_manager_x11.cc;l=251-258) ([and a comment in a much older revision explaining this as well](https://source.chromium.org/chromium/chromium/src/+/9054218f820a9d0502c24db35837c81a8f4ffa73:ui/events/devices/x11/touch_factory_x11.cc;l=86-100)). So maybe this will be useful for us in the future.

src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -1055,21 +1058,30 @@ impl EventProcessor {
         // Set the timestamp.
         wt.xconn.set_timestamp(event.time as xproto::Timestamp);
 
+        let Some(DeviceType::Mouse) = self
+            .devices
+            .borrow()
+            .get(&DeviceId(event.sourceid as xinput::DeviceId))
```
daxpedda, line 1064: This came as a surprise to me, but I had to do quite some digging to figure out why this wasn't working for me.
`deviceId` here uses the cursor, but `sourceId` actually uses the device that moved the cursor. Shouldn't we use that to send to the user?

src/event.rs
```diff
@@ -939,6 +1177,28 @@ pub enum MouseButton {
...
+        match tool {
+            ToolButton::Contact => MouseButton::Left,
+            ToolButton::Barrel => MouseButton::Right,
+            ToolButton::Other(1) => MouseButton::Middle,
+            ToolButton::Other(3) => MouseButton::Back,
+            ToolButton::Other(4) => MouseButton::Forward,
+            ToolButton::Other(other) => MouseButton::Other(other),
+        }
+    }
+}
```
daxpedda, line 1200: I'm not sure if this maps correctly on all backends?

daxpedda:
After some discussion and thinking, I created #3833, which would basically result in a pre-cursor to this PR.
The implementation of #3833 will be partly based on the code here.

I will resume this PR when we are done with #3833.

DorianRudolph:
Great work on this PR and the pointer event overhaul! May I ask what the status of this PR is? This is something I have wanted for quite some time.

I made an implementation based on this PR (with some modifications) for wayland https://github.com/WhiteboardCX/winit/tree/tablet
- I use f64 for tilt/twist/angle because wayland reports these as floats and it feels more natural.
- I only use a single `Tool` kind for `PointerKind`, instead of separate eraser and pen. This aligns more with the wayland interface. The buttons on the pen are handled as PointerButton events and mapping these to pen/eraser/... then can be handled by the application.
- I also added a `ToolType`, currently only pen and eraser, though wayland supports many [more](https://wayland.app/protocols/tablet-v2#zwp_tablet_tool_v2:enum:type). However, I don't have hardware with different tool types. Presumably pens with an eraser on the back should work.
- Wayland provides some extra features like distance, hardware ids for pens, etc., which presumably are also only available on higher end wacom tablets (I only have Huion and "One by Wacom"). I did not implement these.
- Strangely, the wayland [protocol](https://wayland.app/protocols/tablet-v2) does not seem to have a way to associate the tablet tools with the drawing tablet if there are multiple (not sure if it matters). The `device_id` is just assigned in the order in which the pens appear (my tablet's don't report any of the ids in the protocol). So the same tool could get a different id when it reappears.

If my changes are OK, I can also make a PR. I should have hardware to test all platforms, but implementing all of them will take some time.

madsmtm:
> May I ask what the status of this PR is

It's nominated for our next meeting, which should happen sometime in start January

DorianRudolph:
Thanks. In the meantime, I added support for windows to my fork and created a small [demo](https://github.com/WhiteboardCX/demo/blob/main/Cargo.toml) based on vello.

wareya:
Does this PR need to be redone on top of https://github.com/rust-windowing/winit/pull/3876? I have a lot of free time right now and like the above poster I did a test implementation of exposing tablet events on Windows. I could do the same for linux, android, and web too if this PR's creator is busy. https://github.com/wareya/winit/tree/tablet

DorianRudolph:
Hey, I think my version of this was already on top of that commit https://github.com/WhiteboardCX/winit/tree/tablet
I also had Linux/Wayland and windows support.

I would be happy to work with you on this. In principle, I also have Mac and iPad hardware, so I could try adding support for these as well.

At this point my fork probably needs some work again to run on top of the latest master branch, and it would be helpful if someone else went over my design decisions.

madsmtm:
Thanks for the interest. I unfortunately don't think we're equipped to review such work at the moment (I plan to post something about that soon), would it be okay if I maybe got back to you in a few months?

wareya:
Are you planning on doing work that would require us to rewrite whatever we come up with, like swapping out platform integrations? If not, I'm willing to do whatever you guys need to make reviewing easier (e.g. limiting its scope, or sticking to a basic design that won't need breakage to be extended later, limiting the platforms supported at first to ones that are unlikely to need changing, etc).

DasLixou:
Don't want to promise anything but it seems that after the backends are now split in seperate crates, the awaited "refactor" is done?
I might try to adopt this PR, make it work with the split and also try to integrate Apple Pencil support. (Still, nothing promised, but I'd like to know if now is the time and if it would be appreciated)

madsmtm:
Superseded by https://github.com/rust-windowing/winit/pull/4318.

PR #3820, opened 2024-07-24
Remove `DeviceEvent::Added` and `Removed`

daxpedda:
This had no real use because we don't expose any information on `DeviceId` except on Windows. But there we only expose the name.

The assumption is that this was originally added for gamepad support, so this might be added back again, but with a accompanying UUID.

These events were only implemented on Windows and X11.


kchibisov:
maybe users track/build some state based on the ids and remove when device is lost, but I also don't see any use case in those events now.

kchibisov (approved):

PR #3822, opened 2024-07-25
Fix CI for Rust v1.80

daxpedda:
[`clippy::doc_lazy_continuation`](https://rust-lang.github.io/rust-clippy/master/index.html#/doc_lazy_continuation) was added, which needed some fixing from our side.


kchibisov:
I think they should generally format like that with rustfmt. Probably it will be in the future.

PR #3823, opened 2024-07-25
Basic iOS IME support

lucasmerlin:
This implements basic iOS IME support (typing, backspace, support for emojis etc but no autocomplete or copy / paste menu).
I've tested this with egui on winit 0.29 but unfortunately I wasn't able to test this yet with winit 0.30 since egui isn't updated yet.

This is an alternative implementation of #3571, implementing UIKeyInput for the WinitView instead of creating a extra text_field.

I'll leave this in draft until I was able to test this with egui and winit 0.30

### DIFF
--- a/src/platform_impl/apple/uikit/view.rs
+++ b/src/platform_impl/apple/uikit/view.rs
@@ -551,6 +551,7 @@ impl WinitView {
             mtm,
             text.to_string().chars().flat_map(|c| {
                 let text = smol_str::SmolStr::from_iter([c]);
+                // Emit both press and release events
                 [ElementState::Pressed, ElementState::Released].map(|state| {
                     EventWrapper::StaticEvent(Event::WindowEvent {
                         window_id,
@@ -561,7 +562,7 @@ impl WinitView {
                                 } else {
                                     None
                                 },
-                                state: ElementState::Pressed,
+                                state,
                                 location: KeyLocation::Standard,
                                 repeat: false,
                                 logical_key: Key::Character(text.clone()),
--- a/src/platform_impl/apple/uikit/view.rs
+++ b/src/platform_impl/apple/uikit/view.rs
@@ -4,19 +4,21 @@ use std::cell::{Cell, RefCell};
 use objc2::rc::Retained;
 use objc2::runtime::{NSObjectProtocol, ProtocolObject};
 use objc2::{declare_class, msg_send, msg_send_id, mutability, sel, ClassType, DeclaredClass};
-use objc2_foundation::{CGFloat, CGPoint, CGRect, MainThreadMarker, NSObject, NSSet};
+use objc2_foundation::{CGFloat, CGPoint, CGRect, MainThreadMarker, NSObject, NSSet, NSString};
 use objc2_ui_kit::{
     UICoordinateSpace, UIEvent, UIForceTouchCapability, UIGestureRecognizer,
-    UIGestureRecognizerDelegate, UIGestureRecognizerState, UIPanGestureRecognizer,
+    UIGestureRecognizerDelegate, UIGestureRecognizerState, UIKeyInput, UIPanGestureRecognizer,
     UIPinchGestureRecognizer, UIResponder, UIRotationGestureRecognizer, UITapGestureRecognizer,
-    UITouch, UITouchPhase, UITouchType, UITraitEnvironment, UIView,
+    UITextInputTraits, UITouch, UITouchPhase, UITouchType, UITraitEnvironment, UIView,
 };
 
 use super::app_state::{self, EventWrapper};
 use super::window::WinitUIWindow;
 use super::{FingerId, DEVICE_ID};
 use crate::dpi::PhysicalPosition;
-use crate::event::{Event, FingerId as RootFingerId, Force, Touch, TouchPhase, WindowEvent};
+use crate::event::{ElementState, Event, FingerId as RootFingerId, Force, KeyEvent, Touch, TouchPhase, WindowEvent};
+use crate::keyboard::{Key, KeyCode, KeyLocation, NamedKey, NativeKeyCode, PhysicalKey};
+use crate::platform_impl::KeyEventExtra;
 use crate::window::{WindowAttributes, WindowId as RootWindowId};
 
 pub struct WinitViewState {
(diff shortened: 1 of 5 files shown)


src/platform_impl/apple/uikit/view.rs
```diff
@@ -512,4 +539,65 @@ impl WinitView {
...
+        unsafe {
+            let mtm = MainThreadMarker::new().unwrap();
+            // send individual events for each character
+            app_state::handle_nonuser_events(
+                mtm,
+                text.to_string().chars().map(|c| {
+                    EventWrapper::StaticEvent(Event::WindowEvent {
+                        window_id,
+                        event: WindowEvent::KeyboardInput {
+                            event: KeyEvent {
```
nicoburns, line 556: I feel like this ought to be generating `Ime` events (new ones if need be) rather than `KeyEvent` events

nicoburns (changes requested):
The general approach of implementing the relevant traits for `WinitView` seems good to me, but I think we should be generating IME-specific events in Winit rather than synthesising keypress events.

src/platform_impl/apple/uikit/view.rs
kchibisov: The Android IME (at least with the proposed patches) does generate the regular key events.

Some phone keyboards have an option to send things like `Control`, etc, so I'm not sure how stuff like that will go through regular `Commit` stuff though.

And there's probably nothing particularly bad in having input via regular keys, especially given that you can plug keyboard on some ios devices.

lucasmerlin: Since UIKeyInput only generates basic key events (no autocomplete etc.) I think KeyEvents make more sense. I think a next step would be to implement UITextInput, which would allow for all the fancy IME stuff like autocomplete and autocorrect, then it would make sense to generate IME events

lucasmerlin:
Since egui was updated with winit 0.30.x support I was now able to test this and it works great! 

https://github.com/user-attachments/assets/ddbe64c9-2298-4144-87bf-36c64c0213bf

src/platform_impl/apple/uikit/view.rs
```diff
@@ -314,6 +316,11 @@ declare_class!(
...
             app_state::handle_nonuser_event(mtm, gesture_event);
         }
+
+        #[method(canBecomeFirstResponder)]
+        fn can_become_first_responder(&self) -> bool {
+            true
+        }
     }
 
     unsafe impl NSObjectProtocol for WinitView {}
```
lucasmerlin, line 326: @madsmtm I was surprised that the order of the impl Traits here mattered, first I put the `impl UIKeyInput` and `impl UITextInputTraits` above `impl NSObjectProtocol` and I would get `failed to add protocol NSObject` when starting the app. Is that expected? I also couldn't find anything about order in the objc2 docs.

src/platform_impl/apple/uikit/view.rs
madsmtm: Indeed, this is a good first step forwards, and we can always work on making IME better

src/platform_impl/apple/uikit/view.rs
madsmtm: Wow... I've been having trouble with this too, but I never considered that it could be the ordering (in hindsight it makes perfect sense, but at the time it just felt like a fluke) - thank you so much for figuring it out!

Have fixed it in https://github.com/madsmtm/objc2/commit/36b4587fc1840b4c6cc24dfcdb7587ecdd81117f, which I'll ship with the next major release (not worth a patch release by itself IMO).

madsmtm (changes requested):
Again, the better solution is probably to send IME events, but for now, this is a fine initial approach, and definitely better than setting up an extra view as was done in https://github.com/rust-windowing/winit/pull/3571.

src/platform_impl/apple/uikit/window.rs
```diff
@@ -365,12 +365,20 @@ impl Inner {
...
+    pub fn set_ime_allowed(&self, allowed: bool) {
+        if allowed {
+            unsafe {
+                self.view.becomeFirstResponder();
+            }
+        } else {
+            unsafe {
+                self.view.resignFirstResponder();
+            }
+        }
```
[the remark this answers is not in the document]
lucasmerlin: This is what actually opens the keyboard. `becomeFirstResponder` requests focus for the winit view, and since it implements UIKeyInput iOS will now show the keyboard. `resignFirstResponder` resigns focus and hides the keyboard. I'll add a comment to explain this.

Not sure if the winit view not being first responder will ever be a problem, but I don't think it will, since before this PR the winit view never was first responder (the default of `canBecomeFirstResponder` is false)

src/platform_impl/apple/uikit/view.rs
```diff
@@ -512,4 +541,66 @@ impl WinitView {
...
+                [ElementState::Pressed, ElementState::Released].map(|state| {
+                    EventWrapper::StaticEvent(Event::WindowEvent {
+                        window_id,
+                        event: WindowEvent::KeyboardInput {
+                            event: KeyEvent {
+                                text: if state == ElementState::Pressed {
+                                    Some(text.clone())
+                                } else {
+                                    None
+                                },
```
lucasmerlin, line 563: @madsmtm I assume the text should only be sent during the pressed event and not the released one? I sent it on both before, but that would cause the chars to be duplicated in egui, so I changed it to be like this.

madsmtm: That is indeed correct.

madsmtm (approved):
Thanks, I'll merge it even if you didn't end up testing it fully with egui!

hacknus:
Thank you very much for this work! How would we go on and implement copy/paste?

hacknus:
I currently implemented the pasting my self with a button in egui, but this is not pretty:

```toml
[dependencies]
objc = "0.2" # Add objc for iOS native bindings
```

```rust
#[cfg(target_os = "ios")]
use objc::runtime::{Object, Class};
#[cfg(target_os = "ios")]
use objc::sel;
#[cfg(target_os = "ios")]
use objc::sel_impl;
#[cfg(target_os = "ios")]
use objc::msg_send;
```

```rust
if ui.button("Paste").clicked() {
    if let Some(text) = unsafe {
        let pasteboard_class = Class::get("UIPasteboard").unwrap();
        let general_pasteboard: *mut Object = msg_send![pasteboard_class, generalPasteboard];
        let text: *mut Object = msg_send![general_pasteboard, string];

        if !text.is_null() {
            let text_str: *mut Object = msg_send![text, UTF8String];
            let c_str = std::ffi::CStr::from_ptr(text_str as *const i8);
            Some(c_str.to_string_lossy().to_string())
        } else {
            None
        }
    } {
        self.pasted_text = text;
    };
}
```

PR #3825, opened 2024-07-26
Handle URLs on  macOS

### LINKED ISSUE #2190: Feature request: Custom URI handling
Hi!

I'm not sure if it's in scope of winit or not, but it would be fantastic if it handled custom uri's; like this: https://developer.apple.com/documentation/xcode/defining-a-custom-url-scheme-for-your-app

https://github.com/maidsafe-archive/system_uri does provide the registration functionality, and it works if the app isn't already running (the app opens with an argument with the uri). But if the app is already running I think something more needs to be done to handle it (at least on OSX). I haven't looked into it in detail, but it feels like it would fit pretty well as a WindowEvent.

nicoburns:
This allows macOS apps to access the URL when they are opened "as a browser / url handler". This is adapted from Iced's fork of Winit and I believe it is the main things preventing Iced from using upstream Winit.

Fixes #2190

### DIFF
--- a/src/application.rs
+++ b/src/application.rs
@@ -325,6 +325,15 @@ pub trait ApplicationHandler {
     fn memory_warning(&mut self, event_loop: &ActiveEventLoop) {
         let _ = event_loop;
     }
+
+    /// Emitted when the application has received a URL, through a
+    /// custom URL handler.
+    ///
+    /// Only supported on macOS.
+    fn received_url(&mut self, event_loop: &ActiveEventLoop, url: String) {
+        let _ = event_loop;
+        let _ = url;
+    }
 }
 
 #[deny(clippy::missing_trait_methods)]
--- a/src/event.rs
+++ b/src/event.rs
@@ -112,6 +112,21 @@ pub(crate) enum Event {
 
     /// User requested a wake up.
     UserWakeUp,
+
+    /// Emitted when the event loop receives an event that only occurs on some specific platform.
+    PlatformSpecific(PlatformSpecific),
+}
+
+/// Describes an event from some specific platform.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum PlatformSpecific {
+    MacOS(MacOS),
+}
+
+/// Describes an event that only happens in `MacOS`.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum MacOS {
+    ReceivedUrl(String),
 }
 
 /// Describes the reason the event loop is resuming.
--- a/src/platform_impl/apple/appkit/app_state.rs
+++ b/src/platform_impl/apple/appkit/app_state.rs
@@ -6,7 +6,10 @@ use std::sync::Arc;
 use std::time::Instant;
 
 use objc2::rc::Retained;
-use objc2::{declare_class, msg_send_id, mutability, ClassType, DeclaredClass};
+use objc2::runtime::AnyObject;
+use objc2::{
+    class, declare_class, msg_send, msg_send_id, mutability, sel, ClassType, DeclaredClass,
+};
 use objc2_app_kit::{NSApplication, NSApplicationActivationPolicy, NSApplicationDelegate};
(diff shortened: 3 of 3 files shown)


src/application.rs
```diff
@@ -325,6 +325,15 @@ pub trait ApplicationHandler {
...
     }
+
+    /// Emitted when the application has received a URL, through a
+    /// custom URL handler.
+    ///
+    /// Only supported on macOS.
+    fn received_url(&mut self, event_loop: &ActiveEventLoop, url: String) {
+        let _ = event_loop;
+        let _ = url;
+    }
```
kchibisov, line 336: Could you create a macOS specific `ApplicationHandler` and add it there?

kchibisov (commented):
I think @madsmtm has a different idea on how to approach all of that, but I don't mind having it under macOS specific application handler.

Also, it should be added without a specific event, just rely on the newly added callback.

madsmtm (changes requested):
I would very much like to get Iced back to using mainline Winit!

I just merged https://github.com/rust-windowing/winit/pull/3758, could you check if that allows you to do what Iced needs? See in particular [the new docs under `winit::platform::macos`](https://rust-windowing.github.io/winit/winit/platform/macos/index.html).

nicoburns:
@madsmtm Super excited to see https://github.com/rust-windowing/winit/pull/3758 merged. That looks like an absolutely fantastic solution for modular Winit on apple platforms. I'll have to implement it to be sure, but I reckon it will be possible for this to be implemented outside of Winit on top of #3758. And I reckon I can use some of the objc2 crates' functionality to really clean it up too. The one question I have is when/where I ought to run the code that registers the listener for the event that returns the url:

I could use my own AppDelegate and register it in applicationWillFinishLaunching like the current PR does. But I would really like to be able to wrap this up in a library, and my feeling is that the approach you've used in #3758 ought to enable that. But that perhaps the `create_observer` function ought to become public so it can be used in 3rd party libs as well as winit.

Regarding event registration ordering:
- I'm guessing I ought to put my code in Winit's "resume" event?
- Do you think I could register for the `kAEGetURL`  event directly in "resume"?
- Or do you think I would need to register for applicationWillFinishLaunching in "resume" and then register for `kAEGetURL`?

madsmtm:
> perhaps the `create_observer` function ought to become public so it can be used in 3rd party libs

It is quite a simple function, and will become even simpler in the future once `block2` supports references in arguments. If you really feel you need it, just copy the definition.

> Regarding event registration ordering:
> 
>     * I'm guessing I ought to put my code in Winit's "resume" event?
> 
>     * Do you think I could register for the `kAEGetURL`  event directly in "resume"?
> 
>     * Or do you think I would need to register for applicationWillFinishLaunching in "resume" and then register for `kAEGetURL`?

I don't know anything about the `kAEGetURL` event, and I'm unfamiliar with these kinds of lower-level events, but it seems probable that the registration needs to happen before the application has launched. You may be able to do it at the top of main (or possibly right after `EventLoop::new`), but otherwise, doing it in `applicationWillFinishLaunching:` could be done with something like:

```rust
let block = RcBlock::new(|notification: NonNull<NSNotification>| {
    let notification = unsafe { notification.as_ref() };
    // ... Register in here
});
let observer = unsafe {
    center.addObserverForName_object_queue_usingBlock(
        // `applicationWillFinishLaunching:`
        Some(unsafe { NSApplicationWillFinishLaunchingNotification }),
        None,
        None,
        &block,
    )
}
// Keep `observer` around until the end of `main` (or leak it).
```

madsmtm:
Not sure if you got the time to play around with this yet, but again, I'm fairly confident that https://github.com/rust-windowing/winit/pull/3758 will solve Iced's needs here.

Feel free to re-open or discuss further if that's not the case!

nicoburns:
@madsmtm I haven't had time to play around with this yet, but I am also confident that https://github.com/rust-windowing/winit/pull/3758 allows this to be solved outside of Winit.

We probably ought to document the pattern described in https://github.com/rust-windowing/winit/issues/4260#issuecomment-2945457455 somewhere.

PR #3827, opened 2024-07-26
Use `dyn ActiveEventloop` in `ApplicationHandler`

kchibisov:
--

This should allow backends to completely remove the complexity layer from the `RootActiveEventLoop` and other `Root` structs, because they'll be able to use them directly.

This is based on `winit-next` efforts around the `as_any` and dynamic traits.

For now it's done just for Wayland to show how it'll look for other backends.

The next steps forward after that would be to make types that event loop returns into the `dyn` counterparts.

### DIFF
--- a/src/platform/wayland.rs
+++ b/src/platform/wayland.rs
@@ -24,10 +24,10 @@ pub trait ActiveEventLoopExtWayland {
     fn is_wayland(&self) -> bool;
 }
 
-impl ActiveEventLoopExtWayland for ActiveEventLoop {
+impl ActiveEventLoopExtWayland for &dyn ActiveEventLoop {
     #[inline]
     fn is_wayland(&self) -> bool {
-        self.p.is_wayland()
+        self.as_any().downcast_ref::<crate::platform_impl::wayland::ActiveEventLoop>().is_some()
     }
 }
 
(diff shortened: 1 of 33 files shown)


kchibisov:
Draft only because it's done just for wayland.

src/platform/wayland.rs
```diff
@@ -24,10 +24,10 @@ pub trait ActiveEventLoopExtWayland {
     fn is_wayland(&self) -> bool;
 }
 
-impl ActiveEventLoopExtWayland for ActiveEventLoop {
+impl ActiveEventLoopExtWayland for &dyn ActiveEventLoop {
     #[inline]
     fn is_wayland(&self) -> bool {
-        self.p.is_wayland()
+        self.as_any().downcast_ref::<crate::platform_impl::wayland::ActiveEventLoop>().is_some()
```
kchibisov, line 30: In general, though, such APIs should be added to `ActiveEventLoop` with `-> Option<dyn &Extension>`, but for now I just show how you can implement them un-conditionally with the help of casting, etc.

daxpedda (commented):
I think these changes are a great choice for a first step!
LGTM!

kchibisov:
I've pushed X11 for now as well.

daxpedda (approved):
Thanks for doing the legwork!

PR #3828, opened 2024-07-26
android: add `ActiveEventLoopExtAndroid::android_app`

### LINKED ISSUE #3818: Exposing the existing `AndroidApp` in the public API
For my work on accesskit_winit, it would be helpful if winit exposed the existing `AndroidApp` instance in the public API, e.g. through an `android_app` method on `ActiveEventLoopExtAndroid`.

It's true that the top-level application already has the `AndroidApp` instance, since it has to pass that to winit in the first place. But that's not enough for accesskit_winit, for two reasons:

1. Directly passing the `AndroidApp` instance from the application to accesskit_winit would require accesskit_winit to have a platform-specific adapter constructor for Android, and I'm trying to avoid platform-specific things in the accesskit_winit API, since the point of winit is to be cross-platform.
2. Having the application directly pass the `AndroidApp` instance to accesskit_winit would also break the layering that tends to happen in non-trivial applications. Consider that `accesskit_winit::Adapter` is typically instantiated by a cross-platform framework like eframe or masonry.

kchibisov:
This type comes from the user and stored for the entire lifetime, so no need to hide it from them after they've passed it to winit.

Fixes #3818.

--

I can not really test it, but it's rather trivial and should just work, I guess.

@mwcampbell I guess you need something like that?

### DIFF
--- a/src/changelog/unreleased.md
+++ b/src/changelog/unreleased.md
@@ -54,6 +54,7 @@ changelog entry.
   `ActiveEventLoop::request_detailed_monitor_permission()`, access to all monitors and their
   information is available. This "detailed monitors" can be used in `Window::set_fullscreen()` as
   well.
+- On Android, add `{Active,}EventLoopExtAndroid::android_app()` to access the app used to create the loop.
 
 ### Changed
 
--- a/src/platform/android.rs
+++ b/src/platform/android.rs
@@ -75,12 +75,22 @@ use crate::event_loop::{ActiveEventLoop, EventLoop, EventLoopBuilder};
 use crate::window::{Window, WindowAttributes};
 
 /// Additional methods on [`EventLoop`] that are specific to Android.
-pub trait EventLoopExtAndroid {}
+pub trait EventLoopExtAndroid {
+    /// Get the [`AndroidApp`] which was used to create this event loop.
+    fn android_app(&self) -> &AndroidApp;
+}
 
-impl EventLoopExtAndroid for EventLoop {}
+impl EventLoopExtAndroid for EventLoop {
+    fn android_app(&self) -> &AndroidApp {
+        &self.event_loop.android_app
+    }
+}
 
 /// Additional methods on [`ActiveEventLoop`] that are specific to Android.
-pub trait ActiveEventLoopExtAndroid {}
+pub trait ActiveEventLoopExtAndroid {
+    /// Get the [`AndroidApp`] which was used to create this event loop.
+    fn android_app(&self) -> &AndroidApp;
+}
 
 /// Additional methods on [`Window`] that are specific to Android.
 pub trait WindowExtAndroid {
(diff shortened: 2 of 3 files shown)


src/platform/android.rs
```diff
@@ -80,7 +80,10 @@ pub trait EventLoopExtAndroid {}
 impl EventLoopExtAndroid for EventLoop {}
 
 /// Additional methods on [`ActiveEventLoop`] that are specific to Android.
-pub trait ActiveEventLoopExtAndroid {}
+pub trait ActiveEventLoopExtAndroid {
+    /// Get the `AndroidApp` which was used to create an event loop.
```
MarijnS95, line 84: Intradoc link?

src/platform/android.rs
```diff
@@ -80,7 +80,10 @@ pub trait EventLoopExtAndroid {}
 impl EventLoopExtAndroid for EventLoop {}
```
MarijnS95, line 80: Do we need an active event loop or can we implement it on the nonactive one as well? (I don't remember if that's deprecated)

src/platform/android.rs
```diff
@@ -80,7 +80,10 @@ pub trait EventLoopExtAndroid {}
...
 
 /// Additional methods on [`ActiveEventLoop`] that are specific to Android.
-pub trait ActiveEventLoopExtAndroid {}
+pub trait ActiveEventLoopExtAndroid {
+    /// Get the `AndroidApp` which was used to create an event loop.
+    fn android_app(&self) -> &AndroidApp;
```
MarijnS95, line 85: Also unsure if we should document that `AndroidApp` is a cheaply-copyable reference type - that's already in the `android-activity` documentation though.

MarijnS95 (approved):
Also didn't test but looks trivial enough.

mwcampbell:
Yes, that's the API I need. Thanks.

src/platform/android.rs
kchibisov: I mean, users should be able to go to `AndroidApp` docs and they should be aware since they've created it.

src/platform/android.rs
kchibisov: I've added to event loop as well, since there's nothing wrong  with it.

src/platform/android.rs
kchibisov: aye

src/changelog/unreleased.md
```diff
@@ -54,6 +54,7 @@ changelog entry.
   `ActiveEventLoop::request_detailed_monitor_permission()`, access to all monitors and their
   information is available. This "detailed monitors" can be used in `Window::set_fullscreen()` as
   well.
+- On Android, add `{Active,}EventLoopExtAndroid::android_app` to access the app used to create the loop.
```
MarijnS95, line 57: Nit:

```suggestion
- On Android, add `{Active,}EventLoopExtAndroid::android_app()` to access the app used to create the loop.
```

src/platform/android.rs
```diff
@@ -75,12 +75,22 @@ use crate::event_loop::{ActiveEventLoop, EventLoop, EventLoopBuilder};
 use crate::window::{Window, WindowAttributes};
 
 /// Additional methods on [`EventLoop`] that are specific to Android.
-pub trait EventLoopExtAndroid {}
+pub trait EventLoopExtAndroid {
+    /// Get the [`AndroidApp`] which was used to create an event loop.
```
MarijnS95, line 79: Nit: an -> the or this?

MarijnS95 (approved):

src/platform/android.rs
kchibisov: this, I guess.

PR #3829, opened 2024-07-26
Use Taplo for TOML formatting

daxpedda:
We currently don't format our TOML files.
This PR introduces [Taplo](https://taplo.tamasfe.dev) to our CI to do just that.


rustfmt.toml
```diff
@@ -1,20 +1,20 @@
-format_code_in_doc_comments = true
-match_block_trailing_comma = true
+comment_width = 100
```
kchibisov, line 1: This one was actually formatted by length, so it was a nice ladder.

kchibisov (commented):
If it's fast enough, I don't really care. Though, maybe we should just do it manually on-demand? Since it's really annoying to require more and more stuff for contributors to reproduce CI failures locally.

daxpedda:
> If it's fast enough, I don't really care.

Its fast alright.
8s total runtime on CI, 3s for installing and less then 1s for running.

> Though, maybe we should just do it manually on-demand? Since it's really annoying to require more and more stuff for contributors to reproduce CI failures locally.

Doing things manually and on-demand is always more time consuming and annoying in my opinion.
Regular contributors shouldn't be playing around in TOML anyway, so it shouldn't be an issue.

rustfmt.toml
daxpedda: Lol.
I can disable the alphabetical sorting?
But I think that's not a small part of the whole point here.

I prefer alphabetical ordering for ease-of-use over "looking nice" anyway.

kchibisov:
I don't really mind as long as others want stuff like that.

notgull (approved):
I'm a fan

PR #3830, opened 2024-07-26
Move to TypeScript

daxpedda:
This moves our JavaScript code to use TypeScript instead.
In general, TypeScript is much safer to use and code can be much better checked by linters.
The process is a bit more involved though, so I moved all the related files in their own folders to not "pollute" our glorious Rust code.

I know we only have one measly JavaScript file to begin with, but I already found a "bug" (it works correctly in reality), so here we are.

- Moves our JavaScript to TypeScript.
- Setup ESLint.
- Adjusts SWC.


kchibisov:
The amount of JS we have is rather small. IIRC typescript is more intense to get running, since you need to compile it to JS in the end of the day, so the whole pipeline of shipping something gets a bit more complex.

So, do we _really_ need? Like bugs could be found any other day in TS as well.

daxpedda:
> IIRC typescript is more intense to get running, since you need to compile it to JS in the end of the day, so the whole pipeline of shipping something gets a bit more complex.

We actually used SWC for minification before already, its doing the whole transpiling to JS as well now, so actually our pipeline stays the same.
All the additional files are actually not because its TS, but because I added ESLint.

> So, do we _really_ need? Like bugs could be found any other day in TS as well.

We don't _really_ need anything. I could also do the whole compilation and minification offline as well.
This is an improvement, bugs could be found in Rust as well, it doesn't mean we stop using Clippy.

daxpedda:
I'm gonna expand on whats going on here so there is some adequate justification in place.

It all starts at us spawning a [Web Worker](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API). This requires some sort of JavaScript code to get started. No way around that.

The easiest thing would be to just put the code as a string inside a Rust file and call it a day.
There are two issues with that:
- Its gonna be quite hard to minimize that.
- Without any sort of checking, it could be easy to make a mistake. E.g. variable name typos.

### Minification

On Web, size is quite important, because everything has to be transmitted over the wire, the bigger something is, the longer its going to take to load a web site. To that end, the Web ecosystem has a huge ecosystem around code minification, so any JavaScript code is minimized as much as possible before being sent through the wire (this excludes any actual compression, e.g. gzip).

Minification is the process of removing any non-essential things from the code. E.g. whitespace, newlines, variable names (replaced with single characters), unnecessary semi-colons, the list goes on.

In comparison, Desktop doesn't have those issues, more often then not executable size is not really a big concern and there is no need for minification because code is compiled down to machine language anyway.

So the next step would be to put the code into a separate file to be able to run a minimizer tool on top of it, which spits out the minimized code, which we then can use for our Web Worker.

Note: minification is often not something done by libraries, but by end-users. We can't offer that luxury, because we are a Rust library. Our end-users are not consuming our JavaScript file, but is embedded into the binary.

### Correctness

In my humble opinion, JavaScript is a terrible language compared to Rust when it comes to correctness. The language doesn't even have type safety. Its pretty hard to ascertain if your code is entirely correct without at least some tools.

<details>
<summary>Here is an example that originally motivated this PR</summary>

I was refactoring some code in a Rust Wasm threading library and it stopped working. After some digging I found out that the issue was [`this`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this) suddenly didn't refer to [`global`](https://developer.mozilla.org/en-US/docs/WebAssembly/Reference/Variables/Global) anymore, where before it did.

So what went wrong? My code looked something like this:

Before:
```js
addEventListener('message', (event) => {
    this.doSomething()
    ...
}
```
After:
```js
function handler(event) {
    this.doSomething()
}

addEventListener('message', handler)
```

So, probably any half-baked JS programmer could spot the error, but I'm a rookie when it comes to JS. I already knew that using `this` to access `global` might be a bad idea, but I originally did that to save some bytes. Honestly I was surprised it worked at all.
Afterwards I went on a bit of a killing spree and "fixed" all similar code in all my worker scripts.

But why did it work until now?
Apparently `this` works just fine if you use [arrow functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) but not when using regular functions, which rebind `this` to the element calling this function.

For reference: I had to figure this all out without an error message. Unfortunately [Audio Worklets](https://developer.mozilla.org/en-US/docs/Web/API/AudioWorkletNode) are mostly incapable of spitting out any reasonable errors.

</details>

So my lesson from this was: don't write code without at least some machine doing some basic sanity checks for you. In Rust we have the compiler and Clippy. But in JavaScript we have nothing, only when executing the code it is checked to some minor degree before any bugs are revealed during runtime.

Why move to TypeScript? Why not just use a JavaScript linter?
Because its pretty hard to work on JavaScript as a linter, remember, it doesn't even have types!

src/platform_impl/web/script/.gitignore
```diff
@@ -0,0 +1,2 @@
+/node_modules
```
kchibisov, line 1: Could we use toplevel gitignore for that?

src/platform_impl/web/script/worker.min.js
```diff
@@ -0,0 +1 @@
+onmessage=e=>{let[s,a]=e.data,l=()=>{s.postMessage(void 0)};"scheduler"in globalThis?globalThis.scheduler.postTask(l,{delay:a}):setTimeout(l,a)};
```
kchibisov, line 1: new line at the end of file.

.github/dependabot.yaml
```diff
@@ -0,0 +1,11 @@
+version: 2
+
+updates:
+  - package-ecosystem: npm
+    directory: src/platform_impl/web/script
```
kchibisov, line 5: I'd assume this is all for linter?

kchibisov (dismissed):

src/platform_impl/web/script/worker.min.js
daxpedda: Haha, that's the whole point!
Its supposed to save bytes as much as possible, even newlines.

Anyway, this is tool generated.

.github/dependabot.yaml
daxpedda: Yes.

src/platform_impl/web/script/.gitignore
daxpedda: Yes.
The idea was to really not pollute the rest of the codebase with the JavaScript ecosystem.

Would you still prefer it in top-level?

src/platform_impl/web/script/worker.min.js
kchibisov: could you indicate that it's generated?

src/platform_impl/web/script/.gitignore
kchibisov: it's more discoverable at the top one and you can also explain there why/what ignored and why, so yes.

src/platform_impl/web/script/worker.min.js
daxpedda: How would you want me to indicate that?
The indication should be the extension: `.min.js`, it means its minimized, nobody does that by hand.

src/platform_impl/web/script/.gitignore
daxpedda: Sure!

daxpedda: Done.

src/platform_impl/web/script/worker.min.js
kchibisov: ah, so the convention is to name generated with `min`? Then it's fine. I was thinking about `.generated.js`.

.github/workflows/ci.yml
```diff
@@ -243,17 +243,38 @@ jobs:
...
+
     runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./src/platform_impl/web/script
 
     steps:
-      - uses: taiki-e/checkout-action@v1
+      - name: Checkout
+        uses: taiki-e/checkout-action@v1
```
notgull, line 272: No reason to add a name to this action

notgull (approved):
Based

.github/workflows/ci.yml
daxpedda: I actually think all actions should have names to make it easier read the CI.
But it would be inconsistent with our current setup!

Done!

PR #3831, opened 2024-07-26
Remove `raw-window-handle` v0.4 and v0.5 support

daxpedda:
As discussed before (somewhere), this PR removes `raw-window-handle` v0.4 and v0.5 support.
Also renames the `rwh_06` feature to `raw-window-handle`.

We could also remove the whole crate feature now that we support only one version?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,8 +15,6 @@ version = "0.30.4"
 
 [package.metadata.docs.rs]
 features = [
-    "rwh_04",
-    "rwh_05",
     "rwh_06",
     "serde",
     "mint",
@@ -48,8 +46,6 @@ android-game-activity = ["android-activity/game-activity"]
 android-native-activity = ["android-activity/native-activity"]
 default = ["rwh_06", "x11", "wayland", "wayland-dlopen", "wayland-csd-adwaita"]
 mint = ["dpi/mint"]
-rwh_04 = ["dep:rwh_04", "ndk/rwh_04"]
-rwh_05 = ["dep:rwh_05", "ndk/rwh_05"]
 rwh_06 = ["dep:rwh_06", "ndk/rwh_06"]
 serde = ["dep:serde", "cursor-icon/serde", "smol_str/serde", "dpi/serde"]
 wayland = [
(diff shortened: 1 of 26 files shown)


kchibisov:
yeah, let's just remove the features for now and unconditionally require it.

daxpedda:
Done.

notgull (changes requested):
I disagree; I see to keep the feature, that way we can seamlessly add `raw-window-handle` v0.7 support if we decide to release it.

daxpedda:
> I disagree; I see to keep the feature, that way we can seamlessly add `raw-window-handle` v0.7 support if we decide to release it.

I think this begs the question: at what point can we stop having to be prepared for the next version?
Probably with `raw-window-handle` v1 I would say.

On the other hand I don't really believe this is necessary if we actually released Winit versions more often like we decided in the past.

If I *had* to go one way or another, I would probably agree to keeping the feature, so I will go ahead and do just that.

notgull (approved):

notgull:
Once `raw-window-handle` v1.0 releases, there won't be a v2.0.

daxpedda:
> Once `raw-window-handle` v1.0 releases, there won't be a v2.0.

Haha, never say never!

@kchibisov and @MarijnS95 would like your approval here as well.

Cargo.toml
```diff
@@ -92,7 +84,7 @@ tracing = { version = "0.1.40", default-features = false }
 image = { version = "0.25.0", default-features = false, features = ["png"] }
 tracing = { version = "0.1.40", default-features = false, features = ["log"] }
 tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
-winit = { path = ".", features = ["rwh_05"] }
+winit = { path = "." }
```
MarijnS95, line 87: Isn't the current (lib) crate always part of `[dev-dependencies]` by default? It looks like it was only added here to force the `rwh_06` feature.

```suggestion
```

MarijnS95 (approved):
Makes sense. I wonder if there is "some place" where we should mention that `rwh_06` is still a feature and that it _was already_ the default?

Cargo.toml
daxpedda: Ah, right!

daxpedda:
> Makes sense. I wonder if there is "some place" where we should mention that `rwh_06` is still a feature and that it _was already_ the default?

Good point!
Yes, I will mention it alongside the entry in the changelog

Cargo.toml
daxpedda: It did have some use after all, enabling default dependencies.
But I addressed this by using the [`required-features` field](https://doc.rust-lang.org/1.80.0/cargo/reference/cargo-targets.html#the-required-features-field).

MarijnS95: Ah, good catch! `requires-features` is a nicer one (though I do sometimes with Rust had a `cargo run --with-required-features --example ...` to simplify).

MarijnS95 (approved):

kchibisov (approved):

PR #3834, opened 2024-07-27
Add GitHub Actions to Dependabot

daxpedda:
While we are at it, add GitHub Actions to Dependabot.
I didn't add Cargo because we didn't reach consensus on that in the past.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,6 +1,15 @@
 version: 2
 
 updates:
+  - package-ecosystem: github-actions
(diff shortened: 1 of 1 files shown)


notgull:
Last I checked Dependabot for GitHub Actions was buggy, see https://github.com/taiki-e/install-action/issues/43

notgull:
Ah wait, it's been fixed, see https://github.com/dependabot/dependabot-core/pull/6743

notgull (approved):

PR #3838, opened 2024-07-29
Add `ActiveEventLoop::system_theme()`

### LINKED ISSUE #3837: System Theme vs Window Theme
In #3736 the meaning of `Window::theme()` was changed to not return the system theme, ergo returning `None` without a theme override through `Window::set_theme()`.
This is how currently MacOS works, but not how Web and Windows do.

I propose the following change:
- The meaning of `Window::theme()` should be changed to return the current window theme, if no override is in place it returns the system theme. I think a use-case for only returning an override has yet to be observed.
- MacOS should return the system theme in `Window::theme()` if no theme override is in place.
- A new `ActiveEventLoop::system_theme()` method should be introduced that returns the system theme.
- (optional) A new `Window::system_theme()` convenience method should be added that returns the same as `ActiveEventLoop::system_theme()`.

Cc @madsmtm.

daxpedda:
In #3736 the definition of `Window::theme()` was changed to only return theme overrides and never the system theme. Which was only how Wayland worked. MacOS was changed to follow this.

However Web and Windows, and MacOS before #3736, currently do return the system theme in `Window::theme()` when no theme override is set. The changed definition and behavior currently does not make it possible to get the system theme, which leaves a gap in the API that was previously covered.

This PR follows the proposal set in #3837 that reintroduces the old behavior and adds a new `ActiveEventLoop::system_theme()` to retrieve the system theme without any window overrides.

- MacOS is restored to its old behavior, returning the system theme when no override is set (minus the bugs fixed in #3736).
- Wayland, no changes, return `None` without a theme override because it currently doesn't implement querying the system theme.

Follow-up to #3736.
Fixes #3837.

### DIFF
--- a/src/window.rs
+++ b/src/window.rs
@@ -1367,14 +1367,31 @@ impl Window {
         self.window.maybe_queue_on_main(move |w| w.set_theme(theme))
     }
 
-    /// Returns the current window theme override.
+    /// Returns the current system theme.
     ///
-    /// Returns `None` if the current theme is set as the system default, or if it cannot be
-    /// determined on the current platform.
+    /// Returns `None` if it cannot be determined on the current platform.
+    ///
+    /// This is the same as [`ActiveEventLoop::system_theme()`], and is provided for convenience.
     ///
     /// ## Platform-specific
     ///
-    /// - **iOS / Android / Wayland / x11 / Orbital:** Unsupported, returns `None`.
+    /// - **iOS / Android / Wayland / x11 / Orbital:** Unsupported.
+    ///
+    /// [`ActiveEventLoop::system_theme()`]: crate::event_loop::ActiveEventLoop::system_theme
+    #[inline]
+    pub fn system_theme(&self) -> Option<Theme> {
+        let _span = tracing::debug_span!("winit::Window::system_theme").entered();
+        self.window.maybe_wait_on_main(|w| w.system_theme())
+    }
+
+    /// Returns the current window theme.
+    ///
+    /// Returns `None` if it cannot be determined on the current platform.
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **iOS / Android / x11 / Orbital:** Unsupported.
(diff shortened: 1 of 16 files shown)


daxpedda:
Considering that `WindowEvent::ThemeChanged` only reports system theme changes, which was clarified in #3736 as well (unless you can change the window theme externally?), I would argue that the event should be moved to a global `Event` and not be `Window`-specific anymore.

src/window.rs
```diff
@@ -1367,14 +1367,31 @@ impl Window {
...
+    ///
+    /// This is the same as [`ActiveEventLoop::system_theme()`], and is provided for convenience.
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **iOS / Android / Wayland / x11 / Orbital:** Unsupported.
+    ///
+    /// [`ActiveEventLoop::system_theme()`]: crate::event_loop::ActiveEventLoop::system_theme
+    #[inline]
+    pub fn system_theme(&self) -> Option<Theme> {
```
kchibisov, line 1382: Do we really need it on the window? In general, if the system theme is tied to event loop, I'd keep it there, it's not like you have to poll it very often etc, and it would require changes internally to get it on the window, well, at least, sometimes.

daxpedda: I can definitely imagine it being useful. Right now it comes with no downsides, but yes, if it requires further changes internally then lets just remove it when it comes to that.

Would advocate to remove it if it makes problems in an implementation or when we make `Window` not `Send + Sync` anymore, but keep it for now.

daxpedda: > We can always add it if there's a desire for it, but I'd not add _more_ without a request.

That's pretty convincing to me.

daxpedda: Done.

kchibisov (approved):

madsmtm (commented):
Posthumously approving 👍

PR #3839, opened 2024-07-29
Android: set `KeyEvent::text` if possible

podusowski:
Related discussion: https://github.com/rust-windowing/winit/pull/3787#issuecomment-2229058082

### DIFF
--- a/src/changelog/unreleased.md
+++ b/src/changelog/unreleased.md
@@ -83,6 +83,7 @@ changelog entry.
   instead of requiring a `&mut` reference to it.
 - On Web, `Window::canvas()` now returns a reference.
 - On Web, `CursorGrabMode::Locked` now lets `DeviceEvent::MouseMotion` return raw data, not OS
+  accelerated, if the browser supports it.
 - On Android, when `keycode` can be represented as character, it is used as `KeyEvent::text`.
 
 ### Removed
--- a/src/changelog/unreleased.md
+++ b/src/changelog/unreleased.md
@@ -83,7 +83,8 @@ changelog entry.
   instead of requiring a `&mut` reference to it.
 - On Web, `Window::canvas()` now returns a reference.
 - On Web, `CursorGrabMode::Locked` now lets `DeviceEvent::MouseMotion` return raw data, not OS
-  accelerated, if the browser supports it.
+- On Android, when `keycode` can be represented as character, it is used as
+  `KeyboardInput::text`.
 
 ### Removed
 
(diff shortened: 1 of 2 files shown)


src/platform_impl/android/mod.rs
```diff
@@ -399,7 +407,7 @@ impl EventLoop {
                                 logical_key: keycodes::to_logical(key_char, keycode),
```
MarijnS95, line 407: DRY:

```suggestion
                                logical_key,
```

src/changelog/unreleased.md
```diff
@@ -84,6 +84,7 @@ changelog entry.
 - On Web, `Window::canvas()` now returns a reference.
 - On Web, `CursorGrabMode::Locked` now lets `DeviceEvent::MouseMotion` return raw data, not OS
   accelerated, if the browser supports it.
+- On Android, when `keycode` can be represented as character, it is used as `KeyEvent::text`.
```
MarijnS95, line 87:
```suggestion
- On Android, when `keycode` can be represented as character, it is provided in `KeyEvent::text`.
```

Or even:

```suggestion
- On Android, when `keycode` can be represented as character, its textual representation is provided in `KeyEvent::text`.
```

Should we also mention that it applies to logical and not physical keys, in case a mapping is active?

podusowski: I'm not that familiar with this stuff, so I need to count on you on this. :)

MarijnS95: Me neither, which is why I expect the other maintainers to chime in and confirm that this is a sensible change to make.

It's always a bit scary when neither contributor nor reviewer have a definitive answer :)

Zoxc: That seems redundant with the `KeyEvent` documentation?

Zoxc:
This change does make some text input with the software keyboard possible with egui. Android seems to emulate physical keys  when a proper IME connection is missing.

src/platform_impl/android/mod.rs
```diff
@@ -468,15 +469,22 @@ impl EventLoop {
                             &mut self.combining_accent,
                         );
 
+                        let logical_key = keycodes::to_logical(key_char, keycode);
+                        let text = if let Key::Character(ref c) = logical_key {
+                            Some(c.to_owned())
+                        } else {
+                            None
+                        };
```
MarijnS95, line 477: Let's change this to:
```suggestion
                        let text = if state == event::ElementState::Pressed {
                            logical_key.to_text().map(SmolStr::new)
                        } else {
                            None
                        };
```

Which is the abstraction that Windows and AppKit also use to _not only_ convert characters to their textual representation but also keys like `NamedKey::Enter` to `\r`, `NamedKey::Tab` to `\t` andsoforth.


And supposedly this should stay `None` if the key is released, too.

PR #3844, opened 2024-08-02
Make `(Active)EventLoop::create_custom_cursor()` return a `Result<CustomCursor, ExternalError>`

daxpedda:
This PR changes `(Active)EventLoop::create_custom_cursor()` to return a `Result<CustomCursor, ExternalError>`.

- Android, iOS and Orbital now return `ExternalError::Unsupported`.
- MacOS, Windows and X11 can now return an error instead of panicking or doing nothing.
- Wayland doesn't implement custom cursor like other platforms (yet?) and actually creates it in `set_custom_cursor()` instead, so it currently panics there.
- Web will continue to do nothing, because errors are delivered asynchronously. However it already had `create_custom_cursor_async()` to properly return an error.

Related #3306.
Follow-up to #3749.

### DIFF
--- a/src/changelog/unreleased.md
+++ b/src/changelog/unreleased.md
@@ -85,6 +85,7 @@ changelog entry.
 - On Web, `Window::canvas()` now returns a reference.
 - On Web, `CursorGrabMode::Locked` now lets `DeviceEvent::MouseMotion` return raw data, not OS
   accelerated, if the browser supports it.
+- `(Active)EventLoop::create_custom_cursor()` now returns a `Result<CustomCursor, ExternalError>`.
 
 ### Removed
 
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -15,8 +15,7 @@ use tracing::{debug, trace, warn};
 use crate::application::ApplicationHandler;
 use crate::cursor::Cursor;
 use crate::dpi::{PhysicalPosition, PhysicalSize, Position, Size};
-use crate::error;
-use crate::error::EventLoopError;
+use crate::error::{self, EventLoopError, ExternalError, NotSupportedError};
 use crate::event::{self, Force, InnerSizeWriter, StartCause};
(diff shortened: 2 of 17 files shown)


src/changelog/unreleased.md
```diff
@@ -84,6 +84,7 @@ changelog entry.
 - On Web, `Window::canvas()` now returns a reference.
 - On Web, `CursorGrabMode::Locked` now lets `DeviceEvent::MouseMotion` return raw data, not OS
   accelerated, if the browser supports it.
+- `(Active)EventLoop::create_custom_error()` now returns a `Result<CustomCursor, ExternalError>`.
```
MarijnS95, line 87: Here & in PR title:

```suggestion
- `(Active)EventLoop::create_custom_cursor()` now returns a `Result<CustomCursor, ExternalError>`.
```

Perhaps?

daxpedda: Copy pasta spaghetti.
Thank you!

src/platform_impl/android/mod.rs
```diff
@@ -591,9 +590,11 @@ impl ActiveEventLoop {
...
     }
 
-    pub fn create_custom_cursor(&self, source: CustomCursorSource) -> CustomCursor {
-        let _ = source.inner;
-        CustomCursor { inner: PlatformCustomCursor }
+    pub fn create_custom_cursor(
+        &self,
+        _source: CustomCursorSource,
+    ) -> Result<CustomCursor, ExternalError> {
+        Err(ExternalError::NotSupported(NotSupportedError::new()))
```
MarijnS95, line 597: In the future, if we do more JNI wrapping, we can provide custom `Bitmap`s in the future using https://developer.android.com/reference/android/view/PointerIcon. Only when the user attaches a mouse to their Android device :)

MarijnS95 (approved):
Approved for Android.

kchibisov (approved):
Should be fine, though, I'd probably change the type of error we return in the future.

PR #3847, opened 2024-08-04
Web: monitor API improvements

daxpedda:
- Improved the documentation to point users into the right direction from all kinds of methods and types.
- De-duplicated some code and added more comments.
- Implement an ID system to correctly and efficiently implement `Eq`, `Hash`, `Ord`, `PartialEq` and `PartialOrd`.
- Fixed screen locking support being cached thread local, ergo calling from a different thread would require to make the check again, not fulfilling its purpose as a fast-path at all.

### DIFF
--- a/src/monitor.rs
+++ b/src/monitor.rs
@@ -98,6 +98,23 @@ impl std::fmt::Display for VideoModeHandle {
 ///
 /// Allows you to retrieve information about a given monitor and can be used in [`Window`] creation.
 ///
+/// ## Platform-specific
+///
+/// **Web:** A [`MonitorHandle`] created without
+#[cfg_attr(
+    any(web_platform, docsrs),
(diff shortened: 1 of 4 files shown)


kchibisov (approved):

daxpedda:
I did some thinking about the API complexity of this implementation.
There are basically two different `MonitorHandle` types, but they are the same Rust type. Users can query which type it is with `is_detailed()`, but that's all not exactly ideal.

I think what could be done about that in the future is to split up the orientation and screen locking stuff away from `MonitorHandle` entirely, that way there won't be a use for `MonitorHandle` on Web at all unless with "detailed information". However this requires a new API surface that we have yet to think about and implement, which requires some cross-platform coordination.

In any case, its some food for thought.

PR #3849, opened 2024-08-07
Winit version 0.30.5

kchibisov:


ArthurBrussee:
Hiya! Would it by chance be possible to include https://github.com/rust-windowing/winit/commit/15b79b18e1f6b64ec21a0027edbc84868bd19c31 as well? That's blocking egui from going past 0.30.2 as it can't read the current system theme since 0.30.3.

Don't know what the process is but figured I'd ask, thank you :)

kchibisov:
I've excluded it because it changed behavior a bit, but given that old behavior was not entirely correct it's probably fine to add it.

ArthurBrussee:
Perfect, thanks a bunch!

daxpedda (approved):

PR #3851, opened 2024-08-08
api: add `ApplicationHandler::as_any`

kchibisov:
This could be used to subscribe for custom events.

### DIFF
--- a/src/application.rs
+++ b/src/application.rs
@@ -1,5 +1,7 @@
 //! End user application handling.
 
+use std::any::Any;
+
 use crate::event::{DeviceEvent, DeviceId, StartCause, WindowEvent};
 use crate::event_loop::ActiveEventLoop;
 use crate::window::WindowId;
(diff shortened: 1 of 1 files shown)


kchibisov:
Example on how to use it https://github.com/rust-windowing/winit/pull/3852

src/application.rs
```diff
@@ -325,6 +327,14 @@ pub trait ApplicationHandler {
...
         let _ = event_loop;
     }
+
+    /// Get the [`ApplicationHandler`] as mutable [`Any`].
+    ///
+    /// This is required to be implemented when platform specific traits are desired.
+    #[inline(always)]
+    fn as_any(&mut self) -> Option<&mut dyn Any> {
+        None
+    }
```
daxpedda, line 337: I played around a bit to figure out how we can get rid of this:
https://github.com/daxpedda/winit/commit/374014880b0606af9796a0ecfaff00f31a3312e5

Mostly inspired by [`upcast`](https://crates.io/crates/upcast).
I'm warming up to the VTable idea, but I couldn't find any documentation on how stable this is, so I'm unsure what the proposal around this was exactly.

kchibisov: >I'm warming up to the VTable idea, but I couldn't find any documentation on how stable this is, so I'm unsure what the proposal around this was exactly.

vtable is only for completely out of tree stuff, for regular staff you don't need vtables, since users will have `Option<dyn Handler>` and you can just `if let` it. But function pointers are completely safe and stable, there's nothing wrong with them and they are all completely safe code.

it's just in the current state of things I decided with vtable, since it's a bit more reliable, but we can already do other approach actually.

So, we'd have `-> Option<&mut dyn MacOSApplicationHandler>` from the `ApplicationHandler` itself.

daxpedda: Yeah okay, I was thinking of totally different things.
I would like to avoid having to use `register_x()` entirely. That is possible by *unsafely* reading the VTable and figuring out at runtime if a trait is implemented for a certain type.

kchibisov:
This is already present on master.

PR #3854, opened 2024-08-11
Properly implement event loop extension traits

madsmtm:
This was implemented for `&dyn ActiveEventLoop` in https://github.com/rust-windowing/winit/pull/3827, but that's less general than implementing it directly for `dyn ActiveEventLoop`.

The `+ '_` is required to tell the compiler that we want to implement this for all `dyn ActiveEventLoop`s, not just `'static` ones (even though they're all going to be `'static`).

This is especially confusing because the requirement to add `+ '_` on `dyn Trait` types differs depending on the place where `dyn Trait` is used, see [the reference](https://doc.rust-lang.org/reference/lifetime-elision.html#default-trait-object-lifetimes).

  - No entry added, as this isn't yet released

### DIFF
--- a/src/platform/android.rs
+++ b/src/platform/android.rs
@@ -109,7 +109,7 @@ impl WindowExtAndroid for Window {
     }
 }
 
-impl ActiveEventLoopExtAndroid for &dyn ActiveEventLoop {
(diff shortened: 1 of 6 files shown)


kchibisov (approved):
Impressive.

PR #3855, opened 2024-08-11
chore: use our own AsAny<T> upcasting

kchibisov:
This removes the direct requirement to implement `as_any` and it could be just derived.

Also implement HasDisplayHandle for dyn ActiveEventLoop + '_.

Suggested-by: @daxpedda

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -4,6 +4,7 @@
 // This isn't used on every platform, which can come up as dead code warnings.
 #![allow(dead_code)]
 
+use std::any::Any;
 use std::ops::Deref;
 use std::sync::OnceLock;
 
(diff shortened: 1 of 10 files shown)


src/utils.rs
```diff
@@ -26,3 +27,36 @@ impl<T> Deref for Lazy<T> {
         self.cell.get_or_init(self.init)
     }
 }
+
+trait UpcastFrom<T: ?Sized> {
+    fn any_from(value: &T) -> &Self;
+    fn any_mut_from(value: &mut T) -> &mut Self;
+}
+
+pub trait AsAny<U: ?Sized> {
```
madsmtm, line 36: Is this intended to be public API?

kchibisov: you can not require private `trait` on `pub trait`. But you can not actually do anything with it.

In the future I think this particular trait could be all pub.

madsmtm, line 36: Why the generic parameter `U` here, if we only ever use it as `AsAny<dyn Any>`?

kchibisov: You can not bound by trait object though.

madsmtm: I think my question is: Why not just
```rust
pub trait AsAny {
    fn as_any(&self) -> &dyn Any;
    fn as_any_mut(&mut self) -> &mut dyn Any;
}

impl<T: Any> AsAny for T {
    fn as_any(&self) -> &dyn Any {
        self
    }

    fn as_any_mut(&mut self) -> &mut dyn Any {
        self
    }
}
```

kchibisov: Hm, indeed.

madsmtm (approved):

PR #3856, opened 2024-08-11
Make `ActiveEventLoop` conditionally subtrait `HasDisplayHandle`

madsmtm:
Using a hacky macro that wraps the definition of `ActiveEventLoop`.

This allows getting rid of the `rwh_06_handle` method.

In the future, once a new version of `raw-window-handle` is released that we want to support, the macro could be extended (though it'll require 4 macros to support all 4 feature combinations, and 8 if there's 3 `raw-window-handle` versions).

### DIFF
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -276,7 +276,7 @@ impl EventLoop {
 #[cfg(feature = "rwh_06")]
 impl rwh_06::HasDisplayHandle for EventLoop {
     fn display_handle(&self) -> Result<rwh_06::DisplayHandle<'_>, rwh_06::HandleError> {
-        rwh_06::HasDisplayHandle::display_handle(self.event_loop.window_target().rwh_06_handle())
+        rwh_06::HasDisplayHandle::display_handle(self.event_loop.window_target())
     }
 }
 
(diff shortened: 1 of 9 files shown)


src/event_loop.rs
```diff
@@ -308,110 +308,137 @@ impl AsRawFd for EventLoop {
...
-    ///
-    /// This is useful for downcasting to a concrete event loop type.
-    fn as_any(&self) -> &dyn Any;
-
-    /// Get the raw-window-handle handle.
-    #[cfg(feature = "rwh_06")]
-    fn rwh_06_handle(&self) -> &dyn rwh_06::HasDisplayHandle;
+// cfg-gate whether the trait is a supertrait
+#[cfg(feature = "rwh_06")]
+macro_rules! cfg_rwh_supertrait {
```
daxpedda, line 313: I think the comment here could be extended a bit saying why we are doing this.

src/event_loop.rs
```diff
@@ -308,110 +308,137 @@ impl AsRawFd for EventLoop {
...
+}
+
+cfg_rwh_supertrait!(
+    /// Target that associates windows with an [`EventLoop`].
+    ///
+    /// This type exists to allow you to create new windows while Winit executes
+    /// your callback.
+    ///
+    /// This implements `HasDisplayHandle` when the corresponding `rwh_xx` feature is enabled.
+    pub trait ActiveEventLoop: rwh_06::HasDisplayHandle {
```
daxpedda, line 340: You are filtering out the `HasDisplayHandle` trait in the macro, but if we have multiple traits listed here, how are you gonna filter them out then?

I would have otherwise proposed to add them in the macro instead of filtering them out.

daxpedda (approved):
I only see upsides here!

daxpedda:
I guess we implementing it for `dyn ActiveEventLoop` works as well? See #3855.
Will close this then.

kchibisov:
>I guess we implementing it for dyn ActiveEventLoop works as well? See https://github.com/rust-windowing/winit/pull/3855.
Will close this then.

yes, there's no difference and I'd rather define our own trait based on feature combination of `rwh` instead of having `macro`.

Issue #3690, opened 2024-05-08
Semantics of getters/setters and events?


madsmtm:
I'm working on improving the internals of the event handling and queuing on macOS (and iOS), and couldn't seem to find any docs on what the expected behaviour actually is? Take the following example of modifying and querying the outer position of a window:

```rust
match event {
    WindowEvent::Moved(new_position) => {
        println!("moved to: ({}, {})", new_position.x, new_position.y);
    },
    WindowEvent::KeyboardInput { event: KeyEvent { state: ElementState::Pressed, .. }, .. } => {
        println!("setting position");
        window.set_outer_position(PhysicalPosition::new(100, 100));
        let new_position = window.outer_position().unwrap();
        println!("new position: ({}, {})", new_position.x, new_position.y);
    },
    _ => (),
}
```

There are a few possible ways this could behave:
1. The value is immediately updated, and querying it yields the new value:
   ```
   setting position
   new position: (100, 100)
   moved to: (100, 100)
   ```
2. The value is deferred to only be updated once the event executes:
   ```
   setting position
   new position: (712, 234) // Old value
   moved to: (100, 100)
   ```
3. The event is executed immediately, and the new value is available afterwards:
   ```
   setting position
   moved to: (100, 100) // Emitted before window.set_outer_position returns
   new position: (100, 100)
   ```

Of these, I would argue that option 3 is the "better" option, as it allows the user to have one place where updates to the position is always executed, and every part of the system instantly remain up to date about the current state. It is, however, also not possible in Rust with the current event handler, since it'd be re-entrant (and is probably also pretty difficult to implement efficiently).

That leaves us with option 1 and 2, of which I'm honestly not sure which is the best one? I can see the arguments for both sides, with option 1 ensuring consistency within the section, while option 2 ensures consistency within the entire application (forcing you to do stuff that depends on a specific update in that event handler).

Currently though, I'm leaning towards option 1 as that's how the macOS backend currently works (mostly), and I think that's what users expect when they call a method called `set_x`.

Is that how it works on the other platforms too? And for all the other methods? I see that we have `request_inner_size`, I think because it works like option 2 on some platforms (X11? Windows?), can and should we change that to be more consistent?

daxpedda:
> Is that how it works on the other platforms too? And for all the other methods?

Not for Web, where any getter method will always return the current true value regardless of what the last thing is the user did. Though more often then not this turns out to look like 1. anyway because on Web most things are updated immediately.

[...]

PR #3861, opened 2024-08-13
Web: return `MonitorHandle` in `Window::fullscreen()`

daxpedda:
- Return `MonitorHandle` in `Window::fullscreen()`
- Change `OrientationData::natural` type from `bool` to `Orientation`, just telling the user what the natural orientation is instead of letting the user "do the math".
- Fix and improve some monitor related documentation.

Follow-up to #3801.

### DIFF
--- a/src/platform_impl/web/window.rs
+++ b/src/platform_impl/web/window.rs
@@ -312,7 +312,7 @@ impl Inner {
     #[inline]
     pub(crate) fn fullscreen(&self) -> Option<Fullscreen> {
         if self.canvas.is_fullscreen() {
-            Some(Fullscreen::Borderless(None))
+            Some(Fullscreen::Borderless(Some(self.monitor.current_monitor())))
         } else {
             None
         }
--- a/src/changelog/unreleased.md
+++ b/src/changelog/unreleased.md
@@ -51,8 +51,11 @@ changelog entry.
   Without prompting the user for permission, only the current monitor is returned. But when
   prompting and being granted permission through
   `ActiveEventLoop::request_detailed_monitor_permission()`, access to all monitors and their
-  information is available. This "detailed monitors" can be used in `Window::set_fullscreen()` as
-  well.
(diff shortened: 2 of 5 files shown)


src/platform_impl/web/window.rs
```diff
@@ -312,7 +312,7 @@ impl Inner {
     #[inline]
     pub(crate) fn fullscreen(&self) -> Option<Fullscreen> {
         if self.canvas.is_fullscreen() {
-            Some(Fullscreen::Borderless(None))
+            Some(Fullscreen::Borderless(Some(self.monitor.current_monitor())))
```
madsmtm, line 315: Not sure if this would necessarily be better? Isn't the "correct" thing to do here to track which state the user originally requested to fullscreen in (custom monitor chosen, or no monitor chosen), and then return the same here? (Such that the assertion in the sequence `window.set_fullscreen(fullscreen); assert_eq!(window.fullscreen(), fullscreen)` would hold).

madsmtm (approved):

src/platform_impl/web/window.rs
daxpedda: I was unsure as well how this would play out, but after some testing I've learned that the current screen is always the one requested to be fullscreen, regardless of what it was before.

So yes, your assertion would hold.

madsmtm: I mean, if I set `let fullscreen = Fullscreen::Borderless(None)`, the assertion no longer holds.

daxpedda: Ah, I see!
No, in that case it won't.

I'm not entirely sure what we should do here exactly, but I'm actually in favor of the current way Web works: always return the currently correct values, regardless of user input.
E.g. what do I do if `set_fullscreen()` fails? Should the assert hold until I receive the error?

This is all less then ideal and should be documented thoroughly ... so I'm gonna leave a comment in https://github.com/rust-windowing/winit/issues/3690.

PR #3864, opened 2024-08-13
Remove `dummy()` from `WindowId` and `DeviceId` and use `Option<DeviceId>`

daxpedda:
This was discussed in the last meeting but it was so late only @kchibisov and myself were around so I'm gonna re-iterate what we discussed here.

`WindowId`, `DeviceId` and the newly added `FingerId`, have only one purpose: making sure that users are able to differentiate between events coming from different devices.
`dummy()` however entirely circumvents this property, because it is *always equal itself*. Ideally `dummy()` would never be equal itself, but that's not possible while we want to implement `Hash` (unless we add a counter ...).
This is in a similar vein to #3795.

However, this assumption was made under the fact hat adding `Option<DeviceId>` was not necessary because only Web is using `dummy()` internally. Turned out this isn't exactly true: *every* backend has at least some events where it is unable to emit a unique `DeviceId` for a device. Some backends, like Wayland and Orbital, don't use any `DeviceId` at all, making all devices appear equal!

So with that in mind, I also replaced all `DeviceId`s with `Option<DeviceId>`.
Surprisingly enough this wasn't necessary for `FingerId` because *every* backend that implements touch emits a proper finger ID.

Based on #3795.

### DIFF
--- a/src/platform_impl/apple/appkit/mod.rs
+++ b/src/platform_impl/apple/appkit/mod.rs
@@ -24,26 +24,17 @@ pub(crate) use self::monitor::{MonitorHandle, VideoModeHandle};
 pub(crate) use self::window::{Window, WindowId};
 pub(crate) use self::window_delegate::PlatformSpecificWindowAttributes;
 pub(crate) use crate::cursor::OnlyCursorImageSource as PlatformCustomCursorSource;
-use crate::event::DeviceId as RootDeviceId;
 pub(crate) use crate::icon::NoIcon as PlatformIcon;
 pub(crate) use crate::platform_impl::Fullscreen;
 
 #[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct DeviceId;
 
-impl DeviceId {
-    pub const fn dummy() -> Self {
-        DeviceId
-    }
-}
-
-// Constant device ID; to be removed when if backend is updated to report real device IDs.
-pub(crate) const DEVICE_ID: RootDeviceId = RootDeviceId(DeviceId);
-
 #[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct FingerId;
 
 impl FingerId {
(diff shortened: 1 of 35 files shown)


madsmtm (commented):
I think `dummy()` on `WindowId` was mostly useful back when we had `Event::WindowEvent(WindowId, WindowEvent)`, and users wanted to create `Event` for testing.

src/platform_impl/apple/appkit/mod.rs
```diff
@@ -27,26 +27,24 @@ pub(crate) use self::monitor::{MonitorHandle, VideoModeHandle};
...
-use crate::event::DeviceId as RootDeviceId;
 pub(crate) use crate::icon::NoIcon as PlatformIcon;
 pub(crate) use crate::platform_impl::Fullscreen;
 
 #[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct DeviceId;
 
 impl DeviceId {
+    #[cfg(test)]
     pub const fn dummy() -> Self {
```
madsmtm, line 38: Still need to remove this and other `fn dummy()`.

Or do we use it for testing anywhere that I don't know about?

daxpedda: We have some tests that check if things don't panic:
https://github.com/rust-windowing/winit/blob/7fbc2118b6b395cd8028f82a5cc8b1563cc86076/src/event.rs#L1166-L1173

But it will be removed eventually when we split the backends out.

madsmtm: I don't see that much value in that test compared to the many extra functions that we need to make it work - but I don't have a strong preference either way here.

daxpedda: I feel the same and would be in favor of removing it.
Unless I hear disagreement I'm gonna go ahead with that when the PR is ready.

kchibisov (approved):
Brought it up to date, should be good now.

notgull (approved):

PR #3865, opened 2024-08-14
iOS: Refactor event handling to share code with macOS

madsmtm:
Instead of storing the event handler within the AppState, and extracting it our every time we need it, we now use the same event handling implementation as for macOS that ensures we don't re-entrantly call the event handler, and that we un-register the handler again after we're done using it (`UIApplicationMain` won't return, but may still unwind, so this is very important for panic safety).

This is the first step of me trying to untangle the inscrutable mess that is the current event handling on iOS.

### DIFF
--- a/src/platform_impl/apple/uikit/app_state.rs
+++ b/src/platform_impl/apple/uikit/app_state.rs
@@ -1,12 +1,12 @@
 #![deny(unused_results)]
 
-use std::cell::{RefCell, RefMut};
+use std::cell::{OnceCell, RefCell, RefMut};
 use std::collections::HashSet;
 use std::os::raw::c_void;
-use std::sync::atomic::AtomicBool;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Mutex, OnceLock};
 use std::time::Instant;
-use std::{fmt, mem, ptr};
+use std::{mem, ptr};
 
 use core_foundation::base::CFRelease;
 use core_foundation::date::CFAbsoluteTimeGetCurrent;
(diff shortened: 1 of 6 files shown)


src/platform_impl/apple/uikit/app_state.rs
```diff
@@ -42,25 +44,45 @@ macro_rules! bug_assert {
...
+        }
     }
+
+    unsafe impl<T> Send for StaticMainThreadBound<T> {}
+    unsafe impl<T> Sync for StaticMainThreadBound<T> {}
+
+    // SAFETY: Creating `StaticMainThreadBound` in a `const` context, where there is no concept
+    // of the main thread.
+    static GLOBAL: StaticMainThreadBound<OnceCell<EventHandler>> =
+        StaticMainThreadBound(OnceCell::new());
```
daxpedda, line 67: Remembering our talk about not storing anything in static memory.
No clue what the story here is exactly.

daxpedda (approved):

src/platform_impl/apple/uikit/app_state.rs
madsmtm: I think I would still like to achieve that at _some_ point, but it's going to be quite difficult with the event handler (especially so on AppKit, where we have to submit events from inside `-[NSApplication sendEvent:]`, without the ability to store anything there, since I plan to swizzle (replace) the method to avoid requiring `WinitApplication` being registered as the main application).

The `EventHandler` is effectively moved from `AppState`, which is already stored in a `static`, so this PR doesn't really change the status quo (we just have two smaller statics now, instead of one big one).

PR #3866, opened 2024-08-14
Remove `PartialOrd` and `Ord` implementations from `WindowId`, `DeviceId`, `MonitorHandle` and   `VideoModeHandle`

daxpedda:
As discussed in the last meeting, this PR removes `PartialOrd`/`Ord` from:
- `WindowId`
- `DeviceId`
- `FingerId`
- `MonitorHandle`
- `VideoModeHandle`

The reason being that `Ord` is supposed to tell users something meaningful, which it doesn't in this case. E.g. what does comparing the order of two `DeviceId`s mean?

On `VideoModeHandle` there was actually some useful implementation, but it wasn't documented and I don't actually believe it makes sense to compare them like this. We might want to consider implementing `Ord` again on `VideoModeData`.

`FingerId` would have been useful to compare with `Ord` *if* the ID actually signifies the finger used in order, in which case we should expose that information directly to the user as well. It isn't clear yet if every backend functions this way, something to ponder about when we overhaul touch events.

### DIFF
--- a/src/monitor.rs
+++ b/src/monitor.rs
@@ -24,27 +24,6 @@ impl std::fmt::Debug for VideoModeHandle {
     }
 }
 
-impl PartialOrd for VideoModeHandle {
-    fn partial_cmp(&self, other: &VideoModeHandle) -> Option<std::cmp::Ordering> {
-        Some(self.cmp(other))
-    }
-}
-
-impl Ord for VideoModeHandle {
-    fn cmp(&self, other: &VideoModeHandle) -> std::cmp::Ordering {
-        self.monitor().cmp(&other.monitor()).then(
-            self.size()
-                .cmp(&other.size())
-                .then(
-                    self.refresh_rate_millihertz()
-                        .cmp(&other.refresh_rate_millihertz())
-                        .then(self.bit_depth().cmp(&other.bit_depth())),
-                )
-                .reverse(),
-        )
-    }
-}
-
 impl VideoModeHandle {
     /// Returns the resolution of this video mode. This **must not** be used to create your
     /// rendering surface. Use [`Window::inner_size()`] instead.
@@ -112,7 +91,7 @@ impl std::fmt::Display for VideoModeHandle {
 /// to check.
 ///
 /// [`Window`]: crate::window::Window
-#[derive(Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
+#[derive(Clone, PartialEq, Eq, Hash)]
 pub struct MonitorHandle {
     pub(crate) inner: platform_impl::MonitorHandle,
 }
--- a/src/platform_impl/windows/monitor.rs
+++ b/src/platform_impl/windows/monitor.rs
@@ -1,4 +1,4 @@
-use std::collections::{BTreeSet, VecDeque};
+use std::collections::VecDeque;
 use std::hash::Hash;
 use std::num::{NonZeroU16, NonZeroU32};
 use std::{io, mem, ptr};
@@ -91,7 +91,7 @@ impl VideoModeHandle {
     }
 }
 
-#[derive(Debug, Clone, Eq, PartialEq, Hash, PartialOrd, Ord)]
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
 pub struct MonitorHandle(HMONITOR);
 
 // Send is not implemented for HMONITOR, we have to wrap it and implement it manually.
@@ -226,7 +226,7 @@ impl MonitorHandle {
         // EnumDisplaySettingsExW can return duplicate values (or some of the
         // fields are probably changing, but we aren't looking at those fields
         // anyway), so we're using a BTreeSet deduplicate
-        let mut modes = BTreeSet::<RootVideoModeHandle>::new();
+        let mut modes = Vec::new();
         let mod_map = |mode: RootVideoModeHandle| mode.video_mode;
 
         let monitor_info = match get_monitor_info(self.0) {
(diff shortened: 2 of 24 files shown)


src/monitor.rs
```diff
@@ -24,27 +24,6 @@ impl std::fmt::Debug for VideoModeHandle {
...
-    fn cmp(&self, other: &VideoModeHandle) -> std::cmp::Ordering {
-        self.monitor().cmp(&other.monitor()).then(
-            self.size()
-                .cmp(&other.size())
-                .then(
-                    self.refresh_rate_millihertz()
-                        .cmp(&other.refresh_rate_millihertz())
-                        .then(self.bit_depth().cmp(&other.bit_depth())),
-                )
-                .reverse(),
```
madsmtm, line 43: I agree that `VideoModeHandle`'s `Ord` is weird (and likely incorrect, I could imagine two video modes that have the same monitor, size, refresh rate and bit depth, but differed on some other parameter, in which case `Ord` would say they were equal, but `Eq` would say they weren't).

madsmtm (changes requested):
I believe that there is value in `PartialOrd` and `Ord` for all of these types! Even though the ordering may not have _semantic_ meaning, it is still stable and correct (if we're comparing e.g. pointer addresses), and as said, allows using these types in `BTreeMap`/`BTreeSet`.

madsmtm (approved):
Thinking it over a bit more, I took a look at the APIs in `std`, and it seems like a fair number of them also don't implement `PartialOrd` and `Ord`, including `BTreeMap` itself.

I guess I was thinking of [the C-COMMON-TRAITS API guideline](https://rust-lang.github.io/api-guidelines/interoperability.html#types-eagerly-implement-common-traits-c-common-traits), and wanted to follow that, but even that says to implement all the _applicable_ traits.

> what the point is of being able to store it in `BTreeMap`/`BTreeSet` if the order has no meaning.

Hmm, yeah, I thought it would make the order consistent between e.g. re-runs of the same code, but that clearly can't happen if the ordering comes from a pointer.

src/platform_impl/windows/monitor.rs
```diff
@@ -226,7 +226,7 @@ impl MonitorHandle {
         // EnumDisplaySettingsExW can return duplicate values (or some of the
         // fields are probably changing, but we aren't looking at those fields
         // anyway), so we're using a BTreeSet deduplicate
-        let mut modes = BTreeSet::<RootVideoModeHandle>::new();
+        let mut modes = Vec::new();
         let mod_map = |mode: RootVideoModeHandle| mode.video_mode;
```
madsmtm, line 230: This code doesn't need to mention `RootVideoModeHandle` any more, and can be simplified even further

src/platform_impl/windows/monitor.rs
```diff
@@ -226,7 +226,7 @@ impl MonitorHandle {
         // EnumDisplaySettingsExW can return duplicate values (or some of the
         // fields are probably changing, but we aren't looking at those fields
         // anyway), so we're using a BTreeSet deduplicate
-        let mut modes = BTreeSet::<RootVideoModeHandle>::new();
+        let mut modes = Vec::new();
```
madsmtm, line 229: I think we may still need to deduplicate here?

src/platform_impl/windows/monitor.rs
madsmtm: Similarly for some of the other backends.

src/platform_impl/windows/monitor.rs
daxpedda: Ah, good catch!

src/platform_impl/windows/monitor.rs
daxpedda: Done!

src/platform_impl/windows/monitor.rs
daxpedda: Done, though I believe that we should actually not de-duplicate here but instead expose the underlying type with an extension specific trait; a matter for another time.

madsmtm (approved):
Thanks!

kchibisov (approved):

MarijnS95:
> Thinking it over a bit more, I took a look at the APIs in `std`, and it seems like a fair number of them also don't implement `PartialOrd` and `Ord`, including `BTreeMap` itself.

`Vec` neither, likely because its pointer can (also) change when pushing new items causing a `realloc()`.

> I guess I was thinking of [the C-COMMON-TRAITS API guideline](https://rust-lang.github.io/api-guidelines/interoperability.html#types-eagerly-implement-common-traits-c-common-traits), and wanted to follow that, but even that says to implement all the _applicable_ traits.

Wish `applicable` was clarified a bit more explicitly here.

> > what the point is of being able to store it in `BTreeMap`/`BTreeSet` if the order has no meaning.
> 
> Hmm, yeah, I thought it would make the order consistent between e.g. re-runs of the same code, but that clearly can't happen if the ordering comes from a pointer.

The only thing I could think of is different implementation speeds, but `HashMap` is always faster than `BTreeMap` (except amortization): https://doc.rust-lang.org/std/collections/index.html#maps.

madsmtm:
> > Thinking it over a bit more, I took a look at the APIs in `std`, and it seems like a fair number of them also don't implement `PartialOrd` and `Ord`, including `BTreeMap` itself.
> 
> `Vec` neither, likely because its pointer can (also) change when pushing new items causing a `realloc()`.

[`Vec` does implement `PartialOrd`](https://doc.rust-lang.org/std/vec/struct.Vec.html#impl-PartialOrd%3CVec%3CT,+A2%3E%3E-for-Vec%3CT,+A1%3E).

MarijnS95:
> [`Vec` does implement `PartialOrd`](https://doc.rust-lang.org/std/vec/struct.Vec.html#impl-PartialOrd%3CVec%3CT,+A2%3E%3E-for-Vec%3CT,+A1%3E).

Right, and that piggybacks off of the lexicographical ordering of the underlying slices. Not of the pointer that could change.

PR #3867, opened 2024-08-14
Remove unnecessary `#[inline]` hints

daxpedda:
As discussed in the last meeting, this PR removes unnecessary `#[inline]` hints, which hurt compile times and sometimes even performance.
To summarize:
- Adding `#[inline]` to generic functions is not recommended.
- Adding `#[inline]` to non-public functions is not recommended.
- Adding `#[inline]` to public functions because they *only* call a private function is not recommended, unless that private  function is small and is annotated with `#[inline]` as well.

`#[inline]` should only be used for **small** non-generic public functions or after some benchmarking.

See https://matklad.github.io/2021/07/09/inline-in-rust.html and https://github.com/rust-lang/hashbrown/pull/119#issuecomment-537539046 for more detailed information and recommendations.

### DIFF
--- a/dpi/src/lib.rs
+++ b/dpi/src/lib.rs
@@ -143,7 +143,6 @@ impl<P> LogicalUnit<P> {
 }
 
 impl<P: Pixel> LogicalUnit<P> {
-    #[inline]
     pub fn from_physical<T: Into<PhysicalUnit<X>>, X: Pixel>(
         physical: T,
         scale_factor: f64,
@@ -151,13 +150,11 @@ impl<P: Pixel> LogicalUnit<P> {
         physical.into().to_logical(scale_factor)
     }
 
-    #[inline]
     pub fn to_physical<X: Pixel>(&self, scale_factor: f64) -> PhysicalUnit<X> {
         assert!(validate_scale_factor(scale_factor));
         PhysicalUnit::new(self.0.into() * scale_factor).cast()
     }
 
-    #[inline]
     pub fn cast<X: Pixel>(&self) -> LogicalUnit<X> {
         LogicalUnit(self.0.cast())
     }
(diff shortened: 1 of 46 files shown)


MarijnS95:
Fwiw, in `ash`, we added `#[inline]` to all our public "high-level" wrapper APIs that wrap public "sys-like" APIs, and got quite a compile-time performance boost out of it: https://github.com/ash-rs/ash/pull/602#issuecomment-1082137017 / https://github.com/ash-rs/ash/pull/638

So I hope you're leaving public functions that call public functions/constructors in place?

dpi/src/lib.rs
```diff
@@ -143,21 +143,18 @@ impl<P> LogicalUnit<P> {
...
 
-    #[inline]
     pub fn to_physical<X: Pixel>(&self, scale_factor: f64) -> PhysicalUnit<X> {
         assert!(validate_scale_factor(scale_factor));
         PhysicalUnit::new(self.0.into() * scale_factor).cast()
     }
 
-    #[inline]
     pub fn cast<X: Pixel>(&self) -> LogicalUnit<X> {
         LogicalUnit(self.0.cast())
```
MarijnS95, line 159: Small public function calling public code? All of these, it seems like?

daxpedda:
> So I hope you're leaving public functions that call public functions/constructors in place?

Yes!
Let me know if I missed something.

dpi/src/lib.rs
daxpedda: Emphasis on **non-generic**.

MarijnS95: Let's wake up and try reading this again after a few hours :see_no_evil:.

madsmtm (approved):
Wow. that's a lot of `#[inline]` that we've been using

kchibisov (approved):

kchibisov (changes requested):
I keep my approval.

Will just like to _hold it_ until we merge like big PRs, since this one was pretty much done automatically(I hope so) and can be regenerated when needed.

PR #3826, opened 2024-07-26
chore: drop v0.30 deprecated APIs

kchibisov:
--

I think there's no reason to keep them around for the 0.31.

PR #3872, opened 2024-08-15
macOS/iOS: Remove window activation hacks

madsmtm:
No longer necessary after https://github.com/rust-windowing/winit/pull/3447 and https://github.com/rust-windowing/winit/pull/3826.

- [x] Tested on macOS
- [x] Tested on iOS
  - I did not feel it is necessary to note in the changelog, as this 

### DIFF
--- a/src/platform/ios.rs
+++ b/src/platform/ios.rs
@@ -3,15 +3,6 @@
 //! Winit has an OS requirement of iOS 8 or higher, and is regularly tested on
 //! iOS 9.3.
 //!
-//! ## Window initialization
-//!
-//! iOS's main `UIApplicationMain` does some init work that's required by all
-//! UI-related code (see issue [#1705]). It is best to create your windows
-//! inside [`ApplicationHandler::resumed`].
-//!
-//! [#1705]: https://github.com/rust-windowing/winit/issues/1705
-//! [`ApplicationHandler::resumed`]: crate::application::ApplicationHandler::resumed
-//!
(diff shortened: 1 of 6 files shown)


src/window.rs
```diff
@@ -29,8 +29,13 @@ use crate::platform_impl::{self, PlatformSpecificWindowAttributes};
 ///
 /// ## Platform-specific
 ///
-/// **Web:** The [`Window`], which is represented by a `HTMLElementCanvas`, can
-/// not be closed by dropping the [`Window`].
+/// - **Web:** The [`Window`], which is represented by a `HTMLElementCanvas`, can not be closed by
+///   dropping the [`Window`].
+/// - **macOS:** The window can only be created after the application has finished launching. This
+///   is the reason why [`EventLoop`] doesn't contain any functionality to create the window.
+/// - **iOS:** Same as macOS.
```
madsmtm, line 36: I'm unsure if this is the right place to put it, but I wanted the _motivation_ for the current design to be available somewhere.

kchibisov: It sounds like the whole design is motivated solely by macOS, where in reality it's not and window creation being inside the event loop is a bit more complex.

I'd generally just drop it, since it's more confusing then helping, the issue about window design, etc, should provide a _good_ enough reasoning.

madsmtm: Fair enough, I've removed it again

daxpedda (approved):

Issue #2962, opened 2023-07-16
Not receiving any input events on iOS


U32Float:
The event loop is not receiving any input events on iOS. However it does receive window resize/focus events.
I made a simple winit app to demonstrate this: https://github.com/U32Float/winit-ios-test

The main function looks like this:
```
let event_loop = EventLoopBuilder::new().build();
let window = WindowBuilder::new().build(&event_loop).unwrap();
event_loop.run(move |event, _, _| {
    println!("{:?}", event);
});
```
I suspect the window is created too early. But I tried creating the window after `NewEvents(Init)` and `Resumed`, but both didn't work.

Does anyone know what is going wrong?

madsmtm:
What device and OS are you using?

PR #3873, opened 2024-08-15
Improve iOS documentation

madsmtm:
Improve the platform-specific documentation with:
- [x] An updated minimum supported OS version (also on macOS)
- [x] Instructions for running with Mac Catalyst.
- [x] Instructions for setting up an Xcode project, and testing in the simulator and on-device (both using Dinghy, and without).
  - Somewhat, still not really happy with it.
- [x] Remove redundant notes about only being able to call things on the main thread.

Builds upon #3872 to avoid a rebase later on.

### DIFF
--- a/src/platform/ios.rs
+++ b/src/platform/ios.rs
@@ -5,6 +5,24 @@
 //!
 //! [rustc-ios-version]: https://doc.rust-lang.org/rustc/platform-support/apple-ios.html#os-version
 //!
+//! ## Running on Mac Catalyst
+//!
+//! Mac Catalyst allows running applications using UIKit on macOS, which can be very useful for
+//! testing. See [`rustc`'s documentation on Mac Catalyst][rustc-mac-catalyst] for details on how to
+//! use these targets. To use these with Winit, you'll need to bundle your application before
+//! running it, otherwise UIKit will exit with an error.
+//!
+//! To run e.g. the `window` example in the Winit repository, you can use [`cargo-bundle`] as
+//! follows:
+//!
+//! ```console
+//! $ cargo +nightly bundle --format=ios --target=aarch64-apple-ios-macabi --example=window
+//! $ ./target/aarch64-apple-ios-macabi/debug/examples/bundle/ios/winit.app/window
+//! ```
+//!
+//! [rustc-mac-catalyst]: https://doc.rust-lang.org/rustc/platform-support/apple-ios-macabi.html
+//! [`cargo-bundle`]: https://github.com/burtonageo/cargo-bundle
+//!
 //! ## Building app
 //!
 //! To build ios app you will need rustc built for this targets:
--- a/src/platform/ios.rs
+++ b/src/platform/ios.rs
@@ -23,43 +23,57 @@
 //! [rustc-mac-catalyst]: https://doc.rust-lang.org/rustc/platform-support/apple-ios-macabi.html
 //! [`cargo-bundle`]: https://github.com/burtonageo/cargo-bundle
 //!
-//! ## Building app
+//! ## Introduction to building an app
 //!
-//! To build ios app you will need rustc built for this targets:
+//! Building and running your application in the iOS simulator, or on a real device, is a bit more
+//! complicated than Mac Catalyst - fundamentally, you must use Xcode, since the binary needs to be
+//! bundled, signed, notarized and uploaded to the device (there is [an open source work-in-progress
+//! on re-implementing parts of this][apple-platform-rs], but the user-story around it is not yet
+//! clear).
 //!
-//!  - armv7-apple-ios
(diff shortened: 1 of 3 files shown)


src/platform/ios.rs
```diff
@@ -1,7 +1,25 @@
...
+//! Mac Catalyst allows running applications using UIKit on macOS, which can be very useful for
+//! testing. Winit somewhat supports this, though you'll need to bundle your application before
+//! running it.
+//!
+//! To run e.g. the `window` example in the Winit repository, you can use [`cargo-bundle`] as
+//! follows:
+//!
+//! ```console
+//! $ cargo +nightly bundle --format=ios --target=aarch64-apple-ios-macabi --example=window
+//! $ ./target/aarch64-apple-ios-macabi/debug/bundle/ios/window.app/window
```
madsmtm, line 19: This is not the actual command. The actual command right now is:
```suggestion
//! $ cargo +nightly build -Zbuild-std --target=aarch64-apple-ios-macabi --example window
//! $ CARGO_BUNDLE_SKIP_BUILD=1 cargo bundle --example window --target=aarch64-apple-ios-macabi --format=ios
//! $ ./target/aarch64-apple-ios-macabi/debug/bundle/ios/window.app/window
```

Since `cargo-bundle` doesn't support `-Zbuild-std` and https://github.com/rust-lang/rust/pull/126450 has yet to be merged (once that's in stable we can also remove the `+nightly` from this command).

You'll also need to use a patched version of `cargo-bundle` (or set `[package.metadata.bundle.example.window]`) until [a new version is released](https://github.com/burtonageo/cargo-bundle/pull/185).

src/platform/ios.rs
```diff
@@ -1,16 +1,25 @@
...
 //!
-//! iOS's main `UIApplicationMain` does some init work that's required by all
-//! UI-related code (see issue [#1705]). It is best to create your windows
-//! inside [`ApplicationHandler::resumed`].
+//! ## Running on Mac Catalyst
 //!
-//! [#1705]: https://github.com/rust-windowing/winit/issues/1705
-//! [`ApplicationHandler::resumed`]: crate::application::ApplicationHandler::resumed
+//! Mac Catalyst allows running applications using UIKit on macOS, which can be very useful for
+//! testing. Winit somewhat supports this, though you'll need to bundle your application before
```
daxpedda, line 11: Should we elaborate on "somewhat"?

daxpedda (approved):

src/platform/ios.rs
madsmtm: Aaand both are resolved, so with `cargo-bundle v0.6.1` and the next nightly, that is the command

src/window.rs
```diff
@@ -29,8 +29,13 @@ use crate::platform_impl::{self, PlatformSpecificWindowAttributes};
 ///
 /// ## Platform-specific
 ///
-/// **Web:** The [`Window`], which is represented by a `HTMLElementCanvas`, can
-/// not be closed by dropping the [`Window`].
+/// - **Web:** The [`Window`], which is represented by a `HTMLElementCanvas`, can not be closed by
+///   dropping the [`Window`].
+/// - **macOS:** The window can only be created after the application has finished launching. This
+///   is the reason why [`EventLoop`] doesn't contain any functionality to create the window.
+/// - **iOS:** Same as macOS.
```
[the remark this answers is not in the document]
madsmtm: See https://github.com/rust-windowing/winit/pull/3872#discussion_r1719056702, I've removed it in that branch

src/platform/ios.rs
madsmtm: I originally wrote "somewhat" because click events aren't working on Mac Catalyst, but it seems like [they aren't elsewhere either](https://github.com/rust-windowing/winit/issues/2962), so it's not a problem tied to Mac Catalyst, it's a more general problem

daxpedda: Uff.
Personally I would either drop "somewhat" or elaborate, but yeah, hairy situation.

madsmtm: I've removed "somewhat" now, thanks for the input - will request it again once I'm done writing

daxpedda (approved):
What a treasure trove of information!

While most of the backends don't need this much documentation, lets hope we can raise the standard of all the documentation in Winit to this level one day.

EDIT: I was briefly thinking that this is a bit too much for Winit. But considering how much special handling the iOS backend needs, I think a general overview of how to even run it is appropriate.
We might want to consider linking to it from the README in the future when we get the examples working for Android as well.

madsmtm:
> I was briefly thinking that this is a bit too much for Winit

I'd much prefer to link to an external resource, but I don't think such a resource exists (yet), so in Winit it is for now.

Issue #3833, opened 2024-07-27
Pointer Event Overhaul


daxpedda:
### Unified Pointer Handling

We currently handle `Touch` completely separately from mouse events (`WindowEvent::Cursor*`), the initial idea was I assume that the type doesn't really fit all the information that touch wants to express.
This would make sense but comes with a big downside: if users don't *explicitly* handle the `Touch` event, their application just won't work on touch devices.

This issue will only grow if we want to introduce pen/stylus support (#99). So to address this, we take inspiration from the [Web Pointer Events Specification](https://www.w3.org/TR/pointerevents3), which was mainly introduced to address this problem. The idea is to have a single event for all pointer input, which contains details about all possible inputs, be it touch, pen/stylus or anything else. That way users don't have to explicitly handle certain input types, but can still choose to do so, without accidentally not supporting a certain one.

Addresses #1555 (doesn't fix it, because there is a performance component to this on Wayland that would not be addressed by this proposal).
Fixes #336.

### Pen/Stylus Support

To support pen/stylus we run into the same problem as above: we can't introduce a separate event, otherwise applications will suddenly stop working when a pen is used and they don't *explicitly* handle it.

Additionally this will require adding much more data to our current `WindowEvent::Cursor*` events. So we should figure out an API that will allow us to add more pointer type support in the future while we are at it.

Addresses #99 (implementation is being done in #3810).

### Position Data in Pointer Events

We are currently missing position data in all pointer events except `WindowEvent::CursorMoved`. This isn't just annoying to the user, because they have to handle two events to get the information they want, but it requires implementations to account for that as well and send "fake" `WindowEvent::CursorMoved` events to report the position to the users.

The solution should be quite simple: add `position: PhysicalPosition<f64>` to all pointer events.

Fixes #883.
Fixes #1909.

## Implementation

- Rename `DeviceEvent::MouseMotion` to `DeviceEvent::PointerMotion` to represent any pointer device.
- Rename `WindowEvent::Cursor*` to `WindowEvent::Pointer*`. "Cursor" is more commonly used when describing the visual representation of a "Pointer", which we would like to avoid when talking about touch or pen/stylus.
- Add `position: PhysicalPosition<f64>` to `WindowEvent::PointerEntered/Left`.
- Rename `WindowEvent::MouseInput` to `WindowEvent::PointerButton`.
- Fold `WindowEvent::Touch` into pointer events:
[...]

PR #3874, opened 2024-08-16
Remove non-platform-consistent mouse events from X11

daxpedda:
Currently X11 will emit a `WindowEvent::CursorMoved` when the window comes into focus, even if e.g. the cursor isn't inside the window.
This is inconsistent with other platforms and can also be a bit misleading to users.
I believe in the future, we should use `CursorEntered/CursorLeft` on focus and unfocus *on all backends* to make this idea work properly. Emitting the cursor position in these events will be handled by #3833.

The other case, is X11 emitting a `WindowEvent::CursorMoved` on the first touch point made in a touch interaction. I believe the idea behind this is to signify the user that the cursor has jumped to this location. However, this can again be misinterpreted by users and is not consistent between platforms.
We are also unable to express this sort of pointer vs touch vs cursor with the current API, which would need its own separate discussion and motivation first.

I've encountered this while working on #3810 and #3833.


src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -1316,19 +1297,6 @@ impl EventProcessor {
...
-                let event = Event::WindowEvent {
-                    window_id,
-                    event: WindowEvent::CursorMoved {
-                        device_id: mkdid(util::VIRTUAL_CORE_POINTER),
-                        position: location.cast(),
-                    },
-                };
-                callback(&self.target, event);
-            }
-
```
kchibisov, line 1331: I think the point was that the touch moves the mouse pointer? In general, I'd _hold_ it a bit until pointer overhual since it won't break things due to things being _generic_.

Right now I'm not sure, so I'd suggest to research why this code is here in the first place, since it's clearly fixing some _not working_ thing.

kchibisov (changes requested):

src/platform_impl/linux/x11/event_processor.rs
daxpedda: Its this https://github.com/rust-windowing/winit/issues/1261.
Which has a good use-case as well.

Will remove the removal of this part.

src/platform_impl/linux/x11/event_processor.rs
```diff
@@ -1235,21 +1234,6 @@ impl EventProcessor {
...
-            .devices
-            .borrow()
-            .get(&DeviceId(xev.deviceid as xinput::DeviceId))
-            .map(|device| device.attachment)
-            .unwrap_or(2);
-
-        let event = Event::WindowEvent {
-            window_id,
-            event: WindowEvent::CursorMoved { device_id: mkdid(pointer_id as _), position },
-        };
```
daxpedda, line 1251: I tracked this down to #348.

While I do think this is a valid use-case, the current implementation is incorrect, because its also emits an event if the cursor is not inside the window.

For this to be solved we need to do it across all platforms and do it correctly.
My proposition would be to emit a `CursorEntered` when focus is gained _if_ the cursor is inside the window.

On the other hand this might also be addressed by https://github.com/rust-windowing/winit/pull/2648.

